Issue #2002, opened 2016-10-19
Allow the use of custom parsers via the CLI/Node API


kaicataldo:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

I have a custom PostCSS parser I'd like to use with stylelint, but it doesn't appear that stylelint allows for the use of custom parsers via the CLI or Node API. I understand that I can use the PostCSS plugin and make a wrapper around it if I really to do this, but it seems like something stylelint should be able to do. Ideally, it would be available through the `--syntax` flag and in a config file/object.

> Which rule, if any, is this issue related to?

N/A

> What CSS is needed to reproduce this issue?

N/A

> What stylelint configuration is needed to reproduce this issue?

N/A

> Which version of stylelint are you using?

v7.5.0

> How are you running stylelint: CLI, PostCSS plugin, Node API?

N/A

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes, it's related to custom parsers

> What did you expect to happen?

To be able to use stylelint with a custom PostCSS parser through the CLI and through the Node API.

> What actually happened (e.g. what warnings or errors you are getting)?

stylelint only allows for three predefined parsers at the moment.

If this is something the team supported I can try to look into implementing it myself.

davidtheclark:
> Ideally, it would be available through the --syntax flag and in a config file/object.

@kaicataldo: I'm interested in hearing more details about what you think would be a viable API. Also, can you please read the discussion in #1749 and share your opinion on how this relates to all that?

kaicataldo:
Reading through that linked discussion, I'm not sure I understand what the issue of allowing users to define their own parsers is, especially since the PostCSS plugin currently allows a user to run stylelint with a custom parser. It's confusing to me that it's allowed for one use case but not for the rest.

As a working example, ESLint allows alternate parsers (a very, VERY popular one is babel-eslint), but doesn't provide any guarantees that the core rules will be compatible with said parsers. I don't see the problem in taking the position that while custom parsers may be used, stylelint's core rules only support SCSS, Less, and SugarSS, and that the responsibility of creating and maintaining both custom parsers and rules compatible with said parsers lies with the community.

As far as configuration goes, I'd love to see something like the following on the command line
```
stylelint "foo/**/*.scss" --syntax "my-custom-parser"
```

and the following for the configuration file
```
{
  "syntax": "my-custom-parser",
  "rules": {
  // ...
}
```

[...]

Issue #1947, opened 2016-10-01
Rewrite non-rule tests with another test runner


davidtheclark:
While working on #1930 I've become convinced that this project is large enough that we would really benefit from some of the niceties of a more full featured test runner, and that the non-rule tests would be pretty easy to rewrite for another runner. The rule tests are another story.

I'm thinking Mocha or AVA. If somebody really wants to try Jest, I'd be open to that experiment, too. 

@evilebottnawi: I know you've expressed interest in AVA — want to give this a shot? But please wait for me to finish #1930 since I'm making significant changes to those tests in that PR.

evilebottnawi:
@davidtheclark yep. you are good developer, i love you not platonic :smile:

jeddy3:
SGTM. I'll label it up as infrastructure.

evilebottnawi:
@davidtheclark ok, let's wait

davidtheclark:
@evilebottnawi Just merged #1930 so if you'd like to try AVA, the time is ripe.

evilebottnawi:
@davidtheclark let's go, in the near future :smile:

davidtheclark:
@evilebottnawi: Are you going to tackle this? If not, I'm going to look into it soon.

evilebottnawi:
@davidtheclark I have time at the end this week. I you want start now, let's go, sorry

davidtheclark:
I rewrote tests in `src/__tests__/` with AVA and was discouraged by the results. What took 5.5 seconds with `babel-tape-runner` took *37 seconds* with AVA. 

I really like the AVA API, but that huge lag rules it out for me.

Will try again with Jest.

evilebottnawi:
@davidtheclark very strange I used to get different results in other projects, ava worked faster then tap

evilebottnawi:
@davidtheclark let's try jest

davidtheclark:
This is complete.

PR #2017, opened 2016-10-27
Add support for custom syntax/parsers

kaicataldo:
#2002

Yes! This is a WIP PR that adds the ability to use custom syntax/parsers. I have a few questions:
- **I ended up saving the output to a file and just manually searched through it. Is there a better way than that?** ~~This is my first time using tape, and I can't for the life of me figure out how to see exactly which tests are failing (there are currently 7 failures). I've tried different reporters, and none of them show me which file/test suite the failing tests are from. Am I doing something really wrong? Any info here would be appreciated. Once I get that figured out, I plan to add more tests as well.~~
- As described [here](https://github.com/postcss/postcss#syntaxes), the `stringify` for syntaxes is optional:  

> If there is not yet support for your favorite syntax, you can write a parser and/or stringifier to extend PostCSS

However, in the case that `stringify` is missing (and therefore the `syntax` is only a `parse` function (example: [PostCSS's own `postcss-safe-parser`](https://github.com/postcss/postcss-safe-parser)), the option should be set as `parser` rather than `syntax`. Your thoughts on how this should be implemented? We can do a check like I've implemented in this PR, or could be more transparent and have the user specify a `parser` option that gets passed through to PostCSS.
- This PR is rough still, but wanted to get feedback early and to see what you all think. It would be nice to be able to consolidate the checks for `syntax`. The challenge I'm facing right now is that the `configDir` doesn't always seem to be available once `getPostcssResult` has been called, and we don't have access to the file being linted when `augmentConfig` is being called (which we need to do the syntax set by file extension checking behavior). Would love to get some guidance on this from someone more knowledgeable about the project. 

Thanks!


evilebottnawi:
@kaicataldo use template please for PR

kaicataldo:
Sure thing - sorry, am using hub on the command line and didn't realize there was one.

kaicataldo:
@evilebottnawi Yeah, please read the above ^

evilebottnawi:
@kaicataldo :octocat:

kaicataldo:
Mind explaining what you mean by that?

davidtheclark:
> I ended up saving the output to a file and just manually searched through it. Is there a better way than that?

Can't you ctrl+F within the terminal? That's what I do: search for `not ok`. In #1947 we're talking about switching a chunk of our tests to a more full-featured test runner.

Looks like there might be a confusion between "options" and "config" here. "options" are the options you pass at run time, and syntax is an "option", not a property in the config. Therefore, 

- We shouldn't need the code in `augmentConfig`.
- Because it's an option, each API — in this case, the Node API and CLI — will need to accept and interpret it individually, then pass it on. The CLI should expect paths and modules relative to `process.cwd()`; the Node API should expect the user to have put together an absolute path.

Finally: I think you're on the right path with one `syntax` option instead of adding a barely different `parser` option.

Does that above help @kaicataldo?

davidtheclark:
> I ended up saving the output to a file and just manually searched through it. Is there a better way than that?

Can't you ctrl+F within the terminal? That's what I do: search for `not ok`. In #1947 we're talking about switching a chunk of our tests to a more full-featured test runner.

Looks like there might be a confusion between "options" and "config" here. "options" are the options you pass at run time, and syntax is an "option", not a property in the config. Therefore, 

- We shouldn't need the code in `augmentConfig`.
- Because it's an option, each API — in this case, the Node API and CLI — will need to accept and interpret it individually, then pass it on. The CLI should expect paths and modules relative to `process.cwd()`; the Node API should expect the user to have put together an absolute path.

Finally: I think you're on the right path with one `syntax` option instead of adding a barely different `parser` option.

Does that above help @kaicataldo?

kaicataldo:
@davidtheclark Thanks, very helpful! Apologies, still not 100% sure about what the differences between "configs" and "options" are. We discussed in the issue about allowing both

```
stylelint "foo/**/*.scss" --syntax "my-custom-parser"
```

and the following for the configuration file
```
{
  "syntax": "my-custom-parser",
  "rules": {
  // ...
}
```

Was the intention rather to only allow it as a flag but not in the configuration file for the CLI option? Thanks for your help on this :)

davidtheclark:
"config" = the configuration file that is loaded, or a configuration object passed in as a `config` option.
"options" = options that you pass to CLI via flags, or to Node API via an object.

I think for this PR let's just leave `syntax` as an option, not a configuration property. We can discuss adding a configuration property as a separate issue.

kaicataldo:
Great, thanks for clearing that up :+1:

src/getPostcssResult.js
```diff
@@ -43,13 +44,29 @@ export default function (
...
       } else if (syntax) {
-        throw new Error("You must use a valid syntax option, either: scss, less or sugarss")
+        try {
+          // Should check these possibilities:
+          //   a. name of a node_module
+          //   b. absolute path
+          //   c. relative path relative to `process.cwd()`.
+          // If none of the above work, we'll try a relative path starting
+          // in `process.cwd()`.
+          const syntaxModule = resolveFrom(process.cwd(), syntax)
```
kaicataldo: This is the same logic used in `cli.js` to resolve the configFile module. Would you prefer this was extracted into a utility method somewhere, or is the duplication okay?

kaicataldo:
Updated the PR. Do you have any tests set up for the CLI/Node API? Maybe I'm just missing something, but I was only able to find tests for `standalone`. Thanks for the guidance on this :)

kaicataldo:
One more question - would it make sense to allow the Node API to accept the following in addition to a string representing the path to the syntax module?

src/__tests__/standalone-syntax-test.js
```diff
@@ -133,3 +133,87 @@ test("standalone with syntax set by extension", t => {
...
+    },
+  }
+
+  standalone({
+    config,
+    syntax: `${fixturesPath}/custom-parser`,
+    code: ".foo { width: 200px; }\n.bar {}",
+    formatter: stringFormatter,
+  }).then(({ output }) => {
+    const strippedOutput = chalk.stripColor(output)
```
davidtheclark: Instead of reading the `output`, stripping the colors, and using `indexOf`, I suggest using `.then(({ results }) => {`: `results` will be the JSON stylelint results.

src/__tests__/standalone-syntax-test.js
```diff
@@ -133,3 +133,87 @@ test("standalone with syntax set by extension", t => {
...
+    t.equal(typeof output, "string")
+    t.ok(strippedOutput.indexOf("2:6") !== -1)
+    t.ok(strippedOutput.indexOf("block-no-empty") !== -1)
```
davidtheclark: What about this output assures us that the custom parser was used? How would this test fail if stylelint *did not* use the custom parser?

src/cli.js
```diff
@@ -10,6 +10,17 @@ import path from "path"
...
+  // Should check these possibilities:
+  //   a. name of a node_module
+  //   b. absolute path
+  //   c. relative path relative to `process.cwd()`.
+  // If none of the above work, we'll try a relative path starting
+  // in `process.cwd()`.
+  return resolveFrom(process.cwd(), moduleName)
+    || path.join(process.cwd(), moduleName)
+}
+
```
davidtheclark: We should make `getModulePath()` from `augmentConfig.js` its own module, then use it both there and here.

src/__tests__/standalone-syntax-test.js
```diff
@@ -133,3 +133,87 @@ test("standalone with syntax set by extension", t => {
...
+    t.end()
+  }).catch(t.end)
+})
+
+test("standalone with npm-installed custom syntax", t => {
```
davidtheclark: I think both these `with npm-installed` tests can be removed, thereby removing the need to have these installed as `devDependencies`. As long as the code uses `getModulePath()` (see comment below), we can be confident that it will find npm dependencies.

src/cli.js
```diff
@@ -128,18 +139,11 @@ if (cli.flags.quiet) {
 }
 
 if (cli.flags.syntax) {
-  optionsBase.syntax = cli.flags.syntax
+  optionsBase.syntax = resolveModule(cli.flags.syntax)
```
davidtheclark: Won't we have a problem here if `cli.flags.syntax` is one of the standard syntaxes, like `"scss"`?

src/getPostcssResult.js
```diff
@@ -43,13 +43,21 @@ export default function (
...
+        } catch (e) {
+          throw new Error(`Cannot resolve syntax module ${syntax}`)
+        }
       }
 
       const postcssOptions = {
-        syntax,
         from: options.filePath,
       }
+      if (syntax && !syntax.stringify) {
```
davidtheclark: Please add a comment explaining why this move is made.

src/cli.js
kaicataldo: Since the built-in syntax options are dependencies, it should resolve to the same dependency. If that doesn't sound like a good plan, we can either put a guard here for the built-in syntaxes or use the extracted module resolution function to resolve the module in `getPostcssResult` instead (my preference).

src/getPostcssResult.js
```diff
@@ -43,13 +43,21 @@ export default function (
       } else if (syntax === "sugarss" || !syntax && fileExtension === ".sss") {
```
kaicataldo: If we don't move the module resolution logic into here the above can just be `if (!syntax && fileExtension === ".sss")`

src/cli.js
kaicataldo: Sounds good

src/__tests__/standalone-syntax-test.js
kaicataldo: If the syntax module can't be found, it should throw. However, we could use something like the `postcss-safe-parser` and have it try to parse something that the regular postcss parser can't.

kaicataldo:
Thanks for the review - will get back to this shortly!

src/cli.js
davidtheclark: But `scss` or `less` are not going to resolve to the `postcss-scss` and `postcss-less` modules, right?

src/__tests__/standalone-syntax-test.js
davidtheclark: > we could use something like the postcss-safe-parser and have it try to parse something that the regular postcss parser can't.

Yeah, that's what I think we should do. Set up a situation so that there will be a parsing problem if the right parser is not used; or the linting output will change. Because we're not just testing whether the syntax module is *found*, but whether it is used to change the parsing and the eventual output.

src/getPostcssResult.js
davidtheclark: Have you considered the PostCSS plugin use-case — where somebody inserts stylelint as a PostCSS plugin, using neither the Node API nor the CLI?

kaicataldo: Ah oh course, good point

src/cli.js
kaicataldo: You're totally right :+1:

kaicataldo:
Updated once more. Appreciate the guidance on this :+1:

davidtheclark:
Excellent! Next step to finishing this up is to add documentation for the new feature.

@stylelint/core Any other feedback on this new feature?

evilebottnawi:
@kaicataldo can you squash commits in one for easy review and good git history?

kaicataldo:
Updated with documentation. Please let me know if any revisions need to be made or if I missed anything.

jeddy3:
> @stylelint/core Any other feedback on this new feature?

It's an exciting new feature. @kaicataldo Thank you for contributing it!

I'm a bit late to the party (v busy of late), but I have query about the UX... what's the rationale behind overloading the `syntax` option, rather than adding a new `custom-syntax` option? I think from a user's perspective having consistency with the `formatter` and `custom-formatter` option would be beneficial, where`formatter` is for the _built-in_ `string`, `verbose` and `json` formatters.

davidtheclark:
^^ That's a good idea @jeddy3 that didn't come to my mind. @kaicataldo what do you think about making that change?

kaicataldo:
It looks like the `custom-formatter` option is resolved in CLI (i.e. if you use the Node API, you would have to just set it as `formatter`?).  Is this the behavior you're thinking of? Or would it just be passed through to `standalone`? Also, what is the expected behavior if `--syntax` and `--custom-syntax` are set to different values? Would the custom syntax override the syntax option?

I can work on this, but would like to get this hammered out before I go through and rewrite the documentation, etc, since I have some other OSS projects that also need attention!

davidtheclark:
@kaicataldo A couple of answers:

- Looks like we need to add `customFormatter` to the the Node API. That's an oversight. Will handle in another issue.
- If `formatter` and `custom-formatter` are both provided, `custom-formatter` overrides: https://github.com/stylelint/stylelint/blob/master/src/cli.js#L116. Same should go for `syntax` and `custom-syntax`.

kaicataldo:
Updated. I took a different approach from `--custom-formatters` and passed through `custom--syntax` even on the CLI side to allow for more useful error messaging. Please let me know what you think.

jeddy3:
@kaicataldo Thank you for contributing this feature and for the tackling the last minute changes!

Added to changelog as: 

> -   Added: option `customSyntax` (for Node API) and `--custom-syntax` (for CLI).

kaicataldo:
Thanks for this project and for all the guidance on this PR! :heart:

kaicataldo:
As a follow up, was curious about your release cycle and when I might be able to use this new feature - thanks again!

jeddy3:
@kaicataldo I reckon we can release today. David's in the States and so usually checks-in in a few hours time. I'll just need to double check with him there's nothing stopping a release, but hopefully we can release later today :)

jeddy3:
@kaicataldo Released as `7.6.0`. Thanks again for contributing :)

kaicataldo:
Fantastic - thanks!

PR #2020, opened 2016-10-28
Add font-family-no-duplicate-names rule

### LINKED ISSUE #1284: Add font-family-no-duplicate-names rule

A new rule to detect duplicate font names used in `font` or `font-family` properties

`font-family-no-duplicate-names`

> The following patterns are considered warnings:

`a { font-family: 'Times', Times, "serif", serif; }`
`b { font: normal 14px/32px -apple-system, BlinkMacSystemFont, sans-serif, sans-serif;`

> And the following patterns are **not** considered warnings:

`a { font-family: Times, serif; }`
`b { font: normal 14px/32px -apple-system, BlinkMacSystemFont, sans-serif;`

gaidarenko:

Closes #1284 

### DIFF
--- a/src/rules/font-family-no-duplicate-names/README.md
+++ b/src/rules/font-family-no-duplicate-names/README.md
@@ -1,0 +1,47 @@
+# font-family-no-duplicate-names
+
+Disallow duplicate font family names.
+
+```css
+a { font-family: serif, serif; }
+/**              ↑      ↑
+ * These font family names */
+```
+
+This rule checks the `font` and `font-family` properties.
+
+This rule ignores `$sass`, `@less`, and `var(--custom-property)` variable syntaxes.
+
+**Caveat:** This rule will stumble on *unquoted* multi-word font names and *unquoted* font names containing escape sequences. Wrap these font names in quotation marks, and everything should be fine.
+
+## Options
+
+### `true`
+
+The following patterns are considered warnings:
+
+```css
+a { font-family: 'Times', Times, serif; }
+```
+
+```css
+a { font: 1em "Arial", 'Arial', sans-serif; }
+```
+
+```css
+a { font: normal 14px/32px -apple-system, BlinkMacSystemFont, sans-serif, sans-serif; }
+```
+
+The following patterns are *not* considered warnings:
+
+```css
+a { font-family: Times, serif; }
+```
+
+```css
+a { font: 1em "Arial", "sans-serif", sans-serif; }
+```
+
+```css
+a { font: normal 14px/32px -apple-system, BlinkMacSystemFont, sans-serif; }
+```
--- a/src/rules/font-family-no-duplicate-names/__tests__/index.js
+++ b/src/rules/font-family-no-duplicate-names/__tests__/index.js
@@ -1,0 +1,58 @@
+﻿import {
+  messages,
+  ruleName,
+} from ".."
+import rules from "../../../rules"
+import { testRule } from "../../../testUtils"
+
+const rule = rules[ruleName]
+
+testRule(rule, {
+  ruleName,
+  config: [true],
+
+  accept: [ {
+    code: "a { font-family: \"Lucida Grande\", \"Arial\", sans-serif; }",
+  }, {
+    code: "a { font: 1em \"Lucida Grande\", \'Arial\', sans-serif; }",
+  }, {
+    code: "a { font: 1em \"Lucida Grande\", \'Arial\', \"sans-serif\", sans-serif; }",
+  }, {
+    code: "a { font-family: Times, serif; }",
+  }, {
+    code: "b { font: normal 14px/32px -apple-system, BlinkMacSystemFont, sans-serif; }",
+  } ],
+
+  reject: [ {
+    code: "a { font-family: \"Lucida Grande\", \'Arial\', sans-serif, sans-serif; }",
+    message: messages.rejected("sans-serif"),
+    line: 1,
+    column: 56,
+  }, {
+    code: "a { font-family: \'Arial\', \"Lucida Grande\", Arial, sans-serif; }",
+    message: messages.rejected("Arial"),
+    line: 1,
+    column: 44,
+  }, {
+    code: "a { fOnT-fAmIlY: \'  Lucida Grande \', \"Lucida Grande\", sans-serif; }",
+    message: messages.rejected("Lucida Grande"),
+    line: 1,
+    column: 38,
+  }, {
+    code: "a { font-family: \'Times\', Times, \"serif\", serif; }",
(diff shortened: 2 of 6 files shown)


src/rules/font-family-no-duplicate-names/index.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+        else {
+          const family = fontFamilyNode.value.trim()
+
+          if (familyNames.has(family)) {
+            complain(messages.rejected(family), declarationValueIndex(decl) + fontFamilyNode.sourceIndex, decl)
+          }
+          else {
+            familyNames.add(family)
+          }
```
davidtheclark: Looks like the only difference between these two blogs is the value assigned to `family`. The code could probably be simplified significantly, right?

gaidarenko: @davidtheclark Not only `family` but also different `Set`s to store already found font family names. I use `keywords` to store found font family keywords (e.g. `sans-serif`, `cursive`, `monospace` ...). And I use `familyNames` to store other font family names. All this done in order to allow `a { font-family: "sans-serif", sans-serif; }` and don't throw an error.

src/rules/font-family-no-duplicate-names/README.md
```diff
@@ -0,0 +1,47 @@
...
+a { font-family: "Times New Roman", serif; }
+/**               ↑                 ↑
+ *               These font family names */
+```
+
+This rule checks the `font` and `font-family` properties.
+
+This rule ignores `$sass`, `@less`, and `var(--custom-property)` variable syntaxes.
+
+**Caveat:** This rule does not currently understand escape sequences. If you want to use the font family name "Hawaii 5-0" you will need to wrap it in quotes, instead of escaping it as `Hawaii \35 -0` or `Hawaii\ 5-0`.
```
davidtheclark: > If you want to use ...

This is inaccurate, right? You do not need to wrap it in quotes to use it. What is the actual consequence related to this rule of not wrapping the name in quotes?

src/rules/font-family-no-duplicate-names/index.js
davidtheclark: Ok, I guess I wasn't looking closely enough. I still think this code can be simplified, though. Please try to match the conventions in other parts of the codebase for brace spacing and early returns, and see if you can this up a little.

src/rules/font-family-no-duplicate-names/README.md
gaidarenko: Not using quotes brings up a lot of ambiguity. For example, <pre> `Lucida      Grande`</pre> and <pre>`Lucida Grande`</pre> are one and the same valid family font name from CSS point of view. But this rule will treat these fonts as different font names. On the other hand, <pre>`"Lucida     Grande"`</pre> and <pre>`"Lucida Grande"`</pre> are different font names from both CSS and this rule point of view.

gaidarenko:
I have simplified the code to use early returns.

src/rules/font-family-no-duplicate-names/README.md
davidtheclark: Ok, so it sounds like the caveat is not about "escape sequences" but about quotation marks around multi-word font names?

src/rules/font-family-no-duplicate-names/index.js
```diff
@@ -0,0 +1,66 @@
...
+        if (isFamilyNameKeyword(fontFamilyNode)) {
+          const family = fontFamilyNode.value.toLowerCase()
+
+          if (keywords.has(family)) {
+            complain(messages.rejected(family), declarationValueIndex(decl) + fontFamilyNode.sourceIndex, decl)
+            return
+          }
+
+          keywords.add(family)
+        }
```
davidtheclark: Can return early here instead of the else statement below.

src/rules/font-family-no-duplicate-names/README.md
gaidarenko: One can say so. The main idea - Please use quotation marks around multi-word font names or around font names containing disallowed symbols. Rather than use escape sequences in unquoted font family names. Because this rule does not currently support escape sequences.

Here some quotes from w3.org site:

> Font family names other than generic families must either be given quoted as strings, or unquoted as a sequence of one or more identifiers. This means most punctuation characters and digits at the start of each token must be escaped in unquoted font family names.

> To avoid mistakes in escaping, it is recommended to quote font family names that contain white space, digits, or punctuation characters other than hyphens

davidtheclark: > This rule will stumble on *unquoted* multi-word font names and *unquoted* font names containing escape sequences. Wrap these font names in quotation marks, and everything should be fine.

How about ^^?

gaidarenko: Ok. Should I keep `Caveat` word?

davidtheclark: Yep, to match the format from other docs.

gaidarenko:
Done.

davidtheclark:
👍  from me. @jeddy3 any input, or should we merge?

evilebottnawi:
@davidtheclark what about add option `ignore`, sometimes we need this https://github.com/necolas/normalize.css/blob/master/normalize.css#L93

src/rules/font-family-no-duplicate-names/README.md
```diff
@@ -0,0 +1,47 @@
+# font-family-no-duplicate-names
+
+Disallow duplicate font family names.
+
+```css
+a { font-family: "Times New Roman", serif; }
```
jeddy3:
```css
a { font-family: serif, serif; }
/**              ↑      ↑
 * These font family names */
```

src/rules/font-family-no-duplicate-names/__tests__/index.js
```diff
@@ -0,0 +1,58 @@
+﻿import {
+  messages,
+  ruleName,
+  } from ".."
```
jeddy3: `}` should be flush to the left.

jeddy3:
> @jeddy3 any input, or should we merge?

Two very minor requests. Otherwise, looks good to me. @gaidarenko Thanks for yet another rule! :)

> what about add option ignore, sometimes we need this 

This what disable comment are for i.e. one-off instances (e.g. hacks or workarounds). The `ignore` option is there for _repeating_ patterns that occur within a codebase.

evilebottnawi:
@jeddy3 agree

gaidarenko:
Done.

jeddy3:
> -   Added: `font-family-no-duplicate-names` rule.

Issue #2016, opened 2016-10-27
Use Jest snapshot testing on a suite of config-stylesheet pairs


davidtheclark:
To protect against regressions, stylelint might benefit from having a collection of config-stylesheet test cases for end-to-end tests. For each test case, the test would be straightforward: Given this config and this stylesheet, what are the stylelint results? We would never be able to make this collection complete and systematic; it will always be a "last line of defense" effort. We'd gradually fill it with any examples we find valuable, especially of regressions that have occurred in the past.

In the past I've considered this idea unrealistic, because it would involve hand-maintaining the stylelint result objects for every test case. Lately, though, I've been using [Jest](https://facebook.github.io/jest/) for React snapshot testing, and have been thinking that snapshot tests might make these end-to-end tests a real possibility. Creating and updating snapshots is very easy.

What would this look like?
- To start with, we'd ask users (and ourselves) to provide us with config-stylesheet pairs that they want to make sure don't regress. We'd be especially interested in config-stylesheet pairs that have suffered regressions in the past. This would be the beginning of the collection.
- From then on, as users run into regressions, _once the bug is fixed_ we can ask for a config-stylesheet pair to add to our suite.

Since the addition of a new test case to the suite would be so painless, we'd have little reason not to pile them up. And the more we have, the more protected we are against regressions, the more certain we are that the code is working as intended.

What do you think?

evilebottnawi:
@davidtheclark sgtm, can you provide first example in PR?

ntwb:
Sgtm, watched the Jest video and peeked at some of [the examples](https://github.com/facebook/jest/tree/master/examples) 👍

jeddy3:
> We'd gradually fill it with any examples we find valuable, especially of regressions that have occurred in the past.

Excellent.

> In the past I've considered this idea unrealistic, because it would involve hand-maintaining the stylelint result objects for every test case.

Agreed.

> We'd be especially interested in config-stylesheet pairs that have suffered regressions in the past. This would be the beginning of the collection.

We'd definitely throw one of the rules that takes an array as a primary option into one of the first config-stylesheet pairs! :)

> From then on, as users run into regressions, once the bug is fixed we can ask for a config-stylesheet pair to add to our suite.

This seems like the most sensible way of building the suite.

> Since the addition of a new test case to the suite would be so painless, we'd have little reason not to pile them up. And the more we have, the more protected we are against regressions, the more certain we are that the code is working as intended.

[...]

PR #2023, opened 2016-10-30
System tests

davidtheclark:
Definitely not ready to merge yet. But I'm opening a PR so I can see how Travis responds ...


davidtheclark:
Ref #2016.

This is actually looking ok ...

Here's the idea so far:
- `system-tests/` holds everything relevant to this.
- `system-tests/.eslintrc` simply adds the Jest env to the config.
- We add a subfolder for each test case, containing a `stylesheet.*`, a `config.*`, and a `[case number].test.js`. The test file checks stylelint's output against the snapshot.

Case 002 experiments with a config with dependencies — in this case, `stylelint-scss`. What I have going here works but it feels wrong and ugly to me. There's a separate `package.json` and `node_modules` inside `system-tests/` to manage these config dependencies, in order to avoid circular dependency problems when one of those plugins depends on `stylelint` (as `stylelint-scss` does). This means that we need to `cd system-tests && npm install` before we run the system tests. **If anybody has another idea about how to accomplish this, I'd love to hear it.**

ntwb:
A couple of notes/questions regarding `system-tests/001/stylesheet.css` if I may:

• The file is Bootstrap v4.0.0-alpha.5
• I believe Bootstrap v4 is still currently using `scss_lint`
• I believe Bootstrap v4 does not currently use CSSLint.

Firstly, is is that entire file valid CSS? (I presume it is, but lets for a moment assume it isn't)

What I'm curious about in regard to validity of that CSS file is as these Jest tests are to _protect against regressions_ if there are any issues with that Bootstrap CSS file then stylelint will be perpetually testing against _possible_ invalid CSS.

If in a future not too far from now stylelint v8.2.5 fixes "a bug of some sort" then how would that affect the above config-stylesheet pair as the `001.test.js` and `001.test.js.snap` are not using a specific semantic version of stylelint and the returned `.snap` object results would now be different to that of the original commit.

1. Rather than large libraries of CSS such as Bootstrap, should these test pairs be smaller succinct examples of 10, 50, 100 lines of CSS rather than thousands?

2. Should, which leads to can each of these tests be against a specific version of stylelint? ( I'm already envisaging even more circular dependency problems with this if that were the case)

p.s. Will think and research some ideas regarding the extra `package.json` NPM implementation

davidtheclark:
@ntwb Thanks for putting some thought into this!

The file is built from Bootstrap's SCSS, so I think it is valid CSS. But regardless, I'm not sure it matters. The point of the snapshot tests is to ensure that stylelint's output remains consistent, even if that output is a syntax error. If stylelint changes such that this Bootstrap CSS file no longer produces the same results, we want to know.

The snapshots will update as stylelint updates, but only if the result changes are *deliberate*. You can see here that the snapshot is committed to the repo: so if we *update* the snapshots, because a bug was fixed and some test case's output changed *in a good way*, that update will be included in a PR, diffed and reviewed.

Because of this, I think we always want the snapshot tests to run against the stylelint code under development, not against fixed versions of stylelint. We want to make sure that the code that we are about to commit does not break the expected output from these test cases.

evilebottnawi:
@davidtheclark what about install all packages for tests from `package.json` for latest snapshot of code?

davidtheclark:
> what about install all packages for tests from package.json for latest snapshot of code?
@evilebottnawi if what you're suggesting is use the project's `package.json` and add plugins etc. as devDependencies, I believe the problem is that some of those plugins etc. will have a dependency on stylelint, in which case we'll have a problem.

ntwb:
I think `peerDependencies` could work here, I still don't fully grok `peerDependencies` so I'll continue researching this, currently reading:
https://stackoverflow.com/questions/26737819/why-use-peer-dependencies-in-npm-for-plugins

m-allanson:
@davidtheclark this is looking great.

> If anybody has another idea about how to accomplish this, I'd love to hear it.

Would adding a `postinstall` npm script be enough to handle the `cd system-tests && npm install` command?

Another option could be to install the `system-test` dependencies as devDependencies in stylelint's main `package.json`. Then use [Jest's manual mocks](https://facebook.github.io/jest/docs/manual-mocks.html#content) to override any nested stylelint dependencies.

This means that, with an npm dependency structure like this:

```
@stylelint@(development version)
└─┬ @stylelint-scss
  └── @stylelint@(released version)
```

any `stylelint-scss`references to the released version of `stylelint` would be replaced with the development version of `stylelint`. Does that sound crazy? Note that I've not tried it and I don't know if it would actually work.

davidtheclark:
> I think peerDependencies could work here

@ntwb Don't think `peerDependencies` would work in this situation. We'd still end up in `stylelint` having a `node_modules/stylelint` directory, which is problematic.

> Would adding a postinstall npm script be enough to handle the cd system-tests && npm install command?

@m-allanson I don't think we want to do `postinstall` because we don't want these installing all the time, not for end users.

```
@stylelint@(development version)
└─┬ @stylelint-scss
  └── @stylelint@(released version)
```

^^ Seems kind of crazy to me, more crazy than having an install-within-in-install.

davidtheclark:
Well, I think this is ready for more trials. @stylelint/core can you please point me to some code, in repos or gists, that you think we should add to this system test suite?

ntwb:
One config-stylesheet pair I thought would be beneficial, though artificial in that it would more than likely need to be created specifically for this would be for [custom messages](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#custom-messages). 

I was thinking the config-stylesheet pair would only require basic CSS but a thorough configuration with a custom message for all stylelint rules that support them.

Is this the type of config-stylesheet pair that would be a good fit here?

davidtheclark:
@ntwb That couldn't hurt! I don't think, though, you'd need to include a custom message for *every* rule, because the logic to apply custom messages lives outside the rules — therefore, if a few work, we're pretty assuming they'll all work.

davidtheclark:
If these tests pass, I'm going to merge what I have. I think we can best avoid the double-install mess by just *not testing external tools*. It is not the responsibility of this library to test, for example, that `stylelint-scss` works as expected. So I think we just don't include test cases that rely on plugins, etc., that in turn depend on stylelint.

Issue #1974, opened 2016-10-11
Increase Flow coverage


davidtheclark:
We recently introduced [Flow](http://flowtype.org/) types to some of the files in `src/`. We should try to gradually extend that coverage to more of the app, focusing first on the non-rules code and then seeing how we can start type-checking all the rules, as well.

Anybody who is interested in learning how to use Flow is welcome to submit PR's of any size adding annotations anywhere you feel like poking around :)

m-allanson:
Consider me interested :) Although realistically it'll be about a week until I have a chance to dig in. I'd also be interested to see PRs from anyone else that wants to work on this.

ntwb:
Likewise, very interested, but alas I've not much available time at the moment.

Will look forward to checking out any PRs submitted 😄

m-allanson:
PR #2007 is a useful reference for anyone looking at this issue.

PR #2028, opened 2016-11-01
Add Flow definitions to atRule utils

m-allanson:

#1974 

This is a very small PR to confirm that I'm doing the right thing.

### DIFF
--- a/src/utils/atRuleParamIndex.js
+++ b/src/utils/atRuleParamIndex.js
@@ -1,10 +1,11 @@
+/* @flow */
 /**
  * Get the index of a media query's params
  *
  * @param {AtRule} atRule
  * @return {int} The index
  */
-export default function (atRule) {
+export default function (atRule: postcss$atRule) {
   // Initial 1 is for the `@`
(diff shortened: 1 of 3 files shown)


src/utils/atRuleParamIndex.js
```diff
@@ -1,10 +1,11 @@
+/* @flow */
 /**
  * Get the index of a media query's params
  *
  * @param {AtRule} atRule
  * @return {int} The index
  */
-export default function (atRule) {
+export default function (atRule: postcss$atRule) {
```
davidtheclark: It would also be good to specify the return values on these functions.

m-allanson: Done

m-allanson: Ah that makes sense, thanks. I'll get this updated.

m-allanson: This is updated and rebased now

davidtheclark:
👍

PR #2029, opened 2016-11-02
Chore: use `isStandardSyntaxRule` instead pure check.

evilebottnawi:

no

No, it's self explanatory.

/cc @davidtheclark


davidtheclark:
Thanks @evilebottnawi.

PR #2030, opened 2016-11-02
Chore: use `raws` instead `raw()`.

evilebottnawi:

no

No, it's self explanatory.

We don't use `raw()` because it is call `stringify` and return `stringifier` value, it is potential error in all non-standard syntax, special `sugarss`. Related commit: https://github.com/stylelint/stylelint/commit/f4bd57a0c4aaa0faae36eb71a88a956102497d19 .

/cc @davidtheclark @jeddy3 
I think we should add note about this in docs.

### DIFF
--- a/src/rules/at-rule-empty-line-before/index.js
+++ b/src/rules/at-rule-empty-line-before/index.js
@@ -68,7 +68,7 @@ export default function (expectation, options) {
       if (optionsMatches(options, "ignore", "after-comment")
         && isAfterComment()) { return }
 
-      const hasEmptyLineBefore = hasEmptyLine(atRule.raw("before"))
+      const hasEmptyLineBefore = hasEmptyLine(atRule.raws.before)
       let expectEmptyLineBefore = (expectation === "always") ? true : false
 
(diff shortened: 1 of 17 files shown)


davidtheclark:
@evilebottnawi Looks good.

> I think we should add note about this in docs.

Feel free to add a note wherever you think it will be helpful. Plugin development guide?

evilebottnawi:
@davidtheclark yep, i am bad in English, @jeddy3 can you help me?

davidtheclark:
@evilebottnawi How about adding to the Best Practices list on http://stylelint.io/developer-guide/rules/ this point:

> - When accessing raw strings from the PostCSS AST, use `node.raws` instead of `node.raw()`. This will ensure string corresponds exactly to the original

evilebottnawi:
@davidtheclark sound good to me, let's wait @jeddy3 , maybe he decides something else to add

jeddy3:
> How about adding to the Best Practices list on http://stylelint.io/developer-guide/rules/ this point:

SGTM. Perfect place for it.

PR #2033, opened 2016-11-05
Deprecate declaration-block-properties-order

### LINKED ISSUE #2010: Deprecate declaration-block-properties-order in favour of more specific community plugins

I want to test my Sass (scss) without having the properties being in a set order.

declaration-block-properties-order

``` json
{
  "rules": {
    "declaration-block-properties-order": [
      {
        order: "flexible",
        properties: [
          "position",
          "top",
          "right",
          "bottom",
          "left"
        ]
      },
      {
        order: "flexible",
        properties: [
          "flex",
          "flex-direction",
          "flex-order",
          "flex-pack",
          "flex-align"
        ]
      },
      {
        order: "flexible",
        properties: [
          "margin",
          "margin-top",
          "margin-right",
          "margin-bottom",
          "margin-left"
        ]
      },
      {
        order: "flexible",
        properties: [
          "font",
          "font-family",
          "font-size",
          "font-style",
          "font-weight"
        ]
      },
      {
        order: "flexible",
        properties: [
          "background",
          "background-color",
          "background-image",
          "background-repeat",
          "background-attachment",
          "background-position",
          "background-position-x",
          "background-position-y",
          "background-clip",
          "background-origin",
          "background-size",
        ]
      },
      {
        order: "flexible",
        properties: [
          "transition",
          "transition-delay",
          "transition-timing-function",
          "transition-duration",
          "transition-property",
        ]
      },
      {
        order: "flexible",
        properties: [
          "animation",
          "animation-name",
          "animation-duration",
          "animation-fill-mode",
          "animation-play-state",
          "animation-timing-function",
          "animation-delay",
          "animation-iteration-count",
          "animation-direction"
        ]
      }
    ]
  }
}
```

"gulp-stylelint": "3.4.0",

Nope

Code to flag, but with the order in each properties: [...] above (not in the listed order).

Flags loads of errors, because the code is not written in the order of each block above.
## 

In an ideal world, I want the code to check through and validate the code. However I can't expect all my development team to remember the order.

E.G. must use background before transition but only after the flex code. As an example I don't mind where the flex and positioning are placed however when someone codes positioning. I want it to go:

Pass linting:

``` sass
position: absolute;
top: 0;
right: 0;
bottom:
left: 0;
```

Flag in NPM:

``` sass
right: 0;
top: 0;
bottom:
position: absolute;
left: 0;
```

Many thanks.

jeddy3:
Closes #2010

### DIFF
--- a/docs/user-guide/plugins.md
+++ b/docs/user-guide/plugins.md
@@ -4,30 +4,18 @@ Plugins are rules and sets of rules built by the community that support methodol
 
 ## Code style
 
--   [`stylelint-declaration-block-order`](https://github.com/hudochenkov/stylelint-declaration-block-order): Specify the order of groups of at-rules, nested rules, custom properties, declarations and dollar variables in declaration blocks.
--   [`stylelint-property-groups-structure`](https://github.com/hudochenkov/stylelint-property-groups-structure): Require or disallow an empty line before user-defined property groups and specify the order of the properties within these groups (incorporates the `declaration-block-properties-order` rule).
+-   [`stylelint-order](https://github.com/hudochenkov/stylelint-order): A plugin pack of rules for specifying the ordering of things e.g. properties within declaration blocks.
 
 ## Methodologies
 
 -   [`stylelint-rscss`](https://github.com/rstacruz/stylelint-rscss): Validate [RSCSS](http://rscss.io) conventions.
 -   [`stylelint-selector-bem-pattern`](https://github.com/davidtheclark/stylelint-selector-bem-pattern): Specify a BEM pattern for selectors (incorporates [postcss-bem-linter](https://github.com/postcss/postcss-bem-linter)).
 
+## Non-standard syntax
+
+-   [`stylelint-declaration-use-variable`](https://github.com/sh-waqar/stylelint-declaration-use-variable): Specify properties for which a SCSS variable must be used.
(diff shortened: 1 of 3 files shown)


docs/user-guide/plugins.md
```diff
@@ -4,30 +4,18 @@ Plugins are rules and sets of rules built by the community that support methodol
 
 ## Code style
 
--   [`stylelint-declaration-block-order`](https://github.com/hudochenkov/stylelint-declaration-block-order): Specify the order of groups of at-rules, nested rules, custom properties, declarations and dollar variables in declaration blocks.
--   [`stylelint-property-groups-structure`](https://github.com/hudochenkov/stylelint-property-groups-structure): Require or disallow an empty line before user-defined property groups and specify the order of the properties within these groups (incorporates the `declaration-block-properties-order` rule).
+-   [`stylelint-order](https://github.com/hudochenkov/stylelint-order): A plugin pack of rules for specifying the ordering of things e.g. properties within declaration blocks.
```
hudochenkov: Syntax error here.

davidtheclark:
I think we need a deprecation error in the code. People will not be checking up on this text, so if we don't add one in the code the removal will still come as a surprise.

jeddy3:
> I think we need a deprecation error in the code. 

Agreed. I've reverted the changelog change, so we can do this in another issue.

This PR is now just about updating the release planning and tidying up the plugin list.

davidtheclark:
👍

Issue #1341, opened 2016-05-20
Create Code Climate engine


ai:
If Code Climate will move to Stylelint it will be awesome for promotion.

I think @davidtheclark should write a mail to Code Climate team, explain benefits of Stylelint (compare to csslint) and offer any help :) (I will help you they will ask something complicated).

jeddy3:
They are looking for a community contribution, via the [developer program](https://codeclimate.com/dev_form). 

Ref: https://twitter.com/a_baldwinhunter/status/733747324267167749

davidtheclark:
I guess at some point someone has to write that integration for them ...

jeddy3:
I'm hoping that that someone is a person or team who uses both Code Climate and stylelint, and has a need to fill the hole :) 

Labelled as "help wanted" accordingly.

ntwb:
Via https://github.com/codeclimate/codeclimate-eslint/issues/80#issuecomment-223386512

> > ntwb wrote: 
> > "Over here stylelint/stylelint#1341 we have been thinking/wanting to write a stylelint engine for Code Climate, maybe not replace the existing CSSLint CSS Code Climate engine, but to start with as an alternative for sure 😏"

> Code Climate wrote:
> "That's fantastic! We've chatted about stylelint here at Code Climate HQ and it would be great to have an engine available. stylelint would make a great native engine! Let me @dblandin or @mrb know if you have any more questions regarding the engine submission process."
>
> "We have a codeclimate-community slack organization where members hang out to discuss engine ideas and development. If you're interested, you can join our Developer Program to get an invite."
 

👍 
***

A slight caveat on the use of stylelint shared plugins and configs with Code Climate:

>  > > Code Climate Wrote:
> > > "Unlike many popular CI systems, our build infrastructure isn't yet designed to execute customer code. It was designed to run engines statically against source code. Before we allow arbitrary execution of customer code (including NPM packages or RubyGems), we have some work to do to ensure that our systems and customer data remain secure. Unfortunately, this isn't an issue specific to ESLint. It's a shared concern across our platform."

> > ntwb wrote:
> > "If we were to write a stylelint Code Climate engine stylelint would also be affected by the above in that Code Climate wouldn't allow us to load stylelint shared plugins or shared configs to be used with the stylelint engine correct?"

> Code Climate wrote:
[...]

PR #2062, opened 2016-11-16
Chore: add `codeclimate` badge.

evilebottnawi:

No, it's self explanatory.

@stylelint/core what do you think add new label? BTW perhaps we should pay attention to the report


davidtheclark:
Honestly, I'm not interested in what CodeClimate has to say. I clicked through some of the reportrs, and did not find any criticisms that were accurate or suggestions that were helpful. Did you?

evilebottnawi:
@davidtheclark I noticed a lot of duplicated code message, i think that some of the architectural highlights can be reviewed and processed in a more user-friendly and readable form

davidtheclark:
> I noticed a lot of duplicated code message

Yes, but the ones I looked at were not actionable. They were for code that *should* be duplicated, for which an abstraction would be overkill. But, yeah, I'm sure there are some cases where this is not the case.

evilebottnawi:
@davidtheclark I think it would be great if anyone has analyzed all the warnings, but I understand that we have little time for it. Also I think that `codeclimate` have configuration files and after analize maybe good create this file and take into account the subsequent warning

evilebottnawi:
@davidtheclark I think it would be great if anyone has analyzed all the warnings, but I understand that we have little time for it. Also I think that `codeclimate` have configuration files and after analize maybe good create this file and take into account the subsequent warning

ntwb:
I'm pretty sure it was myself who originally queued up the stylelint repo analysis at Code Climate. 

To have any kind of accurate analysis by Code Climate quite a few things need to happen, some of those are listed in #1341, whitelisting by Code Climate of our `eslint-config-stylelint` which was denied in https://github.com/codeclimate/codeclimate-eslint/issues/80, adding a `.codeclimate.yml` with a configuration suitable for stylelint to name but a few.

Per the above I'm going to close this, as adding the badge alone is *not* an accurate reflection of the status of the stylelint repo. 😄

PR #2081, opened 2016-11-19
Add more Flow definitions to utils

m-allanson:

#1974

This adds Flow type definitions to a few more of the util files.

### DIFF
--- a/src/utils/beforeBlockString.js
+++ b/src/utils/beforeBlockString.js
@@ -1,3 +1,4 @@
+/* @flow */
 /**
  * Given a CSS statement, return the string before the block.
  * For rules, this is the selector list (and surrounding whitespace).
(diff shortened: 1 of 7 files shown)


src/utils/beforeBlockString.js
```diff
@@ -10,24 +11,26 @@
...
  * @return {string}
  */
-export default function (statement, { noRawBefore } = {}) {
-  let result = ""
-  if (statement.type !== "rule" && statement.type !== "atrule") { return result }
+export default function (
+  statement: Object,
+  { noRawBefore }: { noRawBefore: ?boolean} = {}
+): string {
+  let result: string = ""
```
davidtheclark: This is ok, but a little excessive. No need to annotate variables, unless they are of a special type.

m-allanson:
Noted, thank you :) As I work through the rest of these, I'll just look at annotating function signatures and any 'unusual' variables.

PR #2083, opened 2016-11-19
Jest tests for src/__tests__

davidtheclark:
- Replaces all the tape tests in `src/__tests__` with Jest tests.
- Changes some configuration to accommodate these tests.
- Makes a couple of very small Flow changes that turned out to be irrelevant, but why not keep them?

The Jest tests don't run quite as quickly as tape tests; but there are some benefits that I think are significant:

- The watch CLI is fantastic. It is very easy to choose which tests to run repeatedly.
- The before/after hooks encourage more isolated assertions, which is good.
- Diffs are very nice when using `toEqual`.
- When a test fail or something else goes wrong, it's very easy to figure out what happened.
- Built-in support for Promises makes tests more simple, easier to read, and less prone to author error that makes debugging difficult.

This porting was pretty tedious. I wonder if there is a codemod we can use to port the `src/utils` tests.


jeddy3:
Oh wow, this is amazing!

> The watch CLI is fantastic

Agreed, it's an amazing developer experience and joy to use. I've had the pleasure of using it on a project for the last couple of weeks.

For anyone else testing this branch, use `npm run jest -- --watch` to trigger the interactive watch mode.

> but there are some benefits that I think are significant:

Agreed, the other benefits definitely outweigh any performance difference.

> I wonder if there is a codemod we can use to port the src/utils tests.

I was reading this "[Migrating to jest](https://medium.com/@kentcdodds/migrating-to-jest-881f75366e7e#.sguo5q4x3)" article the other day, which mentioned [this codemod](https://github.com/skovhus/jest-codemods). Perhaps that will help with the `utils` tests.

Exciting PR that LGTM :)

m-allanson:
> I was reading this "Migrating to jest" article the other day, which mentioned this codemod. Perhaps that will help with the utils tests.

I've just tried this, it looks promising but the codemod doesn't support `t.plan` so those calls will need to be removed manually. I can push up a PR once that's done.

davidtheclark:
Forgot to mention: the distinction between tests that should be read by Jest and those that should not is that I started suffixing Jest tests with `.test.js`. So `integration.test.js` should be for Jest, whereas `integration-test.js` should be for tape.

ntwb:
Cool, Have spent my Sunday night not in jest 😉 

I've created https://github.com/stylelint/stylelint-config-standard/pull/50 using the _codemod_ referenced above to switch from Ava to Jest, with bonus `npm watch` task, this is the coolest _watch_ task I've seen 👍

ntwb:
From the when in Rome department, whilst adding `jest` as an additional ESLint `env` in `eslint-config-stylelint` I switched the tests from Tape to Jest: https://github.com/stylelint/eslint-config-stylelint/pull/11

PR #2085, opened 2016-11-19
Jest tests for utils

m-allanson:

Refs PR #2083

This is a PR to merge into the `new-test-runner` branch, not `master`.


src/utils/__tests__/atRuleParamIndex.test.js
```diff
@@ -1,21 +1,18 @@
...
-    t.equal(atRuleParamIndex(atRule), 8)
+    expect(atRuleParamIndex(atRule)).toBe(8)
   })
   atRules("@import\n'x.css');", atRule => {
-    t.equal(atRuleParamIndex(atRule), 8)
+    expect(atRuleParamIndex(atRule)).toBe(8)
   })
   atRules("@document url-prefix(http://www.w3.org/Style/)", atRule => {
-    t.equal(atRuleParamIndex(atRule), 10)
+    expect(atRuleParamIndex(atRule)).toBe(10)
```
davidtheclark: We are only lucky if this works ... Each call to `atRules` is asynchronous, cannot be treated synchronously, like this.

Instead, I'd suggest something like this: 

- Modify the `atRules` util so it's less fancy. It should just return the Promise that resolves with a PostCSS result.
- Change the test to `describe` and put each assertion within its own `it` statement.
- For each assertion, return the Promise from `atRules`, following by `.then(result => /* do assertion here */)`.

src/utils/__tests__/beforeBlockString.test.js
```diff
@@ -0,0 +1,48 @@
...
+  expect(postcssCheck("a {}")).toBe("a ")
+  expect(postcssCheck("\na\n{}")).toBe("\na\n")
+  expect(postcssCheck("\n\na,\nb,\n\tspan > .foo\n{}")).toBe("\n\na,\nb,\n\tspan > .foo\n")
+})
+
+it("beforeBlockString at-rules", () => {
+  expect(postcssCheck("@media print {}")).toBe("@media print ")
+  expect(postcssCheck("\n@media print, screen\n\t{}")).toBe("\n@media print, screen\n\t")
+  expect(postcssCheck("@supports (animation-name: test) {}")).toBe("@supports (animation-name: test) ")
+  expect(postcssCheck(
```
davidtheclark: I don't see an assertion here ... oh, now I see it. Could you put longer strings like this in variables so the `expect(..).toBe(..)` is more apparent?

src/utils/__tests__/blurInterpolation.test.js
```diff
@@ -0,0 +1,16 @@
...
+
+it("blurInterpolation", () => {
+  expect(blurInterpolation("#{$selector}")).toBe(" $selector ")
+  expect(blurInterpolation("#{$selector}", "`")).toBe("`$selector`")
+  expect(blurInterpolation("#{$selector * 10px}")).toBe(" $selector * 10px ")
+  expect(blurInterpolation("#{$font-size}/#{$line-height}")).toBe(" $font-size / $line-height ")
+  expect(blurInterpolation("url(#{$selector * 10px})")).toBe("url( $selector * 10px )")
+  expect(blurInterpolation("calc(#{$selector} * 2)")).toBe("calc( $selector  * 2)")
+  expect(blurInterpolation(
+    "filter: progid:DXImageTransform.Microsoft.gradient(enabled='false', startColorstr='#{ie-hex-str($green)}', endColorstr='#{ie-hex-str($translucent-red)}');")).toBe(
```
davidtheclark: Same thing about the visibility of `expect(..).toBe(..)`.

src/utils/__tests__/declarationValueIndex.test.js
```diff
@@ -1,24 +1,21 @@
...
   })
   rules("a { a  : b }", decl => {
-    t.equal(declarationValueIndex(decl), 5)
+    expect(declarationValueIndex(decl)).toBe(5)
   })
   rules("a { a:\nb }", decl => {
-    t.equal(declarationValueIndex(decl), 3)
+    expect(declarationValueIndex(decl)).toBe(3)
   })
 })
```
davidtheclark: These tests have the same async issue as mentioned above.

src/utils/__tests__/isKeyframeRule.test.js
```diff
@@ -1,40 +1,37 @@
...
   })
   rules("@media print { a {} }", rule => {
-    t.notOk(isKeyframeRule(rule), "@media")
+    expect(isKeyframeRule(rule)).toBeFalsy()
   })
   rules("@supports (animation-name: test) { a {} }", rule => {
-    t.notOk(isKeyframeRule(rule), "@supports")
+    expect(isKeyframeRule(rule)).toBeFalsy()
   })
 })
```
davidtheclark: Same async problem.

src/utils/__tests__/isStandardSyntaxAtRule.test.js
```diff
@@ -2,58 +2,55 @@ import isStandardSyntaxAtRule from "../isStandardSyntaxAtRule"
...
 import postcss from "postcss"
 import scss from "postcss-scss"
-import test from "tape"
-
-test("isStandardSyntaxAtRule", t => {
-  t.plan(12)
 
+it("isStandardSyntaxAtRule", () => {
   atRules("@charset UTF-8;", atRule => {
-    t.ok(isStandardSyntaxAtRule(atRule), "non nested at-rules without quotes")
```
davidtheclark: In all cases where we *have* descriptions for assertions, we should preserve them. This means changing the `it()` above to `describe()`, and then wrapping these assertions in `it("description", () => {..})`.

src/utils/__tests__/isStandardSyntaxDeclaration.test.js
```diff
@@ -0,0 +1,89 @@
...
+  lessRules("@var: b", decl => {
+    expect(isStandardSyntaxDeclaration(decl)).toBeFalsy()
+  })
+  lessRules("a { @var: b }", decl => {
+    expect(isStandardSyntaxDeclaration(decl)).toBeFalsy()
+  })
+})
+
+function rules(css, cb) {
+  postcss().process(css).then(result => {
```
davidtheclark: Async problem.

src/utils/__tests__/isStandardSyntaxFunction.test.js
```diff
@@ -1,25 +1,22 @@
...
+    expect(isStandardSyntaxFunction(func)).toBeTruthy()
   })
 
   rules("a { $list: (list) }", func => {
-    t.notOk(isStandardSyntaxFunction(func), "scss list")
+    expect(isStandardSyntaxFunction(func)).toBeFalsy()
   })
 
   rules("a { $map: (key: value) }", func => {
-    t.notOk(isStandardSyntaxFunction(func), "scss map")
```
davidtheclark: Losing descriptions.

m-allanson:
@davidtheclark Very good points, thanks for catching those. I've added [one more commit](https://github.com/stylelint/stylelint/pull/2085/commits/968a559f19d2c7ce9e88e650262ea4971f0853b9) with an example of handling the async tests properly. Does that look like a viable pattern to use?

If you're running the code locally you can uncomment the `delayPromise` code to verify that tests are only passing after the promises have resolved.

davidtheclark:
@m-allanson That looks great to me!

m-allanson:
@davidtheclark I'm gradually amending these files. Once finished I'll close this PR and open a fresh one into `master`. In the meantime, let me know if you spot any issues with my recent commits here.

davidtheclark:
@m-allanson Would you mind pointing me at specific lines you might have questions about? That's a lot of commits, and there are a lot of tests.

m-allanson:
@davidtheclark Having just tried to pick out some parts now, I think this will be easier once I've finished and opened a new PR. I should have something up later this week.

PR #2095, opened 2016-11-22
Add `env` values to Travis CI

ntwb:
I've added some Travis CI environment variables to the build matrix.

I've added these as "hacky docs labels"  as when investigating Travis CI builds there is are no "labels" per se describing what each of the 5 jobs are, at the moment differentiating between which of the 3 NodeJS jobs are running what task (lint+flow, jest, or test) is guess work:

no-envs

By adding these "hacky environment labels" we can now identify which npm task is ran by each job a little quicker:

envs

p.s. These have no effect on the actual Travis CI jobs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,16 @@ matrix:
   include:
     - node_js: '6'
       script: npm run lint && npm run flow
+      env: CI=lint+flow
     - node_js: '6'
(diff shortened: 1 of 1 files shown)


davidtheclark:
@ntwb Good idea. Can you rebase then merge?

Issue #2093, opened 2016-11-21
Parse error for nested declarations in selector-max-compound-selectors


maxmx:
In release 7.6.0 

The following sass which used to be valid in 7.5.0 now returns a `Cannot parse selector` error
```css
.foo {
  margin: {
    left: 5px;
    top: 10px;
  };
}
```

> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
    "at-rule-name-case": "lower",
    "at-rule-name-space-after": "always-single-line",
    "at-rule-semicolon-newline-after": "always",
    "block-closing-brace-empty-line-before": "never",
    "block-closing-brace-newline-after": "always",
    "block-closing-brace-newline-before": "always-multi-line",
    "block-closing-brace-space-before": "always-single-line",

    "block-opening-brace-newline-after": "always-multi-line",
    "block-opening-brace-space-after": "always-single-line",
    "block-opening-brace-space-before": "always",
    "color-hex-case": "lower",
    "color-no-invalid-hex": true,
    "comment-empty-line-before": [ "always", {
      except: ["first-nested"],
      ignore: ["stylelint-commands"],
    } ],
    "comment-no-empty": true,
    "comment-whitespace-inside": "always",
    "custom-property-empty-line-before": [ "always", {
      except: [
        "after-custom-property",
        "first-nested",
      ],
      ignore: [
        "after-comment",
        "inside-single-line-block",
      ],
    } ],
    "declaration-bang-space-after": "never",
    "declaration-bang-space-before": "always",
    "declaration-block-no-duplicate-properties": true,
    "declaration-block-no-ignored-properties": true,
    "declaration-block-no-redundant-longhand-properties": true,
    "declaration-block-no-shorthand-property-overrides": true,
    "declaration-block-semicolon-newline-after": "always-multi-line",
    "declaration-block-semicolon-space-after": "always-single-line",
    "declaration-block-semicolon-space-before": "never",
    "declaration-block-single-line-max-declarations": 1,
    "declaration-block-trailing-semicolon": "always",
    "declaration-colon-newline-after": "always-multi-line",
    "declaration-colon-space-after": "always-single-line",
    "declaration-colon-space-before": "never",

    "font-family-name-quotes": "always-where-required",
    "function-calc-no-unspaced-operator": true,
    "function-comma-newline-after": "always-multi-line",
    "function-comma-space-after": "always-single-line",
    "function-comma-space-before": "never",
[...]

PR #2102, opened 2016-11-22
Fixed: false positives Scss selector with nested properties.

evilebottnawi:

https://github.com/stylelint/stylelint/issues/2093

No, it's self explanatory.

I believe that the rule does not work correctly and we should fix this in `8.0.0`.

Example:
```scss
.foo .bar > .foobar { & .test { @media screen { color: red; } } }
```

We this line use for check on nested

```if (rule.nodes.some(node => node.type === "rule" || node.type === "atrule")) { return }```

And we get:
1. We use `walkRules`.
2. Here we have 2 selector `.foo .bar > .foobar` and `& .test`
    First selector have `nodes` and return.
    Second selector have `nodes` and return.
But we should get error here.

Also rule don't work with nested properties.

/cc @davidtheclark @jeddy3

### DIFF
--- a/src/rules/selector-max-compound-selectors/__tests__/index.js
+++ b/src/rules/selector-max-compound-selectors/__tests__/index.js
@@ -170,10 +170,13 @@ testRule(rule, {
   config: [1],
   syntax: "scss",
 
-  accept: [{
+  accept: [ {
     code: "#hello #{$test} {}",
     description: "ignore rules with variable interpolation",
-  }],
+  }, {
+    code: ".foo { margin: { left: 5px; top: 10px; }; }",
+    description: "nested properties",
+  } ],
 })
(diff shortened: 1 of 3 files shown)


evilebottnawi:
strange, some tests failed

jeddy3:
As this addresses the immediate parse error, I think we should merge this.

@evilebottnawi Can you create _two_ new issues to discuss your other findings; one for `.foo .bar > .foobar { & .test { @media screen { color: red; } } }` and another for nested properties.

davidtheclark:
Added note to changelog:

> -   Fixed: `selector-max-compound-selectors` ignores SCSS nested declarations ([#2102](https://github.com/stylelint/stylelint/pull/2102))

Issue #2048, opened 2016-11-10
Add --allow-empty-input CLI option


nikolasleblanc:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

On a brand new project setup with stylelint as part of its linting process, stylelint fails when it can't find any css files. On a new project, it seems safe to assume there may not be any css files just yet. Right now, it seems the best way to avoid this is to add a dummy css file, but that seems unnecessary.

`Error: Files glob patterns specified did not match any files`

> Which rule, if any, is this issue related to?

None.

> What CSS is needed to reproduce this issue?

Literally, none. (Run stylelint on a project without css files.)

> What stylelint configuration is needed to reproduce this issue?

Any, this is related to the glob process.

> Which version of stylelint are you using?

7.5.0

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No.

> What did you expect to happen?

Expect stylelint to simply realize that there are no files to process and either indicate as much and slink away quietly or just slink away quietly without any hoopla.

> What actually happened (e.g. what warnings or errors you are getting)?

`Error: Files glob patterns specified did not match any files`

davidtheclark:
There is no way for stylelint to know that you are working on a new project or that you actually intend your glob *not to work*. And there is good reason to throw this error: most of the time, if your glob does not match any files it is because you've made a mistake. It would be frustrating to find out that when you thought your linting had gone well you were actually just linting nothing. You want to know when your glob is inaccurate.

If you don't have any files to lint, you should not run stylelint. Once there are files to lint, point to them with the glob and run stylelint.

nikolasleblanc:
If the concern here is that a user may confuse a misdirected glob with a clean and linted set of stylesheets, I'm not sure blowing up the rest of their linting toolchain is the most graceful way to handle it. 

Could there not simply be a message on the command line indicating "0 files found" before exiting politely? 

If I run tslint, for example, against `src/**/*.ts` in a project without ts files let alone a src folder, it doesn't throw, it exits politely. If I run karma on a project that doesn't have any tests, it doesn't blow up. It tells me I have no tests.

I'd like to have a reliable, reusable toolchain at my disposal before I write a single line of code, to have to reintroduce stylelint _only after I've written my first stylesheet_ just doesn't seem sound.

davidtheclark:
[...]

PR #2104, opened 2016-11-22
Allow empty input for --allow-empty-input argument in CLI

finico:
Issue: #2048 Add --allow-empty-input CLI option

With `--allow-empty-input` option when there is no input will prevent exit code 80

### DIFF
--- a/src/__tests__/standalone.test.js
+++ b/src/__tests__/standalone.test.js
@@ -90,6 +90,28 @@ it("standalone with non-existent-file should throw error with code 80", () => {
   })
 })
 
+describe("standalone with non-existent-file and allow-empty-input", () => {
+  it("should not throw error", () => {
+    expect(() => standalone({
+      files: `${fixturesPath}/non-existent-file.css`,
+      allowEmptyInput: true,
+    })).not.toThrow()
+  })
+
+  it("should resolve empty results", () => {
+    return standalone({
+      files: `${fixturesPath}/non-existent-file.css`,
+      allowEmptyInput: true,
+    }).then(data => {
+      expect(data).toEqual({
+        errored: false,
+        output: "[]",
+        results: [],
+      })
+    })
+  })
+})
+
 describe("standalone passing code with syntax error", () => {
   let results
 
--- a/src/cli.js
+++ b/src/cli.js
@@ -107,6 +107,10 @@ const meowOptions = {
         If you provide the argument "error", the process will exit with code 2
         if needless disables are found.
 
+      --allow-empty-input
+
+        Prevent a non-zero exit code when there is no input.
+
       --version, -v
 
         Show the currently installed version of stylelint.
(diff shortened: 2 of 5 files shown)


finico:
@davidtheclark I agree
I changed the behavior so that there was no throw

src/cli.js
```diff
@@ -201,7 +211,14 @@ Promise.resolve().then(() => {
...
   process.stdout.write(output)
   if (errored) { process.exitCode = 2 }
-}).catch(err => {
-  console.log(err.stack) // eslint-disable-line no-console
-  process.exit(err.code || 1)
+}).catch(({ code, message, stack }) => {
+  const exitCode = code === undefined ? 1 : code
+
+  if (code > 0) {
+    process.stdout.write(`${stack}${EOL}`)
```
evilebottnawi: Also i don't think good idea use `stdout` instead `console.log`. See https://github.com/nodejs/node/blob/master/lib/console.js#L43 . `log` wrapper on stdout and work fine in all `os` without `EOL` package.

finico: code no longer exists

evilebottnawi: @finico ohh, sorry, I looked at the wrong place :smile:

evilebottnawi: @finico squash your commits in one for best history and convenient review in future :+1:

finico: > log wrapper on stdout

I know, but `stdout` is already present in `cli.js` and it seemed to me as a convention
https://github.com/stylelint/stylelint/blob/master/src/cli.js#L196
https://github.com/stylelint/stylelint/blob/master/src/cli.js#L202

evilebottnawi: @finico Thanks for catch this.
@davidtheclark Why do not we use `console.log` here?

finico: @evilebottnawi I will squash it after approvea.
For best review convenient I keep commits after "request changes" and mark it "(fixup)" for squash in future.

davidtheclark: Feel free to switch to `console.log`.

src/__tests__/standalone.test.js
```diff
@@ -90,6 +90,13 @@ it("standalone with non-existent-file should throw error with code 80", () => {
   })
 })
 
+it("standalone with non-existent-file and allow-empty-input should not throw error", () => {
+  expect(() => standalone({
+    files: `${fixturesPath}/non-existent-file.css`,
+    config: { allowEmptyInput: true },
+  })).not.toThrow()
```
davidtheclark: Let's also assert the expected outcome. What is the `results` and `output` in `standalone().then(({ results, output })`?

finico:
Strange and unstable lint in the TravisCI...
`CHANGELOG.md` not changed by me

src/__tests__/standalone.test.js
finico: @davidtheclark I added a test for that

jeddy3:
> Strange and unstable lint in the TravisCI... CHANGELOG.md not changed by me

My bad. Please rebase from `master` to get the linting fix.

jeddy3:
Done in https://github.com/stylelint/stylelint/pull/2117

finico:
@jeddy3 So it's almost a repetition of my code with a delay in two days. Why my request is a duplicate?

jeddy3:
@finico There were a lot of stylelint issues and PRs to work through this morning and I went from top to bottom. I encountered the other one first, and merged it as I'd forgotten about this one. Sorry about that. The good news is that the option is in `master` now.

davidtheclark:
Agh, sorry @finico! I have been jumping in and out of things kind of quickly of late, and did not notice the duplication.

PR #2106, opened 2016-11-23
fix(package): update table to version 4.0.1

ntwb:
Superseeds #2101

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "stylehacks": "^2.3.0",
     "sugarss": "^0.2.0",
     "svg-tags": "^1.0.0",
-    "table": "^3.7.8"
+    "table": "^4.0.1"
   },
   "devDependencies": {
     "babel-cli": "^6.1.18",
--- a/src/formatters/stringFormatter.js
+++ b/src/formatters/stringFormatter.js
@@ -1,4 +1,4 @@
-import table, { getBorderCharacters } from "table"
+import { getBorderCharacters, table } from "table"
 import _ from "lodash"
 import chalk from "chalk"
 import path from "path"


ntwb:
I'm not sure what is going on with Table v4.x.x to be honest, there is no changelog and all 4.x.x [Table Travis CI build jobs have failed](https://travis-ci.org/gajus/table/builds).

Here's a compare of the last 3.x.x to 4.x.x branch: https://github.com/gajus/table/compare/v3.8.3...v4.0.1

Will need to look into that some more for further investigation

(If no one beats me to it I'll have more time tomorrow to investigate)

evilebottnawi:
@ntwb maybe cache?

ntwb:
@evilebottnawi I don't think so, cache maintenance for Travis CI was added back in https://github.com/stylelint/stylelint/commit/ebfec8a488f6f03e1383389e223bc8ef9a7af117 and not touched in AppVeyor. Also this PR is a fresh PR I created for Table v4.0.1, #2101 was for v4.0.0 and #2103 didn't have any issues.

davidtheclark:
We need to expel Table from our dependencies. I don't like having a dependency that does not include a changelog. Let's find another library.

evilebottnawi:
`eslint` use https://github.com/substack/text-table in `stylish` formatter, but it is seem old library and many tests failed

evilebottnawi:
I am create issue let's wait what says contributors

gajus:
> I create issue, let's wait what says contributors

The only breaking change is how you import table.

See https://github.com/gajus/table/issues/25#issuecomment-262209270

In your code base, you simply need to replace:

```js
import table from 'table';
// or
// const table = require('table').default;
```

with 

```js
import {table} from 'table';
// or
// const {table} = require('table');
```

gajus:
Sorry about the lack of changelog. I don't like the https://github.com/semantic-release/semantic-release tool (because it uses github's releases instead of a static changelog file) and there isn't a feasible alternative at the moment.

Please suggest how to add automated changelog to the CI https://github.com/gajus/table/issues/27

davidtheclark:
@gajus If you don't have an automated changelog you need to have a manual one. That's what we do.

gajus:
> @gajus If you don't have an automated changelog you need to have a manual one. That's what we do.

You, personally, don't appear to maintain 50+ open-source projects. Therefore, we have different problems to solve.

In most of the projects I use semantic-release. However, I have learned that migrating to other repositories (e.g. gitlab) renders the release log useless.

On other projects I have started to use a custom script (https://github.com/gajus/gitinfo/blob/master/.scripts/release.sh), which is working great, but it is in experimental state. I am planning to abstract a variation of it and release it as a tool. As soon as thats out of the door, I will enforce it on table as well as all other packages.

gajus:
> You, personally, don't appear to maintain 50+ open-source projects. Therefore, we have different problems to solve.

Dammit. Said this without looking at the profile. Well, kudos for not being sloppy. :-)

I will do my best to remedy table ASAP.

ntwb:
I restarted the failing Travis CI job and all jobs now pass, thanks @evilebottnawi for the commit 👍

davidtheclark:
@gajus I understand budgeting your energy! Consider mine a vote for changelog a even at the expense of something else. I personally consider it a priority in dependencies that are not super simple. It can save me (the user) a lot of headaches and uncertainty. Anyway, thanks for your work on these OSS projects!

PR #2117, opened 2016-11-24
[2048] - Adds cli argument to fail silently when no files matched

nikolasleblanc:

#2048 

FIrst pass at this, welcome comments on coding conventions, styles etc.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -24,6 +24,7 @@ const minimistOptions = {
     rd: "report-needless-disables",
     s: "syntax",
     v: "version",
+    aei: "allow-empty-input",
   },
 }
 
@@ -101,6 +102,10 @@ const meowOptions = {
 
         Force enabling/disabling of color.
 
+      --allow-empty-input, -aei
+
+        If no files match glob pattern, exits without throwing an error.
+
       --report-needless-disables, --rd
 
         Report stylelint-disable comments that are not blocking a lint warning.
@@ -168,6 +173,10 @@ if (cli.flags.ignorePath) {
 
 if (cli.flags.ignoreDisables) {
   optionsBase.ignoreDisables = cli.flags.ignoreDisables
+}
+
+if (cli.flags.allowEmptyInput) {
+  optionsBase.allowEmptyInput = cli.flags.allowEmptyInput
 }
 
 const { reportNeedlessDisables } = cli.flags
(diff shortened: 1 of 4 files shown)


src/cli.js
```diff
@@ -24,6 +24,7 @@ const minimistOptions = {
     rd: "report-needless-disables",
     s: "syntax",
     v: "version",
+    aei: "allow-empty-input",
```
davidtheclark: No need to create an alias if it's not going to be documented.

src/cli.js
```diff
@@ -101,6 +102,10 @@ const meowOptions = {
 
         Force enabling/disabling of color.
 
+      --allow-empty-input
+
+        If no files match glob pattern, exits gracefully.
```
davidtheclark: Can you please be more precise than "gracefully"?

nikolasleblanc:
I've  been spoiled by Webstorm's auto saving, did this in sublime and never noticed I hadn't saved my alias change to cli.js. Also, rephrased the description as per recommendation. I can be _more_ specific, just trying to balance it with brevity.

jeddy3:
@nikolasleblanc Thank you.

Added to the changelog as:

> -   Added: `allow-empty-input` flag to CLI ([#2117](https://github.com/stylelint/stylelint/pull/2117))

alexgorbatchev:
I can't seem to find any references to `allowEmptyInput` in the code any more other than type defs (https://github.com/stylelint/stylelint/search?q=allowEmptyInput&unscoped_q=allowEmptyInput)

same with grepping through `node_modules`

```
$ ag "allowEmptyInput"
stylelint/lib/cli.js
103:    allowEmptyInput: any,
134:  allowEmptyInput?: any,

stylelint/flow-typed/stylelint.js
149:  allowEmptyInput?: boolean,
```

Using the flag in the command line also makes no difference, I get exit code 0 no matter what

```
$ stylelint not-existing-file.scss && echo $?
0
$ stylelint not-existing-file.scss --allow-empty-input && echo $?
0
```

Am I missing something obvious?

jeddy3:
@alexgorbatchev `allowEmptyInput` [was removed](https://github.com/stylelint/stylelint/pull/2464) in `8.0.0` in favour of allowing empty input by default.

> Am I missing something obvious?

Nope, it wasn't obvious as the CLI `--help` still lists the command. I'd forgotten about the change in behaviour myself too (see https://github.com/stylelint/stylelint/issues/3056).

I've [pushed up a PR](https://github.com/stylelint/stylelint/pull/3400) to remove the option from the CLI `--help` and type defs. I suspect these were reintroduced during some conflict resolution.

PR #2118, opened 2016-11-24
Handle non-integer error codes as exit code 1 (#2115)

### LINKED ISSUE #2115: Exit code 0 when runtime error happened in loaded javascript configuration

When you have a `stylelint.config.js` looking like this:

```javascript
module.exports = require("./someUnknownStuff.js");
```

and you don't have `someUnknownStuff.js`, then stylelint cli breaks, obviously, and shows and error. But returning 0 (success) exit code, so build on jenkins is succeeded, even though configuration is wrong.

Not related to any rule

No CSS involved

```javascript
//styleling.config.js
module.exports = require("./someUnknownStuff.js");
```

`7.6.0`, and verified problem on current `master` as well

CLI

No

Error code 1 or other that != 0;

Error code 0

nkabardin:
Fixes #2115 

When there is an error in require() in configuration file, cli gets an error code which is not actually integer, so passing it to process.exit is not a good idea.

Obvious soultion here is just ensure that code that we got is integer, and we can do that with just covering it with parseInt, which will give the integer back if it's actually an integer, and will return NaN if it's a string, so exit code will default to `1`, exactly what we need.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "globjoin": "^0.1.4",
     "html-tags": "^1.1.1",
     "ignore": "^3.2.0",
-    "known-css-properties": "^0.0.5",
+    "known-css-properties": "^0.0.6",
     "lodash": "^4.0.0",
     "log-symbols": "^1.0.2",
     "meow": "^3.3.0",


src/cli.js
```diff
@@ -203,5 +203,5 @@ Promise.resolve().then(() => {
   if (errored) { process.exitCode = 2 }
 }).catch(err => {
   console.log(err.stack) // eslint-disable-line no-console
-  process.exit(err.code || 1)
+  process.exit(parseInt(err.code) || 1)
```
davidtheclark: How about checking if `typeof err.code === "number")? I think it might be more transparent than `parseInt`.

nkabardin:
@davidtheclark no I did not try that, but I'll do that and add a system test here.

and yes, I was thinking that it might be better to write it more explicit, so will do that as well

Issue #2067, opened 2016-11-17
Cross-reference in docs the whitespace rules that relate to each other


davidtheclark:
As suggested by @jtfairbanks at https://github.com/stylelint/stylelint/issues/2060#issuecomment-260693121.

This would mean, for example, adding a link from `declaration-colon-newline-after` to `declaration-colon-space-after`, mentioning that `"declaration-colon-space-after": "never"` is the only way to force *no* space.

jeddy3:
Closing [as stylistic rules are frozen](https://github.com/stylelint/stylelint/issues/5674).

Issue #1464, opened 2016-06-14
Update remark


wooorm:
Hi! Thanks for using remark. 👍

There’s a new version of remark out, and most importantly the CLI is externalised.  For your usage, I believe the only change needed in `package.json` is:

```diff
     "postcss-import": "^8.0.2",
-    "remark": "^4.0.0",
-    "remark-lint": "^3.0.0",
+    "remark-cli": "^1.0.0",
+    "remark-lint": "^4.0.0",
     "request": "^2.69.0",
```

If you can’t seem to get it to work, [Gitter is a good place to get help](https://gitter.im/wooorm/remark).

👋

wooorm:
Oh and before I forget I suggest checking out if [`remark-validate-links`](https://github.com/wooorm/remark-validate-links) might be a handy addition, it checks if links point to existing headings / files in the docs.

evilebottnawi:
@wooorm Thanks for feedback.

jeddy3:
> Hi! Thanks for using remark. 👍

Thanks for making remark! It's a life-saver for us :)

ntwb:
Fixed via https://github.com/stylelint/stylelint/commit/0143a221d563ef69a66e33aa632dabefb2aa9446

PR #2127, opened 2016-11-26
Prepare docs for #2067

jeddy3:
This PR prepares the docs for #2067 - adding links between related rules.

A while ago we started using hacky “quasi-absolute” URLs within the docs e.g. `(/docs/user-guides/configuration.md)`. This was to overcome a limitation in how the URLs were rewritten to work on the website. This approach is incompatible with linking between rules. In https://github.com/stylelint/stylelint.io/pull/28 the need for this workaround was removed.

This PR updates all the “pseudo-absolute” links to use standard relative URLs.

The PR is split into a handful of commits:

- The first set of commits are of the tweaks I made to the copy and fixing links while going through and validating all the links. 
- The latter commits are todo with the URL rewriting.

Removing the “quasi-absolute” paths has the added bonus that we can now make use of `remark-validate-link`, as [suggested by](https://github.com/stylelint/stylelint/issues/1464#issuecomment-225895964) @wooorm in June.  `remark-validate-link` turns out to be an amazing package, and is capable of spotting invalid anchors too. It would of caught those two broken links in the FAQ that are fixed in this PR.


davidtheclark:
👍

PR #2128, opened 2016-11-26
Allow non-existent codeFilenames

### LINKED ISSUE #2064: --stdin-filename requires file existence

`--stdin-filename` option seems to be broken. Running stylelint via cli with the option leads to `Error: ENOENT: no such file or directory` in case there is no file with given name in the working directory.

e.g. `7.4.0` – `7.5.0`

Output example. Note that `hey.@.css` doesn't exist but it doesn't have to, since it's only a name to infer a syntax.

```
~  ᐅ echo '.class { color: blue }' | /usr/local/bin/node /Users/zolotov/Downloads/webstorm_sample_asset/node_modules/stylelint/dist/cli.js -f json --stdin-filename hey.@.css
Error: ENOENT: no such file or directory, stat '/Users/zolotov/hey.@.css'
    at Error (native)
```

davidtheclark:
Closes #2064.

This was a little tricky to track down and fix without breaking other tests, but I think I got it.

The key is that we need to anticipate non-existent files in `lintSource` if `code` and `codeFilename` are being used.

### DIFF
--- a/src/__tests__/standalone.test.js
+++ b/src/__tests__/standalone.test.js
@@ -311,3 +311,32 @@ it("Setting `plugins` inside `configOverrides` object should overrides the ones 
     expect(results[0].warnings[0].text).toBe("found .bar (plugin/warn-about-bar)")
   })
 })
+
+describe("nonexistent codeFilename with loaded config", () => {
+  let actualCwd
+
+  beforeAll(() => {
+    actualCwd = process.cwd()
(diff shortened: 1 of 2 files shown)


jeddy3:
Very nice! I can see how this one would've been tricky.

davidtheclark:
Changelog entry:

> -   Fixed: regression causing `--stdin-filename` in CLI and `codeFilename` in Node API to error if a non-existent filename is provided ([#2128](https://github.com/stylelint/stylelint/pull/2128))

PR #2130, opened 2016-11-27
[WIP] Jest for rules

davidtheclark:
Experimenting with using Jest for rule tests.

Attach `testRule` to `global`, as done in `jest-setup.js`, seemed to be the only way I could get this going while using the existing `testRule` syntax. I'm hoping that this is an acceptable use of Jest that will not be broken in future releases ...

Besides that, I had to change some non-`testRule` tests from tape to Jest.

It's working pretty well, but not 100% yet. The only trade-off I know of is that Jest is so much slower than tape — 3x slower at least, which means that a 20 second run through the rule tests is now 1 minute. (Given all the hype about Jest's speed, I really am [confused by this](https://twitter.com/davidtheclark/status/802714480895373313) ...)


davidtheclark:
(ง'̀-'́)ง ... Because of the Jest-in-Travis problems documented in #2131 it looks like this 3x slowdown is actually much, much worse. I posted these commits 11 minutes ago and it's looking like those tests might never end ...

jeddy3:
I thought I'd try the `-i` flag as suggested [here]|(https://github.com/facebook/jest/issues/1742), but tests still appear slow.

jeddy3:
Continued in no-babel PR at https://github.com/stylelint/stylelint/pull/2158

Issue #2131, opened 2016-11-27
Random CI failures and slowdowns in Jest tests


davidtheclark:
I'm seeing some problems with the Jest tests that are bumming me out ... (ಠ_ಠ)

Occasionally, not consistently, async tests are timing out with an error like this:

```

 FAIL  src/rules/block-closing-brace-newline-after/__tests__/index.js (100.175s)
  ● block-closing-brace-newline-after › reject › no description
    Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
      
      at ontimeout (timers.js:365:14)
      at tryOnTimeout (timers.js:237:5)
      at Timer.listOnTimeout (timers.js:207:5)
```

Such tests may or may not pass on a retry.

Also, I'm seeing a lot of hang time slowing down Travis. Throughout the suite listing there are messages like this:

```
PASS  src/rules/declaration-block-properties-order/__tests__/grouped-strict.js (76.282s)
```

So running the tests in that one file ended up taking 76 seconds? That's a significant problem.

I can't reproduce these problems locally, and don't see problems with the code suggesting that we're not handling async right (e.g. never resolving a Promise). Maybe I'm missing something ...

Anyway, this needs more investigation. I'm beginning to regret our trial of the hot new thing ... (ಠ⌣ಠ)

davidtheclark:
Running with the `-i` flag prevented the timeout errors, as 

BUT I still was unable to finish running the tests successfully. Hit an allocation failure:

```

<--- Last few GCs --->

[2473:0x3109a00]   341803 ms: Mark-sweep 1404.1 (1532.0) -> 1404.1 (1532.0) MB, 1353.1 / 0.0 ms  allocation failure GC in old space requested
[2473:0x3109a00]   343219 ms: Mark-sweep 1404.1 (1532.0) -> 1404.1 (1531.0) MB, 1414.9 / 0.0 ms  last resort gc 
[2473:0x3109a00]   344601 ms: Mark-sweep 1404.1 (1531.0) -> 1404.1 (1531.0) MB, 1381.7 / 0.0 ms  last resort gc 


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x2ebae67c0d31 <JS Object>
[...]

PR #2135, opened 2016-11-28
Jest util tests

m-allanson:

Refs #2083
Replaces #2085

This should fix the recurring issues noted in #2085. If we're sticking with tape then this PR can just be closed.


davidtheclark:
Awesome work. Too bad it seems to be hitting the same problems noted in #2131.

src/utils/__tests__/findAtRuleContext.test.js
```diff
@@ -14,21 +13,20 @@ test("findAtRuleContext", t => {
     d {}
   `
 
-  t.plan(4)
   postcss().process(css).then(result => {
```
davidtheclark: Should return this Promise, I think.

davidtheclark:
> If we're sticking with tape then this PR can just be closed.

@m-allanson Yeah, I'm afraid we have to see if the Jest issues can be resolved first ... let's circle back to this within a week or two to see if we think it's a good idea or not.

jeddy3:
The current CI performance for using jest for `lib`, `utils` and `system-tests` looks good in this branch (using the `-i` flag), and that's still with babel in the mix. 

I think it could be worth, irrespective of how jest pans out for rule tests, continuing with this initiative and moving the `utils` tests to jest once the `drop-babel` branch is merged. This feels like it might give the best balance between DX and CI performance.

davidtheclark:
I will try to hash this out tonight.

davidtheclark:
I agree with @jeddy3 that we can merge this as long as we can get it consistently passing on CI. Probably the only significant work that will need to be done after the Great Babel Removal is to change `import` and `export` to CommonJS.

m-allanson:
I can take a look at rebasing this tonight (GMT).

PR #2140, opened 2016-11-28
Handle non-integer error codes as exit code 1 (#2115)

### LINKED ISSUE #2115: Exit code 0 when runtime error happened in loaded javascript configuration

nkabardin:
Fixes #2115 

When there is an error in require() in configuration file, cli gets an error code which is not actually integer, so passing it to process.exit is not a good idea.

Now value is getting checked if it's a number or not, and if it's not a number, it's replaced with plain `1`.

System test added that runs cli in ensures that exit code != 0

### DIFF
--- a/system-tests/004/004.test.js
+++ b/system-tests/004/004.test.js
@@ -1,0 +1,28 @@
+/* @flow */
+"use strict"
+const path = require("path")
+const spawn = require("child_process").spawn
+
+// this test ensures that CLI returns exit code which is not 0
+// when there is a runtime error in configuration javascript file
+// https://github.com/stylelint/stylelint/issues/2115
+
+it("004", (done) => {
+  const localPath = path.resolve(__dirname)
+  const cliPath = path.join(localPath, "../../lib/cli.js")
+
+  const cliProcess = spawn("node", [ cliPath, `${localPath}/*.css` ], { cwd: localPath })
+
+  cliProcess.on("error", function (error) {
+    console.log("error running cli:", error) // eslint-disable-line no-console
+  })
+
+  let stdout = ""
+  cliProcess.stdout.on("data", (data) => stdout += data)
+
+  cliProcess.on("close", function (code) {
+    expect(stdout.includes("Cannot find module")).toBeTruthy()
+    expect(code).not.toEqual(0)
+    done()
+  })
+})
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -214,5 +214,6 @@ Promise.resolve().then(() => {
   }
 }).catch(err => {
   console.log(err.stack) // eslint-disable-line no-console
-  process.exit(err.code || 1)
+  const exitCode = typeof err.code === "number" ? err.code : 1
+  process.exit(exitCode)
 })
--- a/system-tests/004/stylelint.config.js
+++ b/system-tests/004/stylelint.config.js
@@ -1,0 +1,2 @@
+"use strict"
+module.exports = require("./someUnknownFile.js")


nkabardin:
@davidtheclark thanks!

something is frozen on ci though. Can't understand what's happening.

nkabardin:
Checked log on Travis, and there might be issues with running this system test on jenkins — any ideas what's happening there?

ntwb:
@nkabardin I've restarted the CI tests so lets see what happens...

ntwb:
Coveralls is failing in both `5569.1` and `5569.2` jobs even though `#5569.1` [is reporting](https://travis-ci.org/stylelint/stylelint/jobs/179432287) as `passed` and `#5569.2` is failing with a Jest test timing out.

nkabardin:
So I guess it's something related to `child_process.spawn`, are there any CI-related limitations for that?

nkabardin:
Playing around with error logging in system test, let's see what CI will tell us...

nkabardin:
did several attempts trying to log out the problems with running stuff, added three `console.log` to test, and didn't catch the problem, nothing logged.

Any ideas?

nkabardin:
Rebased branch over current master (so, support babel drop and no need to use babel-node to run cli) and also  and removed all debug stuff and it's currently ready and well tested, and surprisingly issues on CI are also solved, so we got green build there, so yay!

ntwb:
LGTM, though let's get another review from @davidtheclark 👍

ntwb:
Coverage tests were failing if less than 95%, I've dropped that value to 90%

So ignore the Coveralls failure here in this PR (I couldn't find a way to restart the job)

system-tests/004/004.test.js
```diff
@@ -0,0 +1,24 @@
...
+
+// this test ensures that CLI returns exit code which is not 0
+// when there is a runtime error in configuration javascript file
+// https://github.com/stylelint/stylelint/issues/2115
+
+it("004", (done) => {
+  const localPath = path.resolve(__dirname)
+  const cliPath = path.join(localPath, "../../lib/cli.js")
+
+  const cliProcess = spawn("node", [ cliPath, `${localPath}/*.css` ], { cwd: localPath })
```
davidtheclark: Could you add an `'error'` listener to this process?

nkabardin: done

system-tests/004/004.test.js
```diff
@@ -0,0 +1,28 @@
...
+
+  cliProcess.on("error", function (error) {
+    console.log("error running cli:", error) // eslint-disable-line no-console
```
evilebottnawi: I think we should failed test if have error in `error` event

nkabardin: it fails anyway if there is an error, it just added logging for that.
Couldn't reproduce a case when it didn't fail on error

evilebottnawi: @nkabardin we can use `mock` for reproduce error, we got non zero code (thoretically :smile: ) and test failed, lgtm

nkabardin: I didn't get it, current code looks good to you, or it will look good to you when some mocking will be added? :)

evilebottnawi: @nkabardin no, all ok

davidtheclark:
Thanks again @nkabardin!

davidtheclark:
Changelog entry

> -   Fixed: bug where some Node errors in special cases did not cause the CLI to exit with a non-zero code ([#2140](https://github.com/stylelint/stylelint/pull/2140))

nkabardin:
@davidtheclark glad I was able to help the project! thanks for `stylelint` ❤️

PR #2134, opened 2016-11-27
[WIP] Drop babel

davidtheclark:
Ref #2133.

Experimenting with using as much automation as possible to drop Babel.

What I've done so far is to use a limited number of Babel plugins, covering syntax *not* permitted by Node 4, to overwrite existing files in `src/`. This looks like it might be a suitable starting point. 

Modules still exist, because that would have gotten nasty — I think those might require the most manual labor ... we'll see.

I'll toy with this a little bit more later today ...

PR #2150, opened 2016-12-01
Jest util tests (post-babel)

m-allanson:

Replaces #2135, accounting for the changes in #2134.

Both master and #2135 had renamed the same files in different ways, which is why I've created _another_ PR for this. Due to the file renaming in both branches, this is essentially #2135 with a sort of manual rebase applied. 

Some of the flow types have been ditched because it made this simpler to manage, they can be re-added at a later date.

I've just about run out of time, but I think it's looking ok.


m-allanson:
The system tests are failing in node v4

m-allanson:
I've removed the object destructuring from the system tests, but they're failing with a new error which I don't understand.

```
Test suite failed to run

    SyntaxError: Unexpected token =

      at transformAndBuildScript (node_modules/jest/node_modules/jest-cli/node_modules/jest-runtime/build/transform.js:316:10)
      at Object.<anonymous>.jest (system-tests/001/001.test.js:4:25)
```

Line 4 of `001.test.js` looks like this:

```js
const systemTestUtils = require("../systemTestUtils")
```

davidtheclark:
Looking now ...

davidtheclark:
@m-allanson `Unexpected token =` happened because there was a default parameter in there, which is not supported by Node 4. Not a very helpful stack trace.

davidtheclark:
Thanks @m-allanson! Looks like it's working well now.

PR #2158, opened 2016-12-03
Use jest for rule tests

jeddy3:
This PR ports the last of @davidtheclark's #2130 work to the no-babel code base.

It currently crashes with an out of memory error on my local machine, a 16gb Macbook Pro, even when not using the `--runInBand` flag.

This PR is here:

1. For anyone to take a look at and investigate further why this crash occurs.
2. To track any upstream performance changes in jest (at https://github.com/facebook/jest/issues/2179)

Update: I can complete the run using `npm run test:jest -- -w='4'` in 52s on my local machine.


davidtheclark:
Looks like I got all the tests passing by removing `graceful-fs` from `cosmiconfig`!

Worth noting that the serialized Jest tests are significant slower than our tape script. But not too bad, I think.

Any objections to merging, then? (Good to know that it wouldn't be hard to revert, since we have the `testRule` abstraction.)

jeddy3:
> Any objections to merging, then?

None, it's fantastic we're been able to migrate to jest wholesale!

Absolutely awesome job tracking down that memory leak!

PR #2159, opened 2016-12-03
Include bin directory in package

jeddy3:
`scripts/visual.sh` now works.

Ran `master` against all my local test files. All went well. Also very refreshing not seeing babel transpiling things for a local install :)


davidtheclark:
🌮

PR #2160, opened 2016-12-03
Format code

jeddy3:
A bit of code style tidying up after the Great Babel Removal.

Some of the conditionals looked a bit gnarly, so I've tried to consistently split those across lines.

I only got through about 20% of the rules before running out of steam. We can either push on and try to do it in one go, or just revisit the code style of each rule when there's a bug to fix in it.

I did just notice that some of the code style around the flow comments are [a bit out there](https://github.com/stylelint/stylelint/blob/master/lib/augmentConfig.js#L31). It might be worth tidying those up. What's the ideal structure for them?


davidtheclark:
Thanks! Preservation of code format is one place where PostCSS has Babel beat.

> We can either push on and try to do it in one go, or just revisit the code style of each rule when there's a bug to fix in it. ... I did just notice that some of the code style around the flow comments are a bit out there

I can tackle the Flow comments today. If you want to push ahead with the rules some time after your weariness has worn off, go for it! Otherwise, yeah, we can always just change things as we revisit code. Whatever you feel like.

ntwb:
I went hunting for _fixable_ ESLint rules for this, alas they do not exist, yet.

There are proposals for both [array-element-newline](https://github.com/eslint/eslint/issues/6075) and [array-bracket-newline](https://github.com/eslint/eslint/issues/6073) though.

PR #2162, opened 2016-12-03
Added link to stylelint-config-amp

tinovyatkin:
Documentation fix, added link to my [stylelint-config-amp](https://github.com/tinovyatkin/stylelint-config-amp) config that enforces rules for [Accelerated Mobile Pages](https://ampproject.org) requirements.

### DIFF
--- a/docs/user-guide/plugins.md
+++ b/docs/user-guide/plugins.md
@@ -19,3 +19,4 @@ Plugins are rules and sets of rules built by the community that support methodol
 ## Validators
 
 -   [`stylelint-csstree-validator`](https://github.com/csstree/stylelint-validator): Validate CSS values to match W3C specs and browsers extensions.
+-   [`stylelint-config-amp`](https://github.com/tinovyatkin/stylelint-config-amp): Enforces CSS restrictions of Accelerated Mobile Pages.


jeddy3:
@tinovyatkin Thanks for the PR! It looks like a great config. However, that page is a curated list of [plugins](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/plugins.md), not [shared configs](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#extends). For those we recommend just adding `stylelint-config` to your `package.json`, which you've done. People can then find it [here](https://www.npmjs.com/browse/keyword/stylelint-config), which we link to from the main README.

tinovyatkin:
I understand and respect your position, but this config is actually more validator than a shareable config - if you will check a NPM link you gave me then you will find a big deal of configs like _it is just another set of my opinated rules_, while this one gives you ability to lint CSS according published restrictions of emerging technology.
Hopefully, almost all these checks can be done by just built-in rules of stylelint (and it shows how powerful it is) and RegEx, but it doesn't mean it's not a validator.

From my point of view, it gives more benefits for `stylelint` that struggles from lack of publicity to associate with AMP technology (and Google behind it) and this small config allows to do so, but once again, I respect your opinion if you think that `it just another shareable config`.

Thank you!

PR #2163, opened 2016-12-03
Lots of Flow work

davidtheclark:
Cleans up the aberrant spacing that Babel inflicted on Flow comments, and adds Flow comments to all of the utils.

### DIFF
--- a/lib/utils/isStandardSyntaxRule.js
+++ b/lib/utils/isStandardSyntaxRule.js
@@ -1,15 +1,13 @@
+/* @flow */
 "use strict"
 
 const _ = require("lodash")
 const isCustomPropertySet = require("../utils/isCustomPropertySet")
 
 /**
- * Check whether a rule is standard
- *
- * @param {Rule} postcss rule node
- * @return {boolean} If `true`, the rule is standard
+ * Check whether a Node is a standard rule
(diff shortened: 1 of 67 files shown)


lib/utils/isStandardSyntaxRule.js
```diff
@@ -1,15 +1,13 @@
...
 
 /**
- * Check whether a rule is standard
- *
- * @param {Rule} postcss rule node
- * @return {boolean} If `true`, the rule is standard
+ * Check whether a Node is a standard rule
  */
-module.exports = function (rule) {
+module.exports = function (rule/*: Object*/)/*: boolean*/ {
```
m-allanson: Should this be `postcss$rule` instead of `Object`?

m-allanson:
Nice. One question, otherwise LGTM.

lib/utils/isStandardSyntaxRule.js
davidtheclark: @m-allanson The fact that we check for `rule.params` below means it's confusing — might not actually be a rule.

m-allanson: @davidtheclark 👍

Issue #2165, opened 2016-12-04
Jest tests not running on NodeJS v6.x & 4.x Travis CI jobs


ntwb:
NodeJS v6.x: https://travis-ci.org/stylelint/stylelint/jobs/181025543
NodeJS v4.x: https://travis-ci.org/stylelint/stylelint/jobs/181025544

```shell
$ npm run test:jest -- --runInBand --testPathPattern /lib/ && npm run test:tape
> stylelint@7.6.0 test:jest /home/travis/build/stylelint/stylelint
> jest "--runInBand" "--testPathPattern" "/lib/"
No tests found
  1 file checked.
  testPathDirs: /home/travis/build/stylelint/stylelint/lib,/home/travis/build/stylelint/stylelint/system-tests - 0 matches
  testPathIgnorePatterns: /node_modules/ - 1 match
  testRegex: .*\.test\.js$ - 0 matches
> stylelint@7.6.0 test:tape /home/travis/build/stylelint/stylelint
> node scripts/rules-tape.js
Testing rules ...
ok: 23163
not ok: 0
total: 23163
```

~~The issue appears to be the `,` after `lib` in the `testPathDirs`: i.e `lib,/` rather than `lib/`~~
Edit: The comma is a separator between two paths, this is not the cause 

https://github.com/stylelint/stylelint/blob/master/package.json#L113-L116
```json
    "testPathDirs": [
      "lib",
      "system-tests"
    ],
```

Running `npm run test:jest -- --runInBand --debug --testPathPattern /lib/` locally:
```shell
  "testPathDirs": [
    "/Users/netweb/dev/stylelint/stylelint/lib",
    "/Users/netweb/dev/stylelint/stylelint/system-tests"
  ],
```

ntwb:
Noting NodeJS v7.x on Travis CI and NodeJS v7.x on AppVeyor work as expected.

This issue has me rather stumped at the moment, will revisit in a few hours 🤔

davidtheclark:
Can't seem to reproduce locally on Node 4.

davidtheclark:
By changing the `testPathPattern` flag to `.*lib/.*` I was able to get these tests running on Travis. Unfortunately, Node 4 crashes!

davidtheclark:
Fixed by #2167.

PR #2166, opened 2016-12-04
Debugging #2165

ntwb:
See #2165


jeddy3:
Done in https://github.com/stylelint/stylelint/pull/2158

PR #2169, opened 2016-12-04
Updates docs for testing set of rules

jeddy3:
As an aside, by default I think the `--watch` is only suppose to run the tests on _changed_ files. However, it currently triggers a complete run on my machine. Perhaps I'm missing something?


davidtheclark:
Nicely done.

PR #2171, opened 2016-12-04
Ignore node_modules effectively

### LINKED ISSUE #2149: Ignoring node_modules when using the PostCSS API
Hello,

We're using stylelint in the SUIT CSS preprocessor and one issue we have is around [ignoring imports](https://github.com/suitcss/preprocessor/issues/58) that are located in `node_modules`. We're using it like so:

```js
  var processor = postcss();
  if (options.lint) {
    processor.use(stylelint(options.stylelint));
  }
```

From looking round the source I can see the Node API uses globs with a `alwaysIgnoredGlobs` containing `node_modules` but it looks like the PostCSS goes around this check via `lintSource`. I only looked briefly so do correct me if I'm wildly inaccurate :)

Just wondering if `alwaysIgnoredGlobs` should be respected when using stylelint this way?

I can put together an example of the issue if that helps further

Thanks

davidtheclark:
Closes #2149.

Managed to reproduce that problem in a test case, and made it pass. I'm not sure if this feature ever worked *before* in the PostCSS plugin; but we always intended it to, assumed it did, so I'd consider this a bug fix.

cc @simonsmith

### DIFF
--- a/lib/__tests__/ignore.test.js
+++ b/lib/__tests__/ignore.test.js
@@ -365,3 +365,41 @@ describe("using codeFilename and ignoreFiles with configBasedir", () => {
     expect(_.get(results[0], [ "_postcssResult", "standaloneIgnored" ])).toBeTruthy()
   })
 })
+
+describe("file in node_modules", () => {
+  const lint = () => standalone({
+    code: "a {}",
+    codeFilename: "/something/node_modules/path/to/something.css",
+    config: {
+      rules: {
+        "block-no-empty": true,
+      },
+    },
+  })
+
+  it("is ignored", () => {
+    return lint().then((output) => {
+      expect(output.results[0].warnings.length).toBe(0)
+      expect(output.results[0].ignored).toBeTruthy()
+    })
+  })
+})
+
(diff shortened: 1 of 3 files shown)


lib/__tests__/ignore.test.js
```diff
@@ -365,3 +365,41 @@ describe("using codeFilename and ignoreFiles with configBasedir", () => {
     expect(_.get(results[0], [ "_postcssResult", "standaloneIgnored" ])).toBeTruthy()
   })
 })
+
+describe("file in node_components", () => {
```
simonsmith: `node_modules`?

simonsmith:
Thanks for this @davidtheclark :)

lib/__tests__/ignore.test.js
davidtheclark: :facepalm:

jeddy3:
Changlelog:

> -   Fixed: the `node_modules` and `bower_components` directories are correctly ignored by default when stylelint is used as a PostCSS plugin.

PR #2172, opened 2016-12-04
Simplify NPM test scripts now that Tape has been removed

ntwb:
Now that Tape was removed in #2158 there's no need for multiple `tape:*` scripts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,14 +91,14 @@
   "scripts": {
     "benchmark-rule": "node scripts/benchmark-rule.js",
     "flow": "flow",
+    "jest": "jest",
     "lint:js": "eslint .",
     "lint:md": "remark . --quiet --frail",
     "lint": "npm-run-all --parallel lint:*",
     "pretest": "npm-run-all --serial lint flow",
     "release": "npmpub",
-    "test:jest": "jest",
-    "test": "npm-run-all --serial test:*",
-    "watch": "npm run test:jest -- --watch"
+    "test": "jest",
+    "watch": "npm run jest -- --watch"
   },
   "jest": {
     "setupFiles": [
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "pretest": "npm-run-all --serial lint flow",
     "release": "npmpub",
     "test": "jest",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -91,14 +91,14 @@
...
     "lint:js": "eslint .",
     "lint:md": "remark . --quiet --frail",
     "lint": "npm-run-all --parallel lint:*",
     "pretest": "npm-run-all --serial lint flow",
     "release": "npmpub",
-    "test:jest": "jest",
-    "test": "npm-run-all --serial test:*",
-    "watch": "npm run test:jest -- --watch"
+    "test": "jest",
+    "watch": "npm run jest -- --watch"
```
jeddy3: Can we have just `jest --watch` here please? I sometimes find the extra `--` unreliable when passing _further_ flags into the npm command e.g. `npm run watch -- logHeapSize`.

jeddy3:
Neater. I've made one request though.

package.json
ntwb: Running `npm run watch` works as expected using `"watch": "npm run jest -- --watch"`

Running `npm run watch` does **not** work using `"watch": "npm run jest --watch"`, this results in the Jest tests being ran and the _watch_ mode is never started:

```shell
> stylelint@7.6.0 watch /Users/netweb/dev/stylelint/stylelint
> npm run jest --watch


> stylelint@7.6.0 jest /Users/netweb/dev/stylelint/stylelint
> jest

 PASS  lib/rules/value-keyword-case/__tests__/index.js
...
...
```

ntwb:
Followed up inline, tl;dr: requested change doesn't work

package.json
davidtheclark: @ntwb I think the request is for `"watch": "jest --watch"`.

ntwb: Done in https://github.com/stylelint/stylelint/pull/2172/commits/ff6acfed29dc2c5cb3b6f4bfd5be756ad2588207 :)

ntwb:
All sorted now 👍

jeddy3:
Thanks :)

Issue #2139, opened 2016-11-28
Incorrect severity level for a rule called from other rule


hudochenkov:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

If rule was executed from another rule, it has incorrect severity level `ignore`. stylelint-order's `declaration-block-property-groups-structure` [calling](https://github.com/hudochenkov/stylelint-order/blob/master/rules/declaration-block-property-groups-structure/index.js#L32) `declaration-block-properties-order`. `declaration-block-property-groups-structure` has default `error` severity level, but `declaration-block-properties-order` has `ignore` level. I've [added temporary fix](https://github.com/hudochenkov/stylelint-order/blob/master/rules/declaration-block-property-groups-structure/index.js#L34-L37) in my plugin to set severity level. I wanted to use `options` object for that, but didn't figured out how to get correct severity level. There is an `if` because when I rune tests there is no `result.stylelint.ruleSeverities`.

> What CSS is needed to reproduce this issue?

```css
a {
    width: 0;

    position: relative;
}
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
    "rules": {
        "order/declaration-block-property-groups-structure": [
            {
                "emptyLineBefore": "always",
                "order": "strict",
                "properties": [
                    "position",
                    "top"
                ]
            },
            {
                "emptyLineBefore": "always",
                "order": "strict",
                "properties": [
                    "width"
                ]
            }
        ]
    }
}
```

> Which version of stylelint are you using?

`7.5.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI, Node API.

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No.

> What did you expect to happen?

Called rule has same severity level as rule which called it.

> What actually happened (e.g. what warnings or errors you are getting)?

```js
{
    line: 5,
    column: 4,
    rule: 'declaration-block-properties-order',
    severity: 'ignore',
    text: 'Expected "position" to come before "width" (declaration-block-properties-order)'
}
```

jeddy3:
@hudochenkov Thank you for the report. This looks like it'll require some investigation and any help would be appreciated.

davidtheclark:
Uh oh ... maybe the ability to use other rules is not quite mature. I have an idea ...

hudochenkov:
[...]

PR #2173, opened 2016-12-04
Add checkAgainstRule util

davidtheclark:
Here's an idea to solve #2139.

It exposes another function `stylelint.checkAgainstRule`, which you can use to run a PostCSS `Root` object through a rule, and invoke a callback with every warning that is created.

What I'd like to see in a situation liked #2139 is that the plugin rule that uses a core rule is the only one that calls `report()`. *It* is the reporting rule, and its severity should be used.

I have not yet added documentation, but did add some tests.

@hudochenkov would you please try this branch out and see if it works for your use-case?

@stylelint/core any input on this?

### DIFF
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -38,13 +38,61 @@ In order for your plugin rule to work with the [standard configuration format](.
 
 ## `stylelint.utils`
 
-Three of stylelint's internal utilities are exposed publicly in `stylelint.utils` and should be used:
+stylelint exposes some utilities that are useful. *For details about the APIs of these functions, please look at comments in the source code and examples in the standard rules.*
 
--   `report`: Report your linting warnings. *Do not use `node.warn()` directly.* If you use `report`, your plugin will respect disabled ranges and other possible future features of stylelint, so it will fit in better with the standard rules.
--   `ruleMessages`: Tailor your messages to look like the messages of other stylelint rules.
--   `validateOptions`: Help your user's out by checking that the options they've submitted are valid.
+### `stylelint.utils.report`
 
-For details about the APIs of these functions, please look at comments in the source code and examples in the standard rules.
+Adds warnings from your plugin to the list of warnings that stylelint will report to the user.
+
+*Do not use PostCSS's `node.warn()` method directly.* When you use `stylelint.utils.report`, your plugin will respect disabled ranges and other possible future features of stylelint, providing a better user-experience, one that better fits the standard rules.
+
+### `stylelint.utils.ruleMessages`
+
+Tailors your messages to the format of standard stylelint rules.
+
+### `stylelint.utils.validateOptions`
+
+Validates the options for your rule.
+
+### `stylelint.utils.checkAgainstRule`
+
+Checks CSS against a standard stylelint rule *within your own rule*. This function provides power and flexibility for plugins authors who wish to modify, constrain, or extend the functionality of existing stylelint rules.
+
+Accepts an options object and a callback that is invoked with warnings from the specified rule. The options are:
+- `ruleName`: The name of the rule you are invoking.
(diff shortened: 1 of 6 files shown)


ntwb:
> _"... is that the plugin rule that uses a core rule is the only one that calls `report()`. It is the reporting rule, and its severity should be used."_

SGTM 👍

jeddy3:
SGTM too.

hudochenkov:
I don't know what callback to use in order to push warning where they belong, so I've used `console.log` for testing:

```js
stylelint.checkAgainstRule(
	{
		ruleName: 'declaration-block-properties-order',
		ruleSettings: [ cleanedConfig, options ],
		root,
	},
	(warning) => console.log(warning)
);
```

I've used code above instead of https://github.com/hudochenkov/stylelint-order/blob/master/rules/declaration-block-property-groups-structure/index.js#L32-L40

It shows this warning on setup as in #2139:

```
Warning {
  type: 'warning',
  text: 'Expected "position" to come before "width" (declaration-block-properties-order)',
  line: 5,
  column: 4,
  severity: 'ignore',
  rule: 'declaration-block-properties-order',
  node:
   Declaration {
     raws: { before: '\n\n\t\t\t', between: ': ' },
     type: 'decl',
     parent:
      Rule {
        raws: [Object],
        type: 'rule',
        nodes: [Object],
        parent: [Object],
        source: [Object],
        selector: 'a',
        lastEach: 3,
        indexes: {} },
     source: { start: [Object], input: [Object], end: [Object] },
     prop: 'position',
     value: 'relative' } }
```

Problem with severity lays some where else :(

davidtheclark:
@hudochenkov Here's how you'd use this:

```js
function myPluginRule(root, result) {
  /* ... */
  stylelint.checkAgainstRule({
    ruleName: 'declaration-block-properties-order',
    ruleSettings: someAppropriateSettings,
    root,
  }, (warning) => {
    stylelint.utils.report({
      result,
      ruleName: myPluginRuleName,
      message: myMessage,      
      node: warning.node,
      line: warning.line,
      column: warning.column
    })
  });
}
```

The idea is that you'd actually `report` a warning *for your own rule*. A user doesn't want to see a warning for `declaration-block-properties-order` if they didn't set that rule in their config — that would be confusing. They want to see the warning for `myPluginRuleName`, even if the logic that determined the warning originated in `declaration-block-properties-order`. Also, this way the severity set for `myPluginRuleName` will be reported with the warning.

Does this make sense?

hudochenkov:
Sounds reasonable. I've changed `message: myMessage` to `message: warning.text`, because I can't predict all the messages from core rule. So the output looks like this:

```
 5:4  ✖  Expected "position" to come before "width" (declaration-block-properties-order)  order/declaration-block-property-groups-structure
```

`warning.text` contains core rule name. Is it ok to show warning as above, or it's better to cut core rule name from `warning.text`?

davidtheclark:
> Is it ok to show warning as above, or it's better to cut core rule name from warning.text

@hudochenkov If I were using that plugin, I think I'd find it confusing to see a rule mentioned that I did not deliberately turn on — so I'd suggest using a regex or something to replace the name with that of the rule that is reporting.

jeddy3:
It sounds like this approach is working out well.

I'll label this up as "needs docs".

davidtheclark:
Added some documentation in `developer-guide/plugins`.

jeddy3:
Excellent. That example really highlights the power of this addition!

jeddy3:
@davidtheclark Those flow errors have resurfaced. Should we fix those before merging?

davidtheclark:
Got the tests passing! Please take another look sometime :)

jeddy3:
Changelog 

> -   Added: `stylelint.utils.checkAgainstRule` for checking CSS against a standard stylelint rule *within your own rule* ([#2173](https://github.com/stylelint/stylelint/pull/2173)).

As an aside, I think with new features like this the next release is shaping up as a good one :)

Issue #2170, opened 2016-12-04
Custom rules overrides extended config


hinok:
**Here is repository that reproduce bug described below**
https://github.com/hinok/stylelint-sass-css-modules

I've found the same problem described in another repository
https://github.com/kristerkari/stylelint-scss/issues/86

> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug when you extends a config and try to define custom rules.

> Which rule, if any, is this issue related to?

```
selector-pseudo-class-no-unknown - ignorePseudoClasses
property-no-unknown - ignoreProperties
at-rule-no-unknown - ignoreAtRules
```

> What CSS is needed to reproduce this issue?

I prepared a repository that represents this bug https://github.com/hinok/stylelint-sass-css-modules

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "extends": "stylelint-config-standard",
  "rules": {
    "selector-pseudo-class-no-unknown": [true, {
      "ignorePseudoClasses": [
        "export",
        "import",
        "global",
        "local",
      ]
    }],
    "property-no-unknown": [true, {
      "ignoreProperties": [
        "composes",
      ]
    }],
    "at-rule-no-unknown": [true, {
      "ignoreAtRules": [
        "value",
      ]
    }]
  }
}
```

> Which version of stylelint are you using?

`7.6.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylelint *.scss --syntax scss`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

Rules defined by extended configuration should be merged with my custom rules.

> What actually happened (e.g. what warnings or errors you are getting)?

My custom rules overrides rules from extended config.

davidtheclark:
@hinok Thanks for opening this. After looking into the linked issue, I realized that there's probably a good solution for stylelint-scss, and [I wrote about it over there](https://github.com/kristerkari/stylelint-scss/issues/86#issuecomment-264737962). Will that solution work for you?

> My custom rules overrides rules from extended config.

I don't understand what this means. In the repo you linked to, I do not see any "custom rules". You are in fact overriding `stylelint-config-standard`.

hinok:
> My custom rules overrides rules from extended config.

Frankly I've thought that it works more like eslint where all rules from

```
"rules": { ... }
```

are merged into configuration defined in

```
"extends": { ... }
```

[...]

PR #2178, opened 2016-12-06
Add section about extending and overriding rules.

hinok:

https://github.com/stylelint/stylelint/issues/2170

Show clear example that secondary options are overriden
instead of be extended.

### DIFF
--- a/docs/user-guide/faq.md
+++ b/docs/user-guide/faq.md
@@ -115,3 +115,27 @@ Refer to [this](about-rules.md#-empty-line-before-and--max-empty-lines) section 
 ## How do I "pin" to a particular branch, pull request or fork of stylelint in my `package.json`?
 
 Use [`npm-git-install`](https://github.com/lzrski/npm-git-install) to work around [a current limitation](https://github.com/npm/npm/issues/3055) of `npm`.
+
+## If I extend a configuration, will the options for each rule be merged or overridden?
+
+In most cases rules will be merged but **secondary options will be overriden**.
+
+There is no way at the moment to disable specific secondary option in a rule. As example look at `ignoreAtRules` below.
+
+```json
+{
+  "extends": [
+    "stylelint-config-standard"
+  ],
+  "rules": {
+    "at-rule-no-unknown": [true, {
+      "ignoreAtRules": [
+        "at-root",
+        "value"
+      ]
+    }]
+  }
+}
+```
+
+If you want to disable only `"ignoreAtRules": ["value"]`, you have to overrides the whole `"ignoreAtRules"` array. More details can be found in [#1646](https://github.com/stylelint/stylelint/issues/1646#issuecomment-232779957), [#2170](https://github.com/stylelint/stylelint/issues/2170) issues.


docs/user-guide/faq.md
```diff
@@ -115,3 +115,27 @@ Refer to [this](about-rules.md#-empty-line-before-and--max-empty-lines) section
 ## How do I "pin" to a particular branch, pull request or fork of stylelint in my `package.json`?
 
 Use [`npm-git-install`](https://github.com/lzrski/npm-git-install) to work around [a current limitation](https://github.com/npm/npm/issues/3055) of `npm`.
+
+## If I extend a configuration, will the options for each rule be merged or overridden?
+
+In most cases rules will be merged but **secondary options will be overriden**.
```
jeddy3: Aren't rule consistently overridden?

docs/user-guide/faq.md
```diff
@@ -115,3 +115,27 @@ Refer to [this](about-rules.md#-empty-line-before-and--max-empty-lines) section
...
+
+There is no way at the moment to disable specific secondary option in a rule. As example look at `ignoreAtRules` below.
```
jeddy3: "no way at the moment"?

I think this implies that it's a bug that needs fixing. When in fact it's a design decision.

docs/user-guide/faq.md
```diff
@@ -115,3 +115,27 @@ Refer to [this](about-rules.md#-empty-line-before-and--max-empty-lines) section
...
+      "ignoreAtRules": [
+        "at-root",
+        "value"
+      ]
+    }]
+  }
+}
+```
+
+If you want to disable only `"ignoreAtRules": ["value"]`, you have to overrides the whole `"ignoreAtRules"` array. More details can be found in [#1646](https://github.com/stylelint/stylelint/issues/1646#issuecomment-232779957), [#2170](https://github.com/stylelint/stylelint/issues/2170) issues.
```
jeddy3: I think surfacing that issue isn't a good idea.

docs/user-guide/faq.md
hinok: When I used "merged", I was thinking about this scenario

`config-1.json`

```json
{
 "rules": {
   "font-weight-notation": "numeric",
   "number-leading-zero": "always",
   "at-rule-no-unknown": [true, {
     "ignoreAtRules": [
       "at-root",
       "value"
     ]
   }]
 }
}
```

`config-2.json`

```json
{
 "extends": "./config-1.json",
 "rules": {
   "font-weight-notation": "named-where-possible",
   "at-rule-no-unknown": [true, {
     "ignoreAtRules": [
       "extends"
     ]
   }]
 }
}
```

Result

```json
{
 "extends": "./config-1.json",
 "rules": {
   "font-weight-notation": "named-where-possible", // this is overriden
   "number-leading-zero": "always", // this is merged from config-1
   "at-rule-no-unknown": [true, {
     "ignoreAtRules": [ // this is overriden
       "extends"
     ]
   }]
 }
}
```

docs/user-guide/faq.md
hinok: Okay, I will remove them. Personally I like that kind of connection because sometimes issues contains a lot of useful informations.

docs/user-guide/faq.md
hinok: That was not my intention, sorry. I know that it's not a bug but just the way how stylelint works.  So maybe something like

```
All defined rules are merged but secondary options for rules are always overriden.
```

docs/user-guide/faq.md
jeddy3: Ok, I understand. I suggest we be explicit about this then.

Having said that, this is already explicitly stated within the [configuration](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#extends) guide. Perhaps we should just link, from within the FAQ, to this section of the docs like we do with the other FAQ entries?

Note, all options are overridden... not just the secondary ones.

docs/user-guide/faq.md
jeddy3: Good point.

jeddy3:
I'm going to merge this in and tweak so that we have something in for the pending release.

hinok:
@jeddy3 Sorry, I had to deal with my private stuff at the last weekend thus I couldn't even check anything on web. I'm glad that it was merged ;)

Issue #2009, opened 2016-10-24
False positives with SCSS maps in color-named


ThePeach:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

This is a bug related to `color-named` and happens when a variable name has the same name of a key of a map

The needed SCSS snippet to trigger this is the following:

``` css
@function color($key) {
  @if map-has-key($colors, $key) {
    @return map-get($colors, $key);
  }
  return null;
}

// Color palette
$colors: (
  white: #FFFFFF,
  grey: #333333
);

$color-footer: color(white) !default;
```

> What stylelint configuration is needed to reproduce this issue?

``` json
{
  "rules": {
    "color-named": "never"
  }
}
```

> Which version of stylelint are you using?

`7.5.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

via `gulp-stylelint`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

"Yes, it's related to SCSS maps."

> What did you expect to happen?

"No warnings to be flagged."

> What actually happened (e.g. what warnings or errors you are getting)?

"The following warnings were flagged:"

``` shell
test.css
  10:3   ✖  Unexpected named color "white"    color-named
  11:3   ✖  Unexpected named color "grey"     color-named
  14:22  ✖  Unexpected named color "white"    color-named
```

jeddy3:
@ThePeach Thanks for using the template.

You can the [`"ignore": ["inside-function"]`](https://github.com/stylelint/stylelint/tree/master/src/rules/color-named#ignore-inside-function) secondary option here e.g.

`"color-named": ["never", { "ignore": ["inside-function"] }]`

ThePeach:
@jeddy3 thanks for the explanation. I don't have a use case on all the projects I'm using to do the automatic replacement with PostCSS and I'd rather write that once and forget about it.

I've used the suggested option and it seems to be working fine.
I would also agree that as a default behaviour it might be probably better, but -again- I don't have a use case for this instance.

davidtheclark:
@jeddy3 I'm fine with ignoring maps by default.

jeddy3:
[...]

PR #2182, opened 2016-12-09
Ignore maps and lists in color-named

mmase:

https://github.com/stylelint/stylelint/issues/2009

No, it's self explanatory.

### DIFF
--- a/lib/rules/color-named/__tests__/index.js
+++ b/lib/rules/color-named/__tests__/index.js
@@ -56,6 +56,9 @@ testRule(rule, {
   }, {
     code: "a { font: 10px/14px Brown, sans-serif; }",
     description: "ignore font family names that are colors",
+  }, {
+    code: "$colors: (white: #fff, blue: #00f);",
+    description: "ignore non-standard list and map syntax",
   } ],
 
   reject: [ {
@@ -93,6 +96,11 @@ testRule(rule, {
     message: messages.rejected("white"),
     line: 1,
     column: 24,
+  }, {
+    code: "a { background: color(green); }",
+    message: messages.rejected("green"),
+    line: 1,
+    column: 23,
   } ],
 })
 
@@ -255,6 +263,11 @@ testRule(rule, {
     message: messages.expected("black", "rgb(0,0,0)"),
     line: 1,
     column: 23,
+  }, {
+    code: "a { color: color(#000 a(50%)) }",
+    message: messages.expected("black", "#000"),
+    line: 1,
+    column: 18,
   } ],
 })
 
@@ -300,6 +313,8 @@ testRule(rule, {
     code: "a { color: map-get($colour, blue(60%)); }",
   }, {
     code: "a { color: map-get($colour, blue); }",
+  }, {
+    code: "a { background: color(blue); }",
   } ],
 
   reject: [ {
--- a/lib/rules/color-named/index.js
+++ b/lib/rules/color-named/index.js
@@ -2,6 +2,7 @@
 
 const _ = require("lodash")
 const declarationValueIndex = require("../../utils/declarationValueIndex")
+const isStandardSyntaxFunction = require("../../utils/isStandardSyntaxFunction")
 const isStandardSyntaxValue = require("../../utils/isStandardSyntaxValue")
 const keywordSets = require("../../reference/keywordSets")
 const namedColorData = require("../../reference/namedColorData")
@@ -67,6 +68,10 @@ const rule = function (expectation, options) {
           optionsMatches(options, "ignore", "inside-function")
           && type === "function"
         ) {
+          return false
+        }
+
+        if (!isStandardSyntaxFunction(node)) {
           return false
         }
 


lib/rules/color-named/index.js
```diff
@@ -49,7 +50,7 @@ const rule = function (expectation, options) {
     const namedColors = Object.keys(namedColorData)
 
     root.walkDecls(decl => {
-      if (propertySets.acceptCustomIdents.has(decl.prop)) {
+      if (propertySets.acceptCustomIdents.has(decl.prop) || !isStandardSyntaxDeclaration(decl)) {
```
evilebottnawi: use `isStandardSyntaxDeclaration` we disabled check on color in variables, root and scss nested properties (see https://github.com/stylelint/stylelint/blob/master/lib/utils/isStandardSyntaxDeclaration.js).
@stylelint/core what do you think guys? I vote for the exclusion of non-standard syntax.

mmase: Makes sense! Will dig in and make some changes

mmase: @jeddy3 so far pretty straight forward, but the maps return `false` for the `isStandardSyntaxFunction` as maps aren't given a value by `postcss-value-parser`. Thoughts?

mmase: I pushed a commit up with what seems to produce expected results, though sans `isStandardSyntaxFunction` due to the above.

lib/rules/color-named/index.js
```diff
@@ -58,11 +58,23 @@ const rule = function (expectation, options) {
...
       }
 
+      const functionNodes = []
+
       valueParser(decl.value).walk(node => {
         const value = node.value,
           type = node.type,
           sourceIndex = node.sourceIndex
 
+        if (type === "function") {
```
jeddy3: Could we get a brief comment stating what this is here for as it's not so obvious and I'm sure we'll come back to it in a couple of months time wondering what it does? :)

davidtheclark: Yes, we definitely need a clarifying comment here. Also, I think the variable name `functionNodes` is misleading. You're not actually putting *nodes* in it, right? Could you please try to name that array to match its contents?

mmase:
Made requested changes, let me know if you have any other thoughts on improvements!

lib/rules/color-named/index.js
```diff
@@ -58,11 +58,23 @@ const rule = function (expectation, options) {
...
           type = node.type,
           sourceIndex = node.sourceIndex
 
+        // Find nodes that are inside functions
+        if (type === "function") {
+          node.nodes.forEach((node) => functionNodeIndexes.push(node.sourceIndex))
+        }
+
+        // Return early if node is inside of a function
+        if (functionNodeIndexes.indexOf(sourceIndex) > -1) {
```
davidtheclark: Can we accomplish the same thing in a more standard way with [blurFunctionArguments](https://github.com/stylelint/stylelint/blob/master/lib/utils/blurFunctionArguments.js)? Have you checked out what is happening in other rules that also ignore function arguments?

mmase: Was just about to mention that - as it appears that this is essentially the same behavior as the `ignore` `inside-function` option. Which makes me ask, what is https://github.com/stylelint/stylelint/issues/2009 asking of that can't be achieved with that option?

mmase: Additionally, it appears returning `false` inside of a function node is all that needs to be done.

davidtheclark:
@mmase I think the relevant suggestion in #2009 is

> Perhaps we should change the behaviour to ignore maps by default, and reserve "ignore": ["inside-function"] for just things like $color-footer: color(white) !default;?

from @jeddy3.

If we can't ignore maps without doing the same thing as `"ignore": ["inside-function"]`, then we should just leave things as they are. If we *can* ignore them by some other means, we'd do that.

mmase:
Ah right. Apologies, missed that. Okay, will try a few things as this current implementation feels too harsh.

jeddy3:
> Made requested changes, let me know if you have any other thoughts on improvements!

Thank you!

I think this issue has highlighted a few shortcomings in the tests of this rule as it's not immediately obvious from them if this rule is behaving as expected.

Am I right in thinking that the default behaviour should be:

1. Check color representations inside all standard syntax functions (e.g. when using [color-mod](https://drafts.csswg.org/css-color/#modifying-colors): `a { color: color(black a(50%))`)
2. Ignore non-standard list and map syntax i.e. `(..)`

Then we provide an `"ignore": ["inside-function"]` option for ignoring standard syntax functions.

I think we're at least missing two tests in the `"never"` set that show colours within standard syntax functions are checked e.g. `accept` and `reject` tests for `a { color: color(black a(50%))`.

Add another in `"always-where-possible"` `reject`. (There is one for the `accept` already.)

These I think these new `reject` tests will likely fail with the current implementation in this PR.

> If we can ignore them by some other means, we'd do that.

I thought we might be able to use `postcss-value-parser` to distinguish between standard syntax functions (i.e. those that have a name (e.g. `color(..)` or `map-get(..)) and non-standard syntax ones (i.e. those that don't `(..)`).  @mmase Do you think this might be possible or did you initial investigation turn up a blank?

Thanks for sticking with us on this. It's a deceptively tricky PR.

mmase:
@jeddy3 Sorry for the delay on this. Declaration values _should_ let us see if they are in a named function or non-standard syntax function (no name). However, I'm wondering if we need to go a step further.

For example, in this case, we should ignore `white` and `grey`:

```scss
$colors: (
  white: #000,
  grey: #333
);
```

But, in this case, we should probably still warn:

```scss
$colors: (
  white: white,
  grey: grey
);
```

Do you agree?

davidtheclark:
@mmase Parsing a Sass map is not within the domain of stylelint — rather, a SCSS-specific plugin should take care of that. stylelint itself should just ignore the construction, avoiding mistaken warnings.

mmase:
@davidtheclark fair enough, that makes sense 👍

jeddy3:
> Parsing a Sass map is not within the domain of stylelint — rather, a SCSS-specific plugin should take care of that. stylelint itself should just ignore the construction, avoiding mistaken warnings.

👍 

@mmase  If you're interested, David goes into a bit more background to this problem space [here](http://davidtheclark.com/lessons-i-may-have-learned/#non-standard-syntaxes-are-a-huge-obstacle-to-css-tooling).

And for some context specific to this instance... dollar-variables [are parsed](https://astexplorer.net/#/6l1YfUI8mM) as `decl`s. Nested props are [parsed as](https://astexplorer.net/#/6l1YfUI8mM/1) `decl`s within `rule`s. In both instances, the _value_ part of the declaration is made available within `decl.value`. This _isn't_ the case with lists and maps, [where the `decl.value` contains](https://astexplorer.net/#/6l1YfUI8mM/3) the property-value pairs of the map _as a single string_.

Therefore, it would require _additional_ code within stylelint to parse this non-standard construct.

Which is why the suggestion is to use `isStandardSyntaxFunction` to ignore list and maps without hammering vars and nested props.

mmase:
@jeddy3 thanks for that. `isStandardSyntaxFunction` is great for ignoring non-standard list and map syntax i.e. (..). That makes sense (I was previously using `isStandardSyntaxDeclaration` which is incorrect).

However, it doesn't the remaining issue is ignoring instances of color-mods like `color(red)` or other function calls `myFunction(red)`. Are we looking to only ignore color-mods?

jeddy3:
>  So, that's done and makes sense (I was previously using isStandardSyntaxDeclaration which is incorrect).

👍 

> However, the remaining issue is ignoring instances of color-mods like color(red) or other function calls myFunction(red). Are we looking to only ignore color-mods?

We're ignoring neither _by default_. There is already the `ignore: ["inside-function"]` option that the user can use to opt into this behaviour. (This option will ignore values in standard syntax functions, so that includes functions included in the spec (like `color()`) and custom functions like `myFunction(red)`.)

It sounds like what you've done so far is what we need. If you've also added those tests, then I think you can push up what you've got and we'll be able to see if it's behaving as expected :)

mmase:
I suppose that could be covered by `"ignore": "inside-function"`? If that is the case,

mmase:
Ha, just saw you commented the same :) Okay, will write some tests and push up 👍

mmase:
@jeddy3 @davidtheclark made some changes. thanks for your patience with me on this one. I hope this is the solution you are looking for!

lib/rules/color-named/__tests__/index.js
```diff
@@ -56,6 +56,9 @@ testRule(rule, {
   }, {
     code: "a { font: 10px/14px Brown, sans-serif; }",
     description: "ignore font family names that are colors",
+  }, {
+    code: "colors: (white: #fff, blue: #00f);",
```
jeddy3: Let's use `$colors` here to make it extra clear this is a non-standard.

lib/rules/color-named/index.js
```diff
@@ -70,6 +71,10 @@ const rule = function (expectation, options) {
           return false
         }
 
+        if (!isStandardSyntaxFunction(node)) {
+          return false
```
evilebottnawi: Better use just `return`

mmase: This doesn't appear to work. I don't know a _ton_ about `postcss-value-parser`, but I believe this is because returning  `undefined` on a `value` would work. However, since we are making an assertion on a `node`, you need to return false for it to skip the `value`s within that `node`. Otherwise, I would group it with the `isStandardSyntaxValue` check below this.

jeddy3:
@mmase Ooops, code for my test should be `a { color: color(#000 a(50%)) }`...

mmase:
@jeddy3 ah, yeah I did the same thing a few times as well

mmase:
thanks guys!

davidtheclark:
Changelog entry:

> -   Fixed: `color-named` now ignores SCSS maps, so map property names can be color names ([#2182](https://github.com/stylelint/stylelint/pull/2182))

Issue #2046, opened 2016-11-08
Better test the CLI


davidtheclark:
I do not know of a good way to test the CLI *directly* (maybe others know?); but I do think that we could do more than we are.

What I suggest is that we change `cli.js` such that it passes the `meow` `cli` object to an outside, tested function. That function will be tested against `cli` objects: e.g. what happens if we receive `cli.flags.customReporter`, etc. It should mock `standalone` and just check that `standalone` is called with the expected arguments, and the output is dealt with accordingly.

jeddy3:
> What I suggest is that we change `cli.js` such that it passes the `meow` `cli` object to an outside, tested function. 

This sounds like a good plan to me.

jeddy3:
Closing as stale.

PR #2186, opened 2016-12-10
CLI: Mark some parameters as booleans explicitly

### LINKED ISSUE #2184: --quiet flag behavior is counter-intuitive

A bug (counter-intuitive behavior) in CLI part. A file pattern after the `-q` flag is treated as a flag value, not as a pattern.

_N/A_

_N/A_

_Doesn't matter._

`7.6.0`

CLI with `stylelint -q`

No.

`--quiet` and other boolean flag doesn't accept values

`stylelint -q "**/*.css" "**/*.less"`

is treated as:

`stylelint` `-q` `"**/*.css"` `"**/*.less"`

`stylelint -q "**/*.css" "**/*.less"`

is treated as:

`stylelint` `-q="**/*.css"` `"**/*.less"`

`-q` value is then silently coerced to `true`, so basically it's the same as calling

`stylelint` `-q=1` `"**/*less"`

As a result, instead of expected `**/*.css` and `**/*.less` only `**/*.less` files are checked by the linter with no clue that something went wrong, which is really bad for git hooks.

*Proposed fix:*
Adding 
```
boolean: ["q", "rd"]
```
to `minimistOptions` dict in the `cli.js:34` would explicitly mark `-q` and `-rd` flag as boolean, so they won't accept any parameters.

subzey:

#2184

No.

Fixes #2184

This change states that --allow-empty-input, --color, --help,
--ignore-disables, --no-color, --quiet, --version has no arguments

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -26,6 +26,15 @@ const minimistOptions = {
     v: "version",
     aei: "allow-empty-input",
   },
+  boolean: [
+    "allow-empty-input",
+    "color",
+    "help",
+    "ignore-disables",
+    "no-color",
+    "quiet",
+    "version",
+  ],
 }
 
 const meowOptions = {


jeddy3:
@subzey Thanks! LGTM.

@stylelint/core Tested locally and looks good. Changelog entry is:

> -   Fixed: a boolean CLI flag (e.g. `--quiet`) placed before an input glob no longer causes the input to be ignored ([#2186](https://github.com/stylelint/stylelint/pull/2186)).

Need for better CLI tests is being tracked in https://github.com/stylelint/stylelint/issues/2046

Issue #2148, opened 2016-11-30
color-rgb-only


merlinpatt:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A new rule (feature) that would only allow for rgb and rgba values for color

> What CSS is needed to reproduce this issue?

e.g.

```css
.foo {
  color: rgb(7, 7, 7);
}
```

> Which version of stylelint are you using?

`7.6.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

"CLI with `stylelint --config myconfig *.css`"

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

evilebottnawi:
@merlinpatt How about this http://stylelint.io/user-guide/rules/declaration-property-value-whitelist/ rule?

jeddy3:
@merlinpatt Thanks for the request. What you're after can be achieved with:

- [`color-named`](http://stylelint.io/user-guide/rules/color-named/) and [`function-whitelist`](http://stylelint.io/user-guide/rules/function-whitelist/) (or `function-blacklist`]), if you want to do this for _all_ colours. 
- [`declaration-property-value-whitelist`](http://stylelint.io/user-guide/rules/declaration-property-value-whitelist/), as @evilebottnawi point out, if you want to do this for _just_ the `color` property.

The `*-blacklist` and `*-whitelist` rules are flexible enough, and work together, to enforce most desired conventions when it comes to allowing or disallowing language features.

merlinpatt:
While the whitelist (or blacklist) will work, as wellas the regex options, this seems like something that should be easily declared within the configuration itself as its own rule. It also means that one needs an understanding of regex to configure that, which shouldn't be necessary to configure a linter.

Some of the configuration rules could in fact be done with regex (color-hex-case, color-named for example), yet these are still rules. What determines that these should be rules as opposed to other rules that can also be done via regex?

davidtheclark:
> this seems like something that should be easily declared within the
configuration itself as its own rule

Can you please explain why you think rgb-only enforcement is important to
many users and not just a special case?

On Wed, Nov 30, 2016 at 10:42 AM, Merlin Patterson <notifications@github.com

merlinpatt:
I don't have any research that would prove that this more than just a special case. But I do think that a not insignificant portion of developers use rgb, just as there are a non-insignificant number who use hex.

[...]

PR #2189, opened 2016-12-10
Document how the (dis)allowing rules work together

jeddy3:
PR explained in #2148

### DIFF
--- a/docs/user-guide/about-rules.md
+++ b/docs/user-guide/about-rules.md
@@ -352,7 +352,7 @@ Say you want to disallow the value `none` for the `border` properties. You can d
 
 #### color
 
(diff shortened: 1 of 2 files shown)


docs/user-guide/about-rules.md
```diff
@@ -323,3 +323,66 @@ You can do that with:
...
+
+```js
+"declaration-property-value-blacklist": [{
+  "/^border/": ["none"]
+}]
+```
+
+#### color
+
+Most `<color>` values are *functions*. As such, they can be (dis)allow using either the `function-blacklist` or `function-whitelist` rules. There are two other color representations that aren't functions: named colors and hex colors. There are two specific rules that (dis)allow these: `color-named` and `color-no-hex`, respectively.
```
davidtheclark: (dis)allowed

(-ed)

Issue #2052, opened 2016-11-14
PLEASE HELP: We need more system test cases!


davidtheclark:
We need more test cases for the system tests described in https://github.com/stylelint/stylelint/blob/master/system-tests/writing-system-tests.md.

To provide a test case, all we need is a stylesheet and a configuration file. Preferably the pair ends up with at least one or two lint errors.

**Please share some stylesheet-config pairs that we can use for test cases!** Especially people on @stylelint/core — each of you please try to come up with at least a few.

**Probably the best way to share a test case is to link to a Gist.**

m-allanson:
@davidtheclark Can these cases be used to test some of the path and config based issues that showed up recently? 

If so, here's a list of issues that might be good references for test cases:

https://github.com/stylelint/stylelint/issues/1991
https://github.com/stylelint/stylelint/issues/1990
https://github.com/stylelint/stylelint/issues/1985
https://github.com/stylelint/stylelint/issues/1966
https://github.com/stylelint/stylelint/issues/1646

Or should these cases be more focussed on rules and rule configs?

ntwb:
The PRs above I added, #2097 & #2098 don't have any lint errors, I picked them as they are _popular_ CSS frameworks that already use stylelint so each _pair_ is based on the repos own stylelint config.

davidtheclark:
@m-allanson I think that protecting against a focused and identified regression is best left to the unit tests. We should save the less precise system tests for larger examples that test a lot of things.

jeddy3:
Done in #2192, #2097 & #2098

PR #2192, opened 2016-12-12
System test 005

m-allanson:

#2052 

The config in this test uses every core rule (except `no-empty-source`) and the CSS will trigger a warning from each rule at least once. This generates a large amount of warnings, as a result the snapshot file ~600KB. 

This isn't a particularly realistic test so feel free to close this if it's not suitable.


davidtheclark:
Thanks @m-allanson! Seems like a perfect system test.

Issue #1946, opened 2016-10-01
False negatives with ignore: ["descendant"] in selector-no-type


jeddy3:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Bug.

> Which rule, if any, is this issue related to?

`selector-no-type`

> What CSS is needed to reproduce this issue?

``` css
.class > a {} /* child */
.class + a {} /* next-sibling */
.class ~ a {} /* following-sibling */
```

> What stylelint configuration is needed to reproduce this issue?

``` json
{
  "rules": {
    "selector-no-type": [true, { "ignore": ["descendant"]}],
  }
}
```

> Which version of stylelint are you using?

`7.3.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

Via the demo

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

The following warnings to be flagged:

``` shell
1:9 Unexpected type selector
2:9 Unexpected type selector
3:9 Unexpected type selector
```

> What actually happened (e.g. what warnings or errors you are getting)?

No warnings.

It looks like [this is bit of code](https://github.com/stylelint/stylelint/blob/master/lib/rules/selector-no-type/index.js#L82-L85) too permissive. Checking for _any_ type of combinators, rather than just the descendant one.

Ref: https://github.com/stylelint/stylelint/issues/1942

This is a breaking change. Post `8.0.0` this would be a _minor_ update, but prior to `8.0.0` it's a _major_.

pixelass:
having the same issue. It is one of the most important features (IMHO) therefore I can't use it until it is fixed. It breaks the entire need for a linter.

jeddy3:
> It is one of the most important features (IMHO) therefore I can't use it until it is fixed.

@pixelass Would you like to contribute this fix? We created a [developer userguide](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md) to help people get started in contributing to this open source project. All help is welcomed and appreciated.

pixelass:
@jeddy3 sure, I'd love to contribute. sadly I'm very busy this month. Since I am using this at work I might be able to invest some of that time.

pixelass:
I'm having a few problems running the tests. There are over 23000 tests which are all logged. 
Even if I test only this rule there are 90+ tests.

Wouldn't it make more sense to only logg errors? (not familiar with tape)
[...]

PR #2200, opened 2016-12-17
Fix false negatives with ignore: ["descendant"] in selector-no-type

m-allanson:

Refs #1946 

Note that this will merge into the `v8` branch, as it's a breaking change (until after `v8` is released).


jeddy3:
@m-allanson Thanks!

Added to `v8` changelog as:

> -   Fixed: `selector-no-type` no longer produces false negatives for when child, next-sibling and following-sibling combinators are used with `ignore: ["descendant"]` ([#2200](https://github.com/stylelint/stylelint/pull/2200)).

PR #2197, opened 2016-12-15
Deprecate rules

m-allanson:
> Which issue, if any, is this issue related to?

#2123 

> Is there anything in the PR that needs further explanation?

Merges into a new `deprecations` branch.

Issue #2079, opened 2016-11-19
Plugin candidates for 8.0.0


jeddy3:
There are over 170 rules in stylelint.

There are the 48 `*-before`, `*-after` and `*-inside` variants of the `*-space-*`, `*-newline-*` and `*-unspaced-*` [rules](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/about-rules.md#whitespace-rules). 

There are the 11 `*-empty-line-before` and `*-max-empty-line` rules.

And there is the `indentation` rule. I think these rules work together to allow the enforcement of most [whitespace code styles](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/about-rules.md#rules-work-together). More so than any CSS linter has done before.

Then there are: 

- 10 `*-pattern` rules.
- 10 `*-case` rules.
- 4 `*-quotes` rules.
- 4 `number-*` and `length-*` rules.
- 2 `*-notation` rules. 
- 2 `*-length` rules. 
- `no-eol-whitespace`
- `no-missing-end-of-source-newline`

I think these 34 rules cover off most other _code style_ requirements.

Next, there are 31 rules for (dis)allowing _language constructs_:
- 20 `*-whitelist` and `*-blacklist` rules. 
- 11 applicable `*-no-*` rules:
   - 7 `selector-no-*` rules.
   - 2 `*-no-important` rules.
   - 2 `color-no-*` rules.

I think the breadth of these rules is excellent and covers nearly all of the language constructs available in CSS.

There are these 4 additional _limiting_ rules:

- `declaration-block-single-line-max-declarations`
- `max-nesting-depth`
- `selector-max-compound-selectors`
- `selector-max-specificity`

And these 7 rules that check valid but likely unintentional code:

- 3 `*-duplicate-*` rules:
  - `declaration-block-no-duplicate-properties`
  - `font-family-no-duplicate-names`
  - `no-duplicate-selectors`
- 2 `*-no-redundant-*` rules:
  - `declaration-block-no-redundant-longhand-properties`
  - `shorthand-property-no-redundant-values`
- 2 "overrides"
  - `declaration-block-no-shorthand-property-overrides`
  - `no-descending-specificity`

There are 5 `*-no-vendor-prefix` rules.

And finally, there are the 3 `*-no-empty` rules discussed in https://github.com/stylelint/stylelint/issues/2047#issuecomment-261382105.

[...]

Issue #1663, opened 2016-07-17
Make *-empty-line-before keyword options consistent


jeddy3:
#1653 (comment) - brought some clarity to how the `*-empty-line-before` secondary options should work.

A couple of the existing options are inconsistent.
- [`comment-empty-line-before`'s `"between-comments"`](https://github.com/stylelint/stylelint/tree/master/lib/rules/comment-empty-line-before#between-comments) → `"after-comment"`
- [`at-rule-empty-line-before`'s `"blockless-group"`](https://github.com/stylelint/stylelint/tree/master/lib/rules/at-rule-empty-line-before#blockless-group) → `"blockless-after-blockless"`

This is consistent with the existing keywords of `"after-comment"`, `"after-single-line-comment"` and `"first-nested"`. And the new keywords of `"after-custom-property"` and `"after-declaration"`.

We can decide whether to leave them be, or change them in the next major release nearer the time. I wanted to write this issue up to highlight these inconsistencies so that any new options we added going forward are consistent.

We've a proposal for a new option in https://github.com/stylelint/stylelint/issues/1366. Perhaps we should change the proposal to be more consistent with these conventions:
- `at-rule-empty-line-before`'s `except: "matching-name-blockless-group"` →  `except: "blockless-after-same-name-blockless"`. The first `"blockless"` describes the thing itself and `"after-same-name-blockless"` describes its relationship to its previous sibling.

Sound about right?

Edit: `"all-nested"` -> `"inside-block"`, `"stylelint-commands"` -> `"stylelint-command"`

jeddy3:
https://github.com/stylelint/stylelint/pull/2188 made we realise that our `"first-nested"` options should be called `"after-opening-brace"` for consistency with all the other `"after-*"` options. I think it's also clearer.

As an aside, I think our `*-before` rules with their `"after-*"` options seem to be standing up well as more use cases are discovered and accounted for.

ntwb:
SGTM... 

_We could duplicate the option names and save off having to deprecate them immediately, they could live side by side for a while..._

m-allanson:
Based on the above discussion here's a list of rule options that will be affected by this issue. If there's no objections I can (probably) start implementing these changes later in the week.

| Rule name | Current option name | Updated option name | 
| --- | --- | --- |
[...]

PR #2207, opened 2016-12-24
Ensure absolute paths to processors

### LINKED ISSUE #2194: Document what relative `files` globs are relative to
This seems to be missing from the docs.

I believe they are relative to `process.cwd()`.

### LINKED ISSUE #2195: Only pass processor functions absolute filepaths
I ran into a situation where the `code` processor function's filepath argument was relative, but the `result` processor function's argument was absolute. This makes it hard to associate input and output in the processor code.

We should provide an absolute path to both. We should also document this.

davidtheclark:
Closes #2194, closes #2195.

Although this does not add more tests, it does add some errors that break things (causing tests to fail) if absolute paths are for some reason not coming through. More focused unit tests looked like more trouble to figure out than they were worth, so I think this does the trick.

I'd like to cut a release for this as soon as possible, so processors stop failing. How about finishing up this and #2197 and then releasing?

### DIFF
--- a/lib/postcssPlugin.js
+++ b/lib/postcssPlugin.js
@@ -3,16 +3,24 @@
 const _ = require("lodash")
 const createStylelint = require("./createStylelint")
 const postcss = require("postcss")
+const path = require("path")
 
-module.exports = postcss.plugin("stylelint", function () {
-  const options = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : {}
+module.exports = postcss.plugin("stylelint", function (options) {
+  options = options || {}
 
-  const tailoredOptions/*: Object*/ = options.rules ? { config: options } : options
-
+  const tailoredOptions/*: Object*/ = options.rules
+    ? { config: options }
+    : options
   const stylelint = createStylelint(tailoredOptions)
+
   return (root, result) => {
+    let filePath = options.from || _.get(root, "source.input.file")
+    if (filePath !== undefined && !path.isAbsolute(filePath)) {
+      filePath = path.join(process.cwd(), filePath)
+    }
+
     return stylelint._lintSource({
-      filePath: options.from || _.get(root, "source.input.file"),
+      filePath,
       existingPostcssResult: result,
     })
   }
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -1,5 +1,6 @@
 /* @flow */
 "use strict"
+const path = require("path")
 const formatters = require("./formatters")
 const createStylelint = require("./createStylelint")
 const globby = require("globby")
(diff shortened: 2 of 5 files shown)


m-allanson:
@davidtheclark it might be nice to include #1663 in the next release too, so the deprecations are landed at once. Although I've not done any work on that yet.

lib/postcssPlugin.js
```diff
@@ -3,16 +3,24 @@
...
+  options = options || {}
 
+  const tailoredOptions/*: Object*/ = options.rules
+    ? { config: options }
+    : options
   const stylelint = createStylelint(tailoredOptions)
+
   return (root, result) => {
+    let filePath = options.from || _.get(root, "source.input.file")
+    if (filePath !== undefined && !path.isAbsolute(filePath)) {
```
m-allanson: Is this pattern the same here and in `standalone.js`? If so, is it possible to create a util named something like `absolutizePath()` and use that?

lib/standalone.js
```diff
@@ -51,7 +52,13 @@ module.exports = function (options/*: Object */)/*: Promise<stylelint$standalone
   })
 
   if (!files) {
-    return stylelint._lintSource({ code, codeFilename }).then(postcssResult => {
+    const absoluteCodeFilename = (codeFilename !== undefined && !path.isAbsolute(codeFilename))
```
m-allanson: See comment about creating a util.

lib/standalone.js
```diff
@@ -70,7 +77,12 @@ module.exports = function (options/*: Object */)/*: Promise<stylelint$standalone
     }
 
     const getStylelintResults = filePaths.map(filePath => {
-      return stylelint._lintSource({ filePath }).then(postcssResult => {
+      const absoluteFilepath = (!path.isAbsolute(filePath))
```
m-allanson: See comment about creating a util.

lib/postcssPlugin.js
davidtheclark: I don't think this pattern is util-worthy. It's just very straightforward JS.

jeddy3:
Added to changelog as:

> -   Fixed: ensure only absolute filepaths are passed to processors ([#2207](https://github.com/stylelint/stylelint/pull/2207)).

---

> I'd like to cut a release for this as soon as possible, so processors stop failing. 

Oops, I just saw this. I got caught up eating mince pies. I can probably do a release today - I'll create an issue.

> it might be nice to include #1663 in the next release too, so the deprecations are landed at once.

The [final list of deprecations](https://github.com/stylelint/stylelint/issues/2079) still needs to be decided on. Hence the `deprecations` branch. I hope to have time to contribute to the various rule discussions forward in the next few days. In the meantime I don't think there's anything blocking a release with this processors fix.

davidtheclark:
Hope you enjoyed your holiday!

jeddy3:
I did thank you! I trust you had an enjoyable xmas too! :)

PR #2217, opened 2016-12-29
Add ignore: ["child"] option to selector-no-type

### LINKED ISSUE #1942: Add ignore: ["child"] option to selector-no-type

Request for enhancement in `selector-no-type`

[`selector-no-type`](https://github.com/stylelint/stylelint/tree/master/src/rules/selector-no-type) 

```
.className > label {} // I want to allow this...
.className label {}   // ...but not this
```

e.g.

``` json
{
  "rules": {
        "selector-no-type": [true, {
            ignore: ["compounded", "descendant"],
        }],
  }
}
```

`7.3.1`.

CLI

no

Hmm. With that config, perhaps just the 1st one should warn, because it's a child selector, not descendant. It would be good to control descendants and children individually. Then I could allow the 1st but flag the 2nd line.

With this config, both lines are fine. When I remove `descendant`, both are flagged as errors.

m-allanson:

Closes #1942

Merging into the `v8` branch as it works well with the changes from #2200. I've rebased `v8` on to `master`.

### DIFF
--- a/lib/rules/selector-no-type/README.md
+++ b/lib/rules/selector-no-type/README.md
@@ -50,7 +50,25 @@ The following patterns are *not* considered warnings:
 
 ## Optional secondary options
 
-### `ignore: ["compounded", "descendant"]`
+### `ignore: ["child", "compounded", "descendant"]`
+
+#### `"child"`
+
+Allow child selectors.
+
+The following patterns are *not* considered warnings:
+
+```css
+a > foo {}
+```
+
+```css
+a > #bar {}
+```
+
+```css
+#bar > a {}
+```
 
 #### `"compounded"`
 
--- a/lib/rules/selector-no-type/README.md
+++ b/lib/rules/selector-no-type/README.md
@@ -59,15 +59,11 @@ Allow child selectors.
 The following patterns are *not* considered warnings:
 
 ```css
-a > foo {}
-```
-
-```css
-a > #bar {}
-```
-
-```css
 #bar > a {}
+```
+
+```css
+#bar > a.foo {}
 ```
 
 #### `"compounded"`
--- a/lib/rules/selector-no-type/__tests__/index.js
+++ b/lib/rules/selector-no-type/__tests__/index.js
@@ -165,12 +165,83 @@ testRule(rule, {
 
 testRule(rule, {
   ruleName,
+  config: [ true, { ignore: ["child"] } ],
+  skipBasicChecks: true,
+
+  accept: [ {
+    code: "div > foo {}",
+    description: "type and child",
+  }, {
+    code: "div > #foo {}",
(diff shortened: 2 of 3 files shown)


m-allanson:
Proposed changelog entry: 
-   Added: `ignore: ["child"]` option to `selector-no-type` ([#1942](https://github.com/stylelint/stylelint/issues/1942))

lib/rules/selector-no-type/README.md
```diff
@@ -50,7 +50,25 @@ The following patterns are *not* considered warnings:
...
+a > foo {}
+```
+
+```css
+a > #bar {}
+```
+
+```css
+#bar > a {}
+```
```
davidtheclark: Either I'm not understanding the intention, or something is wrong here. In the first two cases, `a` is *not* a child, therefore it *should* be considered a warning, right? Only the last case has no type selectors that are not children. If the intention is that you always allow types in compound selectors that involve child combinators — I don't think I understand the justification for that. I *think* the original request in #1942 only addressed *type selectors following child combinators* — that is, types that are children.

jeddy3: Whoops, I totally missed that.

> Either I'm not understanding the intention, or something is wrong here. 

You are correct.

> only addressed type selectors following child combinators 

I think this is the correct behaviour and I think is consistent with how the current `"descendant"` option works. Although, it'd be nice to have some tests [here](https://github.com/stylelint/stylelint/blob/2ddb4a5bf3ef5d7518d599bddeb5a210797d9323/lib/rules/selector-no-type/__tests__/index.js#L114-L120) to definitely show that things like `a .foo {}` and `a div {}` are rejected.

m-allanson:
I'm not sure what I was thinking there, thanks for catching it. PR is updated.

lib/rules/selector-no-type/__tests__/index.js
```diff
@@ -165,6 +171,66 @@ testRule(rule, {
...
+    code: ".foo ~ div {}",
+    description: "following-sibling",
+    message: messages.rejected,
+  }, {
+    code: "div > foo {}",
+    description: "type and child type",
+  }, {
+    code: "div > a {}",
+    description: "type and child type",
+    message: messages.rejected,
```
davidtheclark: Good idea to add line and column here so it's clear which type selector is violating the rule.

m-allanson: 👍  PR updated.

jeddy3:
Updated `v8` changelog with @m-allanson suggestion:

> -   Added: `ignore: ["child"]` option to `selector-no-type` ([#1942](https://github.com/stylelint/stylelint/issues/1942)).

Issue #2219, opened 2016-12-29
Add ignorePattern: "" option to max-line-length


handonam:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Using max-line-length with a long import statement causes the error to be thrown.  Should have an option to ignore patterns (much like eslint, with `{ "ignorePattern": "^import\\s.+\\sfrom\\s.+;$" }],`), or some way to ignore `@imports`.

> Which rule, if any, is this issue related to?

`max-line-length`

> What CSS is needed to reproduce this issue?

```css
@import (reference) "some/really/long/path/to/a/file/that/is/over/the/rainbow/or/something/likethat.scss";
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "max-line-length": 100,
  }
}
```

> Which version of stylelint are you using?

`7.7.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

atom plugin / PostCSS Plugin

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No.

> What did you expect to happen?

Ignore imports when something like ignorePattern is specified

> What actually happened (e.g. what warnings or errors you are getting)?

```shell
Expected line length to be no more than 100 characters
```

ntwb:
Thanks for creating this issue and using our template @handonam 👍 

I've confirmed the behaviour you describe though it is only when using `@import "something"`, if including the `url` in the statement such as `@import url(something)` then this rule works as expected. I've added some tests in https://github.com/stylelint/stylelint/pull/2221/commits/480c3fc01db30a748cd900eb3ab640c5a18e661d confirming this.

Currently `max-line-length` [uses the regex](https://github.com/stylelint/stylelint/blob/master/lib/rules/max-line-length/index.js#L37) `/url\(.*\)` to detect URLs, thus it doesn't detect the valid use of `@import` without the URL string.

Via https://developer.mozilla.org/en-US/docs/Web/CSS/@import :
```css
@import url("fineprint.css") print;
@import url("bluish.css") projection, tv;
@import 'custom.css';
@import url("chrome://communicator/skin/");
@import "common.css" screen, projection;
@import url('landscape.css') screen and (orientation:landscape)
```

Rather than adding an ignore option to the rule I've updated the rule regex adding support to also detect `@import "something"` in https://github.com/stylelint/stylelint/pull/2221/commits/926e35e5e4628235aef938ef92c8517f2ddc9094

[...]

PR #2221, opened 2016-12-30
Add `@import` tests for `max-line-length` rule

ntwb:

Initial tests for #2219


ntwb:
Following the changes in https://github.com/stylelint/stylelint/pull/2221/commits/926e35e5e4628235aef938ef92c8517f2ddc9094 the jest snapshot for `005` needed updating, is this the correct way to do that in this commit https://github.com/stylelint/stylelint/pull/2221/commits/fdbd4faa96a21847dc58defcbc242d549dc9f330 ?

jeddy3:
> is this the correct way to do that in this commit fdbd4fa ?

If you updated the snapshot via the interactive testing prompt, then yes.

jeddy3:
Is the plan here to ignore `@import` statements _by default_ and _in addition to_ [adding an](https://github.com/stylelint/stylelint/issues/2219) `ignorePattern` option?

lib/rules/max-line-length/index.js
```diff
@@ -34,7 +34,11 @@ const rule = function (maxLength, options) {
 
     // Collapse all urls into something nice and short,
     // so they do not throw the game
-    const rootString = root.toString().replace(/url\(.*\)/ig, "url()")
+    const rootString = root.toString()
+    // Replace `url(something)` with `url()`
+    .replace(/url\(.*\)/ig, "url()")
+    // Replace `@import "something"` with `@import ""`
+    .replace(/\@import\s+['"].*['"]/ig, "@import ")
```
davidtheclark: Unfortunately, I do not think this regexp is exact enough. for example, `@import 'svg-something<id="horse">';` will become `@important ">';`, right?

system-tests/005/__snapshots__/005.test.js.snap
```diff
@@ -11765,169 +11765,169 @@ Array [
         "text": "Expected line length to be no more than 20 characters (max-line-length)",
       },
       Object {
-        "column": 5,
-        "line": 146,
+        "column": 25,
+        "line": 145,
```
davidtheclark: Something seems wrong with this system test. Am I right that these line numbers do not correspond with lines in `005/stylesheet.css` that make sense? Let's sort this out before accepting changes to it.

lib/rules/max-line-length/index.js
ntwb: Thanks @davidtheclark, I think what I have is OK (for my sanity I use http://regexr.com/ to test/explain regexes). There is probably room for improvement as I'm not the greatest regex architect.

Using `\@import\s+['"].*['"]` and replacing with `@import `

With this string: `@import 'svg-something<id="horse">';`

Results in `@import ;` (There is a single space between the `t` and `;`)

I've added a test to accept that the length of this `@import 'svg-something<id=\"horse\">' projection;` string is 20 characters or under, stripping `'svg-something<id=\"horse\">'` leaves `@import  projection;` which is exactly 20 characters in length.

I've also added a failing test using `@import 'svg-something<id=\"horse\">' screens, tv;` which once stripping `'svg-something<id=\"horse\">'` leaves `@import screens, tv;` which is 21 characters in length.

system-tests/005/__snapshots__/005.test.js.snap
ntwb: Trying to wrap my head around this and it's not going so well 😏 

Currently in `master` before changes by this PR we have:

https://github.com/stylelint/stylelint/blob/master/system-tests/005/stylesheet.css#L146
> `b a {}`  <- 6 characters in length

https://github.com/stylelint/stylelint/blob/master/system-tests/005/__snapshots__/005.test.js.snap#L11767-L11773
```
      Object {
        "column": 5,
        "line": 146,
        "rule": "max-line-length",
        "severity": "error",
        "text": "Expected line length to be no more than 20 characters (max-line-length)",
      },
```

Once applying the changes in this PR the above warning for line 146 is no longer reported in the _test snapshot_ but a new one is, line 149 with `.bar,` is now reported in the  test snapshot whereas it wasn't previously.

It appears that there are numerous cases of the above, the only sane piece of data I've been able to extract thus far is that both before and after this PR there are a total of 84 `max-line-length` warnings, when i manually remove all lines from the `stylesheet.css` file less than 20 characters in length I indeed have 84 lines greater than 20 characters which correlates.

Line numbers, are at best a _crap shoot_ 🎲 

If anyone else in @stylelint/core can make any sense of this it'd be appreciated 😄

m-allanson: There's _something_ going on here, but I'm not sure what. If I can reproduce a smaller test case I'll post it up.

ntwb: One possability is that Jest is _serializing_ the `snapshot.css` file:
https://facebook.github.io/jest/docs/configuration.html#snapshotserializers-array-string

Unconfirmed at this stage though, need to read more docs and or code to confirm.

m-allanson: I think this is the problem: https://github.com/stylelint/stylelint/issues/2286

ntwb: @m-allanson I actually didn't even get to the stage of checking if the column number was correct, I've only really tested line numbers, and it's been clear that the line numbers referenced are not 20 characters in length (based on this PRs settings)

m-allanson: @ntwb Yeah it's a funny one - the incorrect column numbers can compound and spill over into incorrect line numbers too. I added an extra example to the bottom of the issue that demonstrate.

lib/rules/max-line-length/index.js
davidtheclark: Ooooo, I think my misunderstanding was in forgetting that the `*` is *non-greedy by default*, so the last `['"]` will match the last possible quotation mark instead of first one it finds. You're right: carry on!

davidtheclark:
@ntwb Looks like the problems with the `max-line-length` warnings in the snapshot are being addressed in #2286, so I'll merge this PR. There are no changes for the changelog here, right? Just new tests?

ntwb:
Thanks everyone, it was a hoot deep diving through code for this one 👍

PR #2232, opened 2017-01-02
Add stylelint-disable-no-empty

### LINKED ISSUE #1814: Add stylelint-disable-no-empty rule
Disable empty `stylelint-disable` in files.
Invalid:

``` css
/* stylelint-disable */
.selector {
  color: red;
}
```

``` css
.selector {
  color: red; // stylelint-disable
}
```

Valid:

``` css
/* stylelint-disable color-named */
.selector {
  color: red;
}
```

``` css
.selector {
  color: red; // stylelint-disable color-named
}
```

Some people in the team think that turning off `stylelint`, review will be faster accepted. Be good if CI down for this bad guys :smiling_imp:

diagramatics:
Resolves #1814.

What I can't figure out is why the tests are failing though...

### DIFF
--- a/lib/rules/stylelint-disable-no-empty/index.js
+++ b/lib/rules/stylelint-disable-no-empty/index.js
@@ -1,0 +1,49 @@
+"use strict"
+
+const report = require("../../utils/report")
+const ruleMessages = require("../../utils/ruleMessages")
+const validateOptions = require("../../utils/validateOptions")
+
+const ruleName = "stylelint-disable-no-empty"
+
+const messages = ruleMessages(ruleName, {
+  rejected: "Unexpected empty `stylelint-disable` comment",
+})
+
+const stylelintDisableCommand = "stylelint-disable"
+
+const rule = (actual) => {
+  return (root, result) => {
+    const validOptions = validateOptions(result, ruleName, { actual })
+    if (!validOptions) {
+      return
+    }
+
+    root.walkComments((comment) => {
+      if (!isDisableCommand(comment.text)) {
+        return
+      }
+
+      const specificRules = comment.text.split(" ", 2).slice(1)
+
+      if (specificRules.length > 0) {
+        return
+      }
+
+      report({
+        message: messages.rejected,
+        node: comment,
+        result,
+        ruleName,
+      })
+    })
+
+    function isDisableCommand(text) {
+      return text.indexOf(stylelintDisableCommand) === 0
+    }
+  }
+}
+
+rule.ruleName = ruleName
+rule.messages = messages
+module.exports = rule
--- a/lib/rules/stylelint-disable-no-empty/index.js
+++ b/lib/rules/stylelint-disable-no-empty/index.js
@@ -30,12 +30,17 @@ const rule = (actual) => {
         return
       }
 
+      const disabledRanges = result.stylelint.disabledRanges
+      result.stylelint.disabledRanges = false
+
       report({
         message: messages.rejected,
         node: comment,
         result,
         ruleName,
       })
+
+      result.stylelint.disabledRanges = disabledRanges
     })
 
     function isDisableCommand(text) {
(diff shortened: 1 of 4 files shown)


davidtheclark:
@diagramatics Thanks for contributing. You can click on `Details` to see the test output, or run the tests locally. Not sure there's much to review here until we have passing tests. Do you have specific questions?

diagramatics:
Yeah I'm not sure what happened here, I've checked over and over again and I can't find out what's causing the tests to fail.

Logging on `jest-setup.js` tells me that `output.results[0].warnings` would always return an empty array from any of the tests, even though I've debugged my code and `report()` definitely gets called. I'm pretty sure I'm missing something obvious here but after comparing with other rule implementations I still can't figure out what's missing.

Sorry for this, but I'm going to need a bit of handholding on this one 😕

davidtheclark:
@diagramatics Ooh, you have a tricky problem. The errors are not being reported because the line is disabled! See https://github.com/stylelint/stylelint/blob/master/lib/utils/report.js#L53 where `report` determines whether to file a complaint or not.

I can think of two approaches you could take (maybe you can think of more):
- Change the way `disabledRanges` are determined so they do not include the comment itself.
- Add a flag to `report` to bypass `disabledRanges` for a rule. We'd then have to add a note that this particular rule cannot be disabled with comments.

diagramatics:
Ah that makes sense! I just saw the same bypass system in `stylelint-disable-reason` so I just used the same technique in there if that's okay.

Thanks for the help!

lib/rules/stylelint-disable-no-empty/index.js
```diff
@@ -0,0 +1,54 @@
...
+    if (!validOptions) {
+      return
+    }
+
+    root.walkComments((comment) => {
+      if (!isDisableCommand(comment.text)) {
+        return
+      }
+
+      const specificRules = comment.text.split(" ", 2).slice(1)
```
davidtheclark: I don't remember: is `comment.text` trimmed? Just checking if this is precise enough, even in cases where there are multiple spaces, e.g. `/*   stylelint-disable   */`

diagramatics: Yeah it seems like it does get trimmed, but maybe it should be more precise than this.

For example doing `/* stylelint-disable , */` would actually pass the rule, but I don't think I can validate if the rule itself is correct, or maybe it should be separate to this rule. Then you have `/* stylelint-disable  block-no-empty */` with two spaces in the middle causing that part of the code to think there are two rules in that line. I might tweak that to trim spaces around the list of rules being disabled.

Regardless I've added a few test cases that account for weird spacing anyway.

davidtheclark: > I don't think I can validate if the rule itself is correct

🤔  ... without that, I'm not sure this rule has a lot of value. If `stylelint-disable .` passes, it's not doing much to prevent abuse.

diagramatics: Yeah this is a bit problematic. Does this mean we should refactor the code to have a separate place to reference all of the rules?

jeddy3: > Does this mean we should refactor the code to have a separate place to reference all of the rules?

Tricky. I think it'll need access to the `"rules"` property of the _configuration object_ as _plugins_ can also be disabled using `stylelint-disable`.  @davidtheclark Is that even possible and would we want to?

davidtheclark: Tricky, indeed. I think this validation might actually work better not as a rule, but as a different kind of violation, like `reportNeedlessDisables`. I think:

- We should always report the disabling of a rule that does not actually exist.
- We could optionally report disables without a rule, `reportUnscopedDisables` or something.

You can see that [where `assignDisabledRanges`](https://github.com/stylelint/stylelint/blob/master/lib/lintSource.js#L89) is invoked, it has access to the configuration — so this is certainly possible.

jeddy3: > Tricky, indeed. I think this validation might actually work better not as a rule, but as a different kind of violation, like `reportNeedlessDisables`.

Agreed.

@diagramatics Sorry to mess you around. The limitations of this rule didn't become apparent until it was implemented. 

We should close this PR, and try to harvest what we can of it for @davidtheclark proposal. I'll create new separate issues for those.

diagramatics: That's all good, sorry it didn't work out! I'll pick up something else to work on in the near future hopefully 😄

Issue #2220, opened 2016-12-29
Add no-unknown-custom-properties rule


dylang:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

New rule: Check that all css variables have been declared.

> Which rule, if any, is this issue related to?

I think this may need to be a new rule.

> What CSS is needed to reproduce this issue?

```css
:root {
    --colorBlue: #3e92e6;
}
```

```css
:root {
    --colorPrimary: var(--colorBlue); // no problem
    --colorUndefined: var(--colorNeverBeenSeenBefore); // oh no!
}
```

> What stylelint configuration is needed to reproduce this issue?

n/a

> Which version of stylelint are you using?

`7.7.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

cli

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

CSS: https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables

> What did you expect to happen?

```shell
test.css
1:21    Unknown variable colorNeverBeenSeenBefore (no-unknown-variables)
```

> What actually happened (e.g. what warnings or errors you are getting)?

"No warnings to be flagged."

evilebottnawi:
@dylang Good rule :+1:  @stylelint/core what do you think about the support new rules in core (`custom-selectors`, `variables` and other from new spec)

jeddy3:
@dylang Thanks for the suggestion and for using the template. I agree that it's a nice rule. 

I think the discussion around whether it belongs in core or as part of a plugin pack is connected to `no-unknown-animations` inclusion in https://github.com/stylelint/stylelint/issues/2079.

We might as well use this issue to discussion these two rules. (We agreed to group together rules with similar issues for easier discussion).

I think these two rules, arguably, fail to adhere to the following two [criteria](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#criteria-for-inclusion):

> Useful to the majority of users
> Has a clear and unambiguous finished state.

These rules are applicable to users who include their custom properties and at-keyframes definitions _within the same source that they are used_ i.e. those users who use methodologies similar to SuitCSS or CSS Modules. These rules aren't applicable if the custom properties and at-keyframes definitions are located _in sources separate from where they are used_, e.g. within an `animations.css` and `variables.css` files, (unless stylelint is used _after_ the sources have been processed and concatenated - an uncommon use-case?). 

[...]

PR #2244, opened 2017-01-08
Update no-unknown-animations docs

jeddy3:
Ref: https://github.com/stylelint/stylelint/issues/2220#issuecomment-269788153 & https://github.com/stylelint/stylelint/issues/2220#issuecomment-269809806

Vocab ref for extended description: https://www.w3.org/TR/css3-animations/#keyframes


davidtheclark:
👍

PR #2248, opened 2017-01-08
Document warning for standard values in no-vendor-prefix

### LINKED ISSUE #2203: "value-no-vendor-prefix" doesn't catch {white-space: -webkit-pre-wrap;} etc

I'm running stylelint over legacy code, and it failed to catch vendor prefixed pre-wrap values.
It catches other vendor prefixed values such as `a { display: -webkit-flex; }`, so it seems to be specific to pre-wrap.

I'm guessing that pre-wrap may never have required vendor prefixes, which is why the test doesn't work. However, it would be really useful if stylelint could still catch issues like this.

"value-no-vendor-prefix"

```css
.wrapword {
    white-space: -moz-pre-wrap !important;  /* Mozilla, since 1999 */
    white-space: -pre-wrap;      /* Opera 4-6 */
    white-space: -o-pre-wrap;    /* Opera 7 */
    white-space: pre-wrap;       /* css-3 */
    white-space: -webkit-pre-wrap; /* Newer versions of Chrome/Safari */
    white-space: normal;
}
```

e.g.

```json
{
    "rules": {
        "indentation": 4,
        "block-no-empty": null,
        "at-rule-no-vendor-prefix": true,
        "media-feature-name-no-vendor-prefix": true,
        "property-no-vendor-prefix": true,
        "selector-no-vendor-prefix": true,
    }
}
```

7.6.0

gulp-stylelint

not that I'm aware. The file is SCSS, but the code that it's failing to catch is standard CSS

Expected a warning

No warnings

jeddy3:
Closes #2203

### DIFF
--- a/lib/rules/value-no-vendor-prefix/README.md
+++ b/lib/rules/value-no-vendor-prefix/README.md
@@ -8,6 +8,8 @@ a { display: -webkit-flex; }
  *  These prefixes */
 ```
 
+This rule will only warn for prefixed _standard_ values, and not for prefixed _proprietary_ or _unknown_ ones.
+
 ## Options
 
 ### `true`
--- a/lib/rules/value-no-vendor-prefix/README.md
+++ b/lib/rules/value-no-vendor-prefix/README.md
@@ -8,7 +8,7 @@ a { display: -webkit-flex; }
  *  These prefixes */
 ```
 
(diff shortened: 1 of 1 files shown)


davidtheclark:
🎖

Issue #2256, opened 2017-01-13
Bug: selector-no-type - Ignore compounded allows types compounded by sibling selector


hankthewhale:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?
In the [docs](http://stylelint.io/user-guide/rules/selector-no-type/), ignoring compounded seems to describe the ideal ignore pattern. `h1.class`. 

However, it also allows two element types to be compounded by the direct sibling selector `input[type='checkbox']:checked + label // does not fail`.

In this case, you can still style element types and sneak past the lint rule.

> Which rule, if any, is this issue related to?

selector-no-type

> What CSS is needed to reproduce this issue?

```css
input[type='checkbox']:checked + label {
  background: black;
}
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
     "selector-no-type": [true, {
        "ignore": ["compounded"]
      }]
  }
}
```

> Which version of stylelint are you using?

`7.7.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

`npm run stylelint`
`stylelint \"client/**/*.css\"`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

Warnings to return

> What actually happened (e.g. what warnings or errors you are getting)?

No warnings are returned. However, warnings are returned when removing `ignore: ['compounded']` from the rule option

jeddy3:
@hankthewhale Thanks for using the template.

I can't replicate this I'm afraid.

Given:

```css
input[type='checkbox']:checked + label {
  background: black;
}
```

The following config:

```json
{
  "rules": {
     "selector-no-type": [true, {
        "ignore": ["compounded"]
      }]
  }
}
```

Gives the following output:

```console
1:34 error Unexpected type selector (selector-no-type)
```

Whereas the following config:

```json
{
  "rules": {
     "selector-no-type": true
  }
}
```

Gives the following output:

```console
1:1 error Unexpected type selector (selector-no-type)
1:34 error Unexpected type selector (selector-no-type)
```

It looks like each _compounded/simple_ selector within a _complex_ selector is being correctly checked individually.

Can you create a reproducible repo for us so we can get to the bottom of this please?

hankthewhale:
@jeddy3. Interesting. Yeah, I'll try to get an example repo up today. Thanks for looking into it.

hankthewhale:
[...]

PR #2258, opened 2017-01-14
Added milestones and issue search links to the issue template

hankthewhale:

https://github.com/stylelint/stylelint/issues/2256

I didn't check the v8 changelog before creating a duplicate bug report. Having this message in the issue template could help prevent myself and others from doing that in the future.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -61,6 +61,7 @@ test.css
 Here are the best ways to help resolve your issue:
 1. Figure out what needs to be done, propose it, and then write the code and submit a PR.
 2. If your issue is a bug, consider at least submitting a PR with failing tests.
+3. Check the [V8 changelog](https://github.com/stylelint/stylelint/blob/v8/CHANGELOG.md) as the bug may have been already resolved.
 
 Note: GitHub issues are for stylelint bugs and enhancements, if you're looking for help or support with stylelint then stackoverflow is our preferred question and answer forum - http://stackoverflow.com/questions/tagged/stylelint
 -->
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -56,12 +56,18 @@ e.g. "The following warnings were flagged:"
 test.css
 1:21    Expected 2 spaces (indentation)
 ```
+<!--
+Before posting, please check that your issue:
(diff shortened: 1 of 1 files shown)


davidtheclark:
I'm not sure this is necessary. That means part of the issue template will have to change as we do future releases. We could add a note that you should search through issues and pull requests in the repo to see if your concern has come up before — everybody should always do that for any repo before submitting an issue.

hankthewhale:
> That means part of the issue template will have to change as we do future releases

@davidtheclark I think that's a fair assumption, but in the case of v8,  it doesn't have an official release date so I could see this one line hanging out for a little bit. Obviously I'm speculating on how far out the release might be.

In this case, the master changelog it doesn't show anything related to this bug fix (which makes sense as it's not in any current releases). I had no idea that there was a v8 release coming out either. 

Of course it's my fault for not digging through the closed PRs/issues or checking the other branches, but I think that for the cases where you have other people who are lazy like me, this could help prevent the bug being from being filed. If I'm the only one who is too lazy to keep digging, then you're probably right that this isn't required at all.

hankthewhale:
> everybody should always do that for any repo before submitting an issue

Also a fair point but I would argue that all the things that are listed under "Here are the best ways to help resolve your issue:" are things that we should always be mindful of when contributing/opening an issue on a repo. The point of a template and little reminder notes like that are because we know that without them, popular repos will have a lot of incoming duplicate or one-liner bug reports without steps to reproduce.

I don't see the harm in adding a helpful reminder like you describe as it's a win/win for all.

jeddy3:
How's about a note about checking if the issue is resolved in an upcoming [_major milestone_](https://github.com/stylelint/stylelint/milestones)? This means we won't need to update the issue after each release, and I think it also addresses @hankthewhale concerns.

I've updated the `8.0.0` description to include a link to the `v8` branch CHANGELOG either way as that's handy to have on show.

davidtheclark:
OK, @jeddy3's idea seems good.

hankthewhale:
This all sounds good to me. I'll make the update today.

hankthewhale:
@davidtheclark @jeddy3 alright fellas. Made the requested changes!

jeddy3:
@hankthewhale Thanks very much :)

@davidtheclark Can you merge after your review please?

hankthewhale:
Changed the PR description to match what's actually happening now.

Issue #2269, opened 2017-01-18
Validating CSS nesting not working


aaronbeall:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Trying to use rules to help convert a SCSS nested codebase (`postcss-nested`) to CSS standard nesting (`postcss-nesting`)

> Which rule, if any, is this issue related to?

`selector-nested-pattern`

> What CSS is needed to reproduce this issue?

```css
.this-is-wrong {
  .this-should-raise-error {
    color: red;
    .this-is-also-wrong {
      color: darkred;
    }
  }
}
.this-is-right {
  & .this-is-okay {
    color: green;
    & this-is-also-okay {
      color: darkgreen;
    }
  }
}
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "selector-nested-pattern": "^&:\\(\\?:hover\\|focus\\)$",
    "rule-nested-empty-line-before": [ "always", {
      "except": ["first-nested"],
      "ignore": ["after-comment"]
    } ]
  }
}
```

> Which version of stylelint are you using?

e.g. `7.7.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

Webpack PostCSS plugin:

```js
  postcss: [
    require('postcss-import')({
      addDependencyTo: webpack,
      plugins: [
        require("stylelint")({ /* your options */ })
      ]
    }),
    require('postcss-cssnext'),
    require("postcss-reporter")({ clearMessages: true })
  ]
```

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

It relates to CSS spec nesting via `postcss-cssnext`.

> What did you expect to happen?

Warning or error flagged for the invalid nested rules, such as `.this-should-raise-error`, etc.

> What actually happened (e.g. what warnings or errors you are getting)?

No warnings or errors were raised.

jeddy3:
@aaronbeall Thanks for using the template.

I think your regex might need some tweaking as the following config:

```json
{
  "rules": {
    "selector-nested-pattern": "^&"
  }
}
```

With this CSS:

```css
.this-is-wrong {
  .this-should-raise-error {
    color: red;
    .this-is-also-wrong {
      color: darkred;
    }
  }
}
.this-is-right {
  & .this-is-okay {
    color: green;
    & this-is-also-okay {
      color: darkgreen;
    }
  }
}
```

Correctly produces:

```console
[...]

PR #2276, opened 2017-01-21
Use unescaped regex in string

jeddy3:
Ref: https://github.com/stylelint/stylelint/issues/2269#issuecomment-274261515


davidtheclark:
👍

PR #2281, opened 2017-01-23
Add some flow annotations, trailing comma and additional checking

ivanzusko:
ADD/some flow annotations;
ADD/additional check if `range.end` exist, because there is a case when number tries to be compared with undefined;
ADD/trailing comma, just for consistency;


jeddy3:
@ivanzusko Thank you for contributing. 

Eslint is currently failing this PR. You can run the linting and tests locally with `npm test`.

lib/assignDisabledRanges.js
```diff
@@ -2,12 +2,12 @@
 "use strict"
 const _ = require("lodash")
 
-const COMMAND_PREFIX = "stylelint-"
-const disableCommand = COMMAND_PREFIX + "disable"
-const enableCommand = COMMAND_PREFIX + "enable"
-const disableLineCommand = COMMAND_PREFIX + "disable-line"
-const disableNextLineCommand = COMMAND_PREFIX + "disable-next-line"
-const ALL_RULES = "all"
+const COMMAND_PREFIX: string = "stylelint-"
```
davidtheclark: @ivanzusko The problem here is that our Flow annotations need to use [the comment syntax](https://flowtype.org/docs/syntax.html#comment-syntax), to avoid the need to compile the code when running it.

lib/needlessDisables.js
```diff
@@ -72,18 +72,19 @@ module.exports = function (results/*: Array<stylelint$result>*/)/*: stylelint$ne
...
-  }*/
-)/*: boolean*/ {
+  }
+): boolean {
   const rule = warning.rule,
     line = warning.line
 
-  return range.start <= line && (range.end >= line || range.end === undefined) && (!range.rules || range.rules.indexOf(rule) !== -1)
+  // Need to check if range.end exist, because line number type cannot be compared to undefined
+  return range.start <= line && (range.end && range.end >= line || range.end === undefined) && (!range.rules || range.rules.indexOf(rule) !== -1)
```
davidtheclark: `range.end` could be `0`, so this additional needs to be a little more precise.

davidtheclark:
@ivanzusko Interested in finishing up this PR? If so, let's reduce it to function signature annotations and make sure they use comment annotation, then it should be good!

ivanzusko:
@davidtheclark , of course, I'm interested in finishing up! Sorry about slow responses... had a lot of main work. I'll do my best (as far as I've understood you)

davidtheclark:
@ivanzusko: Cool! No hurry, I just wanted to make sure we were on the same page.

lib/assignDisabledRanges.js
```diff
@@ -2,12 +2,12 @@
...
-const enableCommand = COMMAND_PREFIX + "enable"
-const disableLineCommand = COMMAND_PREFIX + "disable-line"
-const disableNextLineCommand = COMMAND_PREFIX + "disable-next-line"
-const ALL_RULES = "all"
+const COMMAND_PREFIX/*: string*/ = "stylelint-"
+const disableCommand/*: string*/ = COMMAND_PREFIX + "disable"
+const enableCommand/*: string*/ = COMMAND_PREFIX + "enable"
+const disableLineCommand/*: string*/ = COMMAND_PREFIX + "disable-line"
+const disableNextLineCommand/*: string*/ = COMMAND_PREFIX + "disable-next-line"
+const ALL_RULES/*: string*/ = "all"
```
davidtheclark: Let's remove these unnecessary variable annotations, then we can merge this PR.

lib/createStylelintResult.js
```diff
@@ -1,5 +1,5 @@
-/* @flow */
 "use strict"
+/* @flow */
```
davidtheclark: We've been putting the Flow pragma *above* `'use strict', right? Why move it here?

ivanzusko: @davidtheclark hm..actually I do not know why it becomes like it is now... I'm pretty sure I could not make it... e.g. `needlessDisables.js` I add `// @flow` above as it should be... anyway I'll fix that, my bad

evilebottnawi:
@ivanzusko remove `.eslintcache` from commit :+1:

ivanzusko:
@evilebottnawi done, sorry for my inattention...

PR #2283, opened 2017-01-24
Maybe update invalid rule options in system test 005?

ntwb:
Regarding the system test 005 added in #2192

The rule `value-no-vendor-prefix` was added with the incorrect option `always` rather than `true`

@m-allanson Do you remember if you explicitly added the invalid option to explicitly test the _invalid option_ syntax? 

Here's the [snapshot message](https://github.com/stylelint/stylelint/blob/master/system-tests/005/__snapshots__/005.test.js.snap#L11-L13):
```
"text": "Unexpected option value \"always\" for rule \"value-no-vendor-prefix\"",
```

Similarly, the same case for the `selector-pseudo-element-no-unknown` rule, [snapshot message](https://github.com/stylelint/stylelint/blob/master/system-tests/005/__snapshots__/005.test.js.snap#L9):
```
"text": "Unexpected option value \"always\" for rule \"selector-pseudo-element-no-unknown\"",
```

If these were in fact done on purpose please close this PR and delete the branch, if they were not I'll update the pull request with the correct options and update the snapshots 😄

### DIFF
--- a/system-tests/005/__snapshots__/005.test.js.snap
+++ b/system-tests/005/__snapshots__/005.test.js.snap
@@ -4,14 +4,7 @@ Array [
     "deprecations": Array [],
     "errored": true,
     "ignored": undefined,
-    "invalidOptionWarnings": Array [
-      Object {
-        "text": "Unexpected option value \"always\" for rule \"selector-pseudo-element-no-unknown\"",
-      },
-      Object {
-        "text": "Unexpected option value \"always\" for rule \"value-no-vendor-prefix\"",
-      },
-    ],
+    "invalidOptionWarnings": Array [],
     "source": "005/stylesheet.css",
     "warnings": Array [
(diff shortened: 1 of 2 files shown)


m-allanson:
That'll be a mistake on my side, updating to the correct options gets a 👍 from me.

ntwb:
Thanks, I've update the second option, am pretty curious to see the outcome of any _snapshot_ issues....

ntwb:
The build job results of the first commit `value-no-vendor-prefix` make sense https://travis-ci.org/stylelint/stylelint/jobs/194701716

The second commit also makes sense from looking at build job errors https://travis-ci.org/stylelint/stylelint/jobs/194757395

Will update the snapshot in a follow up commit now.

jeddy3:
Good catch and investigation guys.

Issue #2123, opened 2016-11-26
Deprecate rules


jeddy3:
- [x] `declaration-block-properties-order` ([ref](https://github.com/stylelint/stylelint/issues/2035))
- [x] `selector-no-empty` ([ref](https://github.com/stylelint/stylelint/issues/2073))
- [x] `media-feature-no-missing-punctuation` ([ref](https://github.com/stylelint/stylelint/issues/2075))
- [x] `declaration-block-no-ignored-properties` ([ref](https://github.com/stylelint/stylelint/issues/2077))
- [x] `block-no-single-line` ([ref](https://github.com/stylelint/stylelint/issues/2125)) - inc. adding FAQ entry about disallowing empty line.
- [x] `no-browser-hacks` ([ref](https://github.com/stylelint/stylelint/issues/2222)) 
- [x] `no-indistinguishable-colors` ([ref](https://github.com/stylelint/stylelint/issues/2222)) 
- [x] `no-unsupported-browser-features` ([ref](https://github.com/stylelint/stylelint/issues/2222)) 
- [x] `stylelint-disable-reason`  ([ref](https://github.com/stylelint/stylelint/issues/2227)) 
- [x] `custom-property-no-outside-root` ([ref](https://github.com/stylelint/stylelint/issues/2226))
- [x] `root-no-standard-properties` ([ref](https://github.com/stylelint/stylelint/issues/2226))
- [x] `selector-root-no-composition` ([ref](https://github.com/stylelint/stylelint/issues/2226))
- [x] `time-no-imperceptible`  ([ref](https://github.com/stylelint/stylelint/issues/2249))

m-allanson:
I can probably do this in the next few days, using the pattern set in [#801](https://github.com/stylelint/stylelint/pull/801/files). See below for a suggested set of deprecation messages for these rules.

# declaration-block-properties-order

**Documentation message**
Deprecated: use the community [`stylelint-order`](https://github.com/hudochenkov/stylelint-order) plugin pack instead. 

**Linter warning message**
declaration-block-properties-order has been deprecated and in 8.0 will be removed. Use the community stylelint-order plugin pack instead.


# selector-no-empty

**Documentation message**
Deprecated: this rule is outside the scope of stylelint's functionality. See [this comment](https://github.com/stylelint/stylelint/issues/2047#issuecomment-261382105) for details. 

**Linter warning message**
selector-no-empty has been deprecated and in 8.0 will be removed.


[...]

Issue #2249, opened 2017-01-08
Add time-min-milliseconds rule


jeddy3:
Almost finished splitting https://github.com/stylelint/stylelint/issues/2079 (woohoo!).

I believe this is the one and only time we've created a rule to limit something _and then make it applicable to **only a single use case**_.

In all other instances, I believe we've created `max`, `no`, `whitelist` and `blacklist` rules (35 of them in total) that target constructs and value types and then allowed the user to configure, where applicable, the rule for their particular use case, e.g. `color-no-hex` and `declaration-property-value-whitelist`. For example, the latter exists instead of individual rules like [BorderZero](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#borderzero).

As an aside, what is "imperceptible" is bit ambiguous e.g. for visually impaired users it could be greater than 100ms.

As such, I think this rule should become deprecated in favour of a new rule (`time-max-milliseconds: int`. In the extended description within the README we can write that a bit about the "imperceptible" use case.

evilebottnawi:
@jeddy3 :+1:

davidtheclark:
I agree that this is a better plugin.

jeddy3:
Tracking in #2079

ntwb:
Reopening this for further clarification:

@jeddy3 you state it should be deprecated in favor of a new rule, @davidtheclark you suggest it should be a plugin ?

Further, if it were to become a _new rule_ rather than a plugin the rule should be named:

> `time-min-milliseconds: int`

This is based on the rule definition: `Disallow animation and transition times under 100ms.`

As such the rule should be based on the _minimum time_ allowed rather than the _maximum_.

Also some notes referenced from the original issue: #597:

> _"...developers write 100ms animations, but human eye can see only >150ms moves."_

Further, I'd see the complementary rule `time-max-milliseconds: int` as a way to limit unnecessarily long slow times such as 10000ms for example.

jeddy3:
> @davidtheclark you suggest it should be a plugin ?

Oops. I totally miss that bit.

And, yes. It should of been `min` :)

```
time-min-milliseconds: int
```

SGTM as a rule to me.

jeddy3:
- Name: `time-min-milliseconds`
- Description: "Limit the minimum number of milliseconds for time values"
- Primary options: `int`
- Secondary options: none
- Message: `Expected no less than ${time} milliseconds`
- Section: "Time"

jeddy3:
[...]

PR #2285, opened 2017-01-24
Deprecate additional rules

m-allanson:

#2123

Deprecates the following rules:

- no-browser-hacks
- no-indistinguishable-colors
- no-unsupported-browser-features
- stylelint-disable-reason
- custom-property-no-outside-root
- root-no-standard-properties
- selector-root-no-composition
- time-no-imperceptible


lib/rules/time-no-imperceptible/README.md
```diff
@@ -1,5 +1,7 @@
 # time-no-imperceptible
 
+***Deprecated: this rule is outside the scope of stylelint's functionality. See [the release planning docs](https://stylelint.io/user-guide/release-planning/) for details.***
```
ntwb: This `time-no-imperceptible` rule is being replaced with a more appropriately and more versatile rule named `time-min-milliseconds`. See #2249

m-allanson:
@ntwb I think that's a good idea. Just looking at the latest changelog, maybe I should add ' If you use this rule, please consider creating a plugin for the community.' to some of the other rules.

lib/rules/time-no-imperceptible/README.md
hudochenkov: But technically new rule isn't here yet. What if it wouldn't be here when these changes will be released?

jeddy3: We won't be releasing the `deprecations` branch until the alternative _within stylelint_ are done. So, updating this message to reference `time-min-milliseconds` SGTM. Otherwise we'll likely forget to do it when `time-min-milliseconds` is added.

hudochenkov: Agreed. `time-min-milliseconds` is on its way https://github.com/stylelint/stylelint/pull/2289

jeddy3: Awesome. Let's update this message to:

***Deprecated: use the `time-min-milliseconds` rule with `100` as its primary option. ***

jeddy3:
Epic job with this @m-allanson btw! `8.0.0` moves ever closer :)

Issue #2286, opened 2017-01-24
Incorrect column reported by max-line-length


m-allanson:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug where `max-line-length` will report an incorrect column number.

> Which rule, if any, is this issue related to?

`max-line-length`

> What CSS is needed to reproduce this issue?

```css
a { background-image: url('some ignored url'); }
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "max-line-length": 20
  }
}
```

> Which version of stylelint are you using?

`7.7.1`

(actually `master@3183227`, but I don't think that makes a difference)

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

A warning to be flagged, with the column indicating the end of the line:

```shell
test.css
1:48    Expected line length to be no more than 20 characters (max-line-length)
```

> What actually happened (e.g. what warnings or errors you are getting)?

The following warning was flagged, listing an incorrect column number.

```shell
test.css
1:30    Expected line length to be no more than 20 characters (max-line-length)
```

---

***Cause***

The rule ignores long urls by replacing `url('some ignored url')` with `url()` before checking the line length.

If a line is still longer than the min length, the reported column number will be at the end of the line minus the `url` value:

```css
a { background-image: url(); }
```

instead of the end of the line including the url `value`:

```css
a { background-image: url('some ignored url'); }
```

Note that this bug is only triggered on lines that are still longer than `max-line-length` _after_ the `url` value has been replaced.

---

***Extra example***

This issue can also cause errors to be reported on incorrect lines. The following CSS will report two errors on line 1, instead of an error on each line.

```css
a { background-image: url('1234567890123456789012345678901234567890'); }
.foo .bar .baz .quux {}
```

```shell
 1:30  ✖  Expected line length to be no more than 20 characters   max-line-length
 1:54  ✖  Expected line length to be no more than 20 characters   max-line-length
```

ntwb:
Fantastic sleuthing @m-allanson 👍 

Can you add a test or tests for that code (or similar) you have in the _Extra example_ above, it'd be great to ensure that that issue is also sorted in the resolution of this issue

PR #2287, opened 2017-01-24
Fix max-line-length column incorrect column

m-allanson:

#2286

The test will fail with:

```shell
FAIL  lib/rules/max-line-length/__tests__/index.js
  ● max-line-length › reject › no description

    expect(received).toBe(expected)

    Expected value to be (using ===):
      48
    Received:
      30

      at stylelint.then.output (jest-setup.js:55:50)
```


davidtheclark:
I added a commit that fixes this test, so closes #2286.

It also fixes the incorrect positions in system test 005 that corresponded to this problem.

davidtheclark:
Added to Changelog:

> -   Fixed: `max-line-length` no longer reports incorrect column positions for lines with `url()` or `import` ([#2287](https://github.com/stylelint/stylelint/pull/2287)).

PR #2288, opened 2017-01-25
Add documentation on turning rules off and setting secondary options

CAYdenberg:
DOCUMENTATION CHANGE ONLY

Although secondary options are documented under each rule, I found there was no general explanation of how to set them. (If I just missed this, feel free to close).

Adds a section to "About rules" explaining that the "value" for each rule can either be the primary rule itself, `null` or a two-part array.

### DIFF
--- a/docs/user-guide/about-rules.md
+++ b/docs/user-guide/about-rules.md
@@ -116,6 +116,20 @@ a { transform: translate( 1, 1 ); }
  * The space inside these two parentheses */
 ```
 
+## About rule options
+
+A complete list of primary rule options can be found in the [example configuration.](example-config.md)
+
+To turn a rule off (when extending a configuration) you can set the value of the rule to `null`.
+
+Many rules have secondary options which provide further customization. To set secondary options, a two-member array replaces the primary option primitive. For example:
+
+```js
+"selector-type-no-unknown": [true, {
+  "ignoreTypes": ["date-input-polyfill"]
+}]
+```
+
 ## Rules work together
 
 The rules can be used together to enforce strict conventions.


hudochenkov:
Thank you for your contribution!

You did miss explanation about secondary options and `null`: https://stylelint.io/user-guide/configuration/#rules

jeddy3:
@CAYdenberg Thank you for contributing.

The [About rules](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/about-rules.md) document is used to outline general naming conventions and interplay _between the rules_. What you're referring to here is the _configuration object_. This is documented in the [Configuration](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md) section of the User guide, specifically the [`rules` bit](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#rules).

As you can see, we already introduce primary and secondary options and turning rules off with `null`. Having said that, I think your explicit examples are well worth adding to this section.

What do you think of?

---

### `rules`

Rules determine what the linter looks for and complains about. There are [over 150](rules.md) built into stylelint. *No rules are turned on by default*, so this is where you turn on everything you want to check. All the rules must be explicitly configured as *there are no default values*.

The `rules` property is *an object whose keys are rule names and values are rule configurations*. Each rule configuration fits one of the following formats:

-   a single value (the primary option)
-   an array with two values (`[primary option, secondary options]`)
-   `null` (to turn the rule off)

```json
{
  "rules": {
    "block-no-empty": null,
    "color-no-invalid-hex": true,
    "comment-empty-line-before": [ "always", {
      "ignore": ["stylelint-commands", "between-comments"]
    } ],
    "declaration-colon-space-after": "always",
    "indentation": ["tab", {
      "except": ["value"]
    }],
    "max-empty-lines": 2,
    "rule-nested-empty-line-before": [ "always", {
      "except": ["first-nested"],
      "ignore": ["after-comment"]
    } ],
    "unit-whitelist": ["em", "rem", "%", "s"]
  }
}
```

Specifying a primary option will turn a rule on. A complete list of primary rule options can be found in the [example configuration](example-config.md).

To turn a rule off (when extending a configuration) you can set the value of the rule to `null`:

```
{
  "rules": {
    "at-rule-empty-line-before": null
  }
}
```

Many rules have secondary options which provide further customization. To set secondary options, a two-member array is used. For example:

```js
"selector-type-no-unknown": [true, {
  "ignoreTypes": ["date-input-polyfill"]
}]
```

CAYdenberg:
Whoops - I totally missed that. 

I'll close this ... feel free to make use of my examples if you like, but I think it's pretty clear as it currently exists.

jeddy3:
> Whoops - I totally missed that.

@CAYdenberg No worries :)

@stylelint/core Shall we update the docs to [the above](https://github.com/stylelint/stylelint/pull/2288#issuecomment-275160374)? There's no harm in being explicit and both the `null` thing and incorrectly configured secondary options pop up _a lot_ in issues.

davidtheclark:
👍

PR #2289, opened 2017-01-25
Add "time-min-milliseconds" rule

hudochenkov:

#2249

No, it's self explanatory.

### DIFF
--- a/docs/user-guide/rules.md
+++ b/docs/user-guide/rules.md
@@ -62,6 +62,7 @@ Here are all the rules within stylelint, grouped by the [*thing*](http://apps.wo
 
 ### Time
 
+-   [`time-min-milliseconds`](../../lib/rules/time-min-milliseconds/README.md): Limit the minimum number of milliseconds for time values.
 -   [`time-no-imperceptible`](../../lib/rules/time-no-imperceptible/README.md): Disallow `animation` and `transition` less than or equal to 100ms.
 
 ### Unit
--- a/docs/user-guide/rules.md
+++ b/docs/user-guide/rules.md
@@ -62,7 +62,7 @@ Here are all the rules within stylelint, grouped by the [*thing*](http://apps.wo
 
 ### Time
 
--   [`time-min-milliseconds`](../../lib/rules/time-min-milliseconds/README.md): Limit the minimum number of milliseconds for time values.
+-   [`time-min-milliseconds`](../../lib/rules/time-min-milliseconds/README.md): Specify the minimum number of milliseconds for time values.
 -   [`time-no-imperceptible`](../../lib/rules/time-no-imperceptible/README.md): Disallow `animation` and `transition` less than or equal to 100ms.
 
 ### Unit
--- a/lib/rules/time-min-milliseconds/README.md
+++ b/lib/rules/time-min-milliseconds/README.md
@@ -1,0 +1,53 @@
+# time-min-milliseconds
+
+Limit the minimum number of milliseconds for time values.
+
+```css
+a { animation: slip-n-slide 150ms linear; }
+/**                         ↑
+ *                  This time */
+```
+
+This rule checks `transition-duration`, `transition-delay`, `animation-duration`, `animation-delay`, and those times as they manifest in the `transition` and `animation` shorthands.
+
+## Options
+
+`int`: Minimum number milliseconds for time values.
+
+For example, with `100`:
+
+The following patterns are considered warnings:
+
+```css
+a { animation: 80ms; }
+```
+
+```css
+a { transition-duration: 0.08s; }
+```
+
+```css
+a { transition: background-color 6ms linear; }
+```
+
+```css
+a { animation: horse-dance 1s linear 0.01s; }
+```
+
+The following patterns are *not* considered warnings:
+
+```css
+a { animation: 8s; }
+```
+
+```css
+a { transition-duration: 0.8s; }
+```
+
+```css
+a { transition: background-color 600ms linear; }
+```
+
+```css
+a { animation: horse-dance 1s linear 1.3s; }
+```
--- a/lib/rules/time-min-milliseconds/README.md
+++ b/lib/rules/time-min-milliseconds/README.md
@@ -1,6 +1,6 @@
 # time-min-milliseconds
 
-Limit the minimum number of milliseconds for time values.
+Specify the minimum number of milliseconds for time values.
 
 ```css
 a { animation: slip-n-slide 150ms linear; }
@@ -12,7 +12,7 @@ This rule checks `transition-duration`, `transition-delay`, `animation-duration`
 
 ## Options
 
-`int`: Minimum number milliseconds for time values.
+`int`: Minimum number of milliseconds for time values.
 
 For example, with `100`:
 
@@ -31,7 +31,7 @@ a { transition: background-color 6ms linear; }
 ```
 
 ```css
-a { animation: horse-dance 1s linear 0.01s; }
+a { animation-delay: 0.01s; }
 ```
 
 The following patterns are *not* considered warnings:
@@ -49,5 +49,5 @@ a { transition: background-color 600ms linear; }
 ```
 
 ```css
-a { animation: horse-dance 1s linear 1.3s; }
+a { animation-delay: 1s; }
 ```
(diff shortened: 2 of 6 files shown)


lib/rules/time-min-milliseconds/__tests__/index.js
```diff
@@ -0,0 +1,207 @@
...
+const rule = rules[ruleName]
+
+const MIN_VALUE = 100
+
+testRule(rule, {
+  ruleName,
+  config: [MIN_VALUE],
+
+  accept: [ {
+    code: "a { transition-delay: 0.2s; }",
```
jeddy3: Can we get a test for `0.1s` please?

lib/rules/time-min-milliseconds/__tests__/index.js
```diff
@@ -0,0 +1,207 @@
...
+  }, {
+    code: "a { -webkit-transition-delay: 0.2s; }",
+  }, {
+    code: "a { tRaNsItIoN-dElAy: 0.2s; }",
+  }, {
+    code: "a { TRANSITION-DELAY: 0.2s; }",
+  }, {
+    code: "a { transition-delay: 3s; }",
+  }, {
+    code: "a { transition-delay: 200ms; }",
```
jeddy3: Can we get a test for `100ms` please?

lib/rules/time-min-milliseconds/__tests__/index.js
hudochenkov: Great catch! It was failing for `0.1s` and `100ms` :)

docs/user-guide/rules.md
```diff
@@ -62,6 +62,7 @@ Here are all the rules within stylelint, grouped by the [*thing*](http://apps.wo
 
 ### Time
 
+-   [`time-min-milliseconds`](../../lib/rules/time-min-milliseconds/README.md): Limit the minimum number of milliseconds for time values.
```
davidtheclark: I think "Specify" is a better word here than "Limit": "Specify the minimum number ..."

lib/rules/time-min-milliseconds/README.md
```diff
@@ -0,0 +1,53 @@
...
+a { animation: slip-n-slide 150ms linear; }
+/**                         ↑
+ *                  This time */
+```
+
+This rule checks `transition-duration`, `transition-delay`, `animation-duration`, `animation-delay`, and those times as they manifest in the `transition` and `animation` shorthands.
+
+## Options
+
+`int`: Minimum number milliseconds for time values.
```
davidtheclark: number *of* milliseconds

lib/rules/time-min-milliseconds/README.md
```diff
@@ -0,0 +1,53 @@
...
+```css
+a { transition-duration: 0.08s; }
+```
+
+```css
+a { transition: background-color 6ms linear; }
+```
+
+```css
+a { animation: horse-dance 1s linear 0.01s; }
```
davidtheclark: I find this to be an ambiguous example. A reader might reasonably come away thinking both `1s` and `0.01s` are not ok. Let's just have one time value in each example, since no text specifies which particular value is the violator.

Same applies to the example below.

lib/rules/time-min-milliseconds/__tests__/index.js
```diff
@@ -0,0 +1,211 @@
...
+  }, {
+    code: "a { animation-delay: 3s; }",
+  }, {
+    code: "a { animation-delay: 200ms; }",
+  }, {
+    code: "a { animation-delay: 5000ms; }",
+  }, {
+    code: "a { animation-duration: 0.15s; }",
+  }, {
+    code: "a { -webkitanimation-duration: 0.15s; }",
```
davidtheclark: Missing hyphen, I think.

lib/rules/time-min-milliseconds/__tests__/index.js
```diff
@@ -0,0 +1,211 @@
...
+  }, {
+    code: "a { animation-duration: 4700ms; }",
+  }, {
+    code: "a { animation: foo 0.8s linear; }",
+  }, {
+    code: "a { -webkit-animation: foo 0.8s linear; }",
+  }, {
+    code: "a { animation: foo 0.8s 200ms ease-in-out; }",
+  }, {
+    code: "a { animation-delay: -2.5s; }",
```
davidtheclark: This is a little confusing and needs to be clarified in the docs. Do *any* negative values work? Or is the "minimum" actually an *absolute* value applying on both sides of zero?

lib/rules/time-min-milliseconds/index.js
```diff
@@ -0,0 +1,80 @@
...
+const ruleMessages = require("../../utils/ruleMessages")
+const validateOptions = require("../../utils/validateOptions")
+const keywordSets = require("../../reference/keywordSets")
+const postcss = require("postcss")
+const valueParser = require("postcss-value-parser")
+
+const ruleName = "time-min-milliseconds"
+
+const messages = ruleMessages(ruleName, {
+  expected: time => `Expected no less than ${time} milliseconds`,
```
davidtheclark: Let's say `Expected of a minimum of ${time} milliseconds` to avoid quarrels about "less" vs "fewer".

lib/rules/time-min-milliseconds/index.js
```diff
@@ -0,0 +1,80 @@
...
+      actual: minimum,
+      possible: _.isNumber,
+    })
+    if (!validOptions) {
+      return
+    }
+
+    root.walkDecls(decl => {
+      if (keywordSets.longhandTimeProperties.has(postcss.vendor.unprefixed(decl.prop.toLowerCase()))) {
+        if (isNotAcceptableTime(decl.value)) {
```
davidtheclark: These two conditions could be combined into one if statement.

lib/rules/time-min-milliseconds/index.js
```diff
@@ -0,0 +1,80 @@
...
+
+      if (parsedTime.unit.toLowerCase() === "s" && absoluteTime * 1000 < minimum) {
+        return true
+      }
+
+      return false
+    }
+
+    function complain(decl) {
+      const offset = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : 0
```
davidtheclark: Ugh this is awkward. Can you use an explicit argument instead?

hudochenkov:
@davidtheclark Thank you for an excellent review. To be honest, I just took an old rule and changed few things to be a new rule ¯﻿\﻿_﻿(﻿ツ﻿)﻿_﻿/﻿¯ Will cleanup things.

lib/rules/time-min-milliseconds/__tests__/index.js
hudochenkov: Good question! I think we should check only positive values.

According to specs `transition-duration` and `animation-duration` can't be negative. Negative values can be in shorthand properties `transition` and `animation`, but it's hard to parse what time is which in shorthands.

Also, negative delay is a rare case. Developer could set negative value in two cases: intentionally for some reason, or by mistake.

What do you think?

lib/rules/time-min-milliseconds/index.js
```diff
@@ -12,7 +12,7 @@ const valueParser = require("postcss-value-parser")
 const ruleName = "time-min-milliseconds"
 
 const messages = ruleMessages(ruleName, {
-  expected: time => `Expected no less than ${time} milliseconds`,
+  expected: time => `Expected of a minimum of ${time} milliseconds`,
```
davidtheclark: Cut the first "of" here: `Expected a minimum of ${time} ...`

hudochenkov: Done.

hudochenkov:
There is an important question lost in outdated diff comments. What numbers should this rule check and [my opinion about it](https://github.com/stylelint/stylelint/pull/2289#discussion_r98077424).

lib/rules/time-min-milliseconds/__tests__/index.js
davidtheclark: Yes, I agree: if `transition-duration` and `animation-duration` cannot be negative, then let's not use the absolute value here, and only check positive values.

davidtheclark:
@hudochenkov I responded above, then realized it might be lost in the diff, too:

> Yes, I agree: if transition-duration and animation-duration cannot be negative, then let's not use the absolute value here, and only check positive values.

davidtheclark:
Added to Changelog:

> -   Added: `time-min-milliseconds` rule, to replace `time-no-imperceptible` ([#2289](https://github.com/stylelint/stylelint/pull/2289)).

Issue #2270, opened 2017-01-19
Implement caching to improve performance


sergesemashko:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Feature request.

I have a project with > 100 .scss files with setup of webpack and stylelint using [stylelint-webpack-plugin](https://github.com/JaKXz/stylelint-webpack-plugin). Introducing stylelint made builds slower:
```
webpack built 17bce08fdf4ec045935d in 9317ms
```
Initially, I thought that webpack plugin is the source of slowness. But profiling of `./lib/standalone.js` showed that it just takes time to lint the entire style folder:
```
Stylelint time: 6168.771ms
```
I tried to lint only changed files while in development mode and linting time went down to ~100ms (max `170.055ms`). 

I'd like to suggest implementing sort of "good file" caching and lint only failed + changed files. We could borrow caching solution from eslint ([Implement cache in order to only operate on changed files since previous run.](https://github.com/eslint/eslint/pull/3597)). 

In short, they cache files metadata for passed files and exclude them from subsequent lint runs. Cache is persisted to `./node_modules/.eslintcache` by default.

Also, we could learn from the issues they had with cache and avoid them:
- [Do not implicitly cache `package.json`](https://github.com/eslint/eslint/pull/4610)
- [Delete cache only when executing on files](https://github.com/eslint/eslint/pull/6540)
- [Cache is basically not working](https://github.com/eslint/eslint/pull/4019/files)
- [Do not cache `.eslintrc.js`](https://github.com/eslint/eslint/pull/5068)

> What stylelint configuration is needed to reproduce this issue?

Stylelint nodejs api options:

```js
{ 
  formatter: require('stylelint').formatters.string,
  configFile: '.stylelintrc',
  files: [ './src/sass/**/*.s?(c|a)ss' ]
 }
```

.stylelintrc:
```json
{
  "extends": "stylelint-config-standard",
    "rules": {
      "at-rule-no-vendor-prefix": true,
      "color-named": "always-where-possible",
      "media-feature-name-no-vendor-prefix": true,
      "property-no-vendor-prefix": true,
      "selector-no-vendor-prefix": true,
      "selector-no-universal": true,
      "value-no-vendor-prefix": true
    }
}
```

> Which version of stylelint are you using?

`7.7.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

[...]

PR #2293, opened 2017-01-27
Implement caching

### LINKED ISSUE #3770: Add support for `x` to unit-no-unknown

`unit-no-unknown ` doesn't know `x` unit, but it is in css spec and supported some browsers.

Input:
```css
div {
  background-image: image-set(
	);
}
```

Improve check in `unit-no-unknown` (maybe need update logic in other `unit-*` rules)

### LINKED ISSUE #6780: Release 15.5.0
- [x] stylelint release
- [x] stylelint-config-recommended update/release
- [x] stylelint-config-standard update/release
- [x] stylelint.io update
- [x] stylelint-demo check
- [x] tweet

Please see the [next changelog](https://github.com/stylelint/stylelint/pull/6762/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed), which adds a few secondary options and includes a few bugfixes.

In addition, we will ship an update of `stylelint-config-recommended` together:

- https://github.com/stylelint/stylelint-config-recommended/issues/186

sergesemashko:

https://github.com/stylelint/stylelint/issues/2270

UPDATED:
Tests are ~~in progress~~ **done**.
~~I decided to submit a PR for initial feedback before going too deep into woods with tests.~~

New dependencies:
- [debug](https://www.npmjs.com/package/debug) - for easy debugging/execution time profiling. To profile lint time just prefix you stylelint command with `DEBUG=stylelint:standalone stylelint "*.scss"`
- [imurmurhash](https://www.npmjs.com/package/imurmurhash) - see [eslint requires nodejs built with ssl support #5522](https://github.com/eslint/eslint/issues/5522)
- [file-entry-cache](https://www.npmjs.com/package/file-entry-cache) - required to get and cache file metadata
- [fs-promise](https://www.npmjs.com/package/fs-promise) - a promisified `fs-extra` version to read, copy, remove files using promises in tests.

Accounted for issues ESlint had with implementing cache:

- [Fix: invalidate cache when config changes. (fixes #3770)](https://github.com/eslint/eslint/pull/3818)
- [Update: --cache-file can take a directory where to store the cache. Fixes #4241](https://github.com/eslint/eslint/pull/4255)
- [Fix: `cacheLocation` handles paths in windows style. (fixes #4285)](https://github.com/eslint/eslint/pull/4287)
- [New: do not remove non visited files from cache. (Fixes #6780)](https://github.com/eslint/eslint/pull/6921)
- [eslint requires nodejs built with ssl support #5522](https://github.com/eslint/eslint/issues/5522)

@jeddy3, @davidtheclark, please, take a look

### DIFF
--- a/docs/user-guide/node-api.md
+++ b/docs/user-guide/node-api.md
@@ -77,6 +77,22 @@ If `true`, all disable comments (e.g. `/* stylelint-disable block-no-empty */`) 
 
 You can use this option to see what your linting results would be like without those exceptions.
 
+## `cache`
+
+Store the info about processed files in order to only operate on the changed ones the next time you run stylelint. Enabling this option can dramatically improve stylelint's speed, because only changed files will be linted.
+
+By default, the cache is stored in `.stylelintcache` in `process.cwd()`. To change this, use the `cacheLocation` option.
+
+**Note:** If you run stylelint with `cache` and then run stylelint without `cache`, the `.stylelintcache` file will be deleted. This is necessary because we have to assume that `.stylelintcache` was invalidated by that second command.
+
+## `cacheLocation`
+
+A path to a file or directory to be used for `cache`. Only meaningful alongside `cache`. If no location is specified, `.stylelintcache` will be created in `process.cwd()`.
+
+If a directory is specified, a cache file will be created inside the specified folder. The name of the file will be based on the hash of `process.cwd()` (e.g. `.cache_hashOfCWD`). This allows stylelint to reuse a single location for a variety of caches from different projects.
+
+**Note:** If the directory of `cacheLocation` does not exist, make sure you add a trailing `/` on \*nix systems or `\` on Windows. Otherwise, the path will be assumed to be a file.
+
 ### `reportNeedlessDisables`
 
 If `true`, `ignoreDisables` will also be set to `true` and the returned data will contain a `needlessDisables` property, whose value is an array of objects, one for each source, with tells you which stylelint-disable comments are not blocking a lint warning.
--- a/lib/__tests__/fixtures/cache/valid.css
+++ b/lib/__tests__/fixtures/cache/valid.css
@@ -1,0 +1,1 @@
+/* This file will not cause a linting error */
--- a/lib/__tests__/standalone-cache.test.js
+++ b/lib/__tests__/standalone-cache.test.js
@@ -1,0 +1,188 @@
+"use strict"
+
+const path = require("path")
+const standalone = require("../standalone")
+const hash = require("../utils/hash")
+const fixturesPath = path.join(__dirname, "fixtures")
+const pify = require("pify")
+const fs = require("fs")
+const readFile = pify(fs.readFile)
+const unlink = pify(fs.unlink)
+const cpFile = require("cp-file")
+const fileExists = require("file-exists-promise")
+
+const cwd = process.cwd()
+const invalidFile = path.join(fixturesPath, "empty-block.css")
+const validFile = path.join(fixturesPath, "cache", "valid.css")
+const newFileDest = path.join(fixturesPath, "cache", "newFile.css")
+
+// Config object is getting mutated internally.
+// Return new object of the same structure to
+// make sure config doesn't change between runs.
+function getConfig() {
+  return {
+    files: path.join(fixturesPath, "cache", "*.css"),
+    config: {
+      rules: { "block-no-empty": true, "color-no-invalid-hex": true },
+    },
+    cache: true,
+  }
+}
+
+describe("standalone cache", () => {
+  const expectedCacheFilePath = path.join(cwd, ".stylelintcache")
+
+  beforeEach(() => {
+    // Initial run to warm up the cache
+    return standalone(getConfig())
+  })
+
+  afterEach(() => {
+    // Clean up after each test case
+    return Promise.all([
+      unlink(expectedCacheFilePath).catch(() => {
+        // fs.unlink() throws an error if file doesn't exist and it's ok. We just
+        // want to make sure it's not there for next test.
+      }),
+      unlink(newFileDest).catch(() => {
+        // fs.unlink() throws an error if file doesn't exist and it's ok. We just
+        // want to make sure it's not there for next test.
+      }),
+    ])
+  })
+
+  it("cache file is created at $CWD/.stylelintcache", () => {
+    // Ensure cache file exists
+    return fileExists(expectedCacheFilePath).then(isFileExist => {
+      expect(!!isFileExist).toBe(true)
+      return readFile(expectedCacheFilePath, "utf8")
+    }).then(fileContents => {
+      return JSON.parse(fileContents)
+    }).then(cacheFile => {
+      // Ensure cache file contains only linted css file
+      expect(typeof cacheFile[validFile] === "object").toBe(true)
+      expect(typeof cacheFile[newFileDest] === "undefined").toBe(true)
+    })
+  })
+
+  it("only changed files are linted", () => {
+    // Add "changed" file
+    return cpFile(validFile, newFileDest).then(() => {
+      // Next run should lint only changed files
+      return standalone(getConfig())
+    }).then(output => {
+      // Ensure only changed files are linted
+      const isValidFileLinted = !!output.results.find(file => file.source === validFile)
+      const isNewFileLinted = !!output.results.find(file => file.source === newFileDest)
+      expect(isValidFileLinted).toBe(false)
+      expect(isNewFileLinted).toBe(true)
+      // Ensure cache file contains linted css files
+      return readFile(expectedCacheFilePath, "utf8")
+    }).then(fileContents => {
+      return JSON.parse(fileContents)
+    }).then(cachedFiles => {
+      expect(typeof cachedFiles[validFile] === "object").toBe(true)
+      expect(typeof cachedFiles[newFileDest] === "object").toBe(true)
+    })
+  })
+
+  it("all files are linted on config change", () => {
+    const changedConfig = getConfig()
+    changedConfig.config.rules["block-no-empty"] = false
+    return cpFile(validFile, newFileDest).then(() => {
+      // All file should be re-linted as config has changed
+      return standalone(changedConfig)
+    }).then(output => {
+      // Ensure all files are re-linted
+      const isValidFileLinted = !!output.results.find(file => file.source === validFile)
+      const isNewFileLinted = !!output.results.find(file => file.source === newFileDest)
+      expect(isValidFileLinted).toBe(true)
+      expect(isNewFileLinted).toBe(true)
+    })
+  })
+
+  it("invalid files are not cached", () => {
+    return cpFile(invalidFile, newFileDest).then(() => {
+      // Should lint only changed files
+      return standalone(getConfig())
+    }).then((output) => {
+      expect(output.errored).toBe(true)
+      // Ensure only changed files are linted
+      const isValidFileLinted = !!output.results.find(file => file.source === validFile)
+      const isInvalidFileLinted = !!output.results.find(file => file.source === newFileDest)
+      expect(isValidFileLinted).toBe(false)
+      expect(isInvalidFileLinted).toBe(true)
+      // Ensure cache file doesn't contain invalid file
+      return readFile(expectedCacheFilePath, "utf8")
+    }).then(fileContents => {
+      return JSON.parse(fileContents)
+    }).then(cachedFiles => {
+      expect(typeof cachedFiles[validFile] === "object").toBe(true)
+      expect(typeof cachedFiles[newFileDest] === "undefined").toBe(true)
+    })
+  })
+  it("cache file is removed when cache is disabled", () => {
+    const noCacheConfig = getConfig()
+
+    noCacheConfig.cache = false
+    let cacheFileExists = true
+    return standalone(noCacheConfig).then(() => {
+      return fileExists(expectedCacheFilePath).then(() => {
+        throw new Error(`Cache file is supposed to be removed, ${expectedCacheFilePath} is found instead`)
+      }).catch(() => {
+        cacheFileExists = false
+        expect(cacheFileExists).toBe(false)
+      })
+    })
+  })
+})
+describe("standalone cache uses cacheLocation", () => {
+  const cacheLocationFile = path.join(fixturesPath, "cache", ".cachefile")
+  const cacheLocationDir = path.join(fixturesPath, "cache")
+  const expectedCacheFilePath = path.join(cacheLocationDir, `.stylelintcache_${hash(cwd)}`)
+  afterEach(() => {
+    // clean up after each test
+    return Promise.all([
+      unlink(expectedCacheFilePath).catch(() => {
+        // fs.unlink() throws an error if file doesn't exist and it's ok. We just
+        // want to make sure it's not there for next test.
+      }),
+      unlink(cacheLocationFile).catch(() => {
+        // fs.unlink() throws an error if file doesn't exist and it's ok. We just
+        // want to make sure it's not there for next test.
+      }),
+    ])
+  })
+  it("cacheLocation is a file", () => {
+    const config = getConfig()
+    config.cacheLocation = cacheLocationFile
+    return standalone(config).then(() => {
+      // Ensure cache file is created
+      return fileExists(cacheLocationFile)
+    }).then(fileStats => {
+      expect(!!fileStats).toBe(true)
+      // Ensure cache file contains cached entity
+      return readFile(cacheLocationFile, "utf8")
+    }).then(fileContents => {
+      return JSON.parse(fileContents)
+    }).then(cacheFile => {
+      expect(typeof cacheFile[validFile] === "object").toBe(true)
+    })
+  })
+  it("cacheLocation is a directory", () => {
+    const config = getConfig()
+    config.cacheLocation = cacheLocationDir
+    return standalone(config).then(() => {
+      return fileExists(expectedCacheFilePath)
+    }).then(cacheFileStats => {
+      // Ensure cache file is created
+      expect(!!cacheFileStats).toBe(true)
+      // Ensure cache file contains cached entity
+      return readFile(expectedCacheFilePath, "utf8")
+    }).then(fileContents => {
+      return JSON.parse(fileContents)
+    }).then(cacheFile => {
+      expect(typeof cacheFile[validFile] === "object").toBe(true)
+    })
+  })
+})
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -6,6 +6,10 @@ const createStylelint = require("./createStylelint")
 const globby/*: Function*/ = require("globby")
 const needlessDisables/*: Function*/ = require("./needlessDisables")
 const alwaysIgnoredGlobs = require("./alwaysIgnoredGlobs")
+const FileCache = require("./utils/FileCache")
+const debug = require("debug")("stylelint:standalone")
+const pkg = require("../package.json")
+const hash = require("./utils/hash")
 
 /*::type CssSyntaxErrorT = {
   column: number;
@@ -38,6 +42,11 @@ module.exports = function (options/*: stylelint$standaloneOptions */)/*: Promise
   const syntax = options.syntax
   const customSyntax = options.customSyntax
   const allowEmptyInput = options.allowEmptyInput
+  const cacheLocation = options.cacheLocation
+  const useCache = options.cache || false
+  let fileCache
+
+  const startTime = Date.now()
 
   const isValidCode = typeof code === "string"
   if (!files && !isValidCode || files && (code || isValidCode)) {
@@ -90,6 +99,17 @@ module.exports = function (options/*: stylelint$standaloneOptions */)/*: Promise
     alwaysIgnoredGlobs.map(file => "!" + file)
   )
 
+  if (useCache) {
+    const stylelintVersion = pkg.version
+    const hashOfConfig = hash(`${stylelintVersion}_${JSON.stringify(config)}`)
+    fileCache = new FileCache(cacheLocation, hashOfConfig)
+  } else {
+    // No need to calculate hash here, we just want to delete cache file.
+    fileCache = new FileCache(cacheLocation)
+    // Remove cache file if cache option is disabled
+    fileCache.destroy()
+  }
+
   return globby(fileList).then(filePaths => {
     if (!filePaths.length) {
       if (allowEmptyInput === undefined || !allowEmptyInput) {
@@ -113,14 +133,27 @@ module.exports = function (options/*: stylelint$standaloneOptions */)/*: Promise
       }
     }
 
-    const getStylelintResults = filePaths.map(filePath => {
+    let absoluteFilePaths = filePaths.map(filePath => {
       const absoluteFilepath = (!path.isAbsolute(filePath))
         ? path.join(process.cwd(), filePath)
-        : filePath
+        : path.normalize(filePath)
+      return absoluteFilepath
+    })
+
+    if (useCache) {
+      absoluteFilePaths = absoluteFilePaths.filter(fileCache.hasFileChanged.bind(fileCache))
+    }
+
+    const getStylelintResults = absoluteFilePaths.map(absoluteFilepath => {
+      debug(`Processing ${absoluteFilepath}`)
       return stylelint._lintSource({
         filePath: absoluteFilepath,
       }).then(postcssResult => {
-        return stylelint._createStylelintResult(postcssResult, filePath)
+        if (postcssResult.stylelint.stylelintError && useCache) {
+          debug(`${absoluteFilepath} contains linting errors and will not be cached.`)
+          fileCache.removeEntry(absoluteFilepath)
+        }
+        return stylelint._createStylelintResult(postcssResult, absoluteFilepath)
       }).catch(handleError)
     })
 
@@ -137,6 +170,10 @@ module.exports = function (options/*: stylelint$standaloneOptions */)/*: Promise
     if (reportNeedlessDisables) {
       returnValue.needlessDisables = needlessDisables(stylelintResults)
     }
+    if (useCache) {
+      fileCache.reconcile()
+    }
+    debug(`Linting complete in ${Date.now() - startTime}ms`)
     return returnValue
   }
 }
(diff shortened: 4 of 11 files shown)


evilebottnawi:
@sergesemashko Awesome, thanks for helping :+1:

IanVS:
Looks like you did a great job researching the lessons we learned over at ESLint.  I'm very excited to see this functionality land in stylelint.  👏

sergesemashko:
I have more free time this week to submit tests as well, I'm close

sergesemashko:
Travis and appveyor are running out of memory in the very beginning, not sure why :/

On my local all tests are passing on node 7.5.0:


@ntwb, @evilebottnawi @davidtheclark, any ideas or suggestions?

m-allanson:
@sergesemashko this is an exciting feature, thanks for working on it!

When we first started using Jest we had some issues with CI failing due running out of memory. It looks like that is what's causing the failures on this PR too.

I've tried running your branch locally and noticed that the tests take significantly longer to run when compared to `stylelint:master`. There are also random failing tests due to async callback timeouts.

Here's the tests on `stylelint:master` running on my machine:

screen shot 2017-02-07 at 22 57 57

Here's the tests on `sergesemashko:implement-caching` running on my machine:

screen shot 2017-02-07 at 23 05 53

(Note that the failing tests will be different on each run)

The `stylelint:master` tests run in around ~90 seconds, and the `implement-caching` tests around ~680 seconds. This is a big jump and I suspect it's why the CI is failing. I haven't taken a close look at the code but hopefully that's enough info someone to do some investigation.

sergesemashko:
@m-allanson, thanks for checking and pointing out to that. I'll investigate the reason. Definitely don't want to introduce any issues for testing :)

docs/user-guide/node-api.md
```diff
@@ -77,6 +77,20 @@ If `true`, all disable comments (e.g. `/* stylelint-disable block-no-empty */`)
 
 You can use this option to see what your linting results would be like without those exceptions.
 
+## `cache`
+
+Store the info about processed files in order to only operate on the changed ones. The cache is stored in `.stylelintcache` by default. Enabling this option can dramatically improve Stylelint's running time by ensuring that only changed files are linted.
+
+**Note:** If you run Stylelint with `--cache` and then run Stylelint without `--cache`, the `.stylelintcache` file will be deleted. This is necessary because the results of the lint might change and make `.stylelintcache` invalid. If you want to control when the cache file is deleted, then use `--cache-location` to specify an alternate location for the cache file.
```
davidtheclark: This is a little confusing to me: does that mean that if I use `--cache-location`, my cache file will become outdated when I run without `--cache`, because it will *not* be deleted? I'm not sure I'd want this implicit behavior change solely because I specified a cache file location.

docs/user-guide/node-api.md
```diff
@@ -77,6 +77,20 @@ If `true`, all disable comments (e.g. `/* stylelint-disable block-no-empty */`)
...
+
+## `cacheLocation`
+
+Path to the cache location. Can be a file or a directory. If no location is specified, `.stylelintcache` will be used. In that case, the file will be created in the directory where the `stylelint` command is executed.
+
+If a directory is specified, a cache file will be created inside the specified folder. The name of the file will be based on the hash of the current working directory (CWD). e.g.: `.cache_hashOfCWD`
```
davidtheclark: Why use a hash filename instead of simply adding `.stylelintcache` to that directory?

lib/__tests__/fixtures/cache/valid.css
```diff
@@ -0,0 +1 @@
+/* This will not cause a CSS syntax error */
```
davidtheclark: Do you mean this will not cause a *linting* error? Those "invalid" files also will not cause CSS syntax errors, right?

lib/__tests__/standalone-cache.test.js
```diff
@@ -0,0 +1,73 @@
...
+      files: csses,
+      config: {
+        rules: { "block-no-empty": true, "color-no-invalid-hex": true },
+      },
+      cache: true,
+    }).then(() => {
+      mockedFileCache.verify()
+    })
+  })
+  afterEach(() => {
```
davidtheclark: Can you please put this up by the `beforeEach`? I find that helps force me to grasp the full context of each test upfront.

lib/__tests__/standalone-cache.test.js
```diff
@@ -0,0 +1,73 @@
...
+  it("cache file is removed", () => {
+    return standalone({
+      files: `${fixturesPath}/empty-block.css`,
+      // Path to config file
+      configFile: path.join(__dirname, "fixtures/config-block-no-empty.json"),
+    }).then(() => {
+      expect(destroyStub.called).toBe(true)
+    })
+  })
+  afterEach(() => {
```
davidtheclark: Same point about moving this up.

lib/standalone.js
```diff
@@ -76,14 +91,28 @@ module.exports = function (options/*: Object */)/*: Promise<stylelint$standalone
...
+    }
+
+    const getStylelintResults = absoluteFilePaths.map(absoluteFilepath => {
+      debug(`Processing ${absoluteFilepath}`)
       return stylelint._lintSource({
         filePath: absoluteFilepath,
       }).then(postcssResult => {
-        return stylelint._createStylelintResult(postcssResult, filePath)
+        if (postcssResult.stylelint.stylelintError && useCache) {
+          debug(`File has problems, skip caching: ${absoluteFilepath}`)
```
davidtheclark: Let's be more clear with this message. How about `${absoluteFilepath} contains linting errors and will not be cached.`?

lib/standalone.js
```diff
@@ -100,6 +129,10 @@ module.exports = function (options/*: Object */)/*: Promise<stylelint$standalone
     if (reportNeedlessDisables) {
       returnValue.needlessDisables = needlessDisables(stylelintResults)
     }
+    if (useCache) {
+      fileCache.reconcile()
+    }
+    debug(`Linting complete in: ${Date.now() - startTime}ms`)
```
davidtheclark: No need for a colon here.

lib/utils/FileCache.js
```diff
@@ -0,0 +1,48 @@
+/* @flow */
+"use strict"
+
+const fileEntryCache = require("file-entry-cache")
+const path = require("path")
+const debug = require("debug")("stylelint:file-cache")
+const getCacheFile = require("./getCacheFile")
+const DEFAULT_CACHE_LOCATION = "./.stylelintcache"
```
davidtheclark: Let's add an empty line above this constant to differentiate it from the dependencies.

lib/utils/FileCache.js
```diff
@@ -0,0 +1,48 @@
...
+FileCache.prototype.hasFileChanged = function (absoluteFilepath) {
+  // get the descriptor for this file
+  // with the metadata and the flag that determines if
+  // the file has changed
+  const descriptor = this._fileCache.getFileDescriptor(absoluteFilepath)
+  const meta = descriptor.meta || {}
+  const changed = descriptor.changed || meta.hashOfConfig !== this._hashOfConfig
+  if (!changed) {
+    // since the file passed we store the result here
+    // successful runs as it will always should be 0 error 0 warnings
```
davidtheclark: This comment is unclear — could you please try rephrasing?

lib/utils/getCacheFile.js
```diff
@@ -0,0 +1,74 @@
...
+
+const hash = require("./hash")
+const path = require("path")
+const fs = require("fs")
+/**
+ * return the cacheFile to be used by stylelint, based on whether the provided parameter is
+ * a directory or looks like a directory (ends in `path.sep`), in which case the file
+ * name will be the `cacheFile/.cache_hashOfCWD`
+ *
+ * if cacheFile points to a file or looks like a file then in will just use that file
```
davidtheclark: I'd like to see capitalization and punctuation in all of these comments. If English is not your first language, feel free to ask for help and someone can pitch in.

lib/utils/getCacheFile.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+module.exports = function getCacheFile(cacheFile, cwd) {
+  /*
+   * make sure the path separators are normalized for the environment/os
+   * keeping the trailing path separator if present
+   */
+  cacheFile = path.normalize(cacheFile)
+
+  const resolvedCacheFile = path.resolve(cwd, cacheFile)
+  const looksLikeADirectory = cacheFile[cacheFile.length - 1 ] === path.sep
```
davidtheclark: Extra space before the `]`.

lib/utils/getCacheFile.js
```diff
@@ -0,0 +1,74 @@
...
+   * so we will try to infer if its a directory if it looks like a directory
+   * for the current operating system.
+   */
+
+  // if the last character passed is a path separator we assume is a directory
+  if (looksLikeADirectory) {
+    return getCacheFileForDirectory()
+  }
+
+  return resolvedCacheFile
```
davidtheclark: Could line 48-73 be simplified to this?

```js
if (looksLikeADirectory || (fileStats && fileStats.isDirectory()) {
  return getCacheFileForDirectory()
}

return resolvedCacheFile
```

lib/utils/getCacheFile.js
```diff
@@ -0,0 +1,74 @@
...
+const fs = require("fs")
+/**
+ * return the cacheFile to be used by stylelint, based on whether the provided parameter is
+ * a directory or looks like a directory (ends in `path.sep`), in which case the file
+ * name will be the `cacheFile/.cache_hashOfCWD`
+ *
+ * if cacheFile points to a file or looks like a file then in will just use that file
+ *
+ * @param {string} cacheFile The name of file to be used to store the cache
+ * @param {string} cwd Current working directory
```
davidtheclark: Let's note that this is only useful for testing.

lib/utils/__tests__/FileCache.test.js
```diff
@@ -0,0 +1,119 @@
...
+  mock("file-entry-cache", mockedFileEntryCache)
+  FileCache = mock.reRequire("../FileCache")
+
+  const fileCache = new FileCache(undefined, hashOfConfig)
+  // assertions
+  expect(fileCache._fileCache).toBe(fileEntryCacheInstance)
+  expect(fileCache._hashOfConfig).toBe(hashOfConfig)
+  mockedFileEntryCache.verify()
+  // restore mocked objects
+  mockedFileEntryCache.restore()
```
davidtheclark: What is the purpose of these two lines? Am I missing an assertion that they relate to?

lib/utils/__tests__/FileCache.test.js
```diff
@@ -0,0 +1,119 @@
...
+  mock.stopAll()
+})
+it("reconcile() stores hash to descriptor", () => {
```
davidtheclark: I'm not sure we need this test. Shouldn't we be able to run tests that determine whether `FileCache` actually *works* as expected without caring about the implementation details? This seems like it's just testing implementation details, not the API of the `FileCache`.

lib/utils/__tests__/FileCache.test.js
```diff
@@ -0,0 +1,119 @@
...
+  const getFileDescriptorStub = sinon.stub()
+  getFileDescriptorStub.withArgs(validFile).returns(validFileDescriptor)
+  getFileDescriptorStub.withArgs(invalidFile1).returns(invalidFileDescriptor1)
+  getFileDescriptorStub.withArgs(invalidFile2).returns(invalidFileDescriptor2)
+  const fileCache = {
+    _fileCache: {
+      getFileDescriptor: getFileDescriptorStub,
+    },
+    _hashOfConfig: hashOfConfig,
+  }
```
davidtheclark: Instead of creating this mock `fileCache` object, can't you use `stub` to `stub` the *real* `getFileDescriptor` method on a real `FileCache` instance? If so, wouldn't that would also avoid the need to awkwardly use `call` below?

lib/utils/__tests__/FileCache.test.js
```diff
@@ -0,0 +1,119 @@
...
+  const hasFileChanged = FileCache.prototype.hasFileChanged
+
+  expect(hasFileChanged.call(fileCache, validFile)).toBe(false)
+  expect(hasFileChanged.call(fileCache, invalidFile1)).toBe(true)
+  expect(hasFileChanged.call(fileCache, invalidFile2)).toBe(true)
+})
+it("file-cache-entry methods are called", () => {
```
davidtheclark: Why does this matter for our tests? Aren't these implementation details?

lib/__tests__/standalone-cache.test.js
```diff
@@ -0,0 +1,73 @@
+"use strict"
+
+const path = require("path")
+const sinon = require("sinon")
+const standalone = require("../standalone")
+const FileCache = require("../utils/FileCache")
+const fixturesPath = path.join(__dirname, "fixtures")
+
+describe("standalone cache is enabled", () => {
+  let mockedFileCache
```
davidtheclark: Why are we mocking the file cache instead of testing the real thing? Couldn't we actually *use* the file cache in our tests, cleaning up after as needed, and then bypass some of the efforts below to test its internals?

docs/user-guide/node-api.md
sergesemashko: Actually, cache file will be deleted when `--cache-location` is not changing between runs with `--cache` and without. Cache file won't be deleted for the following case:
1. `stylelint *.scss --cache` - create cache file in default location
1. `stylelint *.scss --cache-location /tmp` - won't delete cache file from default location
1. `stylelint *.scss --cache` - reuse cache file from default location

docs/user-guide/node-api.md
sergesemashko: This will allow to reuse a single location for the caches from different projects and still receive the benefits of the cache. I'll add this sentence to the doc.
see https://github.com/eslint/eslint/pull/4255#issue-113211086.

lib/utils/getCacheFile.js
sergesemashko: makes sense

lib/utils/__tests__/FileCache.test.js
sergesemashko: agree, line 23 doesn't make sense. Line 24 restores file-entry-cache dependency for further use in case of new tests.

lib/utils/__tests__/FileCache.test.js
sergesemashko: FileCache.prototype.reconcile() wraps `file-cache-entry`.reconcile() to add config hash before saving to every file entry. Just trying to make sure config hash is added on reconcile

docs/user-guide/node-api.md
davidtheclark: Ok, thanks for clarifying. I think we should remove this sentence from the docs because it represents an edge-case that is fairly confusing. If people ask about it, we could add a note somewhere else, like the FAQ.

docs/user-guide/node-api.md
davidtheclark: Sounds good!

davidtheclark:
Awesome work @sergesemashko!

lib/utils/__tests__/FileCache.test.js
sergesemashko: Just was trying to make sure the constructor is covered by tests. Maybe doesn't make sense when real FileCache is used elsewhere, I'll if it's still needed later

lib/utils/getCacheFile.js
sergesemashko: Sorry, didn't read through carefully copied text

lib/__tests__/fixtures/cache/valid.css
sergesemashko: yep, right, I'll update the wording

lib/__tests__/standalone-cache.test.js
sergesemashko: Makes sense, I'll change to the real thing

lib/utils/__tests__/FileCache.test.js
sergesemashko: ok, since I'm going to have more real things in tests, that would be the new way of testing hasFileChanged

sergesemashko:
What's in the latest update: 
- All the tests are rewritten to test on real dependencies. Actually, tests replicate real use cases now. The only thing I'm not sure about whether AppVeyor allows to do file manipulations in tests, because tests are failing there, when everything is ok on local. @evilebottnawi, @davidtheclark, do you have any insights on it or preferences how to deal with file manipulations in tests?
- Reflected comments in the code, @davidtheclark thanks a bunch!
- Reduced the overhead by adding caching and execution time looks great now.

lib/__tests__/standalone-cache.test.js
```diff
@@ -0,0 +1,146 @@
+"use strict"
+
+const path = require("path")
+const standalone = require("../standalone")
+const hash = require("../utils/hash")
+const fixturesPath = path.join(__dirname, "fixtures")
+const fsExtra = require("fs-extra")
```
davidtheclark: I'd rather we avoid this new dependency and just use regular `fs`. For copying you could use https://github.com/sindresorhus/cpy. Additionally, we should be able to avoid sync calls for all of this because Jest has good async/Promise support.

lib/__tests__/standalone-cache.test.js
```diff
@@ -0,0 +1,146 @@
...
+  })
+
+  it("invalid files are not cached", () => {
+    fsExtra.copySync(invalidFile, changedFile)
+    // Should lint only changed files
+    return standalone(getConfig()).then((output) => {
+      expect(output.errored).toBe(true)
+      // Ensure only changed files are linted
+      const isValidFileLinted = !!output.results.find(file => file.source === validFile)
+      const isInvalidFileLinted = !!output.results.find(file => file.source === changedFile)
```
davidtheclark: Should that be `invalidFile` instead of `changedFile`?

lib/__tests__/standalone-cache.test.js
```diff
@@ -0,0 +1,146 @@
...
+      expect(isValidFileLinted).toBe(false)
+      expect(isInvalidFileLinted).toBe(true)
+      // Ensure cache file doesn't contain invalid file
+      const cachedFiles = fsExtra.readJsonSync(expectedCacheFilePath)
+      expect(typeof cachedFiles[validFile] === "object").toBe(true)
+      expect(typeof cachedFiles[changedFile] === "undefined").toBe(true)
```
davidtheclark: Again, wondering if this should be `invalidFile`.

davidtheclark:
@jeddy3 I think we're close on this one. Do you want to give it a review? It's a pretty significant feature, so I want to make sure I'm not the only one who looks it over.

lib/__tests__/standalone-cache.test.js
sergesemashko: I realized that changedFile is a confusing var name here, it should be something like `newFileDest`. The idea here is to create new file so standalone should process both cached and uncached file.

lib/__tests__/standalone-cache.test.js
sergesemashko: I'll update the var as per [comment](https://github.com/stylelint/stylelint/pull/2293#discussion_r102703828)

lib/__tests__/standalone-cache.test.js
sergesemashko: thanks, wasn't aware of async/promise support

sergesemashko:
@davidtheclark, thanks for the copy edits and feedback. 
Maybe you have any ideas why file operations are failing in appveyor build?

davidtheclark:
@sergesemashko I'm afraid not. I try to ignore Appveyor. We have some Windows users in @stylelint/core. One guess: have you used `path` functions to normalize path separators?

sergesemashko:
@davidtheclark, good idea, I'll try normalized paths in tests.

sergesemashko:
@davidtheclark, Here are the latest updates:
1. Sync calls are converted to promises, however I decided to leave `fs-promise`(a promisified `fs-extra` version) because It required ways to read, remove and copy files using promises. And I though implementing the same functions using just `fs` would be redundant.
2. I updated all paths to use normalized paths, however the issue was completly gone when I added normalize to paths returned by globby()

sergesemashko:
@davidtheclark @evilebottnawi @jeddy3, could you please take a look?

evilebottnawi:
@sergesemashko can you squash commits for easy review

sergesemashko:
@evilebottnawi, sorry for the delay, It was tricky to squash all the commits in a nice way because there were merges from master. So, I decided to apply all the changes on a clean master branch within one commit.

cc/ @davidtheclark @jeddy3 

Test time looks good now, BTW:

davidtheclark:
@sergesemashko Sorry for the delay on this ... I've had an exceptionally busy few weeks. I'm hoping I can review and merge the PR this week.

Related question: Would you like to become part of the stylelint org? This is a very significant feature contribution, and we'd certainly love to have you on board if it's something you're interested in.

sergesemashko:
@davidtheclark, thank you. Certainly yes, I'd love to become a part of stylelint org.

evilebottnawi:
@sergesemashko Some question:
1. What about invalidate cache file if i change configuration?
2. Why we use multiple `cache` files? We can use one cache file and meta information about cache as do `eslint`.
Example
```shell
{"/absolute/path/to/index.js":{"size":81,"mtime":1475500756217,"hashOfConfig":"114krsx","results":{"filePath":"/absolute/path/to/index.js","messages":[],"errorCount":0,"warningCount":0}},
```
3. We can use `pify` instead `fs-promise` https://github.com/sindresorhus/pify. More universal package for promising callbacks.
4. I think we should remove all debug function from this commit and do this in separate PR.

sergesemashko:
@evilebottnawi, thanks for your feedback.

> - What about invalidate cache file if i change configuration?

it's already accounted. Every cached entry has the hash of the config. If config changes all the files will be re-linted because hashes don't match. Checkout [this line](https://github.com/stylelint/stylelint/pull/2293/files#diff-9de7ea0960deccf5a8e2c5eaadc6c91aR24)
> - Why we use multiple cache files? 556168c#diff-25bbf797b85423b576b2d765600b119aR33

This will allow to reuse a single location (let's say `/tmp`) for the caches from different projects and still receive the benefits of the cache. I'll add this sentence to the doc.
see eslint/eslint#4255 (comment).
> - We can use pify instead fs-promise https://github.com/sindresorhus/pify. More universal package for promising callbacks.

Thanks, wasn't aware of this package. Looks like it may be helpful for any other cases and it looks much lighter that `fs-promise`. Will update soon.
> - I think we should remove all debug function from this commit and do this in separate PR. 

Didn't quite get your concerns. Do you want to add more debug messages in separate PR? I'd like keep it here, if possible, as profiling time rely on using `debug` and `debug()` usages look pretty straightforward. Separate PR would just mean more work for me and for the reviewers.

evilebottnawi:
@sergesemashko Thanks for answer.

> Didn't quite get your concerns. Do you see potential issues with introducing debug and want keep discussion in separate PR?
TBH, I'd like keep it here, if possible, as cache profiling rely on using debug and its usages look pretty straightforward.
Separate PR would just mean more work for me and for the reviewers. If you know something that might be affected by adding debug within this PR, please let me know.

No i don't see problems, we should add `debug` for all code or nothing. If this is already done, it will be fine :smile: 

SGTM for me :1st_place_medal:

sergesemashko:
@evilebottnawi , I see. Frankly speaking, adding debug for all codebase sounds like a "big deal" for me 😃 I'm not yet that familiar with all the cases where it really should be added.
Maybe, gradually introducing debug to the file as we go would work? 
Debug namespaces are already unique per file, e.g. `stylelint:standalone`, `stylelint:file-cache`, so it can be easily introduced to other files when necessary. 
I know it sounds slippery, but I think it's worth suggesting :) If it still doesn't work, I'll extract `debug` and will keep the discussion in separate PR. 
cc/ @davidtheclark

davidtheclark:
@jeddy3 are you good with merging this?

jeddy3:
Added to changelog:

> -   Added: `cache` option to store the info about processed files in order to only operate on the changed ones the next time you run stylelint ([#2293](https://github.com/stylelint/stylelint/pull/2293)).

evilebottnawi:
@sergesemashko I experienced the work of the cache and it's great! Only I noticed one strangeness:
we ignore by default `node_modules` and `bower_components`, but if package have .scss (or .css) and i update package, this files fall as if not cached. I think problems in here https://github.com/stylelint/stylelint/blob/master/lib/standalone.js#L144 you don't exclude files from `node_modules` and `bower_components`.

My npm script: `stylelint '{**/*,*}.scss' --ignore-path .gitignore`. I don't know it is bug or not, just information.

sergesemashko:
@evilebottnawi, thanks for the input. 
Cache info for specific file is used depending on the output of `file-entry-cache` that checks if file has changed. And cache is invalidated for all changed files. 
Could you please clarify what's your expectation from linting dependency styles?

evilebottnawi:
@sergesemashko default command `stylelint '{**/*,*}.scss' --ignore-path .gitignore` globbing all files also inside in node_modules :(

olegskl:
I was wondering why is `fileCache` [initialized on line 105](https://github.com/stylelint/stylelint/blob/43b18fe4c56e9f177644b50354fb8efe9fb34a84/lib/standalone.js#L105), but `prepareReturnValue` [is called on line 90](https://github.com/stylelint/stylelint/blob/43b18fe4c56e9f177644b50354fb8efe9fb34a84/lib/standalone.js#L90) inside the return statement? When there are no files (linting code as a string) this throws:

> Cannot read property 'reconcile' of undefined

evilebottnawi:
@olegskl can you create issue?

sergesemashko:
@olegskl, thanks for reporting this. fileCache is initialized after 
```js
if (!files) {
    const absoluteCodeFilename = (codeFilename !== undefined && !path.isAbsolute(codeFilename))
      ? path.join(process.cwd(), codeFilename)
      : codeFilename
    return stylelint._lintSource({
      code,
      codeFilename: absoluteCodeFilename,
    }).then(postcssResult => {
      return stylelint._createStylelintResult(postcssResult)
    }).catch(handleError).then(stylelintResult => {
      return prepareReturnValue([stylelintResult])
    })
  }
```
because this block is executed when raw styles are passed. So, there is no need to initialize `fileCache` before this block, because there is no files to cache. 

However, looks like you found a bug, which wasn't covered by tests. @olegskl, could you please submit an issue? And I'll take a look at it.

olegskl:
@sergesemashko here it is #2492, sorry for the late reply.

PR #2300, opened 2017-01-29
Standardise deprecation notes and warnings

jeddy3:


lib/rules/custom-property-no-outside-root/README.md
```diff
@@ -1,6 +1,6 @@
 # custom-property-no-outside-root
 
-***Deprecated: this rule is outside the scope of stylelint's functionality. See [the release planning docs](https://stylelint.io/user-guide/release-planning/) for details.***
+***Deprecated: See [CHANGELOG](../../../CHANGELOG.md).***
```
m-allanson: Is it worth pre-emptively linking to the `v7.8.0` part of the changelog here?

` (../../../CHANGELOG.md#780)`

hudochenkov: I don't think so, because this deprecation notice will be here for the new release only and will be removed in `8.0.0`.

PR #2302, opened 2017-01-29
Revert deprecation of stylelint-commands option

jeddy3:
I see now that it’s a needless deprecation. It also meant we’d have to put out new major version of `stylelint-config-standard`. Reverting the change means `stylelint-config-standard` is `7.8.0`-ready, as this was the only change that affected it.


hudochenkov:
Deprecation isn't removal. All deprecated rules still available for all configs made for stylelint `7.7.1`. I think renaming `stylelint-commands` to `stylelint-command` was made for consistency. Every keyword is in a singular form: `after-comment`, `blockless-after-blockless`, etc. I think one custom config (`stylelint-config-standard`) shouldn't be more important than tool (`stylelint`) itself.

jeddy3:
> I think renaming stylelint-commands to stylelint-command was made for consistency

I know. I proposed it :) 

> Every keyword is in a singular form: after-comment, blockless-after-blockless, etc

I now realise that's not true. They are consistently in the singular form when referring to a previous node, but vary otherwise ([1](https://github.com/stylelint/stylelint/blob/master/lib/rules/declaration-block-no-duplicate-properties/README.md#optional-secondary-options), [2](https://github.com/stylelint/stylelint/blob/master/lib/rules/max-line-length/README.md#optional-secondary-options), [3](https://github.com/stylelint/stylelint/blob/master/lib/rules/max-nesting-depth/README.md#ignore-blockless-at-rules)). So, I now think changing from `stylelint-commands` to `stylelint-command` adds nothing.

Issue #2273, opened 2017-01-20
Improve documentation of "always-multi-line" in at-rule-name-newline-after


davegomez:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

I think there's a bug the rule, it shows an error even when the snippet complies with it. It doesn't happen with all the `@mixin` at-rules but only with these specially long and deeply nested rules that you can find in the [Flexbox Grid](http://flexboxgrid.com/) SCSS code.

> Which rule, if any, is this issue related to?

`at-rule-name-newline-after`

> What CSS is needed to reproduce this issue?

```sass
@mixin columns(
  $selector: $column-selector,
  $properties: $column-properties,
  $columns: $column-count,
  $queries: $queries
) {
  @each $size in $sizes {
    @include queries($size, $queries) {
      @for $i from 0 through $columns {
        @if ($i == 0) {
          #{$selector}#{$block-delimiter}#{$size} {
            @content;

            @include autoWidth();
          }
        }

        @else {
          #{$selector}#{$block-delimiter}#{$size}#{$modifier-delimiter}#{$i} {
            @content;

            @each $prop in $properties {
              @include percentWidth($prop, $columns, $i);
            }
          }
        }
      }
    }
  }
}
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "at-rule-name-newline-after": "always-multi-line"
  }
}
```

> Which version of stylelint are you using?

`7.7.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylelint scss/**/*.scss --syntax scss` and IDE/Editor plugins

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes, it's related to SCSS syntax.

> What did you expect to happen?

No errors to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

The following errors were flagged:

```shell
scss/_fg-grid.scss
 118:6  ✖  Expected newline after at-rule name "@mixin"   at-rule-name-newline-after
```

hudochenkov:
The rule works like expected.

```scss
@mixin columns(
  $selector: $column-selector,
  $properties: $column-properties,
  $columns: $column-count,
  $queries: $queries
) {}
```

After at-rule name `mixin` and before at-rule's block `{}` there are multiple lines. You can check [examples in docs](http://stylelint.io/user-guide/rules/at-rule-name-newline-after/#always-multi-line).

[...]

PR #2304, opened 2017-01-30
Improve documentation of "always-multi-line" in at-rule-name-newline-after

hudochenkov:

#2273


jeddy3:
Thanks!

PR #2308, opened 2017-01-31
Add fr to known units

### LINKED ISSUE #2307: Add support for `fr` unit from CSS Grid spec

The `unit-no-unknown` rule doesn't know about the `fr` unit introduced in the [CSS Grid spec](https://www.w3.org/TR/css-grid-1/)

`unit-no-unknown`

```css
.foo {
  display: grid;
  grid-template-columns: repeat(12, 1fr);
}
```

```json
{
  "rules": {
    "unit-no-unknown": true
  }
}
```

`7.7.1`

CLI with `stylelint packages/**/*.scss`

No, it's related to standard CSS, albeit a relatively new feature which is only starting to ship in browsers.

No warnings to be flagged.

The following warning was flagged:

```shell
packages/mdc-layout-grid/_mixins.scss
 88:91  ✖  Unexpected unknown unit "fr"   unit-no-unknown
```

jeddy3:
Closes #2307

### DIFF
--- a/lib/reference/keywordSets.js
+++ b/lib/reference/keywordSets.js
@@ -44,6 +44,8 @@ keywordSets.lengthUnits = new Set([
   "pt",
   "pc",
   "q",
+  // Flexible length units
+  "fr",
 ])
 
 keywordSets.units = uniteSets(keywordSets.nonLengthUnits, keywordSets.lengthUnits)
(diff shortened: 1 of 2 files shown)


davidtheclark:
Added to Changelog:

> -   Fixed: `unit-no-unknown` accepts `fr` units ([#2308](https://github.com/stylelint/stylelint/pull/2308)).

PR #2309, opened 2017-01-31
Add rule-empty-line-before rule

### LINKED ISSUE #2229: Merge rule-non-nested-empty-line-before and rule-nested-empty-line-before

A bug

rule-nested-empty-line-before with except "after-rule"

```css
a {
  b {
    color:  red;
  }
}
```

```json
{
  "rules": {
    "rule-nested-empty-line-before": [
      "always",
      {
        "except": [
          "after-rule"
        ]
      }
    ],
  }
}
```

7.7.1

VSCode plugin

Yes, SCSS nesting

No errors to be flagged.

```shell
test.css
2:2    Expected empty line before nested rule (rule-nested-empty-line-before)
```

jeddy3:
Closes #2229 

I’ll update the rule to use the new empty line utils (to ignore shared-line comments) in `v8`.


docs/user-guide/rules.md
```diff
@@ -188,8 +188,7 @@ Here are all the rules within stylelint, grouped by the [*thing*](http://apps.wo
 
 ### Rule
 
--   [`rule-nested-empty-line-before`](../../lib/rules/rule-nested-empty-line-before/README.md): Require or disallow an empty line before nested rules.
--   [`rule-non-nested-empty-line-before`](../../lib/rules/rule-non-nested-empty-line-before/README.md): Require or disallow an empty line before non-nested rules.
```
hudochenkov: Do you think it's a good idea delete links to docs? These rules still in stylelint, despite the fact they are deprecated. When user will see a deprecation message, she may want to read what this rule for. But it will be very difficult to find.

lib/rules/rule-empty-line-before/README.md
```diff
@@ -0,0 +1,223 @@
...
+{}
+b
+{}
+```
+
+## Optional secondary options
+
+### `except: ["after-single-line-comment", "inside-block-and-after-rule", "first-nested"]`
+
+#### `"after-single-line-comment"`
```
hudochenkov: Why it's explicitly for a single-line comment?

docs/user-guide/rules.md
jeddy3: Good point. 

Can you create a PR that adds them back in for the other 13 deprecated rules? Please append "(deprecated)." to the end of the description.

lib/rules/rule-empty-line-before/README.md
jeddy3: This rule mimics the options from the previous two rules. Only the `rule-nested...` had the after single line comment exception, as there was a specific request for that functionality.

jeddy3:
> With new rule it won't be possible to enforce this style:

@hudochenkov Thanks for the feedback.

> How about new option `except: ["inside-block-and-after-non-rule"]`?

We don't have the concept of `non-`things as the recommend approach is to set the primary option to the _most common_ occurrence and then invert it using `except` options. So, instead of `"never"` and `except: ["after-non-rule"]`, the recommendation is to use `"always"` and `except: ["after-rule"]`.  

I've added `except: ["after-rule"]`. 

I think that covers off a few use-cases. I suggest to roll with what we have and then wait for requests.

lib/rules/rule-empty-line-before/__tests__/index.js
```diff
@@ -98,6 +98,32 @@ testRule(rule, {
...
+
+  accept: [ {
+    code: "a {} \n b {}",
+  }, {
+    code: "$var: pink;\n\nb {}",
+    description: "scss variable",
+  }, {
+    code: "@media {}\n\na{}",
+    description: "media rule",
+  } ],
```
hudochenkov: Could you add tests for nested rules either, please? New option affects both non-nested and nested rules.

hudochenkov:
`except: ["after-rule"]` makes sense. Thank you!

lib/rules/rule-empty-line-before/__tests__/index.js
jeddy3: Done

docs/user-guide/rules.md
hudochenkov: Will do today.

lib/rules/rule-empty-line-before/README.md
```diff
@@ -0,0 +1,244 @@
...
+There *must always* be an empty line before multi-line rules.
+
+The following patterns are considered warnings:
+
+```css
+a
+{}
+b
+{}
+```
```
davidtheclark: I think these are confusing examples. Multi-line rules typically manifest as multiple lines of declarations — that's what people will recognize. Let's do that.

jeddy3: Done

jeddy3:
@davidtheclark I'm around this evening and can finish the release (woohoo!), if you're about to review https://github.com/stylelint/stylelint-config-standard/pull/60 and feedback on the example changes here.

FYI:

- stylelint-config-standard
- stylelint.io
- stylelint-demo

Are all ready-to-go against this branch.

lib/rules/rule-empty-line-before/README.md
hudochenkov: @jeddy3 As a user I would have better understanding if there would be declarations.

jeddy3: Thing is, "multi-line" doesn't necessarily mean multiple declarations. 

This is multi-line:

```
a {
  color: red;
}
```

i.e. the _rule_ spans multiple lines.

Perhaps a single declaration is the clearest?

hudochenkov: From a user perspective your example with a single declaration isn't always a multi-line rule :) User might assume that this is single-line rule.

In fact you're right, of course.

davidtheclark: @jeddy3 You're right: that seems fine to me, too.

jeddy3:
Changlog:

> -   Deprecated: 15 rules ([#2197](https://github.com/stylelint/stylelint/pull/2197),  [#2285](https://github.com/stylelint/stylelint/pull/2285) & [#2309](https://github.com/stylelint/stylelint/pull/2309)).
        -   `rule-nested-empty-line-before` and `rule-non-nested-empty-line-before`. Use the new `rule-empty-line-before` rule instead.

PR #2311, opened 2017-01-31
Use HTTPS links where applicable.

ntwb:
Maybe also a future a _beginner-friendly_ PR could be accepted that updates our code examples to prefer when possible the use of `https://` in our code examples.


lib/rules/no-invalid-double-slash-comments/README.md
```diff
@@ -1,6 +1,6 @@
 # no-invalid-double-slash-comments
 
-Disallow double-slash comments (`//...`) which are not supported by CSS and [could lead to unexpected results](http://stackoverflow.com/a/20192639/130652).
+Disallow double-slash comments (`//...`) which are not supported by CSS and [could lead to unexpected results](https://stackoverflow.com/a/20192639/130652).
```
hudochenkov: Funny thing: it redirects to HTTP. Shall we use the final URL instead? https://stackoverflow.com/questions/12298890/is-it-bad-practice-to-comment-out-single-lines-of-css-with/20192639#20192639

ntwb: Ha, it does to, I use the _HTTPS Everywhere_ Chrome extension from EFF.

And for me, it remains on the HTTPS site, so either or I guess

I've a preference to make as many as possible links used HTTPS

jeddy3:
LGTM

Issue #2086, opened 2016-11-20
Add ignoreFontFamilyName: [] option to font-family-no-duplicate-names


ntwb:
Follow up to #1284 

There should be a way to ignore, or optionally ignore duplicate `monospace` font names in `font-family-no-duplicate-names`

Ways to approach this:

1. Ignore `monospace, monospace` by default
2. Add an option to ignore only the `monospace` duplicates, i.e. `monospace, monospace`
3. Add an option to add an array of font names using regex/string to be ignored.

----

For context the above is based on the following comments in #1284:

Over in #1284 @ben-eb made the [following comment](https://github.com/stylelint/stylelint/issues/1284#issuecomment-243898839):
> _"Just FYI, this rule should not affect `monospace, monospace`, as specifying this twice corrects rendering issues across browsers. http://code.stephenmorley.org/html-and-css/fixing-browsers-broken-monospace-font-handling/"_

I added a note [below that](https://github.com/stylelint/stylelint/issues/1284#issuecomment-243936785):
> _"One of the many reasons I dislike SEO and their practices, this said post has no date, no comments and makes it impossible to determine if this was posted 10 years ago or yesterday. Hence I place next to zero value on content of this manner."_

I've still no idea if the above is a legitimate concern for _modern browsers_ but though a rudimentary search of GitHub CSS for `monospace, monospace` returned [1 million plus results](https://github.com/search?p=8&q=%22monospace%2C+monospace%3B%22&ref=searchresults&type=Code&utf8=%E2%9C%93). A high probability that this issue would come up in the future based on those results. The article cites Google Chrome & Safari together, Chrome split from Webkit in Chrome 28 in mid 2013. As such mid 2013 browser versions would be my best guess at trying to guess the versions of browsers that this issue refers to, what has changed in regard to this issue for each browser since then I'm also unaware of /shrug

Thoughts @stylelint/core ?

ntwb:
Some further background on this I just noticed that this was included in `normalize.css`:

• https://github.com/necolas/normalize.css/blob/master/README.md#pre-code-kbd-samp:

> `pre`, `code`, `kbd`, `samp`
> 
> The `font-family: monospace, monospace` hack fixes the inheritance and scaling of font-size for preformatted text. The duplication of monospace is intentional. [Source](https://en.wikipedia.org/wiki/User:Davidgothberg/Test59).

From the source cited above: https://en.wikipedia.org/wiki/User:Davidgothberg/Test59:
[...]

PR #2314, opened 2017-02-01
Add ignoreFontFamilyName option to font-family-no-duplicate

CAYdenberg:

Issue #2086 

I think it's pretty self-explanatory but this is my first PR to a major project so let me know if anything needs changing.


lib/rules/font-family-no-duplicate-names/README.md
```diff
@@ -45,3 +45,21 @@ a { font: 1em "Arial", "sans-serif", sans-serif; }
 ```css
 a { font: normal 14px/32px -apple-system, BlinkMacSystemFont, sans-serif; }
 ```
+
+
+### `ignoreFontFamilyNames: ["monospace"]`
```
m-allanson: Can you change this to show the types of values that can be accepted here?
```markdown
### `ignoreFontFamilyNames: ["/regex/", "string"]`
```

lib/rules/font-family-no-duplicate-names/README.md
```diff
@@ -45,3 +45,21 @@ a { font: 1em "Arial", "sans-serif", sans-serif; }
...
+
+Given:
+
+```js
+["monospace"]
```
m-allanson: I think it'd be good to include a regex example here too.

CAYdenberg:
Good suggestions. You got me all confused about case-sensitivity so I added another test as well.

lib/rules/font-family-no-duplicate-names/README.md
CAYdenberg: Addressed in ee5279e

lib/rules/font-family-no-duplicate-names/README.md
CAYdenberg: Addressed in ee5279e

m-allanson:
@CAYdenberg 👍 for extra tests. 

I should have explained myself more clearly for the docs changes in the previous comments. 

The [`ignoreFontFamilyNames` title](https://github.com/stylelint/stylelint/pull/2314/files?diff=split#diff-5c425a57dfd87ced29b7b6a300031f4aR51) should show the _types_ of values it accepts - regexes or strings e.g. `["/regex/", "string"]`

Then the [example usage](https://github.com/stylelint/stylelint/pull/2314/files?diff=split#diff-5c425a57dfd87ced29b7b6a300031f4aR56) should show an _example_ regex and an _example_ string. e.g. `["/^my-/", "custom"]`

You can use the [ignoreAtRules option of at-rule-no-unknown](https://github.com/stylelint/stylelint/blob/master/lib/rules/at-rule-no-unknown/README.md#ignoreatrules-regex-string) as a reference.

Can you update those parts of the readme along with the regex code example [on L66](https://github.com/stylelint/stylelint/pull/2314/files?diff=split#diff-5c425a57dfd87ced29b7b6a300031f4aR66)? This might all seem overly fussy but there are a lot of rules in stylelint so we aim for strong consistency across the docs. 

Other than the docs update this LGTM, thanks again for working on it.

CAYdenberg:
+1 for fussiness - I don't think someone who isn't fussy would be using/contributing to/writing a CSS linter!

Is the issue that the example `("/Regex/")` doesn't look enough like a regular expression? I mean, it _could_ be, but I guess I see your point. Most font family names are capitalised and have spaces (not hyphens) so I guess my thought was there would be some value in providing an example that looks like something someone might actually use. Would ("/^Regex/") be more appropriate? Or would you prefer `("/^my-/") since it's closer to the other examples.

I do see what you mean about the title - I just outright missed that.

m-allanson:
>Is the issue that the example ("/Regex/") doesn't look enough like a regular expression?

Exactly - I think it's confusing if the example regex includes the string "Regex" in it. I see your point about font family names though. We could use a more realistic example regex - `"/^My Font /"` or `"/^CustomFamily/"`?

So we'd have `["/regex/", "string"]` in the title, and something like `["/^My Font /", "custom"]` in the example.

CAYdenberg:
Got it. That should do it? I left `monospace` as the example string since that will probably cover most use cases for this option. Other than that it's exactly as you prescribed.

m-allanson:
That LGTM, thanks again!

davidtheclark:
Looks great. Thanks @CAYdenberg! I notice you're doing a lot to help out!

davidtheclark:
Added to changelog:

> -   Added: `ignoreFontFamilyName` option to `font-family-no-duplicate` ([#2314](https://github.com/stylelint/stylelint/pull/2314)).

webdesignberlin:
Any Example there, how i can use the ignoreFontFamilyName Rule within the stylelintrc?

I have try this:

``
"font-family-no-duplicate-names": [
     true, { "ignore-font-family-name": ["monospace"] }
],
``

But i became an error:
``
Invalid option name "ignore-font-family-name" for rule "font-family-no-duplicate-names"
``

Version:
`
$ ./node_modules/.bin/stylelint -v
8.0.0
`

evilebottnawi:
@webdesignberlin use `ignoreFontFamilyName` instead `ignore-font-family-name`

webdesignberlin:
@evilebottnawi 

thx, but ai tried this also. same error:
`
Invalid Option: Invalid option name "ignoreFontFamilyName" for rule "font-family-no-duplicate-names"
`

webdesignberlin:
Oje, he, she, it the s goes with :) ignoreFontFamilyNames works.

via Source Code: 
https://github.com/stylelint/stylelint/blob/74d9d05f25e8b865bc21801ec5759c4229592f71/lib/rules/font-family-no-duplicate-names/index.js#L46

jeddy3:
For future visitors... the `ignoreFontFamilyNames` option is documented [here](https://github.com/stylelint/stylelint/blob/master/lib/rules/font-family-no-duplicate-names/README.md#ignorefontfamilynames-regex-string).

PR #2315, opened 2017-02-02
Fix "too many files open" error

evilebottnawi:

PoC of parallelism. Why?

1. Speedup linting on multiprocessor system
2. No error: "Too many files..." and etc, each file open, parse and create result in a separate process.

No, it's self explanatory.

### DIFF
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -5,6 +5,8 @@ const formatters = require("./formatters")
 const createStylelint = require("./createStylelint")
 const globby = require("globby")
 const needlessDisables = require("./needlessDisables")
+const createThrottle = require("async-throttle")
+const os = require("os")
 
 module.exports = function (options/*: Object */)/*: Promise<stylelint$standaloneReturnValue>*/ {
   const files = options.files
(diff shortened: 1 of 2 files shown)


davidtheclark:
@evilebottnawi This is a simple change — does it actually speed things up? Can you provide some numbers?

evilebottnawi:
@davidtheclark try to create over 65k files and run linter, i am get too many files opens, about speed i am try about ~2-4s on my code base on 8 core processor, i can explore about speed up as soon as the free time

davidtheclark:
Oooooh ... If I understand correctly, async-throttle does not provide actual "parallelism": it throttles the number of Promises that will execute at once. So the goal here is not to speed things up, but to stop the "too many files open" error? Yeah, I think this is probably a good idea. However, since it's not parallelism, I don't think the number of cores should be used to determine how many files to lint at once. It would probably be helpful if you wrote up an issue explaining the problem you are trying to solve.

davidtheclark:
@evilebottnawi Any interest in following through on this, or should we close it — or, probably better, write up an issue so somebody else can try contributing a PR?

evilebottnawi:
@davidtheclark t think be good add tests to demonstrate problems for this and we can merge after this, maybe i found time for this on this weekend.

hudochenkov:
@evilebottnawi what's the status of this PR?

evilebottnawi:
@hudochenkov we can close this issue, but problem exists, try to create 1m files and linting their

hudochenkov:
Clothing this one, because no reports were received from users for this problem.

PR #2319, opened 2017-02-04
Add flow comments annotation to function signatures

ivanzusko:


lib/assignDisabledRanges.js
```diff
@@ -2,12 +2,12 @@
...
-const enableCommand = COMMAND_PREFIX + "enable"
-const disableLineCommand = COMMAND_PREFIX + "disable-line"
-const disableNextLineCommand = COMMAND_PREFIX + "disable-next-line"
-const ALL_RULES = "all"
+const COMMAND_PREFIX/*: string*/ = "stylelint-"
+const disableCommand/*: string*/ = COMMAND_PREFIX + "disable"
+const enableCommand/*: string*/ = COMMAND_PREFIX + "enable"
+const disableLineCommand/*: string*/ = COMMAND_PREFIX + "disable-line"
+const disableNextLineCommand/*: string*/ = COMMAND_PREFIX + "disable-next-line"
+const ALL_RULES/*: string*/ = "all"
```
davidtheclark: There is no need to annotate constants, since Flow can understand them just fine.

lib/assignDisabledRanges.js
```diff
@@ -20,7 +20,7 @@ const ALL_RULES = "all"
 module.exports = function (
   root/*: Object*/,
   result/*: Object*/
-) {
+)/*: Object*/ {
```
davidtheclark: Does this always return an object? If so, what object? Is there a more precise annotation we can give?

davidtheclark:
@ivanzusko As discussed in that other PR, let's remove all of the annotations on variables and just stick with those on the function signatures. Can you add any more specificity than `Object` to those annotations?

ivanzusko:
@davidtheclark , hope I made a correct annotation to returning object...

lib/assignDisabledRanges.js
```diff
@@ -16,11 +16,20 @@ const ALL_RULES = "all"
...
+    quiet?: boolean,
+  }
+}*/
+
 // Run it like a plugin ...
 module.exports = function (
   root/*: Object*/,
   result/*: Object*/
-) {
+)/*: returningResult*/ {
```
davidtheclark: The return value here is a `postcss$result`, right?

lib/assignDisabledRanges.js
```diff
@@ -121,7 +130,7 @@ module.exports = function (
   }
 
   function checkComment(comment/*: postcss$comment*/) {
-    const text = comment.text
+    const text/*: string*/ = comment.text
```
davidtheclark: No need for this.

lib/assignDisabledRanges.js
```diff
@@ -144,15 +153,15 @@ module.exports = function (
...
+    const rules/*: Array<string>*/ = _.compact(fullText.slice(command.length).split(",")).map(r => r.trim())
     if (_.isEmpty(rules)) {
       return [ALL_RULES]
     }
     return rules
   }
 
   function startDisabledRange(line/*: number*/, ruleName/*: string*/) {
-    const rangeObj = { start: line }
+    const rangeObj/*: Object*/ = { start: line }
```
davidtheclark: No need to annotate this.

lib/assignDisabledRanges.js
```diff
@@ -144,15 +153,15 @@ module.exports = function (
     command/*: string*/,
     fullText/*: string*/
   )/*: Array<string>*/ {
-    const rules = _.compact(fullText.slice(command.length).split(",")).map(r => r.trim())
+    const rules/*: Array<string>*/ = _.compact(fullText.slice(command.length).split(",")).map(r => r.trim())
```
davidtheclark: No need to annotate this.

lib/assignDisabledRanges.js
ivanzusko: @davidtheclark do you mean I have to rename this type?
There are just several points:

- I did not find `postcss$result` in declaration files, that's why I assumed that you've meant renaming
- I've run tests and found out that returning object is much bigger than I described. Should I expand the type annotation of the returning object?

lib/assignDisabledRanges.js
ivanzusko: fixed

lib/assignDisabledRanges.js
ivanzusko: fixed

lib/assignDisabledRanges.js
ivanzusko: fixed

lib/assignDisabledRanges.js
ivanzusko: @davidtheclark , I got it!

davidtheclark:
Thanks @ivanzusko!

ivanzusko:
@davidtheclark thank you :)

PR #2321, opened 2017-02-04
Add flow annotation

ivanzusko:


ivanzusko:
@davidtheclark I'm definitely agree with You about the Flow does a fantastic job inferring types within functions. In my practice, I'm using annotation within the functions just for easyreading the code and less assumption by me 'from the future' and for anybody who will read my code. But you are definitely right that it is not 100% necessary. My bad, sorry

davidtheclark:
@ivanzusko: I appreciate the idea of adding the annotations like comments, to clarify the code, but I think I'd like to restrict the annotations to function signatures and necessary places for now. Thanks for the suggestion!

Issue #2251, opened 2017-01-11
Confused error message on unknown file


ai:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

UX bug

> Which version of stylelint are you using?

`7.7.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

```sh
stylelint file/doesnt/exists
```

> What did you expect to happen?

"File file/doesnt/exists doesn’t exists" error message.

> What actually happened (e.g. what warnings or errors you are getting)?

```sh
Error: Files glob patterns specified did not match any files
    at globby.then.filePaths (node_modules/stylelint/lib/standalone.js:71:33)
```

jeddy3:
@ai Thanks for the request.

Is the issue here that the current error message doesn't distinguish between single files and globs? And you think having a specific error message for single files would be a better user experience?

ai:
1. Of course. User didn’t enter a glob, but see some glob error. Also some users even don’t know what glob is.
2. Also we should write missed file name. If I saw what name was passed to Stylelint by lint-staged, debug could be faster. Maybe related also to issue #649.

davidtheclark:
How about just `whether-you-entered does not match any files`?

```
file/doesnt/exist does not match any files

**/things/*.css does not match any files

file/one file/two and file/three do not match any files
```

ai:
@davidtheclark nice idea :+1:

davidtheclark:
Closed by #2328.

PR #2328, opened 2017-02-09
Print out filenames/globs when files do not exist

CAYdenberg:

Issue #2251 

The formatting of multiple filenames/globs into a string is non-trivial. I addressed this with an immediately-executing anonymous function in the globby promise handler, because I wasn't sure if there's a standard place pure helper functions like this. Let me know if it could be improved.

### DIFF
--- a/lib/__tests__/standalone.test.js
+++ b/lib/__tests__/standalone.test.js
@@ -79,7 +79,7 @@ it("standalone without input css and file(s) should throw error", () => {
 })
 
 it("standalone with non-existent-file should throw error with code 80", () => {
-  const expectedError = new Error("Files glob patterns specified did not match any files")
+  const expectedError = new Error(`${fixturesPath}/non-existent-file.css does not match any files`)
   expectedError.code = 80
 
   return standalone({
--- a/lib/__tests__/standalone.test.js
+++ b/lib/__tests__/standalone.test.js
@@ -92,12 +92,42 @@ it("standalone with non-existent-file should throw error with code 80", () => {
   })
 })
 
-it("standalone with non-existent-file and allowEmptyInput false should throw error with code 80", () => {
-  const expectedError = new Error("Files glob patterns specified did not match any files")
+it("standalone with a single non-existent-file and allowEmptyInput false should throw error with code 80", () => {
+  const expectedError = new Error(`${fixturesPath}/non-existent-file.css does not match any files`)
   expectedError.code = 80
 
   return standalone({
     files: `${fixturesPath}/non-existent-file.css`,
+    config: configBlockNoEmpty,
+    allowEmptyInput: false,
+  }).then(() => {
+    throw new Error("should not have succeeded")
+  }).catch(actualError => {
+    expect(actualError).toEqual(expectedError)
+  })
+})
+
+it("standalone with two non-existent-files and allowEmptyInput false should throw error with code 80", () => {
+  const expectedError = new Error(`${fixturesPath}/non-existent-file.css and ${fixturesPath}/another-non-existent-file.css does not match any files`)
+  expectedError.code = 80
+
+  return standalone({
+    files: [ `${fixturesPath}/non-existent-file.css`, `${fixturesPath}/another-non-existent-file.css` ],
+    config: configBlockNoEmpty,
+    allowEmptyInput: false,
+  }).then(() => {
+    throw new Error("should not have succeeded")
+  }).catch(actualError => {
+    expect(actualError).toEqual(expectedError)
+  })
+})
+
+it("standalone with three or more non-existent-files and allowEmptyInput false should throw error with code 80", () => {
+  const expectedError = new Error(`${fixturesPath}/non-existent-file.css, ${fixturesPath}/another-non-existent-file.css and ${fixturesPath}/third-non-existent-file.css does not match any files`)
+  expectedError.code = 80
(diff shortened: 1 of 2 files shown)


lib/standalone.js
```diff
@@ -69,10 +69,16 @@ module.exports = function (options/*: Object */)/*: Promise<stylelint$standalone
     if (!filePaths.length) {
       if (allowEmptyInput === undefined || !allowEmptyInput) {
         const message = (files => {
-          const [ last, ...initial ] = files.slice(0).reverse()
-          const finalSeparator = (files.length > 1 ? " and" : "")
-          return `${initial.join(", ")} ${finalSeparator} ${last}`
-        })(files)
+          if (typeof files === "string") {
+            return files
+          } else {
```
evilebottnawi: No need here `else`, before you use `return` and already exit from block.

lib/__tests__/standalone.test.js
```diff
@@ -107,6 +107,36 @@ it("standalone with non-existent-file and allowEmptyInput false should throw err
...
+    files: [ `${fixturesPath}/non-existent-file.css`, `${fixturesPath}/another-non-existent-file.css`, `${fixturesPath}/third-non-existent-file.css` ],
+    config: configBlockNoEmpty,
+    allowEmptyInput: false,
+  }).then(() => {
+    throw new Error("should not have succeeded")
+  }).catch(actualError => {
+    expect(actualError).toEqual(expectedError)
+  })
+})
+
```
davidtheclark: Can you please add a test that uses one or more globs?

davidtheclark:
@CAYdenberg: I noticed your commit a little while ago. Do you feel like you've addressed the feedback? I usually wait to look again until the dev pings me and says the PR is ready for another look.

lib/__tests__/standalone.test.js
CAYdenberg: Addressed in db3ff5f

CAYdenberg:
Got it. I'll make sure to reply to the comment in future.

lib/__tests__/standalone.test.js
```diff
@@ -107,6 +107,81 @@ it("standalone with non-existent-file and allowEmptyInput false should throw err
   })
 })
 
+it("standalone with two non-existent-files and allowEmptyInput false should throw error with code 80", () => {
+  const expectedError = new Error(`${fixturesPath}/non-existent-file.css and ${fixturesPath}/another-non-existent-file.css does not match any files`)
```
davidtheclark: If we're using `x and y` syntax we also need to change the verb from `does` to `do`. I'd be fine simplifying it by reversing the syntax: `Could not find files matching x and y`.

lib/__tests__/standalone.test.js
```diff
@@ -107,6 +107,81 @@ it("standalone with non-existent-file and allowEmptyInput false should throw err
...
+
+  return standalone({
+    files: [`${fixturesPath}/nodir/**/*.css`],
+    config: configBlockNoEmpty,
+    allowEmptyInput: false,
+  }).then(() => {
+    throw new Error("should not have succeeded")
+  }).catch(actualError => {
+    expect(actualError).toEqual(expectedError)
+  })
```
davidtheclark: With this Promise usage, the `should not have succeeded` error will actually get caught by the `catch` statement, I believe. Although the test will still fail, the source of failure will be less clear. Can we instead do 
```js
.then(() => { 
  throw new Error("should not have succeeded")
}, actualError => {
 ..
})
```

for all these tests? Does that make sense?

lib/__tests__/standalone.test.js
```diff
@@ -107,6 +107,81 @@ it("standalone with non-existent-file and allowEmptyInput false should throw err
...
+  return standalone({
+    files: [ `${fixturesPath}/nodir/**/*.css`, `${fixturesPath}/anotherdir/**/*.css`, `${fixturesPath}/athird/**/*.css` ],
+    config: configBlockNoEmpty,
+    allowEmptyInput: false,
+  }).then(() => {
+    throw new Error("should not have succeeded")
+  }).catch(actualError => {
+    expect(actualError).toEqual(expectedError)
+  })
+})
```
davidtheclark: I think there's no need to repeat each of these tests with `allowEmptyInput=false`, because you're not really testing anything new there.

CAYdenberg: The repeats are actually for files vs. globs (except for the first two tests on lines 81 and 95, which I didn't write). Maybe declaring `allowEmptyInput=false` is confusing though? Or are there just an excessive number of tests?

lib/__tests__/standalone.test.js
CAYdenberg: It actually wasn't too difficult to just use "do". Addressed in #7782b54

lib/__tests__/standalone.test.js
CAYdenberg: Yeah that makes sense. Addressed in ff174e77

CAYdenberg:
This is ready to be reviewed again

davidtheclark:
Added to Changelog:

> -   Fixed: more helpful messages when file globs do not match any files ([#2328](https://github.com/stylelint/stylelint/pull/2328)).

jeddy3:
@CAYdenberg Awesome job!

Issue #2236, opened 2017-01-04
Disregard default ignores when ignoreFiles is used


bzang:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Target files located in any node_modules folder are ignore even when I specifically want to lint them.

> Which rule, if any, is this issue related to?

N/A

> What CSS is needed to reproduce this issue?

N/A

> What stylelint configuration is needed to reproduce this issue?

N/A

> Which version of stylelint are you using?

7.7.1

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylelint "packages/node_modules/**/*.css" --config .stylelintrc.yml -i node_modules`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

All files with extension css in the `packages/node_modules` folder to be linted

> What actually happened (e.g. what warnings or errors you are getting)?

Nothing happens

> Possible Solutions

- allow ignorePath to overwrite the default ignore
- expose alwaysIgnore as a config option that defaults to the current value (https://github.com/stylelint/stylelint/blob/master/lib/isPathIgnored.js#L28)
- change alwaysIgnore to only ignore root `node_modules` directory.

evilebottnawi:
@bzang why you need linting in `node_modules`? it is quite specific and would like to hear real use case

bzang:
@evilebottnawi. Our project uses the alle pattern <https://github.com/boennemann/alle>, which is a monorepo pattern that places all packages in a `packages/node_modules` folder. This allows node to traverse up the file tree to find modules that are in the repo and make them usable as imports. As it stands, even if we point stylelint to a specific file, the file does not get linted due to the hard coded `alwaysIgnore` values.

jeddy3:
@bzang Interesting use case. Do you encounter the same issue when using eslint as [they also ignore the `node_modules` directory by default](http://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories)?

bzang:
@jeddy3, eslint actually only ignores the base `node_modules` directory. That's why they specify `/node_modules/*`. Any other `node_modules` directory isn't automatically ignored.
I think this is a reasonable setup since the root `node_modules` directory is generally where npm installs everything.

jeddy3:
@bzang Thanks for looking into it.

> only ignores the base node_modules directory.

[...]

PR #2331, opened 2017-02-09
Add a note that node_modules and bower_components are always ignored

sapegin:
If my understanding of the default behavior is correct.


jeddy3:
@sapegin Thanks!

FYI, [this behaviour is going to change](https://github.com/stylelint/stylelint/issues/2236) in `8.0.0`. At the moment your update is accurate though.

PR #2333, opened 2017-02-10
Adds ignorePattern option to max-line-length

CAYdenberg:

#2219 

Pretty straightforward.

### DIFF
--- a/lib/rules/max-line-length/README.md
+++ b/lib/rules/max-line-length/README.md
@@ -108,3 +108,19 @@ a { color: pink; }
  */
 a { color: pink; }
 ```
+
+### `ignorePattern: ["/regex/"]`
+
+Ignore any line that matches the given regex pattern, regardless of whether it is comment or not.
+
+Given:
+
+```js
+['/^\\@import\\s+/']
+```
+
+The following is *not* considered a warning regardless of line length:
+
+```css
+@import "../../../../another/css/or/scss/file/or/something.css";
+```
--- a/lib/rules/max-line-length/README.md
+++ b/lib/rules/max-line-length/README.md
@@ -116,7 +116,7 @@ Ignore any line that matches the given regex pattern, regardless of whether it i
 Given:
 
 ```js
-['/^\\@import\\s+/']
+["/^\\@import\\s+/"]
 ```
 
 The following is *not* considered a warning regardless of line length:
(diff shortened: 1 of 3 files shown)


lib/rules/max-line-length/README.md
```diff
@@ -108,3 +108,19 @@ a { color: pink; }
...
 ```
+
+### `ignorePattern: ["/regex/"]`
+
+Ignore any line that matches the given regex pattern, regardless of whether it is comment or not.
+
+Given:
+
+```js
+["/^\\@import\\s+/"]
```
davidtheclark: I don't think you need to escape `@`, do you?

hudochenkov:
@davidtheclark do you think it should support support RegExp either? I've noticed some of our rule/options, which expect some kind of RegExp, aren't consistent on that matter. [Some of them](https://github.com/stylelint/stylelint/blob/d4260a06f99ec8be6a8a78ea14fcbfc8c5bce347/lib/rules/at-rule-empty-line-before/index.js#L46) support only `string`, [other](https://github.com/stylelint/stylelint/blob/d4260a06f99ec8be6a8a78ea14fcbfc8c5bce347/lib/rules/custom-media-pattern/index.js#L19-L22) support both.

lib/rules/max-line-length/README.md
CAYdenberg: I was not aware of t@ but it appears you're right. 

I changed it in the tests as well.

CAYdenberg:
@hudochenkov The example (at-rule-empty-line-before) only applies to @ rules ... I think this should be more flexible since there are other places you might want it. Someone might want "/linear-gradient/" for example.

davidtheclark:
> it should support support RegExp either

@hudochenkov: Can you please try to rephrase this question? It's not clear to me what you're asking.

Regarding the inconsistencies — yeah, there are a lot of rules :) If you would like to add regexp support to any options that only support strings, please open an issue or submit a PR — I usually can't think of a reason *not* to enable regexps.

hudochenkov:
@davidtheclark sorry :) I wanted to know if `ignorePattern` should support `RegExp` additional to `string`. Now I see that it should support both types.

lib/rules/max-line-length/README.md
CAYdenberg: Addressed in 07f4634

jeddy3:
@CAYdenberg Thanks! 

Changelog updated:

> -   Added: `ignorePattern` option to `max-line-length` ([#2333](https://github.com/stylelint/stylelint/pull/2333)).

Issue #2336, opened 2017-02-10
False positive for selector-type-no-unknown


SDSLeon:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

For code like this rule selector-type-no-unknown show error:

> Which rule, if any, is this issue related to?

selector-type-no-unknown

> What CSS is needed to reproduce this issue?

```css
input {
        &:not(checked) {
            position: absolute;
            opacity: 0;
        }
}
```

> What stylelint configuration is needed to reproduce this issue?

stylelint-config-standard

> Which version of stylelint are you using?

e.g. `^7.8.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

webpack plugin

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

e.g. "Yes, it's related to postCSS with nested"

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

The following error were flagged: Unexpected unknown type selector "checked"

evilebottnawi:
@SDSLeon Thanks for feedback, accepted.

hudochenkov:
Thank you for a report. Are you sure you wrote your CSS correctly? I believe correct way is with a colon before `checked` if you want to target non-checked inputs:

```scss
input {
        &:not(:checked) {
            position: absolute;
            opacity: 0;
        }
}
```

http://codepen.io/hudochenkov/pen/PWxbJd

evilebottnawi:
@hudochenkov oh yep, it slipped from my mind :smile: , but be good added more test.

evilebottnawi:
Rule works as expected. I am add more tests.

PR #2337, opened 2017-02-10
Tests: more tests for selector-type-no-unknown (#2336).

evilebottnawi:
#2336

No, it's self explanatory.

### DIFF
--- a/lib/rules/selector-type-no-unknown/__tests__/index.js
+++ b/lib/rules/selector-type-no-unknown/__tests__/index.js
@@ -50,6 +50,16 @@ testRule(rule, {
   }, {
     code: "a:nth-child(n) {}",
   }, {
+    code: "input:not(*) {}",
+  }, {
+    code: "input:not(:checked) {}",
+  }, {
+    code: "input:first-child:not(:checked) {}",
+  }, {
+    code: "input:not(:nth-child(2n+1)) {}",
+  }, {
+    code: ":not(::first-line) {}",
+  }, {
     code: "circle {}",
     description: "svg tags",
   }, {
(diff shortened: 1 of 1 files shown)


lib/rules/selector-type-no-unknown/__tests__/index.js
```diff
@@ -50,6 +50,16 @@ testRule(rule, {
   }, {
     code: "a:nth-child(n) {}",
   }, {
+    code: "input:not(*) {}",
```
hudochenkov: I'm not sure what elements will be targeted by this selector. Do you have idea? :) It targets `input`, which should not be targeted by `*`. But every `input` is targeted by `*`.

lib/rules/selector-type-no-unknown/index.js
```diff
@@ -85,7 +85,9 @@ const rule = function (actual, options) {
           const tagName = tagNode.value
           const tagNameLowerCase = tagName.toLowerCase()
 
-          if (htmlTags.indexOf(tagNameLowerCase) !== -1 || svgTags.indexOf(tagNameLowerCase) !== -1 || nonStandardHtmlTags.has(tagNameLowerCase)) {
+          if (htmlTags.indexOf(tagNameLowerCase) !== -1 || svgTags.indexOf(tagNameLowerCase) !== -1
```
hudochenkov: I think this change is unrelated to this PR.

evilebottnawi: @hudochenkov i think we should change the length of the string to 120 (щк 80), it's just not possible to read, maybe it should be done in another PR.

lib/rules/selector-type-no-unknown/__tests__/index.js
evilebottnawi: @hudochenkov see first note https://developer.mozilla.org/ru/docs/Web/CSS/:not :smile: it is possible

Issue #2332, opened 2017-02-09
False positive for SVG reference interaction in color-hex-case


Sinistralis:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

I am describing a bug involving interactions between the color-hex-case rule and stroke/fill syntax where you can use the '#' identifier in a url declaration to reference an SVG.

> Which rule, if any, is this issue related to?

color-hex-case

> What CSS is needed to reproduce this issue?

```css
.foo {
  stroke: url(#gradiantA)
}
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "color-hex-case": [ "lower", { "severity": "error" } ]
  }
}

upper would work here as well obviously
```

> Which version of stylelint are you using?

7.8.0

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI: stylelint src/scss/**/*.scss --syntax scss --config stylelint.config.js

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

I am using SCSS but this should be reproducable on CSS as well.

> What did you expect to happen?

No warning/error since the above is not a hex color code.

> What actually happened (e.g. what warnings or errors you are getting)?

```shell
Expected "#gradientA" to be "#gradienta"   color-hex-case
```

I will work on trying to get something put together in my spare time, it might take a bit though.

davidtheclark:
@Sinistralis I think this problem could be avoided if the rule always ignores the content of `url` functions. So if you or anybody else picks this up, that's probably the thing to try first.

Sinistralis:
Agree, that was the route I was going to go if I got to it first.

evilebottnawi:
Done in https://github.com/stylelint/stylelint/pull/2338

PR #2338, opened 2017-02-10
Fixed: `color-hex-case` rule is now respect SVG reference interaction.

evilebottnawi:

#2332

No, it's self explanatory.


davidtheclark:
Changelog update:

> -   Fixed: `color-hex-case` no longer reports false positives for id references in `url` functions ([#2338](https://github.com/stylelint/stylelint/pull/2338).

Issue #2330, opened 2017-02-09
Malformed message in shorthand-property-no-redundant-values


bedney:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

This rule produces a malformed fix. See below.

> Which rule, if any, is this issue related to?

`shorthand-property-no-redundant-values`

> What CSS is needed to reproduce this issue?

```css
h1 {
    font-size: 123.1%;
    margin: 1em 0 2em 0;
}

```

> What stylelint configuration is needed to reproduce this issue?

```json
{
    "extends": "stylelint-config-standard",
    "rules": {
        "at-rule-empty-line-before": null,

        "comment-empty-line-before": null,

        "block-no-empty": [true, { "severity": "warning" }],
        "block-opening-brace-newline-after": "always",

        "declaration-empty-line-before": null,

        "indentation": 4,

        "font-family-name-quotes": "always-unless-keyword",

        "function-url-quotes": "always",

        "rule-nested-empty-line-before": null,

        "selector-attribute-quotes": "always",
        "selector-type-no-unknown": null,

        "string-quotes": "double",

        "value-keyword-case": "lower"
    }
}
```

> Which version of stylelint are you using?

`7.8.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

"CLI with `stylelint *.css`"

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

Proper CSS corrective declarations to be emitted.

> What actually happened (e.g. what warnings or errors you are getting)?

The following corrective CSS was produced. Note the weird `0:2em` corrective suggestion:

```shell
 66:5  ✖  Unexpected longhand value '1em 0 2em   shorthand-property-no-redundant-values
          0' instead of '1em 0:2em'
```

Thanks!

bedney:
I'm not an expert on CSS tri-value syntax, but it looks like this might just be a bug where the output is using a colon (':') rather than a space (' ').

bedney:
Further information: it only looks like this is happening when the value that's being 'repeated' and is 'unnecessary' is the value `0`.

evilebottnawi:
@bedney Thanks for feedback. Fell free to send a PR :smile:

bedney:
[...]

PR #2339, opened 2017-02-10
Tests: `shorthand-property-no-redundant-values` rule produces normal message (#2330).

evilebottnawi:

#2330

No, it's self explanatory.

### DIFF
--- a/lib/rules/shorthand-property-no-redundant-values/__tests__/index.js
+++ b/lib/rules/shorthand-property-no-redundant-values/__tests__/index.js
@@ -230,6 +230,11 @@ testRule(rule, {
     message: messages.rejected("thin thin thin thin", "thin"),
     line: 1,
     column: 5,
+  }, {
+    code: "a { margin: 1em 0 2em 0; }",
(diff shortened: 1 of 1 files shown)


lib/rules/shorthand-property-no-redundant-values/__tests__/index.js
```diff
@@ -230,6 +230,11 @@ testRule(rule, {
     message: messages.rejected("thin thin thin thin", "thin"),
     line: 1,
     column: 5,
+  }, {
+    code: "a { margin: 1em 0 2em 0; }",
+    message: messages.rejected("1em 0 2em 0", "1em 0 2em"),
+    line: 1,
+    column: 6,
```
jeddy3: I think this should be `5`.

evilebottnawi: @jeddy3 done, stupid error :smile:

davidtheclark:
Thanks @evilebottnawi.

Issue #2325, opened 2017-02-07
--report-needless-disables should exit with non-zero code 


evilebottnawi:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

```shell
stylelint '{**/*,*}.?(s)css' --ignore-path .gitignore --rd
```

> Which rule, if any, is this issue related to?

No matter

> What CSS is needed to reproduce this issue?

No matter

> What stylelint configuration is needed to reproduce this issue?

No matter

> Which version of stylelint are you using?

latest

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No matter

> What did you expect to happen?

Error code is not equal to 0.

> What actually happened (e.g. what warnings or errors you are getting)?

```
src/components/logo/type-1/_type-1.scss
start: 5, end: 14
```

Error code is equal to 0.

/cc @jeddy3 @davidtheclark

evilebottnawi:
Done in https://github.com/stylelint/stylelint/pull/2341

PR #2341, opened 2017-02-10
Fixed: `report-needless-disables` argument is now exit with non-zero code.

evilebottnawi:

#2325 

No, it's self explanatory.

It is very strange :smile:  `reportNeedlessDisables` can be only `true` or `false`.

/cc @jeddy3 @davidtheclark

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -207,10 +207,19 @@ Promise.resolve().then(() => {
   return standalone(options)
 }).then((linted) => {
   if (reportNeedlessDisables) {
-    process.stdout.write(needlessDisablesStringFormatter(linted.needlessDisables))
-    if (reportNeedlessDisables === "error") {
+    const hasReportNeedlessDisable = !!linted.needlessDisables && linted.needlessDisables.some((sourceReport) => {
+      if (!sourceReport.ranges || sourceReport.ranges.length === 0) {
+        return false
+      }
+
+      return true
+    })
+
+    if (hasReportNeedlessDisable) {
+      process.stdout.write(needlessDisablesStringFormatter(linted.needlessDisables))
       process.exitCode = 2
     }
+
     return
   }
 


davidtheclark:
> It is very strange 😄 reportNeedlessDisables can be only true or false

@evilebottnawi Why is that strange?

evilebottnawi:
@davidtheclark `reportNeedlessDisables` can never be `error`

evilebottnawi:
@davidtheclark i bad know `flow`, what is wrong? (in CI)

lib/cli.js
```diff
@@ -207,10 +207,19 @@ Promise.resolve().then(() => {
   return standalone(options)
 }).then((linted) => {
   if (reportNeedlessDisables) {
-    process.stdout.write(needlessDisablesStringFormatter(linted.needlessDisables))
-    if (reportNeedlessDisables === "error") {
+    const hasReportNeedlessDisable = linted.needlessDisables.some((sourceReport) => {
```
davidtheclark: Based on the type definitions, `needlessDisables` is not *guaranteed* to be on the `linted` object, so you can't call `some` on it without first checking whether it exists. An easy way to fix this with lodash, I think: `_.get(linted, 'needlessDisables', []).some(..)`.

davidtheclark: Or, without lodash, `const hasReportNeedlessDisable = !!linted.needlessDisables && linted.needlessDisables.some(..)`

evilebottnawi:
@davidtheclark i think it is now breaking changes, because all error should return non zero exit code, i think it is fix

davidtheclark:
Ehhh, maybe you're right. Didn't we have a semver policy doc? I can't find it!

evilebottnawi:
@davidtheclark ohh, i can't find too :smile:
/cc @jeddy3 what do you think about semver doc and about this issue?

hudochenkov:
@davidtheclark @evilebottnawi https://github.com/stylelint/stylelint/blob/v8/README.md#semantic-versioning-policy :)

jeddy3:
> However, this is a breaking change, so needs to be merged into v8

Yep, let's do that as there's a chance this change could break builds.

davidtheclark:
@evilebottnawi Can you please change the target branch to v8, then we can merge?

evilebottnawi:
@davidtheclark yep, in this week

davidtheclark:
👍

jeddy3:
`v8` changelog: 

> -   Changed: `report-needless-disables` now exits with non-zero code ([#2341](https://github.com/stylelint/stylelint/issues/2341)).

Issue #2334, opened 2017-02-10
False positives for resolved and interpolated selectors in selector-class-pattern


kaelig:
**PR with failing test: #2335 (help wanted)**

Note: could be related to https://github.com/stylelint/stylelint/issues/988

> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug where resolved nested selector class name doesn't validate correctly.

> Which rule, if any, is this issue related to?

`selector-class-pattern`

> What CSS is needed to reproduce this issue?

```css
.foo-#{$interpolated}-class-name {
  width: 0;

  // This throws a warning
  &:hover {
    width: 100%;
  }
}

.foo-class-name {
  width: 0;

  // No warning, yay:
  &:hover {
    width: 100%;
  }
}
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "selector-class-pattern": [
      "^foo(-[a-z0-9_-]+|)$",
      {
        "severity": "error",
        "resolveNestedSelectors": true
      }
    ]
  }
}
```

> Which version of stylelint are you using?

`7.8.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

`stylelint index.scss --syntax scss`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes, related to SCSS

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

The following warnings were flagged:

```shell
index.scss
 5:3  ✖  Expected class selector ".foo-#{$interpolated}-class-name" to match specified pattern   selector-class-pattern
```

jeddy3:
@kaelig Thanks for the report, using the template and for the test.

This appears to be instance we've not accounted for. Fixing it should just be the simple case of returning early if the _resolved selector_ is not standard, like [we already do](https://github.com/stylelint/stylelint/blob/master/lib/rules/selector-class-pattern/index.js#L48-L50) with the _unresolved one_.

Labeling as `help wanted` and `beginner friendly`.

evilebottnawi:
Done in https://github.com/stylelint/stylelint/pull/2344

PR #2344, opened 2017-02-11
Fixed: false positives for resolved and interpolated selectors in `selector-class-pattern`.

evilebottnawi:

#2334

For `selector-class-pattern`.

No, it's self explanatory.


davidtheclark:
Added to changelog:

> -   Fixed: `selector-class-pattern` false positive in SCSS when combining interpolated and nested selectors ([#2344](https://github.com/stylelint/stylelint/pull/2344)).

kaelig:
Thank you so much @evilebottnawi!

PR #2345, opened 2017-02-11
Tests: added tests for nested selectors in `selector-id-pattern` rule.

evilebottnawi:

#2334 

Add tests. Note need add `resolveNestedSelectors` option for respect nested selectors and consistently with `selector-class-pattern`. If you agree create issue :smile: 

@kaelig I was a little confusing about name #2334 issue, with `selector-id-pattern` all works fine. Or add tests example where have error.

No, it's self explanatory.

### DIFF
--- a/lib/rules/selector-id-pattern/__tests__/index.js
+++ b/lib/rules/selector-id-pattern/__tests__/index.js
@@ -59,5 +59,17 @@ testRule(rule, {
   }, {
     code: ".#{$a} {}",
     description: "ignore sass var interpolation",
+  }, {
+    code: ".foo { .#{$a} {} }",
+    description: "ignore sass var nested interpolation",
+  }, {
+    code: "a { #ABABA {} }",
   } ],
+
+  reject: [{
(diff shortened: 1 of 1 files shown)


jeddy3:
> Note need add `resolveNestedSelectors` option for respect nested selectors and consistently with `selector-class-pattern`. 

Let's only add it if there's a specific request for it.

> @kaelig I was a little confusing about name #2334 issue, with `selector-id-pattern` all works fine. 

This was my bad. I updated the issue title assuming the issue also impacted `selector-id-pattern`. I've amended the issue title.

evilebottnawi:
@jeddy3 now `selector-id-pattern` report invalid error.
```scss
a {
  &#foo {
    color: red;
  }
}
```
Now: `Expected id selector "#foo" to match specified pattern`.
Right: `Expected id selector "a#foo" to match specified pattern`.

jeddy3:
`#foo` is the _id_ selector. `a` is a _type_ selector. 

So, I think we're fine here.

Issue #2136, opened 2016-11-28
False negative for nested media queries in selector-max-compound-selectors


evilebottnawi:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Expected a different behavior.

Ref: https://github.com/stylelint/stylelint/pull/2102#issue-191053388

> Which rule, if any, is this issue related to?

`selector-max-compound-selectors`

> What CSS is needed to reproduce this issue?

e.g.

```scss
.foo .bar > .foobar { 
  & .test { 
    @media screen { 
      color: red; 
    }
  } 
}
```

> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
  "rules": {
    "selector-max-compound-selectors": "3"
  }
}
```

> Which version of stylelint are you using?

`7.6.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

No matter

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

SCSS

> What did you expect to happen?

Throw error.

> What actually happened (e.g. what warnings or errors you are getting)?

Nothing

PR #2350, opened 2017-02-11
Report every selector resolved selector in selector-max-compound-selectors

hudochenkov:

#2136

The only way I see to avoid incorrectly resolved selector in custom syntaxes is report every resolved selector rather the deeper one. Because variation of nesting may be very different.

I have a question about testing new behaviour. The rule now produced more than one warning and we should test for more than one warning. E. g. [this test](https://github.com/stylelint/stylelint/blob/db2098115a3b53fbc1a2c218328159701057b649/lib/rules/selector-max-compound-selectors/__tests__/index.js#L177) produce two warnings, but we can test only the first one.

### DIFF
--- a/lib/rules/selector-max-compound-selectors/__tests__/index.js
+++ b/lib/rules/selector-max-compound-selectors/__tests__/index.js
@@ -149,13 +149,32 @@ testRule(rule, {
     code: ".cd { .de { .fg > & {} } }",
     message: messages.expected(".fg > .cd .de", 2),
   }, {
+    code: "a { b { c {} } }",
+    description: "standard nesting",
+  }, {
+    code: "a b { c { top: 10px; } }",
+    description: "standard nesting with declaration",
+    message: messages.expected("a b c", 2),
+  }, {
+    code: "a b { c { d {} } }",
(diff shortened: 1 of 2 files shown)


hudochenkov:
@davidtheclark I think we should test more than one to be sure we report every violation, not the first one only.

Previously, with option set to `2` rule `a b { c { d {} } }` report about `a b c d` selector only. Now it report about `a b c` and `a b c d`, but in tests we can check the first one `a b c` only.

davidtheclark:
@hudochenkov I don't know ... we've never had to do this with the gazillion tests we've done so far. But if you think it's essential, feel free to work on a proposal for good way to make it happen, and we'll go from there.

hudochenkov:
I was thinking about it and decided don't make testing more complicated. I believe, it tests what it needs to test right now.

davidtheclark:
👍

jeddy3:
Excellent.

`v8` changelog:

> -   Changed: `selector-max-compound-selectors` now checks all resolved selectors, rather than just the deepest ([#2350](https://github.com/stylelint/stylelint/issues/2350)).

Issue #2352, opened 2017-02-12
Conflict with Modernizr


andrey-hohlov:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

When installed Modernizr and gulp-stylelint lint task throw error.

```
"devDependencies": {
  "gulp": "^3.9.1",
  "gulp-stylelint": "^3.8.0"
},
"dependencies": {
  "modernizr": "^3.3.1"
}
```
1. Run `npm i` for install dependencies
2. Run `gulp lint-css`

No errors if install gulp-stylelint first, then install Modernizr:

1. Run `npm i gulp gulp-stylelint modernizr` instead of `npm i`
2. Run `gulp lint-css`

Issue in gulp-stylelint: https://github.com/olegskl/gulp-stylelint/issues/71

> Which rule, if any, is this issue related to?

irrelevant

> What CSS is needed to reproduce this issue?

irrelevant

> What stylelint configuration is needed to reproduce this issue?

irrelevant

> Which version of stylelint are you using?

gulp-stylelint `3.8.0` (stylelint `7.8.0`)

> How are you running stylelint: CLI, PostCSS plugin, Node API?

``` javascript
const gulp = require('gulp');

gulp.task('lint-css', function lintCssTask() {
  const gulpStylelint = require('gulp-stylelint');
  return gulp
    .src('style.css')
    .pipe(gulpStylelint({
    reporters: [
      {formatter: 'string', console: true}
    ]
  }));
});
```
> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

irrelevant

> What did you expect to happen?

css linted

> What actually happened (e.g. what warnings or errors you are getting)?

```
TypeError: Cannot read property '0' of undefined
    at _.omitBy (/Users/andrey/Work/Repositories/test/node_modules/stylelint/lib/formatters/jsonFormatter.js:8:48)
    at /Users/andrey/Work/Repositories/test/node_modules/lodash/lodash.js:11233:17
    at /Users/andrey/Work/Repositories/test/node_modules/lodash/lodash.js:3085:13
    at /Users/andrey/Work/Repositories/test/node_modules/lodash/lodash.js:3925:15
    at baseForIn (/Users/andrey/Work/Repositories/test/node_modules/lodash/lodash.js:2509:40)
    at basePickBy (/Users/andrey/Work/Repositories/test/node_modules/lodash/lodash.js:3084:7)
[...]

PR #2353, opened 2017-02-12
Update lodash

hudochenkov:

#2352

I think we should release a patch version.


davidtheclark:
@hudochenkov Thanks for looking into that issue. Let's cut a release tomorrow, including whatever we can finish merging today.

davidtheclark:
Added to changelog:

> -   Fixed: update version of `lodash` dependency to match feature usage, fixing a conflict with Modernizr ([#2353](https://github.com/stylelint/stylelint/pull/2353)).

Issue #1780, opened 2016-08-09
False negatives for some properties in declaration-block-no-redundant-longhand-properties and declaration-block-no-shorthand-property-overrides


evilebottnawi:
New shothand:
- `animation`
- `border-block-end`
- `border-block-start`
- `border-image`
- `border-inline-end`
- `border-inline-start`
- `column-rule`
- `columns`
- `flex`
- `flex-flow`
- `grid`
- `grid-area`
- `grid-column`
- `grid-gap`
- `grid-row`
- `grid-template`
- `mask`
- `outline`
- `text-decoration`
- `text-emphasis`
- `transition`

This can do breaking changes.

jeddy3:
Is there a definitive resource where we can be sure we've got these shorthand properties?

What about upcoming specs like grid?

evilebottnawi:
@jeddy3 tomorrow i'll look at all the properties in the specification. Label `need investigation`?

evilebottnawi:
@jeddy3 i refresh all shorthand properties in first message, feel free to edit and discussion.

jeddy3:
Sgtm.

ideadapt:
@jeddy3 I recently found https://drafts.csswg.org/indexes/#properties. Its looks quite interesting too me.
So, I started to scrape that page, see https://github.com/a-axton/css-properties-values/pull/1. In the end you have a list of properties filterable by e.g. css version, css module, name, values etc.

jeddy3:
@evilebottnawi What rules are affected by this change?

evilebottnawi:
@jeddy3 `declaration-block-no-redundant-longhand-properties`, `declaration-block-no-shorthand-property-overrides`

evilebottnawi:
@jeddy3 what about `8.0.0` milestone here?

jeddy3:
Yes. This is a pre-`8.0.0` *major* change. After that it would be *minor*.

davidtheclark:
Closed by #2354.

PR #2354, opened 2017-02-12
Add new shorthand data

hudochenkov:

#1780 

Added new shorthand properties from #1780, except `mask`, because [it's not clear](https://developer.mozilla.org/en-US/docs/Web/CSS/mask) what for this shorthand.

Removed prefixed shorthands. I don't think it's a good idea to keep list of all prefixed properties and shorthand.

I have concerns about `shorthand-property-no-redundant-values`. It uses `shorthandData` also. My concers is that it has a [list of ignored shorthands](https://github.com/stylelint/stylelint/blob/4ad8fbdbc4d126c618bd70d98bf94e4b207aefb8/lib/rules/shorthand-property-no-redundant-values/index.js#L33-L43). With new shorthands ignore list will be much bigger than list of shorthands supported in this rule. Maybe we should specify the list of allowed shorthands, rather using `shorthandData` and ignore.


davidtheclark:
> except mask, because it's not clear what for this shorthand.

davidtheclark:
> except mask, because it's not clear what for this shorthand.

Why is `mask` different from the others? I have never used it, but it's real! e.g. https://css-tricks.com/clipping-masking-css/

> Removed prefixed shorthands. I don't think it's a good idea to keep list of all prefixed properties and shorthand.

I agree, but I don't like that we're removing functionality for that. Can we use Autoprefixer in the rule to remove the need for a manual list?

> Maybe we should specify the list of allowed shorthands, rather using shorthandData and ignore.

Sounds good to me!

Want to add the changes above to this PR?

hudochenkov:
> Why is mask different from the others? I have never used it, but it's real! e.g. https://css-tricks.com/clipping-masking-css/

I know it's real, but I don't know every property for this shorthand. [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/mask) and [W3C](https://drafts.fxtf.org/css-masking-1/#the-mask) don't make it clear. They say about `mask-clip`, `mask-origin`, and `mask-border` only, but `mask` accept more than this properties, as I see.

> I agree, but I don't like that we're removing functionality for that. Can we use Autoprefixer in the rule to remove the need for a manual list?

I will look how `declaration-block-no-redundant-longhand-properties`, `declaration-block-no-shorthand-property-overrides`, and `shorthand-property-no-redundant-values` work and see what can be done.

> > Maybe we should specify the list of allowed shorthands, rather using shorthandData and ignore.

> Sounds good to me!

> Want to add the changes above to this PR?

Sure!

davidtheclark:
@hudochenkov This page looks helpful: https://tympanus.net/codrops/css_reference/mask/

hudochenkov:
Benchmarks for https://github.com/stylelint/stylelint/pull/2354/commits/ac40d7a4a2b82bbd72837b7911cccfa31e4189a3

Before:

```
> stylelint@7.8.0 benchmark-rule /Users/aleks/projects/stylelint
> node scripts/benchmark-rule.js "declaration-block-no-shorthand-property-overrides" "true"

Warnings: 0
Mean: 45.400157870370364 ms
Deviation: 11.06851959653398 ms

Mean: 41.5444920952381 ms
Deviation: 11.320279370467999 ms

Mean: 41.935588403508774 ms
Deviation: 10.520627666926606 ms

Mean: 41.81118664912281 ms
Deviation: 11.763514223194033 ms

Mean: 41.28931149206349 ms
Deviation: 10.750820788142141 ms
```

After:

```
Warnings: 0
Mean: 40.35173133620689 ms
Deviation: 10.237891375944221 ms

Mean: 43.27671090243902 ms
Deviation: 11.477326149577069 ms

Mean: 40.06621774358975 ms
Deviation: 11.043813336306489 ms

Mean: 40.419671480620174 ms
Deviation: 12.434897384556843 ms

Mean: 42.71238819642859 ms
Deviation: 9.455533365058331 ms

Mean: 39.48838472435897 ms
Deviation: 10.162578651704262 ms
```

lib/rules/declaration-block-no-redundant-longhand-properties/README.md
```diff
@@ -36,8 +36,28 @@ This rule warns when the following shorthand properties can be used:
...
+-   `flex-flow`
+-   `grid`
+-   `grid-area`
+-   `grid-column`
+-   `grid-gap`
+-   `grid-row`
+-   `grid-template`
+-   `outline`
+-   `text-decoration`
+-   `text-emphasis`
```
davidtheclark: Wow, this is a lot of new power!

lib/rules/declaration-block-no-shorthand-property-overrides/index.js
```diff
@@ -18,8 +18,13 @@ const rule = function (actual) {
...
 
-    root.walkRules(check)
-    root.walkAtRules(check)
+    root.walk((node) => {
+      if (node.type !== "rule" && node.type !== "atrule") {
+        return
+      }
+
+      check(node)
+    })
```
davidtheclark: Why is this change relevant?

lib/rules/shorthand-property-no-redundant-values/index.js
```diff
@@ -15,7 +14,14 @@ const messages = ruleMessages(ruleName, {
   rejected: (unexpected, expected) => `Unexpected longhand value '${unexpected}' instead of '${expected}'`,
 })
 
-const shorthandableProperties = new Set(Object.keys(shorthandData))
+const shorthandableProperties = new Set([
```
davidtheclark: Can you please add a comment explaining why these specific properties are checked, and not others?

lib/rules/declaration-block-no-shorthand-property-overrides/index.js
hudochenkov: It's kind of irrelevant. Benchmarks [here](https://github.com/stylelint/stylelint/pull/2354#issuecomment-280312263). Should I revert this change?

lib/rules/shorthand-property-no-redundant-values/index.js
hudochenkov: Done.  Not sure if text correct, though. Also added `grid-gap`. Maybe some other Grid properties can be added, but I can't say for sure.

lib/rules/declaration-block-no-redundant-longhand-properties/README.md
jeddy3: This PR is looking great!

lib/rules/declaration-block-no-shorthand-property-overrides/index.js
davidtheclark: Is this change the only thing that caused the slight speed-up? Does such a change reliably cause speed-ups in other rules? What we had before matches a pattern that is widely used throughout the rules, so I'd prefer to stick with it unless we get some significant benefit from changing it.

hudochenkov: Reverted.

davidtheclark:
Added to changelog:

> -   Fixed: `declaration-block-no-redundant-longhand-properties` and `declaration-block-no-shorthand-property-overrides` understand more shorthand properties ([#2354](https://github.com/stylelint/stylelint/pull/2354)).

PR #2358, opened 2017-02-13
Add flow annotations: coverage in `createStylelintResult.js` from 51.…

ivanzusko:


davidtheclark:
All of these annotations are within the function body, not for function signature; so, as we discussed elsewhere, I think we should close it. Let's only focus on function signatures and anything that Flow specifically requires — while allowing Flow to make as many inferences as it can within function bodies.

Issue #2359, opened 2017-02-14
False positive for placeholder selector in selector list in selector-type-case


xi:
I found a bug with `selector-type-case` and SCSS placeholder selectors that only seems to occur when they are used in a selector group.

> Which rule, if any, is this issue related to?

`selector-type-case`

> What CSS is needed to reproduce this issue?

`.foo, %foo {}`

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "selector-type-case": "upper"
  }
}
```

> Which version of stylelint are you using?

7.8.0

> How are you running stylelint: CLI, PostCSS plugin, Node API?

`stylelint --syntax scss ...`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes, it is related to [SCSS placeholder selectors](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#placeholder_selectors_)

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

```
Expected "%foo" to be "%FOO"   selector-type-case
```

evilebottnawi:
@xi Thanks for feedback and PR

jeddy3:
Done in https://github.com/stylelint/stylelint/issues/2360

@xi Thank you!

PR #2360, opened 2017-02-14
Fix placeholder selector case

xi:

#2359

No, it's self explanatory.

### DIFF
--- a/lib/utils/isStandardSyntaxTypeSelector.js
+++ b/lib/utils/isStandardSyntaxTypeSelector.js
@@ -31,5 +31,9 @@ module.exports = function (node/*: Object*/)/*: boolean*/ {
     return false
   }
 
+  if (node.value[0] === '%') {
+    return false
+  }
+
   return true
 }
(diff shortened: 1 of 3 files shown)


lib/utils/isStandardSyntaxTypeSelector.js
```diff
@@ -31,5 +31,9 @@ module.exports = function (node/*: Object*/)/*: boolean*/ {
     return false
   }
 
+  if (node.value[0] === '%') {
```
evilebottnawi: Use `"` quotes

xi:
Thanks for the quick reply! I adjusted the quotes.

davidtheclark:
Great! Feel free to merge and add a changelog entry.

I just remembered that @hudochenkov suggested we release soon to get the Modernizr-conflict resolved. Should we do that after this merge?

jeddy3:
Changelog

> -   Fixed: `selector-type-case` false positive for placeholder selectors ([#2360](https://github.com/stylelint/stylelint/pull/2360)).

PR #2361, opened 2017-02-14
Add flow annotations

ivanzusko:
Added more annotations and a bit increased flow coverage on the getPostcssResult.js (from 77.27% to 79.10%)


lib/getPostcssResult.js
```diff
@@ -1,5 +1,11 @@
 /* @flow */
 "use strict"
+
+/*:: type PostcssOptions = {
```
davidtheclark: This is the right idea, I think. Did you see that we're defining these `type`s in declaration files? The PostCSS one is https://github.com/stylelint/stylelint/blob/master/decls/postcss.js.

lib/getPostcssResult.js
```diff
@@ -19,7 +25,7 @@ module.exports = function (stylelint/*: stylelint$internalApi*/)/*: Promise<?Obj
     customSyntax?: string
   }*/ = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : {}
 
-  const cached = stylelint._postcssResultCache.get(options.filePath)
+  const cached/*: Object*/ = stylelint._postcssResultCache.get(options.filePath)
```
davidtheclark: This should be a PostCSS Result, right — not just an Object? I'm not sure the Object annotation is very helpful, because it does not add any information. If you want to create a new `postcss$result` `type` declaration, that could be helpful!

lib/getPostcssResult.js
```diff
@@ -70,8 +76,8 @@ module.exports = function (stylelint/*: stylelint$internalApi*/)/*: Promise<?Obj
       postcssOptions.syntax = syntax
     }
 
-    const source = options.code ? options.codeFilename : options.filePath
-    let preProcessedCode = code
+    const source/*: ?string*/ = options.code ? options.codeFilename : options.filePath
+    let preProcessedCode/*: string*/ = code
```
davidtheclark: Let's remove these two annotations, as they don't add any new information.

lib/getPostcssResult.js
ivanzusko: yes, you are right, @davidtheclark , it's not just an Object, my bad... besides it could be `undefined` too. I'm trying to figure out the structure of new `postcss$result` for creating type declaration. Hope I'll resolve it

lib/getPostcssResult.js
ivanzusko: @davidtheclark sorry, moved it to declarations file

lib/getPostcssResult.js
```diff
@@ -44,7 +45,7 @@ module.exports = function (stylelint/*: stylelint$internalApi*/)/*: Promise<?Obj
         throw new Error(`Cannot resolve custom syntax module ${customSyntax}`)
       }
     } else {
-      const fileExtension = path.extname(options.filePath || "")
+      const fileExtension/*: string*/ = path.extname(options.filePath || "")
```
davidtheclark: No need for this one. It will be inferred.

decls/postcss.js
```diff
@@ -32,3 +32,9 @@ export type postcss$rule = {
   parent: Object,
   nodes: Array<Object>,
 }
+
+export type postcss$Options = {
```
davidtheclark: To follow the convention of the other types, let's lowercase `options`: `postcss$options`.

decls/postcss.js
```diff
@@ -32,3 +32,9 @@ export type postcss$rule = {
...
+  from?: ?string,
```
davidtheclark: You shouldn't need both question marks. I think just `from?: string,` should be fine.

lib/getPostcssResult.js
```diff
@@ -19,7 +20,7 @@ module.exports = function (stylelint/*: stylelint$internalApi*/)/*: Promise<?Obj
     customSyntax?: string
   }*/ = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : {}
 
-  const cached = stylelint._postcssResultCache.get(options.filePath)
+  const cached/*: ?Object*/ = stylelint._postcssResultCache.get(options.filePath)
```
davidtheclark: This should be a `postcss$result`, right?

davidtheclark: Oh, whoops ... this is the one you emailed about! You proposed a type like this, right?

```
export type postcss$result = {
  propcessor: any,
  messages: <Array>,
  root: Object,
  opts: Object,
  css: string,
  map: any,
}
```

Let's start with just this: 

```
export type postcss$result = {
  root: Object
  css: string
}
```

And see if that works. We can add more of the Result's properties as needed.

decls/postcss.js
ivanzusko: @davidtheclark you are definitely right, thank you.

lib/getPostcssResult.js
```diff
@@ -18,8 +19,14 @@ module.exports = function (stylelint/*: stylelint$internalApi*/)/*: Promise<?Obj
     syntax?: stylelint$syntaxes,
     customSyntax?: string
   }*/ = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : {}
+  // eslint-disable-next-line
+  console.log('----------------------------')
+  // eslint-disable-next-line
+  console.log('OPTIONS ==> ', options)
+  // eslint-disable-next-line
+  console.log('===============')
```
davidtheclark: Please remove these console logs (must be a mistake, right?), then we can merge.

Issue #2198, opened 2016-12-16
False positive for declarationless rules in no-duplicate-selectors


subzey:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A feature request. Add a secondary option `ignoreEmpty` to `no-duplicate-selectors`, so blocks with no content won't count.

The use case:

```less
.some-block {
  & {
    color: black;
  }

  &--error {
    color: red;
  }
}
```

This LESS code is linted as 

```css
.some-block {}
.some-block {
  color: black;
}
.some-block--error {
  color: red;
}
```
triggering a `no-duplicate-selectors` error/warning. `ignoreEmpty` would allow using such a style, yet catching the "real" errors, like this:

```less
.some-block {
  color: black;

  & { // ERROR!
    color: green;
  }
}
```

> Which rule, if any, is this issue related to?

`no-duplicate-selectors`

> What CSS is needed to reproduce this issue?

e.g.

```less
.some-block {
  & {
    color: black;
  }

  &--error {
    color: red;
  }
}
```
> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
  "rules": {
    "no-duplicate-selectors": true
  }
}
```

> Which version of stylelint are you using?

`7.6.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI, as a git hook and with SublimeLinter-contrib-stylelint

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes, it's about LESS nesting

> What did you expect to happen?

`&` within a rule without properties is not treated as a duplicate block

> What actually happened (e.g. what warnings or errors you are getting)?

`&` within a rule without properties is treated as a duplicate block

evilebottnawi:
@subzey I think it is bug, not feature, we should ignore `selectors` which contains only nesting

evilebottnawi:
@subzey it should be easy fix, add `isContainsOnlyNesting` utils (maybe best name) and skip this selectors.

jeddy3:
> it should be easy fix, add isContainsOnlyNesting utils (maybe best name) and skip this selectors.

SGTM. Let's go with`hasOnlyNestedRules` to be in line with our other `has*` utils like `hasEmptyBlock`.

hudochenkov:
This rule fails if a CSS rule has a nested rule with selector `&`. I think #2199 fails because of this too.

[...]

Issue #2199, opened 2016-12-16
False positive for non standard selectors in no-duplicate-selectors


cookch10:
When using stylelint `v7.6.0` as follows: 

package.json (simplified):

``` json
{
  "scripts": {
    "lint:styles": "stylelint --syntax scss \"src/**/*.scss\""
  }
}
```

My .stylelintrc config (uses stylelint-config-standard `v11.0.0`):

``` json
{
  "extends": "stylelint-config-standard",
  "rules": {
    "function-url-quotes": "none",
    "selector-pseudo-element-colon-notation": "single"
  }
}
```

given the following Sass (scss file): [Sassmeister gist](http://www.sassmeister.com/gist/3f1e0e4872439b81ed369b8d972506ff)

``` scss
.widget-container {
  $variation-without-image: #{"&[data-variation^=without_image]"};
  $has-buttons: #{"&[data-has-buttons]"};

  #{$variation-without-image} {
    & {
      #{$has-buttons} {
        .grid.widget__grid {
          border: 1px solid red;
        }
      }
    }

    .grid.widget__grid {
      border: 1px solid orange;
    }
  }
}
```

``` shell
npm run lint:styles
```

The following results in an error:

``` shell
:5  ×  Unexpected duplicate selector ".widget-container #{$variation-without-image}"   no-duplicate-selectors
```

There is no duplicate selector here.  If I change the Sass to the following (which yields the same output):  given the following Sass (scss file), there is no error observed: [Sassmeister gist](http://www.sassmeister.com/gist/ea0e16f3b14eaa86f7f2ba7355c0bbf2)
``` scss
.widget-container {
  $variation-without-image: #{"&[data-variation^=without_image]"};
  $has-buttons: #{"&[data-has-buttons]"};

  #{$variation-without-image} {
    #{"#{$has-buttons}"} {
      .grid.widget__grid {
        border: 1px solid red;
      }
    }

    .grid.widget__grid {
      border: 1px solid orange;
    }
  }
}
```

jeddy3:
> `:5  ×  Unexpected duplicate selector ".widget-container #{$variation-without-image}"   no-duplicate-selectors`

I think we should ignore any resolved non-standard selectors in this rule i.e. those with interpolation.

cookch10:
@jeddy3 , what are your thoughts on this one?  Is this a valid bug?

jeddy3:
Whoops, I didn't notice those thumbs up.

[...]

PR #2364, opened 2017-02-16
Add a test for a parent selector in no-duplicate-selectors

hudochenkov:

https://github.com/stylelint/stylelint/issues/2198

No, it's self explanatory.

### DIFF
--- a/lib/rules/no-duplicate-selectors/__tests__/index.js
+++ b/lib/rules/no-duplicate-selectors/__tests__/index.js
@@ -84,6 +84,12 @@ testRule(rule, {
     message: messages.rejected("a b"),
     line: 1,
     column: 12,
+  }, {
+    code: "a { & {} }",
+    description: "duplicate caused by &-parent selector",
+    message: messages.rejected("a"),
+    line: 1,
+    column: 5,
   } ],
 })
 


lib/rules/no-duplicate-selectors/__tests__/index.js
```diff
@@ -84,6 +84,12 @@ testRule(rule, {
     message: messages.rejected("a b"),
     line: 1,
     column: 12,
+  }, {
+    code: "a { & {} }",
+    description: "duplicate caused by &-parent selector",
+    message: messages.rejected("a"),
```
davidtheclark: Wait ... why is this a duplicate? Don't we end up with `a` and `a a`, which are different selectors?

hudochenkov: It will be `a {} a {}` or `a {}`. We can't predict what preprocessor will do. Example http://www.sassmeister.com/gist/8ad7e0270cf5817804625228e74f9c6b

jeddy3: In the [CSS nesting instance](https://tabatkins.github.io/specs/css-nesting/#nest-selector) it looks like `&` is syntactic sugar for `:matches()`. So, I guess 

```
a { & {} }
```

Just becomes:

```
a {}
:matches(a) {}
```

Technically different selectors, but they do the same thing.

Anyhow, I think you're [excellent write-up of the edge-cases and inconsistencies](https://github.com/stylelint/stylelint/issues/2198#issuecomment-274276106) now makes me think we should just give any rule with just `&` as its selector a wide-berth i.e. don't resolve it and don't check it. If we go down this path, should we do it in stylelint or within `postcss-resolve-nested-selector`?

hudochenkov: Worth noting that [CSS Nesting Module Level 3](https://tabatkins.github.io/specs/css-nesting/) it's just a Tab Atkin's idea. It's not related to W3C. Basically it have the same “power” as any blog post.

I think postcss-resolve-nested-selector is working well and resolves `&` well.

davidtheclark: > we should just give any rule with just & as its selector a wide-berth i.e. don't resolve it and don't check it. If we go down this path, should we do it in stylelint or within postcss-resolve-nested-selector?

I think that for all intents and purposes `a { & { color: pink; } }` includes a duplicate selector, and over in #2198 everybody seems to agree about that. So it should be flagged.

However, because of #2198 and #2199 it looks like some people want an option to *not* consider the lone `&` to be a duplicate selector. That could be a new feature request.

davidtheclark:
👍 merging!

PR #2365, opened 2017-02-16
no-browser-hacks: link to external plugin

Slamdunk:
https://github.com/stylelint/stylelint/issues/2123#issuecomment-280045108

Feedbacks on the entire [plugin](https://github.com/Slamdunk/stylelint-no-browser-hacks) are welcome.


Slamdunk:
2. At the time of writing, only one plugin out of 7 uses `plugin/` namespace, while the others are custom.
The doc states:
> Your plugin's rule name must be namespaced, e.g. `your-namespace/your-rule-name`. If your plugin provides only a single rule or you can't think of a good namespace, you can simply use `plugin/my-rule`.

If you want consistency, it's better you move to mandatory rules instead of optional ones.

Changed.

3. That's what pull requests are for

Changed.

jeddy3:
@Slamdunk Excellent work. Thanks!

Issue #2348, opened 2017-02-11
Add ignore: ["custom-elements"] option to selector-type-no-unknown


abdonrd:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

I want to propose add support to [Custom Elements](https://www.w3.org/TR/custom-elements/).

I add the [specification reference to valid Custom Elements name](https://www.w3.org/TR/custom-elements/#valid-custom-element-name).

> Which rule, if any, is this issue related to?

`selector-type-no-unknown`

> What CSS is needed to reproduce this issue?

```css
my-custom-element {}
```

> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
  "rules": {
    "selector-type-no-unknown": true
  }
}
```

> Which version of stylelint are you using?

`7.8.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI and VSCode plugin.

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No.

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

The following warnings were flagged:

```shell
Unexpected unknown type selector "my-custom-element" (selector-type-no-unknown)
```

hudochenkov:
I don't think this is a good idea. There will be nothing to check by this rule if it allows every custom element. You can just disable rule and receive same result.

davidtheclark:
@abdonrd I think this is a perfect use-case for a [plugin](https://stylelint.io/developer-guide/plugins/). If the spec gains traction, browsers start implementing and people start relying on custom elements — at that time we would reconsider inclusion of the rule in stylelint core.

abdonrd:
@hudochenkov if I disable the rule, we lose the unknown check for non custom elements. Also, maybe we can check the valid Custom Elements name.

abdonrd:
@davidtheclark I think the spec already has traction. It is landed in Chrome, Opera, Safari. And active developing in Firefox.

<img width="1038" alt="screen shot 2017-02-14 at 13 27 24" src="https://cloud.githubusercontent.com/assets/1007051/22928957/5fe68f36-f2b9-11e6-8309-3caa0a9f59f2.png">

hudochenkov:
[...]

Issue #2326, opened 2017-02-08
Add ignore: ["default-namespace"] and ignoreNamespaces: [] options to selector-type-no-unknown


bedney:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

The `selector-type-no-unknown` rule doesn't take CSS namespace into account.

> Which rule, if any, is this issue related to?

`selector-type-no-unknown`

> What CSS is needed to reproduce this issue?

e.g.

```css

@namespace todomvc url(urn:app:todomvc);

todomvc|app {
    display: block;
    position: absolute;
    height: 100%;
    width: 100%;
    margin: 0;
    padding: 0;
}
```

> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
    "extends": "stylelint-config-standard",
    "rules": {
        "selector-type-no-unknown": [true, { "severity": "warning" }]
    }
}
```

> Which version of stylelint are you using?

`7.8.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylelint **/*.css`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No. CSS namespaces are a W3C-approved syntax.

> What did you expect to happen?

"No warnings to be flagged."

> What actually happened (e.g. what warnings or errors you are getting)?

"The following warnings were flagged:"

```shell
   12:8    warn    Unexpected unknown type selector "app"                         selector-type-no-unknown
```

Note that this bug can be fixed by altering `selector-type-no-unknown/index.js` by inserting this at line 85:

```js
if (optionsMatches(options, "ignoreTypes", tagNode.namespace + '|' + tagNode.value)) {
            return
          }
```

and then by updating the `.stylelintrc` to have:

```json
        "selector-type-no-unknown": [true, { 
            "severity": "warning", 
            "ignoreTypes": ["/^.*\\|.*$/"] 
        }],
```

hudochenkov:
Thank you for the report and for using the template.

Rule is working as expected. `todomvc|app` is an `app` type selector in `todomvc` namespace. `app` a is non-standard type, that's why rule shows warning.

Use this config to add `app` as recognised type selector:

```json
{
    "rules": {
        "selector-type-no-unknown": [
            true, 
            {
                "ignoreTypes": ["app"]
             }
        ]
    }
}
```

jeddy3:
> Rule is working as expected.

I agree. 

[...]

PR #2366, opened 2017-02-16
Add ignore:["custom-element"] option to selector-type-no-unknown

evilebottnawi:

#2348

No, it's self explanatory.

Btw, i think be good add option `resolve: ['custom-element']` (maybe best name). Because `not-my-type` is valid as custom element but invalid as known type selector :sob:

### DIFF
--- a/lib/rules/selector-type-no-unknown/README.md
+++ b/lib/rules/selector-type-no-unknown/README.md
@@ -40,7 +40,27 @@ li > a {}
 
 ## Optional secondary options
 
-### `ignore: ["default-namespace"]`
+### `ignore: ["custom-elements", "default-namespace"]`
+
+#### `"custom-elements"`
+
+Allow custom elements.
+
+The following patterns are considered warnings:
+
+```css
+unknown {}
+```
+
+```css
+x-Foo {}
+```
+
+The following patterns are *not* considered warnings:
+
+```css
+x-foo {}
+```
 
 #### `"default-namespace"`
 
--- a/lib/rules/selector-type-no-unknown/index.js
+++ b/lib/rules/selector-type-no-unknown/index.js
@@ -4,12 +4,14 @@ const isKeyframeSelector = require("../../utils/isKeyframeSelector")
 const isStandardSyntaxRule = require("../../utils/isStandardSyntaxRule")
 const isStandardSyntaxSelector = require("../../utils/isStandardSyntaxSelector")
 const isStandardSyntaxTypeSelector = require("../../utils/isStandardSyntaxTypeSelector")
+const isCustomElement = require("../../utils/isCustomElement")
 const optionsMatches = require("../../utils/optionsMatches")
 const parseSelector = require("../../utils/parseSelector")
 const report = require("../../utils/report")
 const ruleMessages = require("../../utils/ruleMessages")
 const validateOptions = require("../../utils/validateOptions")
 const htmlTags = require("html-tags")
+const keywordSets = require("../../reference/keywordSets")
 const _ = require("lodash")
 const svgTags = require("svg-tags")
 const mathMLTags = require("mathml-tag-names")
@@ -20,38 +22,13 @@ const messages = ruleMessages(ruleName, {
   rejected: selector => `Unexpected unknown type selector "${selector}"`,
 })
 
-// htmlTags includes only "standard" tags. So we augment it with older tags etc.
-const nonStandardHtmlTags = new Set([
-  "acronym",
-  "applet",
-  "basefont",
-  "big",
-  "blink",
-  "center",
-  "content",
-  "dir",
-  "font",
-  "frame",
-  "frameset",
-  "hgroup",
-  "isindex",
-  "keygen",
-  "listing",
-  "marquee",
-  "noembed",
-  "plaintext",
-  "spacer",
-  "strike",
-  "tt",
-  "xmp",
-])
-
 const rule = function (actual, options) {
   return (root, result) => {
     const validOptions = validateOptions(result, ruleName, { actual }, {
       actual: options,
       possible: {
         ignore: [
+          "custom-elements",
           "default-namespace",
         ],
         ignoreNamespaces: [_.isString],
@@ -84,6 +61,13 @@ const rule = function (actual, options) {
           }
 
           if (
+            optionsMatches(options, "ignore", "custom-elements")
+            && isCustomElement(tagNode.value)
+          ) {
+            return
+          }
+
+          if (
             optionsMatches(options, "ignore", "default-namespace")
             && !tagNode.hasOwnProperty("namespace")
           ) {
@@ -103,7 +87,7 @@ const rule = function (actual, options) {
 
           if (htmlTags.indexOf(tagNameLowerCase) !== -1
             || svgTags.indexOf(tagNameLowerCase) !== -1
-            || nonStandardHtmlTags.has(tagNameLowerCase)
+            || keywordSets.nonStandardHtmlTags.has(tagNameLowerCase)
             || mathMLTags.indexOf(tagNameLowerCase) !== -1
           ) {
             return
--- a/lib/utils/isCustomElement.js
+++ b/lib/utils/isCustomElement.js
@@ -1,0 +1,40 @@
+/* @flow */
+"use strict"
+
+const svgTags = require("svg-tags")
+const htmlTags = require("html-tags")
+const keywordSets = require("../reference/keywordSets")
+const mathMLTags = require("mathml-tag-names")
+
+/**
+ * Check whether a type selector is a custom element
+ */
+module.exports = function (selector/*: string*/)/*: boolean*/ {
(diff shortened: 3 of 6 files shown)


jeddy3:
Can you update [the docs](https://github.com/stylelint/stylelint/tree/master/lib/rules/selector-type-no-unknown#optional-secondary-options):

> This rule considers tags defined in the HTML and SVG Specifications to be known.

Becomes:

This rule considers tags which are [valid custom elements](https://www.w3.org/TR/custom-elements/#valid-custom-element-name) and those defined in the following Specifications to be known: HTML and SVG.

(We'll [likely be adding](https://github.com/stylelint/stylelint/issues/2326#issuecomment-278416740) MathML to that list).

Can you also update the [examples](https://github.com/stylelint/stylelint/tree/master/lib/rules/selector-type-no-unknown#ignoretypes-regex-string) not to use hyphens.

Lastly, can you remove the hyphen from [the regex tests](https://github.com/stylelint/stylelint/blob/80ed0c7200047bcd47476e7a56e74330c0c34ea4/lib/rules/selector-type-no-unknown/__tests__/index.js#L173-L198) to make them relevant again.

> Btw, i think be good add option resolve: ['custom-element'] (maybe best name). Because not-my-type is valid as custom element but invalid as known type selector

Can you reword this please? It's not clear what you mean.

evilebottnawi:
@jeddy3 if we merge this PR type selector `foo-bar` be valid type selector (because it is valid as custom selector), if we want avoid this situation we should add option `resolve: ["custom-selector"]` (maybe best), if option enable `foo-bar` is valid as type selector, if option disable `foo-bar` is invalid as type selector, sorry for my bad english

jeddy3:
Oh I see. You're suggesting we put this behind an option e.g. `ignore: ["custom-elements"]`?

>  if option enable foo-bar is valid as type selector, if option disable foo-bar is invalid as type selector

That would address this issue.

evilebottnawi:
@jeddy3 yep, it will be good solution

davidtheclark:
I'm wondering if this option should be reversed. Instead of by default complaining about custom elements, it seems we should by default let them pass. Maybe we'd add a `disallowCustomElements` option to get the reverse.

(It's probably unlikely that a legitimate custom element is going to be confused with a type selector typo, because custom elements require hyphens and almost no real type selectors do.)

jeddy3:
> I'm wondering if this option should be reversed. Instead of by default complaining about custom elements, it seems we should by default let them pass.

I agree.

> Maybe we'd add a disallowCustomElements option to get the reverse.

Yep, but I think we can wait until someone actually runs into the scenario and requests it (as, like you said, it seems unlikely to occur).

So, I think this PR is ready to merge once @evilebottnawi addresses the [doc and test requests](https://github.com/stylelint/stylelint/pull/2366#issuecomment-280470173).

evilebottnawi:
@jeddy3 some my project use custom elements some no, i am add this option :smile: , i requests this :smile:

jeddy3:
> some my project use custom elements some no, i am add this option 😄 , i requests this

Sounds good. Let's create a separate issue for that and add that option in a separate PR. Let's handle the [doc and test requests](https://github.com/stylelint/stylelint/pull/2366#issuecomment-280470173) and merge this.

davidtheclark:
@evilebottnawi Do you plan to finish this PR, or should we close it?

evilebottnawi:
@davidtheclark finish on this week :smile:

lib/utils/isCustomElement.js
```diff
@@ -0,0 +1,27 @@
...
+const svgTags = require("svg-tags")
+const htmlTags = require("html-tags")
+const keywordSets = require("../reference/keywordSets")
+const mathmlTagNames = require("mathml-tag-names")
+
+/**
+ * Check whether a type selector is a custom element
+ */
+module.exports = function (selector/*: string*/)/*: boolean*/ {
+  const startWithALowercaseASCII = (/^[a-z]/m).test(selector)
```
davidtheclark: Why the `m` flag?

lib/utils/isCustomElement.js
```diff
@@ -0,0 +1,27 @@
...
+  const isNotHtmlTag = htmlTags.indexOf(selector.toLowerCase()) === -1
+    && !keywordSets.nonStandardHtmlTags.has(selector.toLowerCase())
+  const isNotMathmlTag = mathmlTagNames.indexOf(selector.toLowerCase()) === -1
+
+  return startWithALowercaseASCII
+    && notContainAnyUppercaseASCII
+    && containAHyphen
+    && isNotSvgTag
+    && isNotHtmlTag
+    && isNotMathmlTag
```
davidtheclark: This is kind of a weird way to structure the function. Couldn't you have a return after each statement and avoid the need to assign variables? E.g.

```js
if (!(/^[a-z]/m).test(selector)) return false
if (selector.toLowerCase() !== selector) return false
...
```

jeddy3:
> I'm wondering if this option should be reversed. Instead of by default complaining about custom elements, it seems we should by default let them pass. Maybe we'd add a `disallowCustomElements` option to get the reverse.

I just realised, while thinking through https://github.com/kristerkari/stylelint-scss/issues/103#issuecomment-287649576, that perhaps this behaviour should be behind an option. Sorry to flip-flop.

I believe _every_ `no` rule in stylelint is most restrictive by default and can be made more permissive using `"ignore": []` options. We even changed `indentation` to be consistent with this behaviour a while back, but checking function parameters _by default_ rather than via opt-in. 

I think adding a `disallow...` boolean option would be at odds with what we've done elsewhere... I think it's also the reason https://github.com/stylelint/stylelint/issues/2437 confused me - that is the issue for this `disallow...` option, right? But it looks like @evilebottnawi instinctively, and understandably so, reach for an `ignore*` option. 

Putting this behind an option also fits with the idea of "known" being a discrete list of elements from specifications, whereas `ignore: ["custom-elements"]` is triggering a _heuristic_ ignore. 

Want do you think?

evilebottnawi:
@jeddy3 sound good for option :+1:

jeddy3:
> sound good for option

@evilebottnawi Is this change something you can factor into this PR?

evilebottnawi:
@jeddy3 yep, in near future

lib/utils/isCustomElement.js
evilebottnawi: @davidtheclark yep, remove it

lib/utils/isCustomElement.js
evilebottnawi: @davidtheclark done

lib/rules/selector-type-no-unknown/index.js
```diff
@@ -97,10 +74,20 @@ const rule = function (actual, options) {
             return
           }
 
+          if (
```
jeddy3: Minor request... can we get this shifted to [line 61](https://github.com/stylelint/stylelint/blob/fff628f10cabf424d18c79750c7e5f71a819ad50/lib/rules/selector-type-no-unknown/index.js#L61) so the early returns are in the same alphabetical order as the options config.

Quite a lot of rules now have 5 or more `ignore/except*` options and I think there's no harm adding a bit of consistent structure to them.

lib/rules/selector-type-no-unknown/index.js
```diff
@@ -19,39 +21,14 @@ const messages = ruleMessages(ruleName, {
...
-])
-
 const rule = function (actual, options) {
   return (root, result) => {
     const validOptions = validateOptions(result, ruleName, { actual }, {
       actual: options,
       possible: {
         ignore: [
           "default-namespace",
+          "custom-elements",
```
jeddy3: Alphabetical order please.

lib/rules/selector-type-no-unknown/README.md
```diff
@@ -8,7 +8,7 @@ Disallow unknown type selectors.
  * This type selector */
 ```
 
-This rule considers tags defined in the HTML and SVG Specifications to be known.
+This rule considers tags which are valid custom elements and those defined in the following Specifications to be known: HTML and SVG.
```
jeddy3: This rule considers tags which are defined in the following Specifications to be known: HTML and SVG.

awebdeveloper:
is this merged. Can a new tag be created

evilebottnawi:
@awebdeveloper wait after @davidtheclark accept.

awebdeveloper:
@evilebottnawi just to clarify will this work for all the below cases

```
1. my-element
2. element
3. [my-attr]
```

davidtheclark:
Added to changelog:

> -   Fixed: `selector-type-no-unknown` now ignores valid custom elements ([#2366](https://github.com/stylelint/stylelint/pull/2366)).

abdonrd:
Thanks! 👏  🎉

awebdeveloper:
looking at the docs this doesn't take into account custom attributes. **[my-attr]**. Maybe a different PR but that should also be included

evilebottnawi:
@awebdeveloper please create issue with template for this

PR #2367, opened 2017-02-17
Mandatory plugin/ namespace for plugins

Slamdunk:
https://github.com/stylelint/stylelint/pull/2365#pullrequestreview-22281850


davidtheclark:
Thanks for the PR, @Slamdunk, but I don't think we want to enforce that the namespace is always `plugin/`. I'd prefer to leave it as-is, with `plugin/` as a suggestion if you don't have another plugin namespace in mind.

PR #2368, opened 2017-02-20
[docs] Correct config lookup in node API

jesstelford:
Brings into line with reality.

No, it's self explanatory.

### DIFF
--- a/docs/user-guide/node-api.md
+++ b/docs/user-guide/node-api.md
@@ -35,7 +35,7 @@ This can be useful, for example, when making a text editor plugin that passes in
 
 A [stylelint configuration object](configuration.md).
 
-If no `config` or `configFile` is passed, stylelint will use a [config lookup algorithm](https://stylelint.io/user-guide/configuration/#loading-the-configuration-object) to find the correct config.
+If no `config` or `configFile` is passed, stylelint will use a [config lookup algorithm](./configuration/#loading-the-configuration-object) to find the correct config.
 
 ### `configBasedir`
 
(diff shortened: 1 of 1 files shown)


docs/user-guide/node-api.md
```diff
@@ -35,7 +35,7 @@ This can be useful, for example, when making a text editor plugin that passes in
 
 A [stylelint configuration object](configuration.md).
 
-If no `config` or `configFile` is passed, stylelint will look for a `.stylelintrc` configuration file.
+If no `config` or `configFile` is passed, stylelint will use a [config lookup algorithm](https://stylelint.io/user-guide/configuration/#loading-the-configuration-object) to find the correct config.
```
jeddy3: Please use a relative path here, so that the link works within the docs of the repo i.e. `./configuration.md#loading-the-configuration-object`. This change means the link will also be linted by `remark-lint`, ensuring it is always valid.

jesstelford:
@jeddy3 Good catch - I've fixed that URI 👍

jeddy3:
@jesstelford Thanks for making the change. Looks like `remark-lint` [did its job](https://travis-ci.org/stylelint/stylelint/jobs/205648594) and caught the missing file extension.

Could you possibly make one more change and correct that (changing the path to `configuration.md/#loading-the-configuration-object`, and then we can merge this in.

FYI, you can run `npm run lint` locally to check everything is dandy with the path and anchor.

jesstelford:
Whoops, not sure what I did there to make that mistake. All fixed.

Love that the linter has a linter!

davidtheclark:
Thanks @jesstelford!

Issue #2343, opened 2017-02-11
Incorrect line in max-empty-lines


b8kkyn:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Bugs, opinions and one idea.

> Which rule, if any, is this issue related to?

`max-empty-lines`

> What CSS is needed to reproduce this issue?

```css





/*
  The above should be warning because my config is "max-empty-lines": 2 
  And there is 3 empty lines and no warning, how?
  Because max-empty-lines doesn't consider beginning of the file.
*/

.selector {
  color: #fff;
}

/**
 * Call me Ishmael.
 *
 *
 *
 * Some years ago -- never mind how log precisely -- ...
 *
 */
/* 
 * Lets look at this. Rule's documentation says this comment block has empty lines and should be warning.
 * Again there is no warning. But in my opinion, shouldn't be either.
 * Because comment block has not actually empty lines, right?
 * Please pay attention to the stars on the left.
 */
 
#samerule {
  background-image: url('funnybugishere.png');
}

/*






 40th Line: Now we have empty lines in comment block. Following lines are considered warning: 37, 38, 39 and 40
 Which are totaly weird. 40th line is not empty and you can see that by looking previous line.
 And again in my opinion these lines should be warning for this comment block: 36, 37, 38 and 39
 By letting only 2 adjacent empty lines ("max-empty-lines": 2)
*/
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "max-empty-lines": 2
  }
}
```

> Which version of stylelint are you using?

`7.8.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

Doesn't matter.

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No.

> What did you expect to happen?

I expected this:
```shell
something.css
  3:1  ×  Expected no more than 2 empty line(s)   max-empty-lines
 36:1  ×  Expected no more than 2 empty line(s)   max-empty-lines
 37:1  ×  Expected no more than 2 empty line(s)   max-empty-lines
 38:1  ×  Expected no more than 2 empty line(s)   max-empty-lines
 39:1  ×  Expected no more than 2 empty line(s)   max-empty-lines
```

> What actually happened (e.g. what warnings or errors you are getting)?

The following warnings were flagged:

```shell
something.css
 37:1  ×  Expected no more than 2 empty line(s)   max-empty-lines
[...]

PR #2371, opened 2017-02-20
Fix wrong line count in max-empty-lines

GorwinJororis:

#2343

No, it's self explanatory.

### DIFF
--- a/lib/rules/max-empty-lines/index.js
+++ b/lib/rules/max-empty-lines/index.js
@@ -13,8 +13,6 @@ const messages = ruleMessages(ruleName, {
 })
 
 const rule = function (max) {
-  const maxAdjacentNewlines = max + 1
-
   return (root, result) => {
     const validOptions = validateOptions(result, ruleName, {
       actual: max,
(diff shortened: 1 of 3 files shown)


lib/rules/max-empty-lines/index.js
```diff
@@ -23,47 +21,28 @@ const rule = function (max) {
     if (!validOptions) {
       return
     }
-
+    
+    // `//`-comments are need to be processed by postcss-scss
+    // To do that we have to use .toString()
     const rootString = root.toString()
```
evilebottnawi: Try to use `root.source.input.css`, because we can have potential incorrect position on non standard syntax.

Leucetios: In fact, without `root.toString()`, we have potential incorrect position on non standard syntax (//-comments). Because output of `root.source.input.css` isn't same as `root.toString()` when you used non standard syntax. (PostCSS-SCSS converts `// single line comment` to `/* single line comment*/`)

evilebottnawi: @Leucetios yep, but you use `index += (lines[i] + separator).length` for detect index and it is wrong on scss comments

evilebottnawi: `// single line comment` and `/* single line comment */` - should report on difference index, but your code seems report on equals, btw `root.source.input.css` is good practice, another syntax can have other comments directive.

Leucetios: I guess I couldn't tell properly. Point is if I use `root.source.input.css`, line counts will be wrong on scss comments. And to fix, need to do extra work.

evilebottnawi: @Leucetios yes, need do this

davidtheclark:
@evilebottnawi Do you feel good about these changes? Is this ready for another review?

evilebottnawi:
@davidtheclark this change in rule report invalid column on non standard syntax, not merge.
/cc @GorwinJororis

PR #2382, opened 2017-02-23
Add defaultIgnores: false option

hudochenkov:

#2236

Changing default behaviour for `ignoreFiles`. User's config will override default values.

Should we somewhere in docs write default pattern? Users might want to add some patterns additionally to default patterns.

Also, what to do to [this code](https://github.com/stylelint/stylelint/blob/c6c5a79cea760e65a6e8a3fb651c84f8d79afc15/lib/standalone.js#L69-L75) which was added recently (https://github.com/stylelint/stylelint/commit/9578c4871fba233905e9bc66e803bddb6996362f).

### DIFF
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -286,7 +286,7 @@ If the globs are absolute paths, they are used as is. If they are relative, they
 -   the config's filepath, if the config is a file that stylelint found a loaded;
 -   or `process.cwd()`.
 
-`node_modules` and `bower_components` are always ignored.
+By default, all `node_modules` and `bower_components` are ignored. Default values will be overridden if `ignoreFiles` is set.
 
 The `ignoreFiles` property is stripped from extended configs: only the root-level config can ignore files.
 
@@ -307,4 +307,4 @@ The patterns in your `.stylelintignore` file must match [`.gitignore` syntax](ht
 
 stylelint will look for a `.stylelintignore` file in `process.cwd()`. You can also specify a path to your ignore patterns file (absolute or relative to `process.cwd()`) using the `--ignore-path` (in the CLI) and `ignorePath` (in JS) options.
 
-`node_modules` and `bower_components` are always ignored.
+By default, all `node_modules` and `bower_components` are ignored. Default values will be overridden by `.stylelintignore`.
--- a/docs/user-guide/node-api.md
+++ b/docs/user-guide/node-api.md
@@ -61,7 +61,7 @@ A file glob, or array of file globs. Ultimately passed to [node-glob](https://gi
 
 Relative globs are considered relative to `process.cwd()`.
 
-`node_modules` and `bower_components` are always ignored.
+By default, all `node_modules` and `bower_components` are ignored.
 
 ### `formatter`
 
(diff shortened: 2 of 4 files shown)


davidtheclark:
I think there's a significant problem here: https://github.com/stylelint/stylelint/issues/2399#issuecomment-282603117. The gist of it is that I can't think of an efficient way for `ignoreFiles` to work on large quantities of files that match the original input glob. A file will only be ignored via `ignoreFiles` *after* stylelint has loaded its relevant config and checked that config for `ignoreFiles` — which means using `ignoreFiles` to ignore `node_modules` will be significantly less efficient than the current implement @hudochenkov linked to.

hudochenkov:
On the other side [this code](https://github.com/stylelint/stylelint/blob/c6c5a79cea760e65a6e8a3fb651c84f8d79afc15/lib/standalone.js#L69-L75) made impossible #2236. It ignores files which might not be ignored if `ignoreFiles` are specified.

davidtheclark:
Uuuggggh.

Here's the only decent solution I can come up with for this mess:

- `config.ignoreFiles` comes with the caveat that it is not an efficient way to ignore a lot of files, because it is included in configs and configs are loaded for each file before they can be analyzed — so we don't know if the file is ignored until we've loaded its config.
- If you want to ignore lots of files efficiently, you need to do that in your input glob.
- `node_modules` and `bower_components` are by default ignored *via the input glob* (so, efficiently); and we can add a new option like `defaultIgnores: false` that turns this behavior off.

hudochenkov:
Sorry, I don't have enough knowledge about ignoring files, and stylelint's architecture. I can't have strong opinion on that matter. I believe other team members could have a solution.

TheXardas:
Gentlemen,
maybe a simple flag like "disableNodeModulesIgnore" would do?

jeddy3:
> Here's the only decent solution I can come up with for this mess:

SGTM. Let's document the limitation of `config.ignoreFiles` and how a more tailored input glob can be used as an alternative, and let's add a `defaultIgnores: false` option.

Am I right in thinking that such an option would no longer mean that this is a breaking change and we can point this PR (once it's adapted) to `master`?

> maybe a simple flag like "disableNodeModulesIgnore" would do?

@TheXardas Thanks for chiming in. Yep, that is part of @davidtheclark's [proposal](https://github.com/stylelint/stylelint/pull/2382#issuecomment-284730139).

davidtheclark:
@jeddy3, I think I've realized a key here (https://github.com/stylelint/stylelint/issues/2399#issuecomment-290928039): You can already load only one `.stylelintignore` — so we should change our strategy a bit so those patterns get incorporated into the glob. `ignoreFiles` will be the inefficient way to ignore things; `.stylelintignore` or your files glob the efficient way.

This will be a breaking change, though the break probably will not affect most users: the `ignored` and `standaloneIgnored` properties of stylelint results will no longer be the same, and files that are ignored via `.stylelintignore` will not show up in the results at all.

I'll try to prepare a PR to `v8` today.

jeddy3:
> I think I've realized a key here (#2399 (comment)): 

I saw that. SGTM.

> I'll try to prepare a PR to v8 today.

Awesome! :)

jeddy3:
Closing in favour of #2464

PR #2391, opened 2017-02-24
chore: Update `remark-*` modules

ntwb:
Ref: https://github.com/stylelint/stylelint/pull/2383#issuecomment-282235550


wooorm:
I’ll reply here to https://github.com/stylelint/stylelint/pull/2383#issuecomment-282235550.

Basically, unified changed. That means remark had to change too, and so do almost all plugins.
There’s no changes to plugins, other than “update for remark@7.0.0”. You can view the change log for remark on its [releases](https://github.com/wooorm/remark/releases) page.

Another problem is presets, they didn’t work on the API, now they do.

You’re using presets as follows:

```json
  "remarkConfig": {
    "presets": [
      "lint-recommended",
      "lint-consistent"
    ],
    "plugins": {
      "validate-links": {
        "repository": "stylelint/stylelint"
      }
    }
  }
```

...but with the new ecosystem you’ll need the following instead:

```json
  "remarkConfig": {
    "plugins": [
      "preset-lint-recommended",
      "preset-lint-consistent",
      ["validate-links", {"repository": "stylelint/stylelint"}]
    ]
  }
```

> @wooorm Any chance you could add changelogs in the future to the remark-* modules please, there's nothing documented that I can see at a quick glance, I need to start comparing versions of these modules to determine the cause 😄

The problem is that there’s a breaking change in the ecosystem, meaning I get 400 greenkeeper PRs, and of course I try my best to handle that as fast as possible. I’d rather have the system stable first though, so you can upgrade your four dependencies, instead of having two with nice changelogs, and two broken plugins.
I get that it sucks, I’m doing my best!

For now, the remark [releases](https://github.com/wooorm/remark/releases) page links to the important changes.

ntwb:
Thanks @wooorm, I know you've got hundreds of modules, and are always incredibly responsive to any ping requests we make, so thank you once again for both these modules and the time to reply with the above info, it's greatly appreciated 😄 

I'll update this PR now with those changes

wooorm:
Cheers, thanks a lot, I appreciate it!
Don’t mind me, I’m just a bit grumpy. Too many modules! 😱

jeddy3:
@ntwb Thanks for handling this. @wooorm As @ntwb said, thanks for your responsiveness (and your continued work on remark-lint - it's great!) :)

wooorm:
Cheers, thanks folks! ❤️

PR #2393, opened 2017-02-24
Update stylelint-disable-reason changelog entry

jeddy3:
Ref: https://github.com/stylelint/stylelint/issues/2292#issuecomment-282379774


davidtheclark:
Thanks @jeddy3.

Issue #2395, opened 2017-02-25
False positives for custom property sets followed by a comment in no-extra-semicolons


alienlebarge:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

bug

> Which rule, if any, is this issue related to?

`no-extra-semicolons`

> What CSS is needed to reproduce this issue?

```css
/* ========================================================================
   #TYPOGRAPHY
   ======================================================================== */

:root {
    /**
     * Mor readable bodytext
     * see http://practice.typekit.com/lesson/caring-about-opentype-features/
     */
    --font-bodytext: {
        font-variant-ligatures: common-ligatures, contextual;
        font-variant-numeric: oldstyle-nums, proportional-nums;
        font-feature-settings: "kern", "liga", "clig", "calt", "onum", "pnum";
        font-kerning: normal;
    }; /* <--- get an error */





    /**
     * Mor readable headings
     * see http://practice.typekit.com/lesson/caring-about-opentype-features/
     */
    --font-headings: {
        font-variant-ligatures: common-ligatures, contextual,
            discretionary-ligatures;
        font-feature-settings: "kern", "liga", "clig", "calt", "dlig";
        font-kerning: normal;
    }; /* <--- get an error */





    /**
     * More readable sub text
     * see http://practice.typekit.com/lesson/caring-about-opentype-features/
     */
    --font-sub: {
        font-variant-ligatures: common-ligatures, contextual;
        font-variant-position: sub;
        font-feature-settings: "kern", "liga", "clig", "calt", "subs";
        font-kerning: normal;
    }; /* <--- get an error */





    /**
     * OpenType feature for small caps
     * see http://practice.typekit.com/lesson/caring-about-opentype-features/
     *
     * 1. Safari iOs doesn't support `font-feature` so we use `text-transform`
     *    to make the texte uppercase. http://caniuse.com/#search=font-feature
     */
    --font-small-caps: {
        font-variant-ligatures: common-ligatures, contextual;
        font-variant-numeric: oldstyle-nums, proportional-nums;
        font-variant-caps: all-small-caps;
        font-feature-settings: "kern", "liga", "clig", "calt", "onum", "pnum",
            "smcp", "c2sc";
        font-kerning: normal;
        text-transform: uppercase; /* [1] */
    }; /* <--- get an error */





    /**
     * OpenType feature for data
     * see http://practice.typekit.com/lesson/caring-about-opentype-features/
     */
    --font-data: {
[...]

PR #2396, opened 2017-02-25
Fix custom properties sets followed by comments in no-extra-semicolons

hudochenkov:

#2395

Just failing tests. Have no time to look into the problem yet.

### DIFF
--- a/lib/rules/no-extra-semicolons/index.js
+++ b/lib/rules/no-extra-semicolons/index.js
@@ -14,7 +14,7 @@ const messages = ruleMessages(ruleName, {
 })
 
 function getOffsetByNode(node) {
-  const string = node.root().toString()
+  const string = node.root().source.input.css
   const nodeColumn = node.source.start.column
   const nodeLine = node.source.start.line
   let line = 1
@@ -54,14 +54,31 @@ const rule = function (actual) {
     }
 
     root.walk(node => {
-      const rawBeforeNode = node.raws.before
+      let rawBeforeNode = node.raws.before
 
       if (rawBeforeNode && rawBeforeNode.trim().length !== 0) {
         let allowedSemi = 0
 
         // Forbid semicolon before first custom properties sets
-        if (isCustomPropertySet(node) && node.parent.index(node) > 0) {
+        if ((isCustomPropertySet(node) && node.parent.index(node) > 0)) {
           allowedSemi = 1
+        }
+
+        const next = node.next()
+
+        // Ignore semicolon before comment if next node is custom properties sets or comment
+        if (node.type === "comment" && next
+          && (isCustomPropertySet(next) && node.parent.index(next) > 0 || next.type === "comment")
+        ) {
+          allowedSemi = 1
+        }
+
+        const prev = node.prev()
+
+        // Adding previous node string to custom properties set if previous node is comment
+        if (isCustomPropertySet(node) && node.parent.index(node) > 0 && prev && prev.type === "comment") {
+          rawBeforeNode = prev.toString() + rawBeforeNode
+          allowedSemi = 0
         }
 
         styleSearch({ source: rawBeforeNode, target: ";" }, (match, count) => {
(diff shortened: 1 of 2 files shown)


jeddy3:
Tests look good, thanks. Labelling as `revision needed`.

hudochenkov:
I tried to find a solution, but failed.

evilebottnawi:
@hudochenkov i am champion this rule, i can see this late, it is difficult rule :smile: 
How about blur all comments? Maybe it is help?

evilebottnawi:
it is break my mind, we support custom properties sets which not even working draft, also `postcss` parses them very badly, i think there is always a solution we just need to more think. But i hope the next verision `postcss` help us and change `raws` for `custom-*` (and other stuff).

/cc @ai what we can hear about the future version and `custom` stuff?

ai:
You talking about adding `;` to rules `raws` in this cases:

```css
--font-bodytext: {
    font-variant-ligatures: common-ligatures, contextual;
    font-variant-numeric: oldstyle-nums, proportional-nums;
    font-feature-settings: "kern", "liga", "clig", "calt", "onum", "pnum";
    font-kerning: normal;
};
```

I am waiting some spec stabilization and we can add it in 6.0

evilebottnawi:
@ai also about sets
```css
:root { 
  --foo: { color: red };
  --foo-bar: { color: red };
}
```

evilebottnawi:
We use `*.toString()` it is potential incorrect position in non standard syntax, need create issue and change all `*.toString()` to `*.source.input.css` in all rules before release `8.0.0`.

lib/rules/no-extra-semicolons/index.js
```diff
@@ -54,16 +54,29 @@ const rule = function (actual) {
...
+          || (node.type === "comment" && next
+            && (isCustomPropertySet(next) && node.parent.index(next) > 0 || next.type === "comment"))
+        ) {
           allowedSemi = 1
         }
 
+        const prev = node.prev()
+
+        if (isCustomPropertySet(node) && node.parent.index(node) > 0 && prev && prev.type === "comment") {
+          rawBeforeNode = prev.toString() + rawBeforeNode
```
hudochenkov: Should it be `prev.source.input.css`? :)

lib/rules/no-extra-semicolons/index.js
```diff
@@ -54,16 +54,29 @@ const rule = function (actual) {
...
 
+        const next = node.next()
+
         // Forbid semicolon before first custom properties sets
-        if (isCustomPropertySet(node) && node.parent.index(node) > 0) {
+        // Also ignore comment before custom properties sets
+        if ((isCustomPropertySet(node) && node.parent.index(node) > 0)
+          || (node.type === "comment" && next
+            && (isCustomPropertySet(next) && node.parent.index(next) > 0 || next.type === "comment"))
+        ) {
```
hudochenkov: What do you think about splitting this `if` statement into `else if` or more `if` statements?

lib/rules/no-extra-semicolons/index.js
evilebottnawi: @hudochenkov i do this in another PR :smile: https://github.com/stylelint/stylelint/pull/2396#issuecomment-283448628

lib/rules/no-extra-semicolons/index.js
evilebottnawi: @hudochenkov it's possible, but why :smile:

hudochenkov: Too complicated. Hard to understand in current form.

lib/rules/no-extra-semicolons/index.js
```diff
@@ -60,7 +60,7 @@ const rule = function (actual) {
         let allowedSemi = 0
 
         // Forbid semicolon before first custom properties sets
-        if ((isCustomPropertySet(node) && node.parent.index(node) > 0)) {
+        if (isCustomPropertySet(node) && node.parent.index(node) > 0) {
```
evilebottnawi: Thanks :smile:

hudochenkov:
LGTM! Thanks!

davidtheclark:
Added to Changelog:

> -   Fixed: `no-extra-semicolons` false positives for comments are custom property sets ([#2396](https://github.com/stylelint/stylelint/pull/2396)).

PR #2400, opened 2017-02-27
Flow fixes

### LINKED ISSUE #2322: Flow comments are not stripped from files in `lib`

Bug.

It seems that flowtype comments are not being stripped correctly, and are ending up in files in `lib` (which was causing my own app's flow checks to fail).  It was good, because now I realized I was checking `node_modules` which I shouldn't be doing, but I think you should also be stripping the types from the files that are distributed through npm, right?

core

`7.8.0`

There should be no flow errors from `node_modules/stylelint/lib`

I had errors such as:

```
node_modules/stylelint/lib/utils/isStandardSyntaxAtRule.js:10
 10: module.exports = function (atRule/*: postcss$atRule*/)/*: boolean*/ {
                                          ^^^^^^^^^^^^^^ identifier `postcss$atRule`. Could not resolve name
```

davidtheclark:
Closes #2322.

- Adds `decls` to the package files (and `CHANGELOG.md`, while I was at it, since that would be documentation offline).
- Creates a `dynamicRequire` module to alleviate the need for Flow's `module.ignore_non_literal_requires` option.


jeddy3:
Looking good.

`CHANGELOG.md` is [always included, regardless of settings](https://docs.npmjs.com/files/package.json#files). I don't think there's any harm in being explicit, though? Perhaps we should also add `package.json`, `README.md` and `LICENSE` to the array too? Or just revert this bit of the PR. FYI, `CHANGELOG.md` and `README.md` are needed in the package for the generation of the website, [here](https://stylelint.io/CHANGELOG/) and [here](https://stylelint.io/) respectively.

Do we also want to move the `flow-typed` decls in this PR, as mentioned in https://github.com/stylelint/stylelint/issues/2322#issuecomment-282600362?

Finally, shall we release `7.10.0` after this in? I can do it this evening or tomorrow. FYI, I'm going to be out of the country and off-line (with the possible exception of the two Saturdays) for two weeks from EOP tomorrow. There is a tonne of issues and PRs I've only scraped the surface of, autofixing etc, but I've earmarked a bit of time when I'm back to get my head around these.

davidtheclark:
Added to Changelog:

> -   Fixed: `decl/` folder of Flow types is shipped with the package, for consumers that use Flow ([#2322](https://github.com/stylelint/stylelint/issues/2322)).

PR #2403, opened 2017-03-01
WiP - @deprecated support

FGRibreau:
### Usage

In `_variables.less`:

```less
// @deprecated use @grayTrout instead
@grayLight:             #555B64;
```

In another file:

```less
.a{
  color: @grayLight;
}
```

Output:

capture d ecran 2017-03-01 10 38 34

### Current open questions

- Should we limit this rule to variables?
- Should we make it SaSS/Less agnostic and also support [Css Variables](https://developer.mozilla.org/fr/docs/Web/CSS/Les_variables_CSS)) ?

> Applicable to standard CSS syntax only.
> Has a singular purpose.

So it's not a rule but a plugin. But then a developer should create a plugin for each type of variable (Sass, Less and Css variables?) or the SRP only applies to rules?

### Implementation questions

- How to define a rule that first extract deprecation policies over every files and then loop through decls for each files? Current implementation hope that a variable file exist, was parsed first and then use a global variable to maintain declaration policies between file parsing.
- Use walkComments instead of walkDecls for the first iteration?
- It it possible to keep this rule outside of the main repository?

### Todo

Once the previous questions are answered and implementation is stabilized, documentation and tests will come.

### DIFF
--- a/lib/rules/deprecated-variable-comments/index.js
+++ b/lib/rules/deprecated-variable-comments/index.js
@@ -35,9 +35,9 @@ const rule = function(isActivated) {
     root.walkDecls(decl => {
       const prop = decl.prop;
 
-      if (prop.startsWith('@')) { // currently this rules only match for Less @variables properties, why not generalize this?
+      if (prop.startsWith("@")) { // currently this rules only match for Less @variables properties, why not generalize this?
           const parentNode = decl.parent.nodes[decl.parent.nodes.indexOf(decl) - 1]
-          if (_.get(parentNode, 'text', '').startsWith('@deprecated')) {
+          if (_.get(parentNode, "text", ").startsWith("@deprecated")) {
             deprecatedVariables[prop] = {
               comment: parentNode.text
             }
--- a/lib/rules/deprecated-variable-comments/index.js
+++ b/lib/rules/deprecated-variable-comments/index.js
@@ -28,12 +28,12 @@ const rule = function(isActivated) {
     }
 
     if(!isActivated){
-      return;
+      return
     }
 
     // @todo maybe use ``walkComments` instead and lookup @deprecated comments first
     root.walkDecls(decl => {
-      const prop = decl.prop;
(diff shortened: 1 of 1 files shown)


lib/rules/deprecated-variable-comments/index.js
```diff
@@ -0,0 +1,65 @@
...
+      return;
+    }
+
+    // @todo maybe use ``walkComments` instead and lookup @deprecated comments first
+    root.walkDecls(decl => {
+      const prop = decl.prop;
+
+      if (prop.startsWith('@')) { // currently this rules only match for Less @variables properties, why not generalize this?
+          const parentNode = decl.parent.nodes[decl.parent.nodes.indexOf(decl) - 1]
+          if (_.get(parentNode, 'text', '').startsWith('@deprecated')) {
```
Orbmancer: you mix both single and double quotes in your code, I suppose the philosophy in this repo is double quotes or ES6 backticks right ?

hudochenkov:
Thank you for your contribution! You've ignored everything in [Creating a new rule](https://stylelint.io/developer-guide/rules/#creating-a-new-rule). I believe this PR will be closed and you should [create a plugin](https://stylelint.io/developer-guide/plugins/) instead.

Let's wait for others maintainers opinion.

FGRibreau:
@hudochenkov First, I think you forgot to read the PR title and it's content. It's a WiP (work-in-progress) PR intended to open a discussion about `@deprecated` detection **and gather community feedbacks**.

Good point about the plugin option though :+1:.

evilebottnawi:
@stylelint/core what do you think about this rule?
@FGRibreau i think best name `comment-deprecated-annotation` and do work this rule with all construction, it is be good rule :1st_place_medal:

davidtheclark:
@evilebottnawi Agree that this should be a plugin. Unfortunately, I don't have time to review plugins, because the activity in this repo is already more than I have time for.

Issue #2406, opened 2017-03-02
value-keyword-case doesn't ingore counter-reset or counter(name)


edg2s:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Bug

> Which rule, if any, is this issue related to?

value-keyword-case

> What CSS is needed to reproduce this issue?

e.g.

```css
.myClass {
	counter-reset: myCounter;
	content: '[' counter( myCounter ) ']';
	counter-increment: myCounter;
}

```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "value-keyword-case": "lower"
  }
}
```

> Which version of stylelint are you using?

7.8.0

> How are you running stylelint: CLI, PostCSS plugin, Node API?

Node API (grunt-stylelint)

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

All three rules should pass

> What actually happened (e.g. what warnings or errors you are getting)?

Rules one and two give the following error;

```shell
Expected "myCounter" to be "mycounter"   value-keyword-case
```

The `counter-increment` rule is correctly ignored, and there are tests for it.

edg2s:
See also #2404

evilebottnawi:
Done in https://github.com/stylelint/stylelint/issues/2407

PR #2407, opened 2017-03-02
Fixed: `value-keyword-case` is now ignore counter-reset property and counter, counters and attr functions.

evilebottnawi:

#2406 

### DIFF
--- a/lib/rules/value-keyword-case/__tests__/index.js
+++ b/lib/rules/value-keyword-case/__tests__/index.js
@@ -58,6 +58,32 @@ testRule(rule, {
     code: "a::before { content: \"TeSt tEsT\"}",
     description: "ignore value within quotes",
   }, {
+    code: "a { content: url(http://www.example.com/test.png) ; }",
+  }, {
+    code: "a { content: attr(value-string); }",
+  }, {
+    code: "a { content: attr(vAlUe-StRiNg); }",
+  }, {
+    code: "a { content: attr(VALUE-STRING); }",
+  }, {
+    code: "a { content: counter(counter-name); }",
+  }, {
+    code: "a { content: counter(cOuNtEr-NaMe); }",
+  }, {
+    code: "a { content: counter(COUNTER-NAME); }",
+  }, {
+    code: "a { content: counters(counter-name); }",
+  }, {
+    code: "a { content: counters(cOuNtEr-NaMe); }",
+  }, {
+    code: "a { content: counters(COUNTER-NAME); }",
+  }, {
+    code: "a { content: '[' counter(counter-name) ']' ; }",
+  }, {
+    code: "a { content: '[' counter(cOuNtEr-NaMe) ']'; }",
+  }, {
+    code: "a { content: '[' counter(COUNTER-NAME) ']'; }",
+  }, {
     code: "a { font-size: $fsBLOCK; }",
     description: "ignore preprocessor variable includes value keyword",
   }, {
@@ -673,6 +699,32 @@ testRule(rule, {
   }, {
     code: "a::before { content: \"TeSt tEsT\"}",
     description: "ignore value within quotes",
+  }, {
+    code: "a { content: url(http://www.example.com/test.png) ; }",
+  }, {
+    code: "a { content: attr(value-string); }",
+  }, {
+    code: "a { content: attr(vAlUe-StRiNg); }",
+  }, {
+    code: "a { content: attr(VALUE-STRING); }",
+  }, {
+    code: "a { content: counter(counter-name); }",
+  }, {
+    code: "a { content: counter(cOuNtEr-NaMe); }",
+  }, {
+    code: "a { content: counter(COUNTER-NAME); }",
+  }, {
+    code: "a { content: counters(counter-name); }",
+  }, {
+    code: "a { content: counters(cOuNtEr-NaMe); }",
+  }, {
+    code: "a { content: counters(COUNTER-NAME); }",
+  }, {
+    code: "a { content: '[' counter(counter-name) ']' ; }",
+  }, {
+    code: "a { content: '[' counter(cOuNtEr-NaMe) ']'; }",
+  }, {
+    code: "a { content: '[' counter(COUNTER-NAME) ']'; }",
   }, {
     code: "a { font-size: $fsblock; }",
     description: "ignore preprocessor variable includes value keyword",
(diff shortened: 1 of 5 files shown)


edg2s:
Looks good. You should add duplicate tests to the "upper" blocks (i.e. wherever we currently have counter-increment tests)

evilebottnawi:
@edg2s thanks, fixed

edg2s:
#2408 fixes the other part of this bug (counter functions)

lib/rules/value-keyword-case/index.js
```diff
@@ -4,6 +4,7 @@ const keywordSets = require("../../reference/keywordSets")
 const declarationValueIndex = require("../../utils/declarationValueIndex")
 const getUnitFromValueNode = require("../../utils/getUnitFromValueNode")
 const isCounterIncrementCustomIdentValue = require("../../utils/isCounterIncrementCustomIdentValue")
+const isCounterResetCustomIdentValue = require("../../utils/isCounterIncrementCustomIdentValue")
```
edg2s: If you're going to separate increment and reset you should use separate utils that use the respective keywordSets.

But I think it would better to just merge them into one set: keywordSets.counterKeywords / isCounterCustomIdentValue as they are identical in the spec: https://www.w3.org/TR/CSS2/generate.html#propdef-counter-reset (and also in CSS3 where counter-set is proposed)

edg2s: Should be isCounterResetCustomIdentValue

evilebottnawi:
@edg2s best could keep them separate. They can change late, imho.

edg2s:
Ok - you still have a typo in the 'require' statement.

lib/rules/value-keyword-case/__tests__/index.js
```diff
@@ -58,6 +58,32 @@ testRule(rule, {
...
+  }, {
+    code: "a { content: counter(cOuNtEr-NaMe); }",
+  }, {
+    code: "a { content: counter(COUNTER-NAME); }",
+  }, {
+    code: "a { content: counter(counter-name); }",
+  }, {
+    code: "a { content: counter(cOuNtEr-NaMe); }",
+  }, {
+    code: "a { content: counter(COUNTER-NAME); }",
```
davidtheclark: The same 3 tests are repeated, right?

lib/rules/value-keyword-case/__tests__/index.js
```diff
@@ -58,6 +58,32 @@ testRule(rule, {
...
+  }, {
+    code: "a { content: attr(value-string); }",
+  }, {
+    code: "a { content: attr(vAlUe-StRiNg); }",
+  }, {
+    code: "a { content: attr(VALUE-STRING); }",
+  }, {
+    code: "a { content: counter(counter-name); }",
+  }, {
+    code: "a { content: counter(cOuNtEr-NaMe); }",
```
davidtheclark: If you've already tested the line about `toLowerCase` (with the `value-string` tests above), I don't think we need to multiply every test with alternate cases.

lib/rules/value-keyword-case/__tests__/index.js
```diff
@@ -1028,6 +1112,16 @@ testRule(rule, {
...
+  }, {
+    code: "a { content: open-quote; }",
+    message: messages.expected("open-quote", "OPEN-QUOTE"),
+    line: 1,
+    column: 14,
+  }, {
+    code: "a { content: oPeN-qUoTe; }",
+    message: messages.expected("oPeN-qUoTe", "OPEN-QUOTE"),
+    line: 1,
+    column: 14,
```
davidtheclark: How do these tests relate to the issue?

lib/rules/value-keyword-case/index.js
```diff
@@ -64,7 +65,7 @@ const rule = function (expectation, options) {
         }
 
         // Ignore keywords within `url` and `var` function
-        if (node.type === "function" && (valueLowerCase === "url" || valueLowerCase === "var")) {
+        if (node.type === "function" && (valueLowerCase === "url" || valueLowerCase === "var" || valueLowerCase === "counter" || valueLowerCase === "counters" || valueLowerCase === "attr")) {
```
davidtheclark: How does `attr` relate to the issue?

evilebottnawi: @davidtheclark I thought that I would correct this behavior in one commit

lib/rules/value-keyword-case/__tests__/index.js
evilebottnawi: @davidtheclark yep extra tests

lib/rules/value-keyword-case/index.js
davidtheclark: That's fine. In the future please make sure to point this out in the comment about your PR, so we remember to place it in the Changelog. Is it tested?

lib/rules/value-keyword-case/__tests__/index.js
evilebottnawi: @davidtheclark typo, it should be `counters` (ending on `s`)

lib/rules/value-keyword-case/index.js
evilebottnawi: @davidtheclark yep https://github.com/stylelint/stylelint/blob/a726845c7798dfc4aee75d4f614dd2f5d91121c7/lib/rules/value-keyword-case/__tests__/index.js#L63

lib/rules/value-keyword-case/__tests__/index.js
evilebottnawi: @davidtheclark difference function to ignore `attr` and `counter`

davidtheclark:
@evilebottnawi Great! After you make those couple of changes, want to merge and add a changelog entry?

evilebottnawi:
@davidtheclark A couple of minutes and send a commit :smile:

davidtheclark:
@evilebottnawi Go ahead and merge!

evilebottnawi:
@davidtheclark merge? :smile:

davidtheclark:
Yes, merge!

hudochenkov:
To merge or not to merge...

evilebottnawi:
Added to Changelog:

> - Fixed: false positives for `attr`, `counter`, `counters` functions and ` counter-reset` property in `value-keyword-case` ([#2407](https://github.com/stylelint/stylelint/pull/2407)).

PR #2408, opened 2017-03-02
value-keyword-case: ignore counter/counters functions

edg2s:

#2406 

No, it's self explanatory.


evilebottnawi:
@edg2s i think best solution is ignore `content` property with check default values as i am do in https://github.com/stylelint/stylelint/pull/2407

edg2s:
Perhaps, although there are other functions you can use in `content` such as `attr`. We would need to add tests for all these whereas this patch is more targeted.

evilebottnawi:
@edg2s Thank you very much for helping us. I close this PR and move some logic in one PR https://github.com/stylelint/stylelint/pull/2407 . :1st_place_medal:

PR #2412, opened 2017-03-05
Fix typos

jwilk:
Found using [mwic](http://jwilk.net/software/mwic) and [anorack](https://github.com/jwilk/anorack).

### DIFF
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -203,7 +203,7 @@ Or starting with `stylelint-config-standard`, then layering `myExtendableConfig`
 }
 ```
 
-The value of `"extends"` is a "locater" (or an array of "locaters") that is ultimately `require()`d, so can fit whatever format works with Node's `require.resolve()` algorithm. That means the a "locater" can be:
+The value of `"extends"` is a "locater" (or an array of "locaters") that is ultimately `require()`d, so can fit whatever format works with Node's `require.resolve()` algorithm. That means a "locater" can be:
 
 -   The name of a module in `node_modules` (e.g. `stylelint-config-standard`; that module's `main` file must be a valid JSON configuration)
 -   An absolute path to a file (which makes sense if you're creating a JS object in a Node context and passing it in) with a `.js` or `.json` extension.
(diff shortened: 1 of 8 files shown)


docs/user-guide/configuration.md
```diff
@@ -203,7 +203,7 @@ Or starting with `stylelint-config-standard`, then layering `myExtendableConfig`
 }
 ```
 
-The value of `"extends"` is a "locater" (or an array of "locaters") that is ultimately `require()`d, so can fit whatever format works with Node's `require.resolve()` algorithm. That means the a "locater" can be:
+The value of `"extends"` is a "locater" (or an array of "locaters") that is ultimately `require()`d, so can fit whatever format works with Node's `require.resolve()` algorithm. That means a "locater" can be:
```
hudochenkov: I'm not sure this one is correct.

docs/user-guide/faq.md
```diff
@@ -144,7 +144,7 @@ All these patterns disallow CSS identifiers that start with a digit, two hyphens
 
 Use the [`defaultSeverity`](configuration.md#defaultseverity) configuration option.
 
-## Can I bundle more than one sharable config within a npm package?
+## Can I bundle more than one sharable config within an npm package?
```
hudochenkov: I'm not sure about this one too.

docs/user-guide/configuration.md
CAYdenberg: "the a" (two different articles) is never correct: it is one or the other. Looking at it in context, I think it could be either one, so this is correct.

docs/user-guide/faq.md
CAYdenberg: This is correct. Although "an" is usually used before a word beginning with a vowel, it is also used before any word beginning with a "vowel sound". Since "npm" is pronounced "en pee em" this is right.

docs/user-guide/configuration.md
hudochenkov: @CAYdenberg My fault, I read diff incorrectly :)

davidtheclark:
@jwilk This looks great. Do you have any interest in creating a PR that introduces these tools into our dependencies so we can run them regularly?

PR #2413, opened 2017-03-06
feat(flow): increase coverage from 55% to 84% in standalone.js

ivanzusko:


lib/standalone.js
```diff
@@ -1,13 +1,29 @@
...
+  line: number;
+  message: string;
+  name: string;
+  reason: string;
+  source: string;
+}*/
+
+module.exports = function (options/*: stylelint$standaloneOptions */)/*: Promise<stylelint$standaloneReturnValue>*/ {
   const files = options.files
   const code = options.code
```
ivanzusko: Maybe here utilize destructuring assignment? 

```
const {
  code, 
  codeFilename, 
  config,
  ...
  files,
  ...
  syntax,
} = options
```

hudochenkov: It doesn't work in Node 4 :(

ivanzusko: @hudochenkov oh, my bad, sorry

PR #2416, opened 2017-03-07
Docs: Mention supported .yml extension for config files

kaelig:
Turns out cosmiconfig supports the .yml extension so it's worth mentioning it.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -58,7 +58,7 @@ const meowOptions = {
         the following places, in this order:
           - a stylelint property in package.json
           - a .stylelintrc file (with or without filename extension:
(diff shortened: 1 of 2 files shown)


lib/cli.js
```diff
@@ -58,7 +58,7 @@ const meowOptions = {
         the following places, in this order:
           - a stylelint property in package.json
           - a .stylelintrc file (with or without filename extension:
-            .json, .yaml, and .js are available)
+            .json, .yaml, .yml and .js are available)
```
hudochenkov: Please, add a comma after `.yml`.

kaelig: Done!

PR #2421, opened 2017-03-10
Improve testRule reliability

hudochenkov:
`testRule` is not fully reliable. It won't catch `rejected` test case if this test case has no `message`, `line`, and `column` defined. For example, it's possible to put any `accepted` test case with `code` property only to `rejected` array. And this test would pass!

To find unreliable tests I wrote two custom matchers. I'm not sure which one is better.

Some tests are failing now, and it's a result of improved test reliability. I'll fix failing tests as soon better matcher is chosen.

What matcher is better? Should I change selected matcher's message?

### DIFF
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -6,31 +6,13 @@ const stylelint = require("./lib/standalone")
 
 global.testRule = (rule, schema) => {
   expect.extend({
-    toHaveDetails(testCase) {
-      if (
-        testCase.message === undefined
-        && testCase.line === undefined
-        && testCase.column === undefined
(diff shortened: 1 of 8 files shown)


m-allanson:
Oh, nicely found! 

From the docs it seems [only `message` is required](https://stylelint.io/developer-guide/rule-testers/#the-test-group-schema).

Maybe something like this for the message? 

```
Expected \"reject\" test case to have a \"message\" property
```

hudochenkov:
@m-allanson Thank you!

I agree with you. Have made changes.

PR #2427, opened 2017-03-16
Proof of concept: Add autofixing

hudochenkov:
I had an [idea how to make autofixing](https://github.com/stylelint/stylelint/pull/2096#issuecomment-279242941) before, but I didn't found a way how to do it ¯﻿\﻿_﻿(﻿ツ﻿)﻿_﻿/﻿¯

I came up with another idea. I think it's even simpler than my first idea. The rule will either report a warning or change PostCSS AST, depends on whether `fix` option was enabled.

Added autofixing for two rules:
* `at-rule-name-case`
* `comment-empty-line-before`

Added new tests for fixing. It would fix input and compare to expected output. And then lint fixed input. It's useful if there no expected output provided, and for extra safety.

Tests made from linting tests. Each `accept` test case received `same: true` property, to avoid duplication in tests files. Each `reject` test case received `expected` property. I did this manually, but believe it could be automated.

Also, there is an integration test. It verifies that stylesheet is fixed, and needed warnings shown.

I didn't found out anything better than `output.results[0]._postcssResult.root.toResult().css` to receive processed CSS. But this method transforms inline comments in non-standard syntaxes (`// inline`) to standard comments (`/* inline */`). I don't know yet how to preserve these comments in their non-standard form. This is why I've commented some tests for `comment-empty-line-before`.

Added `--fix` parameter to CLI.

P. S. I didn't understand how stylelint works because the code is very complicated. But I understood one thing:

### DIFF
--- a/decls/stylelint.js
+++ b/decls/stylelint.js
@@ -115,4 +115,5 @@ export type stylelint$standaloneOptions = {
   customSyntax?: string,
   formatter?: "json" | "string" | "verbose" | Function,
   allowEmptyInput?: boolean,
+  fix?: boolean,
 }
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -81,5 +81,43 @@ global.testRule = (rule, schema) => {
         })
       })
     }
+
+    if (schema.fixingCases && schema.fixingCases.length) {
+      describe("fix", () => {
+        schema.fixingCases.forEach((testCase) => {
+          const spec = (testCase.only) ? it.only : it
+          let code = testCase.code
+
+          // Fix testCase
+          spec(testCase.description || "no description. fix", () => {
+            return stylelint({
+              code,
+              config: stylelintConfig,
+              syntax: schema.syntax,
+              fix: true,
+            }).then((output) => {
+              code = output.results[0]._postcssResult.root.toResult().css
+
+              if (testCase.same) {
+                expect(code).toEqual(testCase.code)
+              } else if (testCase.expected) {
+                expect(code).toEqual(testCase.expected)
+              }
+            })
+          })
+
+          // Lint fixed testCase, it shouldn't show warnings
+          spec(testCase.description || "no description. lint fixed", () => {
+            return stylelint({
+              code,
+              config: stylelintConfig,
+              syntax: schema.syntax,
+            }).then((output) => {
+              expect(output.results[0].warnings).toEqual([])
+            })
+          })
+        })
+      })
+    }
   })
 }
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -96,7 +96,9 @@ global.testRule = (rule, schema) => {
               syntax: schema.syntax,
               fix: true,
             }).then((output) => {
-              code = output.results[0]._postcssResult.root.toResult().css
+              const result = output.results[0]._postcssResult
+
+              code = result.root.toString(result.opts.syntax)
 
               if (testCase.same) {
                 expect(code).toEqual(testCase.code)
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -249,7 +249,7 @@ Promise.resolve().then(() => {
   if (optionsBase.fix) {
     linted.results.forEach((result) => {
       if (result._postcssResult) {
-        fs.writeFile(result.source, result._postcssResult.root.toResult().css)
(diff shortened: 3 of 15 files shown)


evilebottnawi:
@hudochenkov good idea :1st_place_medal:
But we can't use  we can use `output.results[0]._postcssResult.root.toResult().css` because it is return processed CSS, this is change `sugarss` to `css`, you can try this and see that happens. But we can use `output.results[0]._postcssResult.root.source.input.css` and working with returned original css. I think we should try

hudochenkov:
`output.results[0]._postcssResult.root.source.input.css` it's just original stylesheet. I want to find a way to find a way to receive stylesheet from transformed AST.

evilebottnawi:
@hudochenkov we can fix all on processed CSS and then use http://api.postcss.org/Root.html#toString with option `stringifier`, which get from `syntax` option.

hudochenkov:
@evilebottnawi thank you for pointing out on `toString(stringifier)`! It works like a charm! Check the last commit.

The only problem is Less syntax. `postcss-less` adds an extra space after `comment` in `// comment`. It's a syntax issue. I can't do anything about it.

evilebottnawi:
@hudochenkov yep, it is related to `postcss-less` and should be fixed in own package. Seem good for me, let's wait what says another people.
/cc @stylelint/core

davidtheclark:
I am leaning towards a solution along these lines rather than the alternative with automutate (https://github.com/stylelint/stylelint/pull/2096). My reasons are:

- PostCSS is designed for mutating code, so why not take advantage of that?
- It's probably better for us to keep such essential functionality in the codebase, as part of the stylelint test suite.

I have a couple of ideas for improving (in my mind) the proof-of-concept in the PR. But before I work on that, I'd like to hear what everybody in @stylelint/core has to say about these two options. Which do you think we should try?

And a word about rollout: I'd imagine that we're going to want to set up the system with a few varied proof-of-concept rules, release, and then gradually release more and more autofixing, one rule at a time.

hudochenkov:
> I have a couple of ideas for improving (in my mind) the proof-of-concept in the PR.

I also have some ideas but don't know how to implement them yet. Tests could be automated. I did them manually. But there can be a test function which will take every `accept` test case, run it through autofixing, and there should be no changes to the code.

> Which do you think we should try?

I'm in favor of my PR. Not because it's my proposal :) I believe it's easier to understand how it works. It means it's easier to contribute an autofixing. And it would be easier to plugins authors to add autofixing to their rules.

> And a word about rollout: I'd imagine that we're going to want to set up the system with a few varied proof-of-concept rules, release, and then gradually release more and more autofixing, one rule at a time.

Agree, we'll make it rule by rule. I believe it's even possible to include this feature in v8. Although it's non-breaking change, it deserves major release.

davidtheclark:
@jeddy3 I'm curious what you think about these two autofixing proposals, before we continue with either one.

jeddy3:
> @jeddy3 I'm curious what you think about these two autofixing proposals, before we continue with either one.

I think each approach has merit and it's fascinating to piece through them.  It feels like there is a general consensus within the team already, and I'm in agreement with the conclusion reached in this PR i.e. PostCSS is designed for mutating code, and it's an API we're all familiar with.

So, I think it's worth exploring the proof of concept in this PR further.

I think stylelint itself has reached a maturity where there is trust in the stability of the project. As such, perhaps we can document this as an experimental feature, much like we did with Less support? 

Thanks everyone for the effort you've put into thinking about how autofixing could work. Two years ago we set out to build a CSS linter, and I think we've achieved that. It feels like we're about to embark on a new (and possibly a more ambitious) project. I think if we strike that same balance between features, sustainable development and the long-term health of the project, this endeavour will, hopefully, work out just as well :)

davidtheclark:
@hudochenkov: If you'd like to make changes to this PR before I take a pass, go for it.

hudochenkov:
@davidtheclark I've just rebased the branch to avoid merge conflicts.

Go ahead! Can't wait to see your ideas! :)

jeddy3:
@hudochenkov Thanks for spearheading the autofixing using PostCSS approach. As your concept has been incorporated into #2467, I'm going to close this PR.

Issue #2404, opened 2017-03-01
False positives for at-variables prefixed with negative sign in value-keyword-case


edg2s:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

bug

> Which rule, if any, is this issue related to?

value-keyword-case

> What CSS is needed to reproduce this issue?

```less
.foo {
  width: @myWidth;
  height: -@myHeight;
  margin: 0 - @myMargin;
}
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "value-keyword-case": "lower"
  }
}
```

> Which version of stylelint are you using?

7.8.0

> How are you running stylelint: CLI, PostCSS plugin, Node API?

Node API (grunt-stylelint)

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

LESS

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

`Expected "-@myHeight" to be "-@myheight"`

The positive value `@myWidth` doesn't trigger an error, nor does the expression `0 - @myMargin`.

edg2s:
See also #2406

jeddy3:
@edg2s FYI, addressing non-standard syntax issues is not a priority for the core team. As such, please consider contributing a fix if you'd like to see this issue resolved.

edg2s:
It looks like the issue here is the valueParser, it doesn't break up expressions unless there are spaces, so while `'0 - @foo'` parses as `['0', ' ', '-', ' ', '@foo']` and each one gets detected as a non-keyword, `'0-@foo'` parses as one node: `['0-@foo']` which can't be easily checked as a keyword

evilebottnawi:
@edg2s can you create issue in `value-parser` repo?

edg2s:
I think the issue is that `value-parser` is specifically designed to only work on regular CSS syntax (as it is currently defined). It seems there is another parser [postcss-values-parser](https://github.com/lesshint/postcss-values-parser/) designed just for this very purpose so I am experimenting with that.

evilebottnawi:
Done in https://github.com/stylelint/stylelint/pull/2558

Issue #1183, opened 2016-05-04
Evaluate postcss-values-parser


jeddy3:
Ref: https://github.com/lesshint/postcss-values-parser

e.g. would consistent walk methods across PostCSS, the selector and value parser be useful?:
https://github.com/lesshint/postcss-values-parser/blob/master/API.md#containerwalk-proxies

evilebottnawi:
@jeddy3 look as we can create https://github.com/stylelint/stylelint/issues/1031

davidtheclark:
PostCSS is considering integration of selector and value parsers for v6, so I might want to hold off on switching libraries if we might just switch again soon: https://github.com/postcss/postcss/issues/754

evilebottnawi:
@jeddy3 i think we can close this issue, and wait postcss v6

evilebottnawi:
@jeddy3 close issue?

jeddy3:
> @jeddy3 close issue?

Yep. I'm just working my way down though the issues and I've finally arrived here :)

PR #2438, opened 2017-03-23
Migrate to postcss-values-parser

edg2s:

#2404

Uses a more detailed parser that can handle Less/SCSS expressions, even if they have no spaces.

### DIFF
--- a/lib/rules/value-keyword-case/index.js
+++ b/lib/rules/value-keyword-case/index.js
@@ -11,7 +11,7 @@ const report = require("../../utils/report")
 const ruleMessages = require("../../utils/ruleMessages")
 const validateOptions = require("../../utils/validateOptions")
 const _ = require("lodash")
-const valueParser = require("postcss-value-parser")
+const valuesParser = require("postcss-values-parser")
 
 const ruleName = "value-keyword-case"
 
@@ -56,7 +56,7 @@ const rule = function (expectation, options) {
       const prop = decl.prop,
         value = decl.value
 
-      valueParser(value).walk(node => {
+      valuesParser(value, { loose: true }).parse().first.walk(node => {
         const valueLowerCase = node.value.toLowerCase()
 
         // Ignore system colors
@@ -65,14 +65,16 @@ const rule = function (expectation, options) {
         }
 
         // Ignore keywords within `url` and `var` function
-        if (node.type === "function" && (valueLowerCase === "url" || valueLowerCase === "var" || valueLowerCase === "counter" || valueLowerCase === "counters" || valueLowerCase === "attr")) {
-          return false
+        if (node.type === "func" && (valueLowerCase === "url" || valueLowerCase === "var" || valueLowerCase === "counter" || valueLowerCase === "counters" || valueLowerCase === "attr")) {
+          // Prevent further descending, but don't return false as that stops all walking
+          node.walk = null
+          return
         }
 
         const keyword = node.value
 
         // Ignore css variables, and hex values, and math operators, and sass interpolation
-        if (node.type !== "word" || !isStandardSyntaxValue(node.value) || value.indexOf("#") !== -1 || ignoredCharacters.has(keyword) || getUnitFromValueNode(node)) {
+        if (node.type !== "word" || !isStandardSyntaxValue(keyword) || value.indexOf("#") !== -1 || ignoredCharacters.has(keyword) || getUnitFromValueNode(node)) {
           return
         }
 
--- a/lib/rules/value-keyword-case/__tests__/index.js
+++ b/lib/rules/value-keyword-case/__tests__/index.js
@@ -285,12 +285,12 @@ testRule(rule, {
     code: "a { transition: -Webkit-transform 2s; }",
     message: messages.expected("-Webkit-transform", "-webkit-transform"),
     line: 1,
-    column: 17,
+    column: 18,
   }, {
     code: "a { transition: -webkit-Transform 2s; }",
     message: messages.expected("-webkit-Transform", "-webkit-transform"),
     line: 1,
-    column: 17,
+    column: 18,
   }, {
     code: "a { background-image: linear-gradient(To right, white calc(100% - 50em), silver); }",
     message: messages.expected("To", "to"),
@@ -593,6 +593,12 @@ testRule(rule, {
     code: "a { margin: @mArGiN + 10px; }",
   }, {
     code: "a { margin: @MARGIN + 10px; }",
+  }, {
+    code: "a { margin: -@margin; }",
+  }, {
+    code: "a { margin: -@mArGiN; }",
+  }, {
+    code: "a { margin: -@MARGIN; }",
   }, {
     code: ".bordered { border-bottom: solid 2px black; }",
     description: "inside mixin",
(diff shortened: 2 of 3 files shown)


lib/rules/value-keyword-case/index.js
```diff
@@ -65,14 +65,16 @@ const rule = function (expectation, options) {
         }
 
         // Ignore keywords within `url` and `var` function
-        if (node.type === "function" && (valueLowerCase === "url" || valueLowerCase === "var" || valueLowerCase === "counter" || valueLowerCase === "counters" || valueLowerCase === "attr")) {
-          return false
+        if (node.type === "func" && (valueLowerCase === "url" || valueLowerCase === "var" || valueLowerCase === "counter" || valueLowerCase === "counters" || valueLowerCase === "attr")) {
+          // Prevent further descending, but don't return false as that stops all walking
+          node.walk = null
```
hudochenkov: `node.walk` is a PostCSS API. I don't think it's a good idea, to break PostCSS API.

jeddy3:
@edg2s Thanks for looking into this. 

The code itself looks great. However, I think this issue will need to trigger a side discussion about whether we wish to introduce another value parser _to support non-standard syntax_.

We're always striving to find the right balance between features and sustainable development. Endeavouring for as much consistency as possible across the rules has helped us keep things maintainable, and development sustainable.

I can think of three options going forward:

- Continue to only use `postcss-value-parser` and mark this bug as wontfix (consistency over features).
- Migrate wholesale (approx 30 files make use of it) to `postcss-values-parser` (consistent and features, but at an effort)
- Allow two value parsers and document when to use one over the other (features over consistency)

Related to https://github.com/stylelint/stylelint/issues/1183

hudochenkov:
Also, we [might drop Less support](https://github.com/stylelint/stylelint/pull/1899) soon. Worth to consider before any changes related to Less support only.

edg2s:
Currently the code has a whole bunch of rules for less/scss support which assume that the value is getting parsed correctly. I think there should either be support or no support (which would be a terrible shame as we'd probably have to stop using it at Wikimedia). The status quo is that we pretend the parser works when in fact it works 90% of the time.

edg2s:
With regards to points 2 & 3, we could create a tech debt bug to migrate to values-parser and either do that in one go as you say, or piece by piece.

edg2s:
Also the APIs were almost identical once I'd worked out how to do the same tree walk, so I don't think such a migration would be that hard, and would be happy to help.

evilebottnawi:
@edg2s What benefits do we get besides supporting less? Speed, more correct ast and etc? Then it will be easier to decide whether it is necessary or not.

edg2s:
Someone else may have a better test rig for determining performance, although as a more sophisticated parser it may end up being slightly slower. Certainly it appears to be better at parsing expressions so it's conceivable it might be useful for vanilla CSS, such as the [calc() function](https://developer.mozilla.org/en/docs/Web/CSS/calc).

jeddy3:
@edg2s Thanks for sharing your further thoughts.

> I think there should either be support or no support (which would be a terrible shame as we'd probably have to stop using it at Wikimedia).

Is there anyone at Wikimedia with experience with parsers (or is wanting to learn) who can pick up the maintenance of `postcss-less`?

> Someone else may have a better test rig for determining performance

You can test the performance difference of your changes to the `value-keyword-case` rule using stylelint's [benchmarking tool](https://stylelint.io/developer-guide/rules/#improving-the-performance-of-a-new-or-an-existing-rule). Just run it against `master` and then again against your branch, and then post your results below.

> Certainly it appears to be better at parsing expressions so it's conceivable it might be useful for vanilla CSS, such as the calc() function.

That's a good point. There's a [summary of other differences](https://github.com/lesshint/postcss-values-parser#postcss-values-parser-vs-postcss-value-parser) within the `postcss-values-parser` README.

> Also the APIs were almost identical once I'd worked out how to do the same tree walk, so I don't think such a migration would be that hard, and would be happy to help.

Thanks for the offer!

If @edg2s is willing to do the legwork, and if the performance difference is acceptable given the new features, I think option 2 is a valid route to go down. It sounds like the behaviour of the parser is more similar to the ones in PostCSS and `postcss-selector-parser`, and so we might be better positioned if value and selector parsing are rolled into PostCSS in the future. Also, I think the consistency will help new contributors maintain the codebase.

davidtheclark:
I agree with @jeddy3. 👍

jeddy3:
>  and if the performance difference is acceptable given the new features,

I ran the `value-keyword-case` rule of `master` and @edg2s fork through the benchmarking script and there appears to be no discernible difference in performance. As such, I think option 2 is good to go.

@edg2s Is this something you'd be willing to champion in this PR?

edg2s:
> Is there anyone at Wikimedia with experience with parsers (or is wanting to learn) who can pick up the maintenance of postcss-less?
Well MediaWiki uses its own markup language so in many teams it is a required skill :) I can certainly ask around.

I'll have to come back to converting the other 30 as it will still take a fair bit more time than this patch, but if anyone else wants to get started on it...

jeddy3:
> I can certainly ask around.

That'd be awesome thanks.

> I'll have to come back to converting the other 30 as it will still take a fair bit more time than this patch, but if anyone else wants to get started on it...

No worries, we'll leave this PR open for now so you can work on it when you have time. I'll also create a new issue to see if anyone else is able to chip away at it in the meantime.

edg2s:
So working on a few new rules - it appears that values-parser has some issues of it's own, mostly with SASS (as it was written with LESS in mind). I filed a few issues, but the maintainer was sadly less than helpful. Given than both parsers have their limitations with extended syntax the only option at this stage would be to use whichever one is best for the job.

davidtheclark:
@edg2s, thanks for investigating. Too bad it didn't work out. Seems the smoothest path right now would be to stick with the value parser we have and add exceptions and workarounds as needed.

jeddy3:
@edg2s Yes, thank you! From reading those `values-parser` issues, it looks like you dived deeply into making this PR work.

> Seems the smoothest path right now would be to stick with the value parser we have and add exceptions and workarounds as needed.

What that in mind, should this PR now be merged? If so, I'd like to see a brief comment in the source code that highlights the use of `values-parser` in the rules and what it works around.

evilebottnawi:
`postcss-values-parser` works as expected, because all mathematical expression should be interpreted as they inside in `calc` function (`-` and `+` always have space before and after sign). I.e `10+@value` and `10-@value` should be interpreted as is, only one thing don't work as expected `10*@value` and `10/@value`. `-@value` and `+@value` should be interpreted as is,
`*@value` and `/@value` should be throw error  (need create issue and send PR).

My opinion on this issue:
Creating a new parser(`postcss-values-parser`) was a bad idea, you could come to an agreement and create one handy parser for values, guys from `postcss` responsive developers. Now we have two parsers with weird and incorrect behavior. 

Nothing complicated, we just need to better interact with each other. 
I apologize if this sounded a little rude.

BTW, if fix this behavior in `stylelint`

jeddy3:
Closing in favour of https://github.com/stylelint/stylelint/pull/2558

Thanks for your work on this, but another way was found that uses the existing parser.

Issue #2435, opened 2017-03-20
Allow lines with only whitespace to be treated as "empty"


chenryhen:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug where empty lines with whitespace still trigger an error of "Expected empty line before rule" even if Stylelint is configured to ignore whitespace in empty lines.

> Which rule, if any, is this issue related to?

`no-eol-whitespace` and `rule-empty-line-before`

> What CSS is needed to reproduce this issue?

```css
.foo {
  color: pink;
}
. . .
.bar {
  color: green;
}
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "no-eol-whitespace": [true, {
      "ignore": ["empty-lines"]
    }],
    "rule-empty-line-before": "always"
  }
}
```

> Which version of stylelint are you using?

`7.9.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

plug-in for webpack

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

no

> What did you expect to happen?

No "expected empty line before rules" errors

The reasoning being that if I'm explicitly ignoring whitespace in empty lines, I would expect Stylelint to consider those empty lines as "empty" for all other rules. Otherwise, you need to go in and delete the whitespace anyway.

> What actually happened (e.g. what warnings or errors you are getting)?

See above. Deleting the whitespace in the empty line above the rule removes the ensuing "Expected empty line before rule" error.

jeddy3:
@chenryhen Thanks for the report and for using the template.

I think `no-eol-whitespace`'s `"ignore": ["empty-lines"]` option is a pretty recent addition to stylelint and before that we didn't really have a consistent notion of what "empty" meant. 

Now that we have defined "empty" to be "contains nothing _or only whitespace_", it seems sensible to update all the `*-empty-line-before` rules to consider a line with _only whitespace on_ to be empty.

Additionally, as it currently stands, I think there is a slight overlap between the `no-eol-whitespace` rule and the `*-empty-line-before` rules as the `*-empty-line-before` rules are _indirectly_ checking for trailing whitespace on "empty" lines.

@stylelint/core Any objections to changing the behaviour of the `*-empty-line-before` rules? I believe this change would _loosen_ the restrictiveness of the rules and would not a breaking change. 

[...]

PR #2440, opened 2017-03-24
*-empty-line-before rules consider line as empty if it contains whitespace only

hudochenkov:

https://github.com/stylelint/stylelint/issues/2435

Based on `v8` because it's a breaking change for `never*` options.


hudochenkov:
Added to changelog:

-   Changed: `*-empty-line-before` rules consider line as empty if it contains whitespace only ([#2440](https://github.com/stylelint/stylelint/pull/2440)).

Issue #2446, opened 2017-03-27
Issue with arrays in function-url-scheme-whitelist


aaronkahlhamer:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Bug

> Which rule, if any, is this issue related to?

function-url-scheme-whitelist

> What CSS is needed to reproduce this issue?

```css
a {
  background-image: url('data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs=');
}

```

> What stylelint configuration is needed to reproduce this issue?

```json
"function-url-scheme-whitelist": ["https", "data"]
```

> Which version of stylelint are you using?

7.9.0 but the issue can be recreated in the demo at https://stylelint.io/demo/

> How are you running stylelint: CLI, PostCSS plugin, Node API?

gulp-stylelint

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

To loop through the array and ignore both https and data.

> What actually happened (e.g. what warnings or errors you are getting)?

When running the following in the demo...

```css
a {
  background-image: url('data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs=');
}
```

```json
"function-url-scheme-whitelist": ["https", "data"]
```

...it reports, Unexpected url scheme "data:"

aaronkahlhamer:
Screenshot of error...

![image](https://cloud.githubusercontent.com/assets/695639/24373270/68974622-12f6-11e7-8a08-7a1616f8a47f.png)

hudochenkov:
Thank you for catching this bug!

hudochenkov:
Done in https://github.com/stylelint/stylelint/pull/2447

aaronkahlhamer:
Thank you!

PR #2447, opened 2017-03-27
Fix not specified array as primary option for "function-url-scheme-whitelist"

hudochenkov:

https://github.com/stylelint/stylelint/issues/2446

For some strange reason it wasn't defined, that rule can accept an array as primary option. Test didn't break, though. I wonder why? I believe we should have tests to test this behavior, but I have no idea how to do them.

I've tested results before and after this change via Node API. It stopped throwing warnings.

### DIFF
--- a/lib/rules/function-url-scheme-whitelist/index.js
+++ b/lib/rules/function-url-scheme-whitelist/index.js
@@ -68,6 +68,8 @@ const rule = function (whitelist) {
   }
 }
 
+rule.primaryOptionArray = true
+
 rule.ruleName = ruleName
 rule.messages = messages
 module.exports = rule


davidtheclark:
@hudochenkov That property ends up being used in https://github.com/stylelint/stylelint/blob/01167e6ed006a9cd2fe8e0777e84b3d07736b4d2/lib/normalizeRuleSettings.js#L23. We could run try settings for tests through this function, to ensure that the settings in tests undergo the same processing they do in the real program.

evilebottnawi:
@davidtheclark it is be good

hudochenkov:
@davidtheclark I'm afraid I didn't understand your idea :(

davidtheclark:
@hudochenkov Could you merge this PR and open up an issue for follow-up, and I will try to find some time this weekend to try some things?

hudochenkov:
Added to changelog:

-   Fixed: `function-url-scheme-whitelist` was working incorrectly if more than one URL scheme were specified ([#2447](https://github.com/stylelint/stylelint/pull/2447)).

Issue #2444, opened 2017-03-27
Add Support for TypeScript


alan-agius4:
Hi,

Are you willing to accept TypeScript typings within the repo itself?

This will make it more convenient for typescript users (and [suggested by typescript](https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package)) to provide typings with the repository itself, rather than another repo (such as [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped)).

Are you willing to have a pull request for this?

I will be happy to create typing for this library myself if you agree.

Thanks!

davidtheclark:
Nope. I do not want to mess with Typescript. If you have any other less intrusive ideas for making package consumption easier for Typescript users, let us know.

alan-agius4:
The least intrusive ideas would be to have typings as a seperate file. Example /typings/index.d.ts. 

Your code won't change a bit. You'll code the way u do now. 

Basicilly the change would be to include this file within you library 

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/stylelint/index.d.ts

evilebottnawi:
@alan-agius4 can you do PR with sample example?

alan-agius4:
Yeah surly I can do it tmr or the day after

Arcanemagus:
The problem I see here is that as the API changes the authors here will need to keep that maintained, when this is something that isn't desired in the first place it will likely to be forgotten and become out of date.

alan-agius4:
In case the API is updated at the moment at the moment is harder to maintain as 3rd party users will have to go and update https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/stylelint and keep it maintained.

jeddy3:
> The problem I see here is that as the API changes the authors here will need to keep that maintained

I agree.

> 3rd party users will have to go and update https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/stylelint and keep it maintained.

I think having it maintained by a 3rd party in the DefinitelyTyped repo correctly signals that the stylelint org isn't responsible for keeping it up to date i.e. the org puts out a release (with a change of API) and then a 3rd party can update the definitions on DefinitelyTyped _after the fact_, rather than the stylelint org being responsible for updating the definitions _before_ the release.

alan-agius4:
Okay, thanks for the feedback and interest anyways.

davidtheclark:
[...]

PR #2448, opened 2017-03-28
feat(typings): add `TypeScript` typings

alan-agius4:
Adding typings for `stylelint` to enable user using `TypeScript`.

By no means the is all the API that you are exposing however so far this is what I discovered and should be a good starting point, to enable users to consume your lib with typescript. If you have any other useful methods that users that use let me know and I can create the typyings for them as well.

Thanks

#2444

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "url": "https://github.com/stylelint/stylelint.git"
   },
   "main": "lib/index.js",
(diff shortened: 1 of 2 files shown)


davidtheclark:
Closing based on [discussion in](https://github.com/stylelint/stylelint/issues/2444#issuecomment-291012481) #2444.

Issue #717, opened 2016-02-09
Lint a directory


silverwind:
With ESLint I can run `eslint public` to recursively lint all supported files in the `public` directory. Trying the same with `stylelint public` results in:

```
Error: EISDIR: illegal operation on a directory, read
    at Error (native)
```

silverwind:
I can possibly work on this if you think it's a worthy feature to have :)

davidtheclark:
We can add something like this but need to determine exactly what rules would work. "Any input that does not [something] should be appended with `**/*.css`" ....

davidtheclark:
And can't forget `.scss` files, I suppose.

silverwind:
Also `.less`, `.styl`, `.gss` and others, if they are supported.

davidtheclark:
Yeah. They are not supported yet, though -- until somebody creates those
parsers for PostCSS.

jeddy3:
And not forgetting `.postcss`, which I've seen around.

This feature [looks to be](https://github.com/Holixus/eslint/commit/3be29956facae148eec43d24a7a29e7ce0ab110b) configurable in eslint. 

Do we want to go down this path? Or do we not mimic eslint and just continue to rely on a glob of files being passed in?

davidtheclark:
I'm fine giving this a shot if it has solid logic -- works without being confusing or surprising.

But it is true that the variety of possible file extensions complicates things.

jeddy3:
> I'm fine giving this a shot if it has solid logic -- works without being confusing or surprising.

Sgtm.

davidtheclark:
I've thought about this and may have lost interest.

By passing a glob, you are explicitly telling stylelint what to load. And you have a standard method, the glob, by which you can specify *exactly* what you want. But if we go down the route of this feature request we are instead making assumptions, and a bunch of questions arise. If I pass a glob that resolves to both files and directories do all of those directories auto-expand to include all their files and directories? How can I tell stylelint to look for my own special extension to CSS files, `.wtfcss`? What other shortcuts can I request?

One quality of ESLint that I do not want to mimic is the enormity of its codebase. It does so much itself. I would much rather outsource what we can, especially when there are standard patterns available like node globs.

So I'm going to close this one. Please pipe up if you disagree.

jeddy3:
Agreed.

merlinpatt:
[...]

PR #2451, opened 2017-03-29
Add text about searching nested directories

merlinpatt:
As suggested in issue #717, I have added a couple paragraphs about how to lint nested directories.

I added this right above the examples because that's the first place that discusses how to use the stylelint cli.

I also added a specific section titled recursively linting. I added this section to specifically call out the need for the quotes. Also, I specifically used the word recursive because it's another word that people may search for when looking for this functionality.

### DIFF
--- a/docs/user-guide/cli.md
+++ b/docs/user-guide/cli.md
@@ -16,6 +16,8 @@ The CLI outputs formatted results into `process.stdout`, which you can read with
 
 ### Examples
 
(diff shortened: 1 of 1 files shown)


davidtheclark:
Right idea, @merlinpatt. I made a couple of modifications to improve the accuracy.

jeddy3:
@merlinpatt Thanks!

PR #2455, opened 2017-03-30
fix: keep scope when using `schema.preceedingPlugins`

joscha:

When creating a `ruleTester`, a `schema` can be passed in the test setup, e.g.:

```js
const postCssModuleValues = require('postcss-modules-values');
{
    preceedingPlugins: [postCssModuleValues],
    ruleName: cssImportsOrder.ruleName,
    config: { preset: 'suit' },
    skipBasicChecks: true,
    accept: [...]
    reject: [...]
}
```
when using `preceedingPlugins`, this currently fails with: 
```
   Uncaught TypeError: Cannot read property 'plugins' of undefined
      at use (node_modules/postcss/lib/processor.js:86:24)
      at Array.forEach (native)
      at postcssProcess (node_modules/stylelint/lib/testUtils/createRuleTester.js:163:32)
      at passingTestCases.forEach.acceptedCase (node_modules/stylelint/lib/testUtils/createRuleTester.js:179:29)
      at Array.forEach (native)
      at processGroup (node_modules/stylelint/lib/testUtils/createRuleTester.js:174:22)
      at process.nextTick (node_modules/stylelint/lib/testUtils/createRuleTester.js:117:9)
      at _combinedTickCallback (internal/process/next_tick.js:67:7)
      at process._tickCallback (internal/process/next_tick.js:98:9)
      at run (bootstrap_node.js:422:7)
      at startup (bootstrap_node.js:143:9)
      at bootstrap_node.js:537:3
```

due to the fact that the `.use` method of the postcss processor needs access to `this`.

Should be fairly straightforward.

### DIFF
--- a/lib/testUtils/__tests__/createRuleTester.test.js
+++ b/lib/testUtils/__tests__/createRuleTester.test.js
@@ -4,14 +4,14 @@ const sinon = require("sinon")
 const createRuleTester = require("../createRuleTester")
 
 function createRuleTesterPromise(rule, schema) {
-  const assertions = []
+  const contexts = []
   return new Promise((resolve, reject) => {
     const ruleTester = createRuleTester((promise, res) => {
-      assertions.push(res)
+      contexts.push(res)
       promise.then(resolve, reject)
     })
     ruleTester(rule, schema)
-  }).then(() => assertions)
+  }).then(() => contexts)
 }
 
 describe("createRuleTester", () => {
@@ -27,8 +27,8 @@ describe("createRuleTester", () => {
     })
 
     return createRuleTesterPromise(() => rule, schema)
-      .then(assertions => {
-        expect(assertions).toEqual(
+      .then(contexts => {
+        expect(contexts).toEqual(
           [
             { comparisonCount: 1,
               caseDescription: "\n> rule: my-test-rule\n> config: \n> code: \"\"\n",
--- a/lib/testUtils/__tests__/createRuleTester.test.js
+++ b/lib/testUtils/__tests__/createRuleTester.test.js
@@ -1,45 +1,63 @@
 "use strict"
 
+const sinon = require("sinon")
 const createRuleTester = require("../createRuleTester")
 
-describe("createRuleTester", () => {
-  it("is possible to create a tester", () => {
-    const results = []
-    return new Promise((resolve, reject) => {
-      const tester = createRuleTester((promise, res) => {
-        results.push(res)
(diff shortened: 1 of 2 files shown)


lib/testUtils/createRuleTester.js
```diff
@@ -160,7 +160,7 @@ function processGroup(rule, schema, equalityCheck) {
     processor.use(assignDisabledRanges)
 
     if (schema.preceedingPlugins) {
-      schema.preceedingPlugins.forEach(processor.use)
+      schema.preceedingPlugins.forEach(processor.use.bind(processor))
```
joscha: can also make this a `schema.preceedingPlugins.forEach(plugin => processor.use(plugin))` if that's favorable

davidtheclark: Let's do that: it fits the code conventions better.

lib/testUtils/__tests__/createRuleTester.test.js
```diff
@@ -0,0 +1,45 @@
+"use strict"
+
+const createRuleTester = require("../createRuleTester")
+
+describe("createRuleTester", () => {
+  it("is possible to create a tester", () => {
```
davidtheclark: Great idea to add tests like this! My one reservation is that the one test covers several different possibilities. It *was* possible to create a tester before, right? — just not with `preceedingPlugins`? Maybe that means there should be a couple of separate tests?

joscha: Sure, I can split them up. I just seem to have been the first that ever introduced a test for `createRuleTester` and the guidelines said to provide a test that exemplifies the issue ;) Happy to add another one before, they are quite verbose however as it pretty much uses the whole stylelint suite, but I agree it would be good to have them.

joscha:
@davidtheclark changes are in 34f092d, please have another look

joscha:
@davidtheclark any chance to get this fix in?

jeddy3:
Changelog:

> -   Fixed: scope is kept when using `schema.preceedingPlugins` ([#2455](https://github.com/stylelint/stylelint/pull/2455)).

joscha:
awesome, thank you @jeddy3 - any chance to get this in dot-release?

jeddy3:
> any chance to get this in dot-release?

Sure thing. Patch releases don't take much time.

Released in `7.10.1`.

joscha:
Thank you!

PR #2464, opened 2017-04-01
Ignoring improvements

### LINKED ISSUE #2236: Disregard default ignores when ignoreFiles is used

### LINKED ISSUE #2399: Improve performance of globbing with ignored paths

Stylelint works extremely slow with large directories by taking a lot of time to discover files targeted by input glob, despite glob contain paths, ignored in Stylelint config with `ignoreFiles` option.

I've digged in code and discovered that it happens, because Stylelin initially passes to `globby` whole input glob and does not account for specified in `ignoreFiles` option path and files.

By stumbling into very large directory, `globby` tries to explore it completely and it takes a lot of time (with 4000 files it almost endless), while it had to ignore it in first place due to specified in `ignoreFiles` option path.

Partially related to `ignoreFiles` option.

Fortunately, no CSS needed to reproduce this issue. However, you will need directory with Craft CMS and Nginx (with around 4000 files) near css files.

You need to launch Stylelint with very wide glob as input:

``` shell
stylelint **/*.css
```

And large directory near your css files.

`7.9.0`

Project with following structure:

```
|- docker <- directory with 4000 files
|- source
    |-- *.css
...
```

CLI with

``` shell
stylelint **/*.css
```

Config stored in `.stylelintrc.yml`:

``` yaml
---
ignoreFiles:
  - docker/{,**/}*
  - docker/***
  - '**/docker/{,**/}*'
  - '**/docker/**'

extends: stylelint-config-standard
```

Funky and in some sense same globs for `ignoreFiles` just to be sure that I passed in correct path.

No

Stylelint should add specified in `ignoreFiles` option paths to passed to `globby` files list as ignored (with `!`), so that `globby` won't try to scan directory which won't be ever used at first place.

Stylelint hangs for very long time.

davidtheclark:
The aim is to close #2399, close #2236 , and replace PR #2382.

- As part of this, I've decided to remove the `allowEmptyInput`. I think that the opinion has come up before that we could just silently not lint any files, if no files match. This is what ESLint does, and I'm happy to stick with that. The reason I'm doing it here is that remove that option removes some complexity we would otherwise face with the new ignoring system.
- Changes syntax of `.stylelintignore` from strict `.gitignore` globs to `node-glob` globs. I *think* this won't bother anybody, because it seems like [the `.gitignore` patterns](https://git-scm.com/docs/gitignore#_pattern_format) are a subset of the `node-glob` patterns — and I preserved the ability to add comments.

There's more work to be done, especially in finding code to delete and tests to clarify.


jeddy3:
It's looking great so far!

davidtheclark:
Added `disableDefaultIgnores` option to standalone and CLI.

davidtheclark:
I think this does the trick now.

@ArmorDarks before we merge this change, can you help us out by trying this branch? I ask because you filed the "slow globbing" issue and have a codebase large enough to notice the difference if we've fixed the situation.

What you'll need to do to test this solution is:
- Move your `ignoreFiles` globs to `.stylelintignore`. (`ignoreFiles` is always going to be inefficient; but `.stylelintignore` should be efficient — we'll recommend this in the docs.)
- Run stylelint out of this branch (`node lib/cli.js ...`) and see if it's faster than before in your codebase.

Let me know what you find!

ArmorDarks:
Hi

Sorry for the delay. Sure, I will try to do it within few days, if it's ok.

jeddy3:
> I think this does the trick now.

Nice! The code looks great.

I'm happy a bit of trouble making this branch work, though. I have this folder structure:

```
/src/
  /1/
    test.css
  /2/
    test.css
.gitignore
.stylelintignore
.stylelintrc.json
```

With the content of `.gitignore` and `.stylelintignore` being:

```
src/1
```

Am I right in thinking that both these should ignore the `tests.css` file in `1`?:

```console
node ../stylelint/lib/cli.js "src/**/*.css"
node ../stylelint/lib/cli.js "src/**/*.css" --ignore-file .gitignore
```

Neither does, though. I've tried different ignore globs too e.g. `src/1/*.css`

I'd like to quickly determine if there's something amiss with my local setup. It seems likely as the tests in the PR look sound. As such, can someone else in @stylelint/core try this branch against their codebase and post their findings.

davidtheclark:
🤔 The problem seems to be in the way that globby deals with relative and absolute paths. Duplicating your setup, here's the globs I send to globby, and the files I get back: 

```
[ 'src/1/test.css',
  'src/2/test.css',
  '!**/node_modules/**',
  '!**/bower_components/**',
  '!/Users/davidclark/Downloads/test/src/1' ]
[ 'src/1/test.css', 'src/2/test.css' ]
```

If, however, I specify absolute paths for the input, here's what I get (it works):

```
[ '/Users/davidtheclark/Downloads/test/src/**/*.css',
  '!**/node_modules/**',
  '!**/bower_components/**',
  '!/Users/davidclark/Downloads/test/src/1' ]
[]
```

I'll look into whether this is expected behavior or not. We could probably fix it if necessary by making the input files absolute ...

davidtheclark:
Ok, so part of this looks to be solved (if tests still passed) by *not* absolutizing the paths we read from the ignore file.

However, another point came up when trying this: I was too hopeful that `.gitignore` would fit nicely into glob syntax. It won't. `src/1` will work in `.gitignore` but not with a glob: you need `src/1/**/*` or something like that.

The trouble here is that we really don't want to analyze things twice, with two separate interpreters. We want to pass globby the right glob, not one without ignored files which we then filter. 

We could try to get around this trouble by adjusting for directories:
- For any line ending with a slash, add `**/*` (e.g. `src/1/`).
- For any line not ending with an extension, add `**/*` (e.g. `src/1`).

I'm not sure if that would cover all the bases. Anybody want to investigate? (I may or may not have time to do so today.)

davidtheclark:
Tests did not pass, so looks like no parts are solved 😞

davidtheclark:
I think I got it ...

Latest commit restores the specific `.gitignore` syntax, because that's preferable to regular globs for this situation. It uses `ignore`'s filter after the glob is analyzed but *before* any files are read. Glob-reading is pretty efficient, I think — I just tried read ~16000 files and it took under a second. So the inefficiency is when we load the file's config before determining whether it's ignored or not. This setup avoids that.

@jeddy3: I believe this works in the test-case you offered now. Want to double-check?

jeddy3:
> I believe this works in the test-case you offered now. Want to double-check?

Worked a treat thanks! I was able to test `.stylelintignore`, `--ignore-path` and `--disable-default-ignores`.

> I just read ~16000 files and it took under a second. So the inefficiency is when we load the file's config before determining whether it's ignored or not. This setup avoids that.

That's excellent news.

jeddy3:
@stylelint/core Anyone else use ignores in the projects? If so, could you possibly give this branch a whirl?

davidtheclark:
Let's merge this now, so the PR does not grow stale; and if anybody has problems or more concerns they can create new issues.

davidtheclark:
Added to changelog:

> -   Added: `disableDefaultIgnores` option (`--disable-default-ignores` in CLI), to allow linting of `node_modules` and `bower_components` directories ([#2464](https://github.com/stylelint/stylelint/pull/2464)).
> -   Added: more efficient file ignoring with `.stylelintignore` ([#2464](https://github.com/stylelint/stylelint/pull/2464)).

davidtheclark:
To be clear: added those to the v8 changelog.

alexgorbatchev:
> we could just silently not lint any files, if no files match. This is what ESLint does, and I'm happy to stick with that.

I'm writing a CLI tool that executed `stylelint`. I found it problematic to test integration because i can't distinguish between the case where I'm passing bad arguments on purpose and the case where file validates OK. Essentially, I'm only able to test if the file has linting issues.

Furthermore, I think if specific file path is passed and file doesn't exist, it should fail. I'm asking for a file to be linted, it's not there, therefore something is wrong. If the glob is passed, current behavior feels ok.

What do you think?

ntwb:
@alexgorbatchev Thanks for the comment, would you mind creating a new issue to discuss this please :)

Issue #2459, opened 2017-03-31
Add tests for rules, which forgot specify `primaryOptionArray`


hudochenkov:
Ref: https://github.com/stylelint/stylelint/pull/2447#issuecomment-290276699

davidtheclark:
Closed by #2466.

PR #2466, opened 2017-04-01
Exemplify testing of primaryOptionArray

davidtheclark:
Relates to #2459, followup to #2447.

@hudochenkov, I think this is what should be done for every rule with `primaryOptionArray`: ensure there is a test that would fail if the primary option were not interpreted correctly.

Without `rule.primaryOptionArray` for `function-ur-schema-whitelist`, this test will fail, because it relies on a correct interpretation of the array as the primary option. *With* `rule.primaryOptionArray` in that rule, the test passes.

All `primaryOptionArray` rules should have something similar.

### DIFF
--- a/lib/rules/at-rule-blacklist/__tests__/index.js
+++ b/lib/rules/at-rule-blacklist/__tests__/index.js
@@ -9,7 +9,7 @@ const rule = rules[ruleName]
 testRule(rule, {
   ruleName,
 
-  config: [[ "extend", "supports", "keyframes" ]],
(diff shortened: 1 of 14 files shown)


jeddy3:
> I think this is what should be done for every rule with primaryOptionArray: ensure there is a test that would fail if the primary option were not interpreted correctly.

SGTM.

> All primaryOptionArray rules should have something similar.

I added a commit that does this.

davidtheclark:
Excellent @jeddy3!

PR #2467, opened 2017-04-02
Add autofixing

davidtheclark:
This builds off of #2427, but I thought it would be cleaner and a little easier for me to create a new PR, modifying things as I copy over changes, instead of adding another commit over there. Much of these changes are lifted from that PR and slightly modified.

This illustrates a way I think we could go about implementing autofixing. Here are the key elements:

- New `fix` option in Node API, `--fix` in CLI.
- Third argument, `context`, passed to rule functions. This is an object so that we can continue adding things to it, as needed. Right now it has `fix: boolean`, indicating whether to fix or not, and `newline: \n | \r\n`, determined by the first newline found in the file, which we'll use to add new newlines when fixing.
- Code in `at-rule-name-case` and `comment-empty-line-before` that implements fixes for those rules.
- Code in `standalone.js` that overwrites the input file, when fixing, with fixed CSS.
- Modifications to `jest-setup.js` to check fixes if a test schema has `fix: true`. For `accept` tests, we always check that the fixed CSS equals the original. Every `reject` test must have a `fixed` property with the fixed CSS, and we check against that. All we need to add to tests, then, when we add fixing to a rule, is `fix: true` to each schema and `fixed: ..` to each `reject` test. This is illustrated in the modified rules.
- System test ensuring both that the PostCSS Result is fixed and that the original file is overwritten with fixed CSS.

I tried via the CLI in a little local real-life test case, too, and it seemed to work!

What do you think @stylelint/core? Does this seem like a promising start? Please try it locally and give it some serious scrutiny.

After we all agree on our methodology, we should also try it out in an editor plugin, probably Atom, to ensure that the way we're doing it will work in that context.

### DIFF
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -104,7 +104,7 @@ In particular, you will definitely want to use `validateOptions()` so that users
 
 ### Adding autofixing
 
-If you're sure it's possible to automatically fix all or some errors rule reports, you can add fixing functionality using [PostCSS API](http://api.postcss.org/) to modify PostCSS AST (Abstract Syntax Tree).
+Depending on the rule, it might be possible to automatically fix the rule's warnings by mutating the PostCSS AST (Abstract Syntax Tree) using the [PostCSS API](http://api.postcss.org/).
 
 Add `context` variable to rule parameters:
 
@@ -119,7 +119,7 @@ function rule(primary, secondary, context) {
 -   `fix`(boolean): If `true`, your rule can apply autofixes.
 -   `newline`(string): Line-ending used in current linted file.
 
-If you can write fixing funtionality, then change `root` using PostCSS API and don't report that error:
+If `context.fix` is `true`, then change `root` using PostCSS API and return early before `report()` is called.
 
 ```js
 if (context.fix) {
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -102,6 +102,34 @@ stylelint has a number of [utility functions](https://github.com/stylelint/style
 
 In particular, you will definitely want to use `validateOptions()` so that users are warned about invalid options. (Looking at other rules for examples of options validation will help a lot.)
 
+### Adding autofixing
+
+If you're sure it's possible to automatically fix all or some errors rule reports, you can add fixing functionality using [PostCSS API](http://api.postcss.org/) to modify PostCSS AST (Abstract Syntax Tree).
+
+Add `context` variable to rule parameters:
+
+```js
+function rule(primary, secondary, context) {
+  return (root, result) => {..}
+}
+```
+
+`context` is an object which could have two properties:
+
+-   `fix`(boolean): If `true`, your rule can apply autofixes.
+-   `newline`(string): Line-ending used in current linted file.
+
+If you can write fixing funtionality, then change `root` using PostCSS API and don't report that error:
+
+```js
+if (context.fix) {
+  // Apply fixes using PostCSS API
+  return // Return and don't report a problem
+}
+
+report(...)
+```
+
 ### Write tests
 
 Each rule must be accompanied by tests that contain:
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -41,12 +41,20 @@ global.testRule = (rule, schema) => {
         passingTestCases.forEach((testCase) => {
           const spec = (testCase.only) ? it.only : it
           spec(testCase.description || "no description", () => {
-            return stylelint({
+            const options = {
               code: testCase.code,
               config: stylelintConfig,
               syntax: schema.syntax,
-            }).then((output) => {
+            }
+            return stylelint(options).then((output) => {
               expect(output.results[0].warnings).toEqual([])
+              if (!schema.fix) return
+
+              // Check the fix
+              return stylelint(Object.assign({ fix: true }, options)).then((output) => {
+                const fixedCode = getOutputCss(output)
+                expect(fixedCode).toBe(testCase.code)
+              })
             })
           })
         })
@@ -58,11 +66,12 @@ global.testRule = (rule, schema) => {
         schema.reject.forEach((testCase) => {
           const spec = (testCase.only) ? it.only : it
           spec(testCase.description || "no description", () => {
-            return stylelint({
+            const options = {
               code: testCase.code,
               config: stylelintConfig,
               syntax: schema.syntax,
-            }).then((output) => {
+            }
+            return stylelint(options).then((output) => {
               const warning = output.results[0].warnings[0]
 
               expect(testCase).toHaveMessage()
@@ -76,10 +85,29 @@ global.testRule = (rule, schema) => {
               if (testCase.column !== undefined) {
                 expect(_.get(warning, "column")).toBe(testCase.column)
               }
+
+              if (!schema.fix) return
+
+              if (!testCase.fixed) {
+                throw new Error("If using { fix: true } in test schema, all reject cases must have { fixed: .. }")
+              }
+
+              // Check the fix
+              return stylelint(Object.assign({ fix: true }, options)).then((output) => {
+                const fixedCode = getOutputCss(output)
+                expect(fixedCode).toBe(testCase.fixed)
+              })
             })
           })
         })
       })
     }
   })
+}
+
+function getOutputCss(output) {
+  const result = output.results[0]._postcssResult
+  return result.root.toString(result.opts.syntax)
+    // Less needs us to manually strip whitespace at the end of single-line comments ¯\_(ツ)_/¯
+    .replace(/(\n?\s*\/\/.*?)[ \t]*(\r?\n)/g, "$1$2")
 }
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -1,6 +1,7 @@
 "use strict"
 
 const _ = require("lodash")
+const less = require("postcss-less")
 const basicChecks = require("./lib/testUtils/basicChecks")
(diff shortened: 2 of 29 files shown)


jest-setup.js
```diff
@@ -76,10 +85,29 @@ global.testRule = (rule, schema) => {
...
       })
     }
   })
 }
+
+function getOutputCss(output) {
+  const result = output.results[0]._postcssResult
+  return result.root.toString(result.opts.syntax)
+    // Less needs us to manually strip whitespace at the end of single-line comments ¯\_(ツ)_/¯
+    .replace(/(\n?\s*\/\/.*?)[ \t]*(\r?\n)/g, "$1$2")
```
hudochenkov: Let's narrow this function to Less syntax only. I. e. apply this transformation only if test case has `syntax: "less"`. This function could potentially mess with our tests results and give us false information.

jeddy3:
> Does this seem like a promising start?

Very much so!

I think the simplicity of the changes to the rule code _and tests_ is a big win here. It lowers the barrier to contributing autofixing PRs.

> Please try it locally

Locally it worked great.

> After we all agree on our methodology, 

This route is looking very promising. The only thing that springs to mind is have we considered how two rules changing the same piece of CSS will work i.e. the "waves" approach in the automutate PoC?

hudochenkov:
> This route is looking very promising. The only thing that springs to mind is have we considered how two rules changing the same piece of CSS will work i.e. the "waves" approach in the automutate PoC?

How rules are working currently? One by one or in parallel? If one by one, then there is no problem. If in parallel, then there could be some problems. I don't know if two PostCSS plugins could work simultaneously on one AST.

If rules works in parallel, then we could force them in sync mode if fix enabled, to be safe.

jeddy3:
OK, scratch that. I was thinking about [something else](https://github.com/postcss/postcss/issues/754#issuecomment-192714884), but I don't think it's relevant here.

lib/rules/comment-empty-line-before/index.js
```diff
@@ -116,6 +117,18 @@ const rule = function (expectation, options) {
         return
       }
 
+      // Fix
+      if (context.fix) {
+        if (expectEmptyLineBefore) {
+          comment.raws.before = /\n/.test(comment.raws.before)
```
davidtheclark: 👍

jest-setup.js
davidtheclark: 👍

davidtheclark:
Great suggestions @hudochenkov. I'll try to get those fixes in soon.

In order to get this ready for release, we'll need documentation. Does anybody feel like contributing that to speed things along?

hudochenkov:
I'll try to write docs tomorrow (on Sunday).

BTW, are we going to introduce this feature in `8.0`? Then this PR should be based on `v8` branch. Which is out of date with master a little bit.

jeddy3:
> BTW, are we going to introduce this feature in 8.0?

It's non-breaking and so I think we can introduce this feature in the next minor release. Is it worth flagging it as experimental?

hudochenkov:
It's not non-breaking change, but it's huge deal! I think it deserves to be in major release. Also minor releases less noticeable by users then major releases.

davidtheclark:
@hudochenkov: If we don't get it done before v8, then it goes out in v8; but I don't think that round version numbers would be a good reason to delay the experiment.

davidtheclark:
@hudochenkov: Addressed your comments.

lib/rules/comment-empty-line-before/index.js
```diff
@@ -116,6 +117,16 @@ const rule = function (expectation, options) {
         return
       }
 
+      // Fix
+      if (context.fix) {
+        if (expectEmptyLineBefore) {
+          addEmptyLineBefore(comment, context.newline)
+        } else {
+          comment.raws.before = comment.raws.before.replace(/(\r?\n\s*\r?\n)+/g, context.newline)
```
hudochenkov: Could you, please, create a new helper function `removeEmptyLineBefore`? I didn't think about it at first, sorry.

hudochenkov:
@davidtheclark nice! Thank you!

hudochenkov:
I was looking to ESLint docs and found [interesting thing about --fix](http://eslint.org/docs/user-guide/command-line-interface#--fix):

> Not all problems are fixable using this option, and the option does not work in these situations:
> 
> 1. This option throws an error when code is piped to ESLint.
> 2. This option has no effect on code that uses processors.

I think we should also disable autofixing in these situations.

hudochenkov:
I've just added docs. What should I change? What I've missed?

lib/rules/comment-empty-line-before/index.js
davidtheclark: This can be saved for the next time we need it.

davidtheclark:
I may not have time to work on this over the next week. If anybody wants to carry it through to completion, feel free; otherwise I'll try to get back to it in a week or two.

hudochenkov:
@davidtheclark I want to carry it through. What do you think should be done before merge?

davidtheclark:
@hudochenkov 
- [ ] Finish documentation (which it sounds like you worked on)
- [ ] Make any code changes that are necessary

That's it, I think.

hudochenkov:
I've added new utility `removeEmptyLinesBefore` for removing all empty lines before. Will be useful for *-empty-line-before rules.

@jeddy3 you're great writer. Could you, please, look at docs and check if everything ok?

@stylelint/core It's a big feature. Could you, please, take a look at it? I think we are ready to merge as soon docs are good.

docs/developer-guide/rules.md
```diff
@@ -102,6 +102,34 @@ stylelint has a number of [utility functions](https://github.com/stylelint/style
 
 In particular, you will definitely want to use `validateOptions()` so that users are warned about invalid options. (Looking at other rules for examples of options validation will help a lot.)
 
+### Adding autofixing
+
+If you're sure it's possible to automatically fix all or some errors rule reports, you can add fixing functionality using [PostCSS API](http://api.postcss.org/) to modify PostCSS AST (Abstract Syntax Tree).
```
jeddy3: "Depending on the rule, it might be possible to automatically fix the rule's warnings by mutating the PostCSS AST (Abstract Syntax Tree) using the [PostCSS API](http://api.postcss.org/).

Determining the feasibility of autofixing a rule's warning is the responsibility of us all during the issue stage. It's not the just the rule author's responsibility.

docs/developer-guide/rules.md
```diff
@@ -102,6 +102,34 @@ stylelint has a number of [utility functions](https://github.com/stylelint/style
...
+  return (root, result) => {..}
+}
+```
+
+`context` is an object which could have two properties:
+
+-   `fix`(boolean): If `true`, your rule can apply autofixes.
+-   `newline`(string): Line-ending used in current linted file.
+
+If you can write fixing funtionality, then change `root` using PostCSS API and don't report that error:
```
jeddy3: If `context.fix` is `true`, then change `root` using PostCSS API and return early before the `report()` is called.

"If you can write fixing funtionality" repeat of above.

docs/developer-guide/rules.md
hudochenkov: Second sentence it's a sentence for docs or just a comment?

hudochenkov:
@jeddy3 thank you! I've made changes. But have one question: https://github.com/stylelint/stylelint/pull/2467#discussion_r111195024.

docs/developer-guide/rules.md
jeddy3: A comment. Sorry, I accidently missed off the closing quotation mark at the end of the first sentence.

hudochenkov:
@stylelint/core looks like everything is ready for merge. Any objections?

docs/developer-guide/rules.md
```diff
@@ -102,6 +102,34 @@ stylelint has a number of [utility functions](https://github.com/stylelint/style
...
+  return (root, result) => {..}
+}
+```
+
+`context` is an object which could have two properties:
+
+-   `fix`(boolean): If `true`, your rule can apply autofixes.
+-   `newline`(string): Line-ending used in current linted file.
+
+If `context.fix` is `true`, then change `root` using PostCSS API and return early before the `report()` is called.
```
jeddy3: "~the~ `report()`"

Soz, this mistake was my bad.

jeddy3:
> Any objections?

Nope. As this is marked as an experimental feature, I don't see any harm in getting in front of users ASAP. We can add something to the changelog about how this release lays the foundation and currently only two rules are autofixed.

hudochenkov:
I've merged it. Thank you, everyone! This feature is awesome! Don't release `7.11.0` yet. Give me day or two, I want to add autofixing for all `*-empty-line-before` rules.

Added to changelog:

    -   `at-rule-name-case`

@jeddy3 could you, please, change changelog if necessary?

davidtheclark:
Thanks for pushing this through @hudochenkov. I'm excited to see how it works with users.

jitendravyas:
I use grunt. Where to have `fix: true`? in gruntfile or stylelintrc file?

gruntfile
```
		stylelint: {
			all: ['app/styles/**/*.scss']
		},
```

hudochenkov:
@jitendravyas It's not released yet.

jitendravyas:
@hudochenkov It's released now :-) Can you help me with my previous question?

hudochenkov:
@jitendravyas :) I believe you should write to grunt plugin author about this ¯﻿\﻿_﻿(﻿ツ﻿)﻿_﻿/﻿¯

PR #2468, opened 2017-04-03
Add Flycheck to the list of editor plugins

PhilippBaschke:
Flycheck is a popular syntax checking extension for Emacs. It recently added support for stylelint.

See: https://github.com/flycheck/flycheck/pull/903

I did not create an issue because it is just a minor addition to the documentation. Hope that is OK 😌


jeddy3:
> Hope that is OK

Totally :)

> It recently added support for stylelint.

That's great to hear!

PR #2469, opened 2017-04-04
Made all instances of (deprecated) bold

jitendravyas:
I think it is better to keep them highlighted.


jitendravyas:
Why are these changes still not showing on the website? https://stylelint.io/user-guide/rules/

hudochenkov:
@jitendravyas stylelint.io updates, when stylelint itself updates. Because all info comes from stylelint package on npm. We haven't any release since your PR.

Issue #2458, opened 2017-03-30
Add selector-max-class rule


jimbo:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Propose adding a **new option** for [`selector-max-specificity`](https://stylelint.io/user-guide/rules/selector-max-specificity/)

- Propose naming the option `granular` and accepting possible values of `true|false`
- With `granular: true`, limit individual components of specificity rather than cumulative specificity

> Which rule, if any, is this issue related to?

[`selector-max-specificity`](https://stylelint.io/user-guide/rules/selector-max-specificity/)

> What CSS is needed to reproduce this issue?

```css
.one .two .three .four .five .six .seven .eight .nine .ten {
  color: pink;
}
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "selector-max-specificity": "1,3,1"
  }
}
```

> Which version of stylelint are you using?

`latest` (`7.9.0`)

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

Stylelint works as expected. However, if my configuration allows for selectors to contain at least one id, I expect there to be some way to limit the number of classnames and/or tagnames.

> What actually happened (e.g. what warnings or errors you are getting)?

Without a new option, the rule is (correctly) not violated because a selector with ten classnames has a lower *cumulative* specificity than the configured maximum, `"1,3,1"`, which represents a selector with one id. As far as I know, there's currently no way to alter this behavior.

---

As it happens, I didn't discover the [rules](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md) until I'd already forked the library and implemented the option. According to the rules, we're supposed to discuss the option and its name here first, but I can put up a PR whenever.

hudochenkov:
I think it's a clear candidate for a [plugin](https://stylelint.io/developer-guide/plugins/), because this option doesn't have a clear and unambiguous finished state, and, I believe, won't be used by majority of stylelint's users.

jeddy3:
@jimbo Thanks for taking the time to writing up a proposal.

It took me a bit to get my head around it, and I'm not sure I understand it completely yet. Am I right in thinking that the way the `selector-max-specificity` currently works is in line with [with the spec](https://drafts.csswg.org/selectors/#specificity-rules), and what you're proposing is some adaptation of that?

[...]

PR #2474, opened 2017-04-06
Add `selector-max-class` rule

jimbo:
This rule verifies that the number of classes in a selector doesn't exceed some maximum. Rather than measuring specificity, it counts classes explicitly.

#2458

- ~~I'll add a README once we agree this is ready~~
- Based this rule on [`selector-max-compound-selectors`](https://stylelint.io/user-guide/rules/selector-max-compound-selectors/), since the operation is similar
- Could write a util, `isFullyResolvedSelector()`, but I'm not sure I want to write a bunch of tests for it 😅

### DIFF
--- a/docs/user-guide/example-config.md
+++ b/docs/user-guide/example-config.md
@@ -132,6 +132,7 @@ You might want to learn a little about [how rules are named and how they work to
     "selector-list-comma-space-after": "always"|"never"|"always-single-line"|"never-single-line",
     "selector-list-comma-space-before": "always"|"never"|"always-single-line"|"never-single-line",
     "selector-max-empty-lines": int,
+    "selector-max-class": int,
     "selector-max-compound-selectors": int,
     "selector-max-specificity": string,
     "selector-nested-pattern": string,
--- a/lib/rules/selector-max-class/README.md
+++ b/lib/rules/selector-max-class/README.md
@@ -1,0 +1,51 @@
+# selector-max-class
+
+Limit the number of classes in a selector.
+
+```css
+div .foo.bar[data-val] > a.baz {}
+/*  ↑   ↑                 ↑
+    |   |                 |
+    1   2                 3  -- this selector contains three classes */
+```
+
+This rule resolves nested selectors before counting the number of classes in a selector. Each selector in a [selector list](https://www.w3.org/TR/selectors4/#selector-list) is evaluated separately.
+
+The `:not()` pseudo-class is also evaluated separately. The rule processes the argument as if it were an independent selector, and the result does not count toward the total for the entire selector.
+
+## Options
+
+`int`: Maximum classes allowed.
+
+For example, with `2`:
+
+The following patterns are considered warnings:
+
+```css
+.foo.bar.baz {}
+```
+
+```css
+.foo .bar {
+  & > .baz {}
+}
+```
+
+The following patterns are *not* considered warnings:
+
+```css
+div {}
+```
+
+```css
+.foo .bar {}
+```
+
+```css
+.foo.bar,
+.lorem.ipsum {} /* each selector in a selector list is evaluated separately */
+```
+
+```css
+.foo .bar :not(.lorem.ipsum) {} /* `.lorem.ipsum` is inside `:not()`, so it is evaluated separately */
+```
--- a/lib/rules/selector-max-class/__tests__/index.js
+++ b/lib/rules/selector-max-class/__tests__/index.js
@@ -1,0 +1,139 @@
+"use strict"
+
+const messages = require("..").messages
+const ruleName = require("..").ruleName
+const rules = require("../../../rules")
+
+const rule = rules[ruleName]
+
+// Sanity checks
+testRule(rule, {
+  ruleName,
+  config: [0],
+
+  reject: [{
+    code: ".foo { left: 0; top: 0; }",
+    description: "disallow classes",
+    message: messages.expected(".foo", 0),
+    line: 1,
+    column: 1,
+  }],
+})
+
+// Standard tests
+testRule(rule, {
+  ruleName,
+  config: [2],
+
+  accept: [ {
+    code: ".ab { left: 0; top: 0; }",
+    description: "fewer than max classes",
+  }, {
+    code: ".ab.cd { left: 0; top: 0; }",
+    description: "exactly max classes",
+  }, {
+    code: ".ab .cd { left: 0; top: 0; }",
+    description: "compound selector",
+  }, {
+    code: ".ab, \n.cd { left: 0; top: 0; }",
+    description: "multiple selectors: fewer than max classes",
+  }, {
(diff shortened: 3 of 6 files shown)


lib/rules/selector-max-class/__tests__/index.js
```diff
@@ -0,0 +1,130 @@
...
+const rules = require("../../../rules")
+
+const rule = rules[ruleName]
+
+testRule(rule, {
+  ruleName,
+  config: [2],
+
+  accept: [ {
+    code: ".cherry { left: 0; top: 0; }",
```
evilebottnawi: Best use `foo`, `bar` classes for tests :smile:

lib/rules/selector-max-class/__tests__/index.js
```diff
@@ -0,0 +1,130 @@
...
+
+const messages = require("..").messages
+const ruleName = require("..").ruleName
+const rules = require("../../../rules")
+
+const rule = rules[ruleName]
+
+testRule(rule, {
+  ruleName,
+  config: [2],
```
evilebottnawi: reoder test first with config `0`, when `1`, when `2` and last `less`, `scss` and etc

lib/rules/selector-max-class/__tests__/index.js
```diff
@@ -0,0 +1,130 @@
...
+  }, {
+    code: ".azure.cyan.aqua :not(.teal) { left: 0; top: 0; }",
+    description: ":not(): greater than max classes, outside",
+    message: messages.expected(".azure.cyan.aqua :not(.teal)", 2),
+    line: 1,
+    column: 1,
+  }, {
+    code: ".orange, .yellow { &:hover > .amber.gold { left: 0; top: 0; } }",
+    description: "nested selectors: greater than max classes",
+    message: messages.expected(".orange:hover > .amber.gold", 2),
```
evilebottnawi: Always add `line` and `column`

evilebottnawi:
@jimbo seems good!

jimbo:
@efedorenko @hudochenkov Thanks for the feedback. I've updated the PR and added the README. 👍

lib/rules/selector-max-class/README.md
```diff
@@ -0,0 +1,51 @@
+# selector-max-compound-selectors
```
hudochenkov: Wrong name :)

lib/rules/selector-max-class/README.md
```diff
@@ -0,0 +1,51 @@
...
+```css
+   div .foo.bar[data-val] > a.baz {}
+/*     ↑   ↑                 ↑
+       |   |                 |
+       1   2                 3  -- this selector contains three classes */
+```
+
+This rule resolves nested selectors before counting the number of classes in a selector. Each selector in a [selector list](https://www.w3.org/TR/selectors4/#selector-list) is evaluated separately.
+
+The `:not()` pseudo-class is also evaluated separately. The rule processes the argument as if it were an independent selector, and the result does not count toward the total for the entire selector.
```
hudochenkov: The rule processes the argument as if it ~were~was an independent selector...

lib/rules/selector-max-class/__tests__/index.js
```diff
@@ -0,0 +1,136 @@
...
+})
+
+// SCSS tests
+testRule(rule, {
+  ruleName,
+  config: [1],
+  syntax: "scss",
+
+  accept: [ {
+    code: ".foo #{$test} { left: 0; top: 0; }",
```
hudochenkov: Could you add `.foo .bar #{$test}` also. Theoretically it should be in `reject`, but based on code it would be ignored by rule.

lib/rules/selector-max-class/README.md
jimbo: Oops. 😅

lib/rules/selector-max-class/README.md
jimbo: I'm no English major, but it looks right to me. [Subjunctive mood](https://en.wikipedia.org/wiki/English_subjunctive#Distinguishing_from_past_indicative_after_if), I think.

lib/rules/selector-max-class/__tests__/index.js
jimbo: Sure. I've noticed a lot of these rules just bail out in the case of nonstandard syntax.

jimbo:
@hudochenkov I've updated it again. 👍

lib/rules/selector-max-class/README.md
hudochenkov: I can't argue with you, because my English is not good enough. Thought it was a typo.

lib/rules/selector-max-class/README.md
```diff
@@ -0,0 +1,51 @@
+# selector-max-class
+
+Limit the number of classes in a [complex selector](https://www.w3.org/TR/selectors4/#complex).
```
jeddy3: Very minor nitpick - let's change this to match the one in the rules list i.e.

"Limit the number of classes in a selector."

lib/rules/selector-max-class/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+```css
+   div .foo.bar[data-val] > a.baz {}
```
jeddy3: Let's shift this so the `<div>` is flush to the left. We only indent when there's not space for the arrow.

There's a tonne of documentation in stylelint and so we try to keep it ship-shape :)

jimbo: Makes sense. I've updated the docs. 👍

jeddy3:
@evilebottnawi Are you happy your requests have been addressed?

docs/user-guide/example-config.md
```diff
@@ -132,6 +132,7 @@ You might want to learn a little about [how rules are named and how they work to
     "selector-list-comma-space-after": "always"|"never"|"always-single-line"|"never-single-line",
     "selector-list-comma-space-before": "always"|"never"|"always-single-line"|"never-single-line",
     "selector-max-empty-lines": int,
+    "selector-max-class": int,
```
davidtheclark: Sorry I'm chiming in late in the game here ... I was going to review this code and then merge, but had at last one or two notes. First one: Should we rename this to `selector-max-classes` (plural)? That seems to fit the convention better.

jeddy3:
@jimbo Thanks very much for this rule!

Changelog:

> -   Added: `selector-max-class` rule.

jimbo:
@jeddy3 @evilebottnawi @hudochenkov 

Thanks for digging in and helping me get this across the finish line!

evilebottnawi:
@jeddy3 can we do todo list with all rules `selector-max-*`?

jeddy3:
Yep, I was hoping to make time to do an audit of the `-no-` rules today.

Issue #2397, opened 2017-02-26
Add MathML tags to list of known tags in selector-type-no-unknown


jeddy3:
Ref: https://github.com/stylelint/stylelint/issues/2326#issuecomment-278416740

evilebottnawi:
@jeddy3 sgtm

evilebottnawi:
http://caniuse.com/#search=mathml spec status `recommendation` change labels

evilebottnawi:
Done in https://github.com/stylelint/stylelint/pull/2478

PR #2478, opened 2017-04-07
Fixed: `selector-type-no-unknown` now ignores MathML tags.

evilebottnawi:

#2397 

No, it's self explanatory.

### DIFF
--- a/lib/rules/selector-type-no-unknown/README.md
+++ b/lib/rules/selector-type-no-unknown/README.md
@@ -8,7 +8,7 @@ Disallow unknown type selectors.
  * This type selector */
 ```
 
-This rule considers tags defined in the HTML and SVG Specifications to be known.
+This rule considers tags defined in the HTML, SVG, and MathML specifications to be known.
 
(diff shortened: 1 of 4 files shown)


lib/rules/selector-type-no-unknown/README.md
```diff
@@ -8,7 +8,7 @@ Disallow unknown type selectors.
  * This type selector */
 ```
 
-This rule considers tags defined in the HTML and SVG Specifications to be known.
+This rule considers tags defined in the HTML and SVG and MathML Specifications to be known.
```
hudochenkov: I think this would be better:

> This rule considers tags defined in the HTML, SVG, and MathML specifications to be known.

evilebottnawi:
@hudochenkov done :smile:

jeddy3:
Changelog:

> -   Fixed: `selector-type-no-unknown` now ignores MathML tags ([#2478](https://github.com/stylelint/stylelint/pull/2478)).

Issue #2481, opened 2017-04-07
False negatives for vendor-prefixed values in value lists in function-linear-gradient-no-nonstandard-direction


pfructuoso:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

It´s a bug. [False positives and negatives when vendor-prefixed values are used in function-linear-gradient-no-nonstandard-direction](https://github.com/stylelint/stylelint/issues/1939) is still happening when using multiple values for `background-image`

> Which rule, if any, is this issue related to?

`function-linear-gradient-no-nonstandard-direction`

 > What CSS is needed to reproduce this issue?

```css
   .node-preview-backlink {
     /* false negative */
     background: url(image.png), -webkit-linear-gradient(top, #012345, #ABCDEF);

     /* Working as expected */
     background: -webkit-linear-gradient(top, #012345, #ABCDEF), url(image.png);
   }
```

> What stylelint configuration is needed to reproduce this issue?

```json
   {
"rules": {
       "function-linear-gradient-no-nonstandard-direction": true
       }
}
```

> Which version of stylelint are you using?

`7.10.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI interface.
`stylelint test.css`

 > Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No, it´s standard syntax

> What did you expect to happen?

Both cases on the example have the same code so no warnings should be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

In "false negative" the following warning wa flagged:

```shell
   test.css
3:55  ✖  Unexpected nonstandard direction   function-linear-gradient-no-nonstandard-direction
```

I tried to figure out what is happening by myself and it looks like in
´lib/rules/function-linear-gradient-no-nonstandard-direction/index.js´

jeddy3:
@pfructuoso Thank you for the report and PR!

pfructuoso:
Can someone with better git-foo than I advise on how best to resolve the problem with the PR #2482? Thanks.

davidtheclark:
Closed by #2496.

PR #2482, opened 2017-04-07
Fix rule for linear-gradient with vendor prefix multivalue #2481

pfructuoso:

#2481 

It just splits the multiple properties to check them individualy


lib/rules/function-linear-gradient-no-nonstandard-direction/__tests__/index.js
```diff
@@ -50,6 +50,33 @@ testRule(rule, {
     code: ".foo { background: -moz-linear-gradient(top, #fff, #000); }",
   }, {
     code: ".foo { background: -o-linear-gradient(top, #fff, #000); }",
+  }, {
+    code: ".foo { background: url(foo.png), -webkit-linear-gradient(bottom, #fff, #000 ); }",
+    message: messages.rejected,
```
jeddy3: `accept` tests shouldn't take a `message` propertry

lib/rules/function-linear-gradient-no-nonstandard-direction/__tests__/index.js
```diff
@@ -97,5 +124,32 @@ testRule(rule, {
   }, {
     code: ".foo { background: -o-linear-gradient(to bottom, #fff, #000 ); }",
     message: messages.rejected,
+  }, {
+    code: ".foo { background: url(foo.png), -webkit-linear-gradient(to bottom, #fff, #000 ); }",
+    message: messages.rejected,
```
jeddy3: Please add `line` and `column` properties to all the `reject` tests - [see here](https://github.com/stylelint/stylelint/blob/master/lib/rules/at-rule-blacklist/__tests__/index.js#L25-L27) as an example.

lib/rules/function-linear-gradient-no-nonstandard-direction/index.js
```diff
@@ -37,40 +38,45 @@ const rule = function (actual) {
...
-            index: expressionIndex,
-            result,
-            ruleName,
-          })
-        }
+          function complain() {
+            report({
+              message: messages.rejected,
+              node: decl,
+              index: expressionIndex,
```
jeddy3: I suspect you'll need to calculate the index based on the `expressionIndex` _and the index of the `postcss-value-parser` node_.

pfructuoso:
@jeddy3 What do you mean by merging it to `v8` branch? Creating another PR for ´v8´ branch? I see it as a breaking change. The index problem is solved and it´s the same it was before. Nevertheless, if you need it or whatever other reason, just ask me again to do it and you´ll have it the faster possible!

lib/rules/function-linear-gradient-no-nonstandard-direction/index.js
```diff
@@ -37,40 +38,45 @@ const rule = function (actual) {
...
-            index: expressionIndex,
-            result,
-            ruleName,
-          })
-        }
+          function complain() {
+            report({
+              message: messages.rejected,
+              node: decl,
+              index: decl.toString().indexOf(valueNode.value) + expressionIndex,
```
jeddy3: I think you'll want to make use of the [`declarationValueIndex` util](https://github.com/stylelint/stylelint/blob/master/lib/utils/declarationValueIndex.js) and the  [`node`'s `sourceIndex`](https://github.com/TrySound/postcss-value-parser#nodes) for consistency with the other rules, [like so](https://github.com/stylelint/stylelint/blob/master/lib/rules/declaration-property-unit-blacklist/index.js#L66), e.g.:

```js
index: declarationValueIndex(decl) + valueNode.sourceIndex + expressionIndex
```

pfructuoso:
> The column and line numbers look good. I've made one further request to make the calculation of the index more consistent with other rules.

I like declarationValueIndex(decl)! I was looking for something like it but didn´t find it.

> I don't think you need to create another PR as you can use github to change the base branch. The v8 branch is for breaking changes. So, if possible, please set v8 as the base branch for this PR.

This is my first base branch change and now we have conflics. I don´t feel confortable trying to solve this in other branch without supervision. Could you guide me with this?

jeddy3:
@pfructuoso Thanks for making the changes. They look good to me.

> I like declarationValueIndex(decl)! I was looking for something like it but didn´t find it.

Good stuff!

> This is my first base branch change and now we have conflics. I don´t feel confortable trying to solve this in other branch without supervision. Could you guide me with this?

@stylelint/core Can someone with better git-foo than I advise on how best to resolve this? Delete this branch and then just port @pfructuoso changes into a new one, or something else?

pfructuoso:
> @stylelint/core Can someone with better git-foo than I advise on how best to resolve this? Delete this branch and then just port @pfructuoso changes into a new one, or something else?

I don´t know how to solve it without creating a new PR. I created a new [branch](https://github.com/pfructuoso/stylelint/tree/multivalue-linear-gradient-v8) from v8 and apply my changes but I can´t change the origin branch.

hudochenkov:
@pfructuoso try to rename your old branch to something else, then rename new branch to `multivalue-linear-gradient` and force push.

pfructuoso:
Ok, I did. The [branch](https://github.com/pfructuoso/stylelint/tree/multivalue-linear-gradient) exists but previous messages say it´ve been deleted and the PR closed. The reopen button is disabled. Is it only me or this doesn´t look good? Is it any problem on creating a new PR and referal it to this (#2482)?
screenshot

hudochenkov:
@pfructuoso I'm so sorry. I believe I had same issues and did what I told you you to do. Maybe Github changed something and now behaviour is different.

hudochenkov:
I've checked it's possible to open a new PR with base `v8`. But let's wait what others can say. I don't want to let you down again :(

Issue #2492, opened 2017-04-11
Error when linting CSS string with cache enabled.


olegskl:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

An error is thrown when trying to lint a CSS string while `cache` option is set to `true.`

> Which rule, if any, is this issue related to?

Not applicable.

> What CSS is needed to reproduce this issue?

Any CSS string.

> What stylelint configuration is needed to reproduce this issue?
```json
{
  "code": ".foo {}",
  "codeFilename": "foo.css",
  "cache": true
}
```

> Which version of stylelint are you using?

`7.10.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

```js
const stylelint = require('stylelint');

stylelint.lint({
  code: '.foo {}',
  codeFilename: 'foo.css',
  cache: true,
  config: {rules: []}
}).catch(console.error);
```

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No.

> What did you expect to happen?

Lint should work. Not sure about caching (probably should be ignored?).

> What actually happened (e.g. what warnings or errors you are getting)?

```shell
TypeError: Cannot read property 'reconcile' of undefined
    at prepareReturnValue (...\node_modules\stylelint\lib\standalone.js:174:16)
    at stylelint._lintSource.then.catch.then.stylelintResult (...\node_modules\stylelint\
lib\standalone.js:90:14)
```

evilebottnawi:
@olegskl thanks for issue :+1:

sergesemashko:
thanks @olegskl, I'll take care of it.

davidtheclark:
Closed by #2494.

PR #2494, opened 2017-04-12
Fix error when linting CSS string with cache enabled (issue #2492)

sergesemashko:

Issue #2492

it's self explanatory

@jeddy3, @davidtheclark, @evilebottnawi, please take a look when you have time.


davidtheclark:
Added to changelog:

> -   Fixed: no error is thrown when linting a string with `cache` enabled ([#2494](https://github.com/stylelint/stylelint/pull/2494)).

PR #1594, opened 2016-07-05
Require NodeJS 4.x LTS

ntwb:
Closes #1580

PR #2497, opened 2017-04-13
chore: Drop NodeJS 4.x and require at least NodeJS 6.x LTS

ntwb:
Previously #1594 Require NodeJS 4.x LTS

NodeJS 4.x went into "maintenance mode" April 1st

NodeJS 6.x is now the "active" LTS version

See https://github.com/nodejs/LTS#lts-schedule

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,6 @@ matrix:
     - node_js: '6'
       script: npm run jest -- --runInBand
       env: CI=tests 6
-    - node_js: '4'
-      script: npm run jest -- --runInBand
-      env: CI=tests 4
-      sudo: required
     - node_js: '7'
       script: npm run jest -- --runInBand  --coverage
       env: CI=coverage
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "!**/__tests__"
   ],
   "engines": {
-    "node": ">=4.2.1"
+    "node": ">=6"
   },
   "dependencies": {
     "autoprefixer": "^6.0.0",


hudochenkov:
It might lead to breaking changes, because we'll stop test for Node 4, so it's safer to base this change on `v8` branch.

davidtheclark:
I really don't want to do this. We don't have any reason to exclude people who are still on Node 4, do we?

hudochenkov:
@davidtheclark I see no reasons to do this. Seems like lots of packages dropped support for Node <4 in the second half of 2016. To early to drop Node 4, I think.

jeddy3:
> We don't have any reason to exclude people who are still on Node 4, do we?

Agreed. Let's only consider this when there's a `node@6` feature we really can't do without.

shellscape:
Jumping into this one a bit late but.... there's a myriad of functional, language, and performance updates that benefit users of newer versions of node, as can be seen here: http://node.green/. Node v4 was first released *_in 2015_*. Not wanting to exclude users that cannot update is noble, but those users have to grapple with projects moving onto to new versions. Bumping the engine to v6 can be a major version change, which won't affect those users stuck in 2015 😄

shellscape:
Adding to this conversation, a few links outlining the benefits of the newer version, plus a fairly influential article from Wikimedia on the topic:

http://www.infoworld.com/article/3061213/javascript/nodejs-6-taps-v8-update-for-more-speed-security.html
https://blog.risingstack.com/whats-new-in-node-v6/
https://blog.wikimedia.org/2017/02/17/node-6-wikimedia/

evilebottnawi:
@shellscape many of our users use `node` v4, `sad but true`.

PR #2500, opened 2017-04-13
Add autofixing for *-empty-line-before rules

hudochenkov:
- [x] `at-rule-empty-line-before`
- [x] `custom-property-empty-line-before`
- [x] `declaration-empty-line-before`
- [x] `rule-empty-line-before`

`comment-empty-line-before` done in initial autofix PR #2467.

I didn't do `block-closing-brace-empty-line-before`, because it's quite different from rules above. Also it's not possible to use [`addEmptyLineBefore`](https://github.com/stylelint/stylelint/blob/1cdd2469a63d6b9751bce191c1478b4893fc6679/lib/utils/addEmptyLineBefore.js) and [`removeEmptyLinesBefore`](https://github.com/stylelint/stylelint/blob/1cdd2469a63d6b9751bce191c1478b4893fc6679/lib/utils/removeEmptyLinesBefore.js) utils, because they change `node.raws.before`, but for this rule `node.raws.after` should be changed.

I'm not sure why [`result` returns in `comment-empty-line-before`](https://github.com/stylelint/stylelint/blob/1cdd2469a63d6b9751bce191c1478b4893fc6679/lib/rules/comment-empty-line-before/index.js#L128). I've tried `return` and `return result`, tests were passing every time. I decided to use just `return`.

### DIFF
--- a/lib/rules/at-rule-empty-line-before/README.md
+++ b/lib/rules/at-rule-empty-line-before/README.md
@@ -12,6 +12,8 @@ a {}
 
 If the at-rule is the very first node in a stylesheet then it is ignored.
 
+The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofixing-errors) can automatically fix some of the problems reported by this rule.
+
 ## Options
 
 `string`: `"always"|"never"`
--- a/lib/rules/at-rule-empty-line-before/README.md
+++ b/lib/rules/at-rule-empty-line-before/README.md
@@ -12,7 +12,7 @@ a {}
 
 If the at-rule is the very first node in a stylesheet then it is ignored.
 
-The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofixing-errors) can automatically fix some of the problems reported by this rule.
+The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofixing-errors) can automatically fix all of the problems reported by this rule.
 
 ## Options
 
--- a/lib/rules/at-rule-empty-line-before/__tests__/index.js
+++ b/lib/rules/at-rule-empty-line-before/__tests__/index.js
@@ -42,33 +42,43 @@ const sharedAlwaysTests = {
 
   reject: [ {
     code: "a {} @media {}",
+    fixed: "a {}\n\n @media {}",
     message: messages.expected,
   }, {
     code: "a {} @mEdIa {}",
+    fixed: "a {}\n\n @mEdIa {}",
     message: messages.expected,
   }, {
     code: "a {} @MEDIA {}",
+    fixed: "a {}\n\n @MEDIA {}",
     message: messages.expected,
   }, {
     code: "@keyframes foo {} @media {}",
+    fixed: "@keyframes foo {}\n\n @media {}",
     message: messages.expected,
   }, {
     code: "@-webkit-keyframes foo {} @media {}",
+    fixed: "@-webkit-keyframes foo {}\n\n @media {}",
     message: messages.expected,
   }, {
     code: "@-webkit-keyframes foo {} @-webkit-keyframes bar {}",
+    fixed: "@-webkit-keyframes foo {}\n\n @-webkit-keyframes bar {}",
     message: messages.expected,
   }, {
     code: "a {}\n@media {}",
+    fixed: "a {}\n\n@media {}",
     message: messages.expected,
   }, {
     code: "a {}\r\n@media {}",
+    fixed: "a {}\r\n\r\n@media {}",
     message: messages.expected,
   }, {
     code: "a {}\n\n/* comment */\n@media {}",
+    fixed: "a {}\n\n/* comment */\n\n@media {}",
     message: messages.expected,
   }, {
     code: "a {}\r\n\r\n/* comment */\r\n@media {}",
+    fixed: "a {}\r\n\r\n/* comment */\r\n\r\n@media {}",
     message: messages.expected,
   } ],
 }
@@ -92,9 +102,19 @@ const sharedNeverTests = {
 
   reject: [ {
     code: "a {}\n\n@media {}",
+    fixed: "a {}\n@media {}",
+    message: messages.rejected,
+  }, {
+    code: "a {}\r\n\r\n@media {}",
+    fixed: "a {}\r\n@media {}",
     message: messages.rejected,
   }, {
     code: "@keyframes foo {}\n/* comment */\n\n@media {}",
+    fixed: "@keyframes foo {}\n/* comment */\n@media {}",
+    message: messages.rejected,
+  }, {
+    code: "@keyframes foo {}\r\n/* comment */\r\n\r\n@media {}",
+    fixed: "@keyframes foo {}\r\n/* comment */\r\n@media {}",
     message: messages.rejected,
   } ],
 }
@@ -102,6 +122,7 @@ const sharedNeverTests = {
 testRule(rule, mergeTestDescriptions(sharedAlwaysTests, {
   ruleName,
   config: ["always"],
+  fix: true,
 
   accept: [{
     code: "a {\n\n  @mixin foo;\n}",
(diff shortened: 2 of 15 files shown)


hudochenkov:
Couldn't be merged until #2501 is fixed. Labeling as `status: blocked`.

hudochenkov:
Thanks to @davidtheclark tests are working consistently regardless of the platform now! This PR is ready for a review.

I thinks these rules are enough for the new version to be released.

lib/rules/at-rule-empty-line-before/README.md
```diff
@@ -12,6 +12,8 @@ a {}
 
 If the at-rule is the very first node in a stylesheet then it is ignored.
 
+The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofixing-errors) can automatically fix some of the problems reported by this rule.
```
davidtheclark: Can it only fix "some"? Not "all"?

lib/rules/at-rule-empty-line-before/__tests__/index.js
```diff
@@ -42,33 +42,43 @@ const sharedAlwaysTests = {
 
   reject: [ {
     code: "a {} @media {}",
+    fixed: "a {}\n\n @media {}",
```
davidtheclark: That extra space before `@` seems undesireable.

lib/rules/at-rule-empty-line-before/__tests__/index.js
```diff
@@ -244,33 +283,39 @@ testRule(rule, mergeTestDescriptions(sharedAlwaysTests, {
 
   reject: [ {
     code: "a {\n\n  @mixin foo;\n  color: pink;\n}",
+    fixed: "a {\n  @mixin foo;\n  color: pink;\n}",
     message: messages.rejected,
   }, {
     code: "a {\n\n  color: pink;\n\n  @mixin foo;\n}",
+    fixed: "a {\n\n  color: pink;\n  @mixin foo;\n}",
```
davidtheclark: Here the spaces before `@` make sense. I think we need to distinguish these cases. If there was a space after a newline — i.e. an indentation — preserve that; but do not preserve the space when replace a space with newlines.

lib/rules/at-rule-empty-line-before/__tests__/index.js
```diff
@@ -632,6 +720,13 @@ testRule(rule, {
       } @else-mixin {
       }
     `,
+    fixed: `
+      @if(true) {
+      }
+
+ @else-mixin {
```
davidtheclark: This looks like bad indentation.

lib/rules/declaration-empty-line-before/__tests__/index.js
```diff
@@ -32,46 +33,49 @@ testRule(rule, {
 
   reject: [ {
     code: "a { top: 15px; }",
+    fixed: "a {\n\n top: 15px; }",
```
davidtheclark: Problem here is that we're going to have to decide how to indent or not indent. I think keeping the space is not right, though: that space should be replaced by the newlines, not added to.

lib/rules/at-rule-empty-line-before/__tests__/index.js
hudochenkov: What if we use similar code? 

```
div {\n  a {}  @media {}\n}
```

Which is looks like:

```
div {
  a {}  @media {}
}
```

We have have two spaces before `a` and `@media`. In this example if we add `\n\n` before two spaces we get correct indentation. These two examples basically the same. There is some amount of spaces before node we're fixing, but we can't know if they are good of bad.

lib/rules/at-rule-empty-line-before/__tests__/index.js
hudochenkov: We can't know if it's indentation or not without interfering with other rules.

lib/rules/at-rule-empty-line-before/__tests__/index.js
hudochenkov: This rule is not about indentation.

lib/rules/declaration-empty-line-before/__tests__/index.js
hudochenkov: We have rules designed to do only their thing. We don't have overlapping rules. So why *-empty-line-before should care about anything except empty lines before? :) We have `indentation` rule to take care of indentation.

hudochenkov:
I think it's not enough to run stylelint only once to autofix errors.

Let's imagine we have autofixing for both rules:

```json
{
    "rules": {
        "indentation": 4,
        "rule-empty-line-before": "always"
    }
}
```

This is input:

```
a { b {} }
```

This is result (1 space before `b`):

```
a {\n\n b {} }
```

What is expected (4 spaces before `b`):

```
a {\n\n    b {} }
```

`indentation` runs first, it found no problems in input code, so it does nothing. Then runs `rule-empty-line-before` and add empty line before. And stylelint is done. But if we run stylelint without --fix flag after that we'll get new error from `indentation`. So we should run with `--fix` two times to eliminate all arrors.

For the first experimental release we're doing fine, in my opinion. But for a release after that we should address this problem. Something like that: with `--fix` run stylelint at least two times. If there errors from rules, which have autofixing run stylelint again. If errors are fixed in the last run, make another run, maybe new errors were introduced.

davidtheclark:
@hudochenkov: Your solution relies on somebody using the `indentation` rule. Maybe they aren't. If I weren't (because maybe I have wacky indentation preferences), it would be a real bummer to have my code automatically formatted in a way that looked terrible — that would be less helpful than *no* automatic formatting, because I'd have to go back and fix the fixing manually.

I agree that our rules each do their own thing, but I don't think that's an excuse to make an automatic fix create code that looks *worse* rather than *better*. If we can't make the code come out *better* after fixing, I don't think we should make the change.

I also agree that multiple runs might be necessary.

I wonder if this problem will crop up *mostly* with indentation. How could we avoid it? Here's a quick brainstorm (any other ideas?):

- Enforce that certain rules can only be autofixed if indentation is set in the config. Then we can use the indentation value when fixing.
- When indentation is not set, infer an indentation style based on the rest of the file, and use that.

hudochenkov:
I think cases like `a {} @media {}` are edge cases. In a real code base these rules most likely are on separate lines.

Also I believe the most of our users have `indent` rule enabled.

Yes, it's just assumptions. But I think they are very close to reality, especially the first one.

I propose don't overengineer this now. Autofixing is an experimental feature. Let's release it in a current state, because for the most use cases it will work great. And we'll see how users react to this feature and what we can improve.

More practical proposal. Let's add autofixing for `indentation` for the first release also. And since we can't run autofixing multiple times yet, we can put `indentation` at the very end of [`rules/index.js`](https://github.com/stylelint/stylelint/blob/master/lib/rules/index.js) so it will run after every *-empty-line-before rule and fix possible inconsistency. We may add a note to *-empty-line-before docs, that we recommend to enable `indentation` rule if user want to use autofixing.

davidtheclark:
> we'll see how users react to this feature and what we can improve

If we already know what needs to be fixed, I think we should fix that before releasing it.

> Let's add autofixing for indentation for the first release also ... we can put indentation at the very end ... We may add a note to *-empty-line-before docs, that we recommend to enable indentation rule if user want to use autofixing.

That's exactly the plan I had in mind 👍 

I'll approve this PR if you want to merge this.

jeddy3:
> we can put indentation at the very end ...

Sounds like a good idea to me. It'll help us mitigate, at least in the short term, having to do waves of fixes.

Great job with this PR btw! 😄

hudochenkov:
Updated changelog to look like:

    -   `at-rule-empty-line-before`
    -   `at-rule-name-case`
    -   `custom-property-empty-line-before`
    -   `declaration-empty-line-before`
    -   `rule-empty-line-before`

PR #2501, opened 2017-04-13
Fix newline for autofixing

hudochenkov:
If source file have no line endings, autofixing will set `newline` depends on current OS. There is no problem for users, I think. But there is problems for tests, because lots of our `code` doesn't have line endings in it. We always assume in tests that it will be `\n`, but on AppVeyor [tests are failing](https://github.com/stylelint/stylelint/pull/2500).

@davidtheclark what do you think?

### DIFF
--- a/lib/lintSource.js
+++ b/lib/lintSource.js
@@ -5,7 +5,6 @@ const _ = require("lodash")
 const assignDisabledRanges = require("./assignDisabledRanges")
 const configurationError = require("./utils/configurationError")
 const path = require("path")
-const os = require("os")
 const ruleDefinitions = require("./rules")
 
 // Run stylelint on a PostCSS Result, either one that is provided
@@ -88,7 +87,7 @@ function lintPostcssResult(
   postcssResult.stylelint.quiet = config.quiet
 
   const newlineMatch = postcssResult.root.toResult().css.match(/\r?\n/)
-  const newline = newlineMatch ? newlineMatch[0] : os.EOL
+  const newline = newlineMatch ? newlineMatch[0] : "\n"
 
   const postcssRoot = postcssResult.root
   assignDisabledRanges(postcssRoot, postcssResult)
--- a/lib/lintSource.js
+++ b/lib/lintSource.js
@@ -4,6 +4,7 @@
 const _ = require("lodash")
 const assignDisabledRanges = require("./assignDisabledRanges")
 const configurationError = require("./utils/configurationError")
+const getOsEol = require("./utils/getOsEol")
 const path = require("path")
 const ruleDefinitions = require("./rules")
 
@@ -87,7 +88,7 @@ function lintPostcssResult(
   postcssResult.stylelint.quiet = config.quiet
 
   const newlineMatch = postcssResult.root.toResult().css.match(/\r?\n/)
-  const newline = newlineMatch ? newlineMatch[0] : "\n"
+  const newline = newlineMatch ? newlineMatch[0] : getOsEol()
 
   const postcssRoot = postcssResult.root
   assignDisabledRanges(postcssRoot, postcssResult)
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -5,6 +5,8 @@ const less = require("postcss-less")
 const basicChecks = require("./lib/testUtils/basicChecks")
 const stylelint = require("./lib/standalone")
 
+jest.mock("./lib/utils/getOsEol", () => () => "\n")
+
 global.testRule = (rule, schema) => {
   expect.extend({
     toHaveMessage(testCase) {
--- a/lib/rules/comment-empty-line-before/__tests__/index.js
+++ b/lib/rules/comment-empty-line-before/__tests__/index.js
@@ -28,17 +28,21 @@ const alwaysTests = {
     description: "Mixed",
   }, {
     code: "a { color: pink;\n\n/** comment */\ntop: 0; }",
+  }, {
+    code: "/** comment */ /** comment */",
+    description: "no newline between comments",
   } ],
 
   reject: [ {
     code: "/** comment */\n/** comment */",
     fixed: "/** comment */\n\n/** comment */",
     message: messages.expected,
+    description: "one newline between comments",
   }, {
     code: "/** comment */\r\n/** comment */",
     fixed: "/** comment */\r\n\r\n/** comment */",
-    description: "CRLF",
     message: messages.expected,
+    description: "CRLF newline between comments",
   }, {
     code: "a { color: pink;\n/** comment */\ntop: 0; }",
     fixed: "a { color: pink;\n\n/** comment */\ntop: 0; }",
--- a/lib/utils/getOsEol.js
+++ b/lib/utils/getOsEol.js
@@ -1,0 +1,12 @@
+/* @flow */
+"use strict"
+
+const os = require("os")
+
+// This function simply provides roundabout way of getting os.EOL
+// so we can mock this for Jest tests
+function getOsEl()/*: string*/ {
+  return os.EOL
+}
+
+module.exports = getOsEl


hudochenkov:
Maybe we should address another “problem” in this PR. If code has mixed line endings, tests are failing, because the first line ending is “correct” line ending for all functions.

[Example](https://github.com/stylelint/stylelint/pull/2500/files#diff-6d86e8fa75d8ffab120b637724892552L150):

```diff
- a {\n /* I am a comment */ \r\n\r\n --custom-prop2: value;}
+ a {\r\n /* I am a comment */ \r\n\r\n --custom-prop2: value;}
```

evilebottnawi:
@hudochenkov i think it is normal, we can add note about this behavior, it is not normal mixed difference line ending, also we can see how it do i eslint

lib/lintSource.js
```diff
@@ -88,7 +87,7 @@ function lintPostcssResult(
   postcssResult.stylelint.quiet = config.quiet
 
   const newlineMatch = postcssResult.root.toResult().css.match(/\r?\n/)
-  const newline = newlineMatch ? newlineMatch[0] : os.EOL
+  const newline = newlineMatch ? newlineMatch[0] : "\n"
```
davidtheclark: I don't think I understand the problem here — I think I need more explanation. `newlineMatch` is supposed to find the first newline that's used in the file, so we can use that newline style again. If no newline is found, we fall back to `os.EOL`. Why would we only want to fall back to `\n`? Can you give an example of the bug that this fixes?

hudochenkov: Failing tests: https://ci.appveyor.com/project/stylelint/stylelint/build/4183. Visually diffs the same, but actually they have different line endings.

In tests we might have no line endings in sources. Rule `at-rule-empty-line-before: "always"`. Test case:

```js
reject: [{
	code: "a {} @media {}",
	fixed: "a {}\n\n @media {}",
	message: messages.expected,
}]
```

On Unix-like system test passes, but on Window it doesn't. Because it expects `a {}\r\n\r\n @media {}`, because there were no line ending in the source, so it uses `os.EOL` as fallback.

We need to find a way to have tests cross-platform.

davidtheclark: Ah, I see. Thanks!

Let's decide first what is the most desirable behavior for users. I assumed that falling back to the OS default was the most typical pattern. Do you agree?

If so, then we should find a way to fix the tests without altering the behavior for actual users. Can we mock `os.EOL` in Jest tests to always be `\n`?

hudochenkov: > I assumed that falling back to the OS default was the best pattern, least likely to surprise users. Do you agree?

Agree.

> we should find a way to fix the tests without altering the behavior for actual users

Also agree. But I don't know how to solve this problem :(

davidtheclark: @hudochenkov: The Jest mocking solution I mentioned should work.

hudochenkov: @davidtheclark I've spent more than hour trying to setup [manual mock](https://facebook.github.io/jest/docs/manual-mocks.html) for `os.EOL` but failed miserably :(

davidtheclark: @hudochenkov: How about adding to `jest-setup.js` something like this?

```js
jest.mock("os", () => ({ EOL: "\n" }))
```

Seems to work for me.

hudochenkov: Holy smokes! You're magician! Thanks! I created `__mocks__/os.js` file, like in documentation, and used `jest.mock("os")` in comment-empty-line-before test. It didn't work.

hudochenkov: @davidtheclark your code is working, but it breaks `system-tests/fix/fix.test.js`, because this test is using other `os` method. I tried to find a way to mock only one method for `os`. But no luck here :(

davidtheclark: Can you extend the original module?

```
jest.mock("os", () => Object.assign({}, os, { EOL: "\n" }))
```

hudochenkov: I wish it works :(

With:

```js
const os = require("os")
jest.mock("os", () => Object.assign({}, os, { EOL: "\n" }))
```

Error:

```
● Test suite failed to run

  /Users/aleks/projects/stylelint/jest-setup.js: babel-plugin-jest-hoist: The module factory of `jest.mock()` is not allowed to reference any out-of-scope variables.
  Invalid variable access: os
  Whitelisted objects: Array, ArrayBuffer, Boolean, DataView, Date, Error, EvalError, Float32Array, Float64Array, Function, Generator, GeneratorFunction, Infinity, Int16Array, Int32Array, Int8Array, InternalError, Intl, JSON, Map, Math, NaN, Number, Object, Promise, Proxy, RangeError, ReferenceError, Reflect, RegExp, Set, String, Symbol, SyntaxError, TypeError, URIError, Uint16Array, Uint32Array, Uint8Array, Uint8ClampedArray, WeakMap, WeakSet, arguments, expect, jest, require, undefined, DTRACE_NET_SERVER_CONNECTION, DTRACE_NET_STREAM_END, DTRACE_HTTP_SERVER_REQUEST, DTRACE_HTTP_SERVER_RESPONSE, DTRACE_HTTP_CLIENT_REQUEST, DTRACE_HTTP_CLIENT_RESPONSE, global, process, Buffer, clearImmediate, clearInterval, clearTimeout, setImmediate, setInterval, setTimeout, console.
  Note: This is a precaution to guard against uninitialized mock variables. If it is ensured that the mock is required lazily, variable names prefixed with `mock` are permitted.

    at invariant (node_modules/babel-plugin-jest-hoist/build/index.js:13:11)
    at newFn (node_modules/babel-traverse/lib/visitors.js:276:21)
    at NodePath._call (node_modules/babel-traverse/lib/path/context.js:76:18)
    at NodePath.call (node_modules/babel-traverse/lib/path/context.js:48:17)
    at NodePath.visit (node_modules/babel-traverse/lib/path/context.js:105:12)
    at TraversalContext.visitQueue (node_modules/babel-traverse/lib/context.js:150:16)
    at TraversalContext.visitMultiple (node_modules/babel-traverse/lib/context.js:103:17)
```

---

With:

```js
const mockos = require("os")
jest.mock("os", () => Object.assign({}, mockos, { EOL: "\n" }))
```

Error:

```
● Test suite failed to run

  ReferenceError: mockos is not defined

    at Object.<anonymous>.jest.mock [as node:os:] (jest-setup.js:141:41)
    at Object.<anonymous> (node_modules/os-homedir/index.js:2:10)
    at Object.<anonymous> (node_modules/cosmiconfig/index.js:4:17)
```

---

With:

```js 
jest.mock("os", () => {
  const nodeos = require("os")

  return Object.assign({}, nodeos, { EOL: "\n" })
})
```

Error:

```
● Test suite failed to run

  RangeError: Maximum call stack size exceeded

    at Resolver.getModulePath (node_modules/jest-resolve/build/index.js:185:16)
    at Object.<anonymous>.jest.mock [as node:os:] (jest-setup.js:141:18)
    at Object.<anonymous>.jest.mock [as node:os:] (jest-setup.js:141:18)
    at Object.<anonymous>.jest.mock [as node:os:] (jest-setup.js:141:18)
```

davidtheclark: @hudochenkov: Sorry ... try putting `jest.mock("os", () => ({ EOL: "\n" }))` in the specific test file running code where that should be mocked, not in `jest-setup`.

hudochenkov: @davidtheclark doesn't work :( I've tried to put `jest.mock("os", () => ({ EOL: "sdfsdfsfd" }))` in:

* `/lib/rules/comment-empty-line-before/__tests__/index.js`
* `jest-setup.js` in `describe` function.
* `jest-setup.js` in `spec` (`it` alias) function.

davidtheclark:
@hudochenkov: in the latest commit I tried a different way of mocking that I think will work.

hudochenkov:
@davidtheclark it works like a charm! Thank you very much! If you don't mind I'm going to merge it and rebase https://github.com/stylelint/stylelint/pull/2500 to rerun tests on that PR.

How did you come up with this solution?

davidtheclark:
> If you don't mind I'm going to merge it and rebase #2500 to rerun tests on that PR.

Yes, go for it!

> How did you come up with this solution?

I have been working with Jest quite a bit lately, so I've tried a lot of different strategies for mocking things. This was another one that I thought I'd try out.

hudochenkov:
> I have been working with Jest quite a bit lately, so I've tried a lot of different strategies for mocking things. This was another one that I thought I'd try out.

I really liked your article about things you've learned working on stylelint. Would you consider writing another one? :)

davidtheclark:
Thanks! It's been a long time since I wrote one. I should try again :)

Issue #2506, opened 2017-04-18
shadow-piercing descendant combinator (/deep/) not supported


binarious:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

The [shadow-piercing descendant combinator](https://drafts.csswg.org/css-scoping/#deep-combinator) (`/deep/` and `>>>`) isn't supported. 

> What CSS is needed to reproduce this issue?

```css
.foo /deep/ .bar {
  color: pink;
}
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "extends": "stylelint-config-standard"
}
```

> Which version of stylelint are you using?

`7.10.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylelint "src/**/*.{scss,css,styl}"`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

It's related to the draft [CSS Scoping Module Level 1](https://drafts.csswg.org/css-scoping/) which is already used in [Angular](https://angular.io/docs/ts/latest/guide/component-styles.html#!#-deep-).

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

The following warnings were flagged using `/deep/`:
```shell
src/styles.scss
 18:7  ✖  Unexpected unknown type selector "/deep/"   selector-type-no-unknown
```

The following warnings were flagged using `>>>`:
```shell
src/styles.scss
 18:7  ✖  Expected single space after ">"    selector-combinator-space-after
 18:7  ✖  Unexpected ">>>"                   selector-descendant-combinator-no-non-space
 18:8  ✖  Expected single space after ">"    selector-combinator-space-after
 18:8  ✖  Expected single space before ">"   selector-combinator-space-before
 18:9  ✖  Expected single space before ">"   selector-combinator-space-before
```

evilebottnawi:
@binarious can you use more bigger css it is allow to understand the whole set of rules that do not handle it :smile:

evilebottnawi:
@binarious also i don't see `/deep/` in spec, can you give me link on spec this feature?

evilebottnawi:
@stylelint/core Someone is well versed in angulare(their css style and other stuff) be good add system tests

evilebottnawi:
Also with regard to me, in our team began to appear projects on Angulyar, now I looked and saw that some guys just turn these rules off, I think we should get support :trophy:

[...]

PR #2508, opened 2017-04-18
Fixed: `selector-type-no-unknown` is now ignore `/deep/` shadow-piercing descendant combinator.

evilebottnawi:

#2506

No, it's self explanatory.

Maybe need update docs :question: or just add example to `valid` :question:

### DIFF
--- a/lib/rules/selector-type-no-unknown/__tests__/index.js
+++ b/lib/rules/selector-type-no-unknown/__tests__/index.js
@@ -77,6 +77,18 @@ testRule(rule, {
   }, {
     code: "abs {}",
     description: "MathML tags",
+  }, {
+    code: ".foo /deep/ .bar {}",
+    description: "shadow-piercing descendant combinator",
+  }, {
+    code: ".foo /dEeP/ .bar {}",
+    description: "shadow-piercing descendant combinator",
+  }, {
+    code: ".foo /DEEP/ .bar {}",
+    description: "shadow-piercing descendant combinator",
+  }, {
+    code: ".foo /for/ .bar {}",
+    description: "reference combinators",
   } ],
 
   reject: [ {
--- a/lib/utils/__tests__/isStandardSyntaxTypeSelector.test.js
+++ b/lib/utils/__tests__/isStandardSyntaxTypeSelector.test.js
@@ -10,8 +10,18 @@ describe("isStandardSyntaxTypeSelector", () => {
       expect(isStandardSyntaxTypeSelector(func)).toBeTruthy()
     })
   })
-  it("nth-child pseudo selector", () => {
+  it("lowercase nth-child pseudo selector", () => {
     return rules(".foo:nth-child(n) {}", func => {
+      expect(isStandardSyntaxTypeSelector(func)).toBeFalsy()
+    })
+  })
+  it("mixedcase nth-child pseudo selector", () => {
+    return rules(".foo:nTh-ChIlD(n) {}", func => {
+      expect(isStandardSyntaxTypeSelector(func)).toBeFalsy()
+    })
+  })
+  it("uppercase nth-child pseudo selector", () => {
+    return rules(".foo:NTH-CHILD(n) {}", func => {
       expect(isStandardSyntaxTypeSelector(func)).toBeFalsy()
     })
   })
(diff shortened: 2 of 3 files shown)


lib/rules/selector-type-no-unknown/index.js
```diff
@@ -89,6 +89,7 @@ const rule = function (actual, options) {
             || svgTags.indexOf(tagNameLowerCase) !== -1
             || keywordSets.nonStandardHtmlTags.has(tagNameLowerCase)
             || mathMLTags.indexOf(tagNameLowerCase) !== -1
+            || tagNameLowerCase === "/deep/"
```
jeddy3: Let's move this to the [`isStandardSyntaxTypeSelector` util](https://github.com/stylelint/stylelint/blob/master/lib/utils/isStandardSyntaxTypeSelector.js). It's where we build on top of `postcss-selector-parser`.

Also, let's return `false` for all [_reference combinators_](https://www.w3.org/TR/2013/WD-selectors4-20130502/#idref-combinators), not just `/deep/` i.e. return `false` for anything that starts and ends with a `/`.

Be sure to comment the line e.g.
```
// Reference combinators like `/deep/`
```

evilebottnawi: @jeddy3 good solution.

evilebottnawi: @jeddy3 one remarks: remove tests on invalid `/deeeps/` (i think it is not related to this rules).

lib/rules/selector-type-no-unknown/__tests__/index.js
```diff
@@ -77,6 +77,18 @@ testRule(rule, {
...
+    description: "shadow-piercing descendant combinator",
+  }, {
+    code: ".foo /dEeP/ .bar {}",
+    description: "shadow-piercing descendant combinator",
+  }, {
+    code: ".foo /DEEP/ .bar {}",
+    description: "shadow-piercing descendant combinator",
+  }, {
+    code: ".foo /for/ .bar {}",
+    description: "reference combinators",
```
davidtheclark: What is a "reference combinator"?

evilebottnawi: @davidtheclark https://www.w3.org/TR/selectors4/#idref-combinators

evilebottnawi: @davidtheclark but seem it is deprecated selector, i don't see infomation about this in other spec, btw some project use `/deep/` and be good just ignore this

davidtheclark: 👍

davidtheclark:
Added to changelog:

> -   Fixed: `selector-type-no-unknown` now ignores the `/deep/` shadow-piercing combinator ([#2508](https://github.com/stylelint/stylelint/pull/2508)).

EmilyRosina:
I still don't understand how to implement this rule in my project :sob: but i need to as we always use `::v-deep` now - HELP :pray:

jeddy3:
@EmilyRosina I don't believe this PR is related to the `::v-deep` pseudo-element.

Please [create a new issue](https://github.com/stylelint/stylelint/issues/new/choose), _following the *bug* template_, if you're getting an unexpected violation.

EmilyRosina:
I just wanted a way to tell the linter that we want to see an error if `/deep/` or `>>>` are used.

jeddy3:
You can use the [`selector-combinator-blacklist` rule](https://github.com/stylelint/stylelint/tree/master/lib/rules/selector-combinator-blacklist) to disallow the `>>>` combinator:

```json
{
  "rules": {
    "selector-combinator-blacklist": [">>>"],
  }
}
```

The `/deep/`[ _reference combinator_](https://www.w3.org/TR/2013/WD-selectors4-20130502/#idref-combinators) is more problematic as the `selector-combinator-blacklist` rule ignores reference combinators. However, you can [write a plugin](https://stylelint.io/developer-guide/plugins) to target and disallow these unusual constructs. I suggest you call the plugin `stylelint-selector-reference-combinator-blacklist`. You can use the code, tests and documentation of the built-in `selector-combinator-blacklist` rule  as a blueprint. Good luck.

EmilyRosina:
Ah ok, so there is no way to do this straight up with stylelint atm :disappointed: 
I guess we'll just rely on Vue/Chrome (unusre who to blame/thank) to tell us for now if someone used `/deep/`. Just a shame we cannot easily do it through the linter.

Thanks anyway :+1:

jeddy3:
> Just a shame we cannot easily do it through the linter.

You can easily do it by writing a plugin. The plugin system exists for exactly these types of scenarios.

EmilyRosina:
It looked a little in-depth on first glance, but maybe it's not as bad as i think :sweat_smile: 
I will have a look once i have more time :smile_cat:

PR #2509, opened 2017-04-18
Fixed: descendant combinator in `selector-descendant-combinator-no-non-space` and `selector-combinator-space-before/after`

evilebottnawi:

#2506

No, it's self explanatory.

### DIFF
--- a/lib/reference/punctuationSets.js
+++ b/lib/reference/punctuationSets.js
@@ -16,6 +16,7 @@ punctuationSets.nonSpaceCombinators = new Set([
   "+",
   "~",
   ">>>",
+  "/deep/",
 ])
 
 module.exports = punctuationSets
--- a/lib/reference/punctuationSets.js
+++ b/lib/reference/punctuationSets.js
@@ -15,6 +15,7 @@ punctuationSets.nonSpaceCombinators = new Set([
   ">",
   "+",
   "~",
+  ">>>",
 ])
 
 module.exports = punctuationSets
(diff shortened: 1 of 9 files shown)


evilebottnawi:
@jeddy3 I'm right now take their implementation (btw i think drop `styleSearch` in this rules and use `postcss-selector-parser`)

jeddy3:
> (btw i think drop styleSearch in this rules and use postcss-selector-parser)

Let's only do that if it adds something e.g. clarity or performance etc.

If the `selector-combinator-space-*` rules work for `>>>` with a simple change the punctuation reference, then let's stick with that for now.

evilebottnawi:
@jeddy3 not working as expected because we have `>` and `>>>` and `styleSearch` found `>` :sob:

evilebottnawi:
Sorry for big commit, but i think it should be done together
One note: `selector-combinator-space-before/after` don't support `/deep/` (and other reference selectors) (because `postcss-selector-parse` don't works with their, theoretically we can write `isReferenceSelector`). I think we can leave this commit as it is and next commit i can see how handle `reference` selectors in this rules (maybe need see other rules).

/cc @jeddy3

jeddy3:
> One note: selector-combinator-space-before/after don't support /deep/ (and other reference selectors) 

Yes, that's fine. Enforcing spacing around `/deep/` will need to be a separate issue as it's a breaking change - feel free to create a new issue for this. The changes in this PR are fixes so that `>>>` isn't mistaken for `>`.

evilebottnawi:
@jeddy3 it think issue already created https://github.com/stylelint/stylelint/issues/2506, even the name it shows :smile:

davidtheclark:
Added to changelog:

> -   Fixed: `selector-descendant-combinator-no-non-space` and `selector-combinator-space-before/after` now understand and check `>>>` shadow-piercing combinator ([#2509](https://github.com/stylelint/stylelint/pull/2509)).

PR #2510, opened 2017-04-18
Use unambiguous terminology for what a linter reports on

ZacharyRSmith:
The current text "The following patterns are considered warnings" is misleading because rules default to throwing errors instead of warnings.  So, this update shows how to set severity to warning.

If this is considered a valuable change, then we should change similar READMEs.


jeddy3:
> is misleading because rules default to throwing errors instead of warnings. 

"The following patterns are (not) considered warnings:" is a standardised string that appears before all examples.

We use "warning(s)" throughout the documentation as a generic word to describe when the linter finds something it doesn't like.

The choice of terminology predates the introduction of severities into stylelint. I can see how it can be confusing now that we have severities.

> If this is considered a valuable change, then we should change similar READMEs.

Thanks for the offer :) Perhaps we can replace the word "warning" with something generic to "errors" and "warnings", e.g. "infraction" or "violation".  Do you think this would be clearer? This approach also takes into account the [`defaultSeverity` configuration option](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#defaultseverity) i.e. it's not just a rules' `"severity"` secondary option that determines if a rule will warn or error.

ZacharyRSmith:
Ah, I did not know about the defaultSeverity opt.  Thanks for pointing it out!

Yes, I think "violation" would be clearer.

jeddy3:
> Yes, I think "violation" would be clearer.

Cool.

@stylelint/core, any objections to changing "warning(s)" to "violation(s)" and "warn" to "complain" (which we [already use in a few places](https://github.com/stylelint/stylelint/search?utf8=%E2%9C%93&q=complains&type=))?

evilebottnawi:
@jeddy3 sounds good

davidtheclark:
@jeddy3: I like those word-replacements.

jeddy3:
> "warning(s)" to "violation(s)" and "warn" to "complain" 

@ZacharyRSmith Would you like to make those changes, across the board, in this PR?

ZacharyRSmith:
@jeddy3 I made those changes for lib/rules/*.md

ZacharyRSmith:
eh i'll check on the failing build

ZacharyRSmith:
i just double-checked, and my changes are not related to the fail.  i need to bounce, sorry i can't fix it

hudochenkov:
@ZacharyRSmith thank you for your work! Indeed build fails not because of your changes. Don't worry.

jeddy3:
FYI, PR to update stylelint.io in prep for this change is at https://github.com/stylelint/stylelint.io/pull/36

PR #1247, opened 2016-05-11
Use parseSelector instead of postcss-selector-parser

davidtheclark:
Addresses #1153.

This illustrates that with a new util function, `parseSelector`, wrapping `postcss-selector-parser`, we can catch that module's errors and deliver a PostCSS warning that should be better handled by various stylelint clients.

@Arcanemagus does this look like it will solve the Atom plugin's problems?

If everyone agrees that this will work, we just need to refactor all the rules using `postcss-selector-parser`.

PR #2512, opened 2017-04-19
Chore: use `parseSelector` instead `postcss-selector-parser` directly.

evilebottnawi:

Just code improved. See https://github.com/stylelint/stylelint/issues/1247

No, it's self explanatory.


davidtheclark:
Thanks @evilebottnawi!

PR #2516, opened 2017-04-22
Allow specifying `codeFilename` to `createStylelintResult` for raw code linting

### LINKED ISSUE #2450: Allow specifying `codeFilename` to `createStylelintResult` for raw code linting

Apologies if this is a dupe, but this is pretty much an issue to relay @shinnn and follow up on shinnn/vscode-stylelint#37 where he was talking about [this line](https://github.com/stylelint/stylelint/blob/master/lib/standalone.js#L88) and the fact that `absoluteCodeFilename` is not being passed as the second argument. This makes the extension default to `getConfigForFile`'s `process.cwd()` call and as a result, tries to get the config from the root of the project directory. If you put the config file in a subdirectory of the project root then Stylelint wouldn't be able to automatically pick it up. Having the second parameter exposed and filled with `absoluteCodeFilename` will allow linting on raw code be a little bit more flexible.

I was taking a quick look at the code and unless if I missed anything I can't find any issues with it. I can create a quick PR if that helps.

None.

Running `vscode` on a project with a config file nested in a directory and not in the root directory should do it.

Anything.

According to shinnn/stylelint-vscode, the language server VSCode is running Stylelint from, it uses `^7.6.0`. 

VSCode with `stylelint` extension.

Shouldn't be.

`createStylelintResult` should return a proper result. 

Nothing happens in VSCode, as if Stylelint hasn't been configured.

diagramatics:
Resolves #2450. Testing with `fs.stat()` for existence before passing it in seems to work fine, and the tests all still work.


evilebottnawi:
@diagramatics thank for pr, be good add test for this

lib/standalone.js
```diff
@@ -81,7 +81,25 @@ module.exports = function (options/*: stylelint$standaloneOptions */)/*: Promise
       code,
       codeFilename: absoluteCodeFilename,
     }).then(postcssResult => {
-      return stylelint._createStylelintResult(postcssResult)
+      // Check for file existence
```
davidtheclark: Why do we need to check for the file's existence?

diagramatics: It makes a test in https://github.com/stylelint/stylelint/blob/master/lib/__tests__/standalone.test.js#L399-L426 fail otherwise.

diagramatics:
Sorry for the wait. I just added a test for nested config checks.

ntwb:
I've just fixed the _prettier_ merge conflicts, everything looks good now I think?

ntwb:
Thanks again @diagramatics, sorry it took a while to get merged 😄 

Changelog updated in https://github.com/stylelint/stylelint/commit/73f99bbe3f5c81b1435b141a8ca98867813f64fe
> -   Added: Allow specifying `codeFilename` to `createStylelintResult` for raw code linting standalone API([#2450](https://github.com/stylelint/stylelint/issues/2450))

diagramatics:
Yay! Thanks for the help everyone!

Issue #2263, opened 2017-01-15
Document use of escaped characters in selector-class-pattern


demrks:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

There seems to be a bug with the `selector-class-pattern` rule that produces an error although the regex works properly [in tests](http://regexr.com/3f2me)

> Which rule, if any, is this issue related to?

`selector-class-pattern`

> What CSS is needed to reproduce this issue?

e.g.

```css
.u-text-left { }
.u-size-11\/12\@sm { }
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "selector-class-pattern": "^.(o-|c-|u-|s-|is-|has-){1}([-]?[\\]?[/]?[a-z0-9]+){2,}(__[a-z0-9]([-]?[a-z0-9]+)*)?(--[a-z0-9-]+)?(\\@[a-z0-9-]+)?$"
  }
}
```

> Which version of stylelint are you using?

`7.7.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

Webpack plugin

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No, vanilla CSS

> What did you expect to happen?

No warnings to be flagged

> What actually happened (e.g. what warnings or errors you are getting)?

```shell
✖  Expected class selector ".u-text-left" to match specified pattern        selector-class-pattern
✖  Expected class selector ".u-size-11\/12\@sm" to match specified pattern   selector-class-pattern
```

jeddy3:
@demrks Thanks for the report and for using the template.

You [shouldn't include](http://stylelint.io/user-guide/rules/selector-class-pattern/#options) the `.` in your regex. This change, however, only catches the `.u-text-left` selector. The `.u-size-11\/12\@sm` is still missed. I think this is a separate issue that needs further investigation.

The rule itself simply [uses `new RegExp(pattern)`], so it seems unlikely to be an issue here. I did notice that PostCSS [parses the selector](https://astexplorer.net/#/mJWRsDuLSx) to _include escaping_. Perhaps this has something to do with it?

@demrks Are you able to dig a little deeper and help us investigate this?

demrks:
@jeddy3 Thanks so much, it works just fine if I change `"\\"` to `"\\\\"`:

```
[...]

PR #2520, opened 2017-04-22
Fix inconsistent behaviour for escaped characters in selector-class-pattern

davidtheclark:
Ref #2263.

cc @hudochenkov

### DIFF
--- a/lib/rules/selector-class-pattern/README.md
+++ b/lib/rules/selector-class-pattern/README.md
@@ -10,6 +10,8 @@ Specify a pattern for class selectors.
 
 This rule ignores non-outputting Less mixin definitions and called Less mixins.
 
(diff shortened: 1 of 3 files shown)


hudochenkov:
@evilebottnawi feel free to jump on it :)

hudochenkov:
I have a feeling that there is nothing we can do code-wise. I think amount of developers who uses this rule and writes escaped selectors is pretty small. We can add a note in docs that escaped selectors (`.u-size-11\/12\@sm`) parsed as escaped one more time (`.u-size-11\\/12\\@sm`), so RegExp should account for that.

davidtheclark:
@hudochenkov: Your plan sounds good to me!

jeddy3:
Let's update the docs in this PR and close the issue.

jeddy3:
@hudochenkov Updated docs.

PR #2527, opened 2017-04-25
Use violation wording in User guides

jeddy3:
Ref: https://github.com/stylelint/stylelint/pull/2510


jeddy3:
FYI, just rebased off `master` for doiuse fix.

Issue #2259, opened 2017-01-14
ROADMAP


davidtheclark:
I'd like to chat about the next steps for stylelint. Our 8.0 release will be mostly focused on standardizing option names; and we already had a major release that was mostly about standardizing rule names; so I expect that more standardization like this will not be necessary for quite a while, if ever. We've create a very fine configuration system. So let's think about what might be our next major goals, if any.

First off, I'll say that I personally have lost all sympathy for and interest in non-standard syntax. I am now a hard-line anti-Sass militant: as far as I'm concerned it should be banished from the realm. I am not planning to spend any of my time fixing issues related to non-standard syntax. I've been toying with the idea of forking stylelint to create a much simplified CSS-only version — that's extreme, but the next paragraph shows why.

There are two directions that interest me the most right now: 

- Automatic fixes
- Spec validation

I agree with the growing opinion in the frontend world (evidence by the immediate popularity of `prettier` for JS) that style format rules are best enforced by an automatic fixer rather than a linter. I *might* say that I'm interested, then, in contributing to stylefmt — *but* I'm discouraged by the idea of getting involved in another project that is going to be poisoned by non-standard syntaxes. Just as I have an inclination to fork stylelint to create a CSS-only linter, I have an inclination to fork stylefmt to create a CSS-only formatter.

Spec validation would involve contributing to [css-values](https://github.com/ben-eb/css-values) or [csstree](https://github.com/csstree/csstree) and their stylelint plugins. These important tools are another reason to hate Sass its kin: these projects will be hindered, stalled, even derailed if they try to accommodate non-standard syntax completely.

My interests in the above come from belief that there are 3 things people typically want from a code quality tool:

1. Clean-up & standardize formatting: stylelint does this, but this is *better* done with an auto-formatter (using stylelint's config).
2. Enforce best practices: stylelint does this as well as anything.
3. Validate that the code is not broken: stylelint does *a little* of this, but this is *better* done via major projects like css-values and csstree.

These priorities make me very, very inclined to steer away from non-standard syntaxes. I could just step back, as mentioned above, and hope that others who are not so hostile will take charge of handling anything non-standard-syntax-related issues. But actually **I'm revisiting @jeddy3's idea in #1749 , that we offload all handling of non-standard syntax, push it out of stylelint. This could be v9.**

[...]

PR #2529, opened 2017-04-25
Indentation autofixing

hudochenkov:
I've made autofixing for most use cases. These use cases was analyzed with PostCSS AST, so I could easily apply fixes. But for big number of other use cases this rule uses `style-search`. It's used for indentation inside declaration values, selectors, and at-rule parameters.

I haven't figured out how to deal with this yet. I've commented all `reject` cases which isn't covered by current autofixing.

Any suggestions how to deal with this problem?

Also there is a problem with Less syntax. I haven't investigate what causes the problem yet.

In order to finish this PR we need:

- [x] Finish autofixing for all other use cases
- [x] Add a note to rule's readme and rules list about autofixing
- [x] Add a note to `*-empty-line-before` rules with recommendation to enable `indentation` rule

P. S. It was total surprise to me, that this rule has secondary options 😯

<details>
<summary>Error log with Less mixins and autofixing</summary>

Semicolon gets removed.

```
 FAIL  lib/rules/indentation/__tests__/functions.js
  ● indentation › accept › Less mixin with multi-line arguments

    expect(received).toBe(expected)

    Expected value to be (using ===):
      ".foo {
      .mixin(
        @foo,
        @bar,
        @baz
      );
    }"
    Received:
      ".foo {
      .mixin(
        @foo,
        @bar,
        @baz
      )
    }"

    Difference:

    - Expected
    + Received

     .foo {
       .mixin(
         @foo,
         @bar,
         @baz
    -  );
    +  )
     }

      at stylelint.then.output (jest-setup.js:59:35)

  ● indentation › accept › Less mixin with multi-line arguments

    Error
      Error: expect(received).toBe(expected)

      Expected value to be (using ===):
        ".foo {
        .mixin(
          @foo,
          @bar,
          @baz
        );
      }"
      Received:
        ".foo {
        .mixin(
          @foo,
          @bar,
          @baz
        )
      }"

      Difference:

      - Expected
      + Received

       .foo {
         .mixin(
           @foo,
           @bar,
           @baz
      -  );
      +  )
       }
      at stylelint.then.output (jest-setup.js:59:35)
```

</details>

### DIFF
--- a/docs/user-guide/rules.md
+++ b/docs/user-guide/rules.md
@@ -255,7 +255,7 @@ Here are all the rules within stylelint, grouped by the [*thing*](http://apps.wo
 
 ### General / Sheet
 
--   [`indentation`](../../lib/rules/indentation/README.md): Specify indentation.
+-   [`indentation`](../../lib/rules/indentation/README.md): Specify indentation (Autofixable).
 -   [`max-empty-lines`](../../lib/rules/max-empty-lines/README.md): Limit the number of adjacent empty lines.
 -   [`max-line-length`](../../lib/rules/max-line-length/README.md): Limit the length of a line.
 -   [`max-nesting-depth`](../../lib/rules/max-nesting-depth/README.md): Limit the depth of nesting.
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -99,6 +99,7 @@ global.testRule = (rule, schema) => {
               return stylelint(Object.assign({ fix: true }, options)).then((output) => {
                 const fixedCode = getOutputCss(output)
                 expect(fixedCode).toBe(testCase.fixed)
+                expect(fixedCode).not.toBe(testCase.code)
               })
             })
           })
--- a/lib/rules/at-rule-empty-line-before/README.md
+++ b/lib/rules/at-rule-empty-line-before/README.md
@@ -12,7 +12,7 @@ a {}
 
 If the at-rule is the very first node in a stylesheet then it is ignored.
 
-The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofixing-errors) can automatically fix all of the problems reported by this rule.
+The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofixing-errors) can automatically fix all of the problems reported by this rule. We recommend to enable [`indentation`](../indentation/README.md) rule for better autofixing results with this rule.
 
 ## Options
 
--- a/lib/rules/comment-empty-line-before/README.md
+++ b/lib/rules/comment-empty-line-before/README.md
@@ -16,7 +16,7 @@ If you're using a custom syntax which support single-line comments with `//`, th
 
 **Caveat:** Comments within *selector and value lists* are currently ignored.
 
-The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofixing-errors) can automatically fix all of the problems reported by this rule.
+The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofixing-errors) can automatically fix all of the problems reported by this rule. We recommend to enable [`indentation`](../indentation/README.md) rule for better autofixing results with this rule.
 
 ## Options
 
--- a/lib/rules/custom-property-empty-line-before/README.md
+++ b/lib/rules/custom-property-empty-line-before/README.md
@@ -12,7 +12,7 @@ a {
  *                   This line */
 ```
 
-The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofixing-errors) can automatically fix all of the problems reported by this rule.
+The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofixing-errors) can automatically fix all of the problems reported by this rule. We recommend to enable [`indentation`](../indentation/README.md) rule for better autofixing results with this rule.
 
 ## Options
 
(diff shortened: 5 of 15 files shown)


jeddy3:
Looking good!

> Also there is a problem with Less syntax. I haven't investigate what causes the problem yet.

I think in https://github.com/stylelint/stylelint/issues/2259 we decided that advanced features like autofixing would only be for standard-syntax. We don't want to repeat the path of stylefmt where the project stagnated due to the quantity of non-standard bugs for this advanced feature. I think we can remove any non-standard tests and make it clear within the documentation that autofix is a standard-syntax only feature. I'm still plugging away at a `VISION.md` doc, that we'll be able to point users towards if they want to know more background reasoning to this decision.

evilebottnawi:
@jeddy3 sounds reasonably, but most of our users use `scss/sass`, And it will be for them not too joyful news, I do not think that we should at the moment not support non-standard syntaxs

hudochenkov:
@evilebottnawi I believe, SCSS will work just fine, because parser is good. Less parser is a mess, and I have never used Less. So I'm not going to do anything for Less, but for SCSS I can do necessary changes, because I'm using it on my new job.

evilebottnawi:
@jeddy3 @hudochenkov i think we should fix only standard construction (`isStandard*`) and etc, rules not related `stylelint` can fix non standard constructions related to own syntax

jeddy3:
> i think we should fix only standard construction (isStandard*) 

This feels like a sensible approach.

evilebottnawi:
@jeddy3 I believe that we should not touch anything what don't pass `isStandard*`, even if we can do it, otherwise it will be just a terrible collapse of errors and logic. But we can think about joining your autofixers for the rules, example:
```
"comment-empty-line-before": [ "always", {}, {fix: "path/fix.js"}],
```
if someone will want to use their fixes, I do not think that it will be difficult to implement

jeddy3:
> "comment-empty-line-before": [ "always", {}, {fix: "path/fix.js"}],

That's very interesting, and very much aligned with our philosophy of extensibility (plugins, processors, custom syntaxes, custom formatters and shareable configs).

evilebottnawi:
@hudochenkov @jeddy3 maybe before `v8` release we should remove unnecessary fixes don't related to standard syntax?

hudochenkov:
@evilebottnawi there is no unnecessary fixes, that I'm aware of.

evilebottnawi:
@hudochenkov it is good :+1:

davidtheclark:
> "comment-empty-line-before": [ "always", {}, {fix: "path/fix.js"}]

@evilebottnawi: This sounds like a pain in the neck, and I doubt it would be used very often if you need different modules for each rule.

evilebottnawi:
@davidtheclark it is just idea :smile: , main thing here https://github.com/stylelint/stylelint/pull/2529#issuecomment-297348450

hudochenkov:
I think if we can add fixing to a rule, we should fix as much as we can. Maybe we can't do fixing for some options of a rule. It's ok, we'll show linting error as usual. Or if we can't be sure about consistency of fixing for particular options or rule, we won't add fixing, because it could be auto-breaking instead of auto-fixing.

Let's return to this PR :)

What should I do with [part](https://github.com/stylelint/stylelint/blob/3d8290da357dc757c7b0116d6852d049e8f20b9d/lib/rules/indentation/index.js#L183-L264) which checks code using `style-search`? Shall I rewrite it without `style-search`?

evilebottnawi:
@hudochenkov Run away from it :running_man: :stuck_out_tongue: If seriously we should get rid of it as much as possible where possible.

hudochenkov:
I haven't found a way how to rewrite this rule away from style-search, because it's very complex. But I finally figured out how to add autofixing for the code that uses style-search!

Less is still broken, because of [semicolon bug](https://github.com/stylelint/stylelint/pull/2547) in postcss-less. That's why tests without `fix: true` for it.

@stylelint/core PR is ready for a review.

evilebottnawi:
@hudochenkov i think before we should fix `semicolon bug` in `postcss-less` :smile: btw seems goods

jeddy3:
Shall we merge this and https://github.com/stylelint/stylelint/pull/2577, and then get out experimental support for autofixing (for standard syntax)? 

We're only a couple of issues away from getting `8.0.0` out too. Although adding autofixing still has some question marks around it, it feels like setting off down this path is a good idea as the end goal of allowing plugins to autofix (e.g. stylelint-scss and stylelint-order) seems solid.

hudochenkov:
@evilebottnawi could you take a look at this PR in the next few day, please? :)

jeddy3:
@hudochenkov Shall we merge this? Tests look good and I think it's better we get the ball rolling on this fix stuff than wait. We're also introducing it an experimental feature.

hudochenkov:
@jeddy3 Yeah, let's do it.

davidtheclark:
@jeddy3 @hudochenkov: Sorry for my absence. I'm afraid my time is being consumed these days and I don't have much left to dedicate to stylelint — so please don't wait on me to make any progress!

hudochenkov:
Thank you for a review!

Updated changelog:

    -   `at-rule-empty-line-before`
    -   `at-rule-name-case`
    -   `custom-property-empty-line-before`
    -   `declaration-empty-line-before`
    -   `indentation`
    -   `rule-empty-line-before`

For some reason Github named [commit](https://github.com/stylelint/stylelint/commit/f0e0780f87b43973f3bf92e405c7c047d0769d70) as “Create CHANGELOG.md”.

jeddy3:
I'm in a similar situation. When we started working on stylelint I was on a hiatus from agency work, but now my time is being consumed by non-open source work. I'm going to try to make some time to help get `8.0.0` out of the door, though. @hudochenkov Thanks for driving these PRs through!

Issue #2110, opened 2016-11-23
False positive for pseudos and Less guard in selector-descendant-combinator-no-non-space


thybzi:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Using CSS pseudo selector (`:hover`, `:not()`, `::after` etc.) and LESS guard (`when`) in the same line leads to `selector-descendant-combinator-no-non-space` false positive.

> Which rule, if any, is this issue related to?

`selector-descendant-combinator-no-non-space`

> What CSS is needed to reproduce this issue?

```less
:hover when (@a = 42) {
    color: red;
}
```

> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
    "rules": {
        "selector-descendant-combinator-no-non-space": true
    }
}
```

> Which version of stylelint are you using?

e.g. `7.6.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylelint test.less`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes, it's related to LESS `when` syntax: http://lesscss.org/features/#css-guards-feature

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

The following warnings were flagged:

```shell
test.less
 1:12  ×  Unexpected " (@a = 42)"   selector-descendant-combinator-no-non-space
```

thybzi:
Concerns not only `:not()`, but all pseudo selectors. Updated the issue description.

evilebottnawi:
@thybzi Thanks for feedback, today i can see this.

evilebottnawi:
@thybzi can you provide full example, i can't find using `pseudo` with `when` in docs?

thybzi:
Hi! I have provided the example in the head of this issue:
```less
:hover when (@a = 42) {
    color: red;
}
```

evilebottnawi:
@thybzi I am on the weekends :smile: Tomorrow it'll :+1:

thybzi:
Okay, waiting for this. Thank you!

evilebottnawi:
@thybzi sorry, very busy last month :sob: i can try resolve himself if you need this immediately :smile:

evilebottnawi:
Done in https://github.com/stylelint/stylelint/pull/2557

PR #2557, opened 2017-05-07
Fixed: ignore less guards in `selector-descendant-combinator-no-non-space`.

evilebottnawi:

#2110

No, it's self explanatory.

### DIFF
--- a/lib/utils/isStandardSyntaxRule.js
+++ b/lib/utils/isStandardSyntaxRule.js
@@ -37,6 +37,11 @@ module.exports = function (rule/*: Object*/)/*: boolean*/ {
     return false
   }
 
+  // Less guards
+  if ((/when\s+(not\s+)*\(/).test(selector)) {
+    return false
+  }
+
   // Ignore Scss nested properties
   if (selector.slice(-1) === ":") {
     return false
--- a/lib/rules/selector-descendant-combinator-no-non-space/__tests__/index.js
+++ b/lib/rules/selector-descendant-combinator-no-non-space/__tests__/index.js
@@ -54,3 +54,15 @@ testRule(rule, {
     column: 5,
   } ],
 })
+
+testRule(rule, {
+  ruleName,
+  config: [true],
+  syntax: "less",
+
+  accept: [ {
+    code: ":hover when (@variable = true) { color: red; }",
+  }, {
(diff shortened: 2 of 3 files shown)


jeddy3:
Does this heuristic introduce a false negative for this valid CSS?

```css
when     a { }
```

Where `when` is an unknown, yet valid, type selector.

evilebottnawi:
@jeddy3 hm, right, let's add more complexly check

jeddy3:
Thanks. Can we add get `truthy` tests in the util for `a  when {}` and `when a {}` please?

lib/utils/isStandardSyntaxRule.js
```diff
@@ -37,6 +37,11 @@ module.exports = function (rule/*: Object*/)/*: boolean*/ {
     return false
   }
 
+  // Less guards
+  if ((/when.*?\(.*?\)/).test(selector)) {
```
davidtheclark: only a whitespace can separate `when` and `(`, right? `when\s+\(.*?\)`

evilebottnawi: @davidtheclark no, allow use `when not ()`

davidtheclark: I think the regexp should account for this limited scenario, instead of allowing for any arbitrary characters between `when` and `(`.

evilebottnawi: @davidtheclark in theory allow `when not not (@variable = true)` (does not make sense)

davidtheclark: @evilebottnawi: You can make the regexp as complicated as you think necessary — but I think it's not great to make it so simple that it might raise false positives.

evilebottnawi: @davidtheclark What solution is better to use here?

davidtheclark: How about `/when\s+(not\s+)*\(/`? I think it will just take writing tests for what should and should not work and fiddling with the regexp until that's what happens.

jeddy3:
Changelog:

> -   Fixed: `selector-descendant-combinator-no-non-space` now ignores Less guards ([#2557](https://github.com/stylelint/stylelint/pull/2557)).

Issue #2519, opened 2017-04-22
Remove Sinon stubs in favor of Jest mocks


davidtheclark:
Since we are using Jest now, we should be able to remove our Sinon dependency.

This will also eliminate the warning we're getting: `sinon.stub(obj, 'meth', fn) is deprecated and will be removed from the public API in a future version of sinon.`

jeddy3:
SGTM.

jeddy3:
Done in https://github.com/stylelint/stylelint/issues/2559

PR #2559, opened 2017-05-07
Replace sinon with jest mocks

CAYdenberg:

Issue #2519 

WIP as I'm doing pieces at a time. Will update as I go.

### DIFF
--- a/lib/__tests__/standalone-deprecations.test.js
+++ b/lib/__tests__/standalone-deprecations.test.js
@@ -5,20 +5,14 @@ const ruleDefinitions = require("../rules")
 const standalone = require("../standalone")
 
 describe("standalone with deprecations", () => {
-  let ruleDefinitionsStub
-
   beforeAll(() => {
-    ruleDefinitionsStub = Object.assign(ruleDefinitions, { "block-no-empty": jest.fn(() => {
+    ruleDefinitions["block-no-empty"] = jest.fn(() => {
       return (root, result) => {
         result.warn("Some deprecation", {
           stylelintType: "deprecation",
         })
       }
-    }) })
-  })
-
-  afterEach(() => {
-    ruleDefinitionsStub["block-no-empty"].mockClear()
+    })
   })
 
   it("works", () => {
--- a/lib/__tests__/standalone-deprecations.test.js
+++ b/lib/__tests__/standalone-deprecations.test.js
@@ -5,9 +5,9 @@ const ruleDefinitions = require("../rules")
 const standalone = require("../standalone")
 
 describe("standalone with deprecations", () => {
-  let ruleDefinitionsStub, results, output
+  let ruleDefinitionsStub
 
-  beforeEach(() => {
+  beforeAll(() => {
     ruleDefinitionsStub = Object.assign(ruleDefinitions, { "block-no-empty": jest.fn(() => {
       return (root, result) => {
         result.warn("Some deprecation", {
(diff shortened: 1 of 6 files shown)


evilebottnawi:
@CAYdenberg thanks for helping to us

lib/__tests__/standalone.test.js
```diff
@@ -298,13 +297,13 @@ describe("standalone with deprecations", () => {
   let ruleDefinitionsStub, results, output
 
   beforeEach(() => {
-    ruleDefinitionsStub = sinon.stub(ruleDefinitions, "block-no-empty", () => {
+    ruleDefinitionsStub = Object.assign({}, ruleDefinitions, { "block-no-empty": jest.fn(() => {
```
CAYdenberg: This fails. I'm uncertain how to accomplish this with Jest :(

CAYdenberg:
OK, I got through everything but I got stuck on how to stub `ruleDefinitions` in `lib/tests/standalone.tests.js`.
Help would be appreciated.

lib/__tests__/standalone.test.js
m-allanson: I think the issue here is that the mocked function isn't being applied to the required `ruleDefinitions` variable. You could try creating a [manual mock](https://facebook.github.io/jest/docs/manual-mocks.html) in `lib/__mocks__/rules.js`, then use `jest.mock(ruleDefinitions)`.

You might also need to move this "standalone with deprecations" test out to a new `standalone-deprecations.test.js` file so that the mocked `ruleDefinitions` doesn't interfere with the other tests in this file.

Let me know if you've any questions.

m-allanson:
@CAYdenberg Thanks! I've added some thoughts on how that could be fixed.

CAYdenberg:
OK, I figured it out. I also took the liberty of rearranging the deprecations test slightly.

I did have to move it to a new file ... it should be possible to save the original method in a temporary variable and recreate the original object, but this feels simpler (to me, anyway).

lib/utils/__tests__/validateOptions.test.js
```diff
@@ -15,66 +14,66 @@ describe("validateOptions for primary options", () => {
       possible: [ "a", "b", "c" ],
       actual: "a",
     })
-    expect(result.warn.calledOnce).toBeFalsy()
-    result.warn.reset()
+    expect(result.warn.mock.calls).toHaveLength(0)
+    result.warn.mockClear()
```
davidtheclark: I think that instead of using `mockClear` invocations maybe we should use [the `clearMocks` configuration option](https://facebook.github.io/jest/docs/en/configuration.html#clearmocks-boolean). Have you looked into that?

lib/utils/__tests__/validateOptions.test.js
```diff
@@ -294,8 +281,8 @@ describe("validateOptions with a function for 'possible'", () => {
       ],
     })
     expect(validArrayOfObjectsAndStrings).toBe(true)
-    expect(result.warn.called).toBeFalsy()
-    result.warn.reset()
+    expect(result.warn.mock.calls).toHaveLength(0)
```
davidtheclark: This is fine; but in all these places, you could have also used [`toHaveBeenCalledTimes`](https://facebook.github.io/jest/docs/expect.html#tohavebeencalledtimesnumber).

lib/__tests__/standalone-deprecations.test.js
```diff
@@ -0,0 +1,35 @@
...
+  let ruleDefinitionsStub
+
+  beforeAll(() => {
+    ruleDefinitionsStub = Object.assign(ruleDefinitions, { "block-no-empty": jest.fn(() => {
+      return (root, result) => {
+        result.warn("Some deprecation", {
+          stylelintType: "deprecation",
+        })
+      }
+    }) })
```
davidtheclark: Oof ... please reformat this block.

davidtheclark:
Looks great! @ntwb do you think this is ready to merge?

CAYdenberg:
How about it @ntwb ?

PR #2570, opened 2017-05-13
Fixed: ignore mergeable less property.

evilebottnawi:

No

No, it's self explanatory.

I don't known why we check non standard less properties, i think we should ignore this.


jeddy3:
Here's [merge](http://lesscss.org/features/#merge-feature) in the Less docs for anyone else, like me, who didn't know there was such thing :)

davidtheclark:
Added to changelog:

> -   Fixed: Less mergeable properties are now ignored ([#2570](https://github.com/stylelint/stylelint/pull/2570)).

PR #2571, opened 2017-05-13
Fixed: ignore `extend` in less.

evilebottnawi:

No

No, it's self explanatory.

Ref: http://lesscss.org/features/#import-options-reference-example


jeddy3:
Added to changelog:

> -   Fixed: Less `:extend` is now ignored ([#2571](https://github.com/stylelint/stylelint/pull/2571)).

Issue #2581, opened 2017-05-17
False positives for parameterless functions in function-parentheses-space-inside


PhilippSoehnlein:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug where the `function-parentheses-space-inside` rule set to `always-single-line` reports a warning for functions with no params.
IMO this shouldn't be flagged or even if, an `except`-option for functions with no params should be possible.

> Which rule, if any, is this issue related to?

`function-parentheses-space-inside`

> What CSS is needed to reproduce this issue?

```scss
@function someFunction() {
    @return 'someUrl';
}
.foo {
    background-image: someFunction();
}
```

> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
  "rules": {
    "function-parentheses-space-inside": "always-single-line"
  }
}
```

> Which version of stylelint are you using?

`7.10.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes, it's related to SCSS.

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

Expected single space before ")" in a single-line function

jeddy3:
@PhilippSoehnlein Thank you for the report and for using the issue template.

> IMO this shouldn't be flagged

Yes, you're right. It's should _not_ flag a violation.

I believe there are two possible fixes, depending on whether parameterless functions are valid construct in CSS. Do you know if they are?

If not, we should ignore them in [`isStandardSyntaxFunction`](https://github.com/stylelint/stylelint/blob/master/lib/utils/isStandardSyntaxFunction.js).

Otherwise, we should bake an early return into [`function-parentheses-space-inside`](https://github.com/stylelint/stylelint/blob/master/lib/rules/function-parentheses-space-inside/index.js#L52) directly.

hudochenkov:
@jeddy3 I did a research. `url()` is valid. Also some [`filter` functions](https://www.w3.org/TR/filter-effects-1/#FilterProperty) could have no parameters.

jeddy3:
Thanks for doing the research!

hudochenkov:
@PhilippSoehnlein thank you for reporting!

Fixed by https://github.com/stylelint/stylelint/pull/2587.

PR #2587, opened 2017-05-20
Ignore functions without parameters in function-parentheses-space-inside

hudochenkov:

#2581


hudochenkov:
Added to changelog:

-   Fixed: `function-parentheses-space-inside` now ignores functions without parameters ([#2587](https://github.com/stylelint/stylelint/pull/2587)).

PR #2596, opened 2017-05-27
Support for HTML, Markdown, Vue component

gucong3000:
- Support for HTML, [Markdown](https://daringfireball.net/projects/markdown/syntax), [Vue component](https://vue-loader.vuejs.org/)
- Fix syntax error when use `--fix`


docs/user-guide/css-processors.md
```diff
@@ -32,7 +33,7 @@ If you're using the linter as a [PostCSS Plugin](postcss-plugin.md), you'll need
 ```js
 var postcss = require("postcss")
 var scss = require("postcss-scss")
-// or use "postcss-less" or "sugarss"
+// or use "postcss-less" or "sugarss", "postcss-sass", "postcss-html"
```
Arcanemagus: This should be:
```js
// or use "postcss-less", "sugarss", "postcss-sass", or "postcss-html"
```

gucong3000: Fixed. 3Q

sergesemashko:
@gucong3000, I was playing around with this feature and found an edge-case with autofix enabled (I just added `fix: true` after [line 161](https://github.com/stylelint/stylelint/pull/2596/files#diff-7d31f5d52b54f165816fa0ff4e0de2dfR161)) - autofix adds extra intends before the `a` selector. So, it invalidates sugarSS syntax. If you remove `fix: true` back to initial state - the tests will fail with updated fixtuers. 

It might be not related to this particular PR, but I think it makes sense to cover autofixing in html by tests as well.

Besides that, looks good to me. 

@stylelint/core, I'd suggest another pair of eyes to take a look at this PR.

gucong3000:
@sergesemashko I found the code for this problem
[sugarss/stringifier.es6#L93](https://github.com/postcss/sugarss/blob/master/stringifier.es6#L93)

Change here to:
```js
while ( node.parent && node.type !== 'root' ) {
```

gucong3000:
@sergesemashko It was fixed.

sergesemashko:
@gucong3000, the autofix issue is gone now, thanks. However, I discovered another tricky case: 
1. Add valid rule to any existing mixture, e.g. updated `style-tag.vue`: 
  
1. It's still expected to receive only one warning for that file, but 0 was returned and the test fails 
  

It looks like the parser doesn't recognize a mix of broken and correct rules and violations are not discovered.

gucong3000:
@sergesemashko in this case, `sugarss` will parse source to:
```css
a div {color: white}
```
So, It is not a mix of broken and correct rules, It is a valid rule.

gucong3000:
> Overall, if I understand correctly, we don't need changes introduced by this PR, because we can use postcss-html as --custom-syntax. And it should be compatible with autofixing.

The main purpose of this PR is to allow users to be compatible with most of the syntax under the default configuration.
By my side, most CSS authors do not have the ability to understand command-line configuration.

gucong3000:
> You removed syntax detection by file extensions. Why? We need this feature.

'postcss-html' will prioritize detection syntax by file content, and if the file is not HTML or Markdown, it will detect the syntax by file extension.

JosefJezek:
@hudochenkov Please any solve.

gucong3000:
Any plan?

evilebottnawi:
@gucong3000 rebase to master please, also be great squash some commits, hard to review

gucong3000:
> rebase to master please, also be great squash some commits, hard to review
@evilebottnawi #2784

ntwb:
Fixed the `package-lock.json` conflict in https://github.com/stylelint/stylelint/pull/2596/commits/85fa160dfd1a12d985bce6e72b672635950da0c9

@evilebottnawi @hudochenkov any new thoughts based on the recent commits please?

gucong3000:
What can I do to merge this PR?

lib/getPostcssResult.js
```diff
@@ -1,13 +1,25 @@
...
 const path = require("path");
 const postcss = require("postcss");
-const scssSyntax = require("postcss-scss");
-const sugarssSyntax = require("sugarss");
+const safeParser = require("postcss-safe-parser");
+const sugarss = require("sugarss");
+const syntaxes /*: {
+  [syntaxName: string]: postcss$syntax,
+}*/ = {
+  // sass: require("postcss-sass"),
```
evilebottnawi: Remove this, right now we don't support `sass`,  it should be added in separate PR

lib/getPostcssResult.js
```diff
@@ -53,25 +65,22 @@ module.exports = function(
...
+      } else if (syntax) {
+        syntax = syntaxes[syntax];
+        if (!syntax) {
           throw new Error(
             "You must use a valid syntax option, either: scss, less or sugarss"
           );
         }
+      } else {
+        syntaxes.css.parse = stylelint._options.fix
+          ? safeParser
```
evilebottnawi: Need tests around `safeParser`

lib/getPostcssResult.js
```diff
@@ -1,13 +1,25 @@
...
-const scssSyntax = require("postcss-scss");
-const sugarssSyntax = require("sugarss");
+const safeParser = require("postcss-safe-parser");
+const sugarss = require("sugarss");
+const syntaxes /*: {
+  [syntaxName: string]: postcss$syntax,
+}*/ = {
+  // sass: require("postcss-sass"),
+  scss: require("postcss-scss"),
+  less: require("postcss-less"),
```
evilebottnawi: Move require on top, it does not make sense because `require` is sync operation

gucong3000:
Close for code rollback.

Issue #2607, opened 2017-06-04
Latest master branch doesn't work in Node 4


hudochenkov:
I was testing `stylelint-order` on latest stylelint `master` and seems like stylelint doesn't work in Node 4.

Installed locally (using [`n`](https://github.com/tj/n)) and in Travis:

```
node v4.8.3
npm 2.15.11
```

If I clone stylelint repository locally, then do `npm install` and `npm test`, then no problems. That what Travis do for us as well.

In different scenarious, when stylelint is dependency, things are getting worse.

**Scenario 1**

Jest test for stylelint-order.

**Scenario 2**

```bash
npm init
npm i github:stylelint/stylelint
node run.js
```

**Scenario 3**

```bash
npm init
npm i github:stylelint/stylelint
node_modules/.bin/stylelint test.css
```

**Scenario 4**

```bash
npm init
npm i ../stylelint # stylelint in sibling folder
node run.js
```

---

```js
// run.js
'use strict';

const stylelint = require('stylelint');

const config = {
	rules: {
		"indentation": "tab"
	},
};

stylelint.lint({
	formatter: 'string',
	code: `
		a {
			display: none;
			--width: 10px;
		}
	`,
	config,
}).then(function (data) {
	console.log(data.output);
	console.log('-----------------');
	console.log(data.results[0].warnings);
	console.log('Done.');
}, function (err) {
	console.log(err);
});
```

Every time I get this error:

```js
$ node run
module.js:327
    throw err;
    ^

Error: Cannot find module 'pify'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.Module._load (module.js:276:25)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/Users/aleks/Downloads/test/sdsdf/node_modules/stylelint/lib/standalone.js:3:14)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
```

I had no luck to find what is going on :( Can we add tests to detect problems like this? That tool doesn't work for users. Because we have green tests, when in fact stylelint doesn't work in Node 4. Maybe there is some kind of problem with npm@2.

Node 6 and 8 work as expected.

hudochenkov:
Fixed in #2614.

PR #2614, opened 2017-06-09
Move pify from devDependencies to dependencies

m-allanson:

#2607


m-allanson:
@hudochenkov Maybe this will work? I haven't tested it in all scenarios from your issue

hudochenkov:
Hm... why it's a devDependency if it's regular dependency? https://github.com/stylelint/stylelint/blob/e1700b47d3a284393723d52f3d3786d6bc7eeb4d/lib/standalone.js#L3

I don't think we should remove it, because we need this dependency.

jeddy3:
> I don't think we should remove it, because we need this dependency.

@m-allanson Good spot. It looks like `pify` was accidentally added as a `devDependency`, rather than a `dependency`. This PR moves it into `dependency`.

hudochenkov:
Looks like I was looking in some other place while was writing a comment above. Sorry ¯﻿\﻿_﻿(﻿ツ﻿)﻿_﻿/﻿¯ I'll test this change in a few hours.

hudochenkov:
I believe we shouldn't add anything to chagelog? Or should we? This problem was introduced in `7.10.0` in this [commit](https://github.com/stylelint/stylelint/blob/43b18fe4c56e9f177644b50354fb8efe9fb34a84/package.json#L91). And as I understand it only affects npm@2 users.

jeddy3:
It's a fix of a regression, right? Best add it to the changelog so that it's discoverable if someone runs into the issue when using `7.10.0`

> - Fixed: "Cannot find module 'pify'" regression in node@4 with npm@2.

hudochenkov:
Thanks. Done.

Issue #2513, opened 2017-04-19
Deprecate `function-url-data-uris` in favor of `function-url-scheme-blacklist`


evilebottnawi:
- [x] Create `function-url-scheme-blacklist`
- [x] Deprecate `function-url-data-uris`

> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Deprecated `function-url-data-uris` in favor `function-url-scheme-blacklist`.

> Which rule, if any, is this issue related to?

If i enable `function-url-data-uris` it is disable use `data` in url, but i can use `function-url-scheme-whitelist` without allowed `data` scheme and get two errors (intersection of rules). I think more right deprecated `function-url-data-uris` and add `function-url-scheme-blacklist` where i can disable `data` and maybe other scheme like as `ftp`, `file` and etc.

In now only `function-url-scheme-whitelist` don't have pair and it is should be `function-url-scheme-blacklist`.

> What CSS is needed to reproduce this issue?

no matter

> What stylelint configuration is needed to reproduce this issue?

no matter

> Which version of stylelint are you using?

latest

> How are you running stylelint: CLI, PostCSS plugin, Node API?

no matter

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

no matter

> What did you expect to happen?

no matter

> What actually happened (e.g. what warnings or errors you are getting)?

no matter

/cc @stylelint/core

davidtheclark:
Sounds correct to me! 👩‍🚀

hudochenkov:
Sounds right.

jeddy3:
Agreed. I was sure we were done with deprecations before `8.0.0`, but it seems there will be a few more (e.g. the `selector-max-*` rules replacing the `selector-no-*` ones). As such, let's make sure we do this deprecation in a new `deprecations-round-2` branch so that we can bundle them all into the same release to minimise inconvenience to the user.

jeddy3:
FYI, I'm WIP on deprecating the old `function-url-data-uris` rule.

Issue #2528, opened 2017-04-25
Deprecate selector-no-* rules in favour of selector-max-*


jeddy3:
Ref: https://github.com/stylelint/stylelint/pull/2474

I did an audit of the 50 or so `-no-` rules. The good news is that it is only some the `selector-no-*` rules that are applying a limit and would be better suited as `-max-` rules.

I believe these are:

Create new rules:

- [x] `selector-no-attribute` -> `selector-max-attribute` ("attribute" is shorthand for _attribute selector**s**_)
- [x] `selector-no-combinator` -> `selector-max-combinators` (pural as not shorthand)
- [x] `selector-no-id` -> `selector-max-id`
- [x] `selector-no-type`  -> `selector-max-type`
- [x] `selector-no-universal`  -> `selector-max-universal`

Deprecate old rules:

- [x] `selector-no-attribute`
- [x] `selector-no-combinator`
- [x] `selector-no-id`
- [x] `selector-no-type`
- [x] `selector-no-universal`

(Additionally, there's scope for a new rule within this system - `selector-list-max-selectors`, but that's for another time).

I think we'll want to make these deprecations at the same time as https://github.com/stylelint/stylelint/issues/2513 - and we'd remove the rules in `8.0.0`.

@stylelint/core Thoughts?

@jimbo Would you be willing to help out with these rules? I believe you're in an ideal position after your work on `selector-max-class`.

evilebottnawi:
@jeddy3 sgtm

davidtheclark:
👌

jimbo:
@jeddy3 Certainly. I'll be traveling for most of May, but I'll have time here and there to work on them, and the new rules will probably be very similar to `selector-max-class` anyway.

jeddy3:
Excellent. I've labelled as `status: help wanted`. 

I'll also add `beginner friendly` as contributing to this issue would be a good my-first-pr.

Anyone feel free to jump in. Just drop a note here about which rule you're working on so that effort is not duplicated.

evilebottnawi:
`selector-max-attribute` wip

evilebottnawi:
`selector-max-universal` wip

evilebottnawi:
`selector-max-combinators` wip

evilebottnawi:
:tada: `selector-no-type` wip

evilebottnawi:
@stylelint/core before close this issue, after all will be merged, please twice check code, docs and tests, I did it quickly and we could have missed something

[...]

PR #2626, opened 2017-06-12
Added: `function-url-scheme-blacklist` rule.

evilebottnawi:

#2513

No, it's self explanatory.

### DIFF
--- a/lib/rules/function-url-scheme-blacklist/README.md
+++ b/lib/rules/function-url-scheme-blacklist/README.md
@@ -18,33 +18,29 @@ This rule ignores URL arguments with variables or variable interpolation (`$sass
 
 ## Options
 
-`array|string`: `["array", "of", "schemes"]|"scheme"`
+`array|string|regex`: `["array", "of", "schemes" or "regex"]|"scheme"|/regex/`
 
 Given:
 
 ```js
-["http", "ftp"]
+["ftp", "/^http/"]
 ```
 
 The following patterns are considered violations:
+
+```css
+a { background-image: url('ftp://www.example.com/file.jpg'); }
+```
 
 ```css
 a { background-image: url('http://www.example.com/file.jpg'); }
 ```
 
 ```css
-a { background-image: url('ftp://www.example.com/file.jpg'); }
-```
-
-The following patterns are *not* considered violations:
-
-```css
 a { background-image: url('https://www.example.com/file.jpg'); }
 ```
 
-```css
-a { background-image: url('HTTPS://www.example.com/file.jpg'); }
-```
+The following patterns are *not* considered violations:
 
 ```css
 a { background-image: url('data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs='); }
--- a/lib/rules/function-url-scheme-blacklist/README.md
+++ b/lib/rules/function-url-scheme-blacklist/README.md
@@ -1,6 +1,6 @@
-# function-url-scheme-whitelist
+# function-url-scheme-blacklist
 
-Specify a whitelist of allowed URL schemes.
+Specify a blacklist of disallowed URL schemes.
 
 ```css
 a { background-image: url('http://www.example.com/file.jpg'); }
@@ -23,7 +23,7 @@ This rule ignores URL arguments with variables or variable interpolation (`$sass
 Given:
 
 ```js
-["https", "data"]
+["http", "ftp"]
 ```
 
 The following patterns are considered violations:
--- a/lib/rules/function-url-scheme-blacklist/__tests__/index.js
+++ b/lib/rules/function-url-scheme-blacklist/__tests__/index.js
@@ -1,0 +1,218 @@
+"use strict"
+
+const messages = require("..").messages
+const ruleName = require("..").ruleName
+const rules = require("../../../rules")
+
+const rule = rules[ruleName]
+
+testRule(rule, {
+  ruleName,
+  config: [ "https", "data" ],
+
+  accept: [ {
+    code: "a { background: url(); }",
+  }, {
+    code: "a { background: url(''); }",
+  }, {
+    code: "a { background: url(\"\"); }",
+  }, {
+    code: "a { background: url(:); }",
+  }, {
+    code: "a { background: url(://); }",
+  }, {
+    code: "a { background: url(//); }",
+  }, {
+    code: "a { background: url(/); }",
+  }, {
+    code: "a { background: url(./); }",
+  }, {
+    code: "a { background: url(./file.jpg); }",
+  }, {
+    code: "a { background: url(../file.jpg); }",
+  }, {
+    code: "a { background: URL(../file.jpg); }",
+  }, {
+    code: "a { background: url('../file.jpg'); }",
+  }, {
+    code: "a { background: url(\"../file.jpg\"); }",
+  }, {
+    code: "a { background: url('/path/to/file.jpg'); }",
+  }, {
+    code: "a { background: url(//www.example.com/file.jpg); }",
+  }, {
+    code: "a { background: url(\"//www.example.com/file.jpg\"); }",
+  }, {
+    code: "a { background: url('http://www.example.com/file.jpg'); }",
+  }, {
+    code: "a { background-image: url('http://example.com:3000'); }",
+  }, {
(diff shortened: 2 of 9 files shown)


lib/rules/function-url-scheme-blacklist/README.md
```diff
@@ -0,0 +1,67 @@
+# function-url-scheme-blacklist
+
+Specify a blacklist of allowed url schemes.
```
jeddy3: Specify a blacklist of disallowed url schemes.

evilebottnawi:
@jeddy3 stupid error, sorry

lib/rules/function-url-scheme-blacklist/index.js
```diff
@@ -0,0 +1,76 @@
...
+const report = require("../../utils/report")
+const ruleMessages = require("../../utils/ruleMessages")
+const validateOptions = require("../../utils/validateOptions")
+const _ = require("lodash")
+const parse = require("url").parse
+
+const ruleName = "function-url-scheme-blacklist"
+
+const messages = ruleMessages(ruleName, {
+  rejected: scheme => `Unexpected url scheme "${scheme}:"`,
```
hudochenkov: Upper case “URL”.

lib/rules/function-url-scheme-blacklist/index.js
```diff
@@ -0,0 +1,76 @@
...
+})
+
+const rule = function (blacklist) {
+  return (root, result) => {
+    const validOptions = validateOptions(result, ruleName, {
+      actual: blacklist,
+      possible: [_.isString],
```
hudochenkov: Readme says it accepts `array|string`. I'm not sure if it checks that. There're no tests if only one `string` was in config (not array with one `string`).

lib/rules/function-url-scheme-blacklist/__tests__/index.js
```diff
@@ -0,0 +1,218 @@
...
+    code: "a { background: no-repeat center/80% url('https://www.example.com/file.jpg'); }",
+    message: messages.rejected("https"),
+    line: 1,
+    column: 42,
+  } ],
+})
+
+testRule(rule, {
+  ruleName,
+  config: [[]],
```
hudochenkov: Why this test group rejects `https` and `data`? Is it an undocumented behaviour?

lib/rules/function-url-scheme-blacklist/__tests__/index.js
```diff
@@ -0,0 +1,218 @@
...
+    code: "a { background-image: url('data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs='); }",
+    message: messages.rejected("data"),
+    line: 1,
+    column: 27,
+  } ],
+})
+
+testRule(rule, {
+  ruleName,
+  config: [""],
```
hudochenkov: Why this test group rejects `https` and `data`? Is it an undocumented behaviour?

evilebottnawi: @hudochenkov this used if your want add all schemes to blacklist analogy in whitelist

lib/rules/function-url-scheme-blacklist/__tests__/index.js
evilebottnawi: @hudochenkov this used if your want add all schemes to blacklist analogy in whitelist

lib/rules/function-url-scheme-blacklist/index.js
evilebottnawi: @hudochenkov we use this in all `*-blacklist/*-whitelist` rules.

lib/rules/function-url-scheme-blacklist/index.js
evilebottnawi: done, also in `whitelist` rule.

lib/rules/function-url-scheme-blacklist/__tests__/index.js
evilebottnawi: @hudochenkov we already have same tests in `function-url-scheme-whitelist`

lib/rules/function-url-scheme-blacklist/__tests__/index.js
hudochenkov: But still behaviour isn't documented anywhere. We should document this.

lib/rules/function-url-scheme-blacklist/__tests__/index.js
hudochenkov: But still behaviour isn't documented anywhere. We should document this.

evilebottnawi: @hudochenkov can your help me with this, I would like the current time to concentrate on the new rules (https://github.com/stylelint/stylelint/issues/2528) and I'm bad at documentation :disappointed:

jeddy3: @hudochenkov Good catch! Sorry I missed this. This isn't the correct behaviour. Both `[""]` and `[[]]` should _allow all_ schemes. 

This is how the other `-blacklist` work. `-whitelist` work differently. 

It's useful to think as

`blacklist: []` = "blacklist nothing" i.e. "everything is ok"
`whitelist: []` = "whitelist nothing" i.e. "nothing is ok".

evilebottnawi:
@jeddy3 can your help with documentation this?

jeddy3:
> @jeddy3 can your help with documentation this?

It's a behaviour issue, not a documentation one. I'm pushing up a commit to fix though.

jeddy3:
- I've added to tests to the `property-*list` rules to show how the whitelist and blacklist rule behaviours are _correctly_ different when given empty strings or arrays as primary options.
- This behaviour comes for free when using the `matchesStringOrRegExp` util. I've updated `function-url-scheme-blacklist` to use it.
- Likewise, support for regex also comes for free when using `matchesStringOrRegExp`.
- I've removed the tests for case insensitive _primary options_. None of the other `*-*list` rules have case-insensitive _primary options_. If we want to revisit this behaviour, so should do so across the board and in a separate issue.

evilebottnawi:
@jeddy3 After commits what are my actions here? Or just wait when @hudochenkov do review?

jeddy3:
> Or just wait when @hudochenkov do review?

Just wait on @hudochenkov as my commits came off the back of his original review.

lib/rules/function-url-scheme-blacklist/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+```css
+a { background-image: url('http://www.example.com/file.jpg'); }
+/**                        ↑
+ *           This URL scheme */
+```
+
+A [URL scheme](https://url.spec.whatwg.org/#syntax-url-scheme) consists of alphanumeric, `+`, `-`, and `.` characters. It can appear at the start of a URL and is followed by `:`.
+
+This rule treats URL schemes as case insensitive (`https` and `HTTPS` are the same).
```
hudochenkov: > I've removed the tests for case insensitive primary options. None of the other *-*list rules have case-insensitive primary options. If we want to revisit this behaviour, so should do so across the board and in a separate issue. — @jeddy3 

Is this line in readme still valid?

jeddy3: Yes. I've added tests to show this.

Like all the other `*-*list` rules, this rule now _only_ accepts lowercase primary options but is case insensitive _when checking for matches_.

So `function-url-scheme-blacklist: "http"` will disallow both `http://g.co/x.jpg` and `HTTP://g.co/x.jpg`.

https://github.com/stylelint/stylelint/issues/2660 is about allowing the user to specify primary options in _any case_ e.g. `function-url-scheme-blacklist: "HTTP"`, or `function-blacklist: "translateX"`.

jeddy3:
> And again about "" and [] configurations. We won't document them?

I'm not sure there's any need to do that now because the behaviour is intuitive. We don't document this behaviour in any of the other `*-blacklist` rules for this reason.

`scheme-blacklist: []` = "blacklist nothing" i.e. "all schemes are ok" seems pretty clear to me.

jeddy3:
Changelog 

> -   Added: `function-url-scheme-blacklist` rule ([#2626](https://github.com/stylelint/stylelint/pull/2626)).

PR #2628, opened 2017-06-12
Added: `selector-max-attribute` rule.

evilebottnawi:

#2528

No, it's self explanatory.

Some Fixes in test `selector-max-class`:
1. Remove properties, because it is unused for tests.
2. Set `config: [0]` for less/sass syntax, because this construction should be ignored **always**, using `1` potential can to skip something, what we should ignore.
3. `.ab { .cd: { left: 0; top: 0; } }` -> `.ab { .cd { left: 0; top: 0; } }` (seems just typo)
4. `.ab, .cd { &:hover > .ef.gh { left: 0; top: 0; } }` -> `.ab { &:hover > .ef.gh { left: 0; top: 0; } }`, because for this case we have two errors, one for `.ab`, second for `.cd`. @davidtheclark @hudochenkov @jeddy3 can we throw error if we have more than one error on test block?

### DIFF
--- a/docs/user-guide/rules.md
+++ b/docs/user-guide/rules.md
@@ -162,6 +162,7 @@ Here are all the rules within stylelint, grouped by the [*thing*](http://apps.wo
 -   [`selector-combinator-space-before`](../../lib/rules/selector-combinator-space-before/README.md): Require a single space or disallow whitespace before the combinators of selectors.
 -   [`selector-descendant-combinator-no-non-space`](../../lib/rules/selector-descendant-combinator-no-non-space/README.md): Disallow non-space characters for descendant combinators of selectors.
 -   [`selector-id-pattern`](../../lib/rules/selector-id-pattern/README.md): Specify a pattern for id selectors.
+-   [`selector-max-attribute`](../../lib/rules/selector-max-attribute/README.md): Limit the number of attribute selectors in a selector.
 -   [`selector-max-class`](../../lib/rules/selector-max-class/README.md): Limit the number of classes in a selector.
 -   [`selector-max-compound-selectors`](../../lib/rules/selector-max-compound-selectors/README.md): Limit the number of compound selectors in a selector.
 -   [`selector-max-specificity`](../../lib/rules/selector-max-specificity/README.md): Limit the specificity of selectors.
--- a/lib/rules/selector-max-attribute/README.md
+++ b/lib/rules/selector-max-attribute/README.md
@@ -1,0 +1,77 @@
+# selector-max-attribute
+
+Limit the number of attribute selectors in a selector.
+
+```css
+    [rel="external"] {}
+/** ↑
+ * This type of selector */
+```
+
+This rule resolves nested selectors before counting the number of attribute selectors. Each selector in a [selector list](https://www.w3.org/TR/selectors4/#selector-list) is evaluated separately.
+
+The `:not()` pseudo-class is also evaluated separately. The rule processes the argument as if it were an independent selector, and the result does not count toward the total for the entire selector.
+
+## Options
+
+`int`: Maximum attribute selectors allowed.
+
+For example, with `2`:
+
+The following patterns are considered violations:
+
+```css
+[type="number"][name="quality"][data-attribute="value"] {}
+```
+
+```css
+[type="number"][name="quality"][disabled] {}
+```
+
+```css
+[type="number"][name="quality"] {
+  & [data-attribute="value"] {}
+}
+```
+
+```css
+[type="number"][name="quality"] {
+  & [disabled] {}
+}
+```
+
+```css
+[type="number"][name="quality"] {
+  & > [data-attribute="value"] {}
+}
+```
+
+```css
+/* `[type="text"][data-attribute="value"][disabled]` is inside `:not()`, so it is evaluated separately */
+input:not([type="text"][data-attribute="value"][disabled]) {}
+```
+
+The following patterns are *not* considered violations:
+
+```css
+[type="text"] {}
+```
+
+```css
+[type="text"][name="message"] {}
+```
+
+```css
+[type="text"][disabled]
+```
+
+```css
+/* each selector in a selector list is evaluated separately */
+[type="text"][name="message"],
+[type="number"][name="quality"] {}
+```
+
+```css
+/* `[disabled]` is inside `:not()`, so it is evaluated separately */
+[type="text"][name="message"]:not([disabled]) {}
+```
--- a/lib/rules/selector-max-attribute/__tests__/index.js
+++ b/lib/rules/selector-max-attribute/__tests__/index.js
@@ -1,0 +1,190 @@
+"use strict"
+
+const messages = require("..").messages
+const ruleName = require("..").ruleName
+const rules = require("../../../rules")
+
+const rule = rules[ruleName]
+
+// Sanity checks
+testRule(rule, {
+  ruleName,
+  config: [0],
+
+  accept: [ {
+    code: "foo {}",
+  }, {
+    code: ".bar {}",
+  }, {
+    code: "foo .bar {}",
+  }, {
+    code: ":root { --foo: 1px; }",
+    description: "custom property in root",
+  }, {
+    code: "html { --foo: 1px; }",
+    description: "custom property in selector",
+  }, {
+    code: ":root { --custom-property-set: {} }",
+    description: "custom property set in root",
+  }, {
+    code: "html { --custom-property-set: {} }",
+    description: "custom property set in selector",
+  } ],
+
+  reject: [ {
+    code: "[foo] {}",
+    message: messages.expected("[foo]", 0),
+    line: 1,
+    column: 1,
+  }, {
+    code: "a[rel=\"external\"] {}",
+    message: messages.expected("a[rel=\"external\"]", 0),
+    line: 1,
+    column: 1,
+  }, {
+    code: "a, .foo[type=\"text\"] {}",
+    message: messages.expected(".foo[type=\"text\"]", 0),
+    line: 1,
+    column: 4,
+  }, {
+    code: "a > [foo] {}",
+    message: messages.expected("a > [foo]", 0),
+    line: 1,
+    column: 1,
+  }, {
+    code: "a[rel='external'] {}",
+    message: messages.expected("a[rel='external']", 0),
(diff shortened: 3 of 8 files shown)


docs/user-guide/rules.md
```diff
@@ -162,6 +162,7 @@ Here are all the rules within stylelint, grouped by the [*thing*](http://apps.wo
 -   [`selector-combinator-space-before`](../../lib/rules/selector-combinator-space-before/README.md): Require a single space or disallow whitespace before the combinators of selectors.
 -   [`selector-descendant-combinator-no-non-space`](../../lib/rules/selector-descendant-combinator-no-non-space/README.md): Disallow non-space characters for descendant combinators of selectors.
 -   [`selector-id-pattern`](../../lib/rules/selector-id-pattern/README.md): Specify a pattern for id selectors.
+-   [`selector-max-attribute`](../../lib/rules/selector-max-attribute/README.md): Limit the number of attributes in a selector.
```
jeddy3: "Limit the number of attribute selectors in a selector."

lib/rules/selector-max-attribute/README.md
```diff
@@ -0,0 +1,70 @@
+# selector-max-attribute
+
+Limit the number of attributes in a selector.
```
jeddy3: "Limit the number of attribute selectors in a selector."

lib/rules/selector-max-attribute/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+```css
+    [rel="external"] {}
+/** ↑
+ * This type of selector */
+```
+
+This rule resolves nested selectors before counting the number of classes in a selector. Each selector in a [selector list](https://www.w3.org/TR/selectors4/#selector-list) is evaluated separately.
```
jeddy3: This rule resolves nested selectors before counting the number of attribute selectors. Each selector in a [selector list](https://www.w3.org/TR/selectors4/#selector-list) is evaluated separately.

lib/rules/selector-max-attribute/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+The `:not()` pseudo-class is also evaluated separately. The rule processes the argument as if it were an independent selector, and the result does not count toward the total for the entire selector.
+
+## Options
+
+`int`: Maximum attribute allowed.
```
jeddy3: Maximum attribute selectors allowed.

lib/rules/selector-max-attribute/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+```css
+[type="text"][name="message"] {}
+```
+
+```css
+[type="text"][disabled]
+```
+
+```css
```
jeddy3: Let's put the comment on top. Long lines are hard to me.

```css
/* each selector in a selector list is evaluated separately */
[type="text"][name="message"],
[type="number"][name="quality"] {}
```

lib/rules/selector-max-attribute/README.md
```diff
@@ -0,0 +1,70 @@
...
+[type="text"][name="message"],
+[type="number"][name="quality"] {} /* each selector in a selector list is evaluated separately */
+```
+
+```css
```
jeddy3: Should this not be a selector list? e.g.

```css
/* [data-attribute="value"]` is inside `:not()`, so it is evaluated separately */
[type="text"][name="message"]:not([data-attribute="value"]) {} 
```

lib/rules/selector-max-attribute/index.js
```diff
@@ -0,0 +1,75 @@
...
+const parseSelector = require("../../utils/parseSelector")
+const report = require("../../utils/report")
+const ruleMessages = require("../../utils/ruleMessages")
+const validateOptions = require("../../utils/validateOptions")
+const resolvedNestedSelector = require("postcss-resolve-nested-selector")
+
+const ruleName = "selector-max-attribute"
+
+const messages = ruleMessages(ruleName, {
+  expected: (selector, max) => `Expected "${selector}" to have no more than ${max} selector attribute`,
```
jeddy3:
```js
`Expected "${selector}" to have no more than ${max} attribute selector(s)`
```

lib/rules/selector-max-attribute/__tests__/index.js
```diff
@@ -0,0 +1,180 @@
...
+    code: "[type=\"text\"], \n[type=\"number\"] {}",
+    description: "multiple selectors: fewer than max classes",
+  }, {
+    code: "[type=\"text\"][name=\"message\"], \n[type=\"password\"][name=\"password\"] {}",
+    description: "multiple selectors: exactly max classes",
+  }, {
+    code: "[type=\"text\"][disabled], \n[type=\"password\"][disabled] {}",
+    description: "multiple selectors: exactly max classes",
+  }, {
+    code: "[type=\"text\"][name=\"message\"] :not([type=\"number\"][name=\"quality\"]) {}",
```
jeddy3: This this be?:

```
[type=\"text\"][name=\"message\"]:not([type=\"number\"][name=\"quality\"]) {}
```

I'm not sure a compounded `:not` pseudo is valid CSS.

lib/rules/selector-max-attribute/__tests__/index.js
```diff
@@ -0,0 +1,180 @@
...
+  }, {
+    code: "[type=\"text\"][name=\"message\"][data-attribute=\"value\"] :not([data-attribute-2=\"value\"]) {}",
+    description: ":not(): greater than max classes, outside",
+    message: messages.expected("[type=\"text\"][name=\"message\"][data-attribute=\"value\"] :not([data-attribute-2=\"value\"])", 2),
+    line: 1,
+    column: 1,
+  }, {
+    code: "[type=\"text\"][name=\"message\"][data-attribute=\"value\"] :not([disabled]) {}",
+    description: ":not(): greater than max classes, outside",
+    message: messages.expected("[type=\"text\"][name=\"message\"][data-attribute=\"value\"] :not([disabled])", 2),
```
jeddy3: Should be?:

```css
[type=\"text\"][name=\"message\"][data-attribute=\"value\"]:not([disabled])
```

lib/rules/selector-max-attribute/README.md
evilebottnawi: @jeddy3 not, example of list above, this example show attributes selectors and attributes selectors in `not` calculate separately

lib/rules/selector-max-attribute/README.md
```diff
@@ -0,0 +1,82 @@
...
+[type="text"][disabled]
+```
+
+```css
+/* each selector in a selector list is evaluated separately */
+[type="text"][name="message"],
+[type="number"][name="quality"] {}
+```
+
+```css
```
jeddy3: I think we can condense the following two examples into just:

```css
/* `[disabled]` is inside `:not()`, so it is evaluated separately */
[type="text"][name="message"]:not([disabled]) {}
```

As that shows _3_ attribute selectors passing because the third attribute selector (`[disabled]`) is within a `:not()`. I think having a fourth attribute selector doesn't add anything to this example.

I think this makes the final example superfluous and can be deleted.

evilebottnawi: @jeddy3 oh, yep, let's remove

evilebottnawi:
@jeddy3 done

lib/rules/selector-max-attribute/README.md
jeddy3: And change the other example to:

```
/* `[disabled]` is inside `:not()`, so it is evaluated separately */
[type="text"][name="message"]:not([disabled]) {}
```

So there are 3, and not 4, attribute selectors.

evilebottnawi:
@jeddy3 seems done, sorry, for stupid error, I'm tired today, but I want to release `8.0.0` version faster

jeddy3:
> but I want to release 8.0.0 version faster

I think we all do :) It feels close now though.

Thanks again for your effort with these rules!

hudochenkov:
> can we throw error if we have more than one error on test block?

It's not supported in our `jest-setup.js`. I believe it's possible to do, if it's really necessary.

lib/rules/selector-max-attribute/index.js
```diff
@@ -0,0 +1,75 @@
...
+const parseSelector = require("../../utils/parseSelector")
+const report = require("../../utils/report")
+const ruleMessages = require("../../utils/ruleMessages")
+const validateOptions = require("../../utils/validateOptions")
+const resolvedNestedSelector = require("postcss-resolve-nested-selector")
+
+const ruleName = "selector-max-attribute"
+
+const messages = ruleMessages(ruleName, {
+  expected: (selector, max) => `Expected "${selector}" to have no more than ${max} attribute selector(s)`,
```
jeddy3: I agree that this is nicer. However, I believe this style ("xxx(s)") is used for all `max` rules. Therefore I suggest we create a separate issue and make this change, at some point, across the board for the sake of consistency.

lib/rules/selector-max-attribute/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+The `:not()` pseudo-class is also evaluated separately. The rule processes the argument as if it were an independent selector, and the result does not count toward the total for the entire selector.
+
+## Options
+
+`int`: Maximum attribute selectors allowed.
+
+For example, with `2`:
+
+The following patterns are considered warnings:
```
jeddy3: Oops, didn't spot this. I had forgotten about this change :)

@evilebottnawi Please change to: "The following patterns are considered violations:"

evilebottnawi:
/cc @hudochenkov @jeddy3 
Already some fixes in `selector-max-class`

hudochenkov:
@evilebottnawi I think squashing commits should be done after PR reviewed and accepted. This way it would be easier for reviewers to check what was changed between reviews ;)

evilebottnawi:
@hudochenkov sometimes there area lot of changes and commits and it is create noise for PR (imho), seems this PR small and squashing it is not bad

jeddy3:
@evilebottnawi Thanks for this!

Changelog:

> -   Added: `selector-max-attribute` rule ([#2628](https://github.com/stylelint/stylelint/pull/2628)).

Let's _not_ squash commit during the review process. Like @hudochenkov I find it much easier to review changes when they are _not_ squashed.

Issue #2627, opened 2017-06-12
TypeError: Cannot read property 'syntax' of undefined with --fix


dptoot:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug where if the --fix flag is used as a CLI argument then the following error results

```
> husky - npm run -s prepush
> husky - node v6.10.0

TypeError: Cannot read property 'syntax' of undefined
    at stylelint._lintSource.then.postcssResult (/data/work/axl/node_modules/stylelint/lib/standalone.js:156:74)
```

Removal of the --fix flag causes stylelint to work as expected

> Which rule, if any, is this issue related to?

n/a

> What CSS is needed to reproduce this issue?
n/a

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "extends": "stylelint-config-standard",
  "rules": {
    "indentation": 4,
  }
}
```

> Which version of stylelint are you using?

└── stylelint@7.11.0

> How are you running stylelint: CLI, PostCSS plugin, Node API?

"lint:css": "stylelint \"packages/**/*.scss\" --syntax scss --fix",

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

I am linting SCSS but I don't feel it is the issue

> What did you expect to happen?

No errors :)

> What actually happened (e.g. what warnings or errors you are getting)?

e.g. "The following warnings were flagged:"

```shell
> husky - npm run -s prepush
> husky - node v6.10.0

TypeError: Cannot read property 'syntax' of undefined
    at stylelint._lintSource.then.postcssResult (/data/work/axl/node_modules/stylelint/lib/standalone.js:156:74)
```

evilebottnawi:
Seems related to https://github.com/stylelint/stylelint/issues/2619

evilebottnawi:
@dptoot weird, locally all work fine, can your provide test minimal reproducible repo?

hudochenkov:
I can't reproduce it either with fresh stylelint install.

@dptoot @jitendravyas could you, please, show us your installed modules? Run `npm list` in directory where stylelint is installed and copy at least stylelint part (with all its dependencies).

And test minimal reproducible repo would be also much appreciated.

jitendravyas:
I have three question before. 

When we use Stylelint CLI do we need to specificy the path of stylelintrc in cli command? 

Can we auto-fix a rule using cli without stylelintrc?

[...]

PR #2652, opened 2017-06-19
Do not try to write back ignored files after --fix.

subzey:

Fixes issue #2627

Accessing `opts.syntax` throws if the file is ignored. _I wonder why Flow didn't catch that?_


hudochenkov:
Thank you for a fix! Would you mind adding test for this issue, please? You can add it to [`fix.test.js`](https://github.com/stylelint/stylelint/blob/master/system-tests/fix/fix.test.js).

subzey:
@hudochenkov Looks like fix.test.js contains tests for lib version of stylelint (`lib/index`), and I need to test the standalone script (`lib/standalone`). I'm not really sure how to do that, in fact I can see no tests that contains `"standalone"` to start off from.

I definitely need help with tests.

subzey:
Added a test. Not sure I did it right, but at least linters are okay, and it's green with the proposed fix code and red with the old one

hudochenkov:
Thank you for thinking outside the box for test :)

I came up with this test:

```js
it("doesn't write to ignored file", () => {
  return stylelint.lint({
    files: [stylesheetPath],
    config: {
      ignoreFiles: stylesheetPath,
      rules: {
        "at-rule-name-case": "lower",
        "comment-empty-line-before": "always",
        "comment-no-empty": true,
      },
    },
    fix: true,
  }).then(() => {
    return pify(fs.readFile)(stylesheetPath, "utf8").then((newFile) => {
      return pify(fs.readFile)(path.join(__dirname, "stylesheet.css"), "utf8").then((oldFile) => {
        expect(newFile).toBe(oldFile)
      })
    })
  })
})
```

I still don't understand promises, so this why it's ugly as hell. If someone could improve code it would be much appreciated :)

evilebottnawi:
@hudochenkov promise is very ugly if your don't use `async/await` :smile:

subzey:
@hudochenkov Actually, I've almost copy-pasted the code chunk from https://github.com/stylelint/stylelint/blob/master/system-tests/004/004.test.js#L11-L14

jeddy3:
@hudochenkov Is this good to go in for `7.12` or did you want to change the test? Is there any advantage to either approach over the other?

hudochenkov:
@jeddy3 it's good to merge, if test looks ok.

> Is there any advantage to either approach over the other?

Test in this PR looks quite complicated. While my version might be improved. I didn't ask to use my test, because not sure if they are better than @subzey's test.

If you accept author's test, then feel free to merge.

jeddy3:
Let's merge and then I'll drop in your test, if only because it's consistent with the other tests in the file.

jeddy3:
Changelog:

> -   Fixed: `--fix` no longer crashes when used with ignored files ([#2652](https://github.com/stylelint/stylelint/pull/2652)).

PR #2665, opened 2017-06-21
Added: `selector-max-type` rule

evilebottnawi:

#2528

No, it's self explanatory.

Need resolve:
1. `html { --foo: 1px; }` and `html { --custom-property-set: {} }` should be rejected? I think yes, because we can have `html { --some-here: 1px; --custom-property-set: { color: red; display: block; }; color: red;}` Later we can add option `ignoreContaintCustom` (or best name for this purpose).
2. After resolve selector some non standard (example scss: `@for $n from 1 through 5 { .foo-#{$n} { div { content: \"#{$n}\"; } } }`) output error `.foo-#{$n} div`, it is valid because we contain `div`, but message is very weird. It is normal or not?

Some tests are commented.

### DIFF
--- a/docs/user-guide/example-config.md
+++ b/docs/user-guide/example-config.md
@@ -139,6 +139,7 @@ You might want to learn a little about [how rules are named and how they work to
     "selector-max-empty-lines": int,
     "selector-max-id": int,
     "selector-max-specificity": string,
+    "selector-max-type": int,
     "selector-max-universal": int,
     "selector-nested-pattern": string,
     "selector-no-attribute": true,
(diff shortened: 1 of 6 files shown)


jeddy3:
> `html { --foo: 1px; }` and `html { --custom-property-set: {} }` should be rejected?

Yes, absolutely. Rules should be _most restrictive by default (when it comes to **standard CSS**)_, and then made more permissive with options. Let's only add a `ignoreContaintCustom` option if it is requested. I suspect using the existing `ignoreTypes: ["html"]` option will suffice for most users.

> `.foo-#{$n} div`

Shouldn't this whole selector be ignored by the rule as it's non-standard? We should check that _resolved_ selectors are standard. If not, return early.

hudochenkov:
Non-standard selectors should be ignored, and tests for them should be accepted.

jeddy3:
@hudochenkov Updated this one.

- Added a few extra reject tests to be sure the `ignore` options are working as expected.
- Fixed test descriptions
- Updated rule description
- Added pattern examples

jeddy3:
@hudochenkov Once you've approved my commits. I'll squash them and rebase this off `master`.

hudochenkov:
Why we [don't ignore](https://github.com/stylelint/stylelint/blob/e7ce9f02df8360ea245a2ec2826e3c3dff5e8c3f/lib/rules/selector-max-type/__tests__/index.js#L263-L307) non-standard selectors? 

> > `.foo-#{$n} div`
>
> Shouldn't this whole selector be ignored by the rule as it's non-standard? We should check that _resolved_ selectors are standard. If not, return early.

> Non-standard selectors should be ignored, and tests for them should be accepted.

jeddy3:
> -   Added: `selector-max-type` rule ([#2665](https://github.com/stylelint/stylelint/pull/2665)).

Issue #2061, opened 2016-11-16
Use Yarn


davidtheclark:
Does anybody have any objections to using Yarn here? — which would basically just mean committing a Yarn lockfile?

ntwb:
No objections at all, was on my to do list 👍

Arcanemagus:
Be careful that it is generating a `node_modules` that actually works, I have a few projects where `yarn` can't be used due to bugs in their resolution of dependencies (`linter-stylelint` was one during the `table` mess 😛).

ntwb:
Thanks @Arcanemagus, I'll read up a little more on this as I've got a good half to a dozen projects that I'm looking at to add Yarn support to 👍

Arcanemagus:
Specifically the bug that has prevented me from using it in a few places is https://github.com/yarnpkg/yarn/issues/681.

ntwb:
Thanks for that, I've subscribed to that issue, I don't have any cases currently that this would cause an issue, but thats more luck than anything, I'm frequently pinning packages in repos 👍

ntwb:
https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn

ntwb:
Adding this here on the issue rather than #2129 as these issues should be addressed separately:

• After merging #2129 we should run "Delete all repository caches" via https://travis-ci.org/stylelint/stylelint/caches so that any _legacy_ cache folders in particular relating to our `master` branch are _flushed_.

• @greenkeeperio / @greenkeeperio-bot currently support Yarn, as such when a new PR is submitted by @greenkeeperio the PR will not include the related `yarn.lock` update, we'd have to manually add that to each PR. This is being discussed in https://github.com/greenkeeperio/greenkeeper/issues/314 and the tl'dr is "not anytime soon".

ntwb:
See also https://medium.com/@boennemann/avoid-yarn-for-packages-and-fully-enjoy-its-benefits-for-application-development-8bdd4deb33cf#.2sbh58tof, I may need to read this article a couple more times to fully grok the ramifications of the issues raised though it looks like if we use packages that _also_ use Yarn there are potential issues with semver conflicts that may occur.

davidtheclark:
[...]

PR #2666, opened 2017-06-21
Ignore package-lock.json

hudochenkov:

Kind of related to https://github.com/stylelint/stylelint/issues/2061#issuecomment-303603179

It's uncomfortable to work with stylelint repository when npm@5 is installed. Because npm@5 automatically creates `package-lock.json`. So this file should be constantly manually deleted when switching branches or rebase. It's tedious. While we haven't decided to use `package-lock.json` we could ignore it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules
+package-lock.json
 *.log
 .coverage
 .eslintcache


evilebottnawi:
@hudochenkov no `package-lock.json` before https://github.com/npm/npm/issues/16866 was resolved, each update your should run `rm package-lock.json` and run `npm install/update`, it is hell :disappointed:

hudochenkov:
@evilebottnawi thanks for heads up! Than my PR has even more sense :)

P. S. Do you think it's better to stick to npm@3 on local machine for now? (I've never used npm@4)

evilebottnawi:
@hudochenkov we now use v4, they are very stable and good, local i am use npm v5, someone must find problems :smile:

ntwb:
I ❤️ with npm 5, but all that love doesn't help here if you are using npm 3 or 4 😢 

I see two options:

1. The proposed adding `package-lock.json` to `.gitignore` which alleviates the problem for the repo, it won't alleviate the issue that @evilebottnawi pointing out above at npm when the `package.json` file is updated and `package-lock.json` is not you'll still need to delete the `package-lock.json` file.

2. Add an `.npmrc` file containing `package-lock=false` for now, I'm trying to find if this is backward compatible for npm 3 and 4 though.

ntwb:
https://twitter.com/netweb/status/877661984677781506
> _"@npm_support Is `package-lock.json` compatible with either npm 3 or 4, if I add a `package-lock.json` can npm 3 or 4 benefit from this?"

https://twitter.com/npm_support/status/877671112028684289
> _"@netweb the `package-lock.json` is not compatible"_

https://twitter.com/netweb/status/877671436281757696
> _"@npm_support Thanks, so if I commit it to a repo and a user checks it out using npm 3 or 4 there won't be any explosions, simply ignored?"_

https://twitter.com/npm_support/status/877672905987043330
> _"@netweb npm3 or npm4 will ignore the package-lock.json files"_

Based on that I think we should go ahead and add a `package-lock.json` file this will benefit npm 5 users and npm 3 and npm 4 users will be unaffected.

hudochenkov:
There more issues I found frustrating in npm@5:

- Installing subdep as toplevel dep causes unmet deps (https://github.com/npm/npm/issues/16741)
- --save-dev is removing installed deps (https://github.com/npm/npm/issues/16794)
- npm does not generate full lockfile unless a full install is run, first (https://github.com/npm/npm/issues/16893)
- Missing dependencies after running `npm install` a second time (https://github.com/npm/npm/issues/16839)
- npm install sometimes fails to install some dependencies on a fresh install (https://github.com/npm/npm/issues/16902) — happened to me twice yesterday

I don't mind `package-lock.json` available in our repo. But how to update it if I update dependency and don't have npm@5 installed?

ntwb:
Per the last tweet in my comment above: "...npm3 or npm4 will ignore the package-lock.json files"

Unless npm 5 is used `package-lock.json` can not be updated

Per your comments above @hankthewhale and mine [here](https://github.com/stylelint/stylelint/pull/2670#issuecomment-310305721) these appear to be the least of our worries, let's merge this PR and revisit once npm addresses a few of these concerns

Issue #2643, opened 2017-06-15
Add stylelint-disable commands support to autofix 


dan-gamble:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

I have a file that has `/* stylelint-disable */` `/* stylelint-enable */` wrapped around it but when using `--fix` it tries to format this file and breaks it quite bad.

> Which rule, if any, is this issue related to?

`--fix`

> What CSS is needed to reproduce this issue?

[Before](https://gist.github.com/dan-gamble/1505b0d34c7a30d3ce4c45f98530b8c6)
[After](https://gist.github.com/dan-gamble/0ae41752b61627538372c9e536084fdb)

A few of the issues:

* Double `;;` on `L38`
* No `;` on `L92` (I imagine this is where the `;;` comes from

> What stylelint configuration is needed to reproduce this issue?

[Config](https://gist.github.com/dan-gamble/30039756b9f8f8096a336b688c3e013f)

The config is pretty huge but i don't think it's the problem here

> Which version of stylelint are you using?

`7.11.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI through `lint-staged`

---
- recommended config
  - [x] `declaration-block-no-duplicate-properties`
  - [x] `function-calc-no-unspaced-operator`
- standard config
  - [x] `alpha-value-notation`
  - [x] `at-rule-empty-line-before`
  - [x] `at-rule-no-vendor-prefix`
  - [x] `color-function-notation`
  - [x] `color-hex-length`
  - [x] `comment-empty-line-before`
  - [x] `comment-whitespace-inside`
  - [x] `custom-property-empty-line-before`
  - [x] `declaration-block-no-redundant-longhand-properties`
  - [x] `declaration-empty-line-before`
  - [x] `font-family-name-quotes`
  - [x] `function-name-case`
  - [x] `function-url-quotes`
  - [x] `hue-degree-notation`
  - [x] `import-notation`
  - [x] `keyframe-selector-notation`
  - [x] `length-zero-no-unit`
  - [x] `lightness-notation`
  - [x] `media-feature-range-notation`
  - [x] `media-feature-name-no-vendor-prefix`
  - [x] `property-no-vendor-prefix`
  - [x] `rule-empty-line-before`
  - [x] `selector-attribute-quotes`
  - [x] `selector-no-vendor-prefix`
  - [x] `selector-not-notation`
  - [x] `selector-pseudo-element-colon-notation`
[...]

PR #2667, opened 2017-06-21
Add autofixing caveats explanation

hudochenkov:

https://github.com/stylelint/stylelint/issues/2643#issuecomment-309893126

Maybe we should even tweet about this, because it's pretty serious caveat.

### DIFF
--- a/docs/user-guide/cli.md
+++ b/docs/user-guide/cli.md
@@ -80,6 +80,10 @@ Linting all `.css` files in the `foo` directory. And fixing source files if viol
 stylelint "foo/*.css" --fix
 ```
 
+**Note:** It's an _experimental_ feature. It ignores special comments for disabling stylelint within style sheet (e. g. `/* stylelint-disable /*`). It means autofixing will apply regardless of these comments.
+
+If you're using both these special comments and autofixing, please run stylelint twice. On the first run, some violations could be missed, or some violations might be reported incorrectly. It's a temporary solution until stylelint learn to respect these comments.
+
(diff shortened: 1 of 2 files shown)


hudochenkov:
Users can use [this trick](https://github.com/npm/npm/issues/17118) for npm scripts to run stylelint twice :)

```json
{
  "name": "project",
  "scripts": {
    "lintcss": "npm run postlintcss",
    "postlintcss": "stylelint . --fix"
  }
}
```

docs/user-guide/cli.md
```diff
@@ -80,6 +80,10 @@ Linting all `.css` files in the `foo` directory. And fixing source files if viol
 stylelint "foo/*.css" --fix
 ```
 
+**Note:** It's an _experimental_ feature. It ignores special comments for disabling stylelint within style sheet (e. g. `/* stylelint-disable /*`). It means autofixing will apply regardless of these comments.
+
+If you're using both these special comments and autofixing, please run stylelint twice. On the first run, some violations could be missed, or some violations might be reported incorrectly. It's a temporary solution until stylelint learn to respect these comments.
```
jeddy3: "It's an _experimental_ feature. It currently does not respect special comments for disabling stylelint within sources (e. g. `/* stylelint-disable /*`). Autofixing will be applied regardless of these comments."

"If you're using both these special comments and autofixing, please run stylelint twice as a temporary solution. On the first run, some violations could be missed, or some violations might be reported incorrectly."

PR #2679, opened 2017-06-26
Deprecate replaced rules

### LINKED ISSUE #2513: Deprecate `function-url-data-uris` in favor of `function-url-scheme-blacklist`

### LINKED ISSUE #2528: Deprecate selector-no-* rules in favour of selector-max-*

jeddy3:
Closes #2528 
Closes #2513

Tests should pass once the remaining two `max-` rules are merged into master and this is rebased.


jeddy3:
@hudochenkov This is rebased now and ready for review.

jeddy3:
Changlog:

> -   Deprecated: 6 rules, each has been replaced by a more configurable alternative ([#2679](https://github.com/stylelint/stylelint/pull/2679)).
    -  `function-url-data-uris` rule. Use either `function-url-scheme-blacklist` or `function-url-scheme-whitelist`.
    -  `selector-no-attribute` rule. Use `selector-max-attribute` with `0` as its primary option.
    -  `selector-no-combinator` rule. Use `selector-max-combinators` with `0` as its primary option.
    -  `selector-no-id` rule. Use `selector-max-id` with `0` as its primary option.
    -  `selector-no-type` rule. Use `selector-max-type` with `0` as its primary option.
    -  `selector-no-universal` rule. Use `selector-max-universal` with `0` as its primary option.

PR #2680, opened 2017-06-26
Don't use deprecated rule in example

nikolajevp:

"None, as it's a documentation fix."

"No, it's self explanatory."


jeddy3:
Thanks!

PR #2682, opened 2017-06-26
Improve search keywords for turning off a rule

### LINKED ISSUE #2669: Add the word "ignore" to "disable rules" docs
In stylelint to turn off a rule we use various "disable" comments to disable rules.

It was pointed out to me that stylelint was unable to "ignore" certain rules as when searching the docs on stylelint.io they found no mention of "ignore" in this context.

Once I pointed out they should be searching for "disable" they were up and running 😄 

I'm thinking we should sneak the word "ignore" into some of our "disable" docs verbiage so that if someone does search for "ignore" they will get some search results for the "disable" rule docs.

A couple of suggested locations:

- https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#turning-rules-off-from-within-your-css
  - _"Rules can be temporarily turned off by using special comments in your CSS. For example, you can either turn all the rules off:"_
  - ...becomes:
  - _"Rules can be ignored by temporarily turning them off by using special comments in your CSS. For example, you can either turn all the rules off:"_

- https://github.com/stylelint/stylelint/blob/master/docs/user-guide/faq.md#how-do-i-disable-a-rule
  - _"How do I disable a rule?"_
  - ...becomes:
  - _"How do I ignore or disable a rule?"_

jeddy3:
Closes #2669

We consistently use the term “turn off”. I suggest we keep it that way in the main docs for consistency.

However, this PR adds “ignore” to the FAQ heading, so that it will appear in the search results.

As we position the FAQ as the first port of call for help, I think it makes sense to make searching improvements only there. It then acts as the jumping off point to the consistent docs.

### DIFF
--- a/docs/user-guide/faq.md
+++ b/docs/user-guide/faq.md
@@ -1,8 +1,8 @@
 # FAQ
 
-## How do I disable a rule?
+## How do I turn off, disable or ignore a rule?
 
-You can disable a rule by setting its config value to `null`.
+You can turn off a rule by setting its config value to `null`.
 
 For example, to use `stylelint-config-standard` without the `at-rule-empty-line-before` rule:
 
@@ -15,7 +15,7 @@ For example, to use `stylelint-config-standard` without the `at-rule-empty-line-
 }
 ```
 
-You can also disable a rule for specific sections of your CSS. Refer to the rules section of the [configuration guide](configuration.md#rules) for more information.
+You can also turn off a rule for specific sections of your CSS. Refer to the rules section of the [configuration guide](configuration.md#rules) for more information.
 
 ## How do I lint from the command line?
 


ntwb:
@jeddy3 Is there any process required to trigger Algolia to update the search results or bust the cache? 

(I'm still seeing the old message rather than the update in this PR ~5 hours after it was merged)

hudochenkov:
> (I'm still seeing the old message rather than the update in this PR ~5 hours after it was merged)

@ntwb stylelint.io was updated only an hour ago https://github.com/stylelint/stylelint.io/commit/334be1e7f53ab08380a88545a2c73465e0052116 Maybe Algolia needs little more than that.

ntwb:
@hudochenkov maybe, but that's why I asked if something other than time is required

jeddy3:
> @jeddy3 Is there any process required to trigger Algolia to update the search results or bust the cache?

I've never looked into it I'm afraid. It'd be good to know, though :)

ntwb:
That probably answers the question, if no one has had to update it previously, then 🤖 🤖 🤖 must

PR #2689, opened 2017-06-28
Fix PostCSS@6 closing-brace test

jeddy3:


lib/rules/block-closing-brace-newline-after/__tests__/index.js
```diff
@@ -83,10 +86,11 @@ testRule(rule, {
     line: 1,
     column: 6,
   }, {
-    code: ":root {\n --x { color: pink; } ;\n --y { color: red; };\n }",
+    only: true,
```
hudochenkov: Oops :)

jeddy3: Indeed :)

jeddy3:
`postcss@6` adds support for `@apply`. As such, a semicolon trailing an `@apply` block (_even if preceded by a space_) should be ignored as it's considered part of the `@apply`.

jeddy3:
In hindsight splitting into separate PRs isn't going to work as the CI is going to fail on the other tests. The suite fails is down to `4`, so the tests specific to the issue pass. Shall we merge?

PR #2690, opened 2017-06-28
Fix postcssPlugin tests for postcss@6

jeddy3:
In 6.0 the `plugin.process` shortcut splits the process and plugin options.


jeddy3:
Going to merge into `v8-postcss-six`. It can be reviewed there.

PR #2702, opened 2017-07-01
Update postcss-less to 1.0

jwilsson:

Superseeds #2637.

This is still a WIP, two issues in `no-extra-semicolons` remains.

The big thing however, which isn't reflected by the tests are stringifying of the new `Import` node type in a couple of rules (`string-quotes` for example) and the use of `node.positionBy()` in `utils/report`. It would fail with the same `TypeError: this[node.type] is not a function` errors since the `Root` node there wasn't from `postcss-less` but from `postcss`. I found those issues while running `stylelint` via the CLI. However, I have a fix for that pending in https://github.com/shellscape/postcss-less/pull/82. So as soon as that is merged and a new version is tagged there we should be good to go (after I've fixed `no-extra-semicolons` too of course).

The current changes I've made are just minor ones, passing the current syntax's stringifier to `toResult` when checking the file's newlines and updating `isStandardSyntaxRule` to use new property names for Less mixins/extends.

### DIFF
--- a/lib/rules/no-extra-semicolons/__tests__/index.js
+++ b/lib/rules/no-extra-semicolons/__tests__/index.js
@@ -555,6 +555,6 @@ testRule(rule, {
     code: "a { .mixin();\ncolor: red;; }",
     message: messages.rejected,
     line: 2,
-    column: 11,
+    column: 12,
   }],
 })
--- a/lib/rules/no-extra-semicolons/__tests__/index.js
+++ b/lib/rules/no-extra-semicolons/__tests__/index.js
@@ -555,6 +555,6 @@ testRule(rule, {
     code: "a { .mixin();\ncolor: red;; }",
     message: messages.rejected,
     line: 2,
-    column: 10,
+    column: 11,
   }],
 })
--- a/lib/rules/no-extra-semicolons/index.js
+++ b/lib/rules/no-extra-semicolons/index.js
@@ -53,6 +53,10 @@ const rule = function (actual) {
     }
 
     root.walk(node => {
+      if (node.mixin) {
+        return
+      }
+
       let rawBeforeNode = node.raws.before
 
       if (rawBeforeNode && rawBeforeNode.trim().length !== 0) {
@@ -87,6 +91,15 @@ const rule = function (actual) {
       const rawAfterNode = node.raws.after
 
       if (rawAfterNode && rawAfterNode.trim().length !== 0) {
+        /**
+         * If the last child is a Less mixin followed by more than one semicolon,
+         * node.raws.after will be populated with that semicolon.
+         * Since we ignore Less mixins, exit here
+         */
+        if (node.last && node.last.mixin) {
+          return
+        }
+
         styleSearch({ source: rawAfterNode, target: ";" }, (match) => {
           const index = getOffsetByNode(node) + node.toString().length - 1 - rawAfterNode.length + match.startIndex
 
--- a/lib/rules/no-extra-semicolons/index.js
+++ b/lib/rules/no-extra-semicolons/index.js
@@ -1,6 +1,7 @@
 "use strict"
 
 const isCustomPropertySet = require("../../utils/isCustomPropertySet")
+const isStandardSyntaxRule = require("../../utils/isStandardSyntaxRule")
 const report = require("../../utils/report")
 const ruleMessages = require("../../utils/ruleMessages")
 const validateOptions = require("../../utils/validateOptions")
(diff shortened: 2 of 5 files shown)


jeddy3:
@jwilsson It's fantastic to see this WIP! I'll label as `review needed` once the `package.json` is updated with that new version of `postcss-less`.

lib/rules/no-extra-semicolons/index.js
```diff
@@ -53,6 +53,10 @@ const rule = function (actual) {
     }
 
     root.walk(node => {
+      if (node.mixin) {
```
jwilsson: I'm not 100% satisfied with this current solution here and below, but we can't use `isStandardSyntaxRule` here since it checks for `--custom-property-set: {}`, thus failing other tests which uses that.

While working on this, though, I found myself thinking of a `isLessRule` util. Would that be of interest or is it too specific since it's a non-standard syntax thing? I was thinking that the `isLessRule` util would consist of the Less-specific checks from `isStandardSyntaxRule` and that `isStandardSyntaxRule` could make use of `isLessRule` instead. 

Using a `isLessRule` here in `no-extra-semicolons` would also catch Less `&:extend(.foo)` rules which aren't currently ignored (`&:extend(.foo)` can of course be caught and ignored without it too, but a util would be a better fit IMO).

lib/rules/no-extra-semicolons/__tests__/index.js
```diff
@@ -555,6 +555,6 @@ testRule(rule, {
     code: "a { .mixin();\ncolor: red;; }",
     message: messages.rejected,
     line: 2,
-    column: 10,
+    column: 11,
```
jwilsson: I actually believe that the old column here was incorrect, `11` makes more sense when checking it in my editor for example (the editor is reporting `11` as the first semicolon after `red`. 

I'm happy to be proved wrong though!

jwilsson:
All tests should pass as of the latest commit, just waiting for a new release of `postcss-less`.

lib/rules/no-extra-semicolons/__tests__/index.js
```diff
@@ -555,6 +555,6 @@ testRule(rule, {
     code: "a { .mixin();\ncolor: red;; }",
     message: messages.rejected,
     line: 2,
-    column: 10,
+    column: 12,
```
jwilsson: This should be even more correct, `12` corresponds to the last of the two semicolons.

jwilsson:
This should be ready for review. The new `postcss-less` version is out and in `package.json`.

lib/rules/no-extra-semicolons/index.js
jeddy3: > Would that be of interest or is it too specific since it's a non-standard syntax thing?

Too specific I'm afraid. Rules make use of the `isStandardSyntax*` utils to make sure only _whitelisted_ constructs are dealt with within the rule. There are too many non-standard syntaxes to add special cases to individual rules, even if it's to explicitly ignore a piece of non-standard syntax. There are a lot of rules to maintain and having this consistent behaviour has benefited us thus far.

As such, I suggest we structure that conditional in terms of standard constuct and standard syntax utils i.e.:

```js
     if (
        node.type === "rule"
        && !isCustomPropertySet(node)
        && !isStandardSyntaxRule(node)
      ) {
        return
      }
```

lib/rules/no-extra-semicolons/index.js
```diff
@@ -87,6 +91,15 @@ const rule = function (actual) {
       const rawAfterNode = node.raws.after
 
       if (rawAfterNode && rawAfterNode.trim().length !== 0) {
+        /**
```
jeddy3: As above, let's check if the last node is standard rule or a custom property set:

```js
        if (
          node.last
          && node.last.type === "rule"
          && !isCustomPropertySet(node.last)
          && !isStandardSyntaxRule(node.last)
        ) {
          return
        }
```

lib/rules/no-extra-semicolons/index.js
jwilsson: Gotcha! That's what I figured, but I thought I'd ask :)

jwilsson:
> Would you like me to add you to the stylelint org? That way any future Less-related updates you might do can happen on a branch here.

Sure, happy to help out!

ntwb:
Welcome @jwilsson 😄

jeddy3:
Tests are now green. LGTM.

PR #2704, opened 2017-07-01
Add VISION.md

### LINKED ISSUE #2418: Add option to color-named rule to prevent bad color names

New option

`color-named`

```css
.foo {
  color: greem;
}
```

I am proposing this syntax:
```json
{
  "rules": {
    "color-named": "no-unknown-names"
  }
}
```
Which would in turn flag the CSS because `greem` is an unknown name.

### LINKED ISSUE #2532: Prettier for CSS
[Prettier](https://github.com/prettier/prettier) become really popular in JS. It automatically formats source code by opinionated code style.

And we have same thing in Go — one code style for everyone with autofix.

Should we create same tool for CSS? On top of PostCSS or Stylelint?

jeddy3:
Ref: https://github.com/stylelint/stylelint/issues/2259#issuecomment-279272009

I’ve been mulling this over for a while now. More so since prettier introduced support for CSS and CSS-like languages. This PR is part of a [wider effort](https://twitter.com/slicknet/status/877194938508656641) to help developers understand the benefits of each tool.   

The `VISION.md` document touches on, what I believe to be, the recurring things from the issue tracker. I’ve tried to build on top of the original goals of the project (completeness, extensibility, consistency etc), while taking into account the changes to those goals and the new goals that have emerged over the last two years (performance, support for non-standard syntax etc). I’ve also tried to address some of the confusion around what are the differences/similarities between stylelint and pretty printers.

I’ve updated the list of rules so that they are first grouped by the three categories listed in the `VISION.md`:

-   Possible errors
-   Limit language features
-   Stylistic issues

I believe doing so:

1. Highlights the unique linting strengths of stylelint i.e. the 26 _error catching_ rules and the 48 _limiting_ rules.
2. Makes it easier for users who wish to use stylelint in conjunction with a pretty printer like prettier.

Additionally, I’ve created [`stylelint-config-recommended`](https://github.com/stylelint/stylelint-config-recommended), which turns on just the _error catching_ rules. Named for parity with eslint’s “recommended” config, which does a similar thing. Users of prettier would use this config (adding any additional _limiting_ rules they need, while avoiding the _stylistic_ rules).

I anticipate [`stylelint-config-standard`](stylelint-config-standard) will extend `stylelint-config-recommended`, and apply its current set of _stylistic_ rules.

The final step is to update the stylelint documentation to reference both these shared configs, and to explain what their use cases are.

Closes #2532
Closes #2418

### DIFF
--- a/VISION.md
+++ b/VISION.md
@@ -82,7 +82,7 @@ Provide multiple points of extensions, including:
 Provide a robust tool with a [comprehensive test suite](docs/developer-guide/rules.md#write-tests), including:
 
 -   High coverage, currently over 95%.
--   A wide range of fixtures for rules, currently over 40000.
+-   A wide range of fixtures for rules, currently over 10000.
 
 ## Consistent
 
@@ -95,4 +95,4 @@ Provide consistency throughout, including:
 
 Provide a fast tool, and the means to test and improve performance, including:
 
--   [Benchmarking](docs/developer-guide/rules.md#improving-the-performance-of-a-new-or-an-existing-rule)of an individual rule's performance.
+-   [Benchmarking](docs/developer-guide/rules.md#improving-the-performance-of-a-new-or-an-existing-rule) of an individual rule's performance.
--- a/VISION.md
+++ b/VISION.md
@@ -54,6 +54,19 @@ There are two approaches to enforcing stylistic conventions:
 
 The former is handled by pretty printers, like [prettier](https://github.com/prettier/prettier), whereas the latter is catered for by the built-in stylistic rules.
 
+Additionally, the built-in stylistic rules and plugins are configurable to support this diverse range of stylistic conventions. This is in contrast to pretty printers, which tend to be opinionated. The ordering of properties within declaration blocks is an example of a divisive topic, where there is no one or two dominant conventions. The [`stylelint-order`](https://www.npmjs.com/package/stylelint-order) plugin adheres to stylelint's philosophies, and can be used to lint and fix a diverse range of ordering conventions.
+
+Another example is the use of single line rules for sets of _related_ rules e.g.
+
+```css
+/* Single line related classes */
+.class-1 { top: 0; bottom: 0; }
+.class-2 { top: 5px; right: 0; }
+.class-3 { top: 8px; left: 0; }
+```
+
+The built-in stylistic rules can be configured to allow both multi-line and single-line rules. The choice of when to use each belongs to the user, who can determine which rules are related.
+
 ## Extensible
 
 Provide multiple points of extensions, including:
--- a/VISION.md
+++ b/VISION.md
@@ -1,0 +1,85 @@
+# Vision
+
+A mighty, modern linter and fixer for CSS and CSS-like languages.
+
+One that is:
+
+-   Complete - with coverage of all standard CSS syntax.
+-   Extensible - with multiple points of extension.
+-   Configurable - with no defaults and plenty of options to tailor the linter.
+-   Robust - with comprehensive test coverage and a wide range of fixtures.
+-   Consistent - with conventions for behaviour, naming and documentation.
+-   Performant - with tools to test and improve performance.
+
+## Complete
+
+Provide built-in rules for the following three areas:
+
+-   [Possible errors](docs/user-guide/rules.md#possible-errors).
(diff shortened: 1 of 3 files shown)


VISION.md
```diff
@@ -0,0 +1,85 @@
...
+-   Whitespace
+-   Case
+-   Quotes
+
+There are two approaches to enforcing stylistic conventions:
+
+-   A machine algorithmically pretty prints the code (usually based on a maximum line length).
+-   A human initially formats the code, and a machine fixes-up/warns-about any mistakes.
+
+The former is handled by pretty printers, like [prettier](https://github.com/prettier/prettier), whereas the latter is catered for by the built-in stylistic rules.
```
davidtheclark: I think this gives a little too much credit to Prettier. Prettier's lack of configurability is the reason that stylelint is venturing into the `--fix` arena, right? To provide a more comprehensively configurable autofixer? I think it might be worth highlighting the difference between vision for these two types of autofixing. In fact, since I'm sure readers of this document will be very curious about this, I wonder if autofixing warrants a section of its own, to describe the goals particular to that (esp. differences from Prettier)?

davidtheclark:
Fantastic work, @jeddy3!

VISION.md
jeddy3: @davidtheclark Thanks for the feedback!

> Prettier's lack of configurability is the reason that stylelint is venturing into the `--fix` arena, right? 

Yep, that is one of the reasons. The other is fixing things that prettier is unlikely to. I've tried (my brain has somewhat gone to mush) to address both in the changes I've made :)

ntwb:
Love it @jeddy3 💯 

Myself, I've not yet grok'd all the _prettier_ things and I've a question:

Can use both Stylistic rules and Prettier together?

For example, if I add the `stylelint-ordering` plugin and can I also use Prettier and will Prettier then print _all the CSS_ in the order I defined in my config?

jeddy3:
> Can use both Stylistic rules and Prettier together?

Yes, they're just be tools in a chain. I anticipate users will run stylelint first (with the `stylelint-order` plugin) to shift around the order of the things (delcarations, rules etc) within blocks. Prettier would run second to format the whitespace and other things.

VISION.md
```diff
@@ -0,0 +1,98 @@
...
+-   [Extendable configs](docs/user-guide/configuration.md#extends) - extend and share configurations.
+-   [Formatters](docs/developer-guide/formatters.md) - format stylelint result objects.
+-   [Custom syntax](docs/user-guide/node-api.md#customsyntax) - use any PostCSS-compatible syntax module.
+
+## Robust
+
+Provide a robust tool with a [comprehensive test suite](docs/developer-guide/rules.md#write-tests), including:
+
+-   High coverage, currently over 95%.
+-   A wide range of fixtures for rules, currently over 40000.
```
hudochenkov: Where this number came from? Jest is reporting only 12 805 tests on `v8` branch.

hudochenkov:
P. S. I won't have much time for stylelint work this and the next week, because I just moved to the Netherlands and have lots of things to do beside open-source :)

VISION.md
jeddy3: Oops, I picked a number out of the air as a placeholder until I checked the actual figure. Which I forgot to do :) 

Will fix.

VISION.md
```diff
@@ -0,0 +1,98 @@
...
+Provide consistency throughout, including:
+
+-   Consistent [names](docs/developer-guide/rules.md#naming-a-rule), [options](docs/developer-guide/rules.md#determining-options), [violation messages](docs/developer-guide/rules.md#determine-violation-messages), [documentation](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#write-the-readme) and [treatment](docs/developer-guide/rules.md#write-the-rule) of non-standard syntax within/of rules.
+-   Consistent [requirements for inclusion](docs/developer-guide/rules.md#criteria-for-inclusion).
+
+## Performant
+
+Provide a fast tool, and the means to test and improve performance, including:
+
+-   [Benchmarking](docs/developer-guide/rules.md#improving-the-performance-of-a-new-or-an-existing-rule)of an individual rule's performance.
```
jeddy3: Will fix.

PR #2706, opened 2017-07-01
Use prettier

### LINKED ISSUE #2518: Use Prettier
I've been using Prettier on several other projects, and loving it.

What do @stylelint/core think about using Prettier here, and removing the stylistic rules from our ESLint config?

If we want to do this, is there anybody who is interested in learning about Prettier and leading the way? (If not, I can do it.)

jeddy3:
Closes #2518

To be merged once `v8` is merged into `master`.

Additionally, prettier will need to be run once on the code base as part of this PR.

I’ve a pending PR in `eslint-config-stylelint` to prepare eslint to work along side prettier - https://github.com/stylelint/eslint-config-stylelint/pull/31

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,11 +86,14 @@
     "file-exists-promise": "^1.0.2",
     "flow-bin": "^0.46.0",
     "fs-promise": "^2.0.0",
+    "husky": "^0.14.2",
     "jest": "^20.0.1",
+    "lint-staged": "^4.0.0",
     "npm-run-all": "^4.0.0",
     "npmpub": "^3.0.1",
     "postcss-import": "^10.0.0",
     "postcss-safe-parser": "^3.0.0",
+    "prettier": "^1.5.2",
     "remark-cli": "^3.0.0",
     "remark-preset-lint-consistent": "^2.0.0",
     "remark-preset-lint-recommended": "^2.0.0",
@@ -99,6 +102,7 @@
     "strip-ansi": "^4.0.0"
   },
   "scripts": {
+    "precommit": "lint-staged",
     "benchmark-rule": "node scripts/benchmark-rule.js",
     "dry-release": "npmpub  --dry --verbose",
     "flow": "flow",
@@ -110,6 +114,12 @@
     "release": "npmpub",
     "test": "jest --coverage",
     "watch": "jest --watch"
+  },
+  "lint-staged": {
+    "*.js": [
+      "prettier --write",
+      "git add"
+    ]
   },
   "eslintConfig": {
     "extends": [


ntwb:
Just for my own curiosity and the new workflow here is how it flows in my head:

1. Do stylelint stuff, i.e. edit files for example
2. Test stylelint stuff before commit, i.e `npm test`
3. Add changes `git add .` ** See incoming PR Review comments
4. Commit changes `git commit`

Now, as part of `4.` during the npm task `precommit` is ran and any JS files are parsed through Prettier and modified the _prettier_ way and there is nothing else needs to be performed by the person committing the code?

I think after this and related PRs are merged we can document the stylelint workflow somewhere, maybe in RELEASES.md, though I think CONTRIBUTING.md might make more sense

ntwb:
Here's an example of what I'm referring to in the PR review:

I've made some changes to `lib/index.js`, here's the diff of the changes (2 hunks):

```diff
diff --git lib/index.js lib/index.js
index f1674d4d..3399b25d 100644
--- lib/index.js
+++ lib/index.js
@@ -1,6 +1,7 @@
 "use strict"
 
 const report = require("./utils/report")
+const someThing = require("./utils/someThing")
 const ruleMessages = require("./utils/ruleMessages")
 const validateOptions = require("./utils/validateOptions")
 const checkAgainstRule = require("./utils/checkAgainstRule")
@@ -27,5 +28,6 @@ api.formatters = formatters
 api.createPlugin = createPlugin
 api.createRuleTester = createRuleTester
 api.createLinter = createStylelint
+api.createSomethingEsle = createSomethingElse
 
 module.exports = api
```

Workflow: `git status`:

```shell
~/dev/stylelint/stylelint issue-2518-prettier* 
❯ git status
On branch issue-2518-prettier
Your branch is up-to-date with 'origin/issue-2518-prettier'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   lib/index.js

no changes added to commit (use "git add" and/or "git commit -a")
```

Workflow: `git add -p`: (I only _stage_ the first _hunk_ as I'd like to commit the 2nd hunk later in a follow up commit)

```shell
~/dev/stylelint/stylelint issue-2518-prettier* 
❯ git add -p
diff --git a/lib/index.js b/lib/index.js
index f1674d4d..3399b25d 100644
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,6 +1,7 @@
 "use strict"
 
 const report = require("./utils/report")
+const someThing = require("./utils/someThing")
 const ruleMessages = require("./utils/ruleMessages")
 const validateOptions = require("./utils/validateOptions")
 const checkAgainstRule = require("./utils/checkAgainstRule")
Stage this hunk [y,n,q,a,d,/,j,J,g,e,?]? y
@@ -27,5 +28,6 @@ api.formatters = formatters
 api.createPlugin = createPlugin
 api.createRuleTester = createRuleTester
 api.createLinter = createStylelint
+api.createSomethingEsle = createSomethingElse
 
 module.exports = api
Stage this hunk [y,n,q,a,d,/,K,g,e,?]? n
```

Workflow: `git status`: Now 1 hunk is staged, 1 is not:

```shell
~/dev/stylelint/stylelint issue-2518-prettier* 12s 
❯ git status
On branch issue-2518-prettier
Your branch is up-to-date with 'origin/issue-2518-prettier'.
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	modified:   lib/index.js

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   lib/index.js
```

Workflow: `git commit`:
```shell
~/dev/stylelint/stylelint issue-2518-prettier* 40s 
❯ git commit -m "fix: Committing Some Things"
husky > npm run -s precommit (node v6.11.0)

 ✔ Running tasks for *.js
[issue-2518-prettier 44e52154] fix: Committing Some Things
 1 file changed, 33 insertions(+), 31 deletions(-)
 rewrite lib/index.js (85%)
```

The end result is the 2nd hunk of the patch that I did *not* stage is committed anyway which is not expected behaviour

package.json
```diff
@@ -112,6 +116,12 @@
     "test": "jest --coverage",
     "watch": "jest --watch"
   },
+  "lint-staged": {
+    "*.js": [
+      "prettier --write",
+      "git add"
```
ntwb: This breaks Git compatibility when using [`git add -p`](https://git-scm.com/docs/git-add#git-add--p) to stage commits

See https://github.com/stylelint/stylelint/pull/2706#issuecomment-312465826

ntwb: This breaks Git compatibility when using [`git add -p`](https://git-scm.com/docs/git-add#git-add--p) to stage commits

See https://github.com/stylelint/stylelint/pull/2706#issuecomment-312465826

ntwb:
Ha, as a follow up I thought I'd go try and find a solution to this and I hit upon the issue instantly via the module readme: https://www.npmjs.com/package/lint-staged#reformatting-the-code

> _~"Starting from v3.1, lint-staged will stash you remaining changes (not added to the index) and restore them from stash afterwards. This allows you to create partial commits with hunks using git add --patch.~ [This is still not resolved](https://github.com/okonet/lint-staged/issues/62)"_

So there we have it, once `lint-staged` has a solution we can add that in a follow up PR 💥 

https://github.com/okonet/lint-staged/issues/62

jeddy3:
Thanks for digging into the partial commits. 

> So there we have it, once lint-staged has a solution we can add that in a follow up PR

Agreed. I don't think we should postpone merging this in the meantime.

jeddy3:
@ntwb This is now ready. I'll merge on green.

hudochenkov:
I think we should add trailing comma `--trailing-comma es5` for prettier, it will lead to better diffs.

ntwb:
Copying what I wrote in https://github.com/stylelint/stylelint/pull/2742#issuecomment-315488060 here:
> > I think we should add trailing comma `--trailing-comma es5`, it will lead to better diffs.
> >
> hmmm, does that mean at the moment if we create a patch right now that our ESLint rules force us to use trailing commas, then prettier strips them?
>
> If that's the case then I'm 👍  for that change 

What's the best way to test this? I think merging this PR as is with the options as they are now and then a follow up PR with this change and then look at what those changes are in the diff and decide from there?

hudochenkov:
> hmmm, does that mean at the moment if we create a patch right now that our ESLint rules force us to use trailing commas, then prettier strips them?

We don't have [`comma-dangle`](http://eslint.org/docs/rules/comma-dangle) rule enabled in [our ESLint config](https://github.com/stylelint/eslint-config-stylelint/blob/master/eslintrc.json). There won't be any conflicts between Prettier and ESLint.

We could add `--trailing-comma es5` in other PR, because it will require to run Prettier over all our files That PR we could open after this one is merged.

jeddy3:
> We could add `--trailing-comma es5` in other PR, because it will require to run Prettier over all our files That PR we could open after this one is merged.

I'll merge this as-is as it's good-to-go.

We can open up another issue if there's further need to discuss whether we want to add `--trailing-comma es5`. Personally, I'm in favour of sticking with the default setting to keep things simple as I think that trumps slightly better diffs.

package.json
```diff
@@ -111,6 +115,12 @@
     "test": "jest --coverage",
     "watch": "jest --watch"
   },
+  "lint-staged": {
+    "*.js": [
+      "prettier --write",
```
Mouvedia, line 120: @ybiquitous what about [pretty-quick](https://github.com/prettier/pretty-quick) instead?

ybiquitous: @Mouvedia We use `lint-staged`, isn't pretty-quick unnecessary?

Issue #2647, opened 2017-06-17
Case of line should be respected by `ignorePattern` in max-line-length


ntwb:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug where trying to ignore multiple patterns using `ignorePattern` in the `max-line-length` rule

This is a follow up to #2630

> Which rule, if any, is this issue related to?

`max-line-length`

> What CSS is needed to reproduce this issue?

```css
/*
Theme Name: Twenty Eleven
Theme URI: Theme URI: https://wordpress.org/themes/twentyeleven/styles/css/url/path/longer/than/80/chars
Author: the WordPress team
Author URI: https://wordpress.org/
Description: The 2011 theme for WordPress is sophisticated, lightweight, and adaptable.
Version: 2.6
License: GNU General Public License v2 or later
License URI: http://www.gnu.org/licenses/gpl-2.0.html
Tags: blog, one-column, two-columns, left-sidebar, right-sidebar, custom-background.
Text Domain: twentyeleven
*/
```

> What stylelint configuration is needed to reproduce this issue?

```js
   "max-line-length": [ 80, {
      "ignore": "non-comments",
      ignorePattern: ["/(https?:\/\/[0-9,a-z]*.*)|(Description\:.+)|(Tags\:.+)/"],
    } ],
```

> Which version of stylelint are you using?

`7.11.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI via `stylelint test.css --config index.js`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

No warnings to be flagged on lines 3, 6, and 10

> What actually happened (e.g. what warnings or errors you are getting)?

The following warnings were flagged:

```shell
test.css
  3:106  ✖  Expected line length to be no more than 80 characters   max-line-length
  6:87   ✖  Expected line length to be no more than 80 characters   max-line-length
 10:84   ✖  Expected line length to be no more than 80 characters   max-line-length
```

Here's what a test of this I think should look like:
```js
testRule(rule, {
  ruleName,
  config: [ 20, { ignorePattern: "/(multiple\:.+)|(should\:.+)" } ],

  accept: [{
    code: "/*\n multiple lines in multiline comments\n should be able to be ignored\n*/",
  }],

})
```

I've tried various regex `g` and `m` modifiers in `ignorePattern` such as these with no luck:
[...]

PR #2709, opened 2017-07-02
Case sensitive white/blacklists and ignore* options

### LINKED ISSUE #2660: Ensure consistent case sensitivity across *-list rules and ignore* options
Ref: https://github.com/stylelint/stylelint/pull/2626#issuecomment-309861380

jeddy3:
Closes #2660

On further my investigation, I believe my [initial suggestion](https://github.com/stylelint/stylelint/issues/2660#issuecomment-309911294) was wrong. Having whitelist, blacklists and `ignore*: []` secondary options be case insensitive by default is too much magic. It leads to problems like https://github.com/stylelint/stylelint/issues/2647 and inconsistent behaviour (see: `comment-word-blacklist` where a case insensitive would not enable to user to allows `"TODO:` while disallowing `"todo:`). Users should _explicitly_ opt into case insensitivity by using regular expressions and the `i` flag.  This approach works well because we have all the `*-case` rules. It also means the behaviour of whitelist and blacklists and `ignore*: []` options are consistent across all rules, including `max-line-length` and `comment-word-blacklist`.

This PR enables the use of the `i` flag in regular expressions, and updates the code, tests and docs of the `*-whitelist` and `*-blacklist` rules, and any rules that have an `ignore*: []` secondary option.

This also fixes the current incorrect behaviour where users are forced to use lowercase values in black/whitelists e.g. a user no longer needs to use `function-whitelist: ["translatex"]`, they can now use the more intuitive `function-whitelist: ["translateX"]`.


ntwb:
Thanks @jeddy3, quite a bit involved here, thanks 👍 

I've just tested this branch with various `/i` tests and switching case for #2647, works a treat.

_`npm run watch` with Jest tests is such a great workflow 😊_

jeddy3:
> npm run watch with Jest tests is such a great workflow 😊

Yep! It was a big time saver whilst working on this PR. After tweaking the underlying utility I ran all tests and jotted down the failing rule names. I was then able to filter by each rule as I worked through them patching up the tests.

ntwb:
Still a thumbs up from me

jeddy3:
`v8`:

> -   Changed: the primary options of the `*-whitelist` & `*-blacklist` rules and the `ignore* []` secondary options are now case sensitive. Use regular expressions with the `i` flag for case insensitivity ([#2709](https://github.com/stylelint/stylelint/issues/2709)).

Issue #2545, opened 2017-05-05
Exit with non-zero code on Cannot parse selector syntax error


FezVrasta:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

When stylelint raises a "Cannot parse selector" warning, I would like it to return with a non zero code to make the linting fail.

> What CSS is needed to reproduce this issue?

```css
data-something="true"] .test
  color: red
```

> What stylelint configuration is needed to reproduce this issue?

No configuration needed.

> Which version of stylelint are you using?

`7.10.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylelint "**/*.sss"`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

I don't think so.

> What did you expect to happen?

An error should be thrown.

> What actually happened (e.g. what warnings or errors you are getting)?

No error, just a log in the output.

jeddy3:
@FezVrasta Thank you for the report and for using the template.

I can reproduce this with standard CSS:

```css
data-something="true"] .test {
  color: red;
}
```

Gives:

```shell
jeddy3:stylelint-test jeddy3$ ./node_modules/.bin/stylelint sources/css/exit.css

sources/css/exit.css
 1:1     Cannot parse selector

jeddy3:stylelint-test jeddy3$ echo $?
0
```

Whereas a PostCSS syntax error (rather than the `postcss-selector-parser` error above):

```css
data-something="true"] .test {
  color: red;
```

Gives:

```
jeddy3:stylelint-test jeddy3$ ./node_modules/.bin/stylelint sources/css/exit.css

sources/css/exit.css
 1:1  ✖  Unclosed block   CssSyntaxError

jeddy3:stylelint-test jeddy3$ echo $?
2
```

Should the behaviour for both syntax errors be consistent i.e. throw an error?

FezVrasta:
Yes it would be Ideal. On SugarSS PostCSS doesn't throw any error

jeddy3:
@stylelint/core Anyone got any ideas about how this could be implemented?

I believe the `Cannot parse selector` warning in produced in [`parseSelector.js`](https://github.com/stylelint/stylelint/blob/master/lib/utils/parseSelector.js) and [`selector-max-specificity`](https://github.com/stylelint/stylelint/blob/master/lib/rules/selector-max-specificity/index.js#L67).

jeddy3:
[...]

PR #2713, opened 2017-07-03
Exit with non-zero code on postcss warnings

evilebottnawi:

https://github.com/stylelint/stylelint/issues/2545

No, it's self explanatory.

I think we should be default exist with non zero code on all postcss warnings. But we can use regexp to filter only `Cannot parse selector`.

### DIFF
--- a/decls/stylelint.js
+++ b/decls/stylelint.js
@@ -75,6 +75,7 @@ export type stylelint$result = {
   invalidOptionWarnings: Array<{
     text: string,
   }>,
+  parseErrors: Array<stylelint$warning>,
   errored?: boolean,
   warnings: Array<stylelint$warning>,
   ignored?: boolean,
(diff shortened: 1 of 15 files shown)


jeddy3:
> I think we should be default exist with non zero code on all postcss warnings.

I think this would defeat the purpose of severity levels in the config (e.g. `warning`, `error` etc).

> But we can use regexp to filter only Cannot parse selector.

That might be the best solution here.

jeddy3:
Actually, perhaps we can create a new warning _type_, called `parseError` to go alongside side our existing `deprecation` and `optionsValidation` types. We can then  exit with non zero code on any of those warnings.

evilebottnawi:
@jeddy3 done, should we improve `stringFormatter` for this?

jeddy3:
`v8`:

> -   Changed: stylelint now exits with non-zero code on parse errors ([#2713](https://github.com/stylelint/stylelint/issues/2713)).

PR #2717, opened 2017-07-04
Correct extends behaviour in document (fix #2716)

### LINKED ISSUE #2716: Correctly document extends behaviour
In [extends section of configuration documentation](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#extends) there is the following statement:
> You can extend an array of existing configurations, with each item in the array taking precedence over the following (so the first item overrides everything else, the second item overrides everything but the first, the last gets overridden by everything else, etc.).

After that there is an example:
> Or starting with stylelint-config-standard, then layering myExtendableConfig on top of that, and then overriding the indentation rule:
```json
{
  "extends": [
    "stylelint-config-standard",
    "./myExtendableConfig"
  ],
  "rules": {
    "indentation": "tab"
  }
}
```
According to the phrase it seems that configs inside `extends` section should be swapped:
```json
  "extends": [
    "./myExtendableConfig",
    "stylelint-config-standard"
  ],
```
Or did I misunderstand the phrase and the example?

gamtiq:

#2716

### DIFF
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -175,7 +175,7 @@ Writing a [custom formatter](../developer-guide/formatters.md) gives you maximum
 
 Your configuration can *extend* an existing configuration (whether your own or a third-party config). When one configuration extends another, it starts with the other's properties then adds to and overrides what's there.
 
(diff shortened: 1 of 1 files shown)


jeddy3:
Thanks!

PR #2730, opened 2017-07-11
Docs: keyframe-declaration-no-important

alexilyaev:
Added link to clarify that using !important within keyframes declarations is completely ignored in some browsers.


jeddy3:
@alexilyaev Thanks! Adding reference links to rule docs is always a good thing.

Issue #2369, opened 2017-02-20
Run --report-* flags before linting, rather than instead of it


evilebottnawi:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Be good can use linting and report needless disables. It is allow reduce unnecessary stylelint disable rules and check on linting error in one command. If you use two scripts (command), then we run twice globby and this increases the load on the file system :disappointed: 

> Which rule, if any, is this issue related to?

No matter

> What CSS is needed to reproduce this issue?

No matter
> What stylelint configuration is needed to reproduce this issue?

e.g.

No matter

> Which version of stylelint are you using?

`7.9.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No matter

> What did you expect to happen?

Warning and messages about unnecessary stylelint disable comments

> What actually happened (e.g. what warnings or errors you are getting)?

Only two scripts (commands) allow this feature.

jeddy3:
Is the idea here that using the `--report-needless-disables` would run that check first and then run the linter afterwards, rather than _instead of the linter_? 

And I assume the process would exit early if `--report-needless-disables` uncovered any issues?

Can anyone foresee any disadvantages to changing the behaviour?

davidtheclark:
Seems fine to me.

jeddy3:
@evilebottnawi Is this issue something you'll be able to look into for `8.0.0`?

evilebottnawi:
@jeddy3 try to find solution in near future

hudochenkov:
Done in https://github.com/stylelint/stylelint/pull/4151

PR #2731, opened 2017-07-11
Enhancement: run `--report-needless-disables` before linting, rather …

evilebottnawi:
…than instead of it.

#2369

No, it's self explanatory.

One note: should be `errored` in true if we contain needless disables, `parseErrors` set `errored` to `true`.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -257,8 +257,6 @@ Promise.resolve().then(() => {
       process.stdout.write(needlessDisablesStringFormatter(linted.needlessDisables))
       process.exitCode = 2
     }
-
-    return
   }
 
   if (!linted.output) {


jeddy3:
I originally thought that resolving this issue would be as simple as removing this `return`. However, in my testing this seemed to change what violations would be detected by the linter. I wasn't able to dig deeper at the time, but I thought it might have something to do with the disable ranges.

Either way, have you been able to locally test this change to check that the behaviour is correct?

evilebottnawi:
@jeddy3 after move ignore logic here https://github.com/stylelint/stylelint/blob/v8/lib/standalone.js#L133, seems there is no problems, tested in several local project.

jeddy3:
When testing this branch locally...

Given:

```json
{
  "rules": {
    "color-hex-case": "lower",
    "color-hex-length": "short"
  }
}
```

And the following CSS (where `#FFFFFF` violates both rules: `hex-case` and `hex-length`):

```css
/* stylelint-disable color-hex-case */ 
a { color: #FFFFFF; }
/* stylelint-enable color-hex-case */
```

When run with:

```shell
./node_modules/.bin/stylelint "sources/css/disable.css"
```

I get:

```console
sources/css/disable.css
 2:12  ✖  Expected "#FFFFFF" to be "#FFF"   color-hex-length
```

This is correct because the `color-hex-case` rule is disabled by the `stylelint-disable` comment, and only the `color-hex-length` violation is reported.

However, when I use the `--report-needless-disables` flag I get:

```console
jeddy3:stylelint-test jeddy3$ ./node_modules/.bin/stylelint "sources/css/disable.css" --report-needless-disables

sources/css/disable.css
 2:12  ✖  Expected "#FFFFFF" to be "#ffffff"   color-hex-case
 2:12  ✖  Expected "#FFFFFF" to be "#FFF"      color-hex-length
```

It seems the use of `--report-needless-disables` turns off the disable ranges. Is this the intended behaviour? It doesn't seem right.

evilebottnawi:
@jeddy3 yep, i will try to found solution tomorrow, don't have time right now :disappointed:

jeddy3:
Going to nuke this PR so we can delete the `v8` branch.

@evilebottnawi Feel free to create a new PR off `master` if you find time to work on it. I now believe this is a non-breaking change as the linter will only run if the flag doesn't return any violations. If the flag does return some violations it'll exit with a non-zero code, which is the current behaviour. So, this is more of a usability fix than a breaking change, and can therefore happen after `8.0.0` is out.

PR #2732, opened 2017-07-11
Refactor: simplify ignore logic.

evilebottnawi:

Based on https://github.com/kaelzhang/node-ignore#addignorefilepath

No, it's self explanatory.

### DIFF
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -64,7 +64,7 @@ module.exports = function (options/*: stylelint$standaloneOptions */)/*: Promise
   } catch (readError) {
     if (readError.code !== FILE_NOT_FOUND_ERROR_CODE) throw readError
   }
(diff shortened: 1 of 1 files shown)


jeddy3:
@evilebottnawi Can you give a few more details about this change please?

According [to the docs](https://github.com/kaelzhang/node-ignore#addpattern), `add()` expects either a string or an array of strings. Aren't we splitting our _multi-line_ string on newlines to pass in an array of ignore strings to `add()`?

evilebottnawi:
@jeddy3 we can avoid splitting, because `node-ignore` do it https://github.com/kaelzhang/node-ignore/blob/master/ignore.js#L42

jeddy3:
Thanks, that makes a lot more sense :)

PR #2737, opened 2017-07-13
Update eslint and config

jeddy3:
Reordering the requires was autofixable. `no-useless-escape` should of been, but didn’t seem to work. It was worth the manual effort though as the rule caught a couple of typos in tests where `\` was being used instead of `\n`.


lib/rules/max-line-length/__tests__/index.js
```diff
@@ -238,7 +238,7 @@ testRule(rule, {
 
 testRule(rule, {
   ruleName,
-  config: [ 20, { ignorePattern: "/https?:\/\/[0-9,a-z]*.*/" } ],
+  config: [ 20, { ignorePattern: "/https?://[0-9,a-z]*.*/" } ],
```
ntwb: My regex tool throws an error for this: `ERROR: Unescaped forward slash.`

lib/utils/__tests__/hasEmptyLine.test.js
```diff
@@ -12,7 +12,7 @@ it("hasEmptyLine", () => {
   expect(hasEmptyLine("\n \n")).toBeTruthy()
   expect(hasEmptyLine("\r\n \r\n")).toBeTruthy()
   expect(hasEmptyLine("\n \n \n \n")).toBeTruthy()
-  expect(hasEmptyLine("\r \n\ r\n\r\n\r\n")).toBeTruthy()
+  expect(hasEmptyLine("\r \n\n r\n\r\n\r\n")).toBeTruthy()
```
ntwb: Nice that it caught these `\n`'s 👌

scripts/benchmark-rule.js
```diff
@@ -26,6 +27,7 @@ if (
 ) {
   parsedOptions = JSON.parse(ruleOptions)
 }
+/* eslint-disable eqeqeq */
```
ntwb: This should be `/* eslint-enable eqeqeq */`

ntwb:
I picked up that one _regex_ issue, going to take a second look, a closer look at a few more regex's

scripts/benchmark-rule.js
ntwb: p.s. This inline comment doesn't show the `/* eslint-disable eqeqeq */` a couple of lines up at line 21

lib/rules/max-line-length/__tests__/index.js
jeddy3: Keep in mind that this isn't a regex itself, but rather a string of a regex

scripts/benchmark-rule.js
jeddy3: Done

jeddy3:
> I picked up that one regex issue, going to take a second look, a closer look at a few more regex's

Please do! I'm going to be out for the next couple of hours, so feel free to pull down this branch, edit and push back up with any changes.

lib/rules/max-line-length/__tests__/index.js
ntwb: Per the rule docs though this should be a regex though should it not?
https://stylelint.io/user-guide/rules/max-line-length/#ignorepattern-regex

I can't remember which rule or rules have the option to use a _string_ or a _regex_ but it doesn't appear that `max-line-length` is one of the rules that allows a _string_ or _regex_ in the pattern, only a _regex_ based on the rule documentation

jeddy3: > Per the rule docs though this should be a regex though should it not?

I meant that it's a regular expression _within_ a string e.g. `"/regex/"`, rather than a regular expression itself e.g. `/regex/`. To be honest, all this regex handling and escaping, especially when they're as strings within JSON configuration objects, is a bit beyond me!

The tests are passing though, so does that mean we're OK? Or should we continue to dig deeper?

jeddy3: @Arcanemagus I hope you don't mind me pinging you, but do you have any thoughts on this? You seemed knowledgeable about this subject in the other issue: https://github.com/stylelint/stylelint/issues/2673.

Arcanemagus: Looking into how the string is used.

Arcanemagus: Looks like the string has the beginning and ending `/` cut off of it and fed into `new RegExp` [here](https://github.com/stylelint/stylelint/blob/8abafb8fe621ada73552cd9fe436553df5b94b53/lib/utils/matchesStringOrRegExp.js#L48), which means the `/` within the string do _not_ need to be escaped:
image

So this change is correct as far as I can see 😉.

jeddy3: Awesome thanks! Great to have a second pair of eyes looks this one over :)

ntwb: SGTM ✅ Merge away 💥

PR #2675, opened 2017-06-24
chore(package): update flow-bin to version 0.48.0

ntwb:
Replaces #2595

PR #2740, opened 2017-07-14
chore(package): update flow-bin to version 0.50.0

ntwb:
Previously #2687 #2675

Suggestion for fix in https://github.com/stylelint/stylelint/pull/2675#issuecomment-311665840


ntwb:
In a somewhat unrelated manner I just discovered that we should probably update the `flow-typed` Jest v20.x definitions, this was previously updated for Jest v19.x in #2400.

I've taken a calculated guess that this change should be included in this PR, I've not made any other changes at this stage, still grok'ing what needs to be done here.

jeddy3:
@ntwb This is green now. As we support node@4 (without transpiling) we [can't use either](http://node.green/) default parameters or the spread operator. So, this [tweaks the annotation](https://flow.org/en/docs/types/functions/#toc-optional-parameters) to work around that.

Shall I merge?

ntwb:
Awesome, merge away 💥

Issue #2703, opened 2017-07-01
Document that Less at-import will be ignored in at-rule-*


jwilsson:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug which will occur in the future.

> Which rule, if any, is this issue related to?

Most (if not all) `at-rule-*` rules and probably other ones where `walkAtRules()` is used.

> What CSS is needed to reproduce this issue?

```less
@IMPORT "foo.less";
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "at-rule-name-case": ["lower"]
  }
}
```

> Which version of stylelint are you using?

`v8` branch

> How are you running stylelint: CLI, PostCSS plugin, Node API?

`./bin/stylelint.js --config config.json file.less`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes, it's related to Less.

> What did you expect to happen?

`at-rule-name-case` to flag the uppercase `@IMPORT`.

> What actually happened (e.g. what warnings or errors you are getting)?

No issue was flagged because `@import` nodes in `.less`-files aren't checked anymore. This is because they're no longer emitted as `atrule` but `import`.

I'd be happy to submit a PR fixing this across all rules.

jeddy3:
@jwilsson Thanks again for working on this! There might not be anything to do here, other than updating the applicable rule docs, as the built-in rules only contain code for parsing and checking standard CSS. A series of `isSyntaxSyntax*` utils _whitelist_ standard CSS syntax and features before any checks are run. Support for non-standard parsing and/or features is done in plugins (see [stylelint-scss](https://github.com/kristerkari/stylelint-scss) as an example).

jwilsson:
Alright, I can take a stab at the docs when the `postcss-less@1.0` work is done.

PR #2741, opened 2017-07-14
Document rules that ignore @import in Less

jwilsson:

#2703

I think I got all of the affected rules, but an extra set of eyes would be appreciated.

### DIFF
--- a/lib/rules/at-rule-blacklist/README.md
+++ b/lib/rules/at-rule-blacklist/README.md
@@ -8,6 +8,8 @@ Specify a blacklist of disallowed at-rules.
  * At-rules like this */
 ```
 
+This rule ignores `@import` in Less.
+
 ## Options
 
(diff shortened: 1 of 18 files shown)


ntwb:
Is it worth adding some Less syntax tests to confirm `@ignore` is ignored in each of these rules?

jeddy3:
> Is it worth adding some Less syntax tests to confirm @import is ignored in each of these rules?

I think so.

jwilsson:
> Is it worth adding some Less syntax tests to confirm @import is ignored in each of these rules?
>
> I think so.

Do we wanna do that in the same PR or should I just create a new one with the tests?

ntwb:
Add them to this same PR, as doing so would also confirm the existing changes here are working as intended

jeddy3:
LGTM. I'll resolve the conflicts.

PR #2742, opened 2017-07-14
Pretty-print code

jeddy3:
Ref #2518 

Uses prettier’s default settings to make it as easy-as-possible for us all to use, should we wish, prettier within our editors.

I don’t think the actual code style matters. Using prettier means we can just write code (without thinking about formatting), save in editor (which optionally formats it there and then) and we’re good to go.

And the resulting code within the repo (for diffs, look-ups etc) is consistent.


hudochenkov:
Semicolons! Finally :)

I think we should add trailing comma `--trailing-comma es5`, it will lead to better diffs.

ntwb:
> I think we should add trailing comma --trailing-comma es5, it will lead to better diffs.

hmmm, does that mean at the moment if we create a patch right now that our ESLint rules force us to use trailing commas, then prettier strips them?

If that's the case then I'm 👍  for that change

jeddy3:
> our ESLint rules force us to use trailing commas, then prettier strips them

I don't believe our ESLint rules enforce a trailing comma.

PR #2743, opened 2017-07-14
Use latest npm

### LINKED ISSUE #2734: chore: Use npm 5.x for all Node.js Travis CI jobs
Historically we've kept the installed version of npm to be that of what is shipped with the Node.js version, this is partially incorrect and has been pointed out by npm that we should use the latest release of npm, the latest release of npm is `5.2.0`, npm does not maintain any LTS releases. npm will push security releases to previous 4.x and 3.x branches as required but no bug fixes.

More here: http://blog.npmjs.org/post/162451604515/the-npm-clis-long-term-support-lts-policy 

Once the v8 branch is merged I propose we switch to using npm 5.x for all Node.js versions.

I'm making a the @stylelint/core team recommendation that a prerequisite to use the latest npm 5.x release by all team members. (Basically if you are not using the latest npm `5.2.0` you are using out of date software)

I've created two pull requests demonstrating the changes I'll make after v8 is merged:
• https://github.com/stylelint/stylelint-config-standard/pull/84
• https://github.com/stylelint/stylelint-config-recommended/pull/3

jeddy3:
Closes #2734

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,20 @@
-language: node_js
 sudo: false
+
+git:
+  depth: 1
+
+branches:
+  only:
+    - master
+
+language: node_js
+
+# cache node modules
 cache:
   directories:
+    - $HOME/.npm
     - node_modules
+
 matrix:
   include:
     - node_js: '8'
@@ -18,11 +30,13 @@ matrix:
       script: npm run jest -- --runInBand
       env: CI=tests 4
       sudo: required
+
 before_install:
-  - npm prune
-  - npm update
+  - npm install -g npm@latest
+
 after_success:
   - if [ "$CI" = "coverage" ]; then cat ./.coverage/lcov.info | ./node_modules/.bin/coveralls; fi
+
 notifications:
   webhooks:
     - https://webhooks.gitter.im/e/d763493612da45967361
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 node_modules
-package-lock.json
 *.log
 .coverage
 .eslintcache
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,7579 @@
+{
+  "name": "stylelint",
+  "version": "7.13.0",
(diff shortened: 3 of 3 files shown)


.travis.yml
```diff
@@ -1,8 +1,20 @@
-language: node_js
 sudo: false
+
+git:
+  depth: 1
+
+branches:
+  only:
+    - master
```
jwilsson: I _think_ this will stop Greenkeeper from testing new versions of dependencies.

Something like `- /^greenkeeper.*$/` might be needed to fix it.

jeddy3: I [think you're right](https://github.com/greenkeeperio/greenkeeper/issues/222).

hudochenkov: Does this also mean, that we can't test our PRs in Travis anymore? Because they are not `master` or `greenkeeper/*`.

jeddy3: I wondered the same thing too. The easiest way to find out is to fail a test in this branch, which I'll do now.

jeddy3: Result: It fails appropriately. 

I'll merge this (and squash down the commits) once we're green again

hudochenkov: What is the point of this `branch` configuration?

jeddy3: I think it reduces load on the CI by only [running on specific branches](https://docs.travis-ci.com/user/customizing-the-build/#Safelisting-or-blocklisting-branches) e.g. if you create a branch and push it up, the CI won't run. However, it looks like [PRs are treated differently](https://docs.travis-ci.com/user/pull-requests) and when a PR is created the CI runs by merging into master, which explains why the test I just did correctly failed.

Shall I merge?

hudochenkov: If you're confident about this part of configuration, then — yes.

ntwb: Previously each PR would trigger 2 Travis CI jobs, now only the required Travis CI PR job testing the merge into `master` is ran.

Issue #2271, opened 2017-01-19
False positives for parametric mixins with rule block/snippet in indentation


timbru31:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug where stylelint does not treat parameters in case a ruleblock/snippet is passed.

> What CSS is needed to reproduce this issue?

```css
#root {
	.foo(@name, @rules) {
		.@{name} {
			@rules();
		}
	}
}

#root > .foo(test, {
	margin-bottom: 15px;
});
```

> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
	"rules": {
		"indentation": "tab"
	}
}

```

> Which version of stylelint are you using?

7.7.1

> How are you running stylelint: CLI, PostCSS plugin, Node API?

./node_modules/.bin/stylelint test.less

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes, Less mixins.

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

The following warnings were flagged

```shell
test.less
 10:2  ✖  Expected indentation of 0 tabs   indentation
```

jeddy3:
@timbru31 Thanks for the report and for using the issue template.

Supporting non-standard syntax is currently a low priority for the core team. As such, please consider investigating and, if necessary, [contributing](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md) a fix to the [`indentation`](https://github.com/stylelint/stylelint/tree/master/lib/rules/indentation) rule.

webOS101:
I have created a patch for this that fixes all the less mixin false positives in our code base.  Be interested if it also fixes your use cases.

jeddy3:
Done in https://github.com/stylelint/stylelint/issues/2744

timbru31:
I can confirm the fix, thanks! 🎉

PR #2744, opened 2017-07-15
First pass at fix

webOS101:

#2271

This is the most naive fix I could come up with for the less parametric mixin.  It's entirely likely that I've gone in the wrong direction.  If so, please give some hints on where to focus.  I added tests for the failing cases and ensure they all pass.  I did discover another false positive I will open an issue on, but it resisted my early attempts at resolution.

~~I'll also note that I just merged master and things are a hot mess.  linting rules are failing left and right due to new code style but non-updated lint rules.  Tests are failing everywhere.  I did check that none of the indentation tests are failing.~~  Apparently I didn't do an `npm install` after merging master.  All clean!

### DIFF
--- a/lib/rules/indentation/__tests__/functions.js
+++ b/lib/rules/indentation/__tests__/functions.js
@@ -831,6 +831,31 @@ testRule(rule, {
         "  );\r\n" +
         "}",
       description: "Less mixin with multi-line arguments"
+    },
+    {
+      code:
+        ".foo {\r\n" +
+        "  .mixin(\r\n" +
+        "    {\r\n" +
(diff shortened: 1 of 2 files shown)


ntwb:
Thanks for the PR @webOS101 👍 

@jwilsson I've assigned this to you, can you have a look please?

jeddy3:
> -   Fixed: `indentation` false positives for Less parametric mixins with rule block/snippet ([#2744](https://github.com/stylelint/stylelint/pull/2744)).

@webOS101 Thanks for your first-time contribution :)

PR #2752, opened 2017-07-16
Revert jest lint staged

jeddy3:
I’m afraid that experiment didn’t last long. It seemingly broke adding commits to Github Desktop.

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,8 +118,7 @@
   "lint-staged": {
     "*.js": [
       "prettier --write",
-      "git add",
-      "jest --bail --findRelatedTests"
+      "git add"
     ]
   },
   "eslintConfig": {


ntwb:
Cool, I've an alternative method to try than might work, I'll test it with GitHub Desktop, if it does I'll create a fresh PR

hudochenkov:
I works for me in the latest Github Desktop. Successfully commits when there is no error, and show Jest errors if test is failing. But commit time is about a minute or so, because Jest run a lot of tests, even if only one test file was changed.

ntwb:
Right, thanks @hudochenkov, it should only run the Jest tests against _staged_ files, so that doesn't seem correct either, I'll also confirm that behaviour with GitHub Desktop and the alternative method I'll test.

jeddy3:
> But commit time is about a minute or so

Ha, I clearly didn't have the patient to wait that long before closing down the program :)

I think we all agree that it's a bit too long. So merging.

Issue #2677, opened 2017-06-26
Cannot parse selector warning for mixins with 2 or more parameters in selector-max-specificity


PhilippeProulx:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug where using the less & operator with nested classes within a mixin that has more than one parameter causes a parsing error.

> Which rule, if any, is this issue related to?

It seems to be related to `selector-max-specificity` since the error will not show up when this rule is not used in the config file.

> What CSS is needed to reproduce this issue?

```less
.foo {
  .myMixin(@a, @b) {
    &.bar {
      color: pink;
    }
  }
}
```
Note: The error does not occur when one parameter or less is used in the mixin.

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "extends": "stylelint-config-standard",
  "rules": {
    "selector-max-specificity": "0,4,0"
  }
}
```

> Which version of stylelint are you using?

`7.11.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylelint foo.less`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes, it's related to LESS nested class with & operator in a mixin.

> What did you expect to happen?

no parsing error

> What actually happened (e.g. what warnings or errors you are getting)?

The following error was outputted:

```shell
3:5     Cannot parse selector
```

hudochenkov:
Please, use issue template, and provide your config.

PhilippeProulx:
Thanks for the quick reply, I updated my issue to comply with the template and added more information.

jeddy3:
Can we add a heuristic to [`isStandardSyntaxSelector`](https://github.com/stylelint/stylelint/blob/master/lib/utils/isStandardSyntaxSelector.js) that ignores something like `.myMixin(@a, @b).bar`?

@PhilippeProulx Non-standard syntax support isn't a priority for the core team. Please consider contributing a fix if you'd like to see this issue resolved.

jwilsson:
I'll take a look at this one.

jeddy3:
Done in https://github.com/stylelint/stylelint/issues/2755

PR #2755, opened 2017-07-17
isStandardSyntaxSelector should not check rules resolved as nested Less mixins

jwilsson:

#2677

While the original issue only covered `.myMixin(@a, @b).bar` I discovered that any nested selector would fail with a `Cannot parse selector` error. 

For example:

```less
.myMixin() {
  & > .bar {
  }
}
```

would fail as well. 

I added a check to `isStandardSyntaxSelector` to check anything that matches `.foo()` since I can't find an instance where that would be valid CSS.

### DIFF
--- a/lib/utils/__tests__/isStandardSyntaxSelector.test.js
+++ b/lib/utils/__tests__/isStandardSyntaxSelector.test.js
@@ -57,7 +57,7 @@ describe("isStandardSyntaxSelector", () => {
   it("SCSS placeholder", () => {
     expect(isStandardSyntaxSelector("%foo")).toBeFalsy();
   });
-  it("Resolved nested selectors", () => {
+  it("Less mixin with resolved nested selectors", () => {
     expect(isStandardSyntaxSelector(".foo().bar")).toBeFalsy();
     expect(isStandardSyntaxSelector(".foo(@a, @b).bar")).toBeFalsy();
     expect(isStandardSyntaxSelector(".foo()#bar")).toBeFalsy();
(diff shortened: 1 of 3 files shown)


lib/utils/__tests__/isStandardSyntaxSelector.test.js
```diff
@@ -57,4 +57,16 @@ describe("isStandardSyntaxSelector", () => {
   it("SCSS placeholder", () => {
     expect(isStandardSyntaxSelector("%foo")).toBeFalsy();
   });
+  it("Resolved nested selectors", () => {
```
jeddy3: "Less mixin with resolved nested selectors"

lib/utils/isStandardSyntaxSelector.js
```diff
@@ -21,5 +21,10 @@ module.exports = function(selector /*: string*/) /*: boolean*/ {
     return false;
   }
 
+  // Resolved nested selectors (e.g. .foo().bar or .foo(@a, @b)[bar])
```
jeddy3: "Less mixin with resolved nested selectors (e.g. .foo().bar or .foo(@a, @b)[bar])"

jwilsson:
True, that's better!

jeddy3:
> -   Fixed: `selector-max-specificity` cannot parse selector violation for Less mixins ([#2677](https://github.com/stylelint/stylelint/pull/2677)).

@jwilsson Thanks. It's great to see some of these long-standing Less bugs squashed!

PR #2802, opened 2017-08-08
Update complementary-tools.md

alexilyaev:
stylelint-find-rules is in good shape now, it can show deprecated rules and unconfigured rules, also follows through extends.


ntwb:
Thanks @alexilyaev 👍

Issue #2810, opened 2017-08-13
Fix false positives for multiple pseudo-class in selector-pseudo-class-no-unknown


Mottie:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

When multiple pseudoclasses (chained) are added to the webkit scrollbar selector, we see a "selector-pseudo-class-no-unknown" error. In this example css:

```css
::-webkit-scrollbar-button:horizontal:decrement {
  background-color: transparent;
}

::-webkit-scrollbar-button:decrement:horizontal {
  background-color: transparent;
}

::-webkit-scrollbar-button:horizontal {
  background-color: transparent;
}

::-webkit-scrollbar-button:decrement {
  background-color: transparent;
}
```

The first two selectors show the error:

    1:38 error Unexpected unknown pseudo-class selector ":decrement" (selector-pseudo-class-no-unknown)
    5:37 error Unexpected unknown pseudo-class selector ":horizontal" (selector-pseudo-class-no-unknown)

The standard stylelint configuration is being used, and the error can be seen after entering the above css into the [main stylelint demo](https://stylelint.io/demo/).

> Which version of stylelint are you using?

`8.0.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

Reproducible in the online demo

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

No "selector-pseudo-class-no-unknown" error.

> What actually happened (e.g. what warnings or errors you are getting)?

"selector-pseudo-class-no-unknown" error.

evilebottnawi:
@Mottie thanks for issue, seems it is bug

ntwb:
Fixed via https://github.com/stylelint/stylelint/commit/982024dd03380338d96bd1f7d48492708c87338d

PR #2813, opened 2017-08-14
Fixed: respect multiple pseudo classes for `webkit` pseudo elements in `selector-pseudo-class-no-unknown` rule.

evilebottnawi:

#2810 

No, it's self explanatory.

### DIFF
--- a/lib/rules/selector-pseudo-class-no-unknown/__tests__/index.js
+++ b/lib/rules/selector-pseudo-class-no-unknown/__tests__/index.js
@@ -90,6 +90,10 @@ testRule(rule, {
     {
       code: "html { --custom-property-set: {} }",
       description: "custom property set in selector"
+    },
+    {
+      code: "::-webkit-scrollbar-button:horizontal:decrement {}",
+      description: "webkit scrollbar multiple pseudo classes"
     }
   ],
 
@@ -141,6 +145,12 @@ testRule(rule, {
       message: messages.rejected(":window-inactive"),
       line: 1,
       column: 18
+    },
+    {
+      code: "::-webkit-scrollbar-button:horizontal:unknown {}",
+      message: messages.rejected(":unknown"),
+      line: 1,
+      column: 38
     },
     {
       code: ":first { }",
--- a/lib/rules/selector-pseudo-class-no-unknown/index.js
+++ b/lib/rules/selector-pseudo-class-no-unknown/index.js
@@ -84,12 +84,17 @@ const rule = function(actual, options) {
               return;
             }
 
-            if (pseudoNode.prev()) {
+            const firstSelectorNode =
+              pseudoNode.parent && pseudoNode.parent.nodes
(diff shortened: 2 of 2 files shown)


lib/rules/selector-pseudo-class-no-unknown/index.js
```diff
@@ -84,9 +84,17 @@ const rule = function(actual, options) {
               return;
             }
 
-            if (pseudoNode.prev()) {
+            const firstSelectorNode =
+              pseudoNode.parent && pseudoNode.parent.nodes
```
evilebottnawi: Should we check `parent` here? Be good to find use case where `pseudoNode` doesn't contain `parent` (if this use case exists).

lib/rules/selector-pseudo-class-no-unknown/__tests__/index.js
```diff
@@ -90,6 +90,10 @@ testRule(rule, {
     {
       code: "html { --custom-property-set: {} }",
       description: "custom property set in selector"
+    },
+    {
+      code: "::-webkit-scrollbar-button:horizontal:decrement {}",
+      description: "webkit scrollbar multiple pseudo classes"
```
ntwb: Let's use `description: "webkit scrollbar chained pseudo classes"` as chained is the correct terminology I believe

lib/rules/selector-pseudo-class-no-unknown/index.js
ntwb: cc @jeddy3 thoughts?

ntwb: We can probably sneak this into 8.1.0 if no other chanfes to the PR are required

lib/rules/selector-pseudo-class-no-unknown/__tests__/index.js
ntwb: I made this change in https://github.com/stylelint/stylelint/pull/2813/commits/21cf021454fe9a07c045513c2193398117daa5ee

evilebottnawi:
@ntwb thanks, no time now, sorry :disappointed:

lib/rules/selector-pseudo-class-no-unknown/index.js
jeddy3: > Should we check parent here?

Sure, no harm in guarding against it.

jeddy3:
This PR just requires a rebase and a small tweak, then I think we can get it in.

evilebottnawi:
@jeddy3 done

ntwb:
Changelog via https://github.com/stylelint/stylelint/commit/20b200979a717526a73d1b094b9d93af3d2bb97d
> -   Fixed: `selector-pseudo-class-no-unknown` false positives when using chained pseudo-classes ([#2810](https://github.com/stylelint/stylelint/issues/2810)).

PR #2827, opened 2017-08-22
Fix `isFirstNested` util to handle shared-line comments

### LINKED ISSUE #2764: False positives for first-nested preceded by a shared-line comment in *-empty-line-before

A bug where i combine comment and sass extend.

e.g. `at-rule-empty-line-before`

e.g.

```scss
&:checked {
  + .e-accordion__title:nth-child(n) { //nth-child: safari not repaint bug fix
    @extend .e-accordion__title-active;
  }
}
```

e.g.

```json
{
  "rules": {
    "at-rule-empty-line-before": [ "always", {
      "except": [ "blockless-after-blockless", "first-nested" ],
      "ignore": ["after-comment"],
      "ignoreAtRules": ["else"]
    }],
  }
}
```

`8.0.0`

CLI with `stylelint app/components/**/**/*.scss`

No, it Can be reproduced with normal css comment.

No warnings to be flagged.

```shell
app/components/components/accordion/_accordion.scss
 99:11  ✖  Expected empty line before at-rule   at-rule-empty-line-before
```

nwoltman:

Fixes #2764

This changes the `isFirstNested` util to ignore comments that are on the same line as the opening brace of a block when checking if the provided node is the first node in the block.

### DIFF
--- a/lib/utils/isFirstNested.js
+++ b/lib/utils/isFirstNested.js
@@ -4,9 +4,44 @@
 module.exports = function(statement /*: postcss$node*/) /*: boolean*/ {
   const parentNode = statement.parent;
 
-  return (
-    parentNode !== undefined &&
-    parentNode.type !== "root" &&
-    statement === parentNode.first
-  );
+  if (parentNode === undefined || parentNode.type === "root") {
+    return false;
+  }
+
+  if (statement === parentNode.first) {
+    return true;
+  }
+
+  /*
+   * Search for the statement in the parent's nodes, ignoring comment
+   * nodes on the same line as the parent's opening brace.
+   */
+
+  const parentNodes = parentNode.nodes;
+  const firstNode = parentNodes[0];
+
+  if (firstNode.type !== "comment" || firstNode.raws.before.includes("\n")) {
+    return false;
+  }
+
+  const openingBraceLine = firstNode.source.start.line;
+
+  if (openingBraceLine !== firstNode.source.end.line) {
+    return false;
+  }
+
+  for (let i = 1; i < parentNodes.length; i++) {
+    const node = parentNodes[i];
+
+    if (node === statement) {
+      return true;
+    }
+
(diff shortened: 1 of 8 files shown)


lib/utils/isFirstNested.js
```diff
@@ -4,9 +4,44 @@
...
+    if (node === statement) {
+      return true;
+    }
+
+    if (node.type !== "comment" || node.source.end.line !== openingBraceLine) {
+      return false;
+    }
+  }
+
+  /* istanbul ignore next: Should always return in the loop */
```
evilebottnawi: Why this line?

nwoltman: It's impossible for the code execution to reach here. One of the `return` statements in the for loop will always be executed. This is guaranteed because it returns if it finds the `statement` node in `parent.nodes` and that must happen because that node must be in its own parent's list of nodes.

If this is too strange I could change the `return false;` in the for loop into `break;` so then this `return false;` at the end of the function will be executed.

evilebottnawi: @nwoltman all good, just question, thanks!

hudochenkov: Why comment starts with “istanbul ignore next:”?

nwoltman: Jest uses Istanbul for analyzing code coverage. That comment tells Istanbul not to count that line in the total code coverage since it's impossible to execute (the [documentation](https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md) is in the old Istanbul repo).

jeddy3:
Changelog:

> -   Fixed: `*-empty-line-before` false positives shared-line comments and `"first-nested"` option ([#2827](https://github.com/stylelint/stylelint/issues/2827)).  

@nwoltman Thanks for this! Sorry about the delay getting it in. A few of us (the stylelint contributors) have been away/relocating.

Issue #2757, opened 2017-07-17
Update to postcss-selector-parser 3


ben-eb:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Feature request (dependency upgrade). I've done a [pre-release of the new 3.x code](https://github.com/postcss/postcss-selector-parser/releases/tag/v3.0.0-rc.0) and as it is breaking a few things I wanted to get some feedback before I land 3.0.0 for real. In my use case it performs well if not better than expected, but stylelint uses postcss-selector-parser much more extensively.

Note that we now can accept PostCSS `rule` nodes and will throw a `CssSyntaxError` when the parser chokes on invalid syntaxes. I hope this provides a better end-user experience. 😄 

Your feedback is important to me! Thanks.

> Which rule, if any, is this issue related to?

n/a

> What CSS is needed to reproduce this issue?

n/a

> What stylelint configuration is needed to reproduce this issue?

n/a

> Which version of stylelint are you using?

`8`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

n/a

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

n/a

> What did you expect to happen?

n/a

> What actually happened (e.g. what warnings or errors you are getting)?

n/a

jeddy3:
@ben-eb Firstly, thanks for your continued work on `postcss-selector-parser`. Your selector parser invaluable to stylelint.

An initial test run using the pre-release gives:

```
Test Suites: 5 failed, 261 passed, 266 total
Tests:       226 failed, 12692 passed, 12918 total
Snapshots:   6 passed, 6 total
Time:        135.296s
Ran all test suites.
```

Those failing suites are for:

- `"selector-attribute-brackets-space-inside"`
- `"selector-attribute-operator-blacklist"`
- `"selector-attribute-operator-space-after"`
- `"selector-attribute-operator-space-before"`
- `"selector-attribute-operator-whitelist"`

I suspect this due to the new attribute selector parsing and whitespace handling. I'm hoping it'll be some minor changes at the stylelint end to fix compatibility.

Unfortunately, I personally doubt I'll be able to take a deeper look into this for a couple of weeks as I'm away. @evilebottnawi has labelled has correctly labelled the issue as `help wanted`, so hopefully someone will some free time to pick this up.

ben-eb:
Any joy?

hudochenkov:
[...]

PR #2828, opened 2017-08-22
Work in progress testing `postcss-selector-parser` pre-release `3.0.0-rc.0`

ntwb:
Ref: #2757 

WIP: Work in progress testing `postcss-selector-parser` pre-release `3.0.0-rc.0`


lib/rules/selector-attribute-operator-blacklist/index.js
```diff
@@ -37,7 +37,7 @@ const rule = function(blacklistInput) {
 
       parseSelector(rule.selector, result, rule, selectorTree => {
         selectorTree.walkAttributes(attributeNode => {
-          const operator = attributeNode.operator;
+          const operator = attributeNode.operator.trim();
```
hudochenkov: What if there no `attributeNode.operator`? `.trim()` will throw an error. In the next condition, there is a check `!operator`.

jeddy3: Yes, you're right. I'll hold off making the change [until we know](https://github.com/stylelint/stylelint/issues/2757#issuecomment-333786942) if this is something we should be addressing in stylelint.

ntwb:
Superseded by #2984

PR #2839, opened 2017-08-30
Fix outdated example rules object in config docs

br0p0p:
Copying this example into my configuration caused an error on build (`stylelint-commands` is expected):

`Invalid value "stylelint-command" for option "ignore" of rule "comment-empty-line-before"`


hudochenkov:
Thank you!

PR #2842, opened 2017-09-01
Prettier JS fixes

ntwb:
As noted at https://github.com/stylelint/stylelint/pull/2834#pullrequestreview-58393680 there was a strange _prettier_ change in that PR that was unexpected, this PR is to to test against `master` any missing changes that should have been added by Prettier already.

• I've added `prettier:check` and `prettier:fix` npm commands, `npm run prettier:check` and `npm run prettier:fix` respectively to help test changes in this PR

Running `npm run prettier:check` results in:
```shell
prettier '**/*.js'  --list-different

.coverage/lcov-report/prettify.js
.coverage/lcov-report/sorter.js
decls/flow-typed/lodash.js

decls/flow-typed/lodash.js: SyntaxError: default is a reserved word (463:55)
  461 |     conforms(source: Object): Function;
  462 |     constant<T>(value: T): () => T;
> 463 |     defaultTo<T1:string|boolean|Object,T2>(value: T1, default: T2): T1;
      |                                                       ^
  464 |     // NaN is a number instead of its own type, otherwise it would behave like null/void
  465 |     defaultTo<T1:number,T2>(value: T1, default: T2): T1|T2;
  466 |     defaultTo<T1:void|null,T2>(value: T1, default: T2): T2;
decls/postcss.js
decls/stylelint.js
lib/__tests__/standalone.test.js
lib/getConfigForFile.js
lib/normalizeRuleSettings.js
lib/postcssPlugin.js
lib/rules/indentation/__tests__/functions.js
lib/rules/indentation/index.js
lib/standalone.js
```

The results after running `npm run prettier:fix` are the file changes in this PR (apart from `package.json` change adding the two script commands)

- [ ] decls/postcss.js
- [ ] decls/stylelint.js
- [ ] lib/__tests__/standalone.test.js
- [ ] lib/getConfigForFile.js
- [ ] lib/normalizeRuleSettings.js
- [ ] lib/postcssPlugin.js
- [ ] lib/rules/indentation/__tests__/functions.js
- [ ] lib/rules/indentation/index.js
- [ ] lib/standalone.js

@stylelint/core could you review this PR please and we'll get this shipped in 8.1.0

PR #2843, opened 2017-09-01
Update flowtype definitions

ntwb:
Related: #2842


ntwb:
This PR resolves 1 issue from #2842 and should be merged into `master` or that PR

Issue #2812, opened 2017-08-14
Add 'peerDependencies' guideline to plugin writing documentation


ybiquitous:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

I propose to add guideline which plugin authors should declare *stylelint* as `peerDependencies`
to [stylelint plugin writing documentation](https://stylelint.io/developer-guide/plugins/) such as [ESLint](http://eslint.org).

[ESLint documentation](http://eslint.org/docs/developer-guide/working-with-plugins#peer-dependency) says:

> To make clear that the plugin requires ESLint to work correctly you have to declare ESLint as a peerDependency in your package.json.
> The plugin support was introduced in ESLint version 0.8.0. Ensure the peerDependency points to ESLint 0.8.0 or later.

And [npm documentation](https://docs.npmjs.com/files/package.json#peerdependencies) says:

> In some cases, you want to express the compatibility of your package with a host tool or library,
> while not necessarily doing a require of this host. This is usually referred to as a plugin.

This proposal reason is that different versions of *stylelint* can be installed **unexpectedly** at same time,
if you will install some plugins depending on different versions of *stylelint*.

For example:

`stylelint-plugin-a` depends to `stylelint@^7.0.0`.

```json
{
  "name": "stylelint-plugin-a",
  "version": "1.0.0",
  "dependencies": {
    "stylelint": "^7.0.0"
  }
}
```

On the other hand, `stylelint-plugin-b` depends to `stylelint@^8.0.0` as `peerDependencies`.

```json
{
  "name": "stylelint-plugin-b",
  "version": "1.0.0",
  "peerDependencies": {
    "stylelint": "^8.0.0"
  }
}
```

Then if installing these plugins, warning about `peerDependencies` will be output.

```shellsession
$ npm install ../stylelint-plugin-a ../stylelint-plugin-b
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN stylelint-plugin-b@1.0.0 requires a peer of stylelint@^8.0.0 but none was installed.
npm WARN using-stylelint-plugins@1.0.0 No description
npm WARN using-stylelint-plugins@1.0.0 No repository field.
npm WARN using-stylelint-plugins@1.0.0 No license field.

+ stylelint-plugin-a@1.0.0
+ stylelint-plugin-b@1.0.0
added 2 packages in 0.981s

$ npm ls | grep stylelint
[...]

PR #2850, opened 2017-09-04
docs: Add description about peer dependency for writing plugin

ybiquitous:

#2812

I am not confident in English, so please give me feedbacks for incorrect or strange wording etc. 😓

### DIFF
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -163,21 +163,11 @@ In addition to the standard parsers mentioned in the ["Working on rules"](rules.
 
 Have a look through [stylelint's internal utils](https://github.com/stylelint/stylelint/tree/master/lib/utils) and if you come across one that you need in your plugin, then please consider helping us extract it out into an external module.
 
-## Peer dependency
+## Peer dependencies
 
-To make clear that your plugin requires which versions of stylelint, you have to declare stylelint as a `peerDependency` (**NOT** `dependency`) in your `package.json`. Because, if a user would try to install plugins declared as a `dependency` which each require different versions of stylelint, different versions of stylelint would be installed **unexpectedly** at same time.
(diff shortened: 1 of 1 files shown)


docs/developer-guide/plugins.md
```diff
@@ -163,6 +163,30 @@ In addition to the standard parsers mentioned in the ["Working on rules"](rules.
 
 Have a look through [stylelint's internal utils](https://github.com/stylelint/stylelint/tree/master/lib/utils) and if you come across one that you need in your plugin, then please consider helping us extract it out into an external module.
 
+## Peer dependency
+
+To make clear that your plugin requires which versions of stylelint, you have to declare stylelint as a `peerDependency` (**NOT** `dependency`) in your `package.json`. Because, if a user would try to install plugins declared as a `dependency` which each require different versions of stylelint, different versions of stylelint would be installed **unexpectedly** at same time.
```
jeddy3: Let's go with the following:

---

## Peer dependencies

You should express, within the `peerDependency` key (and **not** within the `dependency` key) of your plugin's `package.json`, what version(s) of stylelint your plugin can be used with. This is to ensure that different versions of stylelint are not unexpectedly installed.

For example, to express that your plugin can be used with stylelint versions 7 and 8:

```json
{
  "peerDependencies": {
    "stylelint": "^7.0.0 || ^8.0.0"
  }
}
```
---

As much as I love emojis, not using them is more inline with the current tone of the documentation. Also, I've tried to make it a little terser as our documentation is growing a lot.

ybiquitous: @jeddy3 Thank you for review!

Certainly, the use of emojis in this case seems a little strange 😓 

Also, I feel my documentation is roundabout, so I agree you proposal which is more clear and easier to understand 👍

ybiquitous:
@jeddy3 I pushed d76664e. Please review again 🙏

Issue #2424, opened 2017-03-14
Add ignoreProperties: [] to value-keyword-case


oleksiibilous:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

"A feature request"

> Which rule, if any, is this issue related to?

`value-keyword-case`

> What CSS is needed to reproduce this issue?

```css
.elementWithCamelCaseName {}

.element {
  composes: elementWithCamelCaseName;
}
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "value-keyword-case": "lower",
    }
  }
}
```

> Which version of stylelint are you using?

`6.7.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

"WebpackPluginStylelint"

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

"Yes, it's related to CSS modules"

> What did you expect to happen?

"No warnings to be flagged for composes property"

> What actually happened (e.g. what warnings or errors you are getting)?

"The following warnings were flagged: Expected camelCaseName to be lowecasename"

Maybe it is a good idea to add additional option to this rule? Something like `ignoreProperties`?  I can work on PR if anyone is interested. Thank you

evilebottnawi:
@oleksiibilous thanks for feedback, be right ignore `composes` property by default, because it is always `custom-ident`, PR :+1:

jeddy3:
> be right ignore composes property by default,

Let's not add code to core for _methodologies_. That's what the `ignore*` options are for.

> Something like `ignoreProperties`? 

Yes, that would be perfect. @oleksiibilous Feel free [to contribute this option](https://stylelint.io/developer-guide/rules/#adding-a-option-to-an-existing-rule) via a PR.

oleksiibilous:
`ignoreProperties` sounds good👍 Will work on it

CAYdenberg:
@stylelint/core Should this issue be closed? It looks like it was fixed in #2937 ?

PR #2858, opened 2017-09-06
fix: Ignore `:local` and `:global` pseudo classes in the `selector-max-specificity` rule

### LINKED ISSUE #2857: Add support for evaluation contexts pseudo-classes in selector-max-specificity

False positive when using CSS modules

`selector-max-specificity`

```css
.lets.goto :local(.bar) {
  color: pink;
}
```

```json
{
  "rules": {
    "selector-max-specificity": "0,3,0"
  }
}
```

`8.0.0`

CLI with `stylelint --ignore-path .eslintignore '**/*.css'`

Yes, it's related to CSS modules

No warnings to be flagged.

```shell
✖  Expected ".lets.goto :local(.bar)" to have a specificity no more than "0,3,0"   selector-max-specificity
```

> Figure out what needs to be done, propose it

Ignore `:local` and `:global`, as it's done with `:not` and `:matches`

ntwb:
Fixes #2857

### DIFF
--- a/lib/rules/selector-max-specificity/README.md
+++ b/lib/rules/selector-max-specificity/README.md
@@ -12,7 +12,7 @@ Visit the [Specificity Calculator](https://specificity.keegan.st) for visual rep
 
 This rule ignores selectors with variable interpolation (`#{$var}`, `@{var}`, `$(var)`).
 
-This rule ignores selectors containing the `:not()` or `:matches()` pseudo-classes.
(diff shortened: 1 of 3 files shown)


hudochenkov:
I thought we're not adding methodologies related code in our rules. Issues like that came up few times. https://github.com/stylelint/stylelint/issues/2424#issuecomment-287642870

Hypnosphi:
Ok, how do I set up `ignore*` then?

evilebottnawi:
@Hypnosphi need implement in rule, PR welcome

Issue #2866, opened 2017-09-08
Growing the stylelint team


jeddy3:
I think we should try to grow the stylelint team. Over the past two years, the number of stylelint users has grown significantly. The number of active stylelint team members has, however, remained pretty constant. Also, for most of the team, myself included, we don't have as much time as we used to to give to the project. There simply isn't enough time amongst us to review all the PRs and triage all the issues anymore.

I think growing the stylelint team would be healthy and would help the sustainability of the project.

There are a couple of contributors I can think of off-hand:

- @CAYdenberg [has made](https://github.com/stylelint/stylelint/pulls?q=is%3Apr+author%3ACAYdenberg+is%3Aclosed) a number of contributions over the past few months.
- @modosc [has offered, and has started,](https://github.com/stylelint/stylelint/issues/2829) to port some of the stylefmt fixes over as he has experience with the stylefmt code base.
- @gucong3000 [has been active](https://github.com/stylelint/stylelint/pulls?utf8=%E2%9C%93&q=author%3Agucong3000%20) in both the stylelint and stylefmt.

@gucong3000 @CAYdenberg & @modosc Would becoming stylelint members be something you're interested in?

As team members:

1. You'll be able to help review pull requests, merge them and update the CHANGELOG.
2. You'll be able to help triage issues.
3. You'll be able to help put out releases.
4. You'll be able to help decide the direction that stylelint takes going forward (see: [VISION.md](https://github.com/stylelint/stylelint/blob/master/VISION.md)).

(I'll try to knock-up, in the next couple of days, additions to the developer guide to formalise the first two of these. We have an informal approach to issues and PRs that isn't documented anywhere).

I'm sure there are a few more suitable people I've failed to mention. Shout out if anyone else is interested! Either because you wish to support a tool you use, or as an opportunity to get involved in a open source project. As we say in the [developer guide](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md):

> If you like stylelint and open source software (since you're reading this, you almost certainly do), please consider taking some time to pitch in. Not only will you help stylelint thrive, you will also learn a thing or two — about CSS, PostCSS, Node, ES2015, unit testing, open source software, and more.

Also, are there any companies out there that use stylelint as an integral part of their tool chain and are able to spare some of their developers' time? If so, please let us know!

[...]

Issue #2837, opened 2017-08-29
No error on undefined variables


Mottie:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Undefined variables should report an error. Enter the following into the online demo and it reports "No warnings"

```css
/* should show an error */
a { color: var(--undef); }

/* this variable has a fallback, but should also show an error if the config is set */
a { color: var(--undef, red); }

/* this variable has nested fallbacks, but should also show an error if the config is set */
a { color: var(--undef, var(--undef2, red)); }

```

When the browser evaluates these variables, it sets the color to `rgba(0, 0, 0, 0);` for the first entry, the other two fallback to red.

> Which rule, if any, is this issue related to?

A new rule would be required.

> What CSS is needed to reproduce this issue?

See above

> What stylelint configuration is needed to reproduce this issue?

Using the [online demo](https://stylelint.io/demo/) default settings

> Which version of stylelint are you using?

`8.0.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

Online demo

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Nope

> What did you expect to happen?

A warning or error should notify of an undefined variable.

> What actually happened (e.g. what warnings or errors you are getting)?

Nada

jeddy3:
@Mottie Thanks for the suggestion. Labeling as `plugin`. Feel free to [create](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/plugins.md) `no-unknown-custom-properties` as a plugin). See https://github.com/stylelint/stylelint/issues/2220#issuecomment-284677340.

Mottie:
It might be nice to have a plugin repository where we can transfer these rule proposals. When they're closed here, people don't usually go searching for rule ideas, when they get the urge to work on something 😉.

jeddy3:
> It might be nice to have a plugin repository where we can transfer these rule proposals.

It's a good suggestion and I see the value in it, but I don't think it adds enough to warrant the ongoing effort as it's for a small audience. There's already so much to manage, maintain and juggle that we're consciously streamlining and reducing things-we-have-to-maintain.  We had a wishlist on the wiki but it wasn't updated.

PR #2868, opened 2017-09-09
Add pull request and issue sections to Dev Guide

jeddy3:
Ref: https://github.com/stylelint/stylelint/issues/2866

@hudochenkov Thanks for [digging out](https://github.com/stylelint/stylelint/issues/2866#issuecomment-328209568) those comments. I had completely forgotten about them :)

I’ve tried to streamline the pr and issue processes where I can e.g. I’ve removed the labels which we add to closed issues. Adding these labels is unnecessary work as this info is either: 

1. Communicated in the closing comment e.g. “Best suited as a plugin”, “Dupe of xxxx” etc.
2. Implicit e.g. the PR was merged after a review, the PR is open and doesn’t have a “PR: needs tests/docs/revision” label and therefore needs reviewing.

The actual documents themselves are pretty terse (and a bit dry I’m afraid) as I wanted to get something up before we add any new members (and it's for a small audience).

---

I'll also try to make some time over the next few days to help work through the backlog of issues so there's a clean-ish slate for any new members...

### DIFF
--- a/docs/developer-guide/prerequisites.md
+++ b/docs/developer-guide/prerequisites.md
@@ -1,0 +1,4 @@
+# Technical prerequisites to contributing
+
+-   [Node.JS](https://nodejs.org/en/) 8+.
+-   Latest [npm](https://www.npmjs.com/).
--- a/docs/developer-guide/prerequisites.md
+++ b/docs/developer-guide/prerequisites.md
@@ -1,4 +1,4 @@
 # Technical prerequisites to contributing
 
--   [Node.JS](https://nodejs.org/en/) 8+.
+-   [Node.js](https://nodejs.org/en/) 4.2.1+.
 -   Latest [npm](https://www.npmjs.com/).
--- a/docs/developer-guide/pull-requests.md
+++ b/docs/developer-guide/pull-requests.md
@@ -1,0 +1,13 @@
+# Reviewing pull requests
+
+-   Use the [GitHub review system](https://help.github.com/articles/about-pull-request-reviews/).
+-   Review against the [Developer Guide criteria](rules.md).
+-   Assign one or more of the appropriate [`PR: needs *` labels](https://github.com/stylelint/stylelint/labels) when requesting a change.
+-   Merge simple documentation fixes after one approval and all other PRs after two approvals.
+-   Update the [CHANGELOG](https://github.com/stylelint/stylelint/blob/master/CHANGELOG.md) directly via the [GitHub website](https://github.com/stylelint/stylelint/edit/master/CHANGELOG.md) for all merged PRs (except documentation changes).
+    -   Create a `# Head` heading if one does not exist already.
+    -   Prefix the item with either: Removed, Changed, Deprecated, Added, or Fixed.
+    -   Order the item within the group by the widest reaching first to the smallest, and then alphabetically by rule name.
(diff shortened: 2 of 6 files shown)


docs/developer-guide/pull-requests.md
```diff
@@ -0,0 +1,13 @@
+# Reviewing pull requests
+
+-   Use the [GitHub review system](https://help.github.com/articles/about-pull-request-reviews/).
+-   Review against the [Developer Guide criteria](rules.md).
+-   Assign one or more of the appropriate [`PR: needs *` labels](https://github.com/stylelint/stylelint/labels) when requesting a change.
+-   Merge simple documentation fixes after one approval and all other PRs after two approvals.
```
hudochenkov: We could also add:

-   Keep git history clean:
    -   Start a commit message with uppercase letter.
    -   Consider squash and merge commits instead simple merge. It's useful when PR has “Fix lint error,” “Fix typo,” “Update test snapshot”, etc.
    -   Use [git rebase](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase) in your own PRs to avoid unnecessary “Merge ...” commits.

jeddy3: Good idea. I've added bits about commit messages, squashing and rebasing.

jeddy3:
> There is also worth mention, that we need update package-lock.json if we update dependencies. Use latest npm, as it recommended.

Isn't the `package-lock.json` file automatically updated with a `npm install package@latest` command? If so, I think we don't need to document it... but we can always reassess if it becomes an issue.

hudochenkov:
> Isn't the `package-lock.json` file automatically updated with a `npm install package@latest` command?

It is. But maintainer should use `npm@5` for updating dependencies. Not everybody uses latest `npm`.

jeddy3:
> It is. But maintainer should use npm@5 for updating dependencies. Not everybody uses latest npm.

Fair point. I've added a "technical prerequisites to contributing" stub. 

I also broke the developer guide into sections so that prerequisites make sense (and the user guide for consistency).

docs/developer-guide/prerequisites.md
```diff
@@ -0,0 +1,4 @@
+# Technical prerequisites to contributing
+
+-   [Node.JS](https://nodejs.org/en/) 8+.
```
hudochenkov: The correct name is “Node.js.” Also, 8+ isn't a legit requirement. Personally, I'm using Node.js 6, and I'm going to switch to Node.js 8 only when it became LTS. I think the most correct version is 4 because it's a minimal version for stylelint itself.

jeddy3: You're right.

ntwb:
This is 👌 , thanks @jeddy3 😄

hudochenkov:
BTW, @jeddy3 didn't you delete by accident `type: plugin` label, when was deleting other labels? Just noticed, that this label is missing https://github.com/stylelint/stylelint/issues/2837#event-1232380222.

jeddy3:
> BTW, @jeddy3 didn't you delete by accident type: plugin label, when was deleting other labels? Just noticed, that this label is missing #2837 (comment).

I deleted it as part of streamlining the processes (and therefore what I had to document). No 1. in https://github.com/stylelint/stylelint/pull/2868#issue-256440907.

PR #2878, opened 2017-09-11
Linkify changelog releases

ntwb:
A PR looking to standardize our changelog along the lines of http://keepachangelog.com

@stylelint/core What's the teams thoughts on "standardising" our changelog per keepachangelog?


jeddy3:
Are these changes compatible with our release script, which generates a GitHub release from the CHANGELOG? I think the current release process is working well and I'd be reluctant to revisit it unless these changes add a lot of benefits for users.

ntwb:
> Are these changes compatible with our release script, which generates a GitHub release from the CHANGELOG? I think the current release process is working well and, if it isn't compatible, I'd be reluctant to revisit it unless these changes add a lot of benefits for users.

Agreed, I shall confirm this doesn't break that 👍

ntwb:
Closing, will revisit maybe later

ntwb:
"Keep a Changelog" support was added in https://github.com/MoOx/github-release-from-changelog/pull/11

PR #2886, opened 2017-09-13
Add autofix support for CSS with standard syntax

gucong3000:

no

Autofix css standard syntax error by `postcss-safe-parser`.


jeddy3:
@gucong3000 Thanks! 

As per [the guide](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/issues.md), only use `type:` labels for issues. The idea is that you should _create an issue to discuss a feature first_, before submitting a PR. Those issues are then assigned `type:` labels and PRs should reference the original issue.

package.json
```diff
@@ -85,15 +86,13 @@
     "eslint-config-stylelint": "^7.0.0",
     "file-exists-promise": "^1.0.2",
     "flow-bin": "^0.54.0",
-    "fs-promise": "^2.0.3",
```
ntwb: I know this package has been deprecated in favour of `fs-extra` per https://www.npmjs.com/package/fs-promise but the `pify` package has not been added via npm install pify --save-dev` to replace `fs-promise` here in `package.json`

lib/__tests__/fixtures/syntax_error.scss
```diff
@@ -0,0 +1 @@
+a {
```
ntwb: There should be a _end-of-file_ new line here.

package.json
jeddy3: I believe `pify` is already available [as a dependency](https://github.com/stylelint/stylelint/blob/90149a5e650e1adb138206b29e055847da59f4ba/package.json#L60)

ntwb: That it is, my bad, sorry 😏

jeddy3:
@gucong3000 Sorry about the delay merging this in. I've only had time to triage issues and not look into PRs. With 3 approvals, this can be merged by any of the team :)

jeddy3:
> -   Added: autofix of syntax errors in standard CSS e.g. unclosed braces and brackets ([#2886](https://github.com/stylelint/stylelint/issues/2886)).

Issue #2454, opened 2017-03-30
`require('stylelint')` is very slow


chris-morgan:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Loading stylelint (`require('stylelint')`) is very slow. It takes over a second for me when there’s nothing else installed, and when there are various other things installed it increases (e.g. two seconds).

I care about this because in a Makefile world where you spin up the tooling for each file, this is a >1 second delay per file which rapidly changes from unpleasant to awful.

Steps to reproduce:

```
$ cd $(mktemp -d)
$ yarn add stylelint
[…]
$ node -e "s=Date.now();require('stylelint');console.log(Date.now()-s)"
1183
$ node -e "s=Date.now();require('stylelint');console.log(Date.now()-s)"
1128
$ node -e "s=Date.now();require('stylelint');console.log(Date.now()-s)"
1155
```

> Which rule, if any, is this issue related to?

n/a

> What CSS is needed to reproduce this issue?

n/a

> What stylelint configuration is needed to reproduce this issue?

n/a

> Which version of stylelint are you using?

7.9.0

> How are you running stylelint: CLI, PostCSS plugin, Node API?

n/a

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

n/a

> What did you expect to happen?

`require('stylelint')` to be fast (like, well under 100ms).

> What actually happened (e.g. what warnings or errors you are getting)?

It took a long time to import stylelint, a second or two.

davidtheclark:
@chris-morgan Can you contrast this with other programs of similar complexity? Is there anything actionable that can be done?

chris-morgan:
@davidtheclark Well, here are a bunch of other things I was trying with postcss:

postcss: 109
postcss-cssnext: 67
postcss-custom-properties: 3
postcss-at-rules-variables: 2
postcss-conditionals: 21
postcss-clean: 204
autoprefixer: 267
stylelint: 1887

And I consider postcss-clean and autoprefixer to be far too slow (over 200ms) as well. Seriously, 100ms is a *long time*.

I don’t know *why* stylelint is so slow at present, but it is *very* slow and must be doing *something* that it shouldn’t be!

davidtheclark:
@chris-morgan Can you compare it to a module of similar size? I imagine involves stylelint has many, many more files than any of those modules you've listed.

chris-morgan:
[...]

PR #2889, opened 2017-09-14
Generate color functions for named colors rather than hardcoding.

erwinw:
#2454

Generate color functions for named colors rather than hardcoding. Eases the way to future color model changes, reduces namedColorData.js file size by 90% (60KB -> 6KB) (part of issue #2454).


lib/rules/color-named/index.js
```diff
@@ -46,7 +48,14 @@ const rule = function(expectation, options) {
       return;
     }
 
-    const namedColors = Object.keys(namedColorData);
+    const namedColors = Object.keys(namedColorDataHex);
+    const namedColorData = {};
+    Object.entries(namedColorDataHex).forEach(([name, hex]) => {
```
jeddy3: Travis [is failing](https://travis-ci.org/stylelint/stylelint/builds/275372897?utm_source=github_status&utm_medium=notification) on node@6 and node@4 as they [seemingly don't support](http://node.green/#ES2017-features-Object-static-methods-Object-entries). Can you please rewrite this snippet of code for node@4.2.1?

erwinw:
This rule has gone from

```
Mean: 29.60662716949152 ms
Deviation: 13.743477932583556 ms
```

to

```
Mean: 31.658794166666663 ms
Deviation: 14.69102207909526 ms
```

-- as expected, since I've moved pre-computed to runtime computed information. Per #2454 the idea was to reduce general require time by reducing the size of this include; not sure how to measure the impact of that, though?

lib/rules/color-named/index.js
erwinw: Done!

ntwb:
> _"-- as expected, since I've moved pre-computed to runtime computed information. Per #2454 the idea was to reduce general require time by reducing the size of this include; not sure how to measure the impact of that, though?")

I'm fine with this, the amount of code removed i.e. `+358 −2,560` I think offsets this 👍

erwinw:
Note that when [CSS Color Module Level 4](https://drafts.csswg.org/css-color/) hits this rule will need work, for example because they drop the comma separators in color functions (so `rgb(255 127 0)` is valid) -- but at the least it'll be easier to add support for that sort of stuff..

jeddy3:
Thanks for running the benchmarks. Just checking that the impact is acceptable, which it is!

> not sure how to measure the impact of that, though?

Would the following give us a rough idea?

```
node -e "s=Date.now();require('./lib/index.js');console.log(Date.now()-s)"
```

There might be a lot of deviation to make it meaningful, though.

Either way, this PR LGTM as, as you said, it eases the way to future color model changes.

Issue #2853, opened 2017-09-05
Fix false positives for CSS in HTML in *-empty-line-before


gucong3000:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

`*-empty-line-before` not support for HTML

> Which rule, if any, is this issue related to?

`rule-empty-line-before`, `at-rule-empty-line-before`

> What CSS is needed to reproduce this issue?

It is CSS in HTML.

```html
<style>
h1 {
  font-weight: normal;
}
</style>
```

> What stylelint configuration is needed to reproduce this issue?

```js
module.exports = {
  extends: 'stylelint-config-standard',
  // add your custom rules here
  rules: {
  }
};

```

> Which version of stylelint are you using?

`8.1.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

```shell
stylelint "**/*.{vue,html,css,sss,less,scss,sass}" --custom-syntax postcss-html
```

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

no.

> What did you expect to happen?

"No warnings to be flagged."

> What actually happened (e.g. what warnings or errors you are getting)?

"The following errors were flagged:"
https://circleci.com/gh/gucong3000/webpack/16#artifacts/containers/0

```shell
src/App.vue
 15:1  ✖  Expected empty line before rule   rule-empty-line-before

src/components/Hello.vue
 36:1  ✖  Expected empty line before rule   rule-empty-line-before
```

PR #2892, opened 2017-09-14
fix `rule-empty-line-before` false for CSS in HTML with ignore: "inside-block"

gucong3000:

#2853 

No, it's self explanatory.

### DIFF
--- a/lib/__tests__/fixtures/rule-empty-line-before.html
+++ b/lib/__tests__/fixtures/rule-empty-line-before.html
@@ -5,6 +5,9 @@
 h1 {
   font-weight: normal;
 }
+h2 {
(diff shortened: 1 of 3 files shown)


lib/__tests__/fixtures/rule-empty-line-before.html
```diff
@@ -5,6 +5,9 @@
 h1 {
```
jeddy3: Would the following?

```css
@media (min-width: 100px) {
  h1 {
    font-weight: normal;
  }
}
```

And no warnings be a more appropriate test for this change as your fixing false positives, right?

gucong3000:
I submitted the wrong branch name

Issue #2891, opened 2017-09-14
Fix false negatives for ignore: ["inside-block"] in rule-empty-line-before


gucong3000:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Cannot monitor the missing blank line before rule in HTML.

> Which rule, if any, is this issue related to?

`rule-empty-line-before`

> What CSS is needed to reproduce this issue?

```html
<style>
.foo {
  color: pink;
}
.bar {
  color: pink;
}
</style>
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "rule-empty-line-before": ["always", {
      "ignore": "inside-block"
    }]
  }
}
```

> Which version of stylelint are you using?

8.1.1

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI
```shell
stylelint "**/*.html" --custom-syntax postcss-html
```

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

standard syntax in HTML.

> What did you expect to happen?

The following warnings were flagged in line that `.bar {`

> What actually happened (e.g. what warnings or errors you are getting)?

No warnings to be flagged.

gucong3000:
@jeddy3
Maybe we need a new label. `non-standard Syntax: html`

evilebottnawi:
@gucong3000 `non-standard syntax: html` is misleading, because `css` in `style` tag is standard syntax. Maybe `processor: html`

jeddy3:
> Maybe we need a new label. non-standard syntax: html or some other description.

Done.

> because css in style tag is standard syntax. Maybe processor: html

Ummm. Good point. 

`postcss-html` is being used via the `--custom-syntax` option, right?:

```
stylelint "**/*.html" --custom-syntax postcss-html
```

What's the relationship/differences between `postcss-html` and the html processor?

gucong3000:
> `postcss-html` is being used via the `--custom-syntax` option, right?:

yes.

> What's the relationship/differences between `postcss-html` and the html processor?

`postcss-html` can restore HTML files from AST. This means `postcss-html` compatible with that the feature of stylelint  - autofixing.

jeddy3:
> postcss-html can restore HTML files from AST. This means postcss-html compatible with that the feature of stylelint - autofixing.

Thanks for explaining that. And am I right in thinking that not being able to do that is a current limitation of the processes? If so, is this a limitation that can be overcome?

PR #2894, opened 2017-09-14
Fix false negatives for ignore: ["inside-block"] in rule-empty-line-before

gucong3000:

#2891

No, it's self explanatory.

### DIFF
--- a/lib/__tests__/fixtures/rule-empty-line-before.html
+++ b/lib/__tests__/fixtures/rule-empty-line-before.html
@@ -5,6 +5,9 @@
 h1 {
   font-weight: normal;
 }
+h2 {
+  font-weight: normal;
+}
 </style>
 </head>
 </html>
--- a/lib/__tests__/standalone-syntax.test.js
+++ b/lib/__tests__/standalone-syntax.test.js
@@ -75,7 +75,12 @@ it("standalone with postcss-html syntax", () => {
     rules: {
       "no-empty-source": true,
       "comment-empty-line-before": "always",
-      "rule-empty-line-before": "always",
+      "rule-empty-line-before": [
+        "always",
+        {
+          ignore: ["inside-block"]
+        }
+      ],
       "at-rule-empty-line-before": [
         "always",
         {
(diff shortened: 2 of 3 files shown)


lib/__tests__/fixtures/rule-empty-line-before.html
```diff
@@ -5,6 +5,9 @@
 h1 {
```
jeddy3: Would the following?

```css
@media (min-width: 100px) {
  h1 {
    font-weight: normal;
  }
}
```

And no warnings be a more appropriate test for this change as your fixing false positives, right?

gucong3000: > Would the following?
It has nothing to do with this.
The affected is that the <style> tag first-level child element is misjudged as nested.

> And no warnings be a more appropriate test for this change as your fixing false positives, right?

It is not "false positives", It is "missed warnings"
(I don't know if my English is appropriate, title modified, please help correct English grammar)

jeddy3: > The affected is that the <style> tag first-level child element is misjudged as nested.

Oooo, I see now. Sorry, I misunderstood. 

> It is not "false positives", It is "missed warnings"

I've just been using "false negatives" for those :)

lib/__tests__/standalone-syntax.test.js
```diff
@@ -120,8 +125,12 @@ it("standalone with postcss-html syntax", () => {
     const ruleEmptyLineBeforeResult = results.find(r =>
       /[/\\]rule-empty-line-before\.html$/.test(r.source)
     );
-    expect(ruleEmptyLineBeforeResult.errored).toBeFalsy();
-    expect(ruleEmptyLineBeforeResult.warnings.length).toBe(0);
+    expect(ruleEmptyLineBeforeResult.errored).toBe(true);
+    expect(ruleEmptyLineBeforeResult.warnings.length).toBe(1);
+    expect(ruleEmptyLineBeforeResult.warnings[0].line).toBe(8);
```
ntwb: What are the 8 warnings added in this change?

gucong3000: not 8 warnings, [just 1](https://github.com/stylelint/stylelint/pull/2894/files/51e7a7f7b8dfa93e0bc1dc4da241768a06eaef4f#diff-7d31f5d52b54f165816fa0ff4e0de2dfR132).

jeddy3:
> -   Fixed: `rule-empty-line-before` false positives for `ignore: ["inside-block"]` and CSS in HTML ([#2894](https://github.com/stylelint/stylelint/issues/2894)).

Issue #2833, opened 2017-08-24
Fix isPathIgnored regression for .stylelintignore files


proProbe:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

There is a bug where the API apparently doesnt read .stylelintignore-files. CLI works fine but not the API.

> Which rule, if any, is this issue related to?

no rule

> What CSS is needed to reproduce this issue?

No actual css

> What stylelint configuration is needed to reproduce this issue?

doesnt matter

> Which version of stylelint are you using?

^8.0.0

> How are you running stylelint: CLI, PostCSS plugin, Node API?

In vscode with 
https://github.com/shinnn/vscode-stylelint

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Nope

> What did you expect to happen?

That the API adheres to the stylelintignore

> What actually happened (e.g. what warnings or errors you are getting)?

All my files gets linted through the API even though they do not get linted in the CLI

> Additional info

I created an issue on the vscode-stylelint repo. The author stated that there is a problem with the API. The CLI works as expected though.
https://github.com/shinnn/vscode-stylelint/issues/101#event-1219431167

evilebottnawi:
@proProbe `--ignore-path .gitignore`, default ignore files have name `.stylelintignore` (without `-files`)

proProbe:
@evilebottnawi Oh Im sorry, I incorrectly wrote .gitignore instead of .stylelintignore in the title. Changed that now

evilebottnawi:
@proProbe for api you can use `ignorePath` option (https://stylelint.io/user-guide/node-api/#ignorepath and https://github.com/stylelint/stylelint/blob/master/lib/standalone.js#L59). 
```js
stylelint.lint({
  config: { ...confguration, ignorePath: '/path/to/ignore-file' },
  files: "all/my/stylesheets/*.css"
})
  .then(function(data) {
    // do things with data.output, data.errored,
    // and data.results
  })
  .catch(function(err) {
    // do things with err e.g.
    console.error(err.stack);
  });;
```

proProbe:
@shinnn I'm kinda new to this. Any of the above suggestions that can be used?

Arcanemagus:
@proProbe The easiest way to handle this is to ask stylelint to tell you whether the file is ignored before attempting to run a lint:

[...]

PR #2897, opened 2017-09-14
Don't lint file if it is ignored and code is added directly

billybonks:

https://github.com/stylelint/stylelint/issues/2833

### Enviroment

.stylelintignore
`tests/fixtures/ignore/*`

folder structure
image

run `stylelint.lint` via node

```
 this.linterConfig.code = content; //code
 this.linterConfig.codeFilename = path.join(this.inputNodesDirectory, relativePath);// relativepath
 return stylelint.lint(this.linterConfig)....
```

### Code flow
when running the linter with `code` options it flows through https://github.com/billybonks/stylelint/blame/ignore-codefilename/lib/standalone.js#L112

which uses `micromatch` and config.ignoredFiles to test for ignores using absolutePaths
https://github.com/stylelint/stylelint/blame/master/lib/isPathIgnored.js#L4

if you run the linter with an array of file paths it flows through
https://github.com/billybonks/stylelint/blame/ignore-codefilename/lib/standalone.js#L175

which uses ignorer.filter using relativePaths.

I am not sure why there are multiple ignore strategies.

### Code change overview

- include `ignoreText` and `ignorePattern` options to stylelint
- in `isPathIgnored` test the same way that files are filtered without the code option. 
- I test both the above and the original for backward compatiblity.

### Remaining work
- if this is accepted we should add a test for the API change.
- I recommend deprecating the mini match strategy for consistency.

### DIFF
--- a/decls/stylelint.js
+++ b/decls/stylelint.js
@@ -33,6 +33,8 @@ export type stylelint$options = {
   configFile?: string,
   configBasedir?: string,
   configOverrides?: Object,
+  ignoreText?: string,
+  ignorePattern?: array,
   ignoreDisables?: boolean,
   ignorePath?: string,
   reportNeedlessDisables?: boolean,
--- a/lib/isPathIgnored.js
+++ b/lib/isPathIgnored.js
@@ -1,5 +1,6 @@
 /* @flow */
 "use strict";
+const ignore = require("ignore");
 const micromatch = require("micromatch");
 const path = require("path");
 
@@ -20,7 +21,14 @@ module.exports = function(
     const absoluteFilePath = path.isAbsolute(filePath)
       ? filePath
       : path.resolve(process.cwd(), filePath);
-    if (micromatch(absoluteFilePath, config.ignoreFiles).length) {
+    const relativePath = absoluteFilePath.replace(`${process.cwd()}/`, "");
+    const ignorer = ignore()
+      .add(stylelint._options.ignoreText)
+      .add(stylelint._options.ignorePattern);
+    if (
+      micromatch(absoluteFilePath, config.ignoreFiles).length ||
+      ignorer.ignores(relativePath)
+    ) {
       return true;
     }
     return false;
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -103,6 +103,8 @@ module.exports = function(
     reportNeedlessDisables,
     syntax,
     customSyntax,
+    ignoreText,
+    ignorePattern,
     fix
   });
 


Arcanemagus:
If I'm understanding this correctly you have two things going on in this PR:

* Fixing the bug with `isPathIgnored` not handling `.stylelintignore` files
* Adding two entirely new options to the API for some reason

Unifying the way that ignores are handled is great, but does that actually fix the bug here? I thought that would involve fixes in [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) but I could easily be wrong.

Your second part adding new features should be a different PR as it has nothing to do with fixing the bug.

billybonks:
> Adding two entirely new options to the API for some reason

The reason for doing this is to make the ignore data available to the `isPathIgnored` function. it does fix the bug because I use the ignore data to build the ignorer.

The other options would:

1. pass the ignore file path, and to turn this into a helper,

https://github.com/stylelint/stylelint/blob/master/lib/standalone.js#L59-L72.

2. which I find more preferable is to add `ingorer.ignores(relativePath)` just before this line
https://github.com/billybonks/stylelint/blob/a55580438584e39a502381df5d34d747aec57fd9/lib/standalone.js#L112

but then we would have to stub the postCss results since we would skip the step altogether, it will be beneficial for performance.
https://github.com/stylelint/stylelint/blob/f6fbad37e1ffa84aed4a996b3829b4275e7320c5/lib/lintSource.js#L52

It may be beneficial to solve both at the same time, vs building a dirty patch.

I did more digging into the whole ignore story.

The normal flow when you don't pass code directly seems to be
- ignore files using styelint ignore
-while linting keep checking if file matches, globs provided by https://stylelint.io/user-guide/configuration/#ignorefiles.

CAYdenberg:
Hey @billybonks thanks for your work on this.

I think @Arcanemagus's concern is about changing interface design just to support a new implementation or bugfix. If we're going to change the outward-facing API that affects everyone using stylelint, and there needs to be a good reason for doing so (they need to be passing input that isn't redundant with any other options or the .stylelintignore file). And of course it needs documentation.

billybonks:
Hey guys happy to make a revision here, I propose bundling the ignorer into the singleton that will be set up in standalone.js and then required in `isPathIgnored` ?

CAYdenberg:
Awesome! I was actually just looking into this myself last night ...

I think the first thing to do is to write a (failing) test for isPathIgnored that examines the .stylelintignore file. We currently have [only one test](https://github.com/stylelint/stylelint/blob/master/lib/__tests__/createLinter.test.js#L30) for isPathIgnored that only looks at the ignore option passed to the API - we should set up another one that uses the [test .stylelintignore file](https://github.com/stylelint/stylelint/tree/master/lib/__tests__/stylelintignore-test).

Once that's done I'm personally fine with using a singleton especially if it simplifies some of the other logic surrounding how ignores are handled ...

rwwagner90:
@billybonks any updates on this?

CAYdenberg:
Superseded by #3021 branch from that and contribute if you'd like to help.

Issue #2788, opened 2017-07-30
Fix false positives for at-charset and single quotes in string-quotes


LiHaoTan:
When specifying the rule `string-quotes: "single"`, then specifying the encoding through `@charset` single quotes is required as well, e.g. `@charset 'UTF-8';`.

However, this is incorrect as per the specification double quotes are always required:

> Otherwise, check the byte stream. If the first 1024 bytes of the stream begin with the hex sequence
40 63 68 61 72 73 65 74 20 22 XX* 22 3B
where each XX byte is a value between 016 and 2116 inclusive or a value between 2316 and 7F16 inclusive, then get an encoding for the sequence of XX bytes, interpreted as ASCII

https://drafts.csswg.org/css-syntax-3/#determine-the-fallback-encoding

ntwb:
Thanks @LiHaoTan, I've marked this issue up as a bug and help wanted, would you be interested in creating a pull request to fix this?

From the cited source above: (emphasis mine)
> Note: Note that the syntax of an encoding declaration looks like the syntax of an at-rule named `@charset`, but no such rule actually exists, and the rules for how you can write it are much more restrictive than they would normally be for recognizing such a rule. **A number of things you can do in CSS that would produce a valid `@charset` rule (if one existed), such as using multiple spaces, comments, or single quotes, will cause the encoding declaration to not be recognized**. This behavior keeps the encoding declaration as simple as possible, and thus maximizes the likelihood of it being implemented correctly.

evilebottnawi:
@stylelint/core what do your think about this, should we ignore `@charset` or output error if your use `single` for `@charset`?

ntwb:
Generally I think this should throw an error:

```css
@charset 'UTF-8';
```

What about a new rule, `no-invalid-charset-quotes` in the "Possible errors" rule category?

Message `Expected double quotes, @charset encoding declarations must use double quotes`

The `string-quotes` rule would then _ignore_ instances of `@charset`

evilebottnawi:
@ntwb sgtm, I would like to hear one more opinion :smile:

jeddy3:
> The string-quotes rule would then ignore instances of @charset

SGTM

jeddy3:
@LiHaoTan This issue is labelled as "help wanted". If you'd like to see this issue closed, please 
[...]

PR #2902, opened 2017-09-17
Fix false positives for at-charset and single quotes in string-quotes #2788 

CAYdenberg:

#2788 

No, self-explanatory

### DIFF
--- a/lib/rules/string-quotes/README.md
+++ b/lib/rules/string-quotes/README.md
@@ -10,6 +10,8 @@ a[id="foo"] { content: "x"; }
 
 Quotes within comments are ignored.
 
+Also, the quotes in a `charset` @-rule are ignored as using single quotes in this context is incorrect according the [CSS specification](https://www.w3.org/TR/CSS2/syndata.html#x57).
+
 ```css
 /* "This is fine" */
 /* 'And this is also fine' */
--- a/lib/rules/string-quotes/README.md
+++ b/lib/rules/string-quotes/README.md
@@ -10,12 +10,19 @@ a[id="foo"] { content: "x"; }
 
 Quotes within comments are ignored.
 
-Also, the quotes in a `charset` @-rule are ignored as using single quotes in this context is incorrect according the [CSS specification](https://www.w3.org/TR/CSS2/syndata.html#x57).
 
 ```css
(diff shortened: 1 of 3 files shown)


lib/rules/string-quotes/README.md
```diff
@@ -10,6 +10,8 @@ a[id="foo"] { content: "x"; }
 
 Quotes within comments are ignored.
 
+Also, the quotes in a `charset` @-rule are ignored as using single quotes in this context is incorrect according the [CSS specification](https://www.w3.org/TR/CSS2/syndata.html#x57).
```
ntwb: Let's move this line to below the _quotes in comments_ code underneath this sentence

Update the verbiage to `Single quotes in a charset @-rule are ignored as using single quotes in this context is incorrect according the CSS specification.`

Maybe see what adding a code example for this also looks like, it might look weird or add confusion 🤔

lib/rules/string-quotes/__tests__/index.js
```diff
@@ -40,6 +40,10 @@ testRule(rule, {
     {
       code: 'a { /* "horse" */ }',
       description: "ignores comment"
+    },
+    {
+      code: '@charset "utf-8"',
+      description: "ignore @charset rules"
     }
```
ntwb: Could you add this same test to the "double quotes" test around line #119 as a safe guard of sorts please

CAYdenberg: Added below ...

ntwb: Thanks, though you added it to the [SCSS tests](https://github.com/stylelint/stylelint/blob/6d89566dca654f990ff26d4a11d44cb9b5edee58/lib/rules/string-quotes/__tests__/index.js#L160-L163) at lines 160-163, which won't hurt either, can you add it after line 118 please :)

CAYdenberg: Whoops. Yes added there.

jeddy3:
> -   Fixed: `string-quotes` false positives for `@charset` and single quotes ([#2902](https://github.com/stylelint/stylelint/issues/2902)).

PR #2916, opened 2017-09-21
Fix links in PULL_REQUEST_TEMPLATE

pkuczynski:
Seems that links were outdated. With this change, they are not :)


ntwb:
Thanks @pkuczynski 😄

pkuczynski:
You welcome :)

Issue #2829, opened 2017-08-22
Add autofix to rules for parity with stylefmt


modosc:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

feature request: start merging in fix code from https://github.com/morishitter/stylefmt

(see also https://github.com/morishitter/stylefmt/issues/291)

the `stylefmt` repo has been very quiet for a while, the [last commit](https://github.com/morishitter/stylefmt#notice-consider-other-tools-before-adopting-stylefmt) recommends people use `stylint --fix` instead. 

i've written some fixes there and bugfixed others - there is a lot of code from this project copied there for parsing / etc so i think porting many of the rules would be straightforward. 

maybe opening a new issue up for each rule would be a start? i'm happy to help with some this, we rely on this working at my workplace. 

> Which rule, if any, is this issue related to?

none

> What CSS is needed to reproduce this issue?

none

> What stylelint configuration is needed to reproduce this issue?

none

> Which version of stylelint are you using?

current

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylelint --config myconfig *.css`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

no

> What did you expect to happen?

nothing

> What actually happened (e.g. what warnings or errors you are getting)?

nothing

---

Edit - list of rules:

- [x] [at-rule-semicolon-newline-after](https://github.com/stylelint/stylelint/tree/master/lib/rules/at-rule-semicolon-newline-after)
- [x] [block-closing-brace-newline-after](https://github.com/stylelint/stylelint/tree/master/lib/rules/block-closing-brace-newline-after)
- [x]  [block-closing-brace-newline-before](https://github.com/stylelint/stylelint/tree/master/lib/rules/block-closing-brace-newline-before)
- [x]  [block-opening-brace-newline-after](https://github.com/stylelint/stylelint/tree/master/lib/rules/block-opening-brace-newline-after) - #3441
- [x]  [block-opening-brace-newline-before](https://github.com/stylelint/stylelint/tree/master/lib/rules/block-opening-brace-newline-before)
- [ ] once we start fixing/changing newlines we need to revisit [this conversation](https://github.com/stylelint/stylelint/pull/2787) and add appropriate tests
- [x]  [block-opening-brace-space-after](https://github.com/stylelint/stylelint/tree/master/lib/rules/block-opening-brace-space-after)
[...]

PR #2917, opened 2017-09-21
Auto fix for number-leading-zero rule

pkuczynski:
This provides --fix for `number-leading-zero` rule


lib/rules/number-leading-zero/index.js
```diff
@@ -56,16 +56,30 @@ const rule = function(expectation) {
 
         // Check leading zero
         if (expectation === "always") {
-          const match = /(?:\D|^)(\.\d+)/.exec(valueNode.value);
+          const regexp = /((?:\D|^))(\.\d+)/;
```
gucong3000: I think it should be better:
```js
const regexp = /(^|\D)(\.\d+)/;
```

lib/rules/number-leading-zero/index.js
```diff
@@ -77,13 +91,28 @@ const rule = function(expectation) {
         }
 
         if (expectation === "never") {
-          const match = /(?:\D|^)(0+\.\d+)/.exec(valueNode.value);
+          const regexp = /((?:\D|^))(0+)(\.\d+)/;
```
gucong3000: I think it should be better:
```js
const regexp = /(^|\D)(0+)(\.\d+)/;
```

lib/rules/number-leading-zero/index.js
```diff
@@ -56,16 +56,30 @@ const rule = function(expectation) {
...
             return;
           }
 
+          if (context.fix) {
+            if (node.value) {
+              node.value = node.value.replace(regexp, "$10$2");
+              return;
+            }
+
+            if (node.params) {
```
gucong3000: From coverage reports, it's never true:
https://coveralls.io/builds/13380484/source?filename=lib%2Frules%2Fnumber-leading-zero%2Findex.js#L72

lib/rules/number-leading-zero/index.js
```diff
@@ -77,13 +91,28 @@ const rule = function(expectation) {
...
           }
 
-          const capturingGroupIndex = match[0].length - match[1].length;
+          if (context.fix) {
+            if (node.value) {
+              node.value = node.value.replace(regexp, "$1$3");
+              return;
+            }
+
+            if (node.params) {
```
gucong3000: From coverage reports, it's never true:
https://coveralls.io/builds/13380484/source?filename=lib%2Frules%2Fnumber-leading-zero%2Findex.js#L107

lib/rules/number-leading-zero/index.js
pkuczynski: I haven't used it in the beginning, but then it was not working for the example `code: "@media (min-width: .01em)"`, as it does not have a node value, just node params. So maybe I am missing something here. It's my first PR to stylelint. Can you help me on this one?

pkuczynski: @jeddy3 can you have a look at this one too? This one I would actually need quite quick :)

jeddy3: @gucong3000 Can you expand on your comment? 

I believe [the rule also checks](https://github.com/pkuczynski/stylelint/blob/bf5ee41da9b4db384392f9874e06b252bedb2e67/lib/rules/number-leading-zero/index.js#L27-L33) _media features_, as well as _declarations_. `atRule` nodes have a `params` property, rather than a `value` one. 

Are you suggesting the use of [postcss-media-query-parser](https://github.com/dryoma/postcss-media-query-parser) to parse out the _value_ of the _media feature_, thus normalising the behaviour to only changing _values_ regardless of whether it's a _declaration_ or _media feature_? I guess this would be preferable as it would scope the regex to the smallest possible area.

jeddy3: > This one I would actually need quite quick :)

I believe the core team is spread across different time zones and so the review process can take a few days. I'll hold off doing a release until the weekend though, in the hopes that your two PRs make it in.

pkuczynski: Yes, this rule checks both properties and media queries. I am not Post CSS expert, so I would need your advice how to improve this code...

I understand the time zones issue. It would be awesome if this change could hit the nearest release...

gucong3000: What I'm worried about is that using the regular replacement string directly, may not be able to cope with unknown complexities.

gucong3000: And here's the inner of `valueParser(value).walk()`.

pkuczynski: Not sure if regexp could break anything here, but you are right, you never know in regexp :) I do not understand the comment about `valueParser(value).walk()`. You would have to be more specific to help me go through this...

gucong3000: > I do not understand the comment about `valueParser(value).walk()`.
https://github.com/TrySound/postcss-value-parser#valueparserwalknodes-callback-bubble
callback function will recursive calls multiple times.
This line of code is not run once for each `node`.

modosc:
i just noticed this - i duplicated it in #2921 . i was working off the list on #2829 - i'll double check before i start another one (currently almost finished with `number-no-trailing-zeros`).

jeddy3:
> i just noticed this - i duplicated it in #2921 

It's a shame there was a duplication of effort. Oh well, it's unlikely to happened again. But to help, let's all add a comment to https://github.com/stylelint/stylelint/issues/2829 when starting on a new rule, letting everyone else know.

>  (currently almost finished with number-no-trailing-zeros).

Good stuff! As per the comments in this issue, be sure to fix via `postcss-value-parser`'s `valueNode` (of type `word`).

lib/rules/number-leading-zero/index.js
jeddy3: @gucong3000 Can you push your proposed changes to @pkuczynski branch, please? I think that'll help us get this PR in.

@pkuczynski I believe @gucong3000 is pointing out that your fix happens [within the postcss-value-parser `walk` loop](https://github.com/pkuczynski/stylelint/blob/bf5ee41da9b4db384392f9874e06b252bedb2e67/lib/rules/number-leading-zero/index.js#L43-L125), and you are therefore setting `node.value` for _every `word` node_ rather than just once.

jeddy3:
@pkuczynski @modosc What's the best way of combining your two PRs together and getting this autofix in? 

I think this PR needs to address @gucong3000 concerns around redundancy of updating the node's value within the `postcss-value-parser` loop. I think @modosc explicit use of `if (node.type === "atRule")`, rather than `if (node.params)`, is clearer... can we adopt that in this PR?

modosc:
@jeddy3 i can make a new branch off of this and then add those requested changes (unless @pkuczynski wants to do it).

jeddy3:
> @jeddy3 i can make a new branch off of this and then add those requested changes

SGTM

modosc:
@jeddy3 see #2921 - i think it covers everything now?

jeddy3:
@pkuczynski Thanks for your work on this. @modosc Picked it up in https://github.com/stylelint/stylelint/pull/2921

pkuczynski:
Sure. No probs.

PR #2920, opened 2017-09-22
Fix how "except" options are checked in *-empty-line-before rules

### LINKED ISSUE #2919: Fix false positives for shared-line comments in nested blocks in rule-empty-line-before

A bug where a comment at the top of a nested block using the `rule-empty-line-before` rule returns false results. This appears to be a regression from 8.0.0, possibly triggered by #2827.

`rule-empty-line-before`

```css
@media screen { /* comment */
    .foo {
        display: none;
    }
}
```

e.g.

```json
{
  "rules": {
    "rule-empty-line-before": [ "always", {
      "except": ["after-single-line-comment", "first-nested"]
    }]
  }
}
```

`8.1.1` (also reproduced on current master https://github.com/stylelint/stylelint/commit/c856d321db2fd33bbef0c98b443a294877abe3ba)

CLI with `stylelint test.css`

No

No warnings to be flagged.

```shell
test.css
 2:5  ✖  Expected empty line before rule   rule-empty-line-before
```

nwoltman:

Fixes #2919

The bug reported in #2919 was being caused by multiple "except" options being checked individually, which caused the expectation to flip-flop and be incorrect if an even number of exceptions were satisfied.

This PR fixes that type of code in the various `*-empty-line-before` rules so that as soon as one exception is satisfied, it skips the rest. The other `*-empty-line-before` rules that I did not touch were already doing this correctly.

### DIFF
--- a/lib/rules/rule-empty-line-before/__tests__/index.js
+++ b/lib/rules/rule-empty-line-before/__tests__/index.js
@@ -155,6 +155,10 @@ testRule(rule, {
     {
       code: "@media {\n\na {}\nb {} }",
       description: "nested"
+    },
+    {
+      code: "a {} /* comment */\nb {}",
+      description: "shared-line comment"
     }
   ],
 
@@ -199,7 +203,8 @@ testRule(rule, {
       code: "/* comment */\na {}"
     },
     {
-      code: "@media { /* comment */\na {} }"
+      code: "@media { /* comment */\n\na {} }",
+      description: "nested shared-line comment"
     }
   ],
 
@@ -215,9 +220,16 @@ testRule(rule, {
       message: messages.rejected
     },
     {
-      code: "@media { /* comment */\n\na {} }",
-      fixed: "@media { /* comment */\na {} }",
-      message: messages.rejected
+      code: "@media { /* comment */\na {} }",
+      fixed: "@media { /* comment */\n\na {} }",
+      message: messages.expected,
+      description: "nested shared-line comment"
+    },
+    {
+      code: "a {} /* comment */\nb {}",
+      fixed: "a {} /* comment */\n\nb {}",
+      message: messages.expected,
+      description: "shared-line comment"
     }
   ]
 });
@@ -237,7 +249,7 @@ testRule(rule, {
     },
     {
       code: "@media { /* comment */\n  a {}\n\n}",
-      description: "shared-line comment boog"
+      description: "shared-line comment"
     },
     {
       code: "@media {\n  a {}\n\n  b{}\n\n}"
@@ -308,6 +320,19 @@ testRule(rule, {
       code: "@media {\n  b {}\n  a {}\n\n}",
       fixed: "@media {\n  b {}\n\n  a {}\n\n}",
       message: messages.expected
+    }
+  ]
+});
+
+testRule(rule, {
+  ruleName,
+  config: ["always", { except: ["after-single-line-comment", "first-nested"] }],
+
+  accept: [
+    {
+      code:
+        "@media screen { /* comment */\n  .foo {\n    display: none;\n  }\n}",
+      description: "shared-line comment - issue #2919"
     }
   ]
 });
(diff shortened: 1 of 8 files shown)


jeddy3:
@nwoltman Oh wow, thanks for the quick fix!

> The bug reported in #2919 was being caused by multiple "except" options being checked individually, which caused the expectation to flip-flop and be incorrect if an even number of exceptions were satisfied.

It's surprising that this hasn't come to light before! Good catch, though.

Is there still an issue here with _shared-line_ comments not being ignored by `rule-empty-line-before`, though e.g.

```css
@media screen { /* comment */
    .foo {
        display: none;
    }
}
```

The `/* comment */` should be ignored (as all shared-line comments are ignored), but it is triggering the `except: ["after-single-line-comment"]` option. This exception revealed the "flip-flop" bug you've fixed as `.foo {}` is also the `"first-nested"`. However, the `"after-single-line-comment"` option should never have been triggered to begin with.

I think we should address this within this PR. What do you think?

lib/rules/rule-empty-line-before/index.js
```diff
@@ -153,6 +131,20 @@ const rule = function(expectation, options, context) {
   };
 };
 
+function isAfterRule(rule) {
+  const prevRule = rule.prev();
+  return prevRule && prevRule.type === "rule";
+}
+
+function isAfterSingleLineComment(rule) {
+  const prevRule = rule.prev();
```
jeddy3: Should like use the `isSharedLineComment` util?

lib/rules/rule-empty-line-before/index.js
```diff
@@ -153,6 +131,20 @@ const rule = function(expectation, options, context) {
   };
 };
 
+function isAfterRule(rule) {
+  const prevRule = rule.prev();
```
jeddy3: Should this use the `getPreviousNonSharedLineCommentNode` util?

nwoltman:
@jeddy3 Ah yes, you're right that shared-line comments should be ignored and shouldn't trigger `"after-single-line-comment"`. I'll fix up this PR to handle that.

lib/rules/rule-empty-line-before/index.js
nwoltman: Turns out there's already an `isAfterCommentLine` util so this function doesn't even need to exist.

nwoltman: Whoops, never mind about that; `isAfterCommentLine` also counts multi-line comments.

Also, when I fix this function, a few tests start failing:

`rule-empty-line-before › accept › ["always", {"except": ["after-single-line-comment"]}]`

```css
@media { /* comment */
  a {}
}
```

`rule-empty-line-before › reject › ["always", {"except": ["after-single-line-comment"]}]`

```css
@media { /* comment */

  a {}
}
```

They actually need to be reversed now because the shared-line comment doesn't count as a single-line comment, so the first test should be rejected and the second one should be accepted. This might need to be highlighted in the release notes of the release that includes this fix.

jeddy3: > They actually need to be reversed now because the shared-line comment doesn't count as a single-line comment

Sounds about right. I had assumed we weren't testing for this eventually. However, it turns out we are but the tests are wrong.

> This might need to be highlighted in the release notes of the release that includes this fix.

Good point. Due to our [Semantic Versioning Policy](https://github.com/stylelint/stylelint#semantic-versioning-policy) this _can be_ included in a minor release, even though it could break some users' builds, though.

jeddy3: > Whoops, never mind about that; `isAfterCommentLine` also counts multi-line comments.

Feel free to create a `isAfterSingleLineComment` util, though.

And also free feel to rename the `isAfterCommentLine` util to `isAfterComment` for consistency with the other `isAfter*` utils. I'm not sure the `Line` suffix makes sense now that _shared-line_ comments are always ignored.

nwoltman:
@jeddy3 I've updated the PR. Would it be alright if I renamed the `isAfterCommentLine` util in a separate PR? I feel like that change out be a little out of place in this PR.

lib/rules/rule-empty-line-before/__tests__/index.js
```diff
@@ -314,6 +326,19 @@ testRule(rule, {
 
 testRule(rule, {
   ruleName,
+  config: ["always", { except: ["after-single-line-comment", "first-nested"] }],
+
+  accept: [
+    {
+      code:
+        "@media screen { /* comment */\n  .foo {\n    display: none;\n  }\n}",
+      description: "shared-line comment - issue #2919"
```
ntwb: _We should try to do this more often when warranted_ 👌

jeddy3:
> -   Fixed: `*-empty-line-before` false negatives and positives when two or more `except: [*]` options were triggered ([#2920](https://github.com/stylelint/stylelint/issues/2920)).
> -   Fixed: `rule-empty-line-before` false positives for `except: ["after-single-line-comment"]` and preceding shared-line comments ([#2920](https://github.com/stylelint/stylelint/issues/2920)).

PR #2921, opened 2017-09-22
Add autofix support to number-leading-zero

modosc:

#2829 

"No, it's self explanatory."

### DIFF
--- a/lib/rules/number-leading-zero/README.md
+++ b/lib/rules/number-leading-zero/README.md
@@ -8,6 +8,8 @@ a { line-height: 0.5; }
  * This leading zero */
 ```
 
+The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofixing-errors) can automatically fix all of the problems reported by this rule.
+
 ## Options
 
 `string`: `"always"|"never"`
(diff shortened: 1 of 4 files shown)


modosc:
ugh, just saw #2917.

modosc:
i think this should be good now?

jeddy3:
> i think this should be good now?

Thanks, looking good.

@gucong3000 Does this address your concerns in https://github.com/stylelint/stylelint/pull/2917?

gucong3000:
> Does this address your concerns in #2917?

Yes.
And this PR looking good.

modosc:
> It doesn't seem to support multiple values. Like this:

fixed. i need to check the trailing zero code as well since i'm expecting it to fail in the same way.

modosc:
> fixed. i need to check the trailing zero code as well since i'm expecting it to fail in the same way.

actually the `no-trailing-zeros` code did this right and is slightly cleaner - i copied that approach here. i also added tests for this case to `no-trailing-zeros` for completeness.

jeddy3:
> -   Added: `number-leading-zero` autofix ([#2921](https://github.com/stylelint/stylelint/issues/2921)).

Issue #2766, opened 2017-07-19
Add Greenkeeper support for package-lock.json


ntwb:
It looks like we need to add a GitHub token to the repo for Greenkeeper to be able to push to branches from within Travis CI

https://github.com/greenkeeperio/greenkeeper-lockfile#setup

Noticed as part of #2765: https://travis-ci.org/stylelint/stylelint/jobs/255347993#L1879
```shell
0.86s$ greenkeeper-lockfile-update
error: Your local changes to the following files would be overwritten by merge:
	package-lock.json
Please, commit your changes or stash them before you can merge.
Aborting
child_process.js:631
    throw err;
    ^
Error: Command failed: git revert -n HEAD
error: Your local changes to the following files would be overwritten by merge:
	package-lock.json
Please, commit your changes or stash them before you can merge.
Aborting
```

hudochenkov:
@ntwb can you take care of this problem, please? You're the best of us who understand in npm@5 and Greenkeeper things. We've already missed some dependencies updates, which breaks stylelint for npm <5 users.

https://github.com/stylelint/stylelint/issues/2775
https://github.com/stylelint/stylelint/issues/2768

ntwb:
@jeddy3 Now that you're back from 🏖  (which I hope was awesome) could you take a look at this please? 😄 

Following the details here: https://github.com/greenkeeperio/greenkeeper-lockfile#setup

1. We need a GitHub token created for each the stylelint repos that use Greenkeeper
2. For each repo the token added as environment variable `GH_TOKEN` to e.g https://travis-ci.org/stylelint/stylelint/settings 

After that I can then create the PRs for each repo to implement the required changes 🤖

jeddy3:
> Now that you're back from 🏖 (which I hope was awesome)

Thanks :)

> could you take a look at this please? 

I've created a token and added it as the `GH_TOKEN` environment variable on Travis to:

- `stylelint`
- `stylelint-demo`
- `stylelint-config-standard`
- `stylelint-config-standard`
- `eslint-config-stylelint`

Anything else to do?

ntwb:
- `stylelint` Changes required added previously ✅ 
- `stylelint-demo` Changes required added previously ✅ 
- `stylelint-config-recommended` https://github.com/stylelint/stylelint-config-recommended/pull/8
- `stylelint-config-standard` https://github.com/stylelint/stylelint-config-standard/pull/91
[...]

PR #2936, opened 2017-10-02
Disable package-lock

jeddy3:
Closes https://github.com/stylelint/stylelint/issues/2766


ntwb:
As noted on the other issue, if strangeness occurs Travis CI caches may need purging.

PR #2937, opened 2017-10-02
Add ignoreProperties option to value-keyword-case rule

Bilie:

#2424

No, it's self explanatory.

### DIFF
--- a/lib/rules/value-keyword-case/README.md
+++ b/lib/rules/value-keyword-case/README.md
@@ -164,3 +164,52 @@ a {
 }
 ```
 
+### `ignoreProperties: ["/regex/", "non-regex"]`
+
+Ignore case of the values of the listed properties.
+
+For example, with `"lower"`.
+
+```js
+["display"]
+```
+
+The following patterns are considered violations:
+
+```css
+a {
+  text-align: LEFT;
+}
+```
+
+```css
+a {
+  text-align: Left;
+}
+```
+
+The following patterns are *not* considered violations:
+
+```css
+a {
+  display: bloCk;
+},
+```
+
+```css
+a {
+  display: BloCk;
+}
+```
+
+```css
+a {
+  display: BLOCK;
+}
+```
+
+```css
+a {
+  display: block;
+}
+```
--- a/lib/rules/value-keyword-case/README.md
+++ b/lib/rules/value-keyword-case/README.md
@@ -171,7 +171,7 @@ Ignore case of the values of the listed properties.
 For example, with `"lower"`.
 
 ```js
-["display"]
+["/^(b|B)ackground$/", "display"]
 ```
 
 The following patterns are considered violations:
@@ -212,4 +212,16 @@ a {
 a {
   display: block;
 }
+```
+
+```css
+  div {
+    background: Red;
+  }
+```
+
+```css
+  div {
+    Background: deepPink;
+  }
 ```
--- a/lib/rules/value-keyword-case/README.md
+++ b/lib/rules/value-keyword-case/README.md
@@ -193,7 +193,7 @@ The following patterns are *not* considered violations:
 ```css
 a {
   display: bloCk;
-},
+}
 ```
 
 ```css
@@ -215,13 +215,13 @@ a {
 ```
 
 ```css
-  div {
-    background: Red;
-  }
+a {
+  background: Red;
+}
 ```
 
 ```css
-  div {
-    Background: deepPink;
-  }
+a {
+  Background: deepPink;
+}
 ```
--- a/lib/rules/value-keyword-case/__tests__/index.js
+++ b/lib/rules/value-keyword-case/__tests__/index.js
@@ -1998,10 +1998,10 @@ testRule(rule, {
       code: "a { display: block; }"
     },
     {
-      code: "div { background: Red; }"
+      code: "a { background: Red; }"
     },
     {
-      code: "div { Background: deepPink; }"
+      code: "a { Background: deepPink; }"
     }
   ],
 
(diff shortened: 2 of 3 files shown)


lib/rules/value-keyword-case/index.js
```diff
@@ -164,6 +166,10 @@ const rule = function(expectation, options) {
           return;
         }
 
+        if (ignoreProperties.length > 0 && ignoreProperties.includes(prop)) {
```
jeddy3: Please use the `matchesStringOrRegExp` util instead of `.includes`, like we do on [this line](https://github.com/Bilie/stylelint/blob/140f09771b9eb26538537e75469a850313ea617b/lib/rules/value-keyword-case/index.js#L164). This will:

1. Allow the option to accept regular expressions.
2. Address the failing [node@4 tests](https://travis-ci.org/stylelint/stylelint/builds/282262478?utm_source=github_status&utm_medium=notification), as node@4 doesn't support `.includes`.

lib/rules/value-keyword-case/__tests__/index.js
```diff
@@ -1938,3 +1938,79 @@ testRule(rule, {
     }
   ]
 });
+
+testRule(rule, {
```
jeddy3: Please include a couple of regular expression tests, like [we do](https://github.com/Bilie/stylelint/blob/140f09771b9eb26538537e75469a850313ea617b/lib/rules/value-keyword-case/__tests__/index.js#L1877-L1940) for the `ignoreKeywords` option.

lib/rules/value-keyword-case/README.md
```diff
@@ -164,3 +164,52 @@ a {
 }
 ```
 
+### `ignoreProperties: ["/regex/", "non-regex"]`
+
+Ignore case of the values of the listed properties.
+
+For example, with `"lower"`.
+
+```js
```
jeddy3: Please include some regular expression examples [we do here](https://github.com/Bilie/stylelint/tree/140f09771b9eb26538537e75469a850313ea617b/lib/rules/value-keyword-case#ignorekeywords-regex-non-regex) for the `ignoreKeywords` option.

Also, can you please move the docs for this option above the `ignoreKeywords` option, as options should be listed alphabetically.

Bilie: Hi, thank you very much for your feedback! I was not sure if the new option should be called ignoreProperties or ignoreCustomProperties. Currently it is named ignoreProperties. Should it still go above ignoreKeywords in the readme or should I rename it to ignoreCustomProperties? Thank you!

jeddy3: It should be called `ignoreProperties`. _Custom properties_ are [something different entirely](https://www.w3.org/TR/css-variables-1/#defining-variables) in CSS. There might come a time when a user requests an `ignoreCustomProperties` option but that would be for the following example:

```js
"value-keyword-case": ["lower", { ignoreCustomProperties: ["brand-color"] }]
```

```css
:root {
  --brand-color: myMadeUPKeYwOrDColor;
}
```

However, for what you're trying to achieve in this PR, `ignoreProperties` is the correct name.

lib/rules/value-keyword-case/README.md
```diff
@@ -164,3 +164,64 @@ a {
...
+  text-align: Left;
+}
+```
+
+The following patterns are *not* considered violations:
+
+```css
+a {
+  display: bloCk;
+},
```
jeddy3: Extra comma

lib/rules/value-keyword-case/README.md
```diff
@@ -164,3 +164,64 @@ a {
...
+```
+
+```css
+a {
+  display: block;
+}
+```
+
+```css
+  div {
```
jeddy3: 1. Remove indent, please.
2. Use the `a` selector, as per [these guidelines](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#example-patterns).

lib/rules/value-keyword-case/README.md
```diff
@@ -164,3 +164,64 @@ a {
...
+    background: Red;
+  }
+```
+
+```css
```
jeddy3: 1. Remove indent, please.
2. Use the `a` selector, as per [these guidelines](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#example-patterns).

lib/rules/value-keyword-case/__tests__/index.js
```diff
@@ -1938,3 +1938,91 @@ testRule(rule, {
...
+      code: "a { display: BloCk; }"
+    },
+    {
+      code: "a { display: BLOCK; }"
+    },
+    {
+      code: "a { display: block; }"
+    },
+    {
+      code: "div { background: Red; }"
```
jeddy3: Use `a` rather than `div` for consistency please, otherwise a future maintainer might think the different selector is significant to the test when it is not.

lib/rules/value-keyword-case/__tests__/index.js
```diff
@@ -1938,3 +1938,91 @@ testRule(rule, {
...
+    },
+    {
+      code: "div { Background: deepPink; }"
```
jeddy3: Don't forget these last two `div`s please :)

lib/rules/value-keyword-case/README.md
jeddy3: > Also, can you please move the docs for this option above the ignoreKeywords option, as options should be listed alphabetically.

Whoops, I just realised my mistake here... `K` is before `P` in the alphabet... silly me. I can see why you were asking about custom properties now.

jeddy3:
> -   Added: `ignoreProperties` option to `value-keyword-case` ([#2937](https://github.com/stylelint/stylelint/pull/2937)).

jeddy3:
@Bilie Thanks again for your first-time contribution. Your new option was just released in `stylelint@8.2.0`

PR #2947, opened 2017-10-09
[#2829] add autofix for number-no-trailing-zeros

modosc:

#2829 

No, it's self explanatory

### DIFF
--- a/lib/rules/number-no-trailing-zeros/index.js
+++ b/lib/rules/number-no-trailing-zeros/index.js
@@ -13,7 +13,7 @@ const messages = ruleMessages(ruleName, {
   rejected: "Unexpected trailing zero(s)"
 });
 
-const rule = function(actual) {
+const rule = function(actual, secondary, context) {
   return (root, result) => {
     const validOptions = validateOptions(result, ruleName, { actual });
     if (!validOptions) {
@@ -31,6 +31,8 @@ const rule = function(actual) {
     root.walkDecls(decl => check(decl, decl.value, declarationValueIndex));
 
     function check(node, value, getIndex) {
+      const fixPositions = [];
+
       // Get out quickly if there are no periods
       if (value.indexOf(".") === -1) {
         return;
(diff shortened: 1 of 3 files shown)


lib/rules/number-no-trailing-zeros/index.js
```diff
@@ -50,28 +52,69 @@ const rule = function(actual) {
...
+          } else {
+            node.value = removeTrailingZeros(node.value, startIndex, endIndex);
+          }
+        });
+      }
     }
   };
 };
 
+function removeTrailingZeros(input, startIndex, endIndex) {
```
evilebottnawi: Maybe best place `utils`?

jeddy3: Is it used by more than rule? If so, then yes I agree. Otherwise, it's probably more helpful to maintainers to keep it here.

jeddy3:
@modosc Thanks again for this.

Please remember [to update](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/pull-requests.md) the CHANGELOG when merging a PR as it's easy for us to loose track.

I did this one for you in https://github.com/stylelint/stylelint/commit/2277cb717090828bfe022f1d954b1af0f132521d

> -   Added: `number-no-trailing-zeros` autofix ([#2947](https://github.com/stylelint/stylelint/issues/2947)).

modosc:
@jeddy3 does the changelog update become part of the pull request or happen after the merge?

modosc:
> After the merge (as it helps us avoid conflicts). I just realised that the guidelines aren't that clear. I'll tweak them now.

got it - it would be cool to use something to do this automatically, maybe https://skywinder.github.io/github-changelog-generator/ ?

jeddy3:
> got it - it would be cool to use something to do this automatically, maybe https://skywinder.github.io/github-changelog-generator/ ?

We experimented with automating it, but most things ended up being more trouble than they're worth. I think the current process has little overhead.

Issue #2875, opened 2017-09-10
Fix false positives for spaceless multiplication in calc functions in unit-*


CAYdenberg:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Bug, although corner-case. In `calc` functions, multiplication (and division) operations can be spaceless e.g. `calc(2px*3)`. postcss-value-parser however doesn't recognise `*` as a divider so it would assess the units as `px*3`. This is a problem for all `unit-*` rules.

> Which rule, if any, is this issue related to?

It was fixed in the case of `unit-no-unknown` in #2848. So unit-blacklist, unit-whitelist and unit-case. There may be others?

You can see #2551 for details - it is functionally the same.

I'm able to handle this one too as the fix is simple - but I'll leave this open for awhile first for discussion.

jeddy3:
SGTM.

CAYdenberg:
This was, afaik, handled in #2948

PR #2948, opened 2017-10-10
Handle spaceless multiplication in unit-* rules

CAYdenberg:

#2875 

Wrote failing tests but haven't implemented fixes yet.

### DIFF
--- a/lib/rules/unit-blacklist/__tests__/index.js
+++ b/lib/rules/unit-blacklist/__tests__/index.js
@@ -49,6 +49,9 @@ testRule(rule, {
       code: "a { top: calc(10em - 3em); }"
     },
     {
+      code: "a { top: calc(10em*2rem); }"
+    },
+    {
       code:
         "a { background-image: linear-gradient(to right, white calc(100% - 50em), silver); }"
     },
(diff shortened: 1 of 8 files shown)


jeddy3:
Tests look good.

jeddy3:
@CAYdenberg Is this ready for review now?

CAYdenberg:
Yes!

jeddy3:
Is there value in creating a `parseValue` util that houses this in one place, like we do for [`parseSelector`](https://github.com/stylelint/stylelint/blob/master/lib/utils/parseSelector.js)?

CAYdenberg:
I thought about that. We really want the postcss-value-parser to handle `*` as a separator, and it doesn't. So the work around is just to replace it, but we probably only want to make this replacement in unit rules for fear of breaking something else.

If I look at the four `unit-*` rules, most of the code is duplicated. I thought about creating some kind unit analyzer utility - maybe walkUnits? So that the individual rules could just recieve units as a callback when found, then check them. Or they could pass a "checking" function into the utility.

This would be a bigger change though, so I think this option is OK since it's just one additional line of code that's duplicated between all four files (although one that doesn't make sense without explanation/comment). I'll happily implement it though if you think it's warranted and have some feedback on how it should be designed.

jeddy3:
> This would be a bigger change though, so I think this option is OK 

Agreed. Thanks for the write-up too. I think we're OK to go with this, as we can always revisit this if we're so inclined and/or have more time on our hands.

jeddy3:
> -   Fixed: `unit-*` false positives for spaceless multiplication ([#2948](https://github.com/stylelint/stylelint/issues/2948)).

After two approvals, anyone can merge a PR... including the original PR author. So, feel free to merge if you noticed an unmerged PR when you check-in on stylelint. I just did this one as I generally catch-up on stylelint around this time of day.

PR #2959, opened 2017-10-12
Autofix for string-quotes rule

modosc:

#2829

No, it's self explanatory


modosc:
> There are many unintentional changes to system test 005, which lower test coverage. Revert them back, please.

oops - fixed.

modosc:
@hudochenkov is there anything else i need to fix here?

hudochenkov:
@modosc I see violations report by this rule are in quite strange order. It's visible in the system test 005 snapshot. Violations reported not from top to bottom and from left to right of a stylesheet, but in random order.

modosc:
> @modosc I see violations report by this rule are in quite strange order. It's visible in the system test 005 snapshot. Violations reported not from top to bottom and from left to right of a stylesheet, but in random order.

i think it's because i changed the parser from `style-search` to use `postcss`. there are now three separate passes - `root.walkRules`, `root.walkAtRules`, and `root.walkDecls`.  i copied this pattern from other places where we do separate `root.walkAtRules` and `root.walkDecls` passes - is it not appropriate here?

hudochenkov:
I think we should avoid unordered violations report. If other rules do it, then it's a subject to discuss and address inconsistency problem.

For current rule, I think it's good to use following instead of bunch `.walk*`:

```js
root.walk(node => {
	if (node.type === 'at-rule') {
		// ...
	}
	
	if (node.type === 'rule') {
		// ...
	}
	
	if (node.type === 'decl') {
		// ...
	}
});
```

modosc:
@hudochenkov parsing is refactored and the snapshot is the same as `master`

lib/rules/string-quotes/index.js
```diff
@@ -23,31 +28,117 @@ const rule = function(expectation) {
...
-        result,
-        ruleName
+    function checkDeclOrAtRule(node, value, getIndex) {
+      const fixPositions = [];
+      // Get out quickly if there are no erroneous quotes
+      if (value.indexOf(erroneousQuote) === -1) {
+        return;
+      } else if (node.type === "atrule" && node.name === "charset") {
+        // allow @charset rules to have double quotes, in spite of the configuration
+        return;
```
hudochenkov: There is an issue related this specific condition https://github.com/stylelint/stylelint/issues/2788

Maybe we should report it if `@charset` has single quotes, and replace them to double quotes with `--fix`? Or it should be a separate PR?

lib/rules/string-quotes/index.js
```diff
@@ -1,18 +1,23 @@
...
 const ruleName = "string-quotes";
 
 const messages = ruleMessages(ruleName, {
   expected: q => `Expected ${q} quotes`
 });
 
-const rule = function(expectation) {
+const rule = function(expectation, secondary, context) {
   const erroneousQuote = expectation === "single" ? '"' : "'";
+  const correctQuote = expectation === "single" ? "'" : '"';
```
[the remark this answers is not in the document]
modosc: sure.

lib/rules/string-quotes/index.js
modosc: i think a separate pr would make sense - this is a pretty specific case. i just tried to keep the original logic in my pr.

lib/rules/string-quotes/index.js
```diff
@@ -23,31 +28,117 @@ const rule = function(expectation) {
...
+        } else {
+          node.value = replaceQuote(node.value, fixIndex, correctQuote);
+        }
+      });
+    }
   };
 };
 
+function replaceQuote(string, index, replace) {
+  return string.substring(0, index) + replace + string.substring(index + 1);
```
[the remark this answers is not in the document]
modosc: makes sense.

hudochenkov:
Can I ask you to not rebase PRs, which were reviewed and need review again, please? After rebase it's harder to understand what was changed since the last review.

jwilsson:
Sorry for posting this so late, but how do we want to handle cases like this:

Config:
```json
"rules": {
  "string-quotes": "single"
}
```

Source:
```css
.foo::before {
    content: "foo isn't bar";
}
```

Fixed:
```css
.foo::before {
    content: 'foo isn't bar';
}
```

Today, the rule will warn about it but the fix as applied now will actually generate faulty CSS. It might be a bit of an edge case but I at least wanted to bring it up. 

Perhaps something like the `avoidEscape` option in ESLint's [quotes rule](https://eslint.org/docs/rules/quotes#avoidescape) is needed? That's of course something out of scope for this PR but I can create an issue for that.

modosc:
@jwilsson that’s a legitimate bug, i’ll fix it here.

modosc:
i feel like actually implementing the fix for this is more complicated than it seems so for now i copied eslint's `avoidEscape` behavior here. if `avoidEscape` is `true` (which is the default behavior) we won't report an error or try to fix it. if `avoidEscape` is `false` we report an error but don't attempt to fix it. i could be totally wrong in which case a followup pr to add the fix would be nice.

lib/rules/string-quotes/__tests__/index.js
```diff
@@ -168,6 +184,64 @@ testRule(rule, {
 
 testRule(rule, {
   ruleName,
+  config: ["single", { avoidEscape: false }],
+  skipBasicChecks: true,
+  reject: [
```
hudochenkov: Please add `fix: true` for all new tests.

hudochenkov: After some investigation and thinking, I withdraw my comment. `testRule()` designed the way it is on purpose. If a rule has autofix and if a code has a violation, then fixed code should be different than input code. It also prevents us from human error, but copy-pasting `code` to `fixed` and forgetting about it.

Please add a comment before these two `testRule` why there no `fix` set (fix grammar or rephrase if needed):

```js
// Styles which requires escaping isn't autofixed, it's just reported instead
```

modosc:
ok, docs / comments updated.

lib/rules/string-quotes/README.md
```diff
@@ -75,3 +76,63 @@ a { content: "x"; }
 ```css
 a[id="foo"] {}
 ```
+
+
+`avoidEscape`: `true|false`, defaults to `true`
```
jeddy3: Can this be a level 3 heading for consistency with other READMEs, please?

lib/rules/string-quotes/README.md
```diff
@@ -75,3 +76,63 @@ a { content: "x"; }
...
+
+Allows strings to use single-quotes or double-quotes so long as the string contains a quote that would have to be escaped otherwise.
+
+### `true`
```
jeddy3: I think just one set of examples with `false` will suffice.

They'll need to follow the following format (due to a limitation of how the example background color highlighting works on the website):

---

Allows strings to use single-quotes or double-quotes so long as the string contains a quote that would have to be escaped otherwise.

For example, with `"single", { "avoidEscape" : false }`.

The following patterns are considered violations:

```css
a { content: "x'y'z"; }
```

```css
a[id="foo'bar'baz"] {}
```

The following patterns are *not* considered violations:

```css
a { content: 'x'; }
```

```css
a[id='foo'] {}
```

---

lib/rules/string-quotes/README.md
jeddy3: As in 

```
## Optional secondary options
  
### `avoidEscape`: `true|false`, defaults to `true`

```

modosc:
Changelog:

>-   Added: `string-quotes` autofix ([#2959](https://github.com/stylelint/stylelint/pull/2959)).

Issue #2965, opened 2017-10-15
Reduce size of npm package


avaly:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

An improvement to the NPM package file size.

It's not common for devs to read the docs for a package from within `node_modules`.

> Which version of stylelint are you using?

`8.*`

jeddy3:
@avaly Thanks for the suggestion.

See comment in at https://github.com/stylelint/stylelint/pull/2966#issuecomment-336708089 about why the corresponding PR was closed.

I'm going to keep this issue open so we can gauge what priority reducing the npm package should be.

ntwb:
I'm seeing `146,167 bytes (205 KB on disk) for 27 items` for the `/docs` folder via "Get Info" on Mac OS, that compresses down to ~50kb, not an insignificant size IMHO.

It might be worthwhile exploring alternative packaging options for stylelint, I'm thinking [Lerna](https://lernajs.io/) might be an option where we can keep the stylelint repo _as is_ for the most part but publish the docs as a separate npm package. This would allow for little change to the current workflow for the vast majority of contributors including @stylelint/core and a minor workflow change to publishing new releases, instead of one package there would be two published to npm, the second being `stylelint-docs` via new `package.json` file in `/docs`. In my mind this would be a pretty trivial change, happy to explore further and whip up a proof-of-concept pull request.

jeddy3:
> It might be worthwhile exploring alternative packaging options for stylelint

Agreed.

Perhaps another option, for now, is to simply assume the main stylelint repo will be placed alongside the stylelint.io repo on our local machines. We can then build the website from that. We're already assuming that the docs are always going to be in the same place in the package, so why not carry this assumption a little further?. This approach avoids having to add another tool and an additional publishing step. Also, not all the docs are within the `/docs/` folder e.g. all the rule READMEs, which could complicate the publishing step. If we're building from the main repo, then we could exclude all `.md` files from the package... reducing the size further.

It should just be the case of replacing `node_modules` with `..` in [`scripts/copy-stylelint-docs.js`](https://github.com/stylelint/stylelint.io/blob/master/scripts/copy-stylelint-docs.js) to make this work.

> In my mind this would be a pretty trivial change, happy to explore further and whip up a proof-of-concept pull request.

Having said all that, if you feel your approach has advantages then please do whip up a PoC :)

ntwb:
Closing this in favour of #2454

PR #2966, opened 2017-10-15
Remove docs from npm package

avaly:

#2965

No, it's self explanatory.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
     "bin",
     "CONTRIBUTING.md",
     "decls/*.js",
-    "docs",
     "lib",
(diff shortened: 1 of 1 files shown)


jeddy3:
@avaly Thanks for the suggestion. At the moment the website (http://stylelint.io) is [built using the docs published in the package](https://github.com/stylelint/stylelint.io). It was done this way as, at the time, it was deemed to be the quickest and easier to get the website up and running. I'm not sure the decrease in package size is worth the effort required to rebuild how we update and publish the website. However, we can always revisit this if the process of updating the website is ever revised.

Issue #2586, opened 2017-05-19
Add ignore: ["custom-properties"] to length-zero-no-unit


jethrolarson:
A bug where using stylefmt with stylelint causes calc expressions to break

> Which rule, if any, is this issue related to?

length-zero-no-unit 

> What CSS is needed to reproduce this issue?

```css
.foo {
  --x: 0px;
  left: calc(300px - var(--x));
}
```

> What stylelint configuration is needed to reproduce this issue?

```yaml
extends: stylelint-config-standard
```

> Which version of stylelint are you using?

7.10.1

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylefmt -r 'app/**/*.scss' 'app/**/*.css'`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)? 
no

> What did you expect to happen?

Fixing the warning to be safe since it's in the recommended list

> What actually happened (e.g. what warnings or errors you are getting)?

Styles get fixed to
```css
.foo {
  --x: 0;
  left: calc(300px - var(--x));
}
```
calc sees different units on left and right hand side of the operator and returns a 0.

I think that this rule should be removed from the recommended list since it is not safe to fix.

I'm filing this with stylefmt as well.

hudochenkov:
Thank you for reporting. This use case will be helpful for us, if we're going to implement fix functionality for this rule.

But currently stylelint doesn't have fixing functionality. Please, report this problem to stylefmt repository.

hudochenkov:
On the other hand, stylelint will show violation if `length-zero-no-unit` is enabled:

```css
.foo {
  --x: 0px;
  left: calc(300px - var(--x));
}
```

@stylelint/core do you think we should ignore custom properties in this rule? Or it's rare case and user could disable a rule for required line?

evilebottnawi:
@hudochenkov from spec:
```
For example, the following code incorrectly attempts to use a variable as a property name:
.foo {
  --side: margin-top;
  var(--side): 20px;
}
This is not equivalent to setting margin-top: 20px;. Instead, the second declaration is simply thrown away as a syntax error for having an invalid property name.

Similarly, you can’t build up a single token where part of it is provided by a variable:

.foo {
  --gap: 20;
  margin-top: var(--gap)px;
}
[...]

PR #2967, opened 2017-10-15
Add ignore: ["custom-properties"] to length-zero-no-unit

lutien:

#2586

I`ve tried to follow all tips in the issue.

### DIFF
--- a/lib/rules/length-zero-no-unit/README.md
+++ b/lib/rules/length-zero-no-unit/README.md
@@ -52,7 +52,7 @@ a { top: 1.001vh }
 
 Ignore units for zero length in custom properties.
 
-The following patterns are *not* considered violations:
+The following pattern is *not* considered a violation:
 
 ```css
 a { --x: 0px; }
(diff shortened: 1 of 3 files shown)


lib/rules/length-zero-no-unit/README.md
```diff
@@ -43,3 +43,17 @@ a { top: 2in; }
...
+
+## Optional secondary options
+
+### `ignore: ["custom-properties"]`
+
+#### `"custom-properties"`
+
+Ignore units for zero length in custom properties.
+
+The following patterns are *not* considered violations:
```
jeddy3:
```
The following pattern is *not* considered a violation:
```

lutien: Yes, true. Now it is fixed

jeddy3:
> -   Added: `ignore: ["custom-properties"]` option to `length-zero-no-unit` ([#2967](https://github.com/stylelint/stylelint/pull/2967)).

Issue #2974, opened 2017-10-18
\u doesn't support 10000+


Aladdin-ADD:
https://github.com/stylelint/stylelint/blob/adb9158b9d928017a4c5ae92b367482f0a72b026/lib/rules/font-family-name-quotes/index.js#L40

hudochenkov:
Thanks for creating this issue but we are closing it as issues need to follow our issue template, so that we can clearly understand your particular circumstances.

Please help us to help you by [recreating the issue](https://github.com/stylelint/stylelint/issues/new) using the template.

Aladdin-ADD:
aha, not a English speaker~(✿◕‿◕✿)

in short, `\u` doesn't support 10000+, so I guess here is a bug.
![image](https://user-images.githubusercontent.com/13050025/31741746-f3221bea-b487-11e7-8039-3812f6729e0b.png)

refs:

Aladdin-ADD:
not a English speaker d=====(￣▽￣*)b
in short `\u` doesn't support 10000+, so I guess here is a bug..

![image](https://user-images.githubusercontent.com/13050025/31741899-85a06f9e-b488-11e7-9018-597ccd6dd70d.png)

jeddy3:
@Aladdin-ADD Can you submit a Pull Request with failing tests, please?

This might help us overcome the language barrier.

PR #2977, opened 2017-10-19
Fix high-range unicode characters in font-family-name-quotes

Aladdin-ADD:
\u doesn't support 10000+
e.g.
```js
/[\u0000-\u10FFFF]/.test("\u1100");         // false
/[\u{0000}-\u{10FFFF}]/u.test("\u1100")  // true
```

#2974

> Is there anything in the PR that needs reviewer focus?

u flag only supported in node.js v6+.
image

### DIFF
--- a/lib/rules/font-family-name-quotes/__tests__/index.js
+++ b/lib/rules/font-family-name-quotes/__tests__/index.js
@@ -85,6 +85,12 @@ testRule(rule, {
     },
     {
       code: 'a { font: italic 300 16px/30px "Arial", serif; }'
+    },
+    {
+      code: 'a { font-family: "\u1100"; }'
+    },
+    {
+      code: 'a { font-family: "ሀ"; }'
     }
   ]),
 
@@ -150,6 +156,18 @@ testRule(rule, {
       message: messages.expected("Arial"),
       line: 1,
       column: 32
+    },
+    {
+      code: "a { font-family: \u1100; }",
+      message: messages.expected("\u1100"),
+      line: 1,
+      column: 18
+    },
+    {
+      code: "a { font-family: ሀ; }",
+      message: messages.expected("ሀ"),
+      line: 1,
+      column: 18
     }
   ]
 });
--- a/lib/rules/font-family-name-quotes/index.js
+++ b/lib/rules/font-family-name-quotes/index.js
@@ -37,7 +37,8 @@ function quotesRecommended(family) {
 function quotesRequired(family) {
   return family.split(/\s+/).some(word => {
     return (
-      /^(-?\d|--)/.test(word) || !/^[-_a-zA-Z0-9\u00A0-\u10FFFF]+$/.test(word)
+      /^(-?\d|--)/.test(word) ||
+      !/^[-_a-zA-Z0-9\u{00A0}-\u{10FFFF}]+$/u.test(word)
     );
   });
 }


jeddy3:
@Aladdin-ADD Thanks. 

It would be very helpful if you could add some tests to [`lib/rules/font-family-name-quotes/__tests__/index.js`](https://github.com/stylelint/stylelint/blob/master/lib/rules/font-family-name-quotes/__tests__/index.js) so we can see what this change fixes.

> u flag only supported in node.js v6+.

This is a problem as stylelint supports node@4 without transpiling. Is there another solution?

jeddy3:
@evilebottnawi Did you mean to approve this PR?

evilebottnawi:
@jeddy3 yes, but need tests and blocked before we drop node 4

hudochenkov:
Regardless of two approvals do not merge this PR yet.

jeddy3:
> Regardless of two approvals do not merge this PR yet.

Agreed. Please don't approve PRs until the fix or feature completely ready to go in i.e. has tests and documentation.

Arcanemagus:
>>u flag only supported in node.js v6+.
>
>This is a problem as stylelint supports node@4 without transpiling. Is there another solution?

You could maybe use [`String.prototype.codePointAt()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt)? Support for that was added in Node.js v4.

Aladdin-ADD:
as a workaround, we can use babel to precompile the regex. But I prefer using the flag u when dropping supporting node.js v4.

ntwb:
> I think it's worthy to add eslint-plugin-node in eslint-config, there are some very useful rules -- e.g. unsupported /u flag can be detected by eslint

Thanks for the suggestion @Aladdin-ADD, I've created #2978 to explore any benefits in adding `eslint-plugin-node`

Aladdin-ADD:
I'm not very familiar with this rule, could someone give a advice to add tests,~ o(>﹏<)o

jeddy3:
> I'm not very familiar with this rule, could someone give a advice to add tests? ~ o(>﹏<)o

[Guide to fixing bugs](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#fixing-a-bug-in-an-existing-rule).

> Write failing unit tests that exemplify the bug.

1. Duplicate [these lines](https://github.com/stylelint/stylelint/blob/master/lib/rules/font-family-name-quotes/__tests__/index.js#L83-L85) of an "accept" test (i.e. code that shouldn't trigger any violations)
2. Change the `code` to illustrate the bug that the changes in this PR fix will suffice. 
3. If applicable, also [add a reject test](https://github.com/stylelint/stylelint/blob/master/lib/rules/font-family-name-quotes/__tests__/index.js#L92-L96) i.e. code that should trigger a single violation by the linter.

Aladdin-ADD:
@jeddy3 thanks for your feedback!  will update asap.

Aladdin-ADD:
updated. review needed!

jeddy3:
> `code: 'a { font-family: "\u1100"; }'`

Interesting. Do people write font family names like that? Or does that represent a character in some alphabet? If so, is it possible to use some actual characters?

Aladdin-ADD:
@jeddy3 it's a little rough to find a real family name having a character in `\10ff` ~ `\10ffff`, I just added an example to repro. thoughts?

jeddy3:
Does anyone have a mo for 2nd review on this?

Arcanemagus:
Should this be marked as blocked on #3075? After that gets merged the `u` RegEx flag will be available and this will be _greatly_ simplified.

hudochenkov:
@Arcanemagus good call!

ntwb:
@Arcanemagus I/we would still have no issue adding you to the @stylelint/core team, with no expectations of you contributing any more than you already do other than to say you could triage issues (or reopen etc) as you see fit when/if the need arises 😄

Arcanemagus:
@ntwb I'm not _opposed_ to it, as I said before though occasionally checking through issues is all I really have time for here though 😉.

ntwb:
That's more than fine @Arcanemagus, your thoughts, opinions, and contributions to stylelint, the stylelint repos and https://github.com/AtomLinter/linter-stylelint are really appreciated and of great value and significance to the @stylelint community, thank YOU 💯 

@jeddy3, can you please add @Arcanemagus to the @stylelint/core team please :)

jeddy3:
Done

ntwb:
Woot 🎉

ntwb:
As Node.js 4.x was dropped in https://github.com/stylelint/stylelint/commit/bcb46b75e9ca3ed98eb68fbedd544239ad52cdf1, we can now us the `u` regex flag, @Aladdin-ADD can you update the PR and we'll get this into v9.0.0 please?

Aladdin-ADD:
@ntwb updated, and rebased the commits. happy to see its landing. 😄

hudochenkov:
> -   Fixed: Increased unicode range in `font-family-name-quotes` ([#2974](https://github.com/stylelint/stylelint/pull/2974)).

PR #2981, opened 2017-10-20
Document built-in rules and non-standard syntax

### LINKED ISSUE #2979: Document the built-in rules' emphasis on standard syntax better

If a complex selector contains an interpolated SASS variable stylelint does not see a type selector inside it and doesn't report it.

`selector-max-type`

```css
#{$var} span {
    width:inherit;
}
```

where $var can be anything (class, another type selector...). In the last case, the type selector contained in the variable is not reported as well.

```json
{
  "rules": {
    "selector-max-type": 0
  }
}
```

`8.2.0`

`gulp-stylelint`

Yes, it's related to SASS variable interpolation.

An error should be flagged.

Nothing was reported. If you make the type selector to be a part of selector that does not contain a variable (or just use it alone) it does report.

jeddy3:
Closes #2979 

I've added links to the relevant section of the documentation for most of the "Features" item. I believe this can help users dive into the docs and help them better comprehend what stylelint is and isn't.


jeddy3:
@hudochenkov Yes, you're right. I'm now also wondering whether linking into sections of the user-guide is more confusing for the user then pointing them to the index of the user-guide (like we already do). I'll strip out the linking from this PR.

README.md
```diff
@@ -6,19 +6,19 @@ A mighty, modern CSS linter and fixer that helps you avoid errors and enforce co
 
 ## Features
 
--   **Over one hundred and sixty built-in rules.** Including ones that:
+-   **Over one hundred and sixty built-in rules:** Geared towards standard CSS syntax, these can:
```
hudochenkov: Two colons in “one” sentence. Could you change it, please?

README.md
```diff
@@ -6,19 +6,19 @@ A mighty, modern CSS linter and fixer that helps you avoid errors and enforce co
...
     -   **Catch possible errors:** e.g., invalid standard CSS syntax, duplicates, and overrides.
     -   **Limit language features:** e.g.:
         -   Disallow specific units, properties, functions, and at-rules.
         -   Limit the specificity and quantity of selectors.
         -   Enforce patterns for selectors and custom properties.
     -   **Enforce stylistic conventions:** e.g., whitespace, case, notation, and quotes.
 -   **Understands the latest CSS syntax:** Including custom properties, `calc()` and level 4 selectors.
--   **Completely unopinionated:** Only enable the rules you want, and configure them with options that tailor the linter to your needs.
+-   **Completely unopinionated:** Only enable the rules] you want, and configure them with options that tailor the linter to your needs.
```
hudochenkov: Typo.

jeddy3:
@hudochenkov Done. Thanks for close inspection :)

README.md
```diff
@@ -18,7 +18,7 @@ A mighty, modern CSS linter and fixer that helps you avoid errors and enforce co
 -   **Support for plugins:** It's easy to create your own rules and add them to the linter, or make use of plugins written by the community.
 -   **Automatically fixes some stylistic violations:** Save time by having stylelint fix your code with this *experimental* feature.
 -   **Shareable configs:** If you don't want to craft your own config, you can extend a shareable config. Or you can craft your own config and share with your team and/or the community.
--   **Supports *CSS-like* syntaxes:** The linter is powered by [PostCSS](https://github.com/postcss/postcss), so it can be configured and extended to understand any syntax that PostCSS can parse, including SCSS, [SugarSS](https://github.com/postcss/sugarss), and Less.
+-   **Parses *CSS-like* syntaxes:** The linter is powered by [PostCSS](https://github.com/postcss/postcss), so it can be configured and extended (e.g. [stylelint-scss](https://github.com/kristerkari/stylelint-scss)) to understand any syntax that PostCSS can parse, including SCSS, [SugarSS](https://github.com/postcss/sugarss), and Less.
```
hudochenkov: I'm not sure `stylelint-scss` should be there. It doesn't parse style sheet, it provides rules for non-standard syntaxes.

jeddy3: Fair point

PR #2984, opened 2017-10-23
Upgrade to postcss-selector-parser 3.1.0

chriseppstein:
This is a courtesy PR that updates to postcss-selector-parser 3.0 which is just released.

This addresses the issue for the upgrade originally created here:
https://github.com/stylelint/stylelint/issues/2757

The problems with attribute operators are fixed.

Two test cases were updated, the assertions for the column number were actually invalid to begin with and are now correct in this release.

There's still two failing tests, but as noted in this comment:

https://github.com/stylelint/stylelint/issues/2757#issuecomment-331698623

The syntax is non-standard and postcss-selector-parser does not plan to handle it, the suggested  workarounds in that comment seem reasonable.

### DIFF
--- a/lib/rules/selector-attribute-operator-blacklist/index.js
+++ b/lib/rules/selector-attribute-operator-blacklist/index.js
@@ -46,7 +46,8 @@ const rule = function(blacklistInput) {
           report({
             message: messages.rejected(operator),
             node: rule,
-            index: attributeNode.attribute.length + 1,
+            index:
+              attributeNode.sourceIndex + attributeNode.offsetOf("operator"),
             result,
             ruleName
           });
--- a/lib/rules/selector-attribute-operator-whitelist/index.js
+++ b/lib/rules/selector-attribute-operator-whitelist/index.js
@@ -46,7 +46,8 @@ const rule = function(whitelistInput) {
           report({
             message: messages.rejected(operator),
             node: rule,
-            index: attributeNode.attribute.length + 1,
+            index:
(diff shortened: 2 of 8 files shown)


jeddy3:
@chriseppstein Thanks! It's fantastic to see the problems with attribute operators fixed.

> The syntax is non-standard and postcss-selector-parser does not plan to handle it, the suggested workarounds in that comment seem reasonable.

Do you fancy porting those over from #2828 to your PR?

- [updated tests](https://github.com/stylelint/stylelint/pull/2828/files#diff-0d62e4ecc1a04792e4007e38975a295d).
- [using the `isStandardSyntaxSelector` util in `selector-attribute-quotes` to ignore non-standard syntax](https://github.com/stylelint/stylelint/pull/2828/files#diff-0cef3a06685092ae8029c334ae96ca5c).

Once this PR is green across the board, I think we'll be able to merge it.

chriseppstein:
@jeddy3 I've ported over those changes and also fixed a number of issues with how comments and spaces are handled in other parts of the attributes -- there was a number of latent bugs that even this extensive test suite didn't catch (as noted in https://github.com/postcss/postcss-selector-parser/issues/121). I also added some convenience methods for getting access to the offset location of attribute nodes discrete parts for error reporting -- the changes I made, made calculating that ... nontrivial.

All the tests are green now. I've already burned a lot of time on this that I didn't really have. If there's any further small issues, it would be great if those could be addressed by someone else.

PR #2988, opened 2017-10-25
Updated outdated cli doc.

etienne-martin:
The doc said that the cli --fix option does not respect special comments for disabling stylelint within sources. This is no longer true.


hudochenkov:
Thank you for your PR. But it still true, nothing has changed.

Issue #2090, opened 2016-11-21
Add except: ["after-closing-brace"] to block-closing-brace-empty-line-before


thomasjbradley:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Possibly a bug, but most likely a feature request.

I'd like to be able to force an empty line at the end of a media query—we can do it at the top of a media query, but I can't seem to find how to do it at the bottom.

```css
@media print {

  a {
    color: red;
  }
            /* <----- Force this empty line here */
}
```

> Which rule, if any, is this issue related to?

I see the rule: `rule-nested-empty-line-before` which will force a new line at the top of the media query.

And I see the rule: `block-closing-brace-empty-line-before` which will force a new line at the end of the block, but that would also force a new line after `color: red`.

> What CSS is needed to reproduce this issue?

*This CSS should pass:*

```css
@media print {

  a {
    color: red;
  }

}
```

*This CSS should fail:*

```css
@media print {

  a {
    color: red;
  }
}
```

> Which version of stylelint are you using?

Currently using `7.5.0`.

> How are you running stylelint: CLI, PostCSS plugin, Node API?

Node API.

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Nope.

> What did you expect to happen?

I'd like to get a warning message if there isn’t an empty line at the end of a media query.

> What actually happened (e.g. what warnings or errors you are getting)?

Both empty lines and non empty lines pass Stylelint tests.

I feel like an older version used to enforce this, but looking through past commits of my Stylelint config I don't see anything that jumps out at me as being “the rule” that enforced that requirement.

ntwb:
Thanks for using our issue template and the detailed description @thomasjbradley 👍 

p.s. Originally via [this](https://stackoverflow.com/questions/40593334/stylelint-force-empty-line-at-bottom-of-nested-block) Stack Overflow issue.

jeddy3:
I think we can achieve this with the addition of an `exceptAtRules: []` option to `block-closing-brace-empty-line-before`.

[...]

PR #3011, opened 2017-11-06
Add except: ["after-closing-brace"] to block-closing-brace-empty-line…

thomasjbradley:
Add `except: ["after-closing-brace"]` to `block-closing-brace-empty-line-before`

When `"never"` is specified for `block-closing-brace-empty-line-before` the option will be reversed when nested in at-rules.

#2090 

No, I don’t believe so.

### DIFF
--- a/lib/rules/block-closing-brace-empty-line-before/README.md
+++ b/lib/rules/block-closing-brace-empty-line-before/README.md
@@ -65,7 +65,7 @@ a { color: pink; }
 
 ### `except: ["after-closing-brace"]`
 
-When a rule is nested, and `"never"` is specified for empty lines before closing braces, force a new line before the closing brace of the nested block.
+When a rule is nested, `after-closing-brace` brace will reverse the primary option.
 
 For example, with `"never"` and `except: ["after-closing-brace"]`:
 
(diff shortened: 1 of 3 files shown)


lib/rules/block-closing-brace-empty-line-before/README.md
```diff
@@ -60,3 +60,79 @@ a {
 ```css
 a { color: pink; }
 ```
+
+
+### `except: ["after-closing-brace"]`
+
+When a rule is nested, and `"never"` is specified for empty lines before closing braces, force a new line before the closing brace of the nested block.
+
```
CAYdenberg: First off, I apologize for the delay in doing this review. Life. Thanks for creating your contribution.

The precedent for the use of "except" in secondary options is to **reverse** the primary option when the excepted case is satisfied. For example see https://github.com/stylelint/stylelint/tree/master/lib/rules/at-rule-empty-line-before#except-after-same-name-inside-block-blockless-after-same-name-blockless-blockless-after-blockless-first-nested and https://github.com/stylelint/stylelint/tree/master/lib/rules/rule-empty-line-before

If I read this correctly it gets half-way there: it reverses the rule when it is "never" but it doesn't do anything when it is "always".

CAYdenberg:
LGTM. Thanks again.

jeddy3:
> -   Added: `except: ["after-closing-brace"]` to `block-closing-brace-empty-line-before` ([#3011](https://github.com/stylelint/stylelint/pull/3011)).

Issue #3009, opened 2017-11-06
Add ignoreValues: [] to value-no-vendor-prefix


jeddy3:
Ref: https://github.com/stylelint/stylelint/issues/3007

There's [a section in the Developer Guide](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#adding-an-option-to-an-existing-rule) on how to add options if someone would like to pick this up.

Bebersohl:
I'll do this one

PR #3015, opened 2017-11-08
Add ignoreValues: [] to value-no-vendor-prefix

Bebersohl:

#3009

No, I don't think so.

Let me know what you think.

### DIFF
--- a/lib/rules/value-no-vendor-prefix/README.md
+++ b/lib/rules/value-no-vendor-prefix/README.md
@@ -41,3 +41,23 @@ a { max-width: max-content; }
 ```css
 a { background: linear-gradient(bottom, #000, #fff); }
 ```
+
+## Optional secondary options
+
+### `ignoreValues: ["-webkit-", "-moz-", "-ms-", "-o-"]`
+
+Given:
+
+```js
+["-webkit-", "-moz-"]
+```
+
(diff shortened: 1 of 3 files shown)


lib/rules/value-no-vendor-prefix/README.md
```diff
@@ -41,3 +41,23 @@ a { max-width: max-content; }
...
 ```
+
+
+### `ignoreValues: ["-webkit-", "-moz-", "-ms-", "-o-"]`
+
+Given:
+
+```js
+["-webkit-", "-moz-"]
```
sindresorhus: The intention of the option is to ignore properties, not vendor prefixes. So here the user might specify:

```js
["grab", "max-content"]
```

To have `-webkit-grab` and `-moz-max-content` ignored from the rule.

jeddy3:
I believe using the `optionsMatches` util will allows regex (within strings) to be accepted as well, see:

- https://github.com/stylelint/stylelint/blob/master/lib/rules/at-rule-no-unknown/index.js#L39-L42 -> using the util
- https://github.com/stylelint/stylelint/blob/master/lib/rules/declaration-block-no-shorthand-property-overrides/index.js#L36 -> getting unprefixed value

PR #3021, opened 2017-11-12
Fix isPathIgnored regression for .stylelintignore files

CAYdenberg:

#2833 

I have an appropriate test, but I cannot get the implementation. Would appreciate some help from either @stylelint/core or @billybonks on resolving relative/absolute filepaths as I feel like I've tried everything and I'm just going in circles.

The basic approach is moving building the ignorer from standalone.js to createLinter.js, and assigning it as a property of stylelint (main object). We can then call it both where it's needed within standalone as well as isPathIgnored.

### DIFF
--- a/lib/__tests__/createLinter.test.js
+++ b/lib/__tests__/createLinter.test.js
@@ -43,3 +43,21 @@ it("createLinter().isPathIgnored", () => {
     expect(results).toEqual([true, true, false, false]);
   });
 });
+
+it("createLinter().isPathIgnored with a .stylelintignore file", () => {
+  const config = {
+    rules: { "block-no-empty": true }
+  };
+  const linter = stylelint.createLinter({
+    config,
+    ignorePath: "./stylelintignore-test/.stylelintignore"
+  });
+  return Promise.all([
+    linter.isPathIgnored("a.css"),
+    linter.isPathIgnored("foo/bar/baz.css"),
+    linter.isPathIgnored("foo/bar/baz.scss"),
+    linter.isPathIgnored("foo/invalid-hex.css")
+  ]).then(results => {
+    expect(results).toEqual([true, true, false, false]);
+  });
+});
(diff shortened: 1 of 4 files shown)


CAYdenberg:
Some progess on this today although nothing to commit unfortunately.

This feature worked in 9578c48, but a couple of lines of code were later deleted from isPathIgnored, as well as a substantial amount of code from augmentConfig that added `ignorePatterns` property to `config` to be consumed here. 

I'm not yet sure why that change was made.

These are mostly implementation notes for myself unless someone wants to jump in ...

CAYdenberg:
Bug likely introduced here: https://github.com/stylelint/stylelint/pull/2464

Basic issue is - I believe - that `ignorer` reports relative paths, and so stylelint (both isPathIgnored and createLinter) understands these as being relative to process.cwd. But the _expectation_ is that the paths are relative .stylelintignore.

Right now, I don't see any good way of dealing with this.

Arcanemagus:
> I don't know if anyone besides @Arcanemagus actually knows it exists.

Probably not 😛.

If I'm understanding the correctly, setting `process.cwd` to the directory of the `.stylelintignore` will work around this issue from what you are saying?

jeddy3:
Is this PR still valid?

billybonks:
@jeddy3 so happens i was looking into my addon that wraps this. it definately is still a problem.

@CAYdenberg but a little nugget of information, is that if i add a path to my ignore file it actually stylelint stand alone returns 

```
log: '', // this is null  that is the interesting part if i remove from the ignore file this value is populated
  source: 'styles/bootstrap/mixins/_tag.scss',
  deprecations: [],
  invalidOptionWarnings: [],
  parseErrors: [],
  warnings: // and should not have generated warnings.
   [ { line: 8,
       column: 38,
       rule: 'function-comma-space-after',
       severity: 'error',
       text: 'Unexpected whitespace after "," in a single-line function (function-comma-space-after)' } ],
  ignored: undefined, //but this should be true.
```

CAYdenberg:
I never got this working. Basically struggling with the fact that ignorer doesn't report on absolute paths. I think the best path is to parse .stylelintignore and add all the paths just as they are added to the API, but I don't have time to deal with that right now.

PR #3023, opened 2017-11-15
Remove incorrect non-violation

squalrus:
I believe there was an incorrect example of a non-violation for `"never-multi-line"`


squalrus:
Nope, nevermind, the examples are trick for this rule but correct! 😄

Issue #3006, opened 2017-11-06
Add ignoreFeatureNames: {} to unit-blacklist


zaneriley:
Hi team,

Stylelint's made a huge impact on my team, but I'm working to clear up some of the annoyances. Currently, `unit-blacklist` is flagging media queries. Am I doing something wrong here?

> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Unit blacklist isn't ignoring media queries that use `px` unit.

> Which rule, if any, is this issue related to?

`unit-blacklist`

> What CSS is needed to reproduce this issue?

```css
/* should be allowed */
@media screen and (min-width: 500px) {
  
  .my-class {
    /* should be blacklisted */
    min-width: 500px; 
  }
}
```

> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
  "rules": {
    "unit-blacklist": [
      "px",
      {
        message:
          "Unexpected PX unit. Please use relative unit. (unit-blacklist)",
        ignoreProperties: {
          px: ["/^border/", "/^transform/", "/^media/", "/^@media/"]
        }
      }
    ],
  }
}
```

> Which version of stylelint are you using?

e.g. `^8.2.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

e.g. "CLI with `stylelint **/*.css`"

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No. Though I can confirm it happens in `.less`, `.modules.css` as well as vanilla css.

> What did you expect to happen?

`px` unit within `@media` rules to be ignored.

> What actually happened (e.g. what warnings or errors you are getting)?

e.g. "The following warnings were flagged:"

```shell
test.css
 208:31  ✖  Unexpected PX unit. Please use relative unit.               unit-blacklist
 211:16  ✖  Unexpected PX unit. Please use relative unit.               unit-blacklist
```

jeddy3:
@zaneriley Thanks for the issue and for using the template.

> Stylelint's made a huge impact on my team

That's great to hear.

>  but I'm working to clear up some of the annoyances. 

Good stuff. Working out the kinks benefits everyone.

> Unit blacklist isn't ignoring media queries that use px unit.

I believe this rule is behaving correctly. However, I think there's scope to add a new option to address your particular use case.

Right now, there is an optional secondary option to ignore units for specific _properties_ (via `ignoreProperties: {}`). 

[...]

PR #3027, opened 2017-11-18
Add ignoreMediaFeatureNames for unit-blacklist

lucmerceron:

#3006 

The PR is easy to understand.

NB:
I am not sure this solution is the most optimized since I have one loop for the value `valueParser` inside the one for the media `mediaParser` but it feels safer than doing matches only on the returned `mediaFeatureNode` and having to add `column` size for error handling.

Since this is my first PR for stylelint, this is widely open to discussion. I may have missed a util or anything else that could be used.

### DIFF
--- a/lib/rules/unit-blacklist/README.md
+++ b/lib/rules/unit-blacklist/README.md
@@ -113,27 +113,27 @@ Given:
 The following patterns are *not* considered violations:
 
 ```css
-@media (min-width: 960px) { }
+@media (min-width: 960px) {}
 ```
 
 ```css
-@media (max-height: 280px) { }
+@media (max-height: 280px) {}
 ```
 
 ```css
-@media not (resolution: 300dpi) { }
+@media not (resolution: 300dpi) {}
 ```
 
 The following patterns are considered violations:
 
 ```css
-a { @media screen and (max-device-width: 500px) { } }
+@media screen and (max-device-width: 500px) {}
 ```
 
 ```css
-@media all and (min-width: 500px) and (max-width: 200px) { }
+@media all and (min-width: 500px) and (max-width: 200px) {}
 ```
 
 ```css
-@media print and (max-resolution: 100dpi) { }
+@media print and (max-resolution: 100dpi) {}
 ```
--- a/lib/rules/unit-blacklist/README.md
+++ b/lib/rules/unit-blacklist/README.md
@@ -94,3 +94,46 @@ a { -moz-border-radius-topright: 40px; }
 ```css
 a { height: 100vmin; }
 ```
+
+### `ignoreMediaFeatureNames: { unit: ["property", "/regex/"] }`
+
+Ignore units for specific feature names.
+
+For example, with `["px", "dpi"]`.
+
+Given:
+
+```js
+{
+  "px": [ "min-width", "/height$/" ],
+  "dpi": [ "resolution" ]  
+}
+```
+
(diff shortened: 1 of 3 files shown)


lib/rules/unit-blacklist/index.js
```diff
@@ -31,18 +40,64 @@ const rule = function(blacklistInput, options) {
...
         }
       }
     );
     if (!validOptions) {
       return;
     }
 
-    function check(node, value, getIndex) {
+    function checkMedia(node, value, getIndex) {
+      value = value.replace(/\*/g, ",");
```
CAYdenberg: Let's leave this line out. Its sole purpose is handle parsing of calc functions - I don't think it's needed within media queries (and it's a bit of a hack so I'd like to not proliferate it if possible).

lib/rules/unit-blacklist/index.js
```diff
@@ -31,18 +40,64 @@ const rule = function(blacklistInput, options) {
...
+          // Ignore all non-word valueNode and
+          // the values not included in the parentValue string
+          if (
+            valueNode.type !== "word" ||
+            !parentValue.includes(valueNode.value)
+          ) {
+            return;
+          }
+
+          const unit = getUnitFromValueNode(valueNode);
```
CAYdenberg: From here to the end of the function is almost a mirror image of the same thing in walk.decls. I think it should be abstracted.

CAYdenberg:
Thanks. Tests and documentation looks good.

lib/rules/unit-blacklist/index.js
lucmerceron: Done in my last commit b6effdf

lib/rules/unit-blacklist/index.js
lucmerceron: Done in my last commit too b6effdf

CAYdenberg:
Awesome. Thanks again.

lib/rules/unit-blacklist/README.md
```diff
@@ -94,3 +94,46 @@ a { -moz-border-radius-topright: 40px; }
...
+```
+
+```css
+@media not (resolution: 300dpi) { }
+```
+
+The following patterns are considered violations:
+
+```css
+a { @media screen and (max-device-width: 500px) { } }
```
jeddy3: Can we just have `@media screen and (max-device-width: 500px) {}` here please as I don't believe nested media queries are standard syntax.

lib/rules/unit-blacklist/README.md
```diff
@@ -94,3 +94,46 @@ a { -moz-border-radius-topright: 40px; }
...
+{
+  "px": [ "min-width", "/height$/" ],
+  "dpi": [ "resolution" ]  
+}
+```
+
+The following patterns are *not* considered violations:
+
+```css
+@media (min-width: 960px) { }
```
jeddy3: Can you please change the examples to use empty braces i.e. `{}` as per [the guidelines](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#example-patterns).

lucmerceron: Yep, sorry about that. Done in my last [commit](6a0bc46).

lib/rules/unit-blacklist/README.md
lucmerceron: Done in my last [commit](6a0bc46)

Issue #3038, opened 2017-11-27
Fix invalid option warning for secondary option in font-family-no-missing-generic-family-keyword


subzey:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug. The rule `font-family-no-missing-generic-family-keyword` should have a way to specify secondary options, but currently it throws.

> Which rule, if any, is this issue related to?

`font-family-no-missing-generic-family-keyword`

> What CSS is needed to reproduce this issue?

Any css, even the empty file.

> What stylelint configuration is needed to reproduce this issue?

```json
{
	"rules": {
		"font-family-no-missing-generic-family-keyword": [true, {
			"severity": "error"
		}]
	}
}
```

> Which version of stylelint are you using?

`8.3.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

The configuration above should have the rule "font-family-no-missing-generic-family-keyword" enabled as an error.

> What actually happened (e.g. what warnings or errors you are getting)?

The following error is shown:

```shell
Invalid Option: Unexpected option value "[object Object]" for rule "font-family-no-missing-generic-family-keyword
```

hudochenkov:
Thank you for reporting and using a template.

We're taking care of this problem in https://github.com/stylelint/stylelint/pull/3039.

PR #3039, opened 2017-11-27
Fix incorrect font-family-no-missing-generic-family-keyword configuration #3038

hudochenkov:

https://github.com/stylelint/stylelint/issues/3038

I've added a test to `validateOptions`, but it doesn't catch an error. It will only show an error if test with configuration like in #3038 run.

I think we can add a new type of test in `jest-setup.js`. Take first `accept` test, add `severity` secondary option, and run test. There is a little warning: most test cases use `config: ["value"]`, while very few use `config: [["value"]]` (which is probably should be fixed). I can't think straight anymore today, so I'm not including this change for `jest-setup.js` in this PR.

### DIFF
--- a/lib/utils/__tests__/validateOptions.test.js
+++ b/lib/utils/__tests__/validateOptions.test.js
@@ -143,6 +143,18 @@ describe("validateOptions for secondary options objects", () => {
       'Invalid option value 2 for rule "foo": should be an object'
     );
   });
+
+  it(`should have "possible" if "optional" is set`, () => {
+    validateOptions(result, "foo", {
+      optional: true,
+      actual: { foo: "always" }
+    });
+
+    expect(result.warn.mock.calls[0]).toHaveLength(2);
+    expect(result.warn.mock.calls[0][0]).toEqual(
+      'Incorrect configuration for rule "foo". Rule should have "possible" values for options validation'
+    );
+  });
 });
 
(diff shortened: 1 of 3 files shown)


lib/utils/__tests__/validateOptions.test.js
```diff
@@ -143,6 +143,18 @@ describe("validateOptions for secondary options objects", () => {
...
   });
+
+  it(`should have "possible" if "optional" is set`, () => {
+    validateOptions(result, "foo", {
+      optional: true,
+      actual: { foo: "always" }
+    });
+
+    expect(result.warn.mock.calls[0]).toHaveLength(2);
+    expect(result.warn.mock.calls[0][0]).toEqual(
```
CAYdenberg: Can you use `toHaveBeenCalledTimes` and `toHaveBeenCalledWith`? Just easier to read for someone unfamiliar with Jest.

hudochenkov: I used same matchers as other tests in this file for consistency.

CAYdenberg: Weird, I thought I changes all of those when we removed Sinon. Oh well, point taken.

hudochenkov:
@CAYdenberg when you merge PR, please, delete branch and update changelog.

Issue #3035, opened 2017-11-26
postcss-html@0.11.0 requires a peer of postcss-sass@>=0.2.0


nikolay-borzov:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Looks like `postcss-html` got new peer dependency that isn't a `stylelint` dependency

> Which version of stylelint are you using?

 `8.3.0`

> What did you expect to happen?

Probably `postcss-sass` should be installed w/ stylelint

> What actually happened (e.g. what warnings or errors you are getting)?

Warning about unmet peer dependency

```shell
npm i stylelint -D
...
npm WARN postcss-html@0.11.0 requires a peer of postcss-sass@>=0.2.0 but none is installed. You must install peer dependencies yourself.
```

jeddy3:
@gucong3000 Thoughts?

modosc:
fwiw we already have all the other `peerDeps` from `postcss-html@0.11.0` listed in `dependencies`

gucong3000:
We can add `postcss-sass` to `dependencies` when finish #2503

billyjanitsch:
@gucong3000 why does `postcss-html` have a peer dependency on `postcss-sass` at all? It doesn't seem like the former references the latter anywhere in its source.

hudochenkov:
@billyjanitsch https://github.com/gucong3000/postcss-html/blob/6f252374c8ff0d10dc8317b76b5515add1d2397f/lib/get-syntax.js#L11-L12

billyjanitsch:
Ah, thanks for the pointer @hudochenkov! That explains why it didn't turn up in a search.

hudochenkov:
Done in https://github.com/stylelint/stylelint/pull/3040

PR #3040, opened 2017-11-27
Add postcss-sass as dependency, because postcss-html needs it

hudochenkov:

https://github.com/stylelint/stylelint/issues/3035

We don't use `postcss-sass` in our codebase, but it's a peer dependency of `postcss-html`, which we use. Luckily `postcss-html` doesn't fail if not every its dependency installed. But it might change theoretically, and we'll have a problem. Also we don't want annoy people by unmeet dependency messages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "postcss-reporter": "^5.0.0",
     "postcss-resolve-nested-selector": "^0.1.1",
     "postcss-safe-parser": "^3.0.1",
+    "postcss-sass": "^0.2.0",
     "postcss-scss": "^1.0.2",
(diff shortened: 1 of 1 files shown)


GreenGremlin:
Will this change be published soon? The peerDependency warnings are getting a bit annoying.

jeddy3:
@stylelint/core If no one beats me to it, I can do a minor release this evening.

jeddy3:
@GreenGremlin A regression delayed the release, but `8.4` is out now and should remove the peerDependency warning.

GreenGremlin:
Thanks!

PR #3063, opened 2017-12-12
Let processors modify the stylelintResult in case of CssSyntaxError

### LINKED ISSUE #3050: Allow processors to handle PostCSS errors

Feature.

Not related to a certain rule.

Not related to CSS codes.

The default configuration is enough.

8.0.0

Node API.

No.

Even if `postcss` errors were caught, give processors a chance to handle it, or by another configurable way.

For example, `styled-components/stylelint-processor-styled-components` normally extracts styles from React components, passes to `stylelint` to lint, and recovers lint results with right line numbers. But if an error happened, line numbers can not be recovered any more.

The program falls from [_lintSource](https://github.com/stylelint/stylelint/blob/master/lib/standalone.js#L216) into [handleError](https://github.com/stylelint/stylelint/blob/master/lib/standalone.js#L242) directly, skipping [_createStylelintResult](https://github.com/stylelint/stylelint/blob/master/lib/standalone.js#L239). So that [processors](https://github.com/stylelint/stylelint/blob/master/lib/createStylelintResult.js#L142) have no chance to be called.

emilgoldsmith:

resolves #3050 

I think it's pretty self explanatory, I have a few things I was in doubt about style wise / API wise that I would just like to ask / point out though:
1. Is there a specific commit format you'd like me to adhere to?

2. There are a few places I copy pasted from another file, so it therefore wasn't completely straight forward whether to modularize in order to apply DRY, specifically the flow type `resultType`, and the whole `stylelint.getConfigForFile(...).then(...)` block. I thought I'd see what you guys thought as I didn't know exactly which file I would put modularized functions/definitions in etc.

3. Since handleError can also throw an error did I get the return value right? This is my first time using flow.

4. In line 327 and 336 where I specify file path arguments, I was slightly in doubt where exactly to include defaults and which ones in that case in terms of exactly what's expected as I know that for example if you provide the code from stdin with the CLI there's no file name etc. I went with my gut but I would advise an extra thorough look at that part.

### DIFF
--- a/lib/createStylelintResult.js
+++ b/lib/createStylelintResult.js
@@ -58,7 +58,7 @@ const _ = require("lodash");
 }
 */
 
-/*:: type resultType = {
+/*:: type configForFileT = {
   config: {
     codeProcessors?: Array<Function>,
     ignorePatterns?: string,
@@ -162,7 +162,7 @@ module.exports = function(
   }
 
   return stylelint.getConfigForFile(filePath).then((
-    result /*: resultType*/
+    result /*: configForFileT*/
   ) => {
     const config = result.config;
     const file = source || (cssSyntaxError && cssSyntaxError.file);
--- a/lib/createStylelintResult.js
+++ b/lib/createStylelintResult.js
@@ -73,73 +73,105 @@ const _ = require("lodash");
 
 module.exports = function(
   stylelint /*: stylelint$internalApi*/,
-  postcssResult /*: postcssResultType*/,
-  filePath /*:: ?: string*/
+  postcssResult /*: ?postcssResultType*/,
+  filePath /*: ?string*/,
+  cssSyntaxError /*: ?stylelint$cssSyntaxError*/
 ) /*: Promise<stylelint$result>*/ {
-  const source = !postcssResult.root.source
-    ? undefined
-    : postcssResult.root.source.input.file ||
-      postcssResult.root.source.input.id;
+  let stylelintResult;
+  let source;
 
-  // Strip out deprecation warnings from the messages
-  const deprecationMessages = _.remove(postcssResult.messages, {
-    stylelintType: "deprecation"
-  });
-  const deprecations = deprecationMessages.map((
-    deprecationMessage /*: messageType*/
-  ) => {
-    return {
-      text: deprecationMessage.text,
-      reference: deprecationMessage.stylelintReference
-    };
-  });
+  if (postcssResult) {
+    source = !postcssResult.root.source
+      ? undefined
+      : postcssResult.root.source.input.file ||
+        postcssResult.root.source.input.id;
 
-  // Also strip out invalid options
-  const invalidOptionMessages = _.remove(postcssResult.messages, {
-    stylelintType: "invalidOption"
-  });
-  const invalidOptionWarnings = invalidOptionMessages.map((
-    invalidOptionMessage /*: messageType*/
-  ) => {
-    return {
-      text: invalidOptionMessage.text
-    };
-  });
-
-  const parseErrors = _.remove(postcssResult.messages, {
-    stylelintType: "parseError"
-  });
-
-  // This defines the stylelint result object that formatters receive
-  let stylelintResult = {
-    source,
-    deprecations,
-    invalidOptionWarnings,
-    parseErrors,
-    errored: postcssResult.stylelint.stylelintError,
-    warnings: postcssResult.messages.map((message /*: messageType*/) => {
+    // Strip out deprecation warnings from the messages
+    const deprecationMessages = _.remove(postcssResult.messages, {
+      stylelintType: "deprecation"
+    });
+    const deprecations = deprecationMessages.map((
+      deprecationMessage /*: messageType*/
+    ) => {
       return {
-        line: message.line,
-        column: message.column,
-        rule: message.rule,
-        severity: message.severity,
-        text: message.text
+        text: deprecationMessage.text,
+        reference: deprecationMessage.stylelintReference
       };
-    }),
-    ignored: postcssResult.stylelint.ignored,
-    _postcssResult: postcssResult
-  };
+    });
+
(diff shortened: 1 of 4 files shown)


jeddy3:
@emilgoldsmith Thanks for picking this one up.

@stylelint/core Does someone more familiar with flow want to comment on the flow queries. And does anyone have a moment to review the engine/api changes and comment on the related queries?

hudochenkov:
Thanks, @emilgoldsmith. Good PR, and good questions. Unfortunately, I'm not that familiar with the base of stylelint and can't answer any of this questions :(

Have you tried your PR on https://github.com/styled-components/stylelint-processor-styled-components/issues/154? Does it solve the problem?

emilgoldsmith:
Yep, solves the problem :)

ntwb:
@CAYdenberg Any chance you could review this PR please?

CAYdenberg:
Yes I can look tonight or tomorrow night.

lib/standalone.js
```diff
@@ -33,6 +34,19 @@ const ALWAYS_IGNORED_GLOBS = ["**/node_modules/**", "**/bower_components/**"];
   source: string;
 }*/
 
+/*:: type resultType = {
+  config: {
```
CAYdenberg: I think the name resultType here is confusing as there is also a stylelint$result used in the same file (and it's a completely different thing?)

If I understand the code correctly this is the result of getConfigForFile ... so is "ConfigForFileT" maybe a more appropriate name?

I'd prefer using T instead of Type (and CamelCase beginning with UpperCase) as it's consistent with CssSyntaxErrorT.

lib/standalone.js
```diff
@@ -305,4 +321,25 @@ function convertCssSyntaxErrorToResult(
       }
     ]
   };
+
+  // Let custom processors mutate result to for example apply source maps
+  // before sending result to user
+  return stylelint.getConfigForFile(error.file || process.cwd()).then((
+    result /*: resultType */
+  ) => {
```
CAYdenberg: You can consider renaming this variable if you agree with my nitpicking, above.

emilgoldsmith:
Thanks for the review @CAYdenberg.

It seems you missed my initial PR message where I asked a few questions and among other things mentioned that the type I used there was copy pasted from another file (I think createStylelintResult), including the name. If you could address the 4 questions above that would be great! Thanks! And of course if we end up not DRYing it up and just use the copy pasted definition with the more appropriate name you suggested I can also easily handle that, I'll just let you get back to me, now that you have more information, first :).

CAYdenberg:
Oh I see. I did miss that. I guess my critcisms really apply to the original code in createStylelintResult. Lol - it still seems to me that resultType is not really describing a _result_.

I'm generally in favour of DRY code, and I think it would be possible to sufficiently generalize `createStylelintResult` such that you don't need a new `convertCssSyntaxErrorToResult` function in this file. It's getting late for me though and there might be something I'm missing ... maybe tomorrow night I can play with it a bit if you don't mind?

Don't worry too much about commit formats as we squash on merge anyway.

(3) This looks fine ... throw within function doesn't affect the return value type.

(4) Have to think about this one - maybe will discover something during my other investigation.

emilgoldsmith:
Yep no worries, take your time :)

CAYdenberg:
Hey Emil checkout stylelint/issue-3050 and see what you think. If you like you can merge the commits directly into this PR (I couldn't seem to make a PR to your repo).

Or if you like you can steal the basic idea and do your own version of it.

emilgoldsmith:
Will do!

> (I couldn't seem to make a PR to your repo)

That's weird, I even have the allow maintainers to edit box ticked

emilgoldsmith:
I'll get to this very soon, on the way home for Christmas I actually looked at it and think it looks great, I just wanted to test it with our custom processor which I think I didn't have time for before my flight was taking off, and then I went mostly offline for the holidays. But yeah will probably just merge in your changes @CAYdenberg and I'll be back with a more thorough update (and commits) in the very near future

emilgoldsmith:
Okay, looked through it again and tested it with our code, and it seems to still correctly fix our issue at https://github.com/styled-components/stylelint-processor-styled-components/issues/154, and this looks really nice and DRY @CAYdenberg! Thanks for the help! I've looked through the code and it looks good to me and CI is passing, so I guess it's up to you guys whether you want someone that's not @CAYdenberg to review his changes or if I'm sufficient ;). 

Look forward to having this live! :)

CAYdenberg:
OK @stylelint/core not sure given #3093 if my review should count since I've now committed, but I guess speak now, or forever hold your peace!

lib/createStylelintResult.js
```diff
@@ -58,7 +58,7 @@ const _ = require("lodash");
 }
 */
 
-/*:: type resultType = {
+/*:: type configForFileT = {
```
ntwb: @CAYdenberg In my ongoing attempt at learning and understanding Flow better, why the name change here? Also should it be `configForFileType` or are these not really about making them _human-readable_?

CAYdenberg: I found it confusing when reading originally that this type is declared in a file called createStylelintResult, but this is NOT the type of the result. That is stylelint$result (returned on line 79). 

So I looked for another name ... this is the type of the returned value of `getConfigForFile` (encased in a promise), so `configForFile` is what it is. Possibly the variable name should have been changed too.

I believe the capital T is used in other places but not here, so I'll change that too.

ntwb: Sweet, thanks for the explanation, and the follow up commit 👍

emilgoldsmith:
So this is ready for merging now right?

CAYdenberg:
Yes, thanks for your work on this @emilgoldsmith - sorry about the delays, we're having some ongoing discussions around our process and code reviews.

emilgoldsmith:
No problem! Thanks for merging it in, I look forward to having the functionality active in our processor :)

PR #3071, opened 2017-12-18
Update Prettier

hudochenkov:
Run the latest Prettier over the entire code base. Also pinned to version to minor version with `~`, because even minor version might change code formatting.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+.coverage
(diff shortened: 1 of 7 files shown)


Arcanemagus:
Their [documentation](https://prettier.io/docs/en/install.html) actually recommends pinning an _exact_ version, but that does lead to more upgrade maintenance when using it 😉.

ntwb:
We could add a CI job to run `npm run prettier:check` but I think that would add a barrier for contributors where unrelated failures would surface in PRs 🤔

PR #3075, opened 2017-12-19
Bump minimum required Node.js version to `6` and drop 4.x

### LINKED ISSUE #2996: Drop Node.js 4 support
Shall we drop Node.js 4 support in the next major version?

Today [Node.js 8 moved to LTS](https://medium.com/the-node-js-collection/news-node-js-8-moves-into-long-term-support-and-node-js-9-becomes-the-new-current-release-line-74cf754a10a0) (i. e. it's now production ready). Node.js 4 became more legacy and outdated. It's in maintenance mode [since April 2017](https://github.com/nodejs/Release), which will end in April 2018.

It will give our codebase and us a breath of fresh air. There are a bunch of awesome stuff starting Node.js 6, which are missing in Node.js 4: default function parameters, rest parameters, spread operator, destructuring, no need for `'use strict'` to use `let, const`. Currently, we have to use “old” methods to do what these ES6 features do, which leads to more code.

Also, we'll have lower test time in CI (if we won't add Node.js 9 :)).

---

Not related: we've just passed 4000 stars :)

ntwb:

Closes #2996

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,10 +30,6 @@ matrix:
     - node_js: '6'
       script: npm run jest -- --runInBand
       env: CI=tests 6
-    - node_js: '4'
-      script: npm run jest -- --runInBand
-      env: CI=tests 4
-      sudo: required
 
 before_install:
   - npm install -g npm@latest
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "!**/__tests__"
   ],
   "engines": {
-    "node": ">=4.2.1"
+    "node": ">=6"
   },
   "dependencies": {
     "autoprefixer": "^7.1.2",


hudochenkov:
If we merge this, then the next version should be major. It's not convenient if we'd have to release a patch-level fix. But maybe major bump shouldn't be with big content. Like we piled things for 8.0.0 release for more than half a year.

ntwb:
Agreed, agreed, and agreed.

hudochenkov:
@ntwb do you think it's ok to release a major version with just dropped Node.js 4 support?

ntwb:
Yes, I don't see any issue in doing that, @stylelint/core, thoughts?

jeddy3:
> Like we piled things for 8.0.0 release for more than half a year.

I remember that... it wasn't the best approach.

ntwb:
I was thinking of releasing 9.0.0 yesterday/today, now today it's now 8pm Friday night here I think we should hold off until at least the middle of next week.

I'm off for a few days to spend a few days relaxing with family 🎉 

Wishing every contributor and the @stylelint/core team happy holidays 😄

hudochenkov:
Happy holidays!

jeddy3:
Happy holidays! :)

gucong3000:
Rebase into [master](https://github.com/stylelint/stylelint/commit/f38bd78bbae75284ed54e43516c4c8aceafa06e9) a few minutes ago.

ntwb:
Changelog: https://github.com/stylelint/stylelint/commit/c7b04a0eba369ae6d495479971b39bd9e41698ba & https://github.com/stylelint/stylelint/commit/b57467e45d4b5d2a801d9094656bd92b138ec3b6
> `Removed: Node.JS 4.x support. Node.js 6.x or greater is now required (#3075).`

PR #3076, opened 2017-12-20
chore(package): update jest to version 22.0.3

CAYdenberg:
Closes #3069


ntwb:
Let's stick with #3069 for now, I've noted in that PR what fixes have been added in the .1,.2,and.3 releases which do not help us out here, we're leaking memory and we need to investigate the casuse of that before our tests will pass.

CAYdenberg:
I was just trying to see if the tests would pass on Travis as they work for me locally.

PR #3081, opened 2017-12-27
Run Jest in parallel

hudochenkov:

https://github.com/stylelint/stylelint/pull/3069

Jest running in parallel consumes less memory. I didn't find why we use `--runInBand` flag in the first place. We just started using it in https://github.com/stylelint/stylelint/pull/2134

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,13 +22,13 @@ matrix:
       script: npm run pretest
       env: CI=pretest
     - node_js: '8'
-      script: npm run jest -- --coverage
+      script: npm run jest -- --maxWorkers=2 --coverage
       env: CI=coverage
     - node_js: '9'
-      script: npm run jest
+      script: npm run jest -- --maxWorkers=2
       env: CI=tests 9
     - node_js: '6'
-      script: npm run jest
+      script: npm run jest -- --maxWorkers=2
       env: CI=tests 6
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,13 +22,13 @@ matrix:
       script: npm run pretest
       env: CI=pretest
     - node_js: '8'
-      script: npm run jest -- --runInBand  --coverage
+      script: npm run jest -- --coverage
       env: CI=coverage
     - node_js: '9'
-      script: npm run jest -- --runInBand
+      script: npm run jest
       env: CI=tests 9
     - node_js: '6'
-      script: npm run jest -- --runInBand
+      script: npm run jest
       env: CI=tests 6
 
 before_install:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,4 +19,4 @@ install:
 test_script:
   - node --version
   - npm --version
-  - cmd: "npm run jest -- --runInBand --testPathPattern \\lib"
+  - cmd: "npm run jest -- --testPathPattern \\lib"
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "file-exists-promise": "^1.0.2",
     "flow-bin": "^0.63.1",
     "husky": "^0.14.3",
-    "jest": "^22.0.4",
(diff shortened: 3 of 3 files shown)


m-allanson:
@hudochenkov I _think_ `--runInBand` was added to work around some out of memory errors in Travis and an older version of Jest. If the CI works without it, then it's probably ok to remove.

ntwb:
Let's wait and see the upstream response from the Jest team per https://github.com/stylelint/stylelint/pull/3069#issuecomment-354106535

hudochenkov:
I would like to keep this PR for a while. Even with Jest 22.0.6 there some problems in default (parallel) mode. We have some leaks in our code (https://github.com/facebook/jest/issues/5157#issuecomment-356807847). Maybe parallel mode will work once we fix those leaks.

rickhanlonii:
Looks like we're running out of memory when using coverage

I just filed https://github.com/facebook/jest/issues/5285 which is a leak when using any coverage

@hudochenkov one thing you could try until the leak is fixed is `--maxWorkers=2`. This fixes Circle builds and may also work for Travis (see explanation [here](https://github.com/facebook/jest/issues/5239#issuecomment-355867359))

hudochenkov:
@rickhanlonii thank you very much for suggestion and explanation!

I restarted this PR and `master` at the same time to have a relatively same testing environment. Top — `--maxWorkers=2`, bottom — `--runInBand`:

image

@stylelint/core what do you think about this change? I think we should give it a try.

Arcanemagus:
Looks like the slight slowdown in the `CI=pretest` build was due to running on a host with slower I/O:

`master`:
image

This PR:
image

_Note: a rebase on `master` should drop https://github.com/stylelint/stylelint/pull/3081/commits/4268ca9d4cbdf652e12920c9b992ca9b1490c08c as it duplicates https://github.com/stylelint/stylelint/commit/7a67e0990d9018805c93736b91bb9c0609f11cd2._ 😉

hudochenkov:
> Looks like the slight slowdown in the `CI=pretest` build was due to running on a host with slower I/O

Even with a slower I/O Jest tests were faster. This is a good sign.

> _Note: a rebase on `master` should drop https://github.com/stylelint/stylelint/pull/3081/commits/4268ca9d4cbdf652e12920c9b992ca9b1490c08c as it duplicates https://github.com/stylelint/stylelint/commit/7a67e0990d9018805c93736b91bb9c0609f11cd2._ 😉

Yeah, it will go away, because the only option to merge PR is to squash it. So it will go away naturally with squash.

Arcanemagus:
> Even with a slower I/O Jest tests were faster. This is a good sign.

Actually we have no guarantee of that since we don't know that all the jobs within that build ran on the same host. The only thing that can be said is that the `CI=pretest` runs were different due to what looks like I/O time. That job has nothing to do with `jest` though.

> Yeah, it will go away, because the only option to merge PR is to squash it. So it will go away naturally with squash.

Ah! Didn't know the other merge types were disabled here. 👍

CAYdenberg:
LGTM. Not sure it's any faster at least locally.

ntwb:
Here are some local timings here on my Late 2014 Mac Book Pro:

• `npm run jest -- --maxWorkers=2` = `61.588s`
• `npm run jest` = `57.252s`

The `maxWorkers` flag should utilise as many CPU threads as your PC will allow, hence it should be faster to run locally without restricting the `maxWorkers` value ([docs](https://facebook.github.io/jest/docs/en/cli.html#maxworkers-num)

hudochenkov:
Correct comparison is:

* `npm run jest -- --runInBand` = 114.094s — what we had before.
* `npm run jest -- --maxWorkers=2` = 52.399s — now.
* `npm run jest` = 26.991s — without CPU limitation. I guess I have 8 cores.

PR #3087, opened 2017-12-29
Add selector-pseudo-element-*list rules

### LINKED ISSUE #3014: Add selector-pseudo-element-*list
Ref: https://github.com/stylelint/stylelint/issues/3010#issuecomment-342436260

- Names: `selector-pseudo-element-blacklist` and `selector-pseudo-element-whitelist`
- Description: Specify a black/whitelist of disallowed/allowed pseudo-element selectors.
- Primary option: `array` (accepts string and regex)
- Secondary options: none
- Message: "Unexpected pseudo-element "${pseudoElement}""
- Section: "Limit language features -> Selector"

jeddy3:
Closes #3014 

It’s been a while since I had the time to write some actual stylelint code, especially adding new rules.

I found myself reaching for the `selector-pseudo-element-whitelist` rule on my current project, so figured I’d add it.


lib/rules/selector-pseudo-element-blacklist/index.js
```diff
@@ -0,0 +1,78 @@
...
+
+const messages = ruleMessages(ruleName, {
+  rejected: selector => `Unexpected pseudo-element "${selector}"`
+});
+
+const rule = function(blacklist) {
+  return (root, result) => {
+    const validOptions = validateOptions(result, ruleName, {
+      actual: blacklist,
+      possible: [_.isString]
```
hudochenkov: It should also check for `_.isRegExp` according to readme. It would be great to have RegExp in tests as well. Both as a string and as RegExp.

The same comment applies to the second rule.

jeddy3:
@hudochenkov Thanks for the feedback. I've made the changes.

I'm about to head out, so I've pushed up what I've got... and there's a flow error. Any flow pros know how to fix this? Setting the type to `string | Class<RegExp>` gives:

```
^^^^ property `test`. Property not found in statics of `RegExp`
```

CAYdenberg:
Right, so flow doesn't understand that `_.isRegExp` is ensuring this is a regular expression. It thinks `comparison` could be a RegExp or a string, so it's warning you that `.test` doesn't work for strings.

Try using `typeof comparison` (harder for the reader, since `typeof /myregexp/ === "object"`) or 

```js
if (comparison.test) {
    // now we know this is RegExp and we can CALL comparison.test
}
```

Make sense?

jeddy3:
> Make sense?

Yes, thanks very much @CAYdenberg 

---

This PR is ready for review again.

jeddy3:
@hudochenkov Restarted travis and now CI is green.

jeddy3:
Ready for a 2nd review

jeddy3:
> -   Added: `selector-pseudo-element-*list` rules ([#3104](https://github.com/stylelint/stylelint/pull/3087)).

PR #3088, opened 2017-12-29
Add selector-combinator-*list rules

### LINKED ISSUE #3013: Add selector-combinator-*list
Ref: https://github.com/stylelint/stylelint/issues/3010#issuecomment-342436260

- Names: `selector-combinator-blacklist` and `selector-combinator-whitelist`
- Description: Specify a black/whitelist of disallowed/allowed combinators.
- Primary option: `array` (accepts string and regex)
- Secondary options: none
- Message: "Unexpected combinator "${combinator}""
- Section: "Limit language features -> Selector"
- Notes: This rule ignores [reference combinators](https://www.w3.org/TR/selectors4/#idref-combinators).

jeddy3:
Closes #3013


lib/utils/__tests__/isStandardSyntaxCombinator.test.js
```diff
@@ -0,0 +1,75 @@
...
+      expect(isStandardSyntaxCombinator(func)).toBeTruthy();
+    });
+  });
+  it("descendant newline", () => {
+    return rules("a\nb {}", func => {
+      expect(isStandardSyntaxCombinator(func)).toBeTruthy();
+    });
+  });
+  it("descendant (double child)", () => {
+    return rules("a >> b {}", func => {
```
hudochenkov: I can't find it in a [specification](https://www.w3.org/TR/css3-selectors/#combinators). Doesn't look like a standard one.

lib/rules/selector-combinator-blacklist/README.md
```diff
@@ -0,0 +1,50 @@
...
+```css
+a b {}
+```
+
+```css
+a > b {}
+```
+
+```css
+a >>> b {}
```
hudochenkov: It's not a standard combinator. Maybe it's better not to add it to readme?

lib/rules/selector-combinator-blacklist/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+const messages = ruleMessages(ruleName, {
+  rejected: combinator => `Unexpected combinator "${combinator}"`
+});
+
+const rule = function(blacklist) {
+  return (root, result) => {
+    const validOptions = validateOptions(result, ruleName, {
+      actual: blacklist,
+      possible: [_.isString]
```
hudochenkov: It should also check for `_.isRegExp` according to readme. It would be great to have RegExp in tests as well. Both as a string and as RegExp.

The same comment applies to the second rule.

lib/rules/selector-combinator-blacklist/README.md
```diff
@@ -0,0 +1,50 @@
...
+  a + b {}
+/** ↑
+ * These combinators */
+```
+
+This rule ignores [reference combinators](https://www.w3.org/TR/selectors4/#idref-combinators) e.g. `/for/`.
+
+## Options
+
+`array|string|regex`: `["array", "of", "unprefixed", "pseudo-elements" or "regex"]|"pseudo-element"|/regex/`
```
hudochenkov: These rule about combinators.

lib/rules/selector-combinator-whitelist/README.md
```diff
@@ -0,0 +1,50 @@
...
+  a + b {}
+/** ↑
+ * These combinators */
+```
+
+This rule ignores [reference combinators](https://www.w3.org/TR/selectors4/#idref-combinators) e.g. `/for/`.
+
+## Options
+
+`array|string|regex`: `["array", "of", "unprefixed", "pseudo-elements" or "regex"]|"pseudo-element"|/regex/`
```
hudochenkov: These rule about combinators.

lib/rules/selector-combinator-whitelist/README.md
```diff
@@ -0,0 +1,50 @@
...
+```css
+a b {}
+```
+
+```css
+a > b {}
+```
+
+```css
+a >>> b {}
```
hudochenkov: It's not a standard combinator. Maybe it's better not to add it to readme?

lib/rules/selector-combinator-blacklist/README.md
jeddy3: Good point

lib/rules/selector-combinator-blacklist/README.md
jeddy3: Oops

lib/utils/__tests__/isStandardSyntaxCombinator.test.js
jeddy3: It's a [level 4 one](https://drafts.csswg.org/selectors-4/#descendant-combinators). I'll take it out though

jeddy3:
> I'm going to push back a bit on allowing RegExp's ... most combinators are just a single special character (+, ~) and I don't see this feature being all that useful.

That's a fair point. I originally intended to add RegExps for controlling the what whitespace is allowed for descendant combinators. I've updated the tests and examples to show this, rather than use the child and piercing combinators.

> I'm thinking that POSSIBLY -whitelist should assume " " is a valid combinator and add it automatically to the config?

I think there is a use case for not allowing the descendant combinator e.g. a completely flat specificity graph or when (exclusively) using CSS Modules' `composes` system.

Having said all that, perhaps it is better to split this up:

- `selector-descendant-combinator-no-non-space` (already built)
- `selector-no-descendent-combinator`
- `selector-combinator-blacklist`
- `selector-combinator-whitelist`

The latter two would ignore whitespace descendant combinators and not accept RegExp.

What do you think?

CAYdenberg:
> I think there is a use case for not allowing the descendant combinator e.g. a completely flat specificity graph or when (exclusively) using CSS Modules' composes system.

Fair. I'm sold on this the way it is. Having to create another rule and to handle an exception to this one seems like a waste. Maybe just make sure the README specifically mentions that " " is itself a combinator and needs to be whitelisted explicitly.

> I originally intended to add RegExps for controlling what whitespace is allowed for descendant combinators.

Yeah, this seems like a different concern to me. Stylistic rather than functional.

jeddy3:
Ummm... I think you've (despite your conclusion) talked me out of it.

> Having to create another rule and to handle an exception to this one seems like a waste.

This rule is unique amongst the `-*list` rules because it the decedent combinator itself isn't a simple string comparison. So, I think there's a case for splitting `selector-no-descendent-combinator` out and keeping the `-*list` rules consistent.

> Yeah, this seems like a different concern to me. Stylistic rather than functional.

That's a good point. Keeping support for the descendant combinator in makes this rule half functional and half stylistic.

lib/rules/selector-combinator-blacklist/README.md
```diff
@@ -0,0 +1,46 @@
...
+  a + b {}
+/** ↑
+ * These combinators */
+```
+
+This rule ignores [reference combinators](https://www.w3.org/TR/selectors4/#idref-combinators) e.g. `/for/`.
+
+## Options
+
+`array|string|regex`: `["array", "of", "unprefixed", "combinators" or "regex"]|"combinator"|/regex/`
```
hudochenkov: What are prefixed combinators? :)

hudochenkov:
I agree with @CAYdenberg. RegExp here might be redundant.

> > Yeah, this seems like a different concern to me. Stylistic rather than functional.
> 
> That's a good point. Keeping support for the descendant combinator in makes this rule half functional and half stylistic.

I believe these rules should be functional and not about the style of code.

To avoid using RegExp and avoid stylistic concerns we could replace `/\s+/` in selectors with single space and do comparisons against normalized selector.

jeddy3:
> To avoid using RegExp and avoid stylistic concerns we could replace /\s+/ in selectors with single space and do comparisons against normalized selector.

That's a great idea. It does away with the need for the `selector-no-descendent-combinator` rule and can be easily communicated in the README.

I'll make the change tomorrow.

jeddy3:
@hudochenkov & @CAYdenberg This is ready for review again.

hudochenkov:
@CAYdenberg please, add a changelog item when PR merged https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/pull-requests.md

CAYdenberg:
I did ... just took a few extra minutes because I also have kids 🤕

hudochenkov:
@CAYdenberg sorry, man. Saw you started answering other issues, and though you forgot. Sorry.

CAYdenberg:
NP ... appreciate the respect for our process.

ntwb:
Changelog: https://github.com/stylelint/stylelint/commit/741cd6e126280378b0433f9cf24ba713095163be#diff-4ac32a78649ca5bdd8e0ba38b7006a1e

> -   Added: `selector-combinator-*list rules` ([#3088](https://github.com/stylelint/stylelint/pull/3088)).

PR #3089, opened 2017-12-30
property-no-vendor-prefix: Add ignoreProperties

### LINKED ISSUE #3007: Add ignoreProperties: [] to property-no-vendor-prefix

I use Stylelint with an Electron app where I have to use a few vendor prefixed properties.

`property-no-vendor-prefix` and `value-no-vendor-prefix`

e.g.

```css
div {
	-webkit-app-region: drag;
	cursor: -webkit-grab;
}
```

I'm aware I could run Autoprefixer, but that's a waste for only two properties. And it's annoying to have to add a lot of inline ignore comments.

```json
{
  "rules": {
    "property-no-vendor-prefix": true,
  }
}
```

`8.2.0`

`stylelint --config myconfig *.css`

No

I expected these rules to have an `ignore` option like many other rules have.

The vendor prefixes got reported.

nalinbhardwaj:

#3007 

No, it's self explanatory. Very similar to #3015.

### DIFF
--- a/lib/rules/property-no-vendor-prefix/README.md
+++ b/lib/rules/property-no-vendor-prefix/README.md
@@ -38,3 +38,23 @@ columns: 2; }
 ```css
 a { -webkit-touch-callout: none; }
 ```
+
+## Optional secondary options
+
+### `ignoreProperties: ["/regex/", "string"]`
+
+Given:
+
+```js
+["transform", "columns"]
+```
+
+The following patterns are *not* considered violations:
+
+```css
+a { -webkit-transform: scale(1); }
+```
+
+```css
+a { -moz-columns: 2; }
+```
--- a/lib/rules/property-no-vendor-prefix/__tests__/index.js
+++ b/lib/rules/property-no-vendor-prefix/__tests__/index.js
@@ -100,3 +100,49 @@ testRule(rule, {
     }
   ]
 });
+
+testRule(rule, {
+  ruleName,
+  config: [
+    true,
+    { ignoreProperties: ["transform", "columns", "/^animation-/i"] }
+  ],
+
+  accept: [
+    {
+      code: "a { -webkit-transform: scale(1); }"
+    },
+    {
+      code: "a { -moz-columns: 2; }"
+    },
+    {
+      code: "a { -webkit-animation-delay: 0.5s; }"
+    },
+    {
+      code: "a { -webkit-ANIMATION-DeLaY: 0.5s; }"
+    },
+    {
+      code: "a { width: 320px; }"
+    }
+  ],
+  reject: [
+    {
+      code: "a { -webkit-border-radius: 10px; }",
+      message: messages.rejected("-webkit-border-radius"),
+      line: 1,
+      column: 5
+    },
+    {
+      code: "a { -moz-background-size: cover; }",
+      message: messages.rejected("-moz-background-size"),
+      line: 1,
+      column: 5
+    },
+    {
+      code: "a { -WEBKIT-tranSFoRM: translateY(-50%); }",
+      message: messages.rejected("-WEBKIT-tranSFoRM"),
+      line: 1,
+      column: 5
+    }
+  ]
+});
--- a/lib/rules/property-no-vendor-prefix/index.js
+++ b/lib/rules/property-no-vendor-prefix/index.js
@@ -1,6 +1,9 @@
 "use strict";
 
+const _ = require("lodash");
 const isAutoprefixable = require("../../utils/isAutoprefixable");
+const optionsMatches = require("../../utils/optionsMatches");
+const postcss = require("postcss");
 const report = require("../../utils/report");
 const ruleMessages = require("../../utils/ruleMessages");
 const validateOptions = require("../../utils/validateOptions");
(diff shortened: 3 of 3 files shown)


CAYdenberg:
@stylelint/core Can someone else review this at their earliest convenience? This student is working with Google Code-In (GCI) and the PR needs to be merged in order for them to get credit for the assignment.

hudochenkov:
Thank you for the contribution!

I see few issues with documentation and tests. I'll do a proper review later today.

nalinbhardwaj:
@hudochenkov : I looked around in the repo myself to look for possible errors, and realised that other examples were consistent across the page, perhaps that's the error you refer to, if so, I've fixed it. :)

lib/rules/property-no-vendor-prefix/__tests__/index.js
```diff
@@ -100,3 +100,46 @@ testRule(rule, {
...
+
+  accept: [
+    {
+      code: "a { -webkit-transform: scale(1); }"
+    },
+    {
+      code: "a { -moz-columns: 2; }"
+    },
+    {
+      code: ".foo { -webkit-animation-delay: 0.5s; }"
```
hudochenkov: Replace `.foo` with `a` in all tests.

lib/rules/property-no-vendor-prefix/__tests__/index.js
```diff
@@ -100,3 +100,46 @@ testRule(rule, {
...
 });
+
+testRule(rule, {
+  ruleName,
+  config: [
+    true,
+    { ignoreProperties: ["transform", "columns", "/^animation-/"] }
+  ],
+
+  accept: [
```
hudochenkov: Add `a { -wEbKiT-tRaNsFoRm: scale(1); }` `accept` test. That we know it's case insensitive.

Both config value case and property case should be case insensitive.

`ignoreProperties: ["transform"]` should match `a { -webkit-TrAnSfOrM: none; }`. And `ignoreProperties: ["TrAnSfOrM"]` should match `a { -webkit-trAnsfOrm: none; }`, etc.

Could you also add test for the later one, please?

lib/rules/property-no-vendor-prefix/__tests__/index.js
```diff
@@ -100,3 +100,46 @@ testRule(rule, {
...
+    {
+      code: ".foo { -webkit-animation-delay: 0.5s; }"
+    },
+    {
+      code: ".foo { width: 320px; }"
+    }
+  ],
+  reject: [
+    {
+      code: "-webkit-border-radius: 10px;",
```
hudochenkov: Replace with `a { -webkit-border-radius: 10px; }`, because property outside a rule isn't a valid CSS.

lib/rules/property-no-vendor-prefix/__tests__/index.js
```diff
@@ -100,3 +100,46 @@ testRule(rule, {
...
+      column: 1
+    },
+    {
+      code: ".foo { -moz-background-size: cover; }",
+      message: messages.rejected("-moz-background-size"),
+      line: 1,
+      column: 8
+    },
+    {
+      code: ".foo { -WEBKIT-tranSFoRM: translateY(-50%); }",
```
hudochenkov: This shouldn't be rejected because `transform` is in ignored properties.

lib/rules/property-no-vendor-prefix/index.js
```diff
@@ -11,15 +14,33 @@ const messages = ruleMessages(ruleName, {
...
+    );
+
     if (!validOptions) {
       return;
     }
 
     root.walkDecls(decl => {
-      const prop = decl.prop;
+      const prop = decl.prop,
+        unprefixedProp = postcss.vendor.unprefixed(prop);
```
hudochenkov: Use one `const` per assignment, please. We should've put a rule about this in our ESLint config.

lib/rules/property-no-vendor-prefix/__tests__/index.js
jeddy3: > Both config value case and property case should be case insensitive.

I believe this behaviour [was changed in `8.0.0`](https://github.com/stylelint/stylelint/pull/2709). Options are now case sensitive and `optionsMatches` will behave as such. Users should explicitly use the `i` RegExp flag for case-insensitivity e.g. `ignoreProperties: ["/transform/i"]`.

So `ignoreProperties: ["transform"] should not match `a { -webkit-TrAnSfOrM: none; }`.

It might be worth updating the `config` to `{ ignoreProperties: ["transform", "columns", "/^animation-/i"] }` so that this behaviour can be confirmed in this rule i.e. `animation-name` and `ANIMATION-name` will both match, but `TRANSFORM` will not.

lib/rules/property-no-vendor-prefix/README.md
```diff
@@ -38,3 +38,23 @@ columns: 2; }
 ```css
 a { -webkit-touch-callout: none; }
 ```
+
+
+### `ignoreProperties: ["string"]`
```
jeddy3:
```
### `ignoreProperties: ["/regex/", "string"]`
```

RegExp are also accepted.

lib/rules/property-no-vendor-prefix/__tests__/index.js
jeddy3: @nalinbhardwaj  FYI, this requirement is in the [guide](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#example-patterns).

lib/rules/property-no-vendor-prefix/__tests__/index.js
jeddy3: See my comment above as why I believe this is valid.

lib/rules/property-no-vendor-prefix/__tests__/index.js
nalinbhardwaj: Fixing. There are some more tests in the repo that violate this, [example](https://github.com/stylelint/stylelint/blob/master/lib/rules/value-no-vendor-prefix/__tests__/index.js#L137)

nalinbhardwaj:
@jeddy3, @hudochenkov : Fixed.

jeddy3:
> -   Added: `ignoreProperties: []` to `property-no-vendor-prefix` ([#3089](https://github.com/stylelint/stylelint/pull/3089)).

PR #3096, opened 2018-01-02
Update PostCSS to v6.0.15

ntwb:
After updating to PostCSS v6.0.15 a bunch of console warnings are displayed during test runs:

```shell
 PASS  lib/testUtils/__tests__/createRuleTester.test.js
  ● Console

    console.warn node_modules/postcss/lib/warn-once.js:11
      Witout `from` option PostCSS could generate wrong source map or do not find Browserslist config. Set it to CSS file path or to `undefined` to prevent this warning
```

[PostCSS v6.0.15 release notes](https://github.com/postcss/postcss/releases/tag/6.0.15): _"Add warning about missed from option on `process().then()` call."_

### DIFF
--- a/lib/__tests__/defaultSeverity.test.js
+++ b/lib/__tests__/defaultSeverity.test.js
@@ -9,10 +9,12 @@ it("`defaultSeverity` option set to warning", () => {
       "block-no-empty": true
     }
   };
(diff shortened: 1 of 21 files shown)


hudochenkov:
Thanks, @ntwb!

PR #3097, opened 2018-01-02
Update eslint-config-stylelint to v8.1.0

ntwb:
See https://github.com/stylelint/eslint-config-stylelint/pull/42

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "d3-queue": "^3.0.7",
     "del": "^3.0.0",
     "eslint": "~4.14.0",
-    "eslint-config-stylelint": "^8.1.0",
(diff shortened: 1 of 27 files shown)


package.json
```diff
@@ -85,7 +85,7 @@
     "d3-queue": "^3.0.7",
     "del": "^3.0.0",
     "eslint": "~4.14.0",
-    "eslint-config-stylelint": "^8.0.0",
+    "eslint-config-stylelint": "^8.1.0",
```
hudochenkov: I think we should pin it as well as ESLint: 

```
"eslint-config-stylelint": "~8.1.0",
```

ntwb: Done

PR #3100, opened 2018-01-03
Add Table of Contents marker to docs

hudochenkov:
Continue making our site better https://github.com/stylelint/stylelint.io/issues/42. Added TOC for pages, where it makes sense.

Looks pretty good:

screen shot 2018-01-03 at 23 14 56

I have concerns about List of rules, though:

localhost_3333_user-guide_rules_

### DIFF
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -4,6 +4,8 @@ Plugins are rules and sets of rules built by the community.
 
 We recommend familiarising yourself and adhering to stylelint's [conventions for writing rules](rules.md), including those for names, options, messages, tests and docs.
 
+<!-- TOC -->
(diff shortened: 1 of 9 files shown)


jeddy3:
Is it possible to limit the TOC to one level on the rules page?

hudochenkov:
Why? It's quite useful to have every level for many rules https://stylelint.io/user-guide/rules/block-closing-brace-newline-after/

jeddy3:
> Update: Did you mean https://stylelint.io/user-guide/rules/? 

Yep, I meant that one :)

> Then yes, we can limit it. It will be easier to make TOC manually in markdown in this case, rather write a lot of code for TOC plugin.

Makes sense.

Issue #3105, opened 2018-01-05
False negatives nested at-if in selector-max-id


BenjaminWFox:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug where "selector-max-id" is not applied if the selector is followed by an SCSS @if statement.

> Which rule, if any, is this issue related to?

"selector-max-id"

> What CSS is needed to reproduce this issue? (SCSS)

e.g.

```css
$p: 4;

Not a violation:
#temp #two {
  @if ($p == 4) {
    font-size: 4em;
  }
  
  color: #000;
}

A violation:
#temp #two {  
  color: #000;
}

```

> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
  "rules": {
    "selector-max-id": 1
  }
}
```

> Which version of stylelint are you using?

8.4.0

> How are you running stylelint: CLI, PostCSS plugin, Node API?

Sublimetext 3 with SublimeLinter-contrib-stylelint
AND
Webpack with stylelint-webpack-plugin

(issue shows up with both)

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes, it's related to SCSS @if

> What did you expect to happen?

selector-max-id should be applied regardless of what follows in the block

> What actually happened (e.g. what warnings or errors you are getting)?

selector-max-id was not applied when the block started with @if

jeddy3:
@BenjaminWFox Thanks for the report and for using the template.

Please consider [contributing a fix](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#fixing-a-bug-in-an-existing-rule) as the core team is focused on standard syntax. I suspect the issue is to [do with resolving nested blocks](https://github.com/stylelint/stylelint/blob/master/lib/rules/selector-max-id/index.js#L66-L73). At the moment, only the _fullest_ selector is checked. Perhaps rule-sets should be ignored _only if_ any of its children _contain_ id selectors.

BenjaminWFox:
@jeddy3 Thanks for the pointers. I'll take a look and see what I can come up with.

jeddy3:
Done in https://github.com/stylelint/stylelint/pull/3113

PR #3113, opened 2018-01-08
Bugfix/issue#3105 selector max id nested at statements

BenjaminWFox:

#3105 

I haven't done a lot of these (edits/PRs) so I'm not sure if I missed something, but I couldn't find a scenario in the tests where the the check for 'atrule' was useful. Everything seemed to work as I expected when I removed it.


lib/rules/selector-max-id/__tests__/index.js
```diff
@@ -167,6 +167,26 @@ testRule(rule, {
...
       description: "media query: nested"
+    },
+    {
+      code: "#foo { @if ($p == 1) { #bar {} } }",
+      description: "@if statement: nested"
+    },
+    {
+      code: ".foo { @if ($p == 1) { #bar #baz {} } }",
+      description: "@if statement: nested compound selector"
+    },
```
hudochenkov: I think tests containing `@if` should go to `test` function with `syntax: "scss"`.

lib/rules/selector-max-id/__tests__/index.js
```diff
@@ -198,6 +218,52 @@ testRule(rule, {
...
+      column: 1
+    },
+    {
+      code: ".foo { @if ($p == 1) { #bar #baz #foo {} } }",
+      description:
+        "nested compound selector: greater than max id selectors, nested @if statement",
+      message: messages.expected(".foo #bar #baz #foo", 2),
+      line: 1,
+      column: 24
+    },
```
hudochenkov: I think tests containing `@if` should go to `test` function with `syntax: "scss"`.

Arcanemagus:
@BenjaminWFox You should either merge changes into this branch, or rebase the branch on the new changes. Cherry picking the commits from `master` into here just causes confusion in the log when this is merged unless all of these commits are squashed together.

BenjaminWFox:
@Arcanemagus Hmm, I thought I rebased but I haven't dealt with upstream/forked repos before - I probably got confused trying to get the right updates from the right place.

Should I just squash them then? And would that be with something like `git rebase -i <after-specific-commit>`? Thanks.

Arcanemagus:
Actually as @hudochenkov just pointed out to me in a different PR, _all_ PR's here are squashed so it's not a huge issue.

For your own reference though:

The simplest method is to make sure your local `master` matches the upstream `master`, then run `git rebase master`. That makes your commit(s) move to having the current `master` as their parent commit.

Alternatively you can also do `git merge master`, which creates a separate commit that merges the current state of `master` into your branch, leaving your existing commits untouched.

PR #3122, opened 2018-01-11
Fix selector-max-id tests

hudochenkov:

None.

Some tests are using incorrect config value. As it fixes tests, which were not run previously, #3113 might break.


hudochenkov:
Merged with one review as change is pretty trivial.

PR #3123, opened 2018-01-12
Add `jest:detectleaks` npm script for Jest memory leak testing.

ntwb:
Follow up to #3069/ #3121

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,6 +111,7 @@
     "dry-release": "npmpub  --dry --verbose",
     "flow": "flow",
     "jest": "jest",
+    "jest:detectleaks": "jest  --detectLeaks",
     "lint:js": "eslint . --cache",
     "lint:md": "remark . --quiet --frail",
     "lint": "npm-run-all --parallel lint:*",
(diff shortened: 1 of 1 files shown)


ntwb:
I've created #3124 to list, and seek help in fixing the current 17 tests that fail

hudochenkov:
What `weak` for? Is it peer or optional dependency of Jest?

ntwb:
My bad, I should have added a note about `weak`, Jest Leak Detector requires `weak`:
https://github.com/facebook/jest/blob/master/packages/jest-leak-detector/package.json#L14

hudochenkov:
Found it https://github.com/facebook/jest/blob/19376c1e1e8a77717fbb7a216df92087eddbd277/packages/jest-leak-detector/src/index.js#L39-L51 Quite strange approach for dependencies.

ntwb:
I think it's due to the _mono_ repo Facebook use:
• ~35 npm modules in a single repo https://github.com/facebook/jest/tree/master/packages
• Lerna is what's used to manage this repo setup https://lernajs.io/ - _"A tool for managing JavaScript projects with multiple packages."_

hudochenkov:
It shouldn't be because of Lerna. I use it at work and any package could have whatever dependencies it wants.

SimenB:
It's because of https://github.com/facebook/jest/issues/5169, fwiw

PR #3131, opened 2018-01-17
Add "/" as valid character after right parenthesis

dsine-de:
... to allow e.g. line-height shorthand value in "font" property:
font: calc(16px + .2vw)/1 sans-serif;

#3132

No, it's self explanatory.

### DIFF
--- a/lib/rules/function-whitespace-after/index.js
+++ b/lib/rules/function-whitespace-after/index.js
@@ -13,7 +13,7 @@ const messages = ruleMessages(ruleName, {
   rejected: 'Unexpected whitespace after ")"'
 });
 
-const ACCEPTABLE_AFTER_CLOSING_PAREN = new Set([")", ",", "}", ":", undefined]);
+const ACCEPTABLE_AFTER_CLOSING_PAREN = new Set([")", ",", "}", ":", "/", undefined]);
 
 const rule = function(expectation) {
   return (root, result) => {


ntwb:
Thanks for the pull request @dsine-de 👍 

Could you update the _"This rule does not check for space immediately after `)` if the very next character is `,`,` )`, or` }`, allowing some of the patterns exemplified below."_ section of the `function-whitespace-after` rule docs [readme.md](https://github.com/stylelint/stylelint/blob/master/lib/rules/function-whitespace-after/README.md) please.

This will also [need some tests](https://github.com/stylelint/stylelint/blob/master/lib/rules/function-whitespace-after/__tests__/index.js) covering both the `always` and `never` rule options

Arcanemagus:
A concern I had was that this may be valid for `font`, but this PR is making it valid for _any_ position after a `)`. Does that actually make sense? Or should this be limited to `font` somehow?

ntwb:
@Arcanemagus I'm also adding some comments to the #3132 issue, we should probably discuss these there.

jeddy3:
@dsine-de Are you able to add some tests to this PR so that we can merge it?

dsine-de:
I'll do that in the next few days!

evilebottnawi:
Closing due to inactivity in favor https://github.com/stylelint/stylelint/pull/3346

PR #3160, opened 2018-02-08
Add autofix to unit-case

hiroppy:

https://github.com/stylelint/stylelint/issues/2829

## lower
```
a { width: 10PIXEL; } -> a { width: 10pixel; }
a { width: calc(10PX * 2); } -> a { width: calc(10px * 2); }
```

## uppter
```
a { width: 10pixel; } -> a { width: 10PIXEL; }
a { width: calc(10px * 2); } -> a { width: calc(10PX * 2); }
```

### DIFF
--- a/lib/rules/unit-case/__tests__/index.js
+++ b/lib/rules/unit-case/__tests__/index.js
@@ -6,6 +6,7 @@ const { messages, ruleName } = rule;
 testRule(rule, {
   ruleName,
   config: ["lower"],
+  fix: true,
 
   accept: [
     {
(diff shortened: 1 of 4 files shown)


lib/rules/unit-case/__tests__/index.js
```diff
@@ -666,3 +666,51 @@ testRule(rule, {
...
+    },
+    {
+      code: "a { width: calc(10PX * 2); }",
+      fixed: "a { width: calc(10px * 2); }",
+      message: messages.expected("PX", "px")
+    }
+  ]
+});
+
+testRule(rule, {
```
evilebottnawi: need tests around scss/less/css variables

abouthiroppy: added tests.

jeddy3:
@hiroppy Friendly ping to see if you're still working on this PR?

lib/rules/unit-case/index.js
```diff
@@ -57,6 +58,23 @@ const rule = function(expectation) {
...
+
+            const res = `${stringStart}${expectedUnit}${stringEnd}`;
+
+            if (node.name === "media") {
+              node.params = res;
+            } else {
+              node.value = res;
+            }
+          });
+        }
```
hudochenkov: Move it before `report` and add `return` at the end of `if` statement. This way we won't show rule violation when we fixed the problem.

hiroppy:
@hudochenkov updated!

jeddy3:
Can we not use `postcss-value-parser` to update the AST, rather than use `stylesearch`?

hudochenkov:
@jeddy3 you mean, could we _use_ `postcss-value-parser` instead of `stylesearch`?

jeddy3:
Oops, yep... that wasn't very clear.

Yes, I mean can we use `postcss-value-parser` to update the AST?

jeddy3:
@hiroppy Do you think you'll be able to update this PR to use `postcss-value-parser` to update the AST, rather than use `stylesearch`. Updating the more granular values AST will likely be a more robust solution.

hiroppy:
@jeddy3 Thank you for the suggestion. OK, I will try using postcss-value-parser :). However, I am a little busy this week so please wait a moment.

yepninja:
Already solved #3725

PR #3161, opened 2018-02-09
Enable fix option for function-name-case

abouthiroppy:

https://github.com/stylelint/stylelint/issues/2829

### Problems
1. We can not search a function name which has `-` because `style-search` 
 does not include this pattern.(https://github.com/davidtheclark/style-search/blob/master/index.js#L182).

2. A function which is written by CamelCase is broken if a user does not specify `ignoreFunctions`.

Do we have any ideas?


abouthiroppy:
Maybe this is dangerous... so I closed. Thanks.

Issue #3144, opened 2018-01-25
Document postcss-flexbugs-fixes in declaration-block-no-redundant-longhand-properties


AlexWayfer:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A dangerous behavior (a bug, I guess, or false positive): `declaration-block-no-redundant-longhand-properties` triggers for `flex`, where `flex-basis` is `calc()`, [which doesn't work correctly in IE 10–11](https://github.com/philipwalton/flexbugs#flexbug-8).

> Which rule, if any, is this issue related to?

`declaration-block-no-redundant-longhand-properties`

> What CSS is needed to reproduce this issue?

e.g.

```css
.foo {
  flex-basis: calc(100% / 3);
  flex-grow: 1;
  flex-shrink: 0;
}
```

> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
  "rules": {
    "declaration-block-no-redundant-longhand-properties": true
  }
}
```

> Which version of stylelint are you using?

`8.4.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `yarn run stylelint --config .stylelintrc path/to/file.scss`"

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No. It's reproducable with regular CSS.

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

The following warnings were flagged:

```shell
 4:2  ✖  Expected shorthand property "flex"                                                                      declaration-block-no-redundant-longhand-properties
```

evilebottnawi:
@AlexWayfer Thanks for issue, i think we can add option and/or look in browserslist config to avoid this problem. If you don't want support IE10-11 rule works as expected.

AlexWayfer:
> i think we can add option and/or look in browserslist config to avoid this problem

Yes, I'm adding `IE 11` into `browserslist` with `stylelint-no-unsupported-browser-features` plugin and this issue is still here.

CAYdenberg:
I believe the rule is behaving as expected.

According to the specification, that CSS need not be written longhand, and the rule is correctly stating that. In effect, writing out these declarations is a hack to avoid IE's shoddy implementation of flexbox.

I think this could be avoided with e.g. `ignoreShorthands: ["/flex-/"]` - possibly we could expand documentation to suggest that?

AlexWayfer:
[...]

PR #3165, opened 2018-02-12
Expand documentation of declaration-block-no-redundant-longhand-prope…

CAYdenberg:
…rties to include discussion of flexbox browser bugs

#3144 

Documentation change only.

### DIFF
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
@@ -60,6 +60,10 @@ This rule complains when the following shorthand properties can be used:
 -   `text-emphasis`
 -   `mask`
 
+**Please note** that properties are considered to be redundant if they may be written shorthand according to the specification, **regardless of the behavior of any individual browser**. For example, due to Internet Explorer's implementation of Flexbox, [it may not be possible to use the shorthand property `flex`](https://github.com/philipwalton/flexbugs#flexbug-8), but the longhand form is still considered a violation. 
+
+Flexbox-related properties can be ignored using `ignoreShorthands: ["/flex-/"]` (see below).
+
 ## Options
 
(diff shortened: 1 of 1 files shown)


lib/rules/declaration-block-no-redundant-longhand-properties/README.md
```diff
@@ -60,6 +60,10 @@ This rule complains when the following shorthand properties can be used:
 -   `text-emphasis`
 -   `mask`
 
+**Please note** that properties are considered to be redundant if they may be written shorthand according to the specification, **regardless of the behavior of any individual browser**. For example, due to Internet Explorer's implementation of Flexbox, [it may not be possible to use the shorthand property `flex`](https://github.com/philipwalton/flexbugs#flexbug-8), but the longhand form is still considered a violation. 
+
+Flexbox-related properties can be ignored using `ignoreShorthands: ["/flex-/"]` (see below).
```
AlexWayfer: It can be just as `"flexbox"`, without `Regexp`, doesn't it?

AlexWayfer:
There is no information about [postcss-flexbugs-fixes](https://github.com/stylelint/stylelint/issues/3144#issuecomment-363911287) :pensive:

ntwb:
@AlexWayfer Can you create a pull request to add that?

lib/rules/declaration-block-no-redundant-longhand-properties/README.md
CAYdenberg: The documentation is a bit confusing as "padding" is given as an example, but the option is `ignoreShorthands`, and the shorthand properties are `flex-basis`, `flex-grow` and `flex-shrink`. I think I'd have to look at the code to figure out what the actual behaviour is. In any case despite being a regex, this seems more straightforward.

PR #3173, opened 2018-02-17
fix(package): update globby to version 8.0.1

m-allanson:

Replaces #3167


m-allanson:
Looks like the same error as in #3167. Closing this.

Issue #3101, opened 2018-01-04
Add ignore: ["first-nested"] to declaration-empty-line-before


jeddy3:
This option would be useful for when using prettier and stylelint side-by-side (see: https://github.com/prettier/prettier/issues/2621)

hudochenkov:
I don't see how it's related to Prettier issue? That issue is empty lines between rules.

jeddy3:
The same is true for empty lines between declarations. However, prettier will remove any empty lines before the first nested node. This option will allow a user to configure stylelint to ignore first nested declarations and, instead, let handle prettier handle those.

ntwb:
I really should take Prettier and stylelint for a spin, I've yet to try it with CSS or SCSS...

Issue #2211, opened 2016-12-25
Add except: ["after-same-name"] option to at-rule-empty-line-before


hudochenkov:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

I want to add `except: ["after-same-name"]` option to `at-rule-empty-line-before`.

> Which rule, if any, is this issue related to?

`at-rule-empty-line-before`

> What did you expect to happen?

Be able to separate at-rules with empty line, but have no empty lines between at-rules. It's currently possible only with blockless at-rules. I want to make this option work like:

```json
{
    "rule-nested-empty-line-before": ["always", {
        "except": ["after-rule"]
    }]
}
```

So with this config:

```json
{
    "at-rule-empty-line-before": ["always", {
        "except": ["after-same-name"]
    }]
}
```

The following pattern are considered warnings:

```css
a {
    display: block;
    
    @media (min-width: 100px) {}

    @media (min-width: 200px) {}
    @media (min-width: 300px) {}
}
```

```css
@import "_fonts.css";
@media (min-width: 100px) {}

@media (min-width: 200px) {}
```

The following pattern are not considered warnings:

```css
a {
    display: block;
    
    @media (min-width: 100px) {}
    @media (min-width: 200px) {}
    @media (min-width: 300px) {}
}
```

```css
@import "_fonts.css";

@media (min-width: 100px) {}
@media (min-width: 200px) {}
```

I can add this option if you'll approve it.

jeddy3:
@hudochenkov Thanks for the suggestion and for explaining it clearly.

SGTM. Feel free to contribute this option.

m-allanson:
@hudochenkov note that there are some [pending changes](https://github.com/stylelint/stylelint/pull/2213) to `at-rule-empty-line-before`. It might be worth waiting until they've landed in `master`.

hudochenkov:
@m-allanson I see you've changed a branch for that PR. Should I still wait for something?

m-allanson:
@hudochenkov Actually don't wait for anything, make the change whenever you like. I can rebase if necessary.

jeddy3:
Done in https://github.com/stylelint/stylelint/issues/2225

Issue #207, opened 2015-06-19
Adapt -except- rules and the indentation rule to use optional options objects


davidtheclark:
* [x] `rule-non-nested-empty-line-before`
* [x] `rule-nested-empty-line-before`
* [x] `at-rule-empty-line-before`
* [x] `comment-empty-line-before`
* [x] `indentation`

Same as https://github.com/stylelint/stylelint/issues/185 but for at-rules.

davidtheclark:
So now we run into a problem with the simple keyword option convention.

This rule already has the special `"always-except-blockless-group"` option. If we want to add an `"always-except-first"` option, as well, it seems we'd almost have to add a `"always-except-first-or-blockless-group"` option in case the user wanted both exceptions ... and that of course could continue indefinitely.

davidtheclark:
Can anybody see a way out of this? Or does it mean that we should have the first (I think) *object* of options? Something like ...

```
'at-rule-empty-line-before`: ['always', {
  ignoreBlocklessGroup: true,
  ignoreFirst: true
}]
```

davidtheclark:
I'm reminded we already have an object for `indentation`.

As an alternative to the above to avoid `ignore*` options, we could do

```js
'at-rule-empty-line-before`: [2, {
  default: "always",
  blocklessGroups: "never",
  first: "always"
}]
```

I don't know. Maybe.

jeddy3:
I realised the same thing yesterday.

I'm also not sure what the right solution is. Although, I think eslint's approach of an *additional and optional* option object might give us the necessary balance between flexibility and keeping things easy for the user (i.e. them not having to think about the optional options most of the time).

So, `at-rule-empty-line-before` might look something like:

```js
"at-rule-empty-line-before": [2, "always", {
  except: [
    "blockless-group",
    "first",
  ]
}]
```

"Except" means: *enforce the opposite* of the 1st option i.e. *never* an empty line before the first rule or rules in a blockless-group.

If that's too stringent we could also provide an `ignore` array within the optional options object e.g.

```js
"at-rule-empty-line-before": [2, "always", {
  except: ["blockless-group",],
  ignore: ["first"]
}]
```

i.e. always an empty line, except never in a blockless-group, and ignore the first node (so anything goes for that one). 

Another use of `ignore` would be in `comment-empty-line-before` e.g.

```js
"comment-empty-line-before": [2, "always", {
  ignore: ["inline"]
}]
```

[...]

PR #3179, opened 2018-02-19
Add ignore: ["first-nested"] options

### LINKED ISSUE #3178: Add ignore: ["first-nested"] to *-empty-line-before
The `"first-nested"` option was added to `custom-property-empty-line-before` and `declaration-empty-line-before` in 9.0.0. Whilst configuring my config for compatibility with prettier, I realised the same option is needed for `at-rule-empty-line-before` and `rule-empty-line-before` as prettier also formats the empty line before the first nested instances of these.

jeddy3:

Closes #3178 

No, it's self-explanatory.

### DIFF
--- a/lib/rules/rule-empty-line-before/README.md
+++ b/lib/rules/rule-empty-line-before/README.md
@@ -201,7 +201,7 @@ The following patterns are considered violations:
 ```scss
 a {
   color: red;
-  
+
   b {
     color: blue;
   }
@@ -247,7 +247,7 @@ The following patterns are *not* considered violations:
 }
 ```
 
-### `ignore: ["after-comment", "inside-block"]`
+### `ignore: ["after-comment", "first-nested", "inside-block"]`
 
 #### `"after-comment"`
 
@@ -260,6 +260,22 @@ The following patterns are *not* considered violations:
 ```css
 /* comment */
 a {}
+```
+
+#### `"after-comment"`
+
+Ignore rules that are nested and the first child of their parent node.
+
+For example, with `"always"`:
+
+The following patterns are *not* considered violations:
+
+```css
+@media {
+  a {}
+
+  b {}
+}
 ```
 
 #### `"inside-block"`
--- a/lib/rules/at-rule-empty-line-before/README.md
+++ b/lib/rules/at-rule-empty-line-before/README.md
@@ -242,7 +242,7 @@ b {
 }
 ```
 
-### `ignore: ["after-comment", "inside-block", "blockless-after-same-name-blockless", "blockless-after-blockless"]`
+### `ignore: ["after-comment", "first-nested", "inside-block", "blockless-after-same-name-blockless", "blockless-after-blockless"]`
 
 #### `"after-comment"`
 
@@ -267,6 +267,22 @@ The following patterns are *not* considered violations:
 @media {} /* comment */
 
 @media {}
+```
+
+#### `"first-nested"`
+
+Ignore at-rules that are nested and the first child of their parent node.
+
+For example, with `"always"`:
+
+The following patterns are *not* considered violations:
+
+```css
+@supports {
+  @media {}
+
+  @media {}
+}
 ```
 
 #### `"inside-block"`
--- a/lib/rules/at-rule-empty-line-before/__tests__/index.js
+++ b/lib/rules/at-rule-empty-line-before/__tests__/index.js
@@ -304,6 +304,31 @@ testRule(rule, {
   ]
 });
 
+testRule(rule, {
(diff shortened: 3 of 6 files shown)


jeddy3:
>  -   Added: `ignore: ["first-nested"]` to `at-rule-empty-line-before` ([#3179](https://github.com/stylelint/stylelint/pull/3179)).
>  -   Added: `ignore: ["first-nested"]` to `rule-empty-line-before` ([#3179](https://github.com/stylelint/stylelint/pull/3179)).

lib/rules/rule-empty-line-before/README.md
```diff
@@ -262,6 +262,22 @@ The following patterns are *not* considered violations:
 a {}
 ```
 
+#### `"after-comment"`
```
AlexWayfer: Wrong!

jeddy3: Good spot. Would you like to create to PR to fix this?

AlexWayfer:
What's the difference with existing [`except: ["first-nested"]` in `at-rule-empty-line-before`](https://stylelint.io/user-guide/rules/at-rule-empty-line-before/#first-nested) and [the same in `rule-empty-line-before`](https://stylelint.io/user-guide/rules/rule-empty-line-before/#first-nested)?

lib/rules/rule-empty-line-before/README.md
AlexWayfer: I think I can. But I'm still [not sure that this PR is necessary](https://github.com/stylelint/stylelint/pull/3179#issuecomment-367324023).

jeddy3:
From [here](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#secondary):

> - Use `"ignore"` when you want the rule to simply skip-over a particular pattern.
> - Use `"except"` when you want to invert the primary option for a particular pattern.

The `ignore` option is used when you don't want stylelint to enforce anything for that pattern.

AlexWayfer:
Thank you, @jeddy3.

Are there cases (examples) for these rules? I see the issue, but I don't see examples, just "let's make it here like in other rules".

jeddy3:
> I see the issue, but I don't see examples, just "let's make it here like in other rules".

An option is generally added to a rule when someone has made a specific request and stated their use case e.g. here's the issue that requests the [`except: ["after-same-name"]`](https://github.com/stylelint/stylelint/issues/2211) option for `at-rule-empty-line-before`.

> Are there cases (examples) for these rules? 

The use case of the new `ignore` options added in this PR can be found in the original issue (https://github.com/stylelint/stylelint/issues/3178) that this PR closes. It's a bit brief I'm afraid, so I understand the confusion. A more detailed use case [can be found in the issue](https://github.com/stylelint/stylelint/issues/3101#issuecomment-355338437) that requests the `ignore: ["first-nested"]` option for `declaration-empty-line-before` i.e. these options are useful if you use another tool, e.g. prettier, to pretty-print the empty-line before the first nested nodes. 

(I've added a link from the original issue to this more detailed use case to make it easier for future users to understand why these two new options were added).

> Small explanation: we have `except: ["first-nested"]` in our projects, and I'm not sure, should we switch to `ignore` or not.

If you want _stylelint to enforce_ a convention for the first nest node, then stick with `except: ["first-nested"]`. If you want to stylelint to ignore the first nest node because, perhaps, you use _another tool to enforce_ a convention, then use `ignore`.

AlexWayfer:
Thank you.

Why `except` option is not named as `invert`? :grimacing: 

I think, "enforce a convention" and "invert the primary option" are pretty different means.

I think I still can't understand `except` option, but I understood `ignore` option from the first time.

But OK, I think it's my bad. I understood that if you have `always` as primary option, then `except` will set `never` for listed cases, and vice-versa. And `ignore` doesn't set any pimary option for listed cases at all.

jeddy3:
> Why `except` option is not named as `invert`? 😬

We have to [go back a couple of years](https://github.com/stylelint/stylelint/issues/207) to answer that one... it looks like we borrowed the word from eslint. 

I think it still works well enough, though. Take the `indentation` rule for example:

```json
"indentation": [2, { "except": ["param"] } ]
```

i.e. indent by 2 spaces except for params.

Unlike words like "always" and "never", there isn't an intuitive inverted value for the number 2. So, "except" might not be the perfect word but it works well enough in a fair few contexts.

> But OK, I think it's my bad. I understood that if you have `always` as primary option, then `except` will set `never` for listed cases, and vice-versa. And `ignore` doesn't set any pimary option for listed cases at all.

Yes, that's correct.

PR #3180, opened 2018-02-19
Remove Open Collective postinstall message

m-allanson:

Comment at https://github.com/stylelint/stylelint/pull/2955#issuecomment-366766373

This removes the stylelint / Open Collective banner that appears after installing stylelint.

As stylelint is designed to be installed alongside other tools, we should probably err on the side of subtlety. What do you think @stylelint/core? I'm ok with this being closed if folks think we should keep the banner.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
     "meow": "^4.0.0",
     "micromatch": "^2.3.11",
     "normalize-selector": "^0.2.0",
-    "opencollective": "^1.0.3",
     "pify": "^3.0.0",
     "postcss": "^6.0.16",
     "postcss-html": "^0.12.0",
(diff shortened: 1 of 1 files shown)


ntwb:
Agreed, it's one big add, merging this PR as is to remove the Open Collective _postinstall_

I did explore an option of minimalising the post install, but I still didn't think it was ideal.

Let's keep what we have on the READMe.md as is and remove the _postinstall_ for now, we can revisit again down the road if others believe we should have something post `npm install`

----

If:
• `"postinstall": "opencollective postinstall"` from [package.json#L127](https://github.com/stylelint/stylelint/blob/master/package.json#L127) is changed to:
• `"postinstall": "opencollective postinstall stylelint --plain"`

And 
• `"logo": "https://opencollective.com/opencollective/logo.txt"` from [package.json#L200](https://github.com/stylelint/stylelint/blob/master/package.json#L200) is removed the resultant _postinstall_ is:

screen shot 2018-02-20 at 6 38 31 pm

ntwb:
FYI: I've got no idea why I never saw that _huge_ logo, I can't even open link it refers to, Cloudflare throws an error for me /shrug https://opencollective.com/opencollective/logo.txt

hudochenkov:
I didn't saw big logo as well.

PR #3193, opened 2018-03-02
Update downloads badge to point to graph of downloads over time 📈 instead of duplicating link to npm

cheapsteak:

"None, as it's a documentation update."

Hi! I noticed that your "npm" badge and "downloads" badge currently both point to the same page for the project on npmjs.org  

I was wondering if you might prefer having the download badge point to a [download chart](https://npmcharts.com/compare/stylelint?minimal=true) instead?

If not, feel free to close and apologies for the drive-by PR 😄


Arcanemagus:
[Looks like](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/pull-requests.md) documentation changes actually only require one approval, and don't need to be added to the changelog, so I merged this 😉.

PR #3201, opened 2018-03-07
Add link to stylelint-selector-tag-no-without-class plugin

aboks:

None, this is a small documentation addition.

No, it's self explanatory.

### DIFF
--- a/docs/user-guide/plugins.md
+++ b/docs/user-guide/plugins.md
@@ -15,4 +15,5 @@ Plugins are rules and sets of rules built by the community that support methodol
 -   [`stylelint-rscss`](https://github.com/rstacruz/stylelint-rscss): Validate [RSCSS](http://rscss.io) conventions.
 -   [`stylelint-scss`](https://github.com/kristerkari/stylelint-scss): Enforce a wide variety of SCSS-syntax specific linting rules (plugin pack).
 -   [`stylelint-selector-bem-pattern`](https://github.com/davidtheclark/stylelint-selector-bem-pattern): Specify a BEM pattern for selectors (incorporates [postcss-bem-linter](https://github.com/postcss/postcss-bem-linter)).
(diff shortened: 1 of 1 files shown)


hudochenkov:
Congratulations on your first stylelint plugin! :)

I took a quick look at it, and have one suggestion. I think it doesn't need a secondary option. Primary option could have an array. If array isn't empty, then the rule is enabled. You can take any `*-whitelist`/`*-blacklist` rule as an example. E. g. [`function-blacklist`](https://stylelint.io/user-guide/rules/function-blacklist/).

aboks:
Thanks for the suggestion. That option crossed my mind, but I was unsure whether an array as primary option would make sense when the rule would get (other) secondary options in the future (which I totally see happening). Would something like that be considered good practice? I couldn't find other examples of such situations, and the configuration looks a bit messy:
```js
"rule": [ ["tag1", "tag2", "tag3"], {
    "someOtherOption": true
} ]
```

hudochenkov:
It's ok.

[`unit-blacklist`](https://stylelint.io/user-guide/rules/unit-blacklist/):

```json
{
	"unit-blacklist": [
		["px", "em", "deg"],
		{
			"ignoreProperties": {
				"px": ["font-size", "/^border/"]
			}
		}
	]
}
```

Also, every rule could have [`severity`](https://stylelint.io/user-guide/configuration/#severities-error--warning) secondary option.

aboks:
Thank you for your feedback. I will aim to release a version 2.0 with the tags as the primary option.

It seems though that https://github.com/stylelint/stylelint-test-rule-tape does not work well with `primaryOptionArray` for plugins, and thus parses the configuration incorrectly. Are you aware of such a limitation?

ntwb:
@aboks Could you create an issue in that repo documenting what happens and what you expect to happen with some links to or code examples please :)

FYI: stylelint itself pretty much uses Jest for testing these days, so that is more than likely the reason we've not seen this issue before.

PR #3215, opened 2018-03-14
Add description of Sass to CLI help info.

gucong3000:

None, as it's a CLI help info fix.

No, it's self explanatory.


hudochenkov:
Maybe put it after “scss”, because “scss” is more popular.

gucong3000:
> Maybe put it after “scss”, because “scss” is more popular.

Done.

PR #3229, opened 2018-03-20
Fix false positives for non-CSS blocks in max-empty-lines

gucong3000:
> Which issue, if any, is this issue related to?

Closes #3221

> Is there anything in the PR that needs further explanation?

e.g. "No, it's self explanatory."

PR #3225, opened 2018-03-18
Fix false positives for at-root in indentation

AndyOGo:
> Which issue, if any, is this issue related to?

fixes #3168 

> Is there anything in the PR that needs further explanation?

No, it's self explanatory.

PR #3232, opened 2018-03-20
Fix git commit hook for Prettier

hudochenkov:
Change was made to precommit hook in #3184. This change wasn't needed.

People reported this problem: https://github.com/stylelint/stylelint/pull/3225#issuecomment-374129446.

Also, I see that Prettier wasn't run on a code: https://github.com/stylelint/stylelint/pull/3229/files#diff-c65b73210ee92c6286a9cee09c912513.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,7 +128,7 @@
   },
   "lint-staged": {
     "*.js": [
-      "npm run prettier --write",
+      "prettier --write",
       "git add"
     ]
   },


jeddy3:
Oh, it looks like both @ntwb and I misunderstood the changes in [lint-staged@7](https://github.com/okonet/lint-staged/releases). There is no "prettier" npm run-script, only "prettier:fix" and "prettier:check". So, it's kind of obvious now why it's failing.

hudochenkov:
To be honest I also didn't fully understood that change at first :)

ntwb:
Ha, thanks 👍

PR #3240, opened 2018-03-26
Fix false positives for non-CSS files in no-empty-source

### LINKED ISSUE #3233: Fix false positives for non-CSS files in no-empty-source

A bug in combination of [`no-empty-source`](https://github.com/stylelint/stylelint/blob/9.1.3/lib/rules/no-empty-source/README.md) rule and non-CSS files for example HTML and Markdown

`no-empty-source`

An HTML, Markdown or Vue file with only white spaces.

e.g.

```js
module.exports = {
  rules: {
    'no-empty-source': true
  }
};
```

e.g. `9.1.3`

API, but probably the problem is reproducible with any interface.

No.

```js
'use strict';

const {lint} = require('stylelint'); // v9.1.3

(async () => {
  const {results} = await lint({
    code: '',
    codeFilename: 'empty.html',
    config: {
      rules: {
        'no-empty-source': true
      }
    }
  });

  console.log(results[0].warnings[0]);
})();
```

When I run this code, I expect the result should contain no warnings about `no-empty-source`.

The code above prints the following:

```js
{ line: 1,
  column: 1,
  rule: 'no-empty-source',
  severity: 'error',
  text: 'Unexpected empty source (no-empty-source)' }
```

This warning seems valid at first glance, because `example.html` is actually a zero-byte file and `no-empty-source` warning looks appropriate for it. However, as the file is not CSS and has no `<style>` tags, I think stylelint should not report any errors about it.

If users really want to invalidate empty HTML, they can use a HTML linter separately and wouldn't expect stylelint to warn about it.

gucong3000:

Closes #3233

No, it's self explanatory.

### DIFF
--- a/lib/rules/max-empty-lines/index.js
+++ b/lib/rules/max-empty-lines/index.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const _ = require("lodash");
+const eachRoot = require("../../utils/eachRoot");
 const optionsMatches = require("../../utils/optionsMatches");
 const report = require("../../utils/report");
 const ruleMessages = require("../../utils/ruleMessages");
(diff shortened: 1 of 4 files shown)


jeddy3:
> -   Fixed: `no-empty-source` false positives for non-CSS blocks ([#3240](https://github.com/stylelint/stylelint/pull/3240)).

Issue #3236, opened 2018-03-22
Release 9.2


jeddy3:
Quite a few fixes are just about to land.

It'd be great to see the following get in the next release too:

- ~https://github.com/stylelint/stylelint/pull/3195~
- ~https://github.com/stylelint/stylelint/pull/3194~
- ~https://github.com/stylelint/stylelint/pull/3203~
- ~https://github.com/stylelint/stylelint/pull/3176~
- ~https://github.com/stylelint/stylelint/pull/2903~

Any chance of reviews, tweaks on those if we have time?

- [x] stylelint release
- [x] stylelint-config-recommended update/release
- [x] stylelint-config-standard update/release
- [x] stylelint-demo update
- [x] stylelint.io update
- [x] tweet

cc @stylelint/core

jeddy3:
@gucong3000 I'm getting a failed test when trying to do a release on my local machine:

<img width="757" alt="screen shot 2018-03-25 at 23 38 34" src="https://user-images.githubusercontent.com/808227/37880819-22481e22-3086-11e8-96d5-009c9a9d36c0.png">

Am I right in thinking that this is related to this recently merged PR?: https://github.com/stylelint/stylelint/pull/2903

Is there anything we can do to either speed up this test or workaround the timeout period?

ntwb:
I'm also seeing a few slow tests when running `npm test`:

```shell
 PASS  lib/rules/color-hex-length/__tests__/index.js (5.95s)
 PASS  lib/rules/no-duplicate-selectors/__tests__/index.js (6.861s)
 PASS  lib/rules/declaration-property-unit-whitelist/__tests__/index.js (5.304s)
 PASS  lib/rules/at-rule-semicolon-space-before/__tests__/index.js (5.289s)
 PASS  system-tests/001/001.test.js (5.602s)
```

The test `FAIL` you posted @jeddy3 worked for me in that above test run:
```shell
 PASS  system-tests/cli/cli.test.js
```

jeddy3:
Timing out tests in https://github.com/stylelint/stylelint/pull/3241 too.

@stylelint/core Any ideas what might have changed recently to have caused this?

gucong3000:
@jeddy3
> I'm getting a failed test when trying to do a release on my local machine:

Could you try to modify this code, remove `if (parseInt(process.versions.node) < 7)`
[...]

PR #3242, opened 2018-03-27
Fix require (and test) performance

jeddy3:
Ref #3236 
Ref #2454

### DIFF
--- a/lib/requireRule.js
+++ b/lib/requireRule.js
@@ -1,0 +1,12 @@
+"use strict";
+
+const importLazy = require("import-lazy")(require);
+const rules = require("./rules");
+
+module.exports = function(ruleName) {
+  if (rules.indexOf(ruleName) !== -1) {
+    return importLazy(`./rules/${ruleName}`);
+  } else {
+    return false;
+  }
+};
--- a/lib/requireRule.js
+++ b/lib/requireRule.js
@@ -4,7 +4,7 @@ const importLazy = require("import-lazy")(require);
 const rules = require("./rules");
 
 module.exports = function(ruleName) {
-  if (rules.indexOf(ruleName) !== -1) {
+  if (rules.includes(ruleName)) {
     return importLazy(`./rules/${ruleName}`);
   } else {
     return false;
--- a/lib/utils/checkAgainstRule.js
+++ b/lib/utils/checkAgainstRule.js
@@ -2,6 +2,7 @@
 "use strict";
 
 const normalizeRuleSettings = require("../normalizeRuleSettings");
+const requireRule = require("../requireRule");
 const Result = require("postcss/lib/result");
 const rules = require("../rules");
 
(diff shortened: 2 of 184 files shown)


jeddy3:
These changes take about a 3rd off the test run times on the CIs and on my local machine:

- Travis Node 8 (with coverage) 204.411s -> 166.244s
- Travis Node 6: 189.027s -> 133.654s
- Appveyor: 197.769s -> 116.166s
- Local machine (with cleaned jest cache): 248.824s -> 82.653s

These changes also reduce the `require` time of stylelint from 900ms to 600ms on my machine.

I've done a fair amount of investigation in the past two days and I've narrowed down two other possible areas of improvements:

1. Requiring `postcss-html` in `getPostcssResult.js` is responsible for nearly half of the remaining require time. Without `postcss-html` the require time is about 300ms. Which I believe is on par with ESLint. Additionally, running the tests only takes 54s on my local machine without `postcss-html`. @gucong3000 Any thoughts on what we can do about this?
2. It's difficult to pinpoint memory leaks, but there might be one in `isAutoprefixable.js`. The `*-no-vendor-prefix` rules popped up regularly when using `jest:detectleaks` and removing the `require('autoprefixer')` line from `isAutoprefixable.js` seem to quell them. Anyone got any ideas what we could do here?

@stylelint/core Can anyone see any potential problems with the changes in this PR? If people have time, it'd be great to get this PR tested against your project code bases.

gucong3000:
> Any thoughts on what we can do about this?
#3243

jeddy3:
> Any thoughts on what we can do about this?
> #3243

Awesome, thanks! I've merged that and rebased this PR. The change seems to have shaved off _another third_.

Travis Node 8 (with coverage: 204.411s -> 166.244s -> **77.806s** 
Travis Node 6: 189.027s -> 133.654s -> **99.038s**
Appveyor: 197.769s -> 116.166s -> **76.075s**

On my machine, tests (with cleaned Jest cache) are now down to **54s** and require time is around **344ms** (down from 900ms -> 600ms).

I'm removing the "WIP" from the title. I believe this PR is ready to review and test.

lib/utils/checkAgainstRule.js
```diff
@@ -22,7 +23,7 @@ module.exports = function(
   if (!callback) throw new Error("checkAgainstRule requires a callback");
   if (!options.ruleName)
     throw new Error("checkAgainstRule requires a 'ruleName' option");
-  if (!rules[options.ruleName])
+  if (rules.indexOf(options.ruleName) === -1)
```
Arcanemagus: I'd suggest `(!rules.includes(options.ruleName))` as a clearer alternative.

lib/requireRule.js
```diff
@@ -0,0 +1,12 @@
+"use strict";
+
+const importLazy = require("import-lazy")(require);
+const rules = require("./rules");
+
+module.exports = function(ruleName) {
+  if (rules.indexOf(ruleName) !== -1) {
```
Arcanemagus: I'd suggest `(rules.includes(ruleName))` as a clearer alternative.

lib/utils/checkAgainstRule.js
jeddy3: For something reason, I'd gotten it into my head that `.includes` was introduced in node 8... [that's not the case](https://node.green/#ES2016-features-Array-prototype-includes).

I'll update now.

jeddy3:
> -   Fixed: slow `require('stylelint')` time ([#3242](https://github.com/stylelint/stylelint/pull/3242)).

PR #3258, opened 2018-04-09
Remove files with syntax errors from cache

### LINKED ISSUE #3254: Fix cache option hiding CssSyntaxError outputs
When a css file contains syntax errors like:

```css
a {
  {color: red;
}
```

When `cache` option is set to `true`, stylelint will report a `CssSyntaxError` first time it runs, however starting from the second time, when css file is not modified, `CssSyntaxError` disappears from output

otakustay:
Close #3254

I think we should remove files from cache on any types of error, not only `CssSyntaxError`, however I don't know how to create other types of error, so the test case only covers `CssSyntaxError`

### DIFF
--- a/lib/__tests__/standalone-cache.test.js
+++ b/lib/__tests__/standalone-cache.test.js
@@ -13,6 +13,7 @@ const fileExists = require("file-exists-promise");
 
 const cwd = process.cwd();
 const invalidFile = path.join(fixturesPath, "empty-block.css");
+const syntaxErrorFile = path.join(fixturesPath, "syntax_error.css");
 const validFile = path.join(fixturesPath, "cache", "valid.css");
 const newFileDest = path.join(fixturesPath, "cache", "newFile.css");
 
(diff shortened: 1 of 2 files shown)


CAYdenberg:
The code LGTM and I'll approve: just clarify: what other types of errors would we be concerned with? If the file doesn't exist it won't be cached (you could write a test for this I suppose?) - if an error occurs in stylelint it's on our end and we should fix it. I'm sure I'm just not thinking of something ...

jeddy3:
> The code LGTM and I'll approve just clarify: what other types of errors would we be concerned with?

Do we need this clarification to approve and merge, or can we just go ahead?

CAYdenberg:
That's fair enough - if we think of something we can just add it later.

CAYdenberg:
@ntwb @jeddy3 Sorry guys I've been out of the loop. Are we still updating the CHANGELOG on each merge to master? It now says I can't commit directly. I apologize for the breach of protocol here. Just let me know what I should do.

hudochenkov:
@CAYdenberg yes. Try to edit changelog via Github interface. This is how it's intended to be edited.

Issue #3252, opened 2018-04-04
Fix compatibility with stylelint-processor-styled-components


yoannmoinet:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Might be a bug with the latest 9.2.0.

> Which rule, if any, is this issue related to?

A specific processor, [stylelint-processor-styled-components](https://github.com/styled-components/stylelint-processor-styled-components) (the only one I'm using) isn't working after 9.1.3.

> What CSS is needed to reproduce this issue?

It will have to be a css using a processor, in my case a simple js file using [styled-component](https://github.com/styled-components/styled-components)  will do : 

```javascript
// test.js
import styled from 'styled-component';

export default styled.div`
.test {
background:black;
}
`;
```

> What stylelint configuration is needed to reproduce this issue?

Any config as long as it's using a processor.

```json
// config.json
{
  "processors": ["stylelint-processor-styled-components"],
  "extends": [
    "stylelint-config-standard",
    "stylelint-config-styled-components"
  ]
}
```

> Which version of stylelint are you using?

`9.2.0` not working
`9.1.3` working

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylelint --config config.json test.js`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes, it's using styled-components.

> What did you expect to happen?

Having errors written in the terminal.

> What actually happened (e.g. what warnings or errors you are getting)?

No output at all.

I tried the exact same configuration with `9.1.3` and I got errors in the terminal.
Nothing with `9.2.0`.

I'm not sure if it's coming from `stylelint` or `stylelint-processor-styled-components`] so I opened an error in both with as much info I can get. 
styled-components/stylelint-processor-styled-components#180
Hope that will help.

ntwb:
> I believe we need to resolve what the roles of `postcss-html` and processors are. It looks like [changes are happening](https://github.com/stylelint/stylelint/issues/3248#issuecomment-378213074) in `postcss-html` that break existing processors.

[...]

Issue #3248, opened 2018-04-01
Reduce install size


jeddy3:
Ref: https://github.com/stylelint/stylelint/issues/3236

The size of the install has increased significantly (and the number of packages has nearly doubled) between `8.4.0` and `9.2.0`. I believe partly because `postcss-html` [bundles a couple of parsers](https://github.com/gucong3000/postcss-html/blob/10b59ccd92d1c16ca0c06e455074200784e44334/package.json#L48-L53): 

Difference between `8.4.0` and `9.2.0`:
```
+ stylelint@8.4.0
added 260 packages in 12.919s
jeddy3:stylelint-test jeddy3$ du -sh node_modules
 24M    node_modules
```

```
+ stylelint@9.2.0
added 418 packages in 22.232s
jeddy3:stylelint-test jeddy3$ du -sh node_modules
 33M    node_modules
```

For comparsion, here's ESLint:

```
+ eslint@4.19.1
added 141 packages in 8.82s
jeddy3:stylelint-test jeddy3$ du -sh node_modules
 19M	node_modules
```

Is there anything we can do about this? @gucong3000 Is it worth optimising `postcss-html` for `<style>` tags and then making markdown and template literal parsing opt-ins? Am I right in thinking it's just the parsing of `<style>` which is most useful for Vue and Web Component authors?

gucong3000:
With the development of `postcss-html`, I changed its original design. Maybe I should find the right time to change its name. I used `postcss-html` for split file (not just html) for `postcss` syntax parsing

jeddy3:
@gucong3000 Thanks for splitting the syntaxes into separate packages.

Once `9.2.1` is released, we can assess the new install size of stylelint. We can then discuss:

1. which syntaxes we should bundle with stylelint _by default_
2. which syntaxes should be used via the `--custom-syntax` flag

for `stylelint@10.0`.

jeddy3:
> I created a new package to solve: postcss-syntax (No dependencies)

I'll close this issue then.

We can move the discussion what about should be included by default to https://github.com/stylelint/stylelint/issues/3264

PR #3261, opened 2018-04-13
Fix support for processors

gucong3000:

Closes: #3252
Closes: https://github.com/styled-components/stylelint-processor-styled-components/issues/180

No, it's self explanatory.


docs/user-guide/processors.md
```diff
@@ -13,6 +13,11 @@ Processors are community packages that enable stylelint to lint the CSS within n
 
 ## PostCSS syntax
```
jeddy3: Let's just remove this whole section for now. This is the `processors.md` file. Once we resolve the relationship between processors and syntax we can revisit this.

docs/user-guide/css-processors.md
```diff
@@ -17,10 +17,10 @@ By default, the linter can *parse* any the following non-standard syntaxes by us
 -   SCSS (using [`postcss-scss`](https://github.com/postcss/postcss-scss))
 -   Less (using [`postcss-less`](https://github.com/shellscape/postcss-less))
 -   SugarSS (using [`sugarss`](https://github.com/postcss/sugarss))
--   HTML,
```
jeddy3: Let's remove "Javascript or typescript" from this list for now as it's an undocumented feature. And let's remove "Markdown" and "HTML-like" as we're still resolving the processor vs syntax issue.

jeddy3:
@gucong3000 Many thanks for starting on this!

Also, thanks for [splitting the parsers into their own packages](https://github.com/stylelint/stylelint/issues/3248#issuecomment-381385995). I'm confident this will help down the line.

Sorry for the delay reviewing this PR. The changes are far reaching and there is a lot to understand. 

I've tested this branch with all the [processors listed in the docs](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/processors.md):

- [x] [@mapbox/stylelint-processor-arbitrary-tags](https://github.com/mapbox/stylelint-processor-arbitrary-tags)
- [x] [stylelint-processor-glamorous](https://github.com/zabute/stylelint-processor-glamorous)
- [ ] [@mapbox/stylelint-processor-markdown](https://github.com/mapbox/stylelint-processor-markdown)
- [x] [stylelint-processor-styled-components](https://github.com/styled-components/stylelint-processor-styled-components)

The good news is most, including `stylelint-processor-styled-components`, are compatible with this branch. _However, `@mapbox/stylelint-processor-markdown` is not._ 

Given the following config:

```json
{
  "processors": ["@mapbox/stylelint-processor-markdown"],
  "extends": [
    "stylelint-config-standard"
  ]
}
```

And the following markdown:

````
Uh oh, I'm about to use SCSS.

```scss
// Parser-breaking comment
$foo: bar;
.foo {}
```
````

The following error is thrown:

```console
jeddy3:stylelint-test jeddy3$ ./node_modules/.bin/stylelint "sources/md/scss.*" --config configs/markdown.json
TypeError: Cannot read property 'line' of undefined
    at result.warnings.reduce (/Users/jeddy3/Projects/stylelint-test/node_modules/@mapbox/stylelint-processor-markdown/index.js:54:39)
    at Array.reduce (<anonymous>)
    at transformResult (/Users/jeddy3/Projects/stylelint-test/node_modules/@mapbox/stylelint-processor-markdown/index.js:51:39)
    at config.resultProcessors.forEach.resultProcessor (/Users/jeddy3/Projects/stylelint/lib/createStylelintResult.js:174:26)
    at Array.forEach (<anonymous>)
    at stylelint.getConfigForFile.then (/Users/jeddy3/Projects/stylelint/lib/createStylelintResult.js:171:31)
    at <anonymous>
```

It is the `scss` (or `less`or `sass`) on the code fence that breaks it. Using `css` is compatible.

**This _regression_ in this branch will need to be fixed before we can merge this PR.**

Then, after a patch is released, I believe stylelint will once again be compatible with the existing processors. We can then, as we prepare for `stylelint@10.0`, carefully deliberate on:

1. the role of processors and syntax
2. what features (e.g. markdown, `<style>`, styled components) stylelint should support _by default_.

gucong3000:
> However, @mapbox/stylelint-processor-markdown is not.

Still fail when I remove `syntax` from postcssOptions:
https://github.com/stylelint/stylelint/blob/issue-3248/lib/getPostcssResult.js#L92

jeddy3:
@gucong3000 Thanks for making the changes and for looking into the markdown processor regression. Sorry for the slow response. I'm just catching up on 2 weeks of stylelint issues.

Ok, I think we should merge this PR and get the release out. It's more important to fix the styled-components processor regression because users might be relying on the interpolation tagging feature of the processor. Anyone using the markdown-processor (with non-standard syntax code fences) can simply remove processor and I believe they'll get the same functionality. Sound good?

stramel:
@jeddy3 any update on this?

jeddy3:
It would be great to see this merged.  @stylelint/core A second pair of eyes on it would be great, though.

If no one has time, I'll try to make some to test again this evening and get this merged.

hudochenkov:
I'm not sure what this PR does in general. Does it introduce breaking change?

Code looks good.

lib/getPostcssResult.js
```diff
@@ -1,28 +1,29 @@
...
-  scss,
-  sss: sugarss,
-  sugarss
+  }
 };
+[
+  "less",
+  "sass",
+  "scss",
+  "markdown",
```
ntwb: `markdown` was being removed wasn't it?

package.json
```diff
@@ -62,15 +62,17 @@
     "normalize-selector": "^0.2.0",
     "pify": "^3.0.0",
     "postcss": "^6.0.16",
-    "postcss-html": "^0.18.0",
+    "postcss-html": "^0.23.6",
     "postcss-less": "^1.1.5",
+    "postcss-markdown": "^0.23.6",
```
ntwb: Likewise, wasn't `markdown` being removed for now, in this PR and to be added back in a PR of its own?

package.json
```diff
@@ -113,7 +115,7 @@
     "precommit": "lint-staged",
     "benchmark-rule": "node scripts/benchmark-rule.js",
     "dry-release": "npmpub  --dry --verbose",
-    "flow": "flow",
+    "flow": "echo flow",
```
ntwb: Why `echo` the `flow` word instead of running `flow`?

package.json
gucong3000: Previously Markdown's processing logic was written in `postcss-html`, and now it is split into multiple packages: `postcss-syntax`, `postcss-html` and `postcss-markdown`

lib/getPostcssResult.js
gucong3000: `postcss-markdown` split from the original `postcss-html`

jeddy3:
> -   Fixed: regression with processors (e.g. styled-components) ([#3261](https://github.com/stylelint/stylelint/pull/3261)).

Issue #3230, opened 2018-03-20
Fix false positives for nested properties


Grawl:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug when Sass indented properties cause all selectors below as warnings.

> Which rule, if any, is this issue related to?

```yml
  no-duplicate-selectors: true
  indentation: tab
```

> What CSS is needed to reproduce this issue?

```sass
.foo
	a
		color: white
	background:
		color: black // this one
	%placeholder
		background: hsla(0, 0%, 79%, 0.2)
	&-bar
		color: white
.baz
	@extend %placeholder
	.qux
		color: red
```

> What stylelint configuration is needed to reproduce this issue?

```yml
rules:
  no-duplicate-selectors: true
  indentation: tab
```

> Which version of stylelint are you using?

```
➜ stylelint -v
9.1.3
```

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylelint path/to/stylesheet.sass`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Sass `.scss` indented syntax

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

Looks like every selector after indented property is marked as warning.

```
  6:1  ✖  Unexpected duplicate selector ".foo %placeholder ", first used at line 6   no-duplicate-selectors
  7:3  ✖  Expected indentation of 3 tabs                                             indentation
  8:1  ✖  Unexpected duplicate selector ".foo-bar ", first used at line 8            no-duplicate-selectors
  9:3  ✖  Expected indentation of 3 tabs                                             indentation
 10:1  ✖  Unexpected duplicate selector ".baz ", first used at line 10               no-duplicate-selectors
 12:1  ✖  Unexpected duplicate selector ".baz  .qux ", first used at line 12         no-duplicate-selectors
 12:2  ✖  Expected indentation of 2 tabs                                             indentation
 13:3  ✖  Expected indentation of 4 tabs                                             indentation
```

If I replace indented property to a regular one, the problem disappears

```sass
.foo
	a
		color: white
	background-color: black // this one
	%placeholder
		background: hsla(0, 0%, 79%, 0.2)
	&-bar
		color: white
.baz
	@extend %placeholder
	.qux
		color: red
```

jeddy3:
@Grawl Thanks for the detailed report and for using the template. 

[...]

Issue #3130, opened 2018-01-16
Consistently check that selectors are standard before passing to the parser


k0nserv:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Updating to 8.4.0 breaks for us.

> What CSS is needed to reproduce this issue?

```scss
@mixin bpk-locale($locale) {
  html[lang=#{$locale}] &,
  html[lang=#{to_lower_case($locale)}] & {
    @content;
  }
}
```

https://github.com/Skyscanner/backpack/blob/3c53ec780ff922a727923b796be540637ce10e98/packages/bpk-mixins/src/mixins/_utils.scss#L165-L170

> Fails with the following error

```
 166:3  ✖  Cannot parse selector   parseError
```

> Which version of stylelint are you using?

8.4.0
> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

SCSS

> What did you expect to happen?

The linting not to fail

jeddy3:
@k0nserv Thanks for the report and for using the template.

Are you able to paste your config? I believe a particular rule is triggering this error. Help narrowing down which one it could be would be appreciated.

k0nserv:
Hey, thanks for getting back.

This is the [config](https://github.com/Skyscanner/backpack/blob/master/.stylelintrc.json) which extends https://github.com/Skyscanner/stylelint-config-skyscanner/blob/master/index.js

k0nserv:
Is there a way to get verbose output from stylelint? I didn't see any options for it, but I'd be happy to provide more information on the failure if required

jeddy3:
Wow, that's quite the config. It's great to see so many rules and plugins get used.

I did a little digging and there's quite a lot going on here. The following code, i.e. an _unquoted_ attribute selector with non-standard variable interpolation (_containing a function_) and followed by the parent selector, is not parseable postcss-selector-parser:

```scss
a[attribute=#{function($value)}] & {}
```

We normally check that [selectors conform to standard syntax](https://github.com/stylelint/stylelint/blob/master/lib/utils/isStandardSyntaxSelector.js) before passing them along to postcss-selector-parser. This is _not_ happening in numerous rules:

- `selector-attribute-brackets-space-inside`
- `selector-attribute-operator-space-after`
- `selector-attribute-operator-space-before`
- `selector-combinator-space-after`
- `selector-combinator-space-before`
[...]

Issue #2642, opened 2017-06-15
Poor linting of nested properties in SCSS


clarcharr:
The below code messes with the parser:

```
margin: auto {
  top: $left;
  bottom: $right;
}
```

This is valid in SCSS, so, the parser shouldn't emit parsing errors here.

ntwb:
You'll need to use [stylelint-scss](https://github.com/kristerkari/stylelint-scss)

Also GitHub issues are for stylelint bugs and enhancements, if you're looking for help or support with stylelint then stackoverflow is our preferred question and answer forum - http://stackoverflow.com/questions/tagged/stylelint

PR #3283, opened 2018-04-28
Fix nested property scss syntax false positive

brneto:

issues #3230 #3130 #2642  

 The problem was in the `no-descending-specificity` rule. When it was calling PostCSS API method `root.walkRules` it was treating `padding:` as a selector and passing it to parseSelector directly, which was throwing the error `✖ Cannot parse selector parseError`.

### DIFF
--- a/lib/utils/__tests__/isNestedProperty.test.js
+++ b/lib/utils/__tests__/isNestedProperty.test.js
@@ -10,7 +10,7 @@ describe("isNestedProperty", () => {
     });
   });
 
-  it("rejects nested property", () => {
+  it("rejects not nested property", () => {
     return nestedProperty("foo: red;", nestedProperty => {
       expect(isNestedProperty(nestedProperty)).toBeFalsy();
     });
--- a/lib/utils/__tests__/isNestedProperty.test.js
+++ b/lib/utils/__tests__/isNestedProperty.test.js
@@ -1,0 +1,26 @@
+"use strict";
+
+const isNestedProperty = require("../isNestedProperty");
+const postcss = require("postcss");
+
+describe("isNestedProperty", () => {
+  it("accepts nested property", () => {
+    return nestedProperty("foo: {};", nestedProperty => {
+      expect(isNestedProperty(nestedProperty)).toBeTruthy();
+    });
+  });
+
+  it("rejects nested property", () => {
+    return nestedProperty("foo: red;", nestedProperty => {
+      expect(isNestedProperty(nestedProperty)).toBeFalsy();
+    });
+  });
+});
+
+function nestedProperty(css, cb) {
+  return postcss()
+    .process(css, { from: undefined })
+    .then(result => {
+      result.root.walk(cb);
+    });
+}
--- a/lib/rules/no-descending-specificity/__tests__/index.js
+++ b/lib/rules/no-descending-specificity/__tests__/index.js
@@ -76,6 +76,9 @@ testRule(rule, {
     },
     {
       code: ":root { --foo: {}; }"
+    },
+    {
+      code: ":root { foo: {}; }"
     }
   ],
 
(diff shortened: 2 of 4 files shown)


brneto:
Yes, feel free to do that!

evilebottnawi:
@brneto i can't do this in your fork/branch :smile:

brneto:
Give me a little help on that, please! Where should be the right folder to put this test lib/utils/__test__ or  lib/__test__? Should I create a new file or used an existent? If I need to use an existent which would be that?

hudochenkov:
@brneto  Create new file `lib/utils/__tests__/isNestedProperty.test.js`. I think you can use [`isCustomPropertySet.test.js`](https://github.com/stylelint/stylelint/blob/master/lib/utils/__tests__/isCustomPropertySet.test.js) as a reference for how your test will look like.

brneto:
I've just created the test for isNestedProperty, but how about the test for `no-descending-specificity` rule? Did you use to create a test for each rule? If so, in which file is it?

brneto:
I've created the test for `no-descending-specificity` rule.  There are any other tests I need to create?

lib/utils/__tests__/isNestedProperty.test.js
```diff
@@ -0,0 +1,26 @@
...
+const postcss = require("postcss");
+
+describe("isNestedProperty", () => {
+  it("accepts nested property", () => {
+    return nestedProperty("foo: {};", nestedProperty => {
+      expect(isNestedProperty(nestedProperty)).toBeTruthy();
+    });
+  });
+
+  it("rejects nested property", () => {
```
evilebottnawi: rejects **not** nested property

brneto: But I have done that, no?! If `foo: {}`, is a nested property. If `foo: red`, isn't a nested property!

Pardon me, if I don't understand your request, I'm a little new in contributing to projects!

brneto: I get it no! Consider done!

evilebottnawi: test inside need describe what happens, i.e. `should be rejects not nested properties`

brneto: @evilebottnawi Sorry, my misunderstanding! On a second read, I got what you were meaning!

Thanks very much for all the help!!

brneto:
Very thanks you all! I happy to have my first contribution done and accepted!!! 😃

brneto:
When are you planning delivery the version with this fix?

evilebottnawi:
@brneto just add `Fixed: description` to CHANGELOG into `HEAD` section and we can merge :+1:

hudochenkov:
@evilebottnawi users shouldn't update changelog. Maintainers do. See the third point https://stylelint.io/developer-guide/pull-requests/.

evilebottnawi:
@hudochenkov oh, sorry, you are right

hudochenkov:
> -   Fixed: `no-descending-specificity` false positives for Sass nested properties ([#3283](https://github.com/stylelint/stylelint/pull/3283)).

brneto:
I think you right! I don't know very well all the utils, that why I did that!

I'm going to change what you said and send a new PR, ok?

hudochenkov:
> So, I'm not surprised the isStandardSyntaxRule util might have been missed in this instance :)

😌

> I'm going to change what you said and send a new PR, ok?

@brneto It would be great!

brneto:
When are the team planning to release a new version of this linter?

jeddy3:
> When are the team planning to release a new version of this linter?

`9.2.1` was just released with this fix in it. Sorry for the delay. There was [tricky regression](https://github.com/stylelint/stylelint/issues/3252) that [took some time](https://github.com/stylelint/stylelint/pull/3261) to work through.

brneto:
Thanks very much!

Issue #3114, opened 2018-01-08
Add linebreaks


bcullman:
Request new rule, similar to https://eslint.org/docs/rules/linebreak-style, `CRLF` or `LF` (aka windows, or unix). I believe this would fall under `https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#general--sheet-2`

> Which version of stylelint are you using?

`^8.0.0`

> How are you running stylelint: 

"CLI with `stylelint './less/**/*.less'`"

> What did you expect to happen?

Would like to throw errors if a line ending style other than what is specified in the `.stylelintrc` is used.

jeddy3:
@bcullman Thanks for the request.

I'm surprised this hasn't come up before. @stylelint/core Any objections to such a rule? 

`"linebreaks":  "unix"|"windows"`?

ntwb:
No objections here, what I will say is that we should include similar notes to that of which ESlint include on setting this using VCS https://eslint.org/docs/rules/linebreak-style#using-this-rule-with-version-control-systems.

jeddy3:
As there no objections, I'll label this issue up accordingly.

@bcullman Feel free to [contribute](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#creating-a-new-rule) this rule.

- Name: `linebreaks`
- Primary option: `"unix"|"windows"`
- Secondary options: none
- Message: "Expected linebreak to be ${linebreak}"
- Section: "Stylistic -> General"

saulotoledo:
Hi. If no one is working on this one, I will do do the job.

bcullman:
hi @saulotoledo - nope have not had time to look into how to implement. feel free to take

saulotoledo:
Thanks @bcullman! I will work on it then.

saulotoledo:
Hi @jeddy3. I have a first version of the rule. It still needs some work and new test cases (as looking for newlines in the end of the file, for example). I plan to work on it this weekend (or before, if possible), but I would like you or the community to take a look if possible. May I open the MR as a WIP?

saulotoledo:
Please take a look at https://github.com/stylelint/stylelint/pull/3289 when possible. It still needs some work. I will work on it as soon as possible, but please let me know if you have comments.

PR #3289, opened 2018-05-02
Add linebreaks rule

saulotoledo:

Closes https://github.com/stylelint/stylelint/issues/3114

~It is a first draft. Needs some more test cases (e.g. newlines in the end of the file and empty newlines) and possibly some polishment. I will work on these as soon as possible this week. Any comments can be helpful. I will edit this description and title in the final version.~

First complete version available for review.

### DIFF
--- a/docs/user-guide/example-config.md
+++ b/docs/user-guide/example-config.md
@@ -83,6 +83,7 @@ You might want to learn a little about [how rules are named and how they work to
     "indentation": int|"tab",
     "keyframe-declaration-no-important": true,
     "length-zero-no-unit": true,
+    "linebreaks": "unix"|"windows",
     "max-empty-lines": int,
     "max-line-length": int,
     "max-nesting-depth": int,
--- a/docs/user-guide/rules.md
+++ b/docs/user-guide/rules.md
@@ -345,6 +345,7 @@ Here are all the rules within stylelint, grouped first [by category](../../VISIO
 #### General / Sheet
 
 -   [`indentation`](../../lib/rules/indentation/README.md): Specify indentation (Autofixable).
+-   [`linebreaks`](../../lib/rules/linebreaks/README.md): Enforces a consistent linebreak style (Autofixable).
 -   [`max-empty-lines`](../../lib/rules/max-empty-lines/README.md): Limit the number of adjacent empty lines.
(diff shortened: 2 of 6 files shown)


saulotoledo:
@evilebottnawi Sure! I will add that test cases too.

saulotoledo:
@evilebottnawi @jeddy3 Final version available.

I used a simple loop to iterate through the lines for reporting. I was using only the PostCSS API, but the code was getting very complicated because I had to calculate the proper lines and columns through the raw values. But there are too much situations to handle (newlines inside comments, comments with multiple lines, newlines between rules, values, after/before brackets and so on). I have a local copy of an (almost) finished version that works just by using the API, but it is difficult to read because of that. I can share if you want.

So, I decided to use the API only when fixing the code, a place where I do not have to worry about calculating the proper lines and columns. Let me know your opinion about.

saulotoledo:
@evilebottnawi, @jeddy3 Also, note that I created a manual node for reporting. I had to do that because there is no node for empty lines or half comments, for example. Thus, it is impossible to report the correct lines by using the original nodes. If there is a better way to do that, please let me know and I will update the code.

hudochenkov:
If you add changes based on my comment above, please don't rebase, so we can see what changed since the last review :)

saulotoledo:
Hi @hudochenkov . I did not forget about the PR. I am traveling but I will read the request again as fast as possible. :)

saulotoledo:
@hudochenkov I did some tentatives to use `line` and `index`, but no success. A good example are multiline comments (the "newlines" above are just examples):

```javascript
/**\n
 * This is a\n
 * multiline\r\n  <- there is an error here
 * comment\n
 */\n
```
I can just report the last line and do a lot of processing and "ifs" in the code to discover the line and index numbers for the problem in the middle of the comment.

There are also other situations. I am not sure about the exact cases because I was trying to test with comments. I was solving one by one, but the code was a bit complex with with many cases. The above version sounds better for me. If you want I can still share the unfinished solution in another branch.

jeddy3:
> -   Added: `linebreaks` rule ([#3289](https://github.com/stylelint/stylelint/pull/3289)).

ntwb:
Welcome to the @stylelint organisation @saulotoledo  🎉

saulotoledo:
Thank you @ntwb! I'm happy with the invitation.

PR #3292, opened 2018-05-03
Add autofix to color-no-hex rule

### LINKED ISSUE #3291: Add autofix to color-no-hex

`stylelint --fix` should convert hex colors to either rgb or hsl

color-no-hex

Input
```css
.foo {
  color: #FFFFFF;
}
```
Output
```css
.foo {
  color: rgb(255,255,255);
}
```

Since it's a new option, I propose the following:
```json
{
  "rules": {
	"color-no-hex": "rgb"
  }
}
```

The option would accept `rgb`, `hsl`, `rgba`, `hsla` and `true`. True would error but wouldn't try to fix anything.

9.2.0

This should work despite the method chose to run stylelint

No

I expected `stylelint --fix` to convert hex color to rgb (which is the most popular) and was disappointed when I couldn't do it.

Just errors, no fixes.

I already have a branch in my fork where this is working, but I haven't written the tests yet. If people are interested, I'll update the branch and open the pull-request.

filipekiss:

Closes #3291 

I kept the option `true` working, but no fix will be performed if the option is set to `true`. This is to avoid breaking any existing project who uses this setting.

I've added the options `rgb`, `rgba`, `hsl, and `hsla`. The versions with the trailing `a` are there just for compatibility, but `rgb` and `rgba` do the same thing.

An hex color will, most of the time, be converted to `rgb` (or `hsl). The formats with an alpha are only used if the color had an alpha in the first place.

I've updated the tests with the fixes.

I also added a new option to Jest, that is to pass a `unfixable` flag to the test. I needed to do this because of two reasons:

If the options is set to true and the fix flag is passed, I needed to make sure that the test would be rejected, as expected, but I needed to mark it as unfixable. 

The second reason is when a malformed hex color is input: This can't be fixed as the color can't be parsed, so the original value is kept, but I needed a way to ensure the tests would be close to real world, so I've made this change.

Let me know if anything seems off or needs changing and I'd be happy to comply :)

Thank you all for the great plugin!

### DIFF
--- a/lib/rules/color-no-hex/README.md
+++ b/lib/rules/color-no-hex/README.md
@@ -8,9 +8,23 @@ a { color: #333 }
  * These hex colors */
 ```
 
+The `--fix` option on the command line can automatically fix all of the problems reported by this rule if applicable. See the options below.
+
 ## Options
 
-### `true`
+`boolean`: `true|false`
+
+If `false`, the rule is disabled. If `true`, the rule is enabled, but no fix will be performed.
+
+`string`: `rgb|rgba|hsl|hsla`
+
+### `rgb` | `rgba`
+
+`rgb` and `rgba` are the same. If the color format is an hex with transparency (Like #FFFFFFDC), the output will be in rgba format. Otherwise, it will be rgb. There's no way to force one or another.
+
+### `hsl` | `hsla`
+
+`hsl` and `hsla` are the same. If the color format is an hex with transparency (Like #FFFFFFDC), the output will be in hsla format. Otherwise, it will be hsl. There's no way to force one or another.
 
 The following patterns are considered violations:
 
--- a/package.json
+++ b/package.json
@@ -1,197 +1,176 @@
 {
-  "name": "stylelint",
-  "version": "9.2.0",
-  "description": "A mighty, modern CSS linter.",
-  "keywords": [
-    "css",
-    "less",
-    "sass",
-    "scss",
-    "sugarss",
-    "lint",
-    "linter",
-    "stylelint"
-  ],
-  "authors": [
-    "David Clark",
-    "Maxime Thirouin",
-    "Richard Hallows"
-  ],
-  "license": "MIT",
-  "homepage": "https://stylelint.io",
-  "repository": {
-    "type": "git",
-    "url": "https://github.com/stylelint/stylelint.git"
-  },
-  "main": "lib/index.js",
-  "bin": "bin/stylelint.js",
-  "files": [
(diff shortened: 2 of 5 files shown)


evilebottnawi:
@filipekiss please rebase, thanks!

filipekiss:
@evilebottnawi Done! :)

package.json
```diff
@@ -1,197 +1,176 @@
 {
```
evilebottnawi: Please don't modify `package.json`

lib/rules/color-no-hex/README.md
```diff
@@ -8,9 +8,23 @@ a { color: #333 }
...
+
+`string`: `rgb|rgba|hsl|hsla`
+
+### `rgb` | `rgba`
+
+`rgb` and `rgba` are the same. If the color format is an hex with transparency (Like #FFFFFFDC), the output will be in rgba format. Otherwise, it will be rgb. There's no way to force one or another.
+
+### `hsl` | `hsla`
+
+`hsl` and `hsla` are the same. If the color format is an hex with transparency (Like #FFFFFFDC), the output will be in hsla format. Otherwise, it will be hsl. There's no way to force one or another.
```
evilebottnawi: Looks like you implement not only `fix`, also you add new options, can we separate PR on two parts?

package.json
filipekiss: Sorry :( It was my linter. I'll revert the changes and push only the dependency I need!

lib/rules/color-no-hex/README.md
filipekiss: Sure we can, but do you think it makes sense? Because `true` with `fix` will not fix anything, actually. The option is needed to select a format to convert the hex colors to.

Any suggestion on how can we do this?

filipekiss: @evilebottnawi No problem. I could make `true` convert hex colors to rgb but that would change the behavior for anyone who already uses this rule, so I opted for this approach: `true` with `fix` will remain doing nothing, as it is right now. Then I introduced `rgba?` and `hsla?` so the user can select their preferred format :)

filipekiss: @jeddy3 Hey Richard, thank you for your consideration.

Sure, I'll publish it as a `color-format` plugin and link it here :)

jeddy3: @filipekiss Great job with the plugin!

Now that https://github.com/stylelint/stylelint/pull/3322 is merged, I'll close with PR.

PR #3297, opened 2018-05-10
Add autofix to block-opening-brace-space-before

superdem:
**Description**
Enable auto fix option for `block-opening-brace-space-before`

### DIFF
--- a/lib/rules/block-opening-brace-space-before/__tests__/index.js
+++ b/lib/rules/block-opening-brace-space-before/__tests__/index.js
@@ -6,7 +6,7 @@ const { messages, ruleName } = rule;
 testRule(rule, {
   ruleName,
   config: ["always"],
-
+  fixed: true,
   accept: [
     {
       code: "@import url(x.css)"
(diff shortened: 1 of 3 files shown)


jeddy3:
@superdem Thank you for starting on this PR!

I believe you'll need to add a `fixed` property to all of the tests for this rule, [like so](https://github.com/stylelint/stylelint/blob/master/lib/rules/at-rule-name-case/__tests__/index.js).

We can then be confident the code works.

superdem:
Hi @jeddy3 , 
I've just updated the test with fixed. Hope it works now.

Thanks!

lib/rules/block-opening-brace-space-before/__tests__/index.js
```diff
@@ -6,7 +6,7 @@ const { messages, ruleName } = rule;
 testRule(rule, {
   ruleName,
   config: ["always"],
-
+  fixed: true,
```
hudochenkov: It should be `fix: true`. Please change it here and in all other places.

hudochenkov:
Superseded by https://github.com/stylelint/stylelint/pull/3438. Sorry this PR wasn't ready.

PR #3300, opened 2018-05-10
chore: cache node_modules

DanielRuf:
This PR caches the node_modules directory on Appveyor which is not the case currently.

"No, it's self explanatory."

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,6 +16,9 @@ install:
   - ps: Install-Product node $env:nodejs_version
   - npm install
 
+cache:
+  - node_modules
+
 test_script:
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -16,6 +16,9 @@ install:
   - ps: Install-Product node $env:nodejs_version
   - npm install
 
+cache:
+  _ node_modules
```
evilebottnawi: `_` to `-`

DanielRuf: Oops.

DanielRuf: Changed.

PR #3320, opened 2018-05-19
Add media-feature-name-value-whitelist

### LINKED ISSUE #3312: Add media-feature-name-value-whitelist
People sometimes use breakpoints at the weirdest values. This adds unnecessary complexity to understanding the points on which the UI changes.

I would like to propose a _breakpoint-strict_ rule which takes a list of values that can be used in media queries. There would be two lists that can be passed, one for `min-width` and one for `max-width`. This to avoid issues with overlapping values as explained in https://github.com/stylelint/stylelint/issues/3311.

There could be a possibility to add presets such as the [Material breakpoints](https://material.io/design/layout/responsive-layout-grid.html#breakpoints) and [Bootstrap](https://getbootstrap.com/docs/4.1/layout/overview/#responsive-breakpoints).

MatthiasKunnen:

Closes #3312

No, it's self explanatory.

### DIFF
--- a/docs/user-guide/rules.md
+++ b/docs/user-guide/rules.md
@@ -171,7 +171,7 @@ Here are all the rules within stylelint, grouped first [by category](../../VISIO
 
 -   [`media-feature-name-blacklist`](../../lib/rules/media-feature-name-blacklist/README.md): Specify a blacklist of disallowed media feature names.
 -   [`media-feature-name-no-vendor-prefix`](../../lib/rules/media-feature-name-no-vendor-prefix/README.md): Disallow vendor prefixes for media feature names.
--   [`media-feature-name-value-whitelist`](../../lib/rules/media-feature-name-value-whitelist/README.md): Specify a whitelist of allowed media feature and value pairs.
+-   [`media-feature-name-value-whitelist`](../../lib/rules/media-feature-name-value-whitelist/README.md): Specify a whitelist of allowed media feature name and value pairs.
 -   [`media-feature-name-whitelist`](../../lib/rules/media-feature-name-whitelist/README.md): Specify a whitelist of allowed media feature names.
 
 #### Custom media
--- a/docs/user-guide/rules.md
+++ b/docs/user-guide/rules.md
@@ -171,6 +171,7 @@ Here are all the rules within stylelint, grouped first [by category](../../VISIO
 
 -   [`media-feature-name-blacklist`](../../lib/rules/media-feature-name-blacklist/README.md): Specify a blacklist of disallowed media feature names.
 -   [`media-feature-name-no-vendor-prefix`](../../lib/rules/media-feature-name-no-vendor-prefix/README.md): Disallow vendor prefixes for media feature names.
+-   [`media-feature-name-value-whitelist`](../../lib/rules/media-feature-name-value-whitelist/README.md): Specify a whitelist of allowed media feature and value pairs.
 -   [`media-feature-name-whitelist`](../../lib/rules/media-feature-name-whitelist/README.md): Specify a whitelist of allowed media feature names.
 
 #### Custom media
--- a/lib/rules/media-feature-name-value-whitelist/README.md
+++ b/lib/rules/media-feature-name-value-whitelist/README.md
@@ -5,7 +5,7 @@ Specify a whitelist of allowed media feature name and value pairs.
 ```css
 @media screen and (min-width: 768px) {}
 /**                ↑          ↑
- *         These features and these values */
+ *    These features and values */
 ```
 
 This rule ignores media features within range and boolean context.
--- a/lib/rules/media-feature-name-value-whitelist/README.md
+++ b/lib/rules/media-feature-name-value-whitelist/README.md
@@ -19,11 +19,12 @@ This rule ignores media features within range and boolean context.
 }
 ```
 
-If a media feature is found, only its whitelisted property values are allowed. The rule reports all
-non-matching values.
+If a media feature name is found in the object, only its whitelisted values are
+allowed. If the media feature name is not included in the object, anything goes.
 
-If a property or value is surrounded with `/` (e.g. `"/width$/"`), it is interpreted as a regular
-expression. For example, `/width$/` will match `max-width` and `min-width`.
+If a name or value is surrounded with `/` (e.g. `"/width$/"`), it is interpreted
+as a regular expression. For example, `/width$/` will match `max-width` and
+`min-width`.
 
 Given:
 
--- a/lib/rules/media-feature-name-value-whitelist/README.md
+++ b/lib/rules/media-feature-name-value-whitelist/README.md
@@ -14,8 +14,8 @@ This rule ignores media features within range and boolean context.
 
 ```json
 {
-  "max-width": ["Array", "of", "values"],
-  "/regex/": ["/regex/", "non-regex"]
+  "unprefixed-media-feature-name": ["array", "of", "values"],
+  "/unprefixed-media-feature-name/": ["/regex/", "non-regex"]
 }
 ```
 
--- a/lib/rules/media-feature-name-value-whitelist/README.md
+++ b/lib/rules/media-feature-name-value-whitelist/README.md
@@ -28,6 +28,7 @@ expression. For example, `/width$/` will match `max-width` and `min-width`.
 ## Examples
 
 Given
+
 ```json
 {
   "min-width": ["768px"],
--- a/lib/rules/media-feature-name-value-whitelist/README.md
+++ b/lib/rules/media-feature-name-value-whitelist/README.md
@@ -1,0 +1,64 @@
+# media-feature-name-value-whitelist
+
+Specify a whitelist of allowed media feature and value pairs.
+
+```css
+@media screen and (min-width: 768px) {}
+/**                    ↑        ↑
+ *         These features and these values */
+```
+
+**Caveat:** Media feature names within range or boolean context are ignored.
+
+## Options
+
+```json
+{
+  "max-width": ["Array", "of", "values"],
+  "/regex/": ["/regex/", "non-regex"]
+}
+```
+
+If a media feature is found, only its whitelisted property values are allowed. The rule reports all
+non-matching values.
+
+If a property or value is surrounded with `/` (e.g. `"/width$/"`), it is interpreted as a regular
+expression. For example, `/width$/` will match `max-width` and `min-width`.
+
+## Examples
+
+Given
+```json
+{
+  "min-width": ["768px"],
+  "/resolution/": ["/dpcm$/"]
+}
+```
+
+The following pattern are considered violations:
+
+```css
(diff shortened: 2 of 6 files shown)


MatthiasKunnen:
@evilebottnawi, there are two tests for non standard syntax. SCSS in particular. What kind of additional tests do you have in mind?

evilebottnawi:
@stylelint/contributors we ignore non standard syntax in any places right?

MatthiasKunnen:
Correct

MatthiasKunnen:
Non standard syntax is handled as suggested in the issue (https://github.com/stylelint/stylelint/issues/3312#issuecomment-390365412) by @hudochenkov.

docs/user-guide/rules.md
```diff
@@ -171,6 +171,7 @@ Here are all the rules within stylelint, grouped first [by category](../../VISIO
 
 -   [`media-feature-name-blacklist`](../../lib/rules/media-feature-name-blacklist/README.md): Specify a blacklist of disallowed media feature names.
 -   [`media-feature-name-no-vendor-prefix`](../../lib/rules/media-feature-name-no-vendor-prefix/README.md): Disallow vendor prefixes for media feature names.
+-   [`media-feature-name-value-whitelist`](../../lib/rules/media-feature-name-value-whitelist/README.md): Specify a whitelist of allowed media feature and value pairs.
```
jeddy3: A minor tweak, as I believe the "name" and the "value" represent the pair.

> Specify a whitelist of allowed media feature name and value pairs.

lib/rules/media-feature-name-value-whitelist/README.md
```diff
@@ -0,0 +1,65 @@
+# media-feature-name-value-whitelist
+
+Specify a whitelist of allowed media feature and value pairs.
```
jeddy3: Specify a whitelist of allowed media feature name and value pairs.

lib/rules/media-feature-name-value-whitelist/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```css
+@media screen and (min-width: 768px) {}
+/**                    ↑        ↑
```
jeddy3: Please align the arrows with the start of each bit of vocab i.e. the "m" in "min-width" and "7" in "768px"

(There are a lot of rules in stylelint, so we try to be consistent as possible across all the docs).

lib/rules/media-feature-name-value-whitelist/README.md
```diff
@@ -0,0 +1,65 @@
...
+ *         These features and these values */
+```
+
+**Caveat:** Media feature names within range or boolean context are ignored.
```
jeddy3: Please write as the following to be consistent with most of the other rules:

"This rules ignores media feature within range or boolean context."

I'll create a separate issue about updating the other rules that still use the old "**Caveat:**" convention (e.g. `media-feature-name-blacklist`).

lib/rules/media-feature-name-value-whitelist/README.md
```diff
@@ -0,0 +1,65 @@
...
+## Options
+
+```json
+{
+  "max-width": ["Array", "of", "values"],
+  "/regex/": ["/regex/", "non-regex"]
+}
+```
+
+If a media feature is found, only its whitelisted property values are allowed. The rule reports all
```
jeddy3: Let's go with the following to be more consistent with `declaration-property-value-whitelist`:

---

## Options

`object`: `{
  "unprefixed-media-feature-name": ["array", "of", "values"]
}`

If a media feature name is found in the object, only its whitelisted values are allowed. If the media feature name is not included in the object, anything goes.

If a name or value is surrounded with `/` (e.g. `"/width$/"`), it is interpreted as a regular expression. For example, `/width$/` will match `max-width` and `min-width`.

---

lib/rules/media-feature-name-value-whitelist/README.md
```diff
@@ -0,0 +1,65 @@
...
+non-matching values.
+
+If a property or value is surrounded with `/` (e.g. `"/width$/"`), it is interpreted as a regular
+expression. For example, `/width$/` will match `max-width` and `min-width`.
+
+## Examples
```
jeddy3: Remove this title, please.

lib/rules/media-feature-name-value-whitelist/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+Given
```
jeddy3: "Given:"

(It's a little clunky, but the stylelint website [needs the semicolon](https://github.com/stylelint/stylelint.io/blob/master/plugins/remark-stylelint-pattern-validity/index.js#L16) for the pattern background colours).

lib/rules/media-feature-name-value-whitelist/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```json
+{
+  "min-width": ["768px"],
```
jeddy3: I think two values here would be useful to the reader e.g. `"min-width": ["768px", "1024px"]`.

You'll need to add an additional pattern to the "*not* considered violations" list for it.

lib/rules/media-feature-name-value-whitelist/__tests__/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+  reject: [
+    {
+      code: "@media screen and (min-width: 1000px) {}",
+      message: messages.rejected("min-width", "1000px"),
+      line: 1,
+      column: 31
+    },
+    {
+      code: "@media screen (min-width: 768px) and (min-width: 1000px) {}",
```
jeddy3: Can we get this two lines please to test the `line` property value e.g.

```
code: "@media screen (min-width: 768px)\nand (min-width: 1000px) {}",`
description: "Media feature multiple",
message: messages.rejected("min-width", "1000px"),
line: 2,
column: 5
```

lib/rules/media-feature-name-value-whitelist/index.js
```diff
@@ -0,0 +1,67 @@
...
+const atRuleParamIndex = require("../../utils/atRuleParamIndex");
+const matchesStringOrRegExp = require("../../utils/matchesStringOrRegExp");
+const mediaParser = require("postcss-media-query-parser").default;
+const report = require("../../utils/report");
+const ruleMessages = require("../../utils/ruleMessages");
+const validateOptions = require("../../utils/validateOptions");
+
+const ruleName = "media-feature-name-value-whitelist";
+
+const messages = ruleMessages(ruleName, {
```
jeddy3: Formatters will usually append the rule name, therefore the "media feature" bit is a little redundant. Let's go with the following instead:

```
  rejected: (name, value) =>
    `Unexpected value "${value}" for name "${name}"`
});

File: lib/rules/media-feature-name-value-whitelist/README.md (hunk 3 above)
Comment (MatthiasKunnen) [reply lib/rules/media-feature-name-value-whitelist/README.md]: Will do. I've not come across this in the docs so I wasn't aware of this rule.

File: lib/rules/media-feature-name-value-whitelist/README.md (hunk 4 above)
Comment (MatthiasKunnen) [reply lib/rules/media-feature-name-value-whitelist/README.md]: Yeah, I reused some things here and there to the rule as consistent as possible. That the caveat was deprecated, I did not know :)

File: lib/rules/media-feature-name-value-whitelist/README.md (hunk 5 above)
Comment (MatthiasKunnen) [reply lib/rules/media-feature-name-value-whitelist/README.md]: Is this the only rule that uses `object:`? I personally find it atrocious to look at/read in code as well as rendered.  The JSON object is better formatted. 
```json
{
  "max-width": ["Array", "of", "values"],
  "/regex/": ["/regex/", "non-regex"]
}
```
I believe there is another rule that uses this in some form but I'd have to check.

MatthiasKunnen:
@jeddy3, thank you for the review. I've replied on some of your requested changes and will push updates soon.

@evilebottnawi, you've requested changes but that seems to be out-of-date.

lib/rules/media-feature-name-value-whitelist/README.md
jeddy3: Yep, it's not an obvious one to glean. I've [created a PR](https://github.com/stylelint/stylelint/pull/3363) to explicitly highlight the convention (along with the "This rule ..." one).

lib/rules/media-feature-name-value-whitelist/README.md
jeddy3: > Yeah, I reused some things here and there to the rule as consistent as possible.

Yep, I saw. Good job and thanks :)

> That the caveat was deprecated, I did not know :)

We're still ironing out all the kinks in the docs, which is why your kind of feedback is so useful.

We just merged [a PR](https://github.com/stylelint/stylelint/pull/3360) to make the other rule READMEs consistent.

lib/rules/media-feature-name-value-whitelist/README.md
jeddy3: > I believe there is another rule that uses this in some form but I'd have to check.

I think it's only the `declaration-property-*-*list` rules that accept objects. I might be wrong, though.

> I personally find it atrocious to look at/read in code as well as rendered.

Yep, I was thinking the same thing. However, if we can't find another example of it done as a JSON object then I suggest we go with the other approach in this PR to keep things consistent.

We should then create a separate issue to address how we can better communicate the options across the board. To be honest, all the option descriptions need a little love and I don't think we ever rationalised the approach we took. It'd be good to look, as part of that issue, into how other tools communicate their options.

lib/rules/media-feature-name-value-whitelist/__tests__/index.js
MatthiasKunnen: You mean at column 17, right? The rule reports errors on the value, not the media feature name.

jeddy3: > You mean at column 17, right?

Oops, yes.

jeddy3:
@MatthiasKunnen Changes look good!

I think once the `line` test is in and the options section is changed to:

---

## Options

`object`: `{
  "unprefixed-media-feature-name": ["array", "of", "values"]
}`

If a media feature name is found in the object, only its whitelisted values are allowed. If the media feature name is not included in the object, anything goes.

If a name or value is surrounded with `/` (e.g. `"/width$/"`), it is interpreted as a regular expression. For example, `/width$/` will match `max-width` and `min-width`.

---

We'll be good to merge and get this into `9.3.0`

MatthiasKunnen:
@jeddy3, the line test is already added. I suggest to resolve #3372 first. I already have a PR ready but I am holding on to it until my last comment has been addressed.

MatthiasKunnen:
@jeddy3, this PR is ready I believe. The new options format in the README depends on PR #3389.

jeddy3:
> -   Added: `media-feature-name-value-whitelist` rule ([#3320](https://github.com/stylelint/stylelint/pull/3320)).

PR #3321, opened 2018-05-20
Added keyframes-name-pattern rule

### LINKED ISSUE #3262: Add keyframes-name-pattern

I'd like to make an Feature request, where `@keyframes` definition's name needs to match given pattern, something like `selector-class-pattern` rule, but for keyframes.

I'd like to use an namespace validation rule for all my keyframe definitions.

Issue relates to similar rule `selector-class-pattern`.

```css
@keyframes NAMESPACE-spin {
  from {
    transform: rotate(0deg);
  }

  to {
    transform: rotate(360deg);
  }
}
```

n/a

`9.2.0`

CLI with `stylelint --config myconfig *.css` should be ok.

No.

I expect warnings to be flagged if keyframe definition name doesn't match the pattern.

Nothing related to the feature request, because of lacking rules.

MatthiasKunnen:

Closes #3262

No, it's self explanatory.

### DIFF
--- a/lib/rules/keyframes-name-pattern/index.js
+++ b/lib/rules/keyframes-name-pattern/index.js
@@ -1,0 +1,47 @@
+"use strict";
+
+const _ = require("lodash");
+const atRuleParamIndex = require("../../utils/atRuleParamIndex");
+const report = require("../../utils/report");
+const ruleMessages = require("../../utils/ruleMessages");
+const validateOptions = require("../../utils/validateOptions");
+
+const ruleName = "keyframes-name-pattern";
+
+const messages = ruleMessages(ruleName, {
+  expected: keyframeName =>
+    `Expected keyframe name "${keyframeName}" to match specified pattern`
+});
+
+const rule = function(pattern) {
+  return (root, result) => {
+    const validOptions = validateOptions(result, ruleName, {
+      actual: pattern,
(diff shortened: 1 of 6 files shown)


lib/rules/keyframes-name-pattern/index.js
```diff
@@ -0,0 +1,47 @@
...
+      actual: pattern,
+      possible: [_.isString]
+    });
+
+    if (!validOptions) {
+      return;
+    }
+
+    const regex = _.isString(pattern) ? new RegExp(pattern) : pattern;
+    root.walkAtRules(/^keyframes/i, keyframesNode => {
```
evilebottnawi: Not all `keyframes` starts with `keyframes`, we have already prefixed version `--webkit-keyframes`, need add supports and tests. Thanks for work!

MatthiasKunnen:
@evilebottnawi, the failing check seems to be something that needs addressing outside of this PR. Can you confirm?

MatthiasKunnen:
Thank you, let me know if there is any more feedback on this PR.

ntwb:
Added to changelog in https://github.com/stylelint/stylelint/commit/1b152cacb1747ccfa69e6cdd1a5d5c9810130640

> Added: `keyframes-name-pattern` rule (#3321).

Issue #3304, opened 2018-05-11
Add `declaration-block-no-redundant-longhand-properties` autofix


imvetri:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

`stylelint --fix` should convert redundant longhand properties to shorthand
> Which rule, if any, is this issue related to?

`declaration-block-no-redundant-longhand-properties`

> What CSS is needed to reproduce this issue?

Following will throw error

```css
  a {
    padding-top: 1px;
    padding-right: 2px;
    padding-bottom: 3px;
    padding-left: 4px; 
}
```

> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
  "rules": {
    "declaration-block-no-redundant-longhand-properties": [true, {
        "severity": "warning"
    }],
  }
}
```

> Which version of stylelint are you using?

e.g. `9.2.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

This should work despite the method chose to run stylelint

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

```css
  a {
    padding-top: 1px;
    padding-right: 2px;
    padding-bottom: 3px;
    padding-left: 4px; 
}
```
should convert above to following

```css
a {
  padding: 1px 2px 3px 4px;
}
```

> What actually happened (e.g. what warnings or errors you are getting)?

No fix currently available

Further notes: Currently this is being discussed in https://github.com/stylelint/stylelint/issues/3068. But that has lot of other rules, so creating a separate one for this rule in case I could write a fix.

jeddy3:
> Currently this is being discussed in #3068.

I've removed this rule from the list in that issue.

> But that has lot of other rules, so creating a separate one for this rule in case I could write a fix.

Good idea. Good luck writing the fix too!

mattxwang:
I know #3326 was closed due to inactivity; is there still interest in writing this autofix? I maintain just-the-docs, and we embarrassingly made a mistake manually fixing this rule (https://github.com/just-the-docs/just-the-docs/pull/1123 and https://github.com/just-the-docs/just-the-docs/pull/1104).

If so, I'd be happy to take a stab at implementing it. I can first base my solution off of @imvetri's existing PR (#3326), unless there have been any developments in utilities/blueprints since then that would make more sense to look at?

imvetri:
[...]

PR #3326, opened 2018-05-22
Add autofix to declaration-block-no-redundant-longhand-properties

imvetri:
WIP. 
As discussed [here](https://github.com/stylelint/stylelint/issues/3304) I have a initial version of autofix written. Before writing more, I would like to get feedback on the approach.

**Overview of the approach:**
On rule violation, it creates a shorthand value using references/shorthandFormats/resolveValue. For now I have created for padding.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,26 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Jest All",
+            "program": "${workspaceFolder}/node_modules/jest/bin/jest",
+            "args": ["--runInBand"],
+            "console": "integratedTerminal",
+            "internalConsoleOptions": "neverOpen"
+          },
+          {
+            "type": "node",
+            "request": "launch",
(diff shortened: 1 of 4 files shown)


jeddy3:
I think adding tests for `padding` using the `fixed:` property of the schema (like [so](https://github.com/stylelint/stylelint/blob/master/lib/rules/rule-empty-line-before/__tests__/index.js#L46)) would be a good way of validating your approach so far.

imvetri:
@jeddy3 Thanks for the pointer. Greatly helped.

ntwb:
@imvetri Any chance you could update this PR please? We'd love to get this shipped :shipit:

imvetri:
@ntwb Thanks for following up. I will have some updates in < 2 weeks.

imvetri:
@ntwb  I m still working on this, I don't think so I will be able to complete by this weekend.

ntwb:
Cool, thanks for the update @imvetri, let us know if we can help in any way :)

imvetri:
@ntwb I continued working on this by adding unit test with context.fix as true ( refer to the [line](https://github.com/stylelint/stylelint/pull/3326/files#diff-28ddeca0d3df92ea11f69bc5a08b7570R9) ). While running the test what I found is that context.fix is not always as true. Sometimes it is true and sometimes it is undefined. Do you think you can guess what could be wrong? .

I have also pushed with vscode configuration in the latest commit to make debugging easier.

hudochenkov:
Closing due inactivity. We can reopen if requested changes were made.

Issue #3327, opened 2018-05-22
throws CssSyntaxError when config processors is a empty array


Xotic750:
**Describe the bug? (A clear and concise description of what the bug is)**

**Which rule, if any, is this issue related to?**

**What CSS is needed to reproduce this issue?**

```html
<template>
  <div>
    <h5 v-if="title">{{ title }}</h5>
  </div>
</template>

<script>
export default {
  props: {
    title: {
      default: '',
      type: String,
    },
  },
};
</script>

<style lang="less">
</style>
```

**What stylelint configuration is needed to reproduce this issue?**

```javascript
/**
 * @file Manages the root configuration settings for project wide stylelint.
 * @copyright Copyright (c) 2017-present, ProReNata AB
 * @module stylelint/root/configuration
 * @version 1.0.0
 * @since 0.0.2
 * @see {@link https://stylelint.io/} for further information.
 */

module.exports = {
  /**
   * Extend an existing configuration.
   * @type {String|Array.<String>}
   * @see {@link https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#extends}
   */
  extends: [
    /**
     * The standard shareable config for stylelint.
     * @type {string}
     * @see {@link https://github.com/stylelint/stylelint-config-standard}
     */
    'stylelint-config-standard',
    /**
     * Turns off all rules that are unnecessary or might conflict with prettier.
     * @type {string}
     * @see {@link https://github.com/shannonmoeller/stylelint-config-prettier}
     */
    'stylelint-config-prettier',
  ],

  /**
   * @type {array}
   */
  plugins: [],

  /**
   * @type {array}
   */
  processors: [],

  /**
   * @type {!Object}
   */
  rules: {
    'no-empty-source': null,
  },
};
```

**Which version of stylelint are you using?**

`9.2.1`

**How are you running stylelint: CLI, PostCSS plugin, Node API?**

"CLI with `stylelint '**/*.{css,sass,scss,less,vue}'`"
or
Node with with webpack or via Webstorm support

**Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?**

"Yes, it's a vue file which worked/works in v9.2.0?"

**What did you expect to happen?**

"No warnings to be flagged."

**What actually happened (e.g. what warnings or errors you are getting)?**

"See the attached screenshot for Webstorm, same output from CLI. You can see the offending line in the snapshot `<h5 v-if="title">{{title}}</h5>`"

[...]

PR #3329, opened 2018-05-23
fix CssSyntaxError when config processors is a empty array

gucong3000:

#3327

No, it's self explanatory.


gucong3000:
Closes: #3327

hudochenkov:
Is it possible to write a test for this case?

evilebottnawi:
Yep need tests for avoid regressions in future :+1:

jeddy3:
Was there a CHANGELOG update to accompany this merge?

PR #3339, opened 2018-05-28
New feature: add `globbyOptions` in option.

### LINKED ISSUE #3338: Add globbyOptions option
In project [stylelint-webpack-plugin](https://github.com/webpack-contrib/stylelint-webpack-plugin), `files` cannot use negated pattern, because there is no option in `stylelint` to set `cwd` of files, so that `stylelint-webpack-plugin` has convert all files paths into abs path.
I'v figure out a way to solve this problem, is to detect if a pattern starts with `!` in `stylelint-webpack-plugin` when convert relative paths to abs paths. But that is not a good way. Here is our discussions: 
https://github.com/webpack-contrib/stylelint-webpack-plugin/pull/158
https://github.com/webpack-contrib/stylelint-webpack-plugin/pull/159

We want to add `filesCwd` in options, where we can pass `cwd` field to `globby`.

**Describe alternatives you've considered**
Null.

**Additional context**
Null.

tinymins:

Closes #3338

No, it's self explanatory.

### DIFF
--- a/docs/user-guide/node-api.md
+++ b/docs/user-guide/node-api.md
@@ -61,9 +61,17 @@ The difference between the `configOverrides` and `config` options is this: If an
 
 A file glob, or array of file globs. Ultimately passed to [node-glob](https://github.com/isaacs/node-glob) to figure out what files you want to lint.
 
-Relative globs are considered relative to `process.cwd()`.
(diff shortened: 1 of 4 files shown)


jeddy3:
> -   Added: `globbyOptions` option ([#3339](https://github.com/stylelint/stylelint/pull/3339)).

PR #3345, opened 2018-05-31
Add autofix for selector-pseudo-element-colon-notation

### LINKED ISSUE #3344: Add autofix to selector-pseudo-element-colon-notation

I would like to enforce using double-colon notation for pseudo-elements. For example
`a::before { display: block; }` is valid, while `a:before { display:block; }` is invalid.

This is possible using the [selector-pseudo-element-colon-notation](https://stylelint.io/user-guide/rules/selector-pseudo-element-colon-notation/) and setting its value to "double".

However there is currently no autofix logic for this, so running `stylelint my.css --fix` does not transform `a:before` into `a::before`.

I would like autofixing to be implemented for selector-pseudo-element-colon-notation, so that when stylelint is ran with `--fix` the linting error is automatically resolved.

If the value for selector-pseudo-element-colon-notation is "double" then level 1 and 2 pseudo elements (before, after, first-line and first-letter as per [keywordSets](https://github.com/stylelint/stylelint/blob/0f9d17f8df9bf56794caf08edad1f7a1e5d5aaff/lib/reference/keywordSets.js#L163)) preceded by a single colon should be converted to be preceded by a double colon. For instance `a:before` becomes `a::before`.

If the value for selector-pseudo-element-colon-notation is "single" then level 1 and 2 pseudo elements (before, after, first-line and first-letter as per [keywordSets](https://github.com/stylelint/stylelint/blob/0f9d17f8df9bf56794caf08edad1f7a1e5d5aaff/lib/reference/keywordSets.js#L163)) preceded by a double colon should be converted to be preceded by a single colon. For instance `a::before` becomes `a:before`.

BPScott:
selector-pseudo-element-colon-notation allows you to enforce using double or single colons on level 1 and level 2 pseudo-elements (back when `:before` and `:after` were considered valid before spec updates made ::before and ::after the preferred format).

For example using the "double" value, `a::before { display: block; }` is valid, while `a:before { display:block; }` is invalid.

This PR adds autofix support to the rule so that:

- If the value of double is set then `a:before { display: block; }` will be automatically fixed to `a::before { display: block; }`
- If the value of single is set then `a::before { display: block; }` will be automatically fixed to `a:before { display: block; }`

Fixes #3344

### DIFF
--- a/lib/rules/selector-pseudo-element-colon-notation/README.md
+++ b/lib/rules/selector-pseudo-element-colon-notation/README.md
@@ -12,6 +12,8 @@ The `::` notation was chosen for *pseudo-elements* to establish a discrimination
 
 However, for compatibility with existing style sheets, user agents also accept the previous one-colon notation for *pseudo-elements* introduced in CSS levels 1 and 2 (namely, `:first-line`, `:first-letter`, `:before` and `:after`).
 
+The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofixing-errors) can automatically fix all of the problems reported by this rule.
+
 ## Options
 
 `string`: `"single"|"double"`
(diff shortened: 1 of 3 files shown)


jeddy3:
> -   Added: `selector-pseudo-element-colon-notation` autofix ([#3345](https://github.com/stylelint/stylelint/pull/3345)).

@BPScott Thanks for contributing to stylelint :)

Issue #3354, opened 2018-06-03
[Rule]: Using :focus with :hover


YozhikM:
**Is your feature request related to a problem? Please describe.**
To improve the availability of sites. To remind developers of this important pseudo-class.

**Describe the solution you'd like**
Adding a new rule.

**Describe alternatives you've considered**
Include this in other selector rules.

**Additional context**
...

jeddy3:
Done in a plugin in https://github.com/stylelint/stylelint/pull/3364

Issue #2650, opened 2017-06-18
[New rule Idea] Rule to remind to add :focus selector to every :hover for keyboard accessibility.


jitendravyas:
For this purpose https://github.com/postcss/postcss-focus

jeddy3:
I think this is a best handled as a plugin as it's a specific and shifting best practise, rather than configurable rule.

@jitendravyas Perhaps you would like to create your own plugin _pack_ called `stylelint-a11y`, which would group together accessibility plugins?

jitendravyas:
I would really like to start developing stylelint plugins and formatter but current docs are not very clear to me and I'm JS noob too.

jeddy3:
I'd recommend looking at some of the existing plugin (packs). [`stylelint-order`](https://github.com/hudochenkov/stylelint-order/) is a good place to start. 

Within stylelint, [`selector-pseudo-class-case`](https://github.com/stylelint/stylelint/blob/master/lib/rules/selector-pseudo-class-case/index.js) is a fairly simple rule that deals with pseudo-classes. You could use that as a foundation for this `:focus` plugin.

Then it'd be great if you could send a PR that improves the documentation in stylelint based on your experiences.

jitendravyas:
Thanks for the direction.

jeddy3:
I still believe a 3rd party `stylelint-a11y` plugin is the appropriate way forward. It's an exciting and worthwhile endeavour for someone!  Our approach has historically been to push the experimentation of rules of unknown complexity to plugins, _with the prospect of promoting them to core if they fulfil the [criteria for inclusion](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#criteria-for-inclusion)_. @mgifford Would you like to champion this? You appear to have a wealth of a11y experience!

There's a nice precedent for an a11y plugin in the ESLint world: [`eslint-plugin-jsx-a11y`](https://github.com/evcohen/eslint-plugin-jsx-a11y).

> This is also about best practices. You might choose to not address keyboard only users, but you certainly should anticipate that you have some. This isn't a fringe issue.

Quite a few best practices can be enforced at the configuration level. I believe there's also an opportunity here for someone to publish an accessibility-focus configuration (e.g. `stylelint-config-ally`). 

[...]

PR #3355, opened 2018-06-03
Add selector-pseudo-class-focus rule

YozhikM:

Issue #3354 

No.

### DIFF
--- a/docs/user-guide/example-config.md
+++ b/docs/user-guide/example-config.md
@@ -151,6 +151,7 @@ You might want to learn a little about [how rules are named and how they work to
     "selector-no-vendor-prefix": true,
     "selector-pseudo-class-blacklist": string|[],
     "selector-pseudo-class-case": "lower"|"upper",
+    "selector-pseudo-class-focus": true,
     "selector-pseudo-class-no-unknown": true,
     "selector-pseudo-class-parentheses-space-inside": "always"|"never",
     "selector-pseudo-class-whitelist": string|[],
--- a/docs/user-guide/rules.md
+++ b/docs/user-guide/rules.md
@@ -300,6 +300,7 @@ Here are all the rules within stylelint, grouped first [by category](../../VISIO
 -   [`selector-combinator-space-before`](../../lib/rules/selector-combinator-space-before/README.md): Require a single space or disallow whitespace before the combinators of selectors.
 -   [`selector-descendant-combinator-no-non-space`](../../lib/rules/selector-descendant-combinator-no-non-space/README.md): Disallow non-space characters for descendant combinators of selectors.
 -   [`selector-pseudo-class-case`](../../lib/rules/selector-pseudo-class-case/README.md): Specify lowercase or uppercase for pseudo-class selectors.
+-   [`selector-pseudo-class-focus`](../../lib/rules/selector-pseudo-class-focus/README.md): Require or disallow a pseudo-element to the selectors with :hover.
 -   [`selector-pseudo-class-parentheses-space-inside`](../../lib/rules/selector-pseudo-class-parentheses-space-inside/README.md): Require a single space or disallow whitespace on the inside of the parentheses within pseudo-class selectors.
 -   [`selector-pseudo-element-case`](../../lib/rules/selector-pseudo-element-case/README.md): Specify lowercase or uppercase for pseudo-element selectors.
 -   [`selector-pseudo-element-colon-notation`](../../lib/rules/selector-pseudo-element-colon-notation/README.md): Specify single or double colon notation for applicable pseudo-elements.
--- a/lib/rules/index.js
+++ b/lib/rules/index.js
@@ -144,6 +144,7 @@ module.exports = [
   "selector-no-vendor-prefix",
   "selector-pseudo-class-blacklist",
   "selector-pseudo-class-case",
+  "selector-pseudo-class-focus",
   "selector-pseudo-class-no-unknown",
   "selector-pseudo-class-parentheses-space-inside",
   "selector-pseudo-class-whitelist",
--- a/lib/rules/selector-pseudo-class-focus/README.md
+++ b/lib/rules/selector-pseudo-class-focus/README.md
@@ -1,0 +1,34 @@
+# selector-pseudo-class-focus
+
+Checks the presence of a pseudo-class for selectors with `:hover`.
+
+```css
+    a:hover, a:focus { }
+```
+
+This rule considers :focus pseudo-class selector defined in the CSS Specifications.
+
+## Options
+
+### true
+
+The following pattern are considered violations:
+
+```css
+    a:hover { }
+```
+
+The following patterns are *not* considered violations:
+
+```css
+    a:hover, a:focus { }
+```
+
+```css
+     a:focus { }
+```
+
+```css
+    a:hover { }
+    a:focus { }
+```
--- a/lib/rules/selector-pseudo-class-focus/__tests__/index.js
+++ b/lib/rules/selector-pseudo-class-focus/__tests__/index.js
@@ -1,0 +1,30 @@
+"use strict";
+
+const rule = require("..");
+const { messages, ruleName } = rule;
+
+testRule(rule, {
+  ruleName,
+  config: [true],
+
+  accept: [
+    {
+      code: "a:focus { }"
+    },
+    {
+      code: "a:hover, a:focus { }"
+    },
+    {
+      code: "a:hover { } a:focus { }"
+    }
+  ],
+
+  reject: [
+    {
+      code: "a:hover { }",
+      message: messages.expected("a:hover"),
+      line: 1,
+      column: 4
+    }
+  ]
+});
--- a/lib/rules/selector-pseudo-class-focus/index.js
+++ b/lib/rules/selector-pseudo-class-focus/index.js
@@ -1,0 +1,105 @@
+"use strict";
+
+const isCustomSelector = require("../../utils/isCustomSelector");
+const isStandardSyntaxAtRule = require("../../utils/isStandardSyntaxAtRule");
+const isStandardSyntaxRule = require("../../utils/isStandardSyntaxRule");
+const isStandardSyntaxSelector = require("../../utils/isStandardSyntaxSelector");
+const report = require("../../utils/report");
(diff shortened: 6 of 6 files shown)


ai:
This rule will be very good for a11y. And could show that Stylelint is not just a tool to check whitespaces but could also promote best practices.

jeddy3:
@YozhikM Thanks for starting this! The code is looking good.

As I suggested in https://github.com/stylelint/stylelint/issues/2650#issuecomment-388975205, I think this and https://github.com/stylelint/stylelint/pull/3357 should first live in an `stylelint-a11y` plugin (pack). 

Once the pack is published, you can send an PR to update [the list of plugins](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/plugins.md).

ai:
@jeddy3 what is your reason for moving it to a plugin?

Here are my thoughts:

1. This rule should be enabled in ost of projects.
2. It is fit all core plugins rules.
3. RIght now Stylelint needs more rules, which are really helpful in promoting best practices. Most of CSS enthusiasts ignores Stylelint since they see it as a tool just for linting spaces. Not to really important things. If we will have more a11y plugins in core, it will change their minds (plugin will not solve the problem, since nobody looks in plugins ;) )

evilebottnawi:
@jeddy3 i agree with @ai

hudochenkov:
I would agree with @ai on this point. I would include this rule in the core, but won't include [another one](https://github.com/stylelint/stylelint/pull/3357). Because in the current state it's for iPhone/iPad users. And very few people know about this feature.

evilebottnawi:
@hudochenkov webkit also support this (i.e. chrome also and chromium based)

jeddy3:
> what is your reason for moving it to a plugin?

I'm going to have to dig into my memory and recall some original design decisions, so bear with me...

When we first starting working on stylelint we looked to CSSLint for inspiration. We noticed that quite a few [of the rules](https://github.com/CSSLint/csslint/wiki/Rules) were now redundant as they were based on "best practises" that were, well, no longer best practice. As such, we made an early design decision to avoid adding such rules to stylelint, instead, we'd favour rules _that could be configured_ to enforce whatever a team needs (or deems the best practice at the time) by applying limits to _targeted constructs within the language_.

That's why we have rules like [`declaration-property-value-blacklist`](https://github.com/stylelint/stylelint/blob/master/lib/rules/declaration-property-value-blacklist/README.md) instead of rules like `transition-no-all` and [`disallow-outline-none`](https://github.com/CSSLint/csslint/wiki/Disallow-outline%3Anone) (from CSSLint).

This design decision is captured in [the VISION doc](https://github.com/stylelint/stylelint/blob/master/VISION.md#limit-language-features).

A little later we realised that not everything a team might need could be achieved using just these configurable rules, so we added a plugin system. This allowed users to create rules which could do more than just apply limits to targeted constructs, whilst keeping the core stable _and relevant_.

In this particular instance, it [looks like there are two ways](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Accessibility#Using_native_keyboard_accessibility) to maintain accessibility:

- specifying `:focus` and `:hover` together
- relying on the browser default `:focus` behavior and not overiding it

I don't believe this rule can account for the latter, which is why it might not meet the "Has a clear and unambiguous finished state" [rule of inclusion](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#criteria-for-inclusion). Anyway, I think this point is moot as this is the kind of stuff that can get experimented on (and quickly iterated on) within a plugin.

> Most of CSS enthusiasts ignores Stylelint since they see it as a tool just for linting spaces

It sounds like we need to do a better job of promoting the [possible errors](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#possible-errors) and [limiting](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#limit-language-features) rules. I'm writing an article at the moment that highlights the value of the [limiting rules](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#limit-language-features). When I get around to finishing it it will hopefully help us promote this feature of stylelint.

> (plugin will not solve the problem, since nobody looks in plugins ;) )

There are some excellent plugins that I see in regular use. `stylelint-scss`, `stylelint-order` and `stylelint-csstree-validator` and `stylelint-selector-bem-pattern` all spring to mind.

> maybe it is normal put this rules in core, if somebody starts working on a11y plugin, we can move this out of stylelint in 10 version

We went through the pain of doing it that way around circa [7.8.0](https://github.com/stylelint/stylelint/blob/master/CHANGELOG.md#780). I thought the lesson we learnt was that it's less painful to move plugins to core than to move built-in rules out.

----

I'm not seeing any augments in the comments above that invalidate the original design decisions. Unless I'm missing something here? I agree there is an user education issue, though. We need to do a better job at promoting and signposting (within the docs and outside of them) the [possible errors](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#possible-errors) and [limiting](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#limit-language-features) rules, as well as the plugin system (and this `ally` plugin). 

An `a11y` stylelint plugin pack is an exciting proposition! Perhaps @YozhikM would like to create and champion it. Within the stylelint org itself, we can help out by promoting it.

It feels like this is approach will balance the user's need with the still valid original design decisions that have kept the built-in rules relevant. Thoughts?

YozhikM:
@jeddy3 Excellent. Work in progress. Who and where can I ask questions if they appear?

hudochenkov:
@YozhikM thank you for starting working on this plugin! I think you can create an issue in this repository and we'll help you.

jeddy3:
> Work in progress

Good stuff! 

Both [stylelint-scss](https://github.com/kristerkari/stylelint-scss) and [stylelint-order](https://github.com/hudochenkov/stylelint-order) are plugin packs. You can use them for inspiration if you like.

ai:
> There are some excellent plugins that I see in regular use. stylelint-scss, stylelint-order and stylelint-csstree-validator and stylelint-selector-bem-pattern all spring to mind.

And only 1/5 of Stylelint users use these plugins http://www.npmtrends.com/stylelint-vs-stylelint-order-vs-stylelint-scss

evilebottnawi:
I still think it will be great as official plugin

hudochenkov:
> And only 1/5 of Stylelint users use these plugins

Not everyone need these plugins :)

ai:
I am making slides about PostCSS and want to promote Stylelint. Here is an overview of our “Possible errors” rules:

1. Prevent unknown props/selectors/colors, empty values, and wrong syntax
2. Prevent duplicate property
3. Prevent `margin` after `margin-top`
4. Prevent descending specificity in selectors

Only 2 and 3 issues are really popular. Error 1 happens sometimes.

Unfortunately, it doesn’t compete with ESLint [“Possible Errors” rules](https://eslint.org/docs/rules/#possible-errors).

This is why I still believe that a11y rules in the core will be extremely good for promotion. We need more rules to prevent errors.

hudochenkov:
> Unfortunately, it doesn’t compete with ESLint “Possible Errors” rules.

Also, JavaScript is a programming language, while CSS isn't. More errors could be identified by code analysis in JavaScript, then in CSS.

evilebottnawi:
I am glad to see more possible errors rules and i have some ideas for new rules, but i think we should concentrate now on bug fixes and build-in support `html/markdown/css-in-js`

hudochenkov:
@evilebottnawi I would like to see more error targeted rules, rather stylistic, as well (we might cover all stylistic rules already :)).

Please, create new issues for your rules ideas. New contributors are more willingly work on new stuff rather fixing bugs, in my opinion.

evilebottnawi:
@hudochenkov In near future :+1:

ai:
@ntwb when we compare ESLint and Stylelint rules by the number we are making mistakes. Stylelint has 10 rules just to check syntax errors (unknown properties, selectors, etc). If you group rules by error type we will have only 4 errors, which STylelint prevents

@hudochenkov we have JS in Stylelint to be smart. Simpler CSS syntax should even help us in finding real errors.

jeddy3:
> I am making slides about PostCSS and want to promote Stylelint. 

That's fantastic! 

I'd suggest highlighting the [limiting rules](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#limit-language-features) rather than focusing on the ones that [check for possible errors](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md#possible-errors). As @hudochenkov pointed out, CSS and JavaScript are two very different languages. They have different problems that need addressing with linters. 

For me personally, I find that _inconsistency_ is the biggest hurdle to maintaining a stylesheet within a team. There are so many ways for a developer to achieve the same outcome in CSS. As such, a large proportion of stylelint rules were specifically built to help teams combat these inconsistencies.

For example, a team might:

* embrace a "mobile-first" approach and so use `"media-feature-name-whitelist": ["min-width"]` to ensure only `min-width`, rather than `max-width`, media queries are used
* want to build a fluid site that proportionally scales and so use `unit-whitelist: ["%", "em", "rem"]` to allow only proportional units
* only use BEM and so use `selector-no-type: true`, `selector-no-universal: true` etc to make sure only class selectors are used

These are just three examples. Hopefully, it's clear how stylelint can help a team enforce the consistent use of any number of other methodologies. 

There are over 50 limiting rules in stylelint, including fine-grained ones like [`declaration-property-value-blacklist`](https://github.com/stylelint/stylelint/blob/master/lib/rules/declaration-property-value-blacklist/README.md) and the [upcoming](https://github.com/stylelint/stylelint/pull/3320) `media-feature-name-value-whitelist`. I think the limiting rules are worth highlighting as I believe:

1. they give teams the power to solve the problem of inconsistency
2. these rules are pretty unique to stylelint

I think we can close these PRs and issues now that https://github.com/stylelint/stylelint/pull/3364 has been merged.

We can continue the discussion about how best to promote the plugin system (and plugins themselves) in https://github.com/stylelint/stylelint/issues/3362.

@YozhikM Thanks again for your work on these rules and porting them to the `a11y` plugin! Feel free to jump into the https://github.com/stylelint/stylelint/issues/3362 discussion if you have any suggestions about how we can better promote your plugin and the plugin system itself.

Issue #3356, opened 2018-06-03
[Rule]: prefers-reduced-motion


YozhikM:
**Is your feature request related to a problem? Please describe.**
Solves the problem of the availability of sites for people who have the disabled animation for health reasons, for example.

It is a non-vendor-prefixed declaration that allows developers to “create styles that avoid large areas of motion for users that specify a preference for reduced motion in System Preferences.”

[webkit.org](https://webkit.org/blog/7551/responsive-design-for-motion/)
[css-tricks.org](https://css-tricks.com/introduction-reduced-motion-media-query/)

**Describe the solution you'd like**
Adding a new rule.

**Describe alternatives you've considered**
...

**Additional context**
...

jeddy3:
Done in a plugin in https://github.com/stylelint/stylelint/pull/3364

PR #3357, opened 2018-06-03
Add media-prefers-reduced-motion rule

YozhikM:

Issue #3356 

No.

### DIFF
--- a/docs/user-guide/example-config.md
+++ b/docs/user-guide/example-config.md
@@ -97,6 +97,7 @@ You might want to learn a little about [how rules are named and how they work to
     "media-feature-parentheses-space-inside": "always"|"never",
     "media-feature-range-operator-space-after": "always"|"never",
     "media-feature-range-operator-space-before": "always"|"never",
+    "media-prefers-reduced-motion": true,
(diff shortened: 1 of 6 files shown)


ai:
Also good rule for a11y

jeddy3:
See https://github.com/stylelint/stylelint/pull/3355#issuecomment-394389845 about having this live in a plugin first.

> Also good rule for a11y

We can promote this plugin in the documentation e.g. 

* **Support for plugins**: It's easy to create your own rules and add them to the linter, or make use of plugins written by the community. For example, there are plugins for [Scss](..) and [a11y](..).

evilebottnawi:
@jeddy3 looks like no one starting `a11y` plugin, i do not think that this can happen, maybe it is normal put this rules in core, if somebody starts working on `a11y` plugin, we can move this out of stylelint in 10 version. We should will be responsive to community, plugin system is awesome, but it is doesn't make sense when we have little community :disappointed:

jeddy3:
Ported to a plugin in https://github.com/stylelint/stylelint/pull/3364

@YozhikM Thanks again.

PR #3360, opened 2018-06-04
Fix ignores inconsistency in rule READMEs

### LINKED ISSUE #3358: Make the extended descriptions of rules consistent
Ref: https://github.com/stylelint/stylelint/pull/3320#discussion_r192720030

We settled on "This rule ..." convention a while ago, but some of the rules still use the "**Caveat:**" convention. We should use the "This rule ..." convention everywhere to make it easier for first times creating new rules.

e.g. The following:

"**Caveat:** Media feature names within range or boolean context are ignored."

Should be written as:

"This rules ignores media feature within range or boolean context."

jeddy3:

Closes #3358

No, it's self-explanatory.

### DIFF
--- a/lib/rules/at-rule-empty-line-before/README.md
+++ b/lib/rules/at-rule-empty-line-before/README.md
@@ -10,7 +10,10 @@ a {}
  *   This line */
 ```
 
-If the at-rule is the very first node in a stylesheet then it is ignored. `@import` in Less will also be ignored.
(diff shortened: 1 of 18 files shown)


jeddy3:
> Maybe we can add to some ignores what it is can be solved, like PR welcome. 

Good point, but let's handle that in a separate issue/PR.

PR #3363, opened 2018-06-05
Add non-obvious list of rule README conventions

jeddy3:

Ref: https://github.com/stylelint/stylelint/pull/3320#discussion_r192897819

I felt like these two conventions were the most non-obvious and difficult to glean from the existing READMEs.

For anyone interested, we use "This rule" to help distinguish the stylelint rule from CSS rules.

### DIFF
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -189,10 +189,19 @@ Each rule must be accompanied by a README, fitting the following format:
 7.  Example patterns that are *not* considered violations (for each option value).
 8.  Optional options (if applicable).
 
-Look at the READMEs of other rules to glean more conventional patterns. These includes:
+Look at the READMEs of other rules to glean more conventional patterns. These include:
 
--   Aligning the arrows within the prototypical code example with the beginning of the construct being highlighted.
 -   Using "This rule" to refer to the rule e.g. "This rule ignores ..."
+-   Aligning the arrows within the prototypical code example with the beginning of the construct being highlighted.
+-   Aligning the text within the prototypical code example as far to the left as possible.
+
+For example:
+
+```css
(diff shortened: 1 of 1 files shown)


jeddy3:
@MatthiasKunnen I've requested your review on this, if you don't mind, as it's based on your feedback. Thanks.

docs/developer-guide/rules.md
```diff
@@ -189,7 +189,10 @@ Each rule must be accompanied by a README, fitting the following format:
 7.  Example patterns that are *not* considered violations (for each option value).
 8.  Optional options (if applicable).
 
-Look at the READMEs of other rules to glean more conventional patterns.
+Look at the READMEs of other rules to glean more conventional patterns. These includes:
```
MatthiasKunnen: Grammar-wise, I believe it should be either _These include_ or _This includes_. These and those should be followed by a singular verb.

docs/developer-guide/rules.md
```diff
@@ -189,7 +189,10 @@ Each rule must be accompanied by a README, fitting the following format:
...
+
+-   Aligning the arrows within the prototypical code example with the beginning of the construct being highlighted.
```
MatthiasKunnen: I think this would benefit from an example. E.g.
````diff
 @media screen and (min-width: 768px) {}
-/**                    ↑        ↑
+/**                ↑          ↑
  *         These features and these values */
````

Comment (jeddy3) [reply docs/developer-guide/rules.md]: Yep. This also highlights another subtle convention... the description should be as left aligned as possible e.g.

```css
 @media screen and (min-width: 768px) {}
/**                 ↑          ↑
  *       These names and values */
```

I'll add an example and additional bullet.

Issue #3380, opened 2018-06-09
Fix false positives on non-standard syntax in declaration-block-no-shorthand-property-overrides


gucong3000:
**Describe the bug? (A clear and concise description of what the bug is)**

Error was missed for `decl` when it not in `rule`

**Which rule, if any, is this issue related to?**

`declaration-block-no-shorthand-property-overrides`

**What CSS is needed to reproduce this issue?**

```css
a { @media (min-width: 768px) { padding-left: 10px; padding: 20px; } }
```

**What stylelint configuration is needed to reproduce this issue?**

```json
{
  "rules": {
    "declaration-block-no-shorthand-property-overrides": true
  }
}
```

**Which version of stylelint are you using?**

all

**Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?**

Relate to all non-standard syntax.

**What did you expect to happen?**

Should throw a error.

**What actually happened (e.g. what warnings or errors you are getting)?**

Error was missed

hudochenkov:
Can't reproduce in demo https://stylelint.io/demo/

gucong3000:
> Can't reproduce in demo https://stylelint.io/demo/

Updated example to reproduce this issue.

gucong3000:
Related to: https://github.com/emotion-js/emotion/issues/686#issuecomment-394077511

hudochenkov:
```css
a {
  padding-left: 10px;
  padding: 20px;
  padding: 20px;
}
```

Works as expected.

```html
<a stylelint="padding-left: 10px; padding: 20px; padding: 20px;"></a>
```

Should not work, because it's a very non-standard thing. Stylelint shouldn't care about this. 

@gucong3000 maybe I'm missing something. Stylelint doesn't suppot CSS-in-JS officially. Discussion about this in https://github.com/stylelint/stylelint/issues/3264

gucong3000:
> it's a very non-standard thing.

It's a spelling mistake, and the right code is:
```html
<a style="padding-left: 10px; padding: 20px; padding: 20px;"></a>
```

evilebottnawi:
WIP

evilebottnawi:
@jeddy3 looks already fixed https://github.com/stylelint/stylelint/pull/3381?

PR #3381, opened 2018-06-09
Fix false positives on non-standard syntax in declaration-block-no-*

gucong3000:

#3380 
emotion-js/emotion#686 (comment)

No, it's self explanatory.

### DIFF
--- a/lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
+++ b/lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
@@ -49,10 +49,6 @@ testRule(rule, {
 
   reject: [
     {
-      code: "color: pink; color: orange",
-      message: messages.rejected("color")
-    },
-    {
       code: "a { color: pink; color: orange }",
       message: messages.rejected("color")
     },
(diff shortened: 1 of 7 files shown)


lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.js
```diff
@@ -35,6 +35,11 @@ testRule(rule, {
   ],
 
   reject: [
+    {
+      code:
+        "margin-left: 10px; margin-right: 10px; margin-top: 20px; margin-bottom: 30px;",
+      message: messages.expected("margin")
+    },
```
evilebottnawi: Let's add test with html and style tag

gucong3000: Done.

lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
```diff
@@ -18,6 +18,10 @@ testRule(rule, {
       code: "a { color: pink; { &:hover { color: orange; } } }",
       description: "spec nested"
     },
+    {
+      code: "color: pink; @media { color: orange; }",
```
jeddy3: As per [the developer guide](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#write-tests):

> Ensure you use standard CSS syntax by default

I don't believe a declaration outside of rule-set is valid. What are we trying to test here?

If it's the content of a `style=""` attribute then how does this differ from the `syntax: html` tests below which are more explicit?

lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
```diff
@@ -48,6 +52,10 @@ testRule(rule, {
   ],
 
   reject: [
+    {
+      code: "color: pink; color: orange",
```
jeddy3: If this is for `style=""` attributes? If so, how does it differ from the `syntax: html` tests below?

jeddy3:
@evilebottnawi If you've time to review this today, then I'll be able to do a release before I go away for the weekend.

evilebottnawi:
@jeddy3 LGTM

jeddy3:
> -   Fixed: `declaration-block-no-*` false positives for non-standard syntax ([#3381](https://github.com/stylelint/stylelint/pull/3381)).

PR #3389, opened 2018-06-11
Standardized object input documentation for rules

MatthiasKunnen:
> Which issue, if any, is this issue related to?

Closes #3372

> Is there anything in the PR that needs further explanation?

I've not added a paragraph in the developer-guide/rules.md but this can be added if requested.

Issue #3008, opened 2017-11-06
Support regex literals in all options that accept regexes


sindresorhus:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

Many of the Stylelint rule options supports regexes, but they require having the regexes in a string. For JS based configs, it would be much nicer to specify them as regex literals. Some rules supports this already, like the `selector-class-pattern` and `selector-id-pattern` rules, while the `declaration-property-value-blacklist` rule does not and require string `'/thin/'`.

> Which rule, if any, is this issue related to?

`declaration-property-value-blacklist`, but really any rule that supports regexes, should also support regex literals.

> What CSS is needed to reproduce this issue?

Does not apply.

> What stylelint configuration is needed to reproduce this issue?

Does not apply.

> Which version of stylelint are you using?

`8.2.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

`stylelint --config myconfig *.css`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

I expected to be able to use regex literals in options.

> What actually happened (e.g. what warnings or errors you are getting)?

It didn't work.

jeddy3:
Sounds good.

I suspect these inconsistencies crept in as new rules were added with this feature, but the feature wasn't backported to the older rules.

Labelling as enhancement and help wanted.

kevva:
I think https://github.com/stylelint/stylelint/blob/master/lib/utils/matchesStringOrRegExp.js should be updated instead of the rules themselves. You could use https://github.com/sindresorhus/is-regexp to check if it's a regex.

ntwb:
Hey, @kevva Did you take a look at this? If not would you be interested still?

jeddy3:
> I think https://github.com/stylelint/stylelint/blob/master/lib/utils/matchesStringOrRegExp.js should be updated

This was done as part of https://github.com/stylelint/stylelint/pull/3087.

> instead of the rules themselves

I believe that all there is left to do is update the options validator config in each applicable rule so that RegExp are accepted. We'd also want to update/add some appropriate tests. Examples of both can be found in https://github.com/stylelint/stylelint/pull/3087.

CAYdenberg:
This is a bit of a chore but here is everything that I think needs to be updated. I got tired of checking the code when I got into the secondary options, so it's possible some of those are already covered - just check the box or comment in thread if so.

[...]

PR #3390, opened 2018-06-11
Standardise options documentation

### LINKED ISSUE #3372: Document standardized object input for rules
The current examples for rules that take an object as input is as follows: 

```
`object`: `{
  "unprefixed-property-name": ["array", "of", "values"],
  "unprefixed-property-name": ["/regex/", "non-regex"]
}`
```

and renders as such
Input object on https://stylelint.io
Both the code and the rendered view are not easy to write and difficult to read both in Markdown as rendered. 

I would like to propose a new format for objects as input. 

````
```json
{
  "unprefixed-property-name": ["array", "of", "values"],
  "unprefixed-property-name": ["/regex/", "non-regex"]
}
```
````

This would be rendered as follows

```json
{
  "unprefixed-property-name": ["array", "of", "values"],
  "unprefixed-property-name": ["/regex/", "non-regex"]
}
```

The advantages are:
* improved code readability
* improved code writability
* improved readability on https://stylelint.io/user-guide/rules/
* syntax highlighting
* a common used format that developers know, JSON

**Describe alternatives you've considered**
N/A

**Additional context**
This format is currently used for the following rules:

* declaration-property-unit-blacklist
* declaration-property-unit-whitelist
* declaration-property-value-blacklist
* declaration-property-value-whitelist
* and the media-feature-name-value-whitelist that is under review at #3320 

Also take a look at this comment tread https://github.com/stylelint/stylelint/pull/3320#discussion_r192722477 for an earlier discussion that led to this issue.

jeddy3:

Closes #3372

I dug deeper while reviewing PR https://github.com/stylelint/stylelint/pull/3389 as I was unsatisfied with the consistency of options _across the board_, rather than just with the rules that accept objects.

This PR is a WIP to solicit feedback. It contains examples for the different types of options that rules accept e.g. arrays, strings, booleans, regexs, objects etc.

If a rule removes prefixes, I moved this to the rule description. I think dealing with prefixes is an edge-case, and so we don't need to pollute the "Options" section with it.

I think the `"array", "of", ...` stuff is confusing. I believe it's clearer to consistently specify what types are accepted first, then go into more detail only when necessary. Usually, the `"array", "of", ...` stuff is redundant as the "Given:" example communicates this better.

The proposed solution also takes into https://github.com/stylelint/stylelint/issues/3008. The existing approach doesn't really work here e.g. when`property-whitelist` is enhanced to accept naked regex, the existing approach will produce:

> ...
>
> ## Options
>
> `array|regex|string`: `["array", "of", "unprefixed", "properties" or "regex" or regex]|"property"|"/regex/"|regex`
>
> If a string is surrounded with `"/"` (e.g. `"/^background/"`), it is interpreted as a regular expression. This allows, for example, easy targeting of shorthands: `/^background/` will match `background`, `background-size`, `background-color`, etc.
>
> Given:
>
> ```js
> ["display", "animation", "/^background/"]
> ```
>
> ...

Becomes:

> ...
>
> This rule removes vendor prefixes before any checks.
>
> ## Options
>
> Type: `array|regex|string`
>
> An `array` can contain any number of `string` or `regex`.
>
> Any `string` surrounded with `"/"` (e.g. `"/string/"`) is interpreted as a regular expression.
>
> Regular expressions allow, for example, the easy targeting of shorthand properties: `/^background/` will match `background`, `background-size`, `background-color`, etc.
>
> Given:
>
> ```js
> ["text-rendering", "animation", "/^background/"]
> ```
>
> ...

Suggestions or thoughts?

### DIFF
--- a/lib/rules/declaration-property-value-whitelist/README.md
+++ b/lib/rules/declaration-property-value-whitelist/README.md
@@ -8,20 +8,30 @@ a { text-transform: uppercase; }
  * These properties and these values */
 ```
 
+This rule removes vendor prefixes before any checks.
+
 ## Options
 
-`object`: `{
-  "unprefixed-property-name": ["array", "of", "values"],
-  "unprefixed-property-name": ["/regex/", "non-regex"]
-}`
+Type: `object`  
+Structure:   
 
-If a property name is found in the object, only its whitelisted property values are allowed. This rule complains about all non-matching values. (If the property name is not included in the object, anything goes.)
+```json
+{
+  string: array|string
+}
+```
 
-If a property name is surrounded with `"/"` (e.g. `"/^animation/"`), it is interpreted as a regular expression. This allows, for example, easy targeting of shorthands: `/^animation/` will match `animation`, `animation-duration`, `animation-timing-function`, etc.
+Each property in the `object` is compared against a declaration's property name. The paired value in the `object` represents the allowed values for that property name.
+
+Any `string` surrounded with `"/"` (e.g. `"/^string/"`) is interpreted as a regular expression.
+
+Regular expressions allow, for example, the easy targeting of shorthand properties: `/^animation/` will match `animation`, `animation-duration`, `animation-timing-function`, etc.
 
 The same goes for values. Keep in mind that a regular expression value is matched against the entire value of the declaration, not specific parts of it. For example, a value like `"10px solid rgba( 255 , 0 , 0 , 0.5 )"` will *not* match `"/^solid/"` (notice beginning of the line boundary) but *will* match `"/\\s+solid\\s+/"` or `"/\\bsolid\\b/"`.
 
 Be careful with regex matching not to accidentally consider quoted string values and `url()` arguments. For example, `"/red/"` will match value such as `"1px dotted red"` as well as `"\"foo\""` and `"white url(/mysite.com/red.png)"`.
+
+If a property name is found in the object, only its whitelisted property values are allowed. This rule complains about all non-matching values. (If the property name is not included in the object, anything goes.)
 
 Given:
 
--- a/lib/rules/property-blacklist/README.md
+++ b/lib/rules/property-blacklist/README.md
@@ -12,7 +12,7 @@ This rule removes vendor prefixes before any checks.
 
 ## Options
 
-`array|string`
+Type: `array|string`
 
 Any `string` surrounded with `"/"` (e.g. `"/^string/"`) is interpreted as a regular expression. This allows, for example, easy targeting of shorthands: `/^background/` will match `background`, `background-size`, `background-color`, etc.
 
--- a/lib/rules/property-blacklist/README.md
+++ b/lib/rules/property-blacklist/README.md
@@ -8,11 +8,13 @@ a { text-rendering: optimizeLegibility; }
  * These properties */
 ```
 
+This rule removes vendor prefixes before any checks.
+
 ## Options
 
-`array|string`: `["array", "of", "unprefixed", "properties" or "regex"]|"property"|"/regex/"`
+`array|string`
 
-If a string is surrounded with `"/"` (e.g. `"/^background/"`), it is interpreted as a regular expression. This allows, for example, easy targeting of shorthands: `/^background/` will match `background`, `background-size`, `background-color`, etc.
+Any `string` surrounded with `"/"` (e.g. `"/^string/"`) is interpreted as a regular expression. This allows, for example, easy targeting of shorthands: `/^background/` will match `background`, `background-size`, `background-color`, etc.
 
 Given:
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,3 +8,6 @@ indent_size = 2
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
--- a/lib/rules/at-rule-blacklist/README.md
+++ b/lib/rules/at-rule-blacklist/README.md
@@ -10,9 +10,11 @@ Specify a blacklist of disallowed at-rules.
 
 This rule ignores `@import` in Less.
 
+This rule removes vendor prefixes before any checks.
+
 ## Options
 
-`array|string`: `["array", "of", "unprefixed", "at-rules"]|"at-rule"`
+Type: `array|string`
 
 Given:
 
--- a/lib/rules/at-rule-empty-line-before/README.md
+++ b/lib/rules/at-rule-empty-line-before/README.md
@@ -19,7 +19,8 @@ The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofix
 
 ## Options
 
-`string`: `"always"|"never"`
+Type: `string`  
+Value: `"always"|"never"`
 
 ### `"always"`
 
--- a/lib/rules/at-rule-name-case/README.md
+++ b/lib/rules/at-rule-name-case/README.md
@@ -16,7 +16,8 @@ The `--fix` option on the [command line](../../../docs/user-guide/cli.md#autofix
 
 ## Options
 
-`string`: `"lower"|"upper"`
+Type: `string`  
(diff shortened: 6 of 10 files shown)


lib/rules/declaration-property-value-whitelist/README.md
```diff
@@ -8,21 +8,31 @@ a { text-transform: uppercase; }
...
-  "unprefixed-property-name": ["array", "of", "values"],
-  "unprefixed-property-name": ["/regex/", "non-regex"]
-}`
+Type: `object`  
+Structure:   
 
-If a property name is found in the object, only its whitelisted property values are allowed. This rule complains about all non-matching values. (If the property name is not included in the object, anything goes.)
+```json
+{
+  string: array|string
```
hudochenkov: Missing `regex` here.

lib/rules/property-blacklist/README.md
```diff
@@ -8,11 +8,13 @@ a { text-rendering: optimizeLegibility; }
  * These properties */
 ```
 
+This rule removes vendor prefixes before any checks.
+
 ## Options
 
-`array|string`: `["array", "of", "unprefixed", "properties" or "regex"]|"property"|"/regex/"`
+Type: `array|string`
```
hudochenkov: Missing `regex` here.

MatthiasKunnen:
I agree with the structure examples. However, I think we're repeating ourselves too much. 

To get to know what was possible with Stylelint, I read all the rules. Having to read the same thing over and over again adds unnecessary strain. I suggest we employ a reusable footnote to refer to regex usage and its examples. This way we don't have to repeat the same thing in code, giving us a single place to edit this text if necessary and allowing the reader to consume the rule description faster.

jeddy3:
> However, I think we're repeating ourselves too much.

That's a fair point.

> To get to know what was possible with Stylelint, I read all the rules. 

That's impressive :) I've not heard of anyone do that before.

> allowing the reader to consume the rule description faster.

When building the documentation, we assumed the most common scenario would be one member of a team configures and rolls out stylelint, then the other members of the team only dig into the documentation for an individual rule _when then encounter a violation_. As such, we designed each README to be standalone i.e. everything a user, who didn't necessarily set up stylelint, needs is contained within that one file.

Perhaps this assumption isn't valid... but it feels right.

How do reusable footnotes work in markdown files e.g. would it be a link off to another page?

We should also consider that the descriptions tend to be contextual e.g. `property-blacklist` will have "Regular expressions allow, for example, the easy targeting of shorthand properties: `/^background/` will match `background..."

If there's a way of using footnotes that considers the above and improves the experience for users who read through all the rules, then I think we should adopt it. If not, we can roll with what's in the PR as it is.

lib/rules/property-blacklist/README.md
jeddy3: Not quite yet. It'll fail validation until https://github.com/stylelint/stylelint/issues/3008 is done

lib/rules/declaration-property-value-whitelist/README.md
MatthiasKunnen: Add double quotes around the property and value to make this proper JSON

MatthiasKunnen:
As I see it, there are 3 different ways a rule can take a regex as input.

As primary input or array, e.g. `selector-class-pattern` and `property-blacklist`.
As an object property, e.g. `declaration-property-unit-blacklist`.
As both, e.g. `declaration-property-value-whitelist`.

If we create a basic explanation for these three we can reuse them across all the rules.

# Example README code

## property-blacklist
```diff
-If a string is surrounded with `"/"` (e.g. `"/^background/"`), it is interpreted as a regular expression. This allows, for example, easy targeting of shorthands: `/^background/` will match `background`, `background-size`, `background-color`, etc.
+This rule supports regexes partial:regex.
```

## declaration-property-unit-blacklist
```diff
-If a property name is surrounded with `"/"` (e.g. `"/^animation/"`), it is interpreted as a regular expression. This allows, for example, easy targeting of shorthands: `/^animation/` will match `animation`, `animation-duration`, `animation-timing-function`, etc.
+ This rule supports regexes on the property partial:regex-property.
```

## declaration-property-value-whitelist
```diff
Each property in the `object` is compared against a declaration's property name. The paired value in the `object` represents the allowed values for that property name.

-Any `string` surrounded with `"/"` (e.g. `"/^string/"`) is interpreted as a regular expression.
+This rules supports regexes on properties and values partial:regex-both.
-
-Regular expressions allow, for example, the easy targeting of shorthand properties: `/^animation/` will match `animation`, `animation-duration`, `animation-timing-function`, etc.
-
-The same goes for values. Keep in mind that a regular expression value is matched against the entire value of the declaration, not specific parts of it. For example, a value like `"10px solid rgba( 255 , 0 , 0 , 0.5 )"` will *not* match `"/^solid/"` (notice beginning of the line boundary) but *will* match `"/\\s+solid\\s+/"` or `"/\\bsolid\\b/"`.
-
-Be careful with regex matching not to accidentally consider quoted string values and `url()` arguments. For example, `"/red/"` will match value such as `"1px dotted red"` as well as `"\"foo\""` and `"white url(/mysite.com/red.png)"`.
```

# Example regex explanation

## Primary input
> If a string is surrounded with `"/"` (e.g. `"/width$/"`), it is interpreted as a regular expression. This allows, for example, easy targeting of shorthands: `/width$/` will match `min-width`, `width`, etc.

## As an object property
> If a property is surrounded with `"/"` (e.g. `"/width$/"`), it is interpreted as a regular expression. This allows, for example, easy targeting of shorthands: `/width$/` will match `min-width`, `width`, etc.

## On both property and values
> Any `string` surrounded with `"/"` (e.g. `"/^string/"`) is interpreted as a regular expression. This allows, for example, easy targeting of shorthands: `/width$/` will match `min-width`, `width`, etc.
>
> The same goes for values. Keep in mind that a regular expression value is matched against the entire value of the declaration, not specific parts of it. For example, a value like `"10px solid rgba( 255 , 0 , 0 , 0.5 )"` will *not* match `"/^solid/"` (notice beginning of the line boundary) but *will* match `"/\\s+solid\\s+/"` or `"/\\bsolid\\b/"`.

hudochenkov:
> > To get to know what was possible with Stylelint, I read all the rules.
> 
> That's impressive :) I've not heard of anyone do that before.

I did it when I started using stylelint (it was version 6 or 7). Still use my config :)

jeddy3:
> I did it when I started using stylelint (it was version 6 or 7).

It looks like I've underestimated this use case :)

> We could implement this by replacing the key, in this example (::partial:regex*) with a superscript footnote number¹ on the website and adding the corresponding partial as a footnote. 

@MatthiasKunnen I've not used markdown partials before. They do look like they'll be useful in this instance, though. If you're familiar with them, and you think they'll work with the website, I think it's worth giving them a go. As you're a member of the stylelint org, you can push commits to this PR. Do you want to pick up this PR and try the partial approach?

MatthiasKunnen:
I've not used them before either, my proposal is purely theoretical. I believe some debate is required to discuss the merits of markdown partials and the desired functionality.

Do we want partial content to be down in the footnotes?  
Do we want to support basic _partial key_ -> _partial content_ replacement?

That being said, while I currently don't have any knowledge regarding the README -> website conversion, I'm willing to take a look and come with a proposal PR. This PR would be partial/footnote support only, serve as a clearer explanation of my intent/proposal, and could later be used in this PR, when approved.

jeddy3:
> This PR would be partial/footnote support only, serve as a clearer explanation of my intent/proposal, and could later be used in this PR, when approved.

SGTM

jeddy3:
> Do we want partial content to be down in the footnotes?
> Do we want to support basic partial key -> partial content replacement?

I guess go with whatever works in your PoC. I don't think any of us have experience with markdown partials, so it's not possible to comment on these questions.

If you can get something working with partials, then we can go with that. If partials don't end up working out, we can go just go for a bit of simple duplication in the READMEs.

hudochenkov:

PR #3394, opened 2018-06-13
Add link to stylelint-no-indistinguishable-colors plugin

ierhyna:
Rule `no-indistinguishable-colors` [was deprecated](https://github.com/stylelint/stylelint/blob/master/CHANGELOG.md#780) in favour of plugin. So I created plugin `stylelint-no-indistinguishable-colors` and added it to the list in `plugins.md`.


ntwb:
Welcome to the stylelint organisation @ierhyna 👋

PR #3284, opened 2018-04-29
Update to postcss-selector-parser@5

chriseppstein:
This PR supercedes https://github.com/stylelint/stylelint/pull/3275

The `postcss-selector-parser@4.0.1` release is not made yet, I'm waiting for a code review from @ai on https://github.com/postcss/postcss-selector-parser/pull/140. You can pull that branch and run tests locally after `npm link`ing the project to this one.

I'm going to leave a few comments on this PR so you can review the things I had doubts about or that you may want to pay close attention to.

This is a courtesy PR, I don't usually have the time to do this, but it was a big update, and I wanted to be sure there weren't any other regressions. If you want changes to this PR, please make them, I can't commit to any iterations.

PR #3395, opened 2018-06-13
Fix false positives for variables in selector-type-case

### LINKED ISSUE #3391: Fix false positives for variables in selector-type-case
> What is the problem?

`selector-type-case` has no customisation settings, and so fails with styled-components’ `$dummyValue`

eg.

`✖  Expected "$dummyValue" to be "$dummyvalue"                        selector-type-case`

an ignore setting akin to `selector-type-no-unknown`: `'selector-type-no-unknown': [true, {ignoreTypes: ['$dummyValue']}],`

eg.

`'selector-type-case': ['lower', {ignoreTypes: ['/dummyValue/']}],`

Thanks!

YozhikM:

Closes #3391 

No.

### DIFF
--- a/lib/utils/isStandardSyntaxTypeSelector.js
+++ b/lib/utils/isStandardSyntaxTypeSelector.js
@@ -36,6 +36,23 @@ module.exports = function(node /*: Object*/) /*: boolean*/ {
     return false;
   }
 
+  let normalizedValue = node.value;
+
+  // Ignore operators before variables (example -$variable)
+  if (/^[-+*/]/.test(node.value[0])) {
+    normalizedValue = normalizedValue.slice(1);
+  }
+
+  // SCSS variable
+  if (normalizedValue[0] === "$") {
+    return false;
+  }
+
+  // Less variable
+  if (normalizedValue[0] === "@") {
+    return false;
+  }
+
   if (node.value[0] === "%") {
     return false;
   }
--- a/lib/rules/selector-type-case/__tests__/index.js
+++ b/lib/rules/selector-type-case/__tests__/index.js
@@ -241,6 +241,9 @@ testRule(rule, {
     },
     {
       code: ".foo, %foo {}"
+    },
+    {
+      code: "$dummyValue {}"
     }
   ]
 });
--- a/lib/utils/__tests__/isStandardSyntaxTypeSelector.test.js
+++ b/lib/utils/__tests__/isStandardSyntaxTypeSelector.test.js
@@ -90,8 +90,23 @@ describe("isStandardSyntaxTypeSelector", () => {
       expect(isStandardSyntaxTypeSelector(func)).toBeFalsy();
     });
   });
-  it("variable selector", () => {
+  it("scss variable as selector", () => {
     return rules("$variable {}", func => {
+      expect(isStandardSyntaxTypeSelector(func)).toBeFalsy();
+    });
+  });
+  it("less variable as selector", () => {
+    return rules("@variable {}", func => {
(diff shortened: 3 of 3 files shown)


YozhikM:
I don't know why tests not passed
```bash
Now using node v6.14.3 (npm v3.10.10)
iMac--USER:stylelint yozhikm$ yarn test system-tests/fix/fix.test.js
yarn run v1.7.0
$ npm-run-all --serial lint flow
$ npm-run-all --parallel lint:*
$ eslint . --cache
$ remark . --quiet --frail
$ flow
No errors!
$ jest --coverage system-tests/fix/fix.test.js
 PASS  system-tests/fix/fix.test.js
  fix
    ✓ expected stylelint results (162ms)
    ✓ expected fixes to PostCSS Result (34ms)
    ✓ overwrites the original file (38ms)
    ✓ doesn't write to ignored file (8ms)
    ✓ outputs fixed code when input is code string (23ms)
```

hudochenkov:
@YozhikM thank you for adding a test. Can you add a test for every new condition you've added to that utility, please? Currently, it covers only one case.

lib/utils/isStandardSyntaxTypeSelector.js
```diff
@@ -36,6 +36,23 @@ module.exports = function(node /*: Object*/) /*: boolean*/ {
...
   }
 
+  let normalizedValue = node.value;
+
+  // Ignore operators before variables (example -$variable)
+  if (/^[-+*/]/.test(node.value[0])) {
+    normalizedValue = normalizedValue.slice(1);
+  }
+
+  // SCSS variable
```
jeddy3: I think we need to look into this issue more closely.

I think we need to:

- establish how we test against a non-standard custom syntax i.e. `postcss-jsx` in this instance
- confirm exactly what syntax is causing the original bug

The first one could involve a change to `testRule` so that the `syntax` property accepts a parser function as well as a string. We'd then include `postcss-jsx` as a dev dependency.

The latter one needs an investigation into what non-standard syntax can be used in a selector. For example, I'm pretty sure that within Scss and Less you can only use variables within interpolation, which makes those conditionals in the `isStandard*` util redundant.

Additionally, it might be worth waiting until https://github.com/stylelint/stylelint/pull/3284/files is merged before revisiting this issue as version 4 of the selector parser is less likley to categorize these things as `tag` selectors.

YozhikM: In which cases is the percent sign used before the selector?

YozhikM: @jeddy3 I agree with your opinion. Shall we put it off for the present?

hudochenkov:
Assigned label “blocked”, as we agreed to wait for https://github.com/stylelint/stylelint/pull/3284 to be merged first.

brandonkal:
The other issue has been merged so it appears this can now be unblocked.

evilebottnawi:
Yes, this PR can be continue

hudochenkov:
Blocked by https://github.com/stylelint/stylelint/pull/3988

hudochenkov:
@YozhikM looks like PR we though is blocking this PR is not going to be merged soon. Could you resolve merge conflicts, so we can review again, please?

Mouvedia:
> looks like PR we though is blocking this PR is not going to be merged soon.

#3988 was superseded by #4595, which was merged.

jeddy3:
Resolved by addition of [`ignoreTypes` secondary option](https://github.com/stylelint/stylelint/issues/3391#issuecomment-599440184).

PR #3400, opened 2018-06-19
Remove aei references

jeddy3:

Ref: https://github.com/stylelint/stylelint/pull/2117#issuecomment-398280562

The `allowEmptyInput` option [was removed](https://github.com/stylelint/stylelint/pull/2464) in `8.0.0`, but the option still lingers in the CLI `--help` and type defs. I suspect this happened during some conflict resolution work.


Mouvedia:
note

It was added back in #3965.

Issue #3264, opened 2018-04-17
Add CSS-in-JS support


ai:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

feature request

@gucong3000 announced amazing [`postcss-styled`](https://github.com/gucong3000/postcss-styled) parser for all `Button = styled()` CSS-in-JS solutions (styled-components, Emotion.js, etc). Let’s add it to default pack as other parsers (or to the docs).

CSS-in-JS become more and more popular. To bring new user and keep “trendy” feeling (it is very important to find new contributors) we should have official CSS-in-JS support.

Right now we can do it by `stylelint-processor-styled-components`. But, as I understand, processor doesn’t work with `--fix`. Also, it doesn’t work well with all rules.

`postcss-styled` could be used easy as any other PostCSS syntaxes like `postcss-scss`.

@jeddy3 what do you think?

ai:
@jeddy3 any thoughts or questions?

ntwb:
Thanks for creating the issue @ai 👍 

This is part of a refactoring of processors that stylelint uses with the goal of reducing the overall optimizing the package size, see:

> https://github.com/stylelint/stylelint/pull/3261#issuecomment-381768182
> https://github.com/stylelint/stylelint/issues/3248#issuecomment-381385995

We will for sure work toward adding `postcss-styled` as part of all this 😄

jeddy3:
There were two amazing things I ran into while reviewing https://github.com/stylelint/stylelint/pull/3261:

1. CSS being autofixed within markdown fences, `<style>` tags _and `style=""` attributes_! This is due to @gucong3000's hard work on the new PostCSS syntaxes.
2. The advanced features of processors! The fact that stylelint can, through the hard work of processor authors, lint the following blows my mind:

```javascript
import styled from 'styled-component';
const Button = styled.div`
  ${/* sc-selector */ Button} {
    ${/* sc-prop */ something}: papayawhip;
    margin-${/* sc-custom 'left' */ rtlSwitch}: 12.5px;
  }
`
```

Where `/* sc-selector */` etc are [interpolation tags](https://www.styled-components.com/docs/tooling#interpolation-tagging) included within the styled-components processor.

Or:

```javascript
import glamorous from 'glamorous';
const Component = glamorous.div({ margin: 0 });
render(
  <Div css={{ margin: 0 }}/>
)
export const styles = 
// @css
{
  margin: 0
}
```

[...]

PR #3407, opened 2018-06-21
PoC to install on demand postcss-markdown dependency

gucong3000:

- #3248
- #3264

Drop built-in support for markdown fences.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
     "npm-run-all": "^4.0.2",
     "npmpub": "^4.0.1",
     "postcss-import": "^11.0.0",
+    "postcss-markdown": "^0.31.0",
     "prettier": "~1.13.0",
     "remark-cli": "^5.0.0",
     "remark-lint-no-missing-blank-lines": "^1.0.1",
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -10,6 +10,7 @@ const meow = require("meow");
 const needlessDisablesStringFormatter = require("./formatters/needlessDisablesStringFormatter");
 const path = require("path");
 const resolveFrom = require("resolve-from");
+const spawn = require("child_process").spawn;
 const standalone = require("./standalone");
 
 /*:: type meowOptionsType = {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -101,6 +100,7 @@
     "npm-run-all": "^4.0.2",
     "npmpub": "^3.1.0",
     "postcss-import": "^11.0.0",
+    "postcss-markdown": "^0.30.0",
```
ntwb: @gucong3000 Should `postcss-markdown` have moved from `dependencies` to `devDependencies`?

jeddy3:
This is a breaking change.

Let's make sure [we're all happy with the direction we're taking](https://github.com/stylelint/stylelint/issues/3264#issuecomment-399043917) before we merge this.

For example, if we want to go down the route of supporting non-standard things, then we should probably investigate how we can roll-up postcss-markdown into a separate bundle to reduce weight and help with performance.

evilebottnawi:
@jeddy3 I'm against removal `postcss-markdown`, it is help linting a lot of documentation, also i don't see any problems in `postcss-markdown`

ntwb:
@evilebottnawi It was originally proposed to be removed to improve performance. I'd also and I will continue to use it even if I have to install `postcss-markdown` separately lint documentation.

As such per the original decision to remove it for performance reasons is why I approved this PR. Now as noted over in https://github.com/stylelint/stylelint/issues/3264#issuecomment-399043917 we are still considering performance, but we're also considering adding and refactoring stylelint syntaxes supported, if we are going to add those currently suggested then I'm in agreement with you that markdown should also be included.

evilebottnawi:
@ntwb Somebody have benchmark? How we decrease perf using `postcss-markdown` out of box?

ntwb:
@evilebottnawi, apologies, not _performance_ it was to reduce the size of the stylelint package #3248

evilebottnawi:
@ntwb i think `postcss-markdown` it not large :smile:

gucong3000:
I pushed a new commit to implement such a feature:
```bash
$ stylelint "**/*.md"
PostCSS syntax module missing: "postcss-markdown"
Please use "--fix" flag for automatic install or install it by yourself

$ stylelint "**/*.md" --fix
Automatic installation missing module: postcss-markdown
+ postcss-markdown@0.30.0
added 1 package from 1 contributor and audited 25185 packages in 32.202s

lib/__tests__/fixtures/style-tag.markdown
 6:1  ×  Unexpected empty block   block-no-empty
```

jeddy3:
> I pushed a new commit to implement such a feature:

Thanks for this PoC. It looks interesting. I think we should hold off on the idea for now and maybe roll it out at a later date as, I believe, this is a breaking change.

We can add CSS-in-JS support in as a non-breaking change. So, let's do that.

PR #3420, opened 2018-06-23
Fix false positives for no-descending-specificity and scss syntax

### LINKED ISSUE #3417: Fix false positives for #{&} in no-descending-specificity
> Clearly describe the bug

There are false positives for rule `no-descending-specificity` when some manipulations are applied to a selector.

`no-descending-specificity`

e.g.

```css
.foo {
    &--a,
    &--a#{&}--b {
      div {
        padding: 10px;
      }
    }
}
```

default

`9.2.0`

"CLI with `stylelint '**/*.{css,scss}'`"

"Yes, it's related to SCSS injections into selector string."

"No warnings to be flagged."

"The following warnings were flagged:"

```shell
file.scss
4:6    Expected selector ".foo--a#{.foo}--b div" to come before selector ".foo--a div" (no-descending-specificity)
```

YozhikM:

Closes #3417 

No.

### DIFF
--- a/lib/rules/no-descending-specificity/__tests__/index.js
+++ b/lib/rules/no-descending-specificity/__tests__/index.js
@@ -155,7 +155,7 @@ testRule(rule, {
 
   accept: [
     {
-      code: ".foo { &--a, &--a#{&}--b {  div { padding: 10px; } } }"
+      code: ".foo { &--a, &--a#{&}--b {  div {} } }"
     }
   ]
 });
--- a/lib/rules/no-descending-specificity/__tests__/index.js
+++ b/lib/rules/no-descending-specificity/__tests__/index.js
@@ -147,3 +147,15 @@ testRule(rule, {
     }
   ]
 });
+
+testRule(rule, {
+  ruleName,
+  syntax: "scss",
+  config: [true],
+
+  accept: [
+    {
+      code: ".foo { &--a, &--a#{&}--b {  div { padding: 10px; } } }"
+    }
+  ]
+});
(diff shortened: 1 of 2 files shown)


lib/rules/no-descending-specificity/__tests__/index.js
```diff
@@ -147,3 +147,15 @@ testRule(rule, {
...
 });
+
+testRule(rule, {
+  ruleName,
+  syntax: "scss",
+  config: [true],
+
+  accept: [
+    {
+      code: ".foo { &--a, &--a#{&}--b {  div { padding: 10px; } } }"
```
jeddy3: As [per the conventions](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#example-patterns):

> Use the minimum amount of code possible to communicate the pattern e.g. if the rule targets selectors then use an empty rule e.g. {}.

Can we use the following please?:

```js
.foo { &--a, &--a#{&}--b { div {} } }

File: lib/rules/no-descending-specificity/index.js [hunk 2]
```diff
@@ -55,9 +56,11 @@ const rule = function(actual) {
         const index = rule.selector.indexOf(trimSelector);
         // Resolve any nested selectors before checking
         resolvedNestedSelector(selector, rule).forEach(resolvedSelector => {
-          parseSelector(resolvedSelector, result, rule, s =>
-            checkSelector(s, rule, index, comparisonContext)
-          );
+          parseSelector(resolvedSelector, result, rule, s => {
+            if (isStandardSyntaxSelector(resolvedSelector)) {
```
jeddy3: The convention, in this file and others, is to return early when using the `isStandard*` utils. We should probably follow it here too:

```js
parseSelector(resolvedSelector, result, rule, s => {
  if (isStandardSyntaxSelector(resolvedSelector)) {
    return;
  }
  checkSelector(s, rule, index, comparisonContext);
}
```

hudochenkov:
> -   Fixed: `no-descending-specificity` false positives for #{&} ([#3420](https://github.com/stylelint/stylelint/pull/3420)).

Issue #3436, opened 2018-07-03
`keyframes-name-pattern` does not support raw JS RegExp


daltones:
> Clearly describe the bug

Although the docs says that `keyframes-name-pattern` supports options as `regex|string` it only supports `string` as of stylelint v9.3.0.

> Which rule, if any, is the bug related to?

`keyframes-name-pattern`

> What CSS is needed to reproduce the bug?

any

> What stylelint configuration is needed to reproduce the bug?

`.stylelintrc.js`:
```js
module.exports = {
  rules: {
    'keyframes-name-pattern': /^some-regexp$/
  }
};
```

> Which version of stylelint are you using?

`v9.3.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylelint "**/*.css"` and a `.stylelintrc.js`

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

No

> What did you expect to happen?

No errors.

> What actually happened (e.g. what warnings or errors did you get)?

Error `Invalid Option: Invalid option value "/^some-regexp$/" for rule "keyframes-name-pattern`

jeddy3:
@daltones Thanks for the report, for using the template and for the quick PR.

(Ref: https://github.com/stylelint/stylelint/issues/3008 - there are a few other rules that need tweaking, but they can be covered in that PR)

hudochenkov:
Thank you for the report and for the fix!

PR #3437, opened 2018-07-03
Fix keyframes-name-pattern not supporting RegExp notation

daltones:

Closes https://github.com/stylelint/stylelint/issues/3436

No, it's self explanatory.


hudochenkov:
> -   Fixed: `keyframes-name-pattern` support for raw JS RegExp ([#3437](https://github.com/stylelint/stylelint/pull/3437)).

Issue #3362, opened 2018-06-05
Better promote the plugin system


jeddy3:
Ref: https://github.com/stylelint/stylelint/pull/3355#issuecomment-394426248 & https://github.com/stylelint/stylelint/issues/3361#issuecomment-394548640

There are some excellent stylelint plugins. We could do a better job of promoting them and the plugin system itself.

Currently, we:

* give plugins a top-level navigation item on the website
* mention the plugin system within the list of features in the README
* mention and link to the list of plugins from the list of the rules page

A couple of ideas... we could:

1. Move "Support for plugins" up the features list and embellish it with examples e.g. **Support for plugins**: It's easy to create your own rules and add them to the linter, or make use of plugins written by the community. For example, there are plugin packs for [Scss](..), [a11y](..) and [ordering of things]() you can use.
2. Create a "Highlighted plugins" list on the plugins page. We could flesh out the descriptions of each of these plugins.
3. Look to embellish the existing mentions of the plugins system in the docs.
3. Look for other places in the docs where a mention of the plugin system would be appropriate.

Any others?

filipekiss:
I guess we could scrape the project issues to see what people ask the most and that could be solved by plugins and create a few "recipes" to tackle those common issues, showing how powerful stylelint can get exactly because of plugins.

PR #3462, opened 2018-07-13
Add stylelint-prettier plugin

BPScott:

No

### DIFF
--- a/docs/user-guide/plugins.md
+++ b/docs/user-guide/plugins.md
@@ -15,7 +15,7 @@ Plugins are rules and sets of rules built by the community that support methodol
 -   [`stylelint-no-indistinguishable-colors`](https://github.com/ierhyna/stylelint-no-indistinguishable-colors): Disallow colors that are suspiciously close to being identical.
 -   [`stylelint-no-unsupported-browser-features`](https://github.com/ismay/stylelint-no-unsupported-browser-features): Disallow features that are unsupported by the browsers that you are targeting.
 -   [`stylelint-order`](https://github.com/hudochenkov/stylelint-order): Specify the ordering of things e.g. properties within declaration blocks (plugin pack).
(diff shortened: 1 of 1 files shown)


BPScott:
Hi @ntwb,

`prettier-stylelint` is a wrapper that runs prettier then stylelint. This means instead of running `stylelint`, you run a new executable `prettier-stylelint`. This is useful if you want stylelint to make adjustments to prettier's output, but it's a new command for people to remember. This is good for autofixing but makes identifying if changes need to be made without making corrections a little tricky. 

`stylelint-prettier` is a plugin to stylelint that exposes prettier modifications as a stylelint rule. This means you continue to run `stylelint` as you would before and prettier changes are exposed as errors in stylelint's reporting. This means people don't need to change their existing process - running `stylelint` shall report on changes and `stylelint --fix` shall fix them.

docs/user-guide/plugins.md
```diff
@@ -15,6 +15,7 @@ Plugins are rules and sets of rules built by the community that support methodol
 -   [`stylelint-no-indistinguishable-colors`](https://github.com/ierhyna/stylelint-no-indistinguishable-colors): Disallow colors that are suspiciously close to being identical.
 -   [`stylelint-no-unsupported-browser-features`](https://github.com/ismay/stylelint-no-unsupported-browser-features): Disallow features that are unsupported by the browsers that you are targeting.
 -   [`stylelint-order`](https://github.com/hudochenkov/stylelint-order): Specify the ordering of things e.g. properties within declaration blocks (plugin pack).
+-   [`stylelint-prettier`](https://github.com/BPScott/stylelint-prettier): Runs [Prettier](https://prettier.io/) as a stylelint rule
```
hudochenkov: Could you put a dot at the end, please?

ntwb:
Thanks for the detailed description @BPScott 💯

jeddy3:
I think this could be one of the plugins we consider highlighting when we come to do https://github.com/stylelint/stylelint/issues/3362.

@BPScott Thanks for the good work and the plugin!

PR #3463, opened 2018-07-14
Fix false positives for system-ui font-family-name-quotes

### LINKED ISSUE #3454: Fix false positives for system-ui font-family-name-quotes
> Clearly describe the bug

`system-ui` is a [standard](https://caniuse.com/#feat=font-family-system-ui) font keyword that should be exempted from the `font-family-name-quotes` rule.

`font-family-name-quotes`

```css
body {
font-family: system-ui, -apple-system, 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', sans-serif; }
```

```json
{
	"rules": {
		"font-family-name-quotes": "always-unless-keyword",
	}
}
```

`9.3.0`

`$ stylelint 'styles/build/*.css'`

No. See: https://drafts.csswg.org/css-fonts-4/#system-ui-def

`system-ui` should be treated as other system font keywords as referenced in #1295.

```shell
Expected quotes around "system-ui"   font-family-name-quotes
```

ScottSmith95:

Closes #3454

Nope.


ntwb:
Changelog added in: https://github.com/stylelint/stylelint/commit/538124059d5d0be9b7f06ad392021469aacd3f4f

> Fixed: `font-family-name-quotes` treats the `system-ui` system font as a keyword that should never be wrapped in quotes.

ntwb:
Welcome to the @stylelint organisation @ScottSmith95 🎉

ScottSmith95:
Thank you!

PR #3405, opened 2018-06-20
Add bundled support for styles in template literals

gucong3000:
> Which issue, if any, is this issue related to?

#3264

> Is there anything in the PR that needs further explanation?

e.g. "No, it's self explanatory."

Issue #3128, opened 2018-01-15
Add overrides property to configuration object


Grawl:
I want to set different rules to lint `.css` and `.sass` files. It's different syntaxes.

These rules are not compatible with `.sass` indented syntax:

- [braces `{}`](https://stylelint.io/user-guide/rules/#block-1)
- [semicolon `;`](https://stylelint.io/user-guide/rules/#declaration-block-2)

So I want to disable them for `.sass` files, but control them for `.css`

I created a sandbox repository to test out [`postcss-sass`](https://github.com/AleshaOleg/postcss-sass) parser with Stylelint, so you can uncomment any incompatible rules mentioned above to see it's false-positives

https://github.com/Grawl/stylelint-sass-test

`"stylelint": "^8.2.0"`

My propose for configuration:

```json
{
    "rules": {
        "optionOne": true, 
        "optionTwo": false
    },
    "overrides": {
        "syntax": "sass",
        "rules": {
            "optionOne": false,
            "optionTwo": true
        }
    }
}
```

Prettier have this option: https://prettier.io/docs/en/configuration.html#configuration-overrides

#2486 #2593

ai:
I need this feature too. I have different roles for CSS and CSS-in-JS files.

I like `overrides` from ESLint. We could also take it to define different roles for different file patterns.

Grawl:
> We could also take it to define different roles for different file patterns.

Can we use different config files in different folders?

ai:
@Grawl not in my case. I have many folders at root. As result, right now I must copy same config to 3-5 places. And it force mistakes — change config in one folder and forget about other folders.

Grawl:
Appreciate. But we don't have `"files"` option in Stylelint as I know.

hudochenkov:
Overrides is a good idea.

As a temporary solution you can use config [`extends`](https://stylelint.io/user-guide/configuration/#extends). Create a second config for .sass files, extend it from the main config and disable unwanted rules:

`.stylelintrc.sass.js`:

```js
module.exports = {
    extends: './.stylelintrc.js',
    rules: {
        'block-closing-brace-empty-line-before': null,
    },
};
```

It will require two npm tasks to run it, though:

```bash
stylelint "**/*.css"
stylelint "**/*.sass" --config .stylelintrc.sass.js
```

Or it can be combined into one:

```bash
stylelint "**/*.css" && stylelint "**/*.sass" --config .stylelintrc.sass.js
```

[...]

PR #3485, opened 2018-07-20
Disable `no-missing-end-of-source-newline` for inline styles

gucong3000:

https://github.com/gucong3000/postcss-html/issues/70

No, it's self explanatory.


jeddy3:
Interesting. 

This makes the rule aware of the context in which it's used in, right? i.e. `root.source.inline` is a property introduced in postcss-html?

Up until now, I think rules have been unware of their context. It has been the user's responsbility to decide which rules are appropriate to turn on.

A `.vue` files can contain two types of styles in one file e.g.

```html
<style>
  .component { color: red; }
</style>
<script>
  new Vue({
    el: '#demo',
    render: function (h) {
      return (
        <span style="max-width: 10px">{something_else}</span>
      )
    }
  })
</script>
```

Wouldn't a user want to turn off this rule for both these instances?

And isn't that better done with [an `overrides` property](https://github.com/stylelint/stylelint/issues/3128) than this approach?

gucong3000:
> `root.source.inline` is a property introduced in postcss-html?

Yes. there's something else:
```
const document.syntax.parse('............')
document.source.lang  // lang of source, html, jsx, markdown, css, sass, scss, less ...
document.nodes[0].source.lang //lang of first style block, only work for html-link，js，markdown
```

gucong3000:
Wouldn't a user want to turn off this rule for both these instances?

```html
<style>
  .component { color: red; }
  // a newline should be here: ==>>
</style>
```
```html
<style>
  .component { color: red; }
</style>
<script>
  new Vue({
    el: '#demo',
    render: function (h) {
      return (
        // I think stylelint not support next line:
        <span style="max-width: 10px">{something_else}</span>
      )
    }
  })
</script>
```

gucong3000:
> And isn't that better done with an overrides property than this approach?

This is true in other rules. But for this rule, we have a clear disabling condition.

jeddy3:
> But for this rule, we have a clear disabling condition
> Users does not need to choose config.

Ok. I agree it's the better developer experience.

Can you please revert the last commit (adding in JSX)? Then we can merge this.

We should add your last commit in separate PR, like https://github.com/stylelint/stylelint/pull/3405.

gucong3000:
> Can you please revert the last commit

Done.

PR #3486, opened 2018-07-20
Use "ID" and not "id" throughout

### LINKED ISSUE #3476: Use spec and common spelling of “ID” (not “id”)
When referring to unique identifiers, the stylelint docs speak of “ids”, not “IDs”. While this seems like a minor thing, both related specs (like [CSS](https://www.w3.org/TR/CSS22/selector.html#id-selectors)) and the common spelling of the term (cf. [Wikipedia’s entry on identifiers](https://en.wikipedia.org/wiki/Identifier)) suggest the spelling “ID”—strongly so, in my view.

I’d be happy to update the docs if it sounds reasonable to use the same spelling as both specs and common writing.

j9t:

Closes #3476

No, it's self-explanatory.


lib/rules/selector-max-id/index.js
```diff
@@ -12,7 +12,7 @@ const ruleName = "selector-max-id";
 
 const messages = ruleMessages(ruleName, {
   expected: (selector, max) =>
-    `Expected "${selector}" to have no more than ${max} id ${
+    `Expected "${selector}" to have no more than ${max} IDs ${
```
hudochenkov: `ID`.

j9t: I intentionally switched to plural since that seemed to work better for any `${max}` > 1. Is `${max}` 1, most frequently, or do I misunderstand so that this better stays singular?

hudochenkov: Let's ask @jeddy3 or @ntwb, they know English better than I :)

Arcanemagus: > Expected "${selector}" to have no more than _n_ IDs

With _n_ > 1 this is correct.

jeddy3: > do I misunderstand so that this better stays singular?

The message, [like with other `selector-max-*` rules](https://github.com/stylelint/stylelint/blob/2d380692173fb9c67d566d440c6cba0fc33d3b4d/lib/rules/selector-max-attribute/index.js#L15-L20), is ended (via a ternary operator inside the template literal) with the word "selectors" if there is zero or two or more, or "selector" if there is one. This gives:

- Expected "${selector}" to have no more than 0 ID selectors
- Expected "${selector}" to have no more than 1 ID selector
- Expected "${selector}" to have no more than 2 ID selectors

As you can see, pluralising ID wouldn't make sense as it'll give:

- Expected "${selector}" to have no more than 0 IDs selectors
- Expected "${selector}" to have no more than 1 IDs selector
- Expected "${selector}" to have no more than 2 IDs selectors

j9t:
Sorry for skipping further discussion but I just reverted to singular. Should all be good to go.

jeddy3:
@j9t You have a wealth of experience with CSS, and especially with coding guidelines! I've been looking through your posts and little books for ideas of new rules. I believe stylelint can already enforce a lot of your guidance, but there were two areas that stood out to me: 

- organising selectors
- enforcing development principles

There are a few built-in rules for organizing selectors, e.g. [`no-descending-specificity`](https://github.com/stylelint/stylelint/tree/master/lib/rules/no-descending-specificity) and [`no-duplicate-selectors`](https://github.com/stylelint/stylelint/tree/master/lib/rules/no-duplicate-selectors), but I'm going to think about other options for ordering selectors.

I found the enforcement of development principles to be particularly interesting. If I have understood your meaning correctly, then I believe the built-in `*-whitelist` and `*-blacklist` rules are applicable here. For example, we can enforce the principle of:

- mobile-first with [`"media-feature-name-whitelist": ["min-width"]`](https://github.com/stylelint/stylelint/tree/master/lib/rules/media-feature-name-whitelist)
- proportional units with [`"unit-whitelist": ["%", "em", "rem"]`](https://github.com/stylelint/stylelint/tree/master/lib/rules/unit-whitelist)

As with organising selectors, I'm going to think more about what else stylelint can do in this area.

If you have any ideas for new rules, please do create issues as any help making stylelint more useful would be appreciated! 

Thanks again for the PR.

ntwb:
And welcome to the @stylelint organisation @j9t 🎉

j9t:
Thank you for the kind welcome, @jeddy3, @ntwb!

I’ll be more than happy to contribute more. For your ideas there, @jeddy3, I’ll take some time to think more about this. Selector order could be standardized in several ways—personally, I’ve long been working with the approach outlined in https://meiert.com/en/blog/how-to-order-css-selectors/, and on the far end even alphabetical selector sorting may be a (though contentious and not trivial!) option—, so I find that quite interesting.

A side note until I hope to report back again, I’ve turned a bit more lenient when it comes to guidelines (and their enforcement). Just thought experiments so far, how expensive is unordered code really (in “common” code bases, not the obviously broken ones), and how far could one push auto-fixing everything (before production compression and such) so that developers can write any style they want (sure, to then see it adjusted to the organization’s guidelines—cheers stylelint `--fix`)? Well. :)

PR #3508, opened 2018-08-01
Fix false positives for selector-descendant-combinator-no-non-space

### LINKED ISSUE #3449: Fix false positives for calculations with parenthesis in selector-descendant-combinator-no-non-space
> Clearly describe the bug

I'm getting what I think is a false positive with a line that contains lots of calculations with variables, but doesn't have any descending in it. It's probably related to the variable being in parenthesis, because if you remove the parenthesis then the error is not raised.

`selector-descendant-combinator-no-non-space`

```css
div {
    & > :nth-child(#{$i + ($column - 1)}) {
        grid-column: $column;
    }
}
```

```json
{
}

```

`9.2.1`

`npm run stylelint` and package.json contains `cd static/styles-src && stylelint '**/*.scss'`

Yes, it's related to SCSS calculations.

No warnings to be flagged.

```shell
> cd static/styles-src && stylelint '**/*.scss'


xx/xx.scss
 13:86  ✖  Unexpected "+($column - 1)"   selector-descendant-combinator-no-non-space
```

CAYdenberg:

fixes #3449 

This rule is a little bit strange in that a descendent combinator is always whitespace, it's just not always a _single_ space. In fact the logic of the rule checks that the combinator is not one of the standard "non-space" combinators (+, >) and _passes_ if that is the case.

For this particular bug, PostCSS apparently determined that `+($column - 1)` is a combinator, which is odd, and it is not a non-space combinator nor is it a single space.

Anyway, since what are _really_ looking is whitespace combinators that are not single spaces, I changed the logic to do that instead. It's a bit of a work-around but it covers the case in question and most importantly does not regress.

### DIFF
--- a/lib/rules/selector-descendant-combinator-no-non-space/__tests__/index.js
+++ b/lib/rules/selector-descendant-combinator-no-non-space/__tests__/index.js
@@ -52,6 +52,9 @@ testRule(rule, {
     {
       code: "html { --custom-property-set: {} }",
       description: "custom property set in selector"
+    },
+    {
+      code: "div > :nth-child(2n + 1) {}"
     }
   ],
 
@@ -82,6 +85,18 @@ testRule(rule, {
     }
   ]
 });
+
+testRule(rule, {
+  ruleName,
+  config: [true],
+  syntax: "scss",
+
+  accept: [
+    {
+      code: "div > :nth-child(#{$i + ($column - 1)}) {}"
+    }
+  ]
+})
 
 testRule(rule, {
   ruleName,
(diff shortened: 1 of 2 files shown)


jeddy3:
> PostCSS apparently determined that +($column - 1) is a combinator, which is odd

This might change in https://github.com/stylelint/stylelint/pull/3284.

> Anyway, since what are really looking is whitespace combinators that are not single spaces, I changed the logic to do that instead.

LGTM. Even if the behaviour of the parser changes, I think the logic in this PR is clear.

jeddy3:
> -   Fixed: `selector-descendant-combinator-no-non-space` false positives for calculations with parenthesis ([#3508](https://github.com/stylelint/stylelint/pull/3508)).

PR #3524, opened 2018-08-08
Add `unix` formatter

### LINKED ISSUE #3522: Add unix-style formatter

For some unix tools (e.g. Emacs), I think that unix-style formatter is very useful, which is similar to [ESLint's unix formatter](https://eslint.org/docs/user-guide/formatters/#unix).

See also:
- https://github.com/eslint/eslint/issues/2991
- https://github.com/eslint/eslint/blob/137140f737d720fa93a4ac06ce79a86b523327f7/lib/formatters/unix.js#L1

e.g. 

```
$ stylelint --formatter=unix foo.css
/Users/ybiquitous/foo.css:1:26: Unexpected empty block (block-no-empty) [error]

1 problem
```

```
$ stylelint --formatter=unix bar.css
/Users/ybiquitous/bar.css:1:26: Unexpected empty block (block-no-empty) [error]
/Users/ybiquitous/bar.css:10:26: Unexpected empty block (block-no-empty) [error]

2 problems
```

ybiquitous:

Fixes #3522

No, it's self explanatory.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -1,9 +1,11 @@
 /* @flow */
 "use strict";
 
+const _ = require("lodash");
 const chalk = require("chalk");
 const dynamicRequire = require("./dynamicRequire");
 const EOL = require("os").EOL;
+const formatters = require("./formatters");
 const getModulePath = require("./utils/getModulePath");
 const getStdin = require("get-stdin");
 const meow = require("meow");
@@ -11,6 +13,13 @@ const needlessDisablesStringFormatter = require("./formatters/needlessDisablesSt
 const path = require("path");
 const resolveFrom = require("resolve-from");
 const standalone = require("./standalone");
+
+const formatterOptions = _.chain(formatters)
+  .keys()
+  .map(opt => `"${opt}"`)
+  .join(", ")
+  .value()
+  .replace(/, ([a-z"]+)$/u, " or $1");
 
 /*:: type meowOptionsType = {
   argv?: string[],
@@ -234,7 +243,7 @@ const meowOptions /*: meowOptionsType*/ = {
 
       --formatter, -f               [default: "string"]
 
-        The output formatter: "compact", "json", "string" or "verbose".
+        The output formatter: ${formatterOptions}.
 
       --custom-formatter
 
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -1,11 +1,10 @@
 /* @flow */
 "use strict";
 
-const _ = require("lodash");
 const chalk = require("chalk");
 const dynamicRequire = require("./dynamicRequire");
 const EOL = require("os").EOL;
-const formatters = require("./formatters");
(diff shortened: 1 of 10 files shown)


lib/cli.js
```diff
@@ -234,7 +243,7 @@ const meowOptions /*: meowOptionsType*/ = {
 
       --formatter, -f               [default: "string"]
 
-        The output formatter: "compact", "json", "string" or "verbose".
+        The output formatter: ${formatterOptions}.
```
ybiquitous: To avoid a bug such as #3521, I change to generate a formatter options text from `formatters` object.
But there are no tests for help text, so I don't add test cases for this change.

Here is a partial help text related to this change, which is output in my CLI termial:

```
$ ./bin/stylelint.js -h | grep -C3 '\-\-formatter' 
      If the directory for the cache does not exist, make sure you add a trailing "/"
      on *nix systems or "\" on Windows. Otherwise the path will be assumed to be a file.

    --formatter, -f               [default: "string"]

      The output formatter: "compact", "json", "string", "unix" or "verbose".
```

This change may be over-engineering, so I want to hear opinions of reviewes about this change.
(I will go back to hard-coded text, if there are more oppositions)

ybiquitous:
Thanks for a feedback! I will change also `docs/user-guide/node-api.md`

https://github.com/stylelint/stylelint/blob/700b7fb7419678e0c903172e67054405c1a79ccf/docs/user-guide/node-api.md

lib/cli.js
```diff
@@ -12,6 +14,13 @@ const path = require("path");
 const resolveFrom = require("resolve-from");
 const standalone = require("./standalone");
 
+const formatterOptions = _.chain(formatters)
+  .keys()
+  .map(opt => `"${opt}"`)
+  .join(", ")
+  .value()
+  .replace(/, ([a-z"]+)$/u, " or $1");
+
```
ntwb: Rather than defining `formatterOptions` multiple times (above and in `lib/standalone.js`), perhaps making this a util in `lib/utils` with some tests.

The above is not a blocker, but would be nice to have some tests for this.

ybiquitous: Yes, I think it's a nice idea! I will try to change code and add tests. 💪

ybiquitous:
@ntwb Thanks for your reviewing! I fixed your review points, so please review again!

jeddy3:
@ntwb Have your concerns been addressed?

ntwb:
@jeddy3 Yes, thanks a bunch for this @ybiquitous 😄

ntwb:
Changelog update in https://github.com/stylelint/stylelint/commit/e327558152d7b9ea9f48a01fa6f2c77aeab16238
> Added: unix formatter (#3524).

ybiquitous:
Thank you, too. 😊

ntwb:
p.s. Welcome to the @stylelint organisation @ybiquitous 🎉

PR #3530, opened 2018-08-12
Fix incorrect line in max-empty-lines

GorwinJororis:

#2343

No, it's self explanatory.

### DIFF
--- a/lib/rules/max-empty-lines/README.md
+++ b/lib/rules/max-empty-lines/README.md
@@ -1,3 +1,4 @@
+
 # max-empty-lines
 
 Limit the number of adjacent empty lines.
@@ -30,12 +31,12 @@ b {}
 Comment strings are also checked -- so the following is a violation:
 
 ```css
-/**
- * Call me Ishmael.
- *
- *
- *
- * Some years ago -- never mind how log precisely -- ...
+/*
+ Call me Ishmael.
+
+
+
+ Some years ago -- never mind how log precisely -- ...
  */
 ```
 
@@ -67,7 +68,7 @@ Only enforce the adjacent empty lines limit for lines that are not comments.
 
 For example, with `2` adjacent empty lines:
 
-The following patterns are considered violations:
+The following pattern is considered violation:
 
 ```css
 /* horse */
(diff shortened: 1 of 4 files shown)


lib/rules/max-empty-lines/README.md
```diff
@@ -30,12 +31,12 @@ b {}
...
- *
- *
- * Some years ago -- never mind how log precisely -- ...
+/*
+ Call me Ishmael.
+
+
+
+ Some years ago -- never mind how log precisely -- ...
  */
```
ntwb: Should both forms of comments be displayed here?

i.e:

```
/**
 * Call me Ishmael.
 *
 *
 *
 * Some years ago -- never mind how log precisely -- ...
/*
```

And
```
/*
 Call me Ishmael.



 Some years ago -- never mind how log precisely -- ...
 */
```

lib/rules/max-empty-lines/index.js
```diff
@@ -38,60 +39,65 @@ const rule = function(max, options) {
...
-    // warning position
-    if (!ignoreComments) {
-      root.walkComments(comment => {
-        const source =
-          (comment.raws.left || "") + comment.text + (comment.raws.right || "");
-        styleSearch({ source, target: "\n" }, match => {
-          checkMatch(source, match.endIndex, comment, 2);
-        });
-      });
-    }
```
ntwb: Is the above section supposed to have been removed?

Won't the warning position for SCSS be inaccurate with this removed?

lib/rules/max-empty-lines/README.md
GorwinJororis: From my point of view, this comment doesn't have empty lines (please attention to stars):

```
/**
 * Call me Ishmael.
 *
 *
 *
 * Some years ago -- never mind how log precisely -- ...
 */
```

That's why I removed from readme.md.

hudochenkov: I agree with @GorwinJororis. This example doesn't have empty lines.

lib/rules/max-empty-lines/index.js
GorwinJororis: @ntwb Well you found something right that there will be false warning position in non-standard syntaxes but it's not related to what you pointed above. Because above is just a part which looks for comments in input css string based on `ignore: ['comments']` option which I already did after few lines later. 

Beside that problem is `root.toString() `. Because it cannot be used to check css string. Instead of you have to use PostCSS' node functions to achieve this.

Then there is one more problem. It's same problem why I couldn't in my first PR and also now in my second PR because PostCSS' node tree structure is so crap.  

For example, in node tree, `before` is the space symbols before the node. It also stores * and _ symbols before the declaration (IE hack), `after` is the space symbols after the last child of the node to the end of the node, they say. WHY?

```
/* before goes here */
.a {
  color: #fff;
  /* after goes here */
}
```

Seriously WHY? Why before is outter but after is inner? Maybe it's my personal thought but this is how I think. So for final words, don't wait for me if you think you can fix this. You are %100 free to edit my PR to fix this solution. Only thing I know that this bug stands for over one year and has to be fixed.

jeddy3:
> -   Fixed: `max-empty-lines` incorrect line reporting ([#3530](https://github.com/stylelint/stylelint/pull/3530)).

PR #3574, opened 2018-08-19
Add autofix to no-extra-semicolons

### LINKED ISSUE #3572: Add autofix to no-extra-semicolons

Adds autofix of rules [defined in `stylelint-config-recommended`](https://github.com/stylelint/stylelint-config-recommended/blob/master/index.js) and `stylelint-config-standard` so that everyone can start using [`stylelint-config-standard`](https://github.com/stylelint/stylelint-config-standard) easily.

Add autofix to [no-extra-semicolons](https://github.com/stylelint/stylelint/tree/master/lib/rules/no-extra-semicolons)

ota-meshi:

closes #3572

ex.

code:
```css
 ;
a {;}
a { color: red; };
```
fixed:
```css
 
a {}
a { color: red; }
```

### DIFF
--- a/lib/rules/no-extra-semicolons/__tests__/index.js
+++ b/lib/rules/no-extra-semicolons/__tests__/index.js
@@ -6,6 +6,7 @@ const { messages, ruleName } = rule;
 testRule(rule, {
   ruleName,
   config: [true],
+  fix: true,
 
   accept: [
     {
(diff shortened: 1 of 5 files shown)


lib/rules/no-extra-semicolons/__tests__/index.js
```diff
@@ -63,620 +64,779 @@ testRule(rule, {
   reject: [
     {
       code: ";",
+      fixed: "",
```
ntwb: Shouldn't the tests in this `config: [true],` section still include a single semi-colon?

ota-meshi: Are you saying it is better not to delete single semicolons?
if so, I think that this should be deleted because it is an extra semicolon.

jeddy3: > if so, I think that this should be deleted because it is an extra semicolon.

Agreed. I think this semicolon is redundant and should be removed when autofixed.

ntwb:
Changelog in https://github.com/stylelint/stylelint/commit/0dce74b36c46e4d3b80a0a85016095588630d917

-   Added: `no-extra-semicolons` autofix ([#3574](https://github.com/stylelint/stylelint/pull/3574)).

Issue #3587, opened 2018-08-20
Rule 'no-empty-source' fails on 'null' and doesn't accept 'false'


PMK:
I want to use stylelint to lint inside Vue files. When adding Vue files to be linted, I get this error:

```
src/layouts/default.vue
 12:24  ✖  Missed semicolon   CssSyntaxError
```

I need to install processor [@mapbox/stylelint-processor-arbitrary-tags](https://github.com/mapbox/stylelint-processor-arbitrary-tags) (because stylelint-preprocessor-html is deprecated). And when using this, their readme says to disable the rule [no-empty-source](https://stylelint.io/user-guide/rules/no-empty-source/).

I am extending [stylelint-config-recommended](https://github.com/stylelint/stylelint-config-recommended) in my config, and I must overwrite the rule `no-empty-source` to turn it off.

When setting `'no-empty-source': null` in the config, stylelint will return no exit code and all linting will succeed (but it should fail when there are lint-errors).
When setting `'no-empty-source': false` in the config, stylelint returns an error: `Invalid Option: Unexpected option value "false" for rule "no-empty-source"`.

I expect to use `false` if the input is of type Boolean, or using `null` to reset to the default. To set to false or null, I can use the processor to use stylelint in style tags in Vue files.

Configuration file:

```js
module.exports = {
  processors: ['@mapbox/stylelint-processor-arbitrary-tags'],
  extends: ['stylelint-config-standard'],
  rules: {
    'no-empty-source': null
  }
}
```

Stylelint version: `9.5.0`, is also not working in `9.4.0`

CLI: `$ ./node_modules/.bin/stylelint --config ./config/stylelint.config.js --syntax scss ./src/{assets/styles/{,*/}*.scss,**/*.vue}`

To test this, make sure to let your example code fail when using the 'stylelint-config-recommended' rules, then update your configuration to enable the processor and add the rule 'no-empty-source'.

jeddy3:
> I need to install processor @mapbox/stylelint-processor-arbitrary-tags (because stylelint-preprocessor-html is deprecated).

stylelint has built-in support for extracting styles from `.vue` files. You shouldn't need to use a processor.

Can you share the contents of your `src/layouts/default.vue` file with us so we can help determine where the syntax error is occuring?

PMK:
I'm using Nuxt as the framework and Nuxt uses Vue. The declaration of a Vue instance is not needed.

Please see this repo: https://github.com/PMK/stylelint-bug-3587
[...]

PR #3591, opened 2018-08-21
Deemphasize processors and syntax option in the docs

jeddy3:

Ref: https://github.com/stylelint/stylelint/issues/3587

This PR deemphasizes processors and syntax option in the docs (and emphasizes stylelint's built-in support for inferring syntax and extracting styles). It also continues to roll-out the simplified writing styled we started using on the homepage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,13 +46,12 @@ This is the quickest way to get started. We suggest you extend either:
 
 The recommended config turns on just the [possible error](docs/user-guide/rules.md#possible-errors) rules. The standard config extends it by turning on 60 [stylistic rules](docs/user-guide/rules.md#stylistic-issues). We suggest you extend the:
 
-- recommended config if you use a pretty printer like [prettier](https://prettier.io/)
-- standard config if you want stylelint to enforce stylistic conventions
+-   recommended config if you use a pretty printer like [prettier](https://prettier.io/)
+-   standard config if you want stylelint to enforce stylistic conventions
 
 You may want to add rules to your config that [limit language features](docs/user-guide/rules.md#limit-language-features) as these will be specific to your team and/or project.
 
-_If you use language extensions, for example `@if` and `@extends`, you can use a community config like [`stylelint-config-recommended-scss`](https://github.com/kristerkari/stylelint-config-recommended-scss) instead._
-
+*If you use language extensions, for example `@if` and `@extends`, you can use a community config like [`stylelint-config-recommended-scss`](https://github.com/kristerkari/stylelint-config-recommended-scss) instead.*
 
 ### Craft your own config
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -27,30 +27,31 @@ It's mighty because it:
 
 It's easy to get started.
 
-First, decide how you want to use stylelint:
+First, decide if you want to use stylelint:
```
ntwb: I think the user has _already_ decided to use stylelint if they are at this point...

The following 5 bullet items are "how" to use stylelint, they are not sentences, statements, or quotes to aid a user in deciding "if" they should use stylelint.

I suggest reverting this change back to `how`

README.md
```diff
@@ -27,30 +27,31 @@ It's mighty because it:
...
 
 -   [on the command line](docs/user-guide/cli.md)
 -   [in your text editor](docs/user-guide/complementary-tools.md#editor-plugins), for example in VS Code
--   [in for your build tool](docs/user-guide/complementary-tools.md#build-tool-plugins), for example in webpack
+-   [in for your build tool](docs/user-guide/complementary-tools.md#build-tool-plugins), for example in Webpack
```
ntwb: `webpack` is like stylelint, all spelled with lowercase `webpack`

See https://webpack.js.org/branding/#the-name

README.md
jeddy3: Yep, it's a odd one. I was following [this guidance](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#bullet-points-steps):

> - you always use a lead-in line
> - the bullets make sense running on from the lead-in line

"First, decide if you want to use stylelint on the CLI"
"First, decide if you want to use stylelint in your build tool"

Whereas "how" ends up meaning something else "First, decide how you want to use stylelint on the CLI". 

There's probably a better way of writing the lead-in line to remove the oddity you mentioned, but I can't think of one right now :) So, I'll revert the change.

ntwb: Ha, I guess so, language is hard and I only speak one!

README.md
```diff
@@ -10,10 +10,10 @@ It's mighty because it:
 
 -   has over **160 built-in rules** to catch errors, apply limits and enforce stylistic conventions
 -   understands the **latest CSS syntax** including custom properties and level 4 selectors
--   parses **CSS-like syntaxes** like SCSS, Sass, Less and SugarSS
 -   extracts **embedded styles** from HTML, markdown and CSS-in-JS object & template literals
--   automatically **fixes** some violations (*experimental feature*)
+-   parses **CSS-like syntaxes** like SCSS, Sass, Less and SugarSS
 -   supports **plugins** so you can create your own rules or make use of plugins written by the community
+-   automatically **fixes** some violations (*experimental feature*)
```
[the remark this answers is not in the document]
hudochenkov: We have one important [issue](https://github.com/stylelint/stylelint/issues/2643) to solve, before we can recommend `--fix` without risk.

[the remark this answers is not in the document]
jeddy3: > We have one important issue to solve

Yep, definitely that one.

We probably need to resolve the whole order-of-rules-being-important thing too e.g. `indentation` needs to run after a lot of other rules.

PR #3606, opened 2018-08-25
Reset whole return value when ignoring stdin. Fix #3590

### LINKED ISSUE #2833: Fix isPathIgnored regression for .stylelintignore files

### LINKED ISSUE #3590: Fix ignoring of files when code and codeFilename properties are used
> Clearly describe the bug

Even if a file path passed to [`codeFilename`](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/node-api.md#codefilename) option should be ignored by `.stylelintignore`, it's still validated as if it's not ignored.

Rules are not related to this issue.

Any invalid CSS, for example `}`.

Any configurations, even if you enable no rules in your `config`.

`9.5.0`

In the current working directory, you have the following files

**.stylelintignore**

```
style.css
```

**style.css**

```css
}
```

As you can see, `style.css` is not a valid CSS.

**files.js**

```js
'use strict';

const {inspect} = require('util');
const {join} = require('path');
const {lint} = require('stylelint');

(async () => {
  const result = await lint({
    files: join(__dirname, 'style.css'),
    config: {rules: {}}
  });

  console.log(inspect(result, {colors: true, depth: Infinity}));
})();
```

**code.js**

```js
'use strict';

const {inspect} = require('util');
const {join} = require('path');
const {lint} = require('stylelint');

(async () => {
  const result = await lint({
    code: '}',
    codeFilename: join(__dirname, 'style.css'),
    config: {rules: {}}
  });

  console.log(inspect(result, {colors: true, depth: Infinity}));
})();
```

No.

When you run `node files.js`, it prints

```
{ errored: false, output: '[]', results: [] }
```

It's OK. `style.css` is successfully ignored by `.stylelintignore`.

Then, I expect `node code.js` to results in exactly the same output, as `code.js` and `file.js` do essentially the same thing.

`node code.js` prints:

```
{ errored: true,
  output: '',
  results:
   [ { source: '/Users/shinnn/example/style.css',
       deprecations: [],
       invalidOptionWarnings: [],
       parseErrors: [],
       errored: true,
       warnings:
        [ { line: 1,
            column: 1,
            rule: 'CssSyntaxError',
            severity: 'error',
            text: 'Unexpected } (CssSyntaxError)' } ] } ] }
```

This is unexpected result. `style.css` should be ignored by `.stylelintignore`, but it's actually validated and emits `CssSyntaxError`.

thibaudcolas:
~Fixes #2833, and it also~ Fixes #3590 ~which is the same issue~.

Edit: my bad, #2833 is about `isPathIgnored`. This fixes the ignoring for the `lint()` API only, so #3590.

My tests showed everything working as expected, while the repro in #3590 did not. The difference between the two is that I was only checking `output` from the stylelint results… which is the only thing that (half) works.

For inputs via stdin, the results are ignored **after** linting, and it was only resetting [`output`](https://stylelint.io/user-guide/node-api/#output), not all other return values (`errored`, etc):

https://github.com/stylelint/stylelint/blob/2584d15e514fc99fb473e8d24175012c20627ff3/lib/standalone.js#L135-L156

…so all is well if you're using the CLI, or only look at `output` in your API integration (like I did in https://github.com/thibaudcolas/stylelint-repro), but for API usage relying on other return values it fails.

It also doesn't respect the formatter – with the default formatter, if there are no errors, `output` should be `"[]"` not `""`.

---

I went for the simplest fix. If there is no reason to lint files that are ignored (right?), then it would be better to ignore files before linting, over here: https://github.com/stylelint/stylelint/blob/2584d15e514fc99fb473e8d24175012c20627ff3/lib/standalone.js#L98-L104

That's a separate issue though.

### DIFF
--- a/lib/__tests__/stylelintignore-test/stylelintignore.test.js
+++ b/lib/__tests__/stylelintignore-test/stylelintignore.test.js
@@ -3,7 +3,7 @@
 const path = require("path");
 const standalone = require("../../standalone");
 
-describe("standalone with .stylelintignore file ignoring one file", () => {
+describe("stylelintignore", () => {
   let actualCwd;
   let results;
   const fixturesPath = path.join(__dirname, "../fixtures");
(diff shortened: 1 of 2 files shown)


CAYdenberg:
@thibaudcolas Does it fix `isPathIgnored`, or does it just ignore but not properly report? I had a test like this for isPathIgnored specifically: https://github.com/stylelint/stylelint/pull/3021/commits/00a31d6e18ef14ed2e83a956bd054f1d34a9cd70
that I could never get to pass.

Honestly we should probably just merge this anyway, as this has been hanging around for awhile and is affecting a lot of people. We can open another issue for isPathIgnored, but I don't know how many people are really relying on that part of the api? @stylelint/core can anyone take a second look at this?

thibaudcolas:
@CAYdenberg this doesn't change anything to how `isPathIgnored` works. As far as I can tell, this part of the API has an entirely different code path. Looking at its source, it doesn't seem to support ignore files:

https://github.com/stylelint/stylelint/blob/208517b5510b94f970f8db98875c0f5dc51f0a95/lib/isPathIgnored.js#L18-L26

I also don't get why `micromatch` is used there, whereas the ignore checks in the linting API use `ignore`:

https://github.com/stylelint/stylelint/blob/208517b5510b94f970f8db98875c0f5dc51f0a95/lib/standalone.js#L48-L61

PR #3623, opened 2018-08-28
Support for CSS Modules in selector specificity rules (#3604).

### LINKED ISSUE #3604: Fix false positives for functional pseudo classes in no-descending-specificity
> Clearly describe the bug

There are false positives for `no-descending-specificity` when using CSS Modules

`no-descending-specificity`

```css
:global(.buildResultsSummaryTable) span {
  color: #444;
}
 
:global(.buildResultsSummaryTable) span:nth-child(2) {
  color: #666;
}
```

```json
{
  "rules": {
  }
}
```

`9.5.0`

CLI with `stylelint --ignore-path .eslintignore \"src/**/*.css\"`

CSS Modules

No warnings to be flagged.

```shell
Expected selector ":global(.buildResultsSummaryTable) span:nth-child(2)" to come before selector ":global(.buildResultsSummaryTable) span" (no-descending-specificity)
```

keeganstreet:
Closes #3604

### DIFF
--- a/lib/rules/no-descending-specificity/__tests__/index.js
+++ b/lib/rules/no-descending-specificity/__tests__/index.js
@@ -79,6 +79,10 @@ testRule(rule, {
     },
     {
       code: ":root { foo: {}; }"
+    },
+    {
+      code:
+        ":global(.buildResultsSummaryTable) span {} :global(.buildResultsSummaryTable) span:nth-child(2) {}"
     }
   ],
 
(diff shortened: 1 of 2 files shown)


jeddy3:
> -   Fixed: `no-descending-specificity` false positives for CSS Modules functional pseudo-classes ([#3623](https://github.com/stylelint/stylelint/pull/3623)).

PR #3671, opened 2018-09-15
Add auto fix for selector-pseudo-class-case rule

### LINKED ISSUE #3661: Add autofix to selector-pseudo-class-case

Add autofix to [selector-pseudo-class-case](https://github.com/stylelint/stylelint/tree/master/lib/rules/selector-pseudo-class-case)

Bilie:
Closes #3661

No, it's self explanatory.


lib/rules/selector-pseudo-class-case/index.js
```diff
@@ -61,6 +61,11 @@ const rule = function(expectation) {
             return;
           }
 
+          if (context.fix) {
+            rule.selector = rule.selector.replace(pseudo, expectedPseudo);
```
ota-meshi: You can also directly replace the value of `pseudoNode`, and receive it as the return value of `parseSelector`.

lib/rules/selector-pseudo-class-case/__tests__/index.js
```diff
@@ -114,120 +115,161 @@ testRule(rule, {
...
+      line: 1,
+      column: 16
+    },
+    {
+      code: "a:Hover /* comment */ { color: pink; }",
+      fixed: "a:hover /* comment */ { color: pink; }",
+      message: messages.expected(":Hover", ":hover"),
+      line: 1,
+      column: 2
+    },
```
ota-meshi: Could you add the following test cases?

`a:Hover,/*comment*/ .b {color: pink;}`

Bilie: Hi, thank you very much for the example! I have updated the PR

lib/rules/selector-pseudo-class-case/index.js
```diff
@@ -61,6 +61,19 @@ const rule = function(expectation) {
             return;
           }
 
+          if (context.fix) {
+            if (rule.raws.selector) {
+              rule.raws.selector.raw = rule.raws.selector.raw.replace(
```
ota-meshi: I found a problem with the CSS as follow.

```css
a::FIRST-LETTER, a:FIRST {color: pink;}
```

The result of fixed this is as follow.

```css
a::first-LETTER, a:FIRST {color: pink;}
```

To improve this, I think we need to change this as follows.

```js
pseudoNode.value = expectedPseudo;
```

Then we process the result of parseSelector as follows.

```js
const fixedSelector = parseSelector(rule.raws.selector ? rule.raws.selector.row : rule.selector, ...
//...

if (context.fix) {
  if (rule.raws.selector) {
    rule.raws.selector.raw = fixedSelector
  } else {
    rule.selector = fixedSelector;
  }
}
```

Bilie: Hi, thank you for spotting this case! I have updated the PR

lib/rules/selector-pseudo-class-case/index.js
```diff
@@ -28,45 +28,63 @@ const rule = function(expectation) {
...
+
+            if (pseudo === expectedPseudo) {
+              return;
+            }
+
+            if (context.fix) {
+              const fixedSelector = selector.replace(pseudo, expectedPseudo);
+
+              if (rule.raws.selector) {
+                rule.raws.selector.raw = rule.raws.selector.raw.replace(
```
ota-meshi: The following css will not be autofixed.

```css
.foo > /* comment */ a:Hover { color: pink; }
```

I think that this problem will not occur with the method I [proposed before](https://github.com/stylelint/stylelint/pull/3671#discussion_r219185245).
Do you have reason to process with `replace`?

I'm sorry if you can not understand, because I can not English.

Bilie: Hi, certainly not your English! :) I am just still trying to understand the postcss api, all the possible test cases, and what is the best way to solve a certain issue. Thank you for your help!

jeddy3:
> -   Added: `selector-pseudo-class-case` autofix ([#3671](https://github.com/stylelint/stylelint/pull/3671)).

PR #3672, opened 2018-09-15
Add autofix to selector-pseudo-element-case

### LINKED ISSUE #3662: Add autofix to selector-pseudo-element-case

Add autofix to [selector-pseudo-element-case](https://github.com/stylelint/stylelint/tree/master/lib/rules/selector-pseudo-element-case)

boccob:

Closes #3662

Example

code:
```
a:bEfOrE {}
```

fixed:
```
a:before {}
```


lib/rules/selector-pseudo-element-case/__tests__/index.js
```diff
@@ -6,6 +6,7 @@ const { messages, ruleName } = rule;
 testRule(rule, {
   ruleName,
   config: ["lower"],
+  fix: true,
```
ota-meshi: Could you add a test case with comments in the selector?

boccob: Unfortunately it won't work until update to the last version of `postcss-selector-parser` (#3284). 
I don't sure that it makes sense to add some workaround.

ota-meshi: In [selectorCombinatorSpaceChecker.js](https://github.com/stylelint/stylelint/blob/master/lib/rules/selectorCombinatorSpaceChecker.js#L14), it choose the selector to pass to `parseSelector` beforehand.

boccob: I added few new test cases with comments inside selectors.

boccob:
Hi @evilebottnawi, @ntwb, @hudochenkov,
Can you please check the PR?

lib/utils/transformSelector.js
```diff
@@ -0,0 +1,16 @@
...
+
+const selectorParser = require("postcss-selector-parser");
+
+module.exports = function(
+  result /*: Object*/,
+  node /*: Object*/,
+  cb /*: Function*/
+) {
+  try {
+    return selectorParser(cb).processSync(node, { updateSelector: true });
```
hudochenkov: Why `updateSelector: true` is needed here? It's `true` by default. https://github.com/postcss/postcss-selector-parser/blame/v3.1.1/API.md#L678-L680

vankop: Does this really block PR?

hudochenkov: It's a new utility which might work the same as existing utility. It would create more confusion. So I wouldn't merge just yet.

boccob: It doesn't look like `updateSelector` option has default value `true` (probably documentation issue).
https://github.com/postcss/postcss-selector-parser/blob/v3.1.1/src/processor.js#L65
Also, check dts file:
https://github.com/postcss/postcss-selector-parser/blob/v3.1.1/postcss-selector-parser.d.ts#L89

hudochenkov: You're right. Looks like behavior changed, but docs weren't updated.

hudochenkov:
> -   Added: `selector-pseudo-element-case` autofix ([#3672](https://github.com/stylelint/stylelint/pull/3672)).

PR #3685, opened 2018-09-22
feat: add autofix to media-feature-name-case

### LINKED ISSUE #3659: Add autofix to media-feature-name-case

Add autofix to [media-feature-name-case](https://github.com/stylelint/stylelint/tree/master/lib/rules/media-feature-name-case)

EliazTray:

e.g. "Closes #3659 " or "None, as it's a documentation fix."


jeddy3:
> -   Added: `media-feature-name-case` autofix ([#3685](https://github.com/stylelint/stylelint/pull/3685)).

PR #3687, opened 2018-09-22
Update to postcss-less@3

jwilsson:
Closes #3686

The biggest change is probably the removal of `Import` nodes from `postcss-less`, they will now always be a usual `AtRule` node. I've removed tests and doc notes around that.

Otherwise, it should be pretty straightforward. I've left some inline notes too.


lib/rules/unit-case/index.js
```diff
@@ -60,9 +60,13 @@ const rule = function(expectation) {
       });
     }
 
-    root.walkAtRules(/^media$/i, atRule =>
-      check(atRule, atRule.params, atRuleParamIndex)
-    );
+    root.walkAtRules(atRule => {
+      if (!/^media$/i.test(atRule.name) && !atRule.variable) {
```
jwilsson: There were a bunch of tests for `@variable` syntax, hence this change even though it's non-standard syntax. Perhaps those tests should be updated instead?

lib/rules/unit-no-unknown/index.js
```diff
@@ -72,9 +72,13 @@ const rule = function(actual, options) {
       });
     }
 
-    root.walkAtRules(/^media$/i, atRule =>
-      check(atRule, atRule.params, atRuleParamIndex)
-    );
+    root.walkAtRules(atRule => {
+      if (!/^media$/i.test(atRule.name) && !atRule.variable) {
```
jwilsson: Same as above, there were a bunch of tests for `@variable` syntax, hence this change even though it's non-standard syntax. Perhaps those tests should be updated instead?

lib/utils/isStandardSyntaxAtRule.js
```diff
@@ -7,17 +7,21 @@
  * @param {atRule} postcss at-rule node
  * @return {boolean} If `true`, the declaration is standard
  */
-module.exports = function(atRule /*: postcss$atRule*/) /*: boolean*/ {
+module.exports = function(atRule /*: Object*/) /*: boolean*/ {
```
jwilsson: Perhaps someone better than me at Flow can help out with this.

hudochenkov: I have no clue about Flow, to be honest. Maybe new properties should be added to this lines?

https://github.com/stylelint/stylelint/blob/6f86f23fe568daa6185d282079358ffc1a26fb54/flow-typed/postcss.js#L30-L38

jwilsson:
@ hudochenkov Thanks for the review! I've added a new test for `isStandardSyntaxAtRule`.

`no-extra-semicolons` should be fine with the current tests, there are a few for mixins where the node type where changed from `Rule` to `AtRule`.

lib/utils/isStandardSyntaxAtRule.js
ntwb: @CAYdenberg Any chance you've got a spare moment to check on this Flow issue please?

jeddy3:
@hudochenkov Have your concerns been addressed?

jeddy3:
I rebased this on `master`, but node 10 (but not node 8) is failing on appveyor:

```shell
FAIL lib/__tests__/ignore.test.js
  ● using ignoreFiles with input files that would cause a postcss syntax error › no-syntax-error.css found
    expect(received).not.toBe(expected) // Object.is equality
    Expected: -1
    Received: -1
      307 | 
      308 |   it("no-syntax-error.css found", () => {
    > 309 |     expect(results[0].source.indexOf("no-syntax-error.css")).not.toBe(-1);
          |                                                                  ^
      310 |   });
      311 | 
      312 |   it("no-syntax-error.css linted", () => {
      at Object.toBe (lib/__tests__/ignore.test.js:309:66)
```

Anyone got an idea why?

evilebottnawi:
@jeddy3 maybe `-1` is string?

Issue #3638, opened 2018-09-06
Release 9.6.0


jeddy3:
- [x] stylelint release
- [x] stylelint-demo update
- [x] stylelint.io update
- [x] tweet

There are a lot of autofix additions in `master`! It'd be nice to [get the rest in](https://github.com/stylelint/stylelint/pulls) in if anyone has time to review?

Either way, I can run a release this weekend.

cc / @stylelint/core

ntwb:
I'll try to get to those later today or tomorrow if no one beats me to it

jeddy3:
I intend to do a release later today. There's a handful of autofix PRs that need a second review. It'd be good to get those in.

jeddy3:
Pre-deploy tests are failing on my machine.

```shell
Summary of all failing tests
 FAIL  lib/__tests__/standalone-syntax.test.js
  ● standalone with syntax set by extension › By extension › parsed each according to its extension

    expect(received).toBe(expected) // Object.is equality

    Expected: 1
    Received: 0

      223 |
      224 |       results.forEach(result => {
    > 225 |         expect(result.warnings.length).toBe(1);
          |                                        ^
      226 |         expect(result.warnings[0].rule).toBe("color-no-invalid-hex");
      227 |       });
      228 |     });

      at toBe (lib/__tests__/standalone-syntax.test.js:225:40)
          at Array.forEach (<anonymous>)
      at Object.forEach (lib/__tests__/standalone-syntax.test.js:224:15)

 FAIL  lib/__tests__/ignore.test.js
  ● using ignoreFiles with input files that would cause a postcss syntax error › syntax-error-ignored.scss found

    expect(received).not.toBe(expected) // Object.is equality

    Expected: -1
    Received: -1

      315 |
      316 |   it("syntax-error-ignored.scss found", () => {
    > 317 |     expect(results[1].source.indexOf("syntax-error-ignored.scss")).not.toBe(-1);
          |     ^
      318 |   });
      319 |
      320 |   it("syntax-error-ignored.scss not linted", () => {

      at Object.toBe (lib/__tests__/ignore.test.js:317:72)

 FAIL  lib/rules/function-whitespace-after/__tests__/index.js
  ● function-whitespace-after › reject › ["always"] › "export default <img style={{ ['transform']: 'translate(1, 1)scale(3)' }} />;" › no description

    expect(received).toBe(expected) // Object.is equality

    Expected: "Expected whitespace after \")\" (function-whitespace-after)"
[...]

PR #3689, opened 2018-09-24
Fix ESLint Jest warnings

jeddy3:

Ref: https://github.com/stylelint/stylelint/issues/3638

Fixes up the 150 or so ESLint warnings in `master`. The just plugin is useful as it caught these mistakes.

Warnings cames from:

- https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/prefer-to-have-length.md
- https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-identical-title.md

### DIFF
--- a/lib/__tests__/defaultSeverity.test.js
+++ b/lib/__tests__/defaultSeverity.test.js
@@ -13,7 +13,7 @@ it("`defaultSeverity` option set to warning", () => {
     .process("a {}", { from: undefined }, config)
     .then(result => {
       const warnings = result.warnings();
-      expect(warnings.length).toBe(1);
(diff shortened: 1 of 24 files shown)


hudochenkov:
> I don't see ESLint running in Travis, is there a reason for that?

@thibaudcolas it's because it runs in `CI=pretest` job only.

jeddy3:
> As a follow-up I'd suggest turning those two rules into errors (ref: stylelint/eslint-config-stylelint#53 (comment)).

@thibaudcolas SGTM. Feel free to PR the change to the config.

PR #3720, opened 2018-10-15
Add autofix to media-feature-parentheses-space-inside

yepninja:
#3585

### DIFF
--- a/lib/rules/media-feature-parentheses-space-inside/index.js
+++ b/lib/rules/media-feature-parentheses-space-inside/index.js
@@ -31,7 +31,7 @@ const rule = function(expectation, options, context) {
       // will be at atRule.raws.params.raw
       const params = _.get(atRule, "raws.params.raw", atRule.params);
       const indexBoost = atRuleParamIndex(atRule);
-      const errors = [];
+      const violations = [];
 
       const parsedParams = valueParser(params).walk(node => {
         if (node.type === "function") {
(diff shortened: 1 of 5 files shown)


yepninja:
I had to fix snapshots. Now the order of the errors is more correct.

lib/rules/media-feature-parentheses-space-inside/index.js
```diff
@@ -31,50 +31,62 @@ const rule = function(expectation) {
       // will be at atRule.raws.params.raw
       const params = _.get(atRule, "raws.params.raw", atRule.params);
       const indexBoost = atRuleParamIndex(atRule);
+      const errors = [];
```
jeddy3: Can we rename this to `violations`, please?

It's a convention we've adopted to:

- help distinguish between things caught by the linter and actual errors, for example thrown by the parsers.
- act as an intermediary before those things are categorised into a particular severity e.g. "error" or "warning".

yepninja: Agreed. No problem at all.

yepninja: Done!

lib/rules/media-feature-parentheses-space-inside/index.js
```diff
@@ -31,50 +31,62 @@ const rule = function(expectation) {
...
+            }
+            if (node.after === " ") {
+              if (context.fix) node.after = "";
+              errors.push({
+                message: messages.rejectedClosing,
+                index: node.sourceIndex - 2 + len + indexBoost
+              });
+            }
+          }
+          if (expectation === "always") {
```
vankop:
```suggestion
         else if (expectation === "always") {
```
or 

```suggestion
         else {
```
will be enough due to `validOptions` check on the top of rule definition

jeddy3:
> -   Added: `media-feature-parentheses-space-inside` autofix ([#3720](https://github.com/stylelint/stylelint/pull/3720)).

Issue #3723, opened 2018-10-16
max-nesting-depth: ignore pseudoclasses


inspiratweb:
> What is the problem you're trying to solve?

Pseudoclasses shoud not be affected by max-nesting-depth like at-rules

> What solution would you like to see?

A new ignore rule `blockless-pseudoclasses` working in the same way as `blockless-at-rules`does

PR #3724, opened 2018-10-16
Blockless pseudoclasses

inspiratweb:
https://github.com/stylelint/stylelint/issues/3723

Having max-nesting-depth = 1, it could be useful adding a rule to ìgnore` like
```
'max-nesting-depth': [1, {
  ignore: ["blockless-pseudoelements"]
}],
```

Allowing for example this css code
```
a {
  b { /* 1 */
    &:nest { /* 1 */
      &:nest-lvl2 { /* 1 */
        top: 0;
      }
    }
  }
}
```

### DIFF
--- a/lib/rules/max-nesting-depth/README.md
+++ b/lib/rules/max-nesting-depth/README.md
@@ -148,6 +148,65 @@ a {
 }
 ```
 
+### `ignore: ["blockless-pseudoclasses"]`
+
+Ignore pseudoclases that only wrap other rules, and do not themselves have declaration blocks.
+
+For example, with `1`:
+
+The following patterns are considered violations:
+
+As the pseudoclasses have a declarations blocks.
+
+```css
+a {
+  &:hover { /* 1 */
+    b { /* 2 */
+      top: 0;
+    }
+  }
+}
+```
+
+```css
+a {
+  &:nest { /* 1 */
+    &:nest-lvl2 { /* 1 */
+      top: 0;
+      .class { /* 2 */
+        bottom: 0;
+      }
+    }
+  }
+}
+```
+
+The following patterns are *not* considered violations:
+
+As all of the following pseudoclasses rules would have a nesting depth of just 1.
+
+```css
+a {
+  b { /* 1 */
+    &:hover { /* 1 */
+      color: pink;
+    }
+  }
+}
+```
+
+```css
+a {
+  b { /* 1 */
+    &:nest { /* 1 */
+      &:nest-lvl2 { /* 1 */
+        top: 0;
+      }
+    }
+  }
+}
+```
+
 ### `ignoreAtRules: ["/regex/", "string"]`
 
 Ignore the specified at-rules.
--- a/lib/rules/max-nesting-depth/__tests__/index.js
+++ b/lib/rules/max-nesting-depth/__tests__/index.js
@@ -112,6 +112,38 @@ testRule(rule, {
 
 testRule(rule, {
   ruleName,
+  config: [1, { ignore: ["blockless-pseudoclasses"] }],
+
+  accept: [
+    {
+      code: "a { b { top: 0; }}"
+    },
+    {
+      code: "a { b { &:hover { top: 0; }}}"
+    },
+    {
+      code: "a { b { &:nest { &:nest-lvl2 { top: 0; }}}}"
+    }
+  ],
+
+  reject: [
+    {
+      code: "a { b { c { top: 0; }}}",
+      message: messages.expected(1)
+    },
+    {
+      code: "a { &:hover { b { top: 0; }}}",
+      message: messages.expected(1)
+    },
+    {
+      code: "a { &:nest { &:nest-lvl2 { top: 0; .class { bottom: 0; }}}}",
+      message: messages.expected(1)
+    }
+  ]
+});
+
+testRule(rule, {
+  ruleName,
   config: [1, { ignoreAtRules: ["media", "/^my-/"] }],
 
   accept: [
--- a/lib/rules/max-nesting-depth/index.js
+++ b/lib/rules/max-nesting-depth/index.js
@@ -31,7 +31,7 @@ const rule = function(max, options) {
         optional: true,
         actual: options,
         possible: {
-          ignore: ["blockless-at-rules"],
+          ignore: ["blockless-at-rules", "blockless-pseudoclasses"],
           ignoreAtRules: [_.isString]
         }
       }
@@ -82,9 +82,12 @@ const rule = function(max, options) {
     }
 
     if (
-      optionsMatches(options, "ignore", "blockless-at-rules") &&
-      node.type === "atrule" &&
-      node.every(child => child.type !== "decl")
+      (optionsMatches(options, "ignore", "blockless-at-rules") &&
+        node.type === "atrule" &&
+        node.every(child => child.type !== "decl")) ||
+      (optionsMatches(options, "ignore", "blockless-pseudoclasses") &&
+        node.type === "rule" &&
+        node.some(child => child.selector && child.selector.startsWith("&:")))
     ) {
       return nestingDepth(parent, level);
     }


lib/rules/max-nesting-depth/index.js
```diff
@@ -82,9 +82,12 @@ const rule = function(max, options) {
...
     if (
-      optionsMatches(options, "ignore", "blockless-at-rules") &&
-      node.type === "atrule" &&
-      node.every(child => child.type !== "decl")
+      (optionsMatches(options, "ignore", "blockless-at-rules") &&
+        node.type === "atrule" &&
+        node.every(child => child.type !== "decl")) ||
+      (optionsMatches(options, "ignore", "blockless-pseudoclasses") &&
+        node.type === "rule" &&
+        node.some(child => child.selector && child.selector.startsWith("&:")))
```
jeddy3: I think you're missing the `node.every(child => child.type !== "decl")` conditional to enforce the "and do not themselves have declaration blocks." bit of the description.

lib/rules/max-nesting-depth/__tests__/index.js
```diff
@@ -110,6 +110,38 @@ testRule(rule, {
...
+testRule(rule, {
+  ruleName,
+  config: [1, { ignore: ["blockless-pseudoclasses"] }],
+
+  accept: [
+    {
+      code: "a { b { top: 0; }}"
+    },
+    {
+      code: "a { b { &:hover { top: 0; }}}"
```
jeddy3: I believe this should be rejected as there are two levels here.

lib/rules/max-nesting-depth/__tests__/index.js
```diff
@@ -110,6 +110,38 @@ testRule(rule, {
...
+    },
+    {
+      code: "a { b { &:nest { &:nest-lvl2 { top: 0; }}}}"
```
jeddy3: I believe this should be rejected as there are two levels here.

lib/rules/max-nesting-depth/__tests__/index.js
```diff
@@ -110,6 +110,38 @@ testRule(rule, {
...
+    }
+  ],
+
+  reject: [
+    {
+      code: "a { b { c { top: 0; }}}",
+      message: messages.expected(1)
+    },
+    {
+      code: "a { &:hover { b { top: 0; }}}",
```
jeddy3: I believe this should be accepted as there is only one level here.

lib/rules/max-nesting-depth/__tests__/index.js
```diff
@@ -110,6 +110,38 @@ testRule(rule, {
...
+      message: messages.expected(1)
+    },
+    {
+      code: "a { &:nest { &:nest-lvl2 { top: 0; .class { bottom: 0; }}}}",
```
jeddy3: Yes is correctly rejected as there are two levels.

lib/rules/max-nesting-depth/index.js
jeddy3: I believe `child.selector.startsWith("&:")` will also catch [pseudo-elements](https://www.w3.org/TR/css-pseudo-4/) e.g. `::placeholder`. Can you please exclude them by checking that the 3rd character isn't a `:`?

lib/rules/max-nesting-depth/__tests__/index.js
```diff
@@ -110,6 +110,38 @@ testRule(rule, {
   ]
 });
 
+testRule(rule, {
+  ruleName,
+  config: [1, { ignore: ["blockless-pseudoclasses"] }],
```
jeddy3: I believe [the word is hyphenated](https://www.w3.org/TR/selectors-4/#pseudo-classes) e.g. `"blockless-pseudo-classes"`. Please edit in the documentation too.

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,65 @@ a {
...
+
+Ignore pseudoclases that only wrap other rules, and do not themselves have declaration blocks.
+
+For example, with `1`:
+
+The following patterns are considered violations:
+
+As the pseudoclasses have a declarations blocks.
+
+```css
```
jeddy3: I'm not sure this is a violation:

```css
a {
  &:hover { /* ignored */
    b { /* 1 */
      top: 0;
    }
  }
}
```

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,65 @@ a {
...
+a {
+  &:hover { /* 1 */
+    b { /* 2 */
+      top: 0;
+    }
+  }
+}
+```
+
+```css
```
jeddy3:
```css
a {
  &:nest { /* ignored */
    &:nest-lvl2 { /* 1 */
      top: 0;
      .class { /* 2 */
        bottom: 0;
      }
    }
  }
}
```

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,65 @@ a {
...
+    }
+  }
+}
+```
+
+The following patterns are *not* considered violations:
+
+As all of the following pseudoclasses rules would have a nesting depth of just 1.
+
+```css
```
jeddy3: I think this is a violation:

```css
a {
  b { /* 1 */
    &:hover { /* 2 */
      color: pink;
    }
  }
}
```

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,65 @@ a {
...
+a {
+  b { /* 1 */
+    &:hover { /* 1 */
+      color: pink;
+    }
+  }
+}
+```
+
+```css
```
jeddy3: I think this is a violation:

```css
a {
  b { /* 1 */
    &:nest { /* ignore */
      &:nest-lvl2 { /* 2 */
        top: 0;
      }
    }
  }
}
```

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,65 @@ a {
 }
 ```
 
+### `ignore: ["blockless-pseudoclasses"]`
+
+Ignore pseudoclases that only wrap other rules, and do not themselves have declaration blocks.
```
jeddy3: Typo: "pseudo-classes"

lib/rules/max-nesting-depth/index.js
inspiratweb: Sorry I c&p the at-rules block, I have to remove that description line.

lib/rules/max-nesting-depth/__tests__/index.js
inspiratweb:
```
a {
  b { // nesting level 1
    &:hover { // pseudo-class (parent depth level 1 taken)
      top: 0;
    }
  }
}
```

lib/rules/max-nesting-depth/__tests__/index.js
inspiratweb:
```
a {
  b { // nesting level 1
    &:nest { // pseudo-class (parent depth level 1 taken)
      &:nest-lvl2 { // pseudo-class (parent depth level 1 taken)
        top: 0;
      }
    }
  }
}
```

lib/rules/max-nesting-depth/__tests__/index.js
inspiratweb: You are right!
```
a {
  &:hover { // pseudo-class (parent depth level 0 taken)
    b { // nesting level 1
      top: 0;
    }
  }
}
```

lib/rules/max-nesting-depth/__tests__/index.js
inspiratweb: I think the code was wrong, this test should be accepted
```
a {
  &:hover { // pseudo-class (parent depth level 0 taken)
    b { // nesting level 1
      top: 0;
    }
  }
}
```

lib/rules/max-nesting-depth/README.md
inspiratweb: You are right, I update that and also the examples given to be easier to read

inspiratweb:
Yes, it was a mistake naming it `blockless` the second option is what i tried to do.
`ignore: ["pseudo-classes"]` LGTM

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,125 @@ a {
 }
 ```
 
+### `ignore: ["pseudo-classes"]`
+
+Ignore pseudoclases.
```
jeddy3: Typo: `pseudo-classes`.

inspiratweb: In my first approach the first two examples were allowed but I believe what you say may make sense. I'll think about it a bit more and I'll try to propose a solution for this case.

The main reason for doing this rule was to prevent declarations like...
```
.a {
  .b:hover {}
}
```
and allow 
```
.a {
  .b {
    &:hover {}
  }
}
```

Thanks!

inspiratweb: This is how it should work in my opinion

__1st example__
I agree with you because `b` selector is not `:hover` dependent and it's a kind of trick to use it like that. I need to fix this.

__2nd example__
I think it's the main purpose of this issue, I give you a real example for this use case
http://jsfiddle.net/inspiratweb/r1d6qpne/ in my opinion it works as it should because it's just a selector for that nesting level.

__3rd example__
It's OK because although they are different selectors, both of them are pseudo-classes.
It works like...
```
a {
  &:hover {
     top: 0;
  }
  &:focus {
    top: 0;
  }
}
```
Do you agree with me?

jeddy3: Sounds good to me.

Thanks for working it through!

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,145 @@ a {
 }
 ```
 
+### `ignore: ["pseudo-classes"]`
+
+Ignore pseudo-clases.
```
jeddy3: We should be really explicit with what is happening here as the behaviour might not be intuitive to all users. How's about?:

```suggestion
Ignore rules where the first selector in each selector list item is a pseudo-class
```

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,145 @@ a {
...
+    .c { /* 2 */
+      top: 0;
+    }
+  }
+}
+```
+
+```css
+.a {
+  &:hover {
```
jeddy3:
```suggestion
  &:hover { /* ignored */
```

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,145 @@ a {
...
+.a {
+  .b { /* 1 */
+    &::selection { /* 2 */
+      color: #64FFDA;
+    }
+  }
+}
+```
+
+```css
```
jeddy3: I think we can delete this example.

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,145 @@ a {
...
+```
+
+The following patterns are *not* considered violations:
+
+As all of the following pseudoclasses rules would have a nesting depth of just 1.
+
+```css
+.a {
+  .b { /* 1 */
+    &:hover {
```
jeddy3:
```suggestion
    &:hover { /* ignored */
```

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,145 @@ a {
...
+        top: 0;
+      }
+    }
+  }
+}
+```
+
+```css
+.a {
+  &:hover {
```
jeddy3:
```suggestion
  &:hover {  /* ignored */
```

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,145 @@ a {
...
+      top: 0;
+    }
+  }
+}
+```
+
+```css
+.a {
+  .b { /* 1 */
+    &:nest {
```
jeddy3:
```suggestion
    &:nest {  /* ignored */
```

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,145 @@ a {
...
+      &:nest-lvl2 {
```
jeddy3:
```suggestion
      &:nest-lvl2 {  /* ignored */
```

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,145 @@ a {
...
+.a {
+  &:hover {
+    .b { /* 1 */
+      top: 0;
+    }
+  }
+}
+```
+
+```css
```
jeddy3: I think we can delete this example... it's similar to the previous two.

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,145 @@ a {
...
+        }
+      }
+    }
+  }
+}
+```
+
+```css
+.a {
+  &:nest {
```
jeddy3:
```suggestion
  &:nest {  /* ignored */
```

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,145 @@ a {
...
+    &:nest-lvl2 {
```
jeddy3:
```suggestion
    &:nest-lvl2 {  /* ignored */
```

lib/rules/max-nesting-depth/README.md
```diff
@@ -148,6 +148,145 @@ a {
...
+      }
+    }
+  }
+}
+```
+
+```css
+.a {
+  .b { /* 1 */
+    &:hover, &:focus {
```
jeddy3:
```suggestion
    &:hover, &:focus {  /* ignored */
```

inspiratweb:
Perfect! suggestions applied :D

PR #3736, opened 2018-10-21
Add ignoreFunctions: [] to unit-no-unkown

### LINKED ISSUE #3424: Add ignoreFunctions: [] to unit-no-unknown

Ignore an unknown unit format based on the entire value of the attribute rather than just the unit. I find this useful specifically in cases of using background-image: image-set with units such as 1x, 2x, 3x, etc. For example: 

```css
div {
  background-image: image-set(
	);
}
```
Yes, we could simply use the ignoreUnits secondary option with "x" as a string literal or regex, but this is too broad as it will allow values such as 20x or 200x, which are likely typos where "px" was the intended unit. Since image-set units will likely only ever be single digit values, I only want to allow "x" as a unit in situations where the value is between 1x and 4x. If we had another secondary option such as "ignoreValue" it would solve this problem as I could create a regex matching rule such as ["/[1-4]x/"] which would cover those specific cases. 

A new secondary option called "ignoreValue" which takes exactly the same parameter format as the existing ignoreUnits secondary option, but instead of comparing it with the units, it would compare it with the valueNode.value. 

I suspect my use case is achievable via another rule such as declaration-property-value-whitelist, however in the case of an image-set, given the variability of the filenames and the format of background-image: image-set syntax, the regex that would need to be created seems rather complicated compared to how easy it is to create a regex for just a unit value. Also, adding an "ignoreValue" secondary option is a very trivial code and documentation change. In fact..

I have created a fork/branch and implemented this feature (including documentation) here:
https://github.com/minorgod/stylelint/tree/unit-no-unknown_ignoreFormat/lib/rules/unit-no-unknown

The existing "ignoreUnits" secondary option implementation could be modified so that it compares both the units and the value, but this could potentially break the expected behavior of some people's configuration, so it would probably be better just to add an new "ignoreValue" secondary option. 

If this feature is approved I'll be happy to submit a pull request. If there's another way to easily achieve what I'm trying to do via an existing stylelint rule please advise.

Bilie:
Closes #3424

No, it's self explanatory.

### DIFF
--- a/lib/rules/unit-no-unknown/README.md
+++ b/lib/rules/unit-no-unknown/README.md
@@ -83,3 +83,43 @@ a {
   width: 10my-other-unit;
 }
 ```
+
+### `ignoreFunctions: ["/regex/", "string"]`
+
+Given:
+
+```js
+["image-set", "/^my-/", "/^YOUR-/i"]
+```
+
+The following patterns are *not* considered violations:
+
+```css
+a {
+  background-image: image-set(
+    '/images/some-image-1x.jpg' 1x,
+    '/images/some-image-2x.jpg' 2x,
+    '/images/some-image-3x.jpg' 3x
+  );
+}
+```
+
+```css
+a {
+  background-image: my-image-set(
+    '/images/some-image-1x.jpg' 1x,
+    '/images/some-image-2x.jpg' 2x,
+    '/images/some-image-3x.jpg' 3x
+  );
+}
+```
+
+```css
+a {
+  background-image: YOUR-image-set(
+    '/images/some-image-1x.jpg' 1x,
+    '/images/some-image-2x.jpg' 2x,
+    '/images/some-image-3x.jpg' 3x
+  );
+}
+```
--- a/lib/rules/unit-no-unknown/README.md
+++ b/lib/rules/unit-no-unknown/README.md
@@ -33,19 +33,19 @@ The following patterns are *not* considered violations:
 ```css
 a {
   width: 10px;
-}  
+}
 ```
 
 ```css
 a {
   width: 10Px;
-}  
+}
 ```
 
 ```css
 a {
   width: 10pX;
-}  
+}
 ```
 
 ```css
@@ -116,7 +116,7 @@ a {
 
 ```css
 a {
-  background-image: YOUR-image-set(
+  background-image: YoUr-image-set(
     '/images/some-image-1x.jpg' 1x,
     '/images/some-image-2x.jpg' 2x,
     '/images/some-image-3x.jpg' 3x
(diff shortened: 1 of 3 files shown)


lib/rules/unit-no-unknown/__tests__/index.js
```diff
@@ -472,3 +472,30 @@ testRule(rule, {
...
+      code:
+        "a { background-image: image-set('/images/some-image-1x.jpg' 1x,'/images/some-image-2x.jpg' 2x,'/images/some-image-3x.jpg' 3x,); }"
+    },
+    {
+      code:
+        "a { background-image: my-image-set('/images/some-image-1x.jpg' 1x,'/images/some-image-2x.jpg' 2x,'/images/some-image-3x.jpg' 3x,); }"
+    },
+    {
+      code:
+        "a { background-image: YOUR-image-set('/images/some-image-1x.jpg' 1x,'/images/some-image-2x.jpg' 2x,'/images/some-image-3x.jpg' 3x,); }"
```
jeddy3:
```suggestion
        "a { background-image: YoUr-image-set('/images/some-image-1x.jpg' 1x,'/images/some-image-2x.jpg' 2x,'/images/some-image-3x.jpg' 3x,); }"
```

lib/rules/unit-no-unknown/index.js
```diff
@@ -42,9 +43,15 @@ const rule = function(actual, options) {
...
         // Ignore wrong units within `url` function
+        // and within functions listed in the `ignoreFunctions` option
         if (
           valueNode.type === "function" &&
-          valueNode.value.toLowerCase() === "url"
+          (valueNode.value.toLowerCase() === "url" ||
+            optionsMatches(
+              options,
+              "ignoreFunctions",
+              valueNode.value.toLowerCase()
```
jeddy3:
```suggestion
              valueNode.value
```

I don't think we want to lowercase this as it makes the case insensitivity flag (`i`) redundant.

lib/rules/unit-no-unknown/README.md
```diff
@@ -83,3 +83,43 @@ a {
...
+    '/images/some-image-1x.jpg' 1x,
+    '/images/some-image-2x.jpg' 2x,
+    '/images/some-image-3x.jpg' 3x
+  );
+}
+```
+
+```css
+a {
+  background-image: YOUR-image-set(
```
jeddy3:
```suggestion
  background-image: YoUr-image-set(
```

hudochenkov:
> -   Added: `ignoreFunctions:[]` to `unit-no-unkown` ([#3736](https://github.com/stylelint/stylelint/pull/3736)).

PR #3758, opened 2018-10-29
Add ignoreTypes option to selector-type-case

### LINKED ISSUE #3739: Add ignoreTypes: [] to selector-type-case

It's a very rare use case and it might not be the most reasonable reason to add this option, but here it goes. I'm using mixins (which are not standard) and with them, I have the ability to apply custom properties as selectors, like this (I'm using SugarSS syntax):

```sss
@define-mixin focus $parentClass, $element
    $parentClass > $element
      color: red
```

This code works great, but I have a problem with `$parentClass` property, which is camel case. Therefore I get an error of `selector-type-case`.

I had to add `ignoreTypes: ['/\$\w*/']` to `selector-type-no-unknown` to make sure I prevent those errors so I would like to be able to apply same ignore logic here.

I would like to add an ignore option as an array where you can pass regExp or string with selectors to ignore. It would be similar to `ignoreTypes` in `selector-type-no-unknown` rule.

Currently, I use comments to disable this eslint rule which is fine, but it would be easier to have ignore option. If this request sounds reasonable I could come up with some PR.

Bilie:
Closes #3739

No, it's self explanatory.

### DIFF
--- a/lib/rules/selector-type-case/README.md
+++ b/lib/rules/selector-type-case/README.md
@@ -57,3 +57,25 @@ A {}
 ```css
 LI {}
 ```
+
+## Optional secondary options
+
+### `ignoreTypes: ["/regex/", "non-regex"]`
+
+Given:
+
+```js
+["$childClass", "/(p|P)arent.*/"]
+```
+
+The following patterns are not considered violations:
+
+```css
+myParentClass {
+  color: pink;
+}
+
+$childClass {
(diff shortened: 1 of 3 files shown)


jeddy3:
> -   Added: `ignoreTypes:[]` to `selector-type-case` ([#3758](https://github.com/stylelint/stylelint/pull/3758)).

PR #3759, opened 2018-10-30
Fix false positives for Less variables and mixins in at-rule-* (#3749)

### LINKED ISSUE #3749: Fix false positives for Less variables and mixins in at-rule-*
When linting a Less file, there are error since you deployed version 9.7.0.

For instance this basic Less file:
```less
@my-variable: #f7f8f9;
@anotherVariable: 1px;

.some-mixin() {
    margin: 0;
}

span {
    .some-mixin();
    background-color: @my-variable;
    margin-left: @anotherVariable;
}

```

would give:

```shell 
 1:1  ✖  Unexpected unknown at-rule "@my-variable"            at-rule-no-unknown
 2:1  ✖  Expected "anotherVariable" to be "anothervariable"   at-rule-name-case 
 2:1  ✖  Unexpected unknown at-rule "@anotherVariable"        at-rule-no-unknown
 9:5  ✖  Unexpected unknown at-rule "@some-mixin"             at-rule-no-unknown
```

This is due to moving to postcss-less 3.0.1 in version 9.7.0. They now generate AtRule nodes for mixins and variables (with respectively `mixin: true` and `variable: true` properties), I suppose we should handle these cases in all at-rule-* rules.

glutentag:

Closes #3749

No, it's self explanatory.

### DIFF
--- a/lib/rules/at-rule-empty-line-before/__tests__/index.js
+++ b/lib/rules/at-rule-empty-line-before/__tests__/index.js
@@ -260,7 +260,7 @@ testRule(rule, {
     },
     {
       code: "@import 'test'; @include mixin(1) { @content; };",
-      fixed: "@import 'test';\n\n @include mixin(1) {\n\n @content; };",
+      fixed: "@import 'test';\n\n @include mixin(1) { @content; };",
       message: messages.expected
     }
   ]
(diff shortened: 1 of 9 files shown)


lib/rules/at-rule-empty-line-before/__tests__/index.js
```diff
@@ -260,7 +260,7 @@ testRule(rule, {
     },
     {
       code: "@import 'test'; @include mixin(1) { @content; };",
-      fixed: "@import 'test';\n\n @include mixin(1) {\n\n @content; };",
```
jeddy3: This changed is due to this [bit in the](https://github.com/stylelint/stylelint/blob/master/lib/utils/isStandardSyntaxAtRule.js#L11-L14) `isStandardSyntaxAtRule` util. I'd consider the existing behaviour a bug and the changes in this PR fix it, even though it's unrelated to Less Syntax.

jeddy3:
> -   Fixed: `at-rule-*` false positives for Less variables and mixins ([#3759](https://github.com/stylelint/stylelint/pull/3759)).

pvdlg:
It seems the same problem is still happening for the tule `at-rule-semicolon-newline-after`.

PR #3776, opened 2018-11-02
Remove useless code

yepninja:


lib/rules/function-name-case/index.js
```diff
@@ -45,7 +45,6 @@ const rule = function(expectation, options, context) {
     }
 
     root.walkDecls(decl => {
-      let needFix = false;
```
vankop: there is some performance idea in this variable ( `parsed.toString()` is a bit expensive operation, see [stringify.js](https://github.com/TrySound/postcss-value-parser/blob/master/lib/stringify.js)), for instance when function name is correct, but `context.fix === true`, it is not necessary to perform `parsed.toString()`

vankop:
@yepninja I have wrote my suggestion to improve performance

jeddy3:
It seems like the `let needFix = false;` approach is a valid performance addition. Shall we close this PR?

jeddy3:
Closing this PR. Feel free to reopen if there's more to discuss.

PR #3778, opened 2018-11-06
Discard parse errors from ignored files

ccqgithub:
now, below codes will throw a parse error even the `.stylelintignore` has pattern `/**`

```js

stylelint.lint({
  code: 'var a = {',
  codeFilename: 'test.js',
```
because lintsource throw parse error even the file is ignored, this ignore the errors

### DIFF
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -141,6 +141,19 @@ module.exports = function(
             return stylelint._createStylelintResult(postcssResult);
           });
       })
+      .catch(error => {
+        // if file is ignored, ignore parse errors
+        if (
+          absoluteCodeFilename &&
+          !filterFilePaths(ignorer, [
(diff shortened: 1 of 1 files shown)


jeddy3:
@ccqgithub Thank you for your contribution. 

Is it possible to add a test for this behaviour?

jeddy3:
> This is an edge case as the error

Thanks for digging and for the explanation!

> Not really sure what the outcome of this should be

I guess the addition is a bit of defensive coding. However, as you said, this is an edge-case. Happy to take your recommendation on whether we should merge or not (after the tests are in) as you're one of the active contributors who knows the engine well.

CAYdenberg:
Alright, @ccqgithub if you merge or cherry-pick the commit above I will approve this PR.

ccqgithub:
i pull a new PR for this, which is better..

PR #3781, opened 2018-11-06
Jest enhancements

hudochenkov:
1\. Added much needed [jest-watch-typeahead](https://github.com/jest-community/jest-watch-typeahead). Very useful with our amount of tests.

~~2\. When developing locally we most likely use `npm run watch` (`jest --watch`), rather running full test suite all the time. With https://github.com/stylelint/stylelint/commit/67ad0e750e6bc1ad59d9d6ddbfbd650702d3a481 we can use `npm test` locally to run watch mode, but on CI it will run all tests and linters we have. What do you think, @stylelint/contributors?~~

P. S. Little known fact: `npm test` has an alias `npm t` ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ cache:
 matrix:
   include:
     - node_js: '8'
-      script: npm run pretest
+      script: npm run static-analysis
       env: CI=pretest
     - node_js: '8'
       script: npm run jest -- --maxWorkers=2 --coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ cache:
 matrix:
   include:
     - node_js: '8'
-      script: npm run static-analysis
+      script: npm run pretest
       env: CI=pretest
     - node_js: '8'
       script: npm run jest -- --maxWorkers=2 --coverage
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
     "flow-typed": "^2.5.1",
     "husky": "^1.1.0",
     "jest": "^23.4.1",
+    "jest-watch-typeahead": "^0.2.0",
     "lint-staged": "^8.0.0",
     "npm-run-all": "^4.0.2",
     "npmpub": "^4.1.0",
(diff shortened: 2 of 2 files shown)


jeddy3:
> 1. Added much-needed jest-watch-typeahead.

Nice!

> we can use npm test locally to run watch mode, but on CI it will run all tests and linters we have. What do you think, @stylelint/contributors?

Do we know what the most idiomatic approach is? Historically I've encountered `npm test` for running the test suite through and `npm run watch` for watching. This is what we [document](ttps://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#running-tests) in the developer guide. I often run `npm t` locally before I push up a PR to check that everything (e.g. tests _, linting and flow_) is hunky-dory. Likewise, I use `npm t` locally whilst preparing for releases. I think it might feel odd:

1. having the same command do something different locally than on the CI
2. having to remember to run `static-analysis`

Having said all that, if the behaviour of `npm test` in this PR is more idiomatic then I think we should adopt it.

hudochenkov:
I should've explained better :)

The rationale behind this change is: developer run watch mode more often than a full test suite. Which is why it makes sense to type more frequent command faster.

I'm having second thoughts now about this change. Hiding full test suite behind `npm run test:full` will most likely be “harmful” for contributors, because they don't know about it, so they might be sure tests are good in watch mode, while linters fail.

I'll revert this change, because it doesn't make sense for stylelint particularly.

jeddy3:
> I'll revert this change, because it doesn't make sense for stylelint particularly.

Groovy.

Great to see typeahead in, though!

ntwb:
Hmmmm... Am I missing something here? When using `npm run watch`, it first runs the full suite of tests, then as I change or update a file or test, only that test runs. As such whilst developing I'd rarely see the need to type in a filename or testname 🤔

hudochenkov:
`npm run watch` runs files related to modified status in `git status` by default. Sometimes its many files, if some utility changed. While working on something, a developer might modify many files, so many tests will run after every file save. Developer most likely wants to focus on one test file only, when developing, so that when a filter is needed. With typeahead, it's much easier to do. I use it all the time in any project I'm working on.

ntwb:
Thanks for that @hudochenkov, makes sense 👍

PR #3794, opened 2018-11-09
Fix system font values on font property

### LINKED ISSUE #3793: Add support of system font values on font property
The rule `font-family-no-missing-generic-family-keyword` returns false positives when [system font values](https://developer.mozilla.org/en-US/docs/Web/CSS/font#System_font_values) are used in `font` property.

```css
input {
  font: caption;
}
```

```json
{
  "rules": {
    "font-family-no-missing-generic-family-keyword": true
  }
}
```

e.g. `9.2.0`

e.g. "CLI with `stylelint "**/*.css" --config myconfig.json`"

e.g. "Yes, it's related to SCSS nested properties."

e.g. "No warnings to be flagged."

```shell
test.css
2:4    Expected a leading zero (number-leading-zero)
```

regseb:

Closes #3793

No, it's self explanatory.

### DIFF
--- a/lib/reference/keywordSets.js
+++ b/lib/reference/keywordSets.js
@@ -78,6 +78,15 @@ keywordSets.camelCaseFunctionNames = new Set([
 
 keywordSets.basicKeywords = new Set(["initial", "inherit", "unset"]);
 
+keywordSets.systemFontValues = uniteSets(keywordSets.basicKeywords, [
+  "caption",
+  "icon",
+  "menu",
+  "message-box",
+  "small-caption",
+  "status-bar"
+]);
+
 keywordSets.fontFamilyKeywords = uniteSets(keywordSets.basicKeywords, [
   "serif",
   "sans-serif",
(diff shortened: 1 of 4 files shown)


jeddy3:
Fascinating! Thanks for the sharing the link to the spec. I had no idea there were such things are _system fonts_ and they are set using the `font` shorthand property which is sometimes not a shorthand property!

Here's the relevant bit from the spec for those interested:

> **System fonts may only be set as a whole**; that is, the font family, size, weight, style, etc. are all set at the same time. These values may then be altered individually if desired. If no font with the indicated characteristics exists on a given platform, the user agent should either intelligently substitute (e.g., a smaller version of the 'caption' font might be used for the 'small-caption' font), or substitute a user agent default font. As for regular fonts, if, for a system font, any of the individual properties are not part of the operating system's available user preferences, those properties should be set to their initial values.

> **That is why this property is "almost" a shorthand property: system fonts can only be specified with this property, not with 'font-family' itself**, so 'font' allows authors to do more than the sum of its subproperties. However, the individual properties such as 'font-weight' are still given values taken from the system font, which can be independently varied.

[Emphasis mine]

ntwb:
Chencelog updated in: https://github.com/stylelint/stylelint/commit/384f20167bfa085b6b28410d2082c64fb64a305a

> Fixed: false positives for system fonts in `font-family-no-missing-generic-family-keyword` (#3794).

PR #3799, opened 2018-11-09
Allow ordinary regular expressions anywhere a regex string is accepted in rule config

### LINKED ISSUE #3008: Support regex literals in all options that accept regexes

CAYdenberg:

Closes #3008 

My goal is to do one of these per (working) day until it's done ... I can submit them as separate PRs or just keep a running branch.

### DIFF
--- a/lib/rules/at-rule-no-unknown/README.md
+++ b/lib/rules/at-rule-no-unknown/README.md
@@ -42,7 +42,7 @@ The following patterns are *not* considered violations:
 
 ## Optional secondary options
 
-### `ignoreAtRules: ["/regex/", "string"]`
+### `ignoreAtRules: ["/regex/", /regex/, "string"]`
 
 Given:
 
--- a/lib/rules/at-rule-no-unknown/__tests__/index.js
+++ b/lib/rules/at-rule-no-unknown/__tests__/index.js
@@ -160,6 +160,28 @@ testRule(rule, {
   ]
 });
 
+
+testRule(rule, {
+  ruleName,
+  config: [true, { ignoreAtRules: [/^my-/] }],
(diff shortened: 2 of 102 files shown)


CAYdenberg:
This is ready for a review if anyone can manage it.

Mostly, this is pretty formulaic, so if you look at one commit you've seen them all. The only exception is:
https://github.com/stylelint/stylelint/pull/3799/commits/20ad77244b648df4cbf25ae9804124d00c7687dc

where I had to change the util validateObjectWithStringArrayProps.

CAYdenberg:
Should probably tag this @stylelint/core 

Please review.

CAYdenberg:
Thanks everyone! ❤️ 

I still can't edit the CHANGELOG directly, so if someone could do that for me, it'd be swell.

jeddy3:
> I still can't edit the CHANGELOG directly, so if someone could do that for me, it'd be swell.

Strange... done:

-   Added: support for ordinary regular expressions anywhere a regex string is accepted in rule config ([#3799](https://github.com/stylelint/stylelint/pull/3799)).

@CAYdenberg Thanks again for all the work you put into this PR!

PR #3800, opened 2018-11-10
just to ignore the `ignored files` before lint, to avoid parse errors…

ccqgithub:
… in these files.

A replacement for https://github.com/stylelint/stylelint/pull/3778.

now, below codes will throw a parse error even the .stylelintignore has pattern /**

stylelint.lint({
  code: 'var a = {',
  codeFilename: 'test.js',

because lintsource throw parse error even the file is ignored.

this PR ignore files before lint, so avoid that errors.


ccqgithub:
lost test

PR #3801, opened 2018-11-10
Don't parse ignored files

ccqgithub:
… in these files.

A replacement for https://github.com/stylelint/stylelint/pull/3778.

now, below codes will throw a parse error even the .stylelintignore has pattern /**

stylelint.lint({
  code: 'var a = {',
  codeFilename: 'test.js',

because lintsource throw parse error even the file is ignored.

this PR ignore files before lint, so avoid that errors.

### DIFF
--- a/lib/__tests__/stylelintignore-test/.stylelintignore
+++ b/lib/__tests__/stylelintignore-test/.stylelintignore
@@ -2,3 +2,5 @@
 
 fixtures/*.css
 !fixtures/invalid-hex.css
+
+badsyntax.css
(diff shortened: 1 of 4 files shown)


CAYdenberg:
This looks ... good. Don't know why we weren't already doing this :+1:

jeddy3:
@CAYdenberg Thanks for the review. I assume this is a patch and the following changelog entry will suffice?:

> -   Fixed: ignored files are no longer parsed ([#3801](https://github.com/stylelint/stylelint/pull/3801)).

CAYdenberg:
Yeah that looks good

ccqgithub:
@jeddy3 When can publish a version that include this? The issue throw errors when i use stylelint with vscode 😸

jeddy3:
We can do a release today https://github.com/stylelint/stylelint/issues/3826.

ccqgithub:
Oh, very thanks...

Issue #3729, opened 2018-10-19
Fix false positives for flush comments containing a comma in selector-max-universal


ssivanatarajan:
There was a false positive for the selector's list with a commented selector in selector-max-universal rule

> Which rule, if any, is the bug related to?

`selector-max-universal`

> What CSS is needed to reproduce the bug?

e.g.

 ```
.newbackbtn,/* .searchall, */.calNav{
	-webkit-transform: rotateY(180deg);
	-moz-transform:rotateY(180deg);
	-ms-transform:rotateY(180deg);
	-o-transform:rotateY(180deg);
}
```

> What stylelint configuration is needed to reproduce the bug?

e.g.

```json
{
  "rules": {
    
  "selector-max-universal":0
  }
}
```

> Which version of stylelint are you using?

9.2.1

> How are you running stylelint: CLI, PostCSS plugin, Node API?

Node API

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

NO,

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors did you get)?

The following warnings were flagged:

```shell
Expected "*/.calNav" to have no more than 0 universal selectors (selector-max-universal)
```

jeddy3:
@ssivanatarajan Thanks for the report and for using the template.

This is an interesting bug, and is likely within the selector parser.

It seems it is only triggered when:

- there is a comma within the comment
- the comment is flush to the surrounding characters

For example, this will also trigger the bug:

```css
flush/* , */flush {}
```

```
1:11 error Expected "*/flush" to have no more than 0 universal selectors (selector-max-universal)
```

There's a pending PR, https://github.com/stylelint/stylelint/pull/3284, to update the selector parser to the latest version. I suggest we wait for that to be merged to see if it resolves the issue. In the meanwhile, I believe both the following will workaround the issue:

- remove the comma from within the comment
- adding a space either side of the comment

sswebcoder:
I think what postcss not support comment as part of selector and as result we got selector with comment and with comma. After split selector by comma we got incorrect selectors.

sswebcoder:
![image](https://user-images.githubusercontent.com/22288830/48395038-def26080-e726-11e8-9d5e-7b3d101e8043.png)

jeddy3:
> I think what postcss not support comment as part of selector and as result we got selector with comment and with comma.

Do you want to raise an issue upstream in PostCSS to query this behaviour?

[...]

PR #3817, opened 2018-11-22
Fix false positives for flush comments containing a comma in selector-max-universal

sswebcoder:
Fix for #3729


hudochenkov:
> -   Fixed: false positives for flush comments containing a comma in `selector-max-universal` ([#3817](https://github.com/stylelint/stylelint/pull/3817)).

PR #3833, opened 2018-12-02
Add function-calc-no-invalid rule

### LINKED ISSUE #3820: Add function-calc-no-invalid

I want to check invalid expressions of the `calc()` function beforehand.

I think that this rule will solve issue #3806 as well.

A new rule that reports an invalid expression of the `calc()` function.

I think that we can check it with `postcss-value-parser`.

:+1: Examples of **correct** css:

```css
.foo {width: calc(100% - 80px);}
```

:-1: Examples of **incorrect** css:

```css
.foo {width: calc();} /* empty expression */
.foo {width: calc(100% 80px);} /* missing operator */
.foo {width: calc(100% -80px);} /* missing operator */
.foo {width: calc(100% - - 80px);} /* unexpected operator */
.foo {width: calc(100% -);} /* unexpected operator */
.foo {width: calc(- 100%);} /* unexpected operator */
.foo {width: calc(100px + 80);} /* addition (`+`) with mixed `number` and `length` */
.foo {width: calc(100% + 80);} /* addition with mixed `number` and `percentage` */
.foo {width: calc(100px - 80);} /* subtraction (`-`) with mixed `number` and `length` */
.foo {width: calc(100px * 80px);} /* multiplication (`*`) by `length` */
.foo {width: calc(100 / 80%);} /* division (`/`) with mixed, and `number` on the left-hand side */
.foo {width: calc(100% / 0);} /* division by zero */
```

---
#### Additional context

https://www.w3.org/TR/css-values-3/#calc-notation
https://developer.mozilla.org/en-US/docs/Web/CSS/calc

ota-meshi:

Closes #3820

This PR add `function-calc-no-invalid` rules.
A new rule that reports an invalid expression of the calc() function.

---

I created `calc` expression parser with reference to [`parser.jison`](https://github.com/postcss/postcss-calc/blob/master/src/parser.jison) of  [postcss-calc](https://github.com/postcss/postcss-calc).
The main changes from `parser.jison` of `postcss-calc` are below.
- Changed to case insensitive.
- Changed to accept an unknown expression.
- Changed to include location in node.
- Added predefined units with stylelint. (e.g. `vm`)
- Changed the calculation process of minus.
- Changed to integrate length units into `LengthValue`.


jeddy3:
Oh wow, this is great!

My original concern was that this rule would not meet the following [criteria for inclusion](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#criteria-for-inclusion):

> - Has a clear and unambiguous finished state.

However, because this rule is using a spec compliant parser, I think it might meet this criteria.

Can anyone think of a reason not to include this rule in core?

ota-meshi:
@evilebottnawi Thank you for review!
I added test cases and fixed bugs.

evilebottnawi:
@ota-meshi let's add some tests with variables `$var`, `@var` and functions `calc(10px + myFunc())` and we can merge

ota-meshi:
I added the test cases.
Please tell me if there is a shortage in the test case.

PR #3872, opened 2019-01-02
Remove `postcss-styled`

gucong3000:


lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
```diff
@@ -282,7 +282,6 @@ testRule(rule, {
   ruleName,
   config: [true],
   skipBasicChecks: true,
-  syntax: "styled",
```
hudochenkov: Why `syntax: "jsx"` isn't required here? `postcss-syntax` recognizes it automatically? Won't it produce false positives, i. e. report no warnings even if there are there?

gucong3000: `postcss-syntax` will automatically infer syntax based on source code

jeddy3:
@gucong3000 I need a little help understanding this PR. 

Has `postcss-styled` been deprecated?

At the moment `"styled"` is one of the [_documented_](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/node-api.md#syntax) options for the `syntax` option of the Node API. As such, some users might be relying on it and I believe that we'll [break their build](https://github.com/stylelint/stylelint/blob/0042994cda79f2f3dffc87fefcd4f17f6623816e/lib/getPostcssResult.js#L84-L88) if we remove it. 

(Incidentally, the error message in that last link needs to be updated...)

gucong3000:
> I believe that we'll break their build if we remove it.

I updated the code. `styled` exists as an alias of `postcss-jsx`

hudochenkov:
I think it would make sense to add deprecation notice for `styled` option. Could be done in another PR.

Also, I think we should rename `jsx` to `css-in-js`. "jsx" a little bit confusing, because it's not for JSX only, but for many CSS-in-JS syntaxes. This is also something for another PR. If we agree, then we can deprecate both `styled` and `jsx` names, and add `css-in-js` all in one PR.

gucong3000:
Should I make any changes to this PR?

jeddy3:
> -   Added: `css-in-js` syntax option that will replace the existing `styled` and `jsx` ones ([#3872](https://github.com/stylelint/stylelint/pull/3872)).

PR #3873, opened 2019-01-03
Disable --fix while using a processor and display message

### LINKED ISSUE #3223: Disable --fix while using a processor and display message

Minor feature request, I'm thinking this could be a `good first issue`. Just thought of this as I was tending to [this sad user of the Styled Components processor](https://github.com/styled-components/styled-components-website/pull/247), the simplest way I can think of avoiding future issues is to just disable `--fix` when a processor is present in the configuration as it is not currently supported (and since a processor being able to use `--fix` would require the processor to always output the same css input it got, the only type of processor I can think of that could use this is one that only changes the stylelint screen output / maybe doing some statistics), if you think that could be a problem an option could be to add (and of course inform the user what happened and the possibilities when `--fix` fails) a `--fix-processor` option or something of the like that overrides the disable.

N/A

N/A

N/A

N/A

N/A

Using processors

N/A

N/A

gucong3000:

Closes #3223

No, it's self explanatory.

### DIFF
--- a/lib/getPostcssResult.js
+++ b/lib/getPostcssResult.js
@@ -108,7 +108,7 @@ module.exports = function(
         if (stylelint._options.fix) {
           // eslint-disable-next-line no-console
           console.warn(
-            "`--fix` cmd flag is not compatible with `processors` in configuration, so it will be disabled."
+            "Autofix is incompatible with processors and will be disabled. Are you sure you need a processor?"
           );
           stylelint._options.fix = false;
         }
--- a/lib/getPostcssResult.js
+++ b/lib/getPostcssResult.js
@@ -104,7 +104,15 @@ module.exports = function(
       const source = options.code ? options.codeFilename : options.filePath;
       let preProcessedCode = code;
 
-      if (options.codeProcessors) {
+      if (options.codeProcessors && options.codeProcessors.length) {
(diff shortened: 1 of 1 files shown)


lib/getPostcssResult.js
```diff
@@ -104,7 +104,15 @@ module.exports = function(
       const source = options.code ? options.codeFilename : options.filePath;
       let preProcessedCode = code;
 
-      if (options.codeProcessors) {
+      if (options.codeProcessors && options.codeProcessors.length) {
+        if (stylelint._options.fix) {
+          // eslint-disable-next-line no-console
+          console.warn(
+            "`--fix` cmd flag is not compatible with `processors` in configuration, so it will be disabled."
```
jeddy3: Shall we say?:

```suggestion
            "Autofix is incompatible with processors and will be disabled. Are you sure you need a processor?"
```

As `fix` can also be enabled via the Node API. The "Are you sure..." bit is because I suspect a lot of users might be using processors needlessly for things like `<style>` tags.

hudochenkov:
> -   Changed: Autofixing will be disabled if it is used with stylelint processors ([#3873](https://github.com/stylelint/stylelint/pull/3873)).

I've mention “stylelint processors” to avoid confusion with preprocessors like Sass.

Issue #3863, opened 2018-12-22
Fix false positives for styled-components in no-duplicate-selectors


hudochenkov:
> Clearly describe the bug

False positives for `no-duplicate-selectors` in Styled Components. The rule looks into two different components which is not correct. Every styled component should be treated as a separate file if we speak in CSS terms. `Left` and `Right` are two different components, and stylelint can't know if their rules collapse or not.

This is most likely has to do with how `postcss-styled` parses a file.

Related issue #3862.

Test repository: https://github.com/hudochenkov/styled-component-stylelint-bugs/tree/master/false-positive-no-duplicate-selectors

> Which rule, if any, is the bug related to?

`no-duplicate-selectors`

> What CSS is needed to reproduce the bug?

```js
import styled from 'styled-components';
import { ButtonStyled } from '../../button/Button.styled';

export const Left = styled.div`
  a {
    color: red;
  }

  ${ButtonStyled} {
    margin-right: 8px;
  }
`;

export const Right = styled.div`
  a {
    color: blue;
  }

  ${ButtonStyled} {
    margin-left: 8px;
  }
`;
```

> What stylelint configuration is needed to reproduce the bug?

```json
{
    "extends": ["stylelint-config-recommended", "stylelint-config-styled-components"]
}
```

> Which version of stylelint are you using?

`9.9.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

CLI with `stylelint "**/*.styled.js" --syntax styled`

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

Yes, it's related to styled components.

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors did you get)?

The following warnings were flagged:

```shell
Header.styled.js
 15:3  ✖  Unexpected duplicate selector "a", first used at line 5                 no-duplicate-selectors
 19:3  ✖  Unexpected duplicate selector "${ButtonStyled}", first used at line 9   no-duplicate-selectors
```

jeddy3:
> Every styled component should be treated as a separate file if we speak in CSS terms

Interesting. I hadn't really thought of it like that. I recently [answered a SO question](https://stackoverflow.com/questions/53499725/no-duplicate-selectors-error-for-different-selectors-inside-the-same-file) saying that the current behaviour was intended, but now I'm questioning that.

I think it does make sense if postcss-styled parses each component definition as a source.

gucong3000:
[...]

PR #3875, opened 2019-01-03
Run rules for each root of document

### LINKED ISSUE #3862: Fix false positives for styled-components in no-descending-specificity
> Clearly describe the bug

False positives for `no-descending-specificity` in Styled Components. The rule looks into two different components which is not correct. Every styled component should be treated as a separate file if we speak in CSS terms. `Link` and `LinkIcon` are two different components, and stylelint can't know if their rules collapse or not.

This is most likely has to do with how `postcss-styled` parses a file.

Test repository: https://github.com/hudochenkov/styled-component-stylelint-bugs/tree/master/false-positive-no-descending-specificity

`no-descending-specificity`

```js
import styled from 'styled-components';
import { NavLink } from 'react-router-dom';

export const Link = styled(NavLink)`
  font-size: 15px;
  font-weight: 700;

  &:hover,
  &.active {
    color: red;

    svg {
      fill: red;
    }
  }
`;

export const LinkIcon = styled.div`
  flex: none;

  svg {
    width: 15px;
    height: 15px;
  }
`;

```

```json
{
    "extends": ["stylelint-config-recommended", "stylelint-config-styled-components"]
}
```

`9.9.0`

CLI with `stylelint "**/*.styled.js" --syntax styled`

Yes, it's related to styled components.

No warnings to be flagged.

```shell
Sidebar.styled.js
 21:3  ✖  Expected selector "svg" to come before selector "&:hover svg"    no-descending-specificity
 21:3  ✖  Expected selector "svg" to come before selector "&.active svg"   no-descending-specificity
```

gucong3000:

Closes #3862 #3863

"No, it's self explanatory."


jeddy3:
> -   Fixed: `no-descending-specificity` false positives for styled-components ([#3875](https://github.com/stylelint/stylelint/pull/3875)).
> -   Fixed: `no-duplicate-selectors` false positives for styled-components ([#3875](https://github.com/stylelint/stylelint/pull/3875)).

vladfaust:
Thank you! :heart:

lib/rules/no-empty-source/index.js
```diff
@@ -11,25 +10,25 @@ const messages = ruleMessages(ruleName, {
...
-        return;
-      }
-
-      report({
-        message: messages.rejected,
-        node: root,
-        result,
-        ruleName
-      });
+    const rootString = context.fix ? root.toString() : root.source.input.css;
```
Mouvedia, line 21: @jeddy3 do you remember why you had to use `context.fix` here?

Issue #3886, opened 2019-01-13
Fix false positives for focus-visible in selector-pseudo-class-no-unknown


mo:
Running stylelint 9.9.0 using stylelint-config-recommended on the following file:

```css
:focus-visible {
	color: red;
}
```

...reports the error:

```shell
test.css
 1:1  ✖  Unexpected unknown pseudo-class selector ":focus-visible"   selector-pseudo-class-no-unknown
```

:focus-visible is a new pseudo class coming in CSS Selectors v4.

It's currently implemented behind a prefix in Firefox and behind a flag in Chrome, see status here:
https://caniuse.com/#search=focus-visible

Chrome implementation is tracked by this bug:
https://bugs.chromium.org/p/chromium/issues/detail?id=817199&desc=3

Some people are already using it via this polyfill:
https://github.com/WICG/focus-visible

It would be nice if stylelint did not report this pseudo class as "unknown".

jeddy3:
@mo Thanks for the report.

Yes, this should be considered known. For those interested, it's in the [Level 4 Selector Working Draft](https://www.w3.org/TR/selectors-4/#the-focus-visible-pseudo). 

It should just be the case of adding it to [this list](https://github.com/stylelint/stylelint/blob/master/lib/reference/keywordSets.js#L235-L237), if you'd like a submit a PR.

mo:
@jeddy3 If I also wanted to write a testcase, in which file should I put it?

jeddy3:
[Each rule](https://github.com/stylelint/stylelint/tree/master/lib/rules) has it own test folder within.

You should just need to add a `:focus-visible {}` item to the `accept` array around [this line](https://github.com/stylelint/stylelint/blob/master/lib/rules/selector-pseudo-class-no-unknown/__tests__/index.js#L23) of the `selector-pseudo-class-no-unknown` rule.

PR #3887, opened 2019-01-13
Fix false positives for focus-visible in selector-pseudo-class-no-unknown

mo:
There is one known issue with this PR: the test passes even if the fix itself is reverted. Any ideas why?

Fixes: https://github.com/stylelint/stylelint/issues/3886


jeddy3:
Changelog:

> -   Fixed: `selector-pseudo-class-no-unknown` false positives for `focus-visible` ([#3887](https://github.com/stylelint/stylelint/pull/3887)).

PR #3899, opened 2019-01-16
Synchronize the same parts in both `node-api.md` & `postcss-plugin.md`

gucong3000:
Synchronize the same parts in both `node-api.md` & `postcss-plugin.md`

No, it's self explanatory.


gucong3000:
image
This PR will keep the common parts of the options consistent in the two documents.

Issue #3779, opened 2018-11-06
Release 9.8.0


jeddy3:
- [ ] stylelint release
- [ ] stylelint-demo update
- [ ] stylelint.io update
- [ ] tweet

I should have time to put out a release this evening.

It'd be good to get reviews on https://github.com/stylelint/stylelint/pull/3748 and https://github.com/stylelint/stylelint/pull/3775 before I do.

cc / @stylelint/core

jeddy3:
Even though `master` looks good on Travis, I'm getting failing tests locally:

```
Summary of all failing tests
 FAIL  lib/__tests__/standalone-cache.test.js
  ● standalone cache › cache file is created at $CWD/.stylelintcache

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      66 |       .then(cacheFile => {
      67 |         // Ensure cache file contains only linted css file
    > 68 |         expect(typeof cacheFile[validFile] === "object").toBe(true);
         |                                                          ^
      69 |         expect(typeof cacheFile[newFileDest] === "undefined").toBe(true);
      70 |       });
      71 |   });

      at toBe (lib/__tests__/standalone-cache.test.js:68:58)
      at tryCatch (node_modules/es6-promise/dist/lib/es6-promise/-internal.js:195:12)
      at invokeCallback (node_modules/es6-promise/dist/lib/es6-promise/-internal.js:210:13)
      at publish (node_modules/es6-promise/dist/lib/es6-promise/-internal.js:178:7)
      at flush (node_modules/es6-promise/dist/lib/es6-promise/asap.js:94:5)

  ● standalone cache › only changed files are linted

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

       97 |       })
       98 |       .then(cachedFiles => {
    >  99 |         expect(typeof cachedFiles[validFile] === "object").toBe(true);
          |                                                            ^
      100 |         expect(typeof cachedFiles[newFileDest] === "object").toBe(true);
      101 |       });
      102 |   });

      at toBe (lib/__tests__/standalone-cache.test.js:99:60)

  ● standalone cache › invalid files are not cached

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      152 |       })
      153 |       .then(cachedFiles => {
    > 154 |         expect(typeof cachedFiles[validFile] === "object").toBe(true);
          |                                                            ^
[...]

PR #3903, opened 2019-01-17
Update file-entry-cache to avoid the deprecated warning

### LINKED ISSUE #3786: Refactor standalone-cache.test.js
In [standalone-cache.test.js](https://github.com/stylelint/stylelint/blob/master/lib/__tests__/standalone-cache.test.js) we rely on an implementation detail of a dependency of a dependency of a dependency, which leads to problems like this one: https://github.com/stylelint/stylelint/issues/3779#issuecomment-436687938.

We should not do this.

Related issues:

https://github.com/eslint/eslint/issues/11062
https://github.com/royriojas/flat-cache/issues/22

Teamop:

Closes #3877

Update `file-entry-cache` to avoid the deprecated warning

Just as the #3786 and #3779, the only thing to do is update the `standalone-cache.test.js`. Just use the `parse` method from the `flatted` which is used by `file-entry-cache` to create the cache to parse the content, then tests can pass.

### DIFF
--- a/lib/__tests__/standalone-cache.test.js
+++ b/lib/__tests__/standalone-cache.test.js
@@ -10,6 +10,10 @@ const readFile = pify(fs.readFile);
 const unlink = pify(fs.unlink);
 const cpFile = require("cp-file");
 const fileExists = require("file-exists-promise");
+// As 'file-entry-cache' use 'flatted' to create the cache,
+// need use this package to parse the content
+// eslint-disable-next-line node/no-extraneous-require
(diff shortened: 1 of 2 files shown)


Teamop:
The fail test cases is not related to this change, it's related with the 005 test case

hudochenkov:
@Teamop thank you!

> However, am I right in thinking we're still testing an implementation detail of file-entry-cache? As such, I think we can leave #3786 open.

Correct.

Issue #3910, opened 2019-01-17
Fix "fatal: Not a git repository" error


ductiletoaster:
Details can be found here: https://github.com/GerHobbelt/jison/issues/43

Styleling version: `9.10.0` (9.9.0 seems to be fine)

> How are you running stylelint: CLI, PostCSS plugin, Node API?

As a CLI

> What actually happened (e.g. what warnings or errors did you get)?

Running the following git cmd caused this error.
```
+ git clean -fdx
fatal: Not a git repository: node_modules/@gerhobbelt/recast/../../../.git/modules/lib/modules/js/recast
```

ductiletoaster:
@jeddy3 Correct it was part of our CI process.

jeddy3:
> Correct it was part of our CI process.

Thanks for the clarification.

The problem should be fixed in `9.10.1`, which was just released.

ductiletoaster:
Thank you!

PR #3915, opened 2019-01-18
Remove jison-gho dependency

### LINKED ISSUE #3913: Reduce package size after ballooning to 171MB in 9.10.0
> Clearly describe the bug

The rule introduced in #3833 is neat, but depending on `jison-gho` brings in ~140MB of additional dependencies, bringing the install size of stylelint from 31MB (9.9.0) to 171MB (9.10.0). 😕 Is a ~6X increase in install size justifiable for a single rule?

Would you consider using Jison as a dev dependency and using it to generate a parser as a JS file at build time, rather than using it as a dependency and generating the parser at runtime? And perhaps reverting this rule until then?

`function-calc-no-invalid`

`9.10.0`

You can compare the install sizes like this:

```
mkdir old && cd old && npm i stylelint@9.9.0 && du -sh node_modules
mkdir new && cd new && npm i stylelint@9.10.0 && du -sh node_modules
```

hudochenkov:

Closes #3913 and #3910.

Using the same strategy as in `postcss-calc` — [compile parser](https://github.com/postcss/postcss-calc/blob/acecba21b4df978527ed5ddaff96d108472f52a4/package.json#L19) and use the compiled version rather runtime.

@jeddy3 about [your request](https://github.com/stylelint/stylelint/issues/3913#issuecomment-455694871) for size-limit: I would rather do it in a separate PR.


jeddy3:
> @jeddy3 about your request for size-limit: I would rather do it in a separate PR.

SGTM.

hudochenkov:
Now we have a new big JS file, which was included in coverage in this PR. I've excluded it.

billyjanitsch:
Thanks! ❤

PR #3921, opened 2019-01-20
Fix false positives for negative numbers in function-calc-no-invalid

### LINKED ISSUE #3912: Fix false positives for negative numbers in function-calc-no-invalid
> Clearly describe the bug

There are false positives for negative numbers inside the ``calc()`` statement. Where inside the ``calc()`` a negative number is used and whether it has a unit or not does not matter.

``function-calc-no-invalid``

```css
a {
  margin-left: calc(-6 * 20px);
}
```

```json
{
  "rules": {
  }
}
```

``9.10.0``

CLI with ``stylelint "source/**/*.css"``

No.

No warnings to be flagged.

The following warning is flagged:

```shell
39:42  ×  Expected a valid expression   function-calc-no-invalid
```

ota-meshi:

Closes #3912


package.json
```diff
@@ -137,14 +137,17 @@
...
+      "*.js": [
+        "prettier --write",
+        "eslint --max-warnings=0 --fix",
+        "git add"
+      ],
+      "*.md": [
+        "remark --quiet --frail"
+      ]
+    },
+    "ignore": ["lib/utils/parseCalcExpression/parser.js"]
```
hudochenkov: This file is ignored in `.prettierignore` and `.eslintignore`.

ota-meshi: It was not ignored if I committed from Sourcetree.
So I made this change.

hudochenkov:
> -   Fixed: `function-calc-no-invalid` false positives for negative numbers ([#3921](https://github.com/stylelint/stylelint/pull/3921)).

PR #3923, opened 2019-01-22
Apply rule in the order defined in `lib/rules/index.js`

### LINKED ISSUE #3853: auto fix for `declaration-colon-newline-after` lead to another issue
> Clearly describe the bug

Some rule ignore `indentation` rule and as result after auto fix I receive error about indentation and should call auto fix second time to fix my files. If I try to use `prettier-stylelint` than I get unfixable indentation error, because stylelint is called once.

`declaration-colon-newline-after`. I think this happen because of PR #3588

> What LESS is needed to reproduce the bug?

```less
:local {
  .root {
    &:active .box {
      background: #e1e1e1;
      box-shadow: 0 -1px 0 0 rgba(0, 0, 0, 0.1), 0 0 0 1px rgba(0, 0, 0, 0.2),
        inset 0 1px 2px 0 rgba(0, 0, 0, 0.1);
    }
  }
}
```

```yaml
extends:
  - stylelint-config-standard
  - stylelint-config-css-modules
```

```
    "stylelint": "^9.9.0",
    "stylelint-config-css-modules": "^1.3.0",
    "stylelint-config-standard": "^18.2.0",
```

CLI with `stylelint "./**/*.less" --fix`

I think No.

Correctly formatted file:
```less
:local {
  .root {
    &:active .box {
      background: #e1e1e1;
      box-shadow:
        0 -1px 0 0 rgba(0, 0, 0, 0.1),
        0 0 0 1px rgba(0, 0, 0, 0.2),
        inset 0 1px 2px 0 rgba(0, 0, 0, 0.1);
    }
  }
}
```

After first auto fix I receive:
```less
:local {
  .root {
    &:active .box {
      background: #e1e1e1;
      box-shadow:
        0 -1px 0 0 rgba(0, 0, 0, 0.1),
 0 0 0 1px rgba(0, 0, 0, 0.2),        /* <--- Here line with indentation error */
        inset 0 1px 2px 0 rgba(0, 0, 0, 0.1);
    }
  }
}
```

gucong3000:

Ref #3853

No, it's self explanatory.


jeddy3:
Interesting! I'm curious about what effect this has on plugins. What order would the following be executed in?

```json
{
  "plugins": ["stylelint-scss"],
  "rules": {
    "scss/dollar-variable-colon-newline-after": "always",
    "indentation": 2,
  }
}
```

gucong3000:
> I'm curious about what effect this has on plugins.

Plugin rules will be insert before built-in rules.

jeddy3:
This is a tricky one as I believe we're moving control away from some users in favour of making it easier for the majority of them.

Considering that neither approach is ideal, and at some point we're probably going to have to revisit how the autofixing works to properly fix https://github.com/stylelint/stylelint/issues/3853 (and https://github.com/stylelint/stylelint/issues/2643), I think we can merge this as it'll help people using `stylelint-config-standard` with autofix.

I've changed `Closes #3853` to `Ref #3853` above as this is more of a workaround than a long term solution.

hudochenkov:
Added to changelog:

> -   Changed: Apply rules in the order defined in `lib/rules/index.js` ([#3923](https://github.com/stylelint/stylelint/pull/3923)).

@jeddy3 I'm not sure this wording is good.

jeddy3:
> @jeddy3 I'm not sure this wording is good.

Sounds ok to me.

PR #3933, opened 2019-01-30
Fix false negatives for `isStandardSyntaxDeclaration`

### LINKED ISSUE #3930: Fix false negatives for some rules when using CSS-in-JS parser
> Clearly describe the bug

Some rules don't work with CSS-in-JS parser

- `property-no-unknown`
- `value-no-vendor-prefix`
- `max-nesting-depth`

(these are just few random ones I've tested, but probably there are more)

```css
a {
  rule: 1;
}
```

```js
const Component = styled.a`
  rule: 1;
`
```

```json
{
  "rules": {
    "property-no-unknown": true
  }
}
```

`9.10.1`

CLI with `stylelint test.{css,js}`

Yes, `css-in-js`

```shell
test.css
 2:3  ×  Unexpected unknown property "rule"   property-no-unknown

test.js
 2:3  ×  Unexpected unknown property "rule"   property-no-unknown
```

```shell
test.css
 2:3  ×  Unexpected unknown property "rule"   property-no-unknown
```

---

This is happening because `postcss-jsx` produces different AST from regular css, which appears to be expected (#3875), making [isStandardSyntaxDeclaration](https://github.com/stylelint/stylelint/blob/master/lib/utils/isStandardSyntaxDeclaration.js) return `false` and `property-no-unknown` skipping the declaration.

Aside from core rules, I've seen it in `stylelint-order` plugin, which uses it's own utility function that also not handles this case. I think there would be other plugins like this. If it's impossible to make this parser generate similar AST, I think it would be useful to document it that plugins also should be tested with it's output.

It also would be good to test all core rules for compatibility with that AST.

gucong3000:

Closes #3930

No, it's self explanatory.

### DIFF
--- a/lib/utils/__tests__/isStandardSyntaxUrl.test.js
+++ b/lib/utils/__tests__/isStandardSyntaxUrl.test.js
@@ -107,16 +107,16 @@ describe("isStandardSyntaxUrl", () => {
     expect(isStandardSyntaxUrl("some/@url")).toBeTruthy();
   });
   it("url with less interpolation without quotes", () => {
-    expect(isStandardSyntaxUrl("@{less-variable}")).toBeTruthy();
+    expect(isStandardSyntaxUrl("@{less-variable}")).toBeFalsy();
   });
   it("url with less interpolation at the start without quotes", () => {
-    expect(isStandardSyntaxUrl("@{less-variable}/path")).toBeTruthy();
+    expect(isStandardSyntaxUrl("@{less-variable}/path")).toBeFalsy();
   });
   it("url with less interpolation in the middle without quotes", () => {
-    expect(isStandardSyntaxUrl("some/@{less-variable}/path")).toBeTruthy();
+    expect(isStandardSyntaxUrl("some/@{less-variable}/path")).toBeFalsy();
   });
   it("url with less interpolation at the end without quotes", () => {
-    expect(isStandardSyntaxUrl("some/@{less-variable}")).toBeTruthy();
+    expect(isStandardSyntaxUrl("some/@{less-variable}")).toBeFalsy();
   });
 
   it("sass interpolation at the start", () => {
--- a/lib/utils/hasTplInterpolation.js
+++ b/lib/utils/hasTplInterpolation.js
@@ -2,11 +2,11 @@
 "use strict";
 
 /**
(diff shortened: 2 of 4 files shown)


lib/utils/hasTplInterpolation.js
```diff
@@ -2,11 +2,11 @@
...
 /**
- * Check whether a string has template literal interpolation
+ * Check whether a string has JS template literal interpolation or HTML-like template
  *
  * @param {string} string
  * @return {boolean} If `true`, a string has template literal interpolation
  */
 module.exports = function(string /*: string*/) /*: boolean*/ {
-  return /\${.+?}/.test(string);
+  return /{.+?}/.test(string);
```
hudochenkov: This matches everything.

lib/utils/__tests__/isStandardSyntaxUrl.test.js
```diff
@@ -107,16 +107,16 @@ describe("isStandardSyntaxUrl", () => {
...
-    expect(isStandardSyntaxUrl("@{less-variable}/path")).toBeTruthy();
+    expect(isStandardSyntaxUrl("@{less-variable}/path")).toBeFalsy();
   });
   it("url with less interpolation in the middle without quotes", () => {
-    expect(isStandardSyntaxUrl("some/@{less-variable}/path")).toBeTruthy();
+    expect(isStandardSyntaxUrl("some/@{less-variable}/path")).toBeFalsy();
   });
   it("url with less interpolation at the end without quotes", () => {
-    expect(isStandardSyntaxUrl("some/@{less-variable}")).toBeTruthy();
+    expect(isStandardSyntaxUrl("some/@{less-variable}")).toBeFalsy();
```
evilebottnawi: Looks like a regression

hudochenkov: Actually, it's a bug fix. Interpolation makes URL non-standard. There are tests for Sass interpolation and they all negative.

evilebottnawi: :+1:

jeddy3:
> -   Fixed: false negatives in declaration-based rules for CSS-in-JS ([#3933](https://github.com/stylelint/stylelint/pull/3933)).

PR #3936, opened 2019-02-04
Add disallowInList option to no-duplicate-selectors

### LINKED ISSUE #3931: Add disallowInList to no-duplicate-selectors
> No-duplicate-selectors doesn't find error, when there're grouping selectors 

`no-duplicate-selectors `

```css
.first-selector, .second-selector {
  background: #eee;
}

.second-selector {
  background: #000;
}
```

e.g.

```json
{
  "rules": {
    "no-duplicate-selectors":  true
  }
}
```

`^9.7.0`

stylelint "**/*.css" --config myconfig.json

No

Warning to be flagged:

```shell
Unexpected duplicate selector ".second-selector", first used at line 22 (no-duplicate-selectors) [no-duplicate-selectors]
```

There's no warning

devsnice:

Closes #3931

### DIFF
--- a/lib/rules/no-duplicate-selectors/README.md
+++ b/lib/rules/no-duplicate-selectors/README.md
@@ -100,3 +100,22 @@ a {
   & c {}
 }
 ```
+
+## Optional secondary options
+
+### `disallowInList: true | false` (default: `false`)
+
+This option will dissallow duplication selectors within a group selectors
+
+For example, with `true`, the following patterns are considered violations:
+
+```css
+input, textarea {
+  border: 2px;
+}
+
+textarea {
+  border: 1px;
+}
+
+```
--- a/lib/rules/no-duplicate-selectors/README.md
+++ b/lib/rules/no-duplicate-selectors/README.md
@@ -107,7 +107,9 @@ a {
 
 This option will also disallow duplicate selectors within selector lists.
 
-For example, with `true`, the following patterns are considered violations:
+For example, with `true`.
+
+The following patterns are considered violations:
 
 ```css
 input, textarea {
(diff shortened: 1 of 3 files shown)


devsnice:
I don't understand why test in ignore.test.js was failed, maybe it's wired with some deps?

hudochenkov:
@devsnice this happens from time to time. We don't know what causes it. I've restarted the job.

lib/rules/no-duplicate-selectors/README.md
```diff
@@ -100,3 +100,22 @@ a {
   & c {}
 }
 ```
+
+
+### `disallowInList: true | false` (default: `false`)
+
+This option will dissallow duplication selectors within a group selectors
```
jeddy3:
```suggestion
This option will also disallow duplicate selectors within selector lists.
```

We like to keep the [vocab](http://apps.workflower.fi/vocabs/css/en) we use consistent. [Selector list](https://www.w3.org/TR/selectors-4/#grouping) comes from the spec.

lib/rules/no-duplicate-selectors/README.md
```diff
@@ -100,3 +100,22 @@ a {
...
+
+For example, with `true`, the following patterns are considered violations:
```
jeddy3:
```suggestion
For example, with `true`.

The following patterns are considered violations:
```

It's a bit clunky, but the sentence needs to split for the color highlighting on the website to work, [for example](https://stylelint.io/user-guide/rules/selector-class-pattern/#optional-secondary-options).

devsnice:
> @devsnice Thanks. It's looking good.
> 
> I've made two suggestions to the README.

Thanks! Can you restart job?

jeddy3:
Changelog:

> -   Added: `disallowInList` to `no-duplicate-selectors` ([#3936](https://github.com/stylelint/stylelint/pull/3936)).

PR #3940, opened 2019-02-05
Refactor standalone-cache.test.js to avoid relying on 'flatted'

### LINKED ISSUE #3786: Refactor standalone-cache.test.js

Teamop:

Closes #3786

Use `file-entry-cache#createFromFile` to create another instance of the cache, so that we can access to the cache without relying on the `flatted` which is the dependency of `file-entry-cache`

### DIFF
--- a/lib/__tests__/standalone-cache.test.js
+++ b/lib/__tests__/standalone-cache.test.js
@@ -6,14 +6,11 @@ const standalone = require("../standalone");
 const fixturesPath = path.join(__dirname, "fixtures");
 const fs = require("fs");
 const pify = require("pify");
-const readFile = pify(fs.readFile);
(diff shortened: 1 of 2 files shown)


billyjanitsch:
Shouldn't `file-entry-cache` have been added to `devDependencies` rather than `dependencies`?

PR #3957, opened 2019-02-19
Fix false negatives for CSS-In-JS in color-no-invalid-hex #3945

### LINKED ISSUE #3945: Fix false negatives for CSS-In-JS in color-no-invalid-hex
Stylelint does not catch CSS errors in JSX, and according to https://github.com/stylelint/stylelint/issues/3264, out-of-the-box JSX support is expected.

```javascript
// Test.jsx
import React from "react";

export default function Test() {
  return <h1 style={{ color: "#FFFFF" }}>Test</h1>; // Invalid hex code
}
```

```javascript
// .stylelintrc
{
  "extends": ["stylelint-config-recommended"]
}
```

```javascript
// package.json
{
  "scripts": {
    "stylelint": "stylelint"
  },
  "dependencies": {
    "react": "16.8.0"
  },
  "devDependencies": {
    "stylelint": "^9.10.1",
    "stylelint-config-standard": "^18.2.0",
  }
}
```

Then, running `yarn stylelint Test.jsx` has the following output:

```
$ stylelint Test.jsx
✨  Done in 0.84s.
```

It is related to JSX syntax.

It should lint the style prop.

No linting errors.

fogost:

Closes #3945

This solves the problem, but I'm not sure about tests.

### DIFF
--- a/lib/rules/color-no-invalid-hex/__tests__/index.js
+++ b/lib/rules/color-no-invalid-hex/__tests__/index.js
@@ -29,9 +29,6 @@ testRule(rule, {
       code: "a { padding: 000; }"
     },
     {
-      code: 'a::before { content: "#ababa"; }'
-    },
-    {
       code: "a { border-#$side: 0; }",
       description: "ignore sass-like interpolation"
     },
--- a/lib/rules/color-no-invalid-hex/__tests__/index.js
+++ b/lib/rules/color-no-invalid-hex/__tests__/index.js
@@ -29,6 +29,9 @@ testRule(rule, {
       code: "a { padding: 000; }"
     },
     {
+      code: 'a::before { content: "#ababa"; }'
+    },
+    {
       code: "a { border-#$side: 0; }",
       description: "ignore sass-like interpolation"
     },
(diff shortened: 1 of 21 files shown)


lib/rules/color-no-invalid-hex/__tests__/index.js
```diff
@@ -28,9 +28,6 @@ testRule(rule, {
     {
       code: "a { padding: 000; }"
     },
-    {
```
jeddy3: This test needs to stay as we need to prove that this rule doesn't check the contents of strings.

lib/rules/color-no-invalid-hex/index.js
```diff
@@ -23,36 +23,39 @@ const rule = function(actual) {
...
 
-      styleSearch({ source: declString, target: "#" }, match => {
-        // If there's not a colon, comma, or whitespace character before, we'll assume this is
-        // not intended to be a hex color, but is instead something like the
-        // hash in a url() argument
-        if (!/[:,\s]/.test(declString[match.startIndex - 1])) {
-          return;
-        }
+      styleSearch(
+        { source: declString, target: "#", strings: "check" },
```
jeddy3: I suspect the way to fix this is to migrate the rule away from `styleSearch` and to `postcss-value-parser`. There wasn't a value parser around when this rule was created, but now we have one and it is proving a more robust way of dealing with the values of declarations.

There an example of the value parser in use in the [color-named rule](https://github.com/stylelint/stylelint/blob/348c139ef1b1c69237a71225d55dfdee73cce39c/lib/rules/color-named/index.js#L73).

e.g.

```js
root.walkDecls(decl => {
  valueParser(decl.value).walk(({value, type, sourceIndex) => {
    if (type !== "word") return;
    const hexMatch = /^#[0-9A-Za-z]+/.exec(value);
    ..
  }
}
```

fogost:
I changed `styleSearch` to `postcss-value-parser` and added css-in-js tests

jeddy3:
> -   Fixed: `color-no-invalid-hex`  false negatives for CSS-in-JS ([#3957](https://github.com/stylelint/stylelint/pull/3957)).

PR #3962, opened 2019-02-21
Run prettier:check as pretest

jeddy3:

Ref: https://github.com/stylelint/stylelint/pull/3957#pullrequestreview-206572436

We should use the CI to make sure everything is pretty-printed.

This will fail until https://github.com/stylelint/stylelint/pull/3957 is in and rebased.


hudochenkov:
@ntwb now we have broken CI :(

jeddy3:
> now we have broken CI :(

A second thumbs up on https://github.com/stylelint/stylelint/pull/3957, and we can get that in which will fix the CI :)

PR #3965, opened 2019-02-25
Throw error on non-existent files unless allow-empty-input is enabled

### LINKED ISSUE #3880: Change behaviour to exit with non-zero status code when no matching files
with version

``` $ stylelint --version
9.9.0
```

running something like

`$ stylelint /path/that/doesnt/exist`

happily return 0 and doesn't complain.
That's a very strange behaviour for me because it doesn't seem too useful to pass a directory that doesn't exist.

Bilie:
Closes #3880


lib/__tests__/ignore.test.js
```diff
@@ -191,8 +191,11 @@ describe("extending config with ignoreFiles glob ignoring one by negation", () =
 });
 
 describe("specified `ignorePath` file ignoring one file", () => {
-  let results;
+  const noFilesErrorMessage = new Error(
+    "The specified `files` glob returns no results"
```
jeddy3: Shall we adopt the ESLint approach of?:

```
No files matching the pattern "${pattern}" were found.
```

This feels a little more explicit as we're stating the glob pattern in the error message, rather than just referring to it.

jeddy3:
With talk of a major release in https://github.com/stylelint/stylelint/issues/4000, I think we can get this in?

lib/__tests__/standalone.test.js
```diff
@@ -114,10 +115,27 @@ it("standalone without input css and file(s) should throw error", () => {
...
+  const files = `${fixturesPath}/non-existent-file.css`;
+  const expectedError = new NoFilesFoundError(files);
+
   return standalone({
-    files: `${fixturesPath}/non-existent-file.css`,
+    files,
     config: configBlockNoEmpty
+  })
+    .then(() => {
+      throw new Error("should not have succeeded");
```
hudochenkov: I'm just curious why to throw an error here? Isn't stylelint should throw an error and `catch` will catch it?

jeddy3:
@hudochenkov Good catches!

jeddy3:
@Bilie Do you think you'll have time to look into @hudochenkov suggestions before the weekend? We're planning to release `10.0.0` then.

lib/__tests__/standalone.test.js
Bilie: Hi, I removed this, it was left over from testing

Bilie:
Hi, thank you for the feedback! I have updated the PR, let me know if you spot anything else.

hudochenkov:
Look awesome! Thanks, @Bilie!

jeddy3:
> -   Changed: throws error if glob matches no files, use the `--allow-empty-input` flag for the old behaviour ([#3965](https://github.com/stylelint/stylelint/pull/3965)).

Issue #3977, opened 2019-03-08
Page margin at-rules not recognised by `at-rule-no-unknown`


spiltcoffee:
> Clearly describe the bug

I'm seeing what I consider to be false positives for page margin at-rules being unknown when using `at-rule-no-unknown`.

> Which rule, if any, is the bug related to?

`at-rule-no-unknown`

> What CSS is needed to reproduce the bug?

```css
@page {
  @top-center {
  }
}
```

> What stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "at-rule-no-unknown": "true"
  }
}
```

> Which version of stylelint are you using?

`9.10.1`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

CLI with `stylelint "**/*.css"`

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

No: https://www.w3.org/TR/css-page-3/#syntax-page-selector

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors did you get)?

The following warnings were flagged:

```shell
test.css
 2:3   ‼  Unexpected unknown at-rule "@top-center"   at-rule-no-unknown
```

spiltcoffee:
GitHub didn't pick it up, but fixed by #3978

PR #3978, opened 2019-03-08
Added test for at-rule-no-unknown for page margin at-rules

spiltcoffee:

#3977. this is a PR with a currently failing test, as suggested I make in the issue template.

No, it's self explanatory.

### DIFF
--- a/lib/reference/keywordSets.js
+++ b/lib/reference/keywordSets.js
@@ -521,8 +521,28 @@ keywordSets.fontShorthandKeywords = uniteSets(
 
 keywordSets.keyframeSelectorKeywords = new Set(["from", "to"]);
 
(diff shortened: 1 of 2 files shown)


hudochenkov:
Thank you!

Could you add missing margin at-rules to https://github.com/stylelint/stylelint/blob/7452f9fbfc43c2ca3cb4f2c723cf16ddc93e6784/lib/reference/keywordSets.js#L525 It will fix the issue.

spiltcoffee:
Done.

Because there's so many page margin at-rules, I put them in their own set and united them back into the main set. Is that okay? If not, I can put them all into the main at-rule set.

hudochenkov:
It's great they are in their own set. Makes it more clear.

PR #3987, opened 2019-03-14
fix: false positives for vendor specific prefixes in `no-descending-specificity` rule

### LINKED ISSUE #3929: Fix false positives for vendor prefixed pseudo-elements in no-descending-specificity
> Clearly describe the bug

There are false positives for the 'no-descending-specificity' rule when an rule selecting a vendor-specific pseudo-element on an element follows a rule selecting that same element with a pseudo-class.

'no-descending-specificity'

```css
input[type=range]:focus {
  color: #ff0000;
}

input[type=range]::-webkit-slider-thumb {
  color: #ff0000;
}
```

```json
{
  "rules": {
  }
}
```

`9.3.0`

Standard CSS, but with a vendor-specific pseudo-element.

No error to be flagged. The description of this rule states:
> There's one other important feature: Selectors targeting pseudo-elements are not considered comparable to similar selectors without the pseudo-element, because they target other elements on the rendered page. For example, a::before {} will not be compared to a:hover {}, because a::before targets a pseudo-element whereas a:hover targets the actual <a>.

The following error was flagged:

```shell
Expected selector "input[type=range]::-webkit-slider-thumb" to come before selector "input[type=range]:focus" (no-descending-specificity)
```

evilebottnawi:

fixes #3929

No, it's self explanatory.

Unfortunately we don't have repo with list of this pseudo classes so we need maintain this manually

### DIFF
--- a/lib/reference/keywordSets.js
+++ b/lib/reference/keywordSets.js
@@ -195,9 +195,35 @@ keywordSets.levelThreePseudoElements = new Set([
   "content"
 ]);
 
+keywordSets.vendorSpecificPseudoElements = new Set([
+  "-moz-progress-bar",
+  "-moz-range-progress",
+  "-moz-range-thumb",
+  "-moz-range-track",
+  "-ms-browse",
+  "-ms-check",
+  "-ms-clear",
+  "-ms-expand",
+  "-ms-fill",
+  "-ms-fill-lower",
+  "-ms-fill-upper",
+  "-ms-reveal",
+  "-ms-thumb",
+  "-ms-ticks-after",
+  "-ms-ticks-before",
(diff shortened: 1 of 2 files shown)


hudochenkov:
Where did you get this list? It would be nice to have link in a comment near this list, so it's easier to check for updates or reference.

ntwb:
Changelog added in https://github.com/stylelint/stylelint/commit/b6b882ad8729f1968ed47c1ae4dde42e549662dc

> Fixed: `no-descending-specificity` false positives for vendor prefixed pseudo-elements ([#3929](https://github.com/stylelint/stylelint/issues/3929)).

PR #4008, opened 2019-04-06
Add ignore: ["comments"] to block-no-empty

### LINKED ISSUE #3925: Add ignore: ["comments"] to block-no-empty
> Clearly describe the bug

If rule `block-no-empty` is enabled, no error is shown in case a block contains comments-only. Of course a comment can be seen as some kind of content but finally the block will do nothing styling-wise which is why it should be considered as empty.

As an alternative an option could be introduced in order to define whether the rule should show an error in case of a completely empty block or already in case of a block with comments-only.

`block-no-empty`

```css
a {
  /* this block with `a` selector should be considered as empty although a comment is there */
}
```

```json
{
  "rules": {
    "block-no-empty": true
  }
}
```

`9.10.0`

CLI with `stylelint "source/**/*.css"`

No.

An error should be shown.

No error is shown.

cahamilton:

Closes #3925. 

Adds secondary option of `ignore: ["comments"]` to rule `block-no-empty`

Not really. If there's any issues with it, let me know :+1:

### DIFF
--- a/lib/rules/block-no-empty/README.md
+++ b/lib/rules/block-no-empty/README.md
@@ -23,15 +23,45 @@ a { }
 ```
 
 ```css
-@media print { a {} }
+@media print {
+  a {}
+}
 ```
 
 The following patterns are *not* considered violations:
 
 ```css
-a { color: pink; }
+a {
+  /* foo */
+}
 ```
 
 ```css
-@media print { a { color: pink; } }
+@media print {
+  a {
+    color: pink;
+  }
+}
+```
+
+## Optional secondary options
+
+### `ignore: ["comments"]`
+
+Exclude comments from being treated as content inside of a block.
+
+The following patterns are considered violations:
+
+```css
+a {
+  /* foo */
+}
+```
+
+```css
+@media print {
+  a {
+    /* foo */
+  }
+}
 ```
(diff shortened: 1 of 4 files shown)


lib/rules/block-no-empty/README.md
```diff
@@ -23,15 +23,87 @@ a { }
...
 ```css
-a { color: pink; }
+a {
+  color: pink;
+}
 ```
 
 ```css
-@media print { a { color: pink; } }
+a {
```
jeddy3: Let's just keep these two comment examples and remove the other new ones:

```css
a {
  /* foo */
}
```

 ```css
@media print {
  a {
    color: pink;
  }
}
```

I think that's enough to communicate that rules with just comments aren't considered empty by default. Also, the built-in rules are geared towards standard CSS syntax and so we tend to avoid using non-standard syntax (e.g. double slash comments) within the rule docs.

File: lib/rules/block-no-empty/README.md [hunk 2]
```diff
@@ -23,15 +23,87 @@ a { }
...
+
+## Optional secondary options
+
+### `ignore: ["comments"]`
+
+Exclude comments from being treated as content inside of a block.
+
+The following patterns are considered violations:
+
+```css
```
jeddy3: As above, let's go with standard CSS in the examples i.e. just keep these two:

```css
a {
  /* foo */
}
```

 ```css
@media print {
  a {
    /* foo */
  }
}
```

cahamilton:
@jeddy3 Those doc changes have been made, let me know of any other changes 👍.

hudochenkov:
> -   Added: `ignore: ["comments"]` to `block-no-empty` ([#4008](https://github.com/stylelint/stylelint/pull/4008)).

PR #4013, opened 2019-04-08
Update dependencies

ntwb:
This PR seeks to update *all* dependencies to the latest version.

Primarily two _dev_  package updates were required:
• Jest from v23 to v24
• Prettier v1.15.1 to v1.16.4

Also to avoid a merge conflict with #4012 updating `flow-bin` from v0.94 to v0.96 I've included it also in this PR which also includes the updated flow-typed definitions for Lodash v4 and the removal of Jest v23 definitions with the addition of Jest v24 definitions.

FYI: Fingers crossed the other version bumps should cause no issues because semver should already be using these on the latest checkouts without issue.


ntwb:
Looking into the single flow error:
```shell
Error ---------------------------------------------------------- lib/standalone.js:182:54
Cannot coerce `JSON.stringify(...)` to string because undefined [1] should not be coerced.
   lib/standalone.js:182:54
   182|     const hashOfConfig = hash(`${stylelintVersion}_${JSON.stringify(config)}`);
                                                             ^^^^^^^^^^^^^^^^^^^^^^
References:
   /tmp/flow/flowlib_3ff2c94c/core.js:491:17
   491|     ): string | void;
                        ^^^^ [1]
Found 1 error
```

ntwb:
https://github.com/facebook/flow/blob/master/Changelog.md#0951

```md
0.95.1

    Added an overload for JSON.stringify allowing mixed input, which returns string | void. Without this, you can't call JSON.stringify on a mixed value at all, because while Flow does allow refining mixed to "not void" (e.g. x === undefined ? undefined : JSON.stringify(x)), it does not support refining mixed to "not a function" (e.g. imagine you could do x === undefined || typeof x == 'function' ? undefined : JSON.stringify(x)). This rolls back some of the more restrictive behavior introduced in v0.95.0, but is still more restrictive and more accurate than in <= v0.94.0.

0.95.0

Likely to cause new Flow errors:

    Disallow undefined and functions in JSON.stringify: JSON.stringify(undefined) returns undefined instead of string. Rather than make it always return string | void, or use overloads to return void on those inputs, we instead disallow those inputs since they are rarely the intended behavior. (#7447)
```

Having still not quite grasped the _flow_ issue and a fix I've added https://github.com/stylelint/stylelint/pull/4013/commits/41c1f5af43b25310b654af37dc54c96a3b40c656 to ignore the issue for now, although ignoring the issue is not ideal Flow [actually recommend](https://medium.com/flow-type/upgrading-flow-codebases-40ef8dd3ccd8) ignoring the issue with a preference to update to the latest version

See also:
> https://github.com/facebook/flow/pull/7447 Update the JSON.stringify
> https://github.com/facebook/flow/issues/7565 Please revert the JSON.stringify definition patch(es) 
> https://twitter.com/fdsmars/status/1092634853378383872


tl;dr: I'm not sure how `JSON.stringify(config)` can be `undefined` in our use case

cc @CAYdenberg

hudochenkov (approved):
LGTM!

alexander-akait (approved):

CAYdenberg:
Replacing the problematic line with:
```
    const hashOfConfig = hash(`${stylelintVersion}_${JSON.stringify(config || '')}`);
```

works as well. Confirmed locally.

PR #4019, opened 2019-04-11
Fix value-keyword-case in scss list

### LINKED ISSUE #3847: Fix autofix for single-line comment within maps in value-keyword-case
> Clearly describe the bug

Single-line comment is replaced with the multi-line. There is the bug in the `value-keyword-case` auto-fixing functionality. This bug was introduced in this PR https://github.com/stylelint/stylelint/pull/3775.

Repository for reproducing: https://github.com/vkrol/stylelint-scss-single-line-comment-bug.

`value-keyword-case`

```scss
$map: (
	key: 0, // some single-line comment
);

```

```js
module.exports = {
	rules: {
		"value-keyword-case": "lower",
	},
};

```

e.g. `9.9.0`

CLI with `stylelint \"**/*.scss\" --fix`

Yes, it's related to SCSS maps.

No changes.

Single-line comment is replaced with the multi-line.

/cc @nazarov-nikita

alexander-schranz:
The problem is that parsed.toString() converts `//` with `/*` I avoid this converting not calling the fixer in a expected case.

Fixes #3847

### DIFF
--- a/lib/rules/value-keyword-case/__tests__/index.js
+++ b/lib/rules/value-keyword-case/__tests__/index.js
@@ -692,6 +692,10 @@ testRule(rule, {
       description: "ignore unit within comments"
     },
     {
+      code: '$list: (\n key: "vAluE", // single quote comment\n );',
+      description: "ignore list single quote comment"
+    },
+    {
       code: "a { $display: block; }"
     },
     {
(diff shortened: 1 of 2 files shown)


jeddy3:
> -   Fixed: `value-keyword-case` autofix for single-line comments within maps ([#4019](https://github.com/stylelint/stylelint/pull/4019)).

PR #4021, opened 2019-04-12
Add max-lines rule

### LINKED ISSUE #3808: Add max-lines

Developers create very large S/CSS files which cannot be maintained

Rule to set an error when the length of a file exceeds a given integer, same concept as tslint:max-file-line-count

cahamilton:

Closes #3808

Not really, just a really basic implementation to enforce max line count. Let me know if there's any issue with it 👍

### DIFF
--- a/lib/rules/max-lines/__tests__/index.js
+++ b/lib/rules/max-lines/__tests__/index.js
@@ -1,0 +1,136 @@
+"use strict";
+
+const rule = require("..");
+const { messages, ruleName } = rule;
+
+testRule(rule, {
+  ruleName,
+  config: [1],
+
+  accept: [
+    {
+      code: ".a { color: red; } .b { color: green; }",
+      description: "count = 1"
+    }
+  ],
+
+  reject: [
+    {
+      code: ".a { color: red; }\r\n.b { color: green; }",
+      description: "count = 2 (CRLF)",
+      message: messages.expected(1, 2)
+    },
+    {
+      code: ".a { color: red; }\n.b { color: green; }",
+      description: "count = 2 (LF)",
+      message: messages.expected(1, 2)
+    }
+  ]
+});
+
+testRule(rule, {
+  ruleName,
+  config: [2],
+
+  accept: [
+    {
+      code: "\r\n",
+      description: "count = 2 (CRLF)"
+    },
+    {
+      code: "\n",
+      description: "count = 2 (LF)"
+    },
+    {
+      code: ".a { color: red; } .b { color: green; }",
+      description: "count = 1"
+    },
+    {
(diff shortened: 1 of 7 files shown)


lib/rules/max-lines/__tests__/index.js
```diff
@@ -0,0 +1,91 @@
...
+      description: "count (3) > max (2)",
+      message: messages.expected(2)
+    },
+    {
+      code: ".a {\ncolor: red;\n}",
+      description: "count (3) > max (2)",
+      message: messages.expected(2)
+    }
+  ]
+});
```
evilebottnawi: Looks very good! Can we add test for `cssinjs` ?

cahamilton: Hey! I'm not overly familiar with `cssinjs`, but I'll take a look into it now!

evilebottnawi: You can find example in other tests

hudochenkov: I don't think it makes much sense to use this rule with CSS-in-JS, because .js could have many imports and logic besides styling parts.

evilebottnawi: @hugomrdias I think it should work well for `css-in-js`, like:

max: 10

```
const Button = styled.a`
  /* This renders the buttons above... Edit me! */
  display: inline-block;
  border-radius: 3px;
  padding: 0.5rem 0;
  margin: 0.5rem 1rem;
  width: 11rem;
  background: transparent;
  color: white;
  border: 2px solid white;

  /* The GitHub button is a primary button
   * edit this to target it specifically! */
  ${props => props.primary && css`
    background: white;
    color: palevioletred;
  `}
`
```

And here we get warning because styles more than 20 lines. eslint max-lines rule for whole js file. stylelint rule only for styles blocks

jeddy3:
Is there any more to do on this or is it ready for review?

cahamilton:
> Is there any more to do on this or is it ready for review?

Just one question that needs clarifying from my standpoint: https://github.com/stylelint/stylelint/commit/f33428b32a3a3e7b043a0091cfbe44be860c4d1c#r33166473

hudochenkov:
> Just one question that needs clarifying from my standpoint: https://github.com/stylelint/stylelint/commit/f33428b32a3a3e7b043a0091cfbe44be860c4d1c#r33166473

I think it's ok to report the whole file, if it's not possible to report on the first node over the limit. All editors shows line numbers and see where it's over the limit is quite easy.

cahamilton:
Hey, sorry for the delay on this one, been fairly busy the last couple weeks. I’ll try take a look again on the weekend 👍

hudochenkov:

PR #4027, opened 2019-04-14
Fix false positives for dollar variables in function-linear-gradient-no-nonstandard-direction

### LINKED ISSUE #4014: Fix false positives for dollar variables in function-linear-gradient-no-nonstandard-direction
> Clearly describe the bug

There are false positives for nonstandard directions when using an SCSS variable with one of the directions (e.g. `top`) in its name.

`function-linear-gradient-no-nonstandard-direction`

```css
.foo {
  background: linear-gradient($purple-top, $purple-bottom);
}
```

```json
{
  "rules": {
    "function-linear-gradient-no-nonstandard-direction": true
  }
}
```

`9.10.1`

CLI with `stylelint --config stylelint.config.js "src/**/*.scss" -f verbose`

Yes, related to SCSS variables.

No errors to be flagged on the usage of `linear-gradient`.

There is an error because of the variable name.

ccabrales:

Closes #4014 

Returns early if the first value is non-standard syntax.

### DIFF
--- a/lib/rules/function-linear-gradient-no-nonstandard-direction/__tests__/index.js
+++ b/lib/rules/function-linear-gradient-no-nonstandard-direction/__tests__/index.js
@@ -266,3 +266,34 @@ testRule(rule, {
     }
   ]
 });
+
+// SCSS tests
+testRule(rule, {
+  ruleName,
+  config: [true],
+  syntax: "scss",
+
+  accept: [
+    {
+      code:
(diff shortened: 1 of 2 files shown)


jeddy3:
> -   Fixed: `function-linear-gradient-no-nonstandard-direction` false positives for dollar variables ([#4027](https://github.com/stylelint/stylelint/pull/4027)).

PR #4029, opened 2019-04-14
Fix allow-empty-input CLI flag

### LINKED ISSUE #4028: Fix allow-empty-input CLI flag
> Clearly describe the bug

Invalid option "`--allow-empty-input`", reports when i run 
`stylelint empty/*.css --allow-empty-input`

`10.0.0`

"CLI with `stylelint empty/*.css --allow-empty-input`"

"No error reported."

```
Invalid option "--allow-empty-input".
```

jeddy3:

Closes #4028

We forgot to add the flag to the available options.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -281,11 +281,15 @@ const meowOptions /*: meowOptionsType*/ = {
 
         Show the currently installed version of stylelint.
 
-      --allow-empty-input
+      --allow-empty-input, --aei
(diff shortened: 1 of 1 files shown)


jeddy3:
> -   Fixed: `--allow-empty-input` CLI flag ([#4029](https://github.com/stylelint/stylelint/pull/4029)).

Issue #4000, opened 2019-03-26
Drop support for Node.js 6.x


ntwb:
Drop support for Node.js 6.x

• Node.js 6.x LTS entered maintenance mode 30 April 2018
• Node.js 6.x becomes end-of-life 30 April 2019

https://github.com/nodejs/Release#release-schedule

Previously: #2996 Drop Node.js 4 support 

As noted in #3991 we are starting to find modules stylelint use are starting to require Node.js 8.x

This will be a breaking change and will require to be shipped in a _major_ version of stylelint, e.g. 10.0

hudochenkov:
Agreed.

jeddy3:
SGTM.

hudochenkov:
Node.js 8 will reach end-of-life pretty soon as well — end of this year. What if minimum support version for stylelint 10 will be Node.js 10? :)

stylelint will most likely work on Node.js 8 as well for a long time, because we almost don't use language feature from Node.js older than 6. And most likely we won't use them. I don't even know what new in Node.js 8 and 10 except `async`/`await` ¯﻿\﻿_﻿(﻿ツ﻿)﻿_﻿/﻿¯

What are you thoughts?

ntwb:
• Node.js 8.x LTS entered maintenance mode 1 January 2019
• Node.js 8.x becomes end-of-life December 2019
(1: The 8.x Maintenance LTS cycle is currently scheduled to expire early on December 31, 2019 to align with the scheduled End-of-Life of OpenSSL-1.0.2.)

I think with [919,893 weekly downloads](https://www.npmjs.com/package/stylelint) we should urge on the side of caution and not remove Node.js 8.x yet, let's wait until later in the year, November or December to drop Node.js 8.x

tim-soft:
How did you guys arrive at a minimum node version of 8.15.1? I'm realizing now that AWS Lambda only supports 8.10, which means many serverless hosting services will not work with StyleLint v10

hudochenkov:
@tim-soft 8.15.1 is the latest 8.x release at the time of release. I believe stylelint 10 will work with Node.js 8.10 without any problems.

tim-soft:
@hudochenkov building for node 8.10 will cause problems with yarn

```bash
error stylelint@10.0.0: The engine "node" is incompatible with this module. Expected version ">=8.15.1". Got "8.10.0"
```

hudochenkov:
[...]

PR #4032, opened 2019-04-15
Drop Node.js required version from 8.15.0 to 8.7.0

ntwb:
Via https://github.com/stylelint/stylelint/issues/4000#issuecomment-483020108

As AWS Labda currenty only supports Node.js 8.10, let's drop the required version down to 8.0.0

We can add this to the next _minor_ release 10.0.1

### DIFF
--- a/docs/developer-guide/prerequisites.md
+++ b/docs/developer-guide/prerequisites.md
@@ -2,7 +2,7 @@
 
 ## Technical
 
--   [Node.js](https://nodejs.org/en/) 8.15.1+.
+-   Minimum [Node.js](https://nodejs.org/en/) v8.0.0+, though we do recommend using the latest 8.x, 10.x, or 11.x Node.js release
 -   Latest [npm](https://www.npmjs.com/).
 
 ## Rules of thumb
--- a/docs/developer-guide/prerequisites.md
+++ b/docs/developer-guide/prerequisites.md
@@ -2,7 +2,7 @@
 
 ## Technical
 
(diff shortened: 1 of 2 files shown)


hudochenkov:
What if we, or our dependency, uses a feature introduced in 8.3.0 for example?

jeddy3:
> What if we, or our dependency, uses a feature introduced in 8.3.0 for example?

I tested locally with `8.0.0` and stylelint crashes out with

```
/Users/jeddy3/Projects/stylelint/node_modules/micromatch/index.js:44
    let isMatch = picomatch(String(patterns[i]), { ...options, onResult }, true);
                                                   ^^^
SyntaxError: Unexpected token ...
```

This is because micromatch uses the object spread operator [which is behind a flag until 8.6.0](https://node.green/#ES2018-features-object-rest-spread-properties).

micromatch has `>=8` in [its package.json](https://github.com/micromatch/micromatch/blob/master/package.json#L32-L34), which gives me the impression that this engine field is a bit problematic.

According to [node.green](https://node.green) the following versions added new features:

- 8.9.4
- 8.6.0
- 8.2.1

Even though stylelint works with 8.6.0 I suggest we go with 8.9.4 as that's the lowest feature-complete version of 8 (and it's compatible with AWS). This seems to me to be the best compromise.

ntwb:
Thanks, makes sense, have bumped it to 8.7.0

jeddy3:
@hudochenkov Did you have any more thoughts on this? I can put out a patch release this morning if this goes in.

jeddy3:
> -   Fixed: minimum Node.js engine reduced to 8.7.0 ([#4032](https://github.com/stylelint/stylelint/pull/4032)).

PR #4046, opened 2019-04-24
Fix false positives for interpolation in function-calc-no-invalid

### LINKED ISSUE #3908: Fix false positives for interpolation in function-calc-no-invalid
> Clearly describe the bug

False positive on `scss` interpolation

e.g. `function-calc-no-invalid`

e.g.

```scss
a {
  width: calc(
    #{$min-value} +
    #{strip-unit($max-value - $min-value)} *
    (100vw - #{$min-vw})
  );
}
```

e.g.

```json
{
  "rules": {
  }
}
```

`9.10.0`

CLI

SCSS

No warnings to be flagged.

```shell
src/shared/styles/mixins/_fluid-type.scss
 4:42  ✖  Expected a valid expression   function-calc-no-invalid
```

ota-meshi:

Closes #3908

No, it's self explanatory.

### DIFF
--- a/lib/utils/parseCalcExpression/parser.jison
+++ b/lib/utils/parseCalcExpression/parser.jison
@@ -50,6 +50,9 @@
 "("                                                   return 'LPAREN';
 ")"                                                   return 'RPAREN';
 
+#\{([\s\S]*?)\}                                       return 'UNKNOWN'; // scss variable
+@\{([\s\S]*?)\}                                       return 'UNKNOWN'; // less variable
+
 \S[^\s()*/+-]*                                        return 'UNKNOWN';
 
 <<EOF>>                                               return 'EOF';
--- a/lib/utils/parseCalcExpression/parser.jison
+++ b/lib/utils/parseCalcExpression/parser.jison
@@ -50,8 +50,8 @@
 "("                                                   return 'LPAREN';
 ")"                                                   return 'RPAREN';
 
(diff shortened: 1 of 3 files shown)


lib/utils/parseCalcExpression/parser.jison
```diff
@@ -50,6 +50,9 @@
 "("                                                   return 'LPAREN';
 ")"                                                   return 'RPAREN';
 
+#\{([\s\S]*?)\}                                       return 'UNKNOWN'; // scss variable
```
evilebottnawi: // scss interpolation

ota-meshi: Thank you for pointing out!
I fixed this.

ota-meshi:
@hudochenkov Thank you for your review!
I added the following test case.

```css
      a {
        width: calc(
          #{$min-value} +
          #{strip-unit($max-value - $min-value)} *
          (100vw - #{$min-vw})
        );
      }
```

hudochenkov:
-   Fixed: `function-calc-no-invalid` false positives for interpolation ([#4046](https://github.com/stylelint/stylelint/pull/4046)).

Bigdragon13th:
When will this be released? Want to turn on the rule, but this issue prevent me.

benjamincharity:
This issue is also a blocker for us 👍

hudochenkov:
@Bigdragon13th @benjamincharity it will be released this week.

PR #4078, opened 2019-05-13
Fix declaration-colon-space-after README

quidcain:
Small documentation fix


ntwb:
Good catch @quidcain, and thanks for the PR 👍 

To other reviewers, this change is for when the rule uses the `never` option of the rule, see:

> https://github.com/stylelint/stylelint/blob/master/lib/rules/declaration-colon-space-after/README.md#never

PR #4072, opened 2019-05-06
Add Node.js 12 to CIs

ybiquitous:
Node.js 12 was released on 2019-04-23, which is the **Current Release** version.
See <https://github.com/nodejs/Release#release-schedule>.

image

> Which issue, if any, is this issue related to?

Related to #3299.

> Is there anything in the PR that needs further explanation?

See [Node.js 12 changelog](https://github.com/nodejs/node/blob/v12.1.0/doc/changelogs/CHANGELOG_V12.md).

PR #4092, opened 2019-05-29
Test also on Node.js 12

DanielRuf:
This PR adds Node.js 12 which is already the latest release.

None.

No, it's self explanatory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,12 +18,15 @@ cache:
 
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


hudochenkov:
Thank you! But I'm going to close this one as a dupe of #4072. Node.js 10 is more stable, so it's more safe to use it as a main engine for tests (linters, coverage, etc). #4072 is following this.

PR #4107, opened 2019-06-07
Revert adding Node.js 12 in AppVeyor

hudochenkov:
Before [the change](https://github.com/stylelint/stylelint/pull/4072) AppVeyor took 5-7 minutes to complete.

https://ci.appveyor.com/project/stylelint/stylelint/builds/24925881
https://ci.appveyor.com/project/stylelint/stylelint/builds/24825609

After change it takes 13-23 minutes:

https://ci.appveyor.com/project/stylelint/stylelint/builds/24976236 13 minutes.
https://ci.appveyor.com/project/stylelint/stylelint/builds/25120885 23 minutes and false fail.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,6 @@ environment:
   matrix:
     - nodejs_version: 8
     - nodejs_version: 10
-    - nodejs_version: 12
 
 clone_depth: 10
 
@@ -16,13 +15,8 @@ branches:
     only:
       - master
 
-platform:
-  - x86
-
 install:
-  # Workaround from https://github.com/appveyor/ci/issues/2921#issuecomment-486727727
-  # https://www.appveyor.com/docs/lang/nodejs-iojs/#installing-any-version-of-nodejs-or-iojs
-  - ps: Update-NodeJsInstallation (Get-NodeJsLatestBuild $env:nodejs_version) $env:PLATFORM
+  - ps: Install-Product node $env:nodejs_version
   - npm install
 
 cache:


hudochenkov:
Windows builds failing. And when they are not — false error about passed tests.

evilebottnawi:
> And when they are not — false error about passed tests.

It is wrong. It is visual bug. When tests failed they exit with code 1 and tests failed. When it is passed they exit with 0 code and tests passed. I can not understand why this is a problem. This is just a visual bug.

evilebottnawi:
I will fix windows bugs and they ready to merge

ntwb:
I’m currently mobile but will check this issue out an the Azure Pipelines issue, I’m not confident of Azure Pipelines current false reporting and would prefer we wait for node.js 12.x to stabilise on AppVeyor before switching to Azure as it stands right now

evilebottnawi:
> refer we wait for node.js 12.x to stabilise on AppVeyor before switching to Azure as it stands right now

Why? :disappointed: AppVeyor always was very bad and looks as hack for windows tests

ntwb:
AppVeyor has served us reliably for many years, at times it has been slow and they added us to a higher performance plan for free, and yes updates have also not been as fast as we would prefer, but right now the stability of AppVeyor without Node.js v12 is preferable than the unreliable and false reporting Azure Pipelines as. a “new” service is offering

evilebottnawi:
> higher performance plan for free

Where? We still not have parallelism so it doesn't have sense, we still wait each build before run previously? 

> yes updates have also not been as fast as we would prefer

Because we don't have alternative for windows.

> right now the stability of AppVeyor without Node.js v12 is preferable than the unreliable and false reporting Azure Pipelines

Wrong, we have right reporting, it is bug in jest because they output invalid ansi code for windows console and will be fixed in next major release.

> Azure Pipelines as. a “new” service is offering IMHO

It is works since 2016/2017, it is not new service.

IMHO AppVeyor is not listen community, all feedback will be ignored, my emails often was unanswered and any bugs in their contains was ignore, this is not too good behavior for developers.

This is not a Microsoft advertisement, they just did it right and it works well.

PR #4110, opened 2019-06-10
Fix crash in block-no-empty

### LINKED ISSUE #4109: `block-no-empty` should allow `@import` statements
> Clearly describe the bug

`block-no-empty` should allow `@import` statements, but it's currently crashing with error "Cannot read property 'every' of undefined"

`block-no-empty`

```css
@import "foo.css";
```

e.g.

```json
{
  "rules": {
    "block-no-empty": [true, {
      "ignore": ["comments"]
    }]
  }
}
```

e.g. `10.1.0`

CLI

No

No crash, no lint error.

```shell
TypeError: Cannot read property 'every' of undefined
    at check (/node_modules/stylelint/lib/rules/block-no-empty/index.js:50:47)
```

ghiculescu:
`block-no-empty` should allow `@import` statements, but it's currently crashing with error "Cannot read property 'every' of undefined"

Closes #4109

### DIFF
--- a/lib/rules/block-no-empty/__tests__/index.js
+++ b/lib/rules/block-no-empty/__tests__/index.js
@@ -133,6 +133,9 @@ testRule(rule, {
     },
     {
       code: "a {\n/* foo */\ncolor: pink;\n}"
+    },
+    {
+      code: '@import "foo.css";'
     }
   ],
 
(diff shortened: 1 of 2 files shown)


hudochenkov:
> -   Fixed: `block-no-empty` crash for `@import` statements ([#4110](https://github.com/stylelint/stylelint/pull/4110)).

PR #2992, opened 2017-10-27
Use write-file-atomic for writing fixed styles to filesystem

ZhangYiJiang:
> Which issue, if any, is this issue related to?

Fixes #2991 

> Is there anything in the PR that needs further explanation?

This makes writing back the linted results when `--fix` is used atomic - previously if the operation was cancelled or crashed while the styles were been written, the sources could be left in an inconsistent state. 

The correct way to write file is to write it to a temp file first, then replace the original. As it turns out the nice folks at npm has written a package to do just that, and hey, if it's good enough for npm then it's probably good enough for Stylelint :wink:

PR #4124, opened 2019-06-19
Replace write-file-atomic with fs.writeFileSync

### LINKED ISSUE #3527: Stylelint crashes when fixing file if it is not owned by current user
When running `--fix` stylelint attempts to [chown its temp file back to the owner of the original file](https://github.com/stylelint/stylelint/blob/07b44802afaf7dfaa06297d358a28e62a3a9fbfd/lib/vendor/writeFileAtomic.js#L130). This fails and causes stylelint to crash when the chown can not be completed. Other linting tools like prettier are able to rewrite files not owned by the current user so I would expect stylelint to be able to as well.

The Error is `Error: EPERM: operation not permitted, chown '/Users/benscott/projects/stylelint-fixfail/style.css.4151106271'`

I discovered this bug because I am running `stylelint --fix` as part of some integration tests for a tool that wraps stylelint and my tests failed on CI as the initial checkout of the files is performed by the `root` user but the tests (and thus `stylelint --fix`) is ran by the `circleci` user.

All autofixable rules.

> What CSS / configuration is needed to reproduce the bug?

In a new folder, create:

A package.json:

```json
{
  "name": "stylelint-fixfail",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "lint": "stylelint '*.css'",
    "lint:fix": "stylelint '*.css' --fix",
    "prettier": "prettier '*.css' --write"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "prettier": "1.14.2",
    "stylelint": "9.4.0"
  },
  "stylelint": {
    "rules": {
      "selector-pseudo-element-colon-notation": "double"
    }
  }
}
```

and a `style.css`:

```css
.foo:before {top: 0;}
```

This css violates the `selector-pseudo-element-colon-notation` rule and autofixing should transform the file contents to be `.foo::before {top: 0;}` (note the double colon).

`9.4.0`

> Reproduction steps

* `npm install`
* Run `npm run lint:fix` and note that style.css is reformatted correctly, then reset the file back to its original content
* `sudo chown root style.css && sudo chmod 666 style.css` to change the ownership of the file to be not-the-current-user but still be writable by all users.
* Run `npm run lint:fix` Get the error message `Error: EPERM: operation not permitted, chown '/Users/benscott/projects/stylelint-fixfail/style.css.4151106271'` and the file is not autoformatted
* Run `npm run prettier` and note that prettier can successfully reformat the file and maintain its ownership without crashing.

It affects all syntaxes

The file to be autofixed successfully.

The file was not autofixed. The following error was raised:

```shell
Error: EPERM: operation not permitted, chown '/Users/benscott/projects/stylelint-fixfail/style.css.4151106271'
```

kevineinarsson:
Closes #3527 

Reverts https://github.com/stylelint/stylelint/pull/2992


evilebottnawi:
Bad idea fixing one problem and return other problem

kevineinarsson:
Eslint and Prettier both use `fs.writeFileSync` and they're widely used without outcry. Over-engineering to fix a 0.01% issue that also happens to break a common workflow of users isn't useful IMO. I don't like suggesting it but do you think adding an option like `--direct-write` (for CLI) and switching between the methods could be a viable solution?

Issue #4100, opened 2019-06-05
--rd flag makes stylelint not to report any violations at all


sobolevn:
> Clearly describe the bug

Reproduction:

```css
/* index.css */
.empty{}
```

Running `npx stylelint index.css`:

```
1:8  ✖  Unexpected empty block   block-no-empty
```

Then, let's add an unused ignore:

```css
/* index.css */
/* stylelint-disable-next-line property-no-unknown */
.empty{}
```

And run `npx stylelint --rd index.css`. No output will be received. It just passes.

Expected output:
1. `block-no-empty` violation
2. Unused `property-no-unknown` comment violation

> What stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "block-no-empty": true
  }
}
```

> Which version of stylelint are you using?

latest: 9.10.1

sobolevn:
Original issue: https://github.com/wemake-services/wemake-vue-template/issues/608

hudochenkov:
Thanks for the report and for using the template.

> block-no-empty violation

When `--report-needless-disables` stylelint reports only needless disables, and no other violations. We have an [open issue](https://github.com/stylelint/stylelint/issues/4076) to change that.

> Unused property-no-unknown comment violation

I can't reproduce it. With `stylelint@10.1.0` I see the report:

```
$ npx stylelint index.css --report-needless-disables

index.css
unused rule: property-no-unknown, start line: 2, end line: 2
```

hudochenkov:
Closing because of no response from the author.

sobolevn:
I would say that this issue: https://github.com/stylelint/stylelint/issues/4076 tracks the feature I am interested in.

PR #4151, opened 2019-07-16
Stop skipping linting when --report-needless-disables CLI flag is used. Fix #4076

### LINKED ISSUE #4076: Report Needless Disables with Rule Errors and Warnings

Currently, I have to run `stylelint` twice to get all errors and warnings that I need: once with `--report-needless-disables` and another time without it. I should only need to run it once for both.

I would like to change the behavior of `--report-needless-disables` so that such needless disables get reported *with* rule errors/warnings in one shot.

thibaudcolas:

Fixes #4076, #2369, #4100. 

This PR does the same change as #2731, which was closed for cleanup and never reopened. The current behavior can be a source of confusion for people used to ESLint’s [`--report-needless-disables`](https://eslint.org/docs/user-guide/command-line-interface#--report-unused-disable-directives). It also seems a bit wasteful to have to run linting twice just to get the extra reporting of unused disable comments.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -274,7 +274,7 @@ const meowOptions /*: meowOptionsType*/ = {
 
       --report-needless-disables, --rd
 
-        Report stylelint-disable comments that are not blocking a lint warning.
+        Also report errors for stylelint-disable comments that are not blocking a lint warning.
         The process will exit with code ${EXIT_CODE_ERROR} if needless disables are found.
 
       --max-warnings, --mw
@@ -282,9 +282,9 @@ const meowOptions /*: meowOptionsType*/ = {
         Number of warnings above which the process will exit with code ${EXIT_CODE_ERROR}.
         Useful when setting "defaultSeverity" to "warning" and expecting the
         process to fail on warnings (e.g. CI build).
-        
+
       --output-file, -o
-      
+
         Path of file to write report.
 
       --version, -v
@@ -552,8 +552,6 @@ module.exports = (argv /*: string[]*/) /*: Promise<void>|void*/ => {
             if (report) {
               process.exitCode = EXIT_CODE_ERROR;
             }
-
-            return;
           }
 
           if (!linted.output) {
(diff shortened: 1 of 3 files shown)


hudochenkov:
@evilebottnawi do you think it's a breaking change or not?

evilebottnawi:
@hudochenkov And yes and no, for me it is looks like bug (eslint reports erroneous comments and problems with the code together), but some developers potentially could use this command as separate

Will be great hear other opinion

thibaudcolas:
I can see the potential for this change to break people’s builds (whether they are relying on the current behavior and might not expect additional errors, or whether they are expecting the new behavior and they _will_ get errors that previously weren’t reported).

This change looks like the following examples from the project’s [semantic versioning policy](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/semantic-versioning-policy.md):

-   minor release (might break your lint build)
    -   a bug fix in a rule that results in stylelint reporting more errors
    -   a new CLI capability is created
-   major release (likely to break your lint build)
    -   a change in the documented behaviour of an existing rule results in stylelint reporting more errors by default
    -   part of the a CLI is removed or changed in an incompatible way

To me it feels like if the change is considered to be a bug fix then it could be done as a minor release (and if there is breakage so be it), while if it’s considered to be a change in documented behavior of this CLI flag then it’s probably for a major release. A patch release seems inappropriate in any case since there is some potential for this to break people’s linting.

jeddy3:
>> /cc @jeddy3 what do you think about this? For me it is looks like a big bug

> To me it feels like if the change is considered to be a bug fix then it could be done as a minor release (and if there is breakage so be it), while if it’s considered to be a change in documented behavior of this CLI flag then it’s probably for a major release. 

I think the behaviour in this PR is probably what users expect is happening and the original doc ("Report stylelint-disable comments that are not blocking a lint warning.") can be read that way too. So, minor release?

hudochenkov:
> -   Changed: `--report-needless-disables` CLI flag now reports needless disables and runs linting ([#4151](https://github.com/stylelint/stylelint/pull/4151)).

Issue #4156, opened 2019-07-18
Improve --quiet flag wording in CLI help


jackmcpickle:
> Clearly describe the bug

When using the `--quiet` flag in the stylelint cli should I expect the errors to be changed to warnings and the cli to exit with exit code 0 ? Currently the quiet option doesn't seem to do much at all.

> Which rule, if any, is the bug related to?

Related to CLI tool. using `-q` or `--quiet`

> What CSS is needed to reproduce the bug?

```css
.emtpy-block {

}
```

> What stylelint configuration is needed to reproduce the bug?

default config or no config

> Which version of stylelint are you using?

10.1.0`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

running inside as npm script
 ```
"scripts": {
    "test": "stylelint test.css --quiet"
}
```

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

No

> What did you expect to happen?

Errors to be converted to warnings and the cli to exit quietly if it is able to lint correctly, but exit with code 1

> What actually happened (e.g. what warnings or errors did you get)?

`npm ERR! Test failed.  See above for more details.`

When I pass `--print-config` I can see that quiet is set to true.

hudochenkov:
Thanks for the report and for using the template.

Options is working as intended. Warnings are discarded, errors are reported. Though wording is confusing in option description. Now we use word “violation” instead of warning to not confuse with ”warning” severity. 

https://github.com/stylelint/stylelint/blob/ae73c908c26ba800b1575991a558012e51fbb59b/lib/cli.js#L265-L268

It should say:

> Only register violations for rules with an "error"-level severity (ignore "warning"-level).

Would you mind sending a PR?

hassam7:
@hudochenkov may I create PR for this?

jackmcpickle:
Ok will. I get it now. My use case is slightly different, I'm wanting to validate my rules via tests so I can my own preset rules reliably across upgrades. Is there an way to do this? Or something people might be interested in?

jeddy3:
> I'm wanting to validate my rules via tests so I can my own preset rules reliably across upgrades.  Is there an way to do this?

Nothing springs to mind. Can you expand on this use case in a new "Request a feature" issue so that we can look into it, and we'll keep this one around for the documentation tweak?

PR #4157, opened 2019-07-18
Quiet error handling

jackmcpickle:
Closes https://github.com/stylelint/stylelint/issues/4156

fixes issue of errors still being set in quiet mode - should be set to warning which then allows for the CLI to exit quietly too.


hudochenkov:
Closing because it works as intended. https://github.com/stylelint/stylelint/issues/4156#issuecomment-513226929

PR #4163, opened 2019-07-22
fix `no-unit-unknown` rule for variable usage

### LINKED ISSUE #4162: `no-unit-unknown` should ignore less variable
> Clearly describe the bug

`@56789a` is considered as a number with unknown unit in less, instead of a valid variable name in less.

`no-unit-unknown`

```less
@56789a: #56789a; // when you can't find a proper word for this color
a {
  color: @56789a;
}
```

```json
{
  "rules": {
    "no-unit-unknown": "always"
  }
}
```

10.1.0

CLI

Yes, it's related to LESS variable usage

No warning should be reported.

```shell
test.css
2:4    Unexpected unknown unit "a"   unit-no-unknown
```

laysent:

Closes #4162 

I noticed that lint-staged changed the way how it's configured. From [their document](https://github.com/okonet/lint-staged/blob/master/README.md#ignoring-files), they encourage to use ignore configuration of each linter. And since this file is already in `.eslintignore` and `.prettierignore`, I believe it's safe to remote corresponding line in lint-staged config.

### DIFF
--- a/lib/utils/getUnitFromValueNode.js
+++ b/lib/utils/getUnitFromValueNode.js
@@ -25,7 +25,7 @@ module.exports = function(node /*: Object*/) /*: ?string*/ {
 
   if (
     node.type !== "word" ||
-    !isStandardSyntaxValue(value) ||
+    !isStandardSyntaxValue(node.value) ||
     !_.isFinite(parseInt(value)) ||
     node.value[0] === "#"
   ) {
--- a/lib/rules/unit-no-unknown/__tests__/index.js
+++ b/lib/rules/unit-no-unknown/__tests__/index.js
@@ -441,6 +441,10 @@ testRule(rule, {
     {
       code: "a { width: 1em; \n// width: 10pix\n }",
       description: "ignore wrong unit within comments"
+    },
+    {
+      code: "@56789a: #56789a;\na { color: @56789a; }",
+      description: "ignore variable names"
     }
   ],
 


hudochenkov:
I wonder why it wasn't catched by `!isStandardSyntaxValue(value)` function? I think we need to fix this function instead of proposed fix.

evilebottnawi:
hm, https://github.com/stylelint/stylelint/blob/master/lib/utils/isStandardSyntaxValue.js#L23 should be

laysent:
> I wonder why it wasn't catched by `!isStandardSyntaxValue(value)` function? I think we need to fix this function instead of proposed fix.

`blurInterpolation` removes `@` and thus it won't be recognized correctly by `isStandardSyntaxValue` here.

I have changed the way how this issue is fixed, i.e. use `node.value` instead of transformed `value` in `isStandardSyntaxValue`. Please kindly have a look.

laysent:
> Nice solution!
> 
> Could remove lint-stages changes, please? We squash commits on merge and this change would be confusing in unrelated commit.

Sure, I have removed lint-stages changes.

Would you like me to send another PR for this lint-staged changes? Seems at least on my side, lint-staged won't work without changing the config in package.json.

hudochenkov:
> Would you like me to send another PR for this lint-staged changes?

Yes, please :)

hudochenkov:
> -   Fixed: `no-unit-unknown` false positives for at-variables (Less) starting with numbers ([#4163](https://github.com/stylelint/stylelint/pull/4163)).

lib/utils/getUnitFromValueNode.js
```diff
@@ -25,7 +25,7 @@ module.exports = function(node /*: Object*/) /*: ?string*/ {
 
   if (
     node.type !== "word" ||
-    !isStandardSyntaxValue(value) ||
+    !isStandardSyntaxValue(node.value) ||
```
evilebottnawi: /cc @hudochenkov it is break other case, for example `.selector { property: 20px\9; }`

hudochenkov: But it's a hack for IE8.

All tests were passed.

evilebottnawi: @hudochenkov yes, based, on code test should be failed :confounded: https://github.com/laysent/stylelint/blob/0c68261c43123df7e0ffaabbdc350a8a480fb43a/lib/utils/getUnitFromValueNode.js#L29, maybe postcss-value-parser recognize this as other type for node, i will looks on this today

laysent:
@hudochenkov Hi, just curious when would be the next release of stylelint? With this fixed, I could introduce the tool to our team and use it in our projects. ;-)

Issue #4166, opened 2019-07-24
Simplify CHANGELOG format


hudochenkov:
I propose to remove compare tags links from changelog. They not really useful. We create changelog manually for users to read, there is no need to provide “alternative changelog” consists of list of commits. If someone would need to check for some reason it would possible to do from Github.

```diff
- ## [10.1.0][]
+ ## 10.1.0

...

- [10.1.0]: https://github.com/stylelint/stylelint/compare/10.0.1...10.1.0
```

It would solve two issues:

1. It's a little bit annoying when releasing stylelint to rename Head to version and also had to rename link (check [5.3 in release process](https://stylelint.io/developer-guide/releases#process)). And after release when the first PR is merged create a new link for compare.
2. It will fix problem with stylelint.io https://github.com/stylelint/stylelint.io/issues/69.

Compare link in changelog format were introduced in https://github.com/stylelint/stylelint/issues/3901. If I understood correctly one of the reason was to be compatible with `npmpub`. Few recent releases I was using `np` instead of `npmpub` and it was great experience. `npmpub` has not nice experience with one-time passwords. `np` will require to copy release content to Github release though (creating release page opens automatically).

jeddy3:
SGTM

> Few recent releases I was using np instead of npmpub and it was great experience.

That's good to hear!

PR #4093, opened 2019-05-31
ci: azure-pipelines

evilebottnawi:
No issue template - azure pipelines

PR #4167, opened 2019-07-24
Simplify changelog

hudochenkov:

#4166

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,13 @@
 
 All notable changes to this project are documented in this file.
 
-## [Head][]
+## Head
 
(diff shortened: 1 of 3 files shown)


hudochenkov:
Huh, I wonder what Azure pipeline is doing in this PR? We don't use it.

ntwb:
> Huh, I wonder what Azure pipeline is doing in this PR? We don't use it.

It was still enabled on the repo, I've just removed it from the organisation

ntwb:
@jeddy3 See https://github.com/stylelint/stylelint/pull/4093 for some context on the experimental addition of switching from Travis CI to Azure Pipelines

tl;dr; We've switched from Travis-CI.org to Travis-CI.com which gives us access to Windows CI builds on Travis. (We're blocked on this too, alas)

PR #4169, opened 2019-07-24
fix max-line-length rule

### LINKED ISSUE #4059: Fix false positives for multi-line url() in max-line-length
> Clearly describe the bug

There are false positives for max line length when there are spaces inside `url()` content.

`max-line-length`

```css
div {
  background-image: url(
    "data:image/svg+xml;utf8,<svg fill='black' height='24' viewBox='0 0 24 24' width='24' xmlns='http://www.w3.org/2000/svg'><path d='M7 9l5 5 5-5z'/></svg>"
  );
}
```

> What tests are not passing?

```js
{
  code:
    "a {\n  background: url(\n    'something something  something'\n  );\n}"
}
```

```yaml
rules:
  max-line-length:
    - 20
```

`9.10.1`

No.

No warnings to be flagged.

```shell
70:102  ✖  Expected line length to be no more than 80 characters                                               max-line-length
```

vankop:

Fixes #4059

Nothing

### DIFF
--- a/lib/rules/max-line-length/index.js
+++ b/lib/rules/max-line-length/index.js
@@ -10,8 +10,8 @@ const validateOptions = require("../../utils/validateOptions");
 
 const ruleName = "max-line-length";
 const EXCLUDED_PATTERNS = [
-  /url\([\n\r]*((.|[\t\v ])*)\s*\)/gi, // from start of url line to newline symbol
-  /@import\s+(['"].*['"])/gi
+  /url\(\s*((.|[\t\v])*[^\s])\s*\)/gi, // allow tab, whitespace in url content
+  /@import\s+(['"](.|[\t\v])*['"])/gi
 ];
 
 const messages = ruleMessages(ruleName, {
--- a/lib/rules/max-line-length/index.js
+++ b/lib/rules/max-line-length/index.js
@@ -9,6 +9,10 @@ const styleSearch = require("style-search");
 const validateOptions = require("../../utils/validateOptions");
 
 const ruleName = "max-line-length";
+const EXCLUDED_PATTERNS = [
+  /url\([\n\r]*((.|[\t\v ])*)\s*\)/gi, // from start of url line to newline symbol
+  /@import\s+(['"].*['"])/gi
+];
 
 const messages = ruleMessages(ruleName, {
   expected: max =>
@@ -43,6 +47,21 @@ const rule = function(maxLength, options, context) {
     const ignoreNonComments = optionsMatches(options, "ignore", "non-comments");
     const ignoreComments = optionsMatches(options, "ignore", "comments");
     const rootString = context.fix ? root.toString() : root.source.input.css;
+    // Array of skipped sub strings, i.e `url(...)`, `@import "..."`
+    const skippedSubStrings = [];
+    let skippedSubStringsIndex = 0;
+
+    EXCLUDED_PATTERNS.forEach(pattern =>
+      execall(pattern, rootString).forEach(match => {
+        const startOfSubString =
+          match.index + match.match.indexOf(_.get(match, "subMatches[0]", ""));
+
+        return skippedSubStrings.push([
+          startOfSubString,
+          startOfSubString + _.get(match, "subMatches[0].length", 0)
+        ]);
+      })
+    );
 
     // Check first line
     checkNewline(rootString, { endIndex: 0 }, root);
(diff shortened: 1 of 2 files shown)


hudochenkov:
So many changes, I don't understand what it does :(

I wonder would this issue be fixed if it was updating RexExp to [accept new lines](https://stackoverflow.com/a/6883270):

```diff
- const urlArgumentsLength = execall(/url\((.*)\)/gi, lineText).reduce(
+ const urlArgumentsLength = execall(/url\(([\s\S]*)\)/gi, lineText).reduce(
```

vankop:
> So many changes, I don't understand what it does :(
> 
> I wonder would this issue be fixed if it was updating RexExp to [accept new lines](https://stackoverflow.com/a/6883270):
> 
> ```diff
> - const urlArgumentsLength = execall(/url\((.*)\)/gi, lineText).reduce(
> + const urlArgumentsLength = execall(/url\(([\s\S]*)\)/gi, lineText).reduce(
> ```

https://github.com/vankop/stylelint/commit/5677f4766d0f550cbc027d2f1c47968cbacc61d9
This test case will fall

vankop:
current code execute regex line-by-line thats why it falls ( `url(` could be splitted between lines)

lib/rules/max-line-length/index.js
```diff
@@ -43,6 +47,21 @@ const rule = function(maxLength, options, context) {
     const ignoreNonComments = optionsMatches(options, "ignore", "non-comments");
     const ignoreComments = optionsMatches(options, "ignore", "comments");
     const rootString = context.fix ? root.toString() : root.source.input.css;
+    // Array of skipped sub strings, i.e `url(...)`, `@import "..."`
+    const skippedSubStrings = [];
+    let skippedSubStringsIndex = 0;
+
+    EXCLUDED_PATTERNS.forEach(pattern =>
```
vankop: all excluded patterns created array of excluded indexes, for instance:
```
url
 (
 "blabla"
)
```
will return `[[7, 15]]`, where 7 is first `"` and 15 is next after `"blabla"`

lib/rules/max-line-length/index.js
```diff
@@ -75,32 +116,21 @@ const rule = function(maxLength, options, context) {
       }
 
       const rawLineLength = nextNewlineIndex - match.endIndex;
+      const excludedLength = skippedSubStrings[skippedSubStringsIndex]
```
vankop: compute excluded substring length (for current line), for instance:
```
url(
\t\t"bla bla"\u0020);
```
will give `excludedLength` equal to `"bla bla".length`

vankop: it is a bit tricky since
```
url("bla \
bla \
");
```

will give `excludedLength` equals to `"bla \\\n".length` on line 1

lib/rules/max-line-length/index.js
```diff
@@ -62,6 +81,28 @@ const rule = function(maxLength, options, context) {
...
+      if (end < startSubString) {
+        return 0;
+      }
+
+      // Compute excluded substring size regarding to current line indexes
+      const excluded =
+        Math.min(end, endSubString) - Math.max(start, startSubString);
+
+      // Current substring is out of range for next lines
+      if (endSubString <= end) {
```
vankop: current line last index + 1 is over current range end

lib/rules/max-line-length/index.js
```diff
@@ -62,6 +81,28 @@ const rule = function(maxLength, options, context) {
       });
     }
 
+    function tryToPopSubString(start, end) {
+      const [startSubString, endSubString] = skippedSubStrings[
+        skippedSubStringsIndex
+      ];
+
+      // Excluded substring does not presented in current line
+      if (end < startSubString) {
```
vankop: current excluded range is not presented on current line

lib/rules/max-line-length/index.js
```diff
@@ -62,6 +81,28 @@ const rule = function(maxLength, options, context) {
...
+        return 0;
+      }
+
+      // Compute excluded substring size regarding to current line indexes
+      const excluded =
```
vankop: Compute excluded substring size regarding to current line indexes =)

vankop:
@hudochenkov i have add some explanation

hudochenkov:
> -   Fixed: `max-line-length` false positives for multi-line url() ([#4169](https://github.com/stylelint/stylelint/pull/4169)).

Issue #3041, opened 2017-11-28
Always returns stylelintError with boolean values


chinesedfan:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug, maybe. It is a question about internal implementations.

> Which rule, if any, is this issue related to?

Not related to a certain rule.

> What CSS is needed to reproduce this issue?

Not related to CSS codes.

> What stylelint configuration is needed to reproduce this issue?

In fact, it doesn't matter.
```
const rules = {
  'block-no-empty': true,
  'declaration-block-no-duplicate-properties': true,
  indentation: 2
}
```

> Which version of stylelint are you using?

8.0.0

> How are you running stylelint: CLI, PostCSS plugin, Node API?

Node API.
```
stylelint
      .lint({
        files: [fixture],
        syntax: 'scss',
        config: {
          processors: [processor], // styled-components/stylelint-processor-styled-components
          rules
        }
      })
      .then(result => {
        data = result
        done()
      })
      .catch(err => {
        data = err
        done()
      })
```

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No.

> What did you expect to happen?

`stylelint` always returns `stylelintError` with boolean values. Are there any special considerations?

> What actually happened (e.g. what warnings or errors you are getting)?

- `undefined` if no errors exist.
- `null` if the file is ignored. See [lib/lintSource.js](https://github.com/stylelint/stylelint/blob/master/lib/lintSource.js#L239).
- `true` if some rules have called the function `report`. See [/lib/utils/report.js](https://github.com/stylelint/stylelint/blob/master/lib/utils/report.js#L76).

hudochenkov:
Thanks for the report and for using the template.

I'm not sure why it behaves like that. Is there any problem this behavior creates?

chinesedfan:
@hudochenkov Not big problem, but weird. I was writing a [stylelint processor](https://github.com/styled-components/stylelint-processor-styled-components/blob/master/src/index.js#L62), which needed to filter some special warnings. When I tried to recover the `errored` field, I got this question.

jeddy3:
@chinesedfan Would making the return behaviour more consistent help you as someone who wishes to extend stylelint? If so, I suggest we welcome a contribution that makes this change.

jeddy3:
[...]

PR #4174, opened 2019-07-26
always returns boolean stylelintError

vankop:

Fixes https://github.com/stylelint/stylelint/issues/3041

Added `ignoreFile` system test


vankop:
> Looks like small breaking change, but for me it is bug in our design api

Yep, it is breaking change =(

hudochenkov:
> -   Changed: always return `stylelintError` as a boolean ([#4174](https://github.com/stylelint/stylelint/pull/4174)).

PR #4176, opened 2019-07-26
add ignore selectors-within-list to no-descending-specificity rule

### LINKED ISSUE #3196: Add ignore: ["selectors-within-list"] to no-descending-specificity

`no-descending-specificity`

```css
a {
	color: #000;
}

a:hover {
	text-decoration: underline;
}

h1,
h2,
h3,
h4,
h5,
h6,
a,
button {
	user-select: none;
	cursor: default;
}
```

```json
{
  "rules": {
  }
}
```

9.1.1

CLI

No

I expected it to not trigger the `no-descending-specificity` rule as it's common to write CSS like this.

It triggered the rule on the `a,`. It would be unnatural to move the `a:hover` below the `...h5, h6, a,...`.

I don't think it should trigger when the selector is part of multiple selectors. Or there should at least be an option to achieve this.

vankop:

closes #3196

nothing

### DIFF
--- a/lib/rules/no-descending-specificity/README.md
+++ b/lib/rules/no-descending-specificity/README.md
@@ -105,3 +105,24 @@ a {}
   #baz a {}
 }
 ```
+
+### `ignore: ["selectors-within-list"]`
+
+Ignores selectors within list of selectors.
+
+The following patterns are considered violations:
+
+```css
+b a {}
+h1 {}
+h2 {}
+h3 {}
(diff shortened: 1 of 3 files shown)


hudochenkov:
> -   Added: `ignore: ["selectors-within-list"]` to `no-descending-specificity` ([#4176](https://github.com/stylelint/stylelint/pull/4176)).

Issue #2657, opened 2017-06-20
Add support for all logical combinations (:matches, :has) in selector-*


evilebottnawi:
Investigate `:matches()`, `:not()` and `:has()` for `selector-*` rules.

Now we support only `not:()`, but we can add also other, it should not be a long time or we can add `9.0`milestone for this.

Spec: https://drafts.csswg.org/selectors-4/#logical-combination

PR #4179, opened 2019-07-29
support logical combinations

vankop:

closes https://github.com/stylelint/stylelint/issues/2657

nothing

### DIFF
--- a/lib/rules/selector-max-attribute/index.js
+++ b/lib/rules/selector-max-attribute/index.js
@@ -2,6 +2,7 @@
 
 const _ = require("lodash");
 const hasUnresolvedNestedSelector = require("../../utils/hasUnresolvedNestedSelector");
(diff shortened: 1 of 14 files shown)


vankop:
> Looks good, can we add test with nested with multiple `body:not(div):not(span) {}` ?

which rule? or every `selector-pseudo-class-*` ?

evilebottnawi:
Any, just ensure this case also covered

vankop:
> Looks good, can we add test with nested with multiple `body:not(div):not(span) {}` ?

done

hudochenkov:
> -   Added: `selector-*` support for all logical combinations (:matches, :has) ([#4179](https://github.com/stylelint/stylelint/pull/4179)).

jeddy3:
@vankop It's great to see some old issues being addressed! Thank you for your work on all your recent pull requests.

PR #4181, opened 2019-07-30
Add report invalid scope disables

### LINKED ISSUE #2291: Add reportInvalidScopeDisables flag
Ref: https://github.com/stylelint/stylelint/pull/2232#discussion_r96887239

To report the disabling of rules that don't exist within the configuration object.

Treated as a similar violation to `reportNeedlessDisables`, and called `reportInvalidScopeDisables`. 

Given the following config:

```js
{
  "rules": {
    "block-no-empty": true
  }
}
```

The following source will trigger a violation:

```css
/* stylelint-disable indentation */
```

As `indentation` isn't present within the configuration object.

vankop:

closes #2291

Naive idea of collecting rules that don't exist within the configuration object is pretty straight forward -  find all disabled rules that don't exist in configuration using info in `disableRanges` object. However, I faced with problem in tests:
```
/* stylelint-disable not-in-config */ a {} /* stylelint-enable */
/* stylelint-disable */ b {} /* stylelint-enable */
```
and vice verca
```
/* stylelint-disable */ b {} /* stylelint-enable */
/* stylelint-disable not-in-config */ a {} /* stylelint-enable */
```
In both cases rule `not-in-config` will be disabled twice, so `invalidScopeDisables` will report twice as well. I decided to enhance `disableRange` with 2 flags `strictStart: boolean` and `strictEnd?: boolean`, that work as describe below:

for each rule in `stylelint-disable* <rule>` `strictStart` will be true
for each rule in `stylelint-enable* <rule>` `strictEnd` will be true
for each rule in `stylelint-disable*` `strictStart` will be false and for `all` rule will be true
for each rule in `stylelint-enable*` `strictEnd` will be false and for `all` rule will be true
```
/* stylelint-disable not-in-config */ a {} /* stylelint-enable */ (1)
/* stylelint-disable */ b {} /* stylelint-enable */ (2)
```
will result in
```
{
    all: [{start: 2, end: 2, strictStart: true, strictEnd: true}],
    "not-in-config": [{start: 1, end: 1, strictStart: true, strictEnd: false}, {start: 2, end: 2, strictStart: false, strictEnd: false}]
}
```
I updated `disableRanges` tests, so you can find more examples there.


vankop:
Actually PR looks pretty big (and it will be bigger since need to add more tests), if we ok to split it in different tickets I can do it

vankop:
> CLI should support this option too

Thanks, forgot about cli description.

hudochenkov:
Error might be caused by https://github.com/stylelint/stylelint/pull/4174 change.

lib/assignDisabledRanges.js
```diff
@@ -66,38 +85,43 @@ module.exports = function(
 
     if (ruleName === ALL_RULES) {
       Object.keys(disabledRanges).forEach(disabledRuleName => {
-        startDisabledRange(line, disabledRuleName);
-        endDisabledRange(line, disabledRuleName);
+        const strict = disabledRuleName === ALL_RULES;
+
+        startDisabledRange(line, disabledRuleName, strict);
+        endDisabledRange(line, disabledRuleName, strict);
```
vankop: regarding to `strict` flag definition - if `ruleName` is `all`, then `strictStart`/`strictEnd` true, otherwise false

lib/assignDisabledRanges.js
```diff
@@ -66,38 +85,43 @@ module.exports = function(
...
       });
     } else {
-      startDisabledRange(line, ruleName);
-      endDisabledRange(line, ruleName);
+      startDisabledRange(line, ruleName, true);
+      endDisabledRange(line, ruleName, true);
```
vankop: provided `ruleName` is not `all`, e.g.:
`/* stylelint-disable my-rule */`
so `strictStart`/`strictEnd` true

lib/assignDisabledRanges.js
```diff
@@ -66,38 +85,43 @@ module.exports = function(
...
 
-      if (ruleIsDisabled(ruleToDisable)) {
-        throw comment.error(`"${ruleToDisable}" has already been disabled`, {
-          plugin: "stylelint"
+      if (isAllRules) {
+        Object.keys(disabledRanges).forEach(ruleName => {
+          startDisabledRange(
+            comment.source.start.line,
+            ruleName,
+            ruleName === ALL_RULES
```
vankop: disabling `all` rule, so `strictStart` true only for `all` rules

lib/assignDisabledRanges.js
```diff
@@ -116,7 +140,11 @@ module.exports = function(
 
         Object.keys(disabledRanges).forEach(ruleName => {
           if (!_.get(_.last(disabledRanges[ruleName]), "end")) {
-            endDisabledRange(comment.source.end.line, ruleName);
+            endDisabledRange(
```
vankop: same for end

lib/invalidScopeDisables.js
```diff
@@ -0,0 +1,67 @@
...
+}*/
+
+module.exports = function(
+  results /*: Array<stylelint$result>*/,
+  config /*: stylelint$config*/
+) /*: stylelint$disableOptionsReport*/ {
+  const report = [];
+  const usedRules = new Set(Object.keys(config.rules || {}));
+
+  usedRules.add("all");
```
vankop: setup all used rules `Set` (include `all` rule)

lib/invalidScopeDisables.js
```diff
@@ -0,0 +1,67 @@
...
+      return;
+    }
+
+    const sourceReport = { source: result.source, ranges: [] };
+    const rangeData /*: rangeDataType*/ =
+      result._postcssResult.stylelint.disabledRanges;
+    const disabledRules = Object.keys(rangeData);
+
+    disabledRules.forEach(rule => {
+      if (usedRules.has(rule)) {
```
vankop: skip creating warning for rule that exists in config (include `all` rule)

lib/invalidScopeDisables.js
```diff
@@ -0,0 +1,67 @@
...
+        return;
+      }
+
+      rangeData[rule].forEach((range /*: unusedRangeT */) => {
+        if (!range.strictStart && !range.strictEnd) {
```
vankop: if current range is not strict from start or end, then skip it, e.g:
```
/* stylelint-disable */ a { color: red} /* stylelint-enable */
/* stylelint-disable color-hex */ a { color: red} /* stylelint-enable */
```

current code will skip creating warning for `color-hex`

lib/invalidScopeDisables.js
```diff
@@ -0,0 +1,67 @@
...
+          return;
+        }
+
+        sourceReport.ranges.push({
```
vankop: creating warning using `report-needless-disable` warnings format

vankop:
@hudochenkov I have clarified some stuff

hudochenkov:
> -   Added: `--reportInvalidScopeDisables` CLI flag ([#4181](https://github.com/stylelint/stylelint/pull/4181)).

Issue #4113, opened 2019-06-13
 False positive in no-duplicate-selector


ssivanatarajan:
> Clearly describe the bug
  
`Stylelint marks all the selectors as duplicated. Actually, one selector is duplicate`

> Which rule, if any, is the bug related to?

`no-duplicate-selector`

> What CSS is needed to reproduce the bug?

```css
.a,.a,.b,.c,.d{
color:red;
}
```

> What stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "no-duplicate-selector": "always"
  }
}
```

> Which version of stylelint are you using?

 `10.0.1`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

`Node API`

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

`NO`
> What did you expect to happen?

`Only one warning should be reported.`

```
Unexpected duplicate selector ".a", first used at line 2
```

> What actually happened (e.g. what warnings or errors did you get)?

```
Unexpected duplicate selector ".a", first used at line 2 (no-duplicate-selectors)
Unexpected duplicate selector ".b", first used at line 2 (no-duplicate-selectors)
Unexpected duplicate selector ".c", first used at line 2 (no-duplicate-selectors)
Unexpected duplicate selector ".d", first used at line 2 (no-duplicate-selectors)
```

hudochenkov:
Thanks for the report and for using the template.

I can reproduce it.

vankop:
I take this

PR #4183, opened 2019-07-31
Support several warnings in tests

### LINKED ISSUE #4172: Add support to test for several warnings

Currently it not possible to detect several warnings in tests, see
https://github.com/stylelint/stylelint/blob/master/jest-setup.js#L92

However, sometimes it is necessary for tests to detect such cases, see https://github.com/stylelint/stylelint/issues/4113

My suggestion is to rework reject object syntax.

Typescript defintion:
```
type WarningDescription = {
    message: string,
    line?: number,
    column?: number,
    fixed?: string
};

type Reject = {code: string, warnings?: Array<WarningDescription>} & WarningDescription;
```

vankop:

closes #4172 

Nothing

### DIFF
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -91,19 +91,20 @@ global.testRule = (rule, schema) => {
                 return stylelint(options).then(output => {
                   const actualWarnings = output.results[0].warnings;
 
-                  if (
-                    schema.fix &&
-                    testCase.warnings &&
-                    testCase.warnings.length > 1
-                  ) {
-                    throw new Error(
-                      "Several expected warnings can not be used with { fix: true } in schema yet"
-                    );
-                  }
-
                   if (testCase.warnings) {
                     expect(actualWarnings).toHaveLength(
                       testCase.warnings.length
+                    );
+                  }
+
+                  if (
+                    schema.fix &&
+                    !testCase.fixed &&
+                    testCase.fixed !== "" &&
+                    !testCase.unfixable
+                  ) {
+                    throw new Error(
+                      "If using { fix: true } in test schema, all reject cases must have { fixed: .. }"
                     );
                   }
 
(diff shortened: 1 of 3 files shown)


jest-setup.js
```diff
@@ -89,72 +89,91 @@ global.testRule = (rule, schema) => {
...
-
-                  if (testCase.message !== undefined) {
-                    expect(_.get(warning, "text")).toBe(testCase.message);
+                  if (
+                    schema.fix &&
+                    testCase.warnings &&
+                    testCase.warnings.length > 1
+                  ) {
+                    throw new Error(
+                      "Several expected warnings can not be used with { fix: true } in schema yet"
```
vankop: Don't know how to implement it better

hudochenkov: This is not needed. With `--fix` all warnings are fixed.

vankop:
> I wonder why `.a, .a, .a {}` has only one warning then

it was fixed in https://github.com/stylelint/stylelint/issues/4113 , right now we have one warning for one selector

vankop:
So,
```
.a, .b, .a, .b, .a, .b {}
```
Will produce 2 warnings for `a` and for `b`
```
.a, .a, .a {}
```
only for `.a`

jest-setup.js
vankop: So you suggest to allow several `fixed` in `warnings` ?

hudochenkov: Based on how `fix` works we'll need only one `fixed` per case.

vankop: done

jest-setup.js
```diff
@@ -89,26 +89,16 @@ global.testRule = (rule, schema) => {
                 };
 
                 return stylelint(options).then(output => {
-                  const warning = output.results[0].warnings[0];
+                  const actualWarnings = output.results[0].warnings;
 
-                  expect(output.results[0].parseErrors).toEqual([]);
```
hudochenkov: We need this line.

vankop: yep, my fault. fixed

PR #4184, opened 2019-08-01
Property no unknown

### LINKED ISSUE #4020: Fix false positives for overflowX for CSS-in-JS in property-no-unknown
> Clearly describe the bug

Using React and the CSS object notation leads to an error for the following piece of code:
```jsx
<div style={{overflowX: 'scroll', color: 'black'}}>
```
error:
```shell
demo.jsx
68:41  ×  Unexpected unknown property "overflowx"   property-no-unknown
```

`property-no-unknown`

```js
style={overflowX: 'scroll'}
```

e.g.

```json
{
  "extends": ["stylelint-config-standard", "stylelint-config-prettier"],
}
```

`9.10.1`

CLI, VScode

No.

No warnings to be flagged.

```shell
demo.jsx
68:41  ×  Unexpected unknown property "overflowx"   property-no-unknown
```

vankop:

closes #4020 

nothing

### DIFF
--- a/lib/rules/property-no-unknown/__tests__/index.js
+++ b/lib/rules/property-no-unknown/__tests__/index.js
@@ -229,6 +229,16 @@ testRule(rule, {
     },
     {
       code: "const Component = styled.a`\n\t${rule}: 1;\n`"
+    },
+    {
+      code: `
+        import glm from 'glamorous';
+        const Component = glm.a({
+            overflowX: 'auto',
+            overflowWrap: 'auto',
+            flexDirection: 'row',
+        });
+      `
     }
   ],
 
--- a/lib/rules/property-no-unknown/__tests__/index.js
+++ b/lib/rules/property-no-unknown/__tests__/index.js
@@ -261,6 +261,17 @@ testRule(rule, {
       message: messages.rejected("rule"),
       line: 2,
       column: 2
+    },
+    {
+      code: `
+        import glm from 'glamorous';
+        const Component = glm.a({
+            overflwX: 'auto',
+        });
+      `,
+      message: messages.rejected("overflw-x"),
+      line: 4,
+      column: 12
     }
   ]
 });
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "pify": "^4.0.1",
     "postcss": "^7.0.14",
     "postcss-html": "^0.36.0",
-    "postcss-jsx": "^0.36.2",
+    "postcss-jsx": "^0.36.3",
     "postcss-less": "^3.1.4",
     "postcss-markdown": "^0.36.0",
     "postcss-media-query-parser": "^0.2.3",


lib/rules/property-no-unknown/__tests__/index.js
```diff
@@ -229,6 +229,16 @@ testRule(rule, {
...
+    },
+    {
+      code: `
+        import glm from 'glamorous';
+        const Component = glm.a({
+            overflowX: 'auto',
+            overflowWrap: 'auto',
+            flexDirection: 'row',
+        });
+      `
```
evilebottnawi: Can we add test with failed example?

vankop: done

vankop:
Don't have we a problem with node 8 runner in CI? always fails in node 8 runner

evilebottnawi:
Strange error, agree

vankop:
As I understand we using windows runners, so we can use build-in images for runners, see https://www.appveyor.com/docs/windows-images-software/#node-js instead of installing node 8 manually https://github.com/stylelint/stylelint/blob/master/appveyor.yml#L19 . I am not familiar with appveyor CI/CD, but it is pretty common functionality + if we will add lock file we could install packages faster with `npm ci`

Also one idea, we using cache from `node_modules`, but I don't know is it same cache for node-8 and node-10, if it is so it could lead to some uncommon problems, for example, node-sass build is different depends on platform, see https://github.com/sass/node-sass/blob/master/package.json

evilebottnawi:
> As I understand we using windows runners, so we can use build-in images for runners, see https://www.appveyor.com/docs/windows-images-software/#node-js instead of installing node 8 manually https://github.com/stylelint/stylelint/blob/master/appveyor.yml#L19

This will not add much time to us, but will increase the maintenance time on the project

> I am not familiar with appveyor CI/CD, but it is pretty common functionality + if we will add lock file we could install packages faster with npm ci

/cc @stylelint/core what is blocker for using lock files, it is good practice

vankop:
> This will not add much time to us, but will increase the maintenance time on the project

Why it will be increased? AppVeyor is pretty popular and they maintain there images, so it will make behavior of our code *expected* since they create node 8 environment for us instead of using combo => unknown runner env + node 8 (If I understand how appveyor works 😁) For gitlab it is common practice

evilebottnawi:
> Why it will be increased?

Who will update the images? Tomorrow we may be busy and not find it at this time

vankop:
Looks like we have misunderstanding each other.

How it works now:
1. Appveyor starts default image on virtual machine
2. clone repo in it
3. cd in repo root
4. Runs `install` commands: a. Start powershell install node 8 b. runs `npm install` https://github.com/stylelint/stylelint/blob/master/appveyor.yml
5. runs tests

Cost of maintenance each version add new environment

How it will work with predefined images:
1. Appveyor starts **node 8 image** on virtual machine
2. clone repo in it
3. cd in repo root
4. Runs `install` commands: ~~a. Start powershell install node 8~~ b. runs `npm install` https://github.com/stylelint/stylelint/blob/master/appveyor.yml
5. runs tests

example config:
```
environment:
    matrix:
        - image: Node.js 8.16.0
        - image: Node.js 10.16.0
```

Cost of maintenance each version add new image in list

vankop:
If I make mistake in describing it will be great, if you will clarify how it works for me

hudochenkov:
> -   Fixed: `property-no-unknown` for overflowX for CSS-in-JS ([#4184](https://github.com/stylelint/stylelint/pull/4184)).

PR #4185, opened 2019-08-03
Update about DOM limitations

### LINKED ISSUE #2489: Fix false positives for pseudo classes in no-descending-specificity
Based on I'm using sass to develop css,
I don't want to define the class (`.b`) without parents.
however, it will happen error with this rule.

is it(as below example) not usual for define class?

but I still need this rule to validate the other situations.

`no-descending-specificity`

```sass
.a {
    .b {}
}

.a--modify {
    .b {}
}

```

`no-descending-specificity`

```json
{
  "rules": {
  }
}
```

version: `7.8.0`

Expected selector ".a--modify .b" to come before selector ".a .b"
Even I switch the order, it still happened.

nosilleg:

Closes #2489 

Attempted to use simple English, but may need revision.

The example should cover a lot of the examples I've seen, but most people are using SASS instead of straight CSS.  I *think* this this is easy enough for the difference not to matter, but maybe we need different/additional examples.

### DIFF
--- a/lib/rules/no-descending-specificity/README.md
+++ b/lib/rules/no-descending-specificity/README.md
@@ -44,9 +44,9 @@ For example the following will cause an error:
 
 This is a correct error because the `a:hover` on line 2 has a higher specificity than the `a` on line 3.
 
-This may lead to confusion because "the two selectors will never match the same `a` in the DOM". However, since the linter does not have access to the DOM it can not evaluate this, and therefore correctly reports the error about decending specificity.
+This may lead to confusion because "the two selectors will never match the same `a` in the DOM". However, since the linter does not have access to the DOM it can not evaluate this, and therefore correctly reports the error about descending specificity.
 
-It may be possible to restructure your CSS to remove the error, otherwise it is recommended that you disable the rule for that line and leave a comment saying why the error should be ignored. Note that disabling the rule will cause addditional valid errors from being reported.
+It may be possible to restructure your CSS to remove the error, otherwise it is recommended that you disable the rule for that line and leave a comment saying why the error should be ignored. Note that disabling the rule will cause additional valid errors from being reported.
 
 ## Options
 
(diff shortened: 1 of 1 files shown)


lib/rules/no-descending-specificity/README.md
```diff
@@ -12,20 +12,42 @@ Source order is important in CSS, and when two selectors have the *same* specifi
...
+
+```css
+.component1 a {}
+.component1 a:hover {}
+.component2 a {}
+```
+
+This is a correct error because the `a:hover` on line 2 has a higher specificity than the `a` on line 3.
+
+This may lead to confusion because "the two selectors will never match the same `a` in the DOM". However, since the linter does not have access to the DOM it can not evaluate this, and therefore correctly reports the error about decending specificity.
```
JohnTitor:
```suggestion
This may lead to confusion because "the two selectors will never match the same `a` in the DOM". However, since the linter does not have access to the DOM it can not evaluate this, and therefore correctly reports the error about descending specificity.
```

lib/rules/no-descending-specificity/README.md
```diff
@@ -12,20 +12,42 @@ Source order is important in CSS, and when two selectors have the *same* specifi
...
+
+It may be possible to restructure your CSS to remove the error, otherwise it is recommended that you disable the rule for that line and leave a comment saying why the error should be ignored. Note that disabling the rule will cause addditional valid errors from being reported.
```
JohnTitor:
```suggestion
It may be possible to restructure your CSS to remove the error, otherwise it is recommended that you disable the rule for that line and leave a comment saying why the error should be ignored. Note that disabling the rule will cause additional valid errors from being reported.
```

lib/rules/no-descending-specificity/README.md
hudochenkov: @nosilleg could you accept these two suggestions, please?

nosilleg: @hudochenkov Done.

PR #4187, opened 2019-08-05
chore(update): postcss-value-parser

evilebottnawi:

update deps

Refactor from https://github.com/stylelint/stylelint/pull/4163

Also find some bugs with new version, working on new release without bugs


lib/__tests__/printConfig.test.js
```diff
@@ -57,7 +57,7 @@ it("printConfig with multiple paths should throw", () => {
 });
 
 it("printConfig with globs should throw", () => {
-  expect(
+  return expect(
```
evilebottnawi: After update eslint plugin jest throw errors o n this, just fix eslint problem

lib/utils/getUnitFromValueNode.js
```diff
@@ -16,21 +15,26 @@ module.exports = function(node /*: Object*/) /*: ?string*/ {
...
+  if (!isStandardSyntaxValue(node.value)) {
+    return null;
+  }
+
+  // Ignore HEX
+  if (node.value[0] === "#") {
+    return null;
+  }
+
+  // Remove non standard stuff
```
evilebottnawi: Here interesting case, for example:
```
a { margin: calc(100% - #{$margin * 2pix}); }
``` 

Now we throw error on `pix`, but it is part of interpolation (i.e. non standard) and i think we should ignore content in/into non standard syntax, but i am not sure, need some feedback

evilebottnawi:
Some stuff from `unit` and `value` rules should be removed on refactor, i will do this in next PRs

lib/utils/getUnitFromValueNode.js
jeddy3: > i think we should ignore content in/into non standard syntax

Yep, I think it's best if we ignore things within interpolation. The change in behaviour will be non-breaking.

evilebottnawi: @jeddy3 let's do it in other PR

PR #4194, opened 2019-08-08
Added cli flag "--disable-globbing"

### LINKED ISSUE #4193: Allow to use full file path without converting it to glob

theodab:

Closes #4193

Not really. I'm open to renaming the flag as you people see fit; I'm not attached to the specific name of "--disable-globbing".

### DIFF
--- a/flow-typed/stylelint.js
+++ b/flow-typed/stylelint.js
@@ -138,6 +138,7 @@ export type stylelint$standaloneOptions = {
   configFile?: string,
   configBasedir?: string,
   configOverrides?: Object,
+  disableGlobbing?: boolean,
   printConfig?: string,
   ignoreDisables?: boolean,
   ignorePath?: string,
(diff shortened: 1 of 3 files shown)


joeyparrish:
Looks like there is a minor style error reported by `npm run pretest`:

https://travis-ci.com/stylelint/stylelint/jobs/223657863

```
/home/travis/build/stylelint/stylelint/lib/standalone.js
  198:3  error  Expected blank line before this statement  padding-line-between-statements
```

theodab:
Ok, I fixed that linter error, and ran the linter locally to make sure I didn't have any others.

hudochenkov:
Labeling this as “blocked”, because we need to find another solution to the original issue.

hudochenkov:
We decided that [this a bug](https://github.com/stylelint/stylelint/issues/4193) and we need it to fix it, rather create new flag.

PR #4196, opened 2019-08-09
chore(package): update eslint to version 6.1.0

ntwb:
Closes #4128

### DIFF
--- a/lib/rules/declaration-block-no-shorthand-property-overrides/index.js
+++ b/lib/rules/declaration-block-no-shorthand-property-overrides/index.js
@@ -39,7 +39,12 @@ const rule = function(actual) {
         }
 
         overrideables.forEach(longhandProp => {
-          if (!declarations.hasOwnProperty(prefix + longhandProp)) {
+          if (
+            !Object.prototype.hasOwnProperty.call(
+              declarations,
+              prefix + longhandProp
+            )
+          ) {
             return;
           }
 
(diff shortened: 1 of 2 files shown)


ntwb:
> This is the correct behavior because `no-console` rule was removed from `eslint:recommended` config. If you want `no-console` to enable, configure it in your `.eslintrc` file.

> See also: eslint.org/docs/user-guide/migrating-to-6.0.0

I've added `no-console` to `eslint-config-stylelint` in https://github.com/stylelint/eslint-config-stylelint/pull/65

Will release `eslint-config-stylelint` version `11.1.0` shortly to fix

ntwb:
ESLint v6 also added a new `no-prototype-builtins` rule to `eslint/recommended`

Commit https://github.com/stylelint/stylelint/pull/4196/commits/6c49b07a483ef5ca0c9060625e0d65d6b9e93133 fixes the single instance of this reported:
> https://eslint.org/docs/user-guide/migrating-to-6.0.0
> _"`no-prototype-builtins` reports method calls like `foo.hasOwnProperty("bar")` (which are a frequent source of bugs), and suggests that they be replaced with `Object.prototype.hasOwnProperty.call(foo, "bar")` instead."_

ntwb:
Yay, @stylelint/core this is now ready for review and merge please 🙇🏼‍♂️

Issue #4206, opened 2019-08-12
Add support for `color-scheme`


tomayac:
> What is the problem you're trying to solve?

https://drafts.csswg.org/css-color-adjust-1/#color-scheme-prop defines a CSS property `color-scheme` that lets pages express which color schemes they support. This is along the lines of https://github.com/stylelint/stylelint/pull/3951/files.

> What solution would you like to see?

Currently using this property results in an ` Unexpected unknown property "color-scheme"   property-no-unknown` error. A new rule would fix this.

tomayac:
This got fixed in an external dependency, see https://github.com/known-css/known-css-properties/pull/98/files#diff-488eead074948aa18d9072d416c1f495R301, all good :-)

PR #4207, opened 2019-08-12
Add support for `color-scheme`

tomayac:
Defined in https://drafts.csswg.org/css-color-adjust-1/#color-scheme-prop. Fixes https://github.com/stylelint/stylelint/issues/4206.


tomayac:
Seems like this got fixed in https://github.com/known-css/known-css-properties/pull/98/files#diff-488eead074948aa18d9072d416c1f495R301, an external dependency. All good then, thanks for looking into this.

Issue #4144, opened 2019-07-02
Add ignoreBeforeAtRules: [] to block-closing-brace-newline-after


srawlins:
> Clearly describe the bug

The [block-closing-brace-newline-after](https://stylelint.io/user-guide/rules/block-closing-brace-newline-after/#optional-secondary-options) rule has a neat [ignoreAtRules](https://stylelint.io/user-guide/rules/block-closing-brace-newline-after/#optional-secondary-options) option, which allows non-newline characters after an at-rule-associated `}`. The examples given are:

```scss
@if ($var) {
  color: pink;
} @else if ($var2) {
  color: red;
} @else {
  color: blue;
}
```

```scss
@if ($var) { color: pink; } @else { color: blue; }
```

This ignores too many violations. E.g. this also becomes considered clean code:

```scss
@if ($var) { color: pink; } p { margin: 2em; }
```

The `ignoreAtRules` option should only allow more at-rules after the `}` attached to an ignored at-rule. I can't conceive of why one would allow a declaration after such an `}`, or a rule, etc. I think the option was written specifically to support Sass's `@if`/`@else`.

> Which rule, if any, is the bug related to?

`block-closing-brace-newline-after`

> What CSS is needed to reproduce the bug?

```scss
@if ($var) { color: pink; } p { margin: 2em; }
```

> What stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "block-closing-brace-newline-after": ["always", "ignoreAtRules": ["if", "else"]]
  }
}
```

> Which version of stylelint are you using?

`9.6.0`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

Node.js API using `cli.js`

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

Yes, it's related to Sass nested properties.

> What did you expect to happen?

block-closing-brace-newline-after to be flagged for the rule after the `}` associated with the `@if`.

> What actually happened (e.g. what warnings or errors did you get)?

No warnings were flagged.

If this is accepted as a bug, I can likely write a PR to fix it.

evilebottnawi:
For specify sass/scss rules better open issue in stylelint-scss, think option ignore any newlines for `at-rules` in your list

srawlins:
[...]

PR #4210, opened 2019-08-14
Add ignoreBeforeAtRules: [] to block-closing-brace-newline-after

rambleraptor:

Closes https://github.com/stylelint/stylelint/issues/4144

No, should be self-explanatory.

### DIFF
--- a/lib/rules/block-closing-brace-newline-after/README.md
+++ b/lib/rules/block-closing-brace-newline-after/README.md
@@ -174,3 +174,37 @@ The following patterns are *not* considered violations:
 ```css
 @if ($var) { color: pink; } @else { color: blue; }
 ```
+
+## Optional secondary options
+
+### `ignoreBeforeAtRules: ["/regex/", "non-regex"]`
+
+Ignore whitespace suggestions before these at-rules.
+
+For example, with `"always"` or `"always-multi-line"`.
+
+Given:
+
+```js
+["else"]
+```
+
+The following patterns are *not* considered violations:
+
+```css
+@if ($var) {
+  color: pink;
+} @else if ($var2) {
+  color: red;
+} @else {
+  color: blue;
+}
+```
+
+```css
+@media (width < 28em) {
+  background: red;
+} @else {
+  background: green;
+}
+```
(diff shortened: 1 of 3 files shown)


rambleraptor:
@nex3 fyi

lib/rules/block-closing-brace-newline-after/README.md
```diff
@@ -174,3 +174,37 @@ The following patterns are *not* considered violations:
...
+  color: red;
+} @else {
+  color: blue;
+}
+```
+
+```css
+@media (width < 28em) {
+  background: red;
+} @else {
```
hudochenkov: Could you remove this example, please? It's not valid CSS and it's not valid SCSS.

lib/rules/block-closing-brace-newline-after/__tests__/index.js
```diff
@@ -207,6 +207,37 @@ testRule(rule, {
...
+    },
+    {
+      code: `@media (width < 28em) {
+        background: red;
+      } @else {
+        background: green;
+      }`
+    }
+  ],
+  reject: [
```
vankop: @hudochenkov could we please update tests README with suggested reject test case style:
```
warnings: [..]
```
to avoid problem with mismatching amount of warnings

vankop: @rambleraptor could you please rewrite rejected test cases as follow https://github.com/stylelint/stylelint/blob/0bd166f57814946b81515a9d96b9313078b27d11/lib/rules/selector-max-pseudo-class/__tests__/index.js#L189

hudochenkov: > to avoid problem with mismatching amount of warnings

@vankop I don't think we really needed. Our tests are quite verbose already. Having `warnings: []` with only one warning inside will make them more verbose. When [strict check](https://github.com/stylelint/stylelint/pull/4231) will be enabled, we won't miss any warnings.

vankop: you are correct, I did some work toward it =)

lib/rules/block-closing-brace-newline-after/__tests__/index.js
```diff
@@ -207,6 +207,37 @@ testRule(rule, {
...
+      code: `@if ($var) {
+        color: pink;
+      } @else if ($var2) {
+        color: red;
+      } @else {
+        color: blue;
+      }
+      `
+    },
+    {
```
jeddy3: Please remove this test too as it's invalid SCSS.

hudochenkov:

PR #4216, opened 2019-08-20
rework scss comments in comments-no-empty rule

### LINKED ISSUE #4037: Fix false negatives for double slash comments in no-empty-comment
> Clearly describe the bug

`comment-no-empty` doesn't report on SCSS empty comment

`comment-no-empty`

e.g.

```scss
//

a {
  width: 00.10em;
}
```

e.g.

```json
{
  "rules": {
    "comment-no-empty: true
  }
}
```

e.g. `10.0.1`

CLI

SCSS

warning

No warning

vankop:

closes #4037

nothing

### DIFF
--- a/lib/rules/comment-no-empty/README.md
+++ b/lib/rules/comment-no-empty/README.md
@@ -8,6 +8,8 @@ Disallow empty comments.
  * Comments like this */
 ```
 
+This rule ignores SCSS-like comments.
+
 **Caveat:** Comments within *selector and value lists* are currently ignored.
 
 ## Options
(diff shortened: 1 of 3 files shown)


evilebottnawi:
I think it should be in `styleint-scss` as rule, not here

vankop:
@evilebottnawi close issue?

evilebottnawi:
@vankop let's wait answers from other developers, maybe we need only add docs about this limitation for rule

jeddy3:
> I think it should be in styleint-scss as rule, not here

Agreed. I think this should be handled within a `scss/comment-no-empty` plugin as [with other SCSS-like comment rules](https://github.com/kristerkari/stylelint-scss#-comment).

The built-in rules are geared towards standard syntax. They ignore non-standard constructs rather than include code to handle this.

> maybe we need only add docs about this limitation for rule

Yes, I think that is best:

"This rule ignores SCSS-like comments."

vankop:
@jeddy3 @evilebottnawi Please consider to make remarks if needed within editing file

PR #4225, opened 2019-08-24
Introduce unicode-bom rule

### LINKED ISSUE #3792: Add unicode-bom rule

The developers on my team use BOM inconsistently. This clutters the diffs.

A rule that requires that files always begin with the Unicode BOM character U+FEFF or that files must never begin with U+FEFF.

Inspired by https://eslint.org/docs/rules/unicode-bom

I tried to write such a rule myself, but I could not find any traces of BOM in the results of the parsing.

vkrol:

Closes #3792.

No, it's self explanatory.

I think it's better to add the auto-fix functionality separately, because first we need to fix the bug in stylelint that causes BOM to be cut out in any auto-fix run (https://github.com/stylelint/stylelint/issues/4228).

### DIFF
--- a/docs/user-guide/example-config.md
+++ b/docs/user-guide/example-config.md
@@ -170,6 +170,7 @@ You might want to learn a little about [how rules are named and how they work to
     "string-no-newline": true,
     "string-quotes": "single"|"double",
     "time-min-milliseconds": int,
+    "unicode-bom": "always"|"never",
     "unit-blacklist": string|[],
     "unit-case": "lower"|"upper",
     "unit-no-unknown": true,
--- a/lib/rules/unicode-bom/README.md
+++ b/lib/rules/unicode-bom/README.md
@@ -1,0 +1,37 @@
+# unicode-bom
+
+Require or disallow the Unicode Byte Order Mark.
+
+## Options
+
+`string`: `"always"|"never"`
+
+### `"always"`
+
+The following pattern is considered a violation:
+
+```css
+a {}
+```
+
+The following pattern is *not* considered a violation:
+
+```css
+U+FEFF
+a {}
+```
+
+### `"never"`
+
+The following pattern is considered a violation:
+
+```css
+U+FEFF
+a {}
+```
+
+The following pattern is *not* considered a violation:
+
+```css
+a {}
+```
--- a/lib/rules/unicode-bom/__tests__/index.js
+++ b/lib/rules/unicode-bom/__tests__/index.js
@@ -1,0 +1,152 @@
+"use strict";
+
+const rule = require("..");
+const { messages, ruleName } = rule;
+
+testRule(rule, {
+  ruleName,
+  config: ["always"],
+  skipBasicChecks: true,
+
+  accept: [
+    {
+      code: "\uFEFF",
+      description: "empty with BOM"
+    },
+    {
+      code: "\uFEFFa{}",
+      description: "with BOM"
+    }
+  ],
+
+  reject: [
+    {
+      code: "",
+      description: "empty without BOM",
+      message: messages.expected
+    },
+    {
+      code: "a{}",
+      description: "without BOM",
+      message: messages.expected
+    }
+  ]
+});
+
+testRule(rule, {
+  ruleName,
+  config: ["never"],
+  skipBasicChecks: true,
+
+  accept: [
+    {
+      code: "",
+      description: "empty without BOM"
+    },
+    {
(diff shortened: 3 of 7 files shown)


lib/rules/unicode-bom/__tests__/index.js
```diff
@@ -0,0 +1,64 @@
...
+      description: "empty with BOM",
+      message: messages.rejected
+    },
+    {
+      code: "\uFEFFa{}",
+      description: "with BOM",
+      message: messages.rejected
+    }
+  ]
+});
```
evilebottnawi: let's add couple test with `html` and `css-in-js`

vkrol: Done.

lib/rules/unicode-bom/index.js
```diff
@@ -0,0 +1,54 @@
...
+    const validOptions = validateOptions(result, ruleName, {
+      actual: expectation,
+      possible: ["always", "never"]
+    });
+
+    if (
+      !validOptions ||
+      root.source.inline ||
+      root.source.lang === "object-literal" ||
+      root.document !== undefined
```
vkrol: I have some doubts about this check, but I have not found another way to verify that it is not a HTML document.

lib/rules/unicode-bom/__tests__/index.js
vkrol: Done.

docs/user-guide/example-config.md
```diff
@@ -170,6 +170,7 @@ You might want to learn a little about [how rules are named and how they work to
     "string-no-newline": true,
     "string-quotes": "single"|"double",
     "time-min-milliseconds": int,
+    "unicode-bom": "never",
```
hudochenkov:
```suggestion
    "unicode-bom": "always"|"never",
```

lib/rules/unicode-bom/index.js
hudochenkov: Could you add a comment into the code explaining what this condition checks, please.

lib/rules/unicode-bom/index.js
```diff
@@ -0,0 +1,54 @@
...
+
+    if (expectation === "always" && !hasBOM) {
+      report({
+        result,
+        ruleName,
+        message: messages.expected,
+        root,
+        line: 1
+      });
+    } else if (expectation === "never" && hasBOM) {
```
hudochenkov:
```suggestion
    }

    if (expectation === "never" && hasBOM) {
```

It will increase readability.

vkrol:
@hudochenkov thanks! Done.

lib/rules/unicode-bom/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Require or disallow the Unicode Byte Order Mark.
+
+## Options
+
+`string`: `"always"|"never"`
+
+### `"always"`
+
+The following pattern is considered violation:
```
ntwb:
```suggestion
The following pattern is considered a violation:
```

lib/rules/unicode-bom/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+```css
+a {}
+```
+
+The following pattern is *not* considered violation:
```
ntwb:
```suggestion
The following pattern is *not* considered a violation:
```

lib/rules/unicode-bom/README.md
```diff
@@ -0,0 +1,37 @@
...
+### `"never"`
+
+The following pattern is considered violation:
+
+```css
+U+FEFF
+a {}
+```
+
+The following pattern is *not* considered violation:
```
ntwb:
```suggestion
The following pattern is *not* considered a violation:
```

lib/rules/unicode-bom/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+```css
+U+FEFF
+a {}
+```
+
+### `"never"`
+
+The following pattern is considered violation:
```
ntwb:
```suggestion
The following pattern is considered a violation:
```

vkrol:
@ntwb thanks, fixed.
> Ideally we update all instances of this where a single example is used

I can do this for all the rules in a separate PR.

hudochenkov:
> -   Added: `unicode-bom` rule ([#4225](https://github.com/stylelint/stylelint/pull/4225)).

Issue #4228, opened 2019-08-28
Fix autofix stripping BOM


vkrol:
> Clearly describe the bug

Autofix strips BOM

> Which rule, if any, is the bug related to?

N/A.

> What CSS is needed to reproduce the bug?

N/A.

> What stylelint configuration is needed to reproduce the bug?

N/A.

> Which version of stylelint are you using?

`10.1.0`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

CLI with `stylelint "**/*.css" --fix`.

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

No.

> What did you expect to happen?

The BOM must remain in place.

> What actually happened (e.g. what warnings or errors did you get)?

N/A.

Minimal reproducible example: https://github.com/vkrol/stylelint-fix-bom-bug.
This is related to https://github.com/stylelint/stylelint/issues/3792 and https://github.com/stylelint/stylelint/pull/4225.

vkrol:
@jeddy3 `status: wip` is most likely confusing because I was only writing the test, but I am not fixing this bug at the moment.

evilebottnawi:
I think we need option in postcss for this, becaue previously postcss always add BOM if it is exists in source file

ybiquitous:
The `unicode-bom` rule is now deprecated and will be removed in the future. See below:

https://github.com/stylelint/stylelint/blob/74d8da444d500ba69b15801c9165d32af54192f1/lib/rules/unicode-bom/README.md?plain=1#L3

PR #4229, opened 2019-08-29
Add the failing test for the stripping BOM bug

vkrol:

This is the failing test for #4228. 

No, it's self explanatory.

I am not sure that it is the right place for this test.

### DIFF
--- a/system-tests/fix/fix.test.js
+++ b/system-tests/fix/fix.test.js
@@ -135,3 +135,34 @@ describe("fix", () => {
       });
   });
 });
+
+describe("fix with BOM", () => {
+  let tmpDir;
+  let stylesheetPath;
+
+  beforeEach(() => {
+    tmpDir = os.tmpdir();
+    stylesheetPath = path.join(tmpDir, `stylesheet-with-bom.css`);
+
+    return cpFile(path.join(__dirname, "stylesheet.css"), stylesheetPath);
+  });
+
+  afterEach(() => {
+    return del(stylesheetPath, { force: true });
+  });
+
+  // eslint-disable-next-line jest/no-disabled-tests
+  it.skip("doesn't strip BOM", () => {
+    return stylelint
+      .lint({
+        files: [stylesheetPath],
+        configFile: systemTestUtils.caseConfig("fix"),
+        fix: true
+      })
+      .then(() => {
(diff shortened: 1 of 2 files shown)


system-tests/fix/fix.test.js
```diff
@@ -72,6 +72,21 @@ describe("fix", () => {
...
+        configFile: systemTestUtils.caseConfig("fix"),
+        fix: true
+      })
+      .then(() => {
+        return pify(fs.readFile)(stylesheetPath, "utf8").then(fileContent => {
+          expect(fileContent.startsWith("\uFEFF")).toBe(true);
+        });
+      });
+  });
+
```
evilebottnawi: Can you add new test case instead modify current

vkrol: By test case you mean a new CSS file?

evilebottnawi: Yes

vkrol: Do I need to create another folder under `system-tests` or the current solution will be appropriate?

system-tests/fix/fix.test.js
```diff
@@ -134,3 +134,34 @@ describe("fix", () => {
       });
   });
 });
+
+describe("fix with bom", () => {
```
hudochenkov:
```suggestion
describe("fix with BOM", () => {
```

vkrol: OK, thanks!

system-tests/fix/fix.test.js
```diff
@@ -134,3 +134,34 @@ describe("fix", () => {
...
+  // eslint-disable-next-line jest/no-disabled-tests
+  it.skip("doesn't strip BOM", () => {
+    return stylelint
+      .lint({
+        files: [stylesheetPath],
+        configFile: systemTestUtils.caseConfig("fix"),
+        fix: true
+      })
+      .then(() => {
+        return pify(fs.readFile)(stylesheetPath, "utf8").then(fileContent => {
```
vkrol: I need to replace `pify` here https://github.com/stylelint/stylelint/commit/14a0da663a52da10ca8df9a0183294e64e4f66fe.

vkrol: Fixed.

PR #4237, opened 2019-08-31
fix #3147

### LINKED ISSUE #3147: Display a violation at 1:1 for each file instead of throwing an error on unrecognised rules
Firstly I would like to say thanks for a great project, I have been introducing my team to stylelint on a couple of projects.

Feature request - don't throw an error on unrecognised rules.

It could be any, but as an example: `font-family-no-missing-generic-family-keyword`

Not applicable.

This should affect any configuration where a user is using an undefined rule.

Not really applicable, but in this case 7.x.x

CLI with `stylelint 'src/**/*.js` - I am using styled components, but this issue affects others.

No

Stylelint to run correctly and lint my files.

Stylelint fails to run with:

```shell
Error: Undefined rule font-family-no-missing-generic-family-keyword
```

The issue itself has been mentioned before, for example here #3067 - the answer is to use the correct version of stylelint, however there are times when that is not always easy, for example when using other tools which wrap stylelint.

In my case I am using Neutrino with their stylelint preset, which uses an older version of stylelint-webpack-plugin, which uses stylelint v7. I have asked the maintainers of neutrino to update their project, as I don't want to have to start locking down my version of stylelint-config-standard.

This could be avoided if instead of throwing an error and failing to run, stylelint ignored the rule, perhaps displaying a warning in the command line output?

Again, thanks for the project.

vankop:

fixes #3147

I add some ability to make suggestions to improve DX

### DIFF
--- a/lib/reportUnknownRuleNames.js
+++ b/lib/reportUnknownRuleNames.js
@@ -42,31 +42,30 @@ function extractSuggestions(ruleName) {
 function rejectMessage(ruleName, suggestions = []) {
   return (
     `Unknown rule ${ruleName}.` +
-    (suggestions.length > 0
-      ? ` May be you mean ${suggestions.join(", ")}.`
-      : "")
+    (suggestions.length > 0 ? ` Did you mean ${suggestions.join(", ")}?` : "")
   );
 }
 
 const cache = new Map();
 
 module.exports = function reportUnknownRuleNames(
-  unknownRules,
+  unknownRuleName,
   postcssRoot,
   postcssResult
 ) {
-  unknownRules.forEach(ruleName => {
-    const suggestions = cache.has(ruleName)
-      ? cache.get(ruleName)
-      : extractSuggestions(ruleName);
-    const warningProperties /*: Object*/ = {
-      severity: "error",
-      rule: ruleName,
-      node: postcssRoot,
-      index: 0
-    };
+  const suggestions = cache.has(unknownRuleName)
+    ? cache.get(unknownRuleName)
+    : extractSuggestions(unknownRuleName);
+  const warningProperties = {
+    severity: "error",
+    rule: unknownRuleName,
+    node: postcssRoot,
+    index: 0
+  };
 
-    cache.set(ruleName, suggestions);
-    postcssResult.warn(rejectMessage(ruleName, suggestions), warningProperties);
-  });
+  cache.set(unknownRuleName, suggestions);
+  postcssResult.warn(
+    rejectMessage(unknownRuleName, suggestions),
+    warningProperties
+  );
 };
--- a/lib/reportUnknownRuleNames.js
+++ b/lib/reportUnknownRuleNames.js
@@ -1,0 +1,67 @@
+"use strict";
+
+const levenshteinDistance = require("./utils/levenshteinDistance");
+const rules = require("./rules");
+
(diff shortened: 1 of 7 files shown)


lib/reportUnknownRuleNames.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const cache = new Map();
+
+module.exports = function reportUnknownRuleNames(
+  unknownRules,
+  postcssRoot,
+  postcssResult
+) {
+  unknownRules.forEach(ruleName => {
```
hudochenkov: I see `reportUnknownRuleNames()` used only in one place and it used as `reportUnknownRuleNames([ruleName], postcssRoot, postcssResult)`. Could we simplify and not use array and `forEach`, since we use only one rule always?

lib/reportUnknownRuleNames.js
```diff
@@ -0,0 +1,72 @@
...
+    const suggestions = cache.has(ruleName)
+      ? cache.get(ruleName)
+      : extractSuggestions(ruleName);
+    const warningProperties /*: Object*/ = {
```
hudochenkov: Just curious why `/*: Object*/` is here? For Flow? Could Flow understand that it's an object by itself in this case?

lib/utils/levenshteinDistance.js
```diff
@@ -0,0 +1,51 @@
...
+"use strict";
+
+/**
+ * Compute levenshtein distance, see https://en.wikipedia.org/wiki/Levenshtein_distance for more info
+ * @param {string} firstStr first string
+ * @param {string} secondStr second string
+ * @param {number} [maxDistance=99] maximum distance, if distance is greater then returns -1
+ * @returns {number} computed distance
+ */
+module.exports = function levenshteinDistance(
```
hudochenkov: Did you copy this function from somewhere? If yes, it would be nice to add source to the comment inside code. So in future we could reference the source.

lib/reportUnknownRuleNames.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  return result.slice(0, 3);
+}
+
+function rejectMessage(ruleName, suggestions = []) {
+  return (
+    `Unknown rule ${ruleName}.` +
+    (suggestions.length > 0
+      ? ` May be you mean ${suggestions.join(", ")}.`
```
hudochenkov:
```suggestion
      ? ` Did you mean ${suggestions.join(", ")}?`
```

lib/utils/levenshteinDistance.js
vankop: No, I implemented it myself. I provided link to wiki page, I think it is enough

lib/reportUnknownRuleNames.js
vankop: fixed

lib/reportUnknownRuleNames.js
vankop: mistake, it was copy/paste from `~/lib/utils/report.js`. Fixed

lib/reportUnknownRuleNames.js
vankop: done

hudochenkov:
> -   Changed: Display a violation at 1:1 for each file instead of throwing an error on unrecognised rules ([#4237](https://github.com/stylelint/stylelint/pull/4237)).

PR #4239, opened 2019-09-02
chore(deps): remove unused postcss-markdown

chyzwar:
I discovered that postcss-markdown was not used in the codebase. I might be wrong and it is used somewhere or is required for other reason.


hudochenkov:
It's used but it's not clearly visible:

https://github.com/stylelint/stylelint/blob/75326e9183e60be8141d4bef6f9dcfaeb2a2b777/lib/getPostcssResult.js#L75-L78

We have it to lint in markdown.

Issue #3499, opened 2018-07-25
Improve the release process to better handle OTP


jeddy3:
I was able to release `9.4.0` following @ntwb's [instructions](https://github.com/stylelint/stylelint/issues/3456#issuecomment-407595692).

It was a bit clunky because:

1. I had to, as per the instructions, manually delete and re-checkout the `.npmrc` file
2. I had to run `NPM_CONFIG_OTP=123456 npm run release` to trigger the prerelease checks, wait for publish step to fail because the OTP would expire, then run `NPM_CONFIG_OTP=987654 npm run release -- --skip-test --skip-cleanup` to publish before the new OTP expired.

Hopefully number 1 will be addressed by @hudochenkov's [PR to npmpub](https://github.com/stylelint/stylelint/issues/3456#issuecomment-407722043).

Maybe number 2 can also be addressed in npmpub, maybe by an interactive prompt that asks you for a OTP just before the publish step?

It'd be great to keep using npmpub for [these reasons](https://github.com/stylelint/stylelint/issues/3456#issuecomment-407672298).

Thoughts?

---

_Background: the stylelint npm package maintainers have turned on two-factor auth following [ESLint's postmortem recommendations](https://eslint.org/blog/2018/07/postmortem-for-malicious-package-publishes). So we now need to supply a one-time-password when publishing npm packages._

ntwb:
Thanks for persisting with what was for sure a frustrating process to release stylelint 9.4 @jeddy3 🥇 

npm are looking to add "time-limited tokens", it's an npm token generated that only lasts 5-10 minutes and generated using a users existing 2FA auth.

Details https://npm.community/t/time-limited-tokens/502/4

ps. Kudos to the npm team as they jumped on and proposed this idea pretty quickly after the ESLint security incident, I'm hoping this will arrive in npm soon 🤞

jeddy3:
> npm are looking to add "time-limited tokens"

That's good to know. The current process isn't too painful. So, I think we can wait to see if these time-limited tokens land before looking for an alternative solution.

ntwb:
|  | [npmpub](https://www.npmjs.com/package/npmpub) | [unleash](https://www.npmjs.com/package/unleash) | [np](https://www.npmjs.com/package/np) |
| --- | --- | --- | --- |
|Checks you are on the `master` branch:|  ? |? |Yes|
|Checks the working directory is "clean":| Yes| ?| Yes|
[...]

PR #4243, opened 2019-09-03
Use np for publishing

hudochenkov:

Closes https://github.com/stylelint/stylelint/issues/3499

I used `np` for a few recent stylelint and stylelint-order releases. Works pretty good.


ntwb:
Nice, the steps to update the release notes, easy enough you think?

We could still use the https://www.npmjs.com/package/github-release-from-changelog package which is what performs this functionality from `npmpub`?

hudochenkov:
I don't think it supports our changelog structure: https://github.com/MoOx/github-release-from-changelog/blob/948691c6110cf1166c61e23ec7bf316c6fa1a1dd/github-release-from-changelog.js#L123-L145 We have `## d.d.d` format for version, but it supports only `## [d.d.d]`.

ntwb:
We've been using it for years, so I guess it would work, whether that's worth further investigation or just performing as you've suggested in the PR to updating the release notes we can investigate at a later date.

Issue #4244, opened 2019-09-04
Replace Greenkeeper


hudochenkov:
Greenkeeper sends a lot of messages and PR. Most of the time they are not really usefull. There is no real need to update every single dependency to it's every version. Greenkeeper open new PRs even for minor versions, and creates more distraction, rather real help.

I propose to reconsider our choice and choose different tool for updating dependencies.

What I think are good features:

1. Create PRs for major versions and 0.x versions. Because versions could have breaking changes. And breaking changes are easier to fix as soon as possible, rather postpone.
2. Create PRs for vulnerabilities.
3. Don't create PRs for version withing semver range specified in package.json. Or at least create PRs not as soon as possible, but on a schedule, like once per month.
4. Notify if dependency withing semver range breaks our build.
5. Nice to have: able to ignore packages. E. g. Flow is the most annoying. They update way to often, and because they still on 0.x versioning, every version considered a major (could have breaking changes).

We could try https://dependabot.com/ (which is bought by Github). These features could cover Greenkeeper's shortcomings:

* [`ignored_updates`](https://dependabot.com/docs/config-file/#ignored_updates)
* [`version_requirement_updates: widen_ranges`](https://dependabot.com/docs/config-file/#version_requirement_updates) or `increase_versions_if_necessary`

Also, very nice feature is [`automerged_updates`](https://dependabot.com/docs/config-file/#automerged_updates).

And we could manually check/update dependencies before releases. We don't release that often, so shouldn't be a burden.

jeddy3:
> We could try https://dependabot.com/ (which is bought by Github).

SGTM

ntwb:
Likewise, SGTM, happy to give Depanabot a try.

Greenkeeper creates a new branch for each package release, it runs the CI tests and if the CI fails it creates a PR, or creates a PR if it is not covered by the SemVer range, otherwise it then deletes the branch.

I think we've been caught a little of guard recently by the above type PRs because for example the `flow-bin` PRs had not been resolved Greenkeeper would continue commenting on existing PRs and sometimes creating new PRs, which is noisy as noted.

Being able to _ignore updates_ would be nice for issues where we have less experience to address some of these _flow_ or _Windows glob_ issues.

hudochenkov:
Dependabot is cool. I closed PR and it [reacted](https://github.com/stylelint/stylelint/pull/4251#issuecomment-528267042):

[...]

PR #4253, opened 2019-09-05
Add Dependabot config

hudochenkov:

https://github.com/stylelint/stylelint/issues/4244

https://dependabot.com/docs/config-file/


hudochenkov:
At first I put all options in the config, but then decided to keep only options, which are not available in web UI of Dependabot. This will keep things simple and in one place for all our repos.

PR #4268, opened 2019-09-09
Explicitly specify syntax module name

hudochenkov:

Avoid problems like https://github.com/stylelint/stylelint/pull/4264 and https://github.com/stylelint/stylelint/pull/4239.

Now all syntaxes could be found in the code by searching by package name.


vankop:
Totally agree with suggestion use explicit `require` ( if it possible ) since it allows stylelint developers to find module usage by searching module name

PR #4272, opened 2019-09-10
feat: add v8-compile-cache

vankop:

inspired by https://github.com/eslint/eslint/pull/11921 =)

nothing

### DIFF
--- a/bin/stylelint.js
+++ b/bin/stylelint.js
@@ -2,4 +2,7 @@
 
 "use strict";
 
+// to use V8's code cache to speed up instantiation time
+require("v8-compile-cache");
+
 require("../lib/cli")(process.argv.slice(2));
--- a/package.json
+++ b/package.json
@@ -85,7 +85,8 @@
     "style-search": "^0.1.0",
     "sugarss": "^2.0.0",
     "svg-tags": "^1.0.0",
-    "table": "^5.2.3"
+    "table": "^5.2.3",
+    "v8-compile-cache": "^2.1.0"
   },
   "devDependencies": {
     "benchmark": "^2.1.4",


jeddy3:
It would be good to get some before and after benchmarks ([like so](https://github.com/stylelint/stylelint/issues/2454#issuecomment-292794000)), so we know the benefit of this change.

vankop:
Don't how it works exactly =)
For me 
```
time node -e "require('./bin/stylelint')"
```
after several runs `0.2s`    

For code in `master` after several runs `0.23s`
For latest npm package `0.3s`

evilebottnawi:
Try to run `stylelint` multiple times with new `require` and without

vankop:
@evilebottnawi something like that
```javascript
function measureRequire() {
  const t = process.hrtime();
  require("./bin/stylelint");
  console.log('stylelint: ' + process.hrtime(t)[1] / 1e+6);
}

const retry = fn => times => new Array(times).fill(0).forEach((_, i) => fn(i));

retry(measureRequire)(10);
```
???

evilebottnawi:
Here example https://github.com/zertosh/v8-compile-cache/blob/master/bench/require-babel-core.js

vankop:
Снимок экрана 2019-09-10 в 19 35 23

jeddy3:
Are those the results from the PR branch or master?

evilebottnawi:
@vankop just create suite WITH_CACHE (where true) and WITH_CACHE (where unspecific)

vankop:
Sorry, did not understand at the beginning that I need to require code without `v8-compile-cache`

Снимок экрана 2019-09-10 в 19 56 15

evilebottnawi:
Good perf on first run

jeddy3:
Yep, better on first run. LGTM.

PR #4277, opened 2019-09-11
use Set instead of Array in property-no-unknown

vankop:

minor rule performance fix

nothing


hudochenkov:
Please provide before/after results https://stylelint.io/developer-guide/rules#improving-the-performance-of-a-rule

vankop:
> Please provide before/after results https://stylelint.io/developer-guide/rules#improving-the-performance-of-a-rule

Снимок экрана 2019-09-11 в 19 32 13

PR #4278, opened 2019-09-11
Small package loading improvements

vankop:

https://github.com/stylelint/stylelint/issues/2454

master: https://gist.github.com/vankop/fd99b23801a8117f0fc0562cc71a0261

PR: https://gist.github.com/vankop/38a5e597090cd967e40a02124349dddc

### DIFF
--- a/lib/requireRule.js
+++ b/lib/requireRule.js
@@ -3,15 +3,10 @@
 const importLazy = require("import-lazy")(require);
 const rules = require("./rules");
 
-function requireRuleUnsafe(ruleName) {
-  return importLazy(`./rules/${ruleName}`);
-}
-
-module.exports = function requireRuleSafe(ruleName) {
+module.exports = function(ruleName) {
   if (rules.includes(ruleName)) {
-    return requireRuleUnsafe(ruleName);
+    return importLazy(`./rules/${ruleName}`);
   } else {
     return false;
   }
 };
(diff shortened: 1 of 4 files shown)


lib/requireRule.js
```diff
@@ -3,10 +3,15 @@
 const importLazy = require("import-lazy")(require);
 const rules = require("./rules");
 
-module.exports = function(ruleName) {
+function requireRuleUnsafe(ruleName) {
```
hudochenkov: What is the reason to make this function?

vankop: After looking how currently we load lazily rules in our public API I found out that there is unnecessary check [here](https://github.com/stylelint/stylelint/blob/master/lib/index.js#L19), so we can omit it using `unsafe` require. Theoretically this change reduce code complexity by `O(n * x)`, where `n` is amount of rules and `x` is complexity to check entry in `Set`. We don't change `Set` between comparisons and amount of rules is not much(likelihood of collisions is very small), so in real world I think one check takes not much times (this change was the least valued I think)

hudochenkov: Thank you!

I think we need to rename unsafe for something less scary. Usually “unsafe” things are experimental or fragile, which is not the case for us.

I can't come up with a better name. Which makes me think. Is change to requireRule has real benefits. It make code harder to understand. 1-2ms improvement not worth to make code more complex.

PR #4259, opened 2019-09-06
New code formatting settings

hudochenkov:
> Which issue, if any, is this issue related to?

https://github.com/stylelint/stylelint/issues/4246

> Is there anything in the PR that needs further explanation?

Here is configuration only. I run Prettier with this settings and here the output: https://github.com/stylelint/stylelint/tree/new-code-formatting-output

Compare: https://github.com/stylelint/stylelint/compare/new-formatting-configuration...new-code-formatting-output

PR #4280, opened 2019-09-15
Use spaces for *.md, *.css and package.json formatting

hudochenkov:

Follow-up to https://github.com/stylelint/stylelint/pull/4259.

I've configured `*.md`, `*.css` and `package.json` to use spaces, so these files are not be broken by new configuration. They use spaces and in many places it's critical for alignment and/or correct test results (like system tests).

ESLint config change is to correct parsing trailing comma in function arguments.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,3 +7,11 @@ end_of_line = lf
 indent_style = tab
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[package.json]
+indent_style = space
+indent_size = 2
+
+[*.{md,css}]
+indent_style = space
+indent_size = 2
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -7,4 +7,13 @@ module.exports = {
 	singleQuote: true,
 	trailingComma: 'all',
 	useTabs: true,
+	overrides: [
+		{
+			files: ['package.json', '*.md', '*.css'],
+			options: {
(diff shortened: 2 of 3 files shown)


hudochenkov:
I'm going to merge as it kind of overlooked fix for https://github.com/stylelint/stylelint/pull/4259.

ntwb:
I don't believe this should be required, I use tabs for .md, .json, and .css everyday https://github.com/WordPress/wordpress-develop/blob/master/.editorconfig

.md and .json changes to tabs in this repo should not be an issue.

.css changes may require us updating some of the tests, I'm kind of surprised by this anyway, our tests should be neutral in that regard as users do use tabs and spaces and our tests should be able to facilitate that.

I'll take a closer look at this during the week...

hudochenkov:
We use spaces to correctly align arrow in examples for rules. E. g. https://github.com/stylelint/stylelint/blob/master/lib/rules/at-rule-name-newline-after/README.md After rule description there is usually an example with arrow. If we convert this to tabs, then we need to verify every rule readme.

Also, in VS Code, for example, now if I go to a file with spaces as indentation it will use tabs when pressing Tab. Because tab is specified in EditorConfig, and VS Code ignores `"editor.detectIndentation": true` for some reason. So it safer to explicitly specify spaces for files we can't easily convert to tabs at the moment.

Issue #4273, opened 2019-09-10
Use GitHub Actions


jeddy3:
We've recently discussed (and implemented) a few changes to improve our experience maintaining stylelint: https://github.com/stylelint/stylelint/issues/4244, https://github.com/stylelint/stylelint/issues/3499, #4245 etc.

Another area of possible improvement is continuous integration, if only because Appveyor has been troublesome recently.

[GitHub Actions](https://github.com/features/actions) could help us here. We could test on multiple version of node across windows, nix and mac within GitHub [as detailed in this blog post](https://github.blog/2019-08-08-github-actions-now-supports-ci-cd/).

(Later down the line, we could look into things [like publishing to npm via Actions](https://medium.com/faun/automate-your-npm-publish-with-github-actions-dfe8059645dd) too.)

---

I've added stylelint to the [GitHub Actions waiting list](https://github.com/features/actions/signup?account=stylelint). Hopefully, it'll be processed soon and we can start investigating. (I had originally signed up with my personal account as I hadn't realised that wouldn't also be for stylelint.)

ybiquitous:
Recently I tried GitHub Actions. Very fast. 👍 
(Perhaps the back-end may be Azure Pipelines)

jeddy3:
> Recently I tried GitHub Actions. Very fast.

That's good to hear!

> But I feel Travis is easier than GitHub Actions for the YAML settings.

I believe GitHub Actions now use YAML. From the [blog post](https://github.blog/2019-08-08-github-actions-now-supports-ci-cd/) mentioned [above](https://github.com/stylelint/stylelint/issues/4273#issue-491732753), this is all we need to test stylelint on 3 node version across 3 operating systems:

```yaml
jobs:
  test:
    name: Test on node ${{ matrix.node_version }} and ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        node_version: [8, 10, 12]
        os: [ubuntu-latest, windows-latest, macos-latest]

    steps:
    - uses: actions/checkout@v1

    - name: Use Node.js ${{ matrix.node_version }}
      uses: actions/setup-node@v1
      with:
        version: ${{ matrix.node_version }}
    
    - name: npm install, build and test
      run: |
        npm install
        npm run build --if-present
        npm test
```

ybiquitous:
> this is all we need to test stylelint on 3 node version across 3 operating systems:

Yes, this is super easy! 😄

[...]

PR #4285, opened 2019-09-16
Add Github Actions for testing

hudochenkov:

https://github.com/stylelint/stylelint/issues/4273

It's a mix of https://github.com/stylelint/stylelint/issues/4273#issuecomment-529991804 and what Github gives by default for a Node.js project.

It need more work to migrate all features from [`.travis.yml`](https://github.com/stylelint/stylelint/blob/master/.travis.yml).


hudochenkov:
I'm impressed by speed of Github CI. In 5m 18s it done 10 jobs (3 Node version * 3 OS + 1 lint job), while neither Appveyor (2 jobs) nor Travis (4 jobs) is done.

hudochenkov:
There is currently a [bug](https://github.community/t5/GitHub-Actions/Expression-in-step-name-not-being-evaluated/td-p/30652) with a step name:

Screenshot 2019-09-17 at 00 57 04

hudochenkov:
To-do:

- [ ] We want to run CI on PRs (created PR, push to PR, force push). On master. And on `^dependabot/.*$` branches.
https://github.com/stylelint/stylelint/blob/c69e098b0a9590bdebdf9b8d8605bcf443078eee/.travis.yml#L6-L9
- [ ] Preferably to use the latest npm version for every OS/Node.js
https://github.com/stylelint/stylelint/blob/c69e098b0a9590bdebdf9b8d8605bcf443078eee/.travis.yml#L35
- [ ] Add coveralls https://github.com/marketplace/actions/coveralls-github-action. We have it implemented [like this](https://github.com/stylelint/stylelint/blob/c69e098b0a9590bdebdf9b8d8605bcf443078eee/.travis.yml#L38) for Travis, which doesn't work with Github Actions.

I don't know why we use following `env`: https://github.com/stylelint/stylelint/blob/c69e098b0a9590bdebdf9b8d8605bcf443078eee/.travis.yml#L30-L32

I don't know if we need this: https://github.com/stylelint/stylelint/blob/c69e098b0a9590bdebdf9b8d8605bcf443078eee/.travis.yml#L40-L42

@digitaljohn is it something you can help us?

jeddy3:
> We want to run CI on PRs (created PR, push to PR, force push). On master. And on

Yes, please.

> Preferably to use the latest npm version for every OS/Node.js

Yep, we'll want this as we only support the latest npm version.

> Add coveralls https://github.com/marketplace/actions/coveralls-github-action.
 
Nice

The Gitter hook can go. It's from a time when we used Gitter for support.

I can't remember what the following is for:

```
env: CI=tests 12 
```

What happens when it goes?

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,82 @@
...
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+
+    - name: npm install, and test
+      run: |
+        npm install
+        npm run jest --coverage
+      env:
+        CI: true
```
vankop: As I see `CI` variable does not mean anything, right now there is only one way -  pass `--ci` flag to jest. I could be wrong, will be great, if you point me out where it uses

There is PR about `Github Actions`, see https://github.com/watson/ci-info/pull/42 - `ci-info` used in `jest` to determine is current environment CI server/runner/worker

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,82 @@
...
+        CI: pretest
+
+  test:
+    name: Test on Node.js ${{ matrix.node-version }} and ${{ matrix.os }}
+
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        node-version: [8, 10, 12]
```
vankop: could you check will our test work with `*-alpine` node, it extremely reduce image size
Снимок экрана 2019-09-17 в 12 22 39

You can check all available images [here](https://hub.docker.com/_/node)

vankop: Currently this affect only `node-8` since `node-8` and `node-12` are cached on github actions servers, see https://help.github.com/en/articles/software-in-virtual-environments-for-github-actions

hudochenkov:
I noticed that all Windows jobs doesn't run Jest for some reason :( Only `npm install`.

.github/workflows/nodejs.yml
vankop: Oh it does not work this way =) `setup-node` install node using `node-12` 😎

vankop:
> I noticed that all Windows jobs doesn't run Jest for some reason :( Only `npm install`.

Do you have any log?

hudochenkov:
https://github.com/stylelint/stylelint/pull/4285/checks?check_run_id=224510457

vankop:
Straightforward idea can you write `npm install && npm run jest` ?

vankop:
> * We want to run CI on PRs (created PR, push to PR, force push). On master. And on `^dependabot/.*$` branches.

```
on:
  push:
    branches:    
      - master                # Push events on master branch
      - 'dependabot/*'  # Push events on Dependabot branches
    paths:
      - '!*.md'
 pull_request:
   paths:
      - '!*.md'
```

Hope it is enough

hudochenkov:
@vankop feel free to work in this branch :)

vankop:
I think we can use `maxWorkers=2` for `jest` since there is only 2 core CPUs

```
Supported virtual environments and hardware resources
Each virtual machine has the same hardware resources available.

2 core CPUs
7 GB of RAM memory
14 GB of SSD disk space
```
https://help.github.com/en/articles/virtual-environments-for-github-actions#supported-virtual-environments-and-hardware-resources

hudochenkov:
> I think we can use maxWorkers=2 for jest since there is only 2 core CPUs

Will it improve speed of tests? I think we used maxWorkers before because Travis or Jest had problems to detect number of CPUs.

vankop:
> Will it improve speed of tests?

It is a tricky question it depends on how much workers jest will spawn without this option (as well as github doc does not say anything about threads,  cores could support `hyper-threading`).

hudochenkov:
It's kind of easy question. Run tests without these option, and with, then compare time :)

vankop:
> It's kind of easy question. Run tests without these option, and with, then compare time :)

Not at all, infrastructure could be inconsistent, so on run `A` job `N` could run with one pair of cores, on run `B` with another one.

Anyway, I took a look in jest code, they use just [cpu amount](https://github.com/facebook/jest/blob/b7cb5221bb06b6fe63c1a5e725ddbc1aaa82d306/packages/jest-runtime/src/cli/index.ts#L81), so I think this option is not necessary

vankop:
Surprised how big is github actions ecosystem already. There is implementation for jest reporter already https://github.com/rkusa/jest-action 

It does not fit our requirements as I see =( However, anyone can use this work as start point =)

.github/workflows/nodejs.yml
hudochenkov: Try to remove it.

vankop:
@hudochenkov now it works for windows as well

hudochenkov:
It's nice! I wonder why tests on Windows fail. Are the same tests a passing on Appveyor?

vankop:
@hudochenkov could you check is `coveralls` works, please?

hudochenkov:
I just realized that we don't use coveralls. We have it, but we have not badge or link in a readme. And to be honest, I don't to see comments like https://github.com/stylelint/stylelint/pull/4285#issuecomment-532320191 on Github and in my email. There are not helpful.

hudochenkov:
About using latest npm for all builds: https://github.com/actions/setup-node/issues/31. It might not work, but we could try.

hudochenkov:
The last question to answer is to whether we need coveralls or not https://github.com/stylelint/stylelint/pull/4285#issuecomment-532352382. Other things are looking good! And we achieved everything we wanted and even more (like tests on macOS and Windows).

hudochenkov:
One more question to decide what are we going to do with Travis and Appveyor? I don't want to have them running with Github Actions at the same time. We don't need them anymore. When config files removed, then we have this results:

Screenshot 2019-09-20 at 01 29 27

I don't see a solution to keep using Travis and Appveyor for previously opened PRs, but not use them for future PRs and master. I propose disable them right before we merge this PR, and all previous PRs would need to be rebased on master. All PRs should be rebased anyway, because we changed formatting recently.

hudochenkov:
Tests still randomly fail on this test:

Screenshot 2019-09-20 at 01 40 17

But it's outside the scope of this PR. Restart helps.

ntwb:
> I don't see a solution to keep using Travis and Appveyor for previously opened PRs, but not use them for future PRs and master. I propose disable them right before we merge this PR, and all previous PRs would need to be rebased on master. All PRs should be rebased anyway, because we changed formatting recently.

Agreed 👍🏼

I'd like to keep coveralls if we can, I think its still a useful tool for us

system-tests/outputFile/outputFile.test.js
```diff
@@ -43,7 +43,7 @@ describe('outputFile', () => {
 		childProcess.on('close', function() {
 			return readFileAsync(directionPath, 'utf-8').then((content) => {
 				expect(content).toEqual(systemTestUtils.stripColors(stdout));
-				expect(content).toMatchSnapshot();
+				expect(content.replace('×', '✖')).toMatchSnapshot();
```
vankop: @hudochenkov I am curious about this fix =) could you explain where charset is wrong?

hudochenkov: I have no idea what happened. But Jest was telling that two lines are different (Windows output and snapshot), and this symbol was the difference.

hudochenkov:
> I'd like to keep coveralls if we can, I think its still a useful tool for us

If anyone uses it for stylelint? :) If not, then regardless of how useful it is, we can remove it.

system-tests/outputFile/outputFile.test.js
vankop: Looks like we need to use [this util](https://github.com/stylelint/stylelint/blob/master/lib/formatters/__tests__/prepareFormatterOutput.js) here

ntwb:
> If anyone uses it for stylelint? :) If not, then regardless of how useful it is, we can remove it.

I do partially, so even though we no longer have the coveralls badge on the readme I pipe the results into the Slack channel, I think it may still be configured to report on PRs if a PR changes the coverage by more than 5%, I think its a good safeguard to keep track of any potential issues we may run into by granting commit to any user who contributes.

image

hudochenkov:
Make sense. Is it possible to configure to not post a comment in every PR, but only after certain threshold?

ntwb:
> Make sense. Is it possible to configure to not post a comment in every PR, but only after certain threshold?

That's the current behaviour I belive, it will only post a comment if the coverage changes by ~5%


p.s I just noticed we've hit `7,005` GitHub stars ⭐

system-tests/outputFile/outputFile.test.js
hudochenkov: Too much trouble, in my opinion. We would need to move this utility somewhere, then change this test, and test that used this utility. I think it's ok to keep things as they are for less complex codebase.

hudochenkov:
This PR is ready for approvals :)

PR #4286, opened 2019-09-17
Add --noglob cli option

### LINKED ISSUE #4193: Allow to use full file path without converting it to glob

AzazKamaz:
Closes #4193 and is non-breaking alternative for #4254 that can be accepted in minor release.
Tests and flow types have been implemented

### DIFF
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -170,7 +170,7 @@ module.exports = function(
 		fileList = [fileList];
 	}
 
-	if (!options.disableDefaultIgnores) {
+	if (!options.disableDefaultIgnores && !options.noglob) {
 		fileList = fileList.concat(ALWAYS_IGNORED_GLOBS.map((glob) => '!' + glob));
 	}
 
@@ -186,7 +186,7 @@ module.exports = function(
 		fileCache.destroy();
 	}
 
-	return globby(fileList, globbyOptions)
+	return (options.noglob ? Promise.resolve(fileList) : globby(fileList, globbyOptions))
 		.then((filePaths) => {
 			// The ignorer filter needs to check paths relative to cwd
 			filePaths = filterFilePaths(ignorer, filePaths.map((p) => path.relative(process.cwd(), p)));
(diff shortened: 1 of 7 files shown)


system-tests/cli/cli.test.js
```diff
@@ -114,4 +114,46 @@ describe('CLI', () => {
 			);
 		});
 	});
+
+	it('--noglob-enabled-right', () => {
```
vankop: Thanks for PR, it really useful feature. Could you please rename test cases without additional `-` and make more informative description. For example
```
describe('--no-glob', () => {
     it('enabled. file exists');
     it('enabled. file does not exist');
     it('disabled.  file exists');
     it('disabled. file does not exist');
})
```
```

**Comment by evilebottnawi** [comment]
I am :-1: on this, solution should be https://github.com/stylelint/stylelint/issues/4193#issuecomment-520214141, not new option for CLI

File: lib/standalone.js [hunk 2]
```diff
@@ -170,7 +170,7 @@ module.exports = function(
 		fileList = [fileList];
 	}
 
-	if (!options.disableDefaultIgnores) {
+	if (!options.disableDefaultIgnores && !options.noglob) {
```
evilebottnawi: In this case, `ALWAYS_IGNORED_GLOBS` is not working as expected with `noglob`, it is not good idea

vankop:
@evilebottnawi So, if I understand you correctly, you suggest to use kind of heuristic:
- if there is a file in file system using `options.files` as relative path, then we just lint this file
- otherwise expect globby pattern

or there is another idea?

vankop:
eslint use [this code](https://github.com/eslint/eslint/blob/d08683e3c807f92daf266894093c70f8d5ac6afa/lib/cli-engine/file-enumerator.js#L292)

evilebottnawi:
Code from eslint looks good, i think we should use this logic too

vankop:
Any other suggestions?

PR #4291, opened 2019-09-20
Add link to stylelint-scales plugin pack

digitaljohn:

n/a

Adding `stylelint-scales` to the list of stylelint plugins.


vankop:
Looks like you need to add namespace `@signal-noise/stylelint-scales` or you going to take `stylelint-scales` package name?

digitaljohn:
@vankop @jeddy3 This has been updated, thank you for pointing this out.

vankop:
Why you don't want to take `stylelint-scales` ?

jeddy3:
> Why you don't want to take stylelint-scales ?

I believe things are shifting towards scoped packages. For example, GitHub new [package registry](https://github.com/features/package-registry) only supports scoped packages at the moment.

I think we might as well go with the trend and start introducing scoped packages to stylelint plugins.

vankop:
> we use [Lerna](https://lerna.js.org/) to manage everything in the single repo that has ~60 packages with deep dependencies on each other

Did you try [rush][https://rushjs.io/] - solution by Microsoft?

PR #4292, opened 2019-09-23
Add GitHub Action to run stylelint with reviewdog

haya14busa:
Hi!
I created GitHub Action to run stylelint with [reviewdog](https://github.com/reviewdog/reviewdog) which can posts review comments on GitHub PullRequests.

https://github.com/marketplace/actions/run-stylelint-with-reviewdog

In case you're interested in, I created this pull-request. Please feel free to close it if it's not worth mentioning this action.


haya14busa:
Thanks!

>  maybe it will be better to use stylelint Node.js API instead of cli + jq just to be sure that output will be correct

Makes sense. It was just I'm not familiar with Node.js API of stylelint and one of the good points of reviewdog is that the users (including me!) don't need to write specific code for each linter.
But, it's public action shared to everyone, so it's worth using API.
I'm happy to accept pull requests.

PR #1392, opened 2016-06-04
Don't remove feature name from error message

epmatsw:
Fixes #1388. Removing the name makes the message a bit nicer, but it makes adding ignore rules tough (as mentioned in the issue).

PR #4294, opened 2019-09-23
Reconfigure CI

hudochenkov:
We need to run `npm run lint:md` on `*.md` files. I don't know how to configure Github Actions for this. At least we could use old approach and run all tests.


vankop:
We can use different workflow for this =)

hudochenkov:
@vankop is it possible to disable coveralls comment? It reports result to status checks, comment is redundant.

image

hudochenkov:
> We can use different workflow for this =)

We could. But is it worth it to complicate CI for this?

vankop:
> > We can use different workflow for this =)
> 
> We could. But is it worth it to complicate CI for this?

I can suggest another way - 2 flows:
1. tests
2. lint

So lint step will be always triggering

vankop:
> is it possible to disable coveralls comment?

@hudochenkov As I understand from code and job output, this must be a coverall setting. All what code does just creating so called `job`, e.g. current job https://coveralls.io/jobs/53558061 . However, I did not figure out how coverall commit owner authenticated as `github-actions`

hudochenkov:
I checked coveralls settings. And comments were disabled. I redisabled them just in case.

ntwb:
Coveralls comments report as the same as “our” comments, as such Coveralls doesn’t currently comment, it reports status checks which is what we do want to continue, Coveralls will only leave a comment of the percentage test coverage changes by the threshold we have set, which is -5%

ntwb:
This is a Coveralls comment example before we disabled them:
> https://github.com/stylelint/stylelint/pull/1392#issuecomment-223771324

vankop:
Problem is that in your example comment authored by `coveralls`, now it authored by `github-actions`

hudochenkov:
I extracted linting into separate workflow. Also renamed few things so now in statuses text is more concise and we see time for every item (almost).

hudochenkov:
I see one flaw with two workflows. We can't enforce all tests for PR to be merged. Before Github Action we set Travis and Appveyor as a requirement. Github would not allow contributors to merge until tests are green. With two workflows we can't enabled requirements for every job (which we have 10 at the moment), because some of jobs (9 of them) just wouldn't run on PRs with `*.md` changes.

We have two choices:

1. Require only Lint job, and keep eye on build statuses when we merge.
2. Don't make exceptions for `*.md`and run all tests for all PRs.

I would go with the second option for more security. We don't have much PRs with just `*.md` changes. And kind of unnecessary 3-4 minutes of waiting for CI won't make any difference, because we do not react to a PR at the very moment it posted.

evilebottnawi:
:+1: on second

ntwb:
Agreed, on the second option, as we've such a large number of committers who can merge a PR lets ensure that we mitigate issues this may bring stylelint as best as possible 👍🏼

hudochenkov:
Ready for a review.

vankop:
@hudochenkov [fastify](https://github.com/fastify/fastify) faced with same `coveralls` problem, so I hope `coveralls` team will investigate this problem soon

XhmikosR:
I contacted Coveralls via email about the comments issue, we'll see how it goes. It's quite annoying, especially for popular repos with many watchers.

PR #4295, opened 2019-09-23
Switch to GitHub Actions badge and remove Travis CI & AppVeyor badges

ntwb:
See https://help.github.com/en/articles/configuring-a-workflow


hudochenkov:
The other question is do we really need this badge? What problem does it solve?

evilebottnawi:
Shows that the project is maintained and developed

hudochenkov:
Well, I could commit something and don't change anything for 5 years. And it will still show the last result :)

evilebottnawi:
@hudochenkov yep, just common practice for many repositories (eslint has badges also). Don't think it is real problem

hudochenkov:
I'm just questioning things that we were doing for a long time. If they still valid or we are doing them just because we were doing them before :)

vankop:
Actually build status is pretty important. 

> Well, I could commit something and don't change anything for 5 years. And it will still show the last result :)

In real world you will look on the releases and commits activity as well

Issue #4303, opened 2019-09-24
Improve test stability


hudochenkov:
There are some tests that fail occasionally on Windows. It doesn't happen every time and not the same tests (it's always one of mentioned below).

First let's gather what failing tests we get overtime, and then find the problem and fix it.

Below will be mentioned Node.js version where test failed, and a link to result if possible. I don't think failures depend on Node.js version, though.

Node.js 12:

```
FAIL lib/__tests__/ignore.test.js
  ● using ignoreFiles with input files that would cause a postcss syntax error › syntax-error-ignored.scss found

    expect(received).not.toBe(expected) // Object.is equality

    Expected: not -1

      313 | 		const index = Number(results[1].source.indexOf('syntax-error-ignored.scss'));
      314 | 
    > 315 | 		expect(index).not.toBe(-1);
          | 		                  ^
      316 | 	});
      317 | 
      318 | 	it('syntax-error-ignored.scss not linted', () => {

      at Object.toBe (lib/__tests__/ignore.test.js:315:21)
```

[Node.js 8](https://github.com/stylelint/stylelint/runs/233974997):

```
FAIL lib/__tests__/ignore.test.js
  ● using ignoreFiles with input files that would cause a postcss syntax error › no-syntax-error.css found

    expect(received).not.toBe(expected) // Object.is equality

    Expected: not -1

      302 | 
      303 | 	it('no-syntax-error.css found', () => {
    > 304 | 		expect(results[0].source.indexOf('no-syntax-error.css')).not.toBe(-1);
          | 		                                                             ^
      305 | 	});
      306 | 
      307 | 	it('no-syntax-error.css linted', () => {

      at Object.toBe (lib/__tests__/ignore.test.js:304:64)
```

Node.js 12:

```
FAIL lib/__tests__/standalone-syntax.test.js
  ● standalone with postcss-safe-parser

    expect(received).toBeFalsy()

    Received: true

      305 | 					const root = result._postcssResult.root;
      306 | 
    > 307 | 					expect(results[0].errored).toBeFalsy();
          | 					                           ^
      308 | 					expect(results[0].warnings).toHaveLength(0);
      309 | 					expect(root.toString()).not.toBe(root.source.input.css);
      310 | 

      at toBeFalsy (lib/__tests__/standalone-syntax.test.js:307:33)
          at Array.map (<anonymous>)
      at map (lib/__tests__/standalone-syntax.test.js:303:12)
```

Node.js 8:

```
[...]

PR #4305, opened 2019-09-24
Test

hudochenkov:
It's a test to see if CI kick in. Related to https://github.com/stylelint/stylelint/pull/4304

### DIFF
--- a/lib/__tests__/fixtures/standaloneNoParsing/no-syntax-error.css
+++ b/lib/__tests__/fixtures/standaloneNoParsing/no-syntax-error.css
@@ -1,1 +1,1 @@
-/* This will not cause a CSS syntax error */
+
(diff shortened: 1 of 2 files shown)


hudochenkov:
I think @evilebottnawi [could be right](https://github.com/stylelint/stylelint/issues/4303#issuecomment-534508093) and we need to rewrite tests to avoid possible race conditions.

E. g. remove `beforeAll` because each run of beforeAll will give the same result.

vankop:
Is [this problem](https://github.com/stylelint/stylelint/pull/4305/checks?check_run_id=234357165) of github-actions?

hudochenkov:
No. It cancels other jobs if one fails. Check https://github.com/stylelint/stylelint/pull/4305/checks?check_run_id=234357218 for real reason.

PR #4315, opened 2019-10-01
Add "css" option to --syntax parameter

hudochenkov:

Needed to fix https://github.com/stylelint/stylelint-demo/issues/155.

Currently there is no way to specify “css” syntax.

### DIFF
--- a/lib/__tests__/standalone.test.js
+++ b/lib/__tests__/standalone.test.js
@@ -227,7 +227,7 @@ it('unknown syntax option', () => {
 		})
 		.catch((err) => {
 			expect(err.message).toBe(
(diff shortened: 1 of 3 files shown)


ntwb:
This will be a breaking change requiring a _major_ release correct?

evilebottnawi:
@ntwb no, should be as minor release

hudochenkov:
@ntwb it will be a minor version. We add new option, it doesn't change automatic syntax detection.

hudochenkov:
> -   Added: `css` syntax option ([#4315](https://github.com/stylelint/stylelint/pull/4315)).

PR #4318, opened 2019-10-01
Improve test stability

hudochenkov:

Closes https://github.com/stylelint/stylelint/issues/4303.

I rewrote tests to be less complex. Following common sense and inspired by [Avoid Nesting when you're Testing](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing).

And I fixed stability issues. They were pretty simple as it turned out. When we use glob for `files` we can't guarantee what order of files we get. Therefore results we get from stylelint could be in different order between two runs. That's why it was failing on Windows a lot. And luckily it was failing for me on macOS, so I could find the problem :)

### DIFF
--- a/lib/__tests__/ignore.test.js
+++ b/lib/__tests__/ignore.test.js
@@ -6,467 +6,338 @@ const standalone = require('../standalone');
 
 const fixturesPath = path.join(__dirname, 'fixtures');
 
-describe('extending config and ignoreFiles glob ignoring single glob', () => {
-	let results;
-
-	beforeEach(() => {
-		return standalone({
-			files: [`${fixturesPath}/empty-block.css`, `${fixturesPath}/invalid-hex.css`],
-			config: {
-				ignoreFiles: '**/invalid-hex.css',
-				extends: ['./config-block-no-empty', './config-color-no-invalid-hex'],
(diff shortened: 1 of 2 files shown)


vankop:
> When we use glob for `files` we can't guarantee what order of files we get. Therefore results we get from stylelint could be in different order between two runs.

You are totally right! Good job!

lib/__tests__/ignore.test.js
```diff
@@ -6,467 +6,341 @@ const standalone = require('../standalone');
...
+		},
+		configBasedir: path.join(__dirname, 'fixtures'),
+	}).then(({ results }) => {
+		// two files found
 		expect(results).toHaveLength(2);
-	});
 
-	it('empty-block.css found', () => {
+		// empty-block.css found
 		expect(results[0].source.indexOf('empty-block.css')).not.toBe(-1);
```
vankop: Don’t we need to check here that invalid-hex.css possible here instead?

hudochenkov: I think we have it here to be sure that the first result is for file we expect.

jeddy3:
Excellent sleuthing!

PR #4321, opened 2019-10-02
add :is to pseudo-class list. fixes #4236.

### LINKED ISSUE #4236: Fix false positives for :is selector in selector-pseudo-class-no-unknown
> Clearly describe the bug

[`:is`](https://developer.mozilla.org/en-US/docs/Web/CSS/:is) selector is not known by this rule.

`selector-pseudo-class-no-unknown`

e.g.

```css
*:is(*) { color: red; }
```

e.g.

```json
{
  "rules": {
    "selector-pseudo-class-no-unknown": "always"
  }
}
```

`10.1.0`

CLI

No

`:is` to not trigger the rule

`:is` triggering the rule

```shell
 9011:3   ✖  Unexpected unknown pseudo-class selector ":is"   selector-pseudo-class-no-unknown
```

bkinnamon:

Closes #4236 

:is is still an experimental technology, but issue #4236 asked for a PR.

https://developer.mozilla.org/en-US/docs/Web/CSS/:is

### DIFF
--- a/lib/reference/keywordSets.js
+++ b/lib/reference/keywordSets.js
@@ -247,6 +247,7 @@ keywordSets.otherPseudoClasses = new Set([
 	'indeterminate',
 	'in-range',
 	'invalid',
+	'is',
 	'last-child',
 	'last-of-type',
 	'link',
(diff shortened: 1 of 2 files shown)


vankop:
Thanks for PR! Could you create test cases for that, please.

bkinnamon:
I'm not sure I've done this right. I'm new to pull requests and trying to get the feel for things. Did the test case commit come through?

hudochenkov:
I'll close and open this PR to try to kick in Github Actions. They didn't started for the last commit for some reason.

hudochenkov:
> -   Fixed: `selector-pseudo-class-no-unknown` false positives for `:is` selector ([#4321](https://github.com/stylelint/stylelint/pull/4321)).

Issue #4245, opened 2019-09-04
Replace Flow with TypeScript JSDoc


jeddy3:
In https://github.com/stylelint/stylelint/issues/4244#issue-489022754 @hudochenkov mentioned how looking after Flow is a bit of a pain.

I wonder whether it's worth investigating how much effort is involved in migrating away from Flow to [TypeScript JSDocs](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#supported-jsdoc). Does anyone have experience with this? I believe we'd be able to keep the codebase free of a compile step with this approach. A little more about the JSDoc style can be [found here](https://fettblog.eu/typescript-jsdoc-superpowers/), and there's a [(relatively old) discussion](https://github.com/postcss/postcss/issues/1221) about taking this approach in PostCSS.

Labeling as a discussion so people can add their thoughts on the pros and cons of such a replacement.

ntwb:
I'd be interested in taking a closer look at this, I try to grok Flow but it just doesn't come easy to me, one of the projects I'm involved with [Gutenberg](https://github.com/wordpress/gutenberg) has recently started doing the same, switching to TypeScript JSDocs.

As for time to do it, I'll be quote limited, starting a new job next week 🎉

Nevertheless I like the sounds of this idea and thanks for raising the discussion @jeddy3 💯

hudochenkov:
I think it's a good idea. TypeScript is more reliable and future-proof.

Here is one more article on this topic https://ricostacruz.com/til/typescript-jsdoc

evilebottnawi:
:+1: For typescript-jsdoc

vankop:
Typescript jsdoc is pain =) anyway it is better, then Flow (de facto TypeScript is only one optimal solution for types in JS, Flow was better for js before babel team implemented plugin for typescript)

Problem with `typescript-jsdoc` is that it is different from TypeScript in `ts` files and less documented.

Some guys in TypeScript gitter suggested me look on jsdoc tests to find usage example =(

vankop:
Maybe we could find (or write) babel plugin to easy migrate on TypeScript

vankop:
should be closed after: 
https://github.com/stylelint/stylelint/issues/2935

hudochenkov:
[A sandbox project to testing VS Code and JavaScript + JSDoc type checking.](https://github.com/patoi/ultima-thule)

hudochenkov:
[...]

PR #4325, opened 2019-10-04
WIP: replace flow with typescript-jsdoc

vankop:

https://github.com/stylelint/stylelint/issues/4245

Help wanted 🤗


hudochenkov:
I see you removed flow directive from all files. You decided to remove all Flow code first, and add JSDoc whenever you can think of? Or Flow is still running?

Could you add TypeScript to npm scripts, please? Can it also be added to `pretest` npm script so we run TypeScript and Flow together?

vankop:
Flow is running, but does not lint anything, I did it just for development purpose (don’t think about flow types). We can revert this commit and commit that removes flow-typed later

hudochenkov:
Looks like our actual JavaScript got broken in the last commit.

vankop:
> Looks like our actual JavaScript got broken in the last commit.

yes you are correct =) I will revert this for now

vankop:
If someone wants to help I suggest to write what you are taking here.

I will fix `lib/assignDisabledRanges.js`
and take
```
lib/standalone.js
lib/writeOutputFile.js
```

I think files from [this list](https://github.com/stylelint/stylelint/issues/4245#issuecomment-536902723) are prioritize

vankop:
I have created issue related to `~/lib/assignDisabledRanges` https://github.com/stylelint/stylelint/issues/4328

vankop:
@hudochenkov I think main work is done =) so I added types check in lint stage. Mostly need to write types for `utils` and fix `TODO TYPES`

hudochenkov:
Does TS failing because it's a WIP?

vankop:
Output looks a little bit strange, but looks like it works. It fails because it is around 200 warnings still . I did not found out why it is linting some files that are in excluded patterns =(

vankop:
I have fixed excluded patterns, now warnings (~50 at all) are more readable =)

types/meow/index.d.ts
```diff
@@ -0,0 +1,29 @@
+interface Flag {
```
G-Rath: There is an [`@types/meow`](https://www.npmjs.com/package/@types/meow) that looks pretty similar to this, so you should be able to just use that.

types/global-modules/index.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module 'global-modules' {
```
G-Rath: There is an [`@types/global-modules`](https://www.npmjs.com/package/@types/global-modules) that looks pretty similar to this, so you should be able to just use that.

G-Rath:
I've added a few comments about `.d.ts` files that can be replaced w/ their `@types` package.

If there is a library that doesn't have a type definition file published on DefinitelyTyped, I'm happy to create one, and get it published.

vankop:
@G-Rath Do you now any workaround of using `default` definitions. What I mean, for example for `strip-ansi` there are definitions in format https://github.com/chalk/strip-ansi/blob/master/index.d.ts, but they don't provide es6 exports (as defined in types). So maybe there are some workaround?

G-Rath:
@vankop `export default` is ES6-style exporting, so you can use ES6 style importing:

```
import blah from 'strip-ansi';
```

vankop:
Yes, you are correct for ES6, but we have code in CommonJS and `strip-ansi` also provide CommonJS export only (but types provide ES6 export). See

> for `strip-ansi` there are definitions in format https://github.com/chalk/strip-ansi/blob/master/index.d.ts, but they don't provide es6 exports

G-Rath:
Ah yes, sorry now I'm w/ you - the commit you pushed is the workaround; that's what `esModuleInterop` does: it transpiles everything to be exported as `{ default: ... }` to fix that issue.

Since you're in JS land however, you've got to do it yourself.

G-Rath:
FYI I've created a PR for adding definitions for [`globjoin`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38908). They should be merged & published within the week :)

vankop:
@G-Rath I know how it works =) I just did not realizes that so many packages support both ways. Thanks for deep explanation! (babel as well provides fallback for both exports `_interopRequireDefault`)

vankop:
@G-Rath As I know there is no a lot of typescript in jsdoc examples + it poor documented (if compare with common Typescript), could you suggest any source of examples / rich information about typescript in jsdoc? as I know there are a lot of tests here - https://github.com/microsoft/TypeScript/tree/master/tests/cases/fourslash. However, it hard to find usage some use cases there (folder is pretty large)

G-Rath:
@vankop sadly not really; you've pretty much summed up the state of the situation.

The two places I'd recommend are looking at [`tsdoc`](https://github.com/microsoft/tsdoc) - It's a project so underway it's not been announced, but I stumbled upon it myself a few months ago when looking for info like you. That can give you a good idea of at least how MS/TS would *like* JS/TSDocs to endup looking - if you just remember to ignore tags that are not official JSDoc ones, I'd expect their examples to already work in TS.

The other place of information is [the handbook](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html).

Finally, the eslint team are actually looking into this same thing, and so you might find interesting code in the [PR](https://github.com/eslint/eslint/pull/12082).

vankop:
@hudochenkov looks like problem with CI still exists

lib/cli.js
```diff
@@ -505,37 +429,44 @@ module.exports = (argv /*: string[]*/) /*: Promise<void>|void*/ => {
...
-	if (cli.flags.help || cli.flags.h) {
+	if (cli.flags.help) {
 		cli.showHelp(0);
 
-		return;
+		return Promise.resolve();
 	}
 
-	if (cli.flags.version || cli.flags.v) {
+	if (cli.flags.version) {
```
hudochenkov: `-h` and `-v` removed. Please, add them back.

lib/createStylelintResult.js
```diff
@@ -1,87 +1,25 @@
...
+ * @return {Promise<StylelintResult>}
+ */
+module.exports = function(stylelint, postcssResult, filePath, cssSyntaxError) {
+	/** @type {StylelintResult} */
 	let stylelintResult;
+	/** @type {string | undefined} */
 	let source;
 
-	if (postcssResult) {
+	if (postcssResult && postcssResult.root) {
```
hudochenkov: Why add this check?

lib/getPostcssResult.js
```diff
@@ -1,29 +1,27 @@
...
+/**
+ * @param {import('stylelint').StylelintInternalApi} stylelint
+ * @returns {Promise<import('postcss').Result>}
+ */
+module.exports = function(stylelint) {
+	/** @type {import('stylelint').GetPostcssOptions} */
+	const options = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : {};
 
-	const cached /*: ?postcss$result*/ = stylelint._postcssResultCache.get(options.filePath);
+	const cached = options.filePath ? stylelint._postcssResultCache.get(options.filePath) : undefined;
```
hudochenkov: Why this ternary?

lib/getPostcssResult.js
```diff
@@ -58,7 +58,7 @@ module.exports = function(stylelint /*: stylelint$internalApi*/) /*: Promise<?Ob
 						stringify: postcss.stringify,
 					};
 				}
-			} else if (syntax) {
+			} else if (stylelint._options.syntax) {
```
hudochenkov: Why not use `syntax` as before?

lib/getPostcssResult.js
```diff
@@ -122,13 +122,23 @@ module.exports = function(stylelint /*: stylelint$internalApi*/) /*: Promise<?Ob
 			return result;
 		})
 		.then((postcssResult) => {
-			stylelint._postcssResultCache.set(options.filePath, postcssResult);
+			if (options.filePath) {
```
hudochenkov: Why additional check?

hudochenkov:
In a totally new places. Might be that it happens with code changes made in this PR. Though locally nothing fails in this branch.

lib/createStylelintResult.js
vankop: Root possible undefined

lib/getPostcssResult.js
vankop: You can’t pass undefined in Map get parameter

lib/getPostcssResult.js
vankop: options.filePath possible undefined same with `get` method

lib/getPostcssResult.js
vankop: here is a little bit tricky place `stylelint._options.syntax` is `undefined | string` so after assigning on line 41 `syntax` implicitly have type `undefined | string`, but after we assign to `syntax` object - `{parse: any, stringify: any}` as well.

vankop:
> This PR would be close to impossible to review :(

Obviously, it is a problematic to review, if you have an idea how to split this, I will be glad to discuss.

> I don't think it makes much sense to do this comparison before this PR is ready

Sorry, I did not have time to fix last 20 warnings. All welcome to take a look on them =)

G-Rath:
> if you have an idea how to split this, I will be glad to discuss.

I would recommend focusing on adding the types, rather than removing flow.

For example, there are a lot of files that have been changed b/c you're removed `@flow` from the top - but there's no reason to do that in this PR, since they're just comments.

Even if flow is generating errors, the whole point of the PR is to move away from flow, so the errors shouldn't matter.

You can also remove the typing files for flow in another PR, as while sure they're not needed anymore, they're also not doing anything and so don't *need* to be removed in this PR :)

ntwb:
Thanks @G-Rath, and thanks for all your work thus far @vankop, this is all so much out of my wheelhouse, and anything that will help make reviewing this PR easier I'm all for.

So, if we can get away without including the removal of Flow and `@flow` comments that would be great if possible please.

vankop:
First I can suggest create separate PR with tsconfig and provided types. I think we can take types folder from this PR, then we can make PRs step by step with or without flow (on your demand).

> I would recommend focusing on adding the types, rather than removing flow.

I did it for the reason as described above - just for development reasons. As I see now it is not necessary (however some times flow and ts types are misleading)

> You can also remove the typing files for flow in another PR, as while sure they're not needed anymore, they're also not doing anything and so don't _need_ to be removed in this PR :)

Yes you are right, we can revert this commit

> Even if flow is generating errors, the whole point of the PR is to move away from flow, so the errors shouldn't matter.

I think the main idea to stay with flow while TS will not cover stylelint' "core"

vankop:
@hudochenkov after restarting CI still fail for Node.js 12.x.x, maybe it is relates to my changes =)

ntwb:
> after restarting CI still fail for Node.js 12.x.x, maybe it is relates to my changes =)

I cannot get the logs of the failing jobs to display past line `29` 😞

hudochenkov:
@ntwb it's a scrollable area. I always keep scrollbar enabled in OS settings :)

vankop:
types is fine now, I think I will start to split this in several PRs

G-Rath:
@vankop btw the definitions for `globjoin` got published :)

I'll circle back during the week and look to create definitions for the other libraries when I have a chance.

vankop:
Thanks all for your reviews! All this work is done in separate issues 🎉

PR #4329, opened 2019-10-05
bump chalk to 3.0.0-beta.1

vankop:

Unfortunately, correct types provided only in @3 version. Related to https://github.com/stylelint/stylelint/pull/4325

I have tested types, they work correctly.


vankop:
It is not necessary, there is `default` export too

PR #4335, opened 2019-10-10
Fix CSS syntax configuration when `--syntax css` is used

hudochenkov:

Fixing my mistakes https://github.com/stylelint/stylelint-demo/pull/177#issuecomment-540594505

postcss-safe-parser is just a parser, so it needs a stringifier.


hudochenkov:
> -   Fixed: syntax configuration for `--syntax css` ([#4335](https://github.com/stylelint/stylelint/pull/4335)).

Issue #4333, opened 2019-10-08
Release 11.1.0


hudochenkov:
Going to release tomorrow.

- [x] stylelint release
- [x] stylelint-demo update
- [x] stylelint.io update
- [x] tweet

hudochenkov:
No release for now. Out of blue [17 failed tests](https://github.com/stylelint/stylelint/runs/253528783) on Windows.

I the common theme:

```
    expect(received).toBe(expected) // Object.is equality

    Expected: "D:\\a\\stylelint\\stylelint\\lib\\__tests__\\fixtures\\empty-block-with-disables.css"
    Received: "d:\\a\\stylelint\\stylelint\\lib\\__tests__\\fixtures\\empty-block-with-disables.css"
```

Looks like path that known to stylelint and path which we get from Node.js `path` in tests, are different.

Maybe one of our dependencies had release in the past 5 hours and it broke something. 5 hours ago we had green CI.

hudochenkov:
I wish we had `package-lock.json` to confirm this :(

evilebottnawi:
@hudochenkov maybe it is time to lock file? I have been proposing to do this for a long time :smile:

vankop:
It is dep of one of out dep as I see =) because I can not find any released one

vankop:
As I understand current `11.0.0` also affected

hudochenkov:
Does anyone has local stylelint repository with `node_modules` installed at least 7 hours ago? Could you remove `.npmrc` and run `npm i --package-lock-only` to create package-lock and then share it with me?

vankop:
https://gist.github.com/vankop/07c4967f0d5abe3a22c764ab93ec5d81

vankop:
oh sorry it is not master =(

vankop:
latest changes for me
<img width="914" alt="Снимок экрана 2019-10-09 в 16 06 36" src="https://user-images.githubusercontent.com/10380560/66484040-e6dd9080-eaae-11e9-8792-e871ec4e2b03.png">

hudochenkov:
It might be Github Actions fault. Failing are inconsistent. Sometimes they happen, sometimes — not. No restart for master latest commit fix the build. But restarts in #4334 fixed the build once.

What I noticed between successful runs and failed, that Github Actions setup environment with different paths!

On success it uses `D:\`, on fails it uses `d:\`.

<details>
<summary>Fail</summary>

```
Run actions/checkout@v1

[...]

Issue #3895, opened 2019-01-16
Add package-lock.json


ybiquitous:
> What is the problem you're trying to solve?

Greenkeeper [supports lockfiles](https://greenkeeper.io/docs.html#lockfiles) now, so what about adding `package-lock.json`?

The following pros are quoted from [npm documentation](https://docs.npmjs.com/files/package-lock.json)  about `package-lock.json`:

> This file is intended to be committed into source repositories, and serves various purposes:
>
> - Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
> - Provide a facility for users to “time-travel” to previous states of node_modules without having to commit the directory itself.
> - To facilitate greater visibility of tree changes through readable source control diffs.
> - And optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.

The current latest NPM version is [`6.5.0`](https://github.com/npm/cli/releases/tag/v6.5.0) and it's mature enough. I think it's now when we should introduce the lockfile.

Related to #2766

jeddy3:
Thanks for the suggestion.

We adopted the ["Lockfiles for apps, but not for packages" approach](https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514) a while back. This has done us well so far. Has new information come to light that advises we should revisit our choice of approach?

ybiquitous:
Thanks for the quick help!

Surely, I think the opinion "Lockfiles for apps, but not for packages." makes sense, but the following opinion in https://github.com/sindresorhus/ama/issues/479#issuecomment-425722909 seems to be more reasonable to me.

I quotes from the comment:

> While this is correct, the inverse is not necessarily true: it's not because you don't have a lockfile that your local tree would match that of your users. So removing the lockfile doesn't solve the mismatch problem.
> 
> At least the lockfile guarantees that your fellow developers have the same trees.

Unless lockfiles, I'm worried that we may not be able to guarantee the same trees on the local machine of our fellow developers/contributors or on CI environments. 😟 

Now, we have [`npm ci`](https://docs.npmjs.com/cli/ci.html) and [`npm audit`](https://docs.npmjs.com/cli/audit.html) commands, which do many things for us with `package-lock.json`.
For example, in the case of the `event-stream` vulnerability last year, by `npm audit` we could examine whether the unfixed version of the package would have been installed on own local machine.

[...]

PR #4341, opened 2019-10-12
Add package-lock

hudochenkov:
Following discussion in https://github.com/stylelint/stylelint/issues/4333#issuecomment-539983673.


vankop:
I think this blocked by https://github.com/stylelint/stylelint/pull/4342

hudochenkov:
Not really.

vankop:
Just to be sure we won't get false positive/negative CI builds ☝️

ntwb:
Again, stylelint should NOT have a lock file, if there are any new valid arguments not previously covered in the last time this was proposed in #3895 please detail them here.

And please, *do not merge* this pull request without my explicit approval

vankop:
@ntwb thanks for pointing on previous discussion. Since lock file ignored in dependencies(I di not know it previously) you are totally right that stylelint does not need it. However, stylelint contributors need ability to find problematic dependencies fast, here lock file comes from. We can use npm install in CI to use latest semver packages

ntwb:
Thanks, currently re-researching the status and behaviour of `package-lock.json` files and I'll update this issue re-examaning my current/previous understandings of the scenario.

ntwb:
Via https://docs.npmjs.com/files/package-lock.json

> One key detail about `package-lock.json` is that it cannot be published, and it will be ignored if found in any place other than the _toplevel_ package. It shares a format with `npm-shrinkwrap.json`, which is essentially the same file, but allows publication. This is not recommended unless deploying a CLI tool or otherwise using the publication process for producing production packages.

Positive: This is a good thing, so "these days" the lock file will not be published

ntwb:
Via https://github.com/stylelint/stylelint/issues/3895#issuecomment-455116319

>    Now, we have npm ci and npm audit commands

> >Both these would be useful. We could run npm audit as part of our pretests on the CI.

Positive: I manually juggle this by deleting the `.npmrc` frequently so that I can run these commands, which is without a doubt frustrating to have to do.

G-Rath:
@ntwb one argument I'm not quite seeing is pretty much summed up by this PR (or related - It's been a long day so apologies if I'm missing something): 

A `package-lock.json` allows provides proof positive of the minimal versions required for the library to work, preventing "it works on my computer" problems.

In addition, any dependency that has high risk of being broken by a newer version of that package should have it's constraint adjusted accordingly (i.e if you use `*`, and a lack of `package-lock` means a new major is used, it should really be `^<major>`)

The summary of the above is that it tends to be easier to have it, than not. Point 2. tends to be self-proving: typically your constraints are that of patch versions, which means they shouldn't be at high-risk of breaking. If they are, the constraint really needs to be adjusted, regardless on if you have a `package-lock.json` or not.

vankop:
> [Yarn](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/) has written a great article summing my points up far better than I can :)

agree with authors) for package we still need lock file, for development reasons.

ntwb:
> @ntwb one argument I'm not quite seeing is pretty much summed up by this PR (or related conversations that lead to this PR - It's been a long day so apologies if I'm missing something):
>
> A package-lock.json allows provides proof positive of the minimal versions required for the library to work, preventing "it works on my computer" problems.

Thanks, I understand this point of view, I'll more than likely come back to this in further follow up replies, it relates to who to put the onus on, the stylelint devs, the stylelint plugin and config devs, or the end consumers of the packages, anyways I'll come back to this in my summary.

> In addition, any dependency that has high risk of being broken by a newer version of that package should have it's constraint adjusted accordingly (i.e if you use *, and a lack of package-lock means a new major is used, it should really be ^<major>)
> 
> The summary of the above is that it tends to be easier to have it, than not. Point 2. tends to be self-proving: typically your constraints are that of patch versions, which means they shouldn't be at high-risk of breaking. If they are, the constraint really needs to be adjusted, regardless on if you have a `package-lock.json` or not.

Indeed, yes, for the most past stylelint only ever uses `^` to declare semver ranges for our deps and dev-deps, if a package _breaks_ we will _pin_ the version of the package to the exact version until resolved.

> You can also use npm update to get the latest possible versions of all packages that your constraints will allow.

Yes, this once again comes back to where to put the onus on this process, I'll come back to it in a bit.

ntwb:
> [Yarn](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/) has written a great article summing my points up far better than I can :)

Yarn publishes the `package-lock.json` file by default, because it believes the lock file should both be committed to the repo and published with the package.

This directly contravenes _npm_ approach which by default does *not* distribute the _lock_ file which I highlighted previously above.

You can test this behaviour by running `npm pack` and `yarn pack` and extract the archives created from each of these to confirm.

hudochenkov:
> if a package breaks we will pin the version of the package to the exact version until resolved.

Before that we spend quite some time trying to figure out what package breaks. Usually I'm who doing this. It's not fun.

> Yarn publishes the package-lock.json file by default, because it believes the lock file should both be committed to the repo and published with the package.
>
> This directly contravenes npm approach which by default does not distribute the lock file which I highlighted previously above.

We use npm to publish.

ntwb:
> [Yarn](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/) has written a great article summing my points up far better than I can :)

> This lockfile contains information about the exact versions of every single dependency that was installed as well as checksums of the code to make sure the code is identical.
> 
> Yarn needs this info about every dependency because packages are constantly changing. New versions of packages are published all the time and since `package.json` specifies version ranges you need to lock them down to a single version.

This is the fundamental flaw in this Yarn article

Yarn suggests that "exact versions of every single dependency" and "you need to lock them down to a single version" should be recorded

What's the point of using SemVer if the only version it believes that matter is the exact version installed at the time of install?

> Version ranges are a way of specifying which types of changes you want to accept and which versions you want to prevent. These version ranges then resolve down to a single version which is either the version you have installed or the latest published version that matches your version range.
>
> If you don’t store which version you ended up installing, someone could be installing the same set of dependencies and end up with different versions depending on when they installed. This can lead to “Works On My Machine” problems and should be avoided.

Yes, this is what SemVer is all about, but wait, they state "These version ranges then resolve down to a single version which is either the version you have installed or the latest published version that matches your version range.", this is the purpose of SemVer and then state "someone could be installing the same set of dependencies and end up with different versions depending on when they installed.", again, this is the purpose of SemVer, and by conclusion lay blame at SemVer being the cause of "This can lead to “Works On My Machine” problems and should be avoided."

While yes, we do see occasionally a package dependency breaks stylelint because the released package breaks stylelint, it's unfortunate, but it is the purpose of SemVer and if a package breaks because it released a _minor_ or _patch_ release is not the fault of stylelint, the devs, or SemVer, mistakes happen.

This part is the start of who to place the onus on in relation to update dependencies, again, I'll follow this up again later.

> It also works as a security precaution: If the package author is either malicious or is attacked by someone malicious and a bad version is published, you do not want that code to end up running without you knowing about it.

Again, mistakes can happen, but this statement is clearly stating you should never trust a newly released package using SemVer, it is putting the onus here on the stylelint dev team to review every. single. release. of any package we depend upon, the onus thing again here.

> For applications, most developers agree that lockfiles are A Good Idea™. But there has been some question about using them when building libraries.

I agree, apps should use and ship a lock file, I've agreed with since lock files were introduced, nothing new here.

> This seems to logically makes sense, but let’s dive deeper into the problem.
> Development Dependencies

This entire section _dives_ into absolutely nothing, there is no explanation in this section for their argument of why libraries also should contain a lock file, nothing, its entirely useless.

> When a package accidentally publishes a breaking change it starts the clock on who will be the first person to catch it. Whoever installs that breaking change first will (most likely) be the first to discover it.
>
> Let’s imagine we have a package called left-pad which takes a string and adds a specified amount of padding in front of it. This small and seemingly harmless package is used by a really big project, which we’ll just call… Babel.
> 
> One day, the maintainer of left-pad decides they are going to do something unspeakably evil and make it pad the right side instead. They publish it as a patch version which quickly spreads to everyone using it.

Right, mistakes happen, my reading of this is once again suggesting that I should be cautious of SemVer, again, this is what SemVer does, its the core principal of SemVer and why it is used. When a bad actor or bad package release occurs we take action to restrict the package version, roll back the package version, _pin_ the package version to an explicit version etc, the onus here is on the stylelint developers and contributors here.

Onto the "onus" subject...

> The contributor’s burden
> The user’s burden


Nothing much of substance here in either of these sections, its a preamble on the "onus" for the last section of the article.

> In Closing
> The idea that not using lockfiles in libraries somehow saves users from encountering breaking changes is at best extremely rare and at worst is never true.

I'm only going to quote this paragraph from this section...

This is why we use SemVer, this is what Semver is for, it's a set of rules that allows developers to specify a range of package versions to install, we "trust" these packages and the developers that publish these packages, occasionally mistakes or bad actors break SemVer, it's unfortunate when it happens, but also it only ever happens occasionally.

ntwb:
> Before that we spend quite some time trying to figure out what package breaks. Usually I'm who doing this. It's not fun.

Agreed, its no fun whatsover, I'll return to this subject with the "onus" subject I keep mentioning shortly...

hudochenkov:
> This is even more confusing, so when running yarn install any yarn.lock will be ignored, yet they publish the package-lock.json file, so what happens to users of npm who run npm install and the package-lock.json file exists?

It will be ignored:

> One key detail about package-lock.json is that it cannot be published, and it will be ignored if found in any place other than the toplevel package.

https://docs.npmjs.com/files/package-lock.json

ntwb:
Let's get into how the addition of a lock file will affect just the stylelint repo.

## Scenario

Via https://github.com/stylelint/stylelint/issues/3895#issuecomment-455974225

> The current `package.json` file:

> `"postcss-value-parser": "^3.3.0",`

> Using `npm ci` with a lock file will install `postcss-value-parser` version `3.3.0`, if `3.3.1` is released the stylelint Travis CI jobs will ***not*** test that release, it will continue to use `3.3.0` until we bump our dependency to `3.3.1`.

Updating the above to use a current example from the `dependencies` of our `package.json` file as it stands in this branch:

https://github.com/stylelint/stylelint/blob/530b7bf46068a4a4e81bbbfba717655c7be2cc95/package.json#L55

The latest release of `ignore` is `5.1.4`, we declare `^5.0.6` currently.

https://github.com/kaelzhang/node-ignore/releases

This PR uses `ignore` version `5.1.4`:

https://github.com/stylelint/stylelint/pull/4341/files?file-filters%5B%5D=.json&file-filters%5B%5D=dotfile#diff-32607347f8126e6534ebc7ebaec4853dR4362-R4364

```json
    "ignore": {
      "version": "5.1.4",
      "resolved": "https://registry.npmjs.org/ignore/-/ignore-5.1.4.tgz",
```

## Observations

- Although it's obfuscated in the _lock_ file, for the immediate future stylelint will perform all of its CI testing using `ignore` version `5.1.4`

- What happens when a new version of `ignore` is released? e.g. `5.1.5`

  - Ideally the `package-lock.json` file will be updated by Greenkeeper/Dependabot by creating a branch and testing via the CI, if it passes it's committed to the repo.

  - When that _lock file_ change is merged into the repo if and when should stylelint update the `package.json` version be updated from the current `^5.0.6` to `^5.1.5`? I'd suggest that it should be immediately because the _patch_ release of `ignore` resolves a _bug fix

ntwb:
> > This is even more confusing, so when running yarn install any yarn.lock will be ignored, yet they publish the package-lock.json file, so what happens to users of npm who run npm install and the package-lock.json file exists?
> 
> It will be ignored:
>
> > One key detail about package-lock.json is that it cannot be published, and it will be ignored if found in any place other than the toplevel package.
>
> https://docs.npmjs.com/files/package-lock.json

This example relates only if Yarn was used to publish the package, since we don't its irreverent as we do use _npm_ to publish.

I wanted to highlight the ongoing inconsistencies between Yarn and npm

hudochenkov:
I've updated `package.json` to match installed version. Next time any dependency would be updated, `package.json` would be updated as well. So `package.json` and `package-lock.json` would be in sync.

> What happens when a new version of ignore is released? e.g. 5.1.5
>
> Ideally the package-lock.json file will be updated by Greenkeeper/Dependabot by creating a branch and testing via the CI, if it passes it's committed to the repo.

We could publish new stylelint version before PR to update dependency happen. User could have the same old version if depencency. Or they could have latest version of dependency if they installed stylelint from scratch.

> When that lock file change is merged into the repo if and when should stylelint update the package.json version be updated from the current ^5.0.6 to ^5.1.5? I'd suggest that it should be immediately because the patch release of ignore resolves a _bug fix

It would happen as Dependabot configured. Once per month.

The thing is, every user (and developer, and CI) have unique tree of dependencies installed for the same version of stylelint. Because of SemVer. The only time users will have the same tree if they installed stylelint in a clean repository at the same.

`package-lock.json` will help us guarantee that developers and CI (we can exlude CI) will have the same dependency tree. It will help us to find what dependency broke stylelint faster, when it happen.

It won't make us publish more risky software, because again it's a SemVer and everyone has unique dependency tree when stylelint installed.

ntwb:
> I've updated `package.json` to match installed version. Next time any dependency would be updated, `package.json` would be updated as well. So `package.json` and `package-lock.json` would be in sync.

This relates to my "onus" comment incoming momentarily, it adds extra onus on the team update the `package.json` and `package-lock.json` files more frequently.

> We could publish new stylelint version before PR to update dependency happen. User could have the same old version if depencency. Or they could have latest version of dependency if they installed stylelint from scratch.

Again, this relates to my "onus" comment incoming momentarily, it adds extra onus on the team to publish new stylelint versions more frequently.

XhmikosR:
For what is worth you guys should also enable [automated security fixes](https://help.github.com/en/articles/configuring-automated-security-fixes).

This is unrelated to add dependabot to the repo.

For what is worth, the lock file is the way to go nowadays. semver is moot for stylelint itself with a lockfile, that is true. And the end users will get other deps as you already pointed out. The only way is if everyone either pins their versions or use a lockfile or even better a combination of both. But that's outside of the scope of this PR :)

ntwb:
> For what is worth you guys should also enable automated security fixes.

Moving to the addition of a lock file, this is something we don't want to do, extra onus is placed on the team updating the `package.json` file from those PRs.

> This is unrelated to add dependabot to the repo.

I think dependabot can do both the `package.json` and `package-lock.json` files in a single pull request.

> For what is worth, the lock file is the way to go nowadays. semver is moot for stylelint itself with a lockfile, that is true. And the end users will get other deps as you already pointed out. The only way is if everyone either pins their versions or use a lockfile or even better a combination of both. But that's outside of the scope of this PR :)

It's not quite out of scope, and gets to one of my main points, if you have a lock file you no longer have SemVer, you cannot have both.

By adding a lock file you eliminate semver, thus increasing the onus on the team to update dependancies.

ntwb:
## Onus

> Before that we spend quite some time trying to figure out what package breaks. Usually I'm who doing this. It's not fun.

@hudochenkov rightly highlighted this issue, time and time again I know Aleks has spent for more time than what should be necessary going down _rabbit holes_ to debug these types of issues.

I've been down the same rabbit holes, the _windows globby_ one for example I've spend days on, and I still never got it resolved.

This work is frustrating and time consuming and truly drains ones energy and the end result is a reduction of time contributing that could have been better spent elsewhere.

### Onus: On the stylelint dev team

By adding a `package-lock.json` file the team will be consistently testing using _exact_ versions of the dependencies, be they `dependencies` or `devDependencies`

These versions will remain the same until Greenkeeper or Dependabot bump these deps in the `package-lock.json` file and or the `package.json` file.

When using SemVer alone, in other words without a _lock_ file, we cross our fingers and "hope" that any _patch_ or _minor_ releases of packages we depend upon "improve" stylelint by way of bug fixes and optimizations.

The stylelint team need to decide how to manage the `dependencies` and `devDependencies` that are used:

1. How often should the _lock_ file be updated per the ranges we specify in the `package.json` file?
2. How often should the `package.json` file be updated per the latest _patch_ and _minor_ releases of the `dependencies` and `devDependencies` packages we use?
3. How often should stylelint release a new release with the updated `dependencies`, in particular if we know a package depended upon contains a bug fix?

### Onus: On the stylelint community

1. Should we pass on this same philosophy to the community of stylelint plugin and config teams?
2. Should we recommend these plugins and teams also use a _lock_ file that essentially _pins_ their _lock_ files to also use an exact version of stylelint in their lock files?


### Onus: On the stylelint users

1. If stylelint and the community of plugins and configs only test on an exact version of stylelint should they also only use an exact version of styleint due to the worry of SemVer breaking their linting?

### Onus: Scenario

By adding a `package-lock.json` file to test to aid the stylelint development team in my opinion has the following pros and cons:

#### Pros:

1. Debugging what version of a package was used is far easier, because we use _exact_ versions 


#### Cons:

1. Every. Single. Dependency needs to updated by what I hope Greenkeeper or Dependabot can perform via pull request.

### Onus: Summary

#### Dependency Updates

- The process can be somewhat automated by the bots creating pull requests for _patch_, _minor_, and _major_ releases of the `dependencies` and `devDependencies` packages we use

  - Pros: This is an automated process by bots
  - Cons: These pull requests still require reviewing, approving, and merging by the stylelint development team

- Allow the bot pull requests to be automatically merged is one way to lessen this extra "onus"?

  - Pros: Nothing required by the stylelint development team
  - Cons: Little to no control of the automatic updates

#### How often should stylelint be published?

- With the _exact_ versions used by way of the _lock file_ how does this affect the frequency that stylelint releases should be published?

  - Pros: stylelint has from the point of view of the stylelint development team a set of reliable and exact version of packages it depends upon.
  - Cons: Extra onus is placed on the stylelint developer team to publish more frequently because of bug fixes included in dependencies updated.

### Finishing up:

I'm genuinely concerned about placing further "onus" on the stylelint development and contributor teams, even shifting that onus from one group to another.

Contribution time by everyone is incredibly valuable and limited and we should ensure that what we are asking of everyone on the team is fair and sustainable.

Finally, is this extra "onus" justified on the development team and does this process outweigh the current behaviour of simply letting SemVer just do what it is designed to perform?

ntwb:
As the `package-lock.json` file is **not** published let's have a go at this, let's see what (if any) extra onus is placed on the development team with this change.

Let's review how we can make tools such as Greenkeeper or Dependabot help here, I'm OK to switch to Dependabot if that makes sense to switch.

Please keep in mind the onus information I've outlined above and help us to update the tooling and docs that will minimize this onus shift as much as possible.

Worst case scenario we can remove the _lock file_ if the onus shift is not sustainable and the tools cannot help mitigate it, its just code afterall.

cc: @stylelint/owners @stylelint/contributors @stylelint/core

ntwb:
I think as the `package-lock.json` file is no longer published, at least if you use _npm_ to publish it continues to not break the _"Lockfiles for apps, but not for packages."_ mantra, I might even have to update my [post](https://netweb.blog/2018/01/26/lockfiles-breaks-semantic-versioning/) on the subject 😏

XhmikosR:
@ntwb the security fixes is unrelated to adding dependabot is what I meant.

vankop:
I still don't understand why we can not combine semver and lock files together =) we can use semver for CI and lock files for devs, so if we will get problems with deps we will know about it, but development won't stop since devs using lock files

XhmikosR:
You can and you should, that's the only way to have full control.

vankop:
> Let's review how we can make tools such as Greenkeeper or Dependabot help here, I'm OK to switch to Dependabot if that makes sense to switch.

Minor info -  it is [canary build](https://www.thoughtworks.com/radar/techniques/canary-builds) tools. Hope to see canary build from revision number for open source (maybe Microsoft will do it)

vankop:
> You can and you should, that's the only way to have full control.

So @hudochenkov can we revert changes for CI and use `npm install` instead of `npm ci` ?

XhmikosR:
@vankop I didn't suggest this personally. On CI `npm ci` should be used. Why would you want to use `npm i` instead on CI?

vankop:
how we can find problem then? only in CI, but with `npm ci` we will use lock file instead of semver

XhmikosR:
But `npm i` still uses the lock file. The differences are other things, like for example `npm ci` will fail if the lock file isn't what it should be.

So, to answer your question, you will see the failures in each PR that bumps a dep.

XhmikosR:
For example https://github.com/twbs/bootstrap/pull/29482/commits/2548df3d1c8aed4741bb820b6167d8698a908acb

So, if you have a bot for deps, you will see any failures on each PR that bumps a dep.

hudochenkov:
> For what is worth you guys should also enable automated security fixes.

We have it enabled, of course :)

> 1. How often should the lock file be updated per the ranges we specify in the package.json file?
> 2. How often should the package.json file be updated per the latest patch and minor releases of the dependencies and devDependencies packages we use?

As we configured Dependabot (currently monthly). And we can bump all dependencies prior every release.

> 3. How often should stylelint release a new release with the updated dependencies, in particular if we know a package depended upon contains a bug fix?

Only if not releasing would make bad UX. E. g. deep in dependencies package with vulnerability and `npm audit` is not happy.

> 1. Should we pass on this same philosophy to the community of stylelint plugin and config teams?
> 2. Should we recommend these plugins and teams also use a lock file that essentially pins their lock files to also use an exact version of stylelint in their lock files?

I don't think we should do anything here. Every package developer is free to do what they want.

> Allow the bot pull requests to be automatically merged is one way to lessen this extra "onus"?

I think, we would like to enablep [auto merge](https://dependabot.com/docs/config-file/#automerged_updates) for patch-minor versions in two-three month, when we had better picture about Dependabot. Last time it created like 10 PRs which @ntwb had to manually merge.

> With the exact versions used by way of the lock file how does this affect the frequency that stylelint releases should be published?

Package-lock won't change frequency.

> Finally, is this extra "onus" justified on the development team and does this process outweigh the current behaviour of simply letting SemVer just do what it is designed to perform?

We'll see how things are going in the comming month. If it there would be more burden having package-lock, we could disable it again.

> Let's review how we can make tools such as Greenkeeper or Dependabot help here, I'm OK to switch to Dependabot if that makes sense to switch.

We switched to Dependabot like a month ago :)

Issue #4337, opened 2019-10-11
Fix failing Windows tests


hudochenkov:
Following discussion started in https://github.com/stylelint/stylelint/issues/4333. Failing tests are caused by Github Actions inconsistent environment. I've contacted support and they told me to open an issue, which I did https://github.com/actions/virtual-environments/issues/33.

This issue is to have reference about status of tests.

ntwb:
Related: Windows CI tests are switching to PowerShell by default next week

https://github.blog/changelog/2019-10-17-github-actions-default-shell-on-windows-runners-is-changing-to-powershell/

> If your `run` step contains Windows batch scripting you should update it to specify `cmd` as the shell to avoid being broken.

```shell
- run: |
    ...
  shell: cmd
```

As we are not running any batch scripting, in my mind it should not break!

XhmikosR:
Do we know if this issue still happens?

hudochenkov:
It's not happening anymore, but the fix is ugly :(

XhmikosR:
No, I meant if the issue happens without the workaround :)

hudochenkov:
We need to remove workaround and check.

XhmikosR:
I have a branch, I'll try to clean it up and see how it goes.

PR #4342, opened 2019-10-12
Add workaround for failing Windows environment

hudochenkov:

Temporary workaround for https://github.com/stylelint/stylelint/issues/4337.

Used [this](https://github.com/Microsoft/vscode/issues/9448#issuecomment-245024153) idea. Because for Windows case in a path is not important.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -61,7 +61,7 @@ jobs:
     # Workaround until https://github.com/stylelint/stylelint/issues/4337 is fixed
     - name: Change directory to uppercase drive letter
       if: matrix.os == 'windows-latest'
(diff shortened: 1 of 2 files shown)


hudochenkov:
Can't verify that workaround is working, because all environments were created with uppercase letter :)

vankop:
may be handle this in our code?
we can use knowledge about `process.cwd` call result

hudochenkov:
I think we shouldn't fix what is not broken. We have no reports from users on this regard.

hudochenkov:
It worked! At checkout I see lowercase letter, then it changes to uppercase.

vankop:
So we can remove extra comment?)

PR #4344, opened 2019-10-13
Separate dependency installation and test

hudochenkov:
Makes CI steps a little bit more granular.


XhmikosR:
@hudochenkov may I also suggest that you set `CI: true` in the workflow root? It's what Travis CI had and suppresses postinstall messages from some package.

See for example https://github.com/twbs/bootstrap/blob/master/.github/workflows/test.yml#L3-L4

PR #4345, opened 2019-10-13
Update usage docs to introduce npx

### LINKED ISSUE #4324: Fix clarity of CLI Docs
> Clearly describe the documentation issue

The [CLI Docs](https://stylelint.io/user-guide/cli) instructs users to install stylelint with `npm install stylelint --save-dev`, and then suggests that, as a `devDependancy`, `stylelint --help` could be run. 

Unless there's something wrong with my environment which is preventing the intended behaviour, I believe that for the CLI to work, users would either have to install stylelint with the `--global` flag, or use the full path like so: `./node_modules/.bin/stylelint --help`. 
The use of `npm link` being yet another option....

Clarification on intended behaviour and more granular troubleshooting instructions if user runs into a problem. See also #4094.

erasmuswill:

Closes #4324

### DIFF
--- a/docs/user-guide/cli.md
+++ b/docs/user-guide/cli.md
@@ -12,9 +12,11 @@ npm install stylelint --save-dev
 
 ## Usage
 
-`stylelint --help` prints the CLI documentation.
+`npm run stylelint --help` prints the CLI documentation.
 
 The CLI outputs formatted results into `process.stdout`, which you can read with your human eyes or pipe elsewhere (e.g. write the information to a file).
+
+You can also use `npx stylelint --help` to check out the help files before adding stylelint as a dependancy.
 
 ### Examples
 
--- a/docs/user-guide/cli.md
+++ b/docs/user-guide/cli.md
@@ -12,7 +12,7 @@ npm install stylelint --save-dev
 
 ## Usage
 
-`npm run stylelint --help` prints the CLI documentation.
+`./node_modules/stylelint/bin/stylelint.js --help` prints the CLI documentation after adding stylelint as a dependancy.
 
 The CLI outputs formatted results into `process.stdout`, which you can read with your human eyes or pipe elsewhere (e.g. write the information to a file).
 


docs/user-guide/cli.md
```diff
@@ -12,10 +12,12 @@ npm install stylelint --save-dev
 
 ## Usage
 
-`stylelint --help` prints the CLI documentation.
+`npm run stylelint --help` prints the CLI documentation.
```
ntwb: This will not work unless the user has added a `stylelint` scrip in their `package.json` file

Perhaps `npx run stylelint --help`, but again as highlighted in the issue, it's difficult to assume how a user uses stylelint...

I think i'd prefer no change on this line, for the reasons outlined here and in the issue

erasmuswill: Today I learned. It works with `yarn` and I'm so used to almost everything being the same across the CLI interfaces. Sorry for the trouble and thanks for catching it!

docs/user-guide/cli.md
```diff
@@ -12,10 +12,12 @@ npm install stylelint --save-dev
 
 ## Usage
 
-`stylelint --help` prints the CLI documentation.
+`./node_modules/stylelint/bin/stylelint.js --help` prints the CLI documentation after adding stylelint as a dependancy.
```
ntwb:
```suggestion
`./node_modules/stylelint/.bin/stylelint.js --help` prints the CLI documentation after adding stylelint as a dependency.
```
I think you'll find there is a `.` period before `bin`...

hudochenkov: I think @ntwb meant `./node_modules/.bin/stylelint`. This is correct path, which not relies on implementation details of stylelint. `./node_modules/stylelint/bin/stylelint.js` will break if we change filename or location of this file. `./node_modules/.bin/stylelint` always will be the same.

hudochenkov:
I think we should revert all changes to keep refering as `stylelint`, but add explanation how to run locally installed stylelint:

> To use locally installed stylelint, use [`npx`](https://www.npmjs.com/package/npx) (availabled with `npm@5.2.0+`). E. g.:
>
> ```
> npx stylelint --help
> ```

<pre><code>
To use locally installed stylelint, use [`npx`](https://www.npmjs.com/package/npx) (availabled with `npm@5.2.0+`). E. g.:

```
npx stylelint --help
```
</code></pre>

docs/user-guide/cli.md
```diff
@@ -12,10 +12,12 @@ npm install stylelint --save-dev
...
 
 The CLI outputs formatted results into `process.stdout`, which you can read with your human eyes or pipe elsewhere (e.g. write the information to a file).
 
+You can also use `npx stylelint --help` to check out the help files before adding stylelint as a dependancy.
```
XhmikosR:
```suggestion
You can also use `npx stylelint --help` to check out the help files before adding stylelint as a dependency.
```

hudochenkov:
@erasmuswill would you mind making requested change, please?

PR #4348, opened 2019-10-14
Fix false negatives for disallowInList: true no-duplicate-selectors

### LINKED ISSUE #4339: Fix false negatives for disallowInList: true no-duplicate-selectors
> Clearly describe the bug

There are false negatives in `no-duplicate-selectors` with `disallowInList: true`.

`no-duplicate-selectors`

```css
textarea {
  border: 1px;
}

input, textarea {
  border: 2px;
}
```

```json
{
  "rules": {
    "no-duplicate-selectors": [true, {"disallowInList": true}]
  }
}
```

`11.1.1`

CLI with `stylelint 'src/**/*.css'`.

Also this bug reproduces here: https://stylelint.io/demo#N4Igxg9gJgpiBcIAuMAeSCGAnGGAEwAOgHZ54BGEWsW8eAjAA6oDcJAviSQJbGMCuSADR4U6bLgIkylajFp4ATMzbFOxEEJABnAJ7FMqBOG3bN4CMQBm3AObGipPIRBZ+AGxjaXdR2TIuxBAAtFD8jO7cYBgowdownmBIVN4IeADaSG4wIkQgUNzaGO7uEADuAJLEADKFSD6i2ewAutJ46uwg7EA.

No.

The following warning should be flagged:

```text
5:8 error Unexpected duplicate selector "textarea", first used at line 1 (no-duplicate-selectors)
```

No warnings were flagged.

### LINKED ISSUE #4340: Fix false positives with :not() and disallowInList: true in no-duplicate-selectors
> Clearly describe the bug

There are false positives in `no-duplicate-selectors` with `disallowInList: true` when using `:not()`.

`no-duplicate-selectors`

e.g.

```css
input:not([type="button"]) {
  border: 2px;
}

[type="button"] {
  border: 1px;
}
```

e.g.

```json
{
  "rules": {
    "no-duplicate-selectors": [true, {"disallowInList": true}]
  }
}
```

`11.1.1`

CLI with `stylelint 'src/**/*.css'`.

Also this bug reproduces here: https://stylelint.io/demo#N4Igxg9gJgpiBcICWA7ADgVwC7xRLAFANpYCeaMAvADogBG2WEKtAugJQAEw1KnndCACdYQ+JwBMaAB4BuXgF9evEuSq0GWJixCtuvfoJEwxnAIwz5KJShAAaEAGdSKLAENpCcI8f3wzADMkAHMvHj5OWiEMABsYR1pxcP5+WjwAWigMNBikMDcsGHTHGDiwJiEEhE4SaJg7fRAoJEc3GJiIAHcASRQAGRasRM4sOoVWA04bBRAFIA.

No.

No warnings to be flagged.

```text
5:1 error Unexpected duplicate selector "[type="button"]", first used at line 1 (no-duplicate-selectors)
```

Boto1:
Fixes #4339

### DIFF
--- a/lib/rules/no-duplicate-selectors/__tests__/index.js
+++ b/lib/rules/no-duplicate-selectors/__tests__/index.js
@@ -209,7 +209,7 @@ testRule(rule, {
 			],
 		},
 		{
-			code: 'textarea {}; input, textarea {}',
+			code: 'textarea {} input, textarea {}',
 			description: 'duplicate within a grouping selector',
 			warnings: [
 				{
--- a/lib/rules/no-duplicate-selectors/__tests__/index.js
+++ b/lib/rules/no-duplicate-selectors/__tests__/index.js
@@ -209,6 +209,15 @@ testRule(rule, {
 			],
 		},
 		{
+			code: 'textarea {}; input, textarea {}',
+			description: 'duplicate within a grouping selector',
+			warnings: [
+				{
+					message: messages.rejected('textarea', 1),
+				},
+			],
+		},
+		{
 			code: 'input, div {};\n textarea, section {};\n textarea {}',
 			description: 'duplicate within a grouping selector. multiline',
 			warnings: [
--- a/lib/rules/no-duplicate-selectors/index.js
+++ b/lib/rules/no-duplicate-selectors/index.js
@@ -42,6 +42,9 @@ const rule = function(actual, options) {
 		// from other rules that share the same parent and the same source.
 		const selectorContextLookup = nodeContextLookup();
 
+		if (_.get(options, 'disallowInList')) {
+			root.nodes.sort((a, b) => a.selector[0].localeCompare(b.selector[0]));
+		}
 		root.walkRules((rule) => {
 			if (isKeyframeRule(rule)) {
 				return;
--- a/lib/rules/no-duplicate-selectors/index.js
+++ b/lib/rules/no-duplicate-selectors/index.js
@@ -45,6 +45,7 @@ const rule = function(actual, options) {
 		if (_.get(options, 'disallowInList')) {
 			root.nodes.sort((a, b) => a.selector[0].localeCompare(b.selector[0]));
 		}
+
 		root.walkRules((rule) => {
 			if (isKeyframeRule(rule)) {
 				return;


vankop:
@Boto1 Thanks for PR! Unfortunately, linting is failing, try to run `npm run pretest` locally to find problematic places.

vankop:
Could you please add test case provided in issue? (with `:not`)

Boto1:
> Could you please add test case provided in issue? (with `:not`)

I`m so sorry, i referenced wrong issue while creating pr, its #4339

lib/rules/no-duplicate-selectors/index.js
```diff
@@ -42,6 +42,10 @@ const rule = function(actual, options) {
 		// from other rules that share the same parent and the same source.
 		const selectorContextLookup = nodeContextLookup();
 
+		if (_.get(options, 'disallowInList')) {
+			root.nodes.sort((a, b) => a.selector[0].localeCompare(b.selector[0]));
```
hudochenkov: This will mutate PostCSS tree. It will change user files.

Also, I think fix should be somewhere around these lines https://github.com/stylelint/stylelint/blob/739a05b8536196a87567da360b13886e32218ea0/lib/rules/no-duplicate-selectors/index.js#L78-L83 Because it's the logic for `disallowInList` option.

lib/rules/no-duplicate-selectors/__tests__/index.js
```diff
@@ -208,6 +208,15 @@ testRule(rule, {
 				},
 			],
 		},
+		{
+			code: 'textarea {}; input, textarea {}',
```
hudochenkov:
```suggestion
			code: 'textarea {} input, textarea {}',
```

To make CSS valid :)

lib/rules/no-duplicate-selectors/index.js
Boto1: okay, can you please explain what is the difference between `contextSelectorSet` and `selectorList`? Because i found this kind of issue: if css file contains `input, textarea {} textarea{}` then contextSelectorSet will contain `input,textarea`(i mean `selector` in forEach loop) and selectorList will be `textarea`, thus `includes` returns true. But in our case contextSelectorSet contains `textarea` and selectorList is `input,textarea` so includes returns false. I think i dont get the purpose of these two entities right.

hudochenkov: Sorry, I missed you question. I believe `contextSelectorSet` has lists of individual selectors per their scope. E. g.

```css
a {}
b {}

@media print {
    a {}
    b {}
}
```

The first two selectors (`a`, `b`) on a top level doesn't know about selectors in other scope (inside `@media`) and vice versa. It's my assumption from reading the code, I did verify that in debugger.

`selectorList` is current rule selectors. E. g.:

```css
a, b {}

c, d {}
```

If you're checking second rule it will be `['c', 'd']`.

jeddy3:
Fixed.

- https://github.com/stylelint/stylelint/issues/4339#issuecomment-689055760
- https://stylelint.io/demo#N4Igxg9gJgpiBcIAuMAeSCGAnGGAEwAOgHZ54BGEWsW8eAjAA6oDcJAviSQJbGMCuSADR4U6bLgIkylajFp4ATMzbF2IISADOAT2KZUCcFq0bwEYgDNuAcyNFSeQiCz8ANjC3O6DsmWfEEAC0UPyMbtxgGChBWjAeYEhUXgh4ANpIrjAiRCBQ3FoYbm4QAO4AksQAMgVI3qJZ7AC60nicaiDsQA

PR #4350, opened 2019-10-15
Set CI environment variable

hudochenkov:

Ref: https://github.com/stylelint/stylelint/pull/4344#issuecomment-542195301


package.json
```diff
@@ -121,7 +121,7 @@
     "flow": "flow",
     "flow-defs": "./node_modules/.bin/flow-typed install jest@24 lodash@4",
     "jest": "jest",
-    "jest:detectleaks": "jest  --detectLeaks",
+    "jest:detectleaks": "jest --detectLeaks",
```
hudochenkov: Not related for this PR, but I didn't want to open new PR for this.

.github/workflows/nodejs.yml
```diff
@@ -9,6 +9,9 @@ on:
     branches:
       - '**'
 
+env:
+  CI: true
```
vankop: @hudochenkov I think I already mentioned in PR when we was adding github actions, I did not find who is reading this env.

PR #4353, opened 2019-10-15
types: add types to writeOutputFile.js

vankop:

add types to writeOutputFile.js

Since `writeFileAtomic` is a fork, I decided to skip it from type checking and use it as blackbox and just define output interface

> How to check that TS works?

run `npm run lint:types | grep writeOutputFile`
run `npm run lint:types | grep writeFileAtomic`

### DIFF
--- a/lib/writeOutputFile.js
+++ b/lib/writeOutputFile.js
@@ -2,10 +2,32 @@
 'use strict';
 
 const path = require('path');
-const stripAnsi = require('strip-ansi');
 const writeFileAtomic /*: Function*/ = require('./vendor/writeFileAtomic');
+const { default: stripAnsi } = require('strip-ansi');
 const { promisify } = require('util');
+
+/**
+ * @typedef {Object} WriteFileAtomicOptions
+ * @property {{uid: number, gid: number}} chown
+ * @property {string | null} [encoding='utf8']
+ * @property {boolean} [fsync=true]
+ * @property {number} [mode]
+ * @property {Function} [tmpfileCreated]
+ */
+
+/**
+ * @type {((
+ *  filename: string,
+ *  data: string | Buffer,
+ *  options?: string | WriteFileAtomicOptions
+ *  ) => Promise<Error | undefined>)}
+ */
 const writeFileAtomicAsync = promisify(writeFileAtomic);
 
+/**
+ * @param {string} content
+ * @param {string} filePath
+ * @returns {Promise<Error | undefined>}
+ */
 module.exports = (content /*: string*/, filePath /*: string*/) =>
 	writeFileAtomicAsync(path.normalize(filePath), stripAnsi(content));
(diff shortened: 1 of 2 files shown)


hudochenkov:
Do you think it's a good idea to update write-file-atomic? They [removed](https://github.com/npm/write-file-atomic/blob/master/CHANGELOG.md) dependency in question in 3.0. But there still a [leak](https://github.com/npm/write-file-atomic/issues/52), so we could update our copy with copy that has a [fix](https://github.com/npm/write-file-atomic/pull/53).

vankop:
@hudochenkov I don't know which problem they had.. Anyway I think it could be done in another PR

lib/writeOutputFile.js
```diff
@@ -2,10 +2,32 @@
...
+ * @property {number} [mode]
+ * @property {Function} [tmpfileCreated]
+ */
+
+/**
+ * @type {((
+ *  filename: string,
+ *  data: string | Buffer,
+ *  options?: string | WriteFileAtomicOptions
+ *  ) => Promise<Error | undefined>)}
```
hudochenkov: It's just for me to learn:

Here it says: `writeFileAtomicAsync` accept three parameters (and what these parameters are), and returns `Promise` which will resolve to `Error` or `undefined`. I think I'm a little bit wrong about promise part. Could you explain it, please, @vankop?

I'm looking at official TS docs and can't find any description for `Promise` and `Error` types. I found that `Promise<string>` mentioned in examples, but no description. Where do find all available types? Is there good docs for TS?

hudochenkov: Noticed that you wrapped whole type in extra `()`. Is it necessary, because you wrote it multiline? I found this in official docs:

```
/** @type {(s: string, b: boolean) => number} Typescript syntax */
```

No extra parenthesis, maybe because it's a single line.

isaacs:
write-file-atomic 3.0.1 fixes the fd leak.

lib/writeOutputFile.js
vankop: > Noticed that you wrapped whole type in extra `()`. Is it necessary, because you wrote it multiline?

Hm, I think it is not necessary, maybe I mad mistake or prettier did this.. Since you can use parenthesis in TS this is a valid TS too

> Here it says: `writeFileAtomicAsync` accept three parameters (and what these parameters are), and returns `Promise` which will resolve to `Error` or `undefined`. I think I'm a little bit wrong about promise part.

`Promise` is a known type because of defined [libs in config](https://github.com/stylelint/stylelint/blob/master/tsconfig.json#L5)

Place where all predefined libs defined =) - https://github.com/microsoft/TypeScript/blob/master/lib

ES2017 - https://github.com/microsoft/TypeScript/blob/master/lib/lib.es2017.d.ts , here you can see all references to previous es versions in recursive way. ES2017 -> ES2016 -> etc.

Also `///` is a valid syntax, see http://www.typescriptlang.org/docs/handbook/triple-slash-directives.html

Feel free to ask more questions, also visit typescript gitter, you can ask/discuss in real time(I always getting response in 5-20min)

hudochenkov:
@isaacs Amazing! Thank you!

lib/writeOutputFile.js
hudochenkov: Thanks!

PR #4355, opened 2019-10-15
types: add types to requireRule, reportUnknownRuleNames

vankop:

add types to requireRule, reportUnknownRuleNames

nothing

Blocked by:  https://github.com/stylelint/stylelint/pull/4354

### DIFF
--- a/lib/reportUnknownRuleNames.js
+++ b/lib/reportUnknownRuleNames.js
@@ -6,6 +6,10 @@ const rules = require('./rules');
 const MAX_LEVENSHTEIN_DISTANCE = 6;
 const MAX_SUGGESTIONS_COUNT = 3;
 
+/**
+ * @param {string} ruleName
+ * @return {string[]}
+ */
 function extractSuggestions(ruleName) {
 	const suggestions = new Array(MAX_LEVENSHTEIN_DISTANCE);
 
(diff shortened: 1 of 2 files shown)


evilebottnawi:
/cc @vankop @hudochenkov Can we create types in one PR (one commit - one types)? Pros:
- no extra notifications
- good git history
- non blocking PRs (when one types require other types and etc)
- easy to review

vankop:
@evilebottnawi After this PRs I will create 3 more, they will be bigger since all types depends on each other. Types for utility we can create in one PR.

evilebottnawi:
@vankop 

> they will be bigger since all types depends on each other

So should be easy it to do in one PR to avoid blocking better PRs, because if you fix something in an one PR you have to fix it in another PRs and it can be inconvenient and increase work time

> Types for utilities we can create in one PR.

Agree

vankop:
> So should be easy it to do in one PR to avoid blocking better PRs, because if you fix something in an one PR you have to fix it in another PRs and it can be inconvenient and increase work time

@evilebottnawi I will try to make PR with types in such ways:
- to cover as much code as possible
- but keep them relatively easy to review

Obviously, it is hard to find "golden" middle between this points (some reviewers prefer small PRs), so I think to put this on my own vision will be a better decision 👮🏼😁

PR #4356, opened 2019-10-16
Update write-file-atomic

hudochenkov:
`write-file-atomic` doesn't have `graceful-fs` in dependencies anymore and has no reported memory leaks. It's time to use original package, and not fork.

@vankop I have a problem with types. I get the following for `write-file-atomic`:

> Could not find a declaration file for module 'write-file-atomic'. '/Users/ahudochenkov/projects/stylelint/node_modules/write-file-atomic/index.js' implicitly has an 'any' type.
>  Try `npm install @types/write-file-atomic` if it exists or add a new declaration (.d.ts) file containing `declare module 'write-file-atomic';`ts(7016)

There is `@types/write-file-atomic`, but it's for an old version. Now package returns Promise, so types are incorrect.

Is there a better fix than `@ts-ignore`?


vankop:
@hudochenkov take a look on how to define custom types for modules that don't provide types https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#types-typeroots-and-types

Also you can create PR to update types in the same way as @G-Rath did https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38908

Issue #2990, opened 2017-10-27
selector-max-*


amk221:
With `selector-no-type` I could use `// stylelint-disable-line` to add an exception, on the rare occasions where I needed to style by type, e.g:

![screen shot 2017-10-27 at 10 31 17](https://user-images.githubusercontent.com/382061/32097600-fbfcda68-bb01-11e7-9855-271776576943.png)

However, now that it has been deprecated in favour of `selector-max-type` I get notified about the wrong line:

![screen shot 2017-10-27 at 10 26 28](https://user-images.githubusercontent.com/382061/32097634-1694d88a-bb02-11e7-853e-5e971b24483c.png)

I suspect this will happen for other `selector-max-*` rules too.

> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
  "rules": {
    "selector-max-type": 0
  }
}
```

> Which version of stylelint are you using?

e.g. `8.2.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

With `ember-cli-stylelint`, and atom linter

jeddy3:
@amk221 Thanks for the report.

The following:

```css
a {
  color: red;

  &:hover {
    color: blue;
  }
}
```

Produces:

```
4:3 error Expected "a:hover" to have no more than 0 type selectors (selector-max-type)
```

With:

```json
{
  "rules": {
    "selector-max-type": 0,
  }
}
```

This is, a violation for the _resolved_ selector but not for the top level selector. 

@stylelint/core Is that the intended behaviour? I'm not 100% sure what I'd expect to happen here... probably two violations, one for the top level selector (`a`) and one for the _resolved_ nested selector (`a:hover`).

CAYdenberg:
I would say that ideally the line that throws is the one where the selector is declared, not where it referred to (via `&`). Barring that, maybe clarify this in the documentation (for this, and similar rules)?

jeddy3:
> I would say that ideally the line that throws is the one where the selector is declared, not where it referred to (via `&`).

I'm not sure it's that easy as this is a `max` rule that counts the number of selectors. For example, consider:

```
{
  "rules": {
    "selector-max-type": 1
  }
}
```

```css
a {
  color: red;

  & b {
    color: blue;
  }
}
```

Only the selector of the nested rule (resolved to `a b`) violates the `"selector-max-type": 1`.

We could add additional logic to treat `"selector-max-type": 0` as a special case, but I'm not sure it's worth it.

[...]

PR #4357, opened 2019-10-16
False negatives for rules with nested rules at selector-max-*

hudochenkov:

Closes https://github.com/stylelint/stylelint/issues/2990

Now rules report every selector which violates the rule. Before that only deepest selector was reported.


hudochenkov:
> -   Fixed: `selector-max-*` false negatives for rules with nested rules ([#4357](https://github.com/stylelint/stylelint/pull/4357)).

Issue #2835, opened 2017-08-27
Fix false positives for selector lists and shared-line comments in comment-empty-line-before


ntwb:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug where multiple shared-line comments using the `comment-empty-line-before` rule returns false results.

via https://github.com/stylelint/stylelint/tree/master/lib/rules/comment-empty-line-before
> _"If the comment is the very first node in a stylesheet then it is ignored. Shared-line comments are also ignored."_

> Which rule, if any, is this issue related to?

`comment-empty-line-before`

> What CSS is needed to reproduce this issue?

```css
/* comment */
.ok { /* comment */
  color: pink;
}

/* comment */
.foo, /* comment */
.bar { /* comment */
  color: pink;
}
```

> What stylelint configuration is needed to reproduce this issue?

```js
{
  "rules": {
    'comment-empty-line-before': [ 'always', {
      ignore: ['stylelint-commands'],
    } ],
  }
}
```

> Which version of stylelint are you using?

`8.0.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

n/a

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

The following warnings were flagged:
```shell
> test.css            
 8:8  ✖  Expected empty line before comment   comment-empty-line-before
```

ntwb:
Here's a couple of failing tests that should pass:
```diff
diff --git lib/rules/comment-empty-line-before/__tests__/index.js lib/rules/comment-empty-line-before/__tests__/index.js
index be684390..3b867c25 100644
--- lib/rules/comment-empty-line-before/__tests__/index.js
+++ lib/rules/comment-empty-line-before/__tests__/index.js
@@ -76,6 +76,10 @@ testRule(
       {
         code: "a {\n\n  /* comment */\n  color: pink;\n}",
         description: "first-nested with empty line before"
+      },
+      {
+        code: "a {\n /* comment */\n  color: pink;\n}",
+        description: "shared-line comment should be ignored"
       }
     ],
 
@@ -147,6 +151,10 @@ testRule(
         code:
           "a {\ncolor: pink;\n/* stylelint-disable something */\ntop: 0;\n}",
         description: "no newline before a stylelint command comment"
+      },
+      {
[...]

PR #4360, opened 2019-10-16
Fix false positives for shared-line comment as a first child

hudochenkov:

Fixes https://github.com/stylelint/stylelint/issues/2835.

There were false positives when beginning of the rule/at-rule was on a different line than start of the block (`{`).


hudochenkov:
> -   Fixed: `comment-empty-line-before` false positives for selector lists and shared-line comments ([#4360](https://github.com/stylelint/stylelint/pull/4360)).

PR #4361, opened 2019-10-17
Types for stylelint config and postcss result

vankop:

https://github.com/stylelint/stylelint/issues/4245

nothing

### DIFF
--- a/lib/augmentConfig.js
+++ b/lib/augmentConfig.js
@@ -10,9 +10,23 @@ const normalizeRuleSettings = require('./normalizeRuleSettings');
 const path = require('path');
 const requireRule = require('./requireRule');
 
-// - Merges config and configOverrides
-// - Makes all paths absolute
-// - Merges extends
+/** @typedef {import('stylelint').StylelintConfigPlugins} StylelintConfigPlugins */
+/** @typedef {import('stylelint').StylelintConfigProcessors} StylelintConfigProcessors */
+/** @typedef {import('stylelint').StylelintConfigRules} StylelintConfigRules */
+/** @typedef {import('stylelint').StylelintInternalApi} StylelintInternalApi */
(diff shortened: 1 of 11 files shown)


lib/augmentConfig.js
```diff
@@ -98,25 +122,30 @@ function augmentConfigFull(
...
+ * @param {string} configDir
+ * @returns {StylelintConfig}
+ */
 function absolutizePaths(
 	config /*: stylelint$config*/,
 	configDir /*: string*/,
 ) /*: stylelint$config*/ {
 	if (config.ignoreFiles) {
-		config.ignoreFiles = [].concat(config.ignoreFiles).map((glob) => {
+		config.ignoreFiles = /** @type {string[]} */ ([]).concat(config.ignoreFiles).map((glob) => {
```
hudochenkov: What is the difference between `Array<string>` and `string[]`?

types/postcss/index.d.ts
```diff
@@ -0,0 +1,34 @@
+declare module 'postcss/lib/lazy-result' {
```
hudochenkov: Is it possible to use https://github.com/postcss/postcss/blob/master/lib/postcss.d.ts instead?

lib/augmentConfig.js
vankop: I think there is no difference, as I understand array literal `[]` is build-in, `Array` is generic but they follows same logic under the hood.

types/postcss/index.d.ts
vankop: As I see there is no `LazyResult` class in export, so there is no way to use `default` export from postcss. Regarding to [module resolution](https://www.typescriptlang.org/docs/handbook/module-resolution.html) we have to define types ourself

PR #4365, opened 2019-10-18
types for lintSource, createStylelintResult, needlessDisables

vankop:

related to https://github.com/stylelint/stylelint/issues/4245

Here is mutation of `disabledRange` https://github.com/stylelint/stylelint/blob/master/lib/needlessDisables.js#L52 that I handle in types `DisabledRange & {used?: boolean}`, but I think better avoid mutation here

~~I will add some types for utils too. I just need to check that all tests working correctly~~

### DIFF
--- a/lib/augmentConfig.js
+++ b/lib/augmentConfig.js
@@ -402,10 +402,9 @@ function addProcessorFunctions(config /*: stylelint$config*/) /*: stylelint$conf
 			if (processorCache.has(processorKey)) {
 				initializedProcessor = processorCache.get(processorKey);
 			} else {
-				/** @type {[string, Object]} */
 				processorConfig = /** @type {Array<string | Object>} */ ([]).concat(processorConfig);
-				const processorLookup = processorConfig[0];
-				const processorOptions = processorConfig[1];
+				const processorLookup = /** @type {[string, Object]} */ (processorConfig)[0];
+				const processorOptions = /** @type {[string, Object]} */ (processorConfig)[1];
 				let processor = dynamicRequire(processorLookup);
 
 				processor = processor.default || processor;
--- a/lib/createStylelintResult.js
+++ b/lib/createStylelintResult.js
@@ -72,16 +72,29 @@ const _ = require('lodash');
 }
 */
 
+/** @typedef {import('stylelint').PostcssResult} PostcssResult */
+/** @typedef {import('postcss').NodeSource} NodeSource */
+/** @typedef {import('stylelint').StylelintResult} StylelintResult */
+
+/**
+ * @param {import('stylelint').StylelintInternalApi} stylelint
+ * @param {PostcssResult} [postcssResult]
+ * @param {string} [filePath]
+ * @param {import('stylelint').StylelintCssSyntaxError} [cssSyntaxError]
+ * @return {Promise<StylelintResult>}
+ */
 module.exports = function(
 	stylelint /*: stylelint$internalApi*/,
 	postcssResult /*: ?postcssResultType*/,
 	filePath /*: ?string*/,
 	cssSyntaxError /*: ?stylelint$cssSyntaxError*/,
 ) /*: Promise<stylelint$result>*/ {
+	/** @type {StylelintResult} */
 	let stylelintResult;
+	/** @type {string | undefined} */
 	let source;
 
-	if (postcssResult) {
+	if (postcssResult && postcssResult.root) {
 		source = !postcssResult.root.source
 			? undefined
 			: postcssResult.root.source.input.file || postcssResult.root.source.input.id;
@@ -118,7 +131,9 @@ module.exports = function(
 			source,
 			deprecations,
 			invalidOptionWarnings,
-			parseErrors,
+			// TODO TYPES check which types are valid? postcss? stylelint?
+			/* eslint-disable-next-line */
+			parseErrors: /** @type {any} */ (parseErrors),
 			errored: postcssResult.stylelint.stylelintError,
 			warnings: postcssResult.messages.map((message /*: messageType*/) => {
 				return {
@@ -160,7 +175,9 @@ module.exports = function(
 	}
 
 	return stylelint.getConfigForFile(filePath).then((configForFile /*: configForFileType*/) => {
-		const config = configForFile.config;
+		// TODO TYPES handle possible null here
+		const config =
+			/** @type {{ config: import('stylelint').StylelintConfig, filepath: string }} */ (configForFile).config;
 		const file = source || (cssSyntaxError && cssSyntaxError.file);
 
 		if (config.resultProcessors) {
--- a/lib/needlessDisables.js
+++ b/lib/needlessDisables.js
@@ -20,9 +20,18 @@ const _ = require('lodash');
   end?: number,
 }*/
 
+/** @typedef {import('stylelint').RangeType} RangeType */
+/** @typedef {import('stylelint').UnusedRange} UnusedRange */
+/** @typedef {import('stylelint').StylelintDisableOptionsReport} StylelintDisableOptionsReport */
+
+/**
+ * @param {import('stylelint').StylelintResult[]} results
+ * @returns {StylelintDisableOptionsReport}
+ */
 module.exports = function(
 	results /*: Array<stylelint$result>*/,
 ) /*: stylelint$disableOptionsReport*/ {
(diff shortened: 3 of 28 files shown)


lib/needlessDisables.js
```diff
@@ -113,13 +132,11 @@ function isWarningInRange(
...
 ) /*: boolean*/ {
-	const rule = warning.rule;
 	const line = warning.line;
 
 	// Need to check if range.end exist, because line number type cannot be compared to undefined
 	return (
 		range.start <= line &&
-		((range.end !== undefined && range.end >= line) || range.end === undefined) &&
-		(!range.rules || range.rules.indexOf(rule) !== -1)
+		((range.end !== undefined && range.end >= line) || range.end === undefined)
```
vankop: found out from types that `range.rules` can not be defined

lib/augmentConfig.js
```diff
@@ -402,10 +402,9 @@ function addProcessorFunctions(config /*: stylelint$config*/) /*: stylelint$conf
 			if (processorCache.has(processorKey)) {
 				initializedProcessor = processorCache.get(processorKey);
 			} else {
-				/** @type {[string, Object]} */
 				processorConfig = /** @type {Array<string | Object>} */ ([]).concat(processorConfig);
-				const processorLookup = processorConfig[0];
-				const processorOptions = processorConfig[1];
+				const processorLookup = /** @type {[string, Object]} */ (processorConfig)[0];
```
hudochenkov: I noticed that you add parenthesis, where they not needed and no one adds them there, from my experience:

```diff
- processorConfig[0]
+ (processorConfig)[0]
```

```diff
- [].concat()
+ ([]).concat()
```

Is there a reason for this? It makes code a little bit harder to understand, because it's unusual.

lib/createStylelintResult.js
```diff
@@ -118,7 +131,9 @@ module.exports = function(
 			source,
 			deprecations,
 			invalidOptionWarnings,
-			parseErrors,
+			// TODO TYPES check which types are valid? postcss? stylelint?
+			/* eslint-disable-next-line */
+			parseErrors: /** @type {any} */ (parseErrors),
```
hudochenkov: Is it possible to use object shorthand here? Something like this:

```diff
- parseErrors: /** @type {any} */ (parseErrors),
+ /** @type {any} */ parseErrors,
```

Is it possible when it's a TypeScript file and not JSDoc?

lib/needlessDisables.js
```diff
@@ -77,8 +89,10 @@ module.exports = function(
 				// If this range is unused and no equivalent is marked,
 				// mark this range as unused
 				if (!range.used && !alreadyMarkedUnused) {
-					range.unusedRule = rule;
-					unused.ranges.push(_.omit(range, 'strictStart', 'strictEnd'));
+					unused.ranges.push({
+						..._.omit(range, 'strictStart', 'strictEnd'),
```
hudochenkov: Looks like Morse code :) Could you put `_.omit(range, 'strictStart', 'strictEnd')` in a variable so destructuring would look simpler and more understandable, please?

lib/utils/ruleMessages.js
```diff
@@ -7,22 +7,25 @@
...
+				newMessages[messageId] = function() {
+					return `${messageText.apply(null, arguments)} (${ruleName})`;
+				};
+			}
 
-		return newMessages;
-	}, {});
+			return newMessages;
+		},
+		/** @type {{[k: string]: string|Function}} */ ({}),
```
hudochenkov: Why this type is needed here? It's an empty object, which is passed to `reduce` callback. Maybe callback function should have type for its parameters?

lib/utils/validateOptions.js
```diff
@@ -26,15 +28,20 @@ const ignoredOptions = ['severity', 'message'];
...
-module.exports = function(result /*: Object*/, ruleName /*: string*/) /*: boolean*/ {
+module.exports = function(
+	result /*: Object*/,
+	ruleName /*: string*/,
+	...optionDescriptions /*: Object*/
+) /*: boolean*/ {
 	let noErrors = true;
 
-	const optionDescriptions = Array.from(arguments).slice(2);
-
```
hudochenkov: ES6, yay! :)

lib/augmentConfig.js
vankop: `/** @type {any} */` applies to first statement in brackets. It is sort of casting types in typescript in jsdoc, see https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#casts

lib/createStylelintResult.js
vankop: same as I described above, it is form of casting in typescript in jsdoc

lib/utils/ruleMessages.js
vankop: take a look https://www.typescriptlang.org/play/index.html?target=99#code/GYVwdgxgLglg9mABAQwE6uQTwPLAHIgC2ARgKaoDOAFHMQFYBciA3gNoDWTFUqMYA5gF0myMJgC+ASiZgiZVK0EtEAWABQiTYlSkoIVEmz1S0AHTtSmarTqTTOgCYgIpKlWQQIAGkQXMkxABeAD4WdS0IlE9TAAcQCgALKnDI1Ji0ClIASTAoGnoOS0EfAEYABkkU1MkAbnUqyJ09AyiIOo0tcR9FWvVxerVQSFgEFHQsXAIScgoAYTh0EzybJjZORG5eAWEUMSkZOXJFZQam-UNjMz9rejtHZ1d3Tx8-AJCwjtSPCFj4pIbUpp0pRsrl8nRCphiohypVPpFemoAVozi1vu0Il1EMdkBRELJpgpBIj+ki1BAENxEDYgspkEwSj5iEwAEw+CBMADkAHpOYhxO00BgcPhDpRwYihRNRYS5gsdNAJTUgA

lib/needlessDisables.js
vankop: done

lib/utils/ruleMessages.js
hudochenkov: What about this (`acc: number[]`)?

```ts
function arrayOfNumbersCorrect(obj: {[k: string]: any}): number[] { 
    return Object.keys(obj).reduce((acc: number[], key) => {
        acc.push(
            parseInt(obj[key], 10)
        );

        return acc;
    }, []);
}
```

https://www.typescriptlang.org/play/index.html?target=99&ssl=19&ssc=2&pln=11&pc=1#code/GYVwdgxgLglg9mABAQwE6uQTwPLAHIgC2ARgKaoDOAFHMQFYBciA3gNoDWTFUqMYA5gF0myMJgC+ASiZgiZVK0EtEAWABQiTYlSkoIVEmz1S0AHTtSmarTqTTOgCYgIpKlWQQIAGkQXMkxABeAD4WdS0IlE9TAAcQCgALKnDI1Ji0ClIASTAoGnoOS0EfAEYABkkU1MkAbnUqyJ09AyiIOo0tcR9FWvVxerVQSFgEFHQsXAIScgoAYTh0EzybJjZORG5eAWEUMSkZOXJFZQam-UNjMz9rejtHZ1d3TwPphWLfSwCQsI7UjwhYvEkg1Upp0pRsrl8nRCph3uVKr9Ir01CCtGcWv92hEuogeu1+qi1BAENxEDYgspkEwSj5iEwAEw+CBMADkAHpWYhxO00BgcPhDpRoSi+RNBa85gsdNARTUgA

I think it makes more sense.

lib/needlessDisables.js
```diff
@@ -90,7 +90,8 @@ module.exports = function(
 				// mark this range as unused
 				if (!range.used && !alreadyMarkedUnused) {
 					unused.ranges.push({
-						..._.omit(range, 'strictStart', 'strictEnd'),
+						start: range.start,
+						end: range.end,
```
hudochenkov: How can we be sure, that we don't miss any property here? Maybe there are more properties.

lib/utils/ruleMessages.js
vankop: let's do it in this way =) Actually it is all about code style

lib/needlessDisables.js
vankop: Actually I got it from types, but there are tests on this https://github.com/stylelint/stylelint/blob/master/lib/__tests__/needlessDisables.test.js#L40

PR #4369, opened 2019-10-21
Update default Node.js in CI

hudochenkov:
[Node.js 12 is LTS now.](https://nodejs.org/en/blog/release/v12.13.0/)


hudochenkov:
@evilebottnawi I just saw it in a [schedule](https://nodejs.org/en/about/releases/). I think this is the first time with two active LTS. Probably, because of Node.js 8 shorter life span.

PR #4370, opened 2019-10-22
Improve Jest configuration

hudochenkov:
Simplified Jest configuration. Removed a lot of nesting. Removed hacks, we don't need anymore.

### DIFF
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -28,26 +28,25 @@ global.testRule = (rule, schema) => {
 
 					describe(`${util.inspect(schema.config)}`, () => {
 						describe(`${testCase.code}`, () => {
-							spec(testCase.description || 'no description', () => {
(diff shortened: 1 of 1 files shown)


jest-setup.js
```diff
@@ -23,162 +133,28 @@ global.testRule = (rule, schema) => {
...
-										})
-										.then(({ output, warnings }) => {
-											expect(output.results[0].warnings).toEqual(warnings);
-											expect(output.results[0].parseErrors).toEqual([]);
-										});
-								});
-							});
-						});
+				describe(`${util.inspect(schema.config)}`, () => {
+					describe(`${util.inspect(testCase.code)}`, () => {
```
vankop: I think we need to preserve `\n`, `\r`, `\t` here.

hudochenkov: It will show `a\nb` instead of:

```
a
b
```

PR #4372, opened 2019-10-23
Fix #4327

### LINKED ISSUE #4327: Add support for `font-weight` ranges in `@font-face`
> Clearly describe the bug

There are false positives when using `font-weight` inside `@font-face`

https://github.com/stylelint/stylelint/tree/master/lib/rules/font-weight-notation

Reference:

https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face#Descriptors
> `font-weight`
> >  A `font-weight` value. Since Firefox 61 (and in other modern browsers) this also accepts two values to specify a range that is supported by a `font-face`, for example `font-weight: 100 400;`

`font-weight-notation`

```css

@font-face {
	font-family: "Inter var";
	font-weight: 100 900;
	font-style: normal;
	src: url(./assets/fonts/inter/Inter-upright.var.woff2) format("woff2");
}

```

```json
{
  "rules": {
    "font-weight-notation": [ "numeric", {
      ignore: ["relative"],
    } ],
  }
}
```

11.0.0

CLI with `./node_modules/.bin/stylelint style.css --config=.stylelintrc.json`

No

"No warnings to be flagged."

```shell
style.css
260:15  ✖  Expected numeric font-weight notation font-weight-notation
```

fanich37:
It should close #4327 . It's my first contribution to this awesome project so it would be great to get any feedback :blush:

### DIFF
--- a/lib/rules/font-weight-notation/index.js
+++ b/lib/rules/font-weight-notation/index.js
@@ -103,6 +103,16 @@ const rule = function(expectation, options) {
 			const weightValueOffset = decl.value.indexOf(weightValue);
 
 			if (expectation === 'numeric') {
+				if (decl.parent.type === 'atrule' && decl.parent.name.toLowerCase() === 'font-face') {
+					const weightValueNumbers = weightValue.match(/\d{1,}\.?\d+/g);
+
+					if (!weightValueNumbers || !weightValueNumbers.every(isNumbery)) {
+						return complain(messages.expected('numeric'));
+					}
+
+					return;
+				}
+
 				if (!isNumbery(weightValue)) {
 					return complain(messages.expected('numeric'));
 				}
--- a/lib/rules/font-weight-notation/index.js
+++ b/lib/rules/font-weight-notation/index.js
@@ -104,7 +104,7 @@ const rule = function(expectation, options) {
 
 			if (expectation === 'numeric') {
 				if (decl.parent.type === 'atrule' && decl.parent.name.toLowerCase() === 'font-face') {
-					const weightValueNumbers = postcss.list.space(weightValue)
+					const weightValueNumbers = postcss.list.space(weightValue);
 
 					if (!weightValueNumbers.every(isNumbery)) {
 						return complain(messages.expected('numeric'));
--- a/lib/rules/font-weight-notation/index.js
+++ b/lib/rules/font-weight-notation/index.js
@@ -104,9 +104,9 @@ const rule = function(expectation, options) {
 
 			if (expectation === 'numeric') {
 				if (decl.parent.type === 'atrule' && decl.parent.name.toLowerCase() === 'font-face') {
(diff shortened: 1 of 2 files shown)


lib/rules/font-weight-notation/index.js
```diff
@@ -103,6 +103,16 @@ const rule = function(expectation, options) {
 			const weightValueOffset = decl.value.indexOf(weightValue);
 
 			if (expectation === 'numeric') {
+				if (decl.parent.type === 'atrule' && decl.parent.name.toLowerCase() === 'font-face') {
+					const weightValueNumbers = weightValue.match(/\d{1,}\.?\d+/g);
```
evilebottnawi: I think better use `value` parser here instead regexp, because comments can be between values and it will be false positive

hudochenkov: `postcss.list.space()` might works as well. We have it in this rule:

https://github.com/stylelint/stylelint/blob/38d89fefe5569545e9239f7867491043d264dec0/lib/rules/font-weight-notation/index.js#L66

fanich37: Thank you for review @evilebottnawi and @hudochenkov ! The edge case that have to processed properly is
```css
font-weight: 400/* 700 */700;
```
In this case we get the ```weightValue``` equals to ```400700```. Probably we need to analyze the ```decl.raws``` data with ```postcss-value-parser``` in such case.

hudochenkov: I don't think `font-weight: 400/* 700 */700;` would be valid. Let's consider that we have at least one space before or after comment: `font-weight: 400 /* 700 */700;`.

fanich37: In this case I think ```list.space``` would be enough. I updated the PR.

lib/rules/font-weight-notation/index.js
```diff
@@ -103,6 +103,16 @@ const rule = function(expectation, options) {
 			const weightValueOffset = decl.value.indexOf(weightValue);
 
 			if (expectation === 'numeric') {
+				if (decl.parent.type === 'atrule' && decl.parent.name.toLowerCase() === 'font-face') {
+					const weightValueNumbers = postcss.list.space(weightValue)
+
+					if (!weightValueNumbers.every(isNumbery)) {
```
hudochenkov: Could you add `reject` cases, please? Currently this condition is not tested.

fanich37: Done!

fanich37:
> Great job! Thank you!

Wow! Thank you for your help with this PR! And for fixing the lint.

hudochenkov:
> -   Fixed: `font-weight-notation` false positives for `font-weight` ranges in `@font-face` ([#4372](https://github.com/stylelint/stylelint/pull/4372)).

Issue #4374, opened 2019-10-23
Allow configOverrides to be a function


shawnbot:
This is a feature request for the Node API.

In #4331 I described the need to filter the loaded configuration by one or more rules from the CLI to ease refactoring efforts. One way to do this is to make the `configOverrides` option more powerful by allowing it to be a function that _modifies_ the loaded configuration. Filtering the rules would be as simple as:

```js
const rules = ['rule/one', 'rule/two']

stylelint.lint({
  files: 'src/**/*.css',
  configOverrides: config => {
    for (const ruleName of Object.keys(config.rules)) {
      if (!rules.includes(ruleName)) {
        delete config.rules[ruleName]
      }
    }
  }
})
```

Since there's no CLI option for `configOverrides`, it seems like that might actually be a good way to implement a fix for #4331, [right about here](https://github.com/stylelint/stylelint/blob/master/lib/cli.js#L426-L428):

```js
if (cli.flags.quiet) {
  optionsBase.configOverrides.quiet = cli.flags.quiet;
}

if (cli.flags.rules) {
  // `stylelint --rules foo,bar,baz/qux` → {rules: ['foo', 'bar', 'baz/qux']}
  const ruleNames = cli.flags.rules.split(',')
  const {configOverrides} = optionsBase

  optionsBase.configOverrides = config => {
    for (const ruleName of Object.keys(config.rules)) {
      if (!rules.includes(ruleName)) {
        delete config.rules[ruleName]
      }
    }
    // re-apply {quiet: cli.flags.quiet}
    return Object.assign(config, configOverrides)
  }
}
```

It looks like the function would need to be called after the config is [augmented](https://github.com/stylelint/stylelint/blob/e1a40111f530c3c4541ef5b2e4ec4b6f11b95589/lib/augmentConfig.js#L42-L47), and any changes to the config would then need to be resolved ("extended" and "absolutized") again in the same way.

I've done some digging around in the source and I think that I could tackle this if there's any interest. Of course, if there's another way to modify the config after it's been augmented and before it's used in a rule, I would love to hear about that! 🙏

hudochenkov:
Thank you for proposal. We have issue for this feature https://github.com/stylelint/stylelint/issues/3128 We can continue discussion there. We would be happy if you would like to contribute this functionality :)

hudochenkov:
Sorry, I didn't read your proposal carefully enough.

[...]

Issue #4331, opened 2019-10-07
Add --rule and --no-stylelintrc CLI flags


shawnbot:
> What is the problem you're trying to solve?

Stylelint autofixes turn out to be a really good way to refactor large codebases. However, it's painful having to modify your config by hand, and if you're extending an external config it's even trickier.

> What solution would you like to see?

My proposal is a rule filter that could be enabled via the CLI, as in:

```sh
stylelint --rule-filter=some-rule --fix
```

or via the Node API:

```js
const stylelint = require('stylelint')
stylelint.lint({
  files: '**/*.scss',
  fix: true,
  ruleFilter: 'some-rule'
})
```

Ideally, rules would be matched against the fully loaded config. It could be as simple as:

```js
// anymatch allows for the same glob patterns as the "files" option
const anymatch = require('anymatch')
const {rules} = finalConfig
const {ruleFilter} = options
if (ruleFilter) {
  const filterRuleName = anymatch(ruleFilter)
  for (const ruleName of Object.keys(rules)) {
    if (!filterRuleName(ruleName)) {
      delete rules[ruleName]
    }
  }
}
```

#### Prior art
eslint has a `--rule` option, as mentioned in https://github.com/eslint/eslint/issues/6333. Generally I think parity between the two is good, so having a `--rule` option would be nice, too.

There's also a standalone tool called [eslint-nibble](https://github.com/IanVS/eslint-nibble) that provides a nicer UX for tackling lots of rule violations. This is probably the route that I'll go if there isn't any interest in making these changes to stylelint directly. Either way, I'm happy to help with the implementation too.

hudochenkov:
I don't understand the usefulness of this feature.

> However, it's painful having to modify your config by hand

Isn't it the same as modifying CLI command?

shawnbot:
> Isn't it the same as modifying CLI command?

What I'm proposing is that `--rule` would _disable_ all the other rules so that you could focus on a single one. So if your config has dozens (if not hundreds) of rules, then the answer to your question is no.

One problem with doing it in a standalone tool is having to reimplement all of the same configuration loading logic that stylelint does. If that's just using the cosmiconfig API, then I guess it's not a huge deal. But I still think that giving people a way to focus on one or more rules — even just to spot-check a change they're making, particularly in big refactors — would be super useful.

hudochenkov:
[...]

PR #4377, opened 2019-10-24
Config overrides function

shawnbot:
This implements the proposal in #4374. TL;DR: `configOverrides` can be a function that takes a fully resolved config and either modifies it in place and/or returns a new one:

```js
const anymatch = require('anymatch')
module.exports = {
  extends: ['some-other-config'],
  configOverrides: config => {
    // modify, add, or remove plugins config.rules and config.plugins
  }
}
```

### `--rules` CLI option
As a demonstration of how this works, I've also implemented a fix for #4331 with a new CLI flag: `--rules <comma-separated-list>` adds an config override that filters the configuration's `rules` object to include only the named rules in the list.

I haven't added docs for this yet, but I wanted to be sure that this change was welcome before doing so. I think it's super useful and would love not to have to maintain [stylelint-only](https://github.com/shawnbot/stylelint-only). 😁 

### File filters?
This _might_ pave the way for a solution to #3128, but I'm honestly not sure where in stylelint's configuration "lifecycle" the configuration is augmented, and it's entirely possible that this happens before the config is applied to individual files. If so, a follow-up would need to be necessary to either:

1. Apply `configOverrides()` for each file, and pass the file information to it; or
2. Use this function under the hood to implement one of the declarative file override patterns proposed in #3128.

Either way, this mechanism is only available in JS configs, and won't work in JSON or YAML.

### DIFF
--- a/docs/user-guide/node-api.md
+++ b/docs/user-guide/node-api.md
@@ -60,7 +60,9 @@ A partial stylelint configuration object whose properties will override the exis
 
 The difference between the `configOverrides` and `config` options is this: If any `config` object is passed, stylelint does not bother looking for a `.stylelintrc` file and instead just uses whatever `config` object you've passed; but if you want to *both* load a `.stylelintrc` file *and* override specific parts of it, `configOverrides` does just that.
 
-You may also provide a function for `configOverrides` that takes the loaded config object and either returns a new one and/or modifies the existing one in place:
+#### `configOverrides` as a function
+
+The `configOverrides` option can be passed as a function that takes the loaded config object and returns a new one and/or modifies the existing object in place:
 
 ```js
 const {pickBy} = require('lodash')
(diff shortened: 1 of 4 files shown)


vankop:
@shawnbot Thanks for PR, looks good! I think we need to support "smart" merge, too. (maybe in another PR) Could you please create test case with `.stylelintrc` file? We need to support it as well.

@hudochenkov new problem with Windows in CI =( Maybe our previous hack (with commands chaining) is unnecessary now

vankop:
> 1. Apply `configOverrides()` for each file, and pass the file information to it

I think this is more useful

> 2. Use this function under the hood to implement one of the declarative file override patterns proposed in #3128.

I think this is not possible in current stylelint design since stylelint runs with one syntax only (Does not determine syntax for each file). As I can see we can use babel approach here. Something like:
```
{
    ...
   overrides: {
           include: "../components",
           syntax: "css-in-js",
           ...
   }
}
```

So under the hood we can create `overrides` function and call it for each file using first approach API

vankop:
I was meaning to use `mergeConfigs` option https://github.com/stylelint/stylelint/blob/master/lib/augmentConfig.js#L244 , since this feature available only for Node.js API I think it will be enough to just provide this function in public API. Code usage:
```
const configSCSS = require('./config.scss');
const {mergeConfigs} = require('stylelint');
module.exports = {
  extends: ['some-other-config'],
  configOverrides: config => {
    config = mergeConfigs(config, configSCSS)
  }
}
```

hudochenkov:
Thank you for investigating possibilities.

We should discuss `configOverrides` proposal in https://github.com/stylelint/stylelint/issues/4374 first.

Proposed new flag is not discussed yet. https://github.com/stylelint/stylelint/issues/4331

Issue #4378, opened 2019-10-25
Rules are not ignored when using SCSS namespace


kimulaco:
## Overview

If you use the new Sass [@use](https://sass-lang.com/documentation/at-rules/use) rule namespace, the rule will not be ignored and an error will occur.
I know the point to correct, so I will issue a PR!

I got a "font-family-no-missing-generic-family-keyword" error under the following circumstances.

```scss
// src/scss/_font.scss
$font-family-default: Arial, sans-serif;
```

```scss
// src/scss/_button.scss
@use "font" as font;

.button {
  font-family: font.$font-family-default;
}
```

```
src/scss/_button.scss
 4:16  ✖  Unexpected missing generic font family   font-family-no-missing-generic-family-keyword
```

https://github.com/kristerkari/stylelint-scss/issues/387

## Config

- "stylelint": "^11.1.1",
- "stylelint-config-standard": "^19.0.0",
- "stylelint-scss": "^3.12.0"

```js
// .stylelintrc.js
module.exports = {
    extends: 'stylelint-config-standard',
    plugins: [
        'stylelint-scss'
    ],
    rules: {
        'at-rule-no-unknown': null,
        'scss/at-rule-no-unknown': true
    }
}
```

Please refer to this repository for details.
https://github.com/kimulaco/dart-sass-example

PR #4379, opened 2019-10-25
Fix to ignore SCSS namespace

kimulaco:
Issue: Rules are not ignored when using SCSS namespace (https://github.com/stylelint/stylelint/issues/4378)

There is a rule to ignore SCSS variable in "lib/utils/isStandardSyntaxValue.js".
I added a rule there that also ignores the SCSS namespace.

### DIFF
--- a/lib/utils/isStandardSyntaxValue.js
+++ b/lib/utils/isStandardSyntaxValue.js
@@ -14,8 +14,13 @@ module.exports = function(value /*: string*/) /*: boolean*/ {
 		normalizedValue = normalizedValue.slice(1);
 	}
 
-	// SCSS variable
+	// SCSS variable (example $color-primary)
 	if (normalizedValue[0] === '$') {
+		return false;
+	}
+
+	// SCSS namespace (example color.$primary)
+	if (/^.*\.\$/.test(value)) {
 		return false;
 	}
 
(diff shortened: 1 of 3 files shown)


kimulaco:
I added the following test.

- Accept LESS variable
- Accept SCSS variable
- Accept SCSS namespace

Verification, please.

hudochenkov:
Please, add tests to [isStandardSyntaxValue.test.js](https://github.com/stylelint/stylelint/blob/master/lib/utils/__tests__/isStandardSyntaxValue.test.js) as well. And merge `master` into your branch. We fixed issue with our CI in master and would like to see that CI is passes for your PR.

lib/utils/isStandardSyntaxValue.js
```diff
@@ -14,11 +14,16 @@ module.exports = function(value /*: string*/) /*: boolean*/ {
...
 	}
 
-	// SCSS variable
+	// SCSS variable (example $variable)
 	if (normalizedValue[0] === '$') {
 		return false;
 	}
 
+	// SCSS namespace (example namespace.$variable)
+	if (/^.*\.\$/.test(value)) {
```
hudochenkov:
```suggestion
	if (/^.+\.\$/.test(value)) {
```

Because namespace should be at least one character long.

kimulaco: Thanks! I fixed it.

kimulaco:
Added a test to "isStandardSyntaxValue.test.js".
And merged from a master, CI was also successful!

hudochenkov:
> -   Fixed: `font-family-no-missing-generic-family-keyword` false positives for Sass-variables with namespaces ([#4378](https://github.com/stylelint/stylelint/pull/4378)).

PR #4387, opened 2019-10-28
Add types to utils

hudochenkov:

Related to https://github.com/stylelint/stylelint/issues/4245

Types for `balanced-match` are wrong. So I created a mock.

I need help from @vankop with following files:

- [x] `eachDeclarationBlock.js`
- [x] `findFontFamily.js`
- [x] `getNextNonSharedLineCommentNode.js`
- [x] `getPreviousNonSharedLineCommentNode.js`
- [x] `isAutoprefixable.js`
- [x] `isFirstNested.js`
- [x] `isFirstNodeOfRoot.js`
- [x] `isStandardSyntaxAtRule.js`
- [x] `isStandardSyntaxDeclaration.js`
- [x] `isStandardSyntaxRule.js`
- [x] `isStandardSyntaxTypeSelector.js`
- [x] `nextNonCommentNode.js`
- [x] `nodeContextLookup.js`
- [x] `parseSelector.js`
- [x] `transformSelector.js`
- [x] `validateObjectWithArrayProps.js`
- [x] `whitespaceChecker.js`

I didn't know how to add types in some cases. Or there was existing types, like `import('postcss).Node`. But out code relies only on a subset of this type. Or our code relies on custom syntaxes (e. g. Less syntax adds more parameters to nodes and we use them).


lib/utils/functionArgumentsSearch.js
```diff
@@ -21,7 +25,10 @@ module.exports = function(source /*: string*/, functionName /*: string*/, callba
 			target: functionName,
 			functionNames: 'check',
 		},
-		(match) => {
+		(/** @type {{
+			startIndex: number,
+			endIndex: number,
+		}} */ match) => {
```
vankop: @hudochenkov I have provided types to `style-search` instead of providing type to callback

lib/utils/isAutoprefixable.js
```diff
@@ -2,7 +2,9 @@
 'use strict';
 
 const autoprefixer = require('autoprefixer');
+// @ts-ignore
```
vankop: maybe provide some explanation why we ignore this, better in the same way `TODO TYPES <text>`

By the way for `autoprefixer/lib/browsers` and `autoprefixer/lib/prefixes` we can provide types, same as I did for `postcss/*`

lib/utils/isFirstNested.js
```diff
@@ -1,6 +1,11 @@
+// @ts-nocheck
```
vankop: some explanation?

lib/utils/isAutoprefixable.js
hudochenkov: Sorry, I didn't explain good enough. The list of files that I provided, I was hopping that you remove all ignores and fix problems :) I didn't know how to fix, so I ignored things.

lib/utils/isFirstNodeOfRoot.js
```diff
@@ -1,6 +1,11 @@
+// @ts-nocheck
```
vankop: I think we need some explanation here as well

lib/utils/isFirstNested.js
hudochenkov: https://github.com/stylelint/stylelint/pull/4387#discussion_r339769695 😇

lib/utils/isFirstNodeOfRoot.js
hudochenkov: https://github.com/stylelint/stylelint/pull/4387#discussion_r339769695 😇

lib/utils/isAutoprefixable.js
vankop: I will check problematic places later

hudochenkov: They might not be problematic. It's just the first time I'm adding types, and I don't know a lot of things :)

hudochenkov:
@vankop will you have time to help with this PR and remove all ts-ignores? :)

vankop:
@hudochenkov later..

lib/utils/typeGuards.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+/**
+ * @param {import('postcss').Node} node
+ * @returns {node is import('postcss').Root}
+ */
+exports.isRoot = function isRoot(node) {
```
hudochenkov: What is this file for?

Also, I think it should be:

```suggestion
module.exports.isRoot = function isRoot(node) {
```

vankop: for this things https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types

vankop:
As I see `getPreviousNonSharedLineCommentNode` already is fine, so I will mark it as done

vankop:
@hudochenkov types in master are broken

hudochenkov:
Thank you, @vankop, for fixing everything!

@stylelint/contributors would you mind review this. Sorry it's big.

lib/utils/addEmptyLineBefore.js
```diff
@@ -1,11 +1,25 @@
...
+ *
+ * @param {ChildNode} node
+ * @param {'\n' | '\r\n'} newline
+ * @returns {ChildNode}
+ */
 function addEmptyLineBefore(
 	node /*: postcss$node*/,
 	newline /*: '\n' | '\r\n'*/,
 ) /*: postcss$node*/ {
+	if (_.isUndefined(node.raws.before)) {
```
fanich37: Since there is a tend to reduce using `lodash` maybe it's a case to use just
```js
node.raws.before === undefined
```
Under the hood that's all it does.
And also in the `lib/utils/addEmptyLineAfter.js`

lib/utils/beforeBlockString.js
```diff
@@ -1,11 +1,20 @@
...
-	options = options || {};
-
+/** @typedef {import('postcss').Rule} Rule */
+/** @typedef {import('postcss').AtRule} AtRule */
+
+/**
+ * @param {Rule | AtRule} statement
+ * @param {{
+ * 	noRawBefore?: boolean
+ * }} options
```
fanich37: Should be `@returns` type here? Sometimes it's present and sometimes not :)
It's also absent in some files:
`lib/utils/blurComments.js`
`lib/utils/blurInterpolation.js`
`lib/utils/containsString.js`
`lib/utils/declarationValueIndex.js` for example.

lib/utils/addEmptyLineBefore.js
hudochenkov: You're right! `isUndefined` looks strange now 😂

lib/utils/beforeBlockString.js
hudochenkov: Nice catch!

hudochenkov: I was adding missing `@returns` and then I remembered that we don't need to specify all types always, because TypeScript could infer them from code. So I didn't look through all files. Even types that I added in the last commit (https://github.com/stylelint/stylelint/pull/4387/commits/a41fbcb447d62bf8c169c51a8fcc4d8893cba9c4) are probably unnecessary. @vankop correct me if I'm wrong.

fanich37: I think you're right 👍

jeddy3:
It's amazing to see this. When I opened https://github.com/stylelint/stylelint/issues/4245 I had no idea so much progress would be made in just a month. Excellent stuff!

PR #4391, opened 2019-10-30
Remove own implementation of Levenshtein distance utility

hudochenkov:

None.

We are using [`leven`](https://github.com/sindresorhus/leven) for Levenshtein distance and we had own implementation. It's better to have only one implementation for the same thing. And it's good when we don't need to support this implementation :)


vankop:
Oh, I did not find this usage when I was implementing suggestions. Problem of this implementation is that there is no threshold to stop =( I think it is fine without it as well for now, because we have not so much rules (175 now)

PR #4393, opened 2019-10-31
Pass returnValue to formatter with needlessDisables + invalidScopeDisables

### LINKED ISSUE #4136: Fix custom formatters access to needless disables results

I'm using a custom formatter to output logs so that our CI system can automatically create [check annotations](https://github.blog/2018-12-14-introducing-check-runs-and-annotations/#annotations-in-pull-requests) on GitHub. This works great for linting errors, but because the [needless disables results are generated after the formatter is called](https://github.com/stylelint/stylelint/blob/4ca75e78aceabad5424d665003888df2245ff8b1/lib/standalone.js#L283-L291), there's no way to access it without calling the underlying `needlessDisables()` function in the custom formatter.

(Side note: when I `require('stylelint/lib/needlessDisables')` and call it on the results in my custom formatter, things get really weird: the first (and/or last??) rule listed in my `stylelint-disable` comment is _always_ reported as unused, but when I delete it the rule is applied and throws an error. 🙃)

`10.1.0`

CLI with `stylelint --custom-formatter path/to/my/formatter.js`

I expected the formatting function to get `needlessDisables` in one or more of its result objects, or as a second argument.

shawnbot:
Fixes #4136. I've shuffled things around in the standalone `prepareReturnValue()` function so that the `output` is set last and the formatting function gets the `returnValue` object as its second argument, so that needless disables can be accessed like so in a custom formatter:

```js
module.exports = function customFormatter(results, returnValue) {
  const errors = formatErrors(results)
  if (returnValue.needlessDisables) {
    errors.push(...formatNeedlessDisables(returnValue.needlessDisables))
  }
  return errors.join('\n')
}
```

The `returnValue` object includes:
- the `errored` flag
- a `needlessDisables` array if configured with `reportNeedlessDisables: true`
- an `invalidScopeDisables` array if configured with `reportInvalidScopeDisables: true`
- a `maxWarningsExceeded` object if configured with `maxWarnings`

### DIFF
--- a/lib/__tests__/standalone-formatter.test.js
+++ b/lib/__tests__/standalone-formatter.test.js
@@ -50,3 +50,53 @@ it('standalone with invalid formatter option', () => {
 		);
 	});
 });
+
+it('standalone formatter receives {needlessDisables} as second argument', () => {
+	const formatter = jest.fn((results, { needlessDisables }) => {
+		return JSON.stringify({ results, needlessDisables }, null, 2);
+	});
+
+	return standalone({
+		code: `
+			// stylelint-disable yo
+			a {}
+		`,
+		config: configBlockNoEmpty,
+		reportNeedlessDisables: true,
+		formatter,
+	}).then((linted) => {
+		const { output, results, needlessDisables } = linted;
+
+		expect(needlessDisables).not.toBeUndefined();
+		expect(needlessDisables).toHaveLength(1);
+
+		expect(typeof output).toBe('string');
+		expect(formatter).toHaveBeenCalledTimes(1);
+		expect(output).toBe(JSON.stringify({ results, needlessDisables }, null, 2));
+	});
+});
+
+it('standalone formatter receives {invalidScopeDisables} as second argument', () => {
+	const formatter = jest.fn((results, { invalidScopeDisables }) => {
+		return JSON.stringify({ results, invalidScopeDisables }, null, 2);
+	});
+
+	return standalone({
+		code: `
+			/* stylelint-disable indentation */
+			a { color: red; }
+		`,
+		config: configBlockNoEmpty,
+		reportInvalidScopeDisables: true,
+		formatter,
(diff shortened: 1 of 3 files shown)


lib/standalone.js
```diff
@@ -293,11 +293,12 @@ module.exports = function(
...
 			(result) => result.errored || result.parseErrors.length > 0,
 		);
+
 		/** @type {StylelintStandaloneReturnValue} */
 		const returnValue /*: stylelint$standaloneReturnValue*/ = {
 			errored,
-			output: formatterFunction(stylelintResults),
-			results: stylelintResults,
+			results: [],
+			output: '',
```
shawnbot: Initializing these with empty values appears to have been necessary to appease the `lint:types` , which appears to just be running `tsc`. I'm not familiar with TypeScript, but I imagine there's a way to do this in the `@type` comment instead?

vankop: I thinks it is fine to initialize with empty fields

lib/__tests__/standalone-formatter.test.js
```diff
@@ -50,3 +50,53 @@ it('standalone with invalid formatter option', () => {
...
 });
+
+it('standalone formatter receives {needlessDisables} as second argument', () => {
+	const formatter = jest.fn((results, { needlessDisables }) => {
+		return JSON.stringify({ results, needlessDisables }, null, 2);
+	});
+
+	return standalone({
+		code: `
+			// stylelint-disable yo
```
hudochenkov: Please, use standard CSS comments syntax.

shawnbot: Whoops, good catch!
```suggestion
			/* stylelint-disable yo */
```

hudochenkov:
> -   Added: more information for custom formatters ([#4393](https://github.com/stylelint/stylelint/pull/4393)).

PR #4394, opened 2019-11-01
Ignore some properties with 0 value (#4250)

### LINKED ISSUE #4250: False positive `length-zero-no-unit` on `fr` unit
> Clearly describe the bug

There is a false positive for `0fr` when used in multiple values for `grid-template-rows` or similar css-grid properties.

`0fr 1fr` behaves differently to `0 1fr` on this property. The `fr` unit refers to _remaining space_ in the grid. `0` (without a unit) is treated as no height at all.
Using `0 1fr` the second row will overlap the first. Using `0fr 1fr` the height of the first row will be taken in to account and the second row will not overlap. [Here's](https://codepen.io/djfw/pen/ExYbExz) a minimal codepen to display the difference between the two and why this is a false positive.

`length-zero-no-unit`

```css
.grid {
  display: grid;
  grid-template-rows: 0fr 1fr;
}
```

```json
{
  "rules": {
    "length-zero-no-unit": true
  }
}
```

`10.1.0`

`stylelint 'src/sass/**/*.scss'`

I am using scss and `stylelint-scss@3.9.4` but I'm not sure if that matters here (could be wrong of course 🤷‍♂)

No warnings to be flagged.

`Unexpected unit (length-zero-no-unit)`

fanich37:

It should close #4250 . 

Nope, the logic of the PR is quite clear.

### DIFF
--- a/lib/rules/length-zero-no-unit/__tests__/index.js
+++ b/lib/rules/length-zero-no-unit/__tests__/index.js
@@ -214,6 +214,34 @@ testRule(rule, {
 			code: 'a { grId-tEMplate-Rows:0fr  1fr; }',
 			description: 'ignore some grid properties with `fr`',
 		},
+		{
+			code: 'a { font: normal normal 400 16px/0px cursive; }',
+			description: 'ignore line-height in font declaration',
+		},
+		{
+			code: 'a { font: normal normal 400 16px/0 cursive; }',
+			description: 'ignore line-height in font declaration',
+		},
+		{
+			code: 'a { font: normal normal 400 0 / 0px cursive; }',
+			description: 'ignore line-height in font declaration',
+		},
+		{
+			code: 'a { font: normal normal 400 16px / 0px cursive; }',
+			description: 'ignore line-height in font declaration',
+		},
+		{
+			code: 'a { font: normal normal 400 16px/ 0px cursive; }',
+			description: 'ignore line-height in font declaration',
+		},
+		{
+			code: 'a { font: normal normal 400 16px /0px cursive; }',
+			description: 'ignore line-height in font declaration',
+		},
+		{
+			code: 'a { font: normal normal 400 1.2em cursive; }',
+			description: 'do not fail if no line-height in font declaration',
+		},
 	],
 
 	reject: [
@@ -470,6 +498,38 @@ testRule(rule, {
 			message: messages.rejected,
 			line: 1,
 			column: 44,
+		},
+		{
+			code: 'a { font: normal normal 400 0px / 0px cursive; }',
+			fixed: 'a { font: normal normal 400 0 / 0px cursive; }',
+
+			message: messages.rejected,
+			line: 1,
+			column: 30,
+		},
+		{
+			code: 'a { font: normal normal 400 0px /0px cursive; }',
+			fixed: 'a { font: normal normal 400 0 /0px cursive; }',
+
+			message: messages.rejected,
+			line: 1,
+			column: 30,
+		},
+		{
+			code: 'a { font: normal normal 400 0px/0px cursive; }',
+			fixed: 'a { font: normal normal 400 0/0px cursive; }',
+
+			message: messages.rejected,
+			line: 1,
+			column: 30,
+		},
+		{
+			code: 'a { font: normal normal 400 0px/ 0px cursive; }',
+			fixed: 'a { font: normal normal 400 0/ 0px cursive; }',
+
+			message: messages.rejected,
+			line: 1,
+			column: 30,
 		},
 	],
 });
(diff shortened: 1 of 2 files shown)


hudochenkov:
This is strange. Rule file is not formatted by Prettier. But Prettier is run on pre-commit hook. Could you run `npm install` just to be sure git hooks are installed. And run `npm run prettier:fix` to reformat files.

In one of previous PRs you also had lint errors, because file wasn't formatted with Prettier. Maybe you skipping git hook intentionally?

fanich37:
> This is strange. Rule file is not formatted by Prettier. But Prettier is run on pre-commit hook. Could you run `npm install` just to be sure git hooks are installed. And run `npm run prettier:fix` to reformat files.
> 
> In one of previous PRs you also had lint errors, because file wasn't formatted with Prettier. Maybe you skipping git hook intentionally?

Hm... Hooks are installed but doesn't seem to work. Could it depend on an OS platform? I use Win 10 now and can also check it on Ubuntu.

hudochenkov:
> Could it depend on an OS platform?

It shouldn't be a problem because of OS, but who knows, everything is possible.

I have no idea what could be wrong :(

lib/rules/length-zero-no-unit/__tests__/index.js
```diff
@@ -431,3 +494,15 @@ testRule(rule, {
...
 });
+
+testRule(rule, {
+	ruleName,
+	config: [true],
+	fix: true,
+
+	accept: [
+		{
+			code: '.a { grid-template-rows: 0fr 1fr; }',
```
hudochenkov: I think you can put this test where you added other tests, because stylelint configuration is the same.

fanich37: Ooops... Agree. Put it with other tests.
P.S. The last commit I did from Ubuntu and pre-commit hook worked as expected :confused:

lib/rules/length-zero-no-unit/index.js
```diff
@@ -20,6 +20,13 @@ const messages = ruleMessages(ruleName, {
 	rejected: 'Unexpected unit',
 });
 
+const gridPropertiesToIgnore = [
+	'grid-template-columns',
+	'grid-template-rows',
+	'grid-auto-columns',
+	'grid-auto-rows',
+];
```
evilebottnawi: What about move this in https://github.com/stylelint/stylelint/tree/master/lib/reference?

fanich37: Now I think this array can be removed at all. Since ```fr``` unit only works with grid family properties (correct me if it's not so) this check is useless. Do you agree?

hudochenkov: @fanich37 it makes sense! :)

fanich37: @hudochenkov @evilebottnawi removed.

lib/rules/length-zero-no-unit/index.js
```diff
@@ -29,6 +29,10 @@ const rule = function(actual, secondary, context) {
 		}
 
 		root.walkDecls((decl) => {
+			if (decl.prop.toLowerCase() === 'line-height') {
```
evilebottnawi: Not only `line-height` can contains `<length> | <percentage>`, for example https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-width, https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dashoffset and other, you can find them here https://github.com/mdn/data/blob/master/css/properties.json, also let's put them in https://github.com/stylelint/stylelint/tree/master/lib/reference

vankop:
> I think we can create build script to automatically update references required for rules before release

You are totally right =)

lib/rules/length-zero-no-unit/index.js
fanich37: To be honest I completely lost the idea why I'd ignored ```line-height``` 😕 .
This is the same:
```css
line-height: 0;
line-height: 0px;
line-height: 0em;
```
So in case of zero value the dimension unit is useless and the warning should be flagged (or it should be auto-fixed).
The only unit to ignore here is ```fr``` since ```0fr``` and ```0``` mean completely different things.

evilebottnawi: `line-height: 0;` !== `line-height: 0px;`

Сlarifications and examples - https://github.com/cssnano/cssnano/issues/768

fanich37: Make sense.
But what the point to ignore ```stroke-width``` for example. It doesn't matter whether it has unit or not. Example: https://jsfiddle.net/vc2otk06/

evilebottnawi: Oh, yes, my mistake, you are right, don't found other properties with same ability as `line-height` has, let's add tests for this cases:
```css
a { font: normal normal 400 16px/0px cursive; }
```
```css
a { font: normal normal 400 16px/0 cursive; }
```

fanich37: The funny thing is these cases don't break tests in master branch but they have to.
I add some logic to process ```font``` declaration and update test cases.

hudochenkov:
> -   Fixed: `length-zero-no-unit` false positives for `line-height`, and for `fr` units ([#4394](https://github.com/stylelint/stylelint/pull/4394)).

PR #4405, opened 2019-11-10
Ignore less variables (#4200)

### LINKED ISSUE #4200: Fix false positives for Less variables in length-zero-no-unit
> Clearly describe the bug

Automatic fix (` --fix`) for `length-zero-no-unit` does nothing when working with Less variables.

`length-zero-no-unit`

```less
@variable: 0px;
```

```json
{
  "rules": {
    "length-zero-no-unit": true
  }
}
```

`10.1.0`

Node.js

Yes, it's related to Less variables.

Input file to be fixed:
```less
@variable: 0;
```

Input file did not change.

fanich37:

It should close #4200 .

No, it's self explanatory.

### DIFF
--- a/lib/rules/length-zero-no-unit/index.js
+++ b/lib/rules/length-zero-no-unit/index.js
@@ -65,6 +65,10 @@ const rule = function(actual, secondary, context) {
 		});
 
 		root.walkAtRules((atRule) => {
+			if (atRule.variable && !hasBlock(atRule)) {
+				return;
+			}
+
 			const source = hasBlock(atRule)
 				? beforeBlockString(atRule, { noRawBefore: true })
 				: atRule.toString();
--- a/lib/rules/length-zero-no-unit/index.js
+++ b/lib/rules/length-zero-no-unit/index.js
@@ -65,6 +65,7 @@ const rule = function(actual, secondary, context) {
 		});
 
 		root.walkAtRules((atRule) => {
+			// Ignore Less variables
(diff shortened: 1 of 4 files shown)


lib/rules/length-zero-no-unit/index.js
```diff
@@ -65,6 +65,10 @@ const rule = function(actual, secondary, context) {
 		});
 
 		root.walkAtRules((atRule) => {
+			if (atRule.variable && !hasBlock(atRule)) {
```
hudochenkov: Could you add comment before this condition to specify that this condition is for Less variables.

fanich37: Done!

lib/rules/length-zero-no-unit/index.js
```diff
@@ -65,6 +65,11 @@ const rule = function(actual, secondary, context) {
 		});
 
 		root.walkAtRules((atRule) => {
+			// Ignore Less variables
+			if (atRule.variable && !hasBlock(atRule)) {
+				return;
+			}
```
evilebottnawi: Maybe we should move this to utils?

fanich37: Sounds reasonable.
Added util function and some tests.

hudochenkov:
> -   Fixed: `length-zero-no-unit` false positives for Less variables ([#4405](https://github.com/stylelint/stylelint/pull/4405)).

PR #4427, opened 2019-11-21
Ignore `x` unit in some cases

### LINKED ISSUE #3770: Add support for `x` to unit-no-unknown

fanich37:

It should close #3770 .

No, it's self explanatory.

### DIFF
--- a/lib/rules/unit-no-unknown/__tests__/index.js
+++ b/lib/rules/unit-no-unknown/__tests__/index.js
@@ -220,6 +220,22 @@ testRule(rule, {
 			code: "@import 'foo.css'",
 			description: 'ignore non-media and non-variable at-rule',
 		},
+		{
+			code: "a { background-image: image-set('img-1x.jpg' 1x, 'img-2x.jpg' 2x, 'img-3x.jpg' 3x) }",
+			description: 'ignore `x` unit in image-set',
+		},
+		{
+			code: '@media (resolution: 2x) {}',
+			description: 'ignore `x` unit in @media with `resolution`',
+		},
+		{
+			code: '@media ( resOLution:  2x) {}',
+			description: 'ignore `x` unit in @media with `resolution`',
+		},
+		{
+			code: 'a { image-resolution: 1x; }',
+			description: 'ignore `x` unit in image-resolution',
+		},
 	],
 
 	reject: [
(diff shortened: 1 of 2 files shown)


lib/rules/unit-no-unknown/__tests__/index.js
```diff
@@ -220,6 +220,22 @@ testRule(rule, {
...
+			description: 'ignore `x` unit in @media with `resolution`',
+		},
+		{
+			code: '@media ( resOLution:  2x) {}',
+			description: 'ignore `x` unit in @media with `resolution`',
+		},
+		{
+			code: 'a { image-resolution: 1x; }',
+			description: 'ignore `x` unit in image-resolution',
+		},
```
evilebottnawi: Let's add test with `@media (resolution: 2x) and (min-width: 200x) {}`, second should be rejected

fanich37: I had to update logic so that such a case could be handled properly.
Not sure about implementation though.
Any suggestions are welcome.

fanich37:
> I think we can parse media queries and improve logic

That's what I did in the last commit :)
Or what do you mean?

lib/rules/unit-no-unknown/index.js
```diff
@@ -62,10 +63,42 @@ const rule = function(actual, options) {
...
+
+						if (ignoreUnit) {
+							return;
+						}
+					}
+
+					if (
+						node.type === 'decl' &&
+						(node.prop.toLowerCase() === 'image-resolution' || /^image-set/i.test(node.value))
+					) {
```
evilebottnawi: I think we can parse decl here too, because you can use `url("../../media/examples/star.png") left 20x / 15% 60% repeat-x`

fanich37: Nice case. Updated the logic.

hudochenkov:
> -   Added: `unit-no-unknown` support for `x` unit ([#4427](https://github.com/stylelint/stylelint/pull/4427)).

PR #4432, opened 2019-11-22
Remove condescending language in documentation

hudochenkov:
Inspired by [How to remove condescending language from documentation](https://dev.to/meeshkan/how-to-remove-condescending-language-from-documentation-4a5p).


jeddy3:
I like this and will try to adopt it going forward.

PR #4435, opened 2019-11-24
Use consistent code style for JavaScript examples

hudochenkov:
Tried to make all JavaScript examples in a consistent code style.


jeddy3:
It'd be good if we could just get prettier to do this for us. It'd mean adding a tonne of disable commands, though. I'll create a new issue.

PR #4437, opened 2019-11-24
 Use String.startsWith in more places. 

XhmikosR:
Requires #4417 to be merged first and then I'll rebase.

### DIFF
--- a/lib/rules/function-url-no-scheme-relative/index.js
+++ b/lib/rules/function-url-no-scheme-relative/index.js
@@ -25,7 +25,7 @@ const rule = function(actual) {
 			functionArgumentsSearch(decl.toString().toLowerCase(), 'url', (args, index) => {
 				const url = _.trim(args, ' \'"');
 
-				if (!isStandardSyntaxUrl(url) || url.indexOf('//') !== 0) {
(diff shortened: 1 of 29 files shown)


lib/rules/function-url-no-scheme-relative/index.js
```diff
@@ -25,7 +25,7 @@ const rule = function(actual) {
 			functionArgumentsSearch(decl.toString().toLowerCase(), 'url', (args, index) => {
 				const url = _.trim(args, ' \'"');
 
-				if (!isStandardSyntaxUrl(url) || url.indexOf('//') !== 0) {
+				if (!isStandardSyntaxUrl(url) || !url.startsWith('//')) {
```
XhmikosR: This needs double checking but I assume this was the intention

ntwb: If there's no test coverage it would be great to add some to be sure

XhmikosR: It is covered already, but mistakes happen :)

XhmikosR:
I will go through the codebase one more time and check for indexOf and such cases again later.

Issue #4412, opened 2019-11-17
Replace some of lodash functions with native ones


XhmikosR:
Since Node.js 8.7.0 is the minimum supported version, I think it totally makes sense to use native functions when possible.

~~Here's a branch: https://github.com/stylelint/stylelint/compare/master...XhmikosR:lodash~~

Note that I'm doing this in steps. Let me know what you think and if I should continue with this.

~~My plan is also to stop requiring lodash (even though I know it'll be cached the second time) but instead require only the needed functions later.~~

* done in #4413:

  - [x] `chain`
  - [x] `compact`
  - [x] `endsWith`
  - [x] `includes`
  - [x] `isArray`
  - [x] `isFinite`
  - [x] `isUndefined`
  - [x] `omitBy`
  - [x] `repeat`
  - [x] `startsWith`
  - [x] `sum`
  - [x] `toArray`
  - [x] `values`

* these should be sort of "easy"

  - [x] `clone`  #4442
  - [x] `forown` #4441
  - [x] `isNil` #4443
  - [x] `map` #4423 
  - [x] `omit` #4446
  - [x] `uniq` (`Set` should do the job) #4440

* Need more research:

  - [x] `find` ~~(needs further research)~~ #5359
  - [x] `findIndex` ~~(needs further research)~~
  - [x] `last` #5323
  - [x] `partial` ~~(needs further research)~~ #5334
  - [ ] `range`
  - [x] `trim` ~~#4445~~ #5360
  - [x] `isEmpty` #5324

hudochenkov:
Good idea!

XhmikosR:
BTW do we have a benchmark to measure the module load time? I think after I'm done with all this, there should be a small gain in load time.

The module size I doubt it will decrease a lot since many dependencies depend on the whole lodash module anyway.

hudochenkov:
We don't have a benchmark for this. We have benchmarks for specific rules: https://stylelint.io/developer-guide/rules#improving-the-performance-of-a-rule

Also for the whole stylelint lint there are some method described https://github.com/stylelint/stylelint/issues/2454. But bear in mind, that we lazy load all rules/syntaxes and some other dependencies. So depends on configuration load time is different.

XhmikosR:
I updated my original post with a list of the functions that we can change without reinventing the wheel

XhmikosR:
I believe this is as good as it gets without us reinventing the wheel.

We could abstract some common code in the utils later if needed.

[...]

PR #4443, opened 2019-11-25
Remove `_.isNil`

XhmikosR:
Ref #4412

### DIFF
--- a/lib/rules/function-calc-no-invalid/index.js
+++ b/lib/rules/function-calc-no-invalid/index.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
 const declarationValueIndex = require('../../utils/declarationValueIndex');
 const parseCalcExpression = require('../../utils/parseCalcExpression');
 const report = require('../../utils/report');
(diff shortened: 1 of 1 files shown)


lib/rules/function-calc-no-invalid/index.js
```diff
@@ -156,7 +155,8 @@ function getNumber(mathExpression) {
 			? getNumber(right)
 			: null;
 
-	if (_.isNil(leftValue) || _.isNil(rightValue)) {
+	// eslint-disable-next-line eqeqeq
```
ybiquitous: The `smart` option of the `eqeqeq` rule can be helpful.
See https://eslint.org/docs/rules/eqeqeq#smart

```js
/*eslint eqeqeq: ["error", "smart"]*/
```

XhmikosR: Yeah, but I'd rather not touch the ESLint config here since there's an upstream eslint-config-stylelint.

XhmikosR: Also, do note we have at least one more case like this, so it would indeed make sense to switch to `smart` upstream

ybiquitous: I opened the issue: https://github.com/stylelint/eslint-config-stylelint/issues/69

PR #4444, opened 2019-11-25
Update dependencies.

XhmikosR:
* autoprefixer 9.7.2
* postcss 7.0.23
* @types/lodash 4.14.149
* @types/micromatch 3.1.1
* eslint 6.7.1
* lint-staged 9.4.3

Also, remove @types/cosmiconfig; this is no longer needed since it's a stub package.

---

Notes:

1. I know we have dependabot, but this required some manual changes.
2. ~~We save a few packages along the way :)~~
3. I switched to `^` for all dependencies since there is now a lock file so this is just for consistency
4. I kept back flow-bin because the new version introduces some new errors
5. As for typescript, I'm not familiar with it so I didn't touch it either


package.json
```diff
@@ -95,30 +95,29 @@
...
-    "@types/micromatch": "^3.1.0",
+    "@types/micromatch": "^3.1.1",
     "benchmark": "^2.1.4",
     "common-tags": "^1.8.0",
     "cp-file": "^7.0.0",
     "del": "^5.1.0",
-    "eslint": "~6.6.0",
-    "eslint-config-stylelint": "~11.1.0",
+    "eslint": "^6.7.1",
+    "eslint-config-stylelint": "^11.1.0",
```
ntwb: Using `^` caret here has caused us problems in the past, I think continuing to use `~` tillde here would be preffered please

XhmikosR: @ntwb with a lock file this is moot. The behavior does not change, unless there's a security fix

XhmikosR:
The change in jest-setup.js wasn't unnecessary.

XhmikosR:
Anyway, I modified the original patch and just updated the deps. Now it should be like you wanted.

hudochenkov:
@XhmikosR Sorry, we didn't communicate this before. Usually, creator of PR don't merge their PR. Also, new contributors don't do merges until they learn how we handle things in stylelint.

XhmikosR:
@hudochenkov I thought this was OK since it has 2 approvals?

XhmikosR:
And also I addressed your comments.

hudochenkov:
It's more about quality check and consistency. E. g. we ofter rephrase PRs on merge, because original title might not be descriptive for Git history.

XhmikosR:
This was fine AFAICT.

hudochenkov:
You didn't do anything wrong. I appreciate all the work you're doing on stylelint. Just for now it's better that long-time contributors will merge PRs.

XhmikosR:
I understand that, sorry for that. I'll leave the next PRs for you guys to merge even if they are approved :)

PR #4445, opened 2019-11-25
Remove `_.trim`

XhmikosR:
Use `String.trim()` and `String.replace()` instead.

Refs #4412 

What I'm not 100% sure about is if we want to do `trim()` after removing the quotes again. IIRC lodash.trim does this.

### DIFF
--- a/lib/rules/function-url-no-scheme-relative/index.js
+++ b/lib/rules/function-url-no-scheme-relative/index.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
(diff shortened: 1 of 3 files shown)


hudochenkov:
I'm afraid it not that easy to replace, because it does trim recursively https://jsbin.com/hakut/edit?js,console:

```js
console.log(_.trim(' " hello " ', ' "')); // → hello
console.log(_.trim(' """" hello " ', ' "')); // → hello
```

XhmikosR:
Yup, that is what I say in my OP too. I guess we can drop this patch for now.

PR #4446, opened 2019-11-25
Replace `_.omit` with object spread

XhmikosR:
Refs #4412

### DIFF
--- a/lib/augmentConfig.js
+++ b/lib/augmentConfig.js
@@ -53,7 +53,8 @@ function augmentConfigExtended(stylelint, cosmiconfigResult) {
 	if (!cosmiconfigResult) return Promise.resolve(null);
 
 	const configDir = path.dirname(cosmiconfigResult.filepath || '');
-	const cleanedConfig = _.omit(cosmiconfigResult.config, 'ignoreFiles');
+	// eslint-disable-next-line no-unused-vars
+	const { ignoreFiles, ...cleanedConfig } = cosmiconfigResult.config;
 
 	return augmentConfigBasic(stylelint, cleanedConfig, configDir).then((augmentedConfig) => {
 		return {
@@ -162,7 +163,8 @@ function extendConfig(stylelint, config, configDir) {
 
 	const normalizedExtends = Array.isArray(config.extends) ? config.extends : [config.extends];
 
-	const originalWithoutExtends = _.omit(config, 'extends');
+	// eslint-disable-next-line no-unused-vars
+	const { extends: configExtends, ...originalWithoutExtends } = config;
 	const loadExtends = normalizedExtends.reduce((resultPromise, extendLookup) => {
 		return resultPromise.then((resultConfig) => {
 			return loadExtendedConfig(stylelint, resultConfig, configDir, extendLookup).then(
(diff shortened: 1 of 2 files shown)


lib/augmentConfig.js
```diff
@@ -67,7 +67,8 @@ function augmentConfigExtended(
 	if (!cosmiconfigResult) return Promise.resolve(null);
 
 	const configDir = path.dirname(cosmiconfigResult.filepath || '');
-	const cleanedConfig = _.omit(cosmiconfigResult.config, 'ignoreFiles');
+	// eslint-disable-next-line no-unused-vars
```
ntwb: Is there not something we can do with ESLint here instead of this?

ybiquitous: How about using the [`ignoreRestSiblings`](https://eslint.org/docs/rules/no-unused-vars#ignorerestsiblings)?

```js
/*eslint no-unused-vars: ["error", { "ignoreRestSiblings": true }]*/
// 'type' is ignored because it has a rest property sibling.
var { type, ...coords } = data;
```

XhmikosR: No idea, I don't want to spend my time fighting with linting tools :P

That being said, if everyone else agrees with your suggestion, feel free to push it to this branch :)

ybiquitous: I have no strong opinion about this. 😃 
I think that disabling via the magic comment is a low cost solution.

XhmikosR: Feel free to add this in this branch or in the ESLint config here or upstream :)

ybiquitous: Thanks! I pushed 1046a9896284e9535065bc804f99211444c35721.

XhmikosR: Should the suppressions be at the top of the files, though?

ybiquitous: Maybe yes. I read the ESLint document, but "Using Configuration Comments" does not seem to support code blocks to me.

https://eslint.org/docs/user-guide/configuring#using-configuration-comments

XhmikosR: No worries, then. Maybe this should be in our ESLint config, though. Let's wait for others to chime in.

lib/augmentConfig.js
```diff
@@ -1,3 +1,4 @@
+/* eslint no-unused-vars: ["error", { "ignoreRestSiblings": true }] */
```
evilebottnawi: Don't use this on top of file, it is bad practice, use `// eslint-disable-next-line` if you need ignore something on line or use `eslint-disable`/`eslint-enable` only between lines where it is unnecessary

ybiquitous: OK, I think `// eslint-disable-next-line no-unused-vars` is enough in this case. I'll revert and squash it.

ybiquitous: Done.

hudochenkov: @ybiquitous I think we can add it to shared ESLint config. Because rule does make sense. For now we can also have it in this package ESLint configuration until new config is ready.

ybiquitous: @hudochenkov Thanks for your suggestion, but I have no strong opinion on this case. 😅 

If there are some agreements, can I open a new issue on [eslint-config-stylelint](https://github.com/stylelint/eslint-config-stylelint)?

ybiquitous: I opened a new issue: https://github.com/stylelint/eslint-config-stylelint/issues/70

PR #4448, opened 2019-11-26
Add logic to get `fix` setting from config file (#3859)

### LINKED ISSUE #3859: Add support for fix to the configuration object

I always want to use the `--fix` flag, but it's a chore having to use it for every `$ stylelint` invocation.

I would like the ability to specify the CLI flag as a setting in the Stylelint config. That way I can put it in my shareable config and I don't have to specify it each time. The setting could be named `fix`.

---

Related to #3858

fanich37:

It should close #3859 

No, it's self explanatory.

### DIFF
--- a/flow-typed/stylelint.js
+++ b/flow-typed/stylelint.js
@@ -24,7 +24,8 @@ export type stylelint$config = {
   codeProcessors?: Array<Function>,
   resultProcessors?: Array<Function>,
   quiet?: boolean,
-  defaultSeverity?: string
+  defaultSeverity?: string,
+  fix?: boolean
 };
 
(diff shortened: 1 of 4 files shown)


hudochenkov:
I'm not sure this is a good idea to put CLI flags into config. Seems weird. ESLint doesn't have such thing and it's used by far bigger user base.

fanich37:
> I'm not sure this is a good idea to put CLI flags into config. Seems weird. ESLint doesn't have such thing and it's used by far bigger user base.

Seems I was misled by the open issue.
Feel free to close it 👍

hudochenkov:
You understood issue correctly. I wanted to reply there, but I wanted to reply to you. So I replied here.

PR #4450, opened 2019-11-27
Ignore map property name (#4375)

### LINKED ISSUE #4375: FP for unit-no-unknown when SCSS map key starts with digit
Rule `unit-no-unknown` produces FPs for SCSS map key names starting with number

```scss
$foo: ( 1bar  : "value")
```

Config:
```json
{
  "rules": {
     "unit-no-unknown": true,
  }
}
```

Stylelint version `11.1.1`, checked with https://stylelint.io/demo

I expect no warnings to be flagged.

```shell
1:9 error unexpected unknown unit "bar" (unit-no-unknown)
```

fanich37:

It should close #4375 .

No, it's self explanatory.

### DIFF
--- a/lib/rules/unit-no-unknown/index.js
+++ b/lib/rules/unit-no-unknown/index.js
@@ -4,6 +4,7 @@ const _ = require('lodash');
 const atRuleParamIndex = require('../../utils/atRuleParamIndex');
 const declarationValueIndex = require('../../utils/declarationValueIndex');
 const getUnitFromValueNode = require('../../utils/getUnitFromValueNode');
+const isMap = require('../../utils/isMap');
 const keywordSets = require('../../reference/keywordSets');
 const mediaParser = require('postcss-media-query-parser').default;
 const optionsMatches = require('../../utils/optionsMatches');
@@ -17,6 +18,8 @@ const ruleName = 'unit-no-unknown';
 const messages = ruleMessages(ruleName, {
 	rejected: (unit) => `Unexpected unknown unit "${unit}"`,
 });
+
+const mapPropertyNameIndexOffset = 4;
 
 const rule = function(actual, options) {
 	return (root, result) => {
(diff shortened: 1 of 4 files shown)


lib/rules/unit-no-unknown/index.js
```diff
@@ -18,6 +19,8 @@ const messages = ruleMessages(ruleName, {
 	rejected: (unit) => `Unexpected unknown unit "${unit}"`,
 });
 
+const mapPropertyNameIndexOffset = 4;
```
hudochenkov: Could you explain what this variable for, please? I don't understand from the following.

lib/utils/isMap.js
```diff
@@ -0,0 +1,45 @@
+/* @flow */
```
hudochenkov:
```suggestion
```

lib/utils/isMap.js
```diff
@@ -0,0 +1,45 @@
...
+'use strict';
+
+/** @typedef {import('postcss-value-parser').Node} ValueNode */
+
+/**
+ * @param {ValueNode | undefined} valueNode
+ * @returns {boolean}
+ */
+module.exports = function(valueNode /*: Object*/) /*: boolean*/ {
```
hudochenkov:
```suggestion
module.exports = function(valueNode) {
```

We removed Flow in master :)

lib/rules/unit-no-unknown/index.js
fanich37: The map property name (in map cleared from comments & spaces) always has index that being divided by 4 gives 0 (e.g. 0, 4, 8 etc).
The naming of this `const` definitely has to be changed since it's not clear. Any suggestions? Or may be comment would be enough?

lib/utils/isMap.js
fanich37: > We removed Flow in master :)

And this is awesome :+1:

lib/rules/unit-no-unknown/index.js
hudochenkov: I think comment in the code would be enough.

fanich37: Added comment with explanation.

hudochenkov:
> -   Fixed: `unit-no-unknown` false positives for Sass map keys ([#4450](https://github.com/stylelint/stylelint/pull/4450)).

PR #4451, opened 2019-11-27
Remove Flow

hudochenkov:

Related to #4245.

It's a **breaking change**, because we include `flow-typed` into npm package.

I removed Flow in the first commit. And in the second I removed all Flow comments.

---

Just for history: How I removed Flow comments.

Initial idea was to write simple Babel plugin which will remove comments. But after hour digging into plugin development for Babel I couldn't find how to work with comments in Babel ¯﻿\﻿_﻿(﻿ツ﻿)﻿_﻿/﻿¯ Comments aren't tretead as any other AST node for some reason.

So I just used VS Code and search and replace using RegExp :) I was done in 10 minutes.

Here's RegExp I used:

```regexp
\/\*: (.|\n)+?\*\/
\/\*:: (.|\n)+?\*\/
```


hudochenkov:
> -   Removed: types declarations for Flow ([#4451](https://github.com/stylelint/stylelint/pull/4451)).

PR #4456, opened 2019-11-28
Use String.startsWith/endsWith in more places

XhmikosR:
Followup of #4437


XhmikosR:
Pushed a few more changes.

BTW it seems `logFrom()` is duplicated and  could be reused.

PR #4459, opened 2019-11-29
FileCache.js: switch to a Class

XhmikosR:
Better check the [non-whitespace diff](https://github.com/stylelint/stylelint/pull/4459/files?utf8=%E2%9C%93&diff=unified&w=1) for this. Also please confirm this is safe, I didn't test it a lot :)

### DIFF
--- a/lib/utils/FileCache.js
+++ b/lib/utils/FileCache.js
@@ -14,7 +14,15 @@ const DEFAULT_HASH = '';
  * @constructor
  */
 class FileCache {
-	constructor(cacheLocation, hashOfConfig) {
+	constructor(cacheLocation = DEFAULT_CACHE_LOCATION, hashOfConfig = DEFAULT_HASH) {
+		const cacheFile = path.resolve(
+			getCacheFile(cacheLocation, process.cwd()),
+		);
+
+		debug(`Cache file is created at ${cacheFile}`);
+		this._fileCache = fileEntryCache.create(cacheFile);
+		this._hashOfConfig = hashOfConfig;
+	}
 		const cacheFile = path.resolve(
 			getCacheFile(cacheLocation || DEFAULT_CACHE_LOCATION, process.cwd()),
 		);
--- a/lib/utils/FileCache.js
+++ b/lib/utils/FileCache.js
@@ -13,53 +13,55 @@ const DEFAULT_HASH = '';
  * @param {string} [hashOfConfig]
  * @constructor
  */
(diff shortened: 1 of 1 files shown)


lib/utils/FileCache.js
```diff
@@ -13,53 +13,55 @@ const DEFAULT_HASH = '';
  * @param {string} [hashOfConfig]
  * @constructor
  */
-function FileCache(cacheLocation, hashOfConfig) {
-	const cacheFile = path.resolve(
-		getCacheFile(cacheLocation || DEFAULT_CACHE_LOCATION, process.cwd()),
-	);
+class FileCache {
+	constructor(cacheLocation, hashOfConfig) {
```
fanich37:
```suggestion
	constructor(cacheLocation = DEFAULT_CACHE_LOCATION, hashOfConfig = DEFAULT_HASH) {
		const cacheFile = path.resolve(
			getCacheFile(cacheLocation, process.cwd()),
		);

		debug(`Cache file is created at ${cacheFile}`);
		this._fileCache = fileEntryCache.create(cacheFile);
		this._hashOfConfig = hashOfConfig;
	}
```

fanich37: It also should fix the lint issue.

XhmikosR: Thanks, good call :)

fanich37:
Ooops, my bad. I thought multi-line suggestion is supported (or i did smth wrong).
I'll fix it :)

XhmikosR:
@hudochenkov FYI you did a merge commit again

hudochenkov:
Should I did something else instead?

XhmikosR:
I thought you guys only did Squash and merge.

hudochenkov:
You [expressed](https://github.com/stylelint/stylelint/pull/4462#issuecomment-569419098) fair point that all work should be credited to people who did it. So we enabled regular merges. We don't have rebase, because regular merge and squash options allow us to put PR number into commit message. This way we can track down why changes were made and all related discussions.

If all commits in PR were made by one person, then we can safely squash, because author of squashed commits would be author of original commits.

If PR has commit from more than one author, then we'll do merge so all commits are authored by person who did that commit only.

XhmikosR:
Hmm, then I guess you changed the default merge commit message? Because right now it doesn't mention it's a merge commit hence why I was confused :)

hudochenkov:
Yeah, I decided to change default messages to have more descriptive merge commits :)

PR #4460, opened 2019-11-29
Use arg rest/spread when possible.

XhmikosR:
Split from #4453 

Lint needs fixing, any help welcome!

### DIFF
--- a/lib/__tests__/fixtures/processor-fenced-blocks.js
+++ b/lib/__tests__/fixtures/processor-fenced-blocks.js
@@ -2,8 +2,8 @@
 
 const execall = require('execall');
 
-module.exports = function() {
-	const options = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : {};
+module.exports = function(...args) {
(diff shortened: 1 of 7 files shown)


lib/__tests__/fixtures/processor-fenced-blocks.js
```diff
@@ -2,8 +2,8 @@
 
 const execall = require('execall');
 
-module.exports = function() {
-	const options = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : {};
```
hudochenkov: I think it could be replaced with:

```js
module.exports = function(options = {}) {
```

XhmikosR: Agreed, feel free to push it :)

XhmikosR: I pushed this, only lint is left :)

XhmikosR:
BTW we need to add this in our ESLint rules for sure. And test Node.js 8.7.0 to make sure every feature we use still works in the minimum supported version.

PR #4462, opened 2019-11-30
Tighten ESLint rules. 

XhmikosR:
I open this for discussion. Most of these rules, if not all, should be made to the upstream eslint config.

Notes:

* I haven't enabled some of the rules on purpose
* prefer-rest-params needs further changes but IMHO it's totally worth it
* prefer-spread same as above, but we need to test if everything will still work with Node.js 8.7.0; we might need >= 8.10.0

### DIFF
--- a/lib/__tests__/standalone-syntax.test.js
+++ b/lib/__tests__/standalone-syntax.test.js
@@ -249,9 +249,25 @@ it('standalone with postcss-safe-parser', () => {
 
 		expect(results).toHaveLength(6);
 
+		const safeParserExtensionsTest = /\.(css|pcss|postcss)$/i;
+
+		results
+			.filter((result) => !safeParserExtensionsTest.test(result.source))
+			.forEach((result) => {
+				expect(result.warnings).toHaveLength(1);
+
+				const error = result.warnings[0];
+
+				expect(error.line).toBe(1);
+				expect(error.column).toBe(1);
+				expect(error.rule).toBe('CssSyntaxError');
+				expect(error.severity).toBe('error');
+			});
+
 		return Promise.all(
-			results.map((result) => {
-				if (/\.(css|pcss|postcss)$/i.test(result.source)) {
+			results
+				.filter((result) => safeParserExtensionsTest.test(result.source))
+				.map((result) => {
 					const root = result._postcssResult.root;
 
 					expect(result.errored).toBeFalsy();
@@ -259,16 +275,7 @@ it('standalone with postcss-safe-parser', () => {
 					expect(root.toString()).not.toBe(root.source.input.css);
 
 					return promisify(fs.writeFile)(root.source.input.file, root.source.input.css);
-				}
-
-				expect(result.warnings).toHaveLength(1);
-				const error = result.warnings[0];
-
-				expect(error.line).toBe(1);
-				expect(error.column).toBe(1);
-				expect(error.rule).toBe('CssSyntaxError');
-				expect(error.severity).toBe('error');
-			}),
+				}),
 		);
 	});
 });
--- a/lib/__tests__/standalone-syntax.test.js
+++ b/lib/__tests__/standalone-syntax.test.js
@@ -259,15 +259,15 @@ it('standalone with postcss-safe-parser', () => {
 					expect(root.toString()).not.toBe(root.source.input.css);
 
 					return promisify(fs.writeFile)(root.source.input.file, root.source.input.css);
-				} else {
-					expect(result.warnings).toHaveLength(1);
-					const error = result.warnings[0];
-
-					expect(error.line).toBe(1);
-					expect(error.column).toBe(1);
-					expect(error.rule).toBe('CssSyntaxError');
-					expect(error.severity).toBe('error');
 				}
+
+				expect(result.warnings).toHaveLength(1);
+				const error = result.warnings[0];
+
+				expect(error.line).toBe(1);
+				expect(error.column).toBe(1);
+				expect(error.rule).toBe('CssSyntaxError');
+				expect(error.severity).toBe('error');
 			}),
 		);
 	});
--- a/lib/rules/indentation/__tests__/rules.js
+++ b/lib/rules/indentation/__tests__/rules.js
@@ -10,7 +10,7 @@ testRule(rule, {
 
 	accept: [
 		{
-			code: String('/* anything\n' + '    goes\n' + '\t\t\twithin a comment */\n'),
+			code: '/* anything\n' + '    goes\n' + '\t\t\twithin a comment */\n',
 		},
 		{
 			code: 'a { top: 0; } b { top: 1px; }',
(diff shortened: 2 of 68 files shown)


lib/rules/indentation/__tests__/rules.js
```diff
@@ -10,7 +10,7 @@ testRule(rule, {
 
 	accept: [
 		{
-			code: '/* anything\n' + '    goes\n' + '\t\t\twithin a comment */\n' + '',
+			code: String('/* anything\n' + '    goes\n' + '\t\t\twithin a comment */\n'),
```
XhmikosR: This and the next change this in this, I'm not sure it was the original intention since it's already a string?

ntwb:
> ...if everything will still work with Node.js 8.7.0; we might need >= 8.10.0

Node.JS v8 is end-of-life in ~31 days [on 2019-12-31](https://nodejs.org/en/about/releases/)

Happy to plan for this by _ignoring_ that and dropping Node.js support in the next major release

XhmikosR:
Will the next version be a major version one? Regardless, if tests pass on Node.js 8.7.0 (ATM we test the latest 8.x not the minimum supported one), I guess we should be OK.

After we decide on the rules, I will make an upstream PR. IMHO we should also move `parserOptions` to the upstream config.

XhmikosR:
I rebased this. There are still a few errors to fix (any help is welcome) and after we agree on the rules then we can move the newly added rules to our eslint config.

IMHO most of these changes are good, but this stuff is always objective :)

PS. tests seems to pass on Node.js 8.7.0 so we should be covered regarding 8.x compatibility.

package.json
```diff
@@ -162,7 +162,34 @@
...
+        {
+          "object": true,
+          "array": false
+        }
+      ],
+      "prefer-object-spread": "error",
+      "prefer-regex-literals": "error",
+      "prefer-rest-params": "error",
+      "prefer-spread": "error",
+      "prefer-template": "off"
```
hudochenkov:
```suggestion
```

This will conflict with #4429.

package.json
```diff
@@ -162,7 +162,34 @@
...
+        {
+          "allowElseIf": false
+        }
+      ],
+      "no-implicit-coercion": "error",
+      "no-lonely-if": "error",
+      "no-mixed-spaces-and-tabs": "off",
+      "no-unneeded-ternary": "error",
+      "prefer-arrow-callback": "off",
+      "prefer-const": "error",
```
hudochenkov: We have it in main config.

And, frankly, it should be disabled, in my opinion.

package.json
```diff
@@ -162,7 +162,34 @@
...
+        "error",
+        {
+          "allowElseIf": false
+        }
+      ],
+      "no-implicit-coercion": "error",
+      "no-lonely-if": "error",
+      "no-mixed-spaces-and-tabs": "off",
+      "no-unneeded-ternary": "error",
+      "prefer-arrow-callback": "off",
```
hudochenkov: This rule is not enabled by default. Why add this?

package.json
```diff
@@ -162,7 +162,34 @@
...
+      "prefer-const": "error",
+      "prefer-destructuring": [
+        "off",
+        {
+          "object": true,
+          "array": false
+        }
+      ],
```
hudochenkov: Why add disabled rule?

package.json
XhmikosR: I know it will conflict, it's whatever gets merged. I will rebase it anyway

package.json
XhmikosR: I'll remove it since it's the main config. I don't want to go into the trendy discussion about const vs let myself.

package.json
XhmikosR: Because this is what #4453 adds.

package.json
XhmikosR: Because if not, it's enabled and I don't like using destructuring for arrays.

XhmikosR: Oh, you mean the whole rule. It's because I wanted to experiment with enabling or not.

XhmikosR: But I'm not sure if we want this enforced or not. I removed it for now.

lib/rules/shorthand-property-no-redundant-values/index.js
```diff
@@ -35,10 +35,7 @@ function isShorthandProperty(property) {
 	return propertiesWithShorthandNotation.has(property);
 }
 
-function canCondense(top, right) {
-	const bottom = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : null;
-	const left = arguments.length > 3 && arguments[3] !== undefined ? arguments[3] : null;
-
+function canCondense(top, right, bottom, left) {
```
XhmikosR: Are you sure this is equivalent and cannot fail?

XhmikosR:
@hudochenkov if you merge this please don't squash your patches with mine. Feel to squash your patches into one if you prefer, but being you made significant changes, authorship should be preserved.

hudochenkov:
I'm afraid all changes will be squashed into one commit, where I will be Co-Author. We don't have any other methods of merge enabled for this repository.

lib/rules/shorthand-property-no-redundant-values/index.js
hudochenkov: Should be ok. I reviewed relevant code and it should work the same both with `null` and `undefined`.

XhmikosR:
I believe a rebase and merge is possible and if not, it should be possible.

I don't want to show as the author of changes I didn't make myself. Even if the coauthor line is added, it doesn't show who did what.

lib/__tests__/standalone-syntax.test.js
```diff
@@ -249,26 +249,33 @@ it('standalone with postcss-safe-parser', () => {
 
 		expect(results).toHaveLength(6);
 
+		const safeParserExtensionsTest = /\.(css|pcss|postcss)$/i;
```
XhmikosR: Maybe this should be a boolean since in both cases that's what we need?

hudochenkov: We need to test every result. It's possible to create new abstraction, but it will make code more complex and won't solve any problem.

lib/utils/typeGuards.js
```diff
@@ -39,6 +39,6 @@ module.exports.isComment = function isComment(node) {
  * @param {Node} node
  * @returns {node is (Node & {source: NodeSource})}
  */
-module.exports.hasSource = function isComment(node) {
-	return !!node.source;
+module.exports.hasSource = function hasSource(node) {
```
XhmikosR: We could remove the function name here; not sure what's used more in the codebase. This change that ensures consistency.

XhmikosR:
@hudochenkov I have two more rules I forgot to add, but shouldn't need a lot of changes. Can I push them here?

hudochenkov:
Sure, but don't force push :)

XhmikosR:
Alright, I'm done with the rule changes. I have an issue opened in eslint-config-stylelint to backport the new rules.

PR #4466, opened 2019-11-30
Replace request with got

XhmikosR:
Unfortunately this doesn't flatten our deps tree since request is included in other devDependencies:

```
C:\Users\xmr\Desktop\stylelint>npm ls request
stylelint@12.0.0 C:\Users\xmr\Desktop\stylelint
`-- jest@24.9.0
  `-- jest-cli@24.9.0
    `-- jest-config@24.9.0
      `-- jest-environment-jsdom@24.9.0
        `-- jsdom@11.12.0
          `-- request@2.88.0
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -568,6 +568,12 @@
         "any-observable": "^0.3.0"
       }
     },
+    "@sindresorhus/is": {
+      "version": "0.14.0",
+      "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-0.14.0.tgz",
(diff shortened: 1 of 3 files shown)


hudochenkov:
What is the reason to change?

XhmikosR:
https://github.com/request/request/issues/3142

hudochenkov:
Did you try if it works?

XhmikosR:
Yeah, it worked fine, but as always, you should also try :)

Issue #3626, opened 2018-08-29
Add helpful warning for invalid file pattern for Windows users


hudochenkov:
> What is the problem you're trying to solve?

Users on Windows sometimes have problems to start using stylelint, because of invalid CLI file name pattern params. E. g., https://github.com/stylelint/stylelint/issues/3624.

> What solution would you like to see?

I believe we can detect invalid quotes in a file pattern for Windows users, and show a warning with a correct example.

Example for CLI:

```
$ stylelint './file.css'

  Please, use double quotes for a file pattern:

      stylelint "./file.css"
```

Example for npm script:

```json
{
    "scripts": {
        "lint:css": "stylelint './file.css'"
    }
}
```

```
$ npm run lint:css

  Please, use escaped double quotes for a file pattern in your npm scripts:

      stylelint \"./file.css\"
```

I don't know if it's possible to do the second one, but the first one is quite doable.

kajalchaudhary07:
Is this issue still open can i work on this??

Mouvedia:
> Is this issue still open

Yes, it is.

> can i work on this??

Yes, you can.
I've labeled the issue as ready to implement.

EverusLainus:
Is this issue still open can i work on this?

jeddy3:
@EverusLainus Yes, it is. Thank you for wanting to contribute. There have been previous attempts to address this issue. If you have a novel solution, then please open a PR.

VidyasagarAlajangi:
Hi! I'm interested in working on this issue. Is this issue still open for PR ? 
I'm planning to implement Windows-specific file pattern validation with helpful warnings.

CwPranay:
Hi, I’d like to work on this issue. Can I take it?

Shrasti0117:
Hi 👋

I’m interested in contributing to this issue. I’m currently learning and working with Stylelint and would like to try reproducing and debugging this problem.

Could you please assign this issue to me if it’s available? I’d be happy to investigate the cause and attempt a fix.

Thanks!

Mridul-Dev123:
Hi i would like to work on this issue

PR #4471, opened 2019-12-02
Add helpful warning for invalid file pattern for Windows users (#3626)

dom111:
Add information to help out users when single quotes are wrapping the filenames passed to `cli.js`. Only shares help on Windows machines.


lib/utils/noFilesFoundError.js
```diff
@@ -11,9 +13,36 @@ class NoFilesFoundError extends Error {
...
+		const pattern = filteredFileList.join(', ');
 
 		this.message = `No files matching the pattern "${pattern}" were found.`;
+
+		if (
+			os.platform() === 'win32' &&
+			filteredFileList.filter((file) => file.match(/^'.*'$/)).length > 0
+		) {
+			const packageScript = process.env.npm_lifecycle_event || '';
+			const scriptContents = process.env.npm_lifecycle_script || '';
```
evilebottnawi: Bad idea use this env variables

dom111: Ah, this was just a 'nice to have' based on the comments on the original ticket. I was interested in trying to solve this in the 'nicest' way for the end user. I  found the `lifecycle_event` documented at: https://docs.npmjs.com/misc/scripts#current-lifecycle-event but noticed the `lifecycle_script` var in investigation. I originally used dummy data, but it could be less useful. Best to replace with just a static example in your opinion?

lib/utils/noFilesFoundError.js
```diff
@@ -11,9 +13,36 @@ class NoFilesFoundError extends Error {
...
+
+			if (packageScript) {
+				this.message += `
+
+Please use escaped double quotes for a file pattern in your package.json:
+
+        "scripts": {
+                "${packageScript}": "${scriptContents.replace(/'([^']*)'/g, '\\"$1\\"')}"
+        },
```
ntwb: These 3 lines are using spaces instead of tabs

lib/utils/noFilesFoundError.js
jeddy3: Let's use dummy example here to avoid this problematic pattern e.g. 

"Please use escaped double quotes for a file pattern in your package.json e.g. `stylelint \"./file.css\"`

dom111: Updated, thanks for the feedback both!

lib/utils/noFilesFoundError.js
dom111: Updated, thank you!

lib/utils/noFilesFoundError.js
```diff
@@ -11,9 +13,33 @@ class NoFilesFoundError extends Error {
...
+		const filteredFileList = fileList.filter((i) => !i.startsWith('!'));
+		const pattern = filteredFileList.join(', ');
 
 		this.message = `No files matching the pattern "${pattern}" were found.`;
+
+		if (
+			os.platform() === 'win32' &&
+			filteredFileList.filter((file) => file.match(/^'.*'$/)).length > 0
+		) {
+			const packageScript = process.env.npm_lifecycle_event || '';
```
evilebottnawi: What about yarn?

dom111: Hey @evilebottnawi, `yarn` also uses the same env var. Good point though, are there any other package managers we should consider, that perhaps don't adhere to the same spec? I've only ever used `yarn` and `npm` myself.

hudochenkov:
Thank you for adding tests!

`cli.test.js` in `master` is quite different from what you have in your branch. This is why you have a merge conflict. I think you need to revert `cli.test.js` changes. Merge `master` into your branch. And then add tests again following new practices set in `cli.test.js` (practices in old `cli.test.js` were dangerous).

dom111:
> Thank you for adding tests!
> 
> `cli.test.js` in `master` is quite different from what you have in your branch. This is why you have a merge conflict. I think you need to revert `cli.test.js` changes. Merge `master` into your branch. And then add tests again following new practices set in `cli.test.js` (practices in old `cli.test.js` were dangerous).

I didn't notice the conflict to be honest, apologies for that. I've updated now though, hopefully it's all as expected. Thanks!

system-tests/cli/cli.test.js
```diff
@@ -105,4 +106,42 @@ describe('CLI', () => {
...
+			);
+
+			spy.mockRestore();
+		});
+	});
+
+	it('helps windows users with erroneous single quotes', async () => {
+		const spy = jest.spyOn(os, 'platform').mockImplementation(() => 'win32');
+
+		process.env = {};
```
hudochenkov: What this line for? If it's needed for test, maybe we can mock it instead? I'm not sure, but reassigning `process.env` looks dangerous and might cause unexpected issues in other tests.

system-tests/cli/cli.test.js
```diff
@@ -105,4 +106,42 @@ describe('CLI', () => {
...
+				expect.stringContaining(`stylesheet.css'" were found.
+
+Please use escaped double quotes for a file pattern in your package.json, e.g.:
+
+	stylelint \\"file.css\\"
+`),
+			);
+
+			spy.mockRestore();
+		});
```
hudochenkov: I think we can avoid promise here and use async-await instead:

```js
await cli([`'${path.join(__dirname, 'stylesheet.css')}'`]);

expect(console.log).toHaveBeenNthCalledWith(
	1,
	expect.stringContaining(`stylesheet.css'" were found.
Please use escaped double quotes for a file pattern in your package.json, e.g.:
	stylelint \\"file.css\\"
`),
);

spy.mockRestore();
```

Same could be applied to the next test.

This way we will have less nesting, and consistent tests implementation in this file.

system-tests/cli/cli.test.js
dom111: Ah, yeah. I missed this. It's needed because the lifecycle event is set in tests and before it was being replaced (using the unsafe method). I'll look at mocking this too, thank you!

system-tests/cli/cli.test.js
dom111: Makes total sense. Will work on it ASAP!

dom111: I've had a few problems with this because of how we're detecting the run (using the `npm_lifecycle_event` which is perhaps what @evilebottnawi was concerned about. I think I might need to completely re-work the mechanism here. I'll try and take another stab at it as time allows.

jeddy3: > I'll try and take another stab at it as time allows.

Sounds good to me.

jeddy3:
> I think I might need to completely re-work the mechanism here. I'll try and take another stab at it as time allows.

Closing as stale for now. Feel free to reopen if time allows.

PR #3988, opened 2019-03-14
[WIP] chore(update): postcss-selector-parser

evilebottnawi:
> Which issue, if any, is this issue related to?

update parser

> Is there anything in the PR that needs further explanation?

Maybe refactor code base, in some places we can improve speed add more check or simplify code, but it is out of this PR

Very few problems

PR #4491, opened 2019-12-25
Ignore ERB template tags in selectors

### LINKED ISSUE #4489: 'string-no-newline' freezes for template inside css
When the following file is analyzed by `stylelint` with the rule `string-no-newline`, `stylelint` freezes:
https://github.com/rubocop-hq/rubocop/blob/master/assets/output.html.erb

Smaller reproducer, given:
the file `config.json`
```json
{"rules": {"string-no-newline": true}}
```
the file `file.css`
```css
<% COLORS.each do |color| %>
a { }
<% end %>
a { }
```
the following command freezes:
```sh
stylelint --config config.json file.css
```

Stylelint version 10.1.0

hudochenkov:

Fixes #4489.

Problem lies in many places. We are passing selector `<% COLORS.each do |color| %>\na` to `postcss-selector-parser`, and the later falls into endless loop. Which is understandable. This case might be fixed in `postcss-selector-parser`, but it won't happen any time soon. Currently, we're using `postcss-selector-parser@3.1.0`, while the latest version is `6.0.2`. We can't [upgrade](https://github.com/stylelint/stylelint/pull/3988), because there are still some issues in new version of a parser which needs to be resolved in a parser.

So, the only choice we have is to fix problem in stylelint. My solution is not universal, but if will fix the problem. It might create false negatives for `isStandardSyntaxSelector`, when selector like `[data-something="<%"]` is used. I think it's very unlikely to have a selector with `<%` or `%>` in it.

---

There is also a question when to release the fix. The next version of stylelint should be major, because we have some breaking changes in `master`. On the other hand we want to drop Node.js 8 from January 1. Which will require to release new major version. I don't want to release two major versions, as it's bad UX for our users have to update two major versions.

If this fix has to be release as soon as possible, we can create a new branch from latest release tag and cherry pick this PR there, and release patch version from that branch.

### DIFF
--- a/lib/rules/string-no-newline/__tests__/index.js
+++ b/lib/rules/string-no-newline/__tests__/index.js
@@ -109,5 +109,21 @@ testRule(rule, {
       <div style="ttt: auto">actual issues are still reported</div>`,
 			description: 'single quotes',
 		},
+		{
+			// https://github.com/stylelint/stylelint/issues/4489
+			code: `<style>
+				.highlight {
+					padding: 2px;
+				}
+				<%- SEVERITY_COLORS.each do |severity, color| %>
+				.severity.<%= severity %> {
+					color: <%= color %>;
+				}
+				.highlight.<%= severity %> {
+					background-color: <%= color.fade_out(0.4) %>;
+				}
+				</style>`,
+			description: 'ERB templates are ignored',
+		},
 	],
 });
--- a/lib/rules/string-no-newline/index.js
+++ b/lib/rules/string-no-newline/index.js
@@ -2,6 +2,7 @@
 
 const atRuleParamIndex = require('../../utils/atRuleParamIndex');
 const declarationValueIndex = require('../../utils/declarationValueIndex');
+const isStandardSyntaxSelector = require('../../utils/isStandardSyntaxSelector');
 const parseSelector = require('../../utils/parseSelector');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
@@ -40,6 +41,10 @@ const rule = function(actual) {
 		function checkRule(rule) {
 			// Get out quickly if there are no new line
 			if (!reNewLine.test(rule.selector)) {
+				return;
+			}
+
+			if (!isStandardSyntaxSelector(rule.selector)) {
 				return;
 			}
 
--- a/lib/utils/__tests__/isStandardSyntaxSelector.test.js
+++ b/lib/utils/__tests__/isStandardSyntaxSelector.test.js
@@ -69,4 +69,10 @@ describe('isStandardSyntaxSelector', () => {
 		expect(isStandardSyntaxSelector('.foo()[bar]')).toBeFalsy();
 		expect(isStandardSyntaxSelector(".foo()[bar='baz']")).toBeFalsy();
 	});
+
+	it('ERB templates', () => {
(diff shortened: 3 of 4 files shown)


hudochenkov:
What we decided about release? (second part of my top message)

hudochenkov:
@stylelint/core

evilebottnawi:
> If this fix has to be release as soon as possible, we can create a new branch from latest release tag and cherry pick this PR there, and release patch version from that branch.

Sounds good for me

hudochenkov:
> -   Fixed: `string-no-newline` memory leak for ERB templates ([#4491](https://github.com/stylelint/stylelint/pull/4491))

vilchik-elena:
@hudochenkov thanks for the quick fix!
but looks like there [at least] one more place to fix: https://github.com/stylelint/stylelint/blob/42f9bf2185a871cdf2ed529591ba09b59691f6d3/lib/rules/no-descending-specificity/index.js#L79-L81

As far as I understand this condition should be checked before trying to parse selector.

hudochenkov:
@vilchik-elena make sense! Can you send a PR, please?

vilchik-elena:
@hudochenkov I tried to do that, but I fail to reproduce with unit test :(  (while I do with the same `.erb` file and `{ "rules": {   "no-descending-specificity": true } }` as config)

Also I checked that that ~40 usages of `parseSelector`. I'm not sure if all of them should be preceeded with `isStandardSyntaxSelector` check.

hudochenkov:
@vilchik-elena I believe this test will be ok:

```js
testRule(rule, {
	ruleName,
	config: [true],
	syntax: 'html',
	skipBasicChecks: true,

	accept: [
		{
			code: `<style>
				.highlight {
					padding: 2px;
				}
				<%- SEVERITY_COLORS.each do |severity, color| %>
				.severity.<%= severity %> {
					color: <%= color %>;
				}
				.highlight.<%= severity %> {
					background-color: <%= color.fade_out(0.4) %>;
				}
				</style>`,
			description: 'ERB templates are ignored',
		},
	],
});
```

> Also I checked that that ~40 usages of parseSelector. I'm not sure if all of them should be preceeded with isStandardSyntaxSelector check.

Let's fix rules which confirmed has issues.

Maybe we can change [`parseSelector.js`](https://github.com/stylelint/stylelint/blob/master/lib/utils/parseSelector.js) and check for standard syntax there.

vilchik-elena:
Ok, I finally found out what's the problem. I was trying to create patch from master branch. And the problem of `no-descending-specificity` does not exist on master, yet it does exist on fresh release 12.0.1 as for some reason (you know better) commit https://github.com/stylelint/stylelint/commit/0f04539ba1fe1639cbdfe6ba597d6676df2b4019 is not part of 12.0.1. 

So what should be done next according to you @hudochenkov ?

hudochenkov:
Good catch :) In commit you found we added `isStandardSyntaxSelector` check to `isStandardSyntaxRule`, so now `isStandardSyntaxSelector` is not needed in `no-descending-specificity`. https://github.com/stylelint/stylelint/blob/42f9bf2185a871cdf2ed529591ba09b59691f6d3/lib/rules/no-descending-specificity/index.js#L53-L56

vilchik-elena:
@hudochenkov so when this will be part of release?

hudochenkov:
We plan to release in the beginning of January.

PR #4497, opened 2019-12-28
Remove unused condition

### LINKED ISSUE #4328: Incorrect overlapping in disabled ranges
> Clearly describe the bug

During writing types in `~/lib/assignDisabledRanges` I found problem in types [here](https://github.com/stylelint/stylelint/blob/master/lib/assignDisabledRanges.js#L120) in `ranges.end` since ranges is array of disabled ranges it does not have property `end` of type array. If rewrite it to correct syntax `!!_.last(ranges).end`, getting errors in tests with incorrect overlapping in disable ranges [here](https://github.com/stylelint/stylelint/blob/master/lib/__tests__/fixtures/disableOptions/disabled-ranges-2.css), right now `block-no-empty` is closing on line 9 because of `stylelint-disable-line` inserting new disabled range for each rule and next command `stylelint-enable` throws error since all disabled ranges have `end` already

any

e.g.

```css
/* stylelint-disable block-no-empty */
a {}
/* stylelint-enable */

a {} /* stylelint-disable-line color-named */
a { color: red; } /* stylelint-disable-line block-no-empty */

/* stylelint-disable block-no-empty */
a { color: red; } /* stylelint-disable-line */
/* stylelint-enable */
```

```json
{
	"rules": {
			"block-no-empty": true
		},
	}
```

latest

Node.js API

no

I think stylelint must not close `block-no-empty` disabled range on line 9

stylelint assignDisabledRanges throwing error [here](https://github.com/stylelint/stylelint/blob/master/lib/assignDisabledRanges.js#L121)

hudochenkov:

Closes #4328.

I'm not sure I understood the problem correctly ¯﻿\﻿_﻿(﻿ツ﻿)﻿_﻿/﻿¯ But look like removing incorrect condition fixes types issue, and doesn't break any test.

### DIFF
--- a/lib/__tests__/disableRanges.test.js
+++ b/lib/__tests__/disableRanges.test.js
@@ -238,6 +238,24 @@ it('disable single line multiple rules', () => {
 	);
 });
 
+it('disable single line one rule and re-enable all', () => {
+	return testDisableRanges(
+		'a {} /* stylelint-disable-line block-no-empty */\n/* stylelint-enable */',
+	).then((result) => {
+		expect(result.stylelint.disabledRanges).toEqual({
+			all: [],
+			'block-no-empty': [
+				{
+					start: 1,
+					end: 1,
+					strictStart: true,
+					strictEnd: true,
+				},
+			],
+		});
+	});
+});
+
 it('disable next line all rules', () => {
 	return testDisableRanges('/* stylelint-disable-next-line */\na {} ').then((result) => {
 		expect(result.stylelint.disabledRanges).toEqual(
--- a/lib/__tests__/disableRanges.test.js
+++ b/lib/__tests__/disableRanges.test.js
@@ -241,18 +241,11 @@ it('disable single line multiple rules', () => {
 it('disable single line one rule and re-enable all', () => {
 	return testDisableRanges(
 		'a {} /* stylelint-disable-line block-no-empty */\n/* stylelint-enable */',
-	).then((result) => {
-		expect(result.stylelint.disabledRanges).toEqual({
-			all: [],
-			'block-no-empty': [
-				{
-					start: 1,
-					end: 1,
-					strictStart: true,
-					strictEnd: true,
-				},
-			],
-		});
+		() => {
+			throw new Error('should have errored');
+		},
+	).catch((err) => {
+		expect(err.reason).toBe('No rules have been disabled');
 	});
 });
 
--- a/lib/__tests__/needlessDisables.test.js
+++ b/lib/__tests__/needlessDisables.test.js
@@ -81,7 +81,7 @@ it('needlessDisables complex case', () => {
 				ranges: [
 					{ start: 5, end: 5, unusedRule: 'color-named' },
 					{ start: 6, end: 6, unusedRule: 'block-no-empty' },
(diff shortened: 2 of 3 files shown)


vankop:
Sorry for my bad explanation. Your code does not check `end` (right now it does not work also), but I think it is not by design. Example `disabledRanges`:
```
{
   'block-no-empty': [{start: 0, end: 0}]
}
```
for this range command `enable all` should throw error (fall in this condition) by design as I understand. Your fix ignore this behaviour

hudochenkov:
I guess we need tests for this first. Even failing.

hudochenkov:
@vankop do you think we can figure out the problem in the coming week? I would like to release new stylelint soon. Don't want to release another major release in case disabled ranges fix is a breaking change.

vankop:
Actually I don't understand reason of throwing error [here](https://github.com/stylelint/stylelint/blob/master/lib/assignDisabledRanges.js#L166)... Maybe skip command is better

hudochenkov:
We have only one test, where this condition is `true`: https://github.com/stylelint/stylelint/blob/2dfd7796f1d879d84cad1f3a922735fe5e8c6043/lib/__tests__/disableRanges.test.js#L604-L610

vankop:
So we need to add test case:
```
/* stylelint-disable-line block-no-empty */
/* stylelint-enable */
```

and update code  to catch error in this case =)

hudochenkov:
I've added test. This test isn't failing and without removing extra condition. In fact, we have no tests which would use run condition `_.last(ranges.end)` :) So I guess it was just a bug, that this condition was there in the first place :)

lib/__tests__/disableRanges.test.js
```diff
@@ -238,6 +238,24 @@ it('disable single line multiple rules', () => {
 	);
 });
 
+it('disable single line one rule and re-enable all', () => {
+	return testDisableRanges(
+		'a {} /* stylelint-disable-line block-no-empty */\n/* stylelint-enable */',
+	).then((result) => {
+		expect(result.stylelint.disabledRanges).toEqual({
```
vankop: We must expect error here since there is nothing to enable, if I understand correct condition with `No rules have been disabled` error =)

hudochenkov: You're right :)

vankop: Hah, I found a bug in ranges assignment, I will do fix

lib/__tests__/needlessDisables.test.js
```diff
@@ -81,7 +81,7 @@ it('needlessDisables complex case', () => {
 				ranges: [
 					{ start: 5, end: 5, unusedRule: 'color-named' },
 					{ start: 6, end: 6, unusedRule: 'block-no-empty' },
-					{ start: 8, unusedRule: 'block-no-empty' },
+					{ start: 8, end: 10, unusedRule: 'block-no-empty' },
```
vankop: test expectation was wrong..

hudochenkov:
> -   Fixed: overlapping disabled ranges edge case ([#4497](https://github.com/stylelint/stylelint/pull/4497)).

PR #4498, opened 2019-12-28
Fix stylelint config usage

### LINKED ISSUE #4479: report-invalid-scope-disables Cannot read property 'rules' of undefined
Error when running stylelint with the flag `report-invalid-scope-disables`.

```
$ stylelint 'src/**/*.scss' --report-invalid-scope-disables --config .stylelintrc.yml
TypeError: Cannot read property 'rules' of undefined
    at module.exports (./node_modules/stylelint/lib/invalidScopeDisables.js:38:47)
    at prepareReturnValue (./node_modules/stylelint/lib/standalone.js:312:39)
    at ./node_modules/stylelint/lib/standalone.js:286:11
$ stylelint --version
12.0.0
$ ls .stylelintrc.yml 
.stylelintrc.yml
```

```yaml
extends:
  - stylelint-config-standard
```

`12.0.0`

webschik:
There was a comment in `lib/standalone.js:312` that `stylelint._options.config` might be `undefined`, but 
`invalidScopeDisables.js` function always expected an object as 2nd parameter.

For me there were 2 ways to fix that:
* adding  a conditional expression to `stylelint._options.config  || {}` for `invalidScopeDisables` call, or
* changing`invalidScopeDisables` interface to accept `undefined` as 2nd parameter.

I've chosen 2nd option, to prevent adding more conditions in the future in we need to call `invalidScopeDisables` somewhere else with the optional config param.

Closes #4479

### DIFF
--- a/lib/invalidScopeDisables.js
+++ b/lib/invalidScopeDisables.js
@@ -6,10 +6,10 @@
 
 /**
  * @param {import('stylelint').StylelintResult[]} results
- * @param {import('stylelint').StylelintConfig} config
+ * @param {import('stylelint').StylelintConfig|undefined} config
  * @returns {StylelintDisableOptionsReport}
  */
-module.exports = function(results, config) {
+module.exports = function(results, config = {}) {
 	/** @type {StylelintDisableOptionsReport} */
 	const report = [];
 	const usedRules = new Set(Object.keys(config.rules || {}));
--- a/lib/__tests__/invalidScopeDisables.test.js
+++ b/lib/__tests__/invalidScopeDisables.test.js
@@ -93,26 +93,13 @@ it('invalidScopeDisables ignored case', () => {
 });
 
 it('invalidScopeDisables without config', () => {
-	const config = {
-		rules: {
-			'color-named': 'never',
-		},
-	};
-
 	return standalone({
-		config,
-		files: [fixture('disabled-ranges-1.css'), fixture('ignored-file.css')],
+		config: {
+			rules: {},
+		},
+		code: 'a {}',
 		ignoreDisables: true,
-		ignorePath: fixture('.stylelintignore'),
 	}).then((linted) => {
-		expect(invalidScopeDisables(linted.results)).toEqual([
-			{
-				source: fixture('disabled-ranges-1.css'),
-				ranges: [
-					{ start: 1, end: 3, unusedRule: 'color-named' },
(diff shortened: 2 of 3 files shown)


hudochenkov:
I think test should look like this:

```js
it('invalidScopeDisables without config', () => {
	return standalone({
		config: {},
		code: 'a {}',
		ignoreDisables: true,
	}).then((linted) => {
		expect(invalidScopeDisables(linted.results)).toEqual([]);
	});
});
```

Because we want to test when config is empty, and code doesn't have any `stylelint-disable` comments. (I'm not sure about `expect` part).

webschik:
Make sense, but then it should look like

```js
return standalone({
    config: {
        rules: {}
    },
    code: 'a {}',
    ignoreDisables: true,
}).then((linted) => {
    expect(invalidScopeDisables(linted.results)).toEqual([]);
});
```

because there is a check for `config.rules` param in `lib/augmentConfig.js:88`

lib/invalidScopeDisables.js
```diff
@@ -6,10 +6,10 @@
 
 /**
  * @param {import('stylelint').StylelintResult[]} results
- * @param {import('stylelint').StylelintConfig} config
+ * @param {import('stylelint').StylelintConfig|undefined} config
```
hudochenkov:
```suggestion
 * @param {import('stylelint').StylelintConfig} config
```

It works without `undefined`.

webschik: In this case type annotation in JSDoc is not correct, it means the function accepts `config` only with `StylelintConfig` type, but it's not true. I think it's more correct to specify in type that function may accept `undefined`. Also, we can describe `config` as optional - https://jsdoc.app/tags-param.html#optional-parameters-and-default-values, but I think it's a worse approach.

hudochenkov: > By default `null` and `undefined` are subtypes of all other types. That means you can assign `null` and `undefined` to something like `number`.

http://www.typescriptlang.org/docs/handbook/basic-types.html#null-and-undefined

While explicitly specifying `undefined` is correct, it's not necessary.

If you want to keep, it's ok.

hudochenkov:
> -   Fixed: `--report-invalid-scope-disables` crash when no rules specified ([#4498](https://github.com/stylelint/stylelint/pull/4498)).

PR #4500, opened 2019-12-29
Drop Node.js 8

hudochenkov:
End of life on 31st December.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "!lib/testUtils"
   ],
   "engines": {
-    "node": ">=8.7.0"
+    "node": ">=10"
   },
   "dependencies": {
     "autoprefixer": "^9.7.2",
--- a/package.json
+++ b/package.json
@@ -39,9 +39,6 @@
     "!**/__tests__",
     "!lib/testUtils"
   ],
-  "engines": {
-    "node": ">=10"
-  },
   "dependencies": {
     "autoprefixer": "^9.7.2",
     "balanced-match": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -185,7 +185,16 @@
       "prefer-regex-literals": "error",
       "prefer-rest-params": "error",
       "prefer-spread": "error",
-      "prefer-template": "error"
+      "prefer-template": "error",
+      "node/no-unsupported-features/es-builtins": ["error", {
+        "version": ">=10.18.0"
+      }],
+      "node/no-unsupported-features/es-syntax": ["error", {
+        "version": ">=10.18.0"
+      }],
+      "node/no-unsupported-features/node-builtins": ["error", {
+        "version": ">=10.18.0"
+      }]
     }
   },
   "jest": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -40,7 +40,7 @@
     "!lib/testUtils"
   ],
   "engines": {
-    "node": ">=8.7.0"
+    "node": ">=10"
```
evilebottnawi: maybe better start to support with LTS, i.e. 10.13.0

hudochenkov: Some users had problem, when we set minor version too high: https://github.com/stylelint/stylelint/issues/4000. But it's not longer the case:

> Starting with Node 10, AWS Lambda will also automatically update the language minor versions to latest minor version, as specified by https://github.com/nodejs/Release. 

https://aws.amazon.com/about-aws/whats-new/2019/05/aws_lambda_adds_support_for_node_js_v10/

So I guess we can even put latest 10.18.0 as the minimum.

ntwb: I think allowing for edge cases like what happened in #4000 is fine, min v10.0.0 I think will be ok

XhmikosR: Unless  a specific feature is required from a version newer than 10.0.0, we can just go with 10.0.0. Just note that CI tests the latest release line and this can be problematic in some cases. It might be wiser to test the minimum supported Node.js version at the time instead.

hudochenkov:
I would declare minimum supported version as latest Node.js 10 LTS release, but remove `engine` field. It's similar to strategy that many websites use: support last two major browse releases.

We might not be using any Node.js latest features, but I don't want to be locked in old releases. I work on a free software and spent my life to work on it. Users can upgrade their Node.js version with no problem.

If stylelint work on Node.js 10.0.0. It's great, but it won't be supported version. Removing `engine` field will allow to run on “unsupported” Node.js versions without, e. g. `yarn` throwing and error about incompatibility.

Also consider our dependencies which specify they drop Node.js 8. We don't know on what exact version of Node.js 10 they will work (except the very latest release). If we specify supported version as 10.0.0, but our dependency needs 10.2.0, we kind of messing with our users.

ntwb:
Looks like we'd have to update eslint-config-stylelint with this:

https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unsupported-features/es-syntax.md#options
> If you omit the engines field, this rule chooses >=8.0.0 as the configured Node.js version since 8 is the minimum version the community is maintaining (see also Node.js Release Working Group).

```json
{
    "node/no-unsupported-features/es-syntax": ["error", {
        "version": ">=10.0.0",
        "ignores": []
    }]
}
```

hudochenkov:
As per my comment (https://github.com/stylelint/stylelint/pull/4500#issuecomment-570050868) I put latest v10 LTS into ESLint config.

hudochenkov:
@stylelint/contributors does anyone have time to review this, please?

hudochenkov:
> -   Removed: Node.js 8.x support. Node.js 10 is now required. We can guarantee stylelint works on the latest Node.js 10 release. ([#4500](https://github.com/stylelint/stylelint/pull/4500)).

PR #4522, opened 2020-01-09
Update nodejs.yml

imbsky:
Fetch performance has been improved since checkout action v2.0.0.


imbsky:
https://github.com/actions/checkout#checkout-v2

PR #4529, opened 2020-01-09
Tweak CI config

XhmikosR:
Non-whitespace diff: <https://github.com/stylelint/stylelint/pull/4529/files?w=1>

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -14,22 +14,22 @@ env:
 
 jobs:
   lint:
-    name: Lint on Node.js ${{ matrix.node-version }} and ${{ matrix.os }}
+    name: Lint on Node.js ${{ matrix.node }} and ${{ matrix.os }}
 
     runs-on: ${{ matrix.os }}
 
     strategy:
       matrix:
-        node-version: [12]
+        node: [12]
         os: [ubuntu-latest]
 
     steps:
       - uses: actions/checkout@v2
 
-      - name: Use Node.js ${{ matrix.node-version }}
+      - name: Use Node.js ${{ matrix.node }}
         uses: actions/setup-node@v1
         with:
-          node-version: ${{ matrix.node-version }}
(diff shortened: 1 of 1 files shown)


XhmikosR:
I guess I'll do some further testing on my fork first

XhmikosR:
Alright, this works now. I couldn't make this work so I removed the vars. If anyone has any ideas how to make this work feel free to push to this branch.

```yml
name: Coverage on Node.js ${{ env.NODE }}
```

jeddy3:
Can you add a little more about what we hope to achieve with this PR, e.g. is it refactor for code cleanliness or something else?

XhmikosR:
@jeddy3 this simplifies the file and also removes that Windows workaround since it doesn't seem to happen any more.

I also switched to the non-parallel coveralls build since I personally don't see any gain.

Better wait for @hudochenkov to chime in.

.github/workflows/nodejs.yml
```diff
@@ -41,61 +36,46 @@ jobs:
...
   coverage:
-    name: Coverage on Node.js ${{ matrix.node-version }} and ${{ matrix.os }}
-
-    runs-on: ${{ matrix.os }}
-
-    strategy:
-      matrix:
-        node-version: [12]
-        os: [ubuntu-latest]
+    name: Coverage
```
hudochenkov: Could you put name back, please. Because it's a “master” testing suit — latest Node.js on Unix. It's good to see what Node.js version and OS is used.

XhmikosR: Like I said above, this cannot work with environment variables

hudochenkov: Sorry, I missed this comment. Then we shouldn't use these variables. It's not a big of a problem change Node version once a year in multiple places within this file.

XhmikosR: Done, although I find it too repetitive

PR #4530, opened 2020-01-09
Refactor CLI options definition

ybiquitous:
This aims to prepare for type-checking `cli.js` and does NOT type-check actually (I will do it on another PR).

- Camelize option names for type-checking.
- Add missing `type` for some options.
- Sort options alphabetically for readability.
- Add a unit test for the definition (A system test is too much for this purpose).

See <https://github.com/stylelint/stylelint/pull/4504#issuecomment-572354918> for details.

### DIFF
--- a/lib/__tests__/cli.test.js
+++ b/lib/__tests__/cli.test.js
@@ -1,0 +1,140 @@
+'use strict';
+
+const { buildCLI } = require('../cli');
+
+describe('buildCLI', () => {
+	it('flags - default', () => {
+		expect(buildCLI([]).flags).toEqual({
+			allowEmptyInput: false,
+			cache: false,
+			color: false,
+			disableDefaultIgnores: false,
+			fix: false,
+			formatter: 'string',
+			help: false,
+			ignoreDisables: false,
+			printConfig: false,
+			quiet: false,
+			reportInvalidScopeDisables: false,
+			reportNeedlessDisables: false,
+			version: false,
+		});
+	});
+
+	it('flags.allowEmptyInput', () => {
+		expect(buildCLI(['--allow-empty-input']).flags.allowEmptyInput).toBe(true);
+		expect(buildCLI(['--aei']).flags.allowEmptyInput).toBe(true);
+	});
+
+	it('flags.cache', () => {
+		expect(buildCLI(['--cache']).flags.cache).toBe(true);
+	});
+
+	it('flags.cacheLocation', () => {
+		expect(buildCLI(['--cache-location=foo']).flags.cacheLocation).toBe('foo');
+	});
+
+	it('flags.color', () => {
+		expect(buildCLI(['--color']).flags.color).toBe(true);
+		expect(buildCLI(['--no-color']).flags.color).toBe(false);
+	});
+
+	it('flags.config', () => {
+		expect(buildCLI(['--config=/path/to/file']).flags.config).toBe('/path/to/file');
+	});
+
+	it('flags.configBasedir', () => {
+		expect(buildCLI(['--config-basedir=/path/to/dir']).flags.configBasedir).toBe('/path/to/dir');
+	});
+
+	it('flags.customFormatter', () => {
+		expect(buildCLI(['--custom-formatter=foo']).flags.customFormatter).toBe('foo');
+	});
+
+	it('flags.customSyntax', () => {
+		expect(buildCLI(['--custom-syntax=foo']).flags.customSyntax).toBe('foo');
+	});
+
+	it('flags.disableDefaultIgnores', () => {
+		expect(buildCLI(['--disable-default-ignores']).flags.disableDefaultIgnores).toBe(true);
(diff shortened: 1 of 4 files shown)


lib/cli.js
```diff
@@ -264,44 +261,47 @@ const meowOptions = {
...
+		ignorePath: {
 			alias: 'i',
+			type: 'string',
 		},
-		'ignore-pattern': {
+		ignorePattern: {
 			alias: 'ip',
+			type: 'string',
 		},
-		'no-color': {
```
ybiquitous: `no-color` is unneeded so removed (`color` is enough).

lib/utils/checkInvalidCLIOptions.js
```diff
@@ -72,7 +72,7 @@ const buildMessageLine = (invalid, suggestion) => {
  * @return {string}
  */
 module.exports = function checkInvalidCLIOptions(allowedOptions, inputOptions) {
-	const allOptions = buildAllowedOptions(allowedOptions);
+	const allOptions = buildAllowedOptions(allowedOptions).map((opt) => _.kebabCase(opt));
```
fanich37: May be it would be OK to do `map` inside `buildAllowedOptions`?
And also may be it would be more readable if pass `_.kebabCase` in map directly since there are no any other data manipulation? I mean:
```js
.map(_.kebabCase);
```

ybiquitous: Thank you! I've fixed via e4aaa95856a210b8dae20433c4e58cb22a13cd89.

lib/__tests__/cli.test.js
```diff
@@ -0,0 +1,140 @@
+'use strict';
```
hudochenkov: I think these tests are not needed, because they test `meow` package and not stylelint code. `meow` package has own set of tests, we don't need to write and support tests for it.

lib/cli.js
```diff
@@ -220,35 +220,32 @@ const meowOptions = {
         When glob pattern matches no files, the process will exit without throwing an error.
 	`,
 	flags: {
-		'allow-empty-input': {
+		allowEmptyInput: {
```
hudochenkov: Why convert these keys to kebab case?

lib/__tests__/cli.test.js
ybiquitous: Ideally, I agree. But we don't have now system test cases to verify all CLI options.

https://github.com/stylelint/stylelint/blob/d3c4a9f50bb77fd9014e04b84e9b336978239d97/system-tests/cli/cli.test.js#L11

I'd like just to avoid regressions caused by a big refactoring for type-checking. That is, the regression is whether our `meow()` call output is broken or not.
I think easier to write unit test cases instead of system test cases for this purpose.

lib/cli.js
ybiquitous: Assume the following example:

```js
// cli-example.js
const meow = require('meow');

const cli = meow({
  flags: {
    'max-warnings': {
      type: 'number'
    },
  },
});

console.log(cli.flags['max-warnings']);
console.log(cli.flags.maxWarnings);
```

When running the script, we can get this result:

```shell-session
$ node cli-example.js --max-warnings=1
undefined
1
```

But the `meow.Result` type has an incorrect `max-warnings` property instead of `maxWarnngs`.

image

When converting to camel-case:

```diff
  flags: {
-   'max-warnings': {
+   maxWarnings: {
      type: 'number'
```

We can get the correct type:

image

lib/cli.js
```diff
@@ -536,3 +532,16 @@ function handleError(err) {
 
 	process.exit(exitCode); // eslint-disable-line no-process-exit
 }
+
+/**
+ * @param {string[]} argv
+ * @returns {CLIOptions}
+ */
+function buildCLI(argv) {
+	meowOptions.argv = argv;
```
hudochenkov: Maybe instead of mutating `meowOptions` we should do:

```js
return meow({ ...meowOptions, argv });
```

ybiquitous: Make sense. I fixed via 668bbf6a971e7949e9603fa3f51eb40680546370. 👌

lib/utils/checkInvalidCLIOptions.js
```diff
@@ -72,10 +72,10 @@ const buildMessageLine = (invalid, suggestion) => {
  * @return {string}
  */
 module.exports = function checkInvalidCLIOptions(allowedOptions, inputOptions) {
-	const allOptions = buildAllowedOptions(allowedOptions);
+	const allOptions = buildAllowedOptions(allowedOptions).map(_.kebabCase);
 
 	return Object.keys(inputOptions)
```
fanich37: Is it possible to `filter` at first? It seems that in this case there would be no need in `map` at `allOptions` const... I might be wrong :)

ybiquitous: @fanich37 You're right! Thank you. I fixed via 9c9d363116c2a9b7247e5455b33a2a7fda3242f2.

Issue #4533, opened 2020-01-09
Fix tests for function-calc-no-invalid


LolliDepp:
> Clearly describe the bug

With `function-calc-no-invalid` there is a false positive, while invalid syntax is allowed.

> Which rule, if any, is the bug related to?

`function-calc-no-invalid`

> What CSS is needed to reproduce the bug?

This is valid LESS and should be accepted.
```less
// style.less
@myVar: 100px;

a {
  width: @myVar;
}
```

This is invalid LESS, but gets flagged as correct.
```less
// style.less
@myVar: 100px;

a {
  width: @{myVar};
}
```

More than this, the second scenario is used as a positive test case for that rule!
So that's definitely broken.
Stick the LESS in here to verify: https://lesstester.com/

> What stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "function-calc-no-invalid": "on"
  }
}
```

> Which version of stylelint are you using?

`12.0.1`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

VSCode extension running it on the currently open file: `stuartzhang.stylelint-stzhang`

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

Yes, it's related to LESS variables

> What did you expect to happen?

No warnings to be flagged with the valid syntax, warnings flagged with invalid syntax.
I also expect the test cases on the project to not explicitly allow invalid syntax.

> What actually happened (e.g. what warnings or errors did you get)?

Invalid syntax is not flagged, valid syntax is flagged.
Unit tests are wrong.

https://github.com/stylelint/stylelint/blob/d3c4a9f50bb77fd9014e04b84e9b336978239d97/lib/rules/function-calc-no-invalid/__tests__/index.js#L380
That line that is in the list of syntax to parse as valid makes no sense at all, it's just not valid LESS.
@{my-var} is for interpolation in properties names, class name, and internal to string values.
@my-var is for using variables as values for properties.

https://github.com/stylelint/stylelint/blob/d3c4a9f50bb77fd9014e04b84e9b336978239d97/lib/rules/function-calc-no-invalid/__tests__/index.js#L426
This line that is in the list of syntax to parse with errors should be parsed with errors, but not with the `function-calc-no-invalid` error. The error you're supposed to get is "Right curly bracket expected but not found". It just doesn't make any sense and should probably trigger a bunch of errors.

Working on a PR with fixed tests...

LolliDepp:
[...]

PR #4534, opened 2020-01-09
Fix Less test for function-calc-no-invalid

LolliDepp:

This changes 2 test cases to highlight the issues explained on #4533 

The syntax which was replaced and is now gone I don't think has any value inside of that test file.
It's not related to that rule at all.

The syntax that was being allowed was actually just plain wrong, but not for the reason being reported.
The issue is more along the lines of "You cannot directly do variable interpolations in the value section of CSS properties"

See LESS docs... http://lesscss.org/features/#variables-feature

allowed to interpolate in properties: 
```less
@{my-var}: 10px
```
allowed to interpolate in class names: 
```less
.@{my-var} { 
  width: 10px 
}
```
allowed to interpolate in strings: 
```less
background: url("@{base-url}-red.png");
```

### DIFF
--- a/lib/rules/function-calc-no-invalid/__tests__/index.js
+++ b/lib/rules/function-calc-no-invalid/__tests__/index.js
@@ -377,7 +377,7 @@ testRule(rule, {
 		{
 			code: `
       a {
-        width: calc(@{min-value} + @{max-value}*(100vw - @{min-vw}));
+        width: calc(@min-value + @max-value*(100vw - @min-vw));
       }
       `,
 		},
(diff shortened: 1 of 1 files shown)


jeddy3:
@LolliDepp Thank you for fixing the syntax of these tests

> The issue is more along the lines of "You cannot directly do variable interpolations in the value section of CSS properties"

Thank you also for breaking down the valid usage of interpolation in Less so clearly. 

I proposed a fix in https://github.com/stylelint/stylelint/issues/4533 that should address the now failing tests in this PR.

PR #4547, opened 2020-01-14
Added Korean translation of Rules docs.

flamingotiger:

No, it's self explanatory.

### DIFF
--- a/docs/user-guide/rule.ko-KR.md
+++ b/docs/user-guide/rule.ko-KR.md
@@ -12,9 +12,9 @@
 
 다음은 stylelint 내의 모든 규칙 을 [범주](../../VISION.md) 별로 그룹화 한 다음 적용 대상별로 [*그룹화 한 것*](http://apps.workflower.fi/vocabs/css/en) 입니다.
 
--   [오류 가능성](#possible-errors)
--   [언어 기능 제한](#limit-language-features)
--   [문체 이슈](#stylistic-issues)
+-   [오류 가능성](#오류-가능성)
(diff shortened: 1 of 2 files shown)


ntwb:
감사합니다 🤗

This is the first proposal we have had to add localised documentation

I'm not sure this will integrate with stylelint.io to be honest

@stylelint/core What do you think about adding a section in the docs for localised documentation?

My thoughts: I think basic documentation translations are fine, rule translations are a little trickier and harder to maintain, maintenance of docs can also be an issue.

jeddy3:
@flamingotiger Thanks for this.

> This is the first proposal we have had to add localised documentation

We had a Chinese translation from a couple of years ago (See https://github.com/stylelint/stylelint.io/issues/33 and http://stylelint.cn/). It's out of date now.

> I'm not sure this will integrate with stylelint.io to be honest

I agree. Neither Prettier nor ESLint has localised documentation and we have a lot fewer resources.

> What do you think about adding a section in the docs for localised documentation?

I think [awesome stylelint](https://github.com/ntwb/awesome-stylelint) would be a better home. It marks them as unofficial and subject to falling out of date.

jeddy3:
@flamingotiger Thank you very much for writing this. I think hosting this file yourself, perhaps on GitHub, and then adding a link to [awesome stylelint](https://github.com/ntwb/awesome-stylelint) is the best thing to do.

We'll be [promoting awesome-stylelint more in the future](https://github.com/stylelint/stylelint/pull/4561).

PR #4550, opened 2020-01-17
Bump postcss-jsx to 0.36.4

### LINKED ISSUE #4449: Autofix changes files without offenses
> Clearly describe the bug

when running with `--fix` stylelint generates the output that is not equal to the input even when no changes where made

generating output with postcss

```js
import React from 'react'

export default () => {
  const left = 0
  const top = 0

  return (
    <div
      style={{ top, left }}
    >
      Test
    </div>
  )
}

```

empty config

```json
{
  "rules": {
  }
}
```

`stylelint@12.0.0`

CLI with `--fix`

Yes, CSS-in-JS

the file is not changed

```js
import React from 'react'

export default () => {
  const left = 0
  const top = 0

  return (
    <div
      style={{ top:top, left:left }}
    >
      Test
    </div>
  )
}
```

BPScott:

Fixes #4449

Bumps dependency to include this upstream PR (which has tests): https://github.com/gucong3000/postcss-jsx/pull/67

There's a bit of extra noice in the package-lock but that's just the result of me running `npm install` - seems something was outdated previously.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3429,8 +3429,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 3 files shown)


BPScott:
@evilebottnawi done!

XhmikosR:
The package-lock.json changes aren't right FYI. I'm using the latest Node.js 12.x LTS which ships with npm 6.13.4 and most of the changes you have here are reverted.

You should probably make sure you are using a newer npm version.

XhmikosR:
And they are also reverted in dependabot's PRs, see for example https://github.com/stylelint/stylelint/pull/4553/files

ntwb:
Create a new PR updating the lock file once all these are merged sounds like a plan

PR #4564, opened 2020-01-27
Fix TypeError in indentation regression

### LINKED ISSUE #4563: indentation error
I was trying to test the current master with some rules and I hit this by accident:

```
C:\Users\xmr\Desktop\stylelint>npm run benchmark-rule indentation 2

> stylelint@13.0.0 benchmark-rule C:\Users\xmr\Desktop\stylelint
> node scripts/benchmark-rule.js "indentation" "2"

TypeError: Cannot read property 'baseIndentLevel' of null
    at <input css 1>:1:1
    at indentationLevel (C:\Users\xmr\Desktop\stylelint\lib\rules\indentation\index.js:143:64)
    at C:\Users\xmr\Desktop\stylelint\lib\rules\indentation\index.js:54:22
    at C:\Users\xmr\Desktop\stylelint\node_modules\postcss\lib\container.js:135:18
    at Root.each (C:\Users\xmr\Desktop\stylelint\node_modules\postcss\lib\container.js:101:16)
    at Root.walk (C:\Users\xmr\Desktop\stylelint\node_modules\postcss\lib\container.js:131:17)
    at C:\Users\xmr\Desktop\stylelint\lib\rules\indentation\index.js:53:8
    at LazyResult.run (C:\Users\xmr\Desktop\stylelint\node_modules\postcss\lib\lazy-result.js:295:14)
    at LazyResult.asyncTick (C:\Users\xmr\Desktop\stylelint\node_modules\postcss\lib\lazy-result.js:208:26)
    at C:\Users\xmr\Desktop\stylelint\node_modules\postcss\lib\lazy-result.js:250:14
    at new Promise (<anonymous>)
```

If I switch to the v12.0.1 tag, then it works fine. Is this something expected or did it broke accidentally?

XhmikosR:
This was introduced in v13.0.0 and happens because `options` is `null`, thus not falling back to `{}`

Fixes #4563 

We still need to make sure we are not hitting any other similar cases after #4458.

PS. tests didn't catch this :/

### DIFF
--- a/lib/rules/indentation/__tests__/rules.js
+++ b/lib/rules/indentation/__tests__/rules.js
@@ -421,6 +421,27 @@ testRule(rule, {
 
 testRule(rule, {
 	ruleName,
+	config: [2, null],
+
+	accept: [
+		{
+			code: 'a {\n' + '  color: pink;\n' + '}',
+		},
+	],
+	reject: [
+		{
+			code: '\ta {\n' + '  color: pink;\n' + '}',
+			fixed: 'a {\n' + '  color: pink;\n' + '}',
+
+			message: messages.expected('0 spaces'),
+			line: 1,
+			column: 2,
+		},
+	],
+});
+
+testRule(rule, {
+	ruleName,
 	config: ['tab'],
 	fix: true,
 
@@ -508,14 +529,6 @@ testRule(rule, {
 			column: 1,
 		},
 	],
-});
-
-testRule(rule, {
-	ruleName,
-	config: [2, null],
-
-	accept: [],
-	reject: [],
 });
 
 testRule(rule, {
--- a/lib/rules/indentation/__tests__/rules.js
+++ b/lib/rules/indentation/__tests__/rules.js
@@ -512,6 +512,14 @@ testRule(rule, {
 
 testRule(rule, {
 	ruleName,
+	config: [2, null],
+
+	accept: [],
+	reject: [],
+});
+
+testRule(rule, {
+	ruleName,
 	config: [2, { except: ['value'] }],
 	fix: true,
 
--- a/lib/rules/indentation/index.js
+++ b/lib/rules/indentation/index.js
@@ -19,7 +19,9 @@ const messages = ruleMessages(ruleName, {
  *   keyword "tab" for single `\t`
  * @param {object} [options]
  */
-function rule(space, options = {}, context) {
+function rule(space, options, context) {
+	options = options || {};
+
 	const isTab = space === 'tab';
 	const indentChar = isTab ? '\t' : ' '.repeat(space);
 	const warningWord = isTab ? 'tab' : 'space';


XhmikosR:
OK, I was able to make tests fail without this patch, if I do `config: [2, null]` in the indentation tests. We should probably add a new test case.

If anyone wants, feel free to push to this branch.

jeddy3:
>PS. tests didn't catch this :/

It's not the first time, and it won't be the last that, despite all of our tests, a refactor introduced a regression. These things happen :)

XhmikosR:
Should I just duplicate a test case and pass `null` to make sure we cover this case?

Also, I'm not sure if there are more cases like this.

Alternatively we could just revert the default params patch.

jeddy3:
> Should I just duplicate a test case and pass null to make sure we cover this case?

That sounds like a plan.

> Also, I'm not sure if there are more cases like this.

They'll get reported if there are, and we can fix them then.

jeddy3:
> his was introduced in v13.0.0 and happens because options is null

We'll probably want to create a follow-up issue to find out why `null` is being passed and changing that so we can use default parameters.

XhmikosR:
Agreed. Can you help me out with adding a test?

Basically we need a test with whatever input and indentation set to `[whatever, null]` to not throw.

XhmikosR:
```diff
 lib/rules/indentation/__tests__/rules.js | 8 ++++++++
 1 file changed, 8 insertions(+)

diff --git a/lib/rules/indentation/__tests__/rules.js b/lib/rules/indentation/__tests__/rules.js
index 49d81cff..5507bdaf 100644
--- a/lib/rules/indentation/__tests__/rules.js
+++ b/lib/rules/indentation/__tests__/rules.js
@@ -510,6 +510,14 @@ testRule(rule, {
 	],
 });
 
+testRule(rule, {
+	ruleName,
+	config: [2, null],
+
+	accept: [],
+	reject: [],
+});
+
 testRule(rule, {
 	ruleName,
 	config: [2, { except: ['value'] }],

```

Even this seems to trigger the issue on master.

jeddy3:
I'm no longer sure this bug has a high impact on users.

Does it only surface when the benchmarking script is run (because it sets the secondary option to `null`)?

XhmikosR:
That's how I noticed it, not sure if it's high either. Feel free to adapt the labels :)

jeddy3:
Can we add one simple accept case and one simple reject case to your new test, please? Then I think this will be ready for review.

XhmikosR:
@jeddy3 I'm not super familiar with the tests, so I could use some help here.

jeddy3:
> I'm not super familiar with the tests, so I could use some help here.

I pushed a commit.

The only change was adding one accept and one rejects case to your new test to make sure the rule was still accepting and rejecting with a `null` secondary option.

hudochenkov:
I think we should fix the root of the problem, and not the consequences. If we get `options === null`, then we need to replace it with `undefined` in [`lintSource.js`](https://github.com/stylelint/stylelint/blob/23e4dd0b8ddfe4581cdfad5edb39ac201d2efab3/lib/lintSource.js#L170).

XhmikosR:
Agreed, feel free to suggest an alternative patch. I did this quickly because I assumed we wanted to get a patch version fast out.

jeddy3:
I don't think there's a rush on this one as, unlike https://github.com/stylelint/stylelint/pull/4572, it doesn't seem impact users.

PR #4571, opened 2020-01-27
scripts/benchmark-rule.js: Switch to Bootstrap 4.4.1

XhmikosR:
I didn't switch to v4.x here, but let me know if you want to

### DIFF
--- a/scripts/benchmark-rule.js
+++ b/scripts/benchmark-rule.js
@@ -22,7 +22,7 @@ if (!ruleOptions) {
 	throw new Error('You must specify rule options');
 }
 
-const CSS_URL = 'https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.css';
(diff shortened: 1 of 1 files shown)


scripts/benchmark-rule.js
```diff
@@ -22,7 +22,7 @@ if (!ruleOptions) {
 	throw new Error('You must specify rule options');
 }
 
-const CSS_URL = 'https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.css';
+const CSS_URL = 'https://stackpath.bootstrapcdn.com/bootstrap/3.4.1/css/bootstrap.css';
```
evilebottnawi: Maybe we can use 4 version?

XhmikosR: Yeah, I wasn't sure if you wanted that because benchmarks will change a lot

jeddy3: I think we can use 4 as we're looking for real-world tests on modern code.

I don't think we have any historical benchmark results anywhere. Usually, we run the tests before and then after making a performance change to a rule.

Issue #4119, opened 2019-06-17
Fix autofix for nested tagged template literals


hudochenkov:
> Clearly describe the bug

Incorrect output with `--fix` when CSS-in-JS has tagged template literal inside tagged template literal. It's a bug with `postcss-jsx`. I've [opened](https://github.com/gucong3000/postcss-jsx/issues/53) issue in related repository. But it won't be fixed any time soon, because @gucong3000 currently is not available, and the fix will require a lot of time invested.

> Which rule, if any, is the bug related to?

None.

> What CSS is needed to reproduce the bug?

```js
import styled, { css } from 'styled-components';
const Message = styled.p`
	padding: 10px;
	${(props) => css`
		color: #b02d00;
	`}
	${(props2) => css`
		border-color: red;
	`}
`;
```

> What stylelint configuration is needed to reproduce the bug?

Any config as it's related to parser only.

```json
{
  "rules": {
  }
}
```

> Which version of stylelint are you using?

`10.1.0`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

```
stylelint "**/*.js" --fix
```

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

Yes, CSS-in-JS.

> What did you expect to happen?

Output is the same as input, because no rules were run.

> What actually happened (e.g. what warnings or errors did you get)?

Output code is broken JavaScript:

```js
import styled, { css } from 'styled-components';
const Message = styled.p`
	padding: 10px;
	${(props) => css`
		color: #b02d00;
	`}
	${(props2) => css`
		border-color: red;
	`}

		color: #b02d00;
	`}
	${(props2) => css`
		border-color: red;
	`}
`;
```

Temporary solution is to refactor input code to not have nested tagged template literals:

```js
// Before
const Message = styled.p`
	padding: 10px;

	${(props) => css`
		color: #b02d00;
	`}
`;

// After
const Message = styled.p`
	padding: 10px;

	${variant}
`;

const variant = (props) => css`
	color: #b02d00;
`;
```

jeddy3:
We've forked the parser we use for CSS-in-JS [into stylelint organisation](https://github.com/stylelint/postcss-css-in-js), and we're looking for help to fix our fork to close this issue.

See https://github.com/stylelint/stylelint/issues/4574.

jeddy3:
[...]

PR #4561, opened 2020-01-25
Improve documentation

jeddy3:
> Which issue, if any, is this issue related to?

Follow-up to https://github.com/stylelint/stylelint/pull/4548

Closes #4345
Closes #4546
Closes #4362
Closes #4324
Closes #4227
Closes #4118
Closes #3362

> Is there anything in the PR that needs further explanation?

This pull request sets out to:

- break the content down into bite-size pieces 
- use simple and consistent language
- remove needless duplication 
- improve the experience for both first-time users and potential contributors

You can look at the [README](https://github.com/stylelint/stylelint/blob/de6f8ebdb6c4e7c49d790dff7a607112ae12792a/README.md) and [Table of Contents](https://github.com/stylelint/stylelint/blob/de6f8ebdb6c4e7c49d790dff7a607112ae12792a/docs/toc.md) to get a feel for how the structure has evolved.

Notable changes:

1\. Splitting the getting started section into getting started and customising.

The previous content front-loaded an overwhelming amount of choices.  Now there are [three steps](https://github.com/stylelint/stylelint/blob/de6f8ebdb6c4e7c49d790dff7a607112ae12792a/README.md#getting-started) to get a first-time user up and running, followed by the _option_ to customise stylelint (and its usage) to their needs.

The three-step approach is compatible with upcoming features. When https://github.com/stylelint/stylelint/issues/4269 and https://github.com/stylelint/stylelint/issues/3411 land, step 2 becomes `npx stylelint --init` and step 3 `npx stylelint .`. In only three commands, a user can lint _all supported files_ for possible errors and stylistic issues.

2\. Removing an incredible amount of duplication within the usage guides.

The CLI, Node.js API and PostCSS plugin share many options. The previous content duplicated the documentation for these options, often three times with slight variations or inaccuracies in each. Now there is a dedicated [Options section](https://github.com/stylelint/stylelint/blob/c13208503403e13ff7d0a664bb8e46c9329b776d/docs/user-guide/options.md), with each of the usage guidelines building on it.

3\. Encouraging new contributors

The old content lacked guidance for potential new contributors on how to get the project to run. There was a little bit about running tests, but the previous content buried that in the working on rules section. 

[...]

Issue #4560, opened 2020-01-23
Fix autofix transforming media queries when no errors


UltraTempest:
Simillar to https://github.com/stylelint/stylelint/issues/4115 so could be related:

// test.jsx
```
const style = {
  maxWidth: 10,
  '@media(max-width:930px)': {
    width: '64%',
  },
}
```

stylelint --syntax sass "test.jsx" -- no output.
stylelint --syntax sass "test.jsx" --fix edits the file anyway:

// *test.jsx
```
const style = {
   maxWidth: 10,
  '@media(maxWidth:930px)': {
    width: '64%',
  },
}
```

am I missing something?

> Which rule, if any, is the bug related to?

No rule

> What CSS is needed to reproduce the bug?
```
const style = {
   maxWidth: 10,
  '@media(max-width:930px)': {
    width: '64%',
  },
}
```
in a js or jsx file

> What stylelint configuration is needed to reproduce the bug?

```
module.exports = {
  extends: ['stylelint-config-recommended', 'stylelint-config-rational-order'],
  ignoreFiles: ['**/storybook-static/**'],
  rules: {
    'at-rule-no-unknown': null,
    'plugin/rational-order': [
      true,
      {
        'border-in-box-model': true,
        'empty-line-between-groups': false,
      },
    ],
  },
};

```

> Which version of stylelint are you using?

`13.0.0`

> How are you running stylelint: CLI?

 "CLI with `stylelint **/*.{html,jsx} "--fix"`"

> Does the bug relate to non-standard syntax?

Yes, it's related to JSX nested properties.

> What did you expect to happen?

max-width in media query string not to be transformed since it can't be camelCase in order to work for css-in-js with emotion.

> What actually happened (e.g. what warnings or errors did you get)?

No warnings flagged but the media query no longer works

evilebottnawi:
Please use issue template otherwise issue will be closed

UltraTempest:
Done

jeddy3:
@UltraTempest Thanks for the report and for using the template.

To clarify, which command are you running?

- `stylelint --syntax sass "test.jsx" --fix`
- or `stylelint **/*.{html,jsx} --fix`

The first command forces the syntax to be SASS, which won't be able to parse a `.jsx` file.

If it's the second command, can you share a typical snippet of the styling code within one of your `.html` files, please? This will help us understand your use case. We can then advise you on what action to take.

UltraTempest:
Hi @jeddy3 ,

```
[...]

PR #4573, opened 2020-01-29
Disable fixing files if fixing can cause bugs

bartlangelaan:

Currently, it's related to the pinned issues #4119 and #2643 - but more issues can be added with the same logic.

Yes.

Stylelint is an awesome tool, and I want to configure it for all projects we use at work. Because of the fixer, this would create almost no extra work for the team, because it can be run before somebody makes a commit.

There are however a few bugs open that make the autofixer behave unpredicted. I think the most known ones are #4119 and #2643. Looking at the comments in these issues, it looks like finding a solution for these issues is not easy.

I propose disabling the fixer when it detects that one of these issues may change the code in an unpredicted way. This way, the fixer doesn't always work, but it also never changes the code in an unpredicted way.

I added some code that tries to detect these bugs, and then disables the autofixer. I am unfamiliar with the code base of Stylelint, so bear with me 😄

There are probably better ways of coding this - if someone can can give me clues on what the 'right way' is, I'm happy to change my MR.


jeddy3:
@bartlangelaan Thank you! I like this proposal a lot. As you said, it makes stylelint predictable and useable while we work to address the underlying issues, which are non-trivial.

We are aware that #4119 and #2643 are pressing issues, and we've started to drum up support for fixing them. So far, we've:

- relabeled them as ["high priority"](https://github.com/stylelint/stylelint/issues?q=is%3Aopen+is%3Aissue+label%3A%22priority%3A+high%22) and pinned them
- created [a pull request](https://github.com/stylelint/stylelint/pull/4561) to changes the docs to [draw more attention to the syntax parsers](https://github.com/stylelint/stylelint/blob/0bf44f76de65563fbd19f4f31c766fe07e909a5d/docs/about/syntaxes.md)

This week I plan to:

- fork the unmaintained parsers into the stylelint organisation
- create an umbrella issue for those issues associated with this parser

(See https://github.com/stylelint/stylelint/issues/4560#issuecomment-578907499)

I hope that the umbrella issue will become something the community rallies around as we need help supporting the unmaintained CSS-in-JS parser.

Neither fixing parser issues nor the disable comments one will be an easy task. For example, 
the proposal in https://github.com/stylelint/stylelint/issues/2643#issuecomment-535406358 to fix disable comments sounds good to me, but I suspect it'll be a breaking change that will require the adapting of all 81 autofixable rules. We should do this, but it may take some time. That is why I like your pull request as it improves stylelint in the interim. 

It would be great to see this proposal grow to address issues like https://github.com/stylelint/stylelint/issues/4560, but as you said #4119 and #2643 seem to be most common. What does everyone think about trialling this proposal as-is in the next patch release and then incorporating more issues afterwards?

lib/lintSource.js
```diff
@@ -209,3 +215,54 @@ function createEmptyPostcssResult(filePath) {
...
+	// Check for issue #4119
+	if (root && root.source && root.source.lang === 'jsx' && root.nodes) {
+		const startEnds = root.nodes.map((n) =>
+			!n.source || !n.source.start || !n.source.input.css
+				? null
+				: {
+						startLine: n.source.start.line,
+						endLine: n.source.start.line + n.source.input.css.split('\n').length,
+				  },
+		);
```
hudochenkov: What do you think if we replace it with:

```js
		const startEnds = root.nodes.map((n) => {
			if (n.source && n.source.start && n.source.input.css !== undefined) {
				return {
					startLine: n.source.start.line,
					endLine: n.source.start.line + n.source.input.css.split('\n').length,
				};
			}

			return null;
		});
```

I think it's more understandable.

jeddy3:
Once this goes in, we'll need to create a follow-up pull request to https://github.com/stylelint/stylelint/pull/4561 that document this approach within the:

- [about syntaxes section](https://github.com/stylelint/stylelint/blob/855fa8a4b92e16e42d7a896b28e84537d9d87ff3/docs/about/syntaxes.md)
- [usage options section](https://github.com/stylelint/stylelint/blob/855fa8a4b92e16e42d7a896b28e84537d9d87ff3/docs/user-guide/usage/options.md#fix)

The former section also seems like the best place to list the _syntax_ patterns that cause the autofixer to ignore file by this workaround.

I can make the follow-up pull request once we merge this and https://github.com/stylelint/stylelint/pull/4561.

---

I also suggest we don't close any issues. We should add a comment to an issue when we add a workaround and state that the underlying problem still needs fixing instead.

bartlangelaan:
Hey @hudochenkov, I've rewritten the code that was unclear. I hope it's clearer now, please let me know if it's not.

lib/lintSource.js
bartlangelaan: Done! I'm now pushing the objects into an array, so the filters to check for null values are not needed anymore.

lib/lintSource.js
```diff
@@ -209,3 +215,54 @@ function createEmptyPostcssResult(filePath) {
...
+									n1.endLine < n2.startLine ||
+									// [b] [a]
+									n2.endLine < n1.startLine
+								)
+							),
+					),
+			)
+		) {
+			return false;
+		}
```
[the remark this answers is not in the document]
bartlangelaan: Rewritten this using for loops, and added comments so it's clear what every check is doing.

hudochenkov:
> I also suggest we don't close any issues. We should add a comment to an issue when we add a workaround and state that the underlying problem still needs fixing instead.

Agree.

jeddy3:
Changelog message added:

> -   Fixed: autofix will ignore sources containing disable comments or nested tagged template literals - this is workaround to make autofix safer to use until we can resolve the [underlying](https://github.com/stylelint/stylelint/issues/4119) [issues](https://github.com/stylelint/stylelint/issues/2643) ([#4573](https://github.com/stylelint/stylelint/pull/4573)).

jeddy3:
Released in `13.2.0`.

PR #4576, opened 2020-01-31
Fix wrong openIndex calculation in string-quotes (#4395)

### LINKED ISSUE #4395: Fix autofix combination for selector-attribute-brackets-space-inside and string-quotes
> Clearly describe the bug

The combination of the two rules completely destroys the attribute selector :scream: 
```js
"selector-attribute-brackets-space-inside": "always",
"string-quotes": "single",
```

`selector-attribute-brackets-space-inside`
`string-quotes`

```css
[aria-expanded="true"] {
  display: none;
}
```

```js
module.exports = {
  rules: {
    "selector-attribute-brackets-space-inside": "always",
    "string-quotes": "single",
  },
}
```

`11.1.1`

`npx stylelint style.css --config .stylelintrc.js --fix`

"No"

```css
[ aria-expanded='true' ] {
  display: none;
}
```

```css
[ aria-expanded'"tru'" ] {
  display: none;
}
```

PS. I think the problem is in this `string-quotes` rule. Because it can' t fix this `[ aria-expanded="true" ]` selector, but it breaks it.

fanich37:

It should close #4395 .

### DIFF
--- a/lib/rules/string-quotes/__tests__/index.js
+++ b/lib/rules/string-quotes/__tests__/index.js
@@ -79,6 +79,13 @@ testRule(rule, {
 			column: 6,
 		},
 		{
+			code: 'a[ id="foo" ] {}',
+			fixed: "a[ id='foo' ] {}",
+			message: messages.expected('single'),
+			line: 1,
+			column: 7,
+		},
+		{
 			code: 'a\n{ background: url("foo"); }',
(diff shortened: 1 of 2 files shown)


jeddy3:
> -   Fixed: `string-quotes` attribute selector autofix ([#4576](https://github.com/stylelint/stylelint/pull/4576)).

PR #4587, opened 2020-02-09
Format markdown with prettier

### LINKED ISSUE #4539: Use prettier to format docs
We should use prettier to format the markdown documentation for consistency. (See https://github.com/stylelint/stylelint/pull/4435). At the moment I'm using "Save without formatting" whenever I work on markdown files in stylelint.

However, this will require us to add [`/* prettier-ignore */` comments](https://prettier.io/docs/en/ignore.html#css) inside our CSS examples which show idiosyncratic patterns. This should only be for the whitespace rules.

We'll also probably need to strip out these comments as part of the website transformation script.

Any objections to this proposal?

jeddy3:

Closes #4539 

I'm glad I did this piece of work as it highlighted a lot of inconsistencies. Also, having prettier format the markdown on saving, including consistent headings, empty lines, lists and JavaScript and JSON within code blocks, is a delight. As there's more work to do on the documentation, improving the author experience around this is a plus for me!

I combed through the documentation diffs to check everything looks OK. I corrected numerous markdown fences and also a few syntax errors in the examples so they could be formatted by prettier. I think having valid JavaScript and JSON will help users who copy and paste the snippets of code.

I also removed (or replaced with a standard CSS example) the handful of Less and Scss code fences within the rule READMEs. I added a note to the working on rules document about only using standard CSS in the examples.

I also added a note to the document about adding `<!-- prettier-ignore -->` before `css` fences.


CONTRIBUTING.md
```diff
@@ -23,13 +23,14 @@ You should first [create a new issue](https://github.com/stylelint/stylelint/iss
...
--   the latest [npm](https://www.npmjs.com/)
+- a minimum of [Node.js](https://nodejs.org/en/) v10, though we do recommend using the latest LTS release
+- the latest [npm](https://www.npmjs.com/)
 
 Then:
 
-1.  [Fork and clone](https://guides.github.com/activities/forking/) the stylelint repository.
-2.  Install all the dependencies with `npm ci`.
+1. [Fork and clone](https://guides.github.com/activities/forking/) the stylelint repository.
+
```
hudochenkov: Looks like empty line is not needed here.

package.json
```diff
@@ -241,13 +240,7 @@
   "remarkConfig": {
     "plugins": [
       "preset-lint-recommended",
-      "preset-lint-consistent",
-      [
-        "lint-no-missing-blank-lines",
```
hudochenkov: Was this plugin causing issues? Prettier never adds empty lines. And we might want them after headings, for example.

CONTRIBUTING.md
jeddy3: Good catch. I've added [remark-lint-list-item-spacing](https://github.com/remarkjs/remark-lint/tree/master/packages/remark-lint-list-item-spacing), which will catch these for us and also enforce a loose style for nested lists which are easier to read.

package.json
jeddy3: > Was this plugin causing issues? Prettier never adds empty lines.

The plugin was redundant as, unlike with JavaScript, Prettier adds empty lines in markdown.

For example:

```md
title
---
1) list
```js
var = x;
\```
Paragraph
```

Becomes:

```md
# title

1. list

```js
var = x;
\```

Paragraph
```

 It's lovely to work with.

hudochenkov: Nice!

jeddy3:
@hudochenkov Thanks for the review. I've made the changes.

> We would need a remark plugin, which will ensure `prettier-ignore` comment before `css` blocks.

I agree. In fact, I started playing around with one. I think we can create a follow-up issue as the plugin will only remove a bit of manual due diligence when a contributor comes to write the README for a new rule, which doesn't happen very often.

hudochenkov:
I think you forgot to push changes :)

jeddy3:
Oops :)

hudochenkov:
Feels like new plugin for empty lines makes things less consistent. Sometimes it adds empty lines, sometime — not. Hard to follow. Also compare difference in rendering [with empty lines](https://github.com/stylelint/stylelint/blob/59ee4124ba3b42e30560f073d79e747a44d49092/docs/toc.md) and [without them](https://github.com/stylelint/stylelint/blob/53bc10e09979b6f491f152c8392381386d8e1f0e/docs/toc.md). It might also introduce weird rendering on our website. I propose revert this plugin and keep no empty lines between list items.

jeddy3:
> Feels like new plugin for empty lines makes things less consistent.

Yes, I think you're right. It _is_ consistent in that it's loose for multi-line list items and tight for single-line list items, but it _feels_ inconsistent. I've reverted the change.

hudochenkov:
BTW, could you also add prettier to lint-staged for markdown, please?

jeddy3:
> BTW, could you also add prettier to lint-staged for markdown, please?

Good catch. Done.

jeddy3:
I think something went iffy with the revert as the CI is complaining that prettier needs to run on `CHANGELOG.md` but it's correctly formatted at my end. Any ideas?

jeddy3:
Closing in favour of https://github.com/stylelint/stylelint/pull/4588.

PR #4594, opened 2020-02-12
Add --stdin flag

### LINKED ISSUE #4570: Add --stdin flag
> Clearly describe the bug

If `a.css` is an empty file then `cat a.css | stylelint` outputs help instead of an error.

`no-empty-source`

An empty file (e.g. `touch a.css`)

```json
{
  "rules": {
    "no-empty-source": true
  }
}
```

13.0.0

CLI

No

Same output as `stylelint a.css`

```
a.css
 1:1  ✖  Unexpected empty source   no-empty-source
```

Same output as `stylelint --help`

```

  A mighty, modern CSS linter.

  Usage: stylelint [input] [options]

  Input: Files(s), glob(s), or nothing to use stdin.

    If an input argument is wrapped in quotation marks, it will be passed to
    globby for cross-platform glob support. node_modules are always ignored.
    You can also pass no input and use stdin, instead.

  Options:

...
```

Possibly related to #1557

alypeng:
Closes #4570

### DIFF
--- a/docs/user-guide/usage/cli.md
+++ b/docs/user-guide/usage/cli.md
@@ -36,6 +36,10 @@ Print the configuration for the given path. stylelint outputs the configuration 
 
 Only register violations for rules with an "error"-level severity (ignore "warning"-level).
 
+### `--stdin`
+
+Accept stdin input even if it is empty.
+
 ### `--version, -v`
 
 Show the currently installed version of stylelint.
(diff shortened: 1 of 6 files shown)


jeddy3:
Changelog entry:

> - Added: `--stdin` CLI flag that accepts stdin input even if it is empty ([#4594](https://github.com/stylelint/stylelint/pull/4594)).

PR #4611, opened 2020-02-20
Fix false positives for longhand grid-column/row-* properties in value-keyword-case

### LINKED ISSUE #4605: Fix false positives for longhand grid-column-* properties in value-keyword-case
Hello!

`value-keyword-case` flags the following instance:

```
grid-column-end: wideBlockRow;
```

However, the rule seems to be primarily designed to impose consistent use of CSS keywords.

Because camel casing seems desirable for (or at least improves the readability of) `longUserDefinedVariables`, it would be neat to disable this feature for such keywords.

Thank you!

fanich37:

It should close #4605 .

I think its necessary to convert prop to lower case before any comparison can be done.

### DIFF
--- a/lib/rules/value-keyword-case/__tests__/index.js
+++ b/lib/rules/value-keyword-case/__tests__/index.js
@@ -669,6 +669,13 @@ testRule(rule, {
 			line: 1,
 			column: 24,
 		},
+		{
+			code: 'a { GRID-COLUMN-START: SPAN camelCaseArea; }',
+			fixed: 'a { GRID-COLUMN-START: span camelCaseArea; }',
+			message: messages.expected('SPAN', 'span'),
+			line: 1,
+			column: 24,
+		},
 	],
 });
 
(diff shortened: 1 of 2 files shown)


lib/rules/value-keyword-case/__tests__/index.js
```diff
@@ -311,6 +311,10 @@ testRule(rule, {
 			code: 'a { color: ThreeDShadow; }',
 			description: 'another system color',
 		},
+		{
+			code: 'a { grid-column-start: span camelCaseArea; }',
```
evilebottnawi: Can we add a test for `a { GRID-COLUMN-START: SPAN camelCaseArea; }`?

jeddy3:
> - Fixed: `value-keyword-case` false positives for longhand `grid-column/row-*` properties ([#4611](https://github.com/stylelint/stylelint/pull/4611)).

camiblanch:
This change caused an issue with having uppercase in css variable properties that you wish to add to the `ignoreProperties` list. See #4644

PR #4633, opened 2020-03-03
Add `node` property to lint result

galaxycoder:


jeddy3:
@galaxycoder Thanks for the pull request. Can you tell us more about the use case behind the change?

PR #4640, opened 2020-03-05
Add ignoreSelectors option to block-opening-brace-space-before

### LINKED ISSUE #4624: Add ignoreSelectors: [] to block-opening-brace-space-before

I'm running into CSS where developers use more than a single space before an opening brace in some situations, such as the following keyframe example with `0%`, `50%`, and `100%`, to make the code easier to read:

```css
@keyframes example {
	0%   { transform: none; }
	50%  { transform: rotate(45deg); }
	100% { transform: rotate(90deg); }
}
```

An `ignoreSelectors` option (similar to how other rules have this secondary option) for the `block-opening-brace-space-before` rule, which would let me exclude selectors that contain `%`, for example.

Note that the rule currently has an `ignoreAtRules` option, which is only concerned about the opening brace of the block, not of any nested rules inside of them.

charlessuh:

Fixes #4624

No, it's self-explanatory.


lib/rules/block-opening-brace-space-before/__tests__/index.js
```diff
@@ -80,89 +80,103 @@ testRule(rule, {
 	],
 });
 
-testRule(rule, {
-	ruleName,
-	config: ['always', { ignoreAtRules: ['for'] }],
-	fix: true,
+for (const selector of ['for', '/fo/', /fo/]) {
```
jeddy3: Thanks for cleaning up the tests!

I suggest we remove the `for` loop and place the array within the config so that it is idomatic with our other test files, like so:

```
config: ['always', { ignoreAtRules: ['for', '/fo/', /fo/] }],
```

lib/rules/block-opening-brace-space-before/__tests__/index.js
```diff
@@ -80,89 +80,103 @@ testRule(rule, {
...
+			},
+		],
+	});
+}
 
-testRule(rule, {
-	ruleName,
-	config: ['always', { ignoreAtRules: '/fo/' }],
-	fix: true,
+for (const selector of ['a', '/a/', /a/]) {
```
jeddy3: Same here, please. Let's add the array to the config instead of using a loop.

lib/rules/block-opening-brace-space-before/__tests__/index.js
```diff
@@ -80,89 +80,103 @@ testRule(rule, {
...
-			code: 'a{ color: pink; }',
-			fixed: 'a { color: pink; }',
-			message: messages.expectedBefore(),
-			line: 1,
-			column: 1,
-		},
-	],
-});
+		reject: [
+			{
```
jeddy3: I think a single reject test will suffice to show that the option is working as the other parts, e.g. the `optionsMatches` function, are tested elsewhere. We can use the first test, using `b{`.

m-allanson:
Updated changelog:

> - Added: `ignoreSelectors[]` to `block-opening-brace-space-before` ([#4640](https://github.com/stylelint/stylelint/pull/4640))

PR #4654, opened 2020-03-19
Fix false positives for x unit within vendor-prefixed image-set in unit-no-unknown

### LINKED ISSUE #4653: Fix false positives for vendor-prefixed image-set w/ unit "x" in unit-no-unknown
> Clearly describe the bug

The unit-no-unknown rule reports a false positive when using the "x" unit in a value for a vendor-prefixed image-set ~~property~~ value. E.g.

```css
a { background-image: -webkit-image-set('img-1x.jpg' 1x, 'img-2x.jpg' 2x) }
```

caniuse mentions that `-webkit-image-set` is a valid vendor-prefixed version of `image-set`: https://caniuse.com/#feat=css-image-set

(slight change to "ignore `x` unit in image-set" test case in [lib/rules/unit-no-unknown/__tests__/index.js](https://github.com/stylelint/stylelint/blob/master/lib/rules/unit-no-unknown/__tests__/index.js))

`unit-no-unknown`

```css
a { background-image: -webkit-image-set('img-1x.jpg' 1x, 'img-2x.jpg' 2x) }
```

```json
{
  "rules": {
    "unit-no-unknown": "always"
  }
}
```

13.2.1

CLI; `npm run test`

No

No warnings to be flagged.

```json
{
  "column": 54,
  "line": 1,
  "rule": "unit-no-unknown",
  "severity": "error",
  "text": "Unexpected unknown unit \"x\" (unit-no-unknown)",
}
```

I can write and send a PR.

srawlins:

Fixes #4653

No, it's self-explanatory.

However, I'm not super familiar with the ASTs involved, and I'm super open to suggestions if this is not an optimal fix.

### DIFF
--- a/lib/rules/unit-no-unknown/index.js
+++ b/lib/rules/unit-no-unknown/index.js
@@ -8,6 +8,7 @@ const isMap = require('../../utils/isMap');
 const keywordSets = require('../../reference/keywordSets');
 const mediaParser = require('postcss-media-query-parser').default;
 const optionsMatches = require('../../utils/optionsMatches');
+const postcss = require('postcss');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
 const validateOptions = require('../../utils/validateOptions');
@@ -116,8 +117,10 @@ function rule(actual, options) {
 							return;
 						}
 
-						if (/^image-set/i.test(value)) {
-							const imageSet = parsedValue.nodes.find((node) => node.value === 'image-set');
+						if (/image-set/i.test(value)) {
+							const imageSet = parsedValue.nodes.find(
+								(node) => postcss.vendor.unprefixed(node.value) === 'image-set',
+							);
 							const imageSetValueLastIndex = _.last(imageSet.nodes).sourceIndex;
 
 							if (imageSetValueLastIndex >= valueNode.sourceIndex) {
--- a/lib/rules/unit-no-unknown/index.js
+++ b/lib/rules/unit-no-unknown/index.js
@@ -117,7 +117,7 @@ function rule(actual, options) {
 							return;
 						}
 
-						if (/image-set/i.test(value)) {
+						if (/^(?:-webkit-)?image-set[\s(]/i.test(value)) {
 							const imageSet = parsedValue.nodes.find(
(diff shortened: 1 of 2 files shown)


lib/rules/unit-no-unknown/index.js
```diff
@@ -116,8 +117,10 @@ function rule(actual, options) {
 							return;
 						}
 
-						if (/^image-set/i.test(value)) {
-							const imageSet = parsedValue.nodes.find((node) => node.value === 'image-set');
+						if (/image-set/i.test(value)) {
```
evilebottnawi: I think better `/^(?:-webkit-)?image-set$/i`, because in theory we can have `image-set-something()` function (yes, it is not exists in CSS), but strict check always better

jeddy3: Yes, let's go with tweaking the RegEx. 

I think we'll be fine with `/^(?:-webkit-)?image-set$/i` as [only the `-webkit` prefix is currently used](https://developer.mozilla.org/en-US/docs/Web/CSS/image-set#Browser_compatibility).

srawlins: Adding `$` at the end does not work here because of what we're matching on. I was able to use `[\s(]` to delimit the end of the function name.

lib/rules/unit-no-unknown/index.js
```diff
@@ -116,8 +117,10 @@ function rule(actual, options) {
 							return;
 						}
 
-						if (/^image-set/i.test(value)) {
-							const imageSet = parsedValue.nodes.find((node) => node.value === 'image-set');
+						if (/^(?:-webkit-)?image-set[\s(]/i.test(value)) {
```
ybiquitous, line 120: [IMO] If using [`\b` (word boundary)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Assertions#Types), the intent of the code may be clear.

For example:

```suggestion
						if (/^(?:-webkit-)?image-set\b/i.test(value)) {
```

jeddy3:
Changelog entry:

> - Fixed: `unit-no-unknown` false positives for `x` unit within vendor-prefixed `image-set` ([#4654](https://github.com/stylelint/stylelint/pull/4654)).

PR #4656, opened 2020-03-21
Add ignoreFontFamilies option to `font-family-no-missing-generic-family-keyword`

### LINKED ISSUE #3288: Add ignoreFontFamilies: [] to font-family-no-missing-generic-family-keyword

The `stylelint-config-recommended` shared config includes the `font-family-no-missing-generic-family-keyword` rule ([src](https://github.com/stylelint/stylelint-config-recommended/blob/master/index.js#L17))

The issue I have is that I'm using an [icon font](https://css-tricks.com/examples/IconFont/) and as there is (typically) no _fallback_ for icon fonts so `font-family: icons;` will trigger the `font-family-no-missing-generic-family-keyword` rule.

```css
[data-icon]:before {
  font-family: icons; /* BYO icon font, mapped smartly */
  content: attr(data-icon);
  speak: none; /* Not to be trusted, but hey. */
}
```

Use or extend the `stylelint-config-recommended` shared config

`9.2.0`

n/a

No

Warnings happen as expected 

n/a

I can envisage a couple of options to resolve this:

1. Remove the `font-family-no-missing-generic-family-keyword` rule from `stylelint-config-recommended`

2. Override the `font-family-no-missing-generic-family-keyword` rule from my config

3. Add an `ignore` option to the `font-family-no-missing-generic-family-keyword` rule where a font name can be declared e.g. `icons` and the `font-family-no-missing-generic-family-keyword` rule would then ignore these instances.

I've a preference for option 3 which I'm thinking would be the best option, thoughts @stylelint/core ?

mattfelten:

Closes #3288

Seems to work alright so far. Something I'm struggling with is how a font family name in quotes should be written. I couldn't tell if that works and I'm just writing the test poorly or this doesn't work with quoted names.

### DIFF
--- a/lib/rules/font-family-no-missing-generic-family-keyword/README.md
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/README.md
@@ -56,12 +56,12 @@ a { font: caption; }
 
 ## Optional secondary options
 
-### `ignoreFontFamilies: : ["/regex/", /regex/, "string"]`
+### `ignoreFontFamilies: ["/regex/", /regex/, "string"]`
 
 Given:
 
 ```
-["custom-icon-font"]
+["custom-font"]
 ```
 
 The following pattern is _not_ considered a violation:
@@ -69,4 +69,11 @@ The following pattern is _not_ considered a violation:
 <!-- prettier-ignore -->
 ```css
 a { font-family: custom-font; }
+```
+
+The following pattern is considered a violation:
+
+<!-- prettier-ignore -->
+```css
+a { font-family: invalid-custom-font; }
 ```
--- a/lib/rules/font-family-no-missing-generic-family-keyword/README.md
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/README.md
@@ -64,7 +64,7 @@ Given:
 ["custom-icon-font"]
 ```
 
-The following pattern is not considered a violation:
+The following pattern is _not_ considered a violation:
 
 <!-- prettier-ignore -->
 ```css
--- a/lib/rules/font-family-no-missing-generic-family-keyword/README.md
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/README.md
@@ -56,9 +56,15 @@ a { font: caption; }
 
 ## Optional secondary options
 
-### `ignoreValues: ["custom-font"]`
+### `ignoreFontFamilies: : ["/regex/", /regex/, "string"]`
 
-Ignore this rule for specific font names that would not have an appropriate generic font family to fall back to. Commonly, this would be used for icon fonts.
+Given:
+
+```
+["custom-icon-font"]
+```
+
+The following pattern is not considered a violation:
 
 <!-- prettier-ignore -->
 ```css
--- a/lib/rules/font-family-no-missing-generic-family-keyword/README.md
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/README.md
@@ -53,3 +53,14 @@ a { font: inherit; }
 ```css
 a { font: caption; }
 ```
+
+## Optional secondary options
+
+### `ignoreValues: ["custom-font"]`
+
+Ignore this rule for specific font names that would not have an appropriate generic font family to fall back to. Commonly, this would be used for icon fonts.
+
+<!-- prettier-ignore -->
+```css
+a { font-family: custom-font; }
+```
--- a/lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.js
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.js
@@ -135,7 +135,7 @@ testRule(rule, {
 	config: [
 		true,
 		{
-			ignoreValues: ['custom-font'],
+			ignoreFontFamilies: ['custom-font'],
 		},
 	],
 	accept: [
@@ -144,20 +144,11 @@ testRule(rule, {
 		},
 		{
 			code: 'a { font: 1em Lucida Grande, Arial, custom-font }',
-			message: messages.rejected,
-			line: 1,
-			column: 37,
 		},
 	],
 	reject: [
 		{
 			code: 'a { font-family: Arial; }',
-			message: messages.rejected,
-			line: 1,
-			column: 18,
-		},
-		{
-			code: 'a { font-family: Times; }',
 			message: messages.rejected,
 			line: 1,
 			column: 18,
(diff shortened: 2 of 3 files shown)


lib/rules/font-family-no-missing-generic-family-keyword/index.js
```diff
@@ -16,7 +16,7 @@ const messages = ruleMessages(ruleName, {
 const isFamilyNameKeyword = (node) =>
 	!node.quote && keywordSets.fontFamilyKeywords.has(node.value.toLowerCase());
 
-function rule(actual) {
+function rule(actual, options = {}) {
```
jeddy3: Instead of using a default parameter, we should update the parameters of the `validateOptions` function below:

```js
const validOptions = validateOptions(
			result,
			ruleName,
			{ actual },
			{
				actual: options,
				possible: {
					ignoreFontFamilies: [_.isString, _.isRegExp]
				},
				optional: true,
			},
		);
```

lib/rules/font-family-no-missing-generic-family-keyword/index.js
```diff
@@ -48,6 +48,15 @@ function rule(actual) {
 				return;
 			}
 
+			if (
```
jeddy3: We'll want to use the `optionsMatches` util here, [like so](https://github.com/stylelint/stylelint/blob/8c25249173dedd93f72c8b4d966abf67cb54f1ad/lib/rules/property-no-unknown/index.js#L64). So that we support regex as well as strings.

mattfelten: Cool, that makes sense. Simplifies my work a bit

mattfelten:
@jeddy3 Thanks for the review. Just made the requested changes. Let me know if I can make this better any more.

lib/rules/font-family-no-missing-generic-family-keyword/index.js
```diff
@@ -48,6 +62,14 @@ function rule(actual) {
...
+				fontFamilies.some((node) =>
+					optionsMatches(options, 'ignoreFontFamilies', node.value.toLowerCase()),
```
jeddy3:
```suggestion
					optionsMatches(options, 'ignoreFontFamilies', node.value),
```

We'll want to avoid using `toLowerCase` here as the user can use the [`i` flag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/ignoreCase) in their regex if needed. (This differs from line 20, where we use `toLowerCase` because we are comparing against a predefined list of reference keywords there rather than a user-defined regex.)

mattfelten:
@jeddy3 Thanks! Got those taken care of too

lib/rules/font-family-no-missing-generic-family-keyword/README.md
```diff
@@ -53,3 +53,20 @@ a { font: inherit; }
 ```css
 a { font: caption; }
 ```
+
+
+### `ignoreFontFamilies: : ["/regex/", /regex/, "string"]`
```
ybiquitous, line 59: [suggestion] Fix typo:

```suggestion
### `ignoreFontFamilies: ["/regex/", /regex/, "string"]`
```

lib/rules/font-family-no-missing-generic-family-keyword/README.md
```diff
@@ -53,3 +53,20 @@ a { font: inherit; }
...
+
+Given:
+
+```
+["custom-icon-font"]
```
ybiquitous, line 64: [suggestion] Fix typo:

```suggestion
["custom-font"]
```

lib/rules/font-family-no-missing-generic-family-keyword/README.md
```diff
@@ -53,3 +53,20 @@ a { font: inherit; }
...
+```
+
+The following pattern is _not_ considered a violation:
+
+<!-- prettier-ignore -->
+```css
+a { font-family: custom-font; }
+```
```
ybiquitous, line 72: [suggestion] I feel more understandable if a violation example is added, for example:

---

Given:

```
["custom-font"]
```

The following pattern is _not_ considered a violation:

```css
a { font-family: custom-font; }
```

The following pattern is considered a violation:

```css
a { font-family: invalid-custom-font; }
```

lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.js
```diff
@@ -130,3 +129,35 @@ testRule(rule, {
 		},
 	],
 });
+
+testRule(rule, {
+	ruleName,
+	config: [
+		true,
+		{
+			ignoreFontFamilies: ['custom-font'],
```
ybiquitous, line 138:
```suggestion
			ignoreFontFamilies: ['custom-font', /foo/, '/baz[0-9]/'],
```

lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.js
```diff
@@ -130,3 +129,35 @@ testRule(rule, {
...
+		},
+	],
+	accept: [
+		{
+			code: 'a { font-family: custom-font; }',
+		},
+		{
+			code: 'a { font: 1em Lucida Grande, Arial, custom-font }',
+		},
```
ybiquitous, line 147:
```suggestion
		},
		{
			code: 'a { font: foo-font }',
		},
		{
			code: 'a { font: baz7 }',
		},
```

lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.js
```diff
@@ -130,3 +129,35 @@ testRule(rule, {
...
+			message: messages.rejected,
+			line: 1,
+			column: 18,
+		},
+		{
+			code: 'a { font: 1em Lucida Grande, Arial, "sans-serif" }',
+			message: messages.rejected,
+			line: 1,
+			column: 37,
+		},
```
ybiquitous, line 161:
```suggestion
		},
		{
			code: 'a { font: bar-font }',
			message: messages.rejected,
			line: 1,
			column: 11,
		},
		{
			code: 'a { font: baz-7 }',
			message: messages.rejected,
			line: 1,
			column: 11,
		},
```

mattfelten:
Thanks @jeddy3 !

@ybiquitous I got your changes in. Thanks for looking! Lmk if I can improve anything else.

jeddy3:
Changelog:

> - Added: `ignoreFontFamilies: []` to `font-family-no-missing-generic-family-keyword` ([#4656](https://github.com/stylelint/stylelint/pull/4656)).

PR #4659, opened 2020-03-22
Ignore non-standard syntax in function-calc-no-invalid

### LINKED ISSUE #4114: Fix false positives for variables in function-calc-no-invalid
There are false positives for less and scss variables containing `-`

`function-calc-no-invalid`

for less

```css
.foo {
	height: calc(100% - (@a-height + @b-margin-top));
}
```
for scss
```css
.foo {
	height: calc(100% - ($a-height + $b-margin-top));
}
```

```json
{
  "rules": {
  }
}
```

`10.1.0`

CLI with `stylelint file.less --config myconfig.json`

Yes, it's related to less and scss variables

No warnings to be flagged

The following warnings were flagged (for less case), a pair of issues per dash in a variable name

```shell
 2:25  ✖  Expected space before "-" operator   function-calc-no-invalid
 2:26  ✖  Expected space after "-" operator    function-calc-no-invalid
 2:37  ✖  Expected space before "-" operator   function-calc-no-invalid
 2:38  ✖  Expected space after "-" operator    function-calc-no-invalid
 2:44  ✖  Expected space before "-" operator   function-calc-no-invalid
 2:45  ✖  Expected space after "-" operator    function-calc-no-invalid
```

There was a bug fix done (#3908) to fix variables inside interpolation for both less and scss, but for simple variable there are still FPs. Note that for scss as soon as I wrap variables inside interpolation, issues disappear :
```
.foo {
	height: calc(100% - (#{$a-height} + #{$b-margin-top}));
}
```

hudochenkov:

Fixes #4114, #4533.
Closes #4534.

@jeddy3 [said](https://github.com/stylelint/stylelint/issues/4114#issuecomment-573375622):

> Built-in rules ignore non-standard syntax. This rule should only check the validity of `calc` functions containing standard CSS syntax.

To be honest, I don't think this is a good approach for users. `function-calc-no-invalid` catches many cases of invalid `calc` for preprocessor users. But with this PR, we reduce usefulness of this rule for these users.

### DIFF
--- a/lib/utils/isStandardSyntaxMathFunction.js
+++ b/lib/utils/isStandardSyntaxMathFunction.js
@@ -1,0 +1,21 @@
+'use strict';
+
+/**
+ * Check whether a math function is standard
+ *
+ * @param {string} mathFunction
+ * @returns {boolean}
+ */
+module.exports = function isStandardSyntaxMathFunction(mathFunction) {
+	// SCSS variable
+	if (mathFunction.includes('$')) {
+		return false;
+	}
+
+	// Less variable
+	if (mathFunction.includes('@')) {
+		return false;
+	}
+
+	return true;
+};
--- a/lib/rules/function-calc-no-invalid/__tests__/index.js
+++ b/lib/rules/function-calc-no-invalid/__tests__/index.js
@@ -285,70 +285,52 @@ testRule(rule, {
 		},
 		{
 			code: `
-      a {
-        width: calc(
-          #{$min-value} +
-          #{strip-unit($max-value - $min-value)} *
-          (100vw - #{$min-vw})
-        );
-      }
-      `,
+				a {
+					width: calc(
+						#{$min-value} +
+						#{strip-unit($max-value - $min-value)} *
+						(100vw - #{$min-vw})
+					);
+				}
+			`,
 		},
 		{
 			code: `
-      a {
-        width: calc(#{$min-value} + #{$max-value}*(100vw - #{$min-vw}));
-      }
-      `,
+				a {
+					width: calc(#{$min-value} + #{$max-value}*(100vw - #{$min-vw}));
+				}
+     		 `,
+		},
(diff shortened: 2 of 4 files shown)


lib/utils/isStandardSyntaxMathFunction.js
```diff
@@ -0,0 +1,21 @@
...
+		return false;
+	}
+
+	// Less variable
+	if (mathFunction.includes('@')) {
+		return false;
+	}
+
+	return true;
+};
```
evilebottnawi: What about interpolation SCSS `#{var}`?

hudochenkov: Isn't it `#{$var}`?

evilebottnawi: Oh, yes, what about `$` and `@` in css variable name, can these characters used?

hudochenkov: I couldn't get straight answer from the spec. But I tried in Chrome putting these symbols in different position in the custom property name, and all were invalid.

evilebottnawi: Sounds good, approved

hudochenkov:
> The community can create a scss/function-calc-no-invalid plugin (and a less/function-calc-no-invalid one etc). Each would cater to the more specific syntax of their respective languages.

True. Also, there are pieces of code after that PR, that handle most of use cases of variables and interpolations. I don't know where they are, but now it's a dead code. And this is the case most likely with every rule, where we started ignoring non-standard syntax.

jeddy3:
> - Fixed: `function-calc-no-invalid` false positives for SCSS and Less variables ([#4659](https://github.com/stylelint/stylelint/pull/4659)).

jonscottclark:
Thanks everyone involved for your contributions on this :)

PR #4682, opened 2020-04-02
Add a release action

Berkmann18:
With the relevant `semantic-release` configurations

None, since it's related to the following discussion:
https://github.com/orgs/stylelint/teams/core/discussions/19 (**tl;dr:** automating the release process).

It makes uses of **two GH secrets being `GITHUB_TOKEN` and `NPM_TOKEN`** (which can be set on the [Settings/Secrets section](https://github.com/stylelint/stylelint/settings/secrets)).
You can read more on the action itself on https://github.com/semantic-release/semantic-release/blob/master/docs/recipes/github-actions.md

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,25 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - name: Install dependencies
+        run: npm ci
+      - name: Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npx semantic-release
--- a/.releaserc
+++ b/.releaserc
@@ -1,0 +1,14 @@
+{
+  "plugins": [
+    "@semantic-release/commit-analyzer",
+    "@semantic-release/release-notes-generator",
+    ["@semantic-release/changelog", {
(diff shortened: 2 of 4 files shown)


Berkmann18:
> I'm not sure we can use semantic-release. First, we don't write commits in the format it wanted.

`semantic-release` normally picks up Angular-style commits as you probably already know, however, the `commit-analyzer` normally detects the convention used (be it ESLint, JSHint, Ember, ...) so it should be fine here (@gr2m correct me if I'm wrong).

> Second, we write our changelogs, because usually commit message is not a good changelog entry.

That makes sense and I guess getting people to write commits/PR titles in a way that they would be changelog-worthy might not be a good idea given how many people are involved and the project' size.
Unless there's a specification/doc on the commit convention used by Stylelint which could be used to make a `commitizen` transformer (which will ensure commits follow the same structure) which could then be turned into a `semantic-release` preset.

> Third, I don't think we want to release after every pull request merged.

I'm not fully sure how that would play out on projects that don't follow a (custom) commit convention, but `semantic-release` doesn't treat every commits the same way, for example a chore/dependency upgrade won't trigger a release (unless it was configured to do so) while the others will trigger one.
You can read more about that on https://github.com/semantic-release/commit-analyzer#usage

gr2m:
> the commit-analyzer normally detects the convention used (be it ESLint, JSHint, Ember, ...) so it should be fine here (@gr2m correct me if I'm wrong).

No, there is no auto-detection. The preset defaults to `angular`, but other presets are supported, and you can configure the commit conventions to your liking:
https://github.com/semantic-release/commit-analyzer#configuration

Berkmann18:
@gr2m Thanks for the correction.

Berkmann18:
@hudochenkov @jeddy3 Do you guys know if anyone tried using [RollingVersions](https://rollingversions.com), I spoke to Forbes Lindsay (the creator) and the use cases seem to go along with what Stylelint would do regarding releases.

Berkmann18:
@jeddy3 No problem!

Issue #4668, opened 2020-03-31
Document passing arguments to custom formatters


masi:
> What solution would you like to see?

Access to all or parts of the Stylelint configuration data and/or to namespaced CLI arguments.

Proposal one requires a new key "formatter" within the configuration array that may hold options for all custom formatters. The path (without .js) is the key within "formatter" for the actual custom formatter in user.

Proposal two requires the CLI to accept any arguments of the form custom-formatter-x. Where x stands for any string. All options without the prefix custom-formatter- are to be passed on the custom formatter.

Both proposals require the addition of a new argument for the formatter function.

jeddy3:
@masi Thanks for the request and for using the template.

This is the first time we've had a request for this feature. It looks like [ESLint suggests using](https://eslint.org/docs/developer-guide/working-with-custom-formatters#passing-arguments-to-formatters) either:

- environmental variables
- piping the output from the JSON formatter into another program

If neither of these are acceptable, can you tell us more about your use case so that we can decide on the best solution? Otherwise, I'll label this up as a documentation issue.

masi:
Both solutions are acceptable. It didn't occur to me to use environment variables for custom code. Building a wrapper around the JSON output is also a god idea.
Sorry.

jeddy3:
> Both solutions are acceptable. 

It's good to hear.

> Sorry.

It's all good. The solutions aren't documented anywhere in our docs. I'll reopen the issue as "type: documentation". I'll probably have time to rustle up something tomorrow, probably a line or two on the [writing formatters](https://stylelint.io/developer-guide/formatters) page in the developer guide.

PR #4688, opened 2020-04-04
Document passings args to formatters

jeddy3:

Closes https://github.com/stylelint/stylelint/issues/4668

Keeps it simple. We can always expand on it later if there's a need.

### DIFF
--- a/docs/developer-guide/formatters.md
+++ b/docs/developer-guide/formatters.md
@@ -79,6 +79,23 @@ And the second argument (`returnValue`) is an object with one or more of the fol
 }
 ```
 
+## Passing arguments
+
+To pass arguments, you can:
+
+- use environmental variables
+- pipe the output from the JSON formatter into another program
+
+For example:
+
+```console
+SKIP_WARNINGS=true stylelint "*.css" --custom-formatter ./my-formatter.js
+```
+
+```console
+stylelint -f json "*.css" | your-program-that-reads-JSON --option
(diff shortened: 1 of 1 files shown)


docs/developer-guide/formatters.md
```diff
@@ -79,6 +79,23 @@ And the second argument (`returnValue`) is an object with one or more of the fol
 }
 ```
 
+## Passing arguments
+
+To pass arguments, you can:
+
+- use environmental variables
+- pipe the output from the JSON formatter into another program
```
hudochenkov, line 87: It's not clear for me how piping output is passing arguments to formatter. Piping happens after formatter was ran.

jeddy3: You're right. It's not clear that the second approach is a separate formatting program rather than a stylelint formatter. I've updated the docs to hopefully illustrate that.

---

Example formatting program:

```
// my-program-that-reads-JSON.js
const getStdin = require("get-stdin");
(async () => {
  const jsonString = await getStdin();
  const object = JSON.parse(jsonString);
  let colors = false;
  if (process.argv[2] && process.argv[2] === "-c") colors = true;
  console.dir(object, { depth: null, colors });
})();
```

```
stylelint -f json "*.css" | my-program-that-reads-JSON --c
```

This just (optionally) colourises and indents the JSON, but a formatting program could do anything with the data.

PR #4690, opened 2020-04-05
Change dependabot schedule to `weekly`

ntwb:
Example https://github.com/stylelint/stylelint-demo/pull/217, also the VSCode extension just had 12 PRs also, there are too many landing monthly on the same day and the merge conflicts and lock files make this difficult and time consuming to manage so many, increasing the frequency should reduce this, at least worth a try

n/a

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -3,7 +3,7 @@ version: 1
 update_configs:
   - package_manager: 'javascript'
     directory: '/'
-    update_schedule: 'monthly'
+    update_schedule: 'weekly'
     version_requirement_updates: 'increase_versions'
     default_labels:
       - 'pr: dependencies'


hudochenkov:
What about not using dependabot at all? Manually update dependencies once a week/month is much faster, that dealing with PR for every dependency.

jeddy3:
> Manually update dependencies once a week/month is much faster, that dealing with PR for every dependency.

I doubt we'll remember to do, especially across all the repositories in the organisation. I think having dependobot automatically do this weekly, especially when it flags security updates [like so](https://github.com/stylelint/vscode-stylelint/pull/88), is worth trying. We can change tack if doesn't work out.

jeddy3:
It looks like [the feature is coming](https://github.com/dependabot/feedback/issues/5), but not for a while.

One workaround is limiting the allowed open PRs to 1 in the dependobot settings. We can then either:

- merge dependency updates one-at-a-time and avoid rebasing
- amend the PR manually to update all dependencies

The latter combines a manual update with an automatic reminder. This might be the sweet spot.

ntwb:
Same, also happy to try the various options above.

One think I've found very helpful is the new GitHub mobile app, simple PRs are quick and easy to merge from the couch whilst watching TV now 😬

PR #4705, opened 2020-04-15
Fix turning off autofix for sources containing scoped disable comments

### LINKED ISSUE #4704: Fix autofix running on files containing scoped disable comments
> Clearly describe the bug

`npx stylelint --fix` corrects violations of the rule `color-hex-length` even after disabling it with `// stylelint-disable color-hex-length`.

`color-hex-length`

This is part of a bigger file located (in my system) at `_sass/layout/syntax-lighlighting.scss`.

```scss
// https://github.com/richleland/pygments-css/blob/master/vs.css
// stylelint-disable color-hex-length, color-named, font-weight-notation, selector-class-pattern
.highlight .hll { background-color: #ffffcc; }
```

```json
{
  "extends": [
  ],
  "rules": {
    "string-quotes": null,
    "indentation": null,
    "value-list-comma-newline-after": null,
    "value-list-comma-space-after": null,
    "function-calc-no-unspaced-operator": null,
    "scss/operator-no-unspaced": null,
    "declaration-no-important": null,
    "declaration-colon-newline-after": null,
    "order/properties-order": null,
    "scss/at-function-named-arguments": null
  }
}
```

13.3.1

`npx stylelint _sass/**/*.scss --fix`

No.

No corrections made.

`npx stylelint _sass/**/*.scss` yields nothing, as expected. But `npx stylelint _sass/**/*.scss --fix` ignores the disabling comment and yields:
```scss
// https://github.com/richleland/pygments-css/blob/master/vs.css
// stylelint-disable color-hex-length, color-named, font-weight-notation, selector-class-pattern
.highlight .hll { background-color: #ffc; }
```

jeddy3:

Closes #4704

We were only disabling autofix on files containing unscoped disable comments. This pull request disables it for both scoped and unscoped.

### DIFF
--- a/lib/lintSource.js
+++ b/lib/lintSource.js
@@ -205,7 +205,11 @@ function lintPostcssResult(stylelint, postcssResult, config) {
 			Promise.all(
 				postcssRoots.map((postcssRoot) =>
 					ruleFunction(primaryOption, secondaryOptions, {
-						fix: stylelint._options.fix && isFileFixCompatible,
+						fix:
+							stylelint._options.fix &&
+							// Next two conditionals are temporary measures until #2643 is resolved
+							isFileFixCompatible &&
+							!postcssResult.stylelint.disabledRanges[ruleName],
 						newline,
 					})(postcssRoot, postcssResult),
 				),
@@ -257,10 +261,7 @@ function createEmptyPostcssResult(filePath) {
  */
 function isFixCompatible({ stylelint }) {
 	// Check for issue #2643
-	const hasUnscopedDisabledRanges = stylelint.disabledRanges.all.length > 0;
-	const hasScopedDisabledRanges = Object.keys(stylelint.disabledRanges).length > 1;
-
-	if (hasUnscopedDisabledRanges || hasScopedDisabledRanges) return false;
+	if (stylelint.disabledRanges.all.length) return false;
 
 	return true;
 }
--- a/lib/lintSource.js
+++ b/lib/lintSource.js
@@ -251,15 +251,16 @@ function createEmptyPostcssResult(filePath) {
  * There are currently some bugs in the autofixer of Stylelint.
  * The autofixer does not yet adhere to stylelint-disable comments, so if there are disabled
  * ranges we can not autofix this document. More info in issue #2643.
- * Also, if this document is parsed with postcss-jsx and there are nested template
- * literals, it will duplicate some code. More info in issue #4119.
  *
  * @param {PostcssResult} postcssResult
  * @returns {boolean}
  */
 function isFixCompatible({ stylelint }) {
 	// Check for issue #2643
-	if (stylelint.disabledRanges.all.length) return false;
+	const hasUnscopedDisabledRanges = stylelint.disabledRanges.all.length > 0;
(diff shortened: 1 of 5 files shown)


amyspark:
This version of stylelint works correctly with my toolchain (`npm i https://github.com/stylelint/stylelint/tarball/issue-4704`).

lib/lintSource.js
```diff
@@ -251,15 +251,16 @@ function createEmptyPostcssResult(filePath) {
...
- * literals, it will duplicate some code. More info in issue #4119.
  *
  * @param {PostcssResult} postcssResult
  * @returns {boolean}
  */
 function isFixCompatible({ stylelint }) {
 	// Check for issue #2643
-	if (stylelint.disabledRanges.all.length) return false;
+	const hasUnscopedDisabledRanges = stylelint.disabledRanges.all.length > 0;
+	const hasScopedDisabledRanges = Object.keys(stylelint.disabledRanges).length > 1;
```
ota-meshi, line 261: I think you can turn off autofix for each scoped rule.
Change the scope data to be passed to the third argument of `ruleFunction`.

jeddy3: > I think you can turn off autofix for each scoped rule.

Am I right in thinking that would give us a more granular workaround where autofix is turned off (for the entire source):

- for all rules, if the source contains unscoped disable comments
- for only the applicable rules, if the source only contains scoped disable comments

Would you mind pulling down the pull request and showing me what you mean? And is it something we should add to this pull request or a follow-up?

ota-meshi: Hi @jeddy3.
I want to fix it like this:

https://github.com/stylelint/stylelint/compare/issue-4704..issue-4704-2

ota-meshi: The color-hex-length rule does not autofix.

https://github.com/stylelint/stylelint/compare/issue-4704..issue-4704-2#diff-01cbf9c2079c6adc0ce42f035fd16a45R245

jeddy3: That's a much better approach, thanks!

If possible, can you:

- use `require('common-tags').stripIndent` to indent the `indentation` rule code snippets, like so:

```js
                          .lint({
				code: stripIndent`
                                     /* stylelint-disable color-hex-length */
                                    a {
                                       color: #ffffff;
                                    }
                                `,
```


- add a comment above the following:

```
// Next two conditionals are temporary measures until #2643 is resolved
isFileFixCompatible &&
!postcssResult.stylelint.disabledRanges[ruleName],
```

As their intent isn't immediately clear.

We should then merge your changes into this branch.

ota-meshi: I fixed it and merged it into this branch.

jeddy3:
I've added a commit to this pull request to update the docs to reflect the new behaviour introduced by this fix. I believe that autofix is now useful and stable for the majority of users. Those using scoped disable comments are only slightly inconvenienced when using the `fix` option. The minority of users using unscoped disable comments are more inconvenienced, but `fix` remains safe for them to use.

jeddy3:
> - Fixed: autofix will respect scoped disable comments by turning off autofix for the scoped rules for the entire source; this is a continuation of the workaround added in `13.2.0` ([#4705](https://github.com/stylelint/stylelint/pull/4705)).

PR #4714, opened 2020-04-22
Make --report-needless-disables CLI flag respect stylelint-disable commands

### LINKED ISSUE #4203: Make --report-needless-disables CLI flag respect stylelint-disable commands
> Clearly describe the bug

When using `--report-needless-disables`, it ignores all of my `stylelint-disable`s.

all

```css
/* stylelint-disable selector-max-type */
/* stylelint-disable selector-max-universal */

*,
::after,
::before {
	box-sizing: border-box;
	padding: 0;
	margin: 0;
}

a {
	color: inherit;
	text-decoration: inherit;
}

```

e.g.

```json
{
  "rules": {
    "selector-max-type": 0,
    "selector-max-universal": 0
  }
}
```

Commit `d3a6cb47fc2d5c59e61f6fac618c6121b316b97c` (it comes with some `--report-needless-disables` fix that I need since the last release)

`stylelint --report-needless-disables`

The problem applies to both

For it to disable the rules

They were not disabled, receive these errors:
```
src/index.css
  6:1  ✖  Expected "*" to have no more than 0 universal selectors        selector-max-universal
 14:1  ✖  Don't style html elements. It leads to css specificity issues  selector-max-type
```

Without the flag, these errors do not appear.

I messed up reporting #4190, it happens even without `--fix`.

ota-meshi:

Closes #4203

This PR modifies ignoreDisables to not be implicitly used when using the reportNeedlessDisables option.

### DIFF
--- a/lib/__tests__/needlessDisables.test.js
+++ b/lib/__tests__/needlessDisables.test.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const needlessDisables = require('../needlessDisables');
 const path = require('path');
 const replaceBackslashes = require('./replaceBackslashes');
 const standalone = require('../standalone');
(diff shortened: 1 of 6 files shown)


thibaudcolas:
Wow! I only half understand the change, but it’s great to see this fixed. Thank you for figuring this out.

jeddy3:
Changelog:

> - Fixed: `--report-needless-disables` respects stylelint-disable commands ([#4714](https://github.com/stylelint/stylelint/pull/4714)).

Issue #3935, opened 2019-02-01
Building Stylelint for client-side


oskarkrawczyk:
I'm currently trying to replace CSSLint in @jsfiddle with Stylelint. Got it working thanks to a proof-of-concept made here https://github.com/m-allanson/stylelint-browser-demo

Unfortunately turns out this works fine but with a very old version of Stylelint, specifically the 5.x branch. Once updated deps to the newest Stylelint, the whole client-side bundle breaks apart with a slightly cryptic `Error: Cannot find module "."` error.

I was wondering if anyone around here can help out figure out if it's even possible to run the newest Stylelint on the client-side (this is a requirement for JSFiddle as our env only supports linting that way).

---

A bit more on what my config looks like:

`webpack.config`:

```js
import webpack from "webpack"
import path from "path"

export default {
  entry: "./site/index.js",
  output: {
    path: path.resolve(__dirname, "site"),
    publicPath: "/",
    filename: "bundle.js",
  },
  resolve: {
    root: path.resolve(__dirname),
    alias: {
      "resolve-from": "empty-module",
      cosmiconfig: "empty-module",
      doiuse: "empty-module",
      globby: "empty-module",
      globjoin: "empty-module",
      multimatch: "empty-module",
      path: "empty-module",
    },
    modulesDirectories: [
      "node_modules",
      "site",
    ]
  },
  module: {
    loaders: [
      {
        test: /\.js$/,
        loader: "babel-loader",
        exclude: /node_modules/,
      },
      {
        test: /\.json$/,
        loader: "json-loader",
      },
    ],
  },
  plugins: [
    new webpack.optimize.DedupePlugin(),
    new webpack.ContextReplacementPlugin(/stylelint/, /NEVER_MATCH^/),
  ],
  node: {
    fs: "empty",
    module: "empty",
  },
}
```

`package.json`:

```json
{
  "name": "stylelint-browser-demo",
  "version": "0.0.1",
  "description": "A proof-of-concept to demo stylelint running in a browser.",
  "main": "site/bundle.js",
  "scripts": {
    "start": "npm run build && http-server ./site",
    "start-dev": "webpack-dev-server --content-base site/ --hot --inline",
    "build": "webpack --progress --colors && stat -f%z site/bundle.js",
    "lint": "eslint . --ignore-path .gitignore",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "m-allanson",
  "license": "MIT",
  "dependencies": {
[...]

PR #4729, opened 2020-05-01
Standardise lazy requires

jeddy3:

https://github.com/stylelint/stylelint/issues/2454

In this pull request, I've standardised how we lazy/dynamic-require:

- formatters
- rules
- syntaxes

Previously we were using lots of different workarounds to get the require-time down. I'm now only using [import-lazy](https://github.com/sindresorhus/import-lazy#readme).

This lays the foundation for being able to bundle code, closing https://github.com/stylelint/stylelint/issues/2454 and https://github.com/stylelint/stylelint/issues/3935.

This pull request doesn't improve the require-time much, because we'd already optimised all that we could without bundling (using the various workarounds). Require-time does seem more consistent, though, normally coming in at the `120` mark.

This pull request does, however, improve the time taken to run stylelint when the auto syntax is used because all the syntaxes are now lazy-required.

Testing locally with 

```js
{
  "code": "a { color: #fff; }", // or sass equivalent
  "config": "{ rules: { "color-hex-length": "long" } }"
}
```

On `master`:

```console
// no syntax
node test.js  0.44s user 0.14s system 0.651 total

// no syntax, sass file
node test.js  0.43s user 0.10s system 0.787 total

// syntax: “css”
node test.js  0.17s user 0.05s system 0.294 total
```

This branch:

```console
// no syntax
node test.js  0.16s user 0.04s system 0.232 total

// no syntax, sass file
node test.js  0.19s user 0.04s system 0.221 total

// syntax: “css”
node test.js  0.16s user 0.04s system 0.200 total
```

Using the auto syntax is almost comparable to using the `syntax` option in this branch, whereas it's significantly slower in `master`.

This improvement is reflected in the test suite which now runs in 72s (down from 92s).

The import-lazy library has two patterns:
- [standard](https://github.com/sindresorhus/import-lazy#usage)
- [for bundlers](https://github.com/sindresorhus/import-lazy#usage-with-bundlers)

I'm using the former for formatters and rules, and the latter for syntaxes. This will allow us to bundle syntaxes separately in the future.

As a consequence, it exposed a shortcoming in how we were excluding files from TypeScript. Previously all the rule files were ignored because they were excluded _and dynamically required using `require("rules" + rulename)`_, which TypeScript couldn't resolve. As we are now using a lazy-require pattern it can resolve, we have to explicitly mark each rule file with a `// @ts-nocheck`. This is arguably cleaner as we were previously relying on [a quirk of the TypeScript's module resolution](https://www.typescriptlang.org/docs/handbook/module-resolution.html#common-questions). The `include` and `exclude` TypeScript config options are cleaner too now.


types/import-lazy/index.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module 'import-lazy' {
```
m-allanson, line 1: Can this be moved to `types/global.d.ts`?

jeddy3:
> - Fixed: improved performance when auto syntax is used ([#4729](https://github.com/stylelint/stylelint/pull/4729)).

PR #4743, opened 2020-05-06
Add an ignore option to time-min-milliseconds

### LINKED ISSUE #4552: Add ignore:["delay"] to time-min-milliseconds

Animations/Transitions that take less than 150ms should be omitted, since this time is too short to provide a visual improvement.

On the other hand, I'm using a delay of 40ms to defer style changes for hover effects. E.g. if you quickly move the cursor over a input box, scrollbar or button, it should not change its style. Only if you leave the cursor on the element for atleast 40ms. This prevents flickering, while quickly moving the cursor.

The "time-min-milliseconds" rule should provide options to specify different values for delay and duration.

srawlins:

Fixes #4552

Certainly.

* Previously, this rule considered the value to the shorthand properties to be space-separated lists. It is more correct to view them as comma-separated lists of animations (or transitions) whose values themselves are space-separated lists. This distinction becomes important when ignoring delay values in specific animation (or transition) values. ([MDN: [transition](https://developer.mozilla.org/en-US/docs/Web/CSS/transition), [animation](https://developer.mozilla.org/en-US/docs/Web/CSS/animation))
* Some of the test cases had transposed the timing function and the delay. This is fixed.

### DIFF
--- a/lib/rules/time-min-milliseconds/README.md
+++ b/lib/rules/time-min-milliseconds/README.md
@@ -60,3 +60,18 @@ a { transition: background-color 600ms linear; }
 ```css
 a { animation-delay: 1s; }
 ```
+
+## Optional secondary options
+
+### `ignore: ["delay"]`
+
+Ignore time values for an animation or transition delay.
+
+For example, with a minimum of `200` milliseconds.
+
+The following is _not_ considered a violation:
+
+<!-- prettier-ignore -->
+```css
+a { animation-delay: 100ms; }
+```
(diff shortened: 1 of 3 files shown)


srawlins:
Wow thanks for the head start on the bug about violation locations. But I think I'd prefer to tackle that in a new PR.

jeddy3:
> But I think I'd prefer to tackle that in a new PR.

Sounds good to me. I've created a new issue for it at https://github.com/stylelint/stylelint/issues/4751.

jeddy3:
Changelog:

> - Added: `ignore:["delay"]` to `time-min-milliseconds` ([#4743](https://github.com/stylelint/stylelint/pull/4743)).

@srawlins Thanks again for the pull request. Apologies for the delay in getting it merged. Some [big](https://github.com/stylelint/stylelint/pull/4758) [chunks](https://github.com/stylelint/stylelint/pull/4755) of work were happening in the background.

PR #4749, opened 2020-05-06
Fix false positives for objects in declaration-block-trailing-semicolon

### LINKED ISSUE #4650: Fix false positives for objects in declaration-block-trailing-semicolon
> Clearly describe the bug

Stylelint emits a warning when declaration-block-trailing-semicolon meets styles in jsx

`declaration-block-trailing-semicolon`

```js
const Component = () => {
  const background = 'white'

  return <div style={{ background }}>Text</div>
}
```

```json
{
  "rules": {
    "declaration-block-trailing-semicolon": "always"
  }
}
```

e.g. `13.2.1`

CLI

Yes, it's related to JSX

no warnings

```
Expected a trailing semicolon   declaration-block-trailing-semicolon
```

I've monkeypatched this in my code through patch-package by adding 

```diff
diff --git a/node_modules/stylelint/lib/rules/declaration-block-trailing-semicolon/index.js b/node_modules/stylelint/lib/rules/declaration-block-trailing-semicolon/index.js
index 890dd96..052e43c 100644
--- a/node_modules/stylelint/lib/rules/declaration-block-trailing-semicolon/index.js
+++ b/node_modules/stylelint/lib/rules/declaration-block-trailing-semicolon/index.js
@@ -40,6 +40,10 @@ function rule(expectation, _, context) {
 		});
 
 		root.walkDecls((decl) => {
+			if (decl.parent.type === 'object') {
+				return;
+			}
+
 			if (decl !== decl.parent.last) {
 				return;
 			}
```

srawlins:

Fixes #4650 

Yes; @jeddy3 suggests a solution in https://github.com/stylelint/stylelint/issues/4650#issuecomment-599205149 that did not work for me: `isStandardSyntaxRule` only accepts Rule nodes, but if used in `declaration-block-trailing-semicolon`, it would also be given AtRule nodes (like in the SCSS test case: `a { @foo { color: pink } }`). This complicates things further because `isStandardSyntaxRule` examines the `selector` of the Rule which it receives, and AtRule has no `selector`.

This simple solution seems to work for now.

### DIFF
--- a/lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
+++ b/lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
@@ -24,10 +24,6 @@ testRule({
 			code: 'a { color: red; &:hover { color: pink; }}',
 			description: 'nesting with first-level decl',
 		},
-		{
-			code: 'background: white',
-			description: 'not a rule',
-		},
 	],
 
 	reject: [
@@ -79,10 +75,6 @@ testRule({
 		{
 			code: 'a { background: orange; color: pink }',
 			description: 'multi-line declaration block without trailing semicolon',
-		},
-		{
-			code: 'background: white;',
-			description: 'not a rule',
 		},
 	],
 
@@ -176,6 +168,31 @@ testRule({
 			message: messages.rejected,
 			line: 1,
 			column: 22,
+		},
+	],
+});
+
+testRule({
+	ruleName,
+	config: ['always'],
+	syntax: 'css-in-js',
+	fix: true,
+
+	accept: [
+		{
+			code: 'const C = () => { return <a style={{ color: "red" }}></a> }',
+			description: 'css-in-js object',
+		},
+	],
+
+	reject: [
+		{
+			code: 'const C = styled.a`color: red`;',
+			fixed: 'const C = styled.a`color: red;`;',
+			description: 'css-in-js template literal',
+			message: messages.expected,
+			line: 1,
+			column: 29,
 		},
 	],
 });
--- a/lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
+++ b/lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
@@ -24,6 +24,10 @@ testRule({
 			code: 'a { color: red; &:hover { color: pink; }}',
 			description: 'nesting with first-level decl',
 		},
+		{
+			code: 'background: white',
+			description: 'not a rule',
+		},
 	],
 
 	reject: [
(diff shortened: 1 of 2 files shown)


lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
```diff
@@ -25,6 +25,10 @@ testRule(rule, {
 			code: 'a { color: red; &:hover { color: pink; }}',
 			description: 'nesting with first-level decl',
 		},
+		{
```
jeddy3, line 28: Let's replace these two tests with a single one at the bottom of the file that tests the problematic syntax, e.g.:

```js
testRule(rule, {
	ruleName,
	config: ['always'],
	syntax: 'css-in-js',
	fix: true,

	accept: [
		{
			code: 'const C = () => { return <a style={{ color: "red" }}></a> }',
			description: 'css-in-js object',
		},
	],

	reject: [
		{
			code: 'const C = styled.a`color: red`;',
			fixed: 'const C = styled.a`color: red;`;',
			description: 'css-in-js template literal',
			message: messages.expected,
			line: 1,
			column: 29,
		},
	],
});
```

So that [we're consistent with the conventions](https://stylelint.io/developer-guide/rules#write-tests):

> You should:
> - use standard CSS syntax by default, and only swap parsers when testing a specific piece of non-standard syntax

I think this will also highlight that template literals are incorrectly excluded by the proposed fix.

lib/rules/declaration-block-trailing-semicolon/index.js
```diff
@@ -42,6 +42,10 @@ function rule(expectation, _, context) {
 		});
 
 		root.walkDecls((decl) => {
+			if (decl.parent.type !== 'rule' && decl.parent.type !== 'atrule') {
```
jeddy3, line 45: I think we should use:

```js
if (decl.parent.type === 'object') return;
```

As not to exclude template literals, e.g.:

```js
const C = styled.a`color: red;`;
```

lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
jeddy3: I've updated the snippet above to use the new `testRule` signature because we now dogfood our Jest preset for testing rules (https://github.com/stylelint/stylelint/pull/4758).

lib/rules/declaration-block-trailing-semicolon/index.js
srawlins: Done

lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
srawlins: Ah, very nice. Thanks much!

jeddy3:
Changelog entry:

> - Fixed: `declaration-block-trailing-semicolon` false positives for CSS-in-JS object notation ([#4749](https://github.com/stylelint/stylelint/pull/4749)).

PR #4750, opened 2020-05-07
Document explicitly checking type property when using forEach

jeddy3:

Ref https://github.com/stylelint/stylelint/pull/4744#discussion_r420705135

No, it's self-explanatory.

### DIFF
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -25,7 +25,9 @@ You should make use of the:
 
 #### PostCSS API
 
-Use the [PostCSS API](https://api.postcss.org/) to navigate and analyze the CSS syntax tree. We recommend using the `walk` iterators (e.g. `walkDecls`), rather than using `forEach` to loop through the nodes. If `forEach` is a better fit, you should explicitly check the `type` property of the `node` before attempting to access other properties. For example:
+Use the [PostCSS API](https://api.postcss.org/) to navigate and analyze the CSS syntax tree. We recommend using the `walk` iterators (e.g. `walkDecls`), rather than using `forEach` to loop through the nodes.
+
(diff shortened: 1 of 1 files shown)


docs/developer-guide/rules.md
```diff
@@ -25,7 +25,15 @@ You should make use of the:
 
 #### PostCSS API
 
-Use the [PostCSS API](https://api.postcss.org/) to navigate and analyze the CSS syntax tree. We recommend using the `walk` iterators (e.g. `walkDecls`), rather than using `forEach` to loop through the nodes. Use `node.raws` instead of `node.raw()` when accessing raw strings from the [PostCSS AST](https://astexplorer.net/#/gist/ef718daf3e03f1d200b03dc5a550ec60/c8cbe9c6809a85894cebf3fb66de46215c377f1a).
+Use the [PostCSS API](https://api.postcss.org/) to navigate and analyze the CSS syntax tree. We recommend using the `walk` iterators (e.g. `walkDecls`), rather than using `forEach` to loop through the nodes. If `forEach` is a better fit, you should explicitly check the `type` property of the `node` before attempting to access other properties. For example:
```
m-allanson, line 28: How can I identify when `forEach` would be a better fit?

m-allanson:
Looks good! One comment but I think this is good to merge regardless.

docs/developer-guide/rules.md
jeddy3: Good question. It highlights that `forEach` is never a good fit.

I've amended the pull request to clarify:

- that checking the type is for when an array method is used
- it is quite separate from the `forEach` advice

m-allanson: 👍

PR #4758, opened 2020-05-09
Use jest-preset-stylelint

hudochenkov:
~~Unfortunately, we can't use `jest-preset-stylelint` as preset (`"preset": "jest-preset-stylelint"`) in Jest configuration, because preset [`require('stylelint')`](https://github.com/stylelint/jest-preset-stylelint/blob/d6ae61776997fae9b583177420517dbd3e3d6c5f/jest-setup.js#L4) and we don't have any installed :)~~ Damn, Jest is powerful. Solved with `moduleNameMapper`. We might not need to have stylelint passed to `jest-preset-stylelint/getTestRule` for plugin authors.

It works great on macOS and Ubuntu if we don't have `getTestRule` in our `jest-setup`. However Windows tests are failing. Probably, because mock is not applied or applied after `testRule` from preset is added to Jest process.

We also have to live with the following warning after every `npm install`:

```
$ npm i -D jest-preset-stylelint
npm WARN jest-preset-stylelint@2.0.0 requires a peer of stylelint@^13.0.0 but none is installed. You must install peer dependencies yourself.
```


hudochenkov:
Nope, looks like mock doesn't work if preset is specified. I'll live this PR without `preset` specified. However, after this PR I will try to add mock to `jest-preset-stylelint` and try to make stylelint work with it.

hudochenkov:
I found a way to simplify stylelint Jest configuration. Waiting on https://github.com/stylelint/jest-preset-stylelint/pull/21 and new `jest-preset-stylelint` release.

**Don't review this PR yet.**

hudochenkov:
It is ready for a review!

hudochenkov:
Good one, @jeddy3. I made the change.

.prettierignore
```diff
@@ -1,4 +1,5 @@
 .coverage
+.vscode
```
ntwb, line 2: I'm fine with this, though it does peek my interest, should we add some VS Code settings to the repo?

jeddy3: > I'm fine with this, though it does peek my interest

Ditto. For example, I wonder if it'll be helpful to include a launch file for debugging when running jest?

hudochenkov: We should not, because I don't want anyone changing setting on my computer. And I think many people don't want it either. I have test cases in `.vscode` when trying to debug something, for example.

PR #4760, opened 2020-05-09
Add color-function-notation

### LINKED ISSUE #4718: Add color-function-notation

I want to disallow the use of (comma-based) legacy color function notation so that I only wrote (space-based) future-facing syntax, and transpile it if necessary with [postcss-preset-env](https://preset-env.cssdb.org/features).

- [Spec](https://www.w3.org/TR/css-color-4/#rgb-functions)
- [CanIUse](https://caniuse.com/#feat=mdn-css_types_color_space_separated_functional_notation)
- [Relevant tweet](https://twitter.com/mathias/status/1253242715304857601)

Legacy notation:

```css
a { color: rgb(0, 0, 0, 0.5) }
```

Modern notation:

```css
a { color: rgb(0 0 0 / 50%) }
```

A new rule to do this. Some users, like with pseudo-element colon notation, might want to enforce the legacy style. I propose a rule with two options: "space" and "comma" (unless there are better names?).

Any thoughts on the use case, rule name or options?

jeddy3:

Closes #4718

I thought I'd give this one a whirl as it's been a while since I wrote a rule.

The `"modern"` option is autofixable, and I've addressed the edge-cases I can think of.

There may be a neater way to write the fixing code that doesn't rely on a count.

### DIFF
--- a/lib/rules/color-function-notation/README.md
+++ b/lib/rules/color-function-notation/README.md
@@ -1,0 +1,116 @@
+# color-function-notation
+
+Specify modern or legacy notation for applicable color-functions.
+
+<!-- prettier-ignore -->
+```css
+    a { color: rgb(0 0 0 / 0.2) }
+/**            ↑
+ *             This notation */
+```
+
+Modern color-functions use a comma-free syntax because functions in CSS are used to group/name a syntax chunk. They should work by the same rules that CSS grammar does in general: values are optional and re-orderable when possible, space-separated, and commas are used to separate repetitions only.
+
+For legacy reasons, `rgb()` and `hsl()` also supports an alternate syntax that separates all of its arguments with commas. Also for legacy reasons, the `rgba()` and `hsla()` functions exist using the same comma-based syntax.
+
+The [`fix` option](../../../docs/user-guide/usage/options.md#fix) can automatically fix some of the problems reported by this rule when the primary option is `"modern"`.
+
+## Options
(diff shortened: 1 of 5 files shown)


lib/rules/color-function-notation/README.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+<!-- prettier-ignore -->
+```css
+a { color: hsl(.75turn 60% 70%) }
+```
+
+### `"legacy"`
+
+Applicable color-functionss _must always_ use the legacy notation.
```
m-allanson, line 72:
```suggestion
Applicable color-functions _must always_ use the legacy notation.
```

jeddy3: Good catch.

jeddy3:
@malsf21 Thank you for the review! 

> Not the most experienced dev on the topic

With the bugs that you've been squashing, you have more familiarity with stylelint than most! We're a small group of contributors and so any help makes an impact and is appreciated (whether that's squashing bugs, reviewing pull requests or triaging issues!).

---

Let's merge this pull request after we release `13.4.1` (https://github.com/stylelint/stylelint/issues/4791) and when the other two related rules are approved.

jeddy3:
> - Added: `color-function-notation` rule ([#4760](https://github.com/stylelint/stylelint/pull/4760)).

PR #4769, opened 2020-05-13
Add hue-degree-notation

### LINKED ISSUE #4721: Add hue-degree-notation

I'd like to enforce a consistent hue notation.

From https://github.com/stylelint/stylelint/issues/4718#issuecomment-619336037:

"- consider being opinionated about using the `deg` suffix where applicable, e.g. in the first parameter for `hsl()` and the last non-alpha parameter for `lch()`. Again, because this clarifies that the range lies between `0-360`."

References:

- [Spec](https://www.w3.org/TR/css-color-4/#typedef-hue)
- [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#HSL_colors)

A new rule that lets me enforce either a `<number>` or `<angle>` notation for `<hue>`.

Primary option: `string: "number"|"angle"`

With `"number"` the following are valid:

```css
a { 
  color: hsl(198 28% 50%);
  color: lch(56.29% 19.86 236.62 / 50%);
  ..
}
```

With `"angle"` the following are valid:

```css
a { 
  color: hsl(198deg 28% 50%);
  color: lch(56.29% 19.86 1turn / 50%);
  hsl(4.71239rad, 60%, 70%);
  ..
}
```

I can, if desired, enforce the [type of angle](https://www.w3.org/TR/css3-values/#angle-value) using [`unit-whitelist`](https://stylelint.io/user-guide/rules/unit-whitelist).

Any thoughts about the use case, name or options?

jeddy3:

Closes #4721

The rule can autofix both primary options as this rule is about explicitness vs shorthand convenience (with neither option being more right than the other), unlike [`color-function-notation`](https://github.com/stylelint/stylelint/pull/4760) which is concerned with modern vs legacy syntax.

### DIFF
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -77,6 +77,10 @@ Grouped first by the following categories and then by the [_thing_](http://apps.
 
 ## Limit language features
 
+### Hue
+
+- [`hue-degree-notation`](../../../lib/rules/hue-degree-notation/README.md): Specify number or angle notation for degree hues.
+
 ### Color
 
 - [`color-named`](../../../lib/rules/color-named/README.md): Require (where possible) or disallow named colors.
--- a/lib/rules/hue-degree-notation/README.md
+++ b/lib/rules/hue-degree-notation/README.md
@@ -1,0 +1,74 @@
+# hue-degree-notation
+
+Specify number or angle notation for degree hues.
+
+<!-- prettier-ignore -->
+```css
+    a { color: hsl(198deg 28% 50%) }
+/**                ↑
+ *                 This notation */
+```
+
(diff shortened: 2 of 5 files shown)


jeddy3:
> but should we discuss the behaviour of the rule in regards to CSS functions and variables (e.g. calc() and var())?

@malsf21 The documentation for rules tend to focus on the simplest example, as to communicate the intent of the rule. Each rule [usually has countless edge-cases](https://stylelint.io/developer-guide/rules#commonly-overlooked-edge-cases) that would if documented make the intent of the rule more difficult to understand.

jeddy3:
Changelog:

> - Added: `hue-degree-notation` rule ([#4769](https://github.com/stylelint/stylelint/pull/4769)).

PR #4770, opened 2020-05-14
Add alpha-value-notation

### LINKED ISSUE #4720: Add alpha-value-notation

I'd like to enforce a consistent alpha-value notation.

From https://github.com/stylelint/stylelint/issues/4718#issuecomment-619336037:

"consider being opinionated w.r.t. how to represent alpha values. IMHO using percentage notation is most readable, as it clarifies the range: e.g. `70%` vs `0.7`"

References:

- [Spec](https://www.w3.org/TR/css-color-4/#typedef-alpha-value)
- [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/alpha-value)

A new rule that lets me enforce either a `<number>` or `<percentage>` notation for `<alpha-value>`.

Primary option: `string: "number"|"percentage"`

With `"number"` the following are valid:

```css
a { 
  color: rgba(0, 0, 0, 0.5)
  color: rgb(0 0 0 / 0.5)
  color: hsl(198deg 28% 50% / 0.5);
  opacity: 0.5;
  shape-image-threshold: 0.5;
  ..
}
```

With `"percentage"` the following are valid:

```css
a { 
  color: rgba(0, 0, 0, 50%)
  color: rgb(0 0 0 / 50%)
  color: hsl(198deg 28% 50% / 50%);
  opacity: 50%;
  shape-image-threshold: 50%;
  ..
}
```

Any thoughts about the use case, name or options?

jeddy3:

Closes #4720 

I added an `exceptProperties` option as support for percentage alpha-values for `opacity` was added in a [later spec](https://drafts.csswg.org/css-color/#transparency), it [isn't supported in some browsers like Safari](https://developer.mozilla.org/en-US/docs/Web/CSS/opacity#Browser_compatibility), and  I don't think [postcss-preset-env](https://preset-env.cssdb.org/) transpiles it yet.

When this rule and the other two new rules are in, I'll create a follow-up issue to refactor some of the shared functions into utils, e.g. `getValue` and `setValue`.

### DIFF
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -77,9 +77,13 @@ Grouped first by the following categories and then by the [_thing_](http://apps.
 
 ## Limit language features
 
+### Alpha-value
+
+- [`alpha-value-notation`](../../../lib/rules/alpha-value-notation/README.md): Specify percentage or number notation for alpha-values (Autofixable).
+
 ### Hue
 
-- [`hue-degree-notation`](../../../lib/rules/hue-degree-notation/README.md): Specify number or angle notation for degree hues.
+- [`hue-degree-notation`](../../../lib/rules/hue-degree-notation/README.md): Specify number or angle notation for degree hues (Autofixable).
 
 ### Color
 
--- a/lib/rules/alpha-value-notation/README.md
+++ b/lib/rules/alpha-value-notation/README.md
@@ -1,0 +1,110 @@
+# alpha-value-notation
+
+Specify percentage or number notation for alpha-values.
+
+<!-- prettier-ignore -->
+```css
+    a { color: rgb(0 0 0 / 0.5) }
+/**                        ↑
+ *                         This notation */
+```
+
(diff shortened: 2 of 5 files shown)


mathiasbynens:
> I added an `exceptProperties` option as support for percentage alpha-values for `opacity` was added in a [later spec](https://drafts.csswg.org/css-color/#transparency), it [isn't supported in some browsers like Safari](https://developer.mozilla.org/en-US/docs/Web/CSS/opacity#Browser_compatibility), and I don't think [postcss-preset-env](https://preset-env.cssdb.org/) transpiles it yet.

Note that the latest stable Safari version (13.1) supports `opacity: 50%` according to `CSS.supports('opacity', '50%');`. I don't know what the oldest Safari version is that supports it though. Also, Opera is Chromium-based, so it likely actually has support for it too. TL;DR Take the MDN support data with a grain of salt. (I think `exceptProperties` is a good idea regardless 👍)

jeddy3:
@mathiasbynens Thank you for the review. I'll release the new rules now.

---

Changelog entry:

> - Added: `alpha-value-notation` rule ([#4770](https://github.com/stylelint/stylelint/pull/4770)).

PR #4775, opened 2020-05-15
Fix false positives for forced-colors in media-feature-name-* rules

### LINKED ISSUE #4773: Fix support for forced-colors in media-feature-name-*
- Docs: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors
- Spec: https://drafts.csswg.org/mediaqueries-5/#forced-colors
- ChromeStatus entry: https://www.chromestatus.com/feature/5757293075365888

mathiasbynens:

Closes #4773.

- Docs: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors
- Spec: https://drafts.csswg.org/mediaqueries-5/#forced-colors
- ChromeStatus entry: https://www.chromestatus.com/feature/5757293075365888

### DIFF
--- a/lib/reference/keywordSets.js
+++ b/lib/reference/keywordSets.js
@@ -527,8 +527,8 @@ keywordSets.pageMarginAtRules = new Set([
 
 // https://developer.mozilla.org/en/docs/Web/CSS/At-rule
 keywordSets.atRules = uniteSets(keywordSets.pageMarginAtRules, [
-	'apply',
 	'annotation',
+	'apply',
 	'character-variant',
 	'charset',
 	'counter-style',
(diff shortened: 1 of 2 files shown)


mathiasbynens:
> ```diff
> -https://drafts.csswg.org/mediaqueries/
> +https://drafts.csswg.org/mediaqueries-5/
> ```
> 
> The former is for Media Queries Level 4, and the latter is for Media Queries Level 5.

I _think_ the former is just the permanent link to whatever the latest draft is. The latter is a fixed snapshot for level 5, like how https://drafts.csswg.org/mediaqueries-4/ is a fixed snapshot for level 4. @argyleink and @una can probably clarify if this is correct. IMHO we should continue pointing to the latest draft.

ybiquitous:
@mathiasbynens That makes sense. Thank you for your detailed explanation! 🥰

jeddy3:
Changelog entry:

> - Fixed: `media-feature-name-*` false positives for `forced-colors` ([#4775](https://github.com/stylelint/stylelint/pull/4775)).

PR #4778, opened 2020-05-16
Fix false positives for WebExtension replacement keywords in value-keyword-case

### LINKED ISSUE #4707: Fix false positives for WebExtension replacement keywords in value-keyword-case
> Clearly describe the bug

Chrome/Firefox extensions have CSS keywords that are replaced at runtime and are case-sensitive. For example:

```css
body {
	direction: __MSG_@@bidi_dir__;
}
```

Which the browser interprets as:

```css
body {
	direction: ltr;
	/* or direction: rtl; */
}
```

These are mostly used for localization/internalization. Documentation: https://developer.chrome.com/extensions/i18n

`value-keyword-case`

```json
{
  "rules": {
    "value-keyword-case": "lower"
  }
}
```

`13.3.2`

CLI

Kinda. However it's parsed by the browsers, not a transpiler.

Keywords that match `/__MSG_[^\s]+__/` should be skipped by the rule.

```shell
test.css
 2:13  ✖  Expected "__MSG_@@bidi_dir__" to be "__msg_@@bidi_dir__"   value-keyword-case
```

malsf21:
Hi there, 

This is a PR that aims to address #4707, by adding a regex match to `isStandardSyntaxValue` that ignores values of the form `/__MSG_[^\s]+__/`, a set of runtime values often used by browsers for internationalization (more information [here](https://developer.chrome.com/extensions/i18n), as mentioned in the issue discussion). I also added two tests, one that checks for the positive, and one that checks for case-sensitivity.

As `value-keyword-case` uses `isStandardSyntaxValue`, this should resolve the false positive originally brought up in the issue.

Locally, this PR passes `npm run format` and `npm run test`.

Closes #4707.

Not on my end! Let me know if I should change the comments or add more tests.

### DIFF
--- a/lib/utils/__tests__/isStandardSyntaxValue.test.js
+++ b/lib/utils/__tests__/isStandardSyntaxValue.test.js
@@ -42,4 +42,10 @@ describe('isStandardSyntaxValue', () => {
 	it('less interpolation', () => {
 		expect(isStandardSyntaxValue('@{var}')).toBeFalsy();
 	});
+	it('WebExtension replacement keyword', () => {
+		expect(isStandardSyntaxValue('__MSG_@@bidi_dir__')).toBeFalsy();
+	});
+	it('negative WebExtension replacement keyword', () => {
+		expect(isStandardSyntaxValue('__msg_@@bidi_dir__')).toBeTruthy();
+	});
 });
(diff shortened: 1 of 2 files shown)


jeddy3:
> - Fixed: `value-keyword-case` false positives WebExtension replacement keywords ([#4778](https://github.com/stylelint/stylelint/pull/4778)).

Issue #4782, opened 2020-05-17
Fix TypeError: Cannot read property 'length' of undefined in time-min-milliseconds 


klimashkin:
After upgrading from 13.2.0 to 13.4.0 I've started getting following errors:

1. **Animation**
```css
.progressStop {
  animation: none;
}
```
Throws error
<img width="343" alt="Screen Shot 2020-05-17 at 2 37 27 AM" src="https://user-images.githubusercontent.com/576077/82140946-5bb43e00-97e7-11ea-84b1-56bf155a7abb.png">
Changing it to 
```css
.progressStop {
  animation-name: none;
}
```
fixes the problem

2. **Transition**
This code
```css
.progressTo100Only {
  transition: opacity var(--button-progress-fadeout-duration) var(--button-progress-fadeout-delayTo100Only) ease-out;
}
``` 
make stylelint fail with 
<img width="864" alt="Screen Shot 2020-05-17 at 2 36 30 AM" src="https://user-images.githubusercontent.com/576077/82140927-39babb80-97e7-11ea-8eda-12afe6cdeab1.png">
Could not find a workaround

jeddy3:
@klimashkin Thanks for the report. 

It looks to be a regression. I'll try tracking it down.

Are you turning on the [`time-min-milliseconds`](https://stylelint.io/user-guide/rules/time-min-milliseconds) rule in your config, and are you using it with the [new `ignore: ["delay"]` option](https://stylelint.io/user-guide/rules/time-min-milliseconds#ignore-delay)? Or did the errors occur without any changes to your config after updating to `13.4.0`?

And to confirm, you're updating from `13.2.0` and not `13.3.3`?

Lastly, can you provide a stack trace? Running stylelint on the CLI and outside or gulp should do it, e.g.:

```
npx stylelint "**/*.{css,scss}"
```

jeddy3:
I was able to produce a TypeError using the new `ignore: ["delay"]` option for ` time-min-milliseconds`, but I can't reproduce one otherwise using the information you gave. I'll need more information to go further. In addition to the four questions above, please can you also provide a config that triggers the error?

---

I've created a pull request (https://github.com/stylelint/stylelint/pull/4783) to fix the TypeError for the `"delay`" option

klimashkin:
@jeddy3 Oh shoot, you are right! I forgot that I've actually enabled the new `ignore: ["delay"]` option along with the update, sorry about that.

If I remove it, everything is fine.

[...]

Issue #4751, opened 2020-05-07
Fix incorrect location for matching violating times in time-min-milliseconds


jeddy3:
> Clearly describe the bug

If there are two or more of the same violating times within a value they will all be reported at the location of the first violations.

We'll need to use the value parser to find the times so that we can make use of its `sourceIndex` property.

Ref: https://github.com/stylelint/stylelint/pull/4743

> Which rule, if any, is the bug related to?

`time-min-milliseconds`

> What code is needed to reproduce the bug?

```css
a { animation: foo 0.01s ease 0.8s, bar 0.8s ease 0.01s; }
```

> What stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "time-min-milliseconds": "200ms"
  }
}
```

> Which version of stylelint are you using?

`13.3.3`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

CLI with `stylelint "**/*.css"`

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

No.

> What did you expect to happen?

Warnings to be flagged at the correct locations.

> What actually happened (e.g. what warnings or errors did you get)?

Each warning is flagged at the first location.

mattxwang:
I think I can address this in lockstep with adding end positions to the rule; going to assign myself!

kawaguchi1102:
Hi @mattxwang .

Are you working on this issue ?
If not, I am ready to do this .

mattxwang:
Ah - @kawaguchi1102, someone else ended up handling the end positions so I had not. Going to unassign myself, feel free to take this on!

kawaguchi1102:
@mattxwang 
Thanks for the advice!

It seems that the problem has not been resolved even in the latest version.  
So I will work on this issue.

PR #4783, opened 2020-05-17
Fix TypeError for shorthand animation in time-min-milliseconds

jeddy3:

Refs #4782

The TypeError is only triggered when the new `ignore: ["delay"]` option is used (which was added in https://github.com/stylelint/stylelint/pull/4743). So the actual problem in #4782 may be an unrelated regression.

---

We'll need to [refactor the code](https://github.com/stylelint/stylelint/pull/4743#pullrequestreview-406653509) for `time-min-milliseconds` to address https://github.com/stylelint/stylelint/issues/4751 at some point.

### DIFF
--- a/lib/rules/time-min-milliseconds/__tests__/index.js
+++ b/lib/rules/time-min-milliseconds/__tests__/index.js
@@ -10,6 +10,15 @@ testRule({
 
 	accept: [
 		{
+			code: 'a { animation: none; }',
+		},
+		{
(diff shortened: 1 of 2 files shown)


jeddy3:
> - Fixed: `time-min-milliseconds` TypeError for `ignore: ["delay"]` and shorthand animation ([#4783](https://github.com/stylelint/stylelint/pull/4783)).

PR #4796, opened 2020-05-21
Add support for browser bundle

m-allanson:
# Overview

This draft PR demonstrates how stylelint can publish a browser compatible bundle.

It:

- adds a new entrypoint _browser.js_
- adds a command to create a `stylelint-esm.js` bundle
- adds a command to create separate bundles for each of stylelint's [built-in syntaxes](./lib/syntaxes/)
- modifies the API for the `customSyntax` option

This is a draft PR. You can use it to get an overview of how this functionality has been implemented. There may still be some rough edges.

Link to live demo (view source to see how it works): https://stylelint-browser-bundle.netlify.app

Refs: https://github.com/stylelint/stylelint/issues/3935

## Tasks

View the [Next Steps](#next-steps) section to see completed and remaining work.

## Example

Lint using the SCSS syntax:

```html
<html>
  <head>
    <title>demo demo demo</title>
    <script type="module">
      import sl from "./stylelint-esm.js";
      import syntaxScss from "./syntax-scss.js";

      async function run() {
        const options = {
          code: `a.#{var} { color: #fff; }`,
          config: {
            rules: {
              "color-hex-length": "long"
            }
          },
          customSyntax: syntaxScss
        };

        const result = await sl.lint(options);
        console.log(result);
      }

      run();
    </script>
  </head>
  <body></body>
</html>
```

Note that the `syntax` option is not supported by this bundle. Syntaxes must be loaded externally and passed in using the `customSyntax` option. This allows syntaxes to be lazy-loaded as needed.

## customSyntax

The `customSyntax` option which previously accepted a string is now overloaded to accept a string or a syntax object. This allows custom syntaxes to be loaded externally and then passed in to stylelint (see example above).

# The problem

Stylelint was written for NodeJS using the CommonJS module format. This led to a couple of obstacles when attempting to bundle stylelint into a browser-friendly format.

## Runtime module resolution

There are parts of stylelint that depend on `require`'s module resolution algorithm to load modules at runtime.

For example, you can give stylelint a config like:

```
{
  "rules": {
    "indentation": "tab",
    "number-leading-zero": null
  }
}
```

And it will load a module that matches the string identifier `"stylelint-config-standard"`. There is no way for JS running in a browser to reliably resolve a string like `"stylelint-config-standard"` to the appropriate module.

## Filesystem

Other parts of stylelint expect a filesystem to exist, to support things like caching results. There is no filesystem available when running in the browser.

# Approach

To run stylelint in a browser, it needs to be converted to a different (browser-friendly) module format and it needs to run without access to a filesytem.

> Note: there may be other approaches to doing this. I'd love to hear about them if so.

I aimed to create a new API that accepts a code string with a config object, and returns a stylelint result.

This would bypass stylelint's caching and loading functionality, avoiding most of stylelint's filesystem IO. This subset of stylelint should be much easier for a bundler to package up.

## Refactoring

This is where most of the work was centered - separating the parts of stylelint that rely on filesystem access (caching, loading and extending configs, loading plugins, etc) from parts that do not (linting a code string with a config object)

Here's a diagram showing the dependencies in stylelint's `lib` directory, for the currently released version of stylelint:

stylelint-dep-chart-master

With the changes on this branch, the updated dependency diagram looks like this (new files highlighted purple):

stylelint-dep-chart-browser

_browser.js_ is the new entry point. The remaining files are all extracted from their parent:

| Source module                                                                                                                    |     | Function extracted to                                                                                                                       |
| -------------------------------------------------------------------------------------------------------------------------------- | --- | ------------------------------------------------------------------------------------------------------------------------------------------- |
| ~standalone.js~                                                                                                                    |     | ~getFormatterFunction.js~                                                                                                                     |
| standalone.js                                                                                                                    | --> | prepareReturnValue.js                                                                                                                       |
| lintSource.js                                                                                                                    | --> | lintPostcssResult.js                                                                                                                        |
| createStylelintResult.js                                                                                                         | --> | createPartialStylelintResult.js                                                                                                             |
| [augmentConfig.js](https://github.com/stylelint/stylelint/compare/browser-bundle?expand=1#diff-0f2e84e8ffbcfacc3d8fe6e329ee65e3) | --> | [normalizeAllRuleSettings.js](https://github.com/stylelint/stylelint/compare/browser-bundle?expand=1#diff-4d4896c18d8e5c35dfdf058a221ac67b) |

In most cases this involved copying an existing but unexported function out to a new 'child' module, then importing it back into its parent.

---

The result of these changes are that (if you squint a bit) you can see the _lib_ directory has been split into halves.

The left half is concerned with all the filesystem things like caching, configs and plugins. Let's call this 'context'. The right half is more focussed on linting a string + config. We can call this 'core'.

Here's a wiggly line to demonstrate:

stylelint-dep-chart-divider

If you gloss over the exact file names, you might get to something like this:

stylelint-core-context

In the future, it may be useful to further refactor stylelint to make the context / core divide clearer.

## Bundling

The bundling step was relatively straightforward once the refactoring had been done.

Using the new _browser.js_ file as an entrypoint, [Parcel](http://github.com/parcel-bundler/parcel) creates an ES module bundle. Each of stylelint's syntaxes also gets a separate bundle, allowing syntaxes to be loaded on demand.

References: Prettier's [browser bundle](https://prettier.io/docs/en/browser.html), ESLint's [`lintText` method](https://github.com/eslint/eslint/blob/bcafd0f8508e19ab8087a35fac7b97fc4295df3e/lib/eslint/eslint.js#L519-L556). Dependency diagrams created with [Dependency Cruiser](https://github.com/sverweij/dependency-cruiser). Also thanks to @jeddy3 for regularly talking through ideas, approaches and the code.

# Limitations

## Does not support `extends` or `plugins` config keys

The browser bundle does not support the `extends` or `plugins` config keys. Here's [a reminder of how these keys are processed](https://stylelint.io/user-guide/configure#extends) by stylelint:

> The value of `"extends"` is a "locater" (or an array of "locaters") that is ultimately `require()`d. It can fit whatever format works with Node's `require.resolve()` algorithm. That means a "locater" can be:
>
> - the name of a module in `node_modules` (e.g. `stylelint-config-standard`; that module's `main` file must be a valid JSON configuration)
> - an absolute path to a file (which makes sense if you're creating a JS object in a Node.js context and passing it in) with a `.js` or `.json` extension.
> - a relative path to a file with a `.js` or `.json extension`, relative to the referencing configuration (e.g. if configA has `extends: "../configB"`, we'll look for `configB` relative to configA).

There is no file system available when running in a browser, so stylelint has no way to resolve the locators used in these keys.

A follow-up to this PR could add a plugin-loading API, and would allow this limitation to be revised.

## `formatters` options is fixed to `json`

The browser bundle will always output results formatted with stylelint's JSON formatter. 

# Next Steps <a href="#next-steps" id="next-steps">#</a>

I'm going to split this PR into smaller review-friendly chunks. This has the added bonus of making it easier to track down any bugs or regressions.

Tasks:
- [x] merge separate refactoring PRs (see list below)
- [x] update customSyntax API to also accept a syntax object https://github.com/stylelint/stylelint/pull/4839
- [ ] add new browser entry point (and tests)
- [ ] add bundling script
- [ ] add a smoke test for bundled code
- [ ] output bundles as part of publish workflow
- [ ] add docs

<details open><summary>Expand this to see merged PR details</summary>

- [x] refactor `createStylelintResult` to create `createPartialStylelintResult` https://github.com/stylelint/stylelint/pull/4815
- [x] refactor `augmentConfig` to create `normalizeAllRuleSettings` https://github.com/stylelint/stylelint/pull/4810
- [x] refactor `lintSource` to create `lintPostcssResult` https://github.com/stylelint/stylelint/pull/4819
- [x] add syntax export test https://github.com/stylelint/stylelint/pull/4824
- [x] add formatter types https://github.com/stylelint/stylelint/pull/4823
- [x] refactor `standalone` to create `prepareReturnValue` https://github.com/stylelint/stylelint/pull/4832
- [x] refactor `getFormatterFunction` and `resolveCustomSyntax` out from their parent scopes https://github.com/stylelint/stylelint/pull/4837/

</details>


scripts/bundle-esm.js
```diff
@@ -1,27 +1,26 @@
 'use strict';
 
-/* eslint-disable no-console */
-/* eslint-disable guard-for-in */
-/* eslint-disable node/no-extraneous-require */
+/* eslint-disable node/no-extraneous-require -- these are all installed as part of `parcel` */
 
 const bundleReport = require('@parcel/reporter-cli/lib/bundleReport').default;
 const defaultConfigContents = require('@parcel/config-default');
+const defaultConfigFilePath = require.resolve('@parcel/config-default');
```
evilebottnawi: Just interesting, why parcel? rollup and webpack, even esbuild is better, parcel is very buggy

m-allanson: At the time I wrote this, Parcel looked like the simplest way to take CommonJS and output ESM. I didn't put much more thought into it than picking 'any' bundler to use for a proof of concept.

Stvad:
👋 thanks for working on this! I was wondering what is the status of this PR?

m-allanson:
Hi @Stvad :wave:

I lost a bit of steam with this. Though it's been on my mind again recently, as there's [talk of a new major release](https://github.com/stylelint/stylelint/issues/5205) for stylelint later this year. 

A good chunk of the work from this PR has already been merged into stylelint. The biggest missing parts are:

- the file [core-entrypoint.js](https://github.com/stylelint/stylelint/pull/4796/files#diff-e5b0ed25e7a0ae56830ae242d0545181c873d21188bb860290aa0f81a142cacc)  that exposes the new API
- a bundling step that converts from CommonJS to ES Modules, and wraps everything together into a single file

(Note there may be some smaller tasks too, I'd need to dig back into the code to confirm.) 

The bundling is where I ran out of steam. I don't think adding a bundler to stylelint is the right choice for the long-term maintenance of project. Currently you can run stylelint straight from GitHub, there's no bundling step at all. I think this approach has been a boon over the years as it's usually simpler to maintain and debug unbundled code.

However, if stylelint moves to ESM (see earlier link) then bundling could be much simpler (or unnecessary?) as the code will be in a browser-friendly module format already. No CommonJS to ESM conversion would be needed!

I think the "minimum" work needed to make this feature available would be something like:

- merge `core-entrypoint.js` (and tests) into stylelint
- wait for / work on [issue #5205](https://github.com/stylelint/stylelint/issues/5205)
- document browser support as an experimental feature, and leave bundling (if needed) to stylelint users

I'd be interested to hear your thoughts on the above, and also how you'd like to use stylelint in a browser?

Stvad:
@m-allanson thanks a lot for a detailed response!

>a bundling step that converts from CommonJS to ES Modules, and wraps everything together into a single file

I'm curious why that needs to be a part of this project itself? I imagine as long as it's bundleable in principle - people can take a dependency on it and let webpack/etc do the job in the target project? (ah, I see you not is as an option later!)

>I'd be interested to hear your thoughts on the above, 

The proposal makes sense to me!

>and also how you'd like to use stylelint in a browser?

My use-case is that I have a service that allows users to publish websites (https://roam.garden/) and one of the inputs I have is a custom CSS field.  
I want to check the CSS validity at the upload/input time vs having a website build fail later becauese webpack does not like the presented CSS 😅.  

So I don't need a linter exactly, but I need something that would allow me to check CSS for syntax errors and present results to a user on ~per line basis (via CodeMirror integration).  StyleLint seemed like a good fit.  

Upon finding out that I can't make it work - I'm trying my luck with https://github.com/csstree/validator but it seems to ignore some errors 🙁.  Would appreciate suggestions on better ways of solving this! 🙂

Mouvedia:
@m-allanson for `require.resolve` you have `import.meta.resolve`.

m-allanson:
I'm closing this because it is very outdated and unlikely to be mergeable in the near future.

PR #4797, opened 2020-05-21
Fix with workaround a TypeError thrown for "html"

### LINKED ISSUE #4793: Fix TypeError Result.apply is not a function
> Clearly describe the bug

After updating Stylint from 13.3.3 tot 13.5.0 we are having issues where we get the following error on js files.

```shell
npx stylelint fin-button.js
TypeError: result.apply is not a function
    at /Users/nickdek/Sites/test/src/node_modules/@stylelint/postcss-css-in-js/extract.js:266:20
    at Array.some (<anonymous>)
    at isStylePath (/Users/nickdek/Sites/test/src/node_modules/@stylelint/postcss-css-in-js/extract.js:257:33)
    at TaggedTemplateExpression (/Users/nickdek/Sites/test/src/node_modules/@stylelint/postcss-css-in-js/extract.js:351:8)
    at NodePath._call (/Users/nickdek/Sites/test/src/node_modules/@babel/traverse/lib/path/context.js:55:20)
    at NodePath.call (/Users/nickdek/Sites/test/src/node_modules/@babel/traverse/lib/path/context.js:42:17)
    at NodePath.visit (/Users/nickdek/Sites/test/src/node_modules/@babel/traverse/lib/path/context.js:90:31)
    at TraversalContext.visitQueue (/Users/nickdek/Sites/test/src/node_modules/@babel/traverse/lib/context.js:112:16)
    at TraversalContext.visitSingle (/Users/nickdek/Sites/test/src/node_modules/@babel/traverse/lib/context.js:84:19)
    at TraversalContext.visit (/Users/nickdek/Sites/test/src/node_modules/@babel/traverse/lib/context.js:140:19)
```

Doesn't seem to be related to any rule.

```js
import { LitElement, html, css } from 'lit-element'

class FinButton extends LitElement {
  // language=css
  static styles = css`
    :host {
      display: block;
      font-size: var(--fin-button-font-size, var(--font-size, 15px));
      text-decoration: none;
    }
  `

  render() {
    return html`
      Test
    `
  }
}

customElements.define('fin-button', FinButton)
```

```js
module.exports = {
  ignoreFiles: ['!**/*.js', '!**/*.ts', '!**/*.css', '!**/*.less', '!**/*.scss'],
  plugins: ['stylelint-no-unsupported-browser-features'],
  extends: [
    'stylelint-config-standard',
    'stylelint-config-rational-order',
    'stylelint-config-prettier',
  ],
  rules: {
    'plugin/no-unsupported-browser-features': [true],
    'selector-type-no-unknown': [
      true,
      {
        ignore: ['custom-elements'],
      },
    ],
  },
}
```

`13.5.0`

CLI with `npx stylelint fin-button.js`

Yes, it relates to CSS used in JS.

No error to be thrown just like in 13.3.3 (last working version).

The following error was thrown:

```shell
npx stylelint fin-button.js
TypeError: result.apply is not a function
    at /Users/nickdek/Sites/test/src/node_modules/@stylelint/postcss-css-in-js/extract.js:266:20
    at Array.some (<anonymous>)
    at isStylePath (/Users/nickdek/Sites/test/src/node_modules/@stylelint/postcss-css-in-js/extract.js:257:33)
    at TaggedTemplateExpression (/Users/nickdek/Sites/test/src/node_modules/@stylelint/postcss-css-in-js/extract.js:351:8)
    at NodePath._call (/Users/nickdek/Sites/test/src/node_modules/@babel/traverse/lib/path/context.js:55:20)
    at NodePath.call (/Users/nickdek/Sites/test/src/node_modules/@babel/traverse/lib/path/context.js:42:17)
    at NodePath.visit (/Users/nickdek/Sites/test/src/node_modules/@babel/traverse/lib/path/context.js:90:31)
    at TraversalContext.visitQueue (/Users/nickdek/Sites/test/src/node_modules/@babel/traverse/lib/context.js:112:16)
    at TraversalContext.visitSingle (/Users/nickdek/Sites/test/src/node_modules/@babel/traverse/lib/context.js:84:19)
    at TraversalContext.visit (/Users/nickdek/Sites/test/src/node_modules/@babel/traverse/lib/context.js:140:19)
```

jeddy3:

Closes #4793

It's not clear what's happening here, and a deeper investigation is needed. It appears the HTML syntax is somehow interfering with JavaScript files that contain the word "html".

We can workaround the issue by not lazy loading the HTML syntax ourselves and instead relying on postcss-syntax mechanism. We can do this because we are using the original postcss-html module, unlike with postcss-markdown and postcss-css-in-js where we are using our own forks.

I'll create a follow-up issue at some point to discuss what we do about postcss-syntax going forward. At the moment the syntax inferring is being handled by an unmaintained package.

### DIFF
--- a/lib/__tests__/fixtures/html.js
+++ b/lib/__tests__/fixtures/html.js
@@ -1,0 +1,5 @@
+'use strict';
+
+let html;
+
+html.replace();
--- a/lib/__tests__/standalone-syntax.test.js
+++ b/lib/__tests__/standalone-syntax.test.js
@@ -238,6 +238,19 @@ it('standalone with automatic syntax inference', () => {
 	});
 });
 
+it('standalone with js file contains html', () => {
+	return standalone({
+		files: `${fixturesPath}/html.js`,
+		config: {
+			rules: {
+				'block-no-empty': true,
+			},
+		},
+	}).then((data) => {
+		expect(data.results[0].errored).toBe(false);
+	});
+});
+
 it('standalone with postcss-safe-parser', () => {
 	return standalone({
 		files: `${fixturesPath}/syntax_error.*`,
--- a/lib/getPostcssResult.js
+++ b/lib/getPostcssResult.js
@@ -78,10 +78,15 @@ module.exports = function (stylelint, options = {}) {
 			} else if (!(options.codeProcessors && options.codeProcessors.length)) {
 				const autoSyntax = require('postcss-syntax');
 
+				// TODO: investigate why lazy import HTML syntax causes
+				// JS files with the word "html" to throw TypeError
+				// https://github.com/stylelint/stylelint/issues/4793
+				const { html, ...rest } = syntaxes;
+
 				syntax = autoSyntax({
 					css: cssSyntax(stylelint),
 					jsx: syntaxes['css-in-js'],
-					...syntaxes,
+					...rest,
 				});
 			}
 


m-allanson:
Updated changelog:

> - Fixed: workaround CSS-in-JS syntax throws a TypeError in the following conditions: 1. encounter a call or template expression named 'html' and 2. syntax loads via `autoSyntax()`.

m-allanson:
I've created a reproduction for this issue here: https://github.com/m-allanson/stylelint-issue-4793

In case anyone wants to dig into this further. The bug is caused by using a call expression (such as _html()_) or a tagged template expression (such as _html\`foo\`_).

Here's how it's happening.

It's triggered when stylelint autoloads syntaxes. When autoloading syntaxes, stylelint adds the `html` syntax to its own options. 

When running, the `css-in-js` syntax checks the global stylelint options for available syntaxes. 

The `css-in-js` syntax sees that the options specify an `html` syntax, and so (incorrectly) assumes the _html_ expression contains parseable styles. 

It then runs into some unexpected behaviour, triggering a TypeError.

The error is triggered in [the isStylePath function](https://github.com/stylelint/postcss-css-in-js/blob/7a28eabd8851114b3394b89c8d510bd1da2e1690/extract.js#L255-L275).

PR #4799, opened 2020-05-21
Write error information to STDERR

### LINKED ISSUE #4798: Fix CLI printing error information to STDOUT
> Clearly describe the bug

CLI printed error information to STDOUT. This is bad behavior for CLI.

This problem is tricky. If stylelint encounters an error while using the pipeline, the error information may not be output.

N/A

N/A

N/A

`13.5.0`

CLI
```sh
stylelint --print-config /dev/null | jq
```

N/A

The CLI is expected to print error information to standard error.

Error information is printed to standard output.

naokikimura:

Closes #4798

No, it's self-explanatory.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -534,7 +534,7 @@ module.exports = (argv) => {
  * @returns {void}
  */
 function handleError(err) {
-	console.log(err.stack); // eslint-disable-line no-console
+	console.error(err.stack); // eslint-disable-line no-console
 	const exitCode = typeof err.code === 'number' ? err.code : 1;
 
 	process.exit(exitCode); // eslint-disable-line no-process-exit
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -534,10 +534,10 @@ module.exports = (argv) => {
  * @returns {void}
  */
 function handleError(err) {
-	console.error(err.stack); // eslint-disable-line no-console
+	process.stderr.write(err.stack);
 	const exitCode = typeof err.code === 'number' ? err.code : 1;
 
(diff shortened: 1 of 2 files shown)


naokikimura:
Thank you, @malsf21 

Even if there are adverse effects from this change, there is a workaround, so the ripple effect should be small.

Redirecting the standard error output to standard output as follows gives the same behavior as before.

```sh
stylelint "**/*.css" 2>&1
```

It's hard to separate the information once it's mixed, but it's easy to mix the information that's separated.

lib/cli.js
```diff
@@ -534,10 +534,10 @@ module.exports = (argv) => {
  * @returns {void}
  */
 function handleError(err) {
-	console.log(err.stack); // eslint-disable-line no-console
+	process.stderr.write(err.stack);
 	const exitCode = typeof err.code === 'number' ? err.code : 1;
 
-	process.exit(exitCode); // eslint-disable-line no-process-exit
+	process.exitCode = exitCode;
```
m-allanson, line 540: Set an exitCode and wait for node to exit when any pending work is complete. This means we can remove another `// eslint-disable-line` comment. The node docs [recommend against using `process.exit()`](https://nodejs.org/api/process.html#process_process_exit_code) where possible.

naokikimura:
Thank you @m-allanson 

I look forward to the quick release of this fix.

m-allanson:
Updated changelog:

> - Fixed: write error information to `stderr` ([#4799](https://github.com/stylelint/stylelint/pull/4799)).

PR #4802, opened 2020-05-23
Fix error message percentage/number precision for alpha-value-notation

### LINKED ISSUE #4792: Fix error message for alpha-value-notation
> Clearly describe the bug
For whatever reason the error message for alpha-value-notation rule when set to `percentage` incorrectly expects 0.14 to be 14.000000000000002%.

alpha-value-notation set to `percentage`

```css
.testclass {
	box-shadow: 0 3px 3px -2px rgba(0, 0, 0, 0.2), 0 3px 4px 0 rgba(0, 0, 0, 0.14), 0 1px 8px 0 rgba(0, 0, 0, 0.12) !important;
}
```

```json
module.exports = {
	extends: `stylelint-config-standard`,
	rules: {
		'alpha-value-notation': `percentage`,
	},
}
```

13.5.0

CLI with `stylelint  **/*.vue **/*.css` and the stylelint v0.84.0 VS Code extension 

standard css

Error message to state "expects 0.14 to be 14%".

image

As you can see on the same line it's managed to convert the other numbers to percentages, it's only 0.14 that's wrong throughout my whole codebase.

malsf21:
Hi there,

This is a PR that aims to tackle a problem addressed in #4792, namely that with floating-point conversions, the error messages (and "fixed" outputs) have numbers that are unpleasant to look at/unreasonable. 

My preliminary shot at this (which might need some feedback) is to use Javascript's in-built [`.toPrecision()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Number/toPrecision) function. I used this instead of [`.toFixed()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) as the latter rounds down very small numbers to 0 (e.g. `0.003`, which I included as a test case), which would create an incorrect error message. I added these precision-specifies to `asNumber()` and `asPercentage()`.

I am curious if this is the correct approach to solve this problem:
* I (somewhat arbitrarily) chose a precision of `3` digits - is this an alright magic number? should this be user-configurable?
* Is using `.toPrecision()` the right way to go around this? I wasn't able to personally find any problems, but I imagine it might create some problems with too many trailing zeroes, or possibly truncating important data
* Should I be covering more edge cases (e.g. very large or very very small numbers), or is that out of scope of stylelint? Should I add special checks in the functions?

Would love to get @jeddy3's input since they implemented the rule, or @m-allanson who originally responded to the issue!

Closes #4792.

Nothing else not in this giant wall of text of a PR 😄

### DIFF
--- a/lib/rules/alpha-value-notation/__tests__/index.js
+++ b/lib/rules/alpha-value-notation/__tests__/index.js
@@ -124,6 +124,22 @@ testRule({
 				{ message: messages.expected('40%', '0.4'), line: 5, column: 29 },
 			],
 		},
+		{
+			code: 'a { opacity: 14% }',
+			fixed: 'a { opacity: 0.14 }',
+			message: messages.expected('14%', '0.14'),
+			line: 1,
+			column: 14,
+			description: 'properly deals with floating-point conversions',
+		},
+		{
+			code: 'a { opacity: 0.3% }',
+			fixed: 'a { opacity: 0.003 }',
+			message: messages.expected('0.3%', '0.003'),
+			line: 1,
+			column: 14,
+			description: 'properly deals with floating-point conversions',
+		},
 	],
 });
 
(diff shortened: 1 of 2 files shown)


lib/rules/alpha-value-notation/__tests__/index.js
```diff
@@ -271,3 +271,51 @@ testRule({
...
 	],
 });
+
+// testing that floating point-rounded numbers-to-percentages are handled properly
+testRule({
+	ruleName,
+	config: ['percentage'],
+	fix: true,
+
+	reject: [
```
jeddy3, line 281: Let's move these tests into the `testRule`s above and remove the comment in favour of adding a `description: ..` to the test case.

We only tend to create a new `testRule` when the `config` changes or the `syntax` property is used.

malsf21: Yup, sounds good! I updated the PR, moving the test cases and adding a `description` key.

m-allanson:
Updated changelog:

> - Fixed: `alpha-value-notation` number precision errors ([#4802](https://github.com/stylelint/stylelint/pull/4802))

Thanks again @malsf21 👍

PR #4803, opened 2020-05-24
Fix false positives for namespaced variables in property-no-unknown

### LINKED ISSUE #4544: Fix false positives for namespaced variables in property-no-unknown
It appears that the `property-no-unknown` rule is incorrectly flagging SCSS `@use` namespaces, when setting a variable within a namespace. This occurs only within `@mixin`s and `@function`s, it doesn't occur when reading or writing module properties in the global scope. It also only occurs while setting a variable - it doesn't occur when reading the variable.

SCSS `@use` and module namespaces are relatively new - [they're described here](https://sass-lang.com/documentation/at-rules/use#loading-members).

```scss
@use 'accoutrement/sass' as a;
@use 'sass:list';

// No stylelint error
a.$sizes: ();

@function clear-sizes() {
  // Lint error: Unexpected unknown property "a.$sizes"   property-no-unknown
  a.$sizes: ();
  @return a.$sizes; // No lint error
}
$s: clear-sizes();

@mixin dump-sizes() {
  // No stylelint error
  $sizes: a.$sizes;
  @debug '# of sizes registered: #{ list.length($sizes) }';
}
@include dump-sizes();
```

The linked file provides a simple repro case with dependencies.
[stylelint-module-property.zip](https://github.com/kristerkari/stylelint-scss/files/4037883/stylelint-module-property.zip)

property-no-unknown

```scss
@use 'accoutrement/sass' as a;

@function clear-sizes() {
  // Lint error: Unexpected unknown property "a.$sizes"   property-no-unknown
  a.$sizes: ();
  @return a.$sizes; // No error
}
```

stylelint.config.js
```js
module.exports = {
  rules: {
    'property-no-unknown': true,
  }
};
```

12.0.1

CLI: `stylelint **/*.scss`

Yes, it's related to SCSS module namespaces.

No lint error.

```shell
$ stylelint **/*.scss

src/stylelint-use-namespace.scss
 9:3  ×  Unexpected unknown property "a.$sizes"   property-no-unknown

error Command failed with exit code 2.
```

I originally posted this bug [in the stylelint-scss project](https://github.com/kristerkari/stylelint-scss/issues/426), but I was directed to post it here.

ryym:

Closes #4544

I didn't add test cases of `isStandardSyntaxDeclaration` for Sass indented syntax because the `postcss-sass` parser does not understand `@use` syntax yet (https://github.com/AleshaOleg/postcss-sass/issues/145).

### DIFF
--- a/lib/utils/isStandardSyntaxDeclaration.js
+++ b/lib/utils/isStandardSyntaxDeclaration.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const isScssVariable = require('./isScssVariable');
 const { isRoot } = require('./typeGuards');
 
 /**
@@ -29,8 +30,8 @@ module.exports = function (decl) {
 		return false;
 	}
 
-	// Sass var (e.g. $var: x), nested list (e.g. $list: (x)) or nested map (e.g. $map: (key:value))
-	if (prop[0] === '$') {
+	// SCSS var
+	if (isScssVariable(prop)) {
 		return false;
 	}
 
@@ -54,11 +55,6 @@ module.exports = function (decl) {
 	// Less &:extend
 	// @ts-ignore TODO TYPES extend does not exists
 	if (decl.extend) {
-		return false;
-	}
-
-	// Sass variables within a namespace (e.g. namespace.$var: x)
-	if (prop.includes('.$') && prop[0] !== '.' && prop[prop.length - 1] !== '$') {
 		return false;
 	}
 
--- a/lib/utils/isStandardSyntaxDeclaration.js
+++ b/lib/utils/isStandardSyntaxDeclaration.js
@@ -57,5 +57,10 @@ module.exports = function (decl) {
 		return false;
 	}
 
+	// Sass variables within a namespace (e.g. namespace.$var: x)
+	if (prop.includes('.$') && prop[0] !== '.' && prop[prop.length - 1] !== '$') {
+		return false;
+	}
+
 	return true;
 };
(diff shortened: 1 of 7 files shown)


lib/utils/isStandardSyntaxDeclaration.js
```diff
@@ -57,5 +57,10 @@ module.exports = function (decl) {
 		return false;
 	}
 
+	// Sass variables within a namespace (e.g. namespace.$var: x)
+	if (prop.includes('.$') && prop[0] !== '.' && prop[prop.length - 1] !== '$') {
+		return false;
+	}
```
ryym, line 63: I'm not sure what this function should return for invalid declarations such as `.$var` and `namespace.$`.
For now this branch excludes those invalid `prop`s.

lib/utils/isStandardSyntaxDeclaration.js
```diff
@@ -57,5 +57,10 @@ module.exports = function (decl) {
 		return false;
 	}
 
+	// Sass variables within a namespace (e.g. namespace.$var: x)
+	if (prop.includes('.$') && prop[0] !== '.' && prop[prop.length - 1] !== '$') {
```
jeddy3, line 61: Let's just go with `if (prop.includes('.$'))` as the util is more interested in excluding non-standard properties than the validity of name-spaced dollar variables.

ryym:
@ybiquitous @jeddy3 Thanks for your reviewing!
I added the `isScssVariable` function and replaced the duplicate code with it.

m-allanson:
Updated changelog:

> - Fixed: `property-no-unknown` false positives for namespaced variables ([#4803](https://github.com/stylelint/stylelint/pull/4803)).

PR #4806, opened 2020-05-27
Fix false positives for variables in font-family-no-missing-generic-family-keyword

### LINKED ISSUE #4765: Fix false positives for variables in font-family-no-missing-generic-family-keyword
> Clearly describe the bug

There are false positives in `font-family-no-missing-generic-family-keyword` if I use a variable _and a font family_.

[Demo](https://stylelint.io/demo#N4Igxg9gJgpiBcIA6A7AhgAmKjGBmEKALgLR5oC2AlgDYCe8GA5AcUwDQYBuaATgBQkyhIgEoA3KgC+IdiADOdYmgAeCcPPmzwhPFQDm67ClxIQvAK40Y8s42O5HZ1qXLV6JFBBLVNVFPok+jAoMLxUYGSUtHQkANYwdADuELxQdhgA2kSWMJw4jo7AGGYGXrwwAGIildE0VDYZmWY8AkIuomYAuhhSBY5dBX0oMlJAA).

See https://github.com/stylelint/stylelint/issues/3848#issuecomment-617288304.

```css
a {
  font-family: 'font', var(--font);
}
```

```json
{
  "rules": {
     "font-family-no-missing-generic-family-keyword": [true, 
       { "ignoreFontFamilies": ["var(--font)"] }
     ]
  }
}
```

13.3.0

Demo

No

No warnings to be flagged because the rule should ignore values that _contain_ variables.

```shell
3:16 error Unexpected missing generic font family (font-family-no-missing-generic-family-keyword)
```

srawlins:

Fixes #4765

Yes, this introduces a second parse loop through a font-family.

To avoid this, we could add an parameter to findFontFamily which
dictates whether to include variables in the result, but the function
would need to blindly include variables, without sanity checking them
(like whether a variable represents a number or something).

### DIFF
--- a/lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.js
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.js
@@ -81,6 +81,9 @@ testRule({
 		{
 			code: 'a { font: namespace.$font-family; }',
 		},
+		{
+			code: 'a { font-family: "font", var(--font); }',
+		},
 	],
 
 	reject: [
--- a/lib/rules/font-family-no-missing-generic-family-keyword/index.js
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/index.js
@@ -4,8 +4,10 @@
 
 const declarationValueIndex = require('../../utils/declarationValueIndex');
 const findFontFamily = require('../../utils/findFontFamily');
+const isVariable = require('../../utils/isVariable');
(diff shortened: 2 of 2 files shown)


m-allanson:
Updated changelog:

> - Fixed: `font-family-no-missing-generic-family-keyword` false positives for variables ([#4806](https://github.com/stylelint/stylelint/pull/4806))

PR #4815, opened 2020-06-01
Create new 'createPartialStylelintResult' module

m-allanson:
In a future PR this module will be required by the new browser.js module. This change has been extracted from: #4796

Related to issue #3935

### DIFF
--- a/lib/createStylelintResult.js
+++ b/lib/createStylelintResult.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const _ = require('lodash');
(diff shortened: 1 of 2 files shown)


lib/createStylelintResult.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-const _ = require('lodash');
+const createPartialStylelintResult = require('./createPartialStylelintResult');
```
m-allanson, line 3: I don't love this name, but this is only used internally. It should be easy enough to change later 🤷

PR #4819, opened 2020-06-04
Create new 'lintPostcssResult' module

m-allanson:
In a future PR this module will be required by the new `browser.js` module. This change has been extracted from: #4796 

Related to issue #3935

### DIFF
--- a/lib/lintSource.js
+++ b/lib/lintSource.js
@@ -1,16 +1,13 @@
 'use strict';
 
-const _ = require('lodash');
-const assignDisabledRanges = require('./assignDisabledRanges');
-const getOsEol = require('./utils/getOsEol');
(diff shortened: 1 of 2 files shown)


lib/lintSource.js
```diff
@@ -84,7 +81,7 @@ module.exports = function lintSource(stylelint, options = {}) {
 					stylelint: stylelintResult,
 				});
 
-				return lintPostcssResult(stylelint, stylelintPostcssResult, config).then(
+				return lintPostcssResult(stylelint._options, stylelintPostcssResult, config).then(
```
m-allanson, line 84: `stylelint._options` was the only part of the `stylelint` object that was being used, so I changed `lintPostcssResult` to accept the options object directly.

PR #4823, opened 2020-06-06
Add type `Formatter` for formatter functions

m-allanson:
Add a named type for formatter functions.

Related to issue #3935

This change is a small part of the changes in PR #4796. On its own this change is quite small. The new type will be used as part of a follow up PR. See [this diff for details](https://github.com/stylelint/stylelint/pull/4796/files#diff-9b6919035950137210a76ab06b9f141aR290-R293).

### DIFF
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -25,6 +25,8 @@ const writeFileAtomic = require('write-file-atomic');
 /** @typedef {import('stylelint').StylelintStandaloneOptions} StylelintOptions */
 /** @typedef {import('stylelint').StylelintStandaloneReturnValue} StylelintStandaloneReturnValue */
 /** @typedef {import('stylelint').StylelintResult} StylelintResult */
+/** @typedef {import('stylelint').Formatter} Formatter */
+/** @typedef {import('stylelint').FormatterIdentifier} FormatterIdentifier */
 
 /**
  * @param {StylelintOptions} options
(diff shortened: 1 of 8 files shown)


types/stylelint/index.d.ts
```diff
@@ -83,6 +83,13 @@ declare module 'stylelint' {
 		warn(message: string, options?: StylelintWarningOptions): void;
 	};
 
+	export type Formatter = (
+		results: Array<StylelintResult>,
+		returnValue?: StylelintStandaloneReturnValue,
```
m-allanson, line 88: This second value isn't used in any of stylelint's built-in formatters. However, it is mentioned in the docs, and there's a test that checks for it. So I've included it in the type definition here.

But maybe it's an old argument that can be removed?

lib/standalone.js
```diff
@@ -25,6 +25,8 @@ const writeFileAtomic = require('write-file-atomic');
 /** @typedef {import('stylelint').StylelintStandaloneOptions} StylelintOptions */
 /** @typedef {import('stylelint').StylelintStandaloneReturnValue} StylelintStandaloneReturnValue */
 /** @typedef {import('stylelint').StylelintResult} StylelintResult */
+/** @typedef {import('stylelint').Formatter} Formatter */
+/** @typedef {import('stylelint').FormatterIdentifier} FormatterIdentifier */
```
hudochenkov, line 29: This one looks unused.

m-allanson: Good catch, thank you 👍

PR #4824, opened 2020-06-06
Export an object from the CSS-in-JS syntax

m-allanson:

This change is a small part of the changes in PR #4796.

The CSS-in-JS syntax was previously being exported as a function instead of an object. This worked because functions can still have properties attached to them. It was a bit confusing though. 

This PR updates the syntax to export an object.

After the first commit in this PR, the test should fail like this:

```
 FAIL  lib/__tests__/syntaxes.test.js
  ● syntaxes › syntax is an object with parse and stringify keys: css-in-js

    expect(received).toMatchObject(expected)

    Matcher error: received value must be a non-null object

    Received has type:  function
    Received has value: [Function syntax]

      10 | describe('syntaxes', () => {
      11 |      it.each(inputs)('syntax is an object with parse and stringify keys: %s', (name, module) => {
    > 12 |              expect(module).toMatchObject({
         |                             ^
      13 |                      parse: expect.any(Function),
      14 |                      stringify: expect.any(Function),
      15 |              });

      at lib/__tests__/syntaxes.test.js:12:18

```

### DIFF
--- a/lib/__tests__/syntaxes.test.js
+++ b/lib/__tests__/syntaxes.test.js
@@ -1,0 +1,17 @@
+'use strict';
+
+const syntaxes = require('../syntaxes/index');
+
+const inputs = Object.keys(syntaxes).map((name) => [
+	name,
+	require(`../syntaxes/syntax-${name}.js`), // directly require modules to bypass importLazy() in syntaxes/index.js
+]);
+
+describe('syntaxes', () => {
+	it.each(inputs)('syntax is an object with parse and stringify keys: %s', (name, module) => {
+		expect(module).toMatchObject({
+			parse: expect.any(Function),
+			stringify: expect.any(Function),
+		});
+	});
+});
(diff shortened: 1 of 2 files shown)


hudochenkov:
Looks like this should be fixed in `@stylelint/postcss-css-in-js`, because it doesn't export an object.

m-allanson:
A couple of the other syntaxes do the same thing, check out:

https://github.com/stylelint/stylelint/blob/33783f14fba79210b2bdd43d35b2fe4c42b31b44/lib/syntaxes/syntax-markdown.js#L3-L8

https://github.com/stylelint/stylelint/blob/33783f14fba79210b2bdd43d35b2fe4c42b31b44/lib/syntaxes/syntax-html.js#L3-L8

I think the function export is by design, so it can be configured at require time. e.g. https://github.com/stylelint/postcss-css-in-js/tree/7a28eabd8851114b3394b89c8d510bd1da2e1690#advanced-use-cases

This seemed like the most pragmatic "fix" given that some of the other syntaxes work like this too. What do you think?

PR #4837, opened 2020-06-21
A small refactor

m-allanson:
Another refactor. This moves a couple of chunks of functionality out to separate functions. It's not particularly essential, but (imo) makes it easier to understand what's happening in each file.

This should be the last bit of refactoring from PR #4796.  The remaining follow-on PRs should result in new (or changed) functionality instead of rearranging internals.

Related to issue #3935

### DIFF
--- a/lib/getPostcssResult.js
+++ b/lib/getPostcssResult.js
@@ -5,16 +5,19 @@ const LazyResult = require('postcss/lib/lazy-result');
 const postcss = require('postcss');
 const syntaxes = require('./syntaxes');
 
+/** @typedef {import('postcss').Result} Result */
+/** @typedef {import('postcss').Syntax} Syntax */
+/** @typedef {import('stylelint').CustomSyntax} CustomSyntax */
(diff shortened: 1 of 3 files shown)


lib/getPostcssResult.js
```diff
@@ -40,25 +43,7 @@ module.exports = function (stylelint, options = {}) {
 			let syntax = null;
 
 			if (stylelint._options.customSyntax) {
-				try {
-					// TODO TYPES determine which type has customSyntax
-					const customSyntax = /** @type {any} */ require(stylelint._options.customSyntax);
```
m-allanson, line 45: This is now using the type `@typedef {import('postcss').Syntax} Syntax`

types/stylelint/index.d.ts
```diff
@@ -90,6 +90,8 @@ declare module 'stylelint' {
 
 	export type FormatterIdentifier = 'compact' | 'json' | 'string' | 'unix' | 'verbose' | Formatter;
 
+	export type CustomSyntax = string;
```
m-allanson, line 93: This type will be extended later to be `string | Syntax`

PR #4839, opened 2020-06-23
Accept syntax object in `customSyntax` option

m-allanson:
This overloads the `customSyntax` option to also accept a syntax object. The option also maintains its previous behaviour of accepting a string. This should be a fully backwards compatible change. 

The signature of this option has changed from `string` to `string | { parse: Function }`. This follows the convention of the `formatter` option, which [has similar behaviour](https://stylelint.io/user-guide/usage/options#formatter). 

https://github.com/stylelint/stylelint/issues/3935. 

This is based on a section of https://github.com/stylelint/stylelint/pull/4796. This is the only part of that PR that changes existing functionality.

The change itself is fairly straightforward, but it's probably worth detailing some of the _why_.

#### Why?

This API will be used to load non-standard CSS syntaxes (`scss`, `less`, `css-in-js` etc) when running in the browser.

#### What is a syntax?

A syntax is typically an object containing two functions, `parse` and `stringify`.  Stylelint uses a syntax to transform your CSS-like code into a structure that can be linted, and then back again.

#### Why do you need a custom syntax?

stylelint ships with syntaxes for linting the various flavours of CSS-like code that people use. These syntaxes won't be included in the default browser bundle as they are quite large - several MB for all of stylelint's included syntaxes. 

#### What's wrong with using a string, anyway?

The `customSyntax` option uses Node.JS's `require` function to load the syntax. It turns a string like `my-excellent-syntax` into a 'live' JS module. `require` is not available in the browser. 

Allowing the `customSyntax` option to accept a syntax object means the host environment (whether a browser, Node.JS or something else) can deal with _how_ to load the syntax, and then pass the object in for stylelint to use.

### DIFF
--- a/docs/user-guide/usage/cli.md
+++ b/docs/user-guide/usage/cli.md
@@ -106,6 +106,14 @@ Using `bar/mySpecialConfig.json` as config to lint all `.css` files in the `foo`
 stylelint "foo/**/*.css" --config bar/mySpecialConfig.json
 ```
 
+### Example H - using a custom syntax
+
+Recursively linting all `.css` files in the `foo` directory using a custom syntax:
+
+```shell
+stylelint "foo/**/*.css" --customSyntax path/to/my-custom-syntax.js
+```
+
 ## Exit codes
 
 The CLI can exit the process with the following exit codes:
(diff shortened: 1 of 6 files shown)


jeddy3:
> - Added: syntax object acceptance to `customSyntax` option ([#4839](https://github.com/stylelint/stylelint/pull/4839)).

PR #4848, opened 2020-06-28
Add support for descriptions in stylelint command comments

### LINKED ISSUE #4840: Add support for descriptions in stylelint command comments

Describing/justifying disable directives, so that they aren't used unless necessary, and other developers know why they were used.

See the equivalent in ESLint:
- https://github.com/eslint/rfcs/pull/33
- https://github.com/eslint/eslint/issues/11806
- https://github.com/eslint/eslint/pull/12699

Example:

```scss
a {
    // stylelint-disable-next-line declaration-no-important -- A really good reason here.
    color: red !important;
}
```

glen-84:

Closes #4840.

Let me know if there is.

### DIFF
--- a/docs/user-guide/ignore-code.md
+++ b/docs/user-guide/ignore-code.md
@@ -62,6 +62,16 @@ stylelint supports complex, overlapping disabling & enabling patterns:
 
 **Caveat:** Comments within _selector and value lists_ are currently ignored.
 
+You may also include a description at the end of the comment, after two hyphens:
+
+```css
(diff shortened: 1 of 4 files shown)


glen-84:
We decided to switch to a delimiter with a space on either side, so it should not affect rule names _containing_ double hyphens.

An extreme edge case would be a rule named just `--`, but that doesn't seem worth documenting.

jeddy3:
Changelog:

> - Added: support for descriptions in stylelint command comments ([#4848](https://github.com/stylelint/stylelint/pull/4848)).

PR #4849, opened 2020-07-01
Fix false positives for trailing combinator in selector-combinator-space-after

### LINKED ISSUE #4584: Fix false positives for trailing combinator in selector-combinator-space-after
> Clearly describe the bug

This is valid SCSS:

```scss
.foo ~,
.bar {
  .baz {
    display: none;
  }
}
```

It will compile to the following CSS:

```css
.foo ~ .baz,
.bar .baz {
  display: none;
}
```

However, stylelint is complaining about the first selector, because the _general sibling_ combinator, `~`, is not followed by a single whitespace.

`selector-combinator-space-after`

```json
{
  "rules": {
    "selector-combinator-space-after": "always"
  }
}
```

11.1.1

```
stylelint **/*.scss
```

via `@wordpress/scripts`

No.

No error to be flagged.

stylelint is complaining about the first selector, because the `~` combinator is not followed by a single whitespace:

```
Expected single space after "~"   selector-combinator-space-after
```

If I were to add a whitespace (which would look strange anyway), stylelint, of course, complains about a stray whitespace before the comma:

```scss
.foo ~ ,
.bar {
  .baz {
    display: none;
  }
}
```

```
Unexpected whitespace before ","   selector-list-comma-space-before
```

This assumes `"selector-list-comma-space-before": "never"` to be included in the rules.

----

This issue is not specific to the `~` combinator, but all of them. When linting SCSS, stylelint should allow for any combinator to be directly followed by a comma, for example, `~,`.

malsf21:
Hey there!

This is a PR that tries to fix the false positive brought up in #4584, namely that a trailing combinator (such as the general sibling combinator, `~`) is not properly handled by `selector-combinator-space-after`.

This PR implements the suggested fix, by adding a check to `isStandardSyntaxSelector` for trailing combinators. It then uses `isStandardSyntaxSelector` in `selectorCombinatorSpaceChecker`, a util used by both `selector-combinator-space-after` and `selector-combinator-space-before`. Finally, it adds the test suggested in the original issue.

Closes #4584.

I wrote a (what I think is poor) implementation of the trailing combinator check:

```js
for (let combinator of punctuationSets.nonSpaceCombinators) {
	if (selector.includes(`${combinator},`)) {
		return false;
	}
}
```

This runs in `O(n)` in the number of `nonSpaceCombinators`, so there aren't any performance benefits from using a set (and I can't imagine running `.includes()` `n` times is that speedy either).

However, I wasn't able to think of another way to do this. Is there something I'm missing? I may have not understood the "search for a trailing combinator" portion of the issue.


lib/rules/selectorCombinatorSpaceChecker.js
```diff
@@ -17,6 +18,11 @@ module.exports = function (opts) {
...
 		const selector = rule.raws.selector ? rule.raws.selector.raw : rule.selector;
 
+		// Ignore non-standard selectors
+		if (!isStandardSyntaxSelector(selector)) {
+			return;
+		}
+
 		const fixedSelector = parseSelector(selector, opts.result, rule, (selectorTree) => {
 			selectorTree.walkCombinators((node) => {
 				// Ignore spaced descendant combinator
```
jeddy3, line 28: Let's add a `!isStandardSyntaxCombinator(node)` conditional here instead.

malsf21:
Gotcha, I think that makes sense. A bit swamped with work this week but I'll see if I can make the revision!

jeddy3:
> Just to be clear, I should then revert my change to isStandardSyntaxSelector?

That's correct.

> A bit swamped with work this week but I'll see if I can make the revision!

No worries. Only do it if you've time. I'm likely to be swamped with work too going forward, but I'll try to make time for reviewing your pull requests as I enjoy it.

malsf21:
> No worries. Only do it if you've time.

Yup, will do! Aiming for a target of this weekend!

malsf21:
> I think we, therefore, need a new `isStandardSyntaxCombinator` util that will accept a combinator node.

Just to clarify @jeddy3, there's already a util called `isStandardSyntaxCombinator`, which does something else. Should I modify this one, create a new one, do some refactoring, etc?

jeddy3:
> Just to clarify @jeddy3, there's already a util called `isStandardSyntaxCombinator`, which does something else. Should I modify this one, create a new one, do some refactoring, etc? Confused on what you mean by "new util".

My bad, I didn't notice that the util already existed. We'll want to modify this one to include the new conditionals rather than create a new utility.

malsf21:
Implemented the new utility, though I do have a quick question. If I implement the check to just returning false on the first and last nodes, I get an error for this type of test case, involving the SCSS/SASS use case of the nested `>`

```scss
'a { 
  >/*comment*/a {} 
}
```

In this case, the `>` combinator is the first combinator in its container, so the new implementation of the rule would say that it's non-standard, and skip the block. But, the test case in this scenario doesn't like that, and wants a space in between `>` and `a` (ignoring the comment).

In what I've pushed to the branch, I chose to ignore cases of `>`, but I'm not sure if this is the right way to proceed. In this case, should `isStandardSyntaxCombinator` return `true` or `false` for this nested `>`?

```js
if (
  node.parent !== undefined &&
  node.parent !== null &&
  node.value !== '>'
) {
  let parent = node.parent;
  if (node === parent.first) {
    return false;
  }
  if (node === parent.last) {
    return false;
  }
}
```

m-allanson:
@malsf21 it looks like something has gone wonky with the PR, it's showing 131 changed files. Most of them are changes that have already been merged. 

I'm guessing it's caused by the way GitHub doesn't track changes on the target branch after the PR is opened. There are some suggested fixes here: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

malsf21:
> @malsf21 it looks like something has gone wonky with the PR, it's showing 131 changed files. Most of them are changes that have already been merged.
> 
> I'm guessing it's caused by the way GitHub doesn't track changes on the target branch after the PR is opened. There are some suggested fixes here: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

Hm, I tried a few but I'm not getting too much success. I may close this PR and re-open it with new changes. I think it's because I cloned into my fork's `master` to update the files, when in reality I probably should've taken a different approach. Give me a moment to test things out!

m-allanson:
> I may close this PR and re-open it with new changes. 

That's often the quickest way to fix these sorts of things 😅. 

> I think it's because I cloned into my fork's master

Feel free to clone `stylelint/stylelint` and create branches straight from this repo instead of your fork. Though I sometimes prefer working from a forked repo too. 

> Give me a moment to test things out!

Absolutely, no rush.

malsf21:
Closed in favour of #4878, which is the same solution but with a cleaner git history.

Issue #4399, opened 2019-11-06
move to TypeScript 3.7


vankop:
There is a killer feature for stylelint - now it possible to create declaration file from JS https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#--declaration-and---allowjs

Blocked by:  https://github.com/stylelint/stylelint/issues/4245

hudochenkov:
I don't think we need a separate issue for this. Upgrade would be handled as with any other dependency.

vankop:
This issue was not about upgrade at first place, It is about creating definitions =)

hudochenkov:
Then rename it and explain a little bit more in the first comment :)

jeddy3:
This is neat. It's great to see JSDocs TypeScript getting some love.

jeddy3:
With #4245 now closed, I'm removing the '"blocked" label.

hudochenkov:
Currently it's blocked by https://github.com/stylelint/stylelint/pull/4505

ybiquitous:
> Currently it's blocked by #4505

We are using TypeScript 3.7 now,

https://github.com/stylelint/stylelint/blob/c63a45fbbde88010cd43703feb6591475db213ab/package.json#L116

and we will be able to use TypeScript 3.8 soon via #4626! 😄

tyankatsu0105:
Hi.
I have a question.
I just found [types](https://github.com/stylelint/stylelint/tree/master/types).
But what for?
At least currently, the types have not been released.

hudochenkov:
@tyankatsu0105 we use TypeScript to avoid bugs in our code https://github.com/stylelint/stylelint/blob/6cfea296736d3189a60eda91bb83aece26bd51c2/package.json#L44 We don't use TypeScript language, but JSDoc, which is supported by TypeScript type checker.

tyankatsu0105:
@hudochenkov 
I see. So these types are not for creating stylelint plugin or something with stylelint but for supporting to develop stylelint self, right?

hudochenkov:
Right. But we want to expose them in future, so plugin developers could use them. That's why we have this issue.

tyankatsu0105:
>But we want to expose them in future, so plugin developers could use them

Awesome!!!
I created stylelint plugin with TypeScript.
https://github.com/tyankatsu0105/stylelint-typescript.
I was worried because @types/stylelint seemed like this wasn't already being maintained.

[...]

PR #4857, opened 2020-07-08
Replace 3rd-party type definitions

ybiquitous:

Related to #4399

This change replaces 3rd-party type definitions under the `types/` directory with the following `@types/*` npm packages.

- [@types/balanced-match](https://www.npmjs.com/package/@types/balanced-match)
- [@types/imurmurhash](https://www.npmjs.com/package/@types/imurmurhash)
- [@types/postcss-safe-parser](https://www.npmjs.com/package/@types/postcss-safe-parser)
- [@types/style-search](https://www.npmjs.com/package/@types/style-search)
- [@types/svg-tags](https://www.npmjs.com/package/@types/svg-tags)
- [@types/write-file-atomic](https://www.npmjs.com/package/@types/write-file-atomic)
- [@types/file-entry-cache](https://www.npmjs.com/package/@types/file-entry-cache) (added)

Note that this change does **not** replace all type definitions under `types/`.
We should address the remaining types via other pull requests.


lib/utils/blurFunctionArguments.js
```diff
@@ -33,8 +33,13 @@ module.exports = function (source, functionName, blurChar = '`') {
...
 			lowerCaseSource.indexOf(nameWithParen, searchStartIndex) + functionNameLength;
-		const closingParenIndex =
-			balancedMatch('(', ')', lowerCaseSource.slice(openingParenIndex)).end + openingParenIndex;
+		const parensMatch = balancedMatch('(', ')', lowerCaseSource.slice(openingParenIndex));
+
+		if (!parensMatch) {
+			throw new Error(`No parens match: "${source}"`);
+		}
+
+		const closingParenIndex = parensMatch.end + openingParenIndex;
```
ybiquitous, line 42: [note] Unless checking `!parensMatch`, the following type error occurs:

```
lib/utils/blurFunctionArguments.js:42:41 - error TS2339: Property 'end' does not exist on type 'void | Output'.
  Property 'end' does not exist on type 'void'.

42  	const closingParenIndex = parensMatch.end + openingParenIndex;
    	                                      ~~~
```

ybiquitous:
@m-allanson Thank you for the review!

We can replace it with also `@types/file-entry-cache` via 5311a55, so I would appreciate it if you could review this again. ☺️

PR #4858, opened 2020-07-08
Remove `postcss-reporter` package

ybiquitous:

Related to #2454

This change removes the [`postcss-reporter`](https://www.npmjs.com/package/postcss-reporter) package and reduces the dependencies.
As you see the changed `package-lock.json`, this package depends on:

- `chalk`
- `lodash`
- `log-symbols`
- `postcss`

See also https://github.com/stylelint/stylelint/issues/2454#issuecomment-622604821

We've used only the tiny utility function `getLocation()` in the package.
This function just returns an object with the `line` and `column` property in our project, so I think we can replace it with a simple code.

https://github.com/postcss/postcss-reporter/blob/f01a601ea2cd41d626e561969d66a765b3afcb2d/lib/util.js#L4-L13


lib/formatters/stringFormatter.js
```diff
@@ -134,14 +133,14 @@ function formatter(messages, source) {
...
+		const { line, column } = message;
 		const severity = /** @type {keyof import('log-symbols')} */ (message.severity);
 		/**
 		 * @type {[string, string, string, string, string]}
 		 */
 		const row = [
-			location.line ? location.line.toString() : '',
-			location.column ? location.column.toString() : '',
+			line ? line.toString() : '',
+			column ? column.toString() : '',
```
ybiquitous, line 143: [note] The `message` variable is a `StylelintWarning`, so checking `null` seems needless:

https://github.com/stylelint/stylelint/blob/b1187b654466bb15c2e42625de652ec86afd02c6/types/stylelint/index.d.ts#L186

```suggestion
			line.toString(),
			column.toString(),
```

But, if changing it, the following test fails:

```
 FAIL  lib/formatters/__tests__/verboseFormatter.test.js
  stringFormatter
    ✓ outputs no warnings (3 ms)
    ✓ outputs one warnings (of severity 'error') (6 ms)
    ✓ outputs 0 stdout column (1 ms)
    ✓ outputs less than 80 stdout column (1 ms)
    ✓ outputs two of the same warnings of 'error' and one of 'warning' across two files (3 ms)
    ✕ outputs lineless syntax error
    ✓ outputs one ignored file

  ● stringFormatter › outputs lineless syntax error

    TypeError: Cannot read property 'toString' of undefined

      140 | 		 */
      141 | 		const row = [
    > 142 | 			line.toString(),
          | 			     ^
      143 | 			column.toString(),
      144 | 			symbols[severity]
      145 | 				? chalk[/** @type {'blue' | 'red' | 'yellow'} */ (levelColors[severity])](symbols[severity])
```

In this case, the failed test case seems wrong (should have `lines` and `columns`?), but I'm not confident whether I should fix it: 😓 

https://github.com/stylelint/stylelint/blob/b1187b654466bb15c2e42625de652ec86afd02c6/lib/formatters/__tests__/verboseFormatter.test.js#L211-L217

ybiquitous:
@jeddy3 @m-allanson Thanks for the review. As you said, it seems good to leave it now. 👍

PR #4860, opened 2020-07-12
Fix TypeError for custom properties fallback in length-zero-no-unit

shirotech:
e.g.

```css
font: var(--Funnel-header-font, 600 16px/1.5 var(--font-header));
```

lineHeightNode is undefined, so we just need to check for it.

### DIFF
--- a/lib/rules/length-zero-no-unit/index.js
+++ b/lib/rules/length-zero-no-unit/index.js
@@ -43,10 +43,13 @@ function rule(actual, secondary, context) {
 			parsedValue.walk((node, nodeIndex) => {
 				if (decl.prop.toLowerCase() === 'font' && node.type === 'div' && node.value === '/') {
 					const lineHeightNode = parsedValue.nodes[nodeIndex + 1];
-					const lineHeightNodeValue = valueParser.stringify(lineHeightNode);
+					
+					if (lineHeightNode) {
+						const lineHeightNodeValue = valueParser.stringify(lineHeightNode);
 
-					for (let i = 0; i < lineHeightNodeValue.length; i++) {
-						ignorableIndexes[lineHeightNode.sourceIndex + i] = true;
+						for (let i = 0; i < lineHeightNodeValue.length; i++) {
+							ignorableIndexes[lineHeightNode.sourceIndex + i] = true;
+						}
 					}
 
 					return;
--- a/lib/rules/length-zero-no-unit/index.js
+++ b/lib/rules/length-zero-no-unit/index.js
@@ -40,16 +40,13 @@ function rule(actual, secondary, context) {
 			const ignorableIndexes = new Array(stringValue.length).fill(false);
 			const parsedValue = valueParser(stringValue);
 
-			parsedValue.walk((node, nodeIndex) => {
+			parsedValue.walk((node, nodeIndex, nodes) => {
 				if (decl.prop.toLowerCase() === 'font' && node.type === 'div' && node.value === '/') {
-					const lineHeightNode = parsedValue.nodes[nodeIndex + 1];
-					
-					if (lineHeightNode) {
-						const lineHeightNodeValue = valueParser.stringify(lineHeightNode);
+					const lineHeightNode = nodes[nodeIndex + 1];
+					const lineHeightNodeValue = valueParser.stringify(lineHeightNode);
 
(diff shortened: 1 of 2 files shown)


jeddy3:
@shirotech Thank you for the pull request.

Can you add a [test case](https://github.com/stylelint/stylelint/blob/master/lib/rules/length-zero-no-unit/__tests__/index.js) for this, please?

You'll also need to run `npm run format` for the linting check to pass.

You'll find more information about running tests and formatting code in the [contributing guide](https://stylelint.io/contributing#code-contributions).

shirotech:
Actually I found there is an option for `ignore: ["custom-properties"]` but it is not taking effect, the lint should skip if this option is found. I will need to look into fixing this properly and provide tests as well, thanks for looking at this.

lib/rules/length-zero-no-unit/index.js
```diff
@@ -40,9 +40,9 @@ function rule(actual, secondary, context) {
 			const ignorableIndexes = new Array(stringValue.length).fill(false);
 			const parsedValue = valueParser(stringValue);
 
-			parsedValue.walk((node, nodeIndex) => {
+			parsedValue.walk((node, nodeIndex, nodes) => {
 				if (decl.prop.toLowerCase() === 'font' && node.type === 'div' && node.value === '/') {
-					const lineHeightNode = parsedValue.nodes[nodeIndex + 1];
+					const lineHeightNode = nodes[nodeIndex + 1];
```
m-allanson, line 45: When descending into functions, the `nodes` array will be scoped to the current function args. This is the array that `nodeIndex` refers to. [See the postcss-value-parser docs for details](https://github.com/TrySound/postcss-value-parser#valueparserwalknodes-callback-bubble). 

e.g., to parse the value `font: var(--foo, normal normal 400 16px/0 cursive);`, the value parser will run two loops. 

The outer loop operates on the following nodes (equivalent to `parsedValue.nodes`):

```
[
      { type: 'word', sourceIndex: 0, value: 'font' },
      { type: 'div', sourceIndex: 4, value: ':', before: '', after: ' ' },
      {
        type: 'function',
        sourceIndex: 6,
        value: 'var',
        before: '',
        after: '',
        nodes: [
          [Object], [Object],
          [Object], [Object],
          [Object], [Object],
          [Object], [Object],
          [Object], [Object],
          [Object], [Object],
          [Object]
        ]
      }
```

The inner loop (equivalent to `parsedValue.nodes[2].nodes`) runs on the following values:

```
[
      { type: 'word', sourceIndex: 10, value: '--foo' },
      { type: 'div', sourceIndex: 15, value: ',', before: '', after: ' ' },
      { type: 'word', sourceIndex: 17, value: 'normal' },
      { type: 'space', sourceIndex: 23, value: ' ' },
      { type: 'word', sourceIndex: 24, value: 'normal' },
      { type: 'space', sourceIndex: 30, value: ' ' },
      { type: 'word', sourceIndex: 31, value: '400' },
      { type: 'space', sourceIndex: 34, value: ' ' },
      { type: 'word', sourceIndex: 35, value: '16px' },
      { type: 'div', sourceIndex: 39, value: '/', before: '', after: '' },
      { type: 'word', sourceIndex: 40, value: '0' },
      { type: 'space', sourceIndex: 41, value: ' ' },
      { type: 'word', sourceIndex: 42, value: 'cursive' }
    ]
```

Using `nodes` here means we can pick the right node with `nodes[nodeIndex+1]`.

m-allanson:
I've updated this and added a couple of tests. It should be ok for review now.

@shirotech the `ignore: ["custom-properties"]` option will ignore constructs like:

```css
a { --x: 0px; }
```

It doesn't handle var functions like:

```
font: var(--foo, normal normal 400 16px/0 cursive);
```

I'd recommend opening a separate issue to talk about the custom-properties option.

shirotech:
Thanks @m-allanson anything else do we need to get this merged? Very keen on re-enabling this rule.

jeddy3:
Changelog:

> - Fixed: `length-zero-no-unit` TypeError for custom properties fallback ([#4860](https://github.com/stylelint/stylelint/pull/4860)).

---

> We should open another issue discussing the custom properties option if it hasn't been done already.

We can wait for a user to open one as I believe this pull request resolves the issue at hand.

Issue #4855, opened 2020-07-08
Fix no files matching the pattern error square brackets and dash


reumia:
> Clearly describe the bug

Succeed cases
- stylelint "./packages/mobile/pages/[channelName]/**/*.scss"
- stylelint "./packages/mobile/pages/**/*.scss"

Failure cases
- stylelint "./packages/mobile/pages/[channelName]/shopping-stories/*.scss"
- stylelint ./packages/mobile/pages/[channelName]/shopping-stories/*.scss

> Which version of stylelint are you using?

`13.3.3`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

Usually i run stylelint with lint-staged like below,

```json
// package.json
{
  "lint-staged": {
    "*.{scss,css}": [
      "stylelint --fix",
      "git add"
    ]
  },
}
```

but i tested it common usage like i wrote in top of this document, and i got same result.

> What actually happened (e.g. what warnings or errors did you get)?

Following error happened.

```shell
Error: No files matching the pattern "./packages/mobile/pages/[channelName]/shopping-stories/*.scss" were found.
    at /Users/user/Workspace/smartstore-global/node_modules/stylelint/lib/standalone.js:211:12
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
The terminal process terminated with exit code: 1
```

jeddy3:
@reumia Thanks for the report and for using the template.

I've added this issue to https://github.com/stylelint/stylelint/issues/4521, which is a list of the problems with globs and paths that we need help to fix.

I've labelled this issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

m-allanson:
Hey @reumia, thanks for reporting this issue. Could you create a small example that reproduces the problem?

As I couldn't get one of your success cases to work. 

For me, these are the failure cases:

- stylelint `"./packages/mobile/pages/[channelName]/**/*.scss"`
- stylelint `"./packages/mobile/pages/[channelName]/shopping-stories/*.scss"`
- stylelint `./packages/mobile/pages/[channelName]/shopping-stories/*.scss`

These all fail for the same reason. They're globs mixed with directory paths that contain special characters. These special characters are valid globbing characters _and_ valid directory names. There's no simple way for stylelint to know which part of the string is a path, and which part is a glob.

[...]

PR #4863, opened 2020-07-13
Fix no files matching the pattern error square brackets and dash

reumia:

- #4855 
- https://github.com/sindresorhus/globby/issues/81

[The related issue](https://github.com/sindresorhus/globby/issues/81) has already been opened on repository of `globby`, but it's been passed few years since the issue was opened. So everybody is solving themselves as I do with this PR.

It would be better if I could change `globby` to some other packages like `fast-glob` or `node-glub` but I couldn't because I worried that i can't handle whole side-effects of that change.

I just focused the issue that i've met.
Any suggestions will be appreciated.

### DIFF
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -184,6 +184,18 @@ module.exports = function (options) {
 		fileCache.destroy();
 	}
 
+	// Escape square brackets
+	fileList = fileList.map((p) => {
+		let path = p;
+
+		if (path.match(/[[\]]/g)) {
+			path = path.split('[').join('\\[');
+			path = path.split(']').join('\\]');
+		}
+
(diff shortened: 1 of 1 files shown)


m-allanson:
Thanks for opening this @reumia. Unfortunately this change will break stylelint for people that want to use `[` or `]` as part of a glob pattern. 

It's a tricky problem, and I don't think it's possible to automatically fix this for every usecase. Therefore I'm going to close this PR for now.

I've written up [some workarounds](https://github.com/stylelint/stylelint/issues/4855#issuecomment-659424327) that may help for your case. Thanks again for reporting the issue and opening a PR :+1:

Issue #4521, opened 2020-01-08
Path/globs problems


hudochenkov:
We have bunch of input (glob and paths) related problem. I gathered them here:

* [#4193 — Allow to use full file path without converting it to glob](https://github.com/stylelint/stylelint/issues/4193)
* [#4211 — Fix file paths that contain the `+` character being ignored](https://github.com/stylelint/stylelint/issues/4211)
* [#4388 — Path with parenthesis in linux says file not found in stylelint 11.1.1](https://github.com/stylelint/stylelint/issues/4388)
* [#4515 — doesn't work with pathname contains square brackets](https://github.com/stylelint/stylelint/issues/4515)
* [#3272 — Fix negated glob patterns not being ignored](https://github.com/stylelint/stylelint/issues/3272)

Possible solution: https://github.com/stylelint/stylelint/issues/4388#issuecomment-547409590

I think if we implement mechanism similar to ESLint's we'll fix most of these errors and increase stylelint reliability.

I'm marking this high priority, because it prevent our users to use stylelint completely.

Issue #4388, opened 2019-10-29
Path with spaces in linux says file not found in stylelint 11.1.1


thegrandpoobah:
> Clearly describe the bug

I upgraded the dependencies of a project recently and the following packages got upgraded:

stylelint from 8.3.0 to 11.1.1
stylelint-webpack-plugin from 0.10.5 to 1.0.3

Other packages got upgraded, but I think these two are the only relevant ones.

Prior to the upgrade, stylelint would work fine as part of webpack's build process, but after this upgrade, I get the following error in my console whenever I invoke webpack: 

```
Error: No files matching the pattern "/home/redacted/Documents/freelance/path with spaces in it/mftk.ca/src/**/*.s?(c|a)ss" were found.
```

After a bit of headscratching I decided to create a link to this directory somewhere else on my system that doesn't have spaces in the path and run it from there (in particular the path I tried is `/home/sahab/Documents/nomorespaces`). Under the path with no spaces, stylelint works fine as part of webpack.

It looks like a regression between styleline 8.3.0 and 11.1.1

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

As part of webpack using stylelint-webpack-plugin 1.0.3

> What did you expect to happen?

Non breaking change between 8.3.0 and 11.1.1

The reason I think this is part of stylelint and not stylelint-webpack-plugin is because the code responsible for calling stylelint has not really changed in that webpack plugin.

see 0.10.5 version:

https://github.com/webpack-contrib/stylelint-webpack-plugin/blob/b244ff7f6298447eca44f502d6ae750233839ce5/lib/linter.js#L5

see 1.0.3 version:

https://github.com/webpack-contrib/stylelint-webpack-plugin/blob/master/src/linter.js#L10

Stylelint-web-pack-plugin does not do any preprocessing of the path prior to passing it to stylelint. Arguably it should escape the spaces, but again, this behaviour seems to be different now.

hudochenkov:
It's either your configuration or `stylelint-webpack-plugin` issue. Paths with spaces and globs should be in quotes:

```bash
# this won't work
stylelint path with spaces/*.css

# this will work
stylelint "path with spaces/*.css"
```

vankop:
@hudochenkov plugin uses Node.js API

thegrandpoobah:
The webpack configuration used for the webpack plugin: 
[...]

Issue #4211, opened 2019-08-16
Fix file paths that contain the `+` character being ignored


andrewthauer:
> Clearly describe the bug

File paths that contain a `+` are either not linted or will fail with an error.

If the glob pattern itself contains a `+` itself it will throw this error even if it's quoted or escaped:

```
Error: No files matching the pattern "src/app/+route" were found.
    at globby.then.filePaths (/.../node_modules/stylelint/lib/standalone.js:206:17)
    at process._tickCallback (internal/process/next_tick.js:68:7)
error Command failed with exit code 1.
```

If the glob pattern does not contain a `+` but the pattern would match paths that do contain a `+` it will not error, but it also won't lint those files either. It seems to just ignore them.

> Which rule, if any, is the bug related to?

n/a

> What CSS is needed to reproduce the bug?

n/a

> What stylelint configuration is needed to reproduce the bug?

n/a

> Which version of stylelint are you using?

`10.1.0`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

- `stylelint src/app/+route/styles.scss`
- `stylelint src/app/+route`
- `stylelint 'src/app/+route/styles.scss'`
- `stylelint src/app/\+route/styles.scss`

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

n/a

> What did you expect to happen?

It will lint file paths that contain the `+` symbol it does not throw an error and also lints files that might be in paths that contain a `+`.

> What actually happened (e.g. what warnings or errors did you get)?

Doesn't lint the files or throws an error.

Some initial investigation seems to suggest this might be an issue in `globby`. I tried updating `globby` to the latest version and it no longer threw the errors. However, it also didn't pick up actual linting errors for any files contained with paths that contain `+`. I've tried the same with other linters `TSLint` & `ESLint` and both seem to handle these paths fine. However, they also don't use `globby` either.

evilebottnawi:
I.e. new version globby is work as expected?

andrewthauer:
I tried switching to `globby@10.0.1` (latest version) and it still doesn't lint file paths containing `+`. It essentially just ignores any file paths that have a `+` in it somewhere. I can't be sure it's `globby` but feels like a possibility. I'll try and find some time to see if I can reproduce the issue with globby alone.

jeddy3:
@andrewthauer Thanks for digging deeper. I'll mark up as a bug needing help.

hudochenkov:
[...]

Issue #3272, opened 2018-04-18
Negated patterns are not ignored


matikhan:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?
A bug where upgrade to stylelint@9.2.0 is ignoring negated patterns

e.g. !../web/style/less/vendor/**/*.less

> Which rule, if any, is this issue related to?

e.g. `number-leading-zero`

> What CSS is needed to reproduce this issue?

> What stylelint configuration is needed to reproduce this issue?

```json
{
  files: [
	"../web/node/style/less/**/*.less",
	"!../web/node/style/less/vendor/**/*.less"
]
}
```

> Which version of stylelint are you using?

9.2.0

> How are you running stylelint: CLI, PostCSS plugin, Node API?

"CLI with grunt"

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?
No

> What did you expect to happen?

Ignore negated patterns

> What actually happened (e.g. what warnings or errors you are getting)?
Linting negated patterns as well.

ntwb:
Thanks, @matikhan, there's another globby issue in #3273

I've subscribed to the upstream issue and will follow up further once a fix is available.

hudochenkov:
It looks like it should be fixed in the latest `13.7.2` release. @matikhan, could you check, please?

PR #4867, opened 2020-07-16
Fix some path / glob problems

### LINKED ISSUE #4193: Allow to use full file path without converting it to glob

m-allanson:

#4521 

Also:

- #4193
- #4211 
- #4388 
- #3272 
- #4855 

It's an alternate take on this PR: https://github.com/stylelint/stylelint/pull/4863

The main behaviour change is:

- if an input path contains globbing characters, it will be checked to see if it exactly matches an existing file.
- if yes, the input path will be escaped so it gets treated as an exact path, instead of a glob. 

This is broadly the approach used in ESLint, mentioned in https://github.com/stylelint/stylelint/issues/4388#issuecomment-547409590. It closes #4193.

This PR adds passing tests for most of the other problems reported in the above issues. I wasn't able to replicate some of the reported issues. They might be platform-specific, or may have been fixed in dependencies. 

This doesn't fix https://github.com/stylelint/stylelint/issues/4855, but I'll add a comment on that issue instead.


package.json
```diff
@@ -118,6 +118,7 @@
     "cosmiconfig": "^6.0.0",
     "debug": "^4.1.1",
     "execall": "^2.0.0",
+    "fast-glob": "^3.2.4",
```
m-allanson, line 121: This is already a dependency of `globby`, so it's not really a new dependency here.

evilebottnawi: Just note - prettier and many other tools remove `globby` and use `fast-glob` directly, maybe we can do same here

m-allanson: Oh interesting! That could be a good follow-up PR. 

It would need to be a breaking change, I think.  As `globbyOptions` is a config option for stylelint. `globby` has a couple of options that aren't supported in `fast-glob`. Though I'd guess few people are using any of the globby-only option..

m-allanson:
Some of the fixture paths are invalid on Windows. Will update.

lib/__tests__/standalone-globs.test.js
```diff
@@ -0,0 +1,193 @@
...
+
+describe('standalone globbing', () => {
+	describe('paths with special characters', () => {
+		// ref https://github.com/micromatch/micromatch#matching-features
+		const fixtureDirs = [
+			`[digit]/not-digits`,
+			`with spaces`,
+			`extglob!(s)`,
+			`got!negate/negate`,
+			// `extglob+(s)`, // Note: +'s cause errors. Ignoring until it becomes a problem
```
malsf21, line 27: I'm curious, what's the error here? Should we document that there's a problem with the `+` character?

m-allanson: Ah, I should have made a note of it. The `+` seemed to be escaped differently for this case, causing the expected match to fail. 

I can't remember the specific details, but I think I tracked it down to somewhere in one of globby's dependencies. I thought I'd read an issue that described it, but now I can't find it 🤷 

It seemed like enough of an edge-case that we can skip it for now, and investigate later if it becomes a problem.

PR #4871, opened 2020-07-21
Add type definition for `postcss-syntax`

ybiquitous:

This is a part of #4399.

This change adds a type definition for the `postcss-syntax` package.
To find the definition easily, this extracts a new file from the `types/postcss/index.d.ts` file.

Note that there is no `@types/postcss-syntax` package.
(see <https://www.npmjs.com/search?q=%40types%2Fpostcss-syntax>)

Here is a main source of the package:
https://github.com/gucong3000/postcss-syntax/blob/v0.36.2/index.js

Also, `postcss-syntax` is used only in `lib/getPostcssResult.js`:

https://github.com/stylelint/stylelint/blob/cecacb9d4fe3b5bae5f264efa6e60c414a0f4946/lib/getPostcssResult.js#L64-L75

### DIFF
--- a/types/postcss-syntax/index.d.ts
+++ b/types/postcss-syntax/index.d.ts
@@ -1,0 +1,7 @@
+declare module 'postcss-syntax' {
+	import { Syntax } from 'postcss';
+
+	function syntax(config: { [k: string]: Syntax }): Syntax;
+
+	export = syntax;
+}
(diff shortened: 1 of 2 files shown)


ybiquitous:
@m-allanson Oh, I didn't know it, thank you to let me know. 😃 

Thanks to you, I could know that `postcss-syntax` requires dynamically packages like `postcss-html`, `postcss-markdown` etc. 
Because there are no direct dependencies by `postcss-syntax`, I misunderstood it. 😅 👍

ybiquitous:
Now, I don't think we need `@types/postcss-syntax` because of less usage of it in `stylelint`. But it may be needed in the future.

At least for the PostCSS community, it is useful. I probably may create it if I have time. 😃

PR #4876, opened 2020-07-23
Add missing newline to CLI error output

### LINKED ISSUE #4862: Fix inconsistent trailing newlines in CLI errors
> Clearly describe the bug

When the CLI (`stylelint`) fails because of some error (particularly when an exception that's thrown), it prints the message with stack trace and exits the process, but the stack trace doesn't have a trailing newline, so the CLI prompt ends up on the same line as the error message.

None.

Run the CLI in a directory that doesn't have any files that match the glob, for example:

```bash
npx stylelint "**/*.sass"
```

Any one.

`13.6.1`

CLI with `npx stylelint "**/*.sass"`

No.

There should be a trailing newline after the stack trace.

There was no trailing newline stack trace. This happens because [`handleError`](https://github.com/stylelint/stylelint/blob/master/lib/cli.js#L537) in `lib/cli.js` uses `process.stderr.write` instead of `console.error` to write the stack trace. Seeing as `console.error` maps to `STDERR` by default, the latter should be used instead so that the trailing newline is appended to the message.

m-allanson:

Closes #4862

No, it's self-explanatory

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -534,7 +534,7 @@ module.exports = (argv) => {
  * @returns {void}
  */
 function handleError(err) {
-	process.stderr.write(err.stack);
+	process.stderr.write(err.stack + EOL);
 	const exitCode = typeof err.code === 'number' ? err.code : 1;
 
 	process.exitCode = exitCode;


jeddy3:
> - Fixed: inconsistent trailing newlines in CLI error output ([#4876](https://github.com/stylelint/stylelint/pull/4876)).

Issue #4854, opened 2020-07-06
Release 13.7.0


jeddy3:
- [x] stylelint release
- [x] stylelint-config-recommended update/release
- [x] stylelint-config-standard update/release
- [x] stylelint-demo update
- [x] stylelint.io update
- [x] tweet

It'll be great if we could get out a release this week. I'll try to make time for the release itself, but we need:

- a second review on https://github.com/stylelint/stylelint/pull/4842
- reviews on https://github.com/stylelint/stylelint.io/pull/159

It'll be nice to have:

- a second review on https://github.com/stylelint/stylelint/pull/4850

hudochenkov:
It doesn't seem like it's ready for release. Renaming rules causes few issues, which are not taken care of.

1. User has `unit-blacklist` rule in their config. But received warning for `unit-disallowed-list`. User is confused, because they never configured `unit-disallowed-list`.
2. User has `unit-blacklist` rule in their config.  And `/* stylelint-disable unit-blacklist */` in the code. Will be violation suppressed?
3. User finds `unit-blacklist` in their config or receives a violation. They use `stylelint@13.6.0`. They go to website and couldn't find rule descriptions. [List of rules](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules/list.md) doesn't have these rules. The only option is to know direct URL to be redirected to the new rule. Old names should be in the list of rules until they are removed.
4. Change log entry doesn't show that there are 26 rules renamed, what their old and new names. And that users are advised to update their configs, because old names will be removed.

jeddy3:
Thanks for catching these.

> User has unit-blacklist rule in their config. But received warning for unit-disallowed-list. User is confused, because they never configured unit-disallowed-list.

I can't think of a solution to this confusion. 

> User has unit-blacklist rule in their config. And /* stylelint-disable unit-blacklist */ in the code. Will be violation suppressed?

It should be, but it currently produces:

```
 9:10  ignore  Unexpected unit "px"                                      unit-disallowed-list
```

Offhand, I'm unsure where the `ignore` part is coming from.

> User finds unit-blacklist in their config or receives a violation. They use stylelint@13.6.0. They go to website and couldn't find rule descriptions. List of rules doesn't have these rules.

We could mitigate this somewhat by including the alias in each description. For example: "`at-rule-allowed-list`: Specify a list of allowed at-rules (aliased as `at-rule-whitelist`).". But that feels less than ideal.

[...]

PR #4892, opened 2020-08-08
Deprecate *-blacklist/*-requirelist/*-whitelist

kevindew:

#4854

This takes the action suggested in [@jeddy3's comment](https://github.com/stylelint/stylelint/issues/4854#issuecomment-670935654) of replacing the aliasing approach used in https://github.com/stylelint/stylelint/pull/4845 for `*-blacklist`/`*-requirelist`/`*-whitelist` rules with a deprecation approach. As suggested, this uses the copy and paste approach for rules rather than a [factory approach](https://github.com/stylelint/stylelint/commit/e93e44c938d77021eec4893fc20327337a60d49f) to reduce code duplication. This is with the expectation that these rules will be removed in a soon major release and thus won't remain duplicated for long.

As this was quite a manual task, the main issues to try catch in reviewing are copy+paste mistakes and anything that were missed. Given that this introduces a lot of copied and pasted code I've tried to make the reviewing easier by doing a bulk copy and paste in https://github.com/stylelint/stylelint/commit/4df7e92dffcfbbb6b46e4a1af1875474e742156c and then my actual edits in the later commits. When reviewing, it'd be easiest to focus attention on the changes after the first commit.

This change should mean that https://github.com/stylelint/stylelint.io/pull/159 no longer needs merging.


kevindew:
Thanks @malsf21 - I just sorted the merge conflict.

> We used to add code that show to user that some rules are deprecated: https://github.com/stylelint/stylelint/pull/2679/files#diff-29bacdfdd2b7605922ec3355625a795aR30-R35
> 
> We also have this docs about deprecation: https://stylelint.io/developer-guide/rules#deprecate-a-rule
> 
> I think we need to add these warnings. What do you think?

Ack, I missed that documentation and copied a previous deprecation strategy.

Anyway, we've got the warnings in place, see: https://github.com/stylelint/stylelint/blob/99877b61d3de3606329db6caee0672376e6e2f7c/lib/rules/at-rule-blacklist/index.js#L32-L35 as an example (assuming I've understood you correctly)

However, what seems a bit less clear to me is how I achieve the permanent GitHub link that is suggested in this PR, do you have an example? I couldn't find one. My dilemma is that once this branch is merged, the commits will be squashed so if I link directly to a commit ([example](https://github.com/stylelint/stylelint/blob/99877b61d3de3606329db6caee0672376e6e2f7c/lib/rules/at-rule-blacklist/README.md)) then it'll be part of a tree [that is not a branch of this repository](https://github.com/stylelint/stylelint/commit/99877b61d3de3606329db6caee0672376e6e2f7c) and - if there are any further rebases - potentially not part of this PR. However once this is merged the links could be updated to reference a merged commit, so I could open up a follow-up PR for that. Do let me know if you share concern on this and how you'd like me to proceed.

hudochenkov:
> Ack, I missed that documentation and copied a previous deprecation strategy.

No worries. Looks like we have to document it properly, because it's not documented.

> However, what seems a bit less clear to me is how I achieve the permanent GitHub link that is suggested in this PR, do you have an example?

It's not clear for me why it should be a link to Github, but in mentioned PR it's a lint to a our website ¯﻿\﻿_﻿(﻿ツ﻿)﻿_﻿/﻿¯

Anyway I came up with a solution — use tag view:

https://github.com/stylelint/stylelint/blob/13.6.1/lib/rules/at-rule-property-requirelist/README.md

We would need to use future version tag, which is not available now, but will be online when new version is released:

https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/at-rule-property-requirelist/README.md

jeddy3:
> It's not clear for me why it should be a link to Github, 

If I remember correctly, it's because the website isn't versioned. When we remove the rules in the next major release, that will also remove the corresponding pages from the site because we build the website from the source files. The deprecation links will be 404ed.

It's pretty clunky, but we never had time to build anything better.

> Anyway I came up with a solution — use tag view:

Sounds good to me.

@kevindew We've not deprecated any rules for a couple of years, and we're finding our way with this. Thanks for bearing with us.  I've been busy with a new contract of late, but it's a Bank Holiday this weekend, and I should have time to get a release out. Will you have time to update the URLs to use the `https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/{ruleName}/README.md` before then?

kevindew:
Thanks @hudochenkov and @jeddy3 that tag suggestion works and I've applied it in 2794d6f.

jeddy3:
@kevindew Thanks for making the change. 

@hudochenkov The deprecation warnings look good to me know. Shall we merge?

PR #4893, opened 2020-08-10
Adds ignoreSelectors:[] to max-nesting-depth

### LINKED ISSUE #4805: Add ignoreSelectors: [] to max-nesting-depth

```scss
.foo {
    opacity: 0.4;
  }
    color: red;
  }
}
```

mattxwang:
Hey there! This is a (hopefully simple) PR that addresses the option requested in #4805, namely adding `ignoreSelectors: []` to the `max-nesting-depth` rule. As suggested, I made the inputs of a string/regex type, and added a check as we traverse up the tree - very similar to the `ignoreAtRules` option.

I wrote a handful of simple test cases, but I'm not sure if they're enough - please let me know if I've missed anything (especially some edge cases)!

Closes #4805.

Seems good on my end! Let me know if I should add more unit tests, or if there are edge cases I haven't considered.

### DIFF
--- a/lib/rules/max-nesting-depth/__tests__/index.js
+++ b/lib/rules/max-nesting-depth/__tests__/index.js
@@ -219,6 +219,35 @@ testRule({
 
 testRule({
 	ruleName,
+	config: [0, { ignoreSelectors: ['a', '.b', '/\\[.*?\\]/'] }],
+	syntax: 'scss',
+
+	accept: [
+		{
+			code: '.foo { a { }}',
+			description: 'Nested, matches ignored selector string',
+		},
+		{
+			code: '.foo { &[disabled] { }}',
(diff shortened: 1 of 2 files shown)


malsf21:
> I think we'll need to check each selector in the selector list, and only ignore the whole selector only if all the parts match.

Oh that totally slipped my mind, thank you for catching that in code review! I'll update the tests and the functionality to match.

malsf21:
(haven't forgotten this, it's been a busy month. should be able to wrap this up this week!)

mattxwang:
Just realizing now that I never circled back to this. I think for now, I'll close this PR, and look to revisit it another time with a fresh branch - if someone else wants to take this up, feel free to do so!

PR #4895, opened 2020-08-12
Support multi-line disable descriptions

### LINKED ISSUE #4886: Add support for multi-line single-line disable descriptions
> Clearly describe the bug

With CSS-style multiline comments, it's possible to write disable descriptions (#4848) that span multiple lines:

```css
/* stylelint-disable declaration-no-important --
 * some long-winded description */
```

Unfortunately, this doesn't work with Sass-style single-line comments (when parsed as the SCSS syntax of course):

```scss
// stylelint-disable declaration-no-important --
// some long-winded description
```

N/A

```css
a {
  // stylelint-disable color-no-invalid-hex --
  // foo
  color: #FF;
}
```

```json
{
	"rules": {
		"color-no-invalid-hex": true
	}
}
```

Tested with 30bacfc2b6559b8428d44c01e612e343021e6be7

CLI with `stylelint test.css --config scripts/visual-config.json`

Yes, it's related to SCSS comments.

I expected the description to be ignored and the lint to be disabled.

The lint was not disabled and still produced a warning

```shell
test.scss
 4:10  ✖  Unexpected invalid hex color "#FF"   color-no-invalid-hex
```

jathak:

Fixes #4886. 

This works around postcss/postcss-scss#109 by combining adjacent `//` comments into a single node before passing it to `checkComment`.

### DIFF
--- a/lib/__tests__/disableRanges.test.js
+++ b/lib/__tests__/disableRanges.test.js
@@ -749,9 +749,9 @@ it('SCSS // line-disabling comment (with description)', () => {
 
 it('SCSS // disable next-line comment (with multi-line description)', () => {
 	const scssSource = `a {
-		// stylelint-disable-next-line declaration-no-important
-		// --
-		// Long-winded description
+    // stylelint-disable-next-line declaration-no-important
+    // --
+    // Long-winded description
     color: pink !important;
   }`;
 
@@ -798,9 +798,9 @@ it('Less // line-disabling comment (with description)', () => {
 
 it('Less // disable next-line comment (with multi-line description)', () => {
 	const lessSource = `a {
-		// stylelint-disable-next-line declaration-no-important
-		// --
-		// Long-winded description
+    // stylelint-disable-next-line declaration-no-important
+    // --
+    // Long-winded description
     color: pink !important;
   }`;
 
--- a/lib/__tests__/disableRanges.test.js
+++ b/lib/__tests__/disableRanges.test.js
@@ -747,6 +747,32 @@ it('SCSS // line-disabling comment (with description)', () => {
 		});
 });
 
+it('SCSS // disable next-line comment (with multi-line description)', () => {
+	const scssSource = `a {
+		// stylelint-disable-next-line declaration-no-important
+		// --
+		// Long-winded description
+    color: pink !important;
+  }`;
+
+	return postcss()
+		.use(assignDisabledRanges)
+		.process(scssSource, { syntax: scss, from: undefined })
+		.then((result) => {
+			expect(result.stylelint.disabledRanges).toEqual({
+				all: [],
+				'declaration-no-important': [
+					{
+						start: 5,
+						end: 5,
+						strictStart: true,
+						strictEnd: true,
+					},
+				],
+			});
+		});
+});
+
 it('Less // line-disabling comment (with description)', () => {
 	const lessSource = `a {
     color: pink !important; // stylelint-disable-line declaration-no-important -- Description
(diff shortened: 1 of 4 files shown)


lib/assignDisabledRanges.js
```diff
@@ -53,7 +53,41 @@ module.exports = function (root, result) {
...
+
+	// Work around postcss/postcss-scss#109 by merging adjacent `//` comments
+	// into a single node before passing to `checkComment`.
+
+	/** @type {PostcssComment?} */
+	let inlineEnd;
+
+	root.walkComments((/** @type {PostcssComment} */ comment) => {
+		if (inlineEnd) {
+			// Ignore comments that were already processed by grouping with
```
nex3, line 65: Unfinished sentence?

lib/assignDisabledRanges.js
```diff
@@ -53,7 +53,41 @@ module.exports = function (root, result) {
...
+			if (inlineEnd === comment) inlineEnd = null;
+		} else if (comment.inline || comment.raws.inline) {
```
nex3, line 67: Add a comment explaining why both of these need to be checked. Maybe factor this out into an `isInlineComment()` function?

lib/__tests__/disableRanges.test.js
```diff
@@ -747,6 +747,32 @@ it('SCSS // line-disabling comment (with description)', () => {
 		});
 });
 
+it('SCSS // disable next-line comment (with multi-line description)', () => {
+	const scssSource = `a {
+		// stylelint-disable-next-line declaration-no-important
+		// --
+		// Long-winded description
+    color: pink !important;
```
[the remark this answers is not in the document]
jathak: Fixed

lib/assignDisabledRanges.js
```diff
@@ -53,7 +53,41 @@ module.exports = function (root, result) {
...
 	result.stylelint.disabledRanges = disabledRanges;
-	root.walkComments(checkComment);
+
+	// Work around postcss/postcss-scss#109 by merging adjacent `//` comments
+	// into a single node before passing to `checkComment`.
+
+	/** @type {PostcssComment?} */
+	let inlineEnd;
+
+	root.walkComments((/** @type {PostcssComment} */ comment) => {
```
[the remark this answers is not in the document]
jathak: This annotation is necessary since the inferred type is the unmodified `Comment` type from PostCSS, and we want to use the modified type that includes the `inline` properties.

lib/assignDisabledRanges.js
jathak: Fixed

lib/assignDisabledRanges.js
jathak: Done

lib/assignDisabledRanges.js
```diff
@@ -53,10 +53,53 @@ module.exports = function (root, result) {
...
+				if (fullComment.source && current.source) {
+					fullComment.source.end = current.source.end;
+				}
+
+				inlineEnd = current;
+				next = current.next();
+			}
+			checkComment(fullComment);
+		} else {
+			checkComment(comment);
```
vankop, line 88: what do you think about adding second argument `checkComment(comment: Comment, text: string| string[])` this prevents of extra allocations using 
```js
fullComment.text += `\n${current.text}`
```
in checkComment check first one `COMMAND_PREFIX` found

jathak: Would putting all the texts into an array and then joining them and assigning the result to `fullComment.text` before calling `checkComment` work? Passing it separately could be confusing, since the text in the parameter and the text of the comment wouldn't always match (and it would require adding that text parameter to all the `process*Command` functions).

nex3: `+=` is pretty highly-optimized in V8... per [this benchmark page](https://web.archive.org/web/20181117202141/https://jsperf.com/join-concat/2), it seems to actually be faster than `.join()`.

vankop: lets keep `+=`, as I understand it affects only inline comments..

jathak:
Am I okay to merge this or should a core team member do so?

vankop:
@jathak yep, hope we will not forget to update changelog..

PR #4897, opened 2020-08-13
Add a reportDisables secondary option

nex3:
This allows config authors to specify that certain rules may not be disabled by authors.

### DIFF
--- a/lib/reportDisables.js
+++ b/lib/reportDisables.js
@@ -1,0 +1,79 @@
+'use strict';
+
+const _ = require('lodash');
+
+/** @typedef {import('stylelint').RangeType} RangeType */
+/** @typedef {import('stylelint').DisableReportRange} DisabledRange */
+/** @typedef {import('stylelint').StylelintDisableOptionsReport} StylelintDisableOptionsReport */
+
+/**
+ * Returns a report describing which `results` (if any) contain disabled ranges
+ * for rules that disallow disables via `reportDisables: true`.
+ *
+ * @param {import('stylelint').StylelintResult[]} results
+ * @returns {StylelintDisableOptionsReport}
+ */
+module.exports = function (results) {
+	/** @type {StylelintDisableOptionsReport} */
+	const report = [];
+
+	results.forEach((result) => {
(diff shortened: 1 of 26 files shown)


lib/reportDisables.js
```diff
@@ -0,0 +1,79 @@
...
+			return [];
+		}
+
+		Object.keys(rangeData).forEach((rule) => {
+			rangeData[rule].forEach((range) => {
+				if (!reportDisablesForRule(_.get(config, ['rules', rule], []))) return;
+
+				// If an equivalent range has already been reported, don't report this
+				// range as well.
+				const duplicate = reported.ranges.find((reportedRange) => {
```
vankop, line 48: I think this is fine to show duplicate here, because if you are using `reportDisables: true` for rule `N` you want to see report regardless to report of rule `M`

types/stylelint/index.d.ts
```diff
@@ -216,21 +216,22 @@ declare module 'stylelint' {
...
-		unusedRule: string;
+	export type DisableReportRange = {
+		rule: string;
 		start: number;
 		end?: number;
+
+		// This is for backwards-compatibility with formatters that were written
+		// when this name was used instead of `rule`. It should be avoided for new
+		// formatters.
+		unusedRule: string;
```
vankop, line 227: Need to create an issue to remove it for next major version after merging PR..

lib/reportDisables.js
nex3: Done.

types/stylelint/index.d.ts
nex3: Done: https://github.com/stylelint/stylelint/issues/4899

nex3:
Is this good to merge? Should I add a CHANGELOG entry in the PR itself?

vankop:
yep, need to add a changelog entry. Core team should handle it =)

jeddy3:
> yep, need to add a changelog entry. Core team should handle it =)

It's a clunky process, and I've created an issue (https://github.com/stylelint/stylelint/issues/4934) to discuss how we can improve managing the changelog and releases.

Issue #4881, opened 2020-07-26
Fix tests that use callbacks


ybiquitous:
> Clearly describe the bug

I found a bug in the following test code using a callback:

https://github.com/stylelint/stylelint/blob/9f93d1f25ab6c8f8142817269243848fb1bec22a/lib/utils/__tests__/isStandardSyntaxRule.test.js#L58-L62

When I add `done()` to the test as follows, this test fails.

```diff
--- a/lib/utils/__tests__/isStandardSyntaxRule.test.js
+++ b/lib/utils/__tests__/isStandardSyntaxRule.test.js
@@ -55,9 +55,10 @@ describe('isStandardSyntaxRule', () => {
 			expect(isStandardSyntaxRule(rule)).toBeFalsy();
 		});
 	});
-	it('called Less class parametric mixin', () => {
+	it('called Less class parametric mixin', (done) => {
 		lessRules('.mixin-name(@var);', (rule) => {
 			expect(isStandardSyntaxRule(rule)).toBeFalsy();
+			done();
 		});
 	});
 	it('non-outputting parametric Less class mixin definition', () => {
```

```console
$ npx jest --silent lib/utils/__tests__/isStandardSyntaxRule.test.js
 FAIL  lib/utils/__tests__/isStandardSyntaxRule.test.js (5.515 s)
  ● isStandardSyntaxRule › called Less class parametric mixin

    thrown: "Exceeded timeout of 5000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."

      56 | 		});
      57 | 	});
    > 58 | 	it('called Less class parametric mixin', (done) => {
         | 	^
      59 | 		lessRules('.mixin-name(@var);', (rule) => {
      60 | 			expect(isStandardSyntaxRule(rule)).toBeFalsy();
      61 | 			done();

      at lib/utils/__tests__/isStandardSyntaxRule.test.js:58:2
      at Object.<anonymous> (lib/utils/__tests__/isStandardSyntaxRule.test.js:7:1)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 30 passed, 31 total
Snapshots:   0 total
Time:        5.544 s, estimated 6 s
```

Because this Less node `.mixin-name(@var);` is an `AtRule` (not `Rule`), and thus `expect()` is not called in the callback passed to `lessRules()`.

When I rewrite the test without a callback as follows, the test fails still, though.
[...]

PR #4902, opened 2020-08-22
Fix `isStandardSyntaxRule.test.js` that use callbacks

ybiquitous:

This is a part of #4881.

This simply rewriting causes some test failures, but I'm not sure which they are due to the incorrectness of the tests or the implementation... 🤷 

Could you let me know if anyone knows the cause?

### DIFF
--- a/lib/utils/__tests__/isStandardSyntaxRule.test.js
+++ b/lib/utils/__tests__/isStandardSyntaxRule.test.js
@@ -4,114 +4,105 @@ const isStandardSyntaxRule = require('../isStandardSyntaxRule');
 const less = require('postcss-less');
 const postcss = require('postcss');
 
+const node = (code, parser = postcss) => parser.parse(code).first;
+const lessNode = (code) => node(code, less);
+
 describe('isStandardSyntaxRule', () => {
 	it('type', () => {
-		expect(isStandardSyntaxRule(rule('a {}'))).toBeTruthy();
+		expect(isStandardSyntaxRule(node('a {}'))).toBeTruthy();
 	});
 	it('when type selector before selector', () => {
-		expect(isStandardSyntaxRule(rule('when a {}'))).toBeTruthy();
+		expect(isStandardSyntaxRule(node('when a {}'))).toBeTruthy();
 	});
 	it('when type selector after selector', () => {
-		expect(isStandardSyntaxRule(rule('a when {}'))).toBeTruthy();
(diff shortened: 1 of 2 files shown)


jeddy3:
> but I'm not sure which they are due to the incorrectness of the tests or the implementation... 🤷

It looks like the implementation.

```js
it('called Less ID mixin', () => {
	expect(isStandardSyntaxRule(lessRule('#mixin-name;'))).toBeFalsy();
});
```

I think the first conditional in `isStandardSyntaxRule` needs to check if the node is of type `rule` and that it has selector, otherwise it's not standard.

lib/utils/__tests__/isStandardSyntaxRule.test.js
```diff
@@ -4,168 +4,105 @@ const isStandardSyntaxRule = require('../isStandardSyntaxRule');
 const less = require('postcss-less');
 const postcss = require('postcss');
 
+const node = (code, parser = postcss) => parser.parse(code).first;
+const lessNode = (code) => node(code, less);
```
ybiquitous, line 8: [note]
- I've moved these helper functions from last to first in the file to find easily them.
- I've renamed `*rules` to `*node` because some test cases pass a non `Rule` node to the `isStandardSyntaxRule` function. E.g. `lessNode('#mixin-name;')`

ybiquitous (commented):
@jeddy3 Thank you for the feedback! I've added the extra checking for `Rule.type` via 585fc4a:

```js
if (rule.type !== 'rule') {
  return false;
}
```

ybiquitous:
BTW, by adding the `Rule.type` check, some conditionals might be needless (but I'm not confident...).

See the followig coverage report (generated on my local machine):

image

ybiquitous:
I can refactor the function to improve the coverage in this PR if anyone hope. Or we can keep it as-is to suppress the expansion of this PR.

m-allanson:
> I can refactor the function to improve the coverage in this PR if anyone hope. Or we can keep it as-is to suppress the expansion of this PR.

Both choices are good. Maybe it could be done in a new PR after this is merged?

m-allanson (approved):
🎉

ybiquitous:
> Both choices are good. Maybe it could be done in a new PR after this is merged?

Yes, I will do it if accepted.

Thank you for the review! 😊

jeddy3 (approved):

PR #4904, opened 2020-08-24
Add support for stylelintIgnore package.json property

### LINKED ISSUE #3627: Add support for stylelintIgnore package.json property

To be consistent stylelint must recognize ignored files from Package.json like eslint does,  it prevents file pollution in a project, adding a file for ignoring one directory doesn't seem good to me, it is a good idea to keep both eslint and stylelint ignored file patterns and browser list supports in package.json so can be changed easily when wanted.

I forked the repository but I'm having a little bit hard time to make changes and test code but I'm sure this might work in `isPathIgnored` file.

```js
const ignoreFiles = require(path.resolve(process.cwd(),'package.json'));
  if(ignoreFiles.hasOwnProperty('stylelintignore')) {
    filePath = ignoreFiles.stylelintignore;
  }
```

In Package.json
```json
"stylelintIgnore: [
  "file1",
  "file2"
]
```

Thank you

theahmadzai:

Closes #3627

With this change, stylelint will be able to ignore files specified in package.json via property stylelintignore, the value for the property can be either string or array of files and it silently ignores if the property if it contains an invalid value type.
This change also breaks some ignore file tests but tested via CLI worked fine.

### DIFF
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -68,16 +68,24 @@ module.exports = function (options) {
 		if (readError.code !== FILE_NOT_FOUND_ERROR_CODE) throw readError;
 	}
 
-	const ignorePattern = options.ignorePattern || [];
-	const ignoreFiles = require(path.resolve(process.cwd(), 'package.json'));
-	let jsonIgnoreText = '';
+	let pkgIgnoreFiles = '';
 
-	if (Object.prototype.hasOwnProperty.call(ignoreFiles, 'stylelintignore')) {
-		jsonIgnoreText = ignoreFiles.stylelintignore;
+	const pkgRootFilePath = path.resolve(process.cwd(), 'package.json');
+	const pkgRoot = fs.readFileSync(pkgRootFilePath, 'utf-8');
+	const pkgObject = JSON.parse(pkgRoot);
+
+	if (Object.prototype.hasOwnProperty.call(pkgObject, 'stylelintignore')) {
+		pkgIgnoreFiles = pkgObject.stylelintignore;
+
+		// If the stylelintignore property has invalid value type ignore it by
+		// setting it to empty string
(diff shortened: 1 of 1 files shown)


malsf21:
Hey @theahmadzai,

Looks like it's failing all of our testing actions. Have you had a chance to run `npm run test` locally and ensure that your changes aren't breaking tests, or to adjust the tests to account for new behaviour?

theahmadzai:
@malsf21 Yes I ran tests locally and also tested the changes via CLI worked as expected but few tests like [these](https://github.com/stylelint/stylelint/pull/4904/checks?check_run_id=1024419029#step:6:46 ) failed where it expects a package.json file.

malsf21:
Have you added a process to check if the file exists? Took a quick peek at the changes and it doesn't seem like we do anything if `package.json` isn't there.

theahmadzai:
I have not added the tests and perhaps the tests don't check if package.json exists, It's just the EONET error when the file is not found. I will try to write tests for the PR and it would be more than good if someone writes the tests 🥇 . The behaviour is kinda self-explanatory.

malsf21:
Sorry, I could've been a bit clearer - in the implementation for your code, you should add some sort of safeguard if the file isn't found! Right now, from lines 72-74, you just assume that the file is found, then you parse it as a JSON object, etc. But what if the file isn't there? Or if there are permissions errors with opening it? Or if it's not valid JSON? You should probably add a `try/catch` or some other mechanism that ensures that we cover all of our bases!

theahmadzai:
Thanks for clarifying 👍 All green now since the package.json doesn't exist and I hope It's no problem to throw errors if there is any problem with the file or parsing object

malsf21:
Oops, looks like I've neglected this issue for a bit 😨 . I'm not super familiar with the general `package.json` internals, is someone else from the contributor team able to take a look?

PR #4907, opened 2020-08-27
Add a reportDescriptionlessDisables flag

### LINKED ISSUE #4868: Add reportDescriptionlessDisables option

Within Google, we want to require that all exceptions to stylelint rules are documented.

As I understand it, `stylelint-disable-reason` (#2227) was deprecated because there was no inherent connection between a comment existing immediately before a `/* stylelint-disable */` tag and that comment actually providing a *reason* for the disabled lint. Now that #4840 is implemented, there's a more explicit way to provide a description for a disable tag, but there's still no way to enforce it.

Judging by issues like #2292, it seems like the best way to model this would be as a flag rather than as a rule. So I propose a `reportDisablesWithoutDescriptions` flag that flags any disable tag that doesn't have an associated reason as a violation.

If this is something you'd be amenable to, my team can probably implement it.

jathak:

Fixes #4868.

This now adds the description, if any to the `DisabledRange` objects and then reports any ranges without descriptions if the flag is passed. For adding the new flag, I mostly mirrored what already exists for needless and invalid-scope disables, so let me know if I missed anything.

### DIFF
--- a/lib/__tests__/descriptionlessDisables.test.js
+++ b/lib/__tests__/descriptionlessDisables.test.js
@@ -1,0 +1,49 @@
+'use strict';
+
+const standalone = require('../standalone');
+const stripIndent = require('common-tags').stripIndent;
+
+it('descriptionlessDisables', () => {
+	const config = {
+		rules: { 'block-no-empty': true },
+	};
+
+	const css = stripIndent`
+    /* stylelint-disable -- Description */
+    a {}
+    /* stylelint-enable */
+    a {
+      b {} /* stylelint-disable-line block-no-empty -- Description */
+    }
+    /* stylelint-disable */
+    a { color: pink; }
+    /* stylelint-enable */
+    a {
+      b { color: pink; } /* stylelint-disable-line block-no-empty */
+    }
+    `;
+
+	return standalone({
+		config,
+		code: css,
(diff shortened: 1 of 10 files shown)


lib/__tests__/descriptionlessDisables.test.js
```diff
@@ -0,0 +1,49 @@
...
+    a {
+      b {} /* stylelint-disable-line block-no-empty -- Description */
+    }
+    /* stylelint-disable */
+    a { color: pink; }
+    /* stylelint-enable */
+    a {
+      b { color: pink; } /* stylelint-disable-line block-no-empty */
+    }
+    `;
```
nex3, line 24: Might be a good idea to test one with the disable on a separate line, too.

jathak: Done

jeddy3 (approved):
Thanks!

glen-84:
@jathak I think this is missing documentation [here](https://github.com/stylelint/stylelint/blob/cf2f45fd0e165efdc7e5ab5ae3d5b46025e1baf1/docs/user-guide/usage/options.md).

jeddy3:
> I think this is missing documentation here.

It's something I should have caught in review.

I'll create a new issue for it.

PR #4913, opened 2020-08-31
Avoid combining disable comments with a blank line

### LINKED ISSUE #4911: Fix double-slash disable comments followed by another comment
> Clearly describe the bug

When upgrading to 13.7.0, we found that double-slash disable comments:

`// stylelint-disable color-no-hex`

... stopped working (and started throwing errors) in 3 files out of several hundred.  Two of them were first-line, whole-file comments, the third was a disable-next-line for a specific rule.

Converting these comments to the other style fixed the issue:

`/* stylelint-disable color-no-hex */`

N/A

Since this only occurred in a tiny sample of our files, I don't know of a specific repro path.

``

13.7.0

`CLI via npx stylelint \"{apps,libs}/**/*.scss\"`

jathak:

Fixes #4911.

This checks line numbers before combining adjacent `//` comments.

### DIFF
--- a/lib/__tests__/disableRanges.test.js
+++ b/lib/__tests__/disableRanges.test.js
@@ -781,6 +781,54 @@ it('SCSS // disable next-line comment (with multi-line description)', () => {
 		});
 });
 
+it('SCSS // disable comment (with // comment after blank line)', () => {
+	const scssSource = `a {
+    // stylelint-disable declaration-no-important
+
+    // Unrelated
+    color: pink !important;
+  }`;
+
+	return postcss()
+		.use(assignDisabledRanges)
+		.process(scssSource, { syntax: scss, from: undefined })
+		.then((result) => {
+			expect(result.stylelint.disabledRanges).toEqual({
+				all: [],
+				'declaration-no-important': [
+					{
+						start: 2,
+						strictStart: true,
+					},
+				],
+			});
+		});
+});
+
+it('SCSS /* disable comment (with // comment after blank line)', () => {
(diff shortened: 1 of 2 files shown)


hudochenkov (approved):
Looks good!

Thank you for a quick fix!

jeddy3 (approved):

jeddy3:
Thanks for the fix!

Changelog entry added:

> - Fixed: double-slash disable comments when followed by another comment ([#4913](https://github.com/stylelint/stylelint/pull/4913)).

ThiefMaster:
Could this be released as 13.7.1? :)

XhmikosR:
It seems we still get failures even with this patch: https://github.com/twbs/bootstrap/runs/1110881505.

I'll try to pinpoint the cause, but going back to 13.6.1 works fine again.

stof:
This PR fixed it for cases where the `stylelint-disable` double-slash comment is on the first line of a group of commented lines, by avoiding merging it with previous lines. But Bootstrap has cases where such comment happens after other comment but without empty lines between them: https://github.com/twbs/bootstrap/blob/5c25ea647bdc463df5a322c8248460a78305a1f4/scss/bootstrap-grid.scss#L24-L27

Adding an empty line between the 2 comments should fix it with 13.7.1.

XhmikosR:
Thanks for the reply. That does seem to be the case. Although, ideally, this should be handled too because it is a regression.

stof:
@XhmikosR I created an new issue for the remaining regression

Issue #4496, opened 2019-12-28
Fix all types problems


hudochenkov:
During migration from Flow to TypeScript (https://github.com/stylelint/stylelint/issues/4245) we marked many places with `@ts-ignore` comments. We need to remove all these comments and fix types problems, which were suppressed by these comments.

ybiquitous:
Now I'm addressing a new type package `@types/postcss-less`. If you are interested, please take a look: 😉 

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46235

If `@types/postcss-less` is published, we can fix the following declaration and some `@ts-ignore` comments:

https://github.com/stylelint/stylelint/blob/cecacb9d4fe3b5bae5f264efa6e60c414a0f4946/types/global.d.ts#L5

https://github.com/stylelint/stylelint/blob/cecacb9d4fe3b5bae5f264efa6e60c414a0f4946/lib/utils/isLessVariable.js#L10-L11

ybiquitous:
Here is a list of rules using the `// @ts-nocheck` comment.

- [ ] `lib/rules/index.js`
- [ ] `lib/rules/no-descending-specificity/index.js`
- [ ] `lib/rules/no-duplicate-at-import-rules/index.js`
- [ ] `lib/rules/no-duplicate-selectors/index.js`
- [ ] `lib/rules/no-empty-first-line/index.js`
- [ ] `lib/rules/no-empty-source/index.js`
- [ ] `lib/rules/no-eol-whitespace/index.js`
- [ ] `lib/rules/no-extra-semicolons/index.js`
- [ ] `lib/rules/no-invalid-double-slash-comments/index.js`
- [ ] `lib/rules/no-invalid-position-at-import-rule/index.js`
- [ ] `lib/rules/no-irregular-whitespace/index.js`
- [ ] `lib/rules/no-missing-end-of-source-newline/index.js`
- [ ] `lib/rules/no-unknown-animations/index.js`
- [ ] `lib/rules/number-leading-zero/index.js`
- [ ] `lib/rules/number-max-precision/index.js`
- [ ] `lib/rules/number-no-trailing-zeros/index.js`
- [ ] `lib/rules/property-allowed-list/index.js`
- [ ] `lib/rules/property-case/index.js`
- [ ] `lib/rules/property-disallowed-list/index.js`
- [ ] `lib/rules/property-no-unknown/index.js`
- [ ] `lib/rules/property-no-vendor-prefix/index.js`
[...]

PR #4920, opened 2020-09-02
Add `@types/postcss-less`

ybiquitous:

This is a part of #4496 and improves type-checking about LESS. (fixing `// @ts-ignore`)

See <https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/postcss-less>

### DIFF
--- a/lib/utils/isStandardSyntaxAtRule.js
+++ b/lib/utils/isStandardSyntaxAtRule.js
@@ -3,7 +3,7 @@
 /**
  * Check whether a at-rule is standard
  *
- * @param {import('postcss').AtRule} atRule postcss at-rule node
+ * @param {import('postcss').AtRule | import('postcss-less').AtRule} atRule postcss at-rule node
  * @return {boolean} If `true`, the declaration is standard
  */
 module.exports = function (atRule) {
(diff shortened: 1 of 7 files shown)


lib/utils/isStandardSyntaxAtRule.js
```diff
@@ -13,15 +13,13 @@ module.exports = function (atRule) {
...
-	if (atRule.mixin) {
+	if ('mixin' in atRule && atRule.mixin) {
 		return false;
 	}
 
 	// Ignore Less detached ruleset `@detached-ruleset: { background: red; }; .top { @detached-ruleset(); }`
 	if (
-		// @ts-ignore TODO TYPES Is this property really exists?
-		atRule.variable ||
+		('variable' in atRule && atRule.variable) ||
```
m-allanson, line 22: Is it ok to do this instead?

```suggestion
		'variable' in atRule ||
```

Or are there times where having both checks is needed?

ybiquitous: > Is it ok to do this instead?

Currently, yes.

But I'm worried that `'variable' in atRule` would not work if the implementation of `postcss-less` would change to `atRule.variable = false or true` and the type definition would not be changed.
So, I think it safer to check `atRule.variable`.

> Or are there times where having both checks is needed?

Rather, I'd like to remove checking `'variable' in atRule` if possible, but the TypeScript checker doesn't allow it. See the following error:

```
lib/utils/isStandardSyntaxAtRule.js:22:11 - error TS2339: Property 'variable' does not exist on type 'AtRule | ImportAtRule | VariableAtRule | MixinAtRule | FunctionAtRule'.
  Property 'variable' does not exist on type 'AtRule'.

22  	(atRule.variable) ||
    	        ~~~~~~~~
```

Does anyone know how to make a successful type-check without changing the runtime code?

ybiquitous: In addition, the following code seems easier to read (but out of the scope of this PR):

```js
// Ignore Less detached ruleset `@detached-ruleset: { background: red; }; .top { @detached-ruleset(); }`
if ('variable' in atRule && atRule.variable) {
	return false;
}
if (!atRule.nodes && atRule.raws.afterName === '' && atRule.params[0] === '(') {
	return false;
}
```

m-allanson: Thanks for the explanation!

> Does anyone know how to make a successful type-check without changing the runtime code?

I do not know this :(

m-allanson (approved):
:+1:

jeddy3 (approved):
Thank you!

PR #4923, opened 2020-09-02
Removed "hasMagic" check before escaping paths.

### LINKED ISSUE #4193: Allow to use full file path without converting it to glob

theodab:

Closes #4193

Globby's "hasMagic" function, which is meant to detect if a path
contains globbing characters, notably does not detect paths that
contain matched parentheses, unless those matched parentheses
have extra globbing characters attached to them. This is a problem
because globby requires matches parentheses to be escaped, if they
are part of the literal path.
This changes the part of the code that escapes paths that literally
exist to not check "hasMagic" first, to get around that oversight.
I'm not super knowledgeable about your codebase, so if you think this solution might break something somewhere else, feel free to request changes.


m-allanson (approved):
Thanks @theodab, this looks good to me :+1:

I've added a test for your case, and adjusted the comment.

theodab:
Whoops. I just realized I shouldn't have made this PR from the master branch of my fork, for backwards compatibility reasons with old versions of Shaka Player.
Here's the new version of the PR: #4931

Issue #4728, opened 2020-04-29
Fix false positives for nested literals in unit-no-unknown


TLadd:
> Clearly describe the bug

I'm getting a `unit-no-unknown` lint error when using string interpolation to calculate an animation delay.

> Which rule, if any, is the bug related to?

`unit-no-unknown`

> What code is needed to reproduce the bug?

```js
import * as styled from "styled-components";

const StyledDiv = styled.div`
  animation-delay: ${({ index }) => `${index * 0.5}s`};
`;
```

I have a repository illustrating the issue here: https://github.com/TLadd/stylelint-no-unknown-bug. Can clone the repo and run
```shell
yarn
yarn lint:css
```

> What stylelint configuration is needed to reproduce the bug?

```json
{
  "extends": [
    "stylelint-config-recommended",
    "stylelint-config-styled-components"
  ]
}
```

> Which version of stylelint are you using?

13.3.3

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

stylelint './src/**/*.js'

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

Css-in-JS

> What did you expect to happen?

Not to receive a lint error

> What actually happened (e.g. what warnings or errors did you get)?

The following warnings were flagged:

```shell
test.css
✖  Unexpected unknown unit "s`"   unit-no-unknown
```

jeddy3:
@TLadd Thanks for the report and for making a reproducible repo.

I can also reproduce this [using the demo](https://stylelint.io/demo#N4Igxg9gJgpiBcICWBbADhATgFwAQCpcBDAZ1xOwE8AbGKXAM0whVwB0QKa6BaSdCADsYg7CQ4BuNoOmRBFXAGUqtKABEkAN1wBeciroA6KFoAG03MUGoi2JEJ6xqRSvFwASYAApguJINgAD1wAXwBKXQA+XFNPfyCCXAAGQwBWEJJTEKlBUwkQABpOSlEiQIRwEhIefx4AKxJC8CEGJABzCuALdhBMAFdacQRcLsFLSw4+62weQQgeKYBrOYB3GWHsfphukOkQkBCgA).

[...]

Issue #4140, opened 2019-06-30
Fix false positives for array literal expressions


brandonkal:
> Clearly describe the bug

Stylelint is not giving special treatment for expressions inside styled components. This means that rules that use regex to locate css attribute brackets result in parseErrors.

> Which rule, if any, is the bug related to?

There are many. For instance: 'selector-attribute-brackets-space-inside': 'never'

> What CSS is needed to reproduce the bug?

```js
const Card = styled.div`
  &${[p.grid$]} {
   padding: 24px;
  }
`
```

> What stylelint configuration is needed to reproduce the bug?

e.g.

```json
{
  "rules": {
    "selector-attribute-brackets-space-inside": "never"
  }
}
```

> Which version of stylelint are you using?

10.1.0

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

CLI

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

css-in-js

> What did you expect to happen?

Stylelint should not attempt to parse array literals inside expressions. No parse errors should occur here. Arrays inside expressions are mistaken as css attribute brackets.

> What actually happened (e.g. what warnings or errors did you get)?

parseError

brandonkal:
Here is a list of rules with false positives because Stylelint is attempting to read inside template expressions:

- [ ] selector-attribute-brackets-space-inside
- [ ] selector-attribute-operator-space-before
- [ ] selector-attribute-operator-space-after
- [ ] function-whitespace-after
- [ ] declaration-bang-space-before
- [ ] function-name-case
- [ ] no-descending-specificity

brandonkal:
I have found some more broken rules:

- [ ] selector-combinator-space-after
- [ ] selector-combinator-space-before
- [ ] selector-descendant-combinator-no-non-space
- [ ] selector-pseudo-class-parentheses-space-inside

Here you can see my config. Note the long list of rules that had to be disabled due to this issue that remains:
```js
module.exports = {
  extends: ['stylelint-config-standard', 'stylelint-config-prettier'],
  plugins: ['@brandonkal/stylelint-csstree-validator'],
  rules: {
    'comment-empty-line-before': null,
    'declaration-empty-line-before': null,
    'csstree/validator': true,
    'unit-no-unknown': [
      true,
      {
        ignoreUnits: ['emin', 'emax', 'eh', 'ew'],
      },
    ],
    // Fixes for errors in Stylelint rules
[...]

PR #4944, opened 2020-09-16
Substitute css in js

bartlangelaan:

Fixes these issues (partially):
https://github.com/stylelint/stylelint/issues/4140
https://github.com/stylelint/stylelint/issues/4728

And probabily solves a lot more issues if you use template literals often.

This adds a new config key: 'unstable_substituteCssInJs'. It changes all template literals with a dummy value (like $dummyValue1). This is compatible with a lot more rules.

However, the error messages can be very unexpected - it will for example show "$dummyValue4" in the error messages, instead of the code of the user.


jeddy3:
Thanks for working on this! I've been getting myself up to speed on https://github.com/stylelint/postcss-css-in-js/pull/46.

> This adds a new config key: 'unstable_substituteCssInJs'.

Can we pull this into a top-level `syntaxOptions` config key?

PR #4946, opened 2020-09-16
Add missing `media-feature-name-value-whitelist` to rule list

ybiquitous:
This PR adds the missing `media-feature-name-value-whitelist` rule to the rule list document.

https://github.com/stylelint/stylelint/blob/7449dc004c9e75119423cd7e4f3780ba895b5c89/lib/rules/media-feature-name-value-whitelist/README.md

No, it's self-explanatory.


vankop (approved):

malsf21 (approved):
LGTM, thanks for the catch!

PR #4950, opened 2020-09-23
Improve inline comment merging

### LINKED ISSUE #4937: Fix regression for disable commands and adjacent double-slash comments
> Clearly describe the bug

This is a follow-up of #4913 which fixed parts of the regression by avoiding to merge double-slash comments separated by empty lines.
There is still a regression when a disabling comment is in a double-shash comment which is not separated from preceding comments with a empty line, as the `stylelint-disable` is not on the at the beginning of the comment anymore after merging.

Can be any rule as it is related to disabling comments

See https://github.com/twbs/bootstrap/blob/5c25ea647bdc463df5a322c8248460a78305a1f4/scss/bootstrap-grid.scss#L24-L27 for an example

e.g. `13.7.1`

e.g. "CLI with `stylelint "**/*.css" --config myconfig.json`"

Yes, it's related to SCSS double-slash comments

rules are properly disabled and re-enabled, as done in Stylelint 13.6

Some of these control comments are ignored

jathak:

Fixes #4937.

This now only starts a merge when the comment begins with `stylelint-` and either the first line contains `--` or the second line starts with `--`.

The merge will end whenever another line starts with `stylelint-` to allow for adjacent stylelint commands.

### DIFF
--- a/lib/assignDisabledRanges.js
+++ b/lib/assignDisabledRanges.js
@@ -66,35 +66,53 @@ module.exports = function (root, result) {
 		if (inlineEnd) {
 			// Ignore comments already processed by grouping with a previous one.
 			if (inlineEnd === comment) inlineEnd = null;
-		} else if (isInlineComment(comment)) {
-			const fullComment = comment.clone();
-			let next = comment.next();
-			let lastLine = (comment.source && comment.source.end && comment.source.end.line) || 0;
 
-			while (next && next.type === 'comment') {
-				/** @type {PostcssComment} */
-				const current = next;
-
-				if (!isInlineComment(current)) break;
-
-				const currentLine = (current.source && current.source.end && current.source.end.line) || 0;
-
-				if (lastLine + 1 !== currentLine) break;
-
-				fullComment.text += `\n${current.text}`;
-
-				if (fullComment.source && current.source) {
-					fullComment.source.end = current.source.end;
-				}
-
-				inlineEnd = current;
-				next = current.next();
-				lastLine = currentLine;
-			}
-			checkComment(fullComment);
-		} else {
-			checkComment(comment);
+			return;
 		}
+
+		const next = comment.next();
+
+		// If any of these conditions are not met, do not merge comments.
+		if (
+			!(
+				isInlineComment(comment) &&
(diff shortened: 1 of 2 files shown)


nex3 (approved):

vankop (approved):

lib/assignDisabledRanges.js
```diff
@@ -66,35 +66,53 @@ module.exports = function (root, result) {
...
+		}
 
-				if (!isInlineComment(current)) break;
+		const next = comment.next();
+
+		// If any of these conditions are not met, do not merge comments.
+		if (
+			!(
+				isInlineComment(comment) &&
+				comment.text.startsWith('stylelint-') &&
```
hudochenkov, line 79: This will match commits with `stylelint-whatever`. Users could put stylelint plugins names in comments. I think we should change comments starting with `stylelint-disable` and `stylelint-enable` to eliminate problems.

Same applies for a `stylelint-` check below.

lib/assignDisabledRanges.js
```diff
@@ -126,6 +126,13 @@ module.exports = function (root, result) {
 		return comment.inline || comment.raws.inline;
 	}
 
+	/**
+	 * @param {PostcssComment} comment
+	 */
+	function isStylelintCommand(comment) {
+		return /^stylelint-(disable|enable)/.test(comment.text);
```
hudochenkov, line 133: I just remembered that we have `disableCommand` and `enableCommand` in this file. I think we should use them instead of hardcoded values. What do you think?

jathak: Yeah, that makes sense. Should I just call `startsWith` with both or should I put a compiled RegExp in a global variable and use that? (since a regex literal couldn't be used in this case)

hudochenkov: I would choose `startsWith` because it's easier to understand.

hudochenkov (approved):
Great job!

I've checked on https://github.com/twbs/bootstrap repository this changes. No more issues!

hudochenkov:
> - Fixed: regression for disable commands and adjacent double-slash comments ([#4950](https://github.com/stylelint/stylelint/pull/4950)).

PR #4957, opened 2020-09-25
Fix `isStandardSyntaxAtRule.test.js` that use callbacks

ybiquitous:

This is a part of #4881.

It seems [`postcss-sass`](https://github.com/AleshaOleg/postcss-sass) does not support `@mixin`, so maybe the following test case is incorrect:

https://github.com/stylelint/stylelint/blob/65ab12ebfa08cf92525626836bd9847c8e740da9/lib/utils/__tests__/isStandardSyntaxAtRule.test.js#L70-L80

In this PR, I've marked the case above as skipped.

### DIFF
--- a/lib/utils/__tests__/isStandardSyntaxAtRule.test.js
+++ b/lib/utils/__tests__/isStandardSyntaxAtRule.test.js
@@ -8,43 +8,43 @@ const postcssScss = require('postcss-scss');
 
 describe('isStandardSyntaxAtRule', () => {
 	it('non nested at-rules without quotes', () => {
-		expect(isStandardSyntaxAtRule(atRules('@charset UTF-8;')[0])).toBeTruthy();
+		expect(isStandardSyntaxAtRule(atRule('@charset UTF-8;'))).toBeTruthy();
 	});
 
 	it("non nested at-rules with `'` quotes", () => {
-		expect(isStandardSyntaxAtRule(atRules("@charset 'UTF-8';")[0])).toBeTruthy();
+		expect(isStandardSyntaxAtRule(atRule("@charset 'UTF-8';"))).toBeTruthy();
 	});
 
 	it('non nested at-rules with `"` quotes', () => {
-		expect(isStandardSyntaxAtRule(atRules('@charset "UTF-8";')[0])).toBeTruthy();
+		expect(isStandardSyntaxAtRule(atRule('@charset "UTF-8";'))).toBeTruthy();
 	});
 
 	it("non nested at-rules with `'` quotes and without space after name", () => {
-		expect(isStandardSyntaxAtRule(atRules("@charset'UTF-8';")[0])).toBeTruthy();
+		expect(isStandardSyntaxAtRule(atRule("@charset'UTF-8';"))).toBeTruthy();
 	});
 
 	it('non nested at-rules with `"` quotes and without space after name', () => {
-		expect(isStandardSyntaxAtRule(atRules('@charset"UTF-8";')[0])).toBeTruthy();
+		expect(isStandardSyntaxAtRule(atRule('@charset"UTF-8";'))).toBeTruthy();
 	});
 
 	it('non nested at-rules with function and without space after name', () => {
-		expect(isStandardSyntaxAtRule(atRules('@import url("fineprint.css") print;')[0])).toBeTruthy();
+		expect(isStandardSyntaxAtRule(atRule('@import url("fineprint.css") print;'))).toBeTruthy();
 	});
 
 	it('nested at-rules', () => {
-		expect(isStandardSyntaxAtRule(atRules('@media (min-width: 100px) {};')[0])).toBeTruthy();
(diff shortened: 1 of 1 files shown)


lib/utils/__tests__/isStandardSyntaxAtRule.test.js
```diff
@@ -1,136 +1,112 @@
 'use strict';
 
 const isStandardSyntaxAtRule = require('../isStandardSyntaxAtRule');
-const less = require('postcss-less');
 const postcss = require('postcss');
-const sass = require('postcss-sass');
-const scss = require('postcss-scss');
+const postcssLess = require('postcss-less');
+const postcssSass = require('postcss-sass');
+const postcssScss = require('postcss-scss');
```
ybiquitous, line 7: [note] I've renamed the variables not to hide local variables with the same name, e.g.

```js
const sass = '@mixin mixin()\n  @content';
const scss = '@mixin mixin() { @content; };';
```

lib/utils/__tests__/isStandardSyntaxAtRule.test.js
```diff
@@ -1,136 +1,112 @@
...
 	it('ignore variables', () => {
-		const less = `
-        @my-variable: 10px;
-        .top { margin-top: @my-variable; }
-    `;
-
-		lessAtRules(less, (atRule) => {
-			expect(isStandardSyntaxAtRule(atRule)).toBeFalsy();
-		});
+		const rules = lessAtRules('@my-variable: 10px; .top { margin-top: @my-variable; }');
```
ybiquitous, line 87: [note] I think the one-line is easier to read here such as `'@detached-ruleset: { background: red; }; .top { @detached-ruleset(); }'`.

lib/utils/__tests__/isStandardSyntaxAtRule.test.js
```diff
@@ -1,136 +1,112 @@
...
-		atRules('@media(min-width: 100px) {};', (atRule) => {
-			expect(isStandardSyntaxAtRule(atRule)).toBeTruthy();
-		});
+		expect(isStandardSyntaxAtRule(atRules('@media(min-width: 100px) {};')[0])).toBeTruthy();
 	});
 
-	it('ignore `@content` inside mixins newline', () => {
-		const sass = '@mixin mixin()\n  @content';
+	// eslint-disable-next-line jest/no-disabled-tests -- TODO: `postcss-sass` parser does not support `@mixin`.
+	it.skip('ignore `@content` inside mixins newline', () => {
```
ybiquitous, line 51: [note] This skip does not affect on the coverage:

---

image

ybiquitous: See also https://coveralls.io/builds/33727095/source?filename=lib/utils/isStandardSyntaxAtRule.js

lib/utils/__tests__/isStandardSyntaxAtRule.test.js
```diff
@@ -1,136 +1,112 @@
...
+const postcssLess = require('postcss-less');
+const postcssSass = require('postcss-sass');
+const postcssScss = require('postcss-scss');
 
 describe('isStandardSyntaxAtRule', () => {
 	it('non nested at-rules without quotes', () => {
-		atRules('@charset UTF-8;', (atRule) => {
-			expect(isStandardSyntaxAtRule(atRule)).toBeTruthy();
-		});
+		expect(isStandardSyntaxAtRule(atRules('@charset UTF-8;')[0])).toBeTruthy();
```
hudochenkov, line 11: What do you think about adding:

```js
function atRule(code) {
    return atRules(code)[0];
}
```

It will simplify all test cases:

```suggestion
		expect(isStandardSyntaxAtRule(atRule('@charset UTF-8;'))).toBeTruthy();
```

hudochenkov (approved):

lib/utils/__tests__/isStandardSyntaxAtRule.test.js
ybiquitous: Thank you for the feedback! Make sense. 👍 
Fixed via c88a46d2

hudochenkov (approved):

lib/utils/__tests__/isStandardSyntaxAtRule.test.js
malsf21: Has anybody left an issue on `postcss-sass` about this? Maybe we can submit a fix upstream as well.

malsf21 (approved):
LGTM 💯

lib/utils/__tests__/isStandardSyntaxAtRule.test.js
ybiquitous: I found a PR to add the support of `@mixin` at-rule, but it seems WIP (and too old). 😢 

https://github.com/AleshaOleg/postcss-sass/pull/56

PR #4961, opened 2020-09-29
Proposal: Add configuration example at end of docs

schalkneethling:
A proposal to add configuration examples to the docs for the various rules. It is not often that I need to update or change the default shareable config, but when I do, oh boy 😄 I can never remember how the various rules are configured. As such, I think it would be useful to add little examples like this one to rules that are not a simple `true|false`.

Also, I believe in this case you set the rule to `null` to disable?

### DIFF
--- a/lib/rules/selector-no-qualifying-type/README.md
+++ b/lib/rules/selector-no-qualifying-type/README.md
@@ -103,3 +103,26 @@ a#foo {
   margin: 0
 }
 ```
+
+### Configuration example
+
+If you want to allow `class` selectors, update your rule configuration as follows:
+
+```
+"rules": {
+    "selector-no-qualifying-type": [
+    true,
+    {
+      "ignore": ["class"]
+    }
+  ]
+}
+```
+
+To completely disable the rule:
+
+```
+"rules": {
+  "selector-no-qualifying-type": null
+}
+```


ybiquitous:
@schalkneethling Thank you for your contribution. Does the following document satisfy your need already?

https://github.com/stylelint/stylelint/blob/0eaf3d456cdecdd698d5e4a9edc54d205d47d907/docs/user-guide/configure.md#rules

schalkneethling:
> @schalkneethling Thank you for your contribution. Does the following document satisfy your need already?
> 
> https://github.com/stylelint/stylelint/blob/0eaf3d456cdecdd698d5e4a9edc54d205d47d907/docs/user-guide/configure.md#rules
> 
> EDIT: See also https://stylelint.io/user-guide/configure#rules

Wow! Indeed it does :) No idea how I missed that all this time. I wonder though. Perhaps one can add a link to this at the bottom of rule pages? Something like: Need help to configure rules? [Read the guide](https://stylelint.io/user-guide/configure#rules).

ybiquitous:
> Perhaps one can add a link to this at the bottom of rule pages?

Umm..., it's difficult. Instead, we can easily add such a link to the page header like this: ⬇️ 

image

---

In any case, it seems better to open a new issue and discuss the difficulty of finding the rules document there.
Can you close this PR and open an issue with your detailed description, please?

schalkneethling:
> Can you close this PR and open an issue with your detailed description, please?

Sure thing!

Issue #4887, opened 2020-07-31
Add comment-pattern


ericcornelissen:
> What is the problem you're trying to solve?

Linting is used to enforce a common style in code. I want to enforce a (somewhat) common style in comments as well. In particular I would like to enforce all comments to start with a lowercase letter.

> What solution would you like to see?

Either:

1. A specific new rule to enforce comments to start with a lowercase/uppercase letter, e.g. `comment-first-letter: "lowercase"`. Or,
2. A slightly more general rule to enforce comments to adhere to a specific style (with more options then just lowercase/uppercase first letter), e.g. `comment-style: "proper-sentence"` which would require starting with a capital letter and ending with a dot. Or,
3. A general rule that can be used to specify a regular expression that all comments must match. For example, to enforce comments to start with a lowercase letter one could configure: `comment-format: "^[a-z].*$"`

The first two are a little bit easier to use because it doesn't involve coming up with a regular expression that matches your style. The third option is very general and would be useful in many more scenarios then just mine (some of which I probably can't even think of). The advantage of the first one over the second one is that it is more straightforward and doesn't "open the floodgates" on comment styles that _should_ be supported.

I would be in support of both the first and third option, but between those two I don't know which one the maintainers would prefer.

jeddy3:
@ericcornelissen Thanks for the report, using the template and proposing three different solutions.

The third option sounds good to me, i.e. a `comment-pattern` rule.

I'll label up a needing discussion to see if there are any objections. Meanwhile, you're welcome to rustle it up [as a plugin](https://stylelint.io/developer-guide/plugins) to trial the rule in your own projects.

jeddy3:
As there are no objections to this, I'll label the rule as ready to implement:

- Name: `comment-pattern`
- Primary option: `regex`
- Secondary options: None
- Autofixable: No
- Message: "Expected comment to match specified pattern"
- Description: "Specify a pattern for comments."
- Section: "Limit language features" -> "Comment"

@ericcornelissen I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

There are [steps on how to add a new rule](https://stylelint.io/developer-guide/rules#add-a-rule) in the Developer guide.

You'll likely want to use the [`custom-property-pattern` rule](https://github.com/stylelint/stylelint/tree/master/lib/rules/custom-property-pattern) as a blueprint. You'll want to include some `// comment` tests too.

PR #4962, opened 2020-10-04
Add "comment-pattern" rule

omril321:

Closes [#4887](https://github.com/stylelint/stylelint/issues/4887)

No, it's self-explanatory.

I do want to mention that it is my first PR for an open source project. I'll appreciate any input :)

### DIFF
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -230,6 +230,7 @@ Grouped first by the following categories and then by the [_thing_](http://apps.
 
 - [`comment-word-blacklist`](../../../lib/rules/comment-word-blacklist/README.md): Specify a list of disallowed words within comments. **(deprecated)**
 - [`comment-word-disallowed-list`](../../../lib/rules/comment-word-disallowed-list/README.md): Specify a list of disallowed words within comments.
+- [`comment-pattern`](../../../lib/rules/comment-pattern/README.md): Specify a pattern for comments.
 
 ### General / Sheet
 
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -228,9 +228,9 @@ Grouped first by the following categories and then by the [_thing_](http://apps.
 
 ### Comment
 
+- [`comment-pattern`](../../../lib/rules/comment-pattern/README.md): Specify a pattern for comments.
 - [`comment-word-blacklist`](../../../lib/rules/comment-word-blacklist/README.md): Specify a list of disallowed words within comments. **(deprecated)**
 - [`comment-word-disallowed-list`](../../../lib/rules/comment-word-disallowed-list/README.md): Specify a list of disallowed words within comments.
-- [`comment-pattern`](../../../lib/rules/comment-pattern/README.md): Specify a pattern for comments.
 
 ### General / Sheet
 
--- a/lib/rules/comment-pattern/README.md
+++ b/lib/rules/comment-pattern/README.md
@@ -1,0 +1,39 @@
+# custom-property-pattern
+
+Specify a pattern for comments.
+
+<!-- prettier-ignore -->
+```css
+                  /*comment*/
+                 /* ↑ */ a { --foo-: 1px; }
+/**                 ↑
+ * The pattern of this */
+```
+
+## Options
+
+`regex`
+
+The pattern to match the comments with.
+
+Given the regex:
+
+```js
+/foo .+/
+```
+
+The following patterns are considered violations:
+
+<!-- prettier-ignore -->
+```css
+/*not starting with foo*/
+a { color: red; }
+```
+
+The following patterns are _not_ considered violations:
+
+<!-- prettier-ignore -->
+```css
+/*foo at the beginning*/
+a { color: red; }
+```
--- a/lib/rules/comment-pattern/README.md
+++ b/lib/rules/comment-pattern/README.md
@@ -18,7 +18,7 @@ The pattern to match the comments with.
 
 Given the regex:
 
-```js
+```
 /foo .+/
 ```
 
--- a/lib/rules/comment-pattern/README.md
+++ b/lib/rules/comment-pattern/README.md
@@ -4,22 +4,21 @@ Specify a pattern for comments.
 
 <!-- prettier-ignore -->
 ```css
-                  /*comment*/
-                 /* ↑ */ a { --foo-: 1px; }
-/**                 ↑
+/*  comment */
+/** ↑
  * The pattern of this */
 ```
 
 ## Options
 
-`regex`
+`regex|string`
 
-The pattern to match the comments with.
+A string will be translated into a RegExp like so `new RegExp(yourString)` — so be sure to escape properly.
 
-Given the regex:
+Given the string:
 
 ```
-/foo .+/
+"foo .+"
 ```
 
 The following patterns are considered violations:
--- a/lib/rules/comment-pattern/__tests__/index.js
+++ b/lib/rules/comment-pattern/__tests__/index.js
@@ -51,7 +51,7 @@ testRule({
 		{
 			code: '.without-comment {}',
 		},
-	]
+	],
 });
 
 testRule({
@@ -80,5 +80,5 @@ testRule({
 			description: 'checks inline scss comments',
 			message: messages.expected,
 		},
-	]
+	],
 });
--- a/lib/rules/comment-pattern/__tests__/index.js
+++ b/lib/rules/comment-pattern/__tests__/index.js
@@ -20,9 +20,6 @@ testRule({
 			\t
 				*/`,
 		},
-		{
-			code: '.without-comment {}',
-		},
 	],
 
 	reject: [
@@ -54,9 +51,6 @@ testRule({
 			\n\r
 			\t
 				*/`,
-		},
-		{
-			code: '.without-comment {}',
 		},
 	],
 
--- a/lib/rules/comment-pattern/__tests__/index.js
+++ b/lib/rules/comment-pattern/__tests__/index.js
@@ -74,23 +74,6 @@ testRule({
 
 testRule({
 	ruleName,
-	config: [true],
-
-	accept: [
-		{
-			code: '/* some comment */',
-		},
-		{
-			code: '/**/',
-		},
-		{
-			code: '.without-comment {}',
-		},
-	],
-});
-
-testRule({
-	ruleName,
 	config: [/foo-.+/],
 	syntax: 'scss',
 
--- a/lib/rules/comment-pattern/__tests__/index.js
+++ b/lib/rules/comment-pattern/__tests__/index.js
@@ -18,7 +18,7 @@ testRule({
 			\r\n
 			\n\r
 			\t
-			 */`,
+				*/`,
 		},
 		{
 			code: '.without-comment {}',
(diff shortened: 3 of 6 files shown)


docs/user-guide/rules/list.md
```diff
@@ -230,6 +230,7 @@ Grouped first by the following categories and then by the [_thing_](http://apps.
 
 - [`comment-word-blacklist`](../../../lib/rules/comment-word-blacklist/README.md): Specify a list of disallowed words within comments. **(deprecated)**
 - [`comment-word-disallowed-list`](../../../lib/rules/comment-word-disallowed-list/README.md): Specify a list of disallowed words within comments.
+- [`comment-pattern`](../../../lib/rules/comment-pattern/README.md): Specify a pattern for comments.
```
hudochenkov, line 233: Could you put it in the beginning of the list, please? We sort all rules in a section alphabetically.

lib/rules/comment-pattern/README.md
```diff
@@ -0,0 +1,39 @@
+# custom-property-pattern
+
+Specify a pattern for comments.
+
+<!-- prettier-ignore -->
+```css
+                  /*comment*/
+                 /* ↑ */ a { --foo-: 1px; }
+/**                 ↑
+ * The pattern of this */
```
hudochenkov, line 10: Simplify and leave only related CSS.

```suggestion
/*  comment */
/** ↑
 * The pattern of this */
```

lib/rules/comment-pattern/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+`regex`
+
+The pattern to match the comments with.
+
+Given the regex:
+
+```js
+/foo .+/
+```
```
hudochenkov, line 23: I would use [this](https://github.com/stylelint/stylelint/blame/924646681a355b327ccfeb6f00be4e1636333911/lib/rules/custom-property-pattern/README.md#L14-L22) and only adjusted RegEx.

If rule accept RegEx, then we allow it to be both string and RegEx, because JSON and other formats doesn't have RegEx as JavaScript does.

lib/rules/comment-pattern/__tests__/index.js
```diff
@@ -0,0 +1,84 @@
...
+			description: 'ignored inline scss comments',
+		},
+		{
+			code: '// foo-ok',
+			description: 'ignored inline scss comments',
+		},
+		{
+			code: '.without-comment { }',
+			description: 'ignores scss without comments',
+		},
```
hudochenkov, line 74: This test is not necessary, as it's exactly the same as for standard syntax.

lib/rules/comment-pattern/__tests__/index.js
```diff
@@ -0,0 +1,84 @@
...
+	syntax: 'scss',
+
+	accept: [
+		{
+			code: 'a {} // foo-ok',
+			description: 'ignored inline scss comments',
+		},
+		{
+			code: '// foo-ok',
+			description: 'ignored inline scss comments',
```
hudochenkov, line 69: I suggest remove these descriptions, because double-slash comment are not ignored.

lib/rules/comment-pattern/__tests__/index.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+
+const { messages, ruleName } = require('..');
+
+testRule({
+	ruleName,
+	config: [/foo-.+/],
```
hudochenkov, line 7: Could you also make a copy of this `testRule`, but use `'foo-.+'` string instead of RegEx. To make sure both RegEx and string option work exactly the same.

lib/rules/comment-pattern/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+const _ = require('lodash');
+const report = require('../../utils/report');
+const ruleMessages = require('../../utils/ruleMessages');
+const validateOptions = require('../../utils/validateOptions');
+
+const ruleName = 'comment-pattern';
+
+const messages = ruleMessages(ruleName, {
+	expected: 'Expected comment to match specified pattern',
```
hudochenkov, line 13: I think message would be more understandable if we show pattern in it:

```
Expected comment to match specified pattern: /foo.+/
```

What do you think, @stylelint/contributors?

lib/rules/comment-pattern/index.js
```diff
@@ -0,0 +1,47 @@
...
+function rule(pattern) {
+	return (root, result) => {
+		const validOptions = validateOptions(result, ruleName, {
+			actual: pattern,
+			possible: [_.isRegExp],
+		});
+
+		if (!validOptions) {
+			return;
+		}
```
hudochenkov, line 25: It needs to support string as I described in the other comment. You can use this code: https://github.com/stylelint/stylelint/blob/924646681a355b327ccfeb6f00be4e1636333911/lib/rules/custom-property-pattern/index.js#L19-L28

hudochenkov (changes requested):
That's a great first pull request!

Few small changes required.

CI is failing: To fix formatting errors run `npm run format`. To run all test locally use `npm test`, if will ensure all tests are passing and not lint errors.

lib/rules/comment-pattern/__tests__/index.js
omril321: Instead of copy-pasting the entire call for `testRule`, what do you think about iterating the configurations that should behave in the exact same way, and calling `testRule` for each?
Meaning:
```js
[/foo-.+/, 'foo-.+'].forEach((config) => {
	testRule({
		ruleName,
		config: [config],
                 ...
```

This way, it is clear that the two configurations behave exactly the same, and if for some reason the tests will require updates - it will be in a single place.
The downside that it could be less readable when seeing for the first time (I didn't find something similar in the existing tests)

omril321:
Thanks for the feedback! I've implemented the fixes as suggested. Please let me know if further changes are required.
By the way, as a feedback of my own, I didn't know that adding a rule which accepts `regex` as an option requires support for a string option as well. I think I got confused by [the original message here](https://github.com/stylelint/stylelint/issues/4887#issuecomment-683840236)
> ...
> Primary option: regex
> ...

I understand now why a string option needs to be supported as well, but I believe that specifying it in the requirements will make it more obvious for first-timers :)

hudochenkov:
> By the way, as a feedback of my own, I didn't know that adding a rule which accepts regex as an option requires support for a string option as well. I think I got confused by the original message here

Indeed, it was missed. I've added it to mentioned message. Thanks!

lib/rules/comment-pattern/__tests__/index.js
```diff
@@ -28,11 +28,46 @@ testRule({
...
+	],
+
+	reject: [
+		{
+			code: '/* not foo- */',
+			message: messages.expected(/foo-.+/),
+		},
+		{
+			code: '/**/',
+			message: messages.expected(/foo-.+/),
```
hudochenkov, line 70: It should match config, and in config we have string.

```suggestion
			message: messages.expected('foo-.+'),
		},
		{
			code: '/**/',
			message: messages.expected('foo-.+'),
```

hudochenkov (changes requested):
Thank for making changes. One small change to make.

lib/rules/comment-pattern/__tests__/index.js
omril321: @hudochenkov got it, fixed now. 
I thought that we wanted to show the `normalizedPattern` in the message, instead of the pattern that was supplied in the config. I've updated the implementation + tests as suggested :)

hudochenkov: > I thought that we wanted to show the normalizedPattern in the message, instead of the pattern that was supplied in the config.

I didn't thought about that :) Now I'm not sure which one is better :)

@stylelint/contributors what do you think?

ybiquitous: It seems reasonable to me to always show the same form `pattern "foo"`, not `pattern "/foo/"`.
Even in different config file formats, it seems users expect the same output of a message, for example:

```yaml
# .stylelintrc.yml
rules:
  comment-pattern: ["foo"]
```

```js
// .stylelintrc.js
module.exports = {
  rules: {
    "comment-pattern": [/foo/] // equal to ["foo"]
  }
}
```

```console
$ bin/stylelint.js a.css --config .stylelintrc.yml
a.css
 1:1  ✖  Expected comment to match specified pattern "foo"   comment-pattern

$ bin/stylelint.js a.css --config .stylelintrc.js
a.css
 1:1  ✖  Expected comment to match specified pattern "/foo/"   comment-pattern
```

omril321: @ybiquitous forgive me but I might be confused.. Are you suggesting to strip the slashes from the pattern when generating an error message? If so, how should regexes with modifiers look? e.g. `/foo-.+/i` or `/foo-.+/ig`?

ybiquitous: > Are you suggesting to strip the slashes from the pattern when generating an error message?

Yes, I thought we could keep the same appearance of an error message in any config formats.

> If so, how should regexes with modifiers look? e.g. `/foo-.+/i` or `/foo-.+/ig`?

Sorry, I missed regex modifiers. Surely your current implementaion seems better if we need to include the modifiers. 👍

lib/rules/comment-pattern/__tests__/index.js
```diff
@@ -0,0 +1,113 @@
...
+			code: `/* foo--
+			multi-line
+			\n
+			\r
+			\r\n
+			\n\r
+			\t
+				*/`,
+		},
+		{
```
jeddy3, line 23: We can remove these `.without-comment {}` tests as they are duplicated in the [basic checks](https://github.com/stylelint/stylelint/blob/8366b54f536b26c2a994865568336c577ea5625e/lib/testUtils/basicChecks.js).

lib/rules/comment-pattern/index.js
```diff
@@ -0,0 +1,48 @@
...
+
+const _ = require('lodash');
+const report = require('../../utils/report');
+const ruleMessages = require('../../utils/ruleMessages');
+const validateOptions = require('../../utils/validateOptions');
+
+const ruleName = 'comment-pattern';
+
+const messages = ruleMessages(ruleName, {
+	expected: (pattern) => `Expected comment to match specified pattern "${pattern}"`,
```
jeddy3, line 13:
```suggestion
	expected: (pattern) => `Expected comment to match pattern "${pattern}"`,
```

Let's drop "specified" as we show the pattern in the message. I think including the pattern is a good idea, but let's create a follow-up issue to do this consistently across the other `*-pattern` rules.

lib/rules/comment-pattern/__tests__/index.js
```diff
@@ -0,0 +1,113 @@
...
+			message: messages.expected('foo-.+'),
+		},
+		{
+			code: '/**/',
+			message: messages.expected('foo-.+'),
+		},
+	],
+});
+
+testRule({
```
jeddy3, line 75: I think we can remove this `testRule`.

---

I'd expect `config: [true]` to produce an invalid option error. This is likely a short coming in the `validateOptions` function, that we can pick up another time.

jeddy3 (changes requested):
@omril321 Looking good, thank you. Just a couple of minor requests from me.

lib/rules/comment-pattern/index.js
omril321: Great. I can open an issue soon

lib/rules/comment-pattern/__tests__/index.js
omril321: I removed this. I'll open an issue soon 👍

hudochenkov (approved):

lib/rules/comment-pattern/README.md
```diff
@@ -0,0 +1,38 @@
+# custom-property-pattern
```
ybiquitous, line 1: Oh, I noticed the typo. 😅 

```suggestion
# comment-pattern
```

omril321: well, that's embarrassing 😅

ybiquitous (approved):
LGTM! 👏

jeddy3 (changes requested):
This looks good to me, once the redundant `code: '.without-comment {}',` are removed.

omril321:
> This looks good to me, once the redundant `code: '.without-comment {}',` are removed.

Thanks for your patience 🙏  I wasn't thorough enough when I removed redundant tests . I've just removed them

hudochenkov (approved):

jeddy3 (approved):

LGTM. Thank you for contributing to stylelint!

hudochenkov:
> - Added: `comment-pattern` rule ([#4962](https://github.com/stylelint/stylelint/pull/4962)).

Issue #4942, opened 2020-09-15
Migrate to PostCSS 8


hudochenkov:
Steps:

- [x] Update to latest PostCSS dependencies
- [x] Fix failing tests 
- [x] Fix failing TypeScript errors
- [x] Migrate to built-in rules to visitor API to remove deprecation warnings
- [x] Investigate impact on executing order

[Release notes](https://github.com/postcss/postcss/releases/tag/8.0.0).
[Migration guide](https://evilmartians.com/chronicles/postcss-8-plugin-migration).

---

Before we migrate we need all our dependencies be on PostCSS 8.

`dependencies` are ready:

- [x] [`@stylelint/postcss-css-in-js`](https://github.com/stylelint/postcss-css-in-js) — https://github.com/stylelint/postcss-css-in-js/pull/74 (removed)
- [x] [`@stylelint/postcss-markdown`](https://github.com/stylelint/postcss-markdown) (removed)
- [x] [`autoprefixer`](https://github.com/postcss/autoprefixer)
- [x] [`postcss`](https://github.com/postcss/postcss)
- [x] [`postcss-html`](https://github.com/stylelint/postcss-html) — https://github.com/stylelint/postcss-html/pull/1. (removed)
- [x] [`postcss-less`](https://github.com/shellscape/postcss-less)
- [x] [`postcss-media-query-parser`](https://github.com/dryoma/postcss-media-query-parser) — it doesn't use PostCSS.
- [x] [`postcss-resolve-nested-selector`](https://github.com/davidtheclark/postcss-resolve-nested-selector) — doesn't depend on PostCSS.
- [x] [`postcss-safe-parser`](https://github.com/postcss/postcss-safe-parser)
- [x] [`postcss-sass`](https://github.com/AleshaOleg/postcss-sass)
- [x] [`postcss-scss`](https://github.com/postcss/postcss-scss)
- [x] [`postcss-selector-parser`](https://github.com/postcss/postcss-selector-parser) — doesn't have PostCSS as peer or regular dependency.
- [x] [`postcss-syntax`](https://github.com/stylelint/postcss-syntax) — https://github.com/stylelint/postcss-syntax/pull/2. (removed)
- [x] [`postcss-value-parser`](https://github.com/TrySound/postcss-value-parser) — doesn't depend on PostCSS.
- [x] [`sugarss`](https://github.com/postcss/sugarss)

`devDependencies` are ready:

[...]

PR #4963, opened 2020-10-04
Use own vendor utility instead of PostCSS

hudochenkov:

Related to https://github.com/stylelint/stylelint/issues/4942.

`postcss.vendor` was removed in PostCSS 8. I did what Andrey Sitnik did with Autoprefixer: copied `vendor` to project :)

### DIFF
--- a/lib/rules/at-rule-allowed-list/index.js
+++ b/lib/rules/at-rule-allowed-list/index.js
@@ -4,10 +4,10 @@
 
 const _ = require('lodash');
 const isStandardSyntaxAtRule = require('../../utils/isStandardSyntaxAtRule');
-const postcss = require('postcss');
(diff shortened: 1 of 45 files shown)


malsf21 (approved):
LGTM! Can't say I know much about why postcss removed the `postcss.vendor` option, but this seems like a reasonable replacement to me!

hudochenkov:
@malsf21 it was removed because only few projects were using it, so it was removed to make package smaller.

malsf21:
Ah, thanks for letting me know! Still getting used to a lot of the dependencies for stylelint, so it's always great to learn more 😄

ybiquitous (approved):
Looks good! 👏

jeddy3 (approved):
Nice, thank you!

PR #4965, opened 2020-10-06
Add ignoreAtRules to property-no-unknown

### LINKED ISSUE #4928: Add ignoreAtRules: [] to property-no-unknown

I would like me and my colleagues to be able to use edge-of-your-seat new features to CSS. Sometimes features are not known to Stylelint yet and it would be nice if developers wouldn't have to disable linting for a few lines in the meantime.

I was thinking it might be useful to by default disable `property-no-unknown` linting within `@supports` blocks within the scss. Seeing as this marks a point where someone intends to progressive enhance their site with new features, I think it might be redundant to check it and have the developer disable bits of code manually. 

See for [explanation on `@supports` MDN website](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports)

Zacharias3690:

Closes #4928 

No, it's self explanatory

### DIFF
--- a/lib/rules/property-no-unknown/README.md
+++ b/lib/rules/property-no-unknown/README.md
@@ -141,7 +141,7 @@ The following patterns are _not_ considered violations:
 
 <!-- prettier-ignore -->
 ```css
-@supports(display: grid) {
+@supports (display: grid) {
   a {
     my-property: 1;
   }
--- a/lib/rules/property-no-unknown/README.md
+++ b/lib/rules/property-no-unknown/README.md
@@ -129,7 +129,7 @@ The following patterns are _not_ considered violations:
 
 ### `ignoreAtRules: ["/regex/", /regex/, "string"]`
 
-Skips checking properties of the given at-rules.
+Ignores properties nested within specified at-rules.
 
 Given:
 
@@ -141,9 +141,9 @@ The following patterns are _not_ considered violations:
 
 <!-- prettier-ignore -->
 ```css
-@supports(display: my-property) {
-  div {
-    display: my-property
+@supports(display: grid) {
+  a {
+    my-property: 1;
   }
 }
 ```
--- a/lib/rules/property-no-unknown/README.md
+++ b/lib/rules/property-no-unknown/README.md
@@ -143,7 +143,7 @@ The following patterns are _not_ considered violations:
 ```css
 @supports(display: my-property) {
   div {
-      display: my-property
+    display: my-property
   }
 }
 ```
--- a/lib/rules/property-no-unknown/README.md
+++ b/lib/rules/property-no-unknown/README.md
@@ -129,7 +129,7 @@ The following patterns are _not_ considered violations:
 
 ### `ignoreAtRules: ["/regex/", /regex/, "string"]`
 
-Skips checking properties of the given at-rules
+Skips checking properties of the given at-rules.
 
 Given:
 
@@ -142,9 +142,9 @@ The following patterns are _not_ considered violations:
 <!-- prettier-ignore -->
 ```css
 @supports(display: my-property) {
-    div {
-        display: my-property;
-    }
+  div {
+      display: my-property
+  }
 }
 ```
 
--- a/lib/rules/property-no-unknown/README.md
+++ b/lib/rules/property-no-unknown/README.md
@@ -127,6 +127,27 @@ The following patterns are _not_ considered violations:
 }
 ```
 
+### `ignoreAtRules: ["/regex/", /regex/, "string"]`
+
+Skips checking properties of the given at-rules
+
+Given:
+
+```
+["supports"]
+```
+
+The following patterns are _not_ considered violations:
+
+<!-- prettier-ignore -->
+```css
+@supports(display: my-property) {
+    div {
+        display: my-property;
+    }
+}
+```
+
 ### `checkPrefixed: true | false` (default: `false`)
 
 If `true`, this rule will check vendor-prefixed properties.
--- a/lib/rules/property-no-unknown/__tests__/index.js
+++ b/lib/rules/property-no-unknown/__tests__/index.js
@@ -367,16 +367,16 @@ testRule({
 
 	accept: [
 		{
-			code: '@supports(display: grid) { my-property: 1; }',
+			code: '@supports (display: grid) { my-property: 1; }',
 		},
 		{
 			code: '@my-at-rule { foo: 1; }',
 		},
 		{
-			code: '@supports(display:grid) { @media (min-width: 10px) { foo: 1; } }',
+			code: '@supports (display:grid) { @media (min-width: 10px) { foo: 1; } }',
 		},
 		{
-			code: '@supports(display:grid) { @media (min-width: 10px) { a { foo: 1; } } }',
+			code: '@supports (display:grid) { @media (min-width: 10px) { a { foo: 1; } } }',
 		},
 		{
 			code: '@my-other-at-rule { a { foo: 1; } }',
@@ -391,10 +391,10 @@ testRule({
 			column: 21,
 		},
 		{
-			code: '@my_at_rule { foo: 1; }',
+			code: '@not-my-at-rule { foo: 1; }',
 			message: messages.rejected('foo'),
 			line: 1,
-			column: 15,
+			column: 19,
 		},
 		{
 			code: 'a { foo: 1; }',
@@ -403,10 +403,10 @@ testRule({
 			column: 5,
 		},
 		{
-			code: '@my_at_rule foobar { foo: 1; }',
+			code: '@not-my-at-rule foobar { foo: 1; }',
 			message: messages.rejected('foo'),
 			line: 1,
-			column: 22,
+			column: 26,
 		},
 	],
 });
--- a/lib/rules/property-no-unknown/__tests__/index.js
+++ b/lib/rules/property-no-unknown/__tests__/index.js
@@ -376,20 +376,19 @@ testRule({
 			code: '@supports(display:grid) { @media (min-width: 10px) { foo: 1; } }',
 		},
 		{
-			code:
-				'@supports(display:grid) { @media (min-width: 10px) {  @my-property-value: { foo: 1; } } }',
+			code: '@supports(display:grid) { @media (min-width: 10px) { a { foo: 1; } } }',
 		},
 		{
-			code: '@my-other-at-rule { a: { foo: 1; } }',
+			code: '@my-other-at-rule { a { foo: 1; } }',
 		},
 	],
 
 	reject: [
 		{
-			code: '@import { a { foo: 1; } }',
+			code: '@media screen { a { foo: 1; } }',
 			message: messages.rejected('foo'),
 			line: 1,
-			column: 15,
+			column: 21,
 		},
 		{
 			code: '@my_at_rule { foo: 1; }',
--- a/lib/rules/property-no-unknown/__tests__/index.js
+++ b/lib/rules/property-no-unknown/__tests__/index.js
@@ -360,3 +360,64 @@ testRule({
 		},
 	],
 });
+
+testRule({
+	ruleName,
+	config: [true, { ignoreAtRules: ['media', '/^my-/'] }],
+
+	accept: [
+		{
+			code: 'a { @media { foo: 1 } }',
+		},
+		{
+			code: 'a { b { @media { foo: 1 }}}',
+		},
+		{
+			code: 'a { @my-at-rule { foo: 1 } }',
+		},
+		{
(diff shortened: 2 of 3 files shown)


lib/rules/property-no-unknown/README.md
```diff
@@ -127,6 +127,27 @@ The following patterns are _not_ considered violations:
 }
 ```
 
+### `ignoreAtRules: ["/regex/", /regex/, "string"]`
+
+Skips checking properties of the given at-rules
```
ybiquitous, line 132: [nitpick] Missing period (`.`)?

```suggestion
Skips checking properties of the given at-rules.
```

lib/rules/property-no-unknown/README.md
```diff
@@ -127,6 +127,27 @@ The following patterns are _not_ considered violations:
...
+```
+
+The following patterns are _not_ considered violations:
+
+<!-- prettier-ignore -->
+```css
+@supports(display: my-property) {
+    div {
+        display: my-property;
+    }
```
ybiquitous, line 147: [nitpick] Different indentation from other snippets in this page?

```suggestion
  div {
    display: my-property;
  }
```

lib/rules/property-no-unknown/__tests__/index.js
```diff
@@ -360,3 +360,64 @@ testRule({
...
+		{
+			code: 'a { foo: 1; }',
+			message: messages.rejected('foo'),
+			line: 1,
+			column: 5,
+		},
+	],
+});
+
+testRule({
```
ybiquitous, line 405: [question] What reason to split this `testRule` from the `testRule` above?

lib/rules/property-no-unknown/index.js
```diff
@@ -67,12 +70,16 @@ function rule(actual, options) {
 				return;
 			}
 
-			const { selector } = decl.parent;
+			const { selector, name } = decl.parent;
```
ybiquitous, line 73:
```suggestion
			const { selector, type, name } = decl.parent;
```

lib/rules/property-no-unknown/index.js
```diff
@@ -67,12 +70,16 @@ function rule(actual, options) {
...
 
 			if (selector && optionsMatches(options, 'ignoreSelectors', selector)) {
 				return;
 			}
 
+			if (name && optionsMatches(options, 'ignoreAtRules', name)) {
```
ybiquitous, line 79: [suggestion] How about checking also `type`? Because:

- More clear of the code intention
- Safer even if other node types would support the `name` property.

```suggestion
			if (type === 'atrule' && optionsMatches(options, 'ignoreAtRules', name)) {
```

See also <https://postcss.org/api/#atrule-type>

ybiquitous (commented):
Thank you for your contribution! I think it's almost good. 👍 
But I've left some comments, so please confirm them.

lib/rules/property-no-unknown/__tests__/index.js
Zacharias3690: I took the tests from max-nesting-depth and modified them for my needs. The first test uses 2 strings (the second being regex inside a string) while the second tests uses regex directly. I'd be more than happy to merge if you feel it's unnecessary.

ybiquitous: I got it. I feel it more understandable if the two `testRule` are merged to one `testRule` like this:

```js
config: [true, { ignoreAtRules: ['media', '/^my-/', /^custom-/i] }],
```

But this is just my opinion. You can ignore it. 👍

ybiquitous: Also, it seems better to add a test case for an at-rule with arguments, for example:

```js
code: 'a { @my-at-rule(foo: bar) { foo: 1; } }',
```

lib/rules/property-no-unknown/README.md
```diff
@@ -127,6 +127,27 @@ The following patterns are _not_ considered violations:
...
+["supports"]
+```
+
+The following patterns are _not_ considered violations:
+
+<!-- prettier-ignore -->
+```css
+@supports(display: my-property) {
+  div {
+      display: my-property
```
ybiquitous, line 146: [nipick] Thanks for fixing the indentation, but it remains:

```suggestion
    display: my-property
```

lib/rules/property-no-unknown/__tests__/index.js
Zacharias3690: Will do!

lib/rules/property-no-unknown/README.md
Zacharias3690: Whoops! I'll update my editor's spacing defaults and fix.

ybiquitous (approved):
Thank you for the quick fix. LGTM! 👍

lib/rules/property-no-unknown/README.md
```diff
@@ -127,6 +127,27 @@ The following patterns are _not_ considered violations:
 }
 ```
 
+### `ignoreAtRules: ["/regex/", /regex/, "string"]`
+
+Skips checking properties of the given at-rules.
```
jeddy3, line 132: Let's go with the following to be more consistent with other READMEs:

```suggestion
Ignore properties nested within specified at-rules
```

lib/rules/property-no-unknown/README.md
```diff
@@ -127,6 +127,27 @@ The following patterns are _not_ considered violations:
...
+
+The following patterns are _not_ considered violations:
+
+<!-- prettier-ignore -->
+```css
+@supports(display: my-property) {
+  div {
+    display: my-property
+  }
+}
```
jeddy3, line 148: Let's make it clear the property (not the value) is unknown.

```suggestion
@supports(display: grid) {
  a {
    my-property: 1;
  }
}
```

lib/rules/property-no-unknown/__tests__/index.js
```diff
@@ -360,3 +360,56 @@ testRule({
 		},
 	],
 });
+
+testRule({
+	ruleName,
+	config: [true, { ignoreAtRules: ['media', '/^my-/', /^custom-/i] }],
```
jeddy3, line 366: Let's use one string and one regex:

```suggestion
	config: [true, { ignoreAtRules: ['supports', /^my-/] }],
```

lib/rules/property-no-unknown/__tests__/index.js
```diff
@@ -360,3 +360,56 @@ testRule({
...
+
+	accept: [
+		{
+			code: 'a { @media { foo: 1 } }',
```
jeddy3, line 370: Let's use standard CSS in all the tests, i.e. only nesting rule-sets within at-rules (and not the other way around):

For example:

```css
@supports(display: grid) {
  .foo {
    my-property: 1;
  }
}
```

I suspect we'll find that this test will fail because only the parent node is being checked.

lib/rules/property-no-unknown/index.js
```diff
@@ -67,12 +70,16 @@ function rule(actual, options) {
 				return;
 			}
 
-			const { selector } = decl.parent;
+			const { selector, name, type } = decl.parent;
```
jeddy3, line 73: We need to check all the at-rules up to the root, not just the parent. 

For example, given `"supports"` the following should be accepted:

```css
@supports(display: grid) {
  @media (min-width: 10px) {
    a {
      my-property: 1
    }
  }
}
```

jeddy3 (changes requested):
@Zacharias3690 Thanks for making a start on this. I've requested some changes, including a substantial one to how this option should work.

Zacharias3690:
Sorry to be a bit of a bother, but would it be possible to add `hacktoberfest-accepted` label to this MR? I'd like it to be accepted in time, and am worried about how long the review process may take.

lib/rules/property-no-unknown/__tests__/index.js
```diff
@@ -360,3 +360,54 @@ testRule({
...
+		},
+		{
+			code: '@my-at-rule { foo: 1; }',
+		},
+		{
+			code: '@supports(display:grid) { @media (min-width: 10px) { foo: 1; } }',
+		},
+		{
+			code:
+				'@supports(display:grid) { @media (min-width: 10px) {  @my-property-value: { foo: 1; } } }',
```
hudochenkov, line 380:
```suggestion
				'@supports(display:grid) { @media (min-width: 10px) {  a { foo: 1; } } }',
```

`@my-property-value:` is not standard at-rule. Can we use `a` instead?

lib/rules/property-no-unknown/__tests__/index.js
```diff
@@ -360,3 +360,54 @@ testRule({
...
+		},
+		{
+			code: '@my-other-at-rule { a: { foo: 1; } }',
```
hudochenkov, line 383:
```suggestion
			code: '@my-other-at-rule { a { foo: 1; } }',
```

lib/rules/property-no-unknown/__tests__/index.js
```diff
@@ -360,3 +360,54 @@ testRule({
...
+		},
+	],
+
+	reject: [
+		{
+			code: '@import { a { foo: 1; } }',
```
hudochenkov, line 389:
```suggestion
			code: '@media screen { a { foo: 1; } }',
```

Make it more standard.

hudochenkov (changes requested):

hudochenkov (approved):

hudochenkov:
@jeddy3 is it good to go?

@Zacharias3690 I've added the label :)

lib/rules/property-no-unknown/__tests__/index.js
```diff
@@ -360,3 +360,53 @@ testRule({
...
+
+	accept: [
+		{
+			code: '@supports(display: grid) { my-property: 1; }',
+		},
+		{
+			code: '@my-at-rule { foo: 1; }',
+		},
+		{
+			code: '@supports(display:grid) { @media (min-width: 10px) { foo: 1; } }',
```
jeddy3, line 376:
```suggestion
			code: '@supports (display:grid) { @media (min-width: 10px) { foo: 1; } }',
```

lib/rules/property-no-unknown/__tests__/index.js
```diff
@@ -360,3 +360,53 @@ testRule({
...
+		},
+		{
+			code: '@supports(display:grid) { @media (min-width: 10px) { a { foo: 1; } } }',
```
jeddy3, line 379:
```suggestion
			code: '@supports (display:grid) { @media (min-width: 10px) { a { foo: 1; } } }',
```

lib/rules/property-no-unknown/__tests__/index.js
```diff
@@ -360,3 +360,53 @@ testRule({
...
+
+	reject: [
+		{
+			code: '@media screen { a { foo: 1; } }',
+			message: messages.rejected('foo'),
+			line: 1,
+			column: 21,
+		},
+		{
+			code: '@my_at_rule { foo: 1; }',
```
jeddy3, line 394:
```suggestion
			code: '@not-my-at-rule { foo: 1; }',
```

lib/rules/property-no-unknown/__tests__/index.js
```diff
@@ -360,3 +360,53 @@ testRule({
...
+			column: 15,
+		},
+		{
+			code: 'a { foo: 1; }',
+			message: messages.rejected('foo'),
+			line: 1,
+			column: 5,
+		},
+		{
+			code: '@my_at_rule foobar { foo: 1; }',
```
jeddy3, line 406:
```suggestion
			code: '@note-my-at-rule foobar { foo: 1; }',
```

lib/rules/property-no-unknown/README.md
```diff
@@ -127,6 +127,27 @@ The following patterns are _not_ considered violations:
...
+
+```
+["supports"]
+```
+
+The following patterns are _not_ considered violations:
+
+<!-- prettier-ignore -->
+```css
+@supports(display: grid) {
```
jeddy3, line 144:
```suggestion
@supports (display: grid) {
```

jeddy3 (changes requested):
@Zacharias3690 Thanks for making the changes. 

Just a couple of minor nitpicks, then I think we're good!

- whitespace after the at-rule name as more common to see it like that, even though without a space [is valid and parses correctly](https://astexplorer.net/#/gist/5026e869417c52357101f826c907505b/a7f0c93696956797bfc84ef530ccb898ed1af2e5)
- using `not-...` rather than underscores in names, to be consistent with the tests in other rules

(My bad for not including the space after the at-rule name in the original request).

jeddy3 (approved):

PR #4972, opened 2020-10-08
Fix `isStandardSyntaxDeclaration.test.js` that use callbacks

ybiquitous:

This is a part of #4881.

Note that this change removes the test cases about LESS variables.
Because the `@var: b` code is not a declaration, but an at-rule.

```console
$ node -e 'console.log(require("postcss-less").parse("@var: b"))'
<ref *1> Root {
  raws: { semicolon: false, after: '' },
  type: 'root',
  nodes: [
    AtRule {
      raws: [Object],
      type: 'atrule',
      name: 'var',
      parent: [Circular *1],
      source: [Object],
      params: 'b',
      variable: true,
      value: 'b'
    }
  ],
  source: {
    input: Input { css: '@var: b', hasBOM: false, id: '<input css 1>' },
    start: { line: 1, column: 1 }
  }
}
```


lib/utils/__tests__/isStandardSyntaxDeclaration.test.js
```diff
@@ -1,241 +1,179 @@
...
+		expect(
+			isStandardSyntaxDeclaration(scssDecl('a { $list: (key: value, key2: value2) }')),
+		).toBeFalsy();
 	});
 
 	it('sass nested property', () => {
-		sassDecls('a\n\tborder:\n\t\tstyle: solid\n\t\tcolor: red', (decl) => {
-			expect(isStandardSyntaxDeclaration(decl)).toBeFalsy();
-		});
+		expect(isStandardSyntaxDeclaration(sassDecl('a\n\tborder:\n\t\tstyle: solid'))).toBeFalsy();
```
ybiquitous, line 137: [note] Note that this case removes the `color: red` declaration.

```diff
-a\n\tborder:\n\t\tstyle: solid\n\t\tcolor: red
+a\n\tborder:\n\t\tstyle: solid
```

Because:

- `sassDecl()` returns only one declaration
- it is needless to test two declarations

lib/utils/__tests__/isStandardSyntaxDeclaration.test.js
```diff
@@ -1,241 +1,179 @@
...
 	});
 
 	it('scss nested property', () => {
-		scssDecls('a { border: { style: solid; color: red; } }', (decl) => {
-			expect(isStandardSyntaxDeclaration(decl)).toBeFalsy();
-		});
+		expect(isStandardSyntaxDeclaration(scssDecl('a { border: { style: solid; } }'))).toBeFalsy();
```
ybiquitous, line 141: [note] `color: red;` is removed also here.

lib/utils/__tests__/isStandardSyntaxDeclaration.test.js
```diff
@@ -1,241 +1,179 @@
...
-		});
-	});
-
-	it('less var', () => {
-		lessDecls('@var: b', (decl) => {
-			expect(isStandardSyntaxDeclaration(decl)).toBeFalsy();
-		});
-	});
-
-	it('nested less var', () => {
```
ybiquitous, line 214: [note] The following test cases are removed. See the PR description.

- `'less var'`
- `'nested less var'`

jeddy3 (approved):
LGTM, thanks!

lib/utils/__tests__/isStandardSyntaxDeclaration.test.js
```diff
@@ -1,241 +1,179 @@
...
+const postcssLess = require('postcss-less');
+const postcssSass = require('postcss-sass');
+const postcssScss = require('postcss-scss');
 
 describe('isStandardSyntaxDeclaration', () => {
 	it('standard prop and value', () => {
-		decls('a { a: b }', (decl) => {
-			expect(isStandardSyntaxDeclaration(decl)).toBeTruthy();
-		});
+		expect(isStandardSyntaxDeclaration(decl('a { a: b }'))).toBeTruthy();
```
hudochenkov, line 11: Could we use `toBe(true)` and `toBe(false)` for a more strict comparison, please?

lib/utils/__tests__/isStandardSyntaxDeclaration.test.js
```diff
@@ -1,241 +1,179 @@
...
+		expect(isStandardSyntaxDeclaration(lessDecl('a { &:extend(b) }'))).toBeFalsy();
 	});
 });
 
-function decls(css, cb) {
-	postcss.parse(css).walkDecls(cb);
+function decl(css, parser = postcss) {
+	const list = [];
+
+	parser.parse(css).walkDecls((decl) => list.push(decl));
```
hudochenkov, line 160: `no-shadow` :)

hudochenkov (changes requested):

hudochenkov (approved):

ybiquitous:
@hudochenkov Thanks for the review! I've addressed the comments via ac6439e.

PR #4973, opened 2020-10-09
Report disables in the same manner as lints

### LINKED ISSUE #4896: Report disable errors the same way as lint errors

Currently, errors caused by `reportNeedlessDisables`, `reportInvalidScopeDisables`, and eventually `reportUnscopedDisables` (https://github.com/stylelint/stylelint/issues/2292) and `reportDisables` (https://github.com/stylelint/stylelint/issues/4875) are reported in a format that's inconsistent with (and less visually appealing than) the way normal lints are recorded.

I propose that these disable issues instead be converted to the same object structure as lints, and be passed to the standard lint formatter to be reported appropriately. That way they'll automatically play nicely with both default and custom formatters, and look consistent with the most common reports.

nex3:
Rather than returning disables that are considered "invalid" for
various reasons (such as --report-needless-disables) as a separate
list of objects that formatters must handle individually, stylelint
now converts them into standard lint warnings. This allows formatters
to automatically format them the same as normal warnings without any
extra disable-specific code.

Closes #4896


jathak (approved):

lib/needlessDisables.js
```diff
@@ -1,95 +1,92 @@
...
+			}
+		}
 
-				// If this range is used but an equivalent has been marked as unused,
-				// remove that equivalent. This can happen because of the duplication
-				// of ranges in rule-specific range sets and the "all" range set
-				if (range.used && alreadyMarkedUnused) {
-					_.remove(unused.ranges, alreadyMarkedUnused);
-				}
+		Object.entries(rangeData).forEach(([rule, ranges]) => {
```
vankop, line 67: let's not mix `forEach` with `for..of`. btw you can optimize `const entries = Object.entries(rangeData)` and use `entries` identifier both on line 67 and line 60.

lib/needlessDisables.js
```diff
@@ -1,95 +1,92 @@
...
-				// mark this range as unused
-				if (!range.used && !alreadyMarkedUnused) {
-					unused.ranges.push({
-						rule,
-						start: range.start,
-						end: range.end,
-						unusedRule: rule,
-					});
-				}
+				_.remove(ranges, (otherRange) => range.comment === otherRange.comment);
```
vankop, line 63: not sure that mutating `result._postcssResult.stylelint.disabledRanges` is fine

lib/reportDisables.js
```diff
@@ -43,21 +35,20 @@ module.exports = function (results) {
...
-				reported.ranges.push({
-					rule,
-					start: range.start,
-					end: range.end,
-					unusedRule: rule,
+				// If the comment doesn't have a location, we can't report a useful error.
+				// In practice we expect all comments to have locations, though.
+				if (!range.comment.source || !range.comment.source.start) return;
+
+				result.warnings.push({
```
vankop, line 42: maybe we need sort result by line? previously there was some sorting

lib/needlessDisables.js
nex3: Good point, I should be cloning that field.

lib/needlessDisables.js
nex3: Done.

lib/reportDisables.js
nex3: I believe lints are automatically sorted (since they're all initially appended to the lint list in the order the lints are run, they'd have to be). The only reason this had to manually sort was that it wasn't being treated as a lint.

I checked manually and it does look like these lints are already being sorted appropriately relative to existing lints in the report.

vankop (approved):

PR #4975, opened 2020-10-10
Show the pattern in "*-pattern" rule messages (#4974)

### LINKED ISSUE #4974: Add the pattern in "*-pattern" rule messages

Some `*-pattern` rules do not print the pattern on their messages. This may be confusing the the users, since they may not may which pattern they must follow. 

As mentioned on https://github.com/stylelint/stylelint/pull/4962#discussion_r500380497, `*-pattern` rules should specify the pattern on their messages. For example:
`Expected custom media query name to match specified pattern *<pattern goes here>*`

The `*-pattern` rules I've found which do not specify the pattern are:

1. custom-media-pattern
3. custom-property-pattern
4. keyframes-name-pattern
5. selector-class-pattern
6. selector-id-pattern
7. selector-nested-pattern

hudochenkov:

Closes #4974.

No, it's self-explanatory.

### DIFF
--- a/lib/rules/custom-media-pattern/__tests__/index.js
+++ b/lib/rules/custom-media-pattern/__tests__/index.js
@@ -27,25 +27,25 @@ testRule({
 	reject: [
 		{
 			code: '@custom-media --foa-bar (min-width: 0);',
-			message: messages.expected,
+			message: messages.expected(/foo-.+/),
 			line: 1,
 			column: 15,
 		},
 		{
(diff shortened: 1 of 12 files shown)


jeddy3 (approved):
Looks great, thanks!

hudochenkov:
> - Added: configured pattern in `*-pattern` rules violation message ([#4975](https://github.com/stylelint/stylelint/pull/4975)).

PR #4976, opened 2020-10-10
Add Node.js 14 to CI

hudochenkov:

None.

Node.js 14 [goes LTS on 27.10.2020](https://nodejs.org/en/about/releases/). We need to be sure stylelint works there. ~~As soon it went LTS we could switch default Node.js version for linting and coverage to 14.~~ Decided to change it now, because I see no downsides of not doing so now.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -20,7 +20,7 @@ jobs:
 
     strategy:
       matrix:
-        node: [12]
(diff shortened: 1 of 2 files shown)


ybiquitous (approved):
LGTM 👍

jeddy3 (approved):
LGTM, thanks!

Issue #4818, opened 2020-06-03
Release 13.6.0


jeddy3:
- [x] stylelint release
- [x] stylelint-config-recommended update/release
- [x] stylelint-config-standard update/release
- [x] stylelint-demo update
- [x] stylelint.io update
- [x] tweet

I may have time tomorrow to put out a release. It'd be great if anyone has time to check through the [open pull requests](https://github.com/stylelint/stylelint/pulls) for ones needing reviews or minor tweaks, so we can get those in too.

jeddy3:
Done 🎉 

Thank you to everyone who contributed to this release!

PR #4988, opened 2020-10-14
Refactor formatter tests

ybiquitous:
I think that using `stripIndent()` or adding extra indent for the expected output of the formatters leads to inaccuracy and poor readability.
So, this change removes needless indent from the expected output in `lib/formatters/__tests__/*.test.js`.

I noticed the problem when I was looking into the source code for #4818.
(though the issue is not directly related to this PR)

No, it's self-explanatory.


lib/formatters/__tests__/verboseFormatter.test.js
```diff
@@ -18,12 +17,13 @@ describe('verboseFormatter', () => {
...
-      1 source checked
-       path/to/file.css
+		expect(output).toBe(
+			`
+1 source checked
+ path/to/file.css
 
-      0 problems found
-    `);
+0 problems found`.trimStart(),
```
ybiquitous, line 25: [note] Reluctantly, I use `trimStart()` only in this file in order to trim the newline at the beginning of the string.

hudochenkov (commented):
Indeed it look closer to reality now!

I think it would be great to use `.trimStart()` in every result with more than one line. This way we could see that all lines aligned.

ybiquitous:
@hudochenkov Make sense! I've used `.trimStart()` for the importance of readability, as possible (414f3d7).

hudochenkov (approved):

jeddy3 (approved):
Thanks for making these code cleanliness improvements in this pull request and others!

PR #4990, opened 2020-10-16
Fix `isStandardSyntaxTypeSelector.test.js` that use callbacks

ybiquitous:

This is a part of #4881.

I disable some test cases due to failure as follows:

```
● isStandardSyntaxTypeSelector › shadow-piercing descendant combinator

  TypeError: Cannot read property 'parent' of undefined

    14 | 	// The fake-tag's "parent" is actually a selector node, whose parent
    15 | 	// should be the :nth-child pseudo node.
  > 16 | 	if (!node.parent || !node.parent.parent) {
       | 	          ^
    17 | 		return false;
    18 | 	}
    19 |

    at isStandardSyntaxTypeSelector (lib/utils/isStandardSyntaxTypeSelector.js:16:12)
    at Object.<anonymous> (lib/utils/__tests__/isStandardSyntaxTypeSelector.test.js:49:10)
```

The cause seems due to [`postcss-selector-parser`](https://github.com/postcss/postcss-selector-parser) has not supported *Reference Combinator*, but I don't know the details. 🤷 
(maybe #2508 related?)

### DIFF
--- a/lib/utils/__tests__/isStandardSyntaxTypeSelector.test.js
+++ b/lib/utils/__tests__/isStandardSyntaxTypeSelector.test.js
@@ -6,101 +6,73 @@ const selectorParser = require('postcss-selector-parser');
 
 describe('isStandardSyntaxTypeSelector', () => {
 	it('tag', () => {
-		rules('a {}', (func) => {
-			expect(isStandardSyntaxTypeSelector(func)).toBeTruthy();
-		});
+		expect(isStandardSyntaxTypeSelector(tag('a {}'))).toBe(true);
 	});
 	it('lowercase nth-child pseudo selector', () => {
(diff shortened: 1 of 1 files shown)


ybiquitous:
https://github.com/stylelint/stylelint/blob/6ad94ac7253a01c31b69c1a183eeb9d4daa5f104/lib/utils/isStandardSyntaxTypeSelector.js#L47-L50

image

https://coveralls.io/builds/34214757/source?filename=lib/utils/isStandardSyntaxTypeSelector.js#L49

jeddy3 (approved):
LGTM, thanks!

Let's pick up the missing coverage and funkiness around reference combinators in other pull requests.

PR #4991, opened 2020-10-16
Fix `disableRanges.test.js` that uses callbacks

ybiquitous:

This is a part of #4881.

This change uses `async/await` a lot to avoid callbacks.

Also, it uses [`.rejects`](https://jestjs.io/docs/en/tutorial-async.html#rejects) of Jest to test thrown errors, instead of `throw/catch`.

Besides, it uses template literals a lot for the readability of multi-line CSS code.

---

Recommended **splitted** diff view: https://github.com/stylelint/stylelint/pull/4991/files?diff=split&w=1

### DIFF
--- a/lib/__tests__/disableRanges.test.js
+++ b/lib/__tests__/disableRanges.test.js
@@ -2,1006 +2,981 @@
 
 const _ = require('lodash');
 const assignDisabledRanges = require('../assignDisabledRanges');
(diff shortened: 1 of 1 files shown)


jeddy3 (approved):
@ybiquitous Many thanks! This is much cleaner.

Issue #4915, opened 2020-09-01
Add selector-attribute-name-disallowed-list


gbhasha:
> What is the problem you're trying to solve?

The developers on my team use class and id attributes in attribute selectors instead of using id selector or classname selector. This adds complexity and inconsistent coding style.

> What solution would you like to see?

To fix the above mentioned problem, having this linting rule would definitely help.

example use case: 

selector-attribute-disallowed-list: ['class', 'id'] 

should throw error for below selectors....
div[class*='some-thing']
div[id~='some-thing']

jeddy3:
@gbhasha Thanks for the request and for using the template.

> The developers on my team use class and id attributes in attribute selectors instead of using id selector or classname selector.

That's fascinating. I can imagine why they'd want to do that; if you're already using attribute selectors for things like `a[href$=".pdf"]::after { content: "(PDF)"; }`, why not consistently use them for everything else including class and id attributes?

Incidentally, it's possible to enforce that approach using:

```json
{ 
  "selector-max-id": 0,
  "selector-max-class": 0
}
```

Anyhow, you want to enforce the opposite and, as you said, that's not currently possible with stylelint.

> having this linting rule would definitely help... `selector-attribute-disallowed-list: ['class', 'id']`

Yes, I agree. I suggest we go with `selector-attribute-name-disallowed-list` so it's clear which part of the selector we are checking. I also suggest we check it as a CSS qualified name, i.e. include any namespaces ([spec ref](https://drafts.csswg.org/selectors/#attrnmsp)).

I assume you'll use `disallowed-list` approach, rather than `allowed-list`? If so, let's just add the former for now.

- Name: `selector-attribute-name-disallowed-list`
- Primary option: `array|string`
- Secondary options: None
- Autofixable: No
- Message: `Unexpected name "${name}"`
- Description: "Specify a list of disallowed attribute names."
- Section: "Limit language features" -> "Selector"

I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

There are [steps on how to add a new rule](https://stylelint.io/developer-guide/rules#add-a-rule) in the Developer guide.

[...]

PR #4992, opened 2020-10-18
Add selector-attribute-name-disallowed-list

rletsin:
Solution for Issue #4915 

I think no, it's self-explanatory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project are documented in this file.
 - Added: disable comments that are reported as errors for various reasons are now reported as standard lint errors rather than a separate class of errors that must be handled specially ([#4973](https://github.com/stylelint/stylelint/pull/4973)).
 - Added: configured pattern in `*-pattern` rules violation message ([#4975](https://github.com/stylelint/stylelint/pull/4975)).
 - Added: `comment-pattern` rule ([#4962](https://github.com/stylelint/stylelint/pull/4962)).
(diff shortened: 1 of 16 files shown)


rletsin:
I'll try to fix linting and test errors.
At the moment I have some trouble reproducing them locally.

jeddy3 (changes requested):
@rletsin Thanks. It's looks great so far!

In [my original proposal](https://github.com/stylelint/stylelint/issues/4915#issuecomment-684843841), I should have said use both the:

- [`selector-attribute-operator-disallowed-list`](https://github.com/stylelint/stylelint/tree/master/lib/rules/selector-attribute-operator-disallowed-list)
- [`media-feature-name-disallowed-list`](https://github.com/stylelint/stylelint/blob/master/lib/rules/media-feature-name-disallowed-list/)

rules as a blueprints; taking the AST traversal from `selector-attribute-operator-disallowed-list` and the options from `media-feature-name-disallowed-list`. As this rule targets names, rather than operators, it'll need to accept regexes like the `media-feature-name-disallowed-list` rule does.

Can you update the options, docs and tests so they are more aligned with `media-feature-name-disallowed-list` rule, please?

rletsin:
@jeddy3 Thanks for your suggestions!
I've added regex as possible input for this rule.

jeddy3 (approved):
Thanks for making the changes! 

Looks good to me.

ybiquitous (approved):
Looks nice! 👏

jeddy3:
Changelog:

> - Added: `selector-attribute-name-disallowed-list` rule ([#4992](https://github.com/stylelint/stylelint/pull/4992)).

Issue #5003, opened 2020-10-25
Fix TypeError for media queries in indentation


ryprice:
> Which rule, if any, is the bug related to?

indentation

> What code is needed to reproduce the bug?

```css
.fancyButton
  color: red

  @media only screen and (min-width: 600px)
    color: blue
```

> What stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "indentation": 2,
  }
}
```

> Which version of stylelint are you using?

13.7.2
This worked previously in 9.9.0, and I verified that with this isolated example.

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

`npx stylelint src/fancyButton.sass --syntax sass`

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

sass

> What did you expect to happen?

I'm using stylelint inside of webpack and this exception was really hard to debug because the only thing webpack was telling me was `ERROR in Cannot read property 'type' of undefined` without indicating the file or plugin that this exception was coming from. This is not the root issue, but it would have been nice for the stylelint webpack plugin to somehow identify itself as the thrower of the exception, as well as the file it failed to parse.

Anyway, I was able to isolate the issue to the described command line setup, and I would have expected no errors.

> What actually happened (e.g. what warnings or errors did you get)?

From npx stylelint: (line number might be slightly off because I was debugging with console.logs. I believe the error [comes from this line](https://github.com/stylelint/stylelint/blob/master/lib/rules/indentation/index.js#L149).)
```
TypeError: Cannot read property 'type' of undefined
    at indentationLevel (D:\src\listlab-web\node_modules\stylelint\lib\rules\indentation\index.js:157:20)
    at D:\src\listlab-web\node_modules\stylelint\lib\rules\indentation\index.js:62:22
    at D:\src\listlab-web\node_modules\postcss\lib\container.js:135:18
    at Rule.each (D:\src\listlab-web\node_modules\postcss\lib\container.js:101:16)
    at Rule.walk (D:\src\listlab-web\node_modules\postcss\lib\container.js:131:17)
    at D:\src\listlab-web\node_modules\postcss\lib\container.js:148:24
    at Root.each (D:\src\listlab-web\node_modules\postcss\lib\container.js:101:16)
    at Root.walk (D:\src\listlab-web\node_modules\postcss\lib\container.js:131:17)
    at D:\src\listlab-web\node_modules\stylelint\lib\rules\indentation\index.js:56:8
[...]

Issue #4865, opened 2020-07-13
Fix TypeError "Cannot read property 'syntax' of undefined" for nested control structures


Buszmen:
> Clearly describe the bug

A SASS code causes an error when linting:
`Cannot read property 'syntax' of undefined`

> Which rule, if any, is the bug related to?

I don't think it's related to any specific rule.

> What code is needed to reproduce the bug?

```css
$sides: 'top' 'bottom' 'left' 'right'
$sizes: 'xxs' 'xs' 'sm' 'md' 'lg' 'xlg'

@each $side in $sides
  @each $size in $sizes
    .ac-padding-#{$side}-#{$size}
      padding-#{$side}: var(--spacing-#{$size})
```

> What stylelint configuration is needed to reproduce the bug?

Default one?
It's even reproducible here:
[...]

PR #5012, opened 2020-10-30
Wrap rules in a try/catch to isolate rule execution exceptions

ryprice:
This try/catch adds an extra failsafe for when a rule fails so it doesn't crash the build. 
#5003 #4865 

**npx stylelint example**
Screen Shot 2020-10-30 at 3 32 25 PM

**webpack build example**
Screen Shot 2020-10-30 at 3 29 00 PM


jeddy3:
@ryprice Thanks for the pull request. This approach is in line with how we catch parse errors from the selector parser.

Anyone else have thoughts on this pull request?

ybiquitous:
In this approach, it seems that a stack trace of a thrown error is dismissed (we might not be able to know the error details).

ryprice:
@jeddy3 should the wrapper be generalized? Are there a bunch of different classes of rules that should all be wrapped in the same way? 

@ybiquitous You're right about that, but it gives the caller enough info to know what rule failed and what code caused the failure. Is there a way to output extra info in a verbose mode? Do you think isolating errors without failing the build go against the design of stylelint?

Anyway, for my situation, the most important part of this fix is indicating what file and rule failed before crashing. That was missing before, and it made unblocking myself an investigation into stylelint rather than some quick tinkering with the offending sass file.

ybiquitous:
In the structured output mode like `--formatter=json`, we can add extra details like a stack trace.

But, in the one-line output mode like `--formatter=compact`, multi-line text like a stack trace may be dismissed, and so I'm worried that users may be hard to know the cause of errors. 🤔 

> the most important part of this fix is indicating what file and rule failed before crashing

Surely, the error message should be improved.
I'm not sure whether it's best or not, but the following may be a solution:

```js
try {
  // ...
} catch (err) {
  console.error(err) // show original stack trace
  throw new Error(`${err.message}: file=${file}, rule=${rule}`)
}
```

m-allanson:
I agree that losing the stacktrace makes it harder to debug problems. 

It looks like stylelint is 'crashing appropriately' and erroring out when it hits a problem in a rule. But the crash information isn't clearly surfaced when running as part of a webpack build.

@ryprice does that sound right? If that's the case then maybe the stylelint webpack plugin could be updated to output more helpful error information?

ryprice:
@m-allanson I agree that the stylelint-webpack-plugin should have been more graceful as well. I think there's still a stylelint fix here to explain the failure in plain english, so maybe something similar to what @ybiquitous suggested, but still throw the original exception?

```javascript
try {
  // ...
} catch (err) {
  console.error(`Unexpected exception running Stylelint rule '${rule}' on '${fileName}'. Message: ${err.message}`);
  throw err;
}
```

ybiquitous:
@ryprice Surely, it is better to throw the original exception in order to keep the original stack trace. 👍 

Another way is to add a new property to a caught `Error` object as follows:

```js
try {
  // ...
} catch (err) {
  err.ruleDetails = {
    ruleName: ruleName,
    filePath: postcssResult.opts ? postcssResult.opts.from : undefined,
  };
  throw err;
}
```

Like `configurationError()` adds the `code` property:

https://github.com/stylelint/stylelint/blob/5a8465770b4ec17bb1b47f359d1a17132a204a71/lib/utils/configurationError.js#L9-L14

We can control an output error message in `cli.js` as follows:

```js
if (err.ruleDetails) {
  process.stderr.write(`Unexpected error occurs: rule=${err.ruleDetails.ruleName}, file=${err.ruleDetails.filePath}` + EOL);
}
process.stderr.write(err.stack + EOL);
```

https://github.com/stylelint/stylelint/blob/5a8465770b4ec17bb1b47f359d1a17132a204a71/lib/cli.js#L584-L589

m-allanson:
@ryprice I like your approach of throwing the original exception _in addition_ to the helpful error message. 👍 

@ybiquitous also has a good take on this, as his approach could be adapted to handle errors from any part of stylelint. 

Either way works for me!

ryprice:
Adding ruleName and filePath to the error object is interesting, but I feel like it's probably a solution that requires more thought to implement wholesale for all rule processing and as a contract between stylelint and stylelint-webpack-plugin. For now, I'll just spit it out as a console.error and throw the original exception. Then in stylelint-webpack-plugin, I'll at least put a bit more detail to indicate the error came from stylelint.

lib/lintPostcssResult.js
```diff
@@ -107,16 +107,27 @@ function lintPostcssResult(stylelintOptions, postcssResult, config) {
...
+								!postcssResult.stylelint.disabledRanges[ruleName],
+							newline,
+						})(postcssRoot, postcssResult);
+					} catch (ex) {
+						const fileName =
+							postcssRoot.source && postcssRoot.source.input
+								? postcssRoot.source.input.file
+								: 'unknown file';
+
+						// eslint-disable-next-line no-console
```
ryprice, line 126: How do we feel about disabling this rule here?

ybiquitous: IMO, disabling it here seems no problems. 👍 
The current ESLint configuration may be a bit too strict, so it seems good to me to loosen it via the `allow` option as follows:

```json
{
  "no-console": ["error", { "allow": ["warn", "error"] }]
}
```

https://github.com/stylelint/eslint-config-stylelint/blob/aecb506259e5371e162ee590b67d671d1bc72095/.eslintrc.js#L33

See also <https://eslint.org/docs/rules/no-console#options>

ybiquitous: About this, I've opened a new issue: https://github.com/stylelint/eslint-config-stylelint/issues/108

lib/lintPostcssResult.js
```diff
@@ -107,16 +107,27 @@ function lintPostcssResult(stylelintOptions, postcssResult, config) {
...
+								// Next two conditionals are temporary measures until #2643 is resolved
+								isFileFixCompatible &&
+								!postcssResult.stylelint.disabledRanges[ruleName],
+							newline,
+						})(postcssRoot, postcssResult);
+					} catch (ex) {
+						const fileName =
+							postcssRoot.source && postcssRoot.source.input
+								? postcssRoot.source.input.file
+								: 'unknown file';
```
ybiquitous, line 124: When I test this change with `--stdin`, the message includes `undefined` instead of `unknown file`:

```console
$ cat a.sass | bin/stylelint.js --stdin --syntax=sass
Rule indentation threw an unknown exception from undefined
```

So, can we add test cases about this change? It seems good to me to add them to `lib/__tests__/standalone.test.js`.

lib/lintPostcssResult.js
```diff
@@ -107,16 +107,27 @@ function lintPostcssResult(stylelintOptions, postcssResult, config) {
...
+
+						// eslint-disable-next-line no-console
+						console.error(`Rule ${ruleName} threw an unknown exception from ${fileName}`);
```
ybiquitous, line 127: [IMO] I have some suggestions about this message:

- How about quoting a rule name?
- How about quoting a file name?
- How about using `<input css>` if a file name is not present?

For example:

```
Rule "indentation" threw an unknown exception from "/home/foo/button.sass"

Rule "indentation" threw an unknown exception from <input css>
```

See also:
https://github.com/stylelint/stylelint/blob/f0b5aa864e9e1563aef6a697e2dd26d3100eb946/lib/createPartialStylelintResult.js#L81

PR #5019, opened 2020-11-02
Lazy load `autoprefixer` module

### LINKED ISSUE #4971: Fix slow autoprefixer calls in *-no-vendor-prefix

Stylelint is often launched as a pre-commit hook. It is important in this case, how fast the tool starts. If it takes several seconds to start (no matter how many files are committed), developers may notice that commits are slower than usual, and prefer some other tool, if it exists.

The overall performance of stylelint is influenced by a few heavy dependencies, that consume significant time when stylelint starts:
1. `require('@stylelint/postcss-css-in-js')` in https://github.com/stylelint/stylelint/blob/master/lib/syntaxes/syntax-css-in-js.js#L3 may take about 900 ms on my notebook (of course, exact value depends on hardware characteristics and many other factors). In other words, when stylelint decides to use this syntax, 900 ms is added to execution time.
2. When linting with `*-no-vendor-prefix` rules (`at-rule-no-vendor-prefix`, `media-feature-name-no-vendor-prefix`, `property-no-vendor-prefix`, `selector-no-vendor-prefix`, `value-no-vendor-prefix`), `require('autoprefixer')` takes about 500 ms here: https://github.com/stylelint/stylelint/blob/master/lib/utils/isAutoprefixable.js#L3.
3. `require('@stylelint/postcss-markdown')` in syntax-markdown - about 200 ms
4. `require('postcss-less')` in syntax-html, syntax-less, syntax-markdown - about 200 ms

In the worst case, the pre-commit hook may take about two seconds to lint even a one-line change in style.

In some cases, it is possible to require only used properties from dependency. I.e. instead of

```js

const prefixes = new Prefixes(
	autoprefixer.data.prefixes,
	new Browsers(autoprefixer.data.browsers, []),
);
```

one could write

```js
const data = {
	browsers: require('caniuse-lite/dist/unpacker/agents').agents,
	prefixes: require('autoprefixer/data/prefixes')
};

const prefixes = new Prefixes(
	data.prefixes,
	new Browsers(data.browsers, []),
);
```

This code is faster (about 400 ms instead of 500) because it does not evaluate the whole autoprefixer code, but it is more complex. Probably `import-lazy` also could be used more widely. Third-party libraries like autoprefixer and babel-core could be improved in order to provide more granular (== faster) imports.

I understand that there is no common solution, only partial mitigation. But performance tests indicate that developers of stylelint think about its speed, and right now slow startup adds the same or even more to the whole time than the rules themselves.

What do you think?

ybiquitous:
This change is based on my patch on <https://github.com/stylelint/stylelint/issues/4971#issuecomment-705320007>.
This mainly lazy-loads the [`autoprefixer`](https://github.com/postcss/autoprefixer) module in `lib/utils/isAutoprefixable.js`.

Maybe fix #4971

I don't know how to get an effective benchmark for this change. Could you lease help if you have any good ideas?

### DIFF
--- a/lib/utils/isAutoprefixable.js
+++ b/lib/utils/isAutoprefixable.js
@@ -1,9 +1,5 @@
 'use strict';
 
-const autoprefixer = require('autoprefixer');
-const Browsers = require('autoprefixer/lib/browsers');
-const Prefixes = require('autoprefixer/lib/prefixes');
-
 /**
  * Use Autoprefixer's secret powers to determine whether or
  * not a certain CSS identifier contains a vendor prefix that
@@ -13,10 +9,32 @@ const Prefixes = require('autoprefixer/lib/prefixes');
  * vendor prefixes.
  */
 
-const prefixes = new Prefixes(
-	autoprefixer.data.prefixes,
-	new Browsers(autoprefixer.data.browsers, []),
-);
+// NOTE: Lazy loading to avoid performance degrades.
+
+/** @type {import('autoprefixer') | undefined} */
+let autoprefixerCache;
+const autoprefixer = () => {
+	if (!autoprefixerCache) {
+		autoprefixerCache = require('autoprefixer');
+	}
+
+	return autoprefixerCache;
+};
+
+/** @type {import('autoprefixer/lib/prefixes') | undefined} */
+let prefixesCache;
+const prefixes = () => {
+	if (!prefixesCache) {
+		const Browsers = require('autoprefixer/lib/browsers');
+		const Prefixes = require('autoprefixer/lib/prefixes');
+
+		const data = autoprefixer().data;
+
+		prefixesCache = new Prefixes(data.prefixes, new Browsers(data.browsers, []));
+	}
+
+	return prefixesCache;
+};
 
 /**
  * Most identifier types have to be looked up in a unique way,
@@ -28,7 +46,7 @@ module.exports = {
 	 * @returns {boolean}
 	 */
 	atRuleName(identifier) {
-		return Boolean(prefixes.remove[`@${identifier.toLowerCase()}`]);
+		return Boolean(prefixes().remove[`@${identifier.toLowerCase()}`]);
 	},
 
 	/**
@@ -36,7 +54,7 @@ module.exports = {
 	 * @returns {boolean}
 	 */
 	selector(identifier) {
-		return prefixes.remove.selectors.some((/** @type {{ prefixed: string}} */ selectorObj) => {
+		return prefixes().remove.selectors.some((/** @type {{ prefixed: string}} */ selectorObj) => {
 			return identifier.toLowerCase() === selectorObj.prefixed;
 		});
 	},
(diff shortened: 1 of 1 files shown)


lib/utils/isAutoprefixable.js
```diff
@@ -64,8 +82,8 @@ module.exports = {
 	 * @returns {boolean}
 	 */
 	propertyValue(prop, value) {
-		const possiblePrefixableValues =
-			(prefixes.remove[prop.toLowerCase()] && prefixes.remove[prop.toLowerCase()].values) || false;
+		const remove = prefixes().remove[prop.toLowerCase()];
+		const possiblePrefixableValues = (remove && remove.values) || false;
```
ybiquitous, line 86: [note] Fixed duplication of `prefixes.remove[prop.toLowerCase()]`.

jeddy3 (approved):
> I don't know how to get an effective benchmark for this change. Could you lease help if you have any good ideas?

I don't know either, I'm afraid.

Code looks good, though.

hudochenkov:
`isAutoprefixable.js` is only used in `-no-vendor-prefix` rules. We lazy load all rules, and load only rules, which are in the config. Is this change is needed? Autoprefixer will be loaded anyway if it's needed. And it won't be loaded if not needed (no affected rule is enabled).

ybiquitous:
@hudochenkov Thanks for your comment. Surely, I also think we should avoid the hack on this PR if possible (to make the code simple).

But, when I examine whether lazy-loading `autoprefixer` works on the `master` branch, I find it does not. Please check it out below:

1. Output the log when `autoprefixer` is loaded as follows:

```diff
--- a/lib/utils/isAutoprefixable.js
+++ b/lib/utils/isAutoprefixable.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const autoprefixer = require('autoprefixer');
+console.log("autoprefixer loaded!!!")
 const Browsers = require('autoprefixer/lib/browsers');
 const Prefixes = require('autoprefixer/lib/prefixes');
 
```

2. Prepare a tested CSS file and `.stylelintrc.json` as follows:

```css
/* a.css */
@-webkit-keyframes { 0% { top: 0; } }
```

```json
{
  "rules": {
    "at-rule-no-vendor-prefix": null
  }
}
```

3. Run stylelint:

```console
$ bin/stylelint.js a.css --no-cache
autoprefixer loaded!!!
```

4. Remove `at-rule-no-vendor-prefix` from `.stylelintrc.json`:

```json
{
  "rules": {
  }
}
```

5. Run stylelint again:

```console
$ bin/stylelint.js a.css --no-cache
```

Perhaps, may this be another kind of bug?

hudochenkov:
Do you think we can find why it is loaded, when it shouldn't and try to fix that?

ybiquitous:
I'm not sure the following patch is correct, but it can prevent loading a rule module that has `null`:

```diff
--- a/lib/normalizeAllRuleSettings.js
+++ b/lib/normalizeAllRuleSettings.js
@@ -20,6 +20,10 @@ function normalizeAllRuleSettings(config) {
 	Object.keys(config.rules).forEach((ruleName) => {
 		const rawRuleSettings = _.get(config, ['rules', ruleName]);
 
+		if (rawRuleSettings === null) {
+			return;
+		}
+
 		const rule = rules[ruleName] || _.get(config, ['pluginFunctions', ruleName]);
 
 		if (!rule) {
```

https://github.com/stylelint/stylelint/blob/5a8465770b4ec17bb1b47f359d1a17132a204a71/lib/normalizeAllRuleSettings.js#L21

Issue #4831, opened 2020-06-16
Fix false positives for flex shorthand with 2 values in length-zero-no-unit


Sultanidze:
> Clearly describe the bug

There is false negative for `0px` in the flex shorthand rule:  
`flex: 1 0px;`

It is ambiguous situation for two values in flex shorthand property,
according to documentation:
https://developer.mozilla.org/en-US/docs/Web/CSS/flex

#### Two values: flex-grow | flex-basis:
`flex: 1 0px;`

#### Two values: flex-grow | flex-shrink:
`flex: 1 0;`

> Which rule, if any, is the bug related to?

e.g. `length-zero-no-unit`

> What code is needed to reproduce the bug?

```scss
.wrapper {
  flex: 1 0px;
}
```

> What stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "length-zero-no-unit": true,
  }
}
```

> Which version of stylelint are you using?

`13.6.0`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

via `stylelint-webpack-plugin`

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

No

> What did you expect to happen?

No `length-zero-no-unit` error to be flagged.
Possibly it would be good to create another rule to disallow 2 values in flex shorthand property to avoid ambiguous situations, and make intentions evident

> What actually happened (e.g. what warnings or errors did you get)?

The following warnings were flagged:

```shell
Value.module.scss
 32:12  ✖  Unexpected unit   length-zero-no-unit
```

jeddy3:
@Sultanidze Thanks for the report and for using the template.

I can reproduce this [using the demo](https://stylelint.io/demo#N4Igxg9gJgpiBcICGACYKBmAbGAPeKAjCgAwAOuA3CgL4gA0IAzgJ4B2ALkrguE0w3AQ2GAJYBzXsAA6bFCmkgATgFccTRQRlz5CkDjbiOACwC0ALxhKIptjZVtRHTSg6qYs+TVl0aQA).

I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

There are [steps on how to fix a bug in a rule](https://stylelint.io/developer-guide/rules#fix-a-bug-in-a-rule) in the Developer guide.

> Possibly it would be good to create another rule to disallow 2 values in flex shorthand property to avoid ambiguous situations, and make intentions evident

[...]

PR #5024, opened 2020-11-04
Fix false positives for flex shorthand with 2 values in length-zero-no-unit

Sultanidze:

#4831 


docs/user-guide/usage/node-api.md
```diff
@@ -80,7 +80,7 @@ As `config` contains no relative paths for `extends` or `plugins`, you do not ha
 ```js
 stylelint
   .lint({
-    config: { rules: "color-no-invalid-hex" },
+    config: { rules: { "color-no-invalid-hex": true } },
```
malsf21, line 83: Is there a reason why you made this change? Doesn't seem relevant to the rule - was it an autoformat fix or something?

lib/rules/length-zero-no-unit/index.js
```diff
@@ -40,6 +40,15 @@ function rule(actual, secondary, context) {
 			const ignorableIndexes = new Array(stringValue.length).fill(false);
 			const parsedValue = valueParser(stringValue);
 
+			if (decl.prop.toLowerCase() === 'flex') {
```
malsf21, line 43: May just want to add a comment pointing to the issue / MDN docs regarding the flex ambiguity - but will defer to others if they feel like this is appropriate!

malsf21 (commented):
Looks like this has slipped through the cracks! On an initial glance, it looks good to me  - just left some minor feedback. Thanks for your work on this so far @Sultanidze!

(not sure if there is a more portable/elegant way of performing the check - I'll defer to the Stylelint pros 😄 )

docs/user-guide/usage/node-api.md
Sultanidze: Hi @malsf21 and thanks for review. I've encountered broken config in node api user guide (Node v12.19.0 is on my machine if it means).

lib/rules/length-zero-no-unit/index.js
Sultanidze: @jeddy3 Should I add an inline comment here or issue number in commit message is enough?

jeddy3: An inline comment would be nice as this is tricky stuff. Perhaps:

```
// Allow <width> for last value in two-value and three-value flex syntax
// https://developer.mozilla.org/en-US/docs/Web/CSS/flex#syntax
```

jeddy3 (changes requested):
@Sultanidze Apologies for the delay in reviewing this.

Can we add two tests with a comment, please? The logic looks good, but it's worth checking with a test. For example:

```
// accept
code: 'a { flEx: 1 /* comment */ 0px; }'

// reject
code: 'a { flex: 1 0 /* comment */ 0px; }', 
fixed: 'a { flex: 1 0 /* comment */ 0; }', 
```

Sultanidze:
@jeddy3 thanks for advice to add test cases with comments. There are issues, I'm working on it.

jeddy3:
Revisiting this now that the rule was refactored in https://github.com/stylelint/stylelint/pull/5256.

I think we can simply ignore the `flex` property to resolve the issue. Looking at [the syntax again](https://developer.mozilla.org/en-US/docs/Web/CSS/flex#syntax), this rule should ignore all possible values for the flex property as they'll either be a `<number>` or intentionally a `<width>`.

@Sultanidze Would you like to update the pull request?

I think a simple`isFlex(prop)` function on [line 108](https://github.com/stylelint/stylelint/blob/master/lib/rules/length-zero-no-unit/index.js#L108) should do it. And the following two `"accept"` tests should suffice:

```
a { flex: 1 0px; }
a { FLEX: 1 1 0px; }
```

No need to add anything to the description of the rule either.

PR #5026, opened 2020-11-08
Fix incorrect enforce prefix regex documentation

miirinjej:
I'm not sure that I'm right, but your example with negative lookahead regex doesn't work for me. We can write positive lookbehind or just "^my-".

No, it's self-explanatory.


malsf21 (approved):
Hey @miirinjej,

Looks like this PR slipped through the cracks! Great observation though, and after running the suggested fix on [regex101](https://regex101.com/r/dNYtt0/1), it seems like the new change works as intended. LGTM, thank you for the catch!

PR #5044, opened 2020-11-18
Add dashed-ident-no-missing-reference-function

### LINKED ISSUE #5000: Add dashed-ident-no-missing-reference-function

Developers sometimes accidentally forget to include the `var()` function call when trying to dereference custom properties. For example, they might write:

```css
a {
  --my-color: --other-color;
  color: --my-color;
}
```

when they meant

```css
a {
  --my-color: var(--other-color);
  color: var(--my-color);
}
```

A new rule that flags identifiers beginning with `--` that appear outside of the first argument of a `var()` function.

jeddy3:

Closes #5000

I wanted to put together something to help feed into the discussion in #5000.

I've dropped `declaration-value` from the name and went with `dashed-ident-no-missing-reference-function` as I now feel `declaration-value` is redundant because [_dashed-idents_](https://drafts.csswg.org/css-values-4/#dashed-idents) outside of declarations values don't use reference functions.

In the future, we can add `ignoreDashedIdents` and `ignoresFunction` options:

```json
"rules": {
  "dashed-ident-no-missing-reference-function": [true, {
    "ignoreDashedIdents": ["--my-js-ident"],
    "ignoreFunctions": ["--custom"]
  }]
}
```

Which would allow:

```css
a {
  prop: --my-js-ident;
  prop: --custom(--foo);
}
```

Like the [`font-family-no-missing-generic-family-keyword` rule](https://stylelint.io/user-guide/rules/font-family-no-missing-generic-family-keyword), this rule sits in the [possible errors category](https://stylelint.io/user-guide/rules/list#possible-errors).


lib/rules/dashed-ident-no-missing-reference-function/index.js
```diff
@@ -0,0 +1,61 @@
...
+const validateOptions = require('../../utils/validateOptions');
+
+const ruleName = 'dashed-ident-no-missing-reference-function';
+
+const messages = ruleMessages(ruleName, {
+	rejected: (dashedIdent) =>
+		`Unexpected missing dashed-ident reference function for "${dashedIdent}"`,
+});
+
+const REFERENCE_FUNCTIONS = ['color', 'env', 'var'];
```
malsf21, line 19: Curious - are there many other uses cases for this set of reference functions? Can't think of any in my normal use-cases, but thinking about if we should pull this out to the constants files.

malsf21 (approved):
Looking through the linked discussion, test cases, and implementation, this seems good to me! 

P.S. `a { color: calc(var(--foo) + --bar)); }` is a great test case for this to work on!

lib/rules/dashed-ident-no-missing-reference-function/index.js
jeddy3: I think we can pull them out into a reference file if they end up being shared by another rule.

jeddy3:
Added a small commit to use `Set` to align with the changes in https://github.com/stylelint/stylelint/pull/5048.

lib/rules/dashed-ident-no-missing-reference-function/README.md
```diff
@@ -0,0 +1,35 @@
+# dash-ident-no-missing-reference-function
```
ybiquitous, line 1: Typo 😅 

```suggestion
# dashed-ident-no-missing-reference-function
```

ybiquitous (approved):
I found one typo, but this change looks good to me! 👍

alexander-akait:
Found this name is very misleading, in css we have `env` https://developer.mozilla.org/en-US/docs/Web/CSS/env() and in future them will be more

jeddy3:
> Found this name is very misleading, in css we have env https://developer.mozilla.org/en-US/docs/Web/CSS/env() and in future them will be more

Yes, I had a similar thought. 

Opened https://github.com/stylelint/stylelint/pull/5317.

Issue #4985, opened 2020-10-13
Enable `no-shadow` ESLint rule


ybiquitous:
This is a suggestion to enable the ESLint rule [`no-shadow`](https://eslint.org/docs/rules/no-shadow), which can prevent shadowing variables.

A quote from the rule document:

> Shadowing is the process by which a local variable shares the same name as a variable in its containing scope.
> (...)
> This can cause confusion while reading the code and it's impossible to access the global variable.

Here are this issue's purposes:

- To prevent unexpected bugs due to shadowing.
- To reduce our load of code review.
    - I recently couldn't notice shadowing in a PR and was received the review comment. See <https://github.com/stylelint/stylelint/pull/4970#discussion_r502831089>)

---

When I tried the rule on the commit 847fddc9, **100** issues were found.
This rule does not support auto-fixing, so we need to fix the issues manually.

```diff
--- a/package.json
+++ b/package.json
@@ -64,6 +64,9 @@
     "globals": {
       "testRule": true
     },
+    "rules": {
+      "no-shadow": "error"
+    },
     "root": true
   },
   "remarkConfig": {
```

```console
$ npm --silent run lint:js -- --format=codeframe
error: 'next' is already declared in the upper scope (no-shadow) at lib/assignDisabledRanges.js:108:10:
  106 | 
  107 | 			inlineEnd = current;
> 108 | 			const next = current.next();
      | 			      ^
  109 | 
  110 | 			if (!next || next.type !== 'comment') break;
  111 | 


error: 'report' is already declared in the upper scope (no-shadow) at lib/cli.js:508:13:
  506 | 
  507 | 					if (reportNeedlessDisables) {
> 508 | 						const report = disableOptionsReportStringFormatter(
      | 						      ^
  509 | 							linted.needlessDisables || [],
  510 | 							'needless disable',
  511 | 						);


error: 'report' is already declared in the upper scope (no-shadow) at lib/cli.js:519:13:
  517 | 
  518 | 					if (reportInvalidScopeDisables) {
> 519 | 						const report = disableOptionsReportStringFormatter(
      | 						      ^
  520 | 							linted.invalidScopeDisables || [],
  521 | 							'disable with invalid scope',
  522 | 						);


error: 'report' is already declared in the upper scope (no-shadow) at lib/cli.js:530:13:
  528 | 
  529 | 					if (reportDescriptionlessDisables) {
> 530 | 						const report = disableOptionsReportStringFormatter(
      | 						      ^
[...]

PR #5048, opened 2020-11-20
Assorted linting tweaks found with xo

XhmikosR:
I know we don't enforce most of these, but since I have [a branch](https://github.com/stylelint/stylelint/tree/xo) around, I decided to rebase it and submit a PR with the auto-fixes.

PS. I didn't commit the TS changes.
PPS. I didn't switch to `for...of` loops either but I can make this change if you all want.

### DIFF
--- a/lib/rules/alpha-value-notation/index.js
+++ b/lib/rules/alpha-value-notation/index.js
@@ -18,8 +18,8 @@ const messages = ruleMessages(ruleName, {
 	expected: (unfixed, fixed) => `Expected "${unfixed}" to be "${fixed}"`,
 });
 
-const ALPHA_PROPS = ['opacity', 'shape-image-threshold'];
-const ALPHA_FUNCS = ['hsl', 'hsla', 'hwb', 'lab', 'lch', 'rgb', 'rgba'];
+const ALPHA_PROPS = new Set(['opacity', 'shape-image-threshold']);
+const ALPHA_FUNCS = new Set(['hsl', 'hsla', 'hwb', 'lab', 'lch', 'rgb', 'rgba']);
 
 function rule(primary, options, context) {
 	return (root, result) => {
@@ -48,12 +48,12 @@ function rule(primary, options, context) {
 			parsedValue.walk((node) => {
 				let alpha;
 
-				if (ALPHA_PROPS.includes(decl.prop.toLowerCase())) {
+				if (ALPHA_PROPS.has(decl.prop.toLowerCase())) {
 					alpha = findAlphaInValue(node);
 				} else {
 					if (node.type !== 'function') return;
 
-					if (!ALPHA_FUNCS.includes(node.value.toLowerCase())) return;
+					if (!ALPHA_FUNCS.has(node.value.toLowerCase())) return;
 
 					alpha = findAlphaInFunction(node);
 				}
--- a/lib/rules/color-named/generateColorFuncs.js
+++ b/lib/rules/color-named/generateColorFuncs.js
@@ -13,7 +13,7 @@ function lin_sRGB(RGB) {
 			return val / 12.92;
 		}
 
-		return Math.pow((val + 0.055) / 1.055, 2.4);
+		return ((val + 0.055) / 1.055) ** 2.4;
 	});
 }
 
@@ -102,11 +102,7 @@ function rgb2hsl(r, g, b) {
 
 	l = (M + m) / 2;
 
-	if (d === 0) {
-		s = 0;
-	} else {
-		s = d / (1 - Math.abs(2 * l - 1));
-	}
+	s = d === 0 ? 0 : d / (1 - Math.abs(2 * l - 1));
 
 	s *= 100;
 	l *= 100;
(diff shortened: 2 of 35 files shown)


lib/rules/max-empty-lines/index.js
```diff
@@ -165,25 +165,21 @@ function rule(max, options, context) {
...
+				  })
+				: str.replace(/(\n)+/g, ($1) => {
+						if ($1.length > repeatTimes) {
+							return emptyLFLines;
+						}
+
+						return $1;
+				  });
 
 			return result;
```
ybiquitous, line 184: [suggestion] Does it make sense to inline the `result` variable? We can also remove the `TODO: Issue #4985` comment.

```suggestion
			return /(\r\n)+/g.test(str)
				? str.replace(/(\r\n)+/g, ($1) => {
						if ($1.length / 2 > repeatTimes) {
							return emptyCRLFLines;
						}

						return $1;
				  })
				: str.replace(/(\n)+/g, ($1) => {
						if ($1.length > repeatTimes) {
							return emptyLFLines;
						}

						return $1;
				  });
```

ybiquitous (approved):
I like this change! LGTM 👍

lib/rules/max-empty-lines/index.js
XhmikosR: Yup, there are 2 places we could do this now, but I wasn't sure if everyone liked the ternary change so I didn't remove the vars :) I can remove them later.

XhmikosR: Actually @ybiquitous feel free to push any patches to this branch :)

* https://github.com/stylelint/stylelint/pull/5048/files#diff-b3326c81bc17dd9317973d495767164226ba4f78683c2c63a8d5039cd5cf33c3R45
* https://github.com/stylelint/stylelint/pull/5048/files#diff-16a9f35cd6251529efb2d8fa37e0b7a218153be95052f697e35d5cbb6cbcd6edR168

ybiquitous: Thanks. Maybe, the 1st change may need `/** @type {PostcssResult} */` still.

jeddy3 (approved):
All LGTM, thanks!

XhmikosR:
I pushed the the temp var removal too; if tests are green we are good to go :)

ybiquitous (approved):
Thanks! 👏

lib/rules/alpha-value-notation/index.js
```diff
@@ -18,8 +18,8 @@ const messages = ruleMessages(ruleName, {
 	expected: (unfixed, fixed) => `Expected "${unfixed}" to be "${fixed}"`,
 });
 
-const ALPHA_PROPS = ['opacity', 'shape-image-threshold'];
-const ALPHA_FUNCS = ['hsl', 'hsla', 'hwb', 'lab', 'lch', 'rgb', 'rgba'];
+const ALPHA_PROPS = new Set(['opacity', 'shape-image-threshold']);
+const ALPHA_FUNCS = new Set(['hsl', 'hsla', 'hwb', 'lab', 'lch', 'rgb', 'rgba']);
```
hudochenkov, line 22: What is advantage of this over array?

lib/rules/color-named/generateColorFuncs.js
```diff
@@ -102,11 +102,7 @@ function rgb2hsl(r, g, b) {
 
 	l = (M + m) / 2;
 
-	if (d === 0) {
-		s = 0;
-	} else {
-		s = d / (1 - Math.abs(2 * l - 1));
-	}
+	s = d === 0 ? 0 : d / (1 - Math.abs(2 * l - 1));
```
hudochenkov, line 105: I would rather keep if statement here, because this line has so many symbols, it's very hard for a human to understand it.

lib/rules/color-named/generateColorFuncs.js
```diff
@@ -152,7 +148,7 @@ function generateColorFuncs(hexString) {
 	const rgb = [0, 0, 0];
 
 	for (let i = 0; i < 3; i += 1) {
-		rgb[i] = parseInt(hexString.substr(2 * i + 1, 2), 16);
+		rgb[i] = Number.parseInt(hexString.substr(2 * i + 1, 2), 16);
```
hudochenkov, line 151: Why not just keep `parseInt`? [Number.parseInt vs parseInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/parseInt#Number.parseInt_vs_parseInt).

lib/rules/color-named/generateColorFuncs.js
XhmikosR: This whole PR is automatic and I'd rather keep it that way.

lib/rules/alpha-value-notation/index.js
XhmikosR: * https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prefer-set-has.md
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

You can find plenty of other references on Google

lib/rules/color-named/generateColorFuncs.js
XhmikosR: >and is part of ECMAScript 2015 (its purpose is modularization of globals). Please see parseInt() for more detail and examples.

hudochenkov (approved):

Issue #4917, opened 2020-09-01
Fix false positives for two double slashed comments in *-empty-line-before


Eugeno:
> Clearly describe the bug

There are _expected empty line before declaration_ after **2 or more** lines of comments (with `//`). Neither in case of single line nor in case of version 13.6.1.

> Which rule, if any, is the bug related to?

`declaration-empty-line-before`

> What code is needed to reproduce the bug?

```scss
a {
  // Stylelint: Expected empty line before declaration
  // (declaration-empty-line-before)
  margin: 0;
}

b {
  // this is ok
  margin: 0;
}

i {
  /*
   * this is ok
   * too
   */
  margin: 0;
}
```

> What stylelint configuration is needed to reproduce the bug?

`stylelint-config-standard`

```json
"declaration-empty-line-before": [ "always", {
      except: [
        "after-declaration",
        "first-nested",
      ],
      ignore: [
        "after-comment",
        "inside-single-line-block",
      ],
    } ],
```

> Which version of stylelint are you using?

`13.7.0`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

CLI

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

Yep, comments at SCSS.

> What did you expect to happen?

No requirement for an empty line after comments (as said in configuration, `ignore: ["after-comment"]`).

> What actually happened (e.g. what warnings or errors did you get)?

```shell
test.css
31:5  ×  Expected empty line before declaration   declaration-empty-line-before
```

sebastianhaeni:
I'm having a related issue. 

The following SCSS code snippet will skip the suppress comment as well and yield 2 lint errors.
```css
.icon {
  // TODO use variable instead
  // stylelint-disable-next-line
  font-size: 21px;
}
```

Result:
```
my.component.scss
 26:3  ✖  Expected variable, function or keyword for "21px" of "font-size"   scale-unlimited/declaration-strict-value
 26:3  ✖  Expected empty line before declaration                                            declaration-empty-line-before
```

And if I change `// stylelint-disable-next-line` to `/* stylelint-disable-next-line */` it works, but that's not what we want to write in SCSS.

sebastianhaeni:
Ah probably related to #4911

jeddy3:
@Eugeno Thanks for the report and for using the template.

[...]

PR #5050, opened 2020-11-20
GH-4917 Amend always test case

agarzola:

Closes GH-4917.

At the moment, yes. This is just a Draft PR because after amending the test cases for the `at-rule-empty-line-before`, tests pass. I have confirmed that the test cases I added are being executed. I would expect tests to show a failure, since the test cases I’m adding are representative of the bug described in GH-4917.

I’ll appreciate any guidance on this issue, as I may be doing something wrong in my test cases.

## Testing Instructions

For running tests locally, I’ve taken the following steps:
- `npm ci`
- `npm run watch`
- Select option `o` to run files changed since last commit (before committing, of course).
- After tests were done, I searched the console output to confirm that my test cases had indeed been run as part of the suite. Each one of them showed as passing.

If you clone this branch and want to run this suite, you can:
- `npm run watch`
- Select the `p` option.
- Type `at-rule-empty-line-before`.
- Use the down arrow key to select the only file that matches.
- Press `Enter`.

You should see 1 test suite run with 362 tests passing and none failing.

### DIFF
--- a/lib/rules/at-rule-empty-line-before/__tests__/index.js
+++ b/lib/rules/at-rule-empty-line-before/__tests__/index.js
@@ -278,16 +278,6 @@ testRule({
 			code: '/* foo */\r\n\r\n@media {}',
 			description: 'CRLF',
 		},
-		{
-			code: '// foo\n// bar\n@media {}',
-		},
-		{
-			code: '// foo\n// bar\n\n@media{}',
-		},
-		{
-			code: '// foo\r\n// bar\r\n\r\n@media {}',
-			description: 'CRLF',
-		},
 	],
 
 	reject: [
@@ -303,6 +293,26 @@ testRule({
 			line: 2,
 			column: 1,
 			description: 'after shared-line comment',
+		},
+	],
+});
+
+testRule({
+	ruleName,
+	config: ['always', { ignore: ['after-comment'] }],
+	syntax: 'scss',
+	fix: true,
+
+	accept: [
+		{
+			code: '// foo\n// bar\n@media {}',
+		},
+		{
+			code: '// foo\n// bar\n\n@media{}',
+		},
+		{
+			code: '// foo\r\n// bar\r\n\r\n@media {}',
+			description: 'CRLF',
 		},
 	],
 });
--- a/lib/rules/at-rule-empty-line-before/__tests__/index.js
+++ b/lib/rules/at-rule-empty-line-before/__tests__/index.js
@@ -278,6 +278,16 @@ testRule({
 			code: '/* foo */\r\n\r\n@media {}',
 			description: 'CRLF',
 		},
+		{
+			code: '// foo\n// bar\n@media {}',
+		},
+		{
+			code: '// foo\n// bar\n\n@media{}',
+		},
+		{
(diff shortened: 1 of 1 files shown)


lib/rules/at-rule-empty-line-before/__tests__/index.js
```diff
@@ -278,6 +278,16 @@ testRule({
 			code: '/* foo */\r\n\r\n@media {}',
 			description: 'CRLF',
 		},
+		{
+			code: '// foo\n// bar\n@media {}',
+		},
+		{
+			code: '// foo\n// bar\n\n@media{}',
```
agarzola, line 285: I’m expecting this one to fail, given the failing parameters described in GH-4917.

jeddy3:
@agarzola Thanks for starting to look into this.

> I’ll appreciate any guidance on this issue, as I may be doing something wrong in my test cases.

You'll want to move those test cases into a new `testRule` that sets the syntax to `"scss"`, like so:

```js
testRule({
	ruleName,
	config: ['always', { ignore: ['after-comment'] }],
	syntax: "scss",
	fix: true,

	accept: [
		{
			code: '// foo\n// bar\n@media {}',
		},
		{
			code: '// foo\n// bar\n\n@media{}',
		},
		{
			code: '// foo\r\n// bar\r\n\r\n@media {}',
			description: 'CRLF',
		},
	]
});
```

It'll then use the Scss parser rather than CSS one. You may be wondering why the CSS parser didn't error... well, [dollar slashes are sometimes valid CSS](https://www.xanthir.com/b4U10), and can be [parsed by the CSS parser](https://astexplorer.net/#/gist/b9adfeda8a126f7f748d78b484312242/5f76904ef3c0ed7127bd2cb2c32bcb6c0bc845a7).

agarzola:
> It'll then use the Scss parser rather than CSS one. You may be wondering why the CSS parser didn't error... well, [double slashes are sometimes valid CSS](https://www.xanthir.com/b4U10), and can be [parsed by the CSS parser](https://astexplorer.net/#/gist/b9adfeda8a126f7f748d78b484312242/5f76904ef3c0ed7127bd2cb2c32bcb6c0bc845a7).

@jeddy3 Thank you for the speedy and very informative response! I had no idea `//` comments could sometimes be valid CSS. Thanks again!

lib/rules/at-rule-empty-line-before/__tests__/index.js
```diff
@@ -297,6 +297,26 @@ testRule({
...
+		},
+		{
+			code: '// foo\r\n// bar\r\n\r\n@media {}',
+			description: 'CRLF',
+		},
+	],
+});
+
```
agarzola, line 319: @jeddy3 I’ve separated out the test cases as you suggested, but it still passes. Any ideas?

agarzola: Never mind. I’m closing this PR, as the issue seems to have been resolved.

agarzola:
Closing this PR as it is no longer necessary. See https://github.com/stylelint/stylelint/issues/4917#issuecomment-731365669 for details.

Issue #4664, opened 2020-03-30
result object should inlcude a "fixed" field


audunolsen:
> What is the problem you're trying to solve?

Fix option makes me lazy and doesn't help me become a better coder, because I cannot
see what I did wrong. I want stylelint to raise my awareness of good styling practises while simultaneously speeding up my productivity.

> What solution would you like to see?

A fixed field in the result object, or alternatively have each warnings-entry have a "fixed" key.

This would allow me to have logs which may look like
```
_colors.sass
⚠ 4:7 Expected single space before "{" [fixed]
✖︎ 7:9 Expected a trailing semicolon    [fixed]
```

audunolsen:
Here's a workaround for people interested, only drawback is that it call's the linter two times on the same file/code.

```
let [
  {results : [{warnings}]},
  {results : [{warnings: afterFix}]}    
] = await Promise.all([
  lint(options),
  lint(Object.assign(options, {fix: true}))
]);
    
for (entry of warnings) if (!afterFix
  .map(e => JSON.stringify(e))
  .includes(JSON.stringify(entry))
) entry.fixed = true;
```

<img width="580" alt="Screenshot 2020-03-31 04 15 37" src="https://user-images.githubusercontent.com/16224137/77979952-55850380-7306-11ea-8df8-0d201beda90b.png">

jeddy3:
@audunolsen Thanks for the request, outlining your use case and using the template.

I didn't think is architecturally achievable right now, but that may change when we revisit the way rules report their fixable violations to address https://github.com/stylelint/stylelint/issues/4203 and  https://github.com/stylelint/stylelint/issues/2643. That is a sizeable piece of work that no one has found time to tackle yet.

I'll label this issue as an enhancement. It is also linked to those issues so that, hopefully, any solution to them will also factor in this use case.

Mouvedia:
Which core formatters will have to be updated?

ybiquitous:
> Which core [formatters](https://stylelint.io/user-guide/options#formatter) will have to be updated?

I think we can postpone changing the formatters until people request, if there are no special reasons.

Mouvedia:
> I think we can postpone changing the formatters until people request, if there are no special reasons.

Check [OP](https://github.com/stylelint/stylelint/issues/4664#issue-590372872), he wants the boolean _and_ the output.
We should at least dogfood our internal changes.
Maybe add it to the `verbose` one first?

ybiquitous:
[...]

PR #5062, opened 2020-11-25
Add support for a TAP formatter

### LINKED ISSUE #3889: Add TAP formatter

We run stylelint under Jenkins.  Our tooling, including other applications such as ESLint, generally output as TAP, which Jenkins picks up and gives us a pretty output for.  Unlike ESLint, Stylelint doesn't seem to currently have a method of outputting its results to TAP.

It would be extremely helpful to have a formatter which outputs the test results as TAP.

clinejj:
This PR adds support for a TAP formatter using an update to the code snippet from @danielsamuels discussed in #3889.

Closes #3889

No, it is self explanatory. Used #3524 as a starting point.

### DIFF
--- a/lib/formatters/tapFormatter.js
+++ b/lib/formatters/tapFormatter.js
@@ -14,18 +14,16 @@ const tapFormatter = (results) => {
 		);
 
 		if (result.warnings.length > 0) {
-			lines.push(...['---', 'messages:']);
+			lines.push('---', 'messages:');
 
 			result.warnings.forEach((warning) => {
 				lines.push(
-					...[
-						` - message: "${warning.text}"`,
-						`   severity: ${warning.severity}`,
-						`   data:`,
-						`     line: ${warning.line}`,
-						`     column: ${warning.column}`,
-						`     ruleId: ${warning.rule}`,
-					],
+					` - message: "${warning.text}"`,
+					`   severity: ${warning.severity}`,
+					`   data:`,
+					`     line: ${warning.line}`,
+					`     column: ${warning.column}`,
+					`     ruleId: ${warning.rule}`,
 				);
 			});
 
--- a/lib/formatters/tapFormatter.js
+++ b/lib/formatters/tapFormatter.js
@@ -1,0 +1,41 @@
+'use strict';
+
+/**
+ * @type {import('stylelint').Formatter}
+ */
+const tapFormatter = (results) => {
+	let lines = [`TAP version 13\n1..${results.length}`];
+
+	results.forEach((result, index) => {
+		lines.push(
+			`${result.errored ? 'not ok' : 'ok'} ${index + 1} - ${result.ignored ? 'ignored ' : ''}${
+				result.source
+			}`,
+		);
+
+		if (result.warnings.length > 0) {
+			lines.push(...['---', 'messages:']);
+
+			result.warnings.forEach((warning) => {
+				lines.push(
+					...[
+						` - message: "${warning.text}"`,
+						`   severity: ${warning.severity}`,
+						`   data:`,
+						`     line: ${warning.line}`,
+						`     column: ${warning.column}`,
+						`     ruleId: ${warning.rule}`,
+					],
+				);
+			});
+
+			lines.push('---');
+		}
+	});
+
+	lines.push('');
+
+	return lines.join('\n');
+};
+
+module.exports = tapFormatter;
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -91,7 +91,14 @@ declare module 'stylelint' {
 		returnValue?: StylelintStandaloneReturnValue,
 	) => string;
 
-	export type FormatterIdentifier = 'compact' | 'json' | 'string' | 'unix' | 'verbose' | Formatter;
+	export type FormatterIdentifier =
+		| 'compact'
+		| 'json'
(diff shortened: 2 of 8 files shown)


lib/formatters/tapFormatter.js
```diff
@@ -0,0 +1,41 @@
...
+						`     ruleId: ${warning.rule}`,
+					],
+				);
+			});
+
+			lines.push('---');
+		}
+	});
+
+	lines.push('');
```
clinejj, line 36: It's not clear to me why this needs to exist, but with some manual testing in another project if this is removed the last result will not be reported for larger sets of results. Would definitely be interested if there's a better approach based on what happens downstream of the formatter.

types/stylelint/index.d.ts
```diff
@@ -91,7 +91,14 @@ declare module 'stylelint' {
 		returnValue?: StylelintStandaloneReturnValue,
 	) => string;
 
-	export type FormatterIdentifier = 'compact' | 'json' | 'string' | 'unix' | 'verbose' | Formatter;
+	export type FormatterIdentifier =
```
clinejj, line 94: Added tap as an option but format was adjusted by prettier

lib/formatters/tapFormatter.js
```diff
@@ -0,0 +1,41 @@
...
+			`${result.errored ? 'not ok' : 'ok'} ${index + 1} - ${result.ignored ? 'ignored ' : ''}${
+				result.source
+			}`,
+		);
+
+		if (result.warnings.length > 0) {
+			lines.push(...['---', 'messages:']);
+
+			result.warnings.forEach((warning) => {
+				lines.push(
```
vankop, line 20: you can use `N` arguments to push in Array.prototype.push https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push

lib/formatters/tapFormatter.js
vankop: maybe for some tools `\n` is required in the end

vankop (approved):
Thanks for your PR

lib/formatters/tapFormatter.js
clinejj: Great point! Updated

lib/formatters/tapFormatter.js
clinejj: That makes sense, leaving as-is

jeddy3 (approved):
Thank you!

jeddy3:
Changelog entry:
> - Added: TAP formatter ([#5062](https://github.com/stylelint/stylelint/pull/5062)).

lib/formatters/tapFormatter.js
```diff
@@ -0,0 +1,39 @@
...
+					` - message: "${warning.text}"`,
+					`   severity: ${warning.severity}`,
+					`   data:`,
+					`     line: ${warning.line}`,
+					`     column: ${warning.column}`,
+					`     ruleId: ${warning.rule}`,
+				);
+			});
+
+			lines.push('---');
```
Mouvedia, line 30: @ybiquitous

According to https://testanything.org/tap-version-14-specification.html#yaml-diagnostics it should be `...`, right?

lib/formatters/tapFormatter.js
```diff
@@ -0,0 +1,39 @@
...
+
+	results.forEach((result, index) => {
+		lines.push(
+			`${result.errored ? 'not ok' : 'ok'} ${index + 1} - ${result.ignored ? 'ignored ' : ''}${
+				result.source
+			}`,
+		);
+
+		if (result.warnings.length > 0) {
+			lines.push('---', 'messages:');
```
Mouvedia, line 17: I am not sure but it seems to me that `\n` is missing.

> `YAMLBlock   := "  ---\n" (YAMLLine)* "  ...\n"`

lib/formatters/tapFormatter.js
ybiquitous: We have implemented the version 13, not version 14:

https://github.com/stylelint/stylelint/blob/eb6d61357765f39ab7dceeb71d7749e55422564e/lib/formatters/tapFormatter.mjs#L7

But it seems v13 also requires an ending `...`, according to <https://testanything.org/tap-version-13-specification.html>:

> If the test line is immediately followed by an indented block beginning with `/^\s+---/` and ending with `/^\s+.../` that block will be interpreted as an inline YAML document.

So, it makes sense for me to fix `---` to `...`. Additionally, we might be able to update the TAP version from 13 to 14.

Can you create an issue or pull request if you have time?

lib/formatters/tapFormatter.js
ybiquitous: See `return lines.join('\n');` at the end of the function.

lib/formatters/tapFormatter.js
Mouvedia: I was reviewing which formatter would have the `fixed` mention.
see https://github.com/stylelint/stylelint/issues/4664#issuecomment-2159060803

> Can you create an issue or pull request if you have time?

Not right away no, but what I wonder is why it wasn't picked up for all these years.
The tap consumers must be very lax.
This is weird.

Mouvedia: Actually, I am planning to send a PR this weekend.

PR #5079, opened 2020-12-15
Properly set the errored property for --report options

### LINKED ISSUE #5046: Fix regression where report flags don't exit with non-zero
> Clearly describe the bug

When a needless disable is found, the process exits with a zero code.

```json
{
  "extends": [
  ],
  "rules": {
    "function-disallowed-list": [
      "calc",
      "lighten",
      "darken"
    ],
    "property-disallowed-list": [
      "border-radius",
      "border-top-left-radius",
      "border-top-right-radius",
      "border-bottom-right-radius",
      "border-bottom-left-radius",
      "transition"
    ],
    "scss/dollar-variable-default": [
      true,
      {
        "ignore": "local"
      }
    ],
    "scss/selector-no-union-class-name": true
  }
}
```

13.8.0

CLI: `stylelint "**/*.{css,scss}" --cache --cache-location .cache/.stylelintcache --rd`

Probably not?

The process should error out and exit with non-zero code

The needless disables are printed, but the process exits with 0.

```shell
scss/_reboot.scss
 1:1  ×  Needless disable for "at-rule-no-vendor-prefix"  --report-needless-disables
```

---

Test repo:

1. `git clone https://github.com/twbs/bootstrap.git`
2. `npm i && npm run css-lint-stylelint`

---

Notes:

1. Running the same command one more time, does not print the needless disables at all unless I delete the cache file.
2. Those needless disables weren't caught before 13.8.0, I think. But when a needless disable was found, the process exited with non-zero code in the previous Stylelint version. So, it looks like this is a regression in 13.8.0.

nex3:
This sets the `errored` property (and thus the exit code) as long as
any warning has severity "error", which includes warnings generated by
--report options.

Closes #5046

### DIFF
--- a/system-tests/004/fs.test.js
+++ b/system-tests/004/fs.test.js
@@ -5,7 +5,7 @@ const { caseConfigFile, caseFilesForFix, prepForSnapshot } = require('../systemT
 
 const CASE_NUMBER = '004';
 
-it('no-fs - errored state for reportNeedlessDisables', async () => {
+it('fs - errored state for reportNeedlessDisables', async () => {
 	expect(
 		prepForSnapshot(
 			await stylelint.lint({
@@ -17,7 +17,7 @@ it('no-fs - errored state for reportNeedlessDisables', async () => {
 	).toMatchSnapshot();
 }, 10000);
 
-it('no-fs - no errored state', async () => {
+it('fs - no errored state', async () => {
 	expect(
 		prepForSnapshot(
 			await stylelint.lint({
--- a/system-tests/004/fs.test.js
+++ b/system-tests/004/fs.test.js
@@ -1,0 +1,29 @@
+'use strict';
+
+const stylelint = require('../../lib');
+const { caseConfigFile, caseFilesForFix, prepForSnapshot } = require('../systemTestUtils');
+
+const CASE_NUMBER = '004';
+
+it('no-fs - errored state for reportNeedlessDisables', async () => {
+	expect(
+		prepForSnapshot(
+			await stylelint.lint({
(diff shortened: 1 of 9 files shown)


hudochenkov (commented):
Do you think it is possible to add a test for this?

We have tests for different exit codes in https://github.com/stylelint/stylelint/blob/master/lib/__tests__/cli.test.js.

nex3:
There are already tests in there that verify the exit codes for report flags:

https://github.com/stylelint/stylelint/blob/bf6a881d9130447e5af858c021e2979bcdc96a9b/lib/__tests__/cli.test.js#L242-L255

It's not clear to me why that wasn't failing prior to this PR, or how to write a test that wouldn't just recapitulate what that test is already doing.

hudochenkov (approved):

vankop:
@nex3 @hudochenkov please rereview

system-tests/004/fs.test.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const stylelint = require('../../lib');
+const { caseConfigFile, caseFilesForFix, prepForSnapshot } = require('../systemTestUtils');
+
+const CASE_NUMBER = '004';
+
+it('no-fs - errored state for reportNeedlessDisables', async () => {
```
nex3, line 8: Should these tests start with `fs -` rather than `no-fs -`?

vankop: thanks, fixed

hudochenkov (approved):

nex3:
Tests look good to me, just need @vankop to approve the PR and I can land it.

vankop (approved):

Issue #4945, opened 2020-09-16
Add ignoreFunctions: [] to length-zero-no-unit


zhawkins:
> Clearly describe the bug

The `length-zero-no-unit` rule `ignore: ["custom-properties"]` option will only run on custom property declarations like:

```css
a { --x: 0px; }
```

It doesn't handle var functions like:

```
font: var(--x, 0px);
```

I see it being discussed here: https://github.com/stylelint/stylelint/pull/4860#issuecomment-662540528 but don't see a corresponding new issue so I created one. Maybe this would be considered a new feature? 🤔

> Which rule, if any, is the bug related to?

`length-zero-no-unit`

> What code is needed to reproduce the bug?

```css
.foo {
  --x: 0px;
  margin-top: var(--y, 0px);
}
```

> What stylelint configuration is needed to reproduce the bug?

e.g.

```json
{
  "rules": {
    "length-zero-no-unit": [
      true,
       {
         "ignore": ["custom-properties"]
       }
    ]
  }
}
```

> Which version of stylelint are you using?

`13.6.1`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

VScode plugin with `settings.json`

```
    "stylelint.validate": ["scss"],
    "editor.codeActionsOnSave": {
        "source.fixAll.stylelint": true
    }
```

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

No.

> What did you expect to happen?

I expect the unit on `0px` within `var(--y, 0px);` to be ignored.

> What actually happened (e.g. what warnings or errors did you get)?

"The following warnings were flagged:"

```shell
   Expected unit (length-zero-no-unit)
```

patric-eberle:
Same issue here. I MUST define a `0px` value as fallback, but the linter triggers.

jeddy3:
@zhawkins Thanks for creating the issue.

I suggest we add an `ignoreFunctions: []` option for this to offer granular control to the user.

For example:

```json
{
  "rules": {
    "length-zero-no-unit": [
      true,
       {
         "ignoreFunctions": ["var", "--custom"]
       }
    ]
  }
}
```

Would allow:

```css
.foo {
  margin-top: var(--y, 0px);
  margin-bottom: --custom(0px);
}
```

I've labelled the issue as ready to implement. @zhawkins or @patric-eberle Please consider [contributing](https://stylelint.io/contributing) if you have time.

[...]

PR #5082, opened 2020-12-18
Add ignoreFunctions: [] to length-zero-no-unit

JuanFML:
Added condition to ignore when function is detected and a test to check it with.

Fixes issue #4945 

I left some comments in the code as it is another type of solution, this is my first PR, so please let me know if the solution wasn't what you were looking for, in order to fix it. There would be still left to add to the README.md file the new option added, although I wanted to get it checked first before adding it.

### DIFF
--- a/lib/rules/length-zero-no-unit/index.js
+++ b/lib/rules/length-zero-no-unit/index.js
@@ -43,6 +43,8 @@ function rule(actual, options, context) {
 			return;
 		}
 
+		const ignoreFunctions = (options && options.ignoreFunctions) || [];
+
 		root.walkDecls((decl) => {
 			if (decl.prop.toLowerCase() === 'line-height') {
 				return;
@@ -71,7 +73,7 @@ function rule(actual, options, context) {
 				// TODO: Issue #4985
 				// eslint-disable-next-line no-shadow
 				const stringValue = valueParser.stringify(node);
-				const ignoreFlag = isMathFunction(node);
+				const ignoreFlag = isMathFunction(node) || ignoreFunctions.some((fn) => fn === node.value);
 
 				for (let i = 0; i < stringValue.length; i++) {
 					ignorableIndexes[node.sourceIndex + i] = ignoreFlag;
@@ -102,19 +104,6 @@ function rule(actual, options, context) {
 			const fixPositions = [];
 
 			styleSearch({ source: value, target: '0' }, (match) => {
-				const ignoreFunctions = (options && options.ignoreFunctions) || [];
-
-				if (
-					ignoreFunctions.length > 0 &&
-					ignoreFunctions.some((functionName) => {
-						if (value.includes(functionName)) return true;
-
-						return false;
-					})
-				) {
-					return;
-				}
-
 				const index = match.startIndex;
 
 				// Given a 0 somewhere in the full property value (not in a string, thanks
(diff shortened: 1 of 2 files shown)


lib/rules/length-zero-no-unit/index.js
```diff
@@ -16,16 +16,29 @@ const ruleMessages = require('../../utils/ruleMessages');
 const styleSearch = require('style-search');
 const validateOptions = require('../../utils/validateOptions');
 const valueParser = require('postcss-value-parser');
+//const { truncate } = require('lodash');
```
vankop, line 19: could you please delete unneeded comments

lib/rules/length-zero-no-unit/index.js
```diff
@@ -83,13 +96,36 @@ function rule(actual, secondary, context) {
...
+			// });
+
 			const fixPositions = [];
 
 			styleSearch({ source: value, target: '0' }, (match) => {
+				const ignoreFunctions = (options && options.ignoreFunctions) || [];
+
+				if (
+					ignoreFunctions.length > 0 &&
+					ignoreFunctions.map((functionName) => {
```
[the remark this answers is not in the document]
vankop: could you also add another rejected test case, where 
```
{
	code: '.foo { margin-top: var(--y, 0px); margin-bottom: --custom1(0px); };',
	description: 'Should error value in --custom1',
}
```

vankop (commented):
Thanks for your PR, some comments:

lib/rules/length-zero-no-unit/index.js
JuanFML: I have made the changes

vankop:
@JuanFML I have reworked things a little bit, take a look

JuanFML:
@vankop Yeah they work perfectly your changes, I don't have to commit again do I? With your commit is all done?

vankop:
> I don't have to commit again do I?
yes, all is fine

vankop (approved):

hudochenkov (commented):
Thank you for working on this!

Could you add test cases, please?

```css
/* it should report only first 0px */
a { margin: 0px var(--prop, 0px) }

/* it should report only second 0px */
a { margin: calc(5px + var(--prop, 0px)) 0px }
```

jeddy3:
Revisiting after the refactoring in https://github.com/stylelint/stylelint/pull/5256.

I think a simple conditional just after [line 52](https://github.com/stylelint/stylelint/blob/master/lib/rules/length-zero-no-unit/index.js#L52) should work now:

```js
if (isFunction(valueNode) && optionsMatches(secondary, 'ignoreFunctions', value)) return false;
```

With the following tests:

```js
testRule({
	ruleName,
	config: [true, { ignoreFunctions: ['var', /^--/] }],
	accept: [
		{
			code: 'a { top: var(--foo, 0px); }',
		},
		{
			code: 'a { top: --bar(0px); }',
		},
	],
	reject: [
		{
			code: 'a { margin: var(--foo, 0px) 0px --bar(0px); }',
			fixed: 'a { margin: var(--foo, 0px) 0 --bar(0px); }',

			message: messages.rejected,
			line: 1,
			column: 30,
		},
	],
});
```

We'll need to require `lodash` on line 5:

```js
const _ = require('lodash');
```

And update the options validator on line 41:

```js
ignoreFunctions: [_.isString, _.isRegExp],
```

And add a `isFunction` function on line 154:

```js
function isFunction({ type }) {
	return type === 'function';
}
```

The docs can look like [this](https://github.com/stylelint/stylelint/tree/72d7bab805af3b7a02df5081b83f007b3b290af0/lib/rules/property-no-vendor-prefix#optional-secondary-options), e.g.

Given:

```
["var", "/^--/"]
```

The following patterns are _not_ considered violations:


```css
a { top: var(--foo, 0px); }
```


```css
a { top: --bar(0px); }
```

@JuanFML Would you like to update the pull request?

PR #5089, opened 2020-12-22
Add ignoreUnits[] to unit-case

### LINKED ISSUE #5083: Fix false positives for spaceless operators in unit-case
> Clearly describe the bug

I want my Less variable to be itself，but it always changed to lowercase.

 `value-keyword-case`

```css
.u-input(@lineHeight, @paddingTop, @paddingBottom) {
  height: 1px*@lineHeight;
  padding-top: 1px*@paddingTop;
  padding-bottom: 1px*@paddingBottom;
  line-height: 1px*@lineHeight;
}
```

```json
{
  "rules": {
    "value-keyword-case":  ["lower", {
      "ignoreTypes": [/@\w+/]
    }],
  }
}
```

`12.0.0`

"CLI with `stylelint "**/*.css" --config myconfig.json`"

"Yes, it's related to Less nested properties."

`@lineHeight` should not to be `@lineheight` In the example, after run `stylelint "**/*.css" --config myconfig.json`. How to keep it self ?

superchow:

Closes #5083

It is pretty self explanatory, I mostly followed instructions from the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project are documented in this file.
 ## Head
 
 - Fixed: `selector-attribute-name-disallowed-list` false positives for valueless attribute selectors ([#5060](https://github.com/stylelint/stylelint/pull/5060)).
+- Added: `ignoreUnits` to `unit-case` ([#5089](https://github.com/stylelint/stylelint/pull/5089)).
 
 ## 13.8.0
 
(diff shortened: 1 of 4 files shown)


jeddy3:
Thanks for the pull request. I've looked into the issue some more and believe this is best solved by patching the [upstream parser](https://github.com/shellscape/postcss-less) so that `1px*@var` is parsed as a value, operators and variable. Please consider contributing there if you'd like to see the issue fixed.

Closing as to avoid adding code to workaround a bug in the upstream parser.

Issue #4826, opened 2020-06-09
Fix false negatives for object notation and style-search rules


swernerx:
> Clearly describe the bug

When using the jss-react syntax I see that only some rules are working properly. Interestingly this seems to be specific to object notation in JSS. I tried debugging this deep into the code but was not very successful even after deep inside `style-search`

> Which rule, if any, is the bug related to?

The most prominent rules we figured out are not working correctly are:

```
length-zero-no-unit: true
color-hex-length: short
```

> What code is needed to reproduce the bug?

```js
import { createUseStyles } from "react-jss"

const useStyles = createUseStyles({
  button: {
    borderRadius: '0px',
    font: "Arial",
    fontSize: 12,
    background: '#2200DD'
  }
})
```

> What stylelint configuration is needed to reproduce the bug?

```yaml
rules:
  length-zero-no-unit: true
  color-hex-length: short
  color-hex-case: lower
  unit-case: lower
```

> Which version of stylelint are you using?

`13.6.0`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

CLI, just the basic `stylelint --fix filename.js`

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

It is related to CSS-in-JS, in this case a JSS based syntax.

> What did you expect to happen?

Reporting the same issue like when the code was written in a template literal.

```js
import styled from "styled-components"

const Button = styled.button`
  border-radius: 0px;
  font: Arial;
  font-size: 12;
  background: #2200DD;
`
```

results in this output:

```shell
 4:19  ✖  Unexpected unit                          length-zero-no-unit
 5:9   ✖  Unexpected missing generic font family   font-family-no-missing-generic-family-keyword
 7:15  ✖  Expected "#2200DD" to be "#2200dd"       color-hex-case
 7:15  ✖  Expected "#2200DD" to be "#20D"          color-hex-length
```

> What actually happened (e.g. what warnings or errors did you get)?

Interestingly the font-family warning came up. So it seems to work at least in some part. But all of `color-hex-length`, `color-hex-case` and `unit-case` are not detected correctly.

```shell
npx stylelint --fix filename.js
6:11  ✖  Unexpected missing generic font family   font-family-no-missing-generic-family-keyword
```

swernerx:
There is a difference in how `styleSearch` is called:

- JSS: source=`borderRadius: '0px'`; target=`0`
- Styled Components: source=`border-radius: 0px`; target=`0`

[...]

PR #5101, opened 2021-01-06
Fix false negatives for css-in-js object notation in color-hex-case

Dru89:
As discussed in #4826, `style-search` has false positives when used with `css-in-js` object notation. This change updates the `color-hex-case` rule to use `postcss-value-parser` instead, which has support for `css-in-js` as a syntax.

The unit tests now also include a suite for testing both `upper` and `lower` options for `color-hex-case` with the `css-in-js` syntax.

### DIFF
--- a/lib/rules/color-hex-case/__tests__/index.js
+++ b/lib/rules/color-hex-case/__tests__/index.js
@@ -231,7 +231,7 @@ testRule({
 
 			message: messages.expected('#aBABAA', '#ababaa'),
 			line: 3,
-			column: 15,
+			column: 13,
 		},
 		{
 			code: `
@@ -246,9 +246,9 @@ testRule({
 			`,
 
 			warnings: [
-				{ message: messages.expected('#AAA', '#aaa'), line: 3, column: 36 },
-				{ message: messages.expected('#FaFa', '#fafa'), line: 3, column: 42 },
-				{ message: messages.expected('#0000FFcc', '#0000ffcc'), line: 3, column: 49 },
+				{ message: messages.expected('#AAA', '#aaa'), line: 3, column: 34 },
+				{ message: messages.expected('#FaFa', '#fafa'), line: 3, column: 40 },
+				{ message: messages.expected('#0000FFcc', '#0000ffcc'), line: 3, column: 47 },
 			],
 		},
 	],
@@ -320,7 +320,7 @@ testRule({
 
 			message: messages.expected('#aBABAB', '#ABABAB'),
 			line: 3,
-			column: 15,
+			column: 13,
 		},
 		{
 			code: `
@@ -335,9 +335,9 @@ testRule({
 			`,
 
 			warnings: [
-				{ message: messages.expected('#aaa', '#AAA'), line: 3, column: 36 },
-				{ message: messages.expected('#FaFa', '#FAFA'), line: 3, column: 42 },
-				{ message: messages.expected('#0000FFcc', '#0000FFCC'), line: 3, column: 49 },
+				{ message: messages.expected('#aaa', '#AAA'), line: 3, column: 34 },
+				{ message: messages.expected('#FaFa', '#FAFA'), line: 3, column: 40 },
+				{ message: messages.expected('#0000FFcc', '#0000FFCC'), line: 3, column: 47 },
 			],
 		},
 	],
--- a/lib/rules/color-hex-case/__tests__/index.js
+++ b/lib/rules/color-hex-case/__tests__/index.js
@@ -173,154 +173,93 @@ testRule({
 	accept: [
 		{
 			code: `
-import styled from 'styled-components';
-export const s = styled.a({
-	stroke: "url(#gradiantA)",
-});
-`,
+						import styled from 'styled-components';
+						export const s = styled.a({
+							color: "#aaa",
+						});
+			`,
+		},
+		{
+			code: `
+						import styled from 'styled-components';
+						export const s = styled.a({
+							stroke: "url(#AAA)",
+						});
+			`,
 			description: 'href with location',
 		},
 		{
 			code: `
-import styled from 'styled-components';
-export const s = styled.a({
-	color: "pink",
-});
-`,
+						import styled from 'styled-components';
+						export const s = styled.a({
+							color: "PINK",
+						});
+			`,
 		},
 		{
 			code: `
-import styled from 'styled-components';
-export const s = styled.a({
-	color: "#000",
-});
-`,
+						import styled from 'styled-components';
+						export const s = styled.a({
+							background: "linear-gradient(#aaa, #ffff, #0000ffcc)",
+						});
+			`,
 		},
 		{
 			code: `
-import styled from 'styled-components';
-export const s = styled.a({
-	something: "#000, #fff, #ababab",
-});
-`,
+						import styled from 'styled-components';
+						export const s = styled("a::before")({
+							content: '"#ABABAB"',
+						});
+			`,
 		},
 		{
 			code: `
-import styled from 'styled-components';
-export const s = styled.a({
-	color: "#0000ffcc",
-});
-`,
-			description: 'eight digits',
-		},
-		{
-			code: `
-import styled from 'styled-components';
-export const s = styled.a({
-	color: "#00fc",
-});
-`,
-			description: 'four digits',
-		},
-		{
-			code: `
-import styled from 'styled-components';
-export const s = styled.a({
-	padding: "000",
-});
-`,
-		},
-		{
-			code: `
-import styled from 'styled-components';
-export const s = styled("a::before")({
-	content: '"#ABABA"',
-});
-`,
-		},
-		{
-			code: `
-import styled from 'styled-components';
-export const s = styled.a({
-	color: "white /* #FFF */",
-});
-`,
+						import styled from 'styled-components';
+						export const s = styled.a({
+							color: "white /* #FFF */",
+						});
+			`,
 		},
 	],
 
 	reject: [
 		{
 			code: `
-import styled from 'styled-components';
-export const s = styled.a({
-	color: "#aBABA",
-});
-`,
+						import styled from 'styled-components';
+						export const s = styled.a({
+							color: "#aBABAA",
+						});
+			`,
 			fixed: `
-import styled from 'styled-components';
-export const s = styled.a({
-	color: "#ababa",
-});
-`,
+						import styled from 'styled-components';
+						export const s = styled.a({
+							color: "#ababaa",
+						});
+			`,
 
-			message: messages.expected('#aBABA', '#ababa'),
+			message: messages.expected('#aBABAA', '#ababaa'),
 			line: 4,
-			column: 9,
+			column: 15,
 		},
 		{
 			code: `
-import styled from 'styled-components';
-export const s = styled.a({
-	something: "#000F, #fff, #abababA",
-});
-`,
+						import styled from 'styled-components';
+						export const s = styled.a({
+							background: "linear-gradient(#AAA, #FaFa, #0000FFcc)",
+						});
+			`,
 			fixed: `
-import styled from 'styled-components';
-export const s = styled.a({
-	something: "#000f, #fff, #abababa",
-});
-`,
+						import styled from 'styled-components';
+						export const s = styled.a({
+							background: "linear-gradient(#aaa, #fafa, #0000ffcc)",
+						});
+			`,
 
 			warnings: [
-				{ message: messages.expected('#000F', '#000f'), line: 4, column: 13 },
-				{ message: messages.expected('#abababA', '#abababa'), line: 4, column: 26 },
+				{ message: messages.expected('#AAA', '#aaa'), line: 4, column: 36 },
+				{ message: messages.expected('#FaFa', '#fafa'), line: 4, column: 42 },
+				{ message: messages.expected('#0000FFcc', '#0000ffcc'), line: 4, column: 49 },
 			],
-		},
-		{
-			code: `
-import styled from 'styled-components';
-export const s = styled.a({
-	something: "#000, #FFFFAZ, #ababab",
-});
-`,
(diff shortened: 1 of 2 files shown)


Dru89:
This is my first contribution, so let me know if there's something I missed with respect to formatting or code style.

lib/rules/color-hex-case/index.js
```diff
@@ -26,68 +30,56 @@ function rule(expectation, options, context) {
...
+			let needsFix = false;
 
-			styleSearch({ source: declString, target: '#' }, (match) => {
-				const hexMatch = /^#[0-9A-Za-z]+/.exec(declString.substr(match.startIndex));
+			parsedValue.walk((node) => {
+				const { type, value } = node;
 
-				if (!hexMatch) {
-					return;
+				// Only traverse a function if it's a color function.
```
jeddy3, line 39: Let's pull this out into a self-documenting function, e.g.:

```js
if (isIgnoredFunction(node)) return false;

// at the bottom of file
function isIgnoredFunction({type, value}) {
  return type === 'function' && IGNORED_FUNCTIONS.has(value.toLowerCase())
}
```

The comment is redundant then.

lib/rules/color-hex-case/index.js
```diff
@@ -26,68 +30,56 @@ function rule(expectation, options, context) {
...
+				if (type === 'function' && IGNORED_FUNCTIONS.has(value.toLowerCase())) {
+					return false;
 				}
 
-				const hexValue = hexMatch[0];
-				const hexValueLower = hexValue.toLowerCase();
-				const hexValueUpper = hexValue.toUpperCase();
-				const expectedHex = expectation === 'lower' ? hexValueLower : hexValueUpper;
+				// Check to make sure the type is valid and that it matches a hex color code
```
jeddy3, line 44: Let's do the same for this:

```js
if (!isHexColor(node)) return;
```

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,323 @@ testRule(
...
 );
+
+testRule({
+	ruleName,
+	config: ['lower'],
+	syntax: 'css-in-js',
+	fix: true,
+	accept: [
+		{
+			code: `
```
jeddy3, line 175: Let's ident all these template literals for readability, e.g.:

```js
                   {
			code: `
                              import styled from 'styled-components';
                              export const s = styled.a({
	                             color: "pink",
                              });
                         `,
```

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,323 @@ testRule(
...
+import styled from 'styled-components';
+export const s = styled.a({
+	stroke: "url(#gradiantA)",
```
jeddy3, line 178:
```suggestion
	stroke: "url(#AAA)",
```

Let's use a string that looks like an uppercase hex to make sure it's ignored.

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,323 @@ testRule(
...
+});
+`,
+			description: 'href with location',
+		},
+		{
+			code: `
+import styled from 'styled-components';
+export const s = styled.a({
+	color: "pink",
```
jeddy3, line 187:
```suggestion
	color: "PINK",
```

Let's use uppercase to make sure it's ignored.

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,323 @@ testRule(
...
+});
+`,
+		},
+		{
+			code: `
+import styled from 'styled-components';
+export const s = styled.a({
+	color: "#000",
```
jeddy3, line 195:
```suggestion
	color: "#aaa",
```

Let's use lowercase, and let's move this test first as it's the clearest example of a valid pattern.

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,323 @@ testRule(
...
+export const s = styled.a({
+	padding: "000",
+});
+`,
+		},
+		{
+			code: `
+import styled from 'styled-components';
+export const s = styled("a::before")({
+	content: '"#ABABA"',
```
jeddy3, line 237:
```suggestion
	content: '"#ABABAB"',
```

Let's use a valid hex.

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,323 @@ testRule(
...
+			code: `
+import styled from 'styled-components';
+export const s = styled.a({
+	color: "#00fc",
+});
+`,
+			description: 'four digits',
+		},
+		{
+			code: `
```
jeddy3, line 226: I think we can delete this test as `padding: 000` isn't valid CSS value.

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,323 @@ testRule(
...
+import styled from 'styled-components';
+export const s = styled.a({
+	color: "#000",
+});
+`,
+		},
+		{
+			code: `
```
jeddy3, line 200: Let's combine the next three tests into valid CSS code, rather than use a `something` property, e.g.:

```css
background: linear-gradient(#aaa, #ffff, #0000ffcc);
```

This will cover the 4 and 8 digit tests.

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,323 @@ testRule(
...
+`,
+		},
+	],
+
+	reject: [
+		{
+			code: `
+import styled from 'styled-components';
+export const s = styled.a({
+	color: "#aBABA",
```
jeddy3, line 256:
```suggestion
	color: "#aBABAA",
```

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,323 @@ testRule(
...
+});
+`,
+			fixed: `
+import styled from 'styled-components';
+export const s = styled.a({
+	color: "#ababa",
```
jeddy3, line 262:
```suggestion
	color: "#ababaa",
```

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,323 @@ testRule(
...
+});
+`,
+
+			message: messages.expected('#aBABA', '#ababa'),
```
jeddy3, line 266:
```suggestion
			message: messages.expected('#aBABAA', '#ababaa'),
```

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,323 @@ testRule(
...
+
+testRule({
+	ruleName,
+	config: ['lower'],
+	syntax: 'css-in-js',
+	fix: true,
+	accept: [
+		{
+			code: `
+import styled from 'styled-components';
```
jeddy3, line 176: I believe can remove `import styled from 'styled-components';` from all the tests as we're explicitly specifying the syntax.

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,323 @@ testRule(
...
+	color: "#ababa",
+});
+`,
+
+			message: messages.expected('#aBABA', '#ababa'),
+			line: 4,
+			column: 9,
+		},
+		{
+			code: `
```
jeddy3, line 271: Let's combine the next three tests into one valid example:

```
background: linear-gradient(#AAA, #FaFa #0000FFcc);
```

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,323 @@ testRule(
...
+`,
+
+			message: messages.expected('#12345AA', '#12345aa'),
+			line: 4,
+			column: 25,
+		},
+	],
+});
+
+testRule({
```
jeddy3, line 328: Let's apply all the requested changes above to the `upper` tests too.

jeddy3 (changes requested):
@Dru89 Thanks! It's looking great so far, and so much cleaner now that the value parser is being used.

I've made a few requested changes.

Dru89:
Thanks for taking the time to look at all of this @jeddy3!

I've gone through and made the changes you requested. For many of them (like the invalid five hex colors, e.g. `#aBABA`), I pulled those directly from the test cases above. (My assumption was that the original tests still wanted to make sure that "invalid" hex colors still got transformed.)

Should we transform the original tests to also use more valid CSS? Or just leave them as they were before?

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -164,3 +164,181 @@ testRule(
 		],
 	}),
 );
+
+testRule({
```
jeddy3, line 168: I think we got our wires crossed with the indents. Let's indent the template literals as normal blocks, like so:

```js
testRule({
	ruleName,
	config: ['upper'],
	syntax: 'css-in-js',
	fix: true,
	accept: [
		{
			code: `
				export const s = styled.a({
				    color: "#AAA",
				});
			`,
		},
	],

	reject: [
		{
			code: `
				export const s = styled.a({
				    color: "#aBABAB",
				});
			`,
			fixed: `
				export const s = styled.a({
					color: "#ABABAB",
				});
			`,
			message: messages.expected('#aBABAB', '#ABABAB'),
			line: 3,
			column: 15,
		},
	],
});
```

jeddy3 (changes requested):
Fantastic, thanks for making the changes! It's almost there. Just one last tweak with the indents.

> Should we transform the original tests to also use more valid CSS? Or just leave them as they were before?

Let's leave them as they are. We default to using valid CSS, but deviating in this case makes sense as it's testing something particular. Whereas in the `testRule` for `"css-in-js"` were testing that objects are read correctly, rather than the edge cases of the rule itself, and so valid CSS is appropriate here.

lib/rules/color-hex-case/__tests__/index.js
Dru89: Ah, apologies! Github's 8-spaces-per-tab made it look like you were trying to align it with the backtick, not just trying to indent it by one line. I'll take care of this today.

jeddy3 (approved):
@Dru89 Thanks. Looks great to me!

It's awesome to see the improvements with the refactoring to the value parser.

Dru89:
Thanks! I was gonna start in on some of the other `style-search` use cases like `hex-color-length`, `color-no-hex`, and `value-no-vendor-prefix`.

Should I wait until this one gets the second approval and merged? Also, would you prefer one PR per rule?

jeddy3:
> Thanks! I was gonna start in on some of the other style-search use cases...

Fantastic!

> Should I wait until this one gets the second approval and merged? Also, would you prefer one PR per rule?

No need to wait. Yes, one pull request per rule, please.

lib/rules/color-hex-case/index.js
```diff
@@ -14,6 +15,9 @@ const messages = ruleMessages(ruleName, {
 	expected: (actual, expected) => `Expected "${actual}" to be "${expected}"`,
 });
 
+const HEX = /^#[0-9A-Za-z]+/;
```
hudochenkov, line 18: Shall we use RegExp, which looks for valid HEX only?

```suggestion
const HEX = /^#([\da-f]{3,4}|[\da-f]{6}|[\da-f]{8})$/i;
```

https://regexr.com/5k70k

CleanShot 2021-01-13 at 22 49 53

lib/rules/color-hex-case/index.js
```diff
@@ -26,68 +30,60 @@ function rule(expectation, options, context) {
...
+function getValue(decl) {
+	return decl.raws.value ? decl.raws.value.raw : decl.value;
+}
+
+function setValue(decl, value) {
+	if (decl.raws.value) decl.raws.value.raw = value;
+	else decl.value = value;
 
-	return stringStart + replaceString + stringEnd;
+	return decl;
```
hudochenkov, line 86: Return is not used anywhere.

lib/rules/color-hex-case/index.js
Dru89: My goal was to match the previous style search (and not break existing tests) for this one, and the original query was:

```
/^#[0-9A-Za-z]+/
```

But if we're ok with a potentially "breaking" change that will only look at valid hex colors, I'm ok to make that.

hudochenkov: I would say it's not a breaking change, but a bugfix. What do you think, @jeddy3?

jeddy3: I believe the rule was originally designed to check the case of anything that _sort of looks like a hex colour_, i.e. a word that starts with hash and is followed by numbers and letters, as:

1. people may want to come up with their own hex-like languages extensions
2. we already had the [color-no-invalid-hex rule](https://stylelint.io/user-guide/rules/color-no-invalid-hex) rule the people could use to check for hex colour validity against the specs

As it's more extensible, and because users may already be relying on this behaviour, I suggest we keep it as.

lib/rules/color-hex-case/index.js
jeddy3: @Dru89 It looks like this is the only change we need to make for a second approval.

jeddy3: @Dru89 It looks like this is the only change we need to make for a second approval.

Dru89:
Sorry for the delay in getting this updated. Last week was a bit crazy, but I've addressed that last piece of feedback.

hudochenkov (approved):
👍

jeddy3:
Changelog:

> - Fixed: `color-hex-case` false negatives for css-in-js object notation ([#5101](https://github.com/stylelint/stylelint/pull/5101)).

@Dru89 Thanks again for your first two pull request. Excellent job!

PR #5106, opened 2021-01-13
Fix false negatives for css-in-js object notation in color-hex-length

Dru89:
Continuing on the change made in #5101 (for issue #4826), this updates
`color-hex-length` to switch to using `postcss-value-parser` because
`style-search` returns false positives when used with `css-in-js` object
notation.

Unit tests have also been added to check the `css-in-js` syntax for this
rule.

### DIFF
--- a/lib/rules/color-hex-length/index.js
+++ b/lib/rules/color-hex-length/index.js
@@ -120,8 +120,6 @@ function getValue(decl) {
 function setValue(decl, value) {
 	if (decl.raws.value) decl.raws.value.raw = value;
 	else decl.value = value;
-
-	return decl;
 }
 
 rule.ruleName = ruleName;
--- a/lib/rules/color-hex-length/index.js
+++ b/lib/rules/color-hex-length/index.js
@@ -2,10 +2,11 @@
 
 'use strict';
 
-const blurFunctionArguments = require('../../utils/blurFunctionArguments');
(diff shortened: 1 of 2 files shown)


lib/rules/color-hex-length/index.js
```diff
@@ -117,12 +105,23 @@ function longer(hex) {
...
+function getValue(decl) {
+	return decl.raws.value ? decl.raws.value.raw : decl.value;
+}
+
+function setValue(decl, value) {
+	if (decl.raws.value) decl.raws.value.raw = value;
+	else decl.value = value;
 
-	return stringStart + replaceString + stringEnd;
+	return decl;
```
jeddy3, line 124:
```suggestion
```

We can remove this redundant return as per the other pull request.

At some point later down the line, we should refactor to move at least the `getValue` and `setValue` functions out to `/utils`. For now, let's duplicate them as you are doing, though.

jeddy3 (changes requested):
@Dru89 Fantastic. Besides one minor request, this looks good to me right off the bat!

hudochenkov (approved):
👍

jeddy3 (approved):

jeddy3:
Changelog: 

> - Fixed: `color-hex-length` false negatives for css-in-js object notation ([#5106](https://github.com/stylelint/stylelint/pull/5106)).

PR #5005, opened 2020-10-25
Document all CLI parameters in one place

hudochenkov:
It's inconvenient to find documentation for a CLI parameters, because all parameters are split into two pages: [CLI](https://stylelint.io/user-guide/usage/cli) and [Options](https://stylelint.io/user-guide/usage/options).

This pull request introduce some information duplication, however it makes finding description of a CLI parameter easier. As they all gathered, where is is expected — on a CLI page.

Issue #4843, opened 2020-06-25
Add customSyntax option to configuration object


DimaNazdratenko:
It's resolve the problem with integrate stylelint with different languages (for example like Stylus) with many IDEs like WebStorm. 

For example in VS Code there is an option for custom syntax option for stylelint, but in another like WebStorm - NO.

jeddy3:
@DimaNazdratenko Thanks for the request and for using the template.

I thought the same thing when I first encountered @ota-meshi's stylus plugin. 

I can't remember the historical reasons why some things live inside the configuration object, others in options and others as CLI flags. However, I think we should make it consistent i.e. everything that can live in the configuration object should and CLI flags can be used to augment this object. I believe this is how [Jest works](https://jestjs.io/docs/en/jest-object). Eventually, we'd then be able to remove options, which will simplify both the experience for the user and the stylelint codebase.

This would close:

- https://github.com/stylelint/stylelint/issues/3860
- https://github.com/stylelint/stylelint/issues/3859
- https://github.com/stylelint/stylelint/issues/3858

I help us achieve https://github.com/stylelint/stylelint/issues/3128.

Adding `customSyntax` to the configuration object would be part of that work.

I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

PR #5126, opened 2021-02-04
Allow disable reporting to be controlled from the config file

### LINKED ISSUE #3858: Add support for reportNeedlessDisables to the configuration object

I always want to use the `--report-needless-disables` flag, but it's a chore having to use it for every `$ stylelint` invocation.

I would like the ability to specify the CLI flag as a setting in the Stylelint config. That way I can put it in my shareable config and I don't have to specify it each time. The setting could be named `reportNeedlessDisables`.

nex3:
This will also make it easier to control disables on a file-by-file basis
once #3128 is implemented. It will also make it possible to add 
finer-grained configuration to these rules in the future.

Closes #3858


lib/needlessDisables.js
```diff
@@ -72,7 +78,7 @@ module.exports = function (results) {
 
 				// Only emit a warning if this range's comment isn't useful for this rule.
 				// For the special rule "all", only emit a warning if it's not useful for
-				// *any* // rules, becuase it covers all of them.
+				// *any* rules, becuase it covers all of them.
```
jathak, line 81:
```suggestion
				// *any* rules, because it covers all of them.
```

jathak (approved):

vankop (approved):

docs/user-guide/configure.md
```diff
@@ -168,6 +168,84 @@ You can set the default severity level for all rules that do not have a severity
 }
 ```
 
+## `ignoreDisables`
+
+Ignore `styleline-disable` (e.g. `/* stylelint-disable block-no-empty */`) comments.
```
jeddy3, line 173:
```suggestion
Ignore `stylelint-disable` (e.g. `/* stylelint-disable block-no-empty */`) comments.
```

docs/user-guide/configure.md
```diff
@@ -168,6 +168,84 @@ You can set the default severity level for all rules that do not have a severity
...
+
+```json
+{
+  "ignoreDisables": true
+}
+```
+
+## `reportNeedlessDisables`
+
+Emit errors for `stylelint-disable` comments that don't include actually match
```
jeddy3, line 185:
```suggestion
Emit errors for `stylelint-disable` comments that don't match
```

docs/user-guide/configure.md
```diff
@@ -168,6 +168,84 @@ You can set the default severity level for all rules that do not have a severity
...
+any lints that need to be disabled.
```
jeddy3, line 186: Let's remove the hard linebreak to be consistent with the rest of the file

docs/user-guide/configure.md
```diff
@@ -168,6 +168,84 @@ You can set the default severity level for all rules that do not have a severity
...
+```json
+{
+  "reportNeedlessDisables": true
+}
+```
+
+## `reportInvalidScopeDisables`
+
+Emit errors for `stylelint-disable` comments that don't match rules that are
+specified in the configuration object.
```
jeddy3, line 199:
```suggestion
Emit errors for `stylelint-disable` comments that don't match rules that are specified in the configuration object.
```

jeddy3 (changes requested):
@nex3 Thank you. It's amazing to see this go in. As you said, it's the first step towards adding the `overrides` configuration property (which opens up so many possibilities). It's also the first step towards being able to [configure everything](https://github.com/stylelint/stylelint/issues/4843#issuecomment-650137005) from within the configuration object, rather than a hodgepodge of properties, parameters and flags.

I've requested some minor documentation changes.

At the start of last year we [restructured](https://github.com/stylelint/stylelint/pull/4548) and [rewrote](https://github.com/stylelint/stylelint/pull/4561) the majority of the docs. This included removing a lot of duplication where API parameters, CLI flags and configuration properties had fallen out of sync. 

I'm aware this pull request reintroduces duplication, but I suggest we roll with it for now. I'll also merge in https://github.com/stylelint/stylelint/pull/5005, which duplicates the CLI flags. Once everything is configurable within the configuration object, I think we'll be better positioned to revisit the documentation and find a structure that both avoids duplication and is intuitive for users.

jeddy3 (approved):
Thanks for making the requested changes.

LGTM!

jeddy3:
> - Added: `ignoreDisables`, `reportNeedlessDisables`, `reportInvalidScopeDisables` and `reportDescriptionlessDisables` configuration object properties ([#5126](https://github.com/stylelint/stylelint/pull/5126)).

Issue #5129, opened 2021-02-06
Fix false negatives for CSS Variables in alpha-value-notation


jeddy3:
> Clearly describe the bug

There are false negatives for CSS Variables that span multiple parameters in `alpha-value-notation`, e.g.

```css
a { color: hsl(var(--hsl) / 50%) }
```

As an aside, this makes me realise how CSS Variables make static analysis less comprehensive. In the past, we were confident that our rules would catch everything, but with CSS Variables that is no longer the case. CSS Variables muddy the waters every time we use the value parser to find (see the `find*` utilities) a particular value, e.g. a font-size within the `font` shorthand property or an `<alpha-value>` within a colour function. A CSS Variable value is only known at runtime when the browser resolves it based on the cascade and DOM.

The best we can do is narrow down the possibilities of what a particular node could represent.

> Which rule, if any, is the bug related to?

`alpha-value-notation`

> What code is needed to reproduce the bug?

```css
a { color: hsl(var(--hsl) / 0.5) }
```

> What stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "alpha-value-notation": "percentage"
  }
}
```

> Which version of stylelint are you using?

`13.9.0`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

API via [Demo](https://stylelint.io/demo#N4Igxg9gJgpiBcICGACYLIBsICd4oAsBnTACgDckdSBaGgMwggEoUB6FABgDoBWZgNwoAvgB0AdiAA0IIgE9xAFyQAPBOCJFp4COPoBLAObrgElClEgcAV0wwil-KfHnzlpJgAOBJDUqZrGBpxCGVFfV1HCxBPGBwwGCUkQxhLMxEJYRBhIA)

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

No.

> What did you expect to happen?

The following warnings to be flagged:

```shell
test.css
2:4    Expected 0.5 to be 50% (alpha-value-notation)
```

> What actually happened (e.g. what warnings or errors did you get)?

No warnings.

PR #5130, opened 2021-02-06
Fix false negatives for CSS variables in alpha-value-notation

jeddy3:

Closes https://github.com/stylelint/stylelint/issues/5129

I ran into this bug while preparing a new release of [stylelint-scales](https://github.com/signal-noise/stylelint-scales).

As mentioned in https://github.com/stylelint/stylelint/issues/5129, the best we can do is narrow down the possibilities of what a particular node could represent. This pull request uses in the slash within modern colour function notation as a fallback to find the `<alpha-value>`.

### DIFF
--- a/lib/rules/alpha-value-notation/__tests__/index.js
+++ b/lib/rules/alpha-value-notation/__tests__/index.js
@@ -47,6 +47,9 @@ testRule({
 			code: 'a { color: hsla(var(--hue), var(--sat), var(--sat), var(--alpha)) }',
 		},
 		{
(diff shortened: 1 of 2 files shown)


malsf21 (approved):
LGTM! The `/* comment */` test case is a good one!

m-allanson (approved):
👍

jeddy3:
> - Fixed: `alpha-value-notation` false negatives for CSS Variables ([#5130](https://github.com/stylelint/stylelint/pull/5130)).

PR #5151, opened 2021-02-17
fix rule doc on declaration-property-value-allowed

evrom:
the patterns `a { background: pink; }` and `a { color: pink; }` are both considered violations and not considered violations. See https://github.com/stylelint/stylelint/tree/64078f5b7fb7cdf892a32840ee9a9c906c7c5a1e/lib/rules/declaration-property-value-allowed-list#options

Not only does this not make sense, but it is not consistent with the tests
https://github.com/stylelint/stylelint/blob/64078f5b7fb7cdf892a32840ee9a9c906c7c5a1e/lib/rules/declaration-property-value-allowed-list/__tests__/index.js#L26-L31
which checks accepts only the color green, not pink.

No, it's self-explanatory.

### DIFF
--- a/lib/rules/declaration-property-value-allowed-list/README.md
+++ b/lib/rules/declaration-property-value-allowed-list/README.md
@@ -62,11 +62,6 @@ The following patterns are _not_ considered violations:
 
 <!-- prettier-ignore -->
 ```css
-a { color: pink; }
-```
-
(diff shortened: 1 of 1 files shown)


jeddy3 (approved):
@evrom Thanks! 

LGTM.

Issue #5013, opened 2020-10-31
Refactor to address suppressed ESLint no-shadow violations


jeddy3:
Ref: https://github.com/stylelint/stylelint/issues/4985

Our remained task is to fix all `// TODO: Issue #4985`, for example:

https://github.com/stylelint/stylelint/blob/6ad94ac7253a01c31b69c1a183eeb9d4daa5f104/lib/assignDisabledRanges.js#L110-L112

PR #5152, opened 2021-02-18
Resolves some ESLint no-shadow violations

mattxwang:

Part of #5013.

I wanted to take a crack at doing some of the work for this PR, but I also wanted to make sure that I'm taking the right approach! Within some of these commits, I try a few different approaches - I just want to make sure that I'm doing things properly/fitting conventions.

* some of the are trivial renamings (ex `root` -> `rootNode`)
* some of them are logical renmaings (ex `output` -> `accum` for a reducer)
* some times, I move the function out of scope, if there's no dependence/reliance and the function seems bite-sized enough 

Any thoughts here? More than happy to change my strategy on this. I can gradually work through the rest of the cases over several PRs (not to make a review nightmare!).

### DIFF
--- a/lib/formatters/stringFormatter.js
+++ b/lib/formatters/stringFormatter.js
@@ -195,9 +195,7 @@ module.exports = function (results) {
 
 	output += deprecationsFormatter(results);
 
-	// TODO: Issue #4985
-	// eslint-disable-next-line no-shadow
-	output = results.reduce((output, result) => {
+	output = results.reduce((accum, result) => {
 		// Treat parseErrors as warnings
 		if (result.parseErrors) {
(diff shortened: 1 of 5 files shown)


ybiquitous (approved):
Thanks. LGTM 👍

jeddy3 (approved):
I've nothing to add. As @ybiquitous is happy with your approaches, I suggest rolling with them for the rest of the cases.

@mattxwang Thanks for this, btw! It's always great to see incremental improvements to the code base. It's six years old in places!

mattxwang:
Great, thanks for the quick reviews! I'll knock out more of these `no-shadow` violations in bite-sized PRs over the next few weeks.

PR #5165, opened 2021-02-22
Add ignore: [single-declaration] to declaration-block-trailing-semicolon

### LINKED ISSUE #5159: Add except: ["single-declaration"] to declaration-block-trailing-semicolon
allow `a { background: orange }` when declaration-block-trailing-semicolon is set to `"always"`
by introducing an except option that accepts `"single-declaration"`
i.e. if the block has 2 or more declarations, the last declaration still needs a `;`

e.g.
```json
        "declaration-block-trailing-semicolon": [
          "always",
          { "ignore": ["single-declaration"] }
        ],
```

It's useful for declaration blocks that **won't ever be expanded**:
```css
body { -webkit-animation: bugfix infinite 1s; }

/* from and to blocks will never change */
@-webkit-keyframes bugfix { from { padding: 0 } to { padding: 0 } }

### PR DESCRIPTION
**Pull request by Mouvedia**

Closes #5159

It cannot be named `"except"` because it would require `foo { property: value; }` to be valid for `"never"` which is useless/stupid. Or do you imagine a use case for it that would warrant it?

Should it be supported just for consistency?
**Can an option be exclusive to `"always"`?** If so do you have an example?

### DIFF
--- a/lib/rules/declaration-block-trailing-semicolon/README.md
+++ b/lib/rules/declaration-block-trailing-semicolon/README.md
@@ -88,3 +88,38 @@ a { color: pink }
 ```css
 a { background: orange; color: pink }
 ```
+
+## Optional secondary options
+
+### `ignore: ["single-declaration"]`
+
+#### `"single-declaration"`
+
+The following patterns are _not_ considered violations:
+
+<!-- prettier-ignore -->
+```css
+foo { property: value }
+```
+
+<!-- prettier-ignore -->
+```css
+foo { property: value; }
+```
+
+<!-- prettier-ignore -->
+```css
+@keyframes name { from { property: 0 } to { property: 1; } }
+```
+
+With `"always"`:
+
+The following pattern is _still_ considered a violation:
+
+<!-- prettier-ignore -->
+```css
+foo {
+  property: value;
+  bar: qux
+}
+```
--- a/lib/rules/declaration-block-trailing-semicolon/README.md
+++ b/lib/rules/declaration-block-trailing-semicolon/README.md
@@ -93,33 +93,16 @@ a { background: orange; color: pink }
 
 ### `ignore: ["single-declaration"]`
 
-#### `"single-declaration"`
+Ignore declaration blocks that contain a single declaration.
 
 The following patterns are _not_ considered violations:
 
 <!-- prettier-ignore -->
 ```css
-foo { property: value }
+a { color: pink }
 ```
 
 <!-- prettier-ignore -->
 ```css
-foo { property: value; }
-```
-
-<!-- prettier-ignore -->
-```css
-@keyframes name { from { property: 0 } to { property: 1; } }
-```
-
-With `"always"`:
-
-The following pattern is _still_ considered a violation:
-
-<!-- prettier-ignore -->
-```css
-foo {
-  property: value;
-  bar: qux
-}
+a { color: pink; }
 ```
--- a/lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
+++ b/lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
@@ -69,15 +69,15 @@ testRule({
 
 	accept: [
 		{
-			code: 'foo { property: value }',
+			code: 'a { color: pink }',
 			description: 'single declaration without trailing semicolon',
 		},
 		{
-			code: 'foo { property: value; }',
+			code: 'a { color: pink; }',
 			description: 'single declaration with trailing semicolon',
 		},
 		{
-			code: '@keyframes name { from { property: 0 } to { property: 1; } }',
+			code: '@keyframes foo { from { top: 0px } to { top: 1px; } }',
 			description: 'inconsistent case (with and without)',
 		},
 	],
@@ -100,15 +100,15 @@ testRule({
 
 	accept: [
 		{
-			code: 'foo { property: value }',
+			code: 'a { color: pink }',
 			description: 'single declaration without trailing semicolon',
 		},
 		{
-			code: 'foo { property: value; }',
+			code: 'a { color: pink; }',
 			description: 'single declaration with trailing semicolon',
 		},
 		{
-			code: '@keyframes name { from { property: 0 } to { property: 1; } }',
+			code: '@keyframes foo { from { top: 0px } to { top: 1px; } }',
 			description: 'inconsistent case (with and without)',
 		},
 	],
--- a/lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
+++ b/lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
@@ -64,6 +64,58 @@ testRule({
 
 testRule({
 	ruleName,
+	config: ['always', { ignore: 'single-declaration' }],
+	fix: true,
+
+	accept: [
+		{
+			code: 'foo { property: value }',
+			description: 'single declaration without trailing semicolon',
+		},
+		{
+			code: 'foo { property: value; }',
+			description: 'single declaration with trailing semicolon',
+		},
+		{
+			code: '@keyframes name { from { property: 0 } to { property: 1; } }',
+			description: 'inconsistent case (with and without)',
+		},
+	],
+
+	reject: [
+		{
+			code: 'a { background: orange; color: pink }',
+			fixed: 'a { background: orange; color: pink; }',
+			description: 'multi declaration block without trailing semicolon',
+			message: messages.expected,
+			line: 1,
+			column: 35,
+		},
+	],
+});
+
+testRule({
+	ruleName,
+	config: ['never', { ignore: ['single-declaration'] }],
+
+	accept: [
+		{
+			code: 'foo { property: value }',
+			description: 'single declaration without trailing semicolon',
+		},
+		{
+			code: 'foo { property: value; }',
+			description: 'single declaration with trailing semicolon',
+		},
+		{
+			code: '@keyframes name { from { property: 0 } to { property: 1; } }',
+			description: 'inconsistent case (with and without)',
+		},
+	],
+});
+
+testRule({
+	ruleName,
 	config: ['never'],
 	fix: true,
 
--- a/lib/rules/declaration-block-trailing-semicolon/index.js
+++ b/lib/rules/declaration-block-trailing-semicolon/index.js
@@ -3,6 +3,7 @@
 'use strict';
 
 const hasBlock = require('../../utils/hasBlock');
+const optionsMatches = require('../../utils/optionsMatches');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
 const validateOptions = require('../../utils/validateOptions');
@@ -14,12 +15,23 @@ const messages = ruleMessages(ruleName, {
 	rejected: 'Unexpected trailing semicolon',
 });
 
-function rule(expectation, _, context) {
+function rule(expectation, options, context) {
 	return (root, result) => {
-		const validOptions = validateOptions(result, ruleName, {
-			actual: expectation,
-			possible: ['always', 'never'],
-		});
+		const validOptions = validateOptions(
+			result,
+			ruleName,
+			{
+				actual: expectation,
+				possible: ['always', 'never'],
+			},
+			{
+				actual: options,
+				possible: {
+					ignore: ['single-declaration'],
+				},
+				optional: true,
+			},
+		);
 
 		if (!validOptions) {
 			return;
@@ -54,10 +66,16 @@ function rule(expectation, _, context) {
 		});
 
 		function checkLastNode(node) {
+			const hasSemicolon = node.parent.raws.semicolon;
+			const ignoreSingleDeclaration = optionsMatches(options, 'ignore', 'single-declaration');
 			let message;
 
+			if (ignoreSingleDeclaration && node.parent.first === node) {
+				return;
+			}
+
 			if (expectation === 'always') {
-				if (node.parent.raws.semicolon) {
+				if (hasSemicolon) {
 					return;
 				}
 
@@ -75,7 +93,7 @@ function rule(expectation, _, context) {
 
 				message = messages.expected;
 			} else if (expectation === 'never') {
-				if (!node.parent.raws.semicolon) {
+				if (!hasSemicolon) {
 					return;
 				}
 

### CODE REVIEW

**Comment by jeddy3** [comment]
> Can an option be exclusive to "always"? If so do you have an example?

I can't think of one. If it's exclusive, then we've merged it into primary option. For example, `"always-unless-keyword"` in [`font-family-name-quotes`](https://stylelint.io/user-guide/rules/font-family-name-quotes).

Primary options and `except` secondary options avoid inconsistencies. For example: `ignore` would allow:

```css
@keyframes bugfix { from { padding: 0 } to { padding: 0; } }
```

The `ignore` secondary option is usually used when another rule is going to enforce consistency instead, e.g. [in a plugin](https://github.com/kristerkari/stylelint-scss#list-of-rules).

> Should it be supported just for consistency?

I think we support it for consistency, when catering to your use case.

And, just so I'm clear, you always want trailing semis except (i.e. never) when the declaration block as one declaration _regardless of whether the block spans multiple lines_?

Like so:

```css
a { background: pink; color: red; }

a {
  background: pink;
  color: red;
}

a { color: red }
a {
  color: red
}
```

Or do your single declaration blocks always appear within single-line blocks?, e.g.:

```css
a { color: red }
```

Mouvedia:
> Primary options and `except` secondary options avoid inconsistencies. For example: `ignore` would allow:
> 
> ```css
> @keyframes bugfix { from { padding: 0 } to { padding: 0; } }
> ```

Good point. **Ill fix that.**

> Or do your single declaration blocks always appear within single-line blocks?

If you check the examples in the README you will see that it works for multi-line blocks as well.
I did it this way just not to complicate things further. Having this granularity may be added at a later time if someone demands it.

> I think we support it for consistency, when catering to your use case.

**What about `{ "single-declaration": "never" }` instead?** (ignored/harmless for `:[ "never", {`)
And if someone insists on having `{ "single-declaration": "always" }` for `:[ "never", {` he will have to create another issue.

lib/rules/declaration-block-trailing-semicolon/index.js
```diff
@@ -54,28 +66,38 @@ function rule(expectation, _, context) {
 		});
 
 		function checkLastNode(node) {
+			const hasSemicolon = node.parent.raws.semicolon;
 			let message;
 
 			if (expectation === 'always') {
-				if (node.parent.raws.semicolon) {
+				const singleDeclarationNeverEnabled = optionsMatches(options, 'singleDeclaration', 'never');
+				const hasOneDeclaration = singleDeclarationNeverEnabled && node.parent.first === node;
```
Mouvedia, line 74: this variable needs to be renamed, for the rest everything can be reviewed

jeddy3:
I'd like to get a clear picture of your use case so that I'm sure we're going down the right path.

Would disallowing trailing semicolons inside single-line blocks cater to your use case, e.g:

```css
a { color: red }
a { background: pink; color: red }

a {
  color: red;
}
```

This would also apply to the example you gave:

```css
@keyframes bugfix { from { padding: 0 } to { padding: 0 } }
```

If so, then an `except: ["inside-single-line-block"]` secondary option, that inverts the primary option, would be consistent with other rules, e.g. [`declaration-empty-line-before`](https://github.com/stylelint/stylelint/tree/master/lib/rules/declaration-empty-line-before), and, hopefully, easy to implement.

Mouvedia:
> Would disallowing trailing semicolons inside single-line blocks cater to your use case

My personal use-case, would be single-line-block / ignore / for always only. That's not exactly what I have PRd though.

I added the consistent version which always require not having a trailing semicolon as `{ "singleDeclaration": "never" }`.
And complemented it with `{ "singleDeclaration": "ignore-always" }` which accept with _and_ without trailing semicolon for single declarations.

> …hopefully, easy to implement.

Like I said I didn't bother with single/multi line blocks for my second PR. It might be easier than I think; Ill have to check.

> `except: ["inside-single-line-block"]`

That would require `a { color: red; }` to be invalid for `never` which is undesirable/nonsense.

> If it's exclusive, then we've merged it into primary option. For example, "always-unless-keyword" in font-family-name-quotes.

"`always-unless-single-block`" wouldn't solve my use case because we would still need to have two variants:
-consistent/never
- inconsistent/ignore

> I'd like to get a clear picture of your use case

Ignore behaviour is desirable because sometimes you do know that this block may be expanded.
But the consistent case should also be available.

jeddy3:
Thanks, I think I understand now. 

You don't add trailing semicolons when you're sure the block won't expand.

For example, in the case of:

```css
@-webkit-keyframes bugfix { from { padding: 0 } to { padding: 0 } }
```

In that case, I think an `ignore: ["inside-single-line-block"]` option is best.  Like other `ignore` options, it'll make the rule more permissive so that [idiosyncratic patterns (or those connected to language extensions)](https://stylelint.io/user-guide/rules/about#about-rules) can handled by hand or in a plugin.

If there are particular patterns for when a block won't ever be expanded, you can write a plugin to enforce the correct use of the trailing semicolon inside single-line blocks.

Mouvedia:
> I think an ignore: ["inside-single-line-block"] option is best

I am alright with this but how do I convey that it's exclusive to `always`?
@jeddy3  Or do you want to evaluate `a { color: red; }` as _also_ valid for `:[ "never", { ignore: ["inside-single-line-block"] } ]`?

jeddy3:
> I am alright with this but how do I convey that it's exclusive to always?

It won't be exclusive to `always`. The rule will simply ignore single-line blocks when this option is used, regardless of the primary option.

> @jeddy3 Or do you want to evaluate a { color: red; } as also valid for :[ "never", { ignore: ["inside-single-line-block"] } ]?

Yes, it is also valid.

> There's another problem: it doesn't convey that it only matches if there's just one declaration.

It will ignore single-line blocks regardless of how many declarations there are. This will work for your use case, right?

The `ignore` option is used to make the rule more permissive. Trailing semicolons insde single-line blocks can then be handled by hand or via a plugin (which could take into account the number of declaration etc).

Mouvedia:
> It will ignore single-line blocks regardless of how many declarations there are. This will work for your use case, right?

Actually, it's not. In my experience most of this once-set never-touch block declarations have a single declaration.
That's what I requested.  
If it's somehow changed to mean `always` only for multi-line blocks this is a completely different requests and I am not interested in adding it.

Please check the README and tests and tell me what's wrong with it. I think it's a good compromise.

jeddy3:
> In my experience most of this once-set never-touch block declarations have a single declaration.

I think your original `ignore: ["single-declaration"]` will work best then.

> Please check the README and tests and tell me what's wrong with it. 

To keep the codebase sustainable, we tend to avoid adding code for idiosyncratic patterns. Instead, we add `ignore` options so that certain patterns are ignored by the built-in rules, and then be checked by a complementary plugin, e.g. [as how the stylelint-scss plugin does it](https://github.com/kristerkari/stylelint-scss/blob/master/docs/examples/if-else.md).

Mouvedia:
> I think your original ignore: ["single-declaration"] will work best then.

@jeddy3 done.

lib/rules/declaration-block-trailing-semicolon/README.md
```diff
@@ -88,3 +88,38 @@ a { color: pink }
...
+
+### `ignore: ["single-declaration"]`
+
+#### `"single-declaration"`
+
+The following patterns are _not_ considered violations:
+
+<!-- prettier-ignore -->
+```css
+foo { property: value }
```
jeddy3, line 102:
```suggestion
a { color: pink }
```

lib/rules/declaration-block-trailing-semicolon/README.md
```diff
@@ -88,3 +88,38 @@ a { color: pink }
...
+```
+
+<!-- prettier-ignore -->
+```css
+foo { property: value; }
```
jeddy3, line 107:
```suggestion
a { color: pink; }
```

lib/rules/declaration-block-trailing-semicolon/README.md
```diff
@@ -88,3 +88,38 @@ a { color: pink }
...
+```
+
+<!-- prettier-ignore -->
+```css
+@keyframes name { from { property: 0 } to { property: 1; } }
+```
+
```
jeddy3, line 114:
```suggestion
```

Let's remove this example to be consistent with other rules where we only show the simplest example of the option in effect.

lib/rules/declaration-block-trailing-semicolon/README.md
```diff
@@ -88,3 +88,38 @@ a { color: pink }
...
+The following pattern is _still_ considered a violation:
+
+<!-- prettier-ignore -->
+```css
+foo {
+  property: value;
+  bar: qux
+}
+```
```
jeddy3, line 125:
```suggestion
```


Let's remove this example to be consistent with other rules where we only show the rule being more permissive in the examples and, due to a limitation in how the website is built, the example won't be given the correct background colour.

lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
```diff
@@ -62,6 +62,58 @@ testRule({
...
 });
 
+testRule({
+	ruleName,
+	config: ['always', { ignore: 'single-declaration' }],
+	fix: true,
+
+	accept: [
+		{
+			code: 'foo { property: value }',
```
jeddy3, line 72:
```suggestion
			code: 'a { color: pink }',
```

lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
```diff
@@ -62,6 +62,58 @@ testRule({
...
+			description: 'single declaration without trailing semicolon',
+		},
+		{
+			code: 'foo { property: value; }',
```
jeddy3, line 76:
```suggestion
			code: 'a { color: pink; }',
```

lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
```diff
@@ -62,6 +62,58 @@ testRule({
...
+			description: 'single declaration with trailing semicolon',
+		},
+		{
+			code: '@keyframes name { from { property: 0 } to { property: 1; } }',
```
jeddy3, line 80:
```suggestion
			code: '@keyframes foo { from { top: 0px } to { top: 0px; } }',
```

lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
```diff
@@ -62,6 +62,58 @@ testRule({
...
+	],
+});
+
+testRule({
+	ruleName,
+	config: ['never', { ignore: ['single-declaration'] }],
+
+	accept: [
+		{
+			code: 'foo { property: value }',
```
jeddy3, line 103:
```suggestion
			code: 'a { color: pink }',
```

lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
```diff
@@ -62,6 +62,58 @@ testRule({
...
+			description: 'single declaration without trailing semicolon',
+		},
+		{
+			code: 'foo { property: value; }',
```
jeddy3, line 107:
```suggestion
			code: 'a { color: pink; }',
```

lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
```diff
@@ -62,6 +62,58 @@ testRule({
...
+			description: 'single declaration with trailing semicolon',
+		},
+		{
+			code: '@keyframes name { from { property: 0 } to { property: 1; } }',
```
jeddy3, line 111:
```suggestion
			code: '@keyframes foo { from { top: 0px } to { top: 0px; } }',
```

jeddy3 (changes requested):
@Mouvedia Thanks for reverting to the `ignore` option. The code looks good.

I've requested some minor changes to the docs and tests to align them with [our conventions](https://stylelint.io/developer-guide/rules#write-the-readme) and to be consistent with the rule's other examples and tests. The conventions help:
- us maintain the test code
- users when looking through lots of rules

lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
Mouvedia: `code: '@keyframes foo { from { top: 0px } to { top: 1px; } }',`

lib/rules/declaration-block-trailing-semicolon/__tests__/index.js
Mouvedia: ditto

jeddy3 (approved):
Thanks! LGTM.

lib/rules/declaration-block-trailing-semicolon/README.md
```diff
@@ -88,3 +88,21 @@ a { color: pink }
 ```css
 a { background: orange; color: pink }
 ```
+
+
+### `ignore: ["single-declaration"]`
+
+#### `"single-declaration"`
+
```
ybiquitous, line 97: This heading is duplicate, so it seems needless to me.

```suggestion
```

For example, see the following document:

https://github.com/stylelint/stylelint/blob/67ef0a31db46eb8b6a89e7b47d72c26d827b9395/lib/rules/declaration-block-no-duplicate-properties/README.md#optional-secondary-options

lib/rules/declaration-block-trailing-semicolon/README.md
```diff
@@ -88,3 +88,21 @@ a { color: pink }
 ```css
 a { background: orange; color: pink }
 ```
+
+
+### `ignore: ["single-declaration"]`
+
```
ybiquitous, line 95: How about add a brief description about this secondary option. For example:

```suggestion

Ignore the case when there is a single declaration within a block.
```

ybiquitous (commented):
I've left minor comments about the document, but this PR looks good to me. 👍

lib/rules/declaration-block-trailing-semicolon/README.md
jeddy3: Yes, let's remove this heading. We usually only add subheadings when there is more than one possible value for `ignore` and `except` options, [like so](https://github.com/stylelint/stylelint/blob/master/lib/rules/at-rule-empty-line-before/README.md#optional-secondary-options).

(The README `declaration-block-no-duplicate-properties` is out of alignment, in that it should use subheadings.)

lib/rules/declaration-block-trailing-semicolon/README.md
jeddy3: Good catch. 

Let's go with "Ignore declaration blocks that contain a single declaration" to be [consistent with the form we usually use](https://github.com/stylelint/stylelint/blob/master/lib/rules/at-rule-empty-line-before/README.md#ignore-after-comment-first-nested-inside-block-blockless-after-same-name-blockless-blockless-after-blockless).

lib/rules/declaration-block-trailing-semicolon/README.md
Mouvedia: > We usually only add subheadings when there is more than one possible value for ignore and except options, like so.

I did it exactly for that reason; to prepare such a case.
Ill remove it.

ybiquitous (approved):
Thanks. LGTM 👍

jeddy3:
Changelog entry added:

> - Added: `ignore: ["single-declaration"]` to `declaration-block-trailing-semicolon` ([#5165](https://github.com/stylelint/stylelint/pull/5165)).

PR #5167, opened 2021-02-23
Add named-grid-areas-no-invalid

### LINKED ISSUE #5155: Add named-grid-areas-no-invalid

Validating `grid-template-areas` property values. 

A rule for `grid-template-areas`

I  made a `grid-template-areas` validator. The code is here: https://github.com/evrom/grid-validator and a live version is here: https://gridvalidator.com/ . I can convert the code into a rule, but I want to make sure it is within the scope of features that stylelint wants in their project.

evrom:

Closes #5155

Adds rule to validate CSS grid areas to the spec defined here: https://drafts.csswg.org/css-grid/#grid-template-areas-property

### DIFF
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -17,6 +17,10 @@ Grouped first by the following categories and then by the [_thing_](http://apps.
 - [`font-family-no-duplicate-names`](../../../lib/rules/font-family-no-duplicate-names/README.md): Disallow duplicate font family names.
 - [`font-family-no-missing-generic-family-keyword`](../../../lib/rules/font-family-no-missing-generic-family-keyword/README.md): Disallow missing generic families in lists of font family names.
 
+### Named Grid Areas
+
+- [`named-grid-areas-no-invalid`](../../../lib/rules/named-grid-areas-no-invalid/README.md): Disallow invalid named grid areas.
+
 ### Function
 
 - [`function-calc-no-invalid`](../../../lib/rules/function-calc-no-invalid/README.md): Disallow an invalid expression within `calc` functions.
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -17,7 +17,7 @@ Grouped first by the following categories and then by the [_thing_](http://apps.
 - [`font-family-no-duplicate-names`](../../../lib/rules/font-family-no-duplicate-names/README.md): Disallow duplicate font family names.
 - [`font-family-no-missing-generic-family-keyword`](../../../lib/rules/font-family-no-missing-generic-family-keyword/README.md): Disallow missing generic families in lists of font family names.
 
-### Named Grid Areas
+### Named grid areas
 
 - [`named-grid-areas-no-invalid`](../../../lib/rules/named-grid-areas-no-invalid/README.md): Disallow invalid named grid areas.
 
--- a/lib/rules/named-grid-areas-no-invalid/README.md
+++ b/lib/rules/named-grid-areas-no-invalid/README.md
@@ -9,6 +9,7 @@ a { grid-template-areas:
       "b b b"; }
 /**   ↑
  *  This named grid area */
+```
 
 ## Options
 
--- a/lib/rules/named-grid-areas-no-invalid/README.md
+++ b/lib/rules/named-grid-areas-no-invalid/README.md
@@ -4,10 +4,11 @@ Disallow invalid named grid areas.
 
 <!-- prettier-ignore -->
 ```css
-a { color: pink; }···
-/**               ↑
- *  This whitespace */
-```
+a { grid-template-areas: 
+      "a a a"
+      "b b b"; }
+/**   ↑
+ *  This named grid area */
 
 ## Options
 
--- a/lib/rules/named-grid-areas-no-invalid/README.md
+++ b/lib/rules/named-grid-areas-no-invalid/README.md
@@ -1,0 +1,44 @@
+# named-grid-areas-no-invalid
+
+Disallow invalid named grid areas in `grid-template-areas`
+
+<!-- prettier-ignore -->
+```css
+a { color: pink; }···
+/**               ↑
+ *  This whitespace */
+```
+
+## Options
+
+### `true`
+
+The following patterns are considered violations:
+
+All strings must define the same number of cell tokens
+<!-- prettier-ignore -->
+```css
+grid-template-areas: "a a a"
+                     "b b b";
+```
+
+All strings must define at least one cell token
+<!-- prettier-ignore -->
+```css
+grid-template-areas: ""
+```
+
+If a named grid area spans multiple grid cells, but those cells do not form a single filled-in rectangle, the declaration is invalid.
+<!-- prettier-ignore -->
+```css
+grid-template-areas: "a a a"
+                     "b b a";
+```
+
+The following patterns are _not_ considered violations:
+
+<!-- prettier-ignore -->
+```css
+grid-template-areas: "a a a"
+                     "b b b";
+```
--- a/lib/rules/named-grid-areas-no-invalid/README.md
+++ b/lib/rules/named-grid-areas-no-invalid/README.md
@@ -15,7 +15,7 @@ a { color: pink; }···
 
 The following patterns are considered violations:
 
-All strings must define the same number of cell tokens
+All strings must define the same number of cell tokens.
 
 <!-- prettier-ignore -->
 ```css
--- a/lib/rules/named-grid-areas-no-invalid/README.md
+++ b/lib/rules/named-grid-areas-no-invalid/README.md
@@ -1,6 +1,6 @@
 # named-grid-areas-no-invalid
 
-Disallow invalid named grid areas in `grid-template-areas`
+Disallow invalid named grid areas.
 
 <!-- prettier-ignore -->
 ```css
--- a/lib/rules/named-grid-areas-no-invalid/README.md
+++ b/lib/rules/named-grid-areas-no-invalid/README.md
@@ -16,19 +16,22 @@ a { color: pink; }···
 The following patterns are considered violations:
 
 All strings must define the same number of cell tokens
+
 <!-- prettier-ignore -->
 ```css
 grid-template-areas: "a a a"
-                     "b b b";
+                     "b b b b";
 ```
 
 All strings must define at least one cell token
+
 <!-- prettier-ignore -->
 ```css
 grid-template-areas: ""
 ```
 
 If a named grid area spans multiple grid cells, but those cells do not form a single filled-in rectangle, the declaration is invalid.
+
 <!-- prettier-ignore -->
 ```css
 grid-template-areas: "a a a"
@@ -41,4 +44,9 @@ The following patterns are _not_ considered violations:
 ```css
 grid-template-areas: "a a a"
                      "b b b";
+```
+
+<!-- prettier-ignore -->
+```css
+grid-template-areas: none;
 ```
--- a/lib/rules/named-grid-areas-no-invalid/README.md
+++ b/lib/rules/named-grid-areas-no-invalid/README.md
@@ -20,39 +20,39 @@ All strings must define the same number of cell tokens.
 
 <!-- prettier-ignore -->
 ```css
-grid-template-areas: "a a a"
-                     "b b b b";
+a { grid-template-areas: "a a a"
+                         "b b b b"; }
 ```
 
-All strings must define at least one cell token
+All strings must define at least one cell token.
 
 <!-- prettier-ignore -->
 ```css
-grid-template-areas: ""
+a { grid-template-areas: "" }
 ```
 
 All named grid areas that spans multiple grid cells must form a single filled-in rectangle.
 
 <!-- prettier-ignore -->
 ```css
-grid-template-areas: "a a a"
-                     "b b a";
+a { grid-template-areas: "a a a"
+                         "b b a"; }
 ```
 
 The following patterns are _not_ considered violations:
 
 <!-- prettier-ignore -->
 ```css
-grid-template-areas: "a a a"
-                     "b b b";
+a { grid-template-areas: "a a a"
+                         "b b b"; }
 ```
 
 <!-- prettier-ignore -->
 ```css
-grid-template-areas: "a a a" "b b b";
+a { grid-template-areas: "a a a" "b b b"; }
 ```
 
 <!-- prettier-ignore -->
 ```css
-grid-template-areas: none;
+a { grid-template-areas: none; }
 ```
--- a/lib/rules/named-grid-areas-no-invalid/README.md
+++ b/lib/rules/named-grid-areas-no-invalid/README.md
@@ -48,5 +48,10 @@ grid-template-areas: "a a a"
 
 <!-- prettier-ignore -->
 ```css
+grid-template-areas: "a a a" "b b b";
+```
+
+<!-- prettier-ignore -->
+```css
 grid-template-areas: none;
 ```
--- a/lib/rules/named-grid-areas-no-invalid/README.md
+++ b/lib/rules/named-grid-areas-no-invalid/README.md
@@ -31,7 +31,7 @@ All strings must define at least one cell token
 grid-template-areas: ""
 ```
 
-If a named grid area spans multiple grid cells, but those cells do not form a single filled-in rectangle, the declaration is invalid.
+All named grid areas that spans multiple grid cells must form a single filled-in rectangle.
 
 <!-- prettier-ignore -->
 ```css
--- a/lib/rules/named-grid-areas-no-invalid/__tests__/index.js
+++ b/lib/rules/named-grid-areas-no-invalid/__tests__/index.js
@@ -54,25 +54,6 @@ testRule({
 
 	reject: [
 		{
-			code: 'a { grid-template-areas: foo "a a a" bar "b b b" baz; }',
-			warnings: [
-				{ message: messages.noInvalidSymbols('foo'), line: 1, column: 26 },
-				{ message: messages.noInvalidSymbols('bar'), line: 1, column: 38 },
-				{ message: messages.noInvalidSymbols('baz'), line: 1, column: 50 },
-			],
-		},
-		{
-			code: stripIndent`
-				a { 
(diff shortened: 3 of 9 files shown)


lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -0,0 +1,131 @@
...
+	return Array.from(names);
+}
+
+function findNotContiguousOrRectangular(areas) {
+	return Array.from(
+		new Set(namedAreas(areas).filter((name) => !isContiguousAndRectangular(areas, name))),
+	);
+}
+
+function removeComment(value) {
```
evrom, line 76: I didn't know if there was already a way to remove comments from values. If there is, I will use that instead

lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -0,0 +1,131 @@
...
+
+const messages = ruleMessages(ruleName, {
+	noRows: () => 'no rows defined',
+	notRectangular: () => 'template is not rectangular',
+	notContiguousOrRectangular: (areaNames) =>
+		`The following areas are not contiguous or rectangular ${areaNames.join(', ')}`,
+});
+
+function areasArray(gridTemplateString) {
+	const rows = Array.from(gridTemplateString.matchAll(/["']([^"']*)["']/g)).map(
```
evrom: `.matchAll` is not supported by node version 10 in the test matrix. A new commit replaces this with `match` and a `while` loop. If there is a better way than this, let me know.

lib/rules/named-grid-areas-no-invalid/__tests__/index.js
```diff
@@ -0,0 +1,57 @@
...
+			column: 7,
+		},
+		{
+			code: 'div { grid-template-areas: ""; }',
+			message: messages.noRows(),
+			line: 1,
+			column: 7,
+		},
+	],
+});
```
jeddy3, line 57: Can we add two positions tests, e.g.:

```css
a { grid-template-areas: /* comment */ ""; }
```

```css
a { grid-template-areas: 
   ""; }
```

We'll likely need to make use of the value parser's `sourceIndex` value to get the correct positions.

The [`alpha-value-notation` rule](https://github.com/stylelint/stylelint/blob/master/lib/rules/alpha-value-notation/index.js) has an example of this.

lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -0,0 +1,137 @@
...
+	);
+}
+
+function removeComment(value) {
+	const regex = /\s*(?!<")\/\*[^*]+\*\/(?!")\s*/g;
+
+	return value.replace(regex, '');
+}
+
+function rule() {
```
jeddy3, line 88: Let's move this function above the others to be consistent with most other rules. We should be able to get the gist of what a rule is doing, without needing to see the smaller building blocks.

lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -0,0 +1,137 @@
...
+	return (root, result) => {
```
jeddy3, line 89: We'll need the validate the options, [like so](https://github.com/stylelint/stylelint/blob/1d24cf1fad9308adf63fee9bbe36225929afe4f2/lib/rules/color-no-invalid-hex/index.js#L18-L24).

lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -0,0 +1,137 @@
...
+		root.walkDecls('grid-template-areas', (decl) => {
+			const value = decl.value;
+
+			if (value === 'none') {
+				return;
+			}
+
+			const valueWithoutComment = removeComment(value);
+
+			const areas = areasArray(valueWithoutComment);
```
jeddy3, line 99: We can replace these with a call to the value parser and a `findAreas(nodes)` function.

jeddy3 (changes requested):
@evrom Thank you for contributing!

I've requested a few changes. We'll want to [make use of the value parser](https://stylelint.io/developer-guide/rules#construct-specific-parsers). This is probably the biggest revision. I've also requested a few changes to align the rule with our conventions so its consistent with other rules.

Once these are in place, it'll be easier to review the other workings of the rule.

(As this rule is candidate for the recommended config, we'll want to be comphensive in our tests as it's likely this rule will be run against a lot of code bases.)

lib/rules/named-grid-areas-no-invalid/__tests__/index.js
evrom: Do tests need to use only `a` selector as well, if the selector is not being tested (like in this rule)?

lib/rules/named-grid-areas-no-invalid/__tests__/index.js
```diff
@@ -12,14 +13,21 @@ testRule({
 			code: 'a { grid-template-areas: "a a a" "b b b"; }',
 		},
 		{
-			code: `div { grid-template-areas: "head head"
-"nav  main"
-"nav  foot";}`,
+			code: stripIndent`
```
evrom, line 16: is this correct? let me know. https://github.com/stylelint/stylelint/pull/5167#discussion_r580942640

lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -0,0 +1,150 @@
...
+const declarationValueIndex = require('../../utils/declarationValueIndex');
+const report = require('../../utils/report');
+const ruleMessages = require('../../utils/ruleMessages');
+const validateOptions = require('../../utils/validateOptions');
+const valueParser = require('postcss-value-parser');
+
+const ruleName = 'named-grid-areas-no-invalid';
+
+const messages = ruleMessages(ruleName, {
+	noEmptyRows: () => 'empty row',
```
jeddy3, line 15: Let's make these:
- consistent with other rules
- consistent with the spec

Like so:

- `expectedToken` - `Expected cell token within string`
- `expectedSameNumber` - `Expected same number of cell tokens`
- `expectedNoRectangle` - `Expected single filled-in rectangle for "${name}"`

This breaks with our existing convention of using `rejected` for `*-no-*` rules, but I think we made a mistake by introducing double negatives.

jeddy3 (changes requested):
@evrom Thank you for making all those changes! Things are looking good and I think the rule is taking shape.

I've requested some more changes.

We should focus the rule on just this part of the spec:

> All strings must define the same number of cell tokens (named cell tokens and/or null cell tokens), and at least one cell token, or else the declaration is invalid. If a named grid area spans multiple grid cells, but those cells do not form a single filled-in rectangle, the declaration is invalid.

i.e.:

- All strings must define at least one cell token.
- All strings must define the same number of cell tokens.
- All named grid areas that span multiple grid cells must form a single filled-in rectangle.

A few of the requested changes are around doing this.

Does that sound good to you? It may deviate from your original validator, though.

lib/rules/named-grid-areas-no-invalid/index.js
evrom: we can change the name of the rule, if there is a better name for this rule.

jeddy3: Rule names follow [a two-part naming convention](https://stylelint.io/user-guide/rules/about#names). 

We've found that [focused rules work best](https://stylelint.io/user-guide/rules/about#about-rules).

We could further focus this rule on `strings` as that is what the above validations are referencing, i.e:

```json
{
  "rules": {
    "named-grid-areas-string-no-invalid": true
  }
}
```

Where `named-grid-areas-string` is the _thing_ and `no-invalid` the _check_. And where `string` and `named-grid-areas` come from [the spec](https://drafts.csswg.org/css-grid/#grid-template-areas-property).

What do you think?

lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -0,0 +1,150 @@
...
+				parsedValue.nodes[0].value === 'none'
+			) {
+				return;
+			}
+
+			let areas = [];
+			let reportSent = false;
+
+			parsedValue.walk((node) => {
+				if (node.type !== 'string' && node.type !== 'space') {
```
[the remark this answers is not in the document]
evrom: I will remove the test cases as well that test for this

lib/rules/named-grid-areas-no-invalid/index.js
evrom: I think this is a bit more clear:
`expectedSameNumber` - `Expected same number of cell tokens in each string`

evrom: should it be `expectedRectangle` instead of `expectedNoRectangle`?

evrom: Is there another member of stylelint who can comment on what the name of the rule `named-grid-areas-no-invalid'` should be?

lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -60,14 +58,14 @@ function rule(actual) {
 			if (reportSent) return;
 
 			if (!isRectangular(areas)) {
-				complain(messages.notRectangular());
+				complain(messages.expectedSameNumber());
```
evrom, line 61: Should I add a `return`? is it worth checking named areas are rectangular and contiguous if the the grid itself is not rectangular?

lib/rules/named-grid-areas-no-invalid/__tests__/index.js
```diff
@@ -0,0 +1,138 @@
...
+			code: 'div { grid-template-areas: "a a a" "b b"; }',
+			message: messages.notRectangular(),
+			line: 1,
+			column: 7,
+		},
+		{
+			code: 'div { grid-template-areas: "a a a" "b b a"; }',
+			message: messages.notContiguousOrRectangular(['a']),
+			line: 1,
+			column: 7,
```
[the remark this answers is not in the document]
evrom: I think this is working in the latest commit, but check it out to make sure it is right.

lib/rules/named-grid-areas-no-invalid/index.js
evrom: I've decided to make this change to `return` if strings do not not have same number of tokens (grid not rectangular). Let me know if you think this is bad in anyway.

lib/rules/named-grid-areas-no-invalid/index.js
evrom: the name of the rule is the only thing that is very difficult to change later on.

jeddy3: > Is there another member of stylelint who can comment on what the name of the rule `named-grid-areas-no-invalid` should be?

Yes. Pull requests require two reviews to be merged.

> the name of the rule is the only thing that is very difficult to change in the future.

Yes, we try to get them right the first time because of this. Hence the documented conventions.

lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -0,0 +1,129 @@
...
+
+			if (value.toLowerCase().trim() === 'none') return;
+
+			const parsedValue = valueParser(value);
+
+			let areas = [];
+			let reportSent = false;
+
+			parsedValue.walk((node) => {
+				if (node.type === 'string') {
```
jeddy3, line 49: Let's return early when walking (unless reporting):

```js
if (type !== 'string') return;
if (value === '') {
 complain(..)
}
areas.push(..)
```

This will be consistent with other rules.

lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -0,0 +1,129 @@
...
+			const value = decl.value;
+
+			if (value.toLowerCase().trim() === 'none') return;
+
+			const parsedValue = valueParser(value);
+
+			let areas = [];
+			let reportSent = false;
+
+			parsedValue.walk((node) => {
```
jeddy3, line 48: Let's destructure:

```suggestion
			parsedValue.walk(({sourceIndex, type value}) => {
```

lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -0,0 +1,129 @@
...
+				report({
+					message,
+					node: decl,
+					index: declarationValueIndex(decl) + sourceIndex,
+					result,
+					ruleName,
+				});
+			}
+
+			const value = decl.value;
```
jeddy3, line 39: Let's destructure:

```suggestion
			const { value }  = decl;
```

lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -0,0 +1,129 @@
...
+
+function rule(actual) {
+	return (root, result) => {
+		const validOptions = validateOptions(result, ruleName, { actual });
+
+		if (!validOptions) {
+			return;
+		}
+
+		root.walkDecls('grid-template-areas', (decl) => {
```
jeddy3, line 28:
```suggestion
		root.walkDecls(/^grid-template-areas$/i, (decl) => {
```

Looking through the [commonly overlooked edge-cases](https://stylelint.io/developer-guide/rules#commonly-overlooked-edge-cases), we should account for uppercase properties.

Let's add one reject test for `a { GRID-TEMPLATE-AREAS: ""; }`

lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -0,0 +1,129 @@
...
+			function complain(message, sourceIndex = 0) {
```
jeddy3, line 29: As with the other building-block functions, let's drop this to the bottom of the scope.

jeddy3 (changes requested):
@evrom Thank you so much for making the changes. I think we're almost there! 

I've requested a couple of minor changes; an edge case, conventional stuff to align with other rules and some nitpicks.

The conventional stuff helps us manage the 150+ rules in stylelint. (We're slowly refactoring and aligning the rules as we go so you'll likely encounter inconsistencies if you look through them).

Lastly, we'll need to add the rules to [the list of rules](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules/list.md).

> Section: "Possible errors" -> "Named grid areas".

We can put the new section after [Font-family](https://stylelint.io/user-guide/rules/list#font-family).

jeddy3 (approved):
@evrom Once again, thanks for making the changes!

Looks good to me. 

It'll need a second pair of eye on it to merge.  We'll add the rule to the [`stylelint-config-recommended`](https://github.com/stylelint/stylelint-config-recommended), so it'll likely touch quite a lot of people's code.

evrom:
Sould I break out the functions like `isContiguousAndRectangular()` out of `index.js` and into a file in the same directory like `utils.js`? I have a bunch of unit tests for these function, and I was thinking it might be good to have them in `./__tests__/utils.js`

jeddy3:
Can do. I suggest putting them in [`lib/utils`](https://github.com/stylelint/stylelint/tree/master/lib/utils) with one function per file, to match the other utilities.

evrom:
I didn't add tests to some private (non-exported) functions in utils. I am not sure what the style is preferred in regards to testing private functions.

P.S. I just learned Rust recently, and found out about the whole debate regarding testing private functions. Apparently Rust by default does not support the idea of testing private functions, but has default workarounds for this.

jeddy3 (approved):
Refactor and tests LGTM, thanks!

jeddy3:
> P.S. I just learned Rust recently, 

Nice. It's interesting to see JS/TS tools being written the likes of Rust and Go, e.g. Deno's linter and formatter, and esbuild.

lib/rules/named-grid-areas-no-invalid/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+Disallow invalid named grid areas.
+
+<!-- prettier-ignore -->
+```css
+a { grid-template-areas: 
+      "a a a"
+      "b b b"; }
+/**   ↑
+ *  This named grid area */
```
ybiquitous, line 11: Are triple backticks missing at the end of the code block?

````suggestion
 *  This named grid area */
```
````

File: docs/user-guide/rules/list.md [hunk 18]
```diff
@@ -17,6 +17,10 @@ Grouped first by the following categories and then by the [_thing_](http://apps.
 - [`font-family-no-duplicate-names`](../../../lib/rules/font-family-no-duplicate-names/README.md): Disallow duplicate font family names.
 - [`font-family-no-missing-generic-family-keyword`](../../../lib/rules/font-family-no-missing-generic-family-keyword/README.md): Disallow missing generic families in lists of font family names.
 
+### Named Grid Areas
```
ybiquitous, line 20: For consistency with other headings, the following seems better to me:

```suggestion
### Named grid areas
```

lib/rules/named-grid-areas-no-invalid/utils/isRectangular.js
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ *
+ * @param {string[][]} areas
+ * @returns {boolean}
+ */
+function isRectangular(areas) {
+	const columnsPerRow = areas.map((row) => row.length);
+
+	return columnsPerRow.every((val, i, arr) => val === arr[0]);
```
ybiquitous, line 11: Does the following simplification make sense?

```suggestion
	return areas.every((row, i, arr) => row.length === arr[0].length);
```

ybiquitous (commented):
I've pointed out some minor details, but this looks mostly fine to me. 😃

evrom:
I am not sure why the test suite failed on Windows on Node.js v10 . The failing test appears to not have anything to do with the recent changes or the pull request as a whole.

jeddy3:
The tests intermittently fail for Node 10 on Windows. We should probably get to the bottom of it one day, or we can wait until we drop Node 10 stop soon in favour of going ESM :)

I've restarted the tests.

ybiquitous (approved):
Thanks. LGTM 👍

jeddy3:
@evrom Thank you for your contribution!

Changelog entry: 

> - Added: `named-grid-areas-no-invalid` rule ([#5167](https://github.com/stylelint/stylelint/pull/5167)).

I'll hopefully get a release out of stylelint and the two official configs this weekend.

PR #5183, opened 2021-03-05
Refactor getting and setting declaration values into utils

Dru89:
This change adds two utility functions, `getDeclarationValue` and `setDeclarationValue`, that replace a pattern found across many rule files. Extracting the `getValue` and `setValue` functions across the rule files was mentioned by @jeddy3 in #5106. (But if we want a specific issue to track this as well, I can still file one.)

I was about to tackle replacing another `style-search` based rule, but remembered that I'd probably want this utility function. And this change felt large enough to make on its own without also updating an additional rule.

### DIFF
--- a/lib/utils/setDeclarationValue.js
+++ b/lib/utils/setDeclarationValue.js
@@ -1,0 +1,20 @@
+'use strict';
+
+const _ = require('lodash');
+
+/** @typedef {import('postcss').Declaration} Declaration */
+
+/**
+ * @param {Declaration} decl
+ * @param {string} value
+ * @returns {Declaration} The declaration that was passed in.
+ */
+module.exports = function getDeclarationValue(decl, value) {
+	if (_.has(decl, 'raws.value')) {
+		_.set(decl, 'raws.value.raw', value);
+	} else {
+		decl.value = value;
+	}
+
+	return decl;
+};
--- a/lib/utils/setDeclarationValue.js
+++ b/lib/utils/setDeclarationValue.js
@@ -9,7 +9,10 @@ const _ = require('lodash');
  * @param {string} value
  * @returns {Declaration} The declaration that was passed in.
  */
-module.exports = function getDeclarationValue(decl, value) {
+module.exports = function setDeclarationValue(decl, value) {
+	// Lodash is necessary here because the declaration may not strictly adhere
+	// to the current version of PostCSS's Declaration interface.
+	// See also: https://github.com/stylelint/stylelint/pull/5183/files#r588047494
 	if (_.has(decl, 'raws.value')) {
 		_.set(decl, 'raws.value.raw', value);
 	} else {
--- a/lib/rules/alpha-value-notation/index.js
+++ b/lib/rules/alpha-value-notation/index.js
@@ -6,10 +6,12 @@ const _ = require('lodash');
 const valueParser = require('postcss-value-parser');
 
 const declarationValueIndex = require('../../utils/declarationValueIndex');
+const getDeclarationValue = require('../../utils/getDeclarationValue');
 const isStandardSyntaxValue = require('../../utils/isStandardSyntaxValue');
 const optionsMatches = require('../../utils/optionsMatches');
 const report = require('../../utils/report');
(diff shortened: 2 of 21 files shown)


lib/utils/setDeclarationValue.js
```diff
@@ -0,0 +1,20 @@
...
+/** @typedef {import('postcss').Declaration} Declaration */
+
+/**
+ * @param {Declaration} decl
+ * @param {string} value
+ * @returns {Declaration} The declaration that was passed in.
+ */
+module.exports = function getDeclarationValue(decl, value) {
+	if (_.has(decl, 'raws.value')) {
+		_.set(decl, 'raws.value.raw', value);
```
Dru89, line 14: This change uses lodash because `Declaration` doesn't strictly adhere to the version of `postcss.Declaration` that we use. `stylelint` uses PostCSS 7 and, as best as I can tell, these utilities were written so that support for migration to PostCSS 8 would be easier? Or perhaps the typescript definitions in PostCSS 7 are simply wrong.

The [`Declaration`](https://github.com/postcss/postcss/blob/12832f3d203474bd273bd06bd3b2407567bfe09e/lib/postcss.d.ts#L1195-L1222) interface in PostCSS 7 inherits from [`NodeBase`](https://github.com/postcss/postcss/blob/12832f3d203474bd273bd06bd3b2407567bfe09e/lib/postcss.d.ts#L588-L707), where [`raws`](https://github.com/postcss/postcss/blob/12832f3d203474bd273bd06bd3b2407567bfe09e/lib/postcss.d.ts#L599-L603) is defined as [`NodeRaws`](https://github.com/postcss/postcss/blob/12832f3d203474bd273bd06bd3b2407567bfe09e/lib/postcss.d.ts#L712-L748) and does not support a `value` property.

In PostCSS 8, there is a [`declaration.d.ts`](https://github.com/postcss/postcss/blob/4f4e2932fc97e2c117e1a4b15f0272ed551ed59d/lib/declaration.d.ts) file that has a [`DeclarationRaws`](https://github.com/postcss/postcss/blob/4f4e2932fc97e2c117e1a4b15f0272ed551ed59d/lib/declaration.d.ts#L3-L27) interface that *does* supply a type signature that meets the first definition, however.

I thought about creating some complicated `@typedefs`, but it seemed like we were already using lodash in a few places to get around this type signature problem. Let me know if you have any suggestions on how to define this better, though.

lib/utils/setDeclarationValue.js
```diff
@@ -0,0 +1,20 @@
...
+const _ = require('lodash');
+
+/** @typedef {import('postcss').Declaration} Declaration */
+
+/**
+ * @param {Declaration} decl
+ * @param {string} value
+ * @returns {Declaration} The declaration that was passed in.
+ */
+module.exports = function getDeclarationValue(decl, value) {
```
ybiquitous, line 12: Oh, this is a typo. 😅 

```suggestion
module.exports = function setDeclarationValue(decl, value) {
```

lib/utils/setDeclarationValue.js
ybiquitous: Great! 👏 
I think it would be nice if you could leave the reason as a code comment as below:

```js
// The reason to use Lodash here: ...
// See also: https://github.com/stylelint/stylelint/pull/5183/files#r588047494
if (_.has(decl, 'raws.value')) {
```

ybiquitous (approved):
I think this is a great improvement! 👏

ybiquitous:
I believe the upload error to Coverall should be resolved by retry:

https://github.com/stylelint/stylelint/pull/5183/checks?check_run_id=2037304546#step:7:9

jeddy3:
@Dru89 A great refactor, thanks!

LGTM, once @ybiquitous suggestions are in.

I've restarted the tests.

Dru89:
Thanks, y'all! I've updated the `setDeclarationValue` method to have the right name and add a comment about the use of lodash.

jeddy3 (approved):
LGTM, thanks!

ybiquitous (approved):
@Dru89 Thank you! 😄

PR #5186, opened 2021-03-06
Fix false positives for css-in-js in color-no-hex

Dru89:
`style-search` returns false positives when used with `css-in-js` object notation. This change continues with the discussion in #4826 and moves another test towards using `postcss-value-parser`.

### DIFF
--- a/lib/rules/color-no-hex/__tests__/index.js
+++ b/lib/rules/color-no-hex/__tests__/index.js
@@ -125,7 +125,7 @@ testRule({
 		{
 			code: `
 				export const s = styled.a({
-					background: 'url("#aabbcc")',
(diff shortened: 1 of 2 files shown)


jeddy3 (approved):
@Dru89 Fantastic, thanks.

LGTM from the get-go.

ybiquitous (approved):
LGTM, thanks! 😄

jeddy3:
> - Fixed: `color-no-hex` false positives for CSS-in-JS objection notation ([#5186](https://github.com/stylelint/stylelint/pull/5186)).

---

@Dru89 you're welcome to clone the stylelint repo directly, rather than work from a fork.

Issue #5192, opened 2021-03-08
Add support for dependencies to custom formatter flag


ThornWalli:
Current implementation, although using `require`, only provides relative or absolute file path.

https://github.com/stylelint/stylelint/blob/16ae9e4537975d1d755e430a813a74ddfb681cde/lib/cli.js#L348-L354

So it is not possible to import a formatter as dependency correctly.

**Not Correct** `stylelint "src/**/*.css" --custom-formatter node_modules/dependency/index.js`

**Correct** `stylelint "src/**/*.css" --custom-formatter dependency`

Topic was already mentioned here in the past. https://github.com/stylelint/stylelint/issues/2449

PR: #5193

ybiquitous:
I forgot about this issue, but this has been implemented with PR #7343 unintentionally.

### Confirmation

Using [`stylelint-codeframe-formatter`](https://www.npmjs.com/package/stylelint-codeframe-formatter).

`package.json`:

```json
{
  "type": "module",
  "dependencies": {
    "stylelint": "^16.2.0",
    "stylelint-codeframe-formatter": "^1.2.0"
  }
}
```

`stylelint.config.js`:

```js
export default {
	rules: {
		'block-no-empty': true
	}
};
```

`test.css`:

```css
a {}
```

Run:

```sh-session
$ npm i

$ npx stylelint "*.css" --custom-formatter=stylelint-codeframe-formatter
  test.css

  ✖ Unexpected empty block (block-no-empty) at test.css:1:3
> 1 | a {}
    |   ^^
  2 |

  1 error
```

Mouvedia:
We will have to edit the CHANGELOG retroactively to add that enhancement.

ybiquitous:
> We will have to edit the CHANGELOG retroactively to add that enhancement.

We had it already:
https://github.com/stylelint/stylelint/blob/34cdc97149234063c94301cd64ceb378b20a07bf/CHANGELOG.md?plain=1#L47

Mouvedia:
I meant a separate line that states that we now support the package name as an argument to `--custom-formatter`.

ybiquitous:
Ah, do you mean we should add an entry to the previous version (16.0.0) in CHANGELOG.md?
https://github.com/stylelint/stylelint/blob/34cdc97149234063c94301cd64ceb378b20a07bf/CHANGELOG.md?plain=1#L31

Mouvedia:
Yes that's what I meant by "retroactively".

PR #5193, opened 2021-03-08
fix(cli): fix custom formatter import

ThornWalli:
Until now it was not possible to import a `custom formatter` as dependency via the **CLI** call.
The specified path was handled either as a `relative` or `absolute` file path.
With the current customizations it is possible to import `formatter` as `dependency`.

For this the existence of the file path is checked first, if file does not exist path is treated as `dependency`.

The adjustment is particularly important if `formatter` are used as `dependency`.

Issue #5192

### DIFF
--- a/docs/user-guide/usage/options.md
+++ b/docs/user-guide/usage/options.md
@@ -52,7 +52,7 @@ Options are:
 - `unix`
 - `verbose`
 
-The `formatter` Node.js API option can also accept a function, whereas the `--custom-formatter` CLI flag accepts a path to a JS file exporting one. The function in both cases must fit the signature described in the [Developer Guide](../../developer-guide/formatters.md).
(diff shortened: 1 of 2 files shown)


ybiquitous:
@ThornWalli Thank you for the contribution! This change looks good to me. 👍 

We are very sorry for no tests for `--custom-formatter`, but could you please add them?

Perhaps, we can write the tests in `lib/__tests__/cli.test.js` as below:

```js
it('some description about --custom-formatter flag...', async () => {
	await cli([
		'--custom-formatter',
		'custom-formatter-example', //=> `lib/__tests__/__mocks__/custom-formatter-example.js`
		'--config',
		replaceBackslashes(path.join(fixturesPath, 'config-color-no-invalid-hex.json')),
		replaceBackslashes(path.join(fixturesPath, 'invalid-hex.css')),
	]);

	// assertions...
});

// `lib/__tests__/__mocks__/custom-formatter-example.js`
module.exports = () => {
	return 'Custom Formatter Example';
};
```

`custom-formatter-example` is just a mock, so you can write a mock name and content as you like.
(see also <https://jestjs.io/docs/manual-mocks>)

PR #5196, opened 2021-03-10
Fix husky configuration and remove needless packages

ybiquitous:
`husky` no longer requires `pinst` and `is-ci` since v5.1.2.
(see <https://github.com/typicode/husky/commit/84be6755c77ee81b14260765f2224e70f6bca8df>)

So, this change simplifies the `husky` configuration and removes the needless packages.
See also <https://typicode.github.io/husky>.

None.

None.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -3,4 +3,4 @@
 
 . "$(dirname "$0")/_/husky.sh"
 
-npx --no lint-staged
+npx --no-install lint-staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,6 @@
 #!/bin/sh
+[ -n "$CI" ] && exit 0
+
 . "$(dirname "$0")/_/husky.sh"
 
 npx --no lint-staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npx lint-staged
+npx --no lint-staged
--- a/package.json
+++ b/package.json
@@ -38,24 +38,17 @@
   "scripts": {
     "benchmark-rule": "node scripts/benchmark-rule.js",
     "format": "prettier . --write",
-    "postinstall": "is-ci || husky install",
     "jest": "jest",
     "lint": "npm-run-all --parallel lint:*",
     "lint:formatting": "prettier . --check",
     "lint:js": "eslint . --cache --max-warnings=0",
     "lint:md": "remark . --quiet --frail",
     "lint:types": "tsc",
-    "prepublishOnly": "pinst --disable",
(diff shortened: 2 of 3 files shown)


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npx lint-staged
+npx --no lint-staged
```
ybiquitous, line 4: [note] Use the local dependency. See <https://docs.npmjs.com/cli/v7/commands/npm-exec>.

package.json
```diff
@@ -38,25 +38,18 @@
   "scripts": {
     "benchmark-rule": "node scripts/benchmark-rule.js",
     "format": "prettier . --write",
-    "postinstall": "is-ci || husky install",
+    "prepare": "husky install",
```
ybiquitous, line 41: [note] The husky document uses `prepare` instead of `postinstall`:
<https://typicode.github.io/husky/#/?id=install>

package.json
```diff
@@ -38,25 +38,18 @@
...
     "release": "np",
     "pretest": "npm run lint",
     "test": "jest --coverage",
     "watch": "jest --watch"
   },
-  "husky": {
-    "hooks": {
-      "pre-commit": "lint-staged"
-    }
-  },
```
ybiquitous, line 59: [note] This is an old configuration.

package.json
jeddy3: Do we still need to do something to [disable hooks on the CI](https://typicode.github.io/husky/#/?id=disable-hooks-in-ci)?

Perhaps:

```
# .husky/pre-commit
# ...
[ -n "$CI" ] && exit 0
```

ybiquitous: Good catch! Fixed via 57eeb86d.

I think we now don't need the `CI` check for the `prepare` script via `is-ci`:

```json
{
  "scripts": {
    "prepare": "is-ci || husky install"
  }
}
```

If we would need it in the future, then let's install it again.

hudochenkov:
There is so much hustle with `husky`. Maybe we should switch to a simpler solution? https://github.com/toplenboren/simple-git-hooks

ybiquitous:
@hudochenkov Honestly, I also don't like husky v5 so much... 😓 
If `simple-git-hooks` is sufficient, I think it good to switch it. I'll look into it more.

ybiquitous:
`simple-git-hooks` is renaming. The old name is `simple-pre-commit`. It looks a bit unstable yet. 🙄 
https://github.com/toplenboren/simple-git-hooks/pull/21

jeddy3 (approved):
Changes look good to me.

We can shift to `simple-git-hooks` when it's released.

hudochenkov (approved):

ybiquitous:
Thank you for the feedback! 😄

PR #5202, opened 2021-03-18
Add no-invalid-position-at-import-rule

### LINKED ISSUE #5133: Add no-invalid-position-at-import-rule

```
:root {
	--bg: white;
}
@import './some.css';
```

Won't load some.css in chrome 88

A new rule that raises an error when @import is not at the top of the stylesheet.

chuuddo:

Closes #5133.

No, it's self-explanatory.

### DIFF
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -79,6 +79,7 @@ Grouped first by the following categories and then by the [_thing_](http://apps.
 - [`no-empty-source`](../../../lib/rules/no-empty-source/README.md): Disallow empty sources.
 - [`no-extra-semicolons`](../../../lib/rules/no-extra-semicolons/README.md): Disallow extra semicolons (Autofixable).
 - [`no-invalid-double-slash-comments`](../../../lib/rules/no-invalid-double-slash-comments/README.md): Disallow double-slash comments (`//...`) which are not supported by CSS.
+- [`no-invalid-position-at-import-rule`](../../../lib/rules/no-invalid-position-at-import-rule/README.md): Disallow invalid position @import rules within a stylesheet.
 
 ## Limit language features
 
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -79,7 +79,7 @@ Grouped first by the following categories and then by the [_thing_](http://apps.
 - [`no-empty-source`](../../../lib/rules/no-empty-source/README.md): Disallow empty sources.
 - [`no-extra-semicolons`](../../../lib/rules/no-extra-semicolons/README.md): Disallow extra semicolons (Autofixable).
 - [`no-invalid-double-slash-comments`](../../../lib/rules/no-invalid-double-slash-comments/README.md): Disallow double-slash comments (`//...`) which are not supported by CSS.
-- [`no-invalid-position-at-import-rule`](../../../lib/rules/no-invalid-position-at-import-rule/README.md): Disallow invalid position @import rules within a stylesheet.
+- [`no-invalid-position-at-import-rule`](../../../lib/rules/no-invalid-position-at-import-rule/README.md): Disallow invalid position `@import` rules within a stylesheet.
 
 ## Limit language features
 
--- a/lib/rules/no-invalid-position-at-import-rule/README.md
+++ b/lib/rules/no-invalid-position-at-import-rule/README.md
@@ -1,0 +1,49 @@
+# no-invalid-position-at-import-rule
+
+Disallow invalid position `@import` rules within a stylesheet.
+
+<!-- prettier-ignore -->
+```css
+a {}
+@import 'foo.css';
+/**   ↑
+  * Invalid position */
+```
+
+## Options
+
+### `true`
+
+The following patterns are considered violations:
+
+<!-- prettier-ignore -->
+```css
+a {}
+@import 'foo.css';
+```
+
+<!-- prettier-ignore -->
+```css
+@media print {}
+@import 'foo.css';
+```
+
+The following patterns are _not_ considered violations:
+
+<!-- prettier-ignore -->
+```css
+@import 'foo.css';
+a {}
+```
+
+<!-- prettier-ignore -->
+```css
+/* some comment */
+@import 'foo.css';
+```
+
+<!-- prettier-ignore -->
+```css
+@charset 'utf-8';
+@import 'foo.css';
+```
--- a/lib/rules/no-invalid-position-at-import-rule/README.md
+++ b/lib/rules/no-invalid-position-at-import-rule/README.md
@@ -8,6 +8,7 @@ a {}
 @import 'foo.css';
 /** ↑
   * This @import */
+```
 
 ## Options
 
--- a/lib/rules/no-invalid-position-at-import-rule/README.md
+++ b/lib/rules/no-invalid-position-at-import-rule/README.md
@@ -10,7 +10,7 @@ a {}
   * This @import */
 ```
 
-Any `@import` rules must precede all other valid at-rules and style rules in a style sheet (ignoring `@charset`), or else the `@import` rule is invalid.
+Any `@import` rules must precede all other valid at-rules and style rules in a stylesheet (ignoring `@charset`), or else the `@import` rule is invalid.
 
 ## Options
 
--- a/lib/rules/no-invalid-position-at-import-rule/README.md
+++ b/lib/rules/no-invalid-position-at-import-rule/README.md
@@ -10,6 +10,8 @@ a {}
   * This @import */
 ```
 
+Any `@import` rules must precede all other valid at-rules and style rules in a style sheet (ignoring `@charset`), or else the `@import` rule is invalid.
+
 ## Options
 
 ### `true`
--- a/lib/rules/no-invalid-position-at-import-rule/README.md
+++ b/lib/rules/no-invalid-position-at-import-rule/README.md
@@ -6,9 +6,8 @@ Disallow invalid position `@import` rules within a stylesheet.
 ```css
 a {}
 @import 'foo.css';
-/**   ↑
-  * Invalid position */
-```
+/** ↑
+  * This @import */
 
 ## Options
 
--- a/lib/rules/no-invalid-position-at-import-rule/__tests__/index.js
+++ b/lib/rules/no-invalid-position-at-import-rule/__tests__/index.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const stripIndent = require('common-tags').stripIndent;
+
 const { messages, ruleName } = require('..');
 
 testRule({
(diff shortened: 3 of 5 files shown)


lib/rules/no-invalid-position-at-import-rule/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Options
+
+### `true`
+
+The following patterns are considered violations:
+
+<!-- prettier-ignore -->
+```css
+a {}
```
voskresla, line 21: Let's add `;` (and in all rest) 😉

lib/rules/no-invalid-position-at-import-rule/__tests__/index.js
```diff
@@ -0,0 +1,80 @@
...
+			code: `
+			@CHARSET 'utf-8';
+			@imPORT 'foo.css';
+			@import 'bar.css';
+			`,
+			description: 'case insensitive',
+		},
+	],
+
+	reject: [
```
voskresla, line 48: Let's add test for double message?

```
code: 'a {}; @import "foo.css"; @import "bar.css"',
warnings: [
	{
		message: messages.rejected(`foo.css`),
		description: '@import after selector',
	},
	{
		message: messages.rejected(`bar.css`),
		description: '@import after selector',
	}
]
```

docs/user-guide/rules/list.md
```diff
@@ -79,6 +79,7 @@ Grouped first by the following categories and then by the [_thing_](http://apps.
 - [`no-empty-source`](../../../lib/rules/no-empty-source/README.md): Disallow empty sources.
 - [`no-extra-semicolons`](../../../lib/rules/no-extra-semicolons/README.md): Disallow extra semicolons (Autofixable).
 - [`no-invalid-double-slash-comments`](../../../lib/rules/no-invalid-double-slash-comments/README.md): Disallow double-slash comments (`//...`) which are not supported by CSS.
+- [`no-invalid-position-at-import-rule`](../../../lib/rules/no-invalid-position-at-import-rule/README.md): Disallow invalid position @import rules within a stylesheet.
```
jeddy3, line 82:
```suggestion
- [`no-invalid-position-at-import-rule`](../../../lib/rules/no-invalid-position-at-import-rule/README.md): Disallow invalid position `@import` rules within a stylesheet.
```

lib/rules/no-invalid-position-at-import-rule/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Disallow invalid position `@import` rules within a stylesheet.
+
+<!-- prettier-ignore -->
+```css
+a {}
+@import 'foo.css';
+/**   ↑
+  * Invalid position */
+```
```
jeddy3, line 11:
```suggestion
<!-- prettier-ignore -->
```css
a {}
@import 'foo.css';
/** ↑
  * This @import */
```
```

File: lib/rules/no-invalid-position-at-import-rule/index.js [hunk 5]
```diff
@@ -0,0 +1,67 @@
...
+
+const report = require('../../utils/report');
+const ruleMessages = require('../../utils/ruleMessages');
+const validateOptions = require('../../utils/validateOptions');
+const valueParser = require('postcss-value-parser');
+
+const ruleName = 'no-invalid-position-at-import-rule';
+
+const messages = ruleMessages(ruleName, {
+	rejected: (atImport) => `Unexpected invalid position @import rule ${atImport}`,
```
jeddy3, line 13:
```suggestion
	rejected: () => `Unexpected invalid position @import rule`,
```

In my [original blueprint](https://github.com/stylelint/stylelint/issues/5133#issuecomment-774642535), I made a mistake by including the value of the URL in the violation message. It is, unlike with the `no-duplicates` rule, not relevant here.

Let's remove:

- the value from the violation messages
- the associated code in rule
- the value from the tests

Sorry for causing extra effort in writing the rule.

lib/rules/no-invalid-position-at-import-rule/__tests__/index.js
```diff
@@ -0,0 +1,80 @@
...
+
+const { messages, ruleName } = require('..');
+
+testRule({
+	ruleName,
+	config: [true],
+
+	accept: [
+		{
+			code: `
```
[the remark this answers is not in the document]
jeddy3: We haven't settled on a convention yet. We use the following inconsistently:

- `stripIndent` and indent an extra tab in from the opening back tick
- template literals and flush to the left
- strings and explicit characters, e.g. `"/n/t@import"`

I suggest we adopt the first here, i.e. indent the code and use `stripIndent`, as it gives us both clean test reports and clean source files. (And we should use explicit characters only when a rule is testing whitespace, e.g. `indentation`.)

For example:

```js
'use strict';

const stripIndent = require('common-tags').stripIndent;

const { messages, ruleName } = require('..');

testRule({
	ruleName,
	config: [true],

	accept: [
		{
			code: stripIndent`
				@import 'foo.css';
				a {}
			`,
			description: '@import on first line',
		},
	]
});
``` 

Will give:

```
 PASS  lib/rules/no-invalid-position-at-import-rule/__tests__/index.js
  no-invalid-position-at-import-rule
    accept
      [ true ]
        "@import 'foo.css';\na {}"
```

lib/rules/no-invalid-position-at-import-rule/__tests__/index.js
```diff
@@ -0,0 +1,80 @@
...
+			@media print {}
+			@imPort URl('foo.css');
+			`,
+			message: messages.rejected('foo.css'),
+			description: 'case insensitive',
+			line: 3,
+			column: 4,
+		},
```
jeddy3, line 78: Lets test:

```
@import "foo.css"
a {}
@import "bar.css"
```

To make sure only the second `@import` is flagged.

jeddy3 (changes requested):
@chuuddo Thank you for your first contribution. The pull request is looking good! It's great to see [the naming conventions](https://stylelint.io/developer-guide/rules#write-the-readme) being followed too.

I've made some requests.

lib/rules/no-invalid-position-at-import-rule/index.js
```diff
@@ -0,0 +1,67 @@
...
+		root.walk((node) => {
+			if (node.type === 'comment' || (node.type === 'atrule' && /^charset$/i.test(node.name))) {
+				return;
+			}
+
+			if (node.type === 'atrule' && /^import$/i.test(node.name)) {
+				if (invalidPosition) {
+					const url = getAtImportUrl(node);
+
+					if (url) {
```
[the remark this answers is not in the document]
jeddy3: Let's remove the [conditional all together](https://github.com/stylelint/stylelint/pull/5202#discussion_r598248699).

lib/rules/no-invalid-position-at-import-rule/__tests__/index.js
jeddy3: Let's also include `line` and `column` for each warning.

lib/rules/no-invalid-position-at-import-rule/README.md
```diff
@@ -0,0 +1,49 @@
...
+Disallow invalid position `@import` rules within a stylesheet.
+
+<!-- prettier-ignore -->
+```css
+a {}
+@import 'foo.css';
+/** ↑
+  * This @import */
+```
+
```
jeddy3, line 12: Let's add the spec rule to the extended description here:

```suggestion

Any `@import` rules must precede all other valid at-rules and style rules in a style sheet (ignoring `@charset`), or else the `@import` rule is invalid.

```

jeddy3 (changes requested):
Thanks for making the changes!

One last request, then it looks good to me.

jeddy3 (approved):
Thanks! LGTM.

voskresla (approved):
thx!

lib/rules/no-invalid-position-at-import-rule/README.md
```diff
@@ -0,0 +1,51 @@
...
+Any `@import` rules must precede all other valid at-rules and style rules in a style sheet (ignoring `@charset`), or else the `@import` rule is invalid.
```
ybiquitous, line 13: `style sheet` seems a typo?

```suggestion
Any `@import` rules must precede all other valid at-rules and style rules in a stylesheet (ignoring `@charset`), or else the `@import` rule is invalid.
```

lib/rules/no-invalid-position-at-import-rule/index.js
```diff
@@ -0,0 +1,50 @@
...
+		}
+
+		let invalidPosition = false;
+
+		root.walk((node) => {
+			if (node.type === 'comment' || (node.type === 'atrule' && /^charset$/i.test(node.name))) {
+				return;
+			}
+
+			if (node.type === 'atrule' && /^import$/i.test(node.name)) {
```
ybiquitous, line 30: [IMO] It seems a bit readable to me if using `toLowerCase`. But this is just my opinion, so you can ignore it. 😃 

```suggestion
			const nodeName = node.name.toLowerCase();

			if (node.type === 'comment' || (node.type === 'atrule' && nodeName === 'charset')) {
				return;
			}

			if (node.type === 'atrule' && nodeName === 'import') {
```

ybiquitous (approved):
LGTM! 👍 
I've left some trivial comments, but this PR is sufficient. 😄

lib/rules/no-invalid-position-at-import-rule/README.md
chuuddo: Looks like a typo in specs https://drafts.csswg.org/css-cascade/#at-import.
But i think we should fix this in our docs.

lib/rules/no-invalid-position-at-import-rule/index.js
chuuddo: Suggestion looks suitable. Just made some research and `toLowerCase()` used much more often in rules.
@jeddy3 what do you think about it? Do we have any convention on this or should create one if not?

jeddy3: I had the same thought about `toLowerCase` as @ybiquitous when I was first reviewing the pull request as we've used that approach in loads of places. However, the regex approach sort of grew on me because we use it for filtering at-rules and declarations when walking (like we did in the [last rule we added](https://github.com/stylelint/stylelint/pull/5167/files#diff-280597139af609e4ef602c8171d06774830a2e1e98e2bc7536a1203c73abe2caR30)), so I didn't mention it.

> Do we have any convention on this or should create one if not?

I'm all for conventions, especially within rules, because they're such a time saver when reviewing code.

I'm not sure which of the approaches is more readable, though. So, let's go with @ybiquitous's `toLowerCase` suggestion for now to be consistent with how we're treating names and values in other rules. We can then, if we want, refactor across the board later down the line.

jeddy3: Following on, we'll probably stick with `toLowerCase` as the convention because (having looked myself) we often:
- do more string manipulation at the same time, e.g. trimming and slicing
- look for matches in arrays and sets

ybiquitous (approved):

jeddy3:
Changelog entry added:

> - Added: `no-invalid-position-at-import-rule` rule ([#5202](https://github.com/stylelint/stylelint/pull/5202)).

XhmikosR:
@jeddy3 just wondering, should this handle scss files too? Trying to figure out where I should report any issues https://github.com/twbs/stylelint-config-twbs-bootstrap/pull/110/checks?check_run_id=2447711414

XhmikosR:
@jeddy3:

* https://github.com/twbs/bootstrap/blob/main/scss/bootstrap-grid.scss
* https://github.com/twbs/bootstrap/blob/main/scss/bootstrap-reboot.scss

jeddy3:
I think we can rejig the rule so that it ignores non-standard syntax (like dollar variables) while being true to the spec that the rule enforces:

> Any `@import` rules must precede all other valid __at-rules and style rules__ in a style sheet (ignoring `@charset`), or else the `@import` rule is invalid. [Emphasis mine]

@XhmikosR Can you create a bug ticket, please?

XhmikosR:
Done, see #5263. Thanks :)

jeddy3:
@XhmikosR Fixed in `13.13.1`.

XhmikosR:
Thanks!

XhmikosR:
@jeddy3 should this also be fixed here? https://github.com/twbs/bootstrap/blob/v4-dev/scss/bootstrap-grid.scss

Sorry for not catching this earlier, I just updated to the latest versions.

jeddy3:
> should this also be fixed here?

Nope. That's the pattern the rule is warning against.

I recommend moving that code into another file and importing, e.g. `@import reset`, if you'd like to closer align with the CSS specs.

Alternatively, if you don't want to align, you can use `/* stylelint-disable no-invalid-position-at-import-rule */` to disable the rule for that section of code, or turn off the rule entirely.

ybiquitous:
I agree with @jeddy3.

Besides, it looks like out of this PR scope, but it seems good to me to migrate `@import` to `@use` because:

> The Sass team discourages the continued use of the `@import` rule.

-- https://sass-lang.com/documentation/at-rules/import

XhmikosR:
@jeddy3 Thanks, I'll just ignore the warnings in the v4-dev branch 👍 

@ybiquitous That's a different thing which we can't do for v4 because it's a breaking change.

PR #5203, opened 2021-03-19
Fix false positives for level 4 math functions in length-zero-no-unit

### LINKED ISSUE #5199: Fix false positives for level 4 math functions in length-zero-no-unit
I have the following css, and if I take out the unit from 0rem, as stylelint suggests, it breaks.

What I have:

```css
right: calc(-1 * max(100vw, 0rem));
```
What stylelint suggest me to do: (length-zero-no-unit error)

```css
right: calc(-1 * max(100vw, 0));
```

The later option without unit is not a valid css line in my browser.

What to expect:
length-zero-no-unit should be omitted inside max/min functions.

Reference: https://stackoverflow.com/questions/62520998/why-doesnt-min-or-max-work-with-unitless-0

tomyo:
Closes #5199

First commit is simply small styling alignment, not part of the issue, but small enough to simply include here I guess.

### DIFF
--- a/lib/rules/length-zero-no-unit/__tests__/index.js
+++ b/lib/rules/length-zero-no-unit/__tests__/index.js
@@ -32,6 +32,30 @@ testRule({
 			description: 'ignore calc, but not inner functions',
 		},
 		{
+			code: 'a { right: max(100vw, 0rem)); }',
+			description: 'ignore max',
+		},
+		{
+			code: 'a { right: calc(max(100vw, 0rem))); }',
+			description: 'ignore max inside calc',
+		},
+		{
+			code: 'a { right: min(100vw, 0rem)); }',
+			description: 'ignore min',
+		},
+		{
+			code: 'a { right: calc(min(100vw, 0rem))); }',
+			description: 'ignore min inside calc',
+		},
+		{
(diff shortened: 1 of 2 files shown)


lib/rules/length-zero-no-unit/__tests__/index.js
```diff
@@ -12,24 +12,48 @@ testRule({
...
+		{
+			code: 'a { right: calc(min(100vw, 0rem))); }',
+			description: 'ignore min inside calc',
+		},
+		{
+			code: 'a { right: clamp(50rem, 50vw, 0rem))); }',
+			description: 'ignore clamp',
+		},
+		{
+			code: 'a { right: clac(clamp(50rem, 50vw, 0rem)))); }',
```
ybiquitous, line 55: It seems a typo:

```suggestion
			code: 'a { right: calc(clamp(50rem, 50vw, 0rem)))); }',
```

ybiquitous (approved):
LGTM 👍

ybiquitous (approved):

jeddy3 (approved):
Thanks!

jeddy3:
Changelog entry added:

> - Fixed: `length-zero-no-unit` false positives for level 4 math functions ([#5203](https://github.com/stylelint/stylelint/pull/5203)).

PR #5209, opened 2021-03-24
new rule: no-irregular-whitespace

### LINKED ISSUE #5201: Add no-irregular-whitespace
I've noticed that I sometimes end up typing a bit too fast when adding white spaces in my code when I've recently used alt or option key + space in Mac in VS Code (for example when adding a $ for a variable), this results in non-breaking spaces (&nbsp) being added to the code which look identical to regular white spaces, but when parsed through sass this causes issues since the output CSS still ends up getting a non-breaking whitespace and also an additional white space added from the parsing. 

I've also added a feature request for this in https://github.com/sass/dart-sass/issues/1265 but I think it makes sense to empower developers to be able to avoid writing irregular whitespace characters in the first place through a stylelint rule.

A new rule for checking the code for irregular whitespace characters, similar to https://eslint.org/docs/rules/no-irregular-whitespace, this rule should probably not be applicable for `content: " ";` (where the space in content in this example is a non-breaking space.)

### Example:
https://codepen.io/jchiem/pen/dyOBpKW

itutto:
## Implementation proposal for issue: [5201](https://github.com/stylelint/stylelint/issues/5201)

This PR proposes an implementation for `no-irregular-whitespace` rule, which has been previously approved for implementation in the referenced issue.

Closes #5201

### DIFF
--- a/lib/rules/no-irregular-whitespace/__tests__/index.js
+++ b/lib/rules/no-irregular-whitespace/__tests__/index.js
@@ -2,6 +2,36 @@
 
 const { messages, ruleName } = require('..');
 
+const IRREGULAR_WHITESPACES = [
+	'\u000B', // Line Tabulation (\v) - <VT>
+	'\u000C', // Form Feed (\f) - <FF>
+	'\u00A0', // No-Break Space - <NBSP>
+	'\u0085', // Next Line
+	'\u1680', // Ogham Space Mark
+	'\u180E', // Mongolian Vowel Separator - <MVS>
+	'\ufeff', // Zero Width No-Break Space - <BOM>
+	'\u2000', // En Quad
+	'\u2001', // Em Quad
+	'\u2002', // En Space - <ENSP>
+	'\u2003', // Em Space - <EMSP>
+	'\u2004', // Tree-Per-Em
+	'\u2005', // Four-Per-Em
+	'\u2006', // Six-Per-Em
+	'\u2007', // Figure Space
+	'\u2008', // Punctuation Space - <PUNCSP>
+	'\u2009', // Thin Space
+	'\u200A', // Hair Space
+	'\u200B', // Zero Width Space - <ZWSP>
+	'\u2028', // Line Separator
+	'\u2029', // Paragraph Separator
+	'\u202F', // Narrow No-Break Space
+	'\u205F', // Medium Mathematical Space
+	'\u3000', // Ideographic Space
+];
+
+const characterToUnicodeString = (str) =>
+	`\\u${str.charCodeAt(0).toString(16).toUpperCase().padStart(4, '0')}`;
+
 testRule({
 	ruleName,
 	config: true,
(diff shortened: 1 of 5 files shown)


ybiquitous:
[question] In the blueprint <https://github.com/stylelint/stylelint/issues/5201#issuecomment-802144994>, this rule is proposed without secondary options, and I agree with the blueprint.
What reason do you have to provide the secondary options in this PR?

itutto:
> [question] In the blueprint [#5201 (comment)](https://github.com/stylelint/stylelint/issues/5201#issuecomment-802144994), this rule is proposed without secondary options, and I agree with the blueprint.
> What reason do you have to provide the secondary options in this PR?

I thought it would be a nice addition if we gave the users some control over which types of irregular whitespaces should be disallowed. I don't have have a practical example where it definitely could come handy unless if we consider it may be purposefully used in an attribute selector. 
```css
a[title="some content with irregular whitespace"]
```

ybiquitous:
@itutto Thank you for the detailed description. I'm a bit worried if there is more complexity than needed... 🤔 

For example, a case when both `allow` and `only` options are specified at the same time:

```json
{
  "rules": {
    "no-irregular-whitespace": [
      true, { "allow": ["\u0085"], "only": ["\u00A0"] }
    ]
  }
}
```

But I would like to know @jeddy3's opinion about this.

lib/rules/no-irregular-whitespace/index.js
```diff
@@ -0,0 +1,159 @@
...
+	'\u2028', // Line Separator
+	'\u2029', // Paragraph Separator
+	'\u202F', // Narrow No-Break Space
+	'\u205f', // Medium Mathematical Space
+	'\u3000', // Ideographic Space
+];
+
+const generateInvalidWhitespaceValidator = (ignoreArray = []) => {
+	const pattern = new RegExp(
+		`(${irregularWhitespaces.filter((ws) => ignoreArray.indexOf(ws) === -1).join('|')})`,
```
ybiquitous, line 43: [suggestion] How about using [`Array.prototype.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) for readability?

```suggestion
		`(${irregularWhitespaces.filter((ws) => !ignoreArray.includes(ws)).join('|')})`,
```

lib/rules/no-irregular-whitespace/index.js
```diff
@@ -0,0 +1,159 @@
...
+			return;
+		}
+
+		let initializeParam = [];
+
+		if (options) {
+			if (options.allow) initializeParam = options.allow;
+
+			if (options.only)
+				initializeParam = irregularWhitespaces.filter((ws) => options.only.indexOf(ws) > -1);
```
ybiquitous, line 128: Same here for `Array.prototype.includes()`:

```suggestion
				initializeParam = irregularWhitespaces.filter((ws) => options.only.includes(ws));
```

lib/rules/no-irregular-whitespace/index.js
```diff
@@ -0,0 +1,159 @@
...
+	const validatorForKey = {};
+
+	allKeys.forEach((key) => {
+		if (typeof nodeSchema[key] === 'string') validatorForKey[key] = validator;
+
+		if (typeof nodeSchema[key] === 'object')
+			validatorForKey[key] = generateNodeValidator(nodeSchema[key], validator);
+	});
+
+	// this will be called many times, so it's optimized for performance and not readibility. Surprisingly, this seem to be slightly faster then concatenating the params and running the validator once
```
ybiquitous, line 90: [suggestion] How about fixing typos and avoiding the long line?

```suggestion
	// This will be called many times, so it's optimized for performance and not readability. 
	// Surprisingly, this seems to be slightly faster than concatenating the params and running the validator once.
```

lib/rules/no-irregular-whitespace/index.js
```diff
@@ -0,0 +1,159 @@
...
+	return (node) => {
+		for (let i = 0, currentKey; i < allKeysLength; i++) {
+			currentKey = allKeys[i];
+
```
ybiquitous, line 94: [suggestion] [`for...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) seems more readable here:

```suggestion
		for (const currentKey of allKeys) {
```

jeddy3:
@itutto Thank you for the pull request. It's looking good so far.

> Please let me know if I should remove it anyway.

Yes, please remove the secondary options. We only them when there has been a specific user request for them. There are lots of rules in stylelint and so we try to minimise complexity where we can.

itutto:
> @itutto Thank you for the pull request. It's looking good so far.
> 
> > Please let me know if I should remove it anyway.
> 
> Yes, please remove the secondary options. We only them when there has been a specific user request for them. There are lots of rules in stylelint and so we try to minimise complexity where we can.

Thank you both for the reviews. I've removed the secondary options and applied the other suggested changes as well.

lib/rules/no-irregular-whitespace/index.js
```diff
@@ -0,0 +1,136 @@
...
+	'\u3000', // Ideographic Space
+];
+
+const generateInvalidWhitespaceValidator = (ignoreArray = []) => {
+	const pattern = new RegExp(
+		`(${IRREGULAR_WHITESPACES.filter((ws) => !ignoreArray.includes(ws)).join('|')})`,
+	);
+
+	return (str) => typeof str === 'string' && pattern.exec(str);
+};
```
ybiquitous, line 47: [suggestion] Perhaps, it seems we can simplify `generateInvalidWhitespaceValidator()` as below:

```suggestion
const IRREGULAR_WHITESPACES_PATTERN = new RegExp(`([${IRREGULAR_WHITESPACES.join('')}])`);

const generateInvalidWhitespaceValidator = () => {
	return (str) => typeof str === 'string' && IRREGULAR_WHITESPACES_PATTERN.exec(str);
};
```

lib/rules/no-irregular-whitespace/__tests__/index.js
```diff
@@ -0,0 +1,98 @@
+'use strict';
+
+const { messages, ruleName } = require('..');
```
ybiquitous, line 3: [suggestion] Perhaps, it seems we can reduce the duplication by exporting `IRREGULAR_WHITESPACES` as below:

```suggestion
const { messages, ruleName, IRREGULAR_WHITESPACES } = require('..');
```

Edit `lib/rules/no-irregular-whitespace/index.js`:

```js
// For test
module.exports.IRREGULAR_WHITESPACES = IRREGULAR_WHITESPACES;
```

But this change exports an essentially needless object. What do you think?

ybiquitous (commented):
@itutto Thanks for addressing reviews. It seems this PR is almost enough already, but I left some refactoring ideas.

It would be nice if you can check it. 😄

lib/rules/no-irregular-whitespace/__tests__/index.js
itutto: I prefer deduplicating as well, but I haven't seen extra exports used in any other rule, so I wasn't sure if it is allowed to do in this environment.

ybiquitous: Make sense. It seems better to keep it as-is.

lib/rules/no-irregular-whitespace/index.js
itutto: I agree, thanks! 👍

ybiquitous (approved):
LGTM 👍

itutto:
Would you like to see any further changes, @jeddy3 ?

jeddy3 (approved):
Thanks! LGTM.

jeddy3:
Changelog entry added:

- Added: `no-irregular-whitespace` rule ([#5209](https://github.com/stylelint/stylelint/pull/5209)).

PR #5215, opened 2021-03-28
Refactor via new utility `isNonNegativeInteger()`

ybiquitous:
This refactoring aims to reduce duplications and resolve some `no-shadow` violations of ESLint (#4985 and #5013) by a new utility function ~~`isNonNegative()`~~ `isNonNegativeInteger()`.

Note: I considered also `isNaturalNumber()`, but I rejected it because “natural number” is ambiguous if including zero or not.

Related to #4985 and #5013

No, it's self-explanatory.


lib/rules/selector-max-compound-selectors/index.js
```diff
@@ -23,29 +24,21 @@ function rule(max) {
 	return (root, result) => {
 		const validOptions = validateOptions(result, ruleName, {
 			actual: max,
-			possible: [
-				// TODO: Issue #4985
-				// eslint-disable-next-line no-shadow
-				function (max) {
-					return typeof max === 'number' && max > 0;
```
ybiquitous, line 30: [note] This does not allow zero, but I think no problems if using `isNonNegative()` as with the other rules.

jeddy3:
A utility function for the options validator is a great idea.

How about `isPositiveInteger`, and we add [`isInteger()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) to the conditional? For these rules, a maximum of a float, e.g. `1.5`, doesn't make sense.

lib/rules/selector-max-compound-selectors/index.js
jeddy3: A util for this is a great idea.

In the future (bug fix pull request), we could rename to `isNonNegativeInteger`, and we add [`isInteger()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) to the conditional? For these rules, a maximum of a float, e.g. `1.5`, doesn't make sense.

jeddy3 (approved):
LGTM, thanks!

lib/rules/selector-max-compound-selectors/index.js
ybiquitous: Renaming to `isNonNegativeInteger()` is a good idea! 👍 

Since we support Node 10, we need a polyfill of `isInteger()` still. But the polyfill is simple, so I'll do it with this PR. 💪 

https://github.com/stylelint/stylelint/blob/1e81784e3abcd6151e98c5cffc94703aec499d87/package.json#L196-L198

ybiquitous: Renamed. 944fab6a

lib/utils/isNonNegativeInteger.js
```diff
@@ -0,0 +1,17 @@
...
+	Number.isInteger ||
+	function (value) {
+		return typeof value === 'number' && isFinite(value) && Math.floor(value) === value;
+	};
+
+/**
+ * @param {unknown} value
+ */
+module.exports = function (value) {
+	return isInteger(value) && typeof value === 'number' && value >= 0;
```
ybiquitous, line 16: [note] `typeof value === 'number'` may be needless, but this is actually needed. Unless this `typeof`, a TypeScript error occurs:

```diff
-	return isInteger(value) && typeof value === 'number' && value >= 0;
+	return isInteger(value) && value >= 0;
```

```
lib/utils/isNonNegativeInteger.js:16:29 - error TS2571: Object is of type 'unknown'.

16  return isInteger(value) && value >= 0;
                               ~~~~~
```

mattxwang (approved):
Also LGTM on my end, great to knock out more `no-shadow` violations!

ybiquitous:
@jeddy3 Thanks for creating the issue! 👏

jeddy3:
@ybiquitous Feel free to merge your own pull requests when they have two reviews.

PR #5220, opened 2021-03-31
Improve examples for keyframe-declaration-no-important

jeddy3:

- `foo` matches the conventions
- `opacity` is a better real-world example as animating `margin` isn't performant


ybiquitous (approved):
Nice! LGTM 👍

PR #5231, opened 2021-04-06
Refactor `lib/__tests__/postcssPlugin.test.js`

ybiquitous:
This refactoring aims to make the usage of `Promise` in the test safer.
(via `.resolves`, `.rejects`, and other Jest matchers)

See also <https://jestjs.io/docs/expect>.

As a result, we can remove callbacks (which may not be performed) from the test.

Part of #4881

No, it's self-explanatory.


mattxwang (approved):
Similarly LGTM - appreciate that each of these PRs are broken up into byte-sized chunks, it makes it easier to review!

jeddy3 (approved):
Thanks!

PR #5240, opened 2021-04-12
Fix false positives for variables in font-family-no-missing-generic-family-keyword

### LINKED ISSUE #4765: Fix false positives for variables in font-family-no-missing-generic-family-keyword

ybiquitous:
This change aims to ignore any variables (e.g. `var()`, `$var`, `@var` etc.) for the `font-family-no-missing-generic-family-keyword` rule.

Fix #4765 (retry of #4806)

This PR ignores **always** any variables for the `font-family` property, but perhaps some people might want to ignore **explicitly** via the `ignoreFontFamilies` secondary option as below:

```js
{
  "ignoreFontFamilies": [/\$fontFamily/]
}
```

Could you please give me feedback about this point?

### DIFF
--- a/lib/rules/font-family-no-missing-generic-family-keyword/README.md
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/README.md
@@ -54,6 +54,16 @@ a { font: inherit; }
 a { font: caption; }
 ```
 
+<!-- prettier-ignore -->
+```css
+a { font-family: var(--font-family-common); }
+```
+
+<!-- prettier-ignore -->
+```css
+a { font-family: Helvetica, var(--font-family-common); }
+```
+
 ## Optional secondary options
 
 ### `ignoreFontFamilies: ["/regex/", /regex/, "string"]`
--- a/lib/rules/font-family-no-missing-generic-family-keyword/index.js
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/index.js
@@ -4,6 +4,7 @@
 
 const declarationValueIndex = require('../../utils/declarationValueIndex');
 const findFontFamily = require('../../utils/findFontFamily');
+const isStandardSyntaxValue = require('../../utils/isStandardSyntaxValue');
 const isVariable = require('../../utils/isVariable');
 const keywordSets = require('../../reference/keywordSets');
 const optionsMatches = require('../../utils/optionsMatches');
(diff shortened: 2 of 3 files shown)


lib/rules/font-family-no-missing-generic-family-keyword/README.md
```diff
@@ -54,6 +54,16 @@ a { font: inherit; }
...
+<!-- prettier-ignore -->
+```css
+a { font-family: var(--font-family-common); }
+```
+
+<!-- prettier-ignore -->
+```css
+a { font-family: Helvetica, var(--font-family-common); }
+```
+
```
ybiquitous, line 66: [note] I think it valuable to add examples which are using variables.

lib/rules/font-family-no-missing-generic-family-keyword/index.js
```diff
@@ -23,6 +24,12 @@ const messages = ruleMessages(ruleName, {
 const isFamilyNameKeyword = (node) =>
 	!node.quote && keywordSets.fontFamilyKeywords.has(node.value.toLowerCase());
 
+const isLastFontFamilyVariable = (value) => {
+	const lastValue = postcss.list.comma(value).pop();
```
ybiquitous, line 28: [note] `postcss.list.comma()` seems more suitable than `postcss.list.space()` in this case.

See also <https://postcss.org/api/#list>

jeddy3 (approved):
> This PR ignores always any variables for the font-family property, but perhaps some people might want to ignore explicitly via the ignoreFontFamilies secondary option

I think your PR is the right call. Overloading `ignoreFontFamily` for variables is confusing users, and we should reserve it for its original intention of icon fonts, e.g. `font-family: awesomefont;` and `ignoreFontFamily: ["awesomefont"]`

LGTM, thanks.

jeddy3:
> - Fixed: `font-family-no-missing-generic-family-keyword` false positives for variables ([#5240](https://github.com/stylelint/stylelint/pull/5240)).

ybiquitous:
@jeddy3 Thank you for the review. 😊

Issue #5211, opened 2021-03-25
Add autofix to selector-attribute-quotes


ProtonGravity:
I have tried to fix some issues in CSS file together with option --fix, but it don't fix problems only reports about errors.

> Clearly describe the bug

For Example: 

I have the simplest rule file:

{
  "rules": {
"selector-attribute-quotes": "always",
"string-quotes": "double"
  }
}

and run this in the terminal:

stylelint "**/*.css" --fix

and receive this:

 72:12  ×  Expected quotes around "text"   selector-attribute-quotes

but this not fix anything

> Which rule, if any, is the bug related to?

e.g. selector-attribute-quotes

> Which version of stylelint are you using?

13.12.0

OS : WINDOWS 10

> What did you expect to happen?

I expected to fix this and of course inform about a problem in the terminal (what it does).

Please inform me if I am doing something wrongly ?

I also must say that sometimes stylelint fix problems, for example with 
plugin stylelint-order it is working.

jeddy3:
@ProtonGravity Thanks for the report and for using the template.

The `selector-attribute-quotes` isn't autofixable yet. Autofixable rules are labelled with "(Autofixable)" in [the list of rules](https://stylelint.io/user-guide/rules/list).

I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

There are [steps on how to add autofix to a rule](https://stylelint.io/developer-guide/rules#add-autofix) in the Developer guide.

Let's autofix with double quotes. The `string-quotes` rule can then switch them to single if neccessary.

PR #5248, opened 2021-04-15
Add autofix to selector-attribute-quotes

doing-art:

Have added autofix to `selector-attribute-quotes` rule as dicussed in issue [#5211](https://github.com/stylelint/stylelint/issues/5211) 

Some of the unit tests can be a bit confusing. They cover pretty unusual but valid selectors:
```css
[href=te\'s\"t] { }
[href=\"test\"] { }
[href=\'test\'] { }
[href="te'st"] { }
[href='te"st'] { }
[href='te\'st'] { }
[href="te\"st"] { }
```

### DIFF
--- a/lib/rules/selector-attribute-quotes/index.js
+++ b/lib/rules/selector-attribute-quotes/index.js
@@ -15,7 +15,9 @@ const messages = ruleMessages(ruleName, {
 	rejected: (value) => `Unexpected quotes around "${value}"`,
 });
 
-function rule(expectation) {
+const acceptedQuoteMark = '"';
+
+function rule(expectation, secondary, context) {
 	return (root, result) => {
 		const validOptions = validateOptions(result, ruleName, {
 			actual: expectation,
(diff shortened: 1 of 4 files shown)


lib/rules/selector-attribute-quotes/index.js
```diff
@@ -36,25 +38,41 @@ function rule(expectation) {
 			}
 
 			parseSelector(ruleNode.selector, result, ruleNode, (selectorTree) => {
```
jeddy3, line 40:
```suggestion
			parseSelector(getRuleSelector(ruleNode), result, ruleNode, (selectorTree) => {
```

We'll need a new util here.

jeddy3 (changes requested):
@doing-art Thank you for the pull request. It's almost there.

Can we add a test with a comment within a selector list please, e.g.:

```
{
  code: 'a[target="_blank"], /* comment */ a { }',
  fixed: 'a[target=_blank], /* comment */ a { }',
  message: messages.rejected('_blank'),
  line: 1,
  column: 10,
},
```

It'll fail because comments are stripped for the `selector` property of the AST. The full string is available in the `raws` property, and we'll want to pass this full string to the selector parser. ([Ref](https://astexplorer.net/#/gist/77cca53302eb6e016c2ef2516537594a/71abb107772e9c19798c5d5daf543bb2816ebf73)).

Let's create a `getRuleSelector` util based off [`getDeclarationValue`](https://github.com/stylelint/stylelint/blob/master/lib/utils/getDeclarationValue.js) and use it in the rule.

lib/rules/selector-attribute-quotes/index.js
doing-art: > @doing-art Thank you for the pull request. It's almost there.
> 
> Can we add a test with a comment within a selector list please, e.g.:
> 
> ```
> {
>   code: 'a[target="_blank"], /* comment */ a { }',
>   fixed: 'a[target=_blank], /* comment */ a { }',
>   message: messages.rejected('_blank'),
>   line: 1,
>   column: 10,
> },
> ```
> 
> It'll fail because comments are stripped for the `selector` property of the AST. The full string is available in the `raws` property, and we'll want to pass this full string to the selector parser. ([Ref](https://astexplorer.net/#/gist/77cca53302eb6e016c2ef2516537594a/71abb107772e9c19798c5d5daf543bb2816ebf73)).
> 
> Let's create a `getRuleSelector` util based off [`getDeclarationValue`](https://github.com/stylelint/stylelint/blob/master/lib/utils/getDeclarationValue.js) and use it in the rule.

@jeddy3 Thanks for reviewing the PR. I made the requested change.

jeddy3 (approved):

LGTM.

mattxwang (approved):
LGTM, seems like a comprehensive set of test cases to me!

jeddy3:
Changelog entry added:

> - Added: `selector-attribute-quotes` autofix ([#5248](https://github.com/stylelint/stylelint/pull/5248)).

PR #5245, opened 2021-04-12
Fix false positives for functions inside of math functions in length-…

doing-art:
…zero-no-unit rule

> Which issue, if any, is this issue related to?

Fix a small diffect explained in the issue [#5204 ](https://github.com/stylelint/stylelint/issues/5204)

> Is there anything in the PR that needs further explanation?

No, it's self-explanatory.

PR #5250, opened 2021-04-16
Ignore length-zero-no-unit violations found within var() functions if custom properties are ignored

lanthaler:

Related to https://github.com/stylelint/stylelint/issues/4945, but limited to the var() function for now

Should be self-explanatory. More than happy to answer questions if there are any

### DIFF
--- a/lib/rules/length-zero-no-unit/README.md
+++ b/lib/rules/length-zero-no-unit/README.md
@@ -61,9 +61,14 @@ a { top: 1.001vh }
 
 Ignore units for zero length in custom properties.
 
-The following pattern is _not_ considered a violation:
+The following patterns are _not_ considered violations:
 
 <!-- prettier-ignore -->
 ```css
 a { --x: 0px; }
+```
+
+<!-- prettier-ignore -->
+```css
+div { width: calc(var(--x, 0px) + 10px); }
 ```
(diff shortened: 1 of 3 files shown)


jeddy3 (changes requested):
@lanthaler Thanks for the pull request.

Two related pull requests were recently merged (https://github.com/stylelint/stylelint/pull/5203 & https://github.com/stylelint/stylelint/pull/5245) which I think will address the problem most users are facing with this rule, i.e. using `var` within `calc` or some other math function. 

However, users may wish to ignore the contents of functions outright, whether that's `var` or something else. The [convention we use](https://stylelint.io/user-guide/rules/about#user-defined-ignore) is an `ignore*` secondary function, e.g. `ignoreFunctions`.

Rather than overloading the `custom-properties` keyword option, this option will allow users to make the rule more permissive on a granular scale.

Is it possible to rework this pull request to https://github.com/stylelint/stylelint/issues/4945#issuecomment-729069609?

jeddy3:
I hadn't realised the rule was a mesh of style-search and the value parser. I've created an issue to refactor the rule to only use the value parser, which will make adding an `ignoreFunction` option easier.

lanthaler:
I just discovered that there's already a lingering PR to add ignoreFunction #5082 and that you started refactoring this code altogether #5256. If this change isn't of interest as is, let's just close this PR. The reason I "overloaded" the custom-properties option is because the `var` function is exclusively about custom properties.

jeddy3:
> If this change isn't of interest, let's just close this PR.

Yes, let's do that. With the recently merged #5203 & #5245 (which are in the refactoring of #5256), I think the most common problem of using `var` within `calc` is solved.

Thanks again for the pull request.

Issue #5274, opened 2021-05-06
Format lint messages like a C compiler


rdiez:
I am a newcomer and I was surprised at the lint message formatting:

SomeFile1.html
   1:23   ✖  Some warning
   2:34   ✖  Something else

SomeFile2.html
   4:56   ✖  Some other warning

That is hard to parse (for a machine). I guess that most IDEs, like Atom, will have their own plug-in that understands that format. Or maybe there is way to generate that information as a JSON file or whatever.

But wouldn't it be easier to (optionally) generate such messages like a C compiler does?

/home/user/some/file.cpp:123:45: error: some error here

This way, my Emacs would recognise those warnings and automatically hyperlink them.

Emacs is rather flexible about the message formatting, see compilation-error-regexp-alist . But I do not think that this option can adapt to stylelint, because the filename is separate from the warning messages.

ybiquitous:
@rdiez Could you try `-f unix` flag, please?

https://stylelint.io/user-guide/usage/options#formatter

rdiez:
Thanks, I'm sorry I missed that option.

PR #5278, opened 2021-05-09
Explain what --formatter=unix generates

rdiez:
Without the extra explanation, people like myself may easily miss this option and waste time creating spurious issues like this one: #5274

None, as it's a documentation improvement.

No, it's self-explanatory.


rdiez:
Your alternative is fine by me. I suggest that you drop my pull request and commit your version instead.

PR #5286, opened 2021-05-12
Refactor `lib/__tests__/message.test.js`

ybiquitous:
Reduce callbacks as much as possible via `async/await` syntax.
(see also <https://jestjs.io/docs/asynchronous>)

A part of #4881

No, it's self-explanatory.


jeddy3 (approved):
LGTM, thanks!

PR #5296, opened 2021-05-12
Remove function-calc-no-invalid

jeddy3:
> Which issue, if any, is this issue related to?

Closes https://github.com/stylelint/stylelint/issues/4731

> Is there anything in the PR that needs further explanation?

Removes one of the two [code scanning](https://github.com/stylelint/stylelint/security/code-scanning) vulnerabilities. This one is found in the generated parser code used by the rule.

This is the 2nd pull request for the `v14` branch.

Once we merged this pull request and https://github.com/stylelint/stylelint/pull/5295, we'll be able to tackle https://github.com/stylelint/stylelint/issues/5289 (removing the `syntax` option). Once that is done, I believe all the other [14.0.0 release issues](https://github.com/stylelint/stylelint/issues/5205), e.g. moving to ESM and migrating to PostCSS 8, will be unblocked and can be worked on in parallel because they shouldn't conflict with each other.

PR #5295, opened 2021-05-12
Remove deprecated rules

jeddy3:
> Which issue, if any, is this issue related to?

Closes https://github.com/stylelint/stylelint/issues/5290

> Is there anything in the PR that needs further explanation?

The first of the [14.0.0 release steps](https://github.com/stylelint/stylelint/issues/5205#issue-837034702), and the first pull request to be merged into the `v14` branch.

PR #5297, opened 2021-05-12
Remove syntax option

### LINKED ISSUE #5289: Remove syntax option
Ref: https://github.com/stylelint/stylelint/issues/4942#issuecomment-839514125

We'll need to:

1. Replace `syntax` with `customSyntax` in the Jest preset (and release a new version)
2. Move PostCSS 8 compatible syntaxes to dev dependencies and remove the non-compatible ones
3. Update rule tests to require the needed syntaxes, and update `syntax` to `customSyntax`
4. Remove the `syntax` and postcss-syntax logic from the engine

jeddy3:

Closes #5289

A work-in-progress pull request showing the removal of the:
- `syntax` option in favour of `customSyntax`
- syntaxes that aren't compatible with PostCSS 8

The rule tests reliant on the `syntax` option will fail until the Jest preset is updated and the tests themselves moved to `customSyntax`. The `color-function-notation` rule is done, and we can do the rest once https://github.com/stylelint/stylelint/pull/5295 and https://github.com/stylelint/stylelint/pull/5296 are merged.

This would be the third pull request for the `v14` branch.


jeddy3:
I've:

- added a message when the `syntax` option is used
- moved all the rules over to `customSyntax`

I've commented out the PostCSS@8-incompatible syntax tests for each rule. We can reinstate them if the existing syntaxes are made compatible or move them to leaner and more specific syntaxes if they are written.

The tests will past once https://github.com/stylelint/jest-preset-stylelint/pull/32 is merged and released.

jeddy3:
This pull request is now ready for review.

lib/getPostcssResult.js
```diff
@@ -39,40 +39,19 @@ module.exports = function (stylelint, options = {}) {
 
 	return getCode
 		.then((code) => {
+			if (stylelint._options.syntax) {
+				throw new Error(
+					'The syntax option has been removed. Install the appropriate syntax and use the customSyntax option instead',
```
ybiquitous, line 44: [suggestion] How about improving the message as below?

- Emphasize the option names with quotes.
- Add the CLI flag names.

```suggestion
					'The "syntax" (--syntax) option has been removed. Install the appropriate syntax and use the "customSyntax" (--custom-syntax) option instead',
```

package.json
```diff
@@ -180,14 +172,18 @@
...
     "jest-watch-typeahead": "^0.6.3",
     "lint-staged": "^10.5.4",
     "np": "^7.5.0",
     "npm-run-all": "^4.1.5",
     "postcss-import": "^12.0.1",
-    "prettier": "^2.2.1",
+    "postcss-less": "^3.1.4",
+    "postcss-sass": "^0.4.4",
+    "postcss-scss": "^2.1.1",
+    "prettier": "2.2.1",
```
ybiquitous, line 184: [question] Is this `prettier` version change intentional?

```diff
-   "prettier": "^2.2.1",
+   "prettier": "2.2.1",
```

lib/getPostcssResult.js
jeddy3: SGTM. We can revisit the message in https://github.com/stylelint/stylelint/issues/5301 to make it consistent with the other messaging.

package.json
jeddy3: Prettier was updating when I was updating the other dependencies, causing formatting changes in files unrelated to this pull request.

We [should be used using an exact version](https://prettier.io/docs/en/install.html).

ybiquitous: Thanks. I understand. 👍🏼

ybiquitous (commented):
[suggestion] It seems that we can remove this file: ⬇️ 

https://github.com/stylelint/stylelint/blob/b01ed25dfa3e8231a976eef76bd5e81fb535b1e9/types/global.d.ts#L1-L5

ybiquitous:
Also for `postcss-syntax`:

https://github.com/stylelint/stylelint/blob/b01ed25dfa3e8231a976eef76bd5e81fb535b1e9/types/postcss-syntax/index.d.ts#L1

jeddy3:
I've removed the redundant types.

lib/rules/color-function-notation/__tests__/index.js
```diff
@@ -363,7 +364,7 @@ testRule({
 
 testRule({
 	ruleName,
-	syntax: 'scss',
+	syntax: postcssScss,
```
hudochenkov, line 367:
```suggestion
	customSyntax: postcssScss,
```

package-lock.json
```diff
@@ -21531,10 +21797,68 @@
       "resolved": "https://registry.npmjs.org/style-search/-/style-search-0.1.0.tgz",
       "integrity": "sha1-eVjHk+R+MuB9K1yv5cC/jhLneQI="
     },
+    "stylelint": {
```
hudochenkov, line 21800: Hope we won't have problems, that this `stylelint` will be always behind. “Thanks”, npm 7.

hudochenkov (changes requested):

package-lock.json
jeddy3: Where's this entry coming from, is it the jest-preset? We can set the peer to 14 when it's out.

hudochenkov: Yes, peerDependency of jest-preset.

hudochenkov (approved):

package-lock.json
jeddy3: Shall we remove the peer dependency? It's odd having stylelint installed in stylelint automatically.

jeddy3: I can do it as a patch release of the preset. We probably don't want an old version of stylelint (and postcss) hanging around as we migrate to PostCSS 8 it adds a bit of an unknown variable into the mix.

hudochenkov: I don't think anyone uses jest-preset to test stylelint plugin without stylelint installed. So it's probably safe to remove peer dependency.

lib/cli.js
```diff
@@ -130,13 +130,6 @@ const meowOptions = {
...
-      --syntax, -s
-
-        Specify a syntax. Options: "css", "css-in-js", "html", "less",
-        "markdown", "sass", "scss", "sugarss". If you do not specify a syntax,
-        syntaxes will be automatically inferred by the file extensions
-        and file content.
-
       --fix
 
         Automatically fix violations of certain rules.
```
ybiquitous, line 135: [suggestion] We can remove the followings:

https://github.com/stylelint/stylelint/blob/5dfe07c1ba44425227b2c25992663916ff8566f1/lib/cli.js#L44

https://github.com/stylelint/stylelint/blob/5dfe07c1ba44425227b2c25992663916ff8566f1/lib/cli.js#L79

https://github.com/stylelint/stylelint/blob/5dfe07c1ba44425227b2c25992663916ff8566f1/lib/cli.js#L312-L315

lib/rules/color-hex-case/__tests__/index.js
```diff
@@ -165,180 +165,180 @@ testRule(
...
-		},
-		{
-			code: `
-				export const s = styled.a({
-					color: "white /* #FFF */",
-				});
-			`,
-		},
-	],
+// testRule({
```
ybiquitous, line 168: [question] What is the reason for this comment-out? It might be good to leave the reason as a code comment.

E.g.

```suggestion
// NOTE: (describing comment-out reason...)
// testRule({
```

ybiquitous: Also, I prefer `/* ... */` over `// ...` because of fewer diffs, like other files.

ybiquitous: > Also, I prefer `/* ... */` over `// ...` because of fewer diffs, like other files.

Sorry, I understand using `// ...`. 🙇🏼  
Because `/* ... */` is unavailable for tests including `/* ... */` content.

ybiquitous: I've added a feature to skip specific `testRule` via stylelint/jest-preset-stylelint#33.

I can confirm it works on my localhost as below:

```console
npm i -D git+https://github.com/stylelint/jest-preset-stylelint.git#add-test-group-only-and-skip
```

```js
testRule({
	skip: true, //<-- add
	ruleName,
    // ...
});
```

lib/cli.js
jeddy3: Aren't these still needed for showing a warning message when `--syntax` is used?

ybiquitous: Ah, surely 👍🏼

lib/rules/color-hex-case/__tests__/index.js
jeddy3: Thanks for adding this. It's much neater!

I've pushed up a commit to use `skip`.

ybiquitous (approved):
@jeddy3 Thank you so much for your great work! 👏🏼

jeddy3:
Changelog:

> - Removed: `syntax` option ([#5297](https://github.com/stylelint/stylelint/pull/5297)).

@ybiquitous No worries. Thank you for your spot-on suggestions and reviews!

PR #5300, opened 2021-05-13
Refactor `lib/__tests__/plugins.test.js`

ybiquitous:
Reduce callbacks as much as possible via `async/await` syntax.
(see also <https://jestjs.io/docs/asynchronous>)

A part of #4881 

I recommend viewing the diffs by ["Hide whitespace changes"](https://github.com/stylelint/stylelint/pull/5300/files?diff=unified&w=1).


jeddy3 (approved):
Looks great, thanks!

PR #5304, opened 2021-05-15
Update to PostCSS 8

jeddy3:
> Which issue, if any, is this issue related to?

Closes #4942
Closes #5276

> Is there anything in the PR that needs further explanation?

This is the long-lived branch to be merged in `v14`.

PR #5305, opened 2021-05-15
Fix invalid position for attribute selector

jeddy3:

Relates to https://github.com/stylelint/stylelint/pull/5304

The opening quote doesn't seem to be included in the `leftContext` anymore, so we need to account for it.


jeddy3:
Reduces the failing tests by one. I guess we can merge without the CI going green.

hudochenkov (approved):

ybiquitous (approved):

PR #5306, opened 2021-05-15
Update async calls to fs to use fs/promises (#5271)

### LINKED ISSUE #5271: Refactor via `fs/promises` module
We will be able to remove the following custom `readFile()` and replace one with the standard API  [`fsPromises.readFile()`](https://nodejs.org/api/fs.html#fs_fspromises_readfile_path_options).

https://github.com/stylelint/stylelint/blob/361c07084256e0340ede2ab328ac849b6e0ca98d/lib/getPostcssResult.js#L161-L171

Ref: #5215

The experimental warning for `fs.promises` has been removed since Node.js 10.17.0.
See also the [Node.js 10.17.0 changelog](https://github.com/nodejs/node/blob/d85929c32a0f657f1389ccc472121e1c8b3483ba/doc/changelogs/CHANGELOG_V10.md#2019-10-22-version-10170-dubnium-lts-bethgriggs).

qubyte:
Closes #5271

I hope it's ok that I've picked up this issue. It seemed like a small one to pick up. I've followed some existing converstions to fs/promises found in the repo. These look like:

```javascript
const { promises: fs } = require('fs');
```

When these are converted to ESM, they are simpler to do than the alternative `const fs = require('fs').promises`.

```javascript
import { promises as fs } from 'fs';
```

In later versions of node the alternative is available:

```javascript
import fs from 'node:fs/promises';
```

but node 12 does not support `'node:fs/promises'` or `'fs/promises'` as imports as later versions do.

I've targeted this PR at the v14 branch.

### DIFF
--- a/lib/__tests__/plugins.test.js
+++ b/lib/__tests__/plugins.test.js
@@ -195,7 +195,7 @@ describe('module providing an array of plugins', () => {
 	});
 });
 
-it('slashless plugin causes configuration error', () => {
+it('slashless plugin causes configuration error', async () => {
 	const config = {
 		plugins: [path.join(__dirname, 'fixtures/plugin-slashless-warn-about-foo')],
 		rules: {
@@ -203,17 +203,11 @@ it('slashless plugin causes configuration error', () => {
 		},
 	};
 
-	return postcss()
-		.use(stylelint(config))
-		.process('.foo {}', { from: undefined })
-		.then(() => {
-			throw new Error('should not have succeeded');
-		})
-		.catch((err) => {
-			expect(err.message.startsWith('stylelint v7+ requires plugin rules to be namespaced')).toBe(
-				true,
-			);
-		});
+	const postcssPromise = postcss().use(stylelint(config)).process('.foo {}', { from: undefined });
+
+	await expect(postcssPromise).rejects.toThrow(
+		/^stylelint v7\+ requires plugin rules to be namespaced/,
+	);
 });
 
 it('plugin with primary option array', () => {
--- a/lib/__tests__/standalone-cache.test.js
+++ b/lib/__tests__/standalone-cache.test.js
@@ -154,26 +154,16 @@ describe('standalone cache', () => {
 				expect(typeof cache.getKey(newFileDest) === 'undefined').toBe(true);
 			});
 	});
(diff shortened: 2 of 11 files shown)


qubyte:
Ah, forgot to note... An ESlint rule is being disabled for each require updated. This is because `eslint-config-stylelint` sets the node version to `>=10.18.0` for that rule. I'd be happy to open a PR for that package too if it's helpful.

hudochenkov:
@qubyte that would be great! This is what we going to support in v14 https://github.com/stylelint/stylelint/blob/1e4ca41e254037dc4c08e7125be29d9efccc6b21/package.json#L190

jeddy3:
@qubyte Thank you for your pull request. It's totally OK that you've picked up this issue. Any help getting `14.0.0` released is going to be appreciated!

I've asked @ybiquitous to review as they created the original issue.

qubyte:
I'll put together a PR for the eslint config. :+1:

ybiquitous (approved):
@qubyte Great! Thank you! 👏🏼

jeddy3:
Before merging, shall we release the new eslint config and update this pull request to remove the disable commands?

qubyte:
I've opened https://github.com/stylelint/eslint-config-stylelint/pull/132 to allow these comments to be dropped.

jeddy3:
Eslint config was released. You can now update and remove the comments.

qubyte:
I've integrated the new linter version into this PR. It seems to have revealed some issues with how jest/expect is used (I guess the underlying plugin was made more strict). I've done my best to fix those issues, but I think at least one test was broken in a way that was hidden.

jeddy3:
> It seems to have revealed some issues with how jest/expect is used (I guess the underlying plugin was made more strict).

My bad. I didn't notice that we'd previously merged dependabot pull requests that included major version updates of some configs we extend. 

Let's skip that failing test so that we can merge this pull request, and I'll create a new issue to patch that failing test (which will hopefully be picked up by someone familiar with the caching part of the engine).

qubyte:
No worries. I've added a skip for that test to the last commit.

jeddy3 (approved):
Many thanks!

PR #5308, opened 2021-05-16
Adds ignore: [rules] / [at-rules] to block-opening-brace-*-after

### LINKED ISSUE #5262: Newline after at-rule opening brace
When using single-line formatting, it would helpful if the selectors for at-rules could pushed to a new line.  For example: 
```
.container { padding: 0; margin: 5px }
@media (min-width: 992px) {
    .container { padding: 5px; margin: 0 }}
```
A proposed solution would be to add at-rule exceptions for the block-opening-brace-newline-after rule.  Or a separate at-rule-opening-brace-newline-after rule might be simpler.

RhnSharma:
Fixes #5262 
Hi @jeddy3, how does this look?
Thanks

### DIFF
--- a/lib/rules/block-opening-brace-newline-after/index.js
+++ b/lib/rules/block-opening-brace-newline-after/index.js
@@ -6,6 +6,7 @@ const beforeBlockString = require('../../utils/beforeBlockString');
 const blockString = require('../../utils/blockString');
 const hasBlock = require('../../utils/hasBlock');
 const hasEmptyBlock = require('../../utils/hasEmptyBlock');
+const optionsMatches = require('../../utils/optionsMatches');
 const rawNodeString = require('../../utils/rawNodeString');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
@@ -29,13 +30,22 @@ function rule(expectation, options, context) {
 			possible: ['always', 'always-multi-line', 'never-multi-line'],
 		});
 
-		if (!validOptions) {
+		if (
+			!validOptions ||
+			optionsMatches(options, 'ignore', 'rules') ||
+			optionsMatches(options, 'ignore', 'at-rules')
+		) {
 			return;
 		}
 
 		// Check both kinds of statement: rules and at-rules
-		root.walkRules(check);
-		root.walkAtRules(check);
+		if (!optionsMatches(options, 'ignore', 'rules')) {
+			root.walkRules(check);
+		}
+
+		if (!optionsMatches(options, 'ignore', 'at-rules')) {
+			root.walkAtRules(check);
+		}
 
 		function check(statement) {
 			// Return early if blockless or has an empty block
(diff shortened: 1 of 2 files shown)


lib/rules/block-opening-brace-newline-after/index.js
```diff
@@ -29,13 +30,22 @@ function rule(expectation, options, context) {
 			possible: ['always', 'always-multi-line', 'never-multi-line'],
 		});
 
-		if (!validOptions) {
+		if (
```
jeddy3, line 33: We should update the schema on line `30` instead of changing this conditional. Look at other rules with `ignore` secondary options do this.

lib/rules/block-opening-brace-newline-after/index.js
```diff
@@ -29,13 +30,22 @@ function rule(expectation, options, context) {
...
 		}
 
 		// Check both kinds of statement: rules and at-rules
-		root.walkRules(check);
-		root.walkAtRules(check);
+		if (!optionsMatches(options, 'ignore', 'rules')) {
+			root.walkRules(check);
+		}
+
+		if (!optionsMatches(options, 'ignore', 'at-rules')) {
```
jeddy3, line 46: The original issue was to only add `ignore: ["rules"]` to this rule. We should revert this change.

lib/rules/block-opening-brace-space-after/index.js
```diff
@@ -38,13 +39,22 @@ function rule(expectation, options, context) {
 			],
 		});
 
-		if (!validOptions) {
+		if (
```
jeddy3, line 42: We should update the schema instead.

lib/rules/block-opening-brace-space-after/index.js
```diff
@@ -38,13 +39,22 @@ function rule(expectation, options, context) {
...
+			optionsMatches(options, 'ignore', 'at-rules')
+		) {
 			return;
 		}
 
-		// Check both kinds of statements: rules and at-rules
-		root.walkRules(check);
-		root.walkAtRules(check);
+		// Check both kinds of statement: rules and at-rules
+		if (!optionsMatches(options, 'ignore', 'rules')) {
```
jeddy3, line 51: The original issue was to only add ignore: ["at-rules"] to this rule. We should revert this change.

jeddy3 (changes requested):
Thanks for the pull request. 

I've requested changes. Additionally, you'll need to:

- update the docs
- add tests

To both rules. You can look at another rule that has an `ignore` secondary option to see how this is done.

For the tests, one `accept` and one `reject` should suffice. For example, with `ignore: ["rules"]` pick two rejected tests: one for a rule and another for an at-rule. Then make sure the at-rule one still `reject`s but the rule one now `accept`s.

Lastly, please set to merge into the `v14` branch. We are preparing a major release and would like to avoid conflicts.

jeddy3:
Closing as inactive for now.

@RhnSharma You're more than welcome to reopen the pull request at any time with the requested changes from https://github.com/stylelint/stylelint/pull/5308#pullrequestreview-660449339.

PR #5315, opened 2021-05-18
Fix false positives for flex in length-zero-no-unit

### LINKED ISSUE #4831: Fix false positives for flex shorthand with 2 values in length-zero-no-unit

jeddy3:

Closes #4831
Closes #5024

Implements https://github.com/stylelint/stylelint/pull/5024#issuecomment-826058241

### DIFF
--- a/lib/rules/length-zero-no-unit/__tests__/index.js
+++ b/lib/rules/length-zero-no-unit/__tests__/index.js
@@ -221,6 +221,14 @@ testRule({
 			code: 'a { lIne-hEight: 0px; }',
 		},
 		{
+			code: 'a { flex: 0px; }',
+			description: 'ignore flex property',
+		},
+		{
+			code: 'a { Flex: 0px; }',
+			description: 'ignore cased flex property',
+		},
+		{
 			code: 'a { grid-auto-columns: 0fr; }',
 			description: 'ignore some grid properties with `fr`',
 		},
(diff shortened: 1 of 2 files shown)


mattxwang (approved):
Similarly short and sweet, LGTM!

ybiquitous (approved):
LGTM! 👍🏼

darwintantuco (approved):
👍

jeddy3:
> - Fixed: `length-zero-no-unit` false positives for `flex` property ([#5315](https://github.com/stylelint/stylelint/pull/5315)).

PR #4891, opened 2020-08-08
Add color-hex-alpha rule

ssivanatarajan:
> Which issue, if any, is this issue related to?

Have added a new rule as discussed in issue #4880. 

> Is there anything in the PR that needs further explanation?

No need

Issue #4880, opened 2020-07-26
Add color-hex-alpha


ssivanatarajan:
> What is the problem you're trying to solve?

Some of the browsers not supported #rgba and #rrggbbaa kind of hex codes for colors.
refer https://caniuse.com/#feat=css-rrggbbaa

> What solution would you like to see?

So it would be great if we have an option in color-hex-length to prevent 4 & 8 digit hex colors.

malsf21:
Hi @ssivanatarajan,

Thank you for the suggestion! I personally think that this is a reasonable feature to add, especially given that Internet Explorer seems to not support 4-character/8-character hex codes. However, I'll leave it to @stylelint/core to make a final judgement, especially on how to name the new option. Some options include `allow-four-character`, `allow-4-character`, or `allow-transparency-character`, though I'm not sure what's the best naming scheme.

I've labelled this issue as needs discussion, so we can decide on a consistent naming scheme. That being said, you can probably get started now; if you have the time, it would be great if you could add the option yourself! [The stylelint documentation has more information on how to add a rule](https://stylelint.io/developer-guide/rules#add-options).

jeddy3:
@ssivanatarajan Thanks for the request and for using the template.

Let's add a new rule. The `color-hex-length` rule is for enforcing the use of [what the spec calls](https://drafts.csswg.org/css-color/#hex-notation) "shorter variants". We shouldn't overload this rule to enforce whether the alpha channel is required or not.

- Name: `color-hex-alpha`
- Primary option: `"always"|"never"`
- Secondary options: None
- Autofixable: No
- Messages:
   - `"always"` -> `Expected alpha channel in "${hex}"`
   - `"never"` -> `Unexpected alpha channel in "${hex}"`
- Description: "Require or disallow alpha channel for hex colors"
- Section: "Limit language features" -> "Color"

Examples...

With `"color-hex-alpha": "always"`:

```
a { color: #aaa }
```

```console
1:8 Error Expected alpha channel in "#aaa" (color-hex-alpha)
```

With `"color-hex-alpha": "never"`:

```
a { background: linear-gradient(to left, #333a, red 100%}
```

```console
1:20 Error Unexpected alpha channel in "#333a" (color-hex-alpha)
```

I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

There are [steps on how to add a new rule](https://stylelint.io/developer-guide/rules#add-a-rule) in the Developer guide.

[...]

PR #5316, opened 2021-05-19
Add color-hex-alpha rule

jeddy3:

Closes https://github.com/stylelint/stylelint/pull/4891
Closes https://github.com/stylelint/stylelint/issues/4880

Implements the remaining requested changes in https://github.com/stylelint/stylelint/pull/4891.

Most of the tests were failing, so refactored to tidy things up and fixed the bugs.

### DIFF
--- a/lib/rules/color-hex-alpha/README.md
+++ b/lib/rules/color-hex-alpha/README.md
@@ -1,0 +1,66 @@
+# color-hex-alpha
+
+Require or disallow alpha channel for hex colors.
+
+<!-- prettier-ignore -->
+```css
+a { color: #ffff }
+/**            ↑
+ * This alpha channel */
+```
+
+## Options
+
+`string`: `"always"|"never"`
+
+### `"always"`
+
+The following patterns are considered violations:
+
+<!-- prettier-ignore -->
+```css
+a { color: #fff; }
+```
+
+<!-- prettier-ignore -->
+```css
+a { color: #ffffff; }
+```
+
+The following patterns are _not_ considered violations:
+
+<!-- prettier-ignore -->
+```css
+a { color: #fffa; }
+```
+
+<!-- prettier-ignore -->
+```css
+a { color: #ffffffaa; }
+```
+
+### `"never"`
+
+The following patterns are considered violations:
+
+<!-- prettier-ignore -->
+```css
+a { color: #fffa; }
+```
+
+<!-- prettier-ignore -->
+```css
+a { color: #ffffffaa; }
+```
+
+The following patterns are _not_ considered violations:
+
+<!-- prettier-ignore -->
+```css
+a { color: #fff; }
+```
+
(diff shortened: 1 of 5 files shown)


lib/rules/color-hex-alpha/README.md
```diff
@@ -0,0 +1,66 @@
+# color-hex-alpha
+
+Require or disallow alpha channel for hex colors.
+
+<!-- prettier-ignore -->
+```css
+a { color: #ffff }
```
ybiquitous, line 7: [suggestion] How about changing the alpha channel character for clarity and consistency with the following code examples? ⬇️ 

```suggestion
a { color: #fffa }
```

lib/rules/color-hex-alpha/__tests__/index.js
```diff
@@ -0,0 +1,92 @@
...
+			message: messages.unexpected('#ffff'),
+			line: 1,
+			column: 12,
+		},
+	],
+});
+
+testRule({
+	ruleName,
+	config: ['always'],
```
ybiquitous, line 70: [trivial] How about the following change for consistency with other test cases?

```suggestion
	config: 'always',
```

lib/rules/color-hex-alpha/index.js
```diff
@@ -0,0 +1,68 @@
...
+	};
+}
+
+function isUrlFunction({ type, value }) {
+	return type === 'function' && value === 'url';
+}
+
+function isHexColor({ type, value }) {
+	return type === 'word' && HEX.test(value);
+}
```
ybiquitous, line 60: [question] There are similar functions named `isHexColor` in other rules. Does it make sense to extract a utility function, like `lib/utils/isHexColor.js`? (maybe out of the scope of this PR)

https://github.com/stylelint/stylelint/blob/b01ed25dfa3e8231a976eef76bd5e81fb535b1e9/lib/rules/color-hex-case/index.js#L76-L78

https://github.com/stylelint/stylelint/blob/b01ed25dfa3e8231a976eef76bd5e81fb535b1e9/lib/rules/color-hex-length/index.js#L114-L116

ybiquitous (approved):
I've left some trivial comments, but it looks good to me. 👍🏼

lib/rules/color-hex-alpha/index.js
jeddy3: Good idea, but one for a refactor pull request.

We'll need to take into account that `color-hex-case` should check anything hex-like, whereas the other rules should only work on valid hex colours.

ybiquitous: Thanks, I see that regexes are slightly different for each rule with reason.
It seems reasonable to keep them as-is.

jeddy3:
> - Added: `color-hex-alpha` rule ([#5316](https://github.com/stylelint/stylelint/pull/5316)).

PR #5317, opened 2021-05-19
Add custom-property-no-missing-var-function rule

### LINKED ISSUE #5000: Add dashed-ident-no-missing-reference-function

jeddy3:

Closes #5000
Closes #5044

This replaces #5044.

I've thought about this rule on and off for a while now. I thought I'd unpacked it all in https://github.com/stylelint/stylelint/issues/5000#issuecomment-713702408, but there's more.

From the [`<dashed-ident>` spec](https://drafts.csswg.org/css-values-4/#dashed-idents):

> Some contexts accept both author-defined identifiers and CSS-defined identifiers. If not handled carefully, this can result in difficulties adding new CSS-defined values; UAs have to study existing usage and gamble that there are sufficiently few author-defined identifiers in use matching the new CSS-defined one, so giving the new value a special CSS-defined meaning won’t break existing pages.

> While there are many legacy cases in CSS that mix these two values spaces in exactly this fraught way, the `<dashed-ident>` type is meant to be an easy way to distinguish author-defined identifiers from CSS-defined identifiers.

I believe we'll see more unreferenced _dashed-idents_ in the future. For example, if [`list-style-type` property](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type) was designed today, I believe _dashed-idents_ would be required for the `<custom-idents>` to avoid conflicts with CSS-defined identifiers like `disc`, e.g.:

```css
@counter-style --disc {}
a { list-style-type: --disc }
```

The only way to confidently know if a _dashed-ident_ is a _custom property_, and should therefore be referenced with the `var` function, is to check if it has been defined as one.

I've updated the rule to do this and renamed it accordingly. Similarly to the [`no-unknown-animations` rule](https://stylelint.io/user-guide/rules/no-unknown-animations), it has the limitation of only catching unreferenced custom properties that are defined within the same source. Unlike the `no-unknown-animations` rule, these manifest as false negatives rather than false positives. As such, I think we're OK to add the rule with this limitation as it'll be useful to some users, i.e. those using locally scoped variables or defining their variables at the top of the same source. We can remove the limitation when we find a way to provide context for rules like these.

At the moment, it'll catch:

```css
/* same source at-properties */
@property --foo {
  syntax: "<color>";
  initial-value: red;
}
a { color: --foo; }
```

```css
/* same source custom properties */
:root { --foo: red; }
a { color: --foo; }
```

```css
/* locally scoped custom properties */
a { 
  --foo: red;
  color: --foo;
}
```

While avoiding any false positives for unreferenced dashed-idents. Which I think is good enough.

### DIFF
--- a/lib/rules/custom-property-no-missing-var-function/index.js
+++ b/lib/rules/custom-property-no-missing-var-function/index.js
@@ -26,7 +26,7 @@ function rule(actual) {
 
 		const customProperties = new Set();
 
-		root.walkAtRules(/property/i, (atRule) => {
+		root.walkAtRules(/^property$/i, (atRule) => {
 			customProperties.add(atRule.params);
 		});
 
--- a/lib/rules/custom-property-no-missing-var-function/index.js
+++ b/lib/rules/custom-property-no-missing-var-function/index.js
@@ -1,0 +1,76 @@
+// @ts-nocheck
+
+'use strict';
+
+const valueParser = require('postcss-value-parser');
+
+const declarationValueIndex = require('../../utils/declarationValueIndex');
+const isCustomProperty = require('../../utils/isCustomProperty');
+const report = require('../../utils/report');
+const ruleMessages = require('../../utils/ruleMessages');
+const validateOptions = require('../../utils/validateOptions');
+
+const ruleName = 'custom-property-no-missing-var-function';
+
+const messages = ruleMessages(ruleName, {
+	rejected: (customProperty) => `Unexpected missing var function for "${customProperty}"`,
+});
+
+function rule(actual) {
+	return (root, result) => {
+		const validOptions = validateOptions(result, ruleName, {
+			actual,
+		});
+
+		if (!validOptions) return;
+
+		const customProperties = new Set();
+
+		root.walkAtRules(/property/i, (atRule) => {
+			customProperties.add(atRule.params);
+		});
+
+		root.walkDecls(({ prop }) => {
+			if (isCustomProperty(prop)) customProperties.add(prop);
+		});
+
+		root.walkDecls((decl) => {
+			const { value } = decl;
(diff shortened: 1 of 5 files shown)


alexander-akait:
The name is sounds better :star:

lib/rules/custom-property-no-missing-var-function/index.js
```diff
@@ -0,0 +1,76 @@
...
+	return (root, result) => {
+		const validOptions = validateOptions(result, ruleName, {
+			actual,
+		});
+
+		if (!validOptions) return;
+
+		const customProperties = new Set();
+
+		root.walkAtRules(/property/i, (atRule) => {
```
ybiquitous, line 29: [question] Does it make sense to check an at-rule name more strictly as below?

```suggestion
		root.walkAtRules(/^property$/i, (atRule) => {
```

ybiquitous (approved):
Thanks for the detailed explanation. I understand well.
 I've left a trivial question, but this rule addition looks like no problem to me. 👍🏼

lib/rules/custom-property-no-missing-var-function/index.js
jeddy3: Good catch. It's a mistake.

jeddy3:
> - Added: `custom-property-no-missing-var-function` rule ([#5316](https://github.com/stylelint/stylelint/pull/5317)).

PR #5323, opened 2021-05-26
Remove `_.last`

stephenwade:

Ref #4412.

No.

### DIFF
--- a/lib/assignDisabledRanges.js
+++ b/lib/assignDisabledRanges.js
@@ -253,8 +253,8 @@ module.exports = function (root, result) {
 					});
 				}
 
-				Object.keys(disabledRanges).forEach((ruleName) => {
-					if (!_.get(_.last(disabledRanges[ruleName]), 'end')) {
+				Object.entries(disabledRanges).forEach(([ruleName, ranges]) => {
+					if (!_.get(ranges[ranges.length - 1], 'end')) {
 						endDisabledRange(endLine, ruleName, ruleName === ALL_RULES);
 					}
 				});
@@ -269,7 +269,8 @@ module.exports = function (root, result) {
 						createDisableRange(comment, start, false, description, end, false),
 					);
 				} else {
-					const range = _.last(disabledRanges[ALL_RULES]);
+					const ranges = disabledRanges[ALL_RULES];
+					const range = ranges[ranges.length - 1];
 
 					if (range) {
 						disabledRanges[ruleToEnable].push({ ...range });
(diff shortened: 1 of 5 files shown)


lib/assignDisabledRanges.js
```diff
@@ -253,8 +253,8 @@ module.exports = function (root, result) {
 					});
 				}
 
-				Object.keys(disabledRanges).forEach((ruleName) => {
-					if (!_.get(_.last(disabledRanges[ruleName]), 'end')) {
+				Object.entries(disabledRanges).forEach(([ruleName, ranges]) => {
+					if (!_.get(ranges[ranges.length - 1], 'end')) {
```
ybiquitous, line 257: [question] What do you think about using the `slice(-1)` idiom? Is it a bit tricky?

```suggestion
					if (!_.get(ranges.slice(-1), 'end')) {
```

See also <https://learnersbucket.com/examples/array/how-to-get-the-last-element-of-the-array-in-javascript/>

lib/assignDisabledRanges.js
```diff
@@ -269,7 +269,8 @@ module.exports = function (root, result) {
 						createDisableRange(comment, start, false, description, end, false),
 					);
 				} else {
-					const range = _.last(disabledRanges[ALL_RULES]);
+					const ranges = disabledRanges[ALL_RULES];
+					const range = ranges[ranges.length - 1];
```
ybiquitous, line 273: [suggestion] [`_.last()`](https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L7609-L7612) is null-safe, so how about adding a null-check as below, just in case?

```suggestion
					const range = ranges == null ? null : ranges[ranges.length - 1];
```

lib/assignDisabledRanges.js
stephenwade: `array.slice(-1)` returns an array with one element. We'd have to use `array.slice(-1)[0]`, which I don't think is as readable as `array[array.length - 1]`.

ybiquitous: Thanks. It makes sense. 👍🏼

lib/assignDisabledRanges.js
stephenwade: I added a few more null checks in ac07653d.

ybiquitous (approved):
Thank you! LGTM 👍🏼

jeddy3 (approved):
LGTM, thanks!

PR #5327, opened 2021-05-28
Refactor to remove some uses of `_.get` and `_.set`

stephenwade:

Ref #4412.

Most of the code is self-explanatory. I replaced the simpler uses of `_.get` and `_.set` with native code.

I had to change a couple of `array.forEach` loops to regular `for … of array` loops to make `tsc` happy. I also updated the JSDoc types in a couple places for the same reason.

### DIFF
--- a/lib/assignDisabledRanges.js
+++ b/lib/assignDisabledRanges.js
@@ -1,7 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
-
 const COMMAND_PREFIX = 'stylelint-';
 const disableCommand = `${COMMAND_PREFIX}disable`;
 const enableCommand = `${COMMAND_PREFIX}enable`;
@@ -254,7 +252,9 @@ module.exports = function (root, result) {
 				}
 
 				Object.entries(disabledRanges).forEach(([ruleName, ranges]) => {
-					if (!_.get(ranges[ranges.length - 1], 'end')) {
+					const lastRange = ranges[ranges.length - 1];
+
+					if (!lastRange || !lastRange.end) {
 						endDisabledRange(endLine, ruleName, ruleName === ALL_RULES);
 					}
 				});
@@ -404,7 +404,7 @@ module.exports = function (root, result) {
 
 		if (!ranges[ranges.length - 1]) return false;
 
-		if (!_.get(ranges[ranges.length - 1], 'end')) return true;
+		if (!ranges[ranges.length - 1].end) return true;
 
 		return false;
 	}
--- a/lib/assignDisabledRanges.js
+++ b/lib/assignDisabledRanges.js
@@ -402,9 +402,11 @@ module.exports = function (root, result) {
 
 		if (!ranges) return false;
 
-		if (!ranges[ranges.length - 1]) return false;
+		const lastRange = ranges[ranges.length - 1];
 
-		if (!ranges[ranges.length - 1].end) return true;
+		if (!lastRange) return false;
+
+		if (!lastRange.end) return true;
 
 		return false;
 	}
--- a/lib/normalizeAllRuleSettings.js
+++ b/lib/normalizeAllRuleSettings.js
@@ -16,9 +16,7 @@ function normalizeAllRuleSettings(config) {
 
 	if (!config.rules) return config;
 
-	for (const ruleName of Object.keys(config.rules)) {
-		const rawRuleSettings = config.rules[ruleName];
-
+	for (const [ruleName, rawRuleSettings] of Object.entries(config.rules)) {
 		const rule = rules[ruleName] || (config.pluginFunctions && config.pluginFunctions[ruleName]);
 
 		if (!rule) {
--- a/lib/normalizeAllRuleSettings.js
+++ b/lib/normalizeAllRuleSettings.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
 const normalizeRuleSettings = require('./normalizeRuleSettings');
 const rules = require('./rules');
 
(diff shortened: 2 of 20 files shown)


lib/assignDisabledRanges.js
```diff
@@ -404,7 +404,7 @@ module.exports = function (root, result) {
 
 		if (!ranges[ranges.length - 1]) return false;
 
-		if (!_.get(ranges[ranges.length - 1], 'end')) return true;
+		if (!ranges[ranges.length - 1].end) return true;
```
ybiquitous, line 407: [suggestion] How about reducing the duplication as below?

```js
const lastRange = ranges[ranges.length - 1];
if (!lastRange) return false;
if (!lastRange.end) return true;
```

lib/normalizeAllRuleSettings.js
```diff
@@ -17,21 +16,21 @@ function normalizeAllRuleSettings(config) {
 
 	if (!config.rules) return config;
 
-	Object.keys(config.rules).forEach((ruleName) => {
-		const rawRuleSettings = _.get(config, ['rules', ruleName]);
+	for (const ruleName of Object.keys(config.rules)) {
+		const rawRuleSettings = config.rules[ruleName];
```
ybiquitous, line 21: [suggestion] Using `for...of` is nice! 👍🏼 

In addition, how about using `Object.entries()` for more simplicity?

```suggestion
	for (const [ruleName, rawRuleSettings] of Object.entries(config.rules)) {
```

lib/reportDisables.js
```diff
@@ -25,19 +25,21 @@ module.exports = function (results) {
...
-			return [];
+		if (!Object.values(config.rules).some(reportDisablesForRule)) {
+			return;
 		}
 
-		Object.keys(rangeData).forEach((rule) => {
-			rangeData[rule].forEach((range) => {
-				if (!reportDisablesForRule(_.get(config, ['rules', rule], []))) return;
+		for (const rule of Object.keys(rangeData)) {
+			for (const range of rangeData[rule]) {
```
ybiquitous, line 37: [suggestion] Same here:

```suggestion
		for (const [rule, ranges] of Object.entries(rangeData)) {
			for (const range of ranges) {
```

lib/rules/max-line-length/index.js
```diff
@@ -53,12 +53,11 @@ function rule(maxLength, options, context) {
...
 		EXCLUDED_PATTERNS.forEach((pattern) =>
 			execall(pattern, rootString).forEach((match) => {
-				const startOfSubString =
-					match.index + match.match.indexOf(_.get(match, 'subMatches[0]', ''));
+				const startOfSubString = match.index + match.match.indexOf(match.subMatches[0] || '');
 
 				return skippedSubStrings.push([
 					startOfSubString,
-					startOfSubString + _.get(match, 'subMatches[0].length', 0),
+					startOfSubString + (match.subMatches[0] ? match.subMatches[0].length : 0),
```
ybiquitous, line 60: [suggestion] Can we simplify the code more?

```js
const subMatch = match.subMatches[0] || '';
const startOfSubString = match.index + match.match.indexOf(subMatch);

return skippedSubStrings.push([
	startOfSubString,
	startOfSubString + subMatch.length,
]);
```

lib/utils/report.js
```diff
@@ -83,7 +81,11 @@ module.exports = function (violation) {
 		}
 	}
 
-	const severity = _.get(result.stylelint, ['ruleSeverities', ruleName]);
+	if (!result.stylelint.ruleSeverities) {
+		result.stylelint.ruleSeverities = {};
+	}
+
+	const severity = result.stylelint.ruleSeverities[ruleName];
```
ybiquitous, line 88: [suggestion] IMO, it seems safe to avoid the side-effect `result.stylelint.ruleSeverities = {}` here. (I'm not sure if `result.stylelint.ruleSeverities` is falsy or not, though)

How about this? 👇🏼 

```suggestion
	const severity = result.stylelint.ruleSeverities && result.stylelint.ruleSeverities[ruleName];
```

lib/utils/report.js
```diff
@@ -107,7 +109,11 @@ module.exports = function (violation) {
 		warningProperties.word = word;
 	}
 
-	const warningMessage = _.get(result.stylelint, ['customMessages', ruleName], message);
+	if (!result.stylelint.customMessages) {
+		result.stylelint.customMessages = {};
+	}
+
+	const warningMessage = result.stylelint.customMessages[ruleName] || message;
```
ybiquitous, line 116: [sugggestion] Same here:

```suggestion
	const warningMessage = (result.stylelint.customMessages && result.stylelint.customMessages[ruleName]) || message;
```

ybiquitous (commented):
@stephenwade Thank you for making this change! 👏🏼 
Could you check my review comments, please?

lib/utils/report.js
stephenwade: I think I was trying to shorten the line length here by setting the property, but I agree it's better to use code that doesn't add a side effect.

stephenwade:
Hi @ybiquitous, I've implemented your suggested changes.

lib/utils/validateOptions.js
```diff
@@ -42,7 +42,12 @@ module.exports = function (result, ruleName, ...optionDescriptions) {
 		result.warn(message, {
 			stylelintType: 'invalidOption',
 		});
-		_.set(result, 'stylelint.stylelintError', true);
+		result.stylelint = result.stylelint || {
+			disabledRanges: {},
+			ruleSeverities: {},
+			customMessages: {},
+		};
```
ybiquitous, line 49: [note] I'm a bit concerned about the newly added properties like `disabledRanges`, but it seems due to our TypeScript definitions (maybe another issue). 🤔 

Anyway, I think no problems going on as-is. 👍🏼

ybiquitous (approved):
@stephenwade Thank you so much! LGTM 👍🏼

lib/utils/validateOptions.js
stephenwade: Yes, that was because of the TypeScript definitions. It wouldn't let me just set `result.stylelint` to `{}`.

jeddy3 (approved):
Thanks!

PR #5329, opened 2021-05-29
Refactor to remove `_.remove`

stephenwade:

Ref #4412.

No.

### DIFF
--- a/lib/needlessDisables.js
+++ b/lib/needlessDisables.js
@@ -55,19 +55,19 @@ module.exports = function (results) {
 			}
 		}
 
-		const rangeEntries = Object.entries(rangeData);
-
 		// Get rid of the duplicated ranges for each `all` rule. We only care
 		// if the entire `all` rule is useful as a whole or not.
 		for (const range of rangeData.all) {
-			for (const [rule, ranges] of rangeEntries) {
+			for (const rule of Object.keys(rangeData)) {
 				if (rule === 'all') continue;
 
-				_.remove(ranges, (otherRange) => range.comment === otherRange.comment);
+				rangeData[rule] = rangeData[rule].filter(
+					(otherRange) => range.comment !== otherRange.comment,
+				);
 			}
 		}
 
-		for (const [rule, ranges] of rangeEntries) {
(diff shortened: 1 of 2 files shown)


lib/needlessDisables.js
```diff
@@ -55,19 +55,19 @@ module.exports = function (results) {
 			}
 		}
 
-		const rangeEntries = Object.entries(rangeData);
-
 		// Get rid of the duplicated ranges for each `all` rule. We only care
 		// if the entire `all` rule is useful as a whole or not.
 		for (const range of rangeData.all) {
-			for (const [rule, ranges] of rangeEntries) {
+			for (const rule of Object.keys(rangeData)) {
```
ybiquitous, line 61: [question] `Object.keys(rangeData)` will be evaluated multiple times in the loop. Will the impact on performance small enough?

ybiquitous (approved):
Thank you! I have left a question comment, but it looks good to me. 👏🏼

lib/needlessDisables.js
stephenwade: I think the performance impact will be very small. I tested in Node and found that running `Object.keys` on an object with 10,000 keys takes 1 ms on my computer, and I think `rangeData` is not that big.

<details>
<summary>My test code</summary>

```js
const randomString = () => Math.random().toString(36).substring(2);
const o1k = {};
for (let i = 0; i < 1_000; ++i) { o1k[randomString()] = randomString(); }
const o10k = {};
for (let i = 0; i < 10_000; ++i) { o10k[randomString()] = randomString(); }
const o100k = {};
for (let i = 0; i < 100_000; ++i) { o100k[randomString()] = randomString(); }
let date1, date2;

date1 = new Date(); Object.keys(o1k); date2 = new Date(); void 0;
console.log(date2 - date1); // 0

date1 = new Date(); Object.keys(o10k); date2 = new Date(); void 0;
console.log(date2 - date1); // 1

date1 = new Date(); Object.keys(o100k); date2 = new Date(); void 0;
console.log(date2 - date1); // 17
```

</details>

ybiquitous: Thank you for the investigation! 👍🏼

PR #5333, opened 2021-05-31
Add CHANGELOG to files

jeddy3:

https://github.com/stylelint/stylelint.io/pull/193

It seems we may need to explicitly add CHANGELOG to files for it to appear when installing from GitHub...

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   },
   "files": [
     "bin",
+    "CHANGELOG.md",
     "CONTRIBUTING.md",
     "SECURITY.md",
     "docs",


ybiquitous (approved):
LGTM 😅 👍🏼 

But, it's weird because the npm doc says:

> Certain files are always included, regardless of settings:

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#files

jeddy3:
Yep, it's weird. I don't know why the behaviour is different when installing from GitHub. 

No harm in being explicit, though.

PR #5334, opened 2021-05-31
Refactor to remove `_.partial`

stephenwade:

Ref #4412.

`_.partial` does not modify the `this` binding, but `fn.bind` does. I verified that none of the functions touched in this PR use `this`, so the behavior difference is not an issue.

### DIFF
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -145,7 +145,7 @@ module.exports = function (options) {
 						return stylelint._createStylelintResult(postcssResult);
 					});
 			})
-			.catch(handleError.bind(null, stylelint))
+			.catch((error) => handleError(stylelint, error))
 			.then((stylelintResult) => {
 				const postcssResult = stylelintResult._postcssResult;
 				const returnValue = prepareReturnValue([stylelintResult], options, formatterFunction);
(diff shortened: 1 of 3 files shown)


lib/standalone.js
```diff
@@ -146,7 +145,7 @@ module.exports = function (options) {
 						return stylelint._createStylelintResult(postcssResult);
 					});
 			})
-			.catch(_.partial(handleError, stylelint))
+			.catch(handleError.bind(null, stylelint))
```
ybiquitous, line 148: [suggestion] IMO, it seems straightforward to me to call just `handleError()` here, instead of using `.bind()`. What do you think?

```suggestion
			.catch((error) => {
				return handleError(stylelint, error);
			})
```

ybiquitous (approved):
@stephenwade Thank you for the improvement! I've left a trivial suggestion, but this PR looks good to me. 👍🏼

lib/standalone.js
stephenwade: Hi @ybiquitous, that's a good idea. I did this in 94135f04.

ybiquitous: Thank you! 😄

ybiquitous (approved):

PR #5341, opened 2021-06-03
Refactor to remove type validation functions

stephenwade:

Ref #4412.

This PR removes the following type validation functions that are used in most rules to check types:
- `_.isBoolean`
- `_.isNumber`
- `_.isObject`
- `_.isRegExp`
- `_.isString`

Since these are frequently used together in calls to `validateOptions`, I thought it would be better to combine them into one PR.

This PR adds the 4 simple functions in [`lib/utils/validateTypes.js`](https://github.com/stylelint/stylelint/pull/5341/files#diff-4c56d4c1a85a6bd394b261b4f742fa1f3040377022f7917cfb974a8833ef3618). It also adds the [`is-plain-object`](https://www.npmjs.com/package/is-plain-object) package because that is a more complicated check to write manually – see https://github.com/stylelint/stylelint/pull/5341#discussion_r645215098.

### DIFF
--- a/lib/utils/validateTypes.js
+++ b/lib/utils/validateTypes.js
@@ -19,15 +19,6 @@ function isNumber(value) {
 }
 
 /**
- * Checks if the value is a plain object.
- * @param {any} value
- * @returns {boolean}
- */
-function isPlainObject(value) {
-	return Object.prototype.toString.call(value) === '[object Object]';
-}
-
-/**
  * Checks if the value is a RegExp object.
  * @param {any} value
  * @returns {boolean}
@@ -48,7 +39,6 @@ function isString(value) {
 module.exports = {
 	isBoolean,
 	isNumber,
-	isPlainObject,
 	isRegExp,
 	isString,
 };
--- a/lib/utils/validateTypes.js
+++ b/lib/utils/validateTypes.js
@@ -19,12 +19,12 @@ function isNumber(value) {
 }
 
 /**
- * Checks if the value is the language type of Object.
+ * Checks if the value is a plain object.
  * @param {any} value
  * @returns {boolean}
  */
-function isObject(value) {
-	return value instanceof Object;
+function isPlainObject(value) {
+	return Object.prototype.toString.call(value) === '[object Object]';
 }
 
 /**
@@ -48,7 +48,7 @@ function isString(value) {
 module.exports = {
 	isBoolean,
 	isNumber,
-	isObject,
+	isPlainObject,
 	isRegExp,
 	isString,
 };
--- a/lib/utils/validateTypes.js
+++ b/lib/utils/validateTypes.js
@@ -1,0 +1,54 @@
+'use strict';
+
+/**
+ * Checks if the value is a boolean or a Boolean object.
+ * @param {any} value
+ * @returns {boolean}
+ */
+function isBoolean(value) {
+	return typeof value === 'boolean' || value instanceof Boolean;
+}
+
+/**
+ * Checks if the value is a number or a Number object.
+ * @param {any} value
+ * @returns {boolean}
+ */
+function isNumber(value) {
+	return typeof value === 'number' || value instanceof Number;
+}
+
+/**
+ * Checks if the value is the language type of Object.
+ * @param {any} value
+ * @returns {boolean}
+ */
+function isObject(value) {
+	return value instanceof Object;
+}
+
+/**
+ * Checks if the value is a RegExp object.
+ * @param {any} value
+ * @returns {boolean}
+ */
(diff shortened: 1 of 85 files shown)


ybiquitous:
@stephenwade This is a great job! I agree with your thought to create `lib/utils/validateTypes.js`. 👍🏼

lib/utils/validateTypes.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/**
+ * Checks if the value is the language type of Object.
+ * @param {any} value
+ * @returns {boolean}
+ */
+function isObject(value) {
+	return value instanceof Object;
+}
```
ybiquitous, line 28: [suggestion] I wonder if this function should be `isPlainObject` (#5336) because `value instanceof Object` can be applied to almost all objects:

```console
$ node
Welcome to Node.js v16.2.0.
Type ".help" for more information.
> new Function() instanceof Object
true
> new RegExp() instanceof Object
true
> new String() instanceof Object
true
```

So, I think `isPlainObject` should be below according to #5336:

```js
function isPlainObject(value) {
	return Object.prototype.toString.call(value) === '[object Object]';
}
```

What do you think?

stephenwade: @ybiquitous Yes, that should be `isPlainObject`. My change copies the original behavior but you're right that it allowed too many objects. I've made the change in 34f948fc.

lib/utils/validateTypes.js
```diff
@@ -0,0 +1,54 @@
...
+function isNumber(value) {
+	return typeof value === 'number' || value instanceof Number;
+}
+
+/**
+ * Checks if the value is a plain object.
+ * @param {any} value
+ * @returns {boolean}
+ */
+function isPlainObject(value) {
```
alexander-akait, line 26: Here right implementation https://github.com/jonschlinkert/is-plain-object/blob/master/is-plain-object.js

stephenwade: I think the simple check is okay.

ybiquitous: @alexander-akait Thanks for sharing it. I also thought the simple check is enough, but implementing `isPlainObject` seems more challenging than I thought.

So, I try the [`is-plain-object` package's test cases](https://github.com/jonschlinkert/is-plain-object/blob/0a47f0f6cd10e0d2489beb55a32a8d0ba7b04b25/test/server.js) also in our codebase:

```js
// lib/isPlainObject.test.js
const assert = require('assert');
const { isPlainObject } = require('./utils/validateTypes');

test('isPlainObject', () => {
	assert(isPlainObject(Object.create({})));
	assert(isPlainObject(Object.create(Object.prototype)));
	assert(isPlainObject({ foo: 'bar' }));
	assert(isPlainObject({}));
	assert(isPlainObject(Object.create(null)));

	function Foo() {
		this.abc = {};
	}

	assert(!isPlainObject(/foo/));
	assert(!isPlainObject(function () {}));
	assert(!isPlainObject(1));
	assert(!isPlainObject(['foo', 'bar']));
	assert(!isPlainObject([]));
	assert(!isPlainObject(new Foo()));
	assert(!isPlainObject(null));
});
```

As a result, the test fails:

```console
$ npx jest lib/isPlainObject.test.js
 FAIL  lib/isPlainObject.test.js
  ✕ isPlainObject (8 ms)

  ● isPlainObject

    assert(received)

    Expected value to be equal to:
      true
    Received:
      false

      18 | 	assert(!isPlainObject(['foo', 'bar']));
      19 | 	assert(!isPlainObject([]));
    > 20 | 	assert(!isPlainObject(new Foo()));
         | 	^
      21 | 	assert(!isPlainObject(null));
      22 | });
      23 |

      at Object.<anonymous> (lib/isPlainObject.test.js:20:2)

Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        0.692 s, estimated 1 s
Ran all test suites matching /lib\/isPlainObject.test.js/i.
```

Thus, I think it seems better to install [`is-plain-object`](https://www.npmjs.com/package/is-plain-object) for compatibility. Fortunately, that package supports ESM, and has no dependencies and its type definition also.

What do you think?

stephenwade: Thank you @alexander-akait and @ybiquitous for investigating! I also didn't realize it would be challenging.

Given this, I think it is a good idea to install `is-plain-object`. This is done in 8ec74f6c.

ybiquitous (approved):
@stephenwade Thank you! Looks good to me! 👏🏼

jeddy3 (approved):
Fantastic, thank you!

PR #5345, opened 2021-06-05
Removes DisableReportRange.unusedRule from types

### LINKED ISSUE #4899: Remove DisableReportRange.unusedRule
(Added by request in https://github.com/stylelint/stylelint/pull/4897/files#r470950362)

#4897 renames the `DisableReportRange.unusedRule` field to `.rule`, since it's used to refer to rules in multiple contexts (rather than just when they're unused). We're keeping the old field name around for the time being to ensure backwards-compatibility, but in the next major version release the old name should be removed entirely.

mattxwang:

Closes #4899.

Relatively short PR that follows up on the change made in #4897. Pretty sure this is a quick and simple change, but let me know if I've missed anything!

### DIFF
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -256,11 +256,6 @@ declare module 'stylelint' {
 		rule: string;
 		start: number;
 		end?: number;
-
-		// This is for backwards-compatibility with formatters that were written
-		// when this name was used instead of `rule`. It should be avoided for new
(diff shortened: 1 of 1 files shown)


ybiquitous (approved):
I tried `git grep unusedRule` but no matches left. LGTM 👍🏼

jeddy3 (approved):
Thanks!

PR #5350, opened 2021-06-06
Refactor to remove `_.groupBy`

stephenwade:

Ref #4412.

Since `_.groupBy` is used twice in this file and only in this file, I avoided repeating code by adding a `groupBy` function to this file.


ybiquitous (approved):
@stephenwade Thank. LGTM 👍🏼

PR #5353, opened 2021-06-07
Migrate stylelint PostCSS plugin to Visitors API

hudochenkov:

Related to #4942

That was easier than I though. Basically I needed to change only our PostCSS plugin part to Visitors API to avoid warnings in tests. Nothing changes in execution order.

I think for us to embrace Visitors API in rules many changes needs to be done. Currently `stylelint` is not a PostCSS plugin. It uses PostCSS internals like `LazyResult` to go around PostCSS API. And then we call rule functions, which looks like PostCSS 7- plugins, but not exactly. So we don't use PostCSS API to parse code and run rules as plugins.

Project to use Visitors API could be a big refactor, and change in how stylelint works internally. I think we can do it after `v14` release.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -12,22 +12,19 @@ const standalone = require('./standalone');
 const validateOptions = require('./utils/validateOptions');
 
 /**
- * TODO TYPES change any to appropriated options
- * @type {import('postcss').Plugin & Partial<import('stylelint').StylelintPublicAPI>}
+ * @type {import('postcss').PluginCreator<import('stylelint').PostcssPluginOptions> & import('stylelint').StylelintPublicAPI}
  */
-const api = postcssPlugin;
+module.exports = postcssPlugin;
 
-api.utils = {
+module.exports.utils = {
 	report,
 	ruleMessages,
 	validateOptions,
 	checkAgainstRule,
 };
 
-api.lint = standalone;
-api.rules = rules;
-api.formatters = formatters;
-api.createPlugin = createPlugin;
-api.createLinter = createStylelint;
-
-module.exports = api;
+module.exports.lint = standalone;
+module.exports.rules = rules;
+module.exports.formatters = formatters;
+module.exports.createPlugin = createPlugin;
+module.exports.createLinter = createStylelint;
--- a/lib/postcssPlugin.js
+++ b/lib/postcssPlugin.js
@@ -1,26 +1,41 @@
-// @ts-nocheck
 'use strict';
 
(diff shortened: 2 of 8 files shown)


lib/postcssPlugin.js
```diff
@@ -1,26 +1,41 @@
...
-const postcss = require('postcss');
 
-module.exports = postcss.plugin('stylelint', (options = {}) => {
-	const tailoredOptions = options.rules ? { config: options } : options;
-	const stylelint = createStylelint(tailoredOptions);
+/** @typedef {import('stylelint').PostcssPluginOptions} PostcssPluginOptions */
+/** @typedef {import('stylelint').StylelintConfig} StylelintConfig */
 
-	return (root, result) => {
-		let filePath = options.from;
```
hudochenkov, line 13: It's not clear what `options.from` supposed to be used for. It's not documented, and there are no tests that would use it. Couldn't find anything in Git history. [Trace](https://github.com/stylelint/stylelint/blame/ebc9e6a69bed159982eee5e0678d29b5b23747c6/lib/postcssPlugin.js) is lost on `Drop babel` commit.

lib/index.js
```diff
@@ -12,22 +12,19 @@ const standalone = require('./standalone');
 const validateOptions = require('./utils/validateOptions');
 
 /**
- * TODO TYPES change any to appropriated options
- * @type {import('postcss').Plugin & Partial<import('stylelint').StylelintPublicAPI>}
+ * @type {import('postcss').PluginCreator<import('stylelint').PostcssPluginOptions> & import('stylelint').StylelintPublicAPI}
  */
-const api = postcssPlugin;
+module.exports = postcssPlugin;
```
hudochenkov, line 17: Allow to remove `Partial` workaround.

jeddy3 (approved):
Fantastic, thank you!

> Project to use Visitors API could be a big refactor, and change in how stylelint works internally. I think we can do it after v14 release.

I agree. It's one for the bucket list.

ybiquitous (approved):
Great! LGTM 👏🏼 

I have just a question. Do we need to change the following document on the v14 release?
https://github.com/stylelint/stylelint/blob/v14/docs/user-guide/usage/postcss-plugin.md

jeddy3:
> Do we need to change the following document on the v14 release?
https://github.com/stylelint/stylelint/blob/v14/docs/user-guide/usage/postcss-plugin.md

I believe the behaviour is unchanged.

ybiquitous:
I see. Thanks. 👌

PR #5356, opened 2021-06-07
Fix declaration block semicolon newline after false positives

### LINKED ISSUE #5355: Fix false positives for CSS-in-JS in declaration-block-semicolon-newline-after
> Clearly describe the bug

There are `declaration-block-semicolon-newline-after` false positives for css-in-js syntax. 

`declaration-block-semicolon-newline-after`

```jsx
const Component = () => (
  <div style={{ color: pink, top: 0 }} />
);
```

```json
{
  "rules": {
    "declaration-block-semicolon-newline-after": "always"
  }
}
```

`13.8.0`

CLI with `stylelint "**/*.js" --config myconfig.json`

Yes, it's related to css-in-js.

No warnings to be flagged.

```shell
Expected newline after ";" (declaration-block-semicolon-newline-after)
```

dcalhoun:

Fixes #5355. 

This changes the rule to ignore css-in-js object literals, as the object literals should not contain semicolons and therefore should not require a newline. Previously, the rule required a newline after each comma within the object.

### DIFF
--- a/lib/rules/declaration-block-semicolon-newline-after/__tests__/index.js
+++ b/lib/rules/declaration-block-semicolon-newline-after/__tests__/index.js
@@ -344,3 +344,45 @@ testRule({
 		},
 	],
 });
+
+testRule({
+	ruleName,
+	config: ['always'],
+	fix: true,
+	syntax: 'css-in-js',
+
+	accept: [
+		{
+			code: `<div style={{ color: pink, top: 0 }} />`,
+			description: 'css-in-js object',
+		},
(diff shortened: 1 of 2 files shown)


jeddy3:
@dcalhoun Many thanks for contributing and opening the pull request.

I'm going to close for now the way syntaxes work is changing in the next major release, see https://github.com/stylelint/stylelint/issues/5355#issuecomment-856682514, and I think we'll need to decide on the best way forward a little down the line.

PR #5358, opened 2021-06-07
Refactor to remove `_.compact`

stephenwade:

Ref #4412.

No.

### DIFF
--- a/lib/rules/named-grid-areas-no-invalid/index.js
+++ b/lib/rules/named-grid-areas-no-invalid/index.js
@@ -2,7 +2,6 @@
 
 'use strict';
 
-const _ = require('lodash');
 const declarationValueIndex = require('../../utils/declarationValueIndex');
 const findNotContiguousOrRectangular = require('./utils/findNotContiguousOrRectangular');
 const isRectangular = require('./utils/isRectangular');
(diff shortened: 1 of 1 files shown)


lib/rules/named-grid-areas-no-invalid/index.js
```diff
@@ -45,7 +44,13 @@ function rule(actual) {
 					return;
 				}
 
-				areas.push(_.compact(tokenValue.trim().split(' ')));
+				areas.push(
+					tokenValue
+						.trim()
+						.split(' ')
+						// keep only truthy values
+						.filter((t) => t),
```
ybiquitous, line 52: [suggestion] `split()` returns an array of strings, so how about clarifying the use of string argument as below?

```suggestion
						.filter((s) => s.length > 0),
```

ybiquitous (approved):
@stephenwade Thank you. I‘ve left a trivial comment about code readability, but this also looks good enough to me! 👍🏼

lib/rules/named-grid-areas-no-invalid/index.js
stephenwade: Nice, I like that better.

ybiquitous (approved):

PR #5360, opened 2021-06-07
Refactor to remove `_.trim`

stephenwade:

Ref #4412.

The regular expression `/^[ '"]+|[ '"]+$/g` matches one or more space, `'`, or `"` characters at the beginning or end of the string. Therefore, `string.replace(/^[ '"]+|[ '"]+$/g, '')` is equivalent to `_.trim(string, ' \'"')`.

`string.trim()` is a built-in function that removes whitespace, so `string.trim()` is equivalent to `_.trim(string, ' ')`.

### DIFF
--- a/lib/rules/function-url-no-scheme-relative/index.js
+++ b/lib/rules/function-url-no-scheme-relative/index.js
@@ -24,7 +24,7 @@ function rule(actual) {
 
 		root.walkDecls((decl) => {
 			functionArgumentsSearch(decl.toString().toLowerCase(), 'url', (args, index) => {
-				const url = args.replace(/^[ '"]+|[ '"]+$/g, '');
+				const url = args.trim().replace(/^['"]+|['"]+$/g, '');
 
 				if (!isStandardSyntaxUrl(url) || !url.startsWith('//')) {
 					return;
--- a/lib/rules/function-url-no-scheme-relative/index.js
+++ b/lib/rules/function-url-no-scheme-relative/index.js
@@ -2,7 +2,6 @@
 
 'use strict';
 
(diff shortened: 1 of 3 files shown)


lib/rules/function-url-no-scheme-relative/index.js
```diff
@@ -25,7 +24,7 @@ function rule(actual) {
 
 		root.walkDecls((decl) => {
 			functionArgumentsSearch(decl.toString().toLowerCase(), 'url', (args, index) => {
-				const url = _.trim(args, ' \'"');
+				const url = args.replace(/^[ '"]+|[ '"]+$/g, '');
```
ybiquitous, line 27: [question] Does it make sense to use a combination of the native `trim()` and a regex as below?

```suggestion
				const url = args.trim().replace(/^['"]+|['"]+$/g, '');
```

Because the native `trim()` removes not only `' '` but also other whitespace characters like `\t` or `\uFEFF`.
See the [polyfill on MDN](http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim#polyfill).

ybiquitous (commented):
@stephenwade Thank you for creating this change! 👏🏼 
I‘ve left just a question, so please confirm it.

lib/rules/function-url-no-scheme-relative/index.js
stephenwade: Yes, I think that makes sense. Also it matches with the behavior of the other code changed by this PR.

ybiquitous (approved):
Thanks! 👍🏼

PR #5362, opened 2021-06-08
Refactor to remove custom PostCSS type definitions

ybiquitous:

Follow-up of #5304

This refactoring removes the custom PostCSS type definition file: `types/postcss/index.d.ts`.
As a result, the internal use of `postcss/lib/comment` is no longer necessary.


lib/assignDisabledRanges.js
```diff
@@ -117,14 +117,20 @@ module.exports = function (root, result) {
 
 	return result;
 
+	/** @typedef {import('postcss').Comment & { raws: { inline?: boolean } }} PostcssSassComment */
```
ybiquitous, line 120: [note] Extracting the `PostcssSassComment` type is for readability of `isInlineComment()`.

jeddy3 (approved):
Thanks!

hudochenkov (approved):
Amazing!

PR #5365, opened 2021-06-08
Refactor to remove `_.kebabCase`

stephenwade:

Ref #4412.

No.

### DIFF
--- a/lib/utils/checkInvalidCLIOptions.js
+++ b/lib/utils/checkInvalidCLIOptions.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
 const chalk = require('chalk');
 const EOL = require('os').EOL;
 const levenshtein = require('fastest-levenshtein');
@@ -46,6 +45,22 @@ const suggest = (all, invalid) => {
 };
 
 /**
+ * Converts a string to kebab case.
+ * For example, `kebabCase('oneTwoThree') === 'one-two-three'`.
+ * @param {string} opt
+ * @returns {string}
+ */
+const kebabCase = (opt) => {
+	const matches = opt.match(/[A-Z]?[a-z]+|[A-Z]|[0-9]+/g);
+
+	if (matches) {
+		return matches.map((s) => s.toLowerCase()).join('-');
+	}
+
+	return '';
+};
+
+/**
  * @param {string} opt
  * @return {string}
  */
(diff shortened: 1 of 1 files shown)


mattxwang (commented):
Generally LGTM, though I'm curious if we should add a couple of test cases since we're now providing the function ourselves. Any thoughts?

(also, I'm surprised that this is our only use of `kebabCase`!)

lib/utils/checkInvalidCLIOptions.js
```diff
@@ -45,6 +44,22 @@ const suggest = (all, invalid) => {
...
+ */
+const kebabCase = (opt) => {
+	const matches = opt.match(/[A-Z]?[a-z]+|[A-Z]|[0-9]+/g);
+
+	if (matches) {
+		return matches.map((s) => s.toLowerCase()).join('-');
+	}
+
+	return '';
+};
```
ybiquitous, line 61: [suggestion] Can we use a simpler version according to the following `kebab-case` package?

https://github.com/joakimbeng/kebab-case/blob/21cba70cd5e694470cc01402b9df829bec1cada7/index.js#L5-L9

ybiquitous:
> Generally LGTM, though I'm curious if we should add a couple of test cases since we're now providing the function ourselves. Any thoughts?

I think the existing test cases cover, but if shortage, it would be good to add more cases to `checkInvalidCLIOptions.test.js`:

https://github.com/stylelint/stylelint/blob/403248c5586d6015127b5f98cf93442efc7c2547/lib/utils/__tests__/checkInvalidCLIOptions.test.js#L29-L39

lib/utils/checkInvalidCLIOptions.js
stephenwade: The linked code is only a little bit simpler, and I think my code is better at handling edge cases.

My function:
```
> kebabCase('abcDef')
'abc-def'
> kebabCase('AbcDef')
'abc-def'
> kebabCase('Option2')
'option-2'
> kebabCase('Abc123defG')
'abc-123-def-g'
```

The linked function:
```
> kebabCase('abcDef')
'abc-def'
> kebabCase('AbcDef')
'-abc-def'
> kebabCase('Option2')
'-option2'
> kebabCase('Abc123defG')
'-abc123def-g'
```

ybiquitous: I see 👍🏼

ybiquitous (approved):
Thanks! LGTM 👍🏼

PR #5366, opened 2021-06-08
Refactor to remove `_.isEqual`

stephenwade:

Ref #4412.

Since all uses of `_.isEqual` are comparing two arrays, I added a simple utility function to compare arrays.

### DIFF
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/index.js
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/index.js
@@ -2,7 +2,7 @@
 
 'use strict';
 
-const _ = require('lodash');
+const arrayEqual = require('../../utils/arrayEqual');
 const eachDeclarationBlock = require('../../utils/eachDeclarationBlock');
 const optionsMatches = require('../../utils/optionsMatches');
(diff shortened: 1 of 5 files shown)


ybiquitous (commented):
Thank you for creating this PR!
IMO, [`fast-deep-equal`](https://github.com/epoberezkin/fast-deep-equal) seems to be a bit heavy for us. Also, I‘m concerned that the package does not provide an ESM version.

So, I suggest adding a simple utility function as follows:

```js
/**
 * @template T
 * @param {T[]} a
 * @param {T[]} b
 * @returns {boolean}
 */
function arrayEquals(a, b) {
	return a.length === b.length && a.every((elem, index) => elem === b[index]);
}
```

The code above is based on the sample code on <https://masteringjs.io/tutorials/fundamentals/compare-arrays>.

stephenwade:
Hi @ybiquitous,

Good idea! I didn't notice that all the uses of `_.isEqual` are only used to compare arrays. I have updated the PR to add a utility function like you suggested.

ybiquitous (approved):
@stephenwade Thank you for the quick fix! LGTM 👍🏼

PR #5369, opened 2021-06-10
Refactor to remove `_.merge` and `_.mergeWith`

stephenwade:

Ref #4412.

`_.merge` replaces values at each index (`_.merge([1, 2], [3])` becomes `[3, 2]`). The `replaceSameIndex` array merging function in [`lib/augmentConfig.js`](https://github.com/stylelint/stylelint/pull/5369/files#diff-231823848e62de25699c7b9a8e92b05cc45aaf489c7c1fa417286295d37fcdf6) causes `deepmerge` to use that strategy.

`deepmerge` defaults to concatenating arrays, so the `mergeCustomizer` function used with `_.mergeWith` in [`lib/testUtils/mergeTestDescriptions.js`](https://github.com/stylelint/stylelint/pull/5369/files#diff-124f7773abf544ee51d0acd57be0e09ecf1636008b085830434668f0a1a4adb5) can simply be replaced with the default behavior of `deepmerge`.

### DIFF
--- a/lib/augmentConfig.js
+++ b/lib/augmentConfig.js
@@ -1,9 +1,9 @@
 'use strict';
 
-const _ = require('lodash');
 const configurationError = require('./utils/configurationError');
 const getModulePath = require('./utils/getModulePath');
 const globjoin = require('globjoin');
+const merge = require('deepmerge');
 const normalizeAllRuleSettings = require('./normalizeAllRuleSettings');
 const path = require('path');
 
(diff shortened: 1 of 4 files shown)


ybiquitous (commented):
[suggestion] How about defining the merge options out of `augmentConfigBasic()` as below?

- Avoid the redundant creation of the options object.
- Simplify the type annotation.
- Use `@ts-expect-error` for the insufficient option. See https://github.com/TehShrike/deepmerge/pull/231

```js
/** @type {import('deepmerge').Options} */
const MERGE_OPTIONS = {
	arrayMerge(target, source, options) {
		const destination = [...target];

		source.forEach((item, index) => {
			// @ts-expect-error -- The type definition is insufficient. See https://github.com/TehShrike/deepmerge/pull/231
			destination[index] = options.cloneUnlessOtherwiseSpecified(item, options);
		});

		return destination;
	}
};
```

ybiquitous:
I think the selection of [`deepmerge`](https://github.com/TehShrike/deepmerge) is good because:
- it‘s hard to implement such a function ourselves
- `deepmerge` has no dependencies
- `deepmerge` provides the type definitions

Unfortunately, `deepmerge` does not provide an ESM version (see https://github.com/TehShrike/deepmerge/issues/194), but ESM can import CJS modules, so we will have no problems.

stephenwade:
@ybiquitous Good idea, I like the simpler type definition, and it is easier to just ignore the error than write a huge JSDoc comment. Resolved in 9262c074.

ybiquitous (approved):
@stephenwade Thank you! LGTM 👍🏼

PR #5370, opened 2021-06-11
Refactor to remove `_.cloneDeep`

stephenwade:

Ref #4412.

No.

### DIFF
--- a/lib/needlessDisables.js
+++ b/lib/needlessDisables.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
 const optionsMatches = require('./utils/optionsMatches');
 const putIfAbsent = require('./utils/putIfAbsent');
 const validateDisableSettings = require('./validateDisableSettings');
@@ -22,7 +21,7 @@ module.exports = function (results) {
 		const [enabled, options, stylelintResult] = settings;
 
 		/** @type {{[ruleName: string]: Array<DisabledRange>}} */
-		const rangeData = _.cloneDeep(stylelintResult.disabledRanges);
+		const rangeData = stylelintResult.disabledRanges;
 
 		if (!rangeData) return;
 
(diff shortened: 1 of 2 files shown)


ybiquitous:
@stephenwade Thank you for creating this PR! The `deep-clone` package seems good to me. 👍🏼 

However, I doubt that `_.cloneDeep` is necessary in the first place. 🤔 

When I read the `needlessDisables()` function carefully, `_.cloneDeep` seems necessary to change the `rangeData` object:

https://github.com/stylelint/stylelint/blob/dadcf7b407ed5cb09cac70e943713f94b4fae3a0/lib/needlessDisables.js#L58-L68

I think we may be able to replace the logic above as below:

```diff
+		const allRangeComments = new Set(rangeData.all.map(range => range.comment));
+
 		for (const [rule, ranges] of Object.entries(rangeData)) {
 			for (const range of ranges) {
+				if (rule !== 'all' && allRangeComments.has(range.comment)) continue;
+
 				if (enabled === optionsMatches(options, 'except', rule)) continue;
 
 				const useful = usefulDisables.get(range.comment) || new Set();
```

What do you think?

If we could stop using `_.cloneDeep`, we could avoid the useless copy of big objects (`stylelint.DisabledRange`) including a `postcss.Comment` node.

stephenwade:
@ybiquitous I have modified the code like you suggested. I also found a way to use `deepmerge` to replace the other use of `_.cloneDeep` so I can avoid adding the dependency at all.

lib/needlessDisables.js
```diff
@@ -22,7 +21,7 @@ module.exports = function (results) {
 		const [enabled, options, stylelintResult] = settings;
 
 		/** @type {{[ruleName: string]: Array<DisabledRange>}} */
```
ybiquitous, line 23: [trivial] This `@type` annotation seems needless thanks to the type inference:

```suggestion
```

ybiquitous (approved):
@stephenwade Thank you for addressing my suggestion! I‘ve left a trivial comment, but this change looks good to me. 👍🏼

lib/needlessDisables.js
stephenwade: @ybiquitous You're right, I think that type must have been needed when we were deep copying, but it is not important anymore.

ybiquitous (approved):
@stephenwade Thanks! LGTM 👍🏼

PR #5371, opened 2021-06-11
Refactor to remove `_.escapeRegExp`

stephenwade:

Ref #4412.

No.

### DIFF
--- a/lib/formatters/stringFormatter.js
+++ b/lib/formatters/stringFormatter.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
 const chalk = require('chalk');
 const path = require('path');
 const stringWidth = require('string-width');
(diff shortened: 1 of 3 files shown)


ybiquitous:
@stephenwade Thank you for creating this PR!

The `escape-string-regexp` package seems good, but I think it better if we could avoid adding a new dependency.
So, how about replacing 

```js
.replace(new RegExp(_.escapeRegExp('(' + message.rule + ')') + '$'), '')
```

with the following tiny function not using regex?

```js
/**
 * @param {string} text
 * @param {string} rule
 * @returns {string}
 */
const removeRuleSuffix = (text, rule) => {
	if (text.endsWith(rule)) {
		return text.slice(0, text.lastIndexOf(rule));
	}
	return text;
}

// usage
removeRuleSuffix(message.text.replace(/\.$/, ''), `(${message.rule})`);
```

stephenwade:
@ybiquitous I have modified the code like you suggested to avoid using a regular expression.

lib/formatters/stringFormatter.js
```diff
@@ -147,6 +146,27 @@ function formatter(messages, source) {
...
+
+		result = result
+			// Remove all control characters (newline, tab and etc)
+			.replace(/[\u0001-\u001A]+/g, ' ') // eslint-disable-line no-control-regex
+			.replace(/\.$/, '');
+
+		const ruleString = `(${message.rule})`;
+
+		if (result.endsWith(ruleString)) {
+			result = result.slice(0, result.lastIndexOf(ruleString));
```
ybiquitous, line 164: Thank you! In addition, how about adding `.trim()` as below?

```suggestion
			result = result.slice(0, result.lastIndexOf(ruleString)).trim();
```

```js
// input
'Unexpected invalid hex color "#ggg" (color-no-invalid-hex)'
//=> without trim
'Unexpected invalid hex color "#ggg" '
//=> with trim
'Unexpected invalid hex color "#ggg"'
```

Maybe, we may need more test cases of `stringFormatter.js` to verify this behavior change. 🤔

stephenwade: @ybiquitous I added a space to the `ruleString` that we attempt to remove. I also added a test for the behavior.

```diff
-		const ruleString = `(${message.rule})`;
+		const ruleString = ` (${message.rule})`;
```

What do you think?

ybiquitous: It also looks good! 👍🏼

ybiquitous (approved):
@stephenwade Thank you so much! LGTM 👍🏼

PR #5374, opened 2021-06-14
Pre commit

### LINKED ISSUE #5373: Document pre-commit framework integration

I want to be able to use stylelint with [pre-commit framework](https://pre-commit.com). The biggest selling point of pre-commit is that in addition to managing git hooks it also installs them in isolated environments making configuration as simple as specifying them in a single yaml file. It is especially good for projects that use multiple languages as it supports hooks in [over a dozen languages](https://pre-commit.com/#supported-languages).

I saw that [docs](https://github.com/stylelint/stylelint/blob/403248c5586d6015127b5f98cf93442efc7c2547/docs/user-guide/integrations/other.md) currently mention this [repo](https://github.com/awebdeveloper/pre-commit-stylelint). However the solution has no benefits over just using [`repo: local`](https://pre-commit.com/#repository-local-hooks) and since it doesn't mirror stylelint releases it is actually worse than using local repo. One of the downsides of using solutions like that is that the hooks can't be updated using pre-commit's autoupdate command and need to be updated manually. Also I think that they make less use of pre-commit's cache if they are used across many repositories.

To be able to use this repo, pre-commit needs a little bit of metadata that needs to be put inside `.pre-commit-hooks.yaml` file. This is a simple config that shouldn't change much over time and therefore would require little to no maintenance. 

While writing this issue I came across a [comment](https://github.com/pre-commit/pre-commit/issues/1768#issuecomment-765551051) where the author of pre-commit suggests adding pre-commit support to this repo as a preferred solution for using stylelint with pre-commit.

AleksaC:

Closes #5373 

All should be clear from the issue.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+- id: stylelint
+  name: stylelint
+  entry: stylelint
+  language: node
+  types: [css]
+  require_serial: false
--- a/docs/user-guide/integrations/other.md
+++ b/docs/user-guide/integrations/other.md
@@ -11,7 +11,13 @@ Other integrations built and maintained by the community.
 
 ## Version Control
 
-- [Pre-commit](https://github.com/awebdeveloper/pre-commit-stylelint) - Git pre-commit hook for stylelint
+- [pre-commit](https://pre-commit.com/) - by adding this to `.pre-commit-config.yaml`:
+```yaml
+-   repo: https://github.com/stylelint/stylelint
+    rev: ''  # Use the sha / tag you want to point at
+    hooks:
+    -   id: stylelint
+```
 
 ## Command Line Tools
 


mattxwang (changes requested):
Is there any chance you could rebase and/or update your fork from the `v14` branch and then make the merge target the `v14` branch as well? The team is prepping for a major release version, so that would make the diff significantly cleaner + makes things easier once the final release is merged!

AleksaC:
I used `v14` in my fork, just used wrong base. Fixed it now. Thanks for pointing it out @mattxwang!

jeddy3:
@AleksaC Thanks for opening a pull request. I'm going to close for now to keep any discussion in one place.

(https://github.com/stylelint/stylelint/issues/5373#issuecomment-861280043)

PR #5378, opened 2021-06-18
Fix false positives for parametric mixins in selector-class-pattern

### LINKED ISSUE #5258: Fix false positives for parametric mixins in selector-class-pattern
> Clearly describe the bug

Getting `selector-class-pattern` violations on LESS mixin definitions.

`selector-class-pattern`

```less
.my-mixin(@x: 5) {
  line-height: @x;
}
```

```js
{
  "rules": {
    // kebab-case
    'selector-class-pattern': '^[a-z0-9]+(-?[a-z0-9]+)*$'
  }
}
```

`13.12.0`

CLI with `stylelint "src/**/*.less" --syntax less`"

Yes, it's related to [LESS parametric mixins](https://lesscss.org/features/#mixins-feature-mixins-parametric-feature).

The `selector-class-pattern` rule ignores LESS mixins [as documented](https://stylelint.io/user-guide/rules/selector-class-pattern)

Styelint seems not to recognize it as a mixin.

```shell
Expected class selector ".my-mixin(@x: 5)" to      selector-class-pattern
match pattern "^[a-z0-9]+(-?[a-z0-9]+)*$"
```

immitsu:
Fix #5258

Noticed that with the added if-block there is no necessity in checker on lines 59-61. Probably, should be removed?
Furthermore, if to delete checkers on lines 21-44 in `isStandardSyntaxRule.js` all tests will be passed. Is it ok?


jeddy3 (changes requested):
Thanks for the pull request, and apologies for just getting around to reviewing it.

Can you please set to merge into the `v14` branch.

> Noticed that with the added if-block there is no necessity in checker on lines 59-61. Probably, should be removed? Furthermore, if to delete checkers on lines 21-44 in isStandardSyntaxRule.js all tests will be passed. Is it ok?

If the new conditional makes the previous ones redundant and the tests pass, then yes go ahead a remove them.

ybiquitous:
Hi @immitsu, thank you for creating this PR!

When I’m looking into `isStandardSyntaxRule.test.js` in detail, I find some test cases that seem to be inaccurate.

For example, the following test case is close to this case added by this PR:

https://github.com/stylelint/stylelint/blob/403248c5586d6015127b5f98cf93442efc7c2547/lib/utils/__tests__/isStandardSyntaxRule.test.js#L41-L43

While `.mixin-name(@var);` is an **at-rule** node, `.mixin-name(@var) {}` is a **rule** node. Thus, I think the test case should be:

```diff
-expect(isStandardSyntaxRule(lessNode('.mixin-name(@var);'))).toBeFalsy();
+expect(isStandardSyntaxRule(lessNode('.mixin-name(@var) {}'))).toBeFalsy();
```

So, I think we should fix the test and refactor the tested code, and then fix this false positive. What do you think?

immitsu:
Hi @ybiquitous!

I think you're right, have fixed/deleted inaccurate test cases.
Thanks!

lib/utils/isStandardSyntaxRule.js
```diff
@@ -22,36 +22,13 @@ module.exports = function (rule) {
...
-
-	// Less detached rulesets
-	if (selector.startsWith('@') && selector.endsWith(':')) {
-		return false;
-	}
-
-	// Ignore Less &:extend rule
-	if ('extend' in rule && rule.extend) {
-		return false;
-	}
```
ybiquitous, line 39: [suggestion] This check of the `exptend` property seems effective still to me because `postcss-less` add the property. See below:

```console
$ node -pe 'require("postcss-less").parse("&:extend(.inline) {}").nodes[0]'
<ref *1> Rule {
  raws: { before: '', between: ' ', after: '' },
  type: 'rule',
  nodes: [],
  parent: Root {
    raws: { semicolon: false, after: '' },
    type: 'root',
    nodes: [ [Circular *1] ],
    source: { input: [Input], start: [Object] },
    [Symbol(isClean)]: false
  },
  source: {
    start: { offset: 0, line: 1, column: 1 },
    input: Input {
      css: '&:extend(.inline) {}',
      hasBOM: false,
      id: '<input css qtn4s3>',
      [Symbol(fromOffset cache)]: [Array]
    },
    end: { offset: 19, line: 1, column: 20 }
  },
  selector: '&:extend(.inline)',
  extend: true,
  [Symbol(isClean)]: false
}
```

So, how about moving this `extend` check before calling `isStandardSyntaxSelector()` as below?

```js
	if (rule.type !== 'rule') {
		return false;
	}

	// Ignore Less &:extend rule
	if ('extend' in rule && rule.extend) {
		return false;
	}
 ```

File: lib/utils/isStandardSyntaxRule.js [hunk 2]
```diff
@@ -22,36 +22,13 @@ module.exports = function (rule) {
...
 	if (selector.endsWith(')') && !selector.includes(':')) {
 		return false;
 	}
 
-	// Less guards
-	if (/when\s+(not\s+)*\(/.test(selector)) {
+	// Less Parametric mixins (e.g. .mixin(@variable: x) {})
+	if (/\(@.*\)$/.test(selector)) {
 		return false;
 	}
```
ybiquitous, line 33: [suggestion] Does it make sense to move all the checks for the `selector` variable into `isStandardSyntaxSelector()`? If possible, it seems that `isStandardSyntaxRule()` could be simpler like this:

```js
module.exports = function (rule) {
	// ...

	if (!isStandardSyntaxSelector(rule.selector)) {
		return false;
	}

	return true;
}
```

ybiquitous (commented):
@immitsu Thank you for addressing my reviews! I've left a few suggestions, so I would be glad if considering them. 😃

lib/utils/isStandardSyntaxRule.js
immitsu: Done.
Thank you!

lib/utils/isStandardSyntaxRule.js
immitsu: Done.
Thank you!

lib/utils/isStandardSyntaxSelector.js
```diff
@@ -29,6 +34,16 @@ module.exports = function (selector) {
...
+	// Less non-outputting mixin definition (e.g. .mixin() {})
+	if (selector.endsWith(')') && !selector.includes(':')) {
+		return false;
+	}
+
+	// Less Parametric mixins (e.g. .mixin(@variable: x) {})
+	if (/\(@.*\)$/.test(selector)) {
+		return false;
+	}
+
```
ybiquitous, line 46: @immitsu Thank you for addressing my reviews! Additionally, can you add test cases for the logic added to `isStandardSyntaxSelector()`, please?

- `SCSS nested properties`
- `Less non-outputting mixin definition (e.g. .mixin() {})`
- `Less Parametric mixins (e.g. .mixin(@variable: x) {})`

immitsu: @ybiquitous Done!

ybiquitous (approved):
@immitsu Thank you for your work! LGTM 👍🏼

jeddy3 (approved):
LGTM, thanks!

jeddy3:
v14 changelog entry:

> - Fixed: `selector-class-pattern` false positives for Less parametric mixins ([#5378](https://github.com/stylelint/stylelint/pull/5378)).

Issue #5205, opened 2021-03-21
Release 14.0.0


jeddy3:
Node 10 EOL is at the end of April. A lot of the packages we depend on will shift to Node 12 (and ESM) around then (see https://github.com/stylelint/stylelint/pull/5198 and https://github.com/sindresorhus/meta/discussions/15)

We should shoot to migrate stylelint to ESM for May. It'll be a major release of stylelint, so it's an opportunity to roll in other breaking changes. If you've got an itch to scratch that's a breaking change, then this is the time to do it!

We should remove deprecated code, e.g. https://github.com/stylelint/stylelint/issues/4899 and all the deprecated rules.

ybiquitous:
If we can use [dynamic imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports), `import-lazy` may be needless:

https://github.com/stylelint/stylelint/blob/e25b5a203deb63c642e0b3b4bbab8a320180725c/package.json#L128

jeddy3:
Does node@12 supports dynamic imports? If so, that'll be great as we are reliant lazy/dynamic loading for performance.

ybiquitous:
Oh, I missed that the `--experimental-modules` flag is necessary for Node 12, sorry 😓 

<img width="785" alt="image" src="https://user-images.githubusercontent.com/473530/114325435-21742d80-9b6b-11eb-8cac-76cdf7fd363c.png">

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#browser_compatibility

m-allanson:
I think the latest LTS release of node 12 supports ESM without a flag.  See:

- https://github.com/nodejs/node/pull/37718/files
- https://nodejs.org/docs/latest-v12.x/api/esm.html#esm_modules_ecmascript_modules

<img width="636" alt="Screenshot 2021-04-12 at 08 46 35" src="https://user-images.githubusercontent.com/381801/114359250-ab23fb00-9b6b-11eb-9cdc-7b3c50f62947.png">

hudochenkov:
@m-allanson I think we talk about `import().then()`, not regular `import`.

jeddy3:
> Oh, I missed that the --experimental-modules flag is necessary for Node 12, sorry

Perhaps ESM (non-dynamic) with bundling will be our approach for 14.0.0. It'll give us:

- something that's future-facing (for dynamic ESM down the line)
[...]

PR #5386, opened 2021-07-07
Rewrite paths for globby to use forward slash

### LINKED ISSUE #5122: "No files matching the pattern" when using Windows
> Clearly describe the bug

When using stylelint on Windows using a file path with a normal Windows path ex. `C:\dev\myfile.scss`, stylelint reports "No files matching the pattern". This is the same issue that was reported in https://github.com/stylelint/stylelint/issues/4772 but that issue was closed as not enough information was provided.

This _does_ work when forward slashes are used but this is not the standard file path used by Windows. This also affects other tools such as `@speedy/build-angular` which uses stylelint under the hood to pass a list of files to lint.

This is not related to any specific Rule. This is about running stylelint successfully on a Windows platform.

N/A but a simple file may be used such as:

```scss
.foo { color: black; }
```

N/A but the following can be used to match the example above.

```json
{
  "rules": {
    "color-named": "never"
  }
}
```

`13.9.0`

Using the CLI with the following command: `stylelint c:\dev\myfile.scss`

I am using SCSS but this is not a bug with the parsing of the file.

I expect the command to complete successfully when using standard Windows file path with backslashes.

The following error message was written to the console:

```shell
Error: No files matching the pattern "c:\dev\myfile.scss" were found.
    at c:\dev\node_modules\stylelint\lib\standalone.js:212:12
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

Notes:

To fix this, I think stylelint just needs to normalize the file paths for the files passed into it to a format that it expects.

tbowmo:
Fixes windows paths for globby, by replacing backward slash with forward slashes

Closes #5122

No, it's self-explanatory.


lib/standalone.js
```diff
@@ -175,11 +175,14 @@ module.exports = function (options) {
...
 		const absolutePath = !path.isAbsolute(entry) ? path.join(cwd, entry) : path.normalize(entry);
 
+		entry = entry.replace(/\\(?![[\]()])/g, '/');
+
 		if (fs.existsSync(absolutePath)) {
 			// This path points to a file. Return an escaped path to avoid globbing
 			return fastGlob.escapePath(entry);
 		}
 
+		// Flip backward slash to forward slash (windows compatability)
```
woutervanvliet, line 185: Runaway comment

lib/standalone.js
```diff
@@ -175,11 +175,14 @@ module.exports = function (options) {
 		const cwd = _.get(globbyOptions, 'cwd', process.cwd());
 		const absolutePath = !path.isAbsolute(entry) ? path.join(cwd, entry) : path.normalize(entry);
 
+		entry = entry.replace(/\\(?![[\]()])/g, '/');
```
woutervanvliet, line 178: Perhaps you can add a unit test on this behaviour, to make sure it doesn't break in the future

woutervanvliet (changes requested):

lib/standalone.js
tbowmo: Added unit test for windows style backslash in paths

lib/standalone.js
tbowmo: removed it..

woutervanvliet (approved):
LGTM

ybiquitous (commented):
@tbowmo Thanks for creating this PR. Could you consider the following suggestions, please?

- Can you change the base branch to [v14](https://github.com/stylelint/stylelint/tree/v14)?
  We have pushed many commits for the next version 14.0.0. (see #5205)
- According to the [FAQ on `fast-glob`](https://github.com/mrmlnc/fast-glob/blob/47fd82eceb97315138ef6158dbb5d4a8c0409039/README.md#how-to-write-patterns-on-windows), which is a base package of [`globby`](https://github.com/sindresorhus/globby), it seems better to me to use [`normalize-path`](https://github.com/jonschlinkert/normalize-path) because it is well-tested. What do you think?

    > 📖 Use the `normalize-path` or the `unixify` package to convert Windows-style path to a Unix-style path.

tbowmo:
@ybiquitous I'll move the changes to the v14 branch instead.. And also have a look at normalize-path.. It was a quick solution that was made in the first place (that is working..)

tbowmo:
@ybiquitous  I tried to use normalize-path, as suggested. (just replaced the regex that I added in standalone.js, with the following:
```
---		entry = entry = entry.replace(/\\(?![[\]()])/g, '/');
+++		entry = normalize(entry);
```
And added the required dependencies.. However, this breaks one of the tests:
_● standalone globbing › mixed globs and paths with special chars › manual escaping_
_No files matching the pattern "/home/thomas/gits/stylelint/lib/__tests__/fixtures/globs/got/[braces/] and /(spaces/)/*.+(s|c)ss" were found._

tbowmo:
btw. what is the timeframe for releasing v14? (I think @woutervanvliet is eager to get this fix into "production" as the current behavior is breaking a couple of developers machines, as they insist running windows only)

ybiquitous:
@tbowmo Sorry for the late reply. My answers are as below.

---

https://github.com/stylelint/stylelint/pull/5386#issuecomment-878273209

I think we should convert back-slashes (`\`) to forward-slashes (`/`) only when actual file paths are specified (not globs) because:
- `fast-glob` [doesn’t convert back-slashes](https://github.com/mrmlnc/fast-glob/blob/47fd82eceb97315138ef6158dbb5d4a8c0409039/README.md#how-to-write-patterns-on-windows)
- the back-slashes problem on globs is difficult and complex, according to the [`micromatch` notes](https://github.com/micromatch/micromatch/blob/5318752abc2f33153b3ccddf9f1f3b7682000a43/README.md#backslashes)
  (`fast-glob` uses internally `micromatch`)

So hers is my suggestion code, fixing also the test case:

```diff
diff --git a/lib/__tests__/standalone-globs.test.js b/lib/__tests__/standalone-globs.test.js
index 8254794d..ed28cd79 100644
--- a/lib/__tests__/standalone-globs.test.js
+++ b/lib/__tests__/standalone-globs.test.js
@@ -113,8 +113,11 @@ describe('standalone globbing', () => {
 	});
 
 	describe('windows path style with backslash', () => {
+		// The tests fail on non-windows environments.
+		if (process.platform !== 'win32') { return }
+
 		it('should handle normal backslash in glob pattern', async () => {
-			const cssGlob = `${fixturesPath}\\with spaces\\*.+(s|c)ss`;
+			const cssGlob = path.win32.join(fixturesPath, "globs", "with spaces", "styles.css");
 
 			const { results } = await standalone({
 				files: cssGlob,
diff --git a/lib/standalone.js b/lib/standalone.js
index 385d7b33..87b672b1 100644
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -19,6 +19,7 @@ const { default: ignore } = require('ignore');
 const DEFAULT_IGNORE_FILENAME = '.stylelintignore';
 const FILE_NOT_FOUND_ERROR_CODE = 'ENOENT';
 const ALWAYS_IGNORED_GLOBS = ['**/node_modules/**'];
+const normalizePath = require('normalize-path');
 const writeFileAtomic = require('write-file-atomic');
 
 /** @typedef {import('stylelint').StylelintStandaloneOptions} StylelintStandaloneOptions */
@@ -174,11 +175,9 @@ module.exports = function (options) {
 		const cwd = (globbyOptions && globbyOptions.cwd) || process.cwd();
 		const absolutePath = !path.isAbsolute(entry) ? path.join(cwd, entry) : path.normalize(entry);
 
-		entry = entry.replace(/\\(?![[\]()])/g, '/');
-
 		if (fs.existsSync(absolutePath)) {
 			// This path points to a file. Return an escaped path to avoid globbing
-			return fastGlob.escapePath(entry);
+			return fastGlob.escapePath(normalizePath(entry));
 		}
 
 		return entry;
```

- Run `normalizePath()` only when specified paths exists on a file system
- Run the test only for a windows environment

---

https://github.com/stylelint/stylelint/pull/5386#issuecomment-878275700

As you see the v14.0.0 issue, we have a lot of tasks to release v14.0.0. So I cannot promise the release date, unfortunately. 😞 

Perhaps, I think there are the following workarounds for this problem:
- Use your forked version based on this PR’s change, such as `"stylelint": "your-github-orb/stylelint#fix-win-path-problem"`
- Specify unix-like paths, such as `c:/dev/myfile.scss` instead of `c:\dev\myfile.scss`

tbowmo:
I'm not that big a fan of making conditions for when tests are run.. If the majority of devs are using unix like systems, then we never will run the tests for windows platform, unless you run the tests in the CI pipeline somehow, and then we can break the functionality for windows (like it is right now with the latest 13.x release)

I'll re-think testing for the windows paths when I get a bit more headroom (@woutervanvliet  you usually have a couple of hints as well ;))

tbowmo:
The above changes needs to be tested on a windows only system (no wsl), which I do not have access to at the moment (work pc is shutdown for 3 weeks of summer holidays)

tbowmo:
I fired up a win10 in virtualbox, for a quick spin of the tests, which unfortunately fails (on the windows path tests).. So I'll look into making a more permanent test setup on my tinkering machine, while debugging / testing this windows issue..

ybiquitous:
Yes, I also agree that we should avoid environment-specific tests if possible.

tbowmo:
Looked into it, I did the changes requested by @ybiquitous but that also fails when running the tests on windows! The extra test that I wrote, fails as the following bit of code:
```js
	fileList = fileList.map((entry) => {
		const cwd = (globbyOptions && globbyOptions.cwd) || process.cwd();
		const absolutePath = !path.isAbsolute(entry) ? path.join(cwd, entry) : path.normalize(entry);

		if (fs.existsSync(absolutePath)) {
			// This path points to a file. Return an escaped path to avoid globbing
			return fastGlob.escapePath(normalizePath(entry));
		}

		return entry;
	});
```
will not enter the if clause, as f.existsSync(absolutePath) returns false.

Thereby it doesn't rewrite the entry through normalizePath.

I then rewrote it, so I always throw entry through normalize path, like:
```js
	fileList = fileList.map((entry) => {
		const cwd = (globbyOptions && globbyOptions.cwd) || process.cwd();
		const absolutePath = !path.isAbsolute(entry) ? path.join(cwd, entry) : path.normalize(entry);

		entry = normalizePath(entry);

		if (fs.existsSync(absolutePath)) {
			// This path points to a file. Return an escaped path to avoid globbing
			return fastGlob.escapePath(entry);
		}

		return entry;
	});
```
but that also fails tests on windows ?!?!?

I then reverted to just use the regular expression as in my initial PR, and then it passes.. That test is passing on both windows and linux hosts.

ybiquitous:
@tbowmo Sorry, my suggestion code is bad. `globs` is needless. 🙇🏼 

```diff
-const cssGlob = path.win32.join(fixturesPath, 'globs', 'with spaces', 'styles.css');
+const cssGlob = path.win32.join(fixturesPath, 'with spaces', 'styles.css');
```

lib/__tests__/standalone-globs.test.js
```diff
@@ -112,6 +112,35 @@ describe('standalone globbing', () => {
 		);
 	});
 
+	describe('windows path style with backslash', () => {
+		// The tests fail on non-windows environments.
+		if (process.platform !== 'win32') {
+			return;
+		}
+
+		it('should handle normal backslash in glob pattern', async () => {
```
ybiquitous, line 121: The description seems inappropriate because this test no longer uses a glob pattern.
For example, how about fixing it as below?

```suggestion
		it('should handle a backslash-separated file path', async () => {
```

ybiquitous (commented):
Thank you for fixing the test failure (a65f668)!

ybiquitous (approved):
Thanks! LGTM 👍🏼

jeddy3 (approved):
Thank you!

jeddy3:
`v14` changelog entry added:

> - Fixed: "No files matching the pattern" when using backslash paths on Windows ([#5386](https://github.com/stylelint/stylelint/pull/5386)).

lib/standalone.js
```diff
@@ -174,11 +175,9 @@ module.exports = function (options) {
 		const cwd = (globbyOptions && globbyOptions.cwd) || process.cwd();
 		const absolutePath = !path.isAbsolute(entry) ? path.join(cwd, entry) : path.normalize(entry);
 
-		entry = entry.replace(/\\(?![[\]()])/g, '/');
-
 		if (fs.existsSync(absolutePath)) {
 			// This path points to a file. Return an escaped path to avoid globbing
-			return fastGlob.escapePath(entry);
+			return fastGlob.escapePath(normalizePath(entry));
```
VikasAgarwal1984, line 180: This only works when path is absolute, if applying a pattern like "**/*.scss" that is not absolute, it still fails.
I tried to run with 
entry = entry.replace(/\\(?])/g, '/');

it works, I have not read all comments what was the reason to remove it.

VikasAgarwal1984: entry = normalizePath(entry) needs to be before if (fs.existsSync(absolutePath)) {

tbowmo: @VikasAgarwal1984  this is an old (merged) pr, so if you think something needs changes, you should create a new pr for the changes.

PR #5389, opened 2021-07-16
Fix some type problems

ybiquitous:

A part of #4496

- Avoid `any` as possible
- Use `unknown`
- Use `@ts-expect-error`
  (see <https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments>)

### DIFF
--- a/lib/utils/validateTypes.js
+++ b/lib/utils/validateTypes.js
@@ -2,7 +2,7 @@
 
 /**
  * Checks if the value is a boolean or a Boolean object.
- * @param {any} value
+ * @param {unknown} value
  * @returns {boolean}
  */
 function isBoolean(value) {
@@ -11,7 +11,7 @@ function isBoolean(value) {
 
 /**
  * Checks if the value is a number or a Number object.
- * @param {any} value
+ * @param {unknown} value
  * @returns {boolean}
  */
 function isNumber(value) {
@@ -20,7 +20,7 @@ function isNumber(value) {
 
 /**
  * Checks if the value is a RegExp object.
- * @param {any} value
+ * @param {unknown} value
  * @returns {boolean}
  */
 function isRegExp(value) {
@@ -29,7 +29,7 @@ function isRegExp(value) {
 
 /**
  * Checks if the value is a string or a String object.
- * @param {any} value
+ * @param {unknown} value
  * @returns {boolean}
  */
 function isString(value) {
--- a/lib/utils/validateTypes.js
+++ b/lib/utils/validateTypes.js
@@ -3,7 +3,7 @@
 /**
  * Checks if the value is a boolean or a Boolean object.
  * @param {unknown} value
- * @returns {boolean}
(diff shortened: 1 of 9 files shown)


lib/utils/validateTypes.js
```diff
@@ -2,7 +2,7 @@
 
 /**
  * Checks if the value is a boolean or a Boolean object.
- * @param {any} value
+ * @param {unknown} value
  * @returns {boolean}
```
hudochenkov, line 6: Just in idea, don't know if it does make sense and helpful. Make all these functions into type guards:

```
 * @param {unknown} value
 * @returns {value is boolean}
```

Here's how they work as type guards: [Playground Link](https://www.typescriptlang.org/play?filetype=js#code/PQKhCgAIUhhALApgYwNYGdIEsBmkAuSkAbgIYA2ArotpqZAHaUC2ARogE6QD2X9Aci3ZdurAFYp8AOigwAAgAdSHUs0gBvSg1QNuAdwYBfEhWqzIcjonyUODTOrJUaWTEzadDs4OBxbk+FjcDLSCHhwAFE7UAJQa4ACQVjZ2BACeCojceNE0ALwFkADk7sJFkAA+FSbO2Pb4pAzIWXhhwgDc4F7goPL4GTTqjWnGID7Iwej4kAAmpA2QeZAAjAAsneAT9ZDISGiLoUKcEXMNMRu4kBGubcenpDFx6olb6NzkiFLk3ADmJ-MPTrdV7vT7fP73c5AA). Check `data` type in both `console.log`. Inside `if` `data` becomes known type. These functions could catch more issues, theoretically.

Here how it currently is: [Playground Link](https://www.typescriptlang.org/play?filetype=js#code/PQKhCgAIUhhALApgYwNYGdIEsBmkAuSkAbgIYA2ArotpqZAHaUC2ARogE6QD2X9Aci3ZdurAFYp8AOigwAAgAdSHUs0gBvSg1QNuAdwYBfEhWqzIcjonyUODTOtbdu5RKSOzg4HFuT4s3Ay0gmycABRkVIgAlBrgAJBWNnYEAJ4KiNx4kdSQALwFkADkTKEcRZAAPpUmUdj2+O7ImXghwgDc4Ibg4KDy+Ok06u6pxiBeyIHo+JAAJqSN+ZAAjAAsneCTDZDISGhLWOht4fON0Ru4kGGHxxxhp6TRseoJW+guiFLk3ADm9wuPTrdTZTD5fX7-M7tIA). `data` inside `if` is still `any`.

hudochenkov (approved):

lib/utils/validateTypes.js
ybiquitous: @hudochenkov Nice idea! I don't know we can use the feature with JSDoc! Changed 24a6f295. 👍🏼

jeddy3 (approved):
Thanks!

PR #5391, opened 2021-07-16
Refactor `validateOptions()` utility

ybiquitous:

Related to #4496

- Improve types avoiding `any`
- Use `for...of` instead of `forEach` (`continue` is clearer than `return`)
- Use `flat()` instead of `[].concat()` idiom (avoiding type errors)


hudochenkov (approved):

jeddy3 (approved):
Good improvements, thanks!

PR #5418, opened 2021-07-19
Improve types for `at-rule-*` rules

ybiquitous:

Related to #4496

This change removes `// @ts-nocheck` comments in `at-rule-*` rules.

### DIFF
--- a/lib/lintPostcssResult.js
+++ b/lib/lintPostcssResult.js
@@ -22,7 +22,7 @@ function lintPostcssResult(stylelintOptions, postcssResult, config) {
 	postcssResult.stylelint.quiet = config.quiet;
 	postcssResult.stylelint.config = config;
 
-	/** @type {string} */
+	/** @type {string | undefined} */
 	let newline;
 	const postcssDoc = postcssResult.root;
 
--- a/lib/rules/alpha-value-notation/index.js
+++ b/lib/rules/alpha-value-notation/index.js
@@ -1,5 +1,3 @@
-// @ts-nocheck
-
 'use strict';
 
 const valueParser = require('postcss-value-parser');
@@ -23,7 +21,8 @@ const messages = ruleMessages(ruleName, {
 const ALPHA_PROPS = new Set(['opacity', 'shape-image-threshold']);
 const ALPHA_FUNCS = new Set(['hsl', 'hsla', 'hwb', 'lab', 'lch', 'rgb', 'rgba']);
 
-function rule(primary, options, context) {
+/** @type {import('stylelint').StylelintRule} */
+const rule = (primary, secondaryOptions, context) => {
 	return (root, result) => {
 		const validOptions = validateOptions(
 			result,
@@ -33,7 +32,7 @@ function rule(primary, options, context) {
 				possible: ['number', 'percentage'],
 			},
 			{
-				actual: options,
+				actual: secondaryOptions,
 				possible: {
 					exceptProperties: [isString, isRegExp],
 				},
@@ -43,11 +42,23 @@ function rule(primary, options, context) {
 
 		if (!validOptions) return;
 
+		const optionFuncs = Object.freeze({
+			number: {
+				expFunc: isNumber,
+				fixFunc: asNumber,
+			},
+			percentage: {
+				expFunc: isPercentage,
+				fixFunc: asPercentage,
+			},
+		});
+
 		root.walkDecls((decl) => {
 			let needsFix = false;
 			const parsedValue = valueParser(getDeclarationValue(decl));
 
 			parsedValue.walk((node) => {
+				/** @type {import('postcss-value-parser').Node | undefined} */
 				let alpha;
 
 				if (ALPHA_PROPS.has(decl.prop.toLowerCase())) {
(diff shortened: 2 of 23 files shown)


lib/rules/alpha-value-notation/index.js
```diff
@@ -43,11 +42,23 @@ function rule(primary, options, context) {
...
+			number: {
+				expFunc: isNumber,
+				fixFunc: asNumber,
+			},
+			percentage: {
+				expFunc: isPercentage,
+				fixFunc: asPercentage,
+			},
+		});
+
```
ybiquitous, line 55: [note] This `optionFuncs` move aims to reduce needless object creation within the `walkDecls` callback.

lib/rules/alpha-value-notation/index.js
```diff
@@ -68,21 +79,15 @@ function rule(primary, options, context) {
...
 
-				if (optionsMatches(options, 'exceptProperties', decl.prop)) {
-					expectation = Object.keys(optionFuncs).filter((key) => key !== expectation);
+				if (optionsMatches(secondaryOptions, 'exceptProperties', decl.prop)) {
+					if (expectation === 'number') {
+						expectation = 'percentage';
+					} else if (expectation === 'percentage') {
+						expectation = 'number';
+					}
 				}
```
ybiquitous, line 91: [note] `Object.keys()` leads to more inaccurate type (`string[]`), so I've changed the logic.

lib/rules/alpha-value-notation/index.js
```diff
@@ -140,19 +168,27 @@ function findAlphaInFunction(node) {
 		return nodesAfterSlash.find(({ type }) => type === 'word');
 	}
 
-	return false;
+	return undefined;
```
ybiquitous, line 171: [note] I think it is a more common pattern to return `undefined` instead of `false`.

lib/rules/at-rule-allowed-list/index.js
```diff
@@ -15,28 +13,32 @@ const messages = ruleMessages(ruleName, {
 	rejected: (name) => `Unexpected at-rule "${name}"`,
 });
 
-function rule(listInput) {
+/** @type {import('stylelint').StylelintRule} */
+const rule = (primary) => {
 	// To allow for just a string as a parameter (not only arrays of strings)
-	const list = [].concat(listInput);
+	const primaryValues = [primary].flat();
```
ybiquitous, line 19: [note] The `[].concat()` idiom leads to a poor type inference.

lib/rules/at-rule-empty-line-before/index.js
```diff
@@ -133,7 +135,7 @@ function rule(expectation, options, context) {
 			}
 
 			// Fix
-			if (context.fix) {
+			if (context.fix && context.newline) {
```
ybiquitous, line 138: [note] `context.newline` can be `undefined`, so type-check raises an error without this addition.

lib/rules/at-rule-name-space-after/index.js
```diff
@@ -33,12 +32,14 @@ function rule(expectation, options, context) {
 			checkedRuleName: ruleName,
 			fix: context.fix
 				? (atRule) => {
-						atRule.raws.afterName = atRule.raws.afterName.replace(/^\s*/, ' ');
+						if (typeof atRule.raws.afterName === 'string') {
+							atRule.raws.afterName = atRule.raws.afterName.replace(/^\s*/, ' ');
+						}
```
ybiquitous, line 37: [note] `atRule.raws.afterName` can be `undefined`.

lib/utils/ruleMessages.js
```diff
@@ -1,28 +1,30 @@
...
  * that provides the same messages postfixed with the rule
  * that has been violated.
  *
+ * @template {import('stylelint').StylelintRuleMessages} T
  * @param {string} ruleName
- * @param {Messages} messages - Object whose keys are message identifiers
+ * @param {T} messages - Object whose keys are message identifiers
  *   and values are either message strings or functions that return message strings
- * @returns {Messages} New message object, whose messages will be marked with the rule name
+ * @returns {T} New message object, whose messages will be marked with the rule name
```
ybiquitous, line 12: [note] Returning `T` instead of `Messages` is more useful for clients of this `ruleMessages()` function. For example:

```ts
const messages = ruleMessages('some-rule', { expected: "Some rule ..." })
messages.expected //=> No error 👍
```

But using `@ts-expect-error` is ugly.  Is there a better way? 🤔

jeddy3 (approved):
Thank you! It's fantastic to see types added to some of the rules!

(I've been offline for a week and only just getting around to reviewing this).

ybiquitous:
@jeddy3 No problems. Thank you for the review and merge! 😄

lib/lintPostcssResult.js
```diff
@@ -22,7 +22,7 @@ function lintPostcssResult(stylelintOptions, postcssResult, config) {
 	postcssResult.stylelint.quiet = config.quiet;
 	postcssResult.stylelint.config = config;
 
-	/** @type {string} */
+	/** @type {string | undefined} */
```
Mouvedia, line 25: @ybiquitous `postcssDoc` at l29 is always truthy correct?
I am trying to check if the `newline == null` conditions are necessary.
Isn't `newline` always a string?

ybiquitous: Um, I don't remember it well. In the current PostCSS type definitions, the `Result.root` property always seems truthy:
- https://github.com/postcss/postcss/blob/8.4.38/lib/result.d.ts#L144
- https://github.com/postcss/postcss/blob/8.4.38/lib/lazy-result.js#L145

But I'm unsure, honestly. If so, we can make this code much simpler, including `newline`... 🤔

Mouvedia: I think it had consequences elsewhere. Anyway, I am gonna change its type to be `string` in an upcoming refactor.

PR #5426, opened 2021-07-27
Refactor to improve types for `block-*` rules

ybiquitous:

Related to #4496
(see also #5418)

This change removes `// @ts-nocheck` from the `block-*` rules.
In addition, it improves the types of some utilities.

### DIFF
--- a/lib/utils/blockString.js
+++ b/lib/utils/blockString.js
@@ -10,15 +10,14 @@ const rawNodeString = require('./rawNodeString');
 /**
  * Return a CSS statement's block -- the string that starts and `{` and ends with `}`.
  *
- * If the statement has no block (e.g. `@import url(foo.css);`),
- * return false.
+ * If the statement has no block (e.g. `@import url(foo.css);`), returns an empty string.
  *
  * @param {Rule | AtRule} statement - postcss rule or at-rule node
- * @return {string | boolean}
+ * @returns {string}
  */
-module.exports = function (statement) {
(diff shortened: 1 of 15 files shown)


lib/utils/blockString.js
```diff
@@ -10,15 +10,14 @@ const rawNodeString = require('./rawNodeString');
...
  *
  * @param {Rule | AtRule} statement - postcss rule or at-rule node
- * @return {string | boolean}
+ * @returns {string}
  */
-module.exports = function (statement) {
+module.exports = function blockString(statement) {
 	if (!hasBlock(statement)) {
-		return false;
+		return '';
```
ybiquitous, line 20: [note] I don't know why this function has returned a boolean, but it seems to make sense that it returns a string instead.

For example, it appears returning a boolean is incorrect at the following place, but errors never occur.
Perhaps, it has no problems because of calling `hasBlock()` before `blockString()`.

https://github.com/stylelint/stylelint/blob/403248c5586d6015127b5f98cf93442efc7c2547/lib/rules/block-opening-brace-space-after/index.js#L51

https://github.com/stylelint/stylelint/blob/403248c5586d6015127b5f98cf93442efc7c2547/lib/rules/block-opening-brace-space-after/index.js#L56

jeddy3 (approved):
It's great to see the quality of the code base improving with these refactors!

ybiquitous:
@jeddy3 Thank you so much for the quick review! 😄

PR #5475, opened 2021-08-12
Update non-ESM deps

jeddy3:

None.

Updates the remaining non-ESM deps.


mattxwang (approved):
LGTM! The only semver breaking change is `lint-staged`, and that's moving the package to Node 12 min - not a problem for v14!

PR #5485, opened 2021-08-18
Refactor to improve types for `declaration-*` rules

ybiquitous:

Part of #4496

This change removes  `// @ts-nocheck` from the remaining `declaration-*` rules.
Also, this adds 2 type parameters to the `StylelintRule` type to pass the type-check.

### DIFF
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -146,9 +146,9 @@ declare module 'stylelint' {
 
 	export type StylelintRuleMessages = Record<string, string | ((...args: any[]) => string)>;
 
-	export type StylelintRule = ((
-		primaryOption: any,
-		secondaryOptions: Record<string, any>,
+	export type StylelintRule<P = any, S = any> = ((
+		primaryOption: P,
+		secondaryOptions: Record<string, S>,
 		context: StylelintPluginContext,
 	) => (root: Root, result: PostcssResult) => Promise<void> | void) & {
 		ruleName: string;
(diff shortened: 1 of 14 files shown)


types/stylelint/index.d.ts
```diff
@@ -146,9 +146,9 @@ declare module 'stylelint' {
 
 	export type StylelintRuleMessages = Record<string, string | ((...args: any[]) => string)>;
 
-	export type StylelintRule = ((
-		primaryOption: any,
-		secondaryOptions: Record<string, any>,
+	export type StylelintRule<P = any, S = any> = ((
+		primaryOption: P,
+		secondaryOptions: Record<string, S>,
```
ybiquitous, line 151: [note] **P** - **P**rimary, **S** - **S**econdary

alexander-akait: Why do not use full words? :smile:

ybiquitous: I follow the existing style, but is using full words common? (I'm open to change)

https://github.com/stylelint/stylelint/blob/dddf8568cfdb2ec65f49164c1a5b67379d8756a8/types/stylelint/index.d.ts#L12

https://github.com/stylelint/stylelint/blob/dddf8568cfdb2ec65f49164c1a5b67379d8756a8/types/stylelint/index.d.ts#L51

ybiquitous, line 151: Another solution:

```suggestion
	export type StylelintRule<Primary = any, Secondary = any> = ((
		primaryOption: Primary,
		secondaryOptions: Record<string, Secondary>,
```

jeddy3: Let's follow the existing style. We can revisit the lot if it becomes an issue.

jeddy3 (approved):
Thanks as always, LGTM!

PR #5487, opened 2021-08-19
Refactor to improve types for `function-*` rules

ybiquitous:

Part of #4496

This change removes `// @ts-nocheck` from the `function-*` rules.


lib/rules/function-url-quotes/index.js
```diff
@@ -86,16 +97,23 @@ function rule(expectation, options) {
 					return;
 				}
 
+				// @ts-expect-error -- Should the message include the function name?
 				complain(messages.expected(functionName), node, complaintIndex);
```
ybiquitous, line 101: [question] The passed `functionName` is ignored now:

https://github.com/stylelint/stylelint/blob/544d85763f2a36ac83aa5d924e41713b1177d3bf/lib/rules/function-url-quotes/index.js#L15-L18

If we don't need to change the messages, I'll remove the `functionName` variable.

On the other hand, if we would improve the message like `Expected quotes for "url" function`, I'll change the message and test with another PR.

What do you think?

jeddy3 (approved):
Amazing. You're making a lot of progress through the rules!

lib/rules/function-url-quotes/index.js
jeddy3: Let's improve the message to `Expected quotes around "${functionName}" function argument` to be consistent with the messages in the other `*-quote` rules ([spec ref for "argument"](https://drafts.csswg.org/css-values/#functional-notation)).

---

Incidentally, it was interesting to [see that a `src()` function may be coming to the language](https://drafts.csswg.org/css-values/#funcdef-src) so that `var()` can be used.

ybiquitous: Thanks for the advice. I'll create a PR to improve the message! 👍🏼 

> Incidentally, it was interesting to [see that a `src()` function may be coming to the language](https://drafts.csswg.org/css-values/#funcdef-src) so that `var()` can be used.

Oh, I also didn't know the feature. It would be more useful! 😄

ybiquitous: I've opened PR #5488.

jeddy3: Thanks. I've merged it. Feel free to rebase this pull request, remove the `@ts-expect-error` and merge.

ybiquitous: Thanks! I've removed `@ts-expect-error` comments via 5bbae44b. When the CI passes, I'll merge it.

Issue #5433, opened 2021-08-06
Add rule-selector-property-disallowed-list


keganlance:
To further limit the freedom, it would be very welcome to disallow specific properties on certain selectors.

For example, not allowing the background-color property to be set on the selector "html".

Basically the same as property-disallowed-list, but mapped on selector. So you end up with a config like this:

```
'selector-property-disallowed-list': {
  'html': ['background-color']
}
```

jeddy3:
@keganlance Thanks for the request and for providing an example.

Sounds good to me. The `*-disallowed-list` rules are how teams enforce these types of constraints.

We've prior art for selectors in the [`selector-disallowed-list` rule](https://stylelint.io/user-guide/rules/list/selector-disallowed-list/). Like that rule, this rule should apply to the whole _selector list_.

The rule should be scoped to a _rule_ (aka _rule-set_), like the [`declaration-property-value-disallowed-list` rule](https://stylelint.io/user-guide/rules/list/declaration-property-value-disallowed-list/) is scoped to a _declaration_.

- Name: `rule-selector-property-disallowed-list`
- Primary option: `object`
- Secondary options: None
- Autofixable: No
- Message: `Unexpected property "${property}" for selector "${selector}"`
- Description: "Specify a list of disallowed properties for selectors within rules."
- Section: "Limit language features" -> "Rules"

I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

There are [steps on how to add a new rule](https://stylelint.io/developer-guide/rules#add-a-rule) in the Developer guide.

You can look at the two rules I mentioned above, and the [`at-rule-property-required-list`](https://github.com/stylelint/stylelint/tree/v14/lib/rules/at-rule-property-required-list) one, for a blueprint.

Branch off `v14` and not `master`.

aNickzz:
I'd like to request an extension to this slightly, I'd like to disallow certain properties in non-nested selectors.

```scss
.foo {
  width: 35px; //error
}

.bar .foo {
  width: 35px; //ok
}
```

My motivation for this is that there are particular properties that really depend on context. For example the `width` a button depends on its layout, as does the `margin` around a `.heading-1`. Margins especially, since they directly relate to the spacing between components rather than the component itself. The layout depends on context.

That said, I'm inclined to think that this should be valid too

```scss
.foo__bar {
  width: 100%;
}
```

[...]

PR #5489, opened 2021-08-22
Add new rule-selector-property-disallowed-list rule

doing-art:

The new rule rule-selector-property-disallowed-list that is mentioned in #5433 has been added.

### DIFF
--- a/lib/rules/rule-selector-property-disallowed-list/index.js
+++ b/lib/rules/rule-selector-property-disallowed-list/index.js
@@ -1,0 +1,69 @@
+// @ts-nocheck
+
+'use strict';
+
+const isStandardSyntaxRule = require('../../utils/isStandardSyntaxRule');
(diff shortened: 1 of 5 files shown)


lib/rules/rule-selector-property-disallowed-list/index.js
```diff
@@ -0,0 +1,69 @@
+// @ts-nocheck
```
ybiquitous, line 1: [suggestion] Can you remove `// @ts-nocheck` and pass the type-check, please?

Please refer to the following code:

https://github.com/stylelint/stylelint/blob/c0b1773f706614ed6acda8fa35264f918b81188c/lib/rules/alpha-value-notation/index.js#L24-L25

See also #4496

ybiquitous (commented):
Thank you for creating the PR! 👏🏼

PR #5494, opened 2021-08-23
Refactor to improve types for `hue-*` rules

ybiquitous:

Part of #4496

This change removes `// @ts-nocheck` from the `hue-*` rules.

### DIFF
--- a/lib/rules/hue-degree-notation/index.js
+++ b/lib/rules/hue-degree-notation/index.js
@@ -1,5 +1,3 @@
-// @ts-nocheck
-
 'use strict';
 
 const valueParser = require('postcss-value-parser');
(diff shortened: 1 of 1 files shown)


lib/rules/hue-degree-notation/index.js
```diff
@@ -78,18 +77,29 @@ function rule(primary, secondary, context) {
...
 function asDegree(value) {
 	return `${value}deg`;
 }
 
+/**
+ * @param {string} value
+ */
 function asNumber(value) {
-	const { number } = valueParser.unit(value);
+	const dimension = valueParser.unit(value);
```
ybiquitous, line 93: [note] `valueParser.unit()` returns `valueParser.Dimension` or `false`. See below:

https://github.com/TrySound/postcss-value-parser/blob/v4.1.0/lib/index.d.ts#L134

lib/rules/hue-degree-notation/index.js
```diff
@@ -102,19 +112,25 @@ function findHue(node) {
 		return args[2];
 	}
 
-	return false;
+	return undefined;
```
ybiquitous, line 115: [note] The `findHue()` function now returns `valueParser.Node` or `undefined`. I think it's a more common pattern to return `undefined` instead of `false`.

jeddy3 (approved):
LGTM, thanks!

PR #5498, opened 2021-08-25
Refactor to improve types for some rules

ybiquitous:

Part of #4496

This change removes `// @ts-nocheck` from the following rules:
- `keyframe-declaration-no-important`
- `keyframes-name-pattern`
- `length-zero-no-unit`
- `linebreaks`

### DIFF
--- a/lib/rules/linebreaks/index.js
+++ b/lib/rules/linebreaks/index.js
@@ -1,5 +1,3 @@
-// @ts-nocheck
-
 'use strict';
 
 const postcss = require('postcss');
@@ -13,10 +11,11 @@ const messages = ruleMessages(ruleName, {
 	expected: (linebreak) => `Expected linebreak to be ${linebreak}`,
 });
 
-function rule(actual, secondary, context) {
+/** @type {import('stylelint').StylelintRule} */
+const rule = (primary, _secondaryOptions, context) => {
 	return (root, result) => {
 		const validOptions = validateOptions(result, ruleName, {
-			actual,
+			actual: primary,
 			possible: ['unix', 'windows'],
 		});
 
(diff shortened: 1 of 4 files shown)


lib/rules/linebreaks/index.js
```diff
@@ -81,23 +99,28 @@ function rule(actual, secondary, context) {
...
+		 * @param {number} line
+		 * @param {number} column
+		 */
 		function reportNewlineError(line, column) {
 			// Creating a node manually helps us to point to empty lines.
 			const node = postcss.rule({
 				source: {
-					start: { line, column },
+					start: { line, column, offset: 0 },
+					input: new postcss.Input(''),
```
ybiquitous, line 111: [note] 

- [`Position`](https://postcss.org/api/#position) requires `offset`
- [`Source`](https://postcss.org/api/#source) requires `input`

...but, I'm not sure if the added dummy values are appropriate. 🤔

lib/rules/linebreaks/index.js
```diff
@@ -13,35 +11,49 @@ const messages = ruleMessages(ruleName, {
...
 		if (!validOptions) {
 			return;
 		}
 
-		const shouldHaveCR = actual === 'windows';
+		const shouldHaveCR = primary === 'windows';
 
 		if (context.fix) {
-			const propertiesToUpdate = ['selector', 'value', 'text'];
-			const rawsPropertiesToUpdate = ['before', 'after'];
```
ybiquitous, line 31: [note] This is a more type-safe way, so that the property sets are unnecessary.

lib/rules/linebreaks/index.js
jeddy3: I think it's acceptable. We can always revisit.

jeddy3 (approved):
Fantastic, thanks!

The `@param` above functions adds so much clarity, especially when destructuring is being used, e.g.:

```js
/**
 * @param {import('postcss-value-parser').Node[]} nodes
 * @param {number} index
 */
function isLineHeightValue({ prop }, nodes, index) { .. }
```

ybiquitous:
@jeddy3 I always appreciate your review! The progress is almost 50%: 😃 

```console
$ ls -l lib/rules | wc -l
     201

$ git grep -l '@ts-nocheck' -- lib | wc -l
     109
```

PR #5454, opened 2021-08-12
Refactor to improve types for `comment-*` rules

ybiquitous:
> Which issue, if any, is this issue related to?

Related to #4496
(see also #5426)

> Is there anything in the PR that needs further explanation?

This change removes `// @ts-nocheck` from the `comment-*` rules.
Also, this change extracts `lib/utils/isInlineComment.js` from `lib/assignDisabledRanges.js`.

PR #5500, opened 2021-08-26
Fix false positives in nested property declarations unit-no-unknown

### LINKED ISSUE #5434: Fix false positives for nested property declaration that contains `tx` or `qx` in unit-no-unknown
> Clearly describe the bug

There are false positives for  unknown unite when you use `tx` or `qx`

`unit-no-unknown`

e.g.

```scss
// Font Sizes
$font-sizes: (
    2x: $font-size-base * 2,
    2qx: $font-size-base * 2.25,
    2hx: $font-size-base * 2.5,
    tx: $font-size-base * 2.75,
    3x: $font-size-base * 3,
    3qx: $font-size-base * 3.25,
    3hx: $font-size-base * 3.5,
    3tx: $font-size-base * 3.75,
    4x: $font-size-base * 4,
    4qx: $font-size-base * 4.25,
    4hx: $font-size-base * 4.5,
    4tx: $font-size-base * 4.75,
    5x: $font-size-base * 5,
    5qx: $font-size-base * 5.25,
    5hx: $font-size-base * 5.5,
    5tx: $font-size-base * 5.75
)
```

e.g.

```json
{
  "rules": {
    "unit-no-unknown": "true"
  }
}
```

e.g. `13.13.1`

```shell
yarn run stylelint --config .github/linters/.stylelintrc.json "app/**/*.scss"  
```

Yes, it's related to SCSS nested properties declaration.

No warnings to be flagged.

```shell
 428:5  ✖  Unexpected unknown unit "qx"   unit-no-unknown
 432:5  ✖  Unexpected unknown unit "qx"   unit-no-unknown
 434:5  ✖  Unexpected unknown unit "tx"   unit-no-unknown
 436:5  ✖  Unexpected unknown unit "qx"   unit-no-unknown
 438:5  ✖  Unexpected unknown unit "tx"   unit-no-unknown
 440:2  ✖  Unexpected unknown unit "qx"   unit-no-unknown
 442:5  ✖  Unexpected unknown unit "tx"   unit-no-unknown
```

sylwia-lask:

Closes #5434 

The problem was the assumption that property in map always has index divided by 4. It isn't true when there is multiplier ("*") in the value. I changed the condition and now it works fine.


lib/rules/unit-no-unknown/index.js
```diff
@@ -22,10 +22,6 @@ const messages = ruleMessages(ruleName, {
 	rejected: (unit) => `Unexpected unknown unit "${unit}"`,
 });
 
-// The map property name (in map cleared from comments and spaces) always
-// has index that being divided by 4 gives remainder equals 0
-const mapPropertyNameIndexOffset = 4;
```
ybiquitous, line 27: [question] This `mapPropertyNameIndexOffset` constant (and also [`isMap`](https://github.com/stylelint/stylelint/blob/5d98cc3e65f5fc1551dc2d1166efb481e3bba679/lib/utils/isMap.js)) was introduced with PR #4450. I'm not sure the context, but it seems better to me to improve `isStandardSyntaxDeclaration` as @jeddy3 suggested on <https://github.com/stylelint/stylelint/issues/5434#issuecomment-894419800>, instead of changing the current logic.

What do you think?

ybiquitous (commented):
@sylwia-lask Thanks for the contribution! This change sounds good enough, but I've left a question.

lib/rules/unit-no-unknown/index.js
sylwia-lask: @ybiquitous this PR doesn't change current logic. It just changes the assumption, that the mapPropertyNameIndexOffset is always 4 - because it isn't, when map value has "*" inside. I've changed it, that now we check that property ends with ":", which is much more reliable in my opinion. 

If we'll use isStandardSyntaxDeclaration it will disable this rule for all scss maps. Which I guess we don't want to do, because now it works ok, instead of the edge case with "*" in value. And for sure we can't make this kind of logic in 119 line, since upper "if" is `if (unit.toLowerCase() === 'x') ` (what with the other units?)

What do you think @jeddy3, am I thinking correctly?

jeddy3: @sylwia-lask Thanks for your pull request. 

Let's remove the map functionality, including `isMap`, as it's inconsistent with how all the other built-in rules deal with non-standard syntax, i.e. by ignoring them using the `isStandardSyntax*` utils. Code for non-standard syntax only belongs in the `isStandardSyntax*` utils.

The built-in rules are for standard CSS constructs, and so this rule should ignore maps.

Someone can then write a `scss/unit-no-unknown` plugin for [`stylelint-scss`](https://github.com/kristerkari/stylelint-scss) that specifically checks for unknown units in Sass constructs like maps. The plugin can deal with the awkward edge cases that come from working with SCSS syntax using parsers designed for CSS). The plugin would be used in tandem with the built-in rule.

Ref: https://github.com/stylelint/stylelint/pull/5454#pullrequestreview-728372653

sylwia-lask: @jeddy3 in test file for unit-no-unknown are lots of tests for less, sass and scss syntaxes. Should I remove it?

jeddy3: Let's just remove the `reject` tests that will fail when we ignore the constructs outright. The `accept` tests will remain valid as they'll now test that the constructs are ignored.

When you remove the map code and add the (amended) `isStandardSyntaxDeclaration` util, it should become clear which tests these are.

lib/utils/isStandardSyntaxDeclaration.js
```diff
@@ -36,6 +37,11 @@ module.exports = function (decl) {
 		return false;
 	}
 
+	// SCSS map and list declarations
+	if (value[0] === '(' && value[value.length - 1] === ')') {
+		return false;
+	}
+
```
ybiquitous, line 44: @sylwia-lask Thanks for the change! In addition, can you add a test case to `lib/utils/__tests__/isStandardSyntaxDeclaration.test.js`, please?

lib/rules/unit-no-unknown/index.js
```diff
@@ -47,11 +43,14 @@ function rule(actual, options) {
 		}
 
 		function check(node, value, getIndex) {
+			if (node.type === 'decl' && !isStandardSyntaxDeclaration(node)) {
```
jeddy3, line 46: Let's move this into the `walkDecls` at the bottom of the file, e.g.

```js
    if (!/^media$/i.test(atRule.name) && !atRule.variable) {
        return;
    }
    check(atRule, atRule.params, atRuleParamIndex);
});
root.walkDecls((decl) => {
    if (!isStandardSyntaxDeclaration(decl) {
        return;
    }
    check(decl, decl.value, declarationValueIndex));
}
```

So that it's grouped with the at-rule conditional.

---

Note: we should replace, in a separate refactor pull request, the `&& !atRule.variable` with the `isStandardSyntaxAtRule` util.

lib/utils/isStandardSyntaxDeclaration.js
```diff
@@ -36,6 +37,11 @@ module.exports = function (decl) {
 		return false;
 	}
 
+	// SCSS map and list declarations
+	if (value[0] === '(' && value[value.length - 1] === ')') {
```
jeddy3, line 41: Let's use string functions for this, which we've moved to in the other `isStandardSyntax*` utils:

```suggestion
	if (value.startsWith('(') && value.endsWith(')')) {
```

jeddy3 (changes requested):
@sylwia-lask Thanks!

It's looking good. I've requested two changes in addition to @ybiquitous request.

lib/rules/unit-no-unknown/index.js
sylwia-lask: Thanks for review :) I can change the `&& !atRule.variable` in separate pull request, do I need to create an issue for that?

ybiquitous (approved):
@sylwia-lask Thanks! LGTM 👍🏼

jeddy3 (approved):
Thanks for making the changes. LGTM!

lib/rules/unit-no-unknown/index.js
jeddy3: > I can change the `&& !atRule.variable` in separate pull request, do I need to create an issue for that?

Fantastic! No need to create a new issue. Just link to this thread in the pull request note.

jeddy3:
`v14` changelog entry added:

- Fixed: `unit-no-unknown` false positives for nested property declarations ([#5500](https://github.com/stylelint/stylelint/pull/5500)).

PR #5507, opened 2021-09-02
Fix max-empty-lines before comments

yepninja:
> Which issue, if any, is this issue related to?

https://github.com/stylelint/stylelint/issues/4330

> Is there anything in the PR that needs further explanation?

e.g. "No, it's self-explanatory."

PR #5508, opened 2021-09-02
Refactor to improve types for `max-*` rules

ybiquitous:

Part of #4496

This change removes `// @ts-nocheck` from the `max-*` rules.

### DIFF
--- a/lib/rules/max-empty-lines/index.js
+++ b/lib/rules/max-empty-lines/index.js
@@ -1,7 +1,6 @@
-// @ts-nocheck
-
 'use strict';
 
+const isStandardSyntaxComment = require('../../utils/isStandardSyntaxComment');
 const optionsMatches = require('../../utils/optionsMatches');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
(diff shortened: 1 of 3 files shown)


ybiquitous:
### To-do

- [ ] Wait for merging #5507

lib/rules/max-empty-lines/index.js
```diff
@@ -67,18 +64,19 @@ function rule(max, options, context) {
 			const rootRawsAfter = root.raws.after;
 
 			// not document node
+			// @ts-expect-error -- TS2339: Property 'document' does not exist on type 'Root'.
```
jeddy3, line 67: We'll likely need to revisit this `document` stuff at some point in light of https://github.com/postcss/postcss/pull/1582.

I think it's mainly the max whitespace rules that touch it.

jeddy3 (approved):
Looks great, thanks!

lib/rules/max-empty-lines/index.js
ybiquitous: Yes, I also think so. 👍🏼 
I didn't know how I should handle it, so I've left `// @ts-expect-error` comments.

jeddy3: Leaving `// @ts-expect-error` comments was the right thing to do.

Feel free to merge.

ybiquitous: Thank you!

PR #4581, opened 2020-02-03
Add range context support for media-feature-name-* rules

jdiogopeixoto:
> Which issue, if any, is this issue related to?

Closes #1744 

> Is there anything in the PR that needs further explanation?

No, it's self explanatory.

PR #5510, opened 2021-09-02
Refactor to improve types for `media-feature-name-*` rules

ybiquitous:

Part of #4496

This change removes `// @ts-nocheck` from the `media-feature-name-*` rules.

Also, this adds a type declaration for the [`postcss-media-query-parser`](https://github.com/dryoma/postcss-media-query-parser) package.


types/postcss-media-query-parser/index.d.ts
```diff
@@ -0,0 +1,33 @@
+declare module 'postcss-media-query-parser' {
```
ybiquitous, line 1: [note] I've created the minimum required types to pass the type-check.

lib/rules/rangeContextNodeParser.js
```diff
@@ -23,6 +26,9 @@ function getRangeContextOperators(node) {
 	return operators.sort((a, b) => b.length - a.length);
 }
```
ybiquitous, line 28:
```ts
const before = node[match.startIndex - 1];
```

This line has been added via #4581, but I think the type of `node` should be `import('postcss-media-query-parser').Node`. So, the type error occurs like this:

> Element implicitly has an 'any' type because expression of type 'number' can't be used to index type 'Node'.

Why? 🤔

ybiquitous: I've fixed the type-check error via the c26238b1 commit. Perhaps, I guess that `(before === '>' || before === '<')` in the `if` condition may exclude a pattern like `<<` or `>>` etc.

jeddy3 (approved):
LGTM, thanks!

ybiquitous:
@jeddy3 Thanks for the review! 😄

PR #5513, opened 2021-09-04
Refactor to improve types for `named-grid-areas-no-invalid` rule

ybiquitous:

Part of #4496

This change removes `// @ts-nocheck` from the `named-grid-areas-no-invalid` rule.

### DIFF
--- a/lib/rules/named-grid-areas-no-invalid/index.js
+++ b/lib/rules/named-grid-areas-no-invalid/index.js
@@ -1,5 +1,3 @@
-// @ts-nocheck
-
 'use strict';
 
 const declarationValueIndex = require('../../utils/declarationValueIndex');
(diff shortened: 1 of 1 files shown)


jeddy3 (approved):
LGTM, thank you.

---

Incidentally, we should probably apply this rule to the [`grid`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template) and [`grid-template`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template) shorthands. I'll create an issue for it. I've been using the shorthands recently and they're great for combining areas with explicit row/column sizes, e.g.:

```css
a {
  grid-template:
    "a a" 1fr
    "b b" 1rem
    / 1fr max-content;
}
```

ybiquitous:
@jeddy3 Thanks for the review and nice suggestion about `grid-template` shorthand! I agree with the suggestion. 👍🏼

PR #5515, opened 2021-09-05
Refactor atrule statement in unit-no-unknown

sylwia-lask:

PR is related to discussion in #5500 - introduce isStandardAtRule. 

No, it's self-explanatory.

### DIFF
--- a/lib/rules/unit-no-unknown/index.js
+++ b/lib/rules/unit-no-unknown/index.js
@@ -5,6 +5,7 @@
 const atRuleParamIndex = require('../../utils/atRuleParamIndex');
 const declarationValueIndex = require('../../utils/declarationValueIndex');
 const getUnitFromValueNode = require('../../utils/getUnitFromValueNode');
+const isStandardSyntaxAtRule = require('../../utils/isStandardSyntaxAtRule');
 const isStandardSyntaxDeclaration = require('../../utils/isStandardSyntaxDeclaration');
 const keywordSets = require('../../reference/keywordSets');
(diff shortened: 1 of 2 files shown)


lib/rules/unit-no-unknown/index.js
```diff
@@ -129,7 +130,7 @@ function rule(actual, options) {
 		}
 
 		root.walkAtRules((atRule) => {
-			if (!/^media$/i.test(atRule.name) && !atRule.variable) {
+			if (!/^media$/i.test(atRule.name) && !isStandardSyntaxAtRule(atRule)) {
```
ybiquitous, line 133: [suggestion] Can we move the position of `/^media$/i` as below? I think it's more readable. 😃 

```js
			if (!isStandardSyntaxAtRule(atRule)) {

```

ybiquitous (commented):
@sylwia-lask Thanks for creating the PR! I've commented on a refactoring suggestion, so can you check it out?

lib/rules/unit-no-unknown/index.js
jeddy3: Yep, we should be using the `walk*` filter when available. I think most rules do, but it looks like this one must have slipped through at some point.

sylwia-lask: @ybiquitous @jeddy3 thanks a lot guys for your review :)

ybiquitous (approved):
@sylwia-lask Thanks. LGTM! 😄

jeddy3 (approved):
@sylwia-lask Looks great, thank you!

Issue #5142, opened 2021-02-10
Add all options to configuration object


jeddy3:
- [x] [`report*`](https://github.com/stylelint/stylelint/issues/3858)
- [ ] [`files`](https://github.com/stylelint/stylelint/issues/3860)
- [ ] [`fix`](https://github.com/stylelint/stylelint/issues/3859)
- [ ] [`customSyntax`](https://github.com/stylelint/stylelint/issues/4843)

There maybe more, but we can start there.

Once these are in place we can approach the `overrides` property (https://github.com/stylelint/stylelint/issues/3128).

And once that's in place, we can decide what approach we're going to take with non-standard syntax, i.e. continue with trying to autodetect, like Prettier, or move to configuration-based approach, like ESLint.

ybiquitous:
@jeddy3 What status is the checklist in this issue description?

jeddy3:
I've:
- checked `customSyntax` as done
- removed `maxWarnings` as the original request was closed by the author. 

I believe the list is accurate now. 

---

Now that `14.0.0` has been released, I'm going to make some time over the coming weeks to help go through our open issues as I suspect a number of them should either be closed or updated.

Issayev:
Is it possible to add  config file 'ignorePath' option support in the future?  It has been already asked earlier in #4126 along with 'ignorePattern' option. I don't know current state of the second option, but 'ignorePath'  doesn't work when it's being included in config file.

oliviertassinari:
+1 for `ignorePath`. I would like to be able to do

```jsx
export default {
  extends: 'stylelint-config-standard',
  ignorePath: ['.gitignore', '.eslintignore'],
``` 

For two reasons:

1. So that https://github.com/stylelint/vscode-stylelint is aware to not lint the file when it opens it (right now it reports confusing errors)
2. So that I can share this config with other repositories, without having to duplicate --ignore-path.

jeddy3:
@oliviertassinari Thanks for chiming in with your use case. I've [created an issue](https://github.com/stylelint/stylelint/issues/8345) for us all to discuss.

Mouvedia:
In light of #8422, I have untick `cache`.

PR #5521, opened 2021-09-12
Add `overrides` option to config (#3128)

### LINKED ISSUE #3128: Add overrides property to configuration object

hudochenkov:

Closes #3128.

It took quite some time to figure out how config resolutions works. Codebase is very tangled in my opinion :(

For every linted file we produce new config from initial config. All overrides are resolved and applied on a config and file gets normal config with no `overrides` property.

I took `overrides` docs from [ESLint documentation](https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-based-on-glob-patterns) with some modifications.

Interesting thing that all properties that regular config support, are supported by overrides. So when we https://github.com/stylelint/stylelint/issues/5142 is done, all options would be working automatically for `overrides`.

### DIFF
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -421,7 +421,7 @@ _Note that this is not an efficient method for ignoring lots of files._ If you w
 
 ## `overrides`
 
-Sometimes a more fine-controlled configuration is necessary, for example, if the configuration for files within the same directory has to be different. Therefore you can provide configurations under the `overrides` key that will only apply to files that match specific glob patterns, using the same format you would pass on the command line (e.g., `app/**/*.test.css`).
+You can provide configurations under the `overrides` key that will only apply to files that match specific glob patterns, using the same format you would pass on the command line (e.g., `app/**/*.test.css`).
 
 It is possible to override settings based on file glob patterns in your configuration by using the `overrides` key. An example of using the `overrides` key is as follows:
 
@@ -430,14 +430,14 @@ In your `.stylelintrc.json`:
 ```json
 {
   "rules": {
-    "quotes": ["warn", "double"]
+    "string-quotes": "double"
   },
 
   "overrides": [
     {
       "files": ["components/**/*.css", "pages/**/*.css"],
       "rules": {
-        "quotes": ["warn", "single"]
+        "string-quotes": "single"
       }
     }
   ]
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -418,3 +418,35 @@ If the globs are absolute paths, they are used as is. If they are relative, they
 The `ignoreFiles` property is stripped from extended configs: only the root-level config can ignore files.
 
 _Note that this is not an efficient method for ignoring lots of files._ If you want to ignore a lot of files efficiently, use [`.stylelintignore`](ignore-code.md) or adjust your files globs.
+
+## `overrides`
+
+Sometimes a more fine-controlled configuration is necessary, for example, if the configuration for files within the same directory has to be different. Therefore you can provide configurations under the `overrides` key that will only apply to files that match specific glob patterns, using the same format you would pass on the command line (e.g., `app/**/*.test.css`).
+
+It is possible to override settings based on file glob patterns in your configuration by using the `overrides` key. An example of using the `overrides` key is as follows:
+
+In your `.stylelintrc.json`:
+
+```json
+{
+  "rules": {
+    "quotes": ["warn", "double"]
+  },
+
+  "overrides": [
+    {
+      "files": ["components/**/*.css", "pages/**/*.css"],
+      "rules": {
+        "quotes": ["warn", "single"]
+      }
+    }
+  ]
+}
+```
+
+Here is how overrides work in a configuration file:
+
+- The patterns are applied against the file path relative to the directory of the config file. For example, if your config file has the path `/Users/person/workspace/any-project/.stylelintrc.js` and the file you want to lint has the path `/Users/person/workspace/any-project/components/card.css`, then the pattern provided in `.stylelintrc.js` will be executed against the relative path `components/card.css`.
+- Glob pattern overrides have higher precedence than the regular configuration in the same config file. Multiple overrides within the same config are applied in order. That is, the last override block in a config file always has the highest precedence.
+- A glob specific configuration works almost the same as any other stylelint config. Override blocks can contain any configuration options that are valid in a regular config.
+- Multiple glob patterns can be provided within a single override block. A file must match at least one of the supplied patterns for the configuration to apply.
--- a/lib/augmentConfig.js
+++ b/lib/augmentConfig.js
@@ -4,13 +4,16 @@ const configurationError = require('./utils/configurationError');
 const getModulePath = require('./utils/getModulePath');
 const globjoin = require('globjoin');
 const merge = require('deepmerge');
+const micromatch = require('micromatch');
 const normalizeAllRuleSettings = require('./normalizeAllRuleSettings');
 const path = require('path');
+const slash = require('slash');
 
 /** @typedef {import('stylelint').StylelintConfigPlugins} StylelintConfigPlugins */
 /** @typedef {import('stylelint').StylelintConfigProcessor} StylelintConfigProcessor */
 /** @typedef {import('stylelint').StylelintConfigProcessors} StylelintConfigProcessors */
 /** @typedef {import('stylelint').StylelintConfigRules} StylelintConfigRules */
+/** @typedef {import('stylelint').StylelintConfigOverride} StylelintConfigOverride */
 /** @typedef {import('stylelint').StylelintInternalApi} StylelintInternalApi */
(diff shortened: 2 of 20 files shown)


docs/user-guide/configure.md
```diff
@@ -418,3 +418,35 @@ If the globs are absolute paths, they are used as is. If they are relative, they
 The `ignoreFiles` property is stripped from extended configs: only the root-level config can ignore files.
 
 _Note that this is not an efficient method for ignoring lots of files._ If you want to ignore a lot of files efficiently, use [`.stylelintignore`](ignore-code.md) or adjust your files globs.
+
+## `overrides`
+
+Sometimes a more fine-controlled configuration is necessary, for example, if the configuration for files within the same directory has to be different. Therefore you can provide configurations under the `overrides` key that will only apply to files that match specific glob patterns, using the same format you would pass on the command line (e.g., `app/**/*.test.css`).
```
jeddy3, line 424:
```suggestion
You can provide configurations under the `overrides` key that will only apply to files that match specific glob patterns, using the same format you would pass on the command line (e.g., `app/**/*.test.css`).
```

I believe the first line refers to ESLint's behaviour of merging nested configuration files. Stylelint doesn't have this behaviour.

docs/user-guide/configure.md
```diff
@@ -418,3 +418,35 @@ If the globs are absolute paths, they are used as is. If they are relative, they
...
+
+It is possible to override settings based on file glob patterns in your configuration by using the `overrides` key. An example of using the `overrides` key is as follows:
+
+In your `.stylelintrc.json`:
+
+```json
+{
+  "rules": {
+    "quotes": ["warn", "double"]
```
jeddy3, line 433:
```suggestion
    "string-quotes": "double"
```

docs/user-guide/configure.md
```diff
@@ -418,3 +418,35 @@ If the globs are absolute paths, they are used as is. If they are relative, they
...
+  },
+
+  "overrides": [
+    {
+      "files": ["components/**/*.css", "pages/**/*.css"],
+      "rules": {
+        "quotes": ["warn", "single"]
```
jeddy3, line 440:
```suggestion
        "string-quotes": "single"
```

lib/augmentConfig.js
```diff
@@ -370,4 +404,53 @@ function addProcessorFunctions(config) {
...
+function applyOverrides(fullConfig, configDir, filePath) {
+	let { overrides, ...config } = fullConfig;
+
+	if (!overrides) {
+		return config;
+	}
+
+	if (!Array.isArray(overrides)) {
+		throw Error(
+			'`overrides` option in a config should be an array. E. g. { "overrides": [{ "files": "*.css", "rules": {}}] }',
```
jeddy3, line 422:
```suggestion
			'The `overrides` configuration property should be an array, e.g. { "overrides": [{ "files": "*.css", "rules": {}}] }.',
```

We can align with the language we use [in the configuration docs](https://stylelint.io/user-guide/configure).

lib/augmentConfig.js
```diff
@@ -370,4 +404,53 @@ function addProcessorFunctions(config) {
...
+		);
+	}
+
+	for (const override of overrides) {
+		const { files, ...configOverrides } = override;
+
+		if (!files) {
+			throw Error(
+				'Every object in config `overrides` option should have `files` property with globs.',
```
jeddy3, line 431:
```suggestion
				'Every object in the `overrides` configuration property should have a `files` property with globs, e.g. { "overrides": [{ "files": "*.css", "rules": {}}] }.',
```

We can align with other error.

jeddy3 (changes requested):
@hudochenkov This is amazing, thank you!

The code and tests look good to me.

I've suggested some minor tweaks to the documentation and error message wording.

Many thanks again! This is an exciting addition that gets us closer to 14.0.0.

jeddy3 (approved):
Fantastic, thanks!

lib/augmentConfig.js
```diff
@@ -370,4 +404,53 @@ function addProcessorFunctions(config) {
...
+
+		if (!files) {
+			throw Error(
+				'Every object in the `overrides` configuration property should have a `files` property with globs, e.g. { "overrides": [{ "files": "*.css", "rules": {}}] }.',
+			);
+		}
+
+		let filesGlobs = Array.isArray(files) ? files : [files];
+
+		filesGlobs = filesGlobs
```
ybiquitous, line 437: [suggestion] How about simplifying the code a bit to use [`Array.prototype.flat()`](http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)?

```suggestion
		const filesGlobs = [files].flat()
```

lib/augmentConfig.js
```diff
@@ -370,4 +404,53 @@ function addProcessorFunctions(config) {
...
+function applyOverrides(fullConfig, configDir, filePath) {
+	let { overrides, ...config } = fullConfig;
+
+	if (!overrides) {
+		return config;
+	}
+
+	if (!Array.isArray(overrides)) {
+		throw Error(
+			'The `overrides` configuration property should be an array, e.g. { "overrides": [{ "files": "*.css", "rules": {}}] }.',
```
ybiquitous, line 422: [suggestion] How about adding extra whitespace inside of the JSON object?

```suggestion
			'The `overrides` configuration property should be an array, e.g. { "overrides": [{ "files": "*.css", "rules": {} }] }.',
```

lib/augmentConfig.js
```diff
@@ -370,4 +404,53 @@ function addProcessorFunctions(config) {
...
+		);
+	}
+
+	for (const override of overrides) {
+		const { files, ...configOverrides } = override;
+
+		if (!files) {
+			throw Error(
+				'Every object in the `overrides` configuration property should have a `files` property with globs, e.g. { "overrides": [{ "files": "*.css", "rules": {}}] }.',
```
ybiquitous, line 431: [suggestion] Also here.

```suggestion
				'Every object in the `overrides` configuration property should have a `files` property with globs, e.g. { "overrides": [{ "files": "*.css", "rules": {} }] }.',
```

lib/augmentConfig.js
```diff
@@ -370,4 +404,53 @@ function addProcessorFunctions(config) {
...
+			// Glob patterns for micromatch should be in POSIX-style
+			.map(slash);
+
+		if (micromatch.isMatch(filePath, filesGlobs)) {
+			config = mergeConfigs(config, configOverrides);
+		}
+	}
+
+	return config;
+}
```
ybiquitous, line 454: [question] Does it make sense to check the `rules` key existence in addition to `files`?

For example:

```js
if (!rules) {
	throw Error(
		'Every object in the `overrides` configuration property should have a `rules` property ...',
	);
}
```

ybiquitous (commented):
@hudochenkov Thanks, this is awesome work! 😄 
I've left some comments, but I think this PR is mostly fine!

lib/augmentConfig.js
jeddy3: It doesn't necessarily need a `rules` key as the user could override something else, e.g. `customSyntax` when linting both CSS and SCSS with `stylelint "**/*.{css,scss}"`:

```json
{
  "overrides": [{
    "files": "*.scss",
    "customSyntax": "postcss-scss"
  }],
  "rules": {
    ..
  }
}
```

We could check that there's at least one other key than `files`, though.

ybiquitous: @jeddy3 I see. Thank you!

lib/augmentConfig.js
hudochenkov: Great suggestion! I was thinking about it, but thought it's a new API, and didn't check MDN for availability :)

ybiquitous (approved):
@hudochenkov Thank you! LGTM! 👏🏼

jeddy3:
`v14` changelog:

> - Added: `overrides` property to configuration object ([#5521](https://github.com/stylelint/stylelint/pull/5521)).

jeddy3:
> Codebase is very tangled in my opinion :(

I'll create an issue so that we can discuss what can we do about this.

PR #5538, opened 2021-09-14
Add `customSyntax` option to configuration object (#4843)

### LINKED ISSUE #4843: Add customSyntax option to configuration object

hudochenkov:

Closes #4843.

It was easier than I though :) Two previous PRs I made recently definitely helped :)

---

Something that we might need to document somewhere or changes. Probably I had to write this in `overrides` PR.

It is true for v13 (except `overrides`) and for all options which could be specified in config, API, and CLI.

Priority of options:

1. `overrides` in config.
2. Option specified in CLI/API.
3. Option in config.

For example, if we have this config:

```json
{
	"customSyntax": "postcss-sass",
	"rules": {
		"block-no-empty": true
	},
	"overrides": [{
		"files": "*.css",
		"customSyntax": "postcss-safe-parser"
	}]
}
```

And run stylelint like this:

```
stylelint --custom-syntax "postcss-scss" "style.css"
```

File would be linted with `postcss-safe-parser`. If were remove `overrides` file would be linted with `postcss-scss`. And finally if we remove `--custom-syntax` file would be linted with `postcss-sass`.

Maybe we need to switch `overrides` and CLI/API order, so CLI/API has highest priority than `overrides`.

### DIFF
--- a/lib/__tests__/fixtures/custom-parser.js
+++ b/lib/__tests__/fixtures/custom-parser.js
@@ -1,3 +1,3 @@
 'use strict';
 
-module.exports = require('postcss-safe-parser/lib/safe-parse');
+module.exports = require('postcss-safe-parser');
--- a/lib/__tests__/fixtures/custom-syntax.js
+++ b/lib/__tests__/fixtures/custom-syntax.js
@@ -1,3 +1,8 @@
 'use strict';
 
-module.exports = require('postcss-scss/lib/scss-syntax');
+const postcssScss = require('postcss-scss');
+
+module.exports = {
+	parse: postcssScss.parse,
+	stringify: postcssScss.stringify,
+};
--- a/lib/__tests__/standalone-syntax.test.js
+++ b/lib/__tests__/standalone-syntax.test.js
@@ -97,6 +97,52 @@ it('standalone with path to custom syntax', () => {
 	});
 });
 
+it('standalone with custom syntax as npm package name', () => {
+	const config = {
+		rules: {
+			'block-no-empty': true,
+		},
+	};
+
+	return standalone({
+		config,
+		customSyntax: 'postcss-scss',
+		code: '$foo: bar; // foo;\nb {}',
+		formatter: stringFormatter,
+	}).then((linted) => {
+		const results = linted.results;
+
+		expect(results).toHaveLength(1);
+		expect(results[0].warnings).toHaveLength(1);
+		expect(results[0].warnings[0].line).toBe(2);
+		expect(results[0].warnings[0].column).toBe(3);
+		expect(results[0].warnings[0].rule).toBe('block-no-empty');
+	});
+});
+
+it('standalone with custom syntax as npm package', () => {
+	const config = {
+		rules: {
+			'block-no-empty': true,
+		},
+	};
+
+	return standalone({
+		config,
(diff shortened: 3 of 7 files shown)


lib/__tests__/fixtures/custom-parser.js
```diff
@@ -1,3 +1,3 @@
 'use strict';
 
-module.exports = require('postcss-safe-parser/lib/safe-parse');
+module.exports = require('postcss-safe-parser');
```
hudochenkov, line 3: Avoid using internals.

lib/__tests__/fixtures/custom-syntax.js
```diff
@@ -1,3 +1,8 @@
 'use strict';
 
-module.exports = require('postcss-scss/lib/scss-syntax');
+const postcssScss = require('postcss-scss');
+
+module.exports = {
+	parse: postcssScss.parse,
+	stringify: postcssScss.stringify,
+};
```
hudochenkov, line 8: Better represent custom syntax.

lib/__tests__/standalone-syntax.test.js
```diff
@@ -97,6 +97,52 @@ it('standalone with path to custom syntax', () => {
...
+	}).then((linted) => {
+		const results = linted.results;
+
+		expect(results).toHaveLength(1);
+		expect(results[0].warnings).toHaveLength(1);
+		expect(results[0].warnings[0].line).toBe(2);
+		expect(results[0].warnings[0].column).toBe(3);
+		expect(results[0].warnings[0].rule).toBe('block-no-empty');
+	});
+});
```
hudochenkov, line 144: We never had these tests, but should have, as it's a documented behavior.

ybiquitous (approved):
The code changes and "Priority of options" look good to me. 👍🏼

jeddy3 (approved):
Excellent stuff, thanks!

> Maybe we need to switch overrides and CLI/API order, so CLI/API has highest priority than overrides

It's a good point. I think the current order is OK as I see the CLI flags as setting their respective properties in the config file. For example, a hypothetical `overrides` flag would work like...

Given:

```json
{
	"customSyntax": "postcss-sass",
	"rules": {
		"block-no-empty": true
	},
	"overrides": [{
		"files": "*.css",
		"customSyntax": "postcss-safe-parser"
	}]
}
```

And run stylelint like this:

```
stylelint --custom-syntax "postcss-scss" --overrides "files: ["*.css"], customSyntax: "postcss-less" "style.css"
```

The executed config would be:

```json
{
	"customSyntax": "postcss-scss",
	"rules": {
		"block-no-empty": true
	},
	"overrides": [{
		"files": "*.css",
		"customSyntax": "postcss-safe-parser"
	},
        {
		"files": "*.css",
		"customSyntax": "postcss-less"
	}]
}
```

And so postcss-less is used on the `style.css` file.

jeddy3:
> - Added: `customSyntax` option as a property in the configuration object ([#5538](https://github.com/stylelint/stylelint/pull/5538)).

Issue #4476, opened 2019-12-04
Add quiet option to Node API


duanehutchins:
> Clearly describe the bug

The Node API ignores the `quiet` option.  You can confirm that the `quiet` option is not used in `lib/standalone.js`.

Example gulpfile.js:

```gulpfile.js
gulp.task('stylelint', () => {
    const stylelint = require('stylelint');
    const options = Object.assign(argv, {
        files: 'sass/**/*.scss',
        formatter: 'string',
        quiet: true,
      });
    return stylelint.lint(options).then(data => {
        console.log(data.output);
    });
});
```

Even though `quiet: true`, the warnings are not suppressed.

> Which version of stylelint are you using?

`12.0.0`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

Node.js API

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

No

> What did you expect to happen?

Warnings should be suppressed in quiet mode.  Example:

```shell
$ npm run stylelint
/src/sass/_mixins.scss:101:2: Expected single space before "{" (block-opening-brace-space-before) [error]

1 problem
```

> What actually happened (e.g. what warnings or errors did you get)?

Warnings are not suppressed.  Example:
```shell
$ npm run stylelint
/src/sass/_mixins.scss:101:2: Expected single space before "{" (block-opening-brace-space-before) [error]
/src/sass/_mixins.scss:234:18: Unexpected !important (declaration-no-important) [warning]
/src/sass/_mixins.scss:238:20: Unexpected !important (declaration-no-important) [warning]

3 problems
```

evilebottnawi:
i think `quiet` option only for CLI

jeddy3:
Is there any reason we don't allow it to be used by the Node API? If not, we could add it.

jeddy3:
As it's a valid use case (where a user wants `data.output` to be filtered of errors when using the Node.js API) and there are no objections, I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time. 

You'll need to move `quiet` out from [the CLI docs](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/usage/cli.md) and into the [Options](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/usage/options.md) docs in the pull request.

glen-84:
@jeddy3 I think this should be closed? (#5542)

Issue #5529, opened 2021-09-13
Make engine code more straightforward


jeddy3:
The stylelint code base has grown over time, especially around the engine. How can we untangle that part of the code to make it easier to add features, like `overrides`, in the future?

Is it something we can do iteratively through refactors or are more substantial changes needed? For example, if providing stylelint as a PostCSS plugin is a cause, then we could drop that in favour of just the Node API and CLI.

(Caveat: the engine is the part of stylelint that I know the least about. I focused on the rules, even from the very beginning.)

I believe @m-allanson's [pull request for the browser entry point](https://github.com/stylelint/stylelint/pull/4796#issue-421326478) (and the refactoring that came out of it) is the only work we've done to map out and refactor the internals before, so feel free to share which areas are problematic and any ideas you may have for untangling the code.

Ref: https://github.com/stylelint/stylelint/pull/5521#issue-732131887

m-allanson:
Here's a couple of thoughts on this. From memory, so maybe inaccurate.

1. standalone.js is a good candidate for refactoring. The deep nested promises make it difficult to follow the logic. The module is essentially two different paths to:

```js
stylelint._lintsource().then(() => {
  return stylelint._createStylelintResult();
});
```
But it's quite tricky to follow the logic through.

A while ago I tried switching the promises over to use `await` instead. I made some progress before noticing intermittent failures in the cache tests. Some more debugging is needed on that (maybe this is fixed by https://github.com/stylelint/stylelint/pull/5533).

2. Processors. If it were possible to remove processors some of stylelint's internal logic could be simplified. Processors can modify code on the way in, and results on the way out.

I'm not up to date on processors, but maybe the following statements are true... Syntaxes replace processors? But the CSS-in-JS syntax lacks some features available in the styled-components processor? And the recent addition of `Document` nodes (in postcss) means it's now possible to create syntaxes to replace the styled components processor?

jeddy3:
> A while ago I tried switching the promises over to use `await` instead.

Nice. That'd make it easier to grok what's going on in the file.

> And the recent addition of Document nodes (in postcss) means it's now possible to create syntaxes to replace the styled components processor?

Yes, we may be able to remove the deprecated processor functionality in favour of only using syntaxes in `14.0.0`.

jeddy3:
[...]

PR #5542, opened 2021-09-16
Add `quiet` option to Node.js API

hudochenkov:

Related to #5529, but not closes it.

Refactored as much as possible to async-await in core. Also noticed some improvements to code.

Accidentally found that `quite` flag in Node.js API didn't work. It's probably worked only in CLI only for a long time, even thought stylelint have it since [v2.0.0](https://github.com/stylelint/stylelint/blob/master/CHANGELOG.md#200). This option wasn't even documented. I fixed it and added tests. *UPD:* found that there is an issue #4476 :)

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -326,7 +326,7 @@ const meowOptions = {
  * @param {string[]} argv
  * @returns {Promise<any>}
  */
-module.exports = (argv) => {
+module.exports = async (argv) => {
 	const cli = buildCLI(argv);
 
 	const invalidOptionsMessage = checkInvalidCLIOptions(meowOptions.flags, cli.flags);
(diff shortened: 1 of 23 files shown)


lib/getPostcssResult.js
```diff
@@ -19,73 +19,65 @@ const postcssProcessor = postcss();
...
+module.exports = async function getPostcssResult(stylelint, options = {}) {
 	const cached = options.filePath ? stylelint._postcssResultCache.get(options.filePath) : undefined;
 
-	if (cached) return Promise.resolve(cached);
+	if (cached) {
+		return cached;
+	}
+
+	if (stylelint._options.syntax) {
+		return Promise.reject(
```
hudochenkov, line 30: When `throw` and error in async function it is converted to rejected Promise. So for being more explicit I converted all `throw` into `return Promise.reject()` through this PR.

ybiquitous:
> This option wasn't even documented. I fixed it and added tests.

Great! Does it make sense to add the documentation fix to this PR title?

lib/cli.js
```diff
@@ -326,7 +326,7 @@ const meowOptions = {
  * @param {string[]} argv
  * @returns {Promise<any>}
  */
-module.exports = (argv) => {
+module.exports = async (argv) => {
```
ybiquitous, line 329: [suggestion] How about changing to a named function like other modules?

```suggestion
module.exports = async function cli(argv) {
```

lib/lintSource.js
```diff
@@ -34,74 +34,60 @@ module.exports = function lintSource(stylelint, options = {}) {
...
+		(await stylelint._getPostcssResult({
+			code: options.code,
+			codeFilename: options.codeFilename,
+			filePath: inputFilePath,
+			codeProcessors: config.codeProcessors,
+			customSyntax: config.customSyntax,
+		}));
+
+	const stylelintPostcssResult = Object.assign(postcssResult, {
+		stylelint: stylelintResult,
```
m-allanson, line 85: This whole section is much simpler now. 👍 

Style nitpick, feel free to ignore. Could change to something like:

```js
const stylelintPostcssResult = {
  ...postcssResult,
  stylelint: {
    ruleSeverities: {},
    customMessages: {},
    disabledRanges: {},
  },
};
```
And then remove that `stylelintResult` from a few lines above?

m-allanson (approved):
I love it 😆

jeddy3 (approved):
Fantastic. So much cleaner and more understandable!

> This option wasn't even documented. I fixed it and added tests. UPD: found that there is an issue #4476 :)

Let's rename the pull request to:

"Add quiet option to Node.js API"

(with the refactoring being an amazing icing on the cake!)

Closes #4476

lib/lintSource.js
hudochenkov: I'm worried that code could rely on object mutation. I've seen it in some places. `Object.assign` will change original object, while spread will create a new one.

lib/cli.js
hudochenkov: I can't do that because `cli` variable declared right on the next line and used across whole function :)

lib/lintSource.js
m-allanson: 👍 makes sense.

lib/cli.js
ybiquitous: Thanks! I see. 😃

ybiquitous (approved):
LGTM! 👍🏼 

And, sorry for the conflict by #5544 merged... 🙇🏼

ybiquitous (approved):
Thank you 👍🏼

jeddy3:
> - Added: `quiet` option to Node.js API ([#5542](https://github.com/stylelint/stylelint/pull/5542)).

PR #5545, opened 2021-09-17
lint script: add `--continue-on-error`

XhmikosR:
This allows the other scripts to finish instead of the process exiting when any error is first encountered in any of the scripts.

BTW @jeddy3 the v14 isn't protected, not sure if it's intentional :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "benchmark-rule": "node scripts/benchmark-rule.js",
     "format": "prettier . --write",
     "jest": "jest",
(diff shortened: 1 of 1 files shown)


jeddy3 (approved):
LGTM, thanks

> BTW @jeddy3 the v14 isn't protected, not sure if it's intentional :)

Fast and loose :)

It'll be merged into `master` (soon to be `main`) soon enough.

ybiquitous (approved):
LGTM 😄

PR #5546, opened 2021-09-17
Refactor based on latest XO recommendations

XhmikosR:
Split from my xo branch. These are mostly autofixes.

Non-whitespace diff: https://github.com/stylelint/stylelint/pull/5546/files?w=1

### DIFF
--- a/lib/__tests__/disableRanges.test.js
+++ b/lib/__tests__/disableRanges.test.js
@@ -990,8 +990,8 @@ function expectDisableRanges(result, expected) {
 
 		expect(actualRanges).toHaveLength(expectedRanges.length);
 
-		for (let i = 0; i < expectedRanges.length; i++) {
-			expectDisableRange(actualRanges[i], expectedRanges[i]);
+		for (const [i, expectedRange] of expectedRanges.entries()) {
+			expectDisableRange(actualRanges[i], expectedRange);
 		}
 	}
 }
--- a/lib/rules/color-named/index.js
+++ b/lib/rules/color-named/index.js
@@ -113,8 +113,8 @@ const rule = (primary, secondaryOptions) => {
 					const normalizedFunctionString = valueParser.stringify(node).replace(/\s+/g, '');
 					let namedColor;
 
-					for (let i = 0, l = namedColors.length; i < l; i++) {
-						namedColor = namedColors[i];
+					for (const namedColor_ of namedColors) {
+						namedColor = namedColor_;
 
 						if (namedColorData[namedColor].func.includes(normalizedFunctionString.toLowerCase())) {
 							complain(
@@ -133,8 +133,8 @@ const rule = (primary, secondaryOptions) => {
 				// Then by checking for alternative hex representations
 				let namedColor;
 
-				for (let i = 0, l = namedColors.length; i < l; i++) {
-					namedColor = namedColors[i];
+				for (const namedColor_ of namedColors) {
+					namedColor = namedColor_;
 
 					if (namedColorData[namedColor].hex.includes(value.toLowerCase())) {
 						complain(
--- a/lib/rules/color-named/index.js
+++ b/lib/rules/color-named/index.js
@@ -111,11 +111,8 @@ const rule = (primary, secondaryOptions) => {
 				if (type === 'function' && keywordSets.colorFunctionNames.has(value.toLowerCase())) {
 					// Remove all spaces to match what's in `representations`
 					const normalizedFunctionString = valueParser.stringify(node).replace(/\s+/g, '');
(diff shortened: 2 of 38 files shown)


lib/rules/color-named/index.js
```diff
@@ -113,8 +113,8 @@ const rule = (primary, secondaryOptions) => {
 					const normalizedFunctionString = valueParser.stringify(node).replace(/\s+/g, '');
 					let namedColor;
 
-					for (let i = 0, l = namedColors.length; i < l; i++) {
-						namedColor = namedColors[i];
+					for (const namedColor_ of namedColors) {
+						namedColor = namedColor_;
```
jeddy3, line 117: Is this right?

jeddy3 (commented):
Thanks!

Only one query, otherwise LGTM.

lib/rules/color-named/index.js
XhmikosR: TBH this baffled me too. It seems that `namedColor` is shared between the two loops. Maybe they could/should be combined later?

ybiquitous (commented):
Thanks, mostly LGTM 👍🏼 

I think some type annotations are no longer required thanks to the idiom using `flat()`.

XhmikosR:
@ybiquitous can you push your changes please? I'm not familiar with TS almost at all :)

lib/rules/color-named/index.js
ybiquitous: I think this PR is good timing to combine them! 👍🏼

ybiquitous:
OK, just a minute. 👌🏼

ybiquitous:
Done. bf6f4abb

lib/rules/color-named/index.js
XhmikosR: Feel free to push here :) Or someone could do it in another PR later

ybiquitous: Fixed! 9ae83efa

ybiquitous (approved):
@XhmikosR Thank you for cleaning the code! 👏🏼

XhmikosR:
NP, thanks for the help :) I hope nothing broke because I did hit a wrong autofix here: https://github.com/stylelint/stylelint/blob/5c55141f53b62e0d452285fc4460e2ae0fe710c6/lib/rules/no-extra-semicolons/index.js#L30 which resulted in:

```
TypeError: string.entries is not a function or its return value is not iterable
```

ybiquitous:
I see. `String` does not have `entries()` method, not like `Array`. I think it is correct not to auto-fix. 😃

XhmikosR:
Yeah, it's just that it couldn't detect the type hence the wrong fix. Too lazy to report it...

I think the rest `for...of` changes are OK, though.

lib/__tests__/disableRanges.test.js
```diff
@@ -990,8 +990,8 @@ function expectDisableRanges(result, expected) {
 
 		expect(actualRanges).toHaveLength(expectedRanges.length);
 
-		for (let i = 0; i < expectedRanges.length; i++) {
-			expectDisableRange(actualRanges[i], expectedRanges[i]);
+		for (const [i, expectedRange] of expectedRanges.entries()) {
```
alexander-akait, line 993: Just note - simple `for` is faster then `for/of` in many cases

ybiquitous:
Perhaps `Object.entries()` may be available like this, but I think it is good to do it with another pull request. 😃 

```js
for (const [i, char] of Object.entries(string)) { ... }
```

lib/__tests__/disableRanges.test.js
alexander-akait: But it is very small optimizations and most likely will not affect anything, excluding parts that really need to be optimized

XhmikosR: Yeah, I'm aware of that, but for of is used already in plenty of places in the codebase.

```
C:\Users\xmr\Desktop>matcha bench-loop.js

        125,000,000 ops/sec > for (post) (67.5x)
        110,000,000 ops/sec > for (pre) (59.5x)
        111,000,000 ops/sec > for (cached length) (59.9x)
        111,000,000 ops/sec > forEach (60.1x)
         76,000,000 ops/sec > for of (41.2x)
          2,190,000 ops/sec > for in (1.19x)
          1,850,000 ops/sec > for in (own) (1x)

  Benches: 7
  Fastest: for (post)
  Elapsed: 38.8s
```

That being said, I don't think this is a real issue and readability plus the fact that you can have off by one errors is better :)

jeddy3 (approved):
LGTM, thanks!

PR #5548, opened 2021-09-17
Refactor to add missing radix

XhmikosR:
I believe `10` is meant to be used here. Either way, I think the ESLint config should have the rule enabled to catch these :)

### DIFF
--- a/lib/utils/isCounterIncrementCustomIdentValue.js
+++ b/lib/utils/isCounterIncrementCustomIdentValue.js
@@ -12,7 +12,7 @@ module.exports = function (value) {
 
 	if (
 		keywordSets.counterIncrementKeywords.has(valueLowerCase) ||
-		Number.isFinite(Number.parseInt(valueLowerCase))
+		Number.isFinite(Number.parseInt(valueLowerCase, 10))
 	) {
 		return false;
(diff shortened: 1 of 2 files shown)


ybiquitous (approved):
Good catch! It looks like we need the ESLint [`radix`](https://eslint.org/docs/rules/radix) rule in our sharable config. 😃

XhmikosR:
Yup, I noticed it in the xo branch because it's enabled there but you should also add it in your config :)

ybiquitous:
Ah, I understand. 👍🏼 
https://github.com/stylelint/eslint-config-stylelint/issues/77

XhmikosR:
No, I'd just add it in your config without the need for xo. It's a core ESLint rule after all/

ybiquitous:
Add to my config? I thought it would be helpful to add the `radix` rule to [`eslint-config-stylelint`](https://github.com/stylelint/eslint-config-stylelint/blob/a488e36b9dd5c242924d44f4397dfe4cc8afdcf9/.eslintrc.js). 🤔

XhmikosR:
Yeah, that's what I meant :) In the config that's used here.

Issue #5291, opened 2021-05-12
Move to ESM


jeddy3:
We'll need to:

1. following the [ESM migration guide](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c):
2. replace lazy import with dynamic imports

---

> ### How can I move my CommonJS project to ESM?

> - Add `"type": "module"` to your package.json.
> - Replace `"main": "index.js"` with `"exports": "./index.js"` in your package.json.
> - Remove `'use strict';` from all JavaScript files.
> - Replace all `require()`/`module.export` with `import`/`export`.
> - Use only full relative file paths for imports: `import x from '.';` → `import x from './index.js';`.
> - If you have a TypeScript type definition (for example, `index.d.ts`), update it to use ESM imports/exports.
> - Optional but recommended, use the [`node:` protocol](https://nodejs.org/api/esm.html#esm_node_imports) for imports.

ybiquitous:
@jeddy3 Should we create a long-lived branch such as `esm`?

jeddy3:
Yep. I've created a [`v14` branch](https://github.com/stylelint/stylelint/tree/v14) for the issues listed in https://github.com/stylelint/stylelint/issues/5205#issue-837034702

> Replace all `require()/module.export` with `import/export`.

We can use https://github.com/knowbee/rona for this. 

> Optional but recommended, use the node: protocol for imports.

Find and replace should suffice for this.

> replace lazy import with dynamic imports

This is probably the trickiest aspect of this issue.

hudochenkov:
Shall we wait with ESM until Jest has non-experimental support for it?

jeddy3:
Let's transpile for Jest for now:
https://bl.ocks.org/rstacruz/511f43265de4939f6ca729a3df7b001c

So that we can update our growing list of dependencies that require ESM.

We can move from transpiling to native ESM Jest when it's available.

jeddy3:
This issue is now unblocked and can be started on.

We can use [putout](https://github.com/coderaiser/putout) to do the bulk of the `import` conversion.

Putting a `.putout.json` config in the root:

```json
{
    "rules": {
        "convert-commonjs-to-esm/require": "on",
        "convert-commonjs-to-esm/exports": "on",
        "convert-commonjs-to-esm/commons": "off",
        "apply-destructuring/array": "off",
        "convert-to-arrow-function": "off",
        "apply-destructuring/object": "off",
        "remove-iife": "off",
[...]

PR #5550, opened 2021-09-18
Refactor `cli.js` a bit

ybiquitous:

Removing the extra `require` may make the ESM migration easier. See <https://github.com/stylelint/stylelint/issues/5291#issuecomment-920778090>.

- Remove the `pkg` option of `meow`. By default, `meow` finds the closest `package.json`.
  See <https://github.com/sindresorhus/meow/blob/v9.0.0/readme.md#pkg>
- Change `@ts-ignore` to `@ts-expect-error`.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -318,8 +318,6 @@ const meowOptions = {
 			type: 'boolean',
 		},
 	},
-	pkg: require('../package.json'),
-	argv: /** @type {string[]} */ ([]),
 };
 
 /**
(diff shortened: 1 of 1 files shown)


jeddy3 (approved):
LGTM, thanks!

([Coveralls down for maintainance](https://status.coveralls.io/))

ybiquitous:
Oh, thanks for sharing the maintenance! 👍🏼

PR #5557, opened 2021-09-22
Refactor to fix `@ts-ignore` in `isStandardSyntaxDeclaration()`

ybiquitous:

Part of #4496

No, it's self-explanatory.

### DIFF
--- a/lib/utils/isStandardSyntaxDeclaration.js
+++ b/lib/utils/isStandardSyntaxDeclaration.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const isScssVariable = require('./isScssVariable');
-const { isRoot } = require('./typeGuards');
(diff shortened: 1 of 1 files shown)


lib/utils/isStandardSyntaxDeclaration.js
```diff
@@ -54,11 +54,10 @@ module.exports = function (decl) {
 
 	// Sass nested properties (e.g. border: { style: solid; color: red; })
 	if (
-		// @ts-ignore TODO TYPES selector does not exists
+		parent &&
+		isRule(parent) &&
```
ybiquitous, line 58: [note] Only `Rule` node has a `selector` property. See: https://postcss.org/api/#rule-selector

hudochenkov (approved):

jeddy3 (approved):
LGTM, thanks!

PR #5560, opened 2021-09-23
Refactor to replace chalk with nanocolors

ai:
The copy of https://github.com/stylelint/stylelint/pull/5559 but on top for `v14` branch

PR #5562, opened 2021-09-25
Refactor to replace `log-symbols` with custom implementation

ybiquitous:

Related to PR #5560

This refactoring replaces the `log-symbols` package with a custom implementation to remove the dependency on the `chalk` package as a sub-dependency.
(its direct dependency was removed with PR #5560)

Note that this custom implementation does not have a fallback like `log-symbols` which supports Unicode unsupported environments.

See also:
- https://github.com/sindresorhus/log-symbols/blob/v5.0.0/index.js

### DIFF
--- a/lib/formatters/stringFormatter.js
+++ b/lib/formatters/stringFormatter.js
@@ -2,9 +2,8 @@
 
 const path = require('path');
 const stringWidth = require('string-width');
-const symbols = require('log-symbols');
 const table = require('table');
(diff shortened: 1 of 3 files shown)


jeddy3 (approved):
LGTM, thanks!

Once less ESM-only dependency for https://github.com/stylelint/stylelint/issues/5291 too.

ybiquitous:
@jeddy3 Thanks for the review! 😄

PR #5581, opened 2021-10-08
Fix TypeError for spaceless condition in media-feature-name-value-allowed-list

### LINKED ISSUE #5577: Fix TypeError for spaceless condition in media-feature-name-value-allowed-list
> Clearly describe the bug

Stylelint throws error when "media-feature-name-value-allowed-list" rule is defined for the following SCSS code:
SCSS code looks like this:
```scss
// This code causes error (even though it's correct SCSS code). Important note: there is no space between "and" and "("
        @media only screen and(min-width: 640px) {
            display: none;
        }
// Next code is OK, notice that space is aded between "and" and "("
        @media only screen and (min-width: 640px) {
            display: none;
        }
```

Error looks like this:
```
(node:1652) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'value' of undefined
    at base.scss:995:13
    at node_modules\stylelint\lib\rules\media-feature-name-value-allowed-list\index.js:53:42
    at node_modules\postcss-media-query-parser\dist\nodes\Container.js:62:33)
    at node_modules\postcss-media-query-parser\dist\nodes\Container.js:65:28)
    at node_modules\stylelint\lib\rules\media-feature-name-value-allowed-list\index.js:34:31
    at node_modules\postcss-scss\node_modules\postcss\lib\container.js:307:18
    at node_modules\postcss-scss\node_modules\postcss\lib\container.js:139:18
```

It's related to "media-feature-name-value-allowed-list"

```js
{
  "rules": {
        'media-feature-name-value-allowed-list': [
            {
                'max-width': ['640px'],
            },
        ],
  }
}
```

"stylelint": "13.13.1"

It doesn't depend on how it's run. I was able to reproduce it by running via CLI and by running via Webstorm.

No

Rule doesn't throw exception

Rule throws exception

lachieh:

Closes #5577 

Following the discussion in the issue, this code implements the changes requested there.

### DIFF
--- a/lib/rules/no-duplicate-at-import-rules/index.js
+++ b/lib/rules/no-duplicate-at-import-rules/index.js
@@ -38,8 +38,8 @@ const rule = (primary) => {
 					: firstParam.value;
 
 			// extract media queries if any
-			const media = mediaParser(valueParser.stringify(restParams))
-				.nodes.map((n) => n.value.replace(/\s/g, ''))
+			const media = (mediaParser(valueParser.stringify(restParams)).nodes || [])
+				.map((n) => n.value.replace(/\s/g, ''))
 				.filter((n) => n.length);
 
 			const isDuplicate = media.length


lachieh:
@ybiquitous @jeddy3 is this really as simple of a change that needs to happen here? Is there a test that can cover this?

hudochenkov:
@lachieh you could use an example from the issue and add it to the tests for the rule.

ybiquitous:
@lachieh 

As @hudochenkov suggested, it would be helpful to have a test case. 😃 

In addition, it would be helpful if you would improve our custom type for `postcss-media-query-parser` like this:

```diff
-		nodes: Child[];
+		nodes?: Child[];
```

https://github.com/stylelint/stylelint/blob/585d28a64358a6b7c72d1bed4aaf849caa3bf181/types/postcss-media-query-parser/index.d.ts#L10

jeddy3:
@lachieh Thanks for the pull request. If you haven't already seen it, there [are steps in the CONTRIBUTING guide](https://stylelint.io/CONTRIBUTING#run-tests) to run the test runner on just a particular rule. Once you've added an test case (and made @ybiquitous suggested change), we can trigger the workflow for you.

lachieh:
The definition has been updated and the test case added. Thanks!

lachieh:
All done! Thanks everyone!

jeddy3:
LGTM, thanks!

v14 changelog entry added:

> - Fixed: `media-feature-name-value-allowed-list` TypeError for spaceless condition ([#5581](https://github.com/stylelint/stylelint/pull/5581)).

PR #5582, opened 2021-10-09
Add TypeScript type definitions

### LINKED ISSUE #4399: move to TypeScript 3.7

### LINKED ISSUE #5580: Fix inconsistent types
`@types/stylelint` is poorly synchronized with the `stylelint` package, which is why the VS Code extension uses [a script](https://github.com/stylelint/vscode-stylelint/blob/82a8a90/scripts/stylelint-types.js) to pull down the types from the Stylelint repository (which is hackish in and of itself). However, even these type definitions are structured oddly and have some inconsistencies.

Ideally, if these inconsistencies are resolved, it would be optimal to have the Stylelint package provide its own type definitions in-package using `"types": "types/stylelint/index.d.ts"`, which entirely avoids the problem of keeping in sync with `@types/stylelint` and ensures consistent typings in the Stylelint repository and downstream.

Here are a few of the issues that I've encountered downstream in the VS Code extension:

#### [`test/lib/warnings-to-diagnostics/test.js`](https://github.com/stylelint/vscode-stylelint/blob/b94dca9/test/lib/warnings-to-diagnostics/test.js#L3):

> Also in [`test/workspace/rule-doc/test-plugin.js`](https://github.com/stylelint/vscode-stylelint/blob/b94dca9/test/workspace/rule-doc/test-plugin.js#L4)

```js
const { lint } = /** @type {import('stylelint').StylelintPublicAPI} */ (require('stylelint'));
```

- Required because `lint` is not defined on the `stylelint` module. `const { lint } = require('stylelint');` results in a `Property 'lint' does not exist on type 'typeof import("stylelint")'.ts(2339)` error. Having the private and public API being defined as types on the module instead of a namespace export is atypical.

  Changing this may mean that the same types can be consumed both by Stylelint itself and downstream, unblocking the option of exposing types in `package.json`.
- `StylelintPublicAPI.lint` has the type `Function`, which means that its return value is not typed and is not subject to type-checking without manually typing its return value. But doing that would mean that if its return type changed, the change would go unnoticed by Typescript.

#### [`test/workspace/lint/stylelint.config.js`](https://github.com/stylelint/vscode-stylelint/blob/b94dca9/test/workspace/lint/stylelint.config.js#L5-L22):

```js
// TODO: Workaround for bad typings upstream
/** @type {StylelintConfig & { overrides: StylelintConfig }} */
const config = {
	// ...
	overrides: [
		// ...
		{
			files: ['**/*.scss'],
			// Workaround necessary for the following line
			customSyntax: 'postcss-scss',
		},
	],
};
```

`StylelintConfigOverride` has a very limited number of properties that are allowed, even though the override implementation allows almost every property (as far as I can tell from digging through the code). In this case, `customSyntax` isn't allowed by the types even though it's allowed by Stylelint, hence the workaround. Otherwise the `customSyntax` property results in the error:

```
Type '{ files: string[]; customSyntax: string; }' is not assignable to type 'StylelintConfigOverride'.
  Object literal may only specify known properties, and 'customSyntax' does not exist in type 'StylelintConfigOverride'.ts(2322)
```

Instead of it being defined with `Pick<...>`, it may make more sense to define it with `Omit<...>` and block the properties that aren't supported. This would also mean the type wouldn't need to be manually updated anytime a new configuration property is added.

#### [`test/workspace/rule-doc/test-plugin.js`](https://github.com/stylelint/vscode-stylelint/blob/b94dca9/test/workspace/rule-doc/test-plugin.js#L16-L45):

```js
module.exports = stylelint.createPlugin(
	ruleName,
	// TODO: Workaround for bad typings upstream
	/** @type {import('stylelint').StylelintRule} */ (
		/** @type {unknown} */ (
			(/** @type {any} */ primaryOption) => {
				return (
					/** @type {PostCSSRoot} */ postcssRoot,
					/** @type {PostCSSResult} */ postcssResult,
				) => {
					const validOptions = stylelint.utils.validateOptions(postcssResult, ruleName, {
						actual: primaryOption,
					});

					if (!validOptions) {
						return;
					}

					stylelint.utils.report({
						ruleName,
						result: postcssResult,
						message: messages.expected,
						node: postcssRoot,
						index: 5,
					});
				};
			}
		)
	),
);
```

This workaround is necessary because the `StylelintRule` type expects [specific properties](https://github.com/stylelint/stylelint/blob/6c54b4c/types/stylelint/index.d.ts#L173-L175) added to the function, even though `createPlugin` doesn't require these properties, whether in the implementation or in the documentation. Without this type conversion Typescript throws the error:

```
Argument of type '(primaryOption: any) => (postcssRoot: PostCSSRoot, postcssResult: PostCSSResult) => void' is not assignable to parameter of type 'StylelintRule<any, any>'.
  Type '(primaryOption: any) => (postcssRoot: Root, postcssResult: PostcssResult) => void' is missing the following properties from type '{ ruleName: string; messages: StylelintRuleMessages; primaryOptionArray?: boolean | undefined; }': ruleName, messagests(2345)
```

Perhaps these properties are indeed necessary somewhere else in the codebase that I'm unaware of. In that event, it shouldn't be the exact same type that's used in `createPlugin`. Otherwise, if they are not required anywhere, they should be made optional.

---

Let me know if I've missed any context!

adalinesimonian:
Closes #4399, Fixes #5580

- Types are refactored to be exportable and are now provided in the package:

  - `Stylelint` prefix removed where appropriate, [types namespaced under `stylelint`](https://github.com/stylelint/stylelint/blob/b8ae84b/types/stylelint/index.d.ts#L6).

  - Module export definition changed to [point to namespace and public API](https://github.com/stylelint/stylelint/blob/b8ae84b/types/stylelint/index.d.ts#L382-L384), which allows `require('stylelint')` to be typed correctly.

  - Types renamed to match those in the `@types/stylelint` package to reduce incompatibilities/breakages downstream.

  - As a result of these changes, the separate types package can be deprecated in favour of the built-in type definitions.

- Internal API endpoints marked with `@internal` where appropriate.

- PostCSS types in types definition file scoped under `PostCSS` (e.g. `PostCSS.Root`) to make it clearer which types are imported ([example](https://github.com/stylelint/stylelint/blob/b8ae84b/types/stylelint/index.d.ts#L113-L117)).

- API types (`stylelint.lint`, `stylelint.createPlugin`, etc.) [changed from `Function` to more specific types](https://github.com/stylelint/stylelint/blob/b8ae84b/types/stylelint/index.d.ts#L287-L353) that actually reflect each function's signature.

- Refactored export types in public-facing exports (e.g. [`lib/index.js`](https://github.com/stylelint/stylelint/blob/b8ae84b/lib/index.js#L14), [`lib/createPlugin.js`](https://github.com/stylelint/stylelint/blob/b8ae84b/lib/createPlugin.js#L17)) to reference API types in type definition file. This ensures that Typescript will report errors if the implementation/internal type annotations are changed to an incompatible signature. Any types that were only defined in those files as `@typedef` JSDoc comments have been moved to the `.d.ts` definition file (for example, the [`Violation` type](https://github.com/stylelint/stylelint/blob/b8ae84b/types/stylelint/index.d.ts#L277-L285)).

- Fixed problem with `Rule` types that resulted in Typescript errors when using `createPlugin`. Required properties were added in #5418, which broke usage of `createPlugin` as the `rule` parameter would require functions to have the newly added properties, even though this doesn't reflect the implementation accurately. To fix this:

  - [`Rule` type split into `RuleBase` and `Rule`.](https://github.com/stylelint/stylelint/blob/b8ae84b/types/stylelint/index.d.ts#L150-L162) Added the `Plugin` type as an alias for `RuleBase` to achieve parity with former types at `@types/stylelint`. `RuleBase` represents the rule without any properties attached. `Rule` is a union of `RuleBase` and an object with the properties.

  - [`createPlugin` updated to use `Plugin`.](https://github.com/stylelint/stylelint/blob/b8ae84b/types/stylelint/index.d.ts#L304)

- Fixed Typescript errors being thrown when the configs passed to the `overrides` option would contain supported properties that were missing from the types definition:

  - `ConfigOverride` type [updated to use `Omit` instead of `Pick`](https://github.com/stylelint/stylelint/blob/b8ae84b/types/stylelint/index.d.ts#L23-L25), removing the `overrides` property instead of allow-listing properties. This fix means the `ConfigOverride` type shouldn't need to be manually updated whenever a new configuration property is added.

- `RuleMessages` and related types fixed to avoid Typescript errors. The return type of `stylelint.utils.ruleMessages` was incorrectly set to the type of the `messages` parameter, which resulted in type errors in `lib/ruleMessages.js`. This change:

  - [Changes the `ruleMessages` return type to `{ [K in keyof T]: T[K] }`](https://github.com/stylelint/stylelint/blob/b8ae84b/types/stylelint/index.d.ts#L330-L333), which is correct. There is a distinction between returning the same type as being passed to the function and returning a type with the same keys and value types for those keys.

  - `RuleMessages` type [changed to a mapped type](https://github.com/stylelint/stylelint/blob/b8ae84b/types/stylelint/index.d.ts#L135) to facilitate the change to `stylelint.utils.ruleMessages`. This allows Typescript to narrow types for the object returned by `ruleMessages`:

    ```js
    const messages = stylelint.utils.ruleMessages(ruleName, {
      violation: 'This a rule violation message',
      warning: (reason: string) => `This is not allowed because ${reason}`,
    });
    // If a `Record<...>` was used instead, the following lines would
    // result in Typescript error 2322.
    const violationMessage: string = messages.violation;
    const violationFunc: Function = messages.warning;
    ```

  - Type annotations in `lib/utils/ruleMessages.js` [updated to refect these changes](https://github.com/stylelint/stylelint/blob/b8ae84b/lib/utils/ruleMessages.js#L3) and to avoid Typescript errors and allow type-checking.

- Type test file [added to `types/stylelint/type-test.ts`](https://github.com/stylelint/stylelint/blob/b8ae84b/types/stylelint/type-test.ts). This file contains code that references the public API as a consuming package would use it. It is not executed, but is used so that if the API or types are changed in a breaking way, the change will be caught by Typescript. If this existed before, the breaking change to the `StylelintRule` type (now `stylelint.Rule`) would have been caught.

- Documentation comments added to public API endpoints (`stylelint.*`).

- [Removed leftover duplicate `Options` type](https://github.com/stylelint/stylelint/blob/6c54b4c/types/stylelint/index.d.ts#L147-L159) that had no references anywhere in the project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ All notable changes to this project are documented in this file.
 - Added: `ignoreFunctions: []` to `length-zero-no-unit` ([#5314](https://github.com/stylelint/stylelint/pull/5314)).
 - Added: `ignoreAtRules: []` to `no-invalid-position-at-import` ([#5520](https://github.com/stylelint/stylelint/pull/5520)).
 - Added: `ignoreProperties: []` to `number-max-precision` ([#5421](https://github.com/stylelint/stylelint/pull/5421)).
-- Added: Typescript type definitions ([#5582](https://github.com/stylelint/stylelint/pull/5582)).
+- Added: TypeScript type definitions ([#5582](https://github.com/stylelint/stylelint/pull/5582)).
 - Fixed: "No files matching the pattern" when using backslash paths on Windows ([#5386](https://github.com/stylelint/stylelint/pull/5386)).
 - Fixed: `function-url-quotes` violation messages to be consistent with other `*-quotes` rules ([#5488](https://github.com/stylelint/stylelint/pull/5488)).
 - Fixed: `length-zero-no-unit` false positives for `flex` property ([#5315](https://github.com/stylelint/stylelint/pull/5315)).
--- a/docs/migration-guide/to-14.md
+++ b/docs/migration-guide/to-14.md
@@ -108,10 +108,11 @@ The `function-calc-no-invalid` has be removed. You should remove it from your co
 
 ## Plugin authors
 
-There are two changes that may affect you:
+There are three changes that may affect you:
 
 - version 8 of PostCSS is now used in stylelint
 - a [`disableFix` secondary option](../user-guide/configure.md#disableFix) was added
+- TypeScript type definitions were added to the package
 
 ### PostCSS 8
 
@@ -130,3 +131,7 @@ Even though version 8 of PostCSS is used in stylelint, you can't use the [new Vi
 ### `disableFix` secondary option
 
 We previously suggested plugin authors provide this option. It is now available in Stylelint itself, and you should remove the option from your plugin.
+
+### Built-in TypeScript definitions
+
+The `stylelint` package exports its own TypeScript type definitions now. If you are using the `@types/stylelint` package, you should remove it from your dependencies.


adalinesimonian:
Updated PR comment with links to each relevant change in the code, where appropriate.

PR #5593, opened 2021-10-12
Add `ignorePseudoClasses: []` to `max-nesting-depth`

### LINKED ISSUE #5310: max-nesting-depth - ignore specific pseudo classes

I want to ignore only specific pseudo classes. In Angular its very common to use `:host` pseudo class, so this always takes 1 place in calculating nested depth

``` JS
rules: {
    'max-nesting-depth': [
        1,
        {
            ignoreAtRules: [':host'],
        },
    ],
},
```

lachieh:

Closes #5310

Added rule option according to discussion in the issue thread.


jeddy3:
@lachieh Actually, probably best you wait an hour until the release is finished and the main branch is renamed. In case that causes you more headaches.

PR #5603, opened 2021-10-19
Add support for `extends` in `overrides` config

ota-meshi:

None, but related to #3128.

This PR adds support for `extends` in `overrides` config.

e.g.

```json
{
  "overrides": [
    {
      "files": ["*.scss"],
      "extends": "stylelint-config-recommended-scss"
    }
  ]
}
```

This option allows users to use config by language.


ota-meshi:
I don't think we need to update the document.
If we don't merge this PR and don't support `extends`, I think we need to add it to document as an unsupported property.

PR #5650, opened 2021-10-25
Fix false positives for hex colours in color-function-notation

### LINKED ISSUE #5638: Fix false positives for hex colours in color-function-notation
> Clearly describe the bug

After upgrading to the latest major, I experienced an issue with `color-function-notation`, and turns out it wasn't an issue per-say, but I think `stylelint` could've made a better job at telling me the issue.

So I had this rule prior to the upgrade

```scss
background-color: rgba(#fff, 0.85);
```

After running stylelint, I had the warning I should use `color-function-notation` (nice!). So what I did is the following

```scss
background-color: rgba(#fff / 0.85);
```

After that, stylelint passed. However, it was a false-positive as when I built my SCSS, it actually failed. I discovered that hex colors aren't allowed. The correct syntax is the following

```scss
background-color: rgba(255 255 255 / 0.85);
```

`color-function-notation`

```css
a {
  background-color: rgba(#fff / 0.85);
}
```

Default one or the following

```json
{
  "rules": {
    "color-function-notation": "modern"
  }
}
```

```
"stylelint": "14.0.0",
```

```
npx stylelint
```

I don't think so?

A warning from stylelint directly in my code (with a tooltip)

✅  from stylelint when it actually was invalid code

lachieh:
Resolves #5638

* Added `isStandardSyntaxColorFunction` util and tests
* Updated `color-function-notation` rule to use above function

### DIFF
--- a/lib/utils/isStandardSyntaxColorFunction.js
+++ b/lib/utils/isStandardSyntaxColorFunction.js
@@ -1,17 +1,14 @@
 'use strict';
 
 const isStandardSyntaxFunction = require('./isStandardSyntaxFunction');
-const { isValueFunction } = require('./typeGuards');
 
 /**
  * Check whether a function is standard syntax color function
  *
- * @param {import('postcss-value-parser').Node} node
- * @returns { node is import('postcss-value-parser').FunctionNode }
+ * @param {import('postcss-value-parser').FunctionNode} node
+ * @returns {boolean}
  */
 module.exports = function isStandardSyntaxColorFunction(node) {
-	if (!isValueFunction(node)) return false;
-
 	if (!isStandardSyntaxFunction(node)) return false;
 
 	// scss rgba() function can accept a hex as the first param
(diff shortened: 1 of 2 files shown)


lib/utils/isStandardSyntaxColorFunction.js
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @param {import('postcss-value-parser').FunctionNode} node
+ * @returns {boolean}
+ */
+module.exports = function isStandardSyntaxColorFunction(node) {
+	if (!isStandardSyntaxFunction(node)) return false;
+
+	// scss rgba() function can accept a hex as the first param
+	for (const fnNode of node.nodes) {
+		if (fnNode.type === 'word' && fnNode.value.startsWith('#')) return false;
```
kamechb, line 16: Scss users probably use a color variable like `$black` in the function, then it should still return false.

jeddy3: @kamechb See https://github.com/stylelint/stylelint/issues/5671.

That issue is labelled as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

There are [steps on how to fix a bug in a rule](https://stylelint.io/developer-guide/rules#fix-a-bug-in-a-rule) in the Developer guide.

PR #5657, opened 2021-10-26
Fix TypeError for baseIndentLevel: 1 option and vue file in indentation

wolfgangwalther:
This is a follow up to issue #5289, because PR #5297 did not replace `syntax: 'html'` and `syntax: 'css-in-js'` in rules tests. Those test-cases were disabled instead, and some of those were broken at the same time by accidentally removing some whitespace, where it was required for the tests.

The first and second commits replace `syntax: ...` everywhere and remove the `skip: true` in those cases. The third commit brings back the accidentally removed whitespace. After that, a few tests still fail:
- the indentation and max-empty-lines rules were using `raws.beforeStart` and `raws.afterEnd`. Not sure where those came from (`postcss-syntax` maybe?), but it seems they should be replace with `raws.codeBefore` and `raws.codeAfter`. (fourth commit)
- the declaration-empty-line-before rule seems to need another check for `isRoot(parent)` - at least that fixed the failing test... (fifth commit). There's still some issue with typing, but I don't really know typescript well, yet, so I'd need some guidance there.

I guess those changes should have been part of #5304, but were not noticed because of the skipped tests.

---

I came here, because when migrating to stylelint 14 in a vue project using `postcss-html`, I had the following error thrown:
```
TypeError: Cannot read properties of undefined (reading 'match')
...
at inferRootIndentLevel (node_modules/stylelint/lib/rules/indentation/index.js:682:25)
```

This only happens when setting `baseIndentLevel` to something other than `auto`.

Closes https://github.com/stylelint/stylelint/issues/5677

### DIFF
--- a/lib/rules/declaration-bang-space-before/__tests__/index.js
+++ b/lib/rules/declaration-bang-space-before/__tests__/index.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const postcssSass = require('postcss-sass');
+
 const { messages, ruleName } = require('..');
 
 testRule({
@@ -197,7 +199,7 @@ testRule({
 testRule({
 	ruleName,
 	config: ['always'],
-	syntax: 'sass',
+	customSyntax: postcssSass,
 	skipBasicChecks: true,
 	fix: true,
 
@@ -215,7 +217,7 @@ testRule({
 testRule({
 	ruleName,
 	config: ['never'],
-	syntax: 'sass',
+	customSyntax: postcssSass,
 	skipBasicChecks: true,
 	fix: true,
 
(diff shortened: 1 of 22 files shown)


wolfgangwalther:
I dropped the `postcss-css-in-js` changes (still have them on a different branch, if needed) and moved the change regarding the `declaration-empty-line-before` into #5689.

The remaining bits are just enabling `postcss-html` and fixing #5677. This should be GTM.

jeddy3 (approved):
LGTM, thanks!

wolfgangwalther:
(just a rebase on latest main)

ybiquitous (approved):
@wolfgangwalther Thanks. LGTM 👍🏼

jeddy3:
Changelog entry:

> - Fixed: `indentation` TypeError for `baseIndentLevel: 1` option for Vue files ([#5657](https://github.com/stylelint/stylelint/pull/5657)).

PR #5679, opened 2021-10-31
Add rule-selector-property-disallowed-list

### LINKED ISSUE #5433: Add rule-selector-property-disallowed-list

doing-art:

The new rule rule-selector-property-disallowed-list that is mentioned in #5433 has been added.

@jeddy3 I see that the branch `v14` has been merged and my previous [PR #5489](https://github.com/stylelint/stylelint/pull/5489), targeted to that branch, has been closed. So, I am creating this one to the `main` branch. 

I made [the change](https://github.com/stylelint/stylelint/pull/5489#pullrequestreview-780769606) that you asked me to do in comments to the previous PR. Namely, added a new test which checks multiple errors in one CSS block.

---

Closes #5433

### DIFF
--- a/lib/rules/rule-selector-property-disallowed-list/index.js
+++ b/lib/rules/rule-selector-property-disallowed-list/index.js
@@ -14,7 +14,7 @@ const messages = ruleMessages(ruleName, {
 });
 
 /** @type {import('stylelint').StylelintRule} */
-function rule(primary) {
+const rule = (primary) => {
 	return (root, result) => {
 		const validOptions = validateOptions(result, ruleName, {
 			actual: primary,
@@ -59,7 +59,7 @@ function rule(primary) {
 			}
 		});
 	};
-}
+};
 
 rule.primaryOptionArray = true;
 
--- a/lib/rules/rule-selector-property-disallowed-list/index.js
+++ b/lib/rules/rule-selector-property-disallowed-list/index.js
@@ -1,5 +1,3 @@
-// @ts-nocheck
-
 'use strict';
 
 const isStandardSyntaxRule = require('../../utils/isStandardSyntaxRule');
@@ -15,6 +13,7 @@ const messages = ruleMessages(ruleName, {
 	rejected: (property, selector) => `Unexpected property "${property}" for selector "${selector}"`,
 });
 
+/** @type {import('stylelint').StylelintRule} */
 function rule(primary) {
 	return (root, result) => {
 		const validOptions = validateOptions(result, ruleName, {
--- a/lib/rules/rule-selector-property-disallowed-list/index.js
+++ b/lib/rules/rule-selector-property-disallowed-list/index.js
@@ -13,7 +13,7 @@ const messages = ruleMessages(ruleName, {
 	rejected: (property, selector) => `Unexpected property "${property}" for selector "${selector}"`,
 });
 
-/** @type {import('stylelint').StylelintRule} */
+/** @type {import('stylelint').Rule} */
 const rule = (primary) => {
 	return (root, result) => {
 		const validOptions = validateOptions(result, ruleName, {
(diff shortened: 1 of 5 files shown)


lib/rules/rule-selector-property-disallowed-list/index.js
```diff
@@ -0,0 +1,66 @@
...
+						result,
+						ruleName,
+					});
+				}
+			}
+		});
+	};
+};
+
+rule.primaryOptionArray = true;
```
jeddy3, line 62:
```suggestion
```

We can remove this line as the primary option is an object.

jeddy3 (changes requested):
@doing-art Thanks for creating a new pull request

> my previous PR #5489, targeted to that branch, has been closed

Yep, it was automatically closed by GitHub when `v14` was merged.

I've requested one last changes, otherwise LGTM.

ybiquitous:
@doing-art Thanks for recreating the PR!

The current HEAD revision seems a bit old because the project version specifies `13.13.1` which is not the latest version:

https://github.com/doing-art/stylelint/blob/0fcb254923ee48e14df7efc51fe997c599d1cc74/package.json#L3

The latest version is `14.0.1`:

https://github.com/stylelint/stylelint/blob/85cc19c73bc7751edca9745548d00e01252cd1c7/package.json#L3

Just in case, could you rebase the `main` branch, please?

doing-art:
> @doing-art Thanks for recreating the PR!
> 
> The current HEAD revision seems a bit old because the project version specifies `13.13.1` which is not the latest version:
> 
> https://github.com/doing-art/stylelint/blob/0fcb254923ee48e14df7efc51fe997c599d1cc74/package.json#L3
> 
> The latest version is `14.0.1`:
> 
> https://github.com/stylelint/stylelint/blob/85cc19c73bc7751edca9745548d00e01252cd1c7/package.json#L3
> 
> Just in case, could you rebase the `main` branch, please?

Thank you for your comment. I rebased `main`.

ybiquitous (approved):
@doing-art Thank you. LGTM 👍🏼

jeddy3 (approved):
Thanks!

jeddy3:
Changelog entry added:

> - Added: `rule-selector-property-disallowed-list` rule ([#5679](https://github.com/stylelint/stylelint/pull/5679)).

Mouvedia:
@jeddy3 can we reuse rules internally?
By leveraging this rule we could have a meta-rule that disallow useless properties on an element.
i.e. either 1. the property is invalid (hence ignored) for that element or 2. it's superfluous (it's the default e.g. an anchor is inline)

1. is fairly easy to implement (it's just tedious)
2. is impossible (because the property might be set using JS but reset using CSS)

If that's possible then we could have a new `property-no-invalid` rule for 1. (would be limited to type selectors though).
Id advise to support both blacklisting _and_ whitelisting for the implementation.

jeddy3:
@Mouvedia 

> can we reuse rules internally?

[Plugins can](https://stylelint.io/developer-guide/plugins#stylelintutilscheckagainstrule). 

> If that's possible then we could have a new `property-no-invalid` rule for 1

See [this plugin](https://www.npmjs.com/package/stylelint-declaration-block-no-ignored-properties) - which was removed from core a while back.

Mouvedia:
> See this plugin - which was removed from core a while back.

@jeddy3 that's not what I am talking about.
The reference, in my case, is the element.
i.e. on that element the property does nothing (**irregardless of the other properties**)
It wouldn't pose problems unlike that plugin.

PR #5680, opened 2021-10-31
Fix false positives for values wrapped in parentheses in declaration-empty-line-before

### LINKED ISSUE #5639: Fix false positives for values wrapped in parentheses in declaration-empty-line-before
> Clearly describe the bug

In a SCSS file, if you have a rule whose value is surrounded in `()`, the next line will show as an `declaration-empty-line-before` error.

Works:
image

Does not work:
image

`declaration-empty-line-before`

```css
div {
  height: (100%);
  width: 100%;
}
```

This is a contrived example, but it also repros with more appropriate examples, such as:

```css
div {
  height: ($abc + 50px) * ($def + 80px);
  width: 100%;
}
```

I don't think any specific configuration is needed as this rule is on by default. Here's what I'm using, with the irrelevant rules trimmed out:

```yml
customSyntax: postcss-scss
extends:
- stylelint-config-standard
plugins:
- stylelint-order
- stylelint-scss
rules:
  declaration-empty-line-before: "always"
```

`14.0.0`

CLI: `stylelint "src/**/test.scss"`

Yes, it's related to the capability of using `()` parentheses in SCSS, which can be used for order of operations math with variables.

No warnings to be flagged.

```shell
src/test.scss
 9:2  ✖  Expected empty line before declaration  declaration-empty-line-before
```

----

I did a quick search for this error in previous issues and didn't see any that brought up this particular issue.

It seems to be caused by [`isStandardSyntaxDeclaration:41`](https://github.com/stylelint/stylelint/blob/060310cfd6383175ee06fe4d15aee27a4e202671/lib/utils/isStandardSyntaxDeclaration.js#L41) expressly treating SCSS lines that start in `(` and end in `)` as "not a standard property declaration".

For example, in this file:

image

The `height` property is not treated as a declaration:

image

I'm not sure what the expected use case of `isStandardSyntaxDeclaration:41` is, so I don't really want to mess with it &mdash; however, it seems to be causing these simple math parentheses to be ignored as declarations.

ekumlin:
Closes #5639.
Adds a condition to the SCSS maps & lists check to ensure it's not used in cases like `height: (0);`.

### DIFF
--- a/lib/utils/__tests__/isStandardSyntaxDeclaration.test.js
+++ b/lib/utils/__tests__/isStandardSyntaxDeclaration.test.js
@@ -138,16 +138,29 @@ describe('isStandardSyntaxDeclaration', () => {
 	it('less map', () => {
 		expect(isStandardSyntaxDeclaration(lessDecl('@map: { key: value; }'))).toBe(false);
 	});
+
 	it('scss map declaration', () => {
 		expect(isStandardSyntaxDeclaration(scssDecl('$foo: (key: value, key2: value2)'))).toBe(false);
 	});
+
 	it('scss map declaration with quotes', () => {
 		expect(isStandardSyntaxDeclaration(scssDecl('$foo: ("key": value, "key2": value2)'))).toBe(
 			false,
 		);
 	});
+
 	it('scss list declaration', () => {
 		expect(isStandardSyntaxDeclaration(scssDecl('$foo: (value, value2)'))).toBe(false);
+	});
+
+	it('scss pure value in parentheses', () => {
+		expect(isStandardSyntaxDeclaration(scssDecl('a { height: (0) }'))).toBe(true);
+	});
+
+	it('scss complex value in parentheses', () => {
+		expect(isStandardSyntaxDeclaration(scssDecl('a { height: (3px + 5px) * (2px + 4px) }'))).toBe(
+			true,
+		);
 	});
 });
 
--- a/lib/utils/isStandardSyntaxDeclaration.js
+++ b/lib/utils/isStandardSyntaxDeclaration.js
@@ -18,7 +18,6 @@ function isStandardSyntaxLang(lang) {
 module.exports = function (decl) {
 	const prop = decl.prop;
 	const parent = decl.parent;
-	const value = decl.value;
 
 	// Declarations belong in a declaration block or standard CSS source
 	if (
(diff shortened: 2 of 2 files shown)


lib/utils/__tests__/isStandardSyntaxDeclaration.test.js
```diff
@@ -138,17 +138,30 @@ describe('isStandardSyntaxDeclaration', () => {
...
+
+	it('scss pure value in parentheses', () => {
+		expect(isStandardSyntaxDeclaration(scssDecl('a { height: (0) }'))).toBe(true);
+	});
+
+	it('scss complex value in parentheses', () => {
+		expect(isStandardSyntaxDeclaration(scssDecl('a { height: (3px + 5px) * (2px + 4px) }'))).toBe(
+			true,
+		);
+	});
```
ybiquitous, line 164: @ekumlin Thanks for creating this PR! 😄 

The added test cases do not seem to cover the new logic...? 🤔 

```console
$ npx jest lib/utils/__tests__/isStandardSyntaxDeclaration.test.js --coverage
...

$ open .coverage/lcov-report/lib/utils/isStandardSyntaxDeclaration.js.html
```

image

ekumlin: That's... very strange. The tests succeed, but when I remove the change and run the tests, the tests fail, so I can't really think of why this might be the case. I'll see if I can figure it out.

ekumlin: Actually, on looking at that closer, the new logic *is* tested. What doesn't seem to be tested is the path where prop *does* start with `$`. However, I'd expect that to be covered by this:

image

So will take a look at that.

ekumlin: Oh, I see. Before the change, there is this block:

image

That already checks if the prop starts with `$`. Which means that in any event when the prop is a variable, we already return `false`, so this second block shouldn't even be needed. And indeed, if I remove that block entirely, the tests pass.

ybiquitous (approved):
@ekumlin Thanks, LGTM! 👍🏼

jeddy3 (approved):
Thank you!

jeddy3:
Changelog entry added:

> - Fixed: `declaration-empty-line-before` false positives for values wrapped in parentheses ([#5680](https://github.com/stylelint/stylelint/pull/5680)).

PR #5683, opened 2021-11-01
Fix extending rules within overrides

### LINKED ISSUE #5656: Fix extending rules within overrides
> Clearly describe the bug

When using `extends` within the `overrides` property, the rules don't seem to override.

None.

`input.scss`:

```scss
@mixin {}
```

`.stylelintrc.json`:

```json
{
  "rules": {
    "at-rule-no-unknown": true
  },
  "overrides": [
    {
      "files": "**/*.scss",
      "extends": ["./scss.json"]
    }
  ]
}
```

`scss.json`:

```json
{
  "customSyntax": "postcss-scss",
  "rules": {
    "at-rule-no-unknown": null
  }
}
```

14.0.0

CLI with `npx stylelint input.scss`      

N/A

No warnings to be flagged.

```shell
input.scss
 1:1  ✖  Unexpected unknown at-rule "@mixin"  at-rule-no-unknown
```

---

What do we think should be the intended behaviour here? I was expecting the rule to be overridden for `scss` files, i.e. to have the following resolved config:

```json
{
  "rules": {
    "at-rule-no-unknown": [
      null
    ]
  },
  "customSyntax": "postcss-scss"
}
```

But `npx stylelint input.scss --print-config` gives me:

```json
{
  "rules": {
    "at-rule-no-unknown": [
      true
    ]
  },
  "customSyntax": "postcss-scss"
}
```

I'm running into this issue while looking into https://github.com/twbs/stylelint-config-twbs-bootstrap/issues/128, where I'd like to write the following:

```json
{
  "extends": ["stylelint-config-standard"],
  "overrides": [
    {
      "files": "**/*.scss",
      "extends": ["stylelint-config-standard-scss"]
    }
  ]
}
```

At-mentioning @ota-meshi because of https://github.com/stylelint/stylelint/pull/5603.

ota-meshi:

Closes #5656

No, it's self-explanatory.

### DIFF
--- a/lib/__tests__/fixtures/config-overrides/at-rule-no-unknown.json
+++ b/lib/__tests__/fixtures/config-overrides/at-rule-no-unknown.json
@@ -1,0 +1,5 @@
+{
+	"rules": {
+		"at-rule-no-unknown": true
+	}
+}
--- a/lib/__tests__/fixtures/config-overrides/indent4.json
+++ b/lib/__tests__/fixtures/config-overrides/indent4.json
@@ -1,0 +1,5 @@
+{
+	"rules": {
+		"indentation": 4
+	}
+}
--- a/lib/__tests__/fixtures/config-overrides/indent8.json
+++ b/lib/__tests__/fixtures/config-overrides/indent8.json
@@ -1,0 +1,5 @@
+{
+	"rules": {
+		"indentation": 8
+	}
+}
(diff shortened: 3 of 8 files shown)


jeddy3:
The behaviour in the tests LGTM.

ybiquitous (approved):
@ota-meshi Thanks, looks good! 👍🏼

jeddy3:
> - Fixed: extending rules within overrides ([#5683](https://github.com/stylelint/stylelint/pull/5683)).

(I've not had time for Stylelint recently, but going to dip into a few pull requests now).

PR #5696, opened 2021-11-06
Add `ConfigurationError` type

adalinesimonian:
Adds the ConfigurationError type. Currently, we [define this type downstream](https://github.com/stylelint/vscode-stylelint/blob/77c0719/types/index.d.ts#L68-L73) since it isn't defined in Stylelint, but it really belongs here.

### DIFF
--- a/lib/utils/configurationError.js
+++ b/lib/utils/configurationError.js
@@ -1,13 +1,14 @@
 'use strict';
 
+/** @typedef {import('stylelint').ConfigurationError} ConfigurationError */
+
 /**
  * Create configurationError from text and set CLI exit code
  * @param {string} text
- * @returns {Object}
+ * @returns {ConfigurationError}
  */
-module.exports = function (text) /* Object */ {
-	/** @type {Error & {code?: number}} */
-	const err = new Error(text);
+module.exports = function (text) {
+	const err = /** @type {ConfigurationError} */ (new Error(text));
 
 	err.code = 78;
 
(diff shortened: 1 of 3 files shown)


lib/utils/configurationError.js
```diff
@@ -1,13 +1,14 @@
...
  * Create configurationError from text and set CLI exit code
  * @param {string} text
- * @returns {Object}
+ * @returns {ConfigurationError}
  */
-module.exports = function (text) /* Object */ {
-	/** @type {Error & {code?: number}} */
-	const err = new Error(text);
+module.exports = function (text) {
+	const err = /** @type {ConfigurationError} */ (new Error(text));
```
ybiquitous, line 11: [trivial] IMO, separating the lines is a bit readable. But you can keep it as-is since it is just my preference. 😃 

```suggestion
	/** @type {ConfigurationError} */
	const err = new Error(text);
```

ybiquitous (approved):
Looks good! 👍🏼

lib/utils/configurationError.js
adalinesimonian: Placing the comment behind the declaration is equivalent to:

```ts
const err: ConfigurationError = new Error(text);
```

This won't work because `code` isn't defined on `Error`. Hence why I changed it to what it is now, which is the equivalent of the typecast:

```ts
const err = new Error(text) as ConfigurationError;
```

ybiquitous: Ah, I see why. Thanks for the help! 😊

jeddy3 (approved):
Thanks, LGTM.

Can you resolve the conflict please, then I'll merge?

(We typically don't include adding CHANGELOG entries in the pull request themselves (to avoid these conflicts). It's normally done afterwards in a commit by whoever merges.)

jeddy3:
Nevermind, did it in GitHub... 😄 will merge on green.

adalinesimonian:
> Nevermind, did it in GitHub... 😄 will merge on green.

Thanks!

> (We typically don't include adding CHANGELOG entries in the pull request themselves (to avoid these conflicts). It's normally done afterwards in a commit by whoever merges.)

Noted!

PR #5699, opened 2021-11-07
Pre-work to update to ESLint 8

mattxwang:

Not an issue, but part of the discussion in #5645.

This PR slightly took on scope creep (from the initial ask to just update `eslint-shared-config` to major version 15. Here's my thought process:

1. First, I bumped the above
2. Then, this causes an error with `eslint-plugin-jest`. I could resolve this by manually reinstalling it as a dev dependency. **I'm not sure if this is the right move**; if it isn't, let me know and I can move this commit out of the PR.
3. Finally, I address the newly enabled [`jest/prefer-to-be`](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-to-be.md) rule, which was a simple find-and-replace.

Let me know if we'd prefer dependabot to handle. I'm using node 16 to generate the new `package-lock`, and we stay on lockfile v2.

### DIFF
--- a/lib/__tests__/cli.test.js
+++ b/lib/__tests__/cli.test.js
@@ -290,7 +290,7 @@ describe('CLI', () => {
 			fixturesPath('empty-block-with-disables.css'),
 		]);
 
-		expect(process.exitCode).toEqual(1);
+		expect(process.exitCode).toBe(1);
 
 		expect(process.stderr.write).toHaveBeenCalledTimes(1);
 		expect(process.stderr.write).toHaveBeenCalledWith(
@@ -307,7 +307,7 @@ describe('CLI', () => {
 			fixturesPath('empty-block.css'),
 		]);
 
-		expect(process.exitCode).toEqual(2);
+		expect(process.exitCode).toBe(2);
 
 		expect(process.stdout.write).toHaveBeenCalledTimes(1);
 		const output = JSON.parse(process.stdout.write.mock.calls[0][0]);
@@ -329,7 +329,7 @@ describe('CLI', () => {
 			fixturesPath('empty-block.css'),
 		]);
 
-		expect(process.exitCode).toEqual(2);
+		expect(process.exitCode).toBe(2);
 
 		expect(process.stdout.write).toHaveBeenCalledTimes(0);
 	});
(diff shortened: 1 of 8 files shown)


ybiquitous:
@mattxwang Thank you for creating this PR! 😄 

I try removing `eslint-plugin-jest`, but it also works:

```console
$ npm rm eslint-plugin-jest
...

$ npm ls eslint-plugin-jest
stylelint@14.0.1 /Users/koba/git/stylelint/stylelint
└─┬ eslint-config-stylelint@15.0.0
  └── eslint-plugin-jest@25.2.3

$ npm run lint:js -- --no-cache

> stylelint@14.0.1 lint:js
> eslint . --cache --max-warnings=0 "--no-cache"
```

ntwb:
> Then, this causes an error with eslint-plugin-jest. I could resolve this by manually reinstalling it as a dev dependency. I'm not sure if this is the right move; if it isn't, let me know and I can move this commit out of the PR.

Yeah, I had the same issue locally it worked without this, installing resolved it for me also /shrug 
> I try removing eslint-plugin-jest, but it also works:
@ybiquitous Depenadbot created another PR ~8 hours ago, same issue in that PR:
- https://github.com/stylelint/stylelint/pull/5702
- https://github.com/stylelint/stylelint/runs/4133769681?check_suite_focus=true
```shell
TypeError: Failed to load plugin 'jest' declared in 'package.json » eslint-config-stylelint': Class extends value undefined is not a constructor or null
Referenced from: /home/runner/work/stylelint/stylelint/node_modules/eslint-config-stylelint/index.js
```

ntwb (approved):
Per my comment https://github.com/stylelint/stylelint/pull/5699#issuecomment-962966853 I'm happy to approve this and land this as-is, we can revisit removing the `eslint-plugin-jest` plugin later

ybiquitous (approved):
@ntwb Ah, I see. Thanks for the feedback! 😄 

LGTM 👍🏼

mattxwang:
Thanks everyone! I can also try to see if removing `eslint-plugin-jest` resolves this in the future, maybe this is a node version thing and/or something resolved with ESLint 8?

PR #5707, opened 2021-11-08
updates globby to current version

damky:
updates globby which uses glob-parent 6.0.1 which fixes an Improper Regular Expression DoS vulnerability


ybiquitous:
Unfortunately, we cannot bump up `globby` since its latest version supports only ESM.
See https://github.com/stylelint/stylelint/issues/5291#issuecomment-917929184

Maybe, we should bump up only `glob-parent`.

PR #5689, opened 2021-11-02
Fix support for HTML files in declaration-empty-line-before

wolfgangwalther:
This is taken out of #5657 to be able to merge the agreed-on parts ahead of other stuff.

The first 3 commits are from that yet unmerged PR right now, the change is only in the 4th commit. After re-enabling those tests by replacing the `syntax: 'html'` with `customSyntax`, those two tests would still fail and were left skipped in #5657.

The fix seems simple enough, but needs some improvement regarding typescript typing.

PR #5708, opened 2021-11-09
Refactor to use PostCSS `Container` type for more flexibility

ybiquitous:

Fixes the type error on #5689

Instead of `import('postcss').Rule | import('postcss').AtRule`, this refactoring uses the `import('postcss').Container` type.
This type includes also `import('postcss').Root` so has more flexibility. 

See the type definitions of PostCSS below:

- `Container` - https://github.com/postcss/postcss/blob/77c2e73/lib/container.d.ts#L23-L30
- `AtRule` - https://github.com/postcss/postcss/blob/77c2e73/lib/at-rule.d.ts#L73
- `Rule` - https://github.com/postcss/postcss/blob/77c2e73/lib/rule.d.ts#L66
- `Root` - https://github.com/postcss/postcss/blob/77c2e73/lib/root.d.ts#L51

### DIFF
--- a/lib/utils/beforeBlockString.js
+++ b/lib/utils/beforeBlockString.js
@@ -1,10 +1,9 @@
 'use strict';
 
-/** @typedef {import('postcss').Rule} Rule */
-/** @typedef {import('postcss').AtRule} AtRule */
+const { isAtRule, isRule } = require('./typeGuards');
 
 /**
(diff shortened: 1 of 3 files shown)


jeddy3 (approved):
Cleaner and more flexible, thanks!

PR #5709, opened 2021-11-09
Fix all files ignored prompting "No files matching the pattern xx were found."

### LINKED ISSUE #4712: Fix ignore files must add —allow-empty-input
I hava a `.stylelintignore` to ignore the `var.less`
```
src/packages/theme-default/var.less
```
and I set the pre-commit git hook at `package.json`
```
  "gitHooks": {
    "pre-commit": "lint-staged"
  },
  "lint-staged": {
    "*.{js,vue}": [
      "vue-cli-service lint",
      "git add"
    ],
    "*.{less,vue}": [
      "stylelint lint",
      "git add"
    ]
  }
```
when I want to  **ONLY** commit the change at `var.less`, it comes to 
```
Error: No files matching the pattern "lint, /MyPath/src/packages/theme-default/var.less" were found.
```
It's that anything wrong with my config? OR it's a bug?
Cause I thought that if I ignore a file, when I commit it, it should skip the step for stylelint.

zuoxiaobai:
For old projects, this error is a puzzling error, try to fix #4712 

.stylelintignore
```
src/test.css
```
npx stylelint src/test.css

image

Solution:
1. Adding ignoreFiles to your .stylelintrc, and remove from your .stylelintignore 
```js
{
  "ignoreFiles": [
    "src/test.css"
  ]
}
```
2. stylelint --allow-empty-input
3. Merge this PR, not throw error when stylelint .stylelintignore files


jeddy3 (changes requested):
Thanks for the pull request.

Let's add a test for this.

zuoxiaobai:
@jeddy3 done, npm run test PASS

ybiquitous:
@zuoxiaobai Thanks for creating the PR!

To be honest, I'm afraid that this change may confuse users. 🤔 

Assume that there are 3 files below:
- `.stylelintignore`
- `.stylelintrc.yml`
- `a.css`

```ignore
# .stylelintignore
*.css
```

```yaml
# .stylelintrc.yml
rules:
  block-no-empty: true
```

```css
/* a.css has a `block-no-empty` problem. */
a {}
```

Run `stylelint a.css`. The command silently exits with no errors or warnings; it is just ignored.

```console
$ stylelint a.css
```

Run the same command changing an ignore file:

```console
$ stylelint a.css --ignore-path .gitignore

a.css
 2:3  ✖  Unexpected empty block  block-no-empty
```

If `.stylelignore` has long lines, users may forget that `*.css` is ignored and mistakenly assume there are no errors.

ybiquitous:
The ESLint behavior might be helpful. ESLint emits a warning instead of exiting silently.

See below:

```yaml
# .eslintrc.yml
ignorePatterns: ["*.js"]
rules: {}
```

```console
$ eslint a.js

/tmp/eslint-example/a.js
  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override

✖ 1 problem (0 errors, 1 warning)
```

zuoxiaobai:
@ybiquitous Thanks for your help. 

It's a bit difficult for me to add warning function to all ignored files like ESLint. So I changed my mind. 

Since this situation will only happen when all input files are ignored, it is relatively rare. I tried to make the smallest changes. When this happens, give the correct error message: `File "${pattern}" ignored because of a matching ignore pattern. Move ignore config to .stylelintrc or use "--allow-empty-input" to override.`

image

jeddy3:
@ybiquitous Does the behaviour now align with your suggestion in https://github.com/stylelint/stylelint/pull/5709#issuecomment-1001620043?

ybiquitous:
@zuoxiaobai
cc @jeddy3 

Sorry for the late response. The new solution looks good to me. 👍🏼

lib/utils/allFilesIgnoredError.js
```diff
@@ -0,0 +1,16 @@
...
+class AllFilesIgnoredError extends Error {
+	/**
+	 * @param {string[]} fileList
+	 */
+	constructor(fileList) {
+		super();
+
+		const pattern = fileList.filter((i) => !i.startsWith('!')).join(', ');
+
+		this.message = `File "${pattern}" ignored because of a matching ignore pattern. Move ignore config to .stylelintrc or use "--allow-empty-input" to override.`;
```
ybiquitous, line 12: [suggest] It may be better to wrap each file with double quotes. What do you think?

```suggestion
		const pattern = fileList.filter((i) => !i.startsWith('!')).map((i) => `"${i}"`).join(', ');

		this.message = `File ${pattern} ignored because of a matching ignore pattern. Move ignore config to .stylelintrc or use "--allow-empty-input" to override.`;
```

jeddy3: @ybiquitous Thanks for confirming the pull request is heading in the right direction.

Am I right in thinking that this error will only show if all the files matching the input are ignored?

And, if the input glob touches a lot of files, the `fileList` could be long?

Shall we make the error message simple, e.g. "All input files were ignored because of the ignore pattern. Either change your input, ignore pattern or use "--allow-empty-input" to allow no inputs." Rather than have the possibility of outputting a long list of files.

ybiquitous: @jeddy3 That's a good idea! 👍🏼

jeddy3 (approved):
LGTM, thanks!

ybiquitous (approved):
LGTM 👍🏼

jeddy3:
> - Fixed: throwing more informative error when all input files are ignored ([#5709](https://github.com/stylelint/stylelint/pull/5709)).

PR #5718, opened 2021-11-14
Fix false positives for hex with alpha-channel and false negatives for modern syntax in `color-named`

### LINKED ISSUE #5716: Fix false positives for `#RRGGBBAA` in `color-named`

```css
.a { color: #fff0ffff; }
.b { color: azure; }
```

[DEMO](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0BDABMNkA2EBO8aAxAGZkAMZZA3GgL4A6AdkgEYZYS4FooBeAVzwxadEABoQAZwCeTAC4oAHgnBSpE8BCYkAlgHNVwZmjQMQeAdhhTzhY01OnzOfAFomKALYwod8yjYAO4oMlJuQQAWMMJuAA4Q6rqs1uYm9MxidEA)

```json
{
  "rules": {
    "color-named":"always-where-possible"
  }
}
```

cli

14.0.1

No problems to be reported.

### What actually happened?

The `color-named` rule warns you to replace `#fff0ffff` with `azure`, but they are completely different colors.

image

```
Expected "#fff0ffff" to be "azure" (color-named)
```

_No response_

### Proposal to fix the bug

_No response_

ota-meshi:

Closes #5716

This PR fixes false positives for alpha-channel posted in #5716, and false negatives for modern syntax (e.g. `rgba(0 0 0 / 1)`).


package.json
```diff
@@ -109,6 +109,7 @@
   },
   "dependencies": {
     "balanced-match": "^2.0.0",
+    "colord": "^2.9.1",
```
ota-meshi, line 112: Changed to use colord package instead of reference file.

lib/rules/color-named/colord-utils.js
```diff
@@ -0,0 +1,75 @@
...
+	/* Syntaxes that is removed in Color Module Level 4 specification. */
+
+	// hwb() with comma
+	(_colordClass, parsers) => {
+		parsers.string.push([parseHwbWithCommaString, /** @type {any} */ ('hwb-with-comma')]);
+	},
+	// gray()
+	(_colordClass, parsers) => {
+		parsers.string.push([parseGrayString, /** @type {any} */ ('gray')]);
+	},
```
ota-meshi, line 24: `colord` doesn't understand the old syntax, so I extended it.

package.json
ybiquitous: The [`colord`](https://github.com/omgovich/colord) package looks nice to me. Just in case, could you tell us why you choose this package, please?

ota-meshi: [colord](https://www.npmjs.com/package/colord) seems to be used in [cssnano](https://www.npmjs.com/package/cssnano) and I think we can trust it. Also, another well-known package, [color](https://www.npmjs.com/package/color), doesn't seem to be able to parse modern syntax (e.g. `rgb(0 0 0)`), so I think colord is better.

Please let me know if you know any other better packages.

lib/rules/color-named/__tests__/colord-utils.test.js
```diff
@@ -0,0 +1,177 @@
...
+		},
+	];
+
+	for (const { input, output, valid } of TESTS) {
+		it(`input: ${input}`, () => {
+			const result = colord(input);
+
+			expect(result.isValid()).toBe(valid == null ? true : valid);
+
+			expect(result.toHex()).toEqual(output || '#000000');
```
ybiquitous, line 143: [suggestion] How about the following refactoring for a bit more readablity?

```suggestion
	for (const { input, output = '#000000', valid = true } of TESTS) {
		it(`input: ${input}`, () => {
			const result = colord(input);

			expect(result.isValid()).toBe(valid);

			expect(result.toHex()).toEqual(output);
```

lib/rules/color-named/__tests__/colord-utils.test.js
```diff
@@ -0,0 +1,177 @@
...
```
ybiquitous, line 174: [suggestion] Same here:

```suggestion
	for (const { input, output = '#000000', valid = true } of TESTS) {
		it(`input: ${input}`, () => {
			const result = colord(input);

			expect(result.isValid()).toBe(valid);

			expect(result.toHex()).toEqual(output);
```

lib/rules/color-named/colord-utils.js
```diff
@@ -0,0 +1,75 @@
+const { colord, extend } = require('colord');
```
ybiquitous, line 1: [trivial] For consistency in the project, the file name `colordUtils.js` may be better than `colord-utils.js`.

lib/rules/color-named/colord-utils.js
```diff
@@ -0,0 +1,75 @@
...
+
+module.exports = {
+	colord,
+};
+
+/**
+ * Parses a valid hwb with comma CSS color function
+ * https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hwb()#syntax
+ * @param {string} input
+ * @returns { { r: number; g: number; b: number; a: number } | null }
```
ybiquitous, line 35: [suggestion] How about using the type provided by `colord`?

```suggestion
 * @type {import('colord/types').ParseFunction<string>}
```

lib/rules/color-named/colord-utils.js
```diff
@@ -0,0 +1,75 @@
...
+
+	const [, hue, whiteness, blackness, alpha] = match;
+
+	return colord(`hwb(${hue} ${whiteness} ${blackness}${alpha ? ` / ${alpha}` : ''})`).rgba;
+}
+
+/**
+ * Parses a valid gray() CSS color function
+ * @param {string} input
+ * @returns { { r: number; g: number; b: number; a: number } | null }
```
ybiquitous, line 55: [suggestion] Same here:

```suggestion
 * @type {import('colord/types').ParseFunction<string>}
```

lib/rules/color-named/colord-utils.js
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Parses a valid hwb with comma CSS color function
+ * https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hwb()#syntax
+ * @param {string} input
+ * @returns { { r: number; g: number; b: number; a: number } | null }
+ */
+function parseHwbWithCommaString(input) {
+	const match =
+		/^hwb\(\s*([+-]?(?:\d+(?:\.\d+)?|\.\d+)(?:deg|grad|rad|turn)?)\s*,\s*([+-]?(?:\d+(?:\.\d+)?|\.\d+)%)\s*,\s*([+-]?(?:\d+(?:\.\d+)?|\.\d+)%)\s*(?:,\s*([+-]?(?:\d+(?:\.\d+)?|\.\d+)%?)\s*)?\)$/iu.exec(
```
ybiquitous, line 39: [question] The regex seems too hard to me, but is there a way to avoid such a complex regex in this case? 😓 
(Also, I'm a bit worried about ReDOS...)

lib/rules/color-named/index.js
```diff
@@ -96,7 +79,13 @@ const rule = (primary, secondaryOptions) => {
 				}
 
 				// Check for named colors for "never" option
-				if (primary === 'never' && type === 'word' && namedColors.includes(value.toLowerCase())) {
+				if (
+					primary === 'never' &&
+					type === 'word' &&
+					/^[a-z]+$/iu.test(value) &&
+					value.toLowerCase() !== 'transparent' &&
+					colord(value).isValid()
```
ybiquitous, line 87: [question] Is this `/^[a-z]+$/iu` check (lower-case check) necessary, that is, is it enough to use just `colord(value).isValid()`?

```suggestion
					value.toLowerCase() !== 'transparent' &&
					colord(value).isValid()
```

ybiquitous (commented):
@ota-meshi Thank you for the great fix PR! 👍🏼 

I've left some questions and refactoring suggestions, but this PR looks mostly fine to me! 👏🏼

lib/rules/color-named/index.js
ota-meshi: I think we need to check if it is defined in the alphabet to avoid hex (e.g. `#fff`).

lib/rules/color-named/colord-utils.js
ota-meshi: I will try it :+1:

lib/rules/color-named/index.js
ota-meshi: > `/^[a-z]+$/iu` check (lower-case check)

It is not a lower-case check because it has an `i` flag.

ota-meshi:
@ybiquitous I have added your requested changes to this PR. So could you check it again?

lib/rules/color-named/index.js
ybiquitous: Thanks, I see. 👍🏼

ybiquitous (approved):
@ota-meshi Great work! LGTM 👏🏼

jeddy3 (commented):
LGTM.

This rule was problematically slow in the past.

@ota-meshi It'd be good to get a [benchmark](https://stylelint.io/developer-guide/rules#improve-the-performance-of-a-rule) of this branch against `main` before merging.

ota-meshi:
I didn't do that intentionally, but it also seems to be of value in terms of performance.

After:

```
npm run benchmark-rule -- color-named never 

Warnings: 0
Mean: 37.03908788571429 ms
Deviation: 20.985396496488296 ms

npm run benchmark-rule -- color-named always-where-possible

Warnings: 131
Mean: 41.65396371538462 ms
Deviation: 25.849404139328065 ms
```

Before:

```
npm run benchmark-rule -- color-named never 

Warnings: 0
Mean: 57.58543509615383 ms
Deviation: 40.58585746807126 ms

npm run benchmark-rule -- color-named always-where-possible

Warnings: 129
Mean: 69.16574440277776 ms
Deviation: 25.362776611378344 ms
```

ota-meshi:
The result of `npm run benchmark-rule --color-named always-where-possible` is different, so I checked the cause and noticed that the new logic was converted to `transparent`. I will fix it.

ota-meshi:
I fixed this PR. So please check it again.
https://github.com/stylelint/stylelint/pull/5718/commits/45371c0cb2b800dca71864feae95907683351083

jeddy3 (approved):
Excellent, thank you!

jeddy3:
> - Fixed: `color-named` false positives for hex with alpha-channel and false negatives for modern syntax ([#5718](https://github.com/stylelint/stylelint/pull/5718)).

omgovich:
Hey guys! Thanks for choosing `colord`! Let me know if you will need any help!

Mouvedia:
After reading https://github.com/postcss/postcss-color-gray, I am wondering if `gray()` was actually implemented by one browser.

Issue #5694, opened 2021-11-05
Add end positions to warnings


ChillyBots:
Avoid errors:

- [x] at-rule-no-unknown
- [x] block-no-empty
- [x] color-no-invalid-hex
- [x] comment-no-empty
- [x] custom-property-no-missing-var-function
- [x] declaration-block-no-duplicate-custom-properties
- [x] declaration-block-no-duplicate-properties
- [x] declaration-block-no-shorthand-property-overrides
- [x] font-family-no-duplicate-names
- [x] font-family-no-missing-generic-family-keyword
- [x] function-calc-no-unspaced-operator
- [x] function-linear-gradient-no-nonstandard-direction
- [x] function-no-unknown
- [x] keyframe-declaration-no-important
- [x] media-feature-name-no-unknown
- [x] named-grid-areas-no-invalid
- [x] no-descending-specificity
- [x] no-duplicate-at-import-rules
- [x] no-duplicate-selectors
- [x] no-empty-source
- [x] no-invalid-double-slash-comments
- [x] no-invalid-position-at-import-rule
- [x] property-no-unknown
- [x] selector-pseudo-class-no-unknown
- [x] selector-pseudo-element-no-unknown
- [x] selector-type-no-unknown
- [x] string-no-newline
- [x] unit-no-unknown

Enforce conventions:

- [x] alpha-value-notation
- [x] at-rule-allowed-list
- [x] at-rule-disallowed-list
- [x] at-rule-no-vendor-prefix
- [x] at-rule-property-required-list
- [x] color-function-notation
- [x] color-hex-alpha
- [x] color-hex-length
- [x] color-named
- [x] color-no-hex
- [x] comment-pattern
- [x] comment-word-disallowed-list
- [x] custom-media-pattern
- [x] custom-property-pattern
- [x] declaration-block-no-redundant-longhand-properties
- [x] declaration-block-single-line-max-declarations
- [x] declaration-no-important
- [x] declaration-property-unit-allowed-list
- [x] declaration-property-unit-disallowed-list
- [x] declaration-property-value-allowed-list
- [x] declaration-property-value-disallowed-list
- [x] font-family-name-quotes
- [x] font-weight-notation
- [x] function-allowed-list
- [x] function-disallowed-list
- [x] function-url-no-scheme-relative
- [x] function-url-quotes
- [x] function-url-scheme-allowed-list
- [x] function-url-scheme-disallowed-list
- [x] hue-degree-notation
[...]

PR #5725, opened 2021-11-19
Add ranges for warnings

adalinesimonian:
Partially addresses #5694

Adds range support for warnings.

**Work in progress, not ready for merge.**

Tasks:

- [x] Update PostCSS to 8.4
- [x] Update postcss-value-parser to 4.2
- [x] Update `Warning` type
- [x] Update `Problem` type
- [x] Update `report` function
- [x] Update some rules
- [x] Update tests for these rules


adalinesimonian:
What this generally looks like in VS Code with the appropriate changes downstream:

Screenshot of VS Code reporting Stylelint warnings with ranges

jeddy3:
You're doing amazing work extending the capabilities of both PostCSS and Stylelint with this range stuff!

jeddy3:
(Remove the label to be consistent with [our](https://stylelint.io/maintainer-guide/pull-requests) [usage](https://stylelint.io/maintainer-guide/issues#labels), i.e. only `pr: *` labels for pull requests.)

adalinesimonian:
I was running into issues trying to determine the end position for functions and other nodes provided by postcss-value-parser. I've opened up a PR there that will resolve this problem: TrySound/postcss-value-parser#83

adalinesimonian:
Just an update: I think I'm somewhere halfway through all of the rules. It's mind-numbing tedium so I'm slowing down a bit since I started on this. But I'm still seeing it through.

The issue I'm thinking of is that it is going to be hard to review this PR since it will change almost every rule's implementation. While most rules may only have a couple of changed lines, some have very thoroughly reworked internals.

My setup for working on this PR has been to have the VS Code extension running in a workspace where I enable different rules as I work on them and test what ranges are reported to the editor. This makes it way easier to test. I was thinking about maybe creating a repo that has a folder for each rule and its different supported configuration options. This way someone could pull it down, and using a modified version of the extension, test the reported ranges for each rule.

Would this be helpful? My only hesitation is that it would take a bit of tedious work to create this thing. But then again, this whole PR is a bit of a pain in the butt so what's a little bit more if it saves people some time and hassle?

ybiquitous:
@adalinesimonian Thank you so much for this big change! Your suggestion seems interesting, but I'm not sure how the maintenance cost of such repositories will be.

adalinesimonian:
> @adalinesimonian Thank you so much for this big change! Your suggestion seems interesting, but I'm not sure how the maintenance cost of such repositories will be.

It'd be temporary, not a long-term thing. If someone wanted to keep it updated, they could, but I'm mostly focused on trying to get this PR reviewed thoroughly.

adalinesimonian:
[stylelint-rules-test](https://github.com/adalinesimonian/stylelint-rules-test) is now up, from which you can inspect how errors are reported for rules easily.

Currently, only a few rules are up, more are being added. Instructions on how to use the repository are provided in the [readme](https://github.com/adalinesimonian/stylelint-rules-test#readme).

adalinesimonian:
I'm feeling a little unwell, so I'll be away from this PR for a little bit. I hope to get back to it shortly!

ybiquitous:
@adalinesimonian Please take care of yourself!

Just a question. Can we extract only the update of PostCSS into another PR?

adalinesimonian:
> @adalinesimonian Please take care of yourself!
> 
> Just a question. Can we extract only the update of PostCSS into another PR?

Feel free!

jeddy3:
@adalinesimonian I hope you feel better soon.

I'm just looping back to this pull request now - I've had little time for OSS these past few weeks, but I'll have time in the new year (after going offline for 10 days of Xmas). I'll be able to help more with this pull request then (updating rules, testing, reviewing and whatnot).

Thanks again for working on this range stuff, by the way! It vastly improves the experience for the user in the editor.

Am I right in thinking that this change is non-breaking and that each rule can opt-in to the range approach? If that's the case, could we break up the work of migrating the rules over, i.e. we create an umbrella task-list issue that we can pick away at? We could then incrementally improve the quality of our reporting over time.

package.json
```diff
@@ -132,12 +133,12 @@
     "normalize-path": "^3.0.0",
     "normalize-selector": "^0.2.0",
     "picocolors": "^1.0.0",
-    "postcss": "^8.3.11",
+    "postcss": "^8.4.0",
```
ybiquitous, line 136: I've updated `postcss` via PR #5778. Could you rebase the main branch?

ybiquitous:
I agree with @jeddy3 suggestion. I'd like to make progress on this wonderful feature if I can help. 💪🏼

jeddy3:
> I'd like to make progress on this wonderful feature if I can help

Fantastic. 

I suggest we:

1. Get this pull request ready to merge
2. Systematically move through the remaining rules

For the latter, I've converted https://github.com/stylelint/stylelint/issues/5694 into an umbrella issue and added a task list. I suggest we don't update any more stylistic rules.

For the former, I suggest:

1. We resolve the conflicts
2. Update the tests for the rules already adapted in this pull request.

We need to be pragmatic about this. It's not practical to add `endLine` and `endColumn` properties to every reject test as we've thousands. I suggest adding them to just a few choice tests. I also suggest we revert any changes to the stylistic rules in the pull request if they require a lot more work to finish.

We can always patch mistakes if we get the positions wrong for a rule. We don't have the time or resources to get it perfect first time around. 

What do you think about this approach?

ybiquitous:
@jeddy3 Your suggestion looks so nice. 👍🏼

ybiquitous:
I'm going to fix the failed tests. 👀

lib/rules/declaration-no-important/index.js
```diff
@@ -28,10 +28,12 @@ const rule = (primary) => {
 				return;
 			}
 
+			const importantMatch = decl.toString().match(/!\s*important/);
+
 			report({
 				message: messages.rejected,
 				node: decl,
-				word: 'important',
+				word: importantMatch ? importantMatch[0] : 'important',
```
ybiquitous, line 36: [note] It may be better to revert the change for `declaration-no-important` and fix it later (via another pull request).

ybiquitous:
@jeddy3 Does it make sense to mark stylistic rules *deprecated* (adding `meta.deprecated = true`) before advancing this pull request? I want to revert the changes in stylistic rules to reduce the diff.

ybiquitous:
I see. Thanks for the answer. 👍🏼

ybiquitous:
Reverted frozen stylistic rules via faeb583.

lib/rules/declaration-no-important/index.js
ybiquitous: I've fixed the `!important` position via 85540e6.

lib/utils/getImportantPosition.js
```diff
@@ -0,0 +1,15 @@
...
+ * @returns {Position | undefined}
+ */
+module.exports = function getImportantPosition(source) {
+	const pattern = /!\s*important\b/gi;
+	const match = pattern.exec(source);
+
+	if (!match) return;
+
+	return { index: match.index, endIndex: pattern.lastIndex };
+};
```
ybiquitous, line 15: [note] I cannot find how to get a position of `!important` from CSS source code via `postcss` or `postcss-value-parser`. So, I've written the tiny utility `getImportantPosition`.

If anyone knows a better way, please let me know.

lib/rules/function-calc-no-unspaced-operator/index.js
```diff
@@ -68,7 +70,10 @@ const rule = (primary, _secondaryOptions, context) => {
 									return true;
 								}
 
-								complain(messages.expectedOperatorBeforeSign(operator), decl, operatorSourceIndex);
+								const endIndex = valueIndex + operatorSourceEndIndex;
+								const index = valueIndex + currentNode.sourceIndex;
+
+								complain(messages.expectedOperatorBeforeSign(operator), decl, index, endIndex);
```
ybiquitous, line 76: [ask] Honestly, I doubt if this change is valuable. The previous report seems natural to me. Any thoughts?

(Screenshots on VS Code ⬇️ )

### Before

image

### After

image

jeddy3: Before seems more natural. Let's revert.

ybiquitous: Thanks. I'll do it. 👍🏼

lib/rules/declaration-block-single-line-max-declarations/index.js
```diff
@@ -48,7 +47,6 @@ const rule = (primary) => {
 			report({
 				message: messages.expected(primary),
 				node: ruleNode,
-				index: beforeBlockString(ruleNode, { noRawBefore: true }).length,
```
ybiquitous, line 51: [ask] The previous version seems more useful to me because the rule is about a block `{ }` (especially when a selector is long). Any thoughts?

(Screenshots on VS Code ⬇️ )

### Before

image

### After

image

lib/rules/function-calc-no-unspaced-operator/index.js
ybiquitous: Fixed via ec84751f

lib/rules/declaration-block-single-line-max-declarations/index.js
jeddy3: I agree. Let's revert.

The `endIndex` should point to the closing brace.

> because the rule is about a block { } (especially when a selector is long). 

Yes, that's correct. The majority of the rules [are named](https://stylelint.io/user-guide/rules/about#names) so that the first part describes the scope of the rule, e.g. [`declaration-property-unit-allowed-list`](https://stylelint.io/user-guide/rules/list/declaration-property-unit-allowed-list) is named as such (where the scope is `declaration`) rather than named `property-unit-allowed-list`. I think we can generally follow this pattern when deciding where the `index` and `endIndex`s should be.

ybiquitous: @jeddy3 Thank you. I understand well. 👍🏼

ybiquitous: Fixed via 19253dd0

ybiquitous:
I've removed the unused utility functions below:

- `getWhitespaceStartIndex()` 2c596890
- `getWhitespaceEndIndex()` 9c675ce3
- `getWordStartIndex()` acca7dd9
- `getWordEndIndex()` 2755c1b8

Let's add them again if they are necessary.

ybiquitous:
All the rules have not been supported yet, but this pull request is ready to review. 🎉 
(we need to address the remaining rules on other pull requests.)

jeddy3 (changes requested):
@ybiquitous Thank you so much for picking this up! You've done fantastic work on top of @adalinesimonian's. 

All the code LGTM. Let's update our docs too. I think we have a couple of examples that need updating [here](https://github.com/stylelint/stylelint/blob/main/docs/developer-guide/formatters.md), [here](https://github.com/stylelint/stylelint/blob/main/docs/developer-guide/plugins.md#testing) and [here](https://github.com/stylelint/stylelint/blob/main/docs/developer-guide/plugins.md#stylelintutilscheckagainstrule).

I think once that's done we can merge and release.

> All the rules have not been supported yet, but this pull request is ready to review.

I agree. When it comes to the other rules, I don't think we need to add end positions for all the reject tests. We should add just enough of them to be confident that the extra `endIndex` logic is working.

ybiquitous:
Ah, thanks for suggesting updating the docs. I'll do it soon. 👍🏼

ybiquitous:
Updating the docs was done via 13030a9.

jeddy3 (approved):
Fantastic work, thank you!

LGTM.

Feel free to merge. Any idea for the changelog entry? Perhaps:

- Added: ranges for warnings that can be used by formatters and integrations

ybiquitous:
@jeddy3 Thanks for the review and suggestion! I'll do it.

ybiquitous:
Done.

> - Added: ranges for warnings that can be used by formatters and integrations ([#5725](https://github.com/stylelint/stylelint/pull/5725)).

PR #5728, opened 2021-11-21
Document requiring custom syntaxes in shared configs

### LINKED ISSUE #5632: Fix custom syntax module path not properly resolving
> Clearly describe the bug

As per https://github.com/stylelint-scss/stylelint-config-standard-scss/issues/2, custom syntax module path is not properly resolved.

Stylelint tries to resolve custom syntax location ([stylelint/stylelint@`main`/lib/getPostcssResult.js#L95](https://github.com/stylelint/stylelint/blob/main/lib/getPostcssResult.js#L95)) but uses Node `require` algorithm, so it search from that file upwards.

If you have dependancy tree where `postcss-scss` is not in one of the root directories, but is nested inside this config’s `node_modules` (it happens if you have different versions of same package), path resolution fails.

None, it’s related to core functionality

Even empty SCSS file would do.

```json
{
}
```

But `postcss-scss` shouldn’t be installed inside root `node_modules`.

`14.0.0`

`stylelint "**/*.scss"`

Yes, it's related to SCSS custom syntax.

Should return at least error/warning related to style code or exit code `0` if everything is OK.

```shell
Cannot resolve custom syntax module "postcss-scss". Check that module "postcss-scss" is available and spelled correctly.

at getCustomSyntax (node_modules/stylelint/lib/getPostcssResult.js:97:10)
at getPostcssResult (node_modules/stylelint/lib/getPostcssResult.js:41:5)
at lintSource (node_modules/stylelint/lib/lintSource.js:76:20)
at Function.standalone [as lint] (node_modules/stylelint/lib/standalone.js:132:26)
```

jeddy3:

Closes #5632

No, it's self-explanatory.

### DIFF
--- a/docs/developer-guide/syntaxes.md
+++ b/docs/developer-guide/syntaxes.md
@@ -17,13 +17,17 @@ We recommend creating a shared-config that:
 
 For example, if you're creating a syntax for a CSS-in-JS library called "foo" then we recommend creating a shared-config called "stylelint-config-standard-foo" with the following content:
 
-```json
-{
-  "extends": ["stylelint-config-standard"],
-  "customSyntax": "postcss-foo",
-  "rules": {
-    "at-rule-no-unknown": null,
-    ..
+```js
+module.exports = {
+  {
+    "extends": ["stylelint-config-standard"],
+    "customSyntax": require("postcss-foo"),
+    "rules": {
+      "at-rule-no-unknown": null,
+      ..
+    }
   }
 }
 ```
+
+We recommended requiring the custom syntax until PostCSS@7 is no longer in circulation.
--- a/docs/developer-guide/syntaxes.md
+++ b/docs/developer-guide/syntaxes.md
@@ -24,7 +24,7 @@ module.exports = {
     "customSyntax": require("postcss-foo"),
     "rules": {
       "at-rule-no-unknown": null,
-      ..
+      // ..
     }
   }
 }


jeddy3:
Do we need coveralls, or can someone with access remove it?

docs/developer-guide/syntaxes.md
```diff
@@ -17,13 +17,17 @@ We recommend creating a shared-config that:
...
-    "at-rule-no-unknown": null,
-    ..
+```js
+module.exports = {
+  {
+    "extends": ["stylelint-config-standard"],
+    "customSyntax": require("postcss-foo"),
+    "rules": {
+      "at-rule-no-unknown": null,
+      ..
```
ybiquitous, line 27: [trivial] How about changing to valid JS as below?

```suggestion
      // ..
```

ybiquitous (approved):
@jeddy3 Thanks! LGTM! 👍🏼

clown-helang:
Has the new version of this issue been released?

jeddy3:
@clown-helang This was a documentation change that's [live on the website](https://stylelint.io/developer-guide/syntaxes) and [shared configs are now following the recommendation](https://github.com/stylelint-scss/stylelint-config-recommended-scss/blob/d77fcac68f819054b501496476d8acd6085cbc4c/index.js#L5), e.g. the SCSS one uses:

```
customSyntax: require('postcss-scss'),
```

The `customSyntax` configuration property has been available [since `14.0.0`](https://stylelint.io/CHANGELOG#1400), which was released a few months ago.

PR #5734, opened 2021-11-22
Add `resolveConfig` to Node.js API

### LINKED ISSUE #5723: Add --print-config equivalent to Node.js API
In vscode-stylelint, we continue to get issues with users confused as to why the extension doesn't work for their non-CSS syntaxes (e.g. SCSS, Less) when their Stylelint config is configured correctly. I was thinking of ways to make this easier for users and I thought that maybe, if there was a way to auto-detect if a user wants a certain file type to be linted, we could have the extension lint the file without having to have extra configuration in the extension.

However, there are different ways this could be achieved, each with pros and cons, and there still stands the question of _if_ this is the right approach in the first place. I think some discussion and brainstorming could help figure out how we can make this less of a headache for users. I'm opening this issue in this repository since my gut feeling is that if there is a way to do this, it would belong in some form in the API here since Stylelint owns all the logic for reading, parsing, and determining effective configuration.

Excludes don't work for this — just because a user didn't exclude a file doesn't mean that they intend for that file to be linted. Since Stylelint is run on the CLI with the paths that it should lint, that's the only context where we know specifically what files the user wants to be linted. I think we should have a way to know this information in the API.

### Maybe check overrides?

For example, let's say we ask Stylelint, "hey, do you have overrides set up for 'test.scss'?" If the resolved config for that particular file is:

```js
module.exports = {
  // …
  overrides: [
    {
      files: ['**/*.scss'], // aha, there is an override for scss files
      customSyntax: 'postcss-scss',
      // …
    },
  ],
};
```

the response would be yes, and then we'd know downstream that we should lint that file. One way this could look:

```js
const shouldLint = stylelint.hasOverride(path);
// -> shouldLint === true
```

#### Caveats

This would only work if the user is using overrides. If, instead, their configuration is:

```js
module.exports = {
  customSyntax: 'postcss-scss',
  // …
};
```

this would result in a no, even though the user's intent here is that SCSS should be listed.

### Includes, similar to tsc?

Another approach could be having an `include` option, similar to some tools like tsc:

```js
module.exports = {
  include: ['**/*.scss', '**/*.css'],
  // …
};
```

This would indicate that Stylelint should only lint files matching those globs. This also would have the additional benefit that you could have Stylelint lint the files you want it to without having to provide it the globs directly on the CLI: `npx stylelint .` would only lint the files the user has configured in their project.

This would also mean that we now have the precise context to know if a user wants a file to be linted. We could ask Stylelint, downstream, if it has a particular file included in the "project":

```js
const shouldLint = stylelint.included(path);
// -> shouldLint === true
```

While at first glance, this seems similar to what we do in the extension with `validate: ['css', 'scss']`, it is fundamentally different in that the user is now able to put into their config the exact files that they want Stylelint to apply to. Therefore, if the user is using `include`, we can know for absolute certain if a user wants to lint a specific file. There is no gap as there is with checking overrides.

#### Caveats

This would be another property added to the configuration. However, at least it wouldn't break any existing behaviour, as it would only come into effect when used.

---

These are just the solutions I came up with off the top of my head. I'd love to hear your thoughts. I definitely feel that there are ways we can make the user experience better here. Unlike ESLint, there are far more preprocessors and contexts in which CSS is used, even in other languages, so we have many more cases where users are going to have all sorts of configurations for all sorts of files.

adalinesimonian:
Closes #5723 

Allows resolving the effective configuration for a given file, similar to `--print-config` on the CLI.

### DIFF
--- a/lib/__tests__/resolveConfig.test.js
+++ b/lib/__tests__/resolveConfig.test.js
@@ -5,37 +5,37 @@ const pluginWarnAboutFoo = require('./fixtures/plugin-warn-about-foo');
 const replaceBackslashes = require('../testUtils/replaceBackslashes');
 const stylelint = require('..');
 
-it('resolveConfig uses getConfigForFile to retrieve the config', () => {
+it('resolveConfig uses getConfigForFile to retrieve the config', async () => {
 	const filepath = replaceBackslashes(
 		path.join(__dirname, 'fixtures/getConfigForFile/a/b/foo.css'),
 	);
 
-	return stylelint.resolveConfig(filepath).then((result) => {
-		expect(result).toEqual({
-			plugins: [path.join(__dirname, '/fixtures/plugin-warn-about-foo.js')],
-			rules: {
-				'block-no-empty': [true],
-				'plugin/warn-about-foo': ['always'],
-			},
-			pluginFunctions: {
-				'plugin/warn-about-foo': pluginWarnAboutFoo.rule,
-			},
-		});
+	const config = await stylelint.resolveConfig(filepath);
+
+	expect(config).toEqual({
+		plugins: [path.join(__dirname, '/fixtures/plugin-warn-about-foo.js')],
+		rules: {
+			'block-no-empty': [true],
+			'plugin/warn-about-foo': ['always'],
+		},
+		pluginFunctions: {
+			'plugin/warn-about-foo': pluginWarnAboutFoo.rule,
+		},
 	});
 });
 
-it('config overrides should apply', () => {
+it('config overrides should apply', async () => {
 	const filepath = replaceBackslashes(
 		path.join(__dirname, 'fixtures/config-overrides/testPrintConfig/style.css'),
 	);
 
-	return stylelint.resolveConfig(filepath).then((result) => {
-		expect(result).toEqual({
-			rules: {
-				'block-no-empty': [true],
-				'color-named': ['never'],
-			},
-		});
+	const config = await stylelint.resolveConfig(filepath);
+
+	expect(config).toEqual({
+		rules: {
+			'block-no-empty': [true],
+			'color-named': ['never'],
+		},
 	});
 });
 
--- a/lib/__tests__/resolveConfig.test.js
+++ b/lib/__tests__/resolveConfig.test.js
@@ -39,6 +39,6 @@ it('config overrides should apply', () => {
 	});
 });
 
-it('resolveConfig with no path should resolve to null', () => {
-	return expect(stylelint.resolveConfig()).resolves.toBeNull();
+it('resolveConfig with no path should resolve to undefined', () => {
+	return expect(stylelint.resolveConfig()).resolves.toBeUndefined();
 });
--- a/lib/__tests__/resolveConfig.test.js
+++ b/lib/__tests__/resolveConfig.test.js
@@ -1,0 +1,44 @@
+'use strict';
+
+const path = require('path');
+const pluginWarnAboutFoo = require('./fixtures/plugin-warn-about-foo');
+const replaceBackslashes = require('../testUtils/replaceBackslashes');
+const stylelint = require('..');
+
+it('resolveConfig uses getConfigForFile to retrieve the config', () => {
+	const filepath = replaceBackslashes(
+		path.join(__dirname, 'fixtures/getConfigForFile/a/b/foo.css'),
+	);
+
+	return stylelint.resolveConfig(filepath).then((result) => {
+		expect(result).toEqual({
+			plugins: [path.join(__dirname, '/fixtures/plugin-warn-about-foo.js')],
+			rules: {
+				'block-no-empty': [true],
+				'plugin/warn-about-foo': ['always'],
+			},
+			pluginFunctions: {
+				'plugin/warn-about-foo': pluginWarnAboutFoo.rule,
+			},
+		});
+	});
+});
+
+it('config overrides should apply', () => {
+	const filepath = replaceBackslashes(
+		path.join(__dirname, 'fixtures/config-overrides/testPrintConfig/style.css'),
+	);
+
+	return stylelint.resolveConfig(filepath).then((result) => {
+		expect(result).toEqual({
+			rules: {
+				'block-no-empty': [true],
+				'color-named': ['never'],
+			},
+		});
+	});
+});
+
+it('resolveConfig with no path should resolve to null', () => {
+	return expect(stylelint.resolveConfig()).resolves.toBeNull();
+});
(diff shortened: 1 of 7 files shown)


lib/resolveConfig.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+const createStylelint = require('./createStylelint');
+const path = require('path');
+
+/**
+ * Resolves the effective configuation for a given file. Resolves to `null` if
+ * no config is found.
+ * @param {string} filePath The path to the file to get the config for.
```
ybiquitous, line 9: [suggestion] How about inserting `-` to distinguish between a parameter name and description?

```suggestion
 * @param {string} filePath - The path to the file to get the config for.
```

lib/resolveConfig.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param {Pick<
+ *   import('stylelint').LinterOptions,
+ *   | 'cwd'
+ *   | 'config'
+ *   | 'configBasedir'
+ *   | 'configFile'
+ * >} options The options to use when creating the Stylelint instance.
```
ybiquitous, line 16: [suggestion] How about marking the `options` parameter as *optional*?

```suggestion
 * >} [options] - The options to use when creating the Stylelint instance.
```

types/stylelint/index.d.ts
```diff
@@ -324,6 +324,16 @@ declare module 'stylelint' {
 			 * @internal
 			 */
 			createLinter: (options: LinterOptions) => InternalApi;
+			/**
+			 * Resolves the effective configuation for a given file. Resolves to
+			 * `null` if no config is found.
+			 * @param filePath The path to the file to get the config for.
+			 * @param options The options to use when creating the Stylelint instance.
```
ybiquitous, line 331: [suggestion] Same as above.

```suggestion
			 * @param filePath - The path to the file to get the config for.
			 * @param [options] - The options to use when creating the Stylelint instance.
```

types/stylelint/index.d.ts
```diff
@@ -324,6 +324,16 @@ declare module 'stylelint' {
...
+			 */
+			resolveConfig: (
+				filePath: string,
+				options: Pick<LinterOptions, 'cwd' | 'config' | 'configBasedir' | 'configFile'>,
```
ybiquitous, line 335: [suggestion] Is `?` necessary to making `options` optional?

```suggestion
				options?: Pick<LinterOptions, 'cwd' | 'config' | 'configBasedir' | 'configFile'>,
```

lib/__tests__/resolveConfig.test.js
```diff
@@ -0,0 +1,44 @@
...
+		});
+	});
+});
+
+it('config overrides should apply', () => {
+	const filepath = replaceBackslashes(
+		path.join(__dirname, 'fixtures/config-overrides/testPrintConfig/style.css'),
+	);
+
+	return stylelint.resolveConfig(filepath).then((result) => {
```
ybiquitous, line 32: [suggestion] How about using async/await to reduce extra nesting?

```js
it('...', async () => {
  // ...

  const result = await stylelint.resolveConfig(filepath);

  expect(result).toEqual(...);
});
```

lib/__tests__/resolveConfig.test.js
```diff
@@ -0,0 +1,44 @@
...
+				'block-no-empty': [true],
+				'color-named': ['never'],
+			},
+		});
+	});
+});
+
+it('resolveConfig with no path should resolve to null', () => {
+	return expect(stylelint.resolveConfig()).resolves.toBeNull();
+});
```
ybiquitous, line 44: [suggestion] Does it make sense to add test cases passing `options` to `resolveConfig()`?

ybiquitous (commented):
@adalinesimonian Great work! 👏🏼 

I've suggested somes small refactoring ideas, but this PR mostly looks good to me. 👍🏼

types/stylelint/index.d.ts
adalinesimonian: Yes, in fact it is, good catch!

types/stylelint/index.d.ts
adalinesimonian: The brackets aren't used in TSDoc since the type definition already specifies whether or not the parameter is optional, but I'll add the hyphens.

adalinesimonian:
@ybiquitous Resolved comments in dfca589, af61233, and 8d664ef.

types/stylelint/index.d.ts
```diff
@@ -324,6 +324,16 @@ declare module 'stylelint' {
 			 * @internal
 			 */
 			createLinter: (options: LinterOptions) => InternalApi;
+			/**
+			 * Resolves the effective configuation for a given file. Resolves to
+			 * `null` if no config is found.
```
ybiquitous, line 329: [question] Is this a typo?

```suggestion
			 * `undefined` if no config is found.
```

And, just a question, which is better in our public API, `null` or `undefined`?

lib/__tests__/resolveConfig.test.js
```diff
@@ -0,0 +1,129 @@
+'use strict';
+
+const path = require('path');
+const pluginWarnAboutFoo = require('./fixtures/plugin-warn-about-foo');
+const replaceBackslashes = require('../testUtils/replaceBackslashes');
+const stylelint = require('..');
+
+describe('resolveConfig', () => {
+	it('resolve to undefined without a path', () => {
```
ybiquitous, line 9: [trivial] Looks a typo:

```suggestion
	it('should resolve to undefined without a path', () => {
```

or

```suggestion
	it('resolves to undefined without a path', () => {
```

ybiquitous (approved):
I've found some typos, but LGTM! 👍🏼

types/stylelint/index.d.ts
adalinesimonian: Yes, it's a mistake. I switched it to `undefined` so that it would match the type of `config` in `LinterOptions` but forgot to update it everywhere.

ybiquitous: I see. Thank you for the explanation!

lib/__tests__/resolveConfig.test.js
adalinesimonian: Done in 3d2ec8b.

types/stylelint/index.d.ts
adalinesimonian: Done in dfbd98e.

ybiquitous (approved):
@adalinesimonian Thanks for the quick fix! 😄

jeddy3 (approved):
LGTM, thanks!

jeddy3:
> - Added: `resolveConfig` to Node.js API ([#5734](https://github.com/stylelint/stylelint/pull/5734)).

PR #5739, opened 2021-11-25
Fix parseError for double-slash comments within selector lists in no-descending-specificity

xboy2012:

It's a code optimize as it is, avoid unnecessary invoke of `parseSelector()` ahead of time. 

I came to refactor this section because I ran into the issue

https://github.com/shellscape/postcss-less/issues/147

To simplify the issue, see the screenshot.
image

double slashed comments between two selectors are not parsed correctly, thus the following code might throw.

the selector returned by `postcss-less` might be invalid (which is a bug),
then `parseSelector()` will throw.

I think the refactor might reduce the occurrence of the above situation.


ybiquitous (commented):
@xboy2012 Thank you for the contribution! Can you please fix one ESLint error, please?

The change looks good to me. Just in case, is this change just a refactoring, not a bugfix?

xboy2012:
It's a code optimize as it is.

I came to refactor this section because I ran into the issue

https://github.com/shellscape/postcss-less/issues/147

the selector returned by `postcss-less` might be invalid (which is a bug),
then `parseSelector()` will throw.

I think the refactor might reduce the occurrence of the above situation.

ybiquitous:
@xboy2012 Thanks for the explanation!

Can you please update this PR description with your bug description and add a test case for the bug reproduction as below? 

https://github.com/stylelint/stylelint/blob/315d3b66e58a195825cbad6fa465c581b35e1aa9/lib/rules/no-descending-specificity/__tests__/index.js#L160-L170

xboy2012:
I have updated the description and the bug detail.

xboy2012:
The CI is blocked, but I don't why it happened. Help please?

ybiquitous:
@xboy2012 We've migrated Coveralls to Codecov (#5749). Could you please rebase the main branch?

jeddy3:
@xboy2012 Please add a test to the `no-descending-specificity` rule.

PR #5744, opened 2021-11-26
Jest: ignore __tests__ and testUtils from coverage

XhmikosR:
Unsure if you all want these in the coverage report, so if you do, feel free to drop the PR :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "collectCoverage": false,
     "collectCoverageFrom": [
       "lib/**/*.js",
-      "!lib/vendor/**/*.js"
+      "!lib/**/{__tests__,testUtils}/**/*.js"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -76,7 +76,7 @@
     "collectCoverage": false,
     "collectCoverageFrom": [
       "lib/**/*.js",
-      "!lib/vendor/**/*.js"
```
XhmikosR, line 79: BTW the vendor folder is no loner present anyway.

ybiquitous (commented):
@XhmikosR Good catch! This PR seems good to me, but still WIP?

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #5763, opened 2021-12-04
Fix showing of incorrect missing package in customSyntax require handling 

### LINKED ISSUE #5761: Fix showing of incorrect missing package in customSyntax require handling

Given the package required for the customSyntax option itself encounters a `'MODULE_NOT_FOUND'`. Then that error is swallows and replaced by an error stating that the customSyntax module could not be found. 

This caused by an incomplete fix in this PR https://github.com/stylelint/stylelint/pull/5635/files. 

You can test this by setting up your environment and running ` require('@stylelint/postcss-css-in-js')` from a node repl. 

```
  overrides: [
      {
        files: ['**/*.js', '**/*.jsx', '**/*.ts', '**/*.tsx'],
        customSyntax: '@stylelint/postcss-css-in-js',
   }
]
```

You must also be missing a dependency of `@stylelint/postcss-css-in-js` for example `postcss-syntax/load-syntax`. 

`"stylelint": "stylelint **/*.{js,ts,tsx}"` in package.json

14.1.0

I expected the customSyntax error to be accurate. 

### What actually happened?

The customSyntax error was not the actual error. 

No

### Proposal to fix the bug

This line should check for something more specific than a `MODULE_NOT_FOUND` error: `error && typeof error === 'object' && error.code === 'MODULE_NOT_FOUND'`. Perhaps that codeblock should emit the underling error as well.

dlaub3:

Closes #5761

I'm not sure if the added `console.error` is beneficial. What are your thoughts?

### DIFF
--- a/lib/getPostcssResult.js
+++ b/lib/getPostcssResult.js
@@ -102,7 +102,6 @@ function getCustomSyntax(customSyntax) {
 				// @ts-expect-error -- TS2571: Object is of type 'unknown'.
 				error.message.includes(customSyntax)
 			) {
-				console.error('Stylelint MODULE_NOT_FOUND', error);
 
 				throw new Error(
 					`Cannot resolve custom syntax module "${customSyntax}". Check that module "${customSyntax}" is available and spelled correctly.\n\nCaused by: ${error}`,
--- a/lib/getPostcssResult.js
+++ b/lib/getPostcssResult.js
@@ -102,7 +102,6 @@ function getCustomSyntax(customSyntax) {
 				// @ts-expect-error -- TS2571: Object is of type 'unknown'.
 				error.message.includes(customSyntax)
 			) {
-
 				throw new Error(
 					`Cannot resolve custom syntax module "${customSyntax}". Check that module "${customSyntax}" is available and spelled correctly.\n\nCaused by: ${error}`,
 				);
--- a/lib/getPostcssResult.js
+++ b/lib/getPostcssResult.js
@@ -94,8 +94,16 @@ function getCustomSyntax(customSyntax) {
 		try {
 			resolved = require(customSyntax);
 		} catch (error) {
-			// @ts-expect-error -- TS2571: Object is of type 'unknown'.
-			if (error && typeof error === 'object' && error.code === 'MODULE_NOT_FOUND') {
+			if (
+				error &&
+				typeof error === 'object' &&
+				// @ts-expect-error -- TS2571: Object is of type 'unknown'.
+				error.code === 'MODULE_NOT_FOUND' &&
+				// @ts-expect-error -- TS2571: Object is of type 'unknown'.
+				error.message.includes(customSyntax)
+			) {
+				console.error('Stylelint MODULE_NOT_FOUND', error);
+
 				throw new Error(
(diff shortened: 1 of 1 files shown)


lib/getPostcssResult.js
```diff
@@ -94,8 +94,16 @@ function getCustomSyntax(customSyntax) {
...
+				typeof error === 'object' &&
+				// @ts-expect-error -- TS2571: Object is of type 'unknown'.
+				error.code === 'MODULE_NOT_FOUND' &&
+				// @ts-expect-error -- TS2571: Object is of type 'unknown'.
+				error.message.includes(customSyntax)
+			) {
+				console.error('Stylelint MODULE_NOT_FOUND', error);
+
 				throw new Error(
 					`Cannot resolve custom syntax module "${customSyntax}". Check that module "${customSyntax}" is available and spelled correctly.`,
```
ybiquitous, line 108: [suggestion] How about embedding a caused error into the existing message like this? (instead of `console.error()`)

```suggestion
					`Cannot resolve custom syntax module "${customSyntax}". Check that module "${customSyntax}" is available and spelled correctly.\n\nCaused by: ${error}`,
```

---

BTW, `Error.prototype.cause` is available since ES2021 and Node 16.9.0 and seems suitable for this case. But unfortunately, we need to support older Node.js versions for a while. 😓 

See also:
- https://2ality.com/2021/06/error-cause.html
- https://github.com/tc39/proposal-error-cause
- https://nodejs.org/en/blog/release/v16.9.0/#error-cause

ybiquitous (commented):
@dlaub3 Thank you for creating this PR! I've suggested one idea for the usage of `console.error()`.

lib/getPostcssResult.js
```diff
@@ -94,10 +94,18 @@ function getCustomSyntax(customSyntax) {
...
-			if (error && typeof error === 'object' && error.code === 'MODULE_NOT_FOUND') {
+			if (
+				error &&
+				typeof error === 'object' &&
+				// @ts-expect-error -- TS2571: Object is of type 'unknown'.
+				error.code === 'MODULE_NOT_FOUND' &&
+				// @ts-expect-error -- TS2571: Object is of type 'unknown'.
+				error.message.includes(customSyntax)
+			) {
+				console.error('Stylelint MODULE_NOT_FOUND', error);
```
dlaub3, line 105:
```suggestion
```

lib/getPostcssResult.js
dlaub3: TIL `Error.prototype.cause`  :+1:

mattxwang:
@dlaub3 looks like it fails the CI Lint, do you want to try running `npm run format` on your fork?

dlaub3:
> @dlaub3 looks like it fails the CI Lint, do you want to try running `npm run format` on your fork?

This is the downside to applying changes via the GitHub UI. It should be good now.

mattxwang (approved):
I'll defer back to @ybiquitous for a final review if needed, but on my end this LGTM! Thanks for the contribution 😄

ybiquitous (approved):
@dlaub3 Thank you. LGTM 👍🏼

ybiquitous:
> - Fixed: Showing of incorrect missing package in `customSyntax` require handling ([#5763](https://github.com/stylelint/stylelint/pull/5763)).

PR #5771, opened 2021-12-10
Fix TypeError for syntaxes that use Document in indentation

### LINKED ISSUE #5767: Fix TypeError for syntaxes that use Document in indentation

If you have a `Root` in a `Document` whose css begins on line 1, like so:

```css
  .foo {
    color: hotpink;
  }
```

The indentation rule will throw, as it tries to access `root.raws.beforeStart` which doesn't exist.

```json
{
  "rules": {
    "indentation": 2
  }
}
```

CLI

14

Indentation rule would succeed.

### What actually happened?

The rule throws:

```
TypeError: Cannot read properties of undefined (reading 'endsWith')
```

No

### Proposal to fix the bug

https://github.com/stylelint/stylelint/blob/0d2303ef01e065a1005b6b7ebcf51611223655f7/lib/rules/indentation/index.js#L94-L98

Ultimately, we hit this condition. It is comprised of these:

Does `node.raws.before` contain a newline? - No, so we continue.

Is it the first child? - Yes.

Does it's parent have a document? - Yes.

This leads us to then attempting to execute `parent.raws.beforeStart.endsWith('\n')` which will fail if `beforeStart` doesn't exist.

We should null check `parent.raws.beforeStart` as it may never exist, or a plugin could have since removed it.

Btw I am aware of #5674 , so please let me know if you'll never fix this as I can fix it on my end instead (by populating `beforeStart` with an empty string at least).

43081j:
Fixes #5767 

I changed `beforeStart` to `codeBefore`  and added a truthy check

### DIFF
--- a/lib/rules/indentation/__tests__/document-with-roots.js
+++ b/lib/rules/indentation/__tests__/document-with-roots.js
@@ -1,0 +1,37 @@
+'use strict';
+
+const { parse: postcssParse, stringify, Document, Input } = require('postcss');
+const { ruleName } = require('..');
+
+const parse = (source, opts) => {
+	const doc = new Document();
+	const root = postcssParse(source, opts);
+
+	root.parent = doc;
+	root.document = doc;
+	doc.nodes.push(root);
+	doc.source = {
+		input: new Input(source, opts),
+		start: { line: 1, column: 1, offset: 0 },
+	};
+
+	return doc;
+};
+
+testRule({
+	ruleName,
+	config: [2],
+	fix: true,
+	customSyntax: {
+		parse,
+		stringify,
+	},
+
+	accept: [
+		{
+			code: '.foo {\n' + '  color: hotpink;\n' + '}',
+		},
+	],
+
+	reject: [],
+});
--- a/lib/rules/indentation/index.js
+++ b/lib/rules/indentation/index.js
@@ -93,7 +93,9 @@ const rule = (primary, secondaryOptions = {}, context) => {
 			// it is some other kind of separation, checked by some separate rule
 			if (
 				(lastIndexOfNewline !== -1 ||
-					(isFirstChild && (!getDocument(parent) || parent.raws.beforeStart.endsWith('\n')))) &&
+					(isFirstChild &&
+						(!getDocument(parent) ||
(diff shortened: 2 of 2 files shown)


lib/rules/indentation/__tests__/document-with-roots.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+const postcssParse = require('postcss/lib/parse');
+const { stringify, Document, Input } = require('postcss');
```
ybiquitous, line 4: [suggestion] I think it's better to avoid internal importing like `require('postcss/lib/parse')` if possible. So how about refactoring like this?

```suggestion
const { parse: postcssParse, stringify, Document, Input } = require('postcss');
```

lib/rules/indentation/index.js
```diff
@@ -628,7 +630,7 @@ function inferRootIndentLevel(root, baseIndentLevel, indentSize) {
 	}
 
 	const indents = [];
-	const foundIndents = /(?:^|\n)([ \t]*)\S/m.exec(root.raws.beforeStart);
+	const foundIndents = root.raws.codeBefore && /(?:^|\n)([ \t]*)\S/m.exec(root.raws.beforeStart);
```
ybiquitous, line 633: [suggestion] Isn't it needed to use `beforeStart`?

```suggestion
	const foundIndents = root.raws.codeBefore && /(?:^|\n)([ \t]*)\S/m.exec(root.raws.codeBefore);
```

ybiquitous (commented):
@43081j Thank you for creating this PR!
I've left some trivial suggestions, but this is mostly good to me. 👍🏼

lib/rules/indentation/index.js
43081j: ah good catch! just missed it

43081j:
sorry this got a bit lost @jeddy3, i've updated it now with what you said

ybiquitous (approved):
@43081j Thank you! LGTM 👍🏼

jeddy3 (commented):
LGTM, thanks.

jeddy3 (approved):

jeddy3:
> - Fixed: `indentation` TypeError for syntaxes that use Document node type ([#5771](https://github.com/stylelint/stylelint/pull/5771)).

PR #5781, opened 2021-12-15
Refactor benchmark script to ESM and node-fetch

mattxwang:

Stems from the conversation in and closes #5773.

This PR's original goal was to remove `got` as a dependency (and use something more lightweight, in this case `node-fetch`) but the scope has increased a bit. Here's a chronological step-by-step of what I had to do:

1. Removed `got` in favour of `node-fetch`. This involves slightly changing the promise structure.
2. In realizing that the latest version of `node-fetch` is ESM-only, and that the benchmark script is still in CJS, upgrades it to ESM. While this is mostly routine,
    * I rename the file extension to `.mjs`. This doesn't seem breaking, since nobody consumes this script.
    * `picocolors` is still a CJS module, so I implement the suggested workaround to make destructured imports work properly.
    * ~~the `rules` import is a directory import, but this requires an [experimental flag to be enabled for ESM](https://stackoverflow.com/questions/59556130/esm-importing-all-files-from-folder-with-index-js-not-working). I add this to the script in `package.json`~~ resolved, thanks @ybiquitous!

I've confirmed that the functionality (and the stats) have not changed for some trivial uses of the benchmark rule, like so:

```
$ npm run benchmark-rule -- color-no-invalid-hex true                                                                  update-benchmark-esm ] 6:12 PM

> stylelint@14.1.0 benchmark-rule
> node --experimental-specifier-resolution=node scripts/benchmark-rule.mjs "color-no-invalid-hex" "true"

Warnings: 0
Mean: 26.194505096618354 ms
Deviation: 14.998938116856769 ms
```

I don't have much context on how the benchmark script is normally used, so any feedback here is appreciated!

### DIFF
--- a/scripts/benchmark-rule.mjs
+++ b/scripts/benchmark-rule.mjs
@@ -6,7 +6,7 @@ import fetch from 'node-fetch';
 import normalizeRuleSettings from '../lib/normalizeRuleSettings.js';
 import picocolors from 'picocolors';
 import postcss from 'postcss';
-import rules from '../lib/rules';
+import rules from '../lib/rules/index.js';
 const { bold, yellow } = picocolors;
 
 const ruleName = process.argv[2];
--- a/scripts/benchmark-rule.mjs
+++ b/scripts/benchmark-rule.mjs
@@ -1,12 +1,13 @@
 'use strict';
 
 /* eslint-disable no-console */
-const Benchmark = require('benchmark');
-const got = require('got');
-const normalizeRuleSettings = require('../lib/normalizeRuleSettings');
-const postcss = require('postcss');
-const rules = require('../lib/rules');
-const { bold, yellow } = require('picocolors');
+import Benchmark from 'benchmark';
+import fetch from 'node-fetch';
+import normalizeRuleSettings from '../lib/normalizeRuleSettings.js';
+import picocolors from 'picocolors';
+import postcss from 'postcss';
+import rules from '../lib/rules';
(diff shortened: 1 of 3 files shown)


scripts/benchmark-rule.mjs
```diff
@@ -1,12 +1,13 @@
...
-const normalizeRuleSettings = require('../lib/normalizeRuleSettings');
-const postcss = require('postcss');
-const rules = require('../lib/rules');
-const { bold, yellow } = require('picocolors');
+import Benchmark from 'benchmark';
+import fetch from 'node-fetch';
+import normalizeRuleSettings from '../lib/normalizeRuleSettings.js';
+import picocolors from 'picocolors';
+import postcss from 'postcss';
+import rules from '../lib/rules';
```
ybiquitous, line 9: [suggestion] How about specifying the file name to remove the `--experimental-specifier-resolution=node` flag?

```suggestion
import rules from '../lib/rules/index.js';
```

ybiquitous (approved):
@mattxwang Thank you for the quick PR! I've left a refactoring suggestion, but LGTM! 👍🏼

scripts/benchmark-rule.mjs
mattxwang: Ah smart, completely slipped my mind. Will do!

ybiquitous (approved):
Thanks! 👍🏼

jeddy3 (approved):
LGTM, thanks!

PR #5793, opened 2021-12-20
Fix false positives for variables and color functions in color-function-notation

### LINKED ISSUE #5671: Fix false positives for variables and color functions in `color-function-notation`
> Clearly describe the bug

A follow-up from #5638, where the original issue was fixed correctly in my tests but more advanced scenarios still result in false positives, e.g. when using Sass' `color.mix` within `rgba()`.

`color-function-notation`

```scss
@use 'sass:color';

$color: rgba(color.mix(#000, #fff, 35%), 0.6);
```

```json
{
  "rules": {
    "color-function-notation": "modern"
  }
}
```

`14.0.1` as well as `main` from commit e2c3473.

```
stylelint \"src/**/*.scss\"
```

Yes, SCSS color functions

No errors

```shell
 25:27  ✖  Expected legacy color-function notation  color-function-notation
```

lachieh:
Fixes #5671

### DIFF
--- a/lib/rules/color-function-notation/__tests__/index.js
+++ b/lib/rules/color-function-notation/__tests__/index.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const postcssScss = require('postcss-scss');
 const stripIndent = require('common-tags').stripIndent;
 
 const { messages, ruleName } = require('..');
@@ -69,6 +68,28 @@ testRule({
 						/* comment */ / 15%
 					)
 				}`,
+		},
+		{
+			code: '$a: rgb(0 0 0);',
+		},
+		{
+			code: '$a: rgb(0 0 0 / 0);',
+		},
+		{
+			code: 'a { color: rgb($a $b $c / $d) }',
+		},
+		{
+			code: 'a { color: rgba($a, $b, $c, $d) }',
+			description: 'scss color function with 4 arguments',
+		},
+		{
+			code: 'a { color: rgb($color-var, 50%); }',
+		},
+		{
+			code: 'a { color: rgba(color.mix(#000, #fff, 35%), 0.6); }',
+		},
+		{
+			code: 'a { color: rgba(#fff, 0.85) }',
 		},
 	],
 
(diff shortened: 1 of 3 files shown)


lachieh:
Question here for maintainers. The rules that are failing now from the scss syntax section:   
https://github.com/stylelint/stylelint/blob/d630d96ae75a8f983db310ad086ed6872149048f/lib/rules/color-function-notation/__tests__/index.js#L407-L413
https://github.com/stylelint/stylelint/blob/d630d96ae75a8f983db310ad086ed6872149048f/lib/rules/color-function-notation/__tests__/index.js#L441-L446

Are these rules valid tests cases here? Should this be something that is handled by [stylelint-scss](https://github.com/stylelint-scss/stylelint-scss)? It seems like part of the problem is that the contents of the variable is what needs to be checked first and that's not a responsibility of stylelint.

jeddy3:
Thanks for the pull request!

> Are these rules valid tests cases here?

Nope. Those should be accepted now, rather than rejected. 

> Should this be something that is handled by stylelint-scss?

Yes, exactly.

lachieh:
Looking closer at theses tests, is the scss section necessary at all or should I merge those test cases with the main syntax rules?

jeddy3:
> Looking closer at theses tests, is the scss section necessary at all or should I merge those test cases with the main syntax rules?

They should be in a separate `testRule` (that uses the `postcss-scss` syntax) so that we signal that we're testing non-standard constructs (which, in this case, are dollar variables - both defined and used).  It helps us maintain the line between CSS and non-standard stuff more clearly.

lachieh:
Got it, thanks. This one is ready for review now.

ybiquitous (commented):
@lachieh Thank you for creating this pull request!

It would seem better to me if you could add some test cases to `isStandardSyntaxColorFunction.test.js`.

lachieh:
@ybiquitous Thanks, added some test cases. Let me know if it needs any more.

ybiquitous (approved):
@lachieh Thank you! LGTM 👍🏼

jeddy3 (approved):
LGTM, thank you!

ybiquitous:
@jeddy3 Will you update the changelog?

jeddy3:
Whoops, good catch. Thanks @ybiquitous !

Added now:

> - Fixed: `color-function-notation` false positives for variables and color functions ([#5793](https://github.com/stylelint/stylelint/pull/5793))

PR #5806, opened 2021-12-30
Add autofix to font-family-name-quotes

### LINKED ISSUE #5805: Support autofix for `font-family-name-quotes` rule

create a css file with
```css
body {
  margin: 0;
  font-family: "Roboto", sans-serif;
}
```
enable rule `'font-family-name-quotes': 'always-where-recommended'`

```json
{
  "rules": {
    "font-family-name-quotes": "always-where-recommended"
  }
}
```

`stylelint index.css --fix`

`14.2.0`

`"Roboto"` should be converted to `Roboto` after `--fix` flag.

### What actually happened?

`--fix` does not automatically fix the problem.

_No response_

### Proposal to fix the bug

I have a hunch that this was working in `stylelint@13` but I'm not sure. either way shouldn't this be an automatically fixable problem by removing the quotes on font family names that don't have spaces or misc symbols.

MrBrN197:

Closes #5805

- On [this line](https://github.com/MrBrN197/stylelint/blob/502e693451fe53dfe83f64ba6dd99b892e9d526c/lib/rules/font-family-name-quotes/index.js#L96) I am using the `result` object to check against the `string-quotes` rule to determine whether to replace with single or double-quotes. I'm not sure whether this is wise or not. I'm not sure what order the rules run in so if the `font-family-name-quotes` rule runs after the `string-quotes` this might cause it to replace with the wrong type of quotes and `string-quotes` will not get a chance to fix this.
- `checkFamilyName` returns an object when `context.fix` is enabled and `complain` doesn't run. modifying the `decl.value` inside `checkFamilyName` would result in the other `fontFamilyNode.sourceIndex`s' being wrong so I defer this until after the loop. I'm not sure if this is the best way to go about this. but to avoid rewriting a lot of the code this is what I decided to do. some clarification on this would be nice.


ybiquitous:
@MrBrN197 Thank you for creating the pull request and sorry for the late response.

> * On [this line](https://github.com/MrBrN197/stylelint/blob/502e693451fe53dfe83f64ba6dd99b892e9d526c/lib/rules/font-family-name-quotes/index.js#L96) I am using the `result` object to check against the `string-quotes` rule to determine whether to replace with single or double-quotes. I'm not sure whether this is wise or not. I'm not sure what order the rules run in so if the `font-family-name-quotes` rule runs after the `string-quotes` this might cause it to replace with the wrong type of quotes and `string-quotes` will not get a chance to fix this.

With the current implementation, `string-quotes` is executed after `font-family-name-quotes`, that is, by the order defined `lib/rules/index.js`. See below:

https://github.com/stylelint/stylelint/blob/8a4253a56b683c32b127a612383372bb86ca99fc/lib/lintPostcssResult.js#L57-L61

https://github.com/stylelint/stylelint/blob/8a4253a56b683c32b127a612383372bb86ca99fc/lib/rules/index.js#L8

This behavior is undocumented as far as I know and may be changed in the future. However, since it will be as-is for a while, I recommend removing the code referring `config.rules['string-quotes']`.

So, it doesn't seem to be a big problem, even if `font-family-name-quotes` *always* fixes CSS with single quotes (or double quotes).


> * `checkFamilyName` returns an object when `context.fix` is enabled and `complain` doesn't run. modifying the `decl.value` inside `checkFamilyName` would result in the other `fontFamilyNode.sourceIndex`s' being wrong so I defer this until after the loop. I'm not sure if this is the best way to go about this. but to avoid rewriting a lot of the code this is what I decided to do. some clarification on this would be nice.

OK, I'll review the code more.

ybiquitous (commented):
[suggest] Can you please add the following description to the `font-family-name-quotes` doc?

https://github.com/stylelint/stylelint/blob/8a4253a56b683c32b127a612383372bb86ca99fc/lib/rules/string-quotes/README.md#L28

lib/rules/font-family-name-quotes/__tests__/index.js
```diff
@@ -390,3 +390,264 @@ testRule({
...
+	ruleName,
+	config: ['always-unless-keyword'],
+	fix: true,
+	reject: [
+		{
+			code: 'a { font-family: "Lucida Grande", "Arial", "sans-serif"; }',
+			message: messages.rejected('sans-serif'),
+			line: 1,
+			column: 45,
+			fixed: 'a { font-family: "Lucida Grande", "Arial", sans-serif; }',
```
ybiquitous, line 404:
```suggestion
			code: 'a { font-family: "Lucida Grande", "Arial", "sans-serif"; }',
			fixed: 'a { font-family: "Lucida Grande", "Arial", sans-serif; }',
			message: messages.rejected('sans-serif'),
			line: 1,
			column: 45,
```

lib/rules/font-family-name-quotes/__tests__/index.js
```diff
@@ -390,3 +390,264 @@ testRule({
...
+		},
+		{
+			code: 'a { font: 1em "Lucida Grande", "Arial", "sans-serif"; }',
+			message: messages.rejected('sans-serif'),
+			line: 1,
+			column: 42,
+			fixed: 'a { font: 1em "Lucida Grande", "Arial", sans-serif; }',
```
ybiquitous, line 411:
```suggestion
			code: 'a { font: 1em "Lucida Grande", "Arial", "sans-serif"; }',
			fixed: 'a { font: 1em "Lucida Grande", "Arial", sans-serif; }',
			message: messages.rejected('sans-serif'),
			line: 1,
			column: 42,
```

ybiquitous (commented):
[suggest] Can you move `fixed:` next to `code:` like other cases?

lib/rules/font-family-name-quotes/index.js
```diff
@@ -102,12 +132,22 @@ const rule = (primary) => {
...
 			if (keywordSets.fontFamilyKeywords.has(family.toLowerCase()) || isSystemFontKeyword(family)) {
 				if (hasQuotes) {
+					if (context.fix) {
+						const openIndex = fontFamilyNode.sourceIndex;
+						const closeIndex = openIndex + rawFamily.length - 1;
+
+						return {
+							type: 'remove',
+							indexes: [openIndex, closeIndex],
+						};
```
ybiquitous, line 148: [suggest] It seems a bit hard to read to me that `checkFamilyName()` returns an object or `undefined`. 🤔 

For example, can you call a helper function like `removeQuotes()` as below (`addQuotes()` as well)?

```suggestion
						removeQuotes(decl, fontFamilyNode);
						return;
```

ybiquitous (commented):
@MrBrN197 I've left some suggestions. I would appreciate them if you could consider them. 😃

jeddy3:
> So, it doesn't seem to be a big problem, even if font-family-name-quotes always fixes CSS with single quotes (or double quotes).

Yes, let's always fix with double quotes. It's what the standard config uses because double quotes are most commonly used in the specification examples.

jeddy3:
@MrBrN197 Thanks for making those changes.

Can you implement @ybiquitous' [two suggestions](https://github.com/stylelint/stylelint/pull/5806#pullrequestreview-844837296) above, and then I think we'll be ready for a final review?

MrBrN197:
@jeddy3 Hey, Thanks, Yes I'm getting on the other changes.

lib/rules/font-family-name-quotes/index.js
MrBrN197: I'm currently working on this. the declaration value can have multiple font-families. If I add quotes or remove quotes from the `decl.value` the indexes in the reset of the following `fontFamilyNode`s will be offset and be incorrect. so I'm currently still working on this in my free time.

lib/rules/font-family-name-quotes/index.js
```diff
@@ -56,8 +56,67 @@ function quotesRequired(family) {
 		.some((word) => /^(?:-?\d|--)/.test(word) || !/^[-\w\u{00A0}-\u{10FFFF}]+$/u.test(word));
 }
 
+/**
+ * @typedef {{
+ * name: string,
+ * hasQuotes: boolean,
+ * sourceIndex: number,
+ * removeQuotes: function,
+ * addQuotes: function
```
ybiquitous, line 65: [suggest] How about the following refactoring ideas?

- Make indentation for each property
- Avoid `function` for more clarity

```suggestion
 *   name: string,
 *   hasQuotes: boolean,
 *   sourceIndex: number,
 *   removeQuotes: () => void,
 *   addQuotes: () => void,
```

lib/rules/font-family-name-quotes/index.js
```diff
@@ -56,8 +56,67 @@ function quotesRequired(family) {
...
+ * }} MutableNode
+ */
+
+/**
+ *
+ * @param {import('postcss-value-parser').Node[]} fontFamilies
+ * @param {import('postcss').Declaration} decl
```
ybiquitous, line 72: [suggest] How about adding `@returns` for more readability and type-safety?

```suggestion
 * @param {import('postcss').Declaration} decl
 * @returns {MutableNode[]}
```

lib/rules/font-family-name-quotes/index.js
```diff
@@ -56,8 +56,67 @@ function quotesRequired(family) {
...
+			resetIndexes: (/** @type {number} */ offset) => {
+				mutableNodes.slice(idx + 1).forEach((n) => (n.sourceIndex += offset));
+			},
+			removeQuotes() {
+				if (this.hasQuotes === false) return;
+
+				const openIndex = this.sourceIndex;
+				const closeIndex = openIndex + this.name.length + 2;
+
+				this.hasQuotes = true;
```
ybiquitous, line 94: [ask] Is `false` valid?

```suggestion
				this.hasQuotes = false;
```

lib/rules/font-family-name-quotes/index.js
```diff
@@ -69,28 +128,26 @@ const rule = (primary) => {
 		}
 
 		root.walkDecls(/^font(-family)?$/i, (decl) => {
-			const fontFamilies = findFontFamily(decl.value);
+			let fontFamiliess = MakeMutableFontFamilies(findFontFamily(decl.value), decl);
```
ybiquitous, line 131: [suggest]

- `const` is preferred over `let` here
- `fontFamilyNodes` is more readable than `fontFamilies`

```suggestion
			const fontFamilyNodes = MakeMutableFontFamilies(findFontFamily(decl.value), decl);
```

lib/rules/font-family-name-quotes/index.js
```diff
@@ -69,28 +128,26 @@ const rule = (primary) => {
...
+			for (const fontFamilyNode of fontFamiliess) {
+				checkFamilyName(fontFamilyNode, decl);
 			}
 		});
 
 		/**
-		 * @param {string} rawFamily
-		 * @param {import('postcss').Declaration} decl
+		 * @param { MutableNode } fontFamilyNode
+		 * @param { import('postcss').Declaration} decl
```
ybiquitous, line 144: [suggest] How about removing extra spaces?

```suggestion
		 * @param {MutableNode} fontFamilyNode
		 * @param {import('postcss').Declaration} decl
```

lib/rules/font-family-name-quotes/index.js
```diff
@@ -56,8 +56,67 @@ function quotesRequired(family) {
...
+ * addQuotes: function
+ * }} MutableNode
+ */
+
+/**
+ *
+ * @param {import('postcss-value-parser').Node[]} fontFamilies
+ * @param {import('postcss').Declaration} decl
+ */
+const MakeMutableFontFamilies = (fontFamilies, decl) => {
```
ybiquitous, line 74: [suggest] How about a lower-case function name as below?

```suggestion
const makeMutableFontFamilies = (fontFamilies, decl) => {
```

ybiquitous (commented):
@MrBrN197 Thank you and sorry for the late response. Great work. 👍🏼 

I've left some easy refactoring suggestions, so please consider them.

lib/rules/font-family-name-quotes/index.js
MrBrN197: thanks

MrBrN197:
@ybiquitous Thanks for the feedback. I've made those changes.

ybiquitous (approved):
@MrBrN197 Great! LGTM 👍🏼

ota-meshi:
I think we need to mark the rule as `(Autofixable)` in the list of rules.
https://github.com/stylelint/stylelint/blob/main/docs/user-guide/rules/list.md

ybiquitous (approved):

jeddy3 (approved):
@MrBrN197 Thank you! LGTM.

jeddy3:
Changelog entry:

> - Added: `font-family-name-quotes` autofix ([#5806](https://github.com/stylelint/stylelint/pull/5806)).

PR #5811, opened 2022-01-05
Update CI badge

LitoMore:

Badges.

No.

### Changes

This is a workaround to fix the badge in https://stylelint.io homepage. I replaced the sheids.io badge with the GitHub one. As you can see, the URL of the new badge is from GitHub.

The weird backslash `\` in the badge comes from the Remark library. Seems there are some bugs when Remark trying to parse the  `&` in ``.

We can update the workflow name in https://github.com/stylelint/stylelint/blob/main/.github/workflows/testing.yml#L1 if you want to keep the `CI` on the badge. That's up to you.

### Preview

| Before | After |
| :-: | :-: |
|  |  |


jeddy3 (approved):
LGTM, thank you!

I noticed that backslash the other day and did wonder what it was. Thanks for the sleuthing.

ybiquitous (approved):
@LitoMore Thank you! 👍🏼

PR #5825, opened 2022-01-11
Refactor to improve types for `no-irregular-whitespace` rule

ybiquitous:

Part of #4496

- Convert the internal logic to a more type-safe way; not using `***Schema` objects for nodes, but accessing directly to each node's properties.


ybiquitous:
I'll add more test cases for at-rule:

image

ybiquitous:
@jeddy3 Thanks. I think `96.33% (-0.01%) compared to ab6ee66` by Codecov is ignorable in this case. 😃

PR #5828, opened 2022-01-11
Document use of overrides in custom syntax configs

### LINKED ISSUE #5786: Document recommended use of overrides in shared configs that use `customSyntax`

In https://github.com/stylelint/stylelint/issues/5777, I was going to suggest extending the appropriate configs, i.e. [stylelint-config-standard-scss](https://github.com/stylelint-scss/stylelint-config-standard-scss) and [stylelint-config-html](https://github.com/ota-meshi/stylelint-config-html). However, I just noticed that they take different approaches to `overrides`, i.e.:

- stylelint-config-standard-scss [doesn't use them](https://github.com/stylelint-scss/stylelint-config-standard-scss/blob/main/index.js)
- stylelint-config-html [does use them](https://github.com/ota-meshi/stylelint-config-html/blob/main/html.js)

If a user wants to lint CSS, HTML and SCSS they will need to use:

```json
{
  "extends": ["stylelint-config-html"],
  "overrides": [
    { "files": "**/*.scss", "extends": ["stylelint-config-standard-scss"] },
  ],
  "rules": {
    "color-no-invalid-hex": true
  }
}
```

i.e. `stylelint-config-html` can be extended directly, but `stylelint-config-standard-scss` must be in an `overrides`. It would be a better experience for people if they could be consumed _in the same way_.

This will not currently work:

```json
{
  "extends": ["stylelint-config-html", "stylelint-config-standard-scss"],
  "rules": {
    "color-no-invalid-hex": true
  }
}
```

Because `stylelint-config-standard-scss` sets the `customSyntax` to `postcss-scss` for _all_ files.

Let's decide which approach we think is best, [document it](https://stylelint.io/developer-guide/syntaxes) and then help update the community packages so that they're all consistent.

There are advantages and disadvantages to both approaches.

If we use `overrides` in these shared configs then they are a bit easier to use, i.e. users just need to extend the config and not need to learn the `overrides` property. However, the disadvantage is that configs will need to track all the file extensions that could use the custom syntax as the config can't be applied to any arbitrary extension. For example, the following won't work (until `.xyz` is added to the config):

```json
{
  "overrides": [{ "files": "**/*.xyz", "extends": ["stylelint-config-html"] }],
  "rules": {
    "color-no-invalid-hex": true
  }
}
```

Where `.xyz` is the hypothetical extension of a new HTML-like CSS-in-JS library, as `stylelint-config-html` itself contains an `overrides` for a list of extensions.

I'm leaning towards recommending `overrides` in these shared configs because I think it'll be easier for the most number of users. For example, if you want to lint HTML, CSS and SCSS files you can simply use (and in any order):

```json
{
  "extends": [
    "stylelint-config-standard",
    "stylelint-config-html"
  ]
}
```

```console
npx stylelint "**/*.{css,scss,html}"
```

/cc @ota-meshi as author of the HTML and Vue configs

What does everyone think?

jeddy3:

Closes #5786

No, it's self-explanatory.

### DIFF
--- a/docs/developer-guide/syntaxes.md
+++ b/docs/developer-guide/syntaxes.md
@@ -1,6 +1,6 @@
 # Writing custom syntaxes
 
-Custom syntaxes are [PostCSS syntaxes](https://github.com/postcss/postcss#syntaxes) written by the community to support other styling languages or CSS-in-JS libraries using the [`customSyntax` option](../user-guide/usage/options.md#customSyntax)
+Custom syntaxes are [PostCSS syntaxes](https://github.com/postcss/postcss#syntaxes) written by the community to support other styling languages and CSS-in-JS libraries using the [`customSyntax` option](../user-guide/usage/options.md#customSyntax)
 
 To write one, familiarize yourself with PostCSS's [how to write custom syntax](https://github.com/postcss/postcss/blob/main/docs/syntax.md) guide.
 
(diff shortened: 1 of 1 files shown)


docs/developer-guide/syntaxes.md
```diff
@@ -1,33 +1,42 @@
 # Writing custom syntaxes
 
-Custom syntaxes are [PostCSS syntaxes](https://github.com/postcss/postcss#syntaxes) written by the community to support other styling languages or CSS-in-JS libraries using the [`customSyntax` option](../user-guide/usage/options.md#customSyntax)
+Custom syntaxes are [PostCSS syntaxes](https://github.com/postcss/postcss#syntaxes) written by the community to support other styling languages and CSS-in-JS libraries using the [`customSyntax` option](../user-guide/usage/options.md#customSyntax)
```
ybiquitous, line 3: [trivial] Is a period at the end of the sentence missing?

```suggestion
Custom syntaxes are [PostCSS syntaxes](https://github.com/postcss/postcss#syntaxes) written by the community to support other styling languages and CSS-in-JS libraries using the [`customSyntax` option](../user-guide/usage/options.md#customSyntax).
```

docs/developer-guide/syntaxes.md
```diff
@@ -1,33 +1,42 @@
...
 
 To write one, familiarize yourself with PostCSS's [how to write custom syntax](https://github.com/postcss/postcss/blob/main/docs/syntax.md) guide.
 
 Existing syntaxes that you can use for reference include:
 
 - [postcss-scss](https://github.com/postcss/postcss-scss)
 - [postcss-less](https://github.com/shellscape/postcss-less)
```
ybiquitous, line 10: [suggest] How about keeping consistency with the links (npmjs.com or github.com)?

```suggestion
- [postcss-scss](https://www.npmjs.com/package/postcss-scss)
- [postcss-less](https://www.npmjs.com/package/postcss-less)
```

ybiquitous (approved):
@jeddy3 Thanks for creating the pull request. I agree with your idea on #5786.

I've left some trivial suggestions, but this PR seems LGTM! 👍🏼

jeddy3:
@ybiquitous Thanks for the suggestions. I've committed them both.

ybiquitous (approved):
Thanks, LGTM! 👍🏼

PR #5829, opened 2022-01-13
Document freezing of stylistic rules

### LINKED ISSUE #5674: Document freezing of stylistic rules
My proposal may seem too loud to someone, but it is discussion. 

Our main problem is that we can never satisfy stylistic preferences. There will always be someone who wants an extra space or line break. Stylistic rules grow endlessly and create more and more problems. It's getting harder and harder to maintain. 

Eslint has https://eslint.org/blog/2020/05/changes-to-rules-policies and says:

> Stylistic rules are frozen - we won't be adding any more options to stylistic rules. We've learned that there's no way to satisfy everyone's personal preferences, and most of the rules already have a lot of difficult-to-understand options. Stylistic rules are those related to spacing, conventions, and generally anything that does not highlight an error or a better way to do something. Update 2021-01-29: We clarified in the README that we will still support newly-added ECMAScript features.

Honestly, I think we should do the same. 

Currently [`prettier`](https://github.com/prettier/prettier) do good job. I see no reason not to use `prettier`. And most developers have long seen the benefits of using formatters.

Yes, some stylistic rules can still be helpful - order of properties (we have a plugin for this), force using newlines in some places (we can reduce them to a minimum options). I think we can put together a list of style rules here and discuss them.

My opinion has always been:  want your preferred formatting - write your rule/plugin. 

I don't think we should bear the burden of maintaining due to some developers who want to have an extra space. I don't want to show disrespect for someone's preferences. 

I think it's time to say that stylistics rules are outdated. They were all created when we didn't have good formatters. Enough time has passed since that time and a lot has changed.

My proposal:

- Collect stylistics rules here and discussion them, maybe divide them on two sides - can be solved formatter and can't be not solved formatter.
- Create/update documentation about rules police about stylistics rules.
- Freeze stylistics rules.
- Close all issues related to stylistics rules.

@stylelint/contributors

jeddy3:

Closes #5674

The pull request touches on a number of our docs, including our vision and README.

It:
- emphasises that Stylelint is a linter (used to _avoid errors_ and  _enforce (non-stylistic) conventions_), and should be used alongside a pretty printer like Prettier
- updates our examples to avoid using stylistic rules
- juggles how we categorise rules in our list

Hopefully, these changes will help provide clear messaging for our users.


docs/user-guide/rules/combine.md
```diff
@@ -2,190 +2,88 @@
```
jeddy3, line 2: Only change in this file was moving the non-stylistic examples above the stylistic ones, i.e. leading with these two sections:

- "`*-allowed-list`, `*-disallowed-list`, `color-named` and applicable `*-no-*` rules"
- "`color-*` and `function-*` rules"

docs/user-guide/configure.md
```diff
@@ -226,10 +200,7 @@ The value of `"extends"` is a "locater" (or an array of "locaters") that is ulti
 
 Plugins are rules or sets of rules built by the community that support methodologies, toolsets, _non-standard_ CSS features, or very specific use cases.
 
-Popular plugin packs include:
-
-- [stylelint-order](https://github.com/hudochenkov/stylelint-order) - specify the ordering of things, e.g. properties within declaration blocks
-- [stylelint-scss](https://github.com/kristerkari/stylelint-scss) - enforce a wide variety of linting rules for SCSS-like syntax
+For example, [stylelint-scss](https://github.com/kristerkari/stylelint-scss) is a popular plugin pack that enforces a wide variety of linting rules for SCSS-like syntax.
```
ybiquitous, line 203: [trivial] We now can use the `stylelint-scss` organization! 

```suggestion
For example, [stylelint-scss](https://github.com/stylelint-scss/stylelint-scss) is a popular plugin pack that enforces a wide variety of linting rules for SCSS-like syntax.
```

docs/user-guide/get-started.md
```diff
@@ -55,6 +63,18 @@ npx stylelint "**/*.scss"
...
 This config includes the [postcss-scss syntax](https://github.com/postcss/postcss-scss), configures the [built-in rules](../user-guide/rules/list.md) for SCSS, and includes the [stylelint-scss plugin](https://www.npmjs.com/package/stylelint-scss) (a collection of rules specific to SCSS).
 
+If you use Prettier alongside Stylelint, you should also their [SCSS shared config](https://github.com/prettier/stylelint-config-prettier-scss):
+
+```json
+{
+  "extends": [
+    "stylelint-config-standard-scss",
+    "stylelint-config-prettier",
+    "stylelint-config-prettier-scss"
```
ybiquitous, line 73: [suggest] `stylelint-config-prettier` may be unnecessary, since `stylelint-config-prettier-scss` internally extends it.

```suggestion
    "stylelint-config-prettier-scss"
```

See https://github.com/prettier/stylelint-config-prettier-scss/blob/v0.0.1/src/index.js#L4

(BTW, I first learned about `stylelint-config-prettier-scss` here. 😄 )

docs/user-guide/get-started.md
```diff
@@ -55,6 +63,18 @@ npx stylelint "**/*.scss"
 
 This config includes the [postcss-scss syntax](https://github.com/postcss/postcss-scss), configures the [built-in rules](../user-guide/rules/list.md) for SCSS, and includes the [stylelint-scss plugin](https://www.npmjs.com/package/stylelint-scss) (a collection of rules specific to SCSS).
 
+If you use Prettier alongside Stylelint, you should also their [SCSS shared config](https://github.com/prettier/stylelint-config-prettier-scss):
```
ybiquitous, line 66: [typo] missing verb?

```suggestion
If you use Prettier alongside Stylelint, you should use also their [SCSS shared config](https://github.com/prettier/stylelint-config-prettier-scss):
```

docs/user-guide/rules/about.md
```diff
@@ -61,39 +61,39 @@ The first part describes what [_thing_](http://apps.workflower.fi/vocabs/css/en)
...
 ```
 
-_Rules are named to encourage explicit, rather than implicit, options._ For example, `color-hex-case: "upper"|"lower"` rather than `color-hex-uppercase: "always"|"never"`. As `color-hex-uppercase: "never"` _implies_ always lowercase, whereas `color-hex-case: "lower"` makes it _explicit_.
+_Rules are named to encourage explicit, rather than implicit, options._ For example, `font-weight-notation: "numeric"|"named-where-possible"` rather than `font-weight-numeric: "always"|"never"`. As `font-weight-named: "never"` _implies_ always numeric, whereas `font-weight-notation: "numeric"` makes it _explicit_.
 
 ### No rules
 
 Most rules require _or_ disallow something.
 
 For example, whether numbers _must_ or _must not_ have a leading zero:
```
ybiquitous, line 86: [suggest] It seems necessary to change also this sentence: ⬇️ 

```suggestion
For example, whether colors _must_ or _must not_ be named:
```

docs/user-guide/rules/about.md
```diff
@@ -61,39 +61,39 @@ The first part describes what [_thing_](http://apps.workflower.fi/vocabs/css/en)
...
 
-- `number-leading-zero`: `string - "always"|"never"`
-  - `"always"` - there _must always_ be a leading zero
-  - `"never"` - there _must never_ be a leading zero
+- `color-named`: `string - "always-where-possible"|"never"`
+  - `"always"` - there _must always (where possible)_ be named colors
+  - `"never"` - there _must never_ be named colors
```
ybiquitous, line 90: [suggest] It seems better to keep consistency with the sentence above ("For example, whether ..."):

```suggestion
  - `"always-where-possible"` - colors _must always (where possible)_ be named
  - `"never"` - colors _must never_ be named
```

ybiquitous (approved):
@jeddy3 This is great work! LGTM! 👏🏼

PR #5835, opened 2022-01-14
Add links to terminal output

### LINKED ISSUE #3823: Add link to the online docs in formatter output

Easy reach of full documentation when encountering a violation.

A link to the relevant online documentation directly from the formatter output (verbose seems a good candidate, but a new one is cool too).

Something like:

image

ybiquitous:

Closes #3823

This change outputs links of files or rule URLs on terminal via the `terminal-link` package.
The `string` and `verbose` formatters are only supported.

Note that it is needed to install `terminal-link@2` because `terminal-link@3` is pure ESM:

```shell
npm i terminal-link@2
```

See also <https://github.com/sindresorhus/terminal-link>

### Demo

On my local machine's terminal (macOS; iTerm2):

https://user-images.githubusercontent.com/473530/149568712-667dcbd2-00b2-410b-817e-4c59974e9af0.mov


lib/formatters/stringFormatter.js
```diff
@@ -83,9 +84,14 @@ function invalidOptionsFormatter(results) {
...
-	if (fromValue.startsWith('<')) return fromValue;
+	if (fromValue.startsWith('<')) {
+		return underline(fromValue);
+	}
+
+	const filePath = path.relative(cwd, fromValue).split(path.sep).join('/');
+	const fileUrl = `file://${fromValue}`;
 
-	return path.relative(cwd, fromValue).split(path.sep).join('/');
+	return terminalLink(filePath, fileUrl, { fallback: false });
```
ybiquitous, line 94: [note] If not disabling `fallback`, file paths would be output into an unsupported terminal like this:

```console
$ bin/stylelint.js /tmp/*.css --config /tmp/.stylelintrc.yml > out.txt

$ cat out.txt

../../../../../tmp/a.css (file:///tmp/a.css)
 1:2  ✖  Unexpected empty block  block-no-empty

../../../../../tmp/b.css (file:///tmp/b.css)
 1:2  ✖  Unexpected empty block  block-no-empty
```

This may be a breaking change.

hudochenkov: What do you think if we disable fallback and use https://github.com/jamestalmage/supports-hyperlinks (dependency of terminal-links) to add hyperlinks if they supported?

ybiquitous: @hudochenkov Thanks for the feedback!
We can use directly `supports-hyperlinks` instead of `terminal-link`!
If doing so, we need to continue to use `ansiEscapes.link()` or write a similar logic by ourselves.

E.g.

```js
function ourTerminalLink(text, url) {
  if (supportsHyperlinks.stdout) {
    return ansiEscapes.link(text, url)
    // or 
    // return ourAnsiEscapeLink(text, url)
  } else {
    return text
  }
}
```

`ansiEscapes.link()` is a tiny utility, so the idea of writing similar logic sounds good, too. (we don't need to depend on old versions of the packages)

- https://github.com/sindresorhus/terminal-link/blob/a925ecd9f2e6975505d801b5e352a05a9ff5921a/index.js#L14
- https://github.com/sindresorhus/ansi-escapes/blob/2b3b59c56ff77a2afdee946bff96f1779d10d775/index.js#L92-L107

Or, it seems good to use the `hyperlinker` package instead of `ansiEscapes.link()`.

https://github.com/jamestalmage/hyperlinker

What do you think?

lib/formatters/verboseFormatter.js
```diff
@@ -83,3 +84,30 @@ function groupBy(array, keyFn) {
...
+
+/**
+ * @param {string} rule
+ * @returns {string}
+ */
+function ruleLink(rule) {
+	// Plugin rules are unsupported.
+	if (rule.includes('/')) {
+		return rule;
+	}
```
ybiquitous, line 108: [note] It is a bit disappointing not to support plugin rules.
We may need to add a rule’s metadata (rule doc URL etc.) in the future. 🤔 
(but out of the scope of this PR)

See https://eslint.org/docs/developer-guide/working-with-rules#rule-basics

lib/formatters/stringFormatter.js
jeddy3: > ansiEscapes.link() is a tiny utility, so the idea of writing similar logic sounds good, too. (we don't need to depend on old versions of the packages)

This sounds good to me.

lib/formatters/verboseFormatter.js
ybiquitous: I've opened a new issue #5842 to discuss rule metadata.

types/supports-hyperlinks/index.d.ts
```diff
@@ -0,0 +1,8 @@
+declare module 'supports-hyperlinks' {
```
ybiquitous, line 1: [note] `supports-hyperlinks` doesn't have type definitions (even on `@types/`).

https://www.npmjs.com/package/supports-hyperlinks

ybiquitous:
Thanks for the feedback! Ready to review. 🙆🏼‍♂️

types/supports-hyperlinks/index.d.ts
mattxwang: Curious, is this something that would be helpful to contribute upstream? Looking at the package, it doesn't seem too complicated - maybe I can give it a shot?

ybiquitous: @mattxwang That would be great! 👍🏼

lib/formatters/__tests__/verboseFormatter.test.js
```diff
@@ -261,4 +261,74 @@ path/to/file.css
...
+					},
+				],
+				deprecations: [],
+				invalidOptionWarnings: [],
+			},
+		];
+
+		const output = prepareFormatterOutput(results, verboseFormatter);
+
+		expect(output).toBe(
```
jeddy3, line 286: Minor nit, can we use `const { stripIndent } = require('common-tags');` here and below to be consistent with how we mostly do it else where?

jeddy3 (commented):
Really good stuff!

I've requested a very minor formatting change.

lib/formatters/__tests__/verboseFormatter.test.js
ybiquitous: Good idea! Fixed via e8429207.

I also fixed the existing test cases in the module for consistency.

jeddy3 (approved):
LGTM, thanks!

lib/formatters/verboseFormatter.js
ybiquitous: We could use `meta.url` if #5845 would be merged.

types/supports-hyperlinks/index.d.ts
mattxwang: Sounds good - I can try to take a crack at it next weekend; probably shouldn't prevent this PR from being merged!

jeddy3:
- Added: hyperlinks to terminate output ([#5835](https://github.com/stylelint/stylelint/pull/5835)).

ybiquitous:
@jeddy3 This might be a typo? (Please ignore it if not a typo)

```diff
-to terminate output
+to terminal output
```

https://github.com/stylelint/stylelint/blob/2ae0c4cfd91dd226da95b6fe2a8c1273bf7477d5/CHANGELOG.md#L7

types/supports-hyperlinks/index.d.ts
mattxwang: Have created https://github.com/jamestalmage/supports-hyperlinks/issues/15!

mattxwang: Looks like this may have been resolved with https://github.com/jamestalmage/supports-hyperlinks/pull/21?

ybiquitous: @mattxwang Thanks for sharing the news! We can delete our custom type definitions with [`supports-hyperlinks@3.0.0`](https://github.com/jamestalmage/supports-hyperlinks/releases/tag/v3.0.0).

PR #5847, opened 2022-01-19
Fix typo in "About rules" doc

ybiquitous:

This change aims to keep consistency for the use of a rule's primary option type.


jeddy3 (approved):
LGTM, thanks!

PR #5849, opened 2022-01-19
Add option: ["camelCaseSvgKeywords"] to value-keyword-case #4884

### LINKED ISSUE #4884: value-keyword-case: `currentcolor`should be acceptable
> Clearly describe the bug

`currentcolor` (all lowercase) should be accepted. CSS Color Module level 4 defines it like that.

https://www.w3.org/TR/css-color-4/#currentcolor-color

Some historical context from above link:

> it was originally introduced in SVG as a special attribute value spelled "currentColor", rather than a CSS value. Only later did CSS pick it up, at which point the capitalization stopped mattering, as CSS keywords are case-insensitive.

`value-keyword-case`

```css
a {
  color: currentcolor;
}
```

```json
{
  "rules": {
    "value-keyword-case": "lower"
  }
}
```

e.g. `13.6.1`

CLI

No

No error.

```shell
test.css
 23:43  ✖  Expected "currentcolor" to be "currentColor"   value-keyword-case
```

kawaguchi1102:

Closes #4884

Following the discussion in the issue thread, I have added a rule option.
I hope this will satisfy your request.

### DIFF
--- a/lib/rules/value-keyword-case/README.md
+++ b/lib/rules/value-keyword-case/README.md
@@ -318,3 +318,27 @@ a {
   color: Foo(--camelCase);
 }
 ```
+
+### `camelCaseSvgKeywords: true | false` (default: `false`)
+
+If `true`, this rule will not check the camel case keyword.
+
+For example with `true`:
+
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+a {
+  color: currentcolor;
+}
+```
+
+The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+a {
+  color: currentColor;
+}
+```
--- a/lib/rules/value-keyword-case/__tests__/index.js
+++ b/lib/rules/value-keyword-case/__tests__/index.js
@@ -2315,3 +2315,53 @@ testRule({
 		},
 	],
 });
+
+testRule({
+	ruleName,
+	config: ['lower', { camelCaseSvgKeywords: true }],
+	fix: true,
+
+	accept: [
+		{
+			code: 'a { color: currentcolor; }',
+		},
+		{
+			code: 'a { color: lightseagreen; }',
+		},
+	],
+
+	reject: [
+		{
+			code: 'a { color: currentColor; }',
+			fixed: 'a { color: currentcolor; }',
+			message: messages.expected('currentColor', 'currentcolor'),
+			line: 1,
+			column: 12,
+		},
+	],
+});
+
+testRule({
+	ruleName,
+	config: ['upper', { camelCaseSvgKeywords: true }],
+	fix: true,
+
+	accept: [
+		{
+			code: 'a { color: CURRENTCOLOR; }',
+		},
+		{
+			code: 'a { color: LIGHTSEAGREEN; }',
+		},
+	],
+
+	reject: [
+		{
+			code: 'a { color: currentColor; }',
+			fixed: 'a { color: CURRENTCOLOR; }',
+			message: messages.expected('currentColor', 'CURRENTCOLOR'),
+			line: 1,
+			column: 12,
+		},
+	],
+});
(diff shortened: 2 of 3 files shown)


lib/rules/value-keyword-case/README.md
```diff
@@ -318,3 +318,27 @@ a {
...
 }
 ```
+
+### `camelCaseSvgKeywords: true | false` (default: `false`)
+
+If `true`, this rule will not check the camel case keyword.
+
+For example with `true`:
+
```
ybiquitous, line 328:
```suggestion
The following pattern is _not_ considered a problem:
```

lib/rules/value-keyword-case/README.md
```diff
@@ -318,3 +318,27 @@ a {
...
+
+<!-- prettier-ignore -->
+```css
+a {
+  color: currentcolor;
+}
+```
+
+The following patterns are considered problems:
```
ybiquitous, line 337:
```suggestion
The following pattern is considered a problem:
```

ybiquitous (commented):
@kawaguchi1102 Thanks for creating the pull request!

[ask] @jeddy3 said on <https://github.com/stylelint/stylelint/issues/4884#issuecomment-671981123>:

> We change the default behaviour so that the primary option (`lower` or `upper`) is used for every keyword, and add a `camelCaseSvgKeywords: true` option for users who want the existing behaviour.

so, when `camelCaseSvgKeywords: false` (this is the changed default behavior?), should `currentcolor` be always correct, not `currentColor`?

jeddy3:
> so, when `camelCaseSvgKeywords: false` (this is the changed default behavior?), should `currentcolor` be always correct, not `currentColor`?

That's correct. This pull request should also fix the _false negatives_, i.e. `currentColor`, `optimizeLegibility` etc should be lowercase if `"lower"` is the primary option and if `camelCaseSvgKeywords` is set to false or not set at all.

lib/rules/value-keyword-case/README.md
```diff
@@ -318,3 +318,27 @@ a {
   color: Foo(--camelCase);
 }
 ```
+
+### `camelCaseSvgKeywords: true | false` (default: `false`)
+
+If `true`, this rule will not check the camel case keyword.
```
jeddy3, line 324:
```suggestion
If `true`, this rule expects camelCase SVG keywords regardless of the primary option
```

lib/rules/value-keyword-case/README.md
```diff
@@ -318,3 +318,27 @@ a {
...
+
+For example with `true`:
+
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+a {
+  color: currentcolor;
```
jeddy3, line 333:
```suggestion
  color: currentColor;
```

lib/rules/value-keyword-case/README.md
```diff
@@ -318,3 +318,27 @@ a {
...
+}
+```
+
+The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+a {
+  color: currentColor;
```
jeddy3, line 342:
```suggestion
  color: currentcolor;
```

jeddy3 (changes requested):
@kawaguchi1102 Thanks for the pull request.

As @ybiquitous implied, the logic needs inverting i.e. when the `camelCaseSvgKeywords` option is set to `true` the SVG keywords should be camel case regardless of the primary option. When the it is set to `false` or not set, then the primary option, e.g. `lower` or `upper`, should be respected.

kawaguchi1102:
Thank you for pointing out the mistake.
I was mistaken in my assumption.

I will fix this issue.

jeddy3 (approved):
Turns out the SVG keywords were only camelCase for the `"lower"` primary. I've aligned this new option with that, as well as patching up the logic and tests.

> If `true`, this rule expects SVG keywords to be camel case when the primary option is `"lower"`.

LGTM.

ybiquitous (approved):
LGTM👍

jeddy3:
> - Added: `camelCaseSvgKeywords` to `value-keyword-case` - use this option if you want legacy camel case SVG keywords like `currentColor` ([#5849](https://github.com/stylelint/stylelint/pull/5849)).

> - Fixed: `value-keyword-case` false negatives for SVG keywords like `currentcolor` ([#5849](https://github.com/stylelint/stylelint/pull/5849)).

PR #5852, opened 2022-01-20
Fix false positives for revert and revert-layer in font-family-no-missing-generic-family-keyword

### LINKED ISSUE #5850: Fix false positives for revert and revert-layer in font-family-no-missing-generic-family-keyword

We ran into this by setting:

```
.example {
  font-family: revert;
}
```

I see in the docs that `inherit` is allowed, but there are several other 'global keywords' that should also have exceptions:

- initial
- unset
- revert
- revert-layer

```yaml
rules:
  font-family-no-missing-generic-family-keyword: true
```

CLI with `stylelint '**/*.scss'`

`14.2.0`

The global CSS keywords should not throw an error expecting a generic font-family.

### What actually happened?

A lint error was reported.

_No response_

### Proposal to fix the bug

Exempt all global CSS keywords from needing a generic font-family

treemmett:

Closes #5850

No, it's self-explanatory.

### DIFF
--- a/lib/reference/keywordSets.js
+++ b/lib/reference/keywordSets.js
@@ -65,7 +65,7 @@ keywordSets.camelCaseFunctionNames = new Set([
 	'skewY',
 ]);
 
-keywordSets.basicKeywords = new Set(['initial', 'inherit', 'unset']);
+keywordSets.basicKeywords = new Set(['initial', 'inherit', 'revert', 'revert-layer', 'unset']);
 
 keywordSets.systemFontValues = uniteSets(keywordSets.basicKeywords, [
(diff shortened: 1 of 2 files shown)


ybiquitous (approved):
@treemmett Thank you! LGTM 👍🏼

jeddy3 (approved):
Thank you!

jeddy3:
Changelog entry added:

> - Fixed: `font-family-no-missing-generic-family-keyword` false positives for `revert` and `revert-layer` ([#5852](https://github.com/stylelint/stylelint/pull/5852)).

PR #5856, opened 2022-01-20
Refactor to improve types for `selector-max-*` rules

ybiquitous:

Part of #4496

- Improve the code a bit for type safety.

### DIFF
--- a/lib/rules/selector-max-pseudo-class/index.js
+++ b/lib/rules/selector-max-pseudo-class/index.js
@@ -1,5 +1,3 @@
-// @ts-nocheck
-
 'use strict';
 
 const isContextFunctionalPseudoClass = require('../../utils/isContextFunctionalPseudoClass');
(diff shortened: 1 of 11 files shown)


lib/rules/selector-max-pseudo-class/index.js
```diff
@@ -71,14 +76,14 @@ function rule(max) {
 
 			for (const selector of ruleNode.selectors) {
 				for (const resolvedSelector of resolvedNestedSelector(selector, ruleNode)) {
-					parseSelector(resolvedSelector, result, rule, (selectorTree) => {
+					parseSelector(resolvedSelector, result, ruleNode, (selectorTree) => {
```
ybiquitous, line 79: [note] This is a potential bug, but it's an edge case. Because `rule` (`node) is used only when a parse error occurs.

https://github.com/stylelint/stylelint/blob/6df0c6bbd6cb59e0f2027b3a21609bb39cf1bacf/lib/utils/parseSelector.js#L15-L17

lib/rules/selector-max-type/index.js
```diff
@@ -85,16 +88,18 @@ function rule(max, options) {
 					return total;
 				}
 
-				return total + (childNode.type === 'tag');
+				return childNode.type === 'tag' ? total + 1 : total;
```
ybiquitous, line 91: [note] `Number(true)` is `1`, and `Number(false)` is `0`.

jeddy3 (approved):
Excellent stuff, thanks!

PR #5865, opened 2022-01-25
Add function-no-unknown

### LINKED ISSUE #5853: Add function-no-unknown

I'm looking for a list of known CSS functions in order to catch typo errors.

see https://github.com/stylelint-scss/stylelint-scss/issues/581#issuecomment-1017484418

1. Stylelint could export a list of known CSS functions that could be imported inside stylelint.config.js
2. Stylelint could provide a rule `function-no-unknown` which throws errors for unknown functions

gnuletik:

Closes #5853

"No, it's self-explanatory."

### DIFF
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -120,6 +120,7 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 
 - [`function-allowed-list`](../../../lib/rules/function-allowed-list/README.md): Specify a list of allowed functions.
 - [`function-disallowed-list`](../../../lib/rules/function-disallowed-list/README.md): Specify a list of disallowed functions.
+- [`function-no-unknown`](../../../lib/rules/function-no-unknown/README.md): Disallow unknown functions.
 - [`function-url-no-scheme-relative`](../../../lib/rules/function-url-no-scheme-relative/README.md): Disallow scheme-relative urls.
 - [`function-url-quotes`](../../../lib/rules/function-url-quotes/README.md): Require or disallow quotes for urls.
 - [`function-url-scheme-allowed-list`](../../../lib/rules/function-url-scheme-allowed-list/README.md): Specify a list of allowed URL schemes.
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -28,6 +28,7 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 
 - [`function-calc-no-unspaced-operator`](../../../lib/rules/function-calc-no-unspaced-operator/README.md): Disallow an unspaced operator within `calc` functions.
 - [`function-linear-gradient-no-nonstandard-direction`](../../../lib/rules/function-linear-gradient-no-nonstandard-direction/README.md): Disallow direction values in `linear-gradient()` calls that are not valid according to the [standard syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient#Syntax).
+- [`function-no-unknown`](../../../lib/rules/function-no-unknown/README.md): Disallow unknown functions.
 
 ### String
 
@@ -120,7 +121,6 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 
 - [`function-allowed-list`](../../../lib/rules/function-allowed-list/README.md): Specify a list of allowed functions.
 - [`function-disallowed-list`](../../../lib/rules/function-disallowed-list/README.md): Specify a list of disallowed functions.
-- [`function-no-unknown`](../../../lib/rules/function-no-unknown/README.md): Disallow unknown functions.
 - [`function-url-no-scheme-relative`](../../../lib/rules/function-url-no-scheme-relative/README.md): Disallow scheme-relative urls.
 - [`function-url-quotes`](../../../lib/rules/function-url-quotes/README.md): Require or disallow quotes for urls.
 - [`function-url-scheme-allowed-list`](../../../lib/rules/function-url-scheme-allowed-list/README.md): Specify a list of allowed URL schemes.
--- a/lib/rules/function-no-unknown/README.md
+++ b/lib/rules/function-no-unknown/README.md
@@ -1,6 +1,15 @@
 # function-no-unknown
 
-This rule considers functions defined in the CSS Specificationsto be known.
+Disallow unknown functions.
+
+<!-- prettier-ignore -->
+```css
+a { transform: unknown(1); }
+/**            ↑
+ * Functions like this */
+```
+
+This rule considers functions defined in the CSS Specifications to be known.
 
 This rule ignores double-dashed custom functions, e.g. --custom-function().
 
@@ -12,7 +21,7 @@ The following patterns are considered problems:
 
 <!-- prettier-ignore -->
 ```css
-a { transform: doesntexist(1); }
+a { transform: unknown(1); }
 ```
 
 The following patterns are _not_ considered problems:
--- a/lib/rules/function-no-unknown/README.md
+++ b/lib/rules/function-no-unknown/README.md
@@ -11,7 +11,7 @@ a { transform: unknown(1); }
 
 This rule considers functions defined in the CSS Specifications to be known.
 
-This rule ignores double-dashed custom functions, e.g. --custom-function().
+This rule ignores double-dashed custom functions, e.g. `--custom-function()`.
 
 ## Options
 
--- a/lib/rules/function-no-unknown/README.md
+++ b/lib/rules/function-no-unknown/README.md
@@ -1,0 +1,28 @@
+# function-no-unknown
+
+This rule considers functions defined in the CSS Specificationsto be known.
+
+This rule ignores double-dashed custom functions, e.g. --custom-function().
+
+## Options
+
+### `true`
+
+The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+a { transform: doesntexist(1); }
+```
+
+The following patterns are _not_ considered problems:
+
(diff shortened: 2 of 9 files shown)


gnuletik:
There's an issue on Windows builds when loading [`css-functions-list`](https://www.npmjs.com/package/css-functions-list):
>     ENOENT: no such file or directory, open '/D:/a/stylelint/stylelint/node_modules/css-functions-list/cjs/index.json'

Does someone use Windows and have an idea to fix it?

ybiquitous:
@gnuletik How about using `URL` as below?

```js
fs.readFileSync(new URL("file://" + listPath))
```

lib/rules/function-no-unknown/index.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+const fs = require('fs');
+const valueParser = require('postcss-value-parser');
+const functionsListPath = /** @type {string} */ (
+	/** @type {unknown} */ (require('css-functions-list'))
```
alexander-akait, line 6: This package is too old and doesn't contain all functions

ybiquitous: Oh, do you know some alternative or another way?

ybiquitous: If the package is insufficient, I think we should contain a function name set internally like the math functions: 🤔 

https://github.com/stylelint/stylelint/blob/b0523d8e76929d8c7224f884b0b256c2657ce977/lib/reference/mathFunctions.js#L3

Any thoughts?

gnuletik: This package uses data from MDN.
I cloned it locally and run `npm run generate-list` and got 10 new functions in the diff.
I guess we can make a PR to update the package, unless you want to maintain it inside stylelint. Up to you :)

ybiquitous: @gnuletik Thanks for the feedback. Could you please open a PR to [niksy/css-functions-list](https://github.com/niksy/css-functions-list), please?
If the repository owner will maintain the package, we can go ahead.

If not, then we will maintain the function list by ourselves. I'm a bit concerned about our maintenance resources... 🤔

gnuletik: > @gnuletik How about using `URL` as below?
> 
> ```js
> fs.readFileSync(new URL("file://" + listPath))
> ```

I just made the change. I need approval to run the CI.

gnuletik:
> @gnuletik How about using `URL` as below?
> 
> ```js
> fs.readFileSync(new URL("file://" + listPath))
> ```

I just made the change. I need approval to run the CI.

lib/rules/function-no-unknown/index.js
alexander-akait: I think we can have list here, it is not hard to maintenance

gnuletik: Done :) https://github.com/niksy/css-functions-list/pull/1

docs/user-guide/rules/list.md
```diff
@@ -120,6 +120,7 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 
 - [`function-allowed-list`](../../../lib/rules/function-allowed-list/README.md): Specify a list of allowed functions.
 - [`function-disallowed-list`](../../../lib/rules/function-disallowed-list/README.md): Specify a list of disallowed functions.
+- [`function-no-unknown`](../../../lib/rules/function-no-unknown/README.md): Disallow unknown functions.
```
ybiquitous, line 123: Should the rule be in the "Avoid errors" section?

lib/rules/function-no-unknown/index.js
niksy: New version has been pushed! 😉

ybiquitous: @niksy Thank you so much! 😊

lib/rules/function-no-unknown/index.js
```diff
@@ -0,0 +1,72 @@
...
+		root.walkDecls((decl) => {
+			const { value } = decl;
+
+			valueParser(value).walk((node) => {
+				if (node.type !== 'function') {
+					return;
+				}
+
+				// custom function: https://github.com/w3c/css-houdini-drafts/issues/1007
+				if (node.value.startsWith('--')) {
```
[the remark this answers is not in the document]
gnuletik: Yep! Done.

docs/user-guide/rules/list.md
gnuletik: That makes sense! Done.

lib/utils/__tests__/isCustomFunction.test.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const isCustomFunction = require('../isCustomFunction');
+
+it('isCustomMediaQuery', () => {
```
ybiquitous, line 5: [typo]

```suggestion
it('isCustomFunction', () => {
```

package.json
```diff
@@ -111,6 +111,7 @@
     "balanced-match": "^2.0.0",
     "colord": "^2.9.2",
     "cosmiconfig": "^7.0.1",
+    "css-functions-list": "^2.0.0",
```
ybiquitous, line 114: [suggest] Bump up, please?

```suggestion
    "css-functions-list": "^3.0.0",
```

https://github.com/niksy/css-functions-list/releases/tag/v3.0.0

ybiquitous (commented):
@gnuletik Thanks, almost good! 👍🏼

ybiquitous (approved):
Thanks, LGTM 👍🏼

jeddy3 (changes requested):
Thanks for the pull request.

I've requested some changes. Let's also add an accept test for `TRANSFORM` and an reject one for `UNKNOWN`.

gnuletik:
Thanks @jeddy3 !
I applied your changes, added some test for the `if (!isStandardSyntaxFunction) { ... }` branch and rebased.

lib/rules/function-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,48 @@
...
+	ruleName,
+	config: false,
+	skipBasicChecks: true,
+
+	accept: [
+		{
+			code: 'a { transform: shouldskip(4); }',
+		},
+	],
+});
```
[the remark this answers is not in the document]
gnuletik: However, this removal generates a test coverage warning : https://github.com/stylelint/stylelint/pull/5865/checks?check_run_id=5069375681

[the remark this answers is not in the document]
jeddy3: I think it's ok to have that warning.

lib/rules/function-no-unknown/README.md
```diff
@@ -0,0 +1,37 @@
...
+<!-- prettier-ignore -->
+```css
+a { transform: unknown(1); }
+/**            ↑
+ * Functions like this */
+```
+
+This rule considers functions defined in the CSS Specifications to be known.
+
+This rule ignores double-dashed custom functions, e.g. --custom-function().
```
jeddy3, line 14:
```suggestion
This rule ignores double-dashed custom functions, e.g. `--custom-function()`.
```

Minor nit.

lib/rules/function-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,39 @@
...
+		},
+		{
+			code: 'a { transform: --custom-function(1px); }',
+		},
+		{
+			code: 'a { transform: scale(0.5) translate(-100%, -100%); }',
+		},
+		{
+			code: 'a { $list: (list) }',
+		},
```
jeddy3, line 22: Let's move this SCSS into its own `testRule` that uses the `postcss-scss` syntax as we strive to keep standard and non-standard syntax separate.

jeddy3 (changes requested):

I've requested a couple more.

Let's also add an:
- accept test for `TRANSFORM` 
- an reject one for `UNKNOWN`

So that we test that the rule is case insensitive.

lib/rules/function-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,57 @@
...
+			line: 1,
+			column: 23,
+		},
+	],
+});
+
+testRule({
+	ruleName,
+	config: true,
+	skipBasicChecks: true,
```
jeddy3, line 50:
```suggestion
	ruleName,
	config: true,
	customSyntax: "postcss-scss",
	skipBasicChecks: true,
```

Let's specify the SCSS syntax (even though the default parser can parse the construct).

jeddy3 (changes requested):

One last suggestion and I think we're ready to merge.

jeddy3 (approved):

jeddy3:
Changelog entry:

> - Added: `function-no-unknown` rule ([#5865](https://github.com/stylelint/stylelint/pull/5865)).

torhovland:
This is now starting to complain when I have

```
@use 'sass:math';

.service-detail-header-logo__active-icon {
  top: math.div(-$pulse-size, 4);
  ...
```

```
Unexpected unknown function "math.div"  function-no-unknown
```

ybiquitous:
@torhovland It is an issue of `stylelint-scss`: ⬇️ 

https://github.com/stylelint-scss/stylelint-scss/issues/581

ybiquitous:
As a workaround, you can [disable the rule](https://stylelint.io/user-guide/configure/#overrides) like this:

```json
{
  "overrides": [
    {
      "files": ["*.scss", "**/*.scss"],
      "rules": {
        "function-no-unknown": null
      }
    }
  ]
}
```

kristerkari:
The SCSS version is now also available: 
https://github.com/stylelint-scss/stylelint-scss/releases/tag/v4.2.0

PR #5873, opened 2022-02-02
Document overrides with customSyntax

### LINKED ISSUE #5872: Document `overrides` and `customSyntax` working together

My project has both less and styled-components. I want to use stylelint to lint both. For styled-components, I use `"customSyntax": "postcss-jsx"` in the .stylelintrc. For less, I use` "customSyntax": "postcss-less"`. This works well for the corresponding file type. But I can't have stylelint to lint both file types since I can only specify one value for `customSyntax`.

perhaps allow specifying an array of `customSyntax` so that stylint can use different library to parse corresponding type?

jeddy3:

Closes #5872

No, it's self-explanatory.


docs/user-guide/configure.md
```diff
@@ -252,6 +252,10 @@ In your `.stylelintrc.json`:
     "alpha-value-notation": "number"
   },
   "overrides": [
+    {
+      "files": ["*.scss", "**/*.scss"],
```
ybiquitous, line 256: [ask] Are these patterns the same as below?

```suggestion
      "files": ["**/*.scss"],
```

ybiquitous (approved):
I've left a trivial comment, but LGTM 👍🏼

ybiquitous:
Now I've retried the CI jobs.

docs/user-guide/configure.md
jeddy3: I don't think that pattern will match files in the root.

See: 
https://github.com/ota-meshi/stylelint-config-html/blob/af14c2bfa55eab72461f7047c25cc4f842b19996/html.js#L23

ybiquitous: I got it. 👍🏼

PR #5876, opened 2022-02-03
Refactor to improve types for `selector-pseudo-*` rules

ybiquitous:

Part of #4496

- Improve the code a bit for type safety.


lib/rules/selector-pseudo-class-case/index.js
```diff
@@ -85,7 +84,7 @@ function rule(expectation, options, context) {
 				},
 			);
 
-			if (context.fix) {
+			if (context.fix && fixedSelector) {
```
jeddy3, line 87: 👍

jeddy3 (approved):
Thank you!

There can't be that many rules left to do now?

ybiquitous:
> There can't be that many rules left to do now?

Yes, about 20 rules left.
https://github.com/stylelint/stylelint/issues/4496#issuecomment-919337742

PR #5874, opened 2022-02-03
Downgrade npm to 8.3 as workaround

ybiquitous:
> Which issue, if any, is this issue related to?

None.

> Is there anything in the PR that needs further explanation?

Currently, our CI jobs for Windows fail due to npm@8.4 (see <https://github.com/npm/cli/issues/4341>). So this change downgrades to npm@8.3 as a workaround.

E.g. https://github.com/npm/cli/issues/4341

PR #5877, opened 2022-02-04
Revert "Downgrade npm to 8.3 as workaround"

ybiquitous:
Reverts stylelint/stylelint#5874

See:
- https://github.com/npm/cli/issues/4341
- https://github.com/actions/setup-node/issues/411


.github/workflows/testing.yml
```diff
@@ -32,10 +32,14 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: npm
 
-      - name: Install latest npm
+      # https://github.com/npm/cli/issues/4341
+      # https://github.com/actions/setup-node/issues/411
+      - name: Workaround for npm installation on Windows
+        if: ${{ matrix.os == 'windows-latest' }}
         run: npm install --global npm@8.3
```
ybiquitous, line 39: A workaround is still necessary for Windows. 😢

jeddy3 (approved):
LGTM.

This is a satisfactory workaround. Better to have one additional step on windows so that all OSes use the latest NPM, than have NPM pinned to an older version on Windows.

ybiquitous:
> This is a satisfactory workaround. Better to have one additional step on windows so that all OSes use the latest NPM, than have NPM pinned to an older version on Windows.

Thanks. I think so, too. 👍🏼

PR #5885, opened 2022-02-06
Fix typo / inconsistency in stylelint.rules API docs for plugins

thibaudcolas:
I think the snippet is understandable even with the typo, but worth fixing nonetheless.


ybiquitous (approved):
@thibaudcolas Thank you! 👍🏼

jeddy3 (approved):
Thank you.

PR #5896, opened 2022-02-08
Refactor requires of custom syntaxes in tests

jeddy3:

None

I figured we'd get this in while we've no conflicting pull requests.

In our tests, we sometimes require a custom syntax and in other times we use a string. This pull request uses strings consistently as they allow us to add self-contained `testRule`s for custom syntaxes.


ybiquitous (approved):
Nice! LGTM 👍🏼

PR #5901, opened 2022-02-09
Add `ignoreFunctions` option to `function-no-unknown`

### LINKED ISSUE #5900: Add `ignoreFunctions: []` to function-no-unknown

[function-no-unknown](https://stylelint.io/user-guide/rules/list/function-no-unknown) throws an error when using non-standard funcitons such as tailwindcss [theme](https://tailwindcss.com/docs/functions-and-directives#theme) function.

It would be nice to have a feature to ignore certain functions just like `ignoreAtRules` of [at-rule-no-unknown](https://stylelint.io/user-guide/rules/list/at-rule-no-unknown).

ybiquitous:

Closes #5900

No, it's self-explanatory.

### DIFF
--- a/lib/rules/function-no-unknown/README.md
+++ b/lib/rules/function-no-unknown/README.md
@@ -50,23 +50,6 @@ Given:
 ["theme", "/^foo-/"]
 ```
 
-The following patterns are considered problems:
-
-<!-- prettier-ignore -->
-```css
-a { transform: unknown(1); }
-```
-
-<!-- prettier-ignore -->
-```css
-a { transform: THEME(1); }
-```
-
-<!-- prettier-ignore -->
-```css
-a { transform: foo(1); }
-```
-
 The following patterns are _not_ considered problems:
 
 <!-- prettier-ignore -->
--- a/lib/rules/function-no-unknown/README.md
+++ b/lib/rules/function-no-unknown/README.md
@@ -35,3 +35,46 @@ a { transform: scale(1); }
 ```css
 a { transform: --custom-function(1); }
 ```
+
+## Optional secondary options
+
+### `ignoreFunctions: ["/regex/", /regex/, "non-regex"]`
+
+Ignore the specified functions.
+
+For example, with `true`.
+
+Given:
+
+```json
+["theme", "/^foo-/"]
+```
+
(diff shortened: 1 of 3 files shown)


lib/rules/function-no-unknown/README.md
```diff
@@ -35,3 +35,46 @@ a { transform: scale(1); }
...
+<!-- prettier-ignore -->
+```css
+a { transform: THEME(1); }
+```
+
+<!-- prettier-ignore -->
+```css
+a { transform: foo(1); }
+```
+
```
jeddy3, line 69: Let's remove this section.

For `ignore` options we usually just show the rule becoming more permissive in the docs.

lib/rules/function-no-unknown/index.js
```diff
@@ -23,14 +25,34 @@ const meta = {
...
 
+		/** @type {Array<string | RegExp>} */
+		const ignoreFunctions = (secondaryOptions && secondaryOptions.ignoreFunctions) || [];
+
+		/** @type {(funcName: string) => boolean} */
+		const isFunctionIgnored =
+			ignoreFunctions.length === 0
+				? () => false
+				: (funcName) => Boolean(matchesStringOrRegExp(funcName, ignoreFunctions));
+
```
jeddy3, line 55: If possible, let's use the [`optionsMatches`](https://github.com/stylelint/stylelint/blob/main/lib/utils/optionsMatches.js) util here.

jeddy3 (changes requested):
Thanks for getting started on this so quickly!

I've made two suggestions.

lib/rules/function-no-unknown/index.js
ybiquitous: Thanks! I forgot the util. 😅

jeddy3 (approved):
LGTM, thank you!

ybiquitous:
@jeddy3 I have no left for this pull request. Can we merge this?

jeddy3:
@ybiquitous Yes, go ahead.

I've changed the branch protection rules to require one approval.  It's a more sensible default. We can reserve two reviews for only the largest pull requests.

ybiquitous:
Thank you!

PR #5903, opened 2022-02-10
Refactor to replace `matchesStringOrRegExp` with `optionsMatches`

ybiquitous:

None, just a small refactoring.

`optionsMatches` makes the code simpler in some cases.


jeddy3 (approved):
LGTM, thanks!

I suspect these rules predate the creation of the `optionMatches` util, which is why they don't already use it.

Issue #3968, opened 2019-03-01
Add rule to prevent use of 3-4 part shorthand values


twig-openlearning:
> What is the problem you're trying to solve?

Prevent use of 3-4 length shorthand values as it can cause confusion for those less familiar with css.

eg.
```
margin: 1px 2px 3px 4px;
padding: 3px 4px 5px 6px;
border-width: 1px 2px 3px 4px;
```

Apologies if I didn't see it in the docs/issues, I'm not exactly sure what keywords to look for.

> What solution would you like to see?

A new configurable rule which limits the number of values allowed for properties which allow for 1-4 values.

eg. `new-shorthand-limit-rule: 2` (limits to 2 values - top/bottom and left/right)

hudochenkov:
Is it enough? https://stylelint.io/user-guide/rules/shorthand-property-no-redundant-values/

twig-openlearning:
I initially thought it would be, but if the values all differ slightly then it'll pass

It's more to be more explicit when declaring the values so others (who aren't as familiar with CSS) can see what it actually means at a glance.

ie. `padding: 1px 2px 3px;`

Is that 1px top/bottom and 2px left/3px right or 1px top/2px bottom and 3px left/right?

jeddy3:
Interesting proposal.

I originally thought that there might a case for adding a `shorthand-property-max-values` rule to core. However, I remember that not all shorthand properties accept up to 4 values e.g. `grid-gap` is a shorthand that accepts 1 or 2 values. As such, it feels like a problematic rule for core.

It also feels somewhat of an idiosyncratic pattern. However, you have two other options:

1. You can [create your own plugin](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/plugins.md) and [share that with the community](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/plugins.md). The plugin would expect:
   - 1 or 2 values for shorthand properties that accept up to 4 values e.g. `margin`
   - 1 value for shorthand properties that accept up to 2 values e.g. `grid-gap`
2. Use the built-in [`declaration-property-value-whitelist`](https://github.com/stylelint/stylelint/tree/master/lib/rules/declaration-property-value-whitelist) rule and configure it with a regex that expects a single value for all shorthand properties (or two values separated by whitespace for shorthand properties that accept up to 4 values).

twig-openlearning:
What exactly do you mean by "core"? (sorry not too familiar with how stylelint is coded)

[...]

PR #5920, opened 2022-02-18
Add declaration-property-max-values

### LINKED ISSUE #5435: Add declaration-property-max-values
Since the padding and margin shorthands are hard to understand for many people, it can be preferred to write it out fully.

Consider the following scenario:
margin: 2px 8px 3px;

Most people have no idea what this involves. Going for a syntax like this:
margin: 8px;
margin-top: 2px;
margin-bottom: 3px;

Would be preferred.

Rule could be to limit the amount of values padding, margin and border-width can have.

mattxwang:

Closes #5435, related to #3968.

As suggested by @jeddy3, I've scaffolded this off of `declaration-property-value-allowed-list`. I've also implemented a new util, `validateObjectWithProps()`, and written some tests.

This is my first time creating an entirely new rule, so I have three things I'd like to get some feedback on:

1. Is the documentation the correct level of conciseness?
2. In this PR, I haven't yet used the `validateOptions` utility function. How should I use it here, if at all?
3. My method for finding the number of values is finding the number of `word` nodes after running the value through the value parser. Is this the correct approach?


lib/rules/declaration-property-max-values/index.js
```diff
@@ -0,0 +1,56 @@
...
+		`Expected "${property}" to have no more than ${max} ${max === 1 ? 'value' : 'values'}`,
+});
+
+const meta = {
+	url: 'https://stylelint.io/user-guide/rules/list/declaration-property-max-values',
+};
+
+/** @type {import('stylelint').Rule<Record<string, Number>>} */
+const rule = (primary) => {
+	return (root, result) => {
```
ybiquitous, line 22: [suggest] I think `validateOptions` is necessary in almost all cases.

```js
const validOptions = validateOptions(...);
```

lib/rules/declaration-property-max-values/index.js
```diff
@@ -0,0 +1,56 @@
...
+const messages = ruleMessages(ruleName, {
+	rejected: (property, max) =>
+		`Expected "${property}" to have no more than ${max} ${max === 1 ? 'value' : 'values'}`,
+});
+
+const meta = {
+	url: 'https://stylelint.io/user-guide/rules/list/declaration-property-max-values',
+};
+
+/** @type {import('stylelint').Rule<Record<string, Number>>} */
```
ybiquitous, line 20: [suggest] The `number` primitive type is preferred over `Number`:

```suggestion
/** @type {import('stylelint').Rule<Record<string, number>>} */
```

lib/rules/declaration-property-max-values/index.js
```diff
@@ -0,0 +1,56 @@
...
+const rule = (primary) => {
+	return (root, result) => {
+		root.walkDecls((decl) => {
+			const prop = decl.prop;
+			const value = decl.value;
```
ybiquitous, line 25: [trivial] What about the following refactoring? (minor, though)

```suggestion
			const { prop, value } = decl;
```

lib/rules/declaration-property-max-values/index.js
```diff
@@ -0,0 +1,56 @@
...
+			const propLength = valueParser(value).nodes.filter((node) => node.type === 'word').length;
```
ybiquitous, line 26: [ask] In addition to `word`, `string` and `function` may be necessary. What do you think?

Ref: https://github.com/TrySound/postcss-value-parser#nodes

ybiquitous (changes requested):
@mattxwang Thank you for creating this pull request! I think the implementation is in the right direction. 👍🏼

lib/rules/declaration-property-max-values/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+If a property name is surrounded with `"/"` (e.g. `"/^margin/"`), it is interpreted as a regular expression. This allows, for example, easy targeting of shorthands: `/^margin/` will match `margin`, `margin-top`, `margin-inline`, etc.
+
+Given:
+
+```json
+{
+  "border": 1,
+  "/^margin/": 2,
+  "padding": 3,
```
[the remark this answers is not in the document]
jeddy3: Let's go with `border` and `/margin/`.

lib/rules/declaration-property-max-values/index.js
jeddy3: Yes, let's also add 1 accept and 1 reject test for custom props: e.g.

```css
margin: var(--foo) var(--bar);
```

lib/rules/declaration-property-max-values/README.md
```diff
@@ -0,0 +1,83 @@
...
+},
+```
+
+The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+div { margin: 1px 2px 3px; }
```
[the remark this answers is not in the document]
jeddy3: FYI, [conventions here](https://stylelint.io/developer-guide/rules#write-the-readme).

lib/rules/declaration-property-max-values/__tests__/index.js
```diff
@@ -0,0 +1,83 @@
...
+			message: messages.rejected('margin', 2),
+			line: 1,
+			column: 7,
+		},
+		{
+			code: 'div { margin-inline: 1px 2px 3px; }',
+			message: messages.rejected('margin-inline', 2),
+			line: 1,
+			column: 7,
+		},
```
jeddy3, line 57: Let's remove this as not valid CSS.

lib/rules/declaration-property-max-values/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+<!-- prettier-ignore -->
+```css
+div { margin-inline: 1px 2px; }
+```
+
+<!-- prettier-ignore -->
+```css
+div { margin: ; }
+```
```
jeddy3, line 73: Let's remove this example.

We want to docs to show the minimum amount of examples so that users get the gist of the rule, rather than demonstate edge cases.

lib/rules/declaration-property-max-values/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+<!-- prettier-ignore -->
+```css
+div { margin: 1px 2px; }
+```
+
+<!-- prettier-ignore -->
+```css
+div { margin: 1px 2px /* 3px */; }
+```
```
jeddy3, line 63: Let's remove this example.

lib/rules/declaration-property-max-values/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+<!-- prettier-ignore -->
+```css
+div { margin: 0; }
+```
+
+<!-- prettier-ignore -->
+```css
+div { margin: 1px 2px; }
+```
```
jeddy3, line 58: Let's remove this as similar to `margin-inline` below.

lib/rules/declaration-property-max-values/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+<!-- prettier-ignore -->
+```css
+div { margin: 1px 2px 3px 4px; }
+```
+
+<!-- prettier-ignore -->
+```css
+div { margin-inline: 1px 2px 3px; }
+```
```
jeddy3, line 36: Let's remove as invalid CSS.

jeddy3 (changes requested):
@mattxwang Looking good so far!

I've requested a few changes in addition to @ybiquitous'.

lib/rules/declaration-property-max-values/index.js
jeddy3: Should we just use `isPlainObject` introduced in https://github.com/stylelint/stylelint/pull/5922?

Or should be more specific, e.g. this rule expects an object whose values are numbers?

ybiquitous: I think `validateObjectWithArrayProps` may be helpful in this case. E.g.

```js
const validOptions = validateOptions(result, ruleName, {
	actual: primary,
	possible: [validateObjectWithArrayProps(isNumber)],
});
```

https://github.com/stylelint/stylelint/blob/f1ee719fe537bfa41b248df1bc7ae5528c153fff/lib/utils/validateObjectWithArrayProps.js#L22

jeddy3: As this rule expects a single number (rather than an array of them), we may need a new function similar to `validateObjectWithArrayProps`.

Shall we create `validateObjectWith()`, e.g.:

```js
const validOptions = validateOptions(result, ruleName, {
	actual: primary,
	possible: [validateObjectWith(isNumber)],
});
```

ybiquitous: Nice idea! 👍🏼

mattxwang:
Thanks for the in-depth feedback @jeddy3 and @ybiquitous, it's much-appreciated - I still have a lot to learn in terms of the codebase and best practices.

In general, I think I've addressed each point of feedback. I did have a chance to implement `validateObjectWithProps`, though I'm not sure if it's the intended behaviour from the above discussion - is this what you two were thinking? Did you want me to add the array functionality from `validateObjectWithArrayProps` as well?

And, anything else I should be mindful of?

ybiquitous:
@mattxwang I think your implementation of `validateObjectWithProps` is just what we want. 👍🏼

lib/utils/__tests__/validateObjectWithProps.test.js
```diff
@@ -0,0 +1,36 @@
...
+
+const validateObjectWithProps = require('../validateObjectWithProps');
+const { isNumber } = require('../validateTypes');
+
+describe('validateObjectWithProps', () => {
+	it('should return a function', () => {
+		expect(validateObjectWithProps((x) => x)).toBeInstanceOf(Function);
+	});
+
+	it('rejects non-objects', () => {
```
ybiquitous, line 11: [trivial] Typo?

```suggestion
	it('should reject non-objects', () => {
```

lib/utils/validateObjectWithProps.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+const { isPlainObject } = require('is-plain-object');
```
ybiquitous, line 3: [suggest] Please use `./validateTypes` instead of `is-plain-object`. The former improves the type inference.

```suggestion
const { isPlainObject } = require('./validateTypes');
```

lib/utils/validateObjectWithProps.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * @template T
+ * @typedef {(i: T) => boolean} Validator
+ */
+
+/**
+ * Check whether the variable is an object and all its properties agree with the provided validator
```
ybiquitous, line 11: [trivial] Typo.

```suggestion
 * Check whether the variable is an object and all its properties agree with the provided validator.
```

lib/utils/validateObjectWithProps.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * config = {
+ *   value1: 1,
+ *   value2: 2,
+ *   value3: 3,
+ * }
```
ybiquitous, line 17: [suggest] How about improving the example like this?

- Add `@example`.
- Bring it closer to practical use.

```suggestion
 * @example
 * config = {
 *   value1: 1,
 *   value2: 2,
 *   value3: 3,
 * };
 * validateObjectWithProps(isNumber)(config);
 * //=> true
 *
```

lib/utils/validateObjectWithProps.js
```diff
@@ -0,0 +1,30 @@
...
+ * @template T
+ * @param {Validator<T>} validator
```
ybiquitous, line 19: [suggest] How about making the `validator` param type simpler?

```suggestion
 * @param {(value: unknown) => boolean} validator
```

lib/utils/validateObjectWithProps.js
```diff
@@ -0,0 +1,30 @@
...
+ * @returns {(value: unknown) => boolean}
+ */
+module.exports = (validator) => (value) => {
+	if (!isPlainObject(value)) {
+		return false;
+	}
+
+	return Object.values(/** @type {Record<string, T>} */ (value)).every((item) => {
```
ybiquitous, line 27: [suggest] If using `./validateTypes`, we can remove the type cast:

```suggestion
	return Object.values(value).every((item) => {
```

jeddy3 (changes requested):
Thanks for making the changes! Looking good.

In addition to @ybiquitous' suggestion, just one more minor request from me

mattxwang:
Thanks for the suggestions @ybiquitous, didn't realize that we'd get better typing by importing from `./validateTypes`!

@jeddy3, I may be missing something, but what was your requested change? I don't see anything attached to your "requested changes" dialog box.

lib/rules/declaration-property-max-values/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+<!-- prettier-ignore -->
+```css
+a { margin: 1px 2px 3px; }
+```
+
+<!-- prettier-ignore -->
+```css
+a { margin: 1px 2px 3px 4px; }
```
ybiquitous, line 34: [trivial] It may be better if there would be a case matching `/^margin/` like `margin-inline`.

lib/rules/declaration-property-max-values/__tests__/index.js
```diff
@@ -0,0 +1,78 @@
...
+			line: 1,
+			column: 5,
+		},
+		{
+			code: 'a { border: 1px solid blue; }',
+			message: messages.rejected('border', 1),
+			line: 1,
+			column: 5,
+		},
+	],
```
ybiquitous, line 77: [trivial] It may be better if there would be a reject case matching `/^margin/` like `margin-inline`.

lib/rules/declaration-property-max-values/index.js
```diff
@@ -0,0 +1,71 @@
...
+const messages = ruleMessages(ruleName, {
+	rejected: (property, max) =>
+		`Expected "${property}" to have no more than ${max} ${max === 1 ? 'value' : 'values'}`,
+});
+
+const meta = {
+	url: 'https://stylelint.io/user-guide/rules/list/declaration-property-max-values',
+};
+
+const isValueNode = (/** @type {valueParser.Node} */ node) => {
```
ybiquitous, line 23: [trivial] The following may be more consistent with other code:

```suggestion
/**
 * @param {valueParser.Node} node
 */
const isValueNode = (node) => {
```

lib/rules/declaration-property-max-values/index.js
```diff
@@ -0,0 +1,71 @@
...
+		root.walkDecls((decl) => {
+			const validOptions = validateOptions(result, ruleName, {
+				actual: primary,
+				possible: [validateObjectWithProps(isNumber)],
+			});
+
+			if (!validOptions) {
+				return;
+			}
+
```
ybiquitous, line 39: [suggest] It's better to move `validateOptions` outside `walkDecls` for performance:

```suggestion
		const validOptions = validateOptions(result, ruleName, {
			actual: primary,
			possible: [validateObjectWithProps(isNumber)],
		});

		if (!validOptions) {
			return;
		}

		root.walkDecls((decl) => {
```

lib/rules/declaration-property-max-values/index.js
```diff
@@ -0,0 +1,71 @@
...
+			const max = primary[propKey];
+
+			if (propLength <= max) {
+				return;
+			}
+
+			report({
+				message: messages.rejected(prop, max),
+				node: decl,
```
ybiquitous, line 60: [ask] Which column position should we specify: property or value?

```suggestion
				node: decl,
				index: declarationValueIndex(decl),
```

property:

image

value:

image

ybiquitous (commented):
@mattxwang Thank you for addressing the reviews. Almost good. 👍🏼 

I've left more trivial suggestions and one question.

lib/rules/declaration-property-max-values/README.md
mattxwang: Hm, in a previous review comment @jeddy3 mentioned that this is not a great example since it is invalid CSS (ex looking at the [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline)), it looks like that `margin-inline` only takes one or two parameters.

I can circumvent this by setting `margin: 1` in the config - would that get the desired effect you're looking for? Happy to rework the tests if that's the case!

lib/rules/declaration-property-max-values/index.js
mattxwang: I am not sure, though I'm inclined to say property (since to me, this rule is about properties meeting a certain predicate) - what do you think?

ybiquitous: Indeed, this rule aims at a *declaration-property*, not a *declaration property value*. So let's keep it as-is. Thanks for the answer. 👍🏼

lib/rules/declaration-property-max-values/README.md
ybiquitous: Thanks, I understand the background. Keeping it as-is may be suitable. 👍🏼 
(@jeddy3 please left a comment if you have about this point)

ybiquitous (approved):
@mattxwang Thank you for implementing the excellent rule. LGTM 👍🏼

lib/rules/declaration-property-max-values/README.md
jeddy3: Let's keep as-is.

jeddy3: > I can circumvent this by setting margin: 1 in the config - would that get the desired effect you're looking for? Happy to rework the tests if that's the case!

Let's do that, i.e. swap `border` and `margin` around so we still have two different numbers.

```
{
  "border": 2,
  "/^margin/": 1,
},
```

Then we can add `margin-inline` back in.

lib/rules/declaration-property-max-values/index.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+const matchesStringOrRegExp = require('../../utils/matchesStringOrRegExp');
+const report = require('../../utils/report');
+const ruleMessages = require('../../utils/ruleMessages');
+const valueParser = require('postcss-value-parser');
```
jeddy3, line 6:
```suggestion
const valueParser = require('postcss-value-parser');

const matchesStringOrRegExp = require('../../utils/matchesStringOrRegExp');
const report = require('../../utils/report');
const ruleMessages = require('../../utils/ruleMessages');
```

Let's move the external `require` to the top (and separate it out). This seems to be the convention we're adopting.

(We should probably use Prettier or ESLint to do this for us).

jeddy3 (changes requested):
> @jeddy3, I may be missing something, but what was your requested change? I don't see anything attached to your "requested changes" dialog box.

That's weird. It seems it fell into the ether.

I've readded it.

lib/rules/declaration-property-max-values/README.md
mattxwang: Ah, totally missed this comment earlier - will do now!

mattxwang:
Completely missed one of the requested changes, my fault! Should have addressed all the comments - @jeddy3 let me know what you think!

jeddy3 (approved):
Thanks for making the changes. I think the tests and readme work better now.

LGTM.

jeddy3:
> - Added: `declaration-property-max-values` rule ([#5920](https://github.com/stylelint/stylelint/pull/5920)).

PR #5922, opened 2022-02-20
Refactor to add type-guard of `isPlainObject()`

ybiquitous:

None, as it's a refactoring.

- Add a type-guard function `isPlainObject()` to our `validateTypes.js` module.
  - The new type-guard function uses internally the `is-plain-object` package.
- Use the new function via `validateTypes.js`, instead of directly using `is-plain-object`.

### DIFF
--- a/lib/normalizeRuleSettings.js
+++ b/lib/normalizeRuleSettings.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const rules = require('./rules');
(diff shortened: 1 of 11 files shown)


jeddy3 (approved):
Moving to `validateTypes` LGTM.

ybiquitous:
@jeddy3 Thank you for the review!

PR #5925, opened 2022-02-20
fix(custom-property-pattern): false positives for interpolation inside var

### LINKED ISSUE #5921: Fix false positives for interpolation inside var in custom-property-pattern

run stylelint on a sass file that interpolates sass variable like so
```css
background-color: var(#{appearance.$appearance-color-fill});
```

```json
{
  "ignoreFiles": [
    "**/*"
  ],
  "overrides": [
    {
      "files": [
        "**/*.scss"
      ],
      "plugins": [
        "stylelint-no-unsupported-browser-features"
      ],
      "extends": [
        "stylelint-config-idiomatic-order",
        "stylelint-config-prettier"
      ],
      "rules": {
        "indentation": "tab",
        "plugin/no-unsupported-browser-features": [
          true,
          {
            "severity": "warning"
          }
        ],
        "order/properties-alphabetical-order": null,
        "selector-max-id": null,
        "selector-class-pattern": null,
        "max-nesting-depth": null,
        "selector-max-compound-selectors": null,
        "no-duplicate-selectors": true,
        "declaration-block-no-duplicate-properties": true,
        "function-parentheses-space-inside": null,
        "rule-empty-line-before": null
      }
    }
  ]
}
```

`npx nx run-many -- --target=stylelint --all`

`14.5.1`

stylelint `custom-property-pattern` rule should maybe ignore this pattern?

### What actually happened?

stylelint `custom-property-pattern` rule expects custom property name to be kebab-case
image

Yes, it's related to scss interpolation

### Proposal to fix the bug

not familiar with the areas of responsibilities but might this be owned by `stylelint-config-standard-scss`?

yinonov:
Closes #5921

### DIFF
--- a/lib/rules/custom-property-pattern/__tests__/index.js
+++ b/lib/rules/custom-property-pattern/__tests__/index.js
@@ -19,6 +19,12 @@ testRule({
 		{
 			code: ':root { --foo-sub-color: #f00; } a { color: var(--foo-color, var(--foo-sub-color)); }',
 		},
+		{
+			code: 'a { color: var($foo-color); }',
+		},
+		{
+			code: 'a { color: var(tokens.$foo-color); }',
+		},
 	],
 
 	reject: [
--- a/lib/rules/custom-property-pattern/index.js
+++ b/lib/rules/custom-property-pattern/index.js
@@ -45,7 +45,11 @@ const rule = (primary) => {
 
 				const { nodes, sourceIndex } = node;
 
-				if (regexpPattern.test(nodes[0].value.slice(2))) return;
+				const { value: firstNodeValue } = nodes[0];
+
+				if (!isCustomProperty(firstNodeValue)) return;
+
+				if (regexpPattern.test(firstNodeValue.slice(2))) return;
 
 				complain(declarationValueIndex(decl) + sourceIndex, decl);
 			});


ybiquitous (approved):
@yinonov Thank you for the quick fix. LGTM 👍🏼

lib/rules/custom-property-pattern/__tests__/index.js
```diff
@@ -19,6 +19,12 @@ testRule({
 		{
 			code: ':root { --foo-sub-color: #f00; } a { color: var(--foo-color, var(--foo-sub-color)); }',
 		},
+		{
+			code: 'a { color: var($foo-color); }',
+		},
+		{
+			code: 'a { color: var(tokens.$foo-color); }',
+		},
```
jeddy3, line 27: Let's move this into a separate `testRule` that sets the `customSyntax` property ([like so](https://github.com/stylelint/stylelint/blob/63b22e065c5e7fbf1e25e9626b167918b81589cd/lib/rules/alpha-value-notation/__tests__/index.js#L324-L357)) as we like to keep non-standard syntax separate from standard syntax.

lib/rules/custom-property-pattern/__tests__/index.js
```diff
@@ -125,3 +125,18 @@ testRule({
 		},
 	],
 });
+
+testRule({
+	ruleName,
+	customSyntax: 'postcss-scss',
+	config: ['$foo-color'],
```
yinonov, line 132: @jeddy3 any hints on relevant config here? ref to relevant documentation on possible values?

jeddy3, line 132:
```suggestion
	config: [/foo-.+/],
```

Let's use the same config as above as using a separate `testRule` is only about:
- keeping non-standard syntax away from standard syntax
- ensuring a custom syntax is used (even if the default parser can parse the construct)

lib/rules/custom-property-pattern/__tests__/index.js
```diff
@@ -125,3 +125,18 @@ testRule({
...
+
+	accept: [
+		{
+			code: 'a { color: var($foo-color); }',
+		},
+		{
+			code: 'a { color: var(tokens.$foo-color); }',
```
jeddy3, line 139:
```suggestion
			code: 'a { color: var(tokens.$bar-color); }',
```

Let's use `bar` here to show that the rule ignores this construct outright, and it's not connected to the config.

jeddy3 (changes requested):
Thanks for making the change.

I've requested two follow up changes (as committable suggestions).

lib/rules/custom-property-pattern/__tests__/index.js
jeddy3: See https://github.com/stylelint/stylelint/pull/5925#discussion_r811145249

jeddy3 (approved):
Thank you! LGTM.

jeddy3:
Changelog entry added:

> - Fixed: `custom-property-pattern` false positives for interpolation inside `var()` ([#5925](https://github.com/stylelint/stylelint/pull/5925)).

yinonov:
I must say that the contribution experience with styling has been tremendously pleasant. Starting with the report bug template, maintainers assistance, and the development environment setup.

This should be rated 5 stars in a GitHub open source criteria If there was any.

Thanks team stylelint 😀

jeddy3:
> I must say that the contribution experience with styling has been tremendously pleasant.

That's fantastic to hear!

Thank you for your contribution.

PR #5934, opened 2022-02-23
Fix invalid option warnings for strings in `*-list` 

ybiquitous:
> Which issue, if any, is this issue related to?

Closes #5933

> Is there anything in the PR that needs further explanation?

This pull request does:

- loosen up the `validateObjectWithArrayProps()` utility to accept also a single value not only an array. E.g.
    ```js
    validateObjectWithArrayProps({ foo: [1, 2], bar: 3 }) //=> true
    ```
- fix or update the undocumented behavior of all rules using `validateObjectWithArrayProps()`

Issue #5933, opened 2022-02-22
Fix invalid option warning for string properties in *-list


edg2s:
### What steps are needed to reproduce the bug?

The config below fails in 14.5.2 (but worked in 14.5.1):
```
Invalid Option: Invalid option value "{"/^border/":"none","/^outline/":"none"}" for rule "declaration-property-value-disallowed-list"
```

### What Stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "declaration-property-value-disallowed-list": {
      "/^border/": "none",
      "/^outline/": "none"
    }
  }
}
```
### How did you run Stylelint?

CLI

### Which version of Stylelint are you using?

14.5.2

### What did you expect to happen?

No Invalid Option warning

### What actually happened?

Invalid Option warning

edg2s:
Might have been caused by #5924?

jeddy3:
@edg2s Thanks for the report and for using the template.

The rule [expects an array](https://stylelint.io/user-guide/rules/list/declaration-property-value-disallowed-list/#options):

```json
{
  "rules": {
    "declaration-property-value-disallowed-list": {
      "/^border/": ["none"],
      "/^outline/": ["none"]
    }
  }
}
```

But it seems we had undocumented behaviour of accepting strings.

What's the appropriate thing to do in this situation? If people are relying on this behaviour, should we make the options validator more lenient?

jeddy3:
We allow strings for other `*-(dis)allowed-list` rules, [for example](https://stylelint.io/user-guide/rules/list/function-allowed-list#options).

To be consistent, we should officially allow strings in this rule too. (It was a historical design decision mistake to allow them in the first place as it needlessly complicates things, but are where we are now.)

We'll need to loosen up [`validateObjectWithArrayProps`](https://github.com/stylelint/stylelint/blob/main/lib/utils/validateObjectWithArrayProps.js) util.

I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

edg2s:
> But it seems we had undocumented behaviour of accepting strings.

Sounds like a grey area, but it would be best to avoid breaking changes (even for undocumented features) in minor releases.

Right now we are in position where will have to release a new version of our shared config to prevent users seeing errors while doing normal minor version updates.

ybiquitous:
[...]

PR #5936, opened 2022-02-23
Refactor to improve types and docs for `*-list` rules

ybiquitous:

Follow-up of #5934

When I was looking into the problem of #5933 and #5934, I felt to check the `*-list` rules again.

So, this pull request aims to improve the docs and type of that rules for consistency and clarity.
It doesn't change the existing behavior.

### DIFF
--- a/lib/rules/at-rule-allowed-list/__tests__/index.js
+++ b/lib/rules/at-rule-allowed-list/__tests__/index.js
@@ -80,7 +80,7 @@ testRule({
 	ruleName,
 	skipBasicChecks: true,
 
-	config: ['keyframes'],
+	config: 'keyframes',
 
 	accept: [
 		{
--- a/lib/rules/at-rule-allowed-list/index.js
+++ b/lib/rules/at-rule-allowed-list/index.js
@@ -17,14 +17,11 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/list/at-rule-allowed-list',
 };
 
-/** @type {import('stylelint').Rule} */
+/** @type {import('stylelint').Rule<string | string[]>} */
 const rule = (primary) => {
-	// To allow for just a string as a parameter (not only arrays of strings)
-	const primaryValues = [primary].flat();
-
 	return (root, result) => {
 		const validOptions = validateOptions(result, ruleName, {
-			actual: primaryValues,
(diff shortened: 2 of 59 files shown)


lib/rules/at-rule-allowed-list/__tests__/index.js
```diff
@@ -80,7 +80,7 @@ testRule({
 	ruleName,
 	skipBasicChecks: true,
 
-	config: ['keyframes'],
+	config: 'keyframes',
```
ybiquitous, line 83: [note] This change aims to test a single primary option value.

lib/rules/at-rule-allowed-list/index.js
```diff
@@ -17,23 +17,19 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/list/at-rule-allowed-list',
 };
 
-/** @type {import('stylelint').Rule} */
+/** @type {import('stylelint').Rule<string | string[]>} */
 const rule = (primary) => {
-	// To allow for just a string as a parameter (not only arrays of strings)
-	const primaryValues = [primary].flat();
```
ybiquitous, line 23: [note] `validateOptions` allows both a single value and an array, so it's better to call `.flat()` after the validation.

lib/rules/selector-combinator-disallowed-list/__tests__/index.js
```diff
@@ -64,3 +64,21 @@ testRule({
...
+		},
+	],
+
+	reject: [
+		{
+			code: 'a > b {}',
+			message: messages.rejected('>'),
+		},
+	],
+});
```
ybiquitous, line 84: [note] This new test case aims to just check a single primary value.

lib/utils/containsString.js
```diff
@@ -33,10 +33,8 @@ module.exports = function containsString(input, comparison) {
 };
 
 /**
- *
  * @param {string} value
- * @param {string} comparison
- *
+ * @param {unknown} comparison
```
ybiquitous, line 37: [note] `testAgainst()` uses `isString()`, so it accepts the `unknown` argument.

jeddy3 (approved):
Fantastic cleanup and refactor, thank you!

LGTM.

PR #5955, opened 2022-03-08
Fix column position for `font-family-name-quotes` rule

ybiquitous:

Extracted from #5725. This change should make it easier to support a column range.

This change fixes a column position of a `font-family-name-quotes` problem to include quotes.

```css
a { font-family: "Arial"; }
/* Before:        ↑ */

a { font-family: "Arial"; }
/* After:        ↑ */
```


lib/rules/font-family-name-quotes/__tests__/index.js
```diff
@@ -400,36 +436,26 @@ testRule({
 			code: 'a { font-family: "Lucida Grande", "Arial", "sans-serif"; }',
 			fixed: 'a { font-family: "Lucida Grande", "Arial", sans-serif; }',
 			message: messages.rejected('sans-serif'),
-			line: 1,
-			column: 45,
```
ybiquitous, line 404: [note] I've removed checking a position from `fix: true` test cases since they are duplicated with the *non-fix* test cases above. No problems?

lib/rules/font-family-name-quotes/index.js
```diff
@@ -83,11 +85,16 @@ const makeMutableFontFamilies = (fontFamilies, decl) => {
 	const mutableNodes = [];
 
 	fontFamilies.forEach((fontFamily, idx) => {
+		const quote = 'quote' in fontFamily && fontFamily.quote;
```
ybiquitous, line 88: [note] The [`StringNode#quote`](https://github.com/TrySound/postcss-value-parser/blob/d4a582692533fc87cbab348136e1dbea365d5a94/lib/index.d.ts#L62-L65) property holds an actual quote like `"` or `'`.

jeddy3 (approved):
LGTM, thank you!

ybiquitous:
@jeddy3 Thank you for the review! I've updated the changelog:

> - Fixed: font-family-name-quotes column position ([#5955](https://github.com/stylelint/stylelint/pull/5955)).

PR #5957, opened 2022-03-08
Fix column position for `*-no-important` rules

ybiquitous:

Extracted from #5725. This pull request should make it easier to support the column range.

I cannot find how to get a position of `!important` from CSS code via `postcss` or `postcss-value-parser`.
So, I've written a tiny utility to do it.


jeddy3 (approved):
Nice.

LGTM, thanks!

PR #5958, opened 2022-03-08
Refactor to replace deprecated `String.prototype.substr()`

ybiquitous:

None, it's just a small refactoring.

The `String.prototype.substr()` method is deprecated.
This change replaces it with the `String.prototype.slice()` method.

See also:
- https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String/substr
- https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String/slice

`substr()` will be not found after this change:

```console
$ git grep -w substr || echo 'no match'
no match
```

### DIFF
--- a/lib/rules/comment-whitespace-inside/index.js
+++ b/lib/rules/comment-whitespace-inside/index.js
@@ -59,7 +59,7 @@ const rule = (primary, _secondaryOptions, context) => {
 			}
 
 			const rawComment = comment.toString();
-			const firstFourChars = rawComment.substr(0, 4);
+			const firstFourChars = rawComment.slice(0, 4);
 
 			// Return early if sourcemap or copyright comment
 			if (/^\/\*[#!]\s/.test(firstFourChars)) {
--- a/lib/rules/function-whitespace-after/index.js
+++ b/lib/rules/function-whitespace-after/index.js
@@ -75,7 +75,7 @@ const rule = (primary, _secondaryOptions, context) => {
 					return;
 				}
 
-				if (source.substr(index, 2) === '\r\n') {
+				if (source.slice(index, index + 2) === '\r\n') {
 					return;
 				}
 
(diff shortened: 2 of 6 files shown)


lib/rules/comment-whitespace-inside/index.js
```diff
@@ -59,7 +59,7 @@ const rule = (primary, _secondaryOptions, context) => {
 			}
 
 			const rawComment = comment.toString();
-			const firstFourChars = rawComment.substr(0, 4);
+			const firstFourChars = rawComment.slice(0, 4);
```
ybiquitous, line 62: [note] The second argument of `substr()` means a *length*, while the one of `slice()` means an *endIndex*. If the first argument is `0`, they can be easily replaced.

lib/rules/function-whitespace-after/index.js
```diff
@@ -75,7 +75,7 @@ const rule = (primary, _secondaryOptions, context) => {
 					return;
 				}
 
-				if (source.substr(index, 2) === '\r\n') {
+				if (source.slice(index, index + 2) === '\r\n') {
```
ybiquitous, line 78: [note] Since the `index` variable is not necessarily `0`, it's needed to change the second argument.

lib/rules/selector-list-comma-newline-after/index.js
```diff
@@ -53,7 +53,7 @@ const rule = (primary, _secondaryOptions, context) => {
 					functionArguments: 'skip',
 				},
 				(match) => {
-					const nextChars = selector.substr(match.endIndex, selector.length - match.endIndex);
+					const nextChars = selector.slice(match.endIndex);
```
ybiquitous, line 56: [note] Since `selector.length - match.endIndex` means to extract a sub-string from `match.endIndex` to the end of `selector`, it can be replaced by omitting the second argument.

jeddy3 (approved):
LGTM, thank you!

ybiquitous:
@jeddy3 Thank you for the review!

Issue #5964, opened 2022-03-14
Document `fix` and `output` behaviour for Node.js API usage


jlarmstrongiv:
### Describe the documentation issue

I would like to run stylelint to fix styles programmatically with the node api.  

To compare, prettier has a node api that will format my code input string and return my prettified code output string:
```ts
prettier.format(css, { ...prettierConfig, parser: "css" });
```

However, with stylelint, I receive many other objects with no clear path to just get my linted code back.  I dug around in `linted.results[0]._postcssResult`, but couldn’t find it.  I also tried all the formatters.

### What solution would you like to see?

I would love to see documentation on how to pass a css string to the stylelint node api and received my fixed css string back

jeddy3:
@jlarmstrongiv Thanks for the report and for using the template.

The autofixed code is available as the value of the `output` property in the return object, but only when the `fix` property is set to `true`:

```js
import stylelint from "stylelint";

const result = await stylelint.lint({
  code: "a { color: hsl(30 10% 15% / 5%) }",
  config: {
    rules: {
      "hue-degree-notation": "angle",
    },
  },
});

console.log(result);
```

Gives:

```
jeddy3@MacBook stylelint-test % node node-api.js
{
  cwd: '/Users/jeddy3/Projects/stylelint-test',
  errored: true,
  results: [
    {
      source: '<input css M42saI>',
      deprecations: [],
      invalidOptionWarnings: [],
      parseErrors: [],
      errored: true,
      warnings: [Array],
      ignored: undefined,
      _postcssResult: [Result]
    }
  ],
  output: '[{"source":"<input css M42saI>","deprecations":[],"invalidOptionWarnings":[],"parseErrors":[],"errored":true,"warnings":[{"line":1,"column":16,"rule":"hue-degree-notation","severity":"error","text":"Expected \\"30\\" to be \\"30deg\\" (hue-degree-notation)"}]}]',
  reportedDisables: []
}
```

Whereas:

```js
import stylelint from "stylelint";

const result = await stylelint.lint({
  code: "a { color: hsl(30 10% 15% / 5%) }",
  config: {
    rules: {
      "hue-degree-notation": "angle",
    },
  },
  fix: true,
});

console.log(result);
```

Gives:

```
jeddy3@MacBook stylelint-test % node node-api.js
{
  cwd: '/Users/jeddy3/Projects/stylelint-test',
  errored: false,
  results: [
    {
      source: '<input css 1eb_BR>',
      deprecations: [],
[...]

PR #5971, opened 2022-03-16
Document fix and output behaviour for Node.js API usage

jlarmstrongiv:

Closes https://github.com/stylelint/stylelint/issues/5964

No, it’s documentation related.

### DIFF
--- a/docs/user-guide/usage/node-api.md
+++ b/docs/user-guide/usage/node-api.md
@@ -60,9 +60,10 @@ Boolean. If `true`, at least one rule with an "error"-level severity registered 
 
 ### `output`
 
-A string displaying the formatted problems (using the default formatter or whichever you passed).
+A string that contains either the:
 
-If the `fix` option is set to `true`, a string displaying the autofixed code.
+- formatted problems (using the default formatter or whichever you passed)
+- or the autofixed code, if the `fix` option is set to `true`
 
 ### `postcssResults`
 
--- a/docs/user-guide/usage/node-api.md
+++ b/docs/user-guide/usage/node-api.md
@@ -175,6 +175,24 @@ stylelint
 
 Note that the customSyntax option also accepts a string. [Refer to the options documentation for details](./options.md#customsyntax).
 
+### Example F
+
+Using a string and the fix option:
+
+```js
+stylelint
+  .lint({
+    code: "a { color: pink; }",
+    config: { rules: { "hue-degree-notation": "angle" } },
+    fix: true
+  })
+  .then(function () {
+    /* .. */
+  });
+```
+
+The autofixed code will be available as the value of the output property in the returned object.
+
 ## Resolving the effective config for a file
 
 If you want to find out what exact configuration will be used for a file without actually linting it, you can use the `resolveConfig()` function. Given a file path, it will return a Promise that resolves with the effective configuration object:
--- a/docs/user-guide/usage/node-api.md
+++ b/docs/user-guide/usage/node-api.md
@@ -22,6 +22,10 @@ Stylelint does not bother looking for a `.stylelintrc` file if you use this opti
 
 A string to lint.
 
+### `fix`
+
+Boolean. If `true`, the autofixed code is available as the value of the output property in the returned object.
+
 ### `cwd`
 
 The directory from which Stylelint will look for files. Defaults to the current working directory returned by `process.cwd()`.
@@ -57,6 +61,8 @@ Boolean. If `true`, at least one rule with an "error"-level severity registered 
 ### `output`
 
 A string displaying the formatted problems (using the default formatter or whichever you passed).
+
+If the `fix` option is set to `true`, a string displaying the autofixed code.
 
 ### `postcssResults`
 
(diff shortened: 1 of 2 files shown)


ybiquitous (approved):
@jlarmstrongiv Thank you for creating the pull request. LGTM 👍🏼 

It might be better if we could add an example using the `fix` option, but I want to hear @jeddy3's idea.

docs/user-guide/usage/node-api.md
```diff
@@ -22,6 +22,10 @@ Stylelint does not bother looking for a `.stylelintrc` file if you use this opti
 
 A string to lint.
 
+### `fix`
+
+Boolean. If `true`, the autofixed code is available as the value of the output property in the returned object.
+
```
jeddy3, line 28:
```suggestion
```

Let's move this to the [options pages](https://github.com/stylelint/stylelint/blob/main/docs/user-guide/usage/options.md#fix) as it's intentionally deduped.

We can add the line somewhere in the middle, so like:

---

CLI flag: `--fix`

Automatically fix, where possible, problems reported by rules.

For CSS with standard syntax, Stylelint uses [postcss-safe-parser](https://github.com/postcss/postcss-safe-parser) to fix syntax errors.

When using the Node.js API, the autofixed code is available as the value of the `output` property in the returned object.

If a source contains a:

- scoped disable comment, e.g. `/* stylelint-disable indentation */`, any problems reported by the scoped rules will not be automatically fixed anywhere in the source
- unscoped disable comment, i.e. `/* stylelint-disable */`, the entirety of source will not be automatically fixed

This limitation in being tracked in [issue #2643](https://github.com/stylelint/stylelint/issues/2643).

docs/user-guide/usage/node-api.md
```diff
@@ -58,6 +62,8 @@ Boolean. If `true`, at least one rule with an "error"-level severity registered
 
 A string displaying the formatted problems (using the default formatter or whichever you passed).
 
+If the `fix` option is set to `true`, a string displaying the autofixed code.
+
```
jeddy3, line 66:
```suggestion
A string that contains either the:

- formatted problems (using the default formatter or whichever you passed)
- or the autofixed code, if the `fix` option is set to `true`

```

Let's make it more explicit that it can contain one of two things.

jeddy3 (changes requested):
@jlarmstrongiv Thanks for the pull request.

I've requested some changes.

As @ybiquitous suggested, like add an "Example H". Something along the lines of:

---

### Example H

Using a string and the `fix` option:

```js
stylelint
  .lint({
    code: "a { color: pink; }",
    config: { rules: "color-no-invalid-hex" },
    fix: true
  })
  .then(function () {
    /* .. */
  });
```

The autofixed code will be available as the value of the `output` property in the returned object.

jlarmstrongiv:
@jeddy3 great suggestions!  I’ve updated the PR accordingly

docs/user-guide/usage/node-api.md
```diff
@@ -172,6 +175,24 @@ stylelint
 
 Note that the customSyntax option also accepts a string. [Refer to the options documentation for details](./options.md#customsyntax).
 
+### Example F
+
+Using a string and the fix option:
```
jeddy3, line 180:
```suggestion
Using a string and the `fix` option:
```

docs/user-guide/usage/node-api.md
```diff
@@ -172,6 +175,24 @@ stylelint
...
+    code: "a { color: pink; }",
+    config: { rules: { "hue-degree-notation": "angle" } },
+    fix: true
+  })
+  .then(function () {
+    /* .. */
+  });
+```
+
+The autofixed code will be available as the value of the output property in the returned object.
```
jeddy3, line 194:
```suggestion
The autofixed code will be available as the value of the `output` property in the returned object.
```

jeddy3 (changes requested):
@jlarmstrongiv Thanks for making those changes.

I've added two minor nits suggestions.

Let's also update the options page as detailed in https://github.com/stylelint/stylelint/pull/5971#discussion_r828902895

jlarmstrongiv:
@jeddy3 thank you!  Should be all set 🤞

docs/user-guide/usage/options.md
```diff
@@ -34,7 +34,9 @@ CLI flag: `--fix`
 
 Automatically fix, where possible, problems reported by rules.
 
-For CSS with standard syntax, Stylelint uses [postcss-safe-parser](https://github.com/postcss/postcss-safe-parser) to fix syntax errors.
+For CSS with standard syntax, Stylelint uses postcss-safe-parser to fix syntax errors.
```
jeddy3, line 37:
```suggestion
For CSS with standard syntax, Stylelint uses [postcss-safe-parser](https://www.npmjs.com/package/postcss-safe-parser) to fix syntax errors.
```

docs/user-guide/usage/options.md
```diff
@@ -34,7 +34,9 @@ CLI flag: `--fix`
...
+
+When using the Node.js API, the autofixed code is available as the value of the output property in the returned object.
```
jeddy3, line 39:
```suggestion
When using the Node.js API, the autofixed code is available as the value of the `output` property in the returned object.
```

jeddy3 (changes requested):

Almost there... the markdown is lost when copying and pasting.

Two suggestions are to add the link and backticks back in.

jlarmstrongiv:
@jeddy3 great catch!  Thank you for your patience

PR #5970, opened 2022-03-16
Fix false positives for large/small/dynamic viewport units in unit-no-unknown

Mouvedia:
> Which issue, if any, is this issue related to?

- close #5965
- ~~[#5717](https://github.com/stylelint/stylelint/pull/5717#issuecomment-991833468)~~

PR #5973, opened 2022-03-17
Fix false positives for SVG elements in selector-type-case

Mouvedia:
see https://github.com/stylelint/stylelint/pull/5970#pullrequestreview-913559407


ybiquitous (approved):
@Mouvedia Thanks. LGTM 👍🏼

jeddy3:
> - Fixed: `selector-type-case` false positives for SVG elements ([#5973](https://github.com/stylelint/stylelint/pull/5973)).

PR #5982, opened 2022-03-23
Fix TypeError error "Cannot destructure property..." in custom-property-pattern

### LINKED ISSUE #5972: Fix TypeError error "Cannot destructure property..." in custom-property-pattern

```css
a { prop: var(); }
```

[Demo](https://stylelint.io/demo#N4Igxg9gJgpiBcICGACYKAOAnCH4oDcksAKASgG4UBfAHQDsQAaEAZwE96AXJADwXCtWzcBHoAzAJYBzAcAYoUtEFgCuAGxitl+efUWLlYVay4QAtgFpsuGFi7trSLlzv0dKANoKDhkAD0STyRLAC8AXWCwgAZLAE5wgCoyEkso0NiEgGoyRIASZSYfXz1fX2VzLVYkaRgPZQBRXgwYMFcoFGNTC0wcFvt2FHokSpQzFAAjGBQAaxgJpAnLMCRWOpBigzp9A3CfbeoQaiA)

```json
{
  "rules": {
    "custom-property-pattern": [
      "^([a-z][a-z0-9]*)(-[a-z0-9]+)*$",
      {
        "message": "Expected custom property name to be kebab-case"
      }
    ]
  }
}
```

Demo

14.6.0

No TypeError

### What actually happened?

TypeError

No

### Proposal to fix the bug

Check that `nodes` is available before destructuring.

Follow-up PR to add "empty functions" to [commonly overlooked edge cases](https://stylelint.io/developer-guide/rules#commonly-overlooked-edge-cases).

mattxwang:

Closes #5972.

In the issue description, @jeddy3 mentions to check if `nodes` exists before restructuring. My understanding is that `nodes` will always exist, and the bug arises from if the array is empty (and thus, `nodes[0]` doesn't exist). ~~So, this approach checks if `nodes` is an empty array, and returns early if so.~~ So, this approach checks if `nodes[0]` exists instead.

I've also added a test case (that previously failed with the same error, and now is resolved).

### DIFF
--- a/lib/rules/custom-property-pattern/__tests__/index.js
+++ b/lib/rules/custom-property-pattern/__tests__/index.js
@@ -19,6 +19,10 @@ testRule({
 		{
 			code: ':root { --foo-sub-color: #f00; } a { color: var(--foo-color, var(--foo-sub-color)); }',
 		},
+		{
+			code: 'a { --foo-color: var(); }',
+			description: 'empty function',
+		},
 	],
 
 	reject: [
--- a/lib/rules/custom-property-pattern/index.js
+++ b/lib/rules/custom-property-pattern/index.js
@@ -58,6 +58,8 @@ const rule = (primary) => {
 
 				const { nodes, sourceIndex } = node;
 
+				if (nodes.length === 0) return;
+
 				const { value: firstNodeValue } = nodes[0];
 
 				if (check(firstNodeValue)) return;
--- a/lib/rules/custom-property-pattern/index.js
+++ b/lib/rules/custom-property-pattern/index.js
@@ -58,11 +58,9 @@ const rule = (primary) => {
 
 				const { nodes, sourceIndex } = node;
 
-				if (nodes.length === 0) return;
+				const firstNode = nodes[0];
 
-				const { value: firstNodeValue } = nodes[0];
-
-				if (check(firstNodeValue)) return;
+				if (!firstNode || check(firstNode.value)) return;
 
 				complain(declarationValueIndex(decl) + sourceIndex, decl);
 			});


ybiquitous:
@mattxwang Thanks for creating the pull request. 👍🏼 

I think we can use the [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) option of TypeScript to detect such a potential bug:

```diff
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,6 +9,7 @@
 		"strict": true,
 		"noFallthroughCasesInSwitch": true,
 		"noUnusedParameters": true,
+		"noUncheckedIndexedAccess": true,
 		"esModuleInterop": true,
 		"resolveJsonModule": true,
 		"typeRoots": ["./types", "./node_modules/@types"]
```

Then, run `npm run lint:types`:

```console
$ npm run lint:types | grep custom-property-pattern
lib/rules/custom-property-pattern/index.js(61,13): error TS2339: Property 'value' does not exist on type 'Node | undefined'.
```

If we would turn on the option, the following might be better:

```js
const firstNode = nodes[0];

if (firstNode && check(firstNode.value)) return;
```

Any thoughts?

mattxwang:
Thanks for the suggestion!

> If we would turn on the option, the following might be better:
> 
> ```js
> const firstNode = nodes[0];
> 
> if (firstNode && check(firstNode.value)) return;
> ```
> 
> Any thoughts?

I like how concise this is. Just to confirm, should the early return look more like

```js
const firstNode = nodes[0];

if (!firstNode || check(firstNode.value)) return;
```

In your suggestion, the original error fails, since `firstNode` is `undefined` (and so the checker doesn't early-exit). Or, did you mean something else?

> I think we can use the [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) option of TypeScript to detect such a potential bug:

I think this is a good idea. I know @jeddy3 mentioned in the original issue

> Follow-up PR to add "empty functions" to [commonly overlooked edge cases](https://stylelint.io/developer-guide/rules#commonly-overlooked-edge-cases).

Perhaps I can create a follow-up PR to enable the TS rule, resolve the other issues (umbrella-issue style), and add it to the edge case list?

ybiquitous:
I've created the draft pull request #5983 to turn on `noUncheckedIndexedAccess`.

mattxwang:
Great, sounds good - have pushed a copy of your suggested change, thanks for the help! If we think this approach works, I can work on eventually getting #5983 to have no errors 😊

ybiquitous (approved):
LGTM! 👍🏼

jeddy3 (approved):
LGTM too, thanks!

jeddy3:
> - Fixed: `custom-property-pattern` TypeError for "Cannot destructure property..." ([#5982](https://github.com/stylelint/stylelint/pull/5982)).

PR #5983, opened 2022-03-24
Refactor to turn on `noUncheckedIndexedAccess` option for TypeScript

ybiquitous:
https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess

Ref: https://github.com/stylelint/stylelint/pull/5982#issuecomment-1076966625

The `noUncheckedIndexedAccess` option allows us to detect potential bugs like #5982, e.g.

```js
const firstNoe = node.nodes[0];
firstNode.value; // `firstNode` may be undefined, but TS compiler won't complain unless `noUncheckedIndexedAccess`.
```

Note: This pull request just helps us to see the errors if `noUncheckedIndexedAccess` is turned on. **We cannot merge it now.**

See https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess

---

EDIT: If you want to try it on a branch other than this one, you can use the CLI flag `--noUncheckedIndexedAccess`, for example:

```console
$ npm run lint:types -- --noUncheckedIndexedAccess
```


lib/rules/color-hex-length/index.js
```diff
@@ -117,7 +117,7 @@ function longer(hex) {
 	let hexVariant = '#';
 
 	for (let i = 1; i < hex.length; i++) {
-		hexVariant += hex[i] + hex[i];
+		hexVariant += hex.charAt(i).repeat(2);
```
ybiquitous, line 120: [note] [`.charAt()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charAt) always returns a string (e.g. an empty string).

lib/rules/linebreaks/index.js
```diff
@@ -60,9 +60,7 @@ const rule = (primary, _secondaryOptions, context) => {
 
 			const lines = root.source.input.css.split('\n');
 
-			for (let i = 0; i < lines.length; i++) {
-				let line = lines[i];
-
+			for (let [i, line] of lines.entries()) {
```
ybiquitous, line 63: [note] [`.entries()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/entries) makes a traditional `for` loop much readable.

lib/rules/selector-class-pattern/index.js
```diff
@@ -110,16 +110,20 @@ const rule = (primary, secondaryOptions) => {
  * @returns {boolean}
  */
 function hasInterpolatingAmpersand(selector) {
-	for (let i = 0, l = selector.length; i < l; i++) {
-		if (selector[i] !== '&') {
+	for (const [i, char] of Array.from(selector).entries()) {
```
ybiquitous, line 113: [note] [`Array.from()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) can make a string to an array of characters.

ybiquitous:
The CI is now all green. ✅

jeddy3 (approved):
This is fantastic work, thank you!

Having this option enabled will likely catch bugs that we've fallen foul to in the past.

ybiquitous:
@jeddy3 Thank you for reviewing the big pull request! 😄

PR #6002, opened 2022-03-31
Refactor to improve types for `validateTypes` and `whitespaceChecker`

ybiquitous:

Part of #5983

This refactoring does the following:

- Move utility functions in `lib/utils/whitespaceChecker.js` to `lib/utils/validateTypes.js`
- Add new utility functions to `validateTypes.js`
- Add a unit test for `validateTypes.js`
- Resolve type errors in `whitespaceChecker.js` via the `noUncheckedIndexedAccess` TypeScript compiler option

We can see the error resolution via the following commands:

```console
$ git checkout tsconfig-noUncheckedIndexedAccess -- tsconfig.json

$ npm run lint:types
```

### DIFF
--- a/lib/utils/whitespaceChecker.js
+++ b/lib/utils/whitespaceChecker.js
@@ -3,6 +3,7 @@
 const configurationError = require('./configurationError');
 const isSingleLineString = require('./isSingleLineString');
 const isWhitespace = require('./isWhitespace');
+const { assertFunction, isNullish } = require('./validateTypes');
 
 /**
  * @typedef {(message: string) => string} MessageFunction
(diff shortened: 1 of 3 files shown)


lib/utils/whitespaceChecker.js
```diff
@@ -215,20 +216,20 @@ module.exports = function whitespaceChecker(targetWhitespace, expectation, messa
...
 			oneCharBefore === ' ' &&
-			(activeArgs.onlyOneChar || !isWhitespace(twoCharsBefore))
+			(activeArgs.onlyOneChar || isNullish(twoCharsBefore) || !isWhitespace(twoCharsBefore))
 		) {
 			return;
 		}
 
 		assertFunction(messageFunc);
-		activeArgs.err(messageFunc(activeArgs.errTarget || source[index]));
+		activeArgs.err(messageFunc(activeArgs.errTarget || source.charAt(index)));
```
ybiquitous, line 232: [note] `source[index]` can return `undefined`, while `source.charAt(index)` always returns a string. See the `charAt()` doc:

> If `index` is out of range, `charAt()` returns an empty string. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charAt

I think it safe to use `charAt()` in this case to build an error message.

jeddy3 (approved):
Nice. LGTM, thank you!

ybiquitous:
@jeddy3 Thank you for the review.

PR #6003, opened 2022-03-31
Refactor `lib/utils/checkInvalidCLIOptions.js`

ybiquitous:

None, as it's a small refactoring.

This small refactoring aims to make `buildAllowedOptions()` more readable.

The number of loops is doubled (`Object.keys()` and `Object.values()`),
but it will not degrade performance since `allowedOptions` is a small array.


jeddy3 (approved):
LGTM, thanks!

Issue #3158, opened 2018-02-07
Add autofix to font-weight-notation


hiroppy:
I read https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/rules.md#adding-an-option-to-an-existing-rule and write this issue.

I already have written code.
see https://github.com/abouthiroppy/stylelint/commit/297723031bdcb239cf131fc90842a4d88757038e

### numeric
```
a { font-weight: normal; } -> a { font-weight: 400; }
a { font: italic bold 20px; } -> a { font: italic 700 20px; }
```

### named-where-possible
```
a { font-weight: 400; } -> a { font-weight: normal; }
a { font: italic 700 20px; } -> a { font: italic bold 20px; }
```

If this request is accepted, I will submit PR :)

evilebottnawi:
@abouthiroppy PR welcome!

jeddy3:
Fixing https://github.com/stylelint/stylelint/issues/3855#issuecomment-844116177 is a prerequisite to adding autofix.

See also: https://github.com/stylelint/stylelint/pull/3159#issuecomment-855271082

FloEdelmann:
@hiroppy Sorry for the ping in this old issue, but given you've already written the code for the autofix, would you fancy opening a PR?

I think @jeddy3's comment is already resolved (#3855 was closed by #6005). I don't know if that requires changes in your code though.

ybiquitous:
@FloEdelmann Thanks for the reminder. I'm going to work on this. 💪🏼

PR #6005, opened 2022-04-03
Fix column position and ignore Sass function for `font-weight-notation`

### LINKED ISSUE #3855: Fix false positives for functions in font-weight-notation

I use a deep nested SCSS map with all of my theme options and have created a function to access things like font-weights.  I'm looking to include the `font-weight-notation` to be numeric unless its using a function I've whitelisted.

As a developer, I would like to see the [font-weight-notation](https://stylelint.io/user-guide/rules/font-weight-notation/) expanded to include an option to `ignoreFunction` or maybe a regex since I use a deep nested SCSS map and function to access font-weights.

ybiquitous:

- Closes #3855
- Prepares for #5725

This change uses `postcss-value-parser` instead of `postcss.list.space()` for more accurate parsing.

### DIFF
--- a/lib/rules/font-weight-notation/README.md
+++ b/lib/rules/font-weight-notation/README.md
@@ -4,13 +4,17 @@ Require numeric or named (where possible) `font-weight` values. Also, when named
 
 <!-- prettier-ignore -->
 ```css
-a { font-weight: bold }
+a { font-weight: bold; }
 /**              ↑
- *   This notation */
+ *               This notation */
 
 a { font: italic small-caps 600 16px/3 cursive; }
 /**                         ↑
-*      And this notation, too */
+ *                          And this notation, too */
+
+@font-face { font-weight: normal bold; }
+/**                       ↑
+ *                        Multiple notations are available in @font-face */
 ```
 
 Valid font-weight names are `normal`, `bold`, `bolder`, and `lighter`.
@@ -37,6 +41,11 @@ a { font-weight: bold; }
 a { font: italic normal 20px sans-serif; }
 ```
 
+<!-- prettier-ignore -->
+```css
+@font-face { font-weight: normal bold; }
+```
+
 The following patterns are _not_ considered problems:
 
 <!-- prettier-ignore -->
@@ -47,6 +56,11 @@ a { font-weight: 700; }
 <!-- prettier-ignore -->
 ```css
 a { font: italic 400 20px; }
+```
+
+<!-- prettier-ignore -->
+```css
+@font-face { font-weight: 400 700; }
 ```
 
 ### `"named-where-possible"`
--- a/lib/rules/font-weight-notation/__tests__/index.js
+++ b/lib/rules/font-weight-notation/__tests__/index.js
@@ -180,7 +180,7 @@ testRule({
 			code: '@font-face { font-weight: 400 bold; }',
 			message: messages.expected('numeric'),
 			line: 1,
-			column: 27,
+			column: 31,
 		},
 		{
 			code: '@font-face { font-weight: normal 700; }',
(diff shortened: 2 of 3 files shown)


lib/rules/font-weight-notation/README.md
```diff
@@ -4,13 +4,17 @@ Require numeric or named (where possible) `font-weight` values. Also, when named
...
+ *               This notation */
 
 a { font: italic small-caps 600 16px/3 cursive; }
 /**                         ↑
-*      And this notation, too */
+ *                          And this notation, too */
+
+@font-face { font-weight: normal bold; }
+/**                       ↑
+ *                        Multiple notations are available in @font-face */
```
ybiquitous, line 17: [note] I think adding a description about `@font-face` makes sense. Any thoughts?

See also https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-weight

lib/rules/font-weight-notation/__tests__/index.js
```diff
@@ -180,7 +180,7 @@ testRule({
 			code: '@font-face { font-weight: 400 bold; }',
 			message: messages.expected('numeric'),
 			line: 1,
-			column: 27,
+			column: 31,
```
ybiquitous, line 183: [note] The position is changed to `bold`, not `400`.

lib/rules/font-weight-notation/index.js
```diff
@@ -98,73 +103,108 @@ const rule = (primary, secondaryOptions) => {
...
 
+/**
+ * @param {string} value
+ * @returns {import('postcss-value-parser').Node[]}
+ */
+function getWeightValueNodes(value) {
+	return valueParser(value).nodes.filter((node, index, nodes) => {
+		if (node.type !== 'word') return false;
+
+		// Exclude `<font-size>/<line-height>` format like `16px/3`.
```
ybiquitous, line 186: See http://developer.mozilla.org/en-US/docs/Web/CSS/font#syntax

lib/rules/font-weight-notation/index.js
```diff
@@ -98,73 +103,108 @@ const rule = (primary, secondaryOptions) => {
...
 				}
 			}
 
 			/**
 			 * @param {string} message
+			 * @param {import('postcss-value-parser').Node} [valueNode]
 			 */
-			function complain(message) {
+			function complain(message, valueNode) {
+				const index = declarationValueIndex(decl) + (valueNode ? valueNode.sourceIndex : 0);
```
ybiquitous, line 164: [note] I'll add `endIndex` on #5725.

lib/rules/font-weight-notation/index.js
```diff
@@ -98,73 +103,108 @@ const rule = (primary, secondaryOptions) => {
...
 			}
 		}
 	};
 };
 
+/**
+ * @param {string} value
+ * @returns {import('postcss-value-parser').Node[]}
+ */
+function getWeightValueNodes(value) {
```
jeddy3, line 182: We generally use `find*` when looking through sub ASTs for things. I'm not sure where the convention started, but shall we follow it here to be consistent, i.e. `findFontWeights(value)`?

jeddy3 (commented):
Fantastic, thank you!

This is a nice stepping stone to https://github.com/stylelint/stylelint/issues/3158 (and also https://github.com/signal-noise/stylelint-scales/issues/73).

I've asked one question.

lib/rules/font-weight-notation/index.js
ybiquitous: Thank you. Fixed via 9bb54e1b. 👍🏼

jeddy3 (approved):
LGTM, thank you!

ybiquitous:
@jeddy3 Thank you for the review! I've updated the changelog:

> - Fixed: `font-weight-notation` false positives for Sass functions and column position ([#6005](https://github.com/stylelint/stylelint/pull/6005)).

PR #6014, opened 2022-04-15
Refactor to improve types for `lib/assignDisabledRanges.js`

ybiquitous:

Ref: #5983

This change aims to:

- fix errors by the TypeScript compiler option `noUncheckedIndexedAccess`
- remove unreachable code
- reduce duplication


lib/assignDisabledRanges.js
```diff
@@ -259,18 +263,10 @@ module.exports = function assignDisabledRanges(root, result) {
 
 			if (ruleIsDisabled(ALL_RULES) && disabledRanges[ruleToEnable] === undefined) {
 				// Get a starting point from the where all rules were disabled
-				if (!disabledRanges[ruleToEnable]) {
```
ybiquitous, line 262: [note] If `ruleIsDisabled(ALL_RULES) && disabledRanges[ruleToEnable] === undefined` is `true`, `!disabledRanges[ruleToEnable]` should always be `true`. So, the `else` block should never be executed.

jeddy3 (approved):
LGTM, thank you.

ybiquitous:
@jeddy3 Thank you for the review. 👍🏼

PR #6017, opened 2022-04-17
Fix false positives for `ui-*` generic system font keywords in `font-family-name-quotes`

### LINKED ISSUE #6016: Fix false positives for `ui-*` generic system font keywords in font-family-name-quotes

1. Set the `font-family-name-quotes` rule to `always-unless-keyword`
2. Add one the following generic keywords in a `font-family` property:
    - `ui-serif`
    - `ui-sans-serif`
    - `ui-monospace`
    - `ui-rounded`
3. Run Stylelint on the file

```json
{
    "rules": {
        "font-family-name-quotes": "always-unless-keyword"
    }
}
```

CLI with `stylelint style.css --config .stylelintrc`

`14.7.0`

Stylelint would recognize `ui-monospace` et al. as a generic keyword that doesn't require quoting.

### What actually happened?

A warning was given stating that it expected quotes around `ui-monospace`.

No (see: https://www.w3.org/TR/css-fonts-4/#ui-serif-def and https://caniuse.com/extended-system-fonts).

### Proposal to fix the bug

Add `ui-serif`, `ui-sans-serif`, `ui-monospace`, and `ui-rounded` to the list of generic font-family keywords such that they be recognized by rules such as `font-family-name-quotes` and any other relevant rules.

nlydv:

Closes #6016

I don't think so.


jeddy3 (approved):
LGTM, thank you!

jeddy3:
Changelog entry added:

> - Fixed: `font-family-name-quotes` false positives for `ui-*` generic system font keywords ([#6017](https://github.com/stylelint/stylelint/pull/6017)).

nlydv:
Cool, thanks.

PR #6018, opened 2022-04-17
Fix no-disable regression for some rules

### LINKED ISSUE #6015: Fix no-disable regression for some rules

Use e.g.
```
.ProseMirror { /* stylelint-disable-line selector-class-pattern */
  ...
}
```

(taken from https://github.com/maevsi/maevsi/blob/1d6fc6eff10c922b028b8811d0aad25cad6fe85c/nuxt/assets/css/main.css#L90)

```
module.exports = {
  extends: [
    'stylelint-config-standard',
    'stylelint-config-prettier',
    'stylelint-config-recommended-vue',
  ],
  plugins: ['stylelint-no-unsupported-browser-features'],
  rules: {
    'at-rule-no-unknown': [
      true,
      {
        ignoreAtRules: [
          'apply',
          'layer',
          'responsive',
          'screen',
          'tailwind',
          'variants',
        ],
      },
    ],
    'declaration-block-trailing-semicolon': null,
    'plugin/no-unsupported-browser-features': [true, { severity: 'warning' }],
  },
}

```

`stylelint **/*.{vue,css} --ignore-path .gitignore`

`14.7.0`

The linter error is not supposed to show up.

### What actually happened?

I receive the error `Expected class selector to be kebab-case  selector-class-pattern` as if the ignore comment wasn't there. The ignore worked previously up until `14.6.1`.

See Renovate PR https://github.com/maevsi/maevsi/pull/583 with failing GitHub Action https://github.com/maevsi/maevsi/runs/6043511344?check_suite_focus=true#step:5:1294.

_No response_

### Proposal to fix the bug

_No response_

ota-meshi:

Closes #6015

This PR is modified so that stylelint-disable is enabled when the start line of the report matches.

### DIFF
--- a/lib/utils/__tests__/report.test.js
+++ b/lib/utils/__tests__/report.test.js
@@ -162,6 +162,28 @@ it('with relevant rule-specific disabledRange, among others', () => {
 	expect(v.result.warn).toHaveBeenCalledTimes(0);
 });
 
+it('with relevant rule-specific disabledRange with range report', () => {
+	const v = {
+		ruleName: 'foo',
+		result: {
+			warn: jest.fn(),
+			stylelint: {
+				disabledRanges: {
+					all: [],
+					foo: [{ start: 2, end: 2 }],
+				},
+			},
+		},
+		message: 'bar',
+		node: {
+			rangeBy: () => ({ start: { line: 2, column: 1 }, end: { line: 5, column: 1 } }),
+		},
+	};
+
+	report(v);
+	expect(v.result.warn).toHaveBeenCalledTimes(0);
+});
+
 it("with quiet mode on and rule severity of 'warning'", () => {
 	const v = {
 		ruleName: 'foo',
--- a/lib/utils/report.js
+++ b/lib/utils/report.js
@@ -28,12 +28,11 @@ module.exports = function report(problem) {
 		return;
 	}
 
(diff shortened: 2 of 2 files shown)


ota-meshi:
For example, the selector-class-pattern rule currently reports a range of rule node. So the report range in the following example is from line 1 to line 3.

```css
.ProseMirror { /* stylelint-disable-line selector-class-pattern */
  ...
}
```

The stylelint-disable-line cannot suppress range reporting, so it has no effect in the above example.

After this PR change, the suppression comment works because it only refers to the start line of the report (line 1 in the example).

ota-meshi:
In other words, this PR is reverting the next change.

https://github.com/stylelint/stylelint/pull/5725/files#diff-2995c78cc41d5f69754525ce1af5b2f682eca7d5a684b160b214381cf5f60529R53

ybiquitous (approved):
@ota-meshi Thank you so much! LGTM 👍🏼

jeddy3 (approved):
@ota-meshi Many thanks for the quick fix!

I'll put out a release now.

jeddy3:
> - Fixed: a regression for `/* stylelint-disable */` comments ([#6018](https://github.com/stylelint/stylelint/pull/6018)).

PR #6024, opened 2022-04-18
Add `keyframe-block-no-duplicate-selectors`

### LINKED ISSUE #5991: Add keyframe-block-no-duplicate-selectors

When revisiting https://github.com/stylelint/stylelint/issues/4065, I noticed on the MDN site that [duplicate selectors in keyframe-blocks are are thing](https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes#when_a_keyframe_is_defined_multiple_times).

Also show in example 3 [in the spec](https://drafts.csswg.org/css-animations/#keyframes).

I'd like to disallow them as, even though it's valid CSS, I've probably made a mistake.

A new rule like the existing [declaration-block-no-duplicate-properties](https://stylelint.io/user-guide/rules/list/declaration-block-no-duplicate-properties) one.

- Name: `keyframe-block-no-duplicate-selectors`
- Primary option: `true`
- Secondary options: None
- Autofixable: No
- Message: `Unexpected duplicate "${selector}"`
- Description: "Disallow duplicate selectors within keyframe blocks."
- Section: "Avoid errors" -> "Keyframe blocks"

[Language in the spec](https://drafts.csswg.org/css-animations/#keyframes) for `keyframe-block` and `selector`.

It would disallow the likes of:

```css
@keyframes foo {
  0% {}
  0% {}
  100% {}
}
```

Sound good?

mattxwang:

Closes #5991.

Still newish to the overall codebase! A couple of questions I had (and of course, general feedback is much appreciated):

1. Should I explicitly be handling case-insensitivity? Alternatively, should I document that the rule is case-sensitive (as of now)
2. How comprehensive are the test cases and documentation? I'm not sure if I'm missing any common cases where this would be a problem.
3. I've added vendored prefixes to the regex for finding keyframes as well as in the test cases - is this good practice?

### DIFF
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -50,7 +50,7 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 
 - [`keyframe-declaration-no-important`](../../../lib/rules/keyframe-declaration-no-important/README.md): Disallow `!important` within keyframe declarations.
 
-### Keyframe blocks
+### Keyframe block
 
 - [`keyframe-block-no-duplicate-selectors`](../../../lib/rules/keyframe-block-no-duplicate-selectors/README.md): Disallow duplicate selectors within keyframe blocks.
 
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -50,6 +50,10 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 
 - [`keyframe-declaration-no-important`](../../../lib/rules/keyframe-declaration-no-important/README.md): Disallow `!important` within keyframe declarations.
 
+### Keyframe blocks
+
+- [`keyframe-block-no-duplicate-selectors`](../../../lib/rules/keyframe-block-no-duplicate-selectors/README.md): Disallow duplicate selectors within keyframe blocks.
+
 ### Declaration block
 
 - [`declaration-block-no-duplicate-custom-properties`](../../../lib/rules/declaration-block-no-duplicate-custom-properties/README.md): Disallow duplicate custom properties within declaration blocks.
--- a/lib/rules/keyframe-block-no-duplicate-selectors/README.md
+++ b/lib/rules/keyframe-block-no-duplicate-selectors/README.md
@@ -9,6 +9,8 @@ Disallow duplicate selectors within keyframe blocks.
  *                     This duplicate selector */
 ```
 
+This rule is case-insensitive.
+
 ## Options
 
 ### `true`
@@ -23,6 +25,11 @@ The following patterns are considered problems:
 <!-- prettier-ignore -->
 ```css
 @keyframes foo { from {} from {} }
+```
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo { from {} FROM {} }
 ```
 
 The following patterns are _not_ considered problems:
--- a/lib/rules/keyframe-block-no-duplicate-selectors/README.md
+++ b/lib/rules/keyframe-block-no-duplicate-selectors/README.md
@@ -4,11 +4,9 @@ Disallow duplicate selectors within keyframe blocks.
 
 <!-- prettier-ignore -->
 ```css
-@keyframes foo {
-  0% {}
-  0% {}  /* This duplicated selector */
-  100% {}
-}
+@keyframes foo { 0% {} 0% {} }
+/**                    ↑
+ *                     This duplicate selector */
 ```
 
 ## Options
@@ -19,12 +17,12 @@ The following patterns are considered problems:
 
 <!-- prettier-ignore -->
 ```css
-@keyframes foo { 0% {} 0% {} 100% {} }
+@keyframes foo { 0% {} 0% {} }
 ```
 
 <!-- prettier-ignore -->
 ```css
-@keyframes foo { 0% {} 0%, 100% {} }
+@keyframes foo { from {} from {} }
 ```
 
 The following patterns are _not_ considered problems:
--- a/lib/rules/keyframe-block-no-duplicate-selectors/README.md
+++ b/lib/rules/keyframe-block-no-duplicate-selectors/README.md
@@ -1,0 +1,40 @@
+# keyframe-block-no-duplicate-selectors
+
+Disallow duplicate selectors within keyframe blocks.
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo {
+  0% {}
+  0% {}  /* This duplicated selector */
+  100% {}
+}
+```
+
+## Options
+
+### `true`
+
+The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo { 0% {} 0% {} 100% {} }
+```
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo { 0% {} 0%, 100% {} }
+```
+
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo { 0% {} 100% {} }
+```
+
+<!-- prettier-ignore -->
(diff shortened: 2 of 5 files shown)


mattxwang:
For code coverage - is there a specific way I should be testing the rule with invalid options? That's the only uncovered line I have.

ybiquitous (commented):
@mattxwang Thanks for creating the pull request. 👍🏼 

Additionally, it should be better to support end positions for warnings.
See also https://github.com/stylelint/stylelint/issues/5694#issuecomment-1100616663

lib/rules/keyframe-block-no-duplicate-selectors/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Disallow duplicate selectors within keyframe blocks.
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo {
+  0% {}
+  0% {}  /* This duplicated selector */
+  100% {}
+}
```
jeddy3, line 11:
```suggestion
@keyframes foo { 0% {} 0% {} }
/**                    ↑
 *                     This duplicate selector */
```

lib/rules/keyframe-block-no-duplicate-selectors/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Options
+
+### `true`
+
+The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo { 0% {} 0% {} 100% {} }
```
jeddy3, line 22:
```suggestion
@keyframes foo { 0% {} 0% {} }
```

lib/rules/keyframe-block-no-duplicate-selectors/README.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo { 0% {} 0%, 100% {} }
```
jeddy3, line 27:
```suggestion
@keyframes foo { from {} from {} }
```

lib/rules/keyframe-block-no-duplicate-selectors/__tests__/index.js
```diff
@@ -0,0 +1,40 @@
...
+	accept: [
+		{
+			code: '@keyframes foo { 0% {} 100% {} }',
+		},
+		{
+			code: '@keyframes foo { from {} to {} }',
+		},
+	],
+
+	reject: [
```
jeddy3, line 18: Let's add reject tests for:

```
@keyframes foo { 0% {} 0% {} }
```

And

```
@keyframes foo { from {} from {} }
```

lib/rules/keyframe-block-no-duplicate-selectors/__tests__/index.js
```diff
@@ -0,0 +1,40 @@
...
+		{
+			code: '@-moz-keyframes foo { 0% {} 0% {} 100% {} }',
+			message: messages.rejected('0%'),
+		},
+		{
+			code: '@keyframes foo { 0% {} 0%, 100% {} }',
+			message: messages.rejected('0%'),
+		},
+	],
+});
```
jeddy3, line 40: Let's add a `testRule` block for SCSS interpolation in the selector, e.g. an accept testcase for:

```
@keyframes foo { #{$bar} {} #{$bar} {} }
```

lib/rules/keyframe-block-no-duplicate-selectors/index.js
```diff
@@ -0,0 +1,56 @@
...
+		if (!validOptions) {
+			return;
+		}
+
+		root.walkAtRules(/^(-(moz|webkit)-)?keyframes$/i, (atRuleKeyframes) => {
+			const selectors = new Set();
+
+			atRuleKeyframes.walkRules((keyframeRule) => {
+				const ruleSelectors = keyframeRule.selectors;
+
```
jeddy3, line 31: Let's make sure the selector is standard syntax here using https://github.com/stylelint/stylelint/blob/main/lib/utils/isStandardSyntaxSelector.js

lib/rules/keyframe-block-no-duplicate-selectors/__tests__/index.js
```diff
@@ -0,0 +1,40 @@
...
+			message: messages.rejected('0%'),
+		},
+		{
+			code: '@-moz-keyframes foo { 0% {} 0% {} 100% {} }',
+			message: messages.rejected('0%'),
+		},
+		{
+			code: '@keyframes foo { 0% {} 0%, 100% {} }',
+			message: messages.rejected('0%'),
+		},
```
jeddy3, line 38:
```suggestion
		{
			code: '@keyframes foo { 0% {} 0%, 100% {} }',
			message: messages.rejected('0%'),
		},
```

Let's remove this invalid CSS.

jeddy3 (changes requested):
@mattxwang Thank you! Looking good so far.

As @ybiquitous said, we'll want to add column and line numbers (including end ones) to the reject tests.

I've requested a few changes myself.

jeddy3:
> Should I explicitly be handling case-insensitivity?

Yes. Let's add `@keyframes foo { from {} FROM {} }` as a reject test.

> How comprehensive are the test cases and documentation? 

They're comprehensive enough, thank you.

> I've added vendored prefixes to the regex for finding keyframes as well as in the test cases - is this good practice?

Yes, LGTM.

lib/rules/keyframe-block-no-duplicate-selectors/index.js
mattxwang: Add this in the `.forEach` - is this what you were thinking?

mattxwang:
Thank you @ybiquitous and @jeddy3 for the helpful feedback - end positions and nonstandard syntax slipped my mind. I've pushed some changes to address the feedback; let me know if this is what you two were looking for!

mattxwang:
Thank you @jeddy3 and @ybiquitous for the in-depth feedback - end positions and non-standard syntax slipped my mind! I've pushed changes to address the feedback; let me know if you have any other thoughts.

lib/rules/keyframe-block-no-duplicate-selectors/index.js
```diff
@@ -0,0 +1,62 @@
...
+							node: keyframeRule,
+							result,
+							ruleName,
+							word: selector,
+						});
+
+						return;
+					}
+
+					selectors.add(selector.toLowerCase());
```
ybiquitous, line 52: [trivial] It may makes sense a bit to prevent evaluating `selector.toLowerCase()` twice.

```js
const normalizedSelector = selector.toLowerCase();
```

ybiquitous (approved):
@mattxwang Nice! LGTM 👍🏼

lib/rules/keyframe-block-no-duplicate-selectors/index.js
mattxwang: Good suggestion, went a bit on autopilot - I'll add that now!

docs/user-guide/rules/list.md
```diff
@@ -50,6 +50,10 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 
 - [`keyframe-declaration-no-important`](../../../lib/rules/keyframe-declaration-no-important/README.md): Disallow `!important` within keyframe declarations.
 
+### Keyframe blocks
```
ybiquitous, line 53: [ask] Is the singular form more accurate?

```suggestion
### Keyframe block
```

jeddy3 (approved):
@mattxwang Thanks for making the changes. LGTM.

jeddy3:
> - Added: `keyframe-block-no-duplicate-selectors` rule ([#6024](https://github.com/stylelint/stylelint/pull/6024)).

PR #6026, opened 2022-04-19
Fix end positions for `at-rule-no-unknown`

ybiquitous:
> Which issue, if any, is this issue related to?

Part of #5694

> Is there anything in the PR that needs further explanation?

I think the range should be at an at-rule name, not an at-rule block. E.g.,

```css
    @unknown (max-width: 960px) {}
/** ↑------↑ */
```

PR #6032, opened 2022-04-23
Fix end positions for `at-rule-*-list`

mattxwang:

I modelled my approach off of #6026 - though, I did note that for the rule in that PR (`at-rule-no-unknown`), the `message` includes the `@`, while for the rules in this PR, the `message` does not. I've still included the `@` in the word itself; let me know if I should instead be consistent with the `message`.

(also not sure about the add/fix terminology for the PR title, feel free to change if needed)


lib/rules/at-rule-allowed-list/index.js
```diff
@@ -47,6 +47,7 @@ const rule = (primary) => {
 				node: atRule,
 				result,
 				ruleName,
+				word: `@${atRule.name}`,
```
ybiquitous, line 50: [trivial] Reusing `name` seems good 👍🏼 

```suggestion
				word: `@${name}`,
```

ybiquitous (commented):
@mattxwang Thanks, LGTM 👍🏼 

IMO, it's easier to understand if each rule message includes `@`, e.g.

```diff
- Unexpected at-rule "foo"
+ Unexpected at-rule "@foo"
```

lib/rules/at-rule-disallowed-list/index.js
```diff
@@ -47,6 +47,7 @@ const rule = (primary) => {
 				node: atRule,
 				result,
 				ruleName,
+				word: `@${atRule.name}`,
```
ybiquitous, line 50: Ah, this also can be simplified: 😅 

```suggestion
				word: `@${name}`,
```

ybiquitous (approved):

jeddy3 (approved):
LGTM, thank you!

jeddy3:
> - Fixed: `at-rule-*-list` end positions ([#6032](https://github.com/stylelint/stylelint/pull/6032)).

PR #6038, opened 2022-04-25
Fix end positions for `function-no-unknown`

mattxwang:

Let me know if `name` should be called something else.

On a related note, I'm happy to do many of the end positions PRs (they're low-hanging fruit, and a good way for me to get more familiar with rules). How many should I be doing per PR?


ybiquitous (approved):
@mattxwang Thank you. LGTM 👍🏼 

> How many should I be doing per PR?

I think it's also good to change some rules with one PR if the diff is not too huge.

jeddy3:
> - Fixed: `function-no-unknown` end positions ([#6038](https://github.com/stylelint/stylelint/pull/6038)).

PR #6045, opened 2022-04-29
Fix false positives and memory leak for `function-calc-no-unspaced-operator`

### LINKED ISSUE #5695: Fix false positives for '*' and '/' operators in function-calc-no-unspaced-operator

```css
a {
  width: calc(1*2);
}
```

```json
{
  "rules": {
    "function-calc-no-unspaced-operator": true
  }
}
```

https://stylelint.io/demo

14.0.1

No issues should be raised because [W3C CSS Values and Units Module Level 3 section 8.1.1 Syntax](https://www.w3.org/TR/css-values-3/#calc-syntax) says the following:

> In addition, white space is required on both sides of the + and - operators. (The * and / operaters can be used without white space around them.)

### What actually happened?

- Expected single space before "*" operator (function-calc-no-unspaced-operator)
- Expected single space after "*" operator (function-calc-no-unspaced-operator)

No.

### Proposal to fix the bug

At first glance, the rule implementation should no longer consider `*` and `/` operators. In other words, one would just need to remove these operators from:

https://github.com/stylelint/stylelint/blob/e4a2e13c63867c47a4e48cff4fb0d62cc703765b/lib/rules/function-calc-no-unspaced-operator/index.js#L20

### LINKED ISSUE #6044: Fix memory leak with `function-calc-no-unspaced-operator` and `--fix`

Minimal example https://github.com/dmarov/stylelint-bug
Such syntax as
```
calc(100%/var(--row-items))
```
causes memory leak when used with --fix option

{
}

stylelint style.scss --fix

14.8.0

Memory leak should not occur

### What actually happened?

command crashes with error

```
<--- Last few GCs --->

[8224:000001E2E2084530]    76754 ms: Scavenge (reduce) 4094.4 (4099.8) -> 4093.4 (4100.8) MB, 4.5 / 0.0 ms  (average mu = 0.251, current mu = 0.214) allocation failure
[8224:000001E2E2084530]    76764 ms: Scavenge (reduce) 4094.8 (4100.0) -> 4093.8 (4101.0) MB, 4.5 / 0.0 ms  (average mu = 0.251, current mu = 0.214) allocation failure

...
```

No

### Proposal to fix the bug

_No response_

ybiquitous:

- Close #5695
- Close #6044 

This pull request does 2 major things:

- dropping the support of the `*` and `/` operators since they are not required (see also [`calc()` notes on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/calc#notes))
- preventing a potential infinite-loop by mutating an array inside a loop (see also the [issue comment](https://github.com/stylelint/stylelint/issues/6044#issuecomment-11128788))

### DIFF
--- a/lib/rules/function-calc-no-unspaced-operator/README.md
+++ b/lib/rules/function-calc-no-unspaced-operator/README.md
@@ -11,6 +11,8 @@ a { top: calc(1px + 2px); }
 
 Before the operator, there must be a single whitespace or a newline plus indentation. After the operator, there must be a single whitespace or a newline.
 
+Note: The `*` and `/` operators do not require whitespace (but it is usually recommened for consistency).
+
 The [`fix` option](../../../docs/user-guide/usage/options.md#fix) can automatically fix all of the problems reported by this rule.
 
 ## Options
@@ -39,6 +41,11 @@ a { top: calc(1px + 2px); }
 <!-- prettier-ignore -->
 ```css
 a { top: calc(calc(1em * 2) / 3); }
+```
+
+<!-- prettier-ignore -->
+```css
+a { top: calc(calc(1em*2)/3); }
 ```
 
 <!-- prettier-ignore -->
--- a/lib/rules/function-calc-no-unspaced-operator/README.md
+++ b/lib/rules/function-calc-no-unspaced-operator/README.md
@@ -9,9 +9,7 @@ a { top: calc(1px + 2px); }
  * The space around this operator */
 ```
 
-Before the operator, there must be a single whitespace or a newline plus indentation. After the operator, there must be a single whitespace or a newline.
-
-Note: The `*` and `/` operators do not require whitespace (but it is usually recommened for consistency).
+This rule checks that there is a single whitespace or a newline plus indentation before the `+` or `-` operator, and a single whitespace or a newline after that operator.
 
 The [`fix` option](../../../docs/user-guide/usage/options.md#fix) can automatically fix all of the problems reported by this rule.
 
--- a/lib/rules/function-calc-no-unspaced-operator/__tests__/index.js
+++ b/lib/rules/function-calc-no-unspaced-operator/__tests__/index.js
@@ -33,7 +33,25 @@ testRule({
 			code: 'a { top: calc(1px * 2); }',
 		},
 		{
+			code: 'a { top: calc(1px*2); }',
+		},
+		{
+			code: 'a { top: calc(1px *2); }',
+		},
+		{
+			code: 'a { top: calc(1px* 2); }',
+		},
+		{
 			code: 'a { top: calc(1px / 2); }',
+		},
+		{
+			code: 'a { top: calc(1px/2); }',
+		},
+		{
+			code: 'a { top: calc(1px /2); }',
+		},
+		{
+			code: 'a { top: calc(1px/ 2); }',
 		},
 		{
 			code: 'a { top: calc(1px * -0.2); }',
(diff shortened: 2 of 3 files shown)


lib/rules/function-calc-no-unspaced-operator/__tests__/index.js
```diff
@@ -32,9 +32,27 @@ testRule({
 		{
```
ybiquitous, line 32: [note] Testing `--fix` has been done.

https://github.com/stylelint/stylelint/blob/0364db13241d25f3859cf4f56c8c69f6b0a85093/lib/rules/function-calc-no-unspaced-operator/__tests__/index.js#L8

lib/rules/function-calc-no-unspaced-operator/__tests__/index.js
```diff
@@ -153,6 +171,36 @@ testRule({
...
+		},
+		{
+			code: 'a { top: calc(calc(1px* 2px) + 3px); }',
+		},
+		{
+			code: 'a { top: calc(10px*var(--foo)); }',
+		},
+		{
+			code: 'a { top: calc(10px/var(--foo)); }',
+		},
```
ybiquitous, line 203: [note] The case raised a memory leak. See https://github.com/stylelint/stylelint/issues/6044#issuecomment-1112805765

lib/rules/function-calc-no-unspaced-operator/index.js
```diff
@@ -143,12 +138,7 @@ const rule = (primary, _secondaryOptions, context) => {
 					if (currentNode.type === 'function') {
 						if (context.fix) {
 							needsFix = true;
-							nodes.splice(operatorIndex, 0, {
-								type: 'space',
-								value: ' ',
-								sourceIndex: 0,
-								sourceEndIndex: 1,
-							});
+							currentNode.value = isBeforeOp ? `${currentNode.value} ` : ` ${currentNode.value}`;
```
ybiquitous, line 141: Oh, dropping `*` and `/` may make the mutation unnecessary. I'll try removing it.

ybiquitous: I've reconsidered. Test cases may not be enough. But, just in case, I leave it.

jeddy3 (approved):
Thanks for fixing the memory leak @ybiquitous !

I've tweaked the README, but everything else LGTM.

Feel free to merge.

lib/rules/function-calc-no-unspaced-operator/README.md
```diff
@@ -11,6 +11,8 @@ a { top: calc(1px + 2px); }
 
 Before the operator, there must be a single whitespace or a newline plus indentation. After the operator, there must be a single whitespace or a newline.
 
+Note: The `*` and `/` operators do not require whitespace (but it is usually recommened for consistency).
+
```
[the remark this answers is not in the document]
ybiquitous: @jeddy3 Thanks for the review.

What if adding the following description? Is it easy to understand? 

```suggestion

We recommend pretty printers like [Prettier](https://prettier.io/) if you want to add whitespace also for the `*` and `/` operators.
```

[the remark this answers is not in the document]
jeddy3: I don't think there's a need as the rule sits in the "avoid errors" category. It's not a stylistic rule. We also have a [general recommendation in the main README](https://github.com/stylelint/stylelint/blob/main/README.md#how-itll-help-you) to pair Stylelint with a pretty-printer like Prettier.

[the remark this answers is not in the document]
ybiquitous: Make sense. Thanks for the feedback. 👍🏼

ybiquitous:
Changelog entry added:

> - Fixed: function-calc-no-unspaced-operator false positives and memory leak ([#6045](https://github.com/stylelint/stylelint/pull/6045)).

PR #6047, opened 2022-04-29
Fix end positions for `no-duplicate-*`

mattxwang:

I have a couple of questions on how to implement the end positions for these two rules, mostly due to my lack of experience. Would like some opinions!

For `no-duplicate-at-import-rules`:

- what should be the start and end position? The rejection message just has the URI, but the README contains the entire at rule string (i.e. `@import ...`). I've currently opted for the latter, since it seems more useful for editor support.
- if I should be implementing the latter, is writing `atRule.toString()` correct? I feel a bit strange writing this. Should I be recreating it with `raws` instead?

For `no-duplicate-selectors`:

- currently, I have implemented the start/end positions that break the tests the least; namely, that we highlight the first of the duplicated selectors. Is this the correct method, or should I be doing something else? For example, reporting each instance of the duplicate selectors, or the last one?
- I had to adjust a test (see below) - is my adjustment correct?

Once these are clarified, I can add `endLine` and `endColumn` to each of the test cases for standard CSS syntax.


lib/rules/no-duplicate-selectors/__tests__/index.js
```diff
@@ -81,8 +85,10 @@ testRule({
 			warnings: [
 				{
 					message: messages.rejected('a', 1),
-					line: 1,
+					line: 2,
```
mattxwang, line 88: I believe this `line` was previously incorrect; both `a` selectors are on lines `2` and `3` respectively, so I'm not sure how `1` would be reported.

If that is true, what does this mean for the test process? Is it because we didn't report a start index either?

PR #6048, opened 2022-04-29
Add `validateTypes` to `stylelint.utils.*`

### LINKED ISSUE #5669: Add `validateTypes` to `stylelint.utils.*`
It would probably make sense if the validateTypes were exposed for downstream packages to consume.

Refs https://github.com/stylelint-scss/stylelint-scss/pull/554

EDIT: looking at stylelint-scss code, it seems they duplicate a lot more utils.

mattxwang:

Closes #5669.

How should I be adding types to `index.d.ts` (if at all)? I assume manually listing each validator is not the correct approach, nor is using `typeof validateTypes` - but I don't have much experience here.

Otherwise, I'm not sure if I'm overlooking anything in the PR - is there a test I should be adding, or more I should be doing here?

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,6 +10,7 @@ const ruleMessages = require('./utils/ruleMessages');
 const rules = require('./rules');
 const standalone = require('./standalone');
 const validateOptions = require('./utils/validateOptions');
+const validateTypes = require('./utils/validateTypes');
 const resolveConfig = require('./resolveConfig');
 
 /** @type {import('stylelint').PublicApi} */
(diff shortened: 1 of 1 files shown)


ybiquitous:
I reconsider whether we need to expose such small utilities used internally. 🤔 

For example, I've recently added the second argument to `assert()` as a part of refactoring.
If we would provide the function as a public API, I would need to change also the `assert()` type definition and type test cases at that time; it would increase refactoring cost.

Since such small utilities might be replaced with another library like Lodash in the future, I think we need to estimate carefully the cost between the core's maintainability and the plugins' usability. Once we expose an API, we cannot easily change it in the future.

What do you think?

jeddy3:
Yes, it's not a clear-cut decision. I suggest we leave this as a draft for now and wait a while until we feel like there's unlikely to be more refactoring in and around this util.

mattxwang:
Going to close this due to inactivity + a now-existing merge conflict. Don't want to pollute the PR tab for now!

PR #6049, opened 2022-04-29
Fix end positions in `no-descending-specificity` (and refactor index)

mattxwang:

Part of umbrella issue #5694.

I wanted to make a separate PR for this since I change part of the rule body.

I noticed that the only use of the `index` parameter for `checkSelector` was to generate the report; if we use the new `word` parameter, I think we don't need to manually calculate the index, etc. After removing this, the test behaviours haven't changed, but I'm not sure if I've missed any knock-on side effects (or if I'm inadvertently changing how duplicates are dealt with). Any thoughts welcome!


ybiquitous:
https://github.com/stylelint/stylelint/blob/41bc1c3fb153622935704f21d06a690c5d27c2df/lib/rules/no-descending-specificity/index.js#L79

I assume "The edge-case of duplicate selectors" looks like below:

```css
a:hover {} a, b, a {}
/*         ↑     ↑
 */
```

In the case above, I think this rule reports the first `a` and the second `a` has different column positions.

mattxwang:
Gotcha. I think my change doesn't change the behaviour of the rule in that regard, so it should be good to review!

ybiquitous:
I understand the behavior doesn't change. 👍🏼 

I'm concerned about the following edge-case:

```js
{
	code: 'a:hover {} a, b, a {}',
	warnings: [
		{
			message: messages.rejected('a', 'a:hover'),
			line: 1,
			column: 12,
			endLine: 1,
			endColumn: 13,
		},
		{
			message: messages.rejected('a', 'a:hover'),
			line: 1,
			column: 18,    // but actual: 12
			endLine: 1,
			endColumn: 19, // but actual: 13
		},
	],
},
```

To address this case, we may need to iterate a selector node of `postcss-selector-parser` instead of `ruleNode.selectors`:

https://github.com/stylelint/stylelint/blob/391af98da4e18528c08e08a05c20ca04a83b5aa5/lib/rules/no-descending-specificity/index.js#L71

because we need an exact position of each selector to report a more accurate column position.

The implementation would be a bit too complex for the edge case.

ybiquitous (approved):
Thank you. LGTM 👍🏼

mattxwang:
Thanks for the quick review @ybiquitous, and I think I understand now.

To combine what you've mentioned both here and in #6072 - should I write an issue and look to change our implementation to use `postcss-selector-parser` in the long-term (even if it's not immediate)?

ybiquitous:
@mattxwang Thanks for the suggestion. I think both PRs are ready to merge, but it would be helpful if you could open an issue. 👍🏼

mattxwang:
> ... it would be helpful if you could open an issue. 👍🏼

Thanks for the suggestion - have opened #6075, let me know what you think!

ybiquitous:
Thank you!

ybiquitous:
Changelog entry added:

> Fixed: no-descending-specificity end positions (#6049).

PR #6053, opened 2022-05-01
Fix false positives for non-standard variables in `function-calc-no-unspaced-operator`

### LINKED ISSUE #6052: Fix false positives for SCSS variables in function-calc-no-unspaced-operator

```scss
textarea {
  min-height: calc(#{$line-height} * 5);
}
```

```js
module.exports = {
  plugins: [
    'stylelint-scss',
  ],
  extends: ['stylelint-config-standard-scss'],
  rules: { },
}

```

CLI with `stylelint 'css/**/*.scss'`

`14.8.1`

Behavior like in previous versions <= 14.8.0

### What actually happened?

```shell
css/components/_forms.scss
 213:27  ✖  Expected single space before "-" operator  function-calc-no-unspaced-operator
 213:28  ✖  Expected single space after "-" operator   function-calc-no-unspaced-operator

```

Yes it happens with SCSS variables inside calc()

### Proposal to fix the bug

This happens only in the newest version 14.8.1 with this fixed: https://github.com/stylelint/stylelint/pull/6045

ybiquitous:

Closes #6052

This pull request aims to fix false positives for non-standard variables (e.g. SCSS, Less, and so on.) in the `function-calc-no-unspaced-operator` rule.

I think [`isStandardSyntaxValue()`](https://github.com/stylelint/stylelint/blob/e017583128863ee75634cb15e721779163b571e5/lib/utils/isStandardSyntaxValue.js) is more appropriate than [`isStandardSyntaxMathFunction()`](https://github.com/stylelint/stylelint/blob/e017583128863ee75634cb15e721779163b571e5/lib/utils/isStandardSyntaxMathFunction.js) suggested on <https://github.com/stylelint/stylelint/issues/6052#issuecomment-1114004679> because `isStandardSyntaxMathFunction()` fails the cases using non-standard variables. E.g.

https://github.com/stylelint/stylelint/blob/e017583128863ee75634cb15e721779163b571e5/lib/rules/function-calc-no-unspaced-operator/__tests__/index.js#L535-L543

### DIFF
--- a/lib/rules/function-calc-no-unspaced-operator/index.js
+++ b/lib/rules/function-calc-no-unspaced-operator/index.js
@@ -4,6 +4,7 @@ const valueParser = require('postcss-value-parser');
 
 const declarationValueIndex = require('../../utils/declarationValueIndex');
 const getDeclarationValue = require('../../utils/getDeclarationValue');
+const isStandardSyntaxValue = require('../../utils/isStandardSyntaxValue');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
 const setDeclarationValue = require('../../utils/setDeclarationValue');
@@ -165,6 +166,8 @@ const rule = (primary, _secondaryOptions, context) => {
 				assert(firstNode);
 
 				if (firstNode.type !== 'word') return false;
+
+				if (!isStandardSyntaxValue(firstNode.value)) return false;
 
 				const operatorIndex = firstNode.value.search(OPERATOR_REGEX);
 				const operator = firstNode.value.slice(operatorIndex, operatorIndex + 1);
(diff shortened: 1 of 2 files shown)


lib/rules/function-calc-no-unspaced-operator/index.js
```diff
@@ -166,6 +167,8 @@ const rule = (primary, _secondaryOptions, context) => {
 
 				if (firstNode.type !== 'word') return false;
 
+				if (!isStandardSyntaxValue(firstNode.value)) return false;
+
 				const operatorIndex = firstNode.value.search(OPERATOR_REGEX);
```
ybiquitous, line 172: [note] If the `isStandardSyntaxValue()` check is not performed, `firstNode.value` like `#{$foo-bar}` matches `/[+-]/` (`OPERATOR_REGEX`):

https://github.com/stylelint/stylelint/blob/e017583128863ee75634cb15e721779163b571e5/lib/rules/function-calc-no-unspaced-operator/index.js#L26

jeddy3 (approved):
LGTM, thank you!

ybiquitous:
Changelog entry added:

> - Fixed: `function-calc-no-unspaced-operator` false positives for non-standard variables ([#6053](https://github.com/stylelint/stylelint/pull/6053)).

PR #6068, opened 2022-05-06
Remove `normalize-selector` module

ybiquitous:

None.

`postcss-selector-parser` can normalize selectors instead of [`normalize-selector`](https://github.com/getify/normalize-selector).

See <https://github.com/postcss/postcss-selector-parser/blob/v6.0.10/API.md#processprocesssyncselectors-options>

image


mattxwang (approved):
LGTM - always good to slim down the dependency tree!

ybiquitous:
Changelog entry added:

> Fixed: dependency on `normalize-selector` removed (#6068).

PR #6072, opened 2022-05-07
Fix end positions for `no-invalid-*`

mattxwang:

For `no-invalid-double-slash-comments`, there was one test case I was confused about:

```js
code: 'a, //div { color: pink; }',
description: 'between rules',
message: messages.rejected,
line: 1,
column: 1,
endLine: 1,
endColumn: 26,
```

I haven't changed the behaviour of this test case, but to me I'm not sure why it's `column: 1` - the comment starts at index `4` (where the double-slash is). I might be misunderstanding how this rule works, though!

### DIFF
--- a/lib/rules/no-invalid-double-slash-comments/__tests__/index.js
+++ b/lib/rules/no-invalid-double-slash-comments/__tests__/index.js
@@ -28,6 +28,8 @@ testRule({
 			message: messages.rejected,
 			line: 1,
 			column: 1,
+			endLine: 1,
+			endColumn: 22,
 		},
 		{
 			code: 'a {\n//color: pink;\n}',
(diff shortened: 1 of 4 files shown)


ybiquitous:
> For `no-invalid-double-slash-comments`, there was one test case I was confused about:

I think there is the same reason as <https://github.com/stylelint/stylelint/pull/6049#issuecomment-1120510011>. To know the exact position of each selector, we need to use a node of `postcss-selector-parser` instead of `ruleNode.selectors` (just a string array).

ybiquitous (approved):
I think `a, //div { color: pink; }` is an edge-case and wouldn't be a big problem.
LGTM. Thank you 👍🏼

ybiquitous:
Changelog entry added:

> Fixed: `no-invalid-*` end positions (#6072).

PR #6073, opened 2022-05-07
Add display-mode media query

### LINKED ISSUE #6071: Fix false positives for display-mode in media-feature-name-no-unknown

The media-feature-name-no-unknown rule doesn't allow the new [display-mode](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/display-mode) media query feature. 

The media query display-mode to be allowed by the media-feature-name-no-unknown rule.

andreypelykh:
This is a new media query added to CSS Media Queries Level 5 to test the display mode of an application.

Closes #6071


lib/reference/keywordSets.js
```diff
@@ -677,6 +677,7 @@ keywordSets.mediaFeatureNames = uniteSets(keywordSets.deprecatedMediaFeatureName
 	'color-gamut',
 	'color-index',
 	'dynamic-range',
+	'display-mode',
```
ybiquitous, line 680: Sort in alphabetical order, please?

```suggestion
	'display-mode',
	'dynamic-range',
```

ybiquitous (commented):
@andreypelykh Thank you for opening the PR, and it works expectedly. 👍🏼 

Could you address the small suggestion, please?

lib/reference/keywordSets.js
andreypelykh: @ybiquitous You are welcome! Thanks for the suggestion.

ybiquitous (approved):
@andreypelykh Thanks, LGTM 👍🏼

ybiquitous:
Changelog entry added:

> Fixed: media-feature-name-no-unknown false positives for display-mode (#6073).

PR #6081, opened 2022-05-13
Use arrow functions in docs

ybiquitous:

Arrow Function is now widely-used and becomes simpler code.
This change replaces `function` expressions with arrow functions in docs.


mattxwang (approved):
Each of these changes seem reasonable to me!

jeddy3 (approved):
LGTM, thanks!

PR #6093, opened 2022-05-14
Fix `at-rule-no-unknown` false positives for `@layer`

### LINKED ISSUE #6084: [at-rule-no-unknown] incorrectly flags `@layer`

Configure stylelint to use at-rule-no-unknown, then try to use an `@layer` rule in your css.

```json
{
	"at-rule-no-unknown": true
}
```

stylelint 'src/**/*.css'

14.8.2

Should not error

### What actually happened?

```
 2:1  ✖  Unexpected unknown at-rule "@layer"  at-rule-no-unknown
```

_No response_

### Proposal to fix the bug

`@layer` is in [W3C Candidate Recommendation Snapshot](https://www.w3.org/TR/css-cascade-5/) phase, so I believe that means it should not be flagged by this rule.

immitsu:

Resolve #6084 

No

### DIFF
--- a/lib/reference/keywordSets.js
+++ b/lib/reference/keywordSets.js
@@ -642,6 +642,7 @@ keywordSets.atRules = uniteSets(keywordSets.pageMarginAtRules, [
 	'font-feature-values',
 	'import',
 	'keyframes',
+	'layer',
 	'media',
 	'namespace',
 	'nest',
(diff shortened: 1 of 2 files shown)


ybiquitous (approved):
@immitsu Thank you. LGTM 👍🏼

ybiquitous:
Changelog entry added:

> Fixed: `at-rule-no-unknown` false positives for `@layer` ([#6093](https://github.com/stylelint/stylelint/pull/6093)).

PR #6094, opened 2022-05-14
Fix `no-invalid-position-at-import-rule` false positives for `@layer`

### LINKED ISSUE #6085: [no-invalid-position-at-import-rule] Should allow `@layer` to come before `@import`

```css
/* establish a layer order up-front, from lowest to highest priority */
@layer reset, defaults, theme, patterns, components, utilities, overrides;

@import url('colors.css') layer(theme);
```

```json
{
	"no-invalid-position-at-import-rule": true,
}
```

stylelint 'src/**/*.css'

`14.8.2`

This is valid syntax according to the spec: https://www.w3.org/TR/css-cascade-5/#layer-empty

### What actually happened?

```
 4:1  ✖  Unexpected invalid position @import rule  no-invalid-position-at-import-rule
```

No

### Proposal to fix the bug

The rule should be updated to allow `@layer` to come before `@import`, and the rule's docs should be updated as well to reflect this.

immitsu:

Resolve #6085 

No.

### DIFF
--- a/lib/rules/no-invalid-position-at-import-rule/README.md
+++ b/lib/rules/no-invalid-position-at-import-rule/README.md
@@ -50,6 +50,12 @@ a {}
 @import 'foo.css';
 ```
 
+<!-- prettier-ignore -->
+```css
+@layer default;
+@import url(theme.css) layer(theme);
+```
+
 ## Optional secondary options
 
 ### `ignoreAtRules: ["/regex/", /regex/, "string"]`
--- a/lib/rules/no-invalid-position-at-import-rule/README.md
+++ b/lib/rules/no-invalid-position-at-import-rule/README.md
@@ -10,7 +10,7 @@ a {}
   * This @import */
 ```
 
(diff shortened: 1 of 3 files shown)


ybiquitous (changes requested):
@immitsu Thanks for creating the pull request. Almost LGTM 👍🏼 

Can you please update the doc also "ignoring `@charset`"? ⬇️ 

https://github.com/stylelint/stylelint/blob/a08312b80efa011f9302fb94d1f4de017754a19f/lib/rules/no-invalid-position-at-import-rule/README.md?plain=1#L13

immitsu:
@ybiquitous Done!

ybiquitous (approved):
Thank you!

ybiquitous:
Changelog entry added:

> Fixed: `no-invalid-position-at-import-rule` false positives for `@layer` ([#6094](https://github.com/stylelint/stylelint/pull/6094)).

PR #6098, opened 2022-05-18
Fix `length-zero-no-unit` autofix for `.0` values

### LINKED ISSUE #6074: `length-zero-no-unit`  autofix not working

.demo{
  position: absolute;
  right: .0rem;
  top: -.52rem;
}

const config = {
  extends: ['stylelint-config-standard'],
  plugins: ['stylelint-scss'],
  rules: {}
};

module.exports = config;

stylelint --fix

14.8.2

.demo{
  position: absolute;
  right: 0;
  top: -.52rem;
}

### What actually happened?

.demo{
  position: absolute;
  right: ;
  top: -.52rem;
}

_No response_

### Proposal to fix the bug

_No response_

playhardgopro:

Closes #6074

I tested these changes in the reproduction repository and
- works well with .0rem (.0rem -> 0)
- doesn't work with .0 (still .0 -> '')

So should we fix it for the second case? It doesn't seem to be the problem with this rule. @ybiquitous correct me if I'm wrong.

### DIFF
--- a/lib/rules/length-zero-no-unit/__tests__/index.js
+++ b/lib/rules/length-zero-no-unit/__tests__/index.js
@@ -674,6 +674,15 @@ testRule({
 			endColumn: 27,
 		},
 		{
+			code: 'a { right: .0rem; }',
+			fixed: 'a { right: 0; }',
+			message: messages.rejected,
+			line: 1,
+			column: 14,
+			endLine: 1,
+			endColumn: 17,
+		},
+		{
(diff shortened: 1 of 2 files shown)


ybiquitous (approved):
@playhardgopro Thank you for your first contribution! The fix looks good to me. 👍🏼 

The case of `.0` is not related to `length-zero-no-unit`, so there are no problems.

PR #6102, opened 2022-05-20
Add `import-notation` rule

### LINKED ISSUE #4055: Add import-notation

We're creating web components using Shadow DOM. In many situations, the structure and style creation of our components is done off-tree and then the component is appended to the DOM. At this point, the stylesheets are parsed and the component is rendered.

If, when the component is appended to the DOM, one of the stylesheets contains an `@import url(...);` rule, parsing the remaining stylesheet is blocked until the import is completed. Therefore, we'd like to disallow `@import url()`, but allow `@import './otherfile.scss'`, which doesn't seem possible (or at least the docs doesn't make it clear how this can be done).

An emerging standard for Constructable Stylesheets also sees the introduction of a new function called `replaceSync(cssText)`, which will throw if any `@import` rules are encountered when parsing the rules.  Having a linting rule for this would mean that anyone trying to import stylesheets from a URL would be made aware at dev time rather than run time.

A new rule that disables `@import url()`.  My suggestion for the name is `no-async`, both as a future proof rule name and as a nod to the `replaceSync()` method.

playhardgopro:

Closes #4055 

No, it's self-explanatory.


lib/rules/import-notation/index.js
```diff
@@ -0,0 +1,110 @@
+'use strict';
+
+const report = require('../../utils/report');
+const ruleMessages = require('../../utils/ruleMessages');
+const validateOptions = require('../../utils/validateOptions');
+const valueParser = require('postcss-value-parser');
```
ybiquitous, line 6: [suggest] Separate external and internal `require()` between an empty line.

```suggestion
const valueParser = require('postcss-value-parser');

const report = require('../../utils/report');
const ruleMessages = require('../../utils/ruleMessages');
const validateOptions = require('../../utils/validateOptions');
```

lib/rules/import-notation/index.js
```diff
@@ -0,0 +1,110 @@
...
+		const validOptions = validateOptions(result, ruleName, {
+			actual: primary,
+			possible: ['string', 'url'],
+		});
+
+		if (!validOptions) {
+			return;
+		}
+
+		root.walkAtRules('import', checkAtRuleImportParams);
```
ybiquitous, line 33: [suggestion] Better to support case-insensitive `@import`.

```suggestion
		root.walkAtRules(/^import$/i, checkAtRuleImportParams);
```

ybiquitous (changes requested):
@playhardgopro Thank you for the contirubtion! This addition looks almost good to me. 👍🏼 

I've left suggestions about trivial things. Please consider fixing them.

In addition, can you please also the following doc to add a new section `Imports` and a link to the new rule?

https://github.com/stylelint/stylelint/blob/e1e28fec4b4f1a98fc94d81b7ffa0cdb393242ac/docs/user-guide/rules/list.md?plain=1#L94

playhardgopro:
@ybiquitous I added all suggested changes, thanks for review!

docs/user-guide/rules/list.md
```diff
@@ -130,6 +130,10 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 - [`function-url-scheme-allowed-list`](../../../lib/rules/function-url-scheme-allowed-list/README.md): Specify a list of allowed URL schemes.
 - [`function-url-scheme-disallowed-list`](../../../lib/rules/function-url-scheme-disallowed-list/README.md): Specify a list of disallowed URL schemes.
 
+### Imports
+
+- [`import-notation`](../../../lib/rules/import-notation/README.md): Specify string or url notation for imports.
```
ybiquitous, line 135:
```suggestion
- [`import-notation`](../../../lib/rules/import-notation/README.md): Specify string or url notation for imports (Autofixable).
```

lib/rules/import-notation/README.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+<!-- prettier-ignore -->
+```css
+@import "foo.css";
+```
+
+### `"url"`
+
+Imports _must_ use _url_ function.
```
ybiquitous, line 63:
```suggestion
Imports _must_ use the `url` function.
```

ybiquitous (commented):
@playhardgopro Thank you! Additionally, I suggest trivial fixes on the doc.

ybiquitous (approved):
Thanks, LGTM 👍🏼

ybiquitous:
Just in case, I'd like to wait for @jeddy3's review.

lib/rules/import-notation/index.js
```diff
@@ -0,0 +1,111 @@
...
+const ruleMessages = require('../../utils/ruleMessages');
+const validateOptions = require('../../utils/validateOptions');
+const setAtRuleParams = require('../../utils/setAtRuleParams');
+const getAtRuleParams = require('../../utils/getAtRuleParams');
+const atRuleParamIndex = require('../../utils/atRuleParamIndex');
+
+const ruleName = 'import-notation';
+
+const messages = ruleMessages(ruleName, {
+	expected: (importUrlOrString, primary) => `Expected "${importUrlOrString}" to be a "${primary}"`,
```
jeddy3, line 15:
```suggestion
	expected: (unfixed, fixed) => `Expected "${unfixed}" to be "${fixed}"`,
```

My bad, I got this wrong in [the blueprint](https://github.com/stylelint/stylelint/issues/4055#issuecomment-1014768867).

We use the:
- "Expected x to be y" message format when showing the fixed values, e.g. https://github.com/stylelint/stylelint/blob/9d045d928699c0d84b86e08a3d8b035343b06ca4/lib/rules/alpha-value-notation/index.js#L18
- "Expected primary option" message format when we don't, e.g. https://github.com/stylelint/stylelint/blob/9d045d928699c0d84b86e08a3d8b035343b06ca4/lib/rules/color-function-notation/index.js#L17

We should use the former here as:
- it's generally more useful for users
- the URL/string parts of `@import` rules are generally short

e.g. "Expected '"foo.css"' to be 'url("foo.css")' (import-notation)"

We'll need to use single quotes if there are double quotes in the value, and vice-versa.

lib/rules/import-notation/__tests__/index.js
```diff
@@ -0,0 +1,201 @@
...
+	accept: [
+		{
+			code: '@import "foo.css";',
+		},
+		{
+			code: "@import 'foo.css';",
+		},
+		{
+			code: '@import foo.css;',
+		},
```
jeddy3, line 109:
```suggestion
```

Let's ignore invalid CSS.

lib/rules/import-notation/__tests__/index.js
```diff
@@ -0,0 +1,201 @@
...
+			fixed: "@import 'foo.css';",
+			message: messages.expected("url( 'foo.css' )", 'string'),
+			line: 1,
+			column: 9,
+			endLine: 1,
+			endColumn: 25,
+		},
+		{
+			code: '@import url(foo.css);',
+			fixed: '@import foo.css;',
```
jeddy3, line 175:
```suggestion
			fixed: '@import "foo.css";',
```

Let's default to double quotes.

lib/rules/import-notation/__tests__/index.js
```diff
@@ -0,0 +1,201 @@
...
+			message: messages.expected('url(foo.css)', 'string'),
+			line: 1,
+			column: 9,
+			endLine: 1,
+			endColumn: 21,
+		},
+		{
+			code: '@import url(foo.css) print;',
+			fixed: '@import foo.css print;',
```
jeddy3, line 184:
```suggestion
			fixed: '@import "foo.css" print;',
```

Let's default to double quotes (otherwise we're generating invalid CSS).

lib/rules/import-notation/__tests__/index.js
```diff
@@ -0,0 +1,201 @@
...
+	reject: [
+		{
+			code: '@import foo.css;',
+			fixed: '@import url(foo.css);',
+			message: messages.expected('foo.css', 'url'),
+			line: 1,
+			column: 9,
+			endLine: 1,
+			endColumn: 16,
+		},
```
jeddy3, line 55:
```suggestion
```

Let's not fixed invalid CSS.

jeddy3 (changes requested):
@playhardgopro Thanks for the pull request! It's looking good.

I've requested some changes. The two biggest are:
- the structure of the problem message
- how unquoted values are handled

The other changes are mainly trivial doc changes to be consistent with other rules.

jeddy3 (approved):
@playhardgopro Fantastic! Thanks for making all the changes.

LGTM.

jeddy3:
Changelog entry:

> - Added: `import-notation` rule ([#6102](https://github.com/stylelint/stylelint/pull/6102)).

PR #6124, opened 2022-05-30
Added online version of autoprefixer.

rahulkumarsingh73690:
Added online version of Autoprefixer. So users can instantly use it without installing it.

### DIFF
--- a/lib/rules/value-no-vendor-prefix/README.md
+++ b/lib/rules/value-no-vendor-prefix/README.md
@@ -11,7 +11,7 @@ a { display: -webkit-flex; }
 
 This rule ignores non-standard vendor-prefixed values that aren't handled by [Autoprefixer](https://github.com/postcss/autoprefixer).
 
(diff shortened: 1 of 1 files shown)


ybiquitous (commented):
@rahulkumarsingh73690 Thanks for creating the pull request.

It seems other rule docs (e.g. `lib/rules/at-rule-no-vendor-prefix/README.md`) would need such a link to the online version if we would add it.

Also, the Autoprefixer README has a link to its own online version ([autoprefixer.github.io](https://autoprefixer.github.io/)):

https://github.com/postcss/autoprefixer/blob/616c44240a1478e54d8b7bc79ed8fa5db6392d11/README.md?plain=1#L29

Thus, I think it's unnecessary to add such a link. What do you think?

ntwb:
Let’s not, @rahulkumarsingh73690 as I noted in one of your other pull requests that by not disclosing you’re the creator of the tool and site makes this pretty spammy, also as you’ve also created ~50+ PRs to projects to add a link to your site…

PR #6147, opened 2022-06-10
Fix rule message function type to not require users to handle all kind of arguments

devversion:
Currently, users writing custom rules and using `RuleMessageFunc` are required to handle all kinds of parameters with TypeScript's strict mode enabled. e.g.

```ts
const messages = utils.ruleMessages(ruleName, {
  invalid: (rule: string) =>
    `Failed to extract namespace from ${rule}. material/no-unused-` +
    `imports Stylelint rule likely needs to be updated.`,
});
```

results in

```
tools/stylelint/no-unused-import.ts:7:12 - error TS2322: Type '(rule: string) => string' is not assignable to type 'string | RuleMessageFunc'.
  Type '(rule: string) => string' is not assignable to type 'RuleMessageFunc'.
    Types of parameters 'rule' and 'args' are incompatible.
      Type 'string | number | boolean | RegExp' is not assignable to type 'string'.

7   invalid: (rule: string) =>
             ~~~~~~~~~~~~~~~~~
8     `Failed to extract namespace from ${rule}. material/no-unused-` +
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
9     `imports Stylelint rule likely needs to be updated.`,
```

This can be observed in my small playground as well:

https://www.typescriptlang.org/play?#code/C4TwDgpgBASgrgGwgWQgZzQQwOYQGJwB2AxlALxQAUAdLZgE7ZoBcVaw9AlodlAD6wI2AKIAPMAEoA2gF0J5AHxR2XHgG4AUBoBmRYsE4B7QlAC2IABIQEkepV0lW8JKgw58e+QG8AvlvNWNhB2lPSIEABymKYQrCrc2PJkSgAGAILAUGFIUAAkXtmR0RA+UJxoUNqYnAgJAIQpEmpAA

Not specifying any type constraints, allows the user to choose the quantity and types of arguments themselves.  TypeScript will still be able to narrow the function type when it is re-exposed through the generic in `utils.ruleMessages`.

None

See description above

### DIFF
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -137,7 +137,15 @@ declare module 'stylelint' {
 			newline?: string | undefined;
 		};
 
-		export type RuleMessageFunc = (...args: (string | number | boolean | RegExp)[]) => string;
+		// Note: With strict function types enabled, function signatures are checked contravariantly.
+		// This means that it would not be possible for rule authors to narrow the message function
+		// parameters to e.g. just `string`. Declaring the type for rule message functions through
+		// method declarations tricks TypeScript into bivariant signature checking. More details can
+		// be found here: https://stackoverflow.com/questions/52667959/what-is-the-purpose-of-bivariancehack-in-typescript-types.
+		// and in the original discussion: https://github.com/stylelint/stylelint/pull/6147#issuecomment-1155337016.
+		export type RuleMessageFunc = {
+			bivariance(...args: (string | number | boolean | RegExp)[]): string;
+		}['bivariance'];
 
 		export type RuleMessages = { [message: string]: string | RuleMessageFunc };
 
--- a/types/stylelint/type-test.ts
+++ b/types/stylelint/type-test.ts
@@ -82,6 +82,7 @@ const ruleName = 'sample-rule';
 const messages = stylelint.utils.ruleMessages(ruleName, {
 	problem: 'This a rule problem message',
 	warning: (reason) => `This is not allowed because ${reason}`,
+	withNarrowedParam: (mixinName: string) => `Mixin not allowed: ${mixinName}`,
 });
 const problemMessage: string = messages.problem;
 const problemFunc: (...reason: string[]) => string = messages.warning;
@@ -115,3 +116,18 @@ const testPlugin: Plugin = (options) => {
 };
 
 stylelint.createPlugin(ruleName, testPlugin);
+
+messages.withNarrowedParam('should work');
+
+// @ts-expect-error Boolean should not be allowed here.
+messages.withNarrowedParam(true);
+// @ts-expect-error Expected a required parameter.
+messages.withNarrowedParam();
+
+// For non-explicit parameters, boolean, strings etc. are always allowed.
+// Not `null` for example though.
+messages.warning(true);
+messages.warning('some string');
+
+// @ts-expect-error Null is not allowed in `RuleMessageFunc` parameters.
+messages.warning(null);


ybiquitous:
@devversion Thanks for creating the pull request.

If `RuleMessageFunc` would allow `any[]`, the following code would pass type-check.

```js
// We don't want to allow `null` or `undefined`! 
ruleMessages.expected("string", null, undefined);
```

Can you remove the type annotation from your code as below?

```diff
const messages = utils.ruleMessages(ruleName, {
-  invalid: (rule: string) =>
+  invalid: (rule) =>
    `Failed to extract namespace from ${rule}. material/no-unused-` +
    `imports Stylelint rule likely needs to be updated.`,
});
```

devversion:
@ybiquitous Thanks for looking at this. Yes, removing the explicit type is the trick we are using right now.

Also yeah, this modification would allow users to _explicitly_ accept `null` or other types (but only if the custom function would allow so). If the user would not specify any parameter, TS would narrow the parameters to "()" and nothing could be passed. TS always narrows the functions passed to `utils.ruleMessages` but previously it just required the custom message function patterns to  handle `string`, `RegExp` and all the others. Here is an example;

https://www.typescriptlang.org/play?#code/C4TwDgpgBASgrgGwgWQgZzQQwOYQGJwB2AxlALxQAUAdLZgE7ZoBcUmhIA2gLoCU5APihpg9AJaFsAbgBQMgGZFiwMQHtCUALYgAEhASR6AHgAqUCAA9gEQgBM0sRCnRZcBEgMqKSrE719QAN4yUKFQ9BDAcPQa3sQyAL5yxOoiUIQM9KoA7hC2eBoU2noGEPSUlPROAHKYmhCsIuKS-GRCAAYAgsDhTlAAJIFVSLX1CVBiDvKYYggS2ACE7byyMhn0Wbn5hJSEiAgrQA

and

https://www.typescriptlang.org/play?#code/C4TwDgpgBASgrgGwgWQgZzQQwOYQGJwB2AxlALxQAUAdLZgE7ZoBcUmhIA2gLoCU5APihpg9AJaFsAbgBQMgGZFiwMQHtCUALYgAEhASR6AHgAqUCAA9gEQgBM0sRCnRZcBEgMqKSrE719QAN4yUKFQ9BDAcPQa3sQyAL5yxOoiUIQM9KoA7hC2eBoU2noGEPSUlPxkQgAGAILA4U5QYg7ymGIIEtgAhDW8sjIZ9Fm5+YSUhIgIA0A

I definitely might miss something as I don't strictly know what Stylelint would do with the function, except for wrapping it and (just expecting it to return a `string`)

ybiquitous:
@devversion Thank you for the detailed explanation. It makes sense. 👍🏼 

I'm reconsidering that the current argument type `(string | number | boolean | RegExp)[]` may be a bit too strict. 🤔 

If loosening the type to `any[]`, plugin authors could narrow the types they want like this:

- `(foo: string) => string`
- `(foo: number, bar: RegExp) => string`
- etc.

Of course, if authors would not specify type annotations to arguments, their message functions might allow invalid values like `undefined` or `null` due to `any[]`.

Considering pros and cons, I think `any[]` has more benefits. But I would like to hear what others have to say.

ybiquitous:
Node: An example of narrowing argument types in the stylelint codebase ⬇️ 

```diff
--- a/lib/rules/alpha-value-notation/index.js
+++ b/lib/rules/alpha-value-notation/index.js
@@ -15,6 +15,10 @@ const { isRegExp, isString, assert } = require('../../utils/validateTypes');
 const ruleName = 'alpha-value-notation';
 
 const messages = ruleMessages(ruleName, {
+	/**
+	 * @param {string} unfixed
+	 * @param {string} fixed
+	 */
 	expected: (unfixed, fixed) => `Expected "${unfixed}" to be "${fixed}"`,
 });
```

https://github.com/stylelint/stylelint/blob/b7ff75786e20237b0f42c8bee53726cb04d4e0ef/lib/rules/alpha-value-notation/index.js#L18

devversion:
@ybiquitous I found a way to make the function arguments being treated bivariantly. Let me know what you think about this. It is certainly a hack, leveraging how TS decides between bivariant and contravariance (with strict function types).

https://www.typescriptlang.org/play?#code/C4TwDgpgBASgrgGwgWQgZzQQwOYQGJwB2AxlALxQDeUARgJYBumATnZiRABQB0vL2aAFxROaYK0LYoAH1gRsAUQAeYGVEJwAtjQjMAlAG0AunuEMA9nQAmUAL4GARPSat2xCA6MAoLwDMixMB05oRQmiAAEhAIkMwAPAAqUBBKwBCEVmiwiCjoWLgEJAB8nP4kwgmmUEmUXlD1UMwQwHDMoWXEXrY+xCFi6izM5gDuEFZ4hABy5gng0BThUTG6nJyYeuRFUAAGAILAjTlQdFm+mHQIdJIAhNt6ANxevYT9hIMjYxMA6nTAABYAZXEVykC0i0Viq0wwjEEmwGzIWz2B2YRxOUDOFxBtweTz6BzezCGo3GhB+-wAqoQUpBAmMAAosTCachhcHLZhQ4Q-DIjBFI-aHJDHU7nS43O6PLyE4mfKYzOacADkf2iCHMStxMo+pOms0gnA0CAQuOl7xJ31+gOBkmVquNGq15rl5OtcMNiBN9yAA

ybiquitous:
@devversion Thanks for the attempt to improve. It seems what we want now. 👍🏼 

One thing I'm concerned about is if this technique will be valid for future versions of the TypeScript compiler. Is this technique officially supported or is there a resource to mention it? 🤔 

Also, if using this technique, we need to fix the following type error in `utils/ruleMessages.js`:

```
lib/utils/ruleMessages.js:23:40 - error TS2352: Conversion of type '(...args: (string | number | RegExp)[]) => string' to type 'R[keyof T]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.

23    newMessages[messageId] = /** @type {R[K]} */ (
                                          ~~~~
```

We would appreciate it if you could fix the error and add some test cases to `types/stylelint/type-test.ts`.

devversion:
This is a pretty standard field access in the type system. I believe this should not become an issue in the near future. FWIW: React types also leverage this trick: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L85

ybiquitous:
Looks good. Let's change to the bivariance hack!

devversion:
@ybiquitous Updated, let me know what you think. Thanks!

ybiquitous (approved):
Great job! Thank you so much! 👏🏼

jeddy3:
Changelog entry added:

> - Fixed: rule message function type to not require users to handle all kind of arguments ([#6147](https://github.com/stylelint/stylelint/pull/6147)).

PR #6152, opened 2022-06-15
Publicly expose reference data

### LINKED ISSUE #6151: Publicly expose reference data

I'm writing a custom lint that would like to have access to some of the data in `lib/reference` (specifically `shorthandData`). Currently that's not publicly available, although it is used by a number of internal lints which suggests that it's useful for implementing reasonable lints.

Possibly a top-level `reference` export that exposes the various data tables in question.

nex3:
Closes #6151

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project are documented in this file.
 
+## 14.10.0
+
+- Added: `reference` export, which contains reference data from the CSS spec that may be useful for custom lints.
+
 ## 14.9.1
 
 - Fixed: `selector-max-specificity` false positives for `:nth-child` ([#6140](https://github.com/stylelint/stylelint/pull/6140)).
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "stylelint",
-  "version": "14.9.1",
+  "version": "14.10.0",
   "description": "A mighty, modern CSS linter.",
   "keywords": [
     "css-in-js",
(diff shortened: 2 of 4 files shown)


types/stylelint/index.d.ts
```diff
@@ -364,6 +364,77 @@ declare module 'stylelint' {
 			line?: number;
 		};
 
+		export type KeywordSets = {
```
ybiquitous, line 367: [suggest] Could you add a type annotation to each module like this? By doing so, we could improve the code's type safety.

```js
// lib/reference/keywordSets.js

/** @type {import('stylelint').KeywordSets} */
const keywordSets = {};

// ...

module.exports = (keywordSets);
```
```

File: CHANGELOG.md [hunk 2]
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to this project are documented in this file.
 
+## 14.10.0
+
+- Added: `reference` export, which contains reference data from the CSS spec that may be useful for custom lints.
+
```
ybiquitous, line 8: [suggest] Could you revert the change, please? We will update the changelog after merged.

```suggestion
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "stylelint",
-  "version": "14.9.1",
+  "version": "14.10.0",
```
ybiquitous, line 3: [suggest] Could you revert the version, please? It will be automatically updated on release.

```suggestion
  "version": "14.9.1",
```

ybiquitous (changes requested):
@nex3 Thanks for creating the pull request!

I think there are no significant problems. Could you consider addressing my review comments? Of course, you can do it after the discussion.

nex3:
I'm going to close this out in favor of https://github.com/stylelint/stylelint/pull/6168, which is a more principled approach to the same goal.

PR #6153, opened 2022-06-15
Add tally to `string` and `verbose` formatters

### LINKED ISSUE #5590: Add tally to string formatter

I found that the output result of calling stylelint.formatters.string does not contain the prompt of error and warning number. It is necessary to summarize the current lint result：

image

On the contrary, eslint will give an error and warning number prompt at the end:

image

I found that in stylelint / lib / formatters / stringformatter.js, the final result is output as follows:

```
output = output.trim();

	if (output !== '') {
		output = `\n${output}\n\n`;
	}

return output;
```

Can you add error and warning number prompts at the end of output? Or do I need to open a PR to solve this problem?

ybiquitous:

Closes #5590


ybiquitous:
Thank you for the review!

Changelog entry added:

> - Added: tally to `string` and `verbose` formatters ([#6153](https://github.com/stylelint/stylelint/pull/6153)).

PR #6155, opened 2022-06-19
Add `annotation-no-unknown`

### LINKED ISSUE #4508: Add annotation-no-unknown
> Clearly describe the bug

"When there is a spelling mistake for an important declaration it is not reported as parse error"

Parsing error

```css
.a{
color: red !imprtant;
}
```

`Since it is related to parsing, no need for the specific configuration.`

`12.0.0`

 `Node API `

 `No`

 `Parse error needs to be reported.`

`No errors reported`

mattxwang:

Closes #4508.

A late turnaround on this one! I think there's probably a few missing things before this is mergeable.

Possible questions with my implementation:
- how should I be checking for nonstandard syntax? ex just at the declaration level?
- should `!important` be case-insensitive? on one hand, syntax highlighters and the literal reading of the spec only allow for lowercase, but it seems like Chrome respects any capitalization. 
- I've opted to include the `!` in the secondary options; is that too clunky?
- should I add any testcases for nonstandard syntax?

### DIFF
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -11,6 +11,10 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 
 ## Avoid errors
 
+### Annotation
+
+- [`annotation-no-unknown`](../../../lib/rules/annotation-no-unknown/README.md): Disallow unknown annotations.
+
 ### Color
 
 - [`color-no-invalid-hex`](../../../lib/rules/color-no-invalid-hex/README.md): Disallow invalid hex colors.
@@ -90,10 +94,6 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 - [`no-empty-source`](../../../lib/rules/no-empty-source/README.md): Disallow empty sources.
 - [`no-invalid-double-slash-comments`](../../../lib/rules/no-invalid-double-slash-comments/README.md): Disallow double-slash comments (`//...`) which are not supported by CSS.
 - [`no-invalid-position-at-import-rule`](../../../lib/rules/no-invalid-position-at-import-rule/README.md): Disallow invalid position `@import` rules within a stylesheet.
-
-### Annotation
-
-- [`annotation-no-unknown`](../../../lib/rules/annotation-no-unknown/README.md): Disallow unknown annotations.
 
 ## Enforce conventions
 
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -91,6 +91,10 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 - [`no-invalid-double-slash-comments`](../../../lib/rules/no-invalid-double-slash-comments/README.md): Disallow double-slash comments (`//...`) which are not supported by CSS.
 - [`no-invalid-position-at-import-rule`](../../../lib/rules/no-invalid-position-at-import-rule/README.md): Disallow invalid position `@import` rules within a stylesheet.
 
+### Annotation
+
+- [`annotation-no-unknown`](../../../lib/rules/annotation-no-unknown/README.md): Disallow unknown annotations.
+
 ## Enforce conventions
 
 ### Alpha-value
--- a/lib/rules/annotation-no-unknown/README.md
+++ b/lib/rules/annotation-no-unknown/README.md
@@ -40,7 +40,7 @@ a {
 Given:
 
 ```json
-["/^!my-/", "!custom"]
+["/^--foo-/", "--bar"]
 ```
 
 The following patterns are _not_ considered problems:
@@ -48,13 +48,13 @@ The following patterns are _not_ considered problems:
 <!-- prettier-ignore -->
 ```css
 a {
-  color: green !custom;
+  color: green !--foo--bar;
 }
 ```
 
 <!-- prettier-ignore -->
 ```css
 a {
-  color: green !my-custom-annotation;
+  color: green !--bar;
 }
 ```
--- a/lib/rules/annotation-no-unknown/README.md
+++ b/lib/rules/annotation-no-unknown/README.md
@@ -9,7 +9,7 @@ a { color: green !imprtant; }
  * This annotation */
 ```
 
-This rule considers solely the `!important` annotation per the [CSS Cascading and Inheritance spec](https://www.w3.org/TR/css-cascade-3/#importance).
+This rule considers annotations defined in the CSS Specifications, up to and including Editor's Drafts, to be known.
 
 ## Options
 
--- a/lib/rules/annotation-no-unknown/README.md
+++ b/lib/rules/annotation-no-unknown/README.md
@@ -24,7 +24,7 @@ a {
 }
 ```
 
-The following patterns are _not_ considered problems:
+The following pattern is _not_ considered problem:
 
 <!-- prettier-ignore -->
 ```css
--- a/lib/rules/annotation-no-unknown/README.md
+++ b/lib/rules/annotation-no-unknown/README.md
@@ -15,7 +15,7 @@ This rule considers annotations defined in the CSS Specifications, up to and inc
 
 ### `true`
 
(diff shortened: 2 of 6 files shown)


mattxwang:
Actually - going to mark this as draft; I think I've made an implementation mistake that the code coverage sneakily catches. I'll revisit and mark as ready when relevant!

lib/rules/annotation-no-unknown/index.js
```diff
@@ -0,0 +1,88 @@
...
+				if (!isAnnotation(node)) return;
+
+				const value = node.value;
+
+				if (optionsMatches(secondaryOptions, 'ignoreAnnotations', value)) {
+					return;
+				}
+
+				if (keywordSets.annotations.has(value.toLowerCase())) {
+					return;
```
mattxwang, line 63: This does never get called. It "accidentally" works because a valid `!important` is stripped from the declaration somehow. I'll need to investigate more.

lib/rules/annotation-no-unknown/index.js
```diff
@@ -0,0 +1,88 @@
...
+		if (!validOptions) {
+			return;
+		}
+
+		root.walkDecls(checkStatement);
+
+		/**
+		 * @param {import('postcss').Declaration} decl
+		 */
+		function checkStatement(decl) {
```
mattxwang, line 48: Interestingly: if I place a `debugger;` directly under this line, the `value`/`raws` of `decl` never contains `!important` across all the test cases. Seems like I'm doing something wrong!

ybiquitous:
@mattxwang Thanks for creating the pull request! 👍🏼 

Here are my answers to your questions in the description:

> * how should I be checking for nonstandard syntax? ex just at the declaration level?
> * should I add any testcases for nonstandard syntax?

If you would have clear test cases for non-standard syntaxes, it might be nice to check them to reduce potential false positives. But if you wouldn't have them or the rule implementation would be much more complex by adding such a check, it would be acceptable not to check them. 

> * should `!important` be case-insensitive? on one hand, syntax highlighters and the literal reading of the spec only allow for lowercase, but it seems like Chrome respects any capitalization.

It seems good that `!important` be case-sensitive. We have similar code already:

https://github.com/stylelint/stylelint/blob/592b149a6225559cccada30d313814910439d170/lib/utils/getImportantPosition.js#L11

> * I've opted to include the `!` in the secondary options; is that too clunky?

I think it's unnecessary for consistency to include `!` because the `ignoreAtRules` option of `property-no-unknown` doesn't `@`.

https://github.com/stylelint/stylelint/blob/592b149a6225559cccada30d313814910439d170/lib/rules/property-no-unknown/README.md?plain=1#L130-L138

lib/rules/annotation-no-unknown/index.js
ybiquitous: Sorry, I misunderstood. The PostCSS parser assigns the `important` property of a `Declaration` node and omit `!important` from its `value` property if the declaration value includes `!important`. See the example below:

```console
$ node -e 'require("postcss").parse("a { color: red !important }").walkDecls(decl => console.log(decl))'
<ref *1> Declaration {
  raws: { before: ' ', between: ': ' },
  type: 'decl',
  parent: Rule {
    raws: { before: '', between: ' ', semicolon: false, after: ' ' },
    type: 'rule',
    nodes: [ [Circular *1] ],
    parent: Root {
      raws: [Object],
      type: 'root',
      nodes: [Array],
      source: [Object],
      lastEach: 1,
      indexes: [Object],
      [Symbol(isClean)]: false,
      [Symbol(my)]: true
    },
    source: { start: [Object], input: [Input], end: [Object] },
    selector: 'a',
    lastEach: 1,
    indexes: { '1': 0 },
    [Symbol(isClean)]: false,
    [Symbol(my)]: true
  },
  source: {
    start: { offset: 4, line: 1, column: 5 },
    input: Input {
      css: 'a { color: red !important }',
      hasBOM: false,
      id: '<input css 0Is-DD>',
      [Symbol(fromOffsetCache)]: [Array]
    },
    end: { offset: 24, line: 1, column: 25 }
  },
  prop: 'color',
  important: true,
  value: 'red',
  [Symbol(isClean)]: false,
  [Symbol(my)]: true
}
```

So, we may need to treat `!important specially. See also the PostCSS code:

[specially](https://github.com/postcss/postcss/blob/04877f35f481df17737147351b4122944de824f4/lib/parser.js#L259)

docs/user-guide/rules/list.md
```diff
@@ -91,6 +91,10 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 - [`no-invalid-double-slash-comments`](../../../lib/rules/no-invalid-double-slash-comments/README.md): Disallow double-slash comments (`//...`) which are not supported by CSS.
 - [`no-invalid-position-at-import-rule`](../../../lib/rules/no-invalid-position-at-import-rule/README.md): Disallow invalid position `@import` rules within a stylesheet.
 
+### Annotation
```
jeddy3, line 94: Let's move this to the top of the section (below Avoid errors) as the sections are ordered by size and then alphabetically.

lib/rules/annotation-no-unknown/README.md
```diff
@@ -0,0 +1,74 @@
...
+  color: green !IMPORTANT;
+}
+```
+
+<!-- prettier-ignore -->
+```css
+a {
+  color: green !ImPoRtAnT;
+}
+```
```
jeddy3, line 48:
```suggestion
```

Let's keep the examples as simple as possible; case sensitivity isn't important to communicate.

jeddy3 (changes requested):
@mattxwang Thanks for making a start on this. Looking good so far.

I've requested some changes to make the docs, tests and rule consistent with existing rules.

mattxwang:
Thank you @ybiquitous and @jeddy3 for the in-depth feedback, I appreciate both of you and your patience as I still getting familiar with the codebase. I definitely need to do a better job of following the conventions doc!

I think I've resolved all of the outstanding changes, and I've removed the need to use `!` (not sure if the way that I've done it is hacky). 

One note: 

> It seems good that !important be case-sensitive.

With my understanding of the PostCSS parser, the top-level declaration doesn't differentiate between different capitalizations of `!important`. There is a different value in the `raws` field that I could use to check if the annotation is a lowercase `!important`. With that in mind, should I add a check for that in the rule (making the rule case-sensitive)? Or, should I leave it as is (case-insensitive)?

```console
$ node -e 'require("postcss").parse("a { color: red !IMPORTANT }").walkDecls(decl => console.log(decl))'
<ref *1> Declaration {
  raws: { before: ' ', between: ': ', important: ' !IMPORTANT' },
  type: 'decl',
  parent: Rule {
    raws: { before: '', between: ' ', semicolon: false, after: ' ' },
    type: 'rule',
    nodes: [ [Circular *1] ],
    parent: Root {
      raws: [Object],
      type: 'root',
      nodes: [Array],
      source: [Object],
      lastEach: 1,
      indexes: [Object],
      [Symbol(isClean)]: false,
      [Symbol(my)]: true
    },
    source: { start: [Object], input: [Input], end: [Object] },
    selector: 'a',
    lastEach: 1,
    indexes: { '1': 0 },
    [Symbol(isClean)]: false,
    [Symbol(my)]: true
  },
  source: {
    start: { offset: 4, line: 1, column: 5 },
    input: Input {
      css: 'a { color: red !IMPORTANT }',
      hasBOM: false,
      id: '<input css HyYU1z>',
      [Symbol(fromOffsetCache)]: [Array]
    },
    end: { offset: 24, line: 1, column: 25 }
  },
  prop: 'color',
  important: true,
  value: 'red',
  [Symbol(isClean)]: false,
  [Symbol(my)]: true
}

```

ybiquitous:
> With my understanding of the PostCSS parser, the top-level declaration doesn't differentiate between different capitalizations of `!important`.

Sorry, let's follow the current PostCSS parser behavior!

jeddy3:
> - Added: `annotation-no-unknown` rule ([#6155](https://github.com/stylelint/stylelint/pull/6155)).

PR #5555, opened 2021-09-21
Use dashed indents in examples of custom at-rule constructs

sylwia-lask:
> Which issue, if any, is this issue related to?

Ref #5523 

> Is there anything in the PR that needs further explanation?

No

Issue #5523, opened 2021-09-12
Use dashed idents in our examples of custom constructs


jeddy3:
> Clearly describe the documentation issue

We should use dashed idents in our examples of custom constructs to [align with the specs](https://www.w3.org/TR/css-values-4/#dashed-idents) (see example 7), e.g. `@--custom` rather than `@custom`.

This is in line with Stylelint's emphasis on standard CSS in our rules and docs.

Ref: https://github.com/stylelint/stylelint/pull/5520#issuecomment-917625540

- [x] At-rules
- [x] Functions
- [x] Units
- [x] Selectors

sylwia-lask:
Hi @jeddy3, I think I can do that next week, but please tell me should I branch off v14 or master?

hudochenkov:
Use v14, please.

ybiquitous:
I believe we've rewritten all the examples. If there is what we missed, please let us know or open a pull request.

PR #6162, opened 2022-06-20
Use dashed indents in examples of custom unit constructs

mattxwang:

Part of umbrella issue #5523.

Tried to follow #5555 but used the foo, bar, baz naming conventions in the docs. Also updated tests to mostly match. Let me know if I should do something different.

In looking for custom unit constructs (mostly skimming the rule list), I only found them in `unit-no-unknown`. Let me know if I should add any others.

### DIFF
--- a/lib/rules/unit-no-unknown/README.md
+++ b/lib/rules/unit-no-unknown/README.md
@@ -68,7 +68,7 @@ a {
 Given:
 
 ```json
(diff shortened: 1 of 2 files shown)


jeddy3 (approved):

jeddy3:
Thank you!

PR #6163, opened 2022-06-20
Fix end positions for `at-rule-no-vendor-prefix`, `at-rule-property-required-list`

mattxwang:

Part of umbrella issue #5694.

For `at-rule-property-required-list`, I chose to highlight just the at-rule (to me, it seems like the problem is "at" the at-rule). Open to changing it if there is a different suggestion!


ybiquitous (approved):
@mattxwang Thank you. LGTM 👍🏼

jeddy3:
> - Fixed: `at-rule-*` end positions ([#6163](https://github.com/stylelint/stylelint/pull/6163)).

PR #6164, opened 2022-06-20
Add keyframe-selector-notation

### LINKED ISSUE #4065: Add keyframe-selector-notation

In animations the as well the `from/to` syntax as the `0%/100%` syntax can be used.

For example:
```css
@keyframes animation-1 {
  from {
    color: red;
  }
  to {
    color: blue;
  }
}

@keyframes animation-2 {
  0% {
    color: green;
  }
  100% {
    color: yellow;
  }
}
```

A new rule that would determine which approach you should use. Not sure how this should be implemented, because sometimes you'll have an animation which looks like this:

```css
@keyframes animation-2 {
  20% {
    color: green;
  }
  100% {
    color: yellow;
  }
}
```
Which will lead to a mix of `20%` and `to` if the `from/to` syntax is preferred.

mattxwang:

Closes #4065.

Not sure if I'm implementing the detect & autofix features properly; I've made some comments below.

### DIFF
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -134,6 +134,10 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 
 - [`import-notation`](../../../lib/rules/import-notation/README.md): Specify string or URL notation for `@import` rules (Autofixable).
 
+### Keyframe Selectors
+
+- [`keyframe-selector-notation`](../../../lib/rules/keyframe-selector-notation/README.md): Specify keyword or percentage notation for keyframe selectors.
+
 ### Keyframes
 
 - [`keyframes-name-pattern`](../../../lib/rules/keyframes-name-pattern/README.md): Specify a pattern for keyframe names.
--- a/lib/rules/keyframe-selector-notation/README.md
+++ b/lib/rules/keyframe-selector-notation/README.md
@@ -1,0 +1,54 @@
+# keyframe-selector-notation
+
+Specify keyword or percentage notation for keyframe selectors. This rule only enforces the convention for `from`, `to`, `0%` and `100%`.
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo { from {} to {} }
+/**              ↑       ↑
+ *               These notations */
+```
+
+The keyword `from` is equivalent to the value `0%`. The keyword `to` is equivalent to the value `100%`.
+
+The [`fix` option](../../../docs/user-guide/usage/options.md#fix) can automatically fix all of the problems reported by this rule.
+
+## Options
+
+`string`: `"keyword"|"percentage"`
+
+### `"keyword"`
+
+Keyframe selectors _must always_ use the keyword notation.
+
+The following pattern is considered a problem:
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo { 0 {} 100% {} }
+```
+
+The following pattern is _not_ considered a problem:
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo { from {} to {} }
+```
+
+### `"percentage"`
+
+Keyframe selectors _must always_ use the percentage notation.
+
+The following pattern is considered a problem:
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo { from {} to {} }
+```
+
+The following pattern is _not_ considered a problem:
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo { 0 {} 100% {} }
+```
--- a/lib/rules/keyframe-selector-notation/index.js
+++ b/lib/rules/keyframe-selector-notation/index.js
@@ -1,0 +1,110 @@
+'use strict';
+
+const report = require('../../utils/report');
+const ruleMessages = require('../../utils/ruleMessages');
+const validateOptions = require('../../utils/validateOptions');
+
+const ruleName = 'keyframe-selector-notation';
+
+const messages = ruleMessages(ruleName, {
+	expected: (selector, fixedSelector) => `Expected "${selector}" to be "${fixedSelector}"`,
+});
+
+const meta = {
+	url: 'https://stylelint.io/user-guide/rules/list/keyframe-selector-notation',
+};
+
+const INTERCHANGEABLE_KEYFRAME_SELECTORS = new Set(['0%', '100%', 'from', 'to']);
+
+/** @type {import('stylelint').Rule} */
+const rule = (primary, _, context) => {
+	return (root, result) => {
+		const validOptions = validateOptions(result, ruleName, {
+			actual: primary,
+			possible: ['keyword', 'percentage'],
+		});
+
+		if (!validOptions) return;
+
+		const optionFuncs = Object.freeze({
+			keyword: {
+				expFunc: isKeyframeSelectorKeyword,
+				fixFunc: asKeyframeSelectorKeyword,
+			},
+			percentage: {
+				expFunc: isKeyframeSelectorPercentage,
+				fixFunc: asKeyframeSelectorPercentage,
+			},
+		});
+
+		root.walkAtRules(/^(-(moz|webkit)-)?keyframes$/i, (atRuleKeyframes) => {
+			atRuleKeyframes.walkRules((keyframeRule) => {
+				const selector = keyframeRule.selector;
+
+				const normalizedSelector = selector.toLowerCase();
+
+				if (!INTERCHANGEABLE_KEYFRAME_SELECTORS.has(normalizedSelector)) {
+					return;
+				}
+
+				/** @type {'keyword' | 'percentage'} */
+				const expectation = primary;
+
+				if (optionFuncs[expectation].expFunc(selector)) return;
+
+				const fixedSelector = optionFuncs[expectation].fixFunc(selector);
+
+				if (context.fix) {
+					keyframeRule.selector = fixedSelector;
+
+					return;
+				}
+
+				report({
+					message: messages.expected(selector, fixedSelector),
+					node: keyframeRule,
+					result,
+					ruleName,
+					word: selector,
+				});
+			});
+		});
+	};
+};
+
+/**
+ * @param {string} selector
+ * @returns {string}
+ */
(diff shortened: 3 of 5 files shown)


lib/rules/keyframe-selector-notation/index.js
```diff
@@ -0,0 +1,110 @@
...
+
+const messages = ruleMessages(ruleName, {
+	expected: (selector, fixedSelector) => `Expected "${selector}" to be "${fixedSelector}"`,
+});
+
+const meta = {
+	url: 'https://stylelint.io/user-guide/rules/list/keyframe-selector-notation',
+};
+
+const INTERCHANGEABLE_KEYFRAME_SELECTORS = new Set(['0%', '100%', 'from', 'to']);
```
mattxwang, line 17: This was the simplest approach, but I'm not sure if this is maintainable. Should this instead go into `keywordSets`, be the union of two sets, etc.?

lib/rules/keyframe-selector-notation/index.js
```diff
@@ -0,0 +1,110 @@
...
+	return selector === '0%' ? 'from' : 'to';
+}
+
+/**
+ * @param {string} selector
+ * @returns {string}
+ */
+function asKeyframeSelectorPercentage(selector) {
+	return selector === 'from' ? '0%' : '100%';
+}
```
mattxwang, line 89: Is this the right approach for "casting" between keyword and percentage? Should I be enforcing this with some sort of enum, throw on an error, etc.?

lib/rules/keyframe-selector-notation/index.js
ybiquitous: Defining the constant used only in this module seems to make sense, as the first implementation. 👍🏼

lib/rules/keyframe-selector-notation/index.js
ybiquitous: As the first implementation, this seems enough to me. 👍🏼

lib/rules/keyframe-selector-notation/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+### `"keyword"`
+
+Keyframe selectors _must always_ use the keyword notation.
+
+The following pattern is considered a problem:
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo { 0 {} 100% {} }
```
ybiquitous, line 28: [suggest] Is `0%` more consistent if seeing other descriptions in this doc?

```suggestion
@keyframes foo { 0% {} 100% {} }
```

lib/rules/keyframe-selector-notation/index.js
```diff
@@ -0,0 +1,110 @@
...
+		const optionFuncs = Object.freeze({
+			keyword: {
+				expFunc: isKeyframeSelectorKeyword,
+				fixFunc: asKeyframeSelectorKeyword,
+			},
+			percentage: {
+				expFunc: isKeyframeSelectorPercentage,
+				fixFunc: asKeyframeSelectorPercentage,
+			},
+		});
```
ybiquitous, line 38: [suggest] IMO, we can simplify the code like this:

```js
const PERCENTAGE_SELECTORS = new Set(['0%', '100%']);
const KEYWORD_SELECTORS = new Set(['from', 'to']);

// ...

/**
 * @type {Record<primary, {
 *   expFunc: (selector: string) => boolean,
 *   fixFunc: (selector: string) => string,
 * }>}
 */
const optionFuncs = Object.freeze({
	keyword: {
		expFunc: (selector) => KEYWORD_SELECTORS.has(selector),
		fixFunc: (selector) => {
			return selector === '0%' ? 'from' : 'to';
		},
	},
	percentage: {
		expFunc: (selector) => PERCENTAGE_SELECTORS.has(selector),
		fixFunc: (selector) => {
			return selector === 'from' ? '0%' : '100%';
		},
	},
});

// ...

if (!KEYWORD_SELECTORS.has(normalizedSelector) && !PERCENTAGE_SELECTORS.has(normalizedSelector)) {
	return;
}
```

But this is just my opinion. The current code is enough, and you can keep it as-is. 👍🏼

ybiquitous (commented):
@mattxwang Thank you for your excellent work!

I think this pull request looks almost good though I've left some trivial comments. 👍🏼

mattxwang (commented):
Thanks for the quick comments! Let me see if I can rework the code given your suggestion.

lib/rules/keyframe-selector-notation/index.js
mattxwang: Definitely agree that this is cleaner in hindsight! I'll look to bring this in.

mattxwang:
Hm, the failed actions aren't me - it's a 503 on NPM's registry with `npm install --global npm@latest`.

lib/rules/keyframe-selector-notation/README.md
```diff
@@ -0,0 +1,54 @@
...
+<!-- prettier-ignore -->
+```css
+@keyframes foo { from {} to {} }
+```
+
+The following pattern is _not_ considered a problem:
+
+<!-- prettier-ignore -->
+```css
+@keyframes foo { 0 {} 100% {} }
```
ybiquitous, line 53: Ah, I missed here 😅 

```suggestion
@keyframes foo { 0% {} 100% {} }
```

mattxwang: Thanks - evidently, so did I!

ybiquitous:
> Hm, the failed actions aren't me - it's a 503 on NPM's registry with `npm install --global npm@latest`.

It seems to be due to the Cloudflare outage:
https://www.cloudflarestatus.com/

docs/user-guide/rules/list.md
```diff
@@ -134,6 +134,10 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 
 - [`import-notation`](../../../lib/rules/import-notation/README.md): Specify string or URL notation for `@import` rules (Autofixable).
 
+### Keyframe selectors
```
jeddy3, line 137:
```suggestion
### Keyframe selector
```

I'll fix the title above and below to match in another pull request:
- Imports -> Import
- Keyframes - Keyframes name

lib/rules/keyframe-selector-notation/README.md
```diff
@@ -0,0 +1,54 @@
+# keyframe-selector-notation
+
+Specify keyword or percentage notation for keyframe selectors. This rule only enforces the convention for `from`, `to`, `0%` and `100%`.
```
jeddy3, line 3:
```suggestion
Specify keyword or percentage notation for keyframe selectors.
```

Let's be consistent with other descriptions (and in sync with the list of rules). 

FYI, "This rule..." stuff usually sits in the [Expanded description](https://stylelint.io/user-guide/rules/about#readmes). No need this move this particular one down though as it's redundant.

jeddy3 (changes requested):
Looking great! Thank you.

Let's add tests for a couple of edge-cases [that are valid syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes):
- duplicate keyframe selectors, e.g. `@keyframes foo { from {} from {} to {} }`
- mixed notations, e.g. `@keyframes foo { from {} 50% {} to {} }`
- keyframe selectors lists, e.g. `@keyframes foo { from, to {} }`

Let's add accept and reject tests for each of these.

mattxwang:
Thanks @jeddy3 for the suggestion! Upon adding the test cases, I realized I didn't properly handle the selector list use-case. I've written a bit of a hacky solution that requires the selector parser, walks the tags, and then regenerates the selector; while it works, it doesn't feel like the most elegant solution. Do you have any suggestions (or places to look) for a better way to do this?

jeddy3:
@mattxwang Thanks for making the changes. I think it's a valid use of the selector parser. @ybiquitous may have some thoughts on the code structure as he's done lots of refactoring lately to make our code consistent across rules.

lib/rules/keyframe-selector-notation/index.js
```diff
@@ -0,0 +1,110 @@
...
+								(fixedSelector) => (selectorTag.value = fixedSelector),
+							);
+						});
+					}).processSync(keyframeRule.selector);
+				} else {
+					checkSelector(
+						keyframeRule.selector,
+						(fixedSelector) => (keyframeRule.selector = fixedSelector),
+					);
+				}
```
ybiquitous, line 68:
```suggestion
				transformSelector(result, keyframeRule, (selectors) => {
					selectors.walkTags((selectorTag) => {
						checkSelector(
							selectorTag.value,
							(fixedSelector) => (selectorTag.value = fixedSelector),
						);
					});
				});
```

ybiquitous (commented):
How about our `transformSelector()` utility?

https://github.com/stylelint/stylelint/blob/a6318d45623bf0e4abf80516abc79e8304ea8c01/lib/utils/transformSelector.js#L11

lib/rules/keyframe-selector-notation/index.js
mattxwang: Makes sense!

Just to check - I originally added the `if`/`else` logic so that we wouldn't use the selector parser if there was only one selector. I was unsure if the selector parser adds any performance issues on routine checks. Does that matter in the grand scheme of things?

ybiquitous: I try benchmark, for example:

```console
$ npm run benchmark-rule -- keyframe-selector-notation keyword
```

But I cannot find any performance issues. Can you please try it?

And, I think it useful that `transformSelector()` catches a parser error:

https://github.com/stylelint/stylelint/blob/a6318d45623bf0e4abf80516abc79e8304ea8c01/lib/utils/transformSelector.js#L15

jeddy3 (approved):
Thanks for making all the changes.

LGTM!

ybiquitous (approved):
Thank you! LGTM 👍🏼

jeddy3:
> - Added: `keyframe-selector-notation` rule ([#6164](https://github.com/stylelint/stylelint/pull/6164)).

PR #6167, opened 2022-06-22
Refactor to use `Set` for `shorthandData` and `mathFunctions` types

ybiquitous:

Ref: https://github.com/stylelint/stylelint/issues/6151#issuecomment-1161953568

This pull request is prepared for #6151, unifying reference data structures.

### DIFF
--- a/lib/reference/shorthandData.js
+++ b/lib/reference/shorthandData.js
@@ -1,10 +1,9 @@
 'use strict';
 
-/** @type {Record<string, string[]>} */
 module.exports = {
-	margin: ['margin-top', 'margin-bottom', 'margin-left', 'margin-right'],
-	padding: ['padding-top', 'padding-bottom', 'padding-left', 'padding-right'],
-	background: [
+	margin: new Set(['margin-top', 'margin-bottom', 'margin-left', 'margin-right']),
+	padding: new Set(['padding-top', 'padding-bottom', 'padding-left', 'padding-right']),
+	background: new Set([
 		'background-image',
 		'background-size',
 		'background-position',
@@ -13,8 +12,8 @@ module.exports = {
 		'background-clip',
 		'background-attachment',
 		'background-color',
-	],
-	font: [
+	]),
+	font: new Set([
 		'font-style',
 		'font-variant',
 		'font-weight',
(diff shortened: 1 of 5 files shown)


lib/rules/declaration-block-no-shorthand-property-overrides/index.js
```diff
@@ -32,10 +32,12 @@ const rule = (primary) => {
...
 			eachDecl((decl) => {
 				const prop = decl.prop;
-				const unprefixedProp = vendor.unprefixed(prop);
+				const unprefixedProp = vendor.unprefixed(prop).toLowerCase();
 				const prefix = vendor.prefix(prop).toLowerCase();
 
-				const overrideables = shorthandData[unprefixedProp.toLowerCase()];
+				const overrideables = /** @type {Record<string, Set<string>>} */ (shorthandData)[
+					unprefixedProp
+				];
```
ybiquitous, line 40: [note] Without this type-cast using `@type`, TypeScript compiler complains as below:

```
lib/rules/declaration-block-no-shorthand-property-overrides/index.js:38:27 - error TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ margin: Set<string>; padding: Set<string>; background: Set<string>; font: Set<string>; border: Set<string>; 'border-top': Set<string>; 'border-bottom': Set<string>; ... 27 more ...; mask: Set<...>; }'.
  No index signature with a parameter of type 'string' was found on type '{ margin: Set<string>; padding: Set<string>; background: Set<string>; font: Set<string>; border: Set<string>; 'border-top': Set<string>; 'border-bottom': Set<string>; ... 27 more ...; mask: Set<...>; }'.

38     const overrideables = (shorthandData)[
                             ~~~~~~~~~~~~~~~~
39      unprefixedProp
   ~~~~~~~~~~~~~~~~~~~
40     ];
   ~~~~~
```

lib/reference/shorthandData.js
```diff
@@ -1,10 +1,9 @@
 'use strict';
 
-/** @type {Record<string, string[]>} */
```
ybiquitous, line 3: [note] I think this `@type` is needless because the auto-completion via property names works.

jeddy3 (approved):
LGTM, thank you!

PR #6168, opened 2022-06-23
Add shorthand properties reference to public API

### LINKED ISSUE #6151: Publicly expose reference data

ybiquitous:

- Closes #6151
- #6152

This pull request is following the suggestion below by @jeddy3 in <https://github.com/stylelint/stylelint/issues/6151#issuecomment-1160488202>:

> - align and expose data as it's requested (starting with the shorthand data)

The type definition is generated by the following script:

```js
const shorthandData = require("./lib/reference/shorthandData")
console.log('export type ShorthandData = {')
for (const key of Object.keys(shorthandData).sort()) {
	console.log(`\t'${key}': Set<string>;`)
}
console.log('};')
```

### DIFF
--- a/lib/reference/properties.js
+++ b/lib/reference/properties.js
@@ -1,7 +1,7 @@
 'use strict';
 
-/** @type {import('stylelint').ShorthandData} */
-const shorthandData = {
+/** @type {import('stylelint').LonghandSubPropertiesOfShorthandProperties} */
+const longhandSubPropertiesOfShorthandProperties = {
 	margin: new Set(['margin-top', 'margin-bottom', 'margin-left', 'margin-right']),
 	padding: new Set(['padding-top', 'padding-bottom', 'padding-left', 'padding-right']),
 	background: new Set([
@@ -147,4 +147,6 @@ const shorthandData = {
 	]),
 };
 
-module.exports = shorthandData;
+module.exports = {
+	longhandSubPropertiesOfShorthandProperties,
+};
--- a/lib/reference/properties.js
+++ b/lib/reference/properties.js
@@ -1,239 +1,351 @@
 'use strict';
 
 /** @type {import('stylelint').LonghandSubPropertiesOfShorthandProperties} */
-const longhandSubPropertiesOfShorthandProperties = {
-	/* eslint sort-keys: 'error' */
-	animation: new Set([
-		'animation-name',
-		'animation-duration',
-		'animation-timing-function',
-		'animation-delay',
-		'animation-iteration-count',
-		'animation-direction',
-		'animation-fill-mode',
-		'animation-play-state',
-	]),
-	background: new Set([
-		'background-image',
-		'background-size',
-		'background-position',
-		'background-repeat',
-		'background-origin',
-		'background-clip',
-		'background-attachment',
-		'background-color',
-	]),
-	border: new Set([
-		'border-top-width',
-		'border-bottom-width',
-		'border-left-width',
-		'border-right-width',
-		'border-top-style',
-		'border-bottom-style',
-		'border-left-style',
-		'border-right-style',
-		'border-top-color',
-		'border-bottom-color',
-		'border-left-color',
-		'border-right-color',
-	]),
-	'border-block-end': new Set([
-		'border-block-end-width',
-		'border-block-end-style',
-		'border-block-end-color',
-	]),
-	'border-block-start': new Set([
-		'border-block-start-width',
-		'border-block-start-style',
-		'border-block-start-color',
-	]),
-	'border-bottom': new Set([
-		// prettier-ignore
-		'border-bottom-width',
-		'border-bottom-style',
-		'border-bottom-color',
-	]),
-	'border-color': new Set([
-		'border-top-color',
-		'border-bottom-color',
-		'border-left-color',
-		'border-right-color',
-	]),
-	'border-image': new Set([
-		'border-image-source',
-		'border-image-slice',
-		'border-image-width',
-		'border-image-outset',
-		'border-image-repeat',
-	]),
-	'border-inline-end': new Set([
-		'border-inline-end-width',
-		'border-inline-end-style',
-		'border-inline-end-color',
-	]),
-	'border-inline-start': new Set([
-		'border-inline-start-width',
-		'border-inline-start-style',
-		'border-inline-start-color',
-	]),
-	'border-left': new Set([
-		// prettier-ignore
-		'border-left-width',
-		'border-left-style',
-		'border-left-color',
-	]),
-	'border-radius': new Set([
-		'border-top-right-radius',
-		'border-top-left-radius',
-		'border-bottom-right-radius',
-		'border-bottom-left-radius',
-	]),
-	'border-right': new Set([
-		// prettier-ignore
-		'border-right-width',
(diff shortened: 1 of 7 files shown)


ybiquitous:
[ask] The `shorthandData` name seems unclear to me. Is it acceptable, or is there a better name?

jeddy3:
@ybiquitous Thank you. It's looking good.

> [ask] The shorthandData name seems unclear to me. Is it acceptable, or is there a better name?

Let's go with `subPropertiesOfLonghandProperties` as per https://github.com/stylelint/stylelint/issues/6151#issuecomment-1160488202.

Future property sets could include [`acceptCustomIdentsProperties`](https://github.com/stylelint/stylelint/blob/main/lib/reference/propertySets.js)  and these:

https://github.com/stylelint/stylelint/blob/bd8c7fd88172af93d8ae80b38256b5074ebee233/lib/reference/keywordSets.js#L411-L423

(Which are in the `keywordSets.js` at the moment...)

What file structure shall we use? Choices are:

- 1 file per set, e.g. `reference/subPropertiesOfLonghandProperties.js`
- 1 file per group, e.g. `reference/properties.js` (which would contain all the above)
- a single file for all the sets, e.g. `reference/reference.js`

I feel like either of the last two will be easiest to work with and maintain.

ybiquitous:
@jeddy3 Thanks for your helpful feedback! The idea of "1 file per group" seems so nice. 👍🏼 

I'll push commits into this pull request, following the ideas of `subPropertiesOfLonghandProperties` and "1 file per group".

jeddy3:
> I'll push commits into this pull request, following the ideas of `subPropertiesOfLonghandProperties` and "1 file per group".

Oops, I got it back-to-front. From the spec:

> Some properties are shorthand properties, meaning that they allow authors to specify the values of several properties with a single property. A [**shorthand property**](https://www.w3.org/TR/css-cascade-4/#shorthand-property) sets all of its **longhand sub-properties**, exactly as if expanded in place.

Let's go with `longhandSubPropertiesOfShorthandProperties` to be accurate with the wording in the spec, even if it's a bit verbose.

ybiquitous:
I have one more suggestion: what if using `Map` instead of `Object`?

[This MDN doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#objects_vs._maps) describes the differences between `Map` and `Object` well. I think `Map` could be safer and more stable than `Object`, especially as a public API, even if `Map` doesn't support JSON serialization natively. E.g.

```js
// Using Object
const longhandSubPropertiesOfShorthandProperties = {
  margin: new Set(['margin-top', 'margin-bottom', 'margin-left', 'margin-right']),
  // ...
};
longhandSubPropertiesOfShorthandProperties.margin; //=> Set(...)

// Using Map
const longhandSubPropertiesOfShorthandProperties = new Map([
  ['margin', new Set(['margin-top', 'margin-bottom', 'margin-left', 'margin-right'])],
  // ...
]);
longhandSubPropertiesOfShorthandProperties.get('margin'); //=> Set(...)
```

What do you think?

lib/reference/properties.js
```diff
@@ -1,8 +1,18 @@
 'use strict';
 
-module.exports = {
-	margin: new Set(['margin-top', 'margin-bottom', 'margin-left', 'margin-right']),
-	padding: new Set(['padding-top', 'padding-bottom', 'padding-left', 'padding-right']),
+/** @type {import('stylelint').LonghandSubPropertiesOfShorthandProperties} */
+const longhandSubPropertiesOfShorthandProperties = {
+	/* eslint sort-keys: 'error' */
```
ybiquitous, line 5: [note] I've sorted properties alphabetically for findability (the diff increases, though).

lib/reference/properties.js
```diff
@@ -91,6 +47,18 @@ module.exports = {
 		'border-block-start-style',
 		'border-block-start-color',
 	]),
+	'border-bottom': new Set([
+		// prettier-ignore
```
ybiquitous, line 51: [note] To align always multi-line for readability.

jeddy3:
> I have one more suggestion: what if using Map instead of Object?

Sounds good to me. 

All the other changes are looking great, by the way!

types/stylelint/index.d.ts
```diff
@@ -383,6 +383,45 @@ declare module 'stylelint' {
...
+			| 'grid-template'
+			| 'list-style'
+			| 'margin'
+			| 'mask'
+			| 'outline'
+			| 'padding'
+			| 'text-decoration'
+			| 'text-emphasis'
+			| 'transition',
+			ReadonlySet<string>
```
ybiquitous, line 422: [note] `ReadonlyMap` and `ReadonlySet` are built-in types.

lib/rules/declaration-block-no-shorthand-property-overrides/index.js
```diff
@@ -27,34 +27,36 @@ const rule = (primary) => {
...
 				if (!overrideables) {
-					declarations[prop.toLowerCase()] = prop;
+					declarations.set(prop.toLowerCase(), prop);
 
 					return;
 				}
 
 				for (const longhandProp of overrideables) {
-					if (!Object.prototype.hasOwnProperty.call(declarations, prefix + longhandProp)) {
+					const declaration = declarations.get(prefix + longhandProp);
```
ybiquitous, line 49: [note] Refactor to use `Map` instead of `Object`; `hasOwnProperty` is no longer needed (safe access).

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -40,49 +40,47 @@ const rule = (primary, secondaryOptions) => {
...
-			},
-			{},
-		);
+			for (const longhand of longhandProps) {
+				const shorthands = longhandToShorthands.get(longhand) || [];
+
+				shorthands.push(shorthand);
+				longhandToShorthands.set(longhand, shorthands);
+			}
+		}
```
ybiquitous, line 57: [note] Refactor to use `Map` and `for...of` instead of `Object` and `reduce()` for readability.

ybiquitous:
@jeddy3 I've switched from `Object` to `Map` and `Set` (`ReadonlyMap` and `ReadonlySet` in TypeScript), and also refactored `declaration-block-*` rules using `Map`.

lib/rules/declaration-block-no-shorthand-property-overrides/index.js
jeddy3: Much nicer.

jeddy3 (approved):
Fantastic, thank you!

I think we have a good direction here for exposing our reference data.

jeddy3:
> - Added: longhand sub-properties of shorthand properties reference data to public API ([#6168](https://github.com/stylelint/stylelint/pull/6168)).

ybiquitous:
@jeddy3 Thank you for all of your reviews. 😊 👍🏼

PR #6169, opened 2022-06-23
Fix start and end positions for `unit-*`

mattxwang:

Part of umbrella issue #5694.

Following the discussion in #5694, I've adjusted the start and end positions to encapsulate just the unit. This is a "breaking" change for the `column` line.

The way I've done this is inefficient, since it makes two calls to the value parser. Another approach (that would be more efficient, but I'd like to get some feedback) would be to:

1. Adjust `getUnitFromValueNode` to return the unit string (current behaviour) *as well as* the length of the value string and/or the value string (new behaviour); rename it to something else
2. Create a new `getUnitFromValueNode` that consumes the above function and returns just the unit, for simple backwards compatability
3. Use the new function defined in # 1 to replace this entire code block:

```js
const parsedUnit = valueParser.unit(valueNode.value);

if (!parsedUnit) return;

const valueNumberLength = parsedUnit.number.length;
```

I may also be missing another approach. Any thoughts?

(and if we do go ahead with this approach, what should I call it?)

After this, I will then apply the same approach to `unit-case`, `unit-allowed-list`, and `unit-disallowed-list`.


jeddy3:
As `getUnitFromValueNode` is a wrapper around `valueParser.unit()` to make it compatible with non-standard syntax, then it makes sense to also returns the number. We can call it `getNumberAndUnit` and drop the `FromValueNode` as most (with the exception being [`getSchemeFromUrl.js`](https://github.com/stylelint/stylelint/blob/060310cfd6383175ee06fe4d15aee27a4e202671/lib/utils/getSchemeFromUrl.js)) of our other utils don't use the `FromX` convention.

Existing call to the function can be easily updated:

```diff
- const unit = getUnitFromValueNode(valueNode);
+ const { unit } = getNumberAndUnit(valueNode); 
```

mattxwang:
Thanks for the feedback @jeddy3! I've implemented `getNumberAndUnit` as you've suggested, and have also added some test cases for the new `Dimension.number` return functionality.

I've then used the new function to refactor the approach for `unit-no-unknown`, and bring the same approach to the rest of `unit-*`. There were a couple of small nuances to updating the function calls - not all the replacements were 100% drop-in - but they are close enough!

Updating the tests *was* repetitive, so it's possible that I've made some errors. Any ideas on how to do a good job verifying these new numbers? I don't think manual review is helpful for something of this size.

jeddy3:
Perhaps we should call the function `getDimension()` ([spec ref](https://www.w3.org/TR/css-values-3/#dimensions)) and always have it return one with the `number` and `unit` properties. I feel that'll be cleaner to use with destructuring, e.g.:

```js
// getDimension.js
return {
  unit: null,
  number: null
}
```

If working with just a unit:

```js
const { unit } = getDimension(node);
if (!unit) return;
```

If needing both the unit and the number:

```js
const { number, unit } = getDimension(node);
if (!number || !unit) return;
```

Or if checking an option then no early return is needed:

```js
const { unit } = getDimension(node);
if (optionsMatches(secondaryOptions, 'ignoreUnits', unit)) {}
```

mattxwang:
Thanks for the suggestions @jeddy3 - I think I've implemented them. Definitely makes the code easier to use/more readable!

jeddy3 (approved):
Excellent, thank you!

You put a lot of legwork into this, especially updating so many tests. It's appreciated!

lib/utils/getDimension.js
```diff
@@ -11,26 +11,38 @@ const valueParser = require('postcss-value-parser');
  *
  * @param {import('postcss-value-parser').Node} node
  *
- * @returns {string | null}
+ * @returns {{unit: null, number: null} | valueParser.Dimension}
```
ybiquitous, line 14: [suggest] What if also updating the function description?

https://github.com/stylelint/stylelint/blob/788f612b16d42919ddca6552ba486003d0c47dc2/lib/utils/getDimension.js#L8-L10

lib/utils/__tests__/getDimension.test.js
```diff
@@ -0,0 +1,172 @@
...
+const valueParser = require('postcss-value-parser');
+
+it('getDimension', () => {
+	expect(getDimension().number).toBeNull();
+	expect(getDimension().unit).toBeNull();
+	expect(getDimension({}).number).toBeNull();
+	expect(getDimension({}).unit).toBeNull();
+
+	// testing Dimension.unit
+	expect(getDimension(valueParser('1.1s').nodes[0]).unit).toBe('s');
```
ybiquitous, line 13: [trivial] [`.toEqual()`](https://jestjs.io/docs/expect#toequalvalue) can reduce repetations as below:

```suggestion
	expect(getDimension(valueParser('1.1s').nodes[0])).toEqual({ unit: 's', number: '1.1' });
```

But keeping the code as-is has no problems. 👍🏼

lib/rules/unit-allowed-list/index.js
```diff
@@ -87,7 +87,8 @@ const rule = (primary, secondaryOptions) => {
 				}
 
 				report({
-					index: getIndex(node) + valueNode.sourceIndex,
+					index: getIndex(node) + valueNode.sourceIndex + number.length,
+					endIndex: getIndex(node) + valueNode.sourceEndIndex,
```
ybiquitous, line 91: [suggest] How about reducing `getIndex()` calls for a bit improved performance?

```js
const index = getIndex(node);

report({
  index: index + valueNode.sourceIndex + number.length,
  endIndex: index + valueNode.sourceEndIndex,
  // ...
});
```

lib/rules/unit-case/index.js
```diff
@@ -59,7 +57,8 @@ const rule = (primary, _secondaryOptions, context) => {
 				}
 
 				problems.push({
-					index: getIndex(node) + valueNode.sourceIndex,
+					index: getIndex(node) + valueNode.sourceIndex + number.length,
+					endIndex: getIndex(node) + valueNode.sourceEndIndex,
```
ybiquitous, line 61: [suggest] The same applies here:

```js
const index = getIndex(node);

problems.push({
  index: index + valueNode.sourceIndex + number.length,
  endIndex: index + valueNode.sourceEndIndex,
  // ...
});
```

mattxwang:
Thanks for the suggestions @ybiquitous - been busy but will look to implement them tomorrow/the day after!

ybiquitous:
@mattxwang No worry. We can refactor `lib/utils/__tests__/getDimension.test.js` with another PR. 👍🏼

mattxwang:
Have just pushed the last of the requested changes!

ybiquitous (approved):
@mattxwang Thank you. LGTM 👍🏼

jeddy3:
> - Fixed: `unit-*` start and end positions ([#6169](https://github.com/stylelint/stylelint/pull/6169)).

PR #6174, opened 2022-06-25
Fix end positions for `property-*`

mattxwang:

Part of https://github.com/stylelint/stylelint/issues/5694.

Includes: 

- `property-allowed-list`
- `property-case`
- `property-disallowed-list`
- `property-no-vendor-prefix`

Do we think this approach with `word`  is a good idea? For some of my other end positions, it's been a problem with duplicate values - `word` (and the underlying PostCSS) always picks the first instance. It wasn't a problem for these test cases though!


jeddy3 (approved):
LGTM, thank you!

> Do we think this approach with word is a good idea?

It's a step in the right direction. We can improve the accuracy for duplicated words down the line in a follow-up series of pull requests.

ybiquitous (approved):
LGTM! 👍🏼

PR #6175, opened 2022-06-26
Fix exit code when nothing is passed to stylelint command

### LINKED ISSUE #5150: Fix exit code when nothing is passed to stylelint command

I am running stylelint on Jenkins. I pull the diffs of Base PR and Current PR and then pass the filenames to stylelint. 

Sometimes the diff will be nil or an empty string.

`git diff origin/${CHANGE_TARGET}... --name-only | egrep '\.(scss|css)$' | xargs`

I pass the above to stylelint like so

```
stylelint --aei $(git diff origin/${CHANGE_TARGET} --name-only | egrep '\.(scss|css)$' | xargs)
```

It fails...

If no glob/filename is passed, lint check fails. 

**It should not fail**. (ie eslint returns with an exit code of 0 if nothing is passed)

stylelint

```
$ stylelint
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

eslint

```
$ eslint
$ eslint
✨  Done in 0.50s.
```

mattxwang:

Closes #5150.

Seems *too simple* - am I missing something?

Also, is this a breaking change? I did change the behaviour that was captured in a test.

### DIFF
--- a/lib/__tests__/cli.test.js
+++ b/lib/__tests__/cli.test.js
@@ -174,7 +174,7 @@ describe('CLI', () => {
 	it('basic', async () => {
 		await cli([]);
 
-		expect(process.exit).toHaveBeenCalledWith(2);
+		expect(process.exit).toHaveBeenCalledWith(0);
 
 		expect(console.log).toHaveBeenCalledTimes(1);
 		expect(console.log).toHaveBeenCalledWith(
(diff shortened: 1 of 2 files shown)


ybiquitous (approved):
Thanks, LGTM 👍🏼 

I don't want to make this change *breaking* (that is, need a major update), even if this may be strictly changing the current behavior. Because maybe this would not affect most users.

jeddy3 (approved):
LGTM, thank you!

jeddy3:
> - Fixed: exit code when nothing is passed to stylelint command ([#6175](https://github.com/stylelint/stylelint/pull/6175)).

PR #6179, opened 2022-06-27
Refactor reference data into groups

jeddy3:

- Follows on from @ybiquitous great work in https://github.com/stylelint/stylelint/pull/6168
- Original issue https://github.com/stylelint/stylelint/issues/6151

This pull request doesn't add any new reference data to the public API as [we'll only want to do that if it's requested because we'll need to make sure the set names are accurate to the specs](https://github.com/stylelint/stylelint/issues/6151#issuecomment-1160488202). What it does is:

- breaks up the unwieldy `keywordSet.js` file into more accurately named groups
- uses the destructuring convention from https://github.com/stylelint/stylelint/pull/6168 to make it clear what data a rule or util is using
- standardizes a few of the set names, e.g. everything in `keywords` is a `*Keywords` set

Hopefully, this will make the reference data easier to work with internally and better prepare it for being added to the public API when requested.


ybiquitous (approved):
This is great work! I've left trivial suggestions, but it looks good enough to merge. 👍🏼

PR #6180, opened 2022-06-27
Refactor to remove `execall` dependency

ybiquitous:

None.

We can replace `execall` with `String.prototype.matchAll()` which has been supported since Node.js v12.

See also:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll
- https://github.com/sindresorhus/execall

FYI. We have supported Node.js v12+:

https://github.com/stylelint/stylelint/blob/0ffb27fa2dc01d46f9df96e5f2c9377b00b30a46/package.json#L192-L194

### DIFF
--- a/lib/rules/max-line-length/index.js
+++ b/lib/rules/max-line-length/index.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const execall = require('execall');
 const optionsMatches = require('../../utils/optionsMatches');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
@@ -9,10 +8,6 @@ const validateOptions = require('../../utils/validateOptions');
 const { isNumber, isRegExp, isString, assert } = require('../../utils/validateTypes');
 
 const ruleName = 'max-line-length';
-const EXCLUDED_PATTERNS = [
-	/url\(\s*(\S.*\S)\s*\)/gi, // allow tab, whitespace in url content
-	/@import\s+(['"].*['"])/gi,
-];
 
 const messages = ruleMessages(ruleName, {
 	expected: (max) =>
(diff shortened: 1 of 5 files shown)


lib/rules/max-line-length/index.js
```diff
@@ -51,6 +46,11 @@ const rule = (primary, secondaryOptions, context) => {
 			throw new Error('The root node must have a source');
 		}
 
+		const EXCLUDED_PATTERNS = [
+			/url\(\s*(\S.*\S)\s*\)/gi, // allow tab, whitespace in url content
+			/@import\s+(['"].*['"])/gi,
+		];
+
```
ybiquitous, line 53: [note] Bring the definition and the place of use closer together for readability.

lib/rules/max-line-length/index.js
```diff
@@ -59,14 +59,14 @@ const rule = (primary, secondaryOptions, context) => {
...
 		let skippedSubStringsIndex = 0;
 
-		for (const pattern of EXCLUDED_PATTERNS)
-			for (const match of execall(pattern, rootString)) {
-				const subMatch = match.subMatches[0] || '';
-				const startOfSubString = match.index + match.match.indexOf(subMatch);
+		for (const pattern of EXCLUDED_PATTERNS) {
+			for (const match of rootString.matchAll(pattern)) {
+				const subMatch = match[1] || '';
+				const startOfSubString = (match.index || 0) + (match[0] || '').indexOf(subMatch);
```
ybiquitous, line 65: [note] `match.index` is optional, so fallback to `0` is needed.

lib/rules/max-line-length/index.js
```diff
@@ -59,14 +59,14 @@ const rule = (primary, secondaryOptions, context) => {
 		let skippedSubStrings = [];
 		let skippedSubStringsIndex = 0;
 
-		for (const pattern of EXCLUDED_PATTERNS)
-			for (const match of execall(pattern, rootString)) {
-				const subMatch = match.subMatches[0] || '';
-				const startOfSubString = match.index + match.match.indexOf(subMatch);
+		for (const pattern of EXCLUDED_PATTERNS) {
+			for (const match of rootString.matchAll(pattern)) {
```
ybiquitous, line 63: [note] [The MDN doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll#examples) says:

> `matchAll` internally makes a clone of the `regexp`

so we don't need to worry about mutable `RegExp` instances (`lastIndex` never changes).

jeddy3 (approved):
Nice. LGTM, thank you!

PR #6181, opened 2022-06-28
Add `meta.fixable` property to each rule

### LINKED ISSUE #6099: Add `meta.autofixable` property to each rule

Similar discussion: #5842

If each rule would have metadata about *autofixable* or not, I think it would be useful for the following cases:

- the built-in `verbose` formatter could suggest autofixable rules if the rules errors are found
- custom formatters could make the Stylelint report more rich

It just adds the `autofixable` boolean property to the `RuleMeta` type as below:

https://github.com/stylelint/stylelint/blob/23dfa32e1c19930c2f3eec0cb27b276d3deb6ac8/types/stylelint/index.d.ts#L163-L166

```diff
export type RuleMeta = {
	url: string;
	deprecated?: boolean;
+	autofixable?: boolean;
};
```

Perhaps, the name `autofixable` may be debatable. Please give me feedback, if any.

ybiquitous:

Closes #6099

This commit adds a new property `meta.fixable` to each rule.
Also, this fixes inaccurate docs about autofix.

---

I used the following script for [`jscodeshift`](https://github.com/facebook/jscodeshift) to perform the task easily.
(though I needed to fix a few files manually)

```js
/**
 * @type {import('jscodeshift').Transform}
 */
module.exports = function transform(fileInfo, api) {
	const j = api.jscodeshift;
	const root = j(fileInfo.source);

	root.findVariableDeclarators('meta').forEach((decl) => {
		if (root.find(j.Identifier, { name: 'fix' }).length > 0) {
			decl.node.init.properties.push(j.property('init', j.identifier('fixable'), j.literal(true)));
		}
	});

	return root.toSource();
};
```

```console
$ npm i -D jscodeshift @types/jscodeshift
...

$ npx jscodeshift lib/rules/*/index.js
...
```

### DIFF
--- a/lib/rules/__tests__/index.test.js
+++ b/lib/rules/__tests__/index.test.js
@@ -1,12 +1,30 @@
 'use strict';
 
+const fs = require('node:fs');
+const path = require('node:path');
+
 const rules = require('..');
 
-describe('rules', () => {
-	for (const [ruleName, rule] of Object.entries(rules)) {
-		test(`the "${ruleName}" rule has metadata`, () => {
-			expect(rule.meta).toBeTruthy();
-			expect(rule.meta.url).toMatch(/^https:\/\/stylelint\.io\/.+/);
-		});
-	}
+describe('all rules', () => {
+	test.each(Object.entries(rules))('"%s" should have metadata', (_name, rule) => {
+		expect(rule.meta).toBeTruthy();
+		expect(rule.meta.url).toMatch(/^https:\/\/stylelint\.io\/.+/);
+		expect([true, undefined]).toContain(rule.meta.fixable);
+	});
+});
+
+describe('fixable rules', () => {
+	const fixableRules = Object.entries(rules).filter(([, rule]) => rule.meta.fixable);
+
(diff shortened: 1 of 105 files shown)


lib/rules/__tests__/index.test.js
```diff
@@ -1,12 +1,30 @@
...
 
-describe('rules', () => {
-	for (const [ruleName, rule] of Object.entries(rules)) {
-		test(`the "${ruleName}" rule has metadata`, () => {
-			expect(rule.meta).toBeTruthy();
-			expect(rule.meta.url).toMatch(/^https:\/\/stylelint\.io\/.+/);
-		});
-	}
+describe('all rules', () => {
+	test.each(Object.entries(rules))('"%s" should have metadata', (_name, rule) => {
```
ybiquitous, line 9: See the [Jest `test.each()` doc](https://jestjs.io/docs/api#testeachtablename-fn-timeout).

ybiquitous:
@jeddy3 Thank you. Changelog entry added:

> - Added: `meta.fixable` property to each rule ([#6181](https://github.com/stylelint/stylelint/pull/6181)).

PR #6182, opened 2022-06-28
Refactor to remove `get-stdin` dependency

ybiquitous:

None.

The `get-stdin` package is a tiny library, so we can replace it with its utility and reduce an extra dependency.
Also, since the package is used only by `cli.js`, replacing it is easy.

See https://github.com/sindresorhus/get-stdin

### DIFF
--- a/lib/utils/getStdin.js
+++ b/lib/utils/getStdin.js
@@ -1,3 +1,5 @@
+'use strict';
+
 /**
  * @returns {Promise<string>}
  */
--- a/lib/utils/getStdin.js
+++ b/lib/utils/getStdin.js
@@ -1,0 +1,12 @@
+/**
+ * @returns {Promise<string>}
+ */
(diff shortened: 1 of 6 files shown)


lib/utils/getStdin.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * @returns {Promise<string>}
+ */
+module.exports = async function getStdin() {
+	const chunks = [];
+
+	for await (const chunk of process.stdin) {
```
ybiquitous, line 7: [note] [`for await...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of) has been available since Node.js v10.

jeddy3 (approved):
LGTM, thank you.

(I don't know of a better solution for the mock.)

ybiquitous:
Thanks for the quick review. 😄

PR #6183, opened 2022-06-28
Show fixable metadata with `verbose` and `github` formatters

ybiquitous:

Ref: #6099

In this pull request, I try showing *fixable* metadata with the `verbose` and `github` formatters.
But I’m not sure whether this addition is really useful. I would be happy if any feedback.

Demo on localhost:

```console
$ (cd scripts ; ./visual.sh)
...

4 errors found
 color-no-invalid-hex: 1
 import-notation: 1 (fixable)
 selector-max-class: 1
 unit-disallowed-list: 1

1 warning found
 custom-property-pattern: 1
```

image

### DIFF
--- a/lib/formatters/__tests__/githubFormatter.test.js
+++ b/lib/formatters/__tests__/githubFormatter.test.js
@@ -34,11 +34,11 @@ test('githubFormatter', () => {
 	const returnValue = {
 		ruleMetadata: {
 			foo: { url: 'https://stylelint.io/rules/foo' },
-			bar: {},
+			bar: { fixable: true },
 		},
 	};
 
 	expect(githubFormatter(results, returnValue))
 		.toBe(`::error file=path/to/file.css,line=1,col=2,endLine=1,endColumn=5,title=Stylelint problem::Unexpected "foo" (foo) - https://stylelint.io/rules/foo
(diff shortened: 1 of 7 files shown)


jeddy3:
I like the addition, especially to the verbose formatter as I feel that those users want as much information as possible. The thing is, not all problems can be fixed by a rule; sometimes we fix only for certain options or we ignore some edge cases. Do we need to communicate that here? For example, by using "maybe fixable" instead of "fixable":

```
import-notation: 1 (maybe fixable)
```

People may be confused if they see "fixable", run "--fix" and then still see the problem listed.

ybiquitous:
@jeddy3 Thanks for the feedback. Your concern and suggestion make sense to me. 👍🏼 
I've pushed the commit 973935c08 to add the `maybe` word to messages.

ybiquitous:
In addition, what if showing the following message (only when fixable problems are found), for example?

> You may fix some problems with the "--fix" option.

image

Is this too verbose?

ybiquitous:
Changelog entry added:

> - Added: fixable status to `verbose` and `github` formatters ([#6183](https://github.com/stylelint/stylelint/pull/6183)).

Mouvedia:
> sometimes we fix only for certain options or we ignore some edge cases

@jeddy3 Should we enforce the addition of one skipped test per edge case not covered?

Issue #6187, opened 2022-06-29
declaration-property-value-*-list multiple matches are ignored


carlosjeurissen:
### What steps are needed to reproduce the bug?

Have a property match multiple regex in `declaration-property-value-*-list`.

### What Stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "declaration-property-value-disallowed-list": {
      "/.*/": ["rebeccapurple"],
      "background-color": ["none"],
    }
  }
}
```

### How did you run Stylelint?

CLI

### Which version of Stylelint are you using?

`14.9.0`

### What did you expect to happen?

Either "none" is disallowed or both "rebeccapurple" and "none" are disallowed.

### What actually happened?

Only "rebeccapurple" is disallowed.

### Proposal to fix the bug

Most preferred fix is to disallow both values and collapse all matches for `disallowed-list`.

If that is not doable. It would make sense if, instead of the order, non-regex match would have preference over regex match.

No clear preference for `allowed-list`. Up for discussion.

jeddy3:
@carlosjeurissen Thanks for the report and for using the template.

> Most preferred fix is to disallow both values 

SGTM.

I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

There are [steps on how to fix a bug in a rule](https://stylelint.io/developer-guide/rules#fix-a-bug-in-a-rule) in the Developer guide.

ybiquitous:
I've created PR #6188 to fix false negatives in `declaratioin-property-value-disallowed-list`.

---

On the other hand, are there false *positives* (not negatives) in `declaratioin-property-value-allowed-list`? We can see false positives in the [demo](https://stylelint.io/demo#N4Igxg9gJgpiBcICGACYAdAdilkA2EATvCoTAEYxhhIAOArobXjANxY7lJgDWA5oQj1MUALT4iJTBExssAXxAAaEAGcAnpgAuSAB4Jwq1cvAyAZgEs+BjNhToQhei1UOStnDgewweJISQtCxlRWkFaGEItdVEANyQ8ehhRBIIAdxgxPAtVLTc0Dk8vEAB6ADoAKhL8gG0HMkpqOkZmGAcAXSVCoocuXgEhEXEIAkJah2lZDu6UeUK5zEV5IA).

PR #6188, opened 2022-06-30
Fix false negatives for multiple match in `declaration-property-value-disallowed-list`

ybiquitous:

Partially fixing #6187

This change uses `.filter()` instead of `.find()` for `Object.keys(primary)`.

### DIFF
--- a/lib/rules/declaration-property-value-disallowed-list/index.js
+++ b/lib/rules/declaration-property-value-disallowed-list/index.js
@@ -32,20 +32,19 @@ const rule = (primary) => {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


lib/rules/declaration-property-value-disallowed-list/index.js
```diff
@@ -32,20 +32,19 @@ const rule = (primary) => {
 			return;
 		}
 
+		const propKeys = Object.keys(primary);
```
ybiquitous, line 35: [note] Calling `Object.keys(primary)` multiple times in `walkDecls()` loop is wasteful.

PR #6190, opened 2022-06-30
Fix false positives for multiple match in `declaration-property-value-allowed-list`

### LINKED ISSUE #6187: declaration-property-value-*-list multiple matches are ignored

ybiquitous:

Closes #6187

This fix is similar to #6188.

See also https://github.com/stylelint/stylelint/issues/6187#issuecomment-1170618875

### DIFF
--- a/lib/rules/declaration-property-value-allowed-list/__tests__/index.js
+++ b/lib/rules/declaration-property-value-allowed-list/__tests__/index.js
@@ -85,3 +85,17 @@ testRule({
 		},
 	],
 });
+
+testRule({
+	ruleName,
+	config: {
+		'/.*/': ['rebeccapurple'],
+		'background-color': ['none'],
+	},
+
+	accept: [
+		{
+			code: 'a { color: rebeccapurple; background-color: none; }',
+		},
+	],
+});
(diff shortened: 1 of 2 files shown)


jeddy3 (approved):
LGTM, thank you!

ybiquitous:
@jeddy3 Thanks for the quick review! Changelog entry added:

> - Fixed: `declaration-property-value-allowed-list` false positives for multiple match ([#6190](https://github.com/stylelint/stylelint/pull/6190)).

PR #6194, opened 2022-07-05
Add `percentage-unless-within-keyword-only-block` primary option to `keyframe-selector-notation`

### LINKED ISSUE #6191: Add `percentage-unless-within-keyword-only-block` primary option to keyframe-selector-notation

I only want to disallow mixed keyframe selectors:

```css
@keyframes foo {
  from { opacity: 0; }
  50% { opacity: 0.5; }
  to { opacity: 1; }
}
```

```css
@keyframes foo {
  from { opacity: 0; }
  100% { opacity: 0.5; }
}
```

While allowing _either_ of the notations when not mixed:

```css
@keyframes foo {
  from { opacity: 0; }
  to { opacity: 1; }
}
```

```css
@keyframes foo {
  0% { opacity: 0; }
  100% { opacity: 1; }
}
```

Ref: https://github.com/stylelint/stylelint-config-standard/issues/255

What's a good name for this option? Maybe:

```
ignore: ["after-same-notation"]`
```

We use `same` [else where](https://stylelint.io/user-guide/rules/list/at-rule-empty-line-before/#except-after-same-name-inside-block-blockless-after-same-name-blockless-blockless-after-blockless-first-nested) and the rule its scoped to the keyframe-selector rather than the keyframe-block.

ybiquitous:

Closes #6191

It's necessary to call `.walkRules()` twice to get all selectors in a `@keyframes` block. In most cases, I don't believe there will not pay a performance penalty.

### DIFF
--- a/lib/rules/keyframe-selector-notation/README.md
+++ b/lib/rules/keyframe-selector-notation/README.md
@@ -15,7 +15,7 @@ The [`fix` option](../../../docs/user-guide/usage/options.md#fix) can automatica
 
 ## Options
 
-`string`: `"keyword"|"percentage"`
+`string`: `"keyword"|"percentage"|"percentage-unless-within-keyword-only-block"`
 
 ### `"keyword"`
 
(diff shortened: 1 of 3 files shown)


jeddy3 (approved):
Fantastic, thanks!

LGTM.

Issue #5157, opened 2021-02-20
Fix false positives in *-no-unknown


Mouvedia:
Some pseudo classes/elements are missing from the whitelists of:
- selector-pseudo-element-no-unknown
- selector-pseudo-class-no-unknown

> sources

https://www.chromestatus.com/feature/5676301379698688
https://www.chromestatus.com/feature/6537562418053120
https://github.com/css-modules/css-modules

> What code is needed to reproduce the bug?

```css
::file-selector-button { }
:state() { }
```
The list is not exhaustive. Personally I think `:global` and `:local` could be considered as well.

> What stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "selector-pseudo-class-no-unknown": true,
    "selector-pseudo-element-no-unknown": true,
  }
}
```

> Which version of stylelint are you using?

https://stylelint.io/demo

> What did you expect to happen?

no errors

jeddy3:
@Mouvedia Thanks for the report and for using the template.

> The list is not exhaustive

We'll need to update our [reference lists of keywords](https://github.com/stylelint/stylelint/blob/master/lib/reference/keywordSets.js).

I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) this if you have time.

> Personally I think `:global` and `:local` could be considered as well.

The rules consider pseudo-classes and elements to be known. Those that belong to methodologies and language extensions can be allowed using the [`ignorePseudoClasses` and `ignorePseudoElements` options](https://stylelint.io/user-guide/rules/selector-pseudo-class-no-unknown#optional-secondary-options). This lets configuration author explicit signal when they are using something outside of the specs.

Mouvedia:
> Please consider contributing this if you have time.

@jeddy3 If I did, I would update the whitelists of selector-type-no-unknown and unit-no-unknown; is that a problem?

jeddy3:
@Mouvedia That would be fantastic! Since first creating the sets, we've updated them piecemeal as users have reported missing items. It'd be great to give them the once over as new CSS specifications are constantly being added and it's likely the lists have fallen out of date.

The sets are either broken down by the sections of a specification (and usually accompanied by the URL to the specification in a comment beforehand) or by the needs of a particular rule.

Thank you!

Mouvedia:
> It'd be great to give them the once over as new CSS specifications are constantly being added and it's likely the lists have fallen out of date.

[...]

PR #443, opened 2015-10-21
Add rule : color-no-named

makotot:
ref : #424
> is-named-css-color might be useful here.

I used [is-css-color-name](https://www.npmjs.com/package/is-css-color-name).
[is-named-css-color](https://www.npmjs.com/package/is-named-css-color) returns `true` if the color value such as `$color-name-black` and `var(--color-name-white)`. I think that the variables like `$color-name-black` should be valid in this rule because the variable that contains with color name is not always color name.

PR #6195, opened 2022-07-07
Fix `no-descending-specificity` false positives for pseudo-classes

### LINKED ISSUE #4010: Fix false positives for functional pseudo-classes in no-descending-specificity
> Clearly describe the bug

There are some error tips in my selectors was not expected..
It flagged an error about rule ` no-descending-specificity`  when i use css-module presudo function like `:global()` with a selector.
Thought the selectors had not any relations ,it always flagged an error.
If i rewrite the same selectors without `:global`, then it works.

no-descending-specificity

```css
.componentName :global(.ant-modal) :global(.ant-btn) {
    color: #fff;
}
.componentName :global(.ant-menu) {
    color: #fff;
}
```

e.g.

```javascript
module.exports = {
    extends: ['stylelint-config-standard', 'stylelint-config-css-modules', 'stylelint-config-prettier'],
    rules: {
        'at-rule-no-unknown': [true, { ignoreAtRules: ['function', 'return'] }]
    }
}

```

`9.10.1`

CLI with `stylelint "**/*.css" --config .stylelintrc.js`

Yes, it's related to css module pre-sudo function like `:global()`

No warnings to be flagged.

```shell
test.css
4:1  ✖  Expected selector ".componentName :global(.ant-menu)" to come before selector ".componentName :global(.ant-modal) :global(.ant-btn)"   no-descending-specificity
```

ybiquitous:

Closes #4010

This pull request aims to fix false positives for the `no-descending-specificity` rule about the [CSS Modules](https://github.com/css-modules/css-modules) `:global()` pseudo-class.

I'm unfamiliar with CSS Modules, so I'm unsure whether this added code specific to `:global()` is suitable. Perhaps, someone may know a more smart solution.

In addition, I don't know if we should address the `:local()` pseudo-class. Please let me know if someone knows anything about it.

See also https://github.com/stylelint/stylelint/issues/4010#issuecomment-1177319550 and the [reproduction demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0YCGAnKACYAdAdllkgM4DuAlgA4w4FFHwDmANhAEaosAUplN6ALQkWFWINRgALhQBuMAJQAaeg2ZtOPPtRhCRYmIKgBXKqLRTDkmfIV1CDIlPSp8JAGYR0AW3hYSaCww3ACMCgDcqkQAvqqx+PEESGyoUBT4TPaMrBxcvLKugilpGQD6pSRU6fi6dngORJRQUgAWfgAMSAAc6DDekQ7x0SBKICQAnvhSqAAeCOAkJCPgEPjuFEzz9US4IOjGQSS7ftsMu-gQRjABMPglTMI0YBTrz1Ljx1jOxjBxBMPRIA).


lib/rules/no-descending-specificity/index.js
```diff
@@ -92,23 +95,24 @@ const rule = (primary, secondaryOptions) => {
 		/**
 		 * @param {import('postcss-selector-parser').Root} selectorNode
 		 * @param {import('postcss').Rule} ruleNode
-		 * @param {Map<any, any>} comparisonContext
+		 * @param {Map<string, Entry[]>} comparisonContext
 		 */
 		function checkSelector(selectorNode, ruleNode, comparisonContext) {
 			const selector = selectorNode.toString();
-			const referenceSelectorNode = lastCompoundSelectorWithoutPseudoClasses(selectorNode);
+			const referenceSelector = lastCompoundSelectorWithoutPseudoClasses(selectorNode);
```
ybiquitous, line 102: [note] `lastCompoundSelectorWithoutPseudoClasses()` returns a string, not a node. So I think `referenceSelector` is a more readable name.

lib/rules/no-descending-specificity/index.js
```diff
@@ -137,17 +142,27 @@ function lastCompoundSelectorWithoutPseudoClasses(selectorNode) {
...
+	const nodesWithoutPseudoClasses = nodesAfterLastCombinator.filter((node) => {
+		return (
+			isCssModulesGlobal(node) ||
+			node.type !== 'pseudo' ||
+			node.value.startsWith('::') ||
+			pseudoElements.has(node.value.replace(/:/g, ''))
+		);
+	});
+
+	if (nodesWithoutPseudoClasses.length === 0) return undefined;
```
ybiquitous, line 155: [note] If `nodesWithoutPseudoClasses` is empty, the old version of this function returns an empty string (`''`). This makes it more difficult to handle irregular values. So the current code returns `undefined` instead.

richex-cn:
I created another [example](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0B3ATgQwA4AJgB0A7HHKASwGcsAbDAT3hwCNqIwBrAbiIF8ij4Ac1ZMM1ABRIyAW0EBKPERLkqtBs1YduhPoQHCIo6ouI4kaGFBMkSkVmkYWo2krt0gANCAp1CAFwwADwRwCgpPcAhCADMyQRCCU3wQNABXahgKZMZEmxxkwggAWlgKMBhCckJBIqoYMDJYhr86bJw-NJglHDcQHiA), But I'm unsure whether it has anything to do with this problem.

lib/rules/no-descending-specificity/index.js
```diff
@@ -137,17 +142,27 @@ function lastCompoundSelectorWithoutPseudoClasses(selectorNode) {
...
+
+	return nodesWithoutPseudoClasses.join('');
+}
+
+/**
+ * @param {import('postcss-selector-parser').Node} node
+ * @returns {boolean}
+ */
+function isCssModulesGlobal(node) {
```
Mouvedia, line 164: [previously @jeddy3 wanted to not handle `:global()`](https://github.com/stylelint/stylelint/issues/5157#issuecomment-782668548)

related: #443

Mouvedia:
> In addition, I don't know if we should address the :local() pseudo-class. Please let me know if someone knows anything about it.

see https://github.com/stylelint/stylelint/blob/main/docs/user-guide/rules/about.md#user-defined-ignore

ybiquitous:
@Mouvedia Thanks for providing the guidance. Indeed, I think it may be better that built-in rules don't support non-standard syntaxes like CSS Modules for simplicity and maintainability.

Mouvedia:
I don't have an opinion. I am just saying let's be consistent.

jeddy3:
Just getting back around to this. It's a can of worms and I don't think I've fully grokked it yet.

I think we want to address functional pseudo-classes more generally rather than just the non-standard `:global()` one.

The rule currently considers the following comparable when it should not:

```css
a :where(a, b) {}
:where(c, d) {}
```

[See demo of false positive](https://stylelint.io/demo#N4Igxg9gJgpiBcICGACeB3AFjATjAFEgDQoBGAlCsAL4A6AdhtnvmCVJTSESAM4Ce9AC5IAHgnC9e3cBHoAzAJYBzCcAYoUtEDgCuAGxi9t8Khs1aQ9CAFpYvMDHpRF9ZTd4AHGGEVLfQvwmKEJ6MOZ09NQg1EA). 

I haven't dug into the implementation of the rule deeply, but I think we should be including the [`:where` and `:is` functional pseudo-classes](https://drafts.csswg.org/selectors/#logical-combination) (and their arguments) in the reference selector.

Then, as `:global` is a non-standard ([sometimes functional](https://github.com/css-modules/css-modules/issues/264)) pseudo-class, we can add an `ignoreSelectorsContaining: []` option, which would not compare selectors that contain any of the specified parts.

(We could add a different option which lets a user specify a list of functional pseudo-class to treat like `:is` and `:where` but I don't think its worth doing down that rabbit hole.)

jeddy3:
> Then, as :global is a non-standard (https://github.com/css-modules/css-modules/issues/264) pseudo-class, we can add an ignoreSelectorsContaining: [] option, which would not compare selectors that contain any of the specified parts.

Let's add an `ignoreSelectors: []` option instead that checks against the whole selector. It's more straightforward and consistent with other rules that way. Users can then craft a regex to ignore selectors that contain `:global` and whatnot.

ybiquitous:
@jeddy3 Thanks for the suggestion. I mistakenly thought the CSS Modules specific code was required; indeed, it's unnecessary. See the commit 68bcaa which makes the test suite pass.

If the `ignoreSelectors` option you suggested may be needed in the future, but in this pull request, it's not needed for now.

jeddy3 (approved):
LGTM, thank you!

> If the ignoreSelectors option you suggested may be needed in the future, but in this pull request, it's not needed for now.

👍

ybiquitous:
@jeddy3 Thanks for the re-review and correcting the PR title! 😄  

Changelog entry added:

> - Fixed: `no-descending-specificity` false positives for pseudo-classes ([#6195](https://github.com/stylelint/stylelint/pull/6195)).

PR #6201, opened 2022-07-13
Fix end positions for `selector-not-notation`

Mouvedia:

#5694

No, it's self-explanatory.

### DIFF
--- a/lib/rules/selector-not-notation/index.js
+++ b/lib/rules/selector-not-notation/index.js
@@ -110,10 +110,13 @@ const rule = (primary, _, context) => {
 						if (mustFix) return fixSimple(pseudo);
 					}
 
+					assert(pseudo.source && pseudo.source.end);
+
 					report({
 						message: messages.expected(primary),
 						node: ruleNode,
 						index: pseudo.sourceIndex,
+						endIndex: pseudo.source.end.column,
 						result,
 						ruleName,
 					});
--- a/lib/rules/selector-not-notation/__tests__/index.js
+++ b/lib/rules/selector-not-notation/__tests__/index.js
@@ -38,6 +38,8 @@ testRule({
 			message: messages.expected('simple'),
 			line: 1,
 			column: 1,
+			endLine: 1,
+			endColumn: 13,
 		},
 		{
 			code: ':not(::before) {}',
@@ -45,6 +47,8 @@ testRule({
 			message: messages.expected('simple'),
 			line: 1,
 			column: 1,
+			endLine: 1,
+			endColumn: 15,
 		},
 		{
 			code: ':not(:first-line) {}',
@@ -52,6 +56,8 @@ testRule({
 			message: messages.expected('simple'),
 			line: 1,
 			column: 1,
+			endLine: 1,
+			endColumn: 18,
 		},
 		{
 			code: ':not(a, div) {}',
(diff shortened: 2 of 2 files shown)


lib/rules/selector-not-notation/index.js
```diff
@@ -110,10 +110,13 @@ const rule = (primary, _, context) => {
 						if (mustFix) return fixSimple(pseudo);
 					}
 
+					assert(pseudo.source && pseudo.source.end);
+
 					report({
 						message: messages.expected(primary),
 						node: ruleNode,
 						index: pseudo.sourceIndex,
+						endIndex: pseudo.source.end.column - 1,
```
ybiquitous, line 119: [ask] Should `endColumn` be exclusive like other rules?

```suggestion
						endIndex: pseudo.source.end.column,
```

ybiquitous (commented):
@Mouvedia Thanks for creating the pull request! I've left a question, so please check it.

lib/rules/selector-not-notation/index.js
Mouvedia: @ybiquitous AFAIK endIndex starts at 0 but column starts at 1 hence the -1.

e.g. for
`code: ':not(:not()) {}',`
I want to include `:not(:not())` and not `:not(:not()) `, correct?

ybiquitous: In the PostCSS API, `endColumn` means ["exclusive end position"](https://postcss.org/api/#warning-endcolumn). See below:

```css
:not(:not()) {}
/**         ↑    
↑-----------↑
1           13 (exclusive)
*/
```

In other rules, `endColumn` is exclusive, for example:

```css
a { opacity: 10% }
/**          ↑  ↑
             ↑--↑
             14 17
*/
```

https://github.com/stylelint/stylelint/blob/9802b9de2a40eb682df141ef2f121ade1631843c/lib/rules/alpha-value-notation/__tests__/index.js#L65-L71

lib/rules/selector-not-notation/index.js
```diff
@@ -110,10 +110,13 @@ const rule = (primary, _, context) => {
 						if (mustFix) return fixSimple(pseudo);
 					}
 
+					assert(pseudo.source && pseudo.source.end);
```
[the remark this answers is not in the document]
ybiquitous: We cannot use optional chaining since [Node.js 12 doesn't support it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility). 😅 

It should be possible in the next major version. (maybe we will drop the Node.js 12 support)

https://github.com/stylelint/stylelint/blob/9802b9de2a40eb682df141ef2f121ade1631843c/package.json#L190-L192

ybiquitous:
Changelog updated:

> - Fixed: `selector-not-notation` end positions ([#6201](https://github.com/stylelint/stylelint/pull/6201)).

lib/rules/selector-not-notation/index.js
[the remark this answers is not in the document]
Mouvedia: I have just tried `node --harmony-optional-chaining` (version 12.18) and `?.` works as intended.
Is it too cumbersome/impractical to add it?

PR #6205, opened 2022-07-15
Fix `named-grid-areas-no-invalid` end positions

ybiquitous:

Part of #5694

This pull request aims to fix end positions for the [`named-grid-areas-no-invalid`](https://github.com/stylelint/stylelint/blob/fix-end-positions-for-named-grid-areas-no-invalid/lib/rules/named-grid-areas-no-invalid/README.md) rule.

There are 3 patterns, like this:

```css
a { grid-template-areas: "" }
/*                       ↑↑  'Expected cell token within string' */

a { grid-template-areas: "a a a" "b b b b" }
/*                               ↑-------↑  'Expected same number of cell tokens in each string' */

a { grid-template-areas: "a a a" "b b a" }
/*                       ↑-----↑    'Expected single filled-in rectangle for "a"' */
```

The last pattern may be better if it were as below:

```css
a { grid-template-areas: "a a a" "b b a" }
/*                                    ↑ */
```

But it's hard to find positions to match the last pattern precisely because the `findNotContiguousOrRectangular()` logic is complex. I think we can improve them later if needed.

Also, the `isRectangular()` module is removed and inlined for simplicity.


jeddy3 (approved):
Nice! LGTM

> I think we can improve them later if needed.

👍

ybiquitous:
@jeddy3 Thanks! Changelog updated:

> - Fixed: `named-grid-areas-no-invalid` end positions ([#6205](https://github.com/stylelint/stylelint/pull/6205)).

PR #6219, opened 2022-07-26
Fix `declaration-block-no-redundant-longhand-properties` end positions

ybiquitous:

Part of #5694

No, it's self-explanatory.


lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -93,6 +93,7 @@ const rule = (primary, secondaryOptions) => {
 						ruleName,
 						result,
 						node: decl,
+						word: decl.prop,
```
ybiquitous, line 96: [note] If not specify `word`, `endColumn` would point to the end of the declaration's value. However, I think it's too long. So this change indicates just a property name.

mattxwang (approved):
I think just highlighting the property for the first redundant longhand property makes sense. LGTM!

jeddy3 (approved):

ybiquitous:
Changelog entry added:

> - Fixed: `declaration-block-no-redundant-longhand-properties` end positions ([#6219](https://github.com/stylelint/stylelint/pull/6219)).

PR #6222, opened 2022-07-27
Fix `rule-selector-property-disallowed-list` end positions

ybiquitous:

Part of #5694

No, it's self-explanatory.


mattxwang (approved):
Read-through + manually looking at the test cases, this looks good to me! Great work 😊 

(I also had a chance to clone and test with multiple decls, and in this case there's no issue with duplication. Very convenient!)

jeddy3 (approved):
LGTM, thank you!

ybiquitous:
Changelog entry added:

> - Fixed: `rule-selector-property-disallowed-list` end positions ([#6222](https://github.com/stylelint/stylelint/pull/6222)).

PR #6227, opened 2022-07-27
Fix `selector-class-pattern` end positions

ybiquitous:

Part of #5694

This changes the implementation of `messages.expect` a bit:

```diff
- messages.expected('foo', /^[A-Z]+$/)
- Expected class selector ".${selector}" to match pattern "${pattern}"

+ messages.expected('.foo', /^[A-Z]+$/)
+ Expected class selector "${selector}" to match pattern "${pattern}"
```

Because I think it's clear to pass a full selector explicitly. The output message remains the same in both cases.

### DIFF
--- a/lib/rules/selector-class-pattern/__tests__/index.js
+++ b/lib/rules/selector-class-pattern/__tests__/index.js
@@ -54,15 +54,19 @@ testRule({
 	reject: [
 		{
 			code: 'a .foo {}',
-			message: messages.expected('foo', /^[A-Z]+$/),
+			message: messages.expected('.foo', /^[A-Z]+$/),
 			line: 1,
 			column: 3,
+			endLine: 1,
+			endColumn: 7,
 		},
 		{
 			code: '.ABABA > .bar {}',
-			message: messages.expected('bar', /^[A-Z]+$/),
+			message: messages.expected('.bar', /^[A-Z]+$/),
 			line: 1,
 			column: 10,
+			endLine: 1,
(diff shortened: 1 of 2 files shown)


lib/rules/selector-class-pattern/__tests__/index.js
```diff
@@ -141,9 +149,11 @@ testRule({
 	reject: [
 		{
 			code: '.A { &__B { }}',
-			message: messages.expected('A__B', /^[A-Z]+$/),
+			message: messages.expected('.A__B', /^[A-Z]+$/),
 			line: 1,
 			column: 6,
+			endLine: 1,
+			endColumn: 11,
```
ybiquitous, line 156: [note] Correctly, `endColumn` should be `10` but actually `11`:

```scss
.A { &__B { }}
/*   ↑   ↑
     6   10
*/
```

`11` is calculated from `6 + ".A__B".length` (`".A__B"` is a resolved selector).

For details, see the note in the source code:
https://github.com/stylelint/stylelint/pull/6227/files#diff-3cdac1b62a3c129821ddbefec56839ea6af7b04c9b78a7c3f4ea5054197bc74fR92

jeddy3 (approved):
LGTM, thank you!

Fantastic work on all these end position fixes, btw!

ybiquitous:
Changelog entry added:

> - Fixed: `selector-class-pattern` end positions ([#6227](https://github.com/stylelint/stylelint/pull/6227)).

PR #6232, opened 2022-07-28
Add test cases for end positions to `selector-max-*`

ybiquitous:

Part of #5694

The test cases for the following rules:

- `selector-max-attribute`
- `selector-max-class`
- `selector-max-combinators`

Some cases are hard to address due to resolved selectors. I leave `TODO` comments in the test code.


lib/rules/selector-max-class/__tests__/index.js
```diff
@@ -100,48 +102,66 @@ testRule({
...
 		},
 		{
 			code: '.ab { &:hover > .ef.gh {} }',
 			description: 'nested selectors: greater than max classes',
 			message: messages.expected('.ab:hover > .ef.gh', 2),
 			line: 1,
 			column: 7,
+			endLine: 1,
+			endColumn: 26,
+			// TODO: The following case is actually correct. It's hard to get the original selector before resolved.
```
Mouvedia, line 152: Why don't you move it to the accepted ones but skip it instead ?

lib/rules/selector-max-combinators/__tests__/index.js
```diff
@@ -242,11 +282,17 @@ testRule({
...
+					endLine: 1,
+					endColumn: 10,
 				},
 				{
 					message: messages.expected('foo + bar .aa', 0),
 					line: 1,
 					column: 29,
+					endLine: 1,
+					endColumn: 35,
+					// TODO: The following case is actually correct. It's hard to get the original selector before resolved.
```
Mouvedia, line 294: Shouldn't you create a github issue for it, instead of several TODOs?

lib/rules/selector-max-class/__tests__/index.js
ybiquitous: If specifying `skip: true`, the case itself is skipped, and `message` or `column` also is not tested.

lib/rules/selector-max-combinators/__tests__/index.js
ybiquitous: It's a good idea to create an issue and I'll do it.

But I think it's also worth leaving TODOs because we tend to forget an issue in the many current issues.

ybiquitous: I've opened an issue #6234.

jeddy3 (approved):
LGTM, thank you!

We can address the inaccuracies around resolved nested selectors at a later date.

PR #6241, opened 2022-08-04
Restore `Node#rangeBy` method if not present

### LINKED ISSUE #5766: Fix "TypeError: opts.node.rangeBy is not a function" with PostCSS 8.4.4

folder structure:
  package.json
  .stylelintrc.json
  styles.scss (empty file)

 package.json:
```js
{
  "name": "untitled2",
  "version": "0.0.0",
  "private": true,
  "scripts": {
    "lint": "npx stylelint '**/*.scss'"
  },
  "devDependencies": {
    "stylelint": "^14.1.0",
    "stylelint-config-standard-scss": "^3.0.0"
  }
}
```

**npm i**
**npm run lint**

```js
{
}
```

npx stylelint '**/*.scss'

^14.1.0

I expect stylelint to show list of errors.

### What actually happened?

```
TypeError: opts.node.rangeBy is not a function
at new Warning (node_modules/postcss/lib/warning.js:9:29)
at Result.warn (node_modules/postcss/lib/result.js:26:19)
at report (node_modules/stylelint/lib/utils/report.js:103:9)
at node_modules/stylelint/lib/rules/no-empty-source/index.js:28:3
at node_modules/stylelint/lib/lintPostcssResult.js:112:8
at Array.map (<anonymous>)
at lintPostcssResult (node_modules/stylelint/lib/lintPostcssResult.js:103:18)
at lintSource (node_modules/stylelint/lib/lintSource.js:88:8)
at async node_modules/stylelint/lib/standalone.js:218:27
at async Promise.all (index 0)

Process finished with exit code 1
```

The problem is in this line

image

No

### Proposal to fix the bug

Works correctly with exactly 8.3.11 version of postcss. 
I compared the code in `warning.js` between these versions of postcss (8.3.11 and 8.4.4), they changed the method name. :(

Proposal: change dependencies of stylelint to use exactly 8.3.11 version.

ybiquitous:

Closes #5766

This PR adds a hack to a situation where there are different PostCSS versions in a dependency tree like a reproduction described on <https://github.com/stylelint/stylelint/issues/5766#issuecomment-1200483928>.

[PostCSS 8.4.0](https://github.com/postcss/postcss/releases/tag/8.4.0) has added the [`Node#rangeBy`](https://postcss.org/api/#node-rangeby) method, while PostCSS 8.3.x or lower doesn't have the method. In the situation above, a parser like `postcss-less` may create a `Node` by an old PostCSS version like 8.3.6. Because the parser's peer dependency version on PostCSS may not match the PostCSS version that Stylelint requires.

* [`postcss-less@6.0.0` requires `postcss@^8.3.5`](https://github.com/shellscape/postcss-less/blob/v6.0.0/package.json#L49-L51) as a peer dependency
* [`stylelint@14.9.1` requires `postcss@^8.4.14`](https://github.com/stylelint/stylelint/blob/14.9.1/package.json#L136) as a normal dependency

Note: If anyone is concerned about adding such a hack, please let me know.

### DIFF
--- a/lib/utils/report.js
+++ b/lib/utils/report.js
@@ -110,6 +110,8 @@ module.exports = function report(problem) {
 	result.warn(warningMessage, warningProperties);
 };
 
+let warningShown = false;
+
 /**
  * @param {import('postcss').Node | undefined} node
  * @param {number | undefined} index
@@ -124,9 +126,13 @@ function getStartLine(node, index, endIndex) {
 	//       that doesn't have the method.
 	//       See https://github.com/stylelint/stylelint/issues/5766
 	if (typeof node.rangeBy !== 'function') {
-		console.warn(
-			'It appears different PostCSS versions are used in your dependency tree. Please consider updating PostCSS to use one version.',
-		);
+		if (!warningShown) {
+			console.warn(
+				'It appears different PostCSS versions are used in your dependency tree. Please consider updating PostCSS to use one version. See https://github.com/stylelint/stylelint/issues/5766 for details.',
+			);
+			warningShown = true;
+		}
+
 		node.rangeBy = postcss.Node.prototype.rangeBy.bind(node);
 	}
 
--- a/lib/utils/report.js
+++ b/lib/utils/report.js
@@ -123,7 +123,10 @@ function getStartLine(node, index, endIndex) {
 	//       while the `node` may be produced by PostCSS 8.3.x or lower
 	//       that doesn't have the method.
 	//       See https://github.com/stylelint/stylelint/issues/5766
-	if (node.rangeBy === undefined) {
+	if (typeof node.rangeBy !== 'function') {
+		console.warn(
+			'It appears different PostCSS versions are used in your dependency tree. Please consider updating PostCSS to use one version.',
+		);
 		node.rangeBy = postcss.Node.prototype.rangeBy.bind(node);
 	}
 
--- a/lib/utils/report.js
+++ b/lib/utils/report.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const postcss = require('postcss');
+
 /**
  * Report a problem.
  *
@@ -28,11 +30,9 @@ module.exports = function report(problem) {
 		return;
 	}
 
-	const { start } = (node && node.rangeBy({ index, endIndex })) || {};
-
 	// If a line is not passed, use the node.rangeBy method to get the
 	// line number that the complaint pertains to
-	const startLine = line || (start && start.line);
+	const startLine = line || getStartLine(node, index, endIndex);
 
 	if (!startLine) {
 		throw new Error('You must pass either a node or a line number');
@@ -109,3 +109,23 @@ module.exports = function report(problem) {
 
 	result.warn(warningMessage, warningProperties);
 };
+
+/**
+ * @param {import('postcss').Node | undefined} node
+ * @param {number | undefined} index
+ * @param {number | undefined} endIndex
+ * @returns {number | undefined}
+ */
+function getStartLine(node, index, endIndex) {
+	if (node === undefined) return undefined;
+
+	// HACK: The `rangeBy` method has been added since PostCSS 8.4.0,
+	//       while the `node` may be produced by PostCSS 8.3.x or lower
+	//       that doesn't have the method.
+	//       See https://github.com/stylelint/stylelint/issues/5766
+	if (node.rangeBy === undefined) {
+		node.rangeBy = postcss.Node.prototype.rangeBy.bind(node);
+	}
+
+	return node.rangeBy({ index, endIndex }).start.line;
+}
--- a/lib/utils/report.js
+++ b/lib/utils/report.js
@@ -117,7 +117,7 @@ module.exports = function report(problem) {
  * @returns {number | undefined}
  */
 function getStartLine(node, index, endIndex) {
-	if (node === undefined) return undefined;
+	if (!node) return;
 
 	// HACK: The `rangeBy` method has been added since PostCSS 8.4.0,
 	//       while the `node` may be produced by PostCSS 8.3.x or lower
--- a/lib/utils/report.js
+++ b/lib/utils/report.js
@@ -133,7 +133,7 @@ function getStartLine(node, index, endIndex) {
 			warningShown = true;
 		}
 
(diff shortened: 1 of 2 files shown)


lib/utils/report.js
```diff
@@ -28,11 +30,9 @@ module.exports = function report(problem) {
 		return;
 	}
 
-	const { start } = (node && node.rangeBy({ index, endIndex })) || {};
-
 	// If a line is not passed, use the node.rangeBy method to get the
 	// line number that the complaint pertains to
-	const startLine = line || (start && start.line);
+	const startLine = line || getStartLine(node, index, endIndex);
```
ybiquitous, line 35: [note] Call the `rangeBy` method lazily for performance.

lib/utils/report.js
```diff
@@ -109,3 +109,23 @@ module.exports = function report(problem) {
...
+function getStartLine(node, index, endIndex) {
+	if (node === undefined) return undefined;
+
+	// HACK: The `rangeBy` method has been added since PostCSS 8.4.0,
+	//       while the `node` may be produced by PostCSS 8.3.x or lower
+	//       that doesn't have the method.
+	//       See https://github.com/stylelint/stylelint/issues/5766
+	if (node.rangeBy === undefined) {
+		node.rangeBy = postcss.Node.prototype.rangeBy.bind(node);
+	}
```
ybiquitous, line 128: [note] The implementation of `rangeBy` is here:

https://github.com/postcss/postcss/blob/8.4.14/lib/node.js#L260

This access to `postcss.Node.prototype` may be flaky because of a dependency on the internal implementation of PostCSS.

lib/utils/report.js
```diff
@@ -109,3 +109,23 @@ module.exports = function report(problem) {
...
+ * @returns {number | undefined}
+ */
+function getStartLine(node, index, endIndex) {
+	if (node === undefined) return undefined;
+
+	// HACK: The `rangeBy` method has been added since PostCSS 8.4.0,
+	//       while the `node` may be produced by PostCSS 8.3.x or lower
+	//       that doesn't have the method.
+	//       See https://github.com/stylelint/stylelint/issues/5766
+	if (node.rangeBy === undefined) {
```
JounQin, line 126: We can add a warning for the user here?

JounQin (changes requested):

lib/utils/report.js
JounQin:
```suggestion
	if (typeof node.rangeBy !== 'function') {
```

JounQin (changes requested):

lib/utils/report.js
ybiquitous: > We can add a warning for the user here?

Nice idea. Changed via 3745154a8. 👍🏼

lib/utils/report.js
```diff
@@ -109,3 +109,26 @@ module.exports = function report(problem) {
...
 };
+
+/**
+ * @param {import('postcss').Node | undefined} node
+ * @param {number | undefined} index
+ * @param {number | undefined} endIndex
+ * @returns {number | undefined}
+ */
+function getStartLine(node, index, endIndex) {
+	if (node === undefined) return undefined;
```
JounQin, line 120:
```suggestion
	if (!node) return;
```

Maybe a person taste here, or do we have a lint rule to enforce a consistent style?

lib/utils/report.js
```diff
@@ -109,3 +109,26 @@ module.exports = function report(problem) {
...
+
+	// HACK: The `rangeBy` method has been added since PostCSS 8.4.0,
+	//       while the `node` may be produced by PostCSS 8.3.x or lower
+	//       that doesn't have the method.
+	//       See https://github.com/stylelint/stylelint/issues/5766
+	if (typeof node.rangeBy !== 'function') {
+		console.warn(
```
JounQin, line 127: We can link #5766 as reference at the same time?

And we can check whether it has been warned to ensure it's not noise.

lib/utils/report.js
```diff
@@ -109,3 +109,26 @@ module.exports = function report(problem) {
...
+			'It appears different PostCSS versions are used in your dependency tree. Please consider updating PostCSS to use one version.',
+		);
+		node.rangeBy = postcss.Node.prototype.rangeBy.bind(node);
```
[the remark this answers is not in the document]
ybiquitous: It may work without `bind` now, but I've added it just in case. Because it's hard to predict how `rangeBy` is and will be used inside PostCSS.

Or, is there some disadvantage to adding `bind`?

JounQin (changes requested):

lib/utils/report.js
ybiquitous: I have no strong opinion about this style. 😅

lib/utils/report.js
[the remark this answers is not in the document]
JounQin: The prototype itself does not bind, so we don't need to do that neither.

lib/utils/report.js
ybiquitous: Make sense. Changed via 95a1ee6de. 👍🏼

lib/utils/report.js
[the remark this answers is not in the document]
ybiquitous: I see. Remove via 02a9c7518. 👍🏼

JounQin (approved):

JounQin:
Awesome work!

lib/utils/report.js
```diff
@@ -109,3 +109,32 @@ module.exports = function report(problem) {
...
+	//       See https://github.com/stylelint/stylelint/issues/5766
+	if (typeof node.rangeBy !== 'function') {
+		if (!warningShown) {
+			console.warn(
+				'It appears different PostCSS versions are used in your dependency tree. Please consider updating PostCSS to use one version. See https://github.com/stylelint/stylelint/issues/5766 for details.',
+			);
+			warningShown = true;
+		}
+
+		node.rangeBy = postcss.Node.prototype.rangeBy;
```
Mouvedia, line 136: If we ever drop Node.js 12, we could use a dynamic import in the `if`.

JounQin: We don't need this PR if https://github.com/stylelint/stylelint/issues/5766#issuecomment-1207260007 works.

Issue #6234, opened 2022-07-28
Fix inaccurate end positions when `postcss-resolve-nested-selector` is used in `selector-*`


ybiquitous:
### What steps are needed to reproduce the bug?

Because the [`postcss-resolve-nested-selector`](https://github.com/davidtheclark/postcss-resolve-nested-selector) package returns just a list of strings and loses a raw source position, there are some cases that we cannot calculate Stylelint problems' end positions accurately.

### What Stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "selector-class-pattern": ["^[A-Z]+$", { "resolveNestedSelectors": true }]
  }
}
```

### How did you run Stylelint?

CLI with `sylelint *.css`

### Which version of Stylelint are you using?

HEAD `ef37dc54e5e08786750baea0433114b0c56e965a`

### What did you expect to happen?

A correct end positions for nested selectors should be reported.

### What actually happened?

See the following case:

https://github.com/stylelint/stylelint/blob/ef37dc54e5e08786750baea0433114b0c56e965a/lib/rules/selector-class-pattern/index.js#L92-L98

the `endColumn` value should be `10` but actually `11`:

https://github.com/stylelint/stylelint/blob/ef37dc54e5e08786750baea0433114b0c56e965a/lib/rules/selector-class-pattern/__tests__/index.js#L168-L175

```css
.A { &__B { }}
/*   ↑   ↑
     ↑   ↑
     6   10
*/
```

### Does the bug relate to non-standard syntax?

Yes, it's related to nested selectors

### Proposal to fix the bug

I don't have a good idea now.

<hr />

- [x] `no-descending-specificity`
- [x] `no-duplicate-selectors`
- [x] `selector-max-attribute`
- [x] `selector-max-class`
- [x] `selector-max-combinators`
- [x] `selector-max-id`
- [x] `selector-max-type`
- [x] `selector-max-compound-selectors`
- [x] `selector-max-universal`
- [x] `selector-max-pseudo-class`
- [x] `selector-max-specificity`
- [ ] `selector-no-qualifying-type`

see also: https://github.com/stylelint/stylelint/issues/7482

jeddy3:
> I don't have a good idea now.

Nor do I. I've labelled the issue as ready to implement in case any has any ideas.

mattxwang:
[...]

PR #6242, opened 2022-08-05
Add test cases for `selector-max-compound-selectors` end positions

ybiquitous:

Part of #5694
Related to #6234

This PR adds test cases for the `selector-max-compound-selectors` rule's end positions. No runtime code changes.

Note that there are some cases for a nested selector with inaccurate end positions. This is related to `postcss-resolve-nested-selector` and will be resolved by #6234.

https://github.com/stylelint/stylelint/blob/e424be5eca77c5532b13147c4a7dad14085a4276/lib/rules/selector-max-compound-selectors/index.js#L82-L85


lib/rules/selector-max-compound-selectors/__tests__/index.js
```diff
@@ -184,10 +200,16 @@ testRule({
...
+			endLine: 2,
+			endColumn: 18,
 		},
 		{
 			code: '.cd { .de { .fg {} } }',
 			message: messages.expected('.cd .de .fg', 2),
+			line: 1,
+			column: 13,
+			endLine: 1,
+			endColumn: 19,
```
ybiquitous, line 212: [note] The accuracy of the end column is lost due to `postcss-resolve-nested-selector`. It should be helpful that the end column would point to just a selector instead of the rule node, but this is not the case. See the example below.

Expected:

```scss
.cd { .de { .fg {} } }
/*          ^^^
            ↑  ↑
            13 16
*/
```

Actual:

```scss
.cd { .de { .fg {} } }
/*          ^^^^^^
            ↑     ↑
            13    19
*/
```

jeddy3 (approved):
LGTM, thank you!

Issue #5221, opened 2021-03-31
Invalid Option Error thrown with "no-duplicate-selectors" rule.


james0r:
> Clearly describe the bug

Invalid Option Error thrown when using `"no-duplicate-selectors": false` rule even though the rule seems to take effect on my SCSS with stylelint-scss plugin added.

> Which rule, if any, is the bug related to?

no-duplicate-selectors

> What code is needed to reproduce the bug?

```module.exports = {
  extends: "stylelint-config-recommended",
  plugins: [
    /**
     * add plugins
     * docs: https://stylelint.io/user-guide/configure#plugins
     */
    'stylelint-scss'
  ],
  ignoreFiles: [
    /**
     * ignore certain files
     * docs: https://stylelint.io/user-guide/configure#ignorefiles
     */
    // 'my-file.css',
    // '**/my-directory/*.css'
  ],
  rules: {
    /**
     * add custom rules
     * docs: https://stylelint.io/user-guide/rules/list
     */
    "at-rule-no-unknown": null,
    "scss/at-rule-no-unknown": true,
    "no-empty-source": false,
    "no-descending-specificity": null,
    "no-duplicate-selectors": false
  },
};

```

> What stylelint configuration is needed to reproduce the bug?

See above.

> Which version of stylelint are you using?

13.10.0

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

Webpack plugin

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

Yes, SCSS

> What did you expect to happen?

Duplicate selectors are ignored and no errors to stdout.

> What actually happened (e.g. what warnings or errors did you get)?

```ERROR in Invalid Option: Unexpected option value "false" for rule "no-duplicate-selectors"
Invalid Option: Unexpected option value "false" for rule "no-empty-source"
```

hudochenkov:
Thanks for the report and for using the template.

To [disable a rule](https://stylelint.io/user-guide/configure#disable-errors) use `null`, and not `false`.

james0r:
Thanks for your reply.

This would be a helpful addition to the detail page for this rule and perhaps others.

https://stylelint.io/user-guide/rules/no-duplicate-selectors

I do not see the null value indicated anywhere there.

To add to the confusion the nested option takes 'false'.

Issue #3516, opened 2018-08-06
Disable `no-descending-specificity`


verpixelt:
The `no-descending-specificity` rule makes absolute sense with traditional ways of writing CSS. As soon as you enter the realm of selectors with unique hashes on the other hand it doesn't anymore.

After having updated to the latest version[9.4.0] of stylelint this rule breaks every lint run. `false` or `"never"` as values throw the following error `Invalid Option: Unexpected option value "false" for rule "no-descending-specificity"`.

hudochenkov:
Please, use issue template next time.

To [disabled rule](https://stylelint.io/user-guide/configure#rules) use `null` value.

Sowed:
> Please, use issue template next time.
> 
> To [disabled rule](https://stylelint.io/user-guide/faq/#how-do-i-turn-off-disable-or-ignore-a-rule) use `null` value.

@hudochenkov  This link seems to be broken at the moment.

hudochenkov:
We removed this page recently. Now the answer lays in line of how to turn on/turn off rule documentation: https://stylelint.io/user-guide/configure#rules

hudochenkov:
@OZZlE https://stylelint.io/user-guide/configure/#rules

> `null` (to turn the rule off)

OZZlE:
My bad I was tired and doing it inside eslint config instead of `.stylelintrc` :D for reference:

    {
      "rules": {
        "no-descending-specificity": null
      }
    }

olariuromeo:
> Please, use issue template next time.
> 
> To [disabled rule](https://stylelint.io/user-guide/faq/#how-do-i-turn-off-disable-or-ignore-a-rule) use `null` value.

Page Not Found
We could not find what you were looking for.

Please contact the owner of the site that linked you to the original URL and let them know their link is broken.

jeddy3:
I've edited the [post above](https://github.com/stylelint/stylelint/issues/3516#issuecomment-410714945) to use the [correct URL](https://stylelint.io/user-guide/configure#rules).

Complete diff syntax for a `stylelint.config.mjs` file is:

```diff js
/** @type {import('stylelint').Config} */
export default {
  extends: ["stylelint-config-standard"],
+ rules: {
+   "no-descending-specificity": null
+ }
};
```

Issue #2027, opened 2016-11-01
option to disable 'length-zero-no-unit'


dmerc:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

i am extending "stylelint-config-standard", it sets 'length-zero-no-unit' to true. I would like to disable it, but config `"rules": { "length-zero-no-unit": false }` is invalid.

> Which rule, if any, is this issue related to?

length-zero-no-unit

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "extends": "stylelint-config-standard",
  "rules": {
    "length-zero-no-unit": false
  }
}
```

> Which version of stylelint are you using?

7.5.1

> How are you running stylelint: CLI, PostCSS plugin, Node API?

using webpack-stylelint-loader

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

no

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

The following warnings were flagged:

```shell
Invalid Option: Unexpected option value "false" for rule "length-zero-no-unit"
```

m-allanson:
@dmerc try using `null`, rather than `false`.

https://github.com/stylelint/stylelint/blob/master/docs/user-guide/faq.md#how-do-i-disable-a-rule

Issue #1953, opened 2016-10-05
Can't override rules while extends a config


dungdm93:
For example, I want to extend standard rule, but don't wanna use `no-empty-source` rule:
```json
// .stylelintrc file
{
  "extends": "stylelint-config-standard",
  "rules": {
    "no-empty-source": false
  },
  "ignoreFiles": "./dist/**"
}
```
I got an error while run stylelint:
```
Invalid Option: Unexpected option value "false" for rule "no-empty-source"
```

I considered the [offical document](http://stylelint.io/user-guide/rules/) and see that a tons of rules have only one option: [`color-no-hex`](http://stylelint.io/user-guide/rules/color-no-hex/), [`string-no-newline`](http://stylelint.io/user-guide/rules/string-no-newline/), [`property-no-unknown`](http://stylelint.io/user-guide/rules/property-no-unknown/), [`value-no-vendor-prefix`](http://stylelint.io/user-guide/rules/value-no-vendor-prefix/), [`declaration-no-important`](http://stylelint.io/user-guide/rules/declaration-no-important/)... for example. 
So, should we add negative option for this case.

P/S: I used stylelint within gulp and it speed is **extremely** slow.

jeddy3:
This is covered in the [FAQ](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/faq.md#how-do-i-disable-a-rule) sections of the docs, which is linked to from the ["Need Help?](https://github.com/stylelint/stylelint#need-help) section in the main README.

Use `null` e.g. `"no-empty-source": null`.

> P/S: I used stylelint within gulp and it speed is extremely slow

Please raise this in the gulp-stylelint issue tracker.

In the future please use the issue template as it helps is process issues more quickly.

PR #6250, opened 2022-08-08
Add invalid rule primary option message for `false`

### LINKED ISSUE #5091: Fix invalid option message for false for better DX
> Clearly describe the bug

Extending the standard config StyleLint gives me an error about the unit after a zero value. This cannot be undone disabling the corresponding rule because that causes another error.

`length-zero-no-unit`

```css
.foobar {
  margin: 0px; /* Unexpected unit (length-zero-no-unit) */
}
```

```js
module.exports = {
  extends: "stylelint-config-standard",
  rules: {
    "length-zero-no-unit": false
  }
};
```

13.7.2

CLI with `stylelint test.css --config stylelint.config.js`

No.

Being able to disable the rule and not get the complaint about the unit after a zero value, anymore.

```plain
Unexpected option value "false" for rule "length-zero-no-unit"
```

Mouvedia:

Closes #5091
 
see also #2027 #5221 #3516 #1953

~check the message of the first commit~
https://github.com/stylelint/stylelint/commit/101f2740f85315cb48f27714cfe4119b79beb6a3#diff-ea40f7b7d9827960e2d8e1590b37c7490648acb555d8fe8c22d0c2328449c95fR64

### DIFF
--- a/lib/rules/selector-no-qualifying-type/index.js
+++ b/lib/rules/selector-no-qualifying-type/index.js
@@ -63,7 +63,7 @@ const rule = (primary, secondaryOptions) => {
 			ruleName,
 			{
 				actual: primary,
-				possible: [true, false],
+				possible: [true],
 			},
 			{
 				actual: secondaryOptions,
--- a/lib/utils/validateOptions.js
+++ b/lib/utils/validateOptions.js
@@ -66,6 +66,12 @@ function validate(opts, ruleName, complain) {
 	const actual = opts.actual;
 	const optional = opts.optional;
 
+	if (actual === false && !ruleName.startsWith('report')) {
+		return complain(
+			`Invalid option value "false" for rule "${ruleName}". Are you trying to disable this rule? If so use "null" instead`,
+		);
+	}
+
 	if (actual === null || arrayEqual(actual, [null])) {
 		return;
 	}
--- a/lib/__tests__/fixtures/plugin-async.js
+++ b/lib/__tests__/fixtures/plugin-async.js
@@ -7,7 +7,7 @@ const ruleName = 'plugin/async';
 const rule = (enabled) => (root, result) => {
 	const validOptions = stylelint.utils.validateOptions(result, ruleName, {
 		actual: enabled,
-		possible: [true, false],
+		possible: [true],
 	});
 
 	if (!validOptions) {
--- a/lib/utils/__tests__/validateOptions.test.js
+++ b/lib/utils/__tests__/validateOptions.test.js
@@ -26,17 +26,17 @@ describe('validateOptions for primary options', () => {
 		expect(result.warn.mock.calls[0][0]).toBe('Invalid option value "d" for rule "foo"');
 	});
 
-	it('passing boolean equivalence', () => {
+	it('passing null equivalence', () => {
 		validateOptions(result, 'foo', {
-			possible: [true, false],
-			actual: false,
+			possible: [true, null],
+			actual: null,
 		});
 		expect(result.warn).toHaveBeenCalledTimes(0);
 	});
 
-	it('failing boolean equivalence', () => {
+	it('failing null equivalence', () => {
 		validateOptions(result, 'foo', {
-			possible: [true, false],
+			possible: [true, null],
 			actual: 'a',
 		});
 		expect(result.warn).toHaveBeenCalledTimes(1);
@@ -204,7 +204,7 @@ describe('validateOptions for `*-no-*` rule with no valid options', () => {
 		});
 		expect(result.warn.mock.calls[0]).toHaveLength(2);
 		expect(result.warn.mock.calls[0][0]).toBe(
-			'Unexpected option value "false" for rule "no-dancing"',
+			'Invalid option value "false" for rule "no-dancing". Are you trying to disable this rule? If so use "null" instead',
 		);
 	});
 


Mouvedia:
This line
`for (const a of [actual].flat()) {`
in validateOptions.js motivated the change to the config.
Don't squash the commits so that it can be reverted easily.

Mouvedia:
WIP
gotta fix the tests

ybiquitous:
@Mouvedia Thanks for creating the PR!

The 66cdca0f709848f7e11a1e461b66ae07822766ba commit looks good to me. 👍🏼 

> Don't squash the commits so that it can be reverted easily.

We usually "squash and merge", so if you open a new PR including the commit above, I'll approve and merge it soon.

lib/utils/validateOptions.js
```diff
@@ -65,6 +65,17 @@ function validate(opts, ruleName, complain) {
...
 	const optional = opts.optional;
+	const whitelist = [
+		'reportDescriptionlessDisables',
+		'reportInvalidScopeDisables',
+		'reportNeedlessDisables',
+	];
+
+	if (actual === false && !whitelist.includes(ruleName)) {
+		return complain(
+			`Invalid option value \`false\` for rule "${ruleName}". Are you trying to disable this rule? If so use \`null\` instead`,
```
ybiquitous, line 76: [suggest] `validateOptions` uses `"` consistently to embed any values into a string,

https://github.com/stylelint/stylelint/blob/0bd6976fea7fa2a07c47672de424194716d31164/lib/utils/validateOptions.js#L180-L186

so I think using `"` is better for `false` or `null`.

```suggestion
			`Invalid option value "false" for rule "${ruleName}". Are you trying to disable this rule? If so use "null" instead`,
```

lib/utils/validateOptions.js
```diff
@@ -65,6 +65,17 @@ function validate(opts, ruleName, complain) {
 	const possible = opts.possible;
 	const actual = opts.actual;
 	const optional = opts.optional;
+	const whitelist = [
+		'reportDescriptionlessDisables',
+		'reportInvalidScopeDisables',
+		'reportNeedlessDisables',
+	];
```
ybiquitous, line 72: I think it's reasonable that `validateDisableSettings` uses internally `validateOptions` to re-use the logic of `validateOptions`:

https://github.com/stylelint/stylelint/blob/0bd6976fea7fa2a07c47672de424194716d31164/lib/validateDisableSettings.js#L50-L63

However, I don't think it's maintainable the `whitelist` array depends implicitly on the callees of `validateDisableSettings`; we would have to change two locations if we would add a new disable setting: 

https://github.com/stylelint/stylelint/blob/0bd6976fea7fa2a07c47672de424194716d31164/lib/invalidScopeDisables.js#L13

https://github.com/stylelint/stylelint/blob/0bd6976fea7fa2a07c47672de424194716d31164/lib/needlessDisables.js#L17

https://github.com/stylelint/stylelint/blob/0bd6976fea7fa2a07c47672de424194716d31164/lib/descriptionlessDisables.js#L16-L19

Is there a better way?

Mouvedia: > Is there a better way?

If there's a way to know the direct parent, yes there is.
In short, create a function `isRule`.

Mouvedia:
@ybiquitous is
https://github.com/stylelint/stylelint/blob/060310cfd6383175ee06fe4d15aee27a4e202671/lib/__tests__/fixtures/plugin-async.js#L10
a problem?

ybiquitous:
Ah, `possible: [true, null]` is less confusing. Can you fix it also, please?

lib/utils/validateOptions.js
ybiquitous: Nice idea 👍🏼

lib/utils/validateOptions.js
```diff
@@ -66,6 +66,12 @@ function validate(opts, ruleName, complain) {
 	const actual = opts.actual;
 	const optional = opts.optional;
 
+	if (actual === false && !ruleName.startsWith('report')) {
```
ybiquitous, line 69: [suggest] General rule names usually contain at least one hyphen (`-`),

https://github.com/stylelint/stylelint/blob/060310cfd6383175ee06fe4d15aee27a4e202671/docs/user-guide/rules/about.md?plain=1#L54-L57

so how about also checking if the characters are included for accuracy?

```suggestion
    // Ignoring options like `reportNeedlessDisables`
	if (actual === false && !ruleName.startsWith('report') && ruleName.includes("-")) {
```

(IMO, an auxiliary comment will help us understand this code.)

Mouvedia:
> Ah, `possible: [true, null]` is less confusing. Can you fix it also, please?

Nop: `Type 'boolean' is not assignable to type 'RuleOptionsPossible[]'.`

lib/utils/validateOptions.js
Mouvedia: > `&& ruleName.includes("-")`

That's wont work for `indentation` and `linebreaks`.

lib/rules/selector-no-qualifying-type/index.js
```diff
@@ -63,7 +63,7 @@ const rule = (primary, secondaryOptions) => {
 			ruleName,
 			{
 				actual: primary,
-				possible: [true, false],
+				possible: [true],
```
ybiquitous, line 66: Good catch! This is a very minor bug. 😅

ybiquitous (approved):
Thank you. LGTM! 👍🏼

ybiquitous:
Changelog entry added:

> - Added: invalid rule primary option message for `false` ([#6250](https://github.com/stylelint/stylelint/pull/6250)).

PR #6255, opened 2022-08-11
docs: add an 's' because there are several patterns

regseb:

Last two block code of [`keyframe-selector-notation`](https://stylelint.io/user-guide/rules/list/keyframe-selector-notation/#percentage-unless-within-keyword-only-block) are red, but there isn't error. I think it's because the _pattern_ is missing an _s_ (and it doesn't respect the pattern to put in green).


JounQin (approved):

ybiquitous (approved):
@regseb Thank you! 👍🏼

PR #6257, opened 2022-08-12
Add test cases for `selector-max-*` end positions

ybiquitous:

Part of #5694

Some cases for nested selectors have inaccurate end columns. This problem will be resolved with #6234.


jeddy3 (approved):
LGTM, thanks!

PR #6262, opened 2022-08-13
Fix `selector-pseudo-class-allowed-list` end positions and message

ybiquitous:
> Which issue, if any, is this issue related to?

Part of #5694

> Is there anything in the PR that needs further explanation?

This PR also fixes the message in addition to end positions; I think including the `:` character for pseudo-classes is more helpful. E.g.

```diff
-Unexpected pseudo-class "focus"
+Unexpected pseudo-class ":focus"
```

PR #6263, opened 2022-08-13
Fix `selector-pseudo-class-disallowed-list` end positions and message

ybiquitous:

Part of #5694

This PR is similar to #6262.


jeddy3 (approved):
LGTM!

There can't be many of these end position fixes left to do, right? You've done so many recently!

ybiquitous:
> There can't be many of these end position fixes left to do, right? You've done so many recently!

Yes, there are just [four rules left](https://github.com/stylelint/stylelint/issues/5694#issue-1045592380). I'll clean them up in a few days. 😁 💪🏼 

Changelog entry added:

> - Fixed: `selector-pseudo-class-disallowed-list` end positions and message ([#6263](https://github.com/stylelint/stylelint/pull/6263)).

PR #6264, opened 2022-08-13
Fix `plugins.md` document and `createPlugin` type definition

### LINKED ISSUE #6245: Fix URL metadata for plugins

I  reproduced the question in many Plugins Libraries, such as `stylelint-z-index-value-constraint`、`stylelint-scss` and so on.

Next, take `stylelint-scss` as an example：

step1: 
`npm i stylelint-scss` in my project;

step2: 
creat an test file called `test.less`;

fill it with：(from `stylelint-scss/ selector-no-redundant-nesting-selector`'s testFile)
`
called test.less
`
step3: 
`.stylelintrc.js`

`
module.exports = {
  plugins: [
    "stylelint-scss",
  ],
  rules: {
    "scss/selector-no-redundant-nesting-selector": true,
  },
};
`
step4:
it give me error tips: 'Unnecessary nesting selector (&) (scss/selector-no-redundant-nesting-selector)Stylelint(scss/selector-no-redundant-nesting-selector)'

But,  `Stylelint(scss/selector-no-redundant-nesting-selector)` had no url.

module.exports = {
  plugins: [
    "stylelint-scss",
  ],
  rules: {
    "scss/selector-no-redundant-nesting-selector": true,
  },
};

cli

 "stylelint": "^14.9.1","stylelint-scss": "^4.3.0"

I click the error tips "...Stylelint(scss/selector-no-redundant-nesting-selector)"

and I can open the web URL which was set in the library.

### What actually happened?

I clicked and I can‘t open the web Url which was set in the library.

_No response_

### Proposal to fix the bug

_No response_

ybiquitous:

Closes #6245

This PR mainly fixes the incorrect examples on the [plugin document](https://stylelint.io/developer-guide/plugins). Also, this improves the type definition of the `stylelint.createPlugin()` API, although it may be called *refactorings*.

The incorrect part on the doc is here: ⬇️ 

https://github.com/stylelint/stylelint/blob/e424be5eca77c5532b13147c4a7dad14085a4276/docs/developer-guide/plugins.md?plain=1#L53-L55

Because `stylelint.createPlugin()` returns an object `{ ruleName: string, rule: Function }`,

https://github.com/stylelint/stylelint/blob/4772851c0a1745db03f149ac35a2de6fe375c369/lib/createPlugin.js#L10-L15

the following code using `module.exports` produces an incorrect object `{ ruleName: string, rule: Function, messages: Object, meta: Object }`.

```js
module.exports.ruleName = ruleName;
module.exports.messages = messages;
module.exports.meta = meta;
```

To be correct, the `rule` function should have the `ruleName`, `messages`, and `meta` properties as below. The current example on the doc is confusing.

```js
const ruleFunc = () => { /* .. */ };

ruleFunc.ruleName = ruleName;
ruleFunc.messages = messages;
ruleFunc.meta = meta;
```

For example, the `stylelint-scss` plugin [sets the properties correctly](https://github.com/stylelint-scss/stylelint-scss/blob/v4.3.0/src/rules/at-each-key-value-single-line/index.js#L75-L77).

### DIFF
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -26,33 +26,32 @@ const meta = {
   // deprecated: true,
 };
 
-module.exports = stylelint.createPlugin(
-  ruleName,
-  (primaryOption, secondaryOptionObject) => {
-    return (postcssRoot, postcssResult) => {
-      const validOptions = stylelint.utils.validateOptions(
-        postcssResult,
-        ruleName,
-        {
-          /* .. */
-        }
-      );
-
-      if (!validOptions) {
-        return;
-      }
-
-      // ... some logic ...
-      stylelint.utils.report({
+const ruleFunction = (primaryOption, secondaryOptionObject) => {
+  return (postcssRoot, postcssResult) => {
+    const validOptions = stylelint.utils.validateOptions(
+      postcssResult,
+      ruleName,
+      {
         /* .. */
-      });
-    };
-  }
-);
+      }
+    );
 
-module.exports.ruleName = ruleName;
-module.exports.messages = messages;
-module.exports.meta = meta;
+    if (!validOptions) {
+      return;
+    }
+
+    // ... some logic ...
+    stylelint.utils.report({
+      /* .. */
+    });
+  };
+};
+
+ruleFunction.ruleName = ruleName;
+ruleFunction.messages = messages;
+ruleFunction.meta = meta;
+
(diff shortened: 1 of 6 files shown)


jeddy3 (approved):
This all LGTM, thank you!

ybiquitous:
@jeddy3 Thanks for the review!

Changelog entry added:

> - Fixed: plugins document and `createPlugin` type definition ([#6264](https://github.com/stylelint/stylelint/pull/6264)).

PR #6270, opened 2022-08-16
Fix `selector-pseudo-element-{allowed,disallowed}-list` end positions and message

ybiquitous:

Part of #5694

In addition to end positions, this PR fixes the messages to include `::` characters for clarity. E.g.

```diff
- Unexpected pseudo-element "MY-OTHER-pseudo-element"
+ Unexpected pseudo-element "::MY-OTHER-pseudo-element"
```


lib/rules/selector-pseudo-element-allowed-list/index.js
```diff
@@ -47,7 +47,7 @@ const rule = (primary) => {
 					const value = pseudoNode.value;
 
 					// Ignore pseudo-classes
-					if (value[1] !== ':') {
+					if (value.charAt(1) !== ':') {
```
ybiquitous, line 50: [note] I think the [`charAt()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charAt) method is more readable since we can easily see that its method receiver is a string.

jeddy3 (approved):
LGTM

ybiquitous:
Changelog entries added:

> - Fixed: `selector-pseudo-element-allowed-list` end positions and message ([#6270](https://github.com/stylelint/stylelint/pull/6270)).
> - Fixed: `selector-pseudo-element-disallowed-list` end positions and message ([#6270](https://github.com/stylelint/stylelint/pull/6270)).

PR #6272, opened 2022-08-16
Fix `shorthand-property-no-redundant-values` message

ybiquitous:

Part of #5694

This PR fixes the message to use double quotes for consistency with other rules:

```diff
- Unexpected longhand value '1px 1px' instead of '1px'
+ Unexpected longhand value "1px 1px" instead of "1px"
```

This also adds test cases for end positions (#5694).

### DIFF
--- a/lib/rules/shorthand-property-no-redundant-values/__tests__/index.js
+++ b/lib/rules/shorthand-property-no-redundant-values/__tests__/index.js
@@ -115,6 +115,8 @@ testRule({
 			message: messages.rejected('1px 1px', '1px'),
 			line: 1,
 			column: 5,
+			endLine: 1,
+			endColumn: 21,
 		},
 		{
 			code: 'a { margin: 1Px 1pX; }',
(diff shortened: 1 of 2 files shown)


lib/rules/shorthand-property-no-redundant-values/__tests__/index.js
```diff
@@ -115,6 +115,8 @@ testRule({
 			message: messages.rejected('1px 1px', '1px'),
 			line: 1,
 			column: 5,
+			endLine: 1,
+			endColumn: 21,
```
ybiquitous, line 119: [note] I don't think it's needed to change the arguments passed to `report()` since it should be helpful that the column range covers a declaration property and value.

https://github.com/stylelint/stylelint/blob/26f4c353079694584e469db7ca432bcc9a494fe3/lib/rules/shorthand-property-no-redundant-values/index.js#L170-L175

```css
a { margin: 1px 1px }
/*  ^^^^^^^^^^^^^^^
    Unexpected longhand value "1px 1px" instead of "1px"
*/
```

jeddy3 (approved):
Thanks!

ybiquitous:
Changelog entry added:

> - Fixed: `shorthand-property-no-redundant-values` message ([#6272](https://github.com/stylelint/stylelint/pull/6272)).

PR #6275, opened 2022-08-17
Add `ignoreAfterCombinators: []` to `selector-max-universal`

### LINKED ISSUE #5792: Add ignoreAfterCombinators: [] to ignore to selector-max-universal

I guess that selector-max-universal is performance related rule.
I.e. it is supposed to forbid patterns that hit perfomance.

Consider the following patterns (with selector-max-universal = 1):
1. "*" involves O(N) search. So, it is Good.
2. "* *" involves O(N^2) search. So, it is Bad.
3. "* + *" involves O(N) search ("+ *" = next element = it does not require additional search). So, it is also Good and should not be reported.

Add option Not to count "+ *"
because it does not require additional search and thus does not hit performance.

fpetrakov:

Closes #5792

No, it's self-explanatory, I guess.

### DIFF
--- a/lib/rules/selector-max-universal/README.md
+++ b/lib/rules/selector-max-universal/README.md
@@ -77,7 +77,7 @@ The following patterns are _not_ considered problems:
 
 ## Optional secondary options
 
-### `ignoreAfterCombinators: ["array", "of", "combinators"]`
+### `ignoreAfterCombinators: [" ", ">", "*", "~"]`
 
 Ignore universal selectors that come after one of the specified combinators.
 
--- a/lib/rules/selector-max-universal/README.md
+++ b/lib/rules/selector-max-universal/README.md
@@ -74,3 +74,35 @@ The following patterns are _not_ considered problems:
 /* `*` is inside `:not()`, so it is evaluated separately */
 * > * .foo:not(*) {}
 ```
+
+## Optional secondary options
+
+### `ignoreAfterCombinators: [">", "+", "~"]`
+
+Ignore universal selector if it's preceded by specified after combinators.
+
+Given:
+
+```json
+[1] // maximum universal selectors allowed
+[">", "+"]
+```
+
+The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+*.foo * {}
+```
+
+While the following patters are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+*.foo + * {}
+```
+
+<!-- prettier-ignore -->
+```css
+*.foo > * {}
+```
--- a/lib/rules/selector-max-universal/README.md
+++ b/lib/rules/selector-max-universal/README.md
@@ -77,7 +77,7 @@ The following patterns are _not_ considered problems:
 
 ## Optional secondary options
 
-### `ignoreAfterCombinators: ["array "of "combinators"]`
+### `ignoreAfterCombinators: ["array "of" "combinators"]`
 
 Ignore universal selector if it's preceded by specified after combinators.
 
--- a/lib/rules/selector-max-universal/README.md
+++ b/lib/rules/selector-max-universal/README.md
@@ -77,7 +77,7 @@ The following patterns are _not_ considered problems:
 
 ## Optional secondary options
 
-### `ignoreAfterCombinators: ["array "of" "combinators"]`
+### `ignoreAfterCombinators: ["array", "of", "combinators"]`
 
 Ignore universal selector if it's preceded by specified after combinators.
 
--- a/lib/rules/selector-max-universal/README.md
+++ b/lib/rules/selector-max-universal/README.md
@@ -77,7 +77,7 @@ The following patterns are _not_ considered problems:
 
 ## Optional secondary options
 
-### `ignoreAfterCombinators: [" ", ">", "*", "~"]`
+### `ignoreAfterCombinators: ["array", "of", "combinators"]`
 
 Ignore universal selectors that come after one of the specified combinators.
 
--- a/lib/rules/selector-max-universal/README.md
+++ b/lib/rules/selector-max-universal/README.md
@@ -79,29 +79,19 @@ The following patterns are _not_ considered problems:
 
 ### `ignoreAfterCombinators: ["array", "of", "combinators"]`
 
-Ignore universal selector if it's preceded by specified after combinators.
+Ignore universal selectors that come after one of the specified combinators.
 
 Given:
 
 ```json
-[1, [">", "+"]]
+[">", "+"]
 ```
 
-The following patterns are considered problems:
+For example, with `2`.
+
+The following pattern is _not_ considered a problem:
 
 <!-- prettier-ignore -->
 ```css
-*.foo * {}
-```
-
-The following patters are _not_ considered problems:
-
-<!-- prettier-ignore -->
-```css
-*.foo + * {}
-```
-
-<!-- prettier-ignore -->
-```css
-*.foo > * {}
+* * > * {}
 ```
--- a/lib/rules/selector-max-universal/README.md
+++ b/lib/rules/selector-max-universal/README.md
@@ -77,15 +77,14 @@ The following patterns are _not_ considered problems:
 
 ## Optional secondary options
 
-### `ignoreAfterCombinators: [">", "+", "~"]`
+### `ignoreAfterCombinators: ["array "of "combinators"]`
 
 Ignore universal selector if it's preceded by specified after combinators.
 
 Given:
 
 ```json
-[1] // maximum universal selectors allowed
-[">", "+"]
+[1, [">", "+"]]
 ```
 
 The following patterns are considered problems:
@@ -95,7 +94,7 @@ The following patterns are considered problems:
 *.foo * {}
 ```
 
-While the following patters are _not_ considered problems:
+The following patters are _not_ considered problems:
 
 <!-- prettier-ignore -->
 ```css
--- a/lib/rules/selector-max-universal/__tests__/index.js
+++ b/lib/rules/selector-max-universal/__tests__/index.js
@@ -296,11 +296,14 @@ testRule({
 
 testRule({
 	ruleName,
-	config: [0, { ignoreAfterCombinators: ['~', '+', '>'] }],
+	config: [0, { ignoreAfterCombinators: ['~', '+', '>', ' '] }],
 
 	accept: [
 		{
 			code: '.foo {}',
+		},
+		{
+			code: '.foo * {}',
 		},
 		{
 			code: '.foo ~ * {}',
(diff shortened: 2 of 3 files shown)


lib/rules/selector-max-universal/README.md
```diff
@@ -74,3 +74,35 @@ The following patterns are _not_ considered problems:
...
+```
+
+The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+*.foo * {}
+```
+
+While the following patters are _not_ considered problems:
```
ybiquitous, line 98: [suggest] Please keep consistency:

```suggestion
The following patters are _not_ considered problems:
```

Because the following script in our website repo would not work...

https://github.com/stylelint/stylelint.io/blob/4e562026b6c5aa0aa61ad0c437dba8881a936e08/scripts/generate-docs.mjs#L62-L63

ybiquitous (changes requested):
@fpetrakov Thanks for creating the PR! Please consider addressing my reviews.

lib/rules/selector-max-universal/README.md
```diff
@@ -74,3 +74,35 @@ The following patterns are _not_ considered problems:
 /* `*` is inside `:not()`, so it is evaluated separately */
 * > * .foo:not(*) {}
 ```
+
+
+### `ignoreAfterCombinators: [">", "+", "~"]`
```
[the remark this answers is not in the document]
fpetrakov: Sounds good to me because we give more flexibility to users this way

lib/rules/selector-max-universal/README.md
fpetrakov: We should replace it [here](https://github.com/stylelint/stylelint/blob/main/lib/rules/selector-max-id/README.md) too.
image

ybiquitous: Ooops. The following area should be green... 😓 

image

Could you please open a PR to fix the doc?

https://github.com/stylelint/stylelint/blob/2c2a57cdca138aa6afe225b90b3fa6d2110245fc/lib/rules/selector-max-id/README.md?plain=1#L97

fpetrakov: > Could you please open a PR to fix the doc?

Will be glad to help!

fpetrakov: Here we go #6276

fpetrakov:
@ybiquitous I guess, I've added support for user defined combinators. What do you think?

lib/rules/selector-max-universal/__tests__/index.js
```diff
@@ -228,6 +269,93 @@ testRule({
...
+			column: 1,
+			endLine: 1,
+			endColumn: 8,
+		},
+	],
+});
+
+testRule({
+	ruleName,
+	config: [0, { ignoreAfterCombinators: ['~', '+', '>', 'h1', 'article'] }],
```
ybiquitous, line 324: [suggest] The `h1` and `article` cases may be unnecessary since they don't match the `ignoreAfterCombinators` name and may cause confusion. I think `'~', '+', '>'` is sufficient.

```suggestion
	config: [0, { ignoreAfterCombinators: ['~', '+', '>'] }],
```

lib/rules/selector-max-universal/index.js
```diff
@@ -39,14 +52,28 @@ const rule = (primary) => {
...
 					checkSelector(childNode, ruleNode);
 				}
 
-				if (childNode.type === 'universal') total += 1;
+				let prevChildNode = nodes[index - 1];
+				let prevChildNodeValue = prevChildNode && prevChildNode.value;
+
+				// checks if previous child node is descendant combinator
+				if (prevChildNodeValue === ' ') {
+					prevChildNode = nodes[index - 2];
```
ybiquitous, line 67: [suggest]

- What about using the `.prev()` method? It can make `nodes` and `index` unnecessary.
- What about using `.type === 'combinator'` to check a descendant combinator? The code intention becomes clearer.

```suggestion
				let prevChildNode = childNode.prev();
				let prevChildNodeValue = prevChildNode && prevChildNode.value;

				// checks if previous child node is descendant combinator
				if (prevChildNode && prevChildNode.type === 'combinator' && prevChildNodeValue === ' ') {
					prevChildNode = prevChildNode.prev();
```

ybiquitous (approved):
@fpetrakov Thanks. I've left more refactoring suggestions, but it's good enough. LGTM 👍🏼 

@jeddy3 I would appreciate it if you would confirm the PR, just in case.

ybiquitous (approved):
👍🏼

lib/rules/selector-max-universal/README.md
```diff
@@ -74,3 +74,34 @@ The following patterns are _not_ considered problems:
...
+The following patters are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+*.foo + * {}
+```
+
+<!-- prettier-ignore -->
+```css
+*.foo > * {}
```
jeddy3, line 106:
```suggestion
The following pattern is _not_ considered a problem:

<!-- prettier-ignore -->
```css
* * > * {}
```

We generally only show in our docs how `ignore` options make the rule more permissive.

lib/rules/selector-max-universal/__tests__/index.js
```diff
@@ -228,6 +269,87 @@ testRule({
...
+		{
+			code: '* {}',
+			message: messages.expected('*', 0),
+			line: 1,
+			column: 1,
+			endLine: 1,
+			endColumn: 2,
+		},
+	],
+});
```
jeddy3, line 295: Let's merge this into the other `config: 0` testRule below.

lib/rules/selector-max-universal/__tests__/index.js
```diff
@@ -228,6 +269,87 @@ testRule({
 	],
 });
 
+testRule({
+	ruleName,
+	config: [0, { ignoreAfterCombinators: ['+'] }],
+
+	accept: [
+		{
+			code: 'foo {}',
```
jeddy3, line 278:
```suggestion
			code: 'a {}',
```

lib/rules/selector-max-universal/__tests__/index.js
```diff
@@ -228,6 +269,87 @@ testRule({
...
+		},
+		{
+			code: 'foo + * {}',
```
jeddy3, line 281:
```suggestion
			code: 'a + * {}',
```

lib/rules/selector-max-universal/index.js
```diff
@@ -46,7 +59,21 @@ const rule = (primary) => {
...
-				if (childNode.type === 'universal') total += 1;
+				let prevChildNode = childNode.prev();
+				let prevChildNodeValue = prevChildNode && prevChildNode.value;
+
+				// checks if previous child node is descendant combinator
+				if (prevChildNode && prevChildNode.type === 'combinator' && prevChildNodeValue === ' ') {
+					prevChildNode = prevChildNode.prev();
+
+					prevChildNodeValue = prevChildNode && prevChildNode.value;
+				}
```
jeddy3, line 70: What purpose does this bit of code have?

jeddy3 (changes requested):
@fpetrakov Thanks for the pull request. It's looking good.

I've requested a few changes to the tests and docs, and asked one question about the rule code.

lib/rules/selector-max-universal/index.js
fpetrakov: > What purpose does this bit of code have?

As you know, there are 4 types of combinators, according to [MDN](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Selectors/Combinators). This code's purpose is to add support for descendant combinator. This way user can add smth like this to the configuration: `["h1"]`.

For example, we have code like this: `a > * { color: red }`. With combinators `">", "+", "~"` it's very easy to check, in this case we did it this way: 
`let prevChildNode = childNode.prev();`
`let prevChildNodeValue = prevChildNode && prevChildNode.value;`
`prevChildNodeValue` will equal to `">"`, `"+"` or `"~"`, and then we just pass the value to optionMatches function:
`optionsMatches(secondaryOptions, 'ignoreAfterCombinators', prevChildNodeValue)`

If we have code `h1 * { color: red }` and we want to ignore `*` after `h1`, we will stumble upon descendant combinator, which is represented by a single space (" ") character. Inn this case, the code from above will not work as `prevChildNoveValue` will equal to `" "`. We have to go back even farther and check previous node of the prevChildNode to get the value `h1`.

jeddy3: Thanks for the explanation. I think we can remove this code.

> This code's purpose is to add support for the descendant combinator.

I believe the rule supports the descendant combinator without this code, via the configuration of `[" "]`. We could probably do more to accommodate other whitespace descendant combinators (like tabs and newlines), but let's only do that if someone requests it. By supporting `">"`, `"+"`, `"~"` and `" "` (space) we satisfy the original use case in https://github.com/stylelint/stylelint/issues/5792.

> If we have code `h1 * { color: red }` and we want to ignore `*` after `h1`

We can introduce a separate `ignoreAfterSelector: []` secondary option to support this as `h1` is a [type selector](https://drafts.csswg.org/selectors/#type-selectors), not a combinator (and `["h1"]` shouldn't be a valid configuration for the `ignoreAfterCombinators` option). Let's only add this option if someone requests it, though.

jeddy3 (changes requested):
@fpetrakov Thanks for making the changes and for the explanation.

I've requested one last change, otherwise it LGTM.

lib/rules/selector-max-universal/README.md
```diff
@@ -74,3 +74,24 @@ The following patterns are _not_ considered problems:
 /* `*` is inside `:not()`, so it is evaluated separately */
 * > * .foo:not(*) {}
 ```
+
+
+### `ignoreAfterCombinators: [" ", ">", "*", "~"]`
```
jeddy3, line 80:
```suggestion
### `ignoreAfterCombinators: ["array", "of", "combinators"]`
```

We use this format to describe [options that accept user-defined lists of things to ignore](https://stylelint.io/user-guide/rules/about#secondary) rather than list out the possible values as:

- there be many values, e.g. for an option like `ignoreUnits: []`
- more of the things being targeted may be added to the language, e.g. new combinators, like the shadow-piercing descendant combinator (`>>>`)

jeddy3 (changes requested):
Thanks for making the changes, including adding the test to show that the descendant combinator value works as a configuration value.

One last revert request, otherwise LGTM.

jeddy3 (approved):
@fpetrakov Thanks again for making the changes.

LGTM.

@ybiquitous Feel free to merge this as part of the upcoming release and trial of the change-set workflow.

ybiquitous:
@jeddy3 No problems. The code and doc are much more straightforward than before! 👍🏼 

@fpetrakov Thanks again for your contribution! 👏🏼

ybiquitous:
Changelog entry added:

> - Added: `ignoreAfterCombinators: []` to `selector-max-universal` ([#6275](https://github.com/stylelint/stylelint/pull/6275)).

PR #6276, opened 2022-08-17
Fix selector-max-id rule docs so it's parsed properly

fpetrakov:

No, it's self-explanatory.


ybiquitous (approved):
Thank you!

PR #6278, opened 2022-08-17
Add test cases for end positions

ybiquitous:

Ref: https://github.com/stylelint/stylelint/issues/5694#issuecomment-1218079814

This change adds test cases for end positions of the following rules:

- `declaration-property-max-values`
- `no-irregular-whitespace`


lib/rules/declaration-property-max-values/__tests__/index.js
```diff
@@ -51,37 +51,49 @@ testRule({
...
+			endLine: 1,
+			endColumn: 21,
 		},
 		{
 			code: 'a { margin-inline: 1px 2px; }',
 			message: messages.rejected('margin-inline', 1),
 			line: 1,
 			column: 5,
+			endLine: 1,
+			endColumn: 28,
```
ybiquitous, line 63: [note] I don't think it's needed to change the existing behavior:

```css
a { margin-inline: 1px 2px; }
/*  ^^^^^^^^^^^^^^^^^^^^^^^
    Expected "margin-inline" to have no more than 1 value
*/
```

jeddy3 (approved):
LGTM, thank you!

@ybiquitous Amazing job finishing all these end position tasks! ❤️

PR #6292, opened 2022-08-21
Refactor to use `isAtRule` util

### LINKED ISSUE #6291: Refactor to use `isAtRule` typeguard util
There many places in rules where instead of using existing util `isAtRule` we use `if (node.type === 'atrule')`. I suggest replace all `node.type === 'atrule'` with `isAtRule` because it makes code more readable and consistent.

fpetrakov:

Closes #6291

No, it's self-explanatory

### DIFF
--- a/lib/rules/string-quotes/index.js
+++ b/lib/rules/string-quotes/index.js
@@ -9,6 +9,7 @@ const ruleMessages = require('../../utils/ruleMessages');
 const validateOptions = require('../../utils/validateOptions');
 const valueParser = require('postcss-value-parser');
 const { isBoolean, assertString } = require('../../utils/validateTypes');
+const { isAtRule } = require('../../utils/typeGuards');
 
 const ruleName = 'string-quotes';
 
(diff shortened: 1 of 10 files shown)


ybiquitous (approved):
@fpetrakov Thanks. LGTM 👍🏼

jeddy3 (approved):
Thank you!

lib/rules/string-quotes/index.js
```diff
@@ -187,7 +188,7 @@ const rule = (primary, secondaryOptions, context) => {
 				return;
 			}
 
-			if (node.type === 'atrule' && node.name === 'charset') {
+			if (isAtRule(node) && node.name === 'charset') {
 				// allow @charset rules to have double quotes, in spite of the configuration
 				// TODO: @charset should always use double-quotes, see https://github.com/stylelint/stylelint/issues/2788
```
Mouvedia, line 193: seems like you can remove the "TODO: " prefix

ybiquitous: PR welcome!

PR #6296, opened 2022-08-22
Add autofix to declaration-block-no-duplicate-properties

### LINKED ISSUE #6119: Add autofix to declaration-block-no-duplicate-properties

I have set the Stylelint config to remove duplicate CSS properties and values as below. 

```
        "declaration-block-no-duplicate-properties": [
            true,
            {
                "ignore": ["consecutive-duplicates-with-different-values"]
            }
        ],
```
I want to remove the duplicate property values like on the fix lint command or on Ctrl+S but this is not happening at all.
Example as below.

```
.full-screen {
    align-items: center;
    align-items: center;

    //you need to setup file-loader in webpack before you can use images
    background-position: center;
    bottom: 0;
    display: flex;
    display: flex;
    flex-direction: column;

    //_ works with row or column_

    flex-direction: column;
    height: 100%;
    justify-content: center;
    justify-content: center;
    left: 0;
    position: fixed;
    right: 0;
    text-align: center;
    top: 0;
    width: 100%;
}
```

Please let me know how I can achieve this?the 

I want to remove the duplicate property values like on the fix lint command and using Ctrl+S

fpetrakov:

Closes #6119.

No, it's self-explanatory.

### DIFF
--- a/lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
+++ b/lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
@@ -215,7 +215,71 @@ testRule({
 
 testRule({
 	ruleName,
+	config: [true, { ignore: ['consecutive-duplicates'] }],
+	fix: true,
+
+	accept: [
+		{
+			code: 'p { font-size: 16px; font-size: 1rem; }',
+		},
+		{
+			code: 'p { display: inline-block; font-size: 16px; font-size: 1rem; }',
+		},
+		{
+			code: 'p { font-size: 16px; font-size: 1rem; color: red; }',
+		},
+		{
+			code: 'p { display: inline-block; font-size: 16px; font-size: 1rem; color: red; }',
+		},
+	],
+
+	reject: [
+		{
+			code: 'p { font-size: 16px; font-weight: 400; font-size: 1rem; }',
+			fixed: 'p { font-size: 16px; font-weight: 400; }',
+			message: messages.rejected('font-size'),
+		},
+		{
+			code: 'p { display: inline-block; font-size: 16px; font-weight: 400; font-size: 1rem; }',
+			fixed: 'p { display: inline-block; font-size: 16px; font-weight: 400; }',
+			message: messages.rejected('font-size'),
+		},
+		{
+			code: 'p { font-size: 16px; font-weight: 400; font-size: 1rem; color: red; }',
+			fixed: 'p { font-size: 16px; font-weight: 400; color: red; }',
+			message: messages.rejected('font-size'),
+		},
+		{
+			code: 'p { display: inline-block; font-size: 16px; font-weight: 400; font-size: 1rem; color: red; }',
+			fixed: 'p { display: inline-block; font-size: 16px; font-weight: 400; color: red; }',
+			message: messages.rejected('font-size'),
+		},
+	],
+});
+
+testRule({
+	ruleName,
 	config: [true, { ignore: ['consecutive-duplicates-with-different-values'] }],
+	fix: false,
+
+	accept: [{ code: 'p { font-size: 16px; font-size: 1rem; font-weight: 400; }' }],
+
+	reject: [
+		{
+			code: 'p { font-size: 16px; font-weight: 400; font-size: 1rem; }',
+			message: messages.rejected('font-size'),
+		},
+		{
+			code: 'p { font-size: 16px; font-size: 16px; font-weight: 400; }',
+			message: messages.rejected('font-size'),
+		},
+	],
+});
+
+testRule({
+	ruleName,
+	config: [true, { ignore: ['consecutive-duplicates-with-different-values'] }],
+	fix: true,
 
 	accept: [{ code: 'p { font-size: 16px; font-size: 1rem; font-weight: 400; }' }],
 
--- a/lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
+++ b/lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
@@ -5,6 +5,7 @@ const { messages, ruleName } = require('..');
 testRule({
 	ruleName,
 	config: [true],
+	fix: true,
 
 	accept: [
 		{
@@ -48,6 +49,7 @@ testRule({
 	reject: [
 		{
 			code: 'a { color: pink; color: orange }',
+			fixed: 'a { color: orange }',
 			message: messages.rejected('color'),
 			line: 1,
 			column: 18,
@@ -56,6 +58,7 @@ testRule({
 		},
 		{
 			code: 'a { cOlOr: pink; color: orange }',
+			fixed: 'a { color: orange }',
 			message: messages.rejected('color'),
 			line: 1,
 			column: 18,
@@ -64,6 +67,7 @@ testRule({
 		},
 		{
 			code: 'a { color: pink; cOlOr: orange }',
+			fixed: 'a { cOlOr: orange }',
 			message: messages.rejected('cOlOr'),
 			line: 1,
 			column: 18,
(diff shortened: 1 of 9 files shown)


lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
```diff
@@ -231,6 +349,25 @@ testRule({
 	],
 });
 
+testRule({
+	ruleName,
+	config: [true, { ignore: ['consecutive-duplicates-with-different-values'] }],
+	fix: true,
+
```
Mouvedia, line 356: Can you add an example for accept?
i.e. `display: inline-block; display: inline;` should remain `display: inline-block; display: inline;` after fix

lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
```diff
@@ -174,6 +174,95 @@ testRule({
 	],
 });
 
+testRule({
+	ruleName,
+	config: [true],
+	fix: true,
```
ybiquitous, line 180: [suggest] Let's merge new test cases with existing ones like this:

```diff
 testRule({
 	ruleName,
 	config: [true],
+	fix: true,
```

ybiquitous (changes requested):
@fpetrakov Thanks for creating the PR! 👍🏼 

Can you please address the following suggestions?

- add `fixable: true` to the metadata like this:
  https://github.com/stylelint/stylelint/blob/b20e672dece29e50d2508be406c6f249fee9cfe1/lib/rules/alpha-value-notation/index.js#L23
- add a description to the rule readme like this:
  https://github.com/stylelint/stylelint/blob/b20e672dece29e50d2508be406c6f249fee9cfe1/lib/rules/alpha-value-notation/README.md?plain=1#L12

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -58,9 +58,11 @@ const rule = (primary, secondaryOptions) => {
...
 		eachDeclarationBlock(root, (eachDecl) => {
-			/** @type {string[]} */
+			/** @type {import('postcss').Declaration[]} */
 			const decls = [];
 			/** @type {string[]} */
+			const declsProps = [];
+			/** @type {string[]} */
 			const values = [];
 
 			eachDecl((decl) => {
```
ybiquitous, line 68: [suggest] It seems possible to replace a new variable `decls` with the third argument of `eachDecl()` like this:

```suggestion
			eachDecl((decl, _index, decls) => {
```

ybiquitous: Ah, the new variable `decls` can be a smaller array than the third argument `decls`.
Sorry, keep the current code as-is.

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -85,12 +87,18 @@ const rule = (primary, secondaryOptions) => {
 					return;
 				}
 
-				const indexDuplicate = decls.indexOf(prop.toLowerCase());
+				const indexDuplicate = declsProps.indexOf(prop.toLowerCase());
```
ybiquitous, line 90: [suggest] I notice we can do the following refactorings:

- Avoid repetition of `prop.toLowerCase()`
- Make `declsProps` and `values` variables unneeded by `decls.findIndex()`

For instance,

```js
const lowerProp = prop.toLowerCase();

// ...

const indexDuplicate = decls.findIndex(d => d.prop.toLowerCase() === lowerProp);

// ...

const duplicateDecl = decls[indexDuplicate];
const duplicateValue = duplicateDecl ? duplicateDecl.value : '';
```

lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
fpetrakov: > [suggest] Let's merge new test cases with existing ones like this:

You mean adding `fix: true` to the first `testRule` and replacing all rejected tests within it with fixed versions from the second `testRule`?

ybiquitous: I mean adding `fix: true` to each `testRule`. Then, all rejected test cases should have a `fixed` property, and the diff in test cases should be fewer.

lib/rules/declaration-block-no-duplicate-properties/index.js
fpetrakov: Also thought of just using decls. Great suggestion!

fpetrakov:
Hmm, now all tests are failing, let's see...

ybiquitous:
It seems `npx jest -u` (updating snapshots) should resolve the failures.

lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
```diff
@@ -360,6 +398,7 @@ testRule({
 	ruleName,
 	config: [true],
 	customSyntax: 'postcss-css-in-js',
+	fix: true,
```
ybiquitous, line 401: Ah, sorry. The cases for `postcss-css-in-js` has been skipped, so, please revert the change in this `testRule`?

lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
```diff
@@ -377,3 +415,30 @@ testRule({
...
+				"import styled from 'styled-components';\nexport default styled.div` color: orange; `;",
+			message: messages.rejected('color'),
+		},
+		{
+			code: "import styled from 'react-emotion'\nexport default styled.div` color: pink; color: orange; `;",
+			fixed: "import styled from 'react-emotion'\nexport default styled.div` color: orange; `;",
+			message: messages.rejected('color'),
+		},
+	],
+});
```
ybiquitous, line 444: [suggest] This code block using `testRule` is unnecessary because `skip: true` is set. Can you please remove it?

```suggestion
```

fpetrakov: Why keeping tests that have `skip: true` btw?

ybiquitous: We've been discussing deprecating the `postcss-css-in-js` package (see https://github.com/stylelint/postcss-css-in-js/issues/225).

Of course, we can remove the skipped cases (including other rules) soon, but it seems reasonable to do it with one PR.

```console
$ git grep postcss-css-in-js lib/rules | wc -l
      28
```

Also, we can do it after we determine to deprecate the package and remove it entirely from our codebase. It seems unnecessary to me to hurry.

ybiquitous (approved):
@fpetrakov Thank you! LGTM 👍🏼

fpetrakov:
@ybiquitous should we merge it or call jeddy3 to review changes?

ybiquitous:
Let's wait for a few more additional reviews. I may have missed something.

jeddy3 (approved):
@fpetrakov LGTM, thank you. 

@ybiquitous Thanks for the initial review and suggestions!

> should we merge it or call jeddy3 to review changes?

We usually use our judgement to determine how many reviews a pull request needs. Although, it's usually one for simple things and, as @ybiquitous suggested, two when adding new features like options or rules.

JounQin:
If we going to merge, don't forget to add the changesets, but it's already a new iteration.

ybiquitous:
@JounQin Thanks for the reminder. I've pushed fd25dfb. No problems?

JounQin:
> @JounQin Thanks for the reminder. I've pushed [fd25dfb](https://github.com/stylelint/stylelint/commit/fd25dfb8555a96d903d2f56e295d994e2349c77c). No problems?

No problem.

PR #6302, opened 2022-08-24
Added null-ls option for Neovim users

benfrain:
Wasn't obvious for Neovim users that the easiest option is using null-ls as this has built-in support for Stylelint. Added that to the list. 👍

> Is there anything in the PR that needs further explanation

No, it's self-explanatory.


jeddy3 (approved):
Thank you!

PR #6305, opened 2022-08-27
Fix typo in annotation-no-unknown document

srzsrni:

I think this typo is causing a problem in stylelint.io page of the `annotation-no-unknown` rule.
The color of box-shadow in the code block  is red (invalid-pattern color) even though the pattern is not considered a problem.

https://stylelint.io/user-guide/rules/list/annotation-no-unknown/

### DIFF
--- a/lib/rules/annotation-no-unknown/README.md
+++ b/lib/rules/annotation-no-unknown/README.md
@@ -24,7 +24,7 @@ a {
 }
 ```
 
-The following pattern is _not_ considered problem:
+The following pattern is _not_ considered a problem:
 
 <!-- prettier-ignore -->
 ```css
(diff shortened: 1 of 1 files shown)


JounQin (approved):

srzsrni:
I noticed another typo, so I added another commit.

ybiquitous (approved):
@srzsrni Thank you. LGTM 👍🏼 

> I think this typo is causing a problem in stylelint.io page of the `annotation-no-unknown` rule.

That's right. The problem should be fixed after the next release.

srzsrni:
@ybiquitous @JounQin 
Thank you for reviewing this pull request!

Issue #5170, opened 2021-02-25
Allow plugins to bail out of custom message and do it themselfs with substitutions


AndyOGo:
> What is the problem you're trying to solve?

My plugin provides a custom `message` with substituted interpolations.
Unfortunately they are overruled by [stylelint's `message` option](https://stylelint.io/user-guide/configure#message) introduced with https://github.com/stylelint/stylelint/pull/672

Support for substitutions has never been added and the request for it got closed #836

Seem that it is ignored because of this [general solution](https://github.com/stylelint/stylelint/pull/672/files#diff-78f1c80ffb2836008dd194b3b0ca28f9b46e4897b606f0b3d25a29e57a8d3e61R74), so I deleted the `customMessages` entry for my rule, like:

```js
delete result.stylelint.customMessages[ruleName];
```

> What solution would you like to see?

Well, some official way of saying, hey stylelint this plugin is handling the `message` option by itself.

Original issue in my plugin: https://github.com/AndyOGo/stylelint-declaration-strict-value/issues/142

jeddy3:
@AndyOGo Thanks for the request.

Labelling as a discussion for now. Ideally, we would provide a common way for using interpolated values in custom messages that all plugins can use, but in the meantime, an official flag may by the best option.

> Support for substitutions has never been added and the request for it got closed #836

It picked up again in https://github.com/stylelint/stylelint/issues/4117, but the issue has stalled.

SalimBensiali:
I would love the ability to supplement the default message with extra info and recommendation. We have RTL aware Sass variables that we want developers to use instead of `margin-left`, `border-left-color` etc. And a custom message currently lacks the ability to preserve the existing message. Something with string interpolation or an `additional_message` option would be grand.

I would love to be able to do the following for example
```json
{
  "rules": {
    "property-disallowed-list": [
      "/left|right$/",
      {
        "message": "${original_message}. Use the \"$start-direction\" & \"$end-direction\" SASS variables instead of \"left\" & \"right\".",
      }
    ]
  }
}
```

or 

```json
{
  "rules": {
    "property-disallowed-list": [
      "/left|right$/",
      {
        "additional_message": "Use the \"$start-direction\" & \"$end-direction\" SASS variables instead of \"left\" & \"right\".",
      }
    ]
  }
}
```

SalimBensiali:
@jeddy3 Is there anything I can help with on this one? I would be keen to contribute, or is this needing internal discussion first?

Thanks.

[...]

PR #6312, opened 2022-08-30
Add experimental support for custom message arguments

### LINKED ISSUE #4117: Add support for functional custom messages

I would like to inform developers better when their code fails linting.

Currently, I can provide a custom message but:

1. This loses the ability to reference the _value_ that caused the failure. For example:
   * Before: `Unexpected hex color "#fff"`
   * After `Unexpected hex color. Please refer to the brand colours in the Styleguide`
<br><br>
2. The blacklisted values are a great way to prevent developers from adding rules you would otherwise not want. However, it's not possible to do this:
    ```
    'declaration-property-value-blacklist': {
       'font-weight': ['500', { message: 'Please use the semi-bold mixin instead' }]
    },
    ```
    ...you can only add 1 catch-all message for _all_ invalid property values.
<br><br>

1.  It would be great to be able to swap-in values `${hex}`
     or, at least prefix/suffix the original error to the custom message.

2. It would be nice if the custom messages were acknowledged for each invalid property

ybiquitous:

Closes #4117

~~**This PR is still a PoC.**~~ I appreciate any comments or suggestions.

Summary:
- `stylelint.utils.report()` can receive a `message` function and an array of `messageArgs`, in addition to just a `message` string
- ~~Only JS config file support (e.g. `.stylelintrc.js`)~~
- For JSON (or YAML), the `message` property can receive parameters like `%s` for string interpolation. The Node.js [`util.format()`](https://nodejs.org/api/util.html#utilformatformat-args) API is internally used.

Considerations:
- ~~Should we support other formats like JSON, YAML, etc.? If doing so, another method than a message function might be necessary; for example, some template syntax like `message: "a={a}, b={b}"`.~~
- ~~It's necessary to change `report()` call in all the rules in this PR. This would be a tedious job. Is there a better API instead of adding `messageArgs`?~~
  - Now, we can proceed gradually to add `messageArgs` to each rule implementation.

Example:

`.stylelintrc.js`:

```js
module.exports = {
  rules: {
    'custom-property-pattern': [
      '^([a-z][a-z0-9]*)(-[a-z0-9]+)*$',
      {
        message: (pattern) => `Expected custom property name to be kebab-case; pattern=/${pattern}/`,
      },
    ],
  },
};
```

Or `.stylelintrc.json`:

```json
{
  "message": "Expected custom property name to be kebab-case; pattern=/%s/"
}
```

`a.css`:

```css
a {
  --CustomProperty: 1;
}
```

Run:

```console
$ stylelint a.css

a.css
 2:3  ✖  Expected custom property name to be kebab-case; pattern=/^([a-z][a-z0-9]*)(-[a-z0-9]+)*$/  custom-property-pattern

```


Mouvedia:
> Only JS config file support (e.g. .stylelintrc.js)

I am using `.stylelintrc` in my projects; will it be supported?

> Should we support other formats like JSON, YAML, etc.?

The minimum for the PoC would be JSON IMO.

> If doing so, another method than a message function might be necessary

What about
```
message: '(pattern) => `Expected custom property name to be kebab-case, but actual "${pattern}"`'
```
?
i.e. using the result of calling toString on the function

ybiquitous:
@Mouvedia Thanks for the feedback. JSON support is still essential.

Perhaps, using `eval` may cause some security problems. So we need to get a better idea, not using a just JS function.

Mouvedia:
> Perhaps, using `eval` may cause some security problems. So we need to get a better idea, not using a just JS function.

```
message: [
	'pattern',
    'return `Expected custom property name to be kebab-case, but actual "${pattern}"`'
]
```
i.e. the last element is always the body (exactly like `Function`)

for the reversal:
```js
Function('pattern', 'return `Expected custom property name to be kebab-case, but actual "${pattern}"`')
```
i.e. you pass all the elements in the same order

Mouvedia:
> when using Content-Security-Policy

Is stylelint being run in browser context?

ybiquitous:
Not now, but it may be realized in the future (see #4796).

ybiquitous:
Here is an example of a template syntax (in JSON), using `{}` characters for interpolation. The idea specifying an array is inspired by @Mouvedia:

```json
{
  "message": [
    "pattern", // message arguments
    "Expected custom property name to be kebab-case, but actual \"{pattern}\"" // message template
  ]
}
```

ybiquitous:
Examples of interpolation characters:

- `{value}` 
- `{{value}}`
- `#{value}`
- `%{value}`
- etc.

ybiquitous:
Oh, I meant to indicate the candidates in <https://github.com/stylelint/stylelint/pull/6312#issuecomment-1235517495>.

Also, I think it's better to avoid `${value}` because it's easy to confuse with JS template literals. See the ESLint [`no-template-curly-in-string`](https://eslint.org/docs/latest/rules/no-template-curly-in-string) rule and the following example:

```js
{
  message: [
    'pattern',
    'Expected custom property name to be kebab-case, but actual "${pattern}"',
                                                              /* ^^^^^^^^^^ not template literal */
  ]
}
```

The idea of `0 arguments`, `1 argument` and `n arguments` in <https://github.com/stylelint/stylelint/pull/6312#issuecomment-1235527085> sounds good. 👍🏼

ybiquitous:
I find a more straightforward solution; to use the notation (`%s`) of [`util.format()`](https://nodejs.org/api/util.html#utilformatformat-args). Here is an example:

```json
{
  "message": "Expected custom property name to be kebab-case, but actual \"%s\""
}
```

`%s` is also used by [`console.log`](https://developer.mozilla.org/en-US/docs/Web/API/Console#using_string_substitutions).

Mouvedia:
> `%s` is also used by [`console.log`](https://developer.mozilla.org/en-US/docs/Web/API/Console#using_string_substitutions).

For the PoC, yes. We should add the support of `%d` and `%f` in the backlog.
String support will be enough for now.

ybiquitous:
The eddbb4b commit adds JSON support using `util.format()`, and I've updated the PR description. It is now ready to review.

ybiquitous:
### To-dos

- [ ] Update the usage document
- [ ] Add `messageArgs` to each rule implementation (but it's not required in this PR)

Mouvedia:
related: #5170

jeddy3 (commented):
This is fantastic work!

> Update the usage document in this PR

Let's classify this as an experimental feature ([like we did for `--fix](https://stylelint.io/CHANGELOG#7110)) so that we can treat any issues that may emerge as non-breaking. It's possible this will happen when we add support to more rules or plugin authors start to make use of the feature.

As with [exposing our reference data](https://github.com/stylelint/stylelint/issues/6151), we should only add support to a rule when a user requests it as we'll need to discuss and lockdown the message format.

I suggest we add support to [`color-no-hex`](https://stylelint.io/user-guide/rules/list/color-no-hex), rather than `custom-property-pattern`, in this pull request as it was one of the rules [originally requested in the issue](https://github.com/stylelint/stylelint/issues/4117#issue-456799155).

Let's then add the following to the `color-no-hex` rule README:

"The [`message` secondary option](https://stylelint.io/user-guide/configure#message) can accept the arguments of this rule."

It'll be the last item in the _extended description_, i.e. below the following if it's present:

"The [`fix` option](https://stylelint.io/user-guide/usage/options#fix) can automatically fix all of the problems reported by this rule.".

We can also extend the description of the [`message` option](https://stylelint.io/user-guide/configure#message) to say something along the lines of:

"Experimental feature: some rules support message arguments. For example, when configuring the `color-no-hex` rule the hex color can be used in `message` string:

`.stylelintrc.js`:

```js
{
  'color-no-hex': [true, {
    message: (hex) => `Don't use hex colors like "${hex}"`,
  }]
}
```

`.stylelintrc.json`:

```json
{
  "color-no-hex": [true, {
    "message": "Don't use hex colors like '${%s/'"
  }]
}
```

Mouvedia:
> rather than custom-property-pattern

He probably wanted the first rule to have a REGEX.
What about extracting the support of message for the specific rules to different PRs?

ybiquitous:
@jeddy3 Thanks for the feedback. Your suggestion sounds good, so I'll implement it.

By the way, I chose `custom-property-pattern` because the rule is used in `scripts/visual-config.json`, and it's easy to try the new behavior in the `scripts/` directory. No problem to start with any rules to me. 👍🏼

docs/user-guide/configure.md
```diff
@@ -109,6 +109,31 @@ For example, the following rule configuration would substitute in custom message
...
+<!-- prettier-ignore -->
+```json
+{
+  "color-no-hex": [true, {
+    "message": "Don't use hex colors like \"%s\""
+  }]
+}
+```
+
+With formats that don't support a function like JSON, you can use a `printf`-like format (e.g., `%s`).
```
ybiquitous, line 135: [note] I think a description of `%s` is necessary. Please let me know if you have a better expression.

See also the [`util.format`](https://nodejs.org/api/util.html#utilformatformat-args) on the Node.js documentation.

docs/user-guide/configure.md
```diff
@@ -109,6 +109,31 @@ For example, the following rule configuration would substitute in custom message
...
+{
+  'color-no-hex': [true, {
+    message: (hex) => `Don't use hex colors like "${hex}"`,
+  }]
+}
+```
+
+`.stylelintrc.json`:
+
+<!-- prettier-ignore -->
```
ybiquitous, line 126: [note] Unless the disabling comment, Prettier would format the snippet like this:

```json
{
  "color-no-hex": [
    true,
    {
      "message": "Don't use hex colors like \"%s\""
    }
  ]
}
```

Compared to the snippet above for `.stylelintrc.js`, it would not be easier to read.

ybiquitous:
@jeddy3 I've done updating the doc. I'd appreciate your review when you have time.

lib/rules/color-no-hex/index.js
```diff
@@ -42,7 +42,8 @@ const rule = (primary) => {
 				const endIndex = index + node.value.length;
 
 				report({
-					message: messages.rejected(node.value),
+					message: messages.rejected,
+					messageArgs: [node.value],
```
Mouvedia, line 46: remark

If I understand correctly, the order will be arbitrary.
e.g. if you don't use `--fix` but you want to use the fixed _value_ in the message

i.e.
- we would generate the value but not replace with it in the file
- the user will need to know that the bad value is the first element and the good value is the second element

jeddy3 (approved):
LGTM, thank you!

lib/rules/color-no-hex/index.js
ybiquitous: @Mouvedia Thanks for the share. Other problems might be found than you pointed out, but this feature is still experimental. We would try considering it based on reactions or feedback later.

PR #6315, opened 2022-09-02
Improve `function-name-case` document

ybiquitous:
- Use dashed idents in examples
- Remove incorrect example

Part of #5523

- Use dashed idents in examples
- Remove incorrect example

### DIFF
--- a/lib/rules/function-name-case/README.md
+++ b/lib/rules/function-name-case/README.md
@@ -115,7 +115,7 @@ a {
 
 ## Optional secondary options
 
-### `ignoreFunctions: ["/regex-as-string/", /regex/, "non-regex"]`
+### `ignoreFunctions: ["/regex/", /regex/, "non-regex"]`
 
 Ignore case of function names.
 
@@ -124,7 +124,7 @@ For example, with `"lower"`.
 Given:
 
 ```json
-["some-function", "/^get.*$/"]
+["--some-function", "/^--get.*$/"]
 ```
 
 The following patterns are considered problems:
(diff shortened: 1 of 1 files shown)


lib/rules/function-name-case/README.md
```diff
@@ -115,7 +115,7 @@ a {
 
 ## Optional secondary options
 
-### `ignoreFunctions: ["/regex-as-string/", /regex/, "non-regex"]`
+### `ignoreFunctions: ["/regex/", /regex/, "non-regex"]`
```
ybiquitous, line 118: [note] `regex-as-string` rarely used.

lib/rules/function-name-case/README.md
```diff
@@ -124,36 +124,29 @@ For example, with `"lower"`.
...
 a {
-  color: sOmE-FuNcTiOn();
+  color: --sOmE-FuNcTiOn();
 }
 ```
 
 <!-- prettier-ignore -->
 ```css
 a {
-  color: some-other-function();
```
ybiquitous, line 142: [note] `some-other-function` is not a problem.

lib/rules/function-name-case/README.md
```diff
@@ -124,36 +124,29 @@ For example, with `"lower"`.
 Given:
 
 ```json
-["some-function", "/^get.*$/"]
+["--some-function", "/^--get.*$/"]
 ```
 
 The following patterns are considered problems:
```
ybiquitous, line 130: [ask] Do we need to give examples of considered problems in `ignoreFunctions`?

jeddy3: Narh, we don't need to. We can just show the rule being made more permissive. For example, by showing a pattern that was a problem previously without the option being used:

e.g. before the secondary options header we can show `--FOO()` being a problem, and then:

Given:

```
["--FOO"]
```

It no longer being an problem. 

I think we can just take the problem section out in this pull request, and save the suggestions above (showing the permissiveness and using `foo`) for another issue/pr.

ybiquitous: Thanks for the explanation. I understand and agree. 👍🏼
Then, I'll keep the PR change as-is.

jeddy3 (approved):

PR #6316, opened 2022-09-02
Add regex support for `ignorePseudoClasses` option of `selector-pseudo-class-no-unknown`

ybiquitous:

Part of #5523

While working on #5523, I notice the `selector-pseudo-class-no-unknown` rule doesn't support a real regex.

### DIFF
--- a/.changeset/big-crabs-tap.md
+++ b/.changeset/big-crabs-tap.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: regex support for `ignorePseudoClasses` option of `selector-pseudo-class-no-unknown`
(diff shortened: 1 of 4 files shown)


.changeset/big-crabs-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: regex support for `ignorePseudoClasses` option of `selector-pseudo-class-no-unknown`
```
ybiquitous, line 5: [ask] Is it no problem to consider this PR a *minor* change? Or a bug fix?

jeddy3: It's not clear-cut. I think minor is fine.

jeddy3 (approved):
LGTM

PR #6317, opened 2022-09-02
Add regex support for `ignorePseudoElements` option of `selector-pseudo-element-no-unknown`

ybiquitous:

Part of #5523

Similar reason with #6316.


jeddy3 (approved):
LGTM, thank you.

PR #6318, opened 2022-09-02
Fix false positives for container queries in `*-no-unknown`

### LINKED ISSUE #6304: Fix false positives for container queries in `*-no-unknown`

CSS Container Queries will be shipped on Chrome 105 soon. See [Can I use](https://caniuse.com/css-container-query-units):

image

However, the `at-rule-no-unknown` rule doesn't support the new at-rule `@container` from the spec. See the [demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIACkB2AXAhgS1TATgAQAUAtrgLQDu2U6AFvIQOwAMrADgB4CUwAOqkKEAdGEz4ohAUOGEo2AM4cANpgCeTAOb5aAbkFzCO2hXQxSqzOYqQVAV1KpFTAEwAzIgEZPB2QF9BQNQQABoQRXUMTC4EcEVFMPAIVHdsLTiZYX4QfHsVGEUcpiy5HOsKPIKKVAgKe1QAa1qqVGLCdDyYQ0Jg-xB-IA):

image

Ref:

- [CSS Containment Module Level 3](https://drafts.csswg.org/css-contain-3/)
- [CSS Container Queries - CSS: Cascading Style Sheets | MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries)
- [@container and :has(): two powerful new responsive APIs landing in Chromium 105 - Chrome Developers](https://developer.chrome.com/en/blog/has-with-cq-m105/)

It seems to me that we can support the new at-rule by adding `container` to the at-rule list. See the related code:

- https://github.com/stylelint/stylelint/blob/cde3a14ebf8e28f4a21e9ebc0c5e1a98bbffcd0f/lib/rules/at-rule-no-unknown/index.js#L54
- https://github.com/stylelint/stylelint/blob/cde3a14ebf8e28f4a21e9ebc0c5e1a98bbffcd0f/lib/reference/atKeywords.js#L26

fpetrakov:

Closes #6304.

No, it's self-explanatory.

### DIFF
--- a/lib/reference/atKeywords.js
+++ b/lib/reference/atKeywords.js
@@ -28,6 +28,7 @@ const atKeywords = uniteSets(pageMarginAtKeywords, [
 	'apply',
 	'character-variant',
 	'charset',
+	'container',
 	'counter-style',
 	'custom-media',
 	'custom-selector',
@@ -48,7 +49,6 @@ const atKeywords = uniteSets(pageMarginAtKeywords, [
 	'supports',
 	'swash',
 	'viewport',
-	'container',
 ]);
 
 module.exports = {
--- a/lib/reference/atKeywords.js
+++ b/lib/reference/atKeywords.js
@@ -48,6 +48,7 @@ const atKeywords = uniteSets(pageMarginAtKeywords, [
 	'supports',
 	'swash',
 	'viewport',
+	'container',
 ]);
 
 module.exports = {
--- a/lib/reference/units.js
+++ b/lib/reference/units.js
@@ -10,12 +10,6 @@ const lengthUnits = new Set([
 	'rem',
 	'rlh',
 	'lh',
-	'cqw',
-	'cqh',
-	'cqi',
-	'cqb',
-	'cqmin',
-	'cqmax',
 	// Viewport-percentage lengths
 	'dvh',
 	'dvmax',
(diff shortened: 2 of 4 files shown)


Mouvedia:
You gotta add cqw, cqh, cqi, cqb, cqmin, and cqmax to https://github.com/stylelint/stylelint/blob/12f1141d2435a2a52a64764cf5910eab561c2d38/lib/reference/units.js#L6

jeddy3 (approved):
LGTM, thank you!

lib/reference/atKeywords.js
```diff
@@ -48,6 +48,7 @@ const atKeywords = uniteSets(pageMarginAtKeywords, [
 	'supports',
 	'swash',
 	'viewport',
+	'container',
```
ybiquitous, line 51: [suggest] Can you please insert `container` to the position in alphabetical order (`atKeywords`)?

lib/reference/units.js
```diff
@@ -10,6 +10,12 @@ const lengthUnits = new Set([
 	'rem',
 	'rlh',
 	'lh',
+	'cqw',
+	'cqh',
+	'cqi',
+	'cqb',
+	'cqmin',
+	'cqmax',
```
ybiquitous, line 18: [suggest] What about adding a new section for the new units like this?

```diff
 // Flexible length units
 'fr',
+// Container query units
+'cqw',
+'cqh',
+'cqi',
+'cqb',
+'cqmin',
+'cqmax',
```

ybiquitous (commented):
@fpetrakov Thanks for creating the PR!

Almost good, but I've suggested a few refactorings. Please consider addressing them.

ybiquitous (approved):
Thank you! LGTM 👍🏼

PR #6319, opened 2022-09-04
Fix incorrect location for matching violating times in `time-min-milliseconds`

### LINKED ISSUE #4751: Fix incorrect location for matching violating times in time-min-milliseconds

kawaguchi1102:

Closes #4751 

Fixed the following additions

- Fix Test case end positions


lib/rules/time-min-milliseconds/index.js
```diff
@@ -62,47 +62,44 @@ const rule = (primary, secondaryOptions) => {
...
-						// Check all time values
-						for (const value of valueList) {
-							if (!isAcceptableTime(value)) {
-								complain(decl, propertyValue.indexOf(value), value.length);
-							}
-						}
-					}
-				}
+				const parsedValue = valueParser(getDeclarationValue(decl));
+				let isValueTimeCount = 0;
```
ybiquitous, line 66: [suggest] The `is*` prefix often indicates a boolean variable and so the `isValueTimeCount` name seems a bit confusing to me. How about using `timeValueCount` instead?

```suggestion
				let timeValueCount = 0;
```

lib/rules/time-min-milliseconds/index.js
```diff
@@ -62,47 +62,44 @@ const rule = (primary, secondaryOptions) => {
...
+
+				parsedValue.walk(({ value, sourceIndex }) => {
+					isValueTimeCount = valueTimeCheck(value, isValueTimeCount);
```
ybiquitous, line 69: [suggest] To make the purpose of the `valueTimeCheck` function clearer, how about renaming it to `calcTimeValueCount`?

```suggestion
					timeValueCount = calcTimeValueCount(value, timeValueCount);
```

lib/rules/time-min-milliseconds/index.js
```diff
@@ -62,47 +62,44 @@ const rule = (primary, secondaryOptions) => {
...
+
+					if (isAcceptableTime(value) || (ignoreDelay && isValueTimeCount !== 1)) return;
```
ybiquitous, line 71: [suggest] I'm a bit concerned about performance degradation due to extra calls of `valueParse.unit()` inside `valueTimeCheck()` and `isAcceptableTime()`. How about passing a `import('postcss-value-parser').Dimension` object instead of a value string like this?

```js
const dimension = valueParser.unit(value);
isValueTimeCount = valueTimeCheck(dimension, value, isValueTimeCount);

if (isAcceptableTime(dimension) || (ignoreDelay && isValueTimeCount !== 1)) return;
```

ybiquitous (commented):
@kawaguchi1102 Thanks for creating the PR. The change looks almost good to me.

I've left some refactoring suggestions, so please consider addressing them.

lib/rules/time-min-milliseconds/index.js
jeddy3: Agreed. I suggest we also use the [`getDimension()`](https://github.com/stylelint/stylelint/blob/main/lib/utils/getDimension.js) util rather than use `valueParser.unit()` directly as it's safer for non-standard syntax.

kawaguchi1102: @ybiquitous @jeddy3 
Thanks for the great advice!
I think these fixed can be done by tomorrow.

kawaguchi1102:
The refactor has been completed.
I hope I am not mistaken in this refactor.

lib/rules/time-min-milliseconds/index.js
```diff
@@ -52,57 +53,45 @@ const rule = (primary, secondaryOptions) => {
...
+				complain(decl, sourceIndex, value.length);
+			});
 		});
 
 		/**
-		 * Get the duration within an `animation` or `transition` shorthand property value.
-		 *
-		 * @param {string[]} valueList
-		 * @returns {string | undefined}
+		 * @param {import('postcss-value-parser').Dimension | {unit: null, number: null}} dimension
```
ybiquitous, line 82: [trivial] It seems we can simplify the type a bit as follows, but acceptable to keep it as-is. Perhaps, it's good to change the return type of `getDimension()`.

```suggestion
		 * @param {{unit: string | null, number: string | null}} dimension
```

lib/rules/time-min-milliseconds/index.js
```diff
@@ -118,27 +107,27 @@ const rule = (primary, secondaryOptions) => {
 		}
 
 		/**
-		 * @param {string} time
+		 * @param {import('postcss-value-parser').Dimension | {unit: null, number: null}} valueNode
 		 * @returns {boolean}
 		 */
-		function isAcceptableTime(time) {
-			const parsedTime = valueParser.unit(time);
+		function isAcceptableTime(valueNode) {
```
ybiquitous, line 113: [suggest] The name `dimension` seems more accurate:

```suggestion
		function isAcceptableTime(dimension) {
```

lib/rules/time-min-milliseconds/index.js
```diff
@@ -118,27 +107,27 @@ const rule = (primary, secondaryOptions) => {
...
 
-			const numTime = Number(parsedTime.number);
+			const numTime = Number(number);
 
 			if (numTime <= 0) {
 				return true;
 			}
 
-			const unit = parsedTime.unit.toLowerCase();
+			const _unit = unit.toLowerCase();
```
ybiquitous, line 124: [suggest] Using the `_` prefix for variables is rare in our code base. How about renaming the variable like this?

```suggestion
			const timeUnit = unit.toLowerCase();
```

ybiquitous (commented):
@kawaguchi1102 Thank you, the code is now more straightforward and readable! Almost no problem.
I've left easy refactoring suggestions, so please check them.

kawaguchi1102:
@ybiquitous Thank you for reviewing every detail!
Now we have a cleaner code!

ybiquitous (approved):
Thank you. LGTM 👍🏼

jeddy3 (approved):
LGTM, thank you!

PR #6326, opened 2022-09-06
Add regex support to `ignoreTypes` option of `selector-type-case`

ybiquitous:

None.

The `ignoreTypes` option of the `selector-type-case` rule has supported a regex-like string already. Supporting also regex expressions in JS config files should be helpful.


jeddy3 (approved):
LGTM, thank you!

PR #6328, opened 2022-09-06
Improve rule docs using dashed idents

ybiquitous:

Part of #5523 (I believe the last one)

No, it's self-explanatory.

### DIFF
--- a/lib/rules/selector-max-id/README.md
+++ b/lib/rules/selector-max-id/README.md
@@ -91,7 +91,7 @@ The following patterns are considered problems:
 
 <!-- prettier-ignore -->
 ```css
-a:matches(#foo) {}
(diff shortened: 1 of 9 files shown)


lib/rules/selector-max-id/README.md
```diff
@@ -91,7 +91,7 @@ The following patterns are considered problems:
 
 <!-- prettier-ignore -->
 ```css
-a:matches(#foo) {}
+a:is(#foo) {}
```
ybiquitous, line 94: [note] `:matches` was renamed to `:is`. See https://developer.mozilla.org/en-US/docs/Web/CSS/:is

jeddy3 (approved):
LGTM, thank you!

PR #6334, opened 2022-09-10
Add `ignore: ["inside-block"]` and `splitList` to selector-disallowed-list

### LINKED ISSUE #6294: Add `ignore: ["inside-block"]` to selector-disallowed-list
## What is the problem you're trying to solve?
I also want to disallow top-level selectors which are starting with `&` **because that's invalid CSS code**

```css
&.foo { }
```

(common to `SASS` precompiler language but is not restricted to it) 

## What solution would you like to see?
either top-level flexible regex rule or an explicit rule for anything starting specifically with `&`

mattmanuel90:

Closes #6294 

A bit of a shuffle, but I added an additional ```iterateThroughSelectorList``` flag which iterates through ruleNode.selectors rather than just the single ruleNode.selector or raw.selector.raw. The RegExp can get a bit unwieldy.

I consider the following ```.foo, .bar {}``` to be 2 separate selectors, so a dev may prefer to parse each of these individually ( ```.foo``` and ```.bar```) than to have to parse a string like ```.foo,\n.bar```

### DIFF
--- a/lib/rules/selector-disallowed-list/README.md
+++ b/lib/rules/selector-disallowed-list/README.md
@@ -69,7 +69,7 @@ a[href] {}
 
 ## Optional secondary options
 
-### `iterateThroughSelectorList: true | false` (default: `false`)
+### `iterateThroughList: true | false` (default: `false`)
 
 Iterate through each individual selector in a selector list.
 
--- a/lib/rules/selector-disallowed-list/README.md
+++ b/lib/rules/selector-disallowed-list/README.md
@@ -93,6 +93,7 @@ The following pattern is _not_ considered a problem:
 <!-- prettier-ignore -->
 ```css
 .bar .foo {}
+```
 
 ### `ignore: ["inside-block"]`
 
--- a/lib/rules/selector-disallowed-list/README.md
+++ b/lib/rules/selector-disallowed-list/README.md
@@ -71,53 +71,44 @@ a[href] {}
 
 ### `iterateThroughSelectorList: true | false` (default: `false`)
 
-This option will apply your configuration to each individual selector in a selector list where each selector is seperated by a comma separated value. ','
+Iterate through each individual selector in a selector list.
 
-For example, with 'true'.
+For example, with `true`.
 
-Given the string:
+Given:
 
 ```json
-.foo
+[".foo"]
 ```
 
-The following patterns are considered problems:
+The following pattern is considered a problem:
 
 <!-- prettier-ignore -->
 ```css
-.bar, .foo { }
+.bar, .foo {}
 ```
 
-The following patterns are _not_ considered problems:
+The following pattern is _not_ considered a problem:
 
 <!-- prettier-ignore -->
 ```css
-.bar .foo { }
+.bar .foo {}
 
 ###  `ignore: ["inside-block"]`
 
-Excludes any selectors that are inside a block. Use this if you only want to target selectors at the root level.
+Ignore selectors that are inside a block. 
 
-Given the string:
+Given:
 
 ```json
-.foo
+[".foo"]
 ```
 
-The following patterns are considered problems:
-
-<!-- prettier-ignore -->
-```css
-.foo { }
-```
-
-The following patterns are _not_ considered problems:
+The following pattern is _not_ considered a problem:
 
 <!-- prettier-ignore -->
 ```css
 .bar {
-    .foo {
-
-    }
+  .foo {}
 }
 ```
--- a/lib/rules/selector-disallowed-list/README.md
+++ b/lib/rules/selector-disallowed-list/README.md
@@ -69,7 +69,7 @@ a[href] {}
 
 ## Optional secondary options
 
-### `iterateThroughList: true | false` (default: `false`)
+### `splitList: true | false` (default: `false`)
 
 Iterate through each individual selector in a selector list.
 
@@ -94,7 +94,7 @@ The following pattern is _not_ considered a problem:
 ```css
 .bar .foo {}
 
-###  `ignore: ["inside-block"]`
+### `ignore: ["inside-block"]`
 
 Ignore selectors that are inside a block. 
 
--- a/lib/rules/selector-disallowed-list/README.md
+++ b/lib/rules/selector-disallowed-list/README.md
@@ -97,7 +97,7 @@ The following pattern is _not_ considered a problem:
 
 ### `ignore: ["inside-block"]`
 
-Ignore selectors that are inside a block. 
+Ignore selectors that are inside a block.
 
 Given:
 
--- a/lib/rules/selector-disallowed-list/README.md
+++ b/lib/rules/selector-disallowed-list/README.md
@@ -71,7 +71,7 @@ a[href] {}
 
 ### `splitList: true | false` (default: `false`)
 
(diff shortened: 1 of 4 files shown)


jeddy3 (changes requested):
@mattmanuel90 Thanks for the pull request. It's looking good.

Let's drop `Selector` from the option name to be consistent with other rules. Is `iterateThrough` the best wording, or can we think of an alternative? 

I've requested a few changes (as `suggestions` that you can commit) to the docs to be consistent with other rules too.

mattmanuel90:
> @mattmanuel90 Thanks for the pull request. It's looking good.
> 
> Let's drop `Selector` from the option name to be consistent with other rules. Is `iterateThrough` the best wording, or can we think of an alternative?
> 
> I've requested a few changes (as `suggestions` that you can commit) to the docs to be consistent with other rules too.

Thanks for the suggestions. Definitely way more concise.

Between `iterateThrough` and `iterateThroughList`, just based on feeling, I think the latter, it's adequate enough when paired with the description in the README, the former might raise an eyebrow.

Will resolve

lib/rules/selector-disallowed-list/README.md
```diff
@@ -66,3 +66,49 @@ a
...
+.bar, .foo {}
+```
+
+The following pattern is _not_ considered a problem:
+
+<!-- prettier-ignore -->
+```css
+.bar .foo {}
+
+###  `ignore: ["inside-block"]`
```
ybiquitous, line 97: There are some typos.

````suggestion
.bar .foo {}
```

### `ignore: ["inside-block"]`
````

File: lib/rules/selector-disallowed-list/index.js [hunk 2]
```diff
@@ -34,21 +47,44 @@ const rule = (primary) => {
 				return;
 			}
 
-			const { selector, raws } = ruleNode;
+			if (optionsMatches(secondaryOptions, 'ignore', 'inside-block')) {
```
ybiquitous, line 50: [suggest] We can avoid unnecessary repetition of the evaluations by moving the code out of the `walkRules` loop like this:

```js
if (!validOptions) {
	return;
}

const ignoreInsideBlock = optionsMatches(secondaryOptions, 'ignore', 'inside-block');
const iterateThroughList = secondaryOptions && secondaryOptions.iterateThroughList;

```

ybiquitous (changes requested):
@mattmanuel90 Thanks for creating the pull request. Adding the `iterateThroughList` option is a nice idea. 👍🏼 

I've left comments about the doc fix and a tiny refactoring. Please consider addressing them.

jeddy3:
> Between `iterateThrough` and `iterateThroughList`, just based on feeling, I think the latter, it's adequate enough when paired with the description in the README, the former might raise an eyebrow.

I meant is there an alternative to "iterate through list" rather than drop "list", for example:

- `splitList` - Split a selector list into individual selectors before checking
- `checkSelectorsInList` - If `true`, this rule will check individual selectors in selectors lists. 
- ...

Can we think of any others or are we happy with iterate through? We'll want to get the name right as we'll likely use the option again for other `selector-*` rules.

---

Other boolean options we already have include:

- [`checkPrefixed`](https://stylelint.io/user-guide/rules/list/property-no-unknown/#checkprefixed-true--false-default-false)
- [`resolveNestedSelectors`](https://stylelint.io/user-guide/rules/list/selector-class-pattern/#resolvenestedselectors-true--false-default-false)
- [`disallowInList`](https://stylelint.io/user-guide/rules/list/no-duplicate-selectors/#disallowinlist-true--false-default-false)
- [`avoidEscape`](https://stylelint.io/user-guide/rules/list/string-quotes/#avoidescape-truefalse-defaults-to-true)

mattmanuel90:
Ah fair point, I should've factored this in.

I’d be explicit and go with `splitSelectorList` as my first choice.

 I think it’s okay to have selectorList than list because selector list is the official term, list might be too vague for some and maybe we would want to use this option too in non `selector-*` rules in the future. 

Split is also the stronger prefix to use than check (I’d rank split > iterate > check). The split prefix is intentional about separating selectors through a list, where as check (i.e checkSelectorsInList) isnt acute enough in intention. Dev's might currently have regExp that already check for selectors in a list.

lib/rules/selector-disallowed-list/README.md
mattmanuel90: Eek , I'll do another pass through of the README

jeddy3:
Split would be my first choice too. Like [resolve](https://stylelint.io/user-guide/rules/list/selector-class-pattern/#resolvenestedselectors-true--false-default-false), it describes the action taken before [the thing](https://stylelint.io/user-guide/rules/about/#names) is checked.

> I think it’s better to have selectorList than list because selector list is the official term for it, list might be too vague for some and maybe we would want to use this option too in non selector-* rules in the future (just a what-if situation)

It's a good point. However, let's go with list for consistency with [preexisting](https://stylelint.io/user-guide/rules/list/no-descending-specificity/#ignore-selectors-within-list) [options](https://stylelint.io/user-guide/rules/list/no-duplicate-selectors/#disallowinlist-true--false-default-false). If we do add a similar option to non `selector-*` and there is ambiguity, we can use selector list there.

---

Incidentally, I now feel that [`checkPrefixed`](https://stylelint.io/user-guide/rules/list/property-no-unknown/#checkprefixed-true--false-default-false) was a mistake as it makes the rule stricter (and we generally make rules a stick as possible by default and add options to make them more permissive). The `property-no-unknown` rule should have checked prefixed properties by default and had an `ignore: ["prefixed"]` option. It's too late to change it now, but we should try to avoid adding `check*` options in the future.

lib/rules/selector-disallowed-list/index.js
mattmanuel90: Ah yes. Thanks!

mattmanuel90:
Sorry ignore the rerequest spam, i think its bugged

lib/rules/selector-disallowed-list/README.md
```diff
@@ -66,3 +66,49 @@ a
...
+<!-- prettier-ignore -->
+```css
+.bar, .foo {}
+```
+
+The following pattern is _not_ considered a problem:
+
+<!-- prettier-ignore -->
+```css
+.bar .foo {}
```
ybiquitous, line 95: [suggest] Missing closing triple-backslashes?

````suggestion
.bar .foo {}
```
````

Comment (mattmanuel90) [reply lib/rules/selector-disallowed-list/README.md:95]: Eugh. My bad. Think it looks good now. 
Screen Shot 2022-09-13 at 10 15 19 am

**Review by ybiquitous — APPROVED** [review]
@mattmanuel90 Thank you. LGTM 👍🏼

File: lib/rules/selector-disallowed-list/README.md [hunk 4]
```diff
@@ -66,3 +66,50 @@ a
 ```css
 a[href] {}
 ```
+
+
+### `splitList: true | false` (default: `false`)
+
+Iterate through each individual selector in a selector list.
```
jeddy3, line 74:
```suggestion
Split selector lists into individual selectors.
```

jeddy3 (approved):
LGTM, thank you.

PR #6335, opened 2022-09-11
Fix false positives for interpolation and shorthand in `font-family-name-quotes`

### LINKED ISSUE #4154: Fix false positives for interpolation and shorthand in font-family-name-quotes
I'm getting false positives related to a Sass interpolation in a `font` shorthand declaration value. The SCSS looks something like:

```scss
.foo {
  font: 12px/#{customFunc($some-length)} Roboto, sans-serif;
}
```

The unexpected errors are:

```
Expected quotes around "} Roboto" (font-family-name-quotes)
Expected "#{customFunc" to be "#{customfunc" (function-name-case)
```

And the related rules are configured as follows:

```json
{
  "rules": {
    "font-family-name-quotes": "always-where-recommended",
    "function-name-case": ["lower", {"ignoreFunctions": ["customFunc"]}]
  }
}
```

This is happening with stylelint 10.1.0, when running from CLI.

kimulaco:

Closes #4154

I updated to change the timing of executing `isStandardSyntaxValue` to ignore problematic cases.

However, if `customSyntax` is not set to `'postcss-scss'`, `CssSyntaxError` is raised.
I think this error is correct because this bug is SCSS syntax.

### DIFF
--- a/.changeset/famous-cobras-clean.md
+++ b/.changeset/famous-cobras-clean.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `font-family-name-quotes` false positives for interpolation and shorthand
--- a/lib/rules/font-family-name-quotes/__tests__/index.js
+++ b/lib/rules/font-family-name-quotes/__tests__/index.js
@@ -25,6 +25,13 @@ const variablePositiveTests = [
 	},
 ];
 
+const scssPositiveTests = [
+	{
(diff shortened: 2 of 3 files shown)


ybiquitous (approved):
@kimulaco Thanks, LGTM. 👍🏼

jeddy3 (approved):
@kimulaco Thanks for the pull request. LGTM.

PR #6345, opened 2022-09-12
Add support for multiple `--ignore-path` flags

### LINKED ISSUE #4680: Allow multiple files in the --ignore-path flag

My project contains a `.gitignore` file that I'd like to use as the basis for my stylelint ignore rules. I can do this with `--ignore-path .gitignore`,  however I'd like to supplement this with additional ignore rules inside `.stylelintignore`.

I would like to be able to specify more than one ignore file and have the rules combined. For now I think that I can get around this by using a combination of the `--ignore-path .gitignore` flag plus the `ignoreFiles` config option, but I've been lead to believe that the latter is not an efficient way to ignore files.

Would it be possible to allow more than one file in the `--ignore-path` flag? (Or multiple uses of the `--ignore-path` flag to create an array)

kimulaco:

Closes #4680

No, it's self-explanatory.
Please let me know if I'm missing a unit test.

### DIFF
--- a/lib/utils/getFileIgnorer.js
+++ b/lib/utils/getFileIgnorer.js
@@ -10,14 +10,18 @@ const isPathNotFoundError = require('./isPathNotFoundError');
 const DEFAULT_IGNORE_FILENAME = '.stylelintignore';
 
 /**
- * @param {{ cwd: string, ignorePath?: string[], ignorePattern?: string[] }} options
+ * @param {{ cwd: string, ignorePath?: string | string[], ignorePattern?: string[] }} options
  * @return {import('ignore').Ignore}
  */
 module.exports = function getFileIgnorer(options) {
 	const ignorer = ignore();
-	const ignorePath = options.ignorePath || [];
+	let ignorePath = [];
 
-	if (ignorePath.length <= 0) {
+	if (Array.isArray(options.ignorePath) && options.ignorePath.length > 0) {
+		ignorePath = options.ignorePath;
+	} else if (typeof options.ignorePath === 'string') {
+		ignorePath.push(options.ignorePath);
+	} else {
 		ignorePath.push(DEFAULT_IGNORE_FILENAME);
 	}
 
--- a/lib/utils/getFileIgnorer.js
+++ b/lib/utils/getFileIgnorer.js
@@ -15,17 +15,13 @@ const DEFAULT_IGNORE_FILENAME = '.stylelintignore';
  */
 module.exports = function getFileIgnorer(options) {
 	const ignorer = ignore();
-	let ignorePath = [];
+	const ignorePaths = [options.ignorePath || []].flat();
 
-	if (Array.isArray(options.ignorePath) && options.ignorePath.length > 0) {
-		ignorePath = options.ignorePath;
-	} else if (typeof options.ignorePath === 'string') {
-		ignorePath.push(options.ignorePath);
-	} else {
-		ignorePath.push(DEFAULT_IGNORE_FILENAME);
+	if (ignorePaths.length === 0) {
+		ignorePaths.push(DEFAULT_IGNORE_FILENAME);
 	}
 
-	for (const ignoreFilePath of ignorePath) {
+	for (const ignoreFilePath of ignorePaths) {
 		const absoluteIgnoreFilePath = path.isAbsolute(ignoreFilePath)
 			? ignoreFilePath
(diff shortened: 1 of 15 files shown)


types/stylelint/index.d.ts
```diff
@@ -223,7 +223,7 @@ declare module 'stylelint' {
 			 */
 			cwd?: string;
 			ignoreDisables?: boolean;
-			ignorePath?: string;
+			ignorePath?: string[];
```
ybiquitous, line 226: [suggest] Since the `LinterOptions` type is a part of the public API, some plugins or API users' codes may break. Is it possible to keep backward compatibility like this?

```suggestion
			ignorePath?: string | string[];
```

ybiquitous (commented):
@kimulaco Thanks for creating the pull request. I've left a comment about backward compatibility, so please confirm it.

types/stylelint/index.d.ts
kimulaco: I fixed to allow string ignorePath. https://github.com/stylelint/stylelint/pull/6345/commits/450fe87f41a8bdd28eb1a0d60ba3064599881ca4

lib/utils/getFileIgnorer.js
```diff
@@ -10,25 +10,38 @@ const isPathNotFoundError = require('./isPathNotFoundError');
...
+
+	if (Array.isArray(options.ignorePath) && options.ignorePath.length > 0) {
+		ignorePath = options.ignorePath;
+	} else if (typeof options.ignorePath === 'string') {
+		ignorePath.push(options.ignorePath);
+	} else {
+		ignorePath.push(DEFAULT_IGNORE_FILENAME);
+	}
+
+	for (const ignoreFilePath of ignorePath) {
```
ybiquitous, line 28: [suggest] How about refactoring the code be using [`Array.prototype.flat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) like this?

```suggestion
	const ignorePaths = [options.ignorePath || []].flat();

	if (ignorePaths.length === 0) {
		ignorePaths.push(DEFAULT_IGNORE_FILENAME);
	}

	for (const ignoreFilePath of ignorePaths) {
```

ybiquitous (commented):
Thanks. Almost good. 👍🏼 

I've left one more refactoring suggestion, so can you see it?

ybiquitous:
We may need to update the CLI help and the following documents about the `--ignore-path` option.
Any thoughts, @jeddy3 ?

https://github.com/stylelint/stylelint/blob/bb102826e4ccb7c0943562b35ff6d6f2493bd141/lib/cli.js#L125-L129

https://github.com/stylelint/stylelint/blob/bb102826e4ccb7c0943562b35ff6d6f2493bd141/docs/user-guide/usage/cli.md?plain=1#L59-L61

https://github.com/stylelint/stylelint/blob/bb102826e4ccb7c0943562b35ff6d6f2493bd141/docs/user-guide/usage/options.md?plain=1#L132-L136

jeddy3:
> Any thoughts, @jeddy3 ?

ESLint [does a good job of this](https://eslint.org/docs/latest/user-guide/command-line-interface#--ignore-pattern). Let's add "You can repeat the option to provide multiple paths" to those three descriptions, e.g.

"Path to a file containing patterns that describe files to ignore. The path can be absolute or relative to process.cwd(). You can repeat the option to provide multiple paths. By default, Stylelint looks for .stylelintignore in process.cwd(). "

lib/utils/getFileIgnorer.js
kimulaco: Thanks. I updated the code you suggested because it was smarter. https://github.com/stylelint/stylelint/pull/6345/commits/d1e89820a6f3aff0d35fb61f421e3a0766faaff6

kimulaco:
@jeddy3 @ybiquitous Thanks. I have updated the document. https://github.com/stylelint/stylelint/pull/6345/commits/268f2afefa482bf53115b2e7de58b09b5763fe3f https://github.com/stylelint/stylelint/pull/6345/commits/dc3db688339ffcf2733a1b20b8763968d0745174

ybiquitous (approved):
@jeddy3 Thanks, your suggestion is so good. 👍🏼 

@kimulaco Thank you for updating the code and docs. LGTM 👍🏼

jeddy3 (approved):
LGTM, thank you!

PR #6347, opened 2022-09-15
Add autofix to `font-weight-notation`

### LINKED ISSUE #3158: Add autofix to font-weight-notation

ybiquitous:

Closes #3158

I've refactored the rule implementation much to support autofix:

- check each value node by the `postcss-value-parser` module
- unify some logic between the `font` and `font-weight` properties (though a few conditional branches remain)
- remove specific logic for the `@font-face` at-rule

Here is the code outline after the refactoring:

```js
root.walkDecls(/^font(-weight)?$/i, (decl) => {

  for (const [index, valueNode] of valueNodes.entries()) {
    // checking each value node...
  }
});
```

### DIFF
--- a/.changeset/cool-zebras-sparkle.md
+++ b/.changeset/cool-zebras-sparkle.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: `font-weight-notation` autofix
--- a/docs/user-guide/rules/list.md
+++ b/docs/user-guide/rules/list.md
@@ -123,7 +123,7 @@ Within each cateogory, the rules are grouped by the [_thing_](http://apps.workfl
 
 ### Font weight
 
-- [`font-weight-notation`](../../../lib/rules/font-weight-notation/README.md): Require numeric or named (where possible) `font-weight` values. Also, when named values are expected, require only valid names.
+- [`font-weight-notation`](../../../lib/rules/font-weight-notation/README.md): Require numeric or named (where possible) `font-weight` values. Also, when named values are expected, require only valid names (Autofixable).
 
 ### Function
 
--- a/lib/reference/keywords.js
+++ b/lib/reference/keywords.js
@@ -311,6 +311,7 @@ module.exports = {
 	fontFamilyKeywords,
 	fontShorthandKeywords,
 	fontSizeKeywords,
+	fontWeightAbsoluteKeywords,
 	fontWeightKeywords,
 	fontWeightRelativeKeywords,
 	gridAreaKeywords,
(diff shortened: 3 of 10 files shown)


jeddy3 (commented):
@ybiquitous This is a fantastic refactor to implement autofix, thank you! 

I've thoughts on 3 things:

1. Let's fix our references so they're accurate:

https://github.com/stylelint/stylelint/blob/efe5283b9c5486e7323050a28074c2e1c6fae94a/lib/reference/keywords.js#L31

This set [should be](https://csstree.github.io/docs/syntax/#Property:font-weight&hltype=Type&hlname=font-weight-absolute): `["normal", "bold"]`

And then let's unite the sets in the reference file (rather than the rule itself), e.g.

```js
const fontWeightNonNumericKeywords = uniteSets(
	fontWeightRelativeKeywords,
	fontWeightAbsoluteKeywords,
);

const fontWeightKeywords = uniteSets(
	basicKeywords,
	fontWeightNonNumericKeywords,
	fontWeightNumericKeywords,
);
```

2. We generally use the format "Expected x to be y" when a rule supports autofixing of small things, e.g. "Expected '400' to be 'normal' (font-weight-notation)".

3. While we're refactoring we may want to remove the validation path, i.e.:

```js
invalidNamed: (name) => `Unexpected invalid font-weight name "${name}"`,
```

The rule should only be doing one thing.

Shall we do the latter two suggestions in this pull request or in follow-up ones?

ybiquitous:
@jeddy3 Thanks for the review! I agree with your suggestions, and I've pushed the first refactoring via 56863de28.

I'll work on the remaining suggestions on another PR to clarify the scope of this PR. 💪🏼

jeddy3 (approved):
Thanks for implementing the first suggestion.

PR LGTM. Feel free to merge.

> I'll work on the remaining suggestions on another PR to clarify the scope of this PR. 💪🏼

SGTM.

PR #6348, opened 2022-09-15
Fix incorrect URL hashes in documents

### LINKED ISSUE #6346: Fix incorrect URL hashes in documents

If the document's URL hash is camel case, the scroll position does not match.
Because the Docusaurus is to convert the id from camel case to lower case.

**Example:** 
The "More info" hash in the [Command Line Interface (CLI)](https://stylelint.io/user-guide/usage/cli) is a camel case.
https://github.com/stylelint/stylelint/blob/main/docs/user-guide/usage/cli.md?plain=1#L17

However, the actual id set is lower case.
https://stylelint.io/user-guide/usage/options/#allowemptyinput

Other pages may have the same problem.

I think it is one of the following.

- Change hash to lower case.
- Specify the id in the camel case heading.
    - e.g. ``## `allowEmptyInput` {#allowEmptyInput}``
    - https://docusaurus.io/docs/next/markdown-features/toc#heading-ids

I prefer the latter because it is more readable.

I can send a PR.

kimulaco:

Closes #6346

No, it's self-explanatory.
I replaced the docs folder in my local [stylelint.io](https://github.com/stylelint/stylelint.io) and checked the page transitions.

### DIFF
--- a/docs/developer-guide/syntaxes.md
+++ b/docs/developer-guide/syntaxes.md
@@ -1,6 +1,6 @@
 # Writing custom syntaxes
 
-Custom syntaxes are [PostCSS syntaxes](https://github.com/postcss/postcss#syntaxes) written by the community to support other styling languages and CSS-in-JS libraries using the [`customSyntax` option](../user-guide/usage/options.md#customSyntax).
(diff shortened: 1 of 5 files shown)


ybiquitous (approved):
@kimulaco Thank you! LGTM 👍🏼

Issue #6342, opened 2022-09-12
Deprecate stylistic rules


jeddy3:
We should deprecate that [stylistic rules](https://stylelint.io/user-guide/rules/list#stylistic-issues) that [we froze a year ago](https://github.com/stylelint/stylelint/issues/5674).

Even though deprecating is a non-breaking change, we should do this as part of a major release so that we can communicate this significant change in direction.

We'd release a new version of our [standard config](https://github.com/stylelint/stylelint-config-standard) that [removes the deprecated rules](https://github.com/stylelint/stylelint-config-standard/issues/244). Therefore, most of our users will never see the deprecation warnings.

As such, we won't need to prepare a plugin pack of the stylistic rules. If people want to continue using the stylistic rules, they can port the rules they need to a plugin under their own namespace and outside of the Stylelint organisation.

Ref: https://github.com/stylelint/stylelint/issues/6311#issuecomment-1241568191
Ref: https://github.com/stylelint/stylelint/issues/5674

ybiquitous:
@jeddy3 Thanks for opening the issue. I agree with the deprecation included in a major release. 👍🏼

jeddy3:
Following on from https://github.com/stylelint/stylelint/issues/6483, let's only deprecate the rules handled by Prettier as suggested by @ybiquitous in https://github.com/stylelint/stylelint/issues/6311#issuecomment-1238891275.

The 8 rules not handled by Prettier are unlikely to be added to it as they need configuration options to be safe or desired.

They are also not as problematic as some of our other stylistic rules as they:

- don't have the possibility of introducing new lint problems
- use the construct parsers rather than stylesearch
- cleanly work with the AST, either mutating raws or single nodes

We can move these rules to 3 subgroups within our "[Enforce non-stylistic conventions](https://stylelint.io/user-guide/rules#enforce-non-stylistic-conventions)" group:

- Case
- Empty lines
- Whitespace inside

We can then rename the:

- "Enforce non-stylistic conventions" group to "Enforce conventions"
- "Enforce stylistic conventions" group to "Deprecated" (before being removed entirely in v16)

I believe this approach will strike the best balance between our needs to:

- ensure the smoothest migration to v15 for the majority of our users
- move forward with plans like migrating to ESM (so that we can update a bunch of our dependencies)

ybiquitous:
@jeddy3 More balanced idea! 👍🏼

Mouvedia:
> before being removed entirely in v16

[...]

PR #6349, opened 2022-09-16
Fix end-of-source position to last line range

nlydv:
The `no-missing-end-of-source-newline` rule frequently outputs line/column positions at random spot *near* the end of a file (especially in custom syntaxes), which can be confusing when highlighted in a text editor.

This changes it to output a range over the last line of the file. There may be (or at least ought to be) more concise ways of doing that though, I’m not too familiar with PostCSS’ API.


ybiquitous:
@nlydv We're sorry that you took the time to create this PR, but we have frozen and will deprecate the `no-missing-end-of-source-newline` rule in the stylistic rules. See below for details:

- https://stylelint.io/user-guide/rules/list#stylistic-issues
- https://github.com/stylelint/stylelint/issues/6342

nlydv:
Oh, no worries. Thanks for pointing that out though.

PR #6350, opened 2022-09-18
Fix messages of `font-weight-notation`

ybiquitous:

Follow-up of https://github.com/stylelint/stylelint/pull/6347#pullrequestreview-1111412669

This PR includes the following changes:

- Use the "Expected x to be y" message format as possible
- Remove the message for invalid `font-weight` notations
  (it's possible to use the [`stylelint-csstree-validator`](https://www.npmjs.com/package/stylelint-csstree-validator) package instead)


lib/rules/font-weight-notation/index.js
```diff
@@ -17,11 +16,13 @@ const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
 const setDeclarationValue = require('../../utils/setDeclarationValue');
 const validateOptions = require('../../utils/validateOptions');
+const { assertString } = require('../../utils/validateTypes');
 
 const ruleName = 'font-weight-notation';
 
 const messages = ruleMessages(ruleName, {
 	expected: (type) => `Expected ${type} font-weight notation`,
+	expectedWithActual: (actual, expected) => `Expected "${actual}" to be "${expected}"`,
```
ybiquitous, line 25: [note] Add `expectedWithActual` for backward compatibility.

lib/rules/font-weight-notation/index.js
```diff
@@ -32,11 +33,11 @@ const meta = {
 
 const NORMAL_KEYWORD = 'normal';
 
-const KEYWORD_TO_NUMERIC = new Map([
+const NAMED_TO_NUMERIC = new Map([
 	['normal', '400'],
 	['bold', '700'],
 ]);
-const NUMERIC_TO_KEYWORD = new Map([
+const NUMERIC_TO_NAMED = new Map([
```
ybiquitous, line 40: [note] Rename for more accuracy because both `400` and `normal` are keywords.

lib/rules/font-weight-notation/index.js
```diff
@@ -124,45 +125,39 @@ const rule = (primary, secondaryOptions, context) => {
...
-				if (isNumbery(lowerWeightValue) && NUMERIC_TO_KEYWORD.has(lowerWeightValue)) {
-					if (context.fix) {
-						const keyword = NUMERIC_TO_KEYWORD.get(lowerWeightValue);
+				if (isNumbery(lowerWeightValue) && NUMERIC_TO_NAMED.has(lowerWeightValue)) {
+					const namedValue = NUMERIC_TO_NAMED.get(lowerWeightValue);
 
-						if (keyword) {
-							weightValueNode.value = keyword;
-						}
+					assertString(namedValue);
```
ybiquitous, line 152: [note] Use `assertString` for less `if` branches. `NUMERIC_TO_NAMED.has() === true` ensures the return value of ` NUMERIC_TO_NAMED.get()` is a string.

.changeset/wicked-wombats-look.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `font-weight-notation` messages
```
ybiquitous, line 5: [ask]

- This PR removes the validation path, but is it a patch change? Or, a minor change?
- "Fixed: `font-weight-notation` messages" may be inaccurate. Any better description?

jeddy3: I think "Fixed: font-weight-notation messages" is fine.

lib/rules/font-weight-notation/index.js
```diff
@@ -17,11 +16,13 @@ const report = require('../../utils/report');
...
 const setDeclarationValue = require('../../utils/setDeclarationValue');
 const validateOptions = require('../../utils/validateOptions');
+const { assertString } = require('../../utils/validateTypes');
 
 const ruleName = 'font-weight-notation';
 
 const messages = ruleMessages(ruleName, {
 	expected: (type) => `Expected ${type} font-weight notation`,
+	expectedWithActual: (actual, expected) => `Expected "${actual}" to be "${expected}"`,
 	invalidNamed: (name) => `Unexpected invalid font-weight name "${name}"`,
```
jeddy3, line 26:
```suggestion
```

Is this line redundant now?

lib/rules/font-weight-notation/index.js
jeddy3: You're right to add this.

When we first built the rule, it should have only checked `normal` and `bold` rather than relative words too. Now we have an odd mix of the `named-where-possible` and `numeric` primary options along with the `ignore: ["relative"]` secondary option.

It's too late to change it now.

Adding `expectedWithActual` is the best way to deal with backwards compatibility for things like:

https://github.com/stylelint/stylelint/blob/9abc4353de642f7274b1e87ed1211675235b9b79/lib/rules/font-weight-notation/__tests__/index.js#L175-L183

jeddy3 (changes requested):
Thanks for this!

We should also remove this line from the rule README and list of rule docs:

"Also, when named values are expected, require only valid names."

And this line from the rule README:

"Valid font-weight names are normal, bold, bolder, and lighter."

ybiquitous:
@jeddy3 Thanks for the review. I’ve addressed your suggestions.

Issue #4715, opened 2020-04-22
Fix false negatives on second run for cache and plugins


Yivan:
> Clearly describe the bug

Stylelint doesn't report warning of plugins on second run if cache activated

> What stylelint configuration is needed to reproduce the bug?

Install:
`npm install stylelint-no-unsupported-browser-features`

And add in .stylelintrc:
```
  "plugins": [
    "stylelint-no-unsupported-browser-features"
  ],
  "rules": {
    "plugin/no-unsupported-browser-features": [true, {
      "severity": "warning"
    }],
```

And active styelint `cache` to `true`

> Which version of stylelint are you using?

13.3.2

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

Node.js API

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

No

> What did you expect to happen?

On second run, see the same warning than in the first run

> What actually happened (e.g. what warnings or errors did you get)?

On the second run only warning on files having errors are reported

jeddy3:
@Yivan Thanks for the report and for using the template.

I can reproduce this locally on the CLI with:

```json
{
  "plugins": ["stylelint-no-unsupported-browser-features"],
  "rules": {
    "plugin/no-unsupported-browser-features": [
      true,
      {
        "browsers": ["ie 6-8"],
        "severity": "warning"
      }
    ]
  }
}
```

```css
a {
  display: flex;
}
```

I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

kimulaco:
I would like to work on this issue.

jeddy3:
@kimulaco That'd be fantastic, thank you!

By the way, you're welcome to start working on any issue you like without commenting. As a Stylelint member, you've permission to change the `status: ready to implement` label to `status: wip`. That'll let other people know you're working on it.

kimulaco:
@jeddy3 Thanks. I will update the label myself next time and start working!

PR #6354, opened 2022-09-19
Fix type declarations for custom message arguments

stof:
> Which issue, if any, is this issue related to?

Closes #6353

> Is there anything in the PR that needs further explanation?

No, it's self-explanatory.

PR #6356, opened 2022-09-20
Fix cache refresh when config is changed

### LINKED ISSUE #2908: Fix .stylelintcache should be refreshed when config is changed
Hello there,

A config change is not detected when using cache. It should be refreshed.

`8.1.1`

`stylelint  **/*.css --cache`

kimulaco:

Closes #2908

Currently, there is a problem with cached file messages not being displayed. This report already exists. #4715
So the test I have added is not optimal. I don't know how to prove that the file is cached in the CLI.


kimulaco:
This lint error does not occur locally.  I'm checking this error.
https://github.com/stylelint/stylelint/actions/runs/3094063323/jobs/5007077131

kimulaco:
The reason the error did not occur on my local was that [this change](https://github.com/stylelint/stylelint/pull/6354) was not merged.
I fixed type error, please review!

types/stylelint/index.d.ts
```diff
@@ -205,6 +217,8 @@ declare module 'stylelint' {
 
 		export type GetLintSourceOptions = GetPostcssOptions & {
 			existingPostcssResult?: PostCSS.Result;
+			cache?: boolean;
+			fileCache?: FileCache;
```
ybiquitous, line 221: [suggest] How about adding `fileCache` to the `InternalApi` type like this?

```diff
export type InternalApi = {
	_options: LinterOptions & { cwd: string };
	_extendExplorer: ReturnType<typeof cosmiconfig>;
	_specifiedConfigCache: Map<Config, Promise<CosmiconfigResult>>;
	_postcssResultCache: Map<string, PostCSS.Result>;
+	_fileCache: FileCache,
```

```diff
function createStylelint(options = {}) {
...
+	stylelint._fileCache = new FileCache(stylelint._options.cacheLocation, stylelint._options.cwd);
```

Because `InternalApi` has the cache properties of `_specifiedConfigCache` and `_postcssResultCache` already.

lib/utils/FileCache.js
```diff
@@ -28,6 +28,13 @@ class FileCache {
 		this._hashOfConfig = hashOfConfig;
 	}
 
+	/**
+	 * @param {string} hashOfConfig
+	 */
+	set hashOfConfig(hashOfConfig) {
+		this._hashOfConfig = hashOfConfig;
+	}
```
ybiquitous, line 36: [suggest] How about adding the following method instead of the setter method?

```js
/**
 * @param {import('stylelint').Config} config
 */
calcHashOfConfig(config) {
	const stylelintVersion = pkg.version;
	const configString = JSON.stringify(config || {});
	this._hashOfConfig = hash(`${stylelintVersion}_${configString}`);
}
```

This suggestion aims to make it easier to find the location of the calculation logic.

lib/utils/FileCache.js
```diff
@@ -28,6 +28,13 @@ class FileCache {
 		this._hashOfConfig = hashOfConfig;
 	}
```
ybiquitous, line 29: [suggest] Is it possible to simplify the `FileCache` constructor like this?

```js
class FileCache {
	/**
	 * @param {string | undefined} cacheLocation
	 * @param {string} cwd
	 */
	constructor(cacheLocation, cwd) {
```

ybiquitous (commented):
@kimulaco Thanks for creating the pull request.

I've commented on several refactoring suggestions. Please consider addressing them.

lib/lintSource.js
```diff
@@ -65,13 +67,30 @@ module.exports = async function lintSource(stylelint, options = {}) {
 
 	const config = configForFile.config;
 	const existingPostcssResult = options.existingPostcssResult;
+	const fileCache = options.fileCache;
+
+	if (options.cache && fileCache) {
+		const stylelintVersion = pkg.version;
+		const hashOfConfig = hash(`${stylelintVersion}_${JSON.stringify(config || {})}`);
```
ybiquitous, line 74: [ask] I worry we calculate the hash of a config file each time a single file is analyzedf. What do you think?

kimulaco: Thanks. It is true, this it is not a good idea to calculate each file.
I think I can improve it thanks to your suggested comments!

types/stylelint/index.d.ts
kimulaco: I moved the `fileCache` to `InternalApi`. [`ca2fb53`](https://github.com/stylelint/stylelint/pull/6356/commits/ca2fb534ba93e2dba40092945e6f56758c99a97e#diff-7ee3bec822a60d59e7dfdabf3fdf562eb48971e3edfad7687e2f5e8c998e2e89)

lib/utils/FileCache.js
kimulaco: Thanks. I think your suggestion is great. So, I added `calcHashOfConfig`. [`ca2fb53`](https://github.com/stylelint/stylelint/pull/6356/commits/ca2fb534ba93e2dba40092945e6f56758c99a97e#diff-83240954c077d79828b7a92f75b9e1c412d2b228b367c5864011f57b937fd26e)

lib/utils/FileCache.js
kimulaco: I removed `hashOfConfig` from argments of `constructor`.  [`ca2fb53`](https://github.com/stylelint/stylelint/pull/6356/commits/ca2fb534ba93e2dba40092945e6f56758c99a97e#diff-83240954c077d79828b7a92f75b9e1c412d2b228b367c5864011f57b937fd26e)

lib/lintSource.js
kimulaco: I Added a check to `fileCache` has already `hashOfConfig`. [`ca2fb53`](https://github.com/stylelint/stylelint/pull/6356/commits/ca2fb534ba93e2dba40092945e6f56758c99a97e#diff-37813c24c3fa3c6093015b61419f0bc18983cc2bbaddd56a732576bdb53f9023)

lib/utils/FileCache.js
ybiquitous: Thanks. In addition, can you please fix also the JSDoc like this?

```js

class FileCache {
	/**
	 * @param {string | undefined} cacheLocation
	 * @param {string} cwd
	 */
	constructor(cacheLocation = DEFAULT_CACHE_LOCATION, cwd) {
```

- Move the JSDoc from `class FileCache` to `constructor`
- Fix the `cacheLocation` parameter type
- Add the `cwd` parameter and make it required (removing `= process.cwd()`)

lib/__tests__/cli.test.js
```diff
@@ -12,6 +13,12 @@ const replaceBackslashes = require('../testUtils/replaceBackslashes');
 const fixturesPath = (...elems) => replaceBackslashes(path.join(__dirname, 'fixtures', ...elems));
 const { buildCLI } = cli;
 
+const removeFile = async (filePath) => {
+	if (existsSync(filePath)) {
+		await fs.unlink(filePath);
+	}
+};
```
ybiquitous, line 20: [suggest] We can now use the `testUtils/removeFile.js` helper. Can you please replace this code with the helper?

https://github.com/stylelint/stylelint/blob/005f39eea531283608afe5222e05a30c7d984d35/lib/testUtils/removeFile.js#L9

lib/lintSource.js
```diff
@@ -66,6 +66,20 @@ module.exports = async function lintSource(stylelint, options = {}) {
 	const config = configForFile.config;
 	const existingPostcssResult = options.existingPostcssResult;
 
+	if (options.cache) {
+		if (!stylelint._fileCache._hashOfConfig) {
+			stylelint._fileCache.calcHashOfConfig(config);
+		}
```
ybiquitous, line 72: [suggest] I feel it a bit smarter to avoid using the private field `_hashOfConfig`. How about checking the field in `calcHashOfConfig()` like this?

```suggestion
		stylelint._fileCache.calcHashOfConfig(config);
```

```js
calcHashOfConfig(config) {
	if (this._hashOfConfig) return; // avoid re-calculation

	// calculate hash...
```

lib/lintSource.js
```diff
@@ -66,6 +66,20 @@ module.exports = async function lintSource(stylelint, options = {}) {
...
+
+		if (options.filePath && !stylelint._fileCache.hasFileChanged(options.filePath)) {
+			return options.existingPostcssResult
+				? Object.assign(options.existingPostcssResult, {
```
ybiquitous, line 76: [suggest] The code can be shorter thanks to a local variable:

```suggestion
			return existingPostcssResult
				? Object.assign(existingPostcssResult, {
```

types/stylelint/index.d.ts
```diff
@@ -87,6 +88,16 @@ declare module 'stylelint' {
 
 		export type DisabledWarning = { line: number; rule: string };
 
+		type FileCache = {
+			_fileCache: fileEntryCache.FileEntryCache;
+			_hashOfConfig: string;
```
ybiquitous, line 93: [suggest] It seems possible to remove the private fields:

```suggestion
```

lib/__tests__/cli.test.js
```diff
@@ -369,4 +376,70 @@ describe('CLI', () => {
...
+			'--cache-location',
+			fixturesPath('.stylelintcache'),
+			'--config',
+			fixturesPath('config-block-no-empty.json'),
+			fixturesPath('empty-block.css'),
+		]);
+		expect(process.stdout.write).toHaveBeenCalledTimes(2);
+
+		await removeFile(fixturesPath('.stylelintcache'));
+	});
```
ybiquitous, line 444: [ask] I think we should test this time bugfix in the `standalone-cache.test.js` instead of `cli.test.js` because `cli.js` is not directly related to caching.

Is it possible to move the test cases from `cli.test.js` to `standalone-cache.test.js`?

kimulaco: There is already a test when the config changes in `standalone-cache.test.js`.
https://github.com/kimulaco/stylelint/blob/feat/issue-2908/lib/__tests__/standalone-cache.test.js#L92-L107

This issue was caused by the different usage of config in the API and CLI. So I thought I also should add a CLI test.

But this PR will make the use of config the same. So as you say, I move this test to `standalone-cache.test.js`.

lib/lintSource.js
kimulaco: Updated `FileCache` that return if has `_hashOfConfig`. [`b9c8a00`](https://github.com/stylelint/stylelint/pull/6356/commits/b9c8a00f7cf4abf58dafd3b0d0660703aa6c913d#diff-83240954c077d79828b7a92f75b9e1c412d2b228b367c5864011f57b937fd26e)

lib/utils/FileCache.js
kimulaco: Updated JSDoc of `FileCache`. [`b9c8a00`](https://github.com/stylelint/stylelint/pull/6356/commits/b9c8a00f7cf4abf58dafd3b0d0660703aa6c913d#diff-83240954c077d79828b7a92f75b9e1c412d2b228b367c5864011f57b937fd26e)

lib/lintSource.js
kimulaco: Thanks. Changed use local variable. [1bf3607](https://github.com/stylelint/stylelint/pull/6356/commits/1bf36073eb112b6819a42828695a1ed58576d1f2)

lib/__tests__/cli.test.js
kimulaco: Deleted cache test from `cli.test.js`. So removed `removeFile`. https://github.com/stylelint/stylelint/pull/6356/commits/fe3f4411257c48b14539480b4699f12a3f4a6a40

lib/__tests__/cli.test.js
kimulaco: I removed the cache test from `cli.test.js`. https://github.com/stylelint/stylelint/pull/6356/commits/fe3f4411257c48b14539480b4699f12a3f4a6a40
The `standalone-cache.test.js` test looked good enough. So I have not added tests.

Mouvedia:
@kimulaco thanks for tackling the oldest non-upstream bug.

lib/__tests__/standalone-cache.test.js
```diff
@@ -105,8 +109,8 @@ describe('standalone cache', () => {
 		expect(errored).toBe(true);
 
 		// Ensure only changed files are linted
-		expect(results.some((file) => file.source === validFile)).toBe(false);
-		expect(results.some((file) => file.source === newFileDest)).toBe(true);
+		expect(results.some((file) => isChanged(file, validFile))).toBe(false);
+		expect(results.some((file) => isChanged(file, newFileDest))).toBe(true);
 
 		const { cache } = fCache.createFromFile(expectedCacheFilePath);
```
ybiquitous, line 116: [suggest] I notice this changed test cases also pass on the `main` branch. So I guess the following case is necessary to test the fixed behavior. What do you think?

```js
describe('standalone cache uses a config file', () => {
	const cwd = path.join(__dirname, 'tmp', 'standalone-cache-use-config-file');

	safeChdir(cwd);

	const configFile = path.join(cwd, '.stylelintrc.json');
	const lintedFile = path.join(cwd, 'a.css');

	beforeEach(async () => {
		await fs.writeFile(lintedFile, 'a {}');
	});

	afterEach(async () => {
		await removeFile(configFile);
		await removeFile(lintedFile);
	});

	it('cache is discarded when a config file is changed', async () => {
		const config = { files: [lintedFile], config: undefined, cache: true };

		// No warnings when a rule is disabled.
		await fs.writeFile(configFile, JSON.stringify({ rules: { 'block-no-empty': null } }));
		const { results } = await standalone(config);
		expect(results[0].warnings.length).toBe(0);

		// Some warnings when a rule becomes enabled by changing the config.
		await fs.writeFile(configFile, JSON.stringify({ rules: { 'block-no-empty': true } }));
		const { results: resultsNew } = await standalone(config);
		expect(resultsNew[0].warnings.length).toBe(1);
	});
});
```

ybiquitous (commented):
@kimulaco Thank you for addressing the reviews.

I've left a suggestion for the test, but the code looks almost good enough to merge.

lib/__tests__/standalone-cache.test.js
kimulaco: Thanks. You are right. It was not enough to only refer to `ignored`.
I added a test you suggested.  https://github.com/stylelint/stylelint/pull/6356/commits/a3bff965057239859b87e9e9f400b43e975f22cd

ybiquitous (approved):
Thank you for dealing with several reviews. LGTM 👍🏼

kimulaco:
@ybiquitous Thanks for reviews. I understood stylelint better than ever.

jeddy3 (approved):
@kimulaco Thank you for fixing the bug, LGTM!

@ybiquitous Thank you for the fantastic reviewing!

PR #6357, opened 2022-09-22
Add cacheStrategy option

### LINKED ISSUE #6170: Add cacheStrategy option

It seems the current cache strategy of Stylelint is the same with ESLint and Prettier's `metadata` strategy. When running on CI, each file metadata can change at every single CI execution.

ESLint and Prettier supports `content` cache strategy. This strategy doesn't depends on metadata but depends on each file's content, so it is very useful when running them at CI.

- https://eslint.org/docs/user-guide/command-line-interface#--cache-strategy
- https://prettier.io/docs/en/cli.html#--cache-strategy

kaorun343:

close #6170 

### DIFF
--- a/docs/user-guide/usage/options.md
+++ b/docs/user-guide/usage/options.md
@@ -116,7 +116,7 @@ CLI flag: `--cache-strategy`
 
 Strategy for the cache to use for detecting changed files. Can be either "metadata" or "content".
 
-The "content" strategy can be useful in cases where the modification time of your files change even if their contents have not. For example, this can happen during git operations like "git clone" because git does not track file modification time.
+The "content" strategy can be useful in cases where the modification time of your files changes even if their contents have not. For example, this can happen during git operations like "git clone" because git does not track file modification time.
 
 ## `maxWarnings`
 
--- a/docs/user-guide/usage/options.md
+++ b/docs/user-guide/usage/options.md
@@ -110,6 +110,14 @@ If a directory is specified, Stylelint creates a cache file inside the specified
 
 _If the directory of `cacheLocation` does not exist, make sure you add a trailing `/` on \*nix systems or `\` on Windows. Otherwise, Stylelint assumes the path to be a file._
 
+## `cacheStrategy`
+
+CLI flag: `--cache-strategy`
+
+Strategy for the cache to use for detecting changed files. Can be either "metadata" or "content".
+
+The "content" strategy can be useful in cases where the modification time of your files change even if their contents have not. For example, this can happen during git operations like "git clone" because git does not track file modification time.
+
 ## `maxWarnings`
 
 CLI flags: `--max-warnings, --mw`
--- a/lib/__tests__/standalone-cache.test.js
+++ b/lib/__tests__/standalone-cache.test.js
@@ -217,7 +217,7 @@ describe('standalone cache uses cacheStrategy', () => {
 		expect(results.some((file) => file.source === newFileDest)).toBe(true);
 
 		// No content change, but file metadata id changed
-		await fs.utimes(newFileDest, Date.now(), Date.now());
+		await fs.utimes(newFileDest, new Date(), new Date());
 		const { results: resultsCached } = await standalone(getConfig({ cacheStrategy }));
 
 		expect(resultsCached.some((file) => file.source === newFileDest)).toBe(true);
--- a/lib/__tests__/standalone-cache.test.js
+++ b/lib/__tests__/standalone-cache.test.js
@@ -208,6 +208,12 @@ describe('standalone cache uses cacheStrategy', () => {
 		await removeFile(newFileDest);
 	});
 
+	it('cacheStrategy is invalid', async () => {
+		await expect(standalone(getConfig({ cacheStrategy: 'foo' }))).rejects.toThrow(
+			'"foo" cache strategy is unsupported. Specify either "metadata" or "content"',
+		);
+	});
+
 	it('cacheStrategy is "metadata"', async () => {
 		const cacheStrategy = 'metadata';
 
--- a/lib/__tests__/standalone-cache.test.js
+++ b/lib/__tests__/standalone-cache.test.js
@@ -194,3 +194,24 @@ describe('standalone cache uses cacheLocation', () => {
 		expect(cache.getKey(validFile)).toBeTruthy();
 	});
 });
+
+describe('standalone cache uses cacheStrategy', () => {
+	const expectedCacheFilePath = path.join(cwd, '.stylelintcache');
+
+	afterEach(async () => {
+		// clean up after each test
+		await removeFile(expectedCacheFilePath);
+	});
+
+	it('cacheStrategy is "contenet"', async () => {
+		await standalone(getConfig({ cacheStrategy: 'content' }));
+
+		// Ensure cache file is created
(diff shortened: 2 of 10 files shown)


lib/cli.js
```diff
@@ -173,6 +175,10 @@ const meowOptions = {
         If the directory for the cache does not exist, make sure you add a trailing "/"
         on *nix systems or "\\" on Windows. Otherwise the path will be assumed to be a file.
 
+      --cache-strategy              [default: 'metadata']
+
+	    // TODO
```
kaorun343, line 180: Please help me to fill this section ( I'm not good at English.).

jeddy3, line 180:
```suggestion
	    Strategy for the cache to use for detecting changed files. Can be either
	    "metadata" or "content".

        The "content" strategy can be useful in cases where the modification time of
        your files change even if their contents have not. For example, this can happen
        during git operations like "git clone" because git does not track file modification
        time.  
```

Let's adopt [ESLint's wording](https://eslint.org/docs/latest/user-guide/command-line-interface#--cache-strategy).

Let's also add it to these docs:
- https://github.com/stylelint/stylelint/blob/main/docs/user-guide/usage/cli.md
- https://github.com/stylelint/stylelint/blob/main/docs/user-guide/usage/options.md

jeddy3 (changes requested):
@kaorun343 Thanks for this.

I've suggested a description for the option.

lib/cli.js
kaorun343: Thanks a lot!

lib/__tests__/standalone-cache.test.js
```diff
@@ -196,3 +196,24 @@ describe('standalone cache uses cacheLocation', () => {
...
+		await standalone(getConfig({ cacheStrategy: 'content' }));
+
+		// Ensure cache file is created
+		expect(existsSync(expectedCacheFilePath)).toBe(true);
+
+		const { cache } = fCache.createFromFile(expectedCacheFilePath, undefined, true);
+
+		expect(typeof cache.getKey(validFile)).toBe('object');
+		expect(cache.getKey(validFile)).toBeTruthy();
+	});
```
ybiquitous, line 218: [suggest] I think we'd like to test the behavior at a higher level instead of using `fCache`. For example,

```js
const cacheStrategy = 'content';

await fs.copyFile(validFile, newFileDest);

const { results } = await standalone(getConfig({ cacheStrategy }));
expect(results.some((file) => file.source === newFileDest)).toBe(true);

// No content change, but file metadata id changed
await fs.utimes(newFileDest, Date.now(), Date.now());

const { results: resultsCached } = await standalone(getConfig({ cacheStrategy }));
expect(resultsCached.some((file) => file.source === newFileDest)).toBe(false);
```

I guess it's better also to add a test case for `cacheStrategy = 'metadata'` since we can compare the `content` case easily.

What do you think?

docs/user-guide/usage/options.md
```diff
@@ -110,6 +110,14 @@ If a directory is specified, Stylelint creates a cache file inside the specified
 
 _If the directory of `cacheLocation` does not exist, make sure you add a trailing `/` on \*nix systems or `\` on Windows. Otherwise, Stylelint assumes the path to be a file._
 
+## `cacheStrategy`
+
+CLI flag: `--cache-strategy`
+
+Strategy for the cache to use for detecting changed files. Can be either "metadata" or "content".
```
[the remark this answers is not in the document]
jeddy3: > Does it make sense to add a description of the default value?

I think we're fine without it as the description of the "content" strategy hints at the opposite, e.g. "git does not track file modification time.".

(Just an aside to thank you for your comprehensive review of this pull request and all your suggestions, @ybiquitous!. It'd be great to see this and the other cache-related pull request land.)

lib/utils/FileCache.js
```diff
@@ -18,13 +19,15 @@ const DEFAULT_HASH = '';
 class FileCache {
 	constructor(
 		cacheLocation = DEFAULT_CACHE_LOCATION,
+		cacheStrategy = DEFAULT_CACHE_STRATEGY,
 		cwd = process.cwd(),
 		hashOfConfig = DEFAULT_HASH,
 	) {
```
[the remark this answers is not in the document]
kaorun343: Sounds good! Thanks!!

lib/cli.js
```diff
@@ -410,6 +425,10 @@ module.exports = async (argv) => {
 		optionsBase.cacheLocation = cli.flags.cacheLocation;
 	}
 
+	if (cli.flags.cacheStrategy) {
+		optionsBase.cacheStrategy = cli.flags.cacheStrategy;
+	}
+
```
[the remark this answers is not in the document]
kaorun343: Done. [539b75e](https://github.com/stylelint/stylelint/pull/6357/commits/539b75e5192960537feb118fb711eb4324cf32ba)

lib/__tests__/standalone-cache.test.js
kaorun343: This test case looks better than the code I wrote.

https://github.com/stylelint/stylelint/pull/6357/commits/5b5088f043441398bdbdfea76897f9207b5d824d

kaorun343:
@ybiquitous Thanks for the review! I really appreciated for your support.

@jeddy3 Please review this PR again.

lib/__tests__/standalone-cache.test.js
```diff
@@ -194,3 +194,47 @@ describe('standalone cache uses cacheLocation', () => {
...
+	it('cacheStrategy is "metadata"', async () => {
+		const cacheStrategy = 'metadata';
+
+		await fs.copyFile(validFile, newFileDest);
+		const { results } = await standalone(getConfig({ cacheStrategy }));
+
+		expect(results.some((file) => file.source === newFileDest)).toBe(true);
+
+		// No content change, but file metadata id changed
+		await fs.utimes(newFileDest, Date.now(), Date.now());
```
ybiquitous, line 220:
```suggestion
		await fs.utimes(newFileDest, new Date(), new Date());
```

lib/__tests__/standalone-cache.test.js
```diff
@@ -194,3 +194,47 @@ describe('standalone cache uses cacheLocation', () => {
...
+	it('cacheStrategy is "content"', async () => {
+		const cacheStrategy = 'content';
+
+		await fs.copyFile(validFile, newFileDest);
+		const { results } = await standalone(getConfig({ cacheStrategy }));
+
+		expect(results.some((file) => file.source === newFileDest)).toBe(true);
+
+		// No content change, but file metadata id changed
+		await fs.utimes(newFileDest, Date.now(), Date.now());
```
ybiquitous, line 235:
```suggestion
		await fs.utimes(newFileDest, new Date(), new Date());
```

ybiquitous (commented):
[suggest] Umm..., `fs.utimes()` fails on Windows with `Date.now()`. Instead, how about passing `new Date()`?

ybiquitous (commented):
Thanks. I've left a suggestion about the test, but the code looks almost good.

ybiquitous (approved):
Thank you for addressing several reviews and merging the main branch. LGTM 👍🏼

jeddy3 (approved):
@kaorun343 Thank you for adding this option. LGTM.

@ybiquitous Thank you for another comprehensive and excellent reviewing.

PR #6370, opened 2022-09-28
Fix lint settings to honor `.mjs` files

ybiquitous:

None.

There is a file with `.mjs` extension: [`scripts/benchmark-rule.mjs`](https://github.com/stylelint/stylelint/blob/2d3a3fb37a76e2a34f0732a2239d1d350ceffb7a/scripts/benchmark-rule.mjs)


jeddy3 (approved):
LGTM, thank you.

Issue #6343, opened 2022-09-12
Release 14.12.0


jeddy3:
- [x] stylelint release
- [x] stylelint-config-recommended update/release
- [x] stylelint-config-standard update/release
- [x] stylelint-demo update
- [x] stylelint.io update
- [x] tweet

We've got a [few things to release](https://github.com/stylelint/stylelint/pull/6301/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed).

Let's do a release after the following two are merged:

- https://github.com/stylelint/stylelint/pull/6334
- https://github.com/stylelint/stylelint/pull/6312

jeddy3:
The above issues are now merged.

I'm away until the weekend. Happy to release then unless someone fancies releasing sometime over the next couple of days.

jeddy3:
Done 🎉 

Thanks to everyone who contributed to this release, including triaging issues and reviewing pull requests.

jeddy3:
I manually tweaked the order of the changelog items before release. It was easy to do and I think changesets for generating the first version of the changelog is useful.

The avatars of the contributors isn't showing up [in our release](https://github.com/stylelint/stylelint/releases/tag/14.12.0) ([should be like this](https://github.com/postcss/postcss/releases/tag/8.4.14)). Does anyone have any ideas?

ybiquitous:
@jeddy3 Thank you for taking your time for the releasing tasks! ❤️ 

> The avatars of the contributors isn't showing up [in our release](https://github.com/stylelint/stylelint/releases/tag/14.12.0) ([should be like this](https://github.com/postcss/postcss/releases/tag/8.4.14)). Does anyone have any ideas?

It seems we need to use just mentions instead of links on GitHub Releases like this:

```diff
-[@ybiquitous](https://github.com/ybiquitous)
+@ybiquitous
```

We can manually edit a release description after copying and pasting from `CHANGELOG.md`, but it's a bit of a pain (we also need to keep links on `CHANGELOG.md` still). I don't have a good idea. 😓

JounQin:
There is a "Generate release notes" button when publishing a new release on GitHub web UI.

I think it could be a GitHub product problem?

ybiquitous:
I think @jeddy3 probably just copied and pasted the changelog entries as the maintainer guide says:

[...]

Issue #6253, opened 2022-08-11
Use automation for a changelog and release management


JounQin:
### Describe the documentation issue

With https://github.com/changesets/changesets/tree/main/packages/changelog-github, we'll have `thanks` mentions for contributors in our changelog and releases.

### What solution would you like to see?

Enable [`changesets`](https://github.com/changesets/changesets)

---

Alternative, manage `thanks` mentions manually.

ybiquitous:
@JounQin Thanks for the proposal. I have some questions.

1. We have the following workflow already for updating the `CHANGELOG.md` file, but will your proposed tool replace with it?

https://github.com/stylelint/stylelint/blob/4772851c0a1745db03f149ac35a2de6fe375c369/docs/maintainer-guide/pull-requests.md?plain=1#L27-L32

2. Can we keep the existing CHANGELOG format with this tool? Because the format change may affect the [CHANGELOG page](https://stylelint.io/CHANGELOG) in our website (will permalinks be retained?).

3. Will this tool affect the existing release workflow with `np`?

https://github.com/stylelint/stylelint/blob/4772851c0a1745db03f149ac35a2de6fe375c369/docs/maintainer-guide/releases.md?plain=1#L7-L17

If a new tool could reduce the tedious work (e.g. updating the changelog file), that would be great!

JounQin:
> We have the following workflow already for updating the CHANGELOG.md file, but will your proposed tool replace with it?

It will.

> Can we keep the existing CHANGELOG format with this tool? Because the format change may affect the [CHANGELOG page](https://stylelint.io/CHANGELOG) in our website (will permalinks be retained?).

Sure, we can define our own changelog format while keeping previous history changelog unmodified.

> Will this tool affect the existing release workflow with np?

We only need to merge a PR like https://github.com/un-ts/prettier/pull/213 for releasing.

ybiquitous:
@JounQin Thanks for the answers. Sounds good to me. 👍🏼

hudochenkov:
We need to consider quality of a changelog before adding any automated tool.

Changelog is one of the most important parts of documentation of any open source project. As I consumer of many open source packages, I've seen lots of changelogs. In my experience automated changelogs usually provide not enough information to the user. Or changelog items are confusing and only make sense to package developers. Or include information, which is not needed to the users (for example dev dependencies updates).

We need to make sure that our changelog provides only needed information and in a good way.

JounQin:
@hudochenkov 

[...]

PR #6371, opened 2022-09-28
Automate GitHub Release

ybiquitous:

Ref: https://github.com/stylelint/stylelint/issues/6343#issuecomment-1250353973
Ref: #6253

I believe this PR can omit the following step of releasing:

https://github.com/stylelint/stylelint/blob/eee9deb35190f46c5f964769d521fef5d7d1d7a8/docs/maintainer-guide/releases.md?plain=1#L16

This is not just a copy-and-paste of `CHANGELOG.md`. Instead, it rewrites user links to user mentions (e.g. `@ybiquitous`) so that the "Contributors" section should be displayed on a release page [like this](https://github.com/postcss/postcss/releases/tag/8.4.14).

But, one downside. This PR increase complexity: a remark script and a GitHub action. It may be too much for this automation. I'll close the PR if anyone expresses a concern.

### DIFF
--- a/.github/workflows/preparing-github-release.yml
+++ b/.github/workflows/preparing-github-release.yml
@@ -1,0 +1,36 @@
+name: Preparing GitHub Release
+
+on:
+  push:
+    tags:
+      - '**'
+
+permissions:
+  contents: write
+
+jobs:
+  prepare:
+    name: Prepare
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Node.js LTS
+        uses: actions/setup-node@v3
+        with:
+          node-version: 'lts/*'
+          cache: npm
+
+      - name: Install latest npm
+        run: npm install --global npm@latest
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Extract changelog entries
+        run: npm run changelog-to-github-release --silent > /tmp/changes.md
+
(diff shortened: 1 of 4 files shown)


.github/workflows/preparing-github-release.yml
```diff
@@ -0,0 +1,36 @@
...
+        run: npm install --global npm@latest
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Extract changelog entries
+        run: npm run changelog-to-github-release --silent > /tmp/changes.md
+
+      - name: Create draft release
+        run: gh release create "${GITHUB_REF_NAME}" --draft --notes-file /tmp/changes.md
```
ybiquitous, line 34: See also https://cli.github.com/manual/gh_release_create

jeddy3 (changes requested):
Thanks for this! Looks good.

> But, one downside. This PR increase complexity: a remark script and a GitHub action. It may be too much for this automation. I'll close the PR if anyone expresses a concern.

There's no harm in trying it out. If the complexity proves problematic later down the line we can remove it and revert to the manual method.

Let's remove the corresponding step in the release doc as part of this PR.

ybiquitous:
@jeddy3 Thanks for the feedback! I've updated the releasing guide via fddc0da3b.

ybiquitous:
If this new workflow is not as expected in the next release, let's recover it manually.

PR #6374, opened 2022-09-29
Fix `unit-no-unknown` false positives for font-relative length units

### LINKED ISSUE #6373: Fix `unit-no-unknown` false positives for font-relative length units

For example, the CSS code using the `ic` unit:

```css
a { width: 10ic; }
```

We can reproduce the bug on the [demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYKDuBLKAXAFvCgIwAMWYA3CgL4A6AdiADQgDOAng7kgB4Lg2bFuAgMAZlgDmA4IxQo6IAE4BXADYw2SonIYKFS1Qyy4AtAwhnjAa0sYGOlLjUx5tRjRA0gA).

```json
{
  "rules": {
    "unit-no-unknown": true
  }
}
```

CLI with `stylelint *.css`

14.13.0

No problems are to be reported.

### What actually happened?

The following problem is reported:

```
test.css
 1:14  ✖  Unexpected unknown unit "ic"  unit-no-unknown

```

No.

### Proposal to fix the bug

To Update the internal list for the units:

https://github.com/stylelint/stylelint/blob/eee9deb35190f46c5f964769d521fef5d7d1d7a8/lib/reference/units.js#L5-L12

ybiquitous:

Closes #6373

I've updated the unit list based on the ["CSS Values and Units Module Level 4" spec](https://w3c.github.io/csswg-drafts/css-values-4/#font-relative-lengths).

### DIFF
--- a/lib/reference/units.js
+++ b/lib/reference/units.js
@@ -3,13 +3,19 @@
 const uniteSets = require('../utils/uniteSets.js');
 
 const lengthUnits = new Set([
-	// Relative length units
+	// Font-relative length units
+	'cap',
+	'ch',
 	'em',
 	'ex',
-	'ch',
-	'rem',
-	'rlh',
+	'ic',
 	'lh',
+	'rcap',
+	'rch',
+	'rem',
+	'rex',
+	'ric',
+	'rlh',
 	// Viewport-percentage lengths
 	'dvh',
 	'dvmax',
--- a/.changeset/old-moles-sort.md
+++ b/.changeset/old-moles-sort.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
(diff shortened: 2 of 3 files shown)


lib/reference/units.js
```diff
@@ -3,13 +3,19 @@
...
-	'ch',
+	'ic',
+	'lh',
+	'rcap',
+	'rch',
 	'rem',
+	'rex',
+	'ric',
 	'rlh',
-	'lh',
```
ybiquitous, line 12: Sort alphabetically.

jeddy3 (approved):
LGTM, thanks!

ybiquitous:
Thanks for the quick review!

PR #6377, opened 2022-09-29
Refactor to add function names

ybiquitous:

None.

Giving function names (avoiding anonymous) helps debugging.

This refactoring relies on the ESLint [`func-names`](https://eslint.org/docs/latest/rules/func-names) rule.

Also, it seems better to add the rule to our [`eslint-config-stylelint`](https://github.com/stylelint/eslint-config-stylelint) config.

### DIFF
--- a/lib/utils/isOnlyWhitespace.js
+++ b/lib/utils/isOnlyWhitespace.js
@@ -8,15 +8,12 @@ const isWhitespace = require('./isWhitespace');
  * @param {string} input
  * @returns {boolean}
  */
-module.exports = function (input) {
-	let isOnlyWhitespace = true;
-
+module.exports = function isOnlyWhitespace(input) {
 	for (const element of input) {
 		if (!isWhitespace(element)) {
-			isOnlyWhitespace = false;
-			break;
+			return false;
 		}
 	}
 
-	return isOnlyWhitespace;
+	return true;
 };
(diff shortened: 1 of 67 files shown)


lib/utils/isOnlyWhitespace.js
```diff
@@ -8,15 +8,12 @@ const isWhitespace = require('./isWhitespace');
...
 	for (const element of input) {
 		if (!isWhitespace(element)) {
-			isOnlyWhitespace = false;
-			break;
+			return false;
 		}
 	}
 
-	return isOnlyWhitespace;
+	return true;
```
ybiquitous, line 18: [note] I've simplified the logic to avoid using the same local variable name as the function name.

jeddy3 (approved):
Fantastic, thank you!

> Also, it seems better to add the rule to our [eslint-config-stylelint](https://github.com/stylelint/eslint-config-stylelint) config.

I agree. We can continue to turn the rule on locally until we publish a new version of our ESLint config.

While reviewing this pull request and seeing all the function name, it reminded me that I wanted to refactor some of utils so that we don't have ones specific to languages like SCSS and Less. Instead that logic would only be contained in the `isStandardSyntax*` utils. I'll try to make time to either open an issue or fire up a pull request myself.

ybiquitous:
I appreciate your kindful review. 😊 

> Instead that logic would only be contained in the `isStandardSyntax*` utils.

I agree entirely. 👍🏼

PR #6380, opened 2022-09-30
Change `extends` in `overrides` to be consistent with `plugins` behaviour

jasikpark:
- Adds docs describing how different properties get merged in `overrides`
- Merges `extends` arrays together in `overrides`
- Adds a test for that `extends` merging functionality


jeddy3:
@jasikpark Thanks for opening this pull request to show the current behaviour.

> If this is the intended behavior

I've been looking back through some issues (https://github.com/stylelint/stylelint/pull/5603, https://github.com/stylelint/stylelint/issues/5656 etc) and it doesn't look like we ever discussed what the intended behaviour should be. 

It feels safe to merge `plugins` as it simply increases the pool of available rules. `extends`, on the other hand, can change what rules are turned on and off. However, having consistent behaviour for both may make the most sense.

Although the current behaviour is undocumented, we can't know how many users rely on it.

Anyone else got thoughts on this?

Whatever we decide, let's update [our docs](https://stylelint.io/user-guide/configure#overrides) to describe the behaviour for each of the most used configuration properties: `rules`, `customSyntax`, `extends` and `plugins`.

At the moment, I believe it's:

- `rules` - merge
- `customSyntax` - replace
- `extends` - replace
- `plugins` - merge

ybiquitous:
@jeddy3 Thanks for summarizing the issues.

> `extends`, on the other hand, can change what rules are turned on and off. However, having consistent behaviour for both may make the most sense.

I agree with the idea that `extends` should cover merging like `plugins`.
And, also updating our docs. 👍🏼

jeddy3:
OK, let's make the `extends` behaviour consistent with `plugins`. It was an oversight, so shall we classify as a bug fix?

@jasikpark Feel free to fix the behaviour in this pull request and update the docs. 

For the docs, we can add a bullet at the [end of this section](https://stylelint.io/user-guide/configure#overrides).

- Configuration properties with a single value, e.g. `customSyntax`, will be overridden whereas everything else, e.g. `extends`, `plugins`, `rules` etc., will be merged.

jasikpark:
Okay, pushed some basics docs & merging `extends`. What is the explicit list of things that should be merged vs replaced?

jasikpark:
looks like `rules` are not deduplicated, though they are merged

https://github.com/jasikpark/stylelint/blob/7ac953e87cef7da2b7264220fd766b361e4ee5b5/lib/augmentConfig.js#L236-L246

jasikpark:
I guess everything in 

https://github.com/stylelint/stylelint/blob/cf04b2be5b7e14ec3f9c633a9b1a7d1653faa920/types/stylelint/index.d.ts#L36-L57

that is an array should be merged right? since technically any of those can be passed to overrides?

jasikpark:
might be worth looking to eslint's behavior here as well https://eslint.org/docs/latest/user-guide/configuring/configuration-files#how-do-overrides-work

jasikpark:
Seems that `extends`, `plugins`, `processors`, `processorFunctions`, `ignoreFiles`, `codeProcessors`, `resultProcessors` are arrays and might be merged similarly to the existing `plugins` behavior

`rules` and `pluginFunctions` are objects containing a variable number of properties, which could also be merged potentially

`ignorePatterns`, `quiet`, `defaultSeverity`, `ignoreDisables`, `report*`, and `customSyntax` are javscript built-in datatypes like boolean and string and should be replaced.

I'd suggest maybe this PR can just alter `extends` behavior and then the other config values can be updated in a second PR?

jasikpark:
Since I'm not sure what the semantics are of all these config values

jeddy3:
@jasikpark Thanks for making the changes and for digging more into the overrideable options.

> I'd suggest maybe this PR can just alter extends behavior and then the other config values can be updated in a second PR?

That sounds best. The general behaviour you outline in https://github.com/stylelint/stylelint/pull/6380#issuecomment-1267722440, SGTM. 

We're going to remove Processors in the next major release. That leaves the following as user-facing options:

- `extends`
- `plugins`
- `ignoreFiles`
- `quiet`
- `defaultSeverity`
- `ignoreDisables`
- `report*`
- `customSyntax`

If we haven't already, it'd be good to have tests for these.

I've added a changeset entry. I think this pull request is ready for review.

docs/user-guide/configure.md
```diff
@@ -297,6 +297,8 @@ Here is how overrides work in a configuration file:
 - Glob pattern overrides have higher precedence than the regular configuration in the same config file. Multiple overrides within the same config are applied in order. That is, the last override block in a config file always has the highest precedence.
 - A glob specific configuration works almost the same as any other Stylelint config. Override blocks can contain any configuration options that are valid in a regular config.
 - Multiple glob patterns can be provided within a single override block. A file must match at least one of the supplied patterns for the configuration to apply.
+- Singular config values like `defaultSeverity` and `customSyntax` will be replaced by overrides.
+- Complex config values like `plugins`, `extends`, `rules`, etc. will be appended.
```
jeddy3, line 301:
```suggestion
- `customSyntax` will be replaced by overrides.
- `plugins`, `extends`, `rules`, etc. will be appended.
```

As we now have tests for these, let's go with them for now.

jeddy3 (changes requested):
> Maybe I should add tests + docs that describe where we are today && then another PR can fix everything?

Would you mind doing that in another pull request?

I've made a suggestion to tweak the docs for this pull request.

.changeset/grumpy-scissors-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
```
jasikpark, line 2: is this a patch when it's almost certainly a breaking change for existing configs? ig we're saying it's a bugfix?

ybiquitous: The PR change might be a potential breaking change for some people, but I believe we'd like to avoid as frequent major releases as possible. Because we plan to have various features and breaking changes on the next major version, so the merge of this PR would be blocked until the next major version would be prepared.

For people who would like to keep the current behavior, it might be more kind if there would be a code example like this:

```md
`overrides.extends` is fixed to make consistency with the `overrides.plugins` behavior and be merged not replaced.
If you would like to keep the previous behavior, you would need to change your config as below:
```

```diff
{
- "extends": ["config-a"]
  "overrides": [
    {
      "rules": ["*.module.css"],
      "extends": ["config-b"]
    },
+   {
+     "rules": ["*.css"],
+     "extends": ["config-a"]
+   },
  ]
}
```

ybiquitous (approved):
Thanks, LGTM 👍🏼

jeddy3 (approved):
LGTM, thank you!

We can merge once we start actively working on a `v15` branch. 

I've created a tracking issue for preparing 15.0.0 (https://github.com/stylelint/stylelint/issues/6409).

ybiquitous:
I've updated the PR base branch to `v15`. Can we merge this PR? @jeddy3

jeddy3:
I've moved the migrating information out of the changeset and into a migration guide doc.

> I've updated the PR base branch to v15. Can we merge this PR? @jeddy3

Yes.

jeddy3:
This merge behaviour is aligned with how a flat config would work, and so is a step in the right direction for us.

[From this article](https://eslint.org/blog/2022/08/new-config-system-part-2/#goodbye-extends%2C-hello-flat-cascade):

> Inside of the array, ESLint finds all config objects that match the file being linted and merges them together in much the same way that eslintrc did. The only real difference is the merge happens from the top of the array down to the bottom instead of using files in a directory structure.

docs/migration-guide/to-15.md
```diff
@@ -0,0 +1,23 @@
+# Migrating to 15.0.0
+
+## Change of `overrides.`extends` behavior
```
ybiquitous, line 3: Typo?

```suggestion
## Change of `overrides.extends` behavior
```

docs/migration-guide/to-15.md
```diff
@@ -0,0 +1,23 @@
...
+- "extends": ["config-a"]
+  "overrides": [
+    {
+      "rules": ["*.module.css"],
+      "extends": ["config-b"]
+    },
++   {
++     "rules": ["*.css"],
++     "extends": ["config-a"]
++   },
```
ybiquitous, line 20: Looks like a typo because JSON disallows a trailing comma.

```suggestion
+   }
```

ybiquitous (approved):
@jeddy3 Thanks so much for writing up the migration guide. There may be a few typos left, but LGTM! 👍🏼

jasikpark:
thx for getting this merge ready 💜

docs/migration-guide/to-15.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Change of `overrides.extends` behavior
+
+We changed the `overrides.extends` behavior to merge rather than replace, to make it consistent with the `overrides.plugins`.
+
+If you would like to keep the previous behavior, you should change your config to:
+
+```diff json
+{
+- "extends": ["config-a"]
```
ybiquitous, line 11:
```suggestion
- "extends": ["config-a"],
```

ybiquitous:
We can merge this when CI passes. 💪🏼

PR #6384, opened 2022-10-02
Fix false negatives on second run for cache and `severity` option

### LINKED ISSUE #4715: Fix false negatives on second run for cache and plugins

kimulaco:

Closes #4715

No, it's self-explanatory.


ybiquitous:
@kimulaco Thanks for creating the PR.

I am looking into this bug. So, the bug seems related to not plugins but *severity*. See the code below.

If `stylelintError` is `true`, a cache is discarded here: ⬇️ 

https://github.com/stylelint/stylelint/blob/751acda06a0755ddf0a0603013681173c70bf2c3/lib/standalone.js#L219-L222

Then, `stylelintError` is set to `true` here (only if the severity is `error`): ⬇️ 

https://github.com/stylelint/stylelint/blob/751acda06a0755ddf0a0603013681173c70bf2c3/lib/utils/report.js#L79-L81

The following is a production: ⬇️ 

```console
$ cat a.css
a { colo: red; }

$ cat package.json
{
  "dependencies": {
    "stylelint": "14.13.0"
  },
  "stylelint": {
    "rules": {
      "property-no-unknown": [true, { "severity": "warning" }]
    }
  }
}

$ rm -f .stylelintcache

$ npx stylelint *.css --cache -f json | jq -c '.[0].warnings'
[{"line":1,"column":5,"endLine":1,"endColumn":9,"rule":"property-no-unknown","severity":"warning","text":"Unexpected unknown property \"colo\" (property-no-unknown)"}]

$ npx stylelint *.css --cache -f json | jq -c '.[0].warnings'
[]
```

We can see the second run emits no warnings.

On the other hand, the second run emits warnings after changing the severity from `warning` to `error` in the `package.json` file above.

```console
$ npx stylelint *.css --cache -f json | jq -c '.[0].warnings'
[{"line":1,"column":5,"endLine":1,"endColumn":9,"rule":"property-no-unknown","severity":"error","text":"Unexpected unknown property \"colo\" (property-no-unknown)"}]

$ npx stylelint *.css --cache -f json | jq -c '.[0].warnings'
[{"line":1,"column":5,"endLine":1,"endColumn":9,"rule":"property-no-unknown","severity":"error","text":"Unexpected unknown property \"colo\" (property-no-unknown)"}]
```

kimulaco:
@ybiquitous Thanks. I thought I needed to cache the result. But I could understand that it was caused by a condition to remove the cache.
Please give me some time to fix it. 🙏

ybiquitous:
@kimulaco I hope you don't mind. I appreciate your contribution to this project. 😊

kimulaco:
@ybiquitous I fixed. It contains a new approach, so I would like your opinion.

First, I reverted all codes that cache the `postcssResult`. (https://github.com/stylelint/stylelint/pull/6384/commits/aa28b1af1a85acc6123a91b89507b0df66415c15, https://github.com/stylelint/stylelint/pull/6384/commits/db190028f17a2283e1f4d9ac53ea28559dfc1701)

Then, I added `stylelintWarning` to `StylelintPostcssResult`. The cache is discarded also when `stylelintWarning` is `true`.
Because, `stylelintError` is equal to `errored`, so to prevent breaking spec of `errored`. 

https://github.com/stylelint/stylelint/blob/59d6a3f174dc6787202db388763cb36da516d5f2/docs/user-guide/usage/node-api.md?plain=1#L53-L55

ybiquitous (commented):
@kimulaco Thank you, LGTM! 👍🏼 

In addition, could you please add a changelog entry? The entry line should be like this (from the PR title): ⬇️ 

```
Fixed: false negatives on second run for cache and `severity` option
```

kimulaco:
@ybiquitous Thanks for the review! I added a changelog.

PR #6389, opened 2022-10-03
Fix `printf`-like formatting for custom messages

ybiquitous:

Ref: #6312

The Node.js [`util.format()`](https://nodejs.org/api/util.html#utilformatformat-args) API can concatenate extra arguments to the result string. This behavior is not suitable for custom messages. So, this commit replace `util.format()` with our implementation.

> If there are more arguments passed to the `util.format()` method than the number of specifiers, the extra arguments are concatenated to the returned string, separated by spaces:
>
> ```js
> util.format('%s:%s', 'foo', 'bar', 'baz');
> // Returns: 'foo:bar baz'
> ```

### DIFF
--- a/lib/utils/report.js
+++ b/lib/utils/report.js
@@ -1,7 +1,5 @@
 'use strict';
 
-const util = require('util');
-
 /**
  * Report a problem.
  *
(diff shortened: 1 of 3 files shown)


lib/utils/report.js
```diff
@@ -124,8 +122,19 @@ function buildWarningMessage(message, messageArgs) {
...
 }
+
+/**
+ * @param {string} format
+ * @param {Array<unknown>} args
+ * @returns {string}
+ */
+function printfLike(format, ...args) {
+	return args.reduce((/** @type {string} */ result, arg) => {
+		return result.replace(/%[ds]/, String(arg));
```
ybiquitous, line 138: [note] For now, only `%s` and `%d` support seems enough.

jeddy3 (approved):
LGTM, thank you!

PR #6396, opened 2022-10-07
Document using stylelint inside npm-scripts

### LINKED ISSUE #6395: Document need to quote globs

1. Create 3 css files with some invalid code and put them in nested folders.
  styles > fileA, folder1
  styles > folder1 > fileB, folder2
  styles > folder1 > folder2 > fileC                                                           

3. run ```npm run lint```  on windows machine
4. run same command on Mac os
5. On Mac we see one error caught by stylelint against 3 error on Windows machine

This is about CLI command issue so I used a basic config just for demonstration.

"package.json" file:
```
{
  "name": "stylelint-bug",
  "version": "1.0.0",
  "description": "Reproducing inconsistency between Mac and Windows",
  "main": "index.js",
  "scripts": {
    "lint": "stylelint styles/**/*.css"
  },
  "author": "Makan Edrisi",
  "license": "ISC",
  "devDependencies": {
    "stylelint": "^14.13.0",
    "stylelint-config-standard": "^28.0.0"
  }
}

```

".stylelintrc.json" file:
```
{
}
```

npm run lint

14.13.0

It should report same result on Windows machine as Mac machine

### What actually happened?

It seems style lint used glob pattern on Mac machine, so it only looked at styles and checked only one file.
But on windows machine it recursively checked all those folders and checked all 3 files.

I also found https://github.com/stylelint/stylelint/issues/4609 which seems happened because of same reason.

_No response_

### Proposal to fix the bug

Changing the script to ```"lint": "stylelint styles/**/**/**/*.css"```solves this problem.
But depends on how many nested folder we have we should add more "/**/" to glob pattern works on Mac as well.

As [this article](https://medium.com/@jakubsynowiec/you-should-always-quote-your-globs-in-npm-scripts-621887a2a784)  suggests I propose the stylelint cli internally use [node-glob](https://github.com/isaacs/node-glob) to be consistent across different os.

jeddy3:

Closes #6395

As an aside, GitHub copilot wrote the example after I typed "if you are using npm scripts...". Was kinda cool to see 😆

### DIFF
--- a/docs/user-guide/get-started.md
+++ b/docs/user-guide/get-started.md
@@ -27,6 +27,8 @@ npm install --save-dev stylelint stylelint-config-standard
 npx stylelint "**/*.css"
 ```
 
+_You should include quotation marks around file globs._
+
 If you use a pretty printer alongside Stylelint, you should turn off any conflicting rules. For example, you can use [Prettier's shared config](https://www.npmjs.com/package/stylelint-config-prettier) to do that:
 
 ```shell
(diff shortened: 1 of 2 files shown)


ybiquitous (approved):
Thanks, LGTM!

PR #6415, opened 2022-10-17
Refactor the internal API

ybiquitous:

None.

This refactoring aims to make the codebase around the internal API more consistent.

Our codebase is mostly FP style, so the OOP-like code using `.bind()` seems inconsistent to me.

Also, the dependency between the internal modules seems implicit through `.bind()`.


types/stylelint/index.d.ts
```diff
@@ -529,16 +529,6 @@ declare module 'stylelint' {
...
-
-			_getPostcssResult: (options?: GetPostcssOptions) => Promise<PostCSS.Result>;
-			_lintSource: (options: GetLintSourceOptions) => Promise<PostcssResult>;
-			_createStylelintResult: (
-				postcssResult: PostcssResult,
-				filePath?: string,
-			) => Promise<LintResult>;
-
-			getConfigForFile: (searchPath?: string, filePath?: string) => Promise<CosmiconfigResult>;
-			isPathIgnored: (s?: string) => Promise<boolean>;
```
ybiquitous, line 541: [note] I believe `InternalApi` and `createLinter` is private and never break our ecosystem.

https://github.com/stylelint/stylelint/blob/d40288983161124d7b8a29be219835bf7ea6841d/types/stylelint/index.d.ts#L458-L463

However, if this pull request would harm the ecosystem, I'd like to close it.

lib/__tests__/getConfigForFile.test.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
```
ybiquitous, line 1: [note] `createLinter.test.js` is divided into two files.

lib/isPathIgnored.js
```diff
@@ -22,16 +24,14 @@ module.exports = async function isPathIgnored(stylelint, filePath) {
...
 
 	if (!result) {
 		return true;
 	}
 
 	// Glob patterns for micromatch should be in POSIX-style
-	const ignoreFiles = /** @type {Array<string>} */ (result.config.ignoreFiles || []).map((s) =>
-		normalizePath(s),
-	);
+	const ignoreFiles = [result.config.ignoreFiles || []].flat().map((s) => normalizePath(s));
```
ybiquitous, line 34: [note] The technique using `.flat()` has an advantage for the type inference.

lib/lintSource.js
```diff
@@ -51,13 +55,17 @@ module.exports = async function lintSource(stylelint, options = {}) {
...
+	try {
+		configForFile = await getConfigForFile(stylelint, configSearchPath, inputFilePath);
+	} catch (err) {
+		if (isCodeNotFile && isPathNotFoundError(err)) {
+			configForFile = await getConfigForFile(stylelint, cwd);
+		} else {
 			throw err;
-		});
+		}
+	}
```
ybiquitous, line 68: [note] This refactoring makes the code consistent using async/await.

jeddy3 (approved):
LGTM, thank you!

ybiquitous:
Thanks for the quick review! 😄

ricardogobbosouza:
Hi @jeddy3 
This was a significant change...it broke some plugins like for example https://github.com/webpack-contrib/stylelint-webpack-plugin
It should be a breaking change, no?

alexander-akait:
@jeddy3 @ybiquitous Friendly ping

ybiquitous:
@ricardogobbosouza @alexander-akait 
Thanks for sharing the problem. As I commented on <https://github.com/stylelint/stylelint/pull/6415#discussion_r997096049>, I thought `InternalApi` was internal, so safe to change.

https://github.com/stylelint/stylelint/blob/b70929b061051e78d9d711e559e5777eaa9f82f0/types/stylelint/index.d.ts#L521-L526

However, I'll gladly cooperate to fix the problem.

alexander-akait:
@ybiquitous Yeah, I think we should move such things from public api to prevent such situaltion, or use `#` or `_` to indicate them, because other tools can faced with it too

ybiquitous:
Agree. I'll add `@internal` annotations to the internal type definitions.

PR #6419, opened 2022-10-20
Fix `declaration-block-no-redundant-longhand-properties` false positives for `inherit` keyword

### LINKED ISSUE #6394: Fix `declaration-block-no-redundant-longhand-properties` false positives for `inherit` keyword

1. Run stylint for checking the below CSS

```css
.demo {
  border-top-left-radius: 20px;
  border-top-right-radius: inherit;
  border-bottom-right-radius: inherit;
  border-bottom-left-radius: inherit;
}
```

```json
{
}
```

stylelint src/**/*.css

^14.13.0

There's no error for this case because there are longhand properties with `inherit` value.
Per [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties#omitting_properties)

> Only the individual properties values can inherit. As missing values are replaced by their initial value, it is impossible to allow inheritance of individual properties by omitting them. The keyword inherit can be applied to a property, but only as a whole, not as a keyword for one value or another. That means that the only way to make some specific value to be inherited is to use the longhand property with the keyword inherit.

So, it's impossible to convert such set of longhand properties to shorthand and preserve initial semantic.

### What actually happened?

 5:3  ✖  Expected shorthand property "border-radius"  declaration-block-no-redundant-longhand-properties

_No response_

### Proposal to fix the bug

If a longhand property of some shorthand property is set to `inherit`, don't treat such case as a rule violation.

kimulaco:

Closes #6394

I think this CSS spec is often overlooked. So, I added a description and example to the docs. https://github.com/stylelint/stylelint/commit/94a21c3dd197144ac92c4803641b86c4fb6402c0

### DIFF
--- a/.changeset/fair-oranges-love.md
+++ b/.changeset/fair-oranges-love.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fix `declaration-block-no-redundant-longhand-properties` false positives for `inherit` keyword
--- a/.changeset/fair-oranges-love.md
+++ b/.changeset/fair-oranges-love.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fix `declaration-block-no-redundant-longhand-properties` false positives for `inherit` keyword
+Fixed: `declaration-block-no-redundant-longhand-properties` false positives for `inherit` keyword
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
@@ -22,7 +22,7 @@ a {
 }
 ```
 
-This rule will only complain if you've used the longhand equivalent of _all_ the properties that the shorthand will set.
+This rule will only complain if you've used the longhand equivalent of _all_ the properties that the shorthand will set and if their values are not `inherit`.
 
 This rule complains when the following shorthand properties can be used:
 
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
@@ -145,6 +145,18 @@ a {
 }
 ```
 
+You should use a longhand property to have only the specific value inherited. So, the following patterns are also not considered problems:
+
+<!-- prettier-ignore -->
+```css
+a {
+  margin-top: 1px;
+  margin-right: 2px;
+  margin-bottom: 3px;
+  margin-left: inherit;
+}
+```
+
 ## Optional secondary options
 
 ### `ignoreShorthands: ["/regex/", /regex/, "string"]`
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
@@ -145,18 +145,6 @@ a {
 }
 ```
 
-You should use a longhand property to have only the specific value inherited. So, the following patterns are also not considered problems:
-
-<!-- prettier-ignore -->
-```css
-a {
-  margin-top: 1px;
-  margin-right: 2px;
-  margin-bottom: 3px;
-  margin-left: inherit;
-}
-```
-
 ## Optional secondary options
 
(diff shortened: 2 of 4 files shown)


.changeset/fair-oranges-love.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fix `declaration-block-no-redundant-longhand-properties` false positives for `inherit` keyword
```
ybiquitous, line 5: [suggest] Thanks for adding the changelog. In addition, could you please use the standard prefix `FIXED:`? Please see also the [maintainer guide](https://github.com/stylelint/stylelint/blob/d171517376238ff2ba33f98e9e171ee13ca0cd81/docs/maintainer-guide/pull-requests.md#merging).

```suggestion
Fixed: `declaration-block-no-redundant-longhand-properties` false positives for `inherit` keyword
```

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -20,6 +20,8 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/declaration-block-no-redundant-longhand-properties',
 };
 
+const longhandMustValues = ['inherit'];
```
ybiquitous, line 23: [trivial] These suggestions depend on my personal preference. You can ignore them.

- How about using `Set` for a bit more efficiency?
- How about changing to a constant-like name for readability?

```suggestion
const IGNORED_VALUES = new Set(['inherit']);
```

See also:

https://github.com/stylelint/stylelint/blob/d171517376238ff2ba33f98e9e171ee13ca0cd81/lib/rules/color-named/index.js#L27

ybiquitous (commented):
@kimulaco Thank you for creating the pull request. Adding a note about `inherit` to the rule doc is a nice idea!

I've left a few trivial suggestions, but this change almost looks good. 👍🏼

lib/rules/declaration-block-no-redundant-longhand-properties/README.md
```diff
@@ -145,6 +145,18 @@ a {
...
+<!-- prettier-ignore -->
+```css
+a {
+  margin-top: 1px;
+  margin-right: 2px;
+  margin-bottom: 3px;
+  margin-left: inherit;
+}
+```
+
```
jeddy3, line 159: Let's remove this as our patterns don't tend to show edge cases. 

We can add something to do _extended description_ at the top instead. For example, expand the following line

> This rule will only complain if you've used the longhand equivalent of _all_ the properties that the shorthand will set.

To:

"This rule will only complain if you've used the longhand equivalent of _all_ the properties that the shorthand will set and if their values are not `inherit`."

jeddy3 (changes requested):
Thank you.

I've requested a change to the docs.

.changeset/fair-oranges-love.md
kimulaco: Thanks. I fixed the prefix. https://github.com/stylelint/stylelint/pull/6419/commits/117e5f63aa66a80e58ef28223fff0bace1c5b34a

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
kimulaco: Fixed to use `Set`. https://github.com/stylelint/stylelint/pull/6419/commits/39f4f273aba6354ce2ef8d4245bbf55dbb2c8b0d

lib/rules/declaration-block-no-redundant-longhand-properties/README.md
kimulaco: Thanks for review! I removed a description of edge case. https://github.com/stylelint/stylelint/pull/6419/commits/616a536d977f7a6c6ce283634a942737045fce77
Then, I updated the description to your suggestion. https://github.com/stylelint/stylelint/pull/6419/commits/f442a05438dcbcb783ee6c388f34dc801783e47a

ybiquitous (approved):
Thank you. LGTM!

jeddy3 (approved):

PR #6429, opened 2022-10-29
Add coverage badge into README

AlexElin:

 Displaying coverage data on the readme of a repository.

No, it's self-explanatory.

Codecov's badge links:
* https://docs.codecov.com/docs/status-badges
* https://docs.codecov.com/docs/github-3-customizing-codecov

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![npm version](https://img.shields.io/npm/v/stylelint)](https://www.npmjs.com/package/stylelint)
 [![Build Status](https://github.com/stylelint/stylelint/workflows/Testing/badge.svg)](https://github.com/stylelint/stylelint/actions/workflows/testing.yml?query=branch%3Amain)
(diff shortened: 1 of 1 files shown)


jeddy3:
Thank you for the pull request. 

An additional badge for code coverage feels redundant as we already have this line in our README:

> is well tested with over 15000 unit tests

Whereas the existing 3 badges provide information not communicated elsewhere on the page.

Closing so that the header remains uncluttered and concise.

AlexElin:
In my opinion the badge is better than just the phrase "**well tested** with over 15000 unit tests" in README because
* it explicitly shows how **well** the code is tested at now. And it this value can be verified by clicking the number and following to the codecov [portal](https://app.codecov.io/gh/stylelint/stylelint)
* It's a common practice for open-source projects to add badges with coverage. So, many users only look at the top part of README to find the coverage. I suppose few users will go through README to find coverage information.

PR #6437, opened 2022-10-31
Add `--globby-options` flag

### LINKED ISSUE #6109: Add `--globby-options` flag

Create a folder beginning with a dot, and set a css or scss file and run a `stylelint --fix` command. It won't work.
I'm experiencing this problem with `.storybook` folder.

any configuration would be enough

stylelint --fix '**/*.css

13.13.1

The lint fixes also files present in hidden folders

### What actually happened?

Hidden folders files are ignored

No

### Proposal to fix the bug

Read also hidden folders

sidverma32:

closes #6109


silversonicaxel:
Should you also update this documentation file https://github.com/stylelint/stylelint/blob/main/docs/user-guide/usage/options.md ?

sidverma32:
> Should you also update this documentation file https://github.com/stylelint/stylelint/blob/main/docs/user-guide/usage/options.md ?

Sure , Updated!

lib/cli.js
```diff
@@ -236,6 +238,11 @@ const meowOptions = {
       --allow-empty-input, --aei
 
         When glob pattern matches no files, the process will exit without throwing an error.
+	 
+	  --globby-options, --go
+
+		CLI option to traverse dotfiles i.e Hidden folders files will not be ignored(e.g. stylelint --fix '**/*.css)
+
```
ybiquitous, line 245: [suggest] Let's remove extra empty lines and convert TAB to SPACEs.

```suggestion

      --globby-options, --go

        CLI option to traverse dotfiles i.e Hidden folders files will not be ignored(e.g. stylelint --fix '**/*.css)	
```

lib/cli.js
```diff
@@ -72,6 +73,7 @@ const EXIT_CODE_ERROR = 2;
  * @property {boolean} [quiet]
  * @property {any} [printConfig]
  * @property {boolean} [fix]
+ * @property {string} [globbyOptions]
```
ybiquitous, line 76: [suggest] A `globbyOptions` value passed to `standalone()` should be an object:

```suggestion
 * @property {Record<string, unknown>} [globbyOptions]
```

lib/cli.js
```diff
@@ -397,6 +408,10 @@ module.exports = async (argv) => {
 			: path.resolve(process.cwd(), cli.flags.configBasedir);
 	}
 
+	if (cli.flags.globbyOptions) {
+		optionsBase.globbyOptions = cli.flags.globbyOptions;
```
ybiquitous, line 412: [suggest] What do you think about the following idea to pass JSON value?

```suggestion
		const value = cli.flags.globbyOptions;

		try {
			optionsBase.globbyOptions = JSON.parse(value);
		} catch (_) {
			process.stderr.write(
				`Invalid option ${red('"--globby-options"')}. The value ${red(
					`"${value}"`,
				)} is not valid JSON.${EOL}`,
			);
			process.exit(EXIT_CODE_ERROR); // eslint-disable-line no-process-exit
		}
```

For example:

image

ybiquitous (changes requested):
@sidverma32 Thanks for creating the pull request. Please consider addressing my reviews.

docs/user-guide/usage/cli.md
```diff
@@ -68,6 +68,12 @@ Automatically fix, where possible, problems reported by rules. [More info](optio
 
 Specify the formatter to format your results. [More info](options.md#formatter).
 
+## `globbyOptions`
+
+CLI flags: `--globby-options, --go`
+
+CLI option to traverse dotfiles i.e Hidden folders/files will not be ignored. [More info](options.md#globbyOptions).
```
ybiquitous, line 75: [suggest] Let's keep the URL hash lower-case.

```suggestion
CLI option to traverse dotfiles i.e Hidden folders/files will not be ignored. [More info](options.md#globbyoptions).
```

ybiquitous (changes requested):

lib/cli.js
sidverma32: yes, JSON sounds promising!

just a small doubt do we always need to **stringify** and **parse** the `value` or will there be any case where this can fail!!?? 🤔

ybiquitous: > just a small doubt do we always need to stringify or parse the `value` or will there be any case where this can fail!!?? 🤔

I think `JSON.parse` is a way to convert a string to an object. Do you have another way?

> Also this change is causing an error in `cli.test.js file`. not sure if i have update anything their based on this change!

What kind of errors?

sidverma32: Screenshot 2022-11-02 at 7 50 24 AM

When running `jest` command getting this error! @ybiquitous

ybiquitous: @sidverma32 You may need to update failed [snapshots](https://jestjs.io/docs/snapshot-testing) by the `npx jest -u` command. Can you please try it?

sidverma32: oh yeah! i did last time to fix the same ..my bad :)

Updated!

ybiquitous: Thanks. Additionally, can you please add the following test cases to `cli.test.js`?

- Success case: `--globby-options` should work on actual file system, e.g. `--globby-options '{"dot": true}'` should scan files in dot directories
- Failure case: `--globby-options` with an invalid JSON value should fail with the expected message and exit status

sidverma32: > Thanks. Additionally, can you please add the following test cases to `cli.test.js`?
> 
> * Success case: `--globby-options` should work on actual file system, e.g. `--globby-options '{"dot": true}'` should scan files in dot directories
> * Failure case: `--globby-options` with an invalid JSON value should fail with the expected message and exit status

```
it('flags.globbyOptions', () => {
		//success
		expect(buildCLI(['--globby-options={"dot": true}']).flags.globbyOptions).toBe('/path/to/dir');
		expect(buildCLI(['--go={"dot": true}']).flags.globbyOptions).toBe('/path/to/dir');
		
		//failure
		expect(buildCLI(['--globby-options={"dot": foo}']).flags.globbyOptions).toBe(
			expect(process.exit).toHaveBeenCalledWith(0);
			expect(console.log).toHaveBeenCalledWith(
			   expect.stringContaining('Invalid JSON value'),
		    );
		 );
	  );
	});
```

Is this fine?

ybiquitous: No, please keep the `flags.globbyOptions` case. Let's add new cases to test `process.exitCode`, `process.stdout`, or `process.stderr` like this:

https://github.com/stylelint/stylelint/blob/042d1adee8a15f61d4d3fdf5c2e0e38569303942/lib/__tests__/cli.test.js#L364-L376

sidverma32:
```
`it('output a message when wrong --globby-options provided', async () => {
		await cli([
			`--globby-options=stylelint-test-globby-options`,
			'--config',
			fixturesPath('default-severity-warning.json'),
		]);

		expect(process.exitCode).toBe(2);

		expect(process.stdout.write).toHaveBeenCalledTimes(1);

		//expect((process.stdout.write).toBe('Invalid JSON value provided.');
	});

`
```
as per my understanding from your suggestion, added this code-block . Please guide me how to print "Invalid" message as it was giving me error. (I am beginner in `jest`  )

ybiquitous: The expectation code should be as follows:

```js
expect(process.exitCode).toBe(2);
expect(process.stdout.write).toHaveBeenCalledTimes(0);
expect(process.stderr.write).toHaveBeenCalledTimes(1);
expect(stripAnsi(process.stderr.write.mock.calls[0][0])).toContain(...);
```

lib/cli.js
```diff
@@ -397,6 +407,21 @@ module.exports = async (argv) => {
...
+			optionsBase.globbyOptions = JSON.parse(value);
+		} catch (_) {
+			process.stderr.write(
+				`Invalid option ${red('"--globby-options"')}. The value ${red(
+					`"${value}"`,
+				)} is not valid JSON.${EOL}`,
+			);
+			process.exit(EXIT_CODE_ERROR); // eslint-disable-line no-process-exit
+		}
+	}
```
ybiquitous, line 423: [suggest] I notice just `JSON.parse` is insufficient because a value like `--globby-options "true"` is unintentionally passed. So we need to check whether the parsed value is an object. We have a utility for that purpose: `isPlainObject()`. See below:

https://github.com/stylelint/stylelint/blob/042d1adee8a15f61d4d3fdf5c2e0e38569303942/lib/utils/validateTypes.js#L65

Thus, I suggest the following additionally:

- Extract a function to parse the `--globby-options` value
- Use `isPlainObject()` to check the JSON-parsed value

```js
	if (cli.flags.globbyOptions) {
		try {
			optionsBase.globbyOptions = await parseGlobbyOptions(cli.flags.globbyOptions);
		} catch (error) {
			if (typeof error === 'string') {
				process.stderr.write(`${error}${EOL}`);
				process.exitCode = EXIT_CODE_ERROR;

				return;
			}

			throw error;
		}
	}

// ...

/**
 * @param {string} value
 * @returns {Promise<Record<string, unknown>>}
 */
function parseGlobbyOptions(value) {
	const errorMessage = () =>
		`Invalid option ${red('"--globby-options"')}.` +
		` The value ${red(`"${value}"`)} is not valid JSON object.`;

	let options;

	try {
		options = JSON.parse(value);
	} catch (_) {
		return Promise.reject(errorMessage());
	}

	if (isPlainObject(options)) {
		return Promise.resolve(options);
	}

	return Promise.reject(errorMessage());
}
```

What do you think?

sidverma32: > [suggest] I notice just `JSON.parse` is insufficient because a value like `--globby-options "true"` is unintentionally passed. So we need to check whether the parsed value is an object. We have a utility for that purpose: `isPlainObject()`. See below:
> 
> https://github.com/stylelint/stylelint/blob/042d1adee8a15f61d4d3fdf5c2e0e38569303942/lib/utils/validateTypes.js#L65
> 
> Thus, I suggest the following additionally:
> 
> * Extract a function to parse the `--globby-options` value
> * Use `isPlainObject()` to check the JSON-parsed value
> 
> ```js
> 	if (cli.flags.globbyOptions) {
> 		try {
> 			optionsBase.globbyOptions = await parseGlobbyOptions(cli.flags.globbyOptions);
> 		} catch (error) {
> 			if (typeof error === 'string') {
> 				process.stderr.write(`${error}${EOL}`);
> 				process.exitCode = EXIT_CODE_ERROR;
> 
> 				return;
> 			}
> 
> 			throw error;
> 		}
> 	}
> 
> // ...
> 
> /**
>  * @param {string} value
>  * @returns {Promise<Record<string, unknown>>}
>  */
> function parseGlobbyOptions(value) {
> 	const errorMessage = () =>
> 		`Invalid option ${red('"--globby-options"')}.` +
> 		` The value ${red(`"${value}"`)} is not valid JSON object.`;
> 
> 	let options;
> 
> 	try {
> 		options = JSON.parse(value);
> 	} catch (_) {
> 		return Promise.reject(errorMessage());
> 	}
> 
> 	if (isPlainObject(options)) {
> 		return Promise.resolve(options);
> 	}
> 
> 	return Promise.reject(errorMessage());
> }
> ```
> 
> What do you think?

yeah that's what i trying to point out in last [discussion](https://github.com/stylelint/stylelint/pull/6437#discussion_r1010530350) . 

This looks clean and more transparent approach! Will update it :)

lib/__tests__/cli.test.js
```diff
@@ -361,6 +366,22 @@ describe('CLI', () => {
 		expect(output).toBe('Custom formatter reports 1 warning(s).');
 	});
 
+	it('output a message when wrong --globby-options provided', async () => {
+		await cli([
+			`--globby-options=stylelint-test-globby-options`,
+			'--config',
+			fixturesPath('default-severity-warning.json'),
+			fixturesPath('empty-block.css'),
+		]);
```
ybiquitous, line 375: [suggest] We can refactor the code to make it simpler.

```suggestion
		await cli(['--globby-options=wrong']);
```

lib/__tests__/cli.test.js
```diff
@@ -361,6 +366,22 @@ describe('CLI', () => {
...
+
+		expect(process.exitCode).toBe(2);
+		expect(process.stdout.write).toHaveBeenCalledTimes(0);
+		expect(process.stderr.write).toHaveBeenCalledTimes(1);
+		expect(stripAnsi(process.stderr.write.mock.calls[0][0])).toContain(
+			'Invalid option "--globby-options". The value "stylelint-test-globby-options" is not valid JSON object.',
+		);
+	});
```
ybiquitous, line 383: [suggest] Thanks for adding the case. Additionally, could you add more cases to verify the new feature like this?

```js
it('--globby-options', async () => {
	await cli([
		'--globby-options={"dot":true}',
		'--config',
		fixturesPath('config-block-no-empty.json'),
		fixturesPath('globby-options'),
	]);

	expect(process.stdout.write).toHaveBeenCalledTimes(1);
	expect(process.stdout.write).toHaveBeenCalledWith(expect.stringMatching(/block-no-empty/));
});
```

adding the following file tree:

```tree
lib/__tests__/fixtures/globby-options/
└── .dot/
    └── empty-block.css
```

Also, let's add the new cases using the `cli()` function to the last in `cli.test.js`.

docs/user-guide/usage/cli.md
```diff
@@ -68,6 +68,12 @@ Automatically fix, where possible, problems reported by rules. [More info](optio
 
 Specify the formatter to format your results. [More info](options.md#formatter).
 
+## `--globbyOptions, --go`
+
+CLI flags: `--globby-options, --go`
+
```
ybiquitous, line 74: [suggest] Can you please remove the redundant lines?

```suggestion
```

docs/user-guide/usage/cli.md
```diff
@@ -68,6 +68,12 @@ Automatically fix, where possible, problems reported by rules. [More info](optio
...
+CLI option to traverse dotfiles i.e Hidden folders/files will not be ignored. [More info](options.md#globbyoptions).
```
ybiquitous, line 75: [suggest] How about making the description more generic?

```suggestion
Options in JSON format passed to [globby](https://github.com/sindresorhus/globby). [More info](options.md#globbyoptions).
```

docs/user-guide/usage/options.md
```diff
@@ -139,6 +139,29 @@ CLI flags: `--disable-default-ignores, --di`
...
+1 directory, 2 files
+
+$ npx stylelint '.foo/\*_/_.css'
+
+.foo/a.css
+1:12 ✖ Unexpected unit length-zero-no-unit
+
+stylelint '\*_/_.css' --globby-options "dot: false"
+
+# should match `.foo/a.css` and `.foo/.bar/a.css`
```
ybiquitous, line 163: [suggest] How about simplifying the description like this because we have the detailed description in `node-api.md`?

```suggestion
Options passed to [globby](https://github.com/sindresorhus/globby). [More info](node-api.md#globbyoptions).
```

In the future, it may be better to move the `globbyOptions` description in `node-api.md` into this `options.md`.

lib/cli.js
```diff
@@ -236,6 +239,10 @@ const meowOptions = {
       --allow-empty-input, --aei
 
         When glob pattern matches no files, the process will exit without throwing an error.
+	 
+	  --globby-options, --go 
+		
+	    CLI option to traverse dotfiles i.e Hidden folders files will not be ignored(e.g. stylelint --fix '**/*.css)
```
ybiquitous, line 245: [suggest] How about simplifying the description?

```suggestion
	    Options in JSON format passed to globby.
```

Also, let's remove extra trailing whitespace and convert TABs to whitespace. If doing so, we can get less diff in `cli.test.js.snap`.

ybiquitous (changes requested):
Thanks for addressing the reviews! With a little more brushing up, I believe this PR is ready to merge.

lib/__tests__/cli.test.js
sidverma32: after refactoring this , do we need to update the "expect" code also at L-377 to L-381? @ybiquitous

lib/__tests__/cli.test.js
sidverma32: Getting error, test cases are failing when modifying conditions. Do let me know if i have to modify anything further! [tried many things but couldn't resolve this]
Screenshot 2022-11-07 at 4 59 59 PM

lib/__tests__/cli.test.js
ybiquitous: Yes, please also update the line to change the message expectation:

```diff
-'Invalid option "--globby-options". The value "stylelint-test-globby-options" is not valid JSON object.'
+'Invalid option "--globby-options". The value "wrong" is not valid JSON object.'
```

lib/__tests__/cli.test.js
ybiquitous: Umm..., could you please push your commits (you can leave it as failures)? I'd like to look into the changed test code.

ybiquitous: Thanks for pushing the commit. Can you update `empty-block.css` as follows?

```diff
diff --git a/lib/__tests__/fixtures/globby-options/.dot/empty-block.css b/lib/__tests__/fixtures/globby-options/.dot/empty-block.css
index e69de29bb..077f6dd7c 100644
--- a/lib/__tests__/fixtures/globby-options/.dot/empty-block.css
+++ b/lib/__tests__/fixtures/globby-options/.dot/empty-block.css
@@ -0,0 +1 @@
+a {}
```

The `a {}` CSS code violates the `block-no-empty` rule.

sidverma32: my bad.. sure updated!

ybiquitous (changes requested):
Thank you for improving the code. We are very close to the merge.

lib/cli.js
```diff
@@ -236,6 +239,10 @@ const meowOptions = {
       --allow-empty-input, --aei
 
         When glob pattern matches no files, the process will exit without throwing an error.
+	 
+	  --globby-options, --go 
+		
+		Options in JSON format passed to globby.
```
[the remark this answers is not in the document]
ybiquitous: I've fixed it via 9913d853.

ybiquitous (approved):
Thank you. LGTM 👍🏼

sidverma32:
> Thank you. LGTM 👍🏼

Thanks a lot for the guidance and huge learning experience for me .  I think this feature roll-out will give end user's a better option for dot files :)

silversonicaxel:
amazing guys

silversonicaxel:
hello everyone,
I updated in my project the stylelint libs to fix my hidden folders linting issues.
I made some tests in a `Storybook` folder where the application lives.

I just created a linting issue in two files that are not in hidden folders.

with `--go` and `--globby-options` applied to stylelint all works still as expected.



Then I created a linting issue in two files that are in hidden folders in a secondary level.

`apps/foundation-storybook/.storybook/sb-theme/styles/sb-sidebar.css`

and

`apps/foundation-storybook/.storybook/sb-theme/styles/sb-sidebar.scss`

but also in primary level

`.test/sb-sidebar.css`

and

`.test/sb-sidebar.css`

and here is the issue

```
#storybook-explorer-tree .sidebar-item {
  padding-bottom: 8px;
  font-size: 16px;
  padding-top: 8px;
}
```
where `font-size` should be first css property and `padding-bottom` the second one.

and these are the commands of `package.json`

```
    "lint:scss": "stylelint --fix '**/*.scss' --globby-options",
    "lint:css": "stylelint --fix '**/*.css' --globby-options",
```

and 

```
    "lint:scss": "stylelint --fix '**/*.scss' --go",
    "lint:css": "stylelint --fix '**/*.css' --go",
```

but any fix is done.

I am not sure if this is still intended, but I fear something is not working as expected.
any opinions, ideas about it?

thanks

ybiquitous:
@silversonicaxel Did you pass a value to the `--globby-options` flag like this?

```
stylelint --fix '**/*.scss' --globby-options '{"dot":true}'
```

silversonicaxel:
Indeed my bad, it works, thanks a lot

PR #6449, opened 2022-11-03
[Fix] Support `import-lazy` bundler format

### LINKED ISSUE #6440: Unable to bundle `stylelint` using bundlers

Use `rollup` to bundle this library. It fails

_None_

I am trying to bundle stylelint in my project

14.14.0

Stylelint should get bundled, without any errors.

### What actually happened?

Getting `Could not dynamically require "./color-no-invalid-hex"` when using `color-no-invalid-hex` in the project.

_None_

### Proposal to fix the bug

- PR that introduced dynamic requires in the `stylelint`: https://github.com/stylelint/stylelint/pull/4729
- PR that fixed typings due to dynamic requires: https://github.com/stylelint/stylelint/pull/5878

Since the latter does not use the [`import-lazy` pattern for bundlers](https://github.com/sindresorhus/import-lazy#usage-with-bundlers), bundlers are not able to resolve the require paths and fails during runtime.

## Possible Resolution:

- We can keep the `import-lazy` pattern for bundlers intact, i.e. to revert back to using `importLazy(() => require('./color-no-invalid-hex'))()` pattern
- To fix usages of // @ts-nocheck, we can do something like following instead:

```js
/** @type {(fn: () => any) => (() => any)} */
// @ts-ignore
const importLazy = require('import-lazy');
...
  'color-no-invalid-hex': importLazy(() => require('./color-no-invalid-hex'))(),
```

I am not sure how strict this project is in terms of using `@ts-ignore` comments, but I feel this is a decent trade-off to support proper bundling, unless someone has a better solution to overcome typescript type errors in `rules` module.

phoenisx:
Closes #6440


ybiquitous (commented):
@phoenisx Thanks for opening the pull request. In addition, could you also fix `lib/formatters/index.js`?

https://github.com/stylelint/stylelint/blob/42bed8e58fc933ebd201523045c8065304ff1097/lib/formatters/index.js#L3

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #6450, opened 2022-11-03
Refactor testing CI config

ybiquitous:

None.

This change reduces duplication in the `if` block. Also, it uses Node.js 18 LTS for coverage.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -44,16 +44,21 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Enable test coverage
+        id: coverage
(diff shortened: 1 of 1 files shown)


.github/workflows/testing.yml
```diff
@@ -44,16 +44,21 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Enable test coverage
+        id: coverage
+        if: ${{ startsWith(matrix.os, 'ubuntu') && matrix.node == '18' }} # Node.js LTS
+        run: echo "enabled=true" >> "${GITHUB_OUTPUT}"
```
ybiquitous, line 50: See https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

PR #6460, opened 2022-11-09
Add support to `checkAgainstRule` with custom rules

### LINKED ISSUE #6446: Add support to `checkAgainstRule` with custom rules

Currently, the `checkAgainstRule` utility [only resolves built-in rules](https://github.com/stylelint/stylelint/blob/042d1adee8a15f61d4d3fdf5c2e0e38569303942/lib/utils/checkAgainstRule.js#L30). However, my team has a use case in which we'd like to call `checkAgainstRule` with other (local) custom rules e.g.

```js
// config
module.exports = {
  plugins: ['./rules/main-rule', './rules/custom-rule-1', './rules/custom-rule-2'],
  rules: { 'scope/main-rule': settings },
}
// main-rule.js
module.exports = createPlugin(ruleName, (primary) => {
  return (root, result) => {
    stylelint.utils.checkAgainstRule({
      ruleName: 'scope/custom-rule-1', // Error: Rule 'scope/custom-rule-1' does not exist
      ruleSettings: customSettings,
    })
  }
})
```

We'd like to be able to use `checkAgainstRule` with (local) custom rules. Being that all `pluginFunctions` are accessible on `result.stylelint.config.pluginFunctions`, this update seems like it would be a non-breaking change by optionally passing the `result` object to [checkAgainstRule](https://github.com/stylelint/stylelint/blob/042d1adee8a15f61d4d3fdf5c2e0e38569303942/lib/utils/checkAgainstRule.js#L30) e.g.

```js
// main-rule.js
module.exports = createPlugin(ruleName, (primary) => {
  return (root, result) => {
    stylelint.utils.checkAgainstRule({
      result,
      ruleName: 'scope/custom-rule-1', // Resolves and checks against custom rules
      ruleSettings: customSettings,
    })
  }
}

// stylelint/checkAgainstRule.js
const rule = options.result.stylelint.config?.pluginFunctions[options.ruleName] || rules[options.ruleName];

if (!rule) throw new Error(`Rule '${options.ruleName}' does not exist`);
```

aaronccasanova:
Closes #6446

- Added support to `checkAgainstRule` with custom rules
- Added test cases to validate passing/failing custom rules and a missing `result` option
- Updated the internal `normalizeRuleSettings` utility to handle custom rules
  - This utility is tested thoroughly and thus I feel very confident about the refactor
- Updated the `checkAgainstRule` documentation to describe the new behavior

### DIFF
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -232,13 +232,14 @@ Validates the options for your rule.
 
 ### `stylelint.utils.checkAgainstRule`
 
-Checks CSS against a standard Stylelint rule _within your own rule_. This function provides power and flexibility for plugins authors who wish to modify, constrain, or extend the functionality of existing Stylelint rules.
+Checks CSS against a standard or custom Stylelint rule _within your own rule_. This function provides power and flexibility for plugins authors who wish to modify, constrain, or extend the functionality of existing Stylelint rules.
 
 It accepts an options object and a callback that is invoked with warnings from the specified rule. The options are:
 
 - `ruleName`: the name of the rule you are invoking
 - `ruleSettings`: settings for the rule you are invoking
 - `root`: the root node to run this rule against
+- `result?`: the postcss result for resolving and invoking custom rules
 
 Use the warning to create a _new_ warning _from your plugin rule_ that you report with `stylelint.utils.report`.
 
@@ -259,7 +260,8 @@ function myPluginRule(primaryOption, secondaryOptionObject) {
       {
         ruleName: "at-rule-no-unknown",
         ruleSettings: [primaryOption, defaultedOptions],
-        root: postcssRoot
+        root: postcssRoot,
+        result: postcssResult
       },
       (warning) => {
         stylelint.utils.report({
--- a/lib/normalizeRuleSettings.js
+++ b/lib/normalizeRuleSettings.js
@@ -31,7 +31,9 @@ module.exports = function normalizeRuleSettings(rawSettings, rule) {
 	}
 	// Everything below is an array ...
 
-	if (rawSettings.length > 0 && (rawSettings[0] === null || rawSettings[0] === undefined)) {
+	const [primary, secondary] = rawSettings;
+
+	if (rawSettings.length > 0 && (primary === null || primary === undefined)) {
 		return null;
 	}
 
@@ -42,11 +44,11 @@ module.exports = function normalizeRuleSettings(rawSettings, rule) {
 	// (they might also have something else, e.g. rule-properties-order can
 	// have the string "alphabetical")
 
-	if (rawSettings.length === 1 && Array.isArray(rawSettings[0])) {
+	if (rawSettings.length === 1 && Array.isArray(primary)) {
 		return rawSettings;
 	}
 
-	if (rawSettings.length === 2 && !isPlainObject(rawSettings[0]) && isPlainObject(rawSettings[1])) {
+	if (rawSettings.length === 2 && !isPlainObject(primary) && isPlainObject(secondary)) {
 		return rawSettings;
 	}
 
--- a/lib/normalizeRuleSettings.js
+++ b/lib/normalizeRuleSettings.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const rules = require('./rules');
 const { isPlainObject } = require('./utils/validateTypes');
 
 // Rule settings can take a number of forms, e.g.
@@ -19,17 +18,10 @@ const { isPlainObject } = require('./utils/validateTypes');
  * @template T
  * @template {Object} O
  * @param {import('stylelint').ConfigRuleSettings<T, O>} rawSettings
- * @param {string} ruleName
- * @param {boolean} [primaryOptionArray] If primaryOptionArray is not provided, we try to get it from the rules themselves, which will not work for plugins
+ * @param {import('stylelint').Rule<T, O>} [rule]
  * @return {[T] | [T, O] | null}
  */
-module.exports = function normalizeRuleSettings(
-	rawSettings,
-	ruleName,
-	// If primaryOptionArray is not provided, we try to get it from the
-	// rules themselves, which will not work for plugins
-	primaryOptionArray,
-) {
+module.exports = function normalizeRuleSettings(rawSettings, rule) {
 	if (rawSettings === null || rawSettings === undefined) {
 		return null;
 	}
@@ -43,15 +35,7 @@ module.exports = function normalizeRuleSettings(
 		return null;
 	}
 
-	if (primaryOptionArray === undefined) {
-		const rule = rules[ruleName];
-
-		if (rule && 'primaryOptionArray' in rule) {
-			primaryOptionArray = rule.primaryOptionArray;
-		}
-	}
-
-	if (!primaryOptionArray) {
+	if (rule && !rule.primaryOptionArray) {
 		return rawSettings;
 	}
 	// Everything below is a rule that CAN have an array for a primary option ...
--- a/lib/utils/checkAgainstRule.js
+++ b/lib/utils/checkAgainstRule.js
@@ -2,8 +2,8 @@
 
 const normalizeRuleSettings = require('../normalizeRuleSettings');
 const Result = require('postcss/lib/result');
-const rules = require('../rules');
 const { isPlainObject } = require('./validateTypes');
+const getStylelintRule = require('./getStylelintRule');
 
 /**
  * Useful for third-party code (e.g. plugins) to run a PostCSS Root
@@ -28,12 +28,7 @@ function checkAgainstRule(options, callback) {
 
 	if (!ruleName) throw new Error("checkAgainstRule requires a 'ruleName' option");
 
-	const rule =
-		rules[ruleName] ||
-		(result &&
-			result.stylelint.config &&
-			result.stylelint.config.pluginFunctions &&
-			result.stylelint.config.pluginFunctions[ruleName]);
+	const rule = getStylelintRule(ruleName, result && result.stylelint.config);
 
 	if (!rule) throw new Error(`Rule '${ruleName}' does not exist`);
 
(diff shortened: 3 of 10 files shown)


ybiquitous:
@aaronccasanova Thanks for creating the pull request.

> Should this PR include updates to the [checkAgainstRule](https://stylelint.io/developer-guide/plugins#stylelintutilscheckagainstrule) documentation describing the new behavior?

Yes, please.

lib/normalizeRuleSettings.js
```diff
@@ -19,17 +18,10 @@ const { isPlainObject } = require('./utils/validateTypes');
...
  * @return {[T] | [T, O] | null}
  */
-module.exports = function normalizeRuleSettings(
-	rawSettings,
-	ruleName,
-	// If primaryOptionArray is not provided, we try to get it from the
-	// rules themselves, which will not work for plugins
-	primaryOptionArray,
-) {
+module.exports = function normalizeRuleSettings(rawSettings, rule) {
```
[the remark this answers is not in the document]
ybiquitous: [suggest] Good catch! In addition, could you refactor the function to replace repeated `rawSettings[*]` like this?

```js
// Everything below is an array ...

const [primary, secondary] = rawSettings;

if (rawSettings.length > 0 && (primary === null || primary === undefined)) {
	return null;
}
```

lib/utils/checkAgainstRule.js
```diff
@@ -3,39 +3,45 @@
...
+ * }} options
  * @param {(warning: import('postcss').Warning) => void} callback
  * @returns {void}
  */
 function checkAgainstRule(options, callback) {
-	if (!options)
-		throw new Error(
-			"checkAgainstRule requires an options object with 'ruleName', 'ruleSettings', and 'root' properties",
-		);
+	if (!isPlainObject(options)) throw new Error('checkAgainstRule requires an options object');
```
ybiquitous, line 23: [suggest] Thanks for improving the error messages. In addition, could you refactor them as follows?

- Remove the `checkAgainstRule` word from the messages because an error raised in `checkAgainstRule` is explicit from a stack trace
- Use `"` (double) instead of `'` (single) in the messages for consistency because we often have used `"` in this project

E.g.

```diff
-throw new Error("checkAgainstRule requires a 'ruleName' option");
+throw new Error('require a "ruleName" option');
```

lib/utils/checkAgainstRule.js
```diff
@@ -3,39 +3,45 @@
...
-		ruleSettings: import('stylelint').ConfigRuleSettings<T, O>,
-		root: import('postcss').Root,
-	}} options
+ *   ruleName: string,
+ *   ruleSettings: import('stylelint').ConfigRuleSettings<T, O>,
+ *   root: import('postcss').Root,
+ *   result?: import('stylelint').PostcssResult,
+ * }} options
  * @param {(warning: import('postcss').Warning) => void} callback
  * @returns {void}
```
ybiquitous, line 20: [suggest] I noticed it's possible to reduce type annotations like this. How about the refactoring?

```js
/**
 * Useful for third-party code (e.g. plugins) to run a PostCSS Root
 * against a specific rule and do something with the warnings.
 *
 * @type {typeof import('stylelint').utils.checkAgainstRule}
 */
function checkAgainstRule(options, callback) {
```

The following line also can be simplified:

https://github.com/stylelint/stylelint/blob/35c0cfa244f725a3548e8ba884edf70c2bc67732/lib/utils/checkAgainstRule.js#L52

lib/utils/checkAgainstRule.js
```diff
@@ -3,39 +3,45 @@
...
-	const rule = rules[options.ruleName];
+	if (!ruleName) throw new Error("checkAgainstRule requires a 'ruleName' option");
 
-	if (!rule) throw new Error(`Rule '${options.ruleName}' does not exist`);
+	const rule =
+		rules[ruleName] ||
+		(result &&
+			result.stylelint.config &&
+			result.stylelint.config.pluginFunctions &&
+			result.stylelint.config.pluginFunctions[ruleName]);
```
ybiquitous, line 36: [suggest] I think it's time to reduce the following duplication:

- https://github.com/stylelint/stylelint/blob/35c0cfa244f725a3548e8ba884edf70c2bc67732/lib/normalizeAllRuleSettings.js#L20
- https://github.com/stylelint/stylelint/blob/35c0cfa244f725a3548e8ba884edf70c2bc67732/lib/lintPostcssResult.js#L65-L66

So what about extracting the following utility function (`utils/getStylelintRule.js`)?

```js
/**
 * @param {string} ruleName
 * @param {import('stylelint').Config | undefined} [config]
 * @returns {import('stylelint').Rule | undefined}
 */
module.exports = function getStylelintRule(ruleName, config) {
	return rules[ruleName] || (config && config.pluginFunctions && config.pluginFunctions[ruleName]);
};
```

docs/developer-guide/plugins.md
```diff
@@ -232,13 +232,14 @@ Validates the options for your rule.
...
 
-Checks CSS against a standard Stylelint rule _within your own rule_. This function provides power and flexibility for plugins authors who wish to modify, constrain, or extend the functionality of existing Stylelint rules.
+Checks CSS against a standard or custom Stylelint rule _within your own rule_. This function provides power and flexibility for plugins authors who wish to modify, constrain, or extend the functionality of existing Stylelint rules.
 
 It accepts an options object and a callback that is invoked with warnings from the specified rule. The options are:
 
 - `ruleName`: the name of the rule you are invoking
 - `ruleSettings`: settings for the rule you are invoking
 - `root`: the root node to run this rule against
+- `result?`: the postcss result for resolving and invoking custom rules
```
ybiquitous, line 242: [suggest] Thanks for updating the doc. Additionally, could you fix the word [for consistency](https://github.com/stylelint/stylelint/blob/35c0cfa244f725a3548e8ba884edf70c2bc67732/docs/developer-guide/plugins.md?plain=1#L70-L71)?

```suggestion
- `result?`: the PostCSS result for resolving and invoking custom rules
```

ybiquitous (commented):
Thank you so much for adding good test cases. Also, the refactoring you introduced is so nice.

This PR is good enough, but I suggest additional refactorings to improve the code quality. If possible, please consider addressing my reviews.

lib/utils/checkAgainstRule.js
aaronccasanova: Agreed, there's a clear opportunity to remove the redundancy 👍 Expanding on the suggestion, what do you think of the following messages:
```js
throw new Error('An options object is required')
throw new Error('A callback function is required')
throw new Error('The "ruleName" option is required')
throw new Error('The "ruleSettings" option is required')
throw new Error('The "root" option is required')
```
Or following the rule/report conventions:
```js
throw new Error('Expected an options object')
throw new Error('Expected a callback function')
throw new Error('Expected a "ruleName" option')
throw new Error('Expected a "ruleSettings" option')
throw new Error('Expected a "root" option')
```

lib/utils/checkAgainstRule.js
aaronccasanova: Big fan of this clean up: https://github.com/stylelint/stylelint/pull/6460/commits/339005b8d3072a8b6708387c76ce5d9aec4b7e95

lib/normalizeRuleSettings.js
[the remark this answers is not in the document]
aaronccasanova: That's great! I find that much clearer 👍 https://github.com/stylelint/stylelint/pull/6460/commits/fe3723eea210a7487ccbea650a0764e5fb286f0d

lib/utils/checkAgainstRule.js
ybiquitous: Nice suggestion. I prefer the latter, but both are acceptable. I leave it to you to decide.

lib/utils/checkAgainstRule.js
aaronccasanova: 👀 Good eye! This is much cleaner: https://github.com/stylelint/stylelint/pull/6460/commits/9f600025bc4282df765654c0f22908284bc4210d

lib/utils/checkAgainstRule.js
aaronccasanova: The latter it is! https://github.com/stylelint/stylelint/pull/6460/commits/b5636bd966adeda34c6a9d52b7a2dbc137e0b5b9

lib/utils/checkAgainstRule.js
```diff
@@ -2,40 +2,33 @@
...
+	const { ruleName, ruleSettings, root, result } = options;
 
-	const rule = rules[options.ruleName];
+	if (!ruleName) throw new Error('Expected a "ruleName" option');
 
-	if (!rule) throw new Error(`Rule '${options.ruleName}' does not exist`);
+	const rule = getStylelintRule(ruleName, result && result.stylelint.config);
 
-	if (!options.ruleSettings) throw new Error("checkAgainstRule requires a 'ruleSettings' option");
+	if (!rule) throw new Error(`Rule '${ruleName}' does not exist`);
```
ybiquitous, line 25: [trivial] Could you please also convert `'` to `"` here?

```suggestion
	if (!rule) throw new Error(`Rule "${ruleName}" does not exist`);
```

ybiquitous (commented):
Thanks for the refactoring. The code is now more straightforward and readable. I've left comments for tiny suggestions.

---

Could you add a `minor` changelog entry following the guideline on <https://github.com/stylelint/stylelint/pull/6460#issuecomment-1308098433> like this?

```
Added: support to `checkAgainstRule` with custom rules
```

lib/utils/checkAgainstRule.js
aaronccasanova: Good catch 👍 https://github.com/stylelint/stylelint/pull/6460/commits/6a53fc709c4847a360b66dcce429dd6808a5fb51

ybiquitous (approved):
Thank you. LGTM 👍🏼

aaronccasanova:
Thanks for the reviews and feedback @ybiquitous! I'm very happy with the additional code quality wins we were able to include in this PR 👍

PR #6466, opened 2022-11-11
Add support autofix with `checkAgainstRule`

### LINKED ISSUE #6465: Add support to use `--fix` with `checkAgainstRule`

The `--fix` option does not work with the `checkAgainstRule` utility:

```js
stylelint.utils.checkAgainstRule(
  {
    ruleName: 'length-zero-no-unit'
    ruleSettings: true,
    root,
  },
  (warning) => { /* Don't report if the problem can be automatically fixed */ },
)
```

I would like to be able to pass through the `context.fix` option to `checkAgainstRule`:

```js
module.exports = createPlugin(
  'custom/length-zero-no-unit',
  (primary, secondary, context) => (root) => {
    stylelint.utils.checkAgainstRule(
      {
        ruleName: 'length-zero-no-unit'
        ruleSettings: true,
        fix: context.fix,
        root,
      },
      (warning) => { /* Don't report if the problem can be automatically fixed */ },
    )
  },
})
```

My team currently has this behavior patched with the following updates to `checkAgainstRule`:
https://github.com/stylelint/stylelint/blob/64b022a052eba7d5a6358781d19cdd3f84a2a0b3/lib/utils/checkAgainstRule.js#L47
```diff
- rule(settings[0], /** @type {O} */ (settings[1]), {})(options.root, tmpPostcssResult);
+ rule(settings[0], /** @type {O} */ (settings[1]), {fix: options.fix})(options.root, tmpPostcssResult);
```

aaronccasanova:

Closes #6465 

No, it's self-explanatory.

### DIFF
--- a/.changeset/tricky-wolves-collect.md
+++ b/.changeset/tricky-wolves-collect.md
@@ -2,4 +2,4 @@
 "stylelint": minor
 ---
 
-Added: support to use `--fix` with `checkAgainstRule`
+Added: support autofix with `checkAgainstRule`
--- a/.changeset/tricky-wolves-collect.md
+++ b/.changeset/tricky-wolves-collect.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: support to use `--fix` with `checkAgainstRule`
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -251,7 +251,7 @@ const allowableAtRules = [
   /* .. */
 ];
 
-function myPluginRule(primaryOption, secondaryOptionObject, context) {
+function myPluginRule(primaryOption, secondaryOptionObject, pluginContext) {
   return (postcssRoot, postcssResult) => {
     const defaultedOptions = Object.assign({}, secondaryOptionObject, {
       ignoreAtRules: allowableAtRules.concat(options.ignoreAtRules || [])
@@ -263,7 +263,7 @@ function myPluginRule(primaryOption, secondaryOptionObject, context) {
         ruleSettings: [primaryOption, defaultedOptions],
         root: postcssRoot,
         result: postcssResult,
-        context
+        context: pluginContext
       },
       (warning) => {
         stylelint.utils.report({
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -240,7 +240,7 @@ It accepts an options object and a callback that is invoked with warnings from t
 - `ruleSettings`: settings for the rule you are invoking
 - `root`: the root node to run this rule against
 - `result?`: the PostCSS result for resolving and invoking custom rules
-- `context?`: the PluginContext for the rule you are invoking
+- `context?`: the [context](rules.md#add-autofix) for the rule you are invoking
 
 Use the warning to create a _new_ warning _from your plugin rule_ that you report with `stylelint.utils.report`.
 
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -263,7 +263,7 @@ function myPluginRule(primaryOption, secondaryOptionObject, ruleContext) {
         ruleSettings: [primaryOption, defaultedOptions],
         root: postcssRoot,
         result: postcssResult,
-        context: pluginContext
+        context: ruleContext
       },
       (warning) => {
         stylelint.utils.report({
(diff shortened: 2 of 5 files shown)


.changeset/tricky-wolves-collect.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: support to use `--fix` with `checkAgainstRule`
```
ybiquitous, line 5: [suggest] Since this change is related to not only CLI but also the Node.js API, could you change the wording like this? (I've changed the PR title)

```suggestion
Added: support autofix with `checkAgainstRule`
```

docs/developer-guide/plugins.md
```diff
@@ -240,6 +240,7 @@ It accepts an options object and a callback that is invoked with warnings from t
 - `ruleSettings`: settings for the rule you are invoking
 - `root`: the root node to run this rule against
 - `result?`: the PostCSS result for resolving and invoking custom rules
+- `context?`: the PluginContext for the rule you are invoking
```
ybiquitous, line 243: [suggest] Since `PluginContext` is an internal name, could you change it like this?

Also, what about adding a link to the document that explains the `context` object?

```suggestion
- `context?`: the [context](rules.md#add-autofix) for the rule you are invoking
```

docs/developer-guide/plugins.md
```diff
@@ -250,7 +251,7 @@ const allowableAtRules = [
   /* .. */
 ];
 
-function myPluginRule(primaryOption, secondaryOptionObject) {
+function myPluginRule(primaryOption, secondaryOptionObject, pluginContext) {
```
ybiquitous, line 254: [suggest] I suggest also changing here for consistency:

```suggestion
function myPluginRule(primaryOption, secondaryOptionObject, ruleContext) {
```

docs/developer-guide/plugins.md
```diff
@@ -261,7 +262,8 @@ function myPluginRule(primaryOption, secondaryOptionObject) {
         ruleName: "at-rule-no-unknown",
         ruleSettings: [primaryOption, defaultedOptions],
         root: postcssRoot,
-        result: postcssResult
+        result: postcssResult,
+        context: pluginContext
```
ybiquitous, line 266:
```suggestion
        context: ruleContext
```

types/stylelint/index.d.ts
```diff
@@ -518,6 +518,7 @@ declare module 'stylelint' {
 						ruleSettings: ConfigRuleSettings<T, O>;
 						root: PostCSS.Root;
 						result?: PostcssResult;
+						context?: PluginContext;
```
ybiquitous, line 521: [suggest] It seems we've given an incorrect name. Perhaps `RuleContext` would be better than `PluginContext`.

Then, I'd like to mark `PluginContext` deprecated and add `RuleContext` as an alternative like this. What do you think?

```ts
export type RuleContext = {
	fix?: boolean | undefined;
	newline?: string | undefined;
};

/**
 * @deprecated Use `RuleContext` instead.
 */
export type PluginContext = RuleContext;
```

ybiquitous (changes requested):
@aaronccasanova Thank you for your writing the code quickly. I've left several suggestions, but there are no major problems. 👍🏼

types/stylelint/index.d.ts
aaronccasanova: Good idea, that is much more consistent with the other `Rule*` types:
https://github.com/stylelint/stylelint/blob/d91bb5b779ce6ca69058549d58195a9e82b7c8d1/types/stylelint/index.d.ts#L162-L200

aaronccasanova: Updated here: https://github.com/stylelint/stylelint/pull/6466/commits/856b30f30224c30574660bffd9ceaafc111cfc4a

Is there a protocol for tracking/ensuring the `PluginContext` type is removed in the next major release?

docs/developer-guide/plugins.md
aaronccasanova: Great suggestion! Definitely an improvement to the information architecture and developer experience.

types/stylelint/index.d.ts
ybiquitous: Thanks for improving the code.

> Is there a protocol for tracking/ensuring the `PluginContext` type is removed in the next major release?

No. I consider such a type as `PluginContext` private, but some plugins seem to use the type. So marking deprecated instead of removing seems helpful for plugin authors.

Honestly, I don't think we need to consider changes in internal types *breaking* because they're not a public API and affect only type-checking.

With the next major version v15.0.0 at least, the deprecated type will be removed.

ybiquitous (approved):
Thank you! LGTM 👍🏼

PR #6391, opened 2022-10-03
Add `*-pattern` custom message formatting

ybiquitous:
> Which issue, if any, is this issue related to?

Ref: https://github.com/stylelint/stylelint-config-standard/issues/263

> Is there anything in the PR that needs further explanation?

This PR updates the following rules:

- `custom-media-pattern`
- `custom-property-pattern`
- `keyframes-name-pattern`
- `selector-class-pattern`
- `selector-id-pattern`
- `selector-nested-pattern`

Also, this PR makes the rules' message more unified and simpler like this (see also <https://github.com/stylelint/stylelint/pull/6391#discussion_r986307057> for details):

```diff
- Expected custom media query name "${name}" to match pattern "${pattern}"
+ Expected "${name}" to match pattern "${pattern}"
```

PR #6475, opened 2022-11-14
Add missing link about custom message option to rule docs

ybiquitous:

Follow-up of #6391

This change also adds test cases to prevent such mistakes.


jeddy3 (approved):
Thank you!

PR #6476, opened 2022-11-14
Remove deprecated `PluginContext` type

ybiquitous:

Ref: #6466

The `PluginContext` type has been marked as deprecated in PR #6466.

Note that the base branch of this PR is `v15`.


ybiquitous:
[ask] Should we add a changelog entry for removing only types? I'm concerned that it will not be possible to change internal types in the future easily.

jeddy3 (approved):

PR #6479, opened 2022-11-15
Remove support for processors

### LINKED ISSUE #3408: Remove support for processors
Following on from https://github.com/stylelint/stylelint/issues/3264.

We [added support](https://github.com/stylelint/stylelint/pull/1535) for [processors](https://github.com/stylelint/stylelint/blob/master/docs/developer-guide/processors.md) in `7.0.0`. Processors replaced the existing `extractStyleTagsFromHtml` functionality with something more extensible, because we realised that bundling support for all formats that could embed CSS was unsustainable. 

The autofix feature was added after the processors system was built. As such, the processor system wasn't designed with autofix in mind and this has led to incompatibilities.

Recently, @gucong3000 proposed using [PostCSS's syntax option](https://github.com/postcss/postcss#options) to _extract CSS-like code_ from the likes of CSS-in-JS, markdown fences, `<style>` tags and `style=""` attributes etc. We'd previously only used PostCSS's syntax option for CSS-like languages themselves like Scss and Less, rather than as a means to extract CSS-like code from other formats.

I believe @gucong3000 approach has merit as:

1. it can be used by the whole PostCSS ecosystem (whereas processors are stylelint-specific)
2. it's compatible with stylelint's autofix (unlike processors)

There are [currently four syntaxes](https://github.com/postcss/postcss#syntaxes) that extract CSS-like code:

* postcss-html parsing styles in `<style>` tags of HTML-like files.
* postcss-markdown parsing styles in code blocks of Markdown files.
* postcss-jsx parsing CSS in template / object literals of source files.
* postcss-styled parsing CSS in template literals of source files.

There is a known limitation, though. PostCSS's syntax option only accepts a single syntax, whereas many processors can be used at the same time. I believe @gucong3000 has attempted to work around this limitation with [postcss-syntax](https://github.com/gucong3000/postcss-syntax/). When passed to PostCSS's syntax option, it attempts to infer what actual syntax (e.g. postcss-scss, postcss-styled etc) should be used for each file.

Now that we have a replacement for the processor system, should we deprecate it in favour of supporting just syntaxes? 

---

I'm going to mention @zabute and @emilgoldsmith as authors of the [glamorous](https://github.com/zabute/stylelint-processor-glamorous) and [styled-components](https://github.com/styled-components/stylelint-processor-styled-components) processors, respectively. I hope you both don't mind. I wanted to give you a heads-up of this discussion in case you have any thoughts you want to share. For example, is there something your processors do that isn't covered by these syntaxes? Incidentally, thank you for your work on these processors and for allowing stylelint to be used by people using CSS-in-JS.

ybiquitous:

Closes #3408

Should we keep `developer-guide/processors.md` instead of removing it for archival purposes?


jeddy3 (approved):
LGTM, thank you.

Let's remove the doc.

PR #6481, opened 2022-11-18
Add support for plugin objects as config values

### LINKED ISSUE #6452: Add support for JS functions in `plugins` list in Stylelint config

Using `stylelint` as a library in JS (NodeJS/Deno) does not accept JS functions as `plugins: []` config values. It only works with `string` type for now.

Is it possible to support both?

Allow developers to, when using `stylelint` as JS library, to pass locally developed stylelint plugins as config param. For e.g.

```js
stylelint({
  rules: {
    "private/rule-name": true,
  },
  plugins: [require("./plugin-path")],
})
```

I guess this can be resolved easily if we add a check in `lib/augmentConfig.js#addPluginFunctions` to separate `string` and JS function types as `plugins` passed in the config and run the logic accordingly.

phoenisx:
Closes #6452

Allow users to pass Plugin functions, as well as string names for plugin config.

### DIFF
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -234,6 +234,16 @@ To use one, add a `"plugins"` array to your config, containing "locaters" identi
 - npm module name
 - absolute path
 - path relative to the invoking configuration file
+- stylelint plugin object of format
+
+```js
+{
+  ruleName: "plugin-name",
+  rule: function plugin(opts) { /* Plugin Code */ }
+}
+```
+
+> Read more about writing [Stylint Plugins here](../developer-guide/plugins.md).
 
 Once the plugin is declared, within your `"rules"` object _you'll need to add options_ for the plugin's rule(s), just like any standard rule. Look at the plugin's documentation to know what the rule name should be.
 
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -229,21 +229,11 @@ For example, [stylelint-scss](https://github.com/stylelint-scss/stylelint-scss) 
 
 You'll find more in [awesome stylelint](https://github.com/stylelint/awesome-stylelint#plugins).
 
-To use one, add a `"plugins"` array to your config, containing "locaters" identifying the plugins you want to use. As with `extends`, above, a "locater" can be either a:
+To use one, add a `"plugins"` array to your config, containing either [plugin objects](../developer-guide/plugins.md) or "locaters" identifying the plugins you want to use. As with `extends`, above, a "locater" can be either a:
 
 - npm module name
 - absolute path
 - path relative to the invoking configuration file
-- stylelint plugin object of format
-
-```js
-{
-  ruleName: "plugin-name",
-  rule: function plugin(opts) { /* Plugin Code */ }
-}
-```
-
-> Read more about writing [Stylint Plugins here](../developer-guide/plugins.md).
 
 Once the plugin is declared, within your `"rules"` object _you'll need to add options_ for the plugin's rule(s), just like any standard rule. Look at the plugin's documentation to know what the rule name should be.
 
--- a/lib/__tests__/plugins.test.js
+++ b/lib/__tests__/plugins.test.js
@@ -39,6 +39,14 @@ const configRelativeAndExtendRelative = {
 	},
 };
 
+const configPluginRequire = {
+	plugins: [require('./fixtures/plugin-warn-about-foo')],
+	rules: {
+		'plugin/warn-about-foo': 'always',
+		'block-no-empty': true,
+	},
+};
+
 const processorRelative = postcss().use(
 	stylelint({ config: configRelative, configBasedir: __dirname }),
 );
@@ -52,6 +60,7 @@ const processorRelativeAndExtendRelative = postcss().use(
 		configBasedir: __dirname,
 	}),
 );
+const processorPluginRequire = postcss().use(stylelint({ config: configPluginRequire }));
 
 it('one plugin runs', async () => {
 	const result = await processorRelative.process(cssWithFoo, { from: undefined });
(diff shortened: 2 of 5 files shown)


phoenisx:
Need some help fixing the typing errors. The changes are logically correct, but I am not able to understand how to fix the typing errors. 🤦🏽

phoenisx:
The changes should also work with `mergeConfigs()` if a developer uses the same plugin import since stylelint runs on a single nodejs process.

`PostCSS` instead uses a format to pass an array of objects, which is helpful, because each plugin has a name, and if a plugin-name overlaps, we can exclude the previous one, while merging configs.

Not sure which pattern is preferred here. I feel the latter would provide more flexibility , but would be a breaking change, and we will have to keep the API backwards compatible, just in case.

types/stylelint/index.d.ts
```diff
@@ -7,7 +7,10 @@ declare module 'stylelint' {
 		export type Severity = 'warning' | 'error';
 
 		export type ConfigExtends = string | string[];
-		export type ConfigPlugins = string | string[];
+		export type PluginImportType =
+			| { default?: { ruleName: string; rule: Rule } }
+			| { ruleName: string; rule: Rule };
+		export type ConfigPlugins = string | string[] | PluginImportType | PluginImportType[];
```
ybiquitous, line 13: [suggest] Since the existing `Plugin` type name is inaccurate, I believe we should use the type in different semantics.

So could you please fix the types like this? I confirm the type-check passes locally.

```diff
@@ -6,11 +6,14 @@ declare module 'stylelint' {
 	namespace stylelint {
 		export type Severity = 'warning' | 'error';
 
+		// TODO: This type will be renamed to `Plugin` with the next major version.
+		export type PluginType = {
+			ruleName: string;
+			rule: Rule;
+		};
+
 		export type ConfigExtends = string | string[];
-		export type PluginImportType =
-			| { default?: { ruleName: string; rule: Rule } }
-			| { ruleName: string; rule: Rule };
-		export type ConfigPlugins = string | string[] | PluginImportType | PluginImportType[];
+		export type ConfigPlugins = string | PluginType | Array<string | PluginType>;
 		export type ConfigProcessor = string | [string, Object];
 		export type ConfigProcessors = string | ConfigProcessor[];
 		export type ConfigIgnoreFiles = string | string[];
@@ -207,6 +210,7 @@ declare module 'stylelint' {
 			meta?: RuleMeta;
 		};
 
+		// TODO: This type will change semantically with the next major version. See `PluginType`.
 		export type Plugin = RuleBase;
 
 		export type CodeProcessor = (code: string, file: string | undefined) => string;
@@ -466,7 +470,7 @@ declare module 'stylelint' {
 			/**
 			 * Creates a Stylelint plugin.
 			 */
-			createPlugin: (ruleName: string, rule: Rule) => { ruleName: string; rule: Rule };
+			createPlugin: (ruleName: string, rule: Rule) => PluginType;
 			/**
 			 * Internal use only. Do not use or rely on this method. It may
 			 * change at any time.
```

ybiquitous (changes requested):
@phoenisx Thanks for creating the pull request. I've left a suggestion comment to fix the type error.

ybiquitous (commented):
In addition, could you add test cases about the new feature to `lib/__tests__/plugins.test.js`?

types/stylelint/index.d.ts
phoenisx: Thanks, @ybiquitous 
I fixed the linting errors 🙇🏽

lib/__tests__/plugins.test.js
```diff
@@ -108,6 +117,37 @@ it('config with its own plugins extending another config that invokes a plugin w
 	expect(result.warnings()[0].node).toBeTruthy();
 });
 
+it('plugin with CJS require()', async () => {
+	const result = await processorPluginRequire.process(cssWithFoo, { from: undefined });
+
+	expect(result.warnings()).toHaveLength(2);
+	expect(result.warnings()[0].text).toBe('found .foo (plugin/warn-about-foo)');
```
ybiquitous, line 124:
```suggestion
	expect(result.warnings()[0].text).toBe('found .foo (plugin/warn-about-foo)');
	expect(result.warnings()[1].text).toBe('Unexpected empty block (block-no-empty)');
```

lib/__tests__/plugins.test.js
```diff
@@ -108,6 +117,37 @@ it('config with its own plugins extending another config that invokes a plugin w
...
+						'plugin/warn-about-foo': 'always',
+						'block-no-empty': true,
+					},
+				},
+			}),
+		)
+		.process(cssWithFoo, { from: undefined });
+
+	expect(result.warnings()).toHaveLength(2);
+	expect(result.warnings()[0].text).toBe('found .foo (plugin/warn-about-foo)');
```
ybiquitous, line 147:
```suggestion
	expect(result.warnings()[0].text).toBe('found .foo (plugin/warn-about-foo)');
	expect(result.warnings()[1].text).toBe('Unexpected empty block (block-no-empty)');
```

ybiquitous (commented):
Thanks for adding test cases! Just in case, could you add more expectations, as I suggested?

lib/__tests__/plugins.test.js
phoenisx: Thanks for the suggestions, have updated with the same 👍🏽

ybiquitous (approved):
Thanks, LGTM 👍🏼

phoenisx:
Shall I merge this, if this Looks good to everyone here?

docs/user-guide/configure.md
```diff
@@ -234,6 +234,16 @@ To use one, add a `"plugins"` array to your config, containing "locaters" identi
...
+- stylelint plugin object of format
+
+```js
+{
+  ruleName: "plugin-name",
+  rule: function plugin(opts) { /* Plugin Code */ }
+}
+```
+
+> Read more about writing [Stylint Plugins here](../developer-guide/plugins.md).
```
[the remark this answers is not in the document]
ybiquitous: I also agree with @jeddy3's suggestion. Furthermore, **"plugin objects"** may be more accurate than "plugin functions" because a Stylelint plugin is an object:

https://github.com/stylelint/stylelint/blob/2290f557cb028242d978b06465dc108aa04a0c3a/lib/createPlugin.js#L6-L11

phoenisx:
Sure. I’ll fix this once I am back home 👍

jeddy3 (approved):
LGTM, thanks.

I made the doc change myself as it was easy to do.

ybiquitous (approved):
Thank you. LGTM 👍🏼

ybiquitous:
Oops, a changelog entry is needed.

ybiquitous:
Changelog added via 40a3c1a

ybiquitous:
@phoenisx Thanks for the contribution!

phoenisx:
Thanks to you all for making the change, while I was away 🙇🏽

PR #6497, opened 2022-11-29
Add media-feature-range-notation

### LINKED ISSUE #6161: Add `media-feature-range-notation`

jeddy3:

Closes #6161 

Part of cleaning up our pull requests as we prepare for v15.

As mentioned in https://github.com/stylelint/stylelint/pull/6430#issuecomment-1296797593, this pull request doesn't include autofix. I'll open a new issue for that. CssTools dropped [a new media query parser](https://www.npmjs.com/package/@csstools/media-query-list-parser) the other day that may make implementating autofix more viable.

Additionally, I believe the edge-case of:

```css
@media (min-width: 500px) and screen and (max-width: 1200px) {}
```

Should be transformed by this rule into:

```css
@media (width >= 500px) and screen and (width <= 1200px) {}
```

That could then be merged into:

```css
@media screen and (500px <= width <= 1200px) {}
```

By another rule, one that sets the notation of media features in a range context. For example, `media-feature-range-context-notation`: `"separate"|"combined"`. I'll create another issue to discuss this, though. I wanted to mention it here to give context to this pull request.

There are [lots of terms](https://drafts.csswg.org/mediaqueries/#evaluating) in media queries. I've tried to get them right in the rule logic and test descriptions. Casting a second pair of eyes over these would be great.

### DIFF
--- a/lib/rules/media-feature-range-notation/__tests__/index.js
+++ b/lib/rules/media-feature-range-notation/__tests__/index.js
@@ -1,0 +1,177 @@
+'use strict';
+
+const stripIndent = require('common-tags').stripIndent;
+
+const { messages, ruleName } = require('..');
+
+testRule({
+	ruleName,
+	config: ['context'],
+
+	accept: [
+		{
+			code: '@media {}',
+			description: 'empty media query',
+		},
+		{
+			code: '@media () {}',
+			description: 'empty media feature',
+		},
+		{
+			code: '@media screen {}',
+			description: 'keyword',
+		},
+		{
+			code: '@media (color) {}',
+			description: 'range type media feature in boolean context',
+		},
+		{
+			code: '@media (color > 0) {}',
+			description: 'range type media feature in non-boolean context',
+		},
+		{
+			code: '@media (pointer: fine) {}',
+			description: 'discrete type media feature',
+		},
+		{
+			code: '@media (width >= 1px) {}',
+			description: 'range type media feature in context notation',
+		},
+		{
+			code: '@media screen and (width >= 1px) {}',
+			description: 'range type media feature in context notation with keyword',
+		},
+		{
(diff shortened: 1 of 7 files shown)


lib/rules/media-feature-range-notation/__tests__/index.js
```diff
@@ -0,0 +1,177 @@
...
+			code: '@media not print, (width >= 1px) {}',
+			description: 'range type media feature in context notation in media query list',
+			message: messages.expected('prefix'),
+			line: 1,
+			column: 19,
+			endLine: 1,
+			endColumn: 33,
+		},
+		{
+			code: '@media (1px < width >= 2px) {}',
```
ybiquitous, line 168: [trivial] Same here:

```suggestion
			code: '@media (1px < width <= 2px) {}',
```

ybiquitous (commented):
Thanks for creating the pull request. I've left some trivial suggestions, but there are no big problems. 👍🏼

ybiquitous (approved):
Thank you. LGTM 👍🏼 

Let's add a changelog entry before merging the PR!

jeddy3:
> I've left some trivial suggestions, but there are no big problems

All good catches, thanks.

I've added a changelog entry.

ybiquitous (approved):

PR #6503, opened 2022-12-03
Change type names to be more consistent

ybiquitous:

Follow-up for #6481

The `Plugin` type should be used as a return type of the `createPlugin()` function.

Type-checking failures may occur if people use the types changed in this PR, but I believe they accept the failures because of the major update (v15).

### DIFF
--- a/.changeset/early-penguins-know.md
+++ b/.changeset/early-penguins-know.md
@@ -1,0 +1,5 @@
+---
+"stylelint": major
+---
+
+Changed: type names to be more consistent
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -7,10 +7,10 @@ declare module 'stylelint' {
 		export type Severity = 'warning' | 'error';
 
 		export type ConfigExtends = string | string[];
-		export type PluginType =
+		export type Plugin =
 			| { default?: { ruleName: string; rule: Rule } }
 			| { ruleName: string; rule: Rule };
-		export type ConfigPlugins = string | PluginType | (string | PluginType)[];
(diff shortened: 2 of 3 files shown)


jeddy3 (approved):
LGTM, thank you!

> Type-checking failures may occur if people use the types changed in this PR

I've added a changelog entry to reflect this.

.changeset/early-penguins-know.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": major
```
ybiquitous, line 2: I doubt if only type definition change should be considered as a breaking. Such a change does not break anything about runtime behavior but may raise a type-check error.

Do you have any (semver) guidelines about such a change?

jeddy3: > Do you have any (semver) guidelines about such a change?

We don't. Is there a defacto approach for packages that publish types that we could adopt? 

If not, let's just switch back to refactor like you originally had it.

ybiquitous: > Is there a defacto approach for packages that publish types that we could adopt?

I'm not sure if it exists or not... 🤷🏼

ybiquitous: When I googled, the following website is found:

Semantic Versioning for TypeScript Types - https://www.semver-ts.org/

ybiquitous: When I read semver-ts.org, I feel a bit too strict, honestly.

However, aside from the future, because it's obvious that there's a possibility of breaking anything, I have no objections to considering this PR as breaking. 👍🏼

PR #6504, opened 2022-12-04
Deprecate stylistic rules handled by Prettier

### LINKED ISSUE #6342: Deprecate stylistic rules

ybiquitous:

Closes #6342

I've created a one-time script to rewrite a lot of rule files.

Usage:

```shell
node scripts/deprecate-stylistic-rules-handled-by-prettier.mjs
```


lib/rules/at-rule-name-case/README.md
```diff
@@ -1,5 +1,7 @@
 # at-rule-name-case
 
+> **Warning** This rule is deprecated and will be removed in the future.
+
```
ybiquitous, line 4: [ask] I believe the warning message is helpful, but what do you think?

---

[note] GitHub renders the `> **Warning**` text as follows:

image

On the other hand, Docusaurus has a different rendering syntax called [Admonitions](https://docusaurus.io/docs/markdown-features/admonitions). E.g.

```markdown
:::caution

This rule is deprecated and will be removed in the future.

:::
```

If using Admonitions, we need to rewrite the warning syntax on the stylelint.io repo.

scripts/deprecate-stylistic-rules-handled-by-prettier.mjs
```diff
@@ -0,0 +1,47 @@
+// TODO: This script is one-time. Please delete it if no longer needed.
```
ybiquitous, line 1: - [ ] Delete this script before merging this PR

docs/migration-guide/to-15.md
```diff
@@ -1,12 +1,65 @@
...
+Additionally, you may no longer need to extend [Prettier's Stylelint config](https://www.npmjs.com/package/stylelint-config-prettier) as there should be no conflicting rules:
+
+```diff json
+{
+- "extends": ["stylelint-config-prettier"],
+  "rules" { .. }
+}
+```
+
+If you want to continue using Stylelint to enforce stylistic consistency, you can [migrate the deprecated rules you need to a plugin](../developer-guide/plugins.md). We will remove the rules from Stylelint in the next major release.
```
jeddy3, line 39: [Note] This is at the bottom as it's not the recommended route. The rules are problematic and a plugin would suffer the same issues:

- old code reliant on stylesearch
- not cleanly working with the AST
- introducing new lint problems
- a confusing number of configuration options

docs/user-guide/get-started.md
```diff
@@ -29,18 +29,6 @@ npx stylelint "**/*.css"
 
 _You should include quotation marks around file globs._
 
-If you use a pretty printer alongside Stylelint, you should turn off any conflicting rules. For example, you can use [Prettier's shared config](https://www.npmjs.com/package/stylelint-config-prettier) to do that:
```
jeddy3, line 32: [Note] We don't need to suggest using the config anymore as the configs won't conflict.

jeddy3 (commented):
@ybiquitous This is fantastic, thank you!

I've added a commit to add a migration entry. Although, not a breaking change it's a significant one that we need to walk our users through.

lib/rules/at-rule-name-case/README.md
jeddy3: > I believe the warning message is helpful, but what do you think?

LGTM

> we need to rewrite the warning syntax on the stylelint.io repo.

Yes, but only if it's quick to do as:

- I think it'll look good enough on the website
- we'll be removing the rules in the next major release

ybiquitous: Agree. It looks like we don't need to add an extra process.

docs/migration-guide/to-15.md
```diff
@@ -1,12 +1,65 @@
...
+Stylelint and pretty printers are complementary tools that work together to help you write consistent and error-free code. To help facilitate this, we've removed the [deprecated rules](../user-guide/rules.md#deprecated) from the latest version of [our standard config](https://www.npmjs.com/package/stylelint-config-standard). You can [extend](../user-guide/configure.md#extends) the config using:
+
+```diff json
+{
++ "extends": ["stylelint-config-standard"],
+  "rules" { .. }
+}
+```
+
+[Our standard config](https://www.npmjs.com/package/stylelint-config-standard) turns on many of the [other rules that enforce conventions](../user-guide/rules.md#enforce-conventions), e.g. most of the [`*-notation`](../user-guide/rules.md#notation), [`*-pattern`](../user-guide/rules.md#pattern) and [`*-quotes`](../user-guide/rules.md#quotes) rules. We recommend adding more of [the rules that enforce conventions](../user-guide/rules.md#enforce-conventions) to your config as many of them will be specific to your needs, e.g. what [units you allow](../../lib/rules/unit-allowed-list/README.md) in your code.
```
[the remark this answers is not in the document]
ybiquitous: > I suggest we mark the rules in our list that are included in our configs.

Sounds good. Do you already have an idea of how to realize it?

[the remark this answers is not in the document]
jeddy3: > Do you already have an idea of how to realize it?

I figured we could just whack an "R" and/or "S" on the end of the rule description and throw a key at the top. Something along the lines of:


(R & S) = In our recommended and standard configs
(S) = In our standard config

- [`at-rule-no-unknown`](https://stylelint.io/user-guide/rules/at-rule-no-unknown): Disallow unknown at-rules (R).
- [`selector-attribute-quotes`](https://stylelint.io/user-guide/rules/selector-attribute-quotes): Require or disallow quotes for attribute values (Autofixable) (R & S).

What do you think? 

(We can do it in a follow-up pull request.)

scripts/deprecate-stylistic-rules-handled-by-prettier.mjs
ybiquitous: Deleted the script via 23a4e758d

ybiquitous (commented):
@jeddy3 Thank you for writing up the documents. LGTM.

jeddy3 (approved):
LGTM too

I think we can merge this.

docs/migration-guide/to-15.md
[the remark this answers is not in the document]
Mouvedia: Ⓡ and Ⓢ

lib/rules/block-closing-brace-newline-after/index.js
```diff
@@ -23,6 +23,7 @@ const messages = ruleMessages(ruleName, {
 const meta = {
 	url: 'https://stylelint.io/user-guide/rules/block-closing-brace-newline-after',
 	fixable: true,
+	deprecated: true,
```
Mouvedia, line 26: I wanted to check whether all the rules that were being deprecated were properly covered by prettier.

This one has [options](https://github.com/stylelint/stylelint/blob/main/lib/rules/block-closing-brace-newline-after/README.md#options):
- "always"
- "always-single-line"
- "never-single-line"
- "always-multi-line"
- "never-multi-line"

Are all of these properly covered by prettier?
If not, we need to make a list of rules that are being deprecated but are only partially covered by the prettier replacement.
e.g.

name | support | note
--- | --- | ---
rule A | full |
rule B | partial | does not support option X and Y
rule C | partial | only support option Z

nolimitdev: @Mouvedia was please suggested list created? To see which deprecated rule can be covered by the prettier replacement? Thanks.

Mouvedia: @jeddy3 @ybiquitous î

ybiquitous: Ideally, we might make a list, but we need more resources. Also, it may not be precious since the rules will be removed in the next major update.

However, we welcome your contribution if you have time.

jeddy3: Following on from https://github.com/stylelint/stylelint/issues/6657#issuecomment-1427969428, I suggest the people working on the plugin pack draw up the list.

As @ybiquitous said, we've other things to focus on (that will benefit the majority of users). I suspect most people are already use a pretty printer alongside Stylelint, and most of those who don't will soon as it's a better solution.

Anyone wanting to continue using Stylelint for formatting pool work can together on a community plugin pack. As with custom syntaxes, the health of the plugin will be determined by the health of the community that maintains it.

PR #6510, opened 2022-12-06
Reorganize type definitions

### LINKED ISSUE #6508: Reorganize public type definitions
Because it is unclear whether a type definition is public or internal and so available from Stylelint plugin packages, there is a problem in that we cannot safely refactor type definitions. 

There are a few ways to limit the publicity of type definitions:

1. Add the `@internal` JSDoc annotation and a warning message like this:
    https://github.com/stylelint/stylelint/blob/11acf312191e5097692f7428f13977bd1fd12cd4/types/stylelint/index.d.ts#L539-L544
2. Add a `_` prefix to types or type members like this:
    https://github.com/stylelint/stylelint/blob/11acf312191e5097692f7428f13977bd1fd12cd4/types/stylelint/index.d.ts#L544-L550
3. Remove `export` from internal types as possible. This way is smarter, but it may need to remove `namespace`:
    https://github.com/stylelint/stylelint/blob/11acf312191e5097692f7428f13977bd1fd12cd4/types/stylelint/index.d.ts#L6

I believe we need to reorganize type definitions by combining such ways well.
I'm happy to hear any feedback.

Ref:
- https://github.com/stylelint/stylelint/pull/6415#issuecomment-1337682121
- https://github.com/stylelint/stylelint/pull/6503#discussion_r1039551312
- [The TypeScript Declaration Files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html)

ybiquitous:

Closes #6508

- remove unused types
- stop using `namespace` to prevent internal types from being published
- mark `@internal` to internal types

Note that I've marked `@internal` conservatively when a type is unknown whether it's internal or not. If an *internal* type proves to be inconvenient or there are requests from people, let's consider removing `@internal` from the type.

### DIFF
--- a/lib/resolveConfig.js
+++ b/lib/resolveConfig.js
@@ -6,17 +6,7 @@ const createStylelint = require('./createStylelint');
 const getConfigForFile = require('./getConfigForFile');
 
 /**
- * Resolves the effective configuration for a given file. Resolves to `undefined`
- * if no config is found.
- * @param {string} filePath - The path to the file to get the config for.
- * @param {Pick<
- *   import('stylelint').LinterOptions,
- *   | 'cwd'
- *   | 'config'
- *   | 'configBasedir'
- *   | 'configFile'
- * >} options - The options to use when creating the Stylelint instance.
- * @returns {Promise<import('stylelint').Config | undefined>}
+ * @type {import('stylelint').default['resolveConfig']}
  */
 module.exports = async function resolveConfig(
 	filePath,
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -1,541 +1,620 @@
 declare module 'stylelint' {
 	import type * as PostCSS from 'postcss';
 	import type { GlobbyOptions } from 'globby';
-	import type { cosmiconfig } from 'cosmiconfig';
+	import type { cosmiconfig, TransformSync as CosmiconfigTransformSync } from 'cosmiconfig';
 
-	namespace stylelint {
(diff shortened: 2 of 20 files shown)


types/stylelint/index.d.ts
```diff
@@ -1,541 +1,620 @@
 declare module 'stylelint' {
```
ybiquitous, line 1: [note] The deletion of `namespace` causes indent changes in many lines. I recommend the diff by [hiding whitespace mode](https://github.com/stylelint/stylelint/pull/6510/files?w=1).

lib/resolveConfig.js
```diff
@@ -6,17 +6,7 @@ const createStylelint = require('./createStylelint');
...
- * @param {string} filePath - The path to the file to get the config for.
- * @param {Pick<
- *   import('stylelint').LinterOptions,
- *   | 'cwd'
- *   | 'config'
- *   | 'configBasedir'
- *   | 'configFile'
- * >} options - The options to use when creating the Stylelint instance.
- * @returns {Promise<import('stylelint').Config | undefined>}
+ * @type {import('stylelint').default['resolveConfig']}
```
ybiquitous, line 9: [note] To reduce duplication, I've moved the description in JSDoc to the `.d.ts` file.

jeddy3 (approved):
Thank you @ybiquitous !

I'm not very familiar with our types, but from what I can tell LGTM.

ybiquitous:
Thanks for the review! 😄

PR #6511, opened 2022-12-06
Add declaration-property-value-no-unknown

### LINKED ISSUE #6376: Add `declaration-property-value-no-unknown`

```
body {
  padding: auto;
}
```

```
{
  "extends": [
    "stylelint-config-standard",
    "stylelint-config-prettier"
  ],
  "rules": {
    "block-closing-brace-newline-after": [
      "always", {
        "ignoreAtRules": [ "if", "else" ]
      }
    ],
    "at-rule-no-unknown": [
      true,
      {
        "ignoreAtRules": [
          "mixin",
          "if",
          "else",
          "error",
          "include",
          "at-root",
          "extend"
        ]
      }
    ]
  },
  "property-no-unknown": true,
  "defaultSeverity": "warning"
}
```

From VSCode

`14.13.0`

The editor should tell me that 'auto' is not a permitted value for property 'padding'.

### What actually happened?

Nothing, the error is not reported.

Nope, tryed in a plain CSS

### Proposal to fix the bug

_No response_

jeddy3:

Closes #6376

I figured it'd be a good idea to include this rule in `v15`. It is a significant change in our direction to offer more validation rules and a change that counterweighs our decision to deprecate most of the stylistic rules. It'll be useful for people to see that in our migration guide. Additionally, the rule is doing a lot and will flag problems for people using values not defined in the CSS specifications. We can help these people by including an entry in the migration guide 

As discussed in https://github.com/stylelint/stylelint/issues/6376#issuecomment-1279825153, the scoped is limited to checking for mismatched property-value pairs.

Most test cases come from issues people have opened in our tracker over the years. 

I've marked as draft as I'm unsure about two things:
- the `is-plain-object` import
- a TypeScript error

```
lib/rules/declaration-property-value-no-unknown/index.js:4:9 - error TS2305: Module '"css-tree"' has no exported member 'fork'.

4 const { fork, parse } = require('css-tree');
          ~~~
Found 1 error in lib/rules/declaration-property-value-no-unknown/index.js:4
```

Any advice?


ybiquitous:
@jeddy3 I've pushed a `@types/css-tree` package patch to fix the type errors using [`patch-package`](https://www.npmjs.com/package/patch-package), although I'm unsure whether `fork()` is a public API and the patched types are correct.

lib/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -0,0 +1,111 @@
+'use strict';
+
+const { isPlainObject } = require('is-plain-object');
+const { fork, parse } = require('css-tree');
```
alexander-akait, line 4: I'm afraid this might well increase the package size, not agains this rule, just intresting how much. Also I think grammar maching can be bundled (and tree shaken) to solve this problem

lib/rules/declaration-property-value-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,225 @@
...
+		},
+		{
+			code: 'a { top: var(--foo); }',
+		},
+		{
+			code: 'a { foo: 1px; }',
+		},
+		{
+			code: 'a { --foo: 1px; }',
+		},
```
Mouvedia, line 27: Could you add
```
		{
			code: 'a { display: -moz-inline-stack; }',
		},
```
as well?

lib/rules/declaration-property-value-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,225 @@
+'use strict';
+
+const { messages, ruleName } = require('..');
```
Mouvedia, line 3: nit

this type of imports will require a codemod for #5291

lib/rules/declaration-property-value-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,225 @@
...
+		{
+			code: 'a { margin-top: "10px"; }',
+			message: messages.rejected('margin-top', '"10px"'),
+			line: 1,
+			column: 17,
+			endLine: 1,
+			endColumn: 23,
+		},
+		{
+			code: 'a { color: greem; }',
```
Mouvedia, line 96: suggestion

As false negatives you could add `color: light-grey;` and `color: light-gray;`.
i.e. both are valid

jeddy3:
This rule is the final piece of the v15 puzzle now that we merged https://github.com/stylelint/stylelint/pull/6561 & https://github.com/stylelint/stylelint/pull/6573.

Let's decide whether we want it in.

As mentioned in https://github.com/stylelint/stylelint/pull/6511/#issue-1479754720, including it now will help communicate to people the type of problems Stylelint can help them with going forward. We've closed two issues (https://github.com/stylelint/stylelint/issues/6591 & https://github.com/stylelint/stylelint/issues/6576) just in the last couple of weeks requesting it, and [there's a desire for it in this Twitter thread](https://twitter.com/argyleink/status/1619098924907905024). People expect and want Stylelint to do this.

However, it comes at the cost of a larger package size and likely some maintenance. For example, CSSTree flags these two:

- `font-size: clamp(1rem, 1vw + 1rem, 3rem);`
- `color: -moz-initial;`

Support for `clamp` (and other math functions) [is coming](https://github.com/csstree/csstree/issues/164#issuecomment-1223111277).

Fortunately, the syntax CSSTree uses [can be extended](https://github.com/csstree/csstree/issues/174#issuecomment-1337001937). So, CSSTree can do the heavy lifting in most cases, and we can maintain our [own references](https://github.com/stylelint/stylelint/tree/main/lib/reference) for things that we can't quickly patch in CSSTree.

There will be edge cases, but we can limit their impact. Regardless, this rule will be beneficial to most people.

What does everyone think?

Unfortunately, I will not be as available as I thought I would be over the coming weeks. If we decide we want this rule in, would someone like to pick it up and get it over the finish line?

Mouvedia:
> Fortunately, the syntax that CSSTree uses [can be extended](https://github.com/csstree/csstree/issues/174#issuecomment-1337001937).

We will probably have a regular stream of issues regarding false positives that will end up being upstreamed to csstree.
To avoid the back and forth—PR here waiting for their release and one after to remove the useless extension—someone from the stylelint organisation needs to have at least merge rights on their repo if we intend to collaborate that tightly.
This way we won't have to wait, we will just use a commit SHA in the package.json if their release cycles are too long.

If that's the case I am all for it.

ybiquitous:
@jeddy3 Sorry for the late reply. I also believe this rule will benefit most people, but I'm a bit concerned about potential false positives and inactivity of the CSSTree repo, as @Mouvedia says.

> - `font-size: clamp(1rem, 1vw + 1rem, 3rem);`

The math functions support like `clamp()` or `max()` is not coming yet, so we should avoid false positives by ignoring such functions, for example:

```js
if (/^(clamp|min|max)\(/i.test(value)) return;
```

Extending via `fork()` seems not to work in this case.

> - `color: -moz-initial;`

CSSTree seems not to support the legacy keyword for `initial` and extend keywords via `fork()`. So we should avoid false positives by just ignoring the keyword, for example:

```js
if (/^-moz-initial$/i.test(value)) return;
```

Sending a patch to CSSTree may be possible, but this is an edge case, so I believe it is sufficient to add the workaround code to this rule.

---

Additionally, I find this rule may be duplicated with `color-no-invalid-hex`. These two rules reports for the following CSS code at the same time:

```css
a { color: #4f; }
```

```
Unexpected invalid hex color "#4f"                   color-no-invalid-hex
Unexpected unknown value "#4f" for property "color"  declaration-property-value-no-unknown
```

---

To summarize, despite potential false positives, I believe this rule has a big value for most people, and CSSTree is suitable to achieve it now.

Perhaps, the author of CSSTree and [stylelint-validator](https://github.com/csstree/stylelint-validator) may collaborate with us because stylelint-validator could be replaced with this rule and future built-in rules using CSSTree.

---

> If we decide we want this rule in, would someone like to pick it up and get it over the finish line?

I can do it. ✋🏼

Mouvedia:
> Perhaps, the author of CSSTree and [stylelint-validator](https://github.com/csstree/stylelint-validator) may collaborate with us because stylelint-validator could be replaced with this rule and future built-in rules using CSSTree.

@lahmatiy what's your opinion on that?

> Additionally, I find this rule may be duplicated with color-no-invalid-hex.

Nice catch. If `declaration-property-value-no-unknown` is added `color-no-invalid-hex` should be deprecated.

> So we should avoid false positives by just ignoring the keyword…

Ill add all the vendor prefixed keywords.

<hr />

Now concerning the options for the rule, it would be necessary to add, in a subsequent PR, ways to support non-standard stuff. It's not really urgent though, as it only matters for old codebases.

e.g. IE only values (css expressions, progid)

lib/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -0,0 +1,111 @@
...
+	return (root, result) => {
+		const validOptions = validateOptions(
+			result,
+			ruleName,
+			{ actual: primary },
+			{
+				actual: secondaryOptions,
+				possible: {
+					ignoreProperties: [isString, isRegExp],
+					ignoreValues: [isString, isRegExp],
```
Mouvedia, line 38: Can I ignore a regex for a specific property?

e.g. for the property `filter` ignore anything that starts with `progid`

ybiquitous: Good catch. It seems good the `ignoreProperties` option supports an object instead of an array, for example:

```json5
{
  "ignoreProperties": {
    "filter": "/^progid/",
    "top": "/.*/"  // any values match
  }
}
```

Any thoughts?

Mouvedia: As long as the name is consistent with the other rules's options. it's OK with me.

I suppose `ignoreValues` is there for global values.
The only example that comes to mind is `-moz-initial`.
If we supported `"ignoreProperties": { "*": "-foo-revert" }` could we drop `ignoreValues`?

lahmatiy:
@Mouvedia I'm open for a collaboration.

CSSTree is out of my focus currently. That primally affects my ability to work on complex features, I have no enough time for it at the moment. However, I can review simple PRs and publish releases when needed.

> For example, CSSTree flags these two:
> * font-size: clamp(1rem, 1vw + 1rem, 3rem);
> * color: -moz-initial;
> ...
> * [width: min-intrinsic;](https://bugs.chromium.org/p/chromium/issues/detail?id=475104)
> * [height: -webkit-fill-available;](https://caniuse.com/mdn-css_properties_height_stretch)

Let me start with how syntax matching (a validation is based on it) works. There is no specific code in CSSTree for a property matching/validation, like many other libraries do. Instead, it uses a general solution based on [Value Definition Syntax](https://www.w3.org/TR/css-values-4/#value-defs) defined in [Values and Units Module](https://www.w3.org/TR/css-values-4/). Using this approach, you can easily add new properties or change existing by making changes in a dictionary. Mentioned `fork()` method allows to have a fork of based CSSTree, but with altered dictionaries.

CSSTree uses [mdn-data](https://github.com/mdn/data/) as a primary dictionary of syntax definitions, with some [additions and fixes](https://github.com/csstree/csstree/blob/master/data/patch.json). In most cases, you need to add/fix a definition in mdn-data dictionary. However, mdn-data's releases a pretty rare events, so it may be much faster to make changes in CSSTree as well.

In fact, in my opinion, the dictionary should be a separate project, which will be aimed at providing a dictionary for use in tools. mdn has no such goal, although it is used in tools (with hacks, like in CSSTree) for lack of a better one. But this is a topic for a separate discussion.

4 cases above is about 3 different areas. Last two is most simple, it's about a missing in the dictionary. You can see current definition for [width](https://csstree.github.io/docs/syntax/#Property:width) and [height](https://csstree.github.io/docs/syntax/#Property:height) properties in CSSTree syntax reference dashboard. The width property is already patched by CSSTree to add some non-standard values:

image

So you can add missing keywords to these properties the same way.

`-moz-initial` is a vendor-prefixed [CSS-wide keyword](https://www.w3.org/TR/css-values-4/#common-keywords). So, that's not a part of any property definition, but a common value for any property. CSSTree has a build-in [list of CSS-wide keywords](https://github.com/csstree/csstree/blob/master/lib/lexer/generic-const.js#L2). It can't be altered with `fork()` method at the moment, but this ability should be easy to add. I'm not sure that `-moz-initial` should be in the list by default.

`clamp()`, `min()`, `max()` and other math functions are [already in the dictionary](https://csstree.github.io/docs/syntax/#report&noedit&title=CSS%20math%20functions&q=ZGljdC5bbmFtZSBpbiBbJ2NhbGMoKScsJ21pbigpJywnbWF4KCknLCdjbGFtcCgpJywncm91bmQoKScsJ21vZCgpJywncmVtKCknLCdzaW4oKScsJ2NvcygpJywndGFuKCknLCdhc2luKCknLCdhY29zKCknLCdhdGFuKCknLCdhdGFuMigpJywncG93KCknLCdzcXJ0KCknLCdoeXBvdCgpJywnbG9nKCknLCdleHAoKScsJ2FicygpJywnc2lnbigpJ11d&v=ewogICAgdmlldzogJ3RhYmxlJywKICAgIGNvbHM6IHsKICAgICAgICB0eXBlOiBmYWxzZSwKICAgICAgICBuYW1lOiB7CiAgICAgICAgICAgIHZpZXc6ICdub3dyYXAnLAogICAgICAgICAgICBjb250ZW50OiAnYXV0by1saW5rJwogICAgICAgIH0sCiAgICAgICAgc3ludGF4OiAnc3ludGF4OnsuLi4sIG1hdGNoVHlwZTogXCdTdHJpbmdcJ30nLAogICAgICAgIG1hdGNoOiBmYWxzZSwKICAgICAgICByZWZzOiBmYWxzZSwKICAgICAgICBwYXJlbnQ6IGZhbHNlLAogICAgICAgIHNlcmlhbGl6YWJsZTogZmFsc2UKICAgIH0KfQ%3D%3D). These functions, together with `calc()`, are special because they are not a part of any property, but can replace any `dimension`, `percentage` or `number`  value. So processing of these function is different. There is a list of [calc functions](https://github.com/csstree/csstree/blob/master/lib/lexer/generic.js#L37) which can be expanded (and named as math function, with support to alter it with `fork()` method). The problem is that the contents of such functions will not be validated as well as their position is correct (for example, that a function substituted instead of a number returns a number, not pixels). To add such validation takes time, this is a big feature. However, I think a lack of validation is better than warning on such functions.

Btw, I believe that there are much more missing parts in dictionaries. Some time ago I started working on a project to automate the synchronization of syntaxes with CSS specs. Also worked on a [project to collect CSS usage data](https://csstree.github.io/real-web-css/#usage) from real web sites. I believe that such projects can be helpful to fill the gaps.

> ```js
> if (/^(clamp|min|max)\(/i.test(value)) return;
> ```

Avoid using fixes like this. `clamp`/`min`/`max`/etc functions can be placed anywhere in a value.

lib/rules/declaration-property-value-no-unknown/index.js
ybiquitous: `ignoreProperties` with an object is [also supported by `unit-allowed-list`](https://stylelint.io/user-guide/rules/unit-allowed-list/#ignoreproperties--unit-property-regex-regexpropertyregexregex-), so I think it is consistent with other rules.

As you say, we could unify `ignoreValues` to `ignoreProperties` with an object. For example:

```json5
{
  "ignoreProperties": {
    "/.*/": "-foo-revert" // ignore the value of any properties
  }
}
```

Mouvedia:
> Avoid using fixes like this. clamp/min/max/etc functions can be placed anywhere in a value.

Here's an example of such a case : `max: ' (source: ' attr(max) ') ';`.

> CSSTree uses [mdn-data](https://github.com/mdn/data/) as a primary dictionary of syntax definitions, with some [additions and fixes](https://github.com/csstree/csstree/blob/master/data/patch.json?rgh-link-date=2023-02-01T23%3A54%3A22Z). In most cases, you need to add/fix a definition in mdn-data dictionary. However, mdn-data's releases a pretty rare events, so it may be much faster to make changes in CSSTree as well.

I reckon we will need to add around 10 exceptions, _at first_. @lahmatiy What method would you recommend for that?

- something on our side
- csstree
  - edit path.json
  - edit the relevant mdn-data dictionaries
  - a separate stylelint.json that we are responsible of

What we want is to avoid rebuttals/refusals/delays on csstree's side. The goal is to have the lowest number of steps.

> However, I can review simple PRs and publish releases when needed.

This is great to hear.

jeddy3:
@lahmatiy Thank you for chiming in and explaining everything so clearly.

It's fantastic that you're open to collaboration. I think we can bring more awareness to your work on dictionaries, usage and syntax as Stylelint is often the entry point to CSS tooling for people. This will hopefully lead to more people contributing to the dictionary (and even the possibility of it becoming a separate project).

I typically use your [syntax reference website](https://csstree.github.io/docs/syntax/) when proposing new rules for Stylelint and I think it's great! 

> However, I think a lack of validation is better than warning on such functions.

I agree. Out of curiosity, is `var()` handled the same way as I noticed that doesn't flag warnings?

It sounds like we can:

- add [`width: min-intrinsic`](https://bugs.chromium.org/p/chromium/issues/detail?id=475104) [`height: -webkit-fill-available`](https://caniuse.com/mdn-css_properties_height_stretch) to [mdn-data](https://github.com/mdn/data/) _and_ patch in CSSTree
- workaround `-moz-initial` in Stylelint until the  built-in [list of CSS-wide keywords](https://github.com/csstree/csstree/blob/master/lib/lexer/generic-const.js#L2) can be altered with `fork()` 
- use `fork()` on the generic "math function" for `clamp()` etc

Am I right about the last one? Is that what you meant with:

> There is a list of [calc functions](https://github.com/csstree/csstree/blob/master/lib/lexer/generic.js#L37) which can be expanded (and named as math function, with support to alter it with fork() method).

I agree with @ybiquitous when he said:

> To summarize, despite potential false positives, I believe this rule has a big value for most people, and CSSTree is suitable to achieve it now.

The rule will benefit most people and we now have an outline for handling future syntaxes and edge cases.

lib/rules/declaration-property-value-no-unknown/index.js
jeddy3: @Mouvedia This is a great suggestion. 

Let's unify `ignoreValues` to `ignoreProperties` with an object as @ybiquitous said.

jeddy3:
> Additionally, I find this rule may be duplicated with `color-no-invalid-hex`. 

Yes, I touched on this in https://github.com/stylelint/stylelint/issues/6376#issuecomment-1279825153. The rule also overlaps with `unit-no-unknown` and `function-no-unknown` when checking declarations.

We could either leave the rules in, deprecate them or limit them to at-rules. 

(The latter opens up a broader discussion about a new `at-rule-name-prelude-no-unknown` rule that also uses CSSTree.)

We can leave them in for now to keep things simple. We already have overlapping rules, e.g. all the `*-disallowed-list` and `*-allowed-list` ones. Keeping them in gives users a choice of using a more general-purpose property & value pair validating rule or something specific to hex colours, functions, units and so on. We can document this overlapping in the appropriate rule  READMEs.

Mouvedia:
> The rule also overlaps with unit-no-unknown and function-no-unknown when checking declarations.

We won't be able to deprecate them right away.

`function-no-unknown`: because css-functions-list package is more complete (e.g. `oklch`)
`unit-no-unknown`: because it has 2 options that won't—and should probably not—be covered `declaration-property-value-no-unknown`

IMHO we should keep them as is.

> It's fantastic that you're open to collaboration.

👍

ybiquitous:
Thanks a lot, everyone. It's great to hear about this open collaboration! 🙌🏼 
We will be contributing to CSSTree with the help of @lahmatiy in the future.

So now, I'd like to hear your thoughts on whether we should include this `declaration-property-value-no-unknown` rule in the v15.0.0 release.

Because this collaboration will likely take more time and the rule is duplicated with rules like `*-no-unknown` (not suitable to put them in `stylelint-config-recommended` together), I guess it may be good to go toward the v15 release **without the rule**. The `v15` branch has been already long-living and so maintaing `main` and `v15` branches should be a messy job for us.

What do you think? @jeddy3 @Mouvedia

Mouvedia:
> I guess it may be good to go toward the v15 release without the rule.
> What do you think?

We usually have _at least_ one release per month; January was release-less.
Id vote for an early release _if_ the release cadence matters.

jeddy3:
> I guess it may be good to go toward the v15 release without the rule.

It comes down to optics; I suspect some users will be surprised by the deprecation of the stylistic rules. Flagging unknown property & value pairs is our most requested feature, and addressing this in the same release will succinctly demonstrate our new focus on helping people avoid errors and enforce (non-stylistic) conventions. Optics often matter.

> Because this collaboration will likely take more time and the rule is duplicated with rules like *-no-unknown (not suitable to put them in stylelint-config-recommended together)

We can mitigate the rule taking time by making some compromises, we:

- make the rule _good enough_
- don't include it in our recommended config just yet
- draw attention to the rule in the migration guide

We can frame it as a _choice_ over some of the other `*-no-unknown` rules and communicate any known issues in the migration guide. We could even flag it as experimental.

By not including in our recommended config just yet, we'll get a more steady stream of feedback and bug reports about it. Once we're happy it's robust we can include it in a major config release.

To make the rule good enough, it looks like we need to:

- unify [`ignoreValues` to `ignoreProperties`](https://github.com/stylelint/stylelint/pull/6511#pullrequestreview-1278598337) with an object
- workaround `-moz-initial` in Stylelint until the  built-in [list of CSS-wide keywords](https://github.com/csstree/csstree/blob/master/lib/lexer/generic-const.js#L2) can be altered with `fork()` 
- add [`width: min-intrinsic`](https://bugs.chromium.org/p/chromium/issues/detail?id=475104) [`height: -webkit-fill-available`](https://caniuse.com/mdn-css_properties_height_stretch) to [mdn-data](https://github.com/mdn/data/) _and_ patch in CSSTree
- use `fork()` on the generic "math function" for `clamp()` etc

If the latter two prove problematic, we can [work around them in Stylelint](https://github.com/stylelint/stylelint/pull/6511#issuecomment-1411530828). It's OK if we introduce false broad false negatives because we can fix them later down the line.

I'll have time to make the changes to the migration guide, but not to the rule itself. Does anyone have time to do this? 

> We usually have at least one release per month; January was release-less.

While we're making the finishing touches to the pull request, we can release [14.14.0](https://github.com/stylelint/stylelint/blob/a31f2a191874d04d0cef6d83ea00ba53a610af94/CHANGELOG.md) which contains a few new features and fixes.

Mouvedia:
> don't include it in our recommended config just yet

👍 

> add [width: min-intrinsic](https://bugs.chromium.org/p/chromium/issues/detail?id=475104) [height: -webkit-fill-available](https://caniuse.com/mdn-css_properties_height_stretch) to [mdn-data](https://github.com/mdn/data/) and patch in CSSTree

Id probably patch the dictionaries directly on CSSTree side, but that's just me.

> workaround -moz-initial in Stylelint until the built-in [list of CSS-wide keywords](https://github.com/csstree/csstree/blob/master/lib/lexer/generic-const.js?rgh-link-date=2023-02-03T12%3A12%3A12Z#L2) can be altered with fork()

An alternative could be to add an entry to that file directly; that could be considered dirty though.
I am considering that because it's not worth the effort to add the feature to `fork` due to the fact that there is probably only one global value that will ever need to be added.

> Does anyone have time to do this?

Ill do 2 tentative PRs on csstree this weekend. Ill leave the `fork` for later or to @ybiquitous.

> …we can release [14.14.0](https://github.com/stylelint/stylelint/blob/a31f2a191874d04d0cef6d83ea00ba53a610af94/CHANGELOG.md) which contains a few new features and fixes.

Can you please include PR 6608 in it?

ybiquitous:
@jeddy3 Thanks for summarizing. I understood the value of this rule and agreed with including it in v15. 👍🏼 

> * don't include it in our recommended config just yet

Agree.

> We can frame it as a _choice_ over some of the other `*-no-unknown` rules and communicate any known issues in the migration guide. We could even flag it as experimental.

If the rule leaves the possibility of false negatives, I agree with flagging it as experimental (assuming we add an explanation to the rule doc). Though it depends on the degree of patching progress. We also need to describe the possibility of conflicts with the existing rules.

> I'll have time to make the changes to the migration guide, but not to the rule itself. Does anyone have time to do this?

I'll take on the left implementation of this rule. 💪🏼 

> While we're making the finishing touches to the pull request, we can release [14.14.0](https://github.com/stylelint/stylelint/blob/a31f2a191874d04d0cef6d83ea00ba53a610af94/CHANGELOG.md) which contains a few new features and fixes.

If people require it, then we'll do it.

> Ill do 2 tentative PRs on csstree this weekend. Ill leave the `fork` for later or to @ybiquitous.

Thanks, @Mouvedia. I'll do the work left on this rule.

> Can you please include PR 6608 in it?

I'll review it this weekend.

ybiquitous:
Now I'm going to push some commits to this `issue-6376` branch. Please mention me if having some problem.

ybiquitous:
Added a workaround for `-moz-initial` via ddfdbfe2. Let's remove it if it would be unnecessary for a CSSTree update.

ybiquitous:
I've skipped a `clamp()` test to avoid CI failures via 7e91fc29. If we find a suitable workaround or CSSTree supports it, let's remove the skipping.

lib/rules/declaration-property-value-no-unknown/index.js
ybiquitous: Done! d5a5ca88

lib/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -0,0 +1,124 @@
...
+			if (isCustomProperty(prop)) return;
+
+			if (isPropIgnored(prop, value)) return;
+
+			try {
+				parse(value, {
+					context: 'value',
+				});
+			} catch (e) {
+				return;
```
ybiquitous, line 92: [suggest] What about handling a parse error like this?

```suggestion
				result.warn(`Cannot parse property value "${value}"`, {
					node: decl,
					stylelintType: 'parseError',
				});

				return;
```

If the following code would be given:

```css
a { color: &a; }
```

the rule could report:

```
Cannot parse property value "&a"
```

ybiquitous:
What I can do now is finished. Please mention me if more work comes.

lib/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -0,0 +1,124 @@
...
+			properties: propertiesSyntax,
+			types: typesSyntax,
+		}).lexer;
+
+		root.walkDecls((decl) => {
+			const { prop, value } = decl;
+
+			// NOTE: CSSTree's `fork()` doesn't support `-moz-initial`, but it may be possible in the future.
+			// See https://github.com/stylelint/stylelint/pull/6511#issuecomment-1412921062
+			if (/^-moz-initial$/i.test(value)) return;
```
Mouvedia, line 75: let's wait for csstree/csstree#243 to settle on this.

Mouvedia:
Iv done the PRs that I was talking about in https://github.com/stylelint/stylelint/pull/6511#issuecomment-1415821725.
Let's wait for @lahmatiy review/merge.

The `height` property will require more work on my part.
i.e.
1. a PR on mdn/browser-compat-data or mdn/data or w3c/webref
2. a subsequent PR on csstree/csstree
3. a dependency update on stylelint/stylelint
Hence Ill postpone it.
i.e. it won't be ready for _this_ PR

lib/rules/declaration-property-value-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,246 @@
...
+		},
+		{
+			code: 'a { color: -moz-initial; }',
+		},
+		{
+			// TODO: CSSTree doesn't support math functions like `clamp()` yet, but the future update will support them.
+			// See https://github.com/stylelint/stylelint/pull/6511#issuecomment-1412921062
+			skip: true,
+			code: 'a { font-size: clamp(1rem, 1vw + 1rem, 3rem); }',
+		},
```
jeddy3, line 45: Let's favour false negatives over positives.

If we can't use `fork()` in this instance, let's return early if [we find](https://github.com/csstree/csstree/blob/master/docs/traversal.md#findast-fn) a `clamp`, `min` or `max` function in the value.

jeddy3:
@ybiquitous & @Mouvedia Thank you for making commits or pull requests here elsewhere.

I've pushed a commit https://github.com/stylelint/stylelint/pull/6511/commits/5872036a03686f9cde2cc55d4642bb8328dc3d17 to update the docs to address suggestions in https://github.com/stylelint/stylelint/pull/6511#issuecomment-1416520482 & https://github.com/stylelint/stylelint/pull/6511#issuecomment-1415789828.

I've made one suggestion. Otherwise, I think we're almost there!

lib/rules/declaration-property-value-no-unknown/__tests__/index.js
ybiquitous: Thanks for sharing the nice way to find functions. Fixed via 93171400.

patches/@types+css-tree+2.0.1.patch
```diff
@@ -0,0 +1,23 @@
...
++        name: string;
++        rawMessage: string;
++    };
++};
++
++declare class Lexer {
++    matchProperty(propertyName: string, value: string): MatchResult;
++}
++
++export function fork(extension: Record<string, Record<string, string>>): { lexer: Lexer };
```
ybiquitous, line 23: I'll request this change to the upstream `@types/css-tree` package later.

ybiquitous (approved):
Thanks, everyone. LGTM 👍🏼 
I believe we can later apply the patches by @Mouvedia that CSSTree will include.

jeddy3:
Thanks, everyone. Looks good enough for me too.

Mouvedia:
[What I was afraid of](https://github.com/stylelint/stylelint/pull/6511#issuecomment-1407586913) has happened.
None of the PRs opened on csstree's side were merged.

ybiquitous:
@lahmatiy What do you think about delegating CSSTree's maintenance to someone passionate? Since CSSTree is a great product, I believe people in the front-end community hope to make CSSTree sustainable.
(maybe this discussion should be open in CSSTree's repo, though)

Mouvedia:
@ybiquitous in the meantime what about forking the repo and using a github branch as a dependency ?

ybiquitous:
@Mouvedia Um, it may be better at this point.

> using a github branch as a dependency

`css-tree` requires building, right? If forking it, do we need to publish it to npmjs.com?
Ref: https://github.com/csstree/csstree/blob/612cc5f2922b2304869497d165a0cc65257f7a8b/package.json#L85

Mouvedia:
> `css-tree` requires building, right?

Now that becomes annoying. It would require to use GitHub actions to build it every time a PR is merged.

> do we need to publish it to npmjs.com?

Nah that's overkill; I'd prefer waiting for the pending PRs to be merged.

Mouvedia:
The last version of css-tree was released in late 2022.
Some PRs related to filled stylelint issues were merged, but a new version wasn't published.
e.g. #6667, #7241
The last commit on the repository was in October of last year.

Is forking on the table? Or should we wait?

ybiquitous:
Thanks for the catch-up. I'd appreciate it if you could open a new issue for the discussion.

PR #6529, opened 2022-12-16
Add `ignoreRules` to `max-nesting-depth`rule

### LINKED ISSUE #2696: Add ignoreRules: [] to max-nesting-depth

**Feature request:** 
Add: `ignoreSelectors: [".my-custom-selector"]` for `max-nesting-depth`
I often find myself using `& {}` as a means to visually encapsulate some rules. As this counts towards the limit of `max-nesting-depth` I sometimes need to disable this rule, which is a bit annoying.
This could also be useful to exclude `::after` and `::before` elements as these need to be nested (except you opt for repeating their parent selector).

`max-nesting-depth`

`7.12.0`

Not necessarily, although I have in mind to exclusively use it with the SCSS selector `&`

### LINKED ISSUE #4805: Add ignoreSelectors: [] to max-nesting-depth

sjarva:

Closes #2696 and closes #4805 (which are duplicates of each other)

- implements the option to add `ignoreRules` to `options`
- adds unit tests to test that feature is working as expected

### DIFF
--- a/lib/rules/max-nesting-depth/index.js
+++ b/lib/rules/max-nesting-depth/index.js
@@ -28,6 +28,40 @@ const rule = (primary, secondaryOptions) => {
 	const isIgnoreAtRule = (node) =>
 		isAtRule(node) && optionsMatches(secondaryOptions, 'ignoreAtRules', node.name);
 
+	/**
+	 * @param {string} selector
+	 * @returns {string | undefined}
+	 */
+	function extractPseudoRule(selector) {
+		return selector.startsWith('&:') && selector[2] !== ':' ? selector.slice(2) : undefined;
+	}
+
+	/**
+	 * @param {string} selector
+	 */
+	function containsPseudoClassesOnly(selector) {
+		const normalized = parser().processSync(selector, { lossless: false });
+		const selectors = normalized.split(',');
+
+		return selectors.every((sel) => extractPseudoRule(sel));
+	}
+
+	/**
+	 * @param {string[]} selectors
+	 * @returns {boolean}
+	 */
+	function containsIgnoredPseudoClassesOnly(selectors) {
+		if (!(secondaryOptions && secondaryOptions.ignorePseudoClasses)) return false;
+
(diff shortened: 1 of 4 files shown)


ybiquitous (commented):
@sjarva Thank you for opening the pull request. The PR title includes `ignoreSelectors`, but it actually adds `ignoreRules`. Is the title incorrect?

sjarva:
> The PR title includes `ignoreSelectors`, but it actually adds `ignoreRules`. Is the title incorrect?

Good catch, the title is incorrect. It is now fixed in the PR title and PR description. I noticed that I had also put it in the branch name - should I update that too?

ybiquitous:
> I noticed that I had also put it in the branch name - should I update that too?

No. It's OK to keep it as-is.

lib/rules/max-nesting-depth/index.js
```diff
@@ -151,14 +184,6 @@ const rule = (primary, secondaryOptions) => {
...
 };
 
-/**
- * @param {string} selector
- * @returns {string | undefined}
- */
-function extractPseudoRule(selector) {
-	return selector.startsWith('&:') && selector[2] !== ':' ? selector.slice(2) : undefined;
-}
-
```
ybiquitous, line 161: [suggest] Could you please keep the original position of some functions to avoid potential conflicts with other PRs? In other words, I'd like to minimize diffs in the PR.

ybiquitous (changes requested):
I've left several suggestions to improve the doc and code quality, but I don't see significant problems.

sjarva:
Thanks for the review and suggestions @ybiquitous! I'll take a look.

Mouvedia:
@sjarva Thanks for tackling such an old issue.

mattxwang:
Hi @sjarva, just checking in - anything we can do to support you in wrapping up this PR? Wary of the branches diverging too much!

PR #6536, opened 2022-12-22
Fix `customSyntax` resolution with `configBasedir`

### LINKED ISSUE #6525: Fix `overrides.files` in config for glob patterns like `*.vue`

Well, I'll try to be clear here as it is not easy to explain...

I created a simple repo that you can clone to reproduce the bug: https://github.com/sronveaux/stylelint-bug-report

Here's the smallest code base I managed to get to and that can be used to reproduce the bug:

    filesToLint/
    |    |
    |    ------> test.css
    |    ------> test.html
    |
    stylelintInstall/
    |    |
    |    ------> node_modules/
    |    ------> package.json
    |
    .stylelintrc

The .stylelintrc content is very simple:

    {
      "extends": [
        "stylelint-config-standard"
       ],
      "overrides": [
        {
          "files": ["*.html", "**/*.html"],
          "customSyntax": "postcss-html"
        }
      ]
    }

If I get to the stylelintInstall directory and run the following commands:

    npx stylelint ../filesToLint/*.css
    npx stylelint ../filesToLint/*.html

everything works as expected. However, if configBasedir is added:

    npx stylelint --config-basedir . ../filesToLint/*.html

The .html files cannot be linted anymore and the following message is returned:

    When linting something other than CSS, you should install an appropriate syntax, e.g. "postcss-html", and use the "customSyntax" option

    ../filesToLint/test.html

This can be made working back again by changing the content of the .stylelintrc file like this:

    {
      "extends": [
        "stylelint-config-standard"
       ],
      "overrides": [
        {
          "files": ["../*.html", "../**/*.html"],
          "customSyntax": "postcss-html"
        }
      ]
    }

However this manual patch doesn't work when used in conjuction with extensions like `stylelint-config-standard-vue`, not to mention how it fails when used through the `VSCode extension`...

    {
      "extends": [
        "stylelint-config-standard"
       ],
      "overrides": [
        {
          "files": ["*.html", "**/*.html"],
          "customSyntax": "postcss-html"
        }
      ]
    }

CLI

14.16.0 under Windows 11

Shouldn't the glob patterns we write in the .stylelintrc file be relative to this config file instead of being interpreted differently when used in conjunction with --config-basedir ?

### What actually happened?

It seems like the file matching in the `overrides` part of the config is dependent on the `config-basedir` as the problem can be solved by changing the glob pattern written in the `overrides` part.

_No response_

### Proposal to fix the bug

Not sure exactly and haven't read the source code long enough, but this could be linked to this part:

https://github.com/stylelint/stylelint/blob/e30ec863f8bc1d5ed87c23e3c5a554a01bc4350b/lib/augmentConfig.js#L106-L113

Here we can see that `configDir` (which is in fact `configBasedir`) is passed twice to the `augmentConfigBasic` function.

https://github.com/stylelint/stylelint/blob/e30ec863f8bc1d5ed87c23e3c5a554a01bc4350b/lib/augmentConfig.js#L34-L63

Then `augmentConfigBasic` seems to call `applyOverrides` with the `rootConfigDir` parameter which perhaps should be the directory where the config file is and not `configBasedir`...

ybiquitous:

Ref #6525

This PR fixes resolution of the `customSyntax` (`--custom-syntax`) option. The option should be able to be read from `configBasedir` (`--config-basedir`) like `plugins` or `extends`.

### DIFF
--- a/.changeset/fifty-donuts-arrive.md
+++ b/.changeset/fifty-donuts-arrive.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `customSyntax` resolution with `configBasedir`
--- a/docs/user-guide/usage/cli.md
+++ b/docs/user-guide/usage/cli.md
@@ -46,7 +46,7 @@ Force enabling/disabling of color.
 
 ### `--config-basedir`
 
-Absolute path to the directory that relative paths defining "extends" and "plugins" are _relative to_. Only necessary if these values are relative paths. [More info](options.md#configbasedir).
+Absolute path to the directory that relative paths defining "extends", "plugins", and "customSyntax" are _relative to_. Only necessary if these values are relative paths. [More info](options.md#configbasedir).
 
 ### `--config`
 
--- a/docs/user-guide/usage/options.md
+++ b/docs/user-guide/usage/options.md
@@ -28,7 +28,7 @@ The path should be either absolute or relative to the directory that your proces
 
 CLI flag: `--config-basedir`
 
-Absolute path to the directory that relative paths defining "extends" and "plugins" are _relative to_. Only necessary if these values are relative paths.
+Absolute path to the directory that relative paths defining "extends", "plugins", and "customSyntax" are _relative to_. Only necessary if these values are relative paths.
 
 ## `fix`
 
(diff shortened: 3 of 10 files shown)


jeddy3 (approved):
LGTM, thank you.

Although, it doesn't fix the reported case. This is a bug we should fix, right?

ybiquitous:
> Although, it doesn't fix the reported case. This is a bug we should fix, right?

Yes, right. So I've updated `Closes #6525` to `Ref #6525` in this PR description.

PR #5460, opened 2021-08-12
Add disableFix secondary option

kuangrs:
>  Which issue, if any, is this issue related to? 

Closes #5431 

> Is there anything in the PR that needs further explanation?

no

PR #6543, opened 2022-12-23
Fix unexpected `output` in Node.js API lint result when any rule contains `disableFix: true`

### LINKED ISSUE #6542: Unexpected Node API output when any rule contains `disableFix: true`

`disableFix: true` in any stylelint rule breaks the `output` of the node api if the `fix: true` option is set. 
If this configuration is provided for any rule, instead of containing the fixed code, the regular output that would have been shown without the `fix: true` flag is returned.
A symptom of this issue is that the VSCode eidtor integration can not autofix any issues.

Example Script
```ts
import * as stylelint from "stylelint";

async function lint(cssContent: string): Promise<void> {
  try {
    const lintResult = await stylelint.lint({
      code: cssContent,
      fix: true,
    });
    console.log(lintResult.output);
  } catch (error) {
    console.error(error);
  }
}
lint(`.border-top-default {
  border-top: 1px solid red;
  margin: 0;
}`);
```
And configuration:
Example stylelint config:
used packages: stylelint-order, stylelint-scss, stylelint-declaration-strict-value.
```js
module.exports = {
  customSyntax: "postcss-scss",
  plugins: ["stylelint-declaration-strict-value", "stylelint-order"],
  rules: {
    "order/properties-order": [["margin", "border"], { unspecified: "bottom" }],
    "scale-unlimited/declaration-strict-value": [
      "/border/",
      { disableFix: true }, // this one breaks the autofix
    ],
  },
};
```

```js
module.exports = {
  plugins: ["stylelint-declaration-strict-value", "stylelint-order"],
  rules: {
    "order/properties-order": [["margin", "border"], { unspecified: "bottom" }],
    "scale-unlimited/declaration-strict-value": [
      "/border/",
      { disableFix: true }, // this one breaks the autofix
    ],
  },
};
```

NodeJS script, provided in steps to reproduce bug

14.15.0

I expect the `intResult.output` to contain the fixed CSS
```css
.border-top-default {
  margin: 0;
  border-top: : 1px solid red;
}
```

### What actually happened?

But as soon as `disableFix: true` is set, I get the following result:
```json
[{
  "source": "<input css rPadxH>",
  "deprecations": [],
  "invalidOptionWarnings": [],
  "parseErrors": [],
  "errored": false,
  "warnings": []
}]
```
which is similar to the result that I get when I remvoe the `fix: true` option from the node script.
```json
[{
  "source": "<input css qETqql>",
  "deprecations": [],
  "invalidOptionWarnings": [],
  "parseErrors": [],
  "errored": true,
  "warnings": [
    {
      "line": 3,
      "column": 5,
      "endLine": 3,
      "endColumn": 15,
      "rule": "order/properties-order",
      "severity": "error",
      "text": "Expected \"margin\" to come before \"border-top\" (order/properties-order)"
    }
  ]
}]
```

I don't think so

### Proposal to fix the bug

Not yet, will try to look into it.

adrianjost:

Closes #6542
and by that should also fix https://github.com/stylelint/vscode-stylelint/issues/369

Issue: The API did not return the fixed css in `output` if any rule contained `ruleDisableFix`.

I have removed the `ruleDisableFix` code, because it did not affect the code fixing itself and I assume was wrongly used to change the content format of `output`. I have adjusted some tests to adapt to the changed behavior and extended the test conditions to check that the returned code has the correct fixes applied.

Issue was introduced in https://github.com/stylelint/stylelint/pull/5460

### DIFF
--- a/lib/__tests__/standalone-fix.test.js
+++ b/lib/__tests__/standalone-fix.test.js
@@ -246,7 +246,8 @@ it('one rule being disabled', async () => {
 		fix: true,
 	});
 
-	expect(JSON.parse(result.output)[0].errored).toBe(true);
+	expect(result.results[0].errored).toBe(true);
+	expect(result.output).toBe('\n		a {\n			color: red;\n		}');
 });
 
 it('two rules being disabled', async () => {
@@ -276,7 +277,7 @@ it('two rules being disabled', async () => {
 		fix: true,
 	});
 
-	const warnings = JSON.parse(result.output)[0].warnings;
+	const warnings = result.results[0].warnings;
 
 	expect(warnings.some((w) => w.text === 'Expected indentation of 0 spaces (indentation)')).toBe(
 		true,
@@ -284,6 +285,7 @@ it('two rules being disabled', async () => {
 	expect(warnings.some((w) => w.text === 'Expected "COLOR" to be "color" (property-case)')).toBe(
 		true,
 	);
+	expect(result.output).toBe('\n		a {\n			COLOR: red;\n		}');
 });
 
 it('one rule being disabled and another still autofixing', async () => {
@@ -296,24 +298,25 @@ it('one rule being disabled and another still autofixing', async () => {
 		code,
 		config: {
 			rules: {
-				indentation: [
-					2,
+				indentation: [0],
+				'property-case': [
+					'lower',
 					{
 						disableFix: true,
 					},
 				],
-				'property-case': ['lower'],
 			},
 		},
 		fix: true,
 	});
 
-	const warnings = JSON.parse(result.output)[0].warnings;
+	const warnings = result.results[0].warnings;
 
 	expect(warnings.some((w) => w.text === 'Expected indentation of 0 spaces (indentation)')).toBe(
-		true,
-	);
-	expect(warnings.some((w) => w.text === 'Expected "COLOR" to be "color" (property-case)')).toBe(
 		false,
 	);
+	expect(warnings.some((w) => w.text === 'Expected "COLOR" to be "color" (property-case)')).toBe(
+		true,
+	);
+	expect(result.output).toBe('\na {\nCOLOR: red;\n}');
 });
--- a/lib/__tests__/standalone-fix.test.js
+++ b/lib/__tests__/standalone-fix.test.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const path = require('path');
-const { stripIndent } = require('common-tags');
+const { stripIndent, stripIndents } = require('common-tags');
 const fs = require('fs').promises;
 
 const removeFile = require('../testUtils/removeFile');
@@ -247,7 +247,7 @@ it('one rule being disabled', async () => {
 	});
 
 	expect(result.results[0].errored).toBe(true);
-	expect(result.output).toBe('\n		a {\n			color: red;\n		}');
+	expect(result.output).toBe(code);
 });
 
 it('two rules being disabled', async () => {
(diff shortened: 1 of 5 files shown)


lib/__tests__/standalone-fix.test.js
```diff
@@ -246,7 +246,8 @@ it('one rule being disabled', async () => {
 		fix: true,
 	});
 
-	expect(JSON.parse(result.output)[0].errored).toBe(true);
+	expect(result.results[0].errored).toBe(true);
+	expect(result.output).toBe('\n		a {\n			color: red;\n		}');
```
ybiquitous, line 250: [suggest] What about making the code a bit smaller as follows?

```suggestion
	expect(result.output).toBe(code);
```

lib/__tests__/standalone-fix.test.js
```diff
@@ -276,14 +277,15 @@ it('two rules being disabled', async () => {
...
-	const warnings = JSON.parse(result.output)[0].warnings;
+	const warnings = result.results[0].warnings;
 
 	expect(warnings.some((w) => w.text === 'Expected indentation of 0 spaces (indentation)')).toBe(
 		true,
 	);
 	expect(warnings.some((w) => w.text === 'Expected "COLOR" to be "color" (property-case)')).toBe(
 		true,
 	);
+	expect(result.output).toBe('\n		a {\n			COLOR: red;\n		}');
```
ybiquitous, line 288: [suggest] Same here.

```suggestion
	expect(result.output).toBe(code);
```

lib/__tests__/standalone-fix.test.js
```diff
@@ -296,24 +298,25 @@ it('one rule being disabled and another still autofixing', async () => {
...
 
 	expect(warnings.some((w) => w.text === 'Expected indentation of 0 spaces (indentation)')).toBe(
-		true,
+		false,
 	);
 	expect(warnings.some((w) => w.text === 'Expected "COLOR" to be "color" (property-case)')).toBe(
-		false,
+		true,
 	);
+	expect(result.output).toBe('\na {\nCOLOR: red;\n}');
```
ybiquitous, line 321: [suggest] What about using `stripIndent` for readability?

At L292:

```diff
-	const code = `
+	const code = stripIndent`
		a {
			COLOR: red;
		}`;
```

```suggestion
	expect(result.output).toBe(stripIndent`
		a {
		COLOR: red;
		}`);
```

ybiquitous (commented):
Thank you for creating the PR! I've left refactoring suggestions to improve readability a bit, but it looks good to me.

lib/__tests__/standalone-fix.test.js
adrianjost: Applied your suggestion slightly modified.
- I have used `stripIndents` instead of `stripIndent` because I want to remove the indentation of all lines
- I have applied `stripIndent` to the source code, because `stripIndents` and `stripIndent` remove an empty line so that the first linebreak of the source-code caused an expect mismatch with the `stripIndents` version it is compared with, because that line was removed entirely.

lib/__tests__/standalone-fix.test.js
adrianjost: sure, thanks for the suggestion.

lib/__tests__/standalone-fix.test.js
adrianjost: sure, thanks for the suggestion.

ybiquitous (approved):
Thank you. LGTM 👍🏼

jeddy3 (approved):
@adrianjost Thank you for contributing!

LGTM.

PR #6545, opened 2022-12-23
Refactor declaration-block-no-duplicate-properties

### LINKED ISSUE #6544: Refactor `declaration-block-no-duplicate-properties`
Hi there. While reading the code I noticed that the declaration-block-no-duplicate-properties rule has a quite deep nesting. I think the rule could be refactored a bit so the code become more readable. What do you think? If you agree, I can take care of it.

fpetrakov:

Closes #6544

No, it's self-explanatory.

### DIFF
--- a/lib/rules/declaration-block-no-duplicate-properties/index.js
+++ b/lib/rules/declaration-block-no-duplicate-properties/index.js
@@ -105,52 +105,40 @@ const rule = (primary, secondaryOptions, context) => {
 					// fails if duplicates are not consecutive
 					if (indexDuplicate !== decls.length - 1) {
 						if (!context.fix) {
-							report({
+							return report({
 								message: messages.rejected(prop),
 								node: decl,
 								result,
 								ruleName,
 								word: prop,
 							});
-
-							return;
 						}
 
 						if (!important && duplicateImportant) {
-							decl.remove();
-
-							return;
+							return decl.remove();
 						}
 
-						duplicateDecl.remove();
-
-						return;
+						return duplicateDecl.remove();
 					}
 
 					if (ignorePrefixlessSameValues) {
 						// fails if values of consecutive, unprefixed duplicates are equal
 						if (vendor.unprefixed(value) !== vendor.unprefixed(duplicateValue)) {
 							if (!context.fix) {
-								report({
+								return report({
 									message: messages.rejected(prop),
 									node: decl,
 									result,
 									ruleName,
 									word: prop,
 								});
-
-								return;
 							}
 
 							if (!important && duplicateImportant) {
-								decl.remove();
-
-								return;
+								return decl.remove();
 							}
 
-							duplicateDecl.remove();
-
-							return;
+							return duplicateDecl.remove();
 						}
 					}
 
@@ -159,20 +147,16 @@ const rule = (primary, secondaryOptions, context) => {
 					}
 
 					if (context.fix) {
-						duplicateDecl.remove();
-
-						return;
+						return duplicateDecl.remove();
 					}
 
-					report({
+					return report({
 						message: messages.rejected(prop),
 						node: decl,
 						result,
 						ruleName,
 						word: prop,
 					});
-
-					return;
 				}
 
 				if (ignoreDuplicates && indexDuplicate === decls.length - 1) {
@@ -180,21 +164,17 @@ const rule = (primary, secondaryOptions, context) => {
 				}
 
 				if (!context.fix) {
-					report({
+					return report({
 						message: messages.rejected(prop),
 						node: decl,
 						result,
 						ruleName,
 						word: prop,
 					});
-
-					return;
 				}
 
 				if (!important && duplicateImportant) {
-					decl.remove();
-
-					return;
+					return decl.remove();
 				}
 
 				duplicateDecl.remove();
--- a/lib/rules/declaration-block-no-duplicate-properties/index.js
+++ b/lib/rules/declaration-block-no-duplicate-properties/index.js
@@ -105,7 +105,7 @@ const rule = (primary, secondaryOptions, context) => {
 				const unprefixedDuplicatesAreEqual =
 					vendor.unprefixed(value) === vendor.unprefixed(duplicateValue);
 
-				function fixOrReport() {
+				const fixOrReport = () => {
 					if (!context.fix) {
 						return report({
 							message: messages.rejected(prop),
@@ -120,10 +120,8 @@ const rule = (primary, secondaryOptions, context) => {
 						return decl.remove();
 					}
 
-					if (!duplicateDecl) return;
-
 					return duplicateDecl.remove();
-				}
+				};
 
 				if (ignoreDiffValues || ignorePrefixlessSameValues) {
 					if (
--- a/lib/rules/declaration-block-no-duplicate-properties/index.js
+++ b/lib/rules/declaration-block-no-duplicate-properties/index.js
@@ -102,9 +102,14 @@ const rule = (primary, secondaryOptions, context) => {
 				const duplicateImportant = duplicateDecl.important || false;
 				const duplicateIsMoreImportant = !important && duplicateImportant;
 				const duplicatesAreConsecutive = indexDuplicate === decls.length - 1;
+				const unprefixedDuplicatesAreEqual =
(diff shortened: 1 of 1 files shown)


lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -185,25 +171,38 @@ const rule = (primary, secondaryOptions, context) => {
...
+						result,
+						ruleName,
+						word: prop,
+					});
+
+					return;
+				}
+
+				if (!important && duplicateImportant) {
+					decl.remove();
```
Mouvedia, line 195: stylistic

I prefer the form `return fn();`.
ditto for l183

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -117,19 +116,21 @@ const rule = (primary, secondaryOptions, context) => {
...
-						if (ignorePrefixlessSameValues) {
-							// fails if values of consecutive, unprefixed duplicates are equal
-							if (vendor.unprefixed(value) !== vendor.unprefixed(duplicateValue)) {
-								if (context.fix) {
-									if (!important && duplicateImportant) {
-										decl.remove();
-									} else {
-										removePreviousDuplicate(decls, lowerProp);
-									}
+						if (!important && duplicateImportant) {
```
Mouvedia, line 119: minor

put `!important && duplicateImportant` in a const
(being reused at l145)

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -88,24 +88,23 @@ const rule = (primary, secondaryOptions, context) => {
...
+				if (!duplicateDecl) {
+					return;
+				}
+
+				const duplicateValue = duplicateDecl.value || '';
+				const duplicateImportant = duplicateDecl.important || false;
 
+				if (ignoreDiffValues || ignorePrefixlessSameValues) {
+					// fails if duplicates are not consecutive
+					if (indexDuplicate !== decls.length - 1) {
```
Mouvedia, line 106: suggestion

put `indexDuplicate !== decls.length - 1` in a const foo (dunno about the name but you ought to)
and use `!foo` at l178

Mouvedia:
perf

verify that it's at least marginally equivalent
https://github.com/stylelint/stylelint/blob/main/docs/developer-guide/rules.md#improve-the-performance-of-a-rule

fpetrakov:
> perf
> 
> verify that it's at least marginally equivalent https://github.com/stylelint/stylelint/blob/main/docs/developer-guide/rules.md#improve-the-performance-of-a-rule

before:
image

after:
image

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -88,97 +88,84 @@ const rule = (primary, secondaryOptions, context) => {
...
+					return report({
+						message: messages.rejected(prop),
+						node: decl,
+						result,
+						ruleName,
+						word: prop,
+					});
+				}
+
+				if (ignoreDuplicates && duplicatesAreConsecutive) {
```
Mouvedia, line 163: question

Do we have a test for
ignoreDuplicates === true
duplicatesAreConsecutive === false
content.fix === true
duplicateIsMoreImportant === false
?
i.e. in that case should we reach l181? Is this setup possible?

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -88,97 +88,84 @@ const rule = (primary, secondaryOptions, context) => {
...
-									return;
-								}
+						return duplicateDecl.remove();
+					}
 
-								report({
+					if (ignorePrefixlessSameValues) {
+						// fails if values of consecutive, unprefixed duplicates are equal
+						if (vendor.unprefixed(value) !== vendor.unprefixed(duplicateValue)) {
+							if (!context.fix) {
```
Mouvedia, line 128: It seems like you have
```
if (!context.fix) {
	return report({
		message: messages.rejected(prop),
		node: decl,
		result,
		ruleName,
		word: prop,
	});
}

if (duplicateIsMoreImportant) {
	return decl.remove();
}

return duplicateDecl.remove();
```
twice.

What about

```js
// false if values of consecutive, unprefixed duplicates are equal
const foo = vendor.unprefixed(value) !== vendor.unprefixed(duplicateValue;
const bar = ignorePrefixlessSameValues && foo;

if (!duplicatesAreConsecutive || bar) {
  …
}
```
at line 107?

fpetrakov: Not sure how to name `bar` really. I guess, creating `unprefixedDuplicatesAreEqual` var is enough

fpetrakov: Also thought of moving repetitive fix logic to a separate function, what do you think?

Mouvedia: I am just making suggestions, you are the one who instigated that refactor, hence you should know better than me.

Mouvedia: I am just making suggestions, you are the one who instigated that refactor, hence you should know better than me.

All I can say is that
```
  ruleName,
  word: prop,
```
is currently missing after l141.
If the reports' arguments were identical you could have an argument-less function for both.
If the function requires arguments, I would advise against it.

fpetrakov:
It's done.

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -88,97 +88,68 @@ const rule = (primary, secondaryOptions, context) => {
...
+
+				if (ignoreDuplicates && duplicatesAreConsecutive) {
+					return;
+				}
+
+				if (!context.fix) {
+					return report({
 						message: messages.rejected(prop),
 						node: decl,
 						result,
```
Mouvedia, line 155: Is
```
  ruleName,
  word: prop,
```
missing?

fpetrakov: Why do you think it's missing? It's right there

Mouvedia: Not on github. Maybe you forgot to push?

Mouvedia:
> a separate function

You are correct you can reuse
```
const foo = () => {
	if (!context.fix) {
		return report({
			message: messages.rejected(prop),
			node: decl,
			result,
			ruleName,
			word: prop,
		});
	}

	if (duplicateIsMoreImportant) {
		return decl.remove();
	}

	return duplicateDecl.remove();
};
```
at line l113 and l151.

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -88,97 +88,60 @@ const rule = (primary, secondaryOptions, context) => {
...
+						return decl.remove();
 					}
 
-					if (context.fix) {
-						if (!important && duplicateImportant) {
-							decl.remove();
-						} else {
-							removePreviousDuplicate(decls, lowerProp);
-						}
+					if (!duplicateDecl) return;
```
Mouvedia, line 123: either this one is not necessary or the one at l97 is
i.e. if we reach l123 `duplicateDecl` is `true` because it's a const

fpetrakov: If you remove check at l123 `duplicateDecl` at l125 will be of type `Declaration | undefined`.

Mouvedia: at l97 you have
```
				if (!duplicateDecl) {
					return;
				}
```
the l123 check is called at l133 and l157 which are after l97.
It seems to me that `duplicateDecl` cannot be `undefined` at that stage because `Boolean(undefined) === false` hence it would have returned early.

fpetrakov: You can clone the branch

ybiquitous (approved):
Thank you. LGTM 👍🏼

jeddy3 (approved):
@fpetrakov Thank you for the refactor. It reads well.

fpetrakov:
> @fpetrakov Thank you for the refactor. It reads well.

Glad to help 🙌

PR #6547, opened 2022-12-28
Fix `overrides.files` in config to allow basename glob patterns

### LINKED ISSUE #6525: Fix `overrides.files` in config for glob patterns like `*.vue`

ybiquitous:

Closes #6525

This change fixes the `overrides.files` property in config to allow basename glob patterns like `*.css` or `*.html`.

For example, let's consider `stylelint-config-standard-vue`:

```js
{
  overrides: [
    {
      files: ["*.vue", "**/*.vue"],
      extends: [/* ... */],
    },
  ]
}
```

People should expect the `*.vue` pattern matches any `.vue` files in any directories.
See <https://github.com/ota-meshi/stylelint-config-standard-vue/blob/v1.0.0/lib/index.js#L6>

See also the `basename` option `micromatch`:
<https://github.com/micromatch/micromatch/tree/4.0.5#optionsbasename>

### DIFF
--- a/lib/augmentConfig.js
+++ b/lib/augmentConfig.js
@@ -20,6 +20,18 @@ const path = require('path');
 /** @typedef {import('stylelint').ResultProcessor} StylelintResultProcessor */
 
 /**
+ * @param {string} glob
+ * @param {string} basedir
+ * @returns {string}
+ */
+function absolutizeGlob(glob, basedir) {
+	const result = path.isAbsolute(glob.replace(/^!/, '')) ? glob : globjoin(basedir, glob);
+
+	// Glob patterns for micromatch should be in POSIX-style
+	return normalizePath(result);
+}
+
+/**
  * - Merges config and stylelint options
  * - Makes all paths absolute
  * - Merges extends
@@ -142,11 +154,7 @@ async function augmentConfigFull(stylelint, filePath, cosmiconfigResult) {
  */
 function absolutizePaths(config, configDir, cwd) {
 	if (config.ignoreFiles) {
-		config.ignoreFiles = [config.ignoreFiles].flat().map((glob) => {
-			if (path.isAbsolute(glob.replace(/^!/, ''))) return glob;
-
-			return globjoin(configDir, glob);
-		});
+		config.ignoreFiles = [config.ignoreFiles].flat().map((glob) => absolutizeGlob(glob, configDir));
 	}
 
 	if (config.plugins) {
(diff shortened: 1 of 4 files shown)


lib/augmentConfig.js
```diff
@@ -19,6 +19,18 @@ const path = require('path');
...
+ * @param {string} glob
+ * @param {string} basedir
+ * @returns {string}
+ */
+function absolutizeGlob(glob, basedir) {
+	const result = path.isAbsolute(glob.replace(/^!/, '')) ? glob : globjoin(basedir, glob);
+
+	// Glob patterns for micromatch should be in POSIX-style
+	return normalizePath(result);
+}
```
ybiquitous, line 32: [note] Created the `absolutizeGlob()` function to reduce duplications.

jeddy3 (approved):
LGTM, thank you!

ybiquitous:
Thanks for the review!

PR #6550, opened 2022-12-30
Add `media-feature-name-unit-allowed-list`

### LINKED ISSUE #6509: Add `media-feature-name-unit-allowed-list`

I'd like to enforce specific units for my media features so that they are consistent.

For example, only allowing `em` units for `width` media features:

```css
@media (width > 20em) {}
```

A new rule to do this.

- Name: `media-feature-name-unit-allowed-list`
- Primary option: `object: { "name": ["array", "of", "units"]|"unit" }`
- Secondary options: none
- Autofixable: None
- Message: `Unexpected unit "${unit}" for name "${name}"`
- Description: "Specify a list of allowed name and unit pairs within media features."
- Section: "Enforce conventions" -> "Allowed..."

[Spec ref for rule name](https://drafts.csswg.org/mediaqueries/#mq-features).

Prior art is our [`declaration-property-unit-allowed-list`](https://github.com/stylelint/stylelint/tree/main/lib/rules/declaration-property-unit-allowed-list) rule.

This is another candidate for using a new media query parser (https://github.com/stylelint/stylelint/issues/6501#issuecomment-1334618683).

mattxwang:

Closes #6509.

@romainmenke has asked a question about warning on values with no unit; my assumption is no, but I'd love to get that checked!

I've also left some notes where I could use some guidance. Since I'm still a relatively junior contributor (and I'm a bit rusty 😓 ), any advice is appreciated!

And, given all of their help, would be great to include the co-author commit for @romainmenke!


lib/rules/media-feature-name-unit-allowed-list/index.js
```diff
@@ -0,0 +1,94 @@
...
+					}
+
+					const nameList = flattenArray(primary[nameKey]);
+
+					if (!nameList) {
+						return;
+					}
+
+					entry.node.walk((childEntry) => {
+						if (!isTokenNode(childEntry.node) || childEntry.node.value[0] !== TokenType.Dimension) {
```
romainmenke, line 66: Should this rule warn when authors use values without unit or percentages?

lib/rules/media-feature-name-unit-allowed-list/index.js
```diff
@@ -0,0 +1,94 @@
...
+
+						if (nameList.includes(unit.toLowerCase())) {
+							return;
+						}
+
+						report({
+							message: messages.rejected(unit, name),
+							node: atRule,
+							index: childEntry.node.value[2],
+							endIndex: childEntry.node.value[3],
```
romainmenke, line 80: When I wrote this bit I didn't know what these values should be :)

After converting an existing rule I found that this works :

```
const index = atRuleParamIndex(atRule) + nameToken[2];
const endIndex = atRuleParamIndex(atRule) + nameToken[3] + 1;
```

lib/rules/media-feature-name-unit-allowed-list/index.js
mattxwang: My initial gut feeling is no; there are many media features that only have valid values that have no unit/percentage (see: [video-dynamic-range](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/video-dynamic-range) for an example of a number-based one). I've added a few tests that address various cases this rule should skip; not sure if this is the preferred way to test for this.

lib/rules/media-feature-name-unit-allowed-list/__tests__/index.js
```diff
@@ -0,0 +1,170 @@
...
+		{
+			code: '@media (grid: 0) { }',
+			description: 'unitless value, number',
+		},
+		{
+			code: '@media (video-dynamic-range: high) { }',
+			description: 'unitless value, keyword',
+		},
+	],
+});
```
mattxwang, line 170: Is this the appropriate way to write tests for these types of values?

lib/rules/media-feature-name-unit-allowed-list/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+			mediaQueryList.forEach((mediaQuery) => {
+				mediaQuery.walk((entry) => {
+					if (!isMediaFeaturePlain(entry.node) && !isMediaFeatureRange(entry.node)) {
+						return;
+					}
+
+					const name = isMediaFeatureRange(entry.node)
+						? entry.node.name.name.value[1].toString()
+						: entry.node.name.toString();
```
mattxwang, line 52: This differs from @romainmenke's suggested code since it didn't properly handle range contexts for Level 4 Mediq Queries. I only figured out this "hardcoded" position from inspecting the generated tree.

Should I:

- document this somewhere outside of this PR
- have a more robust way of doing this

Or, am I missing a feature of the library?

lib/rules/media-feature-name-unit-allowed-list/index.js
```diff
@@ -0,0 +1,99 @@
...
+					if (!nameList) {
+						return;
+					}
+
+					entry.node.walk((childEntry) => {
+						if (!isTokenNode(childEntry.node) || childEntry.node.value[0] !== TokenType.Dimension) {
+							return;
+						}
+
+						const unit = childEntry.node.value[4].unit;
```
mattxwang, line 73: Another "hardcoded" value indicated by @romainmenke.

lib/rules/media-feature-name-unit-allowed-list/index.js
```diff
@@ -0,0 +1,99 @@
...
+							return;
+						}
+
+						const atRuleIndex = atRuleParamIndex(atRule);
+
+						report({
+							message: messages.rejected(unit, name),
+							node: atRule,
+							index: atRuleIndex + childEntry.node.value[2],
+							endIndex: atRuleIndex + childEntry.node.value[3] + 1,
```
mattxwang, line 85: Two "hardcoded" values indicated by @romainmenke.

lib/rules/media-feature-name-unit-allowed-list/README.md
```diff
@@ -0,0 +1,69 @@
...
+Given:
+
+```json
+{
+  "width": "em",
+  "/height/": ["em", "rem"]
+}
+```
+
+The following patterns are considered problems:
```
mattxwang, line 27: Should I also demonstrate the level 4 media queries? Unsure if that's the convention we've been using for other rules.

mattxwang (commented):
Getting ready to mark this as review; leaving some comments where I could use some guidance!

lib/rules/media-feature-name-unit-allowed-list/index.js
```diff
@@ -0,0 +1,99 @@
...
+						matchesStringOrRegExp(name, nameIdentifier),
+					);
+
+					if (!nameKey) {
+						return;
+					}
+
+					const nameList = flattenArray(primary[nameKey]);
+
+					if (!nameList) {
```
mattxwang, line 64: aside: not sure how I could add a meaningful example here for code coverage

lib/rules/media-feature-name-unit-allowed-list/index.js
romainmenke: This will become simpler in a future version.
Most likely : `entry.node.getName()`

-----

Can you try `entry.node.name.getName()`?
This doesn't yet work for boolean media query features but we are only handling range and plain features here so the `getName` function exists for both.

`getName()` also ensures that you are using unescaped values.

```css
@media (\61 spect-ratio: 1/1) {}
@media (aspect-ratio: 1/1) {}
```

These are exactly the same but `name.toString()` will have the raw (escaped) form and `name.getName()` will have the unescaped form.

mattxwang: Ah, great suggestion - didn't realize this function existed. Just updated!

lib/rules/media-feature-name-unit-allowed-list/index.js
romainmenke: Correct :)

The check on line 69 ensures that this is a dimension token.
A dimension token has extra data :

```js
{
  unit: "px",
  value: 10,
  type: "integer"
}
```

----

Index `4` is where all the usable and unescaped token values can be found.
This is the same for all token types.

```
[<token type>, <raw contents>, <start index>, <end index>, <parsed contents>]
```

lib/rules/media-feature-name-unit-allowed-list/index.js
Mouvedia: I agree, it shouldn't.
e.g. `@media screen and (min-aspect-ratio: 16/9) {)` is unit less

.changeset/few-monkeys-fold.md
```diff
@@ -2,4 +2,4 @@
 "stylelint": minor
 ---
 
-Add `media-feature-name-unit-allowed-list`
+Added `media-feature-name-unit-allowed-list`
```
Mouvedia, line 5:
```suggestion
Added: `media-feature-name-unit-allowed-list` rule
```

lib/rules/media-feature-name-unit-allowed-list/index.js
```diff
@@ -0,0 +1,97 @@
...
+			actual: primary,
+			possible: [validateObjectWithArrayProps(isString)],
+		});
+
+		if (!validOptions) {
+			return;
+		}
+
+		root.walkAtRules(/^media$/i, (atRule) => {
+			const mediaQueryList = parse(atRule.params);
```
ybiquitous, line 42: [ask] If a parse error occurs from the `parse()` function, is it possible to recover the error like the `parseSelector()` utility?

https://github.com/stylelint/stylelint/blob/5e6d46aad5f61ef86f5823a67696a137a0e77a10/lib/utils/parseSelector.js#L13-L19

lib/rules/media-feature-name-unit-allowed-list/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+const messages = ruleMessages(ruleName, {
+	rejected: (unit, name) => `Unexpected unit "${unit}" for name "${name}"`,
+});
+
+const meta = {
+	url: 'https://stylelint.io/user-guide/rules/media-feature-name-unit-allowed-list',
+};
+
+/** @type {import('stylelint').Rule} */
```
ybiquitous, line 29: [trivial] We can make the code a bit type-safe:

```suggestion
/** @type {import('stylelint').Rule<Record<string, string | string[]>>} */
```

lib/rules/media-feature-name-unit-allowed-list/index.js
ybiquitous: [suggest] We can refactor the code to improve performance and readability as follows:

- Use `Object.entries()` instead of `Object.keys()`
- Memoize a return value of `Object.entries()`
- Extract a tiny internal utility
- Avoid using `flattenArray()`

```js
		if (!validOptions) {
			return;
		}

		const primaryPairs = Object.entries(primary);
		const primaryUnitList = (/** @type {string} */ featureName) => {
			for (const [name, unit] of primaryPairs) {
				if (matchesStringOrRegExp(featureName, name)) return [unit].flat();
			}

			return undefined;
		};
```

```js
		const featureName = node.name.getName();
		const unitList = primaryUnitList(featureName);

		if (!unitList) {
			return;
		}
```

lib/rules/media-feature-name-unit-allowed-list/index.js
ybiquitous: [suggest] I believe we can improve readability by using destructuring assignment like this:

```js
entry.node.walk(({ node: childNode }) => {
	if (!isTokenNode(childNode)) {
		return;
	}

	const [tokenType, , startIndex, endIndex, parsedContent] = childNode.value;

	if (tokenType !== TokenType.Dimension) {
		return;
	}
```

package-lock.json
```diff
@@ -32,7 +35,6 @@
         "micromatch": "^4.0.5",
         "normalize-path": "^3.0.0",
         "picocolors": "^1.0.0",
-        "postcss": "^8.4.20",
```
ybiquitous, line 35: [suggest] `postcss` moved `dependencies` to `devDependencies`. Could you restore it?

ybiquitous (commented):
Thank you for creating this pull request. I believe there are no significant problems. 👍🏼 

I've left some suggestions to improve the doc and code.

lib/rules/media-feature-name-unit-allowed-list/index.js
romainmenke: I was thinking of these types of issues :

```css
@media (width <= 100) {} /* number is not a length */
@media (width <= 50%) {} /* percentage is not a length */
@media (min-aspect-ratio: 16px/9px) {} /* length values do not make up a ratio */
```

But maybe that is beyond the scope of this rule?

ybiquitous: Good point. I believe such problems should be detected by a rule to [avoid errors](https://stylelint.io/user-guide/rules#avoid-errors). Perhaps we need a rule like `media-feature-no-invalid-range`.

lib/rules/media-feature-name-unit-allowed-list/index.js
romainmenke: This is not simple because the parser is forgiving and doesn't just throw whenever it encounters something unexpected.

Some details :
- the media query parser will never throw, if it does cause exceptions it would be a bug and must be fixed
- token stream level errors can be emitted
- component value level errors can be emitted
- there are no parser errors possible when applying the media query grammar to component values
- invalid media queries are component values that do not fit "todays" grammar
- invalid media queries can be preserved so that feedback can be given to authors

```ts
parse('something with parse errors', {
  onParseError: (err) => {
    // contains :
    // - tokenizer errors
    // - component value parser errors
    console.log(err);
  },
  preserveInvalidMediaQueries: true
});
```

Tokenizer/parser level errors are highly technical and will be rare.

Invalid media queries can not have a specific error messages because these are unknowns.
We only know that the component values do not fit the known grammar.

It is possible to emit a warning with `failed to parse X` but it is not possible to explain why it failed to parse it.

----

I suggest we handle this separately from this pull request?
Ideally we have good examples of error cases and the messages we want for each.

Then this work can be done once for all rules that would use this parser :)

ybiquitous: Thanks for the explanation. I understand well. As the first step, we can proceed as-is without considering edge cases.

Or, we may have the following utility for other rules using `@csstools/media-query-list-parser` (this PR doesn't need to add it, though)

```js
// lib/utils/parseMediaQuery.js

/**
 * @param {import('postcss').AtRule} atRule
 * @returns {ReturnType<typeof parse>}
 */
module.exports = function parseMediaQuery(atRule, result) {
	const source = atRule.params;
	return parse(source, {
		onParseError: (err) => {
			result.warn(`Cannot parse selector (${err})`, {
				node: atRule,
				word: source,
				stylelintType: 'parseError',
			});
		},
		preserveInvalidMediaQueries: true,
	});
};
```

mattxwang: If we'd like, I'm happy to open an issue to hammer out which cases need to be accounted for, and then draft a PR that uses this util stub + adds tests. Should I do that?

mattxwang:
Thank you for the in-depth review @ybiquitous, appreciate your attention to detail and good eye for readability. I believe I've addressed all the comments for this PR (tabling conversations about the `parseMediaQuery` util and a potential `media-feature-no-invalid-range` for later.

Also, this is an oversight on my part - should we be merging this into `v15` or `main`? I absentmindedly picked `v15` at the start since I assumed the major release was coming soon. I can do some interactive rebase hacking to get this into `main` instead!

lib/rules/media-feature-name-unit-allowed-list/index.js
ybiquitous: Yes, please.

ybiquitous (approved):
@mattxwang Thank you so much for addressing the reviews. Yes, the `v15` branch is suitable as a base. LGTM 👍🏼

lib/rules/media-feature-name-unit-allowed-list/index.js
mattxwang: I have created https://github.com/stylelint/stylelint/issues/6559; I'm not sure if I've captured the nuance of the request, feel free to chime in with more!

lib/rules/media-feature-name-unit-allowed-list/index.js
mattxwang: Drafted https://github.com/stylelint/stylelint/issues/6560. I struggled a bit with both the scope and behaviour of the rule; comments welcome!

PR #6553, opened 2023-01-02
fix: consider ruleless declarations as standard

43081j:
it seems this is leftover logic from back when we bundled various parts of the css-in-js solutions in the past.

by removing this, newer postcss syntaxes can be correctly used by stylelint (as they often have declarations with no parent rule).

cc @jeddy3

### DIFF
--- a/lib/utils/__tests__/isStandardSyntaxDeclaration.test.js
+++ b/lib/utils/__tests__/isStandardSyntaxDeclaration.test.js
@@ -166,6 +166,16 @@ describe('isStandardSyntaxDeclaration', () => {
 			true,
 		);
 	});
+
+	it('supports root-level declarations', () => {
+		const doc = new postcss.Document();
+		const root = postcss.parse('color: yellow;');
+
+		root.parent = doc;
+		doc.nodes.push(root);
+
+		expect(isStandardSyntaxDeclaration(root.nodes[0])).toBe(true);
+	});
 });
 
 function decl(css, parser = postcss) {
--- a/lib/utils/isStandardSyntaxDeclaration.js
+++ b/lib/utils/isStandardSyntaxDeclaration.js
@@ -1,14 +1,7 @@
 'use strict';
 
 const isScssVariable = require('./isScssVariable');
-const { isRoot, isRule } = require('./typeGuards');
-
-/**
- * @param {string} [lang]
- */
-function isStandardSyntaxLang(lang) {
-	return lang && (lang === 'css' || lang === 'custom-template' || lang === 'template-literal');
-}
+const { isRule } = require('./typeGuards');
 
 /**
  * Check whether a declaration is standard
@@ -18,18 +11,6 @@ function isStandardSyntaxLang(lang) {
 module.exports = function isStandardSyntaxDeclaration(decl) {
 	const prop = decl.prop;
 	const parent = decl.parent;
-
-	// Declarations belong in a declaration block or standard CSS source
-	if (
-		parent &&
-		isRoot(parent) &&
-		parent.source &&
-		!isStandardSyntaxLang(
-			/** @type {import('postcss').Source & {lang?: string}} */ (parent.source).lang,
-		)
-	) {
-		return false;
-	}
 
 	// SCSS var; covers map and list declarations
 	if (isScssVariable(prop)) {


jeddy3 (approved):
Thank you, LGTM.

Ref: https://github.com/43081j/postcss-styled-components/issues/2#issuecomment-1369105896

lib/utils/isStandardSyntaxDeclaration.js
```diff
@@ -19,18 +12,6 @@ module.exports = function isStandardSyntaxDeclaration(decl) {
 	const prop = decl.prop;
```
Mouvedia, line 12: `isStandardSyntaxDeclaration` is used in at least 7 rules. Are we sure this is not gonna cause regressions?  
Is being more lax really covered by tests already?

Mouvedia:
It seems to have been introduced in #3933.
Is it still relevant?

43081j:
It seems to have originally existed to detect sources coming from the old css-in-js implementation. Something in the past must have been patching up the node source to contain the `lang` property it was looking for. However, that isn't a property within postcss (if it ever was), the css-in-js code must have been adding it in. 

Those old syntaxes are deprecated afaik and their newer replacements won't set such a property anymore

Mouvedia:
If there are still code bases where the condition is necessary, it should be an option/flag.
The default would be what you are asking for: no parent allowed.

ybiquitous:
I'm not sure, but @Mouvedia's point may be correct. `@stylelint/postcss-css-in-js` package depends on `postcss-syntax`, and the `lang` property is set in the package. So, problems may come for configs or plugins depending on `@stylelint/postcss-css-in-js`. See also:

- https://github.com/stylelint/postcss-css-in-js/blob/0.38.0/index.js#L4
- https://github.com/stylelint/postcss-css-in-js/blob/0.38.0/template-parser-helper.js#L209-L211

Mouvedia:
> So, problems may come for configs or plugins depending on `@stylelint/postcss-css-in-js`.

[It has 49 dependent packages.](https://www.npmjs.com/package/@stylelint/postcss-css-in-js?activeTab=dependents)

ybiquitous:
[suggest] For compatibility with `@stylelint/postcss-css-in-js` and `postcss-syntax`, what about changing the logic as below?

```js
/**
 * @param {unknown} lang
 */
function isStandardSyntaxLang(lang) {
	return lang && (lang === 'css' || lang === 'custom-template' || lang === 'template-literal');
}

// ...

module.exports = function isStandardSyntaxDeclaration(decl) {
	// ...

	// Declarations belong in a declaration block or standard CSS source
	if (
		parent &&
		isRoot(parent) &&
		parent.source &&
		'lang' in parent.source &&
		!isStandardSyntaxLang(parent.source.lang)
	) {
		return false;
	}

	// ...
```

43081j:
i feel like you may be misunderstanding.

the original PR that introduced it didn't come from nothing, it changed from this:

```ts
  if (parent.type === "root") {
    return false;
  }
```

to this:

```ts
  if (parent.type === "root" && !isStandardSyntaxLang(parent.source.lang)) {
    return false;
  }
```

the former would have caused css-in-js syntaxes (**old or new** ones) to always return false in this statement when they have a top-level declaration. the latter just added an escape hatch for those declarations produced by css-in-js syntaxes (such that the `if` would be skipped).

before:
* if the declaration is a direct child of a root, it is non-standard

after:
* if the declaration is a direct child of a root **and isn't from a css-in-js syntax**, it is non-standard

by removing it entirely here, we're saying declarations which are a direct child of a root are _always_ standard/valid.

importantly, we're _not_ going back to the code from before the PR you mentioned. we're getting rid of it entirely.

this means all declarations which are children of roots will be valid, _including_ the ones from before.

we're expanding the set, not reducing it. that regression shouldn't be affected.

Mouvedia:
You are correct.
I guess the approach of waiting for issues to be raised _if_ ppl somehow relied on
https://github.com/stylelint/stylelint/blob/26fa669cd0c28cbd5dc1f94e8dc5a9812bc6da07/src/utils/isStandardDeclaration.js#L10
to catch standard declaration is a pragmatic choice.

ybiquitous:
@43081j Thank you for your explanation. Indeed, I misunderstood. 😅 

> by removing it entirely here, we're saying declarations which are a direct child of a root are always standard/valid, not only when produced by css-in-js syntaxes.

This approach is reasonable. And I believe it will not affect both `@stylelint/postcss-css-in-js` and `postcss-syntax`.

lib/utils/__tests__/isStandardSyntaxDeclaration.test.js
```diff
@@ -166,6 +166,16 @@ describe('isStandardSyntaxDeclaration', () => {
...
 	});
+
+	it('supports root-level declarations', () => {
+		const doc = new postcss.Document();
+		const root = postcss.parse('color: yellow;');
+
+		root.parent = doc;
+		doc.nodes.push(root);
+
+		expect(isStandardSyntaxDeclaration(root.nodes[0])).toBe(true);
```
ybiquitous, line 177: [suggest] If we'd like to focus on "root-level", using `Document` seems confusing. Instead, what about simplifying the test code like this?

```suggestion
		expect(isStandardSyntaxDeclaration(decl('color: yellow;'))).toBe(true);
```

43081j: no preference here but fyi the reason i did it this way was basically to ensure there's a test somewhere that makes sure it works with declarations deeply contained in a document (don't think there are any such tests yet).

at the min the implementation doesn't care, it was just in case someone some day updates this function and breaks it for those cases. its not actually testing any existing code branch

ybiquitous: I've applied the suggestion via 54a5098. Thanks.

ybiquitous (approved):
Thank you! LGTM 👍🏼

PR #6558, opened 2023-01-05
Add `function-url-quotes` autofix

### LINKED ISSUE #6500: Add `function-url-quotes` autofix

The rule doesn't autofix.

To have it autofix as it's:

- the only [`*-quotes` rule](https://stylelint.io/user-guide/rules#quotes) that doesn't
- [turned on](https://github.com/stylelint/stylelint-config-standard/blob/main/index.js#L92) (alongside [`url` imports](https://github.com/stylelint/stylelint-config-standard/blob/main/index.js#L95)) in our standard config

I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if anyone has time.

There are [steps to add autofix to a rule](https://stylelint.io/developer-guide/rules#add-autofix) in the Developer guide.

mattxwang:

Closes #6500.

~~Two possible code smells I want to call out. In both cases, I opted to pick a "dirty" solution that doesn't change existing behaviour and minimizes/localizes the changed code. However, there are probably more correct/elegant ways to do this. Advice is welcome!~~

I was able to address these thanks to code review, but I'll leave the comments for posterity.

---

First, this rule operates on both declarations and at-rules. It seems like in `checkArgs`, this type information is "erased". To resolve this, I have my utility functions `addQuotes` and `removeQuotes` take in a union type, and then use a crude `'params' in node` to differentiate the type. 

This ... feels wrong? I was a bit unsure if there's a better way for me to do this (either from a TS perspective, or perhaps a Stylelint util that I'm missing).

---

Secondly, the previous iteration of this rule lowercases declarations and at-rule parameters before passing them in as arguments to `functionArgumentsSearch`. This ensures that the rule matches `UrL` and the like.

However, this has the side effect of lowercasing the declaration value and/or at-rule parameters. Then, when I apply the autofix, I propagate this lowercase change (which means that the autofix also inadvertently lowercases the code).

To resolve this problem, I pass in a case-insensitive regex to `functionArgumentsSearch`. I don't think this is *technically* wrong, but it is likely a performance regression; I didn't want to change `functionArgumentsSearch` implementation instead. With that in mind, any suggestions on better ways to do this?

Thanks in advance - implementing this autofix was trickier than I thought!

### DIFF
--- a/lib/rules/function-url-quotes/__tests__/index.js
+++ b/lib/rules/function-url-quotes/__tests__/index.js
@@ -5,6 +5,7 @@ const { messages, ruleName } = require('..');
 testRule({
 	ruleName,
 	config: ['always'],
+	fix: true,
 
 	accept: [
 		{
@@ -180,6 +181,7 @@ testRule({
 	reject: [
 		{
 			code: '@import url(foo.css);',
+			fixed: '@import url("foo.css");',
 			message: messages.expected('url'),
 			line: 1,
 			column: 13,
@@ -188,6 +190,7 @@ testRule({
 		},
 		{
 			code: '@import url( foo.css );',
+			fixed: '@import url( "foo.css" );',
 			message: messages.expected('url'),
 			line: 1,
 			column: 14,
@@ -196,6 +199,7 @@ testRule({
 		},
 		{
 			code: '@document url-prefix(http://www.w3.org/Style);',
+			fixed: '@document url-prefix("http://www.w3.org/Style");',
 			message: messages.expected('url-prefix'),
 			line: 1,
 			column: 22,
@@ -204,6 +208,7 @@ testRule({
 		},
 		{
 			code: '@document url-prefix( http://www.w3.org/Style );',
+			fixed: '@document url-prefix( "http://www.w3.org/Style" );',
 			message: messages.expected('url-prefix'),
 			line: 1,
 			column: 23,
@@ -212,6 +217,7 @@ testRule({
 		},
 		{
 			code: "@font-face { font-family: 'foo'; src: url(foo.ttf); }",
+			fixed: '@font-face { font-family: \'foo\'; src: url("foo.ttf"); }',
 			message: messages.expected('url'),
 			line: 1,
 			column: 43,
@@ -220,6 +226,7 @@ testRule({
 		},
 		{
 			code: "@font-face { font-family: 'foo'; src: url( foo.ttf ); }",
+			fixed: '@font-face { font-family: \'foo\'; src: url( "foo.ttf" ); }',
 			message: messages.expected('url'),
 			line: 1,
 			column: 44,
@@ -228,6 +235,7 @@ testRule({
 		},
 		{
 			code: 'a { cursor: url(foo.png); }',
+			fixed: 'a { cursor: url("foo.png"); }',
 			message: messages.expected('url'),
 			line: 1,
 			column: 17,
@@ -236,6 +244,7 @@ testRule({
 		},
 		{
 			code: 'a { background-image: url(foo.css), url("bar.css"), url("baz.css"); }',
+			fixed: 'a { background-image: url("foo.css"), url("bar.css"), url("baz.css"); }',
 			message: messages.expected('url'),
 			line: 1,
 			column: 27,
@@ -244,6 +253,7 @@ testRule({
 		},
 		{
 			code: 'a { background-image: url( foo.css ), url("bar.css"), url("baz.css"); }',
+			fixed: 'a { background-image: url( "foo.css" ), url("bar.css"), url("baz.css"); }',
 			message: messages.expected('url'),
 			line: 1,
 			column: 28,
@@ -252,6 +262,7 @@ testRule({
 		},
 		{
 			code: 'a { background-image: url("foo.css"), url(bar.css), url("baz.css"); }',
+			fixed: 'a { background-image: url("foo.css"), url("bar.css"), url("baz.css"); }',
 			message: messages.expected('url'),
 			line: 1,
 			column: 43,
@@ -260,6 +271,7 @@ testRule({
 		},
 		{
 			code: 'a { background-image: url("foo.css"), url( bar.css ), url("baz.css"); }',
+			fixed: 'a { background-image: url("foo.css"), url( "bar.css" ), url("baz.css"); }',
 			message: messages.expected('url'),
 			line: 1,
 			column: 44,
@@ -268,6 +280,7 @@ testRule({
 		},
 		{
 			code: 'a { background-image: url("foo.css"), url("bar.css"), url(baz.css); }',
+			fixed: 'a { background-image: url("foo.css"), url("bar.css"), url("baz.css"); }',
 			message: messages.expected('url'),
 			line: 1,
 			column: 59,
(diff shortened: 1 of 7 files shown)


lib/rules/function-url-quotes/index.js
```diff
@@ -58,22 +72,64 @@ const rule = (primary, secondaryOptions) => {
...
+		 * @param {import('postcss').Declaration | import('postcss').AtRule} node
 		 * @param {string} args
-		 * @param {import('postcss').Node} node
+		 * @param {number} index
+		 */
+
+		function addQuotes(node, args, index) {
+			const fixedName = `"${args}"`;
+			const openIndex =
+				'params' in node ? index - atRuleParamIndex(node) : index - getDeclarationValueIndex(node);
```
mattxwang, line 95: nit: I'm not sure if the following is clearer (or an alternative)
```suggestion
				index - ('params' in node ? atRuleParamIndex(node) : getDeclarationValueIndex(node));
```

Or, setting a variable to the value of `'params' in node ?  atRuleParamIndex(node) :  getDeclarationValueIndex(node)`, etc.

lib/rules/function-url-quotes/index.js
```diff
@@ -19,8 +19,22 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/function-url-quotes',
 };
 
+/**
+ * @param {import('postcss').Declaration} decl
+ * @returns {number}
+ */
+function getDeclarationValueIndex(decl) {
```
mattxwang, line 26: nit: I didn't see a `get` util for this. Have I missed something? Alternatively, should this go into `utils/`?

ybiquitous: [suggest] `declarationValueIndex` may be used here:

https://github.com/stylelint/stylelint/blob/b4fd2d50f588c5c4adb7f63a1091393b427f1a13/lib/utils/declarationValueIndex.js#L9

lib/rules/function-url-quotes/index.js
```diff
@@ -19,8 +19,22 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/function-url-quotes',
```
ybiquitous, line 19: [suggest] `fixable: true` in the metadata is required:

```suggestion
	url: 'https://stylelint.io/user-guide/rules/function-url-quotes',
	fixable: true,
```

lib/rules/function-url-quotes/__tests__/index.js
```diff
@@ -710,6 +710,148 @@ testRule({
 	],
 });
 
+testRule({
```
ybiquitous, line 713: [suggest] Can you add `fix: true` to the existing `testRule` to reduce duplication?

https://github.com/stylelint/stylelint/blob/3b7a7d0a348494120d6f01d3d4acf949780d37bf/lib/rules/function-url-quotes/__tests__/index.js#L5-L7

lib/rules/function-url-quotes/index.js
ybiquitous: Both seem acceptable, though I prefer the suggested code.

ybiquitous:
> First, this rule operates on both declarations and at-rules. It seems like in `checkArgs`, this type information is "erased". To resolve this, I have my utility functions `addQuotes` and `removeQuotes` take in a union type, and then use a crude `'params' in node` to differentiate the type.

Good point. I think it will be more readable to split into 2 functions with a single purpose. For example:

```js
if (node.type === 'atrule') {
	addQuotesForAtRule(node, trimmedArg, complaintIndex);
} else {
	addQuotesForDecl(node, trimmedArg, complaintIndex);
}
```

The same is true for `removeQuotes`.

ybiquitous:
> To resolve this problem, I pass in a case-insensitive regex to `functionArgumentsSearch`. I don't think this is *technically* wrong, but it is likely a performance regression; I didn't want to change `functionArgumentsSearch` implementation instead. With that in mind, any suggestions on better ways to do this?

I think the solution makes sense and will not degrade performance much. But benchmarking is beneficial. Please consider sharing benchmark results via the `npm run benchmark-rule` script:

https://github.com/stylelint/stylelint/blob/b4fd2d50f588c5c4adb7f63a1091393b427f1a13/docs/developer-guide/rules.md?plain=1#L267

mattxwang:
Thanks @ybiquitous for the quick review, I appreciate it! I believe I've addressed all the comments.

In addition, thanks for suggesting benchmarks! I ran them (results below), and it seems like there's no sizeable impact on the rule behaviour with or without autofix. I'm unsure if I'm using the commands correctly, small timescale, or variance on my local machine (M1 macOS) represents all other use cases, but it seems reasonable. 

---

## Benchmarks

### With Autofix

```
$ npm run benchmark-rule -- function-url-quotes always "{\"fix\": \"true\"}"

> stylelint@14.16.1 benchmark-rule
> node scripts/benchmark-rule.mjs function-url-quotes always {"fix": "true"}

Warnings: 0
Mean: 13.63671329657534 ms
Deviation: 3.1167337864783593 ms
```

```
$ npm run benchmark-rule -- function-url-quotes never "{\"fix\": \"true\"}"

> stylelint@14.16.1 benchmark-rule
> node scripts/benchmark-rule.mjs function-url-quotes never {"fix": "true"}

Warnings: 0
Mean: 13.534337736986304 ms
Deviation: 2.8618163492841986 ms
```

### Without Autofix

Note: we should expect no change, since this PR doesn't touch non-autofix behaviour.

HEAD of this PR:

```
$ npm run benchmark-rule -- function-url-quotes always

> stylelint@14.16.1 benchmark-rule
> node scripts/benchmark-rule.mjs function-url-quotes always

Warnings: 0
Mean: 13.244889730666669 ms
Deviation: 2.8424667783141677 ms
```

```
$ npm run benchmark-rule -- function-url-quotes never 

> stylelint@14.16.1 benchmark-rule
> node scripts/benchmark-rule.mjs function-url-quotes never

Warnings: 14
Mean: 13.58528738150685 ms
Deviation: 3.2287915266088905 ms
```

`main` right now:

```
$ npm run benchmark-rule -- function-url-quotes always

> stylelint@14.16.1 benchmark-rule
> node scripts/benchmark-rule.mjs function-url-quotes always

Warnings: 0
Mean: 13.292001378666669 ms
Deviation: 3.100038171534227 ms
```

```
$ npm run benchmark-rule -- function-url-quotes never

> stylelint@14.16.1 benchmark-rule
> node scripts/benchmark-rule.mjs function-url-quotes never

Warnings: 14
Mean: 13.420950610810808 ms
Deviation: 3.1724774966395293 ms
```

ybiquitous (approved):
Thank you for addressing the reviews and sharing the performance results. LGTM! 👍🏼

Mouvedia:
@mattxwang does it use the `string-quotes` rule value—if not set to `null`—to choose between `"` and `'` for the fix?

mattxwang:
> @mattxwang does it use the `string-quotes` rule value—if not set to `null`—to choose between `"` and `'` for the fix?
> 
> If not, that fix will require multiple runs. i.e. #3853

Ah - no, it doesn't. I implemented this similar to the behaviour in the other two quotes rules (`font-family-name-quotes`, `selector-attribute-quotes`) which both default to `"` as the quote. I believe I've added similar behaviour (i.e. in all three autofix cases, multiple runs are needed).

Is there past precedence / a blueprint for a rule inspecting the value of another rule? It may make sense to simply fix all of these at once in one PR (I am happy to contribute this if desired).

Mouvedia:
> It may make sense to simply fix all of these at once in one PR (I am happy to contribute this if desired).

If you don't I probably will. Either way, you need an issue for it: TBD (lemme add it)

Issue #6409, opened 2022-10-13
Prepare 15.0.0


jeddy3:
- https://github.com/stylelint/stylelint/issues/6311
- https://github.com/stylelint/stylelint/issues/3408
- https://github.com/stylelint/stylelint/issues/6342
- https://github.com/stylelint/stylelint/issues/6414

Discussed:

- https://github.com/stylelint/stylelint/issues/6483

Pull requests:

- https://github.com/stylelint/stylelint/pull/6380
- https://github.com/stylelint/stylelint/pull/6323
- https://github.com/stylelint/stylelint/pull/6332

There's no fixed date for when this might happen. We'll need to investigate what's possible and what we've time to do.

This is an umbrella issue for tracking our progress and for broad discussions. Let's use the individual issues for specific discussions.

Ref: Some initial discussion in https://github.com/stylelint/stylelint/issues/6311.

ybiquitous:
@jeddy3 Thanks for listing up the tasks.

It seems we can remove the deprecated `syntax` option. Is there a related issue?

https://github.com/stylelint/stylelint/blob/d40288983161124d7b8a29be219835bf7ea6841d/types/stylelint/index.d.ts#L243-L244

jeddy3:
> It seems we can remove the deprecated syntax option.

I agree.

> Is there a related issue?

I don't believe so. Let's create one.

ybiquitous:
Thanks. I've opened #6414 and updated this issue description.

ybiquitous:
I've created the [`v15`](https://github.com/stylelint/stylelint/tree/v15) branch just now. Let's push all the breaking changes into the branch.

ybiquitous:
I've cleaned up some tasks. Thank you for your reviews, @jeddy3.

The ESM task is the biggest, so I believe it's recommended to work on the task after reducing the code as possible.

Mouvedia:
I think #6483 should be added to the list of the blocking issues.

ybiquitous:
I've added #6483 to the list above.

ntwb:
⭐ Stylelint hit 10,000 stars this weekend

jeddy3:
> @ybiquitous The ESM task is the biggest, so I believe it's recommended to work on the task after reducing the code as possible.

Yes, I agree that the ESM task is extensive and we want to reduce the code first. 

I suggest we move the ESM task to v16. I think dropping support for node 12 and deprecating a lot of stylistic rules is enough for this major release.

[...]

PR #6561, opened 2023-01-05
Add deprecated rule warnings

ybiquitous:

Ref: #6409

To notify deprecation and make the migration easier.
Such warnings are added to a linting result and reporting output.

For example of a report:

```console
$ cd scripts
$ ../bin/stylelint.js visual.css --config=visual-config.json

Deprecation Warning: The "string-quotes" rule is deprecated and will be removed in the next major release.
Deprecation Warning: The "unit-case" rule is deprecated and will be removed in the next major release.

visual.css
 1:9   ✖  Expected "url("foo.css")" to be ""foo.css""       import-notation
 1:13  ✖  Expected single quotes                            string-quotes
 4:9   ✖  Unexpected invalid hex color "#4f"                color-no-invalid-hex
 6:1   ✖  Expected ".foo.bar" to have no more than 1 class  selector-max-class
 7:10  ✖  Unexpected unit "px"                              unit-disallowed-list
 8:3   ⚠  Expected custom property name to be kebab-case    custom-property-pattern

6 problems (5 errors, 1 warning)
```


lib/lintPostcssResult.js
```diff
@@ -131,4 +135,14 @@ function isFixCompatible({ stylelint }) {
...
 
-module.exports = lintPostcssResult;
+/**
+ * @param {PostcssResult} result
+ * @param {string} ruleName
+ * @returns {void}
+ */
+function warnDeprecatedRule(result, ruleName) {
+	result.warn(
+		`The "${ruleName}" rule is deprecated and will be removed in the next major release.`,
```
jeddy3, line 145:
```suggestion
		`The "${ruleName}" rule is deprecated.`,
```

Let's keep it succinct in the terminal to avoid a wall of text.

jeddy3 (changes requested):
@ybiquitous I've made one minor suggestion, otherwise LGTM.

lib/lintPostcssResult.js
Mouvedia: What about having just one sentence with a list instead of multiple sentences?
e.g. `The "string-quotes", "unit-case" and "foo-bar" rules are deprecated and will be removed in the next major release.`

ybiquitous: Thanks for the advice. I've simplified warning messages via ff83ba86c. Here is a screenshot:

image

I keep multiple sentences because each message may have a `reference`, and it's hard to include reference(s) in one sentence. See the code below:

https://github.com/stylelint/stylelint/blob/ff83ba86caa9d5b3f9db47903fbb29ec1f6bb98d/lib/formatters/stringFormatter.js#L59-L61

Instead, I've avoided repeated "Deprecation Warning" messages in each line and appended a " - " prefix to each line.

What about this change?

types/stylelint/index.d.ts
```diff
@@ -323,7 +323,7 @@ declare module 'stylelint' {
 		source?: string;
 		deprecations: {
 			text: string;
-			reference: string;
+			reference: string | undefined;
```
ybiquitous, line 326: [note] for more accuracte type.

lib/lintPostcssResult.js
Mouvedia: Let's take stylelint config standard as the baseline.
Let's say a user resets about 19% of the deprecated rules; that clocks at around 12 lines out of potentially 64.
In that scénario I would prefer a counter and the usage of `…` or `etc.` for more than 3.
e.g.
```js
`${count} deprecation warnings for rules ${rules[0]},  ${rules[1]}, ${rules[2]}, etc.`
```

types/stylelint/index.d.ts
Mouvedia: [question]

Should it be optional instead?
e.g. `reference?: string;`

i.e. reference must be set when you are explicit
e.g.
```
				deprecations: [
					{
						text: 'Deprecated foo',
						reference: undefined,
					},
				],
```

lib/lintPostcssResult.js
jeddy3: > What about this change?

SGTM, thank you.

lib/formatters/stringFormatter.js
```diff
@@ -47,22 +47,23 @@ function deprecationsFormatter(results) {
...
+		if (reference) {
+			line += dim(` See: ${underline(reference)}`);
 		}
 
-		return `${output}\n`;
-	}, '\n');
+		lines.push(line);
+	}
+
+	return ['', yellow('Deprecation Warning:'), ...lines, ''].join('\n');
```
jeddy3, line 66:
```suggestion
	return ['', yellow('Deprecation warnings:'), ...lines, ''].join('\n');
```

Minor nit, otherwise LGTM to merge.

jeddy3 (changes requested):

types/stylelint/index.d.ts
ybiquitous: Good point. Fixed via d6108a993.

lib/formatters/stringFormatter.js
ybiquitous: Nice. Fixed via bec7cdc7c.

lib/lintPostcssResult.js
ybiquitous: @Mouvedia I understand your concern. But, if `reference` is specified with a rule deprecation, how should we display it with `…` or `etc.`?　I believe we need to consider also plugin rules, not only the built-in ones.

ybiquitous:
Maybe off-topic, but I found the following description in our developer guide:

https://github.com/stylelint/stylelint/blob/f129f3b992accbf37c7d532d0d020c62c4f3749a/docs/developer-guide/rules.md?plain=1#L278

Indeed, the `stylelintReference` property seems not to be used:

```console
$ git grep -w 'stylelintReference' lib/
lib/createPartialStylelintResult.js:32:                reference: deprecationMessage.stylelintReference,
```

Should we follow the guideline still? If doing that, this PR change would show a large amount of text for users.

lib/lintPostcssResult.js
Mouvedia: Alright, as long as we a ENV variable—or something similar—that hides those warnings it should be OK.
@ybiquitous Does the `--quiet` option hide these warnings?

ollie-iterators:
When will this be getting merged?

jeddy3:
> When will this be getting merged?

The original author usually merges once a pull request has one approval. So, it'll be merged whenever @ybiquitous has time. We're all volunteers so things happen when they happen.

> Also, once this PR gets merged, will it cause https://github.com/stylelint/stylelint/pull/6573 to be able to be merged so that stylelint can be updated to version 15.0.0?

Yes, that's likely. I'll update https://github.com/stylelint/stylelint/issues/6409 to bring things together.

lib/lintPostcssResult.js
ybiquitous: > Does the `--quiet` option hide these warnings?

I'm afraid the answer is **no** because these warnings are not related to a *warning* severity produced by lint rules. We currently don't have a way to suppress this kind of warning, except for using a custom formatter.

ybiquitous:
Because we will deprecate many rules in v15.0.0, some may see many deprecation warnings and feel annoyed. But I believe showing warnings should be better than showing nothing.

Also, if people update `stylelint` and `stylelint-config-{recommended,standard}` at the same time, they will not see the warnings, since the configs will remove deprecation rules.

We can consider adding an option to turn off deprecation warnings if people need it.

If there is no other feedback, I will merge this PR within a few days.

jeddy3:
> Also, if people update `stylelint` and `stylelint-config-{recommended,standard}` at the same time, they will not see the warnings, since the configs will remove deprecation rules.

Yep, exactly. 

> If there is no other feedback, I will merge this PR within a few days.

Merging SGTM.

ybiquitous:
> I will for the 15.1.0 version.

Let's open an issue when it becomes necessary.

lib/lintPostcssResult.js
```diff
@@ -131,4 +135,11 @@ function isFixCompatible({ stylelint }) {
...
 }
 
-module.exports = lintPostcssResult;
+/**
+ * @param {PostcssResult} result
+ * @param {string} ruleName
+ * @returns {void}
+ */
+function warnDeprecatedRule(result, ruleName) {
+	result.warn(`The "${ruleName}" rule is deprecated.`, { stylelintType: 'deprecation' });
```
Mouvedia, line 144: @ybiquitous

This PR was merged before #7550.
Should we use `process.emitWarning` instead?
i.e. `--no-deprecation` doesn't suppress these warnings

ybiquitous: Good point. We must consider backward compatibility if switching to `process.emitWarning` would make sense.

For example, we may need to update the following document:
https://github.com/stylelint/stylelint/blob/1ba1a0aeafaa124a4987a0135d511111cec3af0a/docs/developer-guide/rules.md?plain=1#L296-L301

Can you open a new issue for the discussion?

Mouvedia: I considered this not important since the deprecation was added 2 majors ago.
But [this](https://deps.dev/npm/stylelint/16.7.0/versions) made me realize that a lot dependents are still stuck on old majors.
e.g. WordPress/gutenberg#48142

> Can you open a new issue for the discussion?

OK.

PR #6565, opened 2023-01-08
Fix `isStandardSyntaxFunction` false positives for interpolation and backticks in CSS-in-JS

### LINKED ISSUE #6564: `function-no-unknown` false positives for interpolation and backticks in CSS-in-JS

CSS-in-JS code with interpolation `${}` or with backticks <code>`</code>:

```js
const Component = styled.div`
	border-radius: ${(props) => props.shape === 'pill' ? token.radiusPill : token.radiusBase};
	border-radius: ${(props) => props.shape === 'pill' ? `calc(${token.radiusPill} + 5px)` : token.radiusBase};
`;
```

Issues reproduced with `postcss-styled-syntax` and with `@stylelint/postcss-css-in-js`.

```json
{
	"rules": {
		"function-no-unknown": true
	}
}
```

stylelint "**/*.js"

`14.16.1`

No violations.

### What actually happened?

```
test.ts
 2:17  ✖  Unexpected unknown function "${"     function-no-unknown
 3:17  ✖  Unexpected unknown function "${"     function-no-unknown
 3:55  ✖  Unexpected unknown function "`calc"  function-no-unknown
```

CSS-in-JS

### Proposal to fix the bug

Update [`isStandardSyntaxFunction`](https://github.com/stylelint/stylelint/blob/v15/lib/utils/isStandardSyntaxFunction.js) to ignore such syntax.

hudochenkov:

Fixes #6564

Tests for the rule are skipped, because there is no CSS-in-JS syntax added to the repo yet. ~~Waiting for https://github.com/stylelint/stylelint/pull/6556.~~

### DIFF
--- a/lib/rules/function-no-unknown/__tests__/index.js
+++ b/lib/rules/function-no-unknown/__tests__/index.js
@@ -71,6 +71,22 @@ testRule({
 });
 
 testRule({
+	skip: true,
+	ruleName,
+	config: true,
+	customSyntax: 'postcss-styled-syntax',
+
+	accept: [
+		{
+			code: 'const buttonFontSize = css`font-size: ${({ size }) => (size === "small") ? "0.8em" : "1em"};`;',
+		},
+		{
+			code: 'const buttonFontSize = css`border-radius: ${`calc(${token.radiusBase} + 2px)`};`;',
+		},
+	],
+});
+
+testRule({
 	ruleName,
 	config: [true, { ignoreFunctions: ['theme', '/^foo-/', /^bar$/i] }],
 
--- a/lib/rules/function-no-unknown/__tests__/index.js
+++ b/lib/rules/function-no-unknown/__tests__/index.js
@@ -71,22 +71,6 @@ testRule({
 });
 
 testRule({
-	skip: true,
-	ruleName,
-	config: true,
-	customSyntax: 'postcss-styled-syntax',
-
-	accept: [
-		{
-			code: 'const buttonFontSize = css`font-size: ${({ size }) => (size === "small") ? "0.8em" : "1em"};`;',
-		},
-		{
-			code: 'const buttonFontSize = css`border-radius: ${`calc(${token.radiusBase} + 2px)`};`;',
-		},
-	],
-});
-
-testRule({
 	ruleName,
 	config: [true, { ignoreFunctions: ['theme', '/^foo-/', /^bar$/i] }],
 
--- a/lib/utils/__tests__/isStandardSyntaxFunction.test.js
+++ b/lib/utils/__tests__/isStandardSyntaxFunction.test.js
@@ -26,6 +26,30 @@ describe('isStandardSyntaxFunction', () => {
 			false,
 		);
 	});
+
+	it('CSS-in-JS interpolation', () => {
+		const functions = [];
+
+		valueParser('${({ size }) => (size === "small") ? "0.8em" : "1em"}').walk((valueNode) => {
+			if (valueNode.type === 'function') {
+				functions.push(valueNode);
+			}
+		});
+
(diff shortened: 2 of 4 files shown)


lib/utils/__tests__/isStandardSyntaxFunction.test.js
```diff
@@ -26,6 +26,30 @@ describe('isStandardSyntaxFunction', () => {
 			false,
 		);
 	});
+
+	it('CSS-in-JS interpolation', () => {
+		const functions = [];
+
+		valueParser('${({ size }) => (size === "small") ? "0.8em" : "1em"}').walk((valueNode) => {
```
hudochenkov, line 33: Because there is no CSS-in-JS test yet in the repository, skipping parsing CSS as with other tests in this file, and go directly to parsing the value.

lib/rules/function-no-unknown/__tests__/index.js
```diff
@@ -70,6 +70,22 @@ testRule({
...
+	accept: [
+		{
+			code: 'const buttonFontSize = css`font-size: ${({ size }) => (size === "small") ? "0.8em" : "1em"};`;',
+		},
+		{
+			code: 'const buttonFontSize = css`border-radius: ${`calc(${token.radiusBase} + 2px)`};`;',
+		},
+	],
+});
+
```
ybiquitous, line 88: [suggest] I think this PR fix should focus on `isStandardSyntaxFunction()`, so what about reverting the test for `function-no-unknown`? See also my idea details for https://github.com/stylelint/stylelint/pull/6556#issuecomment-1375183292

lib/utils/__tests__/isStandardSyntaxFunction.test.js
ybiquitous: > go directly to parsing the value.

Good point. Furthermore, what about rewriting the `func` helper in this test file and the existing test cases as below?

```js
it('calc', () => {
	expect(isStandardSyntaxFunction(func('calc(a + b)'))).toBe(true);
});

// ...

it('CSS-in-JS syntax', () => {
	expect(isStandardSyntaxFunction(func('`calc(${token.radiusBase} + 2px)`'))).toBe(false);
});

// ...

function func(declValue) {
	const functions = [];

	valueParser(declValue).walk((valueNode) => {
		if (valueNode.type === 'function') {
			functions.push(valueNode);
		}
	});

	return functions[0];
}
```

ybiquitous (commented):
Thanks for creating the pull request. If addressing my reviews, it may be good to update the PR title like this:

```
Fix `isStandardSyntaxFunction` false positives for interpolation and backticks in CSS-in-JS
```

hudochenkov:
@ybiquitous I agree with your suggestion. I've updated PR title, but I don't think we need to mention internal function in a changelog. Because users don't care about internal functions, but they care about rules.

ybiquitous (approved):
Agreed. LGTM. Thank you 👍🏼

lib/utils/__tests__/isStandardSyntaxFunction.test.js
```diff
@@ -1,42 +1,51 @@
...
-			false,
-		);
+	it('scss function in scss interpolation', () => {
+		expect(isStandardSyntaxFunction(getFunction('#{darken(#fff, 0.2)}'))).toBe(false);
+	});
+
+	it('CSS-in-JS interpolation', () => {
+		expect(
+			isStandardSyntaxFunction(
+				getFunction('${({ size }) => (size === "small") ? "0.8em" : "1em"}'),
```
43081j, line 32: this is great, exactly what i was getting at in the other issue - a generic test for all css-in-js function nodes.

nice work 🎉

PR #6571, opened 2023-01-09
Fix `selector-attribute-quotes` false positives for "never"

### LINKED ISSUE #4300: Fix false positives for "never" in selector-attribute-quotes

We try avoid any unnecessary quotes in selectors, so we use `"selector-attribute-quotes": "never"`.

But some selectors really require quotes to be valid.

```less
div[style='*'] { ... }
li[data-id='42'] { ... }
```

And here is why:

According to the spec http://www.w3.org/TR/CSS2/selector.html#attribute-selectors:
> Attribute values must be identifiers or strings

Furthermore, according to the spec http://www.w3.org/TR/CSS2/syndata.html#value-def-identifier: 
> identifiers (including element names, classes, and IDs in selectors) can contain only the characters [a-zA-Z0-9] and ISO 10646 characters U+00A0 and higher, plus the hyphen (-) and the underscore (_); they cannot start with a digit, two hyphens, or a hyphen followed by a digit.

And, according to the spec http://www.w3.org/TR/CSS2/syndata.html#strings:
> Strings can either be written with double quotes or with single quotes.

A new option for `selector-attribute-quotes` (e.g. "only-required" or "only-necessary"), which would work like "never", but allow/require quotes when the correspondent part of selector:
* contains chars other than latin letters, digits, hyphen and underscore
* starts with a digit
* starts with double hyphen
* starts with a hyphen followed by a digit

mattxwang:

Closes #4300.

First - should I add some documentation to explain this new behaviour? Or do we think the current setup is enough.

There's a bit of discussion in #4300. To reiterate @Mouvedia's point: 

> so to me
> ```
> [href='te\'s\'t'] { }
> [href="te\"s\"t"] { }
> ```
> 
> can be processed and become
> ```
> [href=te\'s\'t] { }
> [href=te\"s\"t] { }
> ```
> 
> but
> ```
> [href="te'st"] { }
> [href='te"st'] { }
> ```
> 
> should remain untouched.

As such, I've adjusted the relevant test cases. The latter two are moved from being rejected to accepted; let me know if that was the desired change (or if instead we should still report, but not fix).

### Regex alternative

I realized that the CSS grammar actually has [a FLEX tokenizer](https://www.w3.org/TR/CSS21/grammar.html#scanner), so I tried to directly copy the rule as a Regex:

```js
// directly taken from spec: https://www.w3.org/TR/CSS21/grammar.html#scanner
const h = "[0-9a-f]";
const nonascii = "[\\240-\\377]";
const unicode = "\\\\{h}{1,6}(\\r\\n|[ \\t\\r\\n\\f])?".replace("{h}", h);
const escape = "({unicode}|\\\\[^\\r\\n\\f0-9a-fA-F])".replace("{unicode}", unicode);
const nmstart = "([_a-zA-Z]|{nonascii}|{escape})".replace("{nonascii}", nonascii).replace("{escape}", escape);
const nmchar = "([_a-zA-Z0-9-]|{nonascii}|{escape})".replace("{nonascii}", nonascii).replace("{escape}", escape);
const identifier = "-?{nmstart}{nmchar}*".replace("{nmstart}", nmstart).replace("{nmchar}", nmchar);
	
return Boolean(ident.match(identifier));
```

Unfortunately, this doesn't work! It misses some "basic" cases, including strings with spaces in them. If anybody has any ideas on what I've done wrong, that would be great - I feel like using the CSS spec when possible seems like a good idea.

### DIFF
--- a/.changeset/chilly-ties-trade.md
+++ b/.changeset/chilly-ties-trade.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: false positives for "never" in `selector-attribute-quotes`
--- a/.changeset/chilly-ties-trade.md
+++ b/.changeset/chilly-ties-trade.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: false positives for "never" in `selector-attribute-quotes`
+Fixed: `selector-attribute-quotes` false positives for "never"
--- a/lib/rules/selector-attribute-quotes/index.js
+++ b/lib/rules/selector-attribute-quotes/index.js
@@ -66,8 +66,13 @@ const rule = (primary, _secondaryOptions, context) => {
 
 					if (quoted && primary === 'never') {
 						// some selectors require quotes to be valid;
+						// we pass in the raw string value, which contains the escape characters
+						// necessary to check if escaped characters are valid
 						// see: https://github.com/stylelint/stylelint/issues/4300
-						if (!isValidIdentifier(value)) {
+						if (
+							!attributeNode.raws.value ||
+							!isValidIdentifier(attributeNode.raws.value.slice(1, -1))
+						) {
 							return;
 						}
 
--- a/lib/rules/selector-attribute-quotes/index.js
+++ b/lib/rules/selector-attribute-quotes/index.js
@@ -2,6 +2,7 @@
 
 const getRuleSelector = require('../../utils/getRuleSelector');
 const isStandardSyntaxRule = require('../../utils/isStandardSyntaxRule');
+const isValidIdentifier = require('../../utils/isValidIdentifier');
 const parseSelector = require('../../utils/parseSelector');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
@@ -64,6 +65,12 @@ const rule = (primary, _secondaryOptions, context) => {
 					}
 
 					if (quoted && primary === 'never') {
+						// some selectors require quotes to be valid;
+						// see: https://github.com/stylelint/stylelint/issues/4300
+						if (!isValidIdentifier(value)) {
+							return;
+						}
+
 						if (context.fix) {
 							selectorFixed = true;
 							attributeNode.quoteMark = null;
--- a/lib/utils/__tests__/isValidIdentifier.test.js
+++ b/lib/utils/__tests__/isValidIdentifier.test.js
@@ -7,12 +7,11 @@ it('isValidIdentifier', () => {
 	expect(isValidIdentifier('foo1')).toBeTruthy();
 	expect(isValidIdentifier('1')).toBeFalsy();
 	expect(isValidIdentifier('-1')).toBeFalsy();
-	expect(isValidIdentifier('--foo')).toBeFalsy();
+	expect(isValidIdentifier('--foo')).toBeTruthy();
 	expect(isValidIdentifier('has a space')).toBeFalsy();
 	expect(isValidIdentifier(null)).toBeFalsy();
 	expect(isValidIdentifier('')).toBeFalsy();
 	expect(isValidIdentifier(' ')).toBeFalsy();
-	expect(isValidIdentifier(' --foo')).toBeFalsy();
 	expect(isValidIdentifier('foö')).toBeFalsy();
 	expect(isValidIdentifier('\\26 B')).toBeTruthy(); // ISO 10646 character
 	expect(isValidIdentifier('\\000026B')).toBeTruthy(); // ISO 10646 character
--- a/lib/utils/__tests__/isValidIdentifier.test.js
+++ b/lib/utils/__tests__/isValidIdentifier.test.js
@@ -13,5 +13,13 @@ it('isValidIdentifier', () => {
 	expect(isValidIdentifier('')).toBeFalsy();
 	expect(isValidIdentifier(' ')).toBeFalsy();
 	expect(isValidIdentifier(' --foo')).toBeFalsy();
-	expect(isValidIdentifier(' --foö')).toBeFalsy();
+	expect(isValidIdentifier('foö')).toBeFalsy();
+	expect(isValidIdentifier('\\26 B')).toBeTruthy(); // ISO 10646 character
+	expect(isValidIdentifier('\\000026B')).toBeTruthy(); // ISO 10646 character
+	expect(isValidIdentifier("te'st")).toBeFalsy(); // unescaped quotation mark
+	expect(isValidIdentifier('te"st')).toBeFalsy(); // unescaped quotation mark
+	expect(isValidIdentifier("te\\'st")).toBeTruthy(); // escaped (in CSS) quotation mark
+	expect(isValidIdentifier('te\\"st')).toBeTruthy(); // escaped (in CSS) quotation mark
+	expect(isValidIdentifier("te\\'s\\'t")).toBeTruthy(); // escaped (in CSS) quotation mark
+	expect(isValidIdentifier('te\\"s\\"t')).toBeTruthy(); // escaped (in CSS) quotation mark
 });
--- a/lib/utils/__tests__/isValidIdentifier.test.js
+++ b/lib/utils/__tests__/isValidIdentifier.test.js
@@ -1,0 +1,17 @@
+'use strict';
+
+const isValidIdentifier = require('../isValidIdentifier');
+
+it('isValidIdentifier', () => {
+	expect(isValidIdentifier('foo')).toBeTruthy();
+	expect(isValidIdentifier('foo1')).toBeTruthy();
+	expect(isValidIdentifier('1')).toBeFalsy();
+	expect(isValidIdentifier('-1')).toBeFalsy();
+	expect(isValidIdentifier('--foo')).toBeFalsy();
+	expect(isValidIdentifier('has a space')).toBeFalsy();
+	expect(isValidIdentifier(null)).toBeFalsy();
+	expect(isValidIdentifier('')).toBeFalsy();
+	expect(isValidIdentifier(' ')).toBeFalsy();
+	expect(isValidIdentifier(' --foo')).toBeFalsy();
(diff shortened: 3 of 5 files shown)


lib/rules/selector-attribute-quotes/index.js
```diff
@@ -64,6 +65,17 @@ const rule = (primary, _secondaryOptions, context) => {
...
 
 					if (quoted && primary === 'never') {
+						// some selectors require quotes to be valid;
+						// we pass in the raw string value, which contains the escape characters
+						// necessary to check if escaped characters are valid
+						// see: https://github.com/stylelint/stylelint/issues/4300
+						if (
+							!attributeNode.raws.value ||
+							!isValidIdentifier(attributeNode.raws.value.slice(1, -1))
+						) {
```
mattxwang, line 75: Admittedly, this feels off - but I (think?) that the `raws` is the only way I can keep the escape characters in the value (opposed to `attributeNode.value`, which unescapes everything) so I can check if the resulting string is a valid identifier, etc. Suggestions on how to make this clearer are welcome!

lib/utils/isValidIdentifier.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
```
mattxwang, line 2: Should this be its own file? And hopefully, I haven't missed an existing version of this util!

lib/utils/isValidIdentifier.js
```diff
@@ -0,0 +1,41 @@
...
+	// ident has a leading digit
+	if (hasDigitAt(0)) {
+		return false;
+	}
+
+	// ident has a leading dash followed by a digit or a dash
+	if (trimmedIdent.charAt(0) === '-' && (hasDigitAt(1) || trimmedIdent.charAt(1) === '-')) {
+		return false;
+	}
```
mattxwang, line 38: Would we prefer for me to collapse this into one expression? I did it this way since I felt that it was more naturally readable.

Alt: see my note on the full FLEX-based regex option.

mattxwang (commented):
This ended up being a bit more of a deep dive into the CSS spec than I thought! I think this PR is now ready for review. I've left some comments where I'm unsure of how I'm doing.

lib/utils/isValidIdentifier.js
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * returns whether a string is a valid CSS identifier
+ * (i.e. only alphanumeric characters, `-`, and `_`;
+ * does not have a leading digit, leading dash followed by digit, or two leading dashes)
+ * furthermore, any escaped or ISO 10646 characters are allowed.
+ * see: https://www.w3.org/TR/CSS2/syndata.html#value-def-identifier
+ * @param {string} ident
+ * @returns {boolean}
+ */
+module.exports = function isValidIdentifier(ident) {
```
Mouvedia, line 12: question

I dunno the stylelint conventions, but did you check if a npm already existed?
Even if it does maybe we recommend not relying on dependencies if possible.

lib/utils/__tests__/isValidIdentifier.test.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const isValidIdentifier = require('../isValidIdentifier');
+
+it('isValidIdentifier', () => {
+	expect(isValidIdentifier('foo')).toBeTruthy();
+	expect(isValidIdentifier('foo1')).toBeTruthy();
+	expect(isValidIdentifier('1')).toBeFalsy();
+	expect(isValidIdentifier('-1')).toBeFalsy();
+	expect(isValidIdentifier('--foo')).toBeFalsy();
```
Mouvedia, line 10: > [This was later relaxed to allow `--` at the start of identifiers with the introduction of custom properties.](https://mathiasbynens.be/notes/unquoted-attribute-values)

I think it should valid.

mattxwang: Sounds good!

lib/utils/isValidIdentifier.js
mattxwang: Hm, to be honest I didn't look too in-depth; I did a quick `site:npmjs.com valid css identifier` and found nothing on the first page or so of Google (much more support for valid JS identifiers or CSS selectors). If you're aware of one and we think we should use it though, happy to refactor!

lib/rules/selector-attribute-quotes/index.js
ybiquitous: [suggest] What about using the [`Attribute#smartQuoteMark()` method of `postcss-selector-parser`](https://github.com/postcss/postcss-selector-parser/blob/v6.0.11/src/selectors/attribute.js#L132) instead of the custom logic?

```js
// The value ident is invalid if the attribute needs a quote mark.
if (typeof attributeNode.smartQuoteMark({}) === 'string') {
	return;
}
```

The method doc says:

> Intelligently select a quoteMark value based on the value's contents. If the value is a legal CSS ident, it will not be quoted. Otherwise a quote mark will be picked that minimizes the number of escapes.

The method's implementation relies on the [`cssesc`](https://github.com/mathiasbynens/cssesc) package, which seems reliable and popular (22M downloads/week).

mattxwang: Thanks for the suggestion @ybiquitous, this is very cool - I hadn't heard about it before!

However, I'm concerned that it doesn't completely resolve our problem. In particular, if I use that instead of `isValidIdentifier`, the [cases identified by \@Mouvedia](https://github.com/stylelint/stylelint/issues/4300#issuecomment-1379256755) are not handled well:

```js
[href='te\\'s\\'t'] { }  // currently autofixable, now missed
[href="te\\"s\\"t"] { }  // currently autofixable, now missed
```

I think this is because the [`Attribute#smartQuoteMark()` method](https://github.com/postcss/postcss-selector-parser/blob/v6.0.11/src/selectors/attribute.js#L132) uses the `value` of the attribute, and not the `raws`; the escaped quotes aren't preserved, and so it incorrectly believes that the attribute is not autofixable.

However, I may be using the function wrong or am missing something. Do you have any suggestions?

Mouvedia: If cssesc doesn't do the job just keep what you had, it's probably good enough.

IMHO adding a new dependency should have a minimum requirement.
e.g. used by 3 rules

lib/utils/isValidIdentifier.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+/**
+ * returns whether a string is a valid CSS identifier
```
ybiquitous, line 4: [trivial]

```suggestion
 * Returns whether a string is a valid CSS identifier
```

lib/utils/isValidIdentifier.js
```diff
@@ -0,0 +1,41 @@
...
+ * (i.e. only alphanumeric characters, `-`, and `_`;
+ * does not have a leading digit, leading dash followed by digit, or two leading dashes)
+ * furthermore, any escaped or ISO 10646 characters are allowed.
+ * see: https://www.w3.org/TR/CSS2/syndata.html#value-def-identifier
```
ybiquitous, line 8: [trivial] The JSDoc tag has better IDE support.

```suggestion
 * @see https://www.w3.org/TR/CSS2/syndata.html#value-def-identifier
```

lib/utils/isValidIdentifier.js
```diff
@@ -0,0 +1,41 @@
...
+	const hasDigitAt = (/** @type {number} */ i) => !isNaN(parseInt(ident.charAt(i)));
+
+	// trims, removes ISO 10646 characters, and singly-escaped characters
+	const trimmedIdent = ident
+		.trim()
+		.replace(/\\[0-9a-f]{1,6}(\\r\\n|[ \t\r\n\f])?/gi, '')
+		.replace(/\\./g, '');
+
+	// characters that are not alphanumeric, hyphen or underscore
+	if (trimmedIdent.match(/[^\w-]/)) {
```
ybiquitous, line 26: [suggest] Using `RegExp#test()` instead of `String#match()` seems readable and a bit performant because `RegExp#test()` returns a boolean and doesn't create an array.

```suggestion
	if (/[^\w-]/.test(trimmedIdent)) {
```

lib/utils/isValidIdentifier.js
```diff
@@ -0,0 +1,41 @@
...
+
+	// ident has a leading digit
+	if (hasDigitAt(0)) {
+		return false;
+	}
+
+	// ident has a leading dash followed by a digit
+	if (trimmedIdent.charAt(0) === '-' && hasDigitAt(1)) {
+		return false;
+	}
```
ybiquitous, line 38: [suggest] What about the following refactorings?

- inline `hasDigit()` because the inner function seems a bit too abstract
- replace the combo of `isNaN()` and `parseInt()` with a single regex
- remove comments because the code is clear enough, and the spec is described in the doc of `isValidIdentifier()`

```suggestion
	if (/\d/.test(trimmedIdent.charAt(0))) {
		return false;
	}

	if (trimmedIdent.charAt(0) === '-' && /\d/.test(trimmedIdent.charAt(1))) {
		return false;
	}
```

mattxwang:
Thanks for the quick review @ybiquitous! Think I've addressed each of the comments, let me know what you think.

.changeset/chilly-ties-trade.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: false positives for "never" in `selector-attribute-quotes`
```
ybiquitous, line 5: [suggest] Can you please make the format consistent by following the current guideline? (I've changed the PR title also)

https://github.com/stylelint/stylelint/blob/005ed45ff170f3e3351a610c6cdd8931f6731f49/docs/maintainer-guide/issues.md?plain=1#L41

```suggestion
Fixed: `selector-attribute-quotes` false positives for "never"
```

ybiquitous (approved):
Thank you. LGTM 👍🏼

PR #6580, opened 2023-01-17
Add `declaration-block-no-redundant-longhand-properties` autofix

### LINKED ISSUE #3304: Add `declaration-block-no-redundant-longhand-properties` autofix

mattxwang:

Closes #3304.

See my review comment below - I think this probably needs some reviewing!

In particular, this PR involved me writing a *lot* of manual code/is prone to error.


lib/reference/properties.js
```diff
@@ -311,55 +320,70 @@ const longhandSubPropertiesOfShorthandProperties = new Map([
...
 		'transition',
-		new Set([
-			'transition-delay',
-			'transition-duration',
+		[
+			// prettier-ignore
 			'transition-property',
+			'transition-duration',
 			'transition-timing-function',
-		]),
```
mattxwang, line 359: I *think* this is the only one that had the "wrong" order - see the CSS spec for `single-transition` (ref: [CSS spec](https://w3c.github.io/csswg-drafts/css-transitions/#single-transition)) for the right order.

lib/reference/properties.js
```diff
@@ -311,55 +320,70 @@ const longhandSubPropertiesOfShorthandProperties = new Map([
...
+	for (const [shorthand, orderedLonghand] of orderedLonghandSubPropertiesOfShorthandProperties) {
+		map.set(shorthand, new Set(orderedLonghand));
+	}
+
+	return map;
+}
+
+/** @type {import('stylelint').LonghandSubPropertiesOfShorthandProperties} */
+const longhandSubPropertiesOfShorthandProperties =
+	convertToLonghandSubPropertiesOfShorthandProperties();
```
mattxwang, line 385: I did this instead of using `.entries().map(...)` and then making a `Map` because it seems like `Object.entries` didn't preserve the type of `orderedLonghandSubPropertiesOfShorthandProperties` properly.

This feels a bit hacky - is there better code I could be doing here (or something I've made a mistake  on)

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -42,10 +43,13 @@ const rule = (primary, secondaryOptions) => {
 			return;
 		}
 
-		/** @type {Map<string, string[]>} */
+		/** @type {Map<string, import('stylelint').ShorthandProperties[]>} */
```
mattxwang, line 46: Pretty sure the previous type was slightly incorrect, but could be wrong!

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -80,21 +86,69 @@ const rule = (primary, secondaryOptions) => {
...
+
+								return values
+									.map((shorthandProp) => {
+										const val = transformedDeclarationNodes.get(shorthandProp);
+
+										return val !== undefined ? val.value : '';
+									})
+									.join(' ')
+									.replace(/\s{2,}/g, ' ');
+							};
```
mattxwang, line 137: This feels inefficient - any suggestions?

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -80,21 +86,69 @@ const rule = (primary, secondaryOptions) => {
...
+
+					if (!arrayEqual(copiedPrefixedShorthandData.sort(), longhandDeclaration.sort())) {
 						continue;
 					}
 
+					if (context.fix) {
+						const declNodes = longhandDeclarationNodes.get(prefixedShorthandProperty) || [];
+
+						if (declNodes !== undefined && declNodes[0] !== undefined) {
+							const resolveShorthandValues = () => {
```
mattxwang, line 116: Should I pull out this function? The nesting is pretty nasty; however, I wasn't sure if separating it (and adding several arguments) would make the function less readable; this version is "in-context".

mattxwang (commented):
Just wrapped up a draft for a first pass on this feature. Broadly, I use a similar approach to @imvetri, except:

- instead of redefining the properties, I convert the existing `longhandSubPropertiesOfShorthandProperties` to contain an (ordered) array, and then recreate `longhandSubPropertiesOfShorthandProperties` by calling the `Set` constructor - hopefully reduces code de-dup while requiring few code changes outside of this PR
- then, I use the orders to "replace" the prop + values of the first node where a longhand declaration was used - adjusting @imvetri's original solution

I'd like to get feedback on:

- broadly, does this approach make sense?
- how many more tests should I write (I am considering one per shorthand)
- how am I doing with naming? I found the existing code slightly confusing to read through, but also recognize that I've added some more complexity
- am I doing TS-related stuff properly?

This is certainly a complicated-ish PR and requires a bit of double-checking. I think this PR is ready to review, but I'll certainly do some more rigorous testing (likely adding more test cases) just in case.

lib/reference/properties.js
Mouvedia: The examples on https://developer.mozilla.org/en-US/docs/Web/CSS/transition#syntax seems to indicate that
`name | duration | easing function` and `name | duration | delay` are both valid.

lib/reference/properties.js
```diff
@@ -311,55 +320,70 @@ const longhandSubPropertiesOfShorthandProperties = new Map([
...
+		],
 	],
 	[
 		'text-decoration',
-		new Set([
+		[
 			// prettier-ignore
 			'text-decoration-color',
 			'text-decoration-style',
 			'text-decoration-line',
```
Mouvedia, line 350: 4th postion `text-decoration-thickness` is missing

Mouvedia, line 350:
```suggestion
			'text-decoration-line',
			'text-decoration-style',
			'text-decoration-color',
			'text-decoration-thickness',
```

mattxwang: Thanks for catching the wrong ordering - fixed it!

Where are you getting `text-decoration-thickness` from? The [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration#formal_syntax) seem to indicate that the syntax is

```
text-decoration = 
  <'text-decoration-line'>   ||
  <'text-decoration-style'>  ||
  <'text-decoration-color'>  

```

lib/reference/properties.js
mattxwang: Yeah, I think the general order is agnostic (that's what the [double bar means](https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#double_bar)), moreso that the *first* time value is parsed as a duration, and the *second* is a delay; from your link,

> The first value that can be parsed as a time is assigned to the [transition-duration](https://developer.mozilla.org/en-US/docs/Web/CSS/transition-duration), and the second value that can be parsed as a time is assigned to [transition-delay](https://developer.mozilla.org/en-US/docs/Web/CSS/transition-delay).

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.js
```diff
@@ -5,6 +5,7 @@ const { messages, ruleName } = require('..');
 testRule({
 	ruleName,
 	config: [true],
+	fix: true,
```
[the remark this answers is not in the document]
mattxwang: Not sure - happy to add that (likely in a separate PR?) if we'd like.

ybiquitous:
@mattxwang Thanks for creating the pull request, and sorry for the late reply.

The [`Set` doc on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) says:

> You can iterate through the elements of a set in insertion order.

So, is it possible to use `longhandSubPropertiesOfShorthandProperties` instead of adding `orderedLonghandSubPropertiesOfShorthandProperties`? Of course, sub properties in `longhandSubPropertiesOfShorthandProperties` should be sorted to correspond to shorthand properties (you done it already). If possible, it would be no longer needed to have similar named variables (they are a part of the Stylelint public API).

lib/reference/properties.js
Mouvedia: > Note: The property used to be called text-decoration-width, but was updated in 2019 to text-decoration-thickness.

Should we support both `text-decoration-width` and `text-decoration-thickness`?

> Where are you getting text-decoration-thickness from?

> This property, which is a [sub-property](https://w3c.github.io/csswg-drafts/css-cascade-5/#longhand) of the [text-decoration](https://w3c.github.io/csswg-drafts/css-text-decor-4/#propdef-text-decoration) shorthand…

ref: https://w3c.github.io/csswg-drafts/css-text-decor-4/#text-decoration-width-property

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.js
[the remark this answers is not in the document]
jeddy3: @Mouvedia Can you open a new false negative bug report, please? So that we can address that in a separate issue and pull request.

lib/reference/properties.js
mattxwang: Ah! Thanks for pointing me to the spec; perhaps the MDN docs need to be updated?

The [spec ref for the shorthand](https://w3c.github.io/csswg-drafts/css-text-decor-4/#text-decoration-property) indicates that the syntax is:

[<'text-decoration-line'>](https://w3c.github.io/csswg-drafts/css-text-decor-4/#propdef-text-decoration-line) [||](https://w3c.github.io/csswg-drafts/css-values-4/#comb-any) [<'text-decoration-thickness'>](https://w3c.github.io/csswg-drafts/css-text-decor-4/#propdef-text-decoration-thickness) || [<'text-decoration-style'>](https://w3c.github.io/csswg-drafts/css-text-decor-4/#propdef-text-decoration-style) || [<'text-decoration-color'>](https://w3c.github.io/csswg-drafts/css-text-decor-4/#propdef-text-decoration-color).

I can definitely add thickness, but that would also change the existing behaviour of this rule. What do we think?

mattxwang:
Thanks for the suggestion @ybiquitous, that's a much cleaner way to approach this! I've reverted most of the changes regarding `orderedLonghandSubPropertiesOfShorthandProperties` in [1263e9e](https://github.com/stylelint/stylelint/pull/6580/commits/1263e9e6429e9183028cea2b23cfbf1758f0d971), and the test still pass, etc.

lib/reference/properties.js
Mouvedia: I guess you could put that and the inset thing in the same enhancement issue.

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.js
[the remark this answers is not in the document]
Mouvedia: @jeddy3 #6613

lib/reference/properties.js
Mouvedia: Can you put a TODO comment so that it's not forgotten?

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -61,6 +62,8 @@ const rule = (primary, secondaryOptions) => {
 		eachDeclarationBlock(root, (eachDecl) => {
 			/** @type {Map<string, string[]>} */
 			const longhandDeclarations = new Map();
+			/** @type {Map<string, import('postcss').Declaration[]>} */
```
Mouvedia, line 65: minor

maybe `@typedef` it once since you are using it several times

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
Mouvedia: opinion

Personally I prefer having a separate fixer function at l110, if it doesn't have too many arguments. In your case there's too many so it would have to be in the same scope.

The reason being that it prepares the refactor needed for https://github.com/stylelint/stylelint/issues/2643#issuecomment-535406358

lib/rules/declaration-block-no-redundant-longhand-properties/README.md
```diff
@@ -66,6 +66,8 @@ This rule complains when the following shorthand properties can be used:
 
 Flexbox-related properties can be ignored using `ignoreShorthands: ["/flex/"]` (see below).
 
+The [`fix` option](../../../docs/user-guide/usage/options.md#fix) can automatically fix most of the problems reported by this rule.
```
Mouvedia, line 69: minor

You could add an `unfixable: true,` test to illustrate that.

mattxwang:
Resolved two quick merge conflicts:

- the link to the "Options" page / `fix` subheading has changed since I opened this PR
- `index.d.ts` was slimmed down (I re-applied my changes to the current `main`)

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -80,21 +85,69 @@ const rule = (primary, secondaryOptions) => {
...
+					const copiedPrefixedShorthandData = [...prefixedShorthandData];
+
+					if (!arrayEqual(copiedPrefixedShorthandData.sort(), longhandDeclaration.sort())) {
 						continue;
 					}
 
+					if (context.fix) {
+						const declNodes = longhandDeclarationNodes.get(prefixedShorthandProperty) || [];
+
+						if (declNodes !== undefined && declNodes[0] !== undefined) {
```
ybiquitous, line 114: [suggest] We can simplify the code by checking just the first node like this:

```suggestion
						const [firstDeclNode] = declNodes;

						if (firstDeclNode) {
```

ybiquitous:
> Final question: would we prefer more test cases (`unfixable` or otherwise)? If so, happy to write them up.

The current cases look enough, thank you. If a problem is found, then let's add cases.

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
mattxwang: Ah, this is super clean - thank you for this suggestion! Also excited to use optional chaining more frequently :)

ybiquitous (approved):
Great! LGTM 👍🏼

PR #6592, opened 2023-01-25
Add `ignoreFunctions: []` to `unit-disallowed-list`

### LINKED ISSUE #6574: Add `ignoreFunctions: []` to `unit-disallowed-list`

Use CSS functions that include a disallowed unit character inside a function that expects it, like calc.
```css
h2 {
  position: absolute;
  inset-inline-end: calc(50% - 100px);
}
```

```json
{
  "rules": {
        "unit-disallowed-list": [
            ["px"],
            {
                ignoreFunctions: ["calc"]
            }
        ],
  }
}
```

```sh
stylelint **/*.css
```

14.11.0

CLI would finish with no errors

### What actually happened?

```sh
Invalid Option: Invalid option name "ignoreFunctions" for rule "unit-disallowed-list"
```

No

### Proposal to fix the bug

A solution similar to #5857

mattxwang:

Closes #6574.

Hopefully not!

### DIFF
--- a/lib/rules/unit-disallowed-list/README.md
+++ b/lib/rules/unit-disallowed-list/README.md
@@ -167,30 +167,17 @@ For example, with `["px"]`.
 Given:
 
 ```json
-{
-  "ignoreFunctions": ["calc", "/^translate/"]
-}
+"ignoreFunctions": ["calc", "/^translate/"]
 ```
 
 The following patterns are _not_ considered problems:
 
 <!-- prettier-ignore -->
 ```css
-a {
-  margin: calc(50% - 100px)
-}
+a { margin: calc(50% - 100px) }
 ```
 
 <!-- prettier-ignore -->
 ```css
-a {
-  translateX: calc(50% - 100px)
-}
-```
-
-<!-- prettier-ignore -->
-```css
-a {
-  translateY: calc(50% - 100px)
-}
+a { transform: translateX(50% - 100px) }
 ```
--- a/lib/rules/unit-disallowed-list/README.md
+++ b/lib/rules/unit-disallowed-list/README.md
@@ -157,3 +157,40 @@ The following patterns are considered problems:
 ```css
 @media print and (max-resolution: 100dpi) {}
 ```
+
+### `ignoreFunctions: ["function", "/regex/", /regex/]|"function"|"/regex/"|/regex/`
+
+Ignore units that are inside of the specified functions.
+
+For example, with `["px"]`.
+
+Given:
+
+```json
+{
+  "ignoreFunctions": ["calc", "/^translate/"]
+}
+```
+
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+a {
+  margin: calc(50% - 100px)
+}
+```
+
+<!-- prettier-ignore -->
+```css
+a {
+  translateX: calc(50% - 100px)
+}
+```
+
+<!-- prettier-ignore -->
+```css
+a {
+  translateY: calc(50% - 100px)
+}
+```
--- a/lib/rules/unit-disallowed-list/README.md
+++ b/lib/rules/unit-disallowed-list/README.md
@@ -167,7 +167,7 @@ For example, with `["px"]`.
 Given:
 
 ```json
-"ignoreFunctions": ["calc", "/^translate/"]
+["calc", "/^translate/"]
 ```
 
 The following patterns are _not_ considered problems:
@@ -179,5 +179,5 @@ a { margin: calc(50% - 100px) }
 
 <!-- prettier-ignore -->
 ```css
-a { transform: translateX(50% - 100px) }
+a { transform: translateX(100px) }
 ```
--- a/lib/rules/unit-disallowed-list/__tests__/index.js
+++ b/lib/rules/unit-disallowed-list/__tests__/index.js
@@ -392,6 +392,67 @@ testRule({
 	ruleName,
 
 	config: [
+		['px'],
+		{
+			ignoreFunctions: ['calc', /^translate/],
+		},
+	],
+
+	accept: [
+		{
+			code: 'a { margin: calc(50% - 100px) }',
+		},
+		{
+			code: 'a { translateX: calc(50% - 100px) }',
+		},
+		{
+			code: 'a { translateY: calc(50% - 100px) }',
+		},
+		{
+			code: 'a { translateZ: calc(50% - 100px) }',
+		},
+	],
+
+	reject: [
+		{
+			code: 'a { margin: 100px }',
+			message: messages.rejected('px'),
+			line: 1,
+			column: 16,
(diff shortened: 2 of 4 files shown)


lib/rules/unit-disallowed-list/README.md
```diff
@@ -157,3 +157,40 @@ The following patterns are considered problems:
...
+Ignore units that are inside of the specified functions.
+
+For example, with `["px"]`.
+
+Given:
+
+```json
+{
+  "ignoreFunctions": ["calc", "/^translate/"]
+}
```
ybiquitous, line 172: [suggest] Considering consistency with other secondary options in this doc, it seems better to omit the option name:

```suggestion
["calc", "/^translate/"]
```

lib/rules/unit-disallowed-list/README.md
```diff
@@ -157,3 +157,40 @@ The following patterns are considered problems:
...
+```
+
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+a {
+  margin: calc(50% - 100px)
+}
```
ybiquitous, line 181: [suggest] Like other code snippets in this doc, can you please change to one-line code?

```suggestion
a { margin: calc(50% - 100px); }
```

lib/rules/unit-disallowed-list/README.md
```diff
@@ -157,3 +157,40 @@ The following patterns are considered problems:
...
+```
+
+<!-- prettier-ignore -->
+```css
+a {
+  translateX: calc(50% - 100px)
+}
```
ybiquitous, line 188: [suggest] Should `translateX` be a function notation rather than a property?

```suggestion
a { transform: translateX(100px); }
```

lib/rules/unit-disallowed-list/README.md
```diff
@@ -157,3 +157,40 @@ The following patterns are considered problems:
...
+```
+
+<!-- prettier-ignore -->
+```css
+a {
+  translateY: calc(50% - 100px)
+}
+```
```
ybiquitous, line 196: [suggest] I think one example for `/^translate/` is enough:

```suggestion
```

lib/rules/unit-disallowed-list/__tests__/index.js
```diff
@@ -388,6 +388,67 @@ testRule({
...
+		},
+		{
+			code: 'a { translateX: calc(50% - 100px) }',
+		},
+		{
+			code: 'a { translateY: calc(50% - 100px) }',
+		},
+		{
+			code: 'a { translateZ: calc(50% - 100px) }',
+		},
```
ybiquitous, line 413: [suggest] A case for `/^translate/` seems necessary, and duplicate cases seems unnecessary:

```suggestion
		{
			code: 'a { transform: translateX(100px) }',
		},
```

ybiquitous (changes requested):
Thanks for creating the PR! Please consider addressing the reviews.

mattxwang:
Thank you for the quick review @ybiquitous! Upon reflection, I should have been more careful before submitting this PR - the slip-ups are 100% on me. Appreciate the feedback!

I think I've addressed each comment; let me know what you think.

lib/rules/unit-disallowed-list/README.md
```diff
@@ -157,3 +157,27 @@ The following patterns are considered problems:
...
+### `ignoreFunctions: ["function", "/regex/", /regex/]|"function"|"/regex/"|/regex/`
+
+Ignore units that are inside of the specified functions.
+
+For example, with `["px"]`.
+
+Given:
+
+```json
+"ignoreFunctions": ["calc", "/^translate/"]
```
ybiquitous, line 170: [suggest] For JSON validity and consistency, can you please update it?

```suggestion
["calc", "/^translate/"]
```

lib/rules/unit-disallowed-list/README.md
```diff
@@ -157,3 +157,27 @@ The following patterns are considered problems:
...
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+a { margin: calc(50% - 100px) }
+```
+
+<!-- prettier-ignore -->
+```css
+a { transform: translateX(50% - 100px) }
```
ybiquitous, line 182: [suggest] Calculation in [`translateX()`](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/translateX) seems invalid. Can you please fix it?

```suggestion
a { transform: translateX(100px) }
```

lib/rules/unit-disallowed-list/__tests__/index.js
```diff
@@ -388,6 +388,64 @@ testRule({
...
+
+	accept: [
+		{
+			code: 'a { margin: calc(50% - 100px) }',
+		},
+		{
+			code: 'a { transform: translate(50% - 100px, 100px) }',
+		},
+		{
+			code: 'a { transform: translateX(50% - 100px) }',
```
ybiquitous, line 409: [suggest] Same here.

```suggestion
			code: 'a { transform: translate(100px, 100px) }',
		},
		{
			code: 'a { transform: translateX(100px) }',
```

lib/rules/unit-disallowed-list/__tests__/index.js
```diff
@@ -388,6 +388,64 @@ testRule({
...
+			endColumn: 22,
+		},
+		{
+			code: 'a { translateZ: 100px }',
+			message: messages.rejected('px'),
+			line: 1,
+			column: 20,
+			endLine: 1,
+			endColumn: 22,
+		},
```
ybiquitous, line 445: [suggest] Since properties like `translateX` or `translateY`, it seems possible to reduce duplicate cases:

```suggestion
		{
			code: 'a { translate: 100px }',
			message: messages.rejected('px'),
			line: 1,
			column: 19,
			endLine: 1,
			endColumn: 21,
		},
```

ybiquitous (commented):
Thanks for addressing the reviews. I've left a few suggestions.

lib/rules/unit-disallowed-list/README.md
mattxwang: Thanks! I've applied the change; I noticed a similar pattern in `unit-allowed-list` which I based the documentation off of:

https://github.com/stylelint/stylelint/blob/b447409c2f7a619f663d7888e98050cba6a636d7/lib/rules/unit-allowed-list/README.md?plain=1#L125-L129

It seems inconsistent with the rest of the docs, so I'll also submit a (tiny) PR to resolve that. Great eye!

mattxwang: (opened a micro-PR in https://github.com/stylelint/stylelint/pull/6594; let me know if that was the desired behaviour)

ybiquitous (approved):
Thanks. LGTM! 👍🏼

PR #6605, opened 2023-01-31
Fix `keyframe-selector-notation` false positives for named timeline range

### LINKED ISSUE #6538: Fix `keyframe-selector-notation` false positives for named timeline ranges
### Demand

Add tests that ensure that
- `@keyframes foo { enter 0% {} enter 100% {} }`
- `@keyframes foo { exit 0% {} exit 100% {} }`
- `@keyframes foo { cover 0% {} cover 100% {} }`
- `@keyframes foo { contain 0% {} contain 100% {} }`

are supported/ignored.

ref: https://www.w3.org/TR/scroll-animations-1/#view-timelines-ranges

All three possible values (`"keyword"`,`"percentage"`, `"percentage-unless-within-keyword-only-block"`) must be covered.

kimulaco:

Closes #6538

Fix to stop linting if the selector of the keyframe has multiple tags.
Please let me know if there is a better approach.

### DIFF
--- a/lib/rules/keyframe-selector-notation/__tests__/index.js
+++ b/lib/rules/keyframe-selector-notation/__tests__/index.js
@@ -29,6 +29,10 @@ testRule({
 			code: '@keyframes foo { from, to {} }',
 			description: 'selector lists',
 		},
+		{
+			code: '@keyframes foo { enter 0% {} enter 100% {} }',
+			description: 'timeline-range-name',
+		},
 	],
 
 	reject: [
@@ -166,6 +170,10 @@ testRule({
 			code: '@keyframes foo { 0%, 100% {} }',
 			description: 'selector lists',
 		},
+		{
+			code: '@keyframes foo { enter 0% {} enter 100% {} }',
+			description: 'timeline-range-name',
+		},
 	],
 
 	reject: [
@@ -300,6 +308,10 @@ testRule({
 		},
 		{
 			code: '@keyframes foo { from,to {} }',
+		},
+		{
+			code: '@keyframes foo { enter 0% {} enter 100% {} }',
+			description: 'timeline-range-name',
 		},
 	],
 
--- a/lib/rules/keyframe-selector-notation/__tests__/index.js
+++ b/lib/rules/keyframe-selector-notation/__tests__/index.js
@@ -30,8 +30,24 @@ testRule({
 			description: 'selector lists',
 		},
 		{
+			code: '@keyframes foo { cover 0% {} cover 100% {} }',
+			description: 'timeline-range-name cover',
+		},
+		{
+			code: '@keyframes foo { contain 0% {} contain 100% {} }',
+			description: 'timeline-range-name contain',
+		},
+		{
+			code: '@keyframes foo { entry 0% {} entry 100% {} }',
+			description: 'timeline-range-name entry',
+		},
+		{
 			code: '@keyframes foo { enter 0% {} enter 100% {} }',
-			description: 'timeline-range-name',
+			description: 'timeline-range-name enter',
+		},
+		{
+			code: '@keyframes foo { exit 0% {} exit 100% {} }',
+			description: 'timeline-range-name exit',
 		},
 	],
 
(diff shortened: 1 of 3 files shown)


kimulaco:
I misunderstood the spec.
I will add a fix that ignores when using the `timeline-range-name` (`cover | contain | entry | exit `).
https://www.w3.org/TR/scroll-animations-1/#view-timelines-ranges

lib/rules/keyframe-selector-notation/__tests__/index.js
```diff
@@ -301,6 +341,26 @@ testRule({
 		{
 			code: '@keyframes foo { from,to {} }',
 		},
+		{
```
Mouvedia, line 344: suggestion

put these in a const and destructure it 3 times

lib/rules/keyframe-selector-notation/index.js
```diff
@@ -19,6 +19,7 @@ const meta = {
 
 const PERCENTAGE_SELECTORS = new Set(['0%', '100%']);
 const KEYWORD_SELECTORS = new Set(['from', 'to']);
+const NAMED_TIMELINE_RABGE_SELECTORS = new Set(['cover', 'contain', 'entry', 'enter', 'exit']);
```
Mouvedia, line 22: typo

RA**N**GE

lib/rules/keyframe-selector-notation/index.js
```diff
@@ -148,6 +153,23 @@ function getSelectorsInBlock(atRule) {
 	return selectors;
 }
 
+/**
+ * @param {import('postcss-selector-parser').Root} selectors
+ * @returns boolean
+ */
+function isNamedTimelineRange(selectors) {
+	for (const selectorNode of Array.from(selectors.nodes)) {
+		const firstTagNode = selectorNode.nodes.filter((node) => node.type === 'tag')[0];
```
Mouvedia, line 162: nit

```suggestion
const [firstTagNode] = selectorNode.nodes.filter((node) => node.type === 'tag');
const firstTagValue = (firstTagNode && firstTagNode.value) || '';
```

lib/rules/keyframe-selector-notation/index.js
```diff
@@ -73,6 +74,10 @@ const rule = (primary, _, context) => {
 
 			atRuleKeyframes.walkRules((keyframeRule) => {
 				transformSelector(result, keyframeRule, (selectors) => {
+					if (isNamedTimelineRange(selectors)) {
+						return;
+					}
+
```
ybiquitous, line 80: [suggest] How about the following refactoring to avoid extra selector node iteration within `isNamedTimelineRange()`? When a `walk` callback returns `false`, the iteration stops.

```js
transformSelector(result, keyframeRule, (selectors) => {
	let first = true;

	selectors.walkTags((selectorTag) => {
		if (first && NAMED_TIMELINE_RABGE_SELECTORS.has(selectorTag.value)) {
			return false; // ignore named timeline range
		}

		first = false;

		checkSelector(...);
	});
});
```

ybiquitous (commented):
@kimulaco Thanks for creating the pull request. I've left a refactoring suggestion comment, but there is no problem. 👍🏼

lib/rules/keyframe-selector-notation/index.js
kimulaco: Thanks, I fixed. https://github.com/stylelint/stylelint/pull/6605/commits/80a8351c20c1c14d6dfebd7dfc2580296d2e567a

lib/rules/keyframe-selector-notation/__tests__/index.js
kimulaco: Thanks, I agree with your suggestion. https://github.com/stylelint/stylelint/pull/6605/commits/ca57baf4ed0404f8202514eff9e19e5bd7cca5ae

lib/rules/keyframe-selector-notation/index.js
kimulaco: Thanks. Your suggestion is simpler and better, so I fixed it. https://github.com/stylelint/stylelint/pull/6605/commits/80a8351c20c1c14d6dfebd7dfc2580296d2e567a

lib/rules/keyframe-selector-notation/index.js
kimulaco: Thanks for reviewing. `isNamedTimelineRange()` is removed by the following suggestion:
https://github.com/stylelint/stylelint/pull/6605#discussion_r1095342096

ybiquitous (approved):
Thank you. LGTM 👍🏼

jeddy3 (approved):
LGTM, thank you.

lib/rules/keyframe-selector-notation/__tests__/index.js
```diff
@@ -2,6 +2,29 @@
...
+	{
+		code: '@keyframes foo { cover 0% {} cover 100% {} }',
+		description: 'timeline-range-name cover',
+	},
+	{
+		code: '@keyframes foo { contain 0% {} contain 100% {} }',
+		description: 'timeline-range-name contain',
+	},
+	{
+		code: '@keyframes foo { entry 0% {} entry 100% {} }',
```
Mouvedia, line 15: https://www.w3.org/TR/scroll-animations-1/#issue-c602d121

PR #6608, opened 2023-02-03
Fix `selector-not-notation` autofix for "simple" option

### LINKED ISSUE #6606: Fix `selector-not-notation` autofix for `"simple"`

```css
.foo:not(.bar, .baz) .content {
    display: none;
}
```

```json
{
  "rules": {
    "selector-not-notation": "simple"
  }
}
```

CLI `stylelint **/*.{vue,scss}`

14.16.1

Autofix to

```css
.foo:not(.bar):not(.baz) .content {
    display: none;
}
```

### What actually happened?

Autofix to

```css
.foo:not(.bar) .content.content {
    display: none;
}
```

No, only native CSS syntax is involved.

### Related issues

* https://github.com/stylelint/stylelint/issues/6236
* https://github.com/stylelint/stylelint/issues/3853

Mouvedia:

Closes #6606


lib/rules/selector-not-notation/index.js
```diff
@@ -152,7 +165,9 @@ function fixSimple(not) {
 	not.nodes.push(firstSelector);
 
 	for (const s of simpleSelectors) {
-		const last = not.parent.last;
+		const last = getLastConsecutiveNot(not.parent.nodes);
+
+		assert(last);
```
ybiquitous, line 170: [suggest] This change will be no longer needed if the suggestion above is applied:

```suggestion
```

lib/rules/selector-not-notation/index.js
```diff
@@ -130,6 +130,19 @@ const rule = (primary, _, context) => {
...
+ * @param {Node[]} nodes
+ */
+const getLastConsecutiveNot = (nodes) => {
+	let i = 0;
+
+	for (; i < nodes.length; i++) {
+		if (!isNot(/** @type {Node} */ (nodes[i]))) break;
+	}
+
+	return nodes[i - 1];
```
ybiquitous, line 143: [suggest] What about making the code readable by `findIndex()` like this? IMHO, an old-style loop using an index is not so readable.

In addition, what about returning `Node` instead of `Node | undefined` by moving `assert()`?

```suggestion
 * @param {Node[]} nodes
 * @returns {Node}
 */
const getLastConsecutiveNot = (nodes) => {
	const index = nodes.findIndex((node) => !isNot(node));
	const found = nodes[index - 1] || nodes[0];

	assert(found);

	return found;
```

ybiquitous (commented):
@Mouvedia Thanks for creating the pull request.
I've left a refactoring suggestion, but the direction looks good to me. 👍🏼

lib/rules/selector-not-notation/index.js
Mouvedia: This is obviously more readable and elegant.
Thanks!

Mouvedia: This is obviously more readable and elegant, thanks!
I see a problem with `const found = nodes[index - 1] || nodes[0];` though.
Ill add a test if it's not already covered.

lib/rules/selector-not-notation/__tests__/index.js
```diff
@@ -68,6 +68,24 @@ testRule({
...
+		{
+			code: ':not(.bar, .baz) .qux :not(.foo) {}',
+			fixed: ':not(.bar):not(.baz) .qux :not(.foo) {}',
+			message: messages.expected('simple'),
+			line: 1,
+			column: 1,
+			endLine: 1,
+			endColumn: 17,
+		},
 		{
```
Mouvedia, line 89: I gotta add `:not(.foo) .qux :not(.bar, .baz)` too.

ybiquitous (approved):
Thanks, LGTM 👍🏼

jeddy3 (approved):
LGTM, thank you!

PR #5303, opened 2021-05-14
Drop Node 10 support for ESM

ybiquitous:
This change is a part of the next major version (v14).

- Remove Node 10 from CI
- Update `engines.node`

In addition, this removes needless `CI: true` (just a refactoring).
See <https://github.blog/changelog/2020-04-15-github-actions-sets-the-ci-environment-variable-to-true/>

> Which issue, if any, is this issue related to?

Closes #5292

> Is there anything in the PR that needs further explanation?

See also:
- https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

PR #6611, opened 2023-02-03
downgrade lint-staged

Mouvedia:
> context

see: https://github.com/stylelint/stylelint/commit/2290f557cb028242d978b06465dc108aa04a0c3a#r99179433

> reason/motivation

I had an error while commiting (because of `.?`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,7 +177,7 @@
     "jest": "^28.1.3",
     "jest-preset-stylelint": "^5.0.4",
     "jest-watch-typeahead": "^2.2.0",
-    "lint-staged": "^13.0.4",
+    "lint-staged": "^12.5.0",
     "node-fetch": "^3.3.0",
     "np": "^7.6.2",
     "npm-run-all": "^4.1.5",
--- a/.changeset/pink-geckos-grow.md
+++ b/.changeset/pink-geckos-grow.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: downgrade lint-staged to ^12.5.0
--- a/package-lock.json
+++ b/package-lock.json
@@ -78,7 +78,7 @@
         "jest": "^28.1.3",
         "jest-preset-stylelint": "^5.0.4",
         "jest-watch-typeahead": "^2.2.0",
-        "lint-staged": "^13.0.4",
+        "lint-staged": "^12.5.0",
         "node-fetch": "^3.3.0",
         "np": "^7.6.2",
         "npm-run-all": "^4.1.5",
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -192,6 +192,6 @@
     "typescript": "^4.9.3"
   },
   "engines": {
-    "node": "^12.20.0 || ^14.13.1 || >=16.0.0"
+    "node": "^12.13.0 || ^14.15.0 || >=16.13.0"
```
ybiquitous, line 195: [ask] It is likely to accept `lint-staged` downgrade because it's for dev, but `engines.node` downgrade may affect users. Is the latter necessary?

Mouvedia:
> …but `engines.node` downgrade may affect users. Is the latter necessary?

You are right to have that worry.
We are about to release a major release; I would like to have introduced that change before that.
Right now you can't really say why you have chosen ^14.13.1 over ^14.0.0.
You have three choices either :
1. that specific version introduced a feature before the LTS that is necessary (but that only works for the bottom range i.e. ^12.20.0)
2. always pick the format major.0.0 format i.e. `"^12.0.0 || ^14.0.0 || >=16.0.0"`
3. or use LTS because it's a good convention (what I am proposing)

It cannot be 1. in our case, so we should pick something sensible that can be justified.
i.e. either `"^12.13.0 || ^14.15.0 || >=16.13.0"` or `"^12.0.0 || ^14.0.0 || >=16.0.0"`.

This is not urgent but it's a nice-to-have, that would permit to have [the documentation](https://github.com/stylelint/stylelint/blob/v15/docs/migration-guide/to-15.md#removed-support-for-nodejs-12) expose the reason behind our choices.

ybiquitous:
> Right now you can't really say why you have chosen ^14.13.1 over ^14.0.0.

The reason to chosen `^14.13.1` is to consider ESM support at that time. See https://github.com/stylelint/stylelint/pull/5303#discussion_r632438728

> use LTS because it's a good convention (what I am proposing)

What are the benefits to adopt LTS? Adopting "major.0.0 format" seems better to me because it can support people or environments more widely, unless there is a special reason like ESM support.

---

In addition, we've dropped Node.js 12 support on the [`v15`](https://github.com/stylelint/stylelint/tree/v15) branch, so there seems to be no benefit to downgrade `lint-staged` on the `main` branch. See:

- https://github.com/stylelint/stylelint/blob/673dc468bb1e8e491ba6cf0cc9ec0acc31f44d2b/package.json#L198
- https://github.com/stylelint/stylelint/blob/673dc468bb1e8e491ba6cf0cc9ec0acc31f44d2b/docs/migration-guide/to-15.md?plain=1#L86-L88

Mouvedia:
> See https://github.com/stylelint/stylelint/pull/5303#discussion_r632438728

I didn't know, my bad. Now it makes perfect sense; Ill remove the commit.

> What are the benefits to adopt LTS?

You can read about LTS [here](https://github.com/nodejs/release).

> In addition, we've dropped Node.js 12 support on the [v15](https://github.com/stylelint/stylelint/tree/v15?rgh-link-date=2023-02-05T23%3A54%3A50Z) branch…

This was meant for the last release before v15.

ybiquitous:
I think the v15.0.0 release will be soon, so can we avoid this `lint-staged` downgrade? Perhaps, many conflicts may come in `package-lock.json`... 😓

Mouvedia:
I don't know. I was bothered so I PRd.
Is it urgent or important? It's not.

jeddy3:
> I think the v15.0.0 release will be soon, so can we avoid this lint-staged downgrade?

Yes, let's close.

Mouvedia:
eslint-plugin-jest seems to be affected too:

> `const globalAliases = ((_context$settings$jes = context.settings.jest) === null || _context$settings$jes === void 0 ? void 0 : _context$settings$jes.globalAliases) ?? {};`

gives a `SyntaxError`.

ybiquitous:
@Mouvedia Which version of Node.js do you use?

Mouvedia:
> @Mouvedia Which version of Node.js do you use?

It depends on the tab or if it's a terminal in my IDE.
I use nvm so I switch often. On the machine: 6, 8, 10, 12, 14 ,16.

Issue #4125, opened 2019-06-19
Add `include` and `exclude` options to stylelintrc config


DesignByOnyx:
> What is the problem you're trying to solve?

This was first brought up in #1227 

It would be nice if `.stylelintrc` file could support an "include" (and "exclude" or "ignore") option directly - which is something done by other tools like tsc/tslint and jest. Right now, you have to pass a _single_ glob pattern to the CLI and use a separate `.stylelintignore` file, which is slightly tedious (not complaining... just sayin). Furthermore, this feature could enable the following:

- multiple glob patterns
- dynamic includes (only available with `.stylelintrc.js` files)
- includes and excludes defined right next to each other
- eliminate need for `.stylelintignore` file
- multiple co-located config files (useful for monorepos or other large projects):
    ```
    "scripts": {
        "lint:css": "stylelint",
        "lint:less": "stylelint -c .stylelintrc.less.js",
        "lint:sass": "stylelint -c .stylelintrc.sass.js",
        "lint:package-foo": "stylelint -c .stylelintrc.package-foo.js"
    }
    ```

> What solution would you like to see?

Ability to set includes (and excludes) in the `.stylelintrc` file

hudochenkov:
I think it's a good idea.

evilebottnawi:
Interesting, how it is solve eslint

evilebottnawi:
Only overrides allow to setup in config, maybe we should support this also

jeddy3:
Now that we have an [`overrides` config property](https://stylelint.io/user-guide/configure#overrides), can we close this in favour of https://github.com/stylelint/stylelint/issues/3860 (and using negation in the files array)?

Mouvedia:
The 2 issues should be consolidated.

ybiquitous:
> can we close this in favour of #3860 (and using negation in the files array)?

I agree with closing it. We should continue to discuss this topic on #3860 or #6689.

Mouvedia:
> * multiple co-located config files (useful for monorepos or other large projects):
>   ```
>   "scripts": {
>       "lint:css": "stylelint",
>       "lint:less": "stylelint -c .stylelintrc.less.js",
>       "lint:sass": "stylelint -c .stylelintrc.sass.js",
>       "lint:package-foo": "stylelint -c .stylelintrc.package-foo.js"
>   }
>   ```

#6690

Issue #3860, opened 2018-12-19
Support specifying input files/globs in the config object


sindresorhus:
> What is the problem you're trying to solve?

In my projects, I prefer to always glob for all CSS files `**/*.css` and rather ignore the few CSS files I want instead of explicitly specify every file. When having to specify the files, it's easy to forget some files, especially when they're added later.

> What solution would you like to see?

I would like to see a top-level setting called `files` that accepts an array of strings. I could then set `files: ['**/*.css'],` in my shareable config and use the `ignoreFiles` in my projects to ignore project-specific files.

---

There are other benefits to specifying the files in a config object:

- The run script becomes less messy.
	Before: `{"lint": "xo && stylelint foo.js bar.js 'fixtures/*.css'"}`
	After: `{"lint": "xo && stylelint"}`
- People often forget to quote the globs, which can lead to weird inconsistencies, since if you don't quote them, they're interpreted as shell globs instead of being passed in as strings to stylelint and its glob handling. By specifying the files in the stylelint config, that's not a problem.
- Editor plugins can read the `files` config and know which files to lint. It cannot reliably parse the run script commands.

I have had `files` config in both the [AVA](http://ava.li) and [XO](https://github.com/xojs/xo) projects for years with great success.

---

This is a duplicate of #2580, but I chose to open a new issue as I feel I have a different use-case and many more arguments for why it makes sense.

evilebottnawi:
Sounds reasonable, strange what `eslint` doesn't support this (and other feature like #3858 and #3859) :confused:

jeddy3:
Thanks for listing out the benefits and noting a similar approach in AVA and XO.

This sounds like a good idea. It's related to https://github.com/stylelint/stylelint/issues/3411, which is also about improving the experience of using stylelint.

Mouvedia:
1. should `files` support negation? (dupe of existing [`ignorefiles`](https://stylelint.io/user-guide/configure/#ignorefiles))
    - if it's supported does the cli input get the priority?
    which list overwrite or merge onto the other?
    e.g. "!foo.css" from the cli "foo.css" in files
    should `ignorefiles` be deprecated?
3. what's the merger strategy for ignored files/directories?
    - should [`.stylelintignore`](https://stylelint.io/user-guide/ignore-code) + `ignorefiles` be an union or a concat?
    - take `--ignore-path` into account before/after the merger
[...]

PR #6617, opened 2023-02-06
[DRAFT] placeholder title

### LINKED ISSUE #3411: Add extensions option for better DX experience

Linting all supported syntax (include non css - `html`, `markdown` and etc).
Now i use:

```shell
stylelint '{**/*,*}.{css,sass,scss,md,html}'
```

But it is very not convenient:
1. Long line in `package.json` with all supported extensions.
2. When we add new syntax and it is require add new extension to command
3. Some syntaxes support multiple extensions, example `markdown` support md, markdown, mdown, mdwn, mkd, mkdn, mkdown, ronn, workbook extensions. HTML and CSS-in-JS also have many extensions (future release). CSS

It can be solved using:
1. Contain supported `extensions` inside package.
2. Use `stylelint .` (if extensions doesn't exists in glob patter(s) we filter all files by `extensions` array)
3. Add flag `--ext` (or `--extensions`) to CLI to allow specify extensions (for example `eslint` have this already).
4. Allow set `extensions` array using `stylelint.config.js`

`prettier` will do same in `2` version.

/cc @stylelint/contributors

Mouvedia:

Closes #3411

TBD


docs/user-guide/configure.md
```diff
@@ -221,6 +221,14 @@ The value of `"extends"` is a "locater" (or an array of "locaters") that is ulti
 - an absolute path to a file (which makes sense if you're creating a JS object in a Node.js context and passing it in) with a `.js` or `.json` extension.
 - a relative path to a file with a `.js` or `.json` extension, relative to the referencing configuration (e.g. if configA has `extends: "../configB"`, we'll look for `configB` relative to configA).
 
+## `extensions`
```
Mouvedia, line 224: @jeddy3 can you take care of the documentation?
I wouldn't know what needs to be updated.

lib/augmentConfig.js
```diff
@@ -309,16 +309,21 @@ function mergeConfigs(a, b) {
 		rulesMerger.rules = { ...a.rules, ...b.rules };
 	}
 
-	const result = {
+	const extensionsMerger = {};
+
+	if (a.extensions || b.extensions) {
+		extensionsMerger.rules = { ...a.extensions, ...b.extensions };
```
Mouvedia, line 315: typo
fix incoming

lib/__tests__/overrides.test.js
```diff
@@ -208,6 +208,7 @@ describe('single input file. all overrides are matching', () => {
 			files: [path.join(fixturesPath, 'style.scss')],
 			config: {
 				extends: ['./at-rule-no-unknown.json'],
+				customSyntax: 'postcss-scss',
```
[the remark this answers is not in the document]
hudochenkov: I think it is better to add it to https://github.com/stylelint/stylelint/blob/main/lib/__tests__/fixtures/config-overrides/scss.json instead. Because it makes more sense to set custom syntax for `*.scss` files.

lib/standalone.js
```diff
@@ -187,6 +190,22 @@ async function standalone({
 
 	let stylelintResults;
 
+	{
+		/* --ext start */
+		const directories = fileList
+			.filter((p) => fs.existsSync(p) && fs.lstatSync(p).isDirectory())
+			.map((p) => path.relative(globCWD, p));
+		const extList = extensions && extensions.length ? extensions : DEFAULT_EXTENSIONS;
```
Mouvedia, line 198: I need to assign this array before `getConfigForFile` is called so that I use the extensions _after_ [the merger](https://github.com/stylelint/stylelint/pull/6617/files#diff-231823848e62de25699c7b9a8e92b05cc45aaf489c7c1fa417286295d37fcdf6R281).
It will remain a draft until that's fixed.

Mouvedia:
Only 10 tests from 2 test suites fail.
i.e. `npm run jest -- ignore.test.js`

Mouvedia:
@ybiquitous you are the author of the 2 remaining failing tests.
They explicitly say "with no configBasedir" in their respective names but the fix would be to add a `configBasedir`.
What was the purpose of these tests? Can you fix them for me?

```sh
npm run jest -- /ignore.test.js
```

Mouvedia:
Once Iv tackled https://github.com/stylelint/stylelint/commit/265a564db51ce7258b92cac594c721f3cc7342a3 it will ready to be reviewed.

```sh
npm run jest -- /ignore.test.js
```

ybiquitous:
@Mouvedia Thanks for picking up the old issue.

I have a concern about the `extensions` config. Does the config conflict with the [`overrides[].files`](https://stylelint.io/user-guide/configure#overrides) config?

```json
{
  "extensions": ["css"],
  "overrides": [
    {
      "files": ["**/*.html"],
      "rules": {...}
    }
  ]
}
```

For example, [`stylelint-config-html` has many extensions](https://github.com/ota-meshi/stylelint-config-html/blob/4dc0db6f862eeb0e5d63780c626a37b4f01ec9da/html.js#L3-L18).

The Node.js API also has the [`files`](https://stylelint.io/user-guide/node-api#files) option, so I guess it's enough to support only the `--ext` flag with CLI.

What do you think?

Mouvedia:
> Does the config conflict with the [`overrides[].files`](https://stylelint.io/user-guide/configure#overrides) config?

In what way? Do you have an example of conflict?
Without going into much details, it's just a global `filter`.
i.e. it narrows down the files selected for the upcoming lint

What I can say is, if we drop the config option the implementation will be way simpler.
Right now I have to simulate the extend path to get back the hypothetical resulting `extensions` union of all the configs.
If we drop `extensions` then the PR will be much easier to review.
I just assumed both were requested.
Personally I don't care, I just wanted to implement it because it's an issue with a lot of 👍

> so I guess it's enough to support only the `--ext` flag with CLI

If we go that route you should split #6617 in two and redirect the users to the _needs-discussion_ one.
And maybe add a rational if you have one.

> The Node.js API also has the [files](https://stylelint.io/user-guide/node-api#files) option

I wanted to tackle #3860 next.

> What do you think?

I think the pair `ignoreFiles` and `files` is problematic.
It should either be [include/exclude](https://github.com/stylelint/stylelint/issues/4125) or just `files` with negation support.

ybiquitous:
> In what way? Do you have an example of conflict?
> Without going into much details, it's just a global `filter`.

Ah, I misunderstood. Indeed, it's just a filter and should work without problems.

> I think the pair `ignoreFiles`/`files` is problematic.

Agreed. These configs are so confusing. If we would add `extensions`, it might be worse.

> If we drop extensions then the PR will be much easier to review.

Make sense. I agree with such simplicity.

@jeddy3 What do you think?

lib/augmentConfig.js
```diff
@@ -275,16 +275,21 @@ function mergeConfigs(a, b) {
 		rulesMerger.rules = { ...a.rules, ...b.rules };
 	}
 
-	const result = {
+	const extensionsMerger = {};
+
+	if (a.extensions || b.extensions) {
+		extensionsMerger.extensions = [...new Set([...(a.extensions || []), ...(b.extensions || [])])];
```
ybiquitous, line 281: [trivial] Using `concat()` seems a bit readable. But just my preference. You can keep it.

```suggestion
		extensionsMerger.extensions = [...new Set((a.extensions || []).concat(b.extensions || []))];
```

lib/cli.js
```diff
@@ -432,6 +461,10 @@ module.exports = async (argv) => {
 		optionsBase.disableDefaultIgnores = cli.flags.disableDefaultIgnores;
 	}
 
+	if (cli.flags.ext) {
+		optionsBase.extensions = cli.flags.ext.replace(/\./g, '').split(',');
```
ybiquitous, line 465: [suggest] How about normalizing a comma-separated string like this?

- Trim whitespaces
- Filter empty strings

```suggestion
		optionsBase.extensions = cli.flags.ext.replace(/\./g, '').split(',').map(s => s.trim()).filter(Boolean);
```

lib/cli.js
```diff
@@ -29,6 +29,7 @@ const EXIT_CODE_ERROR = 2;
  * @property {string} [color]
  * @property {string} [customFormatter]
  * @property {boolean} [disableDefaultIgnores]
+ * @property {string} [ext=".css"]
```
ybiquitous, line 32: [suggest] We may be able to omit the default value here to avoid repetition.

```suggestion
 * @property {string} [ext]
```

lib/standalone.js
```diff
@@ -187,6 +192,36 @@ async function standalone({
...
+			: DEFAULT_EXTENSIONS;
+
+		filePaths = filePaths.filter((p) => {
+			if (directories.some((directory) => p.startsWith(directory)))
+				return extList.some((extension) => p.endsWith(`.${extension}`));
+
+			return true;
+		});
+		/* extensions end */
+	}
```
ybiquitous, line 223: [suggest] Can we extract the block into a new function in this module like this?

```js
filePaths = filterFilePathsByExtensions(...);

if (filePaths.length) {
  // ...
}
```

ybiquitous (commented):
@Mouvedia Thanks for the pull request. I've left review comments.

It may be nice to describe a comma-separated string for `--ext` in the document and the CLI help. What do you think?

lib/cli.js
Mouvedia: > Trim whitespaces

This is a flag argument so you shouldn't allow whitespace in between extensions.
i.e. `--ext .css,   .scss` shouldn't be allowed

> Filter empty strings

We maybe need a test for `--ext` followed by nothing.

lib/standalone.js
Mouvedia: I could but it will have too many arguments for my taste.
I am not sure if it's for the best.

lib/cli.js
ybiquitous: > This is a flag argument so you shouldn't allow whitespace in between extensions.

What about the case of `--ext=".css, .scss,"` using quotes?

Mouvedia: I think it shouldn't be allowed. Since _multiple_ is not supported, Id like to keep it simple.
But we maybe need a test for  dangling `,`.

- [ ] dangling `,`
- [ ] lone `--ext`

Mouvedia: I think it shouldn't be allowed. Since _multiple_ is not supported, Id like to keep it simple.
But we maybe need a test for a dangling `,`.

- [ ] dangling `,`
- [ ] lone `--ext`

Mouvedia: @ybiquitous Iv checked the case _lone --ext_ and AFAIK meow doesn't provide a way to know whether the flag is being passed or not when using a default—even using the `isRequired` callback.

I am gonna add a fix for _dangling `,`_ while I am at it Ill add your `map` even if I think it shouldn't be allowed; it doesn't hurt since we currently don't have error messages in place for the case `--ext=".css, .scss"`.

lib/__tests__/overrides.test.js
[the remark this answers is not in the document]
Mouvedia: fixed in #6645

Mouvedia:
> * the `extensions` config can be conflicted with [`ignoreFiles`](https://stylelint.io/user-guide/configure#ignorefiles), for example:
>   ```json
>   { "ignoreFiles": ["**/*.js"], "extensions": [".js"] }
>   ```
> * the `extensions` option of Node.js API can be conflicted [`files`](https://stylelint.io/user-guide/node-api#files), for example:
>   ```js
>   stylelint.lint({ files: ["**/*.css"], extensions: [".scss"] });
>   ```
> 
> A conflict can occur with CLI still, for example:
> 
> ```shell
> stylelint --ext='.css' '**/*.scss'
> ```

All of these scenari are unlikely. I mean who would ignore js files but only pick js files?

> We need to clarify the priority order. For example, in such a case, ESLint [ignores `--ext`](https://eslint.org/docs/latest/use/command-line-interface#--ext)…

That's what Iv done here:

https://github.com/Mouvedia/stylelint/blob/2b33910764330601f29a8f772e9e65aeb1927424/docs/user-guide/cli.md?plain=1#L67

ybiquitous:
> All of these scenari are unlikely. I mean who would ignore js files but only pick js files?

Because we cannot predicate how people pass arguments to Stylelint, we should check to avoid inconsistent behavior.

Now I'm concerned about the behavior when glob inputs and file extensions are passed simultaneously. Should we ignore the extensions like ESLint? Or should we merge globs and extensions filtering?

If we don't decide and document behavior, we might have to deal with many bug reports in the future.

In addition, having `extenstions` in the config might raise other problems. When such configs are overridden, how should we handle the `extensions`? Merge or overwrite? Considering the rest decisions, I think even it may be good to abort introducing `extensions` to the config at least on this PR.

> Like I said before, I fear that the subscribed users will be disappointed if we don't ship what they asked for.

Yeah, I also know it and see this feature nice. But I also think we shouldn't leave any concerns about the new feature.

Mouvedia:
> Now I'm concerned about the behavior when glob inputs and file extensions are passed simultaneously. Should we ignore the extensions like ESLint?

Right now, `--ext` only applies to directories as explained [here](https://github.com/stylelint/stylelint/pull/6617/files#diff-99c40913a909127db6e242dd1a4cf9e939236358cf5d7636ec781f9d06972156R106). If that's still unclear check [this test](https://github.com/stylelint/stylelint/pull/6617/files#diff-0378929f7797bb789a616f16b2a0f5d0e53a28eba60469829002b0b30d60886cR457).

> Or should we merge globs and extensions filtering?

If a glob expands to be a folder, the filter should be applied.
ref https://github.com/stylelint/stylelint/pull/6617#discussion_r1097467613

> If we don't decide and document behavior, we might have to deal with many bug reports in the future.

It's documented but only once in the `--help`. I should probably add a note about it in cli.md.

> When such configs are overridden, how should we handle the `extensions`? Merge or overwrite?

I chose merge and added [a test](https://github.com/stylelint/stylelint/pull/6617/files#diff-bfdbcf2f00abee1b29aefe6ac6f68f5fe86e68d0958ece6ab96bf54b7c98ac32R50) for it.

ybiquitous:
> > When such configs are overridden, how should we handle the `extensions`? Merge or overwrite?
> 
> I chose merge and added [a test](https://github.com/stylelint/stylelint/pull/6617/files#diff-bfdbcf2f00abee1b29aefe6ac6f68f5fe86e68d0958ece6ab96bf54b7c98ac32R50) for it.
> I hope that clears up your uncertainties.

Is merging the correct way? If some people request overwriting over merging, how should we handle the needs?

I don't think the `extensions` config is really needed for now. Also, I think it's enough to provide a way to specify file extensions from entry points (CLI and Node.js API).

How about keeping it simple? If we could give up adding the `extensions` config, we would not need the [extra config loading](https://github.com/stylelint/stylelint/pull/6617/files#diff-37813c24c3fa3c6093015b61419f0bc18983cc2bbaddd56a732576bdb53f9023R312-R320) and would be beneficial for performance.

I want to hear other opinions from @stylelint/contributors.

lib/cli.js
```diff
@@ -259,6 +268,8 @@ const meowOptions = {
 			type: 'boolean',
 		},
 		config: {
+			// TODO
+			// alias: 'c',
```
ybiquitous, line 272: [suggest] Let's open an issue instead of leaving a comment:

```suggestion
```

lib/__tests__/standalone-ext.test.js
```diff
@@ -0,0 +1,35 @@
...
+
+it('extensions default', async () => {
+	const { results } = await standalone({
+		files: [path.join(fixturesPath, 'extensions')],
+		config: { rules: { 'color-named': 'never' } },
+	});
+
+	expect(results).toHaveLength(1);
+	expect(results[0].source.endsWith(path.join('extensions', 'folder', 'qux.css'))).toBe(true);
+});
```
ybiquitous, line 35: [suggest] Could you add cases when specifying glob patterns?

romainmenke:
> I don't think the `extensions` config is really needed for now. Also, I think it's enough to provide a way to specify file extensions from entry points (CLI and Node.js API).

I agree, much easier to add this later if enough people need it, than it is to remove it.

Mouvedia:
> Is merging the correct way? If some people request overwriting over merging, how should we handle the needs?

That's a valid argument but it seemed to me that merging was consistent with the rest.

> I agree, much easier to add this later if enough people need it, than it is to remove it.
> How about keeping it simple? If we could give up adding the extensions config, we would not need the [extra config loading](https://github.com/stylelint/stylelint/pull/6617/files#diff-37813c24c3fa3c6093015b61419f0bc18983cc2bbaddd56a732576bdb53f9023R312-R320) and would be beneficial for performance.

Ill do a separate commit that removes `extensions` for the config.
Remember to squash it when merging.
@ybiquitous could you do what I was advising earlier?
i.e.

> If we go that route you should split #3411 in two and redirect the users to the needs-discussion one.
> And maybe add a rational if you have one.

I wanted to tackle this ticket because a lot of users wanted it done for a long time. I don't want to be the one telling them they won't have it. Ill leave that to someone more brave :)

> Also, I think it's enough to provide a way to specify file extensions from entry points (CLI and Node.js API).

Your argument remains in that case: what happens when you extend using the Node.js API?
i.e. [this test](https://github.com/stylelint/stylelint/pull/6617/files#diff-bfdbcf2f00abee1b29aefe6ac6f68f5fe86e68d0958ece6ab96bf54b7c98ac32R50)
I suppose you want to keep it simple and just ignore the `extensions` set on the other configs, correct?

lib/__tests__/standalone-ext.test.js
Mouvedia: https://github.com/stylelint/stylelint/pull/6617/files#diff-0378929f7797bb789a616f16b2a0f5d0e53a28eba60469829002b0b30d60886cR456

Mouvedia: https://github.com/stylelint/stylelint/pull/6617/files#diff-0378929f7797bb789a616f16b2a0f5d0e53a28eba60469829002b0b30d60886cR457

Mouvedia: https://github.com/stylelint/stylelint/pull/6617/files#diff-0378929f7797bb789a616f16b2a0f5d0e53a28eba60469829002b0b30d60886cR457

Mouvedia: https://github.com/stylelint/stylelint/pull/6617/files#diff-0378929f7797bb789a616f16b2a0f5d0e53a28eba60469829002b0b30d60886cR457

ybiquitous:
> If we go that route you should split #3411 in two and redirect the users to the needs-discussion one.

Done. Please check #6689 out.

By the way, it seems we should consider the new properties suggested on #3411 and #3860 together. For example, see the config object below:

```json
{
  "files": ["src/**/*.{css,scss}"],
  "extensions": [".js", ".html"]
}
```

Note that these properties may be implicit by extended configs. So, they may make people confusing.

This is a reason why I think `extensions` is unnecessary.

Mouvedia:
> Done.

Thanks.

> "extensions": [".js", ".html"]

@ybiquitous check the previous commit it was

```
  "extensions": ["js", "html"]
```

> By the way, it seems we should consider the new properties suggested on https://github.com/stylelint/stylelint/issues/3411 and https://github.com/stylelint/stylelint/issues/3860 together.

[Like I said earlier](https://github.com/stylelint/stylelint/pull/6617#issuecomment-1426007965) I might tackle #3860 if no one does before me.

> This is a reason why I think `extensions` is unnecessary.

You don't have to justify _to me_ :)

docs/user-guide/options.md
```diff
@@ -139,6 +139,15 @@ CLI flags: `--disable-default-ignores, --di`
...
 Disable the default ignores. Stylelint will not automatically ignore the contents of `node_modules`.
 
+## `extensions`
+
+CLI flags: `--ext`
+
+Specify supported file extensions. By default, Stylelint sets it to `['css']`.
+
+Note that it only applies to directories.
+i.e. files and glob patterns matching files will be ignored
```
ybiquitous, line 149: [suggest] We can make it a bit simpler like this:

````suggestion
Note that it only applies to directories, i.e., files and glob patterns matching files will be ignored.
````

**Review by ybiquitous — CHANGES_REQUESTED** [review]

File: .changeset/old-otters-care.md [hunk 12]
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: `--ext` and `extensions` to the Node.js API
```
ybiquitous, line 5: [suggest] I've updated the PR title. Please follow it here:


```suggestion
Added: `--ext` CLI flag and `extensions` Node.js API option
```

Perhaps `extensions` may be renamed to `fileExtensions` though.

docs/user-guide/configure.md
```diff
@@ -218,7 +218,7 @@ The value of `"extends"` is a "locater" (or an array of "locaters") that is ulti
 - an absolute path to a file (which makes sense if you're creating a JS object in a Node.js context and passing it in) with a `.js` or `.json` extension.
 - a relative path to a file with a `.js` or `.json` extension, relative to the referencing configuration (e.g. if configA has `extends: "../configB"`, we'll look for `configB` relative to configA).
 
-You'll find more configs in [Awesome Stylelint](https://github.com/stylelint/awesome-stylelint#readme).
+You'll find more configs in [Awesome Stylelint](https://github.com/stylelint/awesome-stylelint#configs).
```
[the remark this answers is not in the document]
Mouvedia: I think it doesn't hurt and it leaves a clue that we are not adding `extensions` as a config property.
i.e. this is not an oversight, it's purposeful on our part

docs/user-guide/options.md
```diff
@@ -139,6 +139,15 @@ CLI flags: `--disable-default-ignores, --di`
 
 Disable the default ignores. Stylelint will not automatically ignore the contents of `node_modules`.
 
+## `extensions`
```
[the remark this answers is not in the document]
Mouvedia: > it seems we should move it to `node-api.md`

Good catch; Ill do that.

docs/user-guide/options.md
```diff
@@ -139,6 +139,15 @@ CLI flags: `--disable-default-ignores, --di`
...
+
+CLI flags: `--ext`
+
+Specify supported file extensions. By default, Stylelint sets it to `['css']`.
```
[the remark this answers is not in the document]
Mouvedia: -it

docs/user-guide/cli.md
```diff
@@ -60,6 +60,10 @@ Specify a custom syntax to use on your code. [More info](options.md#customsyntax
 
 Disable the default ignores. Stylelint will not automatically ignore the contents of `node_modules`. [More info](options.md#disabledefaultignores).
 
+### `--ext`
+
+Allows you to specify which file extensions Stylelint uses when searching for target files in the directories set in the [input](#usage-examples).
```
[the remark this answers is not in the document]
Mouvedia: > In addition, it seems nice to have a link to the option description on another page.

Which one?

Mouvedia:
I still gotta remove some unused files.

ybiquitous:
@Mouvedia Could you push commits without rebasing during a review? Otherwise, it’s hard to track changes after reviews.

Mouvedia:
@ybiquitous  In other PRs I will in this one it's a bit special: I want the commit which removed `extensions` cfg property to be separate so that we can easily go back to it if needed. Sorry for the annoyance.

docs/user-guide/configure.md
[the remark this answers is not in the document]
ybiquitous: `#configs` are hard to follow than `#readme` when `#configs` may change. `#readme` is more stable. Also, this change seems unrelated to this pull request.

docs/user-guide/cli.md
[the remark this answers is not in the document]
ybiquitous: It seems nice to have a link such as `[More info](node-api.md#extensions)`.

lib/augmentConfig.js
```diff
@@ -280,16 +280,21 @@ function mergeConfigs(a, b) {
 		rulesMerger.rules = { ...a.rules, ...b.rules };
 	}
 
-	const result = {
+	const extensionsMerger = {};
```
Mouvedia, line 283: still need to remove that

lib/cli.js
ybiquitous: Could you catch this up? Or, you can just remove the change.

types/stylelint/index.d.ts
```diff
@@ -46,6 +46,7 @@ declare module 'stylelint' {
 	 */
 	export type Config = {
 		extends?: ConfigExtends;
+		extensions?: string[];
```
ybiquitous, line 49: [suggest] `Config` has no longer `extensions`:

```suggestion
```

ybiquitous (changes requested):

lib/cli.js
```diff
@@ -185,6 +187,13 @@ const meowOptions = {
         during git operations like "git clone" because git does not track file modification
         time.
 
+      --ext                         [default: ".css"]
+
+        Specify supported file extensions.
+
+        Only applies to directories.
+        i.e. files and glob patterns matching files will be ignored
```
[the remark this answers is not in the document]
Mouvedia: alright

lib/cli.js
Mouvedia: Ill remove it but won't _resolve_ until someone creates an issue.

lib/__tests__/standalone-ext.test.js
Mouvedia: I think we need tests in both `cli.test.js` and `standalone.test.js`.
e.g.
- if meow ever let's us detect the case `--ext` no arg
- when meow will support list+multiple

lib/cli.js
```diff
@@ -432,6 +461,14 @@ module.exports = async (argv) => {
 		optionsBase.disableDefaultIgnores = cli.flags.disableDefaultIgnores;
 	}
 
+	if (cli.flags.ext) {
+		optionsBase.extensions = cli.flags.ext
+			.replace(/\./g, '')
```
ybiquitous, line 466: [suggest] I think moving this `replace()` into `standalone.js` is necessary because Node.js API also allows `['.css', '.scss']`, not only `['css', 'scss']`.

lib/standalone.js
```diff
@@ -41,6 +44,7 @@ async function standalone({
 	customSyntax,
 	cwd = process.cwd(),
 	disableDefaultIgnores,
+	extensions,
```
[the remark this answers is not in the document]
Mouvedia: Ill keep this line almost intact: `const extList = extensions?.length ? extensions : DEFAULT_EXTENSIONS;` since `extensions` can be provided but empty. Ill just remove the `?`.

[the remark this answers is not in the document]
ybiquitous: Then, how about changing the arguments of `filterByExtensions()` instead of passing a `stylelint` object like this?

```js
if (filePaths.length) {
	filePaths = filterByExtensions(fileList, filePaths, extensions, globCWD);
}

// ...

function filterByExtensions(inputPaths, resolvedFilePaths, extensions, cwd) {
  // ...
}
```

lib/cli.js
ybiquitous: If needed, please open an issue. At least, the change is unrelated to this pull request.

lib/__tests__/standalone-ext.test.js
ybiquitous: `cli.test.js` should focus on handling CLI arguments. We can test the logic details in `standalone-ext.test.js`.

lib/standalone.js
```diff
@@ -290,6 +300,27 @@ function getFormatterFunction(selected) {
...
+	const { cwd, extensions, globbyOptions } = stylelint._options;
+	const globCWD = globbyOptions?.cwd || cwd;
+	const directories = fileList
+		.filter((p) => fs.existsSync(p) && fs.lstatSync(p).isDirectory())
+		.map((p) => path.relative(globCWD, p));
+	const extList = extensions?.length ? extensions : DEFAULT_EXTENSIONS;
+
+	return filePaths.filter((p) => {
+		if (directories.some((directory) => p.startsWith(directory)))
+			return extList.some((extension) => p.endsWith(`.${extension}`));
```
ybiquitous, line 318: [suggest] We can rename local vars readable and short enough like this:

```suggestion
	return filePaths.filter((file) => {
		if (directories.some((dir) => file.startsWith(dir))) {
			return extList.some((ext) => p.endsWith(`.${ext}`));
		}
```

lib/cli.js
ybiquitous: For example, we can use `replace()` in `filterByExtensions()` like this:

```js
const extList = extensions.map((ext) => ext.replace(/\./g, ''));
```

Mouvedia: > because Node.js API can also allow `['.css', '.scss']`

That's not allowed.

If you are saying it should also be allowed. I think it should be discussed in #6689.
It makes sense to have the `.` with `--ext` but not really with an array.
In this case I favor discoverability/conventions over consistency.
Let's keep it simple for now and add it later?

<hr />

BTW I think you could be interested in [this RFC](https://github.com/eslint/rfcs/pull/9/files#diff-3553ca3bc6ff6b51efa315c689191a841deb1f6fdb83f9b318ffb2c3a6adf6a8R550).
It kinds of support your argument of a possible conflict between `files` and config `extensions` but it also favors the config over the cli—which is the opposite of what we are currently doing.

lib/standalone.js
Mouvedia: You made a typo but I agree with you.

lib/cli.js
ybiquitous: > Let's keep it simple for now and add it later?

Make sense. Let's keep it as-is. 👍🏼 

> It kinds of support your argument of a possible conflict between files and config extensions but it also favors the config over the cli—which is the opposite of what we are currently doing.

Yeah, we can learn many lessons from ESLint. This [ESLint blog](https://eslint.org/blog/2022/08/new-config-system-part-1/#the-need-for-simplification) emphasizes also **simplicity**. The more we fulfill people's requests, the harder we maintain the codebase.

Although I believe the `--ext` support is convenient for now, we may face difficulties if we extend the config system in the future.

Could you share the links on #3411 and #3860?

lib/__tests__/standalone-ext.test.js
Mouvedia: So to be clear you want to move all existing tests for `--ext` currently in `cli.test.js` in `standalone-ext.test.js`?

lib/cli.js
Mouvedia: #6690

lib/standalone.js
```diff
@@ -290,6 +300,27 @@ function getFormatterFunction(selected) {
...
+	const { cwd, extensions, globbyOptions } = stylelint._options;
+	const globCWD = globbyOptions?.cwd || cwd;
+	const directories = fileList
+		.filter((p) => fs.existsSync(p) && fs.lstatSync(p).isDirectory())
+		.map((p) => path.relative(globCWD, p));
+	const extList = extensions?.length ? extensions : DEFAULT_EXTENSIONS;
+
+	return filePaths.filter((file) => {
+		if (directories.some((dir) => file.startsWith(dir)))
+			return extList.some((ext) => file.endsWith(`.${ext}`));
```
ybiquitous, line 318: [suggest]

- `fileList` and `filePaths` are confusing to me. How about renaming them? The following is an example.
- How about explicitly passing `extensions` and `cwd` instead of a `stylelint` object?

```suggestion
function filterByExtensions(inputPathList, resolvedFilePaths, extensions, cwd) {
	const directories = inputPathList
		.filter((p) => fs.existsSync(p) && fs.lstatSync(p).isDirectory())
		.map((p) => path.relative(cwd, p));

	return resolvedFilePaths.filter((file) => {
		if (directories.some((dir) => file.startsWith(dir)))
			return extensions.some((ext) => file.endsWith(`.${ext}`));
```

Caller:

```js
filePaths = filterByExtensions(fileList, filePaths, extensions, globCWD);
```

lib/__tests__/standalone-ext.test.js
ybiquitous: If there are cases specific to CLI arguments, keeping them in `cli.test.js` is possible.

But I think we should have cases about extensions filtering in `standalone-ext.test.js` instead of `cli.test.js`.

lib/standalone.js
Mouvedia: > How about explicitly passing `extensions` and `cwd` instead of a `stylelint` object?

I disagree with this, that's why we have the `stylelint` object for. Check the first commit to see why it doesn't scale to pass everything as a separate argument.

> `fileList` and `filePaths` are confusing to me.

I agree.

> How about renaming them? The following is an example.

Ill take it into account.

ybiquitous: > I disagree with this, that's why we have the `stylelint` object for. Check the first commit to see why it doesn't scale to pass everything as a separate argument.

You may be right outside this module, but I think there is an advantage for a private function inside this module: we can avoid repetitions about `globbyOptions?.cwd` and `DEFAULT_EXTENSIONS`.

lib/__tests__/standalone-ext.test.js
Mouvedia: I found a bug while migrating the tests. Ill convert the PR to a draft while trying to fix it.

Mouvedia:
@ybiquitous thanks for the review. I need to refactor the implementation.
Ill work on it this weekend.

lib/standalone.js
Mouvedia: This part will be refactored so it will be irrelevant. It doesn't work as advertised.
Ill work on it this weekend and fix it.

Mouvedia: > we can avoid repetitions about `globbyOptions?.cwd` and `DEFAULT_EXTENSIONS`

We cannot avoid reusing `DEFAULT_EXTENSIONS` to cover the `extensions` === `[]` edge case.
Concerning `globbyOptions?.cwd`, I need both `globbyOptions` and `cwd`—in the latest commit.
We would end up with 5 arguments. Just passing `stylelint` is cleaner IMHO.

I think it's more readable with fewer arguments and I try to avoid having too many arguments passed to a function.

lib/__tests__/standalone-ext.test.js
Mouvedia: I have just fixed [an edge case](https://github.com/stylelint/stylelint/pull/6617/files#diff-7d9373b505908cd16088030eae195fbcb90036ed455e04eb15b32fffb4feab4aR63) while migrating the tests.
Just to be clear, Iv followed ESLint footsteps: globs are ignored.
Do you have any other cases to cover?

Mouvedia:
TODO

with an input: `foo` `foobar/baz/c.json` and `--ext .css`
`foobar/baz/c.json` should be picked up

Mouvedia:
> Can the leading dot be made optional, so that `css,scss` works the same as `.css,.scss`? This is matching behaviour of the same eslint option.

It's been requested once already by @ybiquitous. I guess we could.
Id prefer if it was an easter egg; if we do add it we would need new tests and documentation updates.

lib/cli.js
Mouvedia: Now that I have added support for `--ext css,scss`, your request is now more justified.
I consider the PR ready as is but if you want I could add support for `['.css', '.scss']` for consistency sake.

ybiquitous: > I consider the PR ready as is but if you want I could add support for `['.css', '.scss']` for consistency sake.

Thanks. I think it's optional, but if we can realize this with a few codes, it's desirable.

lib/__tests__/standalone-ext.test.js
```diff
@@ -0,0 +1,125 @@
...
+			files: inputFiles,
+			customSyntax: 'postcss-scss',
+			extensions: ['css', 'scss'],
+			config: {
+				rules: { 'color-named': 'never' },
+			},
+		});
+
+		expect(results).toHaveLength(3);
+		expect(existsSync(path.join(fixturesPath, 'extensions', 'foo.json'))).toBe(true);
```
ybiquitous, line 27: [ask] What does this line `expect(existsSync(...))` expect? Since `foo.json` exists obviously, this expectation seems unneeded.

lib/__tests__/standalone-ext.test.js
```diff
@@ -0,0 +1,125 @@
...
+		expect(results[0].source.endsWith('no-empty-source.html')).toBe(true);
+		expect(results[1].source.endsWith(path.join('extensions', 'bar.scss'))).toBe(true);
+		expect(results[2].source.endsWith(path.join('extensions', 'folder', 'qux.css'))).toBe(true);
```
ybiquitous, line 30: [suggest] I suggest the following refactoring to make the test code more readable.

First, define a tiny util:

```js
const fixture = (...items) => path.join(__dirname, 'fixtures', ...items);
```

Next, shorten codes with the util like this:

```suggestion
		expect(results[0].source).toBe(fixture('no-empty-source.html'));
		expect(results[1].source).toBe(fixture('extensions', 'bar.scss'));
		expect(results[2].source).toBe(fixture('extensions', 'folder', 'qux.css'));
```

By doing so, we can avoid repetition including `.endsWith` or `.toContain` and use `.toBe` for more accurate expectations.

lib/__tests__/standalone-ext.test.js
```diff
@@ -0,0 +1,125 @@
...
+			cwd,
+		});
+
+		expect(results).toHaveLength(5);
+		expect(results[0].source).toContain('bar.scss');
+		expect(results[1].source).toContain('foo.json');
+		expect(results[2].source).toContain(path.normalize('folder/baz.json'));
+		expect(results[3].source).toContain(path.normalize('folder/qux.css'));
+		expect(results[4].source).toContain(path.normalize('folder-slash/file.json'));
+	});
```
ybiquitous, line 124: [suggest] How about adding a case for no match like this?

```js
test('no match', () => {
	return expect(
		standalone({
			files: ['.'],
			config: configColorNamed,
			extensions: ['js'],
			cwd,
		}),
	).rejects.toThrow('All input files were ignored');
});
```

ybiquitous (commented):
Thanks for the hard work. The behavior seems no problems.
I've suggested a few comments for a bit of improvement and maintainability. Please consider addressing them.

lib/__tests__/standalone-ext.test.js
Mouvedia: If someone removes `foo.json` (because it's empty) this test must fail.
I could have added `expect(existsSync(path.join(fixturesPath, 'extensions', 'folder-slash', 'file.json'))).toBe(true);` but I considered one is enough.
In short, what we are checking is whether the filter works or not, for that it needs _not_ to pick up existing files.
If someone removes both json files the test would become useless.

lib/__tests__/standalone-ext.test.js
Mouvedia: Iv seen something similar in another test. Should it be added to `testUtils/` instead?

ybiquitous: > Should it be added to testUtils/ instead?

Sounds nice.

lib/__tests__/standalone-ext.test.js
Mouvedia: It unearthed a bug, thanks.

lib/__tests__/standalone-ext.test.js
Mouvedia: I don't think `replaceBackslashes` is needed.

Issue #5594, opened 2021-10-12
Fix `ignoreFiles` being evaluated case-sensitively on Windows


adalinesimonian:
Tested on Stylelint 14

`ignoreFiles` is evaluated case-sensitively on Windows, leading to a few issues downstream in vscode-stylelint. I stumbled into this while migrating from the deprecated vscode-languageserver URI API to the new vscode-uri API. The new API generates lowercase drive letters for paths instead of uppercase letters as the old API did.

Let's say the project is in `C:\Users\Adaline\Code\my-project`. Suppose that there is a file named `should-be-ignored.css` and the Stylelint config is:

```json
{ "ignoreFiles": "*-ignored.css" }
```

If you pass `C:\Users\Adaline\Code\my-project\should-be-ignored.css` to Stylelint, it ignores the file, as it should. But if you pass `c:\...\should-be-ignored.css` or any other case of any part of the path, Stylelint doesn't ignore it even though it can read the file correctly using the differently-cased path.

Node.js's API's behaviour, for reference, correctly sees these paths as equal on Windows:

```js
const path = require('path');

path.win32.relative('C:\\Path\\to\\file.css', 'c:\\path\\To\\File.css')
// Evaluates to ''
```

I haven't checked any other config properties or functionality, so I'm unaware if there are other cases of this behaviour elsewhere in the project.

psychobolt:
stylelint is using process.cwd() , which can potentially return lower-case in environments. 
 https://github.com/nodejs/node/issues/6624 
 
I think we should avoid using process.cwd() direct values for comparison.

Mouvedia:
What about using an util to normalize the paths?

e.g.
```js
return path
    .resolve(folder)
    .split(path.sep)
    .join("/")
    .replace(/^([a-z]):/, (_, driveLetter) => driveLetter.toLowerCase() + ":");
```

ybiquitous:
Can we use [`normalize-path`](https://www.npmjs.com/package/normalize-path)? We have already it as a dependency:

https://github.com/stylelint/stylelint/blob/5d907ddefa78f7c21b39640a06276b3eaa4afc45/package.json#L186

Mouvedia:
> Can we use [`normalize-path`](https://www.npmjs.com/package/normalize-path)? We have already it as a dependency:
> 
> https://github.com/stylelint/stylelint/blob/5d907ddefa78f7c21b39640a06276b3eaa4afc45/package.json#L186

[...]

PR #6627, opened 2023-02-07
Fix `configBasedir` being evaluated case-sensitively on Windows

nolimitdev:
Fixes not working "extends", "plugins" and "customSyntax" in case of small drive letter and backward slashes in configBasedir path. After this fix warn `When linting something other than CSS, you should install an appropriate syntax, e.g. "postcss-html", and use the "customSyntax" option.` is not shown and plugins works.

Such as https://github.com/stylelint/stylelint/issues/5594 but for configBasedir

### DIFF
--- a/.changeset/long-toys-boil.md
+++ b/.changeset/long-toys-boil.md
@@ -1,0 +1,5 @@
+---
+'stylelint': patch
+---
+
+Fixes not working "extends", "plugins" and "customSyntax" in case of small drive letter and backward slashes in configBasedir path.
--- a/.changeset/long-toys-boil.md
+++ b/.changeset/long-toys-boil.md
@@ -1,5 +1,0 @@
----
-'stylelint': patch
----
-
-Fixes not working "extends", "plugins" and "customSyntax" in case of small drive letter and backward slashes in configBasedir path.
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -85,14 +85,10 @@ async function standalone({
 		return Promise.reject(error);
 	}
 
-	// convert e.g. c:\foo\bar to C:/foo/bar to avoid not working "extends", "plugins" and "customSyntax"
+	// convert e.g. c:/foo/bar to C:/foo/bar to avoid not working "extends", "plugins" and "customSyntax"
 	if (configBasedir) {
 		if (configBasedir[0] && configBasedir[1] === ':') {
 			configBasedir = configBasedir[0].toUpperCase() + configBasedir.substr(1);
-		}
-
-		if (configBasedir.includes('\\')) {
-			configBasedir = configBasedir.replace(/\\/g, '/');
 		}
 	}
 
--- a/lib/standalone.js
+++ b/lib/standalone.js
@@ -85,6 +85,16 @@ async function standalone({
 		return Promise.reject(error);
 	}
 
+	// convert e.g. c:\foo\bar to C:/foo/bar to avoid not working "extends", "plugins" and "customSyntax"
+	if (configBasedir) {
+		if (configBasedir[1] == ':') {
+			configBasedir = configBasedir[0].toUpperCase() + configBasedir.substr(1);
+		}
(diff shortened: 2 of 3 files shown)


lib/standalone.js
```diff
@@ -85,6 +85,17 @@ async function standalone({
 		return Promise.reject(error);
 	}
 
+	// convert e.g. c:\foo\bar to C:/foo/bar to avoid not working "extends", "plugins" and "customSyntax"
+	if (configBasedir) {
+		if (configBasedir[0] && configBasedir[1] === ':') {
```
Mouvedia, line 90: opinion

1. organisation: the path handler/fixer should probably be in a helper/file in `lib/utils/`.
2. implementation: we should probably use builtin fs methods—maybe `fs.realpathSync.native`?—or an existing npm package.
3. consistency: all known/similar cases need to be fixed at the same time

.changeset/long-toys-boil.md
```diff
@@ -0,0 +1,5 @@
+---
+'stylelint': patch
+---
+
+Fixes not working "extends", "plugins" and "customSyntax" in case of small drive letter and backward slashes in configBasedir path.
```
ybiquitous, line 5: [suggest] Please remove the duplicate changelog.

lib/standalone.js
ybiquitous: [ask] Can we use [`normalize-path`](https://www.npmjs.com/package/normalize-path)? We've already used it.

ybiquitous (changes requested):
@nolimitdev Thanks for creating the pull request. Could you add test cases?

lib/standalone.js
nolimitdev: Yes, `normalize-path` could be used to fix part with backward slashes but when debugging thoroughly again problem is just lower-cased drive letter, so part with changing slashes I have just removed completely

nolimitdev:
@ybiquitous  I have just created demo repo with bug demo: https://github.com/nolimitdev/repo-stylelint-configbasedir-problem

ybiquitous:
@nolimitdev Thanks for providing the demo repo. The root cause is likely that a drive letter becomes lower-cased. I'm unsure which VSCode, VSCode Stylelint extension, or something is responsible.

But we can accept just converting a driver letter. If possible, please check if [`os.platform()`](https://nodejs.org/api/os.html#osplatform) equals `win32`, in order to avoid unexpected side-effects by the conversion.

Also, can you please add a test case for the change to `lib/__tests__/standalone.test.js` like this?

```js
let absoluteConfigBasedir = path.resolve("path/to/config");
absoluteConfigBasedir = absoluteConfigBasedir.charAt(0).toLowerCase() + absoluteConfigBasedir.substring(1);

const { results } = await standalone({
	/* ... */
	configBasedir: absoluteConfigBasedir,
});

// expectations...
```

CI jobs on Windows should pass the case above.

nolimitdev:
Demo repo is based on stylelint v14.16.0 where is a bug. By changelog in v14.16.1 https://github.com/stylelint/stylelint/blob/main/CHANGELOG.md#14161 there was fixing something related by description to this bug `Fixed: customSyntax resolution with configBasedir (https://github.com/stylelint/stylelint/pull/6536) (@ybiquitous])` but when tested demo repo with v14.16.1 `"customSyntax": "postcss-html"` still does not work. But when tested with v15.0.0 it works with any incorrect/nonexistent path for `stylelint.configBasedir` or `stylelint.configBasedir` can be omitted and `"customSyntax": "postcss-html"` works. But in v15 https://github.com/stylelint/stylelint/blob/main/CHANGELOG.md#1500 there is no mention about fixing something about it.

ybiquitous:
@nolimitdev I want to dig into the problem: Stylelint itself or VSCode extension. So, could you please open a new issue with a minimal reproduction repo if you find this PR is not relevant to the second problem?

jeddy3:
Closing as stale.

PR #6629, opened 2023-02-08
Add `configurationComment` configuration property

### LINKED ISSUE #6628: Add `disablesPrefix` configuration property

This might seem like an odd use case, but at larger companies there may be a company wide stylelint running against every team's codebase where the config file needs to meet the needs of the whole company. In order to enforce additional rules it can be convenient in that situation to run a separate instance of stylelint with additional rules. In this case it would be convenient to be able to disable each instance of stylelint with different stylelint commands by making the prefix configurable e.g. `stylelint-disable` -> `stylelint-special-rules-disable`.

I have a pull request with a solution that appears to work. I am happy to take any suggestions to make it more ergonomic or address any edge cases I might have missed.

ifitzpatrick:
Closes #6628

This PR changes the list of stylelint commands to omit the prefix, and updates functions related to stylelint commands to accept the prefix as an argument. Additionally it makes stylelintCommandPrefix configurable to the users needs.


Mouvedia:
[This page](https://stylelint.io/user-guide/configure) will have to be updated.

ifitzpatrick:
Updated the docs and changed the name of the configuration property to `disablesPrefix` (https://github.com/stylelint/stylelint/issues/6628#issuecomment-1425639557)

ybiquitous:
Considering we have also `stylelint-enable`, is the `disablesPrefix` name suitable? Is `commandPrefix` (or something else) better?

jeddy3:
@ybiquitous See https://github.com/stylelint/stylelint/issues/6628#issuecomment-1425639557 as `disables` is consistent with our other relevant options. Although, it's not a strong preference.

jeddy3:
> Is `commandPrefix` (or something else) better?

We may apply this to more than disables and enables in the future, e.g. [configuring rules](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments). So, on further thought, maybe there is a better name than `disablesPrefix`.

ESLint uses "configuration comments" to describe [disabling comments too](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1). Should we adopt that terminology for the option name and internally? It's aligned with our language of [configuring rules](https://stylelint.io/user-guide/configure/).

The option could be `configurationComment`, defaulting to `"stylelint"`. 

```json
{
  "configurationComment": "foo-stylelint"
}
```

Internally, we'd change:

- `isStylelintCommand(comment)` to `isConfigurationComment(comment)`
- `extractStylelintCommand(comment)` to  `extractStylelintComment(comment)`
- and so on

We, rather than the user, should add a hyphen before any of our existing command, e.g.:

- `const DISABLE_COMMAND = 'disable';` becomes `const DISABLE_COMMENT = '-disable';`

How does that sound?

ifitzpatrick:
Is there any concern that changing them names of `isStylelintCommand` and friends would be a breaking change for rules or plugins outside of this repo?

ybiquitous:
@jeddy3's idea sounds good to me. 👍🏼 

> Is there any concern that changing the names of `isStylelintCommand` and friends would be a breaking change for rules or plugins outside of this repo?

I believe there is no concern because `isStylelintCommand` is an internal utility, and we have not supported it as a public API.

Just in case, I [searched the usage of `isStylelintCommand` in GitHub](https://github.com/search?q=isStylelintCommand&type=code), but there seem to be no plugins using the utility.

ifitzpatrick:
Awesome thanks for looking into that @ybiquitous! Updated those functions and filenames to `configurationComment`.

docs/user-guide/configure.md
```diff
@@ -413,6 +413,18 @@ For example:
 
 [More info](options.md#ignoredisables).
 
+### `configurationComment`
```
ybiquitous, line 416: [ask] This new section is inserted under the `ignoreDisables` section, but is there any reason to do that?

docs/user-guide/configure.md
```diff
@@ -413,6 +413,18 @@ For example:
...
+
+You can set the `configurationComment` to change the prefix of stylelint commands used in `stylelint-disable` comments. This can be useful if you expect to have multiple instances of stylelint with different configurations running.
+
+For example, to have stylelint disable rules with `/* stylelint-special-rules-disable */` instead of the default `/* stylelint-disable */`:
+
+```json
+{
+  "configurationComment": "stylelint-special-rules-"
```
ybiquitous, line 424: [suggest] `configurationComment` is not for a prefix, so is the following correct?

```suggestion
  "configurationComment": "stylelint-special-rules"
```

lib/utils/configurationComment.js
```diff
@@ -0,0 +1,70 @@
...
+const ENABLE_COMMAND = '-enable';
+
+const ALL_COMMANDS = new Set([
+	DISABLE_COMMAND,
+	DISABLE_LINE_COMMAND,
+	DISABLE_NEXT_LINE_COMMAND,
+	ENABLE_COMMAND,
+]);
+
+const DEFAULT_COMMENT_PREFIX = 'stylelint';
```
ybiquitous, line 17: [suggest] For more accurate naming, how about the following?

```suggestion
const DEFAULT_COMMENT = 'stylelint';
```

or

```suggestion
const DEFAULT_CONFIGURATION_COMMENT = 'stylelint';
```

docs/user-guide/configure.md
ifitzpatrick: Felt related to disabling stuff but I wasn't sure the best place to put it so definitely open to moving it somewhere else that it fits better

docs/user-guide/configure.md
ifitzpatrick: Good catch, fixed!

lib/utils/configurationComment.js
ifitzpatrick: Updated to `DEFAULT_CONFIGURATION_COMMENT`

.changeset/wet-yaks-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Make stylelintCommandPrefix configurable
```
jeddy3, line 5: Added: `configurationComment` configuration property

docs/user-guide/configure.md
```diff
@@ -413,6 +413,18 @@ For example:
 
 [More info](options.md#ignoredisables).
 
+### `configurationComment`
+
+You can set the `configurationComment` to change the prefix of stylelint commands used in `stylelint-disable` comments. This can be useful if you expect to have multiple instances of stylelint with different configurations running.
```
jeddy3, line 418:
```suggestion
You can set what configuration comments like `/* stylelint-disable */` start with. This can be useful if you use multiple instances of Stylelint with different configurations.

```

docs/user-guide/configure.md
```diff
@@ -413,6 +413,18 @@ For example:
...
+
+For example, to have stylelint disable rules with `/* stylelint-special-rules-disable */` instead of the default `/* stylelint-disable */`:
```
jeddy3, line 420:
```suggestion
For example, to have an instance of Stylelint disable rules with `/* stylelint-foo-instance-disable */` instead of the default `/* stylelint-disable */`:
```

docs/user-guide/configure.md
```diff
@@ -413,6 +413,18 @@ For example:
...
+
+```json
+{
+  "configurationComment": "stylelint-special-rules"
```
jeddy3, line 424:
```suggestion
  "configurationComment": "stylelint-foo-instance"
```

lib/rules/comment-empty-line-before/index.js
```diff
@@ -61,7 +60,7 @@ const rule = (primary, secondaryOptions, context) => {
 
 			// Optionally ignore stylelint commands
 			if (
-				comment.text.startsWith(stylelintCommandPrefix) &&
+				comment.text.startsWith(context.configurationComment || DEFAULT_CONFIGURATION_COMMENT) &&
```
jeddy3, line 63: Can we use the `isConfigurationComment` util here instead?

docs/user-guide/configure.md
jeddy3: Let's move it to above `ignoreDisables` and make it a `##` heading.

jeddy3 (changes requested):
Thank you for making the changes. It's shaping up nicely.

I've requested a couple of documentation changes and asked one questino.

jeddy3:
I pushed 3 commits to standardise the terminology elsewhere in our docs as part of this PR.

lib/rules/comment-empty-line-before/README.md
```diff
@@ -140,7 +140,7 @@ a {
 
 #### `"stylelint-commands"`
 
-Ignore comments that deliver commands to stylelint, e.g. `/* stylelint-disable color-no-hex */`.
+Ignore configuration comments, e.g. `/* stylelint-disable color-no-hex */`.
```
jeddy3, line 143: I hadn't realised we had an option related to commands. I think we're ok to move forward with "configuration comment", though, as we have a few other rule secondary options which are at odds with our conventions.

lib/rules/comment-empty-line-before/index.js
ifitzpatrick: Fixed

docs/user-guide/configure.md
ifitzpatrick: Updated

types/stylelint/index.d.ts
```diff
@@ -192,6 +193,7 @@ declare namespace stylelint {
 	 * A rule context.
 	 */
 	type RuleContext = {
+		configurationComment?: string | undefined;
```
ybiquitous, line 196: [suggest] How about adding `configurationComment` to the developer guide?

https://github.com/stylelint/stylelint/blob/fcff497a3ea121a2721587f93f7b95d1eba3f6c9/docs/developer-guide/rules.md?plain=1#L198-L201

lib/rules/comment-empty-line-before/index.js
```diff
@@ -61,7 +60,7 @@ const rule = (primary, secondaryOptions, context) => {
 
 			// Optionally ignore stylelint commands
 			if (
-				comment.text.startsWith(stylelintCommandPrefix) &&
+				isConfigurationComment(comment) &&
```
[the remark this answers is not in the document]
ifitzpatrick: Fixed

types/stylelint/index.d.ts
ifitzpatrick: Updated

ybiquitous (approved):
Thank you. LGTM 👍🏼

jeddy3 (approved):
LGTM, thank you.

PR #6642, opened 2023-02-10
Document `reportDisables` in the customising guide

### LINKED ISSUE #6584: Document `reportDisables` in customizing guide

I wrote a `no-disable` rule that only works with `"ignoreDisables": true` flag as `root.walkComments` does not contain those declarations that have been disabled. Same issue for disabled files.

```
{
	"plugins":
	[
		"@isnotdefined/stylelint-plugin"
	],
	"rules":
	{
		"@isnotdefined/no-disable": true
	},
	"ignoreDisables": true
}
```

Source: https://github.com/isnotdefinedcom/stylelint-plugin/blob/master/src/rules/no-disable/rule.ts

Either there should be an internal API to set `ignoreDisables` once the rule is enabled or an option of a "pure walker" that includes every declaration, even the disabled ones.

jeddy3:

Closes #6584

None.

### DIFF
--- a/docs/user-guide/customize.md
+++ b/docs/user-guide/customize.md
@@ -106,6 +106,8 @@ You can use the [`report*`](./configure.md#report) properties in your config to 
 
 Each of these properties is configurable if you need to add exceptions to them.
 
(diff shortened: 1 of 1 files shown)


ybiquitous (approved):
Thanks. LGTM 👍🏼

Issue #6649, opened 2023-02-11
Fix `declaration-property-value-no-unknown` parseError even when a file starts with `/* stylelint-disable */`


romainmenke:
### What steps are needed to reproduce the bug?

```css
/* stylelint-disable */
a {
  filter: alpha(opacity=30);
}
```

CSS Tree is unable to parse old CSS hacks.
It also isn't designed to handle all these things.

### What Stylelint configuration is needed to reproduce the bug?

```json
{
  "declaration-property-value-no-unknown": true
}
```

### How did you run Stylelint?

CLI and VSCode

### Which version of Stylelint are you using?

15.1.0

### What did you expect to happen?

I expected to be able to ignore very old CSS with `/* stylelint-disable */` at the top a file.

### What actually happened?

An error was emitted.

### Proposal to fix the bug

https://github.com/stylelint/stylelint/blob/main/lib/rules/declaration-property-value-no-unknown/index.js#L93-L99

```js
try {
	cssTreeValueNode = parse(value, { context: 'value' });

	if (containsUnsupportedFunction(cssTreeValueNode)) return;
} catch (e) {
	result.warn(`Cannot parse property value "${value}"`, {
		node: decl,
		stylelintType: 'parseError',
	});

	return;
}
```

When I use `report()` I am able to use `disable` comments to suppress parse errors.

```js
try {
	cssTreeValueNode = parse(value, { context: 'value' });

	if (containsUnsupportedFunction(cssTreeValueNode)) return;
} catch (e) {
	const index = declarationValueIndex(decl);
	const endIndex = index + decl.value.length;

	report({
		message: messages.parseError(prop, decl.value),
		node: decl,
		index,
		endIndex,
		result,
		ruleName,
	});

	return;
}
```

jeddy3:
@romainmenke Thanks for the report.

Your proposal LGTM.

I've labelled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

There are [steps on how to fix a bug in a rule](https://stylelint.io/developer-guide/rules#fix-a-bug-in-a-rule) in the Developer guide.

PR #6650, opened 2023-02-12
Fix `declaration-property-value-no-unknown` parse error for `alpha(opacity=n)` to report as violation

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/6649

No, it's self-explanatory.

### DIFF
--- a/.changeset/clean-snakes-cheat.md
+++ b/.changeset/clean-snakes-cheat.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `declaration-property-value-no-unknown` parseError even when a file starts with `/* stylelint-disable */`
+Fixed: `declaration-property-value-no-unknown` parse error for `alpha(opacity=n)` to report as violation
--- a/.changeset/clean-snakes-cheat.md
+++ b/.changeset/clean-snakes-cheat.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `declaration-property-value-no-unknown` parseError even when a file starts with `/* stylelint-disable */`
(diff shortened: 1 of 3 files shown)


.changeset/clean-snakes-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `declaration-property-value-no-unknown` parseError even when a file starts with `/* stylelint-disable */`
```
ybiquitous, line 5: [suggest] It seems to me this is a problem of reporting rather than `/* stylelint-disable */`. What about this change (also the PR title)?

```suggestion
Fixed: `declaration-property-value-no-unknown` parse error for `alpha(opacity=n)` to report as violation
```

lib/rules/declaration-property-value-no-unknown/__tests__/index.js
```diff
@@ -49,6 +49,9 @@ testRule({
 		{
 			code: 'a { font-size: max(1rem, 3rem); }',
 		},
+		{
+			code: '/* stylelint-disable */ a { filter: alpha(opacity=30); }',
+		},
```
ybiquitous, line 54: [suggest] This case tests `/* stylelint-disable */` rather than this rule, so only the reject case seems enough to me. What about removing the case to avoid making developers confused?

```suggestion
```

ybiquitous (commented):
@romainmenke Thanks for the quick fix! Please confirm my reviews

lib/rules/declaration-property-value-no-unknown/__tests__/index.js
romainmenke: Thank you for pointing this out, I was also unsure about this one :)

.changeset/clean-snakes-cheat.md
romainmenke: Much better :)

ybiquitous (approved):
Thanks! LGTM 👍🏼

PR #6660, opened 2023-02-14
Fix `overrides.extends` order when including same rules

### LINKED ISSUE #6658: Fix `overrides.extends` order when including same rules

#6380 changes the behaviour of `overrides.extends` @jasikpark

In following config file, `./a` is the last item, it should in the last of the merge array.

But in the pull request

https://github.com/stylelint/stylelint/blob/4daef246942587560baa60177985b0ae6384738f/lib/augmentConfig.js#L303-L305

`./a` in first `extends` is used.

```
modules.export = {
  extends: ["./a", "./b"],
  overrides: [
    {
      files: ["*.module.css"],
      extends: ["./c", "./b", "./a"]
    }
  ]
}
```

none

`15.1.0`

`overrides.extends` work as expect

### What actually happened?

see previous

_No response_

### Proposal to fix the bug

_No response_

kuoruan:

Closes #6658

No, it's self-explanatory.

### DIFF
--- a/.changeset/thin-candles-cross.md
+++ b/.changeset/thin-candles-cross.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `overrides.extends` order when including same rules
--- a/lib/__tests__/applyOverrides.test.js
+++ b/lib/__tests__/applyOverrides.test.js
@@ -237,6 +237,30 @@ describe('two matching overrides', () => {
 
 		expect(applied).toEqual(expectedConfig);
 	});
+
+	test('with ordered extends', () => {
+		const config = {
+			extends: ['stylelint-config1'],
+			overrides: [
+				{
+					files: ['*.module.css'],
+					extends: ['stylelint-config2', 'stylelint-config1'],
+				},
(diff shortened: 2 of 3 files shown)


.changeset/thin-candles-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fix overrides.extends order when including same rules
```
ybiquitous, line 5: [suggest] According to [our convention](https://stylelint.io/maintainer-guide/pull-requests#merging):

```suggestion
Fixed: `overrides.extends` order when including same rules
```

ybiquitous (approved):
Thanks, LGTM 👍🏼 
In addition, please fix the changeset I've suggested.

.changeset/thin-candles-cross.md
kuoruan: @ybiquitous done.

ybiquitous (approved):
Thank you!

jeddy3 (approved):
LGTM, thank you.

PR #6661, opened 2023-02-14
Fix `plugins` and `overrides` duplicates when merged 

kuoruan:
We merge a.plugins and b.plugins, but only remove duplicates when b.plugins exist. Duplicate items can be in a.plugins too.

None.

No, it's self-explanatory.


lib/augmentConfig.js
```diff
@@ -250,8 +252,10 @@ function mergeConfigs(a, b) {
 		}
 
 		if (b.overrides) {
-			overridesMerger.overrides = [...new Set(overridesMerger.overrides.concat(b.overrides))];
+			overridesMerger.overrides = overridesMerger.overrides.concat(b.overrides);
```
Mouvedia, line 255: question

Why didn't you do it on l251 as well?

lib/augmentConfig.js
```diff
@@ -250,8 +252,10 @@ function mergeConfigs(a, b) {
...
 		}
+
+		overridesMerger.overrides = [...new Set(overridesMerger.overrides)];
```
Mouvedia, line 258: If you are doing it here, you don't need to it before

lib/augmentConfig.js
```diff
@@ -235,8 +235,10 @@ function mergeConfigs(a, b) {
 		}
 
 		if (b.plugins) {
-			pluginMerger.plugins = [...new Set(pluginMerger.plugins.concat(b.plugins))];
+			pluginMerger.plugins = pluginMerger.plugins.concat(b.plugins);
 		}
+
+		pluginMerger.plugins = [...new Set(pluginMerger.plugins)];
```
Mouvedia, line 241: ditto

ybiquitous: I also agree with @Mouvedia. We can pass `lib/__tests__/applyOverrides.test.js` even if using `lib/augmentConfig.js` before this change.

kuoruan:
@Mouvedia @ybiquitous Updated, please review again.

lib/augmentConfig.js
```diff
@@ -354,7 +361,8 @@ function applyOverrides(fullConfig, rootConfigDir, filePath) {
 	let { overrides, ...config } = fullConfig;
 
 	if (!overrides) {
-		return config;
+		// Do mergeConfigs to remove duplicate items
+		return mergeConfigs(config, {});
```
Mouvedia, line 365: opinion

that's overkill
i.e. if someone has only one config, which has duplicates, he ought to fix them himself

lib/augmentConfig.js
```diff
@@ -250,8 +252,13 @@ function mergeConfigs(a, b) {
 		}
 
 		if (b.overrides) {
-			overridesMerger.overrides = [...new Set(overridesMerger.overrides.concat(b.overrides))];
+			overridesMerger.overrides = overridesMerger.overrides.concat(b.overrides);
 		}
+
+		// Remove duplicates, last one wins
+		overridesMerger.overrides = overridesMerger.overrides.filter(
```
Mouvedia, line 259: question

Overrides are objects, where are you doing the deep comparison to detect duplicates?

kuoruan:
Seems no need for this PR, closing it.

PR #6666, opened 2023-02-16
Fix `*-case|*-unknown` false positives for CSS-in-JS template literals

### LINKED ISSUE #6659: Fix `*-case|*-unknown` false positives for CSS-in-JS template literals

It seems that rule `function-no-unknown` is triggering a false positive when functions are called within an anonymous function

```
import styled from "styled-components"

function getColor() {
  return "red"
}

const Foo = styled.div`
  color: ${() => getColor()}; // Unexpected unknown function "getColor"
`

const Bar = styled.div`
  color: ${getColor()}; // OK
`
```

```
module.exports = {
  extends: [
    "stylelint-config-standard",
  ],
  rules: {},
  customSyntax: "postcss-styled-syntax",
}
```

`stylelint \"./src/**/*.tsx\"`

`15.1.0`

```
"postcss-styled-syntax": "0.3.1",
"stylelint": "15.1.0",
"stylelint-config-standard": "30.0.1"
```

It should not report `Unexpected unknown function "getColor"` since we are inside a literal expression

### What actually happened?

```
✖  Unexpected unknown function "getColor"  function-no-unknown
```

yes, css-in-js

### Proposal to fix the bug

_No response_

hudochenkov:

Fixes #6659.

All rules has same issues: trying to lint non-standard value. Based on test one rule (`unit-no-unknown`) has a “regression”. Since the fix ignores declaration with non-standard values. It means that previous linted declaration with dollar-variables are now ignored as well. Given focus of Stylelint on standard CSS only, and policy to ignore non-standard syntax, I consider this “regression” acceptable as it doesn't break builds or report new violation.

### DIFF
--- a/.changeset/five-waves-relax.md
+++ b/.changeset/five-waves-relax.md
@@ -2,8 +2,8 @@
 "stylelint": patch
 ---
 
-Fixed: `annotation-no-unknown` false positives in CSS-in-JS syntax
-Fixed: `function-name-case` false positives in CSS-in-JS syntax
-Fixed: `function-no-unknown` false positives in CSS-in-JS syntax
-Fixed: `unit-no-unknown` false positives in CSS-in-JS syntax
-Fixed: `value-keyword-case` false positives in CSS-in-JS syntax
+Fixed: `annotation-no-unknown` false positives for CSS-in-JS template literals
+Fixed: `function-name-case` false positives for CSS-in-JS template literals
+Fixed: `function-no-unknown` false positives for CSS-in-JS template literals
+Fixed: `unit-no-unknown` false positives for CSS-in-JS template literals
+Fixed: `value-keyword-case` false positives for CSS-in-JS template literals
--- a/.changeset/five-waves-relax.md
+++ b/.changeset/five-waves-relax.md
@@ -1,0 +1,9 @@
+---
+"stylelint": patch
+---
+
+Fixed: `annotation-no-unknown` false positives in CSS-in-JS syntax
+Fixed: `function-name-case` false positives in CSS-in-JS syntax
+Fixed: `function-no-unknown` false positives in CSS-in-JS syntax
+Fixed: `unit-no-unknown` false positives in CSS-in-JS syntax
+Fixed: `value-keyword-case` false positives in CSS-in-JS syntax
--- a/lib/rules/annotation-no-unknown/index.js
+++ b/lib/rules/annotation-no-unknown/index.js
@@ -3,6 +3,7 @@
 const valueParser = require('postcss-value-parser');
 
 const getDeclarationValue = require('../../utils/getDeclarationValue');
+const isStandardSyntaxValue = require('../../utils/isStandardSyntaxValue');
 const optionsMatches = require('../../utils/optionsMatches');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
@@ -45,6 +46,10 @@ const rule = (primary, secondaryOptions) => {
 		 * @param {import('postcss').Declaration} decl
 		 */
 		function checkStatement(decl) {
+			if (!isStandardSyntaxValue(decl.value)) {
+				return;
+			}
+
 			if (decl.important) return;
 
 			if (!decl.value.includes('!')) return;
--- a/lib/rules/annotation-no-unknown/index.js
+++ b/lib/rules/annotation-no-unknown/index.js
@@ -46,9 +46,7 @@ const rule = (primary, secondaryOptions) => {
 		 * @param {import('postcss').Declaration} decl
 		 */
 		function checkStatement(decl) {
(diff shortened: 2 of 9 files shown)


lib/rules/annotation-no-unknown/index.js
```diff
@@ -45,6 +46,10 @@ const rule = (primary, secondaryOptions) => {
 		 * @param {import('postcss').Declaration} decl
 		 */
 		function checkStatement(decl) {
+			if (!isStandardSyntaxValue(decl.value)) {
+				return;
+			}
```
jeddy3, line 51:
```suggestion
			if (!isStandardSyntaxValue(decl.value)) return;
```

.changeset/five-waves-relax.md
```diff
@@ -0,0 +1,9 @@
+---
+"stylelint": patch
+---
+
+Fixed: `annotation-no-unknown` false positives in CSS-in-JS syntax
+Fixed: `function-name-case` false positives in CSS-in-JS syntax
+Fixed: `function-no-unknown` false positives in CSS-in-JS syntax
+Fixed: `unit-no-unknown` false positives in CSS-in-JS syntax
+Fixed: `value-keyword-case` false positives in CSS-in-JS syntax
```
jeddy3, line 9:
```suggestion
Fixed: `*-case|*-unknown` false positives for CSS-in-JS template literals
```

lib/rules/function-name-case/index.js
```diff
@@ -53,6 +54,10 @@ const rule = (primary, secondaryOptions, context) => {
 		}
 
 		root.walkDecls((decl) => {
+			if (!isStandardSyntaxValue(decl.value)) {
+				return;
+			}
```
jeddy3, line 59:
```suggestion
			if (!isStandardSyntaxValue(decl.value)) return;
```

lib/rules/function-no-unknown/index.js
```diff
@@ -57,6 +58,10 @@ const rule = (primary, secondaryOptions) => {
 		root.walkDecls((decl) => {
 			const { value } = decl;
 
+			if (!isStandardSyntaxValue(value)) {
+				return;
+			}
```
jeddy3, line 63:
```suggestion
			if (!isStandardSyntaxValue(value)) return;
```

lib/rules/unit-no-unknown/index.js
```diff
@@ -158,7 +159,7 @@ const rule = (primary, secondaryOptions) => {
 			check(atRule, atRule.params, atRuleParamIndex);
 		});
 		root.walkDecls((decl) => {
-			if (!isStandardSyntaxDeclaration(decl)) {
+			if (!isStandardSyntaxDeclaration(decl) || !isStandardSyntaxValue(decl.value)) {
 				return;
 			}
```
jeddy3, line 164:
```suggestion
			if (!isStandardSyntaxDeclaration(decl)) return;
			
			if (!isStandardSyntaxValue(decl.value)) return;
```

lib/rules/value-keyword-case/index.js
```diff
@@ -79,6 +79,10 @@ const rule = (primary, secondaryOptions, context) => {
 			const propLowerCase = decl.prop.toLowerCase();
 			const value = decl.value;
 
+			if (!isStandardSyntaxValue(value)) {
+				return;
+			}
```
jeddy3, line 84:
```suggestion
			if (!isStandardSyntaxValue(value)) return;
```

jeddy3 (changes requested):
Thank you! Just some minor nits, otherwise LGTM.

(We're adopting a shorthand for early returns as we often use them in rules.)

> Given focus of Stylelint on standard CSS only, and policy to ignore non-standard syntax, I consider this “regression” acceptable as it doesn't break builds or report new violation.

SGTM.

.changeset/five-waves-relax.md
hudochenkov: I think this would be confusing to users. I would understand proposed line as if _all_ rules matching the pattern would have fixed false positives. I fixed only rules that I personally had issues with. I don't know about other rules.

hudochenkov:
OMG! Check out this PR number: 6666! :)

.changeset/five-waves-relax.md
jeddy3: > I fixed only rules that I personally had issues with. I don't know about other rules.

Oh, I didn't clock that. Makes sense. I've updated the suggestion.

jeddy3:
> OMG! Check out this PR number: 6666! :)

Ha, I noticed it too. Nice! It will be another 1111 issues/prs before we see a sequence like that again.

jeddy3 (approved):
LGTM, thank you.

ybiquitous (approved):
Thank you. LGTM 👍🏼

Issue #6560, opened 2023-01-05
Add `media-feature-name-value-no-unknown`


mattxwang:
### What is the problem you're trying to solve?

In the discussion for https://github.com/stylelint/stylelint/pull/6550, @romainmenke mentioned certain types of issues involving invalid media query ranges. Their examples include:

```css
@media (width <= 100) {} /* number is not a length */
@media (width <= 50%) {} /* percentage is not a length */
@media (min-aspect-ratio: 16px/9px) {} /* length values do not make up a ratio */
```

I could also see the potential for auditing queries with the wrong number or types or arguments, ex:

```css
@media (width <= hover) /* invalid: hover is a keyword, not a length */
@media (color <= 16px)  /* invalid: color expects an integer */

/* grid is *not* a range, but just to also call out: */
@media (grid: 10)   /* invalid: grid expects a boolean (<mq-boolean>) of 1 or 0 */
@media (grid: 10px) /* invalid: grid expects a boolean (<mq-boolean>) of 1 or 0 */
```

I'm not sure if this is an appropriate rule for Stylelint, especially given its potential complexity.

### What solution would you like to see?

- Name: `media-feature-name-value-no-unknown`
- Primary option: `true`
- Secondary options: none
- Autofixable: No
- Message: `Unexpected unknown value "${value}" for media feature "${name}"`
- Description: Disallow unknown values for media features.
- Section: "Avoid Errors" -> "Unknown"
- Extended description: something about valid value types for specific media features

ybiquitous:
@mattxwang Thanks for opening the issue!

romainmenke:
It would be relatively straightforward with our media query parser to check if a media query is syntactically valid or not.

Types of errors :
1. the `@media` params or not even remotely similar to a media query
   - `isMediaQueryInvalid(x)` will be true
   - e.g. `@media @foo {}`
2. the type is invalid
   - `isMediaQueryWithType(x)` will be true
   - `x.getMediaType()` is not one of `all | print | screen | tty | tv | projection | handheld | braille | embossed | aural | speech`
   - e.g. `@media tablet and (min-width: 800px)`
3. the parenthesis and combinators are incorrect
   - `isGeneralEnclosed(x)` will be true
   - e.g. `@media ((min-width: 300px) and (hover: hover) or (aspect-ratio: 1 / 1))`
4. syntactically incorrect values for media queries in general are used
   - `isGeneralEnclosed(x)` will be true
[...]

PR #6669, opened 2023-02-17
Fix `declaration-property-value-no-unknown` false positives for descriptors

### LINKED ISSUE #6663: Fix `declaration-property-value-no-unknown` false positives for descriptors

```css
@font-face {
    font-family: 'AwesomeVariableFont';
    font-style: normal;
    font-weight: 100 700;
    src: url('./fonts/AwesomeVariableFont.woff2') format('woff2-variations');
}
```

For variable fonts, the `font-weight` descriptor accepts the minimum and maximum weights supported by the font:

* [MDN: `@font-face`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face#font-weight)
* [MDN: `@font-face`'s `font-weight` descriptor](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-weight) (compare also with [MDN: `font-weight` property](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight))

```yaml
rules:
  declaration-property-value-no-unknown: true
```

npx stylelint style.css

15.1.0

No problems to be reported.

### What actually happened?

The following problem was reported:

```
style.css
 4:22  ✖  Unexpected unknown value "700" for property "font-weight"  declaration-property-value-no-unknown
```

No, only native CSS syntax is involved.

### Proposal to fix the bug

_No response_

FloEdelmann:

Closes #6663.

No, it's self-explanatory.

### DIFF
--- a/.changeset/cyan-kangaroos-hope.md
+++ b/.changeset/cyan-kangaroos-hope.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fix false-positives for at-rule descriptors in `declaration-property-value-no-unknown`
--- a/.changeset/cyan-kangaroos-hope.md
+++ b/.changeset/cyan-kangaroos-hope.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fix false-positives for at-rule descriptors in `declaration-property-value-no-unknown`
+Fixed: `declaration-property-value-no-unknown` false positives for at-rule descriptors
--- a/lib/rules/declaration-property-value-no-unknown/index.js
+++ b/lib/rules/declaration-property-value-no-unknown/index.js
@@ -109,15 +109,11 @@ const rule = (primary, secondaryOptions) => {
 				return;
 			}
 
-			let error;
-			const parent = decl.parent;
-
-			if (parent && parent.type === 'atrule') {
-				// @ts-expect-error -- CSSTree types are not complete
-				({ error } = forkedLexer.matchAtruleDescriptor(parent.name, prop, cssTreeValueNode));
-			} else {
-				({ error } = forkedLexer.matchProperty(prop, cssTreeValueNode));
-			}
+			const { error } =
+				decl?.parent?.type === 'atrule'
+					? // @ts-expect-error -- CSSTree types are not complete
+					  forkedLexer.matchAtruleDescriptor(decl.parent.name, prop, cssTreeValueNode)
+					: forkedLexer.matchProperty(prop, cssTreeValueNode);
 
 			if (!error) return;
 
--- a/lib/rules/declaration-property-value-no-unknown/index.js
+++ b/lib/rules/declaration-property-value-no-unknown/index.js
@@ -86,6 +86,7 @@ const rule = (primary, secondaryOptions) => {
 
 			if (isPropIgnored(prop, value)) return;
 
+			/** @type {import('css-tree').CssNode} */
 			let cssTreeValueNode;
 
 			try {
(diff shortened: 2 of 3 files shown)


lib/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -108,7 +109,15 @@ const rule = (primary, secondaryOptions) => {
...
-			const { error } = forkedLexer.matchProperty(prop, cssTreeValueNode);
+			let error;
+			const parent = decl.parent;
+
+			if (parent && parent.type === 'atrule') {
+				// @ts-expect-error -- CSSTree types are not complete
+				({ error } = forkedLexer.matchAtruleDescriptor(parent.name, prop, cssTreeValueNode));
+			} else {
+				({ error } = forkedLexer.matchProperty(prop, cssTreeValueNode));
+			}
```
jeddy3, line 120:
```suggestion
				const {error} = decl?.parent?.type === 'atrule'
				// @ts-expect-error -- CSSTree types are not complete
				? forkedLexer.matchAtruleDescriptor(decl.parent?.name, prop, cssTreeValueNode)
				: forkedLexer.matchProperty(prop, cssTreeValueNode);
```

I believe we can use the optional chaining operator in `15.0.0` as we dropped support for Node.js 12.

@ybiquitous may have a better suggestion for refactoring this snippet. He's amazing at them.

.changeset/cyan-kangaroos-hope.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fix false-positives for at-rule descriptors in `declaration-property-value-no-unknown`
```
jeddy3, line 5:
```suggestion
Fixed: `declaration-property-value-no-unknown` false positives for descriptors
```

nit for [conventions](https://stylelint.io/maintainer-guide/pull-requests).

jeddy3 (changes requested):
@FloEdelmann This is fantastic, thank you!

It's looking good. I've requested a couple of small changes.

.changeset/cyan-kangaroos-hope.md
FloEdelmann: I'd still explicitly mention "at-rules", since most people will probably not know what "descriptors" are.

```suggestion
Fixed: `declaration-property-value-no-unknown` false positives for at-rule descriptors
```

lib/rules/declaration-property-value-no-unknown/index.js
FloEdelmann: We can actually omit the optional chaining in the `matchAtruleDescriptor` function call because it is already checked to exist in the ternary condition.

I refactored the code to this suggestion (minus the superfluous optional chaining operator, plus formatting via Prettier) in [`ca25f9b`](https://github.com/stylelint/stylelint/pull/6669/commits/ca25f9bb0dac9a81c3118ffe96fc20a76a00e6c0).

lib/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -108,7 +109,11 @@ const rule = (primary, secondaryOptions) => {
 				return;
 			}
 
-			const { error } = forkedLexer.matchProperty(prop, cssTreeValueNode);
+			const { error } =
+				decl?.parent?.type === 'atrule'
+					? // @ts-expect-error -- CSSTree types are not complete
+					  forkedLexer.matchAtruleDescriptor(decl.parent.name, prop, cssTreeValueNode)
```
ybiquitous, line 115: I'll send a patch to `@types/css-tree`. 💪🏼

ybiquitous (approved):
Cool. LGTM 👏🏼

jeddy3 (approved):

LGTM.

lib/rules/declaration-property-value-no-unknown/index.js
FloEdelmann: Thanks! Note that more functions from the `Lexer` class are missing, so maybe you can add them all in one batch?

Also, I used `@ts-expect-error` instead of `@ts-ignore`, so that this PR can be merged first and will throw an error once the `css-tree` types are fixed.

FloEdelmann: [`checkAtrulePrelude`](https://github.com/csstree/csstree/blob/612cc5f2922b2304869497d165a0cc65257f7a8b/lib/lexer/Lexer.js#L244) might be needed for #6560.

ybiquitous: I've opened PR https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64410. Any feedback is welcome if you have.

PR #6589, opened 2023-01-24
Add `messageArgs` to 76 rules

kizu:
Adds missing `messageArgs` to most rules.

> Which issue, if any, is this issue related to?

- Original issue: https://github.com/stylelint/stylelint/issues/4117
- Original PR implementing the `messageArgs`: https://github.com/stylelint/stylelint/pull/6312

> Is there anything in the PR that needs further explanation?

No, it's self-explanatory.

PR #6678, opened 2023-02-20
Add `selector-anb-no-unmatchable` rule

### LINKED ISSUE #5907: Add `selector-anb-no-unmatchable`

I'd like to catch `nth-*` selectors with zero index like:

```css
a:nth-child(0) {}
```

Although it's valid CSS, it won't match anything.

From [the spec](https://www.w3.org/TR/css-syntax-3/#anb-microsyntax):

> If both _A_ and _B_ are 0, the pseudo-class represents no element in the list.

See: https://twitter.com/dan_abramov/status/1492277543185096704

Related to https://github.com/stylelint/stylelint/issues/5575 (which I suggest we break into separate rules).

A new rule to help me [avoid errors](https://stylelint.io/user-guide/rules/list#avoid-errors) in structural pseudo-classes:

- [Tree-Structural pseudo-classes](https://www.w3.org/TR/selectors-4/#structural-pseudos)
- [Grid-Structural Selectors](https://www.w3.org/TR/selectors-4/#table-pseudos)

It'll catch:

```css
a:nth-child(0) {}
```

```css
a:nth-child(+0n) {}
```

```css
a:nth-child(0n+0) {}
```

```css
a:nth-child(0n + 0) {}
```

```css
a:nth-child(0n-0) {}
```

```css
a:nth-child(-0n-0) {}
```

```css
a:nth-of-type(0n) {}
```

It'd also disallow zero index _of S_ selectors, e.g.:

```css
:nth-child(0 of a) {}
```

This is a good candidate for using an _An+B_ parser, e.g. the one from SWC if it's available, otherwise, we can release for just the simple case of `0` as that's the most common mistake.

- Name: `selector-anb-no-unmatchable`
- Primary option: `true`
- Secondary options: None
- Autofixable: No
- Message: `Unexpected unmatchable An+B selector "${puesdoClass}"` (e.g. Unexpected unmatchable An+B selector "nth-child(0)")
- Description: "Disallow unmatchable An+B selectors"
- Extended description: (Something about zero indexes being a common mistake)
- Section: "Avoid errors" -> "Unmatchable"

Any thoughts on the rule, its name and scope?

--- 

Incidentally, there's a set of rules to [enforce (non-stylistic) conventions](https://stylelint.io/user-guide/rules/list#enforce-conventions) in this area too, e.g. preferring:

- `:first-child` over `:nth-child(1)` (or vice versa)
- `:nth-child(even)` over `:nth-child(2n+0)` (or vice versa)
- and so on

If you'd like to add any of these to Stylelint, please open a separate issue.

mattxwang:

Closes #5907.

I've left a few comments in review. 

In addition, I'm a bit confused about [Grid-Structural Selectors](https://www.w3.org/TR/selectors-4/#table-pseudos). The spec itself doesn't have an example of `:nth-col()` or `:nth-last-col()`, the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-col)' code example doesn't render for me on either Chrome or Firefox and has no browser compat data, and I haven't found much online on correct use. I must be missing something - some guidance would be helpful!


lib/rules/selector-anb-no-unmatchable/index.js
```diff
@@ -0,0 +1,117 @@
...
+					checkSelector(cssTreeSelector);
+				} catch (e) {
+					report({
+						message: messages.rejectedParseError(ruleNode.selector),
+						node: ruleNode,
+						index: 0,
+						endIndex: 0,
+						result,
+						ruleName,
+					});
```
mattxwang, line 68: Wasn't exactly sure what to write here / what to write tests for. Are there other patterns that I should look at?

lib/rules/selector-anb-no-unmatchable/README.md
```diff
@@ -0,0 +1,75 @@
...
+Disallow unmatchable An+B selectors.
+
+<!-- prettier-ignore -->
+```css
+a:nth-child(0n+0) {}
+/*↑             ↑
+ * This unmatchable An+B selector */
+```
+
+[An+B selectors](https://drafts.csswg.org/css-syntax/#anb) are one-indexed. Selectors that always evaluate to `0` will not match any elements.
```
mattxwang, line 12: Any suggestions on a better thing to highlight / extended description?

docs/user-guide/rules.md
```diff
@@ -73,6 +73,12 @@ Disallow overrides with these `no-overrides` rules.
 
 - [`declaration-block-no-shorthand-property-overrides`](../../lib/rules/declaration-block-no-shorthand-property-overrides/README.md): Disallow shorthand properties that override related longhand properties (Ⓡ & Ⓢ).
 
+### Unmatchable
+
+Disallow unmatchable things with these `no-unmatchable` rules.
+
+- [`selector-anb-no-unmatchable`](../../lib/rules/selector-anb-no-unmatchable/README.md): Disallow unmatchable An+B selectors
```
Mouvedia, line 80: nit

full stop missing

docs/user-guide/rules.md
```diff
@@ -73,6 +73,12 @@ Disallow overrides with these `no-overrides` rules.
 
 - [`declaration-block-no-shorthand-property-overrides`](../../lib/rules/declaration-block-no-shorthand-property-overrides/README.md): Disallow shorthand properties that override related longhand properties (Ⓡ & Ⓢ).
 
+### Unmatchable
+
+Disallow unmatchable things with these `no-unmatchable` rules.
```
Mouvedia, line 78: minor

"things" should be replaced

lib/rules/selector-anb-no-unmatchable/README.md
Mouvedia, line 12: question

Should it link to a draft?
Would https://www.w3.org/TR/css-syntax-3/#anb-microsyntax be better?

lib/rules/selector-anb-no-unmatchable/index.js
```diff
@@ -0,0 +1,117 @@
...
+
+const report = require('../../utils/report');
+const ruleMessages = require('../../utils/ruleMessages');
+const validateOptions = require('../../utils/validateOptions');
+
+const ruleName = 'selector-anb-no-unmatchable';
+
+const messages = ruleMessages(ruleName, {
+	rejected: (pseudoClass) => `Unexpected unmatchable An+B selector "${pseudoClass}"`,
+	rejectedParseError: (selector) => `Cannot parse selector "${selector}"`,
```
Mouvedia, line 18: question

Iv seen one rejectedParseError test for the `declaration-property-value-no-unknown` rule.
Does it need to be covered?

lib/rules/selector-anb-no-unmatchable/__tests__/index.js
```diff
@@ -0,0 +1,145 @@
...
+		},
+		{
+			code: 'a:nth-of-type(1)',
+		},
+		{
+			code: 'a:nth-last-of-type(1)',
+		},
+	],
+
+	reject: [
```
Mouvedia, line 39: question

What happens with https://www.w3.org/TR/css-syntax-3/#example-3d25855e?

lib/rules/selector-anb-no-unmatchable/index.js
```diff
@@ -0,0 +1,117 @@
...
+					pseudoClassSelector.children.forEach((child) => {
+						if (child.type !== 'Nth' || child.nth.type !== 'AnPlusB') {
+							return;
+						}
+
+						if (isUnmatchableNth(child.nth)) {
+							report({
+								message: messages.rejected(pseudoClassSelector.name),
+								node: ruleNode,
+								index: pseudoClassSelector.loc?.start.column,
```
Mouvedia, line 101: question

Is `undefined` allowed as an index?

lib/rules/selector-anb-no-unmatchable/index.js
```diff
@@ -0,0 +1,117 @@
...
+						node: ruleNode,
+						index: 0,
+						endIndex: 0,
+						result,
+						ruleName,
+					});
+				}
+			});
+
+			function checkSelector(/** @type {import('css-tree').CssNode} */ selector) {
```
Mouvedia, line 72: minor

It's used at least twice; maybe @typedef it.

lib/rules/selector-anb-no-unmatchable/index.js
```diff
@@ -0,0 +1,117 @@
...
+		}
+
+		root.walkRules((ruleNode) => {
+			if (!isStandardSyntaxRule(ruleNode)) {
+				return;
+			}
+
+			ruleNode.selectors.forEach((selector) => {
+				/** @type {import('css-tree').CssNode} */
+				let cssTreeSelector;
```
Mouvedia, line 55: It should be a `const` at l58.

lib/rules/selector-anb-no-unmatchable/__tests__/index.js
mattxwang: it errors; I'll add some test cases for this!

lib/rules/selector-anb-no-unmatchable/index.js
mattxwang: I don't think so; there's a "soft" guarantee that if we include `positions` in the parse call, then `loc` is guaranteed to exist. Is there a better way for me to encode this? Ex, should I throw a parse error when `loc` doesn't exist?

docs/user-guide/rules.md
mattxwang: Gotcha, what's a good replacement? I'm not the best at writing docs 😅

Mouvedia: Me neither. Wait for jeddy3.

lib/rules/selector-anb-no-unmatchable/index.js
```diff
@@ -0,0 +1,121 @@
...
+		root.walkRules((ruleNode) => {
+			if (!isStandardSyntaxRule(ruleNode)) {
+				return;
+			}
+
+			ruleNode.selectors.forEach((selector, i) => {
+				try {
+					const cssTreeSelector = parse(selector, { context: 'selector', positions: true });
+
+					checkSelector(cssTreeSelector);
```
ybiquitous, line 57: [suggest] Moving `checkSelector(cssTreeSelector)` out of the `try...catch` block seems better to me because the current code may catch an unexpected error by `checkSelector()` .

So, I suggest a refactoring like this:

```js
let cssTreeSelector;

try {
	cssTreeSelector = parse(selector, { context: 'selector', positions: true });
} catch {
	// ...
	return;
}

checkSelector(cssTreeSelector);
```

lib/rules/selector-anb-no-unmatchable/index.js
```diff
@@ -0,0 +1,121 @@
...
+						.slice(0, i)
+						.map((sel) => sel.length)
+						.reduce((c, a) => c + a, 0);
+					const endIndex = index + selector.length;
+
+					report({
+						message: messages.rejectedParseError(selector),
+						node: ruleNode,
+						index,
+						endIndex,
```
ybiquitous, line 69: [suggest]

- We may simply use `word` instead of calculating `index` and `endIndex` because a parse error seems to rarely occur.
- How about using `messageArgs`? See #6589

```suggestion
					report({
						message: messages.rejectedParseError,
						messageArgs: [selector],
						node: ruleNode,
						word: selector,
```

lib/rules/selector-anb-no-unmatchable/index.js
ybiquitous: Even if passing `undefined` to `index` and `endIndex`, I think there is no inconvenience. Because the lint problem location (`column` and `endColumn` will point to the whole selector. Also, it seems a rare case.

lib/rules/selector-anb-no-unmatchable/index.js
```diff
@@ -0,0 +1,121 @@
...
+					}
+
+					pseudoClassSelector.children.forEach((child) => {
+						if (child.type !== 'Nth' || child.nth.type !== 'AnPlusB') {
+							return;
+						}
+
+						if (isUnmatchableNth(child.nth)) {
+							report({
+								message: messages.rejected(pseudoClassSelector.name),
```
ybiquitous, line 103: [suggest]

- How about prepending a colon (`:`) to a selector in a reported message? Because a pseudo-selector is often described with `:` like [`:nth-child()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child). If doing so, we may need to calculate `index - 1`.
- I suggest using `messageArgs` also here.

```suggestion
								message: messages.rejected,
								messageArgs: [`:${pseudoClassSelector.name}`],
```

ybiquitous (commented):
@mattxwang Thanks for the pull request, great work! 👏🏼 

I've left minor suggestions, so please check them.

mattxwang:
Thank you for the helpful feedback @ybiquitous! I need to get in the habit of adding `messageArgs` to new rules, and I agree that adding the `:` can make it easier for users to understand the message.

I think I've addressed all the comments, let me know what you think!

ybiquitous (approved):
@mattxwang Thanks, LGTM! 👍🏼 

I would suggest waiting for @jeddy3's review.

.changeset/cuddly-olives-learn.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: `selector-anb-no-unmatchable`
```
jeddy3, line 5:
```suggestion
Added: `selector-anb-no-unmatchable` rule
```

docs/user-guide/rules.md
jeddy3: Funnily enough, we've used "thing(s)" since the beginning of [our docs](https://github.com/stylelint/stylelint/blob/main/docs/developer-guide/rules.md#define-the-rule)

> the [thing](http://apps.workflower.fi/vocabs/css/en) the rule applies to, e.g. at-rule

We haven't been able to think of anything better. If anyone does, let's roll it out across all the instances of things in another PR. At the moment, let's keep it as things in this PR to be consistent with the other sections in the doc.

lib/rules/selector-anb-no-unmatchable/README.md
```diff
@@ -0,0 +1,77 @@
...
+The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+a:nth-child(0) {}
+```
+
+<!-- prettier-ignore -->
+```css
+a:nth-child(0n) {}
```
jeddy3, line 29:
```suggestion
a:nth-last-child(0n) {}
```

lib/rules/selector-anb-no-unmatchable/README.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+<!-- prettier-ignore -->
+```css
+a:nth-child(0n+0) {}
```
jeddy3, line 34:
```suggestion
a:nth-of-type(0n+0) {}
```

lib/rules/selector-anb-no-unmatchable/README.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+<!-- prettier-ignore -->
+```css
+a:nth-child(0 of a) {}
```
jeddy3, line 39:
```suggestion
a:nth-last-of-type(0 of a) {}
```

lib/rules/selector-anb-no-unmatchable/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+<!-- prettier-ignore -->
+```css
+a:nth-of-type(0) {}
+```
+
+<!-- prettier-ignore -->
+```css
+a:nth-last-of-type(0) {}
+```
```
jeddy3, line 55:
```suggestion
```

Let's keep the examples to the minimum to explain the rule's intent.

lib/rules/selector-anb-no-unmatchable/README.md
```diff
@@ -0,0 +1,77 @@
...
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+a:nth-child(1) {}
+```
+
+<!-- prettier-ignore -->
+```css
+a:nth-child(1n) {}
```
jeddy3, line 66:
```suggestion
a:nth-last-child(1n) {}
```

lib/rules/selector-anb-no-unmatchable/README.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+<!-- prettier-ignore -->
+```css
+a:nth-child(1n+0) {}
```
jeddy3, line 71:
```suggestion
a:nth-of-type(1n+0) {}
```

lib/rules/selector-anb-no-unmatchable/README.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+<!-- prettier-ignore -->
+```css
+a:nth-child(0n+1) {}
```
jeddy3, line 76:
```suggestion
a:nth-last-of-type(1 of a) {}
```

We can mirror the accept examples in the reject ones.

lib/rules/selector-anb-no-unmatchable/index.js
```diff
@@ -0,0 +1,120 @@
...
+				} catch (e) {
+					report({
+						message: messages.rejectedParseError,
+						messageArgs: [selector],
+						node: ruleNode,
+						word: selector,
+						result,
+						ruleName,
+					});
+
```
jeddy3, line 67:
```suggestion
```

Let's silently catch parse errors, as it's out of the scope of this rule.

CSSTree is less forgiving than the PostCSS selector parser, and this will likely flag non-standard language extensions.

jeddy3 (changes requested):
@mattxwang Thanks. It's looking great.

I've requested one significant change and some minor nits.

mattxwang:
Thanks for the feedback @jeddy3; I think I've addressed all of the comments!

Two quick questions:

1. should I add test cases for nonstandard selectors / situations where CSSTree will swallow the error? basically, the rule would accept/bail out here
    - as a note, this will increase the test coverage!
3. should I also be doing grid-structural selectors? I was having some trouble finding good examples -- see PR description

jeddy3:
Thanks for making the changes:

> should I add test cases for nonstandard selectors / situations where CSSTree will swallow the error?

Yeap, a single test case for that would be swell.

> should I also be doing grid-structural selectors? 

Let's leave them out for now.

lib/rules/selector-anb-no-unmatchable/__tests__/index.js
```diff
@@ -40,6 +40,14 @@ testRule({
...
 			code: 'a:nth-child(even) {}',
 		},
+		{
+			code: 'a:nth-child(3n + - 6) {}',
+			description: 'CSSTree parse error not flagged',
+		},
+		{
+			code: '.mixin(@variable: x):nth-child(0) {}',
+			description: 'nonstandard syntax (LESS parametric mixin)',
+		},
```
jeddy3, line 50:
```suggestion
```

Let's avoid adding CSS-like code in our tests.

jeddy3 (changes requested):
Thanks for making the changes. One minor request, then I think it LGTM.

lib/rules/selector-anb-no-unmatchable/__tests__/index.js
mattxwang: Ah! I think I misinterpreted your comment earlier, thanks for pointing this out. Have removed it.

jeddy3 (approved):
Thank you for making the changes.

LGTM.

PR #6684, opened 2023-02-27
Refactor `declaration-property-value-no-unknown` to fix `@ts-expect-error`

ybiquitous:

Follow-up to #6683 (see also DefinitelyTyped/DefinitelyTyped#64410)

By updating `@types/css-tree`, we no longer need to use `@ts-expect-error` to suppress TypeScript errors about `css-tree`.

In addition, this pull request refactors the code to use the `isAtRule` type guard. If not using the type guard, `parent` is not inferred as `AtRule` and so another error occurs.


ybiquitous:
[off-topic] I've enabled the required status checks for each CI job in the branch protection rule. Purposes:

- To clarify which jobs are required
- To prevent accidental merging even though status checks fail
- To make GitHub merge queue work (currently disabled, though)

jeddy3 (approved):
LGTM, thank you!

PR #6686, opened 2023-02-27
Refactor `safeChdir()` test util to fix `@ts-expect-error`

ybiquitous:

None.

The `@jest/globals` package is included in Jest by default:

```console
$ npm ls '@jest/globals'
stylelint@15.2.0 /Users/masafumi.koba/git/stylelint/stylelint
└─┬ jest@29.4.3
  └─┬ @jest/core@29.4.3
    └─┬ jest-runtime@29.4.3
      └── @jest/globals@29.4.3
```

Also, because the refactored util is used for only tests, we can use the modern module syntax `node:fs/promises` for Node.js. See also: <https://nodejs.org/api/fs.html>


jeddy3 (approved):
LGTM, thank you.

Issue #6694, opened 2023-03-03
Fix type definitions for commonjs


baseballyama:
### What steps are needed to reproduce the bug?

After upgrade stylelint from `14.16.1` to `15.2.0`, below type error comes.
But I can run the program as same as before.
So I think just type definition file is wrong.

```js
const stylelint = require("stylelint");
stylelint.utils
```

![image](https://user-images.githubusercontent.com/19153718/222650245-018f6345-08e0-49f2-b1af-c23d60e7a23b.png)

### What Stylelint configuration is needed to reproduce the bug?

```js
const stylelint = require("stylelint");
stylelint.utils
```

### How did you run Stylelint?

This error occurs on my custom Stylelint rule js file

### Which version of Stylelint are you using?

15.2.0

### What did you expect to happen?

No type error

### What actually happened?

Type error

### Does the bug relate to non-standard syntax?

No

### Proposal to fix the bug

update `index.d.ts`.

```diff
	const stylelint: PublicApi;

-	export default stylelint;
+	export = stylelint;
```

PR #6695, opened 2023-03-03
Fix TypeScript error for CommonJS importing

baseballyama:

close: https://github.com/stylelint/stylelint/issues/6694

No.

### DIFF
--- a/.changeset/yellow-dodos-juggle.md
+++ b/.changeset/yellow-dodos-juggle.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: type definition for CommonJS
--- a/.changeset/yellow-dodos-juggle.md
+++ b/.changeset/yellow-dodos-juggle.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: type definition for CommonJS
+Fixed: TypeScript error for CommonJS importing
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -631,5 +631,5 @@ declare module 'stylelint' {
 
 	const stylelint: PublicApi;
 
-	export default stylelint;
+	export = stylelint;
 }
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -631,5 +631,5 @@ declare module 'stylelint' {
 
 	const stylelint: PublicApi;
 
-	export = stylelint;
+	export default stylelint;
 }
--- a/lib/index.js
+++ b/lib/index.js
@@ -32,4 +32,5 @@ const stylelint = Object.assign(postcssPlugin, {
 	},
 });
 
+// To adjust type for CommonJS, we export the default property.
 module.exports = { ...stylelint, default: stylelint };
--- a/lib/index.js
+++ b/lib/index.js
@@ -32,5 +32,6 @@ const stylelint = Object.assign(postcssPlugin, {
 	},
 });
 
-// To adjust type for CommonJS, we export the default property.
-module.exports = { ...stylelint, default: stylelint };
+module.exports = stylelint;
+// @ts-ignore -- To adjust type for CommonJS, we export the default property.
+module.exports.default = stylelint;
--- a/lib/index.js
+++ b/lib/index.js
@@ -32,4 +32,4 @@ const stylelint = Object.assign(postcssPlugin, {
 	},
 });
 
-module.exports = stylelint;
+module.exports = { ...stylelint, default: stylelint };


baseballyama:
Hmmm. CI failed. so I need to dig into the codebase more.

types/stylelint/index.d.ts
```diff
@@ -631,5 +631,5 @@ declare module 'stylelint' {
 
 	const stylelint: PublicApi;
 
-	export default stylelint;
+	export = stylelint;
```
ybiquitous, line 634: Thanks for the pull request! How about just adding a new `export`?

```suggestion
    export default stylelint;
	export = stylelint; // for CommonJS
```

baseballyama: Thank you for your fast response!
After applying your suggestion, now I gotanother ts error.

```
Module '"stylelint"' can only be default-imported using the 'esModuleInterop' flagts(1259)
index.d.ts(635, 2): This module is declared with 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.
```

image

ybiquitous: Good catch. It's difficult...

Then, does it make sense to use `require("...").default` instead?

```diff
-const stylelint = require("stylelint");
+const stylelint = require("stylelint").default;
```

Of course, we may need to change the following code:

https://github.com/stylelint/stylelint/blob/d2ca95a197669c9c18f87f7b77fa309cf4c2d5f7/lib/index.js#L35

```diff
 module.exports = stylelint; 
+module.exports.default = stylelint; 
```

baseballyama: Ohh! I thought about only how to fix the type, but yeah, I think now it works fine.

ybiquitous:
I'm not sure about CommonJS details, but the following form seems to work and pass tests:

```js
module.exports = stylelint;
module.exports.default = stylelint;
```

In addition, we may need to suppress a TS error like this:

```js
// NOTE: For CommonJS compatibility.
// @ts-expect-error -- TS2339: Property 'default' does not exist on type 'PublicApi'.
module.exports.default = stylelint;
```

baseballyama:
Oops, sorry. I did run the test but I couldn't find the test failure😫
Now I think we can pass the test.(I run `npm run test-coverage` locally and all tests passed.)

.changeset/yellow-dodos-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: type definition for CommonJS
```
ybiquitous, line 5:
```suggestion
Fixed: TypeScript error for CommonJS importing
```

ybiquitous:
Thanks for fixing the error! We can confirm PostCSS also uses the same way:

- https://github.com/postcss/postcss/blob/8.4.21/lib/postcss.js#L101
- https://github.com/postcss/postcss/pull/1459

jeddy3:
> Thanks for fixing the error! We can confirm PostCSS also uses the same way:

I haven't delved deeply into this PR, but is there anything in https://github.com/postcss/postcss/pull/1815 that we need to consider?

ybiquitous:
@jeddy3 Thanks for the information. It seems to be the same problem, but hard to find a suitable way for all the situations... 🤔 

Does anyone have a nice solution?

baseballyama:
Could you add more context about what https://github.com/postcss/postcss/pull/1815 solves?

After read https://github.com/postcss/postcss/pull/1815 , I tried to update `tsconfig.json` as below, still I have a type error even after restarting TS server.
So I couldn't get what issue https://github.com/postcss/postcss/pull/1815 solves with my limited knowledge.😫

```json
  "compilerOptions": {
    "module": "commonjs",
    "moduleResolution": "node16",
  }
```

image

```
Property 'utils' does not exist on type 'typeof import("stylelint")'.
```

ybiquitous:
In our case, it seems there are no problems. We can check the Stylelint types on <https://arethetypeswrong.github.io/>, but no problems are found (this website is introduced on <https://github.com/postcss/postcss/pull/1815#issuecomment-1433257343>).

I have created a small repo to include this PR change:
https://github.com/ybiquitous/stylelint-pr-6695

TypeScript compile (tsc) complains only the following CommonJS code. This code is reported as a problem on #6694, and it should be fixed by this PR.

```js
const stylelint = require("stylelint");
console.log(stylelint.utils);
//                    ~~~~~ TS2339: Property 'utils' does not exist on type 'typeof import("stylelint")'.
```

All the files are executed successfully:

```console
$ node index.mjs ; node index.cjs ; node index-default.cjs

{
  report: [Function: report],
  ruleMessages: [Function: ruleMessages],
  validateOptions: [Function: validateOptions],
  checkAgainstRule: [Function: checkAgainstRule]
}
{
  report: [Function: report],
  ruleMessages: [Function: ruleMessages],
  validateOptions: [Function: validateOptions],
  checkAgainstRule: [Function: checkAgainstRule]
}
{
  report: [Function: report],
  ruleMessages: [Function: ruleMessages],
  validateOptions: [Function: validateOptions],
  checkAgainstRule: [Function: checkAgainstRule]
}
```

In short, I think we can merge safely this PR. Any thoughts?

jeddy3 (approved):
Thanks for checking arethetypeswrong for Stylelint

> In short, I think we can merge safely this PR. Any thoughts?

I think so too.

remcohaszing:
Hey, I saw you referenced my PR for postcss, and you made some wrong conclusions.

This PR actually makes proper type definitions much uglier / more annoying to maintain by making the same mistake that was made in postcss. Can this be reverted, or at least not be released?

It's getting late. I can create a proper PR tomorrow.

jeddy3:
@remcohaszing Thank you for chiming in. We can hold off releasing it. Your input or a PR would be greatly appreciated!

PR #6696, opened 2023-03-04
Document `media-feature-name-no-vendor-prefix` as standard rule

valtlai:

None, as it’s a documentation fix.

The [rule is enabled by stylelint-config-standard](https://github.com/stylelint/stylelint-config-standard/blob/30.0.1/index.js#L81).


jeddy3 (approved):
Thank you, LGTM.

PR #6711, opened 2023-03-13
Fix `function-url-quotes` autofix for multiple `url()`

### LINKED ISSUE #6710: Fix `function-url-quotes` autofix for multiple `url()`

Run stylint for this:

```
.foo {
        background-image: url(../img/footer/needle.svg), url(../img/footer/tel.svg);
}
```

Related standard syntax: [background-image]( https://developer.mozilla.org/en-US/docs/Web/CSS/background-image?retiredLocale=de#formal_syntax)

```json
{
}

```

npx stylelint template/sass/test.scss --fix

`15.2.0`

.foo {
        background-image: url("../img/footer/needle.svg"), url("../img/footer/tel.svg");
}

### What actually happened?

The following stylelint errors were reported:

```
template/sass/test.scss
 2:31  ✖  Expected quotes around "url" function argument  function-url-quotes
 2:62  ✖  Expected quotes around "url" function argument  function-url-quotes

2 problems (2 errors, 0 warnings)
```
With `--fix` stylelint creates malformed Sass:

```
.foo {
        background-image: url("../img/footer/needle.svg"), ur"../img/footer/tel.svg"vg);
}

```

no

### Proposal to fix the bug

It might have to do with multiple urls in the rule. With just one url it works.

ybiquitous:

Closes #6710

This pull request rewrites the rule implementation to use the [`postcss-value-parser`](https://github.com/TrySound/postcss-value-parser) AST node rather than string-based indexes.

### DIFF
--- a/lib/rules/function-url-quotes/index.js
+++ b/lib/rules/function-url-quotes/index.js
@@ -3,6 +3,8 @@
 const atRuleParamIndex = require('../../utils/atRuleParamIndex');
 const declarationValueIndex = require('../../utils/declarationValueIndex');
 const functionArgumentsSearch = require('../../utils/functionArgumentsSearch');
+const getAtRuleParams = require('../../utils/getAtRuleParams');
+const getDeclarationValue = require('../../utils/getDeclarationValue');
 const isStandardSyntaxUrl = require('../../utils/isStandardSyntaxUrl');
 const optionsMatches = require('../../utils/optionsMatches');
 const report = require('../../utils/report');
@@ -44,6 +46,9 @@ const rule = (primary, secondaryOptions, context) => {
 			return;
 		}
 
+		const exceptEmpty = optionsMatches(secondaryOptions, 'except', 'empty');
+		const emptyArgumentPatterns = new Set(['', "''", '""']);
+
 		root.walkAtRules(checkAtRuleParams);
 		root.walkDecls(checkDeclParams);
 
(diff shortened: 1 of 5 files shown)


patches/postcss-value-parser+4.2.0.patch
```diff
@@ -0,0 +1,13 @@
...
++++ b/node_modules/postcss-value-parser/lib/parse.js
+@@ -174,7 +174,7 @@ module.exports = function(input) {
+       };
+       pos = next;
+ 
+-      if (name === "url" && code !== singleQuote && code !== doubleQuote) {
++      if ((name === "url" || name === "url-prefix") && code !== singleQuote && code !== doubleQuote) {
+         next -= 1;
+         do {
+           escape = false;
```
ybiquitous, line 13: [note] We cannot merge this PR including the patch. I've created an upstream issue:
https://github.com/TrySound/postcss-value-parser/issues/88

package.json
```diff
@@ -144,7 +144,7 @@
     "postcss-resolve-nested-selector": "^0.1.1",
     "postcss-safe-parser": "^6.0.0",
     "postcss-selector-parser": "^6.0.11",
-    "postcss-value-parser": "^4.2.0",
+    "postcss-value-parser": "github:TrySound/postcss-value-parser#pull/90/merge",
```
ybiquitous, line 147: [note] I've created the upstream PR TrySound/postcss-value-parser#90.

jeddy3:
If [`url-prefix()` is only used by the deprecated `@document`](https://csstree.github.io/docs/syntax/#Function:url-prefix), we can remove checkin `url-prefix()` from the rule. It'd be a non-breaking change, making the rule more permissive.

ybiquitous:
@jeddy3 Thanks for the feedback. Indeed, when seeing the [browser compatibility table of `@document`](https://developer.mozilla.org/en-US/docs/Web/CSS/@document#browser_compatibility), it seems only outdated Firefox versions support `url-prefix()`, and so it seems safe to remove the function from the test cases.

lib/rules/function-url-quotes/index.js
```diff
@@ -44,96 +46,78 @@ const rule = (primary, secondaryOptions, context) => {
...
-		 */
-		function addQuotesForDecl(node, args, index) {
-			const fixedName = `"${args}"`;
-			const openIndex = index - declarationValueIndex(node);
-			const closeIndex = openIndex + args.length;
+			const params = getAtRuleParams(atRule);
+			const startIndex = atRuleParamIndex(atRule);
+			const parsed = functionArgumentsSearch(
+				params,
+				/^(url|url-prefix|domain)$/i,
```
jeddy3, line 78:
```suggestion
				/^url$/i,
```

> so it seems safe to remove the function from the test cases.

And from the rule itself. Same for [`domain()`](https://csstree.github.io/docs/syntax/#Function:domain).

ybiquitous: Fixed via 9a103c1a

ybiquitous:
Thanks for the quick review!

PR #6712, opened 2023-03-14
Fix `value-keyword-case` false positives for Level 4 system colours

thewilkybarkid:
This adds the [CSS Color Module Level 4 system colours](https://www.w3.org/TR/css-color-4/#css-system-colors) to the list of keywords that are not changed.


jeddy3 (approved):
LGTM, thank you.

ybiquitous (approved):
Thanks, LGTM 👍🏼

Issue #6501, opened 2022-12-01
Add `media-feature-range-notation` autofix


jeddy3:
### What is the problem you're trying to solve?

The rule isn't autofixable.

### What solution would you like to see?

Have it autofixable so that:

- we can turn it on in our standard config
- make it consistent with our other [`*-notation` rules](https://stylelint.io/user-guide/rules#notation)

See https://github.com/stylelint/stylelint/pull/6497#issue-1467251029

It'll only be for the more modern `"context"` option as that addresses the limitations of the prefix approach.

We could reach for [a new media-query-list-parser](https://www.npmjs.com/package/@csstools/media-query-list-parser]).

It won't combine media features. For example:

```css
@media (min-width: 500px) and screen and (max-width: 1200px) {}
```

Becomes:

```css
@media (width >= 500px) and screen and (width <= 1200px) {}
```

Rather than:

```css
@media (500px <= width <= 500px) and screen {}
```

ybiquitous:
> We could reach for [a new media-query-list-parser](https://www.npmjs.com/package/@csstools/media-query-list-parser).

It sounds [postcss-media-query-parser](https://www.npmjs.com/package/postcss-media-query-parser) could be replaced!

romainmenke:
Let us know if you need any help using our media query list parser.
Currently it isn't very simple :/

I needed it for fallback plugins but haven't done the important bits for the wider community.
- no docs
- only a low level API

It is fully typed, so that helps a bit.

I will gladly spend time on docs and some more high level functions if there is a community need for this.

romainmenke:
After reading [this comment](https://github.com/stylelint/stylelint/issues/6509#issuecomment-1366311978) I realized that I left things vaguely here :)

It is not easy for me to provide better documentation and/or a more ergonomic API without specific use cases.

I've opened an issue to collect use cases here : https://github.com/csstools/postcss-plugins/issues/763

romainmenke:
I've been converting a few of the existing stylelint rules to our media query parser and ran into an issue.

Our parser is unable to handle sass, less and other non standard syntaxes.

For example : `@media (MIN-WIDTH: ($var + 10px)) { }`

This looks like a plain media feature with a name and a value but our parser will parse this as `general enclosed` because the value doesn't have the correct syntax.

`$var` for example is not one thing. It is a delim token (`$`) and an ident token (`var`).

[...]

PR #6742, opened 2023-03-28
Add `media-feature-range-notation` autofix

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/6501

Some bits are more complex/messy than I would like but I held of on trying to abstract this.
I personally prefer to only abstract away complexity after having found ±3 use cases.

But I don't mind spending time on refactors now if people with more experience in this code base have good ideas on how to improve this :)

### DIFF
--- a/lib/rules/media-feature-range-notation/index.js
+++ b/lib/rules/media-feature-range-notation/index.js
@@ -1,14 +1,20 @@
 'use strict';
 
-const mediaParser = require('postcss-media-query-parser').default;
-
-const { rangeTypeMediaFeatureNames } = require('../../reference/mediaFeatures.js');
 const atRuleParamIndex = require('../../utils/atRuleParamIndex');
-const isRangeContextMediaFeature = require('../../utils/isRangeContextMediaFeature');
-const isStandardSyntaxMediaFeatureName = require('../../utils/isStandardSyntaxMediaFeatureName');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
 const validateOptions = require('../../utils/validateOptions');
+const parseMediaQuery = require('../../utils/parseMediaQuery');
+const {
+	isMediaFeatureRange,
+	isMediaFeaturePlain,
+	isMediaFeature,
+	MediaFeatureRangeNameValue,
+	MediaFeatureName,
+} = require('@csstools/media-query-list-parser');
+const { rangeTypeMediaFeatureNames } = require('../../reference/mediaFeatures');
+const { TokenNode } = require('@csstools/css-parser-algorithms');
+const { TokenType } = require('@csstools/css-tokenizer');
 
 const ruleName = 'media-feature-range-notation';
 
@@ -18,10 +24,11 @@ const messages = ruleMessages(ruleName, {
 
 const meta = {
 	url: 'https://stylelint.io/user-guide/rules/media-feature-range-notation',
+	fixable: true,
 };
 
 /** @type {import('stylelint').Rule} */
-const rule = (primary) => {
+const rule = (primary, _secondaryOptions, context) => {
 	return (root, result) => {
 		const validOptions = validateOptions(result, ruleName, {
 			actual: primary,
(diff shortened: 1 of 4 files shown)


lib/rules/media-feature-range-notation/index.js
```diff
@@ -33,57 +40,101 @@ const rule = (primary) => {
...
+									{ value: unprefixedMediaFeature },
+								]),
+								node.name.before,
+								node.name.after.length > 0
+									? node.name.after
+									: [[TokenType.Whitespace, ' ', -1, -1, undefined]],
+							),
+							[operator, [TokenType.Delim, '=', -1, -1, { value: '=' }]],
+							node.value,
+						);
```
romainmenke, line 94: Creating a new `MediaFeatureRangeNameValue` is a bit elaborate and painful at this moment, but having easy access to the low level API makes it trivial to have the right whitespace around and in between all the parts.

If I were to create a high level API in the upstream package it would follow the CSS spec serialization rules and it wouldn't work the same.

This might be necessary complexity.

lib/rules/media-feature-range-notation/index.js
```diff
@@ -33,57 +40,101 @@ const rule = (primary) => {
...
+
+					let startIndex = 0;
+					let endIndex = atRule.params.length;
+
+					if (firstToken && lastToken) {
+						startIndex = firstToken[2];
+						endIndex = lastToken[3];
+					}
+
+					const atRuleIndex = atRuleParamIndex(atRule);
```
romainmenke, line 111: This is too elaborate and should have a helper.
My intention is to gather the `index` and `endIndex`.

We have similar but simpler code in `lib/rules/media-feature-name-unit-allowed-list/index.js:67`.

Maybe something with this signature : `function foo(atRule, startToken, endToken)` where `endToken` can be omitted?

lib/rules/media-feature-range-notation/index.js
```diff
@@ -1,14 +1,20 @@
...
+const {
+	isMediaFeatureRange,
+	isMediaFeaturePlain,
+	isMediaFeature,
+	MediaFeatureRangeNameValue,
+	MediaFeatureName,
+} = require('@csstools/media-query-list-parser');
+const { rangeTypeMediaFeatureNames } = require('../../reference/mediaFeatures');
+const { TokenNode } = require('@csstools/css-parser-algorithms');
+const { TokenType } = require('@csstools/css-tokenizer');
```
ybiquitous, line 17: [suggest] Can you move the imports of `@csstools/*` to the top like this? In order to distinguish between public packages and internal modules:

```js
'use strict';

const {
	isMediaFeatureRange,
	isMediaFeaturePlain,
	isMediaFeature,
	MediaFeatureRangeNameValue,
	MediaFeatureName,
} = require('@csstools/media-query-list-parser');
const { TokenNode } = require('@csstools/css-parser-algorithms');
const { TokenType } = require('@csstools/css-tokenizer');

const atRuleParamIndex = require('../../utils/atRuleParamIndex');
const { rangeTypeMediaFeatureNames } = require('../../reference/mediaFeatures');
// ...
```

lib/rules/media-feature-range-notation/index.js
```diff
@@ -33,57 +40,101 @@ const rule = (primary) => {
...
+						/** @type {import('@csstools/css-tokenizer').TokenDelim} */
+						let operator = [TokenType.Delim, '<', -1, -1, { value: '<' }];
+
+						if (/^min-/i.test(featureName)) {
+							operator = [TokenType.Delim, '>', -1, -1, { value: '>' }];
+						}
+
+						const parent = entry.parent;
+
+						if (!isMediaFeature(parent)) return;
```
ybiquitous, line 76: [suggest]

- Move `isMediaFeature()` before creating a `TokenDelim` object to avoid extra object creation
- Initialize the `operator` variable with `if ... else` block for readability

```suggestion
						const parent = entry.parent;

						if (!isMediaFeature(parent)) return;

						/** @type {import('@csstools/css-tokenizer').TokenDelim} */
						const operator = /^min-/i.test(featureName)
							? [TokenType.Delim, '>', -1, -1, { value: '>' }]
							: [TokenType.Delim, '<', -1, -1, { value: '<' }];
```

lib/rules/media-feature-range-notation/index.js
```diff
@@ -33,57 +40,101 @@ const rule = (primary) => {
...
+						index: atRuleIndex + startIndex - 1,
+						endIndex: atRuleIndex + endIndex + 1 + 1,
+						result,
+						ruleName,
+					});
+				});
+			});
 
-				if (primary === 'context' && isRangeContextMediaFeature(value)) return;
+			if (context.fix && hasFixes) {
```
ybiquitous, line 125: [trivial] `context.fix` here seems duplicate and needless, but it's trivial so you can also leave it as-is.

```suggestion
			if (hasFixes) {
```

lib/rules/media-feature-range-notation/index.js
ybiquitous: Surely, this seems too elaborate, but we can accept it as-is.

ybiquitous (commented):
@romainmenke Thanks for the pull request. Great work! 👏🏼 

I've left a few suggestions, but this change has no big problems. 👍🏼

lib/rules/media-feature-range-notation/index.js
romainmenke: Done :)

romainmenke:
Thank you for the review and feedback @ybiquitous 🙇 
I've applied all the suggestions.

ybiquitous (approved):
Wow, thanks for your quick fix! LGTM 👍🏼

yoyo837:
Unable to parse when including variable:

```less
@media screen and(max-width <= @screen-sm-max) {
  .topbox {
    // code
  }
}
```

ybiquitous:
@yoyo837 Thanks for the report. Can you open a new issue, please?

yoyo837:
> @yoyo837 Thanks for the report. Can you open a new issue with reproducible info, please?

Yes, here it is #6754.

PR #6760, opened 2023-04-03
Fix media-feature-range-notation parse error

### LINKED ISSUE #6754: [15.4.0] Can not parse in some case

Unable to parse when :

less file:
```less
@media screen and(width <= @screen-sm-max) {
  .topbox {
    // code
  }
}
```

Need to add spaces manually:

```less
@media screen and (width <= @screen-sm-max) {
  .topbox {
    // code
  }
}
```

```js
module.exports = {
  extends: ['stylelint-config-standard', 'stylelint-prettier/recommended'],
  overrides: [
    {
      files: ['**/*.less'],
      customSyntax: 'postcss-less',
    },
  ],
  rules: {
    'selector-pseudo-class-no-unknown': [
      true,
      {
        ignorePseudoClasses: ['global'],
      },
    ],
    'alpha-value-notation': null,
    'font-family-no-missing-generic-family-keyword': null,
    'declaration-empty-line-before': null,
    'selector-class-pattern': null,
    'color-function-notation': null,
    'import-notation': null,
    'length-zero-no-unit': null,
    'function-no-unknown': [
      true,
      {
        ignoreFunctions: ['env', 'constant'],
      },
    ],
  },
};

```

`.stylelintrc.js`

`15.4.0`

Upgrade `stylelint@15.3.0` to `15.4.0`, Lint should works.

### What actually happened?

image

No

### Proposal to fix the bug

Compatible with previous version.

fpetrakov:
Closes #6754

No, it's self-explanatory

### DIFF
--- a/lib/rules/media-feature-name-unit-allowed-list/index.js
+++ b/lib/rules/media-feature-name-unit-allowed-list/index.js
@@ -44,7 +44,7 @@ const rule = (primary) => {
 		};
 
 		root.walkAtRules(/^media$/i, (atRule) => {
-			const mediaQueryList = parseMediaQuery(atRule, result);
+			const mediaQueryList = parseMediaQuery(atRule);
 
 			mediaQueryList.forEach((mediaQuery) => {
 				mediaQuery.walk((entry) => {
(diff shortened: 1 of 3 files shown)


romainmenke (commented):
Thank you for this change @fpetrakov 🙇 
One thought but the change looks good!

romainmenke (approved):
Change looks good, thank you so much for this 🙇 

I am not yet fully up to speed with all the conventions and best practices, so maybe @ybiquitous can best take another look?

ybiquitous (approved):

PR #6764, opened 2023-04-04
Add raw regex support to `ignoreProperties` for `declaration-block-no-duplicate-properties`

ybiquitous:

None.

The `ignoreProperties` option of other rules also supports raw regex.


mattxwang (approved):
LGTM!

PR #6767, opened 2023-04-09
Fix `custom-property-empty-line-before` false positives for CSS-in-JS

### LINKED ISSUE #6766: Using the "except" option with "first-nested" does not work for "custom-property-empty-line-before" when using `postcss-styled-syntax`

When using `["always", { "except": ["first-nested"] }]` as the option for `custom-property-empty-line-before`, `stylelint` is still demanding an empty line before the very first css variable defined in a rule:

```typescript
const Form = styled.form`
  --max-width: 58rem;
`
```

```json
{
  "extends": [
    "stylelint-config-standard"
  ],
  "rules": {
    "custom-property-empty-line-before": ["always", {
      "except": ["first-nested"]
    }]
  }
}
```

`yarn stylelint '**/*.{ts,tsx}'`

`15.4.0`

I should not have to add a newline before the variable definition, prettier actually automatically removes this line which makes it impossible to add the newline. 

### What actually happened?

The only way for me to work around it is to add a `/* stylelint-disable-next-line */` as the first line since prettier would always reformat the code to remove the line.

I'm using `postcss-styled-syntax` and I don't know if the bug manifests without it

### Proposal to fix the bug

Make the configuration work as documented.

ybiquitous:

Closes #6766

This change fixes the `isFirstNested()` utility that the `custom-property-empty-line-before` rule internally uses.

~~The rule's test does not change in order to avoid adding an extra dependency for CSS-in-JS.~~

See also the PostCSS API for the `Document` node: <https://postcss.org/api/#document>

### DIFF
--- a/lib/__tests__/fixtures/postcss-naive-css-in-js.js
+++ b/lib/__tests__/fixtures/postcss-naive-css-in-js.js
@@ -1,0 +1,34 @@
+const postcss = require('postcss');
+
+/**
+ * @type {postcss.Parser<postcss.Document>}
+ */
+function parse(css) {
+	const source = typeof css === 'string' ? css : css.toString();
+
+	const document = postcss.document({
+		source: {
+			input: new postcss.Input(source),
+		},
+	});
+
+	// E.g. "css` color: red; `;"
+	for (const match of source.matchAll(/\bcss`([^`]+)`;/g)) {
+		document.append(postcss.parse(match[1]));
+	}
+
+	return document;
+}
+
+/**
+ * @type {postcss.Stringifier}
+ */
+function stringify(node, builder) {
+	if (node.type === 'document') {
+		node.each((root) => {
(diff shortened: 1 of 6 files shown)


lib/__tests__/fixtures/postcss-naive-css-in-js.js
ybiquitous, line 1: [note] https://postcss.org/docs/how-to-write-custom-syntax

ybiquitous:
**Note**: Probably, this PR should be applied to not only `postcss-styled-syntax` but also other CSS-in-JS syntaxes supporting template literals.

@hudochenkov I would appreciate it if you could let me any concerns about PostCSS's CSS-in-JS syntaxes.

hudochenkov (approved):
Looks good!

ybiquitous:
Thanks for the review!

PR #6772, opened 2023-04-10
Add `ignore: ["consecutive-duplicates-with-different-syntaxes"]` to `declaration-block-no-duplicate-properties`

### LINKED ISSUE #6632: Add `ignore: ["consecutive-duplicates-with-different-syntaxes"]` to `declaration-block-no-duplicate-properties`

https://stylelint.io/user-guide/rules/declaration-block-no-duplicate-properties/#ignore-consecutive-duplicates-with-different-values

I would like to have errors on duplicate declarations with different values only if the different values have the same syntax.

A different syntax is most likely a fallback/progressive enhancement.

Error :

```css
.foo {
  width: 100px;
  width: 150px;
}
```

```css
.foo {
  width: calc(10px * 2);
  width: calc(20px * 2);
}
```

Ok :

```css
.foo {
  width: 100px;
  width: 10ch;
}
```

```css
.foo {
  width: calc(10px * 2);
  width: 2vi;
}
```

It is possible to analyse the value by tokenizing and checking if the token streams of both are the same.

When comparing the token streams the actual values must be ignored and only the type and unit must be considered.

If the token streams differ the syntax is not the same.

kimulaco:

Closes #6632

No, it's self-explanatory.

### DIFF
--- a/lib/rules/declaration-block-no-duplicate-properties/index.js
+++ b/lib/rules/declaration-block-no-duplicate-properties/index.js
@@ -47,7 +47,10 @@ const isEqualValueNodes = (nodes1, nodes2) => {
 		}
 
 		if (isFunctionNode(node1) && isFunctionNode(node2)) {
-			if (isEqualValueNodes(node1.children.toArray(), node2.children.toArray())) {
+			if (
+				node1.name.toLowerCase() === node2.name.toLowerCase() &&
+				isEqualValueNodes(node1.children.toArray(), node2.children.toArray())
+			) {
 				continue;
 			} else {
 				return false;
--- a/lib/rules/declaration-block-no-duplicate-properties/index.js
+++ b/lib/rules/declaration-block-no-duplicate-properties/index.js
@@ -49,11 +49,11 @@ const isEqualValueSyntaxes = (value1, value2) => {
 	const value1Node = parse(value1, { context: 'value' });
 	const value2Node = parse(value2, { context: 'value' });
 
-	if (!isValueNode(value1Node) || !isValueNode(value2Node)) {
-		return false;
-	}
-
-	if (value1Node.children.size !== value2Node.children.size) {
+	if (
+		!isValueNode(value1Node) ||
+		!isValueNode(value2Node) ||
+		value1Node.children.size !== value2Node.children.size
+	) {
 		return false;
 	}
 
--- a/lib/rules/declaration-block-no-duplicate-properties/index.js
+++ b/lib/rules/declaration-block-no-duplicate-properties/index.js
@@ -27,16 +27,42 @@ const isValueNode = (node) => {
 	return node.type === 'Value';
 };
 
-/** @type {(node1: import('css-tree').CssNode, node2: import('css-tree').CssNode) => boolean} */
-const isEqualValueChildNode = (node1, node2) => {
-	if (node1.type !== node2.type) {
+/** @type {(node: import('css-tree').CssNode) => node is import('css-tree').FunctionNode} */
+const isFunctionNode = (node) => {
+	return node.type === 'Function';
+};
+
+/** @type {(node1: import('css-tree').CssNode[], node2: import('css-tree').CssNode[]) => boolean} */
+const isEqualValueNodes = (nodes1, nodes2) => {
+	if (nodes1.length !== nodes2.length) {
 		return false;
 	}
 
-	const node1Unit = 'unit' in node1 ? node1.unit : '';
-	const node2Unit = 'unit' in node2 ? node2.unit : '';
+	for (let i = 0; i < nodes1.length; i++) {
+		const node1 = nodes1[i];
+		const node2 = nodes2[i];
 
-	return node1Unit === node2Unit;
+		if (typeof node1 === 'undefined' || typeof node2 === 'undefined' || node1.type !== node2.type) {
+			return false;
+		}
+
+		if (isFunctionNode(node1) && isFunctionNode(node2)) {
+			if (isEqualValueNodes(node1.children.toArray(), node2.children.toArray())) {
+				continue;
+			} else {
+				return false;
+			}
+		}
+
+		const node1Unit = 'unit' in node1 ? node1.unit : '';
+		const node2Unit = 'unit' in node2 ? node2.unit : '';
+
+		if (node1Unit !== node2Unit) {
+			return false;
+		}
+	}
+
+	return true;
 };
 
 /** @type {(value1: string, value2: string) => boolean} */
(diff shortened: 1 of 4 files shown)


lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -21,6 +22,60 @@ const meta = {
...
+	return node1Unit === node2Unit;
+};
+
+/** @type {(value1: string, value2: string) => boolean} */
+const isEqualValueSyntaxes = (value1, value2) => {
+	if (value1 === value2) {
+		return true;
+	}
+
+	/** @type {import('css-tree').CssNode} */
```
ybiquitous, line 48: [trivial] We might be able to omit `@type` thanks to type inference.

```suggestion
```

ybiquitous (approved):
Thank you, LGTM 👍🏼 

I commented on a trivial suggestion, but you can also leave it as-is.

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -21,6 +22,60 @@ const meta = {
...
+			typeof node1 === 'undefined' ||
+			typeof node2 === 'undefined' ||
+			!isEqualValueChildNode(node1, node2)
+		) {
+			return false;
+		}
+	}
+
+	return true;
+};
```
romainmenke, line 77: Thank you for this change @kimulaco 🙇 

-------

I did a few tests and this doesn't seem to handle functions. Possible because it parses with css tree into a full value AST while the function to detect similar syntaxes doesn't have recursion?

```css
p { width: calc(10px + 2px); width: calc(10rem + 2rem); }
```

-----

Alternatively you could look purely at the token stream :

```js
const { tokenize, TokenType } = require('@csstools/css-tokenizer');

/** @type {(value1: string, value2: string) => boolean} */
const isEqualValueSyntaxes = (value1, value2) => {
	if (value1 === value2) {
		return true;
	}

	const value1Tokens = tokenize({ css: value1 }).filter((x) => x[0] !== TokenType.Whitespace && x[0] !== TokenType.Comment);
	const value2Tokens = tokenize({ css: value2 }).filter((x) => x[0] !== TokenType.Whitespace && x[0] !== TokenType.Comment);

	if (value1Tokens.length !== value2Tokens.length) return false;

	for (let i = 0; i < value1Tokens.length; i++) {
		const token1 = value1Tokens[i];
		const token2 = value2Tokens[i];

		if (!token1 || !token2) {
			return false;
		}

		if (token1[0] !== token2[0]) {
			// Different token types
			return false;
		}

		if (
			token1[0] === TokenType.Dimension &&
			token2[0] === TokenType.Dimension &&
			token1[4].unit.toLowerCase() !== token2[4].unit.toLowerCase()
		) {
			// Different units
			return false;
		}

		if (
			token1[0] === TokenType.Function &&
			token2[0] === TokenType.Function &&
			token1[4].value.toLowerCase() !== token2[4].value.toLowerCase()
		) {
			// Different functions
			return false;
		}
	}

	return true;
};
```

ybiquitous: @romainmenke Nice catch!

kimulaco: @romainmenke  Thanks! I forgot the function use case.
I will try to fix it to support the function with reference to your suggestion!

kimulaco: @romainmenke @ybiquitous Fixed to support function node. https://github.com/stylelint/stylelint/pull/6772/commits/047218e51c6baa146e9aa22e14c3efdca75f341c
When function, check children recursively.

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -21,6 +22,69 @@ const meta = {
...
+
+/** @type {(node: import('css-tree').CssNode) => node is import('css-tree').FunctionNode} */
+const isFunctionNode = (node) => {
+	return node.type === 'Function';
+};
+
+/** @type {(node1: import('css-tree').CssNode[], node2: import('css-tree').CssNode[]) => boolean} */
+const isEqualValueNodes = (nodes1, nodes2) => {
+	if (nodes1.length !== nodes2.length) {
+		return false;
```
ybiquitous, line 38: [ask] Line 38 was not covered by tests, but could you add a test case easily to cover the line?

lib/rules/declaration-block-no-duplicate-properties/index.js
romainmenke: Awesome! thank you @kimulaco

romainmenke (commented):
Thank you for the latest changes @kimulaco
I tested a few more things and spotted two more additions that would make this better :)

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -21,6 +22,69 @@ const meta = {
...
+
+	for (let i = 0; i < nodes1.length; i++) {
+		const node1 = nodes1[i];
+		const node2 = nodes2[i];
+
+		if (typeof node1 === 'undefined' || typeof node2 === 'undefined' || node1.type !== node2.type) {
+			return false;
+		}
+
+		if (isFunctionNode(node1) && isFunctionNode(node2)) {
```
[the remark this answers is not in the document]
kimulaco: @romainmenke Thanks. Fixed to check function name. https://github.com/stylelint/stylelint/pull/6772/commits/f0b45b64d6ecf5809c714193c75d1b8babb85185

romainmenke (approved):
LGTM 🎉

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -21,6 +22,66 @@ const meta = {
 	fixable: true,
 };
 
+/** @type {(node: import('css-tree').CssNode) => node is import('css-tree').CssNode & { children: List<import('css-tree').CssNode> }} */
+const hasChildren = (node) => 'children' in node && node.children instanceof List;
+
+/** @type {(node1: import('css-tree').CssNode[], node2: import('css-tree').CssNode[]) => boolean} */
```
ybiquitous, line 28: [suggest] For readability, how about adding a type alias like this?

```suggestion
/** @typedef {import('css-tree').CssNode} CssNode */

/** @type {(node: CssNode) => node is CssNode & { children: List<CssNode> }} */
const hasChildren = (node) => 'children' in node && node.children instanceof List;

/** @type {(node1: CssNode[], node2: CssNode[]) => boolean} */
```

kimulaco: Thanks! I didn't know `@typedef`.
Fixed to use `@typedef`. https://github.com/stylelint/stylelint/pull/6772/commits/b81cc2335b4e932c73a11ffde9f040b553e403e6

ybiquitous (approved):
Thank you! LGTM 👍🏼

PR #6773, opened 2023-04-10
Add `ignoreProperties: []` to `declaration-block-no-duplicate-custom-properties`

### LINKED ISSUE #6763: Add ignoreProperties to declaration-block-no-duplicate-custom-properties

Hitting errors on the rule due to my pattern of using customizable effects in my classes that duplicate properties. I know they are duplicated and am fine with it.

```tsx
[view-transition-new='screen'] {
  ${RelativeMoveEffect.apply({ endX: 0, startX: '200px', duration: 500 })};
  ${FadeEffect.apply({ startOpacity: 0, endOpacity: 1, delay: 50, duration: 350 })};
  animation: ${FadeEffect.animation}, ${RelativeMoveEffect.animation};
}
```

Each of these "effects" apply defaults, say a default `--view-transition_mix-blend-mode`. They can each also be used individually. In my case I'm totally fine with the last one "winning" and want to allow this exception.

Add an `ignoreProperties` syntax like in other rules to let me prefix my exceptions.

```
ignoreProperties: ['/--view-transition']
```

mattxwang:

Closes #6763.

No, it's self-explanatory.

### DIFF
--- a/.changeset/swift-falcons-battle.md
+++ b/.changeset/swift-falcons-battle.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: `ignoreProperties: []` to `declaration-block-no-duplicate-custom-properties`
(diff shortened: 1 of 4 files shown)


ybiquitous (approved):
Thanks, LGTM! 👍🏼

PR #6777, opened 2023-04-12
Fix `declaration-block-no-redundant-longhand-properties` autofix for `grid-template`

### LINKED ISSUE #6701: Fix `declaration-block-no-redundant-longhand-properties` autofix for `grid-template`

Broken autofix for `declaration-block-no-redundant-longhand-properties` with `grid-template`

I have the next CSS
DEMO: https://codepen.io/shoonia/pen/qBMVPXr
```css
.page {
  display: grid;
  height: 100vh;
  grid-template-rows: var(--header-h) 1fr var(--footer-h);
  grid-template-columns: var(--toolbar-w) 1fr;
  grid-template-areas: "header header" "toolbar main" "footer footer";
}
```

After using autofix with `--fix` , it broken my layout
DEMO: https://codepen.io/shoonia/pen/GRXOMMG
```diff
.page {
  display: grid;
  height: 100vh;
-  grid-template-rows: var(--header-h) 1fr var(--footer-h);
-  grid-template-columns: var(--toolbar-w) 1fr;
-  grid-template-areas: "header header" "toolbar main" "footer footer";
+ grid-template: var(--toolbar-w) 1fr var(--header-h) 1fr var(--footer-h) "header header" "toolbar main" "footer footer";
}
```

```json
{
  "extends": [
    "stylelint-config-standard",
    "stylelint-config-css-modules"
  ],
  "rules": {
    "selector-class-pattern": null,
    "selector-pseudo-element-no-unknown": [
      true,
      {
        "ignorePseudoElements": [
          "range-track",
          "range-thumb"
        ]
      }
    ]
  }
}

```

`npx stylelint src/**/*.css --fix`

`15.2.0`

it shouldn't be minimized.

### What actually happened?

```diff
-  grid-template-rows: var(--header-h) 1fr var(--footer-h);
-  grid-template-columns: var(--toolbar-w) 1fr;
-  grid-template-areas: "header header" "toolbar main" "footer footer";
+ grid-template: var(--toolbar-w) 1fr var(--header-h) 1fr var(--footer-h) "header header" "toolbar main" "footer footer";
```

NO

### Proposal to fix the bug

Looks like it shouldn't be minimize. I don't know a short syntax for this construction.

mattxwang:

Closes #6701.

A handful of thoughts:

- should I be moving this "custom resolver" logic to a new file?
- I've elected to *not* implement the same fix for the `grid` shorthand, since the situation there is quite a bit more complicated with the `auto-flow` syntax element ([spec ref](https://www.w3.org/TR/css-grid-2/#grid-shorthand))

More broadly, I'm happy to own any other changes that need to be made to this autofixer.


mattxwang:
After thinking on it a bit more, I'm not sure what the other alternative is to resolve this problem (other than a "custom resolver" methodology). Going to mark this as ready to review, but I'm anticipating that it'll need some changes!

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -113,21 +154,24 @@ const rule = (primary, secondaryOptions, context) => {
...
+								prefixedShorthandProperty,
+								prefixedShorthandData,
+								transformedDeclarationNodes,
+							);
 
-							const newShorthandDeclarationNode = firstDeclNode.clone({
-								prop: prefixedShorthandProperty,
-								value: resolvedShorthandValue,
-							});
+							if (resolvedShorthandValue !== null) {
```
ybiquitous, line 163: [suggest] This is just my preference, simply checking just a variable seems a bit more readable:

```suggestion
							if (resolvedShorthandValue) {
```

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -22,8 +22,49 @@ const meta = {
...
 
+const customResolvers = new Map([
+	[
+		'grid-template',
+		(/** @type {Map<string, Declaration>} */ transformedDeclarationNodes) => {
+			const areas = transformedDeclarationNodes.get('grid-template-areas')?.value.trim();
+			const columns = transformedDeclarationNodes.get('grid-template-columns')?.value.trim();
+			const rows = transformedDeclarationNodes.get('grid-template-rows')?.value.trim();
+
+			if (areas === undefined || columns === undefined || rows === undefined) return;
```
ybiquitous, line 33: [suggest] Excluding also `''` seems safer because of using `.trim()` like this:

```suggestion
			if (!(areas && columns && rows)) return;
```

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -22,8 +22,49 @@ const meta = {
...
+
+			const splitAreas = [...areas.matchAll(/".+?"/g)].map((x) => x[0]);
+			const splitRows = rows.split(' ');
+
+			if (splitAreas.length !== splitRows.length) return;
```
ybiquitous, line 38: [suggest] Just in case, how about checking if the arrays are empty?

```suggestion
			if (splitAreas.length === 0 || splitRows.length === 0) return;

			if (splitAreas.length !== splitRows.length) return;
```

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -22,8 +22,49 @@ const meta = {
...
+	[
+		'grid-template',
+		(/** @type {Map<string, Declaration>} */ transformedDeclarationNodes) => {
+			const areas = transformedDeclarationNodes.get('grid-template-areas')?.value.trim();
+			const columns = transformedDeclarationNodes.get('grid-template-columns')?.value.trim();
+			const rows = transformedDeclarationNodes.get('grid-template-rows')?.value.trim();
+
+			if (areas === undefined || columns === undefined || rows === undefined) return;
+
+			const splitAreas = [...areas.matchAll(/".+?"/g)].map((x) => x[0]);
```
ybiquitous, line 35: [suggest] How about changing to a bit more accurate regex?

```suggestion
			const splitAreas = [...areas.matchAll(/"[^"]+?"/g)].map((x) => x[0]);
```

ybiquitous (changes requested):
@mattxwang Thanks for the pull request. I've commented on a few refactoring ideas, so can you check them?

mattxwang:
Thanks for the suggestions! I think I've incorporated all of them (in a slightly-messy git history 😅 )

ybiquitous (approved):
Thanks, LGTM 👍🏼

Issue #6402, opened 2022-10-11
Add `cache` to the configuration object


noahtallen:
### What steps are needed to reproduce the bug?

Add `"cache": true` to a `.stylelintrc` file and then run stylelint. Notice that a cache is not created. (Also note that a cache is created when running with the `--cache` flag.)

### What Stylelint configuration is needed to reproduce the bug?

```json
{
  "cache": true,
  "extends": "stylelint-config-standard-scss"
}
```

### How did you run Stylelint?

`yarn run stylelint "**/*.{css,scss}"`

### Which version of Stylelint are you using?

14.13.0

### What did you expect to happen?

The cache should be created and used.

### What actually happened?

The cache file was not created. (Note that a cache file _is_ created when adding `--cache` as a CLI flag -- just not when it's in the config.)

### Does the bug relate to non-standard syntax?

No

#### Root cause:
The root cause is likely due to how stylelint resolves its config file per-linted file rather than at a global level. (Related discussion here: https://github.com/stylelint/stylelint/issues/6383) As a result, when any global-level config option (such as `cache`) is used before the config is read per-linted file, the config file isn't able to set the option. 

For example, in the standalone linter, stylelint sets `cache` to false by default here:
https://github.com/stylelint/stylelint/blob/168b71a0faf53324c04c97cecd9409b903fca57d/lib/standalone.js#L37

Then, when we go to lint the file, that option is passed in here:

https://github.com/stylelint/stylelint/blob/168b71a0faf53324c04c97cecd9409b903fca57d/lib/standalone.js#L214-L217

In the `lintSource` function, we see that while the config is read in from .stylelintrc, that config is not used for the `cache` option, which was passed through as an argument instead:

https://github.com/stylelint/stylelint/blob/168b71a0faf53324c04c97cecd9409b903fca57d/lib/lintSource.js#L66-L79

#### Config resolution:
As a result, I think this is a deeper issue with configuration resolution, where we have some config options which are meant to be used "globally," like cache, but are only ever read from the CLI and not from the config file. This behavior is unfortunately not documented super well, as an option like this https://stylelint.io/user-guide/usage/options/#cache reads as if it can be used in the config file.

#### Why?
The reason I'd like this feature is so an IDE-level linter (e.g. https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint) could also use the cache if that would improve performance for some devs.

#### Resolution
[...]

PR #6778, opened 2023-04-12
Add `allowEmptyInput`, `cache`, `fix` options to configuration object

mattxwang:

Part of (but does not close) #5142. This PR handles all the `boolean` flags (which should be relatively simple).

Sorry about the CodeQL alerts - it was an approach that I tried (and chose to abandon) that I forgot to remove 😅 

As well as my note below on `cache`, two quick questions:

1. For the configuration docs, what orders should the options go in? I may have missed a pattern - I was confused since it's not in the same order as the flags.
2. Any tests I should add?

---

Note that I didn't `augmentConfig` for `allowEmptyInput` and `cache`, since we never access them through the `config` object outside of `standalone.js`; including them drops code coverage. That being said, maybe there's a cleaner way to use these options "downstream", and be consistent in augmenting the config?

### DIFF
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -473,6 +473,8 @@ For example:
 }
 ```
 
+Note: this config option should not be overridden on a per-file basis.
+
 [More info](options.md#cache).
 
 ## `fix`
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -459,6 +459,8 @@ For example:
 }
 ```
 
+Note: this config option should not be overridden on a per-file basis.
+
 [More info](options.md#allowemptyinput).
 
 ## `cache`
@@ -488,5 +490,7 @@ For example:
   "fix": true
 }
 ```
+
+Note: this config option should not be overridden on a per-file basis.
 
 [More info](options.md#fix).
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -446,3 +446,45 @@ If the globs are absolute paths, they are used as is. If they are relative, they
 - or `process.cwd()`.
 
 _Note that this is not an efficient method for ignoring lots of files._ If you want to ignore a lot of files efficiently, use [`.stylelintignore`](ignore-code.md) or adjust your files globs.
+
+## `allowEmptyInput`
+
+Stylelint does not throw an error when the glob pattern matches no files.
+
+For example:
+
+```json
+{
+  "allowEmptyInput": true
+}
+```
+
+[More info](options.md#allowemptyinput).
+
+## `cache`
+
+Store the results of processed files so that Stylelint only operates on the changed ones.
+
+For example:
+
+```json
+{
+  "cache": true
+}
+```
+
+[More info](options.md#cache).
+
+## `fix`
+
+Automatically fix, where possible, problems reported by rules.
+
+For example:
+
+```json
+{
+  "fix": true
+}
+```
+
+[More info](options.md#fix).
--- a/lib/__tests__/standalone-cache.test.js
+++ b/lib/__tests__/standalone-cache.test.js
@@ -34,6 +34,18 @@ function getConfig(additional = {}) {
 	};
 }
 
+// equivalent function, but cache is enabled as a config option
+function getConfigWithCache(additional = {}) {
+	return {
+		files: replaceBackslashes(path.join(fixturesPath, 'cache', '*.css')),
+		config: {
+			cache: true,
+			rules: { 'block-no-empty': true, 'color-no-invalid-hex': true },
+		},
+		...additional,
+	};
+}
+
 describe('standalone cache', () => {
 	const cwd = path.join(__dirname, 'tmp', 'standalone-cache');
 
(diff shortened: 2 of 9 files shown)


mattxwang:
In #6402, it sounds like `cache` may be more complicated than it first seems. I feel like in this case, I must be missing a test case that addresses the complication in the issue (about per-file config resolution) - is that correct?

mattxwang:
Thanks for the quick response @ybiquitous, I think you bring up some really great points!

> ### Should CLI flags win config files?
> 
> For example, assume we have the following config file named `.stylelintrc.json`:
> 
> ```json
> { "cache": true }
> ```
> 
> Then, should we expect the following? Note that `--no-` prefix is supported by [`meow`](https://github.com/sindresorhus/meow).
> 
> ```shell
> # Cache is enabled via config file
> stylelint *.css
> 
> # Cache is enabled via CLI flag
> stylelint *.css --cache
> 
> # Cache is disabled via CLI flag
> stylelint *.css --no-cache
> ```
> 
> My answer to the question above is **yes**. Because people would like to turn on/off caching without editing the config file.

I agree with you here (in other words, the CLI flag takes precedence over the config file)! Just implemented the change + have added tests for each of those three cases.

> ### Should we support per-file configs for properties like `cache`?
> 
> I don't believe such a per-file overriding for global settings makes sense. Lint performance and code maintainability would degrade. Even usability might decrease due to config files everywhere.
> 
> So, I recommend supporting only properties in the root config file. We can just ignore properties in sub-folder configs. Of course, we should clearly state the restriction in the document.

This makes sense to me as well. Do you think there are other config options that fall into this category? The rest of them feel like they could be enabled/disabled per-file.

If the user has set `cache` in a subdirectory - should we error, print a message saying that we're ignoring, proceed quietly, or do something else? If this is an exception rather than a rule, I think informing users may be a good idea.

ybiquitous:
> Do you think there are other config options that fall into this category? The rest of them feel like they could be enabled/disabled per-file.

Basically, we should prevent per-file config props from increasing in the future (we have now `overrides`). I'm not sure which config props support per-file configs, though. Anyway, not touching the existing props might be safe not to break users' linting.

> If the user has set `cache` in a subdirectory - should we error, print a message saying that we're ignoring, proceed quietly, or do something else? If this is an exception rather than a rule, I think informing users may be a good idea.

I believe erroring is best, but if hard to implement it, just documenting the limitation might be enough.

lib/__tests__/standalone-cache.test.js
```diff
@@ -223,6 +223,91 @@ describe('standalone cache uses cacheLocation', () => {
...
+		await standalone(getConfigWithCache());
+	});
+
+	afterEach(async () => {
+		// Clean up after each test case
+		await removeFile(expectedCacheFilePath);
+		await removeFile(newFileDest);
+	});
+
+	it('uses cache when no CLI argument is provided', async () => {
```
ybiquitous, line 255: [suggest] I think relating `standalone()` with CLI may be confusing since `standalone()` is called not only CLI but also Node.js programs. So, can you rewrite the descriptions around CLI or `--cache`?

lib/standalone.js
```diff
@@ -134,7 +134,7 @@ async function standalone({
 		const postcssResult = stylelintResult._postcssResult;
 		const returnValue = prepareReturnValue([stylelintResult], maxWarnings, formatterFunction, cwd);
 
-		if (fix && postcssResult && !postcssResult.stylelint.ignored) {
+		if ((fix || config?.fix) && postcssResult && !postcssResult.stylelint.ignored) {
```
ybiquitous, line 137: [suggest] I think the [`??`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) operator can clear the code, like this:

```suggestion
		const autofix = fix ?? config?.fix ?? false;

		if (autofix && postcssResult && !postcssResult.stylelint.ignored) {
```

Note: We can use this operator since Node.js 14 supports it.

lib/standalone.js
```diff
@@ -164,6 +164,9 @@ async function standalone({
 		fileList = fileList.concat(ALWAYS_IGNORED_GLOBS.map((glob) => `!${glob}`));
 	}
 
+	// do not cache if config is explicitly overrided by option
+	const useCache = cache || (config?.cache && cache !== false);
```
ybiquitous, line 168: [suggest] I have the same suggestion as above:

```suggestion
	const useCache = cache ?? config?.cache ?? false;
```

lib/standalone.js
```diff
@@ -246,7 +249,7 @@ async function standalone({
 		});
 
 		stylelintResults = await Promise.all(getStylelintResults);
-	} else if (allowEmptyInput) {
+	} else if (allowEmptyInput || config?.allowEmptyInput) {
```
ybiquitous, line 252: [suggest] I have the same suggestion as above:

```suggestion
	} else if (allowEmptyInput ?? config?.allowEmptyInput) {
```

lib/standalone.js
```diff
@@ -30,7 +30,7 @@ const writeFileAtomic = require('write-file-atomic');
  */
 async function standalone({
 	allowEmptyInput = false,
```
ybiquitous, line 32: Oh, tests fail. This change also is needed:

```suggestion
	allowEmptyInput,
```

mattxwang:
Apologies for the delay, got held up with some other work. I think I've addressed most of the code review comments!

To clarify, when you say

> Anyway, not touching the existing props might be safe not to break users' linting.

Do you mean the minimal-invasive change (ex, right now - we follow the "default" behaviour", and I've added a little note in the docs)? Or, do you mean that I should explicitly add a subdirectory check?

ybiquitous:
I think we should add some actions for the new config props in this PR, but I don't think they are needed for the existing props.

docs/user-guide/configure.md
```diff
@@ -446,3 +446,47 @@ If the globs are absolute paths, they are used as is. If they are relative, they
...
+
+For example:
+
+```json
+{
+  "cache": true
+}
+```
+
+Note: this config option should not be overridden on a per-file basis.
```
ybiquitous, line 476: [ask] Can we add this note also to `allowEmptyInput` and `fix`?

ybiquitous (approved):
Thanks, LGTM 👍🏼

ybiquitous:
Let's merge this and include it in the next release. 👍🏼 

I've commented on the reason: https://github.com/stylelint/stylelint/issues/6795#issuecomment-1517018386

PR #6782, opened 2023-04-13
Fix `block-no-empty` false positives with non-whitespace characters

### LINKED ISSUE #6779: Fix `block-no-empty` false positives with the `css` function nested in a styled component

Lint the following styled components code in a `.tsx` file with the provided stylelint configuration and version. Other version numbers:
- `postcss-styled-syntax`: `^0.4.0`
- `stylelint-config-standard`: `^32.0.0`

Failing code snippet:
```tsx
const buttonHover = css`
    color: blue;
    text-decoration: none;
    border-bottom: 3px solid black;
`;

export const Button = styled.button`
    &:hover {
        ${buttonHover}
    }
`;
```

```json
{
    "files": ["*.ts", "*.tsx"],
    "extends": [
        "stylelint-config-standard"
    ],
    "rules": {}
}
```

CLI with `stylelint \"src/**/*.{ts,tsx}\"`

`15.4.0`

The CSS should be valid and pass the `block-no-empty` rule check.

### What actually happened?

`block-no-empty` Returns an error 

```bash
src/shared/components/Button/Button.styled.tsx
 159:41  ✖  Unexpected empty block  block-no-empty
```

Yes, it's related to CSS-in-JS (styled components)

### Proposal to fix the bug

_No response_

ybiquitous:

Closes #6779

This pull request fixes such false positives for not only the standard CSS but also CSS-in-JS like this:

Standard CSS:

```css
a {
  ;
}
```

CSS-in-JS:

```jsx
const Button = styled.button`
  &:hover {
    ${buttonHover}
  }
`;
```

### DIFF
--- a/.changeset/long-trainers-impress.md
+++ b/.changeset/long-trainers-impress.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `block-no-empty` false positives with non-whitespace characters
--- a/lib/rules/block-no-empty/__tests__/index.js
+++ b/lib/rules/block-no-empty/__tests__/index.js
@@ -29,10 +29,19 @@ testRule({
 			code: 'a {\n/*\nfoo\nbar\n*/\n}',
 		},
 		{
+			code: 'a { ; }',
+		},
+		{
 			code: '@media print { a { color: pink; } }',
 		},
 		{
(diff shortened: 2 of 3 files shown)


hudochenkov (approved):
Looks good to me!

ybiquitous:
Thanks for the review!

PR #6783, opened 2023-04-13
Improve CLI help

ybiquitous:

None.

Changes summary:
- Unify wording and format in the help.
- Simplify flag descriptions.
- Add missing flags.
- Refactor type annotations.

For the help format, I referred to the [`git log`](https://www.git-scm.com/docs/git-log) manual.

<details><summary>Screenshot</summary>
<p>

Note: using [`bat`](https://github.com/sharkdp/bat) for coloring.

```shell
bin/stylelint.js --help | bat --no-pager -l cmd-help --style=plain
```

image

</p>
</details>


lib/__tests__/__snapshots__/cli.test.js.snap
```diff
@@ -9,103 +9,105 @@ exports[`CLI --help 1`] = `
...
-      your files changes even if their contents have not. For example, this can happen
-      during git operations like "git clone" because git does not track file modification
-      time.
+      The "content" strategy can be useful in cases where the modification time
+      of your files changes even if their contents have not. For example, this can
+      happen during git operations like "git clone" because git does not track file
+      modification time.
 
-    --formatter, -f               [default: "string"]
+    --formatter, -f <compact|github|json|string|tap|unix|verbose>
```
mattxwang, line 104: Agh! Retrying to post this comment again.

---

A minor question: it is possible that users confuse this argument notation (`<>`, where each item is a literal) opposed to other cases where `<>` describes a type.

For example, the `<...|json|...>` here describes the literal string `"json"`; but, for `--go <json>`, that refers to a set of values that is valid JSON.

(this would also apply to `--cache-strategy`)

The `git-log` docs have a few possible options:

- using braces (`{}`), which feels intuitive like set notation; ref [`--diff-algorithm`](https://www.git-scm.com/docs/git-log#Documentation/git-log.txt---diff-algorithmpatienceminimalhistogrammyers)
- using parens (`()`), which feel intuitive like regex/tuples; ref [`--diff-merges`](https://www.git-scm.com/docs/git-log#Documentation/git-log.txt---diff-mergesoffnoneonfirst-parent1separatemcombinedcdense-combinedccremerger)
- defining a new `<format>` type, which is what they do for their [`--format`](https://www.git-scm.com/docs/git-log#Documentation/git-log.txt---formatltformatgt)

What do you think? I also think leaving it as-is is generally reasonable.

ybiquitous: Thanks, good point! Listing values sounds more understandable like [`--diff-algorithm`](https://www.git-scm.com/docs/git-log#Documentation/git-log.txt---diff-algorithmpatienceminimalhistogrammyers). For example:

```
--formatter, -f <formatter>

  Choose an output formatter. The variants are as follows:
  
    string - human-readable strings (default)
    compact - similar to ESLint's compact formatter 
    github - workflow commands for GitHub Actions
    json - JSON format
    tap - TAP format
    unix - C compiler-like format
    verbose - a file list and a tally, extending the "string" formatter
```

What do you think?

Ref:

https://github.com/stylelint/stylelint/blob/231dac450f345f84966983db5bc6cce3b3e5989a/docs/user-guide/options.md?plain=1#L85-L91

mattxwang: I think that's a great idea! LGTM.

ybiquitous: Thanks, I changed it via d27ea14d.

mattxwang (approved):
I think this is great, nice work! LGTM.

PR #6785, opened 2023-04-16
Remove Node.js 14 support

### LINKED ISSUE #6740: Remove Node.js 14 support

Node.js 14 will [reach EOL on Apr 30, 2023](https://nodejs.dev/en/about/releases/) (in about a month).

The currently supported versions are the following:

https://github.com/stylelint/stylelint/blob/13efb8655c2fc8163596ace2b67635e9bea8e6b9/package.json#L203-L205

We will remove the Node.js 14 support with the next major version.

The supported versions will be:

```json
{
  "engines": {
    "node": "^16.0.0 || >=18.0.0"
  }
}
```

MohitMaheshwari1711:

Closes #6740

The issue mentioned that the minimum version required should be v16.0.0; as per my observations, some packages, such as jest and others need version ^16.14.0. I have attached a screenshot for further clarification. Hence I have added support for v16.14.0 rather than 16.0.0.

Screenshot 2023-04-16 at 7 59 24 PM


ybiquitous:
@MohitMaheshwari1711 Thanks for the pull request. But I think this PR is too much for the reason I commented on https://github.com/stylelint/stylelint/issues/6740#issuecomment-1510520466.

Mouvedia:
I think if it has to be changed it should based on the LTS.
i.e. `>=16.13.0"`

package.json
```diff
@@ -201,6 +201,6 @@
     "typescript": "^5.0.4"
   },
   "engines": {
-    "node": "^14.13.1 || >=16.0.0"
+    "node": "^16.14.0 || >=18.0.0"
```
ybiquitous, line 204: [suggest] I agree with @Mouvedia's suggestion on <https://github.com/stylelint/stylelint/pull/6785#issuecomment-1565834412>. Node.js 16.13.0 is the first version of Node.js 16.x LTS series.

See https://nodejs.org/en/blog/release/v16.13.0

```suggestion
    "node": "^16.13.0 || >=18.0.0"
```

The [`ls-engines`](https://www.npmjs.com/package/ls-engines) check passes:

```console
$ npx ls-engines@latest
`node_modules` found; loading tree from disk...
┌────────┬─────────────────────────────────────────────────────────────────┐
│ engine │ Currently available latest release of each valid major version: │
├────────┼─────────────────────────────────────────────────────────────────┤
│ node   │ v20.3.1, v19.9.0, v18.16.1, v17.9.1, v16.20.1                   │
└────────┴─────────────────────────────────────────────────────────────────┘

┌───────────────────────────────────┬───────────────────────────┐
│ package engines:                  │ dependency graph engines: │
├───────────────────────────────────┼───────────────────────────┤
│ "engines": {                      │ "engines": {              │
│   "node": "^16.13.0 || >= 18.0.0" │   "node": ">= 16.13"      │
│ }                                 │ }                         │
└───────────────────────────────────┴───────────────────────────┘


Your “engines” field allows fewer node versions than your dependency graph does.
...
```

Surely, dev dependencies require higher Node.js versions. But we can ignore them because they're just for *dev*, and we should respect `engines.node` to support more users.

```console
$ npx ls-engines@latest --dev
...
Your “engines” field does not exactly match your dependency graph‘s requirements!
...
```

ybiquitous:
I've changed the base branch from `main` to `v16`. See #6930.

ybiquitous:
See #7020

PR #6792, opened 2023-04-17
Fix `declaration-block-no-duplicate-properties` syntax error

### LINKED ISSUE #6791: Fix `declaration-block-no-duplicate-properties` SyntaxError

lint error:
```scss
@mixin iosSafeAreaBottom($offset: 0px) {
  bottom: $offset;
  bottom: calc(var(--body-padding-offset-bottom) + $offset);
  bottom: calc(var(--body-padding-offset-bottom) + $offset + constant(safe-area-inset-bottom));
  bottom: calc(var(--body-padding-offset-bottom) + $offset + env(safe-area-inset-bottom, 0px));
}
```

```js
module.exports = {
  extends: [
    'stylelint-config-standard',
    'stylelint-config-html/vue',
    'stylelint-config-html',
    'stylelint-config-recommended-vue/scss',
    'stylelint-config-recommended-vue',
    'stylelint-prettier/recommended',
  ],
  overrides: [
    {
      files: ['**/*.vue'],
      customSyntax: 'postcss-html',
    },
    {
      files: ['**/*.scss'],
      customSyntax: 'postcss-scss',
    },
    {
      files: ['**/*.less'],
      customSyntax: 'postcss-less',
    },
  ],
  plugins: ['stylelint-scss'],
  rules: {
    'selector-pseudo-class-no-unknown': [
      true,
      {
        ignorePseudoClasses: ['first'],
      },
    ],
    'font-family-no-missing-generic-family-keyword': null,
    'no-invalid-position-at-import-rule': [
      true,
      {
        ignoreAtRules: ['use', 'tailwind'],
      },
    ],
    'selector-class-pattern': null,
    'at-rule-no-unknown': [
      true,
      {
        ignoreAtRules: [
          'mixin',
          'include',
          'extend',
          'each',
          'use',
          'tailwind',
          'layer',
          'apply',
          'if',
          'else',
          'at-root',
        ],
      },
    ],
    'scss/at-rule-no-unknown': [
      true,
      {
        ignoreAtRules: [
          'mixin',
          'include',
          'extend',
          'each',
          'use',
          'tailwind',
          'layer',
          'apply',
          'if',
          'else',
          'at-root',
        ],
      },
    ],
    'selector-pseudo-element-no-unknown': [
      true,
      {
        ignorePseudoElements: ['v-deep'],
      },
    ],
    'color-function-notation': null,
    'import-notation': null,
    'length-zero-no-unit': null,
    'function-no-unknown': [
      true,
      {
        ignoreFunctions: ['env', 'constant', 'math.div', '-', 'red', 'blue', 'green'],
      },
    ],
    // less 编译 rgb(0 0 0 / 0.03) 会输出 #000000, 但编译 rgba(0, 0, 0, 0.03) 又能正确输出 rgb(0 0 0 / 0.03)
    // 先禁用
    'alpha-value-notation': null,
    'selector-not-notation': null,
    // FIXME: 迁移完毕后修复⬇️
    'scss/at-extend-no-missing-placeholder': null,
    // postcss-preset-env 好像没有默认支持 "context"
    'media-feature-range-notation': 'prefix',
  },
};

```

Cli with `stylelint '**/*.{css,scss}'`

15.5.0

run lint success

### What actually happened?

run lint throw SyntaxError error

scss

### Proposal to fix the bug

_No response_

yoyo837:

close #6791

"No, it's self-explanatory."


lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -76,6 +77,10 @@ const isEqualValueSyntaxes = (value1, value2) => {
 		return true;
 	}
 
+	if (!(isStandardSyntaxValue(value1) && isStandardSyntaxValue(value2))) {
+		return false;
+	}
+
 	const value1Node = parse(value1, { context: 'value' });
 	const value2Node = parse(value2, { context: 'value' });
```
ybiquitous, line 85: [suggest] I rethink it safer to catch a syntax error from `css-tree` than our `isStandardSyntaxValue` util. Can you rewrite it like this, please?

```suggestion
	let value1Node;
	let value2Node;

	try {
		value1Node = parse(value1, { context: 'value' });
		value2Node = parse(value2, { context: 'value' });
	} catch (e) {
		return false;
	}
```

Other rules also do the same for `css-tree`, for example:

https://github.com/stylelint/stylelint/blob/369557b79805ca3daee5d595662ba603f27ba9c2/lib/rules/declaration-property-value-no-unknown/index.js#L93-L97

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -76,6 +77,10 @@ const isEqualValueSyntaxes = (value1, value2) => {
 		return true;
 	}
 
+	if (!(isStandardSyntaxValue(value1) && isStandardSyntaxValue(value2))) {
+		return false;
+	}
```
ybiquitous, line 82: [suggest] Can you add a test case to cover the new code? For example,

```scss
p { width: $a; width: calc(1 + $a); }
```

lib/rules/declaration-block-no-duplicate-properties/index.js
yoyo837: Do you want me to move `containsUnsupportedFunction` to utils?

ybiquitous: No, just catch an error and return `false` from `css-tree`.

ybiquitous: Oh, sorry. I misunderstood. 😓 

Yes, please move it to utils! 🙏🏼

ybiquitous: Um, `containsUnsupportedFunction` is needed in this case?

yoyo837: In my personal opinion, no, but any suggestion is ok.

lib/rules/declaration-block-no-duplicate-properties/index.js
yoyo837: For the current case, I don't know how to write `reject` code in test case.

ybiquitous: Sure. For example:

```diff
--- a/lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
+++ b/lib/rules/declaration-block-no-duplicate-properties/__tests__/index.js
@@ -307,6 +307,7 @@ testRule({
 		{ code: 'p { width: min(10px, 11px); width: max(10px, 11px); }' },
 		{ code: 'p { width: calc((10px + 2px)); width: calc((10rem + 2rem)); }' },
 		{ code: 'p { width: calc((10px + 2px) + 10px); width: calc((10rem + 2rem) + 10px); }' },
+		{ code: 'p { width: $a; width: calc(1 + $a); }', description: 'using SCSS variables' },
 	],
 
 	reject: [
```

lib/rules/declaration-block-no-duplicate-properties/index.js
ybiquitous: Very sorry for my wrong request. 🙇🏼 
Could you revert the `containsUnsupportedFunction` change to minimize this PR diff?

yoyo837: OK, I’ll do that at tomorrow.

lib/rules/declaration-block-no-duplicate-properties/index.js
yoyo837: Done, if it no need to configure `reject`.

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -76,8 +76,17 @@ const isEqualValueSyntaxes = (value1, value2) => {
 		return true;
 	}
 
-	const value1Node = parse(value1, { context: 'value' });
-	const value2Node = parse(value2, { context: 'value' });
+	let value1Node;
+	let value2Node;
+
+	try {
+		// css-tree will throw an exception if !isStandardSyntaxValue(value)
```
ybiquitous, line 83: [suggest] Can you remove this comment because it's inaccurate now?

```suggestion
```

lib/rules/declaration-block-no-duplicate-properties/index.js
ybiquitous: Thanks!

ybiquitous (approved):
Thank you. LGTM 👍🏼

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -76,8 +77,20 @@ const isEqualValueSyntaxes = (value1, value2) => {
...
-	const value1Node = parse(value1, { context: 'value' });
-	const value2Node = parse(value2, { context: 'value' });
+	if (!(isStandardSyntaxValue(value1) && isStandardSyntaxValue(value2))) {
+		return false;
+	}
+
+	let value1Node;
+	let value2Node;
+
+	try {
```
yoyo837, line 87: Looks like more test cases are needed.

ybiquitous: Feel free to add cases. 👌🏼

ybiquitous: The added case is handled at the line of:

```js
if (!(isStandardSyntaxValue(value1) && isStandardSyntaxValue(value2))) {
```

so finding a test case causing a syntax error may be hard. I think the current case is enough, but you may add more cases if possible. 👍🏼

yoyo837: Test case added.

ybiquitous (approved):
Nice addition to the test case! 👍🏼

Issue #6796, opened 2023-04-18
False positives with option `ignore: ["consecutive-duplicates-with-different-syntaxes"]` in `declaration-block-no-duplicate-properties`


romainmenke:
see : https://github.com/stylelint/stylelint/issues/6632#issuecomment-1511226734

### What steps are needed to reproduce the bug?

- two subsequent declarations with the same property
- two different values but having the same AST shape and different keywords

```css
.foo {
  position: fixed;
  position: sticky;
}

.bar {
  color: color(display-p3 0 1 0);
  color: color(xyz 0 1 0);
}
```

### What Stylelint configuration is needed to reproduce the bug?

```js
 	'declaration-block-no-duplicate-properties': [
 		true,
 		{
 			ignore: [
 				'consecutive-duplicates-with-different-syntaxes',
 			],
 		},
 	],
```

### How did you run Stylelint?

js API

### Which version of Stylelint are you using?

latest

### What did you expect to happen?

I expected fewer false positives form this new option.
Some false positives are ok.

### What actually happened?

The most common false positives were two different keywords.
These are currently seen as the same syntax while they are most often fallbacks.

```css
.foo {
  width: auto;
  width: max-content; /* max-content is fairly recent feature */
}
```

### Does the bug relate to non-standard syntax?

No

### Proposal to fix the bug

I think there are multiple ways forward:

- extend the current algorithm to allow keywords with different names
- use two mappings to ignore subsequent keywords, one with common initial values (`auto`) and the css wide keywords (`inherit`), the other with common fallbacks (`fixed`)
- switch to a different algorithm that focusses more on detecting unique keywords, function names, units, delims, ... (see: https://github.com/stylelint/stylelint/issues/6632#issuecomment-1511761204) 

--------

Personally I think it would be best if we reduce the number of false positives as much as possible and make the option smarter/better over time.

The least false positives:
- distinct node types indicate different syntaxes
- distinct keywords indicate different syntaxes
- distinct function names indicate different syntaxes
- different number of child nodes indicate different syntaxes
- distinct units in dimensions indicate different syntaxes

Compared to direct comparison between values as strings this still catches a few types of errors :
- same value but one has different whitespace/comment tokens
- `10px` followed by `11px`

ybiquitous:
@romainmenke Thanks for the report. By the way, is `consecutive-duplicates-with-different-values` in the config example you reported a mistake for `consecutive-duplicates-with-different-syntaxes`, right?

[...]

PR #6797, opened 2023-04-18
Fix `declaration-block-no-duplicate-properties` false positives with option `consecutive-duplicates-with-different-syntaxes`

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/6796

### DIFF
--- a/.changeset/many-toes-smell.md
+++ b/.changeset/many-toes-smell.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed false positives with option `ignore: ["consecutive-duplicates-with-different-syntaxes"]` in `declaration-block-no-duplicate-properties`
--- a/.changeset/many-toes-smell.md
+++ b/.changeset/many-toes-smell.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed false positives with option `ignore: ["consecutive-duplicates-with-different-syntaxes"]` in `declaration-block-no-duplicate-properties`
+Fixed: `declaration-block-no-duplicate-properties` false positives with option `ignore: ["consecutive-duplicates-with-different-syntaxes"]`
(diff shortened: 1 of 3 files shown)


lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -38,34 +39,47 @@ const isEqualValueNodes = (nodes1, nodes2) => {
...
+
+		// Custom properties have unknown value syntaxes but are equal for CSS parsers.
+		if (
+			node1.type === 'Identifier' &&
+			node1Name.startsWith('--') &&
+			node2.type === 'Identifier' &&
+			node2Name.startsWith('--')
+		) {
+			continue;
+		}
```
romainmenke, line 58: Dashed idents are different from regular idents.
These are always known to be the same syntax.

.changeset/many-toes-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed false positives with option `ignore: ["consecutive-duplicates-with-different-syntaxes"]` in `declaration-block-no-duplicate-properties`
```
ybiquitous, line 5: [suggest] Can you fix the message for consistency?

```suggestion
Fixed: `declaration-block-no-duplicate-properties` false positives with option `ignore: ["consecutive-duplicates-with-different-syntaxes"]`
```

lib/rules/declaration-block-no-duplicate-properties/index.js
```diff
@@ -38,34 +39,47 @@ const isEqualValueNodes = (nodes1, nodes2) => {
...
 
+		const node1Name = 'name' in node1 ? String(node1.name) : '';
+		const node2Name = 'name' in node2 ? String(node2.name) : '';
+
+		// Custom properties have unknown value syntaxes but are equal for CSS parsers.
+		if (
+			node1.type === 'Identifier' &&
+			node1Name.startsWith('--') &&
+			node2.type === 'Identifier' &&
+			node2Name.startsWith('--')
```
ybiquitous, line 55: [suggest] Can you use `isCustomProperty()` for consistency?

```suggestion
			isCustomProperty(node1Name) &&
			node2.type === 'Identifier' &&
			isCustomProperty(node2Name)
```

romainmenke: I should have checked what this function does :)
I assumed it took a declaration as its argument.

ybiquitous (approved):
Thanks, LGTM 👍🏼

romainmenke:
Thank you for reviewing 🙇

Issue #6795, opened 2023-04-18
Release 15.6.0


ybiquitous:
- [x] stylelint release
- [x] stylelint-config-recommended update/release
- [x] stylelint-config-standard update/release
- [x] stylelint.io update
- [x] stylelint-demo check
- [x] tweet

This is the [next changelog](https://github.com/stylelint/stylelint/pull/6794/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed). We have a bugfix for a syntax error to be released as soon as possible.

I'll release it in a few days if anyone doesn't have other concerns.

ybiquitous:
PRs waiting for reviews:

- #6800

If possible, I'd like to include the above in v15.5.1.

mattxwang:
On my end, also seems reasonable to release! Let me know what else I can do to help.

ybiquitous:
I've changed the title (version) to include the new features:

- #6778
- #6802

Because these features will not likely break things (I feel safe).

ybiquitous:
@mattxwang Thanks for your great contribution! I'll do the release this weekend.

ybiquitous:
I will start the release work now.

ybiquitous:
https://github.com/stylelint/stylelint/releases/tag/15.6.0 is out! 🎉

ybiquitous:
This release is done! 🎉 

Thanks a lot for your contributions and efforts, everyone!

PR #6802, opened 2023-04-20
Add `ignore: ["with-var-inside"]` to `color-function-notation`

### LINKED ISSUE #6620: Add `ignore: ["with-var-inside"]` to `color-function-notation`

```css
div {
  background-color: rgba(var(--bs-primary-rgb), 0.4);
}
```

```json
{
  "ignoreFiles": ["**/*"],
  "overrides": [
    {
      "files": ["**/*.css"],
      "extends": ["stylelint-config-standard", "stylelint-config-prettier"],
      "rules": {}
    },
    {
      "files": ["**/*.scss"],
      "extends": ["stylelint-config-standard-scss", "stylelint-config-prettier"],
      "rules": {}
    }
  ],
  "rules": {}
}
```

`stylelint app/styles`

14.16.1

No error/warning showing since the syntax is correct.

### What actually happened?

An error is reported.

_No response_

### Proposal to fix the bug

Maybe related to https://github.com/stylelint/stylelint/issues/5671.

mattxwang:

Closes #6620.

No, it's self-explanatory.

### DIFF
--- a/lib/rules/color-function-notation/README.md
+++ b/lib/rules/color-function-notation/README.md
@@ -116,3 +116,37 @@ a { color: hsla(270, 60%, 50%, 15%) }
 ```css
 a { color: hsl(.75turn, 60%, 70%) }
 ```
+
+## Optional secondary options
+
+### `ignore: ["with-var-inside"]`
+
+Ignore color functions containing variables.
+
+Given
+
+```json
+["modern", { "ignore": ["with-var-inside"] }]
+```
+
+The following patterns are _not_ considered problems:
+
+```css
+a {
+  color: rgba(var(--foo), 0.5);
+}
+```
+
+Given
+
+```json
+["legacy", { "ignore": ["with-var-inside"] }]
+```
+
+The following patterns are _not_ considered problems:
+
+```css
+a {
+  color: rgba(var(--foo) / 0.5);
+}
+```
--- a/lib/rules/color-function-notation/README.md
+++ b/lib/rules/color-function-notation/README.md
@@ -123,7 +123,7 @@ a { color: hsl(.75turn, 60%, 70%) }
 
 Ignore color functions containing variables.
 
-Given
+Given:
 
 ```json
 ["modern", { "ignore": ["with-var-inside"] }]
@@ -137,7 +137,7 @@ a {
 }
 ```
 
-Given
+Given:
 
 ```json
 ["legacy", { "ignore": ["with-var-inside"] }]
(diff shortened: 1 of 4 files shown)


lib/rules/color-function-notation/README.md
```diff
@@ -116,3 +116,37 @@ a { color: hsla(270, 60%, 50%, 15%) }
...
 a { color: hsl(.75turn, 60%, 70%) }
 ```
+
+
+### `ignore: ["with-var-inside"]`
+
+Ignore color functions containing variables.
+
+Given
```
ybiquitous, line 126: Typo.

```suggestion
Given:
```

lib/rules/color-function-notation/README.md
```diff
@@ -116,3 +116,37 @@ a { color: hsla(270, 60%, 50%, 15%) }
...
+
+The following patterns are _not_ considered problems:
+
+```css
+a {
+  color: rgba(var(--foo), 0.5);
+}
+```
+
+Given
```
ybiquitous, line 140: Typo.

```suggestion
Given:
```

lib/rules/color-function-notation/index.js
```diff
@@ -46,6 +60,13 @@ const rule = (primary, _secondaryOptions, context) => {
 
 				const { value, sourceIndex, sourceEndIndex, nodes } = node;
 
+				if (
+					ignoreWithVarInside &&
+					nodes.filter((child) => child.type === 'function' && child.value.toLowerCase() === 'var')
+						.length > 0
+				)
+					return;
```
ybiquitous, line 68: [suggest] I recommend using `.some()` over `.filter().length`. Also, using `{}` for `return` at a new line seems more readable.

```suggestion
				if (
					ignoreWithVarInside &&
					nodes.some((child) => child.type === 'function' && child.value.toLowerCase() === 'var')
				) {
					return;
				}
```

ybiquitous: Or, extracting a helper function might be more readable like this:

```js
if (ignoreWithVarInside && containsVariable(nodes)) return;

// ...

/**
 * @param {import('postcss-value-parser').Node[]} nodes
 */
function containsVariable(nodes) {
	return nodes.some(({ type, value }) => type === 'function' && value.toLowerCase() === 'var');
}
```

ybiquitous (changes requested):
@mattxwang Thanks for the pull request! I've commented on refactoring, but this option addition is nice! 👍🏼

mattxwang:
Thanks for the feedback! I think I've addressed all the code suggestions.

ybiquitous (approved):
Thanks, LGTM 👍🏼

mattxwang:
(dup comment as #6778)

> Shall I hold off on merging this since #6795 is a patch?

ybiquitous:
Let's merge it. See https://github.com/stylelint/stylelint/issues/6795#issuecomment-1517018386

PR #6804, opened 2023-04-20
Add `number-no-partial-percent` rule

### LINKED ISSUE #6803: New rule: disallow percent size values that are not `0%` or `100%`.

I would like to have a linting error for percent size values in CSS that are not `0%` or `100%`.
Almost all of the usages I've seen so far of percent size values that are not `0%` or `100%` could have been replaced with a more responsive alternative approach, such as using flexbox and `min`, and `max` size values.
I would like to have a Stylelint rule to discourage using such values in the first place.

Examples of popular bad usages I seen so far:
```scss
.something {
    // using flexbox is better for most cases,
    // as it considers the content of the element and neighboring elements when calculating the size of it.
    // It's possible to also configure flexbox to not consider the content of the element,
    // but when you do need such behavior, you have to deliberately configure it this way
    width: 33%;

    // Centering an element using flexbox is better
    top: 50%;

    // 99% of the usages of this that I've seen so far tried to make an element look circular,
    // but using `999px` is more responsive when the content prevents the element from being circular
    border-radius: 50%;
}
```

I would like to have a new rule for this

giladgd:

Closes #6803

No, it's self-explanatory.

### DIFF
--- a/lib/rules/number-no-partial-percent/README.md
+++ b/lib/rules/number-no-partial-percent/README.md
@@ -1,0 +1,72 @@
+# number-no-partial-percent
+
+Disallow percent size values that are not `0%` or `100%`.
+
+> Usage of any percent values for alpha in colors is still allowed. (e.g. `rgba(0 0 0 / 60%)`)
+
+<!-- prettier-ignore -->
+```css
+a { width: 30%; }
+/**        ↑
+ *  This value */
+```
+
+## Options
+
+### `true`
+
+The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+a {
+  width: 30%;
+}
+```
+
+<!-- prettier-ignore -->
+```css
+a {
+  top: 50%;
+}
+```
+
+<!-- prettier-ignore -->
+```css
+a {
+  transform: translate(10%, 12%);
+}
+```
+
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+a {
+  color: rgb(0 0 0 / 60%);
+  border: solid 1px rgb(0 0 0 / 60%);
+}
+```
+
+> Alpha values in colors are allowed
+
+<!-- prettier-ignore -->
+```css
+a {
+  width: 100%;
+}
+```
+
+<!-- prettier-ignore -->
+```css
+a {
+  top: 0%;
+}
+```
+
+<!-- prettier-ignore -->
+```css
+a {
+  transform: translate(0%, 30px);
+}
+```
--- a/lib/rules/number-no-partial-percent/README.md
+++ b/lib/rules/number-no-partial-percent/README.md
@@ -2,7 +2,7 @@
 
 Disallow percent size values that are not `0%` or `100%`.
 
-> Usage of any percent values for alpha in colors is still allowed. (e.g. `rgba(0 0 0 / 60%)`)
+> Usage of any percent values in colors, color-related functions, and shape functions is still allowed. (e.g. `rgba(0 0 0 / 60%)`, `opacity(60%)`, `circle(40% at center)`)
 
 <!-- prettier-ignore -->
 ```css
@@ -43,16 +43,6 @@ The following patterns are _not_ considered problems:
 <!-- prettier-ignore -->
 ```css
 a {
-  color: rgb(0 0 0 / 60%);
-  border: solid 1px rgb(0 0 0 / 60%);
-}
-```
-
-> Alpha values in colors are allowed
-
-<!-- prettier-ignore -->
-```css
-a {
   width: 100%;
 }
 ```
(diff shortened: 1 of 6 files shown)


lib/rules/number-no-partial-percent/index.js
```diff
@@ -0,0 +1,74 @@
...
+const balancedMatch = require('balanced-match');
+
+const report = require('../../utils/report');
+const ruleMessages = require('../../utils/ruleMessages');
+const validateOptions = require('../../utils/validateOptions');
+
+const ruleName = 'number-no-partial-percent';
+
+const messages = ruleMessages(ruleName, {
+	invalidPercent: 'Percent values must be 0% or 100%',
```
Mouvedia, line 12: minor

maybe add _either_?

lib/rules/number-no-partial-percent/index.js
```diff
@@ -0,0 +1,74 @@
...
+			return;
+		}
+
+		root.walkDecls((decl) => {
+			const value = decl.value;
+
+			if (!value.includes('%')) return;
+
+			// Match that start of color functions and percent values
+			const matches = value.matchAll(/(?:rgba?|color)\s*\(|[\d.]+%/g);
```
Mouvedia, line 34: question

Is the list of functions exhaustive?

lib/rules/number-no-partial-percent/index.js
giladgd: Done

lib/rules/number-no-partial-percent/index.js
giladgd: I went over all the CSS functions that exist and made sure this list is exhaustive

giladgd: I went over all the CSS functions that exist and made sure this list is exhaustive

lib/rules/number-no-partial-percent/index.js
```diff
@@ -30,8 +30,12 @@ const rule = (primary) => {
 
 			if (!value.includes('%')) return;
 
-			// Match that start of color functions and percent values
-			const matches = value.matchAll(/(?:rgba?|color)\s*\(|[\d.]+%/g);
+			// Match the start of color functions,
+			// functions where their content should be ignored and percent value can be found inside of them (such as `url("data:image/svg+xml,203%22")`),
```
Mouvedia, line 34: question

Should it trigger `transform: translateX(50%);`?

giladgd: `transform: translateX(50%);` will trigger a linting error from this rule; there are also a few tests that check such cases in this PR

lib/rules/number-no-partial-percent/index.js
```diff
@@ -0,0 +1,78 @@
...
+			// and percent values
+			const matches = value.matchAll(
+				/(?:rgba?|color|color-mix|color-contrast|device-cmyk|hsl|hwb|lch|oklch|lab|oklab|linear-gradient|radial-gradient|conic-gradient|repeating-linear-gradient|repeating-radial-gradient|repeating-conic-gradient|opacity|brightness|contrast|grayscale|invert|saturate|sepia|url|image|image-set|cross-fade|element|paint|circle|ellipse|inset|polygon|path)\s*\(|[\d.]+%/g,
+			);
+
+			if (matches == null) return;
+
+			let lastFuncEndIndex = -1;
+
+			for (const match of matches) {
```
Mouvedia, line 44: minor nit
```js
for (const { 0: text, index } of matches) {
```
since you are not reusing `match`

lib/rules/number-no-partial-percent/__tests__/index.js
```diff
@@ -0,0 +1,168 @@
...
+	config: [true],
+	accept: [
+		{
+			code: 'a { width: 100%; }',
+		},
+		{
+			code: 'a { width: 0%; }',
+		},
+		{
+			code: 'a { top: 0%; }',
```
Mouvedia, line 18: Can you add a test for `.0`?

lib/rules/number-no-partial-percent/README.md
```diff
@@ -0,0 +1,91 @@
+# number-no-partial-percent
+
+Disallow percent size values that are not `0%` or `100%`.
+
+> Usage of any percent values in colors, color-related functions, and shape functions is still allowed. (e.g. `rgba(0 0 0 / 60%)`, `opacity(60%)`, `circle(40% at center)`)
```
Mouvedia, line 5: question

Should we mention that `calc` evaluation is not supported?

jeddy3:
Thank you for the pull request. It's a great candidate for a plugin.

See https://github.com/stylelint/stylelint/issues/6803#issuecomment-1604240221.

PR #6811, opened 2023-04-24
Fix `selector-pseudo-class-no-unknown` false positive for `:modal`

### LINKED ISSUE #6809: Fix `selector-pseudo-class-no-unknown` false positive for `:modal`

I'm writing CSS for only modal dialogs (not vanilla dialogs). The web spec allows `:modal` to target such cases. But with stylelint I get an error.

`Unexpected unknown pseudo-class selector ":modal" (selector-pseudo-class-no-unknown)`

Would like to see a warning instead of an error due to low range of browser support perhaps?

Yasir761:
Closes #6809.

### DIFF
--- a/.changeset/many-mugs-camp.md
+++ b/.changeset/many-mugs-camp.md
@@ -2,4 +2,4 @@
 "stylelint": minor
 ---
 
-Fixed selector-pseudo-class-no-unknown false positive for :modal
+Fixed: `selector-pseudo-class-no-unknown` false positive for `:modal`
--- a/.changeset/many-mugs-camp.md
+++ b/.changeset/many-mugs-camp.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Fixed selector-pseudo-class-no-unknown false positive for :modal
--- a/lib/rules/selector-pseudo-class-no-unknown/__tests__/index.js
+++ b/lib/rules/selector-pseudo-class-no-unknown/__tests__/index.js
@@ -29,6 +29,9 @@ testRule({
 			code: 'a::before { }',
 		},
 		{
+			code: 'a:modal { }',
+		},
+		{
 			code: "input:not([type='submit']) { }",
 		},
 		{
@@ -122,6 +125,14 @@ testRule({
 		{
 			code: 'a:unknown { }',
 			message: messages.rejected(':unknown'),
+			line: 1,
+			column: 2,
+			endLine: 1,
+			endColumn: 10,
+		},
+		{
+			code: 'a:modal { }',
+			message: messages.rejected(':modal'),
 			line: 1,
 			column: 2,
 			endLine: 1,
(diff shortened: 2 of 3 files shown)


mattxwang (changes requested):
Hi @Yasir761, thanks for submitting a change! Two quick asks:

1. could you add a test case that validates your fix?
2. could you add a changeset using `npx changeset`?

(I've renamed the PR to match our stylistic rules; you can use the title for the changeset; just change the tense)

Yasir761:
> Hi @Yasir761, thanks for submitting a change! Two quick asks:
> 
> 1. could you add a test case that validates your fix?
> 2. could you add a changeset using `npx changeset`?
> 
> (I've renamed the PR to match our stylistic rules; you can use the title for the changeset; just change the tense)

off course I'll add these, can you tell me where I should add test case?

mattxwang:
> off course I'll add these, can you tell me where I should add test case?

@Yasir761 each rule has a `__tests__` subdirectory; so, I'd add it within `lib/rules/selector-pseudo-class-no-unknown/__tests__/index.js`. Since this is a false positive, you'll want to write a simple accept case that uses `:modal`. Try to match the conventions we use for other test cases; for more, see the [writing tests](https://stylelint.io/developer-guide/rules#write-tests) section of our developer guide!

.changeset/many-mugs-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Fixed selector-pseudo-class-no-unknown false positive for :modal
```
mattxwang, line 5: Changeset nit:

```suggestion
Fixed: `selector-pseudo-class-no-unknown` false positive for `:modal`
```

lib/rules/selector-pseudo-class-no-unknown/__tests__/index.js
```diff
@@ -127,6 +130,14 @@ testRule({
...
 			endColumn: 10,
 		},
+		{
+			code: 'a:modal { }',
+			message: messages.rejected(':modal'),
+			line: 1,
+			column: 2,
+			endLine: 1,
+			endColumn: 10,
+		},
```
mattxwang, line 140: Hm, it's unclear to me why this test case is here; we *don't* want to reject `:modal` since it is a known pseudo-class. Could you remove this test case?

```suggestion
```

lib/rules/selector-pseudo-class-no-unknown/__tests__/index.js
```diff
@@ -28,6 +28,9 @@ testRule({
 		{
 			code: 'a::before { }',
 		},
+		{
+			code: 'a:modal { }',
```
mattxwang, line 32: Bit of a nit, but it seems like from documentation ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:modal), [CSS spec](https://www.w3.org/TR/selectors-4/#selectordef-modal)) that no attached selector is more idiomatic:

```suggestion
			code: ':modal { }',
```

mattxwang (changes requested):
Thanks for submitting some changes @Yasir761; I've added some comments, could you take a look at them and let me know what you think!

(as a suggestion: make sure that `npm test` passes locally!)

mattxwang (approved):
Thanks - LGTM!

PR #6815, opened 2023-04-26
Fix `declaration-block-no-redundant-longhand-properties` autofix for `transition`

### LINKED ISSUE #6812: Fix `declaration-block-no-redundant-longhand-properties` autofix for `transition`

Given code like:
```css
a {
  transition-delay: 500ms;
  transition-duration: 250ms;
  transition-timing-function: ease-in-out;
  transition-property: transform, visibility;
}
```

```json
{
  "rules": {
  }
}
```

https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYAdAdilAXATkpgM4CWupEmAtLADZICe8KArAAzsC2xA3FjgJEyFKrQCuhUZhYAmDj37Y8hEuUo0KXUpgDm1AGbjMYaSxhJiMajuoRxuJYNUiN1AA74I7mPlzMVYQMIfC4AGhQAN1IyACNSOnJGJQBfEDCQAwSYADkkLjhEGAAPfPc6GAA6MGJidPAqLN0EEAxldBAS3BhMKGIOlgBtARwUDuJ-CsTMXGpITCbqCaIoJHwoDpGAXSw0jPmmgDEQriRcFoArYip62Hc6xDacccmYadwBsZAGbomOsJGL0YUx0swOpH0y16aw2CC+PxgfxAuxAKSAA

`15.6.0`

No problems to be reported, or, if a `declaration-block-no-redundant-longhand-properties` error is reported, the autofix should be 
```css
a {
  transition: transform 250ms ease-in-out 500ms, visibility 250ms ease-in-out 500ms;
}
```

### What actually happened?

An error of `Expected shorthand property "transition"` with an incorrect suggested autofix of:

```css
a {
  transition: transform, visibility 250ms ease-in-out 500ms;
}
```

No

### Proposal to fix the bug

IMO, this error shouldn't be reported at all for transitions with multiple properties (as reported in https://github.com/stylelint/stylelint/issues/4074), but if it is considered an error, then the autofix at least needs to be changed.

mattxwang:

Closes #6812.

This one required some non-trivial CSS spec reading, hence the complicated-ish behaviour!

### DIFF
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/index.js
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/index.js
@@ -50,12 +50,20 @@ const customResolvers = new Map([
 	[
 		'transition',
 		(decls) => {
-			const delays = decls.get('transition-delay')?.value.trim().split(',');
-			const durations = decls.get('transition-duration')?.value.trim().split(',');
-			const timingFunctions = decls.get('transition-timing-function')?.value.trim().split(',');
-			const properties = decls.get('transition-property')?.value.trim().split(',');
+			/** @type {(input: string | undefined) => string[]} */
+			const commaSeparated = (input = '') =>
+				input
+					.split(',')
+					.map((s) => s.trim())
+					.filter((s) => s.length > 0);
+			const delays = commaSeparated(decls.get('transition-delay')?.value);
+			const durations = commaSeparated(decls.get('transition-duration')?.value);
+			const timingFunctions = commaSeparated(decls.get('transition-timing-function')?.value);
+			const properties = commaSeparated(decls.get('transition-property')?.value);
 
-			if (!(delays && durations && timingFunctions && properties)) return;
+			if (!(delays.length && durations.length && timingFunctions.length && properties.length)) {
+				return;
+			}
 
 			// transition-property is the canonical list of the number of properties;
 			// see spec: https://w3c.github.io/csswg-drafts/css-transitions/#transition-property-property
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/index.js
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/index.js
@@ -71,12 +71,16 @@ const customResolvers = new Map([
 			// the other properties are computed cyclically -- ex with %
 			// see spec example #3: https://w3c.github.io/csswg-drafts/css-transitions/#example-d94cbd75
 			return properties
-				.map(
-					(property, i) =>
-						`${property.trim()} ${durations[i % durations.length]?.trim()} ${timingFunctions[
-							i % timingFunctions.length
-						]?.trim()} ${delays[i % delays.length]?.trim()}`,
-				)
+				.map((property, i) => {
+					return [
+						property,
+						durations[i % durations.length],
+						timingFunctions[i % timingFunctions.length],
+						delays[i % delays.length],
+					]
+						.filter(isString)
+						.join(' ');
+				})
 				.join(', ');
 		},
 	],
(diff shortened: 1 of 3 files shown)


ybiquitous (changes requested):
@mattxwang Thanks for the pull request! Please consider addressing my reviews (mainly refactoring).

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -47,6 +47,39 @@ const customResolvers = new Map([
...
+
+			if (!(delays && durations && timingFunctions && properties)) return;
+
+			// none, inherit and initial are not permitted as items of lists with more than one identifier
+			// see spec: https://w3c.github.io/csswg-drafts/css-transitions/#transition-property-property
+			if (
+				properties.length > 1 &&
+				('none' in properties || 'inherit' in properties || 'initial' in properties)
+			)
+				return;
```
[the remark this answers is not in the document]
mattxwang: Ah! Great timing, I had just realized this independently of you submitting the review (got sidetracked while writing this up). I added a test case for the unpermitted keyword (this is a dupe of a previous change I made for this autofix), as well as one for `none`.

mattxwang:
So sorry for the rushed PR, I got sidetracked (in-person) in the middle of finishing it up - meant to mark it as draft 😅 . Hopefully it wasn't too frustrating to review!

I think I've addressed all of the comments (and slightly changed my approach re: the `none` special case, after re-reading the spec). Let me know what you think!

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -47,6 +47,43 @@ const customResolvers = new Map([
...
+				input
+					.split(',')
+					.map((s) => s.trim())
+					.filter((s) => s.length > 0);
+			const delays = commaSeparated(decls.get('transition-delay')?.value);
+			const durations = commaSeparated(decls.get('transition-duration')?.value);
+			const timingFunctions = commaSeparated(decls.get('transition-timing-function')?.value);
+			const properties = commaSeparated(decls.get('transition-property')?.value);
+
+			if (!(delays.length && durations.length && timingFunctions.length && properties.length)) {
```
mattxwang, line 64: I'm unsure of how to write a test here that isn't invalid syntax (one of the properties would have to be empty) - any thoughts?

ybiquitous: How about specifying an empty declaration value (`transition-delay`) like this?

```css
a { transition-delay: ; transition-duration: 0.1s; transition-timing-function: ease; transition-property: top; }
```

mattxwang: Thanks for the suggestion! Hm, this seems like invalid CSS to me (though I could be wrong) - would that still make it a good test case?

ybiquitous: Surely that CSS seems invalid, but PostCSS can parse it as having an empty value. See the example below:

```js
require("postcss").parse("a { color: ; }")
  .walkDecls(({ prop, value }) => console.log({ prop, value }));
//=> { prop: 'color', value: '' }
```

Adding such a case seems to make sense.

mattxwang: Gotcha; I'm still a bit confused how to proceed, since that case wouldn't be fixable (or in other words, what would I put in `fixed`), but in that case it seems to break the test infrastructure. 

Alternatively, I could extract this out as a utility and have more rigorous tests for the custom resolvers themselves - what do you think?

ybiquitous: How about specifying `unfixable: true` like this?

```js
{
	code: 'a { transition-delay: ; transition-duration: 1s; transition-timing-function: ease; transition-property: top; }',
	unfixable: true,
},
```

ybiquitous (approved):
Thank you! LGTM 👍🏼

PR #6817, opened 2023-04-26
Improve bug report template

ybiquitous:

None.

This pull request improves the bug report template so that we can easily reproduce a reported issue.

Changes summary:
- Add description and placeholder to `reproduce-bug` section
- Avoid deprecated rule `number-leading-zero`
- Add a demo link for easy reproduction
- Add an example of dependencies
- Use `textarea` instead of `input` so that people can easily input

I've prepared a test repository to try this PR's change easily:
https://github.com/ybiquitous/stylelint-pull-6817/issues/new?template=REPORT_A_BUG.yml


mattxwang (approved):
These all look good to me! Hopefully this will encourage more use of the demo!

PR #6818, opened 2023-04-26
Add `ignorePatterns` config option

### LINKED ISSUE #5804: Integrate `ignoreFiles` configuration and `--ignore-pattern` CLI flag

We have the following two options to ignore files from linting:

### `ignoreFiles` (.stylelintrc)

https://github.com/stylelint/stylelint/blob/dbbfe21e1c945c9d54b16ddd5c1570b4ae3f935e/docs/user-guide/configure.md#L368-L380

### `--ignore-pattern` (CLI)

https://github.com/stylelint/stylelint/blob/dbbfe21e1c945c9d54b16ddd5c1570b4ae3f935e/docs/user-guide/usage/cli.md#L63-L65

These options are similar but the usages are different: from Configuration and CLI.
I think this usability is not good. 😕 

So I suggest the following:

- add `ignorePatterns` option to `.stylelintrc`
- deprecate `ignoreFiles` option of `.stylelintrc`

Thus, users can manage only one file `.stylelintrc` for stylelint configuration; that is, it is unnecessary to manage also `.stylelintignore`.

For example, `.stylelintignore`:

```diff
- dist/**/*.css
```

`.stylelintrc`:

```diff
{
+ "ignorePatterns": [
+   "dist/**/*.css"
+ ]
}
```

See also:
- [Ignoring Code - ESLint - Pluggable JavaScript linter](https://eslint.org/docs/user-guide/configuring/ignoring-code)

mattxwang:

Closes #5804.

Draft - I think I Need to improve some test coverage and nail down some of the behavioural changes.

In addition,

- need to add docs
- potentially add deprecation warning?
- changeset

### DIFF
--- a/lib/__tests__/fixtures/config-extending-and-ignoring-patterns.json
+++ b/lib/__tests__/fixtures/config-extending-and-ignoring-patterns.json
@@ -1,0 +1,4 @@
+{
+	"ignorePatterns": ["**/empty-block.css"],
+	"extends": "./config-block-no-empty"
+}
--- a/lib/__tests__/ignore.test.js
+++ b/lib/__tests__/ignore.test.js
@@ -12,6 +12,323 @@ const fixtures = (...elem) => replaceBackslashes(path.join(__dirname, 'fixtures'
 
 safeChdir(__dirname);
 
+// ignorePath
+
+test('specified `ignorePath` file ignoring one file', async () => {
+	const files = [fixtures('empty-block.css')];
+
+	await expect(
+		standalone({
+			files,
+			config: {
+				rules: {
+					'block-no-empty': true,
+				},
+			},
+			ignorePath: [fixtures('ignore.txt')],
+		}),
+	).rejects.toThrow(AllFilesIgnoredError); // all files ignore
+});
+
+test('specified multiple `ignorePath` to ignore files', async () => {
+	const files = [fixtures('empty-block.css'), fixtures('empty-block-with-disables.css')];
+
+	await expect(
+		standalone({
+			files,
+			config: {
+				rules: {
+					'block-no-empty': true,
+				},
+			},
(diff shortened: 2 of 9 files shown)


Mouvedia:
> potentially add deprecation warning?

see #7449

ybiquitous:
@Mouvedia Thanks for the reminder. Right, this issue/PR duplicates with #7449 in our new config format. I'll close the original issue #5804 if @mattxwang agrees.

mattxwang:
Ah! So sorry - to be frank, I've been way behind on many of my open-source obligations.

> I'll close the original issue https://github.com/stylelint/stylelint/issues/5804 if @mattxwang agrees.

Go for it!

Also, I know I've had a couple of PRs languishing (I haven't had the chance to really work on them) and I don't think I'll be able to push them over until at least March. Let me know if I should close #7194 or #7229 (needs more scoping out of what this will look like).

ybiquitous:
@mattxwang Thanks for the quick response. I've just closed #6818, so I'm closing this PR too.

> Let me know if I should close #7194 or #7229 (needs more scoping out of what this will look like).

Let's close them since they need to be rewritten for ESM (`*.mjs`). You could restart with a new pull request if you want.

Mouvedia:
> I'll close the original issue #5804 if @mattxwang agrees.

It should be reduced to

> add `ignorePatterns` option

i.e. it's only partially a duplicate.

Mouvedia:
> I'll close the original issue #5804 if @mattxwang agrees.

It's only partially a duplicate; #7449 doesn't have a flag listed.
i.e. `--ignores` is missing
cf https://eslint.org/docs/latest/use/command-line-interface#--ignore-pattern

ybiquitous:
> #7449 doesn't have a flag listed.

Right. But it's OK since we could revisit the flag after #7449 (and the new config format) is implemented. Leaving the original issue #5804 would be confusing.

PR #6840, opened 2023-05-11
Fix `selector-anb-no-unmatchable` doc to include in recommended and standard configs

sxyazi:
The `selector-anb-no-unmatchable` rule is actually has been included in the recommended and standard.

https://github.com/stylelint/stylelint-config-recommended/blob/b35e61355d4a9853532a9f977219453456d6b1bc/index.js#L36


ybiquitous (approved):
Thank you! LGTM 👍🏼

Issue #6582, opened 2023-01-18
Fix `function-no-unknown` false positives for unspaced operators against nested brackets


IgnatovDan:
### What steps are needed to reproduce the bug?

Add a css file with this class:
```
.class1 {
  height: calc(10 - (10 / 10*(10 - 1)));
  height: calc(10 - (10 / (10 + 1)*(10 - 1)));
}
```

### What Stylelint configuration is needed to reproduce the bug?

```
module.exports = {
  "rules": {
    "function-no-unknown": true,
  }
}
```

### How did you run Stylelint?

"stylelint": "npx stylelint \"**/*.css\"",

### Which version of Stylelint are you using?

14.16.0

### What did you expect to happen?

No problems to be reported.

### What actually happened?

```
  8:27  ✖  Unexpected unknown function "10*"    function-no-unknown
  9:33  ✖  Unexpected unknown function "*"      function-no-unknown
```

### Does the bug relate to non-standard syntax?

No

jeddy3:
@IgnatovDan Thanks for the report and for using the template.

[...]

PR #6842, opened 2023-05-12
Fix `function-no-unknown` false positives for unspaced operators against nested brackets

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/6582

This change migrates this rule from `postcss-value-parser` to `@csstools/css-parser-algorithms`.

### DIFF
--- a/lib/rules/function-no-unknown/index.js
+++ b/lib/rules/function-no-unknown/index.js
@@ -1,15 +1,19 @@
 'use strict';
 
 const fs = require('fs');
-const valueParser = require('postcss-value-parser');
 const functionsListPath = require('css-functions-list');
+const { tokenize } = require('@csstools/css-tokenizer');
+const {
+	parseCommaSeparatedListOfComponentValues,
+	isFunctionNode,
+	isSimpleBlockNode,
+} = require('@csstools/css-parser-algorithms');
 
 const declarationValueIndex = require('../../utils/declarationValueIndex');
 const optionsMatches = require('../../utils/optionsMatches');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
 const validateOptions = require('../../utils/validateOptions');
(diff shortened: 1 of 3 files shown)


lib/rules/function-no-unknown/index.js
```diff
@@ -60,39 +64,42 @@ const rule = (primary, secondaryOptions) => {
...
+				if (!isFunctionNode(componentValue)) return;
 
-				if (node.type !== 'function') {
-					return;
-				}
+				const name = componentValue.getName();
 
-				if (!isStandardSyntaxFunction(node)) {
-					return;
-				}
```
mattxwang, line 72: Since we've removed this check, is it possible that this rule will now flag errors for SASS/CSS-in-JS users? If so, I imagine that this might be slightly breaking for those users (and/or something we propagate to configs)?

(I could also be missing something about how SASS and CSS-in-JS play together with this rule)

mattxwang (commented):
I have one quick question, but other than that this looks good to me - glad to see that the `@csstools` packages are useful!

lib/rules/function-no-unknown/index.js
romainmenke: Good catch :)

```js
module.exports = function isStandardSyntaxFunction(node) {
	// Function nodes without names are things in parentheses like Sass lists
	if (!node.value) {
		return false;
	}

	if (node.value.startsWith('#{')) {
		return false;
	}

	// CSS-in-JS interpolation
	if (node.value.startsWith('${')) {
		return false;
	}

	// CSS-in-JS syntax
	if (node.value.startsWith('`')) {
		return false;
	}

	return true;
};
```

None of those cases would be parsed as a `FunctionNode` by `@csstools/css-parser-algorithms`.

Only `<ident><open-parenthesis>` (e.g. `foo(`) starts a function in standard CSS.

This behavior makes `isFunctionNode` the equivalent of  `isStandardSyntaxFunction`.

ybiquitous (approved):
LGTM 👍🏼

Issue #6823, opened 2023-04-30
Fix `alpha-value-notation` false negatives for `oklab()`, `oklch()`, and `color()`


ArseniyKruglov:
### What steps are needed to reproduce the bug?

```css
body
{
	background-color: oklab(0 0 0 / 50%);
	background-color: oklch(0 0 0 / 50%);
	background-color: color(display-p3 0 0 0 / 50%);
}
```

### What Stylelint configuration is needed to reproduce the bug?

```json
{
	"rules":
	{
		"alpha-value-notation": "number"
	}
}
```

### How did you run Stylelint?

[Stylelint playground](https://stylelint.io/demo/#N4Igxg9gJgpiBcIBG0CeAdAdsL6AuSAhmANYDmAThAK6ZQC0kANhBfAAQAWAzkwBQAGdgICkwsQHp2AVlEBKANz58EgFTsA4hGjtVE3AWLkqtBs1YcIJJmE6DhDqbJGLleNewBChKLv2YAXxAAGhAAMwBLJhgAOUIAWzhEGAAPBIAHaIA6MG5uEPAITEiyBBAcTHZ2dBAKamjuGvgsKoqqqprCJnTOQnoANy7qGHpMCDxCPAiipuqQTGp4pBgKGpb2AKwg0MhiiLIAMVZ4ybKAK24igth0-MQ2ue48VGimCMw8WZqmSZgnmuC6xqTxeMDeH0YRRK9CehDohAoUC+IB+eD+nxAWxAASAA)

### Which version of Stylelint or dependencies are you using?

15.6.0

### What did you expect to happen?

```
ERROR Expected "50%" to be "0.5"
```

### What actually happened?

No error.

### Does the bug relate to non-standard syntax?

No, it's modern syntax from [CSS Color Module Level 4](https://w3c.github.io/csswg-drafts/css-color/#color-type).

### Proposal to fix the bug

Add OKLAB & OKLCH to color functions list in [this](https://github.com/stylelint/stylelint/blob/main/lib/rules/alpha-value-notation/index.js) file. `color()` syntax is different, so it will be harder to implement.

PR #6844, opened 2023-05-12
Fix `alpha-value-notation` false negatives for `oklab()`, `oklch()`, and `color()`

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/6823

This didn't require any logic changes because this rule just scans ahead for the first `/`.
This works for all color functions and for relative color syntax.

I've added a few more tests to verify that everything is working as expected.
Happy to add more if needed.

### DIFF
--- a/.changeset/healthy-planets-hide.md
+++ b/.changeset/healthy-planets-hide.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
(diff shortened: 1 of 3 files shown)


mattxwang (approved):
LGTM!

(glad it didn't require changing any of the rule logic)

Issue #6741, opened 2023-03-28
Fix `function-url-quotes` autofix for double-slash comments in SCSS maps


jgerigmeyer:
### What steps are needed to reproduce the bug?

```scss
$test-map: (
  // test comment
  foo: 'bar'
);
```

Might be a bug with the `postcss-scss` custom syntax imported by stylelint-config-recommended-scss?

### What Stylelint configuration is needed to reproduce the bug?

```yaml
extends:
  - stylelint-config-recommended-scss

rules:
  function-url-quotes: always
```

or

```yaml
extends:
  - stylelint-config-recommended-scss

rules:
  function-url-quotes: never
```

### How did you run Stylelint?

https://stylelint.io/demo/#N4Igxg9gJgpiBcIAkAXGBnFBaAtgQwAd4ACACgB0A7Y4geluLU2MhxxkpSpoDMIISAcgBGeAE6CqASgDcVEABoQPAJYAbGADk87BCBgAPHQQ0A6dGHTpF4CJVUBzPYbSUo6eN2JZimAJ4aaiqcWJD2Kg5YYjCs7G4wUFgWVlRUYgCuGh5ePOmUYCgqdljpYmpYAI7pEEwkeGoA7nh+6PJKYY4AYhBi+Ch6fjpqNrAE1ojAXuQg-oHBXAjE02p4TAsKUzMoATBBIR0RUTEQbBywicmti8urGAtUAL4gD0A

### Which version of Stylelint are you using?

`15.3.0`

### What did you expect to happen?

No change to original code:

```scss
$test-map: (
  // test comment
  foo: 'bar'
);
```

### What actually happened?

```scss
$test-map: (
  /* test comment*/
  foo: 'bar'
);
```

### Does the bug relate to non-standard syntax?

SCSS maps

ybiquitous:
@jgerigmeyer Thanks for the report. We can reproduce the bug with the demo.

When I looked into the `function-url-quotes` implementation, perhaps the following function seems a problem:

https://github.com/stylelint/stylelint/blob/fe4a86f8e341ef9f50dd6163f312da73f004ea02/lib/rules/function-url-quotes/index.js#L58-L68

The `getDeclarationValue()` utility checks `.raws.value.raw`, while `postcss-scss` produces `.raws.value.scss`:

https://github.com/stylelint/stylelint/blob/fe4a86f8e341ef9f50dd6163f312da73f004ea02/lib/utils/getDeclarationValue.js#L7-L11

[...]

PR #6847, opened 2023-05-15
Fix `function-url-quotes` false positives for SCSS `with()` construct

ybiquitous:

Ref: https://github.com/stylelint/stylelint/issues/6741#issuecomment-1546878778

This PR introduces a better logic, so the rule implementation becomes simpler (removing the `containsOnlyComments()` internal function).


mattxwang (approved):
LGTM!

(definitely agree that this is simpler!)

Issue #6848, opened 2023-05-16
Refactor to replace `postcss-media-query-parser` with `@csstools/media-query-list-parser`


ybiquitous:
Applicable rules:

- [x] `media-feature-name-allowed-list` #6852
- [x] `media-feature-name-case` #6849
- [x] `media-feature-name-disallowed-list` #6851
- [x] `media-feature-name-no-unknown` #6850
- [x] `unit-no-unknown` #6879
- [x] `unit-disallowed-list` #6886
- [x] `no-duplicate-at-import-rules` #6996
- [x] `media-feature-name-value-allowed-list` #6999
- [ ] do a final, cross rule review after all refactors

---

While `postcss-media-query-parser` is old and unmaintained, `@csstools/media-query-list-parser` supports modern syntaxes and is used now in this project. Therefore, I believe we can completely switch.

| Name | Repo | npm | Last published |
|--------|--------|--------|---|
| `postcss-media-query-parser` | https://github.com/dryoma/postcss-media-query-parser | https://www.npmjs.com/package/postcss-media-query-parser | Oct 27, 2016 |
| `@csstools/media-query-list-parser` | https://github.com/csstools/postcss-plugins/tree/HEAD/packages/media-query-list-parser | https://www.npmjs.com/package/@csstools/media-query-list-parser | April 10, 2023 |

The current usage:

```console
$ git grep -w 'postcss-media-query-parser' lib/
lib/rules/media-feature-name-allowed-list/index.js:8:const mediaParser = require('postcss-media-query-parser').default;
lib/rules/media-feature-name-case/index.js:7:const mediaParser = require('postcss-media-query-parser').default;
lib/rules/media-feature-name-disallowed-list/index.js:8:const mediaParser = require('postcss-media-query-parser').default;
lib/rules/media-feature-name-no-unknown/index.js:8:const mediaParser = require('postcss-media-query-parser').default;
lib/rules/media-feature-name-value-allowed-list/index.js:3:const mediaParser = require('postcss-media-query-parser').default;
lib/rules/no-duplicate-at-import-rules/index.js:3:const mediaParser = require('postcss-media-query-parser').default;
lib/rules/rangeContextNodeParser.js:22: * @param {import('postcss-media-query-parser').Node} node
lib/rules/unit-disallowed-list/index.js:6:const mediaParser = require('postcss-media-query-parser').default;
[...]

PR #6849, opened 2023-05-18
Refactor to add utility for media feature names and migrate `media-feature-name-case`

romainmenke:

I started with `media-feature-name-case` because it is the only rule for media feature names that also has an autofix option. Any utility that works for that rule will also work for rules without an autofix option.

Almost all of the complexity in this new utility is the result of trying to have some support for non-standard syntax.

There are two issues with non-standard syntax :

1. `@media @foo and (min-width: 320px) {}`

Top level non-standard syntax doesn't have any fallbacks in the media query parser. It's just an invalid list item.

To resolve this I moved to a "two stage" parser.

- first pass tokenizes and parses into component values.
- second pass takes simple blocks that start and end with parenthesis and parses these as media queries

`@media @foo and (min-width: 320px) and @bar {}` -> only `(min-width: 320px)` is fully parsed.

2. `@media (something = @foo)` vs. `@media (@foo = something)`

With range syntax there isn't a fixed order, so we can't make the same assumptions as for non-range syntax.

_`@media (something: @foo)` -> `something` is the feature name because it comes before `:` (colon)_

So the current strategy is to search for ident tokens and to check if these are found in a location that isn't invalid (doesn't mean valid) for media feature names.

This is exactly the same as what we already have for `media-feature-name-no-unknown`.
I didn't touch that rule yet because I only wanted to alter one rule at a time.

------

What is a good commit message for a refactor like this?

What is a good function name for the utility I created?
- it has a callback
- but it also returns something that can be used for autofix options
- did I try to do too much in this utility?

### DIFF
--- a/lib/rules/media-feature-name-case/__tests__/index.js
+++ b/lib/rules/media-feature-name-case/__tests__/index.js
@@ -165,7 +165,7 @@ testRule({
 		},
 		{
 			code: '@media not all and (\\MONOCHROME) {}',
-			fixed: '@media not all and (monochrome) { }',
+			fixed: '@media not all and (monochrome) {}',
 			message: messages.expected('MONOCHROME', 'monochrome'),
 			line: 1,
 			column: 21,
@@ -173,15 +173,15 @@ testRule({
 			endColumn: 32,
 		},
 		{
-			code: '@media not all and (\\@MONOCHROME) { }',
-			fixed: '@media not all and (\\40 monochrome) { }',
+			code: '@media not all and (\\@MONOCHROME) {}',
+			fixed: '@media not all and (\\40 monochrome) {}',
 			message: messages.expected('@MONOCHROME', '@monochrome'),
 			line: 1,
 			column: 21,
 		},
 		{
-			code: '@media not all and (\\40MONOCHROME) { }',
-			fixed: '@media not all and (\\40 monochrome) { }',
+			code: '@media not all and (\\40MONOCHROME) {}',
+			fixed: '@media not all and (\\40 monochrome) {}',
 			message: messages.expected('@MONOCHROME', '@monochrome'),
 			line: 1,
 			column: 21,
--- a/lib/rules/media-feature-name-case/__tests__/index.js
+++ b/lib/rules/media-feature-name-case/__tests__/index.js
@@ -68,6 +68,8 @@ testRule({
 			message: messages.expected('MONOCHROME', 'monochrome'),
 			line: 1,
 			column: 21,
+			endLine: 1,
+			endColumn: 31,
 		},
 		{
 			code: '@media not all and (mOnOcHrOmE) { }',
@@ -75,6 +77,8 @@ testRule({
 			message: messages.expected('mOnOcHrOmE', 'monochrome'),
 			line: 1,
 			column: 21,
+			endLine: 1,
+			endColumn: 31,
 		},
 		{
 			code: '@media (MIN-WIDTH: 700px) { }',
@@ -82,6 +86,8 @@ testRule({
 			message: messages.expected('MIN-WIDTH', 'min-width'),
 			line: 1,
 			column: 9,
+			endLine: 1,
+			endColumn: 18,
 		},
 		{
 			code: '@media (mIn-WiDtH: 700px) { }',
@@ -89,6 +95,8 @@ testRule({
 			message: messages.expected('mIn-WiDtH', 'min-width'),
 			line: 1,
 			column: 9,
+			endLine: 1,
+			endColumn: 18,
 		},
 		{
 			code: '@media (MIN-WIDTH: 700px) and (orientation: landscape) { }',
@@ -96,6 +104,8 @@ testRule({
 			message: messages.expected('MIN-WIDTH', 'min-width'),
 			line: 1,
 			column: 9,
+			endLine: 1,
+			endColumn: 18,
 		},
 		{
 			code: '@media (min-width: 700px) and (ORIENTATION: landscape) { }',
@@ -103,6 +113,8 @@ testRule({
 			message: messages.expected('ORIENTATION', 'orientation'),
 			line: 1,
 			column: 32,
+			endLine: 1,
+			endColumn: 43,
 		},
 		{
 			code: '@media (min-width: 700px) /* comments */ and (ORIENTATION: landscape) {}',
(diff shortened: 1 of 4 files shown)


ybiquitous (commented):
@romainmenke Thanks for the pull request! Does this change actually fix false negatives of `media-feature-name-case`?

lib/rules/media-feature-name-case/__tests__/index.js
```diff
@@ -163,6 +163,27 @@ testRule({
 			line: 1,
 			column: 16,
 		},
+		{
+			code: '@media not all and (\\MONOCHROME) { }',
+			fixed: '@media not all and (monochrome) { }',
+			message: messages.expected('MONOCHROME', 'monochrome'),
+			line: 1,
+			column: 21,
```
ybiquitous, line 171: [suggest] Could you test also `endLine`/`endColumn` since passing `index`/`endIndex` to `report()`? It's OK only for this PR's addition.

```suggestion
			column: 21,
			endLine: 1,
			endColumn: 32,
```

lib/utils/findMediaFeatureNamesSearch.js
```diff
@@ -0,0 +1,147 @@
...
+ * Found tokens are mutable and modifications made to them will be reflected in the output.
+ *
+ * This function supports some non-standard syntaxes like SCSS variables and interpolation.
+ *
+ * @param {string} mediaQueryParams
+ * @param {(mediaFeatureName: TokenIdent) => void} callback
+ *
+ * @returns {MediaQuerySerializer}
+ */
+module.exports = function findMediaFeatureNamesSearch(mediaQueryParams, callback) {
```
ybiquitous, line 36: [suggest] I feel that `find` and `Search` are duplicated. How about this? ⬇️ 

```suggestion
module.exports = function findMediaFeatureNames(mediaQueryParams, callback) {
```

lib/utils/findMediaFeatureNamesSearch.js
```diff
@@ -0,0 +1,147 @@
...
+					) {
+						callback(token);
+					}
+				});
+			}
+		});
+	});
+
+	return {
+		stringify() {
```
ybiquitous, line 106: [ask] What reason is to return a `stringify()` object instead of just a string?

lib/utils/findMediaFeatureNamesSearch.js
```diff
@@ -0,0 +1,147 @@
...
+		},
+	};
+};
+
+/** @param {import('@csstools/media-query-list-parser').GeneralEnclosed} node */
+function topLevelTokenNodes(node) {
+	const components = node.value.value;
+
+	if (isToken(components) || components.length === 0 || isToken(components[0])) {
+		return false;
```
ybiquitous, line 117: [suggest] How about simplifying the return type of this function?

```suggestion
		return [];
```

lib/utils/findMediaFeatureNamesSearch.js
```diff
@@ -0,0 +1,147 @@
...
+					return;
+				}
+
+				relevantTokens.forEach((token, i) => {
+					if (token[0] !== TokenType.Ident) {
+						return;
+					}
+
+					const nextToken = relevantTokens[i + 1];
+					const prevToken = relevantTokens[i - 1];
```
ybiquitous, line 84: [suggest] If `topLevelTokenNodes()` always returns an array, we can simplify the code like this:

```suggestion
				topLevelTokenNodes(node).forEach((token, i, tokens) => {
					if (token[0] !== TokenType.Ident) {
						return;
					}

					const nextToken = tokens[i + 1];
					const prevToken = tokens[i - 1];
```

ybiquitous (changes requested):
I think this refactoring is so nice! 👍🏼 

Please check my reviews to try to improve the code readability.

ybiquitous:
> Do you have a reference or examples for false negative issues?

Ah, sorry. I misunderstood fixing false negatives because this PR adds test cases. 😅 
This PR is just refactoring. No problems. 👍🏼

lib/utils/findMediaFeatureNamesSearch.js
romainmenke: Serializing takes time/resources.
So for rules that do not have an autofix option it is a bit wasteful to always serialize.

I don't have a strong opinion on this :)
We can simplify the code a bit and have everything run a bit slower, or more complexity and a bit faster.

lib/utils/findMediaFeatureNamesSearch.js
romainmenke: Sounds good! Thank you

lib/utils/findMediaFeatureNamesSearch.js
romainmenke: Yes, this better, thanks!

lib/utils/findMediaFeatureNamesSearch.js
ybiquitous: > So for rules that do not have an autofix option it is a bit wasteful to always serialize.

Make sense. Let's keep it. 👍🏼 

Additionally, leaving a code comment describing a reason for returning the `stringify()` object might be better.

romainmenke:
@ybiquitous I think this is ready for a final review.
Thank you for all the feedback 🙇 

Should this PR have a changeset?

ybiquitous (approved):
> Should this PR have a changeset?

No. This PR is a refactoring, so we don't need to update the changelog. 👍🏼 

Thank you for this great improvement. LGTM! 👏🏼

lib/utils/findMediaFeatureNames.js
```diff
@@ -0,0 +1,144 @@
...
+	isTokenNode,
+	parseCommaSeparatedListOfComponentValues,
+	isSimpleBlockNode,
+} = require('@csstools/css-parser-algorithms');
+const {
+	isGeneralEnclosed,
+	isMediaFeature,
+	isMediaQueryInvalid,
+	parseFromTokens,
+} = require('@csstools/media-query-list-parser');
```
alexander-akait, line 14: Just idea for future and future discussions, maybe we can union and write full featured CSS parser + at-rules/values parser from scratch, I am afraid we can't rewrite postcss due some specific logic (and it will probably take longer), so union around CSS parser will be great for any JS tooling, we can open an issue for this

alexander-akait: Yeah, @romainmenke What do you think about it?

romainmenke: Sounds good!
But indeed best to discus in a separate issue,
I have too many thoughts on this subject :D

PR #6850, opened 2023-05-19
Refactor to migrate `media-feature-name-no-unknown` to `findMediaFeatureNames` utility

romainmenke:

No, it's self-explanatory.

### DIFF
--- a/lib/rules/media-feature-name-no-unknown/index.js
+++ b/lib/rules/media-feature-name-no-unknown/index.js
@@ -2,6 +2,7 @@
 
 const atRuleParamIndex = require('../../utils/atRuleParamIndex');
 const findMediaFeatureNames = require('../../utils/findMediaFeatureNames');
+const getAtRuleParams = require('../../utils/getAtRuleParams');
 const isCustomMediaQuery = require('../../utils/isCustomMediaQuery');
 const isStandardSyntaxMediaFeatureName = require('../../utils/isStandardSyntaxMediaFeatureName');
 const { mediaFeatureNames } = require('../../reference/mediaFeatures');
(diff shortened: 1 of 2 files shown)


lib/rules/media-feature-name-no-unknown/index.js
```diff
@@ -52,129 +42,41 @@ const rule = (primary, secondaryOptions) => {
...
-									rangeFeatureOperator.test(prevToken[4].value))
-							) {
-								const [, , startIndex, endIndex, { value: featureName }] = token;
-
-								validateFeatureName(atRule, featureName, startIndex, endIndex);
-							}
-						});
-					}
+			const hasComments = atRule.raws.params && atRule.raws.params.raw;
+			const mediaRule = hasComments ? hasComments : atRule.params;
```
ybiquitous, line 47: [suggest] Modern operators can simplify the code like this:

```suggestion
			const mediaRule = atRule.raws.params?.raw ?? atRule.params;
```

ybiquitous (approved):
Thanks for the pull request! Much cleaner code. LGTM 👍🏼 

I've commented on a trivial refactoring suggestion, but you can merge it.

lib/rules/media-feature-name-no-unknown/index.js
ybiquitous: Ah, I forgot it, but we can use `getAtRuleParams()` utility instead: 😅 

https://github.com/stylelint/stylelint/blob/fd78ef7ad01123b0c4b05ad55d9986e3e1e21a01/lib/utils/getAtRuleParams.js#L7-L11

ybiquitous: We also can refactor the util, although it's out of the PR's scope.

```diff
 module.exports = function getAtRuleParams(atRule) {
-	const raws = atRule.raws;
-
-	return (raws.params && raws.params.raw) || atRule.params;
+	return atRule.raws.params?.raw ?? atRule.params;
 };
```

romainmenke: Awesome!
This utility is exactly what was missing :)

PR #6861, opened 2023-05-24
Fix `string` formatter for unexpected truncation on non-ASCII characters

### LINKED ISSUE #6860: `stringFormatter.js` truncate non-ASCII characters unexpectedly

To put it simply, this test case will fail (in `lib/formatters/__tests__/stringFormatter.test.js`):
```typescript
it('outputs warnings contains non-ASCII characters', () => {
		const results = [
			{
				source: 'path/to/file.css',
				warnings: [
					{
						line: 1,
						column: 1,
						rule: 'bar',
						severity: 'error',
						text: '简体中文こんにちは안녕하세요',
					},
				],
			},
		];

		const output = prepareFormatterOutput(results, stringFormatter);

		expect(output).toBe(stripIndent`
path/to/file.css
 1:1  ×  简体中文こんにちは안녕하세요  bar
1 problem (1 error, 0 warnings)`);
	});
```
in fact,  if a stylelint rule report error which `text` is '简体中文こんにちは안녕하세요', it will display '简体中文こんに'  inside the terminal(some characters are missing).

After locating it,  I found that the reason is:
the package `table` can't wrap non-ascii characters very well.
The smallest reproducible example:
```typescript
const table = require("table");

let config,
  data,
  output;

data = [
  ['简体中文'],
];

config = {
  columns: {
    0: {
      width: 20,
      wrapWord: true, // Note, enable `wrapWord`
    }
  }
};

output = table.table(data, config);

console.log(output); 
/*
╔══════════════════════╗
║ 简体                 
╚══════════════════════╝
*/
```
but if you disbale `wrapWord`, like: 
```typescript
// ...
config = {
  columns: {
    0: {
      width: 20,
      wrapWord: false, // Note, disable `wrapWord`
    }
  }
};
// ...
console.log(output); 
/*
╔══════════════════════╗
║ 简体中文             
╚══════════════════════╝
*/
```
it works well

```typescript
{
  "rules": {
    "selector-class-pattern": [/foo/, { "message": '简体中文' }]
  }
}
```

```css
.bar {}
```

```shell
npx stylelint 'foo.css'
```

15.6.2

output:
```shell
foo.css
 1:1  × 简体中文  selector-class-pattern

1 problem (1 error, 0 warnings)`);
```

### What actually happened?

output:
```shell
foo.css
 1:1  × 简体  selector-class-pattern

1 problem (1 error, 0 warnings)`);
```

note: '中文' is missing

No

### Proposal to fix the bug

disable `wrapWord` if error message contains non-ASCII characters.
if it is possible, i will open a PR for it later

Max10240:

See issus: [stringFormatter.js truncate non-ASCII characters unexpectedly](https://github.com/stylelint/stylelint/issues/6860)

Closes #6860

The reason that why don't I just contribute code to repository `table`, is: make it to support full ability about non-ASCII seems like a lot of work. before it, we can solve the problem in this way first。

### DIFF
--- a/lib/formatters/stringFormatter.js
+++ b/lib/formatters/stringFormatter.js
@@ -11,10 +11,8 @@ const { assertNumber } = require('../utils/validateTypes');
 const preprocessWarnings = require('./preprocessWarnings');
 const terminalLink = require('./terminalLink');
 
+const NON_ASCII_PATTERN = /\P{ASCII}/u;
 const MARGIN_WIDTHS = 9;
-// https://stackoverflow.com/questions/3203190/regex-any-ascii-character
-// eslint-disable-next-line no-control-regex
-const NON_ASCII_OR_WS_PATTERN = /[^\x00-\x7F]/;
 
 /**
  * @param {string} s
@@ -203,7 +201,8 @@ function formatter(messages, source, cwd) {
 		return row;
 	});
 
-	const hasNonAsciiChar = messages.some((msg) => NON_ASCII_OR_WS_PATTERN.test(msg.text));
+	const messageWidth = getMessageWidth(columnWidths);
+	const hasNonAsciiChar = messages.some((msg) => NON_ASCII_PATTERN.test(msg.text));
 
 	output += table
 		.table(cleanedMessages, {
@@ -214,8 +213,8 @@ function formatter(messages, source, cwd) {
 				2: { alignment: 'center', width: columnWidths[2] },
 				3: {
 					alignment: 'left',
-					width: getMessageWidth(columnWidths),
-					wrapWord: getMessageWidth(columnWidths) > 1 && !hasNonAsciiChar,
+					width: messageWidth,
+					wrapWord: messageWidth > 1 && !hasNonAsciiChar,
 				},
 				4: { alignment: 'left', width: columnWidths[4], paddingRight: 0 },
 			},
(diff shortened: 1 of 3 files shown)


lib/formatters/stringFormatter.js
```diff
@@ -200,6 +203,8 @@ function formatter(messages, source, cwd) {
 		return row;
 	});
 
+	const hasNonAsciiChar = messages.some((msg) => NON_ASCII_OR_WS_PATTERN.test(msg.text));
```
ybiquitous, line 206: [suggest] I'm unsure why, but it seems we don't need to call `getMessageWidth()` twice. Could you please reduce evaluation times?

```suggestion
	const messageWidth = getMessageWidth(columnWidths);
	const hasNonAsciiChar = messages.some((msg) => NON_ASCII_OR_WS_PATTERN.test(msg.text));
```

ybiquitous, line 206: [suggset] We can now use a [unicode character class escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape) because the spec has been available since Node.js 10.

So, could you simplify the regex like this?

```suggestion
	const hasNonAsciiChar = messages.some((msg) => /\P{ASCII}/u.test(msg.text));
```

ybiquitous (changes requested):
Thanks for the pull request. There seem to be no big problems. 👍🏼 
Please consider addressing my refactoring suggestions.

Max10240:
> Thanks for the pull request. There seem to be no big problems. 👍🏼 
> Please consider addressing my refactoring suggestions.

OK， thanks for your review，i will refactor it soon～

ybiquitous (approved):
Thank you. LGTM 👍🏼

Issue #6853, opened 2023-05-19
Implement full featured CSS parser


alexander-akait:
Just idea for future and future discussions, maybe we can union and write full featured CSS parser + at-rules/values parser from scratch, I am afraid we can't rewrite postcss due some specific logic (and it will probably take longer), so union around CSS parser will be great for any JS tooling, we can open an issue for this

_Originally posted by @alexander-akait in https://github.com/stylelint/stylelint/pull/6849#discussion_r1198368121_

ybiquitous:
@alexander-akait Thanks for raising this parser discussion! This is an exciting topic.

So, do you think this issue covers #5586?

I think also `@csstools/*` tools like [`@csstools/media-query-list-parser`](https://www.npmjs.com/package/@csstools/media-query-list-parser) or [`@csstools/css-tokenizer`](https://www.npmjs.com/package/@csstools/css-tokenizer) by @romainmenke are great achievements in this area.

ybiquitous:
> We have `postcss` and we have certain problems/issues, which, unfortunately, have not been resolved for a long time, like CSS compliance tokenizer and parser, selectors, at-rules and value parser

Does PostCSS's author recognize these problems? And can we not add parser improvements upstream to PostCSS?

Because Stylelint is currently a part of the PostCSS's ecosystem, I think PostCSS would be best for backward compatibility if PostCSS accepted requests for parser improvements.

romainmenke:
Thank you for bringing this up and for gathering all this info 🙇 

----------------

I think performance of PostCSS and the CSS tooling ecosystem build around PostCSS is a complicated subject :)

On the one hand PostCSS itself is really really fast. But the way it is designed means that there will be a lot of duplicate work on selectors, values, at-rule preludes, ...

It also is written in JS, so it's bound by the constraints of JS engines.
There are a lot of factors here, but in short, it will never be Rust.

Imho it isn't realistically possible to create a new parser that solves everything :
- equally fast or faster than PostCSS
- can support non-standard syntaxes (`less`, `scss`, `css-in-js`, ...)
- can support an ecosystem of plugins
- will still be fast when a lot of plugins are run
- can parse CSS in it's entirety 
- has a user friendly API surface
- has a complete and correct Object Model
- is written in JavaScript

If the constraint that is most important is performance, then it makes more sense to me that Rust or something similar is chosen as a starting point and that other aspects are sacrificed.

[...]

PR #6864, opened 2023-05-27
Fix `alpha-value-notation` performance with improved benchmark script

romainmenke:

See : https://github.com/stylelint/stylelint/issues/6853

I was curious about performance and if there were any benchmark tools in place.

I found that the current benchmark script is just checking PostCSS performance, not rule performance in specific. I've refactored this script to separate the initial parsing phase and the plugin processing.

This gives a more relevant result to Stylelint plugins.

I also found that processing Bootstrap only once doesn't give meaningful results.
PostCSS is just too fast, the deviation too high and the actual duration per cycle with or without a code change isn't clear.

I found that duplicating the Bootstrap source 20 times gives a more useful output.

```
> node scripts/benchmark-rule.mjs alpha-value-notation ["number"]

Warnings: 0
Mean: 94.41161338888888 ms
Deviation: 12.21578916466738 ms
```

```
> node scripts/benchmark-rule.mjs alpha-value-notation ["number"]

Warnings: 0
Mean: 72.47793288888887 ms
Deviation: 6.865530836418062 ms
```

I have a M2 macbook air, which is possibly the worst machine to run benchmarks on.
So someone with a decent machine (no power saving cores, cooled cpu) should verify this work.

-------

This change also adds some improvements to the first rule.
I just picked the first rule at random, not because I noticed anything wrong in terms of performance.

- add a fast abort before any parsing
- some minor code improvements

--------

Changes like these do not take away the underlying issue as described in https://github.com/stylelint/stylelint/issues/6853

But I do think it is worthwhile to look into and apply trivial improvements like these.
Even with a different style parser it will be a good practice to add fast aborts before doing a deeper AST walk.

### DIFF
--- a/lib/rules/alpha-value-notation/index.js
+++ b/lib/rules/alpha-value-notation/index.js
@@ -26,7 +26,7 @@ const meta = {
 const ALPHA_PROPS_REGEXP =
 	/^(?:opacity|shape-image-threshold|fill-opacity|flood-opacity|stop-opacity|stroke-opacity)$/i;
 const ALPHA_FUNCTION_REGEXP = /(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)\(/i;
-const ALPHA_FUNCTION_NAME_REGXP = /^(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)$/i;
+const ALPHA_FUNCTION_NAME_REGEXP = /^(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)$/i;
 const DIGIT_REGEXP = /\d/;
 
 /** @type {import('stylelint').Rule} */
--- a/lib/rules/alpha-value-notation/index.js
+++ b/lib/rules/alpha-value-notation/index.js
@@ -23,11 +23,11 @@ const meta = {
 	fixable: true,
 };
 
-const ALPHA_PROPS_REGEXP =
+const ALPHA_PROPS =
 	/^(?:opacity|shape-image-threshold|fill-opacity|flood-opacity|stop-opacity|stroke-opacity)$/i;
-const ALPHA_FUNCTION_REGEXP = /(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)\(/i;
-const ALPHA_FUNCTION_NAME_REGEXP = /^(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)$/i;
-const DIGIT_REGEXP = /\d/;
+const ALPHA_FUNCTION = /(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)\(/i;
+const ALPHA_FUNCTION_NAME = /^(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)$/i;
+const DIGIT = /\d/;
 
 /** @type {import('stylelint').Rule} */
 const rule = (primary, secondaryOptions, context) => {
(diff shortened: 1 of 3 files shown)


lib/rules/alpha-value-notation/index.js
```diff
@@ -23,27 +23,11 @@ const meta = {
...
-	'lch',
-	'oklab',
-	'oklch',
-	'rgb',
-	'rgba',
-]);
+const ALPHA_PROPS_REGEXP =
+	/^(?:opacity|shape-image-threshold|fill-opacity|flood-opacity|stop-opacity|stroke-opacity)$/i;
+const ALPHA_FUNCTION_REGEXP = /(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)\(/i;
+const ALPHA_FUNCTION_NAME_REGXP = /^(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)$/i;
```
ybiquitous, line 29: Typo:

```suggestion
const ALPHA_FUNCTION_NAME_REGEXP = /^(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)$/i;
```

ybiquitous (changes requested):
@romainmenke Thanks for the pull request. This change definitely makes sense. 👍🏼 

Let's add this to the changelog to easily track it.

---

FYI, here is my benchmark on M1 MacBook Pro and Node.js v20.2.0:

Before:

```
Warnings: 0
Mean: 105.49801229411766 ms
Deviation: 14.985993295535048 ms
```

After:

```
Warnings: 0
Mean: 84.28338958823528 ms
Deviation: 8.832248644160641 ms
```

lib/rules/alpha-value-notation/index.js
romainmenke: Good catch :)

lib/rules/alpha-value-notation/index.js
```diff
@@ -23,27 +23,11 @@ const meta = {
...
-	'lab',
-	'lch',
-	'oklab',
-	'oklch',
-	'rgb',
-	'rgba',
-]);
+const ALPHA_PROPS_REGEXP =
+	/^(?:opacity|shape-image-threshold|fill-opacity|flood-opacity|stop-opacity|stroke-opacity)$/i;
+const ALPHA_FUNCTION_REGEXP = /(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)\(/i;
```
Mouvedia, line 28: question

what's the `\(` there for ?

scripts/benchmark-rule.mjs
```diff
@@ -51,9 +52,43 @@ const processor = postcss().use(rule);
 fetch(CSS_URL)
 	.then((response) => response.text())
 	.then((response) => {
+		let css = '';
+
+		for (let i = 0; i < 20; i++) {
```
Mouvedia, line 57: suggestion

put the magic number in a variable and maybe add a comment to explain why you end up on 20

lib/rules/alpha-value-notation/index.js
romainmenke: Something like `lab` could be part of a larger word.
However `lab(` is much less likely, `\(` in this regexp matches the opening parenthesis.

_It could still have a prefix that makes it something other than the `lab()` function._

example :

```css
.foo {
	/* oklab keyword doesn't match /oklab\(/ */
	background-image: linear-gradient(in oklab 90deg, red 25% 50%, blue 70% 75%);
}

.foo {
	/* oklab function matches /oklab\(/ */
	background-image: oklab(100% 50% 50%);
}
```

scripts/benchmark-rule.mjs
romainmenke: Done, thank you suggesting this 🙇

ybiquitous (approved):
Thank you. LGTM! 👍🏼

PR #6865, opened 2023-05-27
Fix `at-rule-property-required-list` performance

romainmenke:

See:
- https://github.com/stylelint/stylelint/issues/6853
- https://github.com/stylelint/stylelint/issues/6869

In this `at-rule-property-required-list` two types of work were less than ideal :
- formatting of plugin options
- nested loops

### DIFF
--- a/lib/rules/at-rule-property-required-list/index.js
+++ b/lib/rules/at-rule-property-required-list/index.js
@@ -43,6 +43,9 @@ const rule = (primary) => {
 			propLists.set(key, new Set(propList));
 		}
 
+		/** @type {Set<string>} */
+		const currentPropList = new Set();
+
 		root.walkAtRules((atRule) => {
 			if (!isStandardSyntaxAtRule(atRule)) {
 				return;
@@ -56,8 +59,7 @@ const rule = (primary) => {
 				return;
 			}
 
-			/** @type {Set<string>} */
-			const currentPropList = new Set();
+			currentPropList.clear();
 
 			nodes.forEach((node) => {
 				if (!node || node.type !== 'decl') return;
--- a/lib/rules/at-rule-property-required-list/index.js
+++ b/lib/rules/at-rule-property-required-list/index.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const flattenArray = require('../../utils/flattenArray');
+const createMapWithSet = require('../../utils/createMapWithSet');
 const isStandardSyntaxAtRule = require('../../utils/isStandardSyntaxAtRule');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
@@ -30,18 +30,7 @@ const rule = (primary) => {
 			return;
 		}
 
-		/** @type {Map<string, Set<string>|undefined>} */
-		const propLists = new Map();
-
-		for (const key in primary) {
-			if (!Object.hasOwnProperty.call(primary, key)) continue;
-
-			const propList = flattenArray(primary[key]);
-
-			if (!propList) continue;
-
-			propLists.set(key, new Set(propList));
-		}
+		const propLists = createMapWithSet(primary);
 
 		/** @type {Set<string>} */
 		const currentPropList = new Set();
--- a/lib/rules/at-rule-property-required-list/index.js
+++ b/lib/rules/at-rule-property-required-list/index.js
@@ -50,15 +50,15 @@ const rule = (primary) => {
 
 			currentPropList.clear();
 
-			nodes.forEach((node) => {
-				if (!node || node.type !== 'decl') return;
+			for (const node of nodes) {
+				if (!node || node.type !== 'decl') continue;
 
 				const propName = node.prop.toLowerCase();
 
-				if (!propList.has(propName)) return;
+				if (!propList.has(propName)) continue;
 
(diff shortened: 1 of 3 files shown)


lib/rules/at-rule-property-required-list/index.js
```diff
@@ -30,39 +30,60 @@ const rule = (primary) => {
...
+
+		for (const key in primary) {
+			if (!Object.hasOwnProperty.call(primary, key)) continue;
+
+			const propList = flattenArray(primary[key]);
+
+			if (!propList) continue;
+
+			propLists.set(key, new Set(propList));
+		}
```
romainmenke, line 44: Should we create a utility for this?

`Record<string, string[] | string | undefined>` to `Map<string, Set<string>>`

lib/rules/at-rule-property-required-list/index.js
```diff
@@ -30,39 +30,62 @@ const rule = (primary) => {
...
+				if (!node || node.type !== 'decl') return;
+
+				const propName = node.prop.toLowerCase();
+
+				if (!propList.has(propName)) return;
+
+				currentPropList.add(propName);
+			});
+
+			if (currentPropList.size === propList.size) {
```
romainmenke, line 74: By adding prop names to a `Set` we can eliminate the nested loop.
After collecting the expected prop names it is sufficient to compare the `Set` size.

lib/rules/at-rule-property-required-list/index.js
ybiquitous: If you need it with later work, you can create a utility on this PR.

I don't think we have a utility to do the same work.

romainmenke: There are 4 rules with similar option types (`Rule<Record<string, string | string[]>>`)
Looking at those I think it will come up again.

Do you have a name suggestion?

ybiquitous: Hum, how about `createMapWithSet()`? For example:

```js
/**
 * @template T
 * @param {Record<string, T | T[]>} records
 * @returns {Map<string, Set<T>>}
 */
function createMapWithSet(records) {
	// ...
}
```

lib/rules/at-rule-property-required-list/index.js
```diff
@@ -30,39 +30,62 @@ const rule = (primary) => {
 			return;
 		}
 
+		/** @type {Map<string, Set<string>|undefined>} */
+		const propLists = new Map();
+
+		for (const key in primary) {
+			if (!Object.hasOwnProperty.call(primary, key)) continue;
+
+			const propList = flattenArray(primary[key]);
```
ybiquitous, line 39: [suggest] How about reducing the loop count by using `for...of` and `Object.entries()`?

```suggestion
		for (const [key, value] of Object.entries(primary)) {
			const propList = flattenArray(value);
```

lib/rules/at-rule-property-required-list/index.js
romainmenke: I like this!

lib/rules/at-rule-property-required-list/index.js
```diff
@@ -30,39 +30,62 @@ const rule = (primary) => {
...
-				);
+			nodes.forEach((node) => {
+				if (!node || node.type !== 'decl') return;
+
+				const propName = node.prop.toLowerCase();
+
+				if (!propList.has(propName)) return;
+
+				currentPropList.add(propName);
+			});
```
ybiquitous, line 72: [suggest] For consistency and readability, how about using `for...of` here like this?

```suggestion
			for (const node of nodes) {
				if (node.type !== 'decl') continue;

				const propName = node.prop.toLowerCase();

				if (!propList.has(propName)) continue;

				currentPropList.add(propName);
			}
```

lib/rules/at-rule-property-required-list/index.js
romainmenke: Good call

lib/rules/at-rule-property-required-list/index.js
romainmenke: Yes, much cleaner 🙇

ybiquitous (commented):
## Benchmark on my local machine

M1 MacBook Pro / Node.js v20.2.0

```shell
npm run benchmark-rule -- at-rule-property-required-list '{"font-face": ["font-display", "font-family", "font-style"], "page": "margin"}'
```

### Before

```
Warnings: 20
Mean: 81.61474988235294 ms
Deviation: 6.212360461737067 ms
```

### After

```
Warnings: 20
Mean: 82.71980147058825 ms
Deviation: 11.136367856858897 ms
```

romainmenke:
Yeah, the difference is very small.
There is only a single `@page` rule in the benchmark source, no `@font-face` rules.

I am fine with dropping this change.

I also tested with something that has more hits and even then it isn't clear which is faster :

```
node scripts/benchmark-rule.mjs at-rule-property-required-list { "media": ["width", "height"] }
```

ybiquitous (approved):
No problems. Thank you for creating this pull request! 👍🏼 

Even if the improvement is small, the new code is more readable and adds a new utility to be able to improve other rules.

PR #6868, opened 2023-05-28
Fix `color-*` performance

romainmenke:

See:
- #6853
- #6869

- added a fast abort for all `color-*` rules
- added some utilities in support of the `color-*` rules
- ensures that regexp's in utilities are declared once and then reused.

----

I don't mind splitting this up into multiple PR's, one for each rule, but the changes are all similar and minor.

I am unsure what the preference is in this project for balancing the extra noise and review work of multiple PR's vs. the clarity of small chunks. I didn't want to open ±10 PR's without prior discussion :)


Mouvedia:
What about having a `utils/interpolation.js` with 5 function exports?
e.g. `hasLess`, `hasPsv`, `hasScss`, `hasTpl`, `hasInterpolation`
If not, is it a convention to have only one export per file for utils?

lib/utils/hasNamedColor.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+const HAS_NAMED_COLOR =
```
Mouvedia, line 3: This one should be imported either from `/lib/reference/` or from a third party package.

romainmenke: Does this still apply given the context that it needs to be exactly this regexp and that this regexp is only useful for this function?

Mouvedia: My rationale is that the W3C often add new names so it either needs to be shared in reference directory so that it can be reused by other rules or we want to push the responsibility of the maintenance of the list to an npm package. 

see #6712

romainmenke: I get that, but the named colors is a pretty fixed set at this point.
`rebeccapurple` is the most recent addition and it has been in browsers since ±2014.

That is almost 10 years without a new named color :)

The system colors and color keywords are something else.

romainmenke: I've moved this to the `/lib/reference/`.

lib/reference/functions.js
```diff
@@ -16,7 +16,22 @@ const camelCaseFunctions = new Set([
...
+	'hsl',
+	'hsla',
+	'hwb',
+	'lab',
+	'lch',
+	'oklab',
+	'oklch',
+	'color',
+	'color-mix',
+]);
```
Mouvedia, line 31: It seems like it's missing `gray`.

cf
https://github.com/stylelint/stylelint/blob/40b4a6daa6168201e55eedae2acef431e216978b/lib/rules/color-named/README.md?plain=1#L22
https://github.com/stylelint/stylelint/blob/40b4a6daa6168201e55eedae2acef431e216978b/lib/rules/color-named/colordUtils.js#L22

romainmenke: `gray` is not a color function :)

I think it was proposed at some point, but it never shipped in any browser.
https://github.com/postcss/postcss-color-gray

romainmenke:
## `color-hex-alpha`

before :

```bash
git checkout main && npm run benchmark-rule color-hex-alpha always

# Warnings: 12980
# Mean: 101.40357111764703 ms
# Deviation: 13.776733967718119 ms


git checkout main && npm run benchmark-rule color-hex-alpha never 

# Warnings: 0
# Mean: 83.61844911111112 ms
# Deviation: 11.006563626355849 ms
```

after :

```bash
git checkout fix-color-rules-performance--adaptable-asian-elephant-740c875825 && npm run benchmark-rule color-hex-alpha always

# Warnings: 12980
# Mean: 81.74926383333332 ms
# Deviation: 7.977572095601296 ms


git checkout fix-color-rules-performance--adaptable-asian-elephant-740c875825 && npm run benchmark-rule color-hex-alpha never

# Warnings: 0
# Mean: 69.26526377777779 ms
# Deviation: 8.457018103395368 ms
```

lib/rules/color-function-notation/index.js
```diff
@@ -50,6 +52,10 @@ const rule = (primary, secondaryOptions, context) => {
 		const ignoreWithVarInside = optionsMatches(secondaryOptions, 'ignore', 'with-var-inside');
 
 		root.walkDecls((decl) => {
+			if (!LEGACY_NOTATION_FUNCTION.test(decl.value)) return;
+
+			if (primary === 'modern' && !COMMA.test(decl.value)) return;
```
ybiquitous, line 57: [suggest] Simply using `.includes()` seems faster and more readable than using a regex in this case. Any thoughts?

```suggestion
			if (primary === 'modern' && !decl.value.includes(',')) return;
```

lib/rules/color-function-notation/index.js
```diff
@@ -93,7 +99,7 @@ const rule = (primary, secondaryOptions, context) => {
 					});
 
 					// Remove trailing 'a' from legacy function name
-					if (LEGACY_FUNCS.has(node.value.toLowerCase())) {
+					if (LEGACY_FUNCTION_NAME.test(node.value)) {
 						node.value = node.value.slice(0, -1);
```
ybiquitous, line 103: [suggest] Trivial, but it's possible to shorten it a bit:

```suggestion
					if (LEGACY_FUNCTION_NAME.test(value)) {
						node.value = value.slice(0, -1);
```

lib/rules/color-hex-case/index.js
```diff
@@ -21,7 +21,8 @@ const meta = {
 	deprecated: true,
 };
 
-const HEX = /^#[0-9A-Za-z]+/;
+const HEX = /^#[\da-z]+/i;
+const CONTAINS_HEX = /#[\da-z]+/i;
```
ybiquitous, line 25: [add] Does it make sense to reduce characters like this?

```suggestion
const HEX = /^#[\da-f]+/i;
const CONTAINS_HEX = /#[\da-f]+/i;
```

romainmenke: These rules specifically check for `a-z`.
I was surprised by this, but didn't want to change the logic.

ybiquitous: I see. Let's keep it as-is. 👍🏼

lib/rules/color-function-notation/index.js
```diff
@@ -23,8 +23,10 @@ const meta = {
 	fixable: true,
 };
 
-const LEGACY_FUNCS = new Set(['rgba', 'hsla']);
-const LEGACY_NOTATION_FUNCS = new Set(['rgb', 'rgba', 'hsl', 'hsla']);
+const LEGACY_FUNCTION_NAME = /^(?:rgba|hsla)$/i;
+const LEGACY_NOTATION_FUNCTION = /(?:rgba?|hsla?)\(/i;
```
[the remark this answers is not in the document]
romainmenke: Good idea, thank you 🙇

lib/rules/color-no-invalid-hex/index.js
```diff
@@ -32,12 +34,16 @@ const rule = (primary) => {
...
+				return;
+			}
+
 			valueParser(decl.value).walk(({ value, type, sourceIndex }) => {
 				if (type === 'function' && value.endsWith('url')) return false;
 
 				if (type !== 'word') return;
 
-				const hexMatch = /^#[0-9A-Za-z]+/.exec(value);
+				const hexMatch = /^#[\da-z]+/i.exec(value);
```
ybiquitous, line 46: [ask] Does it make sense to extract a constant like this? Does `new RegExp()` have a performance penalty?

```js
const HEX = /^#[\da-z]+/i;
```

```suggestion
				const hexMatch = new RegExp(HEX).exec(value);
```

romainmenke: I am not sure.

I left that one as is because it also has the start anchor `^` and because it uses `exec` which is stateful if I recall correctly.

ybiquitous (commented):
[ask]

- Can you add a changelog entry?
- Will this PR close #6869?

romainmenke:
> Will this PR close

No, I've added a todo list.

ybiquitous (approved):
Thanks. LGTM 👍🏼

romainmenke:
Thank you for reviewing all this @ybiquitous 🙇

Issue #6870, opened 2023-05-28
False positives in `length-zero-no-unit` with new math functions


romainmenke:
### What steps are needed to reproduce the bug?

```css
.a {
  width: atan2(0px, 1px);
}
```

### What Stylelint configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "length-zero-no-unit": true
  }
}
```

### How did you run Stylelint?

https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0BDABMAOgOzWg7gJZQAuAFvGiiStgEwAUADAA4AeANGgIzsCUAbhwBfEBxAAzQgBsYAORQBbOIhhslLWUjABnHWPARsUgOYIQWXGkwgATgFdZOm5Ut48N2dhPkAtAC8YWwhfbBD7bEISFzQAbRIHGABdHDxhEQNIY0ITADEIW0VqcwArHSMDWBZ9RDdrEB0SAE9ZaUJsaIR66WoYRpsOVPrGlpg2jt8s019G2igUWygYz17+kAzhIA

### Which version of Stylelint or dependencies are you using?

latest

### What did you expect to happen?

No error on any of the new math functions

### What actually happened?

```
Unexpected unit (length-zero-no-unit)[2:17-19]
```

### Proposal to fix the bug

Add the new math function to `/lib/reference`

PR #6871, opened 2023-05-28
Fix `length-zero-no-unit` false positives on new math functions

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/6870

### DIFF
--- a/.changeset/calm-rivers-wink.md
+++ b/.changeset/calm-rivers-wink.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `length-zero-no-unit` false positive on new math functions
+Fixed: `length-zero-no-unit` false positives on new math functions
--- a/lib/reference/functions.js
+++ b/lib/reference/functions.js
@@ -32,6 +32,7 @@ const mathFunctions = new Set([
 	'pow',
 	'rem',
 	'round',
+	'sign',
 	'sin',
 	'sqrt',
 	'tan',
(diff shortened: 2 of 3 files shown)


lib/reference/functions.js
```diff
@@ -14,7 +14,28 @@ const camelCaseFunctions = new Set([
...
+	'atan',
+	'atan2',
+	'calc',
+	'clamp',
+	'cos',
+	'exp',
+	'hypot',
+	'log',
+	'max',
+	'min',
```
Mouvedia, line 30: Could you add `minmax` and `sign`?
https://developer.mozilla.org/en-US/docs/Web/CSS/minmax
https://developer.mozilla.org/en-US/docs/Web/CSS/sign

romainmenke: Added `sign`.

`minmax` is not a math function, so I didn't add that in this list.

.changeset/calm-rivers-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `length-zero-no-unit` false positive on new math functions
```
ybiquitous, line 5:
```suggestion
Fixed: `length-zero-no-unit` false positives on new math functions
```

ybiquitous (approved):
Thank you! LGTM 👍🏼

.changeset/calm-rivers-wink.md
romainmenke: good catch, thank you 🙇

PR #6876, opened 2023-05-29
Refactor benchmark script

ybiquitous:

None.

Summary:

- Uninstall no longer `node-fetch`.
  - Global `fetch()` has been available on Node.js current LTS versions (18.x)
    - See https://nodejs.org/api/globals.html#fetch
  - We recommend using a Node.js LTS version during development
    - See https://github.com/stylelint/stylelint/blob/a7cf5abdaa/CONTRIBUTING.md?plain=1#L18-L20
- Refactor the script to improve readability a bit.
- Use `console.error()` (stderr) for error logs.

Test (my local machine):

```console
$ npm run benchmark-rule -- alpha-value-notation '["number", {"exceptProperties": ["/opacity/"]}]'

> stylelint@15.6.2 benchmark-rule
> node scripts/benchmark-rule.mjs alpha-value-notation ["number", {"exceptProperties": ["/opacity/"]}]

Warnings: 580
Mean: 79.97806858823527 ms
Deviation: 7.444096257985445 ms
```

### DIFF
--- a/scripts/benchmark-rule.mjs
+++ b/scripts/benchmark-rule.mjs
@@ -2,16 +2,15 @@
 
 /* eslint-disable no-console */
 import Benchmark from 'benchmark';
-import fetch from 'node-fetch';
-import normalizeRuleSettings from '../lib/normalizeRuleSettings.js';
 import picocolors from 'picocolors';
 import postcss from 'postcss';
+
+import normalizeRuleSettings from '../lib/normalizeRuleSettings.js';
 import rules from '../lib/rules/index.js';
+
 const { bold, yellow } = picocolors;
 
-const ruleName = process.argv[2];
-const ruleOptions = process.argv[3];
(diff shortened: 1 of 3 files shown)


scripts/benchmark-rule.mjs
```diff
@@ -48,8 +47,7 @@ if (
 /* eslint-enable eqeqeq */
 const ruleSettings = normalizeRuleSettings(parsedOptions);
 
-const primary = ruleSettings[0];
-const secondary = ruleSettings[1] || null;
+const [primary, secondary] = ruleSettings;
```
Mouvedia, line 50: question

I suppose switching from `null` to `undefined` doesn't matter?

ybiquitous: Yes, it shouldn't have any problems. If something were wrong, the current validation logic would have problems.

PR #6726, opened 2023-03-22
Fix `unit-no-unknown` false positives for the second and subsequent `image-set()` with `x` descriptor

chalkygames123:
> Which issue, if any, is this issue related to?

Fixes #6721

> Is there anything in the PR that needs further explanation?

I'll do some refactoring later on, as @ybiquitous [suggested](https://github.com/stylelint/stylelint/issues/6721#issuecomment-1477831540) (it may require a few weeks).

PR #6879, opened 2023-05-31
Fix `unit-no-unknown` false positives for the second and subsequent `image-set()` with `x` descriptor

romainmenke:

Supersedes : https://github.com/stylelint/stylelint/pull/6726

- ~~this will fail because it uses unreleased features of the media query parser~~
- also happens to resolve #6726

before :

```
npm run benchmark-rule unit-no-unknown true
Warnings: 0
Mean: 143.3089323125 ms
Deviation: 10.593647978441565 ms
```

after : 

```
npm run benchmark-rule unit-no-unknown true
Warnings: 0
Mean: 109.86287500000002 ms
Deviation: 9.320724134887374 ms
```

### DIFF
--- a/lib/rules/unit-no-unknown/index.js
+++ b/lib/rules/unit-no-unknown/index.js
@@ -1,21 +1,28 @@
 'use strict';
 
+const { tokenize, TokenType } = require('@csstools/css-tokenizer');
+const {
+	parseCommaSeparatedListOfComponentValues,
+	isFunctionNode,
+	isSimpleBlockNode,
+	isTokenNode,
+} = require('@csstools/css-parser-algorithms');
+const { isMediaFeature, parseFromTokens } = require('@csstools/media-query-list-parser');
+
 const atRuleParamIndex = require('../../utils/atRuleParamIndex');
 const declarationValueIndex = require('../../utils/declarationValueIndex');
-const getDimension = require('../../utils/getDimension');
+const getAtRuleParams = require('../../utils/getAtRuleParams');
+const getDeclarationValue = require('../../utils/getDeclarationValue');
 const isStandardSyntaxAtRule = require('../../utils/isStandardSyntaxAtRule');
 const isStandardSyntaxDeclaration = require('../../utils/isStandardSyntaxDeclaration');
-const { units } = require('../../reference/units');
-const mediaParser = require('postcss-media-query-parser').default;
+const isStandardSyntaxValue = require('../../utils/isStandardSyntaxValue');
 const optionsMatches = require('../../utils/optionsMatches');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
 const validateOptions = require('../../utils/validateOptions');
-const valueParser = require('postcss-value-parser');
 const vendor = require('../../utils/vendor');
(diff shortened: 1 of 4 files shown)


romainmenke:
@ybiquitous I intended to just change the parser here, but this change also had the side effect of fixing an issue.

I propose changing the title and adding the correct changelog for that:

```md
Fix `unit-no-unknown` false positives for the second and subsequent `image-set()` with `x` descriptor
```

I would also add `chalkygames123` as a co-author as they wrote the tests that I used to verify the fix in https://github.com/stylelint/stylelint/pull/6726

ybiquitous:
Yes, that's a good idea on https://github.com/stylelint/stylelint/pull/6879#issuecomment-1572262355 

Please update them!

lib/rules/unit-no-unknown/index.js
```diff
@@ -48,122 +58,149 @@ const rule = (primary, secondaryOptions) => {
...
+			if (!tokens) return;
 
-			check(atRule, atRule.params, atRuleParamIndex);
+			parseFromTokens(tokens).forEach((mediaQuery) => {
+				const state = {
+					ignored: false,
+					allowX: false,
+				};
+
+				mediaQuery.walk((entry) => {
```
ybiquitous, line 149: [ask] Can `entry.state` be empty? If no, I guess we could simplify the code by early return like this:

```suggestion
				mediaQuery.walk((entry) => {
					if (!entry.state) return;
```

lib/rules/unit-no-unknown/index.js
```diff
@@ -48,122 +58,149 @@ const rule = (primary, secondaryOptions) => {
...
+
+					if (isMediaFeature(entry.node)) {
+						const name = entry.node.getName();
+
+						if (entry.state && RESOLUTION_MEDIA_FEATURE_NAME.test(name)) {
+							entry.state.allowX = true;
+						}
+					}
+
+					if (isFunctionNode(entry.node) || isTokenNode(entry.node)) {
```
ybiquitous, line 160: [ask] Does converting to `else if` make sense to reduce extra evaluations?

```suggestion
					} else if (isFunctionNode(entry.node) || isTokenNode(entry.node)) {
```

lib/rules/unit-no-unknown/index.js
romainmenke: It should never be empty unless there is a bug or if a parser is used which doesn't have support for this.

lib/rules/unit-no-unknown/index.js
romainmenke: yes! thanks 🙇

ybiquitous (approved):
Thank you. LGTM 👍🏼

ScorpAL:
HI Guys.
Something wrong after update from 15.6.2 to 15.6.3 

```css
.test {
    margin-bottom: 20zrr;
}
```
With **stylelint 15.6.2**
CLI and VSCode reported:
`Unexpected unknown unit "zrr"  unit-no-unknown`

With **stylelint 15.6.3**
CLI:
no errors reported
VSCode:
`Unexpected unknown unit "zrr" (unit-no-unknown)Stylelint[unit-no-unknown](https://stylelint.io/user-guide/rules/unit-no-unknown)`

Issue #6880, opened 2023-06-01
Set a timeout for GitHub Workflows


romainmenke:
see : https://github.com/stylelint/stylelint/actions/runs/5136149333/jobs/9242517643?pr=6879

This workflow job ran for 6 hours.
99% sure this was a glitch unrelated to the actual change.

Setting a maximum duration for workflow jobs should help with cases like these.
Preventing accidental usage of resources is good for the environment :)

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

ybiquitous:
@romainmenke Nice catch! Please feel free to open a pull request. 😄

PR #6881, opened 2023-06-01
Fix: add timeouts for GitHub Workflows

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/6880

CI for Stylelint seems to take ±6 minutes on most runs, sometimes a little bit over 10 minutes.

30 minutes seems like a safe multiple of the current duration.
Sufficient headroom so that no one will ever encounter this timeout during normal operations.

I've lowered this to 20 minutes for jobs that don't do any heavy lifting.

### DIFF
--- a/.github/workflows/code-scanning.yml
+++ b/.github/workflows/code-scanning.yml
@@ -20,6 +20,7 @@ jobs:
   analyze:
     name: Analyze
     runs-on: ubuntu-latest
+    timeout-minutes: 30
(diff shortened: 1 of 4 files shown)


ybiquitous (approved):
Thanks, LGTM! 👍🏼 

It's also good to set timeout to our usable workflows:
https://github.com/stylelint/.github/tree/main/.github/workflows

PR #6886, opened 2023-06-03
Refactor to migrate  to new media query parser

romainmenke:


lib/rules/unit-disallowed-list/index.js
```diff
@@ -96,75 +113,108 @@ const rule = (primary, secondaryOptions) => {
...
-					// the values not included in the parentValue string
-					if (valueNode.type !== 'word' || !parentValue.includes(valueNode.value)) {
-						return;
+		root.walkAtRules(/^media$/i, (atRule) => {
+			const params = getAtRuleParams(atRule);
+
+			if (!hasDimension(params)) return;
+
+			parseFromTokens(tokenizeWithoutPercentageTokens(params)).forEach((mediaQuery) => {
+				/** @type {{ mediaFeatureName?: string }} */
```
ybiquitous, line 122: [suggest] The non-optional property seems a bit more accurate in this case, although it's trivial.

```suggestion
				/** @type {{ mediaFeatureName: string | undefined }} */
```

lib/rules/unit-disallowed-list/index.js
```diff
@@ -96,75 +113,108 @@ const rule = (primary, secondaryOptions) => {
...
 					check(
-						node,
-						getIndex(node),
-						valueNode,
-						mediaName,
-						secondaryOptions ? secondaryOptions.ignoreMediaFeatureNames : {},
+						atRule,
+						atRuleParamIndex,
+						entry.node,
+						entry.state?.mediaFeatureName,
```
ybiquitous, line 140: [ask] Does passing `mediaFeatureName` to `check()` make sense if `mediaFeatureName` is empty? Can we skip it like this?

```js
if (!entry.state.mediaFeatureName) return;

check(
  // ...
  entry.state.mediaFeatureName,
  // ...
);
```

lib/rules/unit-disallowed-list/index.js
```diff
@@ -96,75 +113,108 @@ const rule = (primary, secondaryOptions) => {
...
+
+					componentValue.walk((entry) => {
+						if (!entry.state) return;
 
-		root.walkAtRules(/^media$/i, (atRule) => checkMedia(atRule, atRule.params, atRuleParamIndex));
-		root.walkDecls((decl) => checkDecl(decl, decl.value, declarationValueIndex));
+						if (entry.state.ignored) return;
+
+						if (componentValueIsIgnored(entry.node)) {
+							entry.state.ignored = true;
```
ybiquitous, line 180: [ask] Can we add a case to cover L180?

> Added line #L180 was not covered by tests

ybiquitous (commented):
@romainmenke Thanks for the pull request! This is almost good. But could you confirm my reviews, please?

lib/rules/unit-disallowed-list/index.js
romainmenke: Yes we can skip it!
Good catch 🙇

lib/rules/unit-disallowed-list/index.js
romainmenke: This led me to discover : https://github.com/stylelint/stylelint/issues/6888
I had to do some rewrites to implement it correctly.

romainmenke: Phew, done 😅

lib/rules/unit-disallowed-list/index.js
```diff
@@ -96,75 +113,109 @@ const rule = (primary, secondaryOptions) => {
...
-		root.walkDecls((decl) => checkDecl(decl, decl.value, declarationValueIndex));
+					if (!isFunctionNode(componentValue) && !isSimpleBlockNode(componentValue)) return;
+
+					componentValue.walk(
+						(entry) => {
+							if (!entry.state) return;
+
+							if (entry.state.ignored) return;
+
+							if (isTokenNode(entry.node) && !entry.state.ignored) {
```
ybiquitous, line 176: [ask] Does this `!entry.state.ignored` duplicate `if (entry.state.ignored) return;` at L174? Can we remove the check?

```suggestion
							if (isTokenNode(entry.node)) {
```

romainmenke:
This also requires a check for `<urange>`.
See : https://github.com/stylelint/stylelint/pull/6892

ybiquitous (approved):
Thank you. LGTM 👍🏼

romainmenke:
Thank you for all the reviews @ybiquitous  🙇

Issue #6891, opened 2023-06-04
Fix `unit-no-unknown` false positives for `unicode-range` descriptors


Mister-Hope:
## Reproduction

https://stylelint.io/demo/#N4Igxg9gJgpiBcID0AqABAGwIYBcCWAdgLQwAeOaKSAOgQAIBmEBORDWYMawtaaTLIgHcYeAOYALHPDQAWAAzyA3L37NWAZxwBPDDBl4cWDHjAqCfAa3YBbPBm0yAMrgjm+GgE5gZGCGGMACmoQFxwINABJIxMwEIBKABpMfyCQsIgiaONTBMTVPgBXTwxgkCkcAAcNeCQkKw0AOjEtXFNGyBskDSRscKQANwBGAHYkAGUANkKADiFZACEAWQBVABUABQAJACtSDQBBFbwiAEcAKQBhAE1GoQgGBgAmBIK+NU8bXDL7x5eQeLmVSFAimaAwIieLAEMT6NArADU8iGiiI8iesgAYslEeiAKwAThxCKGAFFUWTMdj4Qinuj5MS6Qd5EQmQt8hZ3oimQARVnyS7U7lDIYAZkZl0mLKelxGQoRBxGdKIiqp5gAvrQQIkQAx7DAAHJYGxwRBkY2VPQdDQabXgZh6sQIEA8TkhMg4GAEKAaEIyADabxCWl0MBMgkgBEdRFa3qwnigIVUAF1aOq7ZHHZiIJ9cM6dhpmHbYNVna6+MGdHpwzg-WgQn0YFoQhyKyAQ9XCKxM+IY0Y4wm6w3cE3ayA0yB1UA

### Which version of Stylelint or dependencies are you using?

15.6.3

### What did you expect to happen?

No problem

### What actually happened?

error

### Does the bug relate to non-standard syntax?

no

Mister-Hope:
Probably caused by https://github.com/stylelint/stylelint/commit/7206cfc12e61d5f3af9308ad960844655b01f666

romainmenke:
Thank you for reporting this @Mister-Hope 

I can confirm the issue.

ybiquitous:
@Mister-Hope @romainmenke Thanks for the report and the quick-fix!

PR #6892, opened 2023-06-04
Fix `unit-no-unknown` false positives for `unicode-range` descriptors

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/6891

I forgot about [`<urange>` productions](https://www.w3.org/TR/css-syntax-3/#urange) when migrating rules to `@csstools/css-tokenizer`.

Parsing `<urange>` requires special handling and should only be done when inside a context that expects such values.

I've added a helper (`isUnicodeRangeDescriptor`) to make it easier for rules to skip `@font-face { unicode-range: U+0100-024F; }`

### DIFF
--- a/.changeset/shiny-toys-chew.md
+++ b/.changeset/shiny-toys-chew.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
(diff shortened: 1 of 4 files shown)


ybiquitous (approved):
Thank you! LGTM 👍🏼

PR #6896, opened 2023-06-05
Add `splitList: boolean` to `selector-nested-pattern`

### LINKED ISSUE #2430: Add splitList: bool to selector-nested-pattern

A bug

`selector-nested-pattern`

```css
.baz {
  .foo,
  .bar {
    color: pink;
  }
}
```

e.g.

```json
{
  "rules": {
    "selector-nested-pattern": "^.*$"
  }
}
```

e.g. `7.1.0`

CLI with `stylelint --config myconfig *.css`

Yes, it's related to postcss/scss nesting

No warnings to be flagged.

```shell
Expected nested selector ".foo,   .bar" to match specified pattern   selector-nested-pattern
```

is2ei:

Closes #2430 

As @jeddy3 said, we could consider a better name for the option.

https://github.com/stylelint/stylelint/issues/2430#issuecomment-1160156212
> (Prehaps there's a better name?. Prior art is [the disallowInList option](https://stylelint.io/user-guide/rules/list/no-duplicate-selectors/#disallowinlist-true--false-default-false). [Spec ref for "list"](https://www.w3.org/TR/selectors-4/#grouping).)

I guess `splitList` is not bad because `selector-disallowed-list` rule has [similar option](https://stylelint.io/user-guide/rules/selector-disallowed-list/#splitlist-true--false-default-false).

### DIFF
--- a/lib/rules/selector-nested-pattern/index.js
+++ b/lib/rules/selector-nested-pattern/index.js
@@ -4,7 +4,7 @@ const isStandardSyntaxRule = require('../../utils/isStandardSyntaxRule');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
 const validateOptions = require('../../utils/validateOptions');
-const { isRegExp, isString } = require('../../utils/validateTypes');
+const { isBoolean, isRegExp, isString } = require('../../utils/validateTypes');
 
 const ruleName = 'selector-nested-pattern';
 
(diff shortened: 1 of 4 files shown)


lib/rules/selector-nested-pattern/index.js
```diff
@@ -39,20 +51,35 @@ const rule = (primary) => {
...
-			});
+				report({
+					result,
+					ruleName,
+					message: messages.expected,
+					messageArgs: [selector, primary],
+					node: ruleNode,
+					word: selector,
+				});
+			}
```
ybiquitous, line 82: [suggest] How about reducing code duplication like this?

```js
const selectors = splitList ? ruleNode.selectors : [ruleNode.selector];

for (const selector of selectors) {
	if (normalizedPattern.test(selector)) {
		continue;
	}

	report({ /*...*/ });
}
```

ybiquitous (commented):
@is2ei Thanks for the pull request! Can you check my reviews?

lib/rules/selector-nested-pattern/index.js
is2ei: @ybiquitous 

Thanks for your comment!
I updated the change.

https://github.com/stylelint/stylelint/pull/6896/commits/33239f984d6606fed515836f656ecda43a6fa93b

ybiquitous (approved):
Thank you! LGTM 👍🏼

Issue #6575, opened 2023-01-13
Add `ratio-notation`


Mouvedia:
### What is the problem you're trying to solve?

Iv recently discovered that Chrome didn't support decimals in ratio data types, but Firefox did.
e.g. `1.7/1` fails but `10/6` is OK

This is probably due to the different types accepted between Media Queries Level 3 and 4.
ref: https://developer.mozilla.org/en-US/docs/Web/CSS/ratio#syntax

### What solution would you like to see?

- Name: `ratio-notation`
- Primary option: `"number"|"integer"`
- Secondary options: none
- Autofixable: Not initially
- Message: `Expected "${primary}" ratio notation`
- Description: "Specify number or integer notation for ratio data type."
- Section: "Enforce conventions" -> "Notation"

A `foo % 1 === 0` condition should be enough to detect it.

Mouvedia:
>  The *-notation rules are for when the same thing can be represented in two or more ways…

That's the case here.
e.g. `1.5/3` === `3/6`

Depending on the level of Media Queries it's supported or not.
It's akin to the selector-not-notation rule that I added previously.

romainmenke:
This PostCSS plugin fixes the browser issues : https://github.com/csstools/postcss-plugins/tree/postcss-preset-env--v8/plugins/postcss-media-queries-aspect-ratio-number-values#readme

Support stats : https://mrhenry.github.io/web-tests/#de023e11-f61a-4635-a74c-e4a152251bdf

Stylelint could have a rule for it, but since it is perfectly possible to polyfill I think it might make sense to do just that :)

Mouvedia:
@romainmenke I am not using PostCSS, so that wouldn't be a solution _for me_.
It's reassuring to see that I am not the only one who saw the need for a fix, though.

FloEdelmann:
I think that rule should not exist to ensure browser compatibility, but to enforce a convention across the codebase:

---

### Primary option `"number"`

> [!NOTE]
> Can always be autofixed by calculating the division.

The following patterns are considered problems:

```css
a { aspect-ratio: 2 / 1; }
a { aspect-ratio: 3 / 2; }
a { aspect-ratio: 16/9; }
a { aspect-ratio: 1.7/1; }
```

The following patterns are *not* considered problems:

```css
a { aspect-ratio: 2; }
a { aspect-ratio: 1.5; }
a { aspect-ratio: 1.7778; }
a { aspect-ratio: 1.7; }
```

[...]

Issue #3026, opened 2017-11-17
unit "x" (image-set)


strarsis:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A new unit to be added to defaults.

> Which rule, if any, is this issue related to?

`unit-no-unknown`

> What CSS is needed to reproduce this issue?

e.g.

```css
.foo {
  background-image:
    image-set(
      url(../images/background-blended.jpg?size=400&quality=10) 1x,
      url(../images/background-blended.jpg?size=800&quality=10) 2x,
      url(../images/background-blended.jpg?size=1200&quality=10) 3x
    );
  background-position: top center;
  background-repeat: no-repeat;
}
```

> What stylelint configuration is needed to reproduce this issue?

e.g.

```json
{
  "extends": "stylelint-config-standard",
    "rules": {
      "no-empty-source": null
    }
}
```

> Which version of stylelint are you using?

e.g. `7.12.0`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

webpack loader

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Unsure.

> What did you expect to happen?

No errors to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

Lint error:
````
Unexpected unknown unit "x"   unit-no-unknown
Unexpected unknown unit "x"   unit-no-unknown
Unexpected unknown unit "x"   unit-no-unknown
````

```shell
test.css
 4:65  ✖  Unexpected unknown unit "x"   unit-no-unknown
 5:65  ✖  Unexpected unknown unit "x"   unit-no-unknown
 6:66  ✖  Unexpected unknown unit "x"   unit-no-unknown
```

I want to use the `image-set` syntax (PostCSS polyfill https://github.com/SuperOl3g/postcss-image-set-polyfill).
In case this is deemed to be not standard enough for being added to default config, what would be the best practice for adding this unit as correct to my stylelint configuration?

CAYdenberg:
Hi @strarsis did you possibly forget the first few questions in the template? Just having trouble following and the answers are still marked "e.g."

Second question for you as I have not used this feature: do you think it would be accurate to consider "x" a unit?

strarsis:
@CAYdenberg: Edited.
`image-set` is a new feature and it uses the `x` unit - hence I think it makes sense to add it to the whitelisted units.

CAYdenberg:
[...]

PR #6906, opened 2023-06-06
Add `media-feature-name-value-no-unknown`

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/6560

No, it's self-explanatory.

------

Tested this with a large monorepo containing ±50 websites and got no false positives and it tracked down a real error :

```diff
- @media (prefers-contrast: high) {
+ @media (prefers-contrast: more) {
```


lib/rules/media-feature-name-value-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,60 @@
...
+		},
+		{
+			code: '@media (color : 1) {}',
+			description: 'Integer value',
+		},
+		{
+			code: '@media screen and (min-width <= 768px) {}',
+			description: 'Range context, media feature in allowed list',
+		},
+	],
```
alexander-akait, line 38: Can we add `-webkit` and `-moz` (do we have other) tests?

romainmenke: Do we want to ignore or validate any vendor prefixed media feature?

I don't have any data for the values of prefixed media features, so my initial reflex is to just ignore these.

I don't have a strong opinion on this.

----

Should be covered by tests either way :)

alexander-akait: @romainmenke Ideally - we should normalize them and validate, `no-vendor-prefixes` should catch them too, but I think it can be done in other PRs

romainmenke: I've added normalization for prefixed idents in feature names en feature values.

lib/reference/mediaFeatures.js
```diff
@@ -6,25 +6,27 @@ const deprecatedMediaFeatureNames = new Set([
...
+	'vertical-viewport-segments',
 	'width',
 ]);
 
-const mediaFeatureNames = uniteSets(deprecatedMediaFeatureNames, rangeTypeMediaFeatureNames, [
+const rangeTypeMediaFeatureNamesWithMinMaxPrefix = new Set(
+	[...rangeTypeMediaFeatureNames].flatMap((name) => {
+		return [`min-${name}`, `max-${name}`];
+	}),
+);
```
romainmenke, line 27: I like have this automatically follow `rangeTypeMediaFeatureNames` but I don't mind changing this to a hard coded list.

lib/rules/media-feature-name-value-no-unknown/__tests__/index.js
Mouvedia: minimum: ignore by default
better: validate based on the value after the prefix
ideal: based on `media-feature-name-no-vendor-prefix` current config, validate using a list (non existing vendor-prefixed names should be detected)

romainmenke: > ideal: based on media-feature-name-no-vendor-prefix current config, validate using a list (non existing vendor-prefixed names should be detected)

I don't think the config of other rules is available 🤔 
Making it available would also open the door for spooky action at a distance.

Vendor prefixes are also a bit of an obsolete issue.
Engines no longer roll out new features with a vendor prefix.
All existing prefixes will eventually die out.

Mouvedia: > Engines no longer roll out new features with a vendor prefix.

Do you have a source for this?

romainmenke: https://developer.mozilla.org/en-US/docs/Glossary/Vendor_Prefix

lib/reference/mediaFeatures.js
Mouvedia: Does `min-horizontal-viewport-segments` and `max-horizontal-viewport-segments` exist?
If not can you add a test for `max-horizontal-viewport-segments`?

lib/rules/media-feature-name-value-no-unknown/index.js
```diff
@@ -0,0 +1,314 @@
...
+const meta = {
+	url: 'https://stylelint.io/user-guide/rules/media-feature-name-value-no-unknown',
+};
+
+const resolutionUnits = new Set([
+	// Resolution
+	'dpi',
+	'dpcm',
+	'dppx',
+	// "x" is omitted in "reference/units".
```
Mouvedia, line 42: minor

we maybe want to add it there and create a separate set so that it can be imported here
i.e. it's probably a false negative for `unit-no-unknown`
not sure if it has to be done in this PR though

lib/reference/mediaFeatures.js
romainmenke: `min-` and `max-` prefixes are required to work in browsers for all `range` type features :

- https://www.w3.org/TR/mediaqueries-5/#mq-min-max
- https://www.w3.org/TR/mediaqueries-5/#media-descriptor-table

> `horizontal-viewport-segments` `<integer>` `range`

-----

Is there anything specific about `horizontal-viewport-segments` that would make it interesting to have extra test coverage?

From a syntax/grammar standpoint this is the same as the `color` feature.

lib/rules/media-feature-name-value-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,196 @@
...
+		{
+			code: '@media (color : 1) {}',
+			description: 'Integer value',
+		},
+		{
+			code: '@media (aspect-ratio : 1 / 1) {}',
+			description: 'Aspect ratio value',
+		},
+		{
+			code: '@media (aspect-ratio : 1.2) {}',
```
Mouvedia, line 47: related: #6575

lib/rules/media-feature-name-value-no-unknown/index.js
romainmenke: Yes, I don't like that I had to add it here.

It might make more sense for rules that want to ignore `x` to create a copy of the `Set` and then call `unitsCopy.delete('x')`

That would make the exception local and explicit.

Other rules that do want to have the full and complete set of units would then be able to use `reference/units`.

lib/rules/media-feature-name-value-no-unknown/__tests__/index.js
romainmenke: It is indeed possible to resolve that issue by adding a secondary rule option.
Something to add extra validation for "only integers separated by a `/` form a valid aspect ratio"

Although a dedicated rule would likely be less complex overal.
Adding it here opens the door for validation checks for all things that were invalid at some point in the past.

lib/reference/mediaFeatures.js
Mouvedia: Here's [some context](https://github.com/w3c/csswg-drafts/issues/5621#issuecomment-713142655) for the ones interested.

lib/rules/media-feature-name-value-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,196 @@
+'use strict';
+
+const { messages, ruleName } = require('..');
+
+testRule({
+	ruleName,
+	config: [true],
+
+	accept: [
```
Mouvedia, line 9: major

Could you add a test for `screen-spanning`?

browser: edge
see: https://daverupert.com/2020/07/gettin-foldy-with-the-dual-screen-web-part-i/

Mouvedia:
There are probably some cases that could covered with a test in [this list](https://www.codeinwp.com/blog/how-to-use-css-media-queries/#less-common-media-features).

lib/rules/media-feature-name-value-no-unknown/__tests__/index.js
romainmenke: `screen-spanning` is an unknown media feature and those are ignored by this rule.

`media-feature-name-no-unknown` already exists to cover those issues.

ybiquitous (changes requested):

lib/rules/media-feature-name-value-no-unknown/index.js
```diff
@@ -0,0 +1,314 @@
...
+const {
+	mediaFeatureNameAllowedValueKeywords,
+	mediaFeatureNameAllowedValueTypes,
+	mediaFeatureNames,
+} = require('../../reference/mediaFeatures');
+
+const ruleName = 'media-feature-name-value-no-unknown';
+
+const messages = ruleMessages(ruleName, {
+	rejected: (name, value) => `Unexpected value "${value}" for name "${name}"`,
```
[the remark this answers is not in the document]
romainmenke: sounds good :)

lib/rules/media-feature-name-value-no-unknown/__tests__/index.js
romainmenke: > So, how about continuing as is?

I started out by just ignored everything with a vendor prefix.
So a vendor prefixed media feature name with an invalid value type would fine, it would not give errors.

I changed this to stripping prefixes, so any vendor prefixed media feature name would be treated as if it were unprefixed.

Do I leave it as is, strip the prefixes, or do I revert to the original code?
I don't have a strong preference between these two. :)

lib/rules/media-feature-name-value-no-unknown/index.js
ybiquitous: adding `resolutionUnits` to `reference/units` sounds good to me. 👍🏼 
Can you do it, please?

lib/rules/media-feature-name-value-no-unknown/index.js
```diff
@@ -0,0 +1,318 @@
...
+					if (isMediaFeature(entry.node)) {
+						const mediaFeatureNameRaw = entry.node.getName();
+						let mediaFeatureName = vendor.unprefixed(mediaFeatureNameRaw.toLowerCase());
+
+						// Unknown media feature names are handled by "media-feature-name-no-unknown".
+						if (!mediaFeatureNames.has(mediaFeatureName)) return;
+
+						if (HAS_MIN_MAX_PREFIX.test(mediaFeatureName)) {
+							mediaFeatureName = mediaFeatureName.slice(4);
+						}
```
ybiquitous, line 262: [suggest] Does it make sense to reactor it like this?

```suggestion
						mediaFeatureName = mediaFeatureName.replace(HAS_MIN_MAX_PREFIX, '');
```

lib/rules/media-feature-name-value-no-unknown/index.js
```diff
@@ -0,0 +1,318 @@
...
+				}, state);
+			});
+		});
+	};
+};
+
+rule.ruleName = ruleName;
+rule.messages = messages;
+rule.meta = meta;
+module.exports = rule;
```
ybiquitous, line 284: [suggest] Can you please move the lines about exporting to the bottom of the file to easily find the exporting?

lib/rules/media-feature-name-value-no-unknown/index.js
```diff
@@ -0,0 +1,318 @@
...
+				report({
+					message: messages.rejected,
+					messageArgs: [state.mediaFeatureNameRaw, valuePart],
+					index: atRuleParamIndexValue + start,
+					endIndex: atRuleParamIndexValue + end,
+					node: atRule,
+					ruleName,
+					result,
+				});
+			};
```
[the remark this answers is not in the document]
romainmenke: `State` is the walker state and that is already a fairly complex concept.
It has a very specific way of being mutated.

Whereas `atRule` is a "constant" during the entire AST traversal.

-----

Maybe this :

- rename `State` to `WalkerState`
- pass `atRule` as an argument to each `check` function
- pass `atRule` as an argument to `reporter`

That makes it explicit that `State` is an object that is suitable for all types of data and also avoids the dual method of passing data (capture vs arguments)

lib/rules/unit-no-unknown/index.js
```diff
@@ -24,7 +24,12 @@ const ruleMessages = require('../../utils/ruleMessages');
 const validateOptions = require('../../utils/validateOptions');
 const vendor = require('../../utils/vendor');
 const { isRegExp, isString } = require('../../utils/validateTypes');
-const { units } = require('../../reference/units');
+const units = new Set(require('../../reference/units').units); // a copy that is safe to mutate
+
+// `x` as a resolution unit is very often a typo for `px`.
+// By removing it from the set of known units, we can catch those typos.
+// Intentional `x` units are supported by manually checking these in specific functions or properties.
+units.delete('x');
```
romainmenke, line 32: see : https://github.com/stylelint/stylelint/pull/6906#discussion_r1225933858

Ignoring `x` is very specific to this one rule.
If we want to have a re-usable reference for `resolutionUnits` and if `resolutionUnits` are included in `units`, then we must move the removal of `x` to this rule.

I think this is fine because it makes the exception local to the rule that requires it.

Mouvedia: see also: https://github.com/stylelint/stylelint/issues/3026#issuecomment-350514289

lib/rules/media-feature-name-value-no-unknown/index.js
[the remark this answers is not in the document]
ybiquitous: Okay, let's leave it as-is 👍🏼

lib/reference/units.js
```diff
@@ -62,7 +62,10 @@ const lengthUnits = new Set([
 	'cqmax',
 ]);
 
-const units = uniteSets(lengthUnits, [
+// lib/reference/units.js
```
ybiquitous, line 65: Oops.

```suggestion
```

lib/rules/media-feature-name-value-no-unknown/index.js
```diff
@@ -0,0 +1,308 @@
...
+			}
+		}
+
+		root.walkAtRules(/^media$/i, (atRule) => {
+			/**
+			 * @type {Reporter}
+			 */
+			const reporter = (state, valuePart, start, end) => {
+				const atRuleParamIndexValue = atRuleParamIndex(atRule);
+
```
ybiquitous, line 223: [suggest] Does it make sense to store `atRuleParamIndexValue` like `atRule` in a closure to reduce calculation?

```suggestion
			const atRuleParamIndexValue = atRuleParamIndex(atRule);

			/** @type {Reporter} */
			const reporter = (state, valuePart, start, end) => {
```

lib/rules/media-feature-name-value-no-unknown/__tests__/index.js
```diff
@@ -0,0 +1,308 @@
...
+
+	reject: [
+		{
+			code: '@media screen and (min-width: 1000khz) {}',
+			description: 'Frequency value when only lengths are allowed',
+			message: messages.rejected('min-width', '1000khz'),
+			line: 1,
+			column: 31,
+			endLine: 1,
+			endColumn: 37,
```
ybiquitous, line 175: [ask] I think `endColumn` should be `38`, not `37`... 🤔 Is it expected?
(`"1000khz".length === endColumn - column`)

lib/rules/media-feature-name-value-no-unknown/index.js
romainmenke: I went back on forth on this.
I think your proposed change would reduce calculations when the number of warnings is more than 1, right?

The current code reduces calculations when there are no warnings at all.
I think this will lead to fewer calculations overall, but happy to make the change :)

lib/rules/media-feature-name-value-no-unknown/__tests__/index.js
romainmenke: No, not expected :)

Thank you so much for catching this.
`endIndex` needs a `+ 1`

lib/rules/media-feature-name-value-no-unknown/index.js
ybiquitous: > The current code reduces calculations when there are no warnings at all.

Okay, leaving it as-is makes sense. Thanks for the explanation! 👍🏼

ybiquitous (approved):
Thanks for the great work. LGTM 👏🏼

ybiquitous:
This new rule might be in our recommended config when the rule is mature. 😄

PR #6910, opened 2023-06-08
Support `.mjs` configuration files

### LINKED ISSUE #6905: Support `.mjs` configuration files

Cosmiconfig has [supported `.mjs` files since v8.2](https://github.com/cosmiconfig/cosmiconfig/blob/v8.2.0/CHANGELOG.md). I believe it'd be useful if Stylelint also could support it.

First, we may need to resolve a problem with loaders. It may not be easy.

https://github.com/stylelint/stylelint/blob/2413e8f186015a0a546ab1fa69c077a196141dd0/lib/createStylelint.js#L22-L27

Then, we must update the CLI help and documentation below:

https://github.com/stylelint/stylelint/blob/2413e8f186015a0a546ab1fa69c077a196141dd0/lib/cli.js#L115-L118

https://github.com/stylelint/stylelint/blob/2413e8f186015a0a546ab1fa69c077a196141dd0/docs/user-guide/configure.md?plain=1#L3-L8

See also:
- #6898
- #6902
- #5291

ybiquitous:

Closes #6905

This PR mainly does the following:

- Update the CLI help.
- Update the user guide.

In addition, this PR improves files in the `scripts` folder so that we can easily try the `.mjs` support. Try running the following at the project root.

```console
$ (cd scripts ; ./visual.sh)
```

### DIFF
--- a/.changeset/swift-pugs-matter.md
+++ b/.changeset/swift-pugs-matter.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: support for `.mjs` configuration files
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -5,7 +5,7 @@ Stylelint expects a configuration object, and looks for one in a:
 - `stylelint` property in `package.json`
 - `.stylelintrc` file
 - `.stylelintrc.{cjs,js,json,yaml,yml}` file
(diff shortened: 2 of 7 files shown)


romainmenke (approved):
LGTM!

I like the new way of describing which config files are supported 🎉

Issue #6920, opened 2023-06-15
Fix `function-no-unknown` false positives for SCSS functions with namespace


re757575:
### File

```scss
// src/assets/sass/utils/_functions.scss

@use 'sass:math';

$base-font-size: 16;

// px to rem
@function rem($px) {
  @return math.div($px, $base-font-size) + rem;
}
```

```scss
// src/assets/sass/utils/_index.scss

@forward 'functions' as f-*;
```

```scss
// src/App.scss

@use "/src/assets/sass/utils" as u;

#root {
  border: u.f-rem(20) solid red;
}
```

### Run stylelint 
Will got Error: Unexpected unknown function "u.f-rem"  scss/function-no-unknown

### Demo
https://codesandbox.io/p/sandbox/stylelint-test-function-no-unknown-t5ylc3?file=%2Fsrc%2FApp.scss

### What Stylelint configuration is needed to reproduce the bug?

```json
{
  "plugins": ["stylelint-scss"],
  "extends": ["stylelint-config-standard-scss"],
  "rules": {
    "scss/function-no-unknown": [
      true,
      {
        "ignoreFunctions": ["/^u\\.f-/"]
      }
    ]
  }
}
```

### How did you run Stylelint?

```shell
stylelint '**/*.{css,scss}'
```

### Which version of Stylelint or dependencies are you using?

`>= 15.6.2`

### What did you expect to happen?

The 15.6.1 version does not have this error, and the version after updating to 15.6.2 has an error.
Or how do I need to adjust `ignoreFunctions`?

### What actually happened?

```
src/App.scss
 4:13  ✖  Unexpected unknown function "u.f-rem"  scss/function-no-unknown
```

### Does the bug relate to non-standard syntax?

SCSS

ybiquitous:
@re757575 Thanks for the report using the template. And for providing the demo!

But I think the reported rule `scss/function-no-unknown` comes from `stylelint-scss`, not `stylelint`, right?
https://github.com/stylelint-scss/stylelint-scss/tree/master/src/rules/function-no-unknown

re757575:
> @re757575 Thanks for the report using the template. And for providing the demo!
> 
> But I think the reported rule `scss/function-no-unknown` comes from `stylelint-scss`, not `stylelint`, right? https://github.com/stylelint-scss/stylelint-scss/tree/master/src/rules/function-no-unknown

Sorry I didn't notice, but I haven't changed stylelint-scss (5.0.1) version.

## Work:
[...]

PR #6921, opened 2023-06-16
 Fix `function-no-unknown` false positives for SCSS functions with namespace

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/6920

No, it's self-explanatory.


ybiquitous (approved):
Thanks for the quick fix! LGTM 👍🏼

Issue #6869, opened 2023-05-28
Fix performance of constructor parsing rules


romainmenke:
Recommended config:

- [x] annotation-no-unknown
- [x] color-no-invalid-hex
- [x] custom-property-no-missing-var-function
- [x] declaration-block-no-duplicate-properties
- [x] font-family-no-duplicate-names
- [x] font-family-no-missing-generic-family-keyword
- [x] function-calc-no-unspaced-operator
- [x] function-linear-gradient-no-nonstandard-direction
- [x] function-no-unknown
- [x] media-feature-name-no-unknown 
- [x] named-grid-areas-no-invalid 
- [x] no-descending-specificity (1687)
- [x] no-duplicate-at-import-rules
- [x] no-duplicate-selectors (363)
- [x] selector-anb-no-unmatchable
- [x] selector-pseudo-class-no-unknown (153)
- [x] selector-pseudo-element-no-unknown
- [x] selector-type-no-unknown
- [x] string-no-newline
- [x] unit-no-unknown

Standard config:

- [x] alpha-value-notation
- [x] color-function-notation
- [x] color-hex-length
- [x] custom-property-pattern
- [x] font-family-name-quotes
- [x] function-name-case
- [x] function-url-quotes
- [x] hue-degree-notation
- [x] import-notation
- [x] length-zero-no-unit
- [x] media-feature-range-notation
- [x] selector-attribute-quotes
- [x] selector-class-pattern (342)
- [x] selector-id-pattern
- [x] selector-no-vendor-prefix
- [x] selector-not-notation
- [x] selector-pseudo-element-colon-notation
- [x] selector-type-case
- [x] shorthand-property-no-redundant-values
- [x] value-keyword-case
- [x] value-no-vendor-prefix

Selector rules:

- [x] selector-attribute-name-disallowed-list
- [x] selector-attribute-operator-allowed-list
- [x] selector-attribute-operator-disallowed-list
- [x] selector-combinator-allowed-list
- [x] selector-combinator-disallowed-list
- [x] selector-max-attribute
- [x] selector-max-class
- [x] selector-max-combinators
- [x] selector-max-compound-selectors
- [x] selector-max-id
- [x] selector-max-pseudo-class
- [x] selector-max-specificity
- [x] selector-max-type
- [x] selector-max-universal
- [x] selector-no-qualifying-type
- [x] selector-pseudo-class-allowed-list
- [x] selector-pseudo-class-disallowed-list
[...]

PR #6922, opened 2023-06-16
Fix `custom-property-no-missing-var-function` performance

romainmenke:

No, it's self-explanatory.

----

before :

```
Warnings: 0
Mean: 106.64779905882351 ms
Deviation: 14.50380702371365 ms
```

after :

```
Warnings: 0
Mean: 98.50316917647058 ms
Deviation: 11.405372458359288 ms
```

### DIFF
--- a/lib/rules/custom-property-no-missing-var-function/index.js
+++ b/lib/rules/custom-property-no-missing-var-function/index.js
@@ -17,9 +17,6 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/custom-property-no-missing-var-function',
 };
 
-const DASHED_IDENT_START = /^--/;
-const DOUBLE_DASH = /--/;
-
 /** @type {import('stylelint').Rule} */
 const rule = (primary) => {
 	return (root, result) => {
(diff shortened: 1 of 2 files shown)


lib/rules/custom-property-no-missing-var-function/index.js
```diff
@@ -32,12 +34,15 @@ const rule = (primary) => {
...
-		root.walkDecls(({ prop }) => {
-			if (isCustomProperty(prop)) knownCustomProperties.add(prop);
+		root.walkDecls(DASHED_IDENT_START, ({ prop }) => {
+			knownCustomProperties.add(prop);
 		});
 
 		root.walkDecls((decl) => {
 			const { value } = decl;
+
+			if (!DOUBLE_DASH.test(value)) return;
```
ybiquitous, line 44: [ask] How about using [`String.prototype.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) instead of a `RegExp`? No significant difference?

```js
const DOUBLE_DASH = "--";

// ...

if (!value.includes(DOUBLE_DASH)) return;
```

ybiquitous (commented):
[suggest] Can you add a changelog entry like this?

https://github.com/stylelint/stylelint/blob/1c44f18c7322449990f790e92ae4a50188a3dac8/CHANGELOG.md?plain=1#L13

lib/rules/custom-property-no-missing-var-function/index.js
romainmenke: I went a slightly different direction.

`if (!value.includes('--')) return;` should be fine because JavaScript has string interning, and strings don't need to be assigned to a `const` for performance.

-------

`root.walkDecls(/^--/, ({ prop }) => {` is more like the existing `root.walkAtRules(/^property$/i, (atRule) => {` above it.

These `RegExp`'s are only created once per stylelint invocation, so also no need to hoist these higher up.

----

thoughts?

ybiquitous: Make sense! 👍🏼

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #6923, opened 2023-06-16
Fix `function-calc-no-unspaced-operator` performance

romainmenke:

No, it's self-explanatory.

----

before :

```
Warnings: 0
Mean: 90.56110877777778 ms
Deviation: 11.54085198893875 ms
```

after : 

```
Warnings: 0
Mean: 83.60294444444445 ms
Deviation: 10.305877029720664 ms
```

### DIFF
--- a/lib/rules/function-calc-no-unspaced-operator/index.js
+++ b/lib/rules/function-calc-no-unspaced-operator/index.js
@@ -48,6 +48,8 @@ const rule = (primary, _secondaryOptions, context) => {
 		}
 
 		root.walkDecls((decl) => {
+			if (!OPERATOR_REGEX.test(getDeclarationValue(decl))) return;
+
 			let needsFix = false;
 			const valueIndex = declarationValueIndex(decl);
 			const parsedValue = valueParser(getDeclarationValue(decl));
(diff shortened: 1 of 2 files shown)


lib/rules/function-calc-no-unspaced-operator/index.js
```diff
@@ -48,6 +48,8 @@ const rule = (primary, _secondaryOptions, context) => {
 		}
 
 		root.walkDecls((decl) => {
+			if (!OPERATOR_REGEX.test(getDeclarationValue(decl))) return;
+
 			let needsFix = false;
 			const valueIndex = declarationValueIndex(decl);
 			const parsedValue = valueParser(getDeclarationValue(decl));
```
ybiquitous, line 55: [suggest] How about reusing a return value from `getDeclarationValue(decl)` like this?

```suggestion
			const value = getDeclarationValue(decl);

			if (!OPERATOR_REGEX.test(value)) return;

			let needsFix = false;
			const valueIndex = declarationValueIndex(decl);
			const parsedValue = valueParser(value);
```

romainmenke: good catch!

romainmenke:
GitHub seems to be having some issues.
It didn't apply the review suggestions correctly and pushed to this branch don't show up.

I am giving it some time before make further changes.

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #6924, opened 2023-06-16
Fix `function-linear-gradient-no-nonstandard-direction` performance

romainmenke:

No, it's self-explanatory.

-----

before :

```
Warnings: 0
Mean: 89.20350494117648 ms
Deviation: 11.294021704297947 ms
```

after : 

```
Warnings: 0
Mean: 77.0127245 ms
Deviation: 7.537474958126632 ms
```

### DIFF
--- a/lib/rules/function-linear-gradient-no-nonstandard-direction/index.js
+++ b/lib/rules/function-linear-gradient-no-nonstandard-direction/index.js
@@ -19,14 +19,22 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/function-linear-gradient-no-nonstandard-direction',
 };
 
+const LINEAR_GRADIENT_FUNCTION = /(?:-webkit-|-moz-|-o-|-ms-)?linear-gradient\(/i;
+const LINEAR_GRADIENT_FUNCTION_NAME = /^(?:-webkit-|-moz-|-o-|-ms-)?linear-gradient$/i;
+
+const DIRECTION_WITH_TO = /^to (top|left|bottom|right)(?: (top|left|bottom|right))?$/i;
+const DIRECTION_WITHOUT_TO = /^(top|left|bottom|right)(?: (top|left|bottom|right))?$/i;
+
+const LEFT_RIGHT_TOP_BOTTOM = /left|right|top|bottom/i;
+const DIGIT = /[\d.]/;
+const ANGLE = /^[\d.]+(?:deg|grad|rad|turn)$/;
+
 /**
  * @param {string} source
  * @param {boolean} withToPrefix
  */
 function isStandardDirection(source, withToPrefix) {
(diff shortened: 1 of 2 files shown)


lib/rules/function-linear-gradient-no-nonstandard-direction/index.js
```diff
@@ -19,14 +19,22 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/function-linear-gradient-no-nonstandard-direction',
 };
 
+const LINEAR_GRADIENT_FUNCTION = /(?:-webkit-|-moz-|-o-|-ms-)?linear-gradient\(/i;
+const LINEAR_GRADIENT_FUNCTION_NAME = /^(?:-webkit-|-moz-|-o-|-ms-)?linear-gradient$/i;
+
+const DIRECTION_WITH_TO = /^to (top|left|bottom|right)(?: (top|left|bottom|right))?$/i;
+const DIRECTION_WITHOUT_TO = /^(top|left|bottom|right)(?: (top|left|bottom|right))?$/i;
+
+const LEFT_RIGHT_TOP_BOTTOM = /left|right|top|bottom/i;
```
Mouvedia, line 28: reminder

Not worth it in this case, but you could reuse `LEFT_RIGHT_TOP_BOTTOM.source` in `DIRECTION_WITH_TO` and  `DIRECTION_WITHOUT_TO` using `RegExp` constructor. If you ever have a very long list, it would be cleaner/shorter.

ybiquitous: Agree. Considering maintainability (and if it never worsens performance), the following refactoring might be worth:

```js
const LINEAR_GRADIENT_FUNCTION = '(?:-webkit-|-moz-|-o-|-ms-)?linear-gradient';
const LINEAR_GRADIENT_FUNCTION_CALL = new RegExp(`${LINEAR_GRADIENT_FUNCTION}\\(`, 'i');
const LINEAR_GRADIENT_FUNCTION_NAME = new RegExp(`^${LINEAR_GRADIENT_FUNCTION}$`, 'i');

const DIRECTION = /top|left|bottom|right/;
const DIRECTION_WITH_TO = new RegExp(`^to (${DIRECTION.source})(?: (${DIRECTION.source}))?$`, 'i');
const DIRECTION_WITHOUT_TO = new RegExp(`^(${DIRECTION.source})(?: (${DIRECTION.source}))?$`, 'i');
```

Any thoughts?

romainmenke: Thank you both for this suggestion 🙇

ybiquitous (approved):
Thanks for your commitment to performance improvement! 👏🏼

PR #6925, opened 2023-06-16
Fix `selector-anb-no-unmatchable` performance

romainmenke:

No, it's self-explanatory.

-----

before : 

```
Warnings: 0
Mean: 171.01108333333332 ms
Deviation: 11.09835497913841 ms
```

after :

```
Warnings: 0
Mean: 115.6843908125 ms
Deviation: 7.734276833118593 ms
```

### DIFF
--- a/lib/utils/hasANPlusBNotationPseudoClasses.js
+++ b/lib/utils/hasANPlusBNotationPseudoClasses.js
@@ -5,13 +5,11 @@ const {
 	aNPlusBOfSNotationPseudoClasses,
 } = require('../reference/selectors');
 
-const HAS_A_N_PLUS_B_NOTATION_PSEUDO_CLASSES = new RegExp(
-	`\\b:(?:${[
-		...aNPlusBNotationPseudoClasses.values(),
-		...aNPlusBOfSNotationPseudoClasses.values(),
-	].join('|')})\\(`,
-	'i',
-);
+const classes = [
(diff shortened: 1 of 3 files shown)


lib/utils/hasANPlusBNotationPseudoClasses.js
```diff
@@ -0,0 +1,24 @@
...
+	aNPlusBOfSNotationPseudoClasses,
+} = require('../reference/selectors');
+
+const HAS_A_N_PLUS_B_NOTATION_PSEUDO_CLASSES = new RegExp(
+	`\\b:(?:${[
+		...aNPlusBNotationPseudoClasses.values(),
+		...aNPlusBOfSNotationPseudoClasses.values(),
+	].join('|')})\\(`,
+	'i',
+);
```
ybiquitous, line 14: [suggest] How about using a local variable to improve readability for the regex like this?

```suggestion
const classes = [
	...aNPlusBNotationPseudoClasses.values(),
	...aNPlusBOfSNotationPseudoClasses.values(),
].join('|');
const HAS_A_N_PLUS_B_NOTATION_PSEUDO_CLASSES = new RegExp(`\\b:(?:${classes})\\(`, 'i');
```

romainmenke: yes, much better, thank you 🙇

ybiquitous (approved):
Thank you, LGTM 👍🏼

PR #6931, opened 2023-06-18
Refactor value parsing to use a simpler algorithm

romainmenke:

None, it's a minor code quality fix that I noticed while working on a different issue.

This isn't a user facing change, so I am omitting a changelog entry.

----

`parseCommaSeparatedListOfComponentValues` :
- list of lists of component values
- split by comma
- doesn't include the comma as a component value
- useful for : 
   - media query lists
   - selectors
   - function params (`rgb(0, 0, 0)`)

`parseListOfComponentValues` :
- list of component values
- no splits
- does include the comma as a `TokenNode`
- useful for anything

This distinction is important when writing specialized parsers (like the media query parser) but it isn't relevant when validating component values.

Using `parseListOfComponentValues` results in simpler, cleaner code.

--------

The diffs appear large because the indent changed.

Actual diff :

```diff
- parseCommaSeparatedListOfComponentValues(tokenize({ css: value }))
- 	.flatMap((x) => x)
+ parseListOfComponentValues(tokenize({ css: value }))
	.forEach((componentValue) => {
```

### DIFF
--- a/lib/rules/function-no-unknown/index.js
+++ b/lib/rules/function-no-unknown/index.js
@@ -4,9 +4,9 @@ const fs = require('fs');
 const functionsListPath = require('css-functions-list');
 const { tokenize } = require('@csstools/css-tokenizer');
 const {
-	parseCommaSeparatedListOfComponentValues,
 	isFunctionNode,
 	isSimpleBlockNode,
+	parseListOfComponentValues,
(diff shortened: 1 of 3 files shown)


ybiquitous (approved):
LGTM. Thanks!

Issue #6856, opened 2023-05-20
Add `insideFunctions: {"function": number}` to `number-max-precision`


firefoxic:
### What is the problem you're trying to solve?

It is very rare to find a need for precision values greater than 2 decimal places. Therefore, the `number-max-precision' rule almost always specifies `2'.

But the `chroma` component of the `oklch()` color function is special in this sense. Even in the `rec2020` color space, the highest value of `chroma` does not exceed `0.5`, and values from `0` to `0.3` are mostly used. With such a short range, even 3 decimal places are not enough, it is desirable to use at least 4.

At the same time, I would not want to completely ignore this component, because automatic converters and generators often produce more than 10 digits, which is not necessary at all.

### What solution would you like to see?

I would like `number-max-precision` to have an additional option, something like `oklchChroma` (suggest a more appropriate name), which takes an integer like the main value of the rule, but does not depend on it.

Example:
```json
{
	"rules": {
		"number-max-precision": [
			2,
			{
				"oklchChroma": 4
			}
		]
	}
}
```

ybiquitous:
@firefoxic Thanks for the proposal. It's an interesting case. So, could you provide CSS code examples so that we can easily understand your problem?

For example, when the following config is given,

```json
{
	"rules": {
		"number-max-precision": [
			2,
			{
				"oklchChroma": 4
			}
		]
	}
}
```

CSS code:

```css
a {
	/* bad */
	background-color: oklch(...);

	/* good */
	background-color: oklch(...);
}
```

firefoxic:
When the following config is given:
```json
{
	"rules": {
		"number-max-precision": [
			2,
			{
				"oklchChroma": 4
			}
		]
	}
}
```

```css
a {
	/* bad, every number */
	color: oklch(75.323% 0.24546 45.234);
	transition: color 0.315s;

	/* good */
	color: oklch(75.32% 0.2455 45.23);
	transition: color 0.32s;
}
```

firefoxic:
But when the following config is given:
```json
{
	"rules": {
		"number-max-precision": 2
	}
}
```
CSS code:
```css
a {
	/* bad, every number */
	color: oklch(75.32% 0.2455 45.23);
	transition: color 0.315s;

	/* good */
	color: oklch(75.32% 0.25 45.23);
	transition: color 0.32s;
}
```
And the approximate correspondence of these colors is:
`oklch(75.32% 0.2455 45.23)` → `#ff6800`
[...]

PR #6932, opened 2023-06-18
Add `insideFunctions: {"function": int}` to `number-max-precision`

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/6856

I've switched this rule to `@csstools/css-parser-algorithms` because the walker state feature is ideal to add support for this new plugin option.

### DIFF
--- a/lib/rules/number-max-precision/README.md
+++ b/lib/rules/number-max-precision/README.md
@@ -155,22 +155,22 @@ a { color: rgb(127.333 0 0); }
 
 <!-- prettier-ignore -->
 ```css
-a { color: rgb(calc(127.333 / 3) 0 0) }
+a { color: rgb(calc(127.333 / 3) 0 0); }
 ```
 
 The following patterns are _not_ considered problems:
 
 <!-- prettier-ignore -->
 ```css
-a { color: oklch(0.333 0 0) }
+a { color: oklch(0.333 0 0); }
 ```
 
 <!-- prettier-ignore -->
 ```css
-a { color: lab(0.3333 0 0) }
+a { color: lab(0.3333 0 0); }
 ```
 
 <!-- prettier-ignore -->
 ```css
-a { color: oklab(calc(127.333 / 3) 0 0) }
+a { color: oklab(calc(127.333 / 3) 0 0); }
 ```
--- a/lib/rules/number-max-precision/README.md
+++ b/lib/rules/number-max-precision/README.md
@@ -134,7 +134,7 @@ a {
 }
 ```
 
-### `insideFunctions: {"/regex/": 1, /regex/: 1, "string": 1}`
+### `insideFunctions: {"/regex/": int, /regex/: int, "string": int}`
 
 The `insideFunctions` option can change a primary option value for specified functions.
 
@@ -150,7 +150,7 @@ The following patterns are considered problems:
 
 <!-- prettier-ignore -->
 ```css
-a { color: rgb(127.333 0 0) }
+a { color: rgb(127.333 0 0); }
 ```
 
 <!-- prettier-ignore -->
--- a/lib/rules/number-max-precision/README.md
+++ b/lib/rules/number-max-precision/README.md
@@ -133,3 +133,44 @@ a {
   width: 10.989my-other-unit;
 }
 ```
+
+### `insideFunctions: {"/regex/": 1, /regex/: 1, "string": 1}`
+
+The `insideFunctions` option can change a primary option value for specified functions.
+
+For example, with `2`.
+
+Given:
+
+```json
+{"/^(oklch|oklab|lch|lab)$/", 4}
+```
+
+The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+a { color: rgb(127.333 0 0) }
+```
+
+<!-- prettier-ignore -->
+```css
+a { color: rgb(calc(127.333 / 3) 0 0) }
+```
+
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+a { color: oklch(0.333 0 0) }
+```
+
+<!-- prettier-ignore -->
+```css
+a { color: lab(0.3333 0 0) }
+```
+
+<!-- prettier-ignore -->
+```css
+a { color: oklab(calc(127.333 / 3) 0 0) }
+```
--- a/lib/rules/number-max-precision/__tests__/index.js
+++ b/lib/rules/number-max-precision/__tests__/index.js
@@ -41,7 +41,7 @@ testRule({
 			code: "@IMPORT '1.123.css'",
 		},
 		{
-			code: 'a { background: url(1.123.jpg) url("foo" 1.123.jpg) }',
+			code: 'a { background: url(1.123.jpg) url("foo" 1.123.jpg); }',
 		},
 		{
 			code: 'a { my-string: "1.2345"; }',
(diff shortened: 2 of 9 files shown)


lib/rules/number-max-precision/index.js
```diff
@@ -72,47 +85,122 @@ const rule = (primary, secondaryOptions) => {
 				return;
 			}
 
-			valueParser(value).walk((valueNode) => {
-				const { unit } = getDimension(valueNode);
+			parseListOfComponentValues(tokenize({ css: value })).forEach((componentValue) => {
+				/** @type {{ ignored: boolean, precision: number }} */
```
ybiquitous, line 89: [suggest] When specifying a primary option type like this:

```diff
-/** @type {import('stylelint').Rule} */
+/** @type {import('stylelint').Rule<number>} */
```

This `@type` annotation seems no longer needed:

```suggestion
```

lib/rules/number-max-precision/index.js
```diff
@@ -72,47 +85,122 @@ const rule = (primary, secondaryOptions) => {
...
+		 * @param {string} functionName
+		 * @param {number} currentPrecision
+		 * @returns {number}
+		 */
+		function precisionInsideFunction(functionName, currentPrecision) {
+			if (functionName in secondaryOptions.insideFunctions) {
+				return secondaryOptions.insideFunctions[functionName];
+			}
+
+			for (const [name, precision] of Object.entries(secondaryOptions.insideFunctions)) {
```
ybiquitous, line 196: [suggest] Constructing `insideFunctions` before walking might be more readable and have a bit performance benefit. Any thoughts?

```js
if (!validOptions) {
	return;
}

/** @type {Map<string, number>} */
const insideFunctions = new Map(Object.entries(secondaryOptions?.insideFunctions ?? {}));
```

```suggestion
			const precision = insideFunctions.get(functionName);
			if (isNumber(precision)) return precision;

			for (const [name, precision] of insideFunctions) {
```

ybiquitous (changes requested):
@romainmenke Thanks for the pull request! Could you address the reviews?

lib/rules/number-max-precision/README.md
```diff
@@ -133,3 +133,44 @@ a {
   width: 10.989my-other-unit;
 }
 ```
+
+### `insideFunctions: {"/regex/": 1, /regex/: 1, "string": 1}`
```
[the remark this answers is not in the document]
romainmenke: Good catch 🙇

lib/rules/number-max-precision/index.js
```diff
@@ -72,47 +89,123 @@ const rule = (primary, secondaryOptions) => {
...
 			});
 		}
+
+		/**
+		 * @param {string} functionName
+		 * @param {number} currentPrecision
+		 * @returns {number}
+		 */
+		function precisionInsideFunction(functionName, currentPrecision) {
+			{
```
romainmenke, line 195: I've added a block here because otherwise `precision` is declared twice.

I don't have a strong preference here.
Changing the names would also work.

lib/rules/number-max-precision/index.js
romainmenke: Yes, much better, thank you

lib/rules/number-max-precision/index.js
ybiquitous: Hum, how about the name `precisionForFunction` instead of `precision`?

lib/rules/number-max-precision/index.js
```diff
@@ -46,21 +55,29 @@ const rule = (primary, secondaryOptions) => {
 			return;
 		}
 
+		/** @type {Map<string, number>} */ const insideFunctions = new Map(
+			Object.entries(secondaryOptions?.insideFunctions ?? {}),
+		);
```
ybiquitous, line 60: [suggest] The following seems more consistent and readable a bit to me. Thoughts?

```suggestion
		/** @type {Map<string, number>} */
		const insideFunctions = new Map(Object.entries(secondaryOptions?.insideFunctions ?? {}));
```

ybiquitous (approved):
Thanks, LGTM! 👍🏼

lib/rules/number-max-precision/__tests__/index.js
```diff
@@ -41,7 +41,7 @@ testRule({
 			code: "@IMPORT '1.123.css'",
 		},
 		{
-			code: 'a { background: url(1.123.jpg) }',
+			code: 'a { background: url(1.123.jpg) url("foo" 1.123.jpg) }',
 		},
 		{
 			code: 'a { my-string: "1.2345"; }',
```
Mouvedia, line 47: maybe replace this one by
```
font-family: "1.2345";
```
ref: https://drafts.csswg.org/css-fonts/#family-name-syntax

lib/rules/number-max-precision/__tests__/index.js
```diff
@@ -41,7 +41,7 @@ testRule({
 			code: "@IMPORT '1.123.css'",
 		},
 		{
-			code: 'a { background: url(1.123.jpg) }',
+			code: 'a { background: url(1.123.jpg) url("foo" 1.123.jpg) }',
 		},
```
Mouvedia, line 45: suggestion

maybe add `image-set`?
https://developer.mozilla.org/en-US/docs/Web/CSS/image/image-set

lib/rules/number-max-precision/__tests__/index.js
romainmenke: The description for this test is :
> ignore all strings rather than only in 'content'

The purpose seems to have been to test a non-specific property with a string value.
In contrast with a known property where string values are expected.

lib/rules/number-max-precision/__tests__/index.js
romainmenke: The tests for `url` are a bit iffy in general.
I think these were intended to catch an implementation detail in `postcss-value-parser`. Not because anyone really wants to ignore floating point precision in numbers in `url`.

In `postcss-value-parser` `url(...)` is always parsed as a function and it's contents can appear to be numbers. e.g. `url(123.jpg)`

This however is incorrect and the CSS specification is very different.

Either `url(...)` is a single `url` token and it's value is a url or it is a function with a name of `url` and it contains a value list that starts with a string token.

- `url(123.jpg)` -> url token
- `url("123.jpg")` -> function with a name of url

Because `@csstools/css-tokenizer` and `@csstools/css-parser-algorithm` follow the CSS specification they will never produce a `url` function where the url value is parsed as a number.

-------

This is also why I had to add a [url modified](https://www.w3.org/TR/css-values-4/#url-modifiers) `url("foo" 1.123.jpg);`. Without this the test coverage wasn't 100%.

Currently no url modifiers exist in the CSS specification, so this isn't testing realistic CSS anymore.

-------

TL;DR;

Either we should remove the special handling of `url` or we should leave it as is.
I don't think we should add more tests for something that is actually an implementation detail in a parser that (as of this change) isn't used anymore :)

Mouvedia: suggestion

Add a concise one sentence _why_ comment above the `if` that ignores `url`?

romainmenke: I've added a note, please let me know if this is unclear in any way.

romainmenke:
Thank you everyone for the thoughtful reviews 🙇

PR #6937, opened 2023-06-19
Introduce rule tables in rules page

ybiquitous:
> Which issue, if any, is this issue related to?

Closes #6928

See https://github.com/stylelint/stylelint.io/pull/328 to preview the website.

> Is there anything in the PR that needs further explanation?

This change introduces rule tables to the rules page to improve visibility.

PR #6939, opened 2023-06-20
Fix rule name's typo in rules doc

ybiquitous:

Follow-up to #6937

Also, this change adds a test case to detect such typos.


docs/user-guide/rules.md
```diff
@@ -425,18 +425,18 @@ These rules are deprecated — we won't fix bugs nor add options, and we will re
...
+| [`selector-descendant-combinator-no-non-space`](../../lib/rules/selector-descendant-combinator-no-non-space/README.md)<br/>Disallow non-space characters for descendant combinators of selectors.                                              |     |     | ✅  |
+| [`selector-max-empty-lines`](../../lib/rules/selector-max-empty-lines/README.md)<br/>Limit the number of adjacent empty lines within selectors.                                                                                                |     |     | ✅  |
+| [`selector-pseudo-class-case`](../../lib/rules/selector-pseudo-class-case/README.md)<br/>Specify lowercase or uppercase for pseudo-class selectors.                                                                                            |     |     | ✅  |
+| [`selector-pseudo-class-parentheses-space-inside`](../../lib/rules/selector-pseudo-class-parentheses-space-inside/README.md)<br/>Require a single space or disallow whitespace on the inside of the parentheses within pseudo-class selectors. |     |     | ✅  |
```
romainmenke, line 438: Good catch!

I had actually expected this to be caught by the existing tests.
Nice that there is extra coverage now.

romainmenke (approved):
LGTM!

docs/user-guide/rules.md
ybiquitous: Ya, this new case is what we had to add first. 😅

PR #6940, opened 2023-06-20
Convert system tests code to ESM

ybiquitous:

Ref: #5291

This pull request rewrites the system tests code to ease ESM migration. There should be no effects on published code.

Also, it adds the `--experimental-vm-modules` flag of Node.js to support ESM with Jest.
See https://jestjs.io/docs/ecmascript-modules

I'll open similar PRs later to proceed with ESM migration gradually.


romainmenke (approved):
LGTM!

ybiquitous:
Always thanks for the quick review! 😊

PR #6941, opened 2023-06-20
Rewrite some tests (`lib/__tests__/*`) to ESM

ybiquitous:

Ref:
- #5291
- #6940

Some tests keep `__dirname` to minimize diffs.

### DIFF
--- a/lib/__tests__/disableRanges-integration.test.mjs
+++ b/lib/__tests__/disableRanges-integration.test.mjs
@@ -1,5 +1,5 @@
 import blockNoEmpty from '../rules/block-no-empty/index.js';
-import maxLineLength from '../rules/max-line-length/index';
+import maxLineLength from '../rules/max-line-length/index.js';
 import selectorCombinatorSpaceBefore from '../rules/selector-combinator-space-before/index.js';
 import stringQuotes from '../rules/string-quotes/index.js';
 
--- a/lib/__tests__/disableRanges-integration.test.mjs
+++ b/lib/__tests__/disableRanges-integration.test.mjs
@@ -1,9 +1,7 @@
-'use strict';
-
-const blockNoEmpty = require('../rules/block-no-empty');
-const maxLineLength = require('../rules/max-line-length');
-const selectorCombinatorSpaceBefore = require('../rules/selector-combinator-space-before');
-const stringQuotes = require('../rules/string-quotes');
+import blockNoEmpty from '../rules/block-no-empty/index.js';
+import maxLineLength from '../rules/max-line-length/index';
+import selectorCombinatorSpaceBefore from '../rules/selector-combinator-space-before/index.js';
+import stringQuotes from '../rules/string-quotes/index.js';
 
 // disabling all rules
 testRule({
--- a/lib/__tests__/extends.test.mjs
+++ b/lib/__tests__/extends.test.mjs
@@ -1,13 +1,20 @@
 import { fileURLToPath } from 'node:url';
 
+import readJSONFile from '../testUtils/readJSONFile.mjs';
 import safeChdir from '../testUtils/safeChdir.js';
 import standalone from '../standalone.js';
 
-import configExtendingAnotherExtend from './fixtures/config-extending-another-extend.json';
-import configExtendingOne from './fixtures/config-extending-one.json';
-import configExtendingThreeWithOverride from './fixtures/config-extending-three-with-override.json';
+const configExtendingAnotherExtend = readJSONFile(
+	new URL('./fixtures/config-extending-another-extend.json', import.meta.url),
+);
+const configExtendingOne = readJSONFile(
+	new URL('./fixtures/config-extending-one.json', import.meta.url),
+);
+const configExtendingThreeWithOverride = readJSONFile(
+	new URL('./fixtures/config-extending-three-with-override.json', import.meta.url),
+);
 
-const fixturesPath = fileURLToPath(new URL('fixtures', import.meta.url));
+const fixturesPath = fileURLToPath(new URL('./fixtures', import.meta.url));
 
 it('basic extending', async () => {
 	const linted = await standalone({
--- a/lib/__tests__/extends.test.mjs
+++ b/lib/__tests__/extends.test.mjs
@@ -1,20 +1,19 @@
-'use strict';
+import { fileURLToPath } from 'node:url';
 
-const path = require('path');
+import safeChdir from '../testUtils/safeChdir.js';
+import standalone from '../standalone.js';
 
-const safeChdir = require('../testUtils/safeChdir');
-const configExtendingAnotherExtend = require('./fixtures/config-extending-another-extend.json');
-const configExtendingOne = require('./fixtures/config-extending-one');
-const configExtendingThreeWithOverride = require('./fixtures/config-extending-three-with-override');
-const standalone = require('../standalone');
+import configExtendingAnotherExtend from './fixtures/config-extending-another-extend.json';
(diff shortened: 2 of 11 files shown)


lib/__tests__/disableRanges-integration.test.mjs
```diff
@@ -1,9 +1,7 @@
-'use strict';
-
-const blockNoEmpty = require('../rules/block-no-empty');
-const maxLineLength = require('../rules/max-line-length');
-const selectorCombinatorSpaceBefore = require('../rules/selector-combinator-space-before');
-const stringQuotes = require('../rules/string-quotes');
+import blockNoEmpty from '../rules/block-no-empty/index.js';
+import maxLineLength from '../rules/max-line-length/index';
```
hudochenkov, line 2: I believe extension is required. There probably an ESLint rule for this somewhere. Maybe it's a good idea to enabled this kind of check for `*.mjs`?

```suggestion
import maxLineLength from '../rules/max-line-length/index.js';
```

lib/__tests__/extends.test.mjs
```diff
@@ -1,20 +1,19 @@
...
-const path = require('path');
+import safeChdir from '../testUtils/safeChdir.js';
+import standalone from '../standalone.js';
 
-const safeChdir = require('../testUtils/safeChdir');
-const configExtendingAnotherExtend = require('./fixtures/config-extending-another-extend.json');
-const configExtendingOne = require('./fixtures/config-extending-one');
-const configExtendingThreeWithOverride = require('./fixtures/config-extending-three-with-override');
-const standalone = require('../standalone');
+import configExtendingAnotherExtend from './fixtures/config-extending-another-extend.json';
```
hudochenkov, line 6: ESM import of `*.json` [looks different](https://nodejs.org/api/esm.html#json-modules).

Shall we use [standard ways](https://www.stefanjudis.com/snippets/how-to-import-json-files-in-es-modules-node-js/) of importing JSON? Or maybe convert JSON to JS?

lib/__tests__/extends.test.mjs
```diff
@@ -68,7 +67,7 @@ it('extending a config that is overridden', async () => {
 });
 
 describe('extending a config from process.cwd', () => {
-	safeChdir(__dirname);
+	safeChdir(new URL('.', import.meta.url));
```
hudochenkov, line 70: I noticed that in other lines `fileURLToPath` is used. Why it's omitted here?

lib/__tests__/disableRanges-integration.test.mjs
ybiquitous: Good catch! I'll try finding such an ESLint rule. Thanks.

lib/__tests__/extends.test.mjs
ybiquitous: Good point, but this is intentional to minimize diffs. 😓 

I'd like to use Import Assertions (the most straightforward), but unfortunately experimental.

Jest supports Import Assertions like this:

```js
// The test passes ✅ 
import configExtendingAnotherExtend from './fixtures/config-extending-another-extend.json' assert { type: 'json' };
```

However, ESLint complains:

```
/Users/masafumi.koba/git/stylelint/stylelint/lib/__tests__/extends.test.mjs
  6:92  error  Parsing error: Unexpected token assert
```

So, I'd like to leave as-is and rewrite it later. Any thoughts?

lib/__tests__/disableRanges-integration.test.mjs
hudochenkov: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md

lib/__tests__/extends.test.mjs
ybiquitous: Now `safeChdir()` also allows an `URL` argument like many Node.js APIs do, e.g. [`fs.readFile()`](https://nodejs.org/api/fs.html#fsreadfilepath-options-callback).

lib/__tests__/extends.test.mjs
ybiquitous: I reconsidered and introduced a new test utility to keep the code small (caa84e07).

lib/__tests__/disableRanges-integration.test.mjs
ybiquitous: Thanks. I've opened PR #6942.

lib/__tests__/disableRanges-integration.test.mjs
```diff
@@ -1,9 +1,7 @@
-'use strict';
-
-const blockNoEmpty = require('../rules/block-no-empty');
-const maxLineLength = require('../rules/max-line-length');
-const selectorCombinatorSpaceBefore = require('../rules/selector-combinator-space-before');
-const stringQuotes = require('../rules/string-quotes');
+import blockNoEmpty from '../rules/block-no-empty/index.js';
+import maxLineLength from '../rules/max-line-length/index.js';
+import selectorCombinatorSpaceBefore from '../rules/selector-combinator-space-before/index.js';
+import stringQuotes from '../rules/string-quotes/index.js';
```
Mouvedia, line 4: You could maybe import them all from `stylelint/lib/rules
/index.js`. But it would require to rename them all due to the hyphens.
Your choice.

lib/testUtils/readJSONFile.mjs
```diff
@@ -0,0 +1,9 @@
+import * as fs from 'node:fs';
+
+/**
+ * @param {string | URL} filePath
+ * @returns {unknown}
+ */
+export default function readJSONFile(filePath) {
+	return JSON.parse(fs.readFileSync(filePath, 'utf8'));
```
Mouvedia, line 8: question

Can't you do 
```
import configExtendingAnotherExtend from './fixtures/config-extending-another-extend.json' assert { type : 'json' };
```
instead?

lib/__tests__/extends.test.mjs
```diff
@@ -89,7 +95,7 @@ describe('extending a config from options.cwd', () => {
 			config: {
 				extends: ['./fixtures/config-string-quotes-single'],
 			},
-			cwd: __dirname,
+			cwd: fileURLToPath(new URL('.', import.meta.url)),
```
Mouvedia, line 98: minor

You could put it in a variable since you have it twice.
i.e. here and line 77
e.g. `const foo = new URL('.', import.meta.url);`

lib/__tests__/disableRanges-integration.test.mjs
Mouvedia: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/group-exports.md would be useful too

lib/testUtils/readJSONFile.mjs
hudochenkov: This is an experimental syntax, that [will change](https://twitter.com/robpalmer2/status/1639014193369354240) to another syntax.

hudochenkov (approved):

lib/testUtils/readJSONFile.mjs
ybiquitous: See https://github.com/stylelint/stylelint/pull/6941#discussion_r1236012487

Jest supports Import Assertions, while ESLint doesn't.

lib/__tests__/extends.test.mjs
ybiquitous: I'd like to avoid `__dirname` as possible because it looks like CommonJS. I think twice occurrences are acceptable.

ybiquitous:
Thanks for the review!

PR #6942, opened 2023-06-21
Add `eslint-plugin-import` to detect missing file extensions in `import`

ybiquitous:

Ref: https://github.com/stylelint/stylelint/pull/6941#discussion_r1236002418

For example:

```diff
-import replaceBackslashes from '../lib/testUtils/replaceBackslashes.js';
+import replaceBackslashes from '../lib/testUtils/replaceBackslashes';
```

```console
$ npm run lint:js
...
/Users/masafumi.koba/git/stylelint/stylelint/system-tests/systemTestUtils.mjs
  6:32  error  Missing file extension "js" for "../lib/testUtils/replaceBackslashes"  import/extensions
```

See https://github.com/import-js/eslint-plugin-import/blob/88dd8153571373151c3c7b508c5944cb2bba1588/docs/rules/extensions.md

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,22 @@
       "stylelint",
       "stylelint/jest"
     ],
+    "overrides": [
+      {
+        "files": [
+          "**/*.mjs"
+        ],
+        "plugins": [
+          "eslint-plugin-import"
+        ],
+        "rules": {
+          "import/extensions": [
+            "error",
+            "ignorePackages"
+          ]
+        }
+      }
+    ],
     "globals": {
       "__dirname": true,
       "module": true,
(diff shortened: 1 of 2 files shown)


alexander-akait:
Yeah, it is a good idea and allow to moving to ESM and start to fix such things

package.json
```diff
@@ -62,6 +62,22 @@
...
+      {
+        "files": [
+          "**/*.mjs"
+        ],
+        "plugins": [
+          "eslint-plugin-import"
+        ],
+        "rules": {
+          "import/extensions": [
+            "error",
```
Mouvedia, line 75: It should probably be a warning for now so that it won't add noise to new PRs.

hudochenkov: I think `error` is correct here, because extension *is* required. Without out code is invalid and won't work.

hudochenkov (approved):

PR #6945, opened 2023-06-21
Avoid at-mark use in Markdown files

ybiquitous:

This change ensures that an at-mark unexpectedly is not handled as a user mention, especially generating the website pages (stylelint.io).


mattxwang (approved):
Good catch!

Issue #6928, opened 2023-06-05
Introduce rules tables in rules page


astorije:
### What is the problem you're trying to solve?

We currently use the `standard` preset. Several times, I skimmed through the whole list of rules to see which extra rules need to be manually enabled. Because the Ⓡ & Ⓢ characters blend in so well on the page, it can't be easily done by skimming visually, and one misses a rule easily.

### What solution would you like to see?

I can think of 2 solutions, neither mutually exclusive.

The first idea and reason why I opened this issue, would be a couple toggles at the top of the page to show/hide rules that are already in a preset. That way the list would only contain rules that one needs to switch on manually.

In comparison, the list of ESLint rules is a lot easier to visually parse: https://eslint.org/docs/latest/rules/
So a second option could be to move align those characters on the right of the list, with maybe some color coding, spacing, etc. to make everything stand out a little.

Thanks for this amazing tool!

ybiquitous:
Since there are no objections, I've labeled the issue as **ready to implement**. Please consider [contributing](https://stylelint.io/contributing) if you have time.

astorije:
The only thing I noticed, looking at the result from my phone: instead of using ✅'s, I wonder if it wouldn't be better to reuse the column icons so it's easier to follow when scrolling, not having to remember which column corresponds to what.

ybiquitous:
@astorije Ah, that sounds good! Can you create a follow-up pull request, please?

ybiquitous:
I'm starting to replace `✅`.

ybiquitous:
Opened PR #6946

ybiquitous:
Oh, sorry for taking away the opportunity for contribution. 🙇🏼 

Does PR #6946 satisfy your expectation?

ybiquitous:
I'd appreciate it if you'd react/comment on that PR.

ybiquitous:
Here's the website preview: https://deploy-preview-328--stylelint.netlify.app/user-guide/rules

PR: https://github.com/stylelint/stylelint.io/pull/328

PR #6946, opened 2023-06-21
Replace `✅` with corresponding symbols in rule doc tables

ybiquitous:

Follow-up to #6937

Instead of `✅`, reusing symbols makes people to read the doc easier, especially on small-width devices.

See https://github.com/stylelint/stylelint/issues/6928#issuecomment-1599111520

Example: https://typescript-eslint.io/rules/


mattxwang (approved):
Nice change; with a manual check, it seems like each replacement is correct.

ybiquitous:
@mattxwang Thanks for the manual check!

PR #6956, opened 2023-06-22
Fix `declaration-block-no-redundant-longhand-properties` autofix for `border-width` shorthand

### LINKED ISSUE #6947: Fix `declaration-block-no-redundant-longhand-properties` autofix for `border-width` shorthand

This syntax will raise the warning `Expected shorthand property "border-width" (declaration-block-no-redundant-longhand-properties)Stylelint[declaration-block-no-redundant-longhand-properties](https://stylelint.io/user-guide/rules/declaration-block-no-redundant-longhand-properties)`.

```CSS
div {
    border-top-width: 0px;
    border-right-width: 1px;
    border-bottom-width: 2px;
    border-left-width: 3px;
}
```

It will be fixed to the following code. (The order of the original items is not relevant.)

```CSS
div {
    border-width: 0 2px 3px 1px;
}
```

If two values should be combined, this leads to the same error:

```CSS
div {
    border-top-width: 0px;
    border-right-width: 1px;
    border-bottom-width: 2px;
    border-left-width: 1px;
}
```

becomes:

```CSS
div {
    border-width: 0 2px 1px 1px;
}
```

{
  "extends": [
    "stylelint-config-standard"
  ]
}

Stylelint demo

[stylelint](https://stylelint.io/)@15.8.0
[stylelint-config-standard](https://www.npmjs.com/package/stylelint-config-standard)@33.0.0

```CSS
div {
    border-top-width: 0px;
    border-right-width: 1px;
    border-bottom-width: 2px;
    border-left-width: 3px;
}
```

Should become:

```CSS
div {
    border-width: 0 1px 2px 3px;
}
```

If two values should be combined, this leads to the same error:

```CSS
div {
    border-top-width: 0px;
    border-right-width: 1px;
    border-bottom-width: 2px;
    border-left-width: 1px;
}
```

Should become:

```CSS
div {
    border-width: 0 1px 2px;
}
```

### What actually happened?

```CSS
div {
    border-top-width: 0px;
    border-right-width: 1px;
    border-bottom-width: 2px;
    border-left-width: 3px;
}
```

Wrongly becomes:

```CSS
div {
    border-width: 0 2px 3px 1px;
}
```

If two values should be combined, this leads to the same error:

```CSS
div {
    border-top-width: 0px;
    border-right-width: 1px;
    border-bottom-width: 2px;
    border-left-width: 1px;
}
```

Wrongly becomes:

```CSS
div {
    border-width: 0 2px 1px 1px;
}
```

_No response_

### Proposal to fix the bug

_No response_

mattxwang:

Closes #6947.

Sorry that I missed this! Happy to quickly cut a new release if we'd like.

### DIFF
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.js
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.js
@@ -206,6 +206,12 @@ testRule({
 			description: 'autofixer should not mangle css functions with comma separated values',
 			message: messages.expected('transition'),
 		},
+		{
+			code: 'a { border-top-width: 0px; border-right-width: 1px; border-bottom-width: 2px; border-left-width: 3px; }',
+			fixed: 'a { border-width: 0px 1px 2px 3px; }',
+			description: 'explicit border-width test',
+			message: messages.expected('border-width'),
+		},
 	],
 });
 
--- a/.changeset/sixty-lies-repair.md
+++ b/.changeset/sixty-lies-repair.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `declaration-block-no-redundant-longhand-properties` autofix for `border-width` shorthand
--- a/lib/reference/properties.js
+++ b/lib/reference/properties.js
@@ -180,9 +180,9 @@ const longhandSubPropertiesOfShorthandProperties = new Map([
 		new Set([
 			// prettier-ignore
 			'border-top-width',
+			'border-right-width',
 			'border-bottom-width',
 			'border-left-width',
-			'border-right-width',
 		]),
 	],
 	[


lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.js
```diff
@@ -206,6 +206,12 @@ testRule({
 			description: 'autofixer should not mangle css functions with comma separated values',
 			message: messages.expected('transition'),
 		},
+		{
+			code: 'a { border-top-width: 0px; border-right-width: 1px; border-bottom-width: 2px; border-left-width: 3px; }',
+			fixed: 'a { border-width: 0px 1px 2px 3px; }',
+			description: 'explicit border-width test',
+			message: messages.expected('border-width'),
+		},
 	],
```
Mouvedia, line 215: Can you add an example for `margin` or `border-width` or `padding` for `top | left/right | bottom`.
If if it's not supported yet add `skip: true,`.

Mouvedia:
Can you add a test for `border-radius` ?
It should be `top-left | top-right | bottom-right | bottom-left`.
see https://github.com/stylelint/stylelint/blob/b900c7b5b310fedfa96ca1ee79f80a84f28ea1ff/lib/reference/properties.js#L144

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.js
mattxwang: Sorry, a bit confused -- what do you mean by `top | left/right | bottom`? Do you mean when all four are specified, or when some of the values are inferred/collapsed? The latter isn't covered by this rule.

ybiquitous (approved):
Thank you. LGTM! 👍🏼

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.js
Mouvedia: > The latter isn't covered by this rule.

That's why I was proposing to add `skip: true` for now.
It could be an enhancement; it's expected by users.
cf #6947

PR #6958, opened 2023-06-23
Fix `declaration-block-no-redundant-longhand-properties` autofix for `border-radius` shorthand

mattxwang:

Ref: https://github.com/stylelint/stylelint/pull/6956#issuecomment-1603347258

### DIFF
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.js
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.js
@@ -206,6 +206,12 @@ testRule({
 			description: 'autofixer should not mangle css functions with comma separated values',
 			message: messages.expected('transition'),
 		},
+		{
+			code: 'a { border-top-left-radius: 1em; border-top-right-radius: 2em; border-bottom-right-radius: 3em; border-bottom-left-radius: 4em; }',
+			fixed: 'a { border-radius: 1em 2em 3em 4em; }',
+			description: 'explicit border-radius test',
+			message: messages.expected('border-radius'),
+		},
 	],
 });
 
--- a/.changeset/violet-otters-rule.md
+++ b/.changeset/violet-otters-rule.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `declaration-block-no-redundant-longhand-properties` autofix for `border-radius` shorthand
--- a/lib/reference/properties.js
+++ b/lib/reference/properties.js
@@ -141,8 +141,8 @@ const longhandSubPropertiesOfShorthandProperties = new Map([
 		'border-radius',
 		new Set([
 			// prettier-ignore
-			'border-top-right-radius',
 			'border-top-left-radius',
+			'border-top-right-radius',
 			'border-bottom-right-radius',
 			'border-bottom-left-radius',
 		]),


ybiquitous (approved):
Thank you. LGTM!

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.js
```diff
@@ -206,6 +206,12 @@ testRule({
 			description: 'autofixer should not mangle css functions with comma separated values',
 			message: messages.expected('transition'),
 		},
+		{
+			code: 'a { border-top-left-radius: 1em; border-top-right-radius: 2em; border-bottom-right-radius: 3em; border-bottom-left-radius: 4em; }',
+			fixed: 'a { border-radius: 1em 2em 3em 4em; }',
+			description: 'explicit border-radius test',
+			message: messages.expected('border-radius'),
+		},
```
Mouvedia, line 214: We will need to add a test for `top-left | top-right-and-bottom-left | bottom-right` and check whether `/` is supported too. 
Thanks for the quick fix.

mattxwang: Hm, I think I'm still misunderstanding something. From my understanding, this rule's autofix only looks at cases like

```css
border-top-left-radius: 1em;
border-top-right-radius: 2em;
border-bottom-right-radius: 3em;
border-bottom-left-radius: 4em;
```

and converts them to

```css
border-radius: 1em 2em 3em 4em;
```

If not all four longhand properties are specified, this rule shouldn't report an error (nor try to do an autofix); is that what you mean?

(i.e. this case [from MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius#syntax), which I don't think is "reachable" from our autofix rules)

```css
/* (first radius values) / top-left | top-right-and-bottom-left | bottom-right */
border-radius: 10px 5px 2em / 20px 25px 30%;
```

Mouvedia: Check #6947 OP, it has a use case for border-width.

e.g. for `border-radius`
```css
border-top-left-radius: 1px;
border-top-right-radius: 4px;
border-bottom-right-radius: 2px;
border-bottom-left-radius: 4px;
```
becomes
```css
border-radius: 1px 4px 2px;
```

mattxwang: Oh, I see what you mean. Is that within scope of the rule? Right now, we don't do any "collapsing", i.e. it would output

```css
border-radius: 1px 4px 2px 4px;
```

Mouvedia: > we don't do any "collapsing"
> Is that within scope of the rule?

It can be a type:enhancement issue for now.
IMHO we should add the cases not covered as tests using `skip: true` and aim to progressively not skip them.

ybiquitous: Interesting. Let's discuss it with a new issue. Could you open an issue? @Mouvedia

Mouvedia:
@mattxwang you might be interested by #8147.

PR #6960, opened 2023-06-23
Document only our standard config

jeddy3:

Follows on from https://github.com/stylelint/stylelint/pull/6937 & https://github.com/stylelint/stylelint/pull/6946

(The use of tables and the introduction of 🔧 for autofix were excellent changes!)

We can improve the onboarding experience for new users by only mentioning our `standard` config in our docs. 

We don't mention our `recommended` config in the first 3 steps of the onboarding process:

- [Getting started](https://stylelint.io/user-guide/get-started)
- [Customising](https://stylelint.io/user-guide/customize)
- [Configuring](https://stylelint.io/user-guide/configure)

Before this pull request, we'd then introduce it in the 4th step ([Rules](https://stylelint.io/user-guide/rules)). At this stage in the onboarding, users are interested in what is _not_ included in the `standard` config that they installed during the getting started step. They want to see at a glance what additional rules they can turn on, and by removing the `recommended` config we declutter the page to make this easier. We also remove any potential for confusion as we now only mention the `standard` config in our docs.

This pull request also reverts to ✅ as:
- we only document one config now
- the polish symbol implies stylistic rules, but we no longer include them

We've been slowly moving away from our historical two-config offering towards a single suggested config that helps people avoid errors and enforce (mostly non-stylistic) conventions. This pull request is part of that effort. People can still use `config-recommended` directly (by discovering the package and inspecting the rules within it), but it's not our suggested approach.

### DIFF
--- a/docs/user-guide/rules.md
+++ b/docs/user-guide/rules.md
@@ -5,11 +5,7 @@ There are over one hundred built-in rules to help you:
 - [avoid errors](#avoid-errors)
 - [enforce conventions](#enforce-conventions)
 
-Many of them are turned on in our [recommended](https://www.npmjs.com/package/stylelint-config-recommended) and [standard](https://www.npmjs.com/package/stylelint-config-standard) configs.
-
-- ⭐️ - [Recommended](https://www.npmjs.com/package/stylelint-config-recommended)
-- 💅 - [Standard](https://www.npmjs.com/package/stylelint-config-standard)
-- 🔧 - Autofixable
+We turn on most of the rules in our [standard config](https://www.npmjs.com/package/stylelint-config-standard) (✅) and many can be autofixed (🔧).
 
 ## Avoid errors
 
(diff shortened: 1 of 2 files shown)


ybiquitous:
> We've been slowly moving away from our historical two-config offering towards a single suggested config that helps people avoid errors and enforce (mostly non-stylistic) conventions. This pull request is part of that effort.

Definitely agreed 👍🏼

ybiquitous (commented):
I agree with reducing the recommended config. 👍🏼 

I just want to confirm one point: is omitting the table header intentional? I feel a bit weird visually. 👀 

image

jeddy3:
Yes, I remove the header to reduce the noise and repetition to make it easier for people to scan the page. I think the key at the top of the page provides enough context.

ybiquitous (approved):
Okay, let's go on. 👍🏼 

We also need to fix the following code in stylelint.io: 
https://github.com/stylelint/stylelint.io/blob/5b11034fdf2296fe7a97e466d38bec682b0c677c/scripts/generate-docs.mjs#L131-L135

jeddy3:
> We also need to fix the following code in stylelint.io:

I'll create an issue.

Issue #6916, opened 2023-06-13
Add `media-query-no-invalid`


romainmenke:
### What is the problem you're trying to solve?

I want to warn on invalid media queries.

see : https://github.com/stylelint/stylelint/issues/6560#issuecomment-1574828599

Types of errors :
1. the `@media` params are not even remotely similar to a media query
   - `isMediaQueryInvalid(x)` will be true
   - e.g. `@media @foo {}`
2. the parenthesis and combinators are incorrect
   - `isGeneralEnclosed(x)` will be true
   - e.g. `@media ((min-width: 300px) and (hover: hover) or (aspect-ratio: 1 / 1))`
3. syntactically incorrect values are used for media queries in general
   - `isGeneralEnclosed(x)` will be true
   - e.g. `@media (min-width: var(--foo))`
   - e.g. `@media (min-width: 50%)`

### What solution would you like to see?

A new rule :
- Name: `media-query-no-invalid`
- Primary option: `true`
- Secondary options: none
- Autofixable: No
- Message: `Unexpected invalid media query "${query}"`
- Description: Disallow invalid media queries.
- Section: "Avoid Errors" -> "Invalid"
- Extended description: something about invalid media queries

ybiquitous:
@romainmenke Thanks for opening the issue. That sounds great to me. 👍🏼 

Let's wait for the feedback. If nobody objects, then let's mark it as ready.

ybiquitous:
Since there are no objections, I've labeled the issue as **ready to implement**. Please consider [contributing](https://stylelint.io/contributing) if you have time.

There are [steps on how to add a new rule](https://stylelint.io/developer-guide/rules#add-a-rule) in the Developer guide.

PR #6963, opened 2023-06-23
Add `media-query-no-invalid`

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/6916

I've added some comments where I thought further explanation might be needed.


lib/rules/media-feature-name-value-no-unknown/index.js
```diff
@@ -239,6 +240,8 @@ const rule = (primary) => {
 			};
 
 			parseMediaQuery(atRule).forEach((mediaQuery) => {
+				if (isMediaQueryInvalid(mediaQuery)) return;
```
romainmenke, line 243: This rule should never warn for invalid media queries.
It should only check feature name and value pairs within grammatically valid media queries.

Otherwise it would overlap with the new rule.

lib/rules/media-feature-name-no-unknown/__tests__/index.js
```diff
@@ -62,6 +62,9 @@ testRule({
 		{
 			code: '@media () { }',
 		},
+		{
+			code: '@media (grid: 1) {}',
+		},
```
romainmenke, line 67: Extra test coverage to check the boundaries of this rule vs. the new rule.

lib/rules/media-feature-no-invalid/README.md
```diff
@@ -0,0 +1,57 @@
...
+Disallow invalid media queries.
+
+<!-- prettier-ignore -->
+```css
+@media not(min-width: 300px) {}
+/**    ↑
+ * This media query */
+```
+
+Media queries must be grammatically valid according to the [Media Queries Level 5](https://www.w3.org/TR/mediaqueries-5/) specification.
```
romainmenke, line 12: I am unsure if this is the best way to explain this rule.  
I know this feature too well on a deeply technical level, so it isn't easy for me to explain it well for CSS authors.

Suggestions to improve would be very welcome :)

package.json
```diff
@@ -130,7 +130,7 @@
   "dependencies": {
     "@csstools/css-parser-algorithms": "^2.2.0",
     "@csstools/css-tokenizer": "^2.1.1",
-    "@csstools/media-query-list-parser": "^2.1.0",
+    "@csstools/media-query-list-parser": "^2.1.1",
```
romainmenke, line 133: see : https://github.com/csstools/postcss-plugins/blob/main/packages/media-query-list-parser/CHANGELOG.md#211

> Fix parsing of `(width == 100px)`. This was erroneously parsed as a range query and will now instead be a general enclosed node.

lib/rules/media-query-no-invalid/README.md
```diff
@@ -0,0 +1,57 @@
...
+Disallow invalid media queries.
+
+<!-- prettier-ignore -->
+```css
+@media not(min-width: 300px) {}
+/**    ↑
+ * This media query */
+```
+
+Media queries must be grammatically valid according to the [Media Queries Level 5](https://www.w3.org/TR/mediaqueries-5/) specification.
```
romainmenke, line 12: I am unsure if this is the best way to explain this rule.
I know this feature too well on a deeply technical level, so it isn't easy for me to explain it well for CSS authors.

Suggestions to improve would be very welcome :)

docs/user-guide/rules.md
```diff
@@ -62,6 +62,7 @@ Disallow invalid syntax with these (sometimes implicit) `no-invalid` rules.
 | [`no-invalid-double-slash-comments`](../../lib/rules/no-invalid-double-slash-comments/README.md)<br/>Disallow invalid double-slash comments. | ✅ | |
 | [`no-invalid-position-at-import-rule`](../../lib/rules/no-invalid-position-at-import-rule/README.md)<br/>Disallow invalid position `@import` rules. | ✅ | |
 | [`string-no-newline`](../../lib/rules/string-no-newline/README.md)<br/>Disallow invalid newlines within strings. | ✅ | |
+| [`media-query-no-invalid`](../../lib/rules/media-query-no-invalid/README.md)<br/>Disallow invalid media queries. | | |
```
ybiquitous, line 65: [suggest] Can you put it alphabetically, please?

lib/rules/index.js
```diff
@@ -203,6 +203,7 @@ const rules = {
 	'media-feature-name-value-no-unknown': importLazy(() =>
 		require('./media-feature-name-value-no-unknown'),
 	)(),
+	'media-query-no-invalid': importLazy(() => require('./media-query-no-invalid'))(),
```
ybiquitous, line 206: [suggest] Please put it alphabetically also here?

lib/rules/media-query-no-invalid/index.js
```diff
@@ -0,0 +1,158 @@
...
+const ruleName = 'media-query-no-invalid';
+
+const messages = ruleMessages(ruleName, {
+	rejected: (query) => `Unexpected invalid media query "${query}"`,
+});
+
+const HAS_MIN_MAX_PREFIX = /^(?:min|max)-/i;
+
+const meta = {
+	url: 'https://stylelint.io/user-guide/rules/media-feature-no-invalid',
```
ybiquitous, line 28: [suggest] Please fix the typo?

```suggestion
	url: 'https://stylelint.io/user-guide/rules/media-query-no-invalid',
```

Note: I've opened #6965 to check such a typo.

ybiquitous (changes requested):
@romainmenke Thanks for the pull request. This is fantastic work!

I've commented a few on trivial things, but I believe there're no big problems 👍🏼

lib/rules/media-query-no-invalid/README.md
Mouvedia: If I understand correctly this rule checks everything between `@media` and `{`, correct?

lib/rules/media-query-no-invalid/index.js
romainmenke: Thank you!

lib/rules/media-query-no-invalid/README.md
romainmenke: Yes, but at a different level, or from a different perspective than other rules.

This rule doesn't check :
- feature names (`width`, `hover`, ...)
- feature values (`10px`, `more`, `dark`, ...)
- feature name and value pairs (`width: 10px`)

You could say it only checks the structure.

Mouvedia: Alright then, if I had to suggest something Id say add a reminder that list the rules that check those.
As a user Id expect this rule to call them internally.

ybiquitous: Surely, it might be helpful to provide such a reminder, e.g., `media-feature-name-no-unknown`, `media-feature-name-value-no-unknown`, etc.

romainmenke: Good idea, listing the rules that cover other parts makes it clearer.

I've updated the rule documentation.

lib/rules/media-query-no-invalid/index.js
```diff
@@ -0,0 +1,158 @@
...
+				if (isMediaQueryInvalid(mediaQuery)) {
+					// Queries that fail to parse are invalid.
+					invalidNodes.push(mediaQuery);
+
+					return;
+				}
+
+				mediaQuery.walk((entry) => {
+					// All general enclosured nodes are invalid.
+					if (isGeneralEnclosed(entry.node)) {
```
Mouvedia, line 54: minor

add `const node = entry.node;`
it's being accessed 8 times

lib/rules/media-feature-no-invalid/README.md
Mouvedia: If you think that this rule will be updated immediately after the release of the hypothetical level 6 specification then you can leave it as is.
If it's unlikely then something more vague would give us more leeway if that happened.

ybiquitous (approved):
Thanks, looks good 👍🏼

lib/rules/media-query-no-invalid/index.js
romainmenke: Thank you, I will keep this in mind!

PR #6964, opened 2023-06-23
Fix `selector-pseudo-class-no-unknown` false negatives for pseudo-elements with matching names

### LINKED ISSUE #6878: Fix `selector-pseudo-class-no-unknown` false negatives for non-standard `:slotted`

```css
:slotted(.foo) {}
```

```json
{
   "rules": {
     "selector-pseudo-class-no-unknown": true
   }
}
```

https://stylelint.io/demo/

Unexpected unknown pseudo-class selector ":slotted"

### What actually happened?

no error

### Proposal to fix the bug

Distinguish between the pseudo-class—which is vue specific—and the standard pseudo-element.

ref:
- pseudo-class: https://vuejs.org/api/sfc-css-features.html#slotted-selectors
- pseudo-element: https://drafts.csswg.org/css-scoping/#slotted-pseudo

Mouvedia:

Closes #6878

I have added a test for `:placeholder` because `:-moz-placeholder` and `:-ms-input-placeholder` exist.
i.e. it's a good example of something that the rule will pick up from now on

### DIFF
--- a/.changeset/great-feet-brake.md
+++ b/.changeset/great-feet-brake.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-pseudo-class-no-unknown` false negatives for pseudo-elements with matching names
--- a/lib/rules/selector-pseudo-class-no-unknown/index.js
+++ b/lib/rules/selector-pseudo-class-no-unknown/index.js
@@ -7,6 +7,7 @@ const isStandardSyntaxRule = require('../../utils/isStandardSyntaxRule');
 const isStandardSyntaxSelector = require('../../utils/isStandardSyntaxSelector');
 const {
 	atRulePagePseudoClasses,
+	levelOneAndTwoPseudoElements,
 	pseudoClasses,
 	pseudoElements,
 	webkitScrollbarPseudoClasses,
@@ -68,8 +69,13 @@ const rule = (primary, secondaryOptions) => {
 						return;
 					}
 
-					// Ignore pseudo-elements
-					if (value.slice(0, 2) === '::') {
+					if (value.startsWith('::')) {
+						return;
+					}
+
+					const name = value.replace(/^:*/, '').toLowerCase();
+
+					if (levelOneAndTwoPseudoElements.has(name)) {
 						return;
 					}
 
@@ -77,8 +83,8 @@ const rule = (primary, secondaryOptions) => {
 						return;
 					}
 
+					const hasVendorPrefix = vendor.prefix(name);
 					let index = null;
-					const name = value.slice(1).toLowerCase();
 
 					if (isAtRule(node) && node.name === 'page') {
 						if (atRulePagePseudoClasses.has(name)) {
@@ -86,8 +92,10 @@ const rule = (primary, secondaryOptions) => {
 						}
 
 						index = atRuleParamIndex(node) + pseudoNode.sourceIndex;
+					} else if (pseudoElements.has(name) && !hasVendorPrefix) {
+						index = pseudoNode.sourceIndex;
 					} else {
-						if (vendor.prefix(name) || pseudoClasses.has(name) || pseudoElements.has(name)) {
+						if (hasVendorPrefix || pseudoClasses.has(name)) {
 							return;
 						}
 
--- a/lib/rules/selector-pseudo-class-no-unknown/__tests__/index.js
+++ b/lib/rules/selector-pseudo-class-no-unknown/__tests__/index.js
@@ -203,6 +203,22 @@ testRule({
 			endColumn: 7,
 		},
 		{
+			code: ':slotted {}',
+			message: messages.rejected(':slotted'),
+			line: 1,
+			column: 1,
+			endLine: 1,
+			endColumn: 9,
+		},
+		{
+			code: ':placeholder {}',
+			message: messages.rejected(':placeholder'),
+			line: 1,
+			column: 1,
+			endLine: 1,
+			endColumn: 13,
+		},
+		{
 			code: '@page :blank:unknown { }',
 			message: messages.rejected(':unknown'),
 			line: 1,


package.json
```diff
@@ -46,10 +46,11 @@
     "release": "np --no-release-draft",
     "pretest": "npm run lint",
     "test": "node --experimental-vm-modules node_modules/jest/bin/jest.js",
-    "test-coverage": "npm run test --ignore-script -- --coverage",
+    "test:only": "npm test --ignore-scripts",
```
Mouvedia, line 49: notice the `s`

ybiquitous: This change makes `npm run watch` not work: `jest.js` doesn't receive `--watch`.

```console
$ npm run watch -- lib/rules/selector-pseudo-class-no-unknown

> stylelint@15.9.0 watch
> npm run test:only -- --watch lib/rules/selector-pseudo-class-no-unknown


> stylelint@15.9.0 test:only
> npm test --ignore-scripts --watch lib/rules/selector-pseudo-class-no-unknown


> stylelint@15.9.0 test
> node --experimental-vm-modules node_modules/jest/bin/jest.js lib/rules/selector-pseudo-class-no-unknown
```

So, could you revert this change for `test:only` in `package.json`?

In addition, `test-coverage` instead of `test:coverage` is intentional. It is never combined like `npm-run-all lint:*`.

.changeset/great-feet-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: reject unknown pseudo-classes even if their names match with existing pseudo-elements
```
ybiquitous, line 5: [suggest] Thanks for creating the changelog. Can you update it for consistency? See also our [convention](https://stylelint.io/maintainer-guide/pull-requests#merging).

```suggestion
Fixed: `selector-pseudo-class-no-unknown` false negatives for non-standard `:slotted` and `:placeholder`
```

Note: I've updated the PR title.

lib/rules/selector-pseudo-class-no-unknown/index.js
```diff
@@ -68,8 +69,13 @@ const rule = (primary, secondaryOptions) => {
...
+					const name = value.replace(/^:*/, '').toLowerCase();
+					const isPseudoElement =
+						value.slice(0, 2) === '::' || levelOneAndTwoPseudoElements.has(name);
+					const hasVendorPrefix = vendor.prefix(name);
+
 					// Ignore pseudo-elements
-					if (value.slice(0, 2) === '::') {
+					if (isPseudoElement) {
 						return;
 					}
```
ybiquitous, line 80: [suggest] For performance, can you evaluate expressions lazily as possible like this?

```js
if (value.startsWith('::')) {
	return;
}

const name = value.replace(/^:*/, '').toLowerCase();

if (levelOneAndTwoPseudoElements.has(name)) {
	return;
}

if (optionsMatches(secondaryOptions, 'ignorePseudoClasses', pseudoNode.value.slice(1))) {
	return;
}

const hasVendorPrefix = vendor.prefix(name);
let index = null;
```

ybiquitous (changes requested):
@Mouvedia Thanks for the pull request! Can you check my reviews?

.changeset/great-feet-brake.md
Mouvedia: "for non-standard `:slotted` and `:placeholder`" is too specific
Ill try to come up with something better

package.json
Mouvedia: It will have to be fixed but in a separate PR.
AFAIK `--ignore-script` doesn't exist.

.changeset/great-feet-brake.md
ybiquitous: Make sense. I've updated the PR title also.

ybiquitous (approved):
Thank you. LGTM 👍🏼

lib/rules/selector-pseudo-class-no-unknown/index.js
```diff
@@ -68,26 +69,33 @@ const rule = (primary, secondaryOptions) => {
...
+						return;
+					}
+
+					const name = value.replace(/^:*/, '').toLowerCase();
+
+					if (levelOneAndTwoPseudoElements.has(name)) {
 						return;
 					}
 
 					if (optionsMatches(secondaryOptions, 'ignorePseudoClasses', pseudoNode.value.slice(1))) {
```
Mouvedia, line 82: @ybiquitous I am seeing another small refactor here.
Ill add it too.

Mouvedia, line 82: It cannot be replaced by `name` because of the lowercasing.

package.json
ybiquitous: See https://docs.npmjs.com/cli/v9/commands/npm-test#ignore-scripts

lib/rules/selector-pseudo-class-no-unknown/index.js
ybiquitous: [suggest] Ah, I missed it. How about the following?

```js
const name = value.replace(/^:*/, '');
const lowerName = value.toLowerCase();

// ...

if (optionsMatches(secondaryOptions, 'ignorePseudoClasses', name)) {
```

package.json
Mouvedia: Iv read it twice I see `--ignore-scripts` not `--ignore-script`.

ybiquitous: Oh, I missed it, sorry. 🙇🏼 
I'll open PR soon.

ybiquitous: Opened #6967.

Mouvedia: If so please add a script to run `test` without `pretest`.

lib/rules/selector-pseudo-class-no-unknown/index.js
Mouvedia: I think it's fine as is, I prefer not having to drill to find the blame for this line.
The less diff the better.

ybiquitous: Okay, leave it as-is. 👍🏼

package.json
ybiquitous: Make sense. Opened #6969.

jeddy3 (approved):
LGTM, thank you.

PR #6965, opened 2023-06-23
Add test case to check `rule.meta.url` more precisely

ybiquitous:

None.

For example, if a URL were incorrect, the test case would fail:

```
  ● all rules › "alpha-value-notation" should have metadata

    expect(received).toBe(expected) // Object.is equality

    Expected: "https://stylelint.io/user-guide/rules/alpha-value-notation"
    Received: "https://stylelint.io/user-guide/rules/alpha-value-notation_"
```

### DIFF
--- a/lib/rules/__tests__/index.test.js
+++ b/lib/rules/__tests__/index.test.js
@@ -17,9 +17,9 @@ beforeAll(async () => {
 });
 
 describe('all rules', () => {
-	test.each(ruleEntries)('"%s" should have metadata', (_name, rule) => {
+	test.each(ruleEntries)('"%s" should have metadata', (name, rule) => {
 		expect(rule.meta).toBeTruthy();
-		expect(rule.meta.url).toMatch(/^https:\/\/stylelint\.io\/.+/);
+		expect(rule.meta.url).toBe(`https://stylelint.io/user-guide/rules/${name}`);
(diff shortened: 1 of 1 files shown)


lib/rules/__tests__/index.test.js
```diff
@@ -17,9 +17,9 @@ beforeAll(async () => {
 });
 
 describe('all rules', () => {
-	test.each(ruleEntries)('"%s" should have metadata', (_name, rule) => {
+	test.each(ruleEntries)('"%s" should have metadata', (name, rule) => {
 		expect(rule.meta).toBeTruthy();
-		expect(rule.meta.url).toMatch(/^https:\/\/stylelint\.io\/.+/);
+		expect(rule.meta.url).toBe(`https://stylelint.io/user-guide/rules/${name}`);
```
ybiquitous, line 22: [note] A string is more readable than a regex.

lib/rules/__tests__/index.test.js
```diff
@@ -69,7 +69,7 @@ describe('custom message option', () => {
 	);
 });
 
-describe('recommended and standard configs', () => {
+describe('standard config', () => {
```
ybiquitous, line 72: see #6960

mattxwang (approved):
LGTM!

PR #6970, opened 2023-06-24
Refactor test file `rules/__tests__/index`

ybiquitous:

Ref: #5291

- Rewrite to ESM (`.mjs`)
- Include missing `stylelint-config-recommended`


lib/rules/__tests__/index.test.mjs
```diff
@@ -0,0 +1,114 @@
...
+	afterAll(async () => {
+		await rm(tmpDir, { recursive: true, force: true });
+	});
+
+	test('tue rules are not empty', () => {
+		expect(standardRules).not.toHaveLength(0);
+	});
+
+	test('the rules are in the rules doc', () => {
+		for (const name of standardRules) {
```
ybiquitous, line 110: [note] Unfortunately, `test.each()` is unavailable for variables assigned by async/await.

lib/rules/__tests__/index.test.mjs
```diff
@@ -0,0 +1,114 @@
...
+		);
+
+		const configRules = (name) => {
+			const config = require(path.join(tmpDir, 'node_modules', name));
+
+			return Object.keys(config.rules);
+		};
+
+		standardRules = configRules('stylelint-config-standard');
+		standardRules.push(...configRules('stylelint-config-recommended'));
```
ybiquitous, line 98: [note] We still need to explicitly include `stylelint-config-recommended`. If `stylelint-config-recommended` is removed in the future, we can also remove this code.

jeddy3 (approved):
LGTM, thank you.

PR #6971, opened 2023-06-24
Refactor to remove no longer used `blurComments()` utility

ybiquitous:

Directly none. But removing unused files is better for ESM migration (#5291).

No, it's self-explanatory.


jeddy3 (approved):
LGTM, thank you.

There will probably be more of these when we remove the deprecated rules, and therefore the need for stylesearch.

Mouvedia:
Was this reported by a tool?
If not, it should not be the result of a manual search.

ybiquitous:
> There will probably be more of these when we remove the deprecated rules, and therefore the need for stylesearch.

Yes, I think so. 👍🏼 

> Was this reported by a tool?
> If not, it should not be the result of a manual search.

No, I found it by accident when rewriting tests to ESM. Do you know any tool to detect unused modules?

Mouvedia:
Do you know any tool to detect unused modules?

@ybiquitous https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-unused-modules.md#options

ybiquitous:
@Mouvedia Thanks for sharing the ESLint rule. But the rule doesn't seem to work with the config below: 🤔 

```diff
diff --git a/package.json b/package.json
index c3e982806..753791edb 100644
--- a/package.json
+++ b/package.json
@@ -59,18 +59,26 @@
   },
   "prettier": "@stylelint/prettier-config",
   "eslintConfig": {
+    "plugins": [
+      "eslint-plugin-import"
+    ],
     "extends": [
       "stylelint",
       "stylelint/jest"
     ],
+    "rules": {
+      "import/no-unused-modules": [
+        "error",
+        {
+          "unusedExports": true
+        }
+      ]
+    },
     "overrides": [
       {
         "files": [
           "**/*.mjs"
         ],
-        "plugins": [
-          "eslint-plugin-import"
-        ],
         "rules": {
           "import/extensions": [
             "error",
```

```console
$ cat lib/utils/foo.js
module.exports = () => 1;

$ npx eslint lib/utils/foo.js
(no output)
```

Do I miss something?

Mouvedia:
@ybiquitous I am not sure, maybe the default—`process.cwd()`—for `src` is not appropriate in our case?
The name of the rule sure sounds like what we want though.
All I am saying is that, there must be tool that does it. Hopefully it would be an ESLint plugin.

Mouvedia:
@ybiquitous I am not sure, <strike>maybe the default—process.cwd()—for src is not appropriate in our case?</strike>
is it the fast that you have `"**/*.mjs"` but your file extension is `.js`?
The name of the rule sure sounds like what we want though.
All I am saying is that, there must be tool that does it. Hopefully it would be an ESLint plugin.

ybiquitous:
The same problem is present even for `.mjs` 🤔 

But thanks for the advice. Please let me know if you'd notice anything later.

Mouvedia:
@ybiquitous you used `module.exports` but [all their examples](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-unused-modules.md#example-for-unused-exports) use `export`.
Well anyway, it's not a priority.

PR #6972, opened 2023-06-24
Refactor to remove no longer used `blurFunctionArguments()` utility

ybiquitous:

Ref: #6971

The `balanced-match` package used in `blurFunctionArguments()` is still used in other modules:
https://github.com/stylelint/stylelint/blob/2bc9e50f63440dc9ef000fa66fbee29ca9a30ff1/lib/utils/functionArgumentsSearch.js#L3


jeddy3 (approved):

Mouvedia:
In your next refactor, can you include using `readJSONFile` [there](https://github.com/stylelint/stylelint/blob/main/lib/rules/function-no-unknown/index.js#L52)?

ybiquitous:
Ah, changing rule codes is going to be a while yet.

Issue #6753, opened 2023-04-02
Document stylistic plugin in migration guide


elirasza:
Hello, just to let you know I have created and published a plugin containing all the stylistic rules that were removed in 15.0.0 : [stylelint-stylistic](https://www.npmjs.com/package/stylelint-stylistic). It still under development but everything should work as before :)

I saw in your [migration guide](https://stylelint.io/migration-guide/to-15/) you wanted to be noticed once such a package were avaiblable, so here we are !

ybiquitous:
@elirasza Great! We appreciate your work! 👏🏼 

Could you add your stylistic plugins to our [awesome list](https://github.com/stylelint/awesome-stylelint) if you don't mind?

elirasza:
Thanks for the quick reply. I have sent a PR !

kerryj89:
I just now saw this, thanks so much @elirasza!  It might be nice if this were part of the migration notes but I guess it's not "official" and vetted enough to recommend there.  This would've made it so much easier.

ybiquitous:
@kerryj89 Nice catch. I agree we would recommend this plugin in the migration guide. 👍🏼 
If you're interested in contributing to Stylelint, could you try PR to improve the migration guide?

GergelyKalmar:
I'd like to mention that the plugin is not very compliance department-friendly at the moment – the repo is owned by a user that is brand new and was seemingly created with the purpose of publishing that one package. From a supply chain security perspective it's not nearly as assuring as the current built-in features that come with Stylelint itself.

elirasza:
I'm glad to help ! This is indeed my first take on open-source, so this account is brand new. I am more used to my company's proprietary codebases so I'm still a noob in understanding how the community works.

I also did not quite understand the underlying reasons of why stylistic rules were removed, but I'd say it does make sense from a single-responsibility point of view : keep stylelint a strict linter, while betting on plugins or other tools for formatting.

Maybe an "official" stylistic plugin would have been a good compromise, but maybe is it not in the stylelint team priorities ?

cenfun:
Actually stylelint becomes even more useless without stylstic-rules, so we also hope that stylstic-rules will be an official plugin.
As we know, stylstic-rules and Prettier have some conflicts, but this should not cause stylstic-rules to be retired, and then need to use another additional tool. We should resolve these conflicts in a mutually beneficial way.

ybiquitous:
[...]

PR #6973, opened 2023-06-24
Document stylistic plugin in migration guide

jeddy3:

Closes https://github.com/stylelint/stylelint/issues/6753

Let's list both for now so that they are at least discoverable by users. If the plugin authors consolidate efforts and one package is deprecated, we can update the migration guide again.


ybiquitous (approved):
So helpful! LGTM 👍🏼

PR #6977, opened 2023-06-25
Fix `media-feature-name-no-vendor-prefix` positions for `*-device-pixel-ratio`

romainmenke:
> Which issue, if any, is this issue related to?

Closes https://github.com/stylelint/stylelint/issues/6975

> Is there anything in the PR that needs further explanation?

No, it's self-explanatory.

PR #3397, opened 2018-06-14
Fix level 5 media queries in media-feature-name-no-unknown

pkuczynski:
Fixes #3396

PR #6978, opened 2023-06-25
Fix `media-feature-name-no-unknown` false positives

romainmenke:

Closes https://github.com/stylelint/stylelint/pull/6977

I found 4 missing features: 
- `environment-blending`
- `nav-controls`
- `prefers-reduced-data`
- `video-color-gamut`


mattxwang (approved):
LGTM!

Re:

> The commit message and changelog entry still need to be changed to reflect this, but listing all four seemed verbose. Thoughts?

My initial thoughts are that listing all four seems reasonable, even if a bit verbose. Are these all missing for a reason (ex were only introduced recently)? Some of the past "false positives" changelog messages are a bit more generic, ex #3397

> Fix level 5 media queries in media-feature-name-no-unknown

ybiquitous (approved):
Thanks, LGTM 👍🏼 

I agree with @mattxwang:

> My initial thoughts are that listing all four seems reasonable, even if a bit verbose.

If just `Fixed: X false positives`, people seeing the changelog may be confused.

PR #6979, opened 2023-06-26
Remove 76 rules deprecated in 15.0.0

### LINKED ISSUE #6961: Remove deprecated rules
As of v15.0.0, we have deprecated many stylistic rules. And, as we announced when the v15.0.0 release, we will remove these rules in the next major version (maybe v16.0.0).

Additionally, as #6753 introduced, the Stylelint community has provided 3rd-party plugins as an alternative, e.g.

- https://github.com/elirasza/stylelint-stylistic

So, I think we can perform this removal without a big confusion.

See also:

- https://github.com/stylelint/stylelint/blob/e56fc6cfdd7c9dab342a1289ab5a596b29c7be57/docs/migration-guide/to-15.md?plain=1#L12-L14
- https://github.com/stylelint/stylelint/blob/e56fc6cfdd7c9dab342a1289ab5a596b29c7be57/docs/user-guide/rules.md?plain=1#L334-L336

mattxwang:

Closes #6961.

I apologize for how large this change is 😅 I'm not sure if this is easier to review (since it is clean + accomplishes it in one go), or if we'd prefer one that splits them up piecemeal (e.g. into sections) - let me know what we'd prefer!

---

I'll quickly explain the methodology for this PR:

1. remove each `lib/rules/*` for each deprecated test
2. remove each (now invalid) path from `lib/rules/index.js`
3. fix all broken tests, which fall into one of several categories:
    - tests that use a deprecated rule as an arbitrary rule; to resolve this, I just picked a different rule (usually `color-named` for `color-hex-case` and `number-max-precision` for `indentation`)
        - one of these is about a custom plugin, so I've rewritten the plugin
    - tests that are *about* deprecated rules; I added a `.skip` (have left comments in the review)
    - fs tests that use deprecated rules; I removed these from the config, then updated the snapshot.
      - the zen garden one doesn't work "out of the box" anymore, i.e. the snapshot has nontrivially changed because of indentation/spacing rules
    - `indentation`-specific behaviour (it being last); I deleted these!
4. remove all documentation related to removed rules (mostly `indentation`)
5. remove all unused utilities (via code coverage, not manual inspection)


lib/__tests__/standalone-quiet-deprecation-warnings.test.mjs
mattxwang, line 1: I wasn't sure if I should delete this file - any thoughts?

lib/rules/__tests__/index.test.mjs
```diff
@@ -46,7 +46,8 @@ describe('fixable rules', () => {
 	});
 });
 
-describe('deprecated rules', () => {
+// eslint-disable-next-line jest/no-disabled-tests
+describe.skip('deprecated rules', () => {
```
mattxwang, line 50: Now, the deprecated rules array is empty, so this ends up being an error. Should we leave this as skipped, delete the test entirely, etc>?

lib/__tests__/standalone-fix.test.mjs
```diff
@@ -39,35 +37,16 @@ it('fixes when enabled in config', async () => {
...
-			},
-		},
-		fix: true,
-	});
-
-	expect(result.output).toBe(
-		'a {\n  box-shadow: 0 -1px 0 0 rgba(0, 0, 0, 0.1),\n    0 0 0 1px rgba(0, 0, 0, 0.2),\n    inset 0 1px 2px 0 rgba(0, 0, 0, 0.1);\n}',
-	);
-});
-
```
mattxwang, line 58: This seems like an `indentation`-specific test, so I deleted it. I can re-add a test if there's a more general behaviour we're testing!

lib/__tests__/standalone-fix.test.mjs
```diff
@@ -117,52 +94,6 @@ it("doesn't fix with multiple scoped stylelint-disable commands", async () => {
...
-		fix: true,
-	});
-
-	expect(result.output).toBe(stripIndent`
-		/* stylelint-disable indentation */
-		a {
-			color: #fff;
-		}`);
-});
-
```
mattxwang, line 165: These feel like duplicate tests, but if there's a general principle we're trying to test, I can rewrite these too!

ybiquitous:
How about creating a `v16` branch? Maybe we should keep the `main` branch for minor changes like bugfixes.

docs/user-guide/rules.md
```diff
@@ -388,196 +388,4 @@ Require or disallow whitespace on the inside with this `whitespace-inside` rule.
 | | | |
 | :-- | :-: | :-: |
 | [`comment-whitespace-inside`](../../lib/rules/comment-whitespace-inside/README.md)<br/>Require or disallow whitespace on the inside of comment markers. | ✅ | 🔧 |
-
-## Deprecated
-
-These rules are deprecated — we won't fix bugs nor add options, and we will remove them in the next major release. We recommend you use a pretty printer (like Prettier) alongside Stylelint rather than these rules.
```
mattxwang, line 394: Removing this header triggers the lint error in CI (since links to `#deprecated` will now be broken). Do we want to keep this header here? If so, any suggestions on what to include?

ybiquitous: I agree with removing the `## Deprecated` header. But it seems good that `to-15.md` will include the removed rules list, instead of the links to `rules.md`.

https://github.com/stylelint/stylelint/blob/5be9a7b996d7846c06fdec00208b27b3c7544727/docs/migration-guide/to-15.md?plain=1#L14

mattxwang: Gotcha - just to confirm, should I then create a removed rules list in the `to-15` migration guide?

ybiquitous: Yes. If so, I believe people could easily find the deprecated rules with v15 later.

mattxwang: Thanks for the suggestion; I gave a first-draft approach in https://github.com/stylelint/stylelint/pull/6979/commits/9074d968a699aee4170629bf7e4791543b766da4; let me know what you think!

docs/migration-guide/to-15.md
```diff
@@ -165,3 +165,179 @@ If you would like to keep the previous behavior, you should change your config t
   ]
 }
 ```
+
+## List of Deprecated Rules
```
ybiquitous, line 169: [suggest] Sorry for the minor suggestion, but how about keeping header consistency?

```suggestion
## List of deprecated rules
```

docs/migration-guide/to-15.md
```diff
@@ -165,3 +165,179 @@ If you would like to keep the previous behavior, you should change your config t
...
+
+These rules are deprecated - we removed them in v16. We recommend you use a pretty printer (like Prettier) alongside Stylelint rather than these rules.
```
ybiquitous, line 171: [suggest] I have minor suggestions:

- how about `x.y.z` format for consistency (e.g. `Migrating to 15.0.0`)?
- how about removing the details because Prettier is mentioned in the section above?

```suggestion
These rules are deprecated - we removed them in 16.0.0.
```

Anyway, thanks for moving the deprecated list. 😊 👍🏼

lib/rules/__tests__/index.test.mjs
ybiquitous: I want to leave the case for future deprecated rules. How about leaving a reason as a comment like this?

```js
// eslint-disable-next-line jest/no-disabled-tests -- Leave the case for future deprecated rules.
```

lib/__tests__/standalone-quiet-deprecation-warnings.test.mjs
ybiquitous: I think we shouldn't delete this file because this test is for the feature, not depending on specific rules.

We can use the Jest mocking for that purpose like this:

https://github.com/stylelint/stylelint/blob/44b238bd7f1cf30941e7f514232d592551721c89/lib/__tests__/standalone-deprecations.test.mjs#L7-L10

I can confirm the following mocking works:

```js
import { jest } from '@jest/globals';

jest.mock('../rules/block-no-empty');

const require = createRequire(import.meta.url);
const deprecatedRule = require('../rules/block-no-empty/index.js');
deprecatedRule.mockImplementation(() => {
	return (root, result) => {
		report({
			ruleName: 'block-no-empty',
			message: 'Foo bar',
			node: root,
			result,
		});
	};
});
deprecatedRule.meta = { deprecated: true };
```

lib/__tests__/standalone-fix.test.mjs
ybiquitous: It appears that this test shouldn't depend on specific rules. 🤔 
Is it testing how multiple rules' autofix works?

lib/__tests__/standalone-fix.test.mjs
ybiquitous: I don't think we shouldn't remove these cases. Could you rewrite them by using existing rules? And, we should make `describe` more generic.

ybiquitous:
[ask] Just to confirm about `Part of #6961`, do you have a plan for follow-up PRs?

lib/__tests__/standalone-fix.test.mjs
jeddy3: > Is it testing how multiple rules' autofix works?

I believe it's testing the workaround we added so that the `indentation` rule runs last. (It was one of the shortcomings of the standalone stylistic rule approach we took.)

I think we're good to delete the test.

jeddy3:
I've pushed a commit to move the list of deprecated rules into the corresponding section of the migration guide. We can include a link to the section from our migrating to 16 guide.

.changeset/rare-squids-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": major
+---
+
+Removed: deprecated rules
```
jeddy3, line 5:
```suggestion
Removed: 76 rules deprecated in 15.0.0
```

jeddy3 (commented):
Can any of the dependencies be removed? I suspect [style-search ](https://github.com/davidtheclark/style-search) is redundant, as only the older stylistic rules used it.

ybiquitous:
> I suspect [style-search ](https://github.com/davidtheclark/style-search) is redundant, as only the older stylistic rules used it.

Agree. 👍🏼 

FYI, here's the output of the tool [`depcheck`](https://github.com/depcheck/depcheck):

```console
$ npx depcheck@latest
...
Unused dependencies
* style-search
...
```

(also `@types/style-search`)

Mouvedia:
Since no one was against tackling them all at once, I surmise this PR can be reviewed.

docs/migration-guide/to-15.md
```diff
@@ -11,7 +11,84 @@ Two significant changes may affect you:
...
+- `selector-pseudo-class-parentheses-space-inside`
+- `selector-pseudo-element-case`
+- `string-quotes`
+- `unicode-bom`
+- `unit-case`
+- `value-list-comma-newline-after`
+- `value-list-comma-newline-before`
+- `value-list-comma-space-after`
+- `value-list-comma-space-before`
+- `value-list-max-empty-lines`
```
Mouvedia, line 91: Maybe add a sentence with a link to https://stylelint.io/user-guide/cli/#--quiet-deprecation-warnings?

docs/user-guide/rules.md
```diff
@@ -388,196 +388,4 @@ Require or disallow whitespace on the inside with this `whitespace-inside` rule.
 | | | |
 | :-- | :-: | :-: |
 | [`comment-whitespace-inside`](../../lib/rules/comment-whitespace-inside/README.md)<br/>Require or disallow whitespace on the inside of comment markers. | ✅ | 🔧 |
-
-## Deprecated
```
Mouvedia, line 392: questions

Are the plugins storing the documentation of the rules? (stylistic and codeguide)
If not could we at least archive all the READMEs of the removed rules in a folder somewhere?

lib/__tests__/standalone-fix.test.mjs
```diff
@@ -39,35 +37,16 @@ it('fixes when enabled in config', async () => {
 	expect(output).toBe('a { color: #fff; }');
 });
 
-it('apply indentation autofix at last', async () => {
```
Mouvedia, line 42: question

If I understand correctly there are no more rules where order matters?
i.e. this was limited to indentation only

Mouvedia:
You forgot a mention of `indentation` [here](https://github.com/stylelint/stylelint/blob/main/docs/user-guide/options.md?plain=1#L45). You should grep the rules on the docs folder.

Mouvedia:
You forgot a mention of `indentation` [here](https://github.com/stylelint/stylelint/blob/main/docs/user-guide/options.md?plain=1#L45). You should grep the rules on the docs folder.
Also depending on the number of tests left you will have to edit [that line](https://github.com/stylelint/stylelint/blob/main/README.md?plain=1#L18).

docs/user-guide/rules.md
ybiquitous: Do you mean the following?

https://github.com/elirasza/stylelint-stylistic/blob/0.4.2/lib/rules/at-rule-name-case/README.md

lib/__tests__/standalone-fix.test.mjs
mattxwang: Sounds good; I've tried to replicate what feels like the spirit of the test in https://github.com/stylelint/stylelint/pull/6979/commits/3556d6d7725e08bd072fdb58fdb74d4725001cc2; let me know if this is what we were looking for!

docs/migration-guide/to-15.md
mattxwang: Since the flag was only introduced past `15.0.0`, would it make sense to add a sentence in the `to-15` guide? And/or any suggestions on how to phrase the sentence?

mattxwang:
Thanks everyone for the timely feedback, appreciate it! I think I've addressed all of the first-round pieces of feedback:

- re-adding the deleted tests (using @ybiquitous's mock for `--quiet-deprecation-warnings` + add comments for disabled tests
- removed `style-search`
- fix an errant `indentation` in a test case + in a doc
- writes a very small version 16 migration guide with the suggested sentence (leaving "space" open for future breaking changes, like the node version, etc.)

Any new thoughts? I haven't written any of the migration guides, so happy to make any adjustments there if necessary!

lib/__tests__/standalone-fix.test.mjs
jeddy3: There are some, e.g. [`property-no-redundant-values`](https://stylelint.io/user-guide/rules/shorthand-property-no-redundant-values) needs to run after [`declaration-block-no-redundant-longhand-properties`](https://stylelint.io/user-guide/rules/declaration-block-no-redundant-longhand-properties), but users less often encounter problems. Many stylistic rules caused `indentation` warnings, so we forced it to run last as it would normally run early on because it starts with an "i".

docs/migration-guide/to-15.md
jeddy3: People would benefit from knowing about that option now, less so when we release 16.0.0. I'll open a PR to add it.

lib/rules/__tests__/index.test.mjs
```diff
@@ -46,7 +46,8 @@ describe('fixable rules', () => {
 	});
 });
 
-describe('deprecated rules', () => {
+// eslint-disable-next-line jest/no-disabled-tests --- Leaving for future cases of deprecated rules.
+describe.skip('deprecated rules', () => {
```
ybiquitous, line 50: [ask] Is it possible to remove `skip` via mocking like `jest.mock('../rules/block-no-empty')`?

lib/__tests__/standalone-fix.test.mjs
Mouvedia: If you can remember please list them here, so that we can refer to the comment if we ever need to document it.

jeddy3: I'll jot them down if I remember any more, but can't offhand at the moment.

docs/migration-guide/to-16.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Breaking changes
+
+This breaking change may affect you:
+
+- removed deprecated stylistic rules
+
+## Removed Deprecated Stylistic Rules
+
+We've removed the stylistic rules we deprecated in 15.0.0. See the [15.0.0 migration guide](./to-15.md) for details of how to migrate.
```
Mouvedia, line 13: minor

> for details of how to migrate

Maybe that's a US thing, but I would prefer _for details on how to migrate_.

docs/migration-guide/to-16.md
```diff
@@ -0,0 +1,13 @@
...
+
+This release contains breaking changes.
+
+## Breaking changes
+
+This breaking change may affect you:
+
+- removed deprecated stylistic rules
+
+## Removed Deprecated Stylistic Rules
```
jeddy3, line 11:
```suggestion
## Removed deprecated stylistic rules
```

docs/migration-guide/to-16.md
```diff
@@ -0,0 +1,13 @@
+# Migrating to 16.0.0
+
+This release contains breaking changes.
+
+## Breaking changes
+
+This breaking change may affect you:
+
+- removed deprecated stylistic rules
```
jeddy3, line 9:
```suggestion
```

We can add the index when we have more than one change. Let's keep it simple for now.

jeddy3 (commented):
Last two suggestion from me, otherwise LGTM.

lib/rules/__tests__/index.test.mjs
mattxwang: Right now, the test checks the README for each rule; would the idea be to mock the rule's README? It feels a bit strange since it doesn't feel like the mock helps test the real behaviour at all (whether or not we documented the deprecation within the rule's README).

ybiquitous: Ah, that makes sense. 👍🏼 Let's leave it as-is.

ybiquitous (approved):
LGTM! 👍🏼

jeddy3:
> Let's address these unused files with follow-up pull requests!

SGTM.

PR #6983, opened 2023-06-26
Fix inconsistent `errored` properties in `stylelint.lint()` return value

### LINKED ISSUE #6982: Fix inconsistent `errored` properties in `stylelint.lint()` return value

1. Enable option ```reportNeedlessDisables```.
2. Create an unused disable comment.
3. Start ```stylelint.lint```.
4. Look at ```results.source.errored```.

I leave a link to the repository where I reproduced the error: https://github.com/dvun0g/issue-stylelint-reportNeedlessDisables-not-error-source.

Description:
When running ```stylelint.list```, if the file contains only ```reportNeedlessDisables``` errors, the errored field will be ```false```. You can clearly see in https://github.com/dvun0g/issue-stylelint-reportNeedlessDisables-not-error-source/blob/trunk/results.stylelint.json.

Any, i used this
```
{
}
```

Start stylelint using the Node JS API ```stylelint.lint```.

```
{
     "stylelint": "^15.9.0",
     "stylelint-config-standard": "^33.0.0"
}
```

It was expected that the errored field would be set to ```true```, if there are ```reportNeedlessDisables``` errors in the file.

### What actually happened?

Field errored equal ```false```.

No, I use css.

### Proposal to fix the bug

_No response_

ybiquitous:

Closes #6982

This pull request fixes the `errored` property calculation in `prepareReturnValue()`.

---

Internal details:

`standalone()` just returns a return value from `prepareReturnValue()`. This value includes a correct `errored` property.

https://github.com/stylelint/stylelint/blob/9ddba420d9f146331a55d0f62d3154331ce4291c/lib/standalone.js#L268-L272

https://github.com/stylelint/stylelint/blob/9ddba420d9f146331a55d0f62d3154331ce4291c/lib/prepareReturnValue.js#L40-L48

Before this pull request's change, `returnValue.errored` could differ from `returnValue.results[*].errored`.

https://github.com/stylelint/stylelint/blob/9ddba420d9f146331a55d0f62d3154331ce4291c/lib/prepareReturnValue.js#L59

### DIFF
--- a/.changeset/hot-planets-rest.md
+++ b/.changeset/hot-planets-rest.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: inconsistent `errored` properties in `stylelint.lint()` return value
--- a/lib/__tests__/descriptionlessDisables.test.mjs
+++ b/lib/__tests__/descriptionlessDisables.test.mjs
@@ -27,19 +27,14 @@ it('descriptionlessDisables', async () => {
 		/* stylelint-disable-next-line block-no-empty */
 		a {}`;
 
-	const linted = await standalone({
+	const { results } = await standalone({
 		config,
 		code: css,
 		reportDescriptionlessDisables: true,
 	});
-	const results = linted.results;
 
 	expect(results).toHaveLength(1);
-	const warnings = results[0].warnings.filter(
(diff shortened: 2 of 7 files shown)


jeddy3 (approved):
Thank you for discovering the cause.

LGTM!

PR #6992, opened 2023-06-27
Refactor to remove unneeded `findListStyleType()` utility

ybiquitous:

None.

No grep match:

```console
$ git grep 'findListStyleType' || echo '(no match)'
(no match)
```


jeddy3 (approved):

jeddy3:
It's excellent to see so much clean up as we move into the ESM era!

PR #6993, opened 2023-06-27
Refactor to inline `flattenArray()` utility

ybiquitous:

None.

`Array.prototype.flat()` is now commonly available, so we no longer need the custom utility.

Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

No grep match:

```console
$ git grep 'flattenArray' || echo '(no match)'
(no match)
```


lib/utils/flattenArray.js
```diff
@@ -1,16 +0,0 @@
-'use strict';
-
-/**
- * Convert the specified value to an array. If an array is specified, the array is returned as-is.
```
Mouvedia, line 4: Looking at the description and the implementation, it seems that, despite the name, it's different from `Array.prototype.flat`.

PR #6996, opened 2023-06-28
Refactor `no-duplicate-at-import-rules` to remove `postcss-media-query-parser`

romainmenke:

I found that this rule doesn't actually require parsing media queries.
It only used the media query parser to split on comma's.

This is something that can be easily done with the `postcss-value-parser` AST nodes.

### DIFF
--- a/lib/rules/no-duplicate-at-import-rules/index.js
+++ b/lib/rules/no-duplicate-at-import-rules/index.js
@@ -41,7 +41,7 @@ const rule = (primary) => {
 					? firstParam.nodes[0].value
 					: firstParam.value;
 
-			const media = listMediaQueryConditions(restParams);
+			const media = listImportConditions(restParams);
 
 			let importedUris = imports[uri];
 			const isDuplicate = media.length
@@ -71,12 +71,12 @@ const rule = (primary) => {
 };
 
 /**
- * List the media query conditions found in the prelude of an `@import` rule
+ * List the import conditions found in the prelude of an `@import` rule
  *
  * @param {import('postcss-value-parser').Node[]} params
  * @returns {Array<string>}
  */
-function listMediaQueryConditions(params) {
+function listImportConditions(params) {
 	if (!params.length) return [];
 
 	/** @type {Array<import('postcss-value-parser').Node>} */
--- a/lib/rules/no-duplicate-at-import-rules/index.js
+++ b/lib/rules/no-duplicate-at-import-rules/index.js
@@ -79,9 +79,6 @@ const rule = (primary) => {
 function listImportConditions(params) {
 	if (!params.length) return [];
 
-	/** @type {Array<import('postcss-value-parser').Node>} */
-	let layerOrSupportConditions = [];
-
 	/** @type {Array<Array<import('postcss-value-parser').Node>>} */
 	let media = [];
 	/** @type {Array<import('postcss-value-parser').Node>} */
@@ -91,20 +88,6 @@ function listImportConditions(params) {
 		const param = params[i];
 
 		if (!param) break;
-
-		if (!media.length) {
-			// @import url(...) layer(base) supports(display: flex)
-			if (param.type === 'function' && (param.value === 'supports' || param.value === 'layer')) {
-				layerOrSupportConditions.push(param);
-				continue;
-			}
-
-			// @import url(...) layer
-			if (param.type === 'word' && param.value === 'layer') {
-				layerOrSupportConditions.push(param);
-				continue;
-			}
-		}
 
 		// remove whitespace to get a consistent key
 		if (param.type === 'space') {
@@ -129,15 +112,7 @@ function listImportConditions(params) {
 		media.push(lastMediaQuery);
 	}
 
-	const layerOrSupportConditionsString = layerOrSupportConditions.length
-		? `${layerOrSupportConditions.map((c) => valueParser.stringify(c)).join(' ')} `
-		: '';
-
-	return media
-		.map((m) => {
-			return layerOrSupportConditionsString + valueParser.stringify(m).trim();
-		})
-		.filter((n) => n.length);
+	return media.map((m) => valueParser.stringify(m).trim()).filter((n) => n.length);
 }
 
 rule.ruleName = ruleName;
--- a/lib/rules/no-duplicate-at-import-rules/index.js
+++ b/lib/rules/no-duplicate-at-import-rules/index.js
@@ -79,6 +79,9 @@ const rule = (primary) => {
 function listMediaQueryConditions(params) {
 	if (!params.length) return [];
 
+	/** @type {Array<import('postcss-value-parser').Node>} */
+	let layerOrSupportConditions = [];
+
 	/** @type {Array<Array<import('postcss-value-parser').Node>>} */
 	let media = [];
 	/** @type {Array<import('postcss-value-parser').Node>} */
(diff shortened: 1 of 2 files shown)


lib/rules/no-duplicate-at-import-rules/index.js
```diff
@@ -73,6 +70,51 @@ const rule = (primary) => {
 	};
 };
 
+/**
+ * List the import conditions found in the prelude of an `@import` rule
+ *
+ * @param {import('postcss-value-parser').Node[]} params
```
Mouvedia, line 76: minor

since you are using `import('postcss-value-parser').Node` 3 times, you could create a custom type before
l76 l82 l84

lib/rules/no-duplicate-at-import-rules/index.js
```diff
@@ -73,6 +70,51 @@ const rule = (primary) => {
...
+		}
+
+		// remove comments to get a consistent key
+		if (param.type === 'comment') {
+			continue;
+		}
+
+		if (param.type === 'div' && param.value === ',') {
+			media.push(lastMediaQuery);
+			lastMediaQuery = [];
```
Mouvedia, line 104: suggestion

since all `if` branches `continue` maybe it would be clearer to just put an additional `if ((i + 1) === params.length)` before l108 so that you can get rid of this reset

pretty minor, you could put `params.length` in a const

lib/rules/no-duplicate-at-import-rules/index.js
```diff
@@ -73,6 +70,51 @@ const rule = (primary) => {
...
+ */
+function listImportConditions(params) {
+	if (!params.length) return [];
+
+	/** @type {Array<Array<import('postcss-value-parser').Node>>} */
+	let media = [];
+	/** @type {Array<import('postcss-value-parser').Node>} */
+	let lastMediaQuery = [];
+
+	for (let i = 0; i < params.length; i++) {
```
Mouvedia, line 87: nit

you could put `params.length` in a const

romainmenke: `.length` is only accessed twice and there is no performance penalty when accessing an array's length.

In this case I think it is fine to leave as is.
But happy to change if there is consensus that assigning to a const is better here :)

ybiquitous: `for...of` seems available to me 👀 

```js
for (const param of params) {
```

lib/rules/no-duplicate-at-import-rules/index.js
```diff
@@ -73,6 +70,52 @@ const rule = (primary) => {
 	};
 };
 
+/**
+ * List the import conditions found in the prelude of an `@import` rule
+ *
+ * @param {Node[]} params
+ * @typedef {import('postcss-value-parser').Node} Node
```
ybiquitous, line 77: `@template` can reduce `Node` repetition like this. But it's OK to leave it as-is since it's trivial. 👌🏼 

```suggestion
 * @template {import('postcss-value-parser').Node} T
 * @param {T[]} params
```

lib/rules/no-duplicate-at-import-rules/index.js
```diff
@@ -73,15 +73,16 @@ const rule = (primary) => {
 /**
  * List the import conditions found in the prelude of an `@import` rule
  *
- * @param {import('postcss-value-parser').Node[]} params
+ * @param {Node[]} params
+ * @typedef {import('postcss-value-parser').Node} Node
```
Mouvedia, line 77: I didn't know typedefs were hoisted.
I learned something today, thanks.

lib/rules/no-duplicate-at-import-rules/index.js
```diff
@@ -73,6 +70,52 @@ const rule = (primary) => {
...
+ * List the import conditions found in the prelude of an `@import` rule
+ *
+ * @param {Node[]} params
+ * @typedef {import('postcss-value-parser').Node} Node
+ * @returns {Array<string>}
+ */
+function listImportConditions(params) {
+	if (!params.length) return [];
+
+	/** @type {Array<Array<Node>>} */
```
ybiquitous, line 83: [suggest] Maybe we can simplify the `media` type:

```suggestion
	/** @type {Array<Node>} */
```

With `.push(...array)` pattern:

```js
if (param.type === 'div' && param.value === ',') {
	media.push(...lastMediaQuery);
	lastMediaQuery = [];
	continue;
}
```

```js
media.push(...lastMediaQuery);
```

lib/rules/no-duplicate-at-import-rules/index.js
```diff
@@ -73,6 +70,52 @@ const rule = (primary) => {
...
+		}
+
+		lastMediaQuery.push(param);
+	}
+
+	if (lastMediaQuery.length) {
+		media.push(lastMediaQuery);
+	}
+
+	return media.map((m) => valueParser.stringify(m).trim()).filter((n) => n.length);
```
ybiquitous, line 116: [suggest] Since `space` nodes are filtered out of `media` nodes, we may simplify the code: 🤔 

```suggestion
	return media.map((m) => valueParser.stringify(m));
```

Although keeping it as-is has no problem.

lib/rules/no-duplicate-at-import-rules/index.js
romainmenke: I've changed this to `for...of`!

Thank you for the suggestion

lib/rules/no-duplicate-at-import-rules/index.js
romainmenke: good call!

lib/rules/no-duplicate-at-import-rules/index.js
romainmenke: That would not be correct.

This function should split media query lists on comma's.
If we use the spread operator `const media = []` would be an array of parts of media queries, not an array of media queries.

`@media (color: 1) and (hover) {}`

`['(color:1)', 'and', '(hover)']`

ybiquitous: I see. Let's leave it. 👍🏼

ybiquitous (approved):
Thanks, LGTM 👍🏼

lib/rules/no-duplicate-at-import-rules/index.js
romainmenke: I went over this function one more time checking this and found that serializing a bit earlier is also a way to simplify the types.

In doing so I also found a bug in the removal of whitespace.

ybiquitous: Thanks! Easier to read now 👍🏼

lib/rules/no-duplicate-at-import-rules/index.js
```diff
@@ -73,6 +70,44 @@ const rule = (primary) => {
...
+ * List the import conditions found in the prelude of an `@import` rule
+ *
+ * @param {Node[]} params
+ * @typedef {import('postcss-value-parser').Node} Node
+ * @returns {Array<string>}
+ */
+function listImportConditions(params) {
+	if (!params.length) return [];
+
+	/** @type {Array<String>} */
```
ybiquitous, line 83: Prefer `string` over `String`?

```suggestion
	/** @type {Array<string>} */
```

ybiquitous: Just trivial, please ignore it 😅

PR #6998, opened 2023-06-29
Add support JS objects for extends config option

### LINKED ISSUE #6952: Add support JS objects for `extends` config option

First, assume that there is a pure ESM package for a sharable config named `foo-config`:

`foo-config/package.json`:

```json
{
  "name": "foo-config",
  "version": "1.0.0",
  "type": "module",
  "main": "index.js"
}
```

`foo-config/index.js`:

```js
export default {
	rules: {
		'color-no-invalid-hex': true,
	},
};
```

Next, consider a case using `foo-config` via npm:

`package.json`:

```json
{
  "dependencies": {
    "foo-config": "^1.0.0",
    "stylelint": "^15.8.0"
  }
}
```

`.stylelintrc.mjs`:

```js
import foo from 'foo-config';

export default {
	// Not working => Error [ERR_REQUIRE_ESM]: require() of ES Module ...
	extends: 'foo-config',

	// Still not working => TypeError: Expected `moduleId` to be of type `string`, got `object`
	extends: foo,
};
```

Running `stylelint` like this raises the errors above:

```shell
echo 'a { color: #f4 }' | npx stylelint --stdin
```

This means that Stylelint cannot use a pure ESM sharable config with the `extends` option.

Related: #5291

I expect Stylelint to work even when using JS objects with `extends` instead of strings (e.g., package name or path). For example:

```console
$ echo 'a { color: #f4 }' | npx stylelint --stdin

<input css qXAh0i>
 1:12  ✖  Unexpected invalid hex color "#f4"  color-no-invalid-hex

```

I find it works by applying the following patch:

```diff
--- a/lib/augmentConfig.js
+++ b/lib/augmentConfig.js
@@ -178,7 +178,13 @@ async function extendConfig(stylelint, config, configDir, rootConfigDir, filePat
 	let resultConfig = originalWithoutExtends;
 
 	for (const extendLookup of normalizedExtends) {
-		const extendResult = await loadExtendedConfig(stylelint, configDir, extendLookup);
+		let extendResult;
+
+		if (typeof extendLookup === 'string') {
+			extendResult = await loadExtendedConfig(stylelint, configDir, extendLookup);
+		} else if (typeof extendLookup === 'object' && extendLookup !== null) {
+			extendResult = { config: extendLookup };
+		}
 
 		if (extendResult) {
 			let extendResultConfig = extendResult.config;
```

This patch simply uses JS objects if they are passed.

What do you think about this idea? Is there some pitfall?

---

For your information, I refer to the current spec of `extends`:

https://github.com/stylelint/stylelint/blob/16754b1dab7590238f9b842ebd25678f8204ae9d/docs/user-guide/configure.md?plain=1#L215-L219

fpetrakov:

Closes #6952.

No, it's self-explanatory.

### DIFF
--- a/lib/__tests__/fixtures/config-extending-with-object.mjs
+++ b/lib/__tests__/fixtures/config-extending-with-object.mjs
@@ -1,0 +1,5 @@
+import configBlockNoEmpty from './config-block-no-empty.mjs';
+
+export default {
+	extends: configBlockNoEmpty,
+};
--- a/lib/augmentConfig.js
+++ b/lib/augmentConfig.js
@@ -178,7 +178,13 @@ async function extendConfig(stylelint, config, configDir, rootConfigDir, filePat
 	let resultConfig = originalWithoutExtends;
 
 	for (const extendLookup of normalizedExtends) {
-		const extendResult = await loadExtendedConfig(stylelint, configDir, extendLookup);
+		let extendResult;
+
+		if (typeof extendLookup === 'string') {
+			extendResult = await loadExtendedConfig(stylelint, configDir, extendLookup);
+		} else if (typeof extendLookup === 'object' && extendLookup !== null) {
+			extendResult = { config: extendLookup };
+		}
 
 		if (extendResult) {
 			let extendResultConfig = extendResult.config;
--- a/.changeset/gorgeous-cobras-beam.md
+++ b/.changeset/gorgeous-cobras-beam.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: support for JS objects with `extends` config option
--- a/lib/__tests__/extends.test.mjs
+++ b/lib/__tests__/extends.test.mjs
@@ -1,5 +1,6 @@
 import { fileURLToPath } from 'node:url';
 
+import configExtendingWithObject from './fixtures/config-extending-with-object.mjs';
 import readJSONFile from '../testUtils/readJSONFile.mjs';
 import safeChdir from '../testUtils/safeChdir.mjs';
 import standalone from '../standalone.js';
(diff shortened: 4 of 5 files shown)


mattxwang (changes requested):
Thanks for submitting a PR @fpetrakov! Do you mind adding a test case (or a few) that validate the behaviour you've added?

fpetrakov:
@ybiquitous just added the test case. Would you mind reviewing it?

ybiquitous (approved):
LGTM, thank you for the quick fix 👍🏼

lib/__tests__/fixtures/config-extending-with-object.mjs
```diff
@@ -0,0 +1,5 @@
+import configBlockNoEmpty from './config-block-no-empty.mjs';
+
+export default {
+	extends: configBlockNoEmpty,
```
Mouvedia, line 4: nit

You could import `lib/__tests__/fixtures/config-block-no-empty.json` instead.
i.e. you don't need to create `lib/__tests__/fixtures/config-block-no-empty.mjs`

mattxwang (approved):
LGTM!

lib/augmentConfig.js
```diff
@@ -178,7 +178,13 @@ async function extendConfig(stylelint, config, configDir, rootConfigDir, filePat
 	let resultConfig = originalWithoutExtends;
 
 	for (const extendLookup of normalizedExtends) {
-		const extendResult = await loadExtendedConfig(stylelint, configDir, extendLookup);
+		let extendResult;
+
+		if (typeof extendLookup === 'string') {
```
Mouvedia, line 183: minor

```js
if (typeof extendLookup === 'object' && extendLookup !== null) {
// …
} else {
// …
}
```
should be enough.
i.e. if we really wanted to check the type we would use something like [is-string](https://www.npmjs.com/package/is-string) or [isstring](https://www.npmjs.com/package/@string/isstring)

lib/__tests__/fixtures/config-extending-with-object.mjs
ybiquitous: Importing JSON with pure ESM currently needs Import assertions, so I think it's fine as it is.

lib/augmentConfig.js
ybiquitous: Strictly checking the type here is important to avoid unexpected failures. Also, it's too much to install another package just for `typeof`.

Mouvedia: related: #5341

PR #6999, opened 2023-06-29
Refactor `media-feature-name-value-allowed-list` to replace  `postcss-media-query-parser` with `@csstools/media-query-list-parser`

romainmenke:

There is some duplication because of `function startAndEndIndex(node) {` between multiple rules.

I don't want to create a utility for this in Stylelint, instead I propose I add this upstream : https://github.com/csstools/postcss-plugins/pull/1022

For now we can keep the duplicate code and refactor when the upstream utility has been released.

-----

This change also removes the `postcss-media-query-parser` as it was the last rule that used this dependency.


ybiquitous:
> See : #6848

Thanks for the pull request! Will this PR `close #6848`?

romainmenke:
> Thanks for the pull request! Will this PR close #6848?

I was thinking that it might be good to do a final review of all refactored rules.
I am sure that some similar things were handled inconsistently between different rules and now might be a good time to go over them one last time.

package-lock.json
ybiquitous, line 1: [suggest] Can you remove `@types/postcss-media-query-parser` also from `package-lock.json`?

docs/user-guide/errors.md
```diff
@@ -65,7 +65,7 @@ The CSS parser built into Stylelint (or the chosen [custom syntax](options.md#cu
 
 The construct-specific parsers are:
 
-- `postcss-media-query-parser`
+- `@csstools/media-query-list-parser`
```
ybiquitous, line 68: [ask] I think it's good to add a changelog entry for this replacement to notify people, especially plugin authors. Any thoughts?

E.g.

```md
Fixed: remove `postcss-media-query-parser` dependency
```

https://github.com/stylelint/stylelint/blob/26f9351d733effaf14b9b0aa50cee8fb6060fe96/CHANGELOG.md?plain=1#L22

lib/rules/media-feature-name-value-allowed-list/index.js
```diff
@@ -37,70 +40,98 @@ const rule = (primary) => {
...
+			parseMediaQuery(atRule).forEach((mediaQuery) => {
+				if (isMediaQueryInvalid(mediaQuery)) return;
+
+				mediaQuery.walk(
+					(entry) => {
+						if (!entry.state) return;
+
+						if (isMediaFeature(entry.node)) {
+							entry.state.mediaFeatureName = entry.node.getName();
+							entry.state.unprefixedMediaFeatureName = vendor.unprefixed(entry.node.getName());
```
Mouvedia, line 52: off topic

I didn't see a test for vendor prefixes.

lib/rules/media-feature-name-value-allowed-list/index.js
```diff
@@ -37,70 +40,98 @@ const rule = (primary) => {
...
-				const mediaFeatureNode = node.nodes.find((n) => n.type === 'media-feature');
-
-				if (mediaFeatureNode == null) throw new Error('A `media-feature` node must be present');
-
-				let mediaFeatureName;
-				let values;
+			parseMediaQuery(atRule).forEach((mediaQuery) => {
+				if (isMediaQueryInvalid(mediaQuery)) return;
+
+				mediaQuery.walk(
```
ybiquitous, line 46: [suggest] This is just my personal preference. You can leave it as-is.

I feel a bit hard to read the code since a deeper indentation and repeated `entry.{var}` variables are present. How about the following refactoring like this?

```js
const initialState = {
	mediaFeatureName: '',
	unprefixedMediaFeatureName: '',
};

mediaQuery.walk(({ node, state }) => {
  // ...
}, initialState);
```

lib/rules/media-feature-name-value-allowed-list/index.js
```diff
@@ -37,70 +40,98 @@ const rule = (primary) => {
...
+
+							return;
+						}
+
+						if (!isMediaFeatureValue(entry.node)) return;
+
+						if (!entry.state.mediaFeatureName || !entry.state.unprefixedMediaFeatureName) return;
+
+						const mediaFeatureName = entry.state.mediaFeatureName;
+						const unprefixedMediaFeatureName = entry.state.unprefixedMediaFeatureName;
```
ybiquitous, line 62: [suggest] It's possible to shorten the code a bit. How about this refactoring?

```suggestion
						const { mediaFeatureName, unprefixedMediaFeatureName } = entry.state;

						if (!mediaFeatureName || !unprefixedMediaFeatureName) return;
```

ybiquitous (changes requested):
[suggest] It seems we can remove also `isRangeContextMediaFeature.js`:

```console
$ git grep -l 'isRangeContextMediaFeature'
lib/utils/__tests__/isRangeContextMediaFeature.test.js
lib/utils/isRangeContextMediaFeature.js
```

lib/rules/media-feature-name-value-allowed-list/index.js
```diff
@@ -37,70 +40,98 @@ const rule = (primary) => {
...
+						if (!entry.state) return;
+
+						if (isMediaFeature(entry.node)) {
+							entry.state.mediaFeatureName = entry.node.getName();
+							entry.state.unprefixedMediaFeatureName = vendor.unprefixed(entry.node.getName());
+
+							return;
+						}
+
+						if (!isMediaFeatureValue(entry.node)) return;
```
Mouvedia, line 57: question

Since the previous behaviour where it would `throw` if no media-feature was found wasn't tested, should we add a test or comment for your change?
Maybe it's too obvious and not necessary?

romainmenke: I had interpreted the old behavior as a fixup for the typing of the old media query parser AST.

- `if (mediaFeatureNode == null) throw new Error('A `media-feature` node must be present');`
- `if (valueNode == null) throw new Error('A `value` node must be present');`

It doesn't make sense for either to ever be `null` in something that parses without error as a media query.

A boolean feature doesn't have a value, but there was check for that a few lines earlier.

-----

The new parser does't have this typing issue.

lib/rules/media-feature-name-value-allowed-list/index.js
romainmenke: Thank you!

I was looking for ways to reduce the indentation a bit, but I didn't consider that the state object at end was causing it :)

Much better with your suggestion 🙇

package-lock.json
romainmenke: Good catch, I forgot to run a second `npm install`.

docs/user-guide/errors.md
romainmenke: Makes sense! I've added this.

.changeset/fifty-hotels-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: remove `postcss-media-query-parser` dependency
```
Mouvedia, line 5: minor

You can use `Removed:`.
cf https://github.com/stylelint/stylelint/blob/main/CHANGELOG.md#1500

romainmenke, line 5:
```suggestion
Removed: `postcss-media-query-parser` dependency
```

romainmenke: thank you 🙇

ybiquitous (approved):
Thanks a lot! Our codebase is now much improved! 👍🏼

jeddy3 (approved):
LGTM, thank you!

romainmenke:
Thank you for all the thorough reviews 🙇

silverwind:
Seems the removal of `lib/utils/isRangeContextMediaFeature.js` broken `stylelint-stylistic` which directly imports a number of files from this repo.

https://github.com/elirasza/stylelint-stylistic/blob/dce7ea440d44a8e6aa28251eca4e3fbd98a376ba/lib/rules/media-feature-name-case/index.js#L4

ybiquitous:
@silverwind Thanks for the feedback. I've left a comment on <https://github.com/elirasza/stylelint-stylistic/issues/12#issuecomment-1620842580>.

TyIsI:
Linking for documentation purposes #7057

Btw, on a related note, does the stylelint team note breaking changes in the changelog?
Also, this issue/PR is not reflected in the CHANGELOG.md.

Mouvedia:
> Also, this issue/PR is not reflected in the CHANGELOG.md.

16110fd

jeddy3:
> Btw, on a related note, does the stylelint team note breaking changes in the changelog?

Yes, but this change was an internal refactor with no changes to our public API so it didn't have a changelog entry. Unfortunately, the stylistic plugin was requiring these internal utils. This issue [is being remedied](https://github.com/elirasza/stylelint-stylistic/pull/13) in that package.

Issue #6930, opened 2023-06-17
Prepare 16.0.0


ybiquitous:
I think it's a good time to start a discussion about the next major version.

Ref:
- Previous major version (Feb 2023): https://github.com/stylelint/stylelint/releases/tag/15.0.0
- https://github.com/stylelint/stylelint/issues/6409
- [future-major milestone](https://github.com/stylelint/stylelint/milestone/20)

Included:
*TBD*

Discussed:
- #6740
- #5291

*Please suggest other topics with a comment if you have any ideas.*

jeddy3:
I'm catching up on my notifications. Seeing all the releases and hard work done over the last few months is fantastic! 

> I think it's a good time to start a discussion about the next major version.

I agree, and with including https://github.com/stylelint/stylelint/issues/6740 & https://github.com/stylelint/stylelint/issues/5291.

Let's remove the deprecated rules in 16.0.0 now that community plugins are available. One of the original motivations for deprecating the stylistic rules was to aid our migration to ESM, as those rules are the bulk of the oldest parts of our codebase.

@ybiquitous As an aside, do you have admin rights [to recover](https://support.1password.com/recovery/) my Stylelint 1password account? Something is up with my secret key.

ybiquitous:
@jeddy3 Welcome back! 😄 

I agree with removing the deprecated rules, and I just updated this issue description.

> As an aside, do you have admin rights [to recover](https://support.1password.com/recovery/) my Stylelint 1password account? Something is up with my secret key.

Oh..., I don't have any admin access to 1password and am unsure about 1password details. 😓

ybiquitous:
@jeddy3 Thanks. I can access it as an admin. 👍🏼 

By the way, Can I invite @mattxwang? He has recently performed releasing work many times (also today).
If he can access Twitter, we can leave releasing work to him completely.

jeddy3:
> By the way, can I invite @mattxwang?

Absolutely. You can invite him to Team Member's group with access to the shared vault. @mattxwang Thanks for doing the recent releases. You're doing a fantastic job!

ybiquitous:
@mattxwang I've sent an invitation to your email on your profile page. Can you check it, please?

ybiquitous:
I've added #6966 to the "Discussing" list.

mattxwang:
> @mattxwang I've sent an invitation to your email on your profile page. Can you check it, please?

Thanks - just signed up with an account with my public email address!

[...]

PR #7002, opened 2023-06-29
Prepare 16.0.0

mattxwang:
Ref #6930

~~(draft PR for breaking v16 changes)~~


ybiquitous:
Hi @mattxwang, thanks for creating this PR so we can easily track v16 changes. 👍🏼 

Just to confirm, we will need to regularly rebase the `main` branch, right?

mattxwang:
Sorry all, I've been a bit MIA as work has been taking up most of my time. Noticed that there's a minor coverage drop; is that something we should be concerned about (I could take a look at adding a missing test)

ybiquitous:
I think there is such a minor coverage drop because `.mjs` and `.cjs` files are mixed now. This will be resolved when we finish the migration to ESM. But surely we need to add test cases for CJS.

ybiquitous:
Finally, #7160 should be the last PR to merge into `v16`. After merging #7160, we will be ready to merge this PR.

Although we may need a few changes in the documentation before the 16.0.0 release, we can also do that on the `main` branch.

ybiquitous (approved):
Now, we should be ready to merge. LGTM.
I want one more approval, just in case, since this is a big PR. 🤯 

**Note to self: we must NOT "squash and merge", but just "merge"!**

types/stylelint/index.d.ts
```diff
@@ -53,574 +56,741 @@ type RuleMessageFunc = {
...
-		start: number;
-		end?: number;
-	};
+	 * Optional severity override for the problem.
+	 */
+	severity?: RuleSeverity;
+};
 
+/** @internal */
+export type ShorthandProperties =
```
Mouvedia, line 605: minor
non-blocking

This union should have been generated from a `Set` imported from a file in the reference directory.

ybiquitous: Yeah, syncing the implementation and type definition is best. I don't know how to achieve it, though. 😅

Mouvedia: It should look something like that:

```
export const bar = /** @type {const} */ (['animation', 'background', 'etc']);
…
import { bar } from 'qux.mjs';
type Baz = typeof foo[number]; // 'animation' | 'background' | 'etc'
```

ybiquitous: Let's discuss this in another place.

jeddy3 (approved):
Amazing work @ybiquitous! This was a colossal effort.

Let's merge so that we can release `next`, start testing, and patch any bugs on `main` before releasing `16.0.0`.

PR #7004, opened 2023-06-30
Fix function-no-unknown performance

jeddy3:

Ref #6869 

Before:
Mean: 104.89167158823528 ms
Deviation: 9.601897806542654 ms

After:
Mean: 71.84089116666665 ms
Deviation: 10.215987936599932 ms

### DIFF
--- a/.changeset/shy-tools-rest.md
+++ b/.changeset/shy-tools-rest.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+- Fixed: `function-no-unknown` performance
--- a/.changeset/shy-tools-rest.md
+++ b/.changeset/shy-tools-rest.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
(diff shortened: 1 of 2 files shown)


jeddy3:
> LGTM! How about adding a changelog entry like this?

Oops, I'm rusty 😆 . I forgot to do that.

.changeset/shy-tools-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+- Fixed: `function-no-unknown` performance
```
ybiquitous, line 5:
```suggestion
Fixed: `function-no-unknown` performance
```

ybiquitous (approved):
Thank you! 👍🏼

.changeset/shy-tools-rest.md
jeddy3: 😅

PR #7008, opened 2023-06-30
Improve `npm run benchmark-rule` script and doc

ybiquitous:

None.

- Add the help output (usage and examples).
- Replace deprecated rule in the doc.

```console
$ npm run benchmark-rule -s
Usage: node benchmark-rule.mjs <ruleName> <ruleOptions> [ruleContext]

Examples:
  # With a primary option
  node benchmark-rule.mjs value-keyword-case lower

  # With secondary options (JSON format)
  node benchmark-rule.mjs value-keyword-case '["lower", {"camelCaseSvgKeywords": true}]'

  # With a context
  node benchmark-rule.mjs value-keyword-case '["lower", {"camelCaseSvgKeywords": true}]' '{"fix": true}'
```

Ref: https://github.com/stylelint/stylelint/blob/15.9.0/lib/rules/value-keyword-case/README.md

### DIFF
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -287,7 +287,7 @@ Deprecating rules doesn't happen very often. When you do, you must:
 
 ## Improve the performance of a rule
 
-You can run a benchmarks on any given rule with any valid config using:
+You can run a benchmark on any given rule with any valid config using:
 
 ```shell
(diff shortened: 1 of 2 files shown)


jeddy3 (approved):
Nice. LGTM, thank you!

PR #7009, opened 2023-06-30
Fix custom-property-pattern performance

jeddy3:

Ref https://github.com/stylelint/stylelint/issues/6869

Before: 
Mean: 148.86722335714288 ms
Deviation: 14.868270718260428 ms

After:
Mean: 120.50095842857144 ms
Deviation: 18.926086943302593 ms

### DIFF
--- a/lib/rules/custom-property-pattern/index.js
+++ b/lib/rules/custom-property-pattern/index.js
@@ -20,6 +20,8 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/custom-property-pattern',
 };
 
+const VAR_FUNC_REGEX = /var\(/i;
+
 /** @type {import('stylelint').Rule} */
 const rule = (primary) => {
 	return (root, result) => {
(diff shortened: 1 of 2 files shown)


lib/rules/custom-property-pattern/index.js
```diff
@@ -40,30 +40,32 @@ const rule = (primary) => {
...
+				!isStandardSyntaxProperty(property) ||
 				regexpPattern.test(property.slice(2))
 			);
 		}
 
 		root.walkDecls((decl) => {
 			const { prop, value } = decl;
 
-			const parsedValue = valueParser(value);
+			if (/var\(/i.test(value)) {
```
ybiquitous, line 52: [suggest] Caching a regex seems to improve performance a bit. Any thoughts?

For example:
https://github.com/stylelint/stylelint/blob/518367a28c8e742516bc8882b6cdf0a843c15ff2/lib/rules/function-calc-no-unspaced-operator/index.js#L28

ybiquitous (approved):
Thanks, LGTM 👍🏼 

I've left a suggestion, but it's minor.

ybiquitous (approved):
Thank you. 👍🏼

PR #7013, opened 2023-06-30
Fix selector-id-pattern performance

jeddy3:

Ref https://github.com/stylelint/stylelint/issues/6869

Mean: 240.67508675000005 ms
Deviation: 20.382065001610304 ms

To

Mean: 85.73348046666666 ms
Deviation: 8.678100769643914 ms

Avoiding unnecessary selector parsing is a win. We seldom see id selectors in the wild.

### DIFF
--- a/lib/rules/selector-id-pattern/index.js
+++ b/lib/rules/selector-id-pattern/index.js
@@ -32,6 +32,8 @@ const rule = (primary) => {
 		const normalizedPattern = isString(primary) ? new RegExp(primary) : primary;
 
 		root.walkRules((ruleNode) => {
+			if (!ruleNode?.selector.includes('#')) return;
+
 			if (!isStandardSyntaxRule(ruleNode)) {
 				return;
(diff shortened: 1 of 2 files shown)


lib/rules/selector-id-pattern/index.js
```diff
@@ -32,6 +32,8 @@ const rule = (primary) => {
 		const normalizedPattern = isString(primary) ? new RegExp(primary) : primary;
 
 		root.walkRules((ruleNode) => {
```
ybiquitous, line 34: [ask] How about specifying a selector filter (regex) instead?

```suggestion
		root.walkRules(/#/, (ruleNode) => {
```

jeddy3: That's nicer! (Benchmark is comparable.)

ybiquitous (approved):
Great improvement with a few characters! 🎉

Issue #7014, opened 2023-06-30
Fix `hue-degree-notation` false negatives for `oklch`


romainmenke:
### What minimal example or steps are needed to reproduce the bug?

```css
a {
	color: lch(50% 50% 50);
	color: oklch(50% 50% 50);
}
```

### What minimal configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "hue-degree-notation": "angle"
  }
}
```

### How did you run Stylelint?

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYAdAdugLpANhAE7wr5gAWAFAKwAMApCvU-QJQDcWuBxpEAa3LUWzRmM5YAviAA0IAGYBLfDABySALZxEMAB5aADqoB0YAM7m54CJmUBzBCAyYUKdCCIBXVeY+kXNzcPCi8YAFpYeyIYCMwIHCQcJVt-d2RMe1UPLDcpaWtIOyV7ADFiTSSnACtzW2tYQytEQPTzHABPVXwlTBw0j3wkmHaPWVy2zu7enHCih3D2pEwoJCIoAZAhnBH+kAKpIA)

### Which Stylelint-related dependencies are you using?

```json
{
  "stylelint": "latest",
  "stylelint-config-standard": "latest"
}
```

### What did you expect to happen?

I expected an error on `oklch`

### What actually happened?

No error was reported

### Do you have a proposal to fix the bug?

Add `oklch` in the same way as `lch` already exists today.

PR #7015, opened 2023-06-30
Fix `hue-degree-notation` false negatives for `oklch`

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/7014

No, it's self-explanatory.


ybiquitous (approved):
Good catch! LGTM 👍🏼

PR #7016, opened 2023-06-30
Fixed selector|value-no-vendor-prefix performance

jeddy3:

https://github.com/stylelint/stylelint/issues/6869

selector-no-vendor-prefix

Mean: 246.52507291666666 ms
Deviation: 22.75678614753106 ms

To

Mean: 91.62227779999999 ms
Deviation: 7.8451629248369095 ms

value-no-vendor-prefix

Mean: 102.0994388 ms
Deviation: 11.78154094781758 ms

To

Mean: 91.27013606666667 ms
Deviation: 15.351019516794766 ms

### DIFF
--- a/.changeset/short-onions-play.md
+++ b/.changeset/short-onions-play.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `selector|value-no-vendor-prefix` performance
+Fixed: `{selector,value}-no-vendor-prefix` performance
--- a/.changeset/short-onions-play.md
+++ b/.changeset/short-onions-play.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector|value-no-vendor-prefix` performance
--- a/lib/reference/prefixes.js
+++ b/lib/reference/prefixes.js
@@ -1,0 +1,5 @@
+const prefixes = new Set(['-webkit-', '-moz-', '-ms-', '-o-']);
+
+module.exports = {
+	prefixes,
+};
--- a/lib/rules/selector-no-vendor-prefix/index.js
+++ b/lib/rules/selector-no-vendor-prefix/index.js
@@ -42,7 +42,7 @@ const rule = (primary, secondaryOptions, context) => {
 		}
 
 		root.walkRules((ruleNode) => {
-			if (!hasPrefix(ruleNode?.selector)) return;
+			if (!hasPrefix(ruleNode.selector)) return;
 
 			if (!isStandardSyntaxRule(ruleNode)) {
 				return;
--- a/lib/rules/selector-no-vendor-prefix/index.js
+++ b/lib/rules/selector-no-vendor-prefix/index.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const hasPrefix = require('../../utils/hasPrefix');
 const isAutoprefixable = require('../../utils/isAutoprefixable');
 const isStandardSyntaxRule = require('../../utils/isStandardSyntaxRule');
 const optionsMatches = require('../../utils/optionsMatches');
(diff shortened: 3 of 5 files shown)


.changeset/short-onions-play.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector|value-no-vendor-prefix` performance
```
ybiquitous, line 5: [suggest] How about a glob-like pattern? It's minor, though. 😅 

```suggestion
Fixed: `{selector,value}-no-vendor-prefix` performance
```

lib/rules/selector-no-vendor-prefix/index.js
```diff
@@ -41,6 +42,8 @@ const rule = (primary, secondaryOptions, context) => {
 		}
 
 		root.walkRules((ruleNode) => {
+			if (!hasPrefix(ruleNode?.selector)) return;
```
ybiquitous, line 45: [suggest] Is `ruleNode` always truthy, right?

```suggestion
			if (!hasPrefix(ruleNode.selector)) return;
```

lib/utils/hasPrefix.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const { prefixes } = require('../reference/prefixes');
+
+const HAS_PREFIX_REGEX = new RegExp(`(?:${[...prefixes.values()].join('|')})`, 'i');
```
ybiquitous, line 5: [ask] Is omitting `^` (starts with) intentional, right? Acceptable?

jeddy3: We may test broader strings, e.g. a selector list where the first selector isn't prefixed.

ybiquitous: I see. I'm a bit concerned that `value-no-vendor-prefix` is using `startsWith`. 🤔

ybiquitous: Ah, I misunderstood! `value-no-vendor-prefix` checks a value by `isAutoprefixable.propertyValue()`. Okay, no problem. 👍🏼

ybiquitous (approved):
Great! LGTM 👍🏼

lib/utils/hasPrefix.js
ybiquitous, line 5: [suggest] Sorry, I found it after approve. 😅 We can omit `.values()`:

```suggestion
const HAS_PREFIX_REGEX = new RegExp(`(?:${[...prefixes].join('|')})`, 'i');
```

ybiquitous (approved):
Thank you! 👍🏼

lib/reference/prefixes.js
```diff
@@ -0,0 +1,5 @@
+const prefixes = new Set(['-webkit-', '-moz-', '-ms-', '-o-']);
```
Mouvedia, line 1: If this is to be reused, Ill eventually add some more rare ones to the set.

ref: https://stackoverflow.com/questions/11705144/webkit-and-its-legacy-prefixes

lib/rules/value-no-vendor-prefix/index.js
```diff
@@ -59,11 +48,9 @@ const rule = (primary, secondaryOptions, context) => {
 		root.walkDecls((decl) => {
 			const { value } = decl;
 
-			if (
-				!isStandardSyntaxDeclaration(decl) ||
-				!isStandardSyntaxProperty(decl.prop) ||
-				!value.startsWith('-')
-			) {
+			if (!hasPrefix(value)) return;
```
Mouvedia, line 51: A more agnostic way would be to use
https://github.com/stylelint/stylelint/blob/3065425dc94be536dcf4369dcc1ed47735259e07/lib/utils/vendor.js#L22-L30
instead.

PR #7018, opened 2023-06-30
Fix `unit-disallowed-list` false negatives with percentages

romainmenke:

See https://github.com/stylelint/stylelint/issues/6848

I noticed inconsistent handling of values in at rule preludes vs. declarations.

### DIFF
--- a/lib/rules/unit-disallowed-list/index.js
+++ b/lib/rules/unit-disallowed-list/index.js
@@ -119,7 +119,7 @@ const rule = (primary, secondaryOptions) => {
 
 			if (!hasDimension(params)) return;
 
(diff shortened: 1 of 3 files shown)


lib/rules/unit-disallowed-list/index.js
```diff
@@ -194,10 +194,14 @@ const rule = (primary, secondaryOptions) => {
 };
 
 /**
+ * In the CSS syntax percentages are a different token type than dimensions.
+ * For CSS authors however this distinction doesn't make sense, so we convert
+ * percentage tokens to dimension tokens with a unit of "%".
+ *
  * @param {string} value
```
Mouvedia, line 201: nit

You could call it `css` so that it doesn't have to be renamed.

ybiquitous (approved):
Good catch! LGTM 👍🏼

Issue #6866, opened 2023-05-28
Document approach to using private utils


Max10240:
### What is the problem you're trying to solve?

I'm writting a custom stylelint rule in TS,  as mentioned in the document:
[doc link](https://stylelint.io/developer-guide/rules#utility-functions)
> Stylelint has [utility functions](https://github.com/stylelint/stylelint/tree/main/lib/utils) that are used in existing rules and might prove useful to you, as well. Please look through those so that you know what's available. (And if you have a new function that you think might prove generally helpful, let's add it to the list!).

I want to use some utils functions to help me implement the rule, like:
`example-rule.ts`
```typescript
import postcss from 'postcss';
import stylelint, { utils } from 'stylelint';

import eachDeclarationBlock from 'stylelint/lib/utils/eachDeclarationBlock';
import isKeyframeSelector from 'stylelint/lib/utils/isKeyframeSelector';
import isStandardSyntaxRule from 'stylelint/lib/utils/isStandardSyntaxRule';
import isStandardSyntaxSelector from 'stylelint/lib/utils/isStandardSyntaxSelector';
import { isDeclaration } from 'stylelint/lib/utils/typeGuards';
import { isBoolean, isRegExp, isString } from 'stylelint/lib/utils/validateTypes';

const rule: stylelint.Rule = (primary, secondaryOptions) => {
  return (root, result) => {
    // use utils functions bellow
    eachDeclarationBlock(root, eachDecl => {
      eachDecl(decl => {
        isDeclaration(decl);
      });
    });

    isDeclaration(root);
    isKeyframeSelector('');
    isStandardSyntaxSelector('');
    isStandardSyntaxRule(null! as postcss.Rule);
    isBoolean(null! as unknown);
    isRegExp(null! as unknown);
    isString(null! as unknown);

    utils.report(null! as stylelint.Problem);
  }
};

rule.ruleName = '';
rule.messages = null!;

```

But TS complained that  it could not find the declaration file:
```text
example-rule.ts:4:34 - error TS7016: Could not find a declaration file for module 'stylelint/lib/utils/eachDeclarationBlock'. 'stylelint-plugin-example/node_modules/stylelint/lib/utils/eachDeclarationBlock.js' implicitly has an 'any' type.
  If the 'stylelint' package actually exposes this module, try adding a new declaration (.d.ts) file containing `declare module 'stylelint/lib/utils/eachDeclarationBlock';`

[...]

PR #7021, opened 2023-06-30
Refactor to remove redundant calls to `isStandardSyntaxMediaFeatureName` utility

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/6848

These utilities aren't needed anymore.  
`@csstools/media-query-list-parser` only parses standard CSS as media feature AST nodes.

------

I think #6848 can be closed as completed after this change.


Mouvedia:
Can we wait for elirasza/stylelint-stylistic to be refactored first?

https://github.com/elirasza/stylelint-stylistic/blob/dce7ea440d44a8e6aa28251eca4e3fbd98a376ba/lib/rules/media-feature-name-case/index.js#L5

romainmenke:
I didn't know the functions from `utils` were part of the public API of Stylint.
I assumed they were not given this comment : https://github.com/stylelint/stylelint/blob/main/docs/developer-guide/plugins.md#external-helper-modules

---

I will restore these files.

Mouvedia:
@romainmenke it's complicated: #6866

ybiquitous:
To be honest, I'm not in favor of leaving unused and internal modules (see also <https://github.com/stylelint/stylelint/issues/6866#issuecomment-1567565404>). As a maintainer, I feel it isn't easy to care about internal modules used by plugins. 😓 

However, only this time, we've removed many utilities toward v16 (see #7002), so we can delay removing unused modules. I think it's acceptable to remove `isStandardSyntaxMediaFeatureName` on the `v16` branch.

Could you open a new PR to remove it for `v16`?

ybiquitous (approved):
Anyway, the series of refactoring work for #6848 is great! 👏🏼 

Thank you so much for your effort! @romainmenke

lib/rules/media-feature-name-allowed-list/__tests__/index.js
```diff
@@ -39,6 +39,10 @@ testRule({
 		{
 			code: '@media only screen and(min-width: 640px) { }',
 		},
+		{
+			code: '@media ($WIDTH: 50em) { }',
+			description: 'ignore non-standard syntax media feature names',
+		},
```
jeddy3, line 45:
```suggestion
```

This is covered by line 208.

(For future reference, we typically put non-standard cases into their own `testRule` with the appropriate non-standard syntax to separate standard from non-standard.)

lib/rules/media-feature-name-case/__tests__/index.js
```diff
@@ -59,6 +59,13 @@ testRule({
 			code: '@media (--VIEWPORT-MEDIUM) { }',
 			description: 'ignore css variables',
 		},
+		{
+			code: '@media ($WIDTH: 700px) { }',
+			description: 'ignore non-standard syntax media feature names',
+		},
```
jeddy3, line 65:
```suggestion
```

Covered by line 312.

jeddy3 (changes requested):
Thank you.

Two minor requests, otherwise LGTM.

> Anyway, the series of refactoring work for https://github.com/stylelint/stylelint/issues/6848 is great! 👏🏼

Totally agree!

lib/rules/media-feature-name-case/__tests__/index.js
```diff
@@ -59,9 +59,6 @@ testRule({
 			code: '@media (--VIEWPORT-MEDIUM) { }',
 			description: 'ignore css variables',
 		},
-		{
-			code: '@media; @media{}',
-		},
```
romainmenke: I was checking if this would hit a specific code bath, but it didn't.
I forgot to remove it again.

jeddy3 (approved):
LGTM, thank you!

romainmenke:
Thank you again for the reviews on all this 🙇 
Awesome to wrap this up!

PR #7024, opened 2023-07-02
Refactor `lib/rules/__tests__/*` test files to migrate to ESM

ybiquitous:
> Which issue, if any, is this issue related to?

Ref #5291

> Is there anything in the PR that needs further explanation?

This change is performed by the following script (though it requires a bit of manual).

```shell
node to-esm.mjs lib/rules/*/__tests__/index.js
```

```js
// to-esm.mjs
import { argv, stdout } from 'node:process';
import { readFile, unlink, writeFile } from 'node:fs/promises';
import { execSync } from 'node:child_process';

async function processFile(file) {
	const ruleFile = file.replace('/__tests__/index.js', '/index.js');

	if ((await readFile(ruleFile, 'utf8')).includes('\tdeprecated: true,')) {
		stdout.write(`Skip deprecated ${file}\n`);

		return;
	}

	let lines = (await readFile(file, 'utf8')).split('\n');

	lines = lines.filter((line) => !line.includes('use strict'));

	lines.forEach((line, i) => {
		line = line.replace(
			"const stripIndent = require('common-tags').stripIndent",
			"const { stripIndent } = require('common-tags')",
		);

		line = line.replace(
			"const { messages, ruleName } = require('..');",
			"import rule from '../index.js'; const { messages, ruleName } = rule;",
		);

		line = line.replace(/const (.+) = require\('(\..+)'\)/, 'import $1 from "$2.js"');
		line = line.replace(/const (.+) = require\('(.+)'\)/, 'import $1 from "$2"');

		lines[i] = line;
	});

	const newFile = file.replace(/\.js$/, '.mjs');
	const newContent = `${lines.join('\n')}\n`;

	await writeFile(newFile, newContent);

	stdout.write(`=> ${newFile}\n`);

	await unlink(file);
}

const files = argv.slice(2);

Promise.all(files.map((f) => processFile(f))).then(() => {
	execSync(`npx prettier --loglevel silent --write lib/rules`, { encoding: 'utf8' });
});
```

Note that this change does NOT include deprecated rules which were removed on the `v16` branch.

PR #7026, opened 2023-07-02
Fix `no-descending-specificity` performance

romainmenke:

This rule warned multiple times for the same selector.  
It would emit one warning for each selector with lower specificity that preceded the current selector.

I don't think this is very useful as it just causes extra noise.

With this change this rule will only emit a single warning for each selector.  
It will always reference the earliest other selector with lower specificity.

All these extra warnings came at a cost.  
fewer warnings -> faster execution.

I've also added extra tests to cover this scenario.

---

before :

```
Warnings: 407860
Mean: 1452.8959879999998 ms
Deviation: 85.89908815057446 ms
```

after :

```
Warnings: 7117
Mean: 367.08543172727275 ms
Deviation: 14.07386884681592 ms
```

---

The excessive number of warnings is the result of the duplicate CSS in the test source.

Ideally we would have a large, but well formed and realistic benchmark source.


lib/rules/no-descending-specificity/index.js
```diff
@@ -141,11 +142,12 @@ const rule = (primary, secondaryOptions) => {
 function lastCompoundSelectorWithoutPseudoClasses(selectorNode) {
 	const firstChild = selectorNode.nodes[0];
 
-	assert(firstChild);
+	if (!firstChild) return undefined;
```
romainmenke, line 145: `assert` seemed to be slower than just a falsy check.

It was a minor improvement and I think it would also be ok to keep using `assert` if it makes the code more readable/maintainable.

jeddy3 (approved):
Nice, this was the slowest rule I encountered while taking a pass at https://github.com/stylelint/stylelint/issues/6869. So, it's great to see this improvement.

(I saw your performance-related pull request upstream in PostCSS too. Excellent stuff.)

lib/rules/no-descending-specificity/index.js
jeddy3: @ybiquitous is best positioned to comment on this one.

lib/rules/no-descending-specificity/index.js
```diff
@@ -126,6 +125,8 @@ const rule = (primary, secondaryOptions) => {
 						messageArgs: [selector, priorEntry.selector],
 						word: selector,
 					});
+
+					break;
```
Mouvedia, line 129: I am surprised this change didn't break any existing test.

lib/rules/no-descending-specificity/index.js
ybiquitous: Agree with replacing `assert` 👍🏼

ybiquitous (approved):
Great improvement! 👏🏼 

`Rule#selectors` property actually has some calculation. Honestly, I feel it's hard to notice it... 😓 
https://github.com/postcss/postcss/blob/8.4.24/lib/rule.js#L13-L15

ybiquitous:
Sorry for the conflict in the test file due to #7024. Could you resolve the conflict, please? 🙏🏼

ybiquitous (approved):

PR #7027, opened 2023-07-02
Fix `selector-type-no-unknown` performance

romainmenke:

No, it's self-explanatory.


lib/rules/selector-type-no-unknown/index.js
```diff
@@ -47,13 +49,15 @@ const rule = (primary, secondaryOptions) => {
 		}
 
 		root.walkRules((ruleNode) => {
-			const selector = ruleNode.selector;
-			const selectors = ruleNode.selectors;
-
 			if (!isStandardSyntaxRule(ruleNode)) {
```
Mouvedia, line 52: minor

You could put this condition after l56 if we are sure that the type of selector is `string`.

ybiquitous: [suggest] Agree with delaying `isStandardSyntaxRule`.

Also, we can delay `ruleNode.selectors` (computed property).

I suggest the following. Any thoughts?

```js
	const { selector } = ruleNode;

	if (!STARTS_A_TAG_NAME_REGEX.test(selector)) return;

	if (!isStandardSyntaxRule(ruleNode)) return;

	const { selectors } = ruleNode;
```

ybiquitous (approved):
LGTM. Thanks 👍🏼

PR #7037, opened 2023-07-03
Build CJS files via Rollup

ybiquitous:

Ref #5291

Steps for this PR:

1. Run `npm i -D rollup`
2. Create `rollup.config.mjs`
3. Rewrite `.js` to `.mjs` in ESM
    - Note that this PR includes just a few `.mjs`. Following-up PRs will gradually migrate `.js` files to ESM.

Discussion points:
- [x] Should we commit `.cjs` files? => Yes.

See https://rollupjs.org/

---

Note: This PR is still a draft, but I welcome any feedback. 👍🏼


lib/writeOutputFile.mjs
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * @param {string} content
+ * @param {string} filePath
+ * @returns {Promise<void>}
+ */
+export default async function writeOutputFile(content, filePath) {
+	await mkdir(path.dirname(filePath), { recursive: true });
+
+	await writeFileAtomic(path.normalize(filePath), stripAnsi(content));
+}
```
ybiquitous, line 15: CJS output:

```js
'use strict';

const promises = require('node:fs/promises');
const path = require('node:path');
const stripAnsi = require('strip-ansi');
const writeFileAtomic = require('write-file-atomic');

/**
 * @param {string} content
 * @param {string} filePath
 * @returns {Promise<void>}
 */
async function writeOutputFile(content, filePath) {
	await promises.mkdir(path.dirname(filePath), { recursive: true });

	await writeFileAtomic(path.normalize(filePath), stripAnsi(content));
}

module.exports = writeOutputFile;
```

package.json
```diff
@@ -28,14 +28,16 @@
...
+    "bin/**/*.{js,mjs,cjs}",
+    "lib/**/*.{js,mjs,cjs}",
     "!**/__tests__/**",
     "!lib/testUtils/**",
     "types/stylelint/index.d.ts"
   ],
   "scripts": {
     "benchmark-rule": "node scripts/benchmark-rule.mjs",
+    "build": "rollup -c",
+    "postinstall": "npm run build",
```
ybiquitous, line 40: TODO: Remove this `postinstall` hook when the migration finishes.

This hook is necessary to prevent `npm run lint` errors on CI.

package.json
```diff
@@ -45,7 +51,7 @@
     "prepare": "husky install && patch-package",
     "release": "np --no-release-draft",
     "pretest": "npm run lint",
-    "test": "node --experimental-vm-modules node_modules/jest/bin/jest.js",
+    "test": "npm run build && node --experimental-vm-modules node_modules/jest/bin/jest.js",
```
Mouvedia, line 54: related: https://github.com/stylelint/stylelint/issues/5291#issuecomment-1328717676

romainmenke:
> Should we commit .cjs files?

For everything under `@csstools/postcss-plugins` we use this approach :
- every build output is committed
- CI rebuilds changed plugins and fails if the build output in git doesn't match a fresh build

_this steps are fairly complicated because it is a monorepo housing many different packages_

We switched to this strategy because building every plugin took to long and we didn't want CI duration to be a factor when deciding to create something new.

In practice we found that it is very rare that more than one person is changing the same plugin at the same time. So we rarely have conflicts in build outputs. I can imagine that the same is true for Stylelint rules.

We also added some git attributes to make it easier to resolve conflicts : https://github.com/csstools/postcss-plugins/blob/main/.gitattributes#L3-L6

When there is a conflict it will always result in a broken build output file, but that is fine.
This allows us to stay focused on the merge conflicts in the source files.
After resolving those we rebuild and commit the new outputs.

package.json
ybiquitous: Unfortunately, we cannot easily switch to another testing framework because of the dependency on [`jest-preset-stylelint`](https://github.com/stylelint/jest-preset-stylelint). 😓 

But when we finish the migration to ESM, it's valuable to consider switching.

ybiquitous:
@romainmenke Thanks for providing the actual example. There are many pros and cons here.

Checking built CJS files' validity on CI sounds good. 👍🏼 

I'm concerned about verbose Git commit history. 🤔

ybiquitous:
I found an advantage in committing `.cjs` files. Committed `.cjs` enables people to install `stylelint` from GitHub, for example:

```shell
npm install github:stylelint/stylelint
```

The second advantage is that we can easily compare `.mjs` and `.cjs` without building, for example, on GitHub. I'm leaning towards committing `.cjs` now, though we can go back before releasing `v16`.

scripts/build-check.mjs
ybiquitous, line 1: Usage:

```console
$ rm -f lib/*.cjs
$ npm run build-check -s
You must commit the following files:
lib/constants.cjs
```

ybiquitous:
Now ready for review!

package.json
```diff
@@ -45,7 +52,7 @@
     "prepare": "husky install && patch-package",
     "release": "np --no-release-draft",
     "pretest": "npm run lint",
```
Mouvedia, line 54: question

Shouldn't `npm run build` be in pretest instead?
That way `test-only` won't build.

scripts/build-check.mjs
Mouvedia: suggestion

1. add https://www.npmjs.com/package/rimraf
2. prepend `rimraf --glob lib/*.cjs` to _build-check_

ybiquitous: I don't want to add am extra dependency to avoid potential conflicts. Instead, 2adec1a5 cleans up `*.cjs` files in the Rollup config file.

rollup.config.mjs
```diff
@@ -1,6 +1,13 @@
 import fg from 'fast-glob';
 
-const inputFiles = fg.globSync(['lib/**/*.mjs', '!**/__tests__/**', '!lib/testUtils/**']);
+const inputFiles = fg.globSync([
+	'lib/**/*.mjs',
+	'!**/__tests__/**',
+	'!lib/testUtils/**',
+
+	// NOTE: We cannot support CJS for `cli.mjs` since the `meow` dependency is pure ESM.
```
Mouvedia, line 10: There's an alternative approach: switching to yargs.
e.g. https://github.com/conventional-changelog/get-pkg-repo/issues/60#issuecomment-888615817

ybiquitous: Ya, I still have the same concern. Can you open an issue about that to share the `meow` CJS problem?

ybiquitous: Okay, let's revisit it if a problem happens.

rollup.config.mjs
```diff
@@ -0,0 +1,14 @@
...
+const inputFiles = fg.globSync(['lib/**/*.mjs', '!**/__tests__/**', '!lib/testUtils/**']);
+
+export default inputFiles.map((input) => {
+	return {
+		input,
+		output: {
+			format: 'cjs',
+			file: input.replace('.mjs', '.cjs'),
+			generatedCode: 'es2015',
+		},
```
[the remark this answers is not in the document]
ybiquitous: Sounds good since we have `"esModuleInterop": true` in `tsconfig.json` 👍🏼 
24ff03e6

package.json
ybiquitous: `pretest` isn't run since we've specified `--ignore-scripts` to split `test` and `lint` jobs:

https://github.com/stylelint/.github/blob/8091030182f57ca1704710cfb926d9be45eb4a0e/.github/workflows/test.yml#L69

So I've just removed `npm run build` from `test` via c9a99a30.

rollup.config.mjs
```diff
@@ -0,0 +1,14 @@
+import fg from 'fast-glob';
+
+const inputFiles = fg.globSync(['lib/**/*.mjs', '!**/__tests__/**', '!lib/testUtils/**']);
+
+export default inputFiles.map((input) => {
+	return {
```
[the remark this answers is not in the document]
ybiquitous: `rollup-plugin-cleanup` seems unmaintained, so I'd like to avoid installing it.
Let's revisit it if comments might raise a problem.

ybiquitous:
All reviews are addressed. Please review this again. 🙏🏼

lib/constants.cjs
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+Object.defineProperty(exports, Symbol.toStringTag, { value: 'Module' });
```
Mouvedia, line 3: question

Can you add a comment that explains why it's needed?
Or is it a practice that we have introduced in another PR? In that case which one?
I am more familiar with
```js
if (typeof Symbol !== 'undefined' && Symbol.toStringTag) {
  Object.defineProperty(exports, Symbol.toStringTag, { value: 'Module' });
}
Object.defineProperty(exports, '__esModule', { value: true });
```

tsconfig.json
```diff
@@ -20,5 +20,5 @@
 		}
 	},
 	"include": ["lib", "types", "package.json"],
-	"exclude": ["**/*.test.js", "**/__tests__/"]
+	"exclude": ["**/*.test.js", "**/__tests__/", "lib/**/*.cjs"]
```
Mouvedia, line 23: off topic

Why would we ignore `"**/*.test.js"` exclusively?
i.e. why not mjs/cjs as well?

lib/constants.cjs
ybiquitous: This is an auto-generated code by Rollup. And [`Symbol` has been supported since older Node.js versions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#browser_compatibility), so I think extra checking is unnecessary.

tsconfig.json
ybiquitous: Because many modifications are necessary to pass type-check for test files.

`"**/*.test.js"` is redundant. We can simplify the config. I'll open another PR to fix it.

ybiquitous: I've opened PR #7059 to support also `"**/*.test.mjs"`. And I reconsider it makes sense to keep `"**/*.test.js"`, for example, see `system-tests/001/*.test.mjs`.

package.json
XhmikosR: Won't this fire for consumers of the npm package too?

jeddy3:
@ybiquitous Thank you for already moving the migration to ESM along so far!

I've been catching up on the discussions in https://github.com/stylelint/stylelint/issues/5291, and I wanted to check how we feel about https://github.com/stylelint/stylelint/issues/5291#issuecomment-1603467243:

> Are you sure it's ok for some plugins to import from the ESM Stylelint modules, and some to require from the CJS ones, at the same time when linting a project occurs? That won't cause problems?

Is there any way for us to test this?

jeddy3:
> I think we need to change require() to import() ([dynamic import](https://nodejs.org/api/esm.html#import-expressions)) to support both ESM and CJS. This means we need to introduce a Promise-based API, keeping backward compatibility as possible.

I believe so. We can see if keeping backward compatibility is possible with our resources.

How do we move forward with this pull request, e.g. shall we merge so we can keep moving forward with trying dynamic imports to uncover any potential issues? 

(I'm going offline for a week for a festival. I've approved the PR in case this is the route we want to go down, and I don't want to block you.)

jeddy3 (approved):
Many thanks again for driving the migration to ESM forward!

ybiquitous:
> How do we move forward with this pull request, e.g. shall we merge so we can keep moving forward with trying dynamic imports to uncover any potential issues?

I suggest gradually rewriting the current code to ESM to make each PR easier to review. I guess the most difficult part may be around importing (e.g. `require()` or `import-lazy`)

scripts/build-check.mjs
Mouvedia: FYI one of our dependencies has it as a dependency so it's available in the `.bin` folder.
Using `rmSync` is fine too.

PR #7039, opened 2023-07-03
Revert removed changelog entry

jeddy3:

None.

We've consistently followed the conventions of:

- Fixed & Security = Patch
- Added & Deprecated = Minor
- Removed & Changed = Major

This pull request removes the inconsistent changelog entry, which isn't user-facing and is "patch" for "Removed".

### DIFF
--- a/.changeset/fifty-hotels-admire.md
+++ b/.changeset/fifty-hotels-admire.md
@@ -1,5 +1,0 @@
----
-"stylelint": patch
----
-
-Removed: `postcss-media-query-parser` dependency


romainmenke:
> Removed & Changed = Major

Good to know 🙇 

-----

This was initially : 

```md
Fixed: remove `postcss-media-query-parser` dependency
```

https://github.com/stylelint/stylelint/pull/6999#discussion_r1246568158

Should it be reverted to that?

jeddy3:
> Should it be reverted to that?

I don't think so. It's not user-facing, and "Fixed: remove" will be inconsistent with the other entries in the changelog.

ybiquitous:
Um... I acknowledge the inconsistency, but will not PR #6999 affect people, especially plugin authors using `postcss-media-query-parser`?

I think we should inform people about the removed dependency in some way. Any thoughts?

jeddy3:
I don't think we need to as those changes were a refactor that didn't change our public API (for consumers and plugin authors). However, how about we include a second tweet as part of the release to let people know about and draw attention to the new parser? Something along the lines of:

"Under the hood, we've replaced our media query parser with @romainmenke's more modern and spec-compliant one. If you're a plugin author targeting media queries, we recommend using this parser too: https://www.npmjs.com/package/@csstools/media-query-list-parser"

That way people are informed and the changelog remains for our public API.

ybiquitous (approved):
That sounds good. 👍🏼

PR #7041, opened 2023-07-04
Fix `selector-type-case` performance

romainmenke:

Before :

```
Warnings: 0
Mean: 298.4487499090909 ms
Deviation: 21.21196081587678 ms
```

After :

```
Warnings: 0
Mean: 103.26133328571427 ms
Deviation: 9.239254247553538 ms
```

### DIFF
--- a/lib/rules/selector-type-case/index.js
+++ b/lib/rules/selector-type-case/index.js
@@ -22,6 +22,10 @@ const meta = {
 	fixable: true,
 };
 
+const STARTS_A_TAG_NAME_REGEX = /(?:[^.#[:\w-]|^)\w/;
+const ANY_UPPER_CASE_REGEX = /[A-Z]/;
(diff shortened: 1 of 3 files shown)


lib/rules/selector-type-case/index.js
```diff
@@ -22,6 +22,10 @@ const meta = {
 	fixable: true,
 };
 
+const STARTS_A_TAG_NAME_REGEX = /(?:[^.#[:\w-]|^)\w/;
```
ybiquitous, line 25: [ask] Can we use `[a-zA-Z]` instead of `\w`?

romainmenke: yes!

https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name

ybiquitous (approved):
Thank you. LGTM 👍🏼

Issue #7040, opened 2023-07-04
Security fix for vulnerability in `semver`


1EDExg0ffyXfTEqdIUAYNZGnCeajIxMWd2vaQeP:
### What minimal example or steps are needed to reproduce the bug?

npm install stylelint@15.9.0
```
added 176 packages, and audited 177 packages in 8s

33 packages are looking for funding
  run `npm fund` for details

6 moderate severity vulnerabilities

To address all issues, run:
  npm audit fix

Run `npm audit` for details.
```

Running `npm audit:`

```
# npm audit report

semver  <7.5.2
Severity: moderate
semver vulnerable to Regular Expression Denial of Service - https://github.com/advisories/GHSA-c2qf-rxjj-qqgw
fix available via `npm audit fix --force`
Will install stylelint@7.13.0, which is a breaking change
node_modules/read-pkg/node_modules/semver
  normalize-package-data  <=2.5.0
  Depends on vulnerable versions of semver
  node_modules/read-pkg/node_modules/normalize-package-data
    read-pkg  <=5.2.0
    Depends on vulnerable versions of normalize-package-data
    node_modules/read-pkg
      read-pkg-up  <=7.0.1
      Depends on vulnerable versions of read-pkg
      node_modules/read-pkg-up
        meow  3.4.0 - 9.0.0
        Depends on vulnerable versions of read-pkg-up
        node_modules/meow
          stylelint  >=8.0.0
          Depends on vulnerable versions of meow
          node_modules/stylelint

6 moderate severity vulnerabilities

To address all issues (including breaking changes), run:
  npm audit fix --force
```

### What minimal configuration is needed to reproduce the bug?

N/A

### How did you run Stylelint?

I don't

### Which Stylelint-related dependencies are you using?

stylelint@15.9.0

### What did you expect to happen?

no vulnerabilities

### What actually happened?

npm install outputs

### Do you have a proposal to fix the bug?

Possibly update meow to version 10 or later, which removed the vulnerable verison of `semver` from the dependency tree.

Mouvedia:
blocked for the same reason as #5042
see https://github.com/stylelint/stylelint/issues/5042#issuecomment-1609508268

Issue #5042, opened 2020-11-18
Fix invalid option regression on Windows 10


bmpeg:
Hi, I get the following error in my project. i am using vuetify. 

package.json
"lint:js": "eslint --fix --ext .js,.vue --ignore-path .gitignore .",
"lint:style": "stylelint **/*.{vue,css} --ignore-path .gitignore",
"lint": "npm run lint:js && npm run lint:style --fix"

run npm run lint

Invalid option "--allow-emptyinput". Did you mean "--allow-empty-input"?
Invalid option "--disable-defaultignores". Did you mean "--disable-default-ignores"?
Invalid option "--reportinvalid-scope-disables". Did you mean "--report-invalid-scope-disables"?

why could it be? Thanks.

jeddy3:
Try quoting your glob, e.g.

```
"lint:style": "stylelint \"**/*.{vue,css}\" --ignore-path .gitignore",
```

If that doesn't work, please create a new issue using the template and link to a reproducible repo.

bmpeg:
OS: Win 10
I am using visual studio code
Nuxtjs project and vuetify

I am run > npx stylelint --help

Invalid option "--allow-emptyinput". Did you mean "--allow-empty-input"?
Invalid option "--disable-defaultignores". Did you mean "--disable-default-ignores"?
Invalid option "--reportinvalid-scope-disables". Did you mean "--report-invalid-scope-disables"?

I guess it's been running the following codes respectively;
npm install --save-dev node-sass sass-loader 
npm uninstall --save-dev node-sass sass-loader

bmpeg:
I also stick with the same problem when I create a new nuxt project. 

There may be a problem with the computer. How can I repair the stylelint?

jeddy3:
Can you run `npx stylelint` outside of any nuxt project folders, please? And then paste the complete output here.

bmpeg:
![image](https://user-images.githubusercontent.com/45894806/99522407-a5ecc680-29a6-11eb-96f6-4636e21ebc13.png)

jeddy3:
Thank you. There looks to be a regression on Windows with `13.8.0`.

Do any contributors have access to a Windows 10 machine to try to reproduce this?

hudochenkov:
I just checked on Windows 10 machine. I don't have such regression. Running `npx stylelint` and `npx stylelint --version` works as expected. And I don't have any globally installed npm packages there.

bmpeg:
I use the following packages

[...]

PR #7043, opened 2023-07-04
Security fix for `semver` vulnerability

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/5042
Closes https://github.com/stylelint/stylelint/issues/7040

Not sure how to describe this change within the conventions of the project.
Suggestions welcome.

-----

How do we verify that #5042 is fixed?

Comments within that issue describe that updating `meow` is the solution, but it would be nice to verify this and to have a test that prevents regressions.

----

#7040 :

```
npm audit --production
found 0 vulnerabilities
```

### DIFF
--- a/.changeset/twenty-camels-promise.md
+++ b/.changeset/twenty-camels-promise.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Security: fix for `semver` vulnerability
--- a/lib/cli.mjs
+++ b/lib/cli.mjs
@@ -1,5 +1,3 @@
-'use strict';
-
 import picocolors from 'picocolors';
 const { dim, red } = picocolors;
 
--- a/lib/cli.mjs
+++ b/lib/cli.mjs
@@ -1,23 +1,30 @@
 'use strict';
 
-const { EOL } = require('os');
-const meow = require('meow');
-const path = require('path');
-const { red, dim } = require('picocolors');
-const resolveFrom = require('resolve-from');
+import picocolors from 'picocolors';
+const { dim, red } = picocolors;
 
-const { isPlainObject } = require('./utils/validateTypes');
-const checkInvalidCLIOptions = require('./utils/checkInvalidCLIOptions');
-const printConfig = require('./printConfig');
-const standalone = require('./standalone');
-const writeOutputFile = require('./writeOutputFile');
-const resolveCustomFormatter = require('./resolveCustomFormatter');
-const {
+import { EOL } from 'os';
+import meow from 'meow';
+import path from 'path';
+import resolveFrom from 'resolve-from';
+
+import checkInvalidCLIOptions from './utils/checkInvalidCLIOptions.js';
+import { isPlainObject } from './utils/validateTypes.js';
+import printConfig from './printConfig.js';
+import resolveCustomFormatter from './resolveCustomFormatter.js';
+import standalone from './standalone.js';
+import writeOutputFile from './writeOutputFile.js';
+
+import {
 	DEFAULT_CACHE_LOCATION,
-	DEFAULT_IGNORE_FILENAME,
 	DEFAULT_FORMATTER,
+	DEFAULT_IGNORE_FILENAME,
 	EXIT_CODE_ERROR,
-} = require('./constants');
+} from './constants.js';
+
+import { createRequire } from 'module';
+// @ts-expect-error
+const require = createRequire(import.meta.url);
 
 /**
  * @typedef {{
@@ -376,7 +383,7 @@ const meowOptions = {
  * @param {string[]} argv
  * @returns {Promise<any>}
  */
-module.exports = async function main(argv) {
+export default async function main(argv) {
 	const cli = buildCLI(argv);
 
 	const invalidOptionsMessage = checkInvalidCLIOptions(meowOptions.flags, cli.flags);
@@ -574,7 +581,7 @@ module.exports = async function main(argv) {
 			}
 		})
 		.catch(handleError);
-};
+}
 
 /**
  * @param {{ stack: any, code: any }} err
@@ -628,9 +635,7 @@ async function getStdin() {
  * @param {string[]} argv
  * @returns {CLIOptions}
  */
-function buildCLI(argv) {
+export function buildCLI(argv) {
 	// @ts-expect-error -- TS2322: Type 'Result<AnyFlags>' is not assignable to type 'CLIOptions'.
-	return meow({ ...meowOptions, argv });
+	return meow({ ...meowOptions, argv, importMeta: import.meta });
 }
-
-module.exports.buildCLI = buildCLI;
--- a/package.json
+++ b/package.json
@@ -25,11 +25,13 @@
   "main": "lib/index.js",
   "types": "types/stylelint/index.d.ts",
   "bin": {
-    "stylelint": "bin/stylelint.js"
+    "stylelint": "bin/stylelint.mjs"
   },
   "files": [
     "bin/**/*.js",
+    "bin/**/*.mjs",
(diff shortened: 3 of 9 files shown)


jeddy3:
> Not sure how to describe this change within the conventions of the project.

Let's lead with the security fix. We can manually add a changeset entry for the windows 10 regression.

jeddy3:
> How do we verify that https://github.com/stylelint/stylelint/issues/5042 is fixed?

Not sure. In the past, for intermittent platform-specific issues we've just waited for users to verify it's working after release. Particularly when the issue was patched in one of our dependencies.

lib/cli.mjs
```diff
@@ -1,23 +1,30 @@
 'use strict';
```
ybiquitous, line 2: [suggest] We can now remove `use strict` since it's ESM 👍🏼 

```suggestion
```

ybiquitous (commented):
@romainmenke I don't realize this solution! Thanks!

I still have two concerns:

1. `meow` requires Node.js >=14.16. Is this a breaking change? That is, should we update `engines.node` to `^14.18.0 || >= 16.0.0`?

```console
$ npm view meow@11.0.0 engines
{ node: '>=14.16' }

$ npx ls-engines@latest
...
┌───────────────────────────────────┬───────────────────────────┐
│ package engines:                  │ dependency graph engines: │
├───────────────────────────────────┼───────────────────────────┤
│ "engines": {                      │ "engines": {              │
│   "node": "^14.13.1 || >= 16.0.0" │   "node": ">= 14.18"      │
│ }                                 │ }                         │
└───────────────────────────────────┴───────────────────────────┘
...
```

2. People will not be able to `require "stylelint/lib/cli.js"`. Should this be acceptable? I guess there may be few people who need such a CJS way... 🤔

jeddy3:
> meow requires Node.js >=14.16. Is this a breaking change? That is, should we update engines.node to ^14.18.0 || >= 16.0.0?

That's a shame. It is breaking (although probably for only a handful of users).

No easier answer here. Major releases are painful for plugin authors, so we only do it once a year.

Shall we move this change to `v16`? It's a security fix, but as we're a dev tool the impact should be low.

> People will not be able to require "stylelint/lib/cli.js". Should this be acceptable?

I believe so, as it's not part of our public API.

romainmenke:
> meow requires Node.js >=14.16. Is this a breaking change? That is, should we update engines.node to ^14.18.0 || >= 16.0.0?

`meow` version 11 requires these node versions, but version 10 should be fine.
I did get a typescript warning on 10, but that might be possible to solve.

Version 10 also resolves the security issue.

I will look into this shortly :)

Mouvedia:
> `meow` version 11 requires these node versions, but version 10 should be fine.

I agree but it needs to be at least version 10.1.0.
cf https://github.com/stylelint/stylelint/issues/5042#issuecomment-1416791902

patches/meow++type-fest+1.4.0.patch
```diff
@@ -0,0 +1,14 @@
...
+--- a/node_modules/meow/node_modules/type-fest/ts41/get.d.ts
++++ b/node_modules/meow/node_modules/type-fest/ts41/get.d.ts
+@@ -92,7 +92,9 @@ type PropertyOf<BaseType, Key extends string> =
+ 		? Item
+ 		: unknown
+ 	)
++	// @ts-expect-error
+ 	: Key extends keyof WithStringKeys<BaseType>
++	// @ts-expect-error
+ 	? WithStringKeys<BaseType>[Key]
```
romainmenke, line 12: The types were incorrect at this version, but that doesn't affect the runtime of the cli.

tsconfig.json
```diff
@@ -13,11 +13,18 @@
...
 	},
 	"include": ["lib", "types", "package.json"],
-	"exclude": ["**/*.test.js", "**/__tests__/"]
+	"exclude": [
+		"**/*.test.js",
+		"**/__tests__/",
+		"**/ts41/get.d.ts",
+		"**/meow/**",
+		"./node_modules/meow/node_modules/type-fest/ts41/get.d.ts"
+	]
```
romainmenke, line 29:
```suggestion
	"exclude": ["**/*.test.js", "**/__tests__/"]
```

ybiquitous:
I also agree with updating to `meow@10` instead of `meow@11`. 👍🏼 

For your information:

```console
$ npm view meow@^10.1.0 engines
meow@10.1.0 { node: '>=12.17' }
meow@10.1.1 { node: '>=12.17' }
meow@10.1.2 { node: '^12.20.0 || ^14.13.1 || >=16.0.0' }
meow@10.1.3 { node: '^12.20.0 || ^14.13.1 || >=16.0.0' }
meow@10.1.4 { node: '^12.20.0 || ^14.13.1 || >=16.0.0' }
meow@10.1.5 { node: '^12.20.0 || ^14.13.1 || >=16.0.0' }
```

Perhaps, it may be better to update the latest version of `meow@^10.1.0` like this:

```diff
- "meow": "^10.1.0",
+ "meow": "^10.1.5",
```

ybiquitous:
In addition, here's a result of `ls-engines`:

```console
$ npx ls-engines@latest
...
┌───────────────────────────────────┬───────────────────────────┐
│ package engines:                  │ dependency graph engines: │
├───────────────────────────────────┼───────────────────────────┤
│ "engines": {                      │ "engines": {              │
│   "node": "^14.13.1 || >= 16.0.0" │   "node": ">= 14.18"      │
│ }                                 │ }                         │
└───────────────────────────────────┴───────────────────────────┘
...
Your “engines” field does not exactly match your dependency graph‘s requirements!
...
```

`ls-engines` still complains as above, but I think `meow@^10.1.0` is acceptable because `meow` itself requires `node: '^12.20.0 || ^14.13.1 || >=16.0.0'`. 👍🏼

.changeset/twenty-camels-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Security: fix for `semver` vulnerability
```
ybiquitous, line 5: [note] Our maintainer guide now doesn't mention the `Security:` prefix. This doesn't block this PR, but I'm a bit curious. 😅 

https://github.com/stylelint/stylelint/blob/a42f95582c2a928122fc1e59f9f8acc078fb9b91/docs/maintainer-guide/pull-requests.md?plain=1#L22-L23

package.json
```diff
@@ -153,7 +155,7 @@
     "is-plain-object": "^5.0.0",
     "known-css-properties": "^0.27.0",
     "mathml-tag-names": "^2.1.3",
-    "meow": "^9.0.0",
+    "meow": "^10.1.0",
```
ybiquitous, line 158: [suggest] This is a minor suggestion (also `package-lock.json`), but you can ignore it.

```suggestion
    "meow": "^10.1.5",
```

ybiquitous (approved):
Thank you for the excellent fix. LGTM 👍🏼

ybiquitous:
I think this PR needs one more approvement.

romainmenke:
> `ls-engines` still complains as above

Seems to be because of `supports-hyperlinks`, which is a direct dependency.

```json5
"node_modules/supports-hyperlinks": {
	"version": "3.0.0",
	// ...
	"engines": {
		"node": ">=14.18"
	}
},
```

Luckily not a sub dependency of `meow` v10.x :)

.changeset/twenty-camels-promise.md
jeddy3: It's an oversight. It should list all the ones from https://keepachangelog.com/en/1.0.0/. I'll open a PR.

jeddy3 (approved):
LGTM, thank you very much.

PR #7047, opened 2023-07-05
Document check of configs in release process

### LINKED ISSUE #7045: Document check of configs in release process
Ref: https://github.com/stylelint/stylelint/pull/7044

We have a chicken and egg problem. Our test in this repo will fail if we mark a rule as included in the standard config before releasing it. If we wait until after realising the config then the tests on `main` will fail.

We also have to mark it as included before releasing the config as we use a tarball version of this package for the website.

Any ideas on how to resolve this or do we need to remove this test?

jeddy3:

Closes #7045 

No, it's self-explanatory.

(Diff is mainly whitespace for test.)

### DIFF
--- a/docs/maintainer-guide/releases.md
+++ b/docs/maintainer-guide/releases.md
@@ -3,7 +3,7 @@
 1. Create a [new issue](https://github.com/stylelint/stylelint/issues/new?title=Release+%7Bversion%7D&labels=status%3A+needs+discussion) to announce the planned release:
    - include the [template checklist](#new-release-issue-template)
    - if applicable, list any new rules so that we can discuss including them in our configs
-2. If necessary, open a pull request to mark those rules in `docs/user-guide/rules.md`.
+2. If necessary, open a pull request to mark those rules in [`docs/user-guide/rules.md`](../user-guide/rules.md).
 3. If necessary, test `main` locally in the:
    1. [`stylelint/stylelint-config-recommended`](https://github.com/stylelint/stylelint-config-recommended)
    2. [`stylelint/stylelint-config-standard`](https://github.com/stylelint/stylelint-config-standard)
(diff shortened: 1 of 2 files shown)


docs/maintainer-guide/releases.md
```diff
@@ -2,30 +2,32 @@
 
 1. Create a [new issue](https://github.com/stylelint/stylelint/issues/new?title=Release+%7Bversion%7D&labels=status%3A+needs+discussion) to announce the planned release:
    - include the [template checklist](#new-release-issue-template)
-2. If necessary, test `main` locally in the:
+   - if applicable, list any new rules so that we can discuss including them in our configs
+2. If necessary, open a pull request to mark those rules in `docs/user-guide/rules.md`.
```
ybiquitous, line 6: [suggest] How about linking to the actual document?

```suggestion
2. If necessary, open a pull request to mark those rules in [`docs/user-guide/rules.md`](../user-guide/rules.md).
```

lib/rules/__tests__/index.test.mjs
```diff
@@ -74,36 +74,39 @@ describe('custom message option', () => {
...
-
-	// NOTE: The use of Promised-based APIs may cause flaky test on CI.
-	rmSync(tmpDir, { recursive: true, force: true });
-	mkdirSync(tmpDir, { recursive: true });
-	writeFileSync(path.join(tmpDir, 'package.json'), '{}');
-	execSync(
-		'npm install --silent --no-package-lock --no-audit --omit=peer stylelint-config-standard',
-		{ cwd: tmpDir },
-	);
+	// eslint-disable-next-line jest/no-disabled-tests
```
ybiquitous, line 77: [suggest] How about leaving a reason to skip it?

For example:

```suggestion
	// eslint-disable-next-line jest/no-disabled-tests -- To prevent a failure when adding a new rule to the sharable config. See #7045.
```

ybiquitous (approved):
Thank you, LGTM 👍🏼 

I've suggested two, but they're minor. You can ignore them.

Issue #7061, opened 2023-07-09
Fix `declaration-property-value-no-unknown` false negatives for typed custom properties


romainmenke:
### What is the problem you're trying to solve?

The `@property` rule makes it possible to type custom properties.

```css
@property --foo {
	syntax: "<color>";
	inherits: false;
	initial-value: #c0ffee;
}
```

Custom properties registered in this way behave subtly different, for one they are animatable.  
But they are also type safe.

`declaration-property-value-no-unknown` doesn't validate the value types for these custom properties.

```css
.foo {
  --foo: bar; /* no warning */
}
```

### What solution would you like to see?

I would like to receive a warning when assigning values of an incorrect type to a typed custom property.

----

This depends on https://github.com/csstree/csstree/pull/256

----

The rule can be modified with this bit of code :

```js
/** @type {Record<string, string>} */
const discoveredProperties = {};

const typeSafeCustomProperties = new Set();

root.walkAtRules(/^property$/i, (atRule) => {
	const propName = atRule.params.trim();

	if (!propName) return;

	const syntax = atRule.nodes?.find((x) => isDeclaration(x) && /^syntax$/i.test(x.prop));

	if (!syntax || !isDeclaration(syntax)) return;

	typeSafeCustomProperties.add(propName);
	discoveredProperties[propName] = syntax.value;
});

const forkedLexer = fork({
	properties: {
		...propertiesSyntax,
		...discoveredProperties,
	},
	types: typesSyntax,
}).lexer;
```

romainmenke:
Unsure how to best label this issue :)

- are these false negatives?
- is this a new sub-feature?
- should this be behind a plugin option?

Mouvedia:
To me, it should a separate rule.
Sadly it would also be affected by #4088.

romainmenke:
I hadn't considered this to be a separate rule because the subject (values of declarations) is the same and what is being checked (no unknown values) is also the same.

If it would be a separate rule it would also overlap with `declaration-property-value-no-unknown` if support is added for type checking `var()`.

```css
@property --foo {
	syntax: "<color>";
	inherits: false;
	initial-value: #c0ffee;
}

:root {
  --foo: red;
}

.component {
  width: var(--foo); /* I expect a warning */
}
```

I think this sub-feature would only add support for modern CSS syntax on the same subjects.

Mouvedia:
[...]

PR #7078, opened 2023-07-13
Fix `declaration-property-value-no-unknown` false negatives for typed custom properties

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/7061

I chose not to validate the contents of the `syntax` property in a `@property` rule.

It is always valid if it also works in a browser.
It might be invalid in browsers and still be processed by CSSTree as a valid syntax definition.

I think this is fine as it future proofs this rule.
We will not have to change this rule each time the restrictions here are relaxed and more of the full [value definition syntax](https://drafts.csswg.org/css-values-3/#value-defs) becomes allowed.


lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
```diff
@@ -138,6 +207,36 @@ testRule({
...
+					syntax: "<color>";
+					inherits: false;
+					initial-value: #c0ffee;
+				}
+			`,
+			message: messages.rejected('--foo', '10px'),
+			line: 2,
+			column: 16,
+			endLine: 2,
+			endColumn: 20,
```
ybiquitous, line 223: [suggest] This isn't required, but I suggest using `stripIndent` to easily check `column` values. The helper removes TAB characters in a template string literal. For example:

https://github.com/stylelint/stylelint/blob/c1aefb4d5c3c15d035a658a8e6f53be2fd5de283/lib/rules/color-function-notation/__tests__/index.mjs#L129-L136

romainmenke: Nice, thank you 🙇

romainmenke:
Intentionally a draft :)
I would like to do a self review after merging the previous change in this same plugin, to make sure that the combined change still makes sense.

lib/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -84,10 +124,22 @@ const rule = (primary, secondaryOptions) => {
...
+			// `initial-value` has a incorrect syntax definition.
+			// In reality everything is valid.
+			if (
+				/^initial-value$/i.test(prop) &&
+				decl.parent &&
+				isAtRule(decl.parent) &&
+				/^property$/i.test(decl.parent.name)
+			) {
+				return;
+			}
```
romainmenke, line 141: I didn't cache the regexp's here because this entire block should be removed.

I don't mind optimizing the code anyway.
But I am really hoping that csstree will receive some attention in the near future.

lib/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -63,9 +65,47 @@ const rule = (primary, secondaryOptions) => {
 			return valuePattern && matchesStringOrRegExp(value, valuePattern);
 		};
 
-		const propertiesSyntax = (secondaryOptions && secondaryOptions.propertiesSyntax) || {};
+		const propertiesSyntax = {
+			// Take a shallow clone as this object will be appended to.
+			...((secondaryOptions && secondaryOptions.propertiesSyntax) || {}),
+		};
 		const typesSyntax = (secondaryOptions && secondaryOptions.typesSyntax) || {};
```
ybiquitous, line 72: [suggest] Using the `?.` operator can simplify the code a bit:

```suggestion
		const propertiesSyntax = {
			// Take a shallow clone as this object will be appended to.
			...(secondaryOptions?.propertiesSyntax ?? {}),
		};
		const typesSyntax = secondaryOptions?.typesSyntax ?? {};
```

lib/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -63,9 +65,47 @@ const rule = (primary, secondaryOptions) => {
...
 
+		const typedCustomPropertyNames = new Map();
```
ybiquitous, line 74: [suggest] Giving type parameters makes the code a bit type-safer:

```suggestion
		/** @type {Map<string, string>} */
		const typedCustomPropertyNames = new Map();
```

lib/rules/declaration-property-value-no-unknown/index.js
ybiquitous: Considering consistency, it seems better to fix `ignorePrpoperties` too:

```diff
-Object.entries((secondaryOptions && secondaryOptions.ignoreProperties) || {})
+Object.entries(secondaryOptions?.ignoreProperties ?? {})
```

https://github.com/stylelint/stylelint/pull/7078/files#diff-e97a8a2ce49674252ada6c3a6f0dde2f19422f662d0afe69be865ddc97b418a5R57

But it's not required, and it's possible to keep it as-is. 👌🏼

ybiquitous (commented):
@romainmenke Thanks for the Pull Request! I think this is great. 👍🏼 

I've commented on a few refactoring suggestions, but they're minor things. 😃

lib/rules/declaration-property-value-no-unknown/index.js
romainmenke: Good catch! 🙇

lib/rules/declaration-property-value-no-unknown/index.js
romainmenke: Thank you!

ybiquitous (approved):
Thank you. LGTM! 👍🏼

Issue #7062, opened 2023-07-09
Fix `declaration-property-value-no-unknown` false negatives for nested declarations


romainmenke:
### What minimal example or steps are needed to reproduce the bug?

```css
.foo {
	color: 10px; /* has a warning */
	color: red;
}

.foo {
	@media (screen) {
		color: 10px; /* no warning */
		color: red;
	}
}
```

_I added `color: red` to make sure there weren't any false positives._

### What minimal configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "declaration-property-value-no-unknown": true
  }
}
```

### How did you run Stylelint?

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0AzCEAEwA6A7bALpADYQBO8GAjAAwAOAHgNx6EnmVkxQu4C+rXKnRZWBAAIBbbgEsAhhgAUAZzBcYuAJSj8BNhFIVq9ZmP2HO3XoQH8QAGhAoZxGADk50hCBgNPdVyQwZWUHcAhcZwBzbxxcDAxsEDIAV1dlJMo4hISk2DBiOTI5AhkIgFo6Mgg6GDICAE9ygDc5YhSYctwIcpTcAGtugHd8BAwCVJg8BNs+MMhImSiAMXJJEu8AK2UIsNg6UMRsxJBlRtdiGVw9MaTCghgzpPtpk7OGi6uCcoXo8rO5LgoEUoJkTvdHjc8HM+EA)

### Which Stylelint-related dependencies are you using?

```json
{
  "stylelint": "latest",
  "stylelint-config-standard": "latest"
}
```

### What did you expect to happen?

I expected to see two errors.  
One for each `color: 10px;`.

### What actually happened?

Only one error was reported, for the regular, unnested rule.

romainmenke:
https://github.com/stylelint/stylelint/blob/c812dae9da050f903d31f4786fff89929c56b0f6/lib/rules/declaration-property-value-no-unknown/index.js#L114

I think this was intended to differentiate at rule descriptors from declarations.

Maybe we can use the list of allowed nested at rules : https://drafts.csswg.org/css-nesting/#conditionals

> Specifically, these rules are capable of being [nested group rules](https://drafts.csswg.org/css-nesting/#nested-group-rules):
[...]

Issue #6421, opened 2022-10-22
Rules for nested CSS following css-nesting-1


romainmenke:
### What is the problem you're trying to solve?

`css-nesting-1` is being implemented in Chrome. The specification isn't final yet and it might change in the next few weeks.

There are some issues where linters might help authors write correctly nested CSS.

```css
.foo {
  bar & {} /* ambiguous nesting */
}
```

```css
::before {
  bar & {} /* pseudo elements are not allowed */
}
```

At the moment we still need to wait for the final specification for all issues to surface.

### What solution would you like to see?

Some stylelint rules that warn authors when they write nested selectors that are invalid or ambiguous.

Is there interest to add this to stylelint?

jeddy3:
@romainmenke Thanks for the request and for using the template.

> Is there interest to add this to stylelint?

Yes, absolutely! I'm glad you created an issue. I've been meaning to open one myself since the discussion about the nesting spec in https://github.com/stylelint/stylelint/issues/6294#issuecomment-1236131389.

In that issue, we touched on these two nesting patterns:

```css
/* this is valid, but probably unintentional */
{
  && {}
}
```

```css
/* these are equivalent */
a {
  @media (orientation: landscape) { color: red; }
}

a {
  @media (orientation: landscape) { & { color: red; } }
}
```

> Some stylelint rules that warn authors when they write nested selectors that are invalid or ambiguous.

Rules that warn about invalid or ambiguous patterns are a perfect fit for Stylelint. We should eventually add rules to catch:

```css
.foo {
  bar & {} /* ambiguous nesting */
}
```

And

```css
::before {
  bar & {} /* pseudo elements are not allowed */
}
```

The patterns above could fit into one of our existing or upcoming categories:

 - [`no-invalid`](https://stylelint.io/user-guide/rules#invalid) - disallow invalid syntax
 - `no-unmatchable` - disallow valid but unmatchable things, e.g.:
     - https://github.com/stylelint/stylelint/issues/5907
     - https://github.com/stylelint/stylelint/issues/5575
- [`notation`](https://stylelint.io/user-guide/rules#notation) - enforce one representation of things that have multiple 

Or we may need a new category or two, e.g. `no-ambiguous`.

>  The specification isn't final yet and it might change in the next few weeks.

[...]

PR #7079, opened 2023-07-13
Fix `declaration-property-value-no-unknown` false negatives for nested declarations

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/7062

No, it's self-explanatory.

### DIFF
--- a/lib/rules/declaration-property-value-no-unknown/index.js
+++ b/lib/rules/declaration-property-value-no-unknown/index.js
@@ -28,6 +28,8 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown',
 };
 
+const CONDITIONAL_RULES_ALLOWED_IN_NESTING = /^(?:media|supports|layer|scope|container)$/i;
+
 /** @type {import('stylelint').Rule} */
 const rule = (primary, secondaryOptions) => {
 	return (root, result) => {
@@ -111,7 +113,7 @@ const rule = (primary, secondaryOptions) => {
 			}
 
 			const { error } =
-				parent && isAtRule(parent)
+				parent && isAtRule(parent) && !CONDITIONAL_RULES_ALLOWED_IN_NESTING.test(parent.name)
 					? forkedLexer.matchAtruleDescriptor(parent.name, prop, cssTreeValueNode)
 					: forkedLexer.matchProperty(prop, cssTreeValueNode);
 
--- a/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
+++ b/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
@@ -10,6 +10,9 @@ testRule({
 			code: 'a { top: 0; }',
 		},
 		{
+			code: 'a { @media screen { top: 0; } }',
+		},
+		{
 			code: 'a { margin: auto 10em; }',
 		},
 		{
@@ -65,6 +68,14 @@ testRule({
 			column: 10,
 			endLine: 1,
 			endColumn: 17,
+		},
+		{
+			code: 'a { @media screen { top: unknown; } }',
+			message: messages.rejected('top', 'unknown'),
+			line: 1,
+			column: 26,
+			endLine: 1,
+			endColumn: 33,
 		},
 		{
 			code: 'a { top: red; }',


lib/rules/declaration-property-value-no-unknown/index.js
```diff
@@ -28,6 +28,8 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown',
 };
 
+const CONDITIONAL_RULES_ALLOWED_IN_NESTING = /^(?:media|supports|layer|scope|container)$/i;
```
ybiquitous, line 31: [ask] This isn't required, but does defining these rules in `reference/atKeywords.js` make sense like this?

```js
// https://drafts.csswg.org/css-nesting/#conditionals
const nestingSupportedAtKeywords = new Set([
	'container',
	'layer',
	'media',
	'scope',
	'supports',
]);
```

This might be available for other rules in the future. Any thoughts?

Mouvedia: > Any thoughts?

It might be useful for #6421
see https://www.w3.org/TR/css-nesting-1/#example-f2b8b339

lib/reference/atKeywords.js
```diff
@@ -2,6 +2,9 @@
 
 const uniteSets = require('../utils/uniteSets.js');
 
+// https://www.w3.org/TR/css-nesting-1/#conditionals
+const nestingSupportedAtKeywords = new Set(['container', 'layer', 'media', 'scope', 'supports']);
```
romainmenke, line 6: This is different from `pageMarginAtKeywords`.

These rules or their grouping is not defined by nesting.
They each have their own definition but happen to have extra allowances in nesting.

There might be a future groupings with only `media, scope, container` or `media, supports, when, else`.

But `atKeywords` must always contain all known at keywords.
Even if a utility grouping is altered or removed.

----

I made it a separate Set and did not include it in the list of `uniteSets`.

Instead I added a unit test to make sure that all items in `nestingSupportedAtKeywords` also exist in `atKeywords`.

Thoughts?

lib/reference/atKeywords.js
```diff
@@ -54,4 +58,5 @@ const atKeywords = uniteSets(pageMarginAtKeywords, [
 
 module.exports = {
 	atKeywords,
+	nestingSupportedAtKeywords,
 };
```
Mouvedia, line 62: offtopic

maybe we could replace `uniteSets` by https://www.npmjs.com/package/set.prototype.union?

lib/reference/__tests__/atKeywords.test.mjs
```diff
@@ -0,0 +1,9 @@
+import { atKeywords, nestingSupportedAtKeywords } from '../atKeywords.js';
+
+describe('atKeywords', () => {
+	it('nestingSupportedAtKeywords is a subset of all atKeywords', () => {
+		for (const keyword of nestingSupportedAtKeywords) {
+			expect(atKeywords).toContain(keyword);
+		}
+	});
+});
```
Mouvedia, line 9: minor

That would be the first test for the reference folder.
I think just by using `uniteSets` at l29 you can avoid adding it.
i.e. be explicit in the code

romainmenke: > But atKeywords must always contain all known at keywords.
Even if a utility grouping is altered or removed.

Then you could run into the scenario that an at keyword was only added in a subset and then later removed from that subset without being added to the full set.

This is easy to spot when there is only one sub set, but would be harder to notice when there are several.

-------

Adding it to `uniteSets` would be explicit and it is how I initially wrote it.
But it wouldn't be obvious why there is a subset being added to the whole list when each item in the subset already exists in the whole list.

I don't mind changing it back :)

Mouvedia: I meant

```js
const atKeywords = uniteSets(pageMarginAtKeywords, [
	'annotation',
	'apply',
	'character-variant',
	'charset',
	'counter-style',
	'custom-media',
	'custom-selector',
	'document',
	'font-face',
	'font-feature-values',
	'import',
	'keyframes',
	'namespace',
	'nest',
	'ornaments',
	'page',
	'property',
	'scroll-timeline',
	'styleset',
	'stylistic',
	'swash',
	'viewport',
], nestingSupportedAtKeywords);
```

lib/reference/atKeywords.js
ybiquitous: Sounds good 👍🏼

lib/reference/atKeywords.js
ybiquitous: Interesting. But I'd like to minimize dependencies as possible.

ybiquitous (approved):
Thank you. LGTM 👍🏼

jeddy3 (approved):
LGTM, thank you!

romainmenke:
Thank you all for the reviews 🙇

PR #7101, opened 2023-07-26
Migrate small utility modules to ESM

ybiquitous:

Ref #5291

This change rewrites small utility modules (less dependent on 3rd-party packages) to ESM.


lib/utils/getSchemeFromUrl.cjs
```diff
@@ -1,15 +1,14 @@
 'use strict';
 
-const { URL } = require('url');
-
```
ybiquitous, line 4: [note] `URL` is now a global object. See https://nodejs.org/api/url.html#class-url

lib/utils/isAfterSingleLineComment.mjs
```diff
@@ -0,0 +1,19 @@
+import isSharedLineComment from './isSharedLineComment.mjs';
+
+/**
+ * @param {import('postcss').Node} node
+ * @returns {boolean}
```
ybiquitous, line 5: [note] Specifying a return type makes code clearer as documentation. Also, it's valid for type-check.

lib/__tests__/ignore.test.mjs
```diff
@@ -176,7 +176,7 @@ test('specified `ignorePath` file ignoring one file', async () => {
 			},
 			ignorePath: [fixtures('ignore.txt')],
 		}),
-	).rejects.toThrow(AllFilesIgnoredError); // all files ignore
+	).rejects.toThrow(new AllFilesIgnoredError());
```
ybiquitous, line 179: [note] Just specifying a constructor fails the test. `new` is needed.

jeddy3 (approved):
Excellent, thank you!

PR #7108, opened 2023-07-27
Group Dependabot updates by `@csstools/*` parser packages

ybiquitous:

Ref:
- #7105 
- #7106
- #7107

This change aims to assemble Dependabot Pull Requests about `@csstools/*` parser packages.

@romainmenke Any thoughts?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,9 @@ updates:
       - 'pr: dependencies'
     versioning-strategy: increase
     groups:
+      csstools-parser:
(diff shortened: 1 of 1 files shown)


romainmenke (approved):
I think this will help 👍 

You will always want all of these updates together or none of them.

If updating any of these packages introduces a regression you want to prevent updating the others. If this happens it is most likely because we shipped a bug and then it should be fixed upstream.

ybiquitous:
Thanks for the feedback! 👍🏼

PR #7118, opened 2023-08-07
Fix CLI to print unfixed errors when --fix is used with --stdin

### LINKED ISSUE #6604: Fix `--fix` not reporting unfixable errors to stderr

Honestly not sure if this was by design or a bug, but I would expect `stdin` to work similarly to linting files. When you lint files it fixes what it can in the files, and any remaining lint errors are output to `stdout`. Of course with `stdin` the changes are output to `stdout` as expected; it's the errors that aren't output at all, when I would expect them in `stderr`.

Use a CSS file that has lint errors that cannot be automatically fixed. Here's a contrived "test.css" that violates the unfixable `no-descending-specificity`, which is set to true by `stylelint-config-recommended`:

```css
.component {
  &.dark .header {
    color: white;
  }

  .header {
    color: black;
  }
}
```

Then in CLI: `stylelint --fix --stdin --stdin-filename "test.css" < "test.css"`

Here's my current configuration in package.json:

```json
    "stylelint": {
        "extends": [
            "stylelint-config-standard",
            "stylelint-config-tailwindcss"
        ],
        "ignoreFiles": [
            "node_modules/**",
            "vendor/**"
        ],
        "rules": {
            "import-notation": ["string", { "severity": "warning", "disableFix": true }],
            "number-max-precision": 7,
            "selector-max-universal": 1,
            "value-keyword-case": ["lower", { "camelCaseSvgKeywords": true }]
        }
    },
```

CLI

14.16.1

`npx stylelint --fix --stdin --stdin-filename "test.css" < "test.css"` (output of file, output of errors):

<pre>.component {
  &.dark .header {
    color: white;
  }

 .header {
    color: black;
  }
}

test.css
 6:3  ✖  Expected selector ".component .header" to come before selector ".component.dark .header"  no-descending-specificity

1 problem (1 error, 0 warnings)</pre>

exit code is `2`

---

Or when outputting to files (easier way to see this), for example:

`npx stylelint --fix --stdin --stdin-filename "test.css" < "test.css" 1>changed-file.css 2>lint-errors.txt`

changed-file.css:
```css
.component {
  &.dark .header {
    color: white;
  }

  .header {
    color: black;
  }
}
```

lint-errors.txt:
```

test.css
 6:3  ✖  Expected selector ".component .header" to come before selector ".component.dark .header"  no-descending-specificity

```

exit code is `2`

### What actually happened?

`npx stylelint --fix --stdin --stdin-filename "test.css" < "test.css"` (output of file only):

<pre>.component {
  &.dark .header {
    color: white;
  }

 .header {
    color: black;
  }
}</pre>

exit code is `2`

---

Or when outputting to files (easier way to see this), for example:

`npx stylelint --fix --stdin --stdin-filename "test.css" < "test.css" 1>changed-file.css 2>lint-errors.txt`

changed-file.css:
```css
.component {
  &.dark .header {
    color: white;
  }

  .header {
    color: black;
  }
}
```

lint-errors.txt (empty):
```
```

exit code is `2`

No

### Proposal to fix the bug

Take this with a grain of salt since I'm not familiar with all the code, but I think it would work to add the reported lint errors to `process.stderr.write` within this conditional for setting the exit code 🤷:
https://github.com/stylelint/stylelint/blob/f8988a10455be84b27c300631c902e289cfc730c/lib/cli.js#L543

haocheng6:

Closes #6604 

- This PR makes stylelint output fixed code to stdout when `--fix` is used with `--stdin`.
- Currently stylelint outputs linting errors, deprecation warnings, descriptionless disables, etc to stdout. This PR changes it to stderr.

### DIFF
--- a/lib/__tests__/getCleanOutput.mjs
+++ b/lib/__tests__/getCleanOutput.mjs
@@ -1,0 +1,23 @@
+import stripAnsi from 'strip-ansi';
+
+const symbolConversions = new Map();
+
+symbolConversions.set('ℹ', 'i');
+symbolConversions.set('✔', '√');
+symbolConversions.set('⚠', '‼');
+symbolConversions.set('✖', '×');
+
+/**
+ * @param {string} output
+ *
+ * @returns {string}
+ */
+export default function getCleanOutput(output) {
+	let cleanOutput = stripAnsi(output).trim();
+
+	for (const [nix, win] of symbolConversions.entries()) {
+		cleanOutput = cleanOutput.replace(new RegExp(nix, 'g'), win);
+	}
+
+	return cleanOutput;
+}
--- a/lib/cli.mjs
+++ b/lib/cli.mjs
@@ -555,11 +555,15 @@ export default async function main(argv) {
 
 	return standalone(options)
 		.then((linted) => {
-			if (!linted.output) {
+			if (!linted.output && !linted.code) {
 				return;
 			}
 
-			process.stdout.write(linted.output);
+			if (linted.code) {
+				process.stdout.write(linted.code);
+			}
+
+			process.stderr.write(linted.output);
 
 			if (options.outputFile) {
 				writeOutputFile(linted.output, options.outputFile).catch(handleError);
--- a/lib/__tests__/cli.test.mjs
+++ b/lib/__tests__/cli.test.mjs
@@ -4,7 +4,7 @@ import { Readable } from 'node:stream';
 import { fileURLToPath } from 'node:url';
 import path from 'node:path';
 
-import { jest } from '@jest/globals';
+import { expect, jest } from '@jest/globals';
 import stripAnsi from 'strip-ansi';
 
 import readJSONFile from '../testUtils/readJSONFile.mjs';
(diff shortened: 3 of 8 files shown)


lib/__tests__/getCleanOutput.mjs
ybiquitous, line 1: [suggest] Since this file is almost the same as `lib/formatters/__tests__/prepareFormatterOutput.mjs`, could you unify them with a new file `lib/testUtils/getCleanOutput.mjs`?

https://github.com/stylelint/stylelint/blob/48a28b4bbca215499569c9dc56ce478d194a35b9/lib/formatters/__tests__/prepareFormatterOutput.mjs#L10

ybiquitous (changes requested):
@haocheng6 Thanks a lot for creating this PR! I'm so afraid, but could you please recreate a PR setting `v16` as a base branch as I commented on <https://github.com/stylelint/stylelint/issues/6604#issuecomment-1668742296>?

lib/cli.mjs
```diff
@@ -555,11 +555,15 @@ export default async function main(argv) {
 
 	return standalone(options)
 		.then((linted) => {
-			if (!linted.output) {
+			if (!linted.output && !linted.code) {
```
Mouvedia, line 558: question

Are you sure it's not a `||`?

lib/__tests__/getCleanOutput.mjs
haocheng6: Thank you for your suggestion!

I have created a new PR (#7119) based on the `v16` branch and reduced duplicate code by reusing the `getCleanOutput` function (although I still added a new `getCleanFormatterOutput` function to make it easier to provide a default parameter because otherwise the default value needs to be explicitly passed in many test cases.).

PR #7119, opened 2023-08-08
Change CLI and Node.js API to respectively print fixed code and problems to `stdout`/`code` and `stderr`/`output`

### LINKED ISSUE #6604: Fix `--fix` not reporting unfixable errors to stderr

haocheng6:

Closes #6604 

- This PR makes stylelint output fixed code to stdout when `--fix` is used with `--stdin`.
- Currently stylelint outputs linting errors, deprecation warnings, descriptionless disables, etc to stdout. This PR changes that to stderr.
- A new test util function is added and an existing test util function is rewritten to reduce duplicate code based on the [suggestion](https://github.com/stylelint/stylelint/pull/7118#discussion_r1286489633) of @ybiquitous.
- This PR also fixes a bug in `cli.test.mjs`.
  - Previously, assertions like `expect(process.exitCode).toBeUndefined();` might get affected by the test above them because it seems `process.exitCode` is shared by all test cases. You can verify that by moving the `--version` test case to the bottom of the `describe('CLI')` block.
  - The `--quiet` test case is another example of the bug. The actual exit code is `undefined`, but the test expects it to be 2. The real exit code can be verified by `npx stylelint --quiet --config lib/__tests__/fixtures/default-severity-warning.json lib/__tests__/fixtures/empty-block.css` and `echo $?`.
  - This PR adds `process.exitCode = 0;` at the beginning of the `main` function so that tests will not interfere with each other.


lib/testUtils/getCleanOutput.mjs
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	return cleanOutput;
+}
+
+/**
+ *
+ * @param {StylelintResult[]} results
+ * @param {Formatter} formatter
+ * @param {any} returnValue
```
haocheng6, line 32: I use `any` here because the default value `{ ruleMetadata: {} }` in the original `prepareFormatterOutput` is not a valid `LinterResult` (some properties are missing).

Mouvedia: You can use [`Omit`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys) for that.

Mouvedia: I wonder if there's an [`Omit`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys)-like alternative in JSDOC.

ybiquitous: [suggest] Here's my suggestion. How about it?

```js
/**
 * @param {import('stylelint').LintResult[]} results
 * @param {import('stylelint').Formatter} formatter
 * @param {Pick<import('stylelint').LinterResult, 'ruleMetadata'>} [returnValue]
 * @returns {string}
 */
export function getCleanFormatterOutput(results, formatter, returnValue = { ruleMetadata: {} }) {
	// @ts-expect-error -- This error is acceptable because of the test use only.
	return getCleanOutput(formatter(results, returnValue));
}
```

Although the code increases, I believe it is more explicit and will make fewer bugs.

haocheng6: @Mouvedia @ybiquitous Thank you for the suggestion! I didn't realize that I could use `Pick` and `@ts-expect-error` here. Updated here: 9780acb15cdf0f1f11a60475817f1e84ebbbdde2

jeddy3 (commented):
@haocheng6 Thank you for opening the pull request.

I've added one comment. I'll also add a changelog entry and update the title to reflect that this is a breaking change.

lib/cli.mjs
```diff
@@ -383,6 +383,8 @@ const meowOptions = {
  * @returns {Promise<any>}
  */
 export default async function main(argv) {
+	process.exitCode = 0;
```
Mouvedia, line 386: question

Isn't `0` the default?
If so, shouldn't the reset be handled in the tests themselves?

lib/standalone.js
```diff
@@ -138,7 +138,7 @@ async function standalone({
 		const autofix = fix ?? config?.fix ?? false;
 
 		if (autofix && postcssResult && !postcssResult.stylelint.ignored) {
-			returnValue.output =
+			returnValue.code =
```
Mouvedia, line 141: minor

The newly added LinterResult.code should be `string | undefined` but `!postcssResult.stylelint.disableWritingFix` is boolean.

lib/cli.mjs
```diff
@@ -556,11 +558,15 @@ export default async function main(argv) {
...
+			if (!linted.output && !linted.code) {
 				return;
 			}
 
-			process.stdout.write(linted.output);
+			if (linted.code) {
+				process.stdout.write(linted.code);
+			}
+
+			process.stderr.write(linted.output);
```
ybiquitous, line 569: [suggest] How about writing conditionally like this to avoid outputting an empty string?

```suggestion
			if (linted.output) {
				process.stderr.write(linted.output);
			}
```

lib/cli.mjs
```diff
@@ -556,11 +558,15 @@ export default async function main(argv) {
 
 	return standalone(options)
 		.then((linted) => {
-			if (!linted.output) {
+			if (!linted.output && !linted.code) {
```
ybiquitous, line 561: [suggest] How about refactoring the local variables like this to avoid repetition?

```suggestion
			const { output, code } = linted;

			if (!output && !code) {
```

lib/cli.mjs
ybiquitous, line 569: [suggest]

```suggestion
			if (code) {
				process.stdout.write(code);
			}

			process.stderr.write(output);
```

lib/cli.mjs
```diff
@@ -556,11 +558,15 @@ export default async function main(argv) {
...
 
-			process.stdout.write(linted.output);
+			if (linted.code) {
+				process.stdout.write(linted.code);
+			}
+
+			process.stderr.write(linted.output);
 
 			if (options.outputFile) {
 				writeOutputFile(linted.output, options.outputFile).catch(handleError);
```
ybiquitous, line 572: [suggest]

```suggestion
				writeOutputFile(output, options.outputFile).catch(handleError);
```

types/stylelint/index.d.ts
```diff
@@ -375,7 +375,8 @@ declare namespace stylelint {
 		cwd: string;
 		results: LintResult[];
 		errored: boolean;
-		output: any;
+		output: string;
+		code?: string;
```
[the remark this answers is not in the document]
ybiquitous: Good catch! I forgot that description in the doc. 😅

lib/standalone.js
ybiquitous: Um, if `!postcssResult.stylelint.disableWritingFix === false`, then run `returnValue.code = code`, right?

Mouvedia: foo = false && whatever

foo will always be false

Mouvedia: My bad, prettier tricked me.

lib/cli.mjs
haocheng6: Good catch! Done here: 8d7bfdaada3d713dab564add646547e9026901ca.

lib/cli.mjs
haocheng6: @Mouvedia I think not setting `process.exitCode` is effectively the same as setting it to 0. Yeah, I agree resetting `process.exitCode` in tests is better. Let me double check with @ybiquitous on this because I am a bit confused of his comment.

@ybiquitous Could you clarify a bit more the reason of adding `process.exitCode = undefined;` in `afterAll()`? Is the purpose to make sure the final exit code of `npm run test` not affected by our test cases? Or do you mean resetting `process.exitCode` in `afterEach()` rather than `afterAll()`?

I agree with @Mouvedia about resetting `process.exitCode` directly in tests rather than in `cli.mjs`. Here is my plan if that sounds good to you:
- Remove `process.exitCode = EXIT_CODE_SUCCESS;` from the `main` function.
- Add `afterEach(() => { process.exitCode = undefined; });` in `cli.test.mjs`.
- Replace `expect(process.exitCode).toBe(0);` with `expect(process.exitCode).toBeUndefined();` in `cli.test.mjs`.

@ybiquitous what do you think?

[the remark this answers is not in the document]
haocheng6: Thanks for the suggestion! Done here: 73f66818a540d9a06dc58bb53d76d22a937454c6

I am not sure if we want to keep this line, because as @Mouvedia suggested above, we could also resetting `process.exitCode` in tests. Please see my [comment](https://github.com/stylelint/stylelint/pull/7119#discussion_r1289122158) above and let me know what you think. Thanks!

Mouvedia: 👍 for `afterEach`

types/stylelint/index.d.ts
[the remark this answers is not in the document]
haocheng6: Thanks for bringing up the documentation issue. I have updated some docs in 4de6a5f8fbaae570d0a50009cfe85010daa4ab96.

lib/cli.mjs
ybiquitous: Ah, sorry. `afterEach` is better 👍🏼

docs/user-guide/node-api.md
```diff
@@ -46,6 +46,10 @@ For more detail usage, see [Globby Guide](https://github.com/sindresorhus/globby
 
 `stylelint.lint()` returns a Promise that resolves with an object containing the following properties:
 
+### `code`
+
+A string that contains the autofixed code, if the `fix` option is set to `true` and the `code` option is provided. Otherwise, it is `undefined`.
```
Mouvedia, line 51: question

If the rule is not fixable—e.g. `color-no-invalid-hex`—Id expect to get the same code back and not `undefined`.
Am I wrong?

jeddy3: I believe it refers to the `fix` option rather than the fixability of individual rules. @haocheng6 Is that right? 

If so, the same code will return in `code` if the `fix` option is used on a config containing only unfixable rules like `color-no-invalid-hex`. 

`undefined` is used when the `fix` option isn't set, which feels intuitive as Stylelint will not attempt to transform the inputted code.

docs/migration-guide/to-16.md
```diff
@@ -12,3 +12,20 @@ Node.js 14 has reached end-of-life. We've removed support for it so that we coul
 
 - 16.13.0
 - 18.0.0
+
+## Changed CLI behavior
+
+Starting from 16.0.0, the CLI prints problems to stderr instead of stdout.
```
jeddy3, line 18:
```suggestion
## Changed CLI to print problems to stderr instead of stdout
```

docs/migration-guide/to-16.md
```diff
@@ -12,3 +12,20 @@ Node.js 14 has reached end-of-life. We've removed support for it so that we coul
...
+
+If you use the CLI to fix a source string by using the [`--fix`](../user-guide/cli.md#--fix) and [`--stdin`](../user-guide/cli.md#--stdin) options, the CLI will print the fixed code to stdout and any problems to stderr.
+
+## Changed Node.js API behavior
+
+The Promise returned by `stylelint.lint()` has the following changes:
+
+- The [output](../user-guide/node-api.md#output) property of the resolved object now only contains formatted problems.
+- A new [code](../user-guide/node-api.md#code-1) property which contains the autofixed code has been added to the resolved object.
```
jeddy3, line 27:
```suggestion
## Changed Node.js API returned resolved object

We changed the resolved object of the Promise returned by `stylelint.lint()` so that:

- the [output](../user-guide/node-api.md#output) property now only contains formatted problems
- a new [code](../user-guide/node-api.md#code-1) property contains the autofixed code
```

jeddy3 (changes requested):
Thank you for adding the docs!

I've requested minor changes to align with our house style, e.g. scannable headings.

lib/cli.mjs
haocheng6: Thanks! I have updated the CLI code and the corresponding tests in fa98abe77234da7e949c2351cb80611af4c8a408.

docs/user-guide/node-api.md
haocheng6: @jeddy3 In this doc, I think the `fix` option refers to the `fix` option for the Node.js API or the `fix` option in the config file.

@Mouvedia @jeddy3 Both of you are correct. I have added two new test cases in 5b1c778686959eb929807d85628ac8a73f7f7235. Hopefully, the new test cases can help us ensure the `standalone` function is working as expected.

lib/__tests__/standalone-fix.test.mjs
```diff
@@ -289,8 +340,60 @@ it('one rule being disabled and another still autofixing', async () => {
...
+			},
+		},
+		fix: false,
+		formatter: 'string',
+	});
+
+	expect(result.code).toBeUndefined();
+	expect(getCleanOutput(result.output)).toBe(stripIndent`
+		test.css
+		 1:12  ×  Expected \"#ffffff\" to be \"#fff\"  color-hex-length
```
Mouvedia, line 371: question

Is escaping double quotes necessary?

haocheng6: Good catch! I copied the text from VS Code terminal and maybe the terminal added those backslashes for me. 😂 

Updated here: 0f5f531ffe54bba6dfb44ec0e62ab99538b26b83

ybiquitous (approved):
Thanks a lot! LGTM 👍🏼

jeddy3 (approved):
LGTMT, thank you!

lib/__tests__/cli.test.mjs
```diff
@@ -352,9 +362,10 @@ describe('CLI', () => {
 			fixturesPath('empty-block.css'),
 		]);
 
-		expect(process.exitCode).toBe(2);
+		expect(process.exitCode).toBeUndefined();
```
Mouvedia, line 365: question

Are you sure it's what was intended?
@hudochenkov do you confirm?
cf https://github.com/stylelint/stylelint/pull/5542/files#diff-0378929f7797bb789a616f16b2a0f5d0e53a28eba60469829002b0b30d60886cR324

hudochenkov: I guess exit code should be 0. Because warnings are ignored and there is no other violations.

haocheng6: The actual exit code is 0 if we run the CLI in a terminal. I did some testing on my machine to verify that.

On the latest `v16` branch:

```console
root@LAPTOP-7C2UU06T:~/my-fork/stylelint# git branch --show-current
v16
root@LAPTOP-7C2UU06T:~/my-fork/stylelint# git rev-parse HEAD
9827e0d9cef369b8af097e97ccbdc32c8b454fd3
root@LAPTOP-7C2UU06T:~/my-fork/stylelint# node bin/stylelint.mjs --quiet --config lib/__tests__/fixtures/default-severity-warning.json lib/__tests__/fixtures/empty-block.css
root@LAPTOP-7C2UU06T:~/my-fork/stylelint# echo $?
0
```

On the `6604-code-stdout-error-stderr-v16` branch (the branch of this PR):

```console
root@LAPTOP-7C2UU06T:~/my-fork/stylelint# git branch --show-current
6604-code-stdout-error-stderr-v16
root@LAPTOP-7C2UU06T:~/my-fork/stylelint# git rev-parse HEAD
0f5f531ffe54bba6dfb44ec0e62ab99538b26b83
root@LAPTOP-7C2UU06T:~/my-fork/stylelint# node bin/stylelint.mjs --quiet --config lib/__tests__/fixtures/default-severity-warning.json lib/__tests__/fixtures/empty-block.css
root@LAPTOP-7C2UU06T:~/my-fork/stylelint# echo $?
0
```

Using `stylelint@latest`:

```console
root@LAPTOP-7C2UU06T:~/my-fork/stylelint# npx stylelint@latest --quiet --config lib/__tests__/fixtures/default-severity-warning.json lib/__tests__/fixtures/empty-block.css
root@LAPTOP-7C2UU06T:~/my-fork/stylelint# echo $?
0
```

In this test, `process.exitCode` is expected to be `undefined` because that is the default value of `process.exitCode` in Node.js ([Reference](https://nodejs.org/api/process.html#processexitcode_1)) and we also explicitly reset `process.exitCode` to `undefined` in the `afterEach` hook.

ybiquitous:
Oops, my merging another PR causes conflicts. 😖 
I will try fixing them.

ybiquitous:
> I will try fixing them.

Done! 72f0dd7d7

ybiquitous:
I will merge this after all CI jobs are green.

PR #7020, opened 2023-06-30
Remove Node.js 14 support

ybiquitous:
> Which issue, if any, is this issue related to?

Closes #6740
Closes #6785
(Unfortunately, the PR's author doesn't respond)

> Is there anything in the PR that needs further explanation?

Node.js **16.13.0** is the first version of Node.js 16.x LTS series.
See https://nodejs.org/en/blog/release/v16.13.0

---

Here's a [`ls-engines`](https://www.npmjs.com/package/ls-engines) report:

```console
$ npx ls-engines@latest
`node_modules` found; loading tree from disk...
┌────────┬─────────────────────────────────────────────────────────────────┐
│ engine │ Currently available latest release of each valid major version: │
├────────┼─────────────────────────────────────────────────────────────────┤
│ node   │ v20.3.1, v19.9.0, v18.16.1, v17.9.1, v16.20.1                   │
└────────┴─────────────────────────────────────────────────────────────────┘

┌───────────────────────────────────┬───────────────────────────┐
│ package engines:                  │ dependency graph engines: │
├───────────────────────────────────┼───────────────────────────┤
│ "engines": {                      │ "engines": {              │
│   "node": "^16.13.0 || >= 18.0.0" │   "node": ">= 16.13"      │
│ }                                 │ }                         │
└───────────────────────────────────┴───────────────────────────┘


Your “engines” field allows fewer node versions than your dependency graph does.

If you want to widen your support, you can run `ls-engines --save`, or manually add the following to your `package.json`:
"engines": {
  "node": ">= 16.13"
}

┌────────┬─────────────────┬─────────────────┬──────────────────────────┐
│ engine │ current version │ valid (package) │ valid (dependency graph) │
├────────┼─────────────────┼─────────────────┼──────────────────────────┤
│ node   │ v20.3.0         │ yes!            │ yes!                     │
└────────┴─────────────────┴─────────────────┴──────────────────────────┘
```

PR #7121, opened 2023-08-09
Update `tsconfig.json` for ES2021

ybiquitous:

None.

Node.js 16 is 100% compatible with ES2021.
See https://node.green/#ES2021

<details>
<summary>Screenshot</summary>

image

</details>

In addition, this change uses [`String.prototype.replaceAll()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) as possible for readability. This method has been available since ES2021.

---

Context: Stylelint 16.0.0 will support Node.js 16 or later. See #7020.


jeddy3 (approved):
LGTM, thank you.

PR #7122, opened 2023-08-09
Migrate `lib/reference/*` to ESM

ybiquitous:

Ref #5291

- Rewrite `lib/reference/*.js` to `*.mjs` in ESM.
- Generate `lib/reference/*.cjs` via `npm run build`.


jeddy3 (approved):
LGTM, thank you for moving the migration to ESM forward!

PR #7124, opened 2023-08-10
Use `invalidConfig` property of `jest-preset-stylelint` new feature

ybiquitous:

Ref https://github.com/stylelint/jest-preset-stylelint/pull/68

Note that this PR uses the unpublished version of `jest-preset-stylelint`.


jeddy3 (approved):
LGTM, thank you.

Is the intent to add this to other rules or keep it in our back pocket to use when fixing a bug with the option validation of a rule?

PR #7127, opened 2023-08-11
Migrate `lib/formatters/*` to ESM except `formatters/index`

ybiquitous:

Ref #5291

`formatters/index` internally uses `lazy-import`. This is a bit hard to resolve, so I will migrate it later (with another PR).

```console
$ git ls-files 'lib/formatters/*.js'
lib/formatters/index.js
```

### DIFF
--- a/lib/formatters/terminalLink.mjs
+++ b/lib/formatters/terminalLink.mjs
@@ -1,6 +1,4 @@
-'use strict';
-
-const supportsHyperlinks = require('supports-hyperlinks');
+import supportsHyperlinks from 'supports-hyperlinks';
 
 // ANSI escapes
 const OSC = '\u001B]';
@@ -14,10 +12,10 @@ const SEP = ';';
  * @param {string} url
  * @returns {string}
  */
-module.exports = function terminalLink(text, url) {
+export default function terminalLink(text, url) {
 	if (supportsHyperlinks.stdout) {
 		return [OSC, '8', SEP, SEP, url, BEL, text, OSC, '8', SEP, SEP, BEL].join('');
 	}
 
 	return text;
-};
+}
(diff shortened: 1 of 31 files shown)


jeddy3 (approved):
LGTM, thank you!

lib/formatters/terminalLink.mjs
```diff
@@ -0,0 +1,21 @@
...
+ * @param {string} url
+ * @returns {string}
+ */
+export default function terminalLink(text, url) {
+	if (supportsHyperlinks.stdout) {
+		return [OSC, '8', SEP, SEP, url, BEL, text, OSC, '8', SEP, SEP, BEL].join('');
+	}
+
+	return text;
+}
```
Mouvedia, line 21: question

Since you don't have an index.mjs for formatters, this file is only used by terminalLink.test.mjs?

ybiquitous: No. `terminalLink.mjs` is also used by `stringFormatter.mjs` and `verboseFormatter.mjs`.

https://github.com/stylelint/stylelint/blob/b02c39aaef66c49d93263433be8ef5508da7f0ec/lib/formatters/stringFormatter.mjs#L11

https://github.com/stylelint/stylelint/blob/b02c39aaef66c49d93263433be8ef5508da7f0ec/lib/formatters/verboseFormatter.mjs#L6

Mouvedia: …and stringFormatter.mjs and verboseFormatter.mjs are not used in formatters.index.mjs—which doesn't exist yet—, hence it's only actually used in terminalLink.test.mjs, correct?

ybiquitous: Ah, yes. I will create `formatters/index.mjs` later, but the job seems difficult since `import-lazy` is related.

PR #7129, opened 2023-08-13
Refactor `lib/cli.mjs`

ybiquitous:

None.

- Remove no longer needed JSDoc types.
- Use type-guard functions for type-safety.
- Make the code more readable with new local variables.

If anyone knows a way to use type inference by `meow()` in this `cli.mjs` module, please let me know.

### DIFF
--- a/lib/cli.mjs
+++ b/lib/cli.mjs
@@ -1,5 +1,6 @@
+import { isAbsolute, join, resolve } from 'node:path';
 import { EOL } from 'node:os';
-import path from 'node:path';
+import process from 'node:process';
 
 import picocolors from 'picocolors';
 const { dim, red } = picocolors;
@@ -7,8 +8,8 @@ const { dim, red } = picocolors;
 import meow from 'meow';
 import resolveFrom from 'resolve-from';
 
+import { isBoolean, isNumber, isObject, isPlainObject, isString } from './utils/validateTypes.mjs';
 import checkInvalidCLIOptions from './utils/checkInvalidCLIOptions.mjs';
-import { isPlainObject } from './utils/validateTypes.mjs';
 import printConfig from './printConfig.js';
 import resolveCustomFormatter from './resolveCustomFormatter.js';
 import standalone from './standalone.js';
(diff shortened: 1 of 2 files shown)


jeddy3 (approved):
Great refactor, thank you!

lib/cli.mjs
```diff
@@ -452,114 +424,87 @@ export default async function main(argv) {
...
 			})
 			.catch(handleError);
+
+		return;
 	}
 
-	if (!options.files && !options.code && !cli.flags.stdin) {
-		cli.showHelp(EXIT_CODE_SUCCESS);
+	if (!options.files && !options.code && !stdin) {
+		showHelp();
```
Mouvedia, line 501: So the default is `===` to `EXIT_CODE_SUCCESS`?

lib/cli.mjs
```diff
@@ -572,14 +517,14 @@ export default async function main(argv) {
...
+			if (isString(outputFile)) {
+				writeOutputFile(output, outputFile).catch(handleError);
 			}
 
-			if (linted.errored) {
+			if (errored) {
 				process.exitCode = EXIT_CODE_ERROR;
-			} else if (maxWarnings !== undefined && linted.maxWarningsExceeded) {
-				const foundWarnings = linted.maxWarningsExceeded.foundWarnings;
+			} else if (isNumber(maxWarnings) && maxWarningsExceeded) {
```
Mouvedia, line 526: nit

Since you were using predicates everywhere else, I was expecting `isPlainObject(maxWarningsExceeded)`.

lib/cli.mjs
Mouvedia: question

@ybiquitous Do we have a regression ?
If so do we need an additional test?
cf #5150

ybiquitous: I didn't change the behavior. `showHelp()` is a local helper and exits with `0`. See https://github.com/stylelint/stylelint/pull/7129/files#diff-d1470268bb8c3a42692b2c5249d7d31832ceeed1b1f16fbdbc814f8013c36b31R357

lib/cli.mjs
ybiquitous: This `isNumber(maxWarnings)` is needed for [narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html) the `maxWarnings` type from `unknown` to `number`. On the other hand, `maxWarningsExceeded` from `standalone()` already has the concrete type, so `isPlainObject()` is unnecessary.

PR #7131, opened 2023-08-14
Fix CLI to avoid waiting for stdin without any input and avoid different outputs

### LINKED ISSUE #3401: Fix different outputs on empty files and empty stdin in CLI
> Clearly describe the bug

For an empty file `empty.css`, `stylelint empty.css` gives normal results, while `stylelint < empty.css` shows the help message with exit code 2.

The latter is how the Emacs plugin flycheck uses stylelint. So this makes flycheck confused when we open an empty file (flycheck/flycheck#1458).

No.

An empty file.

I am using `stylelint-config-recommended`.

`9.3.0`.

CLI with `stylelint < empty.css`.

No.

`stylelint < empty.css` should give the same output as `stylelint empty.css`.

`stylelint empty.css` gives (with `stylelint-config-recommended`)

```
empty.css
1:1   Unexpected empty source   no-empty-source
```

`stylelint < empty.css` shows the help message

```
A mighty, modern CSS linter.

Usage: stylelint [input] [options]

...[truncated]
```

Thanks!

ybiquitous:

Closes #3401

This commit changes the following:

1. To avoid waiting for stdin without any input. For example, just running the `stylelint` command
   immediately exits, instead of waiting. I believe this behavior is more user-friendly.
2. To avoid different outputs on empty files and empty stdin. This is achieved by allowing an empty
   string input (`''`) from stdin. See the example below:
   ```sh
   # code is undefined
   bin/stylelint.mjs

   # code is ''
   echo -n '' | bin/stylelint.mjs
   ```

See also [`process.stdin.isTTY`](https://nodejs.org/api/tty.html#readstreamistty).

### DIFF
--- a/.changeset/orange-dancers-compete.md
+++ b/.changeset/orange-dancers-compete.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: CLI to avoid waiting for stdin without any input (regression since 14.10.0)
+Fixed: CLI regression to avoid waiting for stdin without any input
--- a/.changeset/orange-dancers-compete.md
+++ b/.changeset/orange-dancers-compete.md
@@ -1,5 +1,5 @@
 ---
-"stylelint": minor
+"stylelint": patch
 ---
 
-Changed: CLI to avoid waiting for stdin without any input
+Fixed: CLI to avoid waiting for stdin without any input (regression since 14.10.0)
--- a/.changeset/orange-dancers-compete.md
+++ b/.changeset/orange-dancers-compete.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Changed: CLI to avoid waiting for stdin without any input
--- a/lib/__tests__/cli.test.mjs
+++ b/lib/__tests__/cli.test.mjs
@@ -183,24 +183,22 @@ describe('buildCLI', () => {
 });
 
 describe('CLI', () => {
-	beforeAll(() => {
+	beforeEach(() => {
 		jest.spyOn(process, 'exit').mockImplementation(() => {});
 		jest.spyOn(process.stdout, 'write').mockImplementation(() => {});
 		jest.spyOn(process.stderr, 'write').mockImplementation(() => {});
-		jest.spyOn(process, 'stdin', 'get').mockImplementation(() => Readable.from([Buffer.from('')]));
 		jest.spyOn(console, 'log').mockImplementation(() => {});
 		jest.spyOn(console, 'error').mockImplementation(() => {});
 	});
 
 	afterEach(() => {
+		jest.restoreAllMocks();
 		process.exitCode = undefined;
 	});
 
-	afterAll(() => {
-		jest.restoreAllMocks();
-	});
+	it('no arguments', async () => {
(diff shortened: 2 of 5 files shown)


lib/__tests__/cli.test.mjs
```diff
@@ -183,24 +183,22 @@ describe('buildCLI', () => {
 });
 
 describe('CLI', () => {
-	beforeAll(() => {
+	beforeEach(() => {
```
ybiquitous, line 186: [note] To avoid effects on each `it`.

lib/__tests__/cli.test.mjs
```diff
@@ -365,8 +390,8 @@ describe('CLI', () => {
 
 		expect(process.exitCode).toBeUndefined();
 
-		expect(process.stdout.write).toHaveBeenCalledTimes(0);
-		expect(process.stderr.write).toHaveBeenCalledTimes(0);
+		expect(process.stdout.write).not.toHaveBeenCalled();
+		expect(process.stderr.write).not.toHaveBeenCalled();
```
ybiquitous, line 394: [note] To clarify the difference from `toHaveBeenCalledWith(0)`.

jeddy3 (commented):
Great to see consistency for empty files and std.

> Changed: CLI to avoid waiting for stdin without any input

We used to show help on `npx stylelint`. Did we intentionally change it to wait for stdin? If not, then it's a regression, and we can change the entry to patch/fixed and drop the migration guide entry.

ybiquitous:
> We used to show help on `npx stylelint`. Did we intentionally change it to wait for stdin?

I found out the reason. This bug was mixed in when I removed the `get-stdin` dependency with PR #6182. 😓 

I'll update the changelog and migration guide. Thanks for pointing it out!

ybiquitous:
FYI. This regression happened with `stylelint@14.10.0`. See below:

```console
$ npm i -s 'stylelint@14.10.0' && npx stylelint
# waiting for stdin...
```

```console
$ npm i -s 'stylelint@14.9.1' && npx stylelint | grep 'Usage:'
  Usage: stylelint [input] [options]
```

ybiquitous:
> I'll update the changelog and migration guide. Thanks for pointing it out!

Fixed via 88bb9b4d3.

.changeset/orange-dancers-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: CLI to avoid waiting for stdin without any input (regression since 14.10.0)
```
jeddy3, line 5:
```suggestion
Fixed: CLI regression to avoid waiting for stdin without any input
```

Avoids brackets, as we typically only use them for contributor and pull requests.

jeddy3 (approved):
One minor suggestion for the changelog entry. Otherwise, LGTM thank you!

lib/cli.mjs
```diff
@@ -580,12 +580,18 @@ function parseGlobbyOptions(value) {
 }
 
 /**
- * @returns {Promise<string>}
+ * @returns {Promise<string | undefined>}
  */
 async function getStdin() {
+	const { stdin } = process;
+
+	if (stdin.isTTY) {
```
Mouvedia, line 588: So I was right, it was the right way.
Thanks @ybiquitous.

PR #7135, opened 2023-08-15
Migrate `lib/formatters/index` to ESM

ybiquitous:

Ref #5291 #7127

Note that this change still keeps using `import-lazy` to minimize this changeset.


jeddy3 (approved):
LGTM, thank you!

> Note that this change still keeps using `import-lazy` to minimize this changeset.

I agree it's a viable stepping stone towards dynamic imports.

PR #7139, opened 2023-08-16
Reconfigure ESLint with bumping eslint-config-stylelint to 20.0.0

ybiquitous:

None.

Summary:

- Remove redundant `globals`.
- Make `testRule*` globals [readonly](https://eslint.org/docs/latest/use/configure/language-options#using-configuration-files-1).
- Turn on `n/prefer-global/process` to disallow global `process` for consistency.
  See https://github.com/eslint-community/eslint-plugin-n/blob/v11.1.0/docs/rules/prefer-global/process.md

Note:

Global variables are often harmful. Unlike global variables defined in standard specs (ECMAScript or WHATWG etc.),
`process` is unique to Node.js. This may be a problem when running Stylelint on other platforms.

So, I prefer using `process` through the module system. But this is just my opinion. If anyone is against it, I'll withdraw this change.


Mouvedia:
To be clear you just want to use `import process from 'node:process';`?

jeddy3 (commented):
Having an explicit import SGTM.

ybiquitous:
> To be clear, you just have to add `import process from 'node:process';`?

Yes, since there seem to be no objections, I'll add `import`. 👍🏼

ybiquitous:
Fixed lint errors via 4aa84e418

ybiquitous:
I've also opened upstream PR https://github.com/stylelint/eslint-config-stylelint/pull/233

jeddy3 (approved):
LGTM, thank you.

ybiquitous:
> I'll update the shared config in this PR. 💪🏼

Done 141de20.

PR #7146, opened 2023-08-21
Fix test expectations after PostCSS update

romainmenke:

https://github.com/stylelint/stylelint/pull/7143

With the latest PostCSS version `Root.source.end` is also present.
This was previously missing.

Any warning or error associated with a `Root` node will now span the entire contents, not just the first line and column.

_this pr will merge into `dependabot/npm_and_yarn/postcss-8.4.28`_


jeddy3 (approved):
Thank you!

PR #7152, opened 2023-08-27
Migrate `lib/postcssPlugin` to ESM

ybiquitous:

Ref #5291

No, it's self-explanatory.


lib/postcssPlugin.mjs
```diff
@@ -0,0 +1,51 @@
...
+/** @typedef {import('stylelint').PostcssPluginOptions} PostcssPluginOptions */
+/** @typedef {import('stylelint').Config} StylelintConfig */
+
+/**
+ * @type {import('postcss').PluginCreator<PostcssPluginOptions>}
+ */
+export default function postcssPlugin(options = {}) {
+	const [cwd, tailoredOptions] = isConfig(options)
+		? [process.cwd(), { config: options }]
+		: [('cwd' in options && isString(options.cwd) && options.cwd) || process.cwd(), options];
```
ybiquitous, line 17: [note] Unless `'cwd' in options`, the following TS error is raised: ⬇️ 

```
TS2339: Property 'cwd' does not exist on type '{}'.
```

lib/postcssPlugin.mjs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+postcssPlugin.postcss = true;
+
+/**
+ * @param {PostcssPluginOptions} options
+ * @returns {options is StylelintConfig}
+ */
+function isConfig(options) {
+	return 'rules' in options;
```
Mouvedia, line 50: nit

Maybe `hasRules` would be better?
I mean Id like to be able to detect that something is wrong with that assumption at the call site without having to check the implementation.
Or simply inline it at line 15 and remove the 2 typedefs?

ybiquitous: Good point. I've refactored the code via 9fcd17380, but it's still insufficient.

First of all, renaming `isConfig` to `hasRules` or inlining `isConfig` cannot improve the code. The root cause is the complexity of the `PostcssPluginOptions` type as the argument type of `postcssPlugin()`:

https://github.com/stylelint/stylelint/blob/b509d18ff155d07a3c839639a15040654e47d9d7/types/stylelint/index.d.ts#L574

---

https://github.com/stylelint/stylelint/blob/b509d18ff155d07a3c839639a15040654e47d9d7/docs/user-guide/postcss-plugin.md?plain=1#L9

As our doc above says, I think `postcssPlugin()`'s argument type should be just the following: ⬇️ 

```typescript
type PostcssPluginOptions = Omit<LinterOptions, 'customSyntax'>;
```

But this change should be done with another PR rather than this one.

Mouvedia: It's the boyscout rule.

If you were just renaming from js to cjs I wouldn't have said anything.
But I think the new mjs file warrants a real review as if it was a new file and not just a conversion verbatim.
If it slows you down then you could postpone these changes to a later PR.
The ternary is not unreadable but it could help to add some clarity.

ybiquitous: > If you were just renaming from js to cjs I wouldn't have said anything.

Please see https://github.com/stylelint/stylelint/pull/7152#discussion_r1306609801
I need more changes to pass type-check.

ybiquitous:
@Mouvedia If you have no other concerns, I'd like to merge this to continue the migration. No problem?

Mouvedia (approved):

ybiquitous:
Thanks!

PR #7158, opened 2023-08-27
Migrate `lib/standalone` to ESM

ybiquitous:

Ref #5291

No, it's self-explanatory.

### DIFF
--- a/lib/standalone.cjs
+++ b/lib/standalone.cjs
@@ -1,0 +1,325 @@
+'use strict';
+
+const node_path = require('node:path');
+const node_fs = require('node:fs');
+const process = require('node:process');
+const createDebug = require('debug');
+const fastGlob = require('fast-glob');
+const globby = require('globby');
+const normalizePath = require('normalize-path');
+const writeFileAtomic = require('write-file-atomic');
+const allFilesIgnoredError = require('./utils/allFilesIgnoredError.cjs');
+const noFilesFoundError = require('./utils/noFilesFoundError.cjs');
+const validateTypes = require('./utils/validateTypes.cjs');
+const createPartialStylelintResult = require('./createPartialStylelintResult.cjs');
+const createStylelint = require('./createStylelint.cjs');
+const filterFilePaths = require('./utils/filterFilePaths.cjs');
+const index = require('./formatters/index.cjs');
+const getConfigForFile = require('./getConfigForFile.cjs');
+const getFileIgnorer = require('./utils/getFileIgnorer.cjs');
+const getFormatterOptionsText = require('./utils/getFormatterOptionsText.cjs');
+const lintSource = require('./lintSource.cjs');
+const prepareReturnValue = require('./prepareReturnValue.cjs');
+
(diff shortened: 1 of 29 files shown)


lib/standalone.cjs
```diff
@@ -1,29 +1,29 @@
...
-const createStylelint = require('./createStylelint.cjs');
+const writeFileAtomic = require('write-file-atomic');
+const allFilesIgnoredError = require('./utils/allFilesIgnoredError.cjs');
+const noFilesFoundError = require('./utils/noFilesFoundError.cjs');
+const validateTypes = require('./utils/validateTypes.cjs');
 const createPartialStylelintResult = require('./createPartialStylelintResult.cjs');
+const createStylelint = require('./createStylelint.cjs');
 const filterFilePaths = require('./utils/filterFilePaths.cjs');
-const formatters = require('./formatters/index.cjs');
+const index = require('./formatters/index.cjs');
```
Mouvedia, line 17: nit

Why the rename?

lib/standalone.mjs
```diff
@@ -0,0 +1,323 @@
...
+
+/**
+ * @typedef {import('stylelint').CssSyntaxError} CssSyntaxError
+ *
+ * @param {unknown} error
+ * @return {import('stylelint').LintResult}
+ */
+function handleError(error) {
+	if (error instanceof Error && error.name === 'CssSyntaxError') {
+		return createPartialStylelintResult(undefined, /** @type {CssSyntaxError} */ (error));
```
Mouvedia, line 309: off topic

> `'createPartialStylelintResult must be called with either postcssResult or CssSyntaxError',`

I think this function should be refactored.
The fact that both arguments are optional is a red flag.
Maybe differentiate based on the content of a single object being passed?
Looks like a code smell to me.

lib/standalone.cjs
ybiquitous: `.cjs` files are automatically generated by Rollup. So I don't know why. 😅

Mouvedia: I mean the const name.
Why rename it index when you kept formatters in the mjs counterpart?

ybiquitous: I don't touch the const name. Rollup determines the name from the `.mjs` file, but I don't know why Rollup renamed it.

lib/standalone.mjs
ybiquitous: I agree that this code needs refactoring, although the function is old. Can you please open an issue or a pull request? I hope it will be after merging `v16` into `main`.

lib/standalone.cjs
mattxwang: Just to chime in - I think it's because the file is named `index`?

mattxwang (approved):
Getting back into the swing of things! On my end, this LGTM.

(I'm not too concerned if we don't resolve the `index` naming, since we're not exporting it anyways)

ybiquitous:
Thanks for the review, guys!

Issue #6258, opened 2022-08-12
Use `exports` field in `package.json` to forbid unexpected API usage


JounQin:
`stylelint/lib/utils/*` seems not a public API at all. Should we use `exports` field in `package.json` to forbid this usage on `v15`? cc @ybiquitous

_Originally posted by @JounQin in https://github.com/stylelint/stylelint/issues/6256#issuecomment-1213537779_

ybiquitous:
Help from @regseb on <https://github.com/stylelint/stylelint/pull/7159/files#r1318318623>.

```json
{
  "exports": {
    ".": {
      "import": "./lib/index.mjs",
      "require": "./lib/index.cjs"
    },
    "./package.json": "./package.json"
  }
}
```

Exporting also `package.json` may be better. See https://github.com/depcheck/depcheck/issues/754#issuecomment-1582636531

Mouvedia:
https://github.com/stylelint/stylelint/commit/310c9ad39add73c7d9af0f5c6d0a1832a272aa4d#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R31

PR #7159, opened 2023-08-27
Migrate `lib/index` to ESM

ybiquitous:

Ref #5291

This PR needs to wait for the merge of:
- [x] #7152
- [x] #7158

### DIFF
--- a/lib/__tests__/integration.test.mjs
+++ b/lib/__tests__/integration.test.mjs
@@ -7,7 +7,7 @@ import sassSyntax from 'postcss-sass';
 import scssSyntax from 'postcss-scss';
 import { stripIndent } from 'common-tags';
 
-import stylelint from 'stylelint';
+import stylelint from '../index.mjs';
 
 const fixtures = (...elem) =>
 	path.join(fileURLToPath(new URL('./fixtures', import.meta.url)), ...elem);
--- a/lib/__tests__/integration.test.mjs
+++ b/lib/__tests__/integration.test.mjs
@@ -7,7 +7,7 @@ import sassSyntax from 'postcss-sass';
 import scssSyntax from 'postcss-scss';
 import { stripIndent } from 'common-tags';
 
-import stylelint from '../index.js';
+import stylelint from 'stylelint';
 
 const fixtures = (...elem) =>
 	path.join(fileURLToPath(new URL('./fixtures', import.meta.url)), ...elem);
--- a/lib/index.cjs
+++ b/lib/index.cjs
@@ -3,21 +3,21 @@
 const checkAgainstRule = require('./utils/checkAgainstRule.cjs');
 const createPlugin = require('./createPlugin.cjs');
 const createStylelint = require('./createStylelint.cjs');
-const formatters = require('./formatters/index.cjs');
-const postcssPlugin = require('./postcssPlugin.js');
+const index$1 = require('./formatters/index.cjs');
+const properties = require('./reference/properties.cjs');
+const postcssPlugin = require('./postcssPlugin.cjs');
 const report = require('./utils/report.cjs');
 const resolveConfig = require('./resolveConfig.cjs');
 const ruleMessages = require('./utils/ruleMessages.cjs');
-const rules = require('./rules/index.cjs');
-const { longhandSubPropertiesOfShorthandProperties } = require('./reference/properties.cjs');
-const standalone = require('./standalone');
+const index = require('./rules/index.cjs');
+const standalone = require('./standalone.cjs');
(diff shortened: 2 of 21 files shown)


lib/__tests__/integration.test.mjs
```diff
@@ -7,7 +7,7 @@ import sassSyntax from 'postcss-sass';
 import scssSyntax from 'postcss-scss';
 import { stripIndent } from 'common-tags';
 
-import stylelint from '../index.js';
+import stylelint from 'stylelint';
```
Mouvedia, line 10: This uses https://github.com/stylelint/stylelint/pull/7159/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R124, which is a commonjs file, correct?

ybiquitous: Good catch! I missed `lib/index.cjs` (commonjs) was used.
Fixed via 3bf3fc030. 👍🏼

package.json
```diff
@@ -22,6 +22,10 @@
   },
   "license": "MIT",
   "author": "stylelint",
+  "exports": {
+    "import": "./lib/index.mjs",
+    "require": "./lib/index.cjs"
+  },
```
regseb, line 28: Export also `package.json`:

```suggestion
  "exports": {
    ".": {
      "import": "./lib/index.mjs",
      "require": "./lib/index.cjs"
    },
    "./package.json": "./package.json"
  },
```

I had a problem with a package that didn't export `package.json`: https://github.com/depcheck/depcheck/issues/754#issuecomment-1582636531

I haven't found any official documentation about this, but other packages export `package.json`: [yargs](https://github.com/yargs/yargs/pull/1818), [typedoc](https://github.com/TypeStrong/typedoc/commit/e021751e5fb599b944994e6c2225349a88b51caa#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR18), [eslint](https://github.com/eslint/eslint/blob/926a286/package.json#L11)...

Mouvedia: This is already covered by #6258.
i.e. these migration PRs have to avoid introducing regressions

ybiquitous: Right. I've reverted the change via 80a1ef014.

ybiquitous: @regseb Thanks for the info. Commented on https://github.com/stylelint/stylelint/issues/6258#issuecomment-1710035791

ybiquitous:
This is now ready for review. 👌🏼

lib/index.mjs
```diff
@@ -0,0 +1,33 @@
...
+import createStylelint from './createStylelint.mjs';
+import formatters from './formatters/index.mjs';
+import { longhandSubPropertiesOfShorthandProperties } from './reference/properties.mjs';
+import postcssPlugin from './postcssPlugin.mjs';
+import report from './utils/report.mjs';
+import resolveConfig from './resolveConfig.mjs';
+import ruleMessages from './utils/ruleMessages.mjs';
+import rules from './rules/index.mjs';
+import standalone from './standalone.mjs';
+import validateOptions from './utils/validateOptions.mjs';
```
Mouvedia, line 12: minor

Maybe having a `./utils/index.mjs` would be useful so that we could have one import for utils instead of 4.

ybiquitous: I have no strong objection, but `utils/index.mjs` (or `utils.mjs`) would bring two patterns into the codebase, for example:

```js
import ruleMessages from "./utils/ruleMessages.mjs"
import validateOptions from "./utils/validateOptions.mjs"

// vs

import { ruleMessages, validateOptions } from "./utils/index.mjs"
```

This might spoil consistency. And we would need to maintain the new file (`utils/index.mjs`) additionally.

If we'd perform the refactoring, I recommend it after this migration, at least.

mattxwang (approved):
LGTM! Thanks for your continued work on this!

lib/index.cjs
```diff
@@ -3,21 +3,21 @@
 const checkAgainstRule = require('./utils/checkAgainstRule.cjs');
 const createPlugin = require('./createPlugin.cjs');
 const createStylelint = require('./createStylelint.cjs');
-const formatters = require('./formatters/index.cjs');
+const index$1 = require('./formatters/index.cjs');
```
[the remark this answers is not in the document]
ybiquitous: I searched for an option you thought, but unfortunately, I cannot find it. (Please let me know if you find it instead 🙏🏼 )
https://rollupjs.org/configuration-options/

The simplest way could be to rename from `formatters/index.mjs` to `formatters.mjs`, but I'd like to avoid doing it for now so that we could easily trace commit history.

PR #7182, opened 2023-09-11
Migrate `alpha-value-notation` rule to ESM

ybiquitous:

Ref #5291

No, it's self-explanatory.

### DIFF
--- a/lib/rules/index.mjs
+++ b/lib/rules/index.mjs
@@ -6,7 +6,7 @@ import importLazy from 'import-lazy';
 
 /** @type {import('stylelint')['rules']} */
 const rules = {
-	'alpha-value-notation': importLazy(() => require('./alpha-value-notation/index.js'))(),
+	'alpha-value-notation': importLazy(() => require('./alpha-value-notation/index.cjs'))(),
 	'annotation-no-unknown': importLazy(() => require('./annotation-no-unknown/index.js'))(),
 	'at-rule-allowed-list': importLazy(() => require('./at-rule-allowed-list/index.js'))(),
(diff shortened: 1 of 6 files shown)


lib/rules/index.mjs
```diff
@@ -6,7 +6,7 @@ import importLazy from 'import-lazy';
 
 /** @type {import('stylelint')['rules']} */
 const rules = {
-	'alpha-value-notation': importLazy(() => require('./alpha-value-notation/index.js'))(),
+	'alpha-value-notation': importLazy(() => require('./alpha-value-notation/index.cjs'))(),
```
Mouvedia, line 9: question

Is this a typo? Don't you want `mjs`?

ybiquitous: Not a typo. We still need `.cjs` when using `require()`; if using requiring `.cjs`, then the following error occurs:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/masafumi.koba/git/stylelint/stylelint/lib/rules/alpha-value-notation/index.mjs not supported.
```

Like #7184, if we could replace `import-lazy` with a Promise-based code, we could rewrite `.cjs` to `.mjs` in `rules/index.mjs`.

mattxwang (approved):
Ah! Didn't realize you were waiting on a review, sorry.

This looks good to me! Thanks 😊

ybiquitous:
Never mind, thanks for the review!

By the way, the coverage degradation is due to `jest-preset-stylelint` uses `require` instead of `import`. We need to change the behavior after we finish to rewrite all rules in ESM.
https://github.com/stylelint/jest-preset-stylelint/blob/6.2.0/getTestRule.js#L18

Mouvedia:
`getTestRule` stopped me from migrating to vitest

PR #7183, opened 2023-09-12
Revert Node.js API returned resolved object change

### LINKED ISSUE #7141: Revert Node.js API returned resolved object change
As pointed out in https://github.com/stylelint/stylelint/pull/7136#pullrequestreview-1579157299, we've introduced a potentially disruptive change in https://github.com/stylelint/stylelint/pull/7119. 

The [original request](https://github.com/stylelint/stylelint/issues/6604) was to change the behaviour of the CLI to make it easier to use husky and lint-staged. We also made breaking changes to the Node.js API as part of the pull request, likely impacting editor integrations. By adding the `code` property, we cleaned up the API but at a potentially high cost because integrations are slow-moving.

I believe most integrations use `warnings[]` for lint problems and `output` for the fixed code, e.g. VS Code's [warnings](https://github.com/stylelint/vscode-stylelint/blob/main/src/utils/stylelint/process-linter-result.ts) and [fixes](https://github.com/stylelint/vscode-stylelint/blob/main/src/utils/documents/get-fixes.ts). If we revert to using `output` for fixed code, integrations will not need to make changes.

Rather than add a `code` property, we can add one that always contains the formatted lint problems so that they are accessible even when the `fix` option is used:

- `errors` - a string that contains the formatted problems (regardless of whether `fix` is used)
- `output` - a string that contains the autofixed code when the `fix` option is used. For backwards compatibility, it also contains the formatted problems otherwise.

What would be the best name for the property? `errors` would align with:

- the CLI's behaviour of stdout (for fixed code) and stderr (for formatted problems)
- the `errored` flag

But it is not factually accurate as it could contain formatted warnings, but then `warnings[]` also contains errors... (if we clean up the API, we should do the whole lot in one go and assess if the impact is worth it).

Another option is `formatterOutput`.

Thoughts?

ybiquitous:

Closes #7141
Follow-up to #7119 (partially reverted)

This Pull Request does:

- reverts the `output` property change in Node.js API returned promise object
- deprecates the `output` property
- adds a new `report` property instead of `output`


docs/migration-guide/to-16.md
```diff
@@ -27,10 +27,11 @@ You should use the following or higher versions of Node.js:
 
 We changed the resolved object of the Promise returned by `stylelint.lint()` so that:
 
-- the [`output`](../user-guide/node-api.md#output) property now only contains formatted problems
+- a new [`report`](../user-guide/node-api.md#report) property contains the formatted problems
 - a new [`code`](../user-guide/node-api.md#code-1) property contains the autofixed code
+- the [`output`](../user-guide/node-api.md#output) property is now deprecated and will be removed in the next major version.
```
Mouvedia, line 32: suggestion

`…deprecated in favor of the new `code` and `report` properties…`

lib/prepareReturnValue.cjs
```diff
@@ -42,11 +42,26 @@ function prepareReturnValue(stylelintResults, maxWarnings, formatter, cwd) {
 		cwd,
 		errored,
 		results: [],
-		output: '',
+		report: '',
+
+		// TODO: Deprecated. Remove in the next major version.
```
Mouvedia, line 47: minor

maybe use https://jsdoc.app/tags-deprecated.html instead?

lib/prepareReturnValue.cjs
```diff
@@ -42,11 +42,26 @@ function prepareReturnValue(stylelintResults, maxWarnings, formatter, cwd) {
...
+		get output() {
+			if (!this._outputWarned) {
+				console.warn('`output` is deprecated. Use `report` or `code` instead.');
+				this._outputWarned = true;
```
Mouvedia, line 51: question

We only have 4 more files that use that pattern:
https://github.com/stylelint/stylelint/blob/main/lib/utils/FileCache.js#L35
https://github.com/stylelint/stylelint/blob/main/lib/rules/max-line-length/__tests__/index.js#L6
https://github.com/stylelint/stylelint/blob/main/lib/utils/isStandardSyntaxTypeSelector.js#L25
https://github.com/stylelint/stylelint/blob/main/lib/utils/whitespaceChecker.js#L212

Should we review if it's consistent?

system-tests/systemTestUtils.mjs
```diff
@@ -37,15 +37,22 @@ export async function caseFilesForFix(caseNumber, ext = 'css') {
...
 
-export function prepForSnapshot({ results, cwd, output, ...rest }) {
+export function prepForSnapshot({ results, cwd, output, report, ...rest }) {
 	// If output isn't fixed code
 	if (output.startsWith('[')) {
 		// The `source` of each file varies between platforms or if a tmp file is used
-		output = JSON.parse(output).map((warning) => {
-			delete warning.source;
+		output = JSON.parse(output).map((warning) => ({
+			...warning,
```
Mouvedia, line 45: question

For consistency sake, should we refactor l61 and l62 at the same time?

docs/migration-guide/to-16.md
ybiquitous: Good catch. Fixed 29af41098 👍🏼

lib/prepareReturnValue.cjs
ybiquitous: We have used the `_` prefix for internal (private) properties, for example:

https://github.com/stylelint/stylelint/blob/3d79ac7ea5202af65bf990878656722bfd71f13a/types/stylelint/index.d.ts#L352-L357

system-tests/systemTestUtils.mjs
ybiquitous: Nice catch. Fixed ad05959ae.

lib/prepareReturnValue.cjs
Mouvedia: There's a difference between `@internal` and [`@private`](https://jsdoc.app/tags-private.html).
TBD in a refactor PR i.e. not this one

types/stylelint/index.d.ts
```diff
@@ -375,7 +375,23 @@ declare namespace stylelint {
 		cwd: string;
 		results: LintResult[];
 		errored: boolean;
+		/**
+		 * @deprecated Use `output` for the formatted problems, or use `code`
+		 *   for the autofixed code instead. This will be removed in the next major version.
+		 */
```
mattxwang, line 381:
```suggestion
		/**
		 * @deprecated Use `report` for the formatted problems, or use `code`
		 *   for the autofixed code instead. This will be removed in the next major version.
		 */
```

types/stylelint/index.d.ts
```diff
@@ -375,7 +375,23 @@ declare namespace stylelint {
...
 		output: string;
+		/** @internal To show the depreaction warning. */
+		_output?: string;
+		/** @internal To show the depreaction warning. */
+		_outputWarned?: boolean;
```
mattxwang, line 386:
```suggestion
		/** @internal To show the deprecation warning. */
		_output?: string;
		/** @internal To show the deprecation warning. */
		_outputWarned?: boolean;
```

.changeset/wise-parents-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: `report`/`code` properties to Node.js API returned object and deprecate `output` property instead
```
mattxwang, line 5: Suggestion: I think separating the "added" and "deprecated" actions could make it easier for users to notice that a deprecation happened as they scan the changelog. What are your thoughts?

mattxwang (commented):
I'm slowly catching up on the context, but this generally looks good to me. left a few minor nits!

.changeset/wise-parents-rule.md
Mouvedia: I wonder if you can add 2 lines/entries or if you need 2 separate markdown files.

ybiquitous: Since 2 lines per file are unsupported, I've split the entry into two files with 24eee76c9.

lib/standalone.mjs
```diff
@@ -60,9 +60,7 @@ export default async function standalone({
 }) {
 	const startTime = Date.now();
 
-	const isValidCode = typeof code === 'string';
-
-	if ((!files && !isValidCode) || (files && (code || isValidCode))) {
+	if ((!files && !isString(code)) || (files && isString(code))) {
```
Mouvedia, line 63: nit

```js
const hasOneValidInput = (!files && !isString(code)) || (files && isString(code));
```

lib/standalone.mjs
```diff
@@ -60,9 +60,7 @@ export default async function standalone({
...
 		return Promise.reject(
 			new Error('You must pass stylelint a `files` glob or a `code` string, though not both'),
```
Mouvedia, line 65: nit

It's missing `--stdin`.
i.e. from the user perspective `code` usage is an internal implementation detail
see https://github.com/stylelint/stylelint/blob/main/lib/cli.mjs#L539

types/stylelint/index.d.ts
```diff
@@ -375,7 +375,23 @@ declare namespace stylelint {
 		cwd: string;
 		results: LintResult[];
 		errored: boolean;
+		/**
+		 * @deprecated Use `report` for the formatted problems, or use `code`
+		 *   for the autofixed code instead. This will be removed in the next major version.
+		 */
 		output: string;
+		/** @internal To show the deprecation warning. */
```
Mouvedia, line 383: suggestion

I am not seeing `@internal` on https://jsdoc.app/
Maybe use https://jsdoc.app/tags-private.html instead?

lib/standalone.mjs
ybiquitous: No, `code` is a part of the Node.js API.

types/stylelint/index.d.ts
ybiquitous: TypeScript uses `@internal`, for example:
https://github.com/microsoft/TypeScript/blob/1a68590c9cd881b23d96424ea49bc9f42e524765/src/services/jsDoc.ts#L183

But the TypeScript document doesn't mention `@internal`:
https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html

I am unsure about this and don't have a strong opinion, but it's off-topic. Can you please open an issue about the usage of `@internal`?

lib/standalone.mjs
ybiquitous: Good catch. Fixed ef631f080.

lib/standalone.mjs
ybiquitous: Um, I don't think it's necessary to mention `--stdin` here because this module has an interest in Node.js API, not CLI.

mattxwang (approved):
I'm not sure if @Mouvedia has other comments left, but from my perspective this LGTM.

---

It's a bit of a bummer that code coverage through codecov is not working super well with the ESM migration. I'm happy to help wrap that up as we merge in the final `v16` branch, since I'm sure I will have missed some things.

ybiquitous:
Yeah, decreasing coverage is a headache. I guess this is caused by `*.cjs` addition. And I'll add test cases for CJS after rewriting to ESM.

Mouvedia:
@ybiquitous not really a big deal right now but you forgot to remove https://github.com/stylelint/stylelint/blob/v16/.changeset/thin-guests-smell.md

It needs to be done before the merge into main.

ybiquitous:
@Mouvedia Thanks. I completely forgot it. Opened PR #7227

types/stylelint/index.d.ts
Mouvedia: related: Microsoft/TypeScript-TmLanguage#480

Should we use [`--stripInternal`](https://www.typescriptlang.org/tsconfig#stripInternal)?

PR #7208, opened 2023-10-02
Fix `declaration-block-no-redundant-longhand-properties` false negatives for `*-block` and `*-inline` logical properties

### LINKED ISSUE #7200: Fix `declaration-block-no-redundant-longhand-properties` false negatives for `*-block` and `*-inline` logical properties

  margin-block-start: 2px;
  margin-block-end: 2px;
  margin-inline-start: 2px;
  margin-inline-end: 2px;

  margin-block: 2px;
  margin-inline: 2px;

### What actually happened?

  margin-block-start: 2px;
  margin-block-end: 2px;
  margin-inline-start: 2px;
  margin-inline-end: 2px;

mattxwang:

Closes #7200.

Bundling many properties into this PR as suggested in the original issue. I'm only including the ones that are trailed with `-block` and `-inline` (as scoped in the original issue title); I will address further false negatives in another PR.

I've chosen to base this off of `main` since:

- we can probably publish a patch off of this
- I don't think this'll introduce many (or any?) merge conflicts with `v16`

Affected properties:

- [x] `border-block`
- [x] `border-inline`
- [x] `inset-block`
- [x] `inset-inline`
- [x] `scroll-margin-block`
- [x] `scroll-margin-inline`
- [x] `scroll-padding-block`
- [x] `scroll-padding-inline`
- [x] `margin-block`
- [x] `margin-inline`
- [x] `padding-block`
- [x] `padding-inline`


mattxwang:
Going to mark this as ready-for-review since I think it's big enough!

ybiquitous (commented):
Great. Thanks for making this change!

One question, should we update the property list in the rule's README?

https://github.com/stylelint/stylelint/blob/fe3087002c99383f07ec25cca044bbfa8588cb62/lib/rules/declaration-block-no-redundant-longhand-properties/README.md?plain=1#L27

mattxwang:
> One question, should we update the property list in the rule's README?

Ah! Totally slipped my mind, thanks. Just added those and sorted the list alphabetically (I'm not sure what the previous order was).

(accidentally force-pushed - sorry about that! only the latest commit is different from the last review)

ybiquitous (approved):
Great. LGTM 👍🏼

PR #7213, opened 2023-10-06
Fix `declaration-block-no-redundant-longhand-properties` false negatives for `overflow`, `overscroll-behavior`, `scroll-margin`, `scroll-padding`, and new Box Alignment shorthands

mattxwang:

Ref: https://github.com/stylelint/stylelint/issues/7200#issuecomment-1734354995 & other missing shorthands.

Wasn't sure about `background-position` or `border-image-width` in the linked issue, so I left a comment.

Any thoughts on how to name this PR?

Current properties:

- [x] `gap`: [94% on caniuse](https://caniuse.com/mdn-css_properties_gap)
- [x] `overflow`
- [x] `overscroll-behavior`
- [x] `place-content`: [95% on caniuse](https://caniuse.com/mdn-css_properties_place-content)
- [x] `place-items`: [95% on caniuse](https://caniuse.com/mdn-css_properties_place-items)
- [x] `place-self`: [95% on caniuse](https://caniuse.com/mdn-css_properties_place-self)
- [x] `scroll-margin` (we already supported the `block`/`inline` properties)
- [x] `scroll-padding` (we already supported the `block`/`inline` properties)

Properties in the [shorthand list](https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties#see_also) that I'm choosing not to do (but could bundle in?):

- `container`: [88% on caniuse](https://caniuse.com/mdn-css_properties_container). I think this is supported by all major browsers, but is slightly smaller than the other shorthands we've supported
- `contain-intrinsic-size`: [74% on caniuse](https://caniuse.com/mdn-css_properties_contain-intrinsic-size), which seems low?
- `font-synthesis`: [92% on caniuse](https://caniuse.com/mdn-css_properties_font-synthesis), but the way this shorthand behaves is a bit different than the standard shorthand (it's almost like a flag?) so I'll do this in a different PR
- `font-variant`: [96% on caniuse](https://caniuse.com/mdn-css_properties_font-variant) same as `font-synthesis` re: shorthand behavior
- `scroll-timeline`: [62% on caniuse](https://caniuse.com/mdn-css_properties_scroll-timeline), definitely too low!


mattxwang:
Going to unmark this as draft, but before we merge this - I think a clearer PR title is necessary (which I'll then put into the changeset for this, which will be a patch).

While this would be accurate:

> Fix `declaration-block-no-redundant-longhand-properties` false negatives for `overflow`, `overscroll-behavior`, `scroll-margin`, `scroll-padding`, and new Box Alignment shorthands

It is likely too long. Any suggestions?

ybiquitous:
> It is likely too long. Any suggestions?

Surely, it's long bug clear to me. I think it's not a problem. 👍🏼

ybiquitous (approved):
Thank you. LGTM 👍🏼 

Please don't forget to update the changelog. 😄

Mouvedia:
> Please don't forget to update the changelog. 😄

Please add all the properties either in the changelog partial or the commit message during the squash.

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
```diff
@@ -301,6 +301,54 @@ testRule({
...
+			message: messages.expected('place-items'),
+		},
+		{
+			code: 'div { justify-self: center; align-self: end; }',
+			fixed: 'div { place-self: end center; }',
+			description: 'explicit place-self test',
+			message: messages.expected('place-self'),
+		},
+		{
+			code: 'a { scroll-margin-left: 10px; scroll-margin-right: 10px; scroll-margin-top: 20px; scroll-margin-bottom: 30px; }',
```
Mouvedia, line 341: minor nit

always use different values
i.e. we couldn't tell if `right` and `left` were badly positioned

mattxwang: Thanks! I also adjusted the original `margin` test that I got this line from.

PR #7214, opened 2023-10-07
Fix `declaration-block-no-redundant-longhand-properties` false negative for `font-synthesis`

mattxwang:

Ref: https://github.com/stylelint/stylelint/pull/7213#issue-1930795469 (and transitively, https://github.com/stylelint/stylelint/issues/7200#issuecomment-1734354995).

First, do we like this custom resolver? It is a bit long!

Secondly, `font-synthesis` actually supports *four* longhand properties. However, `font-synthesis-position` currently has [**0.41%** adoption according to caniuse](https://caniuse.com/mdn-css_properties_font-synthesis-position), which is far too small for it to be relevant. I've elected to not include it for now; however, we could also include it if we think that better future-proofs the rule?


lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -28,6 +28,42 @@ const meta = {
...
 const customResolvers = new Map([
+	[
+		'font-synthesis',
+		(decls) => {
+			const weight = decls.get('font-synthesis-weight')?.value.trim();
+			const style = decls.get('font-synthesis-style')?.value.trim();
+			const smallCaps = decls.get('font-synthesis-small-caps')?.value.trim();
+
+			const isValidFontSynthesisValue = (/** @type {string | undefined} */ s) =>
+				s === 'none' || s === 'auto';
```
ybiquitous, line 39: [suggest] Minor. The following seems a bit more readable to me:

```suggestion
			/** @type {(s: string | undefined) => boolean} */
			const isValidFontSynthesisValue = (s) => s === 'none' || s === 'auto';
```

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -28,6 +28,42 @@ const meta = {
...
+
+			if (
+				!isValidFontSynthesisValue(weight) ||
+				!isValidFontSynthesisValue(style) ||
+				!isValidFontSynthesisValue(smallCaps)
+			)
+				return;
```
ybiquitous, line 46: [suggest] Minor. Using brackets seems a bit clearer to me.

```suggestion
			) {
				return;
			}
```

ybiquitous (approved):
@mattxwang Thanks for creating the pull request!

Not supporting `font-synthesis-position` for now makes sense to me. We can revisit it later if a request comes. 👍🏼 

I've posted a few minor suggestion comments, but they are my preference. So you can ignore them (please resolve them if you won't address them).

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -28,6 +28,43 @@ const meta = {
...
+			/** @type {(s: string | undefined) => boolean} */
+			const isValidFontSynthesisValue = (s) => s === 'none' || s === 'auto';
+
+			if (
+				!isValidFontSynthesisValue(weight) ||
+				!isValidFontSynthesisValue(style) ||
+				!isValidFontSynthesisValue(smallCaps)
+			) {
+				return;
+			}
```
mattxwang, line 47: I realized that earlier, this line was uncovered. I've added a test that involves some invalid CSS (in https://github.com/stylelint/stylelint/pull/7214/commits/8fddced20cb68a0b539cce463d1f319abb49ad48) - is this the right way to go about this? 

(I noticed some of the earlier test cases also involve slightly malformed CSS, ex the transition test cases)

If not, I can remove the test case and/or try something else?

Mouvedia: > which are the only allowed values in the spec

I think that global values are also accepted.
i.e. inherit; initial; revert; revert-layer; unset;

I wonder if `decls.get` retrieves it raw or if it does the mapping automatically.
i.e. is `font-synthesis-weight: initial;` considered invalid? (should map to `auto` and hence be considered valid)

ybiquitous: I think the added test case is enough to cover the path. 👍🏼 

And supporting global values is ideal, but I don't think it's required for now. We can revisit it later if people request.

Mouvedia: > We can revisit it later if people request.

It should be done for all properties at once; let's leave it for a later enhancement.

PR #7215, opened 2023-10-07
Add `ignoreRules` to `max-nesting-depth`

### LINKED ISSUE #2696: Add ignoreRules: [] to max-nesting-depth

### LINKED ISSUE #4805: Add ignoreSelectors: [] to max-nesting-depth

mattxwang:

Closes #6529 (and thus, closes #2696 and closes #4805).

1. I've add co-author annotations to credit @sjarva and @ybiquitous with their contributions in the original PR; we should keep these when/if we merge. Also happy to merge this back into #6529 if @sjarva would like; I think we should give her the authorship for this commit if possible.
2. I noticed that there's a possibility where if both `ignoreRules` and `ignorePseudoClasses` are used together (as siblings), the original PR does not properly treat that combination as fully ignored. I've added a test case to verify that and then fix it in https://github.com/stylelint/stylelint/commit/cfa98ab42904aa6953819fabe140ade4649e87b9. The code isn't very elegant, so suggestions are welcome!

### DIFF
--- a/lib/rules/max-nesting-depth/index.js
+++ b/lib/rules/max-nesting-depth/index.js
@@ -28,6 +28,13 @@ const rule = (primary, secondaryOptions) => {
 	const isIgnoreAtRule = (node) =>
 		isAtRule(node) && optionsMatches(secondaryOptions, 'ignoreAtRules', node.name);
 
+	/**
+	 * @param {import('postcss').Node} node
+	 */
+	const isIgnoreRule = (node) => {
+		return isRule(node) && optionsMatches(secondaryOptions, 'ignoreRules', node.selector);
+	};
+
 	return (root, result) => {
 		const validOptions = validateOptions(
 			result,
@@ -42,6 +49,7 @@ const rule = (primary, secondaryOptions) => {
 				possible: {
 					ignore: ['blockless-at-rules', 'pseudo-classes'],
 					ignoreAtRules: [isString, isRegExp],
+					ignoreRules: [isString, isRegExp],
 					ignorePseudoClasses: [isString, isRegExp],
 				},
 			},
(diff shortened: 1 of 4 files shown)


ybiquitous (approved):
Thanks. LGTM 👍🏼

lib/rules/max-nesting-depth/index.js
```diff
@@ -120,10 +132,24 @@ const rule = (primary, secondaryOptions) => {
 		 * @param {string[]} selectors
 		 * @returns {boolean}
 		 */
-		function containsIgnoredPseudoClassesOnly(selectors) {
-			if (!(secondaryOptions && secondaryOptions.ignorePseudoClasses)) return false;
+		function containsIgnoredPseudoClassesOrRulesOnly(selectors) {
```
Mouvedia, line 135: nit

What about `hasIgnoredSelector`?

mattxwang: Hm, in this case it would have to be something like `hasOnlyIgnoredSelectors`, and there are some other functions in this rule that are also `contains*`. 

I do feel like these could be named better - how about I follow up with another PR to quickly make the names consistent? The nice thing about that is that it's not a user-facing change (so there's no urgency, etc.).

Mouvedia: > how about I follow up with another PR

No need it was really minor.

PR #7227, opened 2023-10-11
Remove unneeded changelog entry

ybiquitous:

Follow-up to #7183

No, it's self-explanatory.

### DIFF
--- a/.changeset/thin-guests-smell.md
+++ b/.changeset/thin-guests-smell.md
@@ -1,5 +1,0 @@
----
-"stylelint": major
----
-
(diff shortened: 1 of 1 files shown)


mattxwang (approved):

Mouvedia:
> this changelog entry seems still needed

The section in the guide is, not the changelog entry which is erroneous and outdated by the partial revert.

ybiquitous:
Okay. Let's move on. There's room to confirm the changelog and guide before the v16 release.

PR #7230, opened 2023-10-13
Fix `declaration-block-no-redundant-longhand-properties` autofix for `grid-template` with `repeat()`

### LINKED ISSUE #7228: Fix `declaration-block-no-redundant-longhand-properties` autofix for `grid-template` with `repeat()`

I've provided a [demo](https://github.com/paceaux/stylelint-template-repeat-bug) of the bug. 

1. Create a CSS ruleset containing these properties

```css
  grid-template-columns: repeat(3, 1fr);
  grid-template-rows: auto;
  grid-template-areas: "left center right";
```

2. Using stylelint, apply the `--fix` parameter and observe that it produces this

```css
    grid-template: "left center center right" auto / repeat(3, 1fr);


```

3. Observe the modified CSS in the browser, specifically that it is invalid.

```json
{
    "extends": [
      "stylelint-config-standard"
    ],
    "configBasedir": [
        "./"
    ],
    "ignoreFiles": [
      "node_modules/**/*",
      "dist/**/*.scss"
    ],
    "rules": {
      "indentation": 2
    }
  }
```

See the [demo](https://github.com/paceaux/stylelint-template-repeat-bug)

I've run the command

```bash
stylelint 'styles/**/*' --config .stylelintrc.json --fix
``` 

as an npm script

* stylelint-config-standard
* stylelint 15.10.3

autofix would produce valid syntax. [According the the CSS specifications](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template#values), MDN states that

> The [repeat()](https://developer.mozilla.org/en-US/docs/Web/CSS/repeat) function isn't allowed in these track listings, as the tracks are intended to visually line up one-to-one with the rows/columns in the "ASCII art".

So this expected output should not be a duplication from `grid-template-columns` but an _interpretation_ of `repeat()` for each column: (interpretation? extraction? parsing? IDK)

```css
grid-template-areas: "left center right" auto / 1fr 1fr 1fr;
```

### What actually happened?

The autofixed output copied the property from `grid-template-columns`, which, in the _specific_ case of having `repeat()`, is invalid:

```css
    grid-template: "left center center right" auto / repeat(3, 1fr);
```

Honestly I think the smartest thing to do is to skip grid-template with this rule because you're going to get some crazy use-cases. 

In lieu of that, you'll need to place a trailing `/` and then "unravel/parse/apply" the `repeat` into an appropriate number of units, which doesn't sound fun at all.

mattxwang:

Closes #7228.

Added the test case from the issue + the swapped version with `rows` instead of `columns`.

Are we interested in making a `repeat` parser? I could add it to my personal backlog (though this would probably be after the ESM migration, etc.). I'm also not sure what existing tooling is here (maybe there's some existing csstools/postcss/other parser that we can use)?

### DIFF
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
@@ -373,6 +373,20 @@ testRule({
 			description: 'explicit scroll-padding test',
 			message: messages.expected('scroll-padding'),
 		},
+		{
+			code: 'a { grid-template-columns: repeat(3, 1fr); grid-template-rows: auto; grid-template-areas: "left center right"; }',
+			unfixable: true,
+			description:
+				'the repeat() notation has non-trivial semantics and is currently not fixable (columns)',
+			message: messages.expected('grid-template'),
+		},
+		{
+			code: 'a { grid-template-columns: auto; grid-template-rows: repeat(3, 1fr); grid-template-areas: "left center right"; }',
+			unfixable: true,
+			description:
+				'the repeat() notation has non-trivial semantics and is currently not fixable (rows)',
+			message: messages.expected('grid-template'),
+		},
 	],
 });
 
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/index.js
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/index.js
@@ -100,7 +100,7 @@ const customResolvers = new Map([
 			// related issue: https://github.com/stylelint/stylelint/issues/7228
 			// spec ref: https://drafts.csswg.org/css-grid/#explicit-grid-shorthand
 
-			if (columns.includes('repeat') || rows.includes('repeat')) return;
+			if (columns.includes('repeat(') || rows.includes('repeat(')) return;
 
 			const splitAreas = [...areas.matchAll(/"[^"]+"/g)].map((x) => x[0]);
 			const splitRows = rows.split(' ');
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/index.js
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/index.js
@@ -96,6 +96,12 @@ const customResolvers = new Map([
 
 			if (!(areas && columns && rows)) return;
 
+			// repeat() is not allowed inside track listings for grid-template.
+			// related issue: https://github.com/stylelint/stylelint/issues/7228
+			// spec ref: https://drafts.csswg.org/css-grid/#explicit-grid-shorthand
+
+			if (columns.includes('repeat') || rows.includes('repeat')) return;
+
 			const splitAreas = [...areas.matchAll(/"[^"]+"/g)].map((x) => x[0]);
 			const splitRows = rows.split(' ');
 
--- a/.changeset/tricky-horses-watch.md
+++ b/.changeset/tricky-horses-watch.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `declaration-block-no-redundant-longhand-properties` autofix for `grid-template` with `repeat()`


lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
```diff
@@ -373,6 +373,20 @@ testRule({
 			description: 'explicit scroll-padding test',
 			message: messages.expected('scroll-padding'),
 		},
+		{
+			code: 'a { grid-template-columns: repeat(3, 1fr); grid-template-rows: auto; grid-template-areas: "left center right"; }',
```
Mouvedia, line 377: nit

Only one declaration is necessary.
ditto for l384
e.g. `a { grid-template-columns: repeat(3, 1fr); }`

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
```diff
@@ -373,6 +373,20 @@ testRule({
...
+				'the repeat() notation has non-trivial semantics and is currently not fixable (columns)',
+			message: messages.expected('grid-template'),
+		},
+		{
+			code: 'a { grid-template-columns: auto; grid-template-rows: repeat(3, 1fr); grid-template-areas: "left center right"; }',
+			unfixable: true,
+			description:
+				'the repeat() notation has non-trivial semantics and is currently not fixable (rows)',
+			message: messages.expected('grid-template'),
+		},
```
Mouvedia, line 389: minor refactor

maybe 1. add the fixed code 2. remove unfixable 3. add `skip: true,`
so if someone ever do fix it, the test case will be ready

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
mattxwang: Hm, I don't think that will work: it won't trip the rule (it requires *all* of the items represented by the shorthand to fire), and so no error will be reported by the rule. I just tried this locally to confirm. Did you mean something else?

Mouvedia (approved):

Mouvedia:
> Are we interested in making a `repeat` parser?

As a package dependency or as an util?

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
ybiquitous: I think keeping this as-is is better because `skip: true` is ambiguous about whether to mark TODO or skip the case for some reason (e.g. flakiness). If we could modify this unsupported autofix in the future, probably some issue would be opened.

lib/rules/declaration-block-no-redundant-longhand-properties/index.js
```diff
@@ -96,6 +96,12 @@ const customResolvers = new Map([
 
 			if (!(areas && columns && rows)) return;
 
+			// repeat() is not allowed inside track listings for grid-template.
+			// related issue: https://github.com/stylelint/stylelint/issues/7228
+			// spec ref: https://drafts.csswg.org/css-grid/#explicit-grid-shorthand
+
+			if (columns.includes('repeat') || rows.includes('repeat')) return;
```
ybiquitous, line 103: [suggest] For more accuracy, how about including "("?

```suggestion
			if (columns.includes('repeat(') || rows.includes('repeat(')) return;
```

ybiquitous (commented):
@mattxwang Thanks for the pull request. This is almost good to me. 👍🏼 

> Are we interested in making a `repeat` parser? 

Such a parser may be present, but I think this PR's simple string check is enough for now (the parser seems over-engineering to me for now).

ybiquitous (approved):
Thanks. LGTM 👍🏼

ybiquitous:
> As part of an existing package dependency or as an internal util?

FYI. Note this is not required. Maybe `@csstools/css-tokenizer` might achieve that:
https://github.com/csstools/postcss-plugins/tree/HEAD/packages/css-tokenizer#readme

cc @romainmenke

PR #7248, opened 2023-10-17
Refactor to add `isHexColor` utility

### LINKED ISSUE #7245: Refactor to add `isHexColor` utility
We have copy-and-pasted a similar function: `isHexColor`. This is a good chance for refactoring to add a new utility and reduce duplication.

See:
- https://github.com/search?q=repo%3Astylelint%2Fstylelint+isHexColor&type=code
- https://github.com/stylelint/stylelint/blob/f1d3228d21d197720785b41e2b505ce320146377/lib/rules/color-hex-length/index.js#L138-L143

0scvr:
Added `isHexcolor` utility function and replaced duplicated isHexColor functions and added unit test for the function.
Also removed unused `isValidHex` in `.../color-hex-alpha/index.js`.

Closes #7245

No, it's self-explanatory.

### DIFF
--- a/lib/utils/__tests__/isHexColor.test.mjs
+++ b/lib/utils/__tests__/isHexColor.test.mjs
@@ -10,6 +10,8 @@ test.each([
 	['#123123', true],
 	['#0080FF44', true],
 	['ffffff', false],
+	['#12345', false],
+	['#123456a', false],
 	['#-111', false],
 	[' #fff', false],
 ])('isHexColor(%s)', (argument, expected) => {
--- a/lib/utils/__tests__/isHexColor.test.mjs
+++ b/lib/utils/__tests__/isHexColor.test.mjs
@@ -1,0 +1,17 @@
+import isHexColor from '../isHexColor.js';
+import { test } from '@jest/globals';
+import valueParser from 'postcss-value-parser';
+
+test.each([
+	['#fff', true],
+	['#abcabc', true],
+	['#f1f2f3', true],
+	['#123', true],
+	['#123123', true],
+	['#0080FF44', true],
+	['ffffff', false],
+	['#-111', false],
+	[' #fff', false],
+])('isHexColor(%s)', (argument, expected) => {
+	expect(isHexColor(valueParser(argument).nodes[0])).toBe(expected);
+});
--- a/lib/utils/isHexColor.js
+++ b/lib/utils/isHexColor.js
@@ -1,0 +1,11 @@
+'use strict';
+
+const isValidHex = require('./isValidHex');
+
+/**
+ * @param {import('postcss-value-parser').Node} node
+ * @returns {boolean}
+ */
+module.exports = function isHexColor({ type, value }) {
+	return type === 'word' && isValidHex(value);
+};
--- a/lib/rules/color-hex-alpha/index.js
+++ b/lib/rules/color-hex-alpha/index.js
@@ -2,7 +2,7 @@
 
 const declarationValueIndex = require('../../utils/declarationValueIndex');
 const hasValidHex = require('../../utils/hasValidHex');
-const isValidHex = require('../../utils/isValidHex');
+const isHexColor = require('../../utils/isHexColor');
 const report = require('../../utils/report');
 const ruleMessages = require('../../utils/ruleMessages');
 const validateOptions = require('../../utils/validateOptions');
@@ -67,13 +67,6 @@ const rule = (primary) => {
  */
 function isUrlFunction({ type, value }) {
 	return type === 'function' && value === 'url';
-}
-
-/**
- * @param {import('postcss-value-parser').Node} node
- */
-function isHexColor({ type, value }) {
-	return type === 'word' && isValidHex(value);
 }
 
 /**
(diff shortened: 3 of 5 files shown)


lib/rules/color-hex-case/index.js
ybiquitous, line 1: [suggest] Can you revert the deprecated `color-hex-case` change? This rule will removed in v16.0.0 and has been removed in the `v16` branch: https://github.com/stylelint/stylelint/tree/v16/lib/rules

lib/utils/__tests__/isHexColor.test.mjs
```diff
@@ -0,0 +1,13 @@
...
+test('isHexColor', () => {
+	expect(isHexColor(valueParser('#fff').nodes[0])).toBe(true);
+	expect(isHexColor(valueParser('#abcabc').nodes[0])).toBe(true);
+	expect(isHexColor(valueParser('#f1f2f3').nodes[0])).toBe(true);
+	expect(isHexColor(valueParser('#123').nodes[0])).toBe(true);
+	expect(isHexColor(valueParser('#123123').nodes[0])).toBe(true);
+	expect(isHexColor(valueParser('ffffff').nodes[0])).toBe(false);
+	expect(isHexColor(valueParser('#-111').nodes[0])).toBe(false);
+	expect(isHexColor(valueParser(' #fff').nodes[0])).toBe(false);
+});
```
ybiquitous, line 13: [suggest] Minor. How about using [`test.each()`](https://jestjs.io/docs/api#testeachtablename-fn-timeout) for more simplicity?

```js
test.each([
  ['#fff', true],
  // other cases...
])('isHexColor(%s)', (argument, expected) => {
	expect(isHexColor(valueParser(argument).nodes[0])).toBe(expected);
});
```

ybiquitous (changes requested):
@0scvr Thanks for the first-time contribution! I've left a few suggestions, so could you please address them?

lib/utils/isHexColor.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+const HEX = /^#[\da-z]+$/i;
```
Mouvedia, line 3: suggestion
enhancement

Maybe we could restrict the accepted lengths to 3, 6 and 8?

lib/utils/__tests__/isHexColor.test.mjs
```diff
@@ -0,0 +1,13 @@
+import isHexColor from '../isHexColor.js';
+import valueParser from 'postcss-value-parser';
+
+test('isHexColor', () => {
+	expect(isHexColor(valueParser('#fff').nodes[0])).toBe(true);
+	expect(isHexColor(valueParser('#abcabc').nodes[0])).toBe(true);
+	expect(isHexColor(valueParser('#f1f2f3').nodes[0])).toBe(true);
+	expect(isHexColor(valueParser('#123').nodes[0])).toBe(true);
+	expect(isHexColor(valueParser('#123123').nodes[0])).toBe(true);
```
Mouvedia, line 9: minor

Could you add a test case for alpha?
e.g. `#0080FF44` should pass

lib/utils/__tests__/isHexColor.test.mjs
```diff
@@ -0,0 +1,13 @@
+import isHexColor from '../isHexColor.js';
+import valueParser from 'postcss-value-parser';
```
Mouvedia, line 2: question

Do you really need `valueParser` to test `isHexColor`?
i.e. just use the output directly

Mouvedia: @ybiquitous maybe we should change the type of the parameter in the JSDoc comment?
e.g. `@param {{ type: string, value: string }}`

lib/utils/isHexColor.js
Mouvedia: Good idea. Import `isValidHex` instead of using the regex.

lib/utils/__tests__/isHexColor.test.mjs
Mouvedia: Yeah, you are right.
Consistency is more important.

Mouvedia:
minor
slightly related

We will probably have to merge `isValidHex` and `hasValidHex` utils.
Not in this PR though.

lib/utils/__tests__/isHexColor.test.mjs
ybiquitous: Not only for consistency, I think clarifying the argument type is a node of valueParser is also important. If valueParser's type changed, these tests could tell us about the change.

ybiquitous:
@0scvr CI fails. Can you address it?

ybiquitous:
@0scvr The `color-no-hex` rule's tests fail. This rule detects an *invalid* hex value, so we cannot use `isValidHex()` for the rule.

https://github.com/stylelint/stylelint/blob/8914d463e6e2622ba0d82ae69d36e61f3c08eb89/lib/rules/color-no-hex/README.md?plain=1#L35

Now, I have 2 ideas to resolve this:

1. Add `strict: true` option to `isHexColor()` like this:
   ```js
   isHexColor(valueNode); // defaults to {strict: true} => use `isValidHex`
   isHexColor(valueNode, {strict: false}); // => use less strict regex
   ```
2. Just revert the change in `color-no-hex`

Any thoughts?

ybiquitous:
Thanks. Feel free to choose the one you want.

ybiquitous (approved):
Thank you. LGTM 👍🏼

PR #7252, opened 2023-10-19
Add `--rule` flag, `--rules` flag, `--merge-rules` flag and `mergeRules` option

Mouvedia:

#4331

- [x] standalone
- [x] cli
- [x] documentation
- [x] tests  

levn has limited uses; I could refactor and use `JSON.parse` instead. 
e.g. having a schema for an object requires explicit property names

###  missing features

- value-less rule selection

Passing a rule without its value seemed confusing.
What should `--rule color-no-invalid-hex` do exactly?
Probably fetch the corresponding rule option in the config and, if found, discard all other rules.
I chose not to add this feature, for now; it needs to be discussed in a separate issue.

- `--rules` should accept a file path (JSON)


lib/__tests__/cli.test.mjs
```diff
@@ -352,7 +452,7 @@ describe('CLI', () => {
 			fixturesPath('empty-block.css'),
 		]);
 
-		expect(process.exitCode).toBe(2);
+		expect(process.exitCode).toBeUndefined();
```
Mouvedia, line 455: see l208

It seems like `process.exitCode` was leaking from a previous test.

lib/getConfigForFile.js
```diff
@@ -57,19 +58,52 @@ module.exports = async function getConfigForFile(
...
+) {
+	const { config, mergeRules } = stylelint._options;
+
+	if (config && !mergeRules) {
+		return getAugmentedConfig(stylelint, filePath);
+	}
+
+	const cosmiconfigResult = await createCosmiconfig(stylelint, { filePath, searchPath });
+
+	if (config?.rules && mergeRules) {
```
Mouvedia, line 99: This file has been refactored to improve readability.
The only relevant change is this `if` and the one at l93.

lib/__tests__/standalone.test.mjs
```diff
@@ -287,6 +287,66 @@ it('unknown formatter option', async () => {
 	).rejects.toThrow(/^You must use a valid formatter option/);
 });
 
+describe('mergeRule', () => {
+	it('ignores configFile when config is passed and mergeRules is disabled (code)', async () => {
```
Mouvedia, line 291: Since `mergeRules` default is `false` this test and l305 _should_ have probably already be covered elsewhere, but I wanted to prevent regressions so I tested the status quo.

ybiquitous:
@Mouvedia Can you please wait until the `v16` branch is merged into `main`? Otherwise, we would have a pain to have to resolve so many conflicts.

Mouvedia:
@ybiquitous I am fine with waiting of the merge of v16, after all this issue was opened in 2019, it can wait a bit more.
But in the meantime can you review it?

Also Id say we should remove some issues from #6930 so that it can be released sooner; Ill comment there.

ybiquitous:
@Mouvedia Since `v16` has renamed most files to `.mjs`, `.js` files will have to change after `v16` is merged. Also, I'd like to focus on the ESM work for now.

As you said, we might be able to drop a few features from v16.

Mouvedia:
> Also, I'd like to focus on the ESM work for now.

Ill remove you from the reviewers list then.

> `.js` files will have to change after v16 is merged

That's not a big deal, Ill handle the conflicts after the merge.

mattxwang:
Hey, sorry - slowly catching up to all of my OSS obligations 😅 . I can review this in a couple of days, though I'd prefer to review this after `v16` is merged (so I'd have a better understanding of how it works with the new codebase). 

As a thought: should we shoot to include this in `16.1`? I anticipate that `16.0.x` will mostly just be bugfixes related to `v16` (and in part, the breaking changes / ESM migration)?

Mouvedia:
> I anticipate that 16.0.x will mostly just be bugfixes related to v16 (and in part, the breaking changes / ESM migration)?

We are planning a `-next` prerelease to avoid that. Even if we missed some, these bug fixes can be part of `16.1.0` as well.

ybiquitous:
@Mouvedia You can restart this. But I'm concerned about where the `--merge-rules` idea comes from.

Mouvedia:
Closed in favour of #7418.
@ybiquitous I am planning to finish it this week.

PR #7258, opened 2023-10-19
Refactor `lib/rules/selector-not-notation/index.js`

ybiquitous:

Extracted from #7211

No, it's self-explanatory.

### DIFF
--- a/lib/rules/selector-not-notation/index.js
+++ b/lib/rules/selector-not-notation/index.js
@@ -1,19 +1,15 @@
 'use strict';
 
+const selectorParser = require('postcss-selector-parser');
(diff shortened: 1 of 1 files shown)


lib/rules/selector-not-notation/index.js
```diff
@@ -1,19 +1,15 @@
 'use strict';
 
+const selectorParser = require('postcss-selector-parser');
+const { isAttribute, isClassName, isIdentifier, isPseudoClass, isTag, isUniversal } =
+	selectorParser;
```
Mouvedia, line 5: prettier is sure weird sometimes o0

PR #7268, opened 2023-10-20
Refactor `fixtures/custom-{parser,syntax}.js` to CJS

ybiquitous:

Ref #5291

This change explicitly appends the `.cjs` extension.

---

Note: The `customSyntax` doesn't support ESM yet.

https://github.com/stylelint/stylelint/blob/958bf369127f867747c613ec9c7025344e220eed/lib/getPostcssResult.js#L74


lib/__tests__/standalone-syntax.test.mjs
Mouvedia, line 1: Did you forget one at line 222?

ybiquitous: Good catch! Thank you. e965ce086

ybiquitous:
FYI. I've opened an issue (#7269) to support a custom syntax for ESM.

ybiquitous:
I'm also unsure, but Jest may internally change the `require()` function. The following is a debug log (`console.log(require)`):

```
[Function: bound requireModuleOrMock] {
  extensions: [Object: null prototype] {},
  resolve: [Function: resolve] { paths: [Function (anonymous)] },
  cache: [Object: null prototype] {},
  main: null
}
```

Outside the Jest context, it is:

```
[Function: require] {
  resolve: [Function: resolve] { paths: [Function: paths] },
  main: undefined,
  extensions: [Object: null prototype] {
    '.js': [Function (anonymous)],
    '.json': [Function (anonymous)],
    '.node': [Function (anonymous)]
  },
  cache: [Object: null prototype] {}
}
```

PR #7270, opened 2023-10-20
Refactor plugin fixtures to CJS

ybiquitous:

Ref #5291

Currently, we support plugins only for CJS. So, this change explicitly appends the `.cjs` extension.


lib/__tests__/fixtures/config-relative-plugin.json
Mouvedia, line 1: Did you miss https://github.com/stylelint/stylelint/blob/v16/lib/__tests__/fixtures/config-relative-plugin-nested.json#L2?

ybiquitous: No need to change the file because `plugin-nested-warn-about-foo` doesn't exist.

mattxwang (approved):
Overall, this LGTM.

One question (that can be addressed later), you said:

> Currently, we support plugins only for CJS. So, this change explicitly appends the `.cjs` extension.

Should we document this somewhere (e.g. in [Writing plugins](https://stylelint.io/developer-guide/plugins) or in the migration guide? I know this is "implicit" with the current state of Stylelint. but may be good to call out. I know it's not directly related to this PR (since it's only on test fixtures), but was just wondering.

ybiquitous:
I don't think adding that to the migration guide is needed, but adding that to the existing document may be good.

But I believe it's best to support ESM plugins, not only adding it to the doc.

Issue #7240, opened 2023-10-16
Remove Node.js 16 support


ybiquitous:
We will remove Node.js 14 support in the next Stylelint version 16.0.0 (#6740).

But Node.js 16 also reached EOL in September 2023. See the announcement about Node.js 16 EOL:
[Bringing forward the End-of-Life Date for Node.js 16](https://nodejs.org/en/blog/announcements/nodejs16-eol)

> We are moving the End-of-Life date of Node.js 16 by seven months to coincide with the end of support of OpenSSL 1.1.1 on September 11th, 2023.

Thus, some of the dependencies or libraries have dropped the support for Node.js 16. So should we also drop Node.js 16 in the next version 16.0.0 (#6930)?

I want any feedback. Thank you.

Mouvedia:
> So should we also drop Node.js 16 in the next version 16.0.0 (https://github.com/stylelint/stylelint/issues/6930)?

👎 for me
Id say dropping one node LTS per Stylelint major is enough.
Deprecating it? Maybe if the `console.warn` can be hidden using a flag.

Mouvedia:
Also could you add a list of the dependencies that are currently affected?
It will help out in our decision making.
e.g. lint-staged 15.0.0

mattxwang:
I don't think I have a strong opinion on this, though personally I have no issue dropping two node LTS issues in a major release (since it is EOL). We could also always release a v17 of Stylelint relatively soon after v16, if we want to pace it out (there are some other major-breaking changes that are smaller-scale that I could work on there).

> e.g. lint-staged 15.0.0

Would that be affected? `lint-staged` is just a dev dependency, so it doesn't affect downstream users of Stylelint (to my understanding).

Mouvedia:
> is just a dev dependency

Exactly my point.
I wanna have, at least, a list of the dependencies that are already dropping v16.
It would be better to have a list of the packages that are planning to drop it, but that would be quite time consuming to compile.

ybiquitous:
Listing up dependencies is a good idea.

Sharing [`ls-engines`](https://www.npmjs.com/package/ls-engines) report (on `v16` branch):

```console
$ npx ls-engines@latest
`node_modules` found; loading tree from disk...
┌────────┬─────────────────────────────────────────────────────────────────┐
│ engine │ Currently available latest release of each valid major version: │
├────────┼─────────────────────────────────────────────────────────────────┤
[...]

PR #7273, opened 2023-10-23
Revert "Bump postcss-safe-parser from 6.0.0 to 7.0.0"

ybiquitous:
Reverts stylelint/stylelint#7223

We should revert the update to `postcss-safe-parser@7.0.0` because the version drops the support of Node.js 16. We need to support Node.js 16 until the next major version (v16.0.0) of Stylelint. See #7240.

---

```console
$ npm view postcss-safe-parser@7.0.0 engines
{ node: '>=18.0' }

$ npm view postcss-safe-parser@6.0.0 engines
{ node: '>=12.0' }
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,7 +38,7 @@
         "picocolors": "^1.0.0",
         "postcss": "^8.4.31",
         "postcss-resolve-nested-selector": "^0.1.1",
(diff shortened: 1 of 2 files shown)


mattxwang (approved):
Ah! So sorry for missing this earlier. Yes, this looks good to me!

PR #7277, opened 2023-10-25
Refactor `standalone-parseErrors.test.mjs` to avoid CJS

ybiquitous:

Ref #5291

I believe we should avoid mocking in tests if possible.

### DIFF
--- a/lib/__tests__/standalone-parseErrors.test.mjs
+++ b/lib/__tests__/standalone-parseErrors.test.mjs
@@ -7,7 +7,7 @@ test('standalone with deprecations', async () => {
 		config: { rules: { 'selector-attribute-quotes': 'always' } },
 	});
 
-	expect(data.output).toContain('"stylelintType":"parseError"');
+	expect(data.report).toContain('"stylelintType":"parseError"');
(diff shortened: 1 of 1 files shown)


lib/__tests__/standalone-parseErrors.test.mjs
```diff
@@ -1,37 +1,18 @@
...
 test('standalone with deprecations', async () => {
 	const data = await standalone({
-		code: 'a {}',
-		config: configBlockNoEmpty,
+		code: 'a[=] {}',
+		config: { rules: { 'selector-attribute-quotes': 'always' } },
 	});
 
-	expect(data.output).toContain('Some parseError');
+	expect(data.output).toContain('"stylelintType":"parseError"');
```
Mouvedia, line 10: question

It should be `report` not `output`, correct?

ybiquitous: Good catch. Fixed with 2d89bf201. 👍🏼

Mouvedia (approved):
Apart from the title l4 which seems inaccurate, LGTM.

PR #7295, opened 2023-11-04
Add background-clip-text exception in property-no-vendor-prefix

### LINKED ISSUE #7294: Rejected -webkit-background-clip with text value

```css
.gradient {
  -webkit-background-clip: text;
  background-clip: text;
  -webkit-text-fill-color: transparent;
  -moz-text-fill-color: transparent;
}
```

```json
{
  "defaultSeverity": "error",
  "rules": { "property-no-vendor-prefix": true }
}
```

```shell
npx stylelint --config=.stylelintrc.json test.css
```

15.11.0

Since the rule does not delete the prefixed `text-fill-color` properties I expect no problems to be reported as `-webkit-background-clip: text` also seems to have [very widespread support](https://caniuse.com/background-clip-text).

`postcss/autoprefixer` has a 5 year old test to accept this [here](https://github.com/postcss/autoprefixer/tree/main/test/cases), see `background-clip.css`

### What actually happened?

```
test.css
 2:3  ✖  Unexpected vendor-prefix "-webkit-background-clip"  property-no-vendor-prefix

```

and it is autocorrectable, so it gets replaced with `background-clip: text` if `--fix` is given

_No response_

brunodccarvalho:
Closes #7294

### DIFF
--- a/lib/rules/property-no-vendor-prefix/__tests__/index.mjs
+++ b/lib/rules/property-no-vendor-prefix/__tests__/index.mjs
@@ -20,8 +20,18 @@ testRule({
 			code: 'a { box-sizing: border-box; }',
 		},
 		{
+			code: 'a { background-clip: text; }',
+		},
+		{
+			code: 'a { background-clip: content-box; }',
+		},
+		{
 			code: 'a { -webkit-font-smoothing: antialiased; }',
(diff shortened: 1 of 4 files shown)


ybiquitous:
#7294 has been closed. This PR is no longer needed.

PR #7300, opened 2023-11-06
Refactor tests to remove deprecated `output` property as possible

ybiquitous:

Follow-up to #7183

This change aims to reduce redundant warnings in test reports.
Note that it still cannot be removed completely where needed.


jeddy3 (approved):
LGTM, thank you.

PR #7307, opened 2023-11-07
Add `exports` field to `package.json`

### LINKED ISSUE #6258: Use `exports` field in `package.json` to forbid unexpected API usage

ybiquitous:

Closes #6258
Ref #5291

This Pull Request is necessary for [Conditional exports](https://nodejs.org/api/packages.html#conditional-exports) (ESM and CJS).

Note: For backward compatibility, we may need to export `stylelint/lib/utils/*` files.
The following script can easily output many `lib/utils/` settings.

<details><summary>Script</summary>
<p>

```js
import path from 'node:path';

import fg from 'fast-glob';

const exports = {
	'.': {
		import: './lib/index.mjs',
		require: './lib/index.cjs',
	},
	'./package.json': './package.json',
};

const files = fg.sync('lib/utils/isStandardSyntax*.mjs', { ignore: '**/__tests__/**' });

for (const file of files) {
	const base = path.basename(file, '.mjs');
	exports[`./lib/utils/${base}`] = {
		import: `./lib/utils/${base}.mjs`,
		require: `./lib/utils/${base}.cjs`,
	};
}

console.log(JSON.stringify(exports, null, 2));
```

</p>
</details> 

I'd like to hear feedback about exporting `lib/**` files. Should we export all files? I worry about frequent breaking changes in the future, if we exposed all modules (including internal ones).

Ideally, we should export utilities only through `stylelint.utils`, but actually we don't. Especially, `isStandardSyntax*` utilities are necessary to write a plugin or a custom rule.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
       "require": "./lib/index.cjs"
     },
     "./package.json": "./package.json",
-    "./lib/utils/isStandardSyntax*": "./lib/utils/isStandardSyntax*"
+    "./lib/utils/*": "./lib/utils/*"
   },
   "main": "lib/index.cjs",
   "types": "types/stylelint/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -29,62 +29,7 @@
       "require": "./lib/index.cjs"
     },
     "./package.json": "./package.json",
-    "./lib/utils/isStandardSyntaxAtRule": {
-      "import": "./lib/utils/isStandardSyntaxAtRule.mjs",
-      "require": "./lib/utils/isStandardSyntaxAtRule.cjs"
-    },
-    "./lib/utils/isStandardSyntaxColorFunction": {
-      "import": "./lib/utils/isStandardSyntaxColorFunction.mjs",
-      "require": "./lib/utils/isStandardSyntaxColorFunction.cjs"
-    },
-    "./lib/utils/isStandardSyntaxCombinator": {
-      "import": "./lib/utils/isStandardSyntaxCombinator.mjs",
-      "require": "./lib/utils/isStandardSyntaxCombinator.cjs"
-    },
-    "./lib/utils/isStandardSyntaxComment": {
-      "import": "./lib/utils/isStandardSyntaxComment.mjs",
-      "require": "./lib/utils/isStandardSyntaxComment.cjs"
-    },
-    "./lib/utils/isStandardSyntaxDeclaration": {
-      "import": "./lib/utils/isStandardSyntaxDeclaration.mjs",
-      "require": "./lib/utils/isStandardSyntaxDeclaration.cjs"
-    },
-    "./lib/utils/isStandardSyntaxFunction": {
-      "import": "./lib/utils/isStandardSyntaxFunction.mjs",
-      "require": "./lib/utils/isStandardSyntaxFunction.cjs"
-    },
-    "./lib/utils/isStandardSyntaxHexColor": {
-      "import": "./lib/utils/isStandardSyntaxHexColor.mjs",
-      "require": "./lib/utils/isStandardSyntaxHexColor.cjs"
-    },
-    "./lib/utils/isStandardSyntaxKeyframesName": {
-      "import": "./lib/utils/isStandardSyntaxKeyframesName.mjs",
-      "require": "./lib/utils/isStandardSyntaxKeyframesName.cjs"
-    },
-    "./lib/utils/isStandardSyntaxProperty": {
-      "import": "./lib/utils/isStandardSyntaxProperty.mjs",
-      "require": "./lib/utils/isStandardSyntaxProperty.cjs"
-    },
-    "./lib/utils/isStandardSyntaxRule": {
-      "import": "./lib/utils/isStandardSyntaxRule.mjs",
-      "require": "./lib/utils/isStandardSyntaxRule.cjs"
-    },
-    "./lib/utils/isStandardSyntaxSelector": {
-      "import": "./lib/utils/isStandardSyntaxSelector.mjs",
-      "require": "./lib/utils/isStandardSyntaxSelector.cjs"
-    },
-    "./lib/utils/isStandardSyntaxTypeSelector": {
-      "import": "./lib/utils/isStandardSyntaxTypeSelector.mjs",
-      "require": "./lib/utils/isStandardSyntaxTypeSelector.cjs"
-    },
-    "./lib/utils/isStandardSyntaxUrl": {
-      "import": "./lib/utils/isStandardSyntaxUrl.mjs",
-      "require": "./lib/utils/isStandardSyntaxUrl.cjs"
-    },
-    "./lib/utils/isStandardSyntaxValue": {
-      "import": "./lib/utils/isStandardSyntaxValue.mjs",
-      "require": "./lib/utils/isStandardSyntaxValue.cjs"
-    }
+    "./lib/utils/isStandardSyntax*": "./lib/utils/isStandardSyntax*"
   },
   "main": "lib/index.cjs",
   "types": "types/stylelint/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
   },
   "license": "MIT",
   "author": "stylelint",
-  "type": "module",
   "exports": {
     ".": {
       "import": "./lib/index.mjs",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,71 @@
   },
   "license": "MIT",
   "author": "stylelint",
-  "main": "lib/index.js",
+  "type": "module",
+  "exports": {
+    ".": {
+      "import": "./lib/index.mjs",
+      "require": "./lib/index.cjs"
+    },
+    "./package.json": "./package.json",
+    "./lib/utils/isStandardSyntaxAtRule": {
+      "import": "./lib/utils/isStandardSyntaxAtRule.mjs",
+      "require": "./lib/utils/isStandardSyntaxAtRule.cjs"
+    },
+    "./lib/utils/isStandardSyntaxColorFunction": {
+      "import": "./lib/utils/isStandardSyntaxColorFunction.mjs",
+      "require": "./lib/utils/isStandardSyntaxColorFunction.cjs"
+    },
+    "./lib/utils/isStandardSyntaxCombinator": {
+      "import": "./lib/utils/isStandardSyntaxCombinator.mjs",
+      "require": "./lib/utils/isStandardSyntaxCombinator.cjs"
+    },
+    "./lib/utils/isStandardSyntaxComment": {
+      "import": "./lib/utils/isStandardSyntaxComment.mjs",
+      "require": "./lib/utils/isStandardSyntaxComment.cjs"
+    },
+    "./lib/utils/isStandardSyntaxDeclaration": {
+      "import": "./lib/utils/isStandardSyntaxDeclaration.mjs",
+      "require": "./lib/utils/isStandardSyntaxDeclaration.cjs"
+    },
+    "./lib/utils/isStandardSyntaxFunction": {
+      "import": "./lib/utils/isStandardSyntaxFunction.mjs",
+      "require": "./lib/utils/isStandardSyntaxFunction.cjs"
+    },
+    "./lib/utils/isStandardSyntaxHexColor": {
+      "import": "./lib/utils/isStandardSyntaxHexColor.mjs",
+      "require": "./lib/utils/isStandardSyntaxHexColor.cjs"
+    },
+    "./lib/utils/isStandardSyntaxKeyframesName": {
+      "import": "./lib/utils/isStandardSyntaxKeyframesName.mjs",
+      "require": "./lib/utils/isStandardSyntaxKeyframesName.cjs"
+    },
+    "./lib/utils/isStandardSyntaxProperty": {
+      "import": "./lib/utils/isStandardSyntaxProperty.mjs",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -22,7 +22,71 @@
...
+    "./lib/utils/isStandardSyntaxUrl": {
+      "import": "./lib/utils/isStandardSyntaxUrl.mjs",
+      "require": "./lib/utils/isStandardSyntaxUrl.cjs"
+    },
+    "./lib/utils/isStandardSyntaxValue": {
+      "import": "./lib/utils/isStandardSyntaxValue.mjs",
+      "require": "./lib/utils/isStandardSyntaxValue.cjs"
+    }
+  },
+  "main": "lib/index.cjs",
```
ybiquitous, line 89: [note] `main` is for backward compatibility.

Mouvedia: question
nit

There's a risk of revert for this change.
Maybe it would be safer to have it in a separate PR?
i.e. must it be slipstreamed at the same time as l27–30?

ybiquitous: Make sense. Opened a new PR #7308. 👍🏼

package.json
Mouvedia, line 1: Do we need a `default`?

> ["default" - the generic fallback that always matches. Can be a CommonJS or ES module file. This condition should always come last.](https://nodejs.org/api/packages.html#conditional-exports)

ybiquitous: Hum. If using `default`, what should be specified as a fallback? A CommonJS file?

```json
{
  "import": "./lib/index.mjs",
  "require": "./lib/index.cjs",
  "default": "./lib/index.cjs"
}
```

But this doesn't seem to make sense... 🤔

Mouvedia: 1. do we need `default`?
2. if so, which extension should be preferred?

webpack documentation says

> `exports` should be written to use fallbacks for unknown future cases. `default` condition can be used for that.

Mouvedia:
Could you also update the paragraph [here](https://github.com/stylelint/stylelint/blob/v16/docs/developer-guide/plugins.md#stylelintutils)?

package.json
ybiquitous: From the point of view of bundlers? To be honest, I'd like to keep the code simple as much as possible until it's really needed, if there are no clear advantages.

I'm unsure if this example is proper, but PostCSS doesn't use `default`. See:
https://github.com/postcss/postcss/blob/fa8cd158fd174429bd3f2234dc1339d9aee1b84b/package.json#L9-L12

package.json
```diff
@@ -22,7 +22,71 @@
...
   "author": "stylelint",
-  "main": "lib/index.js",
+  "type": "module",
+  "exports": {
+    ".": {
+      "import": "./lib/index.mjs",
+      "require": "./lib/index.cjs"
+    },
+    "./package.json": "./package.json",
+    "./lib/utils/isStandardSyntaxAtRule": {
```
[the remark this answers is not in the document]
ybiquitous: Thanks, I missed the [Subpath patterns](https://nodejs.org/api/packages.html#subpath-patterns) feature. It can simplify the settings. Fixed with 027567a07.

ybiquitous:
> Could you also update the paragraph [here](https://github.com/stylelint/stylelint/blob/v16/docs/developer-guide/plugins.md#stylelintutils)?

Good point. We should clarify the policy for the internal utilities here.

The current documentation says:

- https://github.com/stylelint/stylelint/blob/25140c070f1e5130ab5dd92e4ab541b62b2dcd36/docs/developer-guide/plugins.md?plain=1#L216-L220
- https://github.com/stylelint/stylelint/blob/25140c070f1e5130ab5dd92e4ab541b62b2dcd36/docs/developer-guide/rules.md?plain=1#L106-L113

In my opinion, we should keep the policy that we should expose utilities at a minimum because it may cause frequent breaking changes.

However, if we restricted importing the internals by the `exports` field, many packages depending on the internals would not work. This should be a pain in the community.

So, I suggest the following:
- Keep the copy-paste policy in the documentation. This means an internal utility change won't be a breaking change.
- Allow plugin authors to import `lib/utils/*` modules by `exports`

Any thoughts?

ybiquitous:
Fixed to `lib/utils/*` with e1597eac4

ybiquitous:
The following is a test of `./lib/utils/*` in the `exports` field.

`package.json`:

```json
{
  "dependencies": {
    "stylelint": "github:stylelint/stylelint#add-exports-field"
  }
}
```

Run:

```console
$ npm install
...

$ node -pe 'require("stylelint/lib/utils/typeGuards.cjs")'
{
  hasSource: [Function: hasSource],
  isAtRule: [Function: isAtRule],
  isComment: [Function: isComment],
  isDeclaration: [Function: isDeclaration],
  isDocument: [Function: isDocument],
  isRoot: [Function: isRoot],
  isRule: [Function: isRule],
  isValueFunction: [Function: isValueFunction]
}

$ node -e 'import("stylelint/lib/utils/typeGuards.mjs").then(console.log)'
[Module: null prototype] {
  hasSource: [Function: hasSource],
  isAtRule: [Function: isAtRule],
  isComment: [Function: isComment],
  isDeclaration: [Function: isDeclaration],
  isDocument: [Function: isDocument],
  isRoot: [Function: isRoot],
  isRule: [Function: isRule],
  isValueFunction: [Function: isValueFunction]
}
```

package.json
```diff
@@ -22,6 +22,15 @@
   },
   "license": "MIT",
   "author": "stylelint",
+  "type": "module",
```
Mouvedia, line 25: question

> Regardless of the value of the "type" field, .mjs files are always treated as ES modules and .cjs files are always treated as CommonJS.

Why do we need `type`?
It's only useful for `.js` files; I don't think we have any anymore.

Is it for other tools?
Or to declare our intent to transition to pure ESM?

ybiquitous: > to declare our intent to transition to pure ESM in the next major?

This is my intent, but `type` is actually unneeded if there are any downsides.

Mouvedia: > any downsides

If you are planning to add `.js` files in the future, remove it.
Adding `type` without any reason seems weird to me, but I am sure it will have consequences whether it's missing or not. I hope the next tag will help us resolve that.

minor

If you keep it, we will need to add somewhere in the documentation something about always using either `.cjs` or `.mjs` until v17; no need to do it in this PR though. Or alternatively just use [this rule](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md) with `"js": "never"`.

ybiquitous: Make sense. Removed `type` with 9a6621c35.

Mouvedia: For the record, I am not sure either way.
Let's wait for someone to request it or v17.

Mouvedia (approved):
LGTM

package.json
```diff
@@ -22,6 +22,14 @@
   },
   "license": "MIT",
   "author": "stylelint",
+  "exports": {
+    ".": {
+      "import": "./lib/index.mjs",
+      "require": "./lib/index.cjs"
+    },
+    "./package.json": "./package.json",
+    "./lib/utils/*": "./lib/utils/*"
```
ybiquitous, line 31: [suggest] If we returned to `.js` from `.mjs` in the next major version (pure ESM), avoiding `.mjs` access might be better for compatibility. Any thoughts?

```suggestion
    "./lib/utils/*": {
      "import": "./lib/utils/*.mjs",
      "require": "./lib/utils/*.cjs"
    }
```

jeddy3 (commented):
> I'd like to hear feedback about exporting lib/** files. Should we export all files? I worry about frequent breaking changes in the future, if we exposed all modules (including internal ones).

In https://github.com/stylelint/stylelint/issues/6866#issuecomment-1621545398, we settled on not exporting more utils (e.g. because people shifting to standard CSS makes the `isStandardSyntax*` utils less relevant and so on).

Am I right in thinking that our move to ESM will break people's imports anyway, regardless of whether we export the utils as the extensions have changed:

```diff js
- require("stylelint/lib/utils/typeGuards.js")'
+ require("stylelint/lib/utils/typeGuards.cjs")'
```

Rather than apply another bandaid, we could take this opportunity to make it clear in our migration guide that plugin authors need to copy any internal utils they want (like [stylelint-scss does](https://github.com/stylelint-scss/stylelint-scss/tree/master/src/utils)). This clarification is likely worth it as the ambiguity has been problematic in the past, e.g. with stylelint-stylistic.

This will also address https://github.com/stylelint/stylelint/pull/7307#discussion_r1385969381, where changing back to `.js` would be another breaking change for those importing internal `.mjs` utils.

ybiquitous:
@jeddy3 Thanks for the feedback. I also don't want to change the policy we settled on <https://github.com/stylelint/stylelint/issues/6866#issuecomment-1621545398>.

And yes, the following code would break in 16.0.0:

```js
require("stylelint/lib/utils/typeGuards.js")
```

But, if we have `./lib/utils/*` like <https://github.com/stylelint/stylelint/pull/7307#discussion_r1385969381>, the following code will NOT break (the difference is just the existence of `.js`):

```js
require("stylelint/lib/utils/typeGuards")
```

On the other hand, if we do NOT provide `./lib/utils/*`, the code above will break, that is the following error will occur:

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './lib/utils/typeGuards' is not defined by "exports"
```

I'm worried that this breaking might be too strict. So I suggest:

- In 16.0.0, we deprecate the use of internal importing like `require('stylelint/lib/**')` or `import 'stylelint/lib/**'`, but allow the use though `exports`.
- In 17.0.0, we remove `exports` for `./lib/**`.

Any thoughts?

jeddy3:
I'd assumed most people were using the `.js` extension, but (after a [cursory look](https://github.com/search?q=stylelint%2Flib%2Futils%2F&type=code&p=1)) it seems most aren't.

So, your suggestion SGTM. It'll stagger the amount of people affected over two major releases, with the most happening in 17.0.0.

> And yes, the following code would break in 16.0.0:

Let's add something to the migration guide as part of this pull request for this. And let's also use the opportunity to steer people in the right direction. Something along the lines of:

## Changed extension of internal files

We've changed the extension of our internal files from `.js` to `.mjs` and `.cjs`. If you're a plugin or integration author that `import`s or `require`s any internal files (e.g. our `util/*` ones), we recommend copying these to your project instead as we:

- intend to disallow access to them in the next major release
- may remove and change files that are not part of our public API outside of a major release  

We don't recommend it, but you can unsafely continue to `import` or `require` the files until the next major release by using the appropriate new extension. For example:

```diff js
- require("stylelint/lib/utils/typeGuards.js")'
+ require("stylelint/lib/utils/typeGuards.cjs")'
```

```diff js
- import isStandardSyntaxAtRule from 'stylelint/lib/utils/isStandardSyntaxAtRule.js';
+ import isStandardSyntaxAtRule from 'stylelint/lib/utils/isStandardSyntaxAtRule.mjs';
```

ybiquitous:
@jeddy3 Sounds good! I've updated the migration guide (565ba5dd0) and added a changelog entry (72d77508b). Please take a look at them. 🙏🏼

ybiquitous:
Feel free to modify the documents directly if necessary. I'm not so confident in my English wording. 😅

package.json
ybiquitous: We realize importing without an extension is unnecessary because we will disallow access to internal utils in the next major release.

Also, `import` without an extension doesn't seem to fit the ESM spec.

See https://github.com/stylelint/stylelint/pull/7307#issuecomment-1801789503

Mouvedia (approved):
Talking about the removal of `type`, here's an example of where it would have been useful: if we were keeping the utils extension as `.js` to avoid breaking changes. In this hypothetical case it would be `"type": "commonjs"` though.
[As @jeddy3 said](https://github.com/stylelint/stylelint/pull/7307#issuecomment-1801789503), I was assuming—wrongly?—that extension-less usage would be rare. If it's really not the case then reverting to `.js` from `.cjs` for this folder might not be necessary.

ybiquitous:
For your information, suppose we provide pure ESM in 17.0.0 (the next to 16.0.0).

- `package.json` will be:
  ```json
  {
    "type": "module",
    "exports": {
      ".": "./lib/index.js",
      "./package.json": "./package.json"
    }
  }
  ```
- `.mjs` will be converted to `.js`
- `.cjs` will be removed

Then, most codes depending on `stylelint` should be broken:

```js
import("stylelint"); //=> ✅ Not broken
import("stylelint/lib/utils/typeGuards.mjs"); //=> ❌ Broken due to invalid `.mjs`
import("stylelint/lib/utils/typeGuards"); //=> ❌ Broken due to no `exports`
require("stylelint"); //=> ❌ Broken due to ERR_REQUIRE_ESM
```

Mouvedia:
> `.mjs` will be converted to `.js`

I don't think that's necessary. Once we have `"type": "module"`, both `.js` and `.mjs` can coexist.
In v17, at least the main `index.mjs` and the utils should remain `.mjs` to smooth the transition for the users.
Anyhow, that will probably be discussed in the upcoming __Prepare 17.0.0_.

ybiquitous:
> > `.mjs` will be converted to `.js`
>
> I don't think that's necessary. Once we have `"type": "module"`, both `.js` and `.mjs` can coexist.

Ah, my bad. You're correct.

jeddy3 (approved):
LGTM, thank you!

PR #7308, opened 2023-11-07
Fix incorrect `main` field in `package.json`

ybiquitous:

Extract from #7307

No, it's self-explanatory.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "license": "MIT",
   "author": "stylelint",
-  "main": "lib/index.js",
+  "main": "lib/index.cjs",
   "types": "types/stylelint/index.d.ts",
   "bin": {
     "stylelint": "bin/stylelint.mjs"


package.json
```diff
@@ -22,7 +22,7 @@
   },
   "license": "MIT",
   "author": "stylelint",
-  "main": "lib/index.js",
+  "main": "lib/index.cjs",
```
Mouvedia, line 25: I am not sure, but I think the agnostic `"lib/index"` should work.
Can you try it out?

Mouvedia:
Can you add a changeset?

ybiquitous:
> Can you add a changeset?

I don't think it's needed because of no behavior change.

package.json
Mouvedia: [It seems that automatic extension resolution has been removed.](https://stackoverflow.com/posts/46900211/revisions)
You can disregard my suggestion.

Mouvedia (approved):

PR #7309, opened 2023-11-07
Change TypeScript definitions for ESM

ybiquitous:

Follow-up to #5291
~~Wait for #7307~~
Block #7160

This Pull Request mainly updates the TypeScript definitions for ESM.

Summary:
- Rewrite TypeScript definitions to use ESM syntax, avoiding the specific syntax like `namespace`.
- Update `tsconfig.json` for ESM and Node.js 18.
- Remove no longer needed `test.d.ts`.
- Add a patch for the `is-plain-object` package to avoid type-check errors.

See also:
- https://www.typescriptlang.org/tsconfig
- https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing
- https://github.com/jonschlinkert/is-plain-object/pull/28

### DIFF
--- a/lib/createStylelint.cjs
+++ b/lib/createStylelint.cjs
@@ -11,7 +11,7 @@ const IS_TEST = process.env.NODE_ENV === 'test';
 const STOP_DIR = IS_TEST ? process.cwd() : undefined;
 
 /**
- * @type {import('stylelint')['_createLinter']}
+ * @type {import('stylelint').PublicApi['_createLinter']}
  */
 function createStylelint(options = {}) {
 	const cwd = options.cwd || process.cwd();
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,8 +1,9 @@
 {
 	"compilerOptions": {
-		"target": "ES2021",
-		"module": "commonjs",
-		"lib": ["ES2021", "DOM"],
+		"module": "Node16",
+		"moduleResolution": "Node16",
+		"target": "ES2022",
+		"lib": ["ES2022"],
 		"checkJs": true,
 		"noEmit": true,
 		"strict": true,
@@ -12,13 +13,13 @@
 		"noImplicitReturns": false,
 		"noUnusedParameters": true,
 		"noUncheckedIndexedAccess": true,
-		"resolveJsonModule": true,
+		"resolveJsonModule": false,
 		"esModuleInterop": true,
 		"skipLibCheck": false,
 		"paths": {
-			"*": ["./node_modules/@types/*", "./types/*"]
+			"stylelint": ["./types/stylelint/index.d.ts"]
 		}
 	},
-	"include": ["lib", "types", "package.json"],
-	"exclude": ["**/*.test.{js,mjs}", "**/__tests__/", "lib/**/*.cjs"]
+	"include": ["lib/**/*.mjs", "types/**/*.ts"],
+	"exclude": ["**/__tests__/**"]
 }
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -3,17 +3,20 @@ import type { GlobbyOptions } from 'globby';
 import type { cosmiconfig, TransformSync as CosmiconfigTransformSync } from 'cosmiconfig';
 
 type ConfigExtends = string | string[];
-type ConfigPlugins = string | stylelint.Plugin | (string | stylelint.Plugin)[];
+
+type ConfigPlugins = string | Plugin | (string | Plugin)[];
+
 type ConfigIgnoreFiles = string | string[];
 
 type ConfigRules = {
-	[ruleName: string]: stylelint.ConfigRuleSettings<any, Object>;
+	[ruleName: string]: ConfigRuleSettings<any, Object>;
 };
-type ConfigOverride = Omit<stylelint.Config, 'overrides'> & {
+
+type ConfigOverride = Omit<Config, 'overrides'> & {
 	files: string | string[];
 };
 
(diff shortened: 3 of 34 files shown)


types/stylelint/index.d.ts
ybiquitous, line 1: [note] I recommend reviewing this file ignoring whitespaces:
https://github.com/stylelint/stylelint/pull/7309/files?w=1

tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
 	"compilerOptions": {
-		"target": "ES2021",
-		"module": "commonjs",
-		"lib": ["ES2021", "DOM"],
+		"module": "Node16",
+		"moduleResolution": "Node16",
+		"target": "ES2022",
+		"lib": ["ES2022"],
```
Mouvedia, line 6: According to https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping#node-16 it should be

```
		"target": "ES2021",
		"lib": ["ES2021"],
```

tsconfig.json
```diff
@@ -12,13 +13,13 @@
 		"noImplicitReturns": false,
 		"noUnusedParameters": true,
 		"noUncheckedIndexedAccess": true,
-		"resolveJsonModule": true,
+		"resolveJsonModule": false,
```
Mouvedia, line 16: Import assertions are required since node version 17.5.0.
This change aims to prohibit json imports?

tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
 	"compilerOptions": {
-		"target": "ES2021",
-		"module": "commonjs",
-		"lib": ["ES2021", "DOM"],
+		"module": "Node16",
+		"moduleResolution": "Node16",
```
Mouvedia, line 4: info

> node16 and nodenext are currently identical, with the exception that they [imply different target option values](https://www.typescriptlang.org/docs/handbook/modules/reference.html#implied-and-enforced-options). If Node.js makes significant changes to its module system in the future, node16 will be frozen while nodenext will be updated to reflect the new behavior.

tsconfig.json
ybiquitous: > Import assertions are required since node version 17.5.0.

The Import assertions (Import attributes?) feature seems still experimental, according to the [Node.js document](https://nodejs.org/api/esm.html#import-attributes).

> This change aims to prohibit json imports?

Yes.

lib/createStylelint.cjs
```diff
@@ -11,7 +11,7 @@ const IS_TEST = process.env.NODE_ENV === 'test';
 const STOP_DIR = IS_TEST ? process.cwd() : undefined;
 
 /**
- * @type {import('stylelint')['_createLinter']}
+ * @type {import('stylelint').PublicApi['_createLinter']}
```
Mouvedia, line 14: nit

It's kind of weird to access something that is expected to be private through `PublicApi`.

tsconfig.json
```diff
@@ -12,13 +13,13 @@
...
 		"skipLibCheck": false,
 		"paths": {
-			"*": ["./node_modules/@types/*", "./types/*"]
+			"stylelint": ["./types/stylelint/index.d.ts"]
 		}
 	},
-	"include": ["lib", "types", "package.json"],
-	"exclude": ["**/*.test.{js,mjs}", "**/__tests__/", "lib/**/*.cjs"]
+	"include": ["lib/**/*.mjs", "types/**/*.ts"],
+	"exclude": ["**/__tests__/**"]
```
Mouvedia, line 24: question

The removal of `"**/*.test.{js,mjs}"` didn't have any consequence?
What's special about `"**/__tests__/**"` that warrants to keep excluding it?

ybiquitous: > The removal of `"**/*.test.{js,mjs}"` didn't have any consequence?

No consequence. Because all `*.test.{js,mjs}` files are under the `__tests__` directory:

```console
$ git ls-files lib/ | grep '\.test\.' | grep -v '__tests__'
(no output)
```

> What's special about `"**/__tests__/**"` that warrants to keep excluding it?

This exclusion aims to prevent so many type-check errors in the test files.

jeddy3 (approved):
LGTM, thank you!

remcohaszing:
This change broke type definitions for CJS. It also didn’t add type definitions for ESM, meaning ESM types now fallback to the broken CJS types.

https://arethetypeswrong.github.io?p=stylelint@16.0.0

types/stylelint/index.d.ts
```diff
@@ -805,22 +811,22 @@ type PublicApi = PostCSS.PluginCreator<stylelint.PostcssPluginOptions> & {
...
 	reference: {
-		longhandSubPropertiesOfShorthandProperties: stylelint.LonghandSubPropertiesOfShorthandProperties;
+		longhandSubPropertiesOfShorthandProperties: LonghandSubPropertiesOfShorthandProperties;
 	};
 };
 
 declare const stylelint: PublicApi;
 
-export = stylelint;
+export default stylelint;
```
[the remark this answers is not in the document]
remcohaszing: Yes, but the removal of the `stylelint` namespace is related.

If you publish both CJS and ESM JavaScript code, you also need to publish both CJS and ESM type definitions. I have a hunch this is as simple as:

```ts
// index.d.mts
export * from './index.js'
export { default } from './index.js'
```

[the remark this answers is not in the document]
ybiquitous: @remcohaszing Thank you. Is there a way to test both types on CI or localhost?

Issue #7322, opened 2023-11-13
Fix ERR_REQUIRE_ESM when importing ESM config with `.js`


rakleed:
### Describe the documentation issue

I don't know what the syntax for `stylelint.config.js` should be since all the examples are for `.stylelintrc` only. But ok, I found in the documentation of other linters that I can use `module.exports = {}` for CJS, but what is the syntax for ESM? I tried `export default {}` and `export config = {}` but it didn't work as there was always an `Error [ERR_REQUIRE_ESM]: require() of ES Module` error in the console even in [version 16](https://github.com/stylelint/stylelint/releases/tag/16.0.0-0).

### What solution would you like to see?

In https://stylelint.io/user-guide/configure it is worth adding at least one example of what the syntax of the `stylelint.config.js` file should look like for CJS and ESM.

newives:
may be related to this https://github.com/cosmiconfig/cosmiconfig/issues/312

ybiquitous:
@rakleed Thanks for the feedback. We can reproduce the error as follows:

`package.json` (see `type: module`):

```json
{
  "type": "module",
  "dependencies": {
    "stylelint": "16.0.0-0"
  },
  "scripts": {
    "test": "echo \"a{}\" | stylelint"
  }
}
```

`stylelint.config.js`:

```js
export default {
  rules: {
    'block-no-empty': true
  }
};
```

Run:

```console
$ npm t

> test
> echo "a{}" | stylelint

Error [ERR_REQUIRE_ESM]: require() of ES Module /private/tmp/stylelint-issue-7322/stylelint.config.js from /private/tmp/stylelint-issue-7322/node_modules/cosmiconfig/dist/loaders.js not supported.
Instead change the require of stylelint.config.js in /private/tmp/stylelint-issue-7322/node_modules/cosmiconfig/dist/loaders.js to a dynamic import() which is available in all CommonJS modules.
    at module.exports (/private/tmp/stylelint-issue-7322/node_modules/import-fresh/index.js:32:59)
    at Object.loadJsSync (/private/tmp/stylelint-issue-7322/node_modules/cosmiconfig/dist/loaders.js:17:12)
    at .js (file:///private/tmp/stylelint-issue-7322/node_modules/stylelint/lib/getConfigForFile.mjs:56:76)
    at #loadConfiguration (/private/tmp/stylelint-issue-7322/node_modules/cosmiconfig/dist/Explorer.js:92:30)
[...]

PR #7334, opened 2023-11-14
Improve config file description in "Configuring" user guide

ybiquitous:

Ref #7322

The [Prettier document](https://prettier.io/docs/en/configuration) was referenced.

### DIFF
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -2,14 +2,46 @@
 
 Stylelint expects a configuration object, and looks for one in a:
 
+- `stylelint.config.js` or `.stylelintrc.js` file
+  - Which module system to use depends on your [default module system configuration](https://nodejs.org/api/packages.html#determining-module-system) for Node.js (e.g., `"type": "module"` in `package.json`).
+- `stylelint.config.mjs` or `.stylelintrc.mjs` file using `export default` (ES module)
+- `stylelint.config.cjs` or `.stylelintrc.cjs` file using `module.exports` (CommonJS)
+- `.stylelintrc.json`, `.stylelintrc.yml`, or `.stylelintrc.yaml` file
+- `.stylelintrc` file in JSON or YAML format
+  - We recommend adding an extension (e.g., `.json`) to help your editor provide syntax checking and highlighting.
 - `stylelint` property in `package.json`
-- `.stylelintrc` file
-- `.stylelintrc.{cjs,js,json,yaml,yml}` file
-- `stylelint.config.{cjs,mjs,js}` file
+
+ES module example:
+
+```js
+export default {
+  rules: {
+    "block-no-empty": true
+  }
+};
+```
+
+CommonJS example:
+
+```js
+module.exports = {
+  rules: {
+    "block-no-empty": true
+  }
+};
+```
+
+JSON example:
+
+```json
+{
+  "rules": {
+    "block-no-empty": true
+  }
+}
+```
 
 Starting from the current working directory, Stylelint stops searching when one of these is found. Alternatively, you can use the [`--config` or `configFile` option](options.md#configfile) to short-circuit the search.
-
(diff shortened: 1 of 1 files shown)


docs/user-guide/configure.md
```diff
@@ -2,14 +2,46 @@
...
+  - Which module system to use depends on your [default module system configuration](https://nodejs.org/api/packages.html#determining-module-system) for Node.js (e.g., `"type": "module"` in `package.json`).
+- `stylelint.config.mjs` or `.stylelintrc.mjs` file using `export default` (ES module)
+- `stylelint.config.cjs` or `.stylelintrc.cjs` file using `module.exports` (CommonJS)
+- `.stylelintrc.json`, `.stylelintrc.yml`, or `.stylelintrc.yaml` file
+- `.stylelintrc` file in JSON or YAML format
+  - We recommend adding an extension (e.g., `.json`) to help your editor provide syntax checking and highlighting.
 - `stylelint` property in `package.json`
-- `.stylelintrc` file
-- `.stylelintrc.{cjs,js,json,yaml,yml}` file
-- `stylelint.config.{cjs,mjs,js}` file
```
ybiquitous, line 8: [note]

- Move `stylelint.config.js` to the top because we are considering a flat configuration in a future major release
- Move `stylelint` property (in `package.json`) to the bottom because it's likely to be less used

docs/user-guide/configure.md
```diff
@@ -2,14 +2,46 @@
...
+
+```js
+module.exports = {
+  rules: {
+    "block-no-empty": true
+  }
+};
+```
 
-The `.stylelintrc` file (without extension) can be in JSON or YAML format. We recommend adding an extension to help your editor provide syntax checking and highlighting.
```
ybiquitous, line 12: [note] Move this description near the list item `.stylelintrc` above.

docs/user-guide/configure.md
```diff
@@ -2,14 +2,46 @@
...
+JSON example:
+
+```json
+{
+  "rules": {
+    "block-no-empty": true
+  }
+}
+```
```
ybiquitous, line 42: [ask] Are these examples redundant? Feel free to give me feedback.

newives: Sample code is so important, it can clearly express how users need to migrate.

In addition to rules, can you add additional examples such as `extends`, `plugins`, `overrides`, `ignoreFiles`?
Or are these the same APIs as before?

ybiquitous: Thanks for the feedback.

> In addition to rules, can you add additional examples such as `extends`, `plugins`, `overrides`, `ignoreFiles`?
> Or are these the same APIs as before?

Yes, the same as before. I think keeping these examples simple in the top section may be good since each section already has a more detailed example.

newives: OK, I understand. Thank you for your reply.

docs/user-guide/configure.md
```diff
@@ -2,14 +2,46 @@
...
+
+Starting from the current working directory, Stylelint stops searching when one of these is found. Alternatively, you can use the [`--config` or `configFile` option](options.md#configfile) to short-circuit the search.
```
Mouvedia, line 44: > [`--config` or `configFile` option](options.md#configfile)

What about

```md
[`--config`](cli.md#--config--c) or [`configFile` option](options.md#configfile)
```
instead?

ybiquitous: Nice catch. Fixed with 1d444b573.

jeddy3 (approved):
LGTM, thank you.

rakleed:
@jeddy3 @ybiquitous when will these changes appear on the site? As I understand it, they do not automatically appear after a commit.

ybiquitous:
@rakleed The changes are unavailable yet on our website until the 16.0.0 release (see https://github.com/stylelint/stylelint.io/pull/359). Please check the documents on GitHub. 🙏🏼

Issue #7338, opened 2023-11-14
Add tally of fixable problems to string formatter


rakleed:
### What is the problem you're trying to solve?

There is currently no way to know how many problems will be fixed automatically with the `--fix` flag. When calling `stylelint`, only the number of problems found will be shown.

![CleanShot 2023-11-13 at 06 19 06](https://github.com/stylelint/stylelint/assets/19418601/044b02ae-c299-470e-8503-b157e309bde6)

### What solution would you like to see?

When calling `stylelint`, show how many problems can be fixed with the `--fix` flag, as ESLint does:

![CleanShot 2023-11-13 at 06 47 50](https://github.com/stylelint/stylelint/assets/19418601/20366b57-2e51-4027-8c5b-6e4f9ce27b56)

ybiquitous:
@rakleed Thanks for the proposal. However, the `verbose` formatter shows the info:

<img width="529" alt="image" src="https://github.com/stylelint/stylelint/assets/473530/d956dff5-065f-4b25-86de-4aee7ef9a08f">

It may be helpful if the default formatter also shows the same info.

jeddy3:
> It may be helpful if the default formatter also shows the same info.

SGTM. 

Let's move the 'You may fix some problems with the "--fix" option.' from the `verbose` formatter into the `string` one, rewrite it to include a tally and change the colour to white as it's informational.

For example, the string formatter:

```console
% npx stylelint "**/*.css"
src/components.css
 10:10  ✖  Unexpected unknown value "1" for property "gap"  declaration-property-value-no-unknown

74 problems (24 errors, 50 warnings)
10 errors and 3 warnings potentially fixable with the "--fix" option.
```

For example, the verbose formatter:

```console
% npx stylelint "**/*.css"
src/components.css
 10:10  ✖  Unexpected unknown value "1" for property "gap"  declaration-property-value-no-unknown

74 problems (24 errors, 50 warnings)
10 errors and 3 warnings potentially fixable with the "--fix" option.

1 sources checked
 /Users/jeddy3/components.css

10 errors found
 font-family-name-quotes: 5 (maybe fixable)
```

Let's only show numbers greater than one, e.g.

- 0 fixable errors & 3 fixable warnings = "3 warnings potentially fixable with the "--fix" option."
- 10 fixable errors & 0 fixable warnings = "10 errors potentially fixable with the "--fix" option."
- 0 fixable errors & 0 fixable warnings = ""

[...]

PR #7341, opened 2023-11-15
Document `fixable: true` meta

jeddy3:

It's a documentation fix related to https://github.com/stylelint/stylelint/issues/7338

Although already used by the `verbose` formatter, the `fixable` `meta` property becomes more important if we tally the fixable problems in the default `string` formatter, especially for plugin authors.

### DIFF
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -190,7 +190,7 @@ Add `fixable: true` to the rule's `meta`:
 
 ```diff js
 const meta = {
-	url: /* .. */,
+  url: /* .. */,
 + fixable: true,
 };
 ```
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -186,7 +186,7 @@ If the rule has autofix use:
 
 Depending on the rule, it might be possible to automatically fix the rule's problems by mutating the PostCSS AST (Abstract Syntax Tree) using the [PostCSS API](http://api.postcss.org/).
 
-Add `fixable: true` to the rule's `meta`:
+Set `meta.fixable = true` to the rule:
 
 ```diff js
 const meta = {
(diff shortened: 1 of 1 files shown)


docs/developer-guide/rules.md
Mouvedia, line 1: l209
> `context` is an object which could have two properties:

I count 3 properties.

Mouvedia (changes requested):

docs/developer-guide/rules.md
```diff
@@ -186,17 +186,27 @@ If the rule has autofix use:
 
 Depending on the rule, it might be possible to automatically fix the rule's problems by mutating the PostCSS AST (Abstract Syntax Tree) using the [PostCSS API](http://api.postcss.org/).
 
+Add `fixable: true` to the rule's `meta`:
+
+```diff js
+const meta = {
+	url: /* .. */,
```
ybiquitous, line 193: [suggest] Incorrect tab indentation:

```suggestion
  url: /* .. */,
```

docs/developer-guide/rules.md
```diff
@@ -186,17 +186,27 @@ If the rule has autofix use:
 
 Depending on the rule, it might be possible to automatically fix the rule's problems by mutating the PostCSS AST (Abstract Syntax Tree) using the [PostCSS API](http://api.postcss.org/).
 
+Add `fixable: true` to the rule's `meta`:
```
ybiquitous, line 189: [suggest] Minor. The following seems a bit more intuitive to me:

```suggestion
Set `meta.fixable = true` to the rule:
```

But it's not a strong opinion. No problems with leaving it as-is. 🙆🏼‍♂️

ybiquitous (commented):
Good addition 👍🏼

ybiquitous (approved):
Thanks. LGTM 👍🏼

Mouvedia (approved):

PR #7343, opened 2023-11-15
Add support for ESM custom formatters

### LINKED ISSUE #7267: Support custom formatter for ESM

Currently, we support only a custom formatter written in CommonJS (CJS). But, it would be useful if people could also specify a custom formatter in ESM.

https://github.com/stylelint/stylelint/blob/958bf369127f867747c613ec9c7025344e220eed/lib/cli.mjs#L396-L400

Related to #6930

I suggest supporting a custom formatter written in ESM. This means we need to change the current *requiring* logic.

ybiquitous:

Closes #7267

This PR adds the support for ESM custom formatters. A custom formatter can be specified via CLI.

Also, this change refactors code as below to reduce the published size:
- move the logic in `lib/resolveCustomFormatter.{cjs,mjs}` files into `cli.mjs`
- remove the files


lib/__tests__/cli.test.mjs
```diff
@@ -17,13 +17,17 @@ import {
...
+import customFormatter from './fixtures/custom-formatter.mjs';
+
 import cli, { buildCLI } from '../cli.mjs';
 
 const pkg = readJSONFile(new URL('../../package.json', import.meta.url));
 
 const fixturesPath = (...elems) =>
 	replaceBackslashes(path.join(fileURLToPath(new URL('./fixtures', import.meta.url)), ...elems));
 
+jest.unstable_mockModule('stylelint-test-custom-formatter', () => ({ default: customFormatter }));
```
ybiquitous, line 29: [note] See https://jestjs.io/docs/ecmascript-modules#module-mocking-in-esm

jeddy3 (changes requested):
Excellent, thank you.

Let's add an entry to the migration guide too. (It's okay if there's repetition at this stage as we can give the guide a clean-up before releasing.)

ybiquitous:
@jeddy3 Thanks for the reminder. I've updated the guide via d896cd71c.

If you have any concerns, feel free to edit the guide directly. 🙏🏼

jeddy3 (approved):
LGTM, thank you.

PR #7344, opened 2023-11-15
Fix flaky system test case for CommonJS

ybiquitous:

None. Just test fix.

This change tries fixing a flaky system test case for CommonJS.
The case sometimes fails only on Node.js 20 and Windows OS. See the example:

```
FAIL system-tests/005/commonjs.test.cjs (8.098 s)
  ● CommonJS API and config

    thrown: "Exceeded timeout of 5000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

       9 | const testFn = Number(nodeMajorVersion) >= 20 ? test : test.skip;
      10 |
    > 11 | testFn('CommonJS API and config', async () => {
         | ^
      12 | 	const result = await stylelint.lint({
      13 | 		files: [caseFilePath('stylesheet.css')],
      14 | 		configFile: caseFilePath('config.cjs'),

      at Object.<anonymous> (system-tests/005/commonjs.test.cjs:11:1)
```

https://github.com/stylelint/stylelint/actions/runs/6878175633/job/18707274491#step:8:645

### DIFF
--- a/system-tests/005/commonjs.test.cjs
+++ b/system-tests/005/commonjs.test.cjs
@@ -8,14 +8,18 @@ const stylelint = require('../../lib/index.cjs');
 const [nodeMajorVersion] = process.versions.node.split('.', 1);
 const testFn = Number(nodeMajorVersion) >= 20 ? test : test.skip;
 
-testFn('CommonJS API and config', async () => {
-	const result = await stylelint.lint({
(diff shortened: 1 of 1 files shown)


jeddy3 (approved):
LGTM, thank you.

PR #7351, opened 2023-11-20
Add support for ESM custom syntaxes

### LINKED ISSUE #7269: Support custom syntax for ESM

Currently, we support a custom syntax module only for CommonJS (CJS). It would be useful if we could support ESM as well.

https://github.com/stylelint/stylelint/blob/958bf369127f867747c613ec9c7025344e220eed/lib/getPostcssResult.js#L74

Related to #6930, #7267

I suggest supporting a custom syntax written in ESM. This means we need to change the current requiring logic.

ybiquitous:

Closes #7269

The dynamic `import()` supports both ESM and CJS.


lib/getPostcssResult.mjs
```diff
@@ -65,22 +65,23 @@ export default async function getPostcssResult(stylelint, options = {}) {
...
-			resolved = require(customSyntaxLookup);
+			resolved = await import(customSyntaxLookup);
+			resolved = resolved.default ?? resolved;
 		} catch (error) {
 			if (
 				error &&
 				typeof error === 'object' &&
 				'code' in error &&
-				error.code === 'MODULE_NOT_FOUND' &&
+				(error.code === 'MODULE_NOT_FOUND' || error.code === 'ERR_MODULE_NOT_FOUND') &&
```
ybiquitous, line 84: [note] The path of `error.code === 'ERR_MODULE_NOT_FOUND'` is not covered in tests for some reason. I guess Babel under Jest may transform the `import()` code, but I'm not sure.

docs/user-guide/options.md
```diff
@@ -70,7 +70,7 @@ This option should be a string that resolves to a JS module that exports a [Post
 
 If you want to lint two or more different languages, you can combine `customSyntax` with the [`overrides`](./configure.md#overrides) configuration property.
 
-Using the Node.js API, the `customSyntax` option can also accept a [Syntax object](https://postcss.org/docs/how-to-write-custom-syntax#syntax). Stylelint treats the `parse` property as a required value.
+Using the Node.js API, the `customSyntax` option can also accept a [Syntax object](https://postcss.org/docs/how-to-write-custom-syntax#syntax). Stylelint treats the `parse` property as a required value. Also, Stylelint supports syntax modules written in ESM.
```
Mouvedia, line 73: > Stylelint treats the `parse` property as a required value

Add `in that case`?

lib/getPostcssResult.mjs
Mouvedia, line 84: minor

Add above
```
// TODO remove MODULE_NOT_FOUND condition @v17
// i.e CJS specific path will be unnecessary
```

see https://nodejs.org/docs/latest-v20.x/api/all.html#all_errors_module_not_found

jeddy3 (approved):
Nice work! LGTM, thank you.

docs/user-guide/options.md
ybiquitous: Hum, adding `in that case` might confuse, instead. But I'm open to a better description.

Of course, we can improve it later.

PR #7353, opened 2023-11-21
Deprecate CommonJS Node.js API

### LINKED ISSUE #7340: Deprecate CommonJS plugins

For 16.0.0, we've migrated our codebase to ESM (with tool-based hybrid CommonJS support) and added support for them.

We intend to be pure ESM for our next major release to update our dependencies, reduce our maintenance load and move with the times. We should deprecate support for CommonJS plugins in 16.0.0 to communicate our intent so that plugin authors are encouraged to migrate their plugins to ESM after 16.0.0 and before 17.0.0.

Ref: https://github.com/stylelint/stylelint/pull/7339#discussion_r1393908910

The minimum would be changelog and migration guide entries, e.g.:

---
### Deprecated support for CommonJS plugins

We've deprecated support for CommonJS plugins, and we'll remove support for them in our next major release. 

You should update how your plugin imports and exports:

```diff js
-const stylelint = require('stylelint');
+import stylelint from 'stylelint';

/* .. */

-module.exports = createPlugin(ruleName, ruleFunction);
+export default createPlugin(ruleName, ruleFunction);
```

You should also:

- Add `"type": "module"` to your package.json.
- Replace `"main": "index.js"` with `"exports": "./index.js"` in your package.json.
- Update the `"engines"` field in package.json to Node.js 18: `"node": ">=18"`.
- Remove `'use strict';` from all JavaScript files.
- Use only full relative file paths for imports: `import x from '.';` → `import x from './index.js';`.
- Use the [`node:` protocol](https://nodejs.org/api/esm.html#esm_node_imports) for Node.js built-in imports.

---

Ideally, we'd also generate a deprecation warning (that respects the [`quietDeprecationWarnings`](https://stylelint.io/user-guide/options#quietdeprecationwarnings) option) when commonjs plugins are used. I'm not sure if this is possible, though?

ybiquitous:

Closes #7340

This change deprecates the use of CommonJS Node.js API and updates the migration guide to 16.0.0.

Stylelint will show the deprecation warning when using CommonJS API. This deprecation code is generated by a Rollup custom plugin.

<details><summary>Outdated description of STYLELINT_CJS_IGNORE_WARNING</summary>
<p>

Examples:

```sh-session
$ node -e 'import("stylelint")'
(no output)

$ node -e 'require("stylelint")'
The CJS build of Stylelint's Node.js API is deprecated. See https://stylelint.io/migration-guide/to-16

$ STYLELINT_CJS_IGNORE_WARNING=true node -e 'require("stylelint")'
(no output)
```

### Why not use `quietDeprecationWarnings`

It's difficult to write a Rollup plugin code using the `quietDeprecationWarnings` option in this case. But I'm open to feedback for a better solution.

</p>
</details>


jeddy3 (commented):
@ybiquitous Thank you for tackling this!

Injecting the deprecation warning into the CJS code using rollup is smart.

Is there a way to silence the warning as a CJS plugin author when using Jest?

```
console.warn
      The CJS build of Stylelint's Node.js API is deprecated. See https://stylelint.io/migration-guide/to-16

      17 |
      18 |      if (!options.quietDeprecationWarnings) {
    > 19 |              console.warn("The CJS build of Stylelint's Node.js API is deprecated. See https://stylelint.io/migration-guide/to-16");
         |                      ^
      20 |      }
      21 |
      22 |      const cwd = options.cwd || process.cwd();

      at createStylelint (../stylelint/lib/createStylelint.cjs:19:11)
      at standalone (../stylelint/lib/standalone.cjs:94:20)
      at Object.<anonymous> (node_modules/jest-preset-stylelint/getTestRule.js:43:21)
```

docs/user-guide/node-api.md
```diff
@@ -96,118 +94,101 @@ It resolves with an object (see [The returned promise](#the-returned-promise)) t
 As `config` contains no relative paths for `extends` or `plugins`, you do not have to use `configBasedir`:
 
 ```js
-stylelint
-  .lint({
+try {
+  const result = await stylelint.lint({
```
Mouvedia, line 98: minor
suggestion

IIRC top-level await is not supported by node in CJS modules.
Since this version will support both, should we have a compatible example to permit easy copy paste?
i.e. use `then`/`catch` instead (like the previous example)

lib/createStylelint.mjs
```diff
@@ -12,6 +12,7 @@ const STOP_DIR = IS_TEST ? process.cwd() : undefined;
  * @type {import('stylelint').PublicApi['_createLinter']}
  */
 export default function createStylelint(options = {}) {
+	// [INSERT HERE] CJS deprecation code
```
Mouvedia, line 15: major

Should we have one [here](https://github.com/stylelint/stylelint/blob/b0c7876d21a1900aaa3edfee6aa6e878bd4822e0/lib/augmentConfig.mjs#L324-L325)
```js
                // Handle either ES6 or CommonJS modules
                pluginImport = pluginImport.default || pluginImport;
```
as well?

docs/migration-guide/to-16.md
```diff
@@ -4,10 +4,53 @@ This release contains significant or breaking changes.
...
+#### How to migrate from CJS to ESM
+
+You should also:
+
+- add `"type": "module"` to `package.json`
+- add `"exports": "./index.js"` in `package.json`
+- update the `"engines"` field in `package.json` to Node.js 18: `"node": ">=18"`
+- remove `'use strict';` from all JavaScript files
+- use only full relative file paths for imports: `import x from '.';` → `import x from './index.js';`
+- use the `node:` protocol for Node.js built-in imports
```
Mouvedia, line 50: suggestion
medium

```md
You must:

- add `"type": "module"` to `package.json` if you are not using `.mjs` extension for your files
- use only full relative file paths for imports
   e.g. `import x from '.';` → `import x from './index.js';`

You may:

- have to update the `"engines"` field in `package.json` to Node.js 18: `"node": ">=18.12.0"`
- remove `'use strict';` from all JavaScript files
- add `"exports": "./index.js"` in `package.json`
- use the `node:` protocol for Node.js built-in imports
```

notice in particular this change : ">=18.0.0" to ">=18.12.0" (match ours)

ybiquitous:
> Is there a way to silence the warning as a CJS plugin author when using Jest?

Ah, good catch. We also need to adjust [`jest-preset-stylelint`](https://github.com/stylelint/jest-preset-stylelint) toward ESM. 😅 

At this time, using `loadLint` with dynamic `import()` seems good:

```js
// jest.setup.js
global.testRule = getTestRule({ loadLint: () => import("stylelint").then(m => m.default.lint) });

// or a test file
testRule({
  loadLint: () => import("stylelint").then(m => m.default.lint),
  ruleName: 'block-no-empty',
  config: true,
  accept: [/*...*/],
  reject: [/*...*/],
});
```

Otherwise, changing the [default `loadLint` function](https://github.com/stylelint/jest-preset-stylelint/blob/9693b30a6809db4e141877a9330a53a1e0fbccda/getTestRule.js#L14) may be better:

```diff
- schema.loadLint || options.loadLint || (() => Promise.resolve(require('stylelint').lint));
+ schema.loadLint || options.loadLint || (() => import('stylelint').then(m => m.default.lint));
```

although this may be a breaking change.

ybiquitous:
@jeddy3 I've opened PR stylelint/jest-preset-stylelint#87 as I suggested on <https://github.com/stylelint/stylelint/pull/7353#issuecomment-1821916255>. If this new `jest-preset-stylelint` would be available, the `console.warn()` warning should not be shown by default (since the ESM version would be loaded).

docs/migration-guide/to-16.md
ybiquitous: Sounds good. Updated via ef77af64c.

docs/user-guide/node-api.md
ybiquitous: Yes, top-level await is unsupported in CJS, but I don't think we have to stick to whether top-level await or not because these examples are just code snippets.

Personally, I don't prefer `then`/`catch` so much since there are many nestings.

lib/createStylelint.mjs
ybiquitous: Your concern makes sense, and we should show the warning when importing a CJS plugin. But unfortunately, we cannot distinguish ESM and CJS by `pluginImport.default`: both modules should have the `default` property.

Is there any good solution?

docs/user-guide/node-api.md
Mouvedia: I guess, it can be considered a subtle nudge to move towards ESM before v17.

lib/createStylelint.mjs
Mouvedia: > Is there any good solution?

Since the path for CJS uses files with `.cjs` extensions, isn't it just a matter of adding the `if` only for `augmenConfig.cjs`?

ybiquitous: `.cjs` checks sounds good. Added it with 6434e98ce.

For `__esModule` check, it's non-standard, right? I'm concerned about supporting non-standard flags.

Mouvedia: ref: nodejs/node#40891

docs/user-guide/node-api.md
```diff
@@ -96,118 +94,101 @@ It resolves with an object (see [The returned promise](#the-returned-promise)) t
...
+});
+
+// do things with result.report
 ```
 
+The report will be available as the value of the `report` property in the returned object.
+
 ### Example D
 
-Using your own custom formatter function and parse `.scss` source files:
```
Mouvedia, line 148: nit

```
Using your own custom formatter function:
```

lib/__tests__/plugins.test.mjs
```diff
@@ -360,3 +370,30 @@ describe('loading a plugin from options.cwd', () => {
...
+			expect.stringContaining('plugin-warn-about-foo.cjs" is a CJS plugin and has been deprecated'),
+		);
+	});
+
+	it('does not show the warning when `quietDeprecationWarnings` is enabled', async () => {
+		await runLint({ quietDeprecationWarnings: true });
+
+		expect(consoleWarn).not.toHaveBeenCalled();
+	});
+});
```
Mouvedia, line 399: questions

Can we use a commonjs `.js` file fixture?
If it fails to load, can we add a test that checks the error?
If it succeeds, then the compatibility should be mentioned in the documentation.

Mouvedia (approved):
LGTM

lib/__tests__/plugins.test.mjs
ybiquitous: Yes, we can add a test case importing a CJS plugin with `.js`. This case should succeed like `.cjs` cases.

For "compatibility" you said, do you have any specific suggestion in the documentation?

Mouvedia: It's reassuring to know that the cjs plugins will remain compatible in a mixed environment.

> do you have any specific suggestion in the documentation?

Not really :)

lib/augmentConfig.mjs
```diff
@@ -317,6 +318,13 @@ async function addPluginFunctions(config) {
 
 		if (typeof pluginLookup === 'string') {
 			pluginImport = await import(pluginLookup);
+
+			// NOTE: This '.cjs' check is limited. Some CJS plugins may have the '.js' extension.
+			if (!quietDeprecationWarnings && pluginLookup.endsWith('.cjs')) {
+				console.warn(
+					`"${pluginLookup}" is a CJS plugin and has been deprecated since 16.0.0. See https://stylelint.io/migration-guide/to-16`,
```
jeddy3, line 325:
```suggestion
					`"CommonJS plugins are deprecated ("${pluginLookup}"). See https://stylelint.io/migration-guide/to-16`,
```

docs/migration-guide/to-16.md
```diff
@@ -4,10 +4,56 @@ This release contains significant or breaking changes.
 
 ## Significant changes
 
+- added support for ESM Node.js API and deprecated CJS API
```
jeddy3, line 7:
```suggestion
- added support for ESM Node.js API and deprecated CommonJS API
```

jeddy3 (changes requested):
@ybiquitous Thank you! 

I've requested some minor copy changes.

(I was offline for the weekend and looping back to this PR now.)

rollup.config.mjs
```diff
@@ -39,7 +43,32 @@ export default inputFiles.map((input) => {
...
+		name: 'stylelint-add-warning-for-cjs',
+		generateBundle(_options, bundle) {
+			const asset = bundle['lib/createStylelint.cjs'];
+
+			if (asset?.type === 'chunk') {
+				asset.code = asset.code.replace(
+					'	// [INSERT HERE] CJS deprecation code',
+					`
+	if (!options.quietDeprecationWarnings) {
+		console.warn("The CJS build of Stylelint's Node.js API is deprecated. See https://stylelint.io/migration-guide/to-16");
```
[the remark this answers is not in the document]
ybiquitous: Make sense. Fixed with e386c1609.

jeddy3 (approved):
LGTM, thank you!

ybiquitous:
Thanks for the review! 😄

PR #7357, opened 2023-11-27
Change default syntax behaviour to always use safe-parser with `fix` regardless of extension

### LINKED ISSUE #7116: Change default syntax behaviour to always use safe-parser with `fix` regardless of extension

Currently the `cssSyntax()` helper only allows three file extensions to be treated as standard CSS without a warning:

https://github.com/stylelint/stylelint/blob/247c94c65c58eec6e5573fb35ff6ab7d45356bac/lib/getPostcssResult.js#L127

I'm writing what is essentially standard CSS, but with a `.mcss` extension for various vite-related reasons.

I can specify a `customSyntax` that copies what `cssSyntax()` outputs, but that file doesn't have access to `stylelint._options` and can't make choices about using the standard postcss or safe postcss parser like `cssSyntax()` can & does when `--fix` is specified:

https://github.com/stylelint/stylelint/blob/247c94c65c58eec6e5573fb35ff6ab7d45356bac/lib/getPostcssResult.js#L135-L141

Add a config option to extend the list of file extensions treated as CSS, or pass the stylelint options to custom syntax creators somehow (but this would be a breaking change afaict).

ybiquitous:

Closes #7116

We no longer maintain the special file extension list for the `customSyntax` warning and the `fix` option. This means that `postcss-safe-parser` is always applied when autofix.


lib/__tests__/fixtures/syntax_error.sass
ybiquitous, line 1: [note] Since `.sass` files are no longer excluded from the safe-parse autofix, the autofix test case about syntaxes now fails. See the example below. The Sass code `a`  with syntax error cannot be fixed:

```js
require('postcss-safe-parser')('a{').toString()
//=> 'a{}'

require('postcss-safe-parser')('a').toString()
//=> 'a'
```

Thus, removing this `.sass` file makes it easier to resolve the test failure.

jeddy3 (approved):
Perfect. LGTM, thank you!

jeddy3:
We can do another preprelease after we merge this one.

PR #7366, opened 2023-12-02
Add lightness-notation

### LINKED ISSUE #7359: Add `lightness-notation`

I would like to specify the type of value of [lightness](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/oklch#l) component

A rule with two options: "number" | "percent". 

**"number":**
The following patterns are considered problems:
```css
a {
  color: oklch(70% 0.2 10);
}
```

The following patterns are NOT considered problems:
```css
a {
  color: oklch(0.7 0.2 10);
}
```

**"percent":**
The following patterns are considered problems:
```css
a {
  color: oklch(0.7 0.2 10);
}
```

The following patterns are NOT considered problems:
```css
a {
  color: oklch(70% 0.2 10);
}
```

fpetrakov:
Closes #7359 

No, it's self explanatory.


lib/rules/lightness-notation/index.cjs
```diff
@@ -93,28 +102,44 @@ const rule = (primary, _secondaryOptions, context) => {
...
  */
-function asPercentage(value) {
-	const num = parseFloat(value) * 100;
+function asPercentage(value, func) {
+	let num = parseFloat(value);
 
-	if (Number.isInteger(num)) return `${num}%`;
+	if (LIGHTNESS_ZERO_TO_HUNDRED_RANGE_FUNCS.has(func)) {
+		return `${num}%`;
+	}
```
romainmenke, line 112: I think this can be omitted?

lib/rules/lightness-notation/index.cjs
```diff
@@ -93,28 +102,44 @@ const rule = (primary, _secondaryOptions, context) => {
...
 	return `${roundToNumberOfDigits(num, value)}%`;
 }
 
 /**
  * @param {string} value
+ * @param {import('postcss-value-parser').FunctionNode['value']} func
  */
-function asNumber(value) {
+function asNumber(value, func) {
 	const dimension = valueParser.unit(value);
```
romainmenke, line 130: Why not reuse the parsed number that is returned here?

I think it's better to use that, than to have `parseFloat(value)` and `value.slice(0, -1)`.

romainmenke (commented):
Thank you for the fixed @fpetrakov

lib/rules/lightness-notation/index.cjs
romainmenke: thank you for explaining

lib/rules/lightness-notation/index.mjs
```diff
@@ -0,0 +1,183 @@
...
+		root.walkDecls((decl) => {
+			if (!HAS_LIGHTNESS_COLOR_FUNC.test(decl.value)) return;
+
+			let needsFix = false;
+			const parsedValue = valueParser(getDeclarationValue(decl));
+
+			parsedValue.walk((node) => {
+				if (node.type !== 'function') return;
+
+				if (!LIGHTNESS_FUNCS.has(node.value.toLowerCase())) return;
```
ybiquitous, line 50: [suggest] Can we move the `functionName` declaration at [L66](https://github.com/stylelint/stylelint/pull/7366/files#diff-e55c9791f54478934c2ad8ddd81e2913441c88b5ba16638f72d10ba02214c11aR66) to here?

```suggestion
				const functionName = node.value.toLowerCase();

				if (!LIGHTNESS_FUNCS.has(functionName)) return;
```

lib/rules/lightness-notation/index.mjs
```diff
@@ -0,0 +1,183 @@
...
+ */
+function findLightness(node) {
+	const args = node.nodes.filter(({ type }) => type === 'word' || type === 'function');
+	const value = node.value.toLowerCase();
+
+	if (LIGHTNESS_FUNCS.has(value)) {
+		return args[0];
+	}
+
+	return undefined;
```
ybiquitous, line 161: [suggest] How about simplifying `findLightness()` like this? `LIGHTNESS_FUNCS.has()` check seems duplicate at [L50](https://github.com/stylelint/stylelint/pull/7366/files#diff-e55c9791f54478934c2ad8ddd81e2913441c88b5ba16638f72d10ba02214c11aR50).

```suggestion
function findLightness({ nodes }) {
	return nodes.find(({ type }) => type === 'word' || type === 'function');
```

lib/rules/lightness-notation/index.mjs
```diff
@@ -0,0 +1,183 @@
...
+
+				if (!LIGHTNESS_FUNCS.has(node.value.toLowerCase())) return;
+
+				const lightness = findLightness(node);
+
+				if (!lightness) return;
+
+				const { value } = lightness;
+
+				if (!isStandardSyntaxValue(value)) return;
```
ybiquitous, line 58: [suggest] I'm concerned about redundant calls of `valueParser.unit()`. How about reducing the calls like this?

```suggestion
				if (!isStandardSyntaxValue(value)) return;

				const dimension = valueParser.unit(value);

				if (!dimension) return;

				const isPercentage = dimension.unit === '%';
				const isNumber = dimension.unit === '';
```

lib/rules/lightness-notation/README.md
ybiquitous, line 1: [suggest] How about adding supported functions like `oklch` to this README?

ybiquitous (changes requested):
@fpetrakov Thanks for creating the PR. I don't believe there are major problems. 👍🏼 

Can you take a look at my suggestion comments? If you have rationale, you don't need to address all.

ybiquitous:
Note: I'd like to merge this PR after the 16.0.0 release (#7374).

fpetrakov:
@ybiquitous Thanks for reviewing the PR. All of yours suggestions are sensible, so I will fix them in the coming days

lib/rules/lightness-notation/index.mjs
```diff
@@ -0,0 +1,183 @@
...
+	url: 'https://stylelint.io/user-guide/rules/lightness-notation',
+	fixable: true,
+};
+
+const LIGHTNESS_ZERO_TO_ONE_RANGE_FUNCS = new Set(['oklch', 'oklab']);
+const LIGHTNESS_ZERO_TO_HUNDRED_RANGE_FUNCS = new Set(['lch', 'lab']);
+const LIGHTNESS_FUNCS = new Set([
+	...LIGHTNESS_ZERO_TO_ONE_RANGE_FUNCS,
+	...LIGHTNESS_ZERO_TO_HUNDRED_RANGE_FUNCS,
+]);
```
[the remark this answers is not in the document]
Mouvedia: https://github.com/stylelint/stylelint/blob/7e54fd597230170c30ad30303d0d41fc2b845396/lib/reference/functions.mjs#L15 could be constructed from 2 sets.
IMHO the granularity of ZERO_TO_ONE vs ZERO_TO_HUNDRED is localized and not worth exposing outside of this rule.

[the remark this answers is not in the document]
ybiquitous: Make sense.

lib/rules/lightness-notation/index.mjs
```diff
@@ -0,0 +1,163 @@
...
+
+				if (primary === 'percentage' && isPercentage) return;
+
+				if (primary === 'number' && isNumber) return;
+
+				const fixed =
+					primary === 'percentage'
+						? asPercentage(value, functionName)
+						: asNumber(value, functionName);
+				const unfixed = value;
```
Mouvedia, line 78: minor

You could instead rename it at l57.
i.e. `const { value: unfixed } = lightness;`

lib/utils/dimensionUnitIsEmpty.cjs
ybiquitous, line 1: Can you remove this file, too?

ybiquitous (approved):
Thanks, LGTM. I'll revisit this after the 16.0.0 release.

Also, we need to determine a value (maybe `percentage`?) added in the standard config before publishing this new rule.

fpetrakov:
> Also, we need to determine a value (maybe `percentage`?) added in the standard config before publishing this new rule.

I agree on `percentage`

ybiquitous:
@fpetrakov Can you resolve the conflicts raised by the 16.0.0 release?

lib/rules/lightness-notation/README.md
```diff
@@ -0,0 +1,116 @@
...
+<!-- prettier-ignore -->
+```css
+a { color: oklab(0.86 0.2 154) }
+```
+
+<!-- prettier-ignore -->
+```css
+a { color: lch(85 0.17 88) }
+```
+
```
jeddy3, line 42:
```suggestion
```

We tend to keep our README examples as small as possible. Two should be enough: one showing a 0-1 percentage range and the other a 1-100 range.

lib/rules/lightness-notation/README.md
```diff
@@ -0,0 +1,116 @@
...
+<!-- prettier-ignore -->
+```css
+a { color: oklab(86% 0.2 154) }
+```
+
+<!-- prettier-ignore -->
+```css
+a { color: lch(85% 0.17 88) }
+```
```
jeddy3, line 63:
```suggestion
```

lib/rules/lightness-notation/README.md
```diff
@@ -0,0 +1,116 @@
...
```
jeddy3, line 89:
```suggestion
```

lib/rules/lightness-notation/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+<!-- prettier-ignore -->
+```css
+a { color: oklab(0.86 0.2 154) }
+```
+
+<!-- prettier-ignore -->
+```css
+a { color: lch(85 0.17 88) }
+```
```
jeddy3, line 111:
```suggestion
```

jeddy3 (changes requested):
Thank you! It's looking great.

I've requested a change to the README, otherwise LGTM.

ybiquitous (approved):
Thanks, LGTM 👍🏼 

One more request, can you add a changelog entry via `npx changeset add`?
Ref https://github.com/stylelint/stylelint/blob/e28036e8c6121e2293a57986998af941ef2acc01/.changeset/early-jobs-hunt.md?plain=1

lib/rules/lightness-notation/__tests__/index.mjs
Mouvedia, line 1: minor

Maybe add an example using `deg`.

Mouvedia (approved):
LGTM

There might be potential futur refactors seeing the similarities between the 2 rules.

jeddy3 (approved):
LGTM, thank you!

(The rule falls neatly into [the wish](https://github.com/stylelint/stylelint/issues/7396#issue-2035832595) of supporting modern CSS in the `16.x`)

PR #7375, opened 2023-12-05
Refactor to replace `@changesets/apply-release-plan` patch

ybiquitous:

Ref #7367

With the Changesets package updates (#7367), the patch doesn't work.
So, this change replaces the patch with a custom script processing the changelog text.

You can locally check if this change works via:

```sh-session
$ export GITHUB_TOKEN=<your_personal_access_token>
$ npm run version
$ git diff CHANGELOG.md
```


jeddy3 (approved):
LGTM, thank you.

JounQin:
The patch does not work because the version is mismatched, I believe it can be fixed by changing the patch file name, of course some changes maybe needed. But I'm also fine with current solution without patch required.

PR #7377, opened 2023-12-08
Fix type definitions for CommonJS

remcohaszing:
This fixes the type definitions for CJS by leveraging a namespace. ESM type definitions were added, which export all types from the CJS types, and exports the exported stylelint object as default.

Some tweaks were made to the package.json exports, to resolve the types correctly for both CJS and ESM.

The type test was modified to fix CJS usage. Also a test was added for ESM usage. Stylelint was using tsconfig `paths`, which is a commonly misunderstood / misused feature that should rarely be used. This was removed, so TypeScript resolves the types based on package exports.

Refs [7309](https://github.com/stylelint/stylelint/pull/7309#issuecomment-1847067266)

https://arethetypeswrong.github.io/?p=stylelint@16.0.0 before this PR:

The result of ATTW showing a lot of errors

After uploading the result of `npm pack` after this PR:

The result of ATTW showing green checkmarks

---

This is best reviewed with whitespace changes disabled.


types/stylelint/index.d.mts
```diff
@@ -0,0 +1,46 @@
...
+	RuleMeta,
+	RuleOptions,
+	RuleOptionsPossible,
+	RuleSeverity,
+	Severity,
+	ShorthandProperties,
+	StylelintPostcssResult,
+	Utils,
+	Warning,
+	WarningOptions,
```
remcohaszing, line 45: Unfortunately `export * from './index.js'` doesn’t work if the referenced module uses `export =`.

ybiquitous:
@remcohaszing Thank you so much! This change looks good to me. 👍🏼 

In addition, the following type test fails, but do you know the cause?

```diff
--- a/types/stylelint/type-test.ts
+++ b/types/stylelint/type-test.ts
@@ -129,7 +129,9 @@ const testRule: RuleBase = (option) => {
 (testRule as Rule).primaryOptionArray = true;
 (testRule as Rule).meta = meta;
 
-stylelint.createPlugin(ruleName, testRule as Rule);
+const plugin = stylelint.createPlugin(ruleName, testRule as Rule);
+plugin.ruleName.toLowerCase();
+plugin.rule();
 
 messages.withNarrowedParam('should work');
```

```sh-session
$ npm -s run lint:types
types/stylelint/type-test.ts:133:8 - error TS2339: Property 'ruleName' does not exist on type 'Plugin'.
  Property 'ruleName' does not exist on type '{ default?: { ruleName: string; rule: Rule<any, any>; } | undefined; }'.

133 plugin.ruleName.toLowerCase();
           ~~~~~~~~

types/stylelint/type-test.ts:134:8 - error TS2339: Property 'rule' does not exist on type 'Plugin'.
  Property 'rule' does not exist on type '{ default?: { ruleName: string; rule: Rule<any, any>; } | undefined; }'.

134 plugin.rule();
           ~~~~


Found 2 errors in the same file, starting at: types/stylelint/type-test.ts:133
```

ybiquitous:
`{ default?: { ruleName: string; rule: Rule } }` in `Plugin` may be unnecessary. 🤔 

https://github.com/stylelint/stylelint/blob/bc28565824633060265bc2982ef45906380511f0/types/stylelint/index.d.ts#L78-L80

.changeset/large-jars-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fix type definitions
```
ybiquitous, line 5: [suggest] Thanks for creating the changelog entry. Can you update it for our convention like this?

```suggestion
Fixed: type definitions for CommonJS
```

ybiquitous (approved):
Got it. Thank you for the quick fix! 👍🏼

PR #7383, opened 2023-12-10
Fix `ERR_UNSUPPORTED_ESM_URL_SCHEME` error on Windows

### LINKED ISSUE #7382: Fix `ERR_UNSUPPORTED_ESM_URL_SCHEME` error on Windows

Running `stylelint` command on Windows fails after upgrading to 16.0.0. 
`stylelint src/**/*.css`

{
    "defaultSeverity": "warning",
    "plugins": ["stylelint-prettier"],
    "rules": {
        "prettier/prettier": true
    }
}

stylelint src/**/*.css

{
"stylelint": "16.0.1",
    "stylelint-config-prettier": "9.0.5",
    "stylelint-config-standard": "35.0.0",
    "stylelint-order": "6.0.4",
    "stylelint-prettier": "4.1.0",
}

Stylelint should run with no issues

### What actually happened?

I got this error

`Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'`

I did some research and it seems to be related to ESM. Not sure why it should fail though. My project is not ESM ready.

JounQin:

close #7382

No, it's self-explanatory.


jeddy3:
@JounQin Thank you for looking into this! 

Is it possible to add tests to cover this scenario?

And do we also need to use this approach for plugins and custom formatters?

JounQin:
> Is it possible to add tests to cover this scenario?

I'm not quite familiar with how to test such edge cases.

JounQin:
@ybiquitous I think the latest change https://github.com/stylelint/stylelint/pull/7383/commits/e547bf420f04ab8f7b076c562343465a8bb98ec8 requested makes more problems than it wants resolve.

I think we should just revert this commit, only `import()` will be affected by this issue.

lib/getPostcssResult.mjs
```diff
@@ -1,5 +1,8 @@
 import { readFile } from 'node:fs/promises';
```
JounQin, line 2: image

I don't think [`sort-imports`](https://eslint.org/docs/latest/rules/sort-imports) working correctly here.

ybiquitous (commented):
@JounQin Can we fix `getModulePath()` instead as <https://github.com/stylelint/stylelint/issues/7382#issuecomment-1848971544>?

JounQin:
@ybiquitous https://github.com/stylelint/stylelint/pull/7383#issuecomment-1848978667

ybiquitous:
> more problems than it wants to resolve

Just curious, what problem?

JounQin:
@ybiquitous https://github.com/stylelint/stylelint/actions/runs/7158350468?pr=7383

ybiquitous:
Ah, you didn't need to force-push since we couldn't find the failed CI easily.

ybiquitous:
I found the failure cause of e547bf420f04ab8f7b076c562343465a8bb98ec8. `extends` doesn't use dynamic `import()`. So, we cannot fix `getModulePath()` now.

However, I think we must fix dynamic `import()` also plugins, not only custom syntaxes:
https://github.com/stylelint/stylelint/blob/2a82f4d2c3147a823241327fe6d6e0a069894d21/lib/augmentConfig.mjs#L320

So, how about adding a tiny utility to wrap dynamic `import()`? For example,

```js
// lib/utils/dynamicImport.mjs
export default dynamicImport(path) {
  const fixedPath = isAbsolute(path) ? pathToFileURL(path).toString() : path;

  return import(fixedPath);
}
```

JounQin:
Sure, a tiny util is fine enough. I didn't find that usage previously.

ybiquitous (approved):
Thanks a lot! LGTM 👍🏼

CrOrc:
The same error is thrown in cli.mjs when running
```stylelint "**/*.css" --custom-formatter="./path/to/formatter.js"

JounQin:
@CrOrc You need to create a new issue with online and runnable reproduction.

PR #7385, opened 2023-12-10
Add system test for importing external packages

ybiquitous:

Ref #7382

This PR aims to test real use cases importing packages: `extends`, `customSyntax`, `plugins`.


ybiquitous:
Um, `ERR_UNSUPPORTED_ESM_URL_SCHEME` cannot be reproduced with the new system test... 🤔

PR #7406, opened 2023-12-16
Fix `declaration-property-value-no-unknown` and other false positives for multiline SCSS interpolation

### LINKED ISSUE #7403: Fix `declaration-property-value-no-unknown` parse error for multiline interpolation

```scss
.foo {
  color: #{
    some-scss-function()
  };
}
```

```json
{
  "rules": {
  }
}
```

```shell
stylelint file.scss
```

```json
{
  "stylelint": "^15.11.0",
  "stylelint-config-standard-scss": "^11.1.0"
}
```

No error, since interpolations should be ignored in the rule:

https://github.com/stylelint/stylelint/blob/da7ce21272d34928f40bbc68c1e01be1c076bb8a/lib/rules/declaration-property-value-no-unknown/index.mjs#L122

### What actually happened?

A problem was reported:

```
Cannot parse property value "#{
    some-scss-function()
  }" for property "color" (declaration-property-value-no-unknown)
```

I think the regex here needs to be changed to also take newline characters into account:

https://github.com/stylelint/stylelint/blob/da7ce21272d34928f40bbc68c1e01be1c076bb8a/lib/utils/hasScssInterpolation.mjs#L1

FloEdelmann:

Closes #7403.

No, it's self-explanatory.

### DIFF
--- a/.changeset/new-bananas-sell.md
+++ b/.changeset/new-bananas-sell.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: Ignore multiline SCSS interpolation e.g. in `declaration-property-value-no-unknown`
+Fixed: `declaration-property-value-no-unknown` and other false positives for multiline SCSS interpolation.
--- a/.changeset/new-bananas-sell.md
+++ b/.changeset/new-bananas-sell.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: Ignore multiline SCSS interpolation e.g. in `declaration-property-value-no-unknown`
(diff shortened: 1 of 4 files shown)


.changeset/new-bananas-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: Ignore multiline SCSS interpolation e.g. in `declaration-property-value-no-unknown`
```
jeddy3, line 5:
```suggestion
Fixed: `declaration-property-value-no-unknown` and other false positives for multiline SCSS interpolation.
```

To follow our conventions.

jeddy3 (changes requested):
Thank you.

I've requested a tweak to the changelog message, otherwise LGTM.

ybiquitous (approved):
Thanks. LGTM 👍🏼

jeddy3 (approved):

PR #7407, opened 2023-12-17
Document testing options in more detail in the v16 migration guide

### LINKED ISSUE #7399: Document testing options in more detail in the v16 migration guide

I was following https://stylelint.io/migration-guide/to-16#plugins and I had some issues with getting the tests working.

The first suggested migration path doesn't actually work when following the guide step by step :

```css
 {
   "scripts": {
-    "test": "jest",
+    "test": "cross-env NODE_OPTIONS=\"--experimental-vm-modules --no-warnings\" jest --runInBand",
   }
 }
```

I think it could work but I don't have `cross-env` installed so that gave an error.
I didn't want to install `cross-env` because the README says that it is in maintenance mode and I didn't want to investigate further as that felt like a sidetrack.

This also made me look closer at the actual test command and I didn't really like depending on experimental flags.

----------

The second path suggests using `jest-light-runner` if a specific error is encountered

> If you get a segmentation fault while running the preset on Node.js 18, you can use [jest-light-runner](https://www.npmjs.com/package/jest-light-runner):

But that wasn't my error, so I ignored this.

----------

The third option is actual gold :

> Alternatively, you can try our new [stylelint-test-rule-node](https://www.npmjs.com/package/stylelint-test-rule-node) package that uses the test runner and assertions built into Node.js:

Adding this and dropping `jest` removed almost 300 dev dependencies.
It also just worked.

Maybe a paragraph needs to be added about `cross-env`?
So that users know that it needs to be added?

------------------

Maybe `stylelint-test-rule-node` can be suggested as the first option to try?

To me this feels like the better path forward and if the first options hadn't failed I wouldn't have tried it.

jeddy3:

Closes #7399

None.

### DIFF
--- a/docs/migration-guide/to-16.md
+++ b/docs/migration-guide/to-16.md
@@ -143,7 +143,14 @@ For example, to migrate your plugin to use `import` and `export`:
 +export default createPlugin(ruleName, ruleFunction);
 ```
 
-We've published a new version of [Jest preset](https://www.npmjs.com/package/jest-preset-stylelint) for plugin tests. If you use it, you should update your npm-run-script:
+If you test your plugin using our `testRule` schema, you can either:
+
+- update to the last version of our [`jest-preset-stylelint`](https://www.npmjs.com/package/jest-preset-stylelint) package
+- try our new [`stylelint-test-rule-node`](https://www.npmjs.com/package/stylelint-test-rule-node) package that uses `node:test|assert`
+
+##### `jest-preset-stylelint`
+
+The preset needs the `--experimental-vm-modules` Node.js flag to support ESM. You can use the [`cross-env`](https://www.npmjs.com/package/cross-env) package to set the `NODE_OPTIONS` environment variable in your npm-run-script:
 
 ```diff json
  {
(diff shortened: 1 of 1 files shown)


docs/migration-guide/to-16.md
```diff
@@ -154,7 +161,9 @@ We've published a new version of [Jest preset](https://www.npmjs.com/package/jes
  }
 ```
 
-If you get a segmentation fault while running the preset on Node.js 18, you can use [`jest-light-runner`](https://www.npmjs.com/package/jest-light-runner):
+(`cross-env` is in maintainence-mode as it's [considered finished](https://github.com/kentcdodds/cross-env/issues/257#issue-754591323).)
```
romainmenke, line 164:
```suggestion
(`cross-env` is in maintenance-mode as it's [considered finished](https://github.com/kentcdodds/cross-env/issues/257#issue-754591323).)
```

romainmenke (commented):
Thank you for these changes, they fully address my issue 🙇

docs/migration-guide/to-16.md
```diff
@@ -154,7 +161,9 @@ We've published a new version of [Jest preset](https://www.npmjs.com/package/jes
  }
 ```
 
-If you get a segmentation fault while running the preset on Node.js 18, you can use [`jest-light-runner`](https://www.npmjs.com/package/jest-light-runner):
+(`cross-env` is in maintenance-mode as it's [considered finished](https://github.com/kentcdodds/cross-env/issues/257#issue-754591323).)
```
Mouvedia, line 164: minor

add `> [!NOTE]`

docs/migration-guide/to-16.md
```diff
@@ -183,6 +198,10 @@ Alternatively, you can try our new [`stylelint-test-rule-node`](https://www.npmj
  }
 ```
 
+The package uses `node:test` which is experimental within Node.js 18, but stable in 20. Coverage support is also experimental.
```
Mouvedia, line 201: minor

add `> [!IMPORTANT]`

docs/migration-guide/to-16.md
jeddy3: Let's only use those outside the migration guide, e.g. [Node.js API docs](https://stylelint.io/user-guide/node-api#output), as the migration guide has a flatter hierarchy.

ybiquitous (approved):
Great. LGTM 👍🏼

PR #7416, opened 2023-12-18
Fix `function-url-quotes` false positives for SCSS variable and `@` character 

### LINKED ISSUE #7404: Fix `function-url-quotes` false positives for SCSS variable and `@` character

```scss

.class {
  background-image: url($asset-prefix + 'assets/images/fa_calculator_blue@2x.png');
}

.otherclass {
    background: url($asset-prefix + 'assets/images/HSA_Icons-31.png');
}
```

```json
{
}
```

```shell
"stylelint \"apps/**/*.scss\" \"libs/**/*.scss\"" --fix
```

```json
"stylelint": "^16.0.2",
"stylelint-config-standard-scss": "^12.0.0",
 ```

nothing should happen. format is valid/conforming.

```scss
.class {
  background-image: url($asset-prefix + 'assets/images/fa_calculator_blue@2x.png');
}

.otherclass {
    background: url($asset-prefix + 'assets/images/HSA_Icons-31.png');
}
```

### What actually happened?

added extraneous double quotes around entire thing 

```scss
.class {
  background-image: url("$asset-prefix + 'assets/images/fa_calculator_blue@2x.png'");
}

.otherclass {
    background: url($asset-prefix + 'assets/images/HSA_Icons-31.png');
}
```

the `@` symbol in the image path seems to be the problem, when i remove that it does not add wrapping double quotes...

mattxwang:

Closes #7404 (though, see discussion below).

tl;dr: this PR (right now) contains a bandaid fix, but there may be a bigger problem of `isStandardSyntaxUrl`'s heuristic regex not doing what it should be. Would love to get some opinions since I don't have all the context for the util!

## long explanation / context for problem

First, to ~~briefly~~ explain why #7404 is happening: `function-url-quotes` uses `isStandardSyntaxUrl` to bail out for nonstandard urls within `url()`:

https://github.com/stylelint/stylelint/blob/b34a1846e50203b4a378e2fac195aa8e62f989ab/lib/utils/isStandardSyntaxUrl.mjs#L41-L46

The regex `IS_SCSS_VARIABLE_IN_URL` is:

https://github.com/stylelint/stylelint/blob/b34a1846e50203b4a378e2fac195aa8e62f989ab/lib/utils/isStandardSyntaxUrl.mjs#L7

Based off of the name and the comment in the condition, it seems like the purpose of this regex is to capture an SCSS variable within `url`. Makes sense! In fact, we have tests for SCSS variables + interpolation (which left me *very* confused as to why adding an `@` sign caused an error):

https://github.com/stylelint/stylelint/blob/b34a1846e50203b4a378e2fac195aa8e62f989ab/lib/utils/__tests__/isStandardSyntaxUrl.test.mjs#L137-L139

However, I want to point out that the regex is *anchored*. It only matches if the entire string matches that regex. So, in `$sass-variable + 'foo'`, `IS_SCSS_VARIABLE_IN_URL` is matching **all of `$sass-variable + 'foo'`**, instead of just `$sass-variable`. So, we're not "really" dealing with concatenation; we're just pretending that there is one SCSS variable called `$sass-variable + 'foo'`. Here's a [regex101 explaining this](https://regex101.com/r/uuNFM7/1).

This then explains why #7404 occurred: because `@` is not part of the character set, `IS_SCSS_VARIABLE_IN_URL` doesn't match the *entire* string `$asset-prefix + 'assets/images/fa_calculator_blue@2x.png'`, even though it probably "should" match `$asset-prefix`. Here is an [accompanying regex101](https://regex101.com/r/nL2A9W/1).

I've introduced some failing tests in https://github.com/stylelint/stylelint/commit/48485ef2130067df92ffbc96804c5344aa79dc8b to showcase this.

## solutions?

The minimum solution to fix this problem is to just add `@` to the regex. I did this in https://github.com/stylelint/stylelint/commit/90422c27bd0320a9f0f9f99461812be0f1c68cab, which resolves the failing tests in https://github.com/stylelint/stylelint/commit/48485ef2130067df92ffbc96804c5344aa79dc8b. In terms of the above explanation, this now just lets us treat `$asset-prefix + 'assets/images/fa_calculator_blue@2x.png'` as "one-giant SCSS variable".

However, I'm not super satisfied with this, since (in my eyes) it punts the problem down the line. For example, the similar declaration still has an error (see [accompanying regex101](https://regex101.com/r/QQN3U5/1)):

```scss
'https://' + $asset-prefix + 'assets/images/fa_calculator_blue@2x.png'
```

Unanchoring itself doesn't work, since it will then incorrectly flag strings with `$` within them, like this current test case:

```
http://domain.com:8080/path/to$to/file$2x.ext?$1=$2
```

So, a "robust"/full solution probably needs to actually parse the value. I'd be happy to spend some time exploring this if we think it's worthwhile, but we could also stick with the current heuristic and add `@` to resolve the immediate issue. I'm also cognizant that I may be missing context (e.g. I didn't write the original util). What are our thoughts?


lib/utils/__tests__/isStandardSyntaxUrl.test.mjs
```diff
@@ -81,13 +81,13 @@ describe('isStandardSyntaxUrl', () => {
...
-	it('url with ampersand at the start', () => {
+	it('url with at sign at the start', () => {
 		expect(isStandardSyntaxUrl('@url/path')).toBeTruthy();
 	});
-	it('url with ampersand in the middle', () => {
+	it('url with at sign in the middle', () => {
 		expect(isStandardSyntaxUrl('some/@url/path')).toBeTruthy();
 	});
-	it('url with ampersand at the end', () => {
+	it('url with at sign at the end', () => {
```
mattxwang, line 90: Separately fixing typos I noticed!

lib/utils/__tests__/isStandardSyntaxUrl.test.mjs
```diff
@@ -137,18 +137,18 @@ describe('isStandardSyntaxUrl', () => {
 	it('sass variable concat with single quotes string', () => {
 		expect(isStandardSyntaxUrl("$sass-variable + 'foo'")).toBeFalsy();
 	});
-	it('sass variable concat with escaped single quotes string', () => {
-		expect(isStandardSyntaxUrl("$sass-variable + 'foo'")).toBeFalsy();
-	});
 	it('sass variable concat with double quotes string', () => {
-		expect(isStandardSyntaxUrl("$sass-variable + 'foo'")).toBeFalsy();
+		expect(isStandardSyntaxUrl('$sass-variable + "foo"')).toBeFalsy();
```
mattxwang, line 141: This test case didn't do what it said earlier (?)

ybiquitous:
@mattxwang Thanks for the pull request!

First, don't hesitate to make a PR ready for review if you finish implementation. If a PR is still draft, we may ignore it ("Oh, it's not ready"). 😃 

Second, I don't think it's necessary to parse SCSS code in `isStandardSyntaxUrl()`. It's just a heuristic and not expected to be more accurate than needed. If we implemented a parser for SCSS, the code would be much more complex, and the package size would increase. I believe it's a plugin's task like `stylelint-scss`.

To resolve the example below that you mentioned, adding `:` to the regex seems enough. Such a case is rare, so I don't believe it's worth the increased code.

```scss
'https://' + $asset-prefix + 'assets/images/fa_calculator_blue@2x.png'
```

In conclusion, I agree with your current solution adding `@` to the regex. 👍🏼

mattxwang:
> First, don't hesitate to make a PR ready for review if you finish implementation. If a PR is still draft, we may ignore it ("Oh, it's not ready"). 😃

Ah good point, my mistake - I'll make sure not to do that in the future!

> Second, I don't think it's necessary to parse SCSS code in `isStandardSyntaxUrl()`. It's just a heuristic and not expected to be more accurate than needed. If we implemented a parser for SCSS, the code would be much more complex, and the package size would increase. I believe it's a plugin's task like `stylelint-scss`.
> 
> To resolve the example below that you mentioned, adding `:` to the regex seems enough. Such a case is rare, so I don't believe it's worth the increased code.
> 
> ```css-scss
> 'https://' + $asset-prefix + 'assets/images/fa_calculator_blue@2x.png'
> ```

Gotcha, I agree with the overall rationale. Unfortunately, this actually (by itself) _doesn't_ resolve the problem, since this will still then (incorrectly) trip this condition first:

https://github.com/stylelint/stylelint/blob/b34a1846e50203b4a378e2fac195aa8e62f989ab/lib/utils/isStandardSyntaxUrl.mjs#L26-L32

So, we'd need to shuffle around the order of checks - though that introduces more of a headache 😓 

> In conclusion, I agree with your current solution adding `@` to the regex. 👍🏼

Maybe I'll leave this PR as-is for now, and address the `:`-related use-case if a user submits a bug?

(with that in mind, have marked this PR as ready to review)

ybiquitous:
> Maybe I'll leave this PR as-is for now, and address the `:`-related use-case if a user submits a bug?

Ya, sounds good for now 👍🏼

ybiquitous (commented):
Thank you. LGTM 👍🏼

ybiquitous (approved):

PR #7417, opened 2023-12-19
Add `ignore: ["keyframe-selectors"]` to `selector-disallowed-list`

### LINKED ISSUE #7162: Add `ignore: ["keyframe-selectors"]` to `selector-disallowed-list`

```css
@keyframes fade-in {
  from {
    opacity: 0;
  }

  to {
    opacity: 1;
  }
}
```

```json
{
  "rules": {
    "selector-disallowed-list": [
      ["/^[a-z]+$/"]
    ]
  }
}
```

[Demo here](https://stylelint.io/demo/#N4Igxg9gJgpiBcIACBrGBPAZgJwIYFsYBnAAk11gFoBLAOxOAB16zsJ8HmTuSIAHXGGoAXdPBIAGANxcSAX2azhETi279BIsSQCMMlgtpyQAGhCZqAGxgA5AnEQwAHgT7WAdGCJFT4CLQsAcwQQJhZGEGwAV2siCPEwnhIIohhrMGVsSihqIlxLSwgAdxgoSktc4XiSAG1ZJLqQAHoAPRrcSgAvAF0AagASJoju+pIRg2ZjM0gA6kCAMQhsfFxhEIArIn9fWD4fRETkkCJRawraKoQjy1XiS5NZFNO0umFKGaDKE9xaKFxsKDVCI3YR3CKTEByIA)

```json
    "stylelint": "15.10.3",
    "stylelint-config-idiomatic-order": "9.0.0",
    "stylelint-config-prettier-scss": "1.0.0",
    "stylelint-config-recommended": "13.0.0",
    "stylelint-config-standard-scss": "10.0.0",
    "stylelint-prettier": "4.0.2",
```

Keyframes values not to be included in this rule as they are not "selectors".

### What actually happened?

See demo link above:

```
error Unexpected selector "from"
    [(selector-disallowed-list)](https://stylelint.io/user-guide/rules/selector-disallowed-list)
    [2:3-7]

error Unexpected selector "to"
    [(selector-disallowed-list)](https://stylelint.io/user-guide/rules/selector-disallowed-list)
    [6:3-5]
```

Either ignore `@keyframes` idents, or expand the rule's `ignore` option to include a `keyframes` option.

mattxwang:

Closes #7162.

Two quick questions:

1. ~~I used the `isKeyframeSelector` util (used elsewhere) to heuristically guess if the item is a keyframe selector, rather than seeing if it's actually *in* a keyframe. Does this make sense?~~ Am now using `isKeyframeRule` instead!
2. ~~I assumed that there should be nothing "special" about this option's interaction with `splitList`; in particular, if the list is not split, given `/from/` the current implementation will still flag `from, to`; it's only if the rule is given `/from/, { splitList: true }` that `from, to` is properly ignored. However, it'd be simple for me to reverse this behaviour - what are our thoughts?~~ Resolved by the former :)

### DIFF
--- a/lib/rules/selector-disallowed-list/README.md
+++ b/lib/rules/selector-disallowed-list/README.md
@@ -115,3 +115,29 @@ The following pattern is _not_ considered a problem:
   .foo {}
 }
 ```
+
+### `ignore: ["keyframe-selectors"]`
+
+Ignore selectors that are keyframe selectors.
+
+Given:
+
+```json
+[/from/, { ignore: ['keyframe-selectors'] }]
+```
+
+The following pattern is _not_ considered a problem:
+
+<!-- prettier-ignore -->
+```css
+@keyframes fade-in {
+  from {}
+}
+```
+
+The following pattern is considered a problem:
+
+<!-- prettier-ignore -->
+```css
+.from {}
+```
--- a/lib/rules/selector-disallowed-list/README.md
+++ b/lib/rules/selector-disallowed-list/README.md
@@ -123,7 +123,7 @@ Ignore keyframe selectors.
 Given:
 
 ```json
-["/from/", { "ignore": ["keyframe-selectors}"] }]
+["/from/", { "ignore": ["keyframe-selectors"] }]
 ```
 
 The following pattern is _not_ considered a problem:
--- a/lib/rules/selector-disallowed-list/README.md
+++ b/lib/rules/selector-disallowed-list/README.md
@@ -80,7 +80,7 @@ For example, with `true`.
 Given:
 
 ```json
-[".foo"]
+[".foo", { "splitList": true }]
 ```
 
 The following pattern is considered a problem:
@@ -104,7 +104,7 @@ Ignore selectors that are inside a block.
 Given:
 
 ```json
-[".foo"]
+[".foo", { "ignore": ["inside-block"] }]
 ```
 
 The following pattern is _not_ considered a problem:
--- a/lib/rules/selector-disallowed-list/README.md
+++ b/lib/rules/selector-disallowed-list/README.md
@@ -123,7 +123,7 @@ Ignore selectors that are keyframe selectors.
 Given:
 
 ```json
-[/from/, { ignore: ['keyframe-selectors'] }]
+["/from/", { "ignore": ["keyframe-selectors}"] }]
 ```
 
 The following pattern is _not_ considered a problem:
@@ -133,11 +133,4 @@ The following pattern is _not_ considered a problem:
 @keyframes fade-in {
   from {}
 }
-```
-
-The following pattern is considered a problem:
-
-<!-- prettier-ignore -->
-```css
-.from {}
 ```
--- a/lib/rules/selector-disallowed-list/README.md
+++ b/lib/rules/selector-disallowed-list/README.md
@@ -118,7 +118,7 @@ The following pattern is _not_ considered a problem:
 
 ### `ignore: ["keyframe-selectors"]`
 
-Ignore selectors that are keyframe selectors.
+Ignore keyframe selectors.
 
 Given:
 
--- a/lib/rules/selector-disallowed-list/__tests__/index.mjs
+++ b/lib/rules/selector-disallowed-list/__tests__/index.mjs
@@ -225,7 +225,7 @@ testRule({
 
 testRule({
 	ruleName,
-	config: [[/from/, /[0-9]/], { ignore: ['keyframe-selectors'] }],
+	config: [[/[A-Za-z]/], { ignore: ['keyframe-selectors'] }],
 
 	accept: [
 		{
@@ -238,7 +238,7 @@ testRule({
 		{
 			code: stripIndent`
 				@keyframes fade-in {
-					0% {}
+					to {}
 				}
 			`,
 		},
@@ -261,12 +261,12 @@ testRule({
 			endColumn: 6,
 		},
 		{
-			code: '.color-0% {}',
-			message: messages.rejected('.color-0%'),
+			code: '.to {}',
+			message: messages.rejected('.to'),
 			line: 1,
 			column: 1,
 			endLine: 1,
-			endColumn: 10,
+			endColumn: 4,
 		},
 	],
 });
--- a/lib/rules/selector-disallowed-list/__tests__/index.mjs
+++ b/lib/rules/selector-disallowed-list/__tests__/index.mjs
@@ -262,11 +262,11 @@ testRule({
 			endColumn: 10,
 		},
 		{
-			code: stripIndent(`
+			code: stripIndent`
 				@keyframes fade-in {
 					from, to {}
 				}
-			`),
+			`,
 			message: messages.rejected('from, to'),
 			line: 2,
 			column: 2,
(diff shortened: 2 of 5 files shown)


lib/rules/selector-disallowed-list/README.md
```diff
@@ -115,3 +115,29 @@ The following pattern is _not_ considered a problem:
...
 ```
+
+### `ignore: ["keyframe-selectors"]`
+
+Ignore selectors that are keyframe selectors.
+
+Given:
+
+```json
+[/from/, { ignore: ['keyframe-selectors'] }]
```
Mouvedia, line 126: minor

- add missing `"`
- replace `'` by `"`

ybiquitous: [suggest] Yes, the current snippet is invalid JSON. Should be:

```suggestion
["/from/", { "ignore": ["keyframe-selectors}"] }]
```

Besides, can you fix other secondary option snippets in this README, too? To keep consistency, they should be:

```json
[".foo", { "splitList": true }]
```

```json
[".foo", { "ignore": ["inside-block"] }]
```

lib/rules/selector-disallowed-list/__tests__/index.mjs
```diff
@@ -220,3 +222,88 @@ testRule({
...
 	],
 });
+
+testRule({
+	ruleName,
+	config: [[/from/, /[0-9]/], { ignore: ['keyframe-selectors'] }],
+
+	accept: [
+		{
+			code: stripIndent(`
```
ybiquitous, line 232: [suggest] Minor. Parens are unnecessary for consistency with other codes.

```suggestion
			code: stripIndent`
```

lib/rules/selector-disallowed-list/__tests__/index.mjs
```diff
@@ -220,3 +222,88 @@ testRule({
...
+				@keyframes fade-in {
+					from {
+						opacity: 0;
+					}
```
ybiquitous, line 236: [suggest] Minor. We can save a few lines by doing this: 😄 

```suggestion
					from { opacity: 0; }
```

ybiquitous (changes requested):
@mattxwang Thanks for the pull request!

I think it's better to check if the rule node is inside `@keyframes {}` block rather than using `isKeyframeSelector()` like this:

```js
if (ignoreKeyframeSelectors && isInsideKeyframes(ruleNode)) {
	return;
}

// ...

const keyframesPattern = /^keyframes$/i;

/**
 * @param {import('postcss').Node} node
 * @returns {boolean}
 */
function isInsideKeyframes({ parent }) {
	return Boolean(parent && isAtRule(parent) && keyframesPattern.test(parent.name));
}
```

If doing so, `splitList` is unnecessary for `ignore: ['keyframe-selectors']`. Any thoughts?

Mouvedia:
> I think it's better to check if the rule node is inside `@keyframes {}` block rather than using `isKeyframeSelector()` like this:
> 
> ```js
> […]
> 
> const keyframesPattern = /^keyframes$/i;
> 
> /**
>  * @param {import('postcss').Node} node
>  * @returns {boolean}
>  */
> function isInsideKeyframes({ parent }) {
> 	return Boolean(parent && isAtRule(parent) && keyframesPattern.test(parent.name));
> }
> ```

This looks similar to `/lib/utils/isKeyframeRule.mjs`.

ybiquitous:
Ah, I forgot `isKeyframeRule`. Nice catch! 👍🏼

lib/rules/selector-disallowed-list/README.md
```diff
@@ -115,3 +115,29 @@ The following pattern is _not_ considered a problem:
...
+  from {}
+}
+```
+
+The following pattern is considered a problem:
+
+<!-- prettier-ignore -->
+```css
+.from {}
+```
```
jeddy3, line 143:
```suggestion
```

We typically only show what the ignore option allows to manage the size of the READMEs.

lib/rules/selector-disallowed-list/__tests__/index.mjs
```diff
@@ -237,6 +237,15 @@ testRule({
 				}
 			`),
 		},
+		{
+			code: stripIndent(`
+				@keyframes fade-in {
+					0% {
+						opacity: 0;
+					}
```
jeddy3, line 245:
```suggestion
					0% {}
```

[Conventions](https://stylelint.io/developer-guide/rules/#write-tests).

lib/rules/selector-disallowed-list/__tests__/index.mjs
```diff
@@ -248,6 +257,28 @@ testRule({
 			endLine: 1,
 			endColumn: 6,
 		},
+		{
+			code: '.color-0% {}',
```
jeddy3, line 261: Is this valid CSS? If not, let's remove it or try to find a valid alternative.

lib/rules/selector-disallowed-list/__tests__/index.mjs
```diff
@@ -248,6 +257,28 @@ testRule({
...
+			column: 1,
+			endLine: 1,
+			endColumn: 10,
+		},
+		{
+			code: stripIndent(`
+				@keyframes fade-in {
+					from, to {
+						opacity: 0;
+					}
```
jeddy3, line 273:
```suggestion
					from, to {}
```

lib/rules/selector-disallowed-list/__tests__/index.mjs
```diff
@@ -256,6 +287,15 @@ testRule({
 	config: [/from/, { ignore: ['keyframe-selectors'], splitList: true }],
 
 	accept: [
+		{
+			code: stripIndent(`
+				@keyframes fade-in {
+					from {
+						opacity: 0;
+					}
```
jeddy3, line 295:
```suggestion
					from {}
```

jeddy3 (changes requested):
Looking good. I've requested some changes to align the docs and tests with our conventions.

lib/rules/selector-disallowed-list/__tests__/index.mjs
mattxwang: Great point, thanks - for some reason I thought `%` was allowed with an ident, my mistake!

mattxwang:
Thanks for the feedback everyone! Have just made the requested changes - let me know what y'all think!

lib/rules/selector-disallowed-list/README.md
```diff
@@ -115,3 +115,22 @@ The following pattern is _not_ considered a problem:
...
 ```
+
+### `ignore: ["keyframe-selectors"]`
+
+Ignore keyframe selectors.
+
+Given:
+
+```json
+["/from/", { "ignore": ["keyframe-selectors}"] }]
```
ybiquitous, line 126: Oops. 😅 

```suggestion
["/from/", { "ignore": ["keyframe-selectors"] }]
```

ybiquitous (approved):
Thank you! LGTM 👍🏼

lib/rules/selector-disallowed-list/README.md
mattxwang: Great catch, thanks!

(I should have caught it!)

lib/rules/selector-disallowed-list/__tests__/index.mjs
```diff
@@ -220,3 +222,73 @@ testRule({
...
+
+testRule({
+	ruleName,
+	config: [[/[A-Za-z]/], { ignore: ['keyframe-selectors'] }],
+
+	accept: [
+		{
+			code: stripIndent`
+				@keyframes fade-in {
+					from { opacity: 0; }
```
jeddy3, line 234:
```suggestion
					from {}
```

jeddy3 (changes requested):

jeddy3 (approved):
LGTM, thank you!

PR #7418, opened 2023-12-19
Add `--rule` flag, `--rules` flag, `--merge-rules` flag and `mergeRules` option

Mouvedia:

#4331

- [x] standalone
- [x] cli
- [x] documentation
- [x] tests  

1. levn has limited uses; I could refactor and use `JSON.parse` instead
e.g. having a schema for an object requires explicit property names
cf https://github.com/stylelint/stylelint/pull/7418/commits/d321cdd8c1791665970dad569b4f5e0587ca1d5d#diff-d1470268bb8c3a42692b2c5249d7d31832ceeed1b1f16fbdbc814f8013c36b31R687
2. levn is too permissive when using `Object` as the type
3. levn custom types are too restrictive
4. levn is already a sub-dependency of ESLint hence it's costless size-wise

###  missing features

- value-less rule selection

Passing a rule without its value seemed confusing.
What should `--rule color-no-invalid-hex` do exactly?
Probably fetch the corresponding rule option in the config and, if found, discard all other rules.
I chose not to add this feature, for now; it needs to be discussed in a separate issue.

- `--rules` should accept a file path (JSON)


lib/getConfigForFile.mjs
Mouvedia, line 1: This file has been refactored for readability.
Only 2 `if` blocks at l87 and l93 are to be reviewed.

Mouvedia:
https://github.com/stylelint/stylelint/issues/4331#issuecomment-1880155770

PR #7434, opened 2023-12-31
Refactor to use `@csstools/css-parser-algorithms` `walk` helper

romainmenke:

None, this is a refactor after an upstream API addition

Starting an AST walk when you have a list of component values was a bit cumbersome because the list is plain JavaScript array. So you have to iterate the top levels manually and do all your checks there first and then start walking down simple blocks and functions.

I've added a `walk` function that takes a list of component values as its argument.
This makes it possible to remove some duplicate code here.

-----

_This isn't a user facing change so I don't think a changelog entry is needed._

### DIFF
--- a/lib/rules/function-no-unknown/index.cjs
+++ b/lib/rules/function-no-unknown/index.cjs
@@ -60,13 +60,10 @@ const rule = (primary, secondaryOptions) => {
 
 			if (!isStandardSyntaxValue(value)) return;
 
-			/**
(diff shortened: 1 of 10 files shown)


jeddy3 (approved):
Nice addition to the API and a nice refactor! LGTM, thank you.

> This isn't a user facing change so I don't think a changelog entry is needed.

Yep, no changelog entry needed.

PR #7438, opened 2024-01-01
Fix `at-rule-no-unknown` false positives for `@starting-style`

### LINKED ISSUE #7420: Fix `at-rule-no-unknown` false positive for `@starting-style`

```css
@starting-style {}
```

```json
{
  "rules": {
    "at-rule-no-unknown": true
  }
}
```

https://stylelint.io/demo

```json
{
  "stylelint": "latest",
}
```

`@starting-style` to be _known_ since it has a decent browser support.
see https://caniuse.com/mdn-css_at-rules_starting-style

### What actually happened?

Unexpected unknown at-rule "@starting-style"

Add `starting-style` to
https://github.com/stylelint/stylelint/blob/b92260f415ae2e1c1ee053ea01658be0b063899b/lib/reference/atKeywords.mjs#L4
and add `nestingSupportedAtKeywords` to
https://github.com/stylelint/stylelint/blob/b92260f415ae2e1c1ee053ea01658be0b063899b/lib/reference/atKeywords.mjs#L33

fpetrakov:

Closes #7420

No, it's self-explanatory.

### DIFF
--- a/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
+++ b/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
@@ -12,9 +12,6 @@ testRule({
 			code: 'a { top: 0; }',
 		},
 		{
-			code: 'a { @starting-style { opacity: 0; } }',
-		},
-		{
 			code: 'a { @media screen { top: 0; } }',
 		},
 		{
--- a/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
+++ b/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
@@ -12,6 +12,9 @@ testRule({
 			code: 'a { top: 0; }',
 		},
 		{
+			code: 'a { @starting-style { opacity: 0; } }',
+		},
+		{
 			code: 'a { @media screen { top: 0; } }',
 		},
 		{
--- a/.changeset/nervous-seas-retire.md
+++ b/.changeset/nervous-seas-retire.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `at-rule-no-unknown` false positives for `@starting-style`
--- a/lib/reference/atKeywords.cjs
+++ b/lib/reference/atKeywords.cjs
@@ -10,6 +10,7 @@ const nestingSupportedAtKeywords = new Set([
 	'layer',
 	'media',
 	'scope',
+	'starting-style',
 	'supports',
 ]);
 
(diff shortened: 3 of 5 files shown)


lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
```diff
@@ -11,6 +11,9 @@ testRule({
 		{
 			code: 'a { top: 0; }',
 		},
+		{
+			code: 'a { @starting-style { opacity: 0; } }',
+		},
```
ybiquitous, line 16: [suggest] This change also should fix the following *false negative* problem for `declaration-property-value-no-unknown`:

```css
a { @starting-style { padding: auto; } } /* no problem */

a { padding: auto; } /* problem for "auto" */
```

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYKACBnALkgTjgJYB2A5gLS4CeANjGigA5JRSlnwpICuOEAbhQBfEQB0SE1OhZsOXXvyHCQAGhAAzIvQBySALZxEMAB4Gm9AHRgsWNeAgktZBCGASUKMSHw96Wby53Ek9Pb1gwWgIkYkcKJnwIJhhCagoANyRaHhgKEggKHhIAa3yAd0kEFBxfGA9xEhV1SCciMgAxCHx9GNcAKyxHe1gmO0RgsJAaelpSHECvEABGADZLAAZLACZvCRVhIA)

So, how about opening another Pull Request for `declaration-property-value-no-unknown`?

ybiquitous (commented):
@fpetrakov Thanks for the pull request! The change for `at-rule-no-unknown` looks good to me. 👍🏼

lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
fpetrakov: @ybiquitous thanks for clarifying it. I removed code related to `declaration-property-value-no-unknown`. Is there an opened issue for it?

Mouvedia: > So, how about opening another Pull Request for `declaration-property-value-no-unknown`?

I would prefer adding 2 changesets instead.

ybiquitous (approved):
Thanks. LGTM 👍🏼

lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
ybiquitous: > Is there an opened issue for it?

@fpetrakov No. It's OK to directly open a PR.

jeddy3 (approved):

lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
ybiquitous: #7461 was opened.

Issue #7442, opened 2024-01-04
Fix `function-no-unknown|value-keyword-case` false positives for template literals with line breaks


Sh031224:
### What minimal example or steps are needed to reproduce the bug?

```ts
const mergeWithVar = (token: Token) => `var(--${token.replace(/\./g, '-')})`

const First = styled.div`
  border: 1px solid ${mergeWithVar('color.primary.normal')}; // works fine
`

const Second = styled.div`
  border: 1px solid 
    ${mergeWithVar('color.primary.normal')}; // works fine
`

const isPrimary = false;

const Third = styled.div`
  border: 1px solid
    ${mergeWithVar( // Unexpected unknown function "mergeWithVar" (function-no-unknown)
      isPrimary ? 'color.primary.normal' : 'color.secondary.normal' // Expected "isPrimary" to be "isprimary" (value-keyword-case)
    )};
`

const Fourth = styled.div`
  border: 1px solid ${mergeWithVar( // Unexpected unknown function "mergeWithVar" (function-no-unknown)
    isPrimary ? 'color.primary.normal' : 'color.secondary.normal' // Expected "isPrimary" to be "isprimary" (value-keyword-case)
  )};
`
```

### What minimal configuration is needed to reproduce the bug?

```js
// stylelint.config.js
module.exports = {
  customSyntax: "postcss-styled-syntax",
  extends: [
    "stylelint-config-standard",
  ],
  rules: {}
}
```

### How did you run Stylelint?

```sh
stylelint **/*.ts
```

### Which Stylelint-related dependencies are you using?

```json
{
  "postcss-styled-syntax": "^0.6.3",
  "stylelint": "^16.1.0",
  "stylelint-config-standard": "^36.0.0",
}
```

### What did you expect to happen?

I want the task to complete normally without errors.

### What actually happened?

An incorrect error occurs only for elements that have line breaks during templete literals.

- `Unexpected unknown function "mergeWithVar" (function-no-unknown)`
- `Expected "isPrimary" to be "isprimary" (value-keyword-case)`

PR #7443, opened 2024-01-04
Fix `function-no-unknown|value-keyword-case` false positives for template literals with line breaks

Sh031224:

#7442 

The regular expression that checks template literals in js has been changed.

### DIFF
--- a/.changeset/beige-peaches-add.md
+++ b/.changeset/beige-peaches-add.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fix: `Template literal` behavior with line breaks is incorrect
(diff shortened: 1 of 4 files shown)


.changeset/beige-peaches-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fix: `Template literal` behavior with line breaks is incorrect
```
jeddy3, line 5:
```suggestion
Fixed: `function-no-unknown|value-keyword-case` false positives for template literals with line breaks
```

jeddy3 (approved):
LGTM, thank you!

ybiquitous (approved):
Thanks, LGTM!

PR #7455, opened 2024-01-08
Document `--output-file` flag in migration guide for CLI stderr change

### LINKED ISSUE #7441: Document `--output-file` flag in migration guide for CLI stderr change

I am in the process of migrating from stylelint v15 to stylelint v16. I have a formatter which generates a report (both success & error cases) of the stylelint result and save it in a file. In v15 I used the following command to achieve this:

```
stylelint "**/*.css" --custom-formatter my-formatter.js > report.json
```

Since v16 the above script doesn't work anymore, the `report.json` file is not created. After debugging a bit I found out that now the result of the formatter is now written to stderr, so I need to use `&>` instead of `>`. For formatters which exclusively print errors, this makes sense, but this is not necessarily the case for all formatters.

This is inconvenient for me, because I cannot be sure that my `report.json` file contains valid JSON, as it could also contain an error message from stylelint. As a workaround, I'm checking the exit code of stylelint, and remove the report file if it's neither 0 nor 2.

1. Update the migration guide and mention that outputs of all custom formatters will now be printed to stderr as well
2. Offer a new `--custom-formatter-write-stdout` CLI flag to write the result coming from a custom formatter to stdout instead of stderr

jeddy3:

Closes #7441

No, it's self-explanatory.

### DIFF
--- a/docs/migration-guide/to-16.md
+++ b/docs/migration-guide/to-16.md
@@ -328,7 +328,7 @@ We've changed the CLI to print problems to stderr instead of stdout.
 
 If you use the [`--fix`](../user-guide/cli.md#--fix) and [`--stdin`](../user-guide/cli.md#--stdin) options, the CLI will print the fixed code to stdout and any problems to stderr.
 
-If you use `>` to redirect standard output, e.g. from a custom formatter, can use use the [`--output-file`](../user-guide/cli.md#--output-file--o) flag instead:
+If you use `>` to redirect standard output, e.g. from a custom formatter, you can use the [`--output-file`](../user-guide/cli.md#--output-file--o) flag instead:
 
 ```diff shell
 -npx stylelint "*.css" --custom-formatter custom-formatter.js > output.txt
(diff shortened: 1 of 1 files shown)


docs/migration-guide/to-16.md
```diff
@@ -328,6 +328,13 @@ We've changed the CLI to print problems to stderr instead of stdout.
 
 If you use the [`--fix`](../user-guide/cli.md#--fix) and [`--stdin`](../user-guide/cli.md#--stdin) options, the CLI will print the fixed code to stdout and any problems to stderr.
 
+If you use `>` to redirect standard output, e.g. from a custom formatter, can use use the [`--output-file`](../user-guide/cli.md#--output-file--o) flag instead:
```
ybiquitous, line 331: [suggest] A typo is found (double "use"):

```suggestion
If you use `>` to redirect standard output, e.g. from a custom formatter, you can use the [`--output-file`](../user-guide/cli.md#--output-file--o) flag instead:
```

ybiquitous (approved):
Looks good! A typo may be found, but you can go on without my additional approval. 👌🏼

ybiquitous (approved):

PR #7456, opened 2024-01-08
Fix `Error: Could not find <package>` message clarity

### LINKED ISSUE #7448: Fix `Error: Could not find <package>` message clarity

```
npm i stylelint
```

(Note that I didn't install the standard config.)

```
{
  "extends": ["stylelint-config-standard"]
}
```

```shell
npx stylelint "test.css"
```

```json
{
  "stylelint": "16.1.0"
}
```

To get an error message informing me that I probably forgot to install the shared config.

Ref: https://github.com/stylelint/stylelint/issues/7445

### What actually happened?

I got:

```console
Error: Could not find "stylelint-config-standard". Do you need the "configBasedir" or "--config-basedir" option?
```

Reword the error message for the more common use case of a missing package, whereas a local config needing `configBasedir` is uncommon.

```console
Error: Could not find "stylelint-config-standard". Do you need to install the package or use the "configBasedir" option?
```

jeddy3:

Closes #7448

No, it's self-explanatory.

### DIFF
--- a/lib/utils/getModulePath.cjs
+++ b/lib/utils/getModulePath.cjs
@@ -29,7 +29,7 @@ function getModulePath(basedir, lookup, cwd = process.cwd()) {
 
 	if (!path) {
 		throw configurationError(
-			`Could not find "${lookup}". Do you need the "configBasedir" or "--config-basedir" option?`,
+			`Could not find "${lookup}". Do you need to install the package or use the "configBasedir" option?`,
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


ybiquitous (approved):
LGTM, thank you.

Just to confirm, this change doesn't need a changelog entry, right?

jeddy3:
> Just to confirm, this change doesn't need a changelog entry, right?

Good catch, it should. Added.

ybiquitous (approved):
Thanks! 👍🏼

PR #7461, opened 2024-01-13
Fix `declaration-property-value-no-unknown` false negatives for `@starting-style`

fpetrakov:

None, it's a false negative fix for `declaration-property-value-no-unknown` rule.

This false negative is explained here - https://github.com/stylelint/stylelint/pull/7438#discussion_r1439118910

### DIFF
--- a/.changeset/cyan-mice-enjoy.md
+++ b/.changeset/cyan-mice-enjoy.md
@@ -1,5 +1,5 @@
 ---
-"stylelint": patch
+"stylelint": minor
 ---
 
 Fixed: `declaration-property-value-no-unknown` false negatives for `@starting-style`
--- a/.changeset/cyan-mice-enjoy.md
+++ b/.changeset/cyan-mice-enjoy.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `declaration-property-value-no-unknown` false negatives for `@starting-style`
--- a/lib/reference/atKeywords.mjs
+++ b/lib/reference/atKeywords.mjs
@@ -6,8 +6,8 @@ export const nestingSupportedAtKeywords = new Set([
 	'layer',
 	'media',
 	'scope',
-	'supports',
 	'starting-style',
+	'supports',
 ]);
 
 // https://www.w3.org/TR/css-page-3/#syntax-page-selector
@@ -31,12 +31,11 @@ const pageMarginAtKeywords = new Set([
 ]);
 
 // https://developer.mozilla.org/en/docs/Web/CSS/At-rule
-export const atKeywords = uniteSets(pageMarginAtKeywords, [
+export const atKeywords = uniteSets(nestingSupportedAtKeywords, pageMarginAtKeywords, [
 	'annotation',
 	'apply',
 	'character-variant',
 	'charset',
-	'container',
 	'counter-style',
 	'custom-media',
 	'custom-selector',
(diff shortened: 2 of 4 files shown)


lib/reference/atKeywords.mjs
```diff
@@ -7,6 +7,7 @@ export const nestingSupportedAtKeywords = new Set([
 	'media',
 	'scope',
 	'supports',
+	'starting-style',
```
Mouvedia, line 10: Id suggest to add `nestingSupportedAtKeywords` to l34 and remove the duplicates from it.
i.e. since `nestingSupportedAtKeywords` are always \@ keywords they are automatically part of `atKeywords`

ybiquitous: Make sense. Let's refactor `atKeywords.mjs` like this:

- Add `nestingSupportedAtKeywords` to `atKeywords`:
  ```diff
  -export const atKeywords = uniteSets(pageMarginAtKeywords, [
  +export const atKeywords = uniteSets(nestingSupportedAtKeywords, pageMarginAtKeywords, [
  ```
- Remove duplicate at-rules (e.g. `container`) from `atKeywords`:
  ```diff
  -	'container',
  ```

Ref: #7420 (original issue)

ybiquitous, line 10: [suggest] Please sort the at-rules alphabetically:

```suggestion
	'starting-style',
	'supports',
```

ybiquitous (changes requested):
@fpetrakov Thanks for the follow-up pull request! I've commented on the refactoring as @Mouvedia suggested. Can you address it?

lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
```diff
@@ -11,6 +11,9 @@ testRule({
 		{
 			code: 'a { top: 0; }',
 		},
+		{
+			code: 'a { @starting-style { opacity: 0; } }',
```
Mouvedia, line 15: minor nit

> - the color property by default
> - the red value by default

Since the rest of the examples don't respect it, you don't have to though.
cf https://stylelint.io/developer-guide/rules#write-tests

Mouvedia (approved):

ybiquitous (approved):
Thank you. LGTM 👍🏼

.changeset/cyan-mice-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
```
jeddy3, line 2:
```suggestion
"stylelint": minor
```

To align with our semver policy.

jeddy3 (changes requested):

.changeset/cyan-mice-enjoy.md
ybiquitous: Oh, it looks like I misunderstood the policy. False negatives are included in **minor** updates, right?

For example, https://github.com/stylelint/stylelint/releases/tag/15.10.3 includes a false negative fix released as a **patch**. Was this incorrect?

jeddy3: > False negatives are included in minor updates, right?

Yes, "a bug fix in a rule that results in Stylelint reporting more errors".

> For example, https://github.com/stylelint/stylelint/releases/tag/15.10.3 includes a false negative fix released as a patch. Was this incorrect?

It was incorrect, but it's okay.

ybiquitous: Ah, I see it. I'll care about the policy more. 👍🏼

jeddy3 (approved):

Issue #7401, opened 2023-12-12
Add `media-query-no-invalid` specific problem messages


coderaiser:
### What minimal example or steps are needed to reproduce the bug?

What is wrong with this code? How should I fix it?

```css
@media only screen and (device-width >= 375px) and (device-width <= 667px) and (orientation: landscape),
@media only screen and (max-height: 360px) and (max-width: 640px) {
    .fm {
        height: 55%;
    }

    .files {
        height: 60%;
    }
}
```

### What minimal configuration is needed to reproduce the bug?

defaults

### How did you run Stylelint?

as 🐊[**Putout** Processor CSS](https://github.com/coderaiser/putout/tree/v34.0.7/packages/processor-css)

### Which Stylelint-related dependencies are you using?

latest

### What did you expect to happen?

More detailed error, auto fix

### What actually happened?

 Unexpected invalid media query "only screen and (device-width >= 375px) and (device-width <= 667px) and (orientation: landscape)

romainmenke:
Hi @coderaiser,

You have written `@media` twice in a list of queries.

Your code :

```css
@media foo,
@media bar {
  /* some styles */
}
```

The correct form is to only write `@media` at the start :

```css
@media foo,
bar {
  /* some styles */
}
```

----

```diff
@media only screen and (device-width >= 375px) and (device-width <= 667px) and (orientation: landscape),
- @media only screen and (max-height: 360px) and (max-width: 640px) {
+ only screen and (max-height: 360px) and (max-width: 640px) {
    .fm {
        height: 55%;
    }

    .files {
        height: 60%;
    }
}
```

coderaiser:
Suggested solution works good, but would be great more clear message, or better autofix

jeddy3:
I've relabelled as a discussion around adding either autofix or specific problem message.

Does `@csstools/media-query-list-parser` identify specific problems we can surface in our problem message?

romainmenke:
I don't think we can be more specific for this specific case because it is completely invalid and just can't be parsed in any way. There will always be more ways that source code can be broken than a parser can specifically handle and detect.

This error message is accurate and the best we can do :

```
Unexpected invalid media query "
@media only screen and (max-height: 360px) and (max-width: 640px)"
[(media-query-no-invalid)](https://stylelint.io/user-guide/rules/media-query-no-invalid)
```

It is confusing because of the source, not because of the reporting.

[...]

PR #7462, opened 2024-01-13
Add `media-query-no-invalid` specific problem messages

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/7401

I've added a second argument for the `rejected` message.
- is there an existing pattern in other rules for what I want to do here?
- is `hint` a good name for the second argument? (maybe `reason`, ...)

I am also unsure what the best wording is.
I don't think most people are familiar with the wording of the specification, e.g. `range context`, `plain feature`, ...
So I've opted to label everything as a query : `range query`, `boolean query`.

### DIFF
--- a/.changeset/healthy-donkeys-appear.md
+++ b/.changeset/healthy-donkeys-appear.md
@@ -1,5 +1,5 @@
 ---
-"stylelint": patch
+"stylelint": minor
 ---
 
 Added: `media-query-no-invalid` specific problem messages
--- a/lib/rules/media-query-no-invalid/index.cjs
+++ b/lib/rules/media-query-no-invalid/index.cjs
@@ -15,7 +15,11 @@ const validateOptions = require('../../utils/validateOptions.cjs');
 const ruleName = 'media-query-no-invalid';
 
 const messages = ruleMessages(ruleName, {
-	rejected: (query) => `Unexpected invalid media query "${query}"`,
+	rejected: (query, hint) => {
+		if (!hint) return `Unexpected invalid media query "${query}"`;
+
+		return `Unexpected invalid media query "${query}", ${hint}`;
+	},
 });
 
 const HAS_MIN_MAX_PREFIX = /^(?:min|max)-/i;
@@ -34,13 +38,12 @@ const rule = (primary) => {
 		}
 
 		root.walkAtRules(/^media$/i, (atRule) => {
-			/** @type {Array<{tokens(): Array<import('@csstools/css-tokenizer').CSSToken>}>} */
-			let invalidNodes = [];
+			const atRuleParamIndexValue = atRuleParamIndex(atRule);
 
 			parseMediaQuery(atRule).forEach((mediaQuery) => {
 				if (mediaQueryListParser.isMediaQueryInvalid(mediaQuery)) {
 					// Queries that fail to parse are invalid.
-					invalidNodes.push(mediaQuery);
+					complain(atRule, atRuleParamIndexValue, mediaQuery);
 
 					return;
 				}
@@ -48,7 +51,7 @@ const rule = (primary) => {
 				mediaQuery.walk(({ node, parent }) => {
 					// All general enclosed nodes are invalid.
 					if (mediaQueryListParser.isGeneralEnclosed(node)) {
-						invalidNodes.push(node);
+						complain(atRule, atRuleParamIndexValue, node);
 
 						return;
 					}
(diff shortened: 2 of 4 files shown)


lib/rules/media-query-no-invalid/index.mjs
```diff
@@ -37,21 +41,20 @@ const rule = (primary) => {
...
 		root.walkAtRules(/^media$/i, (atRule) => {
-			/** @type {Array<{tokens(): Array<import('@csstools/css-tokenizer').CSSToken>}>} */
-			let invalidNodes = [];
+			const atRuleParamIndexValue = atRuleParamIndex(atRule);
 
 			parseMediaQuery(atRule).forEach((mediaQuery) => {
 				if (isMediaQueryInvalid(mediaQuery)) {
 					// Queries that fail to parse are invalid.
-					invalidNodes.push(mediaQuery);
+					complain(atRule, atRuleParamIndexValue, mediaQuery);
```
Mouvedia, line 49: question

Can we lint files in separate threads?
If not just resolve this comment.
I wonder if other reports from other rules could insert themselves in between the reports of this rule.
i.e. unlike before the reports are not grouped

lib/rules/media-query-no-invalid/index.mjs
```diff
@@ -62,7 +65,12 @@ const rule = (primary) => {
 
 						if (isCustomMediaQuery(name)) {
 							// In a plain context, custom media queries are invalid.
-							invalidNodes.push(parent);
+							complain(
```
Mouvedia, line 68: nit

Since you are reusing the same arguments 5 times you could maybe set a new function at l65.
e.g.
```
const foo = complain.bind(bar, atRule, atRuleParamIndexValue, parent);
[…]
foo('custom media queries can only be used in boolean queries');
```

ybiquitous:
@romainmenke Thanks for the pull request!

> * is there an existing pattern in other rules for what I want to do here?

As far as I know, there is no such case. But I think what this PR try is not a problem. 👌🏼 

> * is `hint` a good name for the second argument? (maybe `reason`, ...)

I prefer `reason` a bit because I think people want to know a reason of the lint problems.

> I don't think most people are familiar with the wording of the specification

I think so, too. Your try looks good to me. 👍🏼

lib/rules/media-query-no-invalid/index.mjs
ybiquitous: [suggest] Good point. How about reducing duplications like this? 👇🏼 

```js
const reasons = {
	custom: 'custom media queries can only be used in boolean queries',
	min_max_in_range: '"min-" and "max-" prefixes are not needed when using range queries',
	min_max_in_boolean: '"min-" and "max-" prefixes are not needed in boolean queries',
	discrete: 'discrete features can only be used in plain and boolean queries',
};

// ...

complain(atRule, atRuleParamIndexValue, parent, 'custom');

// ...

/**
 * @param {keyof reasons} [reason]
 */
function complain(atRule, index, node, reason) {
	report({
		// ...
		messageArgs: [node.toString(), reason ? reasons[reason] : '']
		// ...
	});
}
```

lib/rules/media-query-no-invalid/index.mjs
```diff
@@ -104,30 +127,37 @@ const rule = (primary) => {
 
 						if (HAS_MIN_MAX_PREFIX.test(name)) {
 							// In a range context, min- and max- prefixed feature names are invalid
```
ybiquitous, line 129: [ask] Is this "In a range context" in the comment above incorrect (boolean context)?

romainmenke: Yes, it is indeed incorrect.
I've changed this to :

```js
// In a boolean feature, min- and max- prefixed feature names are invalid
```

lib/rules/media-query-no-invalid/index.cjs
```diff
@@ -97,34 +107,36 @@ const rule = (primary) => {
 
 					// Invalid boolean media features.
 					if (mediaQueryListParser.isMediaFeatureBoolean(node)) {
-						const name = node.getName().toLowerCase();
+						const name = node.getName();
```
romainmenke, line 110: `.toLowerCase()`

Wasn't needed here because `HAS_MIN_MAX_PREFIX` is already case insensitive and this branch doesn't do any other checks on `name`

romainmenke:
Thank you for the feedback @Mouvedia, @ybiquitous 🙇 
I've made some changes following your suggestions.

ybiquitous (approved):
Thank you. LGTM 👍🏼

.changeset/healthy-donkeys-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
```
ybiquitous, line 2: Oops. A typo, right?

```suggestion
"stylelint": minor
```

romainmenke: good catch, thanks!

ybiquitous (approved):

PR #7468, opened 2024-01-17
Fix `overrides.files` negated pattern regression

### LINKED ISSUE #7056: Fix `overrides.files` negated pattern regression

Create file `fontSizeAllowedFile.tsx`:

```ts
import { css } from '@emotion/react';

export const someStyling = css`
  font-size: 16px;
  color: white;
`;
```

Using glob pattern to exclude files (`!(files)`) does not work in Stylelint 15. It does work in previous versions.

Using the positive glob does include only this file `["*/**/fontSizeAllowedFile.tsx"]`, just negating it does not work.

```json
{
  "rules": {
    "property-disallowed-list": null
  },
  "overrides": [
    {
      "files": ["!(*/**/fontSizeAllowedFile.tsx)"],
      "rules": {
        "property-disallowed-list": ["font-size", "font-weight"]
      }
    }
  ]
}
```

Via nx stylelint command

```json 
{
    "postcss-styled-syntax": "^0.4.0",
    "stylelint": "^15.10.1",
    "stylelint-config-standard": "^33.0.0"
}
```

File should not run disallowed list on the excluded file.

### What actually happened?

File is not excluded
```
 4:3  ✖  Unexpected property "font-size"    property-disallowed-list
```

_No response_

ybiquitous:

Closes #7056

No, it's self-explanatory.

### DIFF
--- a/.changeset/ten-rivers-hammer.md
+++ b/.changeset/ten-rivers-hammer.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `overrides.files` negated pattern regression
+Fixed: `overrides.files` negated pattern regression introduced in 15.0.0
--- a/.changeset/ten-rivers-hammer.md
+++ b/.changeset/ten-rivers-hammer.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `overrides.files` negated pattern regression
--- a/lib/__tests__/overrides.test.mjs
+++ b/lib/__tests__/overrides.test.mjs
@@ -358,3 +358,24 @@ it('two files', async () => {
 	expect(linted.results[1].warnings).toHaveLength(1);
 	expect(linted.results[1].warnings[0].rule).toBe('property-disallowed-list');
 });
+
+it('negation glob pattern', async () => {
+	const { results } = await standalone({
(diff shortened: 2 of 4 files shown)


.changeset/ten-rivers-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `overrides.files` negated pattern regression
```
Mouvedia, line 5: medium

I would add the version—15.0.0—that introduced the regression.
It's an important information.

ybiquitous: Good catch. How about this? 👇🏼 

```suggestion
Fixed: `overrides.files` negated pattern regression since 15.0.0
```

Mouvedia: _since_ sounds fine to me.
_introduced in_ works as well.

ybiquitous: Thanks. I'll apply your suggestion. 👍🏼 

```suggestion
Fixed: `overrides.files` negated pattern regression introduced in 15.0.0
```

jeddy3 (approved):
LGTM, thank you.

PR #7474, opened 2024-01-20
Fix `report()` error message responsibility for a missing node or line number

### LINKED ISSUE #7172: Fix unexpected error message in `report()`

Empty sass file

```json
{
    "customSyntax": "postcss-sass",
}
```

```shell
npx stylelint -f verbose "**/*.sass"
```

```json
  "devDependencies": {
    "postcss-sass": "^0.5.0",
    "stylelint": "^15.10.3",
    "stylelint-config-standard": "^34.0.0"
  }
```

No problem to report

### What actually happened?

Stylelint crashed:
```
Error: You must pass either a node or a line number
    at report (/home/georges/data/tmp/dev/stylelint-empty-file/node_modules/stylelint/lib/utils/report.js:42:9)
    at /home/georges/data/tmp/dev/stylelint-empty-file/node_modules/stylelint/lib/rules/no-empty-source/index.js:32:3
    at /home/georges/data/tmp/dev/stylelint-empty-file/node_modules/stylelint/lib/lintPostcssResult.js:121:8
    at Array.map (<anonymous>)
    at lintPostcssResult (/home/georges/data/tmp/dev/stylelint-empty-file/node_modules/stylelint/lib/lintPostcssResult.js:111:18)
    at lintSource (/home/georges/data/tmp/dev/stylelint-empty-file/node_modules/stylelint/lib/lintSource.js:110:8)
    at async /home/georges/data/tmp/dev/stylelint-empty-file/node_modules/stylelint/lib/standalone.js:211:27
    at async Promise.all (index 1)
    at async standalone (/home/georges/data/tmp/dev/stylelint-empty-file/node_modules/stylelint/lib/standalone.js:254:22)
```

Stylelint fails to process an empty sass file.

Again, I am not sure if it comes from stylelint or postcss-sass plugin. It works properly with the `postcss-scss` plugin though.

Also, the error message is not very practical, it does not even specify which file made stylelint crash, even with the `-f verbose` option. Any way to improve debugging messages?

ybiquitous:

Closes #7172

Including a rule name in the error message helps us understand why an error happens.

### DIFF
--- a/.changeset/weak-maps-tease.md
+++ b/.changeset/weak-maps-tease.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: unexpected error message in `report()`
--- a/.changeset/weak-maps-tease.md
+++ b/.changeset/weak-maps-tease.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: unexpected error message in `report()`
+Fixed: `report()` error message responsibility for a missing node or line number
--- a/lib/utils/report.mjs
+++ b/lib/utils/report.mjs
@@ -43,7 +43,9 @@ export default function report(problem) {
 	const startLine = line || (start && start.line);
 
 	if (!startLine) {
-		throw new Error(`The "${ruleName}" rule failed to pass either a node or a line number to the `report()` function.`);
+		throw new Error(
+			`The "${ruleName}" rule failed to pass either a node or a line number to the \`report()\` function.`,
+		);
 	}
 
 	const { ignoreDisables } = result.stylelint.config || {};
--- a/lib/utils/report.mjs
+++ b/lib/utils/report.mjs
@@ -43,7 +43,7 @@ export default function report(problem) {
 	const startLine = line || (start && start.line);
 
 	if (!startLine) {
-		throw new Error('You must pass either a node or a line number');
+		throw new Error(`You must pass either a node or a line number (rule: ${ruleName})`);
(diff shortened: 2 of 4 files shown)


lib/utils/report.mjs
```diff
@@ -43,7 +43,7 @@ export default function report(problem) {
 	const startLine = line || (start && start.line);
 
 	if (!startLine) {
-		throw new Error('You must pass either a node or a line number');
+		throw new Error(`You must pass either a node or a line number (rule: ${ruleName})`);
```
jeddy3, line 46: Am I right in thinking that only the (custom) rule author can resolve this issue and not the user? If so, let's rewrite the error to address the original problem: lack of clarity around the message.

```suggestion
		throw new Error(`The "${ruleName}" (custom) rule failed to pass either a node or a line number to the `report()` function.`);
```

.changeset/weak-maps-tease.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: unexpected error message in `report()`
```
jeddy3, line 5:
```suggestion
Fixed: `report()` error message clarity for missing node or line number
```

lib/utils/report.mjs
ybiquitous: Nice catch! Your suggestion is much better. 👍🏼

jeddy3 (approved):
LGTM, thank you.

PR #7475, opened 2024-01-20
Fix `selector-max-specificity` false positives with `ignoreSelectors` option for `of <selector>` syntax

### LINKED ISSUE #6615: Fix `selector-max-specificity` false positives with `ignoreSelectors` option for `of <selector>` syntax

The following second CSS snippet (selector) reports a `selector-max-specificity` problem:

```css
/* pass */
foo:nth-child(even) { color: red; }

/* fail */
foo:nth-child(even of a) { color: red; }
```

```json
{
  "rules": {
    "selector-max-specificity": ["0,1,0", { "ignoreSelectors": ["/foo/"] }]
  }
}
```

CLI

`14.16.1`

No problems to be reported by the `ignoreSelectors` option.

### What actually happened?

The following problem is reported:

```
5:1  Expected "foo:nth-child(even of a)" to have a specificity no more than "0,1,0"
```

See the [demo](https://stylelint.io/demo#N4Igxg9gJgpiBcID0AqABABwIYGcdpSQB0A7AMwgnhIBcALAWjDoEsAbKAChgDcYSAlGmBpIbCACd4aCTCgBuNAF9SpVGjJZ2BYuUrV6TVh258SaCGTRYhIsZOmyFy0iAA0IHAE9aWAB4I4Hju4BDkLADmgcCkaGhEIBIArmwwOAnSMeZx8Z4wqWA0kgwAtv4MOBgwYCxkLDU0XhloANoJAAxuAIxu7QluwrmRJJIwAMr51UUS6QitCUgUEMQgALrKq7EuJEogSkA).

_No response_

### Proposal to fix the bug

Relevant code:

- https://github.com/stylelint/stylelint/blob/b3af7d2bb8cce81bb7e51e81d980274bb4d14643/lib/rules/selector-max-specificity/__tests__/index.js#L463-L466
- https://github.com/stylelint/stylelint/blob/b3af7d2bb8cce81bb7e51e81d980274bb4d14643/lib/rules/selector-max-specificity/index.js#L129-L131

See also #6140

ybiquitous:
Config:

```json
{"ignoreSelectors": [".foo"]}
```

CSS code:

```css
:nth-child(even of .foo) {}

/* ↓ Calculates specificity ignoring 'of .foo' in `ignoreSelectors` */
:nth-child(even) {}
```

Closes #6615

No, it's self-explanatory.

### DIFF
--- a/lib/rules/selector-max-specificity/index.cjs
+++ b/lib/rules/selector-max-specificity/index.cjs
@@ -119,10 +119,10 @@ const rule = (primary, secondaryOptions) => {
 		 */
 		const ignoreOfSelectorIfAny = (pseudo) => {
 			/** @type {(node: import('postcss-selector-parser').Node | undefined) => boolean} */
-			const isOfSelector = (node) => Boolean(node && node.type === 'tag' && node.value === 'of');
+			const isOfSelector = (node) => node?.type === 'tag' && node.value === 'of';
 
 			/** @type {(node: import('postcss-selector-parser').Node | undefined) => boolean} */
-			const isSpace = (node) => Boolean(node && node.type === 'combinator' && node.value === ' ');
+			const isSpace = (node) => node?.type === 'combinator' && node.value === ' ';
 
 			const nodes = pseudo.nodes[0]?.nodes ?? [];
 			const ofSelectorIndex = nodes.findIndex((child, i, children) => {
--- a/lib/rules/selector-max-specificity/index.cjs
+++ b/lib/rules/selector-max-specificity/index.cjs
@@ -77,6 +77,10 @@ const rule = (primary, secondaryOptions) => {
 			return;
 		}
 
+		/** @type {(selector: string) => boolean} */
+		const isSelectorIgnored = (selector) =>
+			optionsMatches(secondaryOptions, 'ignoreSelectors', selector);
+
 		/**
 		 * Calculate the specificity of a simple selector (type, attribute, class, ID, or pseudos's own value).
 		 *
@@ -84,7 +88,7 @@ const rule = (primary, secondaryOptions) => {
 		 * @returns {Specificity}
 		 */
 		const simpleSpecificity = (node) => {
-			if (optionsMatches(secondaryOptions, 'ignoreSelectors', node.toString())) {
+			if (isSelectorIgnored(node.toString())) {
 				return zeroSpecificity();
 			}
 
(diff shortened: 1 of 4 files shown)


jeddy3 (approved):

lib/rules/selector-max-specificity/index.cjs
```diff
@@ -104,6 +108,45 @@ const rule = (primary, secondaryOptions) => {
...
+			/** @type {(node: import('postcss-selector-parser').Node | undefined) => boolean} */
+			const isOfSelector = (node) => Boolean(node && node.type === 'tag' && node.value === 'of');
+
+			/** @type {(node: import('postcss-selector-parser').Node | undefined) => boolean} */
+			const isSpace = (node) => Boolean(node && node.type === 'combinator' && node.value === ' ');
+
+			const nodes = pseudo.nodes[0]?.nodes ?? [];
+			const ofSelectorIndex = nodes.findIndex((child, i, children) => {
+				// Find ' of <selector>' nodes
+				return isSpace(child) && isOfSelector(children[i + 1]) && isSpace(children[i + 2]);
```
Mouvedia, line 130: minor nit

not sure if the `node.type ===` check are needed
which could allow a refactoring
e.g. `const foo = (node, value) => Boolean(node && && node.value === value);`

lib/rules/selector-max-specificity/index.mjs
```diff
@@ -105,6 +109,45 @@ const rule = (primary, secondaryOptions) => {
...
+			/** @type {(node: import('postcss-selector-parser').Node | undefined) => boolean} */
+			const isOfSelector = (node) => Boolean(node && node.type === 'tag' && node.value === 'of');
+
+			/** @type {(node: import('postcss-selector-parser').Node | undefined) => boolean} */
+			const isSpace = (node) => Boolean(node && node.type === 'combinator' && node.value === ' ');
+
+			const nodes = pseudo.nodes[0]?.nodes ?? [];
+			const ofSelectorIndex = nodes.findIndex((child, i, children) => {
+				// Find ' of <selector>' nodes
+				return isSpace(child) && isOfSelector(children[i + 1]) && isSpace(children[i + 2]);
```
Mouvedia, line 131: minor nit

not sure if the `node.type ===` checks are needed
which could allow a refactoring
e.g. `const foo = (node, value) => Boolean(node?.value === value);`

ybiquitous: Yeah, checking `node.type` is necessary for safer parsing. I've refactored the code via 9e0fea060 as you suggested. 👍🏼

Mouvedia: > for safer parsing

That's reasonable.

PR #7476, opened 2024-01-20
Fix `validateOptions` to report when secondary option object is an empty object or null

### LINKED ISSUE #6154: Fix `validateOptions` to report when secondary option object is an empty object or null

First, set up a rule that requires via `validateOptions` that the secondary option object is defined. I have an example here: https://github.com/ashleemboyer/stylelint-validate-options-test/blob/main/disallow-variable-name.js#L28-L44.

Second, either with Jest tests or the CLI tool & a config file, try running Stylelint with an empty object as the secondary option to the rule. I have an example here: https://github.com/ashleemboyer/stylelint-validate-options-test/blob/main/disallow-variable-name.test.js#L12-L28.

For a custom rule named `disallow-variable-name`, for example, two configurations that don't go as expected:

Secondary option is an empty object:
```json
{
  "plugins": ["./disallow-variable-name.js"],
  "rules": {
    "my-test/disallow-variable-name": [true, {}],
  },
}
```

Secondary option is `null`:
```json
{
  "plugins": ["./disallow-variable-name.js"],
  "rules": {
    "my-test/disallow-variable-name": [true, null],
  },
}
```

With the standalone `lint` in a Jest test: https://github.com/ashleemboyer/stylelint-validate-options-test/blob/main/disallow-variable-name.test.js

14.9.1

I would expect `validateOptions` to treat an empty object the same way it treats `undefined` and an empty array as "nothing possible" here: https://github.com/stylelint/stylelint/blob/e870430a37a5df5b077a2a54f2470c1b82f6e461/lib/utils/validateOptions.js#L73-L74.

### What actually happened?

When [the suggested steps for writing a plugin](https://stylelint.io/developer-guide/plugins#the-anatomy-of-a-plugin) are followed for using `validateOptions` (code block below), and the rule returns early if `validateOptions` returns a falsy value, the rest of the rule continues to run. If a rule depends on and expects a certain option to be set, a runtime error happens where that option is referenced.

See example error related to other links I've shared: https://github.com/ashleemboyer/stylelint-validate-options-test/runs/6941567039?check_suite_focus=true#step:4:17

```js
module.exports = stylelint.createPlugin(
  ruleName,
  (primaryOption, secondaryOptionObject) => {
    return (postcssRoot, postcssResult) => {
      const validOptions = stylelint.utils.validateOptions(
        postcssResult,
        ruleName,
        {
          /* .. */
        }
      );

      if (!validOptions) {
        return;
      }

      // ... some logic ...
      stylelint.utils.report({
        /* .. */
      });
    };
  }
);
```

_No response_

### Proposal to fix the bug

Update `nothingPossible` in `validateOptions` to check for empty objects, like it checks for empty arrays. Also proposes checking `possible === null` in addition to `possible === undefined` since a value of `null` can cause runtime errors too.

Current definition ([link to code](https://github.com/stylelint/stylelint/blob/e870430a37a5df5b077a2a54f2470c1b82f6e461/lib/utils/validateOptions.js#L73-L74))):

```js
const nothingPossible =
  possible === undefined ||
  (Array.isArray(possible) && possible.length === 0);
```

Proposed definition:

```js
const nothingPossible =
  possible === undefined ||
  possible === null ||
  (Array.isArray(possible) && possible.length === 0) ||
  (isPlainObject(possible) && Object.keys(possible).length === 0);
```

ybiquitous:

Closes #6154

No, it's self-explanatory.

### DIFF
--- a/lib/utils/validateOptions.mjs
+++ b/lib/utils/validateOptions.mjs
@@ -41,23 +41,23 @@ export default function validateOptions(result, ruleName, ...optionDescriptions)
  * @param {string} ruleName
  * @param {(message: string) => void} complain
  */
-function validate(opts, ruleName, complain) {
-	const possible = opts.possible;
-	const actual = opts.actual;
-	const optional = opts.optional;
-
+function validate({ possible, actual, optional }, ruleName, complain) {
 	if (actual === false && !ruleName.startsWith('report')) {
 		return complain(
 			`Invalid option value "false" for rule "${ruleName}". Are you trying to disable this rule? If so use "null" instead`,
 		);
 	}
 
+	// `null` means to turn off a rule.
 	if (actual === null || arrayEqual(actual, [null])) {
 		return;
 	}
 
 	const nothingPossible =
-		possible === undefined || (Array.isArray(possible) && possible.length === 0);
+		possible === undefined ||
+		possible === null ||
+		(Array.isArray(possible) && possible.length === 0) ||
+		(isPlainObject(possible) && Object.keys(possible).length === 0);
 
 	if (nothingPossible && actual === true) {
 		return;
--- a/.changeset/olive-experts-breathe.md
+++ b/.changeset/olive-experts-breathe.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `validateOptions` to report when secondary option object is an empty object or null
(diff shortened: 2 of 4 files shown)


jeddy3 (approved):

lib/utils/validateOptions.mjs
```diff
@@ -41,23 +41,23 @@ export default function validateOptions(result, ruleName, ...optionDescriptions)
...
 
+	// `null` means to turn off a rule.
 	if (actual === null || arrayEqual(actual, [null])) {
 		return;
 	}
 
 	const nothingPossible =
-		possible === undefined || (Array.isArray(possible) && possible.length === 0);
+		possible === undefined ||
+		possible === null ||
```
Mouvedia, line 58:
```suggestion
		possible == null ||
```

PR #7478, opened 2024-01-20
Fix `custom-property-no-missing-var-function` false positives for properties that can contain author-defined identifiers

### LINKED ISSUE #5742: Fix `custom-property-no-missing-var-function` false positives for properties that can contain author-defined identifiers

Thanks to the CSS Houdini API, the property name of a custom property could appear in the value of some declarations, such as `transition` (or more precisely, `transition-property`). However, the current `custom-property-no-missing-var-function` rule will report errors for these custom property names that are not wrapped in `var()`.

[Online demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIACAHAThFM0BcCeABALREBmEEBwAOgHYEEDOetOAhgB7wEDkAPJAA2ENAD4eAbjoMAlrQAW2GTkbdSbQYxhT6BOcpkaiANw0BXGNxxo2tRijZoYrHQF86deBgg4q04mQU3E5QOgzWtowGELTcJOSUAAwAdABMjG4etGx+uvFBBABGghaZtHRs8PIQxti5DPkQ3GB4tmUgADQgzKycCOCMjJ3gMaQyAOb9NLrUIGhmgjCMs9zTDAyzYGaMOBAAtkTomNj4RLQQRHsyg3LjJo5kZrRgODIxKwTWFv7utK4griAA)

One possibility is to not report `transition`, etc. by default, but this could also produce breaking changes.

It is also possible to add options to the rule to ignore certain property names, as we did in `property-no-vendor-prefix`.

```js
module.exports = {
  rules: {
    'custom-property-no-missing-var-function': [true, { ignoreProperties: ['transition', 'transition-property'] }],
  }
}
```

ybiquitous:

Closes #5742

No, it's self-explanatory.

### DIFF
--- a/.changeset/curly-seals-confess.md
+++ b/.changeset/curly-seals-confess.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `custom-property-no-missing-var-function` false positives for properties that can contain author-defined identifiers
--- a/lib/rules/custom-property-no-missing-var-function/README.md
+++ b/lib/rules/custom-property-no-missing-var-function/README.md
@@ -10,7 +10,10 @@ Disallow missing `var` function for custom properties.
  *             This custom property */
 ```
 
-This rule only reports custom properties that are defined within the same source.
+This rule has the following limitations:
+
+- It only reports custom properties that are defined within the same source.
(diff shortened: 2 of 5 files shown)


jeddy3 (approved):
LGTM, thank you!

PR #7483, opened 2024-01-21
Fix report flags not reporting on subsequent runs when cache is used

### LINKED ISSUE #5049: Fix report flags not reporting on subsequent runs when cache is used
> Clearly describe the bug

When using the `--rd` switch, if there is a needless disable, the second or any subsequent run doesn't report it.

None

Probably doesn't matter:

```json
{
  "extends": [
  ],
  "rules": {
    "function-disallowed-list": [
      "calc",
      "lighten",
      "darken"
    ],
    "property-disallowed-list": [
      "border-radius",
      "border-top-left-radius",
      "border-top-right-radius",
      "border-bottom-right-radius",
      "border-bottom-left-radius",
      "transition"
    ],
    "scss/dollar-variable-default": [
      true,
      {
        "ignore": "local"
      }
    ],
    "scss/selector-no-union-class-name": true
  }
}
```

13.7.2 and 13.8.0

CLI: `stylelint "**/*.{css,scss}" --cache --cache-location .cache/.stylelintcache --rd`

Probably not

Any needless disables should be reported on subsequent runs.

No needless disables were reported. **If I delete the cache or remove the `--cache` CLI flag, then the needless disables are reported**.

@jeddy3 I verified this and indeed happens with 13.7.0, 13.7.2 and 13.8.0 that I could quickly test.

ybiquitous:
The bug's reason is that the `stylelintError`/`stylelintWarning` properties in a lint result object are not set when `styelint-disable` comment problems are reported.

Also, this change refactors the code around `styelint-disable` comments to make it more readable and prevent possible bugs.

The new code uses [`Result#warn()`](https://postcss.org/api/#result-warn) in PostCSS for a comment node, like reporting a normal rule problem.

Closes #5049

No, it's self-explanatory.


lib/reportDisables.mjs
```diff
@@ -1,58 +1,43 @@
...
-		}
+			if (!reportDisablesForRule(configRules[rule])) continue;
 
-		for (const [rule, ranges] of Object.entries(rangeData)) {
-			for (const range of ranges) {
-				if (!reportDisablesForRule(config.rules[rule] || [])) continue;
-
-				// If the comment doesn't have a location, we can't report a useful error.
-				// In practice we expect all comments to have locations, though.
-				if (!range.comment.source || !range.comment.source.start) continue;
```
Mouvedia, line 38: question

Do all comments have "locations"?

Mouvedia:
> Also, this change refactors the code around styelint-disable comments to make it more readable and prevent possible bugs.

I think keeping the refactors in a separate commit would help the reviewers.

lib/reportDisables.mjs
ybiquitous: Do you mean the comments "// If the comment doesn't have a location, we can't report..."? If so, yes.

ybiquitous:
> I think keeping these refactors in a separate commit would help the reviewers.

Sorry. I forgot to do that at this time. 🙏🏼

jeddy3 (approved):
LGTM, thank you!

Issue #7495, opened 2024-01-24
Fix `selector-max-specificity` README examples of the `ignoreSelectors` secondary option


romainmenke:
see : https://github.com/stylelint/stylelint/blob/382961fb6eb67557910a1ff7430a7a180f888bcc/lib/rules/selector-max-specificity/__tests__/index.mjs#L451

These tests are problematic for multiple reasons:
- `:global` and `:local` are non-standard and it isn't defined what their specificity is.
- `:global` and `:local` don't have any specificity of their own in `@csstools/selector-specificity`, so ignoring them to reduce their specificity seems weird?
- `:global` and `:local` do not support selector lists in implementations
  - `lightningcss` errors
  - `postcss-modules` mangles the output (`:global(.page, .bar)` -> `.page .bar`)

They also dominate the implementation of `selector-max-specificity`.

As far as I can tell the `ignoreSelectors` was also created with css modules too much in mind.
It acts like a hole punch. The AST node that matches doesn't contribute to specificity, but its child nodes do. Which works really well for `:global` and `:local` because they happen to be removed during compilation.

But I am not sure if this makes sense for any case other than css modules.
There isn't any test coverage for other pseudo selectors with child nodes, so I can't tell what the exact intention was here.

--------------

Proposal :
- fix the tests so that `:global` and `:local` are used as supported by implementations
- better define how `ignoreSelectors` works when it matches AST nodes with child nodes

ybiquitous:
@romainmenke Thanks for the report. `ignoreSelectors` seemed to be added by PR #2874, quite a while ago (2017)! I'm honestly unsure about the background, but it seems a good opportunity to reconsider CSS Modules selectors (e.g. `:global`). 👍🏼 

In addition, thanks for trying `@csstools/selector-resolve-nested` with PR #7496!

romainmenke:
Regressions or CSS modules compat are not really my primary concern here :)

CSS modules are supported in `@csstools/selector-specificity` so there is nothing that needs to happen for that in Stylelint when calculating specificity.

But the test however write CSS for CSS modules that isn't correct.
This is just bogus : 

```css
:global(.page, .bar)
:global(.page, :local(.bar))
```

Because these are bogus we can safely assume that no user will be writing this and that changes won't cause regressions.

--------------------

At the same time these test are the only coverage we have for `ignoreSelectors` in combination with functional pseudo classes.

There is no coverage with standard functional pseudo classes and `ignoreSelectors`.

[...]

PR #7496, opened 2024-01-24
Demo implementation of `@csstools/selector-resolve-nested`

romainmenke:

I wanted to gain a better understanding of the complexity that CSS nesting introduces for all rules that analyse selectors.

I also want to validate how `@csstools/selector-resolve-nested` works to make sure it works correctly and its API has the right shape.

This is a draft PR and it isn't intended to be merged.


lib/rules/selector-max-specificity/index.cjs
```diff
@@ -27,29 +29,127 @@ const meta = {
...
+			node.replaceWith(selectorParser.universal({
+				value: '*',
+				sourceIndex: node.sourceIndex,
+				source: node.source,
+			}));
+		}
+	});
+
+	return clone;
+}
```
romainmenke, line 71: I removed all the custom logic for specificity and instead manipulated the AST so that ignored nodes were replaced with `*` which has zero specificity. This has obvious differences but I think it can be an interesting approach to `ignoreSelector` in this context.

lib/rules/selector-max-specificity/index.cjs
```diff
@@ -27,29 +29,127 @@ const meta = {
...
+ * @typedef {import('postcss-selector-parser').Selector} Selector
+ * @typedef {import('postcss-selector-parser').Root} SelectorRoot
+ * @param {import('postcss').Rule} rule
+ * @param {import('stylelint').PostcssResult} result
+ * @param {(selector: string) => boolean} isIgnored
+ * @returns {Array<{selector: Selector, resolvedSelectors: Array<Selector>}> | undefined}
  */
-const specificitySum = (specificities) => {
-	const sum = zeroSpecificity();
+function gatherSelectorsForRule(rule, result, isIgnored) {
```
romainmenke, line 81: This should be a shared helper.

It walks the `postcss` AST to gather all the ancestor nodes and resolves all the nested selectors.

lib/rules/selector-max-specificity/index.cjs
```diff
@@ -27,29 +29,127 @@ const meta = {
...
+			// `.foo, #bar { @media screen { color: red; } }`
+			// equivalent to
+			// `@media screen { .foo, #bar { & { color: red; } } }`
+			// `@media screen { :is(.foo, #bar) { color: red; } }`
+			const childAST = selectorAST('&', result, child);
+
+			if (!childAST) return;
+
+			ast = selectorResolveNested.resolveNestedSelector(childAST, ast);
+		}
```
romainmenke, line 133: for more info and demo's see : https://github.com/mrhenry/stylelint-mrhenry/issues/66

lib/utils/selectorAST.mjs
```diff
@@ -0,0 +1,33 @@
...
+		// }
+
+		// if ('walk' in root) {
+		// 	root.walk((x) => {
+		// 		if (x.source) {
+		// 			// @ts-ignore the `node` field doesn't exist on the source type.
+		// 			x.source.node = node;
+		// 		}
+		// 	});
+		// }
```
romainmenke, line 25: _bogus code I forgot to cleanup._

romainmenke (commented):
This currently errors, mostly because of : https://github.com/stylelint/stylelint/issues/7495

jeddy3:
This looks fantastic and is a significant step towards spec compliance!

lib/utils/resolveNestedSelectorsForRule.cjs
```diff
@@ -0,0 +1,94 @@
...
+			// equivalent to
+			// `@media screen { .foo, #bar { & { color: red; } } }`
+			// `@media screen { :is(.foo, #bar) { color: red; } }`
+			const childAST = selectorAST('&', result, child);
+
+			if (!childAST) return;
+
+			ast = selectorResolveNested.resolveNestedSelector(childAST, ast);
+		}
+	}
```
romainmenke, line 71: Note : the specification has since changed again, this is no longer true.

romainmenke:
See : https://developer.mozilla.org/en-US/docs/Web/CSS/@scope#the_difference_between_scope_and_inside_scope

Resolving / Flattening nested selectors should also take `@scope` into account.

romainmenke:
> Resolving / Flattening nested selectors should also take @scope into account.

This is really tricky to implement.
`@scope` start selectors only impact the outcome when having an explicit `&` selector in a child rule. (e.g. `@scope (.a) { & { color: red; } }`)

But there isn't really a way to syntactically express all of `@scope` without using `@scope` itself.

Simply combining the selectors found in both rules and `@scope`, as we do for nested rules, would be incorrect in many cases.

I suggest we ignore `@scope` for now and first see if users encounter any issues.

PR #7504, opened 2024-01-30
Fix benchmark script

ybiquitous:
This fixes the broken benchmark script (`npm run benchmark`)
The error reason is that `rules[ruleName]` has returned a `Promise` since 16.0.0.

```sh-session
$ npm run benchmark-rule -- value-keyword-case lower
...
const rule = ruleFunc(primary, secondary, context);
             ^

TypeError: ruleFunc is not a function
...
```

Also, this changes `parserOptions.ecmaVersion` in the ESLint config to allow top-level `await`.

None.

If we release `eslint-config-stylelint` including stylelint/eslint-config-stylelint#252 and bump the package version first, the ESLint config's change in this PR will be needless.

Of course, we can revisit it later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,9 @@
   },
   "prettier": "@stylelint/prettier-config",
   "eslintConfig": {
+    "parserOptions": {
+      "ecmaVersion": 2023
+    },
     "extends": [
       "stylelint",
       "stylelint/jest"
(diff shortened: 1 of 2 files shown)


ybiquitous:
Success! 🎉 

```sh-session
$ npm run benchmark-rule -- value-keyword-case lower

> stylelint@16.2.0 benchmark-rule
> node scripts/benchmark-rule.mjs value-keyword-case lower

Warnings: 120
Mean: 223.59690633333332 ms
Deviation: 14.505605610006747 ms
```

romainmenke (approved):
LGTM!

PR #7505, opened 2024-01-30
Fix `function-calc-no-unspaced-operator` performance

ybiquitous:

Follow-up to #6923

Wait for #7504 (Note that the base branch is `fix-benchmark-script`)

How to measure:

```shell
npm run -s benchmark-rule -- function-calc-no-unspaced-operator true
```

Before:

```
Warnings: 0
Mean: 125.72145828571429 ms
Deviation: 14.960783775263804 ms
```

After:

```
Warnings: 0
Mean: 110.21394642857142 ms
Deviation: 10.1318139617483 ms
```

### DIFF
--- a/lib/rules/function-calc-no-unspaced-operator/index.cjs
+++ b/lib/rules/function-calc-no-unspaced-operator/index.cjs
@@ -28,6 +28,7 @@ const meta = {
 const OPERATORS = new Set(['+', '-']);
 const OPERATOR_REGEX = /[+-]/;
 const ALL_OPERATORS = new Set([...OPERATORS, '*', '/']);
+const CALC_FUNC_REGEX = /calc\(/i;
 
 /** @type {import('stylelint').Rule} */
 const rule = (primary, _secondaryOptions, context) => {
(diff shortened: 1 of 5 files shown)


lib/rules/function-calc-no-unspaced-operator/index.cjs
```diff
@@ -53,6 +54,8 @@ const rule = (primary, _secondaryOptions, context) => {
 
 			if (!OPERATOR_REGEX.test(value)) return;
 
+			if (!CALC_FUNC_REGEX.test(value)) return;
```
romainmenke, line 57: Does this rule also check other match functions?

```
width: min((10px+20px), 5px);
```

ybiquitous: No, this rule checks only `calc()`. See also:

https://github.com/stylelint/stylelint/blob/7c0cc197f5a8040115810795a14ae275509e5e67/lib/rules/function-calc-no-unspaced-operator/index.mjs#L318

ybiquitous:
@romainmenke Thanks for the quick review! Can you take a quick look at #7504, too? It's a blocker for this PR.

PR #7509, opened 2024-02-01
Fix inaccurate end positions in `selector-max-type` (breaking)

romainmenke:

No, it's self-explanatory.


lib/rules/selector-max-type/__tests__/index.mjs
```diff
@@ -33,15 +33,6 @@ testRule({
...
 		},
-		{
-			code: '.foo { &-bar {} }',
-		},
-		{
-			code: '.foo { &__bar {} }',
-		},
-		{
-			code: '.foo { [&] {} }',
-		},
```
romainmenke, line 44: This can not be parsed and gives a warning, but it also isn't a valid selector.

I would prefer to just drop this test because bogus inputs will always give bogus outputs.

lib/rules/selector-max-type/__tests__/index.mjs
```diff
@@ -138,6 +129,24 @@ testRule({
...
+		},
+		{
+			code: '.foo { &__bar {} }',
+			description: '&__bar is an invalid selector, types must precede `&`',
+			message: messages.expected('&__bar', 0),
+			line: 1,
+			column: 8,
+			endLine: 1,
+			endColumn: 14,
+		},
```
romainmenke, line 149: This is a breaking change.

If people use this rule while writing sass style nesting they will get errors after updating. Given the popularity of sass style nesting it will definitely be noticed when we change this.

lib/rules/selector-max-type/__tests__/index.mjs
Mouvedia: Don't know if our policies have changed since 

> [Non-standard selectors should be ignored, and tests for them should be accepted.](https://github.com/stylelint/stylelint/pull/2665#issuecomment-311105801)

but shouldn't it fail gracefully?

lib/rules/selector-max-type/__tests__/index.mjs
Mouvedia: Maybe add a specific `if` to detect the pattern and `console.warn` a hint?

lib/rules/selector-max-type/__tests__/index.mjs
romainmenke: Are non-standard selectors all selectors that aren't standard, or are they only those that are valid in sass, less,...

Nuance between :
- completely bogus
- valid in a CSS flavor, but not in standard CSS

Mouvedia:
I am not seeing `selector-max-type` under https://github.com/stylelint-scss/stylelint-scss/tree/master/src/rules, does that mean it will have to be turned off?
e.g. `&__foo {}`

lib/rules/selector-max-type/__tests__/index.mjs
Mouvedia: I don't know but I think that we should at least communicate with `stylelint-scss` so that they can prepare a rule synchronised with our next release to offer as a drop-in solution to annoyed users.
That way you may drop anything that isn't standard.

romainmenke: I'm sorry, I think there is some confusion here :)

`[&]` is not a sass or less selector.

Screenshot 2024-02-01 at 21 00 38

It isn't a valid selector anywhere as far as I know.
If it is valid anywhere, please let me know because then my entire premise is incorrect.

`selector-max-type` already used and continues to use `isStandardSyntaxSelector` to ignore any selector that is detectable as sass, less, ...

That `[&]` was fine in this rule was a false negative, likely a result of `postcss-resolve-nested-selector` doing find/replace with strings.

lib/rules/selector-max-type/__tests__/index.mjs
romainmenke: Or should we add detection for this in `isStandardSyntaxSelector`?

`&__bar` and `&-bar` are never standard selectors, they can only be sass, less, ...
So I think `isStandardSyntaxSelector` should return false in this case.

I think this might be the least intrusive way forward.

lib/rules/selector-max-type/__tests__/index.mjs
Mouvedia: > Editted to clarify that this comment is only about [&]

Sorry my bad, I was commenting on the basis that it was about the lines removed from line 36 to 44 which contained.

`code: '.foo { &__bar {} }',`

lib/rules/selector-max-type/__tests__/index.mjs
Mouvedia: > are _never_ standard selectors

If true, it ought to be fixed in a separate commit.
IIRC this util is being exported so it might have consequences.
We shouldn't care though, utils ought to match their names.

ybiquitous: I agree with adding detection to `isStandardSyntaxSelector` by another pull request, as @Mouvedia suggested.

ybiquitous: I think this PR is the right approach we should take 👍🏼 
But one big concern is a breaking change for non-standard syntaxes, e.g., SCSS, as you considered.

My current guess is that we have to bring the breaking change in the next major version, especially for the SCSS users (I guess it's a big community).

romainmenke: Yup, definitely prefer saving this for a next major.

I do find these changes very informative.
There are a lot of past choices encoded in these rules and refactoring them is a good way to get a better feel for the complexity.

Maybe we can setup a long lived branch for selector rule refactors?
Its main purpose would be to explore the coming changes so that possible issues can be encountered early. Either we then pick specific parts of that branch or we delete it entirely and start over manually, but with a fresh and comprehensive understanding of the work that should be done.

Mouvedia: Maybe there are others selector-* rules that use `postcss-resolve-nested-selector` that don't have problematic test cases. If so, the regression won't be considered as severe because it wasn't acknowledged though the uses of test cases. It's a technicality but that would pass as sub-major change in my book.

ybiquitous: I agree with the refactoring using the nice package `"@csstools/selector-resolve-nested` unless it introduces breaking changes. 👍🏼 

I believe a new rule might need this package for spec-compliant selector parsing.

romainmenke: > I agree with the refactoring using the nice package "@csstools/selector-resolve-nested unless it introduces breaking changes.

I am a bit confused now :)
I think that any refactor using this package will have some amount of breaking changes and should be saved for a next major.

I think we have these options:
1. do not use this package at this time, save it for a next major. We close these PR's and focus purely on the issue of inaccurate end positions which might also have a different solution.
2. Start work early on the next major and start refactoring all selector rules to use this package. How exactly we work on the next major, this early, is a different topic :)
3. A mix of 1 and 2 where we fix inaccurate end positions in a way that facilitates a future switch to this package. This also implies having a good understanding of this package and how it will integrate in the future.

Option 1 might not find a solution and it can not solve issues with nesting. But it could be a non-breaking step forward.

Notes :
- I am not attached to these changes, the proposed package, ... These can all be discarded if they are not the right way forward.
- I am fine with shelving this work for a few months. Closing these PR's and reviving the effort later.

ybiquitous: Thank you for the detailed explanation. I misunderstood there was room to refactor via `@csstools/selector-resolve-nested` somewhere in the code base. 👍🏼 

We just released the current major version (16.0.0) a few months ago (Dec 2023), so I predict the next major release will be several months later. I guess staring the work for the next major version is a bit too early at this point.

Of course, I welcome non-breaking changes. Also, since end-position bugs are a bit less critical, we could address other works, e.g. considering the migration path from `postcss-resolve-nested-selector` to `@csstools/selector-resolve-nested`.

Mouvedia: For the record, I vote 3.

romainmenke: I have a vague idea of what 3 looks like.
I will try to prepare a PR with that shortly.

Then this is less abstract :)

romainmenke: see : https://github.com/stylelint/stylelint/pull/7518

romainmenke:
Closing this, but also stored the interesting bits in a gist here : https://gist.github.com/romainmenke/32b8e196d0bde94b4429710e3c6ada38

PR #7518, opened 2024-02-06
Fix `selector-max-type` end positions

romainmenke:

See : https://github.com/stylelint/stylelint/pull/7509

No, it's self-explanatory.


lib/rules/selector-max-type/__tests__/index.mjs
```diff
@@ -39,9 +39,6 @@ testRule({
 		{
 			code: '.foo { &__bar {} }',
 		},
-		{
-			code: '.foo { [&] {} }',
-		},
```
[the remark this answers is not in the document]
Mouvedia: ref https://stackoverflow.com/q/925994/248058

ybiquitous (commented):
@romainmenke Thanks for the pull request. Can you go on? I think this PR can be released as a patch version.

lib/utils/flattenNestedSelectorsForRule.cjs
```diff
@@ -0,0 +1,48 @@
...
+
+			flattenedSelectors.push({
+				selector: selectorAST.clone(),
+				resolvedSelectors: resolvedRoot.nodes,
+			});
+		}
+	}
+
+	return flattenedSelectors;
+}
```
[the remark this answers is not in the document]
romainmenke: Also saved the breaking implementation in a gist here : https://gist.github.com/romainmenke/32b8e196d0bde94b4429710e3c6ada38

romainmenke:
I was actually waiting for feedback here : https://github.com/stylelint/stylelint/issues/6234#issuecomment-1921970026
And then I lost track of these pull requests :)

But I guess if someone had major concerns they would have commented by now on that issue, so yea this is ready for review :)

lib/rules/selector-max-type/__tests__/index.mjs
```diff
@@ -109,25 +106,25 @@ testRule({
 			line: 1,
 			column: 13,
 			endLine: 1,
-			endColumn: 20,
+			endColumn: 21,
 		},
 		{
 			code: '.foo { div {} }',
```
Mouvedia, line 112: question

I suppose `code: 'p { div {} }',` gives `message: messages.expected('p div', 0),`?
Or does it count as 2 separate errors?

lib/rules/selector-max-type/index.mjs
Mouvedia, line 1: l198, l95, etc.

What about

```js
const { isCombinator, isTag } = require('postcss-selector-parser');
```

?

lib/utils/getSelectorAST.mjs
```diff
@@ -0,0 +1,19 @@
...
+ * @param {import('postcss').Node} node
+ * @returns {import('postcss-selector-parser').Root | undefined}
+ */
+export default function getSelectorAST(selector, result, node) {
+	if (!selector) return undefined;
+
+	try {
+		return selectorParser().astSync(selector);
+	} catch (err) {
+		result.warn(`Cannot parse selector (${err})`, { node, stylelintType: 'parseError' });
```
Mouvedia, line 15: nit

I would expect it to be in `err.message`.

Mouvedia (commented):
note

I wonder why there's no ignore option for the subsequent/general sibling combinator.

lib/rules/selector-max-type/__tests__/index.mjs
romainmenke: Currently it would be reported as `div`.
My thinking was that the resolved selector (`p div`) wouldn't exist in source.

So a user who got a warning for `p div` might do a find throughout their codebase for exactly `p div` and wouldn't get any results.

But it can be easily changed so that the resolved selector is reported instead or together with the current value.

I don't have a strong preference here.

lib/rules/selector-max-type/index.mjs
romainmenke: This beyond the scope of this change :)

lib/utils/getSelectorAST.mjs
romainmenke: This function mirrors `lib/utils/parseSelector.mjs`.
I'd prefer to keep this as is for now.

Maybe we should consider changing all instances of `catch (err) {}`?
But that is beyond the scope of this change.

lib/rules/selector-max-type/__tests__/index.mjs
Mouvedia: If you thought about it already, it's fine.

ybiquitous:
> I wonder why there's no ignore option for the subsequent/general sibling combinator.

I'm unsure, but I guess there are no requests from people.

lib/rules/selector-max-type/__tests__/index.mjs
```diff
@@ -109,25 +106,25 @@ testRule({
 			line: 1,
 			column: 13,
 			endLine: 1,
-			endColumn: 20,
+			endColumn: 21,
```
ybiquitous, line 109: [ask] `13` to `20` seems correct to me 👀 

```css
.foo, .bar, foo.baz {}
/*          ↑      ↑
12345678901234567890
*/
```

ybiquitous: Ah, I see. The selector is trimmed, right?

lib/utils/flattenNestedSelectorsForRule.mjs
```diff
@@ -0,0 +1,43 @@
+import resolvedNestedSelector from 'postcss-resolve-nested-selector';
```
ybiquitous, line 1: [suggest] Minor. Just a typo:

```suggestion
import resolveNestedSelector from 'postcss-resolve-nested-selector';
```

lib/utils/flattenNestedSelectorsForRule.mjs
```diff
@@ -0,0 +1,43 @@
...
+
+import getRuleSelector from './getRuleSelector.mjs';
+import getSelectorAST from './getSelectorAST.mjs';
+import isStandardSyntaxSelector from './isStandardSyntaxSelector.mjs';
+
+/**
+ * @typedef {import('postcss-selector-parser').Selector} Selector
+ * @typedef {import('postcss-selector-parser').Root} SelectorRoot
```
ybiquitous, line 9: [suggest] Remove the unused type:

```suggestion
```

lib/utils/flattenNestedSelectorsForRule.mjs
```diff
@@ -0,0 +1,43 @@
...
+ * @typedef {Array<{selector: Selector, resolvedSelectors: Array<Selector>}> | undefined} FlattenedSelectors
```
ybiquitous, line 10: [suggest] We can make `FlattenedSelectors` unnecessary:

```suggestion
```

lib/utils/flattenNestedSelectorsForRule.mjs
```diff
@@ -0,0 +1,43 @@
...
+ * @param {import('postcss').Rule} rule
```
ybiquitous, line 11: [suggest] Let's insert an empty line to distinguish the typedefs and the function spec:

```suggestion
 *
 * @param {import('postcss').Rule} rule
```

lib/utils/flattenNestedSelectorsForRule.mjs
```diff
@@ -0,0 +1,43 @@
...
+ * @param {import('stylelint').PostcssResult} result
+ * @returns {Array<{selector: Selector, resolvedSelectors: Array<Selector>}> | undefined}
```
ybiquitous, line 13: [suggest] Does returning an empty array instead of `undefined` make sense?

```suggestion
 * @returns {Array<{selector: Selector, resolvedSelectors: Array<Selector>}>}
```

lib/utils/flattenNestedSelectorsForRule.mjs
```diff
@@ -0,0 +1,43 @@
...
+ */
+export default function flattenNestedSelectorsForRule(rule, result) {
+	/** @type {FlattenedSelectors} */
```
ybiquitous, line 16:
```suggestion
```

lib/utils/flattenNestedSelectorsForRule.mjs
```diff
@@ -0,0 +1,43 @@
...
+	const flattenedSelectors = [];
+
+	const ownAST = getSelectorAST(getRuleSelector(rule), result, rule);
+
+	if (!ownAST) return;
+
+	for (const selectorAST of ownAST.nodes) {
```
ybiquitous, line 23: [suggest] Minor. I feel initializing `flattenedSelectors` lazily makes sense.

```suggestion
	const ownAST = getSelectorAST(getRuleSelector(rule), result, rule);

	if (!ownAST) return;

	const flattenedSelectors = [];

	for (const selectorAST of ownAST.nodes) {
```

lib/utils/flattenNestedSelectorsForRule.mjs
```diff
@@ -0,0 +1,43 @@
...
+		const resolvedSelectors = resolvedNestedSelector(selectorAST.toString(), rule);
+
+		for (const resolvedSelector of resolvedSelectors) {
+			if (!isStandardSyntaxSelector(resolvedSelector)) return;
```
ybiquitous, line 27: [ask] Just to confirm, does not skipping a non-standard selector make sense?

```suggestion
			if (!isStandardSyntaxSelector(resolvedSelector)) continue;
```

lib/utils/getSelectorAST.mjs
```diff
@@ -0,0 +1,19 @@
+import selectorParser from 'postcss-selector-parser';
+
+/**
+ * @param {string} selector
+ * @param {import('stylelint').PostcssResult} result
+ * @param {import('postcss').Node} node
+ * @returns {import('postcss-selector-parser').Root | undefined}
+ */
+export default function getSelectorAST(selector, result, node) {
```
ybiquitous, line 9: [suggest] Nice-to-have. This new function `getSelectorAST` is almost the same as `parseSelector`, so I feel the former will be replaced with the latter. For example:

```diff
-import parseSelector from '../../utils/parseSelector.mjs';
+import { parseSelectorAST } from '../../utils/parseSelector.mjs';

-parseSelector(resolvedSelector, result, ruleNode, (s) => {
+parseSelectorAST(resolvedSelector, result, ruleNode)?.each((s) => {
 	checkSelector(resolvedSelector, s, ruleNode, comparisonContext);
 });
```

So, how about adding this new function to `parseSelector.mjs` like this? I guess the migration is easier.

```js
// lib/utils/parseSelector.mjs

export default function parseSelector(selector, result, node, callback) {
	// ...
}

// Move from `getSelectorAST` to here.
export function parseSelectorAST(selector, result, node) {
	// ...
}
```

ybiquitous (commented):
I've left several refactoring suggestions, so can you take a look at them?

lib/utils/getSelectorAST.mjs
Mouvedia: > how about adding this new function to `parseSelector.mjs`

_If_ you do that, `getSelectorAST.test.mjs` will have to be renamed `parseSelector.test.mjs`.

lib/utils/flattenNestedSelectorsForRule.mjs
romainmenke: I would prefer to keep it as it currently is and skip the entire selector list.

I am concerned that the updated implementation using `@csstools/selector-resolve-nested` will not be able to handle partially non-standard selectors because it always first parses and then processes selectors.

lib/utils/getSelectorAST.mjs
romainmenke: I wanted to make this change, but it has the side effect of changing all the `.cjs` files that use `parseSelector ` 

Screenshot 2024-03-19 at 22 38 50


```diff
- parseSelector(resolvedSelector, result, ruleNode, (s) => checkSelector(s, ruleNode));
+ parseSelector.default(resolvedSelector, result, ruleNode, (s) => checkSelector(s, ruleNode));
```

-----

Options:
- leave as is
- separate file, but named `parseSelectorAST.mjs` and a function `parseSelectorAST`
- refactor `export default parseSelector` to `export parseSelector` in a separate change
- ...

lib/rules/selector-max-type/__tests__/index.mjs
romainmenke: Thank you for calling this out.

To be honest I am never sure about these and I am prone to having off by one errors here.

lib/utils/getSelectorAST.mjs
ybiquitous: > separate file, but named `parseSelectorAST.mjs` and a function `parseSelectorAST`

Okay. Can you choose the option above?

lib/utils/flattenNestedSelectorsForRule.mjs
ybiquitous: Make sense. Let's keep it as-is 👍🏼

lib/rules/selector-max-type/__tests__/index.mjs
romainmenke: I had an issue here, but I want to take a closer look again later.

I think the end index needs to be one less than it currently is.
The doc comment for `endIndex` also states that it is an exclusive index, so that would make sense.

But for `a {}` it is then `[0, 0]` and this causes other issues.
The final `endColumn` then becomes `5`.

Not sure if this makes sense and if someone knows what issue I've encountered? :)

ybiquitous: I'm sorry, I didn't understand your question well. 😓 

The `a {}` CSS code has the location `1:1-1:5` when parsed by PostCSS, right?

```sh-session
$ node -pe 'require("postcss").parse("a {}").source'
{
  input: Input {
    css: 'a {}',
    hasBOM: false,
    id: '<input css CWnXw->',
    [Symbol(fromOffsetCache)]: [ 0 ]
  },
  start: { column: 1, line: 1, offset: 0 },
  end: { column: 5, line: 1, offset: 4 }
}
```

romainmenke: This is a bug in both PostCSS and Stylelint as far as I can tell.

see:  https://github.com/postcss/postcss/blob/0fd1d86f2d33926c782463e63d8e1c816d207bc8/lib/node.js#L244-L248

see: https://github.com/stylelint/stylelint/blob/a02e4dd6d99e9159fd0f51d04caf324c1d990066/lib/utils/report.mjs#L111

If `endIndex` is `0` it is treated as if it isn't set, which isn't correct.

It is just really really rare for the `endIndex` to be `0` in CSS.
Only happens for a type selector with only a single `char`.

So users would have to start their CSS file with `a {}`, `i {}`, `b {}`, ... or something similar.

lib/utils/getSelectorAST.mjs
romainmenke: Done!

lib/rules/selector-max-type/__tests__/index.mjs
ybiquitous: [suggest] Thanks for the fix. However, it seems that we still need to fix the calculation of `endIndex` when trimming whitespaces in a selector string.

- `.length` has to be calculated before trimming
- `endIndex` has to be without `-1` (excluding)

```js
const selectorStr = selectorNode.toString().trim();
const index = selectorNode.first?.sourceIndex ?? 0;
const endIndex = index + selectorStr.length;

report({
	ruleName,
	result,
	node: ruleNode,
	message: messages.expected,
	index,
	endIndex,
});
```

```css
/* start: 31, end: 42 */
.foo + .bar { & .baz + .foo { & foo + bar {} } }
                              ^          ^
123456789012345678901234567890123456789012
```

Any thoughts?

romainmenke: 🤔 I am not sure I follow.

If I test these I get the ranges I would expect:

Screenshot 2024-03-21 at 19 34 26

The selector text is also only trimmed for the message. The length value is from the untrimmed selector.

But happy to make any needed adjusted, just trying to understand it better :)

ybiquitous: [`stylelint-codeframe-formatter`](https://www.npmjs.com/package/stylelint-codeframe-formatter) helps us see the reported position.

```jsonc
// .stylelintrc.json
{
  "rules": {
    "selector-max-type": 0
  }
}
```

```css
/* test.css */
.foo + .bar { & .baz + .foo { & foo + bar {} } }
```

Run:

```sh-session
$ npx stylelint test.css --custom-formatter node_modules/stylelint-codeframe-formatter/index.js
  test.css

  ✖ Expected "& foo + bar" to have no more than 0 type selectors (selector-max-type) at test.css:2:31
  1 | /* test.css */
> 2 | .foo + .bar { & .baz + .foo { & foo + bar {} } }
    |                               ^^^^^^^^^^
  3 |

  1 error
```

As the output above shows, the end position is short by one character.

romainmenke: It isn't fixed yet with my last commit, but I fully understand the issue now.
Will look at it later today, or early tomorrow.

Thank you for the extra information here 🙇

romainmenke: `sourceIndex` on `postcss-selector-parser` AST nodes doesn't include the length of `rawSpaceBefore`.

Any selector with a comma followed by a space (e.g. `foo, bar`) had unexpected results. `sourceIndex` on `bar` would be without the leading space. While the string representation would include the leading space.

This means that `sourceIndex` + `length` wouldn't be correct when used as `endIndex`.

Trimming all surrounding whitespace first and then taking the `length` seems to fix it.

@ybiquitous Please let me know if this makes sense :)

lib/rules/selector-max-type/__tests__/index.mjs
```diff
@@ -741,47 +829,63 @@ testRule({
...
 			line: 1,
 			column: 15,
+			endLine: 1,
+			endColumn: 20,
 		},
 		{
 			code: '.foo + .bar { & .baz + .foo { & foo + bar {} } }',
-			message: messages.expected('.foo + .bar .baz + .foo foo + bar', 0),
+			message: messages.expected('& foo + bar', 0),
 			line: 1,
```
ybiquitous, line 886: [suggest] Oh, `column` is missing 😓 

```suggestion
			line: 1,
			column: 31,
```

lib/rules/selector-max-type/__tests__/index.mjs
ybiquitous: Thanks a lot! The changes make sense!

ybiquitous (approved):
Great work! Thank you! 👍🏼

ybiquitous:
@Mouvedia Do you have any other thoughts?

lib/utils/__tests__/parseSelectorAST.test.mjs
```diff
@@ -0,0 +1,48 @@
...
+	it('parses selectors into an AST', async () => {
+		const result = await postcss().process('a .foo, b::before {}');
+		const rule = result.root.first;
+		const ast = parseSelectorAST(rule.selector, result, rule);
+
+		expect(ast).toBeTruthy();
+		expect(result.warnings()).toHaveLength(0);
+	});
+
+	it('short circuits on empty selectors', async () => {
```
Mouvedia, line 15: nit

This describes the internals.
i.e. I would expect a `toThrow`

lib/utils/__tests__/parseSelectorAST.test.mjs
```diff
@@ -0,0 +1,48 @@
...
+	});
+
+	it('warns on invalid selectors, but does not throw', async () => {
+		const result = await postcss().process('[&] {}');
+		const rule = result.root.first;
+		const ast = parseSelectorAST(rule.selector, result, rule);
+
+		expect(ast).toBeUndefined();
+		expect(
+			result.warnings().map((x) => {
```
Mouvedia, line 31: question

Is there more than 1 `node`?

lib/utils/__tests__/parseSelectorAST.test.mjs
```diff
@@ -0,0 +1,48 @@
...
+	it('short circuits on empty selectors', async () => {
+		const result = await postcss().process('{}');
+		const rule = result.root.first;
+		const ast = parseSelectorAST(rule.selector, result, rule);
+
+		expect(ast).toBeUndefined();
+		expect(result.warnings()).toHaveLength(0);
+	});
+
+	it('warns on invalid selectors, but does not throw', async () => {
```
Mouvedia, line 24: ditto

Mouvedia (approved):
> @Mouvedia Do you have any other thoughts?

Not much.
LGTM

lib/utils/__tests__/parseSelectorAST.test.mjs
romainmenke: No

lib/utils/__tests__/parseSelectorAST.test.mjs
romainmenke: I am unsure if your comment is about the message for this test, the test itself, or the implementation of `lib/utils/parseSelectorAST.mjs`.

Mouvedia: It's about the string passed as the first argument to the `it` function.

The _name_ should match the test and not the implementation.
i.e. in the case of a refactor—where the return type doesn't change—the test name shan't be changed
It's non-blocking, you can disregard it if you are disagreeing.

romainmenke: I've changed the first description to :

> returns `undefined` when there is no selector

lib/utils/__tests__/parseSelectorAST.test.mjs
romainmenke: This describes the public interface, not the implementation, so I am going to keep this as is.

romainmenke:
Thank you @ybiquitous, @Mouvedia for all the feedback and help 🙇

Issue #7396, opened 2023-12-11
Prepare 17.0.0


jeddy3:
This is an umbrella issue to capture some of the decisions we made while working on 16.0.0:

- [ ] Remove CommonJs Node.js API (and update our pure ESM deps)
- [ ] Remove deprecated `output` property in Node.js API result object

I'll create issues for these, but we'll likely want to wait at least a few months before touching them.

In the meantime, we can share any thoughts on what may be, if people have time, good things to look at in `16.x`.

It'd be amazing to see:

- more rules that help people write modern CSS consistently
- a built-in resolver so that people can lint in files in context

I think rules like [`color-format-notation`](https://stylelint.io/user-guide/rules/color-function-notation/) help our users write modern CSS code. I'm seeing code in the wild using modern CSS features like:

- `@container`
- `@layer`
- `clamp()` and other math comparison functions
- `has()`
- logical properties & values
- Nesting
- `oklch()`
- relative colours

For example:

```css
:root {
  --dk-color-black-20: oklch(20% 0 0deg);
  --dk-color-black-20-a-10: oklch(from var(--dk-color-black-20) l c h / 10%);
}

@layer defaults {
  :where(body) {
    color: var(--dk-color-black-20);
  }

  :where(a) {
    &:any-link {
      cursor: pointer;
    }

    &:hover {
      text-decoration: none;
    }
  }
}

parent-component {
  margin-inline: auto;

  &:has(child-component) {
    container: child-component / inline-size;
    display: grid;
  }
}

child-component {
  padding: clamp(1rem, 1rem + 1cqi, 2rem);

  @container parent-component (width > 40rem) {
    display: flex;
  }
}
```

There are already a [couple](https://github.com/stylelint/stylelint/issues/5430) of [open](https://github.com/stylelint/stylelint/issues/6421) issues around these, and I will try to make some time to revisit them. And I'll think about what new rules we can create to help people consistently use these modern features.

The built-in resolver already has an issue (https://github.com/stylelint/stylelint/issues/4088). It's a biggy, though, and I've added a "needs help" label because it's likely non-trivial. It'd be great to have as it'd open up linting CSS in the context of a whole CSS codebase. 

What are the things that you'd like? We can broadly discuss them here, or you can create `needs: discussion` issues for them.

jeddy3:
@ybiquitous Just saw your https://github.com/stylelint/stylelint/issues/7395 issue. What coincidently timing 😆

ybiquitous:
[...]

Issue #2498, opened 2017-04-13
Add smoke test


alexander-akait:
I think this would not be superfluous, and would have allowed to avoid stupid mistakes in releases.

Example smoke test https://github.com/babel/babel-preset-env/pull/240/commits/8101857b893c99834dc853400eca6b64d7cc05fe

jeddy3:
SGTM. I currently run a manual smoke test before each release, but having an automated one will help ensure that happens when any contributor does a release.

evilebottnawi:
@jeddy3 It will be fine if you show all the steps of the test, if the other wants to help us in this, nothing will miss

jeddy3:
>  It will be fine if you show all the steps of the test

Just before a release I simply install `master` into the `stylelint.io` directory, add a lint error to one of the CSS files, and run it via `stylelint.io`'s npm-run-scripts e.g.

```
cd stylelint.io
npm i ../stylelint
// add lint error, usually to `/src/components/Header/index.css`
npm run lint
```

The code in https://github.com/babel/babel-preset-env/pull/240/commits/8101857b893c99834dc853400eca6b64d7cc05fe seems to provide the same type of smoke test, without the manual work.

jeddy3:
Covered by our [ecosystem tester](https://github.com/stylelint/stylelint-ecosystem-tester).

PR #7532, opened 2024-02-20
Fix false positive CJS deprecation warning for dual-package plugins

### LINKED ISSUE #7460: Fix handling of plugins that are hybrid CommonJS and ESM packages

The stylelint plugin that I build with rollup outputs a `index.cjs` and `index.mjs` file.

In `package.json` file:

```
"main": "index.cjs",
"module": "index.mjs"
```

# Package structure

grafik

Stylelint 15 takes the `cjs` file and Stylelint 16 the `mjs` file

### What actually happened?

```
CommonJS plugins are deprecated ("/home/.../.../node_modules/@isnotdefined/stylelint-plugin/index.cjs")
```

Related issue: https://github.com/isnotdefinedcom/stylelint-plugin/issues/14

JounQin:

fix #7460
close #7535

~~I don't know how to keep `lib/utils/getModulePath.mjs` different with `lib/utils/getModulePath.cjs` because they should use different strategies.~~

This PR currently uses [`@dual-bundle/import-meta-resolve`](https://github.com/un-es/import-meta-resolve), when we work on next major release which should be ESM only, we should change to use original [`import-meta-resolve`](https://github.com/wooorm/import-meta-resolve) instead. This task should be added into #7396 if this PR is merged.


rollup.config.mjs
```diff
@@ -12,6 +12,12 @@ const inputFiles = fg.globSync([
 
 	// NOTE: We cannot support CommonJS for `cli.mjs` since the `meow` dependency is pure ESM.
 	'!lib/cli.mjs',
+	/**
+	 * NOTE: We are using pure ESM dependency `import-meta-resolve`
+	 *
+	 * @see https://github.com/stylelint/stylelint/pull/7532
+	 */
+	'!lib/utils/getModulePath.mjs',
```
JounQin, line 20: I'm not sure why it does not work expected.

cc @ybiquitous

ybiquitous:
@JounQin [`import-meta-resolve`](https://github.com/wooorm/import-meta-resolve) is a pure ESM package, right? If so, we cannot use it because a CommonJS module cannot load a pure ESM package.

ybiquitous:
Also, I'm not sure why this PR could resolve #7460...

JounQin:
@ybiquitous 

https://github.com/stylelint/stylelint/issues/7460#issuecomment-1953999411

I only enable `import-meta-resolve` for `.mjs` entry, for `.cjs` it still uses `resolve-from` that's why commonjs entry is resolved over ESM one previously.

This PR is not finally finished yet.

rollup.config.mjs
JounQin: @ybiquitous I'd expect this file wouldn't be transformed into `.cjs` but failed, so I have to reset that `.cjs` file after building.

ybiquitous:
> I only enable `import-meta-resolve` for `.mjs` entry, for `.cjs` it still uses `resolve-from` that's why commonjs entry is resolved over ESM one previously.

Hum, this solution is not so good because the building becomes complex. 😕

JounQin:
@ybiquitous I'd like to skip building that `.mjs` file just like `cli.mjs`.

Mouvedia:
Based on the name, [`resolveImportMeta`](https://rollupjs.org/plugin-development/#resolveimportmeta) might be useful.

JounQin:
> Based on their names, [`resolveImportMeta`](https://rollupjs.org/plugin-development/#resolveimportmeta) and [`import.meta.resolve()`](https://nodejs.org/docs/latest/api/esm.html#importmetaresolvespecifier) might be useful.

@Mouvedia I don't think that helps for our situation.

1. We need to resolve module path instead of only `import.meta.url`
2. `import.meta.resolve` is not supported for all node versions we're targeting, that's why I'm using https://github.com/wooorm/import-meta-resolve ponyfill.

If we don't want pure ESM dependency, we can bundle it for commonjs.

ybiquitous:
I feel like it's better to unify `getModulePath()` and `dynamicImport()` than this `resolve-from` and `import-meta-resolve` approach.

https://github.com/stylelint/stylelint/blob/c49f9b5a7b45cb35c58b5bc1cd669733896d0183/lib/utils/dynamicImport.mjs#L11

JounQin:
Indeed, it seems we don't need `getModulePath` at all, right?

ybiquitous:
> Indeed, it seems we don't need `getModulePath` at all, right?

Yes. I feel we can refactor `getModulePath()` and `dynamicImport()`. The dynamic `import()` can load both ESM and CJS packages or files, right?

ybiquitous:
It may be a quite big refactoring, though.

JounQin:
@ybiquitous I have several solutions:

1. Use different solution for `getModulePath.mjs` vs `getModulePath.cjs` like this PR
2. bundle `import-meta-resolve` into cjs with esbuild
3. use `import('import-meta-resolve')` + https://github.com/un-ts/synckit, no internal API change
4. change `getModulePath`'s API to async
5. drop `getModulePath` and refactor to use `dynamicImport`

4 and 5 could be a huge task, and we're very likely to drop commonjs support in the next major version, I don't know is it worth.

JounQin:
@ybiquitous Any news?

ybiquitous:
@JounQin Honestly, I don't want to use two build tools at the same time because of the dependency management and more complex build flow, which will no longer be necessary in the next major version.

ybiquitous:
We have already used rollup. In addition, PR #7535 tries adding esbuild. I doubt if the way is a suitable solution for this problem.

> 4 and 5 could be a huge task and may cause breaking change itself

I think non-breaking because `getModulePath` and `dynamicImport` are a private utility, but do I miss anything?

JounQin:
So what is your preferred way to fix that issue considering the contribution cost?

Notice: `getModulePath` is used in many places, so I think option 5 will not work. Option 4 will change many internal or some `stylelint.utils` API into async. (I haven't confirmed it)

ybiquitous:
I prefer a way not to increase dependencies and not to bring a more complex build flow as much as possible. They often would produce maintenance problems.

JounQin:
Did you mean you don't want `import-meta-resolve`? Then maybe `try/catch` + `dynamicImport`?

ybiquitous:
Although `import-meta-resolve` is well maintained, but it would increase an extra dependency and loading time, right? (because it includes many polyfill codes) If so, I do want to avoid adding dependencies.

To minimize potential side effects, I think we can fix the following code block, which loads a plugin function from a string:

https://github.com/stylelint/stylelint/blob/c49f9b5a7b45cb35c58b5bc1cd669733896d0183/lib/augmentConfig.mjs#L320-L329

At the same time, we have to remove the following code:

https://github.com/stylelint/stylelint/blob/c49f9b5a7b45cb35c58b5bc1cd669733896d0183/lib/augmentConfig.mjs#L150-L158

I imagine the updated code would be like this:

```diff
if (typeof pluginLookup === 'string') {
-	pluginImport = await dynamicImport(pluginLookup);
+	pluginImport = await importPlugin(pluginLookup);
}

// ...

async function importPlugin(lookup) {
	// Using dynamicImport()...

	// If dynamicImport() fails, try reading a file...
}
```

JounQin:
@ybiquitous

But I don't think we can support global plugins without `import-meta-resolve`? `import()` only support local dependence by default, we need to resolve global plugins manually first. And `resolve-global` doesn't support ESM only package well.

And also, we have many dependencies already, `resolve-from` is also as slow `import-meta-resolve`/native `import.meta.resolve`, I still don't get the point. Correctness is more important than performance trade off IMO.

ybiquitous:
Just to confirm, does [`import.meta.resolve()`](https://nodejs.org/api/esm.html#importmetaresolvespecifier) in Node.js support importing global modules?

JounQin:
I don't think so, and we have to use something like ［`global-directory`］(https://github.com/sindresorhus/global-directory) which `resolve-global` is also using.

See also https://github.com/conventional-changelog/commitlint/pull/3939 which I just fixed for @commitlint yesterday.

ybiquitous:
I see. Thanks for the info. I'm still concerned about using `import.meta.resolve()` because:

- [`require.resolve()`](https://nodejs.org/api/modules.html#requireresolverequest-options) can also resolve ESM modules
- `require.resolve()` supports loading [global modules](https://nodejs.org/api/modules.html#loading-from-the-global-folders) via `NODE_PATH` etc.
- [`import.meta.resolve()`](https://nodejs.org/api/esm.html#importmetaresolvespecifier) experimentally supports the non-standard second argument `parent`

If Stylelint continues supporting global modules since the next major version (17.0.0), does it make sense to depend on the non-standard API of `import.meta.resolve()`?

ybiquitous:
> I noticed that you were concerning about using native `import.meta.resolve`, I still recommend to use the `import-meta-resolve` ponyfill, no above issues need to be concerned about.

But we will replace the ponyfill with the native API some day, right?

JounQin:
> But we will replace the ponyfill with the native API some day, right?

Sure! When it's available without flag and supports all engines we're targeting.

ybiquitous:
I try summarizing points:

- Stylelint now loads CJS plugins rather than ESM ones in a dual package (want this PR tries to resolve)
- The `import-meta-resolve` ponyfill is necessary because `import.meta.resolve()` is experimental yet
- `resolve-from` is still necessary because `import-meta-resolve` is a pure ESM package
- `import.meta.resolve()` doesn't support global modules
- The `parent` second argument of `import.meta.resolve()` in Node.js is non-standard (this may be problematic on other platforms)

Currently, I think we should wait for `import.meta.resolve()` (`import-meta-resolve`) until the next major version (`17.0.0`) because Stylelint will switch to a pure ESM package in the version.

For global modules, which is another topic, I'd like to stop the support (only local modules supported) honestly, but it may be challenging for compatibility reasons.

JounQin:
> Currently, I think we should wait for `import.meta.resolve()` (`import-meta-resolve`) until the next major version (17.0.0) because Stylelint will switch to a pure ESM package in the version.

So #7460 would be marked as a known issue until the next major version, right? I don't understand why we cannot fix it in current major version, but I respect your decision.

lib/augmentConfig.cjs
```diff
@@ -2,7 +2,7 @@
 // please instead edit the ESM counterpart and rebuild with Rollup (npm run build).
 'use strict';
 
-const node_path = require('node:path');
+const path = require('node:path');
```
Mouvedia, line 5: minor

Even if I agree with the rename, I think you should reduce the diff if possible.
i.e. minor refactors should be avoided if you don't change anything in a file apart from the rename

lib/augmentConfig.cjs
```diff
@@ -2,7 +2,7 @@
 // please instead edit the ESM counterpart and rebuild with Rollup (npm run build).
 'use strict';
 
-const node_path = require('node:path');
```
Mouvedia, line 5: minor

Even if I agree with the rename, I think you should reduce the diff if possible.
i.e. minor refactors should be avoided if you don't change anything in a file apart from the rename

JounQin: It is automatically changed by rollup.

ybiquitous:
> So #7460 would be marked as a known issue until the next major version, right?

In my understanding, the downside of a dual package is only not to show the deprecation warning about using the CJS version. Showing the warning is ideal, but a dual package works still until the next major version (17.0.0). I really worry that this PR solution may cause other problems.

> What if `import-meta-resolve` is commonjs compatible?

It's worth considering. But compiling and bundling `import-meta-resolve` is too complex for this project. And it may cause other problems.

> Besides, do we support yarn P'n'P officially?

No, we don't. Honestly, I don't know the problem of yarn P'n'P, but I think the problem should not stop our move toward pure ESM.

JounQin:
> It's worth considering. But compiling and bundling import-meta-resolve is too complex for this project. And it may cause other problems.

OK, how about that, like https://github.com/es-tooling/cjs-bundle/issues/1, I'd like to fork `import-meta-resolve` and republish it as `@dual-bundle/import-meta-resolve` which would be a dual package in the meantime. And like https://github.com/un-es/eslint-plugin-i:

> It will rebase and try to release in order to sync with the upstream every day, see [.github/workflows/rebase-upstream.yml](https://github.com/un-es/eslint-plugin-i/blob/fork-release/.github/workflows/rebase-upstream.yml) for details.

Mouvedia:
> > Besides, do we support yarn P'n'P officially?
> 
> No, we don't. […]

Wait up. If there's an incompatibility with yarn PNP it either needs to be discussed in the issue or have its own separate issue—i.e. not in this PR. I disagree: we should support pnpm, npm and yarn pnp to the best of our abilities.

JounQin:
For yarn P'n'P compatibility, we may still need `resolve-from` as fallback although it may resolve commonjs incorrectly even in next major version like my another PR https://github.com/conventional-changelog/commitlint/pull/3941.

JounQin:
@ybiquitous I've changed to use [`@dual-bundle/import-meta-resolve`](https://github.com/un-es/import-meta-resolve), no extra build step is needed now.

lib/utils/resolveSilent.mjs
```diff
@@ -0,0 +1,46 @@
+import fs from 'node:fs';
+import path from 'node:path';
+
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+import { moduleResolve } from '@dual-bundle/import-meta-resolve';
+import resolveFrom from 'resolve-from';
+
+const pathSuffixes = ['', '.js', '.json', `${path.sep}index.js`, `${path.sep}index.json`];
```
Mouvedia, line 9: minor
question

We don't need to support `.cjs` nor `.mjs` I suppose?

lib/utils/resolveSilent.mjs
```diff
@@ -0,0 +1,46 @@
+import fs from 'node:fs';
+import path from 'node:path';
```
Mouvedia, line 2: minor

Maybe `{ isAbsolute, resolve }` instead?

.changeset/clean-needles-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: false positive CJS deprecation warning for dual plugin packages
```
Mouvedia, line 5: nit

there's a word missing after "dual"

lib/utils/resolveSilent.mjs
JounQin: Those two suffixes must be used specifically, and these suffixes are only here for backward compatibility.

.changeset/clean-needles-deliver.md
JounQin: Oh sorry, I'm not a native English speaker, can you point out what word should I add or how should I change the sentence?

Mouvedia: I am not really sure, maybe `…for hybrid plugin packages` would be better?
Or perhaps `…for dual-package plugins`?
Either one sounds better to me.

JounQin:
```suggestion
Fixed: false positive CJS deprecation warning for dual-package plugins
```

lib/utils/resolveSilent.mjs
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * @param {string} parent
+ * @param {string} lookup
+ * @return {string | undefined}
+ */
+export default function resolveSilent(parent, lookup) {
+	if (path.isAbsolute(lookup)) {
+		for (const suffix of pathSuffixes) {
+			const filename = lookup + suffix;
```
ybiquitous, line 21: [ask] I don't understand the necessity of the suffixes. Can you explain?

lib/utils/resolveSilent.cjs
```diff
@@ -0,0 +1,50 @@
...
+
+			if (fs.existsSync(filename)) {
+				return filename;
+			}
+		}
+
+		return;
+	}
+
+	const resolvedParent = node_url.pathToFileURL(path.resolve(parent, 'noop.js'));
```
ybiquitous, line 33: [ask] Why is `noop.js` necessary?

lib/utils/resolveSilent.cjs
```diff
@@ -0,0 +1,50 @@
...
+		} catch {
+			//
+		}
+	}
+
+	/**
+	 * Yarn P'n'P does not support pure ESM well by default, this is only a workaround for it
+	 * @see https://github.com/wooorm/import-meta-resolve/issues/23
+	 */
+	return resolveFrom.silent(parent, lookup);
```
ybiquitous, line 47: [ask] This workaround will no longer be needed when Stylelint becomes pure ESM, right?

ybiquitous (commented):
Thanks for the new package. I just asked a few questions about the implementation.

lib/utils/resolveSilent.cjs
JounQin: To mock `import.meta.url` which is a filename.

lib/utils/resolveSilent.mjs
JounQin: In pure ESM, those anonymous suffixes will not be tried automatically anymore while for commonjs (`resolve-from`) will try them automatically, so for compatibility we need to try all theses files inside.

lib/utils/resolveSilent.cjs
JounQin: Nope. `import-meta-resolve` does not support such options.

https://github.com/wooorm/import-meta-resolve#differences-to-node

image

lib/utils/resolveSilent.cjs
```diff
@@ -0,0 +1,50 @@
...
+			return node_url.fileURLToPath(importMetaResolve.moduleResolve(lookup + suffix, resolvedParent));
+		} catch {
+			//
+		}
+	}
+
+	/**
+	 * Yarn P'n'P does not support pure ESM well by default, this is only a workaround for it
+	 * @see https://github.com/wooorm/import-meta-resolve/issues/23
+	 */
```
JounQin, line 46:
```suggestion
	/**
	 * Yarn P'n'P does not support pure ESM well by default, this is only a workaround for it
	 * @see https://github.com/wooorm/import-meta-resolve/issues/23
	 *
	 * TODO: this workaround is still necessary util native `import.meta.resolve` available
	 */
```

lib/utils/resolveSilent.cjs
JounQin: Sure! How about `TODO: this workaround is still necessary util native `import.meta.resolve` available`

ybiquitous: Sounds good 👍🏼

lib/utils/resolveSilent.mjs
ybiquitous: > I doubt will we remove the compatibility in the future

Why? In pure ESM, supporting these implicit suffixes (e.g. `index.js`) will not make sense, right? Because `import` requires an accurate path unlike `require()`?

lib/utils/resolveSilent.cjs
ybiquitous: Is the mock necessary? Is a parent directory not working? In my PoC below, the parent directory seems to work: 🤔 

`tmp/a.js`:

```js
import { pathToFileURL } from 'node:url';

import { resolve } from '@dual-bundle/import-meta-resolve';

const parent = pathToFileURL(process.env.NODE_PATH);

console.log('parent:', parent.href);
console.log('import.meta.resolve:', import.meta.resolve('npm', parent));
console.log('import-meta-resolve:', resolve('npm', parent));
```

Run:

```sh-session
$ NODE_PATH=$(npm root -g) node --experimental-import-meta-resolve tmp/a.mjs
parent: file:///Users/masafumi.koba/.nvm/versions/node/v21.6.1/lib/node_modules
import.meta.resolve: file:///Users/masafumi.koba/.nvm/versions/node/v21.6.1/lib/node_modules/npm/index.js
import-meta-resolve: file:///Users/masafumi.koba/.nvm/versions/node/v21.6.1/lib/node_modules/npm/index.js

$ node -v
v21.6.1
```

lib/utils/resolveSilent.mjs
```diff
@@ -0,0 +1,46 @@
+import fs from 'node:fs';
+import path from 'node:path';
+
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+import { moduleResolve } from '@dual-bundle/import-meta-resolve';
```
ybiquitous, line 6: [ask] Why does it use `moduleResolve` instead of [`resolve`](https://github.com/wooorm/import-meta-resolve#resolvespecifier-parent)? `resolve` seems closer to the native `import.meta.resolve`.

JounQin: `resolve` does not apply [Resolver algorithm](https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_resolver_algorithm), so `moduleResolve` is slower than `resolve`.

lib/utils/resolveSilent.mjs
JounQin: That's good if we don't need such compatibility. I'll add such comment.

```
TODO: These suffixes are here for commonjs compatibility reason, we should remove these codes after migrating to pure ESM because in ESM accurate paths are required
```

lib/utils/resolveSilent.cjs
Mouvedia: typo `util` > `until`

```md
…until native `import.meta.resolve` availability
```
or
```md
…before native `import.meta.resolve` replacement
```

lib/utils/resolveSilent.mjs
ybiquitous: Sounds good 👍🏼

lib/utils/resolveSilent.cjs
ybiquitous: Make sense. Let's keep it as-is. 👍🏼

lib/utils/resolveSilent.mjs
ybiquitous: I see. Thanks for the link 👍🏼

lib/utils/resolveSilent.mjs
```diff
@@ -0,0 +1,53 @@
...
+		} catch {
+			//
+		}
+	}
+
+	/**
+	 * Yarn P'n'P does not support pure ESM well by default, this is only a workaround for it
+	 * @see https://github.com/wooorm/import-meta-resolve/issues/23
+	 *
+	 * TODO: this workaround is still necessary before native `import.meta.resolve` replacement
```
ybiquitous, line 50: [ask] Thanks for the TODO comment. I'm reading the Yarn doc (https://yarnpkg.com/getting-started/qa#does-yarn-support-esm) from your shared issue. So let me ask.

The Yarn doc says:

> Loaders aren't perfect yet, and until they are ESM-only packages cannot be recommended, but there's a way forward and as we work together we'll get there. We just have to be careful not to push people aside as we run towards our goal.

Does this mean that Yarn will resolve this ESM issue in the future, i.e. will this workaround no longer be necessary? If so, is there any GitHub issue we should track?

JounQin: When we migrate to native `import.meta.resolve`, yarn P'n'P will be supported as well.

The issue here is that `import-meta-resolve` does not support `--loader`.

ybiquitous: I see. Thanks for the answer 👍🏼

ybiquitous (dismissed):
Thanks for your hard work! LGTM 👍🏼 

Hopefully, we'd like one more person's approval.

Mouvedia:
> I believe @Mouvedia can help to review.

I would prefer if @jeddy3 did.
The reason being that a test would be too tricky and this is a critical PR.

related: #2498

JounQin:
cc @jeddy3

JounQin:
Now that @jeddyjeis missing for reviewing, @Mouvedia can you help to review for not blocking?

Mouvedia:
@JounQin I am using https://github.com/stylelint/stylelint/issues/7460#issuecomment-1955866907 to check whether it's fixed or not.

I get

```sh
~/[…]/stylelint/test % echo 'a{}' | ../bin/stylelint.mjs
CommonJS plugins are deprecated ("/Users/ggozlan/Code/stylelint/test/node_modules/@isnotdefined/stylelint-plugin/index.cjs"). See https://stylelint.io/migration-guide/to-16
```

If the steps provided by @ybiquitous are incorrect could you provide a MRE or a link to one ?

JounQin:
image

Interesting, I'll report to https://github.com/wooorm/import-meta-resolve.

Mouvedia:
> Interesting, I'll report to [wooorm/import-meta-resolve](https://github.com/wooorm/import-meta-resolve).

I suppose that means it's not ready to be approved yet.

JounQin:
@Mouvedia fixed in [`bebcc68` (#7532)](https://github.com/stylelint/stylelint/pull/7532/commits/bebcc68614f2afcd7d75362e44782375cbf9193d)

ybiquitous:
Just curious, why does our test suite cause `ERR_PACKAGE_PATH_NOT_EXPORTED`?

JounQin:
> Just curious, why does our test suite cause `ERR_PACKAGE_PATH_NOT_EXPORTED`?

`moduleResolve` does not have default `conditions` param, so `exports` will not be resolved correctly.

It is mentioned at https://github.com/wooorm/import-meta-resolve/issues/24#issuecomment-1980982765

ybiquitous:
I understand the commit resolved the error, but I want to know why our test suite didn't detect the error before the commit.

JounQin:
> I understand the commit resolved the error, but I want to know why our test suite didn't detect the error before the commit.

Because the error is silent and we have `resolve-from` as fallback which resolves `cjs`.

ybiquitous:
I see. You mean discarding a caught error here, right?

https://github.com/stylelint/stylelint/blob/bebcc68614f2afcd7d75362e44782375cbf9193d/lib/utils/resolveSilent.mjs#L41-L45

But since we encountered the error to be resolved, should we handle unexpected errors like this?

```js
try {
	return fileURLToPath(moduleResolve(lookup + suffix, base, conditions));
} catch (e) {
	// Are there other errors than ERR_PACKAGE_PATH_NOT_EXPORTED?
	if (e.code === 'ERR_PACKAGE_PATH_NOT_EXPORTED') {
		throw e;
	}

	// Ignore other errors
}
```

```

**Comment by JounQin** [comment]
> But since we encountered the error to be resolved, should we handle unexpected errors like this?

But we also have https://github.com/stylelint/stylelint/pull/7532/files#diff-f432f021d32e88d2e0cd52d76e2887dbd527250f4beba1021a0df8eca15e230eR54 for Yarn P'n'P?

And also, we don't expected `resolveSilent` to throw anyway? See also `getModulePath`, which will fallback to `cwd` then `global`.

And after adding `conditions` correctly, I don't think we should care about `ERR_PACKAGE_PATH_NOT_EXPORTED` any longer.

**Comment by ybiquitous** [comment]
Um..., I still worry about ignoring other critical errors, not only `ERR_PACKAGE_PATH_NOT_EXPORTED`. If a critical error occurs, we cannot notice it, right?

**Comment by ybiquitous** [comment]
Although the silently fallback can avoid the worst case... 🤔

**Comment by ybiquitous** [comment]
By the way, I'm reading https://github.com/wooorm/import-meta-resolve/issues/24#issuecomment-1982970965 and am worried about the comment:

> 99% of users should not use `moduleResolve`.

**Comment by JounQin** [comment]
Here is my response: https://github.com/wooorm/import-meta-resolve/issues/24#issuecomment-1983068175

I'm using `moduleResolve` because it doesn't work. It catches some errors which could be unexpected for some cases: https://github.com/wooorm/import-meta-resolve/blob/411516aaefa86bbe4d8566a98e79370f78f48338/index.js#L37-L43 and the logic is a bit different: https://github.com/wooorm/import-meta-resolve/blob/411516aaefa86bbe4d8566a98e79370f78f48338/lib/resolve.js#L1266

`moduleResolve` is lower-level than `resolve`, which works fine in our test cases.

**Comment by ybiquitous** [comment]
I can fix the test failures by using `resolve` instead of `moduleResolve`. Is this correct?

```js
for (const suffix of specifierSuffixes) {
	const resolved = fileURLToPath(resolve(lookup + suffix, base));

	if (fs.existsSync(resolved)) {
		return resolved;
	}
}
```

JounQin:
@ybiquitous It doesn't catch possible error thrown from `resolve`, and then that could just be same as `moduleResolve`, I don't get why should we change to use `resolve` over lower-level `moduleResolve`.

Mouvedia:
> If a critical error occurs, we cannot notice it, right?

@JounQin if we add

```js
	if (e.message) {
		console.error(e.message);
	}
```

in the `catch`, would it be too noisy for the Yarn PnP fallback?

ybiquitous:
I think this `fs.existsSync` check makes sense because `import.meta.resolve` returns a URL even if the given path doesn't exist. See the example below. Although `./lib/foo.js` doesn't exist, the resolved URL is printed:

```sh-session
$ cat a.mjs
console.log(import.meta.resolve('./lib/foo.js'));

$ node a.mjs
file:///Users/masafumi.koba/git/stylelint/stylelint/lib/foo.js

$ ls lib/foo.js
ls: lib/foo.js: No such file or directory

$ node -v
v21.6.2
```

So, if `import-meta-resolve` raises an error, it should be a critical case that we shouldn't catch. For example, `import-meta-resolve` doesn't raise an error of `ERR_MODULE_NOT_FOUND`.

In addition, I think using `resolve` instead of the low-level `moduleResolve` makes sense because:

- `moduleResolve` is discouraged by the `import-meta-resolve` author on <https://github.com/wooorm/import-meta-resolve/issues/24#issuecomment-1982970965>.
- We will switch from `import-meta-resolve` (ponyfill) to `import.meta.resolve` someday. Then, `resolve` can be migrated more easily than `moduleResolve`.

Mouvedia:
> @Mouvedia fixed in [`bebcc68` (#7532)](https://github.com/stylelint/stylelint/pull/7532/commits/bebcc68614f2afcd7d75362e44782375cbf9193d)

This is indeed resolved.

Mouvedia (approved):
> @Mouvedia fixed in [`bebcc68` (#7532)](https://github.com/stylelint/stylelint/pull/7532/commits/bebcc68614f2afcd7d75362e44782375cbf9193d)

This is indeed resolved.
@ybiquitous is this ready to approved? Or the switch from `moduleResolve` to `resolve` a blocker?

JounQin:
> We will switch from import-meta-resolve (ponyfill) to import.meta.resolve someday. Then, resolve can be migrated more easily than moduleResolve.

This seems making sense to me. 🤔

@Mouvedia How do you think? And at the same time we need to catch `resolve`'s error silently for Yarn P'n'P or custom resolver actually.

ybiquitous:
> And at the same time we need to catch `resolve`'s error silently for Yarn P'n'P or custom resolver actually.

When using `resolve` with Yarn P'n'P, what error is raised?

Mouvedia:
> @Mouvedia How do you think? [sic]

Find better arguments to convince @ybiquitous :)
For me the bug seems fixed, hence the rest is minor and can be resolved in a separate `refactor` PR.  
Now if there are cases not covered by @ybiquitous's MRE I would like either to know them or for you to test them.

Being cautious was the right call; we can release a new version once the question on whether the refactor is necessary or not is settled.

JounQin:
> When using resolve with Yarn P'n'P, what error is raised?

@ybiquitous 

https://github.com/wooorm/import-meta-resolve/issues/10#issuecomment-1968831154

ybiquitous:
`ERR_MODULE_NOT_FOUND` is raised, right? If so, it seems that we can write like this:

```js
try {
  // resolve(...)
} catch (e) {
  if (e.code === 'ERR_MODULE_NOT_FOUND') {
    // Ignore the error for Yarn P'n'P
  } else {
    throw e;
  }
}
```

ybiquitous:
I was also wondering if we could add a test case to reproduce this issue. 🤔 

E.g.

```jsonc
// packages/dual-package/package.json
{
  "name": "@stylelint/dual-package",
  "version": "1.0.0",
  "exports": {
    "require": "./index.cjs",
    "import": "./index.mjs"
  }
}
```

```jsonc
// package.json
{
  "devDependencies": {
    "@stylelint/dual-package": "file:packages/dual-package"
  }
}
```

JounQin:
I don't quite get why we need to care about the error in a **silent** resolving function, but if you stick to it, I'm fine with that. However, as I understand, there could be still other undefinded/unknown errors could be thrown or even in tbe further.

ybiquitous:
> However, as I understand, there could be still other undefinded/unknown errors could be thrown or even in tbe further.

Okay, let's keep it as-is.

> > I was also wondering if we could add a test case to reproduce this issue. 🤔
> I can give it a try if you want.

Yes, please.

JounQin:
@ybiquitous I just added the test case at [`a836e90` (#7532)](https://github.com/stylelint/stylelint/pull/7532/commits/a836e90888a10d71dbf706f2c5e562c2a776b064)

lib/__tests__/resolveSilent.test.mjs
```diff
@@ -0,0 +1,13 @@
+import path from 'path';
+
+import { fileURLToPath } from 'url';
```
ybiquitous, line 3: [suggest] Minor. I prefer `node:`, so can you change it?

```suggestion
import path from 'node:path';

import { fileURLToPath } from 'node:url';
```

lib/__tests__/resolveSilent.test.mjs
```diff
@@ -0,0 +1,13 @@
...
+
+import resolveSilent from '../utils/resolveSilent.mjs';
+
+const fixturesPath = path.resolve(fileURLToPath(import.meta.url), '../fixtures');
```
ybiquitous, line 7: [suggestion] Minor. One library is required here:

```suggestion
const fixturesPath = fileURLToPath(new URL('./fixtures', import.meta.url));
```

ybiquitous (commented):
@JounQin Thanks for adding the test case. Putting `node_modules` in the fixtures directory is a nice idea!

ybiquitous:
@JounQin I still have a concern about using `moduleResolve` instead of `resolve` as I commented on <https://github.com/stylelint/stylelint/pull/7532#issuecomment-1983397322>. Don't your idea change?

JounQin:
> @JounQin I still have a concern about using `moduleResolve` instead of `resolve` as I commented on [#7532 (comment)](https://github.com/stylelint/stylelint/pull/7532#issuecomment-1983397322). Don't your idea change?

> Okay, let's keep it as-is.

I was thinking we don't need change at this PR.

I'll change to use `resolve` then.

ybiquitous:
>> Okay, let's keep it as-is.

> I was thinking we don't need change at this PR.

Sorry, this comment meant "try...catch", not `moduleResolve`. 🙏🏼

lib/__tests__/resolveSilent.test.mjs
ybiquitous, line 1: [suggset] Thanks for the quick fix. But I just realized that `lib/__tests__/resolveSilent.test.mjs` should be in `lib/utils/__tests__/`. Can you move it and its related fixtures?

JounQin: Done.

.gitignore
```diff
@@ -1,4 +1,6 @@
 node_modules
+!lib/__tests__/fixtures/node_modules
```
Mouvedia, line 2: medium

remove this line

JounQin: There is https://github.com/stylelint/stylelint/blob/main/lib/__tests__/fixtures/node_modules/test.css actually.

Mouvedia: So you are saying it was an oversight of be558d4.

https://github.com/stylelint/stylelint/blob/d2c8d0dd20924706c3075c9d208234c373a41f42/lib/__tests__/ignore.test.mjs#L386

ybiquitous (approved):
Thank you for your hard work! LGTM 👍🏼

PR #7535, opened 2024-02-22
fix: replace `resolve-from` with `import-meta-resolve` for dual package

### LINKED ISSUE #7460: Fix handling of plugins that are hybrid CommonJS and ESM packages

JounQin:

Fixes #7460

This is another solution which close #7532, it bundles `import-meta-resolve` as vendor so that it can be used directly in commonjs


ybiquitous:
Blocked by the discussion in #7532.

PR #7539, opened 2024-02-25
Add tally of fixable problems to `string` and `verbose` formatters

### LINKED ISSUE #7338: Add tally of fixable problems to string formatter

m-allanson:

Closes #7338 

This PR adds a count of potentially fixable problems to the string formatter's output. e.g.

```
file.css
  1:2  ×  Unexpected foo  no-foo
  1:2  ×  Unexpected bar  no-bar
  1:2  ‼  Unexpected baz  no-baz

× 3 problems (2 errors, 1 warning)
  1 error and 1 warning potentially fixable with the "--fix" option.
```

The line `1 error and 1 warning potentially fixable with the "--fix" option.` is new.

In adddition, I've adjusted the formatting slighty, to ensure that the most important line (the problem count) is still visually distinct. This new formatting imitates eslint's default output.

The formatting changes are:

- add an `×` in front of the problem count, to highlight that line
- indent the list of problems, so that the line numbers are vertically aligned with the problem count and fix count text.

<details><summary>Formatting examples</summary>

Here are examples of the formatting changes, excluding the new fixable counts.

Old formatting:

```
file.css
 1:2  ×  Unexpected foo  no-foo
 1:2  ×  Unexpected bar  no-bar
 1:2  ‼  Unexpected baz  no-baz

3 problems (2 errors, 1 warning)
```

New formatting:

```
file.css
  1:2  ×  Unexpected foo  no-foo
  1:2  ×  Unexpected bar  no-bar
  1:2  ‼  Unexpected baz  no-baz

× 3 problems (2 errors, 1 warning)
```

</details>


ybiquitous:
@m-allanson Thanks for the pull request.

> * add an `×` in front of the problem count, to highlight that line

That sounds good to me. But I am concerned about a case of 0 errors and some warnings, e.g.

```
file.css
 1:2  ‼  Unexpected baz  no-baz

× 1 problem (0 errors, 1 warning)
```

In this case, using `‼` instead of `×` seems a bit more consistent and readable for me. What do you think?

.changeset/selfish-points-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: Show a count of fixable problems in the string formatter
```
ybiquitous, line 5: [suggest] Can you change the description for wording consistency in CHANGELOG.md?

```suggestion
Added: tally of fixable problems to `string` and `verbose` formatters
```

lib/formatters/__tests__/verboseFormatter.test.mjs
```diff
@@ -368,20 +368,19 @@ describe('verboseFormatter', () => {
...
+			  1:2  ×  Unexpected foo  no-foo
+			  1:2  ×  Unexpected bar  no-bar
+			  1:2  ×  Unexpected baz  no-baz
 
-			3 problems (3 errors, 0 warnings)
+			× 3 problems (3 errors, 0 warnings)
+			  2 errors potentially fixable with the "--fix" option.
 
 			1 source checked
 			 file.css
```
ybiquitous, line 379: [suggest] The same indent level also in the file list seems a bit more readable. Any thoughts?

```suggestion
			  file.css
```

m-allanson: Yes good idea. Updated 👍

lib/formatters/stringFormatter.mjs
```diff
@@ -249,7 +251,12 @@ export default function stringFormatter(results, returnValue) {
 		);
 
 		for (const warning of result.warnings) {
-			calcSeverityCounts(warning.severity, counts);
+			calcSeverityCounts(warning.severity, resultCounts);
+			calcMaybeFixableCounts(
+				warning.severity,
+				returnValue.ruleMetadata?.[warning.rule]?.fixable ?? false,
+				fixableCounts,
+			);
```
ybiquitous, line 259: [suggest] How about reusing `calcSeverityCounts` like this? If so, `calcMaybeFixableCounts` seems unneeded.

```suggestion
			const fixable = returnValue.ruleMetadata?.[warning.rule]?.fixable;

			if (fixable === true) {
				calcSeverityCounts(warning.severity, fixableCounts);
			}
```

m-allanson:
Thanks for the review @ybiquitous 👍 This is updated and ready for another look.

ybiquitous (commented):
Thank you. This PR is almost okay. But I just commented on a refactoring, so can you take a look?

lib/formatters/stringFormatter.mjs
m-allanson: That's much better. Updated!

ybiquitous (approved):
Thanks! LGTM 👍🏼

lib/formatters/__tests__/stringFormatter.test.mjs
```diff
@@ -158,10 +443,10 @@ path/to/file.css
 
 		expect(output).toBe(stripIndent`
 path/to/file.css
- 1:1  ×  Unexpected very very very very very very very  bar-very-very-very-very-very-long
-         very very very very very very long foo
+  1:1  ×  Unexpected very very very very very very very  bar-very-very-very-very-very-long
```
Mouvedia, line 446: minor nit

I would have expected this one to have required one character to be removed to pass.

m-allanson: Huh, yeah. I dug into this and it's an off-by-one error in the existing implementation. On `main`, this line is only 89 characters long. Adding the one extra space in this branch pushes it up to 90.

The cause of the error is the hardcoded `MARGIN_WIDTHS` value: https://github.com/stylelint/stylelint/blob/c49f9b5a7b45cb35c58b5bc1cd669733896d0183/lib/formatters/stringFormatter.mjs#L17

By default, `table` adds one character of padding to each side of each column. So a five column table would have 10 characters of padding. This can be overridden in the table config by setting paddingRight and paddingLeft. For `main`, that's done like this: https://github.com/stylelint/stylelint/blob/c49f9b5a7b45cb35c58b5bc1cd669733896d0183/lib/formatters/stringFormatter.mjs#L213-L221

So this table only had 8 characters of padding, but `MARGIN_WIDTHS` was set to 9. Which resulted in the table calculations coming out one character short.

This PR modifies the table config, bringing the actual padding value up to 9:  https://github.com/stylelint/stylelint/blob/ef5133519b8fe989ce92e678b6c683eefa122bfa/lib/formatters/stringFormatter.mjs#L213-L221

I've added a test to ensure the formatter will use up to the maximum available space.

A real fix would be to derive the margin widths from the table config itself, instead of using the `MARGIN_WIDTHS` constant. But I think that's outside the scope of this PR.

Mouvedia: > it's an off-by-one error

So my hunch was correct.

> I've added a test to ensure the formatter will use up to the maximum available space.

You have gone above and beyond of what I was expecting.

Mouvedia (approved):
The tests look good to me.

PR #7543, opened 2024-02-26
Refactor Cosmiconfig types

ybiquitous:

None.

Allowing the `CosmiconfigResult` type to receive a type parameter can make the code a bit type-safer because the `any` type is no longer used.
Ref https://github.com/cosmiconfig/cosmiconfig/blob/v9.0.0/src/types.ts#L11

Note this change includes a patch for the type defined in the `cosmiconfig` package. I will open a pull request in the package repository soon.


ybiquitous:
I just transitioned this PR to ready for review because this patch benefits. I would be happy if someone could take a look. 🙏🏼

ybiquitous:
There is no review yet, but I believe this helps detect issues like 9c7bbe3. So I'll merge this.

PR #7544, opened 2024-02-26
Add `ignoreSelectors: []` to `selector-max-compound-selectors`

### LINKED ISSUE #7536: Add `ignoreSelectors: []` to `selector-max-compound-selectors`

Vue 2[^vue3] uses a special syntax to allow component-scoped styles to affect child components:

[^vue3]: In Vue 3, the syntax was changed to `.foo :deep(.bar)`, see <https://vuejs.org/api/sfc-css-features.html#deep-selectors>, and the last "pass" example.

```css
/* all equivalent: */
.foo >>> .bar {}
.foo::v-deep .bar {}
.foo ::v-deep .bar {}
.foo /deep/ .bar {}
```

See https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors.

When using Sass, `>>>` and `/deep/` are not supported correctly, so we are stuck with using `::v-deep`. However, that "fake" selector counts towards the [`selector-max-compound-selectors`](https://stylelint.io/user-guide/rules/selector-max-compound-selectors/) limit.

Since that "fake" selector rather acts as a combinator, I'd like to ignore it in the `selector-max-compound-selectors` rule, while keeping the current limit:

#### Rule

```json
"selector-max-compound-selectors": [3, { "ignore": ["::v-deep"] }]
```

#### Fail

```css
.foo .bar .baz ::v-deep .lorem {}

.foo .baz {
  & ::v-deep > .bar .lorem {}
}
```

#### Pass

```css
#foo ::v-deep #bar > #baz {}

.foo .baz {
  & ::v-deep > .bar {}
}

.foo + div :deep(a b ~ c) {} /* `a b ~ c` is inside `:deep()`, so it is evaluated separately */
```

FloEdelmann:

Closes #7536.

[I don't think ignoring `:not` should also ignore contents of it.](https://github.com/stylelint/stylelint/pull/7544#discussion_r1510802573)


lib/rules/selector-max-compound-selectors/__tests__/index.mjs
```diff
@@ -384,3 +384,82 @@ testRule({
 		},
 	],
 });
+
+// Testing ignoreSelectors
+testRule({
+	ruleName,
+	config: [2, { ignoreSelectors: ['::v-deep', '/-moz-.*/', /-screen$/] }],
+
+	accept: [
```
Mouvedia, line 393: _if the status is changed to ready to implement_

add a nesting example
i.e. would `:global(.foo)` count as 1 or 0?
If it's 1, then similar selectors like `:not` will require _unpacking_ if ignored

lib/rules/selector-max-compound-selectors/index.mjs
```diff
@@ -51,7 +65,11 @@ const rule = (primary) => {
 				// Compound selectors are separated by combinators, so increase count when meeting one
 				if (childNode.type === 'combinator') {
 					compoundCount++;
+				} else if (optionsMatches(secondaryOptions, 'ignoreSelectors', childNode.toString())) {
+					compoundCount -= previousNodeWasCompoundSeparator ? 1 : 0;
```
Mouvedia, line 69: stylistic

```suggestion
					compoundCount -= Number(previousNodeWasCompoundSeparator);
```

.changeset/big-trains-learn.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Add `ignoreSelectors` option to `selector-max-compound-selectors` rule
```
ybiquitous, line 5: [suggest] Let's follow the pull request title:

```suggestion
Added: `ignoreSelectors: []` to `selector-max-compound-selectors`
```

lib/rules/selector-max-compound-selectors/index.mjs
ybiquitous: I prefer the current code because `Number()` is slightly less explicit.

ybiquitous (changes requested):
Thanks for the pull request. Can you add the option to the rule's README?

.changeset/big-trains-learn.md
Mouvedia: As long as `Add` is corrected I am good.

```suggestion
Added: `ignoreSelectors` option to `selector-max-compound-selectors` rule
```

would be acceptable too.

ybiquitous: For consistency. See:

https://github.com/stylelint/stylelint/blob/b6446093297c9a6a147c3c967067f06c12d76ab1/CHANGELOG.md?plain=1#L16

lib/rules/selector-max-compound-selectors/README.md
```diff
@@ -59,3 +59,43 @@ div {}
...
+
+<!-- prettier-ignore -->
+```css
+.foo ::v-deep .baz {}
+```
+
+<!-- prettier-ignore -->
+```css
+.some {
+  &-ignored-class ::v-deep > .bar {}
```
ybiquitous, line 99: [suggest] `&-ignored-class` is non-standard CSS (e.g. SCSS), right? I think we should avoid non-standard syntaxes even in examples.

See also: https://github.com/stylelint/stylelint/blob/b6446093297c9a6a147c3c967067f06c12d76ab1/docs/developer-guide/rules.md?plain=1#L11-L13

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
ybiquitous: Agree. We want test cases using `:not()` like this:

https://github.com/stylelint/stylelint/blob/da72434387fde008df95a13eb4f8ecbb22b8c68c/lib/rules/selector-max-compound-selectors/__tests__/index.mjs#L123

lib/rules/selector-max-compound-selectors/README.md
FloEdelmann: You're right, see https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Using_CSS_nesting#concatenation_is_not_possible. I changed it in [`1df6e37`](https://github.com/stylelint/stylelint/pull/7544/commits/1df6e37fee0043b8da359035cd294d7307027b36).

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
FloEdelmann: I'm not sure which tests I should add. Could you tell me (high-level) the code you would like to be tested as accepted and not accepted?

Mouvedia: It depends on my question :

> `:global(.foo)` count as 1 or 0?

A If it's currently 0 then we don't need a new test case.
B If it's not then you gotta at least check that the behaviour is coherent when unpacking the _complex real selector list_ of the ignored selector.
e.g. (if B) `:not(.a, div)` should be 2 based on

https://github.com/stylelint/stylelint/blob/b6446093297c9a6a147c3c967067f06c12d76ab1/lib/rules/selector-max-compound-selectors/README.md?plain=1#L60

ybiquitous: `:global()` is non-standard, so let's ignore it here.

I think it's enough to add cases for a functional pseudo-class e.g. `:not()`:

Accept:

```css
:not(.foo ::v-deep .bar) {}
```

Reject:

```css
:not(.foo ::v-deep .bar .baz) {}
```

Mouvedia: > Reject:
> 
> ```css
> :not(.foo ::v-deep .bar .baz) {}
> ```

If it's for `2` with an ignore of `::v-deep`, it's an incorrect test.
If it's for `2` with an ignore of `:not`, it's a correct test.

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
```diff
@@ -384,3 +384,82 @@ testRule({
...
+			description: 'selector ignored by regex',
+		},
+		{
+			code: '.foo::v-deep :-webkit-full-screen a ::-moz-placeholder {}',
+			description: 'multiple ignored selectors',
+		},
+		{
+			code: '.foo { & ::v-deep > .bar {} }',
+			description: 'nested ignored selector',
+		},
```
Mouvedia, line 433: question
major

How should `'.foo .baz .bar.quz {}'` behave for `2` ignore `[".bar", ".quz"]`?
i.e. does it remove one by one from the compound selector or does it skip it if it's not exactly a match?

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
ybiquitous: I meant using `config: [2, { ignoreSelectors: ['::v-deep', '/-moz-.*/', /-screen$/] }]` at L31. So, I need clarification on why the following is incorrect.

> If it's for `2` with an ignore of `::v-deep`, it's an incorrect test.

Mouvedia: > I meant using `config: [2, { ignoreSelectors: ['::v-deep', '/-moz-.*/', /-screen$/] }]` at L31. So, I need clarification on why the following is incorrect.
> 
> > If it's for `2` with an ignore of `::v-deep`, it's an incorrect test.

It was not immediately clear to me until I read :

> `:not()` is considered one compound selector irrespective to the complexity of the selector inside it. The rule does process that inner selector, but does so separately, independent of the main selector.

so

> Reject:
> 
> ```css
> :not(.foo ::v-deep .bar .baz) {}
> ```

wasn't 4 but 1 *and* 3.
i.e. two separate evaluations

My bad, I missed that warning in the documentation. The test is correct.
Could we add a `> [!WARNING]` to

https://github.com/stylelint/stylelint/blob/b6446093297c9a6a147c3c967067f06c12d76ab1/lib/rules/selector-max-compound-selectors/README.md?plain=1#L17

?

Mouvedia: > Could you tell me (high-level) the code you would like to be tested as accepted and not accepted?

@FloEdelmann

`config: [2, { ignoreSelectors: [':not'] }]`

Accept:

```css
p a :not(.foo .bar) {}
```

Reject:

```css
p a :not(.foo .bar .baz) {}
```

i.e. if `not` is ignored, does the inner selector gets evaluated so that it can be rejected or should you bail and get rid of it completely and hence

```css
p a :not(.foo .bar .baz) {}
```

should pass too?

FloEdelmann: > Could we add a `> [!WARNING]` to […]?

I added it in [`0cba157`](https://github.com/stylelint/stylelint/pull/7544/commits/0cba1578168089aee3d931e47a4fad64cb066ffa). I needed to add a `<!-- prettier-ignore -->` comment above it, otherwise Prettier would merge the two lines together into one, apparently because the second line starts with inline code (`` ` ``).

> […] hence `p a :not(.foo .bar .baz) {}` should pass too?

I don't think ignoring `:not` should also ignore contents of it.

---

I changed the implementation to support ignoring `:not` and updated the tests and docs accordingly in [`e174fc5`](https://github.com/stylelint/stylelint/pull/7544/commits/e174fc57a6949389cec58773879cef5e146a7ff2).

Mouvedia: > I don't think ignoring :not should also ignore contents of it.

Ill add that to OP.

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
Mouvedia: @FloEdelmann ping

Mouvedia:
@ybiquitous should we move on and leave [the note](https://github.com/stylelint/stylelint/pull/7544#discussion_r1509938213) for later?  
Or push it to the next version?

ybiquitous:
@Mouvedia Good point. I think adding a not is a good idea to reduce people's confusion.

In the example you suggested,

```json
{
  "selector-max-compound-selectors": [2, {"ignoreSelectors": [".bar", ".quz"]}]
}
```

```css
.foo .baz .bar.quz {}
```

it should be considered as:

```css
.foo .baz {} /* ".bar" and ".quz" are ignored */
```

because it seems easier to implement by using the current selector parser (`postcss-selector-parser`).

Mouvedia:
Whatever the statu quo is, I am fine with it; but it needs to have a pass/fail example or a note.
It's not a blocker so it could be punted to another PR.

ybiquitous:
Although it's not a blocker, I think adding a note in this PR is preferable.

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
```diff
@@ -384,3 +384,108 @@ testRule({
...
```
ybiquitous, line 433:
```suggestion
		},
		{
			code: 'p a > .ignored .ignored {}',
			description: 'doubled class selectors ignored',
		},
		{
			code: 'p a > .ignored.ignored {}',
			description: 'concatenated class selectors ignored one by one',
		},
```

ybiquitous (commented):
@Mouvedia, I think adding a code example instead of a note is enough, but what do you think?

@FloEdelmann I would like to include this PR in the next version. If you don't have time, I'll push additional commits instead.

lib/rules/selector-max-compound-selectors/index.mjs
```diff
@@ -41,6 +68,7 @@ const rule = (primary) => {
 		 */
 		function checkSelector(selectorNode, ruleNode) {
 			let compoundCount = 1;
+			let previousNodeWasCompoundSeparator = true; // first node is implicitly preceded by a separator
```
ybiquitous, line 71:
```suggestion
```

lib/rules/selector-max-compound-selectors/index.mjs
```diff
@@ -41,6 +68,7 @@ const rule = (primary) => {
...
 
 			selectorNode.each((childNode) => {
```
ybiquitous, line 73:
```suggestion
			// @ts-expect-error -- TODO: The `index` argument type is missing. See https://github.com/postcss/postcss-selector-parser/pull/289
			selectorNode.each((childNode, index) => {
```

lib/rules/selector-max-compound-selectors/index.mjs
```diff
@@ -51,7 +79,11 @@ const rule = (primary) => {
 				// Compound selectors are separated by combinators, so increase count when meeting one
 				if (childNode.type === 'combinator') {
 					compoundCount++;
+				} else if (isSelectorIgnored(childNode)) {
+					compoundCount -= previousNodeWasCompoundSeparator ? 1 : 0;
 				}
+
+				previousNodeWasCompoundSeparator = childNode.type === 'combinator';
```
ybiquitous, line 86:
```suggestion
					compoundCount++;
					return;
				}

				// Try ignoring the selector if the current node is the first, or the previous node is a combinator
				if (
					(index === 0 || selectorNode.at(index - 1)?.type === 'combinator') &&
					isSelectorIgnored(childNode)
				) {
					compoundCount--;
				}
```

ybiquitous (commented):
[suggest] I found that we can remove the `previousNodeWasCompoundSeparator` variable outside the `each` loop by using the `index` second argument. As a result, we can put the ignoring code in one `if` condition.

lib/rules/selector-max-compound-selectors/index.mjs
```diff
@@ -22,17 +24,42 @@ const meta = {
...
+		 * @returns {boolean}
+		 */
+		function isSelectorIgnored(selectorNode) {
+			const selectorString = selectorNode.toString();
+			const selectorStringWithoutParentheses = selectorString.replace(/\(.*\)$/, '');
+
+			return (
+				optionsMatches(secondaryOptions, 'ignoreSelectors', selectorString) ||
+				optionsMatches(secondaryOptions, 'ignoreSelectors', selectorStringWithoutParentheses)
+			);
```
ybiquitous, line 60: [suggest] How about simplifying the code like this?

```suggestion
			const selector =
				selectorNode.type === 'pseudo' ? selectorNode.value : selectorNode.toString();

			return optionsMatches(secondaryOptions, 'ignoreSelectors', selector);
```

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
ybiquitous: See https://github.com/stylelint/stylelint/pull/7544#pullrequestreview-1931665796

FloEdelmann:
@ybiquitous Sorry for not responding. I currently don't have much time to work on this PR, so please go ahead and edit to your liking!

Mouvedia:
> @Mouvedia I think adding a code example instead of a note is enough, but what do you think?

Either one is fine.

lib/rules/selector-max-compound-selectors/README.md
```diff
@@ -59,3 +61,53 @@ div {}
...
+```css
+p a :not(.foo .bar) {}
+```
+
+<!-- prettier-ignore -->
+```css
+.foo {
+  &.some-ignored-class ::v-deep > .bar {}
+}
+```
```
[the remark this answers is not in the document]
ybiquitous: Fixed via 97797463b

lib/rules/selector-max-compound-selectors/index.mjs
ybiquitous: Fixed via af3d4047e

ybiquitous (approved):
I pushed additional commits.

@Mouvedia Can you take a look again?

ybiquitous:
I bumped `postcss-selector-parser` and removed the patch (91c2a8edc) since the new version has been released:
https://github.com/postcss/postcss-selector-parser/releases/tag/v6.0.16

lib/rules/selector-max-compound-selectors/README.md
```diff
@@ -59,3 +61,63 @@ div {}
...
+
+<!-- prettier-ignore -->
+```css
+p a :not(.foo .bar) {}
+```
+
+<!-- prettier-ignore -->
+```css
+.foo {
+  &.some-ignored-class ::v-deep > .bar {}
```
Mouvedia, line 111: nit
non-blocking

rename to `.ignored`
for consistency with the following examples

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
```diff
@@ -384,3 +384,116 @@ testRule({
...
+		{
+			code: '.foo > ::v-deep .bar {}',
+			description: 'ignored selector preceded by a combinator',
+		},
+		{
+			code: '.foo input::-moz-placeholder {}',
+			description: 'selector ignored by string regex',
+		},
+		{
+			code: '.foo :-webkit-full-screen a {}',
```
Mouvedia, line 423: nit
non-blocking

You could refactor by removing `/-screen$/` and just depending on `/-moz-.*/`.
 The reason being that both `::-moz-placeholder`  *and* `:-moz-placeholder` exist.
i.e. `.foo:-moz-placeholder a {}`

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
```diff
@@ -384,3 +384,116 @@ testRule({
...
+			description: 'selector ignored by regex',
+		},
+		{
+			code: '.foo::v-deep :-webkit-full-screen a ::-moz-placeholder {}',
```
Mouvedia, line 427: If you do the refactor, this would become `.foo ::v-deep input ::-moz-placeholder {}`.

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
```diff
@@ -384,3 +384,116 @@ testRule({
...
+			code: '.foo .bar:not(.baz) {}',
+		},
+		{
+			code: 'p a :not(.foo .bar) {}',
+		},
+	],
+
+	reject: [
+		{
+			code: 'p a :not(.foo .bar .baz) {}',
```
Mouvedia, line 490: minor
non-blocking

We could also check that `,` is supported.
i.e. `p a :not(.foo .bar, .foo .baz .qux) {}`

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
```diff
@@ -384,3 +384,116 @@ testRule({
...
 });
+
+// Testing ignoreSelectors
+testRule({
+	ruleName,
+	config: [2, { ignoreSelectors: ['::v-deep', '/-moz-.*/', /-screen$/, '.ignored'] }],
+
+	accept: [
+		{
+			code: '::v-deep .foo .bar {}',
```
Mouvedia, line 395: minor

`::v-deep` doesn't make sense without a selector before
replace with `'.ignored .foo .bar {}'`

Mouvedia (approved):
All of the changes I have requested are non-blocking.
Feel free to integrate them at your convenience.

LGTM

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
ybiquitous: I think this case checks a regex-like string (`'/-moz-.*/'`) and a regex (`/-screen$/`), so keeping it as-is makes sense for me.

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
ybiquitous: I have the same reason as https://github.com/stylelint/stylelint/pull/7544#discussion_r1524024793

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
ybiquitous: Good catch. Added via 411ed1e9e

lib/rules/selector-max-compound-selectors/__tests__/index.mjs
ybiquitous: Fixed via 411ed1e9e

ybiquitous:
Thank you for creating and reviewing the pull request!

FloEdelmann:
Thanks for bringing this over the finish line!

Mouvedia:
@ybiquitous we have missed something apparently.
relevant ref: https://github.com/stylelint/stylelint/pull/7544#discussion_r1509938213

for `config: [2, { ignoreSelectors: ['.ignored'] }],`
and `'.foo .bar > .ignored.qux {}'`
what would expect to happen ?

Since it's exact—i.e. not `"/ignored/"`—I would expect the compound selector to not be ignored and hence end up with 3 and the test to fail instead of passing currently.
Am i wrong?

ybiquitous:
@Mouvedia Thanks, your concern is right. We have to fix the case.

The following makes sense to me, so I'll open a follow-up PR soon:

```css
.foo .bar > .ignored.qux {}
.foo .bar > .qux.ignored {}
/* considered as ↓ */
.foo .bar > .qux {}
```

ybiquitous:
Opened follow-up PR #7559

Issue #7506, opened 2024-02-01
Fix `declaration-block-no-shorthand-property-overrides` false negatives for `border`


hudochenkov:
### What minimal example or steps are needed to reproduce the bug?

```css
a {
	border-color: red;
	border: 1px solid blue;
}
```

### What minimal configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "declaration-block-no-shorthand-property-overrides": true
  }
}
```

### How did you run Stylelint?

All ways.

### Which Stylelint-related dependencies are you using?

```json
"stylelint": "16.2.0"
```

### What did you expect to happen?

To get a violation:

```
Unexpected shorthand "border" after "border-color"
```

### What actually happened?

No violation is reported.

### Do you have a proposal to fix the bug?

Update [shorthands reference](https://github.com/stylelint/stylelint/blob/328544d8dfe4abfb1d75ab6631528f2e3c605a74/lib/reference/properties.mjs#L15).

Looking at the reference, it could be further improved. E. g. `border` doesn't have `border-block`, `border-block-end`, etc. in it's set. Which means that there are more unreported violations.

ybiquitous:
@hudochenkov Thanks for the report. Right, this is a bug of false negatives.

> Update [shorthands reference](https://github.com/stylelint/stylelint/blob/328544d8dfe4abfb1d75ab6631528f2e3c605a74/lib/reference/properties.mjs#L15).

This update only may not be enough. It may be necessary to fix the rule.

> Looking at the reference, it could be further improved. E. g. `border` doesn't have `border-block`, `border-block-end`, etc. in it's set. Which means that there are more unreported violations.

Sounds good. Let's fix them together.

I've labeled the issue as **ready to implement**. Please consider [contributing](https://stylelint.io/contributing) if you have time.

There are [steps on how to fix a bug in a rule](https://stylelint.io/developer-guide/rules#fix-a-bug-in-a-rule) in the Developer guide.

G-Rath:
I might have a crack at this

G-Rath:
This is fixed for the given example but not when the property order is switched:

```
a {
  border: 1px solid blue;
  border-color: red;
}
```

While you could argue that might not be a problem of the specifically mentioned rule since it's setting a different value, it also applies when there isn't e.g. a color on `border:

```
a {
  border: 1px solid;
  border-color: red;
}
```

PR #7546, opened 2024-03-06
fix: deduplicate `border-style` and `border-color`

G-Rath:

Relates to #7506

From what I can tell, I think it'll be a lot more work to support deduplicating `border-width` (because you can have multiple values, so it's only safe to interact when there's just one value) and with `border` (because you need to consider what options are already support).

I've done this in the meantime because it's safe and easy improvement to land straight away.

~I've not added a changeset yet because it seems to want a GitHub Personal Token which doesn't seem right - if someone wants to handle that instead, be my guest :)~

### DIFF
--- a/.changeset/new-chefs-give.md
+++ b/.changeset/new-chefs-give.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+fix: deduplicate `border-style` and `border-color`
--- a/lib/reference/properties.cjs
+++ b/lib/reference/properties.cjs
@@ -65,6 +65,14 @@ const longhandSubPropertiesOfShorthandProperties = new Map([
 		]),
 	],
 	[
+		'border',
+		new Set([
+			// prettier-ignore
+			'border-style',
+			'border-color',
+		]),
+	],
+	[
 		'border-block',
 		new Set([
 			// prettier-ignore
--- a/lib/reference/properties.mjs
+++ b/lib/reference/properties.mjs
@@ -61,6 +61,14 @@ export const longhandSubPropertiesOfShorthandProperties = new Map([
 		]),
 	],
 	[
+		'border',
+		new Set([
+			// prettier-ignore
+			'border-style',
+			'border-color',
+		]),
+	],
+	[
 		'border-block',
 		new Set([
 			// prettier-ignore
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
@@ -48,6 +48,15 @@ testRule({
 			code: 'a { top: 0; right: 0; bottom: 0; }',
 		},
 		{
+			code: 'a { border: solid; }',
+		},
+		{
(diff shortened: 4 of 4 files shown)


ybiquitous (changes requested):
@G-Rath Thanks for the pull request. It's worth fixing these false negatives around the `border` shorthand.

But this fix doesn't make sense because the newly added `border` overwrites the existing `border` entry:

https://github.com/stylelint/stylelint/blob/74145ce67ae9c5afe8b61f6917ffd73dfc99c16c/lib/reference/properties.mjs#L45-L62

This case is tricky... 🤔

G-Rath:
> But this fix doesn't make sense because the newly added border overwrites the existing border entry:

I'm not sure what you mean - I've added a test that shows with this change the rule will now autocorrect code that it didn't before, which to me indicates its not being overwritten?

So to me this is at least an _improvement_ even though it doesn't resolve everything.

ybiquitous:
Please check out the failure on CI:

```
FAIL lib/rules/declaration-block-no-shorthand-property-overrides/__tests__/index.mjs
  ● declaration-block-no-shorthand-property-overrides › reject › [ true ] › 'a { border-top-width: 1px; top: 0; bottom: 3px; border: 2px solid blue; }' › no description

    expect(received).toHaveLength(expected)

    Expected length: 1
    Received length: 0
    Received array:  []

      at Object.<anonymous> (node_modules/jest-preset-stylelint/getTestRule.js:81:29)
```

https://github.com/stylelint/stylelint/actions/runs/8166175436/job/22369564288?pr=7546#step:7:1379

G-Rath:
huh that's weird - I had that locally but fixed it before pushing:

```
Test Suites: 1 passed, 1 total
Tests:       81 passed, 81 total
Snapshots:   0 total
Time:        0.382 s, estimated 1 s
Ran all test suites matching /declaration-block-no-redundant-longhand-properties/i
```

ybiquitous:
According to the [CSS spec](https://drafts.csswg.org/css-backgrounds/#propdef-border), I think `longhandSubPropertiesOfShorthandProperties` should have the `border` entry as:

```js
	[
		'border',
		new Set([
			// prettier-ignore
			'border-width',
			'border-style',
			'border-color',
		]),
	],
```

If it makes sense, we have to fix the `declaration-block-no-redundant-longhand-properties` and `declaration-block-no-shorthand-property-overrides` logic, using the `border` properties.

G-Rath:
@ybiquitous unless I've missed something my expectation is that will be hard to support as that's only true if each long-hand property has only a single value which I'm not sure if the current implementation is already expecting; if its not, then there will need to be some way to express that in the configuration of `longhandSubPropertiesOfShorthandProperties`.

Digging a bit deeper, I think part of the issue is that the logic currently doesn't expect there to be multiple combinations of long-forms for the same properties, which I thought it did - and so right now its implemented to detect when someone has _all_ of the properties for setting color, width, and style for all four borders.

Unfortunately I think I'm going to have to put this down as it looks like improving this will require a decent refactor and I've not got enough experience with the codebase and related CSS to tackle that at this point.

Issue #7488, opened 2024-01-21
Fix `rule-selector-property-disallowed-list` false positives for nesting selectors


emmacharp:
### What minimal example or steps are needed to reproduce the bug?

```css
.search {
   & input {
      width: 100%;
   }
}
```

### What minimal configuration is needed to reproduce the bug?

```json 
{ 
  "rules": { 
    "rule-selector-property-disallowed-list": [{ "/^((?!input).)*$/": "width" }]
  }
}
```

### How did you run Stylelint?

```shell
stylelint *.css
```

[Online demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0BnGBDATmAFgAmAB0A7PPAMjwEtiAHAVwBcCSyyB3KqRnePARgAMggKQBuPAHoAVHhQ4I9ADZQ8AIxh4IAawA06pnnRhG9dEqUBPPJhi0ImRjFXTJrPAF8SX4iVQZsfBoGZiJSPE5uXgFhcSlZYghmW3tHZ30dPFdvEj80LFw8ADMHAFsWcMiePiFRCRkbOwcnKAztLLdiDxBdECKqJRgAOXRSuEQYAA9R2kGkMBQUHvAIYn6AcwQQMLJCEExlGBQ9vh22PYPBgFo0QZMHK9pMCFoYR0srqCoUcyUIdmcVyU30YJzwAG1gHg9pIAHoACnhAH4AITBJgASiQGOkABI3AhoSAqjg9p4ALruHzdXqQNZUdYAMTK6EYWwAVihVstYLQlogznsUIxLINgcRQYS9kpWUdJd4QB4gA)

### Which Stylelint-related dependencies are you using?

```json
{
   "stylelint": "16.2.0"
}
```

### What did you expect to happen?

The nested block should not generate an error.

### What actually happened?

`.search` is evaluated and an error is thrown.

### Do you have a proposal to fix the bug?

Don't evaluate nested selectors... though I'm not sure about collateral damage.

ybiquitous:
@emmacharp Thanks for the report with a reproduction. To clarify, do you want to disallow a `width` prop within all `input` selectors?

[...]

Issue #7557, opened 2024-03-13
Rule options where users set selectors


romainmenke:
_Opening a new issue to discuss this separate from any individual bug report or refactor because I am first looking for some guidance on which direction is more favorable._

--------

Examples : 
- `rule-selector-property-disallowed-list`: disallow specific properties when a selector matches a pattern
- `selector-disallowed-list`: disallow a list of selectors

It used to be relatively fine to treat selectors as strings and to use tools like regex but I don't think this works well anymore.

With nesting it is ambiguous if users want to match the resolved or unresolved string.
Even when they agree they want to match the resolved string they might not agree on the method (sass like or following the spec).

But that is not the only bit of complexity.

`:not()` is a negation, so should `:not(input) { width: 100% }` error when disallowing `width` for `/input/` ?

`:has()` matches elements when other elements match a selector, not when it itself is something. Same as before, should this error `:has(input) { width: 100% }`?

--------

The underlying issue is that string matching can only go so far and that the outcome will always be surprising to some users.

I personally think that there should be an effort to use a more semantic approach.
Alternative we could embrace that these rules operate on strings and that they are agnostic of parent rules, nesting, selector matching mechanics,...

_Any rule that enforces something onto tags, classnames, types, attributes will be unaffected by this. They already target specific selector parts._

A concrete example is : https://github.com/stylelint/stylelint/issues/7488

This rule treats selectors as strings and doesn't consider their semantic meaning or which elements they would match.

Updating that rule so that nested selectors are resolved would be surprising as it can radically change what does or doesn't give errors.

> [!Note]
> Keep in mind that nesting has an implicit `&`
> so `.foo { input {} }` would resolve to `.foo input` and would no longer match `"input"` as a plugin option.
> That alone would be a breaking change.

But the same rule would also be really helpful if it was implemented differently.
It could look at selectors in the same way browsers do and only disallow properties when the right most compound selector would be a semantic match.

----------

I think that handling this case by case always favors the current bug reporter as we all want to help people :) This might mean that we accidentally end up somewhere that is worse instead of better.

So I am looking for input here as I don't know what others think is the better solution here.

ybiquitous:
@romainmenke Thank you for opening the discussion. I think your concern absolutely makes sense. With the modern syntaxes or components (e.g. nested selectors, `:not()`, etc.), I agree that string-based or regex-based filtering has been gradually confusing.

At this point, the ignoring behavior of each rule depends on the selector parser mostly.

[...]

PR #7558, opened 2024-03-13
Fix `rule-selector-property-disallowed-list` false positives for nesting selectors

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/7488

I noticed that this walks down the entire sub tree for each rule node.
This might not be what users expect.

But changing it to `.each()` also has issues.
It will no longer give errors on :

```css
.bad-selector {
  @media screen {
    color: red;
  }
}
```

Edit: I've found a more complete patch

### DIFF
--- a/lib/rules/rule-selector-property-disallowed-list/__tests__/index.mjs
+++ b/lib/rules/rule-selector-property-disallowed-list/__tests__/index.mjs
@@ -104,3 +104,49 @@ testRule({
 		},
 	],
 });
+
+testRule({
+	ruleName,
+	config: {
+		'/^((?!input).)*$/': 'color',
+	},
+
+	accept: [
+		{
+			code: '.foo input { color: red; }',
+		},
+		{
+			code: '.foo { & input { color: red; } }',
+		},
+		{
+			code: 'input .foo { color: red; }',
+		},
+	],
+
+	reject: [
+		{
+			code: '.foo form { color: red; }',
+			message: messages.rejected('.foo form', 'color'),
+			line: 1,
+			column: 13,
+			endLine: 1,
+			endColumn: 18,
+		},
+		{
+			code: '.search { & form { color: red; } }',
+			message: messages.rejected('& form', 'color'),
+			line: 1,
+			column: 20,
+			endLine: 1,
+			endColumn: 25,
+		},
+		{
+			code: 'input { & .foo { color: red; } }',
+			message: messages.rejected('& .foo', 'color'),
+			line: 1,
+			column: 18,
+			endLine: 1,
+			endColumn: 23,
+		},
+	],
+});
--- a/lib/rules/rule-selector-property-disallowed-list/__tests__/index.mjs
+++ b/lib/rules/rule-selector-property-disallowed-list/__tests__/index.mjs
@@ -47,6 +47,14 @@ testRule({
 			endColumn: 27,
 		},
 		{
+			code: 'a { @media screen { color: red; } }',
+			message: messages.rejected('a', 'color'),
+			line: 1,
+			column: 21,
+			endLine: 1,
+			endColumn: 26,
+		},
+		{
 			code: 'a { margin-top: 0px; }',
 			message: messages.rejected('a', 'margin-top'),
 			line: 1,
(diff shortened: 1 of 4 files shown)


ybiquitous:
@romainmenke Thank you for the pull request! This is just an idea, but how about resolving and flattening selectors like #7518?

ybiquitous:
> resolving and flattening selectors

Umm... I realize that this way is problematic, as #7557 describes.

lib/rules/rule-selector-property-disallowed-list/index.mjs
```diff
@@ -49,14 +49,18 @@ const rule = (primary) => {
 				return;
 			}
 
-			ruleNode.walkDecls((decl) => {
-				const { prop } = decl;
+			ruleNode.walk((node) => {
+				if (node.type !== 'decl') return;
+
+				if (!declarationIsAChildOfRule(node, ruleNode)) return;
```
romainmenke, line 55: @ybiquitous I think that this check matches the original intent of this rule and fixes the issues for nesting.

The original bug report happend when validation `.foo {}` for `.foo { & input {} }`. This change adds an early return so that nested declarations are only tested against their most immediate ancestor rule. (if that makes sense?)

ybiquitous: Yes, that makes sense! 👍🏼

lib/rules/rule-selector-property-disallowed-list/__tests__/index.mjs
```diff
@@ -89,6 +97,14 @@ testRule({
...
 			endColumn: 27,
 		},
+		{
+			code: 'a { b { width: 100% } color: red; }',
+			message: messages.rejected('a', 'color'),
+			line: 1,
+			column: 23,
+			endLine: 1,
+			endColumn: 28,
+		},
```
ybiquitous, line 107: [suggest] How about adding the following case?

```js
{
	code: 'b { a { color: red; } }',
	message: messages.rejected('a', 'color'),
	// ...
}
```

Besides, how about adding more *accepted* cases? E.g.

```js
	{
		code: 'a { b { color: red; } }',
	},
	{
		code: 'b { & a { color: red; } }',
	},
```

lib/rules/rule-selector-property-disallowed-list/index.mjs
```diff
@@ -49,14 +49,18 @@ const rule = (primary) => {
 				return;
 			}
 
-			ruleNode.walkDecls((decl) => {
-				const { prop } = decl;
+			ruleNode.walk((node) => {
+				if (node.type !== 'decl') return;
```
ybiquitous, line 53: [suggest] Is this the same as the original code?

```suggestion
			ruleNode.walkDecls((decl) => {
```

lib/rules/rule-selector-property-disallowed-list/index.mjs
```diff
@@ -71,3 +75,26 @@ rule.ruleName = ruleName;
...
 export default rule;
+
+/**
+ * Check that a given declaration is a child of this rule and not a nested rule
+ *
+ * @param {import('postcss').Declaration} decl
+ * @param {import('postcss').Rule} ruleNode
+ */
+function declarationIsAChildOfRule(decl, ruleNode) {
+	if (decl.parent === ruleNode) return true;
```
ybiquitous, line 86: [suggest] This code seems to be the same as `if (parent === ruleNode) return true;` in the `while` block, so maybe we may remove it:

```suggestion
```

lib/rules/rule-selector-property-disallowed-list/index.mjs
```diff
@@ -71,3 +75,26 @@ rule.ruleName = ruleName;
...
+
+	/** @type {import('postcss').Document|import('postcss').Root|import('postcss').Container|undefined} */
```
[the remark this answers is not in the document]
romainmenke: I didn't know about this type, thank you!

lib/rules/rule-selector-property-disallowed-list/index.mjs
romainmenke: Ah, yes, I changed it back and forth a couple of times, but the original is better and is what we want.

romainmenke:
Thank you for the review @ybiquitous 🙇

ybiquitous (approved):
Thank you! LGTM 👍🏼

PR #7559, opened 2024-03-14
Fix `selector-max-compound-selectors` with `ignoreSelectors` for class selectors

ybiquitous:

Follow-up to PR #7544
See https://github.com/stylelint/stylelint/pull/7544#issuecomment-1996915259

Note that this change doesn't need any changelog item since the `ignoreSelectors` option is unreleased.

### DIFF
--- a/lib/rules/selector-max-compound-selectors/index.cjs
+++ b/lib/rules/selector-max-compound-selectors/index.cjs
@@ -2,24 +2,26 @@
 // please instead edit the ESM counterpart and rebuild with Rollup (npm run build).
 'use strict';
 
-const resolvedNestedSelector = require('postcss-resolve-nested-selector');
+const resolveNestedSelector = require('postcss-resolve-nested-selector');
+const selectorParser = require('postcss-selector-parser');
 const validateTypes = require('../../utils/validateTypes.cjs');
 const isContextFunctionalPseudoClass = require('../../utils/isContextFunctionalPseudoClass.cjs');
 const isNonNegativeInteger = require('../../utils/isNonNegativeInteger.cjs');
 const isStandardSyntaxRule = require('../../utils/isStandardSyntaxRule.cjs');
 const optionsMatches = require('../../utils/optionsMatches.cjs');
 const parseSelector = require('../../utils/parseSelector.cjs');
+const pluralize = require('../../utils/pluralize.cjs');
 const report = require('../../utils/report.cjs');
 const ruleMessages = require('../../utils/ruleMessages.cjs');
 const validateOptions = require('../../utils/validateOptions.cjs');
 
+const { isCombinator, isPseudo, isRoot, isSelector } = selectorParser;
+
 const ruleName = 'selector-max-compound-selectors';
 
 const messages = ruleMessages(ruleName, {
-	expected: (selector, max) =>
-		`Expected "${selector}" to have no more than ${max} compound ${
-			max === 1 ? 'selector' : 'selectors'
-		}`,
+	expected: (selector, /** @type {number} */ max) =>
+		`Expected "${selector}" to have no more than ${max} compound ${pluralize('selector', max)}`,
 });
 
 const meta = {
@@ -54,8 +56,7 @@ const rule = (primary, secondaryOptions) => {
 		 * @returns {boolean}
 		 */
 		function isSelectorIgnored(selectorNode) {
-			const selector =
-				selectorNode.type === 'pseudo' ? selectorNode.value : selectorNode.toString();
+			const selector = isPseudo(selectorNode) ? selectorNode.value : selectorNode.toString();
 
 			return optionsMatches(secondaryOptions, 'ignoreSelectors', selector);
 		}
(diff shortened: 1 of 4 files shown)


lib/rules/selector-max-compound-selectors/index.cjs
```diff
@@ -67,34 +68,35 @@ const rule = (primary, secondaryOptions) => {
...
-			if (
-				selectorNode.type !== 'root' &&
-				selectorNode.type !== 'pseudo' &&
-				compoundCount > primary
-			) {
+			// Normalize selector nodes and count combinators
+			const combinatorCount = filteredChildNodes
+				.filter((node, i, nodes) => {
+					// Filter out a consecutive combinator
+					return !(isCombinator(node) && i > 0 && isCombinator(nodes[i - 1]));
```
Mouvedia, line 89: question

I might be missing something here but the comment says "consecutive combinator".
AFAIK that be ` `,  `+`, `~` and `>`.
Why would `.foo + + .bar` matter? Isn't that invalid?

lib/rules/selector-max-compound-selectors/index.mjs
```diff
@@ -64,34 +65,35 @@ const rule = (primary, secondaryOptions) => {
...
-			if (
-				selectorNode.type !== 'root' &&
-				selectorNode.type !== 'pseudo' &&
-				compoundCount > primary
-			) {
+			// Normalize selector nodes and count combinators
+			const combinatorCount = filteredChildNodes
+				.filter((node, i, nodes) => {
+					// Filter out a consecutive combinator
+					return !(isCombinator(node) && i > 0 && isCombinator(nodes[i - 1]));
```
Mouvedia, line 86: minor

```js
const isLeftover = isCombinator(node) && i > 0 && isCombinator(nodes[i - 1]);

return !isLeftover;
```

<hr />

refactor

It doesn't matter which one you remove, but I would have expected to remove the left one.
That would require to check whether `i + 1` exists which would enable to have only 2 filters.
i.e. the l91 one could be performed at the same time
e.g. `return !(isLeftover || isFirst || isLast);`

Mouvedia (changes requested):

Mouvedia:
You can ignore my review if you are in a hurry.
What matters is that the behaviour is now consistent.
Hence this can be merged as is.

Mouvedia:
> Note that this change doesn't need any changelog item since the ignoreSelectors option is unreleased.

You could add `& ([@ybiquitous](https://github.com/ybiquitous))` to

https://github.com/stylelint/stylelint/blob/6f24e261193adb08ce7936cd582c17a22768e72f/CHANGELOG.md?plain=1#L5

lib/rules/selector-max-compound-selectors/index.mjs
ybiquitous: @Mouvedia Sorry for the late reply. Based on your feedback, I just tried refactoring (b7a723440).

ybiquitous:
> You could add `& ([@ybiquitous](https://github.com/ybiquitous))` to

Don't mind the mention in the changelog since I'll release it. 😅

lib/rules/selector-max-compound-selectors/index.mjs
```diff
@@ -64,34 +65,35 @@ const rule = (primary, secondaryOptions) => {
...
+				})
+				.filter((node, i, nodes) => {
+					// Filter out a combinator at the edge
+					return !(isCombinator(node) && (i === 0 || i === nodes.length - 1));
+				})
+				.filter(isCombinator).length;
+
+			const compoundCount = combinatorCount + 1;
+
+			if (!isRoot(selectorNode) && !isPseudo(selectorNode) && compoundCount > primary) {
```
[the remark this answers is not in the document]
ybiquitous: In preference, extracting a local variable is unnecessary since this `if` block immediately calls `report()`; the current code is clear.

lib/rules/selector-max-compound-selectors/index.mjs
```diff
@@ -64,34 +65,37 @@ const rule = (primary, secondaryOptions) => {
...
+			}, 0);
+
+			const compoundCount = combinatorCount + 1;
 
-			if (
-				selectorNode.type !== 'root' &&
-				selectorNode.type !== 'pseudo' &&
-				compoundCount > primary
-			) {
+			if (!isRoot(selectorNode) && !isPseudo(selectorNode) && compoundCount > primary) {
```
Mouvedia, line 98: minor
perf

You can put an early `if (isRoot(selectorNode) || isPseudo(selectorNode)) return;` in between l80 and l83 so that we don't call the `reduce` in those cases; ending up with

```js
if (compoundCount > primary) {
```

Mouvedia (approved):

lib/rules/selector-max-compound-selectors/index.mjs
ybiquitous: Nice catch! Fixed via debfdec55 👍🏼

PR #7565, opened 2024-03-20
Fix `report()` for `index`/`endIndex` zero values

romainmenke:

None, but related to https://github.com/stylelint/stylelint/pull/7518#discussion_r1532505205

```css
a {}
```

This stylesheet has a selector with a length `1` at the start of a stylesheet.
This is passed to PostCSS as `index: 0` and `endIndex: 0` since `endIndex` is exclusive.

However both Stylelint and PostCSS have boolean checks for `endIndex` instead of a check that it is a number.

So the warning is reported as if no `endIndex` was set.
This does a fallback to the entire length of the node. 

```
a {}
^  ^
```

-------

PostCSS has already been fixed but awaiting release : https://github.com/postcss/postcss/pull/1932

-------

This should have a unit test, but I noticed that there aren't any unit tests for `report.mjs` with `endIndex`.

So I wasn't sure on what the best and most succinct test would be here.

### DIFF
--- a/.changeset/red-fans-wait.md
+++ b/.changeset/red-fans-wait.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `report()` for `index`/`endIndex` zero values
--- a/lib/utils/report.cjs
+++ b/lib/utils/report.cjs
@@ -112,7 +112,7 @@ function report(problem) {
 
 	if (problem.end) {
 		warningProperties.end = problem.end;
-	} else if (endIndex) {
+	} else if (typeof endIndex === 'number') {
 		warningProperties.endIndex = endIndex;
 	}
 
--- a/lib/utils/report.cjs
+++ b/lib/utils/report.cjs
@@ -106,7 +106,7 @@ function report(problem) {
 
 	if (problem.start) {
 		warningProperties.start = problem.start;
-	} else if (index) {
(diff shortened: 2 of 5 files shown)


ybiquitous (commented):
@romainmenke Thanks for the pull request! In addition to `endIndex`, does also checking `index` make sense?

```diff
-	} else if (index) {
+	} else if (typeof index === 'number') {
```

https://github.com/stylelint/stylelint/blob/b05abc8e44e4bb2335404d26b24ac49e1b1afd9e/lib/utils/report.mjs#L105

romainmenke:
Tested with also a check for `index` and that seems safe to me, so I've also added this.

I don't think it makes a practical difference because in the absence of a start index there already are fallbacks to the start in PostCSS, so "not specified" is always equal to `0`.

But I like the consistency :)

-----

I've also updated PostCSS so that the fix is complete.

ybiquitous (approved):
Thank you. LGTM 👍🏼 

Besides, can you add a changelog item according to this PR title?

```markdown
Fixed: `report()` for `index`/`endIndex` zero values
```

After adding the item, feel free to merge 👍🏼

romainmenke:
Thank you for the feedback @ybiquitous 🙇

Mouvedia:
@ybiquitous should we trigger dependabot to get v8.4.38?

PR #7568, opened 2024-03-22
Add experimental support for post processors

### LINKED ISSUE #7567: Add experimental support for post processors

I'm creating a stylelint custom syntax for a css-in-js engine. While lint result from `stylelint.lint` successfully identifies errors/warnings in css, the result is based on their location in css AST. And I would like to post process these locations into original locations in the js code (and therefore have correct squiggly lines from vscode extension).

Btw in ESlint, there's [custom processors](https://eslint.org/docs/latest/extend/custom-processors) `preprocess` which handles similar situation as Stylelint `customSyntax`. And ESlint `postprocess` allows for adjusting the linting result. I have not found a similar feature in Stylelint that would allow for a `postprocess` - like manipulation of the linting result object.

A config option in style lint. Something like a callback `(lintResult: Promise<stylelint.LinterResult>) => Promise<stylelint.LinterResult>`. And `stylelint.lint` will call this callback at the very end, allowing post process of the lint result.

If this request makes sense, I would be happy to contribute :) - edit: here's a draft implementation #7568

YuanboXue-Amber:

Closes #7567

Adding **experimental** config `processors` that allows modifying lint `results` object:

### Type:
```js
// config type:
type ConfigProcessors = string[];

// type of each processor object:
type Processor = {
  name: string;
  postprocess: PostProcessFunction;
};

type PostProcessFunction = (result: LintResult) => void;
```

### Usage:
- In config:
    
    JSON:
    ```json
    {
      "processors": ["some_module"]
    }
    ```
    or multiple processors:
    ```json
    {
      "processors": ["module_A", "module_B"]
    }
    ```

- Example processor:
    ```js
    export default function myProcessor() {
      return {
        name: 'remap-location',
    
        /** @type {(result: LintResult) => void} */
        postprocess(result) {
          result.warnings.forEach(warning => {
            warning.endLine = warning.endLine === undefined ? warning.line : warning.line + 5;
          });
        },
      };
    }
    ```


.changeset/neat-walls-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Add config `processors` that allows modifying lint results.
```
ybiquitous, line 5: [suggest] Let's adjust the format for consistency. Ref https://stylelint.io/maintainer-guide/pull-requests#merging

```suggestion
Added: experimental support for post processors
```

docs/user-guide/configure.md
```diff
@@ -375,6 +375,35 @@ Patterns are applied against the file path relative to the directory of the conf
 
 Overrides have higher precedence than regular configurations. Multiple overrides within the same config are applied in order. That is, the last override block in a config file always has the highest precedence.
 
+## `processors`
+
+> ⚠️ Warning:
```
ybiquitous, line 380: [suggest] Let's use the special notation supported by GitHub. Ref https://github.com/orgs/community/discussions/16925

```suggestion
> [!WARNING]
```

docs/user-guide/configure.md
```diff
@@ -375,6 +375,35 @@ Patterns are applied against the file path relative to the directory of the conf
...
+> This is an experimental feature. The API may change in the future.
```
ybiquitous, line 381: [suggest] Adding a note about the previous `processors` may reduce users' confusion. For example:

```suggestion
> This is an experimental feature. The API may change in the future.
>
> This `processors` property was [removed in 15.0.0](../../migration-guide/to-15.md#removed-processors-configuration-property), but has revived for post-processing. Note that this is different from the previous behavior.
```

Any thoughts?

docs/user-guide/configure.md
```diff
@@ -375,6 +375,35 @@ Patterns are applied against the file path relative to the directory of the conf
...
+
+Processors are functions that hook into Stylelint's pipeline.
+Currently, processors contains only two properties: a string `name` and a function `postprocess`. `postprocess` runs after all rules have been evaluated. This function receives the `result` object of the linting process and can modify it.
+
+```js
+// post-processor.js
+export default {
```
ybiquitous, line 388: [suggest] I prefer a function module over an object module as follows because:

- A function can change the behavior by the `options` argument (more flexibility)
- A function is similar to the previous processor API, so it may reduce the learning cost
- A function can support async execution, e.g. for dynamic `import()`

```js
export default function myProcessor(options = {}) {
  return {
    name: "...",
    postprocess(result) { /* ... */ }
  };
};
```

Any thoughts?

docs/user-guide/configure.md
```diff
@@ -375,6 +375,35 @@ Patterns are applied against the file path relative to the directory of the conf
...
+      warning.endLine =
+        warning.endLine === undefined ? warning.line : warning.line + 5;
+    });
+  }
+};
+```
+
+```json
+{
+  "processors": "path/to/post-processor.js"
```
ybiquitous, line 403: [suggest] To simplify things, how about supporting only an array?

```suggestion
  "processors": ["path/to/post-processor.js"]
```

ybiquitous (changes requested):
@YuanboXue-Amber Thanks for the pull request. Can you take a look at my comments?

docs/user-guide/configure.md
YuanboXue-Amber: I like it 👍 it helps clarify things for users from v14

docs/user-guide/configure.md
YuanboXue-Amber: I don't have a strong preference about supporting string format. My implementation was influenced by config `plugins` and `ignoreFiles` that accept a string or an array. Also supporting string is straightforward and efficient ( just an additional `[config.processors].flat()`). However, if stylelint is moving towards standardizing on array format, I'm open to removing string support.
I will update the documentation to use an array format as it represents a better practice :)

docs/user-guide/configure.md
YuanboXue-Amber: I'm up for changing the object module to function. I'm going to update both documentation and test fixture. 

For 
>A function can change the behavior by the options argument 

Just want to call out that currently I don't have the implementation for accepting options in processors.

YuanboXue-Amber: Here's the [commit](https://github.com/stylelint/stylelint/pull/7568/commits/0ef94c895810bdecea2a86f4828816a835013899) that supports function module processor.

YuanboXue-Amber: I'm thinking a different API is necessary if we want to support adding processor options. 

### Current API:
  The config `processors` type is defined as: `type ConfigProcessors = string | stylelint.Processor | (string | stylelint.Processor)[];`
  And `stylelint.Processor` is the `{name: ..., postprocess: ...}` object.

### API with processor options
To support options to function module processor, the config can look something like:
  ```json
  {
    "processors": [
      { "module": "./path-to-processor.js", "options": { "enabled": true } }
    ]
  }
  ```
  An example of ./path-to-processor.js can be
  ```js
  export default function myProcessor(options = {}) {
    return {
      name: 'remap-location',
      /** @type {(result: LintResult) => void} */
      postprocess(result) { 
        if (options.enabled) { 
          // do something... 
        } 
      }
    };
  }

  ```
  Accordingly, the `processors` API type would be `type ConfigProcessors = string | stylelint.ProcessorConfig | (string | stylelint.ProcessorConfig)[];`
  And `stylelint.ProcessorConfig` is the `{module: ..., options: ...}` object.

### Proposal for initial implementation (this PR)
To keep things simple, how about:
1. initially, we do not support processor options.
2. have just string/string array for config `processors`: `type ConfigProcessors = string | string[];` (or just `string[]` as suggested by another comment). 

This allow us to potentially extend the API later to include processor options, should there for a demand for it.

ybiquitous: Completely agree 👍🏼 Let's avoid supporting `options` for now and simplify the type as `type ConfigProcessors = string[]`.

YuanboXue-Amber:
@ybiquitous I made some updates since your last review. Appreciate another round of check 🙏

docs/user-guide/configure.md
```diff
@@ -375,6 +375,39 @@ Patterns are applied against the file path relative to the directory of the conf
...
+Processors are functions that hook into Stylelint's pipeline.
+Currently, processors contains only two properties: a string `name` and a function `postprocess`. `postprocess` runs after all rules have been evaluated. This function receives the `result` object of the linting process and can modify it.
+
+```js
+// post-processor.js
+export default function myProcessor() {
+  return {
+    name: "remap-location",
+
+    /** @type {(result: LintResult) => void} */
```
ybiquitous, line 394:
```suggestion
// my-processor.js

/** @type {import("stylelint").Processor} */
export default function myProcessor() {
  return {
    name: "remap-location",

```

docs/user-guide/configure.md
```diff
@@ -375,6 +375,39 @@ Patterns are applied against the file path relative to the directory of the conf
...
+          warning.endLine === undefined ? warning.line : warning.line + 5;
+      });
+    }
+  };
+}
+```
+
+```json
+{
+  "processors": ["path/to/post-processor.js"]
```
ybiquitous, line 407:
```suggestion
  "processors": ["path/to/my-processor.js"]
```

ybiquitous (commented):
[suggest] How about improving the code example to make it closer to real a use-case?

lib/__tests__/fixtures/postprocess-update-text.mjs
```diff
@@ -0,0 +1,33 @@
+export default {
+	name: 'update-rule-text',
+
+	/** @type {(result: LintResult) => void} */
+	postprocess(result) {
+		const updatedWarnings = result.warnings.map((warning) => {
+			let updatedWarning = { ...warning };
+
+			// Find the rule that the warning is on and update the warning text to include it
+			result._postcssResult.root.walk((node) => {
```
ybiquitous, line 10: [suggest] I don't want to encourage using the internal API (`_postcssResult`). If it's useful, how about adding an argument for `PostCSS.Root`?

```js
postprocess(result, root) {
  // result is stylelint.LinterResult
  // root is postcss.Root
}
```

lib/augmentConfig.mjs
```diff
@@ -362,6 +389,34 @@ async function addPluginFunctions(config, { quietDeprecationWarnings }) {
...
+			processorImport = await dynamicImport(processorLookup);
+			processorImport = processorImport.default || processorImport;
+		} else {
+			processorImport = processorLookup;
+		}
+
+		return typeof processorImport === 'function' ? await processorImport() : processorImport;
+	});
+
+	config.processors = await Promise.all(processors);
```
ybiquitous, line 415: [suggest] I want to keep `config.processors` as an original value to make debugging easier. So, what about saving processor functions in an internal state like plugins?

https://github.com/stylelint/stylelint/blob/d3351244957e76945241e4b091d0e745b35835ab/lib/augmentConfig.mjs#L387

E.g.

```js
config._processorFunctions = await Promise.all(processors);
```

lib/createReturnedResults.mjs
```diff
@@ -0,0 +1,33 @@
...
+ * @param {import('stylelint').CssSyntaxError} [cssSyntaxError]
+ * @return {Promise<import('stylelint').LintResult>}
+ */
+export default async function createStylelintResult(
+	stylelint,
+	postcssResult,
+	filePath,
+	cssSyntaxError,
+) {
+	let stylelintResult = createPartialStylelintResult(postcssResult, cssSyntaxError);
```
ybiquitous, line 17: [suggest] I want to avoid one more nest for the experimental feature. Does it make sense to modify `createPartialStylelintResult()` instead of adding the `createStylelintResult()` module?

types/stylelint/index.d.ts
```diff
@@ -194,6 +197,14 @@ declare namespace stylelint {
 	/** @internal */
 	export type CustomSyntax = string | PostCSS.Syntax;
 
+	/** @internal */
+	export type PostProcessFunction = (result: LintResult) => void;
+
+	export type Processor = {
+		name: string;
+		postprocess: PostProcessFunction;
+	};
```
ybiquitous, line 206: [suggest] I have a few suggestions about the type:

- Fix `Processor` to a function type
- Add a warning comment around the experiment
- Inline the internal type

```suggestion
	/**
     * WARNING: This is an experimental feature. The API may change in the future.
     */
	export type Processor = () => {
		name: string;
		postprocess: (result: LintResult) => void;
	};
```

ybiquitous (changes requested):

lib/__tests__/fixtures/postprocess-update-text.mjs
YuanboXue-Amber: Yes sounds good. This is the updated Processor type:
```js
        export type Processor = () => {
                name: string;
                postprocess: (result: LintResult, root?: PostCSS.Root) => void;
        };
```
https://github.com/YuanboXue-Amber/stylelint/blob/985bdf1a7f7cd7999867f97153ca4cc440dbb643/types/stylelint/index.d.ts#L206-L209
I updated this test accordingly.

lib/createReturnedResults.mjs
YuanboXue-Amber: Agree. I'm keeping the original createPartialStylelintResult, and adding a new function underneath:
```js
  const stylelintResult = createPartialStylelintResult(postcssResult);
  await postProcessStylelintResult(stylelint, stylelintResult, absoluteFilepath);
```
https://github.com/YuanboXue-Amber/stylelint/blob/985bdf1a7f7cd7999867f97153ca4cc440dbb643/lib/standalone.mjs#L250-L252

lib/augmentConfig.mjs
YuanboXue-Amber: Good point on keeping processors as original value! 
I added:
```js
                _processorFunctions?: {
                        [processorName: string]: Omit<ReturnType<Processor>, 'name'>;
                }; // Internal use only
```
https://github.com/YuanboXue-Amber/stylelint/blob/985bdf1a7f7cd7999867f97153ca4cc440dbb643/types/stylelint/index.d.ts#L121-L123

lib/augmentConfig.mjs
```diff
@@ -362,6 +389,52 @@ async function addPluginFunctions(config, { quietDeprecationWarnings }) {
...
+
+	const processorPromises = config.processors.map(async (processorLookup) => {
+		let processorImport;
+
+		if (typeof processorLookup === 'string') {
+			processorImport = await dynamicImport(processorLookup);
+			processorImport = processorImport.default || processorImport;
+		} else {
+			processorImport = processorLookup;
+		}
```
ybiquitous, line 410: [suggest] Since `config.processors` should be an array of string, how about simplifying the code like this?

```suggestion
		let processor = await dynamicImport(processorLookup);

		processor = processor.default ?? processor;

		if (!isFunction(processor)) {
			throw configurationError(`The processor "${processorLookup}" must be a function`);
		}
```

`isFunction` is available by importing `utils/validateTypes.mjs` as below:

```js
import { isFunction } from './utils/validateTypes.mjs';
```

lib/augmentConfig.mjs
```diff
@@ -362,6 +389,52 @@ async function addPluginFunctions(config, { quietDeprecationWarnings }) {
...
+
+		const { name, ...functions } =
+			(typeof processorImport === 'function' ? await processorImport() : processorImport) ?? {};
```
ybiquitous, line 413: [suggest] How about simplify the code thanks to `isFunction()` type guard?

```suggestion
		const { name, postprocess } = processor();
```

lib/augmentConfig.mjs
```diff
@@ -362,6 +389,52 @@ async function addPluginFunctions(config, { quietDeprecationWarnings }) {
...
+
+		if (!name) {
```
ybiquitous, line 415: [suggest] How about checking the `name` property?

```suggestion
		if (!isString(name) || !name) {
```

`isString` can be imported by:

```js
import { isString } from './utils/validateTypes.mjs';
```

lib/augmentConfig.mjs
```diff
@@ -362,6 +389,52 @@ async function addPluginFunctions(config, { quietDeprecationWarnings }) {
...
+			throw configurationError(
+				`stylelint requires processors to expose a name. The processor "${processorLookup}" is not doing this, so will not work with stylelint. Please file an issue with the processor.`,
+			);
+		}
+
+		return { name, functions };
```
ybiquitous, line 421:
```suggestion
		if (!isFunction(postprocess)) {
			throw configurationError(`The processor "${processorLookup}" must return an object with the "postprocess" property`);
		}

		return { name, postprocess };
```

types/stylelint/index.d.ts
```diff
@@ -115,6 +117,10 @@ declare namespace stylelint {
 		configurationComment?: string;
 		overrides?: ConfigOverride[];
 		customSyntax?: CustomSyntax;
+		processors?: ConfigProcessors;
+		_processorFunctions?: {
+			[processorName: string]: Omit<ReturnType<Processor>, 'name'>;
+		}; // Internal use only
```
ybiquitous, line 123: [suggest]

- How about using `Map` instead of `Object` because `Map` has a more explicit property order than `Object`? ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#objects_vs._maps))
- How about using `@internal` as other internal types?

```suggestion
		/** @internal */
		_processorFunctions?: Map<string, ReturnType<Processor>['postprocess']>;
```

YuanboXue-Amber: Sounds good :) Yep agree keeping the order is better

YuanboXue-Amber:
> Great! The implementation is overall good 👍🏼
> 
> Can you take a look at my some suggestions?

Thank you :) I addressed all comments. Appreciate another round of review 🙏

ybiquitous (approved):
Thank you! LGTM 👍🏼

YuanboXue-Amber:
@ybiquitous thank you and could you merge this? I don't have access. Also when will be the next release 🛳️?

ybiquitous:
Since there are no concerns or objections, I'm merging this PR. And I'll release the next minor version soon.

@YuanboXue-Amber Again, thank you for the contribution!

YuanboXue-Amber:
Thank you @ybiquitous. Looking forward!

PR #7571, opened 2024-03-24
Fix `selector-max-id` end positions

romainmenke:

Follows the same patterns as : https://github.com/stylelint/stylelint/pull/7518

No, it's self-explanatory.


lib/rules/selector-max-id/__tests__/index.mjs
```diff
@@ -236,11 +236,11 @@ testRule({
...
-			message: messages.expected('#foo:hover > #bar #baz', 2),
+			message: messages.expected('&:hover > #bar #baz', 2),
 			line: 1,
 			column: 8,
 			endLine: 1,
-			endColumn: 30,
+			endColumn: 27,
 		},
 		{
 			code: '#foo #bar #baz { @media print { #bar {} } }',
```
Mouvedia, line 246: question

Are you sure this is valid CSS?
i.e. `@media` can't be nested in that case
I suppose it's SASS.

> The main extensions compared to level 2 are allowing nesting of certain at-rules inside [@media](https://www.w3.org/TR/css-conditional-3/#at-ruledef-media)…

lib/rules/selector-max-id/index.mjs
```diff
@@ -66,15 +66,17 @@ const rule = (primary, secondaryOptions) => {
 			}, 0);
 
 			if (selectorNode.type !== 'root' && selectorNode.type !== 'pseudo' && count > primary) {
```
Mouvedia, line 68: minor
perf

You can put `if (selectorNode.type === 'root' || selectorNode.type === 'pseudo') return;` before l53 so that you don't call the `reduce` in those cases, ending up with

```js
if (count > primary)
```

lib/rules/selector-max-id/index.mjs
```diff
@@ -106,13 +108,13 @@ const rule = (primary, secondaryOptions) => {
...
 			}
 
-			for (const selector of ruleNode.selectors) {
-				for (const resolvedSelector of resolvedNestedSelector(selector, ruleNode)) {
-					parseSelector(resolvedSelector, result, ruleNode, (container) =>
-						checkSelector(container, ruleNode),
-					);
-				}
-			}
+			if (!isStandardSyntaxSelector(ruleNode.selector)) return;
```
Mouvedia, line 111: nit
style

either both `return;` or both `{ return; }`

lib/rules/selector-max-id/__tests__/index.mjs
romainmenke: This is certainly valid CSS :)

romainmenke: The CSS nesting spec can be found here : https://drafts.csswg.org/css-nesting/

lib/rules/selector-max-id/index.mjs
romainmenke: Nice find!

Would be nice to apply this everywhere the same pattern exists.

romainmenke: I tried this and we can't make this change because `checkSelector` is recursive.

lib/rules/selector-max-id/index.mjs
romainmenke: Good catch, thank you!

ybiquitous (approved):
@romainmenke Thanks for the pull request. LGTM 👍🏼

PR #7585, opened 2024-03-30
Fix `declaration-block-no-shorthand-property-overrides` false negatives for `border`

### LINKED ISSUE #7506: Fix `declaration-block-no-shorthand-property-overrides` false negatives for `border`

fpetrakov:

Closes #7506

No, it's self-explanatory.


lib/rules/declaration-block-no-shorthand-property-overrides/index.mjs
```diff
@@ -37,9 +37,9 @@ const rule = (primary) => {
 					longhandSubPropertiesOfShorthandProperties
 				).get(unprefixedProp);
 
-				if (!overrideables) {
-					declarations.set(prop.toLowerCase(), prop);
+				declarations.set(prop.toLowerCase(), prop);
```
Mouvedia, line 40: nit

This can moved after l44.

lib/rules/declaration-block-no-shorthand-property-overrides/__tests__/index.mjs
```diff
@@ -47,6 +47,46 @@ testRule({
...
+			endColumn: 40,
+		},
+		{
+			code: 'a { border-block-start: 1px solid; border: 20px dashed black; }',
+			message: messages.rejected('border', 'border-block-start'),
+			line: 1,
+			column: 36,
+			endLine: 1,
+			endColumn: 42,
+		},
```
Mouvedia, line 73: These are incorrect and have to be removed.
You could add them as passing if you want to detect regressions though.

lib/rules/declaration-block-no-shorthand-property-overrides/index.mjs
```diff
@@ -37,9 +37,9 @@ const rule = (primary) => {
 					longhandSubPropertiesOfShorthandProperties
 				).get(unprefixedProp);
 
-				if (!overrideables) {
-					declarations.set(prop.toLowerCase(), prop);
```
Mouvedia, line 41: question
medium

Shouldn't this being in this `if` block have produced a bug?
i.e. a test is probably missing to cover that case or the setter was useless

Mouvedia (changes requested):

lib/reference/properties.mjs
```diff
@@ -58,6 +58,11 @@ export const longhandSubPropertiesOfShorthandProperties = new Map([
 			'border-right-color',
 			'border-bottom-color',
 			'border-left-color',
+			'border-color',
+			'border-style',
+			'border-block',
+			'border-block-end',
+			'border-block-start',
```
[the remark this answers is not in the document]
ybiquitous: Right. This `longhandSubPropertiesOfShorthandProperties` is exposed under the `reference` type, so we need to keep its correctness.

https://github.com/stylelint/stylelint/blob/a5965a09b557fd4a5420a6b48f1c1353ecf83aa2/types/stylelint/index.d.ts#L826-L831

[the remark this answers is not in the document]
ybiquitous: So, I guess some specific logic for this rule is necessary.

---

Also, supporting `border-image` seems reasonable to me because the [CSS spec](https://drafts.csswg.org/css-backgrounds/#propdef-border) says:

> The [border](https://drafts.csswg.org/css-backgrounds/#propdef-border) shorthand also resets [border-image](https://drafts.csswg.org/css-backgrounds/#propdef-border-image) to its initial value.

E.g.

```css
a {
  border-image: url('...'); /* this is ignored by border */
  border: 1px solid black; /* here, border-image is reset */
}
```

.changeset/mighty-cows-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
```
ybiquitous, line 2: [note] This fix will increase Stylelint errors, so "minor" is correct according to our versioning policy.

https://github.com/stylelint/stylelint/blob/a5965a09b557fd4a5420a6b48f1c1353ecf83aa2/docs/about/semantic-versioning.md?plain=1#L17-L21

lib/rules/declaration-block-no-shorthand-property-overrides/index.mjs
ybiquitous: I think this cannot be moved after L44. The move breaks some test cases.

lib/reference/properties.mjs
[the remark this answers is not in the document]
ybiquitous: > So, I guess some specific logic for this rule is necessary.

Sorry, this comment was inaccurate. I mean that the following 3 properties addition makes sense:

```js
'border-width',
'border-style',
'border-color',
```

This order reflects the `border` shorthand's [values order](https://developer.mozilla.org/en-US/docs/Web/CSS/border#values).

But `border-image` cannot be added here because it's for resetting a value, unlike other properties. If we'd like to support `border-image`, we should write custom logic in `declaration-block-no-shorthand-property-overrides/index.mjs`.

lib/rules/declaration-block-no-shorthand-property-overrides/index.mjs
ybiquitous: I feel like this logic change makes sense because test cases other than `border-block*` properties pass.

lib/reference/properties.mjs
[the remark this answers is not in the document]
ybiquitous: If supporting `border-image` is difficult, I think only `border-width`, `border-style`, and `border-color` are enough for this PR. 👌🏼

lib/rules/declaration-block-no-shorthand-property-overrides/index.mjs
Mouvedia: If I understand the change, a new passing test is needed or it needs to be kept in the `if`.
I gotta checkout your branch to check.

Mouvedia (approved):
LGTM

lib/reference/properties.mjs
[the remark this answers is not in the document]
Mouvedia: > I think only border-width, border-style, and border-color are enough for this PR.

👍 

> The border shorthand also resets border-image to its initial value.

@ybiquitous can you create a separate issue?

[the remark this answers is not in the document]
ybiquitous: Sure. Just opened #7586.

ybiquitous (approved):
@fpetrakov LGTM. Thanks a lot for actively fixing bugs! 😊

fpetrakov:
@ybiquitous thanks for the kindness and actively reviewing my PRs!

PR #7594, opened 2024-04-06
Add `no-unknown-custom-media`

### LINKED ISSUE #6362: Add no-unknown-custom-media

I'd like to disallow unknown custom media queries. For example:

```css
@custom-media --foo (max-width: 30em);
@media (--bar) {}
```

A new rule to catch this. Like https://github.com/stylelint/stylelint/issues/6361, it's currently [available as a plugin](https://github.com/csstools/stylelint-media-use-custom-media) but it [(perhaps) meets our criteria](https://stylelint.io/user-guide/rules/about) to be a built-in rule. Unlike https://github.com/stylelint/stylelint/issues/6361, it's for [a draft spec](https://drafts.csswg.org/mediaqueries-5/#custom-mq). However, we already have a custom media queries rule ([custom-media-pattern](https://stylelint.io/user-guide/rules/list/custom-media-pattern)) and both [PostCSS Preset Env](https://preset-env.cssdb.org/features/#custom-media-queries) and [Lightening CSS](https://github.com/parcel-bundler/lightningcss#features) offer syntax lowering for it.

Interestingly, the plugin also enforces using custom media queries through its [`always` and `always-known` primary options](https://github.com/csstools/stylelint-media-use-custom-media#options). Shall we take the same approach or create two rules?:

- one that checks for unknown custom media queries 
- and another that ensures matching queries use the corresponding name 

I'm leaning towards the latter.

- Name: `no-unknown-custom-media`
- Primary option: `true`
- Secondary options: `importForm` & `resolver`?
- Autofixable: No
- Message: - `Unexpected unknown custom media`
- Description: "Disallow unknown custom media queries."
- Extended description: "This rule considers custom media queries defined within the same source to be known. Use the `importForm` secondary option to specify more sources."
- Section: "Avoid errors" -> "General"

fpetrakov:

Closes #6362

No, it's self-explanatory.


lib/rules/no-unknown-custom-media/index.mjs
```diff
@@ -0,0 +1,84 @@
...
+				return;
+			}
+
+			if (atRuleName === 'custom-media') {
+				const customMediaQuery = parseCustomMediaQuery(atRule);
+				const customMediaQueryName = customMediaQuery && customMediaQuery.getName();
+
+				if (!customMediaQueryName) return;
+
+				declaredCustomMediaQueries.add(customMediaQueryName);
```
ybiquitous, line 42: [suggest] How about reducing one variable like this?

```suggestion
				const customMediaQuery = parseCustomMediaQuery(atRule);

				if (!customMediaQuery) return;

				declaredCustomMediaQueries.add(customMediaQuery.getName());
```

lib/rules/no-unknown-custom-media/index.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+			const mediaQueryList = parseMediaQuery(atRule);
+
+			mediaQueryList.forEach((mediaQuery) => {
+				if (isMediaQueryInvalid(mediaQuery)) return;
+
+				mediaQuery.walk(({ node }) => {
+					if (isMediaFeature(node) && isCustomMediaQuery(node.getName())) {
+						usedCustomMediaQueryName = node.getName();
+					}
```
ybiquitous, line 57: [suggest] How about fixing multiple lint problems and positions? E.g.

```suggestion
					if (!isMediaFeature(node)) return;

					const [, name, start, end] = node.getNameToken();

					if (!isCustomMediaQuery(name)) return;

					if (declaredCustomMediaQueries.has(name)) return;

					const baseIndex = atRuleParamIndex(atRule);
					const index = baseIndex + start;
					const endIndex = baseIndex + end + 1;

					report({ /*...*/ });
```

lib/rules/no-unknown-custom-media/index.mjs
```diff
@@ -0,0 +1,84 @@
...
+			if (declaredCustomMediaQueries.has(usedCustomMediaQueryName)) {
+				return;
+			}
+
+			const index = atRuleParamIndex(atRule);
+			const endIndex = index + usedCustomMediaQueryName.length + 1;
+
+			report({
+				message: messages.expected,
+				messageArgs: [usedCustomMediaQueryName, primary],
```
ybiquitous, line 70: [suggest] Let's remove the unnecessary argument:

```suggestion
				messageArgs: [usedCustomMediaQueryName],
```

types/stylelint/index.d.ts
```diff
@@ -329,6 +329,7 @@ declare namespace stylelint {
 		readonly 'no-irregular-whitespace': Promise<Rule>;
 		readonly 'no-unknown-animations': Promise<Rule>;
 		readonly 'no-unknown-custom-properties': Promise<Rule>;
+		readonly 'no-unknown-custom-media': Promise<Rule>;
```
ybiquitous, line 332: [suggest] Let's reorder alphabetically:

```suggestion
		readonly 'no-unknown-custom-media': Promise<Rule>;
		readonly 'no-unknown-custom-properties': Promise<Rule>;
```

ybiquitous (changes requested):
@fpetrakov Thank you for creating this pull request! The direction is correct, but I've suggested some bugfixes and refactorings. Can you take a look at them?

fpetrakov:
@ybiquitous thanks a lot for the review, learned a bunch of things

lib/rules/no-unknown-custom-media/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+@custom-media --sm (min-width: 40rem);
+
+@media (--sm) {}
+
+@media (--sm), (min-width: 40rem) {}
```
Mouvedia, line 41: minor

This rule makes no sense: you have the same media query twice.
Maybe change it to `@media (--sm), (man-height: 40rem) {}`?

lib/rules/no-unknown-custom-media/README.md
```diff
@@ -0,0 +1,42 @@
...
+@custom-media --sm (min-width: 40rem);
+/**             ↑
+*   This custom media query name */
+
+@media (--sm) {}
+/**      ↑
+*   And this one */
+```
+
+This rule considers custom media queries defined within the same source to be known.
```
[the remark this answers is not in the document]
ybiquitous: +1 for emphasis

lib/rules/no-unknown-custom-media/__tests__/index.mjs
```diff
@@ -0,0 +1,92 @@
...
+		{
+			code: '@custom-media --foo (max-width: 30em); @media screen and (--bar) {};',
+			message: messages.expected('--bar'),
+			line: 1,
+			column: 59,
+			endLine: 1,
+			endColumn: 64,
+		},
+		{
+			code: '@custom-media --foo (max-width: 30em); @media (--bar) {};',
```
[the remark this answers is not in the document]
ybiquitous: I can't find a description of "hoisting" in <https://www.w3.org/TR/mediaqueries-5/>, but I'm fine with adding such a reject case because it should not request any logic change.

Mouvedia:
There's something interesting about this rule addition.
AFAIK no browsers currently support `@custom-media`.
Which means the rule is being added on the premise that you are using some kind of tooling.
Maybe we could link to those in the rule README?

lib/rules/no-unknown-custom-media/__tests__/index.mjs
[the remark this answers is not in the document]
Mouvedia: > I suppose that custom media rules must be declared before their usage.

see https://github.com/parcel-bundler/lightningcss/issues/49#issuecomment-1013758907

lib/rules/no-unknown-custom-media/README.md
```diff
@@ -0,0 +1,49 @@
...
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+@custom-media --sm (min-width: 40rem);
+
+@media (--sm), (max-height: 40rem) {}
+```
+
+> [!NOTE] `@custom-media` can be used before its defined
```
Mouvedia, line 42:
```suggestion
> [!NOTE]
> The `@custom-media` rule can be used before its declaration
```

lib/rules/no-unknown-custom-media/__tests__/index.mjs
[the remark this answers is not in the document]
Mouvedia: see https://github.com/stylelint/stylelint/pull/7594#pullrequestreview-1988807526

Mouvedia (approved):
LGTM

ybiquitous:
For https://github.com/stylelint/stylelint/pull/7594#issuecomment-2043069915, I don't think adding some links to transpiling tools (e.g. postcss plugins or parcel) is necessary, even if there is insufficient support in the current browsers. We already have another rule for `@custom-media`, so we must keep such links up-to-date across documents.
 
See https://github.com/stylelint/stylelint/blob/08910de8bb3775b06efed7f7990e81d47548bace/lib/rules/custom-media-pattern/README.md

ybiquitous (approved):

PR #7603, opened 2024-04-10
Refactor to silence PostCSS stderr warning 

Mouvedia:

relates to #7518

```sh
  console.warn
    Without `from` option PostCSS could generate wrong source map and will not find Browserslist config. Set it to CSS file path or to `undefined` to prevent this warning.

      at warnOnce (node_modules/postcss/lib/warn-once.js:11:13)
      at NoWorkResult.then (node_modules/postcss/lib/no-work-result.js:67:9)
```


ybiquitous (approved):
Nice catch 👍🏼

PR #7606, opened 2024-04-12
Fix `declaration-block-no-shorthand-property-overrides` false negatives for `font` and `border`

Mouvedia:

Closes #7585

`border` resets `border-image` which itself resets its own longhands
see https://www.w3.org/TR/css-backgrounds-3/#border-image


lib/reference/properties.mjs
```diff
@@ -249,6 +283,14 @@ export const longhandSubPropertiesOfShorthandProperties = new Map([
 			'font-size',
 			'line-height',
 			'font-family',
+			/**
+			 * reset explicitly: normal | small-caps
+			 * reset implicitly: all-small-caps | petite-caps | all-petite-caps | unicase | titling-caps
+			 * i.e. either way it will be reset
+			 * @todo add font-variant shorthand to longhandSubPropertiesOfShorthandProperties
```
Mouvedia, line 290: `font-variant` addition will be covered in another PR.

lib/rules/declaration-block-no-shorthand-property-overrides/index.mjs
```diff
@@ -32,19 +36,21 @@ const rule = (primary) => {
...
-
-				const overrideables = /** @type {Map<string, Set<string>>} */ (
+				const subProperties = /** @type {Map<string, Set<string>>} */ (
 					longhandSubPropertiesOfShorthandProperties
 				).get(unprefixedProp);
+				const resettables = /** @type {Map<string, Set<string>>} */ (
+					shorthandToResetToInitialProperty
+				).get(unprefixedProp);
+				// see microsoft/TypeScript#57228
+				const union = new Set([...(subProperties ?? []), ...(resettables ?? [])]);
```
Mouvedia, line 46: `union` from https://github.com/tc39/proposal-set-methods cannot be used yet

types/stylelint/index.d.ts
```diff
@@ -681,6 +681,10 @@ declare namespace stylelint {
 		| 'text-emphasis'
 		| 'transition';
 
+	type ImplicitResetters = Extract<ShorthandProperties, 'border' | 'flex' | 'font'>;
```
Mouvedia, line 684: `flex` used to be an implicit resetter.
I chose to not support it for now because it's unclear what's preferable as a default.
i.e. a separate issue needs to be created

lib/reference/properties.mjs
```diff
@@ -11,6 +11,40 @@ export const acceptCustomIdentsProperties = new Set([
 	'list-style-type',
 ]);
 
+/** @type {import('stylelint').ShorthandToResetToInitialProperty} */
```
Mouvedia, line 14: WIP
It seems like I gotta edit `type-test.ts`.

Mouvedia, line 14: @ybiquitous do you know how to introduce a `type` and reference it in the same PR?

lib/rules/declaration-block-no-shorthand-property-overrides/index.mjs
ybiquitous: [suggest] Let's use `utils/uniteSets` instead:

```suggestion
				const union = uniteSets(subProperties ?? [], resettables ?? []);
```

types/stylelint/index.d.ts
ybiquitous: [suggest] It seems that adding the `ImplicitResetters` and `ShorthandToResetToInitialProperty` types is unnecessary for now because they're not required to be published, not `LonghandSubPropertiesOfShorthandProperties`.

So, how about removing the types?

ybiquitous (changes requested):
@Mouvedia Thanks for the pull request, and sorry for the late response. I have been very busy since last week.

I just left some suggestions. Can you take a look?

lib/rules/declaration-block-no-shorthand-property-overrides/index.mjs
```diff
@@ -32,19 +37,21 @@ const rule = (primary) => {
...
 				const unprefixedProp = vendor.unprefixed(prop).toLowerCase();
 				const prefix = vendor.prefix(prop).toLowerCase();
-
-				const overrideables = /** @type {Map<string, Set<string>>} */ (
+				const subProperties = /** @type {Map<string, Set<string>>} */ (
 					longhandSubPropertiesOfShorthandProperties
 				).get(unprefixedProp);
+				const resettables = /** @type {Map<string, Set<string>>} */ (
+					shorthandToResetToInitialProperty
+				).get(unprefixedProp);
```
ybiquitous, line 45: [suggest] Casting is no longer necessary. Let's simplify the code:

```suggestion
				const resettables = shorthandToResetToInitialProperty.get(unprefixedProp);
```

lib/rules/declaration-block-no-shorthand-property-overrides/index.mjs
```diff
@@ -32,19 +37,21 @@ const rule = (primary) => {
...
+				// see microsoft/TypeScript#57228
```
ybiquitous, line 46: [suggest] This link comment is outdated, so let's remove it:

```suggestion
```

ybiquitous (changes requested):
Thanks for the quick fix. Can you address additional reviews?

ybiquitous (approved):
Thanks. LGTM 👍🏼

.changeset/silent-flies-float.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Fixed: `declaration-block-no-shorthand-property-overrides` false negatives for `font` and `border`
```
Mouvedia, line 5: @ybiquitous should I

```suggestion
Fixed: `declaration-block-no-shorthand-property-overrides` false negatives for `font`
```

and add myself to https://github.com/stylelint/stylelint/pull/7582/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR8?

ybiquitous: Why removing `border`?

Mouvedia: I was thinking of replacing this line

```md
- Fixed: `declaration-block-no-shorthand-property-overrides` false negatives for `border` ([#7585](https://github.com/stylelint/stylelint/pull/7585)) ([@fpetrakov](https://github.com/fpetrakov)).
```

by

```md
- Fixed: `declaration-block-no-shorthand-property-overrides` false negatives for `border` ([#7585](https://github.com/stylelint/stylelint/pull/7585) & [#7606](https://github.com/stylelint/stylelint/pull/7606)) ([@fpetrakov](https://github.com/fpetrakov) & [@Mouvedia](https://github.com/Mouvedia)).
```

lib/utils/uniteSets.cjs
```diff
@@ -6,6 +6,7 @@
  * Unite two or more sets
  *
  * @param {Iterable<string>[]} args
+ * @see {@link https://github.com/microsoft/TypeScript/issues/57228|GitHub}
  */
 function uniteSets(...args) {
```
Mouvedia, line 11: @ybiquitous

Should we add https://github.com/es-shims/Set.prototype.union?

ybiquitous: No, I don't think we need to add the polyfill package for this tiny utility function.
The package has many dependencies, which increase Stylelint's download size.

See https://pkg-size.dev/set.prototype.union
image

Issue #7608, opened 2024-04-13
Consider adding meta.schema


TKDev7:
It would be equivalent to [ESLint’s meta.schema](https://eslint.org/docs/latest/extend/custom-rules#options-schemas). This can be used for generating plugin/rule documentation websites and typegens.

ybiquitous:
Sounds good 👍🏼 

Maybe one downside is increasing a dependency that processes the JSON schema, right?

ybiquitous:
Great. `ajv` is already included in our runtime dependency tree (`stylelint` depends on `table`).

```sh-session
$ npm ls ajv
stylelint@16.3.1 /Users/masafumi.koba/git/stylelint/stylelint
├─┬ eslint@8.57.0
│ ├─┬ @eslint/eslintrc@2.1.4
│ │ └── ajv@6.12.6 deduped
│ └── ajv@6.12.6
└─┬ table@6.8.2
  └── ajv@8.12.0
```

ybiquitous:
@MrHBS If you're interested, could you please show more specific idea to implement this feature so that others could image it easily?

ybiquitous:
@MrHBS Thanks for the helpful info.

@ota-meshi Do you have any ideas to bring this feature from ESLint?

Mouvedia:
Here's another use case for autocompletion: [knowing the message arguments' names and positions](https://github.com/stylelint/stylelint/pull/6312/files#r968253349).

Mouvedia:
I am working on a pre-requisite for `schema` that will expedite the conversion to JSON schema.

Mouvedia:
@MrHBS

I have a few questions.

1. Would `schema` be required if the rule is not `CoreRule<true>`?
i.e. doesn't accept `true` or `null` as a primary and global secondary options

> [Rules with options must specify a meta.schema property…](https://eslint.org/docs/latest/extend/custom-rules#options-schemas)

2. If introduced, would it sunset the `...optionDescriptions` argument of the `validateOptions` function?
i.e. `possible` and `optional` would be covered by the schema

3. Would introducing a way to generate a schema for the current config be sufficient?
e.g.
- `stylelint schema > foo.jtd.json` defaults to JSON Type Definition (default `--type` is `"RFC8927"`)
- `stylelint schema -c ../bar/.stylelintrc > foo.jtd.json` doesn't use the default config
- `stylelint schema --type="2020-12" > foo.schema.json` sets the format to JSON Schema version 2020-12

---

[...]

Issue #7510, opened 2024-02-02
Do we care about features behind browser flags?


Mouvedia:
### What is the problem you're trying to solve?

Ill give a concrete example: `<selectmenu>` is not part of [html-tags](https://github.com/sindresorhus/html-tags/blob/main/html-tags.json) which is used by `selector-type-no-unknown` rule.
It already has been renamed `<selectlist>` and can have a `<popup>` child.

Should we have policy that states that we wait for the experimentation to end before we add support for such features ?
If so we would have to give some pointers on what we mean by the _experimentation phase_.

### What solution would you like to see?

https://chromestatus.com/feature/5737365999976448
https://hidde.blog/custom-select-with-selectlist/
https://bugs.chromium.org/p/chromium/issues/detail?id=1121840

ybiquitous:
> Should we have a policy that states that we have to wait for the experimentation to end before we add support for such features?

It seems ideal to me, but indeed, I guess it's difficult because we have to care about dependency change details. At the same time, that work doesn't seem to be worth it.

Mouvedia:
@ybiquitous are you advocating for a case by case basis?

ybiquitous:
No. My concern is how we can achieve this goal. It would be great if there was a great solution.

Mouvedia:
Let's review the other examples that I have added.

- 4 examples of pseudo-classes that are missing

The additions can be argued either way. If it's not resolved here I will eventually have to open an issue.

- `text-decoration-width` requires just one `if`
- `scroll()` has already been added

If we had a policy, the _ready to implement_ status could be added without much discussion.  
Something as simple as _if it has ever been implemented in any spec conformant browser it can be added_ would do.

ybiquitous:
Make sense. Then, do you assume such a new policy to document? For example, adding an item to "A rule must be:" etc.

https://github.com/stylelint/stylelint/blob/f44de2e414b9e70cba3477e50c67e41ad97b0375/docs/developer-guide/rules.md?plain=1#L11-L14

Mouvedia:
I don't know where it should be added in the documentation. It would mainly be used to help with the _ready to implement_ labelling. So I would expect to find it here: https://github.com/stylelint/stylelint/blob/main/docs/maintainer-guide/issues.md

My proposal is uncomplicated: if a browser supported it, it's OK to add. That's a completist dream.  
But requiring the experimental phase to be finished to be added would _also_ be acceptable.
[...]

PR #7611, opened 2024-04-14
Add `ignoreLonghands: []` to `declaration-block-no-redundant-longhand-properties`

### LINKED ISSUE #7232: Add `"ignoreLonghands": []` to `declaration-block-no-redundant-longhand-properties`
## use cases

- [`text-decoration`](#text-decoration)
- [`background`](#background)

## rationale

If a browser doesn't support the new arity for a given shorthand, it will drop it altogether.
For example if we were to introduce an enhancement for the current fixer of `text-decoration` by adding support to
`text-decoration-thickness`, for a lot of users of the rule it would introduce a hard to detect regression.  
They would have to rely on `"ignoreShorthands": ["text-decoration"]` which is arguably a downgrade compared to the status quo. Hence a new option is required to be more granular and cross-browser.

### rationale for new fix

https://github.com/stylelint/stylelint/pull/6580#discussion_r1095994177
#7060

### without option

```css
a {
  text-decoration-line: underline;
  text-decoration-style: solid;
  text-decoration-color: green;
  text-decoration-thickness: 1px;
}
```
becomes
```css
a {
  text-decoration: underline solid green 1px;
}
```

### with `"ignoreLonghands": ["text-decoration-thickness"]`

```css
a {
  text-decoration-line: underline;
  text-decoration-style: solid;
  text-decoration-color: green;
  text-decoration-thickness: 1px;
}
```
becomes
```css
a {
	text-decoration: underline solid green;
	text-decoration-thickness: 1px;
}
```

### references

https://bugs.webkit.org/show_bug.cgi?id=190774
https://drafts.csswg.org/css-text-decor-4/#text-decoration-property

### implementation details

[Firefox 69 supports this feature (preffed off) under the name `text-decoration-width`, behind the pref layout.css.text-decoration-width.enabled](https://bugzilla.mozilla.org/show_bug.cgi?id=1567282#c8).
[`text-decoration-skip` is not part of the shorthand](https://drafts.csswg.org/css-text-decor-4/#text-decoration-property) but will have to be eventually added in a separate PR. 

### browser support

https://caniuse.com/?search=text-decoration-thickness

### rationale

https://caniuse.com/background-img-opts

### example

```json
{
  "fix": true,
  "rules": {
    "declaration-block-no-redundant-longhand-properties": [
      true,
      { "ignoreLonghands": ["background-size", "background-origin", "background-clip"] }
    ],
  }
}
```

```css
a {
  background-image: none;
  background-position: 0% 0%;
  background-size: auto auto;
  background-repeat: repeat;
  background-origin: padding-box;
  background-clip: border-box;
  background-attachment: scroll;
  background-color: transparent;
}
```
becomes
```css
a {
  background: repeat scroll 0% 0% none transparent;
  background-size: auto auto;
  background-origin: padding-box;
  background-clip: border-box;
}
```

Mouvedia:

Closes #7232

To understand why a fix is being accompanied by an option check the discussion starting with https://github.com/stylelint/stylelint/issues/7232#issuecomment-1762654301

tl;dr: CSS Text Decoration Module Level 3 vs CSS Text Decoration Module Level 4 means that a granular option is necessary for the `text-decoration` fix not be considered a downgrade

### DIFF
--- a/.changeset/selfish-plums-dream.md
+++ b/.changeset/selfish-plums-dream.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Fixed: `declaration-block-no-redundant-longhand-properties` autofix for `text-decoration`
--- a/.changeset/sharp-buses-drop.md
+++ b/.changeset/sharp-buses-drop.md
@@ -2,4 +2,4 @@
 "stylelint": minor
 ---
 
-Added: `ignoreLonghands` option to `declaration-block-no-redundant-longhand-properties`
+Added: `ignoreLonghands: []` to `declaration-block-no-redundant-longhand-properties`
--- a/.changeset/sharp-buses-drop.md
+++ b/.changeset/sharp-buses-drop.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: `ignoreLonghands` option to `declaration-block-no-redundant-longhand-properties`
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
@@ -208,7 +208,7 @@ a {
 }
 ```
 
-The following pattern are _not_ considered problems:
+The following patterns are _not_ considered problems:
 
 <!-- prettier-ignore -->
 ```css
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
@@ -174,6 +174,60 @@ a {
 
 ## Optional secondary options
 
+### `ignoreLonghands: ["string"]`
+
+Given:
+
+<!-- prettier-ignore -->
+```json
+["text-decoration-thickness", "background-size", "background-origin", "background-clip"]
+```
+
+The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+a {
+  text-decoration-line: underline;
+  text-decoration-style: solid;
+  text-decoration-color: purple;
+}
+```
+
+<!-- prettier-ignore -->
+```css
+a {
+  background-repeat: repeat;
+  background-attachment: scroll;
+  background-position: 0% 0%;
+  background-color: transparent;
+  background-image: none;
+  background-size: contain;
+  background-origin: border-box;
+  background-clip: text;
+}
+```
+
+The following pattern are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+a {
+  text-decoration: underline solid purple;
+  text-decoration-thickness: 1px;
+}
+```
+
+<!-- prettier-ignore -->
+```css
+a {
+  background: none 0% 0% repeat scroll transparent;
+  background-size: contain;
+  background-origin: border-box;
+  background-clip: text;
+}
+```
+
 ### `ignoreShorthands: ["/regex/", /regex/, "string"]`
 
 Given:
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
@@ -59,6 +59,14 @@ testRule({
 			code: 'a { transition-delay: 500ms, 1s; transition-duration: 250ms,2s; transition-timing-function: ease-in-out; transition-property: inherit; }',
 			description: 'transition property contains basic keyword (inherit)',
 		},
+		{
+			code: 'a { text-decoration-line: underline; text-decoration-style: solid; text-decoration-color: purple; }',
+			description: 'missing text-decoration-thickness',
+		},
+		{
+			code: 'a { background-repeat: repeat; background-attachment: scroll; background-position: 0% 0%; background-color: transparent; background-image: none; background-origin: border-box; background-clip: text; }',
+			description: 'missing background-size',
+		},
 	],
 
 	reject: [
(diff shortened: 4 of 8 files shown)


lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
```diff
@@ -387,6 +395,50 @@ testRule({
...
+		{
+			code: 'a { text-decoration-line: underline; text-decoration-style: solid; text-decoration-color: purple; }',
+			fixed: 'a { text-decoration: underline solid purple; }',
+			description: 'CSS Text Decoration Module Level 3',
+			message: messages.expected('text-decoration'),
+		},
+		{
+			code: 'a { text-decoration-color: purple; text-decoration-thickness: 1px; text-decoration-style: solid; text-decoration-line: underline; text-decoration-width: 1px; }',
+			fixed:
+				'a { text-decoration: underline solid purple; text-decoration-thickness: 1px; text-decoration-width: 1px; }',
```
Mouvedia, line 432: I chose not to add an exception for `text-decoration-width` because #7510 doesn't have a resolution.

.changeset/sharp-buses-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: `ignoreLonghands` option to `declaration-block-no-redundant-longhand-properties`
```
ybiquitous, line 5: [suggest] Let's keep consistency in CHANGELOG:

```suggestion
Added: `ignoreLonghands: []` to `declaration-block-no-redundant-longhand-properties`
```

lib/rules/declaration-block-no-redundant-longhand-properties/README.md
```diff
@@ -174,6 +174,60 @@ a {
 
 ## Optional secondary options
 
+### `ignoreLonghands: ["string"]`
```
ybiquitous, line 177: [suggest] I feel supporting also regexes makes sense. Any thoughts?

```suggestion
### `ignoreLonghands: ["/regex/", /regex/, "string"]`
```

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
```diff
@@ -387,6 +395,50 @@ testRule({
...
+			ignoreLonghands: [
+				'background-size',
+				'background-origin',
+				'background-clip',
+				'text-decoration-thickness',
+			],
+		},
+	],
+	fix: true,
+
```
ybiquitous, line 421: [suggest] Can you add `accept` cases, too?

ybiquitous (changes requested):
Thanks for the pull request. Can you address my comments?

.changeset/selfish-plums-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
```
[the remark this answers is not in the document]
Mouvedia: > Typo?

Nop

> [a bug fix in a rule that results in Stylelint reporting more errors](https://stylelint.io/about/semantic-versioning/#minor-release)

i.e. without the option

```css
a {
  text-decoration-line: underline;
  text-decoration-style: solid;
  text-decoration-color: purple;
  text-decoration-thickness: 1px; /* this one will newly be reported */
}

File: lib/rules/declaration-block-no-redundant-longhand-properties/README.md (hunk 3 above)
Comment (Mouvedia) [reply lib/rules/declaration-block-no-redundant-longhand-properties/README.md:177]: I don't think so for 2 reasons:

- it will complicate the implementation
- I couldn't find a use case for regexes

File: lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs (hunk 4 above)
Comment (Mouvedia) [reply lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs:421]: I couldn't think of good ones.
i.e. the accept ones needs to help with potential regressions

Do you have a suggestion?

Comment (ybiquitous) [reply lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs:421]: I see. Let's keep it.

File: lib/rules/declaration-block-no-redundant-longhand-properties/README.md (hunk 3 above)
Comment (ybiquitous) [reply lib/rules/declaration-block-no-redundant-longhand-properties/README.md:177]: I see. Let's leave it as-is. We can revisit later if people request.

File: lib/rules/declaration-block-no-redundant-longhand-properties/index.mjs [hunk 6]
```diff
@@ -234,6 +235,8 @@ const rule = (primary, secondaryOptions, context) => {
 			}
 
 			for (const longhand of longhandProps) {
+				if (secondaryOptions?.ignoreLonghands?.includes(longhand)) continue;
```
ybiquitous, line 238: [suggest] How about defining a local variable outside the loop like this?

```js
		if (!validOptions) {
			return;
		}

		/** @type {string[]} */
		const ignoreLonghands = secondaryOptions?.ignoreLonghands ?? [];
```

lib/rules/declaration-block-no-redundant-longhand-properties/index.mjs
```diff
@@ -275,11 +278,17 @@ const rule = (primary, secondaryOptions, context) => {
...
 
-					const shorthandProps = longhandSubPropertiesOfShorthandProperties.get(shorthandProperty);
-					const prefixedShorthandData = Array.from(shorthandProps || [], (item) => prefix + item);
+					const shorthandProps = new Set(
+						longhandSubPropertiesOfShorthandProperties.get(shorthandProperty),
+					);
 
+					secondaryOptions?.ignoreLonghands?.forEach((/** @type {string} */ value) => {
+						if (shorthandProps.has(value)) shorthandProps.delete(value);
+					});
```
ybiquitous, line 287: [suggest] Extracting a local variable `ignoreLonghands`

```js
/** @type {string[]} */
const ignoreLonghands = secondaryOptions?.ignoreLonghands ?? [];
```

can simplify this code like this:

```suggestion
					ignoreLonghands.forEach((value) => shorthandProps.delete(value));
```

Just `.delete` seems enough (`.has` seems unneeded). Any thoughts?

lib/rules/declaration-block-no-redundant-longhand-properties/index.mjs
Mouvedia: I wish we didn't have to cast it.
Can you take a look at #7608?

ybiquitous: The idea of `meta.schema` sounds interesting, but casting is necessary for now, right?

ybiquitous (approved):
Thanks. LGTM 👍🏼

Mouvedia:
@ybiquitous @jeddy3 I would advise to add `"ignoreLonghands": ["text-decoration-thickness"]` to `stylelint-config-standard`.

ybiquitous:
> I would advise to add `"ignoreLonghands": ["text-decoration-thickness"]`

Hum, [this example on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration#examples) makes sense, right?

```css
.thick {
  text-decoration: solid underline purple 4px; /* includes `text-decoration-thickness: 4px` */
}
```

Mouvedia:
It does if you consider that 95%+ support on caniuse is enough. If not it would be better to add the option so that standard doesn't change.

ybiquitous:
> It does if you consider that 95%+ support on caniuse is enough.

I think the percentage is enough for now. Let's keep things simple. If people complain, then we can revisit it.

Mouvedia:
@ybiquitous 

```
error - 2024-04-18 12:59:34,223 -- Commit creating failed: {"detail":"Tokenless has reached GitHub rate limit. Please upload using a token: https://docs.codecov.com/docs/adding-the-codecov-token. Expected available in 136 seconds."}
Error: Codecov: Failed to properly create commit: The process '/home/runner/work/_actions/codecov/codecov-action/v4/dist/codecov' failed with exit code 1
```

ybiquitous:
Have you merge the commit fca3269efcabe1291746a6832cd197e688712bdd into this branch?

PR #7612, opened 2024-04-14
Fix `selector-type-no-unknown` false positives for experimental and deprecated HTML tags

Mouvedia:

check the "concrete example" in #7510

compare https://github.com/sindresorhus/html-tags/blob/main/html-tags.json and https://github.com/sindresorhus/html-tags/commit/d3b931ee8dc7c5bbbe20761ebbe4dad257cf39bc

- `selectlist`: https://github.com/openui/open-ui/issues/773#issuecomment-1664421419
- `listbox`: https://issues.chromium.org/issues/40146374#comment98
- `portal`: https://caniuse.com/mdn-html_elements_portal
- for the rest check https://developer.mozilla.org/en-US/docs/Web/HTML/Element/

I chose not to include the previous names of the tag names that I am introducing; we need to settle #7510 first.
e.g. not `selectmenu`

### DIFF
--- a/.changeset/flat-bananas-hug.md
+++ b/.changeset/flat-bananas-hug.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Fixed: `selector-type-no-unknown` false positives
--- a/.changeset/flat-bananas-hug.md
+++ b/.changeset/flat-bananas-hug.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `selector-type-no-unknown` false positives
+Fixed: `selector-type-no-unknown` false positives for experimental and deprecated HTML tags
--- a/.changeset/flat-bananas-hug.md
+++ b/.changeset/flat-bananas-hug.md
@@ -1,5 +1,5 @@
 ---
-"stylelint": minor
+"stylelint": patch
 ---
 
 Fixed: `selector-type-no-unknown` false positives
--- a/lib/reference/selectors.cjs
+++ b/lib/reference/selectors.cjs
@@ -10,6 +10,7 @@ const deprecatedHtmlTypeSelectors = new Set([
 	'applet',
 	'basefont',
 	'big',
+	'bgsound',
 	'blink',
 	'center',
 	'content',
@@ -17,14 +18,17 @@ const deprecatedHtmlTypeSelectors = new Set([
 	'font',
 	'frame',
 	'frameset',
-	'hgroup',
 	'isindex',
 	'keygen',
 	'listing',
 	'marquee',
+	'multicol',
+	'nextid',
 	'nobr',
 	'noembed',
+	'noframes',
 	'plaintext',
+	'param',
 	'spacer',
 	'strike',
 	'tt',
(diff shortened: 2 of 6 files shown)


lib/reference/selectors.cjs
```diff
@@ -10,21 +10,26 @@ const deprecatedHtmlTypeSelectors = new Set([
...
 	'big',
+	'bgsound',
 	'blink',
 	'center',
 	'content',
 	'dir',
 	'font',
 	'frame',
 	'frameset',
-	'hgroup',
```
Mouvedia, line 20: already part of html-tags package
used to be obsolete

lib/reference/selectors.cjs
```diff
@@ -10,21 +10,26 @@ const deprecatedHtmlTypeSelectors = new Set([
...
+	'image',
```
Mouvedia, line 21: was silently converted

lib/reference/selectors.mjs
```diff
@@ -7,21 +7,26 @@ const deprecatedHtmlTypeSelectors = new Set([
...
 	'big',
+	'bgsound',
 	'blink',
 	'center',
 	'content',
 	'dir',
 	'font',
 	'frame',
 	'frameset',
-	'hgroup',
```
Mouvedia, line 17: already part of html-tags package
used to be obsolete

lib/reference/selectors.mjs
```diff
@@ -7,21 +7,26 @@ const deprecatedHtmlTypeSelectors = new Set([
...
+	'image',
```
Mouvedia, line 18: was silently converted to ``

lib/utils/isCustomElement.cjs
```diff
@@ -31,10 +30,6 @@ function isCustomElement(selector) {
 		return false;
 	}
 
-	if (selectors.htmlTypeSelectors.has(selectorLowerCase)) {
-		return false;
-	}
-
```
Mouvedia, line 37: check l20
i.e. HTML tags don't include `-`

lib/utils/isCustomElement.mjs
```diff
@@ -28,10 +27,6 @@ export default function isCustomElement(selector) {
 		return false;
 	}
 
-	if (htmlTypeSelectors.has(selectorLowerCase)) {
-		return false;
-	}
```
Mouvedia, line 33: check l20 it's unnecessary
i.e. HTML tag names don't include `-`

.changeset/flat-bananas-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
```
ybiquitous, line 2: [ask] Is this PR a patch update?

```suggestion
"stylelint": patch
```

ybiquitous (commented):
@Mouvedia Thanks for the pull request. Overall LGTM, but I have one question.

.changeset/flat-bananas-hug.md
Mouvedia: stylelint will report more errors so it's a minor
see https://stylelint.io/about/semantic-versioning/#minor-release

ybiquitous: So, "false negatives" are correct?

ybiquitous: This PR seems to reduce false errors, for example: 🤔 

```css
fencedframe, listbox, portal, selectlist {}
```

image

Mouvedia: I don't know.
I will put `patch` to be consistent with previous false-positive PRs.

.changeset/flat-bananas-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-type-no-unknown` false positives
```
ybiquitous, line 5: [suggest] Just to be sure, does clarifying targets of this change make sense? E.g.

```suggestion
Fixed: `selector-type-no-unknown` false positives for deprecated HTML tags
```

ybiquitous (approved):
LGTM 👍🏼

PR #7625, opened 2024-04-18
Fix `declaration-block-no-redundant-longhand-properties` autofix for `border`

### LINKED ISSUE #7609: Fix `declaration-block-no-redundant-longhand-properties` autofix for `border`

```css
a {
	border-width: 7px;
	border-style: double;
	border-color: green;
}
```

```json
{
  "rules": {
  },
  "fix": true
}
```

https://stylelint.io/demo

.

```css
a {
	border: 7px double green;
}
```

### What actually happened?

.

add a custom resolver for `border`

Mouvedia:

Closes #7609

To avoid regressions we didn't remove

```
    "border-top-width",
    "border-bottom-width",
    "border-left-width",
    "border-right-width",
    "border-top-style",
    "border-bottom-style",
    "border-left-style",
    "border-right-style",
    "border-top-color",
    "border-bottom-color",
    "border-left-color",
    "border-right-color",
```

in #7585.
The error was introduced in 2015; it's too late to fix because at this point it could be considered a feature.
These properties do get reset by `border` after all.
i.e. for `declaration-block-no-shorthand-property-overrides`
It's just not consistent with the rest nor with the original `shorthand.js` name intent.
Just a minor annoyance IMHO.

### DIFF
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/index.mjs
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/index.mjs
@@ -284,7 +284,14 @@ const rule = (primary, secondaryOptions, context) => {
 						longhandSubPropertiesOfShorthandProperties.get(shorthandProperty),
 					);
 
-					ignoreLonghands.forEach((value) => shorthandProps.delete(value));
+					// see 9ab4fb2 for the historical reason behind the exception
+					const isBorder = shorthandProperty === 'border';
+
+					shorthandProps.forEach((value) => {
+						const mustDiscard = isBorder && value.length !== 12;
+
+						if (ignoreLonghands.includes(value) || mustDiscard) shorthandProps.delete(value);
+					});
 					const prefixedShorthandData = Array.from(shorthandProps, (item) => prefix + item);
 					const copiedPrefixedShorthandData = [...prefixedShorthandData];
 
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/index.mjs
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/index.mjs
@@ -286,9 +286,10 @@ const rule = (primary, secondaryOptions, context) => {
 
 					// see 9ab4fb2 for the historical reason behind the exception
 					const isBorder = shorthandProperty === 'border';
+					const borderSubPropertyLength = 12;
 
 					shorthandProps.forEach((value) => {
-						const mustDiscard = isBorder && value.length !== 12;
+						const mustDiscard = isBorder && value.length !== borderSubPropertyLength;
 
 						if (ignoreLonghands.includes(value) || mustDiscard) shorthandProps.delete(value);
 					});
--- a/.changeset/swift-jobs-fold.md
+++ b/.changeset/swift-jobs-fold.md
@@ -2,4 +2,4 @@
 "stylelint": minor
 ---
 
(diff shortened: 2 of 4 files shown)


lib/rules/declaration-block-no-redundant-longhand-properties/index.mjs
```diff
@@ -284,7 +284,14 @@ const rule = (primary, secondaryOptions, context) => {
...
 
-					ignoreLonghands.forEach((value) => shorthandProps.delete(value));
+					// see 9ab4fb2 for the historical reason behind the exception
+					const isBorder = shorthandProperty === 'border';
+
+					shorthandProps.forEach((value) => {
+						const mustDiscard = isBorder && value.length !== 12;
+
+						if (ignoreLonghands.includes(value) || mustDiscard) shorthandProps.delete(value);
+					});
```
ybiquitous, line 294: [suggest] I think we can just customize `longhandSubPropertiesOfShorthandProperties` for this rule like this:

```js
const shorthandToLonghands = new Map(longhandSubPropertiesOfShorthandProperties);
shorthandToLonghands.set('border', new Set(['border-width', 'border-style', 'border-color']));
```

I'd like to avoid the magic number `12` (in `value.length !== 12`).

ybiquitous (commented):
@Mouvedia Thanks for the pull request. Can you check my review?

lib/rules/declaration-block-no-redundant-longhand-properties/index.mjs
Mouvedia: I thought about it, I chose not to. I hesitated because this change is hacky and preferred the diff to be small.
Ill add a `const`.

ybiquitous: Even after extracting the magic number to a local variable via 63b41d7, the number is still present. This means that we have to fix the hack later.

I think we should write non-hack code if possible.

Mouvedia: I meant that not fixing `properties.mjs` is sloppy and I think that this exception will be temporary.
i.e. `border` should be updated in the next major

ybiquitous: I suggest adding the following code to `lib/rules/declaration-block-no-redundant-longhand-properties/index.mjs`, not `properties.mjs`. E.g.

```diff
const meta = {
	url: 'https://stylelint.io/user-guide/rules/declaration-block-no-redundant-longhand-properties',
	fixable: true,
};

+const shorthandToLonghands = new Map(longhandSubPropertiesOfShorthandProperties);
+shorthandToLonghands.set('border', new Set(['border-width', 'border-style', 'border-color']));
```

This code doesn't mutate the `longhandSubPropertiesOfShorthandProperties` object, so there is no effect on other modules.

Mouvedia:
It's not as simple as it looks.
I was wrong about the `border` shorthand.
It's not an error: it's by design.
`border` will require a custom resolver.

lib/rules/declaration-block-no-redundant-longhand-properties/index.mjs
Mouvedia: Thanks for the review but I realised that I was wrong along the way.
So this is moot :)

PR #7626, opened 2024-04-19
Fix `declaration-block-no-redundant-longhand-properties` autofix conflicts

### LINKED ISSUE #7610: Fix `declaration-block-no-redundant-longhand-properties` autofix conflicts

```css
a {
	border-top-width: foo;
	border-right-width: foo;
	border-bottom-width: foo;
	border-left-width: foo;

	border-top-color: foo;
	border-right-color: foo;
	border-bottom-color: foo;
	border-left-color: foo;

	border-top-style: foo;
	border-right-style: foo;
	border-bottom-style: foo;
	border-left-style: foo;
}
```

```json
{
  "rules": {
  },
  "fix": true
}
```

https://stylelint.io/demo

```css
a {
	border-width: foo foo foo foo;
	border-color: foo foo foo foo;
	border-style: foo foo foo foo;
}
```
or
```css
a {
	border-top: foo foo foo;
	border-right: foo foo foo;
	border-bottom: foo foo foo;
	border-left: foo foo foo;
}
```

### What actually happened?

demo

```css
a {
	border-width: foo foo foo foo;
	border-color: foo foo foo foo;
}
```

locally

```css
a {
	border-width: foo foo foo foo;
	border-color: foo foo foo foo;
    border-left-style: foo;
}
```

@ybiquitous @jeddy3
It seems to me that it's a **major bug**.

Mouvedia:

Closes #7610

No, it's self-explanatory.

### DIFF
--- a/.changeset/large-olives-press.md
+++ b/.changeset/large-olives-press.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fix `declaration-block-no-redundant-longhand-properties` autofix conflicts
--- a/.changeset/large-olives-press.md
+++ b/.changeset/large-olives-press.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fix `declaration-block-no-redundant-longhand-properties` autofix conflicts
+Fixed: `declaration-block-no-redundant-longhand-properties` autofix conflicts
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
@@ -220,6 +220,389 @@ testRule({
 			],
 		},
 		{
+			code: stripIndent`
+				a {
+					border-top-width: 1px;
+					border-right-width: 2px;
+					border-bottom-width: 3px;
+					border-left-width: 4px;
+					border-top-style: dotted;
+					border-right-style: solid;
+					border-bottom-style: dashed;
+					border-left-style: double;
+					border-top-color: red;
+					border-right-color: green;
+					border-bottom-color: blue;
+					border-left-color: yellow;
+				}
+			`,
+			fixed: stripIndent`
+				a {
+					border-width: 1px 2px 3px 4px;
+					border-style: dotted solid dashed double;
+					border-color: red green blue yellow;
+				}
+			`,
+			warnings: [
+				{
+					line: 5,
+					column: 2,
+					endLine: 5,
+					endColumn: 19,
+					message: messages.expected('border-width'),
+				},
+				{
+					line: 9,
+					column: 2,
+					endLine: 9,
+					endColumn: 19,
+					message: messages.expected('border-style'),
+				},
+				{
+					line: 10,
+					column: 2,
+					endLine: 10,
+					endColumn: 18,
+					message: messages.expected('border-top'),
+				},
+				{
+					line: 11,
+					column: 2,
+					endLine: 11,
+					endColumn: 20,
+					message: messages.expected('border-right'),
+				},
+				{
+					line: 12,
+					column: 2,
+					endLine: 12,
+					endColumn: 21,
+					message: messages.expected('border-bottom'),
+				},
+				{
+					line: 13,
+					column: 2,
+					endLine: 13,
+					endColumn: 19,
+					message: messages.expected('border-color'),
+				},
+				{
+					line: 13,
+					column: 2,
+					endLine: 13,
+					endColumn: 19,
+					message: messages.expected('border-left'),
+				},
+			],
+		},
+		{
+			code: stripIndent`
+				a {
+					border-top-width: 1px;
+					border-top-style: dotted;
+					border-top-color: red;
+					border-right-width: 2px;
+					border-right-style: solid;
+					border-right-color: green;
+					border-bottom-width: 3px;
+					border-bottom-style: dashed;
+					border-bottom-color: blue;
+					border-left-width: 4px;
+					border-left-style: double;
+					border-left-color: yellow;
+				}
+			`,
+			fixed: stripIndent`
+				a {
+					border-top: 1px dotted red;
+					border-right: 2px solid green;
+					border-bottom: 3px dashed blue;
+					border-left: 4px double yellow;
+				}
+			`,
+			warnings: [
+				{
+					line: 4,
+					column: 2,
+					endLine: 4,
+					endColumn: 18,
+					message: messages.expected('border-top'),
+				},
+				{
+					line: 7,
+					column: 2,
+					endLine: 7,
+					endColumn: 20,
+					message: messages.expected('border-right'),
+				},
+				{
+					line: 10,
+					column: 2,
+					endLine: 10,
+					endColumn: 21,
+					message: messages.expected('border-bottom'),
+				},
+				{
+					line: 11,
+					column: 2,
+					endLine: 11,
+					endColumn: 19,
+					message: messages.expected('border-width'),
+				},
+				{
+					line: 12,
+					column: 2,
+					endLine: 12,
+					endColumn: 19,
+					message: messages.expected('border-style'),
+				},
+				{
+					line: 13,
+					column: 2,
+					endLine: 13,
+					endColumn: 19,
+					message: messages.expected('border-color'),
+				},
+				{
+					line: 13,
+					column: 2,
+					endLine: 13,
+					endColumn: 19,
+					message: messages.expected('border-left'),
+				},
+			],
+		},
+		{
+			code: stripIndent`
+				a {
+					border-top-width: 1px;
+					border-top-style: dotted;
+					border-top-color: red;
+					border-bottom-width: 3px;
+					border-bottom-style: dashed;
+					border-bottom-color: blue;
+					border-right-width: 2px;
+					border-right-style: solid;
+					border-right-color: green;
+					border-left-width: 4px;
+					border-left-style: double;
+					border-left-color: yellow;
+				}
+			`,
+			fixed: stripIndent`
+				a {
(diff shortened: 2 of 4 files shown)


lib/rules/declaration-block-no-redundant-longhand-properties/index.mjs
```diff
@@ -314,9 +314,14 @@ const rule = (primary, secondaryOptions, context) => {
 								});
 
 								firstDeclNode.replaceWith(newShorthandDeclarationNode);
-
 								declNodes.forEach((node) => node.remove());
 
+								longhandDeclarations.forEach((v, k) => {
+									const array = v.filter((element) => !longhandDeclaration.includes(element));
+
+									longhandDeclarations.set(k, array);
```
Mouvedia, line 322: I could have done `v.splice(0, v.length,...v.filter((element) => !longhandDeclaration.includes(element)));` instead.

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
```diff
@@ -142,6 +142,78 @@ testRule({
 			endLine: 1,
 			endColumn: 138,
 		},
+		{
+			code: `a {
```
ybiquitous, line 146: [suggest] I recommend using `stripIndent` like other test codes:

```js
```

```suggestion
			code: stripIndent`a {
```

If doing so, we won't need to care about `column` values.

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
```diff
@@ -142,6 +142,78 @@ testRule({
...
+				border-width: 1px 2px 3px 4px;
+				border-color: red green blue yellow;
+				border-style: dotted solid dashed double;
+			}`,
+			warnings: [
+				{
+					column: 5,
+					endColumn: 22,
+					endLine: 5,
+					line: 5,
```
ybiquitous, line 170: [suggest] The order of `line`, `column`, `endLine`, and `endColumn` is more consistent:

```suggestion
					line: 5,
					column: 5,
					endLine: 5,
					endColumn: 22,
```

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
Mouvedia: It didn't change `endColumn` and `column`.

Mouvedia: It didn't change `endColumn` and `column`.
What's the desired outcome? I don't get it.

ybiquitous: Sorry, I mean the following should be:

```js
			code: stripIndent`
				a {
					border-top-width: 1px;
					border-right-width: 2px;
					border-bottom-width: 3px;
					border-left-width: 4px;
					border-top-color: red;
					border-right-color: green;
					border-bottom-color: blue;
					border-left-color: yellow;
					border-top-style: dotted;
					border-right-style: solid;
					border-bottom-style: dashed;
					border-left-style: double;
				}
			`,
			fixed: stripIndent`
				a {
					border-width: 1px 2px 3px 4px;
					border-color: red green blue yellow;
					border-style: dotted solid dashed double;
				}
			`,
```

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
ybiquitous: 2c3d386 doesn't resolve it. Please see the order in the same file: https://github.com/stylelint/stylelint/blob/2c3d38652c3b9a8be6abf829ab02c45ac5bcd8f7/lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs#L95-L98

.changeset/large-olives-press.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fix `declaration-block-no-redundant-longhand-properties` autofix conflicts
```
ybiquitous, line 5: [suggest] Let's clarify the resolve bug:

```suggestion
Fix `declaration-block-no-redundant-longhand-properties` autofix conflicts for `border-*`
```

ybiquitous:
I notice that this problem is about `border-*-style` false positive, rather than generic autofix. See the [demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYAdAdugLgIwgCdZCBaHCAB1IHcBLKHAC3hQEZKAPAbi1wOIwyhOgHMmOWg2asATF17Z8REqQI4KAWymMWKAMwK+ywWQA2MAGaT6u1gBYjSgaorVIZoq0Iwoi-ipCpCLikh5eKKI+MJj+JqrqWqThhKx4ZgCuMHEuQRbWyRCeqSgAnjBmnjQ5gWRupADOOKUWrFAQGr41psFiEo3NrSgNRQzdCR1JTS0wbUgNTF3GueZWktND7Rnp2VgAviAANCCWdBYAckiacIgwnFeUFgB0YA0NR+AQmKeiCCAYmBQKHQIDuOBiUAaINYAG0sECgSCNhU6Jgwl8fgMkJgoEhiCD4SgALr7D6Qb5iABiRE0SBwfwAViNMB9YJR3ogAYiQMizKicNDgSAzHSYE0QYdCUjBii0YUKaIsTi8VBBSCReDxSBSXsgA):

image

`border-*-style` props are reported too much, comparing `border-*-width` and `border-*-color`.

.changeset/large-olives-press.md
ybiquitous: See https://github.com/stylelint/stylelint/pull/7626#issuecomment-2065876959. The following may be more accurate:

```suggestion
Fix `declaration-block-no-redundant-longhand-properties` false positives for `border-*-style`
```

Mouvedia: This is the only example that I could find.
The underlying question is, if we care about perf, should we put an `if` around the `forEach`?
i.e. `if (blacklist.includes(foo)) {…}`

Mouvedia:
Apart from

> `Expected shorthand property "border"
[(declaration-block-no-redundant-longhand-properties)](https://stylelint.io/user-guide/rules/declaration-block-no-redundant-longhand-properties)
[13:2-19]`

which will be covered by #7609 in a particular case:

```css
a {
	border-top-width: foo;
	border-right-width: foo;
	border-bottom-width: foo;
	border-left-width: foo;
	border-top-color: bar;
	border-right-color: bar;
	border-bottom-color: bar;
	border-left-color: bar;
	border-top-style: qux;
	border-right-style: qux;
	border-bottom-style: qux;
	border-left-style: qux;
}
```

the rest looks OK.
i.e. I expect 7 warnings but only 3 fixes attempted.

.changeset/large-olives-press.md
Mouvedia: > for `border-*`

It's inaccurate. Only a subset of `border-*` are affected and the current implementation doesn't differentiate.
i.e. it's future proof but probably could use a blacklist instead

Mouvedia:
@ybiquitous I have added an `if` around the `forEach` so that it doesn't loop for other shorthands.
I feel that this is a major bug and that this PR is fine as is.
If you have suggestions for other tests Ill welcome them.
Concerning a refactor, this will have to wait.

lib/rules/declaration-block-no-redundant-longhand-properties/__tests__/index.mjs
Mouvedia: Sorry about that I must read it incorrectly.

.changeset/large-olives-press.md
Mouvedia: I have added the blacklist in the last commit.

ybiquitous:
@Mouvedia I'm against adding this workaround, and again we should reconsider the implementation even if it would be a big change. Honestly, I don't want to review PRs so many times only for one rule, and this workaround may occur other problems.

In addition, this bug has been not found for a long time. You say this is "major", but I guess many people are not in trouble.

I think it's a good timing to detect the rule's implementation design.

Mouvedia:
Ill just explain what this fix does then, maybe that will change your mind.
The rule creates an array that it populates with longhands that it encounters inside a declaration block, for each shorthand supported. Once the array matches it can move on to the fixer.
Before the fix, once the replacement of the longhands by the shorthand was done it didn't remove those longhands from the other arrays that were being populated at the same time. Inevitably it would cause useless autofixes to be attempted because the arrays weren't cleaned up. The first iteration of the fix was doing the cleanup without any regard of whether the shorthands were affected by the bug. The last iteration only does it for a shortlist of shorthands, it's less future proof and requires to maintain a blacklist.

I hope that's clearer and I managed to convince you.

.changeset/large-olives-press.md
ybiquitous, line 5:
```suggestion
Fixed: `declaration-block-no-redundant-longhand-properties` autofix conflicts
```

ybiquitous (approved):
Okay, I agree with merging this since it resolves at lease autofix problem. But I still have the opinion to reconsider this rule's implementation as soon as possible.

If you fix the changelog, you can merge. LGTM.

ybiquitous:
The discussions about this rule have been opened in #7630 and #7631. They indicates there is some problem in the rule implementation.

Mouvedia:
> …you can merge…

I have added more tests to cover more permutations.

> The discussions about this rule have been opened in https://github.com/stylelint/stylelint/issues/7630 and https://github.com/stylelint/stylelint/issues/7631. They indicates there is some problem in the rule implementation.

We can continue the discussion around maintenance and refactors there.

Issue #7623, opened 2024-04-18
Fix `function-no-unknown` false positive for `palette-mix`


Mouvedia:
### What minimal example or steps are needed to reproduce the bug?

```css
.bar { foo: palette-mix(); }
```

### What minimal configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "function-no-unknown": true
  }
}
```

### How did you run Stylelint?

https://stylelint.io/demo

### What actually happened?

`Unexpected unknown function "palette-mix"`

<hr/>

https://caniuse.com/mdn-css_properties_font-palette_palette-mix_function
niksy/css-functions-list/issues/7

ybiquitous:
@Mouvedia Thanks for the report. Sounds good 👍🏼 

Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/font-palette/palette-mix

Mouvedia:
🎉 
We have the green light of @niksy to fix it upstream; no more exceptions on our side.

Mouvedia:
@ybiquitous can you trigger a dependabot scan for https://www.npmjs.com/package/css-functions-list/v/3.2.2?

ybiquitous:
@Mouvedia Opened #7640

PR #7642, opened 2024-04-23
Refactor `function-no-unknown` to remove exceptions

Mouvedia:

#7623
#7640
niksy/css-functions-list#8

At one point we will have to vendor the package if the non-existing functions are not removed.
For now it's just a minor bother.

https://github.com/stylelint/stylelint/blob/main/lib/rules/function-no-unknown/__tests__/index.mjs#L25
https://github.com/stylelint/stylelint/blob/main/lib/rules/function-no-unknown/__tests__/index.mjs#L31
https://github.com/stylelint/stylelint/blob/main/lib/rules/function-no-unknown/__tests__/index.mjs#L34


ybiquitous (approved):
@Mouvedia Thanks. LGTM.

PR #7647, opened 2024-04-24
Refactor utilities to unify `parseSelector()` and `parseSelectorAST()`

ybiquitous:

None.

This unifies similar utilities using `postcss-selector-parser` into one. The new utility returns a root object for selectors.

> [!TIP]
> I recommend [ignoring whitespace changes](https://github.com/stylelint/stylelint/pull/7647/files?w=1) for code review.


lib/rules/no-descending-specificity/index.cjs
```diff
@@ -88,9 +88,11 @@ const rule = (primary, secondaryOptions) => {
 						continue;
 					}
 
-					parseSelector(resolvedSelector, result, ruleNode, (s) => {
-						checkSelector(resolvedSelector, s, ruleNode, comparisonContext);
-					});
+					const selectorRoot = parseSelector(resolvedSelector, result, ruleNode);
```
Mouvedia, line 91: major

I think that changing the signature of that function will break many plugins that use the callback argument.
It's under utils so it's public.
Let's wait for v17?

see https://github.com/search?q=%22stylelint%2Flib%2Futils%2FparseSelector%22&type=code

lib/utils/parseSelector.mjs
```diff
@@ -4,12 +4,13 @@ import selectorParser from 'postcss-selector-parser';
  * @param {string} selector
  * @param {import('stylelint').PostcssResult} result
  * @param {import('postcss').Node} node
- * @param {(root: import('postcss-selector-parser').Root) => void} callback
- * @returns {string | undefined}
+ * @returns {import('postcss-selector-parser').Root | undefined}
  */
-export default function parseSelector(selector, result, node, callback) {
+export default function parseSelector(selector, result, node) {
```
Mouvedia, line 9: major

I think that changing the signature of that function will break many plugins that use the callback argument.
It's under utils so it's public.
Let's wait for v17?
Or fallback to the previous implementation if the callback is passed?
e.g.
```js
return callback ? selectorParser(callback).processSync(selector) : selectorParser().astSync(selector);
```

see https://github.com/search?q=%22stylelint%2Flib%2Futils%2FparseSelector%22&type=code

```suggestion
export default function parseSelector(selector, result, node, callback) {
```

ybiquitous: Good point. I've added a fallback via d81656b8e. 👍🏼

lib/utils/parseSelector.mjs
```diff
@@ -4,12 +4,20 @@ import selectorParser from 'postcss-selector-parser';
  * @param {string} selector
  * @param {import('stylelint').PostcssResult} result
  * @param {import('postcss').Node} node
- * @param {(root: import('postcss-selector-parser').Root) => void} callback
- * @returns {string | undefined}
+ * @param {(root: import('postcss-selector-parser').Root) => void} [callback] - Deprecated. It will be removed in the future.
```
Mouvedia, line 7: minor nit

https://github.com/jsdoc/jsdoc.github.io/issues/39#issuecomment-18919123

ybiquitous: I feel that not using `@deprecated` is more explicit though.

lib/rules/selector-max-specificity/index.mjs
```diff
@@ -246,19 +246,21 @@ const rule = (primary, secondaryOptions) => {
...
-								ruleName,
-								result,
-								node: ruleNode,
-								message: messages.expected,
-								messageArgs: [resolvedSelector, primary],
-								word: selector,
-							});
-						}
-					});
+					const selectorTree = parseSelector(resolvedSelector, result, ruleNode);
```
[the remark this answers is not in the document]
ybiquitous: I don't like callbacks here because they hide a `if` statement. Checking a return value is more explicit, and these refactoring at this level is acceptable, I think.

lib/rules/selector-not-notation/index.mjs
```diff
@@ -78,48 +78,49 @@ const rule = (primary, _, context) => {
...
+					messageArgs: [primary],
+					node: ruleNode,
+					index: pseudo.sourceIndex,
+					endIndex: pseudo.source.end.column,
+					result,
+					ruleName,
 				});
 			});
 
-			if (context.fix && fixedSelector) {
```
Mouvedia, line 121: note

This was a potential bug.

https://github.com/stylelint/stylelint/blob/713f36f1f7e02723e1949524e53aab7b000853c4/lib/rules/selector-not-notation/__tests__/index.mjs#L35-L36

could potentially return `''` if it ever was supported.

lib/rules/selector-pseudo-element-colon-notation/index.mjs
```diff
@@ -48,43 +48,45 @@ const rule = (primary, _secondaryOptions, context) => {
...
-
-					report({
-						message: messages.expected(primary),
-						node: ruleNode,
-						index: pseudo.sourceIndex,
-						endIndex: pseudo.sourceIndex + (isDouble ? 2 : 1),
-						result,
-						ruleName,
-					});
+			const selectorRoot = parseSelector(selector, result, ruleNode);
```
Mouvedia, line 51: question

Can't you use `?.walkPseudos(`?

lib/utils/__tests__/parseSelector.test.mjs
```diff
@@ -45,4 +45,17 @@ describe('parseSelectorAST', () => {
...
+		expect.assertions(3);
+
+		const result = await postcss().process('a {}', { from: undefined });
+		const rule = result.root.first;
+		const processed = parseSelector(rule.selector, result, rule, (selectorRoot) => {
+			expect(selectorRoot).toHaveProperty('type', 'root');
+			expect(selectorRoot).toHaveProperty('nodes[0].nodes[0].value', 'a');
+		});
+
+		expect(typeof processed).toBe('string');
```
Mouvedia, line 59: note

jest-extended has [`toBeString`](https://jest-extended.jestcommunity.dev/docs/matchers/String#tobestring)

lib/rules/selector-pseudo-element-colon-notation/index.mjs
ybiquitous: Code you suggested has the same effect, right? I think early return is more understandable.

Mouvedia (approved):

ybiquitous:
Thanks for the review 👍🏼

PR #7650, opened 2024-04-25
Add regex support to `ignoreValues` for `value-no-vendor-prefix`

### LINKED ISSUE #7645: Add regex support to `value-no-vendor-prefix` secondary option

see the _Caveat_ section of #7624

`ignoreValues: ["string"]`
becomes
`ignoreValues: ["/regex/", /regex/, "string"]`

Mouvedia:

Closes #7645

The previously **non-documented** _partial_ regex support that performed the comparison on the unprefixed value has been scrapped in favor of a documented support which compares exactly.
cf [`8a1a968` (#3015)](https://github.com/stylelint/stylelint/pull/3015/commits/8a1a9685e81251c7ad9e27fe8df1835badb796c0)


Mouvedia:
```
error - 2024-04-25 13:03:37,676 -- Commit creating failed: {"detail":"Tokenless has reached GitHub rate limit. Please upload using a token: https://docs.codecov.com/docs/adding-the-codecov-token. Expected available in 850 seconds."}
```

@ybiquitous In the meantime, you can review.

lib/rules/value-no-vendor-prefix/README.md
```diff
@@ -72,5 +72,20 @@ cursor: -webkit-grab;
...
+
+<!-- prettier-ignore -->
+```css
+.foo { -webkit-appearance: -apple-pay-button; }
+```
+
+> [!WARNING]
+> An _exact_ match comparison will be performed for non-regex strings in the next major version.
+> If you want to keep the legacy behavior, please consider using a regex instead.
+> e.g. `[/^(-webkit-|-moz-|-ms-)?inline-/]`
```
Mouvedia, line 91:
```suggestion
> An _exact_ match comparison will be performed for non-regex strings in the next major version.  
> If you want to keep the legacy behavior, please consider using a regex instead.  
> e.g. `[/^(-webkit-|-moz-|-ms-)?inline-/]`
```

ybiquitous:
https://github.com/stylelint/stylelint/pull/7650#issuecomment-2077135495

Probably, forked branches cannot read secrets stored in the repository, such as a Codecov token.

lib/rules/value-no-vendor-prefix/index.mjs
```diff
@@ -52,8 +53,23 @@ const rule = (primary, secondaryOptions, context) => {
 				return;
 			}
 
-			if (optionsMatches(secondaryOptions, 'ignoreValues', vendor.unprefixed(value))) {
-				return;
+			if (secondaryOptions?.ignoreValues) {
+				const isIgnored = secondaryOptions.ignoreValues.some(
+					(/** @type { string | RegExp } */ element) => {
```
ybiquitous, line 58: [suggest] It's possible to remove the argument `@type`:

```suggestion
					(element) => {
```

by specifying type parameters to `rule`:

```diff
-/** @type {import('stylelint').Rule} */
+/** @type {import('stylelint').Rule<boolean, Array<string | RegExp>>} */
 const rule = (primary, secondaryOptions, context) => {
```

lib/rules/value-no-vendor-prefix/README.md
ybiquitous: We don't care about small screens and already have long lines, for example, see L14 in this file.

In addition, our Editorconfig settings may remove trailing whitespaces. So, it seems better to avoid trailing whitespaces as possible:

https://github.com/stylelint/stylelint/blob/8c23fcecc93edb3be32eaad313a7270fae6451e4/.editorconfig#L9

lib/rules/value-no-vendor-prefix/index.mjs
```diff
@@ -52,8 +53,23 @@ const rule = (primary, secondaryOptions, context) => {
...
+						const isStringifiedRegex =
+							isString(element) && element[0] === '/' && element.at(-1) === '/';
+						const useRawValue = element instanceof RegExp || isStringifiedRegex;
```
ybiquitous, line 61: [suggest]

- `isRegExp()` is available also here
- `startsWith()` and `endsWith()` can make it a bit readable, IMO
- Lazily evaluating a string `element` (e.g. `isString()` etc.) seems a bit more performant

```suggestion
						const useRawValue = isRegExp(element) || (isString(element) && element.startsWith('/') && element.endsWith('/'));
```

lib/rules/value-no-vendor-prefix/README.md
```diff
@@ -72,5 +72,20 @@ cursor: -webkit-grab;
...
+
+<!-- prettier-ignore -->
+```css
+.foo { -webkit-appearance: -apple-pay-button; }
+```
+
+> [!WARNING]
+> An _exact_ match comparison will be performed for non-regex strings in the next major version.  
+> If you want to keep the legacy behavior, please consider using a regex instead.  
+> e.g. `[/^(-webkit-|-moz-|-ms-)?inline-/]`
```
ybiquitous, line 91:
```suggestion
> An _exact_ match comparison will be performed for non-regex strings in the next major version.
> If you want to keep the legacy behavior, please consider using a regex instead.
> E.g. `[/^(-webkit-|-moz-|-ms-)?inline-/]`.
```

lib/rules/value-no-vendor-prefix/README.md
ybiquitous: Okay, how about this? https://github.com/stylelint/stylelint/pull/7650/files#r1579504581

lib/rules/value-no-vendor-prefix/index.mjs
```diff
@@ -52,8 +53,23 @@ const rule = (primary, secondaryOptions, context) => {
...
+						/**
+						 * @todo consolidate in the next major
+						 * @see stylelint/stylelin#7542
+						 */
+						const input = useRawValue ? value : vendor.unprefixed(value);
+
+						return Boolean(testAgainstStringOrRegExp(input, element));
```
ybiquitous: [suggest] I find a way to avoid exporting `testAgainstStringOrRegExp`. This uses `optionsMatches()` as before. How about this?

```js
		if (!validOptions) {
			return;
		}

		/** @type {{ raw: Array<string | RegExp>, unprefixed: Array<string | RegExp> }} */
		const ignoreValues = { raw: [], unprefixed: [] };

		secondaryOptions?.ignoreValues?.forEach((element) => {
			const useRawValue =
				isRegExp(element) ||
				(element.startsWith('/') && (element.endsWith('/') || element.endsWith('/i')));

			if (useRawValue) {
				ignoreValues.raw.push(element);
			} else {
				ignoreValues.unprefixed.push(element);
			}
		});

		root.walkDecls((decl) => {
			// ...

			if (optionsMatches(ignoreValues, 'raw', value)) {
				return;
			}

			if (optionsMatches(ignoreValues, 'unprefixed', vendor.unprefixed(value))) {
				return;
			}

			// ...
		});
```

PR #7651, opened 2024-04-25
Refactor `matchesStringOrRegExp()` utility

ybiquitous:

None.

This change aims to improve readability in the `matchesStringOrRegExp()` utility module.

### DIFF
--- a/lib/utils/matchesStringOrRegExp.mjs
+++ b/lib/utils/matchesStringOrRegExp.mjs
@@ -6,10 +6,11 @@
  * Any strings starting and ending with `/` are interpreted
  * as regular expressions.
  *
+ * @typedef {{match: string, pattern: (string | RegExp), substring: string} | false} MatchResult
+ *
  * @param {string | Array<string>} input
  * @param {string | RegExp | Array<string | RegExp>} comparison
- *
- * @returns {false | {match: string, pattern: (string | RegExp), substring: string}}
+ * @returns {MatchResult}
  */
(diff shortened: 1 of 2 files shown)


lib/utils/matchesStringOrRegExp.mjs
```diff
@@ -47,39 +49,47 @@ function testAgainstStringOrRegExpOrArray(value, comparison) {
...
-
+	const regexFlag = 'i';
 	const comparisonIsRegex =
-		firstComparisonChar === '/' &&
-		(lastComparisonChar === '/' ||
-			(secondToLastComparisonChar === '/' && lastComparisonChar === 'i'));
-
-	const hasCaseInsensitiveFlag = comparisonIsRegex && lastComparisonChar === 'i';
+		comparison.startsWith('/') &&
+		(comparison.endsWith('/') || comparison.endsWith(`/${regexFlag}`));
```
Mouvedia, line 78:
```suggestion
		comparison.match(`\/${regexFlag}?$`);
```

ybiquitous: `endsWith` seems a bit faster than a regex, right?

Mouvedia (approved):
LGTM

lib/utils/matchesStringOrRegExp.mjs
ybiquitous: And, I feel `endsWith` is a bit clearer than a regex, considering `startsWith`.

PR #7654, opened 2024-04-25
Fix `value-no-vendor-prefix` false negatives

### LINKED ISSUE #7624: Fix `value-no-vendor-prefix` false negatives
While working on a fix I realised how the rule was intended to work. We can't change that behaviour after the fact.
It reports if the value can be unprefixed safely, else it bails-out.

In practice this is what you want and actually that's better for us: we don't have to update `stylelint-config-standard` with an `-apple-system` exception anymore because it is not an unprefixable value.

<hr/>

```css
body {
  font: 16px/1.2 BlinkMacSystemFont, -apple-system, "Segoe UI", Roboto, Helvetica, Arial, sans-serif;
}
```

```json
{
  "rules": {
  }
}
```

https://stylelint.io/demo

replace

https://github.com/stylelint/stylelint/blob/99591106dd306143625a89cca3ac8cbd9b53afea/lib/reference/prefixes.mjs#L1

by

```js
export const prefixes = new Set(['-webkit-', '-moz-', '-ms-', '-o-', '-konq-', '-apple-', '-khtml-']);
```

### Resources

- https://github.com/stylelint/stylelint/pull/7016#pullrequestreview-1507406116
- https://booking.design/implementing-system-fonts-on-booking-com-a-lesson-learned-bdc984df627f
- https://webkit.org/blog/3709/using-the-system-font-in-web-content/

### Caveat

`system` will have to be added to

https://github.com/stylelint/stylelint-config-standard/blob/3c610c147cc6c20b9e24292ff002b4637d55d3bc/index.js#L117

because it's used in many code bases as part of the default system font stack values.
Other values, like `-apple-system-body`, are rarely used, so it doesn't matter as much.

<hr />

There are 64 values that start with "-apple-system" supported by Safari.
see https://github.com/WebKit/WebKit/blob/main/Source/WebCore/css/CSSValueKeywords.in
Hence being able to ignore them using a regex is essential.
e.g. `["/^-apple-system/"]`
**So this fix will have to be introduced alongside an update of `ignoreValues`.**
e.g.

from
```md
ignoreValues: ["string"]
```
to
```md
ignoreValues: ["/regex/", /regex/, "string"]
```

Mouvedia:

related to #7624

While working on #7624 I realised how the rule was intended to work. We can't change that behaviour after the fact.
It reports if the value can be unprefixed safely, else it bails-out.

**In practice this is what you want and actually that's better for us: we don't have to update `stylelint-config-standard` with an `-apple-system` exception anymore because it is not an unprefixable value.**

This PR mainly fixes the documentation and some false negatives.
We can close #7624 after that because it's not on scope with the rule.
I didn't add all the missing values; I might some day.
I couldn't find use cases for `-konq-` so I left it out.

### DIFF
--- a/.changeset/serious-bees-burn.md
+++ b/.changeset/serious-bees-burn.md
@@ -1,5 +1,5 @@
 ---
-"stylelint": patch
+"stylelint": minor
 ---
 
-Fixed: false negatives for `value-no-vendor-prefix`
+Fixed: `value-no-vendor-prefix` false negatives
--- a/.changeset/serious-bees-burn.md
+++ b/.changeset/serious-bees-burn.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: false negatives for `value-no-vendor-prefix`
--- a/lib/reference/prefixes.mjs
+++ b/lib/reference/prefixes.mjs
@@ -1,1 +1,13 @@
-export const prefixes = new Set(['-webkit-', '-moz-', '-ms-', '-o-']);
+// used by value-no-vendor-prefix and selector-no-vendor-prefix
+// i.e. this list is deliberately not exhaustive
+// cf https://www.w3.org/TR/CSS22/syndata.html#vendor-keyword-history
+export const prefixes = new Set([
+	'-webkit-',
+	'-moz-',
+	'-ms-',
+	'-o-',
+	'-xv-',
+	'-apple-',
+	'-wap-',
+	'-khtml-',
+]);
--- a/lib/rules/value-no-vendor-prefix/README.md
+++ b/lib/rules/value-no-vendor-prefix/README.md
@@ -9,7 +9,7 @@ a { display: -webkit-flex; }
  *  This prefix */
 ```
 
-This rule ignores non-standard vendor-prefixed values that aren't handled by [Autoprefixer](https://github.com/postcss/autoprefixer).
+This rule does not fix vendor-prefixed values that weren't handled by [Autoprefixer](https://github.com/postcss/autoprefixer) version 10.2.5. Exceptions may be added on a case by case basis.
 
 The [`fix` option](../../../docs/user-guide/options.md#fix) can automatically fix all of the problems reported by this rule. However, it will not remove duplicate values produced when the prefixes are removed. You can use [Autoprefixer](https://github.com/postcss/autoprefixer) itself, with the [`add` option off and the `remove` option on](https://github.com/postcss/autoprefixer#options), in these situations.
 
@@ -60,7 +60,7 @@ a { background: linear-gradient(bottom, #000, #fff); }
 Given:
 
 ```json
-["grab", "hangul", "/^-apple-/"]
+["grab", "max-content", "/^-moz-all$/"]
 ```
 
 The following patterns are _not_ considered problems:
@@ -72,20 +72,15 @@ a { cursor: -webkit-grab; }
 
 <!-- prettier-ignore -->
 ```css
-a { list-style-type: -moz-hangul; }
+a { max-width: -moz-max-content; }
 ```
 
 <!-- prettier-ignore -->
 ```css
-a { list-style-type: -moz-hangul-consonant; }
-```
-
-<!-- prettier-ignore -->
-```css
-a { -webkit-appearance: -apple-pay-button; }
+a { -moz-user-select: -moz-all; }
 ```
 
 > [!WARNING]
 > An _exact_ match comparison will be performed for non-regex strings in the next major version.
 > If you want to keep the legacy behavior, please consider using a regex instead.
-> E.g. `[/^(-webkit-|-moz-|-ms-)?inline-/]`.
+> E.g. `[/^(-webkit-|-moz-|-ms-)?max-content$/]`.
--- a/lib/rules/value-no-vendor-prefix/__tests__/index.mjs
+++ b/lib/rules/value-no-vendor-prefix/__tests__/index.mjs
@@ -8,25 +8,41 @@ testRule({
 
 	accept: [
 		{
-			code: '.foo { display: flex; }',
+			code: 'a { display: flex; }',
 		},
 		{
-			code: '.foo { background: linear-gradient(to top, #000, #fff); }',
+			code: 'a { background: linear-gradient(to top, #000, #fff); }',
 		},
 		{
-			code: '.foo { max-width: max-content; }',
+			code: 'a { max-width: max-content; }',
 		},
 		{
-			code: '.foo { -webkit-transform: translate(0, 0); }',
+			code: 'a { -webkit-transform: translate(0, 0); }',
 			description: 'ignores property vendor prefixes',
 		},
 		{
-			code: '.foo { -wEbKiT-tRaNsFoRm: translate(0, 0); }',
+			code: 'a { -wEbKiT-tRaNsFoRm: translate(0, 0); }',
 			description: 'ignores property vendor prefixes',
 		},
 		{
-			code: '.foo { -WEBKIT-TRANSFORM: translate(0, 0); }',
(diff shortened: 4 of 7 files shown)


lib/rules/value-no-vendor-prefix/README.md
```diff
@@ -9,7 +9,7 @@ a { display: -webkit-flex; }
  *  This prefix */
 ```
 
-This rule ignores non-standard vendor-prefixed values that aren't handled by [Autoprefixer](https://github.com/postcss/autoprefixer).
+This rule does not fix vendor-prefixed values that weren't handled by [Autoprefixer](https://github.com/postcss/autoprefixer) version 10.2.5. Exceptions may be added on a case by case basis.
```
Mouvedia, line 12: I have added some examples of such cases that happen to be the false negatives that I am fixing in this PR.
see https://github.com/stylelint/stylelint/pull/7654/files#diff-db7ee227393b97a75cdd1e359512825b05fc5e14bfd94e304762622814f41a14

Mouvedia, line 12: For the version check https://github.com/stylelint/stylelint/pull/5312#discussion_r636094886

lib/rules/value-no-vendor-prefix/README.md
```diff
@@ -75,11 +75,6 @@ a { cursor: -webkit-grab; }
 a { list-style-type: -moz-hangul; }
 ```
 
-<!-- prettier-ignore -->
-```css
-a { list-style-type: -moz-hangul-consonant; }
```
Mouvedia, line 80: This was technically true that it was being ignored but it was misleading.
i.e. it was ignored because it wasn't part of the variables that are considered unprefixable
see https://github.com/stylelint/stylelint/pull/7654/files#diff-5f9be0b5e6e40c3ad07fc3db2a8895c25da53046e10664c0112bbb9e237893c5R295

lib/rules/value-no-vendor-prefix/__tests__/index.mjs
```diff
@@ -8,26 +8,42 @@ testRule({
...
 			description: 'ignores property vendor prefixes',
 		},
 		{
-			code: '.foo { -WEBKIT-TRANSFORM: translate(0, 0); }',
+			code: 'a { -WEBKIT-TRANSFORM: translate(0, 0); }',
 			description: 'ignores property vendor prefixes',
 		},
+		{
+			code: 'a { white-space: -pre-wrap; }',
+			description: 'ignores non-vendor prefixed values',
```
Mouvedia, line 33: see https://web.archive.org/web/20100502064159/http://users.tkk.fi:80/tkarvine/pre-wrap-css3-mozilla-opera-ie.html

lib/rules/value-no-vendor-prefix/__tests__/index.mjs
```diff
@@ -94,6 +110,15 @@ testRule({
 			endLine: 1,
 			endColumn: 28,
 		},
+		{
+			code: '.foo { display: -khtml-box; }',
+			fixed: '.foo { display: box; }',
```
Mouvedia, line 115: see https://github.com/Igalia/webkit/commit/a659cc29a3fe51efecedf48ef10c9bd96f105936#diff-29c17e702b85143afffde80667b4c59b196f34d765b2f77e1586d3c3de7f1679

lib/rules/value-no-vendor-prefix/__tests__/index.mjs
```diff
@@ -112,6 +137,24 @@ testRule({
 			endLine: 1,
 			endColumn: 35,
 		},
+		{
+			code: '.foo { speak: -xv-digits; }',
+			fixed: '.foo { speak: digits; }',
```
Mouvedia, line 142: see https://issues.chromium.org/issues/40813740

lib/rules/value-no-vendor-prefix/README.md
```diff
@@ -75,11 +75,6 @@ a { cursor: -webkit-grab; }
 a { list-style-type: -moz-hangul; }
```
Mouvedia, line 75: WIP todo replace with `max-content`

lib/rules/value-no-vendor-prefix/__tests__/index.mjs
```diff
@@ -112,6 +137,33 @@ testRule({
...
+			fixed: '.foo { white-space: pre-wrap; }',
+			message: messages.rejected('-o-pre-wrap'),
+			line: 1,
+			column: 21,
+			endLine: 1,
+			endColumn: 32,
+		},
+		{
+			code: '.foo { -webkit-user-select: -moz-all; }',
+			fixed: '.foo { -webkit-user-select: all; }',
```
Mouvedia, line 160: see https://bugzilla.mozilla.org/show_bug.cgi?id=1492958

lib/rules/value-no-vendor-prefix/README.md
```diff
@@ -72,20 +72,15 @@ a { cursor: -webkit-grab; }
...
-<!-- prettier-ignore -->
-```css
-a { list-style-type: -moz-hangul-consonant; }
+a { max-width: -moz-max-content; }
 ```
 
 <!-- prettier-ignore -->
 ```css
-a { -webkit-appearance: -apple-pay-button; }
+a { -moz-user-select: -moz-all; }
```
Mouvedia, line 80: The previous example was misleading, `-apple-pay-button` is always ignored.

lib/reference/prefixes.mjs
```diff
@@ -1 +1,13 @@
...
+// i.e. this list is deliberately not exhaustive
+// cf https://www.w3.org/TR/CSS22/syndata.html#vendor-keyword-history
+export const prefixes = new Set([
+	'-webkit-',
+	'-moz-',
+	'-ms-',
+	'-o-',
+	'-xv-',
+	'-apple-',
+	'-wap-',
```
Mouvedia, line 11: see https://www.developershome.com/wap/wcss/wcss_tutorial.asp?page=marqueeExtension

lib/reference/prefixes.mjs
```diff
@@ -1 +1,13 @@
...
+	'-khtml-',
```
Mouvedia, line 12: e.g. `:-khtml-drag`

.changeset/serious-bees-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
```
ybiquitous, line 2:
```suggestion
"stylelint": minor
```

.changeset/serious-bees-burn.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Fixed: false negatives for `value-no-vendor-prefix`
```
ybiquitous, line 5:
```suggestion
Fixed: `value-no-vendor-prefix` false negatives for more vendor prefixes
```

ybiquitous (commented):
Thanks, overall LGTM. I've left a few suggestions only for the changelog.

.changeset/serious-bees-burn.md
Mouvedia: > for more

That sounds weird. Ill try to reformulate so that the rule is first.

ybiquitous (approved):
LGTM 👍🏼

lib/rules/value-no-vendor-prefix/README.md
```diff
@@ -72,20 +72,15 @@ a { cursor: -webkit-grab; }
...
 ```
 
 > [!WARNING]
-> E.g. `[/^(-webkit-|-moz-|-ms-)?inline-/]`.
+> E.g. `[/^(-webkit-|-moz-|-ms-)?max-content$/]`.
```
Mouvedia, line 86: typo

```suggestion
E.g. `[/^(-webkit-|-moz-)?max-content$/]`.
```

PR #7655, opened 2024-04-26
Fix `function-calc-no-unspaced-operator` false negatives

### LINKED ISSUE #7181: Fix `function-calc-no-unspaced-operator` false negatives when combined with properly spaced operator

```css
.a {
  /* Error */
  top: calc(5px+3px);
}

.b {
  /* No error */
  top: calc(5px+3px - 2px);
}
```

```json
{
  "rules": {
    "function-calc-no-unspaced-operator": true
  }
}
```

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0BDABMAOgOzWg9AFRoCiATmRGWofjngC4QAO8aYKANmABQCszAB4BqAMxCAlAG4cAXxw4kAIwz0CxAHIQ0MClRp1caJq3ZdeAkeMFoAtGgBMkmdnnYQAGhAAzAJacYDRQAWzhEGEEQ5gCkMABnOM9wCGw-AHMEECwjTBAyAFcAuNy2bLw8XO987DAGXxTbDm5bbAhbarjmFDAYKFsWXRQmMhLjApg1NzckyFTfNIAxKmChzIArOJSk2GZExDK0XLiGAE8Azl9sBlHcziGYY9yPNSPT88uGRpT022OUbCgKDIUBuIDuDAe1xAchAsiAA)

```json
{
  "stylelint": "^15.10.3",
  "stylelint-config-standard": "^34.0.0"
}
```

Rule should report an error for each instance of an unspaced operator.

### What actually happened?

No error is reported if the expression contains at least one properly spaced operator.

This actually caused (well, failed to catch) a real error in our codebase. We had the following code:
```css
.foo {
  top: max(-75px, min(-75px, calc(-0.33 * (100vh - 311px -61px - 50px -64px) + 15px)));
}
```
which is not valid (apparently the `-` in `-61px` and `-64px` is interpreted as the unary minus). It worked for a long time because we were using `cssnano` which changed the spacing and "fixed" the expression - but when I switched to [@csstools/postcss-minify](https://www.npmjs.com/package/@csstools/postcss-minify) which leaves the whitespace alone, the style stopped working.

@romainmenke Thought this might interest you (it's not a bug in @csstools/postcss-minify but rather a bug that was previously hidden by another minifier) :)

_No response_

ybiquitous:

Closes #7181

This significantly changes the implementation by switching `postcss-value-parser` to `@csstools/css-parser-algorithms`.
Because `postcss-value-parser` is inactive and leaves bugs for a while, and we've used already `@csstools/css-parser-algorithms` for some rules.

In addition, this updates `tsconfig.json` to support ES2023 libraries. This change uses `Array.prototype.findLast()`, which has been supported since Node.js 18.0.0.
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLast


Mouvedia:
> support ES2023 libraries

🎉

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -66,167 +71,174 @@ const rule = (primary, _secondaryOptions, context) => {
...
-
-				const message =
-					position === 'before'
-						? messages.expectedBefore(operator)
-						: messages.expectedAfter(operator);
-
-				complain(message, decl, valueIndex + operatorSourceIndex, operator);
-			}
+			function checkWhitespaceAroundOperator(operatorNode, parent, index, position) {
+				const aroundNode = parent.value[index + (position === 'before' ? -1 : 1)];
```
Mouvedia, line 81: see l129 and l110

```
const isBefore = position === 'before';
```

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -242,20 +254,84 @@ function insertCharAtIndex(str, index, char) {
...
+ */
+function isScssInterpolation(node, parent) {
+	// E.g. "#{$foo}"
+	if (getDelimToken(node)?.value !== '#') return false;
+
+	const afterNode = parent.at(Number(parent.indexOf(node)) + 1);
+
+	if (!isSimpleBlockNode(afterNode)) return false;
+
+	const varNode = afterNode.at(0);
```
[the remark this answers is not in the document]
ybiquitous: > Do you trim whitespace?

What kind of case do you assume?

Considering `hasScssInterpolation.mjs`, this `varNode` seems unnecessary.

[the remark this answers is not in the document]
Mouvedia: If you can reuse the util, it would be preferable.
Because if the util is too permissive or not strict enough it can be fixed later on.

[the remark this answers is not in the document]
ybiquitous: We cannot reuse the util `hasScssInterpolation.mjs` for this rule because the util is regex-based. This module's private util is based on `css-parser-algorithms` AST.

tsconfig.json
```diff
@@ -3,7 +3,7 @@
 		"module": "Node16",
 		"moduleResolution": "Node16",
 		"target": "ES2022",
-		"lib": ["ES2022"],
+		"lib": ["ES2023"],
```
Mouvedia, line 6: question

Would it be safer to have something like

```js
declare global {
  interface Array<T> {
    findLastIndex(
      predicate: (value: T, index: number, obj: T[]) => unknown,
      thisArg?: any
    ): number;
    findLast<S extends T>(predicate: (value: T, index: number, array: T[]) => value is S, thisArg?: any): S | undefined;
    findLast(predicate: (value: T, index: number, array: T[]) => unknown, thisArg?: any): T | undefined;
  }
}
```

so that ppl don't use `toSorted`, `with`, etc.?

ybiquitous: Methods like `toSorted` are unsupported on Node.js 18, so the CI should fail even if not adding such custom types.

romainmenke:
quick heads up, I will review this by Monday at the latest :)

tsconfig.json
Mouvedia: related: microsoft/TypeScript#56796

lib/reference/keywords.mjs
```diff
@@ -474,3 +474,6 @@ export const namedColorsKeywords = new Set([
 	'yellow',
 	'yellowgreen',
 ]);
+
+/** @see https://drafts.csswg.org/css-values/#typedef-calc-keyword */
+export const calcKeywords = new Set(['e', 'pi', 'infinity', '-infinity', 'NaN']);
```
romainmenke, line 479: It might be better to treat any `ident` token as a valid calc keyword.

Specific features can define their own calc keywords and it will be more maintenance work to keep this list up to date. It will also add significant complexity to keep track of which keywords are valid in which context.

Given that this rule doesn't aim to validate the keywords used it might be better to consider any `ident` token a valid calc keyword.

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -1,14 +1,18 @@
...
-import { assert } from '../../utils/validateTypes.mjs';
+	isFunctionNode,
+	isSimpleBlockNode,
+	isTokenNode,
+	isWhitespaceNode,
+	parseListOfComponentValues,
+	stringify,
+	walk,
+} from '@csstools/css-parser-algorithms';
+import { tokenize } from '@csstools/css-tokenizer';
```
romainmenke, line 10:
```suggestion
import { tokenize, TokenType } from '@csstools/css-tokenizer';
```

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -242,20 +251,80 @@ function insertCharAtIndex(str, index, char) {
...
 /**
- * @param {import('postcss-value-parser').Node | undefined} node
- * @returns {node is import('postcss-value-parser').SpaceNode}
+ * @param {import('@csstools/css-parser-algorithms').TokenNode} node
+ * @returns {{value: string, startIndex: number, endIndex: number} | undefined}
+ */
+function getDelimToken(node) {
+	const [type, value, startIndex, endIndex] = node.value;
+
+	if (type !== 'delim-token') return;
```
romainmenke, line 260:
```suggestion
	if (type !== TokenType.Delim) return;
```

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -242,20 +251,80 @@ function insertCharAtIndex(str, index, char) {
...
+}
+
+/**
+ * @param {import('@csstools/css-parser-algorithms').WhitespaceNode} node
+ * @returns {string}
  */
-function isSingleSpace(node) {
-	return node != null && isSpace(node) && node.value === ' ';
+function getWhitespace(node) {
+	return node.value[0]?.[1] ?? '';
```
romainmenke, line 281:
```suggestion
	return node.toString();
```

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -242,20 +251,80 @@ function insertCharAtIndex(str, index, char) {
...
 }
 
+/** @see https://drafts.csswg.org/css-values/#typedef-calc-value */
+const OPERAND_TOKEN_TYPES = new Set([
+	'number-token',
+	'dimension-token',
+	'percentage-token',
+	'ident-token',
```
romainmenke, line 289:
```suggestion
	TokenType.Number,
	TokenType.Dimension,
	TokenType.Percentage,
	TokenType.Ident,
```

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -242,20 +251,80 @@ function insertCharAtIndex(str, index, char) {
...
+ * @returns {boolean}
  */
-function isOperator(node, operators = OPERATORS) {
-	return node != null && isWord(node) && operators.has(node.value);
+function isOperandToken(token) {
+	const [type, value] = token;
+
+	if (!OPERAND_TOKEN_TYPES.has(type)) return false;
+
+	if (type === 'ident-token' && !calcKeywords.has(value)) return false;
```
romainmenke, line 301:
```suggestion
	if (type === TokenType.Ident && !calcKeywords.has(value)) return false;
```

Suggestion : also remove the requirement for it to be a known calc keyword.

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -242,20 +251,80 @@ function insertCharAtIndex(str, index, char) {
...
+	return true;
+}
+
+/**
+ * @param {import('@csstools/css-parser-algorithms').ComponentValue | undefined} node
+ * @returns {boolean}
+ */
+function isOperandNode(node) {
+	return isTokenNode(node) && isOperandToken(node.value);
+}
```
romainmenke, line 312: Aren't these also valid operands?

```
calc((10px) + 5px)
calc(calc(10px) + 5px)
```

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -66,167 +71,171 @@ const rule = (primary, _secondaryOptions, context) => {
...
 
-			const parsedValue = valueParser(value);
+			const nodes = parseListOfComponentValues(tokenize({ css: value }));
 
-			parsedValue.walk((node) => {
-				if (!isFunction(node) || !FUNC_NAMES_REGEX.test(node.value)) return;
+			if (nodes.length === 0) return;
 
-				const { nodes } = node;
+			walk(nodes, ({ node: funcNode }) => {
```
romainmenke, line 212: Should we add beter recursion support?

```
calc((1px+2px) + 3px)
```

The `(1px+2px)` simple block is also a `calc-sum` and should be detected as an error.

This is currently also ignored : https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYAdAdugLjiADvCmEgDZgAUlAjAQB4DUATAwJQqMoDM7A3FgC+IADQgAZgEsyMAHJIAtnEQx6igjIB0YAM47R4CJikBzBCAyYUKdCFU4YmKDtvEA2lmvXbOnAE8ZMklMHABaSGNJE1DfJCckACcoW08UAF0hAwjTADEIBIUkHHMAKx0jA1gCfURLbxBfAJggkNcbEDIimF9bEVSff0DgsOyomJw4qETkhHbOhx6QTMEgA

Adding support for this is relatively simple with the walker states.

romainmenke (commented):
This looks really good @ybiquitous, thank you for this!

I've left a few notes here and also took a deep dive into this rule to see how I would approach it : https://github.com/stylelint/stylelint/pull/7670

There aren't any major issues with this change as far as I can tell 🎉

lib/rules/function-calc-no-unspaced-operator/index.mjs
Mouvedia: You have added examples of such cases in your other PR.
Let's add one test case here.
If it fails add `skip: true` for now.

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -70,169 +74,306 @@ const rule = (primary, _secondaryOptions, context) => {
...
 
-				if (!delimToken) return;
+					remainderTokens.forEach((remainderToken) => {
+						remainderToken[2] += token[3];
+						remainderToken[3] += token[3];
+					});
 
-				const { value: operator, startIndex } = delimToken;
-				const message = messages[isBefore ? 'expectedBefore' : 'expectedAfter'](operator);
+					tokens.splice(i + 1, 0, ...remainderTokens);
```
Mouvedia, line 120: Could you check that we didn't regress #6045?
see https://github.com/stylelint/stylelint/issues/6044#issuecomment-1112878879
It seems to be covered by https://github.com/stylelint/stylelint/pull/7655/files#diff-4aac0dd17733f1428b79c71d6c9d05c314d49cf340415478c12bf5d9dfd330c6R225 but I am being cautious.

Mouvedia:
Can you update [this line](https://github.com/stylelint/stylelint/blob/c56f71482f578c442839fb2d934919d16c9ee0c0/lib/rules/function-calc-no-unspaced-operator/README.md?plain=1#L3)?

ybiquitous:
> Can you update [this line](https://github.com/stylelint/stylelint/blob/c56f71482f578c442839fb2d934919d16c9ee0c0/lib/rules/function-calc-no-unspaced-operator/README.md?plain=1#L3)?

Done! a4c2707c8

lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
```diff
@@ -548,15 +566,139 @@ testRule({
...
+			code: 'a { padding: calc(1px +-2px); }',
+			fixed: 'a { padding: calc(1px + -2px); }',
+			message: messages.expectedAfter('+'),
+			line: 1,
+			column: 23,
+			endLine: 1,
+			endColumn: 24,
+		},
+		{
+			code: 'a { padding: calc(1px+2px+3px); }',
```
Mouvedia, line 588: This one should be `'a { padding: calc(1px+2px + 3px); }'` to catch regressions.
cf #7181

lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
```diff
@@ -548,15 +566,139 @@ testRule({
...
+			description: 'nested math expression',
+			warnings: [
+				{ message: messages.expectedBefore('+'), line: 1, endLine: 1, column: 22, endColumn: 23 },
+				{ message: messages.expectedBefore('+'), line: 1, endLine: 1, column: 37, endColumn: 38 },
+			],
+		},
+		{
+			code: 'a { top: rgb(from red calc(r+1) calc(g-+2) calc(clamp(10px+2px, g+2, none))); }',
+			fixed:
+				'a { top: rgb(from red calc(r + 1) calc(g- + 2) calc(clamp(10px + 2px, g + 2, none))); }',
```
Mouvedia, line 689: questions
medium

What's `g-`?
Shouldn't it be fixed to `calc(g - +2)` instead?
i.e. `-` is subtraction and `+` is the unary operator

lib/utils/typeGuards.mjs
Mouvedia, line 1: question
minor

Can you add `isTokenColon`?
see https://github.com/stylelint/stylelint/blob/fb54d13c79ec9094ed233dbb24a7050f9c702c6f/lib/utils/findMediaFeatureNames.mjs#L80

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -65,197 +86,486 @@ const rule = (primary, _secondaryOptions, context) => {
...
+	const after = [];
+	/** @type {ComponentValue | undefined} */
+	let operator = undefined;
+	/** @type {import ('@csstools/css-tokenizer').TokenDelim | undefined} */
+	let operatorToken = undefined;
+
+	let currentNode = container.value[cursor];
+
+	// Consume as much whitespace and comments as possible
+	while (isWhitespaceNode(currentNode) || isCommentNode(currentNode)) {
```
Mouvedia, line 489: suggestion
nit

```js
// Consume as much whitespace and comments as possible
const isConsumed = (node) => isWhitespaceNode(currentNode) || isCommentNode(currentNode);
```

see l502 l522 l536

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -65,197 +86,486 @@ const rule = (primary, _secondaryOptions, context) => {
...
+	/** @type {ComponentValue | undefined} */
+	let firstOperand = undefined;
+	/** @type {ComponentValue | undefined} */
+	let secondOperand = undefined;
+	/** @type {Array<ComponentValue>} */
+	const before = [];
+	/** @type {Array<ComponentValue>} */
+	const after = [];
+	/** @type {ComponentValue | undefined} */
+	let operator = undefined;
```
Mouvedia, line 482: style
refactor

```suggestion
	/** @type {Array.<ComponentValue | undefined>} */
	let [firstOperand, secondOperand, operator] = [];
	/** @type {Array<ComponentValue>} */
	const before = [];
	/** @type {Array<ComponentValue>} */
	const after = [];
```

lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
romainmenke: `g-` is an ident token while we currently only scan for dimension tokens (e.g. `10px-`)

https://github.com/stylelint/stylelint/blob/4055b41b85ea9c1c1e2c529c5c4a97c7e289bc98/lib/rules/function-calc-no-unspaced-operator/index.mjs#L210

We can extend L210 to also look for idents and add extra logic to support both the `value` of an ident and the `unit` of a dimension.

Mouvedia: I think that @ybiquitous wouldn't mind if you were to commit on this branch.

lib/utils/typeGuards.cjs
```diff
@@ -93,13 +95,85 @@ function hasSource(node) {
...
+function isTokenComma(token) {
+	return token !== undefined && token[0] === cssTokenizer.TokenType.Comma;
+}
+
+/**
+ * @param {import('@csstools/css-tokenizer').CSSToken | undefined} token
+ * @returns {token is import('@csstools/css-tokenizer').TokenDelim}
+ */
+function isTokenDelim(token) {
+	return token !== undefined && token[0] === cssTokenizer.TokenType.Delim;
```
romainmenke, line 111: I've added type guards upstream.

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -65,197 +86,486 @@ const rule = (primary, _secondaryOptions, context) => {
...
+					if (indexOfDash === -1) return;
+
+					const remainder = unit.slice(indexOfDash);
+
+					if (remainder.length === 1) return;
 
-				assert(firstNode);
+					token[4].unit = unit.slice(0, indexOfDash);
+					token[1] = `${signCharacter === '+' ? '+' : ''}${dimensionValue}${token[4].unit}`;
+					token[3] = token[2] + token[1].length;
```
romainmenke, line 123: I've added a `mutateUnit` helper upstream.

lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
ybiquitous: But, autofixing to `calc(1px + 2px + 3px)` is better than `calc(1px+2px + 3px)`, isn't it?

lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
ybiquitous: This problem is resolved thanks to PR #7676.

lib/rules/function-calc-no-unspaced-operator/index.mjs
ybiquitous: Nice catch. Fixed with c8d7979e8 👍🏼

lib/rules/function-calc-no-unspaced-operator/index.mjs
ybiquitous: To be honest, I'm against this idea because `undefined` in the `ComponentValue | undefined` type is hidden and unclear.

lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
Mouvedia: Look at the [demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0BDABMAOgOzWg9AFRoCiATmRGWofjngC4QAO8aYKANmABQCszAB4BqAMxCAlAG4cAXxw4kAIwz0CxAHIQ0MClRp1caJq3ZdeAkeMFoAtGgBMkmdnnYQAGhAAzAJacYDRQAWzhEGEEQ5gCkMABnOM9wCGw-AHMEECwjTBAyAFcAuNy2bLw8XO987DAGXxTbDm5bbAhbarjmFDAYKFsWXRQmMhLjApg1NzckyFTfNIAxKmChzIArOJSk2GZExDK0XLiGAE8Azl9sBlHcziGYY9yPNSPT88uGRpT022OUbCgKDIUBuIDuDAe1xAchAsiAA) linked in the issue; as you can see it's not detected as invalid right now but `calc(1px + 2px + 3px)` is.
Hence we want `calc(1px+2px + 3px)` way more because it is/was a bug.
i.e. we want to pick an eventual regression

Mouvedia:
Thanks @romainmenke, having such a proactive member adapting our dependencies that fast is invaluable.

romainmenke (approved):
LGTM!

Thank you @ybiquitous for this update 🙇

ybiquitous:
Thanks for the review, all!

PR #7657, opened 2024-04-26
Fix `shorthand-property-no-redundant-values` false negatives for functions

### LINKED ISSUE #7656: Fix `shorthand-property-no-redundant-values` false negatives for `var()`

```css
a { padding: 0 var(--padding) 0; }
```

```json
{ "rules": { "shorthand-property-no-redundant-values": true } }
```

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcCBOEIBcAEwA6A7baAtIQA4CGUUAlrgObwYBMADCQB4DceAvnnmVj5pylGvQzMMANzJIAFMRHU6ASgldcvXCAA0IAGZUANjAByZALZxEMNpZImAdGADOL3eAi5DtBCBy4GBjYILZoMLhQLiEMANp4QUEhLmgAniZGNESQ3lS0hClkkbJQIQkYALo8Hjk+AGIQSBZkaH4AVi5eHrAk7ogBSSAp6TCZuGgxwSBGLTApITrlyWkZWYS1eQVoRVAlkyEz4fMg1dxAA)

```json { "stylelint": "15.6.0" } ```

fix redundant values

### What actually happened?

nothing

--

### PR DESCRIPTION
**Pull request by ybiquitous**

Ref #7656

I think supporting all functions makes sense, including ~~`var()`~~ or `calc()` etc.
-> Dropped `var()` support. See https://github.com/stylelint/stylelint/issues/7656#issuecomment-2079826514

### DIFF
--- a/lib/rules/shorthand-property-no-redundant-values/__tests__/index.mjs
+++ b/lib/rules/shorthand-property-no-redundant-values/__tests__/index.mjs
@@ -78,23 +78,11 @@ testRule({
 			description: 'ignore not shorthandable property',
 		},
 		{
-			code: 'a { border-radius: 1px / 1px }',
+			code: 'a { border-radius: 1px / 1px; }',
 			description: 'ignore ellipse',
 		},
 		{
-			code: 'a { margin: calc(1px + 1px) calc(1px + 1px); }',
-			description: 'ignore calc function',
-		},
-		{
-			code: 'a { margin: some-function(1px + 1px) some-function(1px + 1px); }',
-			description: 'ignore all function',
-		},
-		{
-			code: 'a { margin: var(--margin) var(--margin); }',
-			description: 'ignore variables',
-		},
-		{
-			code: 'a { border-color: #FFFFFF transparent transparent }',
+			code: 'a { border-color: #FFFFFF transparent transparent; }',
 			description: 'ignore upper case value',
 		},
 		{
@@ -378,6 +366,30 @@ testRule({
 			code: 'a { inset: 1px 0 1px 0; }',
 			fixed: 'a { inset: 1px 0; }',
 			message: messages.rejected('1px 0 1px 0', '1px 0'),
+			line: 1,
+			column: 5,
+		},
+		{
+			code: 'a { margin: var(--margin) var(--margin); }',
+			fixed: 'a { margin: var(--margin); }',
+			message: messages.rejected('var(--margin) var(--margin)', 'var(--margin)'),
+			line: 1,
+			column: 5,
+		},
+		{
+			code: 'a { margin: calc(1px + 1px) calc(1px + 1px); }',
+			fixed: 'a { margin: calc(1px + 1px); }',
+			message: messages.rejected('calc(1px + 1px) calc(1px + 1px)', 'calc(1px + 1px)'),
+			line: 1,
+			column: 5,
+		},
+		{
+			code: 'a { margin: some-function(1px + 1px) some-function(1px + 1px); }',
+			fixed: 'a { margin: some-function(1px + 1px); }',
+			message: messages.rejected(
+				'some-function(1px + 1px) some-function(1px + 1px)',
+				'some-function(1px + 1px)',
+			),
 			line: 1,
 			column: 5,
 		},
--- a/lib/rules/shorthand-property-no-redundant-values/__tests__/index.mjs
+++ b/lib/rules/shorthand-property-no-redundant-values/__tests__/index.mjs
@@ -82,6 +82,10 @@ testRule({
 			description: 'ignore ellipse',
 		},
 		{
+			code: 'a { border-radius: 1px 1px / 1px; }',
+			description: 'ignore ellipse',
+		},
+		{
 			code: 'a { border-color: #FFFFFF transparent transparent; }',
 			description: 'ignore upper case value',
 		},
--- a/lib/rules/shorthand-property-no-redundant-values/__tests__/index.mjs
+++ b/lib/rules/shorthand-property-no-redundant-values/__tests__/index.mjs
@@ -86,6 +86,10 @@ testRule({
 			description: 'ignore ellipse',
 		},
 		{
+			code: 'a { margin: var(--margin) var(--margin); }',
+			description: 'ignore variables',
+		},
+		{
 			code: 'a { border-color: #FFFFFF transparent transparent; }',
 			description: 'ignore upper case value',
 		},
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: lib/rules/shorthand-property-no-redundant-values/index.mjs [hunk 1]
```diff
@@ -29,16 +32,6 @@ const propertiesWithShorthandNotation = new Set([
 	'inset',
 ]);
```
Mouvedia, line 33: Can you add the alias of `grid-gap`?
i.e. `gap`

lib/rules/shorthand-property-no-redundant-values/index.mjs
```diff
@@ -122,40 +115,44 @@ const rule = (primary, _secondaryOptions, context) => {
...
 				return;
 			}
 
-			const shortestForm = canCondense(
-				valuesToShorthand[0] || '',
-				valuesToShorthand[1] || '',
-				valuesToShorthand[2] || '',
-				valuesToShorthand[3] || '',
-			);
+			const [top, right, bottom, left] = valuesToShorthand;
```
Mouvedia, line 154:
```suggestion
const [first, second, third, fourth] = valuesToShorthand;
```

e.g. `grid-gap: <'row-gap'> <'column-gap'>`

lib/rules/shorthand-property-no-redundant-values/index.mjs
ybiquitous: Out of scope of the PR. Please add it in another PR?

lib/rules/shorthand-property-no-redundant-values/__tests__/index.mjs
```diff
@@ -78,23 +78,11 @@ testRule({
 			description: 'ignore not shorthandable property',
 		},
 		{
-			code: 'a { border-radius: 1px / 1px }',
+			code: 'a { border-radius: 1px / 1px; }',
 			description: 'ignore ellipse',
 		},
```
Mouvedia, line 83: minor
question

Can you add a test for `border-radius: 1px 1px / 1px;`?

lib/utils/typeGuards.cjs
```diff
@@ -53,6 +53,14 @@ function isDocument(node) {
 	return node.type === 'document';
 }
 
+/**
+ * @param {import('postcss-value-parser').Node} node
+ * @returns {node is import('postcss-value-parser').DivNode}
+ */
+function isValueDiv(node) {
```
Mouvedia, line 60: minor
suggestion

Maybe update 

https://github.com/stylelint/stylelint/blob/68cb920c45641702d76670afb111960cbd743296/lib/rules/font-weight-notation/index.mjs#L188-L190

lib/utils/typeGuards.mjs
```diff
@@ -49,6 +49,14 @@ export function isDocument(node) {
 	return node.type === 'document';
 }
 
+/**
+ * @param {import('postcss-value-parser').Node} node
+ * @returns {node is import('postcss-value-parser').DivNode}
+ */
+export function isValueDiv(node) {
```
Mouvedia, line 56: minor
suggestion

Maybe update 

https://github.com/stylelint/stylelint/blob/68cb920c45641702d76670afb111960cbd743296/lib/rules/font-weight-notation/index.mjs#L188-L190

lib/rules/shorthand-property-no-redundant-values/index.mjs
ybiquitous: Need to be consistent with canCondense arguments. It is another refactoring chance.

lib/utils/typeGuards.mjs
ybiquitous: Out of scope. Feel free to do it in another PR.

lib/rules/shorthand-property-no-redundant-values/index.cjs
```diff
@@ -140,6 +140,16 @@ const rule = (primary, _secondaryOptions, context) => {
 					break;
 				}
 
+				// `var()` should be ignored. E.g.
+				//
+				//   --padding: 20px 5px;
+				//   padding: 0 var(--padding) 0;
+				//
+				if (typeGuards.isValueFunction(valueNode) && valueNode.value.toLowerCase() === 'var') {
+					shouldIgnore = true;
```
Mouvedia, line 149: suggestion
refactor

What about a `return` inside both this `if` and the preceding one?
i.e. `shouldIgnore` flag seems unnecessary

lib/rules/shorthand-property-no-redundant-values/index.mjs
```diff
@@ -138,6 +138,16 @@ const rule = (primary, _secondaryOptions, context) => {
...
 
+				// `var()` should be ignored. E.g.
+				//
+				//   --padding: 20px 5px;
+				//   padding: 0 var(--padding) 0;
+				//
+				if (isValueFunction(valueNode) && valueNode.value.toLowerCase() === 'var') {
+					shouldIgnore = true;
+					break;
+				}
```
Mouvedia, line 149: suggestion
refactor

What about a `return` inside both this `if` and the preceding one?
i.e. `shouldIgnore` flag seems unnecessary

ybiquitous: Nice catch. Fixed via 101003d36 👍🏼

lib/rules/shorthand-property-no-redundant-values/index.mjs
```diff
@@ -122,40 +115,46 @@ const rule = (primary, _secondaryOptions, context) => {
...
+			if (!isStandardSyntaxProperty(prop)) return;
+
+			if (!isStandardSyntaxValue(value)) return;
 
 			const normalizedProp = vendor.unprefixed(prop.toLowerCase());
 
-			if (hasIgnoredCharacters(value) || !isShorthandProperty(normalizedProp)) {
-				return;
-			}
+			if (!isShorthandProperty(normalizedProp)) return;
```
Mouvedia, line 128: nit

rename to `isSupportedShorthand`
else we would have used `longhandSubPropertiesOfShorthandProperties.keys()`

Mouvedia (approved):
LGTM

lib/rules/shorthand-property-no-redundant-values/index.mjs
ybiquitous: Make sense. Done b290aee10

lib/rules/shorthand-property-no-redundant-values/__tests__/index.mjs
```diff
@@ -381,6 +377,23 @@ testRule({
...
+		{
+			code: 'a { margin: some-function(1px + 1px) some-function(1px + 1px); }',
+			fixed: 'a { margin: some-function(1px + 1px); }',
+			message: messages.rejected(
+				'some-function(1px + 1px) some-function(1px + 1px)',
+				'some-function(1px + 1px)',
+			),
+			line: 1,
+			column: 5,
+		},
```
Mouvedia, line 396: @ybiquitous

```suggestion
		},
		{
			code: 'a { margin: 0 foo(1px + 1px) 0; }',
			fixed: 'a { margin: 0 foo(1px + 1px); }',
			message: messages.rejected('0 foo(1px + 1px) 0', '0 foo(1px + 1px)',
			),
			line: 0,
			column: 0,
		},
```

PR #7658, opened 2024-04-27
Fix `value-no-vendor-prefix` false positives/negatives

Mouvedia:

follow-up of #7654

- unprefixed value doesn't exist
  - [autoprefixer](https://github.com/postcss/autoprefixer/blob/main/test/cases/flexbox.out.css), [whatwg](https://compat.spec.whatwg.org/#css-keyword-mappings)
    - `-khtml-box`, `-moz-box` and `-webkit-box`
    - `-moz-inline-box`, `-ms-inline-flexbox` and `-webkit-inline-box`
  - https://caniuse.com/intrinsic-width
    - `-moz-available`
    - `-webkit-fill-available`
- https://caniuse.com/css-crisp-edges
  - Opera only supported `-o-crisp-edges` (i.e. not `-o-pixelated`)
  - `-webkit-optimize-contrast`
- https://caniuse.com/mdn-css_properties_unicode-bidi_plaintext
  - `-webkit-plaintext` was missing

I reviewed each values one by one.

### DIFF
--- a/lib/rules/value-no-vendor-prefix/README.md
+++ b/lib/rules/value-no-vendor-prefix/README.md
@@ -83,4 +83,4 @@ a { -moz-user-select: -moz-all; }
 > [!WARNING]
 > An _exact_ match comparison will be performed for non-regex strings in the next major version.
 > If you want to keep the legacy behavior, please consider using a regex instead.
-> E.g. `[/^(-webkit-|-moz-|-ms-)?max-content$/]`.
+> E.g. `[/^(-webkit-|-moz-)?max-content$/]`.
--- a/lib/rules/value-no-vendor-prefix/__tests__/index.mjs
+++ b/lib/rules/value-no-vendor-prefix/__tests__/index.mjs
@@ -11,6 +11,12 @@ testRule({
 			code: 'a { display: flex; }',
 		},
 		{
+			code: '.foo { display: -webkit-box; }',
+		},
+		{
+			code: '.foo { display: -moz-box; }',
+		},
+		{
 			code: 'a { background: linear-gradient(to top, #000, #fff); }',
 		},
 		{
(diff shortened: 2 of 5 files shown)


lib/rules/value-no-vendor-prefix/README.md
```diff
@@ -83,4 +83,4 @@ a { -moz-user-select: -moz-all; }
 > [!WARNING]
 > An _exact_ match comparison will be performed for non-regex strings in the next major version.
 > If you want to keep the legacy behavior, please consider using a regex instead.
-> E.g. `[/^(-webkit-|-moz-|-ms-)?max-content$/]`.
```
Mouvedia, line 86: typo that slipped in the last PR
see https://caniuse.com/mdn-css_properties_width_max-content

lib/rules/value-no-vendor-prefix/__tests__/index.mjs
```diff
@@ -175,7 +172,7 @@ testRule({
 
 	accept: [
 		{
-			code: 'a { $margin: -webkit-box; }',
+			code: 'a { $foo: -webkit-plaintext; }',
```
Mouvedia, line 175: This was setup so that it could pick regressions, hence the value needs to be one of the unprefixable ones.

lib/rules/value-no-vendor-prefix/README.md
```diff
@@ -83,4 +83,4 @@ a { -moz-user-select: -moz-all; }
...
+> E.g. `[/^(-webkit-|-moz-)?max-content$/]`.
```
Mouvedia, line 86: typo that slipped in the last PR
see https://caniuse.com/mdn-css_properties_width_max-content

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #7670, opened 2024-04-29
Add support for multi-argument functions to `function-calc-no-unspaced-operator`

### LINKED ISSUE #7618: `function-calc-no-unspaced-operator` should support newer calc functions

> [fix the false negatives for `round` and other calc-sum functions by broadening the meaning of calc to calc-sum](https://github.com/stylelint/stylelint/issues/5695#issuecomment-1079960257)

> [It doesn't validate any of the newer math functions.](https://github.com/stylelint/stylelint/issues/7181#issuecomment-1713643664)

compare https://www.w3.org/TR/css-values-4/#calc-syntax and https://www.w3.org/TR/css-values-3/#calc-syntax

I already have it working on a local branch for a subset of these.

- [x] `abs`
- [x] `acos`
- [x] `asin`
- [x] `atan`
- [x] `calc`
- [x] `cos`
- [x] `exp`
- [x] `sign`
- [x] `sin`
- [x] `sqrt`
- [x] `tan`
- [x] `min`
- [x] `max`
- [x] `hypot`
- [x] `clamp`
- [x] `round`
- [x] `mod`
- [x] `rem`
- [x] `atan2`
- [x] `pow`
- [x] `log`

romainmenke:

Closes #7618

<hr />

@ybiquitous This rule and the implementation was a lot more complex that I originally thought it would be. Doing pure code review was tricky for me and I needed to take a deep dive.

_The code in this PR is a mess, but it also isn't intended to be used as is.
It can be used, we can continue work from this. But I haven't spend much time on self-review to get it into a really good state._

As I understand it know there are multiple issues and layers.

Steps:
1. Fix units : `10px-20px` -> `10px -20px` (`px-20px` is the entire unit here)
2. Fix missing operators : `10px -20px` -> `10px -|20px` (`|` to indicate that these are separate)
3. Insert whitespace : `10px -|20px` -> `10px - 20px`
4. Normalize whitespace : `10px -\t20px` -> `10px - 20px`

The csstools tokenizer and parsing packages have several aspects that help with some of these steps.

The tokenizer is just arrays, strings and numbers. So Step 1 can be handled purely on the token stream. It is a messy function, but it gives us more control and nice separation of concerns.

The parsing algorithm walker can be stateful. This is useful to keep track of being in a math context or not e.g. `calc((((10px+2px))))`

In general I try to do as much work and analysis on the parsed values (e.g. `token[4].value`, `token[4].unit`, ...) and as little as possible on the string representation (e.g. `token[1]`, `node.toString()`)

My hope is that doing the same work from different angles is a good way to transfer some knowledge :)

Ref #7655

### DIFF
--- a/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
+++ b/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
@@ -239,6 +239,11 @@ testRule({
 		},
 		{
 			code: 'a { padding: 10px calc([10px+5px]); }',
+			description: 'simple blocks with square brackets are ignored',
+		},
+		{
+			code: 'a { padding: 10px calc({10px+5px}); }',
+			description: 'simple blocks with curly braces are ignored',
 		},
 		{
 			code: 'a { top: calc(5--6px-7px); }',
@@ -620,6 +625,16 @@ testRule({
 			column: 19,
 			endLine: 1,
 			endColumn: 20,
+		},
+		{
+			code: 'a { top: calc(calc(1px+ 1px)); }',
+			fixed: 'a { top: calc(calc(1px + 1px)); }',
+			description: 'nested math expression',
+			message: messages.expectedBefore('+'),
+			line: 1,
+			column: 23,
+			endLine: 1,
+			endColumn: 24,
 		},
 		{
 			code: 'a { top: calc(((1px+ 1px))); }',
--- a/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
+++ b/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
@@ -225,12 +225,6 @@ testRule({
 			code: 'a { top: calc(10px/var(--foo)); }',
 		},
 		{
-			code: 'a { top: calc(10px*var(--foo, 10px 20px)); }',
-		},
-		{
-			code: 'a { top: calc(10px*var(--foo, 10px+20px)); }',
-		},
-		{
 			code: 'a { padding: calc(); }',
 		},
 		{
@@ -249,6 +243,12 @@ testRule({
 		{
 			code: 'a { top: calc(5--6px-7px); }',
 			description: 'custom units',
+		},
+		{
+			code: 'a { top: calc(10px*var(--foo, 10px 20px)); }',
+		},
+		{
+			code: 'a { top: calc(10px*var(--foo, 10px+20px)); }',
 		},
 	],
 
--- a/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
+++ b/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
@@ -687,57 +687,57 @@ testRule({
 	],
 });
 
-// testRule({
-// 	ruleName,
-// 	config: [true],
-// 	customSyntax: 'postcss-scss',
-// 	fix: true,
+testRule({
+	ruleName,
+	config: [true],
+	customSyntax: 'postcss-scss',
+	fix: true,
 
-// 	accept: [
-// 		{
-// 			code: 'a { top: calc(100%*#{$foo}); }',
-// 		},
-// 		{
-// 			code: 'a { top: calc(100% *#{$foo}); }',
-// 		},
-// 		{
-// 			code: 'a { top: calc(100%* #{$foo}); }',
-// 		},
-// 		{
-// 			code: 'a { top: calc(100% * #{$foo}); }',
-// 		},
-// 		{
-// 			code: 'a { top: calc(100%/#{$foo}); }',
-// 		},
-// 		{
-// 			code: 'a { top: calc(100% /#{$foo}); }',
-// 		},
-// 		{
-// 			code: 'a { top: calc(100%/ #{$foo}); }',
-// 		},
-// 		{
-// 			code: 'a { top: calc(100% / #{$foo}); }',
-// 		},
-// 	],
+	accept: [
+		{
+			code: 'a { top: calc(100%*#{$foo}); }',
+		},
+		{
+			code: 'a { top: calc(100% *#{$foo}); }',
+		},
+		{
+			code: 'a { top: calc(100%* #{$foo}); }',
+		},
+		{
+			code: 'a { top: calc(100% * #{$foo}); }',
+		},
+		{
+			code: 'a { top: calc(100%/#{$foo}); }',
+		},
+		{
+			code: 'a { top: calc(100% /#{$foo}); }',
+		},
+		{
+			code: 'a { top: calc(100%/ #{$foo}); }',
+		},
+		{
+			code: 'a { top: calc(100% / #{$foo}); }',
+		},
+	],
 
-// 	reject: [
-// 		{
-// 			code: 'a { top: calc(100%- #{$foo}); }',
-// 			fixed: 'a { top: calc(100% - #{$foo}); }',
-// 			message: messages.expectedBefore('-'),
-// 			line: 1,
-// 			column: 19,
-// 			endLine: 1,
-// 			endColumn: 20,
-// 		},
-// 		{
-// 			code: 'a { top: calc(100% -#{$foo}); }',
-// 			fixed: 'a { top: calc(100% - #{$foo}); }',
-// 			message: messages.expectedAfter('-'),
-// 			line: 1,
-// 			column: 20,
-// 			endLine: 1,
-// 			endColumn: 21,
-// 		},
-// 	],
-// });
+	reject: [
+		{
+			code: 'a { top: calc(100%- #{$foo}); }',
+			fixed: 'a { top: calc(100% - #{$foo}); }',
+			message: messages.expectedBefore('-'),
+			line: 1,
+			column: 19,
+			endLine: 1,
+			endColumn: 20,
+		},
+		{
+			code: 'a { top: calc(100% -#{$foo}); }',
+			fixed: 'a { top: calc(100% - #{$foo}); }',
+			message: messages.expectedAfter('-'),
+			line: 1,
+			column: 20,
+			endLine: 1,
+			endColumn: 21,
+		},
+	],
+});
--- a/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
+++ b/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
@@ -195,6 +195,10 @@ testRule({
 			description: 'CRLF newline and tab before operator',
 		},
 		{
+			code: 'a { top: calc(5--6px-7px); }',
+			description: 'custom units',
+		},
+		{
 			code: 'a { padding: 10px calc(calc(1px + 2px)* 3px); }',
 		},
 		{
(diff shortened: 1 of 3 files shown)


lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
```diff
@@ -586,60 +599,145 @@ testRule({
...
+			],
 		},
 	],
 });
+
+// testRule({
+// 	ruleName,
+// 	config: [true],
+// 	customSyntax: 'postcss-scss',
+// 	fix: true,
```
romainmenke, line 694: I ignored this aspect of this rule purely to keep things a bit simpler.

Implementing this is definitely possible.

lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
```diff
@@ -586,60 +599,145 @@ testRule({
...
+			line: 1,
+			column: 20,
+			endLine: 1,
+			endColumn: 21,
 		},
 		{
-			code: 'a { top: calc(100% /#{$foo}); }',
+			code: 'a { font-size: clamp(1rem, 2.5vw, 1rem+1rem); }',
+			fixed: 'a { font-size: clamp(1rem, 2.5vw, 1rem + 1rem); }',
+			description: 'multiple argument functions',
```
Mouvedia, line 637: Once this is ready to review add

```md

Close #7618
```

lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
```diff
@@ -225,17 +229,13 @@ testRule({
...
-			code: 'a { font-size: clamp(1rem, 2.5vw, 1rem+1rem); }',
-			description:
-				'multiple argument: functions that accept more than one argument are not supported yet',
+			code: 'a { top: calc(10px*var(--foo, 10px+20px)); }',
 		},
 		{
-			code: 'a { width: min(25vw+25vw); }',
-			description:
-				'single argument: functions that accept more than one argument are not supported yet',
+			code: 'a { padding: calc(); }',
```
Mouvedia, line 238: Can you keep the original order of the examples to help with the review?

romainmenke: Let me check!

Mouvedia:
If you can't make a benchmark, do you feel confident that your new approach is faster or at least as fast as the current one?

lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
romainmenke: Should be better now.

Mouvedia: e.g.  l635

romainmenke:
> If you can't make a benchmark, do you feel confident that your new approach is faster or at least as fast as the current one?

This change:

```
;node benchmark-rule.mjs function-calc-no-unspaced-operator 'true'
Warnings: 0
Mean: 92.64952612499998 ms
Deviation: 11.291622367245036 ms
```

Main :

```
;node benchmark-rule.mjs function-calc-no-unspaced-operator 'true'
Warnings: 0
Mean: 78.32598956249998 ms
Deviation: 8.242243635683067 ms
```

`function-calc-no-unspaced-operator-bugfix` :

```
;node benchmark-rule.mjs function-calc-no-unspaced-operator 'true'
Warnings: 0
Mean: 96.32743500000002 ms
Deviation: 9.753154793016495 ms
```

These results match my expectations.
This rule is doing more work than before, so it will be slightly slower.

Mouvedia:
> so it will be _slightly_ slower

Great.  
I am sure we will find some ways to gain some ms during the review.

> My hope is that doing the same work from different angles is a good way to transfer some knowledge :)

I think that once the other PR is merged, you should fix the conflicts and continue your effort.

romainmenke:
Thank you @Mouvedia for the feedback 🙇

Mouvedia:
I didn't check the code yet :)
Personally I don't care about the approach as long as it passes the current tests and if we are gaining something while we are losing on the perf side. The tests are looking good.
The only thing I would like is for both PRs to be part of the same version.

Feel free to switch from draft to ready if you want a proper review.

ybiquitous (commented):
@romainmenke Thanks a lot! I feel this PR token-based approach is much better than the current character-based approach. 👍🏼 

So, as @Mouvedia commented, can you fix the conflict and make this PR ready for review? Since this PR will be merged into PR #7655, I will hand over refactoring in #7655 after merging this PR.

ybiquitous:
Or, if you don't mind, I can take on this issue.

romainmenke:
@ybiquitous I've made a few more edits:
- bring back support for `scss` interpolation
- improve coverage
- improve support for comments (e.g. `calc(1px/* a comment */+ 2px)`)

Feel free to take this!

lib/rules/function-calc-no-unspaced-operator/index.cjs
```diff
@@ -65,169 +62,301 @@ const rule = (primary, _secondaryOptions, context) => {
...
+				// These might be typo's.
+				// For example: `10px-20px` has a unit of `px-20px`
+				for (let i = 0; i < tokens.length; i++) {
+					const token = tokens[i];
 
-					if (whitespace === ' ') return;
+					// NOTE:
+					// This is a mess, I am fairly certain that it is correct, but too much is happening with strings.
+					// This will be brittle and might break when the css-tokenizer is updated.
+					// It will also produce invalid CSS for exotic units, nu such units exists today.
```
Mouvedia, line 79: typo

nu >  no

<hr/>

question

> nu such units exists today

There's `rpx` which is a weapp unit.
e.g.
```
    'unit-no-unknown': [
      true,
      {
        ignoreUnits: ['rpx'],
      },
    ],
```

Should we add a test?

lib/rules/function-calc-no-unspaced-operator/index.cjs
```diff
@@ -65,169 +62,301 @@ const rule = (primary, _secondaryOptions, context) => {
...
+					});
 
-				const delimToken = getDelimToken(operatorNode);
+					tokens.splice(i + 1, 0, ...remainderTokens);
+				}
 
-				if (!delimToken) return;
+				// Step 1.2
+				// Re-tokenize scss interpolation blocks
+				// Grouping `#` and `{` into a single token allows us to parse these as simple blocks with curly braces.
```
Mouvedia, line 110: question

Would less be supported just by doing the same thing for `@`?
e.g. `@{variable}`

lib/rules/function-calc-no-unspaced-operator/index.cjs
```diff
@@ -65,169 +62,301 @@ const rule = (primary, _secondaryOptions, context) => {
...
+							token[4].unit = token[4].unit.slice(0, -1);
+							token[1] = token[1].slice(0, -1);
 						}
+
+						return;
 					}
+				}
 
-					const afterOperator = operand.charAt(operatorIndex + 1);
+				if (
```
Mouvedia, line 221: nit

Maybe add some const declarations?
e.g.
```js
const bar = operation.secondOperand?.value[0];
const foo = bar === cssTokenizer.TokenType.Dimension || bar === cssTokenizer.TokenType.Percentage || bar === cssTokenizer.TokenType.Number;
```

lib/rules/function-calc-no-unspaced-operator/index.cjs
```diff
@@ -28,9 +26,8 @@ const meta = {
 
 const OPERATORS = new Set(['+', '-']);
 const OPERATOR_REGEX = /[+-]/;
-const ALL_OPERATORS = new Set([...OPERATORS, '*', '/']);
 // #7618
```
Mouvedia, line 29: You can remove this comment now.

lib/rules/function-calc-no-unspaced-operator/index.cjs
```diff
@@ -236,90 +365,184 @@ const rule = (primary, _secondaryOptions, context) => {
...
 	if (!cssParserAlgorithms.isTokenNode(node)) return false;
 
-	return operators.has(getDelimToken(node)?.value ?? '');
+	return OPERAND_TOKEN_TYPES.has(node.value[0]);
 }
 
 /**
- * @param {import('@csstools/css-parser-algorithms').WhitespaceNode} node
- * @returns {string}
+ * NOTE: this is messy.
```
Mouvedia, line 397:
```js
/** @typedef {import ('@csstools/css-parser-algorithms').ComponentValue} ComponentValue */
```
will help.

Mouvedia (commented):
I did a first quick pass; nothing major.
This is well documented except for the token positions but that can't be helped.
e.g. `token[4]`
If a regression slipped by I couldn't tell but I think our test suite would catch it.

Mouvedia:
Can you remove the `export` from https://github.com/stylelint/stylelint/blob/f87f784a429f626e67710ab1a8026c0725414a3b/lib/reference/functions.mjs#L29?

lib/rules/function-calc-no-unspaced-operator/index.cjs
romainmenke: This is an exotic unit : `1🧛🏼`

```
'1🧛🏼'.length // 5
```

However, it seems that I did end up wiring everything up correctly and these just work.
I will clarify the comments and add tests

romainmenke: And this is also an exotic unit `1\\23`.
This is not working correctly.

But again, no such units exist today.

lib/rules/function-calc-no-unspaced-operator/index.cjs
Mouvedia: > So anything that can easily become such a simple block is also easy to ignore.

If someone ever request it, we will.

lib/rules/function-calc-no-unspaced-operator/index.cjs
romainmenke: thanks

lib/rules/function-calc-no-unspaced-operator/index.cjs
ybiquitous: > But I am unsure if we want to increase our exposure to non-standard syntaxes.

For your reference, we want to support *only* standard syntax. Supporting non-standard syntaxes should be exceptional, e.g. for backward compatibility.

See also the [document](https://stylelint.io/developer-guide/rules#define-the-rule):

> for standard CSS syntax only

lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
```diff
@@ -247,6 +237,19 @@ testRule({
 		{
 			code: 'a { padding: calc(1px + 2- ); }',
 		},
+		{
+			code: 'a { padding: 10px calc([10px+5px]); }',
```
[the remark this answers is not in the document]
romainmenke: They are not used for anything, which is why they must be ignored.

`[]`, `{}` and `()` are all the same structure in CSS.
They are all simple blocks.

So the parsing algorithms will happily visit `calc((10px + 20px))` and `calc([10px + 20px])` as if these are the same thing.

This test ensures that the rule only handles `()` and ignores `[]`, `{}`.
I've added one more test with `{}`.

lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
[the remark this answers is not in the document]
romainmenke: Done, thanks

romainmenke:
I've made a few edits, but will defer to @ybiquitous now for the remainder of the work on this rule :)

ybiquitous:
@romainmenke Thank you for your work! I'll take over this. 👍🏼

Mouvedia:
Remember to change the author in the 16.X.0 changelog.

PR #7578, opened 2024-03-26
Fix import errors for configs and plugins omitting `/index.js`

ota-meshi:
> Which issue, if any, is this issue related to?

Closes #7576
Closes #7577

Ref #7532

> Is there anything in the PR that needs further explanation?

This PR fixes an issue where cannot resolve paths in configurations that expect `index.js` to resolve implicitly by Node.js.

PR #7675, opened 2024-05-03
Refactor to slightly simplify the code

Mouvedia:

N/A

- eslint-plugin-perf-standard/check-function-inline for 3 functions
- 2 try…catch refactors
- 1 typo

### DIFF
--- a/lib/augmentConfig.mjs
+++ b/lib/augmentConfig.mjs
@@ -150,7 +150,8 @@ function absolutizePaths(config, configDir, cwd) {
 		config.ignoreFiles = [config.ignoreFiles].flat().map((glob) => absolutizeGlob(glob, configDir));
 	}
 
-	const cb = (/** @type { any } */ lookup) => {
+	/** @type {<T>(lookup: T) => (string | T)} */
+	const toAbsolutePath = (lookup) => {
 		if (typeof lookup === 'string') {
 			return getModulePath(configDir, lookup, cwd);
 		}
@@ -159,11 +160,11 @@ function absolutizePaths(config, configDir, cwd) {
 	};
 
 	if (config.plugins) {
-		config.plugins = [config.plugins].flat().map(cb);
+		config.plugins = [config.plugins].flat().map(toAbsolutePath);
 	}
 
 	if (config.processors) {
-		config.processors = config.processors.map(cb);
+		config.processors = config.processors.map(toAbsolutePath);
 	}
 
 	return config;
--- a/lib/augmentConfig.mjs
+++ b/lib/augmentConfig.mjs
@@ -150,24 +150,20 @@ function absolutizePaths(config, configDir, cwd) {
 		config.ignoreFiles = [config.ignoreFiles].flat().map((glob) => absolutizeGlob(glob, configDir));
 	}
 
-	if (config.plugins) {
-		config.plugins = [config.plugins].flat().map((lookup) => {
-			if (typeof lookup === 'string') {
-				return getModulePath(configDir, lookup, cwd);
-			}
+	const cb = (/** @type { any } */ lookup) => {
+		if (typeof lookup === 'string') {
+			return getModulePath(configDir, lookup, cwd);
+		}
 
-			return lookup;
-		});
+		return lookup;
+	};
+
+	if (config.plugins) {
+		config.plugins = [config.plugins].flat().map(cb);
 	}
 
 	if (config.processors) {
-		config.processors = config.processors.map((lookup) => {
-			if (typeof lookup === 'string') {
-				return getModulePath(configDir, lookup, cwd);
-			}
-
-			return lookup;
-		});
+		config.processors = config.processors.map(cb);
 	}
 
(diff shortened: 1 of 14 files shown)


lib/augmentConfig.mjs
```diff
@@ -150,24 +150,20 @@ function absolutizePaths(config, configDir, cwd) {
 		config.ignoreFiles = [config.ignoreFiles].flat().map((glob) => absolutizeGlob(glob, configDir));
 	}
 
-	if (config.plugins) {
-		config.plugins = [config.plugins].flat().map((lookup) => {
-			if (typeof lookup === 'string') {
-				return getModulePath(configDir, lookup, cwd);
-			}
+	const cb = (/** @type { any } */ lookup) => {
```
Mouvedia, line 153: not sure how to get rid of the `any`

lib/utils/getCacheFile.mjs
```diff
@@ -33,15 +33,9 @@ export default function getCacheFile(cacheFile, cwd) {
...
-	try {
-		fileStats = lstatSync(resolvedCacheFile);
-	} catch {
-		fileStats = null;
-	}
-
-	if (looksLikeADirectory || (fileStats && fileStats.isDirectory())) {
+	if (looksLikeADirectory || fileStats?.isDirectory()) {
 		// Return path to provided directory with generated file name.
 		return getCacheFileForDirectory();
```
ybiquitous, line 40: [suggest] I'm not sure why `getCacheFileForDirectory` was created, but it seems we can inline this function. Any thoughts?

```suggestion
		return join(resolvedCacheFile, `.stylelintcache_${hash(cwd)}`);
```

lib/utils/getFileIgnorer.mjs
```diff
@@ -25,15 +23,16 @@ export default function getFileIgnorer({ ignorePath, ignorePattern, cwd }) {
...
 
-		try {
-			const ignoreText = readFileSync(absoluteIgnoreFilePath, 'utf8');
+		if (!existsSync(absoluteIgnoreFilePath)) continue;
+
+		const ignoreText = readFileSync(absoluteIgnoreFilePath, {
+			// utf must remain lowercased to hit the fast path
+			// see nodejs/node#49888
+			encoding: 'utf8',
+			flag: 'rs',
```
ybiquitous, line 32: [ask] Just curious, what difference is between `rs` and unspecified for `readFileSync`? The [Node.js doc](https://nodejs.org/api/fs.html#file-system-flags) says:

> `'rs'`: Open file for reading in synchronous mode. An exception occurs if the file does not exist.

The default flag is `'r'`. Ref: https://nodejs.org/api/fs.html#fsreadfilesyncpath-options

ybiquitous (commented):
@Mouvedia Thanks for the refactoring. Overall LGTM, but can you check my review comments?

lib/utils/getCacheFile.mjs
```diff
@@ -33,15 +33,9 @@ export default function getCacheFile(cacheFile, cwd) {
...
+	const fileStats = lstatSync(resolvedCacheFile, { throwIfNoEntry: false });
 
-	try {
-		fileStats = lstatSync(resolvedCacheFile);
-	} catch {
-		fileStats = null;
-	}
-
-	if (looksLikeADirectory || (fileStats && fileStats.isDirectory())) {
+	if (looksLikeADirectory || fileStats?.isDirectory()) {
```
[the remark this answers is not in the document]
Mouvedia: > I didn't know the `throwIfNoEntry` option

Actually [you did](https://github.com/stylelint/stylelint/pull/7578#discussion_r1538509311) :)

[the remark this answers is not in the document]
Mouvedia: > I didn't know the `throwIfNoEntry` option

Actually [you did](https://github.com/stylelint/stylelint/pull/7578#discussion_r1538509311) :)

lib/utils/getCacheFile.mjs
Mouvedia: I am trying to limit the diff to the minimum.
It's weird, I agree, but it might have a use if exported.
e.g. check l517 of standalone-cache.test.mjs

lib/utils/getFileIgnorer.mjs
Mouvedia: I just wanted to be specific about it.
In this case it might not make much difference with the default.
I shouldn't change this if not backed up by a benchmark.

lib/utils/getCacheFile.mjs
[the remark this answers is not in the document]
ybiquitous: Oh, I completely forgot it. 😅

lib/utils/getCacheFile.mjs
ybiquitous: Okay, no problem 👍🏼

ybiquitous (approved):
Thanks, looks great to me 😄

PR #7676, opened 2024-05-05
Fix `function-calc-no-unspaced-operator` unit and ident escaping

romainmenke:

See : https://github.com/stylelint/stylelint/pull/7655

- Switch to upstream type guards
- Use the ident and unit mutation helpers from `@csstools/css-tokenizer`
- Also support trailing `-` found in ident tokens `calc(g- 10px)`


ybiquitous (approved):
@romainmenke Thanks for the nice addition of `mutateUnit` and many type guards! 👍🏼

PR #7685, opened 2024-05-10
Fix `selector-max-specificity` end positions

romainmenke:

No, it's self-explanatory.


romainmenke:
I will take a look at the coverage issue shortly.

lib/rules/selector-max-specificity/__tests__/index.mjs
```diff
@@ -260,47 +260,47 @@ testRule({
...
 		{
 			code: '.cd { .de { .fg {} } }',
-			message: messages.expected('.cd .de .fg', '0,2,1'),
+			message: messages.expected('.fg', '0,2,1'),
 			line: 1,
 			column: 13,
 			endLine: 1,
 			endColumn: 16,
 		},
 		{
```
Mouvedia, line 269: minor

Can you add `.abc { & & {} }`?
i.e. `.abc .abc .abc {}` fails too

romainmenke: We don't try to determine from which part of a selector the limit was exceeded. We always compare the final specificity of a complete selector agains the user config.

So in your example the message would contain `& &`.

lib/rules/selector-max-specificity/__tests__/index.mjs
```diff
@@ -65,6 +65,9 @@ testRule({
...
 			code: ':nth-child(2n + 1) {}',
 		},
+		{
+			code: '[value] {}',
```
Mouvedia, line 69: nit

It's kind of contradictory.
i.e. that's the attribute name not its value

ybiquitous (approved):
Thanks, LGTM 👍🏼

lib/rules/selector-max-specificity/index.mjs
```diff
@@ -234,35 +232,27 @@ const rule = (primary, secondaryOptions) => {
...
-
-					if (!selectorTree) continue;
+			if (!isStandardSyntaxRule(ruleNode)) return;
 
+			flattenNestedSelectorsForRule(ruleNode, result).forEach(({ selector, resolvedSelectors }) => {
+				resolvedSelectors.forEach((resolvedSelector) => {
 					// Check if the selector specificity exceeds the allowed maximum
-					if (compare(maxChildSpecificity(selectorTree), maxSpecificity) > 0) {
+					if (compare(nodeSpecificity(resolvedSelector), maxSpecificity) > 0) {
+						const index = selector.first?.sourceIndex ?? 0;
```
ybiquitous, line 241: [note] The coverage diff is caused by this line, but I think it's ignorable.

If improving the coverage, we might change it like this, but I'm unsure it's really safe...

```suggestion
						const index = selector.first.sourceIndex;
```

romainmenke: Yeah, I think it is safe in practice, but I'd rather keep it anyway.

If the AST is mutated without setting `sourceIndex` or if something clears out a selector it would throw errors.

But maybe we can add a helper so that we can document this better?

ybiquitous: I am fine with keeping it as-is, and I agree with adding a helper. 👍🏼

romainmenke: I should actually fix this one : https://github.com/postcss/postcss-selector-parser/issues/287

Then we change `const index = selector.first?.sourceIndex ?? 0` to `const index = selector.sourceIndex`

ybiquitous: nice👍

Issue #4189, opened 2019-08-07
no-descending-specificity trips up on pseudo-elements


Zirak:
> Clearly describe the bug

When choosing a selector to compare specificity to, `no-descending-specificity` discards any pseudo-classes like `last-child` before finding a selector to compare to. This seems sort of related to #2489, but from what I can tell that issue talks about more complex selectors where relationships may be tricky to define, while this has a more straightforward repro (see below).

If the pseudo-class was a regular class, I think we could agree that there'd be no reason to emit a warning, and intuitively seems like it should exhibit the same behaviour.

> Which rule, if any, is the bug related to?

`no-descending-specificity`

> What CSS is needed to reproduce the bug?

In `/tmp/blah.css`:

```css
.foo > span:last-child {
    color: #abc;
}

.bar > span {
    color: #fed;
}
```

> What stylelint configuration is needed to reproduce the bug?

In `/tmp/blah.stylelintrc.json`:

```json
{
    "rules": {
        "no-descending-specificity": true
    }
}
```

> Which version of stylelint are you using?

Reproduced in master:
```sh
$ ./bin/stylelint.js --version
10.1.0
```

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

CLI from master, `./bin/stylelint.js --config /tmp/blah.stylelintrc.json /tmp/blah.css`

> What did you expect to happen?

No warnings to be flagged. The output should (imo) be equivalent to linting the following css:

```css
.foo > span.something-else {
    color: #abc;
}

.bar > span {
    color: #fed;
}
```

Whose output it empty.

> What actually happened (e.g. what warnings or errors did you get)?

```shell
/tmp/blah.css
 5:1  ✖  Expected selector ".bar > span" to come before selector ".foo > span:last-child"   no-descending-specificity
```

nosilleg:
@zirak Thanks for the details in your issue, it makes things a lot easier to check.

The example you have provided works as intended. The specificity of those selectors is 0-2-1 and 0-1-1. Therefore the second is less specific and should come first.  The linting error is correct.

See https://www.codecaptain.io/tools/css-specificity-calculator for confirmation.

Zirak:
@nosilleg My pleasure, thank you for taking a look !

You're right that the selector with the pseudo-class has a higher specificity, however so does the selector with a regular class. My intuition was that they'd be treated similarly. To better elaborate on what I mean:

```css
/* A: specificity 0-2-1 */
.foo span:last-child { ... }
[...]

PR #7688, opened 2024-05-11
Refactor `selector-max-specificity` tests to align with standard CSS

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/7495

As explained in the linked issue `:global()` and `:local()` are non-standard functional pseudo classes that have these properties:
- they are non-standard
- have no specificity of their own
- do not support selector lists, only a single selector at a time (i.e. `:local(a), :local(b)` and not `:local(a, b)`)

But they were used in tests to verify the `ignoreSelectors` feature.
This feature acts like a hole punch. Only the ignored node is skipped when calculating specificity. The child nodes are part of the equation.

Given that `:global` and `:local` have no specificity of their own and do not allow lists, it isn't ideal that we are using these to test `ignoreSelectors`. All of the complexity for `ignoreSelectors` is in calculating the specificity of lists of child selectors.

This change replaces `:global` and `:local` with `:is` and `:has` because these do support lists of selectors (e.g. `:is(a, b)`).

It also adds tests with `:host` because `:host` does have its own specificity.

This gives us meaningful coverage for `ignoreSelectors` in a way that aligns with standard CSS.

------------

The goal of this change is to have coverage in a way that makes it easier to change the implementation of this rule.

The goal is not to make breaking changes that are noticeable to end users, only to make sure that we are testing the right things in the right way.

------------

Side note:

Calculating correct specificity for `:global`, `:local` or any other selector is purely an upstream concern in `@csstools/selector-specificity`. So I think it is fine if we don't have explicit coverage for these (non-standard) selectors.

We need coverage for the rule and for `ignoreSelectors` here, but not for the specificity of any individual selector.

### DIFF
--- a/lib/rules/selector-max-specificity/__tests__/index.mjs
+++ b/lib/rules/selector-max-specificity/__tests__/index.mjs
@@ -439,22 +439,25 @@ testRule({
 	config: [
 		'0,1,0',
 		{
-			ignoreSelectors: [':global', ':local', '/my-/'],
+			ignoreSelectors: [':is', ':has', ':host', '/my-/'],
 		},
 	],
 
 	accept: [
 		{
-			code: ':global(.b) {}',
+			code: ':is(.b) {}',
 		},
 		{
-			code: ':global(.b, :local(.c)) {}',
+			code: ':is(.b, :has(.c)) {}',
 		},
 		{
-			code: ':local(.b) {}',
+			code: ':has(.b) {}',
 		},
 		{
-			code: ':local(.b, :global(.c)) {}',
+			code: ':has(.b, :is(.c)) {}',
+		},
+		{
+			code: ':has(.b, :is(a a)) {}',
 		},
 		{
 			code: 'my-tag.a {}',
@@ -462,48 +465,75 @@ testRule({
 		{
 			code: ':nth-child(even of my-tag) {}',
 		},
+		{
+			code: '.a:host {}',
+		},
+		{
+			code: '.a:host(:where(.b)) {}',
+		},
+		{
+			code: ':host(.a) {}',
+		},
 	],
 
 	reject: [
 		{
-			code: '.a:global(.b) {}',
-			message: messages.expected('.a:global(.b)', '0,1,0'),
+			code: '.a:is(.b) {}',
+			message: messages.expected('.a:is(.b)', '0,1,0'),
 			line: 1,
 			column: 1,
 		},
 		{
-			code: '.a:global(.b, .c) {}',
-			message: messages.expected('.a:global(.b, .c)', '0,1,0'),
+			code: '.a:is(.b, .c) {}',
+			message: messages.expected('.a:is(.b, .c)', '0,1,0'),
 			line: 1,
 			column: 1,
 		},
 		{
-			code: ':global(.b, .c.d) {}',
-			message: messages.expected(':global(.b, .c.d)', '0,1,0'),
+			code: ':is(.b, .c.d) {}',
+			message: messages.expected(':is(.b, .c.d)', '0,1,0'),
 			line: 1,
 			column: 1,
 		},
 		{
-			code: '.a:local(.b) {}',
-			message: messages.expected('.a:local(.b)', '0,1,0'),
+			code: '.a:has(.b) {}',
+			message: messages.expected('.a:has(.b)', '0,1,0'),
 			line: 1,
 			column: 1,
 		},
 		{
-			code: '.a:local(.b, .c) {}',
-			message: messages.expected('.a:local(.b, .c)', '0,1,0'),
+			code: '.a:has(.b, .c) {}',
+			message: messages.expected('.a:has(.b, .c)', '0,1,0'),
 			line: 1,
 			column: 1,
 		},
 		{
-			code: ':local(.b, .c.d) {}',
-			message: messages.expected(':local(.b, .c.d)', '0,1,0'),
+			code: ':has(.b, .c.d) {}',
+			message: messages.expected(':has(.b, .c.d)', '0,1,0'),
 			line: 1,
 			column: 1,
 		},
 		{
 			code: 'my-tag.a.b {}',
 			message: messages.expected('my-tag.a.b', '0,1,0'),
+			line: 1,
+			column: 1,
+		},
+		{
+			code: 'a.b:host {}',
+			message: messages.expected('a.b:host', '0,1,0'),
+			line: 1,
+			column: 1,
+		},
+		{
+			code: 'a:host(.b) {}',
+			message: messages.expected('a:host(.b)', '0,1,0'),
+			line: 1,
+			column: 1,
+		},
+		{
+			code: ':host(a.b) {}',
+			message: messages.expected(':host(a.b)', '0,1,0'),
 			line: 1,
 			column: 1,
 		},
@@ -527,8 +557,8 @@ testRule({
 
 	reject: [
 		{
-			code: '.a:global(.b) {}',
-			message: messages.expected('.a:global(.b)', '0,1,0'),
+			code: '.a:is(.b) {}',
+			message: messages.expected('.a:is(.b)', '0,1,0'),
 			line: 1,
 			column: 1,
 		},


lib/rules/selector-max-specificity/__tests__/index.mjs
Mouvedia, line 1: question
minor

Should we update the README accordingly?
see https://github.com/stylelint/stylelint/blob/main/lib/rules/selector-max-specificity/README.md?plain=1#L95-L145

Mouvedia:
Can you update

https://github.com/stylelint/stylelint/blob/22f288402c87bb33fb93f834395ae008aa4f103a/lib/rules/no-descending-specificity/__tests__/index.mjs#L84

so that it uses either `:host()`, `:is()` or `:has()` instead?

lib/rules/selector-max-specificity/__tests__/index.mjs
romainmenke: Good catch, yes, I would also update it.

It could be misunderstood that you need to set this config if you want to use `:local` or `:global`, which isn't true.

See these examples : 

without `ignoreSelectors`
https://stylelint.io/demo/#N4Igxg9gJgpiBcCDmAbCAjAhigFAOkwEoACYAXwB0A7a+NMbfI0ymq+VDRzPdE86rXrde-MiAA0IAGYBLFDABymALZxEMAB6qADgrxgAzocngIVOUgQhg1YsQogATgFcFhx-FJ37DkIZgFMAAXCCcAWhVMTXDDHRgwWTlE4IBPT2IAbUcABgkARgkcxwBdH1ZxKUgLWSQAMTCo4OsAK0NzU1gdE0RbKntHQzSFFFkqYIzHFExgmCHHCR9B4cCx4PDqy1jgzCooTCcoSZBp2fmQanEyIA

with `ignoreSelectors`
https://stylelint.io/demo/#N4Igxg9gJgpiBcCDmAbCAjAhigFAOkwEoACYAXwB0A7a+NMbfI0ymq+VDRzPdE86rXrde-MiAA0IAGYBLFDABymALZxEMAB6qADgrxgAzocngIVOUgQhg1YsQogATgFcFhx-FJ37DkIZgFMAAXCCcAWhVMTXDDHRgwWTlE4IBPT2IAbUcABgkARgkcxwlSP1kkKjCYAGVAhNCnDwQsz04sFBK-OggGTpAAXWIyAZ9WcSlICwqAMTCo4OsAK0NzU1gdE0RbKntHQzSFFFkqYIzHFExgmAOSn33DwJPg8KnLWODMKihMJyhzkCXa63EDUcRkIA

lib/rules/selector-max-specificity/README.md
```diff
@@ -48,7 +48,7 @@ The following patterns are considered problems:
 ```css
 .foo {
   color: red;
-  @nest .baz .bar & {
```
romainmenke, line 51: `@nest` doesn't exist anymore.

I don't mind splitting this out into a separate PR.

Mouvedia: It was part of the a specification at one time. Let's not touch it for now.

romainmenke: let's keep both for now, to make the change easier to review :)

lib/rules/no-descending-specificity/__tests__/index.mjs
Mouvedia, line 1: off-topic
remark

I suppose that `:has()` is already covered by `:not()` or `:is()` test cases.

romainmenke: Yes, `:has` has the same properties as `:not` and `:is` in the context of these rules.

lib/rules/selector-max-specificity/README.md
```diff
@@ -109,17 +109,17 @@ The following patterns are _not_ considered problems:
...
 ```css
-:local(.foo.bar) {}
+:has(.foo.bar) {}
 ```
 
 <!-- prettier-ignore -->
 ```css
-:local(.foo, :global(.bar).baz) {}
+:has(.foo, :is(.bar).baz) {}
 ```
```
Mouvedia, line 123: minor
remark

These examples make no sense to me since

> The :has() pseudo-class takes on the specificity of the most specific selector in its arguments the same way as :is() and :not() do.

it should regardless of the `ignoreSelectors`.
Are those even relevant examples?

romainmenke: Ah, yes, needs to be something like `:host` which does have its own specificity.

lib/rules/selector-max-specificity/__tests__/index.mjs
```diff
@@ -439,65 +439,77 @@ testRule({
 	config: [
 		'0,1,0',
 		{
-			ignoreSelectors: [':global', ':local', '/my-/'],
+			ignoreSelectors: [':is', ':has', ':host', '/my-/'],
```
Mouvedia, line 442: question

Do we have something covering https://chromium.googlesource.com/chromium/src/+/674ef6ac2499fb51d82e752ccf723fd76c505caf/third_party/WebKit/LayoutTests/external/wpt/css/css-scoping/host-context-specificity-003.html?
ref https://chromestatus.com/feature/5177305701810176

lib/rules/selector-max-specificity/README.md
```diff
@@ -109,17 +109,17 @@ The following patterns are _not_ considered problems:
...
 <!-- prettier-ignore -->
 ```css
-:local(.foo.bar) {}
+:host-context(.foo.bar) {}
 ```
 
 <!-- prettier-ignore -->
 ```css
-:local(.foo, :global(.bar).baz) {}
+:host-context(.foo, :host(.bar).baz) {}
```
Mouvedia, line 122: question
typo

Should it be `:host-context(.foo, :host(.bar), .baz) {}` instead?

lib/rules/selector-max-specificity/README.md
```diff
@@ -131,17 +131,17 @@ The following patterns are considered problems:
...
 <!-- prettier-ignore -->
 ```css
-:local(.foo.bar.baz) {}
+:host-context(.foo.bar.baz) {}
 ```
 
 <!-- prettier-ignore -->
 ```css
-:local(.foo, :global(.bar), .foo.bar.baz) {}
+:host-context(.foo, :host(.bar), .foo.bar.baz) {}
```
Mouvedia, line 144: nit

```suggestion
:host-context(.foo, :host(.bar), .baz.qux) {}
```

lib/rules/selector-max-specificity/__tests__/index.mjs
romainmenke: Yes, but upstream : https://github.com/csstools/postcss-plugins/blob/3ac6363e487f6e22637c46107498d1ece1fa8692/packages/selector-specificity/test/tests-from-bramus-specificity.mjs#L111-L116

lib/rules/selector-max-specificity/README.md
romainmenke: The config in the readme allows the specificity of two classes.

`:host-context(.foo, :host(.bar).baz) {}` after all the ignores is `.foo, .bar.baz`.

I think the intention was to illustrate how nesting selectors works when ignoring.
I'd rather keep it as is because it isn't wrong.

lib/rules/selector-max-specificity/README.md
romainmenke: This has one class selector less and wouldn't produce an error.

ybiquitous (commented):
Thanks, I like this change. Can you fix the test failures?

lib/rules/selector-max-specificity/README.md
Mouvedia: Did I find a bug then?

https://stylelint.io/demo/#N4Igxg9gJgpiBcCAWEDOAXAtJAdumAHugBQB0AZhBADQAE8KGZARgIYBOAlKWwF6e1gAXwA6OMQzRZc+ImUo16jEjw6c6q-oNHic8ADYQwrffKp14Ac0NtTqrhoX3NA4SGohyAS30wAcqwAtnCIhEEADr6kYKio7uAQON6WCCDAYrS0IiDsAK6+qNnwghmZWSCoML5g6BDsmIGsBJio4TBgXt4d6ACeRbQA2tkADNQATNTD2XTA5V6WOHUwAMpV7bXshQiDRcrYibLo2QC6tELHpTpC8bjJAGJ1jeipAFaoifGw4XGI6TiZ2QwPV8+i8eH62X0rHwGGmpUBvRBYOkiWSLXQrBwUA4UAhIChMKOIDE1yEQA

romainmenke: `:host` has its own specificity, if you also ignore that you get the expected outcome : https://stylelint.io/demo/#N4Igxg9gJgpiBcCAWEDOAXAtJAdumAHugBQB0AZhBADQAE8KGZARgIYBOAlKWwF6e1gAXwA6OMQzRZc+ImUo16jEjw6c6q-oNHickjNgh5CKhXX0q2XDQtXtNA4SGohyASwA2MAHKsAtnCIhP4ADl6kYKiozuBG7gDmCCDAYrS0IiDsAK5eqBnwgqlp6SCoMF5g6BDsmH6sBJioITBgbu6t6ACe+bQA2hkADNQATNQDGXTAJW7xONUwAMrlLVXseQh9+cqGxkQTJRYZALq0QkdFOkIxuAkAYtV16EkAVqhGMbAh0YgpOGkZGE6Xg8bjwPQyHlY+AwEyKAK6wNB0jiM0a6FYOCgHCg4JAkOh6AyYiuQiAA

romainmenke:
Thank you for the reviews, in particular from @Mouvedia 🙇

Mouvedia (approved):

Mouvedia:
@romainmenke #4189 might interest you.

PR #7689, opened 2024-05-11
Refactor `selector-max-specificity` to leverage upstream helper

romainmenke:

None

While working on `selector-max-specificity` for the end position fix I noticed that all of the complexity in the rule is for the `ignoreSelectors` secondary option.

Because this option is present it is needed to manually walk the selector and re-implement large sections of `@csstools/selector-specificity` so that ignoring selectors works as expected.

I've added an option upstream to customize the specificity calculations.
See : https://github.com/csstools/postcss-plugins/blob/main/packages/selector-specificity/docs/selector-specificity.selectorspecificity.md

Because this is integrated into `@csstools/selector-specificity` we can offload a lot more complexity to that package.

A good example was all of the logic around `:nth-child(n of .foo)`.
All this logic already existed upstream but couldn't be used before.

-------

A side effect of this change is that this rule is also a bit faster.

Without secondary options.

Main:
```
node scripts/benchmark-rule.mjs selector-max-specificity '10,10,10'
Warnings: 0
Mean: 370.16875366666665 ms
Deviation: 16.015808005438103 ms
```

This change:
```
node scripts/benchmark-rule.mjs selector-max-specificity '10,10,10'
Warnings: 0
Mean: 301.923139 ms
Deviation: 13.039688509741097 ms
```

With secondary options.

Main:
```
node scripts/benchmark-rule.mjs selector-max-specificity '["10,10,10", { "ignoreSelectors": "foo" }]'
Warnings: 0
Mean: 373.55708683333336 ms
Deviation: 17.037700655785976 ms
```

This change:
```
node scripts/benchmark-rule.mjs selector-max-specificity '["10,10,10", { "ignoreSelectors": "foo" }]'
Warnings: 0
Mean: 317.46817016666665 ms
Deviation: 19.0016953176133 ms
```


Mouvedia:
> A _side effect_ of this change is that this rule is also a bit faster.

Be bold and say that's what you intended all along :)

lib/rules/selector-max-specificity/index.mjs
```diff
@@ -237,7 +108,16 @@ const rule = (primary, secondaryOptions) => {
...
+					if (
+						compare(
+							selectorSpecificity(resolvedSelector, {
+								customSpecificity: secondaryOptions?.ignoreSelectors
+									? isSelectorNodeIgnored
+									: undefined,
+							}),
+							maxSpecificity,
+						) > 0
+					) {
```
Mouvedia, line 120: suggestion
refactor

prettier is getting funky with the formatting
maybe set `const customSpecificity = secondaryOptions?.ignoreSelectors ? isSelectorNodeIgnored : undefined;` before the `if`?

lib/rules/selector-max-specificity/index.mjs
```diff
@@ -80,146 +57,40 @@ const rule = (primary, secondaryOptions) => {
...
 				case 'class':
 				case 'id':
 				case 'tag':
-					return simpleSpecificity(node);
-				case 'pseudo':
-					return pseudoSpecificity(node);
-				case 'selector':
-					// Calculate the sum of all the direct children
-					return specificitySum(node.map((n) => nodeSpecificity(n)));
-				default:
```
Mouvedia, line 221: Can you add a default with either a comment or a `throw`?

lib/rules/selector-max-specificity/index.mjs
```diff
@@ -80,146 +57,40 @@ const rule = (primary, secondaryOptions) => {
...
+		const isSelectorNodeIgnored = (node) => {
 			switch (node.type) {
 				case 'attribute':
 				case 'class':
 				case 'id':
 				case 'tag':
-					return simpleSpecificity(node);
-				case 'pseudo':
-					return pseudoSpecificity(node);
-				case 'selector':
```
Mouvedia, line 218: question

Is there a reason for this case removal?

romainmenke: The old function that included a switch statement over node types also did the recursive decent down the AST. This is no longer needed.

The updated version only has a switch statement over node types to customize their specificity. Given that `selector` type nodes aren't ignorable they are omitted.

------------

Git diff seems to align these two switch statements, but they are actually unrelated in function and purpose.

lib/rules/selector-max-specificity/index.mjs
romainmenke: I've made a few more changes in the same spirit.
Let me know what you think

lib/rules/selector-max-specificity/index.mjs
```diff
@@ -80,148 +57,46 @@ const rule = (primary, secondaryOptions) => {
...
+							const emptySpecificity = selectorSpecificity(node);
+
+							return {
+								a: entireSpecificity.a - emptySpecificity.a,
+								b: entireSpecificity.b - emptySpecificity.b,
+								c: entireSpecificity.c - emptySpecificity.c,
+							};
+						}
+						default:
+						// Other node types are not ignorable.
```
ybiquitous, line 96: [suggest] Super-minor, you can ignore it. If adding the `return` statement, Prettier would align indent:

```suggestion
							// Other node types are not ignorable.
							return;
```

lib/rules/selector-max-specificity/index.mjs
```diff
@@ -80,148 +57,46 @@ const rule = (primary, secondaryOptions) => {
...
+							if (!node.nodes?.length) {
+								return zeroSpecificity();
+							}
+
+							// We only ignore the current pseudo-class, not the specificity of the child nodes.
+							// Calculate the diff between specificity with and without child nodes.
+							const entireSpecificity = selectorSpecificity(node);
+
+							node.nodes = [];
+							const emptySpecificity = selectorSpecificity(node);
```
ybiquitous, line 87: [ask] If it was necessary to respect the original code, we would clone the node instead of mutating it like this:

```suggestion
							const emptySpecificity = selectorSpecificity(node.clone({ nodes: [] }));
```

This mutation is fine, isn't it?

ybiquitous (commented):
@romainmenke Thanks for the pull request. This is a great improvement! Overall LGTM, but I've left a few comments, so I'd appreciate if you would check them out.

lib/rules/selector-max-specificity/index.mjs
romainmenke: Good idea!

Theoretically it's possible to have an auto fix for this rule with `:where()`
Mutating the selector AST without cloning could be surprising in the future.

lib/rules/selector-max-specificity/index.mjs
romainmenke: Eslint errors on this suggestion with `Unnecessary return statement`, so I've reverted this back to what it was.

But I agree that the indent should be as suggested :)

Mouvedia (approved):

ybiquitous (approved):
Thanks, LGTM 👍🏼

romainmenke:
Thank you for the reviews 🙇

PR #7702, opened 2024-05-22
Refactor to add `selectorSourceIndex` utility

romainmenke:

See : https://github.com/stylelint/stylelint/pull/7685#discussion_r1597462061

I had actually expected this to be simpler but I now understand that this isn't possible.

The whitespace between `,` and the selector needs to go somewhere in the AST and the selector node is the right place. This implies that a selector `sourceIndex` includes leading whitespace.

We want to find the `sourceIndex` value without leading whitespace.

`selectorNode.first?.sourceIndex ?? 0`: this was still not ideal because `0` is not the right fallback.

For example with : `.foo,,`

This is a list with 3 entries. The middle entry is definitely empty.
We don't want to use `0` as the source index value because that is the location of `.` in `.foo`.

Using `node.first?.sourceIndex ?? node.sourceIndex` is more correct.

This also updates `postcss-selector-parser` to `6.1.0` as this version includes the patches to make `node.sourceIndex` available.


lib/utils/selectorSourceIndex.mjs
```diff
@@ -0,0 +1,13 @@
+/**
+ * Get the source index of a selector, excluding leading whitespace.
+ *
+ * @typedef {import('postcss-selector-parser').Container<string | undefined>} Container
+ * @param {Container} node
+ *
+ * @returns {number}
+ */
+export default function selectorSourceIndex(node) {
```
Mouvedia, line 9: question
nit

Are all type of `node`s accepted?

Mouvedia (approved):

lib/utils/selectorSourceIndex.mjs
ybiquitous, line 1: [suggest] For consistency, how about renaming the function to `getSelectorSourceIndex` like other `get*` utils (e.g. `getAtRuleParams`)? But this suggestion depends on my feelings, and you can leave the name as-is if you have any thoughts. 😅

ybiquitous (approved):
Nice catch! I left a comment on styling, but LGTM 👍🏼

lib/utils/selectorSourceIndex.mjs
romainmenke: Thank you!

Naming these things is always tricky for me, so this feedback is very much appreciated!

ybiquitous (approved):
LGTM!

romainmenke:
Thank you for the reviews 🙇

PR #7703, opened 2024-05-22
Fix `selector-not-notation` autofix of the `"simple"` option

Mouvedia:

N/A

While I was working on #2643 I uncovered a bug.
`:not()` is mostly used as a suffix so no one reported it.
@ybiquitous  we should release once that's merged.

i.e. `foo:not()` is not affected


ybiquitous (approved):
Nice catch! LGTM 👍🏼

PR #7704, opened 2024-05-23
Refactor types for `flattenNestedSelectorsForRule` and `getSelectorSourceIndex`

romainmenke:

See : https://github.com/stylelint/stylelint/pull/7702#discussion_r1610745608

@Mouvedia correctly noted that `getSelectorSourceIndex` was maybe too ambiguously typed.

I used `Container<string | undefined>` as this was the type that was already present and was expected to be used.

In this change however I used the more correct `Selector` type and made the needed changes to get there.

Notable changes:
- `flattenNestedSelectorsForRule` now returns `Root` objects, not arrays of selectors -> easier to walk
- walking is now done with the `postcss-selector-parser` AST helpers and not with a recursive `checkSelector`

This feels more solid to me and more in line with the `postcss-selector-parser` API.

_Submitting as a separate PR because it ended up being a larger change than the original PR._


lib/rules/selector-max-attribute/index.mjs
```diff
@@ -92,8 +90,21 @@ const rule = (primary, secondaryOptions) => {
...
+
+						return;
+					}
+
+					if (
+						isPseudo(childSelector.parent) &&
+						isContextFunctionalPseudoClass(childSelector.parent)
+					) {
+						checkSelector(childSelector, selector, ruleNode);
+					}
```
romainmenke, line 107: I don't really like that we are now checking the parent of a node multiple times.

Before we checked that a node itself was a specific kind of pseudo and then recursively called `checkSelector` on it.

Now we are walking down the AST and then we check the parent.

This is more work, marginally more work, but still more work.

lib/rules/selector-max-attribute/index.mjs
```diff
@@ -92,8 +90,21 @@ const rule = (primary, secondaryOptions) => {
 			if (!isStandardSyntaxRule(ruleNode)) return;
 
 			flattenNestedSelectorsForRule(ruleNode, result).forEach(({ selector, resolvedSelectors }) => {
-				resolvedSelectors.forEach((resolvedSelector) => {
-					checkSelector(resolvedSelector, selector, ruleNode);
+				resolvedSelectors.walk((childSelector) => {
```
ybiquitous, line 93: [note] I wish `postcss-selector-parser` would add `walkSelector` like the existing `walkPseudos`! 😄

lib/rules/selector-max-attribute/index.mjs
```diff
@@ -92,8 +90,21 @@ const rule = (primary, secondaryOptions) => {
...
+					if (childSelector.type !== 'selector') return;
```
ybiquitous, line 94: [suggest] Minor. The `isSelector` type-guard might be a bit more consistent and readable (for example, see `isRoot` and `isPseudo`). But this is not required, you can leave it as-is. 👍🏼 

```suggestion
					if (!isSelector(childSelector)) return;
```

lib/rules/selector-max-attribute/index.mjs
ybiquitous: Fair point. We can accept it for now. 👍🏼

lib/rules/selector-max-attribute/index.mjs
```diff
@@ -92,8 +90,21 @@ const rule = (primary, secondaryOptions) => {
...
+					if (isRoot(childSelector.parent)) {
+						checkSelector(childSelector, selector, ruleNode);
+
+						return;
+					}
+
+					if (
+						isPseudo(childSelector.parent) &&
+						isContextFunctionalPseudoClass(childSelector.parent)
+					) {
```
ybiquitous, line 105: [suggest] How about simplifying callers of `isContextFunctionalPseudoClass`?

```suggestion
					if (isContextFunctionalPseudoClass(childSelector.parent)) {
```

Perhaps we could achieve this by changing the parameter type of `isContextFunctionalPseudoClass` like this:

```diff
- * @param {import('postcss-selector-parser').Node} node - postcss-selector-parser node (of type pseudo)
+ * @param {import('postcss-selector-parser').Node | import('postcss-selector-parser').Container | undefined} node - postcss-selector-parser node (of type pseudo)
```

ybiquitous (commented):
Overall LGTM, thanks for the improvements!

I've left a few comments, so please take a look. 🙏🏼

lib/rules/selector-max-attribute/index.mjs
romainmenke: I was also surprised that it was missing!

lib/rules/selector-max-attribute/index.mjs
romainmenke: This also made it possible to combine the two conditions (`isRoot` and `isContextFunctionalPseudoClass`) in a single `if` while keep it readable.

Let me know what you think 🙇

ybiquitous: Great! 👏🏼

ybiquitous (approved):
Thank you, LGTM! 👍🏼

romainmenke:
Thank you for the reviews 🙇

PR #7705, opened 2024-05-23
Refactor to jsonFormatter

Connormiha:
`JSON.stringify` has a second argument for filtering output. So we can easily simplify this function without reduce, filter, Object.entries.


ybiquitous (approved):
Great catch! Thank you! 👍🏼

Issue #7192, opened 2023-09-18
Add replaced text and corresponding range to `Warning` type


pamelalozano16:
### What is the problem you're trying to solve?

Code review toolings, such as Github API or Google's internal lint checks, often require linters to provide a fix for each finding individually rather than fixing all findings in the entire file. Stylelint doesn't support such feature as it can only fix the entire file with `--fix`, therefore it is not able to fix just one finding or to limit its changes to a region in a file. 

### What solution would you like to see?

Stylelint stores each individual finding in a structured data format.

keywords: granular, eslint-nibble

ybiquitous:
@pamelalozano16 Thanks for the request. That's so interesting to me. To clarify, can you provide a specific example with other tools, please?

pamelalozano16:
Sure, here's a [JSON reporter](https://github.com/pylint-dev/pylint/blob/main/pylint/reporters/json_reporter.py) from Pylint and a [Javascript reporter](https://github.com/pylint-dev/pylint/blob/main/pylint/reporters/json_reporter.py) from JSLint. As you can see each finding is reported individually.

Mouvedia:
related: #4331

ybiquitous:
@pamelalozano16 Thanks for the info. I understand your request well (although the link to JSLint may be incorrect). 👍🏼 

We can probably refer to ESLint, too. It provides the API to express info about autofix:
https://eslint.org/docs/latest/integrate/nodejs-api#-lintmessage-type

With the following example of ESLint JSON results, the `fix` property seems what we want.

```json
{
  "ruleId": "no-var",
  "severity": 2,
  "message": "Unexpected var, use let or const instead.",
  "line": 2,
  "column": 1,
  "nodeType": "VariableDeclaration",
  "messageId": "unexpectedVar",
  "endLine": 2,
  "endColumn": 7,
  "fix": {
    "range": [6, 9],
    "text": "let"
  }
}
```

pamelalozano16:
Yes, that's a good example!

ybiquitous:
@pamelalozano16 Please let us know your specific use case. What kind of data do you need? Like ESLint, are only `fix.range` and `fix.text` enough?

pamelalozano16:
Yes! `fix.range` and `fix.test` would be enough information for our purposes.

jeddy3:
[...]

PR #7706, opened 2024-05-23
refactor fixers of `selector-not-notation` and `function-name-case`

Mouvedia:

#2643

1.
Most of the rules won't require as much work as `selector-not-notation` to transition.
That's why I have chosen to refactor a simpler rule to demonstrate how easy it could be.

2.
This PR unlocks #7192 and #4664.

3.
`selector-not-notation` now reports new line ranges correctly

4.
`function-name-case` reports the full name instead of the first letter previously

5.
see https://github.com/stylelint/stylelint/pull/7706#issuecomment-2131429567

6.
tests can be found here: [`c162413` (#7729)](https://github.com/stylelint/stylelint/pull/7729/commits/c162413e0a1d3ae9e2d50879a87df1ca2fc5acb8)

> TODO

- [x] `stylelint-disable`
- [x] `stylelint-disable-line`
- [x] `stylelint-disable-next-line`
- [x] `stylelint-* foo-bar`
- [x] `stylelint-* foo-bar, qux-baz`
- [x] add changeset
- [x] `ignoreDisables`
- [x] create experimental branch
	- revert to unsafe behaviour
	- update [`stylelint.utils`](https://github.com/stylelint/stylelint/blob/main/docs/developer-guide/plugins.md#stylelintutils)
	- add tests


lib/rules/selector-not-notation/index.mjs
```diff
@@ -197,6 +220,23 @@ function fixComplex(previousNot) {
...
+ * @param {RuleNode} ruleNode
+ */
+function getSource(pseudo, ruleNode) {
+	assert(ruleNode.source?.start && pseudo.source?.start && pseudo.source?.end);
+	const startline = ruleNode.source.start.line + pseudo.source.start.line - 1;
+	const endline = ruleNode.source.start.line + pseudo.source.end.line - 1;
+
+	return {
+		start: { line: startline, column: pseudo.source.start.column },
+		end: { line: endline, column: pseudo.source.end.column + 1 },
```
Mouvedia, line 238: minor
question

@alexander-akait do you know why do I need `+ 1` here?

lib/lintSource.mjs
```diff
@@ -151,3 +154,29 @@ function createEmptyPostcssResult(filePath) {
...
+	const array = Object.values(fixersData).flat();
+	/** @type { Array<[number, number]> | undefined } */
+	const ranges = all?.map(({ start, end }) => [Number(start), Number(end)]);
+
+	array.forEach(({ source, callback, args, unfixable }) => {
+		const isInRange = (/** @type {[number, number]} */ [start, end]) =>
+			source?.start?.line && source.start.line >= start && source.start.line <= end;
+
+		if (!ranges || !ranges.some(isInRange)) {
+			!unfixable && callback(args);
```
Mouvedia, line 179: note

We don't bail out on the fixer addition because we will eventually have to expose the fix data regardless of the fixability.
i.e. it's simpler to do it that way than having different paths

types/stylelint/index.d.ts
```diff
@@ -252,6 +254,16 @@ declare namespace stylelint {
 		fixable?: boolean;
 	};
 
+	type FixerArguments = { [x: string]: any };
+
+	export type Fixer = (args?: FixerArguments) => string | void;
```
Mouvedia, line 259: note

`string` is to cover the `text` upcoming exposition
i.e. probably not in this PR but it's planned

Mouvedia:
@romainmenke @ybiquitous
Concerning

> We can either
> 
> - keep the safe behaviour introduced in v13.2.0 until all rules are migrated
> - revert to the v13.1.0 unsafe behaviour
> - or create a long-lived experimental branch

I would like to have your opinions.

lib/rules/selector-not-notation/index.mjs
Mouvedia: minor
question

@romainmenke
Why do I need + 1 here?
Is this a bug?

romainmenke: Not sure 🤔 

Different projects, parsers, ... use different conventions for these indices and positions.

Could be a bug, or could just be such a difference in implementation.

You could create an isolated case in the test suite of `postcss-selector-parser` to check if there are any bugs there.

Mouvedia: I wouldn't know where to start and I don't wish it to change.
i.e. if it was a bug it wouldn't be fixed because users depend on the off-by-one

Mouvedia:
> @romainmenke @ybiquitous Concerning
> 
> > We can either
> > 
> > * keep the safe behaviour introduced in v13.2.0 until all rules are migrated
> > * revert to the v13.1.0 unsafe behaviour
> > * or create a long-lived experimental branch
> 
> I would like to have your opinions. Depending on this decision Ill have to make a few more changes to this branch. FYI there are 33 fixable rules. cf [49b343a](https://github.com/stylelint/stylelint/commit/49b343a44e6dc4f9d14bf94ed2f65783fe158c30)

Thinking about it more, I think a mix of the 3 is possible.

1. create umbrella issue
2. create experimental branch
3. add fixers to the main branch
4. test them on the experimental branch
5. revert #4573 once all the rules have transitioned

For the users it's transparent and the migration will be smoother.
I don't think I can create the experimental branch with my rights though.

types/stylelint/index.d.ts
```diff
@@ -36,9 +36,7 @@ type FileCache = {
 
 type EmptyResult = {
 	root: {
-		nodes?: undefined;
 		source: {
-			lang?: undefined;
```
Mouvedia, line 41: note

These were forgotten during https://github.com/stylelint/stylelint/commit/e1c8d2de681671a9698a16c724baf93ecef593f8#diff-efea4e624a129cbb42ab5b2d771e4d6749b632549e135d5e77e90145097a870f

lib/rules/selector-not-notation/index.mjs
```diff
@@ -82,43 +90,49 @@ const rule = (primary, _, context) => {
...
+					const second = selectors.length > 1 && selectors[1];
+					const isFixable =
+						second && (!second.nodes.length || selectors.every(({ nodes }) => nodes.length === 1));
+
+					if (context.fix) {
+						return addFixer(result, ruleName, {
+							source,
+							callback: fixer,
+							args: { cb: fixSimple, nodes: { root: selectorRoot, rule: ruleNode, pseudo } },
+							unfixable: !isFixable,
```
Mouvedia, line 122: note

We don't bail out on the fixer addition because we will eventually have to expose the fix data regardless of the fixability status.
i.e. it's simpler to do it that way than having different paths down the line

lib/rules/selector-not-notation/index.mjs
ybiquitous: I'm reading the code in this PR, but this seems related to `endColumn` in PostCSS API: ⬇️ 

> Column for exclusive end position in the input file with this warning’s source.

https://github.com/postcss/postcss/blob/f9a9868eb3a591cd99d504ac57bf7ee77380856b/lib/warning.d.ts#L67

---

The following test code:

```js
import postcss from 'postcss';

const print = ({ start, end }) => {
	const { line: sLine, column: sCol } = start;
	const { line: eLine, column: eCol } = end;
	console.log('start: %o, end: %o', { line: sLine, column: sCol }, { line: eLine, column: eCol });
};

console.log('.abc {}');
console.log('1234567');
console.log('---');

print(postcss.parse('.abc {}').nodes[0].source);
print(selectorParser().astSync('.abc').nodes[0].nodes[0].source);
```

prints below: ⬇️ 

```
.abc {}
1234567
---
start: { line: 1, column: 1 }, end: { line: 1, column: 7 }
start: { line: 1, column: 1 }, end: { line: 1, column: 4 }
```

It seems that `postcss` and `postcss-selector-parser` are consistent for source positions.

ybiquitous (commented):
@Mouvedia Thanks for working on old issues! I'm still reading the code, but I've left a few comments for you to address.

lib/rules/selector-not-notation/index.mjs
Mouvedia: I found the reason:

```
  export interface Range {
    /**
     * End position, exclusive.
     */
    end: Position

    /**
     * Start position, inclusive.
     */
    start: Position
  }

  export interface Source {
    /**
     * The inclusive ending position for the source
     * code of a node.
     */
    end?: Position

   // …

    /**
     * The inclusive starting position for the source
     * code of a node.
     */
    start?: Position
  }
```

lib/rules/selector-not-notation/index.mjs
```diff
@@ -199,6 +225,41 @@ function fixComplex(previousNot) {
...
+ * @param {number} [offset]
+ */
+function getNewSource(pseudo, ruleNode, offset = 0) {
+	assert(ruleNode.source?.start && pseudo.source?.start);
+	const line = ruleNode.source.start.line + pseudo.source.start.line - 1;
+	const column = pseudo.source.start.column;
+
+	return {
+		start: { line, column },
+		end: { line, column: column + pseudo.toString().length + offset },
```
Mouvedia, line 261: WIP

This one doesn't return the right column.
I gotta choose between not returning it—which is not necessary—or fixing it.

types/stylelint/index.d.ts
```diff
@@ -1,6 +1,7 @@
 import type * as PostCSS from 'postcss';
 import type { GlobbyOptions } from 'globby';
 import type { cosmiconfig, TransformSync as CosmiconfigTransformSync } from 'cosmiconfig';
+import { RequiredDeep } from 'type-fest';
```
Mouvedia, line 4: opinion

I think it's worth adding that dependency.
I won't mind if you disagree; if so I will just have to duplicate the types.
see https://github.com/sindresorhus/type-fest/commit/c2bf37491077ee5fbb580d913ba0c80152d65ce0

types/stylelint/index.d.ts
Mouvedia: [fc5aede](https://github.com/stylelint/stylelint/commit/fc5aede1ef5dfdff7d7e7c24370d02d1a546142a)

types/stylelint/index.d.ts
ybiquitous: If using `type-fest`, will there be any effect on projects depending on Stylelint's public types?

Mouvedia: I am not sure I understand, `RequiredDeep` is just a type provided by a dependency.
It could easily be vendored on our side and it would work the same.

lib/applyFixes.mjs
```diff
@@ -0,0 +1,67 @@
...
+ * @param {PostcssResult['stylelint']['disabledRanges']} disabledRanges
+ * @param {number} endLine
+ * @param {number} diff
+ */
+function offsetDisabledRanges(disabledRanges, endLine, diff) {
+	const values = Object.values(disabledRanges);
+
+	for (const element of values) {
+		element.forEach((disabledRange) => {
+			if (disabledRange.start > endLine) disabledRange.start += diff;
```
Mouvedia, line 62: **todo**

```
:not(
a
):not(div) {} /* stylelint-disable-next-line */
:not(a):not(div) {}
```
to
```
:not(a, div) {} /* stylelint-disable-next-line */
:not(a):not(div) {}
```

This is an edge case that I have missed.

types/stylelint/index.d.ts
ybiquitous: When I try this PR as a dependency in a sample project, I get the following error: ⬇️ 

```sh-session
$ npx tsc
node_modules/stylelint/types/stylelint/index.d.ts:4:35 - error TS2307: Cannot find module 'type-fest' or its corresponding type declarations.

4 import type { RequiredDeep } from 'type-fest';
                                    ~~~~~~~~~~~


Found 1 error in node_modules/stylelint/types/stylelint/index.d.ts:4
```

Maybe, it's necessary to have `type-fest` in `dependencies`, not `devDependencies`.

ybiquitous: However, I don't think adding `type-fest` to `dependencies` is worth just for `RequiredDeep`... 🤔

Mouvedia: > However, I don't think adding type-fest to dependencies is worth just for RequiredDeep... 🤔

I will remove it.

lib/applyFixes.mjs
```diff
@@ -0,0 +1,67 @@
...
+	} = result.stylelint;
+	const rules = Object.entries(fixersData);
+
+	rules.forEach(([ruleName, array]) => {
+		const ruleRanges = disabledRanges[ruleName]?.map(cb) || [];
+		const ranges = all.map(cb).concat(ruleRanges);
+
+		array.forEach(({ source, callback, args, unfixable }) => {
+			const isInRange = (/** @type {Tuple} */ [start, end]) =>
+				source.start.line >= start && (!end || source.start.line <= end);
```
Mouvedia, line 34: todo wip

types/stylelint/index.d.ts
Mouvedia: [33a0bc6](https://github.com/stylelint/stylelint/commit/33a0bc61a67845b2cd819e15885ece4b231cb68e)

Mouvedia:
I have rebased on `main` due to your refactor PR.

types/stylelint/index.d.ts
```diff
@@ -252,6 +253,26 @@ declare namespace stylelint {
...
+			column?: number /** exclusive */;
+		};
+	};
+
+	export type Fixer = (args?: FixerArguments) => Range;
+	export type FixerData = {
+		source: { start: Required<Exclude<Range['start'], undefined>>; end: Required<Range['end']> };
+		callback: Fixer;
+		args?: FixerArguments;
+		unfixable?: boolean;
```
ybiquitous, line 273: [ask] Is `unfixable` necessary? If `unfixable` is `true`, we don't need to add a fixer, right?

types/stylelint/index.d.ts
```diff
@@ -252,6 +253,26 @@ declare namespace stylelint {
...
+			line: number;
+			column?: number /** exclusive */;
+		};
+	};
+
+	export type Fixer = (args?: FixerArguments) => Range;
+	export type FixerData = {
+		source: { start: Required<Exclude<Range['start'], undefined>>; end: Required<Range['end']> };
+		callback: Fixer;
+		args?: FixerArguments;
```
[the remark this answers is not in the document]
ybiquitous: See the alternative API in <https://github.com/stylelint/stylelint/pull/7706#discussion_r1617499836> that I suggested. If we eventually expose a new API, the API should be usable.

types/stylelint/index.d.ts
ybiquitous: Note that `context.fix` does not equal `meta.fixable`.

Here, we can refer to the ESLint API:
- [How to report problems](https://eslint.org/docs/latest/extend/custom-rules#reporting-problems)
  - `context.report()` provides the `fix` function property
- [`LintMessage` type](https://eslint.org/docs/latest/integrate/nodejs-api#-lintmessage-type)
  - This is almost the same as Stylelint's `Warning` type:
  - https://github.com/stylelint/stylelint/blob/d902c0b7cc1233b74cc4959eff864cb3e2d87f73/types/stylelint/index.d.ts#L480

I believe the Stylelint API would also become similar to ESLint's one.

Mouvedia: > Note that `context.fix` does not equal `meta.fixable`.

I know.

> I believe the Stylelint API would also become similar to ESLint's one.

Maybe ESLint has rules that can be fixed only in certain cases like us.

> fix (EditInfo | undefined)
> The EditInfo object of autofix. **This property is undefined if this message is not fixable.**

types/stylelint/index.d.ts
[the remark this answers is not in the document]
Mouvedia: If we fix we don't report and vice versa.
i.e. you only report if you don't fix

So I don't understand why you would give the responsibility of the _fixer callback addition_ to `report`.
This seems illogical to me.

<hr />

Do you want me to 

> put `fixSimple` and `fixComplex` in the scope of l91–136
> remove `args` from `FixerData` type

?

types/stylelint/index.d.ts
ybiquitous: Still, I don't understand why we cannot use `report()` to add a fixer. 🤔 

To achieve #7192, we should add autofix info to our `Warning` type, right? For this purpose, I think the already exposed `report()` API is most appropriate instead of adding a new API like `addFixer()`, because plugin authors' learning curve may be kept lower.

lib/applyFixes.mjs
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * if a fix is attempted a stylelint-disable comment can either be before or after
+ * i.e. stylelint-disable-line would have prevented the fix in the first place
+ * @summary emend the start/end lines of the disabled ranges
+ * @param {PostcssResult['stylelint']['disabledRanges']} disabledRanges
+ * @param {number} endLine
+ * @param {number} diff
+ */
+function offsetDisabledRanges(disabledRanges, endLine, diff) {
```
Mouvedia, line 57: refactor
wip

This function might be unnecessary.
I am currently adding tests on `autofix-disable-setup` to prove it.

types/stylelint/index.d.ts
Mouvedia: > Still, I don't understand why we cannot use `report()` to add a fixer. 🤔

You must have missed https://github.com/stylelint/stylelint/pull/7706#discussion_r1617595101.
You should review our rules, you will notice that we often have this pattern

```js
if (context.fix) {
	/* … */
	return;
}

report({
	/* … */
})
```

In short the `fixer` is never added through `report` because we are in the fix path not in the report path.
I anticipate your next question if you want to put it on `rule` it won't work either because you will lack the dynamic aspect.
i.e. the `fixer callback` must be added while the rule is being run

ybiquitous: Can we not change `report()` internally to support a fixer, as ESLint does?

Mouvedia: > Can we not change `report()` internally to support a fixer, as ESLint does?

I don't think we can without revamping how every rule currently works.
This would completely out of scope for this refactor and frankly quite the undertaking.

<hr/>

`addFixer` will exposed as an util to the plugin author here:

https://github.com/stylelint/stylelint/blob/d902c0b7cc1233b74cc4959eff864cb3e2d87f73/package.json#L43

and users may use the `--safe-disables` flag once we are done.

Mouvedia: I am just saying plugin authors will have a way to refactor their rules as well.

off topic

`--safe-disables` will fallback to the safe behaviour _if_ the config contains a rule which has not been transitioned else it will use the fixers in files that contain stylelint commands.
For that to work I was planning to add a `fixer: boolean` to `RuleMeta`.
The default being falsy if missing, we could easily detect plugin rules that have not yet been refactored.
Hopefully that helps you envision how everything will end up.

types/stylelint/index.d.ts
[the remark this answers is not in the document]
Mouvedia: Let's go back to your original question:

> Is `args` necessary? Cannot we add a callback as a closure?

Do you want me to remove `args`?
It was intended as a nicety so that you don't have to move your fixers in the scope.
I am fine with it.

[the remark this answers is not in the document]
Mouvedia: > Is `args` necessary? Cannot we add a callback as a closure?

I went ahead and removed `args`.

types/stylelint/index.d.ts
```diff
@@ -252,6 +253,23 @@ declare namespace stylelint {
...
+			column: number /** inclusive */;
+		};
+		end: {
+			line: number;
+			column: number /** exclusive */;
+		};
+	};
+
+	export type FixerData = {
+		source: Range;
```
Mouvedia, line 268: At this point I consider this PR ready to be merged as is but I have still one last refactor in mind that might be an improvement.
As explained [here](https://github.com/stylelint/stylelint/pull/7706/files#r1615360093) there could be confusion over the terminology.
i.e. what you would expect from `end.column` to be, based on its parent name

Therefore I think renaming `source` to `range` would make sense for people who have already used a `Source` before to avoid unnecessary confusion even if the type is clear.
@ybiquitous Ill move on with the renaming shortly if you don't have anything against it.
@romainmenke does that make sense to you?

types/stylelint/index.d.ts
romainmenke: I think we can choose any path here. I don't agree that it is impossible to add this to `report()`. Not saying that it must be part of `report()` persé, but rather that both options are available and we should weigh the benefits of both.

----------

How I see it :

Code you write in a Stylelint rule itself is "user land" code.

However when you call `report()` you are reporting information to Stylelint and handing over control back to the framework.

The framework then decides what to do with the report based on provided info, user preferences, disable comments, ...

From that perspective I think it makes sense to include the fixers into `report()`.

That doesn't imply that auto fixes need to use `report()`.
Only those that want to make use of this new API and have support for all the niceties that come with it need to update.

Any other rule will work as before and can still do anything they want. Rules remain "user land" code and rule authors retain the ability to mutate PostCSS AST nodes in rules.

(even today rule authors can write code that ignores `context.fix` and don't have to set `meta.fixable = true`)

Integrating into `report()` means that authors don't even need to add conditions for fixes.

Migrating to this new API could like roughly like this :

_ignore the exact field naming and API shape, the dropped `if` is more interesting than what is passed to `report()`_

```diff
- if (context.fix) {
- 	node.value = expectedFunctionName;
- 	setDeclarationValue(decl, parsed.toString());
- 
- 	return;
- }

report({
	message: messages.expected,
	messageArgs: [functionName, expectedFunctionName],
	node: decl,
	index,
	endIndex,
	result,
	ruleName,
+ 	fix: () => {
+ 		node.value = expectedFunctionName;
+ 		setDeclarationValue(decl, parsed.toString());
+ 	},
+ 	fixRange: decl.rangeBy({ index, endIndex }),
});
```

Another benefit of leveraging `report()` is that it blurs the lines between fixes en reports. Things like a dry run then become quasi trivial to implement.

--------

On the other hand I wonder if `fix()` isn't better than `addFixer()`?

This would be nicer imho:
- `report({...})`
- `fix({...})`

I do think it is important that these closely mirror each other.
Currently `addFixer` takes a list of arguments while `report` takes a single object.
They should both take a single object.

romainmenke (commented):
Thank you for taking up this effort @Mouvedia!

I have one concern with this change and that is that it seems to defers the fixes.

Is that correct?

The order of execution seems to change to roughly:
- lint source with Rule A
- capture fixes for Rule A in a callback
- lint source with Rule B
- capture fixes for Rule B in a callback
- lint source with Rule C
- capture fixes for Rule C in a callback
- ...
- apply fixes for Rules A, B, C, ...

While the current order of execution is roughly :
- lint source with Rule A
- apply fixes for Rule A
- lint source with Rule B
- apply fixes for Rule B
- lint source with Rule C
- apply fixes for Rule C
- ...

This change means that we have to define what happens in these scenario's and add sufficient test coverage:
- two rules mutate the same AST node
- a rule mutates an AST node that has since been removed
- a rule inserts a new node and this new node can be autofixed by another rule (e.g. insert `:not()` and something else enforces allcaps `:NOT()`)
- ...

types/stylelint/index.d.ts
Mouvedia: > On the other hand I wonder if `fix()` isn't better than `addFixer()`?

I chose `addFixer` so that the temporality is clear.
i.e. I would expect `fix` to fix immediately
The notion of queuing/collection is important to convey.
Could this be considered a detail of implementation?
Maybe it's better to hide that to the rule's author.
I am not sure.

> I do think it is important that these closely mirror each other.
> Currently `addFixer` takes a list of arguments while report takes a single object.
> They should both take a single object.

You are correct; I will change its signature.

<hr />

Concerning `report`, making the changes you are describing implies

1. changing how rules are expected to work
2. fixing existing tests
3. fixing documentation
e.g. https://stylelint.io/developer-guide/rules#add-autofix, https://stylelint.io/developer-guide/plugins#the-anatomy-of-a-plugin
4. conflating reporting with fixing
5. giving `report` access to `context.fix`
6. `report` not calling `result.warn` if `context.fix` is `true`
7. increasing the diff of all refactors (longer to review)
8. etc.

I am aiming for these 33 refactors to be easy to undergo.
i.e. 2 or 3 developers should be able to tackle them in a single release
Some rules—e.g. `function-calc-no-unspaced-operator`—are already quite hard to migrate using `addFixer`.
I am trying to be pragmatic here.

**I am not saying it's not doable but that list seems daunting to me.**

`report`'s description is `Report a problem.`.
I can be interpreted in 2 ways:
 - reports a problem to stylelint
 - reports a problem to the user

What it actually does is `Creates an instance of Warning and adds it to Result#messages`.
That doesn't sound like it should have the responsibility of collecting fixes.

Mouvedia:
> I have one concern with this change and that is that it seems to defer the fixes.
> 
> Is that correct?
> 
> The order of execution seems to change to roughly:
> 
> * lint source with Rule A
> * capture fixes for Rule A in a callback
> * lint source with Rule B
> * capture fixes for Rule B in a callback
> * lint source with Rule C
> * capture fixes for Rule C in a callback
> * ...
> * apply fixes for Rules A, B, C, ...
> 
> While the current order of execution is roughly :
> 
> * lint source with Rule A
> * apply fixes for Rule A
> * lint source with Rule B
> * apply fixes for Rule B
> * lint source with Rule C
> * apply fixes for Rule C
> * ...
> 
> This change means that we have to define what happens in these scenario's and add sufficient test coverage:
> 
> * two rules mutate the same AST node
> * a rule mutates an AST node that has since been removed
> * a rule inserts a new node and this new node can be autofixed by another rule (e.g. insert `:not()` and something else enforces allcaps `:NOT()`, not sure if we have such conflicts, but they will definitely exist)
> * ...

I am gonna test one of these scenarios.
In the meantime, I will turn this PR into a draft.
Having the tests run on another branch is annoying but cannot be avoided if we are to merge into `main`.

PR #7708, opened 2024-05-24
Temporarily exclude `eslint` from Dependabot group updates

ybiquitous:

Closes #7695

We cannot bump `eslint` to v9 until all dependent plugins support v9.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,7 +15,7 @@ updates:
       changesets:
         patterns: ['@changesets/*']
       eslint:
-        patterns: ['eslint', 'eslint-*']
+        patterns: ['eslint-*'] # TODO: Include 'eslint' when we're ready for ESLint v9.
       jest:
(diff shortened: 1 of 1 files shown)


ybiquitous:
Just `.github/dependabot.yml` changes, so I'll merge without approvals.

.github/dependabot.yml
```diff
@@ -15,7 +15,7 @@ updates:
       changesets:
         patterns: ['@changesets/*']
       eslint:
-        patterns: ['eslint', 'eslint-*']
+        patterns: ['eslint-*'] # TODO: Include 'eslint' when we're ready for ESLint v9.
```
Mouvedia, line 18: Can we only ignore major updates instead?

ref https://github.blog/changelog/2021-05-21-dependabot-version-updates-can-now-ignore-major-minor-patch-releases/

ybiquitous: Ah, much better 👍🏼 Please open a PR 🙏🏼

PR #7716, opened 2024-05-24
Change `dependabot.yml` to only ignore eslint major updates and add postcss group

Mouvedia:

https://github.com/stylelint/stylelint/pull/7708#discussion_r1613715827

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,19 +9,23 @@ updates:
       - 'pr: dependencies'
     versioning-strategy: increase
     groups:
-      csstools-parser:
+      csstools:
         patterns: ['@csstools/*']
-        exclude-patterns: ['@csstools/selector-specificity'] # unrelated to other parser packages
       changesets:
         patterns: ['@changesets/*']
       eslint:
-        patterns: ['eslint-*'] # TODO: Include 'eslint' when we're ready for ESLint v9.
+        patterns: ['eslint', 'eslint-*']
       jest:
         patterns: ['jest', 'jest-*', '@jest/*']
+      postcss:
+        patterns: ['postcss', 'postcss-*']
       stylelint:
         patterns: ['@stylelint/*']
       typescript:
         patterns: ['typescript', '@types/*']
+    ignore:
+      - dependency-name: 'eslint'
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -9,19 +9,23 @@ updates:
       - 'pr: dependencies'
     versioning-strategy: increase
     groups:
-      csstools-parser:
+      csstools:
```
ybiquitous, line 12:
```suggestion
      csstools-parser:
```

.github/dependabot.yml
```diff
@@ -9,19 +9,23 @@ updates:
...
         patterns: ['@csstools/*']
-        exclude-patterns: ['@csstools/selector-specificity'] # unrelated to other parser packages
       changesets:
```
ybiquitous, line 14: [suggest] Let's keep this group because `@csstools/*` group update has a problem on Dependabot side. See https://github.com/dependabot/dependabot-core/issues/7646

```suggestion
        exclude-patterns: ['@csstools/selector-specificity'] # unrelated to other parser packages
      changesets:
```

@romainmenke Feel free to comment if you have any thoughts on this.

ybiquitous (changes requested):
Thanks for the pull request. I agree with ignoring eslint and grouping postcss. 👍🏼 

But for csstools, please consider reverting the change.

.github/dependabot.yml
```diff
@@ -9,19 +9,23 @@ updates:
       - 'pr: dependencies'
     versioning-strategy: increase
     groups:
-      csstools-parser:
+      csstools:
         patterns: ['@csstools/*']
```
Mouvedia, line 13:
```suggestion
      csstools-parser:
        patterns: ['@csstools/*']
        exclude-patterns: ['@csstools/selector-specificity'] # unrelated to other parser packages
```

.github/dependabot.yml
romainmenke: I agree to keep it as is.

I hope that they eventually fix that bug so that we can remove this exception :)

ybiquitous (approved):
Thanks, LGTM! 👍🏼

PR #7717, opened 2024-05-25
Fix typo in rules.md

Mouvedia:

N/A

https://github.com/stylelint/stylelint/commit/aa9be6f181613c695e80ddf97b0518b1ac674198#diff-b49ac02da5ff95ce9ef8c2de7f0b2ae24f8bed1f6789825bec7d7d133905a01fL36

### DIFF
--- a/lib/rules/__tests__/index.test.mjs
+++ b/lib/rules/__tests__/index.test.mjs
@@ -36,9 +36,9 @@ describe('fixability', () => {
 	test.each(ruleNames)('"%s" should be properly documented', async (name) => {
 		const rule = await rules[name];
 		const cell = rule.meta.fixable ? '\\s+🔧\\s+' : ' ';
-		const rulesPattern = `^.+\`${name}\`.+\\|.+\\|${cell}\\|$`;
 		const needle = '`fix` option';
 		const rulePattern = rule.meta.fixable ? needle : `^((?!${needle})(\\s|\\S))*$`;
+		const rulesPattern = `^.+\`${name}\`.+\\|.+\\|${cell}\\|$`;
 		const ruleDoc = await readFile(new URL(`../${name}/README.md`, import.meta.url), 'utf8');
 
 		expect(ruleDoc).toMatch(new RegExp(rulePattern));
--- a/lib/rules/__tests__/index.test.mjs
+++ b/lib/rules/__tests__/index.test.mjs
@@ -32,16 +32,17 @@ describe('all rules', () => {
 	});
 });
 
-describe('fixable rules', () => {
-	test.each(ruleNames)('"%s" should describe fixable in the documents', async (name) => {
+describe('fixability', () => {
+	test.each(ruleNames)('"%s" should be properly documented', async (name) => {
 		const rule = await rules[name];
-
-		if (!rule.meta.fixable) return;
-
+		const cell = rule.meta.fixable ? '\\s+🔧\\s+' : ' ';
+		const rulesPattern = `^.+\`${name}\`.+\\|.+\\|${cell}\\|$`;
+		const needle = '`fix` option';
+		const rulePattern = rule.meta.fixable ? needle : `^((?!${needle})(\\s|\\S))*$`;
 		const ruleDoc = await readFile(new URL(`../${name}/README.md`, import.meta.url), 'utf8');
 
-		expect(ruleDoc).toMatch('`fix` option');
-		expect(rulesListDoc).toMatch(new RegExp(`^.+\`${name}\`.+\\|.+\\|\\s+🔧\\s+\\|$`, 'm'));
+		expect(ruleDoc).toMatch(new RegExp(rulePattern));
+		expect(rulesListDoc).toMatch(new RegExp(rulesPattern, 'm'));
 	});
 });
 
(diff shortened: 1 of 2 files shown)


lib/rules/__tests__/index.test.mjs
```diff
@@ -35,13 +35,15 @@ describe('all rules', () => {
...
+		const cell = rule.meta.fixable ? '\\s+🔧\\s+' : ' ';
+		const wrapper = (x) => {
+			return rule.meta.fixable ? x : x.not;
+		};
 
 		const ruleDoc = await readFile(new URL(`../${name}/README.md`, import.meta.url), 'utf8');
 
-		expect(ruleDoc).toMatch('`fix` option');
-		expect(rulesListDoc).toMatch(new RegExp(`^.+\`${name}\`.+\\|.+\\|\\s+🔧\\s+\\|$`, 'm'));
+		wrapper(expect(ruleDoc)).toMatch('`fix` option');
```
ybiquitous, line 45: [suggest] How about just using `if` block and disabling the lint error like this, instead of a `wrapper` function? It seems a bit readable.

```js
if (rule.meta.fixable) {
	expect(ruleDoc).toMatch('`fix` option'); // eslint-disable-line jest/no-conditional-expect
}
```

ybiquitous (commented):
Nice catch! 👍🏼

lib/rules/__tests__/index.test.mjs
Mouvedia: > It seems a bit readable.

You mean that it hurts the readability?
I was just trying to whip up a quick fix because the current tests let slip that typo.
I think that if we have to add a eslint-disable-line comment it's a sign that we are doing something wrong; **we should have two separate tests**.

ybiquitous: My previous example code was insufficient. The more accurate suggested code is below: ⬇️ 

```js
if (rule.meta.fixable) {
	expect(ruleDoc).toMatch('`fix` option');
} else {
	expect(ruleDoc).not.toMatch('`fix` option');
}
```

This does the same thing as `wrapper`, but more explicitly. I feel `wrapper` is a bit hacky.

> I think that if we have to add a eslint-disable-line comment it's a sign that we are doing something wrong

I don't think so for every `eslint-disable` comment. Linters may produce false positives (as Stylelint does), and sometimes, there are some cases that we can just ignore. 😅 (For example, `eslint-disable no-console` etc.)

I think the `wrapper` function just hides the conditional here and works around `jest/no-conditional-expect` issues.

Mouvedia: Lemme work on something gimme a minute.

Mouvedia: Tell me if 7ccbf91 is to your liking.

ybiquitous: Ah, the updated code uses the negative lookahead assertion (`(?!pattern)`); much better 👍🏼

lib/rules/__tests__/index.test.mjs
```diff
@@ -32,16 +32,17 @@ describe('all rules', () => {
...
+describe('fixability', () => {
+	test.each(ruleNames)('"%s" should be properly documented', async (name) => {
 		const rule = await rules[name];
-
-		if (!rule.meta.fixable) return;
-
+		const cell = rule.meta.fixable ? '\\s+🔧\\s+' : ' ';
+		const rulesPattern = `^.+\`${name}\`.+\\|.+\\|${cell}\\|$`;
+		const needle = '`fix` option';
+		const rulePattern = rule.meta.fixable ? needle : `^((?!${needle})(\\s|\\S))*$`;
```
ybiquitous, line 41: [suggest] I feel a bit more readable if the definitions are in the same order as `expect(...)` lines. How about this?

- `rulePatter` -> `expect(ruleDoc)`
- `rulesPattern` -> `expect(rulesListDoc)`

```suggestion
		const needle = '`fix` option';
		const rulePattern = rule.meta.fixable ? needle : `^((?!${needle})(\\s|\\S))*$`;
		const rulesPattern = `^.+\`${name}\`.+\\|.+\\|${cell}\\|$`;
```

ybiquitous, line 41: [suggest] To be sure, is it not enough to use only `(?!pattern)` form like this?

```suggestion
		const rulePattern = rule.meta.fixable ? needle : `(?!${needle})`;
```

Mouvedia: I cannot test right now. Ill come back to you later.

Mouvedia: I just tried, it doesn't work.

ybiquitous: Okay, let's keep it.

Just curious, can you tell me the reason not to work?

Mouvedia: > Okay, let's keep it.

👍 

> Just curious, can you tell me the reason not to work?

I just put
```
`fix` option
```
in the README of `font-family-no-duplicate-names` and it doesn't fail the test with `(?!${needle})`.
Maybe there's something simpler that works too but I didn't bother.
Regular expressions are hard to get right :)

ybiquitous: I understand why, thanks 👍🏼 

> Regular expressions are hard to get right :)

Absolutely 😅

ybiquitous (approved):
Thank you, LGTM 👍🏼

PR #7722, opened 2024-05-27
Refactor to use shared constant for `disabledRanges` object

ybiquitous:

See #7706

`'all'` is a special rule name for `stylelint-disable` comments. This change makes this name a constant shared in multiple files.

In addition, I think this PR helps us understand the PR #7706 changes more easily. Because this special rule name often confused me when reading the PR (`'all'` or `.all`).


ybiquitous:
@Mouvedia I'm reviewing PR #7706, but understanding the code changes takes time. Can you take a look at this PR to help me understand?

Mouvedia:
This change is compatible with #7706.
Since the removal of `indentation` and `linebreaks` we don't allow rules without hyphens/scopes hence having `all` has a catch-all is safe.

lib/lintPostcssResult.mjs
```diff
@@ -135,7 +136,7 @@ export default async function lintPostcssResult(stylelintOptions, postcssResult,
  */
 function isFixCompatible({ stylelint }) {
 	// Check for issue #2643
-	if (stylelint.disabledRanges.all && stylelint.disabledRanges.all.length) return false;
+	if (stylelint.disabledRanges[RULE_NAME_ALL]?.length) return false;
 
 	return true;
```
Mouvedia, line 141:
```suggestion
	return Boolean(!stylelint.disabledRanges[RULE_NAME_ALL]?.length);
```

ybiquitous: Make sense. Fixed via ee92462be. I omitted `Boolean()` since it would be redundant.

lib/needlessDisables.mjs
```diff
@@ -12,8 +14,8 @@ export default function needlessDisables(postcssResult) {
 
 	if (!options) return;
 
-	const rangeData = postcssResult.stylelint.disabledRanges;
-	const disabledWarnings = postcssResult.stylelint.disabledWarnings || [];
+	const { disabledRanges, disabledWarnings = [] } = postcssResult.stylelint;
+	const disabledRangesAll = disabledRanges[RULE_NAME_ALL] ?? [];
```
Mouvedia, line 18:
```suggestion
	const {
	  disabledRanges,
	  disabledRanges: { [RULE_NAME_ALL]: disabledRangesAll = [] },
	  disabledWarnings = [],
	} = postcssResult.stylelint;
```

lib/utils/report.mjs
```diff
@@ -49,10 +51,10 @@ export default function report(problem) {
 	}
 
 	const { ignoreDisables } = result.stylelint.config || {};
+	const { disabledRanges } = result.stylelint;
```
Mouvedia, line 54:
```suggestion
	const { disabledRanges, config: { ignoreDisables = {} } } = result.stylelint;
```

lib/needlessDisables.mjs
ybiquitous: Thanks. Fixed via 1056ed3ca. 👍🏼

lib/utils/report.mjs
ybiquitous: Unfortunately, the suggested code emits the type error below...

```
TS2339: Property 'ignoreDisables' does not exist on type 'Config | undefined'
```

Mouvedia (approved):

lib/utils/report.mjs
Mouvedia: I think you read my comment before my edit.

ybiquitous:
Thanks for the quick review!

Mouvedia:
> @Mouvedia I'm reviewing PR #7706, but understanding the code changes takes time. Can you take a look at this PR to help me understand?

Until I setup `refactor-autofix-disable` I would advise you to cherry-pick e8cbefd on top of `mouvedia-2643`.
i.e. that way you can easily test it using `./bin/stylelint.mjs --fix foo.css`

ybiquitous:
Ah, thanks for the good advice! 👍🏼

Issue #7723, opened 2024-05-27
Fix `no-descending-specificity` false positives for nested selectors


bbrk24:
### What minimal example or steps are needed to reproduce the bug?

```scss
a {
    color: red;

    .foo & {
        color: blue;
    }

    .foo.bar & {
        color: green;
    }
}

b {
    color: cyan;

    .foo & {
        color: yellow;
    }

    .foo.bar & {
        color: magenta;
    }
}
```

### What minimal configuration is needed to reproduce the bug?

```yml
extends: 
  - stylelint-config-standard-scss
```

### How did you run Stylelint?

`npx stylelint filename.scss` in bash

### Which Stylelint-related dependencies are you using?

```json
    {
        "stylelint": "^16.6.0",
        "stylelint-config-standard-scss": "^13.0.0"
    }
```

### What did you expect to happen?

No errors

### What actually happened?

With the Sass written above, stylelint complains:

> ```
>   16:5  ✖  Expected selector ".foo &" to come before selector ".foo.bar a"  no-descending-specificity
> ```

Obviously, it gives a very similar error message if I put the rules for `b` before the rules for `a`. The only way to avoid this is to manually expand the `&`s and interleave the rules, which defeats the entire point of using Sass.

### Do you have a proposal to fix the bug?

Revert to whatever the logic was in v16.5? There must have been a reason for changing it to what it is now, but with my code I can't get stylelint 16.6 to give me any fewer than *eight* errors for a single Sass file.

romainmenke:
Hi @bbrk24,

Thank you for reporting this and for providing a reproduction.

A slightly more succinct reproduction : https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYAdAdugLgOgDMII8AjJAJxQDI0tddIAbCC+FAcwphkwG56OAL5YR2TKTrZ8RCDSkMczVu25QB0sUJAAaEAQCWTGADkkAWziIYADwsAHY3jABnF7vARMhjghAZMFBR0EFscXigXEPYAbSwgoJCXHABPYyYDTBwAWkhvAw5s5KRMKEooEPiUAF1RDzyfADFWcyQcPwArFy8PWHt3RADEkGS0mAys6OCQJjaYZJCdKqTU9MychoKinBKyigqEadnwhZA6oSA

This is most likely an unexpected and unwanted regressions after this change : https://github.com/stylelint/stylelint/pull/7701

[...]

PR #7724, opened 2024-05-27
Fix `no-descending-specificity` false positives for nested selectors

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/7723

We didn't have coverage for nested selectors where the nested part is the same while the parent is different.

I should have double checked this when refactoring this rule.

I am certain that this case is very common.
Do we have capacity to ship a patch in the next few days?

### DIFF
--- a/.changeset/hot-pens-joke.md
+++ b/.changeset/hot-pens-joke.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `no-descending-specificity` false positives for nested selectors
--- a/lib/rules/no-descending-specificity/__tests__/index.mjs
+++ b/lib/rules/no-descending-specificity/__tests__/index.mjs
@@ -104,6 +104,9 @@ testRule({
 		{
 			code: ':is(.foo) {} :host { .bar {} }',
 		},
+		{
+			code: 'a { .foo.bar & {} } b { .foo & {} }',
+		},
 	],
 
 	reject: [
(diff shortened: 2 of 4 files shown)


Mouvedia (approved):
This LGTM but could you update line 132 to be
```
 * @param {import('postcss-selector-parser').Selector} resolvedSelectorNode
```
?

Mouvedia:
> I am certain that this case is very common.
> Do we have capacity to ship a patch in the next few days?

We have discussed this with @ybiquitous. He's around thankfully but even if he wasn't you can release in his absence.

ybiquitous (approved):
@romainmenke Thanks for the quick patch! LGTM 👍🏼 

If you want, I can grant you permission to publish the `stylelint` package (you don't have it, right?).

I have time to release a new version within a few days, though.

romainmenke:
It's fine that I don't have publishing access.
Users can stay on 16.5 for now.

Having a next patch out in the next few days is still a really quick response, thank you 🙇 

Also thank you for the reviews!

romainmenke:
> This LGTM but could you update line 132 to be
> ` * @param {import('postcss-selector-parser').Selector} resolvedSelectorNode`

That function is correct as it is now imho.

This function does transform the provided selector but it is not this transform that was key to this bug. It was which selector was used as the reference.

The transform could be useful for any selector, even those not produced by resolving nested selectors or when the result isn't used as a reference.

PR #7726, opened 2024-05-28
Ignore `tmp/` from Prettier

ybiquitous:

None.

The `.gitignore` file already has the same setting:
https://github.com/stylelint/stylelint/blob/62b58bc2616fd1420c0adb7090e1e3c03b64597a/.gitignore#L10


romainmenke (approved):
LGTM

PR #7730, opened 2024-05-29
Add `fix` callback to the object passed to `stylelint.utils.report()`

Mouvedia:

#2643, #4573, #4705

- [x] future exposition
  - [x] `range` for #7192
  - [x] `fixed` for #4664
- [x] tests
  - [x] `ignoreDisables`
  - [x] `disableFix`
  - [x] stylelint commands
- [x] add changeset

Documentation will be added in a separate PR.
e.g. rules#add-autofix, plugins#the-anatomy-of-a-plugin


lib/utils/fix.mjs
```diff
@@ -0,0 +1,57 @@
...
+ * @see stylelint/stylelint#7192
+ * @summary apply fix while taking into account the disabled ranges
+ * @param {object} o
+ * @param {import('stylelint').PostcssResult} o.result
+ * @param {string} o.ruleName
+ * @param {FixerData['range']} o.range
+ * @param {FixerData['callback']} o.callback
+ * @param {FixerData['unfixable']} [o.unfixable]
+ */
+const fix = ({ result, ruleName, range, callback, unfixable }) => {
```
Mouvedia, line 25: @romainmenke 
This PR addresses your demands, namely:

- rename to `fix`
- flatten the object i.e. share the same shape
- immediate instead of delayed

I didn't add many tests though.
It's not ready to review yet though.

ybiquitous (changes requested):
@Mouvedia I appreciate your effort to address the very old issue. However, it seems that we cannot accept this PR easily because:

- This PR adds a few features is not a refactoring. The PR title is misleading.
- This PR is too big to discuss each topic.
- This PR does not add enough test cases for the new changes.

ybiquitous:
[note] Related (closed) PRs:

- #7706 
- #7729

Mouvedia:
> @Mouvedia I appreciate your effort to address the very old issue. However, it seems that we cannot accept this PR easily because:
> 
> * This PR adds a few features is not a refactoring. The PR title is misleading.

I can edit the title.

> * This PR is too big to discuss each topic.

I can move the 3 rule refactors and related tests to another PR if you want.
That would remove about 16 files. It will be easier to review that way.

> * This PR does not add enough test cases for the new changes.

If you have suggestion for tests, Ill add them but that would be on the other PR because it relies on having refactored rules.

lib/lintPostcssResult.mjs
```diff
@@ -95,7 +95,11 @@ export default async function lintPostcssResult(stylelintOptions, postcssResult,
 		// Log the rule's severity in the PostCSS result
 		const defaultSeverity = config.defaultSeverity || 'error';
 		// disableFix in secondary option
-		const disableFix = (secondaryOptions && secondaryOptions.disableFix === true) || false;
+		const disableFix = secondaryOptions?.disableFix || false;
+		const fileFix = !disableFix && config.fix && isFileFixCompatible;
+		// until #2643 is resolved
+		const ruleFix =
```
Mouvedia, line 101: medium

This condition should use `ruleMetadata` instead of `safeDisables`.
i.e. it's rule specific not file specific

Mouvedia:
@ybiquitous

> @Mouvedia I appreciate your effort to address the very old issue.

Thanks.

> However, it seems that we cannot accept this PR easily because:
> 
> * This PR adds a few features is not a refactoring. The PR title is misleading.

I edited the title.

> * This PR is too big to discuss each topic.

I removed 3 rule refactors and their related tests.
That clocks at about 16 files less. It should be easier to review that way.

> * This PR does not add enough test cases for the new changes.

I have added 7 plugin tests and 4 skipped tests.
The skipped tests require at least 2 core rules to be refactored.

ybiquitous:
Can you explain why this new flag and option are needed so that maintainers can understand?

Mouvedia:
> Can you explain why this new flag and option are needed so that maintainers can understand?

In the previous PR, the plan was to add the refactors on the `main` branch and test the _stylelint commands_ interaction on an experimental branch. That would have been really tedious/inconvenient for the reviewers unless all the refactors were added in less than 3 batches—which is unlikely.

We will add a `describe.each` that will loop on the core rules that have a truthy `RuleMeta['fixer']` and run a bunch of tests _directly_ on the main branch.
Once all the fixable rules have been migrated we can decide whether we want to change the default of the flag/option to `false` or not. The past changes related to that behaviour were introduced either in a minor (13.2.0) or a patch (13.3.3) hence it could be reasonable to consider it wouldn't be a breaking change; that's still up for debate though.

tl;dr: to ease the refactor reviews, to centralize the changes and to facilitate the hypothetical switch of the default behaviour

romainmenke:
> Can you explain why this new flag and option are needed so that maintainers can understand?

I have the same question :)

If I understand it correctly this flag/option is added mostly as an opt-in for experimental behavior? If so, I find it confusing as `--safe-disables` is not a feature we really want to add and expose to users, right?

We also have other tools available to us to aid in reviews and to get early feedback on partial work. We don't need to merge partial work to main.

Given the complexity of the problem in https://github.com/stylelint/stylelint/issues/2643 I suspect we will have several rounds of feedback to find the correct approach. I think it's fine if we have a longer lived branch and make sure we get this right.

_I don't mean to say that this PR should be closed/restarted. I hope it can be re-focussed on purely fixing https://github.com/stylelint/stylelint/issues/2643_

-----

Going through the code next :)

lib/utils/fix.mjs
romainmenke, line 1: I still don't really understand why we can't use `report()`.
Or why `fix()` is still substantially different when it is separate.

This seems harder to maintain to me.
Things like the range checks are non-trivial and prone to "off by one" errors. Having two implementations of this will lead to hard to debug problems.

-------

Example of a `fix()` implementation that is based on `report()`:

_I did the minimal effort to merge these, this is in no way a perfect thing :)`

```js
/** @typedef {import('stylelint').FixerData} FixerData */

import { RULE_NAME_ALL } from '../constants.mjs';

/**
 * even though stylelint-disable comments cannot be inserted inside a declaration or a selector list,
 * new lines cannot be disregarded because FixerData['range'] is exposed through StylelintPostcssResult['fixersData']
 * i.e. ranges must be accurate to be exploited
 * @see stylelint/stylelint#7192
 * @summary apply fix while taking into account the disabled ranges
 * @param {object} o
 * @param {import('stylelint').PostcssResult} o.result
 * @param {import('postcss').Node} o.node
 * @param {number} o.index
 * @param {number} o.endIndex
 * @param {number} o.line
 * @param {string} o.ruleName
 * @param {FixerData['range']} o.range
 * @param {FixerData['callback']} o.callback
 * @param {FixerData['unfixable']} [o.unfixable]
 */
export default function fix({
	result,
	ruleName,
	node,
	index,
	endIndex,
	line,
	callback,
	unfixable,
}) {
	const fixableRange = (node && node.rangeBy({ index, endIndex })) || {};

	const startLine = line || fixableRange.start?.line;

	if (!startLine) {
		throw new Error(
			`The "${ruleName}" rule failed to pass either a node or a line number to the \`fix()\` function.`,
		);
	}

	if (unfixable) {
		addFixerData(result, ruleName, { range: fixableRange, fixed: false });

		return;
	}

	const { ignoreDisables } = result.stylelint.config || {};
	const { disabledRanges } = result.stylelint;

	if (disabledRanges) {
		const ranges = disabledRanges[ruleName] ?? disabledRanges[RULE_NAME_ALL] ?? [];

		for (const range of ranges) {
			if (
				// If the problem is within a disabledRange,
				// and that disabledRange's rules include this one,
				// do not attempt to fix it
				range.start <= startLine &&
				(range.end === undefined || range.end >= startLine) &&
				(!range.rules || range.rules.includes(ruleName))
			) {
				addFixerData(result, ruleName, { range: fixableRange, fixed: false });

				if (!ignoreDisables) {
					return;
				}

				break;
			}
		}
	}

	callback();

	addFixerData(result, ruleName, { range: fixableRange, fixed: true });
}

/**
 * @param {string} name
 * @param {{ range: FixerData['range'], fixed: boolean }} data
 */
function addFixerData(result, name, data) {
	const list = result.stylelint.fixersData[name];

	if (!list) result.stylelint.fixersData[name] = [data];
	else list.push(data);
}
```

Also notice how this becomes easier to use from the perspective of the caller :

```js
if (context.fix) {
	return fix({
		result,
		ruleName,
		callback: () => {
			ruleNode.selector = selector.replace('foo', 'bar');
		},
		node: ruleNode,
		index: i,
		endIndex: i + 'foo'.length,
	});
}

report({
	result,
	ruleName,
	message: messages.rejected(selector),
	node: ruleNode,
	index: i,
	endIndex: i + 'foo'.length,
});
```

They are the same, except one has `message` and the other has `callback`.

I still think it would be better as :

```js
report({
	result,
	ruleName,
	message: messages.rejected(selector),
	fix: () => {
		ruleNode.selector = selector.replace('foo', 'bar');
	},
	node: ruleNode,
	index: i,
	endIndex: i + 'foo'.length,
});
```

But I also want to hear others opinion on this before changes are made :)

Mouvedia:
> If so, I find it confusing as `--safe-disables` is not a feature we really want to add and expose to users, right?

The users will need it when we are done with the migration.
The reasons being if

1. we keep the safe behaviour as default then the users will want a way to switch that fix behaviour on files that only report errors/warnings for core rules or plugins that have been migrated
2. we change to the unsafe behaviour by default then the users will want a way to switch it off if they have files with stylelint commands that report errors/warnings for plugins that don't use the `fix` util

Do we absolutely need it now? No we don't. But it sure is useful to not have to jungle between `main` and an experimental branch. And we will eventually have to add it, so it doesn't hurt.

> Going through the code next :)

Thanks.
I am busy today but Ill check your comments tonight.

romainmenke:
> The reasons being if
> - we keep the safe behaviour as default then the users will want a way to switch that fix behaviour on files that only report errors/warnings for core rules or plugins that have been migrated
> - we change to the unsafe behaviour by default then the users will want a way to switch it off if they have files with stylelint commands that report errors/warnings for plugins that don't use the fix util

Can't we make this work as expected in all cases without such a flag?

This seems solvable to me.

We only want to retain the old behavior of skipping fixes for old plugins, plugins that do not have `meta.fixer`.

I might be overlooking something here :)

Mouvedia:
> I might be overlooking something here :)

A simplistic approach would be to detect whether the current config contains at least one rule with a `fixer` falsy.  
What's desirable would be to know dynamically. But even if we did check while running the rule, we couldn't know whether the plugin rules have a warning/error in advance.

To summarize even if we were to run the rules in a predefined order, it would require to run the plugin rules twice.
i.e. put all the plugin rules without a truthy `fixer` first, and if none are reporting warnings/errors during the dry run for that file switch to the unsafe behavior and start the fix run for all rules

tl;dr: having a flag/option is simpler and faster

romainmenke:
@Mouvedia can you explain what goes wrong?
You immediately explain the remediation, but you don't say what you are preventing.
This makes it very hard for me to understand why this flag/option is needed.

Mouvedia:
Here's an example.

A code base which may or may not have stylelint commands, which themselves could be scoped or not.
You are using the version of stylelint which has migrated all the fixable core rules.
Some of your plugins may or may not have been migrated, as a user you don't want to investigate.
You want to fix the files that have disable commands without any worry.

<sub>safe behaviour means #4573 + #4705 (statu quo)</sub>

There are 3 **non-equivalent** alternatives that may be considered:

1.
We check all the rules of the resolved config and if we find a rule which has `fixer == true` we fallback to the safe behaviour for all the files that contain disable commands. As a user, it would require to create two configs, one with the migrated rules and one with rest. You would have to run stylelint twice.

2.
We check the resolved config and separate the rules which match `fixer == false`. We do a dry run for these and compile the files that don't have any errors/warnings reported. We switch to the unsafe behaviour during the fix run that immediately follows only for those aforementioned files; the rest being fixed using the safe behaviour.

3.
We add a flag/option that switch the behaviour. As a user you have to opt-out of the default which will probably be the safe behaviour. It gives more control to the user. Like 1. it may require two configs depending on the chosen linting strategy. We could even add a sentence to the verbose formatter that list the rules incompatible with the flag/option.
i.e. if he decides to do `--fix --no-safe-disables` followed by a `git add -p` he can

lib/__tests__/fixtures/plugin-meta-fixer.mjs
```diff
@@ -0,0 +1,63 @@
...
+
+		if (!validOptions) return;
+
+		root.walkRules((ruleNode) => {
+			const { selector } = ruleNode;
+			const i = selector.indexOf('foo');
+
+			if (i === -1) return;
+
+			const range = ruleNode.rangeBy({
```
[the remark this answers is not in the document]
Mouvedia: Not all fixer refactors will use `RangeBy` to generate `range`.
e.g. `selector-not-notation` has [`getRange`](https://github.com/stylelint/stylelint/pull/7731/commits/36586ed4652dcaa19446612187e2c9583af02a4e#diff-bd94fe99a5bd46c7e6b11e8afd5f71ca4aa1897e2d52801cc1e7e3be5de7d925R217-R231)
i.e. `RangeBy` will be prevalent/recommended but not required

lib/utils/fix.mjs
Mouvedia: > Having two implementations of this will lead to hard to debug problems.

Most of the refactors will pass `...range` to report.
i.e. it will be shared by `fix` and `report`

romainmenke:
Unfortunately no. Again you are explaining what you want to do, what you want users to be able to do, but you aren't saying what problem is in the way.

What does `--safe-disables` change or enable that isn't possible without it?

----------

However, I was curious about what you could have meant and I think I might have found it.
Not totally sure.

In `lib/lintPostcssResult.mjs`: 

```js
const isFileFixCompatible = isFixCompatible(postcssResult);

if (!isFileFixCompatible) {
 	postcssResult.stylelint.disableWritingFix = true;
}
```

`postcssResult.stylelint.disableWritingFix` is set to true when disable comments are present in a file.

When this flag is set to true `result.code` will always be the original CSS.
Even when fixers run, the fixed CSS won't be what is stored in `result.code`.

Adding an extra flag allows you to bypass the `postcssResult.stylelint.disableWritingFix`

Is that correct?

Mouvedia:
@romainmenke yes that's about right I though you knew about the PRs that I was referencing.
i.e. one of these introduced `disableWritingFix`
The other PR introduced a limitation related to scoped stylelint commands.
i.e. `/* styleint-disable foo-bar */`

So in short, the flag/option reverts what has been introduced as safe measures by these 2 PRs.
Sorry if I didn't understand what you meant.

romainmenke:
I understand now that it is unknowable beforehand that some plugins might ignore `context.fix` being false and will mutate the PostCSS AST.

However is that something that should influence this work to this extends?

This only affects rules that do not honor `context.fix` and only in files with disable comments. Such rules would already be very buggy as they would apply fixes even when a user doesn't want them to. They would only stop applying fixes in files that have disable comments.

Any rule that does honor `context.fix` isn't a problem for the work in this PR, right?

Some open questions:
- is this a real problem or hypothetical, the original [PR](https://github.com/stylelint/stylelint/pull/4573) doesn't seem to explain why this snippet was needed. Why wasn't it sufficient to set `context.fix` to `false`?
- is this something we want/need to preserve?

Mouvedia:
_currently_

A file that contains `/* stylelint-disable */` won't be fixed at all.
A file that contains `/* stylelint-disable foo-bar */` won't fix foo-bar rule but will fix the rules that themselves don't have a corresponding stylelint-disable scoped comment in the same file.

> This only affects rules that do not honor `context.fix`

I don't think that's correct. Or I might have misunderstood you again.

> is this something we want/need to preserve?

If by _this_ you mean the changes introduced by the aforementioned 2 PRs.
I would say that my aim was for the refactors to be transparent for the users.
If we were to switch the behaviour—i.e. 6e7afd0—it could almost be considered a breaking change.
i.e. all the non-migrated plugin rules would not honor the stylelint commands and people would complain (because `--fix` is now considered safe)

romainmenke:
> > This only affects rules that do not honor context.fix
> I don't think that's correct. Or I might have misunderstood you again.

see : https://github.com/stylelint/stylelint/pull/7733/files

_Removing that doesn't even trigger test failures._

Mouvedia:
> > > This only affects rules that do not honor context.fix
> > > I don't think that's correct. Or I might have misunderstood you again.
> 
> see : [#7733 (files)](https://github.com/stylelint/stylelint/pull/7733/files)
> 
> _Removing that doesn't even trigger test failures._

Are tests failing if you copy what I did in 6e7afd001?
i.e. [these lines](https://github.com/stylelint/stylelint/pull/7733/files#diff-8ba0772440729ba7a5ddd0acc62ed5374586c296388e4a1f29c2f01f33993adfL110-L115) are untouched in your change

romainmenke:
Yes, they are untouched and that is exactly my point. 

There are currently two mechanics to prevent fixes when disable comments are present while only one is really needed.
- `context.fix` is set to `false`
- `postcssResult.stylelint.disableWritingFix` is set to true.

`context.fix` needs to be honored by rule authors (e.g. `if (context.fix) { /* mutations here */ }`)

While `postcssResult.stylelint.disableWritingFix` is a catch-all that doesn't actually stop any AST mutations, but rather forces `result.code` to be the original input.

`--safe-disables` isn't needed for rules that honor `context.fix`, it is only needed for those that don't.

Arguably it is wrong if rules apply fixes without considering `context.fix`.

I don't know why this was written in this way and https://github.com/stylelint/stylelint/pull/4573 doesn't explain this in detail.

Mouvedia:
The line `isFileFixCompatible &&` could have been written `!postcssResult.stylelint.disableWritingFix &&`.
If it was, would you have removed l114 of `lintPostcssResult`?

<hr/>

`context.fix` is a _per rule_ boolean
`disableWritingFix` is a _per file_ boolean

The `fix` coming from the config/flag is never mutated and the `fix` passed to `ruleFunction` is never stored.

Maybe I am completely missing your point. All I gotta say about #4573 is that it's a hack, and a hard revert would be ill advised.

romainmenke:
Please look at where `postcssResult.stylelint.disableWritingFix` is set and used.

It is only set once here : https://github.com/stylelint/stylelint/blob/b5b3799cc107ede5b418bef6ef64bab072b2deab/lib/lintPostcssResult.mjs#L45

It is only used twice :
- https://github.com/stylelint/stylelint/blob/b5b3799cc107ede5b418bef6ef64bab072b2deab/lib/standalone.mjs#L148
-  https://github.com/stylelint/stylelint/blob/b5b3799cc107ede5b418bef6ef64bab072b2deab/lib/standalone.mjs#L237


It is set to `true` when `isFixCompatible` is `false`.

When `true` it forces Stylelint to output the original CSS even if the user requested fixes to be applied.

------

But `isFixCompatible` being `false` also causes `context.fix` to be `false`.

See : https://github.com/stylelint/stylelint/blob/b5b3799cc107ede5b418bef6ef64bab072b2deab/lib/lintPostcssResult.mjs#L108-L116

From this we know with absolute certainty that `context.fix` will always be `false` when `postcssResult.stylelint.disableWritingFix` is `true`.

In other words rules that honor `context.fix` will never apply fixes in this case.

------

The only scenario in which `postcssResult.stylelint.disableWritingFix` has an observable effect is :
- if there is a rule that always mutates the AST. it doesn't honor `context.fix`
- the user has any disable comments in the source
- the user requests the source code to be fixed

In that case `postcssResult.stylelint.disableWritingFix` will prevent the badly behaved Stylelint rule from altering the source code.

There is no scenario in which a rule honors `context.fix` and it still requires `postcssResult.stylelint.disableWritingFix`.

I will open an issue to discuss removing this entirely.

Mouvedia:
Now I understand what you meant.
There must have been plugins that didn't respect `context.fix` usage convention in the wild.

lib/utils/fix.mjs
ybiquitous: I'd prefer `report()` over `fix()` because of:

- low learning cost by enhancing the existing API
- similar to the ESLint API
- less duplication between `report()` and `fix()`

Mouvedia: Alright, Ill attempt the change on the other PR after #7735 resolution.
Ill keep the previous commits so that both approaches can be compared.

Mouvedia: a04fdba

Mouvedia:
I have rebased due to #7733.

lib/__tests__/fixtures/plugin-meta-fixer.mjs
```diff
@@ -0,0 +1,59 @@
...
+			}
+
+			report({
+				result,
+				ruleName,
+				message: messages.rejected(selector),
+				node: ruleNode,
+				...range,
+				fix,
+			});
```
romainmenke, line 50:
```suggestion
			report({
				result,
				ruleName,
				message: messages.rejected(selector),
				node: ruleNode,
				...range,
				fix: () => {
					ruleNode.selector = selector.replace('foo', 'bar');

					return range;
				},
			});
```

You can internalize the logic on when to apply `fix`.

If you do, you can remove the flag/option added in this PR.
This was part of the motivation behind : https://github.com/stylelint/stylelint/pull/7733

Instead of requiring users to set a flag you can dynamically pick the expected behavior.

----

Is it fine if I demonstrate how this could work with in a pull request?

Mouvedia: So you wanna pass `fix`, `unfixable` _and_ `context` (or `context.fix`) to `report`?
If you meant doing
```
				fix: context.fix ? (() => {
					ruleNode.selector = selector.replace('foo', 'bar');
					return range;
				}) : undefined,
```
instead, I don't think that would be cleaner.
Also, you have rules with multiple fixers so inlining wouldn't be recommended in that case.

> Is it fine if I demonstrate how this could work in a pull request?

I would prefer if you opened another PR.

romainmenke: I've opened another PR to discuss what I had in mind : https://github.com/stylelint/stylelint/pull/7741

Mouvedia:
@romainmenke I have made the changes requested. I have also updated [OP](https://github.com/stylelint/stylelint/pull/7730#issue-2324194258).
We cannot reuse this for loop:

https://github.com/stylelint/stylelint/blob/9a2e6540f92509628df76032262293ae57d32dda/lib/utils/report.mjs#L53

because it either contains the scoped ranges for the rule or the unscoped ranges, not both.
Another reason which is less important is that having them separated is easier to maintain and to reason about.
From a performance point of view it doesn't matter because it's either fix or warn.

### codecov

The second warning is unrelated to this PR.
The first one, Ill comment directly on the line to have your feedback.

### documentation

It will require a lot of work so Id prefer if we separated into its own PR.
We should release once at least one refactor has been added and the documentation is on point.

### follow up PRs are

- warning related to the use of `context.fix` and/or absence of `RuleMeta['fixer']`
- documentation
- refactor of rules (one PR per refactor)

How does that sound ?

lib/utils/applyFix.mjs
```diff
@@ -0,0 +1,47 @@
...
+	const mayFix = config?.ignoreDisables || !ranges.length || !ranges.some(isInRange);
+	const array = fixersData[ruleName] || (fixersData[ruleName] = []);
+
+	try {
+		if (mayFix) fix();
+
+		array.push({ range, fixed: mayFix });
+	} catch {
+		array.push({ range, fixed: false });
+	}
```
Mouvedia, line 46: @ybiquitous @romainmenke 

I consider that the `fix` may throw and that is not up to us to dictate that behaviour.
If so should I update the return type of `fix`?
e.g. `void | never`

types/stylelint/index.d.ts
```diff
@@ -110,7 +110,7 @@ declare namespace stylelint {
 		rules?: ConfigRules;
 		quiet?: boolean;
 		defaultSeverity?: Severity;
-		ignoreDisables?: DisableSettings;
```
Mouvedia, line 113: see https://github.com/stylelint/stylelint/commit/b47df94515ca8e1114216e2345ed7e12030561ab#diff-9a4b261e295bef1ccce813aa194ac4023a1dcf98236667889964b1ba6fbc551fR42

lib/utils/applyFix.mjs
romainmenke: I think we can say that `fix` must not throw.

We definitely can't erase errors as this code would do because that would make it much harder to surface bugs in fixers.

So I would remove this `try {} catch {}`.

Mouvedia: > I think we can say that `fix` must not throw.

I think you are right for one more reason: it might throw after the fix.

lib/__tests__/fixtures/plugin-meta-fixer.mjs
```diff
@@ -0,0 +1,48 @@
...
+					ruleNode.selector = selector.replace('foo', 'bar');
+				},
+			});
+		});
+	};
+};
+
+rule.ruleName = ruleName;
+rule.messages = messages;
+rule.meta = { fixer: true };
```
romainmenke, line 46: `fixer` is not needed right at a functional level, right?

Should we omit it altogether?

lib/utils/report.mjs
```diff
@@ -15,13 +17,38 @@ import { RULE_NAME_ALL } from '../constants.mjs';
...
+		const range = hasRange ? { start, end } : node.rangeBy({ index, endIndex });
+
+		if (hasFix) return applyFix({ fix, result, ruleName, range });
+	}
 
 	const ruleSeverityOption = severity || result.stylelint.ruleSeverities[ruleName];
-
-	const defaultSeverity = result.stylelint.config?.defaultSeverity || 'error';
-
+	const defaultSeverity = config?.defaultSeverity || 'error';
```
Mouvedia, line 51: nit

config? > config
l51 too

lib/__tests__/fixtures/plugin-meta-fixer.mjs
Mouvedia: I thought about it. It depends on how you want the granularity of the warning to be for unmigrated plugins.
i.e. either we trust the `fixer` property or we assume trickery from plugin authors and add a getter to `context.fix`

What do you reckon?

romainmenke:
> because it either contains the scoped ranges for the rule or the unscoped ranges, not both.

You mean this : 

```js
const ranges = disabledRanges[ruleName] ?? disabledRanges[RULE_NAME_ALL] ?? [];
```

Which ignores the ranges in `disabledRanges[RULE_NAME_ALL]` is values are present in `disabledRanges[ruleName]`.

Is this a bug?

Should it be something like:

```js
const ranges = [
  ...disabledRanges[ruleName] ?? [],
  ...disabledRanges[RULE_NAME_ALL] ?? []
];
```

Either way this alone doesn't justify duplicating the logic for the range checks.

-------

> Another reason which is less important is that having them separated is easier to maintain and to reason about.

I strongly disagree here.
If you had used the exact same implementation it could have been debatable but you implemented a complex bit of code in a very different way.

A change that needs to happen in both places will require twice the effort.

I really think we can and should fully integrate the fixers into `report()` and only once have the complexity of checking if something is in a range.

Mouvedia:
> Is this a bug?

see https://github.com/stylelint/stylelint/pull/7741#discussion_r1629819352

Mouvedia:
> A change that needs to happen in both places will require twice the effort.

Can you give an example?
It seems to me way cleaner to have a well defined function that does only one thing.

> I really think we can and should fully integrate the fixers into report()…

That's already the case.

> …and only once have the complexity of checking if something is in a range.

This is gonna need to fix a bug, add a regression test and complicate the existing code further.
Let's postpone that refactor for the bug fix?

lib/__tests__/fixtures/plugin-meta-fixer.mjs
Mouvedia: @ybiquitous what's your take on this?
i.e. what should trigger a warning/output for plugins that haven't been migrated yet.
e.g. when the formatter is verbose we could have a count of the outdated plugins and a link to the documentation

romainmenke:
Looked into `const ranges = disabledRanges[ruleName] ?? disabledRanges[RULE_NAME_ALL] ?? [];` and there is no bug here.

As far as I can tell all ranges present in `disabledRanges[RULE_NAME_ALL]` are also always present in any `disabledRanges[ruleName]` if a file contains both rule specific and `all` disable comments.

There are three scenario's as far as I can tell:

(1)

- only `all` disables
- `disabledRanges[ruleName]` is empty
- `disabledRanges[RULE_NAME_ALL]` is populated

```css
/* stylelint-disable */
.foo {}
/* stylelint-disable */
```

(2)

- only rule specific disables
- `disabledRanges[ruleName]` is populated
- `disabledRanges[RULE_NAME_ALL]` is an empty list

```css
/* stylelint-disable plugin/foo */
.foo {}
/* stylelint-disable plugin/foo */
```

(3)

- mixed disables
- `disabledRanges[ruleName]` is populated
- `disabledRanges[RULE_NAME_ALL]` is populated
- all disable ranges present in `disabledRanges[RULE_NAME_ALL]` are also present in each `disabledRanges[ruleName]`

```css
/* stylelint-disable */
.foo {}
/* stylelint-disable */

/* stylelint-disable plugin/foo */
.foo {}
/* stylelint-disable plugin/foo */
```

```css
/* stylelint-disable plugin/foo */
.foo {}
/* stylelint-disable plugin/foo */

/* stylelint-disable */
.foo {}
/* stylelint-disable */
```

-----

This explains the shape of the code in `report()` and why `const ranges = disabledRanges[ruleName] ?? disabledRanges[RULE_NAME_ALL] ?? [];` works as expected.

> because it either contains the scoped ranges for the rule or the unscoped ranges, not both.

So I would say that this premise is untrue.

`ranges` from `const ranges = ...` always contains all the disable ranges. 

-----

Am I overlooking something here?
I tried to test this really carefully.

Mouvedia:
So it's by design then. I have added a regression test just in case.
Ill try to refactor and share the same for loop then.
No need to `concat` that's better for performance.

lib/utils/report.mjs
```diff
@@ -15,13 +17,38 @@ import { RULE_NAME_ALL } from '../constants.mjs';
...
+
+	const hasRange = end?.column && start;
+	const hasIndexes = isNumber(index) && isNumber(endIndex);
+	const hasFixData = config.fix && (hasRange || hasIndexes);
+
+	if (hasFixData) {
+		const hasFix = typeof fix === 'function';
+		const range = hasRange ? { start, end } : node.rangeBy({ index, endIndex });
+
+		if (hasFix) return applyFix({ fix, result, ruleName, range });
```
romainmenke, line 47: Validating the input and building ranges is also duplicated between warnings and fixes.

Can we remove this duplication?

Is the end position critical in `FixerData.range`? Making this optional would align better with warnings and disable ranges.

Mouvedia: It is critical to introduce this expectation for future refactors.
We want all plugins authors that switch to the fix callback to at least include `endIndex` so that #7192 can be achieved. It's not a lot of effort on their part once they have started the process and it will be a clear win for the community in general. Our effort started in #5694 and we _still_ have core rules that don't have an `endIndex`.

e.g.

- stylelint commands will be suddenly honored
- you will have a _precise_ wavy underline for the errors/warnings
- a few release later we will expose the errors/warnings for other tools to exploit

romainmenke: 🤔 I agree.

I need to think on what the best shape could be for this code to aid in this transition.

At this time I would suggesting throwing if a `fix` argument is present and it isn't possible to construct a full range. That is the clearest signal to rule authors.

But I might gain new insights if I think about it more :)

Mouvedia:
I tried to share the for loop but it would be too complicated.
We would be doing a lot of unnecessary logic that we don't need for just fixing.
We would have to change the `if` of
```
	if (result.stylelint.quiet && ruleSeverity !== 'error') {
		return;
	}
```
Overall the readability would also be worse.

On the good side, thanks to your help I could remove the "complex bit of code" that you were talking about.
The only thing that is left is whether `applyFix` merit being a util or not. I don't think it might be reused.
Should I add it directly in `report.mjs`?

lib/utils/report.mjs
Mouvedia: > At this time I would suggesting throwing if a fix argument is present and it isn't possible to construct a full range. That is the clearest signal to rule authors.

That's interesting.

lib/utils/report.mjs
```diff
@@ -15,13 +16,38 @@ import { RULE_NAME_ALL } from '../constants.mjs';
...
+		ruleName,
+		severity,
+		word,
+	} = problem;
+
+	const hasRange = end?.column && start;
+	const hasIndexes = isNumber(index) && isNumber(endIndex);
+	const hasFixData = config.fix && (hasRange || hasIndexes);
+
+	if (hasFixData) {
```
Mouvedia, line 42: > [At this time I would suggest throwing if a `fix` argument is present and it isn't possible to construct a full range. That is the clearest signal to rule authors.](https://github.com/stylelint/stylelint/pull/7730#discussion_r1630258272)

lib/utils/report.mjs
Mouvedia: > At this time I would suggesting throwing if a fix argument is present and it isn't possible to construct a full range. That is the clearest signal to rule authors.

That's interesting.
I went ahead and added the `throw` and its test.

lib/__tests__/fixtures/plugin-meta-fixer.mjs
ybiquitous: This new `fixer` property would be temporary, right? If all plugins migrated to the new API, would this property become unnecessary?

Also, I wonder if plugin authors would feel annoyed to have to add `fixer`.

If the suggested `context.fix` change works, I am in favor of `context.fix`.

Mouvedia: > This new fixer property would be temporary, right? If all plugins migrated to the new API, would this property become unnecessary?

The real benefit would be for us to know **without** linting if some outdated rules are present in the resolved config.
i.e. without `fixer` we wouldn't be able to implement a `--outdated` flag to list those rules

> I am in favor of `context.fix`

If @romainmenke agrees too Ill remove it.

ybiquitous: But, having to specify both `fix` and `fixer` doesn't make sense, right?

Mouvedia: If the name the problem, it can be changed. Anyway I thought having that information in `meta` would be more versatile.
Ill just remove it for now; it's not needed for that particular PR anyway.

lib/utils/report.mjs
```diff
@@ -15,13 +16,45 @@ import { RULE_NAME_ALL } from '../constants.mjs';
...
+			`The fix callback for rule "${ruleName}" requires either index/endIndex or start/end to be passed to the \`report()\` function.`,
+		);
+	}
 
-	const ruleSeverityOption = severity || result.stylelint.ruleSeverities[ruleName];
+	if (hasFixData) {
+		const range = hasRange ? { start, end } : node.rangeBy({ index, endIndex });
 
-	const defaultSeverity = result.stylelint.config?.defaultSeverity || 'error';
+		if (hasFix) return applyFix({ fix, result, ruleName, range });
```
Mouvedia, line 53: major

`config.rules[ruleName][1].disableFix` is missing.

lib/utils/report.mjs
```diff
@@ -15,13 +16,46 @@ import { RULE_NAME_ALL } from '../constants.mjs';
...
+	const hasFixData = hasFix && hasRangeData;
+
+	if (hasFixCallback && !hasRangeData) {
+		throw new Error(
+			`The fix callback for rule "${ruleName}" requires either index/endIndex or start/end to be passed to the \`report()\` function.`,
+		);
+	}
 
-	const ruleSeverityOption = severity || result.stylelint.ruleSeverities[ruleName];
+	if (hasFixData) {
```
Mouvedia, line 51: note

I chose not to expose the `FixerData` in the case of `context.fix` usage. It wouldn't incentivize enough to move towards using a fix callback.
This decision can be changed later on.

types/stylelint/index.d.mts
```diff
@@ -13,6 +13,7 @@ export {
 	DisabledRange,
 	DisabledRangeObject,
 	DisabledWarning,
+	FixerData,
```
romainmenke, line 16: Do we need to bikeshed the new API?

romainmenke (commented):
Thank you @Mouvedia for your continued efforts of this change 🙇 

Some final thoughts, nothing blocking.

types/stylelint/index.d.ts
Mouvedia: I have reviewed the usage of `ignoreDisables` in the codebase and the documentation.
It's a `boolean`, i.e. not `DisableSettings`.

Iv just linked when it was changed, to give some context.

types/stylelint/index.d.mts
Mouvedia: Not really, this is internal and follows what we have done so far for similar data.
The real bikeshedding will be about the details of the exposition of `range` and `fixed` in JSON.

e.g.
do we create `text` from `range`?
what should be the shape of the object?
etc.

In short this is on topic for #7192 and #4664.

lib/__tests__/fixtures/plugin-fixer.mjs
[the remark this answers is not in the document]
Mouvedia: > Maybe this can be part of `plugin-fixes.mjs` ?

I suppose you mean `plugin-fixer.mjs`.
If so do you mean replace?

I think you just want me to add a new test.
I don't mind acknowledging that plugin authors may do unconventional things.

types/stylelint/index.d.mts
Mouvedia: Thinking about it. You have a point.
From a user standpoint, at this stage having `Range` is more useful than `FixerData`.
i.e. the only usage in JSDoc is `FixerData['range']`

change incoming

romainmenke: Before making further changes we should design this :)
Maybe the current form is the best?

--------

A way to get around this, at this time, is to not expose the `FixerData` and to limit the API surface additions to the `fix()` function.

Can we add the `fix()` function, collect the `FixerData` but without exposing any other new API surface or types than `fix()`?

Maybe even preferable to not collect the `FixerData` just yet?

The main purpose is to make this change easier to review and to find consensus on all parts :)

lib/__tests__/fixtures/plugin-fixer.mjs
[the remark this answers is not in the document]
Mouvedia: 782828a

lib/utils/report.mjs
```diff
@@ -146,3 +181,37 @@ function printfLike(format, ...args) {
...
+	const isInRange = (/** @type {DisabledRange} */ { start, end }) =>
+		range.start.line >= start && (!end || range.start.line <= end);
+	const ranges = disabledRanges[ruleName] ?? all;
+	const mayFix = config?.ignoreDisables || !ranges.length || !ranges.some(isInRange);
+	const array = fixersData[ruleName] || (fixersData[ruleName] = []);
+
+	array.push({ range, fixed: mayFix });
+
+	if (mayFix) fix();
+}
```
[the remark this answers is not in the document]
Mouvedia: I think it's better for maintenance, for documentation, for keeping the blame, separating concerns.
In this case, after this is merged you can attempt to refactor and see for yourself if it's really worth it.
From what I gathered, it would also be a regression hazard; i.e. it's safer not to refactor.
The logic behind `isInRange` do look similar indeed, but there are differences.
I disagree that this is still "complex" though.

types/stylelint/index.d.mts
Mouvedia: This PR is a precursor for 3 issues by design.
If we want to decide to start working on any of the 3, we can.
It's meant to be an unblocker.

i.e.

If you add a refactor, you have the expectation related to the `throw` that we added.
So if anyone fixes #7192 while we have partially migrated the core rules, it will be available.
Same thing for #4664.
`Range` won't change and it's useful so it's OK to expose.
BTW `FixerData` is not exported anymore.

I could maybe not expose it if JSDoc supports something crazy like `(typeof StylelintPostcssResult['FixersData'][keyof typeof StylelintPostcssResult['FixersData']])[number]['range']`.
IMHO it's not worth the effort.

lib/utils/applyFix.mjs
Mouvedia: `selector-type-case` is an example of such cases.
i.e. its fixer will throw

I will add `never` to the return type.

lib/utils/report.mjs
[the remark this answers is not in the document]
Mouvedia: > do not have half of it in `report`

I am gonna move that part in a `handleFix` function.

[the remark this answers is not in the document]
Mouvedia: > do not have half of it in `report`

Ill try something.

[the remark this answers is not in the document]
romainmenke: > The logic behind isInRange do look similar indeed, but there are differences.

That is an issue :)

I was assuming that these were doing the same thing.
If they are doing different things then this is a blocking issue.

At first glance I already notice one potential bug:

```js
range.start.line >= start && (!end || range.start.line <= end)
```

This requires the disable range to fully enclose the code to fix.
A partial enclosure would still be fixed, right?

```css
/* stylelint disable */
.foo {
/* stylelint enable */
}
```

The `.foo {}` block ends after the disable range end. So a rule that would try to fix the range of the block would fix it, correct?

While the logic for the disable ranges for warnings would not emit a warning, correct?
I don't see why this **needs** to be different.

I feel really strongly that we must have only one implementation of how disable ranges work. Having two different ways makes it much harder to use and document Stylelint.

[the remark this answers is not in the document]
Mouvedia: > do not have half of it in `report`

6f1751c

This is much cleaner _and_ it preserves the blame.
Thanks !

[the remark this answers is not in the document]
Mouvedia: > I don't see why this needs to be different.

Ill check tonight and add a test if we are missing one.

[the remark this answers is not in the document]
Mouvedia: > I don't see why this needs to be different.

I have extracted the shared logic into `isDisabled` function.
I agree that it should be identical.

> A partial enclosure would still be fixed, right?

Now that the logic is shared, if there's bug it can be fixed for both warn and fix.
Let's have a separate issue to not clutter this PR too much?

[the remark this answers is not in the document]
romainmenke: How I would refactor it :

```js
import { RULE_NAME_ALL } from '../constants.mjs';

/**
 * Report a problem.
 *
 * This function accounts for `disabledRanges` attached to the result.
 * That is, if the reported problem is within a disabledRange,
 * it is ignored. Otherwise, it is attached to the result as a
 * postcss warning.
 *
 * It also accounts for the rule's severity.
 *
 * You *must* pass *either* a node or a line number.
 *
 * @type {import('stylelint').Utils['report']}
 */
export default function report(problem) {
	if (handleReportFixes(problem)) return;

	handleReportWarnings(problem);
}

/** @typedef {import('stylelint').Problem} Problem */

/**
 * Handle warnings from report
 *
 * @param {Problem} problem
 * @returns {boolean}
 */
function handleReportWarnings(problem) {
	const { ruleName, result, message, messageArgs, node, index, endIndex, word, severity } =
		problem;

	const ruleSeverityOption = severity || result.stylelint.ruleSeverities[ruleName];

	const defaultSeverity = result.stylelint.config?.defaultSeverity || 'error';

	const ruleSeverity =
		typeof ruleSeverityOption === 'function'
			? ruleSeverityOption(...(messageArgs || [])) || defaultSeverity
			: ruleSeverityOption;

	if (result.stylelint.quiet && ruleSeverity !== 'error') {
		return false;
	}

	const startLine = getReportStartLine(problem);

	if (!startLine) {
		throw new Error(
			`The "${ruleName}" rule failed to pass either a node or a line number to the \`report()\` function.`,
		);
	}

	const { ignoreDisables } = result.stylelint.config || {};
	const { disabledRanges } = result.stylelint;

	if (isDisabled(ruleName, startLine, disabledRanges)) {
		// Collect disabled warnings
		// Used to report `needlessDisables` in subsequent processing.
		const disabledWarnings =
			result.stylelint.disabledWarnings || (result.stylelint.disabledWarnings = []);

		disabledWarnings.push({
			rule: ruleName,
			line: startLine,
		});

		if (!ignoreDisables) {
			return false;
		}
	}

	if (!result.stylelint.stylelintError && ruleSeverity === 'error') {
		result.stylelint.stylelintError = true;
	}

	if (!result.stylelint.stylelintWarning && ruleSeverity === 'warning') {
		result.stylelint.stylelintWarning = true;
	}

	/** @type {import('stylelint').WarningOptions} */
	const warningProperties = {
		severity: ruleSeverity,
		rule: ruleName,
	};

	if (node) {
		warningProperties.node = node;
	}

	if (problem.start) {
		warningProperties.start = problem.start;
	} else if (typeof index === 'number') {
		warningProperties.index = index;
	}

	if (problem.end) {
		warningProperties.end = problem.end;
	} else if (typeof endIndex === 'number') {
		warningProperties.endIndex = endIndex;
	}

	if (word) {
		warningProperties.word = word;
	}

	const { customMessages } = result.stylelint;
	const warningMessage = buildWarningMessage(
		(customMessages && customMessages[ruleName]) || message,
		messageArgs,
	);

	result.warn(warningMessage, warningProperties);

	return true;
}

/**
 * Handle warnings from report
 *
 * @param {Problem} problem
 * @returns {boolean}
 */
function handleReportFixes(problem) {
	const { ruleName, result, fix } = problem;

	if (!fix) return false;

	const config = result.stylelint.config || {};
	const fixerRange = getFixerRange(problem);

	const startLine = getReportStartLine(problem);

	if (!startLine) return false;

	const { disabledRanges } = result.stylelint;
	const fixersData = result.stylelint.fixersData?.[ruleName] || [];

	if (
		config.rules?.[ruleName]?.[1]?.disableFix ||
		isDisabled(ruleName, startLine, disabledRanges)
	) {
		fixersData.push({ range: fixerRange, fixed: false });

		return false;
	}

	fixersData.push({ range: fixerRange, fixed: true });

	fix();

	return true;
}

/**
 * @param {import('stylelint').RuleMessage} message
 * @param {import('stylelint').Problem['messageArgs']} messageArgs
 * @returns {string}
 */
function buildWarningMessage(message, messageArgs) {
	const args = messageArgs || [];

	if (typeof message === 'string') {
		return printfLike(message, ...args);
	}

	return message(...args);
}

/**
 * @param {string} format
 * @param {Array<unknown>} args
 * @returns {string}
 */
function printfLike(format, ...args) {
	return args.reduce((/** @type {string} */ result, arg) => {
		return result.replace(/%[ds]/, String(arg));
	}, format);
}

/** @typedef {import('stylelint').DisabledRangeObject} DisabledRangeObject */
/** @typedef {import('stylelint').Range} Range */

/**
 * @param {string} ruleName
 * @param {number} startLine
 * @param {DisabledRangeObject | undefined} disabledRanges
 * @returns {boolean}
 */
function isDisabled(ruleName, startLine, disabledRanges) {
	if (!disabledRanges) return false;

	const ranges = disabledRanges[ruleName] ?? disabledRanges[RULE_NAME_ALL] ?? [];

	for (const range of ranges) {
		if (
			// If the problem is within a disabledRange,
			// and that disabledRange's rules include this one,
			// do not register a warning
			range.start <= startLine &&
			(range.end === undefined || range.end >= startLine) &&
			(!range.rules || range.rules.includes(ruleName))
		) {
			return true;
		}
	}

	return false;
}

/**
 * @param {Problem} problem
 * @returns {Range}
 */
function getFixerRange(problem) {
	const { node, index, endIndex, start, end, ruleName  } = problem;

	if (start?.line && start?.column && end?.line && end?.column) {
		return {
			start,
			end,
		};
	}

	if (node && typeof index === 'number' && typeof endIndex === 'number') {
		return node.rangeBy({ index, endIndex });
	}

	if (typeof index === 'number' || typeof endIndex === 'number' || node) {
		throw new Error(
			`The "${ruleName}" rule failed to pass both a \`node\`, \`index\` and \`endIndex\` to the \`report()\` function.`,
		);
	}

	throw new Error(
		`The "${ruleName}" rule failed to pass both \`start\` and \`end\` to the \`report()\` function.`,
	);
}

/**
 * @param {Problem} problem
 * @returns {number | undefined}
 */
function getReportStartLine(problem) {
	const { node, index, endIndex, line } = problem;

	// endIndex is optional
	const { start } = (node && node.rangeBy({ index, endIndex })) || {};

	return line || (start && start.line);
}
```

----

I agree that fully integrating it into a single function likely isn't ideal.

This proposed versions shares as much as possible without interleaving logic for fixes and warnings.

lib/utils/report.mjs
```diff
@@ -146,3 +146,79 @@ function printfLike(format, ...args) {
...
+function applyFix({ fix, result, ruleName, range }) {
+	const {
+		disabledRanges,
+		disabledRanges: { all = [] },
+		config,
+		fixersData,
+	} = result.stylelint;
+	const fn = isDisabled.bind(null, ruleName, range.start.line);
+	const ranges = disabledRanges[ruleName] ?? all;
+	const mayFix = config?.ignoreDisables || !ranges.length || !ranges.some(fn);
```
romainmenke, line 218: I think that no fixes should be applied in disable ranges even when `ignoreDisables` is set.

see : https://stylelint.io/user-guide/options#ignoredisables

> You can use this option to see what your linting results would be like without those exceptions.

From this I read that the intention of this option is to **preview** warnings and errors.

This does not hint at permanent effects, like fixes being applied.

lib/utils/report.mjs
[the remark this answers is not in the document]
Mouvedia: You are going for the full-on refactor.
My goal is to have the less diff possible.
i.e. preserve the blame, avoid regressions

Ill check it out tonight and try to mix the two.

lib/utils/report.mjs
Mouvedia: The `--help` says `Ignore "stylelint-disable" comments.`.
The name is "ignore disables".
I think it's pretty clear what it ought to do.
I would expect `--ignore-disables` to actually ignore disables when is conjunction with `--fix`.

I propose to update the documentation instead and add details about `--fix`/`fix`.

romainmenke: @ybiquitous any thoughts on this?

Either option has merit but we can't easily change this later in a patch/minor.

lib/utils/report.mjs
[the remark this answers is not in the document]
romainmenke: > My goal is to have the less diff possible.
i.e. preserve the blame, avoid regressions

A small diff should not be the goal and if you look closely then you will find that the diff isn't as large as you might expect.

Anyway, this was a very soft suggestion.
I don't mind submitting this patch in a subsequent PR.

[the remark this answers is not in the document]
Mouvedia: I have included your version of `isDisabled` in 1ae8523.
For the rest, I went for a more conservative approach.

> Anyway, this was a very soft suggestion.
> I don't mind submitting this patch in a subsequent PR.

Let's do this.
Thanks for your patience.
Ill put you as a co-author on the squash commit obviously.

lib/utils/report.mjs
```diff
@@ -146,3 +134,84 @@ function printfLike(format, ...args) {
...
+
+		addFixData({ fixersData, ruleName, range, fixed });
+
+		if (fixed) {
+			fix();
+
+			return true;
+		}
+	} else if (hasRangeData) {
+		addFixData({ fixersData, ruleName, range, fixed: false });
```
Mouvedia, line 200: note

Without passing `unfixable` to `report` nor having `rule.meta.fixer` we can't distinguish unmigrated rules that did fix from migrated rules that, for some cases, are unfixable.
i.e. it might be picked up by unmigrated rules and set `fixed` to false (which may be inaccurate)

@romainmenke should we add back `unfixable?: boolean` to `Problem`?
Your draft PR didn't have it and I can't handle that case without an extra boolean.

lib/utils/report.mjs
Mouvedia: I think that makes perfect sense to the point that I added a test for it.
Now since that's the last contentious point for this PR to be merged, I can remove it if you are feeling strongly about it.
But Ill probably open an issue afterward to request that it can be supported or we will have at least have to mention that it doesn't work as expected with `--fix`/`fix` in the documentation.

ybiquitous: From my feelings, `--ignore-disables` can be valid even with `--fix`.

> From this I read that the intention of this option is to preview warnings and errors.

I think users can use `--ignore-disables` *also* for "the intention", but it's up to users. Some people may expect `--ignore-disables` to work with `--fix`.

`--ignore-disables` is a quite strong option, so I think users should be careful to use it. 😅

ybiquitous (commented):
[ask] It's necessary to encourage plugin or custom rule authors to use the new way for autofix after this PR. So, I think we should update the following documents etc. somehow. Do you have any plans about that?

- https://github.com/stylelint/stylelint/blob/9a2e6540f92509628df76032262293ae57d32dda/docs/developer-guide/plugins.md#L234
- https://github.com/stylelint/stylelint/blob/9a2e6540f92509628df76032262293ae57d32dda/docs/developer-guide/rules.md?plain=1#L215

ybiquitous:
Ah, I'm sorry to miss the following description. It's fine. 👍🏼 

> Documentation will be added in a separate PR.

.changeset/friendly-snakes-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: `fix` to `Problem` type
```
ybiquitous, line 5: [suggest] This changelog message (also the PR title) seems internal and unhelpful to me.
Instead, what about focusing on the public `report()` API? For example:

```suggestion
Added: `fix`/`unfixable` arguments to `stylelint.utils.report()`
```

types/stylelint/index.d.ts
```diff
@@ -616,6 +634,7 @@ declare namespace stylelint {
 		 * Optional severity override for the problem.
 		 */
 		severity?: RuleSeverity;
+		unfixable?: boolean;
```
ybiquitous, line 637: [ask] To make sure, what case do you assume using `unfixable`?

types/stylelint/index.d.ts
```diff
@@ -577,6 +594,7 @@ declare namespace stylelint {
 	 * A lint problem.
 	 */
 	export type Problem = {
+		fix?: () => void | never;
```
ybiquitous, line 597: [suggest] Can you move `fix?` to the end of this type's definition to clarify that it's added later?

lib/utils/report.mjs
romainmenke: I would rather we don't have `unfixable` or `rule.meta.fixer`.
I am not convinced either are needed.

This change is also perfectly fine without either.

----

I think we can have sufficient forwards pressure by making it easy to migrate and clearly documenting the benefits.

We also have ways to deprecated the old fixers without introducing new API.
We can add warnings when rules access `context.fix` by making it a getter. This is only used by unmigrated rules.

lib/utils/report.mjs
romainmenke: 🤔 Users also have things like git history to undo unexpected changes.

So fine with me to go with the current implementation :)

types/stylelint/index.d.ts
```diff
@@ -157,6 +157,7 @@ declare namespace stylelint {
 		ruleSeverities: { [ruleName: string]: RuleSeverity };
 		customMessages: { [ruleName: string]: RuleMessage };
 		ruleMetadata: { [ruleName: string]: Partial<RuleMeta> };
+		fixersData: { [ruleName: string]: Array<FixerData> };
```
ybiquitous, line 160: [ask] Where is `fixersData` used? If autofix is performed only in `report()`, `fixersData` seems no longer needed.

lib/utils/report.mjs
ybiquitous: If we don't need `fixable`, it's better. 👍🏼 

It's worth considering `context.fix`.

Mouvedia: OK I have an idea.
For the very few rules that do have `unfixable` cases we will call `addFixdata` directly in the rule itself.
`addFixdata` doesn't have to be exported for now, but it will become an util eventually for that purpose.

.changeset/friendly-snakes-sparkle.md
Mouvedia:
```suggestion
Added: `fix` callback to the object passed to `stylelint.utils.report()`
```

types/stylelint/index.d.ts
Mouvedia: explained in the comments, but I found a better alternative
in short some fixable migrated rules have cases where they can't fix (so you won't have a fix callback)
these cannot be distinguished from unmigrated fixable rules when adding fix data

types/stylelint/index.d.ts
Mouvedia: I won't ask you to read all our exchanges with romain but this comment might be relevant : https://github.com/stylelint/stylelint/pull/7730#discussion_r1632050103

ybiquitous: I understand your thoughts. We can go on anyway and return later if needed since `fixersData` is just internal.

types/stylelint/index.d.ts
```diff
@@ -251,6 +252,22 @@ declare namespace stylelint {
...
+	export type Range = {
+		start: {
+			line: number;
+			column: number /** inclusive */;
+		};
+		end: {
+			line: number;
+			column: number /** exclusive */;
+		};
+	};
```
ybiquitous, line 264: [suggest] We already have the same data structure for line and column here:

https://github.com/stylelint/stylelint/blob/06e93eecfb0bea0823a1544d873232a12b7d7fa6/types/stylelint/index.d.ts#L617-L629

So, it seems to make sense to add a new data type like PostCSS's [`Position`](https://postcss.org/api/#position). What do you think?

```ts
type Position = {
  line: number;
  column: number;
};
```

types/stylelint/index.d.ts
```diff
@@ -251,6 +252,22 @@ declare namespace stylelint {
...
 	};
 
+	export type Range = {
+		start: {
+			line: number;
+			column: number /** inclusive */;
+		};
+		end: {
+			line: number;
+			column: number /** exclusive */;
```
ybiquitous, line 262: [ask] What "inclusive" and "exclusive" in the comments mean?

lib/utils/report.mjs
```diff
@@ -146,3 +134,83 @@ function printfLike(format, ...args) {
...
+/** @typedef {import('stylelint').Problem} Problem */
+
+/**
+ * @param {string} ruleName
+ * @param {number} startLine
+ * @param {DisabledRangeObject} [disabledRanges]
+ */
+function isDisabled(ruleName, startLine, disabledRanges) {
+	if (!disabledRanges) return false;
+
```
ybiquitous, line 150: [suggest] `disabledRanges` seems not undefined: 🤔 

```suggestion
 * @param {DisabledRangeObject} disabledRanges
 */
function isDisabled(ruleName, startLine, disabledRanges) {
```

types/stylelint/index.d.ts
Mouvedia: I didn't add a common type because I wanted to document the inclusive/exclusive difference inline.
I guess we could add `Position`, yes.

types/stylelint/index.d.ts
Mouvedia: It works like `column` and `endColumn` in our tests.
i.e. `end.column` is the index of the `\0`

lib/utils/report.mjs
```diff
@@ -44,37 +46,23 @@ export default function report(problem) {
...
-					return;
-				}
-
-				break;
-			}
+	if (isDisabled(ruleName, startLine, disabledRanges)) {
+		// Collect disabled warnings
+		// Used to report `needlessDisables` in subsequent processing.
+		const disabledWarnings =
+			result.stylelint.disabledWarnings || (result.stylelint.disabledWarnings = []);
```
ybiquitous, line 55: [note] It seems reasonable that the initialization of `disabledWarnings` is performed like other `PostcssResult.stylelint` properties (see #7728). After this PR is merged, let's consider if there is any chance.

lib/utils/report.mjs
```diff
@@ -146,3 +134,83 @@ function printfLike(format, ...args) {
...
+		}
+	}
+
+	return false;
+}
+
+/** @param {Problem} problem */
+function isFixApplied(problem) {
+	const { end, start, index, endIndex, fix, node, result, ruleName } = problem;
+	const hasRange = end?.column && start;
```
ybiquitous, line 171: [suggest] Small refactorings:

- The order of start and end seems more natural.
- Checking if the variable is a number seems safer. (since `0` is falsy)

```suggestion
	const hasRange = isNumber(start) && isNumber(end?.column);
```

lib/utils/report.mjs
Mouvedia: I won't remove `if (!disabledRanges) return false;` just to keep it consistent with the previous code.
We can refactor it later.

lib/utils/report.mjs
```diff
@@ -146,3 +134,83 @@ function printfLike(format, ...args) {
...
+		) {
+			return true;
+		}
+	}
+
+	return false;
+}
+
+/** @param {Problem} problem */
+function isFixApplied(problem) {
```
ybiquitous, line 169: [suggest] Not required, but I recommend explicitly having a return type. It's clear and prevents unintentional bugs (e.g. returning undefined).

```suggestion
/**
 * @param {Problem} problem
 * @returns {boolean}
 */
function isFixApplied(problem) {
```

lib/utils/report.mjs
```diff
@@ -146,3 +134,83 @@ function printfLike(format, ...args) {
...
+	const { end, start, index, endIndex, fix, node, result, ruleName } = problem;
+	const hasRange = end?.column && start;
+	const hasIndexes = isNumber(index) && isNumber(endIndex);
+	const hasRangeData = hasRange || hasIndexes;
+	const hasFixCallback = typeof fix === 'function';
```
ybiquitous, line 174: [suggest] How about using early return like this to reduce a nest level?

```js
if (typeof fix !== 'function') return false;
// or
if (!isFunction(fix)) return false;

if (!hasRangeData) {
// ...
```

lib/utils/report.mjs
Mouvedia: I think that's not necessary if the function follows the convention `hasFoo` or `isFoo`.
Returning `undefined` would be falsy so it wouldn't be a big deal.

types/stylelint/index.d.ts
ybiquitous: So, let's add `Position`. If it's added, the `Problem` and `Range` types should be as follows:

```ts
type Problem = {
  // ...
  start?: Position;
  end?: Position;
  // ...
};

type Range = {
  start: Position;
  end: Position;
};
```

lib/utils/report.mjs
ybiquitous: Make sense.

lib/utils/report.mjs
ybiquitous: > The first is an `object`.

Sorry, I misunderstood. So, I suggest instead:

```js
const hasRange = Boolean(start) && isNumber(end?.column);
```

ybiquitous: Ah, `Boolean(start)` emits type errors. Okay, how about this?

```js
const hasRange = start && isNumber(end?.column);
```

types/stylelint/index.d.ts
```diff
@@ -251,6 +252,22 @@ declare namespace stylelint {
 		fixable?: boolean;
 	};
 
+	export type Range = {
```
ybiquitous, line 255: [suggest] How about marking it as an unpublished type? Keeping published types minimal can make refactoring efficient later.

```suggestion
	/** @internal */
	export type Range = {
```

lib/utils/report.mjs
Mouvedia: Yes please, postpone it.

lib/utils/report.mjs
```diff
@@ -146,3 +134,80 @@ function printfLike(format, ...args) {
...
+}
+
+/** @param {Problem} problem */
+function isFixApplied(problem) {
+	const { end, start, index, endIndex, fix, node, result, ruleName } = problem;
+	const hasRange = start && isNumber(end?.column);
+	const hasIndexes = isNumber(index) && isNumber(endIndex);
+	const hasRangeData = hasRange || hasIndexes;
+
+	if (typeof fix !== 'function') return false;
```
Mouvedia, line 175: wip
minor

`addFixdata` for unmigrated rules (report path) and migrated rules (unfixable path)

lib/utils/report.mjs
Mouvedia: Indeed, it's not necessary.

Mouvedia:
If there's no more feedback, I will merge it tonight GMT+2.

lib/utils/report.mjs
```diff
@@ -146,3 +134,84 @@ function printfLike(format, ...args) {
...
+	// even though stylelint-disable comments cannot be inserted inside a declaration or a selector list,
+	// new lines cannot be disregarded because Range is exposed through StylelintPostcssResult['fixersData']
+	// i.e. ranges must be accurate to be exploited
+	if (!hasRangeData) {
+		throw new Error(
+			`The fix callback for rule "${ruleName}" requires either index/endIndex or start/end to be passed to the \`report()\` function.`,
+		);
+	}
+
+	const hasFix = config.fix && !config.rules?.[ruleName][1]?.disableFix;
```
ybiquitous, line 192: [suggest] `config.rules?.[ruleName]` may not be an array, and `config.rules?.[ruleName][1]` may cause a type error. So, we have to check it more strictly. For example:

```js
const hasFix = config.fix && !isFixDisabled(config.rules?.[ruleName]);

/**
 * @param {unknown} ruleSettings
 * @returns {boolean}
 */
function isFixDisabled(ruleSettings) {
	if (Array.isArray(ruleSettings)) {
		const [, secondaryOptions] = ruleSettings;

		if (isPlainObject(secondaryOptions)) {
			return Boolean(secondaryOptions.disableFix);
		}
	}

	return false;
}
```

ybiquitous: Ah, wait. We may already normalize `config.rules` before calling `report()`.

https://github.com/stylelint/stylelint/blob/ba2aca55b6b7f1e94b025d91fc81166071a059f3/lib/normalizeAllRuleSettings.mjs#L19-L23

ybiquitous: Maybe safe. But we should improve the `config.rules` type for accuracy. Let's refactor it later. Sorry for bothering you.

ybiquitous (approved):
Thanks for your patience. Great step. LGTM 👍🏼

Mouvedia:
Thanks for all the feedback.
I have added the tasks left on #2643.

PR #7733, opened 2024-06-02
remove global fix prevention

romainmenke:

None, this is a demo

No, it's self-explanatory.

### DIFF
--- a/lib/lintPostcssResult.mjs
+++ b/lib/lintPostcssResult.mjs
@@ -41,10 +41,6 @@ export default async function lintPostcssResult(stylelintOptions, postcssResult,
 
 	const isFileFixCompatible = isFixCompatible(postcssResult);
 
-	if (!isFileFixCompatible) {
-		postcssResult.stylelint.disableWritingFix = true;
-	}
-
 	const postcssRoots = /** @type {import('postcss').Root[]} */ (
 		postcssDoc && postcssDoc.constructor.name === 'Document' ? postcssDoc.nodes : [postcssDoc]
 	);
--- a/lib/standalone.mjs
+++ b/lib/standalone.mjs
@@ -144,12 +144,11 @@ export default async function standalone({
 		const autofix = fix ?? config?.fix ?? false;
 
 		if (autofix && postcssResult && !postcssResult.stylelint.ignored) {
-			returnValue.code =
-				!postcssResult.stylelint.disableWritingFix && postcssResult.opts
-					? // If we're fixing, the output should be the fixed code
-					  postcssResult.root.toString(postcssResult.opts.syntax)
-					: // If the writing of the fix is disabled, the input code is returned as-is
-					  code;
+			returnValue.code = postcssResult.opts
+				? // If we're fixing, the output should be the fixed code
+				  postcssResult.root.toString(postcssResult.opts.syntax)
+				: // If the writing of the fix is disabled, the input code is returned as-is
+				  code;
 			returnValue._output = returnValue.code; // TODO: Deprecated. Remove in the next major version.
 		}
 
@@ -229,13 +228,7 @@ export default async function standalone({
 				/**
 				 * If we're fixing, save the file with changed code
 				 */
-				if (
-					postcssResult.root &&
-					postcssResult.opts &&
-					!postcssResult.stylelint.ignored &&
-					fix &&
-					!postcssResult.stylelint.disableWritingFix
-				) {
+				if (postcssResult.root && postcssResult.opts && !postcssResult.stylelint.ignored && fix) {
 					const fixedCss = postcssResult.root.toString(postcssResult.opts.syntax);
 
 					if (
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -163,7 +163,6 @@ declare namespace stylelint {
 		ignored?: boolean;
 		stylelintError?: boolean;
 		stylelintWarning?: boolean;
-		disableWritingFix?: boolean;
 		config?: Config;
 	};
 
--- a/lib/lintPostcssResult.cjs
+++ b/lib/lintPostcssResult.cjs
@@ -44,10 +44,6 @@ async function lintPostcssResult(stylelintOptions, postcssResult, config) {
 
 	const isFileFixCompatible = isFixCompatible(postcssResult);
 
-	if (!isFileFixCompatible) {
-		postcssResult.stylelint.disableWritingFix = true;
-	}
-
 	const postcssRoots = /** @type {import('postcss').Root[]} */ (
 		postcssDoc && postcssDoc.constructor.name === 'Document' ? postcssDoc.nodes : [postcssDoc]
 	);
--- a/lib/standalone.cjs
+++ b/lib/standalone.cjs
@@ -146,12 +146,11 @@ async function standalone({
 		const autofix = fix ?? config?.fix ?? false;
 
 		if (autofix && postcssResult && !postcssResult.stylelint.ignored) {
-			returnValue.code =
-				!postcssResult.stylelint.disableWritingFix && postcssResult.opts
-					? // If we're fixing, the output should be the fixed code
-					  postcssResult.root.toString(postcssResult.opts.syntax)
-					: // If the writing of the fix is disabled, the input code is returned as-is
-					  code;
+			returnValue.code = postcssResult.opts
+				? // If we're fixing, the output should be the fixed code
+				  postcssResult.root.toString(postcssResult.opts.syntax)
+				: // If the writing of the fix is disabled, the input code is returned as-is
+				  code;
 			returnValue._output = returnValue.code; // TODO: Deprecated. Remove in the next major version.
 		}
 
@@ -231,13 +230,7 @@ async function standalone({
 				/**
 				 * If we're fixing, save the file with changed code
 				 */
-				if (
-					postcssResult.root &&
-					postcssResult.opts &&
-					!postcssResult.stylelint.ignored &&
-					fix &&
-					!postcssResult.stylelint.disableWritingFix
-				) {
+				if (postcssResult.root && postcssResult.opts && !postcssResult.stylelint.ignored && fix) {
 					const fixedCss = postcssResult.root.toString(postcssResult.opts.syntax);
 
 					if (


lib/__tests__/standalone-fix.test.mjs
```diff
@@ -402,3 +402,67 @@ it('returns the original code if the fix option is true but the code is not fixa
...
+		},
+		fix: true,
+		formatter: 'string',
+	});
+
+	expect(result.code).toBe(stripIndent`
+		/* stylelint-disable plugin/fixes-always */
+		.bar {}`);
+	expect(result.report).toBe('');
+});
```
romainmenke, line 468: Notice how rule specific disable comments also do not work today.

`postcssResult.stylelint.disableWritingFix` does not prevent writes in this scenario.

Mouvedia: Indeed the order of the booleans* should have been different to cater for the plugins not respecting the convention of using `context.fix` block for fixes.

*l114-118 lintPostcssResult

lib/__tests__/standalone-fix.test.mjs
```diff
@@ -402,3 +402,26 @@ it('returns the original code if the fix option is true but the code is not fixa
...
+				'plugin/fixes': true,
+			},
+		},
+		fix: false,
+		formatter: 'string',
+	});
+
+	expect(result.code).toBeUndefined();
+	expect(result.report).toBe('');
+});
```
romainmenke, line 427: I don't want to add explicit test coverage for rules that do not respect `context.fix` when the `fix` option is `true` and disable comments are present.

My thinking is that that behavior is outside of the defined boundaries of Stylelint and the outcome is undefined. Adding explicit test coverage might give the wrong impression.

However adding extra test coverage for when the `fix` option is false seemed like a good idea.

romainmenke:
I am unsure if the test coverage should be addressed.

No code branches were added, only deleted.
So I assumed coverage would go up, not down.

I don't want to get sidetracked adding coverage for pre-existing code :)

lib/lintPostcssResult.mjs
Mouvedia, line 1: nit
suggestion

l110–115

```js
						fix:
							!disableFix &&
							config.fix &&
							// #4573
							isFileFixCompatible &&
							// #4705
							!postcssResult.stylelint.disabledRanges[ruleName],
```

`isFixCompatible` function already has a lengthy description.

lib/standalone.mjs
```diff
@@ -144,12 +144,11 @@ export default async function standalone({
 		const autofix = fix ?? config?.fix ?? false;
 
 		if (autofix && postcssResult && !postcssResult.stylelint.ignored) {
-			returnValue.code =
-				!postcssResult.stylelint.disableWritingFix && postcssResult.opts
-					? // If we're fixing, the output should be the fixed code
-					  postcssResult.root.toString(postcssResult.opts.syntax)
-					: // If the writing of the fix is disabled, the input code is returned as-is
-					  code;
+			returnValue.code = postcssResult.opts
```
Mouvedia, line 147: question
minor

Does reverting to https://github.com/stylelint/stylelint/pull/4573/files#diff-37813c24c3fa3c6093015b61419f0bc18983cc2bbaddd56a732576bdb53f9023L168-L169 ends up with a test failure?
It should.

Could you add a comment that details in which case(s) `postcssResult.opts` is falsy?
e.g.
```js
// this may be falsy depending on rules' disableFix secondary option
// and/or the presence of stylelint commands
```

ybiquitous:
> It is almost completely an internal refactor. But it might be noticeable to some users in (hopefully) very rare cases.
> 
> I prefer to include it in the change log so that users can more easily discover what changed.

Sounds good to me. I just left a comment on that, too: https://github.com/stylelint/stylelint/issues/7735#issuecomment-2146344716

lib/__tests__/standalone-fix.test.mjs
ybiquitous: This added test case looks good to me. 👍🏼 

Additionally, especially in this case, adding a case when "the `fix` option is `true` and disable comments are present" makes sense to me, because such a case can clearly show what this PR changed and why this bug has been left for a long time.

As you commented, such a new case may confuse us in the future, but leaving a comment and a link to this issue/PR could mitigate it.

If such a case is no longer needed in the future, then we can remove it.

ybiquitous:
> What is an appropriate and clear message for this purpose?

I'm not confident so much, but what about this?

```
Fixed: autofix by incorrectly implemented plugins when `/* stylelint-disable */` comments are specified
```

Feel free to give me any suggestions. 😃

Mouvedia:
> `/* stylelint-disable */` comments

I would replace that part with _unscoped styelint commands_ so that it covers `-next-line` and `-line`.

> autofix by

_autofix of_

Personally I think a changeset is not needed, but it doesn't hurt if it's accurate.

ybiquitous:
@Mouvedia Thanks, your suggestion sounds good to me. 👍🏼

lib/__tests__/standalone-fix.test.mjs
romainmenke: I've added such a case, please let me know if this covers what you had in mind

ybiquitous: Sounds good. Thank you. 👍🏼

lib/__tests__/standalone-fix.test.mjs
```diff
@@ -402,3 +402,73 @@ it('returns the original code if the fix option is true but the code is not fixa
...
+
+it('returns partially fixed code if the fix option is true, depending on the plugin implementation', async () => {
+	// see: https://github.com/stylelint/stylelint/issues/7735
```
ybiquitous, line 430: [suggest] In addition, how about a TODO comment to help us? e.g.

```suggestion
	// TODO: This case should fail if the bug in #2643 is fixed. Then we can remove this.
	// see: https://github.com/stylelint/stylelint/issues/2643
	// see: https://github.com/stylelint/stylelint/issues/7735
```

lib/standalone.mjs
romainmenke: > Does reverting to https://github.com/stylelint/stylelint/pull/4573/files#diff-37813c24c3fa3c6093015b61419f0bc18983cc2bbaddd56a732576bdb53f9023L168-L169 cause a test failure?
It should.

It didn't cause a test failure before because I was hesitant to add test coverage for undefined behavior. Any rule can do unexpected things.

An extra test was added following https://github.com/stylelint/stylelint/pull/7733#discussion_r1625195087

This test fails when reverting the code change.

```
  ● returns partially fixed code if the fix option is true, depending on the plugin implementation

    expect(received).toBe(expected) // Object.is equality

    - Expected  - 2
    + Received  + 2

      /* stylelint-disable */
      a {
    -   always: fixed;
    +   always: 1;
        when-context: 2;
      }
      /* stylelint-enable */

      /* stylelint-disable plugin/fixes */
      b {
    -   always: fixed;
    +   always: 1;
        when-context: 2;
      }
      /* stylelint-enable plugin/fixes */

      457 |     });
      458 |
    > 459 |     expect(result.code).toBe(stripIndent`
          |                         ^
      460 |             /* stylelint-disable */
      461 |             a {
      462 |                     always: fixed;

      at Object.<anonymous> (lib/__tests__/standalone-fix.test.mjs:459:22)


```

romainmenke: > Could you add a comment that details in which cases `postcssResult.opts` may be falsy?
e.g.

That seems unrelated to this change?

ybiquitous: > Does reverting to [#4573 (files)](https://github.com/stylelint/stylelint/pull/4573/files#diff-37813c24c3fa3c6093015b61419f0bc18983cc2bbaddd56a732576bdb53f9023L168-L169) cause a test failure?
> It should.

If my understanding is correct, thanks to `isFixCompatible`, it doesn't cause a test failure, right?

https://github.com/stylelint/stylelint/blob/1c974ecfe8a65d71ce49ae4a8463b81facd80361/lib/lintPostcssResult.mjs#L130-L132

ybiquitous: Oh, sorry. My comment is duplicated. 🙇🏼 

---

>> Could you add a comment that details in which cases `postcssResult.opts` may be falsy?
>> e.g.
>
> That seems unrelated to this change?

I don't believe it's needed to add such a comment, too.

It seems to be caused by `EmptyResult.opts`.

https://github.com/stylelint/stylelint/blob/1c974ecfe8a65d71ce49ae4a8463b81facd80361/types/stylelint/index.d.ts#L37-L49

romainmenke: Running the added tests with the code as exists on main does cause a test failure.
As the last test that was added specifically tests the edge case that is affected.

romainmenke:
I've added the changeset with a slight tweak to make it less verbose.

lib/standalone.mjs
ybiquitous: > Running the added tests with the code as exists on main does cause a test failure.

Yes, I also confirmed it. 👍🏼

ybiquitous (approved):
Thanks a lot for making this change. LGTM 👍🏼

lib/__tests__/fixtures/plugin-fixes.mjs
```diff
@@ -0,0 +1,38 @@
...
+
+const ruleName = 'plugin/fixes';
+
+const messages = ruleMessages(ruleName, {});
+
+const rule = (primary, _, context) => {
+	return (root, result) => {
+		const validOptions = validateOptions(result, ruleName, {
+			actual: primary,
+			possible: [true],
```
Mouvedia, line 16: off-topic

We have 3 `possible: [true, false],` in the codebase that need to be changed to `possible: [true, null],`.

types/stylelint/index.d.ts
```diff
@@ -163,7 +163,6 @@ declare namespace stylelint {
 		ignored?: boolean;
 		stylelintError?: boolean;
 		stylelintWarning?: boolean;
-		disableWritingFix?: boolean;
```
Mouvedia, line 166: note

was introduced alongside l39 and l41

Mouvedia (approved):

romainmenke:
Thank you for the reviews and feedback 🙇

PR #7744, opened 2024-06-09
Fix `comment-whitespace-inside` end positions

Mouvedia:

#5694

There are at least 2 more rules that don't have either `endIndex` nor `word` passed to `report`.
This one is slightly more finicky compared to the other 2 so I wanted this rule to be ready before its upcoming fix callback refactor.

tl;dr: a quantifier of the regular expressions has been changed


lib/rules/comment-whitespace-inside/index.mjs
```diff
@@ -65,44 +65,53 @@ const rule = (primary, _secondaryOptions, context) => {
...
 				throw new Error(`Invalid comment: "${rawComment}"`);
 			}
 
 			const opener = leftMatches[1];
-			const leftSpace = leftMatches[2] || '';
-			const rightSpace = rightMatches[1] || '';
+			const leftWord = leftMatches[2] || '';
+			const leftSpace = leftWord[0] || '';
+			const rightWord = rightMatches[1] || '';
+			const rightSpace = rightWord[0] || '';
```
ybiquitous, line 84: [suggest] This is not required, but the following would be a bit more readable for me:

- Use `??` instead of `||`.
- Use `charAt()`, which always returns a string.

```suggestion
			const leftWord = leftMatches[2] ?? '';
			const leftSpace = leftWord.charAt(0);
			const rightWord = rightMatches[1] ?? '';
			const rightSpace = rightWord.charAt(0);
```

ybiquitous (approved):
Thank you, nice catch! 👍🏼 

I've left one comment, but you can merge this as-is.

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #7746, opened 2024-06-10
`keyframe-selector-notation` fix callback refactor

Mouvedia:

#2643

I wanted to start with a rule that uses `word`.
After that PR most rule refactors will be quite straight forward.

### DIFF
--- a/.changeset/mighty-chefs-rhyme.md
+++ b/.changeset/mighty-chefs-rhyme.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Changed: use `fix` callback instead of `context.fix` for `keyframe-selector-notation`
--- a/.changeset/mighty-chefs-rhyme.md
+++ b/.changeset/mighty-chefs-rhyme.md
@@ -1,5 +1,0 @@
----
-"stylelint": patch
----
-
-Changed: use `fix` callback instead of `context.fix` for `keyframe-selector-notation`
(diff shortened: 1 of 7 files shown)


lib/rules/keyframe-selector-notation/index.mjs
```diff
@@ -107,20 +103,18 @@ const rule = (primary, _, context) => {
...
+					const index = tag.sourceIndex;
+					const endIndex = index + selector.length;
 
 					report({
 						message: messages.expected,
 						messageArgs: [selector, fixedSelector],
 						node: keyframeRule,
 						result,
 						ruleName,
-						word: selector,
```
Mouvedia, line 123: Should we support `word` in `isFixApplied`?
i.e. instead of requiring `index`/`endIndex`

.changeset/mighty-chefs-rhyme.md
ybiquitous, line 1: [ask] This PR is refactoring, so the changelog item isn't necessary, right?

ybiquitous (commented):
Thanks. LGTM except for changelog 👍🏼

.changeset/mighty-chefs-rhyme.md
Mouvedia: It can be interpreted both ways. I will remove it.

Mouvedia:
@ybiquitous codecov is complaining about l93–94. I can't merge.
Ill let you handle it.

ybiquitous (approved):
The reason is that there were no approvals. Just approved ✅

PR #7747, opened 2024-06-10
Fix `function-name-case` end positions and refactor to use `fix` callback instead of `context.fix`

Mouvedia:

#2643, #5694

N/A

### DIFF
--- a/.changeset/grumpy-shoes-cross.md
+++ b/.changeset/grumpy-shoes-cross.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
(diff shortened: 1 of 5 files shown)


.changeset/grumpy-shoes-cross.md
ybiquitous, line 1: [suggest] Let's revert this changelog since it's just refactoring.

ybiquitous (commented):
Overall LGTM. Can you address my review?

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #7752, opened 2024-06-12
Fix `selector-type-case` end positions and refactor to use `fix` callback instead of `context.fix`

Mouvedia:

#2643, #5694

N/A

### DIFF
--- a/.changeset/real-insects-flash.md
+++ b/.changeset/real-insects-flash.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-type-case` end positions
--- a/lib/rules/selector-type-case/__tests__/index.mjs
+++ b/lib/rules/selector-type-case/__tests__/index.mjs
@@ -112,6 +112,10 @@ testRule({
 			code: 'DIV::before {}',
 			fixed: 'div::before {}',
 			message: messages.expected('DIV', 'div'),
+			line: 1,
+			column: 1,
+			endLine: 1,
+			endColumn: 4,
 		},
 		{
 			code: 'a B {}',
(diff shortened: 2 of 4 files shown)


ybiquitous (approved):
Thanks, LGTM 👍🏼

Mouvedia:
@ybiquitous what's the objective concerning the release which will have the documentation update?
i.e. should the refactors be scattered between several releases ?

I can't really do all the refactors myself.
If we were to split the work between the maintainers we could probably aim for the next release.
That's roughly 8 rules per person if I only count you and @romainmenke.

ybiquitous:
@Mouvedia I recommend creating a new umbrella issue to list all the rules to be refactored. This issue would be helpful for contributors other than the three of us.

Also, I think it's not a problem to release the refactors gradually.

Mouvedia:
> @Mouvedia I recommend creating a new umbrella issue to list all the rules to be refactored. This issue would be helpful for contributors other than the three of us.

Do you want me to remove the task list from #2643?
 
> Also, I think it's not a problem to release the refactors gradually.

👍

ybiquitous:
> Do you want me to remove the task list from #2643?

Ah, sorry. #2643 is enough. Let's use the issue. 👍🏼 

It might be nice to add the "good first issue" label to attract the attention of many contributors.

Mouvedia:
> It might be nice to add the "good first issue" label to attract the attention of many contributors.

Ill do that once we have done one clean PR; to use as a reference.
Right now I concentrate on tricky ones or more challenging ones for that very reasons.
i.e. I leave the easy ones last so that they can be picked-up by newcomers

ybiquitous:
Sounds good. It's up to you. 👍🏼

PR #7754, opened 2024-06-12
Refactor `comment-empty-line-before` to use `fix` callback instead of `context.fix`

Mouvedia:

#2643

1.
I didn't bother refactoring this line

https://github.com/stylelint/stylelint/blob/131b84f49676578a677d64b3277f66976654d6a9/lib/utils/checkAgainstRule.mjs#L15

because it's unrelated.

2.
no need for a changeset, it already used the `node` for the range


types/stylelint/index.d.ts
```diff
@@ -235,7 +235,7 @@ declare namespace stylelint {
 	export type RuleContext = {
 		configurationComment?: string | undefined;
 		fix?: boolean | undefined;
-		newline?: string | undefined;
```
Mouvedia, line 238: Is

https://github.com/stylelint/stylelint/blob/131b84f49676578a677d64b3277f66976654d6a9/lib/lintPostcssResult.mjs#L30

always truthy?
If so shouldn't

https://github.com/stylelint/stylelint/blob/131b84f49676578a677d64b3277f66976654d6a9/lib/lintPostcssResult.mjs#L37

shouldn't always be a string?

cf https://github.com/stylelint/stylelint/pull/5418/files#diff-9a4b261e295bef1ccce813aa194ac4023a1dcf98236667889964b1ba6fbc551fR144

lib/rules/comment-empty-line-before/index.mjs
```diff
@@ -100,18 +100,29 @@ const rule = (primary, secondaryOptions, context) => {
...
+			const endIndex = comment.toString().length;
+			const range = comment.rangeBy({ index: 0, endIndex });
+
 			// Fix
-			if (context.fix) {
-				if (typeof context.newline !== 'string') return;
+			const fix = () => {
+				let count = -1;
+
+				assert(typeof context.newline === 'string');
```
Mouvedia, line 110: Is

https://github.com/stylelint/stylelint/blob/131b84f49676578a677d64b3277f66976654d6a9/lib/lintPostcssResult.mjs#L30

always truthy?
If so shouldn't

https://github.com/stylelint/stylelint/blob/131b84f49676578a677d64b3277f66976654d6a9/lib/lintPostcssResult.mjs#L37

shouldn't always be a string?

cf
https://github.com/stylelint/stylelint/pull/5418/files#diff-9a4b261e295bef1ccce813aa194ac4023a1dcf98236667889964b1ba6fbc551fR144
https://github.com/stylelint/stylelint/commit/7c7ee09d775597ccfba0d15a749681c41e8522ae#diff-488b06a075b40dca0232e14e2583371556ff9895854619c74e4d85112e24d38bR121-R130

ybiquitous: Let's refactor `postcssDoc` and `context.newline` with another PR!

Mouvedia: I am not confident in refactoring that one. Maybe the types are wrong or it was different in a previous version?
Yeah let's postpone this to another PR.

ybiquitous (approved):
Thanks. LGTM 👍🏼

PR #7755, opened 2024-06-12
Fix `selector-not-notation` end positions and refactor to use `fix` callback instead of `context.fix`

Mouvedia:

#2643

N/A

### DIFF
--- a/lib/__tests__/standalone-fix.test.mjs
+++ b/lib/__tests__/standalone-fix.test.mjs
@@ -48,7 +48,6 @@ it('fixes when enabled in config', async () => {
 });
 
 describe('fix callback', () => {
-	// eslint-disable-next-line jest/no-disabled-tests
 	suite('core rules', () => {
 		test('out of range', async () => {
 			const code = stripIndent`
--- a/lib/__tests__/standalone-fix.test.mjs
+++ b/lib/__tests__/standalone-fix.test.mjs
@@ -49,7 +49,7 @@ it('fixes when enabled in config', async () => {
 
 describe('fix callback', () => {
 	// eslint-disable-next-line jest/no-disabled-tests
-	suite.skip('core rules', () => {
+	suite('core rules', () => {
 		test('out of range', async () => {
 			const code = stripIndent`
 				:not(
--- a/lib/rules/selector-not-notation/index.cjs
+++ b/lib/rules/selector-not-notation/index.cjs
@@ -98,7 +98,7 @@ const rule = (primary) => {
 
 					if (!hasConsecutiveNot) return;
 
-					fix = fixComplex.bind({ ruleNode, selectorRoot }, prev);
+					fix = () => fixComplex(prev, ruleNode, selectorRoot);
 				} else {
 					const selectors = pseudo.nodes;
 
@@ -108,7 +108,7 @@ const rule = (primary) => {
 					const isFixable =
 						second && (!second.nodes.length || selectors.every(({ nodes }) => nodes.length === 1));
 
-					if (isFixable) fix = fixSimple.bind({ ruleNode, selectorRoot }, pseudo);
+					if (isFixable) fix = () => fixSimple(pseudo, ruleNode, selectorRoot);
 				}
 
 				report({
@@ -143,9 +143,10 @@ const getLastConsecutiveNot = ({ parent, sourceIndex }) => {
 
 /**
  * @param {Pseudo} not
- * @this {{ ruleNode: Rule, selectorRoot: Root }}
+ * @param {Rule} ruleNode
+ * @param {Root} selectorRoot
  */
-function fixSimple(not) {
+function fixSimple(not, ruleNode, selectorRoot) {
 	const simpleSelectors = not.nodes
 		.filter(({ nodes }) => nodes[0] && isSimpleSelector(nodes[0]))
 		.map((s) => {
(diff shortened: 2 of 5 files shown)


lib/rules/selector-not-notation/index.cjs
```diff
@@ -83,46 +85,109 @@ const rule = (primary, _, context) => {
...
+					.filter(({ nodes }) => nodes[0] && isSimpleSelector(nodes[0]))
+					.map((s) => {
+						validateTypes.assert(s.nodes[0]);
+						s.nodes[0].rawSpaceBefore = '';
+						s.nodes[0].rawSpaceAfter = '';
+
+						return s;
+					});
+				const firstSelector = simpleSelectors.shift();
+				let offset = 0;
```
Mouvedia, line 100: note

This is the only notable change that I have done to the existing fixers.

lib/rules/selector-not-notation/index.mjs
```diff
@@ -79,46 +81,109 @@ const rule = (primary, _, context) => {
...
+					.filter(({ nodes }) => nodes[0] && isSimpleSelector(nodes[0]))
+					.map((s) => {
+						assert(s.nodes[0]);
+						s.nodes[0].rawSpaceBefore = '';
+						s.nodes[0].rawSpaceAfter = '';
+
+						return s;
+					});
+				const firstSelector = simpleSelectors.shift();
+				let offset = 0;
```
Mouvedia, line 96: note

This is the only notable change that I have done to the existing fixers.

lib/__tests__/standalone-fix.test.mjs
```diff
@@ -49,7 +49,7 @@ it('fixes when enabled in config', async () => {
 
 describe('fix callback', () => {
 	// eslint-disable-next-line jest/no-disabled-tests
```
ybiquitous, line 51: [suggest] Let's remove the unused comment. It will be automatically detected by ESLint, thanks to #7757.

```suggestion
```

lib/rules/selector-not-notation/index.mjs
```diff
@@ -79,46 +81,109 @@ const rule = (primary, _, context) => {
 
 			if (!selectorRoot) return;
 
+			/** @param {Pseudo} not */
+			const fixSimple = (not) => {
```
ybiquitous, line 85: [suggest] Can we keep the original position of `fixSimple()` to minimize diff?

Mouvedia: No we can't unless we pass tons of arguments to the function.
i.e. it needs to be in the right scope

see also https://github.com/stylelint/stylelint/pull/7755#discussion_r1637024503

Mouvedia: Lemme try something.

Mouvedia: done

lib/rules/selector-not-notation/index.mjs
```diff
@@ -82,43 +85,38 @@ const rule = (primary, _, context) => {
...
+				let fix;
+
 				if (primary === 'complex') {
 					const prev = pseudo.prev();
 					const hasConsecutiveNot = prev && isNot(prev);
 
 					if (!hasConsecutiveNot) return;
 
-					if (context.fix) return fixComplex(prev);
+					fix = fixComplex.bind({ ruleNode, selectorRoot }, prev);
```
ybiquitous, line 97: [suggest] Using `bind` and `this` is a bit too tricky. Instead, how about just creating a function like this?

```js
fix = () => {
	fixComplex(prev);
	ruleNode.selector = selectorRoot.toString();
	return getNewRange(prev, ruleNode);
};
```

lib/rules/selector-not-notation/index.mjs
```diff
@@ -82,43 +85,38 @@ const rule = (primary, _, context) => {
...
-						context.fix &&
-						selectors.length > 1 &&
-						selectors[1] &&
-						(selectors[1].nodes.length === 0 || selectors.every(({ nodes }) => nodes.length === 1));
+					const second = selectors.length > 1 && selectors[1];
+					const isFixable =
+						second && (!second.nodes.length || selectors.every(({ nodes }) => nodes.length === 1));
 
-					if (mustFix) return fixSimple(pseudo);
+					if (isFixable) fix = fixSimple.bind({ ruleNode, selectorRoot }, pseudo);
```
ybiquitous, line 107: [suggest] Same here. https://github.com/stylelint/stylelint/pull/7755/files#r1637409423

```js
if (isFixable) {
	fix = () => {
		const offset = fixSimple(pseudo);
		ruleNode.selector = selectorRoot.toString();
		return getNewRange(pseudo, ruleNode, offset);
	};
}
```

Mouvedia: I am not convinced that's better.
What about not using `this`?
i.e. adding 2 new parameters instead
e.g. `fixSimple.bind(null, pseudo, ruleNode, selectorRoot);`

ybiquitous: Instead of `bind`, can we just do this? ⬇️ 

```js
fix = () => fixSimple(pseudo, ruleNode, selectorRoot)
```

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #7756, opened 2024-06-12
Refactor `comment-whitespace-inside` to use `fix` callback instead of `context.fix`

Mouvedia:

#2643

It used to fix both sides for `"always"` regardless of the current `complain`.
i.e. in certain cases it was attempting 2 more fixes for nothing

### DIFF
--- a/lib/rules/comment-whitespace-inside/index.mjs
+++ b/lib/rules/comment-whitespace-inside/index.mjs
@@ -41,7 +41,7 @@ function addWhitespaceAfter(comment) {
 }
 
 /** @type {import('stylelint').Rule} */
-const rule = (primary, _secondaryOptions, context) => {
+const rule = (primary) => {
 	return (root, result) => {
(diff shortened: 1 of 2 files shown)


ybiquitous (approved):
LGTM 👍🏼

lib/rules/comment-whitespace-inside/index.mjs
```diff
@@ -112,23 +112,22 @@ const rule = (primary, _secondaryOptions, context) => {
...
 					}
 
-					return;
-				}
+					const mutate =
+						message === messages.expectedClosing ? addWhitespaceAfter : addWhitespaceBefore;
+
+					mutate(comment);
+
+					return comment.rangeBy({ index, endIndex: index + 1 });
```
Mouvedia, line 129: I think this is incorrect.
i.e. it should be `endIndex: index`

PR #7757, opened 2024-06-13
Refactor to turn on `reportUnusedDisableDirectives` for ESLint

ybiquitous:
This `reportUnusedDisableDirectives` option reports unused `// eslint-disable` comments.

For example, assume the following JS file:

```js
const a = 1; // eslint-disable-line no-unused-vars

a;
```

Run ESLint with this option on or off:

```sh-session
# reportUnusedDisableDirectives: false
$ npx eslint tmp/eslint-test.js
(success)

# reportUnusedDisableDirectives: true
$ npx eslint tmp/eslint-test.js

/Users/masafumi.koba/git/stylelint/stylelint/tmp/eslint-test.js
  1:14  warning  Unused eslint-disable directive (no problems were reported from 'no-unused-vars')
```

None.

No, it's self-explanatory.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
       "testRule": "readonly",
       "testRuleConfigs": "readonly"
     },
+    "reportUnusedDisableDirectives": true,
     "root": true
   },
   "remarkConfig": {


package.json
```diff
@@ -113,6 +113,7 @@
       "testRule": "readonly",
       "testRuleConfigs": "readonly"
     },
+    "reportUnusedDisableDirectives": true,
```
Mouvedia, line 116: > [For legacy compatibility, true is equivalent to "warn" and false is equivalent to "off".](https://eslint.org/docs/latest/use/configure/configuration-files#reporting-unused-disable-directives)

So if we ever use a flat config for ESLint it will have to be changed to `"error"`.

Mouvedia (approved):

PR #7758, opened 2024-06-13
Fix `declaration-block-no-duplicate-properties` reported ranges and refactor to use `fix` callback instead of `context.fix`

Mouvedia:

#2643

Previously the rule always reported the last duplicate which wasn't the one being removed—if fixed—in the majority of cases.
An example of a case in which the rule did report correctly the duplicate that would be removed, was if the duplicate that remained was `!important` and placed before the last duplicate.
i.e. [demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIAOACYAdAdugLgdwEsocALeFARiQA8UBCAgWyQgCccBDTHAbi10LEylGn0wBfEABoQAMwIAbGADkOjOIhjU1SJQDowAZ0PTwETPIDmCEBkwoU6EFpwxMUQ0-IBtLA4dOhjgAnkoKBNwAtJAWBJaRQVxQHKxQTn4oALpYkjIxVgBibIwcODYAVobmprBIJoh2ASBBoTDh3F6OIAqlMEFOUhmBIWERONHmVgmc7ilpCF09rv0gOSDiQA)

Now it always reports the duplicate that will be removed.
It also helps for the diff between the range of the warning and the range returned by `fix`.


Mouvedia:
@ybiquitous previously I could re-run individual tests that failed due to `cacheStrategy is "metadata"`.
Now the retry button is missing from the interface, did something change with the action?

ybiquitous:
@Mouvedia Sorry for the inconvenience. We (owners) are reconsidering and assessing the Contributors team's permissions for several reasons. I just granted you permission to write since you have been so active recently. Please let me know if you are still having any trouble. 🙏🏼

Mouvedia:
> We (owners) are reconsidering and assessing the Contributors team's permissions for several reasons.

Is one of the reason the fallout of the xz backdoor? Did contributors had too many rights?

ybiquitous:
Right, it's one reason.

ybiquitous:
> Now it always reports the duplicate that will be removed.

Sounds good. Can you add a changelog item for this? I think it's a bug fix.

Mouvedia:
> I think it's a bug fix.

I consider it a preference or laziness on our part.

> Can you add a changelog item for this?

Ill add a changeset, sure.

lib/rules/declaration-block-no-duplicate-properties/index.mjs
```diff
@@ -246,17 +250,20 @@ const rule = (primary, secondaryOptions, context) => {
...
 					}
 
-					if (context.fix) {
-						return duplicateDecl.remove();
-					}
+					const fix = () => {
+						duplicateDecl.remove();
+
+						return duplicateDecl.rangeBy({ index: 0, endIndex: 0 });
+					};
```
Mouvedia, line 257: refactor
minor

We could have only one function declaration.
i.e.
```js
const fix = fixer.bind(null, node);
// …
const fix = fixer.bind(null, duplicateDecl);
```

.changeset/green-lions-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Changed: `declaration-block-no-duplicate-properties` warning ranges are now always matching the duplicate that will be removed after autofixing
```
ybiquitous, line 5: [suggest] "Changed" has often been used for significant changes, so I'd like to use "Fixed" here. So, how about this?

```suggestion
Fixed: `declaration-block-no-duplicate-properties` reported code ranges
```

Mouvedia:
```suggestion
Fixed: `declaration-block-no-duplicate-properties` reported ranges
```

lib/rules/declaration-block-no-duplicate-properties/index.mjs
```diff
@@ -200,26 +201,31 @@ const rule = (primary, secondaryOptions, context) => {
 				const unprefixedDuplicatesAreEqual =
 					vendor.unprefixed(value) === vendor.unprefixed(duplicateValue);
 
+				/** @param {Declaration} node */
+				const fixer = (node) => {
+					node.remove();
+
+					return node.rangeBy({ index: 0, endIndex: 0 });
+				};
```
ybiquitous, line 209: [suggest] How about simplifying `fixer()` like this?

- Avoid `bind`
- Move outside the `eachDeclarationBlock` loop

```js
/** @param {Declaration} node */
const fixer = (node) => () => {
	node.remove();

	return node.rangeBy({ index: 0, endIndex: 0 });
};
```

PR #7760, opened 2024-06-14
Fix `value-keyword-case` end positions and refactor to use `fix` callback instead of `context.fix`

Mouvedia:

#5694, #2643

N/A


ybiquitous (approved):
Thank you. LGTM 👍🏼

PR #7762, opened 2024-06-15
Refactor `*-empty-line-before` rules to use `fix` callback instead of `context.fix`

Mouvedia:

#2643

1.
This supersedes #7754 because the description of the `addEmptyLineBefore` function had me overlooking an edge case.

2.
This introduces an util because all of these rules behave the same regarding fixes.

3.
The range restriction is being removed because you can get an accurate range without an `index`/`endIndex` or `word`; it's replaced by a more lax node restriction.
i.e. for `declaration-empty-line-before` passing `endIndex` would require to use `fast-deep-equal` to find out whether the last declaration has a semicolon or not
cf 4d4556a31c49690bd8865c7779a6ff7a9186f169

4.
I recommend `?file-filters%5B%5D=.mjs&show-viewed-files=true` for your review.


lib/rules/declaration-empty-line-before/index.mjs
```diff
@@ -127,22 +128,27 @@ const rule = (primary, secondaryOptions, context) => {
...
-					removeEmptyLinesBefore(decl, context.newline);
-				}
-
-				return;
-			}
-
 			const message = expectEmptyLineBefore ? messages.expected : messages.rejected;
+			const mutate = expectEmptyLineBefore ? addEmptyLineBefore : removeEmptyLinesBefore;
+			const semicolon = decl.parent?.raws.semicolon ? 1 : 0;
+			const endIndex = decl.toString().length + semicolon;
```
Mouvedia, line 134: This is the only rule that breaks the pattern of passing the length directly.

see also https://github.com/stylelint/stylelint/pull/7762/files#diff-fa27f0406ab7dfe7dfe29dfd44816f632ce1ccd274bc13ccde805c27e6ec92fbR52-R58

Mouvedia, line 134: major

this is an incorrect assumption
i.e. it fails for `a { bottom: 12px; top: 17px }`

package.json
```diff
@@ -169,6 +169,7 @@
     "css-functions-list": "^3.2.2",
     "css-tree": "^2.3.1",
     "debug": "^4.3.5",
+    "fast-deep-equal": "^3.1.3",
```
Mouvedia, line 172: It's already a dependency of `eslint`.

lib/utils/emptyLineBeforeFixer.mjs
Mouvedia, line 1: Maybe rename `fixEmptyLineBefore`?

lib/__tests__/standalone-fix.test.mjs
```diff
@@ -205,7 +205,7 @@ describe('fix callback', () => {
 					},
 				}),
 			).rejects.toThrow(
-				'The fix callback for rule "plugin/selector-no-qux" requires index/endIndex or start/end or word to be passed to the `report()` function.',
+				'The fix callback for rule "plugin/selector-no-qux" requires a node to be passed to the `report()` function.',
 			);
 		});
```
Mouvedia, line 210: @romainmenke this change will allow the refactor you wanted between fix and report to happen.
Tell me, if you want to work on it after this PR.
Ill eventually do a small scale version of that refactor if you don't.

lib/utils/report.mjs
ybiquitous, line 1: [suggest] Can you add test cases to `lib/utils/__tests__/report.test.mjs` for this new behavior?

lib/utils/fixEmptyLineBefore.mjs
```diff
@@ -0,0 +1,39 @@
...
+ */
+export default ({ node, newline, mutate, range }) => {
+	let count;
+
+	switch (mutate.name) {
+		case 'addEmptyLineBefore':
+			count = !/\r?\n/.test(node.raws.before) ? -2 : -1;
+			break;
+		case 'removeEmptyLinesBefore':
+			count = (node.raws.before.match(/\n/g) || []).length - 1;
```
ybiquitous, line 24: [ask] It doesn't seem that this `removeEmptyLinesBefore` case is executed. Will this be used later?

lib/utils/fixEmptyLineBefore.mjs
```diff
@@ -0,0 +1,39 @@
...
+
+/** @typedef {import('stylelint').Range} Range */
+/** @typedef {import('postcss').Node} Node */
+/** @typedef {(node: Node, newline: string) => Node} Callback */
+
+/**
+ * fix callback for *-empty-line-before rules
+ * @param {object} o
+ * @param {Node} o.node
+ * @param {string=} o.newline
```
ybiquitous, line 11: [ask] Typo?

```suggestion
 * @param {string} o.newline
```

lib/utils/fixEmptyLineBefore.mjs
```diff
@@ -0,0 +1,39 @@
...
+ * @param {Callback} o.mutate
+ * @param {Range} [o.range]
+ * @returns {Range}
+ */
+export default ({ node, newline, mutate, range }) => {
```
ybiquitous, line 16: [suggest] Let's specify a function name for consistency and debuggability:

```suggestion
export default function fixEmptyLineBefore({ node, newline, mutate, range }) {
```

lib/rules/comment-empty-line-before/index.mjs
```diff
@@ -1,5 +1,6 @@
-import { assert, isRegExp, isString } from '../../utils/validateTypes.mjs';
+import { isRegExp, isString } from '../../utils/validateTypes.mjs';
 import addEmptyLineBefore from '../../utils/addEmptyLineBefore.mjs';
+import fixer from '../../utils/fixEmptyLineBefore.mjs';
```
ybiquitous, line 3: [suggest] Let's use the actual function name for readability.

```suggestion
import fixEmptyLineBefore from '../../utils/fixEmptyLineBefore.mjs';
```

ybiquitous (changes requested):

lib/utils/report.mjs
Mouvedia: Either me (or Romain) is gonna refactor `report` in the coming weeks and the `throw` already has a test.
So I don't think it's necessary in this PR.

see https://github.com/stylelint/stylelint/pull/7762/files#diff-a5338b3127d4362334c745fc36f7eb15c3b029c550b326b6a4c9ac8f578c85b7R208

lib/utils/fixEmptyLineBefore.mjs
Mouvedia: No it's optional.
There are two ways supported in JSDOC.
Since a refactor—that we talked about—is planned which will change `newline` to be never `undefined`, it won't be missed.

lib/utils/fixEmptyLineBefore.mjs
Mouvedia: It will if `expectEmptyLineBefore` is false in the `walk*` loop of the rules.

lib/utils/fixEmptyLineBefore.mjs
ybiquitous: Then, I prefer `string | undefined` because it's explicit. I believe it's more readable for many people.

ybiquitous: But it's fine for now since it will be refactored.

lib/rules/at-rule-empty-line-before/index.mjs
```diff
@@ -138,20 +139,23 @@ const rule = (primary, secondaryOptions, context) => {
...
 			const message = expectEmptyLineBefore ? messages.expected : messages.rejected;
+			const mutate = expectEmptyLineBefore ? addEmptyLineBefore : removeEmptyLinesBefore;
 
-			report({ message, node: atRule, result, ruleName });
+			// Fix
+			const fix = fixEmptyLineBefore.bind(null, {
+				node: atRule,
+				newline: context.newline,
+				mutate,
+			});
```
ybiquitous, line 150: [suggest] Using arrow functions instead of `.bind()` seems more readable for me:

```suggestion
			const fix = () => fixEmptyLineBefore({
				node: atRule,
				newline: context.newline,
				mutate,
			});
```

lib/rules/declaration-empty-line-before/__tests__/index.mjs
```diff
@@ -45,6 +45,37 @@ testRule({
 			message: messages.expected,
 			line: 1,
 			column: 5,
+			endLine: 1,
+			endColumn: 15,
+		},
+		{
+			code: 'a { top: 15px }',
+			fixed: 'a {\n\n top: 15px }',
```
ybiquitous, line 53: [ask] Doesn't this case duplicate with the case at L43 (`a { top: 15px; }`)?

lib/utils/fixEmptyLineBefore.mjs
```diff
@@ -0,0 +1,39 @@
...
+ * @param {Node} o.node
+ * @param {string=} o.newline
+ * @param {Callback} o.mutate
+ * @param {Range} [o.range]
+ * @returns {Range}
+ */
+export default function fixEmptyLineBefore({ node, newline, mutate, range }) {
+	let count;
+
+	switch (mutate.name) {
```
ybiquitous, line 19: [suggest] I want to avoid depending on a function name because it's not type-safe. So, I suggest more refactoring as follows:

- Introduce the `action` argument (type: `'add' | 'remove'`) instead of `mutate`.
- Internally call `addEmptyLineBefore()` or `removeEmptyLinesBefore()`.
- If possible, inline `addEmptyLineBefore.mjs` and `removeEmptyLinesBefore.mjs` into this file.

E.g.

```js
switch (action) {
	case 'add':
		count = addEmptyLineBefore(node, newline);
		break;
	case 'remove':
		count = removeEmptyLinesBefore(node, newline);
		break;
	default:
		throw new Error(`Unknown action '${action}'`);
}
```

lib/utils/fixEmptyLineBefore.mjs
```diff
@@ -0,0 +1,39 @@
...
+ * fix callback for *-empty-line-before rules
+ * @param {object} o
+ * @param {Node} o.node
+ * @param {string=} o.newline
+ * @param {Callback} o.mutate
+ * @param {Range} [o.range]
+ * @returns {Range}
+ */
+export default function fixEmptyLineBefore({ node, newline, mutate, range }) {
+	let count;
```
ybiquitous, line 17: [suggest] Can you rename the variable for clarity?

```suggestion
	let lineCount;
```

lib/utils/fixEmptyLineBefore.mjs
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * fix callback for *-empty-line-before rules
+ * @param {object} o
+ * @param {Node} o.node
+ * @param {string=} o.newline
+ * @param {Callback} o.mutate
+ * @param {Range} [o.range]
+ * @returns {Range}
+ */
+export default function fixEmptyLineBefore({ node, newline, mutate, range }) {
```
ybiquitous, line 16: [suggest] It seems more accurate to use plural (`line` or `lines`). Any thoughts?

```suggestion
export default function fixEmptyLinesBefore({ node, newline, mutate, range }) {
```

lib/utils/fixEmptyLineBefore.mjs
ybiquitous: I forgot one point. I'd like to move the logic of `/\r?\n/.test` or `match(/\n/g)` into each add/remove function because of duplication.

lib/rules/declaration-empty-line-before/__tests__/index.mjs
Mouvedia: No it doesn't, there's a semicolon missing.
It matters for regression detection.
e.g. if someone eventually passes an `endIndex`
See 6cfca7d8.

lib/utils/fixEmptyLineBefore.mjs
Mouvedia: > If possible, inline addEmptyLineBefore.mjs and removeEmptyLinesBefore.mjs into this file.

I won't do that because these functions have tests.

> I'd like to move the logic of /\r?\n/.test or match(/\n/g) into each add/remove function because of duplication.

This can be done in a different refactor PR.
I would like this PR be about the rule refactor and not the utils refactors.

ybiquitous: It's also fine to address these refactoring on another PR 👍🏼

lib/rules/at-rule-empty-line-before/index.mjs
Mouvedia: I don't like how prettier format it but I don't mind the change.

lib/utils/report.mjs
```diff
@@ -179,15 +176,14 @@ function isFixApplied(problem) {
 		return false;
 	}
 
-	// even though stylelint-disable comments cannot be inserted inside a declaration or a selector list,
-	// a complete range is requested as a prerequisite
-	if (!hasRangeData) {
+	if (!node) {
 		throw new Error(
-			`The fix callback for rule "${ruleName}" requires index/endIndex or start/end or word to be passed to the \`report()\` function.`,
+			`The fix callback for rule "${ruleName}" requires a node to be passed to the \`report()\` function.`,
```
ybiquitous, line 181: [ask] I just noticed `node` has been required already (see the `report()` type definition). So, if checking `node`, it seems to make sense to do that at the start in `report()`, e.g.

```js
export default function report(problem) {
	const { ruleName, result, message, messageArgs, line, node, index, endIndex, word, severity } =
		problem;

	if (!node) {
		throw new Error(...);
	}
```

Any thoughts?

Mouvedia: This is related to https://github.com/stylelint/stylelint/pull/7762#discussion_r1641343910.
i.e. this is postponed to another refactor PR

lib/rules/at-rule-empty-line-before/index.mjs
```diff
@@ -138,20 +137,24 @@ const rule = (primary, secondaryOptions, context) => {
...
-				}
-
-				return;
-			}
-
 			const message = expectEmptyLineBefore ? messages.expected : messages.rejected;
+			const action = expectEmptyLineBefore ? 'add' : 'remove';
 
-			report({ message, node: atRule, result, ruleName });
+			// Fix
```
ybiquitous, line 143: [suggest] Super minor. This comment no longer makes sense.

```suggestion
```

lib/rules/comment-empty-line-before/index.mjs
```diff
@@ -100,38 +99,22 @@ const rule = (primary, secondaryOptions, context) => {
 				return;
 			}
 
-			const endIndex = comment.toString().length;
-			const range = comment.rangeBy({ index: 0, endIndex });
+			const message = expectEmptyLineBefore ? messages.expected : messages.rejected;
+			const action = expectEmptyLineBefore ? 'add' : 'remove';
 
 			// Fix
```
ybiquitous, line 105: [suggest] Same here.

```suggestion
```

lib/rules/custom-property-empty-line-before/index.mjs
```diff
@@ -101,26 +100,23 @@ const rule = (primary, secondaryOptions, context) => {
...
-					removeEmptyLinesBefore(decl, context.newline);
-				}
-
-				return;
-			}
-
 			const message = expectEmptyLineBefore ? messages.expected : messages.rejected;
+			const action = expectEmptyLineBefore ? 'add' : 'remove';
+
+			// Fix
```
ybiquitous, line 106:
```suggestion
```

lib/rules/declaration-empty-line-before/index.mjs
```diff
@@ -127,22 +126,25 @@ const rule = (primary, secondaryOptions, context) => {
...
-				}
-
-				return;
-			}
-
 			const message = expectEmptyLineBefore ? messages.expected : messages.rejected;
+			const action = expectEmptyLineBefore ? 'add' : 'remove';
 
-			report({ message, node: decl, result, ruleName });
+			// Fix
```
ybiquitous, line 132:
```suggestion
```

lib/rules/rule-empty-line-before/index.mjs
```diff
@@ -115,30 +114,23 @@ const rule = (primary, secondaryOptions, context) => {
...
-					removeEmptyLinesBefore(ruleNode, newline);
-				}
-
-				return;
-			}
-
 			const message = expectEmptyLineBefore ? messages.expected : messages.rejected;
+			const action = expectEmptyLineBefore ? 'add' : 'remove';
+
+			// Fix
```
ybiquitous, line 120:
```suggestion
```

ybiquitous (approved):
Thanks. LGTM.

I've left minor comments, but you can ignore them. 👍🏼

lib/__tests__/standalone-fix.test.mjs
Mouvedia: #7769

PR #7764, opened 2024-06-15
Fix `property-no-unknown` false positives for `navigation` and legacy properties

### LINKED ISSUE #7751: `property-no-unknown` false positive for descriptor `navigation`

```css
@view-transition {
  navigation: auto;
}
```

```json
{
  "rules": {
    "property-no-unknown": true
  }
}
```

```shell
stylelint "**/*.css"
```

```json
{
    "stylelint": "^16.6.1",
    "stylelint-config-standard": "^36.0.0",
}
```

Lint should pass.

### What actually happened?

```
20:3  ✖  Unexpected unknown at-rule "@view-transition"  at-rule-no-unknown
21:5  ✖  Unexpected unknown property "navigation"       property-no-unknown
```

_No response_

Mouvedia:

Closes #7751

ref
https://www.w3.org/TR/css-view-transitions-2/#view-transition-navigation-descriptor
https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariCSSRef/Articles/StandardCSSProperties.html
https://stackoverflow.com/a/11759353/248058


.changeset/stupid-cheetahs-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fix: `property-no-unknown` false positives
```
ybiquitous, line 5: [suggest] I want to clarify this PR's target like this:

```suggestion
Fixed: `property-no-unknown` false positives for `navigation` and legacy properties
```

lib/reference/properties.mjs
```diff
@@ -618,3 +618,50 @@ export const longhandTimeProperties = new Set([
 ]);
 
 export const shorthandTimeProperties = new Set(['transition', 'animation']);
+
+export const legacyProperties = new Set([
```
ybiquitous, line 622: [ask] Is it possible to add them to `known-css-properties` upstream?

lib/rules/property-no-unknown/index.mjs
```diff
@@ -21,9 +22,16 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/property-no-unknown',
 };
 
+const knownProperties = [
+	...properties,
+	...legacyProperties,
+	// postcss doesn't distinguish between descriptors and properties
+	'navigation',
```
ybiquitous, line 29: [ask] Is it possible to add `navigation` to `known-css-properties` upstream?

lib/reference/properties.mjs
ybiquitous, line 622: [ask] Is there a reliable reference to make this list?

Mouvedia: It's possible but, for the same reasons that `css-functions-list` or `csstree` can be slow to release, we want to fix it on our side first.
Also, in the same way that `html-tags` doesn't include deprecated elements, we run the risk that their package policy would be against adding legacy properties.
i.e. we should open a PR there but we should still fix it in the meantime.

lib/rules/property-no-unknown/index.mjs
Mouvedia: If they consider descriptors not to be properties we run the risk that it would be rejected.

lib/reference/properties.mjs
Mouvedia: Yes, I have added two links in [OP](https://github.com/stylelint/stylelint/pull/7764#issue-2355325301).
Only `-apple-dashboard-region` is included by `known-css-property`.

lib/reference/properties.cjs
```diff
@@ -623,7 +623,55 @@ const longhandTimeProperties = new Set([
 
 const shorthandTimeProperties = new Set(['transition', 'animation']);
 
+const legacyProperties = new Set([
+	'-apple-line-clamp',
```
Mouvedia, line 627: missing `-apple-dashboard-region`

lib/reference/properties.mjs
Mouvedia: known-css/known-css-properties#174

ybiquitous: Okay, can you add the links to this variable doc? E.g.

```suggestion
// See ...
export const legacyProperties = new Set([
```

ybiquitous: Thanks for opening the PR upstream. It's okay to fix them on our side for now. 👍🏼 

In addition, can you add a TODO comment with the PR link so as not to forget to remove duplicate properties when `known-CSS-properties` releases a new version in the future?

Mouvedia: The issue will do.

Mouvedia:
```suggestion
/** @see known-css/known-css-properties#174 */
const legacyProperties = new Set([
```

Mouvedia:
```suggestion
/** @see known-css/known-css-properties#174 */
export const legacyProperties = new Set([
```

lib/rules/property-no-unknown/index.mjs
```diff
@@ -21,9 +22,16 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/property-no-unknown',
 };
 
+const knownProperties = [
+	...properties,
+	...legacyProperties,
```
Mouvedia, line 27:
```suggestion
	// https://stackoverflow.com/a/11759353/
	...legacyProperties,
```

.changeset/stupid-cheetahs-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `property-no-unknown` false positives for `navigation` and legacy properties
```
Mouvedia, line 5:
```suggestion
Fixed: `property-no-unknown` false positives for `navigation` and miscellaneous legacy properties
```

lib/reference/properties.mjs
Mouvedia: https://github.com/known-css/known-css-properties/pull/174#issuecomment-2176842525

ybiquitous: Opened #7771. Can you take a look? Thanks.

Mouvedia: I tried to commit and got:

```console
stylelint $ git push --no-verify
remote: Permission to stylelint/stylelint.git denied to Mouvedia.
fatal: impossible d'accéder à 'https://github.com/stylelint/stylelint.git/' : The requested URL returned error: 403
```

We had the same problem on #7713.

ybiquitous: Um, weird. You now have the *Write* role in this repo... 🤔

ybiquitous: Um..., I think you can push commits like this https://github.com/stylelint/stylelint/commit/f4e8b63cdcb097d3855f07dfe487bada104c50a6

Mouvedia: Really? I must be doing something wrong then.
I thought I needed the aforementioned fined-grained token to be accepted to push on `dependabot/*` branches.
I am not really in the mood to investigate right now; Ill let you do the `legacyProperties` removal.

ybiquitous: Ah, I see. It seems that Dependabot branches are especially protected. Alright, I'll push later. 👌🏼

Issue #6075, opened 2022-05-09
Fix imprecise start/end indices in rules using `node.selectors`


mattxwang:
### What is the problem you're trying to solve?

In the process of working on #5694, I've noticed a couple of edge-cases that could use more precise/improved warning indices.

For example, in `no-invalid-double-slash-comments` (related PR: #6072), I note:

> For `no-invalid-double-slash-comments`, there was one test case I was confused about:
> 
> ```js
> code: 'a, //div { color: pink; }',
> description: 'between rules',
> message: messages.rejected,
> line: 1,
> column: 1,
> endLine: 1,
> endColumn: 26,
> ```
> 
> I haven't changed the behaviour of this test case, but to me I'm not sure why it's `column: 1` - the comment starts at index `4` (where the double-slash is). I might be misunderstanding how this rule works, though!

And in `no-descending-specificity` (related PR: #6049), @ybiquitous notes:

> I'm concerned about the following edge-case:
> 
> ```js
> {
> 	code: 'a:hover {} a, b, a {}',
> 	warnings: [
> 		{
> 			message: messages.rejected('a', 'a:hover'),
> 			line: 1,
> 			column: 12,
> 			endLine: 1,
> 			endColumn: 13,
> 		},
> 		{
> 			message: messages.rejected('a', 'a:hover'),
> 			line: 1,
> 			column: 18,    // but actual: 12
> 			endLine: 1,
> 			endColumn: 19, // but actual: 13
> 		},
> 	],
> },
> ```

There might be many more of these edge-cases that report slightly-mismatched/erroneous indices for the offending warnings. More precise indices can be very helpful downstream, especially for the vscode-stylelint team.

### What solution would you like to see?

@ybiquitous has suggested to use `postcss-selector-parser` instead of `ruleNode.selectors` to provide us with exact positions for selectors instead of guessing. Depending on whether or not other issues involve other elements, we may have to use other postcss parsers as well.

Overall, I'm not very familiar with the internals, so any comments are helpful! If we agree this is a good idea, I can get some of this work done - at the very least, address the issue for the two rules mentioned above, and start working on these as I encounter more rules in #5694.

---

- [ ] `no-duplicate-selectors`
- [x] `no-invalid-double-slash-comments`
- [x] `no-descending-specificity`

ybiquitous:
I think there is the same problem in `no-duplicate-selectors`. See #6047.

ybiquitous:
[...]

PR #7768, opened 2024-06-17
Fix `no-invalid-double-slash-comments` reported ranges

Mouvedia:

#6075

N/A

### DIFF
--- a/.changeset/yellow-feet-provide.md
+++ b/.changeset/yellow-feet-provide.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `no-invalid-double-slash-comments` autofix
+Fixed: `no-invalid-double-slash-comments` reported ranges
--- a/.changeset/yellow-feet-provide.md
+++ b/.changeset/yellow-feet-provide.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `no-invalid-double-slash-comments` autofix
--- a/lib/rules/no-invalid-double-slash-comments/index.mjs
+++ b/lib/rules/no-invalid-double-slash-comments/index.mjs
@@ -28,22 +28,36 @@ const rule = (primary) => {
 					node: decl,
 					result,
 					ruleName,
-					word: decl.toString(),
 				});
 			}
 		});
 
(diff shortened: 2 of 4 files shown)


ybiquitous:
This PR will close #6075, right?

lib/rules/no-invalid-double-slash-comments/index.mjs
```diff
@@ -28,22 +28,36 @@ const rule = (primary) => {
...
+					const word = lines.length > 1 ? `${comment}\n` : comment;
+
 					report({
 						message: messages.rejected,
 						node: ruleNode,
 						result,
 						ruleName,
-						word: ruleNode.toString(),
+						index: i,
+						word,
```
ybiquitous, line 56: [ask] Does it make sense to specify `endIndex` instead of `word` because `index` is already set?

Mouvedia:
> This PR will close https://github.com/stylelint/stylelint/issues/6075, right?

It won't.

lib/rules/no-invalid-double-slash-comments/index.mjs
Mouvedia: It's more complicated than using `word`.
This will eventually be refactored again in the upcoming weeks so it's acceptable for now.

.changeset/yellow-feet-provide.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `no-invalid-double-slash-comments` autofix
```
ybiquitous, line 5: [suggest] This PR is not related to autofix, but end positions, right?

```suggestion
Fixed: `no-invalid-double-slash-comments` end positions
```

If so, please update the PR title.

Mouvedia:
```suggestion
Fixed: `no-invalid-double-slash-comments` reported ranges
```

ybiquitous (approved):
Thanks, LGTM 👍🏼

lib/rules/no-invalid-double-slash-comments/index.mjs
Mouvedia: My bad I was wrong. It will be addressed in an upcoming PR.

PR #7769, opened 2024-06-17
Simplify `report` util even further

Mouvedia:

https://github.com/stylelint/stylelint/pull/7762#discussion_r1642584242

N/A

### DIFF
--- a/lib/utils/report.mjs
+++ b/lib/utils/report.mjs
@@ -1,5 +1,5 @@
+import { isFunction as isFn, isNumber, isString } from '../utils/validateTypes.mjs';
 import { RULE_NAME_ALL } from '../constants.mjs';
-import { isFunction as isFn } from '../utils/validateTypes.mjs';
 
 /**
  * Report a problem.
@@ -77,13 +77,13 @@ export default function report(problem) {
 
 	if (start) {
 		warningProperties.start = start;
-	} else if (typeof index === 'number') {
+	} else if (isNumber(index)) {
 		warningProperties.index = index;
 	}
 
 	if (end) {
 		warningProperties.end = end;
-	} else if (typeof endIndex === 'number') {
+	} else if (isNumber(endIndex)) {
 		warningProperties.endIndex = endIndex;
 	}
 
@@ -111,7 +111,7 @@ export default function report(problem) {
  * @returns {string}
  */
 function buildWarningMessage(message, messageArgs) {
-	if (typeof message === 'string') {
+	if (isString(message)) {
 		return printfLike(message, ...messageArgs);
 	}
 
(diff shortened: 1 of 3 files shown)


lib/utils/report.mjs
```diff
@@ -170,28 +154,18 @@ function isDisabled(ruleName, startLine, disabledRanges) {
...
 
-/** @param {Problem} problem */
-function isFixApplied(problem) {
-	const { end, start, index, endIndex, fix, node, result, ruleName, word } = problem;
-	const { disabledRanges, config = {}, fixersData } = result.stylelint;
+/** @param {Problem & { line: number }} problem */
+function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
+	const { disabledRanges, config = {}, fixersData } = stylelint;
 
 	if (typeof fix !== 'function') {
```
ybiquitous, line 161: [suggest] How about replacing `typeof` here as well? Other `typeof` in this file can be addressed, too.

```suggestion
	if (!isFn(fix)) {
```

lib/utils/report.mjs
```diff
@@ -16,54 +16,45 @@ import { isNumber } from './validateTypes.mjs';
...
+	const ruleSeverity = isFn(severity) ? severity(...messageArgs) || defaultSeverity : severity;
 
 	// In quiet mode, mere warnings are ignored
-	if (result.stylelint.quiet && ruleSeverity !== 'error') {
-		return;
-	}
-
-	// endIndex is optional
-	const { start } = (node && node.rangeBy({ index, endIndex })) || {};
+	if (quiet && ruleSeverity !== 'error') return;
```
ybiquitous, line 32: [suggest] How about extracting the severity literals to constants (`'error'` and `'warning'`)? Put the constants in `lib/constants.mjs` seems good.

```suggestion
	if (quiet && ruleSeverity !== SEVERITY_ERROR) return;
```

Mouvedia: There are too matches in the code base. Id prefer if it was done in a separate refactor PR. 
i.e. it should be done for all files at once

ybiquitous: I'm fine with another PR 👍🏼

ybiquitous (approved):
Now much simpler. LGTM 👍🏼

lib/utils/report.mjs
```diff
@@ -16,54 +16,45 @@ import { isNumber } from './validateTypes.mjs';
...
 
+	const range = node?.rangeBy({ index, endIndex, word }) ?? {};
```
romainmenke, line 34: @Mouvedia Can you elaborate on this change?
Why is `word` included here?

I am fairly certain that this is a breaking change.

See : https://github.com/postcss/postcss/blob/f9a9868eb3a591cd99d504ac57bf7ee77380856b/lib/node.js#L222-L229

Mouvedia: #7779

PR #7770, opened 2024-06-18
Fix `media-feature-name-no-vendor-prefix` autofix and refactor to use `fix` callback instead of `context.fix`

Mouvedia:

#2643

[demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIACBbGUCWBDABACgFoB3GAIwGsMAXAlDAOwNgDcMwYCAHDADxgBsCAJyxUMEeDgAMAShxZ6UfMTKUaKLDyYxW7LrwHDR4yQCY5wAL4gANCABmGfjAByWNAhAwe7zs4B0YADOQbbgEPSOAOaewAA69Dg4cSBCAK7OQSmS8YlJySBomFgE9jCiaUIc9O7VEATMMIoQQlxVjjzZOFTpMAlJlgnWdpCRGFEAYi0aVJ4AVkERYbCcoYi5SSlBVACezvwMVF0p-KIw2yk2-QXbewKHBKPRBNsKUFhCUMcgp1TnRyAhiBLEA) of the bug

### DIFF
--- a/lib/rules/media-feature-name-no-vendor-prefix/index.mjs
+++ b/lib/rules/media-feature-name-no-vendor-prefix/index.mjs
@@ -15,7 +15,7 @@ const meta = {
 };
 
 /** @type {import('stylelint').Rule} */
-const rule = (primary, _secondaryOptions, context) => {
+const rule = (primary) => {
 	return (root, result) => {
 		const validOptions = validateOptions(result, ruleName, { actual: primary });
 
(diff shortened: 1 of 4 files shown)


lib/rules/media-feature-name-no-vendor-prefix/index.mjs
```diff
@@ -30,25 +30,41 @@ const rule = (primary, _secondaryOptions, context) => {
 				return;
 			}
 
-			const matches = atRule.toString().match(/(?:min-|max-)?-[a-z-]+device-pixel-ratio/gi);
+			const features = [
```
ybiquitous, line 33: [suggest] How about extracting a constant for the features outside the loop like this?

```js
const FEATURES = [ /*...*/ ];

/** @type {import('stylelint').Rule} */
const rule = (primary) => {
  // ...

  const matches = atRuleString.match(new RegExp(`(?:${FEATURES.join('|')})`, 'gi'));
```

ybiquitous (commented):
Thank you. Good catch. 👍🏼 
Overall LGTM, but can you take a look at my refactoring suggestion?

ybiquitous (approved):

PR #7772, opened 2024-06-19
Refactor `at-rule-no-vendor-prefix` to use `fix` callback instead of `context.fix` and fix false negatives

Mouvedia:

#2643

- false negatives
  - `@-moz-document`: https://developer.mozilla.org/en-US/docs/Web/CSS/@document#browser_compatibility
  - `@-webkit-viewport`
    - https://trac.webkit.org/browser/webkit/trunk/LayoutTests/css3/device-adapt/viewport-at-rule-parsing.html?rev=133084
    - https://github.com/WebKit/WebKit/commit/faab0fc796cf0ac3af2a05f928f032d433d3a726
- don't exist
  - `@resolution`
    - https://github.com/postcss/autoprefixer/blob/main/data/prefixes.js#L815-L823
    - https://caniuse.com/css-media-resolution
  - `@-khtml-keyframes`
    - https://github.com/postcss/autoprefixer/blob/main/data/prefixes.js#L109
    - https://github.com/postcss/autoprefixer/commit/fa4d7263ce9e6a3792201577d57ebcfdce8afde9

### DIFF
--- a/.changeset/afraid-windows-play.md
+++ b/.changeset/afraid-windows-play.md
@@ -1,5 +1,5 @@
 ---
-"stylelint": patch
+"stylelint": minor
 ---
 
-Fixed: `at-rule-no-vendor-prefix` false negatives
+Fixed: `at-rule-no-vendor-prefix` false negatives for `@-moz-document` and `@-webkit-viewport`
--- a/.changeset/afraid-windows-play.md
+++ b/.changeset/afraid-windows-play.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `at-rule-no-vendor-prefix` false negatives
--- a/lib/utils/isAutoprefixable.cjs
+++ b/lib/utils/isAutoprefixable.cjs
@@ -25,12 +25,13 @@ const vendor = require('./vendor.cjs');
 const AT_RULES = new Set([
 	'@-khtml-keyframes',
 	'@-moz-keyframes',
+	'@-moz-document',
 	'@-ms-keyframes',
 	'@-ms-viewport',
 	'@-o-keyframes',
 	'@-o-viewport',
 	'@-webkit-keyframes',
-	'@resolution',
+	'@-webkit-viewport',
 ]);
 
 /**
--- a/lib/utils/isAutoprefixable.cjs
+++ b/lib/utils/isAutoprefixable.cjs
@@ -23,7 +23,6 @@ const vendor = require('./vendor.cjs');
  * Object.keys(prefixes.remove).filter((s) => s.startsWith('@'));
  */
 const AT_RULES = new Set([
-	'@-khtml-keyframes',
 	'@-moz-keyframes',
 	'@-moz-document',
 	'@-ms-keyframes',
--- a/lib/utils/isAutoprefixable.mjs
+++ b/lib/utils/isAutoprefixable.mjs
@@ -21,12 +21,13 @@ import vendor from './vendor.mjs';
 const AT_RULES = new Set([
 	'@-khtml-keyframes',
 	'@-moz-keyframes',
+	'@-moz-document',
 	'@-ms-keyframes',
 	'@-ms-viewport',
 	'@-o-keyframes',
 	'@-o-viewport',
 	'@-webkit-keyframes',
-	'@resolution',
+	'@-webkit-viewport',
 ]);
 
 /**
--- a/lib/utils/isAutoprefixable.mjs
+++ b/lib/utils/isAutoprefixable.mjs
@@ -19,7 +19,6 @@ import vendor from './vendor.mjs';
  * Object.keys(prefixes.remove).filter((s) => s.startsWith('@'));
  */
 const AT_RULES = new Set([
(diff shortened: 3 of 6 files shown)


ybiquitous:
Thanks for opening this PR. Can you provide a reproducible demo for the false positive/negative?

Mouvedia:
> Can you provide a reproducible demo for the false <strike>positive</strike>/negative?

[demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIACBaAthAXiqEwFc0YA7AFwAJ8AnAGwAoAdEAC1NIAcBneAeh4HdBAOhgAPAIZp2NGEMhoeTAJTlgAXwbFNqfjABGAawCWpFADcjMfuwhUK6zdpQRzl67ftqQAGhAAzIxkAOUk4RDFJaVkwTk4fcAhiAIBzBBBgTXJyJip8GU4meFVMrOyQcVNcmRRiFzMSXCoUdioYANFC8lJcmBKNYi9fSCSjZIAxWzQKtIArTkT42C40jOIspk5SAE8ZGiMyTqYaCphNpk0vNSA)

lib/utils/isAutoprefixable.mjs
```diff
@@ -21,12 +21,13 @@ import vendor from './vendor.mjs';
 const AT_RULES = new Set([
 	'@-khtml-keyframes',
 	'@-moz-keyframes',
+	'@-moz-document',
 	'@-ms-keyframes',
 	'@-ms-viewport',
 	'@-o-keyframes',
 	'@-o-viewport',
 	'@-webkit-keyframes',
-	'@resolution',
```
ybiquitous, line 29: [ask] This `@resolution` is output from the `@example` code. I'm not sure why this unprefixed rule is included. Can you confirm autoprefixer, just in case?

lib/utils/isAutoprefixable.cjs
```diff
@@ -25,12 +25,13 @@ const vendor = require('./vendor.cjs');
 const AT_RULES = new Set([
 	'@-khtml-keyframes',
```
Mouvedia, line 26: WIP

This one doesn't exist AFAIK.

lib/utils/isAutoprefixable.mjs
Mouvedia: I have already put a link that goes directly to the lines in their codebase.

Mouvedia:
I thought that refactors would be a cinch but every time I attempt one, there's a fix looming.

lib/utils/isAutoprefixable.mjs
ybiquitous: Honestly, I'm not sure about `@resolution`, but this PR's changes will be away from the code generation by the `@example` scripts:

- https://github.com/stylelint/stylelint/blob/d273483ac942380d98c28b591d7005e613d0776e/lib/utils/isAutoprefixable.mjs#L8-L14
- https://github.com/stylelint/stylelint/blob/d273483ac942380d98c28b591d7005e613d0776e/lib/utils/isAutoprefixable.mjs#L18-L19

I worry that we might be confused about the difference between this PR's changes and output from `@example` scripts in the future.

So, I suggest leaving some comments to clarify this code change intention like this:

```js
AT_RULES.delete("@-khtml-keyframes"); // <some reason>
AT_RULES.delete("@resolution"); // <some reason>
AT_RULES.add("@-moz-document"); // <some reason>
```

Of course, I understand this way is a bit ugly, so I welcome a better idea if any.

ybiquitous:
You can focus on refactoring if you want.

lib/utils/isAutoprefixable.mjs
Mouvedia: I think we should do away with the examples and replace them with an explanation of the util in its entirety placed [before the `export`](https://github.com/stylelint/stylelint/blob/main/lib/utils/isAutoprefixable.mjs#L306-L309).
For now what matters is the code itself not that we weren't careful enough during #5312.
I would like for it to be a separate documentation PR though.

Mouvedia:
> You can focus on refactoring if you want.

There's no way I would prioritize the refactor over fixing a bug :)

lib/utils/isAutoprefixable.mjs
ybiquitous: Agree. It seems we need to improve the document in this file. 👍🏼

ybiquitous (commented):
Thank you. Overall LGTM. But I just left a few suggestions for the changelog. Can you have a look?

.changeset/afraid-windows-play.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `at-rule-no-vendor-prefix` false negatives
```
Mouvedia, line 5:
```suggestion
"stylelint": minor
---

Fixed: `at-rule-no-vendor-prefix` false negatives for `@-moz-document` and `@-webkit-viewport`
```

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #1537, opened 2016-06-25
New rule `at-rule-no-unknown` (#1515)

cs-forever-ex:

Issue #1515, opened 2016-06-23
at-rule-no-unknown


evilebottnawi:
Valid:
```css
@media {
  /* Some code */
}
```
Invalid:
```css
@some-strange-at-rule {
  /* Some code */
}
```
With option `ignoreAtRule: ['string', '/regex/']`.

jeddy3:
Sure. 

- Name: `at-rule-no-unknown`
- Primary option: `true`
- Secondary options: `ignoreAtRules: []` (user defined array of custom at-rules)
- Message: "Unexpected unknown at-rule"
- Section: "At-rule"

We're going to have to draw a line in the sand here to define "unknown" though. Else where we've gone with present and past (read: obsolete) spec things being *known*, and preprocessor language extensions and "future" spec stuff as *unknown*. This latter one is ambiguous though and we need to nail it down for this rule as there a [lots](https://tabatkins.github.io/specs/css-when-else/) [of](http://tabatkins.github.io/specs/css-extend-rule/) ["future"](http://tabatkins.github.io/specs/css-apply-rule/) stuff around at-rules.

Anyone got any ideas? Perhaps ["Recommendation"](https://www.w3.org/2005/10/Process-20051014/tr.html#maturity-levels) and above?

As an aside, even when this is label as ready-to-go can we continue to focus on squashing bugs, updating docs, and the [next and `7.0`](https://github.com/stylelint/stylelint/milestones) milestones before adding new features? This also gives the wider community the opportunity to step up and have a go at adding these new features, and give us the time to [support](https://github.com/stylelint/stylelint/pull/1506) [them](https://github.com/stylelint/stylelint/pull/1410) in their endeavour.

evilebottnawi:
@jeddy3 agree

evilebottnawi:
@jeddy3 good question, stylelint should stick `Recomedation` only or not?
/cc @davidtheclark What do you think?

jeddy3:
> If we choise Recommendation, some rules need change behavior (color-hex-length and etc)

I'm not suggesting a stylelint-wide change of behaviour here. I'm suggestion just removing the ambiguity of what we mean by "unknown" in the `*-no-unknown"` rules.

evilebottnawi:
@jeddy3 I think this is a bit excessive fragmentation.

jeddy3:
> Anyone got any ideas? Perhaps "Recommendation" and above?

[...]

PR #7774, opened 2024-06-19
Fix `at-rule-no-unknown` false positives for `@historical-forms` and `@font-palette-values`

Mouvedia:

#1515

1.
`@historical-forms` was added after #1537 was merged.
ref: https://www.w3.org/TR/2018/WD-css-fonts-4-20180920/#font-feature-values-syntax

2.
[`@font-palette-values`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-palette-values)

### DIFF
--- a/lib/rules/at-rule-no-unknown/__tests__/index.mjs
+++ b/lib/rules/at-rule-no-unknown/__tests__/index.mjs
@@ -96,10 +96,6 @@ testRule({
 		{
 			code: '@scroll-timeline foo {}',
 		},
-		{
-			code: '@color-profile --swop5c {}',
-			skip: true,
-		},
 	],
 
 	reject: [
--- a/lib/rules/at-rule-no-unknown/__tests__/index.mjs
+++ b/lib/rules/at-rule-no-unknown/__tests__/index.mjs
@@ -96,6 +96,10 @@ testRule({
 		{
 			code: '@scroll-timeline foo {}',
 		},
+		{
+			code: '@color-profile --swop5c {}',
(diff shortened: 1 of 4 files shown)


lib/rules/at-rule-no-unknown/__tests__/index.mjs
```diff
@@ -96,6 +96,10 @@ testRule({
 		{
 			code: '@scroll-timeline foo {}',
 		},
+		{
+			code: '@color-profile --swop5c {}',
+			skip: true,
```
ybiquitous, line 101: [ask] Why skipped? Even if there is no browser support, is adding `@color-profile` a problem?

ybiquitous (commented):
Thanks, nice catch. 👍🏼 

But I have a question for `@color-profile`.

lib/rules/at-rule-no-unknown/__tests__/index.mjs
Mouvedia: The problem is that we don't have a policy regarding unimplemented features.
#7510 is not even resolved after all. This will do for now.
i.e. if it was behind a flag the addition could be considered

see also Fyrd/caniuse#5958

ybiquitous: Okay, so can you remove the case to avoid any confusion? If it's requested, then we can add it again.

ybiquitous (approved):
Thanks. LGTM 👍🏼

PR #7776, opened 2024-06-20
Refactor `selector-attribute-quotes` to use `fix` callback instead of `context.fix`

Mouvedia:

#2643

nit
off-topic
`color-hex-alpha` and `no-irregular-whitespace` should be refactored to follow the `rejected` convention


ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #7777, opened 2024-06-20
Refactor `font-family-name-quotes` to use `fix` callback instead of `context.fix` and fix false negatives

Mouvedia:

#2643

ref
https://github.com/WebKit/WebKit/blob/main/Source/WebCore/css/CSSValueKeywords.in
https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariCSSRef/Articles/StandardCSSProperties.html#//apple_ref/doc/uid/TP30001266-font
https://developer.mozilla.org/en-US/docs/Web/CSS/Mozilla_Extensions#font-family
http://help.dottoro.com/lclmgrtl.php
https://groups.google.com/a/chromium.org/g/blink-dev/c/TVRkV9XrFGo

### DIFF
--- a/lib/reference/keywords.mjs
+++ b/lib/reference/keywords.mjs
@@ -2,7 +2,6 @@ import uniteSets from '../utils/uniteSets.mjs';
 
 export const basicKeywords = new Set(['initial', 'inherit', 'revert', 'revert-layer', 'unset']);
 
-/** @todo add to {@link fontFamilyKeywords} */
 export const systemFontKeywords = uniteSets(basicKeywords, [
 	'caption',
 	'icon',
--- a/lib/reference/keywords.mjs
+++ b/lib/reference/keywords.mjs
@@ -2,6 +2,7 @@ import uniteSets from '../utils/uniteSets.mjs';
 
 export const basicKeywords = new Set(['initial', 'inherit', 'revert', 'revert-layer', 'unset']);
 
+/** @todo add to {@link fontFamilyKeywords} */
 export const systemFontKeywords = uniteSets(basicKeywords, [
 	'caption',
 	'icon',
@@ -23,6 +24,56 @@ export const fontFamilyKeywords = uniteSets(basicKeywords, [
 	'ui-monospace',
 	'ui-rounded',
 ]);
+
+const appleSystemFonts = new Set([
+	'-apple-system',
+	'-apple-system-headline',
+	'-apple-system-body',
+	'-apple-system-subheadline',
+	'-apple-system-footnote',
+	'-apple-system-caption1',
+	'-apple-system-caption2',
+	'-apple-system-short-headline',
+	'-apple-system-short-body',
+	'-apple-system-short-subheadline',
+	'-apple-system-short-footnote',
+	'-apple-system-short-caption1',
+	'-apple-system-tall-body',
+	'-apple-system-title0',
+	'-apple-system-title1',
+	'-apple-system-title2',
+	'-apple-system-title3',
+	'-apple-system-title4',
+]);
+
+const mozillaSystemFonts = new Set([
+	'-moz-button',
+	'-moz-desktop',
+	'-moz-dialog',
+	'-moz-document',
+	'-moz-field',
+	'-moz-fixed',
+	'-moz-info',
+	'-moz-list',
+	'-moz-pull-down-menu',
+	'-moz-window',
+	'-moz-workspace',
+]);
+
+const webkitSystemFonts = new Set([
+	'-webkit-body',
+	'-webkit-control',
+	'-webkit-mini-control',
+	'-webkit-pictograph',
+	'-webkit-small-control',
+	'-webkit-standard',
+]);
+
+export const prefixedSystemFonts = uniteSets(
+	appleSystemFonts,
+	mozillaSystemFonts,
+	webkitSystemFonts,
+);
 
 export const fontWeightRelativeKeywords = new Set(['bolder', 'lighter']);
 
--- a/lib/rules/font-family-name-quotes/index.mjs
+++ b/lib/rules/font-family-name-quotes/index.mjs
@@ -1,5 +1,5 @@
+import { fontFamilyKeywords, prefixedSystemFonts } from '../../reference/keywords.mjs';
 import findFontFamily from '../../utils/findFontFamily.mjs';
-import { fontFamilyKeywords } from '../../reference/keywords.mjs';
 import isStandardSyntaxValue from '../../utils/isStandardSyntaxValue.mjs';
 import isVariable from '../../utils/isVariable.mjs';
 import report from '../../utils/report.mjs';
@@ -23,7 +23,7 @@ const meta = {
  * @returns {boolean}
  */
 function isSystemFontKeyword(font) {
-	if (font.startsWith('-apple-')) {
+	if (prefixedSystemFonts.has(font)) {
 		return true;
 	}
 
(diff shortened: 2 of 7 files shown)


Mouvedia:
@ybiquitous 

> Error: Codecov token not found. Please provide Codecov token with -t flag.
> Error: Codecov:
>                         Failed to properly create report: The process '/home/runner/work/_actions/codecov/codecov-action/v4/dist/codecov' failed with exit code 1

Is this something new or being worked on?

lib/reference/keywords.mjs
```diff
@@ -2,6 +2,7 @@ import uniteSets from '../utils/uniteSets.mjs';
 
 export const basicKeywords = new Set(['initial', 'inherit', 'revert', 'revert-layer', 'unset']);
 
+/** @todo add to {@link fontFamilyKeywords} */
```
Mouvedia, line 5: note

This might be a bug. This needs to be investigated.
i.e. it should be `export const fontFamilyKeywords = uniteSets(basicKeywords, systemFontKeywords, [`
I don't want to change this in this PR.
i.e. `fontFamilyKeywords` is imported by other rules

ybiquitous:
> Is this something new or being worked on?

No changes at all. Can you retry?

lib/reference/keywords.mjs
ybiquitous: What reason of `const fontFamilyKeywords = uniteSets(basicKeywords, systemFontKeywords)`?

Mouvedia:
> Can you retry?

Already retried once.
Could https://github.com/codecov/codecov-action/commit/7594baaf0c8b9ecdddba6269a6b089f60a3c859a be related?

lib/reference/keywords.mjs
Mouvedia: Let's discuss that in its own PR or issue.
In short those keywords are accepted by `font-family` too.

lib/rules/font-family-name-quotes/index.mjs
```diff
@@ -187,75 +181,57 @@ const rule = (primary, _secondary, context) => {
...
-						return complain(messages.expected(family), rawFamily, decl);
+						return complain('expected', fontFamilyNode, decl);
 					}
 			}
 		}
 
 		/**
-		 * @param {string} message
-		 * @param {string} family
+		 * @param {'expected'|'rejected'} messageType
```
ybiquitous, line 212: [suggest] `keyof` can reduce duplication:

```suggestion
		 * @param {keyof messages} messageType
```

lib/rules/font-family-name-quotes/index.mjs
```diff
@@ -187,75 +181,57 @@ const rule = (primary, _secondary, context) => {
...
+			const { name, rawName } = fontFamilyNode;
+			const fix = () => {
+				const mutate = messageType === 'expected' ? 'addQuotes' : 'removeQuotes';
+				const end = rawName.length - 1;
+				const word = messageType === 'expected' ? `"${rawName}"` : rawName.slice(1, end);
+
+				fontFamilyNode[mutate]();
+
+				return decl.rangeBy({ word });
+			};
```
ybiquitous, line 226: [suggest] What about an idea that `addQuotes` and `removeQuotes` return a range since they already have the `slice` logic? For example:

```js
{
	removeQuotes() {
		// ...
		return decl.rangeBy({ word: this.name });
	},
	addQuotes() {
		// ...
		return decl.rangeBy({ word: fixedName });
	},
}
```

Then, `fix` would be much simpler:

```js
const fix = () => {
	return messageType === 'expected'
		? fontFamilyNode.addQuotes()
		: fontFamilyNode.removeQuotes();
	};
```

lib/reference/keywords.mjs
ybiquitous: Okay. At this point, can you remove this comment to avoid any confusion by other people? If this is not a bug, this comment would be unnecessary.

ybiquitous:
> A new version of codecov-action was released 2 days ago.

Hum, a similar issue was reported (https://github.com/codecov/codecov-action/issues/1487). I'll look into it. 👀

ybiquitous:
Maybe, https://github.com/codecov/codecov-action/pull/1471 seems a related change in https://github.com/codecov/codecov-action/releases/tag/v4.5.0 🤔

lib/rules/font-family-name-quotes/index.mjs
Mouvedia: `no-confusing-arrow` rule conflicts with prettier; it's kinda annoying.

ybiquitous:
@Mouvedia Codecov CLI seems a cause. Can you try the following in this PR? 👇🏼 

```diff
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -53,6 +53,7 @@ jobs:
           files: ./.coverage/lcov.info
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true
+          version: '0.6.0' # TODO: Workaround for https://github.com/codecov/codecov-action/issues/1487
 
   build:
     runs-on: ubuntu-latest
```

lib/reference/keywords.mjs
Mouvedia: I put it here so that it wouldn't be forgotten. Ill remove it.

lib/rules/font-family-name-quotes/index.mjs
ybiquitous: Agree. I noticed the [`no-confusing-arrow`](https://eslint.org/docs/latest/rules/no-confusing-arrow) rule had been deprecated because of the stylistic rule.

I will remove it from our shared config:
https://github.com/stylelint/eslint-config-stylelint/blob/c3d49f9cd9040aee87132b2b070b4685be963703/index.js#L26-L31

Mouvedia:
> Can you try the following in this PR? 👇🏼
> 
> ```diff
> --- a/.github/workflows/testing.yml
> +++ b/.github/workflows/testing.yml
> @@ -53,6 +53,7 @@ jobs:
>            files: ./.coverage/lcov.info
>            token: ${{ secrets.CODECOV_TOKEN }}
>            fail_ci_if_error: true
> +          version: '0.6.0' # TODO: Workaround for https://github.com/codecov/codecov-action/issues/1487
>  
>    build:
>      runs-on: ubuntu-latest
> ```

> gpg: no valid OpenPGP data found.
> gpg: the signature could not be verified.
>Please remember that the signature file (.sig or .asc) should be the first file given on the command line.
> Error: Codecov: Error verifying gpg signature: Command failed: gpg --logger-fd 1 --verify /home/runner/work/_actions/codecov/codecov-action/v4/dist/codecov.SHA256SUM.sig /home/runner/work/_actions/codecov/codecov-action/v4/dist/codecov.SHA256SUM

Mouvedia:
You can now commit on this branch yourself if you want to.

lib/rules/font-family-name-quotes/index.mjs
ybiquitous: Opened https://github.com/stylelint/eslint-config-stylelint/pull/276

ybiquitous (approved):
The workaround for codecov-action works (see 7e1ac212d and 9ae7505dc). 👍🏼 

Thank you. LGTM 👍🏼

Mouvedia:
> The workaround for codecov-action works.

Can you open a separate PR and reset this PR to 1489268?
If we ever upgrade to 0.7 we will want a specific PR to refer to not an unrelated PR like this one.
Ill rebase on main once that's merged.

ybiquitous:
Make sense. Opened PR #7778 👍🏼

ybiquitous:
I've merged #7778 and taken the main change into this branch. Now ready to merge. 👍🏼

Mouvedia:
Thanks for the review.

PR #7778, opened 2024-06-21
Add workaround for codecov-action

ybiquitous:

Ref #7777

This aims to avoid codecov-action upload failures in a PR by forked repositories. The old Codecov CLI version is used, instead of `latest`.

See also https://github.com/codecov/codecov-action/issues/1487


Mouvedia (approved):
I cannot be sure that it's the same issue though.

PR #7780, opened 2024-06-21
Refactor `report()` variable names

romainmenke:

None

See review comments.

These are some minor and superficial nitpicks.
I don't feel strongly about these, but I think they improve readability.

### DIFF
--- a/lib/utils/report.cjs
+++ b/lib/utils/report.cjs
@@ -168,16 +168,16 @@ function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
 		return false;
 	}
 
-	const hasFix = Boolean(config.fix && !config.rules?.[ruleName][1]?.disableFix);
-	const fixed = hasFix && (config.ignoreDisables || !isDisabled(ruleName, line, disabledRanges));
+	const shouldFix = Boolean(config.fix && !config.rules?.[ruleName][1]?.disableFix);
+	const mayFix = config.ignoreDisables || !isDisabled(ruleName, line, disabledRanges);
 
-	if (fixed) {
-		addFixData({ fixersData, ruleName, fixed, range: fix() ?? undefined });
+	if (shouldFix && mayFix) {
+		addFixData({ fixersData, ruleName, fixed: true, range: fix() ?? undefined });
 
 		return true;
 	}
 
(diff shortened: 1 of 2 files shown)


lib/utils/report.cjs
```diff
@@ -168,16 +168,16 @@ function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
 		return false;
 	}
 
-	const hasFix = Boolean(config.fix && !config.rules?.[ruleName][1]?.disableFix);
-	const fixed = hasFix && (config.ignoreDisables || !isDisabled(ruleName, line, disabledRanges));
+	const shouldFix = Boolean(config.fix && !config.rules?.[ruleName][1]?.disableFix);
```
romainmenke, line 171: `hasFix` is easily confused with "has a fixer callback".
While these conditions check the config to make sure that fixes **should** be applied when possible.

lib/utils/report.cjs
```diff
@@ -168,16 +168,16 @@ function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
...
+	const mayFix = config.ignoreDisables || !isDisabled(ruleName, line, disabledRanges);
```
romainmenke, line 172: `fixed` is confusing because it implies mutation of the source, which is not what happens on this line.

The conditions only check if fixing is allowed.

lib/utils/report.cjs
```diff
@@ -191,9 +191,9 @@ function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
  * @todo stylelint/stylelint#7192
  */
 function addFixData({ fixersData, ruleName, range, fixed }) {
-	const array = (fixersData[ruleName] ??= []);
+	const ruleFixersData = (fixersData[ruleName] ??= []);
```
romainmenke, line 194: I always try to avoid using broad programming concepts like `array` as variable names. I even prefer something abstract like `x` (`x.push()` over `array`.

ybiquitous: That makes sense. In addition, I prefer `ruleFixers` or `fixers` a bit more. " Data" seems a bit verbose to me.

But it's just my preference. Let's keep it as-is if you're fine. 👍🏼

PR #7781, opened 2024-06-21
Refactor to remove `word` from the object passed to `rangeBy` in `report` util

### LINKED ISSUE #7779: Address potentially breaking changes in #7769
I am fairly certain that an unwanted breaking change was introduced in : https://github.com/stylelint/stylelint/pull/7769#pullrequestreview-2132231017

Adding `word` as an extra param to the `node.rangeBy` has side-effects:
- extra serialization of AST nodes, which will have a performance impact
- `word` has precedence over `index` and `endIndex` so the resulting `Range` might be different than it was before for plugins that pass both `index`, `endIndex` and `word` to `report()`

This change was not discussed in that PR.
It should at least be addressed before we release a new version.

Mouvedia:

Closes #7779

This PR

1. revert this change

```diff
-rangeBy({ index, endIndex })
+rangeBy({ index, endIndex, word })
```

2. addresses the recent changes that assumed that `index` would shift the search for the needle when used with `word`

</hr>

As an aside, we have legitimate cases where passing just `node` is simpler than passing let's say `node`+`word`.
e.g. if we have to take into account the last semicolon
i.e. we cannot be as strict as not allowing passing `word` or at least it won't be covered by that PR

### DIFF
--- a/lib/rules/no-invalid-double-slash-comments/index.mjs
+++ b/lib/rules/no-invalid-double-slash-comments/index.mjs
@@ -1,4 +1,3 @@
-import { assert } from '../../utils/validateTypes.mjs';
 import report from '../../utils/report.mjs';
 import ruleMessages from '../../utils/ruleMessages.mjs';
 import validateOptions from '../../utils/validateOptions.mjs';
@@ -45,10 +44,9 @@ const rule = (primary) => {
 					const offset = value.length - selector.length;
 					const i = index + offset;
 					const lines = ruleNode.toString().slice(i).split('\n');
-					const comment = lines[0];
-					const word = lines.length > 1 ? `${comment}\n` : comment;
+					let comment = lines[0] ?? '';
 
-					assert(word);
+					if (lines.length > 1) comment += '\n';
 
 					report({
 						message: messages.rejected,
@@ -56,7 +54,7 @@ const rule = (primary) => {
 						result,
 						ruleName,
 						index: i,
-						endIndex: i + word.length,
+						endIndex: i + comment.length,
 					});
 				}
 
--- a/lib/rules/no-invalid-double-slash-comments/index.mjs
+++ b/lib/rules/no-invalid-double-slash-comments/index.mjs
@@ -35,6 +35,7 @@ const rule = (primary) => {
 		root.walkRules((ruleNode) => {
 			// ruleNode.selectors elements are trimmed
 			const selectors = ruleNode.selector.split(',');
+			const ruleStringified = ruleNode.toString();
 			let index = 0;
 
 			for (const value of selectors) {
@@ -43,7 +44,7 @@ const rule = (primary) => {
 				if (selector.startsWith('//')) {
 					const offset = value.length - selector.length;
 					const i = index + offset;
-					const lines = ruleNode.toString().slice(i).split('\n');
+					const lines = ruleStringified.slice(i).split('\n');
 					let comment = lines[0] ?? '';
 
 					if (lines.length > 1) comment += '\n';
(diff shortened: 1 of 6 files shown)


lib/rules/no-invalid-double-slash-comments/index.mjs
romainmenke, line 1: Is it correct that these rules were previously changed, after support for `word` was added in the `report()` range calculation? 

And that now they need to be changed again so that they correctly report `index` and `endIndex`.

-----

Or are these purely new changes?

lib/rules/no-invalid-double-slash-comments/index.mjs
```diff
@@ -47,13 +48,15 @@ const rule = (primary) => {
 					const comment = lines[0];
 					const word = lines.length > 1 ? `${comment}\n` : comment;
 
+					assert(word);
```
romainmenke, line 51: This is purely because TypeScript always assumes that `arr[0]` can be nullish right?

lib/rules/selector-no-vendor-prefix/index.mjs
```diff
@@ -64,10 +64,13 @@ const rule = (primary, secondaryOptions) => {
 				}
 
 				const fix = () => {
+					const index = pseudoNode.sourceIndex;
+					const endIndex = index + pseudoNode.value.length;
```
romainmenke, line 68:
```suggestion
					const endIndex = index + value.length;
```

At this point `value` is the same as `pseudoNode.value`, but maybe that is not intentional?

If the intention is to return the range of the fixed text then these indices should be calculated after the mutation.

romainmenke (commented):
Some small questions :)
Thanks!

lib/rules/no-invalid-double-slash-comments/index.mjs
ybiquitous: I have the same question. And, instead of `word`, reusing `comment` seems better to me like this:

```js
let comment = lines[0] ?? '';
if (lines.length > 1) comment += '\n';

endIndex: i + comment.length,
```

lib/rules/no-invalid-double-slash-comments/index.mjs
ybiquitous, line 1: [ask] From a performance perspective, moving `ruleNode.toString()` outside the `for...of` loop may be better. Any thoughts?

https://github.com/stylelint/stylelint/blob/942500e2e38d16fdc7bf4a2f34f63441c1bc852e/lib/rules/no-invalid-double-slash-comments/index.mjs#L41

https://github.com/stylelint/stylelint/blob/942500e2e38d16fdc7bf4a2f34f63441c1bc852e/lib/rules/no-invalid-double-slash-comments/index.mjs#L47

lib/rules/no-invalid-double-slash-comments/index.mjs
Mouvedia: At this point it should be string that's why I used an `assert`.
If it's not we have a problem.

I don't mind replacing the assert with `?? ''`.

lib/rules/selector-no-vendor-prefix/index.mjs
Mouvedia: All the ranges currently return the range of fixed text.
We can discuss to change that behaviour in another PR; this is off topic.

> these indices should be calculated after the mutation

My bad. Fix incoming.

lib/rules/no-invalid-double-slash-comments/index.mjs
Mouvedia: As explained in [OP](https://github.com/stylelint/stylelint/pull/7781#issue-2366769934):

> addresses the recent changes that assumed that `index` would shift the search for the needle when used with `word`

If you are asking whether we need a new changeset, no we don't the current one covers this change.
see https://github.com/stylelint/stylelint/pull/7768#discussion_r1643579489 for details

ybiquitous (approved):
Thanks. LGTM 👍🏼

romainmenke (approved):
LGTM!

Issue #7784, opened 2024-06-21
Autofix refactors


romainmenke:
Requests:
- https://github.com/stylelint/stylelint/issues/2643
- https://github.com/stylelint/stylelint/issues/7192
- https://github.com/stylelint/stylelint/issues/4664

Supporting tasks:
- https://github.com/stylelint/stylelint/issues/7787
- https://github.com/stylelint/stylelint/issues/7788
- https://github.com/stylelint/stylelint/issues/7836

(@Mouvedia were there others you think are related to this effort?)

---------

Thanks to the efforts by @Mouvedia we have already made a lot of progress on #2643 and I think that in doing so we also have a good active mental model for the parts involved.

I think that it would be good at this time to take a moment to evaluate where we want to end up.
I also want to ensure that we are happy with both the public interface and the internal implementations.

---------

###  Add stylelint-disable commands support to autofix #2643

- fixes should be gated behind range checks
- range checks must align with disable range checks for warnings

### Add replaced text and corresponding range to `Warning` type #7192

- no fixes can be applied
- data to enable an external driver to apply fixes must be exposed
- interop with other tools makes this feature more useful

### Add fixed field to Result object #4664

- for each fix we must also capture a report so that users can see why something was fixed

### Fix data as part of warnings

ESLint includes data for unfixed, but fixable issues in lint messages.
This approach makes it easy to resolve #7192.

But at the same time it makes it impossible to resolve #4664.

We can't expose fixed issues as part of warnings because the source of the warning would already have been mutated. Doing so anyway implies that all surrounding tools (editor integrations, formatters, ...) would also need to update.

At this time I would challenge the need to resolve #4664.
I think it is more important to solve the other requests and to align with tools like ESLint.

----------

#### exposed as: 
- https://eslint.org/docs/latest/integrate/nodejs-api#-editinfo-type

This is a `fix` property on `LintMessage` with type [`fix (EditInfo | undefined)`](https://eslint.org/docs/latest/integrate/nodejs-api#-editinfo-type)

#### implemented as:
- https://eslint.org/docs/latest/extend/custom-rules#applying-fixes

```js
context.report({
    node: node,
    message: "Missing semicolon",
    fix(fixer) {
        return fixer.insertTextAfter(node, ";");
    }
});
```

#### Note worthy:
- data about fixes is part of lint messages, it is not something separate
- `fix()` provides a `fixer` object with helper methods for mutations
[...]

PR #7786, opened 2024-06-22
POC for #7784

Mouvedia:

#7192, #7784

- [x] new `text`
- [x] old `range`
- [x] fallback to warn range
- [x] does not require `fix` to be called (store the data even for the warn path)
- [x] ~~handles non-consecutive ranges (e.g. `declaration-block-no-redundant-longhand-properties`)~~

> What's the purpose of this PR?

1. To bikeshed the name.
I chose `editInfo` but I am sure there's something better.
ref: https://eslint.org/docs/latest/integrate/nodejs-api#-editinfo-type

2. To confirm the type of the `fix` function

3. to discuss if `editInfo` should be set on `fix` or passed directly to `report` or returned by `fix`

> What is this PR not about?

- the refactor of `alpha-value-notation`
- the eventual test errors of the other rules

### DIFF
--- a/lib/rules/alpha-value-notation/index.cjs
+++ b/lib/rules/alpha-value-notation/index.cjs
@@ -31,6 +31,8 @@ const ALPHA_FUNCTION = /(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)\(/i;
 const ALPHA_FUNCTION_NAME = /^(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)$/i;
 const DIGIT = /\d/;
 
+/** @import {Problem} from 'stylelint' */
+
 /** @type {import('stylelint').Rule} */
 const rule = (primary, secondaryOptions) => {
 	return (root, result) => {
@@ -102,14 +104,18 @@ const rule = (primary, secondaryOptions) => {
 				const unfixed = value;
 				const index = declarationValueIndex(decl) + alpha.sourceIndex;
 				const endIndex = index + alpha.value.length;
-				const fix = () => {
-					alpha.value = fixed;
-					setDeclarationValue(decl, parsedValue.toString());
-				};
 
-				fix.editInfo = {
-					// range: decl.rangeBy({ index, endIndex }),
-					text: fixed,
+				/** @type {Problem['fix']} */
+				const fix = (apply) => {
+					if (apply) {
+						alpha.value = fixed;
+						setDeclarationValue(decl, parsedValue.toString());
+					}
+
+					return {
+						// range: decl.rangeBy({ index, endIndex }),
+						text: fixed,
+					};
 				};
 
 				report({
--- a/lib/rules/alpha-value-notation/index.cjs
+++ b/lib/rules/alpha-value-notation/index.cjs
@@ -31,8 +31,6 @@ const ALPHA_FUNCTION = /(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)\(/i;
 const ALPHA_FUNCTION_NAME = /^(?:color|hsla?|rgba?|hwb|lab|lch|oklab|oklch)$/i;
 const DIGIT = /\d/;
 
-/** @import {Problem} from 'stylelint' */
-
 /** @type {import('stylelint').Rule} */
 const rule = (primary, secondaryOptions) => {
 	return (root, result) => {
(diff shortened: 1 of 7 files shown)


types/stylelint/index.d.ts
```diff
@@ -260,10 +260,17 @@ declare namespace stylelint {
 		end: Position;
 	};
 
-	type FixerData = {
-		range?: Range;
-		fixed: boolean;
-	};
+	type FixerData =
```
Mouvedia, line 263: @ybiquitous @romainmenke
Once you have read [OP](https://github.com/stylelint/stylelint/pull/7786#issue-2367884702), can you tell me if this type and [`Fix`](https://github.com/stylelint/stylelint/pull/7786/files#diff-9a4b261e295bef1ccce813aa194ac4023a1dcf98236667889964b1ba6fbc551fR647-R653) correspond to everything we laid out?

types/stylelint/index.d.ts
```diff
@@ -634,9 +641,17 @@ declare namespace stylelint {
...
 		 */
-		fix?: () => void | never | Range;
+		fix?: Fix;
 	};
 
+	interface Fix {
+		(): void | never;
+		editInfo?: {
+			range?: Range;
+			text: string;
```
Mouvedia, line 651: typo

fix incoming

types/stylelint/index.d.ts
romainmenke: Can we spend a bit more time in design phase before creating a POC?

Mouvedia: The POC is there to help the design phase going forward.
We have a concrete example that we can use to iterate further.
Apart from the name and where `editInfo` should be set/passed, every demand you laid out should be fulfilled.
Can you at least review the index.d.ts and give me your feedback?

romainmenke: I personally do not feel that code review is the most efficient way to design architectural changes like these. This is too taxing for me. Maybe someone else has an easier time to do design work as code review?

Mouvedia: Ill just copy paste the types in the issue then.
For me, at least, it would seem way easier to comprehend with a concrete example.
Ill leave the PR open so that @ybiquitous can give his feedback.

Mouvedia: https://github.com/stylelint/stylelint/issues/7784#issuecomment-2184094675

lib/rules/alpha-value-notation/index.cjs
```diff
@@ -102,9 +104,18 @@ const rule = (primary, secondaryOptions) => {
 				const unfixed = value;
 				const index = declarationValueIndex(decl) + alpha.sourceIndex;
 				const endIndex = index + alpha.value.length;
-				const fix = () => {
-					alpha.value = fixed;
-					setDeclarationValue(decl, parsedValue.toString());
+
+				/** @type {Problem['fix']} */
```
Mouvedia, line 108: Unless you pass `fix` directly to `report`, with the proposal 3, you will have to add the type for every declaration. That's a big CON but not a blocker.

types/stylelint/index.d.ts
```diff
@@ -634,9 +635,19 @@ declare namespace stylelint {
...
 
+	type EditInfo =
+		| {
+				range?: Range;
+				text: string;
+		  }
+		| {
+				range: Range[];
+				text: string[];
+		  };
```
Mouvedia, line 649: @romainmenke Should I remove this and `| Range[]`/`| string[]` from `FixerData`?
This commit illustrate your proposal.

romainmenke: >  Should I remove this and | Range[]/| string[] from FixerData?

Yes :)
I don't we think we need it at this time and anything that makes the POC simpler is always nice :)

lib/rules/alpha-value-notation/index.mjs
```diff
@@ -99,9 +101,18 @@ const rule = (primary, secondaryOptions) => {
 				const unfixed = value;
 				const index = declarationValueIndex(decl) + alpha.sourceIndex;
 				const endIndex = index + alpha.value.length;
-				const fix = () => {
-					alpha.value = fixed;
-					setDeclarationValue(decl, parsedValue.toString());
+
+				/** @type {Problem['fix']} */
+				const fix = (apply) => {
```
Mouvedia, line 106: questions
minor
@romainmenke @ybiquitous

1.
I find it annoying to have to add the type for every `const fix` declaration.
Can it be avoided?
Maybe having to import the `Problem` type for all fixable rules is not such a big deal?

2.
Are we sure we want the default to be `apply`?
i.e. it could also be `doNotApply`

ybiquitous: > I find it annoying that you have to add the type for every single `const fix` declaration.
> Can it be avoided?

Inlining the `fix` variable is a way. Otherwise, I think some type annotations are necessary.

For other questions about the function signature, we're discussing it on #7784. Let's continue there.

Mouvedia:
@ybiquitous you now have 3 commits for 3 of the 4 proposals listed in #7784.
That way, it's easier to compare.

PR #7800, opened 2024-06-24
Refactor `color-hex-length` to use `fix` callback instead of `context.fix`

Mouvedia:

#2643

N/A


ybiquitous (approved):
Thanks, LGTM 👍🏼

lib/rules/color-hex-length/index.cjs
Mouvedia, line 1: @ybiquitous I didn't refactor `isIgnoredFunction` in this PR.
Should we add `attr` to the set of `isIgnoredFunction` util?

ybiquitous: Sorry, I don't know the background.

PR #7801, opened 2024-06-24
Fix `function-no-unknown` performance by reducing file read count

ybiquitous:

None.

I happened to notice this chance for performance improvement.

In the `(root, result) => {}` function, `fs.readFileSync()` is called many times for each scanned file.
However, `functionsList` is immutable, so we can call it only once.

Benchmark:

```sh-session
$ cat tmp/stylelint.config.mjs
export default {
	rules: { 'function-no-unknown': true },
};

$ cat tmp/test.css
a {
  width: cal(1 + 2);
}

$ for i in $(seq 1 1000); do cp tmp/test.css "tmp/test-${i}.css"; done

$ time bin/stylelint.mjs -c tmp/stylelint.config.mjs "tmp/*.css" 2>/dev/null
```

On the main branch:

```
0.92s user 0.32s system 133% cpu 0.928 total
```

On this branch:

```
0.88s user 0.29s system 132% cpu 0.878 total
```

### DIFF
--- a/lib/rules/function-no-unknown/index.cjs
+++ b/lib/rules/function-no-unknown/index.cjs
@@ -25,6 +25,9 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/function-no-unknown',
 };
 
+/** @type {string[] | undefined} */
+let functionsListMemo;
+
 /** @type {import('stylelint').Rule} */
 const rule = (primary, secondaryOptions) => {
 	return (root, result) => {
@@ -45,7 +48,10 @@ const rule = (primary, secondaryOptions) => {
 			return;
 		}
 
-		const functionsList = JSON.parse(fs.readFileSync(functionsListPath.toString(), 'utf8'));
+		// Read the file lazily for performance.
+		functionsListMemo ??= JSON.parse(fs.readFileSync(functionsListPath.toString(), 'utf8'));
+
+		const functionsList = /** @type {string[]} */ (functionsListMemo);
 
 		root.walkDecls((decl) => {
 			const { value } = decl;
--- a/lib/rules/function-no-unknown/index.cjs
+++ b/lib/rules/function-no-unknown/index.cjs
@@ -2,9 +2,8 @@
 // please instead edit the ESM counterpart and rebuild with Rollup (npm run build).
 'use strict';
 
-const fs = require('node:fs');
+const node_module = require('node:module');
 const cssParserAlgorithms = require('@csstools/css-parser-algorithms');
-const functionsListPath = require('css-functions-list');
 const cssTokenizer = require('@csstools/css-tokenizer');
 const validateTypes = require('../../utils/validateTypes.cjs');
 const declarationValueIndex = require('../../utils/declarationValueIndex.cjs');
@@ -15,6 +14,13 @@ const report = require('../../utils/report.cjs');
 const ruleMessages = require('../../utils/ruleMessages.cjs');
 const validateOptions = require('../../utils/validateOptions.cjs');
 
+var _documentCurrentScript = typeof document !== 'undefined' ? document.currentScript : null;
+const require$1 = node_module.createRequire((typeof document === 'undefined' ? require('u' + 'rl').pathToFileURL(__filename).href : (_documentCurrentScript && _documentCurrentScript.src || new URL('lib/rules/function-no-unknown/index.cjs', document.baseURI).href)));
+
+// TODO: Replace this with Import attributes when dropping the support for older Node.js versions.
+// See https://nodejs.org/api/esm.html#import-attributes
+const functionsList = require$1('css-functions-list/index.json');
+
 const ruleName = 'function-no-unknown';
 
 const messages = ruleMessages(ruleName, {
@@ -24,9 +30,6 @@ const messages = ruleMessages(ruleName, {
 const meta = {
 	url: 'https://stylelint.io/user-guide/rules/function-no-unknown',
 };
-
-/** @type {string[] | undefined} */
-let functionsListMemo;
 
 /** @type {import('stylelint').Rule} */
 const rule = (primary, secondaryOptions) => {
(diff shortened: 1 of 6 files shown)


lib/rules/function-no-unknown/index.cjs
```diff
@@ -45,7 +48,10 @@ const rule = (primary, secondaryOptions) => {
 			return;
 		}
 
-		const functionsList = JSON.parse(fs.readFileSync(functionsListPath.toString(), 'utf8'));
+		// Read the file lazily for performance.
+		functionsListMemo ??= JSON.parse(fs.readFileSync(functionsListPath.toString(), 'utf8'));
```
Mouvedia, line 52:
```suggestion
functionsListMemo ??= readJSONFile(functionsListPath.toString());
```

That's in `testUtils` folder though.

ybiquitous: Unfortunately, `testUtils` is unpublished:
https://github.com/stylelint/stylelint/blob/a6b9b92e3f7d09d4b044b2eb24b9007ab33230f1/package.json#L57

Mouvedia:
I don't think it's worth it, but I thought I would mention it nonetheless.
If you were to change the `tsconfig.json` like so

```
{
	"compilerOptions": {
		"module": "NodeNext",
		"moduleResolution": "NodeNext",
		"resolveJsonModule": true,
		…
```

then you could do

```js
import foo from 'css-functions-list/index.json' assert { type: 'json' };
```

ybiquitous:
Unfortunately, it seems we cannot use [Import attributes](https://nodejs.org/api/esm.html#import-attributes) yet because:

- Node.js 18 doesn't support the feature.
- Node.js 20 supports it, but it's experimental.

See:

```sh-session
$ cat test-import-attributes.mjs
import functionsList from 'css-functions-list/index.json' with { type: 'json' };

console.log(functionsList[0]);

$ nvm exec 20 node test-import-attributes.mjs
Running node v20.14.0 (npm v10.8.1)
abs
(node:55164) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)

$ nvm exec 18 node test-import-attributes.mjs
Running node v18.12.0 (npm v8.19.2)
file:///Users/masafumi.koba/git/stylelint/stylelint/test-import-attributes.mjs:1
import functionsList from 'css-functions-list/index.json' with { type: 'json' };
                                                          ^^^^

SyntaxError: Unexpected token 'with'
    at ESMLoader.moduleStrategy (node:internal/modules/esm/translators:119:18)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:468:14)
    at async link (node:internal/modules/esm/module_job:67:21)

Node.js v18.12.0
```

Mouvedia:
That's why I was using `assert` and not `with` but it's not necessary in our case.
i.e. just `"resolveJsonModule": true` with `import foo from 'css-functions-list/index.json';` is sufficient

ybiquitous:
Even if `with`/`assert` is absent, we cannot use it:

```sh-session
$ cat test-import-attributes.mjs
import functionsList from 'css-functions-list/index.json';

console.log(functionsList[0]);

$ node test-import-attributes.mjs
node:internal/errors:484
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_IMPORT_ASSERTION_TYPE_MISSING]: Module "file:///Users/masafumi.koba/git/stylelint/stylelint/node_modules/css-functions-list/esm/index.json" needs an import assertion of type "json"
    at new NodeError (node:internal/errors:393:5)
    at validateAssertions (node:internal/modules/esm/assert:82:15)
    at defaultLoad (node:internal/modules/esm/load:84:3)
    at nextLoad (node:internal/modules/esm/loader:163:28)
    at ESMLoader.load (node:internal/modules/esm/loader:605:26)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:457:22)
    at new ModuleJob (node:internal/modules/esm/module_job:63:26)
    at #createModuleJob (node:internal/modules/esm/loader:480:17)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:434:34)
    at async ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:78:21) {
  code: 'ERR_IMPORT_ASSERTION_TYPE_MISSING'
}

Node.js v18.12.0
```

Mouvedia (approved):

Mouvedia:
There's also

```js
import { createRequire } from 'node:module';
const require = createRequire(import.meta.url);

const foo = require('css-functions-list/index.json');
```

ybiquitous:
@Mouvedia Your suggestion sounds better since it's the standard way to import a JSON file. Please take a look at 4fd66f901 again. 🙏🏼

lib/utils/mathMLTags.mjs
```diff
@@ -2,12 +2,11 @@ import { createRequire } from 'node:module';
...
+// TODO: mathml-tag-names v3 is a pure ESM package,
 // so we cannot update it while supporting both ESM and CommonJS.
 //
 // In addition, mathml-tag-names v2 provides only a JSON file,
 // so ESM cannot import it (raises the "ERR_IMPORT_ASSERTION_TYPE_MISSING" error).
-// @ts-expect-error -- mathml-tag-names v2 doesn't have type declarations.
-const mathMLTags = require('mathml-tag-names');
+const mathMLTags = require('mathml-tag-names/index.json');
 
 export default mathMLTags;
```
Mouvedia, line 12: minor

Can you add
```js
/** @type {string[]} */
```
?

Mouvedia (approved):
Have you seen what rollup is generating?
That's surprisingly inefficient.

ybiquitous:
> Have you seen what rollup is generating?
> That's surprisingly inefficient.

Yes, this is a known thing. Rollup transforms `createRequire` to such.

lib/utils/mathMLTags.mjs
ybiquitous: Yes, this type is correctly inferred as `string[]`. However, specifying a type annotation may reduce type inference cost by `tsc` and help us understand the type more efficiently.

So I've added a type annotation via 6760c2cef.

Mouvedia (approved):

PR #7802, opened 2024-06-24
Refactor `media-feature-range-notation` to use `fix` callback instead of `context.fix`

Mouvedia:

#2643

N/A

### DIFF
--- a/lib/rules/media-feature-range-notation/index.mjs
+++ b/lib/rules/media-feature-range-notation/index.mjs
@@ -28,7 +28,7 @@ const meta = {
 };
 
 /** @type {import('stylelint').Rule} */
-const rule = (primary, _secondaryOptions, context) => {
+const rule = (primary) => {
 	return (root, result) => {
 		const validOptions = validateOptions(result, ruleName, {
 			actual: primary,
@@ -41,7 +41,6 @@ const rule = (primary, _secondaryOptions, context) => {
 
 		root.walkAtRules(/^media$/i, (atRule) => {
 			const mediaQueryList = parseMediaQuery(atRule);
-			let hasFixes = false;
 
 			mediaQueryList.forEach((mediaQuery) => {
 				if (isMediaQueryInvalid(mediaQuery)) return;
@@ -61,17 +60,18 @@ const rule = (primary, _secondaryOptions, context) => {
 
 					if (!rangeTypeMediaFeatureNames.has(unprefixedMediaFeature)) return;
 
-					if (context.fix && primary === 'context' && isMediaFeaturePlain(node)) {
-						if (!isMediaFeature(parent)) return;
-
-						hasFixes = true;
-
+					/**
+					 * @param {object} entry
+					 * @param {import('@csstools/media-query-list-parser').MediaFeaturePlain} entry.node
+					 * @param {import('@csstools/media-query-list-parser').MediaFeature} entry.parent
+					 */
+					const contextFixer = (entry) => () => {
 						/** @type {import('@csstools/css-tokenizer').TokenDelim} */
 						const operator = /^min-/i.test(featureName)
 							? [TokenType.Delim, '>', -1, -1, { value: '>' }]
 							: [TokenType.Delim, '<', -1, -1, { value: '<' }];
 
-						parent.feature = new MediaFeatureRangeNameValue(
+						entry.parent.feature = new MediaFeatureRangeNameValue(
 							new MediaFeatureName(
 								new TokenNode([
 									TokenType.Ident,
@@ -80,20 +80,26 @@ const rule = (primary, _secondaryOptions, context) => {
 									-1,
 									{ value: unprefixedMediaFeature },
 								]),
-								node.name.before,
-								node.name.after.length > 0
-									? node.name.after
+								entry.node.name.before,
+								entry.node.name.after.length > 0
+									? entry.node.name.after
 									: [[TokenType.Whitespace, ' ', -1, -1, undefined]],
 							),
 							[operator, [TokenType.Delim, '=', -1, -1, { value: '=' }]],
-							node.value,
+							entry.node.value,
 						);
 
-						return;
-					}
+						const expectedMediaQueryList = mediaQueryList.map((mq) => mq.toString()).join(',');
 
+						if (expectedMediaQueryList === atRule.params) return;
+
+						atRule.params = expectedMediaQueryList;
+					};
+
+					const hasFix =
+						primary === 'context' && isMediaFeaturePlain(node) && isMediaFeature(parent);
+					const fix = hasFix ? contextFixer({ node, parent }) : undefined;
 					const [startIndex, endIndex] = sourceIndices(node);
-
 					const atRuleIndex = atRuleParamIndex(atRule);
 
 					report({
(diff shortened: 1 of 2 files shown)


lib/rules/media-feature-range-notation/index.mjs
```diff
@@ -80,20 +80,26 @@ const rule = (primary, _secondaryOptions, context) => {
...
 									: [[TokenType.Whitespace, ' ', -1, -1, undefined]],
 							),
 							[operator, [TokenType.Delim, '=', -1, -1, { value: '=' }]],
-							node.value,
+							entry.node.value,
 						);
 
-						return;
-					}
+						const expectedMediaQueryList = mediaQueryList.map((mq) => mq.toString()).join(',');
```
ybiquitous, line 92: [ask] This PR increases `mq.toString()` in the loop, but I'm not sure about the effects on performance.

@romainmenke Can you please let us know if you have any concerns? 🙏🏼

ybiquitous (commented):
Thank you. Overall LGTM, but I have one concern.

lib/rules/media-feature-range-notation/index.mjs
romainmenke: Serialization for these objects isn't fast, also not slow.

It's a bunch of recursive calls to `foo.toString()` where eventually tokens are found which each have string representations. All those bits of strings are concatenated.

So it's slow-ish because it can be a lot of function calls.
But it's fast-ish because the operations are always simple.

------

Does this change really increase the number of calls?
I thought it was only moved around a bit 🤔

ybiquitous: Thanks for the answer. It sounds like no significant issue. 👍🏼 

> Does this change really increase the number of calls?

If I understand correctly, this PR moves `mediaQueryList.map((mediaQuery) => mediaQuery.toString())` from outside to inside the `mediaQueryList.forEach()` loop. So, the fix count increases by the loop count.

https://github.com/stylelint/stylelint/blob/d1d11f2c9556b9274e900188df1416a161ba9ddf/lib/rules/media-feature-range-notation/index.mjs#L45

But if the loop count is a few, this may not be a problem. 😅

lib/rules/media-feature-range-notation/index.mjs
```diff
@@ -80,20 +80,26 @@ const rule = (primary, _secondaryOptions, context) => {
...
 
-					const [startIndex, endIndex] = sourceIndices(node);
+						if (expectedMediaQueryList === atRule.params) return;
+
+						atRule.params = expectedMediaQueryList;
+					};
 
+					const hasFix =
+						primary === 'context' && isMediaFeaturePlain(node) && isMediaFeature(parent);
+					const fix = hasFix ? contextFixer({ node, parent }) : undefined;
```
ybiquitous, line 101: [suggest] Super minor. `mayFix` (like a variable in `report()`) seems a bit more accurate to me. It may be just an issue of preference, though.

```suggestion
					const mayFix =
						primary === 'context' && isMediaFeaturePlain(node) && isMediaFeature(parent);
					const fix = mayFix ? contextFixer({ node, parent }) : undefined;
```

ybiquitous (approved):
Thank you, LGTM 👍🏼 You can ignore my review comments.

lib/rules/media-feature-range-notation/index.mjs
romainmenke: Oh, wow, this was completely invisible to me in the git diff 😅 

------

https://github.com/stylelint/stylelint/blob/d1d11f2c9556b9274e900188df1416a161ba9ddf/lib/rules/media-feature-range-notation/index.mjs#L45

vs. 

https://github.com/stylelint/stylelint/blob/main/lib/rules/media-feature-range-notation/index.mjs

The larger issue is that multiple reports are possible. If this is a complex query it can contain multiple warnings.

This rule assumes that fixes and warnings are separate branches and that there is no overlap.

I need to think on how we can handle this better.
But we shouldn't merge this as is.

I will also try to formulate a failing test.

Mouvedia: > But we shouldn't merge this as is.
> I will also try to formulate a failing test.

Please do.

Mouvedia:
> You can ignore my review comments.

Even https://github.com/stylelint/stylelint/pull/7802#discussion_r1651086393?

Mouvedia:
> Need to investigate

Let's postpone the release until this is resolved.

lib/rules/media-feature-range-notation/index.mjs
romainmenke:
```js
{
	code: `@media (min-width: 1px), (max-width: 2px) and (min-height: 3px) {}`,
	fixed: `@media (width >= 1px), (width <= 2px) and (height >= 3px) {}`,
	description: 'two range type media features in prefix notation',
	warnings: [
		{ message: messages.expected('context'), line: 1, column: 8, endLine: 1, endColumn: 24 },
		{ message: messages.expected('context'), line: 1, column: 26, endLine: 1, endColumn: 42 },
		{ message: messages.expected('context'), line: 1, column: 47, endLine: 1, endColumn: 64 },
	],
},
```

Before we started this effort the fix and warning execution branches were completely separate and fixes could do anything they needed to fix the source.

The final goal however is something somewhat distant from the current state.
We want fixes to be independent operations that do not have side effects or bleed.

In the above example we do have side effects and bleed.

After the first fix the AST will have changed and serialized it will look like:

```css
(width >= 1px), (max-width: 2px) and (min-height: 3px)
```

So the second fix will start from that point and won't be independent.
It will produce:

```css
(width >= 1px), (width <= 2px) and (min-height: 3px)
```

While it should instead produce:

```css
(min-width: 1px), (width <= 2px) and (min-height: 3px)
```

This however is not observable at this time because we don't yet return ranges or replacement texts. The end result will not be wrong. We will have a few too many mutations and serializations, but we will never have a hard bug.

I will open a separate issue for this because I don't want to block the progress here while I still want to formulate general guidance for this challenge. I think it will be most prominent with the resolved/flattened selectors for nesting.

ybiquitous: @romainmenke Thank you for the clear description with the actual example. I now realize this problem that you're concerned about, and I agree that we can discuss it in the follow-up issue. 👍🏼

Mouvedia: > we don't yet return ranges or replacement texts

This means that the way we will test will require to test only the range returned in correlation with the text.
i.e.
```
[
//...
  {
    range: { start: { line: 1, column: 27 }, end: { line: 1, column: 39 } },
    text: 'width <= 2',
  }
/...
]
```

Mouvedia:
> I will open a separate issue for this because I don't want to block the progress here while I still want to formulate general guidance for this challenge.

@romainmenke is this PR good to go?

Mouvedia:
> I will open a separate issue

@romainmenke If this is a queuing system, it could be interesting.
i.e. array of functions
But I would expect a `fix: ''` per warning in the tests in that case.
i.e. snapshots between fixes 
There are probably one or two core rules that might be easier to refactor with the help of such a feature.

Mouvedia:
offtopic

We should document on the READMEs which option has a fix instead of stating "fix _some_ of the problems reported".
Another example would be `color-function-notation`.
The reason being that it would probably spark the interest of contributors to implement the counterpart(s).
If a fix for a certain option is just not doable, it is also worth mentioning.

ybiquitous:
@Mouvedia Documenting autofix limitations sounds good. Feel free to open a pull request. 👍🏼

PR #7803, opened 2024-06-24
Refactor to use JSDoc `@import` tag in some files

ybiquitous:

Ref #7797

This introduces the JSDoc `@import` tag since TypeScript 5.5, see below:
https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#the-jsdoc-import-tag

Note that this replaces only `lib/*.mjs` files because there are so many files to be refactored.
We can gradually change other files.

```sh-session
$ git grep '@typedef \{import' lib/*.mjs || echo '(no match)'
(no match)
```


romainmenke (approved):
Thank you for this!
This is much simpler to read and write :)

PR #7806, opened 2024-06-27
Refactor `length-zero-no-unit` to use `fix` callback instead of `context.fix`

Mouvedia:

#2643

N/A

### DIFF
--- a/lib/rules/length-zero-no-unit/index.mjs
+++ b/lib/rules/length-zero-no-unit/index.mjs
@@ -84,7 +84,7 @@ const rule = (primary, secondaryOptions) => {
 			if (isFraction(unit)) return;
 
 			const fix = () => {
-				valueNode.value = number[0] === '.' ? number.slice(1) : number;
+				valueNode.value = number.startsWith('.') ? number.slice(1) : number;
 
 				switch (node.type) {
 					case 'atrule':
--- a/lib/rules/length-zero-no-unit/index.mjs
+++ b/lib/rules/length-zero-no-unit/index.mjs
@@ -28,8 +28,11 @@ const meta = {
 	fixable: true,
 };
 
+/** @import { AtRule, Declaration } from 'postcss' */
+/** @import { ParsedValue, Node } from 'postcss-value-parser' */
+
 /** @type {import('stylelint').Rule} */
-const rule = (primary, secondaryOptions, context) => {
+const rule = (primary, secondaryOptions) => {
 	return (root, result) => {
 		const validOptions = validateOptions(
 			result,
@@ -49,14 +52,13 @@ const rule = (primary, secondaryOptions, context) => {
 
 		if (!validOptions) return;
 
-		let needsFix;
-
 		/**
-		 * @param {import('postcss').Node} node
+		 * @param {AtRule | Declaration} node
 		 * @param {number} nodeIndex
-		 * @param {import('postcss-value-parser').Node} valueNode
+		 * @param {Node} valueNode
+		 * @param {ParsedValue} parsedValue
 		 */
-		function check(node, nodeIndex, valueNode) {
+		function check(node, nodeIndex, valueNode, parsedValue) {
 			const { value, sourceIndex } = valueNode;
 
 			if (isMathFunction(valueNode)) return false;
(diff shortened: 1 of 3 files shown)


lib/rules/length-zero-no-unit/index.mjs
```diff
@@ -49,14 +52,13 @@ const rule = (primary, secondaryOptions, context) => {
...
 		/**
-		 * @param {import('postcss').Node} node
+		 * @param {AtRule | Declaration} node
 		 * @param {number} nodeIndex
-		 * @param {import('postcss-value-parser').Node} valueNode
+		 * @param {Node} valueNode
+		 * @param {ParsedValue} parsedValue
 		 */
-		function check(node, nodeIndex, valueNode) {
+		function check(node, nodeIndex, valueNode, parsedValue) {
```
Mouvedia, line 61: I don't like having functions with more than 2 arguments but it seemed like the most straightforward refactor.
I am trying to limit the diff to the minimum.

lib/rules/length-zero-no-unit/index.mjs
```diff
@@ -206,6 +194,13 @@ function isZero(number) {
 	return Number.parseFloat(number) === 0;
 }
 
+/**
+ * @param {string} unit
+ */
+function isNumber(unit) {
```
Mouvedia, line 200: This follows the convention in this file.

lib/utils/__tests__/checkAgainstRule.test.mjs
```diff
@@ -86,24 +86,25 @@ describe('checkAgainstRule', () => {
...
+		const root = postcss.parse('a { font-weight: bold; }');
 		const context = { fix: true };
 
 		const warnings = [];
 
 		await checkAgainstRule(
 			{
-				ruleName: 'length-zero-no-unit',
-				ruleSettings: true,
+				ruleName: 'font-weight-notation',
```
Mouvedia, line 98: This rule will probably be the last one fixed.
It's neither included in recommended nor standard.

lib/rules/length-zero-no-unit/index.mjs
```diff
@@ -66,32 +68,33 @@ const rule = (primary, secondaryOptions, context) => {
...
 			if (!isLength(unit)) return;
 
 			if (isFraction(unit)) return;
 
-			if (!isZero(number)) return;
-
-			if (context.fix) {
-				let regularNumber = number;
+			const fix = () => {
+				valueNode.value = number[0] === '.' ? number.slice(1) : number;
```
ybiquitous, line 87: [suggest] I prefer `startsWith` over `[0]` because of readability, like natural English, as the original code does:

```suggestion
				valueNode.value = number.startsWith('.') ? number.slice(1) : number;
```

lib/rules/length-zero-no-unit/index.mjs
ybiquitous: [suggest] I think `isNumber` is confusing with `isNumber` in `validateTypes.mjs`. How about this?

```suggestion
function isNumberUnit(unit) {
```

Of course, I'm also okay with renaming to `isLengthUnit` and `isFractionUnit` for consistency.

ybiquitous (commented):
Thanks, overall LGTM. I've left a few refactoring suggestions.

lib/rules/length-zero-no-unit/index.mjs
Mouvedia: Just curious, how long does the `npm run build` take on your machine?

lib/rules/length-zero-no-unit/index.mjs
Mouvedia: It's named `isNumber` in 2 other rules so I think it's fine as is.
i.e. `hue-degree-notation` and `lightness-notation`

lib/rules/length-zero-no-unit/index.mjs
ybiquitous: My machine is MacBook with Apple M1. Here's the `time` command result. I feel about less than 10 seconds.

```sh-session
$ time npm run build

npm run build  7.43s user 5.82s system 221% cpu 5.985 total
```

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #7808, opened 2024-06-27
Fix `shorthand-property-no-redundant-values` false negatives for logical properties, `overflow`, `overscroll-behavior`, `scroll-margin` and `scroll-padding`

Mouvedia:

#7396

I wanted to separate this fix from the upcoming refactor.

> see also

https://caniuse.com/css-logical-props
https://www.w3.org/TR/css-scroll-snap-1/

### DIFF
--- a/.changeset/tough-bottles-play.md
+++ b/.changeset/tough-bottles-play.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `shorthand-property-no-redundant-values` false negatives for logical properties, `overflow`, `overscroll-behavior`, `scroll-margin` and `scroll-padding`
--- a/.changeset/tough-bottles-play.md
+++ b/.changeset/tough-bottles-play.md
@@ -1,5 +1,5 @@
 ---
-"stylelint": patch
+"stylelint": minor
 ---
 
 Fixed: `shorthand-property-no-redundant-values` false negatives for logical properties, `overflow`, `overscroll-behavior`, `scroll-margin` and `scroll-padding`
--- a/lib/reference/properties.mjs
+++ b/lib/reference/properties.mjs
@@ -610,6 +610,38 @@ export const longhandSubPropertiesOfShorthandProperties = new Map([
 	],
 ]);
 
+export const hasCollapsibleValues = new Set([
+	'margin',
+	'margin-block',
+	'margin-inline',
+	'padding',
+	'padding-block',
+	'padding-inline',
+	'border-color',
+	'border-style',
+	'border-width',
+	'border-radius',
+	'border-block-color',
+	'border-block-style',
+	'border-block-width',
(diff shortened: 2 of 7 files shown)


lib/rules/shorthand-property-no-redundant-values/index.cjs
```diff
@@ -23,23 +24,12 @@ const meta = {
...
 };
 
-const SUPPORTED_SHORTHANDS = new Set([
-	'margin',
-	'padding',
-	'border-color',
-	'border-radius',
-	'border-style',
-	'border-width',
-	'grid-gap',
```
Mouvedia, line 33: https://developer.mozilla.org/en-US/docs/Web/CSS/gap

lib/reference/properties.mjs
```diff
@@ -610,6 +610,38 @@ export const longhandSubPropertiesOfShorthandProperties = new Map([
 	],
 ]);
 
+export const hasCollapsibleValues = new Set([
```
ybiquitous, line 613: [ask] `has***` seems to return a boolean. And, `***Properties` seems more consistent than `***Values` in this module. Do you have any other better name?

.changeset/tough-bottles-play.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
```
ybiquitous, line 2: [suggest] Fixing false negatives will produce more errors.

```suggestion
"stylelint": minor
```

ybiquitous (commented):
Thanks, overall LGTM. Can you take a look at my comments?

lib/reference/properties.mjs
Mouvedia: > Do you have any other better name?

I am a hard time finding something better.
`import { hasCollapsibleValues } from '../../reference/properties.mjs';` looks OK to me.
Since it's not being used by another rule yet, should I inline the set directly in the file?
i.e. `const SUPPORTED_SHORTHANDS = new Set([…`

ybiquitous (approved):
Inlining the constant is good. LGTM 👍🏼

lib/rules/shorthand-property-no-redundant-values/index.cjs
```diff
@@ -23,15 +23,20 @@ const meta = {
...
+	'padding', 'padding-block', 'padding-inline',
+	'border-color', 'border-style', 'border-width',
 	'border-radius',
-	'border-style',
-	'border-width',
-	'grid-gap',
-	'inset',
+	'border-block-color', 'border-block-style', 'border-block-width',
+	'border-inline-color', 'border-inline-style', 'border-inline-width',
+	'gap', 'grid-gap',
```
Mouvedia, line 34: https://caniuse.com/mdn-css_properties_gap

PR #7817, opened 2024-07-03
Require `meta.fixable` to be truthy when using `fix` callback for `report()`

ybiquitous:

Related to #7784

This change aims to prevent inconsistency between `fixable` metadata and `fix` callback.

> [!NOTE]
> The `fix` callback is not published yet, so this PR doesn't need a changelog entry.

### DIFF
--- a/lib/utils/__tests__/report.test.mjs
+++ b/lib/utils/__tests__/report.test.mjs
@@ -411,6 +411,7 @@ describe('with fix callback', () => {
 			result: {
 				warn: jest.fn(),
 				stylelint: {
+					ruleMetadata: { foo: { fixable: true } },
 					ruleSeverities: {},
 					fixersData,
 					config: {
@@ -451,6 +452,7 @@ describe('with fix callback', () => {
 			result: {
 				warn: jest.fn(),
 				stylelint: {
+					ruleMetadata: { foo: { fixable: true } },
 					ruleSeverities: {},
 					fixersData,
 					config: {
@@ -473,6 +475,25 @@ describe('with fix callback', () => {
 
 		expect(fixerData.fixed).toBe(false);
 		expect(fixerData.range).toBeUndefined();
+	});
+
+	test('failing due to non-fixable metadata', () => {
+		const v = {
+			ruleName: 'foo',
+			result: {
+				stylelint: {
+					ruleMetadata: { foo: { fixable: false } },
+					ruleSeverities: {},
+					fixersData: {},
+				},
+			},
+			node: {
+				rangeBy: defaultRangeBy,
+			},
+			fix: () => {},
+		};
+
+		expect(() => report(v)).toThrow('The "foo" rule requires "fixable=true" metadata for autofix');
 	});
 });
 
--- a/lib/utils/__tests__/report.test.mjs
+++ b/lib/utils/__tests__/report.test.mjs
@@ -493,7 +493,9 @@ describe('with fix callback', () => {
 			fix: () => {},
 		};
 
-		expect(() => report(v)).toThrow('The "foo" rule requires "fixable=true" metadata for autofix');
+		expect(() => report(v)).toThrow(
+			'The "foo" rule requires "meta.fixable" to be truthy if the "fix" callback is being passed',
+		);
 	});
 });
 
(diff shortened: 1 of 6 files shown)


lib/utils/__tests__/report.test.mjs
```diff
@@ -474,6 +476,25 @@ describe('with fix callback', () => {
...
+					fixersData: {},
+				},
+			},
+			node: {
+				rangeBy: defaultRangeBy,
+			},
+			fix: () => {},
+		};
+
+		expect(() => report(v)).toThrow('The "foo" rule requires "fixable=true" metadata for autofix');
```
Mouvedia, line 496:
```suggestion
expect(() => report(v)).toThrow('The "foo" rule requires "fixable" to be truthy if the "fix" callback is being passed');
```

lib/utils/report.cjs
```diff
@@ -20,18 +20,23 @@ const constants = require('../constants.cjs');
...
 		customMessages: { [ruleName]: message = rest.message } = {},
 		customUrls: { [ruleName]: customUrl } = {},
+		ruleMetadata: { [ruleName]: metadata } = {},
 	} = result.stylelint;
 	const { messageArgs = [], severity = ruleSeverities[ruleName] } = rest;
 	const ruleSeverity = validateTypes.isFunction(severity) ? severity(...messageArgs) || defaultSeverity : severity;
 
+	if (validateTypes.isFunction(fix) && metadata && !metadata.fixable) {
+		throw new Error(`The "${ruleName}" rule requires "fixable=true" metadata for autofix`);
+	}
```
Mouvedia, line 38: nit

You want this to be checked after line 41.
I don't know why someone would depend on it but that was the legacy behaviour.
i.e. if `(quiet && ruleSeverity !== 'error')` is truthy you wouldn't have errors (e.g. the startLine condition that throws is placed after)

lib/utils/__tests__/report.test.mjs
ybiquitous: Sounds good. Changed via 1b40515c5 👍🏼 

I've changed `fixable` to `meta.fixable` since the doc has used it already:
https://github.com/stylelint/stylelint/blob/cb3a7f8d5e89f2e467a4166293c17d3c2d1e72cc/docs/developer-guide/rules.md?plain=1#L192

lib/utils/report.cjs
ybiquitous: That makes sense. Changed via 1b40515c5. 👍🏼

lib/utils/report.mjs
```diff
@@ -16,21 +16,28 @@ import { RULE_NAME_ALL } from '../constants.mjs';
...
 		customUrls: { [ruleName]: customUrl } = {},
+		ruleMetadata: { [ruleName]: metadata } = {},
 	} = result.stylelint;
 	const { messageArgs = [], severity = ruleSeverities[ruleName] } = rest;
 	const ruleSeverity = isFn(severity) ? severity(...messageArgs) || defaultSeverity : severity;
 
 	// In quiet mode, mere warnings are ignored
 	if (quiet && ruleSeverity !== 'error') return;
 
+	if (isFn(fix) && metadata && !metadata.fixable) {
```
Mouvedia, line 35: question

If `metadata` is missing, this would never `throw`?

Mouvedia (approved):

lib/utils/report.mjs
ybiquitous: Right. I assume plugins without metadata.

ybiquitous: However, it may be good to require metadata itself when using `fix()` 🤔

ybiquitous: But unfortunately, the following change fails the `checkAgainstRule()` test:

```diff
-	if (isFn(fix) && metadata && !metadata.fixable) {
+	if (isFn(fix) && !metadata?.fixable) {
```

```sh-session
$ npm -s run test-only -- lib/utils/__tests__/checkAgainstRule.test.mjs
...
  ● checkAgainstRule › handles array rule settings

    The "at-rule-empty-line-before" rule requires "meta.fixable" to be truthy if the "fix" callback is being passed
...
```

Because there are no metadata loaded in `checkAgainstRule()`:

https://github.com/stylelint/stylelint/blob/cb3a7f8d5e89f2e467a4166293c17d3c2d1e72cc/lib/utils/checkAgainstRule.mjs#L47

Issue #4057, opened 2019-04-26
Fix false positives for escaped multi-line in `string-no-newline`


AlexWayfer:
> Clearly describe the bug

I want to split large strings for aesthetic reasons with enabled `string-no-newline` rule.

There is the quote from [the specs](https://www.w3.org/TR/CSS2/syndata.html#strings), including:

> And also: "It is possible to break strings over several lines, for aesthetic or other reasons, but in such a case the newline itself has to be escaped with a backslash (\)."

But for some reason escaped newlines are forbidden in Stylelint.

> Which rule, if any, is the bug related to?

`string-no-newline`

> What CSS is needed to reproduce the bug?

```css
background-image: url(
  "data:image/svg+xml;utf8,\
  <svg fill='black' height='24' \
    viewBox='0 0 24 24' width='24' \
    xmlns='http://www.w3.org/2000/svg'\
  ><path d='M7 9l5 5 5-5z'/></svg>"
);
```

> What stylelint configuration is needed to reproduce the bug?

```yaml
extends: stylelint-config-standard
```

> Which version of stylelint are you using?

`9.10.1`

> How are you running stylelint: CLI, PostCSS plugin, Node.js API?

CLI with `stylelint assets/styles/`.

> Does the bug relate to non-standard syntax (e.g. SCSS, Less etc.)?

No.

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors did you get)?

The following warnings were flagged:

```shell
 116:197  ✖  Expected line length to be no more than 80 characters                                               max-line-length
```

hudochenkov:
Thanks for the report and for using the template.

I can't reproduce this error with the following code:

```css
.hello {
  background-image: url(
    "data:image/svg+xml;utf8,\
    <svg fill='black' height='24' \
      viewBox='0 0 24 24' width='24' \
      xmlns='http://www.w3.org/2000/svg'\
    ><path d='M7 9l5 5 5-5z'/></svg>"
  );
}
```

Looks like you're also has `max-line-length: 80` in your config. I can't reproduce error with this rule enabled as well.

Could you check please what is actual CSS, which triggers warning? Maybe there are many indentation symbols, and line is actually bigger than 80 symbols.

hudochenkov:
Looks like our demo is misleading, it doesn't show problem, when CLI shows.

For the following CSS:

```css
.hello {
  background-image: url("data:image/svg+xml;utf8,\
[...]

PR #7818, opened 2024-07-03
Fix `string-no-newline` false positives for escaped multi-line

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/4057

No, it's self-explanatory.

### DIFF
--- a/lib/rules/string-no-newline/__tests__/index.mjs
+++ b/lib/rules/string-no-newline/__tests__/index.mjs
@@ -13,12 +13,16 @@ testRule({
 			description: 'without newline',
 		},
 		{
-			code: "a::before { content: 'one\\ntwo'; }",
+			code: "a::before { content: 'one\\\ntwo'; }",
 			description: 'with escaped slash-slash-n newline',
 		},
 		{
 			code: "a::before { content: 'one\\Atwo'; }",
 			description: 'with escaped slash-A newline',
+		},
+		{
+			code: "a[href^='one\\\r\ntwo'] { display: block; }",
+			description: 'attribute CRLF',
 		},
 		{
 			code: stripIndent`
@@ -46,8 +50,8 @@ testRule({
 			message: messages.rejected,
 			line: 1,
 			column: 22,
-			endLine: 2,
-			endColumn: 5,
+			endLine: 1,
+			endColumn: 26,
 		},
 		{
 			code: "a::before { content: 'one\r\ntwo'; }",
@@ -55,8 +59,8 @@ testRule({
 			message: messages.rejected,
 			line: 1,
 			column: 22,
-			endLine: 2,
-			endColumn: 5,
+			endLine: 1,
+			endColumn: 26,
 		},
 		{
 			code: "a[href^='one\r\ntwo'] { display: block; }",
@@ -64,8 +68,8 @@ testRule({
 			message: messages.rejected,
 			line: 1,
 			column: 9,
-			endLine: 2,
-			endColumn: 5,
+			endLine: 1,
+			endColumn: 13,
 		},
 		{
 			code: "@charset 'utf-8\n';",
@@ -73,8 +77,8 @@ testRule({
 			message: messages.rejected,
 			line: 1,
 			column: 10,
-			endLine: 2,
-			endColumn: 2,
+			endLine: 1,
+			endColumn: 16,
 		},
 	],
 });
--- a/lib/rules/string-no-newline/index.mjs
+++ b/lib/rules/string-no-newline/index.mjs
@@ -1,4 +1,4 @@
-import { isTokenBadString, tokenize } from '@csstools/css-tokenizer';
+import { ParseErrorMessage, ParseErrorWithToken, tokenize } from '@csstools/css-tokenizer';
 
 import atRuleParamIndex from '../../utils/atRuleParamIndex.mjs';
 import declarationValueIndex from '../../utils/declarationValueIndex.mjs';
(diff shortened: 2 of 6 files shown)


lib/rules/string-no-newline/__tests__/index.mjs
```diff
@@ -13,13 +13,17 @@ testRule({
 			description: 'without newline',
 		},
 		{
-			code: "a::before { content: 'one\\ntwo'; }",
+			code: "a::before { content: 'one\\\ntwo'; }",
```
romainmenke, line 16: This is the correct representation of `backslash + newline` in CSS :

```css
"foo \
bar"
```

With only a single `\` and `\n` it becomes an escaped newline, not a backslash followed by a newline.

lib/rules/string-no-newline/__tests__/index.mjs
```diff
@@ -46,35 +50,35 @@ testRule({
 			message: messages.rejected,
 			line: 1,
 			column: 22,
-			endLine: 2,
-			endColumn: 5,
+			endLine: 1,
+			endColumn: 26,
```
romainmenke, line 54: `bad-string-token` stops at the end of the line.
Before it would scan ahead to a subsequent quote mark.

package.json
```diff
@@ -159,7 +159,7 @@
   },
   "dependencies": {
     "@csstools/css-parser-algorithms": "^2.7.0",
-    "@csstools/css-tokenizer": "^2.3.2",
+    "@csstools/css-tokenizer": "^2.3.3",
```
romainmenke, line 162: see: https://github.com/csstools/postcss-plugins/blob/main/packages/css-tokenizer/CHANGELOG.md#233

> Fix tokenization of `string-token` containing a backslash followed by CRLF

lib/rules/string-no-newline/index.mjs
```diff
@@ -56,67 +55,54 @@ const rule = (primary) => {
...
+			/** @type {Array<ParseError>} */
+			let errors = [];
+			const tokens = tokenize(
+				{ css: value },
+				{
+					onParseError: (/** @type {ParseError} */ err) => {
+						if (err.message === 'Unexpected newline while consuming a string token.') {
+							errors.push(err);
+						}
+					},
```
romainmenke, line 84: This rule happens to align perfectly with a specified parse error: https://drafts.csswg.org/css-syntax/#consume-string-token

> newline
> This is a parse error. Reconsume the current input code point, create a `<bad-string-token>`, and return it.

lib/rules/string-no-newline/index.mjs
```diff
@@ -56,67 +55,54 @@ const rule = (primary) => {
...
+
+				const [, , start, end] = token;
 
-				if (!reNewLine.test(valueNode.value)) {
-					return;
-				}
+				const error = errors.find((err) => err.sourceStart === start);
 
-				const nodeIndex = getIndex(node);
+				if (!error) return;
```
romainmenke, line 99: Theoretically there could be future changes to the CSS syntax with more conditions that result in `bad-string-token`.

Currently there is only one, which is the parse error for incorrect newlines.

This makes it impossible to have coverage for this condition.
It would still like to keep it for forwards compat.

lib/rules/string-no-newline/index.mjs
```diff
@@ -56,67 +55,54 @@ const rule = (primary) => {
...
+
+			tokens.forEach((token) => {
+				if (!isTokenBadString(token)) return;
+
+				const [, , start, end] = token;
 
-				if (!reNewLine.test(valueNode.value)) {
-					return;
-				}
+				const error = errors.find((err) => err.sourceStart === start);
```
ybiquitous, line 97: [suggest] If the `error` object is used only in the `if` condition, we might replace `find()` with `some()`:

```suggestion
				if (!errors.some((err) => err.sourceStart === start)) return;
```

Furthermore, replacing `errors` with `errorStartIndexes` might make the condition a bit simpler, for example:

```js
if (err instanceof ParseError) {
	errorStartIndexes.push(err.sourceStart);
}

// ...

if (!errorStartIndexes.includes(start)) return;
```

ybiquitous (changes requested):
@romainmenke Thank you for creating the pull request with the new version of `@csstools/css-tokenizer`!

I've left a few refactoring suggestions, but this PR overall LGTM 👍🏼

lib/rules/string-no-newline/index.mjs
```diff
@@ -56,67 +55,54 @@ const rule = (primary) => {
...
-					return;
-				}
+			/** @typedef {import('@csstools/css-tokenizer').ParseError} ParseError */
+			/** @type {Array<ParseError>} */
+			let errors = [];
+			const tokens = tokenize(
+				{ css: value },
+				{
+					onParseError: (/** @type {ParseError} */ err) => {
+						if (err.message === 'Unexpected newline while consuming a string token.') {
```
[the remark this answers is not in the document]
romainmenke: 🤔 It can only be a `ParseError` :

```ts
export function tokenize(
	input: {
		css: { valueOf(): string },
		unicodeRangesAllowed?: boolean
	},
	options?: {
		onParseError?: (error: ParseError) => void
	},
): Array<CSSToken>
```

Checking the string is how we can check that the problem is a newline in a string and not for example a bad url token.

-----

I could provide exported constants for these errors in `@csstools/css-tokenizer`.
That would be easier.

I am not sure what I can do about the type signature so that TypeScript correctly determines that `err` is always of type `ParseError` in this callback.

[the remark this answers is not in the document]
ybiquitous: > I could provide exported constants for these errors in `@csstools/css-tokenizer`.
> That would be easier.

That sounds better 👍🏼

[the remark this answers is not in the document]
romainmenke: I've added a new `ParseErrorWithToken` subclass and a `ParseErrorMessage` object holding all the possible error messages.

The `ParseErrorWithToken` (as the name implies) is a parse error that holds a reference to the token that caused the error.

With this we have the error and token in one place without extra loops.

romainmenke:
Thank you for the feedback @ybiquitous,

I've update `@csstools/css-tokenizer` to expose a nice API for checking and handling parse errors. This simplifies the amount of code and logic that is needed here.

Please let me know what you think.

lib/rules/string-no-newline/index.mjs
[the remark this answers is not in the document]
ybiquitous: Thanks. I love this change. 👍🏼

ybiquitous (approved):
@romainmenke Thanks a lot for publishing `@csstools/css-tokenizer` at a few times to fix this bug. The code is now much simpler. LGTM! 👍🏼

Mouvedia:
@romainmenke thanks for fixing a 5+yo bug.

PR #7819, opened 2024-07-03
Do not return the range of the replaced text

Mouvedia:

#7784

None of the proposals that are being retained have `fix` returning the range of the replaced text.
The return type of `fix` doesn't have to be changed for now.


romainmenke (approved):
Thank you for this @Mouvedia 🙇 

I think it's good to clean this up at this time.

Maybe @ybiquitous also has thoughts on this?

ybiquitous:
Thanks for the pull request. I'm okay with this change for now.

In addition, I suggest updating the type, too:

```diff
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -634,7 +634,7 @@ declare namespace stylelint {
 		 * @internal
 		 * WARNING: Don't use this feature now. It may change in the future.
 		 */
-		fix?: () => void | never | Range;
+		fix?: () => void | undefined;
 	};
 
 	/** @internal */
```

PR #7822, opened 2024-07-04
Refactor `font-weight-notation` to use `fix` callback instead of `context.fix`

Mouvedia:

#2643

N/A

### DIFF
--- a/lib/rules/font-weight-notation/index.mjs
+++ b/lib/rules/font-weight-notation/index.mjs
@@ -40,7 +40,7 @@ const NUMERIC_TO_NAMED = new Map([
 ]);
 
 /** @type {import('stylelint').Rule<'numeric' | 'named-where-possible'>} */
-const rule = (primary, secondaryOptions, context) => {
+const rule = (primary, secondaryOptions) => {
 	return (root, result) => {
 		const validOptions = validateOptions(
 			result,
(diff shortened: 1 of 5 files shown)


lib/rules/font-weight-notation/index.mjs
```diff
@@ -84,26 +82,24 @@ const rule = (primary, secondaryOptions, context) => {
...
+					if (checkWeight(decl, valueNode, parsedValue)) {
 						break; // Stop traverse if font-weight is processed
 					}
 				}
 
-				checkWeight(decl, valueNode);
-			}
-
-			if (context.fix) {
-				// Autofix after the loop ends can prevent value nodes from changing their positions during the loop.
```
Mouvedia, line 96: We don't have a test case that failed.
@ybiquitous do you have one?

lib/reference/keywords.mjs
```diff
@@ -103,11 +103,11 @@ export const fontWeightKeywords = uniteSets(
 
 const fontStyleKeywords = uniteSets(basicKeywords, ['normal', 'italic', 'oblique']);
 
+/** @todo add sub, super, text, emoji, unicode */
```
Mouvedia, line 106: This will be handled in a future issue/PR.
i.e. this is a refactor PR but I don't want this to be forgotten

lib/rules/font-weight-notation/index.mjs
ybiquitous: No. I guess there is no actual problem, so let's go on.

ybiquitous (approved):
Thank you. LGTM 👍🏼

PR #1433, opened 2016-06-12
New util `findAnimationName` and some fixes.

evilebottnawi:
/cc @davidtheclark @jeddy3

PR #7823, opened 2024-07-04
Update `fontSizeKeywords` set and fix `fontShorthandKeywords` set

Mouvedia:

N/A

This _might_ impact these rules:

- `font-family-name-quotes`
- `font-family-no-duplicate-names`
- `font-family-no-missing-generic-family-keyword`

https://github.com/stylelint/stylelint/blob/4ccec36357b9669f1ea5158c5fd2be10f434c1e8/lib/utils/findFontFamily.mjs#L75-L91

> References

https://developer.mozilla.org/en-US/docs/Web/CSS/font
https://developer.mozilla.org/en-US/docs/Web/CSS/font-size

### DIFF
--- a/.changeset/slow-timers-relax.md
+++ b/.changeset/slow-timers-relax.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Fixed: `font-family-no-missing-generic-family-keyword` false negatives for font families which names match a CSS3 `font-variant` keyword
--- a/lib/reference/keywords.cjs
+++ b/lib/reference/keywords.cjs
@@ -107,44 +107,7 @@ const fontWeightKeywords = uniteSets(
 
 const fontStyleKeywords = uniteSets(basicKeywords, ['normal', 'italic', 'oblique']);
 
-const fontVariantKeywords = uniteSets(basicKeywords, [
-	'normal',
-	'none',
-	'historical-forms',
-	'none',
-	'common-ligatures',
-	'no-common-ligatures',
-	'discretionary-ligatures',
-	'no-discretionary-ligatures',
-	'historical-ligatures',
-	'no-historical-ligatures',
-	'contextual',
-	'no-contextual',
-	'small-caps',
-	'small-caps',
-	'all-small-caps',
-	'petite-caps',
-	'all-petite-caps',
-	'unicase',
-	'titling-caps',
-	'lining-nums',
-	'oldstyle-nums',
-	'proportional-nums',
-	'tabular-nums',
-	'diagonal-fractions',
-	'stacked-fractions',
-	'ordinal',
-	'slashed-zero',
-	'jis78',
-	'jis83',
-	'jis90',
-	'jis04',
-	'simplified',
-	'traditional',
-	'full-width',
-	'proportional-width',
-	'ruby',
-]);
+const fontVariantCSS2Keywords = uniteSets(basicKeywords, ['normal', 'none', 'small-caps']);
 
 const fontStretchKeywords = uniteSets(basicKeywords, [
 	'semi-condensed',
@@ -165,8 +128,10 @@ const fontSizeKeywords = uniteSets(basicKeywords, [
 	'large',
 	'x-large',
 	'xx-large',
+	'xxx-large',
 	'larger',
 	'smaller',
+	'math',
 ]);
 
 const lineHeightKeywords = uniteSets(basicKeywords, ['normal']);
@@ -174,7 +139,7 @@ const lineHeightKeywords = uniteSets(basicKeywords, ['normal']);
 const fontShorthandKeywords = uniteSets(
 	basicKeywords,
 	fontStyleKeywords,
-	fontVariantKeywords,
+	fontVariantCSS2Keywords,
 	fontWeightKeywords,
 	fontStretchKeywords,
 	fontSizeKeywords,
--- a/lib/reference/keywords.mjs
+++ b/lib/reference/keywords.mjs
@@ -103,44 +103,7 @@ export const fontWeightKeywords = uniteSets(
 
 const fontStyleKeywords = uniteSets(basicKeywords, ['normal', 'italic', 'oblique']);
 
-const fontVariantKeywords = uniteSets(basicKeywords, [
-	'normal',
-	'none',
-	'historical-forms',
-	'none',
-	'common-ligatures',
-	'no-common-ligatures',
-	'discretionary-ligatures',
-	'no-discretionary-ligatures',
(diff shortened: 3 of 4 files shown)


ybiquitous:
Thanks for the pull request. The addition to `fontSizeKeywords` looks good. 👍🏼 

But I don't understand why `fontVariantKeywords` was removed. Can you explain?
It seems that [The MDN doc for `font-variant`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant) still describes removed keywords like `slashed-zero`.

---

> This *might* impact these rules:

If this PR fixes some bugs for these rules, can you add test cases to reproduce them?

Mouvedia:
> But I don't understand why `fontVariantKeywords` was removed. Can you explain? It seems that [The MDN doc for `font-variant`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant) still describes removed keywords like `slashed-zero`.

> The small-caps value is the only non-normal font variant valid within the [font](https://developer.mozilla.org/en-US/docs/Web/CSS/font) shorthand property.

> If this PR fixes some bugs for these rules, can you add test cases to reproduce them?

I didn't link any issue for the "Which issue, if any, is this issue related to?" question, so no.
The 3 conditions listed in [OP](https://github.com/stylelint/stylelint/pull/7823#issue-2391004467) will be affected.

ybiquitous:
> But I don't understand why `fontVariantKeywords` was removed. Can you explain? It seems that [The MDN doc for `font-variant`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant) still describes removed keywords like `slashed-zero`.

@Mouvedia Can you answer this question?

romainmenke:
> And, as I said earlier, font doesn't accept CSS3 values.
i.e. only `small-caps` and `normal` are valid

How do you mean?
Do you have a browser bug or some other reference?

Seems to work fine to me:
> CSS.supports('font: italic small-caps bold 16px/2 cursive');
> // true

----

Should we instead increase the test coverage for these keywords?
-> https://github.com/stylelint/stylelint/pull/7832

romainmenke:
After some more reading 😅 

These properties are relevant:
- `font` : https://developer.mozilla.org/en-US/docs/Web/CSS/font
- `font-variant` : https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant
- `font-variant-caps` : https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-caps

`font-variant-caps` and `font-variant` support values (e.g. `unicase`) that are not supported in `font`.

Only a specific set is supported in `font`:
- `normal` and `small-caps` (see: https://developer.mozilla.org/en-US/docs/Web/CSS/font#syntax)

So what @Mouvedia says is correct :)

ybiquitous:
@romainmenke Thank you for the explanation. I now understand what this PR wants to fix. 

---

@Mouvedia I believe this PR's title and body lack enough information. Is this PR a refactoring, a bug fix, or something else? Also, what is the purpose or background? I genuinely want you to understand that our OSS work is volunteer, and we don't have enough time. Honestly, it seems to me that your behavior lacks respect a little.

ybiquitous (dismissed):

Mouvedia:
> Is this PR a refactoring, a bug fix, or something else?

- `fontShorthandKeywords`

It's something that should have been detected in the PR that added the set.
i.e. https://github.com/stylelint/stylelint/pull/1433/files#diff-80e963d0841f584ac75f640a1db78410a3cecfe84bc6cff14a2b58e7b7d6a299R450
So to answer your question I would say it's an old oversight that no one corrected until now.

- `fontSizeKeywords`

Chromium had `-webkit-xxx-large` before https://issues.chromium.org/issues/41460591.
Since the set was added 3 years before the first implementation by a browser, this change should be considered a very late update.

https://github.com/stylelint/stylelint/commit/393a9f9539052315eca7d2f60b5e54ffbb87f6a6
https://caniuse.com/mdn-css_properties_font-size_xxx-large

Mouvedia:
> I genuinely want you to understand that our OSS work is volunteer, and we don't have enough time. Honestly, it seems to me that your behavior lacks respect a little.

My bad I genuinely though that the references that I put in the [OP](https://github.com/stylelint/stylelint/pull/7823#issue-2391004467) were enough to justify the change.
i.e. the answer to your question was in the reference, then I answered it directly and a second time with more details

ybiquitous:
Thanks for providing information. People could see the PR comments have enough information for the background.

To make sure, isn't this PR refactoring? If this is a bug fix, we should add a changelog entry and, hopefully, a few test cases to reproduce the bug. 🤔

Mouvedia:
> If this is a bug fix, we should add a changelog entry and, hopefully, a few test cases to reproduce the bug.

If you are not in a hurry, I can try to find a bug; gimme some time.
I am of the opinion that fixing a potential bug before someone encountered it and filled it is fine.

Mouvedia:
OK I have found bug.
It's highly unlikely to happen.
Here's a [demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYAdAdugLgMwkx3hQEYA2ABwA8V0QAnAVwCMBPegbiwF8sQANCDwBLADYwAckgC2cRDGqzKEgHRgAzhsHhCogOYIQGTCjqMmEjfRImzZ+gSIBaPLPFtnmCM5kitIpj6zvowmDAMImCu7mKeANYwbADuEAxQNig4zDBYZnyYPDqQmAYAYmkySDhGAFYahDqwlNqIduYaOGwSYoE4mfRi1TCd9AJ5HV09fc4lBs6dSJhQSOkDIEM4I-0gvCA8QA).

You have to `@font-face` load a font which name correspond to one of the CSS3 values supported by the `font-variant` property. And then use the `font` property without a generic-family fallback.
e.g. https://www.myfonts.com/fr/collections/ruby-font-sparkytype?tab=techSpecs
In that very rare scenario you would have false negative for `font-family-no-missing-generic-family-keyword`.

ybiquitous:
Thanks for providing the demo. Once this PR is released, the following lint problem will be reported:

```css
a {
  font: 16px "ruby";
/*           ^^^^^^ Unexpected missing generic font family (font-family-no-missing-generic-family-keyword) */
}
```

In this case, how do users fix the problem?

> Do you want me to add a changeset?

Yes, please.

Mouvedia:
> In this case, how do users fix the problem?

They add a generic-family fallback.
e.g. `font: 16px "ruby", sans-serif;`

> > Do you want me to add a changeset?
> 
> Yes, please.

OK.

ybiquitous (approved):

PR #7825, opened 2024-07-05
Refactor `function-url-quotes` to use `fix` callback instead of `context.fix`

Mouvedia:

#2643

This is the last refactor that Ill send.

@ybiquitous @romainmenke
I am leaving `function-calc-no-unspaced-operator` in your competent hands.
Looking at #7655, you are probably way more equipped to refactor that rule than me.


lib/rules/function-url-quotes/index.mjs
```diff
@@ -54,48 +58,83 @@ const rule = (primary, secondaryOptions, context) => {
...
+
+		/**
+		 * @param {AtRule|Declaration} node
+		 * @param {string} source
+		 * @param {number} startIndex
 		 */
+		function complain(node, source, startIndex) {
+			/** @type {ReportQueue} */
+			const reportQueue = [];
+			const parsedValue = functionArgumentsSearch(source, /^url$/i, (args, index, funcNode) => {
```
Mouvedia, line 99: You don't have access to `parsedValue` inside the callback.

ybiquitous: [suggest] How about adding `parsedValue` to the callback's arguments in `functionArgumentsSearch()` like this?

```diff
--- a/lib/utils/functionArgumentsSearch.mjs
+++ b/lib/utils/functionArgumentsSearch.mjs
@@ -15,7 +15,7 @@ import { assert, isRegExp, isString } from './validateTypes.mjs';
  *
  * @param {string} source
  * @param {string | RegExp} functionName
- * @param {(expression: string, expressionIndex: number, funcNode: valueParser.FunctionNode) => void} callback
+ * @param {(expression: string, expressionIndex: number, funcNode: valueParser.FunctionNode, parsedValue: valueParser.ParsedValue) => void} callback
  * @returns {valueParser.ParsedValue}
  */
 export default function functionArgumentsSearch(source, functionName, callback) {
```

lib/rules/function-url-quotes/index.mjs
```diff
@@ -54,48 +58,83 @@ const rule = (primary, secondaryOptions, context) => {
...
+				const problem = checkArgs(args, startIndex + index, funcNode);
+
+				if (problem) {
+					reportQueue.push([
+						problem,
+						[node, funcNode, problem.message.startsWith('Expected') ? 'add' : 'remove'],
```
ybiquitous, line 105: [suggest] I want to avoid depending on the message (`message.startsWith('Expected')`) for maintainability. Instead, how about adding the message type? For example:

```diff
--- a/lib/rules/function-url-quotes/index.mjs
+++ b/lib/rules/function-url-quotes/index.mjs
@@ -60,7 +60,7 @@ const rule = (primary, secondaryOptions) => {
 		/**
 		 * @param {Declaration|AtRule} node
 		 * @param {FunctionNode} fn
-		 * @param {'add'|'remove'} mutateType
+		 * @param {keyof messages} messageType
 		 * @param {ParsedValue} parsedValue
 		 */
 		function fixer(node, fn, mutateType, parsedValue) {
```

ybiquitous (changes requested):
@Mouvedia Thanks for the pull request. Can you take a look at my review comments?

lib/rules/function-url-quotes/index.mjs
Mouvedia: > How about adding parsedValue to the callback's arguments in functionArgumentsSearch()

Search for it in the codebase.
i.e. 4 other rules uses it as well

ybiquitous: Right. I think there is no effect on other rules, even if adding an argument to the callback.

ybiquitous: > And what you are proposing is not possible.

How about this? 👇🏼 We can simplify the code around `reportQueue` without any effects on other rules:

```diff
--- a/lib/rules/function-url-quotes/index.mjs
+++ b/lib/rules/function-url-quotes/index.mjs
@@ -94,23 +94,17 @@ const rule = (primary, secondaryOptions) => {
 		 * @param {number} startIndex
 		 */
 		function complain(node, source, startIndex) {
-			/** @type {ReportQueue} */
-			const reportQueue = [];
-			const parsedValue = functionArgumentsSearch(source, /^url$/i, (args, index, funcNode) => {
+			functionArgumentsSearch(source, /^url$/i, (args, index, funcNode, parsedValue) => {
 				const object = checkArgs(args, startIndex + index, funcNode);
 
 				if (object) {
 					const { messageType, ...problem } = object;
 
-					reportQueue.push([problem, [node, funcNode, messageType]]);
+					const fix = fixer(node, funcNode, messageType, parsedValue);
+
+					report({ ...problem, node, fix });
 				}
 			});
-
-			reportQueue.forEach(([problem, argsArray]) => {
-				const fix = fixer.apply(null, [...argsArray, parsedValue]);
-
-				report({ ...problem, node, fix });
-			});
 		}
 
 		/** @param {Declaration} decl */
diff --git a/lib/utils/functionArgumentsSearch.mjs b/lib/utils/functionArgumentsSearch.mjs
index 044ad0522..ff49cce34 100644
--- a/lib/utils/functionArgumentsSearch.mjs
+++ b/lib/utils/functionArgumentsSearch.mjs
@@ -15,11 +15,13 @@ import { assert, isRegExp, isString } from './validateTypes.mjs';
  *
  * @param {string} source
  * @param {string | RegExp} functionName
- * @param {(expression: string, expressionIndex: number, funcNode: valueParser.FunctionNode) => void} callback
+ * @param {(expression: string, expressionIndex: number, funcNode: valueParser.FunctionNode, parsedValue: valueParser.ParsedValue) => void} callback
  * @returns {valueParser.ParsedValue}
  */
 export default function functionArgumentsSearch(source, functionName, callback) {
-	return valueParser(source).walk((node) => {
+	const parsedValue = valueParser(source);
+
+	return parsedValue.walk((node) => {
 		if (node.type !== 'function') return;
 
 		const { value } = node;
@@ -36,6 +38,6 @@ export default function functionArgumentsSearch(source, functionName, callback)
 		const parenLength = 1; // == '('
 		const expressionIndex = node.sourceIndex + value.length + parenLength;
 
-		callback(expression, expressionIndex, node);
+		callback(expression, expressionIndex, node, parsedValue);
 	});
 }
```

Mouvedia: This is better and will permit the other rules to have fixes added more easily.

lib/rules/function-url-quotes/index.mjs
```diff
@@ -54,48 +58,72 @@ const rule = (primary, secondaryOptions, context) => {
...
 		 */
+		function complain(node, source, startIndex) {
+			functionArgumentsSearch(source, /^url$/i, (args, index, funcNode, parsedValue) => {
+				const object = checkArgs(args, startIndex + index, funcNode);
+
+				if (object) {
+					const { messageType, ...problem } = object;
+					const fix = getFix(node, funcNode, messageType, parsedValue);
+
+					report({ ...problem, node, fix });
```
ybiquitous, line 101: [suggest] I also want to simplify `checkArgs()` and make `report()` arguments clearer for readability. The `checkArgs()` logic actually does not relate to `funcNode`, `ruleName`, nor `result`. For example:

```js
const object = checkArgs(args, startIndex + index);

if (object) {
	const { messageType, index, endIndex } = object;
	const message = messages[messageType](funcNode.value.toLowerCase());
	const fix = getFix(node, funcNode, messageType, parsedValue);
	
	report({ ruleName, result, node, index, endIndex, message, fix });
}
```

Mouvedia: `index` clashes with other `index`; Ill use `...rest`.

ybiquitous (approved):
Thanks, much simpler. LGTM 👍🏼

ybiquitous:
> I am leaving `function-calc-no-unspaced-operator` in your competent hands.
> Looking at #7655, you are probably way more equipped to refactor that rule than me.

Okay, I'll take the task.

Mouvedia:
Thanks for the review.

PR #7826, opened 2024-07-06
Add `formatter` property in configuration

### LINKED ISSUE #7621: Add `formatter` property in configuration

If a formatter in a configuration (e.g. `stylelint.config.js`) is supported, we can simplify `stylelint` CLI arguments or specify a formatter via an environment variable. For example:

Before:

```sh
stylelint a.css -f json
stylelint b.css -f json
# ...
```

After:

```sh
stylelint a.css
stylelint b.css
# ...
```

```jsonc
// .stylelintrc.json
{
  "formatter": "json"
}
```

```js
// .stylelintrc.js
export default {
  formatter: process.env['STYLELINT_FORMATTER']
};
```

- Add a `formatter` property in a configuration

```js
// stylelint.config.js
import myCustomFormatter from 'my-custom-formatter';

export default {
  // A built-in formatter name
  formatter: 'string', // | 'compact' | 'json' | ...

  // Or a path to a custom formatter module
  formatter: './path/to/my-custom-formatter.js',

  // Or a custom formatter module
  formatter: myCustomFormatter,

  rules: { /* ... */ },
};
```

- It's possible to override the property by CLI arguments: `--formatter` or `--custom-formatter`

Background: https://github.com/stylelint/stylelint/issues/7617#issuecomment-2061279950

emmacharp:

Closes #7621

Tests and doc are pending. Not sure about the logic of precedence for cli flags vs. config property.


ybiquitous:
@emmacharp Thanks for the pull request. I've marked this PR as a draft to make our maintenance easier. Feel free to remove the draft state once you are ready. 🙏🏼

emmacharp:
@ybiquitous, I was waiting for your input to go ahead here. Do you prefer I complete the tests and everything before having on a look on the feature implementation?

ybiquitous:
Ah, I missed your concern, sorry. I'll look at the direction later.

ybiquitous:
@emmacharp Generally speaking, I think a CLI flag should be prioritized over a configuration because CLI operations are usually easier than changing configuration files.

E.g.

```jsonc
// .stylelintrc.json
{
	"formatter": "json"
}
```

```shell
# Use "json" formatter in the config without a CLI flag
$ stylelint *.css

# Use "verbose" formatter overriding "json" in the config
$ stylelint *.css --formatter=verbose
```

emmacharp:
Hello, @ybiquitous! 

Tests and doc added. Still got problems with the snapshots though. I've run this but to no avail:

```sh
npm run test --ignore-scripts -- --updateSnapshot system-tests/
```

Any idea what is wrong?

ybiquitous:
Hum, currently, snapshot tests seem to be free of failures...
I recommend first fixing the type errors. 😃

types/stylelint/index.d.ts
```diff
@@ -109,6 +109,7 @@ declare namespace stylelint {
 		ignorePatterns?: string;
 		rules?: ConfigRules;
 		quiet?: boolean;
+		formatter?: string;
```
ybiquitous, line 112: [suggest] Probably, to resolve the type errors, this might be required to be:

```suggestion
		formatter?: FormatterType | Formatter;
```

lib/cli.mjs
```diff
@@ -368,10 +369,22 @@ export default async function main(argv) {
...
-	} else if (isString(formatterInput)) {
+	} else if (cliFlagExists && isString(formatterInput)) {
 		formatter = /** @type {import('stylelint').FormatterType} */ (formatterInput);
+	} else if (configFormatter && !configFormatterExists) {
+		formatter = await importCustomFormatter(configFormatter);
+	} else if (configFormatter) {
+		formatter = /** @type {import('stylelint').FormatterType} */ (configFormatter);
+	} else {
+		formatter = DEFAULT_FORMATTER;
 	}
```
ybiquitous, line 388: [suggest] I now understand `cli.mjs` must load a configuration to determine the default formatter. This would make the code more complex than necessary.

So, how about adding an internal default formatter option to `LinterOptions`? This could allow the default formatter logic to be determined in one place in `standalone.mjs`. For example:

```diff
export type LinterOptions = {
		// ...
 		formatter?: FormatterType | Formatter;
+		/** @internal */
+		_defaultFormatter?: FormatterType;
		// ...
};
```

types/stylelint/index.d.ts
ybiquitous: Or we could do this:

```suggestion
		formatter?: FormatterType;
```

ybiquitous:
To make the first step easier, we could start with the minimum feature: only supporting built-in formatters in a configuration. After it is released, this feature can be extended.

```js
{
  formatter: 'json' // or 'compact', 'verbose', ...
}
```

Any thoughts?

lib/standalone.cjs
```diff
@@ -108,6 +99,17 @@ async function standalone({
 		quietDeprecationWarnings,
 	});
 
+	const configForFile = await getConfigForFile(stylelint, configFile, configFile);
```
Mouvedia, line 102: If a config is passed through the cli it should be used instead.
i.e. `config ?? […]`

lib/standalone.cjs
```diff
@@ -299,11 +301,15 @@ async function standalone({
...
  * @param {FormatterType | Formatter | undefined} selected
+ * @param {string | undefined} configFormatter // Added parameter
  * @returns {Promise<Formatter>}
  */
-function getFormatterFunction(selected) {
-	if (typeof selected === 'string') {
-		const formatterFunction = index[selected];
+function getFormatterFunction(selected, configFormatter) {
+	// Modified function signature
+	const formatterToUse = selected || configFormatter; // Choose formatter from config if not specified
```
Mouvedia, line 309: minor
refactor

This should probably be done before calling this function.
i.e. this function should have only one argument

lib/cli.mjs
```diff
@@ -368,10 +369,22 @@ export default async function main(argv) {
 
 	let formatter = undefined;
 
+	const cwd = process.cwd();
+	const resolvedConfig = await resolveConfig(cwd, { configFile, configBasedir });
+	const configFormatter = resolvedConfig?.formatter;
+	const cliFlagExists = Object.prototype.hasOwnProperty.call(cli.flags, 'formatter');
```
Mouvedia, line 375: nit

This variable name is too generic.

lib/cli.mjs
Mouvedia: question

I don't know about @ybiquitous solution but I agree that something is wrong about this.
i.e. it seems to me that there's code that doesn't belong in cli.mjs or
Maybe an util could be created that would return the formatter?
e.g. `getFormatter(cli.flags)`

emmacharp:
> To make the first step easier, we could start with the minimum feature: only supporting built-in formatters in a configuration. After it is released, this feature can be extended.
> 
> ```js
> {
>   formatter: 'json' // or 'compact', 'verbose', ...
> }
> ```
> 
> Any thoughts?

Personally, I would make use of the custom formatter. So I'd prefer if it was implemented now, but I'll follow your lead, obviously.

lib/cli.mjs
emmacharp: which name?

Mouvedia: `cliFlagExists`

It could be any flag. It's really minor though.
All the code related to `formatter` being assigned should probably be refactored.
e.g. add a function
see https://github.com/stylelint/stylelint/pull/7826#discussion_r1685183481

Mouvedia:
> Personally, I would make use of the custom formatter. So I'd prefer if it was implemented now […]

I agree.

related: Phillip9587/nx-stylelint#634

lib/cli.mjs
emmacharp: oh. I think I got it. changing for `cliConfigFlagExists`

lib/cli.mjs
emmacharp: I opted for @Mouvedia approach.

lib/standalone.cjs
emmacharp: I think the requested changes to the getFormatterFunction resolved this too.

lib/standalone.cjs
emmacharp: Changed.

lib/cli.mjs
emmacharp: Changed.

emmacharp:
I've broken other tests. I'll have a look in the next few days. I'm out for today.

ybiquitous:
Okay. Let's implement the support for custom formatters in this PR. 👍🏼

lib/standalone.cjs
```diff
@@ -126,6 +126,8 @@ async function standalone({
 				? path.join(cwd, codeFilename)
 				: codeFilename;
 
+		formatterFunction = await getFormatter(stylelint, { filePath: absoluteCodeFilename });
```
Mouvedia, line 129: What happens if `absoluteCodeFilename` is falsy?
e.g. if `codeFilename` is `undefined`

lib/standalone.mjs
Mouvedia, line 1: Since it's a draft I didn't go too deep in my analysis, but at first glance I would say that having to reset a variable that many times is probably a code smell.

emmacharp: I agree. But I didn't know how to get the changing paths in another way. Maybe you have an idea?

Mouvedia: This is not a critical regression so you can take your time thinking about it.
I am just giving you preliminary advices.
Once it's ready to review, Ill check it out in details.

emmacharp:
Almost there, @ybiquitous!

There are two tests still failing but I think that they may be changed so they pass. Here is one of two (it's the same error):
https://github.com/emmacharp/stylelint/blob/9bc35126c63677954dbb28aeecc6efee9cda0325/lib/__tests__/stylelintignore-test/stylelintignore.test.mjs#L234

The "configuration not found" error message is now returned but the test is expecting an empty result. If I understand correctly, the new result seems better so we may change the expected result in the test.

What do you think?

lib/__tests__/cli.test.mjs
```diff
@@ -36,7 +36,6 @@ describe('buildCLI', () => {
 			color: false,
 			disableDefaultIgnores: false,
 			fix: false,
-			formatter: 'string',
```
Mouvedia, line 39: Which tests are failing when you leave it there?
i.e. I suppose both

https://github.com/stylelint/stylelint/blob/9bc35126c63677954dbb28aeecc6efee9cda0325/lib/__tests__/cli.test.mjs#L91-L93

and

https://github.com/stylelint/stylelint/blob/9bc35126c63677954dbb28aeecc6efee9cda0325/lib/__tests__/cli.test.mjs#L108-L111

were expected to overwrite it.

ybiquitous:
> The "configuration not found" error message is now returned but the test is expecting an empty result. If I understand correctly, the new result seems better so we may change the expected result in the test.

I think we should adjust the logic around `getConfigForFile` in `getFormatter.mjs` to keep the current behavior. For example, as this code, if a config file is not found, we should fall back to the default formatter, I guess:

```js
try {
	await getConfigForFile(stylelint, configPath, filePath)
} catch (err) {
	if (err instanceof Error && 'code' in err && err.code === EXIT_CODE_INVALID_CONFIG) {
		// fallback
	} else {
		throw err;
	}
}
```

lib/__tests__/cli.test.mjs
emmacharp: Only `flags - default` fail since the default is no longer only a flag, but in separate function (ie: `getFormatter`).

emmacharp:
@ybiquitous, I've followed your advice, thanks.

Only code coverage issues remain in checks. Not sure how to test for it though. Any idea (again)?

lib/utils/getFormatter.cjs
```diff
@@ -0,0 +1,64 @@
...
+	if (!formatter) {
+		let configForFile;
+
+		try {
+			configForFile = await getConfigForFile(stylelint, configPath, filePath);
+		} catch (err) {
+			if (err instanceof Error && 'code' in err && err.code === constants.EXIT_CODE_INVALID_CONFIG) {
+				configForFile = undefined;
+			} else {
+				throw err;
```
Mouvedia, line 34:
```
throw new Error('foo bar qux');
```

lib/utils/getFormatter.mjs
```diff
@@ -0,0 +1,58 @@
...
+	if (!formatter) {
+		let configForFile;
+
+		try {
+			configForFile = await getConfigForFile(stylelint, configPath, filePath);
+		} catch (err) {
+			if (err instanceof Error && 'code' in err && err.code === EXIT_CODE_INVALID_CONFIG) {
+				configForFile = undefined;
+			} else {
+				throw err;
```
Mouvedia, line 30: This should have a message.
e.g. `throw new Error('foo bar qux');`

ybiquitous:
@emmacharp Okay, let's sync this feature branch with the `main` branch and make this PR ready for review!

lib/utils/getFormatter.mjs
ybiquitous: I don't think it's necessary to throw the original `err`.

lib/standalone.mjs
```diff
@@ -116,6 +100,24 @@ export default async function standalone({
...
+
+	try {
+		/** @see https://github.com/stylelint/stylelint/issues/7447 */
+		if (!quietDeprecationWarnings && formatter === 'github') {
+			emitDeprecationWarning(
+				'"github" formatter is deprecated.',
+				'GITHUB_FORMATTER',
+				'Please use "stylelint-actions-formatters" instead.',
+			);
+		}
```
ybiquitous, line 114: [suggest] Can you move this `emitDeprecationWarning()` into `getFormatter()`?

lib/standalone.mjs
```diff
@@ -124,6 +126,8 @@ export default async function standalone({
 				? join(cwd, codeFilename)
 				: codeFilename;
 
+		formatterFunction = await getFormatter(stylelint, { filePath: absoluteCodeFilename });
```
ybiquitous, line 129: [suggest] I think we should call `getFormatter()` at once. In other words, the `formatter` should be determined in a root-level config. If it's possible, the `filePath` parameter will be no longer necessary.

ybiquitous (changes requested):

lib/standalone.mjs
emmacharp: I tried to but getFormatter can be called more than one time in certain cases so the message was printed more than once. Tried to keep it from happening but in vain. Any idea?

lib/standalone.mjs
emmacharp: Not sure I understand correctly... root-level instead of on a per-file basis? Maybe by using `resolveConfig()` instead of `getConfigForFile()` ?

lib/cli.mjs
```diff
@@ -1,4 +1,5 @@
 import { isAbsolute, join, resolve } from 'node:path';
+import { DEFAULT_FORMATTER } from './constants.mjs';
```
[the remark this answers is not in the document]
emmacharp: Is a simple commit enough or would you prefer a full-on rebase?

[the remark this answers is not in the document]
ybiquitous: A simple commit is enough. 👍🏼 

We will eventually squash into one commit when merging a PR.

docs/user-guide/configure.md
```diff
@@ -648,3 +648,9 @@ For example:
 > This config option should not be overridden on a per-file basis.
 
 [More info](options.md#fix).
+
+## `formatter`
+
+Specify the formatter to format your results.
+
+[More info](options.md#formatter).
```
ybiquitous, line 656: [suggest] Can you add a sample config code for `formatter`? It should help people understand the usage.

lib/standalone.mjs
ybiquitous: > root-level instead of on a per-file basis?

Yes.

> Maybe by using `resolveConfig()` instead of `getConfigForFile()` ?

No. I think it's unnecessary to pass the third `filePath` argument to `getConfigForFile()` if I understand correctly.

lib/standalone.mjs
ybiquitous: Okay, let's keep it as-is.

lib/standalone.mjs
```diff
@@ -116,6 +100,24 @@ export default async function standalone({
...
+				'"github" formatter is deprecated.',
+				'GITHUB_FORMATTER',
+				'Please use "stylelint-actions-formatters" instead.',
+			);
+		}
+
+		formatterFunction = await getFormatter(stylelint, { formatterType: 'json' });
+	} catch (error) {
+		return Promise.reject(error);
+	}
```
ybiquitous, line 119: [ask] Can we remove this `cache`? It seems unnecessary here.

lib/standalone.mjs
emmacharp: Hmmm. I'm pretty sure at one point some tests were failing without the filePath argument... but, hey it seems to work now without it. Hehe. Not sure I understand why/how though, sadly.

docs/user-guide/configure.md
emmacharp: Done. Not sure about the format though.

docs/user-guide/configure.md
```diff
@@ -658,3 +658,37 @@ For example:
...
+
+- A formatter function
+
+```js
+export default {
+  formatter: () => {
+    // ...
+  }
+};
+```
```
ybiquitous, line 692: [suggestion] Thanks for improving the document. Additionally, can you make one list by indenting the code blocks? E.g.

````suggestion
- The name of a provided formatter.
  ```json
  {
    "formatter": "string"
  }
  ```
- A path to a custom formatter function.
  ```json
  {
    "formatter": "path/to/customformatter.js"
  }
  ```
- A formatter function.
  ```js
  export default {
    formatter: () => { /*..*/ }
  };
  ```
````

File: docs/user-guide/configure.md [hunk 16]
```diff
@@ -658,3 +658,37 @@ For example:
...
+
+- A path to a custom formatter function.
+
+```json
+{
+  "formatter": "path/to/customformatter.js"
+}
+```
+
+- A formatter function
```
ybiquitous, line 684: [suggestion] For consistency with other items above, adding a period seems better:

```suggestion
- A formatter function.
```

docs/user-guide/configure.md
```diff
@@ -658,3 +658,37 @@ For example:
...
+
+```js
+export default {
+  formatter: () => {
+    // ...
+  }
```
ybiquitous, line 690: [suggestion] How about saving a few lines like this?

```suggestion
  formatter: () => { /* ... */ }
```

lib/__tests__/standalone-formatter.test.mjs
```diff
@@ -57,3 +58,52 @@ it('standalone with input css and custom promised formatter', async () => {
...
+		config: configWithFormatter,
+	});
+
+	const strippedOutput = stripAnsi(report);
+
+	expect(strippedOutput).toContain('1:3');
+	expect(strippedOutput).toContain('block-no-empty');
+});
+
+it('standalone with CLI formatter overriding configuration formatter', async () => {
```
ybiquitous, line 79: [suggestion] This code isn't actually related to CLI, avoiding the word seems better:

```suggestion
it('standalone with formatter argument overriding configuration formatter', async () => {
```

lib/augmentConfig.mjs
```diff
@@ -523,5 +523,9 @@ function addOptions(stylelint, config) {
 		augmentedConfig.fix = stylelint._options.fix;
 	}
 
+	if (stylelint._options.formatter) {
+		augmentedConfig.formatter = stylelint._options.formatter;
+	}
```
ybiquitous, line 528: [suggestion] This code looks like unnecessary. Is it possible to remove?

lib/standalone.mjs
```diff
@@ -116,6 +100,20 @@ export default async function standalone({
...
+	/** @see https://github.com/stylelint/stylelint/issues/7447 */
+	if (!quietDeprecationWarnings && formatter === 'github') {
+		emitDeprecationWarning(
+			'"github" formatter is deprecated.',
+			'GITHUB_FORMATTER',
+			'Please use "stylelint-actions-formatters" instead.',
+		);
+	}
+
+	formatterFunction = await getFormatter(stylelint, { formatterType: 'json' });
```
ybiquitous, line 115: [suggestion] I think calling `getFormatter()` once is enough and `formatterType: 'json'` is no longer needed:

```suggestion
	const formatterFunction = await getFormatter(stylelint);
```

lib/utils/getFormatter.mjs
```diff
@@ -0,0 +1,64 @@
...
+ * @param {InternalApi} stylelint
+ * @param {Object} [params]
+ * @param {FormatterType | Formatter | undefined} [params.formatterType]
+ * @returns {Promise<Formatter>}
+ */
+export default async function getFormatter(stylelint, params = {}) {
+	const { formatterType } = params;
+	const cwd = stylelint._options.cwd;
+	const configPath = stylelint._options.configFile || cwd;
+	let formatter = formatterType;
```
ybiquitous, line 21: [suggestion] I think we can simplify this function logic without `params`.

First, it tries getting a formatter from linter options (i.e., `standalone()` arguments):

```suggestion
	let formatter = stylelint._options.formatter;
```

lib/utils/getFormatter.mjs
```diff
@@ -0,0 +1,64 @@
...
+			} else {
+				throw err;
+			}
+		}
+
+		const resolvedConfig = stylelint._options.config ? stylelint._options : configForFile;
+		const configFormatter = resolvedConfig?.config.formatter;
+
+		formatter =
+			stylelint._options.formatter || configFormatter || stylelint._options._defaultFormatter;
```
ybiquitous, line 40: [suggestion] Next, it tries getting a formatter from the config:

```suggestion

		formatter = resolvedConfig?.config.formatter;
```

lib/utils/getFormatter.mjs
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	if (typeof formatter === 'string') {
+		const formatterFunction = formatters[formatter];
+
+		if (formatterFunction === undefined) {
+			if (existsSync(formatter)) {
+				formatter = await dynamicImport(resolve(formatter)).then((m) => m.default);
```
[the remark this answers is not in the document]
ybiquitous: [suggestion] Can you fix the code and add a test case to resolve the warnings?

lib/utils/getFormatter.mjs
```diff
@@ -0,0 +1,64 @@
...
+				const formattersText = getFormatterOptionsText(', ', '"');
+
+				throw new Error(`You must use a valid formatter option: ${formattersText} or a function`);
+			}
+		}
+
+		return formatterFunction;
+	}
+
+	if (typeof formatter === 'function') {
```
ybiquitous, line 59: [suggestion] If a formatter is present, assume it's a function or a promise of function.

```suggestion
	// Assume a function or a promise of a function.
	if (typeof formatter === 'function' || formatter) {
```

lib/utils/getFormatter.mjs
```diff
@@ -0,0 +1,64 @@
...
+		return Promise.resolve(formatter);
+	}
+
+	return formatters.json;
```
ybiquitous, line 63: [suggestion] Last, if a formatter is not found, it falls back to the default formatter option or `json`:

```suggestion
	formatter ??= stylelint._options._defaultFormatter ?? 'json';

	return formatters[formatter];
```

ybiquitous (changes requested):
@emmacharp Sorry for the late response. 🙇🏼
The code has been getting better, but I still have some concerns. Can you have a look at my reviews? Thanks.

lib/standalone.mjs
emmacharp: Well, one seems like it's enough indeed. Pretty sure at one point it wasn't though. Hehe. To be honest, I'm not sure why or how it works now, by being called only once before... Is it because we removed the getConfigFromFile logic?

lib/utils/getFormatter.mjs
emmacharp: The _defaultFormatter seems to be needed here or the default will be JSON. Should I keep it?

lib/utils/getFormatter.mjs
```diff
@@ -0,0 +1,57 @@
...
+
+		return formatterFunction;
+	}
+
+	// Assume a function or a promise of a function.
+	if (typeof formatter === 'function' || formatter) {
+		return Promise.resolve(formatter);
+	}
+
+	formatter ??= stylelint._options._defaultFormatter ?? 'json';
```
Mouvedia, line 54: question

Is `_defaultFormatter` only being used here?
If so what about removing it completely and just using `DEFAULT_FORMATTER` directly here?

ybiquitous: Sadyly we cannot do that because `standalone` uses `json` by default, while `cli` uses `string` by default. The default formatter depends on the context.

lib/__tests__/standalone-formatter.test.mjs
```diff
@@ -57,3 +58,52 @@ it('standalone with input css and custom promised formatter', async () => {
...
+		formatter: customFormatter,
+	};
+
+	const { report } = await standalone({
+		code: 'a {}',
+		config: configWithCustomFormatter,
+	});
+
+	expect(report).toContain('errored: true');
+});
```
ybiquitous, line 109: [suggestion] It seems there is no case for `"formatter": "path/to/customformatter.js"`. Can you add it?

lib/utils/getFormatter.mjs
Mouvedia: @ybiquitous [this line](https://github.com/stylelint/stylelint/pull/7826/files#diff-d1470268bb8c3a42692b2c5249d7d31832ceeed1b1f16fbdbc814f8013c36b31R379) is the only line where `_defaultFormatter` is set, correct?
Are you saying that it's necessary so that it can be set in our tests?

lib/standalone.mjs
ybiquitous: > Is it because we removed the getConfigFromFile path logic?

Right. In Stylelint (internally using [Cosmiconfig](https://github.com/cosmiconfig/cosmiconfig)), there can be multiple config files in a project, e.g.

```
(root)
├── src/
│   ├── a.css 
│   └── .stylelintrc.json
├── b.css 
└── .stylelintrc.json
```

So, which configuration file to use depends on the source file that is currently being analyzed. That's why `getConfigFromFile()` is used.

However, the `formatter` option makes sense only at the root config file. So, `getConfigFromFile()` is not required in this case.

lib/utils/getFormatter.mjs
ybiquitous: Yes, `_defaultFormatter` is set only in `cli` for now. This is not for the test but for CLI. Users shouldn't use it, but the Stylelint implementation needs it.

Mouvedia: > **for now**

As I understand it, `_defaultFormatter` should be set elsewhere; does it mean that the context switch will be done in this PR?

ybiquitous: No more changes are necessary in this PR, I think.

Mouvedia: > but the Stylelint implementation needs it

What I am trying to get at is _where/when_?
i.e. if `standalone` uses `'json'` by default, where or when is `_defaultFormatter` supposed to be set to `'json'`?
In a separate PR?
If it's not meant to be used for our tests nor be used by users, then what will set it to `'json'`?

lib/utils/getFormatter.mjs
```diff
@@ -0,0 +1,64 @@
...
+			if (err instanceof Error && 'code' in err && err.code === EXIT_CODE_INVALID_CONFIG) {
+				configForFile = undefined;
+			} else {
+				throw err;
+			}
+		}
+
+		const resolvedConfig = stylelint._options.config ? stylelint._options : configForFile;
+
+		formatter = resolvedConfig?.config.formatter;
```
ybiquitous, line 37: [suggestion] `stylelint._options.config` seems no longer needed. Can we simplify the code like this?

```suggestion
		formatter = configForFile?.config.formatter;
```

lib/utils/getFormatter.mjs
```diff
@@ -0,0 +1,64 @@
...
+	let formatter = stylelint._options.formatter;
+
+	if (!formatter) {
+		let configForFile;
+
+		try {
+			configForFile = await getConfigForFile(stylelint, configPath);
+		} catch (err) {
+			if (err instanceof Error && 'code' in err && err.code === EXIT_CODE_INVALID_CONFIG) {
+				configForFile = undefined;
```
ybiquitous, line 29: [suggestion] This line seems to be uncovered by test cases. Can you add a test case to `standalone-formatter.test.mjs` when a config file doesn't exist and the `formatter` argument is unspecified?

ybiquitous (commented):
@emmacharp I appreciate your continuous contribution for this feature, and please don't mind. We could find many things overlooked while the PR has been going.

Overall LGTM, but let me leave some additional reviews. They are tiny refactoring and tests to edge cases.

emmacharp:
@ybiquitous, there is still a check failing about test coverage. I'm not sure how to test for the line mentioned. Do I have to force getConfigForFile to throw an expected error?

ybiquitous:
No need to test the uncovered line because of an edge case. 👍🏼

ybiquitous (approved):
@emmacharp I appreciate your continuous efforts on this new feature! LGTM 👍🏼 

I'll merge this PR after a while if anyone else has no concerns.

lib/cli.mjs
Mouvedia, line 1: questions

@ybiquitous

1. minor

`_defaultFormatter` is set before

https://github.com/stylelint/stylelint/blob/123d1bf74e993718c7031c19e8e8994164776a10/lib/cli.mjs#L489

Do we want to print it?
If not, what about setting it after?

2. medium

Since `options` is only being passed to

https://github.com/stylelint/stylelint/blob/123d1bf74e993718c7031c19e8e8994164776a10/lib/cli.mjs#L504

could `_defaultFormatter` be replaced by `formatter = DEFAULT_FORMATTER,` at [line 90](https://github.com/stylelint/stylelint/pull/7826/files#diff-a9211bfb8f0453e9d6e413fe613dbdcd388621ae34881d7400c93f10dafe516bR90) of `standalone.mjs`?

ybiquitous: > `_defaultFormatter` is set before
>
> Do we want to print it?
> If not, what about setting it after?

No worry. `printConfig()` doesn't output the `options`, but a resolved config. It doesn't include `_defaultFormatter`.

https://github.com/stylelint/stylelint/blob/7ee28f4e266ecb2ac35f093e46606798634fa067/lib/printConfig.mjs#L35-L42

---

> could `_defaultFormatter` be replaced by `formatter = DEFAULT_FORMATTER,`

Unfortunately, no. Depending on the context, `_defaultFormatter` in the `standalone()` arguments is changed:

| Context | `_defaultFormatter` |
|--------|--------|
| CLI | `string` |
| non CLI | `json` |

If replacing `_defaultFormatter` with `formatter = DEFAULT_FORMATTER`, a configuration would be never read in `getFormatter()`:

https://github.com/stylelint/stylelint/blob/123d1bf74e993718c7031c19e8e8994164776a10/lib/utils/getFormatter.mjs#L18-L21

emmacharp:
Great!

And thanks to you for your patience, @ybiquitous! I'll keep an eye on issues and try to continue helping on a  more modest level. Hehe.

A nice day to you

lib/standalone.mjs
```diff
@@ -118,6 +102,19 @@ export default async function standalone({
 		validate,
 	});
 
+	/** @type {Formatter} */
```
Mouvedia, line 105: This should have been just before `const formatterFunction = await getFormatter(stylelint);`.

PR #7831, opened 2024-07-08
Refactor `function-calc-no-unspaced-operator` to use `fix` callback instead of `context.fix`

ybiquitous:

Part of #2643

This PR aims to perform refactoring for #2643. But, this work was not as easy as just replacing `context.fix` with the `fix` callback. Before switching to the `fix` callback, I needed a big refactoring to rewrite the code to class-based logic for the code readability. Please look at each commit for details.

### DIFF
--- a/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
+++ b/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
@@ -280,18 +280,16 @@ testRule({
 			fixed: 'a { scale: rem(10 + 2, 1.7); }',
 			warnings: [
 				{ message: messages.expectedBefore('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
-				{ message: messages.expectedAfter('+'), line: 1, column: 19, endLine: 1, endColumn: 20 },
+				{ message: messages.expectedAfter('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
 			],
-			skip: true,
 		},
 		{
 			code: 'a { rotate: rem(10turn, 2turn+1turn); }',
 			fixed: 'a { rotate: rem(10turn, 2turn + 1turn); }',
 			warnings: [
-				{ message: messages.expectedBefore('+'), line: 1, endLine: 1 },
-				{ message: messages.expectedAfter('+'), line: 1, endLine: 1 },
+				{ message: messages.expectedBefore('+'), line: 1, column: 30, endLine: 1, endColumn: 31 },
+				{ message: messages.expectedAfter('+'), line: 1, column: 30, endLine: 1, endColumn: 31 },
 			],
-			skip: true,
 		},
 		{
 			code: 'a { top: calc(1px +\t-1px); }',
--- a/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
+++ b/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
@@ -263,32 +263,32 @@ testRule({
 			fixed: 'a { top: calc(2px + 1px); }',
 			description: 'no space before or after operator',
 			warnings: [
-				{ message: messages.expectedBefore('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
 				{ message: messages.expectedAfter('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
+				{ message: messages.expectedBefore('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
 			],
 		},
 		{
 			code: 'a { transform: rotate(acos(2+1)); }',
 			fixed: 'a { transform: rotate(acos(2 + 1)); }',
 			warnings: [
-				{ message: messages.expectedBefore('+'), line: 1, column: 29, endLine: 1, endColumn: 30 },
 				{ message: messages.expectedAfter('+'), line: 1, column: 29, endLine: 1, endColumn: 30 },
+				{ message: messages.expectedBefore('+'), line: 1, column: 29, endLine: 1, endColumn: 30 },
 			],
 		},
 		{
 			code: 'a { scale: rem(10+2, 1.7); }',
 			fixed: 'a { scale: rem(10 + 2, 1.7); }',
 			warnings: [
-				{ message: messages.expectedBefore('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
 				{ message: messages.expectedAfter('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
+				{ message: messages.expectedBefore('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
 			],
 		},
 		{
 			code: 'a { rotate: rem(10turn, 2turn+1turn); }',
 			fixed: 'a { rotate: rem(10turn, 2turn + 1turn); }',
 			warnings: [
-				{ message: messages.expectedBefore('+'), line: 1, column: 30, endLine: 1, endColumn: 31 },
 				{ message: messages.expectedAfter('+'), line: 1, column: 30, endLine: 1, endColumn: 31 },
+				{ message: messages.expectedBefore('+'), line: 1, column: 30, endLine: 1, endColumn: 31 },
 			],
 		},
 		{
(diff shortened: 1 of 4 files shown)


lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
```diff
@@ -280,18 +280,16 @@ testRule({
 			fixed: 'a { scale: rem(10 + 2, 1.7); }',
 			warnings: [
 				{ message: messages.expectedBefore('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
-				{ message: messages.expectedAfter('+'), line: 1, column: 19, endLine: 1, endColumn: 20 },
+				{ message: messages.expectedAfter('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
 			],
-			skip: true,
```
ybiquitous, line 285: [note] For some reason, the two cases were skipped.

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -222,21 +135,14 @@ const rule = (primary, _secondaryOptions, context) => {
...
-						const { unit } = token[4];
-						const operatorChar = unit.slice(-1);
+						const [, , , endPos, { unit }] = token;
+						const operatorChar = unit.at(-1);
 
 						if (operatorChar && OPERATOR_REGEX.test(operatorChar)) {
-							fixOperationWithoutOperator(operation, token, operatorChar);
+							operation.completeMissingOperator(operatorChar, endPos, 'append');
 
 							if (context.fix) {
```
ybiquitous, line 144: [note] Note that `context.fix` is still necessary because it's impossible to call `report()` at this point. After completing a missing operator, we need to insert whitespaces by autofix.

lib/utils/__tests__/checkAgainstRule.test.mjs
```diff
@@ -87,27 +87,6 @@ describe('checkAgainstRule', () => {
 		expect(warnings[0].column).toBe(1);
 	});
 
-	/** @todo remove once all fixable rules have been been migrated from context.fix */
-	it('outputs fixed code when provided a context object', async () => {
```
ybiquitous, line 91: [ask] Don't we ensure the compatibility of `checkAgainstRule()`? 🤔

lib/rules/function-calc-no-unspaced-operator/index.mjs
romainmenke: I was looking at this rule too and noticed this aswel.

My thinking at the time was that one function can extract and restore the operator.
And another function can be dedicated to fixing whitespace.

But maybe better to refactor this so that each mutation fully fixes each error?

ybiquitous: Sounds good. Can you push a commit to this PR or create a new PR?

romainmenke: I don't have much time today, but hopefully tomorrow I can look into this :)

(separate PR is maybe is easier? I'll see how large the change is)

ybiquitous: Thanks. I'm fine with either. As you said, this PR is already quite big. 😅

Mouvedia: > Note that `context.fix` is still necessary…

@ybiquitous When I looked at this refactor, the only solution that didn't refactor the rule from the ground up was to have an array that would collect functions which all would be called in a `forEach` inside the fixer.

lib/utils/__tests__/checkAgainstRule.test.mjs
Mouvedia: The only way would be to test a plugin rule.

ybiquitous: Just opened PR #7833, which blocks this PR.

lib/rules/function-calc-no-unspaced-operator/index.mjs
ybiquitous: @Mouvedia, your idea sounds possible, but I'm still unsure if it works. So, let's wait for @romainmenke. He might bring a better solution.

romainmenke: This was resolved in https://github.com/stylelint/stylelint/pull/7835

romainmenke:
> I needed a big refactoring to rewrite the code to class-based logic for the code readability.

Thank you for this @ybiquitous 🙇 
The code was very clear and easy to make changes to.

ybiquitous:
@romainmenke I appreciate it, too. 😄 

I've pushed one more commit d5290413d for a small refactoring.

Now, this PR is ready for review again.

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -85,134 +85,47 @@ const rule = (primary, _secondaryOptions, context) => {
...
-						);
-					} else {
-						complain(
-							messages.expectedBefore(operation.operatorToken[4].value),
-							decl,
-							valueIndex + operation.operatorToken[2],
-							operation.operatorToken[4].value,
-						);
-					}
+				if (!operation.before.some(isWhitespaceNode)) {
```
Mouvedia, line 102: suggestion
refactor

```js
/** @param {keyof messages} method */
const foo = (method) => {
	const position = method === 'expectedAfter' ? 'after' : 'before';

	if (!operation[position].some(isWhitespaceNode)) {
		complain(
			messages[method](operation.operatorChar),
			decl,
			valueIndex + operation.operatorCharPosition,
			operation.operatorChar,
			() => {
				operation.insertWhitespace(node, position);
				fixDeclarationValue();
			},
		);
	}
};
```

I am not sure if it's necessary though.
It depends on the clarity of the function name chosen.

lib/rules/function-calc-no-unspaced-operator/index.mjs
```diff
@@ -222,46 +135,46 @@ const rule = (primary, _secondaryOptions, context) => {
 					const token = operation.firstOperand.value;
 
 					if (isTokenDimension(token)) {
-						const { unit } = token[4];
-						const operatorChar = unit.slice(-1);
+						const [, , , endPos, { unit }] = token;
```
Mouvedia, line 138: question
minor

Is there a way to keep the `return` after `complain`?

```js
/** @param {'unit'|'value'} property */
const bar = (property) => {
	const [, , , endPos, { [property]: operator }] = token;
	const operatorChar = operator.at(-1);

	if (operatorChar && OPERATOR_REGEX.test(operatorChar)) {
		operation.completeMissingOperator(operatorChar, endPos, 'append');

		complain(
			messages.expectedBefore(operation.operatorChar),
			decl,
			valueIndex + operation.operatorCharPosition,
			operation.operatorChar,
			() => {
				operation.insertOperatorAfterFirstOperand(node);
				mutateUnit(token, operator.slice(0, -1));
				fixDeclarationValue();
			},
		);
	}
};
```

lib/rules/function-calc-no-unspaced-operator/index.mjs
ybiquitous: Good catch. Improved via 6944be586.

romainmenke (approved):
LGTM!
Thank you for refactoring this rule (again 😄)

lib/rules/function-calc-no-unspaced-operator/index.mjs
ybiquitous: Possible. How about the new commit 04df33fdf?

lib/rules/function-calc-no-unspaced-operator/index.cjs
```diff
@@ -200,82 +100,85 @@ const rule = (primary, _secondaryOptions, context) => {
...
 			}
 
 			/**
-			 * @param {CompleteOperation} operation
-			 * @param {CompleteOperation["before"] | CompleteOperation["after"]} whitespaceNodes
+			 * @param {Operation} operation
+			 * @param {Operation["before"] | Operation["after"]} whitespaceNodes
 			 * @param {'before' | 'after'} position
 			 */
 			function checkOperandWhitespace(operation, whitespaceNodes, position) {
```
Mouvedia, line 184: nit
refactor

Since you have `operation` and `position`, you don't need to pass `whitespaceNodes` anymore.
i.e. `const whitespaceNodes = operation[position];`

lib/rules/function-calc-no-unspaced-operator/index.cjs
```diff
@@ -358,15 +252,16 @@ const rule = (primary, _secondaryOptions, context) => {
...
 						// Step 4
 						// If there is no operator, try to find one
-						if (isOperationWithoutOperator(operation)) {
+						if (!operation.operator) {
 							checkOperationWithoutOperator(node, operation);
 						}
 
 						// Step 5
 						// If the operation is complete, ensure there is whitespace around the operator
 						// The operation might have started without an operator and might have been repaired by Step 4
```
Mouvedia, line 261: minor
question

> and might have been repaired by Step 4

Is this still true?

romainmenke: yes :)

lib/rules/function-calc-no-unspaced-operator/index.cjs
```diff
@@ -381,14 +276,74 @@ const rule = (primary, _secondaryOptions, context) => {
...
+		// Set the string representation of the open curly to `#{`
+		nextToken[1] = '#{';
+		// Remove the `#` token
+		tokens.splice(i, 1);
+	});
+
+	return cssParserAlgorithms.parseListOfComponentValues(tokens);
+}
+
 /** @see https://drafts.csswg.org/css-values/#typedef-calc-value */
```
Mouvedia, line 347: https://www.w3.org/TR/css-values-4/#typedef-calc-value

lib/rules/function-calc-no-unspaced-operator/index.cjs
```diff
@@ -418,51 +373,99 @@ function isOperandNode(node) {
...
- * @returns {operation is Operation}
- */
-function isOperationWithoutOperator(operation) {
-	return (
-		Boolean(operation.firstOperand) &&
-		Boolean(operation.secondOperand) &&
-		!operation.operator &&
-		!operation.operatorToken
-	);
+class Operation {
```
Mouvedia, line 382: Having a `class` is a great idea.

Mouvedia:
I knew you would do a better job, with the help of Romain, than me, being so familiar with the rule.
If you are in a hurry, you don't have to address my last comments.

lib/rules/function-calc-no-unspaced-operator/index.cjs
romainmenke: The draft versions always contain the latest information and are also the documents used by browser implementers.

The published versions are more of a snapshot for historical documentation purposes.
So that you can link back to a previous iteration of the spec.

lib/rules/function-calc-no-unspaced-operator/index.cjs
ybiquitous: Nice catch! Fixed via c40a3da88.

ybiquitous:
Thank you for your help with this big refactoring. 😄

PR #380, opened 2015-09-12
Stream, standalone, cli; addresses #288

davidtheclark:
This introduces a first draft of few things. I'll write up a little proto-documentation of what's in place to help explain. I made a lot of assumptions about APIs, etc., so please share feedback and I'll revise as we see fit.

There are two new exposed functions and a CLI. This also makes rule severity levels meaningful and means `.stylelintrc` files can be a reality.

**In everything below, if no `config` is explicitly provided, stylelint will use [rc](https://github.com/dominictarr/rc) (via [rc-loader](https://github.com/MoOx/rc-loader)) to find a `.stylelintrc` configuration file in the usual configuration-file places, in JSON, YML, or CommonJS format.**

---

## stylelint.lint(options)

Options:
- `files` (*required*): A file glob, ultimately passed to [node-glob](https://github.com/isaacs/node-glob) to figure out what files you're referencing
- `config` (optional): A stylelint configuration object
- `formatter` (optional): By default, [postcss-reporter](https://github.com/postcss/postcss-reporter)'s minimal formatter is used. Alternately, you can pass a function that
  - accepts an object containing a `messages` array and a `source` string
  - returns a string

Returns a Promise that resolves with an object containing the following properties:

- `output`: String. A string displaying the formatted warnings (using the default formatter or whichever you passed)
- `errored`: Boolean. If `true`, at least one rule with a severity of `2` registered a warning
- `results`: An array containing all the [PostCSS LazyResults](https://github.com/postcss/postcss/blob/master/docs/api.md#lazyresult-class) that were accumulated during processing

```js
var stylelint = require('stylelint');
var myConfig = require('./myConfig.js');

stylelint.lint({ files: 'all/my/stylesheets/*.css', config: myConfig })
  .then(function(data) {
      /* do things with data.output, data.errored, and data.results */
  });
```

## stylelint.lintStream([options])

Options:
- `files` (optional): A file glob, ultimately passed to [node-glob](https://github.com/isaacs/node-glob) to figure out what files you're referencing
- `config` (optional): A stylelint configuration object

Returns a Transform stream, which expects a readable stream of CSS or a file glob as input, runs each chunk (string or file) through stylelint, and for each chunk provides a [PostCSS LazyResult](https://github.com/postcss/postcss/blob/master/docs/api.md#lazyresult-class) object containing the messages logged by stylelint.

```js
// Using a file glob and a config object
var stylelint = require('stylelint');
var myConfig = require('./myConfig.js');

[...]

PR #7833, opened 2024-07-09
Fix `utils.checkAgainstRule()` to support `fix` callback in `utils.report()`

ybiquitous:

Follow-up to PR #7730

To make the `fix` callback work with `utils.checkAgainstRule()`, we need to use `result.stylelint` in `checkAgainstRule()` because `report()` now depends on `result.stylelint.config.fix` instead of `context.fix`.

See:
https://github.com/stylelint/stylelint/blob/2dfe7cea14159b536b83a35ab9423cd643b36a7e/lib/utils/report.mjs#L174

### DIFF
--- a/.changeset/short-goats-mix.md
+++ b/.changeset/short-goats-mix.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `stylelint.utils.checkAgainstRule()` to use `result.stylelint` if possible
+Fixed: `stylelint.utils.checkAgainstRule()` to use `result.stylelint` if present
--- a/.changeset/short-goats-mix.md
+++ b/.changeset/short-goats-mix.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `stylelint.utils.checkAgainstRule()` to use `result.stylelint` if possible
--- a/lib/utils/__tests__/checkAgainstRule.test.mjs
+++ b/lib/utils/__tests__/checkAgainstRule.test.mjs
@@ -26,6 +26,8 @@ const mockResult = {
 				},
 			},
 		},
+		ruleSeverities: {},
+		fixersData: {},
 	},
 };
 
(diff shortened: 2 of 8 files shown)


.changeset/short-goats-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `stylelint.utils.checkAgainstRule()` to use `result.stylelint` if possible
```
Mouvedia, line 5: nit
suggestions

if present
if provided
if passed

lib/utils/__tests__/checkAgainstRule.test.mjs
```diff
@@ -26,6 +26,8 @@ const mockResult = {
 				},
 			},
 		},
+		ruleSeverities: {},
+		fixersData: {},
```
Mouvedia, line 30: question

That means that

```
		customMessages: {},
		customUrls: {},
		ruleMetadata: {},
		disabledRanges: {},
```
are optional?

ybiquitous: No, they are not optional from the type definition:
https://github.com/stylelint/stylelint/blob/2dfe7cea14159b536b83a35ab9423cd643b36a7e/types/stylelint/index.d.ts#L156-L169

I pushed 09a724ee3 to clarify that. Any thoughts?

lib/utils/report.mjs
```diff
@@ -142,8 +142,6 @@ function printfLike(format, ...args) {
  * @param {DisabledRangeObject} disabledRanges
  */
 function isDisabled(ruleName, startLine, disabledRanges) {
-	if (!disabledRanges) return false;
```
Mouvedia, line 145: note

https://github.com/stylelint/stylelint/pull/380/files#diff-8d2b4819892fef9fe283797c5d2a4ef3ccd4aac43b1e25da50f5db420f5bbb27R31

Mouvedia (approved):

lib/utils/report.mjs
ybiquitous: `disabledRanges` should always be truthy, thanks to #7728.

PR #7834, opened 2024-07-09
Refactor to extract utility for deprecation warnings

ybiquitous:

None.

This change aims to facilitate the management of warning codes.


lib/utils/emitDeprecationWarning.mjs
```diff
@@ -0,0 +1,25 @@
...
+ * @param {string} message
+ * @param {keyof CODES} codeKey
+ * @param {string} detail
+ * @returns {void}
+ */
+export default function emitDeprecationWarning(message, codeKey, detail) {
+	const code = CODES[codeKey];
+
+	assertString(code);
+	process.emitWarning(message, { type: 'DeprecationWarning', code: `stylelint:${code}`, detail });
```
Mouvedia, line 24: question

Why wouldn't the details not be mapped to the code keys like the codes are?
Are you expecting using the same code with different details?

ybiquitous: I want to put a `message` and `detail` in the same place because they are strongly related.

`detail` has not been repeated yet, so I don't think it's worth coupling the code and detail for now.

Mouvedia:
Can you add `--no-deprecation` to the `test-only` script?

lib/prepareReturnValue.mjs
```diff
@@ -35,11 +35,11 @@ export default function prepareReturnValue(stylelintResults, maxWarnings, format
 		// TODO: Deprecated. Remove in the next major version.
 		get output() {
 			if (!this._outputWarned) {
```
Mouvedia, line 37: suggestion
minor

Could you change the signature of `prepareReturnValue` to be
```
 * @param {StylelintResult[]} stylelintResults
 * @param {Formatter} formatter
 * @param {object} opts
 * @param {string} opts.cwd
 * @param {maxWarnings} opts.maxWarnings
 * @param {boolean} opts.quietDeprecationWarnings
```
so that we can add `&& !opts.quietDeprecationWarnings` to the condition?

ybiquitous:
> Can you add `--no-deprecation` to the `test-only` script?

What is the purpose/reason?

lib/prepareReturnValue.mjs
ybiquitous: Your suggestion makes sense, but it's out of the scope of this PR, which is just a refactoring.

I want to work on that after merging this PR.

Mouvedia:
> > Can you add `--no-deprecation` to the `test-only` script?
> 
> What is the purpose/reason?

Not having the deprecation warnings outputted in the console when you run the tests.
i.e. less noise and shorter output

ybiquitous:
I understood, but I'm against because it's hard to specify `--no-deprecation` only to `test-only`. Changing `scripts` like this would affect also on `test` and `test-coverage`:

```diff
-"test": "node --experimental-vm-modules node_modules/jest/bin/jest.js", 
+"test": "node --experimental-vm-modules --no-deprecation node_modules/jest/bin/jest.js", 
```

https://github.com/stylelint/stylelint/blob/cd69f4c0078ffa452c909bf2739c4ef9ed1823bd/package.json#L77-L79

Additionally, `--no-deprecation` would hide other warnings than Stylelint, so it would be harder for us to notice some deprecations.

Instead, I recommend specifying `export NODE_OPTIONS='--no-deprecation'` on your environment.

Anyway, this requirement is out of the scope of this PR.

lib/utils/emitDeprecationWarning.mjs
```diff
@@ -0,0 +1,25 @@
+import process from 'node:process';
+
+import { assertString } from './validateTypes.mjs';
+
+// Each code must be unique.
+const CODES = {
```
Mouvedia, line 6: You might be able to get rid of `assertString` if you set a `@type` here.

Mouvedia (approved):

lib/utils/emitDeprecationWarning.mjs
ybiquitous: I found `assertString` was unnecessary. Removed it via 6b4d95ee9. 👍🏼

lib/prepareReturnValue.mjs
ybiquitous: Just opened follow-up PR #7837

lib/utils/emitDeprecationWarning.cjs
```diff
@@ -0,0 +1,28 @@
...
+	COMMONJS_NODEJS_API: '002',
+	RESULT_OUTPUT_PROPERTY: '003',
+};
+
+/**
+ * Use this function for deprecation warnings, instead of `process.emitWarning()`.
+ *
+ * @param {string} message
+ * @param {keyof CODES} codeKey
+ * @param {string} detail
```
Mouvedia, line 19: @ybiquitous we could possibly share [`DeprecatedInfo`](https://jsr.io/@eslint/core@0.11.0/doc/~/DeprecatedInfo) with ESLint.

ybiquitous: I'm not sure about the type of ESLint, but is it possible to pass such more complex info to `process.emitWarning`?

I'm okay with discussing a new issue if there's feasibility.

PR #7835, opened 2024-07-09
Refactor `function-calc-no-unspaced-operator` to have atomic fixes

romainmenke:

Each mutation of the AST now corresponds to a single warning and a single fix.

See : https://github.com/stylelint/stylelint/pull/7831

The order of the warnings changes, but I hope this is fine?

### DIFF
--- a/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
+++ b/lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
@@ -263,32 +263,32 @@ testRule({
 			fixed: 'a { top: calc(2px + 1px); }',
 			description: 'no space before or after operator',
 			warnings: [
-				{ message: messages.expectedBefore('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
 				{ message: messages.expectedAfter('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
+				{ message: messages.expectedBefore('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
 			],
 		},
 		{
 			code: 'a { transform: rotate(acos(2+1)); }',
 			fixed: 'a { transform: rotate(acos(2 + 1)); }',
 			warnings: [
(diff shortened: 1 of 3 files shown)


lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
```diff
@@ -263,32 +263,32 @@ testRule({
 			fixed: 'a { top: calc(2px + 1px); }',
 			description: 'no space before or after operator',
 			warnings: [
-				{ message: messages.expectedBefore('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
 				{ message: messages.expectedAfter('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
+				{ message: messages.expectedBefore('+'), line: 1, column: 18, endLine: 1, endColumn: 19 },
```
ybiquitous, line 267: [ask] Can we prevent this before/after order change?

ybiquitous (commented):
@romainmenke Thank you. Overall, this looks good 👍🏼 

I've left a comment only for the lint problem order change.

lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
ybiquitous: I now understand it seems quite difficult for now because an example of `2px+1px` is initially parsed as `2px` and `+1px` tokens. The algorithm is like this:

1. `2px+1px` → [`2px`, `+1px`] (the first and second operands)
2. [`2px`, `+1px`] → [`2px`, `+`, ` `, `1px`] with the `expectedAfter` message
3. [`2px`, `+`, ` `, `1px`] → [`2px`, ` `, `+`, ` `, `1px`] with the `expectedBefore` message

This is not a significant issue since the number of reported lint problems is unchanged. 👍🏼

ybiquitous (approved):
LGTM 👍🏼

lib/rules/function-calc-no-unspaced-operator/__tests__/index.mjs
romainmenke: Correct :)

Issue #3111, opened 2018-01-08
Fix configuration comments within selector and value lists being ignored


fiskhandlarn:
I was told in https://github.com/rstacruz/stylelint-rscss/issues/11#issuecomment-350542523 that this is a problem in `stylelint` rather than `stylelint-rscss`.

> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

```
    &:checked ~ .single-steward-header, // stylelint-disable-line rscss/class-format
    &:checked ~ .toggle {
      //
    }
```

still triggers an error for the first line: 
`Invalid element name: '.single-steward-header'   rscss/class-format`

ATM you have to do this instead:
```
    // stylelint-disable
    &:checked ~ .single-steward-header,
    // stylelint-enable
    &:checked ~ .toggle {
      //
    }
```

> Which rule, if any, is this issue related to?

`stylelint-disable-line`

> What CSS is needed to reproduce this issue?

See above.

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "extends": [
    "stylelint-rscss/config"
  ]
}
```

> Which version of stylelint are you using?

8.2.0

> How are you running stylelint: CLI, PostCSS plugin, Node API?

Webpack with `stylelint-webpack-plugin`.

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes, SCSS.

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

The following warnings were flagged:

```shell
Invalid element name: '.single-steward-header'   rscss/class-format
```

jeddy3:
@fiskhandlarn Thanks for the report and for using the template.

Unfortunately, this is a known limitation. It's document at the bottom of [this section of the configuration guide](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#turning-rules-off-from-within-your-css). "Comments within _selector and value lists_ are currently ignored.". 

Having said that, digging through the history it looks like [it was an issue](https://github.com/stylelint/stylelint/issues/1628) with postcss@5. stylelint now uses postcss@6. Does anyone [know if this behaviour changed](https://github.com/postcss/postcss/issues/845) in postcss@6? If so, can this limitation be addressed now?

fiskhandlarn:
Ah, I missed that, @jeddy3. Thanks for following up anyway. :)

jeddy3:
Closing as a documented limitation.

[...]

Issue #5348, opened 2021-06-06
Fix multi-line double-slash command comments


jeddy3:
Refs: https://github.com/stylelint/stylelint/issues/4942#issuecomment-841786453 & https://github.com/stylelint/stylelint/pull/5346#discussion_r645994299

This issue only impacts users who are using Less and multi-line double-slash command comments, e.g.:

```scss
// stylelint-disable color-hex-case
// --
// description

a {color: #FFF; }
```

I doubt many, if any, users do this as Less isn't popular and support for multi-line double-slash command comments is a relatively new feature. As the workaround is to simply switch to `/*` style comments for these multi-line commands, I'd say this is a low priority. I'll add it to the `14.0.0` umbrella issue in case it does impact someone and they want to pick it up, but I don't see it as a blocker to releasing.

nwalters512:
Wanted to add a +1 to this. We use a lot of such comments, and we started seeing Stylelint report issues in really weird places. For instance, a block of `//` comments in the middle of a file caused a lint error at the end of the file to be reported on a token in the middle of the comment. Here's a stripped-down version of that file illustrating the issue:

<img width="714" alt="Screen Shot 2021-08-09 at 3 46 41 PM" src="https://user-images.githubusercontent.com/1476544/128783712-5495e5d1-6a08-4ee2-a4e1-4a296f03a611.png">

I finally managed to figure out that the common link between all files with weird errors was the use of multi-line `//` comments and trace it back to this GitHub issues, but it was a reallllly confusing 30 minutes spent banging my head against the table.

Mouvedia:
fixed by #5458

PR #7839, opened 2024-07-14
Fix configuration comments within selector and value lists being ignored

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/3111


lib/assignDisabledRanges.mjs
```diff
@@ -61,101 +73,162 @@ export default function assignDisabledRanges(root, result) {
...
+			const nextComment = node.next();
+
+			// If any of these conditions are not met, do not merge comments.
+			if (
+				!(
+					!isStandardSyntaxComment(node) &&
+					isConfigurationComment(node.text, configurationComment) &&
+					nextComment &&
+					isComment(nextComment) &&
+					(node.text.includes('--') || nextComment.text.startsWith('--'))
```
Mouvedia, line 94: once it's ready to be implemented

Could you put `(node.text.includes('--') || nextComment.text.startsWith('--')` in a `const`?
This would serve as documentation.
i.e. it's an optional disable description
e.g. `hasDisableDescription`

medium
fix

> There must be a space on both sides of the hyphens.

Hence, it should be `' -- '`, not `'--'`.

romainmenke: This is unrelated to this change :)
Only the indent changed and variable names changed, I didn't make any logic changes to this part.

Do you mind filing a separate issue for this?

Mouvedia: > Do you mind filing a separate issue for this?

If it doesn't get the `ready to implement` label, I will.
Else you could just add another changeset for the `' -- '` fix.

romainmenke: I prefer not to make that change here, as it is completely unrelated.

lib/__tests__/disableRanges.test.mjs
```diff
@@ -953,8 +1062,7 @@ it('Less // line-disabling comment (with description)', async () => {
 	});
 });
 
-// eslint-disable-next-line jest/no-disabled-tests -- see #5348
```
Mouvedia, line 956: This issue was fixed by #5458 in version 14.0.0-0.
i.e. one day before the release of 14.0.0
We can close #5348.

romainmenke: Thank you for digging this up @Mouvedia 🙇

.changeset/pretty-beans-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: stylelint-disable-line commands within selector and value lists being ignored
```
romainmenke, line 5:
```suggestion
Fixed: stylelint commands within selector and value lists being ignored
```

lib/__tests__/disableRanges.test.mjs
```diff
@@ -202,22 +203,58 @@ it('disable single line all rules', async () => {
...
-	const result = await testDisableRanges(
-		`a, /* stylelint-disable-line */
-		b /* stylelint-disable-line */ {} `,
-	);
+	const result = await testDisableRanges(stripIndent`
+		/* offset by one line */
+		a,/* stylelint-disable-line */
+		b/* stylelint-disable-line */,
+		c,/* stylelint-disable-line */
+		long-selector/* stylelint-disable-line */,
```
Mouvedia, line 211: nit
suggestion

Type selectors don't contain hyphens.
Maybe rename to .long-selector?

lib/assignDisabledRanges.mjs
```diff
@@ -147,7 +147,7 @@ export default function assignDisabledRanges(root, result) {
 		}
 
 		if (isAtRule(node)) {
-			let offset = node.name.length + 1;
+			let offset = node.name.length + 1; // `@` + name
```
Mouvedia, line 150: nit
style

In those cases I prefer putting the documentation in the code itself.
e.g. `'@'.length + node.name.length`

lib/assignDisabledRanges.mjs
```diff
@@ -268,34 +345,58 @@ export default function assignDisabledRanges(root, result) {
...
-			throw comment.error(`"${ruleToEnable}" has not been disabled`, {
+			throw node.error(`"${ruleToEnable}" has not been disabled`, {
 				plugin: 'stylelint',
 			});
 		}
 	}
 
 	/**
-	 * @param {PostcssComment} comment
+	 * @param {PostcssNode} node
```
Mouvedia, line 355: minor

The type could be more specific.
e.g. `{AtRule|Rule|Declaration}`

lib/assignDisabledRanges.mjs
```diff
@@ -268,34 +345,58 @@ export default function assignDisabledRanges(root, result) {
...
+				processDisableNextLineCommand(node, source, text);
 				break;
 			case DISABLE_COMMAND:
-				processDisableCommand(comment);
+				processDisableCommand(node, source, text);
 				break;
 			case ENABLE_COMMAND:
-				processEnableCommand(comment);
+				processEnableCommand(node, source, text);
 				break;
```
Mouvedia, line 400: question

Is it OK if the `node` passed to these functions is not a `Comment`?

lib/__tests__/disableRanges.test.mjs
romainmenke: Could be a custom element, this is perfectly valid CSS/HTML :)

lib/assignDisabledRanges.mjs
romainmenke: But that is a massive increase in computational effort.

I now that the absolute effort is still small, but still feels wrong to waste CPU cycles when a small code comment can serve the same purpose.

lib/assignDisabledRanges.mjs
romainmenke: Why would it not be ok?
Can you elaborate, I don't know which cases you are thinking of :)

lib/assignDisabledRanges.mjs
romainmenke: Is there a need to make it more specific?

We are using this value to throw errors `PostCSS`, the common type that can throw errors is `postcss.Node`.

lib/assignDisabledRanges.mjs
Mouvedia: I was just checking if passing `AtRule|Rule|Declaration` to `checkComment` was intended or not.

lib/assignDisabledRanges.mjs
Mouvedia: > Is there a need to make it more specific?

It sometimes help to surface potential bugs down the line.
It's minor anyway.

lib/assignDisabledRanges.mjs
romainmenke: Yes :)

If you read through the changes you will see that it is needed to separate these things:
- the source location
- the comment text
- error reporting

So pass three values to these functions, each dedicated to a specific purpose.

lib/assignDisabledRanges.mjs
```diff
@@ -61,101 +73,162 @@ export default function assignDisabledRanges(root, result) {
...
-		let lastLine = (comment.source && comment.source.end && comment.source.end.line) || 0;
-		const fullComment = comment.clone();
+		if (isRule(node)) {
+			let offset = 0;
+			const selector = node.raws?.selector?.raw;
 
-		let current = nextComment;
+			if (selector && containsComments(selector)) {
+				checkCommentContainingNode(node, selector, offset);
+			}
```
ybiquitous, line 139: [suggest] This combination of `containsComments` and `checkCommentContainingNode` often appears, so I feel we can refactor like this: 👇🏼 

```suggestion
			checkCommentContainingNode(node, selector, offset);
```

Inlining `containsComments()`, e.g.

```js
/**
 * @param {PostcssNode} node
 * @param {string | undefined} part
 * @param {number} offset
 */
function checkCommentContainingNode(node, part, offset) {
	if (!(part && part.includes('/*') && part.includes('*/'))) return;

	// ...
}
```

lib/assignDisabledRanges.mjs
```diff
@@ -61,101 +73,162 @@ export default function assignDisabledRanges(root, result) {
...
-			!(
-				!isStandardSyntaxComment(comment) &&
-				isConfigurationComment(comment, configurationComment) &&
-				nextComment &&
-				nextComment.type === 'comment' &&
-				(comment.text.includes('--') || nextComment.text.startsWith('--'))
-			)
-		) {
-			checkComment(comment);
+			let lastLine = (node.source && node.source.end && node.source.end.line) || 0;
```
ybiquitous, line 102: [suggest] Minor.

```suggestion
			let lastLine = node.source?.end?.line ?? 0;
```

lib/assignDisabledRanges.mjs
```diff
@@ -61,101 +73,162 @@ export default function assignDisabledRanges(root, result) {
...
+			const fullComment = node.clone();
 
-			return;
+			let current = nextComment;
+
+			while (
+				!isStandardSyntaxComment(current) &&
+				!isConfigurationComment(current.text, configurationComment)
+			) {
+				const currentLine = (current.source && current.source.end && current.source.end.line) || 0;
```
ybiquitous, line 111: [suggest] Minor.

```suggestion
				const currentLine = current.source?.end?.line ?? 0;
```

lib/assignDisabledRanges.mjs
```diff
@@ -268,34 +345,58 @@ export default function assignDisabledRanges(root, result) {
...
+	 * @param {string} part
+	 * @param {number} offset
+	 */
+	function checkCommentContainingNode(node, part, offset) {
+		tokenize({ css: part })
+			.filter((token) => {
+				return isTokenComment(token);
+			})
+			.forEach((token) => {
+				const [, text, start, end] = token;
```
ybiquitous, line 365: [suggest] Minor. For the performance, reducing the loop count may be a bit more effective:

```suggestion
		tokenize({ css: part })
			.forEach((token) => {
				if (!isTokenComment(token)) return;

				const [, text, start, end] = token;
```

lib/assignDisabledRanges.mjs
```diff
@@ -268,34 +345,58 @@ export default function assignDisabledRanges(root, result) {
...
+			});
+	}
+
+	/**
+	 * @param {PostcssNode} node
+	 * @param {Source | undefined} source
+	 * @param {string} text
 	 */
-	function checkComment(comment) {
+	function checkComment(node, source, text) {
```
ybiquitous, line 381: [ask] Is the following early return for `source` effective? It seems that we can skip `processDisable*()` function calls if `source` is undefined. 🤔 

```suggestion
	function checkComment(node, source, text) {
		if (!source) return;
```

lib/assignDisabledRanges.mjs
```diff
@@ -10,23 +10,27 @@ import {
...
 import { assert, assertNumber, assertString } from './utils/validateTypes.mjs';
+import { isAtRule, isComment, isDeclaration, isRule } from './utils/typeGuards.mjs';
 import isStandardSyntaxComment from './utils/isStandardSyntaxComment.mjs';
 
-/** @import {Comment as PostcssComment, Document as PostcssDocument, Root as PostcssRoot} from 'postcss' */
+import { isTokenComment, tokenize } from '@csstools/css-tokenizer';
+
+/** @import {Node as PostcssNode, Comment as PostcssComment, Document as PostcssDocument, Root as PostcssRoot, Position as PostcssPosition} from 'postcss' */
 /** @import {DisabledRange, DisabledRangeObject, PostcssResult} from 'stylelint' */
+/** @typedef {{start?: PostcssPosition, end?: PostcssPosition}} Source */
```
ybiquitous, line 20: [suggest] Minor. Using the `Source` type may be a bit clearer. But you can keep it as-is.

```suggestion
/** @import {Node as PostcssNode, Comment as PostcssComment, Document as PostcssDocument, Root as PostcssRoot, Source as PostcssSource} from 'postcss' */
/** @import {DisabledRange, DisabledRangeObject, PostcssResult} from 'stylelint' */
/** @typedef {Pick<PostcssSource, 'start' | 'end'>} Source */
```

ybiquitous (commented):
@romainmenke Thanks for the pull request. Overall LGTM. 👍🏼 

I've left some reviews about readability and performance. Please take a look. 🙏🏼

.changeset/pretty-beans-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: stylelint commands within selector and value lists being ignored
```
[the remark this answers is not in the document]
Mouvedia: Romain was probably using [this option](https://stylelint.io/user-guide/rules/comment-empty-line-before/#stylelint-commands) as a precedent.
Interestingly the description says "configuration comments".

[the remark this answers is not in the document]
romainmenke: No, I used the original issue title :)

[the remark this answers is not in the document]
ybiquitous: The *"configuration command"* was introduced in 15.3.0 (see 4fadc6c7184efe6e516c8ceae28081b6f2ef6517). Perhaps, that secondary option may have been added before that.

lib/assignDisabledRanges.mjs
romainmenke: yeah, that makes sense :)

lib/assignDisabledRanges.mjs
romainmenke: Good call, thank you

romainmenke:
Thank you for the review and suggestions 🙇

I've applied the requested changes and also renamed `checkCommentContainingNode` to `checkCommentsInNode`. I think this is clearer after the latest changes.

lib/assignDisabledRanges.mjs
```diff
@@ -61,101 +65,148 @@ export default function assignDisabledRanges(root, result) {
...
+			checkCommentsInNode(node, value, offset);
+		}
 	});
 
 	return result;
 
 	/**
-	 * @param {PostcssComment} comment
+	 * @param {PostcssNode} node
+	 * @param {Source | undefined} source
```
ybiquitous, line 171: [suggest] Thanks for refactoring! Furthermore, we can simplify the `source` argument type like this, right?

```suggestion
	 * @param {Source} source
```

If so, using early return could make the code a bit more readable, I think.

```js
if (!source.start) return;

const line = source.start.line;
// ...
```

I think this could be applied to other `processDisable*` functions.

romainmenke: good catch 🙇

lib/__tests__/disableRanges.test.mjs
Mouvedia, line 1: question

Is _next-line_ handled too?
e.g.

```css
:is(
  div
/* stylelint-disable-next-line selector-type-case */) { color: green; }
SPAN { color: red; }
```

romainmenke: Yes, no changes were made to how configuration comments work :)
We only detect and gather more comments.

lib/assignDisabledRanges.mjs
```diff
@@ -61,101 +65,148 @@ export default function assignDisabledRanges(root, result) {
...
+	 * @param {PostcssNode} node
+	 * @param {Source} source
+	 * @param {string} text
 	 */
-	function processDisableLineCommand(comment) {
-		if (comment.source && comment.source.start) {
-			const line = comment.source.start.line;
-			const description = getDescription(comment.text);
+	function processDisableLineCommand(node, source, text) {
+		if (source?.start) {
```
ybiquitous, line 175:
```suggestion
		if (source.start) {
```

lib/assignDisabledRanges.mjs
```diff
@@ -61,101 +65,148 @@ export default function assignDisabledRanges(root, result) {
...
+	 * @param {PostcssNode} node
+	 * @param {Source} source
+	 * @param {string} text
 	 */
-	function processDisableNextLineCommand(comment) {
-		if (comment.source && comment.source.end) {
-			const line = comment.source.end.line;
-			const description = getDescription(comment.text);
+	function processDisableNextLineCommand(node, source, text) {
+		if (source?.end) {
```
ybiquitous, line 191:
```suggestion
		if (source.end) {
```

lib/assignDisabledRanges.mjs
```diff
@@ -201,27 +254,29 @@ export default function assignDisabledRanges(root, result) {
 				);
 			}
 
-			if (comment.source && comment.source.start) {
-				const line = comment.source.start.line;
+			if (source?.start) {
```
ybiquitous, line 257:
```suggestion
			if (source.start) {
```

lib/assignDisabledRanges.mjs
```diff
@@ -201,27 +254,29 @@ export default function assignDisabledRanges(root, result) {
...
+	 * @param {Source} source
+	 * @param {string} text
 	 */
-	function processEnableCommand(comment) {
-		for (const ruleToEnable of getCommandRules(ENABLE_COMMAND, comment.text)) {
+	function processEnableCommand(node, source, text) {
+		for (const ruleToEnable of getCommandRules(ENABLE_COMMAND, text)) {
 			// need fallback if endLine will be undefined
-			const endLine = comment.source && comment.source.end && comment.source.end.line;
+			const endLine = source?.end?.line;
```
ybiquitous, line 279:
```suggestion
			const endLine = source.end?.line;
```

ybiquitous (commented):
[suggest] We can remove unneeded `?.`, right?

ybiquitous (approved):
Thank you. LGTM 👍🏼

Mouvedia:
@romainmenke Just curious, is this

https://github.com/stylelint/stylelint/blob/1232b539dfebc7a14aaa4b2d99af3af96114d776/lib/rules/comment-whitespace-inside/README.md?plain=1#L14-L15

related to this change?

romainmenke:
Related in that they are conceptually the same issue, yes :)
But completely different code and features.

Those notes could be removed by updating those rules in a similar way.

lib/assignDisabledRanges.mjs
Mouvedia: The regex would be too complicated because javascript doesn't support the `\Z` metacharacter.

Issue #7846, opened 2024-07-16
Fix `no-descending-specificity` false positives for nested rules without declarations


romainmenke:
### What minimal example or steps are needed to reproduce the bug?

```css
.foo {
	& h2 {
		& + p {
			color: blue;
		}
	}
}

h1 {
	& h2 {
		& + #baz {
			color: green;
		}
	}
}
```

### What minimal configuration is needed to reproduce the bug?

```json
{
  "extends": [
    "stylelint-config-standard"
  ]
}
```

### How did you run Stylelint?

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0AzCEAEwA6A7bALgGQYAWATFnoYSQNQYAOV+BNBkANhAE7wYAjTgFcYAbmpsCAX0mzc8vKQCMLWmUo5W6hgGIBAQwBeaqYS69+Acx4wYuCdply80kABoQKAJacYAOQMAWzhEGAAPYMY-JDAAZziPcAhcHysEEC0MDGwQCIJ7KDjc-gBtPGzs3LiCAE8-Tm9cAgBaSFTvKxaag1woAx4oXIqMAF1XJPa0gDFeIIMCDIArOJSk2EZExCyckBr6mEbmkt3OBZga3PcR6rqGptapzu6CXv7Bk9yzgsuQCekgA)

### Which Stylelint-related dependencies are you using?

```json
{
  "stylelint": "latest",
  "stylelint-config-standard": "latest"
}
```

### What did you expect to happen?

No error to be reported

### What actually happened?

> ERROR
Expected selector "& h2" to come before selector ".foo h2" [(no-descending-specificity)](https://stylelint.io/user-guide/rules/no-descending-specificity) [10:2-6]

### Do you have a proposal to fix the bug?

Only declarations are affected by specificity.
Nested rules are not. The combined selector of nested and parent rules do have specificity. But when a parent **rule** only contains other rules it never has any effect by itself.

-----

With nesting it possible to have parent rules with lower specificity and child rules with much higher specificity.

- `.foo h2` has specificity `[0, 1, 1]` -> highest
- `h1 h2` has specificity `[0, 0, 2]`

While the children are the inverse:
- `.foo h2 + p` has specificity `[0, 1, 2]`
- `h1 h2 + #baz` has specificity `[1, 0, 2]` -> highest

We should only warn about specificity when this actually matters, when a rule contains declarations.

-----

[...]

PR #7850, opened 2024-07-17
Fix `no-descending-specificity` false positives for nested rules without declarations

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/7846

No, it's self-explanatory.


lib/rules/no-descending-specificity/README.md
Mouvedia, line 1: minor
fix

Check line 28 and 30.

lib/rules/no-descending-specificity/index.mjs
```diff
@@ -151,6 +157,21 @@ function lastCompoundSelectorWithoutPseudoClasses(selectorNode) {
...
 }
 
+/**
+ * Specificity only has an effect on declarations.
+ * We only want to check rules that contain declarations either directly or in nested at-rules.
+ *
+ * @param {import('postcss').Container} node
+ * @returns {boolean}
+ */
+function nodeContainsDeclarations(node) {
```
Mouvedia, line 167: nit
style

`hasDeclaration`
i.e. only one is enough

lib/rules/no-descending-specificity/index.mjs
```diff
@@ -151,6 +157,21 @@ function lastCompoundSelectorWithoutPseudoClasses(selectorNode) {
...
+ * We only want to check rules that contain declarations either directly or in nested at-rules.
+ *
+ * @param {import('postcss').Container} node
+ * @returns {boolean}
+ */
+function hasDeclaration(node) {
+	if (!node.nodes) return false;
+
+	return node.some((child) => {
+		return isDeclaration(child) || (isAtRule(child) && hasDeclaration(child));
```
Mouvedia, line 171: question
minor

Is `from { margin-top: 50px; }` considered a declaration?
i.e. inside an `@keyframes`

lib/rules/no-descending-specificity/__tests__/index.mjs
Mouvedia, line 1: minor
non-blocking

I thought of a weird edge case: rule-less layers.
i.e. `a { @layer foo; }`
Could you add skipped test with a `description` to show its not supported?
e.g. `description: not supported because it's useless`

lib/rules/no-descending-specificity/index.mjs
romainmenke: Yes, this is a declaration

lib/rules/no-descending-specificity/__tests__/index.mjs
romainmenke: I don't understand.
Why would we add a skipped test for this, why is it not supported?

Mouvedia: So `foo;` is also considered a declaration? Interesting.
I would have thought it would be an identifier list.

ref: https://drafts.csswg.org/css-cascade-5/#typedef-layer-name

romainmenke: No, it isn't a declaration. It's an at rule.

But I don't understand why you say it isn't supported or why you think it is useless :)

Name defining at rules loke `@layer` aren't affected by specificity.

There is no difference between these:

```css
@layer foo;

a {}
```

```css
a { @layer foo; }
```

```css
#a { @layer foo; }
```

Naming defining at rules are only affected by source order and cascade layers themselves.

Mouvedia: > No, it isn't a declaration. It's an at rule.

So `(isAtRule(child) && hasDeclaration(child)` returns `false` and it was intended.

romainmenke: Indeed :)

Currently the CSS specification does not have any at rules that define things (like `@layer foo;` defines a layer named `foo`) and where the definition is affected by specificity of parent rules.

This is unlikely (but not impossible) to change in the future.

It is extremely complex for browsers to implement named things like keyframes, cascade layers, fonts, that are effected by matched elements and specificity.

ybiquitous (approved):
Thank you, nice fix 👍🏼 

`hasDeclaration()` could be put in `lib/utils/` if it will be used by other rules. 😃

Issue #7855, opened 2024-07-18
Fix `selector-type-*`, `selector-max-type` and `selector-class-pattern` false positives for `<timeline-range-name>` in keyframe selectors


romainmenke:
Scroll driven animations extended the valid syntax for keyframe selectors:
- https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes
- https://drafts.csswg.org/scroll-animations-1/#named-timeline-range

### What minimal example or steps are needed to reproduce the bug?

```css
@keyframes animate-in-and-out {
	entry 0%  {
		opacity: 0; transform: translateY(100%);
	}

	entry 100%  {
		opacity: 1; transform: translateY(0);
	}

	exit 0% {
		opacity: 1; transform: translateY(0);
	}
	
	exit 100% {
		opacity: 0; transform: translateY(-100%);
	}
}
```

### What minimal configuration is needed to reproduce the bug?

```json
{
  "extends": [
    "stylelint-config-standard"
  ]
}
```

### How did you run Stylelint?

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIACBrGBPAZgJwIYFsYBnAAlwDsBLfXAFxgFpLyGKoGIBXWk4AHXJ965WtnQkADAFISvAUKEQADrjCVa6eJIDcJURSKYI2fFv3kiAGzowAmgAoAjBOkBKbfNoBfeYOGjxZ2lZfj9FFTUNLUddc0NjUz08C2t6Bwl3Tx9BPxgAD3VJGVCFWmVVdU0SGKSDIxMzZKsbdMy-bIVcgp4g4s9wiqidWot6xLjUu3sGXrahbK8QABoQTEpLGAA5AjhEfIIlDYA6MCIiZfAIcjWAcwQQUNk+EHzhKCJnrQBtAVknkCIGg2lmYtAYkGulBuDEBbFw2Cgz1+JAAugJFisIbcAGIJOj3ABWRCuF1gSnOiEeJGegPQwNBn2pIEmgOeS2RNKBMBBInBV1uMNocIRjOeLNoSPIiy8QA)

### Which Stylelint-related dependencies are you using?

```json
{
  "stylelint": "latest",
  "stylelint-config-standard": "latest"
}
```

### What did you expect to happen?

No problems to be reported

### What actually happened?

[...]

PR #7856, opened 2024-07-18
Fix `selector-type-*`, `selector-max-type` and `selector-class-pattern` false positives for `<timeline-range-name>` in keyframe selectors

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/7855

This is a blocker for https://github.com/stylelint/stylelint/pull/7856#discussion_r1682845978.


.changeset/afraid-crabs-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-*` false positives for `timeline-range-name` in `keyframe` selectors
```
Mouvedia, line 5:
```suggestion
Fixed: `selector-type-*`, `selector-max-type` and `selector-class-pattern` false positives for `<timeline-range-name>` in keyframe selectors
```

lib/reference/keywords.mjs
```diff
@@ -286,6 +286,16 @@ export const camelCaseKeywords = new Set([
 
 export const keyframeSelectorKeywords = new Set(['from', 'to']);
 
+// https://drafts.csswg.org/scroll-animations-1/#view-progress-timelines
+export const namedTimelineRangeKeywords = new Set([
```
Mouvedia, line 290: Can you export this set in `keyframe-selector-notation/index.mjs` and replace

https://github.com/stylelint/stylelint/blob/e7e54941b9376bc63d5c7343c1d1ef56eb0ab04c/lib/rules/keyframe-selector-notation/index.mjs#L20

romainmenke: I think we best do that in a separate PR.

The list used by `keyframe-selector-notation` is either wrong or outdated.
`enter` does not exist (anymore) and `entry-crossing`, `exit-crossing` do exist.

Fixing that is out of scope.

Mouvedia: We will have to wait for this PR to be merged then.

lib/utils/isKeyframeSelector.mjs
```diff
@@ -1,4 +1,9 @@
-import { keyframeSelectorKeywords } from '../reference/keywords.mjs';
+import { keyframeSelectorKeywords, namedTimelineRangeKeywords } from '../reference/keywords.mjs';
+
+const HAS_TIMELINE_RANGE = new RegExp(
+	`^(?:${[...namedTimelineRangeKeywords.values()].join('|')})\\s+(?:\\d+|\\d*\\.\\d+)%$`,
```
Mouvedia, line 4: trivial
note

I wonder if browser support unusual `<number>`s.
e.g. `10e3%`, `-3.4e-2%`

romainmenke: Out of scope?

I doubt that anyone writes CSS likes this If anyone does, then they can open an issue and let us know :)

Mouvedia, line 4: note

This requires the selector to be trimmed before being passed.

.changeset/afraid-crabs-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-type-*`, `selector-max-type` and `selector-class-pattern` false positives for `<timeline-range-name>` in keyframe selectors
```
Mouvedia, line 5: question
minor

Does it underline the percentage too?
i.e. Should it be "…for `<timeline-range-name> <percentage>` in…"?

Mouvedia (approved):

lib/utils/isKeyframeSelector.mjs
romainmenke: Correct, like the other checks in this function

.changeset/afraid-crabs-ring.md
romainmenke: I guess it depends on the rule?
I think the current text is descriptive enough to communicate what was fixed.

ybiquitous (approved):
@romainmenke Thanks. LGTM 👍🏼 

As a bonus, it might be great to add a test case to each rule affected by this PR, so that users might understand this bug fix more easily. Of course, this is not required, though. We can also merge this as-is. 😃

romainmenke:
I've added some extra tests, also good to have those independent of the helper :)
Thank you

Issue #7838, opened 2024-07-12
New repository for testing ecosystem


ybiquitous:
Hello Stylelint folks, 👋🏼 

This issue aims just to share information, not for any feature requests or bug fixes.

Before releasing 16.7.0 in #7824, I created a repository to test a bunch of plugins and configs with the new version:
https://github.com/ybiquitous/stylelint-ecosystem-tester

See also the current test results:
https://github.com/ybiquitous/stylelint-ecosystem-tester/actions/runs/9903754245

There are still many errors, but most are unrelated to Stylelint. This repository doesn't aim to resolve *all* errors but to observe our ecosystem situation. I will try to fix as many errors as possible, though.

If there are no concerns or problems, I would like to move this repository under the Stylelint organization so that more people, like maintainers or plugin authors, can efficiently find any issues.

I would appreciate it if you would be interested in this topic and provide any feedback. 😃

romainmenke:
No concerns!

I really like this way of unreleased updates.
Thank you for setting this up @ybiquitous 🙇

Mouvedia:
> If there are no concerns or problems, I would like to move this repository under the Stylelint organization so that more people, like maintainers or plugin authors, can efficiently find any issues.

👍 

> provide any feedback

1.
Maybe add criteria related to package addition.
i.e. if someone request his package to be added you need your refusal to be based on/justified by something already mentioned in the README

2.
add popular formatters

3.
This new repository goes hand in hand with #2498.

ybiquitous:
Thanks for the feedback. I agree with adding such a criteria. 👍🏼 If you're interested, creating a PR that updates the README would be great.

ybiquitous:
I've added criteria. See https://github.com/ybiquitous/stylelint-ecosystem-tester#criteria

Mouvedia:
> I've added criteria.

👍 

> was published within three years

Most of the formatters won't fit the bill because they often don't have any dependencies; they don't need to be updated often.

> move this repository under the Stylelint organization

Do it once you are satisfied with the list.

ybiquitous:
I relaxed the criteria a bit: https://github.com/ybiquitous/stylelint-ecosystem-tester#criteria

ybiquitous:
Once I arrange the packages based on the criteria, I'll move the repo under the <https://github.com/stylelint> org.

Please let me know if you have any concerns.

Mouvedia:
[...]

PR #7857, opened 2024-07-18
Encourage using common keywords for plugins, configs, and formatters in documentation

ybiquitous:

Related to #7838

These keywords would help us find packages more easily.

Common keywords: `stylelint-plugin`, `stylelint-config`, `stylelint-formatter`

See:
- https://www.npmjs.com/search?q=keywords%3Astylelint-plugin
- https://www.npmjs.com/search?q=keywords%3Astylelint-config
- https://www.npmjs.com/search?q=keywords%3Astylelint-formatter

### DIFF
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -362,4 +362,32 @@ To make a single module provide multiple rules, export an array of plugin object
 
 ## Sharing plugins and plugin packs
 
-Use the `stylelint-plugin` keyword within your `package.json`.
+Use the `stylelint-plugin` keyword within your `package.json`. For example:
+
+```json
+{
+  "keywords": ["stylelint", "stylelint-plugin"]
+}
+```
+
+## Sharing configurations
+
+You may want to create a shareable configuration for your plugins to provide a default ruleset. For example:
+
+```json
+{
+  "plugins": ["stylelint-plugin-foo"],
+  "rules": {
+    "foo/rule-1": true,
+    "foo/rule-2": true
+  }
+}
+```
+
+When publishing such configurations, use the `stylelint-config` keyword within your `package.json`. For example:
+
+```json
+{
+  "keywords": ["stylelint", "stylelint-config"]
+}
+```
--- a/docs/developer-guide/formatters.md
+++ b/docs/developer-guide/formatters.md
@@ -83,3 +83,13 @@ stylelint -f json "*.css" 2>&1 | my-program-that-reads-JSON --option
 ## `stylelint.formatters`
 
 Stylelint's internal formatters are exposed publicly in `stylelint.formatters`.
+
+## Sharing formatters
+
+Use the `stylelint-formatter` keyword within your `package.json`. For example:
+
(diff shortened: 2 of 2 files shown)


romainmenke (approved):
LGTM!

----

In our test framework for PostCSS plugins we also check `package.json` for the presence of `postcss-plugin` in keywords.

https://github.com/csstools/postcss-plugins/blob/e8809370ba61f406c3063b7a7566faf3223c4aae/packages/postcss-tape/src/index.ts#L98-L141

Is there an analogue to something like that for Stylelint?

We used to forget and overlook things like the keywords but having a test for it makes it easier.

docs/developer-guide/plugins.md
```diff
@@ -362,4 +362,32 @@ To make a single module provide multiple rules, export an array of plugin object
...
-Use the `stylelint-plugin` keyword within your `package.json`.
+Use the `stylelint-plugin` keyword within your `package.json`. For example:
+
+```json
+{
+  "keywords": ["stylelint", "stylelint-plugin"]
+}
+```
+
+## Sharing configurations
```
Mouvedia, line 373: We will eventually need a dedicated `docs/developer-guide/config.md` page for _Writing configurations_.

docs/developer-guide/plugins.md
```diff
@@ -362,4 +362,32 @@ To make a single module provide multiple rules, export an array of plugin object
...
+
+You may want to create a shareable configuration for your plugins to provide a default ruleset. For example:
+
+```json
+{
+  "plugins": ["stylelint-plugin-foo"],
+  "rules": {
+    "foo/rule-1": true,
```
Mouvedia, line 381: question
minor

Should be `stylelint-plugin-foo/rule-x`?

docs/developer-guide/plugins.md
ybiquitous: Probably yes. You can write it.

docs/developer-guide/plugins.md
ybiquitous: Actually, we don't need to keep consistency between a package name and a rule namespace. So, I think it's good enough as it is.

E.g.

https://github.com/stylelint/stylelint/blob/67c21375e12f695a3929bde27281d772dcdeaf93/docs/developer-guide/plugins.md?plain=1#L27-L35

ybiquitous:
@romainmenke Thanks for the review!

> Is there an analogue to something like that for Stylelint?
> 
> We used to forget and overlook things like the keywords but having a test for it makes it easier.

As far as I know, there is no such thing, but that sounds great to me. 😃 
If we provide such a feature, [stylelint-test-rule-node](https://github.com/stylelint/stylelint-test-rule-node) and [jest-preset-stylelint](https://github.com/stylelint/jest-preset-stylelint) seems appropriate to me.

Or do you have a better idea about this?

romainmenke:
I filed an issue here : https://github.com/stylelint/jest-preset-stylelint/issues/110 😄

PR #7859, opened 2024-07-18
Fix `keyframe-selector-notation` keywords for `<timeline-range-name>`

romainmenke:

See: https://github.com/stylelint/stylelint/pull/7856#discussion_r1682845978

I looked around but I couldn't find any trace of `enter` as a named timeline range keyword.
It might have existed in an earlier version of the spec.

This change moves the list of keywords to `reference/keywords.mjs` and updates `keyframe-selector-notation` to match the latest list.

----

This PR will be merged after https://github.com/stylelint/stylelint/pull/7856#discussion_r1682845978


.changeset/angry-spies-allow.md
```diff
@@ -0,0 +1,6 @@
+---
+"stylelint": patch
+---
+
+Fixed: `keyframe-selector-notation` false positives for `entry-crossing` and `exit-crossing`
+Fixed: `keyframe-selector-notation` false negatives for `enter`
```
romainmenke, line 6: Can we do two entries?
I vaguely remember that changeset has an issue with this.

Mouvedia: I tested `parse.default` using https://npm.runkit.com/%40changesets%2Fparse; it worked as intended.
i.e. produces `{releases: [{name: "stylelint", type: "patch"}], summary: "Fixed: foo\nFixed: bar"}`

Mouvedia:
> I looked around but I couldn't find any trace of enter as a named timeline range keyword.
> It might have existed in an earlier version of the spec.

https://github.com/stylelint/stylelint/pull/6605#discussion_r1097029879

romainmenke:
Ah, yes there was a note:

>  Should this be `entry` or `enter`?

Mouvedia (approved):

.changeset/angry-spies-allow.md
```diff
@@ -0,0 +1,6 @@
+---
+"stylelint": patch
```
romainmenke, line 2:
```suggestion
"stylelint": minor
```

lib/rules/keyframe-selector-notation/__tests__/index.mjs
```diff
@@ -14,14 +14,18 @@ const ACCEPT_TIMELINE_RANGE_NAME_TEST_CASES = [
 		code: '@keyframes foo { entry 0% {} entry 100% {} }',
 		description: 'timeline-range-name entry',
 	},
-	{
-		code: '@keyframes foo { enter 0% {} enter 100% {} }',
-		description: 'timeline-range-name enter',
-	},
```
romainmenke, line 20: I did not add a test with a warning for `enter 0% {}` because this rule does not aim to validate the named timeline ranges, only `from` vs. `0%`.

I think it is better to leave this as undefined behavior for now.

.changeset/angry-spies-allow.md
Mouvedia: @romainmenke check https://github.com/stylelint/stylelint/pull/7867#discussion_r1686023873

lib/rules/keyframe-selector-notation/__tests__/index.mjs
romainmenke: 🤔 Maybe better to omit it entirely from the changelog?

No one is writing `enter 10%` and this rule doesn't validate the named timeline ranges themselves.

I added it to the changelog before I fully understood why `enter` was allowed in this rule. Having a better understanding I think we can omit it.

-----

`enter` was at one point considered by the specification editors.
But as far as I know it was never implemented. So I think we can assume that no one is writing `@keyframes foo { enter 0% {} enter 100% {} }`.

ybiquitous: Understood. Thanks for the explanation. So, let's omit `enter` from the changelog. 👍🏼

ybiquitous (approved):
Nice to reuse the references. 👍🏼 

Once updating the changelog, feel free to merge it.

.changeset/angry-spies-allow.md
```diff
@@ -0,0 +1,6 @@
+---
+"stylelint": minor
+---
+
+Fixed: `keyframe-selector-notation` false positives for `entry-crossing` and `exit-crossing`
+Fixed: `keyframe-selector-notation` false negatives for `enter`
```
romainmenke, line 6:
```suggestion
```

romainmenke:
Thank you for the reviews everyone 🙇

Issue #7860, opened 2024-07-18
CI is failing on Windows with `Error: Cannot find module '..'`


romainmenke:
see : https://github.com/nodejs/node/issues/53922, nodejs/node#53902

We might need to pin node 22 to 22.4 or lower if this issue is blocking progress.

romainmenke:
53922 was marked as a duplicate of : https://github.com/nodejs/node/issues/53902

romainmenke:
Almost there, waiting on : https://github.com/actions/node-versions/pull/182

ybiquitous:
Let's wait for a new release of actions/node-versions. We can allow failures in only one CI job for Windows and Node.js 22 for a while.

ybiquitous:
Node.js 22.5.1 is now available on actions/setup-node. The retried CI passed: ✅ 
https://github.com/stylelint/stylelint/actions/runs/10035926924/job/27742175426

PR #7862, opened 2024-07-18
Refactor to remove `Range` from the return type of the `fix` callback

Mouvedia:

https://github.com/stylelint/stylelint/pull/7819#issuecomment-2210064821

#7860

### DIFF
--- a/lib/utils/report.mjs
+++ b/lib/utils/report.mjs
@@ -160,7 +160,11 @@ function isDisabled(ruleName, startLine, disabledRanges) {
 function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
 	const { disabledRanges, config = {}, fixersData } = stylelint;
 
-	if (!isFn(fix)) return addFixData({ fixersData, ruleName, fixed: false }), false;
+	if (!isFn(fix)) {
+		addFixData({ fixersData, ruleName, fixed: false });
+
+		return false;
+	}
 
 	const shouldFix = Boolean(config.fix && !config.rules?.[ruleName][1]?.disableFix);
 	const mayFix =
@@ -168,9 +172,11 @@ function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
 
 	addFixData({ fixersData, ruleName, fixed: mayFix });
 
-	if (mayFix) return fix(), true;
+	if (!mayFix) return false;
 
-	return false;
+	fix();
+
+	return true;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


lib/utils/report.mjs
```diff
@@ -163,23 +160,15 @@ function isDisabled(ruleName, startLine, disabledRanges) {
 function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
 	const { disabledRanges, config = {}, fixersData } = stylelint;
 
-	if (!isFn(fix)) {
-		addFixData({ fixersData, ruleName, fixed: false });
-
-		return false;
-	}
+	if (!isFn(fix)) return addFixData({ fixersData, ruleName, fixed: false }), false;
```
ybiquitous, line 163: [suggest] Can you restore the original form? Because:

- `return addFixData(...)` looks like the `addFixData()` function would return something regardless nothing.
- The `return x, y` pattern is uncommon in this code base: let's keep it less surprising.

```suggestion
	if (!isFn(fix)) {
		addFixData({ fixersData, ruleName, fixed: false })

		return false;
	}
```

ybiquitous (changes requested):
Thanks for the pull request. Overall LGTM, but I've left a comment on code style.

lib/utils/report.mjs
Mouvedia: > The `return x, y` pattern is uncommon…

Should we add an ESLint rule to our config to disallow it?
I liked the terseness.

ybiquitous: > Should we add an ESLint rule to our config to disallow it?

If we could add such an ESLint rule easily (e.g., a built-in rule), I'd be okay with it. But I don't have a strong hope since such a pattern is rare in this code base.

Mouvedia: https://eslint.org/docs/latest/rules/no-sequences

ybiquitous (approved):
Now much simpler. LGTM 👍🏼

lib/utils/report.mjs
ybiquitous: Good. Can you open a PR to add the `no-sequences` rule to our ESLint config if you have time?
https://github.com/stylelint/eslint-config-stylelint/blob/f80257c6116306f7ee7e38457a6f7d98cffbe259/index.js

Mouvedia: Sorry, but this time I won't.
I actually like that pattern so you will understand that I don't want to initiate that change :)

PR #7864, opened 2024-07-18
Fix document to correct stdout to stderr since v16

### LINKED ISSUE #7863: Prints to stderr

.h1 {
}

/* Duplicate properties - can be auto-fixed */
a { color: pink; color: orange; }

export default {
  rules: {
    "block-no-empty": true,
    "declaration-block-no-duplicate-properties": true,
  },
};

```
 example % ./node_modules/.bin/stylelint >stdout 2>stderr --fix src/hello.css
alexeagle@aspect-build example % cat stderr                                                        

src/hello.css
  11:5  ✖  Unexpected empty block  block-no-empty

✖ 1 problem (1 error, 0 warnings)

alexeagle@aspect-build example % cat stdout                                                        
alexeagle@aspect-build example % 
```

{
  "dependencies": {
  },
  "devDependencies": {

    "stylelint": "^16",

https://stylelint.io/user-guide/cli#--output-file--o says it should print to "Standard Output"

### What actually happened?

went to stderr

_No response_

alexeagle:
Fixes #7863

### DIFF
--- a/docs/user-guide/cli.md
+++ b/docs/user-guide/cli.md
@@ -90,7 +90,7 @@ Set a limit to the number of warnings accepted. [More info](options.md#maxwarnin
 
 ### `--output-file, -o`
 
-Path of file to write a report. Stylelint outputs the report to the specified `filename` in addition to standard error.
+Path of file to write a report. Stylelint outputs the report to the specified `filename` in addition to `process.stderr`.
 
 ### `--print-config`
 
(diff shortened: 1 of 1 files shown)


ybiquitous:
@alexeagle Thanks for the pull request! Since I found one more mistake, so I fixed it via 5a62b0d4a.

PR #7865, opened 2024-07-19
Deprecate `github` formatter

### LINKED ISSUE #7447: Deprecate GitHub formatter

Ref: https://github.com/stylelint/stylelint/issues/7430#issuecomment-1872329954

Deprecate the GitHub formatter and remove it in the next major release.

Ideally, we'd link to a newly published GitHub custom formatter as part of the deprecation changelog entry.

marcalexiei:

Close #7447

No, it's self-explanatory


.changeset/twenty-apes-shave.md
```diff
@@ -0,0 +1,7 @@
+---
+"stylelint": minor
+---
+
+Deprecate `github` formatter
+
+You can use [stylelint-actions-formatters](https://github.com/xt0rted/stylelint-actions-formatters) instead
```
Mouvedia, line 7:
```suggestion
Deprecated: GitHub formatter
```

lib/standalone.mjs
```diff
@@ -83,6 +83,16 @@ export default async function standalone({
 	let formatterFunction;
 
 	try {
+		/** @see https://github.com/stylelint/stylelint/issues/7447 */
+		if (isString(formatter) && formatter === 'github') {
+			process.emitWarning(
```
Mouvedia, line 88: medium

Use `emitDeprecationWarning` util.

types/stylelint/index.d.ts
```diff
@@ -188,6 +188,7 @@ declare namespace stylelint {
 	type Formatters = {
 		readonly compact: Promise<Formatter>;
 		readonly github: Promise<Formatter>;
+		/** @deprecated */
```
Mouvedia, line 191: It seems to be on the wrong line.

Mouvedia (changes requested):
@marcalexiei 
Thanks for the PR.
I have made some comments, please review them.

lib/standalone.mjs
marcalexiei: Added a new code to `emitDeprecationWarning`:  `GITHUB_FORMATTER: '004'`

lib/standalone.mjs
```diff
@@ -83,6 +84,15 @@ export default async function standalone({
 	let formatterFunction;
 
 	try {
+		/** @see https://github.com/stylelint/stylelint/issues/7447 */
+		if (isString(formatter) && formatter === 'github') {
```
Mouvedia, line 88: question

Is `isString(formatter) &&` necessary?

lib/__tests__/cli-deprecation.test.mjs
```diff
@@ -0,0 +1,33 @@
+import { fileURLToPath } from 'node:url';
+import path from 'node:path';
+import process from 'node:process';
+
+import { expect, jest } from '@jest/globals';
```
Mouvedia, line 5: nit

remove `expect`

lib/standalone.mjs
marcalexiei: Actually no... even if it is a function or `undefined` the `isString` check is not necessary

Mouvedia:
Unless you wanna add a skipped test to cover #7621, this looks good to me.

Mouvedia (approved):

marcalexiei:
IMHO #7826 introduces a new config option so I think it's better to add the test there directly.

marcalexiei:
Obviously If #7826 is merged before this PR I have no problem adding the tests here 😃

lib/standalone.mjs
```diff
@@ -83,6 +84,15 @@ export default async function standalone({
...
 	try {
+		/** @see https://github.com/stylelint/stylelint/issues/7447 */
+		if (formatter === 'github') {
+			emitDeprecationWarning(
+				'"github" formatter is deprecated',
+				'GITHUB_FORMATTER',
+				'Please use "stylelint-actions-formatters" instead.',
+			);
+		}
+
```
Mouvedia, line 95: minor
refactor

I think this should be moved in `getFormatterFunction` inside the first `if`.

lib/standalone.cjs
```diff
@@ -303,6 +304,15 @@ async function standalone({
  */
 function getFormatterFunction(selected) {
 	if (typeof selected === 'string') {
+		/** @see https://github.com/stylelint/stylelint/issues/7447 */
+		if (selected === 'github') {
```
Mouvedia, line 308: @marcalexiei
Thinking about it more, you will need to add `!quietDeprecationWarnings && ` to the condition, hence you should put that `if` where it was originally.
i.e. l88–94

---

@ybiquitous after this PR we should put the `if (!quietDeprecationWarnings)`inside the `emitDeprecationWarning` util.

Mouvedia (approved):

.changeset/twenty-apes-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Deprecated: GitHub formatter
```
ybiquitous, line 5: [suggest] Super minor, but for consistency:

```suggestion
Deprecated: `github` formatter
```

ybiquitous (approved):
@marcalexiei Thanks for the first contribution. LGTM! 👍🏼

ybiquitous:
Oh, I forgot the document: 😅 
https://github.com/stylelint/stylelint/blob/0850a28a1780c84ae55f179f60dcf0ea6971f4a6/docs/user-guide/options.md?plain=1#L87

@marcalexiei Can you also update it?

docs/user-guide/options.md
```diff
@@ -84,7 +84,7 @@ Specify the formatter to format your results.
 Options are:
 
 - `compact` - generates output similar to ESLint's compact formatter
-- `github` - generates messages via [workflow commands for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions)
```
marcalexiei, line 87: @ybiquitous I added a `(DEPRECATED)` at the end of the string.
I also changed the same string in `cli.mjs` so it is more visible.

ybiquitous (approved):
Thanks. LGTM 👍🏼

PR #7866, opened 2024-07-19
Fix error message when a selector is duplicated inside a selector list

### LINKED ISSUE #7583: Fix `no-duplicate-selectors` false negatives for three or more duplicates

```css
a[href*="abc"],
a[href*=""],
a[href*=""],
a[href*=""] {
  color: #fd0;
}
```

default

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGBtAFgJxgMwFQF4AdZAIzGIF0AaQgO1UxwOMpvvSzyJFboc+Y8QFAATA6IkZAA2EDPBEBibFAAMAbjoBfEFRDYAltJgA5JAFs4iGAA8LAB2MA6MAGdXu8BFqGA5ghBxWkliWwAXGFooV2IFFAlJEWJXMIBPY2kDWjCAWkgfA18clKQopAwoFmCRCm1PfL8AMTlzJDCAgCtXb09Yew9EIJCQFPSYTOzYpJBpNpgU4jZh0Yys3IbC4rDSqHLKhGnZiIWeWh0tIA)

```json
{
  "stylelint": "16.3.0",
  "stylelint-config-standard": "36.0.0"
}
```

Error should occur on line 3 & 4 and refer to line 2 e.g.
Line 2 selector is not a duplicate of line 1 selector.

---
PS.  (on another note) Is there a rule for empty selector value? :thinking: 

### What actually happened?

ERROR Unexpected duplicate selector "a[href*=""]", first used at line 1 (no-duplicate-selectors) [2:1-12]

_No response_

marcalexiei:

Close #7583

Information about the updated implementation can be found in the [comment below](https://github.com/stylelint/stylelint/pull/7866#issuecomment-2238582384).

<details>
<summary>Outdated description</summary>
I tackled a little with `no-duplicate-selectors`, unfortunately I wasn't able to find a clean and efficient way to highlight duplicate selectors inside a single selector list.
The alternative solution I'm proposing is to change the message of the rule adding to it the occurrences of the duplicated selector within the list.
Something like:
```text
`Selector "${selector}" is present ${count} times in selector list at line ${line}`
```
Inevitably this will affect other scenarios but I think that this make the error more understandable
</details>

### DIFF
--- a/.changeset/gorgeous-maps-juggle.md
+++ b/.changeset/gorgeous-maps-juggle.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Fix no-duplicated-selectors error position when duplication is in the same selector list
--- a/lib/__tests__/standalone-fix.test.mjs
+++ b/lib/__tests__/standalone-fix.test.mjs
@@ -593,7 +593,7 @@ it('two rules being disabled', async () => {
 	expect(result.code).toBe(code);
 	expect(getCleanOutput(result.report)).toBe(stripIndent`
 		test.css
-		  2:3   ×  Unexpected duplicate selector "a", first used at line 2  no-duplicate-selectors
+		  2:6   ×  Unexpected duplicate selector "a", first used at line 2  no-duplicate-selectors
 		  3:11  ×  Expected "#ffffff" to be "#fff"                          color-hex-length
 
 		× 2 problems (2 errors, 0 warnings)
@@ -644,7 +644,7 @@ it('one rule being disabled and another still autofixing', async () => {
 		}`);
 	expect(getCleanOutput(result.report)).toBe(stripIndent`
 		test.css
-		  2:3  ×  Unexpected duplicate selector "a", first used at line 2  no-duplicate-selectors
+		  2:6  ×  Unexpected duplicate selector "a", first used at line 2  no-duplicate-selectors
 
 		× 1 problem (1 error, 0 warnings)
 	`);
--- a/lib/rules/no-duplicate-selectors/__tests__/fixtures/many-duplicates.css
+++ b/lib/rules/no-duplicate-selectors/__tests__/fixtures/many-duplicates.css
@@ -1,0 +1,6 @@
+div,
+div,
+body,
+div {
+  color: #fd0;
+}
--- a/lib/rules/no-duplicate-selectors/__tests__/index.mjs
+++ b/lib/rules/no-duplicate-selectors/__tests__/index.mjs
@@ -1,4 +1,5 @@
 import { fileURLToPath } from 'node:url';
+import fs from 'node:fs';
 
 import postcss from 'postcss';
 import postcssImport from 'postcss-import';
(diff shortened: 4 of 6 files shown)


Mouvedia:
> `Selector "${selector}" is present ${count} times in selector list at line ${line}`

To me that sounds like Ill find the duplicates if I check that line; but the duplicates may be on different lines.
Is is the line of the selector that ought to be kept?
If so, that's unclear.

marcalexiei:
Gotcha, maybe something like this can be more straight forward:
```text
`Selector "${selector}" is present ${count} times in selector list starting at line ${line}`
```
What do you think?

Mouvedia:
> What do you think?

This should have been discussed in the issue itself but since you are asking here: I would expect that each duplicate would get its own warning. If you agree with me you can turn this PR into a draft while you are working on it.
If not you can wait for more feedback from the other reviewers; they might have another opinion.

marcalexiei:
> This should have been discussed in the issue itself

Gotcha, apologies for that.

---

> I would expect that each duplicate would get its own warning

Having one warning for each duplicate is more coherent with the current implementation so I'll try to comply with it.

---

 I'll also try to find a way to retrieve the correct selector list item source position to correctly reference the warning.

marcalexiei:
I changed the implementation to report a warning for each duplicated selector.
Now the warning is positioned right on the duplicated value.

- Before cycling on the rule selectors I clone the rule
- Then every time I found a duplicated selector I replace the previous occurrences on the clone rule selector  with a placeholder with the same length
- By doing so I can use [Rule.rangeBy](https://postcss.org/api/#atrule-rangeby) to get the exact `start` and `end` position of the selector item!

romainmenke:
Hi @marcalexiei,
Thank you for your contribution!

However I am not sure of this direction is what we want.

I had this rule on my todo list for another issue: https://github.com/stylelint/stylelint/issues/6234

I think both issues have the same underlying cause, that this rule has too much logic and equation checks around the string representation of selectors.

For `a, a, a {}` the rule can't distinguish between the 1st, 2nd or 3rd `a`.

However if we convert the rule to use a proper AST then each will be a different class instance. Each having a correct link back to the source location.

I am concerned that applying this fix, in this way, would make it harder to improve the rule in general.

marcalexiei:
Greetings @romainmenke!
I understand your point, It's not a problem. Thanks for the explanation.
I will close this PR since is not needed

Mouvedia:
@romainmenke #7583 has the _good first issue_ label which explains why @marcalexiei attempted to fix it.
Should we remove it?

> I had this rule on my todo list for another issue: https://github.com/stylelint/stylelint/issues/6234

Do you want to assign yourself so that no one attempts to fix it in the meantime?

romainmenke:
Unfortunately I wasn't aware of https://github.com/stylelint/stylelint/issues/7583 :/

It was only in seeing this PR by @marcalexiei (which I very much appreciate) and reading through the changes that I realized how the work for https://github.com/stylelint/stylelint/issues/6234 was connected.

I've opened a PR with a patch for both issues here : https://github.com/stylelint/stylelint/pull/7867

@marcalexiei If you have time, please check my changes to ensure that I did indeed fix it :)

marcalexiei:
No problem, I'll take a look in the next yours!

PR #7867, opened 2024-07-19
Fix `no-duplicate-selectors` false negatives for three or more duplicates and end positions

### LINKED ISSUE #7583: Fix `no-duplicate-selectors` false negatives for three or more duplicates

romainmenke:

See: https://github.com/stylelint/stylelint/issues/6234
Closes: https://github.com/stylelint/stylelint/issues/7583

This takes the same approach as other changes in this series:
- parse with `postcss-selector-parser`
- flatten the selectors
- do checks against the AST, not against strings
- report warnings using the source locations from the AST

Because an AST is used we also automatically fix #7583 as we can now distinguish between the 2nd and 3rd selector in this list: `a, a, a {}`. I didn't really have to do anything for this. It just works as expected and requested in that issue when switching to an AST approach.

-------

Sorting the selector AST, as part of normalizing selectors, is a bit tricky.
It needs to sort inside out.

```css
:is(b, c), :is(c, a) {}
```

Sorting these outside in results in:

```css
:is(b, c), :is(a, c) {}
```

Because `:is(b, c)` comes before `:is(c, a)` and only later is there a sort of the child nodes.

Doing the sort correctly (inside out) means that these selectors are equal after normalization:

```css
:is(a, c), :is(b, c) {}
:is(b, c), :is(c, a) {}
```


lib/rules/no-duplicate-selectors/__tests__/index.mjs
```diff
@@ -100,7 +103,16 @@ testRule({
...
+					endLine: 1,
+					endColumn: 7,
+				},
+				{
+					message: messages.rejected('.a', 1),
+					line: 1,
+					column: 9,
+					endLine: 1,
+					endColumn: 11,
 				},
```
romainmenke, line 116: We now have two warnings instead of only one, relevant for : https://github.com/stylelint/stylelint/issues/7583

.changeset/stale-pianos-teach.md
```diff
@@ -0,0 +1,6 @@
+---
+"stylelint": patch
```
Mouvedia, line 2:
```suggestion
"stylelint": minor
```

ref: https://stylelint.io/about/semantic-versioning/#minor-release
e.g. `a, a, a {}`

romainmenke: Is this really reporting **new** errors?

Before this change it would first report for a single `a` and after the user fixes that error it would report for a second `a`.

lib/rules/no-duplicate-selectors/__tests__/index.mjs
```diff
@@ -134,6 +150,21 @@ testRule({
 					message: messages.rejected('b, a', 1),
 					line: 1,
 					column: 9,
+					endLine: 1,
+					endColumn: 13,
+				},
+			],
+		},
+		{
+			code: ':is(a, c), :is(b, c) {} :is(b, c), :is(c, a) {}',
```
Mouvedia, line 159: question
off-topic

Should we detect `is:(a), a {}` too?

lib/rules/no-duplicate-selectors/index.cjs
```diff
@@ -53,122 +55,217 @@ const rule = (primary, secondaryOptions) => {
...
-			);
-			const resolvedSelectorList = [
-				...new Set(
-					ruleNode.selectors.flatMap((selector) => resolvedNestedSelector(selector, ruleNode)),
-				),
-			];
-
-			const normalizedSelectorList = resolvedSelectorList.map(normalize);
+		/**
+		 * @param {import('postcss').Rule} node
```
Mouvedia, line 59: suggestion
refactor

You have got 4 `import('postcss').Rule`, maybe add a `/** @typedef {import('postcss').Rule} Foo */`?

.changeset/stale-pianos-teach.md
Mouvedia: Iv been asked to switch to `minor` for the same reason before.
It depends on your interpretation of

> a bug fix in a rule that results in Stylelint reporting _more errors_

i.e. 2 warnings/errors instead of 1 with my example

marcalexiei (commented):
I found a small regressions, given the following code:

```css
div,
div {
  color: red;
}

div { // ⬅️ not marked as error with the new patch
  color: red;
}
```

---

With 16.7.0 the result is:

```text
  1:1  ✖  Unexpected duplicate selector "div", first used at line 1  no-duplicate-selectors
  6:1  ✖  Unexpected duplicate selector "div", first used at line 1  no-duplicate-selectors

✖ 2 problems (2 errors, 0 warnings)
```

---

With this PR change the selector on line 6 is not marked as error:
```text
  2:1  ✖  Unexpected duplicate selector "div", first used at line 1  no-duplicate-selectors

✖ 1 problem (1 error, 0 warnings)
```

.changeset/stale-pianos-teach.md
romainmenke: Sure!

I thought that the intention was that if Stylelint runs without errors and still runs without errors after an update that it implies a patch.

But happy to adjust the changeset.

romainmenke:
Thank you @marcalexiei

On main this test case produces these results: 

```js
		{
			code: '.a, .a {} .a {}',
			description: "duplicated selectors within two rule's selector list. 3 duplicates",
			warnings: [
				{
					message: messages.rejected('.a', 1),
					line: 1,
					column: 1,
				},
				{
					message: messages.rejected('.a', 1),
					line: 1,
					column: 11,
				},
			],
		},
```

.changeset/stale-pianos-teach.md
Mouvedia: > I thought that the intention was that if Stylelint runs without errors and still runs without errors after an update that it implies a patch.

Probably, yes.

> But happy to adjust the changeset.

I guess this bullet point would be less subject to interpretation if we had a tool like https://github.com/hjylewis/esplint but for Stylelint.

lib/rules/no-duplicate-selectors/index.cjs
romainmenke: I've added this instead:

```js
/** @import { Rule as PostcssRule } from 'postcss' */
/** @import { Node as SelectorNode, Root as SelectorRoot, Selector } from 'postcss-selector-parser' */
```

lib/rules/no-duplicate-selectors/__tests__/index.mjs
Mouvedia, line 1: question
off-topic

If `.a.b` and `.b.a` are equivalent, should we detect these cases as well?

romainmenke: That is indeed off-topic :)
You can file a new issue to discuss.

Mouvedia: I would like your opinion though, is it expected from that rule? Is it desirable?

romainmenke: You might be interested in this: https://github.com/csstools/postcss-plugins/blob/main/packages/selector-resolve-nested/src/compound-selector-order.ts

This sorts selectors so that you don't accidentally change the meaning after serializing.
So that joining `.a` and `b` doesn't accidentally become `.ab`, but is sorted to `b.a`.

To same code can be used but you need to add an extra sort by "value" if the types are the same.

This sorting should be applied during normalization.

Mouvedia: OK, what I am getting is that you are not against the idea.
I thought I missed something obvious.

romainmenke: I think it is better to open a new issue and to not discuss things there :)

Discussing these off topic questions adds friction for me.
It makes it harder to keep focus on the task and to know when all points have been addressed. It also often requires me to switch context to this new scenario, which is taxing.

If the same question was asked in a separate issue I could choose when to read it and if my input would be helpful. By asking it here you force me to spend time on it. I know that you mean well, but it would help me if we can keep threads more on topic.

Mouvedia: > I think it is better to open a new issue and to not discuss things there :)

#7869
I just wanted to avoid opening an issue for something that might have been already discussed and discarded.
i.e. when something seems obvious to me and is not implemented I err on the side that I am missing something
Sorry for the noise.

marcalexiei (approved):
I've done some more tests and everything works like a charm as far as I can tell 😃!

.changeset/stale-pianos-teach.md
```diff
@@ -0,0 +1,6 @@
+---
+"stylelint": minor
+---
+
+Fixed: `no-duplicate-selectors` end positions
+Fixed: `no-duplicate-selectors` false negatives for three or more duplicates
```
ybiquitous, line 6: [suggest] Just to be safe, can you separate the entries to each file? I don't remember if `changeset` supports multiple entries, and each entry should have different effects:

- "Fixed: `no-duplicate-selectors` end positions" -> patch
- "Fixed: `no-duplicate-selectors` false negatives for three or more duplicates" -> minor

ybiquitous (commented):
@romainmenke Thanks for the pull request! This change looks great 👍🏼 

The `JSON` way makes sense for me, but I'm a bit concerned about the performance since I heard `JSON` methods were slow. Any thoughts?

.changeset/stale-pianos-teach.md
Mouvedia: related: https://github.com/stylelint/stylelint/pull/7859#discussion_r1682913500

ybiquitous: I tried `npx changeset version` locally. Unfortunately, such double entries don't generate the expected content. Here's the diff in `CHANGELOG.md`:

```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index f5bf24418..073dfd9b8 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 16.8.0
+
+### Minor Changes
+
+- Fixed: `no-duplicate-selectors` end positions Fixed: `no-duplicate-selectors` false negatives for three or more duplicates ([#7867](https://github.com/stylelint/stylelint/pull/7867)) ([@romainmenke](https://github.com/romainmenke)).
+
 ## 16.7.0
 
 - Changed: `tap` formatter to support TAP14 ([#7759](https://github.com/stylelint/stylelint/pull/7759)) ([@Mouvedia](https://github.com/Mouvedia)).
```

So we have to separate entries for each file.

romainmenke:
> The `JSON` way makes sense for me, but I'm a bit concerned about the performance since I heard `JSON` methods were slow. Any thoughts?

I agree :)

If I recall correctly node has internal optimizations for this pattern.
Where you `parse` immediately after `stringify` without reading or writing to the result of `stringify` as if it were a string value. It does the actual work lazily.

But we also have a regexp in there and it definitely is more work and complexity than really needed.

I am going to move it to a helper because then I can write some unit tests, if that is ok?

lib/utils/normalizeSelector.mjs
```diff
@@ -0,0 +1,84 @@
...
+	node.spaces.after = '';
+
+	if (parser.isAttribute(node)) {
+		delete node.spaces.insensitive;
+		delete node.spaces.operator;
+		delete node.spaces.attribute;
+		delete node.spaces.value;
+
+		delete node.raws.spaces;
+	}
```
romainmenke, line 45: This was a lot less code and complexity than I was expecting it to be 😅 

The typings in `postcss-selector-parser` are really complex and gave me the false impression that each node has some specific fields, while in reality it limited to attributes 🎉

romainmenke:
@ybiquitous Thank you for the feedback.
I've applied some changes and this is now ready for a second look 🙇

lib/rules/no-duplicate-selectors/__tests__/index.mjs
Mouvedia, line 1: major
regression

Try

```css
a, /* test */a {}
```

It passes and it used to report an error.
[demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGAaABAegFRoC4wGdc0sMk1gBfEFEAMwEsAbGAOSQFs5EYAPTgA4sAdGAIEa4CADtGAcwQhgAHWlo0ykACcAriwKb4FVevWbpEALRQdQhmCT5LBGCzC4IWgwjy6YJtEpValpIWQY5ADFPDkdFACsCGUlYAQlEFTUNECIATxYmBmlcQ2ymR0ISmgDNPIKi3Esw+WdcJGkoJC0oUs1y-CJNYJBKIA)

Mouvedia (dismissed):

lib/rules/no-duplicate-selectors/__tests__/index.mjs
romainmenke: Can you check again, I think I have already fixed this, or maybe I am misunderstanding what the issue is?

I added extra test coverage anyway, just to be sure.

Mouvedia: I must have tested a previous commit then, my bad.

romainmenke: No problem, thank you for checking this thoroughly 🙇

lib/rules/no-duplicate-selectors/index.mjs
```diff
@@ -1,12 +1,14 @@
...
+import getRuleSelector from '../../utils/getRuleSelector.mjs';
+import getSelectorSourceIndex from '../../utils/getSelectorSourceIndex.mjs';
 import { isBoolean } from '../../utils/validateTypes.mjs';
 import isKeyframeRule from '../../utils/isKeyframeRule.mjs';
-import isStandardSyntaxSelector from '../../utils/isStandardSyntaxSelector.mjs';
+import isStandardSyntaxRule from '../../utils/isStandardSyntaxRule.mjs';
 import nodeContextLookup from '../../utils/nodeContextLookup.mjs';
+import normalizeSelector from '../../utils/normalizeSelector.mjs';
 import parseSelector from '../../utils/parseSelector.mjs';
+import parser from 'postcss-selector-parser';
```
ybiquitous, line 11: [suggest] Can you move importing `postcss-selector-parser` to the top of this file? I want to distinguish between importing outside packages and inside modules.

Also, `parser` is ambiguous, so I'd like to rename it. For example:

```suggestion
import parserSelector from 'postcss-selector-parser';
```

lib/rules/no-duplicate-selectors/index.mjs
```diff
@@ -50,124 +55,140 @@ const rule = (primary, secondaryOptions) => {
...
+		function checkSelectorListInDifferentRules(ruleNode, contextSelectorSet) {
+			const combinedRoot = parser.root({
+				nodes: flattenNestedSelectorsForRule(ruleNode, result).flatMap(
+					({ resolvedSelectors }) => resolvedSelectors.nodes,
+				),
+				value: '',
+			});
 
-			const selectorLine = ruleNode.source.start.line;
+			const normalized = normalizeSelector(combinedRoot.clone()).toString();
```
ybiquitous, line 121: [ask] Just curious, is it worth avoiding `.clone()` like this?

```suggestion
			const normalized = normalizeSelector(combinedRoot).toString();
```

It seems that cloning may be unnecessary since `parser.root()` creates a new object. But, if the performance diff is too low, please leave it as-is.

ybiquitous (commented):
@romainmenke Great improvements, thanks a lot! 👍🏼 

I've left a few questions, but they should be trivial. Overall LGTM.

lib/rules/no-duplicate-selectors/index.mjs
romainmenke: Good catch!
We can avoid it here and at line 151.

We still need it at line 88 because that variable is used in error reporting and should not be mutated.

ybiquitous (approved):
Thanks. LGTM 👍🏼

romainmenke:
Thank you for the feedback and thorough review everyone 🙇

Mouvedia:
@romainmenke did this fix the last task of #6075?

Issue #7887, opened 2024-07-29
Fix `no-duplicate-selectors` false positives with Less syntax


FritsvanCampen:
### What minimal example or steps are needed to reproduce the bug?

```less
.foo() {
	.bar > & {
		color: #000;
	}
	
	&.readonly {
		color: #fff;
	}
}
```
Note: it's important that the root rule is a mixin (needs brackets)

### What minimal configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "no-duplicate-selectors": true
  }
}
```

### How did you run Stylelint?

https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0AzCEAUBKABMAOgHb4AuSARgIYBOOAfDgGR5GmmQA2E18OAxAAYhAblYkAvmLGMk1GJSgRCHAJ4tiJdhC49+KfaI2TC4kABoQKAJYcYAOUoBbOIhgAPJwAdbSWwGc-c3AlawBzBBACQhwcfBB3EhhCKD843gBtIhiYuL8SFVsOK0ISAFpIQjDSvMpkmig4rJwAXSJTCwqwgDFuR0oSCIArPyUg2E9AxCickDyCmCKStNiQDn6YPLizJtz8wuKyzqtQ6pJaqHrluLXEzZA2kHEgA

### Which Stylelint-related dependencies are you using?

16.8.0

### What did you expect to happen?

No error

### What actually happened?

Stylelint: Unexpected duplicate selector "&.readonly", first used at line 2 (no-duplicate-selectors)

For larger files the errors become non-sensical, so I expect there might be some contextual error.

### Do you have a proposal to fix the bug?

I suspect some logic was changed where less-tokens (such as `&` and `+`) are now incorrectly detected as selectors.

romainmenke:
Thank you for reporting this @FritsvanCampen 

This regression was likely introduced in https://github.com/stylelint/stylelint/pull/7867

I've labeled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

There are [steps on how to fix a bug in a rule](https://stylelint.io/developer-guide/rules#fix-a-bug-in-a-rule) in the Developer guide.

FritsvanCampen:
Thanks for the quick fix! 🙏

PR #7888, opened 2024-07-29
Fix `no-duplicate-selectors` false positives with Less syntax

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/7887

`flattenNestedSelectorsForRule` can return an empty list on invalid syntax.
This forgiving behavior is intended as it also allows us to lint partially standard selector lists.

However in `no-duplicate-selectors` we also use these lists (after normalizing) to detect duplicates. And when the entire list is non-standard the resulting selector will be an empty string.

Each empty string is equal to any other empty string, so in turn each non-standard selector list is considered a duplicate of every other non-standard selector list.

Adding a check that `flattenNestedSelectorsForRule` did return at least one selector resolves any issues with non-standard syntaxes.

### DIFF
--- a/lib/rules/no-duplicate-selectors/__tests__/index.mjs
+++ b/lib/rules/no-duplicate-selectors/__tests__/index.mjs
@@ -521,5 +521,8 @@ testRule({
 			code: 'a { b, .@{var} {} b {} }',
 			description: 'Non standard Less nested interpolation (3',
 		},
+		{
+			code: '.foo() { .bar > & { color: #000; } &.readonly { color: #fff; } }',
+		},
 	],
 });
--- a/lib/rules/no-duplicate-selectors/index.mjs
+++ b/lib/rules/no-duplicate-selectors/index.mjs
@@ -119,9 +119,7 @@ const rule = (primary, secondaryOptions) => {
 			if (!flattenedNestedSelectors.length) return;
 
 			const combinedRoot = selectorParser.root({
-				nodes: flattenNestedSelectorsForRule(ruleNode, result).flatMap(
-					({ resolvedSelectors }) => resolvedSelectors.nodes,
-				),
+				nodes: flattenedNestedSelectors,
 				value: '',
 			});
 
--- a/lib/rules/no-duplicate-selectors/index.mjs
+++ b/lib/rules/no-duplicate-selectors/index.mjs
@@ -112,6 +112,12 @@ const rule = (primary, secondaryOptions) => {
 		 * @param {Map<string, { line: number }>} contextSelectorSet
 		 */
 		function checkSelectorListInDifferentRules(ruleNode, contextSelectorSet) {
+			const flattenedNestedSelectors = flattenNestedSelectorsForRule(ruleNode, result).flatMap(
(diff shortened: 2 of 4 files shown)


lib/rules/no-duplicate-selectors/__tests__/index.mjs
Mouvedia, line 1: If you want more coverage you could add tests for `& + &` and `foo &`.
e.g. `'a { div & { & + & { color: red; } } }'`

romainmenke: But that is standard syntax, so it doesn't actually increase coverage for this issue?

See: https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYKoEsBuKBka+KA1EegM4AOSAdgepADYQBO8KLMUA3CgL79BfEABoQAM0yMYAOSQBbOIhgAPBZWkA6aeXKjwEGpIDmCEMAA6NFCgshVAFxg0o5O+wDaVmzbvkHAJ7SjJg0DgC0kEaYxuH+tFBILFB23igAulbCYlEmAGKs8kgOZgBW5Ib6sJR6iJbWtiD+QTAhYe6NjMUw-nYiaX6BwaERuTFxDglJKQid3b0gWSB8QA

Mouvedia: I was testing with Chrome 111, my bad.

lib/rules/no-duplicate-selectors/index.mjs
```diff
@@ -112,6 +112,12 @@ const rule = (primary, secondaryOptions) => {
 		 * @param {Map<string, { line: number }>} contextSelectorSet
 		 */
 		function checkSelectorListInDifferentRules(ruleNode, contextSelectorSet) {
+			const flattenedNestedSelectors = flattenNestedSelectorsForRule(ruleNode, result).flatMap(
```
Mouvedia, line 115: refactor
l122–124

```
				nodes: flattenedNestedSelectors,
```

romainmenke: Good catch, thanks.

I added the early check, but forgot to use the variable here.

lib/rules/no-duplicate-selectors/__tests__/index.mjs
```diff
@@ -521,5 +521,8 @@ testRule({
 			code: 'a { b, .@{var} {} b {} }',
 			description: 'Non standard Less nested interpolation (3',
 		},
+		{
+			code: '.foo() { .bar > & { color: #000; } &.readonly { color: #fff; } }',
```
Mouvedia, line 525: I suppose the mixin is being ignored in its entirety.
i.e. if there are duplicates in `.foo() { /* here */ }`, they won't be detected

Mouvedia (approved):

Mouvedia:
@ybiquitous Should we release 16.8.1 ?
Or wait for more PRs?
By we, I mean you :)

ybiquitous (approved):
Thanks, LGTM 👍🏼 

I'll release 16.8.1 soon.

Issue #7904, opened 2024-08-06
Fix reported ranges of warnings in selectors


pamelalozano16:
### What minimal example or steps are needed to reproduce the bug?

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0YA0ACA9J9BhCAtgTAHYAu6AjADolJQbZ6HHnoBMtwt66UAlgGcADgBsAhgE946AEaiIYANYBuWgF9atTACp0gspNExR-cgFoBg8fJjmSMAB5lzph+hIRzZgG7jTUJYQAK625oISggAW5pBEpGSC6DqYtEjijDj48Ww0JLr6hsZuLlY2xvZOLm4wHl6+-vyBUCFhEeLRsSwJSSlpspnMORScJNwkvFZiUjKeDmokmiQgqCAAZvzGAHLixAggTrtiMCiCgivgECQbAOb747zUB86kUIJPMgDaPLzoTwZGExmFyQa78G7hMjiEhQcQAJygTx+AF0NBdQbcAGIQOEEcRkfYAK0EVwusGE50QDz+IABxWBHxpEjIMAMT1QP3+RSBFgx4Mh0NhCMZT2ZrLISMWIHUQA)

### What minimal configuration is needed to reproduce the bug?

```json
{
  "extends": [
    "stylelint-config-standard"
  ]
}
```

### How did you run Stylelint?

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0YA0ACA9J9BhCAtgTAHYAu6AjADolJQbZ6HHnoBMtwt66UAlgGcADgBsAhgE946AEaiIYANYBuWgF9atTACp0gspNExR-cgFoBg8fJjmSMAB5lzph+hIRzZgG7jTUJYQAK625oISggAW5pBEpGSC6DqYtEjijDj48Ww0JLr6hsZuLlY2xvZOLm4wHl6+-vyBUCFhEeLRsSwJSSlpspnMORScJNwkvFZiUjKeDmokmiQgqCAAZvzGAHLixAggTrtiMCiCgivgECQbAOb747zUB86kUIJPMgDaPLzoTwZGExmFyQa78G7hMjiEhQcQAJygTx+AF0NBdQbcAGIQOEEcRkfYAK0EVwusGE50QDz+IABxWBHxpEjIMAMT1QP3+RSBFgx4Mh0NhCMZT2ZrLISMWIHUQA)

### Which Stylelint-related dependencies are you using?

[...]

PR #7905, opened 2024-08-07
Fix `no-invalid-double-slash-comments` reported ranges

romainmenke:

I also noticed that rules were eagerly serialized, this is quite expensive.

This change:

```
;node benchmark-rule.mjs no-invalid-double-slash-comments 'true'                               
Warnings: 0
Mean: 95.35613458823532 ms
Deviation: 9.862092026552252 ms
```

Main:

```
;node benchmark-rule.mjs no-invalid-double-slash-comments 'true'
Warnings: 0
Mean: 118.80464825 ms
Deviation: 23.471924727353816 ms
```

### DIFF
--- a/lib/rules/no-invalid-double-slash-comments/__tests__/index.mjs
+++ b/lib/rules/no-invalid-double-slash-comments/__tests__/index.mjs
@@ -1,6 +1,8 @@
 import rule from '../index.mjs';
 const { messages, ruleName } = rule;
 
+import { stripIndent } from 'common-tags';
+
 testRule({
 	ruleName,
 	config: [true],
@@ -17,6 +19,18 @@ testRule({
 		{
 			code: 'a { background: url(//foo.com/bar.png) }',
 			description: 'url with double slash',
+		},
+		{
+			code: stripIndent`
+				/* stylelint-disable-next-line no-invalid-double-slash-comments */
+				.a, // Comment 1
+				/* stylelint-disable-next-line no-invalid-double-slash-comments */
+				.b, // Comment 2
+				.c {
+					color: red;
+				}
+			`,
+			description: 'configuration commands interleaved with double slash comments',
 		},
 	],
 
@@ -74,6 +88,33 @@ testRule({
 			column: 1,
 			endLine: 1,
 			endColumn: 13,
+		},
+		{
+			code: stripIndent`
+				.a, // Comment 1
+				/* foo */
+				.b, // Comment 2
+				.c {
+					color: red;
+				}
+			`,
+			description: 'regular comments interleaved with double slash comments',
+			warnings: [
+				{
+					message: messages.rejected,
+					line: 1,
+					column: 5,
+					endLine: 2,
+					endColumn: 1,
+				},
+				{
+					message: messages.rejected,
+					line: 3,
+					column: 5,
+					endLine: 4,
+					endColumn: 1,
+				},
+			],
 		},
 	],
 });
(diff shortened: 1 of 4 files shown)


lib/rules/no-invalid-double-slash-comments/__tests__/index.mjs
```diff
@@ -75,6 +89,33 @@ testRule({
...
+					color: red;
+				}
+			`,
+			description: 'regular comments interleaved with double slash comments',
+			warnings: [
+				{
+					message: messages.rejected,
+					line: 1,
+					column: 5,
+					endLine: 2,
```
romainmenke, line 107: I am not sure if the `endLine` here is correct or helpful, but out of scope for this issue. This issue is limited to incorrect ranges as a result of interleaved comments. It might be that this is fine :)

lib/rules/no-invalid-double-slash-comments/index.cjs
```diff
@@ -37,9 +37,9 @@ const rule = (primary) => {
 		});
 
 		root.walkRules((ruleNode) => {
-			// ruleNode.selectors elements are trimmed
```
romainmenke, line 40: I removed this comment because there is test coverage for this aspect and we already have guidance against using `node.selectors`.

lib/rules/no-invalid-double-slash-comments/index.mjs
```diff
@@ -33,9 +33,9 @@ const rule = (primary) => {
 		});
 
 		root.walkRules((ruleNode) => {
-			// ruleNode.selectors elements are trimmed
-			const selectors = ruleNode.selector.split(',');
-			const ruleStringified = ruleNode.toString();
+			const selectors = (ruleNode.raws?.selector?.raw ?? ruleNode.selector).split(',');
```
ybiquitous, line 36: getRuleSelector util seems available instead.

lib/rules/no-invalid-double-slash-comments/index.cjs
Mouvedia: > we already have guidance against using `node.selectors`

I didn't know about that. If we already have a regression test coverage it can be removed.
i.e. the comment was there so that no one would refactor it back

ybiquitous (approved):
Thank you for your quick fix! I left a review but LGTM 👍

lib/rules/no-invalid-double-slash-comments/index.mjs
romainmenke: Thank you @ybiquitous 🙇

Maybe interesting to have some custom eslint rules to check such stuff :)

lib/rules/no-invalid-double-slash-comments/__tests__/index.mjs
Mouvedia: It has something to do with the trick of using `{}`.
i.e. `// bar {}` vs `// bar`

Mouvedia: It has something to do with the trick of using `{}`.
The others tests—except l74–83—all use a `{}` construct.
i.e. `// bar {}` vs `// bar`

So we should either have the real commented code (including the following construct) or the part before the `\n` selected.
i.e. either `// Comment 2\n` or `// Comment 2\n.c {\ncolor: red;\n}` (not `// Comment 2\n.`)
Either way it should be fixed.
From an UX standpoint it should be `// Comment 2\n`.

tl;dr `(lines.length > 1)` is false when using `{}`

PR #7906, opened 2024-08-07
Update `user-guide/ignore-code.md`

romainmenke:

No, it's self-explanatory.

### DIFF
--- a/docs/user-guide/ignore-code.md
+++ b/docs/user-guide/ignore-code.md
@@ -62,7 +62,7 @@ Stylelint supports complex, overlapping disabling & enabling patterns:
 /* stylelint-enable foo */
 ```
 
-> [!NOTICE]
+> [!NOTE]
 > Non-standard syntax comments (i.e. //) are only minimally supported.
 > They will only work when found outside of selectors or value lists.
 
--- a/docs/user-guide/ignore-code.md
+++ b/docs/user-guide/ignore-code.md
@@ -62,8 +62,8 @@ Stylelint supports complex, overlapping disabling & enabling patterns:
 /* stylelint-enable foo */
 ```
 
-> [!NOTE]
-> Non-standard syntax comments (i.e. //) are only minimally supported.
+> [!WARNING]
+> Non-standard syntax comments (i.e. `//`) are only minimally supported.  
 > They will only work when found outside of selectors or value lists.
 
 You may also include a description at the end of the comment, after two hyphens:
--- a/docs/user-guide/ignore-code.md
+++ b/docs/user-guide/ignore-code.md
@@ -63,7 +63,7 @@ Stylelint supports complex, overlapping disabling & enabling patterns:
 ```
 
 > [!WARNING]
-> Non-standard syntax comments (i.e. `//`) are only minimally supported.  
+> Configuration commands in non-standard syntax comments (e.g. `// stylelint-disable`) are only minimally supported.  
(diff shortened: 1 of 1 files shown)


docs/user-guide/ignore-code.md
```diff
@@ -62,9 +62,6 @@ Stylelint supports complex, overlapping disabling & enabling patterns:
 /* stylelint-enable foo */
 ```
 
-> [!WARNING]
```
Mouvedia, line 65:
```
> [!WARNING]
> Invalid CSS comments—i.e. starting with `//`—are ignored if they are within a selector or value list.
```

romainmenke: I am not sure that we need to warn about things that do not work when those are things we do not officially support.

But I don't have a strong opinion about this, happy to make changes here.

Mouvedia: I am sure my proposed wording sucks. Nonetheless we should at least have a `!NOTE`.

romainmenke: The wording is fine :)

I was mainly thinking that we already state elsewhere that our focus is on supporting standard CSS only. Mentioning each aspect/feature that doesn't work for non-standard CSS might not be that needed as we already mention it in general?

Mouvedia: Maybe mentioning the non-standard support would be warranted then?
i.e. "Invalid CSS comments—i.e. starting with `//`—are supported when found outside of selectors and/or value lists"

romainmenke: Maybe something like this:

> Non-standard syntax comments (i.e. `//`) are only minimally supported. They will only work when found outside of selectors or value lists.


----

I prefer `non-standard` over `invalid` in this context.

`//` is widely known syntax for comments in `less` and `scss`.

While most developers won't know that even standard CSS has quirks that make it swallow the next construct after `//`. Those quirks do not make `//` a comment since no comment token or node will be produced by standard CSS tokenizers or parsers.

(the `no-invalid-double-slash-comments` rule has a different context, requiring different terms)

docs/user-guide/ignore-code.md
```diff
@@ -62,8 +62,9 @@ Stylelint supports complex, overlapping disabling & enabling patterns:
...
-> Comments within _selector and value lists_ are currently ignored.
+> [!NOTICE]
```
romainmenke, line 65:
```suggestion
> [!NOTE]
```

Mouvedia (approved):

Mouvedia:
Can you take a look at #7907 while you are at it?

docs/user-guide/ignore-code.md
```diff
@@ -63,7 +63,8 @@ Stylelint supports complex, overlapping disabling & enabling patterns:
 ```
 
 > [!WARNING]
-> Comments within _selector and value lists_ are currently ignored.
+> Non-standard syntax comments (i.e. `//`) are only minimally supported.
```
romainmenke, line 66:
```suggestion
> Configuration commands in non-standard syntax comments (e.g. `// stylelint-disable`) are only minimally supported.  
```

romainmenke: To make it clear that we are not referring to support of comments, but rather the commands in such comments

Mouvedia (approved):

PR #7895, opened 2024-08-03
Deprecate `context.fix` in favour of the `fix` callback

Mouvedia:
> Which issue, if any, is this issue related to?

Closes #2643

> Is there anything in the PR that needs further explanation?

- codecov warning can be ignored
- this change should be part of the same version that will ship #7896

PR #7907, opened 2024-08-07
Fix `no-invalid-double-slash-comments` end positions

Mouvedia:

https://github.com/stylelint/stylelint/pull/7905#discussion_r1706912140

This needs to be merged in the same version as #7905.


lib/rules/no-invalid-double-slash-comments/index.mjs
```diff
@@ -36,6 +37,8 @@ const rule = (primary) => {
 		root.walkRules((ruleNode) => {
 			const selectors = getRuleSelector(ruleNode).split(',');
 
+			assert(ruleNode.source);
```
Mouvedia, line 40: @romainmenke can I assume that `source` is always present?

lib/rules/no-invalid-double-slash-comments/index.mjs
```diff
@@ -46,19 +49,19 @@ const rule = (primary) => {
 					const offset = value.length - selector.length;
 					const i = index + offset;
 
-					ruleStringified ??= ruleNode.toString();
-					const lines = ruleStringified.slice(i).split('\n');
-					let comment = lines[0] ?? '';
+					ruleStringified ??= ruleNode.source.input.css;
```
Mouvedia, line 52: This rule is not fixable so it should be safe to use.

lib/rules/no-invalid-double-slash-comments/index.mjs
```diff
@@ -46,19 +49,19 @@ const rule = (primary) => {
...
 
-					if (lines.length > 1) comment += '\n';
+					const lines = ruleStringified.slice(i).split('\n');
+					const comment = lines[0] ?? '';
```
Mouvedia, line 55: note

TS is overzealous.
i.e. `comment[0]` can be `undefined` only if `''` is passed to `split`
e.g. `''.split('')`

lib/rules/no-invalid-double-slash-comments/index.mjs
romainmenke: No, `source` is correctly typed and can be absent.
Impossible to test with `testRule` as this does ensure that `source` is present, but other plugins with rules might insert nodes without `source` or other runners might not provide it.

Mouvedia: > but other plugins with rules might insert nodes without source or other runners might not provide it.

More concretely is `ruleStringified ??= ruleNode.source.input.css;` safe to use?
If not, what would be the fallback?

romainmenke: I am not sure, I would have to investigate and I don't currently have time :/

In general I know that `source` is in reference to the source CSS (as a string) that was used to construct the AST.

When manually creating AST nodes through the programmatic API's there is no source CSS as a string for those nodes.

So you should create some AST nodes programmatically and then experiment from there.

Mouvedia:
@ybiquitous ping

.changeset/few-fans-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fix: `no-invalid-double-slash-comments` reported `endIndex`
```
ybiquitous, line 5: [suggest] How about adjusting it a bit? Because `endIndex` is an internally related thing.

```suggestion
Fixed: `no-invalid-double-slash-comments` reported end positions
```

lib/rules/no-invalid-double-slash-comments/index.mjs
```diff
@@ -47,18 +47,18 @@ const rule = (primary) => {
 					const i = index + offset;
 
 					ruleStringified ??= ruleNode.toString();
-					const lines = ruleStringified.slice(i).split('\n');
-					let comment = lines[0] ?? '';
 
-					if (lines.length > 1) comment += '\n';
+					const lines = ruleStringified.slice(i).split('\n');
```
ybiquitous, line 51: [suggest] I remember the existence of `context.newline`. Should we use it as much as possible like this?

```suggestion
					const lines = ruleStringified.slice(i).split(context.newline ?? '\n');
```

ybiquitous (commented):
@Mouvedia Thanks for the pull request. The change looks good to me, but I've left a few small reviews. Can you take a look?

.changeset/few-fans-bathe.md
Mouvedia:
```suggestion
Fixed: `no-invalid-double-slash-comments` end positions

lib/rules/no-invalid-double-slash-comments/index.mjs
Mouvedia: In #7895, we confirmed that newline is always truthy.
A PR that updates the type is long overdue.

ref https://github.com/stylelint/stylelint/pull/5418#discussion_r1636001255

ybiquitous (approved):
Thanks. LGTM 👍🏼

PR #7909, opened 2024-08-08
Add deprecation notice to `github` formatter's output

### LINKED ISSUE #7447: Deprecate GitHub formatter

Mouvedia:

Closes #7447

https://github.com/stylelint/stylelint/issues/7447#issuecomment-2269174803

### DIFF
--- a/lib/formatters/githubFormatter.mjs
+++ b/lib/formatters/githubFormatter.mjs
@@ -23,8 +23,7 @@ export default function githubFormatter(results, returnValue) {
 		});
 	});
 
-	if (lines.length) lines.push(`::notice title=${title}::The github formatter is deprecated`);
-
+	lines.push(`::notice title=Stylelint deprecation::The github formatter is deprecated`);
 	lines.push('');
 
 	return lines.join('\n');
--- a/lib/formatters/githubFormatter.mjs
+++ b/lib/formatters/githubFormatter.mjs
@@ -23,6 +23,8 @@ export default function githubFormatter(results, returnValue) {
 		});
 	});
 
+	if (lines.length) lines.push(`::notice title=${title}::The github formatter is deprecated`);
+
 	lines.push('');
 
 	return lines.join('\n');
--- a/.changeset/loud-humans-join.md
+++ b/.changeset/loud-humans-join.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Added: deprecation notice annotation to the output of the `github` formatter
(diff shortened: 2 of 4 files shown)


Mouvedia:
@ybiquitous should I add a changeset?

lib/formatters/githubFormatter.mjs
```diff
@@ -23,6 +23,8 @@ export default function githubFormatter(results, returnValue) {
 		});
 	});
 
+	if (lines.length) lines.push(`::notice title=${title}::The github formatter is deprecated`);
```
romainmenke, line 26:
```suggestion
	lines.push(`::notice title=${title}::The github formatter is deprecated`);
```

I think we always want to emit this notice.
Not only when there are other warnings.

romainmenke (commented):
Thank you for this change @Mouvedia
nice to see that it is fairly straightforward :)

lib/formatters/githubFormatter.mjs
Mouvedia: Check the `outputs no warnings` tests.
i.e. all the formatters `output === ""` when `warnings.length === 0`

Mouvedia: I found another formatter—`verbose`—that returns a truthy value when `warnings` array is empty so I will remove the `if`.
I still think it's a bit risky and will have unintended/unknown consequences.

romainmenke: If it draws peoples attention to this formatter and makes them consider the upcoming deprecation then maybe those are still good consequences 😛 

If there really is something major that we could not have foreseen then we can revert the change and regroup from there.

Mouvedia: > If there really is something major that we could not have foreseen then we can revert the change and regroup from there.

That's what I was trying to avoid.
i.e. no regression test has been foreseen

romainmenke (approved):
LGTM! Thank you for this 🙇

ybiquitous:
Yes, please add a changelog item.

And is warning more proper than notice?

Mouvedia:
> And is warning more proper than notice?

It's not because it could be confused with real warnings.
i.e. it's shown even when `warnings.length === 0`

PR #7912, opened 2024-08-09
Add 2 tests

Mouvedia:

#5742, #6712

`grid-template-rows`, `grid-template-columns` and `grid-template` support `[<custom-ident>*]` values.
I was expecting to have to update [this set](https://github.com/stylelint/stylelint/blob/main/lib/rules/custom-property-no-missing-var-function/index.mjs#L18) but it wasn't necessary.


romainmenke:
@Mouvedia can you split this change in three?
These changes do not have a common theme or purpose.

Mouvedia:
> These changes do not have a common theme or purpose.

@romainmenke these are grouped in the same PR because they are all minor.
i.e. no changeset necessary

> can you split this change in three?

If we don't squash and keep the commits, it can be merged as is.

romainmenke:
We always squash when merging :)
https://github.com/stylelint/stylelint/blob/main/docs/maintainer-guide/pull-requests.md#merging

Splitting them up makes it easier to correctly describe changes and also to review them.
Currently this PR is misleading because contrary to its title it isn't just a test change.
If they were split it up it would be more straightforward to explain and discuss each change.

Mouvedia:
If you are seeing problems with these seemingly minor changes we can discuss them here.
Ill split them in several PRs when Ill have time.
IMHO always squashing regardless of the situation is not a smart approach.

lib/rules/custom-property-no-missing-var-function/__tests__/index.mjs
```diff
@@ -56,6 +56,10 @@ testRule({
 			code: '@property --foo {} a { transition: --foo; }',
 			description: 'ignore a property that can receive a custom-ident',
 		},
+		{
+			code: 'a { grid-template-rows: [--foo]; }',
```
Mouvedia, line 60:
```suggestion
			code: '@property --foo {} a { grid-template-rows: [--foo]; }',
```

lib/rules/custom-property-no-missing-var-function/__tests__/index.mjs
Mouvedia, line 1: `running()` could have been added to the functions ignored alongside `var()` but no browsers support it.

ref https://chromestatus.com/feature/5400251359821824

ybiquitous:
@Mouvedia Please separate a PR to each PR with a specific purpose and clarify a PR title. This PR may be a bad example for other contributors.

PR #7913, opened 2024-08-09
Fix `selector-type-no-unknown` false positives for custom identifiers passed to `::highlight()` and `::view-transition-*()`

### LINKED ISSUE #7911: Fix `selector-type-no-unknown` false positives for `<custom-ident>` passed to pseudo-elements

```css
html::view-transition-group(lightdark) {
	animation-timing-function:step-end;
}

:root::view-transition-old(lightdark) {
	animation-name: keep-old;
	z-index: 1;
}

:root::view-transition-new(lightdark) {
	animation-name: keep-new;
	z-index: 9999;
}
```

```js
export default {
	extends: [
		'stylelint-config-standard',
		'stylelint-config-html',
	],
	rules: {
	},
}
```

from astro

```
"stylelint": "^16.8.1",
```

No error, since this is [valid CSS syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/::view-transition-group#custom-ident).

### What actually happened?

`Unexpected unknown type selector "lightdark" (selector-type-no-unknown)`

Accept any identifier in those pseudo selectors, since they can be used by the Javascript-based View Transitions API.

Mouvedia:

Closes #7911

N/A

### DIFF
--- a/.changeset/gentle-spoons-scream.md
+++ b/.changeset/gentle-spoons-scream.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-type-no-unknown` false positives for custom identifiers passed to `::highlight()` and `::view-transition-*()`
--- a/.changeset/gentle-spoons-scream.md
+++ b/.changeset/gentle-spoons-scream.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `selector-type-no-unknown` false positives for custom identifiers passed to `::highlight()` and `::view-transition-*()`
+Fixed: `selector-type-no-unknown` false positives for `::highlight()` and `::view-transition-*()`
--- a/lib/rules/selector-type-no-unknown/index.mjs
+++ b/lib/rules/selector-type-no-unknown/index.mjs
@@ -1,3 +1,4 @@
+import parser from 'postcss-selector-parser';
 import svgTags from 'svg-tags';
 
 import { isRegExp, isString } from '../../utils/validateTypes.mjs';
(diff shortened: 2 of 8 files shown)


romainmenke (commented):
Can you add some tests for at least one of the `view-transition-*` pseudo elements?

.changeset/gentle-spoons-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-type-no-unknown` false positives for custom identifiers passed to `::highlight()` and `::view-transition-*()`
```
romainmenke, line 5:
```suggestion
Fixed: `selector-type-no-unknown` false positives for `::highlight()` and `::view-transition-*()`
```

Same for the PR and issue title.

-----

The `custom-ident` aspect is actually irrelevant and inaccurate :)
These pseudo's do not take selectors as their argument.
Although those arguments do sometimes look like selectors.

This is also valid for example: `::view-transition-group(*)`, where `*` is also not a selector.

So we ignore them because their arguments aren't selectors, not because their arguments might be custom-idents.

Simplifying the changelog entry makes it more correct :)

lib/rules/selector-type-no-unknown/index.mjs
```diff
@@ -79,13 +84,13 @@ const rule = (primary, secondaryOptions) => {
...
+				if (optionsMatches(secondaryOptions, 'ignoreNamespaces', tagNode.namespace)) return;
 
-				if (optionsMatches(secondaryOptions, 'ignoreTypes', tagNode.value)) {
-					return;
-				}
+				if (optionsMatches(secondaryOptions, 'ignoreTypes', tagNode.value)) return;
+
+				const value = tagNode.parent?.parent?.value?.toLowerCase() ?? '';
+
+				if (IGNORED_PSEUDO_ELEMENTS.has(value)) return;
```
ybiquitous, line 93: [suggest] I suggest the two points:

1. Extract a helper function in this file for readability. E.g.
   ```suggestion
   				if (isWithinIgnoredPseudoElement(tagNode)) return;
   ```
2. Use the check function of `postcss-selector-parser` for AST safety. E.g.
   ```js
   // ...
   const pseudoElement = tagNode?.parent?.parent;
   if (selectorParser.isPseudoElement(pseudoElement) && IGNORED_PSEUDO_ELEMENTS.has(pseudoElement.value)) return;
   ```

ybiquitous (commented):
@Mouvedia Thanks for the pull request. Overall LGTM 👍🏼 

But I've left a review for refactoring. Can you have a look?

ybiquitous (approved):
Thanks. LGTM 👍🏼

PR #7916, opened 2024-08-10
Fix `selector-max-type` reported ranges

romainmenke:

See:  https://github.com/stylelint/stylelint/issues/7904#issuecomment-2275024780

No


lib/utils/getSelectorMeaningfulSource.mjs
```diff
@@ -0,0 +1,45 @@
...
+	const first = node.nodes[firstIndex];
+	let last = node.nodes[lastIndex];
+
+	if (!first || !last) {
+		const selectorStr = node.toString().trim();
+		const index = node.sourceIndex ?? 0;
+		const endIndex = index + selectorStr.length;
+
+		return {
+			selectorStr,
```
romainmenke, line 23: - `selector`
- `selectorString`
- `source`
- `representation`

?

lib/utils/getSelectorMeaningfulSource.mjs
```diff
@@ -0,0 +1,45 @@
+import selectorParser from 'postcss-selector-parser';
+
+/**
+ * Get the source index of a selector, excluding leading whitespace.
+ *
+ * @param {import('postcss-selector-parser').Selector} node
+ *
+ * @returns {{selectorStr: string, index: number, endIndex: number}}
+ */
+export default function getSelectorMeaningfulSource(node) {
```
romainmenke, line 10: - `getSelectorTrimmed`
- `getSelectorTrimmedSource`
- `getTrimmedSelectorSource`

?

lib/utils/getSelectorMeaningfulSource.mjs
```diff
@@ -0,0 +1,45 @@
...
+	const firstIndex = node.nodes.findIndex((x) => !selectorParser.isComment(x));
+	const lastIndex = node.nodes.findLastIndex((x) => !selectorParser.isComment(x));
+
+	const first = node.nodes[firstIndex];
+	let last = node.nodes[lastIndex];
+
+	if (!first || !last) {
```
romainmenke, line 17: When the selector has an empty array as `nodes`

lib/utils/getSelectorMeaningfulSource.mjs
```diff
@@ -0,0 +1,45 @@
...
+		const selectorStr = node.toString().trim();
+		const index = node.sourceIndex ?? 0;
```
romainmenke, line 19: `?? 0` when the AST node was constructed instead of being the result of parsing

lib/utils/getSelectorMeaningfulSource.mjs
Mouvedia: `getStrippedSelector`?

lib/utils/getSelectorMeaningfulSource.mjs
```diff
@@ -0,0 +1,45 @@
+import selectorParser from 'postcss-selector-parser';
+
+/**
+ * Get the source index of a selector, excluding leading whitespace.
```
Mouvedia, line 4: question

I thought it was supposed to remove comments too?

lib/utils/getSelectorMeaningfulSource.mjs
Mouvedia: If this gets merged, I would expect a comment in the source.

lib/utils/getSelectorMeaningfulSource.mjs
romainmenke: It does, comment was incorrect, should be clearer now. 🙇

lib/utils/getSelectorMeaningfulSource.mjs
Mouvedia: No need to go hungarian notation; we have JSDoc.
`selector` seems fine to me.

lib/utils/getSelectorMeaningfulSource.mjs
```diff
@@ -0,0 +1,45 @@
...
+
+		return {
+			selectorStr,
+			index,
+			endIndex,
+		};
+	}
+
+	last = last.rawSpaceAfter.length ? last.clone() : last;
+	last.rawSpaceAfter = '';
```
Mouvedia, line 30: I suppose there's no need to set `rawSpaceBefore`?

lib/utils/getSelectorMeaningfulSource.mjs
Mouvedia: No need to go hungarian notation; we have JSDoc.
Maybe `value`?
If anything it's probably the shape of the object returned that we could bikeshed.

package.json
```diff
@@ -188,7 +188,7 @@
     "postcss": "^8.4.41",
     "postcss-resolve-nested-selector": "^0.1.6",
     "postcss-safe-parser": "^7.0.0",
-    "postcss-selector-parser": "^6.1.1",
+    "postcss-selector-parser": "^6.1.2",
```
Mouvedia, line 191: https://github.com/postcss/postcss-selector-parser/commit/fb38876194130c7518790c116886186fc1d0e37f

lib/utils/getSelectorMeaningfulSource.mjs
romainmenke: `selectorStr` is often used by the caller:

```js
const { selectorStr, index, endIndex } = getSelectorMeaningfulSource(selectorNode);

report({
	ruleName,
	result,
	node: ruleNode,
	message: messages.expected,
	index,
	endIndex,
});
```

Because we use destructuring so much it is indeed important to pick a good name here that doesn't isn't already in use for other variables in rules.

lib/utils/getSelectorMeaningfulSource.mjs
romainmenke: It is important not to alter this.
We only want to trim then endings, we don't want to alter the middle.

Altering `rawSpaceBefore` for the last node means that we would alter the middle part of a selector.

package.json
ybiquitous: Nice patch! Once #7928 is merged, please rebase it 🙏🏼

ybiquitous (commented):
@romainmenke Thanks for the pull request. This PR is a bug fix for `selector-max-type` rather than a refactoring, right?

romainmenke:
Ah, indeed, do you prefer that we rename or that I remove the changes for `selector-max-type`?

I had initially planned to make this a more pure refactor. I changed the rule to verify the fix and forgot to clean that up.

ybiquitous:
Let's change this PR title and add a changelog item for this bug fix. 😃

lib/utils/getSelectorMeaningfulSource.mjs
romainmenke: Yes, I will likely submit a patch for this but I am not looking forward to all the type errors everyone will get after that 😅

lib/utils/getSelectorMeaningfulSource.mjs
ybiquitous: I prefer a bit `selector` over `selectorStr` or `value`. And we can use `selectorStr` in destructuring like this:

```js
const { selector: selectorStr, index, endIndex } = getSelectorMeaningfulSource(selectorNode);
```

But this is not a strong opinion. 👍🏼

lib/utils/getSelectorMeaningfulSource.mjs
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Get the source of a selector, excluding surrounding whitespace and comments.
+ *
+ * @param {import('postcss-selector-parser').Selector} node
+ *
+ * @returns {{selectorStr: string, index: number, endIndex: number}}
+ */
+export default function getSelectorMeaningfulSource(node) {
+	const firstIndex = node.nodes.findIndex((x) => !selectorParser.isComment(x));
+	const lastIndex = node.nodes.findLastIndex((x) => !selectorParser.isComment(x));
```
ybiquitous, line 12: [suggest] Super minor. This may be a bit more readable, but you can keep it as-is. 👍🏼 

```js

const { isComment } = selectorParser;
```

```suggestion
	const firstIndex = node.nodes.findIndex((x) => !isComment(x));
	const lastIndex = node.nodes.findLastIndex((x) => !isComment(x));
```

lib/utils/getSelectorMeaningfulSource.mjs
Mouvedia: `selector` is fine too.

romainmenke: `selector` it is!

lib/utils/getSelectorMeaningfulSource.mjs
romainmenke: `getStrippedSelectorSource` seems fine to me!
Thank you both for the suggestions 🙇

ybiquitous (approved):
Thank you. LGTM 👍🏼

romainmenke:
Thank you all for the feedback 🙇

PR #7250, opened 2023-10-17
Refactor to add `isIgnoredFunction` utility

0scvr:
Added `isIgnoredFunction` utility function and replaced duplicated isIgnoredFunction functions throughout codebase.
Also added a unit test for the function. 

> Which issue, if any, is this issue related to?

Closes #7249

> Is there anything in the PR that needs further explanation?

No, it's self-explanatory.

PR #7918, opened 2024-08-11
Refactor to add `isVarFunction` utility and rename `isIgnoredFunction` to `isUrlFunction`

Mouvedia:

#7250

N/A


lib/rules/no-duplicate-at-import-rules/index.mjs
```diff
@@ -36,7 +37,7 @@ const rule = (primary) => {
 
 			// extract uri from url() if exists
 			const uri =
-				firstParam.type === 'function' && firstParam.value === 'url' && firstParam.nodes[0]
+				isValueFunction(firstParam) && firstParam.value === 'url' && firstParam.nodes[0]
```
Mouvedia, line 40: TS doesn't narrow the type when using `isUrlFunction`.

ybiquitous: [suggest] How about changing the return type of `isUrlFunction` like this?

```diff
 /**
  * @param {import('postcss-value-parser').Node} node
- * @returns {boolean}
+ * @returns {node is import('postcss-value-parser').FunctionNode}
  */
 export default function isUrlFunction(node) {
 	return isValueFunction(node) && node.value.toLowerCase() === 'url';
```

Ref:
https://github.com/stylelint/stylelint/blob/ee3118d2460cddf469f959b5438d8aeab7e9584e/lib/utils/typeGuards.mjs#L60-L64

ybiquitous (commented):
@Mouvedia Thanks for the pull request. I like this change.

Additionally, can we use `isVarFunction` also for the following code?

https://github.com/stylelint/stylelint/blob/ee3118d2460cddf469f959b5438d8aeab7e9584e/lib/rules/custom-property-pattern/index.mjs#L57-L59

https://github.com/stylelint/stylelint/blob/ee3118d2460cddf469f959b5438d8aeab7e9584e/lib/rules/no-unknown-custom-properties/index.mjs#L43

lib/rules/no-duplicate-at-import-rules/index.mjs
ybiquitous: The same change can be applied to `isVarFunction`, too.

Mouvedia:
@ybiquitous thanks for the review.
I know what you are gonna say about the force push but I would like to keep the commits.
i.e. not squash them
It's cleaner for refactors to be granular.

lib/rules/color-hex-alpha/index.mjs
```diff
@@ -61,13 +62,6 @@ const rule = (primary) => {
 	};
 };
 
-/**
- * @param {import('postcss-value-parser').Node} node
- */
-function isUrlFunction({ type, value }) {
-	return type === 'function' && value === 'url';
```
Mouvedia, line 68: Technically it could be considered a fix.
The lack of lowercasing is just an oversight.
i.e. not worth a changeset

ybiquitous (approved):
Thank you. LGTM 👍🏼

Mouvedia:
I forgot to remove one cjs file.

ybiquitous:
Oh, our CI doesn't detect unused `.cjs` files. 😓

Mouvedia:
> Oh, our CI doesn't detect unused `.cjs` files. 😓

No one bothers doing refactors.
i.e. not worth looking into it
fix upcoming

ybiquitous (approved):

Mouvedia:
@ybiquitous
I just realised that "Rebase and merge" is "not enabled for this repository".
I don't wanna create a merge commit nor "squash and merge"; could you enable the option or should I separate it into 2 PRs?

ybiquitous:
We have used "squash and merge" except for rare cases:

https://github.com/stylelint/stylelint/blob/2cf322882ef9048a8d10f323bc78333fa7503cab/docs/maintainer-guide/pull-requests.md?plain=1#L25

So yes, you have to separate this PR if you don't want to squash commits.

PR #7922, opened 2024-08-12
Fix `.cjs` files clean up

ybiquitous:

We found this issue in <https://github.com/stylelint/stylelint/pull/7918#issuecomment-2283016175>.

This change eliminates *all* `.cjs` files before the build to prevent us from forgetting to delete unused `.cjs` files.

Verification:

```shell
# Add `.mjs`/`.cjs` files for a module
echo "export default 'foo';" > lib/foo.mjs
npm run build
git add lib/foo.*
git commit -m 'add lib/foo module'

# Remove the module files
rm lib/foo.mjs
npm run build
npm run build-check
```

Then, we see the following message:

```
You must commit the following files:
lib/foo.cjs
lib/foo.mjs
```

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -5,19 +5,20 @@ import { rmSync } from 'node:fs';
 
 import fg from 'fast-glob';
 
+const testFileGlobs = ['!**/__tests__/**', '!lib/testUtils/**'];
+
+// clean up
+for (const cjs of fg.globSync(['lib/**/*.cjs', ...testFileGlobs])) {
+	rmSync(cjs);
+}
+
 const inputFiles = fg.globSync([
 	'lib/**/*.mjs',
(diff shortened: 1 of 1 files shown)


rollup.config.mjs
```diff
@@ -5,20 +5,21 @@ import { rmSync } from 'node:fs';
 
 import fg from 'fast-glob';
 
+const testFileGlobs = ['!**/__tests__/**', '!lib/testUtils/**'];
+
+// clean up
+for (const cjs of fg.globSync(['lib/**/*.cjs', ...testFileGlobs])) {
+	rmSync(cjs);
```
ybiquitous, line 12: [note] Since `cjs` always exists, `{ force: true }` is unnecessary.

romainmenke (approved):
LGTM

PR #7926, opened 2024-08-12
Add details on keys used for `--cache` option in documentation

### LINKED ISSUE #6590: Document how `--cache` handles plugin and Stylelint versions

https://stylelint.io/user-guide/usage/options/#cache has no explanation of if plugin versions and implementations are used, nor has it information about if the Stylelint version, config etc. is used.

Prettier adds more verbose information on what is cached, see https://prettier.io/docs/en/cli.html#--cache

The ideal solution would however be for the caching to be able to handle the Prettier version, Prettier config as well as plugin versions and implementations. The rationale for this is that unless it is done, caching in CI is going to be impossible to use since the cache will not be correct, potentially missing new faults introduced in CI.

3w36zj6:

Closes #6590

I added details on keys used for `--cache` option by consulting the [Prettier documentation](https://prettier.io/docs/en/cli.html#--cache). Additionally, for the hash calculation's implementation, I referred to [`lib/utils/FileCache.mjs`](https://github.com/stylelint/stylelint/blob/1fba291f3b7e88970d54566b360841cd123fc565/lib/utils/FileCache.mjs#L43-L53).


docs/user-guide/options.md
```diff
@@ -97,8 +97,16 @@ CLI flag: `--cache`
 
 Store the results of processed files so that Stylelint only operates on the changed ones. By default, the cache is stored in `./.stylelintcache` in `process.cwd()`.
 
+the following values are used as cache keys:
```
ybiquitous, line 100: [suggest] Super minor typo 😃 

```suggestion
The following values are used as cache keys:
```

ybiquitous (commented):
@3w36zj6 Thanks for the pull request! Can you address my one review, please? But this is almost LGTM 👍🏼

PR #7932, opened 2024-08-13
Fix `keyframe-block-no-duplicate-selectors` reported ranges

romainmenke:

No, it's self-explanatory.


lib/rules/keyframe-block-no-duplicate-selectors/index.mjs
```diff
@@ -26,25 +29,26 @@ const rule = (primary) => {
 			const selectors = new Set();
 
 			atRuleKeyframes.walkRules((keyframeRule) => {
-				const ruleSelectors = keyframeRule.selectors;
+				if (!isStandardSyntaxRule(keyframeRule)) {
```
Mouvedia, line 32: note

`isStandardSyntaxSelector` is called by `isStandardSyntaxRule`.

ybiquitous (approved):
Thanks, LGTM 👍🏼

Issue #7845, opened 2024-07-16
Use `emitDeprecationWarning` for rule deprecations


Mouvedia:
### Request

https://github.com/stylelint/stylelint/pull/6561#discussion_r1678660168

### Reasons

1. for consistency
2. `--no-deprecation` should suppress _all_ deprecation warnings

### Do you have a proposal to fix the bug?

- update
https://github.com/stylelint/stylelint/blob/1ba1a0aeafaa124a4987a0135d511111cec3af0a/lib/lintPostcssResult.mjs#L138-L140
- add a new code to
https://github.com/stylelint/stylelint/blob/1ba1a0aeafaa124a4987a0135d511111cec3af0a/lib/utils/emitDeprecationWarning.mjs#L4-L8

ybiquitous:
@Mouvedia Thanks for opening the issue. If we switched to `emitDeprecationWarning`, would this be a breaking change? If not, we could ship this change in the next *minor* release.

Mouvedia:
I don't know but if we had priority labels this would be _very low_ because we already have the `--quiet-deprecation-warnings` flag.
i.e. not respecting `--no-deprecation` is not a big deal if we have an alternative

ybiquitous:
We have already had the [`--quiet-deprecation-warnings`](https://github.com/stylelint/stylelint/blob/1ba1a0aeafaa124a4987a0135d511111cec3af0a/docs/user-guide/options.md#quietdeprecationwarnings) flag/option, so this isn't a big deal.

If we were more careful, this should be shipped in the next *major* version because the reporting way for deprecated rules would change. See also our [versioning policy](https://github.com/stylelint/stylelint/blob/1ba1a0aeafaa124a4987a0135d511111cec3af0a/docs/about/semantic-versioning.md).

Mouvedia:
> If we were more careful, this should be shipped in the next _major_ version because the reporting way for deprecated rules would change. See also our [versioning policy](https://github.com/stylelint/stylelint/blob/1ba1a0aeafaa124a4987a0135d511111cec3af0a/docs/about/semantic-versioning.md).

@ybiquitous `deprecated` rules are pretty rare outside of stylelint/stylelint, which has none currently.
The only one I could find is [`stylelint-scss/at-import-partial-extension`](https://github.com/stylelint-scss/stylelint-scss/blob/a689b55e41a4c582377f2b291236757e4594b429/src/rules/at-import-partial-extension/index.js#L18).
[...]

PR #7936, opened 2024-08-17
Refactor to add `WarningOptions.StylelintWarningType` union type

Mouvedia:

slightly related to #7845

1. There are 2 places where we set `stylelintType` to `'parseError'`:

https://github.com/stylelint/stylelint/blob/860acdccb666ee3429d8509e596ce81708448b3f/lib/utils/transformSelector.mjs#L13

https://github.com/stylelint/stylelint/blob/860acdccb666ee3429d8509e596ce81708448b3f/lib/utils/parseSelector.mjs#L22

I chose the first one for the formatter tests.
i.e. the second one would have 2 parentheses

https://github.com/stylelint/stylelint/blob/860acdccb666ee3429d8509e596ce81708448b3f/lib/formatters/preprocessWarnings.mjs#L39

The tests wouldn't fail without these changes; I just thought it would be more accurate.

2. `Result.Message`'s index signature is `any` so you won't get any auto-suggestions in your IDE.
3. I didn't update this comment:
 https://github.com/stylelint/stylelint/blob/860acdccb666ee3429d8509e596ce81708448b3f/lib/createPartialStylelintResult.mjs#L61-L62

### DIFF
--- a/lib/formatters/__tests__/githubFormatter.test.mjs
+++ b/lib/formatters/__tests__/githubFormatter.test.mjs
@@ -51,8 +51,8 @@ describe('githubFormatter', () => {
 					{
 						line: 20,
 						column: 1,
-						stylelintType: 'foo-error',
-						text: 'Cannot parse foo',
+						stylelintType: 'ParseError',
+						text: 'Cannot parse selector',
 					},
 				],
 			},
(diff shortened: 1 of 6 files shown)


types/stylelint/index.d.ts
```diff
@@ -168,9 +168,11 @@ declare namespace stylelint {
 		config?: Config;
 	};
 
+	type StylelintType = 'deprecation' | 'invalidOption' | 'parseError';
```
ybiquitous, line 171: [suggest] Since `StylelintType` is too generic, I suggest:

```suggestion
	type StylelintWarningType = 'deprecation' | 'invalidOption' | 'parseError';
```

lib/formatters/__tests__/unixFormatter.test.mjs
```diff
@@ -143,8 +143,8 @@ path/to/file.css:1:1: Unexpected very very very very very very very very very ve
 					{
 						line: 1,
 						column: 1,
-						stylelintType: 'foo-error',
-						text: 'Cannot parse foo',
+						stylelintType: 'ParseError',
```
ybiquitous, line 146: [suggest] Typo?

```suggestion
						stylelintType: 'parseError',
```

ybiquitous (commented):
Thanks for the pull request. Overall LGTM. Can you check my reviews?

lib/formatters/__tests__/githubFormatter.test.mjs
```diff
@@ -51,8 +51,8 @@ describe('githubFormatter', () => {
 					{
 						line: 20,
 						column: 1,
-						stylelintType: 'foo-error',
-						text: 'Cannot parse foo',
+						stylelintType: 'ParseError',
```
Mouvedia, line 54: typo

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #7937, opened 2024-08-17
Fix `selector-no-qualifying-type` reported ranges

romainmenke:

See:
- https://github.com/stylelint/stylelint/issues/7904
- https://github.com/stylelint/stylelint/issues/6234

This was the last one in this series and I was still unhappy with the reduced precision of reported ranges for regular, not-nested, selectors.

For all nested selectors it becomes impossible to know if the problematic part of the selector appears exactly in the same form in source or if it was artificially constructed when resolving the nested selector. So it also becomes impossible to have a report range for a part of a selector.

However this restriction only exists for nested selectors.

I've added a boolean flag to `flattenNestedSelectorsForRule` so that we have better report ranges for regular selectors.

### DIFF
--- a/.changeset/stupid-walls-glow.md
+++ b/.changeset/stupid-walls-glow.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-no-qualifying-type` reported ranges
--- a/lib/utils/flattenNestedSelectorsForRule.cjs
+++ b/lib/utils/flattenNestedSelectorsForRule.cjs
@@ -8,13 +8,19 @@ const getRuleSelector = require('./getRuleSelector.cjs');
 const isStandardSyntaxSelector = require('./isStandardSyntaxSelector.cjs');
 const parseSelector = require('./parseSelector.cjs');
 
+/** @import { Selector, Root } from 'postcss-selector-parser' */
+/** @import { Rule } from 'postcss' */
+/** @import { PostcssResult } from 'stylelint' */
+
 /**
- * @typedef {import('postcss-selector-parser').Selector} Selector
- * @typedef {import('postcss-selector-parser').Root} Root
+ * Flatten the selectors of the current rule against it's parent rules and at-rules.
+ *
+ * The selectors for the current rule are assumed to be standard CSS selectors.
+ *
  * @typedef {{selector: Selector, resolvedSelectors: Root, nested: boolean}} FlattenedSelector
  *
- * @param {import('postcss').Rule} rule
- * @param {import('stylelint').PostcssResult} result
+ * @param {Rule} rule
+ * @param {PostcssResult} result
  * @returns {Array<FlattenedSelector>}
  */
 function flattenNestedSelectorsForRule(rule, result) {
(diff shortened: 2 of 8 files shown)


lib/utils/flattenNestedSelectorsForRule.cjs
```diff
@@ -3,27 +3,45 @@
...
+					nodes: [selectorAST],
+				}),
+				nested: false,
+			});
+
+			continue;
+		}
 
 		for (const resolvedSelector of resolvedSelectors) {
 			if (!isStandardSyntaxSelector(resolvedSelector)) return [];
```
Mouvedia, line 47: question

When the `length === 1` shouldn't `isStandardSyntaxSelector` be called too?

lib/utils/flattenNestedSelectorsForRule.mjs
```diff
@@ -7,20 +8,37 @@ import parseSelector from './parseSelector.mjs';
...
+	/** @type {Array<FlattenedSelector>} */
 	const flattenedSelectors = [];
 
 	for (const selectorAST of ownAST.nodes) {
-		const resolvedSelectors = resolveNestedSelector(selectorAST.toString(), rule);
+		const selector = selectorAST.toString();
+
+		const resolvedSelectors = resolveNestedSelector(selector, rule);
+
+		if (resolvedSelectors.length === 1 && resolvedSelectors[0] === selector) {
```
Mouvedia, line 30: question

When the `length === 1` shouldn't `isStandardSyntaxSelector` be called too?

romainmenke: `isStandardSyntaxSelector` is called on the output of `resolveNestedSelector` to make sure that the resolved selector is still a standard selector.

For example:

```scss
.icon-#{$name} {
  .foo & {
    color: red;
  }
}
```

- `.foo &` is a standard selector
- `.foo .icon-#{$name}` is non-standard

------

We make the assumption that any input to `flattenNestedSelectorsForRule` is already checked to be a standard selector.

(This assumption is true in practice).

Since the result of calling `resolveNestedSelector` was the same as the input we can also assume that the result will be a standard selector.

Mouvedia: > We make the assumption that any input to flattenNestedSelectorsForRule is already checked to be a standard selector.

That's the kind of details you can add to the [`@description`](https://jsdoc.app/tags-description).
BTW you can remove the `@returns`.

.changeset/stupid-walls-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-no-qualifying-type` reported ranges
```
Mouvedia, line 5:
```suggestion
Fixed: `selector-no-qualifying-type` end positions
```

I prefer your wording but it's not consistent with the past changesets.

romainmenke: This changes fixes the entirety of the reported ranges, as raised in https://github.com/stylelint/stylelint/issues/7904

While https://github.com/stylelint/stylelint/issues/6234 was only about the end positions.

The current changelog entry best describes both fixes.

lib/utils/flattenNestedSelectorsForRule.mjs
romainmenke: I've added a description in the doc comment but I didn't use `@description` as no other part of Stylelint uses this. I didn't want to deviate from existing practices.

ybiquitous (approved):
Great! LGTM 👍🏼

PR #7940, opened 2024-08-19
`property-no-vendor-prefix`: handle `-webkit-background-size` as a special case when it has less than two values per background

### LINKED ISSUE #7727: Add exception to `property-no-vendor-prefix` for `-webkit-background-size` when it only has one value

```css
a {
  -ms-interpolation-mode: bicubic;
  -webkit-background-size: 1px;
}
```

```json
{
  "rules": {
    "property-no-vendor-prefix": true
  }
}
```

1. remove this exception:

https://github.com/stylelint/stylelint/blob/d902c0b7cc1233b74cc4959eff864cb3e2d87f73/lib/utils/isAutoprefixable.mjs#L342-L346

2. remove this property:

https://github.com/stylelint/stylelint/blob/d902c0b7cc1233b74cc4959eff864cb3e2d87f73/lib/utils/isAutoprefixable.mjs#L78

### References

the discussion starting with https://github.com/stylelint/stylelint/issues/7294#issuecomment-1793474109
whatwg/compat#28
w3c/csswg-drafts#7802

Mouvedia:

Closes #7727

1. `-webkit-background-size` doesn't support `cover` nor `contain`.
ref: https://developer.mozilla.org/en-US/docs/Web/CSS/background-size
2. The exception for `-ms-interpolation-mode` wasn't needed in the first place.
i.e. `interpolation-mode` doesn't exist and only the properties that can be safely unprefixed should be reported by this rule
ref: https://web.archive.org/web/20140407235146/http://msdn.microsoft.com/en-us/library/ms530822(VS.85).aspx
cf https://github.com/stylelint/stylelint/commit/3a44bff74052414db0e3432a6ba67a0fdf9407a2#diff-e5bf71a0292e1668512d2ca5b489f678f2bb144050241cc80e1351f04a3c5e3bR101
3. `decl.value` is already trimmed
4. codecov wants me to test `-webkit-background-size:;` or ≥3 values which is nonsensical


lib/rules/property-no-vendor-prefix/__tests__/index.mjs
```diff
@@ -35,9 +35,57 @@ testRule({
...
+		},
+		{
+			code: 'a { -moz-background-size: contain; }',
+			fixed: 'a { background-size: contain; }',
+			message: messages.rejected('-moz-background-size'),
+			line: 1,
+			column: 5,
+			endLine: 1,
+			endColumn: 25,
+		},
```
Mouvedia, line 79: see also l197–205

.changeset/few-tigers-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: do not attempt to autofix `-webkit-background-size` for `property-no-vendor-prefix` if it has less than two values per background
```
Mouvedia, line 5:
```suggestion
Fixed: do not report `-webkit-background-size` for `property-no-vendor-prefix` when the resulting autofix is unsafe```

File: .changeset/few-tigers-deliver.md [hunk 3]
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: do not report `-webkit-background-size` for `property-no-vendor-prefix` when the resulting autofix is unsafe```
```
Mouvedia, line 5:
```suggestion
Fixed: do not report `-webkit-background-size` for `property-no-vendor-prefix` when the resulting autofix is unsafe
```

.changeset/few-tigers-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: do not report `-webkit-background-size` for `property-no-vendor-prefix` when the resulting autofix is unsafe
```
ybiquitous, line 5: [suggest] I'd like to put the rule name ahead to make the changelog items sortable. Can you update the description to fit the following format?

```
Fixed: `property-no-vendor-prefix` <bug summary> <bug details>
```

lib/rules/property-no-vendor-prefix/index.mjs
```diff
@@ -57,6 +58,27 @@ const rule = (primary, secondaryOptions) => {
...
+					const length = values.length;
+					const [first] = values;
+
+					switch (length) {
+						case 2:
+							return true;
+						case 1:
+							return first && (first === 'auto' || basicKeywords.has(first));
+						default:
+							return false;
```
ybiquitous, line 75: [ask] Just to make sure, is it possible to add a test case to cover this line?

ybiquitous (commented):
Thanks for the pull request. Overall LGTM, but I've left a few trivial reviews.

.changeset/few-tigers-deliver.md
Mouvedia:
```suggestion
Fixed: `property-no-vendor-prefix` report for `-webkit-background-size`
```

lib/rules/property-no-vendor-prefix/index.mjs
ybiquitous: That makes sense. 👍🏼

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #7943, opened 2024-08-19
Honour `--no-deprecation` flag for deprecated rules

### LINKED ISSUE #7845: Use `emitDeprecationWarning` for rule deprecations

Mouvedia:

Closes #7845

N/A

### DIFF
--- a/.changeset/tall-pumpkins-do.md
+++ b/.changeset/tall-pumpkins-do.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `--no-deprecation` to suppress `stylelint:000` warning
+Fixed: `--no-deprecation` to suppress rule deprecation warnings
--- a/.changeset/tall-pumpkins-do.md
+++ b/.changeset/tall-pumpkins-do.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `--no-deprecation` to suppress `stylelint:000` warning
--- a/.changeset/tall-pumpkins-do.md
+++ b/.changeset/tall-pumpkins-do.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `--no-deprecation` to suppress rule deprecation warnings
+Fixed: honour Node.js `--no-deprecation` flag for rule deprecation warnings
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -319,7 +319,7 @@ For example:
 
 ```js
 result.warn(
-  `"your-namespace/old-rule" has been deprecated and will be removed in 7.0. Use '"your-namespace/new-rule" instead.`,
+  `"your-namespace/old-rule" has been deprecated and will be removed in 7.0. Use "your-namespace/new-rule" instead.`,
   {
     stylelintType: "deprecation",
     stylelintReference:
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -321,9 +321,10 @@ For example:
 result.warn(
   `"your-namespace/old-rule" has been deprecated and will be removed in 7.0. Use '"your-namespace/new-rule" instead.`,
   {
-    stylelintType: 'deprecation',
-    stylelintReference: 'https://github.com/your-org/your-stylelint-plugin/blob/v6.3.0/src/rules/old-rule/README.md'
-  },
+    stylelintType: "deprecation",
+    stylelintReference:
+      "https://github.com/your-org/your-stylelint-plugin/blob/v6.3.0/src/rules/old-rule/README.md"
+  }
 );
 ```
 
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -311,8 +311,9 @@ You should:
 
 Deprecating rules doesn't happen very often. When you do, you must:
 
-1. Point the `stylelintReference` link to the specific version of the rule README on the GitHub website, so that it is always accessible.
-2. Add the appropriate metadata to mark the rule as deprecated like `rule.meta = { deprecated: true }`.
+1. Add the appropriate metadata to mark the rule as deprecated like so: `rule.meta = { deprecated: true }`
+2. Set the `stylelintType` to `deprecation`
+3. Optionaly set `stylelintReference` to a link that points to a specific version of the rule's README on your source-code-hosting facility of choice—e.g. GitHub—so that it always remains accessible.
 
 ## Improve the performance of a rule
 
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -311,9 +311,21 @@ You should:
 
 Deprecating rules doesn't happen very often. When you do, you must:
 
-1. Add the appropriate metadata to mark the rule as deprecated like so: `rule.meta = { deprecated: true }`
-2. Set the `stylelintType` to `deprecation`
-3. Optionaly set `stylelintReference` to a link that points to a specific version of the rule's README on your source-code-hosting facility of choice—e.g. GitHub—so that it always remains accessible.
+1. Add the appropriate metadata to mark the rule as deprecated like so: `rule.meta = { deprecated: true }`.
+2. Set the `stylelintType` to `'deprecation'`.
+3. Optionally set `stylelintReference` to a link that points to a specific version of the rule's document so that it always remains accessible.
+
+For example:
+
+```js
+result.warn(
+  `"your-namespace/old-rule" has been deprecated and will be removed in 7.0. Use '"your-namespace/new-rule" instead.`,
+  {
+    stylelintType: 'deprecation',
+    stylelintReference: 'https://github.com/your-org/your-stylelint-plugin/blob/v6.3.0/src/rules/old-rule/README.md'
+  },
+);
+```
 
 ## Improve the performance of a rule
 
--- a/lib/__tests__/standalone-quiet-deprecation-warnings.test.mjs
+++ b/lib/__tests__/standalone-quiet-deprecation-warnings.test.mjs
@@ -45,6 +45,16 @@ it('standalone does not silence deprecation warnings by default', async () => {
 });
 
 describe('flags', () => {
+	let originalNoDeprecation;
+
+	beforeEach(() => {
+		originalNoDeprecation = process.noDeprecation;
+	});
+
+	afterEach(() => {
+		process.noDeprecation = originalNoDeprecation;
+	});
+
 	it('standalone silences deprecation warnings when passed --quiet-deprecation-warnings', async () => {
 		const config = {
 			rules: {
@@ -63,7 +73,7 @@ describe('flags', () => {
 	});
 
 	it('standalone silences deprecation warnings when passed --no-deprecation', async () => {
-		process.noDeprecation = true;
+		jest.replaceProperty(process, 'noDeprecation', true);
 
 		const config = {
 			rules: {
@@ -78,10 +88,6 @@ describe('flags', () => {
 
 		expect(results[0].deprecations).toHaveLength(0);
 		expect(results[0].warnings).toHaveLength(1);
-	});
-
-	afterEach(() => {
-		process.noDeprecation = undefined;
 	});
 });
 
--- a/lib/__tests__/standalone-quiet-deprecation-warnings.test.mjs
+++ b/lib/__tests__/standalone-quiet-deprecation-warnings.test.mjs
@@ -21,6 +21,7 @@ const { jest } = import.meta;
 jest.unstable_mockModule('../rules/block-no-empty/index.mjs', () => ({ default: deprecatedRule }));
 
 it('standalone does not silence deprecation warnings by default', async () => {
+	const mock = jest.spyOn(process, 'emitWarning').mockImplementation(() => {});
 	const config = {
 		rules: {
 			'block-no-empty': true,
@@ -28,9 +29,19 @@ it('standalone does not silence deprecation warnings by default', async () => {
 	};
 
 	const { results } = await standalone({ code: 'a { color: #FFF }', config });
+	const text = 'The "block-no-empty" rule is deprecated.';
 
 	expect(results[0].deprecations).toHaveLength(1);
+	expect(results[0].deprecations[0].text).toBe(text);
 	expect(results[0].warnings).toHaveLength(1);
+	expect(mock).toHaveBeenCalledWith(text, {
+		code: 'stylelint:000',
+		detail:
+			'Please be aware that the "block-no-empty" rule will soon be either removed or renamed.',
+		type: 'DeprecationWarning',
+	});
+
+	mock.mockClear();
 });
 
 it('standalone silences deprecation warnings when passed --quiet-deprecation-warnings', async () => {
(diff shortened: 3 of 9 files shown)


lib/lintPostcssResult.mjs
```diff
@@ -137,5 +138,15 @@ function isFixCompatible({ stylelint: { disabledRanges } }, name) {
  * @returns {void}
  */
 function warnDeprecatedRule(result, ruleName) {
```
Mouvedia, line 140: off-topic

I am not sure, but it seems like this function may produces duplicates warnings for plugins.
e.g. https://github.com/stylelint-scss/stylelint-scss/blob/a689b55e41a4c582377f2b291236757e4594b429/src/rules/at-import-partial-extension/index.js#L51-L59

If that's the case, it might be wiser to modify it, in another PR, so that it only warns for core rules.

lib/lintPostcssResult.mjs
```diff
@@ -137,5 +138,15 @@ function isFixCompatible({ stylelint: { disabledRanges } }, name) {
...
+
+	emitDeprecationWarning(
+		message,
+		'RULE',
+		`Please be aware that the "${ruleName}" rule will soon be either removed or renamed.`,
+	);
+
+	// @ts-expect-error
+	// https://github.com/nodejs/node/blob/main/doc/api/process.md#processnodeprecation
+	if (!process.noDeprecation) result.warn(message, { stylelintType: 'deprecation' });
```
Mouvedia, line 151: ref https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/process.d.ts

docs/developer-guide/rules.md
```diff
@@ -311,8 +311,9 @@ You should:
 
 Deprecating rules doesn't happen very often. When you do, you must:
 
-1. Point the `stylelintReference` link to the specific version of the rule README on the GitHub website, so that it is always accessible.
-2. Add the appropriate metadata to mark the rule as deprecated like `rule.meta = { deprecated: true }`.
+1. Add the appropriate metadata to mark the rule as deprecated like so: `rule.meta = { deprecated: true }`
+2. Set the `stylelintType` to `deprecation`
+3. Optionaly set `stylelintReference` to a link that points to a specific version of the rule's README on your source-code-hosting facility of choice—e.g. GitHub—so that it always remains accessible.
```
Mouvedia, line 316: see https://github.com/stylelint/stylelint/pull/6561#pullrequestreview-1248758992

.changeset/tall-pumpkins-do.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `--no-deprecation` to suppress `stylelint:000` warning
```
Mouvedia, line 5:
```suggestion
Fixed: `--no-deprecation` to suppress rule deprecation warnings
```

lib/utils/emitDeprecationWarning.mjs
```diff
@@ -2,6 +2,7 @@ import process from 'node:process';
 
 // Each code must be unique.
 const CODES = {
+	RULE: '000', // either removal or renaming
```
ybiquitous, line 5: [suggest] I think using the normally incremented number is fine because `000` looks like a special number. E.g.

```suggestion
	RULE: '006',
```

docs/developer-guide/rules.md
```diff
@@ -311,8 +311,9 @@ You should:
 
 Deprecating rules doesn't happen very often. When you do, you must:
 
-1. Point the `stylelintReference` link to the specific version of the rule README on the GitHub website, so that it is always accessible.
-2. Add the appropriate metadata to mark the rule as deprecated like `rule.meta = { deprecated: true }`.
+1. Add the appropriate metadata to mark the rule as deprecated like so: `rule.meta = { deprecated: true }`
+2. Set the `stylelintType` to `deprecation`
```
ybiquitous, line 315: [suggest] Minor:

```suggestion
1. Add the appropriate metadata to mark the rule as deprecated like so: `rule.meta = { deprecated: true }`.
2. Set the `stylelintType` to `'deprecation'`.
```

docs/developer-guide/rules.md
ybiquitous: [suggest] How about simplifying the description a bit more and adding an example to help people understand it? E.g.

````suggestion
3. Optionally set `stylelintReference` to a link that points to a specific version of the rule's document so that it always remains accessible.

For example:

```js
result.warn(
  `"your-plugin/old-rule" has been deprecated and will be removed in 7.0. Use '"your-plugin/new-rule" instead.`,
  {
    stylelintType: 'deprecation',
    stylelintReference: 'https://github.com/your-org/your-stylelint-plugin/blob/v6.3.0/src/rules/old-rule/README.md'
  },
);
```
````

File: lib/lintPostcssResult.mjs (hunk 1 above)
Comment (ybiquitous) [reply lib/lintPostcssResult.mjs:140]: This point is still important. The current Stylelint collects deprecation warnings in a PostCSS result and shows them through the `string` formatter:

https://github.com/stylelint/stylelint/blob/30e1027b93520193b73a692394db124132144ba5/lib/createPartialStylelintResult.mjs#L25-L33

https://github.com/stylelint/stylelint/blob/30e1027b93520193b73a692394db124132144ba5/lib/formatters/stringFormatter.mjs#L45-L47

So, I think we have to fix the formatter code in this PR to work `--no-deprecation` and leave `result.warn()` as-is here to keep backward compatibility of a lint result.

File: .changeset/tall-pumpkins-do.md [hunk 7]
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `--no-deprecation` to suppress rule deprecation warnings
```
ybiquitous, line 5: [suggest] How about adding a few words to help people distinguish which flag of Node.js or Stylelint?

```suggestion
Fixed: Node.js `--no-deprecation` flag to suppress rule deprecation warnings
```

lib/lintPostcssResult.mjs
ybiquitous: It looks like a contribution chance to `@types/node`. 😃

lib/lintPostcssResult.mjs
```diff
@@ -137,5 +138,15 @@ function isFixCompatible({ stylelint: { disabledRanges } }, name) {
...
-	result.warn(`The "${ruleName}" rule is deprecated.`, { stylelintType: 'deprecation' });
+	const message = `The "${ruleName}" rule is deprecated.`;
+
+	emitDeprecationWarning(
+		message,
+		'RULE',
+		`Please be aware that the "${ruleName}" rule will soon be either removed or renamed.`,
+	);
+
+	// @ts-expect-error
```
ybiquitous, line 149: [suggest] Let's add a detailed error to help understand the reason for suppressing the error:

```suggestion
	// @ts-expect-error -- TS2339: Property 'noDeprecation' does not exist on type 'Process'.
```

ybiquitous (changes requested):
@Mouvedia Thanks for the pull request! Can you have a look at my reviews?

lib/utils/emitDeprecationWarning.mjs
Mouvedia: It's the oldest type of deprecation, that's why I chose `000`.
But `006` will do too.

docs/developer-guide/rules.md
Mouvedia:
````suggestion
3. Optionally set `stylelintReference` to a link that points to a specific version of the rule's document so that it always remains accessible.

For example:

```js
result.warn(
  `"your-namespace/old-rule" has been deprecated and will be removed in 7.0. Use '"your-namespace/new-rule" instead.`,
  {
    stylelintType: 'deprecation',
    stylelintReference: 'https://github.com/your-org/your-stylelint-plugin/blob/v6.3.0/src/rules/old-rule/README.md'
  },
);
```
````

File: lib/lintPostcssResult.mjs (hunk 1 above)
Comment (Mouvedia) [reply lib/lintPostcssResult.mjs:140]: I don't understand your point.
I am saying we might have 2 warnings for each deprecation.
Am I correct in my assumption?
If so, I am proposing to fix it in another PR.
This PR is about adding `006` for rule deprecation and suppressing `warn` when `noDeprecation` is `true`.

Comment (Mouvedia) [reply lib/lintPostcssResult.mjs:140]: Are you proposing to add another `if (process.noDeprecation)`?

File: lib/lintPostcssResult.mjs (hunk 8 above)
Comment (Mouvedia) [reply lib/lintPostcssResult.mjs:149]: Actually the comment under this one shows that it's wrong.
It does exist and it's just missing from the d.ts.
This message would be misleading.

File: lib/lintPostcssResult.mjs (hunk 2 above)
Comment (Mouvedia) [reply lib/lintPostcssResult.mjs:151]: Be my guest.
I don't want to be the guy that will generate errors all over the word due to the `// @ts-expect-error` suddenly not being accurate anymore.
😃

File: .changeset/tall-pumpkins-do.md (hunk 7 above)
Comment (Mouvedia) [reply .changeset/tall-pumpkins-do.md:5]:
```suggestion
Fixed: honour Node.js `--no-deprecation` flag for rule deprecation warnings
```

lib/lintPostcssResult.mjs
ybiquitous: I don't think the message would be misleading because `tsc` outputs the error message exactly. We would be rather a bit confused if `@ts-expect-error` has no reason (e.g. "What error does this `@ts-expect-error` suppress?").

Mouvedia: So we are adding as the reason the exact opposite of why it's being added?
I guess, I could add it, sure.

lib/lintPostcssResult.mjs
Mouvedia: I moved the `if (!process.noDeprecation)` in `createPartialStylelintResult.mjs`.

lib/lintPostcssResult.mjs
ybiquitous: If adding "why" more exactly, we should include why `@types/note` doesn't have `process.noDeprecation`. However, the link to the Node.js doc is enough for me.

docs/developer-guide/rules.md
```diff
@@ -311,8 +311,22 @@ You should:
...
-2. Add the appropriate metadata to mark the rule as deprecated like `rule.meta = { deprecated: true }`.
+1. Add the appropriate metadata to mark the rule as deprecated like so: `rule.meta = { deprecated: true }`.
+2. Set the `stylelintType` to `'deprecation'`.
+3. Optionally set `stylelintReference` to a link that points to a specific version of the rule's document so that it always remains accessible.
+
+For example:
+
+```js
+result.warn(
+  `"your-namespace/old-rule" has been deprecated and will be removed in 7.0. Use '"your-namespace/new-rule" instead.`,
```
Mouvedia, line 322:
```suggestion
  `"your-namespace/old-rule" has been deprecated and will be removed in 7.0. Use "your-namespace/new-rule" instead.`,
```

lib/createPartialStylelintResult.mjs
```diff
@@ -22,9 +23,12 @@ export default function createPartialStylelintResult(postcssResult, cssSyntaxErr
 			}
 		}
 
-		const deprecationMessages = postcssResult.messages.filter(
-			(message) => message.stylelintType === 'deprecation',
-		);
+		// @ts-expect-error -- TS2339: Property 'noDeprecation' does not exist on type 'Process'.
```
Mouvedia, line 26:
```suggestion
		// @ts-ignore -- TS2339: Property 'noDeprecation' does not exist on type 'Process'.

File: lib/createPartialStylelintResult.mjs [hunk 11]
```diff
@@ -22,9 +23,12 @@ export default function createPartialStylelintResult(postcssResult, cssSyntaxErr
 			}
 		}
 
-		const deprecationMessages = postcssResult.messages.filter(
-			(message) => message.stylelintType === 'deprecation',
-		);
+		// @ts-ignore -- TS2339: Property 'noDeprecation' does not exist on type 'Process'.
```
ybiquitous, line 26: [suggest] I think `@ts-expect-error` has more advantages. See:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#ts-ignore-or-ts-expect-error

```suggestion
		// @ts-expect-error -- TS2339: Property 'noDeprecation' does not exist on type 'Process'.
```

lib/__tests__/standalone-quiet-deprecation-warnings.test.mjs
```diff
@@ -21,33 +21,68 @@ const { jest } = import.meta;
...
+			config,
+		});
+
+		expect(results[0].deprecations).toHaveLength(0);
+		expect(results[0].warnings).toHaveLength(1);
+	});
+
+	afterEach(() => {
+		process.noDeprecation = undefined;
+	});
```
ybiquitous, line 85: [suggest] If doing this more exactly, we should restore the original `process.noDeprecation` like this:

```js
let originalNoDeprecation;

beforeEach(() => { originalNoDeprecation = process.noDeprecation });

afterEach(() => { process.noDeprecation = originalNoDeprecation });
```

Otherwise, using [`jest.replaceProperty`](https://jestjs.io/docs/jest-object#jestreplacepropertyobject-propertykey-value) may be more smart.

Mouvedia: > [suggest] If doing this more exactly, we should restore the original `process.noDeprecation` like this:
> 
> ```js
> let originalNoDeprecation;
> 
> beforeEach(() => { originalNoDeprecation = process.noDeprecation });
> 
> afterEach(() => { process.noDeprecation = originalNoDeprecation });
> ```

👍 

> Otherwise, using [`jest.replaceProperty`](https://jestjs.io/docs/jest-object#jestreplacepropertyobject-propertykey-value) may be more smart.

This gives me

> Property `noDeprecation` does not exist in the provided object

lib/createPartialStylelintResult.mjs
ybiquitous: We don't need to release a patch version. Once a `@types/node` bump PR comes and `@ts-expect-error` becomes an error, we just need to remove the `@ts-expect-error` comment in the bump PR.

If using `@ts-ignore`, it's hard for us to notice the needless `@ts-expect-error`.

Mouvedia: > We don't need to release a patch version. 
> […] we just need to remove the `@ts-expect-error` comment in the bump PR.

That would be a patch release :)
It seems you are confident about it, let's change it then.

lib/createPartialStylelintResult.cjs
```diff
@@ -26,9 +27,12 @@ function createPartialStylelintResult(postcssResult, cssSyntaxError) {
 			}
 		}
 
-		const deprecationMessages = postcssResult.messages.filter(
-			(message) => message.stylelintType === 'deprecation',
-		);
+		// @ts-ignore -- TS2339: Property 'noDeprecation' does not exist on type 'Process'.
```
Mouvedia, line 30:
```suggestion
		// @ts-expect-error -- TS2339: Property 'noDeprecation' does not exist on type 'Process'.
```

lib/createPartialStylelintResult.mjs
ybiquitous: > That would be a patch release :)

I don't know why just removing the `@ts-expect-error` comment will be a patch release. 🤷🏼

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #7944, opened 2024-08-19
Fix `declaration-property-value-no-unknown` false positives

### LINKED ISSUE #6667: Fix `declaration-property-value-no-unknown` false positives for `overflow: overlay`

https://stylelint.io/demo/#N4Igxg9gJgpiBcIBG0CeACYAdAdu9EAbjAE4BmANhAO7wHEkUCGqA3LgL64gA0IAzqhwAXJgA8E4fv17gIOMgEsA5pOx50WECQCuFGPy111+fFthhmJJsMXyAtAAcSER6WGp7hJhR0x7OBD2OjgA1oHUOEbowrowuPhcOBwgHEA

This is a lesser-known property that is still supported in Blink. [MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#overlay).

See demo

demo

15.1.0

No error

### What actually happened?

`Unexpected unknown value "overlay" for property "overflow"`

No

### Proposal to fix the bug

Add the property to upstream dependencies. I checked `csstree` and `mdn/data` repos, but wasn't able to find where the definition for valid `overflow` values actually resides. Pointers appreciated.

### LINKED ISSUE #7939: Fix `declaration-property-value-no-unknown` false positives for `word-break: auto-phrase`

```css
.foo {
  word-break: auto-phrase;
}
```

```json
{
  "rules": {
  }
}
```

[word-break: auto phrase - Demo | Stylelint](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0AzCEAEwA6A7bALgO4QBOUAtAEakwCGA1vBnQK4EQUAOAFqXQGcYAbjwBfPCAA0IFAEsANjABydALZxEMAB7quSpGAEDp4CLnkBzBCBy4MGbCFKslAp8zsOHT2GAV0-ARy5tykEFwwpAQAnhQAbnQKrDAUuJysuAzpRPgIGAQuMHgOErhippAWcpYAYmRqdAQ2AFYC5qawXCaIXo4gArFKCnK4BB79AQQwg05SJf2DMcOjBBRVVhSDdLhQgVATTlMz4yDiIGJAA)

stylelint 16.8.2 locally, and in the Stylelint demo, 16.8.2 too.

No problems to be reported.

### What actually happened?

Unexpected unknown value "auto-phrase" for property "word-break"
[(declaration-property-value-no-unknown)](https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown)

_No response_

### More information

On [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break), it's listed as experimental.

Mouvedia:

Closes #6667
Closes #7939

[I warned about the very situation we are in](https://github.com/stylelint/stylelint/pull/6511#issuecomment-1407586913) when we were adding this rule.
I was getting impatient so I went ahead and fixed some false positives.
If `css-tree` eventually catches up, we will remove the default extension.
I only added properties that have at least some browser support.
i.e. not block-step-insert, block-step-size, line-fit-edge, text-autospace, text-group-align, etc.

https://caniuse.com/css-text-box-trim
https://caniuse.com/mdn-css_properties_text-spacing-trim
https://caniuse.com/mdn-css_properties_text-wrap-style
https://caniuse.com/mdn-css_properties_field-sizing
https://developer.mozilla.org/en-US/docs/Web/CSS/@property/syntax

The last version of `css-tree` was released in 2022, which predates these PRs:

- csstree/csstree#242
- csstree/csstree#248

### DIFF
--- a/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
+++ b/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
@@ -358,7 +358,7 @@ testRule({
 			code: 'a { text-box-trim: none; }',
 		},
 		{
-			code: 'a { text-spacing-trim: space-first }',
+			code: 'a { text-spacing-trim: space-first; }',
 		},
 		{
 			code: 'a { text-wrap-style: stable; }',
--- a/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
+++ b/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
@@ -57,6 +57,18 @@ testRule({
 			code: 'a { font-size: max(1rem, 3rem); }',
 		},
 		{
+			code: 'a { word-break: auto-phrase; }',
+		},
+		{
+			code: 'a { field-sizing: content; }',
+		},
+		{
+			code: 'a { overflow: overlay; }',
+		},
+		{
+			code: 'a { width: min-intrinsic; }',
+		},
+		{
 			code: stripIndent`
 				a { font-weight: bolder; }
 				@font-face { font-weight: 100 200; }
@@ -135,6 +147,14 @@ testRule({
 	],
 
 	reject: [
+		{
+			code: 'a { text-box-edge: text ex; }',
+			message: messages.rejected('text-box-edge', 'ex'),
+			line: 1,
+			column: 25,
+			endLine: 1,
+			endColumn: 27,
+		},
 		{
 			code: 'a { top: unknown; }',
 			message: messages.rejected('top', 'unknown'),
@@ -327,6 +347,21 @@ testRule({
 	accept: [
 		{
 			code: 'a { size: 10px; }',
+		},
+		{
+			code: 'a { text-box-edge: ideographic-ink; }',
+		},
+		{
+			code: 'a { text-box-edge: ex text; }',
+		},
+		{
+			code: 'a { text-box-trim: none; }',
+		},
+		{
+			code: 'a { text-spacing-trim: space-first }',
+		},
+		{
+			code: 'a { text-wrap-style: stable; }',
 		},
 	],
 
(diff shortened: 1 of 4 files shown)


lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
```diff
@@ -328,6 +348,21 @@ testRule({
...
+			code: 'a { text-box-edge: ideographic-ink; }',
+		},
+		{
+			code: 'a { text-box-edge: ex text; }',
+		},
+		{
+			code: 'a { text-box-trim: none; }',
+		},
+		{
+			code: 'a { text-spacing-trim: space-first }',
```
Mouvedia, line 361:
```suggestion
			code: 'a { text-spacing-trim: space-first; }',
```

lib/rules/declaration-property-value-no-unknown/index.mjs
```diff
@@ -62,10 +62,19 @@ const rule = (primary, secondaryOptions) => {
 		};
 
 		const propertiesSyntax = {
-			// Take a shallow clone as this object will be appended to.
-			...(secondaryOptions?.propertiesSyntax ?? {}),
+			overflow: '| overlay', // csstree/csstree#248
+			width: '| min-intrinsic | -moz-min-content | -moz-available | -webkit-fill-available', // csstree/csstree#242
+			'field-sizing': 'content | fixed',
+			'text-box-edge':
+				'[ text | cap | ex | ideographic | ideographic-ink ] [ text | alphabetic | ideographic | ideographic-ink ]?',
```
ybiquitous, line 69: [ask] Does `text-box-edge` support `auto`?
Ref https://drafts.csswg.org/css-inline-3/#text-box-edge

ybiquitous (commented):
Thank you. Overall LGTM. I've left one question.

ybiquitous (approved):
Great work! LGTM 👍🏼

PR #7948, opened 2024-08-22
`color-function-notation`: add more criteria to the function that determines the notation

### LINKED ISSUE #7945: `color-function-notation`: do not attempt to autofix when the syntax is invalid

stylelint --fix changes ```a { color: rgb(white, .5) }``` to ```a { color: rgb(white .5) }```, which sass fails to compile.

```json
{ "rules": { "color-function-notation": "modern" } }
```

```sh
pnpm exec stylelint path/to/file.scss -c path/to/stylelint.config.js --fix
```

```json
{
    "stylelint": "16.8.2",
    "stylelint-config-standard-scss": "13.1.0",
    "stylelint-order": "6.0.4",
    "stylelint-scss": "6.5.0",
}
```

stylelint should suggest a valid syntax in this case (probably `rgb(white, 50%)`)

### What actually happened?

stylelint replaces `rgb(white, .5)` with `rgb(white .5)` which sass fails to compile

_No response_

Mouvedia:

Closes #7945

N/A

### DIFF
--- a/lib/rules/color-function-notation/index.mjs
+++ b/lib/rules/color-function-notation/index.mjs
@@ -63,11 +63,11 @@ const rule = (primary, secondaryOptions) => {
 
 				const { value, sourceIndex, sourceEndIndex, nodes } = node;
 
-				if (!nodes.length) return;
-
 				if (ignoreWithVarInside && containsVariable(nodes)) return;
 
 				if (!LEGACY_NOTATION_FUNCTION_NAME.test(value)) return;
+
+				if (nodes.length < 5) return;
 
 				if (primary === 'modern' && !isLikelyLegacy(nodes)) return;
 
@@ -146,8 +146,6 @@ function isComma(node) {
 
 /** @param {import('postcss-value-parser').Node[]} nodes */
 function isLikelyLegacy(nodes) {
-	if (nodes.length < 5) return false;
-
 	let commas = 0;
 
 	for (const node of nodes) {
--- a/lib/rules/color-function-notation/index.mjs
+++ b/lib/rules/color-function-notation/index.mjs
@@ -63,13 +63,15 @@ const rule = (primary, secondaryOptions) => {
 
 				const { value, sourceIndex, sourceEndIndex, nodes } = node;
 
+				if (!nodes.length) return;
+
 				if (ignoreWithVarInside && containsVariable(nodes)) return;
 
 				if (!LEGACY_NOTATION_FUNCTION_NAME.test(value)) return;
 
-				if (primary === 'modern' && !hasCommas(node)) return;
+				if (primary === 'modern' && !isLikelyLegacy(nodes)) return;
 
-				if (primary === 'legacy' && hasCommas(node)) return;
+				if (primary === 'legacy' && isLikelyLegacy(nodes)) return;
 
 				const modernFixer = () => {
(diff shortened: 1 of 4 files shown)


ybiquitous:
> 1. No need for new tests because we are not officially supporting Sass.

I think adding a case like `rgb(white, 50%)` is still worth it because such an invalid CSS is not limited to Sass, for example, a user's misunderstanding etc.

lib/rules/color-function-notation/index.mjs
```diff
@@ -142,11 +144,18 @@ function isComma(node) {
...
+	if (nodes.length < 5) return false;
+
+	let commas = 0;
+
+	for (const node of nodes) {
+		if (isComma(node)) commas++;
+		else if (node.value === '/') return false;
+	}
+
+	return commas && (commas === 2 || commas === 3);
```
ybiquitous, line 158: [suggest] `commas` (equals `commas > 0`) seems unnecessary: 👀 

```suggestion
	return (commas === 2 || commas === 3);
```

Mouvedia: That's an early return when you have a modern notation.
i.e. `=== 0`

Mouvedia:
> > 1. No need for new tests because we are not officially supporting Sass.
> 
> I think adding a case like `rgb(white, 50%)` is still worth it because such an invalid CSS is not limited to Sass, for example, a user's misunderstanding etc.

I disagree. I think it sets a bad precedent.
i.e. we are only fixing regressions related to Sass/Less at this point
If the category of the rule was [invalid](https://stylelint.io/user-guide/rules/#invalid), I would agree with you.
This rule should do the strict minimum.
i.e. performance over perfection

Why should I add `a { color: rgba(,,,); }` to the tests?
What does it accomplish?

lib/rules/color-function-notation/index.mjs
Mouvedia: It's an early return so that you don't have to check the second condition.

ybiquitous:
I don't think stricter test cases are required, but this PR causes some behavior change, so I think we still need some test cases to prove the behavior change.

ybiquitous (commented):
Thanks, overall LGTM. Besides, can you add a changelog item? This PR fixes `color-function-notation` false positives.

lib/rules/color-function-notation/index.mjs
```diff
@@ -67,9 +67,13 @@ const rule = (primary, secondaryOptions) => {
 
 				if (!LEGACY_NOTATION_FUNCTION_NAME.test(value)) return;
 
-				if (primary === 'modern' && !hasCommas(node)) return;
+				const hasEnoughArguments = nodes.length >= 5;
 
-				if (primary === 'legacy' && hasCommas(node)) return;
+				if (!hasEnoughArguments) return;
+
+				if (primary === 'modern' && !isLikelyLegacy(nodes)) return;
```
romainmenke, line 74: Maybe assign the result of `isLikelyLegacy(nodes)` to a variable so that it can be reused?

romainmenke: nvm :)

Only one is ever executed because `primary` is always `modern` or `legacy`.

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #7950, opened 2024-08-22
Change `secondaryOptions` argument type of the `Rule` function

Mouvedia:

N/A

While changing the type of `RuleMeta['fixable']`, I stumbled on a typo in `selector-class-pattern/index.mjs` which lead me to a wrongly defined type in `index.d.ts`.

This PR:

- fixes a typo
- changes `secondaryOptions`
- aligns `value-no-vendor-prefix` with the other 2 rules
- blocks #7954

Only 3 rules were using the second parameter—`S`:
- `selector-disallowed-list`
- `selector-class-pattern`
- `value-no-vendor-prefix`

`RuleBase` is internal so it is safe to update.
i.e. `Rule` wouldn't be

### DIFF
--- a/.changeset/poor-boats-reflect.md
+++ b/.changeset/poor-boats-reflect.md
@@ -2,4 +2,4 @@
 "stylelint": minor
 ---
 
-Changed: `secondaryOptions` type
+Changed: `secondaryOptions` argument type of the `Rule` function
--- a/.changeset/poor-boats-reflect.md
+++ b/.changeset/poor-boats-reflect.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Changed: `secondaryOptions` type
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -246,7 +246,7 @@ declare namespace stylelint {
 	/** @internal */
 	export type RuleBase<P = any, S = any> = (
 		primaryOption: P,
-		secondaryOptions: S,
+		secondaryOptions: S extends Record<string, any> ? S : never,
 		context: RuleContext,
 	) => (root: PostCSS.Root, result: PostcssResult) => Promise<void> | void;
 
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -243,7 +243,14 @@ declare namespace stylelint {
 		newline?: string | undefined;
 	};
 
-	type Primary = string | number | boolean | RegExp | Array<string | RegExp> | Record<string, any>;
+	type Primary =
+		| string
+		| number
+		| true
+		| null
+		| RegExp
+		| Array<string | RegExp>
+		| Record<string, any>;
 
 	/** @internal */
 	export type RuleBase<P extends Primary = any, S = any> = (
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -246,7 +246,7 @@ declare namespace stylelint {
 	/** @internal */
 	export type RuleBase<P = any, S = any> = (
 		primaryOption: P,
-		secondaryOptions: S extends Record<string, any> ? S : never,
+		secondaryOptions: S,
 		context: RuleContext,
 	) => (root: PostCSS.Root, result: PostcssResult) => Promise<void> | void;
 
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -243,14 +243,7 @@ declare namespace stylelint {
 		newline?: string | undefined;
 	};
 
-	type Primary =
-		| string
-		| number
-		| true
-		| null
-		| RegExp
-		| Array<string | RegExp>
-		| Record<string, any>;
+	type Primary = string | number | true | RegExp | Array<string | RegExp> | Record<string, any>;
 
 	/** @internal */
 	export type RuleBase<P extends Primary = any, S = any> = (
@@ -259,13 +252,11 @@ declare namespace stylelint {
 		context: RuleContext,
 	) => (root: PostCSS.Root, result: PostcssResult) => Promise<void> | void;
 
-	type Fixable<P> = P extends string ? Record<P, boolean> | boolean : boolean;
-
 	/** @internal */
-	export type RuleMeta<P = any> = {
+	export type RuleMeta = {
 		url: string;
 		deprecated?: boolean;
-		fixable?: Fixable<P>;
+		fixable?: boolean;
 	};
 
 	/** @internal */
@@ -286,7 +277,7 @@ declare namespace stylelint {
 		ruleName: string;
 		messages: RuleMessages;
 		primaryOptionArray?: boolean;
-		meta?: RuleMeta<P>;
+		meta?: RuleMeta;
 	};
 
 	type BuiltInRules = {
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -243,18 +243,22 @@ declare namespace stylelint {
 		newline?: string | undefined;
 	};
 
+	type Primary = string | number | boolean | RegExp | Array<string | RegExp> | Record<string, any>;
+
 	/** @internal */
-	export type RuleBase<P = any, S = any> = (
+	export type RuleBase<P extends Primary = any, S = any> = (
 		primaryOption: P,
-		secondaryOptions: Record<string, S>,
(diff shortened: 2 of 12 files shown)


ybiquitous:
Is this PR a refactoring or not? I don't understand the reason to change the `fixable` type.

Mouvedia:
> Is this PR a refactoring or not?

It is not.

> I don't understand the reason to change the fixable type.

So that we can introspect/document rules that have partial fixers.
i.e. rules that only have a fixer for certain options
Right now, `fixable` is not granular enough.
It doesn't have much consequences because `fixable: {}` remains truthy.
i.e. `fixable: { foo: false, bar: false }` is `fixable: false` which I have never seen
tl;dr: the only test that I could add would be one that checks whether at least one property is true when fixable is an object

ybiquitous:
Can you please separate the `fixable` change from other changes? I think we should consider the `fixable` change as a minor update because the `fixable?: true` type is exposed both the types and documents.

Mouvedia:
> Can you please separate the `fixable` change from other changes?

OK

Mouvedia:
> Can you please separate the `fixable` change from other changes?

I suppose you mean 2 PRs?

ybiquitous:
Maybe 2 or 3 PRs. I think this PR consists of three parts: `primaryOption`, `secondaryOptions`, and `fixable`. We may combine `primaryOption` and `secondaryOptions`

Also, since the `Rule` type is documented below, I think a changelog item is necessary:

https://github.com/stylelint/stylelint/blob/27b8e16df61e69f0883ea62d220bb39ffc147c70/docs/developer-guide/plugins.md?plain=1#L37-L38

Mouvedia:
> Maybe 2 or 3 PRs. I think this PR consists of three parts: `primaryOption`, `secondaryOptions`, and `fixable`. We may combine `primaryOption` and `secondaryOptions`.
> 
> Also, since the `Rule` type is documented below, I think a changelog item is necessary:
> 
> https://github.com/stylelint/stylelint/blob/27b8e16df61e69f0883ea62d220bb39ffc147c70/docs/developer-guide/plugins.md?plain=1#L37-L38

There is no changes for `primaryOption` because it always fallback to `any`.
i.e. `P extends Primary ? P : any`
By separating the `fixable` change, it's harder to tell why it matters to have `Primary` because `RuleMeta` is not generic anymore.

I agree that by having a separate `fixable` PR, this PR will be easier to review.
tl;dr: only `secondaryOptions` needs a changeset

.changeset/poor-boats-reflect.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Changed: `secondaryOptions` type
```
Mouvedia, line 5: It's changed and not fixed because we had 2 cases that matched the new type and one that matched the old one.

Mouvedia:
@ybiquitous I have split the PR as requested; can you take another look?

Mouvedia:
@ybiquitous I found only one plugin rule in the wild that would be affected by this type change.
It's `@mrhenry/stylelint-mrhenry-nesting`](https://github.com/mrhenry/stylelint-mrhenry/blob/6939dd319ec125ae7ae9b3939bf106e9f39b452a/packages/nesting/index.mjs#L32) which is luckily being maintained by Romain; sorry @romainmenke :)
I am not surprised that the type itself being inconsistently used on our very own codebase wouldn't be adopted by plugin authors. It's obviously less useful in its current form and probably wasn't intended in the first place.

Mouvedia:
@ybiquitous I found only one plugin rule in the wild that would be affected by this type change.
It's [`@mrhenry/stylelint-mrhenry-nesting`](https://github.com/mrhenry/stylelint-mrhenry/blob/6939dd319ec125ae7ae9b3939bf106e9f39b452a/packages/nesting/index.mjs#L32) which is luckily being maintained by Romain; sorry @romainmenke :)
I am not surprised that the type itself being inconsistently used on our very own codebase wouldn't be adopted by plugin authors. It's obviously less useful in its current form and probably wasn't intended in the first place.

ybiquitous:
Thanks for the investigation. Luckily, this type change doesn't seem to have a big impact.

types/stylelint/index.d.ts
```diff
@@ -246,7 +246,7 @@ declare namespace stylelint {
 	/** @internal */
 	export type RuleBase<P = any, S = any> = (
```
ybiquitous, line 247: [suggest] I'm not sure, but is it worth making the `S` default type more specific? For example:

```suggestion
	export type RuleBase<P = any, S = Record<string, any>> = (
```

.changeset/poor-boats-reflect.md
ybiquitous, line 5: [suggest] Minor. The following may be more explicit:

```suggestion
Changed: `secondaryOptions` argument type of the `Rule` function
```

ybiquitous: FYI, the `Rule` type is exposed:
https://github.com/stylelint/stylelint/blob/18dff4f65360b20e434c572850a24d88e035c614/types/stylelint/index.d.ts#L271-L274

romainmenke:
> It's [@mrhenry/stylelint-mrhenry-nesting](https://github.com/mrhenry/stylelint-mrhenry/blob/6939dd319ec125ae7ae9b3939bf106e9f39b452a/packages/nesting/index.mjs#L32) which is luckily being maintained by Romain; sorry @romainmenke :)

That is fine :)
I've removed these type hints for now:

https://github.com/mrhenry/stylelint-mrhenry/blob/884ac304d5a2fd5ed70591d12c5325a1c9e4eb3f/packages/nesting/index.mjs#L32

types/stylelint/index.d.ts
Mouvedia: It's slightly more likely to be more error-prone but I agree with you nonetheless.
i.e. it's the de facto convention

Mouvedia: I don't think it should be the default.
Do you mean `secondaryOptions: S extends Record<string, any> ? S : never,` instead?

types/stylelint/index.d.ts
```diff
@@ -246,7 +246,7 @@ declare namespace stylelint {
...
 		primaryOption: P,
-		secondaryOptions: Record<string, S>,
+		secondaryOptions: S,
```
Mouvedia, line 249:
```suggestion
		secondaryOptions: S extends Record<string, any> ? S : never,
```

see https://github.com/stylelint/stylelint/pull/7950/files#r1730844881

types/stylelint/index.d.ts
ybiquitous: No. If my suggestion cannot get a value, let's leave it as-is. I was just curious.

types/stylelint/index.d.ts
ybiquitous: Sorry for my late comment. Keeping it simple (i.e., `secondaryOptions: S`) seems enough for now.

types/stylelint/index.d.ts
Mouvedia: Are you fine with [17cd6ca](https://github.com/stylelint/stylelint/pull/7950/commits/17cd6ca5eeb35b435539eeb9fd763f9eee600032) or should I revert to `S`.
It's slightly more strict.
I am not sure myself if we want to restrict the type for plugins or not.
i.e. maybe an array could be acceptable

.changeset/poor-boats-reflect.md
Mouvedia: Yes but RuleBase is internal :)

```ts
	/** @internal */
	export type RuleBase<P = any, S = any> = (
```

It flows back to `Rule`, you are right.
I thought I would be spared on a technicality.

types/stylelint/index.d.ts
```diff
@@ -246,7 +246,7 @@ declare namespace stylelint {
 	/** @internal */
 	export type RuleBase<P = any, S = any> = (
 		primaryOption: P,
-		secondaryOptions: Record<string, S>,
+		secondaryOptions: S extends Record<string, any> ? S : never,
```
Mouvedia, line 249:
```suggestion
		secondaryOptions: S,
```

see https://github.com/stylelint/stylelint/pull/7950#discussion_r1730886345

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #7954, opened 2024-08-23
Refactor to add `CoreRules` type

Mouvedia:

slightly related to #7608

- short term
	- facilitate detecting inconsistencies between what's documented and what can actually be passed to the options
	- auto completion
	- enable further refactors
	e.g. `possible` type
- long term
  - config tab of https://stylelint.io/demo
  - auto-generated documentation
  - runtime validation

~~blocked by #7950~~

### DIFF
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -92,7 +92,7 @@ declare namespace stylelint {
 
 	/** @internal */
 	export type DisableOptions = {
-		except?: (string | RegExp)[];
+		except?: StringsOrRegexes;
 		severity?: Severity;
 	};
 
@@ -282,18 +282,22 @@ declare namespace stylelint {
 		readonly [name in keyof CoreRules]: Promise<Rule>;
 	};
 
-	type Primary = string | number | true | RegExp | Array<string | RegExp> | Record<string, any>;
+	type StringOrRegex = string | RegExp;
+	type StringsOrRegexes = Array<StringOrRegex>;
+	type OneOrMoreStrings = string | string[];
+	type Primary = number | true | StringOrRegex | StringsOrRegexes | Record<string, any>;
 	type Secondary = Record<string, any>;
 	type CoreRule<P extends Primary, S extends Secondary = any> = Rule<P, S>;
 
+	/** @internal */
 	export type CoreRules = {
 		'alpha-value-notation': CoreRule<
 			'number' | 'percentage',
-			{ exceptProperties: Array<string | RegExp> }
+			{ exceptProperties: StringsOrRegexes }
 		>;
-		'annotation-no-unknown': CoreRule<true, { ignoreAnnotations: Array<string | RegExp> }>;
-		'at-rule-allowed-list': CoreRule<string | string[]>;
-		'at-rule-disallowed-list': CoreRule<string | string[]>;
+		'annotation-no-unknown': CoreRule<true, { ignoreAnnotations: StringsOrRegexes }>;
+		'at-rule-allowed-list': CoreRule<OneOrMoreStrings>;
+		'at-rule-disallowed-list': CoreRule<OneOrMoreStrings>;
 		'at-rule-empty-line-before': CoreRule<
 			'always' | 'never',
 			{
@@ -314,35 +318,32 @@ declare namespace stylelint {
 				ignoreAtRules: string[];
 			}
 		>;
-		'at-rule-no-unknown': CoreRule<true, { ignoreAtRules: Array<string | RegExp> }>;
+		'at-rule-no-unknown': CoreRule<true, { ignoreAtRules: StringsOrRegexes }>;
 		'at-rule-no-vendor-prefix': CoreRule<true>;
-		'at-rule-property-required-list': CoreRule<Record<string, string | string[]>>;
+		'at-rule-property-required-list': CoreRule<Record<string, OneOrMoreStrings>>;
 		'block-no-empty': CoreRule<true, { ignore: 'comments'[] }>;
-		'color-function-notation': CoreRule<
-			'modern' | 'legacy',
-			{ ignore: 'with-var-inside'[] | 'with-var-inside' }
-		>;
+		'color-function-notation': CoreRule<'modern' | 'legacy', { ignore: 'with-var-inside'[] }>;
 		'color-hex-alpha': CoreRule<'always' | 'never'>;
 		'color-hex-length': CoreRule<'short' | 'long'>;
 		'color-named': CoreRule<
 			'never' | 'always-where-possible',
-			{ ignoreProperties: Array<string | RegExp>; ignore: ['inside-function'] }
+			{ ignoreProperties: StringsOrRegexes; ignore: 'inside-function'[] }
 		>;
 		'color-no-hex': CoreRule<true>;
 		'color-no-invalid-hex': CoreRule<true>;
 		'comment-empty-line-before': CoreRule<
 			'always' | 'never',
 			{
-				except: ['first-nested'];
+				except: 'first-nested'[];
 				ignore: ('stylelint-commands' | 'after-comment')[];
-				ignoreComments: Array<string | RegExp>;
+				ignoreComments: StringsOrRegexes;
 			}
 		>;
 		'comment-no-empty': CoreRule<true>;
-		'comment-pattern': CoreRule<string | RegExp>;
+		'comment-pattern': CoreRule<StringOrRegex>;
 		'comment-whitespace-inside': CoreRule<'always' | 'never'>;
-		'comment-word-disallowed-list': CoreRule<string | RegExp | Array<string | RegExp>>;
-		'custom-media-pattern': CoreRule<string | RegExp>;
+		'comment-word-disallowed-list': CoreRule<StringOrRegex | StringsOrRegexes>;
+		'custom-media-pattern': CoreRule<StringOrRegex>;
 		'custom-property-empty-line-before': CoreRule<
 			'always' | 'never',
 			{
@@ -351,10 +352,10 @@ declare namespace stylelint {
 			}
 		>;
 		'custom-property-no-missing-var-function': CoreRule<true>;
-		'custom-property-pattern': CoreRule<string | RegExp>;
+		'custom-property-pattern': CoreRule<StringOrRegex>;
 		'declaration-block-no-duplicate-custom-properties': CoreRule<
 			true,
-			{ ignoreProperties: Array<string | RegExp> }
+			{ ignoreProperties: StringsOrRegexes }
 		>;
 		'declaration-block-no-duplicate-properties': CoreRule<
 			true,
@@ -365,12 +366,12 @@ declare namespace stylelint {
 					| 'consecutive-duplicates-with-different-syntaxes'
 					| 'consecutive-duplicates-with-same-prefixless-values'
 				>;
-				ignoreProperties: Array<string | RegExp>;
+				ignoreProperties: StringsOrRegexes;
 			}
 		>;
 		'declaration-block-no-redundant-longhand-properties': CoreRule<
 			true,
-			{ ignoreShorthands: Array<string | RegExp>; ignoreLonghands: string[] }
+			{ ignoreShorthands: StringsOrRegexes; ignoreLonghands: string[] }
 		>;
 		'declaration-block-no-shorthand-property-overrides': CoreRule<true>;
 		'declaration-block-single-line-max-declarations': CoreRule<number>;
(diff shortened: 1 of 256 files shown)


Mouvedia:
@ybiquitous it reflects our current documentation.
Whether the implementation or the documentation should be considered the source of truth can be discussed in a subsequent issue.
i.e. inconsistencies between the two will have to be settled one way or another

I think this change should please Romain who likes automising processes.
e.g. set the type and the documentation regenerates in a pre-release hook

types/stylelint/index.d.ts
```diff
@@ -279,133 +279,340 @@ declare namespace stylelint {
 	};
 
 	type BuiltInRules = {
```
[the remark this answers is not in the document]
Mouvedia: I think we might decide to change `: Promise<Rule>;` soon.
Let's rename the type if we ever do.
i.e. rename if it actually changes

types/stylelint/index.d.ts
```diff
@@ -279,133 +279,340 @@ declare namespace stylelint {
...
+	};
+
+	type Primary = string | number | true | RegExp | Array<string | RegExp> | Record<string, any>;
+	type Secondary = Record<string, any>;
+	type CoreRule<P extends Primary, S extends Secondary = any> = Rule<P, S>;
+
+	export type CoreRules = {
+		'alpha-value-notation': CoreRule<
+			'number' | 'percentage',
+			{ exceptProperties: Array<string | RegExp> }
```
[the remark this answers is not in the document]
Mouvedia: I thought of it but I thought we would want to find the inconsistencies before doing micro optimisations.
I guess it can be done before just as well.

types/stylelint/index.d.ts
```diff
@@ -279,133 +279,340 @@ declare namespace stylelint {
...
+					| 'blockless-after-same-name-blockless'
+					| 'blockless-after-blockless'
+				>;
+				ignoreAtRules: string[];
+			}
+		>;
+		'at-rule-no-unknown': CoreRule<true, { ignoreAtRules: Array<string | RegExp> }>;
+		'at-rule-no-vendor-prefix': CoreRule<true>;
+		'at-rule-property-required-list': CoreRule<Record<string, string | string[]>>;
+		'block-no-empty': CoreRule<true, { ignore: 'comments'[] }>;
```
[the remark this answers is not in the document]
Mouvedia: I think this notation is clearer when you have unions.
I have been consistent in this.

types/stylelint/index.d.ts
```diff
@@ -279,133 +279,340 @@ declare namespace stylelint {
...
+		'color-function-notation': CoreRule<
+			'modern' | 'legacy',
+			{ ignore: 'with-var-inside'[] | 'with-var-inside' }
```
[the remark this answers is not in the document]
Mouvedia: Actually it supports both IIRC—e.g. like all `Array<string>`—but you are right the documentation doesn't state it.

types/stylelint/index.d.ts
```diff
@@ -279,133 +279,340 @@ declare namespace stylelint {
...
+		>;
+		'color-hex-alpha': CoreRule<'always' | 'never'>;
+		'color-hex-length': CoreRule<'short' | 'long'>;
+		'color-named': CoreRule<
+			'never' | 'always-where-possible',
+			{ ignoreProperties: Array<string | RegExp>; ignore: ['inside-function'] }
```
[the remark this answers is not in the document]
Mouvedia:
```suggestion
			{ ignoreProperties: Array<string | RegExp>; ignore: 'inside-function'[] }
```

types/stylelint/index.d.ts
```diff
@@ -279,133 +279,340 @@ declare namespace stylelint {
...
+		>;
+		'declaration-block-no-shorthand-property-overrides': CoreRule<true>;
+		'declaration-block-single-line-max-declarations': CoreRule<number>;
+		'declaration-empty-line-before': CoreRule<
+			'always' | 'never',
+			{
+				except: ('first-nested' | 'after-comment' | 'after-declaration')[];
+				ignore: [
+					'after-comment' | 'after-declaration' | 'first-nested' | 'inside-single-line-block',
+				];
```
Mouvedia, line 383:
```suggestion
				ignore:
					('after-comment' | 'after-declaration' | 'first-nested' | 'inside-single-line-block')[]
```

types/stylelint/index.d.ts
```diff
@@ -279,133 +279,340 @@ declare namespace stylelint {
...
-		readonly 'unit-no-unknown': Promise<Rule>;
-		readonly 'value-keyword-case': Promise<Rule>;
-		readonly 'value-no-vendor-prefix': Promise<Rule>;
+		readonly [name in keyof CoreRules]: Promise<Rule>;
+	};
+
+	type Primary = string | number | true | RegExp | Array<string | RegExp> | Record<string, any>;
+	type Secondary = Record<string, any>;
+	type CoreRule<P extends Primary, S extends Secondary = any> = Rule<P, S>;
+
```
Mouvedia, line 288:
```suggestion

type StringOrRegex = string | RegExp;
type StringsOrRegexes = Array<StringOrRegex>;
type OneOrMoreStrings = string | string[];
```

Mouvedia:
> Thanks, great work!

Thanks.

> Just to be sure, this is just a refactoring, right?

It should be.
This change only produced one type error which has been fixed in 94e53d76d5dd8b78b75a739ebf8b2547316b9986.

types/stylelint/index.d.ts
romainmenke, line 1: Is it correct that we would have to keep this file in sync with each rule manually?
And that the types aren't actually tested or validated?

Can't we keep the type definition for each rule closer to the actual rule?

Mouvedia: This PR is a pre-requisite for future refactors and features.
see [OP](https://github.com/stylelint/stylelint/pull/7954#issue-2482663449)

In short, yes, in the short term the gains are not obvious but it will become evident if we ever decide to move towards these goals.

You can continue to type your rules as you did before.
i.e. `Rule` is still available

romainmenke: I am sorry @Mouvedia but can we please keep these threads in the right tone?
I, myself might also fail to do this from time to time and please point this out when I do. As contributors we should be careful to set the right example and to make sure that other contributors that read through these conversations feel that this is a safe and welcoming space.

My question was in no way confrontational.

I am worried that often when I ask questions about a proposed change that these questions are shut down.

I think it is important that everyone feels that they can ask questions, give comments and that those will be fully and fairly considered.

I strongly believe that hearing multiple opinions and working towards consensus is important.

------

Is there room to consider alternatives?
Maybe we can have all of the benefits without any of the drawbacks?

Mouvedia: Hum, I didn't think I shut you down.
Ill be mindful of that from now on.

> Is there room to consider alternatives?

This comment is not resolved; of course there is but we would have to take into account that, again, the gains aren't obvious.
i.e. this PR as itself isn't that great
IMHO typing the rules that way is long overdue.

> Maybe we can have all of the benefits without any of the drawbacks?

I don't know, maybe.
I am working towards bigger goals in small steps.
If you think it's not going in the right direction and the goals I laid out are not pursuing, I would understand if you were against merging this PR.

romainmenke: Thank you 🙇 

----

I am all for auto-generated docs.

I however can not see the end result you have envisioned and there isn't any design thread to detail/explain where we will end up :)

So I have to judge this PR purely on the work that is being done here.
Separating the rules from their type definitions is something I think we should try to avoid.

Mouvedia: > I however can not see the end result you have envisioned and there isn't any design thread to detail/explain where we will end up :)
> […]
> Separating the rules from their type definitions is something I think we should try to avoid.

Try to imagine having to generate the documentation from the types.
Now imagine having the types scattered between all the rules themselves.
Wouldn't having to access all these files separately to extract the types from the JSDoc comments be less efficient than having it centralised in the d.ts.
I am not saying it's not doable nor that having the types inlined in the rules isn't more handy.

I am also not saying that I am assuming that it should be expected of you to share a vision that I didn't lay out in details. Nonetheless, I still think you should see the immediate benefit of having **all** the rules primary and secondary options be—as faithful to the current documentation as we could—typed.

Mouvedia:
Another example of future usage would be to validate configs on the demo site.
e.g. [demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYLIBsICd45igG4UBfAHQDsQAaEAMwEtMYA5JAWzkRgA9OADiwB0YAM5ja4CJSYBzBCGBUUKciBwBXFmPXw0K1WunYcAWnqbKYAC6MZZyhBtI7MvSgDa6gEZIc6jQGlEZG6oxyTgQe3gwQEOoAuoaqpMkhZFSkUpCyEQBiuByuigBWYjJSsAKSiMoZ6mI2AJ4smIyUNh7qmK4wTYGGjS1tHTZmufJmTUiUUP5Q3SC9Nv1dIFkgpEA) doesn't produce an error

types/stylelint/index.d.ts
romainmenke: > Now imagine having the types scattered between all the rules themselves.
Wouldn't having to access all these files separately to extract the types from the JSDoc comments (or possible values) be less efficient than having it centralised in the d.ts?

Yes, this is less efficient, but it can be done by a script?

While having the type declaration close to the rule helps contributors to more easily create, modify and understand rules.

I think these things are important:
- the types should be verified and tested for correctness (at some point in the future)
- contributors must easily find how to update type definitions
- if there is a test that gives a type error, it must point to all relevant files

Separating the types from the rules might be fine if we have sufficient tooling to help contributors. But if we end up needing to add extra tooling then maybe better to have tooling to extract the current JSDoc comments?

I don't think there is an obvious best or better approach here.

--------

Which tool or technique do you have in mind to generate the docs?

I am currently a big fan of https://www.npmjs.com/package/@microsoft/api-extractor
It makes it easy to generate a single `.d.ts` file for only a public interface.

Mouvedia: I think the PROs outweighs the CONs or else I wouldn't have sent this PR.
The long-term goal list summarise is succinctly:

> - config tab of [stylelint.io/demo](https://stylelint.io/demo)
> - auto-generated documentation
> - runtime validation

> Which tool or technique do you have in mind to generate the docs?

I didn't dive into that yet but I already had https://api-extractor.com/ as a bookmark.
Ill ping you if I start anything related to that.
Here's the order of the changes that I would follow through if I had the time to do everything myself:

1. restrict `SecondaryOptions` after discussion
2. restrict `possible`
3. generate JSON Type Definition or JSON schema on the fly
4. exploit that JSON file for features like
  - auto-generated documentation
  - demo
    - validate options
    - auto-suggest options
  - etc.
 5. discuss [`defineConfig`](https://github.com/eslint-types/eslint-define-config#usage)
 6. refactor the current runtime validation of rules if deemed warranted

romainmenke: Sounds good, thank you for elaborating a bit here 🙇

ybiquitous:
> e.g. [demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYLIBsICd45igG4UBfAHQDsQAaEAMwEtMYA5JAWzkRgA9OADiwB0YAM5ja4CJSYBzBCGBUUKciBwBXFmPXw0K1WunYcAWnqbKYAC6MZZyhBtI7MvSgDa6gEZIc6jQGlEZG6oxyTgQe3gwQEOoAuoaqpMkhZFSkUpCyEQBiuByuigBWYjJSsAKSiMoZ6mI2AJ4smIyUNh7qmK4wTYGGjS1tHTZmufJmTUiUUP5Q3SC9Nv1dIFkgpEA) doesn't produce an error

This looks like a bug. Can you [open an issue](https://github.com/stylelint/stylelint-demo/issues/new/choose) as a bug report?

types/stylelint/index.d.ts
```diff
@@ -279,133 +279,329 @@ declare namespace stylelint {
...
-		readonly 'value-no-vendor-prefix': Promise<Rule>;
+		readonly [name in keyof CoreRules]: Promise<Rule>;
+	};
+
+	type StringOrRegex = string | RegExp;
+	type StringsOrRegexes = Array<StringOrRegex>;
+	type OneOrMoreStrings = string | string[];
+	type Primary = number | true | StringOrRegex | StringsOrRegexes | Record<string, any>;
+	type Secondary = Record<string, any>;
+	type CoreRule<P extends Primary, S extends Secondary = any> = Rule<P, S>;
```
ybiquitous, line 290: [suggest] Minor. How about renaming the types more specifically?

```suggestion
	type PrimaryOption = number | true | StringOrRegex | StringsOrRegexes | Record<string, any>;
	type SecondaryOptions = Record<string, any>;
	type CoreRule<P extends PrimaryOption, S extends SecondaryOptions = any> = Rule<P, S>;
```

ybiquitous (approved):
Thanks, LGTM 👍🏼 

I've left a minor suggestion, but it's not a blocker.

Mouvedia:
@ybiquitous I am surprised that you managed to review this that fast; thanks.

The follow-up PR will be the about the inconsistencies.
i.e. https://github.com/stylelint/stylelint/pull/7954#discussion_r1731611549
You don't have to wait for it, to release the next version though.

types/stylelint/index.d.ts
Mouvedia: > extract the types from the JSDoc comments (or possible values)
> […]
> but it can be done by a script?

As promised when I said

> Ill ping you if I start anything related to that.

I have a working POC that generates a JSON schema from rule/rules that use the `validateOptions` util.
I just did it for fun; it doesn't have tests.
Its purpose would be to be used as a fallback when a plugin doesn't provide types in the form of .d.ts or it's not part of the plugins that are maintained by [`stylelint-types`](https://github.com/stylelint-types).
i.e. 1. try d.ts 2. try `@stylelint-types/*` 3. fallback to my script

Issue #7241, opened 2023-10-16
Fix `declaration-property-value-no-unknown` false positives for `oklch` and `oklab`


ghost:

Mouvedia:
We are waiting for the next release of csstree.
It's fixed on HEAD.

ref: csstree/csstree#244

jeddy3:
@meerhimmel While waiting on the next release of csstree (the upstream package that powers this rule), you can amend the syntax in your own config:

```json
{
  "rules": {
    "declaration-property-value-no-unknown": [true, {
      "typesSyntax": { 
        "color": "| oklch( [ <percentage> | <number> | none] [ <percentage> | <number> | none] [ <hue> | none] [ / [<alpha-value> | none] ]? )"
      }
    }]
  }
}
```

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYLIBsICd4ogGtMwALACgBYAGAOgEYBSFO+gJgGYU37aBWAOwBKANwoAviAA0IAGYBLTDABySALZxEMAB7qADktpgAzsengIAOwUBzBCGAAdSyhSOQOAK5Lj7-E5dXNxBYMEwkHCQAF3krAFo9HAg9GBwogE84gDckTE8YOMsIOM9LQiKAd0s-FABtKK8YKTRnIKD3DJTjAGV0yyikbRr0Vrb2i2wcGvcAHwJiMnI6lAAeFJwwGH6kGxgAPhQ5lctPNQAjVIO5ossYAF1ltdTN7d2r1ZPzy8OUG-vH0j5d5-B61FAAejqK1yelISGyuSBPxBKDuAH4UEJ3KM2uIcRI7qM8ZZJDJINZ5DYAGK4NTRewAK2MVnMsD0ZkQAVc7mMGSUmHk-WmIHCURgvPcUlGPL5MAF-Ti5NscV5SEsUAiUGFovFUWxJJA4iAA).

Mouvedia:
I've labeled the issue as ready to implement. Please consider [contributing](https://stylelint.io/contributing) if you have time.

There are [steps on how to fix a bug in a rule](https://stylelint.io/developer-guide/rules#fix-a-bug-in-a-rule) in the Developer guide.

Mouvedia:
> Unexpected unknown value \"c\" for property \"background-color\" (declaration-property-value-no-unknown)

The challenge for this one is to add support for the channel keywords of `lab` and `lch`.
[...]

Issue #7917, opened 2024-08-11
Add `transition-behavior` support to `declaration-block-no-redundant-longhand-properties` and `declaration-block-no-shorthand-property-overrides`


Mouvedia:
### What minimal example or steps are needed to reproduce the bug?

```css
a {
  transition-behavior: foo;
  transition-delay: foo;
  transition-duration: foo;
  transition-property: foo
  transition-timing-function: foo;
}
```

```css
a {
  transition-behavior: foo;
  transition: foo;
}
```

### What minimal configuration is needed to reproduce the bug?

.

### How did you run Stylelint?

.

### Which Stylelint-related dependencies are you using?

.

### What did you expect to happen?

.

### What actually happened?

https://github.com/stylelint/stylelint/blob/ee3118d2460cddf469f959b5438d8aeab7e9584e/lib/reference/properties.mjs#L601-L610

### Do you have a proposal to fix the bug?

The low browser support—[under 90%](https://caniuse.com/mdn-css_properties_transition-behavior)—is not a problem for `declaration-block-no-redundant-longhand-properties` because we provide the `ignoreLonghands` option.
Nonetheless Id prefer if the issue was labeled _status: agreed but held_ for now.

ybiquitous:
@Mouvedia Thanks for the report. Supporting `transition-timing-function` sounds good to me. Just to make sure, we can confirm this bug on [this demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYAdAdugLjgTkpgM4CWOpEmAtAA74S0z44Ce8KSANlwNxa4CRMhSrUoAV0KjMHAAwA6AEwBWYv2x5CJcpRoUAtqUwBzagDMJmMDI4wkxGBsHaRe8TC5J2KOeoFawrpiAEYwABZIAG6U+BzcXBAA7uKkxGD4MDhOWAC+IAA0IOakXDAAckgGcIgwAB5VtGUKYMTEheBUJSYIIBiYKCjoIPXZmFDEwxwA2liDg8PEbGVcxjjUkJjd1EtEUEj4UMNzKAC6eR2b3QBiEPgGSDi9AFbEVB2wtO2I-QsgS6wVmspkMQF5skthgUTotlp41hsuqQzLtxgcjghQeCYJCQBdckA):

```css
a {
	transition-property: all;
	transition-duration: 0.25s;
	transition-timing-function: ease;
	transition-delay: 0s;
	transition-behavior: allow-discrete;
}

[...]

PR #7957, opened 2024-08-24
Fix `declaration-property-value-no-unknown` false negatives

Mouvedia:

follow-up of #7944

[demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYAdAdilTNgAsIAnAWkyQFsZ4BuLHANwEsYB3UgF2eoBtnMMUkgAezAM7wUAMwgR62FC3ZceMfoNIDxMTlNnyGS1h258BQitTpHlptRpoy5Cpia7E845twiZyVE4GCgC+WCAANCDSzLwwAHKBCCAwIlQADnEAdGDi4pHgfjEA5skYiuggxACuceKVUuU4OJWwYLxInr7+6cQQ6TDEnACepIxIvNWWEKTVmADWmBBsmA0onDUwRmGYIQWQmCUAYiSUSJzJAFbifgWw6fmITSiV4iNxGpxrlR2cMG+VCJGV7vdQCTikA4lUhvPBQTpQb4gX7-L4gLB7EJAA)

https://caniuse.com/mdn-css_properties_view-transition-name
https://caniuse.com/?search=view-timeline
https://caniuse.com/mdn-css_properties_anchor-name
https://csstree.github.io/docs/validator.html

The changeset of #7944 will be modified in #7946 to include #7957.
see https://github.com/stylelint/stylelint/pull/7946/files#r1730474501

### DIFF
--- a/lib/rules/declaration-property-value-no-unknown/index.cjs
+++ b/lib/rules/declaration-property-value-no-unknown/index.cjs
@@ -67,18 +67,32 @@ const rule = (primary, secondaryOptions) => {
 		const propertiesSyntax = {
 			overflow: '| overlay', // csstree/csstree#248
 			width: '| min-intrinsic | -moz-min-content | -moz-available | -webkit-fill-available', // csstree/csstree#242
+			'anchor-name': 'none | <custom-property-name>#',
 			'field-sizing': 'content | fixed',
 			'text-box-edge':
 				'auto | [ text | cap | ex | ideographic | ideographic-ink ] [ text | alphabetic | ideographic | ideographic-ink ]?',
 			'text-box-trim': 'none | trim-start | trim-end | trim-both',
 			'text-spacing-trim': 'normal | space-all | space-first | trim-start',
-			'word-break': '| auto-phrase',
 			'text-wrap-mode': 'wrap | nowrap',
 			'text-wrap-style': 'auto | balance | pretty | stable',
 			'text-wrap': "<'text-wrap-mode'> || <'text-wrap-style'>",
+			'view-timeline-axis': '[ block | inline | x | y ]#',
+			'view-timeline-inset': '[ [ auto | <length-percentage> ]{1,2} ]#',
+			'view-timeline-name': '[ none | <custom-property-name> ]#',
+			'view-timeline':
+				"[ <'view-timeline-name'> [ <'view-timeline-axis'> || <'view-timeline-inset'> ]? ]#",
+			// <custom-ident> represents any valid CSS identifier that would not be misinterpreted as a pre-defined keyword in that property’s value definition
+			// i.e. reserved keywords don't have to be excluded explicitly
+			// w3c/csswg-drafts#9895
+			'view-transition-name': 'none | <custom-ident>',
+			'word-break': '| auto-phrase',
 			...secondaryOptions?.propertiesSyntax,
 		};
-		// TODO csstree/csstree#244
+
+		/**
+		 * @todo add support for oklab(), oklch(), color(), color-mix(), light-dark(), etc.
+		 * @see https://drafts.csswg.org/css-color-5/
+		 */
 		const typesSyntax = { ...secondaryOptions?.typesSyntax };
 
 		/** @type {Map<string, string>} */
--- a/lib/rules/declaration-property-value-no-unknown/index.mjs
+++ b/lib/rules/declaration-property-value-no-unknown/index.mjs
@@ -64,18 +64,32 @@ const rule = (primary, secondaryOptions) => {
 		const propertiesSyntax = {
 			overflow: '| overlay', // csstree/csstree#248
 			width: '| min-intrinsic | -moz-min-content | -moz-available | -webkit-fill-available', // csstree/csstree#242
+			'anchor-name': 'none | <custom-property-name>#',
 			'field-sizing': 'content | fixed',
 			'text-box-edge':
 				'auto | [ text | cap | ex | ideographic | ideographic-ink ] [ text | alphabetic | ideographic | ideographic-ink ]?',
 			'text-box-trim': 'none | trim-start | trim-end | trim-both',
 			'text-spacing-trim': 'normal | space-all | space-first | trim-start',
-			'word-break': '| auto-phrase',
 			'text-wrap-mode': 'wrap | nowrap',
 			'text-wrap-style': 'auto | balance | pretty | stable',
 			'text-wrap': "<'text-wrap-mode'> || <'text-wrap-style'>",
+			'view-timeline-axis': '[ block | inline | x | y ]#',
+			'view-timeline-inset': '[ [ auto | <length-percentage> ]{1,2} ]#',
+			'view-timeline-name': '[ none | <custom-property-name> ]#',
+			'view-timeline':
+				"[ <'view-timeline-name'> [ <'view-timeline-axis'> || <'view-timeline-inset'> ]? ]#",
+			// <custom-ident> represents any valid CSS identifier that would not be misinterpreted as a pre-defined keyword in that property’s value definition
+			// i.e. reserved keywords don't have to be excluded explicitly
+			// w3c/csswg-drafts#9895
+			'view-transition-name': 'none | <custom-ident>',
+			'word-break': '| auto-phrase',
 			...secondaryOptions?.propertiesSyntax,
 		};
-		// TODO csstree/csstree#244
+
+		/**
+		 * @todo add support for oklab(), oklch(), color(), color-mix(), light-dark(), etc.
+		 * @see https://drafts.csswg.org/css-color-5/
+		 */
 		const typesSyntax = { ...secondaryOptions?.typesSyntax };
 
 		/** @type {Map<string, string>} */
--- a/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
+++ b/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
@@ -75,6 +75,18 @@ testRule({
 			code: 'a { width: min-intrinsic; }',
 		},
 		{
+			code: 'a { view-timeline-name: --foo; }',
+		},
+		{
+			code: 'a { view-timeline: --bar x; }',
+		},
+		{
+			code: 'a { view-transition-name: qux-baz; }',
+		},
+		{
+			code: 'a { anchor-name: --bar, --qux; }',
+		},
+		{
 			code: stripIndent`
 				a { font-weight: bolder; }
 				@font-face { font-weight: 100 200; }


lib/rules/declaration-property-value-no-unknown/index.mjs
```diff
@@ -64,16 +64,30 @@ const rule = (primary, secondaryOptions) => {
...
 			'text-wrap-style': 'auto | balance | pretty | stable',
+			'view-timeline-axis': '[ block | inline | x | y ]#',
+			'view-timeline-inset': '[ [ auto | <length-percentage> ]{1,2} ]#',
+			'view-timeline-name': '[ none | <custom-property-name> ]#',
+			'view-timeline':
+				"[ <'view-timeline-name'> [ <'view-timeline-axis'> || <'view-timeline-inset'> ]? ]#",
+			// <custom-ident> represents any valid CSS identifier that would not be misinterpreted as a pre-defined keyword in that property’s value definition
+			// i.e. reserved keywords don't have to excluded
+			// w3c/csswg-drafts#9895
+			'view-transition-name': 'none | <custom-ident>',
```
Mouvedia, line 82: I couldn't find any way to exclude a keyword.
i.e. something that could merge `<custom-ident>` with `auto#{0}`

lib/rules/declaration-property-value-no-unknown/index.mjs
```diff
@@ -64,16 +64,30 @@ const rule = (primary, secondaryOptions) => {
 		const propertiesSyntax = {
 			overflow: '| overlay', // csstree/csstree#248
 			width: '| min-intrinsic | -moz-min-content | -moz-available | -webkit-fill-available', // csstree/csstree#242
+			'anchor-name': 'none | <custom-property-name>#',
```
Mouvedia, line 67: `<dashed-ident>` is not exposed.

see https://github.com/csstree/csstree/blob/ba6dfd8bb0e33055c05f13803d04825d98dd2d8d/lib/lexer/generic.js#L225-L227

lib/rules/declaration-property-value-no-unknown/index.cjs
```diff
@@ -67,16 +67,30 @@ const rule = (primary, secondaryOptions) => {
...
 			'text-box-trim': 'none | trim-start | trim-end | trim-both',
 			'text-spacing-trim': 'normal | space-all | space-first | trim-start',
 			'text-wrap-style': 'auto | balance | pretty | stable',
+			'view-timeline-axis': '[ block | inline | x | y ]#',
+			'view-timeline-inset': '[ [ auto | <length-percentage> ]{1,2} ]#',
+			'view-timeline-name': '[ none | <custom-property-name> ]#',
+			'view-timeline':
+				"[ <'view-timeline-name'> [ <'view-timeline-axis'> || <'view-timeline-inset'> ]? ]#",
+			// <custom-ident> represents any valid CSS identifier that would not be misinterpreted as a pre-defined keyword in that property’s value definition
+			// i.e. reserved keywords don't have to excluded
```
Mouvedia, line 83:
```suggestion
			// i.e. reserved keywords don't have to excluded explicitly
```

romainmenke (commented):
I'd rather we fix false negatives upstream in `css-tree` and try to limit local patches to false positives.

I am concerned that we will end up maintaining to much data/config here in Stylelint.

----

Edit: I've asked in `css-tree` about the maintenance status of `csstree` : https://github.com/csstree/csstree/issues/279

Mouvedia:
> I'd rather we fix false negatives upstream in css-tree and try to limit local patches to false positives.

Me too but for now this will do.
I value the users more than _temporary_ annoyances for maintainers.
i.e. we will eventually fork its repo if the license let's us
I also waited years before finally taking the time to send a PR.
It's not ideal but it's the best we can do for _now_.

Good luck with your issue nonetheless.

romainmenke:
My concern is exactly because this is not a temporary situation.
It would be a permanent increase in our workload.

I'd appreciate it if we can open this up for feedback and other opinions and do not immediately assume that this is something that we need to do.

Keep in mind that even if you are willing to submit all the patches to keep this rule up to date, someone else also needs to volunteer their time to review and release :)

Mouvedia:
We have accepted to add the rule on the premise that css-tree would be maintained or accept our PRs in a timely manner.
It wasn't the case.
This rule had accumulated false positives/negatives.
I went and fixed some of them.

> and do not immediately assume that this is something that we need to do.

If there's a false negative/positive, it's an immediate _ready to implement_ unless a browser doesn't support it.
i.e. it's up to be discussed if it's behind a flag or if it has unwanted consequences
i.e. #7510
e.g. #7917

If you don't to maintain this rule anymore, you can open a new issue to discuss its removal/deprecation.
For now Ill continue fixing false negatives/positives for non-deprecated rules.

ybiquitous:
I appreciate @Mouvedia's work to fix bugs in this rule, but I well understand @romainmenke's concern. Especially, this rule has to support many property value definitions to be *truly useful*. We want to rely on a reliable library like csstree to work on this rule, but *for now*, we have to accept ad-hoc fixes like this PR.

Of course, I believe this rule should be useful for many CSS authors because of its error detection.

If too many bug fixes would come for this rule or we would feel overloaded, let's reconsider making it deprecated or stopping the maintenance until csstree becomes active or an alternative appears.

Generally speaking, this is an example of challenging things in OSS management. 😞

Mouvedia:
> If too many bug fixes would come for this rule or we would feel overloaded, let's reconsider making it deprecated or stopping the maintenance until csstree becomes active or an alternative appears.

If it comes to that, we will add a new `stylelintWarningType`.
e.g. `frozen`

If we get no answers from css-tree maintainers, I would prefer if we'd go the fork route.
Well let's not get ahead of ourselves; the options are almost enough for now.
I am afraid we won't be able to fix #7241 in a spec-compliant manner through them though.

ybiquitous (commented):
The changes look good to me. 👍🏼 

Can you add a changelog item? If possible, I want to clarify the target properties by this bug fix, e.g.

```
Fix `declaration-property-value-no-unknown` false negatives for <property names...>
```

Mouvedia:
> The changes look good to me. 👍🏼
> 
> Can you add a changelog item? If possible, I want to clarify the target properties by this bug fix, e.g.
> 
> ```
> Fix `declaration-property-value-no-unknown` false negatives for <property names...>
> ```

> The changeset of https://github.com/stylelint/stylelint/pull/7944 will be modified in https://github.com/stylelint/stylelint/pull/7946 to include https://github.com/stylelint/stylelint/pull/7957.
> see [#7946 (files)](https://github.com/stylelint/stylelint/pull/7946/files#r1730474501)

ybiquitous (approved):
Okay, let's manually add this PR in #7946. 👍🏼

romainmenke:
I think we are overlooking a third option :)

`csstree` is extensible, exactly this is what allows us to patch it here in the first place.

That I think that Stylelint is not the right context doesn't mean that I don't want to have a more modern syntax definition.

I've already referred to `@webref/css` multiple times and I think it also offers a solution here.

------

My main concern with these patches:
- they are adhoc
- they are manual work
- they are not fully and exhaustively tested (we don't even call and check `lexer.validate()`)
- they are not shareable

----

I think a better approach is to have a shared library with syntax patches.

A library that uses external references like `@webref/css` to automated most of the work.
A library that is separate from Stylelint so that it can be used by others facing the same issue.

Something roughly like this: https://github.com/csstools/postcss-plugins/pull/1466

I am not looking forward to maintaining these syntax definitions by hand, here in Stylelint.
But I don't mind maintaining a generated and fully test definition in a context that can support such an effort :)

I'll try to allocate some time to this over the next few weeks and open an issue if I think it is ready enough for feedback.

ybiquitous:
I'm so curious about the work of [`@webref/css`](https://www.npmjs.com/package/@webref/css), thanks for the info, @romainmenke. 😄 

Let me ask a question, isn't there a smarter way to replace csstree by using `@webref/css` for now? I'm concerned about whether we will have to care about csstree.

Mouvedia:
@romainmenke I am curious to know how you handled [this problem](https://github.com/stylelint/stylelint/issues/7241#issuecomment-2302913135).
i.e. I was under the impression that it would require to update css-tree itself

romainmenke:
What `csstree` got right was being able to parse the value definitions (e.g. `'alpha-value': '<number> | <percentage>'`)

This works really well and CSS value syntax evolves slowly enough that this will continue to work. We can still consider forking `csstree` if there are sufficient future issues with this aspect.

At this time I don't see an urgent need to make updates to `csstree` internals.

The more immediate problem is that `csstree` isn't publishing regular updates with the latest syntax definitions.

I propose not doing these syntax definition updates, here, by hand. But instead do them automated and shareable, separate from Stylelint.

These patches are just a mapping of properties and types types and their respective syntax.
`@webref/css` already contains all this information and it is very actively maintained by the W3C.

This approach is also relatively safe going forwards:
- the current script I am writing detects if `csstree` has an update and will request updated patches for those entries. So when `csstree` is more actively maintained the list of patches will shrink.
- the list of patches is just a json and can be copy/pasted and manually maintained if I stop doing maintenance.

I am just a very big proponent of not doing manual and error prone work when even remotely possible 😛

------

> I guess they can maybe be added individually.

Indeed, that is the way to go :)

Mouvedia:
Diversities in approaches is a boon for OSS projects.
If I didn't act and patched it, it wouldn't have tickled you enough.
This was worth it after all, let's switch to what you are proposing once it's ready.

ybiquitous:
I understand it well; thanks for the explanation. 😃 

I'm looking forward to the published list of patches to `csstree` based on `@webref/css`. Then, let's replace the current implementation of the new package. 👍🏼

romainmenke:
> If I didn't act and patched it, it wouldn't have tickled you enough.

No, I am already working 10-12 hours a day and rarely take a day off 😓

By patching these here you made it more urgent do something to illustrate a healthier way forwards. I was well aware of how this could be improved and it was on my very long todo list. I could have proactively communicated this, but that also takes time, that is on me. 

Making changes that result in even more work is alarming to me exactly because of the already massive workload. Quickly showing how we can steer this back to less maintenance work is taxing but important to make sure things don't escalate.

The only way I see that OSS maintenance can be done in a sustainable way is by leveraging automations and minimizing manual labor. This type of manual labor just doesn't scales in OSS. We can never keep up with the collective effort of all spec editors and implementers working for Apple, Mozilla, Google, Samsung, Microsoft, ...

Again this doesn't mean that I don't appreciate these improvements in PR's like this 🙇 

But be mindful that there is a cost to all this.
I have to carefully choose where I spend my time to make sure I don't burn myself out.

Mouvedia:
> > If too many bug fixes would come for this rule or we would feel overloaded, let's reconsider making it deprecated or stopping the maintenance until csstree becomes active or an alternative appears.
> 
> If it comes to that, we will add a new `stylelintWarningType`. e.g. `frozen`

related: https://github.com/eslint/rewrite/pull/149/files#diff-4c2458380e50c967c70d7afe819dc7412180f9f817043f27441de3498e4d0479R136
i.e. `frozen` would be boolean property of `RuleMeta`

Issue #7482, opened 2024-01-21
Deprecate `resolveNestedSelectors` option of `selector-class-pattern`


ybiquitous:
### What is the problem you're trying to solve?

The [`resolveNestedSelectors`](https://github.com/stylelint/stylelint/blob/eebb78694ee87e76245414108b0ae2626c78729f/lib/rules/selector-class-pattern/README.md#resolvenestedselectors-true--false-default-false) secondary option of the `selector-class-pattern` rule can resolve Sass-like identifier concatenation (e.g. `&__foo`), but this behavior is against the current rule policy of "standard CSS syntax only":

https://github.com/stylelint/stylelint/blob/eebb78694ee87e76245414108b0ae2626c78729f/docs/developer-guide/rules.md?plain=1#L11-L13

The CSS Nesting spec [forbids](https://www.w3.org/TR/css-nesting-1/#syntax) such a syntax explicitly:

> Unfortunately, this string-based interpretation is ambiguous with the author attempting to add a type selector in the nested rule.

### What solution would you like to see?

I think 3rd-party plugins like [`stylelint-scss`](https://github.com/stylelint-scss/stylelint-scss) should provide this option instead. So, I suggest deprecating the option and finally removing it in the next major version.

For example:
1. Depreacate the option with a warning in the next _minor_ version (e.g. 16.3.0)
2. Remove the option in the next _major_ version (e.g. 17.0.0)

Currently, `selector-class-pattern` supports the standard nesting with `resolveNestedSelectors: false` (default). See the [demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYAdAdugLgOgCMII0BfLXAMhUOLK1JABoQAzASwBsYA5JAWziIYADwEAHbnjABnGc3ARMHAOYIQGTChToQAJwCu3Gbvhos27bpkxuYHBD0BaMJyRyn4pDhww92BBQAbV1WYl0mc30YGQhOADdeGN8oAGVbGHtHE0DWJE4bUgBdCxRyTEYWSGV2FQAxR35vdQArWMwFWHF5RE0rEBkcAE9uTnZMHFMdEABGADY8ACY8AAZdBhBSIA). So, I think most users would not have trouble without this option.

In addition, from the point of view of a maintainer, the following code in the rule would be no longer necessary and be able to be much simpler, including a dependency on [`postcss-resolve-nested-selector`](https://github.com/davidtheclark/postcss-resolve-nested-selector):

[...]

PR #7959, opened 2024-08-25
Fix `selector-class-pattern` first message argument

### LINKED ISSUE #7958: `selector-class-pattern`: ensure the first message argument is trimmed

```css
.88x, .88x>img {width:88px}
```

```json
{"extends": ["stylelint-config-standard"]}
```

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0AOFAPANAAlRgfAJYC2A5tsAO6FQAuAFvGgA7oC+ImIAZoQDYwAcgENicRDHSjmApGADO8zuAgA7XqQQhgAHRCTaMVVHl742ANp75tAJ4C+hVbQC0kdYVIubw48IBOUHoAuhxc7hoAYhD+xMK0WgBW8mrKsMxKiLqq2NjWdg5OtGZ5IHzxMDZ6mDo5pTb2MI7Obmoa3rS+UAFBCKXlhlUgtRxsQA)

stylelint: '^16.8.2',
'stylelint-config-standard': '^36.0.1',

class selector should be shown as ".88x"

### What actually happened?

in second case class selector is shown as " .88x", which has a space at the start, not part of a selector
```
1:1-5 Expected class selector ".88x" to be kebab-case
1:7-12 Expected class selector " .88x" to be kebab-case
```
any amount of whitespace can be added after the comma, tabs, newlines, etc.

ignore whitespace after commas

Mouvedia:

Closes #7958

N/A


lib/rules/selector-class-pattern/index.mjs
```diff
@@ -89,15 +89,15 @@ const rule = (primary, secondaryOptions) => {
 					return;
 				}
 
-				const selector = String(classNode);
```
Mouvedia, line 92: `String(classNode).trim()` or `classNode.spaces.before = ''` would fix it too.

Mouvedia:
@romainmenke since you were heavily involved in #6234, could you take a look at this PR?

lib/rules/selector-class-pattern/index.mjs
romainmenke: I think I prefer `classNode.toString().trim()` so that it aligns with other cases similar to this.

lib/rules/selector-class-pattern/index.mjs
```diff
@@ -89,15 +89,15 @@ const rule = (primary, secondaryOptions) => {
...
+				const selector = `.${value}`;
```
Mouvedia, line 92:
```suggestion
				const selector = String(classNode).trim();
```

lib/rules/selector-class-pattern/index.cjs
```diff
@@ -92,15 +92,15 @@ const rule = (primary, secondaryOptions) => {
 					return;
 				}
 
-				const selector = String(classNode);
+				const selector = `.${value}`;
```
Mouvedia, line 95:
```suggestion
				const selector = String(classNode).trim();
```

.changeset/selfish-numbers-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-class-pattern` selector message argument be trimmed
```
romainmenke, line 5:
```suggestion
Fixed: `selector-class-pattern` start positions
```

This is more in line with the other change log entries.

romainmenke (commented):
I think this change is good minimal patch until we refactor it for https://github.com/stylelint/stylelint/issues/7482

Thank you @Mouvedia

.changeset/selfish-numbers-sparkle.md
Mouvedia: Check the actual change again.
This is not about the reported ranges.
Granted the actual changeset is bad but yours is not accurate.

Mouvedia:
```suggestion
Fixed: `selector-class-pattern` first message argument
```

romainmenke: I don't think it is inaccurate.

The `endIndex` is the sum of the selector length and the start `index`.
This changes affects the selector length and thereby also the `endIndex`

You can also see how the `endIndex` is incorrect in the [demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0AOFAPANAAlRgfAJYC2A5tsAO6FQAuAFvGgA7oC+ImIAZoQDYwAcgENicRDHSjmApGADO8zuAgA7XqQQhgAHRCTaMVVHl742ANp75tAJ4C+hVbQC0kdYVIubw48IBOUHoAuhxc7hoAYhD+xMK0WgBW8mrKsMxKiLqq2NjWdg5OtGZ5IHzxMDZ6mDo5pTb2MI7Obmoa3rS+UAFBCKXlhlUgtRxsQA):

Screenshot 2024-08-25 at 16 43 06

Mouvedia: So there are two fixes in this PR.
You are right.

.changeset/selfish-numbers-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-class-pattern` message's first argument
```
Mouvedia, line 5:
```suggestion
Fixed: `selector-class-pattern` end positions
```

.changeset/selfish-numbers-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-class-pattern` end positions
```
ybiquitous, line 5: [suggest] Minor. The following seems more accurate:

```suggestion
Fixed: `selector-class-pattern` reported ranges
```

https://github.com/stylelint/stylelint/blob/f1fd5748f9c926b1c5360df5f5e874b22a65357f/CHANGELOG.md?plain=1#L8

Mouvedia: `index` didn't change but fine by me.

ybiquitous (approved):
Thanks. LGTM 👍🏼

PR #7966, opened 2024-08-27
Change `PublicApi['rules']` type

Mouvedia:

see https://github.com/stylelint/stylelint/pull/7954#discussion_r1731601791 and https://github.com/stylelint/stylelint/pull/7954#discussion_r1731266595

`index.d.mts` exposed too many types.

### DIFF
--- a/types/stylelint/index.d.mts
+++ b/types/stylelint/index.d.mts
@@ -1,6 +1,5 @@
 export {
 	default,
-	BuiltInRules,
 	Config,
 	ConfigRuleSettings,
 	CosmiconfigResult,
@@ -16,7 +15,6 @@ export {
 	DisabledWarning,
 	Formatter,
 	FormatterType,
-	Formatters,
 	GetLintSourceOptions,
 	GetPostcssOptions,
 	InternalApi,
@@ -39,7 +37,6 @@ export {
 	RuleMeta,
 	RuleOptions,
 	RuleOptionsPossible,
-	RuleSeverity,
 	Severity,
 	ShorthandProperties,
 	StylelintPostcssResult,
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -278,10 +278,6 @@ declare namespace stylelint {
 		meta?: RuleMeta;
 	};
 
-	type BuiltInRules = {
-		readonly [name in keyof CoreRules]: Promise<Rule>;
-	};
-
 	type StringOrRegex = string | RegExp;
 	type StringsOrRegexes = Array<StringOrRegex>;
 	type OneOrMoreStrings = string | string[];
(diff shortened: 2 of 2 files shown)


Mouvedia:
@ybiquitous do we continue exposing `BuiltinRules`?
Since it wasn't annotated as internal, it would be safer.

ybiquitous:
`BuiltInRules` has not been exposed since it doesn't have `export` (that's why it doesn't have `@internal`).

Mouvedia:
> `BuiltInRules` has not been exposed since it doesn't have `export` (that's why it doesn't have `@internal`).

@ybiquitous 
Are you sure?

https://github.com/stylelint/stylelint/blob/5cf18b19c7ef2a81a0357c7ff28eebb17220defd/types/stylelint/index.d.mts#L3

ybiquitous:
Ah, `BuiltInTypes` has been mistakenly added to `index.d.mts` in #7377. It shouldn't be exposed, and it should be unavailable actually?

ybiquitous:
See:

```sh-session
$ cat package.json
{
  "dependencies": {
    "@tsconfig/node20": "^20.1.4",
    "stylelint": "^16.8.2",
    "typescript": "^5.5.4"
  }
}

$ cat tsconfig.json
{
  "extends": "@tsconfig/node20/tsconfig.json"
}

$ cat test.ts
import {type BuiltInTypes, type Rule} from 'stylelint';

$ npx tsc --noEmit
test.ts:1:14 - error TS2305: Module '"stylelint"' has no exported member 'BuiltInTypes'.

1 import {type BuiltInTypes, type Rule} from 'stylelint';
               ~~~~~~~~~~~~


Found 1 error in test.ts:1
```

Mouvedia:
> Ah, `BuiltInTypes` has been mistakenly added to `index.d.mts` in #7377. It shouldn't be exposed, and it should be unavailable actually?

OK so if that's confirmed, I think we should have a separate PR that removes all the exposed types that lack an export.

ybiquitous:
So, I think it's safe to inline `BuiltInRules` and remove it from `index.d.mts` without any changelog items. It's just a refactoring.

Mouvedia:
> So, I think it's safe to inline `BuiltInRules` and remove it from `index.d.mts` without any changelog items. It's just a refactoring.

Ill check if there are other cases.

ybiquitous:
Thanks. If we drop the CJS support in the next major version, this confusion will disappear. 😅

ybiquitous (approved):
LGTM, thanks 👍🏼

Mouvedia:
@ybiquitous I am gonna revert the title change because the type actually changed.
i.e. if it was just inlined all rules would have `Promise<Rule>` as a return type

ybiquitous:
The reason I added "Refactor to" is to express that this PR doesn't need changelog updates, although there is no way around it since it's already merged. 😓

Mouvedia:
> The reason I added "Refactor to" is to express that this PR doesn't need changelog updates, although there is no way around it since it's already merged. 😓

It's fine; check the commit message, it doesn't match the PR's title.

PR #7967, opened 2024-08-27
Refactor to fix `CoreRules` type inconsistencies

Mouvedia:

https://github.com/stylelint/stylelint/pull/7954#discussion_r1731291043 and https://github.com/stylelint/stylelint/pull/7954#discussion_r1731611549

### explanation

This PR switch from the documentation to the implementation as the source of truth for the type of the rules.
This is required if we want to validate configs in the wild.
e.g. ESLint has [`defineConfig`](https://github.com/eslint-types/eslint-define-config)

#7954 was not controversial because it was just reflecting the documentation.
Now that we are planning to actually use the type, it needs to be accurate.
e.g. for `PublicApi['rules']` or to auto-generate the documentation
i.e. you should consider this PR a documentation fix because in the long run it will be

There's a convention which a lot of rule READMEs currently omit: if an option accepts an array of string and/or regex it also accepts a string and/or regex directly. This commit is trying to fix that specifically.
e.g. `alpha-value-notation: ["percentage", { exceptProperties: "foo" }]` vs `alpha-value-notation: ["percentage", { exceptProperties: ["foo"] }]`
i.e. options that previously had a type of `string[]` or `StringsOrRegexes`
The inverse is not always true though.
e.g. pattern rules and `selector-max-specificity`

### disclaimer

I am expecting the review of this PR to be tedious and lengthy.

### exceptions found

- `ignoreValues` of `value-no-vendor-prefix`
- `ignoreLonghands` of `declaration-block-no-redundant-longhand-properties`

### follow-up

Either create an issue to manually fix the documentation based on this PR or add the _status: ready to implement_ label to #7608.

### DIFF
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -92,7 +92,7 @@ declare namespace stylelint {
 
 	/** @internal */
 	export type DisableOptions = {
-		except?: StringsOrRegexes;
+		except?: Array<StringOrRegex>;
 		severity?: Severity;
 	};
 
@@ -279,9 +279,8 @@ declare namespace stylelint {
 	};
 
 	type StringOrRegex = string | RegExp;
-	type StringsOrRegexes = Array<StringOrRegex>;
-	type OneOrMoreStrings = string | string[];
-	type Primary = number | true | StringOrRegex | StringsOrRegexes | Record<string, any>;
+	type OneOrMany<S> = S | S[];
+	type Primary = number | true | OneOrMany<StringOrRegex> | Record<string, any>;
 	type Secondary = Record<string, any>;
 	type CoreRule<P extends Primary, S extends Secondary = any> = Rule<P, S>;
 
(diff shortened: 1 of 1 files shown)


types/stylelint/index.d.ts
```diff
@@ -279,147 +279,159 @@ declare namespace stylelint {
 	};
 
 	type StringOrRegex = string | RegExp;
-	type StringsOrRegexes = Array<StringOrRegex>;
-	type OneOrMoreStrings = string | string[];
-	type Primary = number | true | StringOrRegex | StringsOrRegexes | Record<string, any>;
+	type OneOrMany<S> = S | S[];
```
Mouvedia, line 282: `type-fest` calls it [`Arrayable`](https://github.com/sindresorhus/type-fest/blob/main/source/arrayable.d.ts).

types/stylelint/index.d.ts
```diff
@@ -279,147 +279,159 @@ declare namespace stylelint {
...
-				ignoreProperties: StringsOrRegexes;
+				ignoreProperties: OneOrMany<StringOrRegex>;
 			}
 		>;
 		'declaration-block-no-redundant-longhand-properties': CoreRule<
 			true,
-			{ ignoreShorthands: StringsOrRegexes; ignoreLonghands: string[] }
+			{
+				ignoreShorthands: OneOrMany<StringOrRegex>;
+				ignoreLonghands: string[];
```
ybiquitous, line 374: [note] Out-of-scope. `ignoreLonghands` should be `OneOrMany<StringOrRegex>`. 😅

ybiquitous (approved):
Thank you. LGTM 👍🏼 I love the new `OneOrMany<S>` type. 😄 

If you're okay, can you merge this PR itself, not including documentation changes? If doing so, changing the PR title is required, though.

Step-by-step changes could reduce our work load, and this PR is worth being included in the next release, I think.

types/stylelint/index.d.ts
Mouvedia: https://github.com/stylelint/stylelint/pull/7611#discussion_r1568916551

Mouvedia:
> If you're okay, can you merge this PR itself, not including documentation changes? If doing so, changing the PR title is required, though. (e.g. "Refactor to fix CoreRules type inconsistency"?)

OK.

types/stylelint/index.d.ts
```diff
@@ -501,103 +517,114 @@ declare namespace stylelint {
...
 			{
-				ignoreContextFunctionalPseudoClasses: StringsOrRegexes;
-				checkContextFunctionalPseudoClasses: StringsOrRegexes;
+				ignoreContextFunctionalPseudoClasses: OneOrMany<StringOrRegex>;
+				checkContextFunctionalPseudoClasses: OneOrMany<StringOrRegex>;
 			}
 		>;
 		'selector-max-pseudo-class': CoreRule<number>;
-		'selector-max-specificity': CoreRule<string, { ignoreSelectors: StringsOrRegexes }>;
+		'selector-max-specificity': CoreRule<string, { ignoreSelectors: OneOrMany<StringOrRegex> }>;
```
Mouvedia, line 553: being reviewed
WIP

Mouvedia:
@pengzhanbo 

Since you are the maintainer of https://github.com/stylelint-types/stylelint-define-config I think you might be interested by `CoreRules`. Based on https://github.com/stylelint-types/stylelint-define-config/commits/main/src/rules/stylelint it seems that you had to update these types regularly.

PR #7969, opened 2024-08-28
Add support for `string | RegExp` to `ignoreValues` option of `value-no-vendor-prefix`

Mouvedia:

see exceptions section of https://github.com/stylelint/stylelint/pull/7967#issue-2489168630

```diff
-               'value-no-vendor-prefix': CoreRule<true, { ignoreValues: Array<StringOrRegex> }>;
+               'value-no-vendor-prefix': CoreRule<true, { ignoreValues: OneOrMany<StringOrRegex> }>;
```

### DIFF
--- a/lib/rules/value-no-vendor-prefix/index.mjs
+++ b/lib/rules/value-no-vendor-prefix/index.mjs
@@ -24,10 +24,7 @@ const meta = {
 	fixable: true,
 };
 
-/** @typedef {import('stylelint').CoreRules[ruleName]} Rule */
-/** @typedef {Parameters<Rule>[1]['ignoreValues']} IgnoreValues */
-
-/** @type {Rule} */
+/** @type {import('stylelint').CoreRules[ruleName]} */
 const rule = (primary, secondaryOptions) => {
 	return (root, result) => {
 		const validOptions = validateOptions(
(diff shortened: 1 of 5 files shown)


lib/rules/value-no-vendor-prefix/index.mjs
```diff
@@ -48,17 +45,21 @@ const rule = (primary, secondaryOptions) => {
...
 
 		/**
-		 * @type {{ raw: IgnoreValues, unprefixed: IgnoreValues }}
+		 * @typedef {string | RegExp} IgnoreValue
+		 * @type {{ raw: IgnoreValue[], unprefixed: IgnoreValue[] }}
 		 * @todo use Object.groupBy once Node v20 support is dropped
 		 */
 		const groups = { raw: [], unprefixed: [] };
-
-		secondaryOptions?.ignoreValues?.forEach((value) => {
```
ybiquitous, line 56: [suggest] The idiom using `flat()` seems suitable here:

```diff
-	secondaryOptions?.ignoreValues?.forEach((value) => {
+	[secondaryOptions?.ignoreValues ?? []].flat().forEach((value) => {
```

lib/rules/value-no-vendor-prefix/index.mjs
```diff
@@ -48,17 +45,21 @@ const rule = (primary, secondaryOptions) => {
...
+		const ignoreValues = secondaryOptions?.ignoreValues;
+		const fillGroups = (/** @type {IgnoreValue} */ value) => {
 			const useRawValue = isRegExp(value) || (value.startsWith('/') && value.match(/\/i?$/));
```
ybiquitous, line 55: [note] Out-of-scope. We want to unify the similar logic with another opportunity.

https://github.com/stylelint/stylelint/blob/84f13fea65658a0ec6916567086ae73e8f77af00/lib/utils/matchesStringOrRegExp.mjs#L74-L78

ybiquitous (approved):
Thanks, overall LGTM. I've left one refactoring suggestion.

lib/rules/value-no-vendor-prefix/index.mjs
Mouvedia: Just to be sure, you mean at line 62?

ybiquitous: No, I think the change above is only necessary and `if...else if` at line 61-62 is unnecessary.

ybiquitous: I think the following works:

```js
[secondaryOptions?.ignoreValues ?? []].flat().forEach((value) => {
	const useRawValue = isRegExp(value) || (value.startsWith('/') && value.match(/\/i?$/));
	const group = useRawValue ? 'raw' : 'unprefixed';

	groups[group].push(value);
});
```

Mouvedia: So no type check, I get it.

Mouvedia: It needs some tweaking, but I see what you mean.

PR #7971, opened 2024-08-28
Add support for `string` to `ignoreLonghands` option of `declaration-block-no-redundant-longhand-properties`

Mouvedia:

see exceptions section of https://github.com/stylelint/stylelint/pull/7967#issue-2489168630

```diff
-                               ignoreLonghands: string[];
+                               ignoreLonghands: OneOrMany<string>;
```

### DIFF
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
@@ -175,7 +175,7 @@ a {
 
 ## Optional secondary options
 
-### `ignoreLonghands`: `["array", "of", "properties"]|"property"`
+### `ignoreLonghands: ["array", "of", "properties"]`
 
 Given:
 
--- a/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
+++ b/lib/rules/declaration-block-no-redundant-longhand-properties/README.md
@@ -175,7 +175,7 @@ a {
 
 ## Optional secondary options
 
-### `ignoreLonghands: ["string"]`
+### `ignoreLonghands`: `["array", "of", "properties"]|"property"`
 
 Given:
 
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -371,7 +371,7 @@ declare namespace stylelint {
 			true,
 			{
 				ignoreShorthands: OneOrMany<StringOrRegex>;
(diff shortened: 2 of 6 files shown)


lib/rules/declaration-block-no-redundant-longhand-properties/README.md
```diff
@@ -175,7 +175,7 @@ a {
 
 ## Optional secondary options
 
-### `ignoreLonghands: ["string"]`
+### `ignoreLonghands`: `["array", "of", "properties"]|"property"`
```
ybiquitous, line 178: [suggest] Let's keep consistency in the heading:

```suggestion
### `ignoreLonghands: ["array", "of", "properties"]`
```

For example, see:
- https://github.com/stylelint/stylelint/blob/ca3c2f82d23d2629e987b246345cdeb80faf78a5/lib/rules/declaration-block-no-redundant-longhand-properties/README.md?plain=1#L232
- https://github.com/stylelint/stylelint/blob/ca3c2f82d23d2629e987b246345cdeb80faf78a5/lib/rules/declaration-block-no-duplicate-custom-properties/README.md?plain=1#L46

types/stylelint/index.d.ts
```diff
@@ -371,7 +371,7 @@ declare namespace stylelint {
 			true,
 			{
 				ignoreShorthands: OneOrMany<StringOrRegex>;
-				ignoreLonghands: string[];
+				ignoreLonghands: OneOrMany<string>;
```
ybiquitous, line 374: [ask] Is it not easy to support also regex?

```suggestion
				ignoreLonghands: OneOrMany<StringOrRegex>;
```

ybiquitous (commented):
Thanks, overall LGTM. Can you look at a few reviews?

types/stylelint/index.d.ts
Mouvedia: It could be done in a separate PR. I cannot find any cases where you would want to use regexes though.
i.e. it's either `"text-decoration-thickness"` or `"background-size", "background-origin", "background-clip"` or `"transition-behavior"`

If someone requests it, we will add it.

ybiquitous: I see. Keeping it makes sense for now. 👍🏼

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #7975, opened 2024-08-31
Fix `possible` values

Mouvedia:

N/A

I am unsure if it can be considered a refactor or not.


lib/utils/validateOptions.mjs
```diff
@@ -36,7 +36,7 @@ export default function validateOptions(result, ruleName, ...optionDescriptions)
  * @param {(message: string) => void} complain
  */
 function validate({ possible, actual, optional }, ruleName, complain) {
-	if (actual === false && !ruleName.startsWith('report')) {
+	if (actual === false && /^report[A-Z]/.test(ruleName)) {
```
Mouvedia, line 39: see https://github.com/stylelint/stylelint/pull/6250#discussion_r941493067

lib/utils/__tests__/validateOptions.test.mjs
```diff
@@ -31,15 +31,15 @@ describe('validateOptions for primary options', () => {
...
-			possible: [true, null],
+			possible: [true],
 			actual: null,
 		});
 		expect(result.warn).toHaveBeenCalledTimes(0);
 	});
 
 	it('failing null equivalence', () => {
 		validateOptions(result, 'foo', {
-			possible: [true, null],
```
Mouvedia, line 42: https://github.com/stylelint/stylelint/blob/a8e598f18f683b0d1274e67400cce8e950bad5d7/lib/lintPostcssResult.mjs#L70

lib/rules/block-no-empty/index.mjs
```diff
@@ -26,7 +25,7 @@ const rule = (primary, secondaryOptions, context) => {
 			ruleName,
 			{
 				actual: primary,
-				possible: isBoolean,
+				possible: [true],
```
Mouvedia, line 28: `false` is not possible

lib/rules/number-max-precision/README.md
```diff
@@ -143,7 +143,7 @@ For example, with `2`.
 Given:
 
 ```json
-{"/^(oklch|oklab|lch|lab)$/", 4}
+{ "/^(oklch|oklab|lch|lab)$/": 4 }
```
Mouvedia, line 146: typo

lib/utils/validateOptions.mjs
```diff
@@ -36,7 +36,7 @@ export default function validateOptions(result, ruleName, ...optionDescriptions)
  * @param {(message: string) => void} complain
  */
 function validate({ possible, actual, optional }, ruleName, complain) {
-	if (actual === false && !ruleName.startsWith('report')) {
+	if (actual === false && !/^report[A-Z]/.test(ruleName)) {
```
Mouvedia, line 39: see https://github.com/stylelint/stylelint/pull/6250#discussion_r941493067

lib/rules/number-max-precision/index.mjs
```diff
@@ -188,8 +189,9 @@ const rule = (primary, secondaryOptions) => {
 		 */
 		function precisionInsideFunction(functionName, currentPrecision) {
 			const precisionForFunction = insideFunctions.get(functionName);
+			const hasPrecision = typeof precisionForFunction !== 'undefined';
 
-			if (isNumber(precisionForFunction)) return precisionForFunction;
```
Mouvedia, line 192: If `get` returns it's either a number or undefined because we already validated the values.

lib/utils/__tests__/validateOptions.test.mjs
```diff
@@ -63,14 +63,30 @@ describe('validateOptions for primary options', () => {
...
+		expect(result.warn).toHaveBeenCalledTimes(0);
+	});
+
+	it('false `actual` for rules named report-*', () => {
+		validateOptions(result, 'report-bar-qux', {
+			possible: [true, false],
+			actual: false,
+		});
+		expect(result.warn).toHaveBeenCalledTimes(1);
+	});
```
Mouvedia, line 89: A plugin might have a rule which name starts with `"report"` which primary option is `true`.

lib/utils/validateOptions.mjs
ybiquitous: [suggest] I'm not 100% sure, but can we explicitly specify `report*` rule names like this?

```js
const REPORT_RULE_NAMES = new Set(['reportDescriptionlessDisables', 'reportNeedlessDisables', 'reportInvalidScopeDisables']);
```

```suggestion
	if (actual === false && !REPORT_RULE_NAMES.has(ruleName)) {
```

ybiquitous (commented):
Thanks for the pull request.

> I am unsure if it can be considered a refactor or not.

I think this change will not affect most users, but adding a changelog item for a patch release seems fine for me, just in case.

.changeset/light-elephants-applaud.md
```diff
@@ -0,0 +1,10 @@
+---
+"stylelint": patch
+---
+
+Changed: stricter `possible` values for:
+
+- `block-no-empty`
+- `declaration-property-max-values`
+- `declaration-property-value-no-unknown`
+- `number-max-precision`
```
ybiquitous, line 10: [suggest] Can you change it to one line? I believe multiple lines are not supported by Changeset.

Mouvedia, line 10:
```suggestion
Changed: stricter `possible` values for `block-no-empty`, `declaration-property-max-values`, `declaration-property-value-no-unknown` and `number-max-precision`
```

.changeset/light-elephants-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Changed: stricter `possible` values for `block-no-empty`, `declaration-property-max-values`, `declaration-property-value-no-unknown` and `number-max-precision`
```
ybiquitous, line 5: [suggest] Sorry for my repeated comments. How about using "Fixed" instead of "Changed" to clarify this PR is a bug fix? In addition, since "possible" seems internal, so I'd like to use "primary" and "secondary" instead for general users.

```suggestion
Fixed: primary or secondary options for `block-no-empty`, `declaration-property-max-values`, `declaration-property-value-no-unknown` and `number-max-precision` to be stricter
```

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #7981, opened 2024-09-02
Fix string formatter handling of log symbols

### LINKED ISSUE #5741: Fix use of unicode symbols in formatters for native Windows cmd

1. Run stylelint on Windows 10 native cmd (haven't tried 11)

Doesn't matter as long as you get an error or one of the symbols here:

https://github.com/stylelint/stylelint/blob/e3615618321833fe4b0b57c065416da977c87def/lib/formatters/stringFormatter.mjs#L34-L39

stylelint "**/*.{css,scss}" --cache --cache-location .cache/.stylelintcache --rd

14.1.0

The characters should show properly

### What actually happened?

image

_No response_

### Proposal to fix the bug

_No response_

Mouvedia:

Closes #5741

https://github.com/stylelint/stylelint/pull/5562/files#diff-0d10acb017f5fc20b4b5eb99470f83c79384c9014f32f1c9c7ddf4cfed760211R192
https://github.com/stylelint/stylelint/issues/5741#issuecomment-979541612
https://github.com/sindresorhus/is-unicode-supported/blob/main/index.js


.changeset/four-cameras-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: string formatter handling of log symbols
```
ybiquitous, line 5: [suggest] How about this for consistency and clarity? 👇🏼 

```suggestion
Fixed: `string` formatter handling of log symbols for non-Unicode environments 
```

ybiquitous (commented):
Thank you, overall LGTM 👍🏼 

Can you look at my review about the changelog?

.changeset/four-cameras-dream.md
Mouvedia:
```suggestion
Fixed: `string` formatter handling of log symbols for non-Unicode terminals 
```

ybiquitous: Sounds good 👍🏼

ybiquitous (approved):
Thank you. LGTM 👍🏼

Mouvedia:
@ybiquitous since it's the default formatter, I am being extra cautious.

ybiquitous (approved):
LGTM. Thank you.

Just out of curiosity, but do you have a plan to return the support for `rxvt-unicode` upstream? In the future of our pure ESM release, we may get back to `is-unicode-supported`.

Mouvedia:
> Just out of curiosity, do you have a plan to return the support for `rxvt-unicode` upstream? In the future of our pure ESM release, we may get back to `is-unicode-supported`.

I might submit an issue.

To be honest, I think the original package is probably handling it roughly.
e.g. maybe checking the encoding or the version might be necessary
So if I were patching it, it would require a lot of time to test older versions of the terminals listed.
I suppose this package assumes the latest version of these terminals; which is also a valid interpretation.

ybiquitous:
I see. So, can you leave a code comment on the difference between `is-unicode-supported` and `lib/utils/isUnicodeSupported.mjs`? The comment could help us understand why we have stopped using `is-unicode-supported` in the future.

Mouvedia:
> I see. So, can you leave a code comment on the difference between `is-unicode-supported` and `lib/utils/isUnicodeSupported.mjs`? The comment could help us understand why we have stopped using `is-unicode-supported` in the future.

I will add a permalink to the JSDoc comment.
The rest can be handled with blame.

ybiquitous (approved):
Thanks. The added comment is great. 👍🏼

Mouvedia:
@ybiquitous https://github.com/sindresorhus/is-unicode-supported/commit/c2d514b417d058774c5e7cb0c98ca9316aae1a7f

ybiquitous:
Nice. Can you backport?

Mouvedia:
> […] In the future of our pure ESM release, we may get back to `is-unicode-supported`.

@ybiquitous We can replace the `isUnicodeSupported` utility by [`is-unicode-supported`](https://www.npmjs.com/package/is-unicode-supported).

ybiquitous:
@Mouvedia Good catch! Appreciated it if you could open a PR to switch to the `is-unicode-supported` package. 👍🏼

Issue #7836, opened 2024-07-10
Constructing replacement text for fixes without bleed or side effects


romainmenke:
see :
- https://github.com/stylelint/stylelint/pull/7802#discussion_r1653241078
- https://github.com/stylelint/stylelint/issues/7784

The purpose of this issue is to explore how we can construct replacement texts for fixes at a more theoretical level.

This issue assumes that we will require rules to expose replacement text and ranges to solve : https://github.com/stylelint/stylelint/issues/7192

### Multiple parsers / AST's

We don't have a single AST, we use multiple parsers and produce multiple distinct AST's.

This implies that some rules work in multiple stages:
1. parse a string with a specialized parser
2. detect a problem in the specialized AST
3. warn about the problem or apply a fix
4. serialize the specialized AST
5. apply the modified string to the PostCSS AST

Multiple subsequent fixes will mutate the same AST and serializing the entire AST will contain previous fixes.

This is bad because an external tool should be able to apply fixes individually.

This is the example given in : https://github.com/stylelint/stylelint/pull/7802#discussion_r1653241078

### Constructing replacement text is non-trivial

We prefer to work with parsed values from an AST when detecting and fixing problems.
The replacement text however is what is present in the output.

One is decoded and unescaped, the other is encoded and might contain escape sequences.

With the `@csstools/css-tokenizer` we store and expose both:

```css
/**
 * The CSS Token interface
 *
 * @remarks
 * CSS Tokens are fully typed and have a strict structure.
 * This makes it easier to iterate and analyze a token stream.
 *
 * The string representation and the parsed value are stored separately for many token types.
 * It is always assumed that the string representation will be used when stringifying, while the parsed value should be used when analyzing tokens.
 */
export interface Token<T extends TokenType, U> extends Array<T | string | number | U> {
	/**
	 * The type of token
	 */
	0: T,
	/**
	 * The token representation
	 *
	 * @remarks
	 * This field will be used when stringifying the token.
	 * Any stored value is assumed to be valid CSS.
	 *
	 * You should never use this field when analyzing the token when there is a parsed value available.
	 * But you must store mutated values here.
	 */
	1: string,
	/**
	 * Start position of representation
	 */
	2: number,
	/**
	 * End position of representation
	 */
	3: number,
	/**
	 * Extra data
	 *
	 * @remarks
	 * This holds the parsed value of each token.
	 * These values are unescaped, unquoted, converted to numbers, etc.
	 *
[...]

PR #7985, opened 2024-09-05
Add fixed section to the output of the `verbose` formatter

Mouvedia:

#4664

1. https://github.com/stylelint/stylelint/issues/4664#issuecomment-2167186271

2. `fixed` was available non-officially since version 16.7.0
i.e. `_postcssResult` is internal

3. if we want to actually close #​4664 we will have to decide how to expose that data publicly


romainmenke:
> `fixed` was available non-officially since version 16.7.0
i.e. `_postcssResult` is internal

Is there a guideline on this?

Should all formatters only use publicly available information and API's?

In other words, should we first discuss making this publicly available and how?

Mouvedia:
> > `fixed` was available non-officially since version 16.7.0
> > i.e. `_postcssResult` is internal
> 
> Is there a guideline for using features in core formatters? Should all formatters only use publicly available information and API's?
> 
> In other words, should we first discuss making this publicly available and how?

If it was a third-party formatter I would be against it; for a core formatter it doesn't matter from where we are getting the data.
i.e. we can switch the source to the public interface once it becomes available

If there's a problem with this PR, it would be that the new section doesn't account for the plugins that are still using `context.fix`. I wish it could be somehow hinted at, that there are N deprecated rules.
e.g. if the `CONTEXT_FIX` getter in `lintPostcssResult` stored those rules in an array we could display that information
i.e. the total count of fixes is inaccurate due to these rules

Mouvedia:
Thinking about it a bit more, I think there's a way to get a somewhat accurate total fixes count.
I would imply adding two things:

- an array for the deprecated rules
- something that checks this array in `report.mjs` and then set their `fixerData`

I would like to separate this change from this PR.
i.e. I would to add the 80% solution for the non-deprecated rules and then add the enhancement—the last 20%—subsequently

romainmenke:
> If there's a problem with this PR, it would be that the new section doesn't account for the plugins that are still using context.fix. I wish it could be somehow hinted at, that there are N deprecated rules.
e.g. if the CONTEXT_FIX getter in lintPostcssResult stored those rules in an array we could display that information
i.e. the total count of fixes is inaccurate due to these rules

I don't think that is really important.

It's a short term pain and we are already causing sufficient noice in the plugin ecosystem with the deprecation on `context.fix` 😅 

Emitting even more messages that some rules are using a deprecated API seems redundant to me :)

----

My thoughts were more that if we are exposing a private API indirectly, then maybe we should take the time to finalize it and make it available for 3rd party formatters.

If users find this information useful, they might also want it in their preferred formatter.

Not really a blocker, just someone I thought might be nice.

Mouvedia:
> Not really a blocker, just _something_ I thought might be nice.

I agree but I would like to expose both `fixed` _and_ `range` at the same time to close #4664.
If you check his demand you will see that the output contains the range of the errors/warnings that have been fixed.
i.e. I feel that exposing `fixed` without `range` would be too hasty
For `range` to be exposed, we will have to come up with a resolution for #7836.
That's the summary.

Mouvedia:
> Not really a blocker, just _something_ I thought might be nice.

I agree but I would like to expose both `fixed` _and_ `range` at the same time to close #4664.
If you check his demand you will see that the output contains the range of the errors/warnings that have been fixed.
i.e. I feel that exposing `fixed` without `range` would be too hasty

ybiquitous:
This concern on <https://github.com/stylelint/stylelint/pull/7985#issuecomment-2336836266> makes sense to me. I think it's better not to expose fixers' data yet.

lib/formatters/verboseFormatter.mjs
```diff
@@ -82,6 +82,24 @@ export default function verboseFormatter(results, returnValue) {
...
 	}
 
+	const fixedRules = getFixedRules(results);
+
+	if (fixedRules.length) {
+		let lines = '\n';
+		let total = 0;
+
+		for (const [name, count] of fixedRules) {
+			const meta = returnValue.ruleMetadata[name] || {};
```
ybiquitous, line 92: [suggestion] `|| {}` seems to produce an extra object because `ruleLink()` accepts `undefined` at the 2nd argument:

```suggestion
			const meta = returnValue.ruleMetadata[name];
```

lib/formatters/verboseFormatter.mjs
```diff
@@ -133,3 +152,28 @@ function ruleLink(rule, metadata) {
 
 	return rule;
 }
+
+/** @param {LintResult[]} results */
+function getFixedRules(results) {
+	/** @type {Record<string, number>} */
+	const rules = {};
```
ybiquitous, line 159: [suggestion] `Map` seems easier to handle here:

```suggestion
	/** @type {Map<string, number>} */
	const rules = new Map();
```

lib/formatters/verboseFormatter.mjs
```diff
@@ -133,3 +152,28 @@ function ruleLink(rule, metadata) {
...
+		} = _postcssResult;
+
+		for (const ruleName in fixersData) {
+			if (Object.hasOwn(fixersData, ruleName)) {
+				const filteredArray = fixersData[ruleName]?.filter(({ fixed }) => fixed);
+				const count = filteredArray?.length;
+
+				if (count) rules[ruleName] = rules[ruleName] ? rules[ruleName] + count : count;
+			}
+		}
```
ybiquitous, line 175: [suggestion] `for...of` with `Object.entries()` can replace `for...in` with `Object.hasOwn()`. I think the code can be more straightforward.

```suggestion
		for (const [ruleName, fixerData] of Object.entries(fixersData)) {
			const count = fixerData.filter(({ fixed }) => fixed).length;

			if (count) rules.set(ruleName, (rules.get(ruleName) ?? 0) + count);
		}
```

lib/formatters/verboseFormatter.mjs
```diff
@@ -133,3 +152,28 @@ function ruleLink(rule, metadata) {
...
+	}
+
+	return Object.entries(rules);
```
ybiquitous, line 178: [suggestion] If this function returns a `Map`:

```suggestion
	return rules;
```

lib/formatters/verboseFormatter.mjs
```diff
@@ -82,6 +82,24 @@ export default function verboseFormatter(results, returnValue) {
 		printProblems('warning');
 	}
 
+	const fixedRules = getFixedRules(results);
+
+	if (fixedRules.length) {
```
ybiquitous, line 87: [suggestion] If `getFixedRules()` returns a `Map`:

```suggestion
	if (fixedRules.size) {
```

ybiquitous (changes requested):
Thanks for the pull request, and sorry for the late review.

Overall LGTM, but I have a few requests around refactoring. Can you have a look at my comments? Thanks.

ybiquitous (approved):
Thank you, LGTM 👍🏼

Issue #3052, opened 2017-12-04
Document `!important` caveat in `no-descending-specificity`


pke:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

A bug where the rule claims wrong ordering.

> Which rule, if any, is this issue related to?

`no-descending-specificity`

> What CSS is needed to reproduce this issue?

```css
a,
a:hover,
a:active {
  color: #6c9cd3;
}

footer a {
  color: white !important; /* override all links */
}
```

> What stylelint configuration is needed to reproduce this issue?

"stylelint": "^8.3.1",
"stylelint-config-standard": "^18.0.0",

> Which version of stylelint are you using?

`8.3.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

npm with `stylelint "src/**/*.css"`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

no

> What did you expect to happen?

No warnings to be flagged

> What actually happened (e.g. what warnings or errors you are getting)?

The following warnings were flagged:

```shell
 72:1  ×  Expected selector "footer a" to come before selector "a:hover"          no-descending-specificity
 72:1  ×  Expected selector "footer a" to come before selector "a:active"         no-descending-specificity
 78:1  ×  Expected selector "a:focus" to come before selector "footer a:hover"    no-descending-specificity
 78:1  ×  Expected selector "a:focus" to come before selector "footer a:active"   no-descending-specificity
```

hudochenkov:
Thanks for the report and for using the template.

Rule checks selectors only. It doesn't check declarations in a rule, and whether rule with lower selector specificity has declaration which is outweights declaration in a rule with higher selector specificity.

`a:hover` has specificity `0,0,1,1`. `footer` has specificity `0,0,0,2`. Rule works as expected.

On the other side it's still show violations if there is a clearly a set of more specific selectors:

```css
a,
a:hover,
a:active {
  color: #6c9cd3;
}

footer a,
footer a:hover,
footer a:active {
  color: white;
}
```

```
 7:1  ✖  Expected selector "footer a" to come before selector "a:hover"    no-descending-specificity
 7:1  ✖  Expected selector "footer a" to come before selector "a:active"   no-descending-specificity
```

While rule is still works as expected and reports correctly, it feels incorrect.

You can disable violation for this use case:

```css
a,
a:hover,
a:active {
    color: #6c9cd3;
}

[...]

PR #7986, opened 2024-09-08
Document `!important` caveat in `no-descending-specificity`

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/3052

No, it's self-explanatory.

### DIFF
--- a/lib/rules/no-descending-specificity/README.md
+++ b/lib/rules/no-descending-specificity/README.md
@@ -33,11 +33,19 @@ This rule resolves nested selectors before calculating the specificity of the se
 
 ## Limitations
 
+This rule has the following limitations:
+
+- does not have access to HTML or DOM structure
+- does not consider `!important`
+- does not consider individual properties
+
+This can lead to valid linting errors appearing to be invalid at first glance.
+
+It may be possible to restructure your CSS to remove the error, otherwise it is recommended that you disable the rule for that line and leave a comment saying why the error should be ignored. Note that disabling the rule will cause additional valid errors from being reported.
+
 ### DOM structure
 
 The linter can only check the CSS to check for specificity order. It does not have access to the HTML or DOM in order to interpret the use of the CSS.
-
-This can lead to valid linting errors appearing to be invalid at first glance.
 
 For example the following will cause an error:
 
@@ -52,13 +60,9 @@ This is a correct error because the `a:hover` on line 2 has a higher specificity
 
 This may lead to confusion because "the two selectors will never match the same `a` in the DOM". However, since the linter does not have access to the DOM it can not evaluate this, and therefore correctly reports the error about descending specificity.
 
-It may be possible to restructure your CSS to remove the error, otherwise it is recommended that you disable the rule for that line and leave a comment saying why the error should be ignored. Note that disabling the rule will cause additional valid errors from being reported.
-
 ### `!important`
 
 The linter only checks selectors and doesn't take `!important` into account.
-
-This can lead to valid linting errors appearing to be invalid at first glance.
 
 For example the following will cause an error:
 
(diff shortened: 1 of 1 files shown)


lib/rules/no-descending-specificity/README.md
```diff
@@ -52,6 +54,46 @@ This may lead to confusion because "the two selectors will never match the same
...
+The linter only checks selectors and doesn't take `!important` into account.
+
+This can lead to valid linting errors appearing to be invalid at first glance.
+
+For example the following will cause an error:
+
+<!-- prettier-ignore -->
+```css
+a:hover { top: 10px; }
+a { top: 10px; !important; }
```
ybiquitous, line 68: [typo]

```suggestion
a { top: 10px !important; }
```

lib/rules/no-descending-specificity/README.md
```diff
@@ -52,6 +54,46 @@ This may lead to confusion because "the two selectors will never match the same
...
+```css
+a:hover { top: 10px; }
+a { left: 10px; }
+```
+
+This is a correct error because the `a:hover` on line 1 has a higher specificity than the `a` on line 2.
+
+This may lead to confusion because both rules contain different declarations and there isn't any conflict between either. However, the linter only evaluates selectors, and therefore correctly reports the error about descending specificity.
+
+It may be possible to restructure your CSS to remove the error, otherwise it is recommended that you disable the rule for that line and leave a comment saying why the error should be ignored.
```
ybiquitous, line 95: [suggest] This is not a strong opinion. "It may be possible to restructure ..." sentence is repeated in each section under "Limitations". If reducing this repetition would make sense, moving it right after Limitation might be good, e.g.

```markdown
## Limitations

This rule has the following limitations. It may be possible to restructure your CSS to ...

### DOM structure
...
```

ybiquitous (commented):
Thanks for the pull request! Overall LGTM except for my trivial reviews 👍🏼

ybiquitous:
@jeddy3 It'd be helpful if you could be available and take a quick look at this PR. 🙏🏼

lib/rules/no-descending-specificity/README.md
romainmenke: Good idea! Thank you

I've moved a bit more to right after `Limitations`:

```md
## Limitations

This rule has the following limitations:

- does not have access to HTML or DOM structure
- does not consider `!important`
- does not consider individual properties

This can lead to valid linting errors appearing to be invalid at first glance.

It may be possible to restructure your CSS to remove the error, otherwise it is recommended that you disable the rule for that line and leave a comment saying why the error should be ignored. Note that disabling the rule will cause additional valid errors from being reported.
```

Please let me know what you think

ybiquitous: Thank you, it's much better! 👍🏼

ybiquitous (approved):
LGTM 👍🏼 

Let's wait for @jeddy3's feedback.

lib/rules/no-descending-specificity/README.md
```diff
@@ -31,12 +31,22 @@ This rule only compares rules that are within the same media context. So `a {top
...
 
-## DOM Limitations
+## Limitations
 
-The linter can only check the CSS to check for specificity order. It does not have access to the HTML or DOM in order to interpret the use of the CSS.
+This rule has the following limitations:
+
+- does not have access to HTML or DOM structure
+- does not consider `!important`
+- does not consider individual properties
```
jeddy3, line 40:
```suggestion
This rule doesn't:

- have access to HTML or DOM structure
- consider `!important`
- consider individual properties
```

Minor style nit when writing lists.

jeddy3 (commented):
The additions nicely explain the limitations. It's a bit of a gnarly rule, especially when nesting is used...

I've suggested one stylistic change to be consistent with how we write lists elsewhere.

lib/rules/no-descending-specificity/README.md
romainmenke: I agree, this is better, thank you @jeddy3 🙇

jeddy3 (approved):

romainmenke:
Thank you for the reviews 🙇

PR #7992, opened 2024-09-09
Fix `keyframe-selector-notation` reported ranges

romainmenke:

There were quite a few issues in this rule:
- parsing of keyframe selectors wasn't quite right
- inconsistent normalization of selectors
- [using `word` in `report()`](https://stylelint.io/demo/#N4Igxg9gJgpiBcIACBrGBPAZgJwIYFsYBnAAkwghOBIB0A7GgFwEYAGVgUgBp6nOrejJpAA2EbPBLYYUANyCAvvQUguITAEsRMAHIE4iGAA8CAB20A6MESKrwEOpoDmCEMHolaIY4xh0oRDQIJADaHp5eRIzo2iIadIwAtJCOGk6JUbj+uNhQQeEAujx0nkHYAK7agcHuJRFBaFh4hBkw2mCM4ol0EIy4jBoOQZJhII0A7uJ5IAXhSnQqainOAGLi+P2uAFZEDnawpraItaUgUTFt8ULBQSL9xNfFp+exV8kOzhl92bnDXne+KL5BYgBRAA)

All fixed issues where either false positives or [pure code bugs](https://stylelint.io/demo/#N4Igxg9gJgpiBcIACBrGBPAZgJwIYFsYBnAAkwghOBIB0A7GgFxwnwBp6mAVAeSs8ZNIAGwjZ4JbDCgBuAQF968kGxCYAlsJgA5AnEQwAHgQAOWgHRgiRFeAh0NAcwQhg9ErRBHGMOlCI0CCQA2u4enkSM6FrC6nSMALSQDuqOCZG4frjYUIFhALocdB6B2ACuWgFBbsXhgWhYeITpMFpgjGIJdBCMuIzq9oESoSAmMNhgvr2OMAlldJVECQDu6owAFnEJDctiUAn2wugJAEaiYCiB+WGKdMqqyU4AYmL4fS4AVkT2trAmNogaiUQJFoq04oIgoFhH1iJCisDQTEIUl7E50r0sjkhp4YT5Ink7iB5EA), e.g. `Expected "TO" to be "undefined"
(keyframe-selector-notation)`

I think `Fixed: keyframe-selector-notation reported ranges` still covers most issues and is fine as a changelog entry. But happy to make adjustments :)


lib/rules/keyframe-selector-notation/index.mjs
```diff
@@ -72,27 +75,18 @@ const rule = (primary) => {
...
-				});
+					if (!keyframeSelector) return;
+
+					const { expFunc, fixFunc } = optionFuncs[primary];
 
-				/**
-				 * @param {import('postcss-selector-parser').Tag} tag
-				 * @param {OptionFuncs} funcs
-				 */
-				function checkSelector(tag, { expFunc, fixFunc }) {
```
romainmenke, line 93: `checkSelector` was only called in one place and I ran into some typing issues.

`selectors` is either `Root` or `undefined`.

From the perspective of `checkSelector` it is still possibly `undefined` even when we do a check for `undefined` in the outer scope. I don't fully understand why TypeScript considers this to be the case.

Instead of wrangling with the types or adding even more checks, I decided to inline this function as it was only called once anyway.

lib/utils/transformSelector.mjs
romainmenke, line 1: This helper was only used once in the entire project.
I think it is better to remove it and to have more consistency between rules.

Did I overlook something here? Is it maybe part of the public API?

Mouvedia: > Did I overlook something here? Is it maybe part of the public API?

AFAIK, no.

> [We recommended copying files to your project instead of importing them as we'll remove access to them in the next major release.](https://stylelint.io/migration-guide/to-16)

https://github.com/stylelint/stylelint/blob/92102823d70cd70e02bc009340e23c026724a6e1/lib/index.mjs#L22-L27

lib/rules/keyframe-selector-notation/__tests__/index.mjs
Mouvedia, line 1: question

If you were to add a failing test for `@keyframes foo { entry 0% {} entries 100% {} }` it would bail out?

lib/rules/keyframe-selector-notation/__tests__/index.mjs
```diff
@@ -193,6 +248,18 @@ testRule({
...
+					from, /* stylelint-disable-next-line keyframe-selector-notation */
+					from,
+					/* stylelint-disable-next-line keyframe-selector-notation */
+					from
+					{}
+				}
+			`,
+		},
 		...ACCEPT_TIMELINE_RANGE_NAME_TEST_CASES,
 	],
```
Mouvedia, line 264: question

Should it also handles typos when at least one keyword match?
Could you add a failing test for `@keyframes foo { to {}, fro {} }`?
i.e. I would expect a report for `to`

romainmenke: Nothing changed with respect to this aspect.
Is there a specific bit of code that makes you suspect an issue?

lib/rules/keyframe-selector-notation/__tests__/index.mjs
romainmenke: Any keyframes selector with more than 1 part would return `undefined`

lib/rules/keyframe-selector-notation/__tests__/index.mjs
Mouvedia: I felt that we were missing regression tests.
If you think it's not necessary, don't bother.

lib/rules/keyframe-selector-notation/index.mjs
```diff
@@ -72,27 +75,18 @@ const rule = (primary) => {
...
-						first = false;
+				selectors.each((selector) => {
+					const keyframeSelector = parseKeyframeSelector(selector);
 
-						checkSelector(selectorTag, optionFuncs[primary]);
-					});
-				});
+					if (!keyframeSelector) return;
+
+					const { expFunc, fixFunc } = optionFuncs[primary];
```
ybiquitous, line 87: [suggestion] Small refactoring ideas:

- It seems better to move `const { expFunc, fixFunc } = optionFuncs[primary];` before `root.walkAtRules()` call because `optionFuncs[primary]` is used only once.
- `Object.freeze()` for `optionFuncs` seems a bit overkill because it's clear not to be mutated.

E.g.

```js
const optionFuncs = {
	// as-is...
};

const { expFunc, fixFunc } = optionFuncs[primary];

root.walkAtRules(/^(-(o|moz|ms|webkit)-)?keyframes$/i, (atRuleKeyframes) => {
	// ...
```

lib/rules/keyframe-selector-notation/index.mjs
```diff
@@ -101,27 +95,56 @@ const rule = (primary) => {
...
 	};
 };
 
+/** @import { Selector, Tag } from 'postcss-selector-parser' */
+
+/**
+ * Full syntax is:
+ * `<keyframe-selector> = from | to | <percentage [0,100]> | <timeline-range-name> <percentage>`
+ *
+ * Only parses `from | to | <percentage [0,100]>` and returns `undefined` in any other case.
```
ybiquitous, line 127: [suggestion] From this comment, how about checking `keyframeSelectorKeywords` and `PERCENTAGE_SELECTORS` inside this `parseKeyframeSelector` function? If this idea makes sense, `parseKeyframeSelector()` would return two values: `keyframeSelector` and `normalizedSelector`, though.

ybiquitous (commented):
Thanks, overall LGTM 👍🏼 

I've left a few refactoring suggestions, so can you have a look? Thanks.

lib/rules/keyframe-selector-notation/index.mjs
romainmenke: Yes, makes sense! good catch 🙇

lib/rules/keyframe-selector-notation/index.mjs
romainmenke: I agree that it makes sense to group these checks.

ybiquitous (approved):
Thank you. LGTM 👍🏼

romainmenke:
Thank you for the reviews @ybiquitous, @Mouvedia 🙇

PR #7996, opened 2024-09-15
Fix `no-descending-specificity` reported ranges

romainmenke:

No, it's self-explanatory.


lib/rules/no-descending-specificity/index.cjs
```diff
@@ -142,6 +146,10 @@ const rule = (primary, secondaryOptions) => {
  * @returns {string | undefined}
  */
 function lastCompoundSelectorWithoutPseudoClasses(selectorNode) {
+	if (selectorNode.nodes.length === 0) return undefined;
+
+	selectorNode = normalizeSelector(selectorNode.clone());
```
romainmenke, line 151: This rule works by comparing selectors against previously seen selectors.
But any comments interleaved in any of those alter the selectors themselves.

For example:

```css
a, b {}
a /* comment */, b {}
```

Where `a` != `a /* comment */`

Using the same `normalizeSelector` utility as for `no-duplicate-selectors` to make selectors comparable.

ybiquitous (approved):
Thank you. LGTM 👍🏼

PR #8001, opened 2024-09-16
Fix `selector-anb-no-unmatchable` reported ranges and message

romainmenke:

The reported ranges for this rule were incorrect in various ways :)
Most impactful was the use of `column` instead of `offset` leading to cases like these: https://stylelint.io/demo/#N4Igxg9gJgpiBcICG8B2AXAFgWjJglgDZQAUADAJQA0AOqgEZpa4HHnV1hM55GmXA6NdJEIQATvAAE4mFADcdAL4gqIAGZEYAOSQBbOIhgAPfQAdCMAHRgAzrdXgIqTQHMEIQailSaIE+gwqFC2ftIA2nQ+Pn626ACeloT4GLjObthxSMFI4lB+UVIAusqOkC74rgBiEnpI6B4AVrbOjrBmDoheMSBxiTDJGGG+IIT1MHF+tN4jfUkp6GkVrpno2VC5+QgjY4GTIKVKQA

Screenshot of Stylelint errors for this rule showing that all errors are reported as if happening on the first line, while an error actually occurred on each of three lines

### DIFF
--- a/.changeset/funny-pets-cover.md
+++ b/.changeset/funny-pets-cover.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `selector-anb-no-unmatchable` reported ranges
+Fixed: `selector-anb-no-unmatchable` reported ranges and message
--- a/.changeset/funny-pets-cover.md
+++ b/.changeset/funny-pets-cover.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-anb-no-unmatchable` reported ranges
--- a/lib/rules/selector-anb-no-unmatchable/__tests__/index.mjs
+++ b/lib/rules/selector-anb-no-unmatchable/__tests__/index.mjs
@@ -1,3 +1,5 @@
+import { stripIndent } from 'common-tags';
+
 import rule from '../index.mjs';
 const { messages, ruleName } = rule;
 
@@ -43,6 +45,16 @@ testRule({
 			code: 'a:nth-child(3n + - 6) {}',
 			description: 'CSSTree parse error not flagged',
 		},
+		{
+			code: stripIndent`
+				/* stylelint-disable-next-line selector-anb-no-unmatchable */
+				a:nth-child(0), /* stylelint-disable-next-line selector-anb-no-unmatchable */
+				b:nth-child(0),
+				/* stylelint-disable-next-line selector-anb-no-unmatchable */
+				c:nth-child(0)
+				{}
+			`,
+		},
 	],
 
 	reject: [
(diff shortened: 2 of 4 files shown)


lib/rules/selector-anb-no-unmatchable/__tests__/index.mjs
```diff
@@ -43,112 +45,163 @@ testRule({
...
 	reject: [
 		{
 			code: 'a:nth-child(0) {}',
 			message: messages.rejected(':nth-child'),
 			line: 1,
-			column: 3,
+			column: 2,
 			endLine: 1,
-			endColumn: 16,
+			endColumn: 15,
```
ybiquitous, line 67: [suggestion] Should we align the reported range and the message? For example, in this case, should the column range be `2`-`12` (`:nth-child`) instead of `3`-`15` (`nth-child(0)`)?

```css
a:nth-child(0) {}
 ↑         ↑
123456789012345678

/* Unexpected unmatchable An+B selector ":nth-child" */
```

lib/rules/selector-anb-no-unmatchable/index.mjs
```diff
@@ -48,15 +49,20 @@ const rule = (primary) => {
...
+					context: 'selectorList',
+					positions: true,
+				});
+			} catch (e) {
+				return;
+			}
+
+			if (cssTreeSelectors.type !== 'SelectorList') return;
 
+			cssTreeSelectors.children.forEach((cssTreeSelector) => {
```
ybiquitous, line 65: [note] Nice. This should improve performance 👍🏼

ybiquitous (commented):
Thanks for the pull request. Overall LGTM, but I've left one concern.

lib/rules/selector-anb-no-unmatchable/__tests__/index.mjs
romainmenke: Or should we change the message so that it includes the problematic selector and not just the pseudo name?

`:nth-child` -> `:nth-child(0)`

Given that the pseudo is incorrect because specific arguments are used it might be better to include those arguments in the message?

ybiquitous: Make sense. Let's change the message to include arguments. 👍🏼

lib/rules/selector-anb-no-unmatchable/index.mjs
```diff
@@ -85,13 +93,20 @@ const rule = (primary) => {
...
 								message: messages.rejected,
-								messageArgs: [`:${pseudoClassSelector.name}`],
+								messageArgs: [
+									isNumber(index) && isNumber(endIndex)
+										? ruleSelector.slice(index, endIndex)
+										: generate(pseudoClassSelector),
+								],
 								node: ruleNode,
-								index: pseudoClassSelector.loc?.start.column,
-								endIndex: pseudoClassSelector.loc?.end.column,
```
romainmenke, line 94: As far as I could tell `csstree` doesn't seem to have a way to re-serialize to source. It only has a `generate` function which takes the AST and generates equivalent CSS.

This implies that some source might alter.
For example:
```diff
- :nth-child(+0n)
+ :nth-child(0n)
```

-----

At the same time `loc` in `pseudoClassSelector.loc` seems to be possibly `undefined`.
So we can't reliably `slice` the original input using `index` and `endIndex`.

-----

I chose to navigate this by first trying to use a `slice` of the original selector when `index` and `endIndex` are numbers and not `undefined`.

If that fails we fallback to `generate()`.

-----

Alternatively we could always use `generate()` and accept that it can produce selectors in messages that are subtly different from the actual source.

This would reduce complexity a bit in this rule.

.changeset/funny-pets-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-anb-no-unmatchable` reported ranges
```
ybiquitous, line 5: [suggestion] It makes sense to mention the improved message, right?

```suggestion
Fixed: `selector-anb-no-unmatchable` reported ranges and message
```

lib/rules/selector-anb-no-unmatchable/index.mjs
ybiquitous: Thanks for the elaboration. The fallback makes sense to me. 👍🏼 

Leaving a brief code comment about why this uses `generate()` seems valuable to me, but what do you think?

romainmenke: After fixing https://github.com/stylelint/stylelint/pull/8003 I think I understand better why and when `loc` will be `undefined`.

If I understand it correctly it will never be `undefined` when parsing with `{ positions: true }`

So I think we can safely do this:

```js
assert(pseudoClassSelector.loc);

const index = pseudoClassSelector.loc.start.offset;
const endIndex = pseudoClassSelector.loc.end.offset;

if (isUnmatchableNth(child.nth)) {
	report({
		message: messages.rejected,
		messageArgs: [ruleSelector.slice(index, endIndex)],
		node: ruleNode,
		index,
		endIndex,
		result,
		ruleName,
	});
}
```

Then we don't have the more complex code and also don't have to explain the complexity in a comment :)

ybiquitous: Much better! 👍🏼

romainmenke:
Thank you for the review @ybiquitous
I've made some changes, hopefully all good now :)

ybiquitous (approved):
LGTM. Thank you 👍🏼

Issue #1739, opened 2016-07-29
CSSLint compat


evilebottnawi:
# All rules

## Possible Errors

- [ ] [box-model](https://github.com/CSSLint/csslint/wiki/Beware-of-box-model-size)
- [ ] [display-property-grouping](https://github.com/CSSLint/csslint/wiki/Require-properties-appropriate-for-display) -> `"declaration-block-no-ignored-properties": true`, but we have more properties than in `CSSLint` maybe write plugin for this based on this rule and remove all checks not compat with `CSSLint`? I can do this.
- [ ] [duplicate-properties](https://github.com/CSSLint/csslint/wiki/Disallow-duplicate-properties) -> `"declaration-block-no-duplicate-properties": true` and we should add option `ignoreDifferenceValue`(maybe search best name)
- [x] [empty-rules](https://github.com/CSSLint/csslint/wiki/Disallow-empty-rules) -> `"block-no-empty": true`
- [ ] [known-properties](https://github.com/CSSLint/csslint/wiki/Require-use-of-known-properties) -> `"property-no-unknown"` i offer a little remake rule, ignore vendor by default, because all rules default ignore this, and add option `noVendorIgnore`(maybe best name), it is allow in future extend all `*-no-unknown` rules checking vendor prefixes if we have time to write or found repo with list.

## Compatibility

- [ ] [adjoining-classes](https://github.com/CSSLint/csslint/wiki/Disallow-adjoining-classes)
- [x] [box-sizing](https://github.com/CSSLint/csslint/wiki/Disallow-box-sizing) -> `"property-blacklist": ["box-sizing"]`
- [ ] [compatible-vendor-prefixes](https://github.com/CSSLint/csslint/wiki/Require-compatible-vendor-prefixes)
- [ ] [gradients](https://github.com/CSSLint/csslint/wiki/Require-all-gradient-definitions)
- [ ] [text-indent](https://github.com/CSSLint/csslint/wiki/Require-standard-property-with-vendor-prefix)
- [ ] [vendor-prefix](https://github.com/CSSLint/csslint/wiki/Require-standard-property-with-vendor-prefix)
- [ ] [fallback-colors](https://github.com/CSSLint/csslint/wiki/Require-fallback-colors)
- [ ] [star-property-hack](https://github.com/CSSLint/csslint/wiki/Disallow-star-hack)
- [ ] [underscore-property-hack](https://github.com/CSSLint/csslint/wiki/Disallow-underscore-hack)
[...]

PR #8008, opened 2024-09-25
Changed `shorthand-property-no-redundant-values` and `color-hex-length` report ranges

Mouvedia:

prepares #7192

see https://github.com/stylelint/stylelint/issues/7192#issuecomment-2338068550

The reported ranges of `*-no-vendor-prefix` rules will be fixed in a separate PR.


lib/rules/color-hex-length/index.cjs
```diff
@@ -65,12 +65,12 @@ const rule = (primary) => {
...
 
-				const index = declarationValueIndex(decl) + node.sourceIndex;
-				const endIndex = index + node.value.length;
+				const index = declarationValueIndex(decl) + 1 + node.sourceIndex;
+				const endIndex = index + node.value.length - 1;
 
 				report({
 					message: messages.expected,
-					messageArgs: [hexValue, expectedHex],
+					messageArgs: [hexValue.substring(1), expectedHex.substring(1)],
```
romainmenke, line 73: I don't fully understand why this is being changed.
Is this because `#` is present before and after the fix?

I think it is fine if we include `#` in both the reported range and the message. Only having the hex values, without the leading `#` is less clear.

I also think that this is less correct on a technical level.
`#fff` is a single token and a single component value in CSS.

lib/rules/shorthand-property-no-redundant-values/index.mjs
```diff
@@ -172,10 +173,15 @@ const rule = (primary) => {
...
 
+			const index = declarationValueIndex(decl);
+			const endIndex = index + decl.value.length;
+
 			report({
 				message: messages.rejected,
 				messageArgs: [value, shortestFormString],
 				node: decl,
+				index,
+				endIndex,
```
romainmenke, line 184: Same question here.

Is the issue you are trying to solve that there is a shared prefix both before and after the fix?

In this case the value and property are tightly coupled.
There is nothing wrong with writing `10px 10px` when looked at in isolation. The property name adds crucial context.

romainmenke (commented):
Thank you for this effort @Mouvedia,

However I don't think we should be making changes for https://github.com/stylelint/stylelint/issues/7192 at this time.

We should first resolve: https://github.com/stylelint/stylelint/issues/7836

I don't think the changes in this pull request are an improvement for users and I'd rather not make supporting changes for https://github.com/stylelint/stylelint/issues/7192 until we have a good understanding of how to resolve it fully.

If this were an obvious bug fix we could ship it, but I don't think that is the case here.

Mouvedia:
> I don't think the changes in this pull request are an improvement for users […]
> If this were an obvious bug fix we could ship it, but I don't think that is the case here.

I disagree for `shorthand-property-no-redundant-values`, it's an improvement over what we had before.
i.e. the rule name refers to the value hence the value should be reported (not the all declaration)

For `color-hex-length` it can be discussed.
Either "hex" in the name means hexadecimal or [`<hex-color>`](https://drafts.csswg.org/css-color/#typedef-hex-color).
i.e. it could be argued that we want to keep the `#`

What about removing the change for `color-hex-length` and merging the rest?

lib/rules/color-hex-length/index.cjs
Mouvedia: see https://github.com/stylelint/stylelint/pull/8008#issuecomment-2374012907

lib/rules/shorthand-property-no-redundant-values/index.mjs
Mouvedia: > If we want to align replacement text with the ranges we can also emit the entire declaration as the replacement text.

We could but that would be less precise / too large.
i.e. we will want the range to be as small as possible

> The property name adds crucial context.

The formatters already give enough context.
i.e. range and rule
I think that reporting the value is better and would be consistent with the current message arguments.

romainmenke:
For context: https://github.com/stylelint/stylelint/blob/148a71718f52f0c3c8cde2d28d6df4f25f348706/src/rules/property-shorthand-no-unnecessary-value/index.js#L96-L100

This rule has since the beginning always reported both the property name and the value.
That tells us at least that it hasn't been an issue for users and that we don't _need_ to change this :)

But I agree that it is arguably more consistent with other rules.
Is there an existing pattern/guide of when to include the property name and when not to?
Maybe this is just a legacy artifact?

`declaration-property-value-no-unknown` also only reports the invalid part and not the entire declaration even when it is the value that is invalid within the context of that specific declaration. Same as for `shorthand-property-no-redundant-values`.

So changing the reported range and message to only the value would bring it in line with `declaration-property-value-no-unknown`. Maybe there are also examples of rules that report the entire declaration?

Making rules more consistent in reported ranges and messages is generally useful and can be done separate from https://github.com/stylelint/stylelint/issues/7192 Such changes would be fine for me :)

-------

> For color-hex-length it can be discussed.
Either "hex" in the name means hexadecimal or [`<hex-color>`](https://drafts.csswg.org/css-color/#typedef-hex-color).
i.e. it could be argued that the # is part of "hex"
For me it was obviously hexadecimal length, I guess I was wrong in my assumption.

Given that the context is CSS colors it can only be `<hex-color>`, which starts with a `#` :)

Mouvedia:
> Is there an existing pattern/guide of when to include the property name and when not to?

AFAIK we only have a guide for messages.
It's mainly based on common sense and copy/pasting.
#1739 probably had an impact on the our previous conventions.
@jeddy3 would know more.

> Maybe this is just a legacy artifact?

Probably.

> Such changes would be fine for me :)

👍 

> Given that the context is CSS colors it can only be <hex-color>, which starts with a # :)

Alright.
Ill update the table.

lib/rules/shorthand-property-no-redundant-values/__tests__/index.mjs
```diff
@@ -121,286 +121,286 @@ testRule({
...
-			endColumn: 21,
+			endColumn: 20,
 		},
 		{
 			code: 'a { margin: 1Px 1pX; }',
 			fixed: 'a { margin: 1Px; }',
 			message: messages.rejected('1Px 1pX', '1Px'),
 			line: 1,
-			column: 5,
+			column: 13,
```
romainmenke, line 151: [suggestion] Maybe include the end line and end column in each test to have more coverage?

romainmenke (commented):
@Mouvedia This looks good to me, thank you.
I've left a single suggestions.

lib/rules/shorthand-property-no-redundant-values/__tests__/index.mjs
Mouvedia: Ill add `endLine`/`endColumn` to the `!important` test cases.

romainmenke (approved):
LGTM

thank you @Mouvedia

PR #8017, opened 2024-10-03
Refactor to replace `strip-ansi` with `stripVTControlCharacters` from `node:util`

### LINKED ISSUE #8016: chore(deps): replace `strip-ansi` with native utils
## Overview

Currently, this module uses `strip-ansi` to remove ANSI escape characters from formatter. This package is no longer needed as Node has provided [`util.stripVTControlCharacters`](https://nodejs.org/api/util.html#utilstripvtcontrolcharactersstr) that behaves the same way, thus eliminating the need for third-party package.

Namchee:

- Closes #8016 

This pull request replaces all instances of `strip-ansi` in all file(s) with `util.stripVTControlCharacters`.

The utility is available since Node 16.11, so there shouldn't be any compatibility issues as the latest version of `stylelint` requires Node 18.12 at minimum.

### DIFF
--- a/lib/__tests__/cli.test.mjs
+++ b/lib/__tests__/cli.test.mjs
@@ -4,9 +4,9 @@ import { Readable } from 'node:stream';
 import { fileURLToPath } from 'node:url';
 import path from 'node:path';
 import process from 'node:process';
+import { stripVTControlCharacters } from 'node:util';
 
 import { jest } from '@jest/globals';
-import stripAnsi from 'strip-ansi';
 import { stripIndent } from 'common-tags';
 
 import {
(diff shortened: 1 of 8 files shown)


lib/__tests__/cli.test.mjs
```diff
@@ -4,9 +4,9 @@ import { Readable } from 'node:stream';
 import { fileURLToPath } from 'node:url';
 import path from 'node:path';
 import process from 'node:process';
+import { stripVTControlCharacters } from 'node:util';
```
Mouvedia, line 7: nit

```suggestion
import { stripVTControlCharacters as stripAnsi } from 'node:util';
```

(if you want to reduce the diff)

ybiquitous (approved):
Thank you for this great refactoring 👍🏼 

I want to leave this change in the changelog with the prefix `Refactored`, but what do you think?

The following is only one case using `Refactored`:


https://github.com/stylelint/stylelint/blob/006cf1fa077f0cbff244300c523007a794fd2a17/CHANGELOG.md?plain=1#L193

jeddy3:
> I want to leave this change in the changelog with the prefix Refactored, but what do you think?

We've historically only used the changelog for user-facing changes to our public API, avoiding the noise of refactors.

We added the unique "Refactored: `.js` extension to `.mjs` and `.cjs`" entry and broke from the [Keep A Changelog](https://keepachangelog.com/en/1.1.0/) types because the change impacted users who import/required files, even though they weren't part of our public API.

jeddy3 (approved):

ybiquitous:
@jeddy3 Thanks for the info. That makes sense 👍🏼

PR #8024, opened 2024-10-08
Add `reportUnscopedDisables` and `--report-unscoped-disables`

### LINKED ISSUE #2292: Add `reportUnscopedDisables` flag

Mouvedia:

Closes #2292

This PR doesn't use

https://github.com/stylelint/stylelint/blob/3fca68706b744b18df53a2691dbd164508cd5d32/types/stylelint/index.d.ts#L145

because it's not getting populated by `assignDisabledRanges`. It can be changed later if it ever gets fixed.


docs/user-guide/cli.md
```diff
@@ -116,6 +116,10 @@ Produce a report of the `stylelint-disable` comments that used for rules that do
 
 Produce a report to clean up your codebase, keeping only the `stylelint-disable` comments that serve a purpose. [More info](options.md#reportneedlessdisables).
 
+### `--report-unscoped-disables, --rud`
+
+Produce a report of the configuration comments that are not scoped to one or more rules.
```
ybiquitous, line 121: [suggestion] How about adding a "More info" link like other `--report-*` flags?

```suggestion
Produce a report of the configuration comments that are not scoped to one or more rules. [More info](options.md#reportunscopeddisables).
```

docs/user-guide/configure.md
```diff
@@ -559,6 +560,18 @@ For example:
...
+Report configuration comments that are not scoped to at least one rule. A [`report*`](#report) property.
+
+For example:
+
+```json
+{
+  "reportUnscopedDisables": true
+}
+```
+
```
ybiquitous, line 574: [suggestion] Same here for "More info" link:

```suggestion

[More info](options.md#reportunscopeddisables).

```

docs/user-guide/options.md
```diff
@@ -212,6 +212,12 @@ CLI flags: `--report-needless-disables, --rd`
 
 Report `stylelint-disable` comments that don't actually match any lints that need to be disabled.
 
+## `reportUnscopedDisables`
+
+CLI flags: `--report-unscoped-disables, --rud`
+
+Report [configuration comments](../user-guide/ignore-code.md#parts-of-a-file) that aren't scoped to one or more rules.
```
ybiquitous, line 219: [note] Using "configuration comments" link instead of "`stylelint-disable` comments" is better. 👍🏼 

If you are interested, can you replace other "`stylelint-disable` comments" with such "configuration comments" links in this document on another pull request?

lib/unscopedDisables.mjs
```diff
@@ -0,0 +1,64 @@
+import { RULE_NAME_ALL } from './constants.mjs';
+
+import optionsMatches from './utils/optionsMatches.mjs';
+import reportCommentProblem from './utils/reportCommentProblem.mjs';
+import validateDisableSettings from './validateDisableSettings.mjs';
+
+/** @typedef {import('postcss').Node} Node */
+
+/** @param {import('stylelint').PostcssResult} postcssResult */
+export default function unscopedDisables(postcssResult) {
```
ybiquitous, line 10: [suggestion] For clarity, the `reportUnscopedDisables` name seems better to me (see also `reportDisables()`). How about renaming it like this?

```suggestion
export default function reportUnscopedDisables(postcssResult) {
```

We should have started with `report` for other functions, such as `needlessDisables()` or `invalidScopeDisables()`. I always feel it's hard to find such reporting functions. 😅

lib/unscopedDisables.mjs
```diff
@@ -0,0 +1,64 @@
...
+			for (const { start, end, node } of unscopedComments) {
+				if (alreadyReported.has(node)) continue;
+
+				if (start <= line && (end === undefined || end >= line)) {
+					report(node);
+					alreadyReported.add(node);
+				}
+			}
+		}
+	} else if (enabled) for (const { node } of unscopedComments) report(node);
```
ybiquitous, line 63: [suggestion] Please avoid the one-liner like this for readability. Using `{}` block seems more explicit:

```suggestion
	} else if (enabled) {
		for (const { node } of unscopedComments) report(node);
	}
```

ybiquitous (changes requested):
Thanks for the pull request. Can you take a look at my reviews?

docs/user-guide/cli.md
Mouvedia: I thought about it but there's nothing more there.
If you have something to add to _options.md#reportunscopeddisables_ that would warrant the _More info_ link, I would do it.
I could add a _See also_ link but _More info_ is ungenuine.

docs/user-guide/configure.md
Mouvedia: see https://github.com/stylelint/stylelint/pull/8024#discussion_r1796435578

lib/unscopedDisables.mjs
Mouvedia: > We should have started with `report` for other functions, such as `needlessDisables()` or `invalidScopeDisables()`.

I wanted to follow the previous convention.
I don't mind the rename.

docs/user-guide/cli.md
ybiquitous: I think we still need the "More info" link to associate the CLI flag with the Node.js API option. Also, for consistency, we should have the "More info" link like other flags have it.

docs/user-guide/options.md
Mouvedia: > If you are interested, can you replace other "stylelint-disable comments" with such "configuration comments" links in this document on another pull request?

Ill do it in this PR.

docs/user-guide/configure.md
```diff
@@ -559,6 +560,18 @@ For example:
...
+Report configuration comments that are not scoped to at least one rule. A [`report*`](#report) property.
+
+For example:
+
+```json
+{
+  "reportUnscopedDisables": true
+}
+```
```
Mouvedia, line 573:
````suggestion
```

[More info](options.md#reportunscopeddisables).

````

lib/unscopedDisables.mjs
ybiquitous: I still want to follow a better convention for a new code so that we can easily refactor the old code later. Since `reportDisables()` already exists, I think `reportUnscopedDisables()` also makes sense and isn't weird.

Mouvedia: I already did the rename.

Mouvedia:
@ybiquitous is there anything left to do?

Mouvedia:
@romainmenke Koba seems to be busy; can you take a look at this PR?
Apart from turning [this todo](https://github.com/stylelint/stylelint/pull/8024/files#diff-cba25f954347e963307879ca83b95b319e9003d9b4b51fd350d3f4110566d3a8R158) into a proper issue, I am not sure what else to do.

lib/unscopedDisables.mjs
romainmenke: I didn't even know that you could do this 😅 
I personally follow the "soft rule" that each line should only do one thing.
So I never even attempted to write a for loop as a one liner.

romainmenke (approved):
LGTM!

Thank you for working on this @Mouvedia 

I think this is a nice addition as disabling stylelint entirely is almost always an anti-pattern :)

lib/unscopedDisables.mjs
Mouvedia: Note that you can't go any further.
i.e. `bool && for (const foo of [1, 2]) fn(foo))` will trigger a syntax error because the for loop is a statement hence does does not return an expression

Mouvedia:
Ill merge it in the upcoming days.

> Thank you for working on this @Mouvedia

This is the oldest issue, it was first requested in #1814 before #​1871; 8 years ago.

Issue #8034, opened 2024-10-09
Fix `length-zero-no-unit` false positives for Sass variables


ArianeBouchardConformit:
### What is the problem you're trying to solve?

We left the rule enabled, and it broke some of our Sass calculations.

Consider the following calculation (simplified, as getting there in our actual codebase requires a mixin and a function that would be useless to pollute this thread with).

```scss
.my-element {
  $padding-top-base-offset: 5px;
  $padding-top-extra-offset: 0px;

  padding-top: calc($padding-top-base-offset + $padding-top-extra-offset); // Simplified in the output by Sass to 5px
  padding-left: 0px;
}
```

After autofixing, Stylelint gives the following result:

```scss
  $padding-top-base-offset: 5px;
  $padding-top-extra-offset: 0;

  padding-top: calc($padding-top-base-offset + $padding-top-extra-offset); // Error!
  padding-left: 0;
```

Sass now gives me an error like this:

![Image](https://github.com/user-attachments/assets/b84f6570-53db-4a68-b77b-367c1e8c4343)

### What solution would you like to see?

It's a bit unfortunate that Sass can't handle unitless zeros in calculations (and I've opened an issue [here](https://github.com/sass/sass/issues/3959)), but unless they decide to support them, it might be wise for Stylelint to either directly implement an exception for Sass variables or give us an option to exclude properties with a regex.

Sass has historically not been very open to my improvement suggestions on philosophical/edge case grounds, so I would be grateful if you guys could move on this.

As things stand, we sadly just cannot use this rule here in our project.

ybiquitous:
[...]

PR #8035, opened 2024-10-11
Fix `length-zero-no-unit` false positives for Sass variables

ryo-manba:

Closes https://github.com/stylelint/stylelint/issues/8034

No, it's self-explanatory.

### DIFF
--- a/lib/rules/length-zero-no-unit/__tests__/index.mjs
+++ b/lib/rules/length-zero-no-unit/__tests__/index.mjs
@@ -763,6 +763,12 @@ testRule({
 	fix: true,
 	customSyntax: 'postcss-scss',
 
+	accept: [
+		{
+			code: 'a { $padding-top-base-offset: 5px; $padding-top-extra-offset: 0px; }',
+			description: 'Ignore Sass variables with zero length units',
+		},
+	],
 	reject: [
 		{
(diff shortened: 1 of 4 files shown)


lib/rules/length-zero-no-unit/__tests__/index.mjs
```diff
@@ -806,38 +812,17 @@ testRule({
...
 		{
 			code: '@detached-ruleset: { font-size: 0px; };',
-			fixed: '@detached-ruleset: { font-size: 0; };',
-
-			message: messages.rejected,
-			line: 1,
-			column: 34,
-			endLine: 1,
-			endColumn: 36,
+			description: 'ignore Less detached ruleset with font-size',
```
ryo-manba, line 817: This change will also prevent variables in `@detached-ruleset` from being updated. 
Do you think this is appropriate?

jeddy3: Yes, I believe it's fine. These days, we typically prefer to ignore non-standard constructs in the built-in rules.

jeddy3 (approved):
LGTM, thank you.

ybiquitous (approved):
@ryo-manba Thanks for the quick fix! LGTM 👍🏼

PR #8036, opened 2024-10-11
Document issue types for maintainers

jeddy3:

To remove duplication of labels and issue types.


Mouvedia:
While we are at it what about a _os: windows_ label?

PR #8037, opened 2024-10-13
Fix `selector-attribute-name-disallowed-list` reported ranges

ryo-manba:

No, it's self-explanatory.

### DIFF
--- a/lib/rules/selector-attribute-name-disallowed-list/__tests__/index.mjs
+++ b/lib/rules/selector-attribute-name-disallowed-list/__tests__/index.mjs
@@ -1,4 +1,5 @@
 import rule from '../index.mjs';
+import { stripIndent } from 'common-tags';
 const { messages, ruleName } = rule;
 
 testRule({
@@ -18,6 +19,16 @@ testRule({
 		},
 		{
 			code: 'div[lang~="en-us"] { }',
+		},
+		{
+			code: stripIndent`
+				/* stylelint-disable-next-line selector-attribute-name-disallowed-list */
+				div[class*="foo"], /* stylelint-disable-next-line selector-attribute-name-disallowed-list */
+				div[id*="foo"],
+				/* stylelint-disable-next-line selector-attribute-name-disallowed-list */
+				div[data-foo*="foo"]
+				{ color: red; }
+			`,
 		},
 	],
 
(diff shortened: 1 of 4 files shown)


lib/rules/selector-attribute-name-disallowed-list/index.mjs
```diff
@@ -37,24 +37,26 @@ const rule = (primary) => {
 				return;
 			}
 
-			parseSelector(ruleNode.selector, result, ruleNode)?.walkAttributes((attributeNode) => {
-				const attributeName = attributeNode.qualifiedAttribute;
+			flattenNestedSelectorsForRule(ruleNode, result).forEach(({ selector }) => {
```
romainmenke, line 40: Do we need to switch to `flattenNestedSelectorsForRule`?

This rule only looks at individual attribute selector parts and doesn't do any analysis of the entire selector. So resolving nested selectors shouldn't be needed.

I think we can use the `getRuleSelector` utility instead.

e.g.

```js
parseSelector(getRuleSelector(ruleNode), result, ruleNode)?.walkAttributes((attributeNode) => {
```

romainmenke (commented):
Thank you for working on this issue @ryo-manba 🙇 
I've left a question about a potentially smaller change.

lib/rules/selector-attribute-name-disallowed-list/index.mjs
ryo-manba: Got it, using `getRuleSelector` helps to include comments in the selector. I've made the update.

[refactor to use getRuleSelector](https://github.com/stylelint/stylelint/pull/8037/commits/4d81b4ddea71f73da346ead5115e9d2a7db70b8d)

romainmenke (approved):
LGTM!
Thank you for your contribution.

(leaving this open in case other maintainers want to take a look)

lib/rules/selector-attribute-name-disallowed-list/__tests__/index.mjs
```diff
@@ -1,4 +1,5 @@
 import rule from '../index.mjs';
+import { stripIndent } from 'common-tags';
 const { messages, ruleName } = rule;
```
romainmenke, line 3:
```suggestion
import { stripIndent } from 'common-tags';

import rule from '../index.mjs';
const { messages, ruleName } = rule;
```

@ryo-manba We split 3rd party and internal module imports.
3rd party modules come first and are separated by an empty line before listing the internal module imports.

PR #8039, opened 2024-10-14
Fix `selector-attribute-operator-disallowed-list` reported ranges

ryo-manba:

No, it's self-explanatory.


lib/rules/selector-attribute-operator-disallowed-list/__tests__/index.mjs
```diff
@@ -1,4 +1,5 @@
 import rule from '../index.mjs';
+import { stripIndent } from 'common-tags';
 const { messages, ruleName } = rule;
```
romainmenke, line 3:
```suggestion
import { stripIndent } from 'common-tags';

import rule from '../index.mjs';
const { messages, ruleName } = rule;
```

romainmenke (approved):
LGTM!
Thank you for your contribution 🙇

PR #8042, opened 2024-10-14
Fix `selector-class-pattern` reported ranges

ryo-manba:

No, it's self-explanatory.

### DIFF
--- a/lib/rules/selector-class-pattern/__tests__/index.mjs
+++ b/lib/rules/selector-class-pattern/__tests__/index.mjs
@@ -1,5 +1,6 @@
-import rule from '../index.mjs';
 import { stripIndent } from 'common-tags';
+
+import rule from '../index.mjs';
 const { messages, ruleName } = rule;
 
 const basicAZTestsAccept = {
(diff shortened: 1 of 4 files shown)


lib/rules/selector-class-pattern/__tests__/index.mjs
```diff
@@ -182,6 +235,39 @@ testRule({
 			endLine: 1,
 			endColumn: 11,
 		},
+		{
+			code: stripIndent`
```
ryo-manba, line 239: The logic of `resolveNestedSelectors` wasn't changed, but confirmed to work.

image

lib/rules/selector-class-pattern/__tests__/index.mjs
```diff
@@ -1,4 +1,5 @@
 import rule from '../index.mjs';
+import { stripIndent } from 'common-tags';
 const { messages, ruleName } = rule;
```
romainmenke, line 3:
```suggestion
import { stripIndent } from 'common-tags';

import rule from '../index.mjs';
const { messages, ruleName } = rule;
```

romainmenke (approved):
LGTM!
Thank you for your contribution 🙇

PR #8045, opened 2024-10-14
Fix `selector-id-pattern` reported ranges

ryo-manba:

No, it's self-explanatory.

### DIFF
--- a/lib/rules/selector-id-pattern/__tests__/index.mjs
+++ b/lib/rules/selector-id-pattern/__tests__/index.mjs
@@ -1,5 +1,6 @@
-import rule from '../index.mjs';
 import { stripIndent } from 'common-tags';
+
+import rule from '../index.mjs';
 const { messages, ruleName } = rule;
 
 const basicAZTests = {
--- a/lib/rules/selector-id-pattern/__tests__/index.mjs
+++ b/lib/rules/selector-id-pattern/__tests__/index.mjs
@@ -1,4 +1,5 @@
 import rule from '../index.mjs';
+import { stripIndent } from 'common-tags';
 const { messages, ruleName } = rule;
 
 const basicAZTests = {
@@ -37,6 +38,16 @@ const basicAZTests = {
 			code: 'html { --custom-property-set: {} }',
 			description: 'custom property set in selector',
 		},
+		{
+			code: stripIndent`
+				/* stylelint-disable-next-line selector-id-pattern */
+				a #foo, /* stylelint-disable-next-line selector-id-pattern */
+				a #bar,
+				/* stylelint-disable-next-line selector-id-pattern */
+				a #foo
+				{}
+			`,
+		},
 	],
 };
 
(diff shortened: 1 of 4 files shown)


lib/rules/selector-id-pattern/index.cjs
```diff
@@ -38,20 +39,24 @@ const rule = (primary) => {
...
+				const selector = String(selectorNode).trim();
+				const endIndex = index + selector.length;
 
 				report({
 					result,
 					ruleName,
 					message: messages.expected,
 					messageArgs: [selector, primary],
 					node: ruleNode,
-					word: selector,
```
ryo-manba, line 54: Passing `word` only alerts the first selector in case of duplicate selectors, so it is not passed.
See: https://github.com/postcss/postcss/blob/1cc6ac3bf43a12460d3396e2cb76f14075daea1b/lib/node.js#L227-L229

lib/rules/selector-id-pattern/index.cjs
```diff
@@ -38,20 +39,24 @@ const rule = (primary) => {
...
-				if (normalizedPattern.test(selectorNode.value)) {
+			parseSelector(getRuleSelector(ruleNode), result, ruleNode)?.walkIds((selectorNode) => {
+				const { value, sourceIndex: index } = selectorNode;
+
+				if (normalizedPattern.test(value)) {
 					return;
 				}
 
-				const selector = String(selectorNode);
+				const selector = String(selectorNode).trim();
```
ryo-manba, line 49: The warnings were incorrect for the following pattern.

```css
/* a comment */
#A, /* a comment */
#B,
/* a comment */
#C {
  color: red;
}
```

image

Mouvedia: see https://github.com/stylelint/stylelint/pull/7959#discussion_r1730295653

i.e.

```suggestion
				const selector = selectorNode.toString().trim();
```

lib/rules/selector-id-pattern/index.cjs
Mouvedia: @ybiquitous if this is fixing another bug, should we add another changeset?

romainmenke: Only the reported ranges are incorrect by using `word`.
So the current changeset entry covers both fixes :)

lib/rules/selector-id-pattern/index.cjs
romainmenke: This change should be made in the `.mjs` file and then the project needs to be built.

lib/rules/selector-id-pattern/index.mjs
```diff
@@ -34,20 +35,24 @@ const rule = (primary) => {
...
-				if (normalizedPattern.test(selectorNode.value)) {
+			parseSelector(getRuleSelector(ruleNode), result, ruleNode)?.walkIds((selectorNode) => {
+				const { value, sourceIndex: index } = selectorNode;
+
+				if (normalizedPattern.test(value)) {
 					return;
 				}
 
-				const selector = String(selectorNode);
+				const selector = String(selectorNode).trim();
```
Mouvedia, line 45:
```suggestion
				const selector = selectorNode.toString().trim();
```

lib/rules/selector-id-pattern/__tests__/index.mjs
```diff
@@ -1,4 +1,5 @@
 import rule from '../index.mjs';
+import { stripIndent } from 'common-tags';
 const { messages, ruleName } = rule;
```
romainmenke, line 3:
```suggestion
import { stripIndent } from 'common-tags';

import rule from '../index.mjs';
const { messages, ruleName } = rule;
```

romainmenke (approved):
LGTM!

Thank you so much for working on these issues @ryo-manba 🙇

PR #8046, opened 2024-10-14
Fix `selector-combinator-allowed-list` reported ranges

ryo-manba:

No, it's self-explanatory.

### DIFF
--- a/lib/rules/selector-combinator-allowed-list/index.mjs
+++ b/lib/rules/selector-combinator-allowed-list/index.mjs
@@ -1,3 +1,4 @@
+import getRuleSelector from '../../utils/getRuleSelector.mjs';
 import isStandardSyntaxCombinator from '../../utils/isStandardSyntaxCombinator.mjs';
 import isStandardSyntaxRule from '../../utils/isStandardSyntaxRule.mjs';
 import { isString } from '../../utils/validateTypes.mjs';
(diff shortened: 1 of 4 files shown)


lib/rules/selector-combinator-allowed-list/index.mjs
Mouvedia, line 1: refactor
off topic

```js
					const { value } = combinatorNode;
					const normalizedValue = normalizeCombinator(value);
					if (primary.includes(normalizedValue)) {
						return;
					}
					const { sourceIndex: index, raws } = combinatorNode;
```

Could become

```js
					const { raws, sourceIndex: index, value } = combinatorNode;
					const normalizedValue = normalizeCombinator(value);

					if (primary.includes(normalizedValue)) return;
```

romainmenke:
> should `getDeclarationValue` become agnostic?
e.g. accept combinators

I am not sure what you mean? :)
I don't see the connection between declaration values and combinators in selectors.

romainmenke (approved):
LGTM 🎉

Mouvedia:
> I am not sure what you mean? :)

We could have reused `getFooValue` [here](https://github.com/stylelint/stylelint/pull/8046/files#diff-f310d487cc8a2d8be9f46a961bfb04a6fd4c07441af62d5088d7828815b4df72R51).

PR #8047, opened 2024-10-14
Fix `selector-combinator-disallowed-list` reported ranges

ryo-manba:

No, it's self-explanatory.

### DIFF
--- a/lib/rules/selector-combinator-disallowed-list/__tests__/index.mjs
+++ b/lib/rules/selector-combinator-disallowed-list/__tests__/index.mjs
@@ -1,4 +1,5 @@
 import rule from '../index.mjs';
+import { stripIndent } from 'common-tags';
 const { messages, ruleName } = rule;
 
 testRule({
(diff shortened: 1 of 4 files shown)


lib/rules/selector-combinator-disallowed-list/__tests__/index.mjs
```diff
@@ -1,4 +1,5 @@
 import rule from '../index.mjs';
+import { stripIndent } from 'common-tags';
 const { messages, ruleName } = rule;
```
romainmenke, line 3:
```suggestion
import { stripIndent } from 'common-tags';

import rule from '../index.mjs';
const { messages, ruleName } = rule;
```

ryo-manba:
Thanks for the review! Could you enable [automatic branch deletion](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-the-automatic-deletion-of-branches) after merges?

romainmenke:
> Could you enable [automatic branch deletion](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-the-automatic-deletion-of-branches) after merges?

That is already enabled, but I don't think it affects forks, which makes sense :)

Screenshot 2024-10-15 at 14 15 02

ryo-manba:
> That is already enabled, but I don't think it affects forks, which makes sense :)

Ah, right. Thanks for confirming.

romainmenke (approved):
LGTM 🎉

Issue #7873, opened 2024-07-22
Add "nesting-selector" as an available ignore option for `max-nesting-depth` rule


e-kulikov:
### What is the problem you're trying to solve?

In some cases it's worth using `&` nesting selector explicitly and as for me it shouldn't increase nesting count in the same logic as `pseudo-classes`.

The following rules are considered problems with `max-nesting-depth: [1, { ignore: ["pseudo-classes"] }]`
```css
.button {
  color: green;

  &:hover,
  &:focus {
    color: blue;
  }

  &[disabled] { /* <-- level 1 */
    &, /* <-- nesting selector is considered as level 2 */
    &:hover,
    &:focus {
      color: grey;
    }
  }
}
```

### What solution would you like to see?

Add `nesting-selector` as an available value for `ignore` option of the rule:

`max-nesting-depth: [1, { ignore: ["pseudo-classes", "nesting-selector"] }]`

romainmenke:
Thank you @e-kulikov for creating this issue.

I think this would be good to somehow support.

I am not sure about the exact implementation as `ignore` `pseudo-classes` only ignores rules when **each** selector is a pseudo class. https://github.com/stylelint/stylelint/tree/main/lib/rules/max-nesting-depth#ignore-pseudo-classes

```css
&, /* not a pseudo class */
&:hover,
&:focus {
  color: grey;
}
```

So I think it needs to a new option to ignore any `&` and/or and pseudo classes.

ybiquitous:
[...]

PR #8048, opened 2024-10-15
Fix `max-nesting-depth` ignore option for '&' and pseudo-classes

ryo-manba:

Closes https://github.com/stylelint/stylelint/issues/7873

No, it's self-explanatory.

### DIFF
--- a/lib/rules/max-nesting-depth/__tests__/index.mjs
+++ b/lib/rules/max-nesting-depth/__tests__/index.mjs
@@ -361,6 +361,73 @@ testRule({
 
 testRule({
 	ruleName,
+	config: [1, { ignoreRules: ['&'], ignore: ['pseudo-classes'] }],
+
+	accept: [
+		{
+			code: stripIndent`
+			a {
+				color: red;
+				& {
+					color: blue;
+				}
+				b {
+					color: gray;
+					& {
+						color: white;
+					}
+					&:hover {
+						color: orange;
+					}
+					&,
+					&:hover {
+						color: purple;
+					}
+				}
+			}
+		`,
+			description: 'Ensures "&" selectors and pseudo-classes are correctly ignored',
+		},
+	],
+});
+
+testRule({
+	ruleName,
+	config: [1, { ignoreRules: ['&'] }],
+
+	reject: [
+		{
+			code: stripIndent`
+					a {
+						color: red;
+						& {
+							color: blue;
+						}
+						b {
+							color: gray;
+							& {
+								color: white;
+							}
+							&,
+							&:hover {
+								color: purple;
+							}
+						}
+					}
+				`,
+			message: messages.expected(1),
+			line: 11,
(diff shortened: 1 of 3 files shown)


lib/rules/max-nesting-depth/__tests__/index.mjs
```diff
@@ -359,6 +359,73 @@ testRule({
...
+					&:hover {
+						color: orange;
+					}
+					&,
+					&:hover {
+						color: purple;
+					}
+				}
+			}
+		`,
```
ybiquitous, line 388: [suggestion] Minor. To save lines and increase visibility, how about making this change? (although from my preference)

```suggestion
			code: stripIndent`
				a {
					color: red;
					& { color: blue; }
					b {
						color: gray;
						& { color: white; }
						&:hover { color: orange; }
						&, &:hover { color: purple; }
					}
				}
			`,
```

lib/rules/max-nesting-depth/__tests__/index.mjs
```diff
@@ -359,6 +359,73 @@ testRule({
...
+							& {
+								color: white;
+							}
+							&,
+							&:hover {
+								color: purple;
+							}
+						}
+					}
+				`,
```
ybiquitous, line 417: [suggestion] The same idea.

```suggestion
			code: stripIndent`
				a {
					color: red;
					& { color: blue; }
					b {
						color: gray;
						& { color: white; }
						&, &:hover { color: purple; }
					}
				}
			`,
```

ybiquitous (commented):
@ryo-manba Thanks for the pull request. Overall LGTM, but could you take a look at my comments?

lib/rules/max-nesting-depth/index.mjs
```diff
@@ -124,7 +124,11 @@ const rule = (primary, secondaryOptions) => {
 			const normalized = parser().processSync(selector, { lossless: false });
 			const selectors = normalized.split(',');
 
-			return selectors.every((sel) => extractPseudoRule(sel));
+			const filteredSelectors = selectors.filter(
+				(sel) => !optionsMatches(secondaryOptions, 'ignoreRules', sel),
+			);
+
+			return filteredSelectors.every((sel) => extractPseudoRule(sel));
```
Mouvedia, line 131:
```suggestion
			const matches = optionsMatches.bind(null, secondaryOptions, 'ignoreRules');

			return selectors.filter((s) => !matches(s))
					.every((s) => extractPseudoRule(s));
```

I don't know which one is better but that's an alternative.
It's a matter of taste I guess.

ryo-manba: Thank you for the suggestion. I appreciate your input, but I feel that the code you proposed might have added a bit more complexity.

ybiquitous (approved):
Thank you. LGTM 👍🏼

PR #8060, opened 2024-10-20
Add `ignoreFunctions: []` to `media-query-no-invalid`

### LINKED ISSUE #7934: Add `ignoreFunctions: []` to `media-query-no-invalid`

```css
/* tailwindcss *\
@media (min-width: theme(screens.mt)) {
  ...
}
```

```js
{
'function-no-unknown': [
      true,
      {
        ignoreFunctions: ['theme'],
      },
    ],
}
```

vs code extension

```json 
{
   "stylelint": "^16.8.2",
   "stylelint-order": "^6.0.4",
   "stylelint-selector-bem-pattern": "^4.0.0",
   "tailwindcss": "^3.4.9",
}
```

It will not be underlined
image

### What actually happened?

https://stylelint.io/user-guide/rules/media-query-no-invalid/

_No response_

ryo-manba:

Closes #7934 

No, it's self-explanatory.


lib/rules/media-query-no-invalid/README.md
```diff
@@ -62,3 +62,39 @@ The following patterns are _not_ considered problems:
 ```css
 @media (color) and ((hover) or (width)) {}
 ```
+
+
+### `ignoreFunctions: ["/regex/", /regex/, "string"]`
```
ryo-manba, line 68: This PR isn't directly related, but it might be good to standardize the `ignoreFunctions` docs wording. https://github.com/search?q=repo%3Astylelint%2Fstylelint%20%23%23%23%20%60ignoreFunctions%3A%20%5B&type=code

romainmenke (commented):
Thank you @ryo-manba,

I've left a suggestion, please let me know what you think

ryo-manba:
@romainmenke 
It seems that the review hasn't been reflected yet. 
Could you please check it?

lib/rules/media-query-no-invalid/index.mjs
```diff
@@ -59,9 +71,25 @@ const rule = (primary) => {
...
+								if (!optionsMatches(secondaryOptions, 'ignoreFunctions', functionName)) {
+									shouldIgnore = false;
+								}
+							}
+						});
+
+						// Do not report if all custom functions are ignored.
+						if (!hasFunction || !shouldIgnore) {
+							complain(atRule, atRuleParamIndexValue, node);
+						}
```
romainmenke, line 92: I think it might be better to ignore the entire `GeneralEnclosed` if it contains anything that is ignored. (currently we only ignore functions)

This allows for a simpler implementation and is more in line with how we handle non-standard syntax in general. (any non-standard function will always become `GeneralEnclosed`)

I also don't think there is any way we can partially ignore contents of a `GeneralEnclosed`. The only option to have such behavior would be by replacing the non-standard syntax with a standard equivalent and to then re-evaluate.

```js
if (isGeneralEnclosed(node)) {
	let ignored = false;

	if ('ignoreFunctions' in secondaryOptions) { // no need to walk if there are no ignore options
		node.walk(({ node: childNode }) => {
			if (!isFunctionNode(childNode)) {
				return;
			}

			ignored = optionsMatches(secondaryOptions, 'ignoreFunctions', childNode.getName());

			if (ignored) {
				return false; // returning `false` stops iterations
			}
		});
	}

	if (!ignored) {
		complain(atRule, atRuleParamIndexValue, node);
	}

	return;
}
```

lib/rules/media-query-no-invalid/__tests__/index.mjs
```diff
@@ -282,3 +282,55 @@ testRule({
...
 	],
 });
+
+testRule({
+	ruleName,
+	config: [true, { ignoreFunctions: ['theme', 'fooFunction', '/^get.*$/'] }],
+
+	accept: [
+		{
+			code: '@media (min-width: theme(screens.md)) {}',
```
romainmenke, line 292: Functions could also be used to level: `@media theme(screens.md) {}`
Is that also covered by the current implementation?

romainmenke (commented):
Somehow my review comment was lost, thank you for pointing that out

lib/rules/media-query-no-invalid/index.mjs
romainmenke:
```css
@media (min-width: var(300px)), (min-width: theme(screens.md)) {}
```

Comma separated media query lists should still report errors

lib/rules/media-query-no-invalid/README.md
Mouvedia: I agree.
I think it should be done at the same time as #3372.

lib/rules/media-query-no-invalid/__tests__/index.mjs
```diff
@@ -282,3 +282,55 @@ testRule({
...
+		},
+		{
+			code: '@media (min-width: theme(screens.md) && fooFunction(300px)) {}',
```
Mouvedia, line 295: question

Should it be `and` instead of `&&`?

lib/rules/media-query-no-invalid/index.mjs
ryo-manba: @romainmenke 
Let me confirm your suggestion.

```css
@media (min-width: theme(screens.md) && fooFunction(300px)) {}
```

If `theme` is in the `ignoreList`:

- Currently, `theme` is ignored, but since `fooFunction` is not, an error is triggered.
- In your proposal, because `theme` is in the ignore list, the entire `GeneralEnclosed` block (including `fooFunction`) would be ignored, so no error would be thrown.

Is this understanding correct?

romainmenke: Yes, that is correct.

There simply is no way to recover after ignoring a part of a `GeneralEnclosed`.

Some samples:
- `@media (min-width: calc(theme(screens.md) + 10px) {}` -> no error?
- `@media (min-width: calc(theme(screens.md) + 10not-a-px) {}` -> an error?
- `@media (min-width: theme(screens.md fooFunction()) {}` -> no error?

Because we don't know what the ignored function is supposed to be we can only assume that the entire `GeneralEnclosed` becomes valid and standard CSS once the ignored function is de-sugared.

ryo-manba: Ah, I understand. Thank you for the clear explanation!

ryo-manba: @romainmenke 

> `@media (min-width: var(300px)), (min-width: theme(screens.md)) {}`
Comma separated media query lists should still report errors

Since commas split media queries, each part can be validated independently, and errors should be triggered for any invalid part. So, doesn’t the current behavior seem correct?

Or, similar to how we handle `GeneralEnclosed`, should the entire media query list be skipped if one part contains an ignored function?

romainmenke: Yeah, it's just an extra test to ensure that we are not ignoring too much if a single `GeneralEnclosed` is ignored :)

Current implementation/behavior should already be correct for that.
Just a bit of extra test coverage.

Mouvedia (changes requested):
Can you update
https://github.com/stylelint/stylelint/blob/eb20f640b5949701cbf4c963db6dd34d28e081d3/types/stylelint/index.d.ts#L474
to be
```
		'media-query-no-invalid': CoreRule<true, { ignoreFunctions: OneOrMany<StringOrRegex> }>;
```

lib/rules/media-query-no-invalid/index.mjs
```diff
@@ -143,6 +158,31 @@ const rule = (primary) => {
...
 
+/**
+ * @param {import('@csstools/media-query-list-parser').MediaQuery|import('@csstools/media-query-list-parser').GeneralEnclosed} node
+ * @param {Object|undefined} secondaryOptions
+ * @returns {boolean}
+ */
+function shouldIgnoreNode(node, secondaryOptions) {
+	let ignored = false;
+
+	if (secondaryOptions && 'ignoreFunctions' in secondaryOptions) {
```
Mouvedia, line 169: question
nit

Why not `if (secondaryOptions?.ignoreFunctions) {`?

lib/rules/media-query-no-invalid/index.mjs
```diff
@@ -143,6 +158,31 @@ const rule = (primary) => {
...
+		node.walk(({ node: childNode }) => {
+			if (!isFunctionNode(childNode)) {
+				return;
+			}
+
+			ignored = optionsMatches(secondaryOptions, 'ignoreFunctions', childNode.getName());
+
+			if (ignored) {
+				return false; // Stop iteration if an ignored function is found
```
Mouvedia, line 178: question

Can you have multiple functions per media query?
If so, can you add a test where you have an invalid function that follows an ignored function?

Just checking if you are not returning too early.
i.e. missing potential reports

lib/rules/media-query-no-invalid/index.mjs
ryo-manba: I've fixed it.
[chore: add strict type definition for SecondaryOptions](https://github.com/stylelint/stylelint/pull/8060/commits/ae03274b5b8221bf7d49fab86a95bbb5e8e6e545)

lib/rules/media-query-no-invalid/index.mjs
```diff
@@ -143,6 +158,31 @@ const rule = (primary) => {
 	};
 };
 
+/**
+ * @param {import('@csstools/media-query-list-parser').MediaQuery|import('@csstools/media-query-list-parser').GeneralEnclosed} node
+ * @param {Parameters<import('stylelint').CoreRules[ruleName]>[1]|undefined} secondaryOptions
```
Mouvedia, line 163: suggestion
non-blocking

maybe `/** @typedef {import('stylelint').CoreRules[ruleName]} Rule */` before l42
so that you can do `* @param {Parameters<Rule>[1]} [secondaryOptions]`

ryo-manba: I'm passing `secondaryOptions` to use `optionsMatches`, and I applied your suggestion to use `typedef`.

lib/rules/media-query-no-invalid/index.cjs
```diff
@@ -35,10 +37,22 @@ const meta = {
 	url: 'https://stylelint.io/user-guide/rules/media-query-no-invalid',
 };
 
+/** @typedef {import('stylelint').CoreRules[ruleName]} Rule */
 /** @type {import('stylelint').CoreRules[ruleName]} */
```
Mouvedia, line 41:
```suggestion
/** @type {Rule} */
```

Mouvedia, line 41:
```suggestion
/** @type {Rule} */
```

lib/rules/media-query-no-invalid/README.md
```diff
@@ -62,3 +62,39 @@ The following patterns are _not_ considered problems:
...
+
+Ignore the specified functions.
+
+For example, with `true`.
```
Mouvedia, line 72: This is not necessary for `true`/`null` rules.

lib/rules/media-query-no-invalid/index.mjs
```diff
@@ -143,6 +159,31 @@ const rule = (primary) => {
...
+ * @param {import('@csstools/media-query-list-parser').MediaQuery|import('@csstools/media-query-list-parser').GeneralEnclosed} node
+ * @param {Parameters<Rule>[1]} secondaryOptions
+ * @returns {boolean}
+ */
+function shouldIgnoreNode(node, secondaryOptions) {
+	let ignored = false;
+
+	if (secondaryOptions?.ignoreFunctions) {
+		node.walk(({ node: childNode }) => {
+			if (!isFunctionNode(childNode)) {
```
Mouvedia, line 172: question

If you can have multiple child nodes and non-functional nodes may be encountered, couldn't we have an ignored function following a non-functional node?
In that case, wouldn't it return too early?
i.e. before `ignored` can be set to true at the next iteration

ybiquitous:
Sorry for the late response. I'm looking at this 👀

lib/rules/media-query-no-invalid/README.md
ybiquitous: Right. `true` is obvious. We can remove this line and simplify the description. 👍🏼

lib/rules/media-query-no-invalid/README.md
```diff
@@ -62,3 +62,39 @@ The following patterns are _not_ considered problems:
...
+
+<!-- prettier-ignore -->
+```css
+@media (max-width: get-default-width()) {}
+```
+
+<!-- prettier-ignore -->
+```css
+@media (max-width: get_default_width()) {}
+```
```
ybiquitous, line 100: [suggestion] The 2 examples for `"/^get.*$/"` seem redundant. I think having only one example is enough.

Personally, `get-default-width` is preferred, but it's not a strong opinion.

lib/rules/media-query-no-invalid/index.mjs
```diff
@@ -52,16 +65,19 @@ const rule = (primary) => {
 
 			parseMediaQuery(atRule).forEach((mediaQuery) => {
 				if (isMediaQueryInvalid(mediaQuery)) {
-					// Queries that fail to parse are invalid.
-					complain(atRule, atRuleParamIndexValue, mediaQuery);
+					if (!shouldIgnoreNode(mediaQuery, secondaryOptions)) {
+						// Queries that fail to parse are invalid.
+						complain(atRule, atRuleParamIndexValue, mediaQuery);
+					}
```
ybiquitous, line 71: [suggestion] Minor. Early return can minimize the diff. How about this?

```suggestion
					if (shouldIgnoreNode(mediaQuery, secondaryOptions)) return;

					// Queries that fail to parse are invalid.
					complain(atRule, atRuleParamIndexValue, mediaQuery);
```

lib/rules/media-query-no-invalid/index.mjs
```diff
@@ -52,16 +65,19 @@ const rule = (primary) => {
...
 
 					return;
 				}
 
 				mediaQuery.walk(({ node, parent }) => {
-					// All general enclosed nodes are invalid.
```
ybiquitous, line 62: [question] Is this comment removal intentional? It seems still necessary to me... 🤔

lib/rules/media-query-no-invalid/index.mjs
```diff
@@ -52,16 +65,19 @@ const rule = (primary) => {
...
 					if (isGeneralEnclosed(node)) {
-						complain(atRule, atRuleParamIndexValue, node);
+						if (!shouldIgnoreNode(mediaQuery, secondaryOptions)) {
+							complain(atRule, atRuleParamIndexValue, node);
+						}
```
ybiquitous, line 80: [suggestion] Minor. The same idea as <https://github.com/stylelint/stylelint/pull/8060/files#r1832706629>:

```suggestion
						if (shouldIgnoreNode(mediaQuery, secondaryOptions)) return;

						complain(atRule, atRuleParamIndexValue, node);
```

lib/rules/media-query-no-invalid/index.mjs
```diff
@@ -143,6 +159,31 @@ const rule = (primary) => {
...
 
+/**
+ * @param {import('@csstools/media-query-list-parser').MediaQuery|import('@csstools/media-query-list-parser').GeneralEnclosed} node
+ * @param {Parameters<Rule>[1]} secondaryOptions
+ * @returns {boolean}
+ */
+function shouldIgnoreNode(node, secondaryOptions) {
+	let ignored = false;
+
+	if (secondaryOptions?.ignoreFunctions) {
```
ybiquitous, line 170: [suggestion] Minor. Also here, early return can reduce nesting and make the code a bit readable to me:

```js
	if (!secondaryOptions?.ignoreFunctions) return false;

	let ignored = false;

	// walk...
	
	return ignored;
```

lib/rules/media-query-no-invalid/index.mjs
Mouvedia: I couldn't, my question was probably unfounded.

---

I found something weird though; in the previous conversation it was said that if one function is ignored, everything must be ignored.
I tried

```
		{
			code: '@media (getMinWidth <= width <= foo()) {}',
		},
```
and it didn't pass with `config: [true, { ignoreFunctions: ['/^get.*$/'] }],`
is it by design?

Mouvedia: I couldn't, my question was probably unfounded.

romainmenke: Does it pass with `@media (getMinWidth() <= width <= foo()) {}`?
(I haven't tried)

Mouvedia: > (I haven't tried)

If you don't know yet, it might be useful if you plan to make a lot of reviews:

```sh
git fetch upstream pull/8060/head:add-ignore-functions-to-media-query-no-invalid && git checkout add-ignore-functions-to-media-query-no-invalid && git pull upstream pull/8060/head:add-ignore-functions-to-media-query-no-invalid
```

You gotta replace `upstream` by what you are using for the remote of `stylelint/stylelint`.

> Does it pass with `@media (getMinWidth() <= width <= foo()) {}`?

It passes.

lib/rules/media-query-no-invalid/README.md
ryo-manba: I've fixed it.
[Update README.md](https://github.com/stylelint/stylelint/pull/8060/commits/fdf7de1a74e8b46739ac864b0ce46c2098262cfd)

lib/rules/media-query-no-invalid/README.md
ryo-manba: Thank you for all the comments!
I've applied your suggestion.
[apply code review suggestion](https://github.com/stylelint/stylelint/pull/8060/commits/de4e39246897a8378945352306b0538f8afd0515)

lib/rules/media-query-no-invalid/README.md
```diff
@@ -62,3 +62,32 @@ The following patterns are _not_ considered problems:
...
+
+<!-- prettier-ignore -->
+```css
+@media (min-width: getDefaultWidth()) {}
+```
+
+<!-- prettier-ignore -->
+```css
+@media (max-width: get-default-width()) {}
+```
```
ybiquitous, line 93: [question] Are both `getDefaultWidth` and `get-default-width` still necessary? Either one seems enough to me to show an example for `/^get.*$/`. 🤔

ybiquitous (approved):
Thanks for addressing the reviews! 
I've left a minor question, but it's not a blocker. LGTM 👍🏼

lib/rules/media-query-no-invalid/README.md
ryo-manba: Sorry, I misunderstood for your [previous suggestion](https://github.com/stylelint/stylelint/pull/8060#discussion_r1832678871).
Fixed this commit.
[remove redundant examples](https://github.com/stylelint/stylelint/pull/8060/commits/c3a89d71d005f9e6652f32408654261236b2e0b0)

ybiquitous (approved):

ybiquitous:
@Mouvedia Do you have any concerns still? If no, we can merge.

lib/rules/media-query-no-invalid/__tests__/index.mjs
```diff
@@ -282,3 +282,67 @@ testRule({
...
+			code: '@media (min-width: theme(screens.md) and fooFunction(300px)) {}',
+		},
+		{
+			code: '@media (min-width: getDefaultWidth()) {}',
+		},
+		{
+			code: '@media (min-width: theme(screens.md) and var(300px)) {}',
+		},
+		{
+			code: '@media (min-width: var(300px) or theme(screens.md)) {}',
```
Mouvedia, line 304: question

@ybiquitous asked me to take a last look.
Based on:

> […such as in `@media` rules, where the `var()` function would not be valid.](https://drafts.csswg.org/css-env-1/#intro)

@romainmenke should we add a `customSyntax` or use `env()` instead?
e.g. `@custom-media --foo-bar (min-width: 300px);`

lib/rules/media-query-no-invalid/index.cjs
```diff
@@ -140,6 +158,31 @@ const rule = (primary) => {
...
+
+	node.walk(({ node: childNode }) => {
+		if (!cssParserAlgorithms.isFunctionNode(childNode)) {
+			return;
+		}
+
+		ignored = optionsMatches(secondaryOptions, 'ignoreFunctions', childNode.getName());
+
+		if (ignored) {
+			return false; // Stop iteration if an ignored function is found
```
Mouvedia, line 179: question

If it uses [`Container.walk`](https://github.com/postcss/postcss/blob/1cc6ac3bf43a12460d3396e2cb76f14075daea1b/lib/container.js#L309-L313)—which I didn't verify—wouldn't the comment be confusing and/or erroneous?

Mouvedia:
> Do you have any concerns still? If no, we can merge.

@ybiquitous Just go ahead, I don't like that it doesn't have range syntax example but that's minor.

lib/rules/media-query-no-invalid/__tests__/index.mjs
romainmenke: I am not sure what you mean :)

Given the scope of this change, can you give some example CSS input, the ignore config and what you would expect to happen?

romainmenke:
> I don't like that it doesn't have range syntax tests but that's minor.

That would not meaningfully increase test coverage.

Anything unrecognized becomes `GeneralEnclosed`.
It doesn't matter which other tokens you add to it (or remove from it).

`(min-width: foo())` is the same as `(width < foo())` and also as `(foo())` it's all a `GeneralEnclosed` containing a `foo()` function.

lib/rules/media-query-no-invalid/__tests__/index.mjs
Mouvedia: Put simply, is `var(300px)` valid in this scope?

romainmenke: From the perspective of "what is valid CSS":

No `var()` can't be used and there aren't any plans to change that.

From the perspective of this rule and this change:

You can now add config to ignore `var()`, which should silence any warnings.

[Current behavior demo]([Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIACBbGUCWBDABAChQwDsBaAdwygBcALeHANywCc8SSAzCCASh52AAdIoKoA6LhAHDRoyABsIzes3QBuGVQC+wrSAA0IDhnkwAcljQIQMAB6WADqbFgAzq4PgIRYwHNrQkQ4OII2tlQwRFCuofQA2sLBwaGuVACepvLEVCSQPhi+JKlYUSxQoYk4ALq6nnl+AGJKKFhU1gBWrt6esA4eiIHJIKkZMFlEVLEhIPKtMKmh+pUp6ZnZud5+RVQlUGVTobMRCyC1WkA))

Screenshot 2024-11-09 at 12 53 47

----

Still unsure what you mean or what your concern is :)

Mouvedia:
@romainmenke like I said, it's minor, but test cases shouldn't be strongly tied to implementation details.

lib/rules/media-query-no-invalid/__tests__/index.mjs
Mouvedia: > Still unsure what you mean or what your concern is :)

My concern is that at line 288 we have

> `config: [true, { ignoreFunctions: ['theme', 'fooFunction', '/^get.*$/'] }],`

which doesn't ignore `var` AFAIK?
Hence shouldn't l303–305 be part of `reject: []` instead?
Maybe I am missing something obvious.

romainmenke: I've already tried explaining this multiple times :)

- https://github.com/stylelint/stylelint/pull/8060#discussion_r1807781359
- https://github.com/stylelint/stylelint/pull/8060#issuecomment-2466184844
- https://github.com/stylelint/stylelint/pull/8060#discussion_r1815551527

Anything that is unrecognized becomes `GeneralEnclosed`.
`GeneralEnclosed` literally means:
- try and fail to parse a [media feature](https://drafts.csswg.org/mediaqueries-5/#typedef-media-feature)
- find the nearest balanced closing parenthesis
- capture everything in a `GeneralEnclosed` node
- continue parsing

There is no way to recover from this.

We can't ignore the function itself because we were never validating functions to begin with.

We must ignore the nearest `GeneralEnclosed`.
Ignoring the nearest `GeneralEnclosed` literally means to ignore anything in that block.

For this input: `@media (min-width: var(300px) or theme(screens.md)) {}` the nearest `GeneralEnclosed` is `(min-width: var(300px) or theme(screens.md))`.

Because the `var(300px)` is also contained within that ignored `GeneralEnclosed` it is also ignored, even though it is not covered by the config.

This is not am implementation detail as you implied in https://github.com/stylelint/stylelint/pull/8060#issuecomment-2466187361

This is how the CSS specification defines media queries must be parsed and how their error recovery works.

This Stylelint rule intends to parse as a browser would and give feedback on parse errors.

romainmenke: And to clarify why this test is correct and intended.
We want to have coverage for the behavior I just described.

This test adds that coverage.

romainmenke (approved):
@ryo-manba LGTM 👍

Thank you everyone for working on this new plugin option 🙇

PR #8061, opened 2024-10-20
Fix `selector-pseudo-class-allowed-list` reported ranges

ryo-manba:

No, it's self-explanatory.

### DIFF
--- a/lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
+++ b/lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
@@ -1,3 +1,5 @@
+import { stripIndent } from 'common-tags';
+
 import rule from '../index.mjs';
 const { messages, ruleName } = rule;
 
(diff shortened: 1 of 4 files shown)


lib/rules/selector-pseudo-class-allowed-list/index.cjs
```diff
@@ -63,12 +66,13 @@ const rule = (primary) => {
 				}
 
 				report({
-					word: value,
```
ryo-manba, line 66: See: https://github.com/stylelint/stylelint/pull/8045#discussion_r1800201009

lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
```diff
@@ -49,6 +51,16 @@ testRule({
...
 		},
+		{
+			code: stripIndent`
+				/* stylelint-disable-next-line selector-pseudo-class-allowed-list */
+				:HOVER, /* stylelint-disable-next-line selector-pseudo-class-allowed-list */
+				:HOVER,
+				/* stylelint-disable-next-line selector-pseudo-class-allowed-list */
+				:HOVER
+				{}
+			`,
```
romainmenke, line 62: Maybe better to use `:focus` in the tests.

That `:hover` and `:HOVER` are treated differently in this rule is peculiar and not what we want to test in this case is:)

romainmenke (commented):
One suggestion, otherwise looks good!

lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
ryo-manba: Ah, that sounds good. I've fixed it.

[chore: rename :HOVER to :focus](https://github.com/stylelint/stylelint/pull/8061/commits/7de08c98903b032bcfad92e5e1566b8423dab9c0)

romainmenke (approved):
LGTM!

Thank you for the speedy fix 🙇

PR #8062, opened 2024-10-20
Fix `selector-pseudo-class-disallowed-list` reported ranges

ryo-manba:

No, it's self-explanatory.


lib/rules/selector-pseudo-class-disallowed-list/index.cjs
```diff
@@ -59,12 +62,13 @@ const rule = (primary) => {
 				}
 
 				report({
-					word: value,
```
ryo-manba, line 62: See: https://github.com/stylelint/stylelint/pull/8045#discussion_r1800201009

romainmenke (approved):
LGTM 🎉

PR #8066, opened 2024-10-21
Fix `no-irregular-whitespace` reported ranges

romainmenke:

No


lib/rules/no-irregular-whitespace/index.mjs
```diff
@@ -39,16 +50,19 @@ const IRREGULAR_WHITESPACES = [
...
-	const result = IRREGULAR_WHITESPACES_PATTERN.exec(str);
-
-	return (result && result[1]) || null;
+	return Array.from(str.matchAll(IRREGULAR_WHITESPACES_PATTERN)).map((match) => {
+		return {
+			index: match.index,
+			length: match[0].length,
+		};
+	});
 };
```
romainmenke, line 66: The old implementation only reported a single irregular whitespace per AST field.
Instead of reporting each contiguous instance.

lib/utils/atRuleRawsAfterIndex.mjs
```diff
@@ -0,0 +1,7 @@
+/**
+ * @param {import('postcss').AtRule} atRule
+ * @returns {number}
+ */
+export default function atRuleRawsAfterIndex(atRule) {
+	return atRule.source?.end?.offset ?? atRule.toString().length;
```
Mouvedia, line 6: refactor
minor

This looks similar to [`return rule.source?.end?.offset ?? rule.toString().length;`](https://github.com/stylelint/stylelint/pull/8066/files#diff-15dd133c5fee2346a62b4e8f9ab0b5da1430f6008a4080cd9c885e9aa4c5fee8R6).
Could you have only one function which accept both node types?
i.e.
```
@import {AtRule, Rule} from 'postcss'
…
@param {AtRule | Rule} node
```

Mouvedia: > the actual implementation in PostCSS is very different for each node type

Interesting.

> but this would need a very specific name

If you pass a `Node` instead of `Rule | AtRule` it could be named `nodeRawsAfterIndex`.

> I am unsure.

Ill leave the decision to you.
We normally don't add that many one-line utils in a single PR so I was trying to lower that number.

romainmenke: > If you pass a `Node` instead of `Rule | AtRul`e it could be named `nodeRawsAfterIndex`.

I think this is the best option, but implies even more tests 😛 
Then we also need coverage for `Declaration`, `Root` and `Document`

I agree that too many small helpers aren't ideal.

Mouvedia: To find the name, just try to add a JSDoc @description; that will help in the process.
Else just don't refactor and resolve this thread.

romainmenke: When looking at this again I noticed that I misunderstood the `raws.after` field.

I thought this was about this:

```css
.foo {}  .bar{}
/*     ↑↑       */
```

But it is not, it is about this:


```css
.foo { color: blue;  }
/*                 ↑↑   */
```

So my previous implementation was incorrect, which also meant that we didn't have any test coverage for `raws.after` in the rule itself.

I've fixed the implementation and added more test coverage.

-----

I ultimately decided to keep these as separate functions because I am unsure if I understand it well enough at this time to pick the right abstraction.

lib/utils/ruleRawsAfterIndex.mjs
```diff
@@ -0,0 +1,13 @@
+/**
+ * @param {import('postcss').Rule} node
+ * @returns {number}
+ */
+export default function ruleRawsAfterIndex(node) {
+	// always subtract 1 for `}`
```
romainmenke, line 6: This is similar to `@media screen` where we have to do `+1` to account for the `@` when calculating any index after the name.

lib/utils/ruleRawsAfterIndex.mjs
```diff
@@ -1,7 +1,13 @@
...
-export default function ruleRawsAfterIndex(rule) {
-	return rule.source?.end?.offset ?? rule.toString().length;
+export default function ruleRawsAfterIndex(node) {
+	// always subtract 1 for `}`
+
+	if (!node.source?.end?.offset) return node.toString().length - 1;
+
+	if (!node.raws?.after?.length) return node.source.end.offset - 1;
+
+	return node.source.end.offset - (node.raws.after.length + 1);
```
Mouvedia, line 12:
```suggestion
export default function ruleRawsAfterIndex(node) {
	const endOffset = node.source?.end?.offset;
	const afterLength = node.raws?.after?.length;
	const closingParenthesisLength = 1;

	if (!offset) return node.toString().length - closingParenthesisLength;

	if (!afterLength) return endOffset - closingParenthesisLength;

	return endOffset - afterLength + 1;
```

All in all, I am not sure which one is better.
You can disregard it if you don't like it.

lib/utils/atRuleRawsAfterIndex.mjs
Mouvedia: > I ultimately decided to keep these as separate functions because I am unsure if I understand it well enough at this time to pick the right abstraction.

OK.

> I've fixed the implementation and added more test coverage.

Can you add another changeset.
This fix is not about reported ranges per se.
i.e. false negative

romainmenke: > Can you add another changeset.
> This fix is not about reported ranges per se.

It is still purely about reported ranges :)
Before this change we used `word` in report, which also gave ok results.
It just wasn't tested.

romainmenke: [Demo for `raws.after` with the current Stylelint version](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0AzCEAEwOQDYQCd4MAjXAVxgG4AdFAXxABoQUBLXGAOQEMBbOIhgAPfgAdOSMAGdpzcBAB27AOYIQwGoowYaIEQBcYiqNL3EA2lp0690gwE9OuNooMBaSMrYr39niY8BFB61hgAulqMLF6qAGKEfDwG6gBW0krysGJyiJrauiD2TjAubuaFuMkw9npMYXaOzq4esT5+BgFQQSEIldW1IFEg9EA)

lib/utils/__tests__/atRuleRawsAfterNameIndex.test.mjs
```diff
@@ -0,0 +1,28 @@
...
+
+	it('has a "supports" condition', () => {
+		expect(atRuleRawsAfterNameIndex(atRule('@supports (color) {}'))).toBe(9);
+	});
+
+	it('has a space before the @', () => {
+		expect(atRuleRawsAfterNameIndex(atRule(' @media (color) {}'))).toBe(6);
+	});
+
+	it('has a newline before the param', () => {
```
Mouvedia, line 17: nit

If `@media` is not "param", can you change the description?

lib/rules/no-irregular-whitespace/index.mjs
```diff
@@ -61,48 +75,65 @@ const rule = (primary) => {
...
 
+function zeroIndex() {
+	return 0;
+}
+
+function oneIndex() {
+	return 1;
+}
+
```
Mouvedia, line 136:
```suggestion

/**
 * @param {import('postcss').Rule} rule
 * @returns {number}
 */
function ruleRawsBetweenIndex(rule) {
	return rule.selector.length;
}

```

What about putting it here instead?
It's unlikely it will be reused and even if it was it's too simple to warrant being an utility IMHO.

lib/utils/__tests__/atRuleRawsAfterNameIndex.test.mjs
romainmenke: Good catch, thank you

lib/utils/atRuleParamIndex.mjs
```diff
@@ -1,10 +1,11 @@
+import atRuleRawsAfterNameIndex from './atRuleRawsAfterNameIndex.mjs';
+
 /**
  * @param {import('postcss').AtRule} atRule
  * @returns {number}
  */
 export default function atRuleParamIndex(atRule) {
```
Mouvedia, line 7: minor
question

`atRuleParamIndex` is used by 14 other rules.
I suppose this modification fixes a bug hence the changeset could have 14 rules except `selector-pseudo-class-no-unknown` which already has its own changeset in this release.

lib/utils/atRuleParamIndex.mjs
Mouvedia, line 1: I am trying to lower the number of util that we won't reuse.

What about having

```js
/**
 * @param {import('postcss').AtRule} atRule
 * @returns {number}
 */
export const atRuleRawsAfterNameIndex = function (atRule) {
	// Initial 1 is for the `@`
	return 1 + atRule.name.length;
}
```

in this file instead?

romainmenke: Given that we already have `atRuleParamIndex`, `declarationValueIndex` I thought I was following existing best practices and patterns by adding these utilities.

We have seen so many bugs around the reported ranges and having these utilities has significant benefits:
- shared code
- documented complexity
- unit tests

I'd rather have clear utilities that are well tested and documented, each in their own file.

With that in mind I have also reverted the precies commit where I inlined one of the utilities.

romainmenke: I've filed a separate issue to discuss the structure of the codebase separate from this specific change: https://github.com/stylelint/stylelint/issues/8097

ybiquitous (commented):
[suggestion] Can you extract a few pull requests from this one to make reviewing/reverting easier?

For example, one PR for the declaration utils is fine to me.

---

For the declaration utils, does it make sense to combine them to one function? I imagine the interface would be:

```js
/**
 * @param {'between' | 'value'} component
 * @returns {number}
 */
export function declarationComponentIndex(decl, component) {
  // ...
}
```

I'd like to reduce similar files/codes for maintainability even they're internal.

Mouvedia:
> I'd like to reduce similar files/codes for maintainability even if they're internal.

@ybiquitous see #8097.

romainmenke:
This should be ready for another look, thank you for the feedback

lib/rules/no-irregular-whitespace/__tests__/index.mjs
```diff
@@ -113,6 +149,33 @@ testRule({
...
+			description: 'irregular whitespace in the after of an at rule',
+			message: messages.unexpected,
+			line: 1,
+			column: 16,
+			endLine: 1,
+			endColumn: 17,
+		},
+		{
+			code: '.a {\f}',
+			description: 'irregular whitespace in the after of an at rule',
```
Mouvedia, line 163: typo
minor

this is not an at rule

lib/rules/no-irregular-whitespace/__tests__/index.mjs
```diff
@@ -113,6 +149,33 @@ testRule({
...
+			message: messages.unexpected,
+			line: 1,
+			column: 5,
+			endLine: 1,
+			endColumn: 6,
+		},
+		{
+			code: '.a {\f\f}',
+			description: 'irregular whitespace in the after of an at rule',
```
Mouvedia, line 172: ditto at rule
whitespace > whitespaces

lib/rules/no-irregular-whitespace/index.cjs
```diff
@@ -65,48 +72,65 @@ const rule = (primary) => {
...
 		root.walkDecls((decl) => {
-			validate(decl, decl.prop);
-			validate(decl, decl.value);
-			validate(decl, decl.raws.before);
-			validate(decl, decl.raws.between);
+			validate(decl, decl.raws.before, zeroIndex);
+			validate(decl, decl.prop, zeroIndex);
+			validate(decl, decl.raws.between, nodeFieldIndices.declarationBetweenIndex);
+			validate(decl, getDeclarationValue(decl), nodeFieldIndices.declarationValueIndex);
 		});
```
Mouvedia, line 122: I suppose `decl.raws.after` doesn't exist.

Mouvedia (commented):
Iv left some additional comments; nothing important though.

lib/rules/no-irregular-whitespace/index.cjs
romainmenke: Correct, `after` does not come after the rule/atrule, it is what comes before the closing `}`:

```
.foo { bar: 1;  }
/*            ^^ this is the after of the rule
```

```
.foo { bar: 1  ; }
/*           ^^ this is part of `raws.values.raw`, `1  `
```

Declarations have no closing `}`

lib/rules/no-irregular-whitespace/__tests__/index.mjs
romainmenke: `whitespace` is both plural and singular.
`whitespaces` isn't actually correct spelling, while `whitespace characters` is correct if you want to be explicit about the plural :)

I think `whitespace` is fine in this case.

ybiquitous (approved):
Thank you for your effort. The code is now much cleaner while fixing the bug! LGTM 😄 👍🏼

PR #8067, opened 2024-10-22
Fix `selector-disallowed-list` reported ranges

ryo-manba:

No, it's self-explanatory.


lib/rules/selector-disallowed-list/__tests__/index.mjs
```diff
@@ -1,6 +1,7 @@
-import rule from '../index.mjs';
 import { stripIndent } from 'common-tags';
 
+import rule from '../index.mjs';
+
```
ryo-manba, line 4: See: https://github.com/stylelint/stylelint/pull/8037#discussion_r1800724431

romainmenke (approved):
Thank you. LGTM 👍

PR #8072, opened 2024-10-23
Fix `selector-nested-pattern` reported ranges

ryo-manba:

No, it's self-explanatory.


lib/rules/selector-nested-pattern/index.mjs
```diff
@@ -49,21 +52,41 @@ const rule = (primary, secondaryOptions) => {
...
+						return;
+					}
 
-				report({
-					result,
-					ruleName,
-					message: messages.expected,
-					messageArgs: [selector, primary],
-					node: ruleNode,
-					word: selector,
```
ryo-manba, line 65: See: https://github.com/stylelint/stylelint/pull/8045#discussion_r1800201009

lib/rules/selector-nested-pattern/index.cjs
```diff
@@ -53,21 +56,41 @@ const rule = (primary, secondaryOptions) => {
...
+						messageArgs: [selectorStr, primary],
+						node: ruleNode,
+						index,
+						endIndex,
+					});
 				});
+			} else {
+				const selectors = [ruleNode.selector];
+
+				for (const selector of selectors) {
```
romainmenke, line 80: Now that there is differentiation between the two modes I think it is fine to optimize this is a bit.

Can you remove the for loop?

romainmenke (commented):
Thank you @ryo-manba,
I've left a suggestion, can you take a look?

lib/rules/selector-nested-pattern/index.cjs
ryo-manba: Ah, you're right, good point. I've fixed it!
[chore: remove unnecessary for loop](https://github.com/stylelint/stylelint/pull/8072/commits/87bc953002cd75b0f16117f9d1fae8eed6c02a02)

lib/rules/selector-nested-pattern/index.mjs
```diff
@@ -49,20 +52,36 @@ const rule = (primary, secondaryOptions) => {
 				return;
 			}
 
-			const selectors = splitList ? ruleNode.selectors : [ruleNode.selector];
+			if (splitList) {
+				parseSelector(getRuleSelector(ruleNode), result, ruleNode)?.each((selector) => {
```
romainmenke, line 56:
```suggestion
			const selectors = getRuleSelector(ruleNode);

			if (splitList) {
				parseSelector(selectors, result, ruleNode)?.each((selector) => {
```

How about storing `getRuleSelector(ruleNode)` into a variable and re-using that in both branches?

```js
			} else {
				if (normalizedPattern.test(selectors)) {
					return;
				}

				report({
					result,
					ruleName,
					message: messages.expected,
					messageArgs: [selectors, primary],
					node: ruleNode,
					index: 0,
					endIndex: selectors.length
				});
			}
```

lib/rules/selector-nested-pattern/__tests__/index.mjs
romainmenke, line 1: Maybe also some tests for the default config?

± ln39

```js
		{
			code: stripIndent`
				a {
					/* stylelint-disable-next-line selector-nested-pattern */
					b /* stylelint-disable-next-line selector-nested-pattern */
					b
					/* stylelint-disable-next-line selector-nested-pattern */
					b
					{}
				}
			`,
		},
```

± ln89

```js
		{
			code: 'a { b /* a comment */ b {} }',
			warnings: [
				{
					message: messages.expected('b /* a comment */ b', /^[A-Z]+$/),
					line: 1,
					column: 5,
					endLine: 1,
					endColumn: 24,
				},
			]
		},
```

romainmenke (commented):
Thank you for applying the feedback @ryo-manba 

I've left one more suggestion but looks good overall

lib/rules/selector-nested-pattern/index.mjs
romainmenke: You could also add a `check()` function to reduce duplication.

Something like :

```js
/**
 * @param {string} selectorStr
 * @param {number} index
 * @param {number} endIndex
 */
function check(selectorStr, index, endIndex) {
	if (normalizedPattern.test(selectorStr)) return;

	report({
		result,
		ruleName,
		message: messages.expected,
		node: ruleNode,
		index,
		endIndex,
	});
}
```

ryo-manba: Thanks for the feedback! I applied your suggestion.
[apply suggestions from code review](https://github.com/stylelint/stylelint/pull/8072/commits/6076e57cea3f061cf86868cf8f9a35284d34ea99)

romainmenke (approved):
Thank you. LGTM 👍

PR #8073, opened 2024-10-23
Fix `selector-no-vendor-prefix` reported ranges

ryo-manba:

No, it's self-explanatory.


lib/rules/selector-no-vendor-prefix/index.mjs
```diff
@@ -68,13 +67,17 @@ const rule = (primary, secondaryOptions) => {
...
+				const index = pseudoNode.sourceIndex;
+				const endIndex = index + value.length;
+
 				report({
 					result,
 					ruleName,
 					message: messages.rejected,
 					messageArgs: [value],
 					node: ruleNode,
-					word: value,
```
ryo-manba, line 77: See: https://github.com/stylelint/stylelint/pull/8045#discussion_r1800201009

Issue #7908, opened 2024-08-08
Fix autofix removing comments


romainmenke:
### What minimal example or steps are needed to reproduce the bug?

```css
:not(
	a,
	/* stylelint-disable-next-line */
	DIV
) {}

:not(
	a,
	/* foo */
	DIV
) {}
```

### What minimal configuration is needed to reproduce the bug?

```json
{
  "rules": {
    "selector-type-case": "lower",
    "selector-not-notation": "simple"
  }
}
```

### How did you run Stylelint?

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcCB2EAuAKAOkzqCGANNrgPQBUABAM6oCeANjPQJZKoC0UzVeARo+yQwAHhxZCKZEsVQARAJIA1bAEoKwAL7Ek8FBhmEZ5CgDMIESdJxylq9RpAEQJ5owByeALZxEIrwAdGADowKipHcAgkFwBzBBBgbAoKTBAAJwBXRipU+HUk5JSQKiYYMFQINPY6fxh2MDwS3KL6CAB3GDTUoiRCopLGcsrBNBH8VGYo5tSqZk9AmFSCrSQHJ0ho5hiAMUrPPFR4gCsqKIjYf3DERN7+ukZxVGmQegOYGm6CmfumVg4N2LsGh4JBQPBpKDPV6od5PEDYBwaIA)

### Which Stylelint-related dependencies are you using?

```json
{
  "stylelint": "latest",
  "stylelint-config-standard": "latest"
}
```

### What did you expect to happen?

I expected the comments to be preserved and the final output to be:

```css
:not(a):not(/* stylelint-disable-next-line */ DIV) {}

:not(a):not(/* foo */ DIV) {}
```

### What actually happened?

Any comments are removed

```css
:not(a):not(DIV) {}

:not(a):not(DIV) {}
```

### Do you have a proposal to fix the bug?

Make sure that all rules read the version of the AST that contains comments.

i.e. `rule.raws.selector.raw` over. `rule.selector`

Or to use the correct helper when this exists, i.e. `getRuleSelector`

romainmenke:
@Mouvedia Can you check that I accurately described the issue?

romainmenke:
This issue is not specific to values and selectors.
It can also be reproduced with rules and declarations:

[...]

PR #8074, opened 2024-10-23
Fix `selector-not-notation` reported ranges

ryo-manba:

No, it's self-explanatory.


lib/rules/selector-not-notation/__tests__/index.mjs
```diff
@@ -140,6 +152,40 @@ testRule({
...
+		{
+			code: stripIndent`
+				/* a comment */
+				:not(::before), /* a comment */
+				:not(::before),
+				/* a comment */
+				:not(::before)
+				{}
+			`,
+			unfixable: true,
```
romainmenke, line 164: Maybe another test for a case that is fixable?
Good to verify that we don't erase or mangle the comments.

Mouvedia: That would be covered by #7908.
see https://github.com/stylelint/stylelint/issues/7908#issuecomment-2287136983

ryo-manba: As @Mouvedia mentioned, this issue involves a different problem. 
Therefore, we can't add the test at this time.

See: [demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcID0AqABAQzZAtjmAdgC5opIA6B8BERAFBgDQCUjaqm2EehJZl1WgxaNKHLLnzFSFKjXpNmaSsAC+lEIxAAzAJYAbGADkM+BCBgAPUwAdDAOjABnJ5vAQCegObnglNMogAE4AroZO5AhofgQBAZFOMIZgRBBBALTymbQYRLoekfCBTro4djCR-mjqBKpukJ66XgBiaTi55gBWTh5usDauiDHxIE5EAJ6G+rrEhYH6uTBjkaKxxRNTM0TpDd7pYxgEUBhBUHORC0RLRJW1IKpAA)

romainmenke: No, this is directly related to this issue.

On line `78` you need to change:

```diff
- const selector = ruleNode.selector;
+ const selector = getRuleSelector(ruleNode);
```

Some examples of issues that currently exist:
- https://stylelint.io/demo/#N4Igxg9gJgpiBcID0AqABAZwC4E8A2MeAlgHZYC0URGAhgEYHkkwAeFxzmhMYWEATkwgUSwmliIQSaFEgA6JeKKwAKGgBoAlOrSo0NNJAC2RmGRnzFytVvUK9B46fOyFS4Tc1oFwAL4KQdRAAMyICADkaUwQQViiABwIAOjAMDEDwKVCAcxjgBTRvEH4AVwIMOQQ0fOlCooxuXgEhETEJKUr4eqIjRJhKgrR-El8MyBIcgDEBI3EYgCsMKQzYePTEGsLK7HxCUixOorxxGGxKu1rt3AIOCnGc8mwaEigafihDyuOsU4OQBVGviAA
- https://stylelint.io/demo/#N4Igxg9gJgpiBcCB2EAuAKAhgGgJTYHoAqAAgGdUBPAGxmoEslUBaKeszAI1uaRgA8WDPuTowwqCACdeaWakyp6EJCSIEAOkngoMOfFuIlMJSAFszMJms3bdWPCS3AAvlpDYQAM3q0AcpiWCCACgQAOtAB0YGRkHuAqPgDmwcBaJE4gUgCutGQaCCRpqhmZZGIS0vLyispIBfBl9GYRMAXpJG5ILvGQSMkAYtJmisEAVmQq8bBhcYjFGQUUNHSMqA2Z1IowFAXYHUtUtMIsfcnMFJhIUJhSUBsFW6g76yBaPS5AA

ryo-manba: Ah, I see. Sharp insight! I've applied the fix.

[Fix selector-not-notation autofix](https://github.com/stylelint/stylelint/pull/8074/commits/79ab762cde48f867928ac2cb9d4a2087f8d58746)

lib/rules/selector-not-notation/__tests__/index.mjs
```diff
@@ -140,6 +152,81 @@ testRule({
...
+				:not(a, ), /* stylelint-disable-next-line selector-not-notation */
+				:not(b, ),
+				/* a comment */
+				:not(c, )
+				{}
+			`,
+			fixed: stripIndent`
+				/* a comment */
+				:not(a), /* stylelint-disable-next-line selector-not-notation */
+				:not(b,),
```
ryo-manba, line 208: It removes spaces even for ignore cases. 
`selectorParser` appears to strip them out, even with [`{lossless: true}`](https://github.com/postcss/postcss-selector-parser/blob/6158750aab0aed3046a5920bd58e0e0266a4ada2/API.md#processoroptions) set.
https://github.com/stylelint/stylelint/blob/eb20f640b5949701cbf4c963db6dd34d28e081d3/lib/utils/parseSelector.mjs#L20

lib/rules/selector-not-notation/__tests__/index.mjs
ryo-manba: Fixed. I've also left one more comment regarding a concern.

[Fix to ensure stylelint-disable-next-line works correctly](https://github.com/stylelint/stylelint/pull/8074/commits/fa9e135936ad886f41fd6dd58c50ddc42d7bb35f)

lib/rules/selector-not-notation/index.mjs
```diff
@@ -113,6 +114,7 @@ const rule = (primary) => {
 					node: ruleNode,
 					result,
 					ruleName,
+					line: range.start.line,
```
Mouvedia, line 117: The current order in `report.mjs` is

```js
	const range = node?.rangeBy({ index, endIndex }) ?? {};
	const startLine = line ?? range.start?.line;
```

we would need to change it to

```js
	const startLine = line ? line : node?.rangeBy({ index, endIndex })?.start?.line;
```

for it to matter.

ryo-manba: That's a good point. Fixed it.
[chore: call rangeBy only when needed](https://github.com/stylelint/stylelint/pull/8074/commits/21f1c034192812db511580f94c20d2d1e2dea59b)

Mouvedia: That was a just a remark; I should have said so, my bad.

Since it's a refactor, it's all good though.

lib/utils/report.mjs
```diff
@@ -47,10 +47,9 @@ export default function report(problem) {
 		);
 	}
 
-	const range = node?.rangeBy({ index, endIndex }) ?? {};
 	// If a line is not passed, use the node.rangeBy method to get the
 	// line number that the complaint pertains to
-	const startLine = line ?? range.start?.line;
+	const startLine = line ?? node?.rangeBy({ index, endIndex })?.start?.line;
```
Mouvedia, line 52: If `rangeBy` returns an object then `start` is always present hence one `?` could be dropped.
i.e. `node?.rangeBy({ index, endIndex })?.start.line`;

We could also drop the first `?` but let's not do that in that PR.
i.e. `node` is not optional

Mouvedia, line 52: `rangeBy` returns an object which always contains `start` hence some `?` could be dropped.
i.e. `node?.rangeBy({ index, endIndex }).start.line`;

We could also drop the first `?` but let's not do that in that PR.
i.e. `node` is not optional

lib/utils/report.mjs
romainmenke, line 1: Can we undo all changes in this file?

I'd like to keep refactors to `report` separate from changes to individual rules when possible.

romainmenke (changes requested):

lib/utils/report.mjs
Mouvedia: Well then let's not pass `line` to report then.
i.e. it doesn't make sense without the change to `report.mjs`

ryo-manba: Okay, I've reverted it.
[Revert "chore: call rangeBy only when needed"](https://github.com/stylelint/stylelint/pull/8074/commits/baf5b11a3a1536be5b3915e372138c35f17fb9b6)

ryo-manba: I created a separate PR.
https://github.com/stylelint/stylelint/pull/8083

romainmenke: Thank you @ryo-manba

lib/rules/selector-not-notation/__tests__/index.mjs
romainmenke: This is interesting!

Please consider opening an upstream issue about this in https://github.com/postcss/postcss-selector-parser if you have time.

lib/rules/selector-not-notation/__tests__/index.mjs
```diff
@@ -140,6 +152,81 @@ testRule({
...
 			endColumn: 12,
 		},
+		{
+			code: stripIndent`
+				/* a comment */
+				:not(a, ), /* a comment */
+				:not(a, ),
+				/* a comment */
+				:not(a, )
+				{}
```
romainmenke, line 162: I think the `:not(a, )` test cases are extremely exotic.
Can we rewrite the new tests to use more common patterns?

Maybe something like `:not(a, div)`?

lib/rules/selector-not-notation/index.mjs
romainmenke, line 117: Blocked until we resolve: https://github.com/stylelint/stylelint/pull/8086

-----

You shouldn't need to pass `line` to get the expected result.

Can you remove this after we resolve #8086 ?

-----

I also noticed that this rule uses a bit of custom code to calculate a range.

I don't think this is actually needed.
Bringing this rule in line with others by calculating an `index` and `endIndex` might be better.

This would also allow us to remove this custom code in this rule.

romainmenke (commented):
Thank you for these changes @ryo-manba,

While reviewing the changes and test cases you added I encountered a couple issues in other parts of Stylelint.

I've opened https://github.com/stylelint/stylelint/pull/8086 to address those.

Ideally we resolve `8086` first and then review and merge this change.

------

I've also left a few more suggestions.
Please take a look when you have time.

lib/rules/selector-not-notation/index.mjs
Mouvedia: > I also noticed that this rule uses a bit of custom code to calculate a range.
> I don't think this is actually needed.

It is. Ill explain when we will get there.
In short it's related to cloning and the loss of information.

romainmenke: I tried with `index` and `endIndex` while workin on `8086` and that just worked 🤔
If there is complexity maybe it is better as a separate change. I don't want to derail @ryo-manba's work here :)

lib/rules/selector-not-notation/__tests__/index.mjs
ryo-manba: Just opened an upstream issue.
https://github.com/postcss/postcss-selector-parser/issues/298

lib/rules/selector-not-notation/index.mjs
romainmenke: We still can't remove the `line` argument because of : https://github.com/stylelint/stylelint/issues/8126

I tested `index` and `endIndex` again and this just works fine. (ignoring upstream bugs around serialization of `:not(a, )`)
Using `index` and `endIndex` allows for the removal of the `line` argument.

I'm focussing on this extra `line` argument because it reveals obvious bugs in our code for `report()`. The code to calculate the ranges is already unique to this rule and adding even more code to cover other bugs seems like an incorrect path to me.

For now however I think we can keep the `line` argument while we resolve surrounding issues?

lib/rules/selector-not-notation/__tests__/index.mjs
ryo-manba: Thank you for the clarifications. I've removed in [this commit](https://github.com/stylelint/stylelint/pull/8074/commits/518bd98461e90f990ed85e40afe001efd4e3ff33).

romainmenke: Thank you for these further edits 🙇

PR #8081, opened 2024-10-25
Document `scope` in the context of configuration comments

### LINKED ISSUE #8077: Document `scope` in the context of configuration comments

see: https://stylelint.io/user-guide/cli#--report-invalid-scope-disables---risd

> `--report-invalid-scope-disables, --risd`
> Produce a report of the `stylelint-disable` comments that used for rules that don't exist within the configuration object. More info.

This comment doesn't explain what `scope` means and leaves it up to the reader to deduce that it means "a comment controlling behavior for a specific rule".

----

see: https://stylelint.io/user-guide/ignore-code

> Or you can turn off individual rules:
> 
> ```css
> /* stylelint-disable selector-max-id, declaration-no-important */
> #id {
>   color: pink !important;
> }
> /* stylelint-enable selector-max-id, declaration-no-important */
> ```

I think we should explain this a bit more explicitly.

Maybe something like this:

> Or you can **scope configuration comments to** turn off individual rules:
> 
> ```css
> /* stylelint-disable selector-max-id, declaration-no-important */
> #id {
>   color: pink !important;
> }
> /* stylelint-enable selector-max-id, declaration-no-important */
> ```

(new text in bold)

ryo-manba:

Closes #8077 

No, it's self-explanatory.

### DIFF
--- a/docs/user-guide/ignore-code.md
+++ b/docs/user-guide/ignore-code.md
@@ -9,7 +9,7 @@ You can ignore:
 
 You can temporarily turn off rules using configuration comments in your CSS.
 
-For example, you can either turn all the rules off:
+For example, you can either turn all the rules off with an unscoped disable comment:
 
 <!-- prettier-ignore -->
 ```css
@@ -18,7 +18,7 @@ a {}
 /* stylelint-enable */
 ```
 
-Or you can turn off individual rules:
+Or you can turn off individual rules with a scoped disable comment:
 
 <!-- prettier-ignore -->
 ```css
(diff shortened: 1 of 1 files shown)


docs/user-guide/ignore-code.md
```diff
@@ -63,7 +63,7 @@ Stylelint supports complex, overlapping disabling & enabling patterns:
 ```
 
 > [!WARNING]
-> Configuration commands in non-standard syntax comments (e.g. `// stylelint-disable`) are only minimally supported.
```
Mouvedia, line 66: > The rules regarding line breaks in quotes is the same as outside of quotes. If you want to have a single newline, you need to either add two spaces after the preceding line, or add a <br> . 

This change seems out of scope.

ryo-manba: It seems that auto-formatting was applied. Thanks for checking!

[chore: revert spaces](https://github.com/stylelint/stylelint/pull/8081/commits/170f8e3ab723b1cf5e7a93fac0f5b26127638fdb)

romainmenke (approved):
Thank you. LGTM 👍

PR #8009, opened 2024-09-25
Add `--[no-]validate` flag and `validate` option

Mouvedia:
> Which issue, if any, is this issue related to?

It was inspired by:

> […by it's very nature Stylelint runs should gravitate towards becoming a `noop`.](https://github.com/stylelint/stylelint/issues/6869#issue-1729397616)

> Is there anything in the PR that needs further explanation?

1.

Most of the time your config is valid. Hence we ought to provide a way to turn `validateOptions` into a noop.
i.e. if you know what you are doing then skipping the validation of the options should be possible
e.g. faster CI, vscode extension

The default is `true`.
i.e. it's not a disruptive change

2.

My very basic/crude perf test compared `main` and `flag-option-validate` using time and [bulma](https://cdn.jsdelivr.net/npm/bulma@1.0.2/css/bulma.min.css):

```console
$ time ./bin/stylelint.mjs bulma.min.css -f compact
```

```json
{
  "extends": [
    "stylelint-config-standard"
  ],
  "validate": false
}
```
After removing the lower and upper outliers, I was getting between 50 and 500 ms of gains on my old macbook air.
Thorough testing is not needed because the outcome should be obvious.

3.

Caching is another alternative.
The flag/option is easier to introduce.
i.e. overloading `cache` is riskier

4.

only review the second commit

PR #8085, opened 2024-10-27
Do not assume `stylelint` presence

### LINKED ISSUE #8044: Fix unexpected `cannot read properties of undefined` error when reading `config`
> We are using stylelint with our vite project — and something in 16.10.0 is causing errors for us: "Internal server error: Cannot read properties of undefined (reading 'config')"
>
>Current setup is that we have a .stylelintrc.yml that extends our global .stylelintrc.json; but we keep getting this error even though 16.9.0 works just fine! This is the only part of the changelog referencing a config > so starting here.

_Originally posted by @michwads in https://github.com/stylelint/stylelint/issues/8030#issuecomment-2411951545_

Mouvedia:

Related to #8044

https://github.com/stylelint/stylelint/pull/7728/files#diff-cba25f954347e963307879ca83b95b319e9003d9b4b51fd350d3f4110566d3a8L21-L25
https://github.com/stylelint/stylelint/pull/8009/files#diff-a9078006ea40b10e08ed8dd13f68bb5a9585641b76e827cef5bc80216d1aa877R18

https://github.com/stylelint/stylelint/blob/b6e6a719e0e9924b60ca2b7fc1b989ab3e2f8a07/types/stylelint/index.d.ts#L186

The previous code was assuming everyone is respecting the JSDoc comments.
Since even our own script wasn't updated, it doesn't hurt to add one `?`.

### DIFF
--- a/scripts/benchmark-rule.mjs
+++ b/scripts/benchmark-rule.mjs
@@ -64,8 +64,19 @@ const [primary, secondary] = ruleSettings;
 const context = ruleContext ? JSON.parse(ruleContext) : {};
 
 const rule = (await rules[ruleName])(primary, secondary, context);
+const fn = (root, result) => {
+	result.stylelint ||= {
+		ruleSeverities: {},
+		customMessages: {},
+		customUrls: {},
+		fixersData: {},
+		ruleMetadata: {},
+		disabledRanges: {},
+	};
 
-const processor = postcss().use(rule);
+	rule(root, result);
+};
+const processor = postcss().use(fn);
 
 // eslint-disable-next-line n/no-unsupported-features/node-builtins -- This script is only for development. We can tolerate it.
 fetch(CSS_URL)
--- a/lib/utils/validateOptions.cjs
+++ b/lib/utils/validateOptions.cjs
@@ -20,7 +20,7 @@ const REPORT_OPTIONS = new Set([
  * @type {import('stylelint').Utils['validateOptions']}
  */
 function validateOptions(result, ruleName, ...optionDescriptions) {
-	const mustValidate = result.stylelint.config?.validate;
+	const mustValidate = result.stylelint?.config?.validate;
 
 	if (!mustValidate) return true;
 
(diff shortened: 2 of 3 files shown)


scripts/benchmark-rule.mjs
```diff
@@ -64,8 +64,19 @@ const [primary, secondary] = ruleSettings;
...
-const processor = postcss().use(rule);
+const fn = (root, result) => {
+	result.stylelint ||= {
+		ruleSeverities: {},
+		customMessages: {},
+		customUrls: {},
+		fixersData: {},
+		ruleMetadata: {},
+		disabledRanges: {},
+	};
```
Mouvedia, line 75: @ybiquitous I chose not to revert your changes to the `report` utility.
i.e. it could also be done directly in `report` but that couldn't be justified by the current type of `PostcssResult['stylelint']`

lib/standalone.cjs
```diff
@@ -60,7 +60,7 @@ async function standalone({
 	reportInvalidScopeDisables,
 	reportNeedlessDisables,
 	reportUnscopedDisables,
-	validate = true,
+	validate = config?.validate ?? true,
```
Mouvedia, line 63: remark

the config has the priority

ybiquitous:
> Adds missing documentation.

What is this change for? If it's not for #8044, can you please separate this PR?

scripts/benchmark-rule.mjs
ybiquitous: I think we should modify this benchmark script by using the public API, such as `utils.checkAgainstRule`, instead of increasing copy-and-paste.

Mouvedia:
> If it's not for #8044, can you please separate this PR?

#8087

ybiquitous:
FYI, I've updated the PR description not to close #8044.

```diff
-Closes #8044
+Related to #8044
```

scripts/benchmark-rule.mjs
ybiquitous: I just tried another approach with PR #8090, but this is quite slow (about 2x slower in my machine).

Mouvedia:
@ybiquitous Ill remove

```diff
-  const mustValidate = result.stylelint.config?.validate;
+  const mustValidate = result.stylelint?.config?.validate;
```
and change the title if we don't get an MRE.

ybiquitous:
@Mouvedia I just merged PR #8090. Would you mind closing this PR? I think this PR is no longer necessary.

PR #8083, opened 2024-10-26
Refactor `report` to call `rangeBy` only when needed

ryo-manba:
> Which issue, if any, is this issue related to?

None, this PR addresses feedback from https://github.com/stylelint/stylelint/pull/8074#discussion_r1817684975.

> Is there anything in the PR that needs further explanation?

No, it's self-explanatory.

Issue #8050, opened 2024-10-15
Deprecate `word` in `report()`


romainmenke:
Should we deprecate `word`?

https://github.com/stylelint/stylelint/blob/31c24c4f017892f4348872625725a3db4661af77/lib/utils/report.mjs#L27

We are moving away from it in function of various changes around fixers, reported ranges, ...

The main issue with using `word` is that a single word can appear multiple times in a given bit of CSS text.

-------

I like the ease of use of the `word` API as it allows plugin authors to quickly create a rule that works well enough in most cases.

But I am unsure if we should have an API that is known to have such issues

Mouvedia:
I recommend to start with reviewing our usage of `word`.
It is currently being used by 35 core rules.
e.g. check if we have cases where always picking the first match induces bugs

If we have bugs then it will warrant deprecating `word` in the next major.
i.e. reviewing our rules is way more valuable for our users than deprecating `word`

romainmenke:
We should first decide if we want to consider deprecating `word` at all :)

For the core rules we are consistently moving away from using `word` and each time we do, this fixes bugs.

But despite that it might still be a valuable API to offer rule authors?
I find it hard to balance the easy of use of this API against the chance of encountering real issues because of it.

Core rules are widely used and should be held to a very high standard.
But some rule authors might not need such accuracy?

----

> check if we have cases where always picking the first match induces bugs

The only scenario where `word` might be acceptable is single word constructs:
- a name of an `at-rule` e.g. `media` for `@media screen {}`
- a property e.g. `color` for `.foo { color: red; }`

Any other case will result in bugs.

But these bugs are of a very specific nature.
- all report will be valid
- all report will point to a relevant source location
- fixing errors one by one will eventually reveal all errors at their correct source position

The only issue is that initially multiple errors will be reported at the first instance in source.

-------

So to summarize
- `word` is fairly handy when writing rules
- `word` can produce reports on incorrect source locations
- end users can easily work out where all issues really are by resolving issues one by one

this is about technical purity vs. ease of use

ybiquitous:
@romainmenke Thanks for raising this up and for the good summary. 👍🏼 

I really understand your motivation. And we should avoid `word` in core rules as much as possible. But to be honest, deprecating `word` might not be accepted by light rule authors because using `word` makes it very easy to write a custom rule in private projects. `word` enables users to be free from index calculation.

FYI, PostCSS also supports `word`:
https://postcss.org/api/#warningoptions

[...]

PR #8086, opened 2024-10-27
Add test cases for report ranges following fixes

romainmenke:

See: https://github.com/stylelint/stylelint/pull/8074/files#diff-bd94fe99a5bd46c7e6b11e8afd5f71ca4aa1897e2d52801cc1e7e3be5de7d925R117
Supersedes: https://github.com/stylelint/stylelint/pull/8083

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0BDABMAOgO0wF0gBsIAneNHffAYgDN7K98B6AKjQGc8BPQmQgJbY8AWigCOKAEZ8R2GAA9Rg+WlbNGNerU146DXHgDcOAL6VsSKRkZFS5XW048+K0eMkyYcxcqEw1DUMtAyo9bUYTbFMQABoQWgE+ADkUAFs4REV0gAc+JDAODjjwCGxEgHMEECxsNAoQXxhsKA5MBDQAbRx6+vauXn4hUUhygQqRLhQWlBIodp60AF1YxfaSAFc+No7a3oa7EhEAC0URPmwKvGP2+HbiS4W6tHNoktHKgDFSNJQ8aoAVhwyiVYDliog9g0Bq5hrcGoQ-jAuO1Vs9+i4hsIRB9xpM8NMoLN5h17kiUSAzCBTEA)

Notice how the fix is applied on the incorrect line.

Screenshot 2024-10-30 at 11 07 28

At first I didn't quite understand why `line` was being added to `report()` in the change and what issue that was expected to solve.

As I understood it before this shouldn't matter.
My expectation was that the `node.rangeBy` call would return the same line as was being passed as an argument.

After a lot of digging I found that PostCSS doesn't calculate the source location relative to the original input CSS. [It calculates it based on the original input **and** the current string representation of the current node](https://github.com/postcss/postcss/blob/1cc6ac3bf43a12460d3396e2cb76f14075daea1b/lib/node.js#L190).

That however is not what we want.
The current string representation might have already been altered by rule fixers, causing unexpected offsets.

If a fixer causes a node to become shorter or longer the offsets of subsequent warnings will be incorrect.

They might become sufficiently incorrect to flip the result of `isDisabled` checks.

-------

To resolve this issue I've inlined and modified `Node.prototype.rangeBy` and `Node.prototype.positionInside` from PostCSS.

Instead of using the current string representation of the given node we always use the original input CSS.

Some of these fields are optional and might be `undefined`/`null`.
If that is the case we do a fallback to `node.rangeBy`.
This will only happen for constructed nodes that lack a `source` object.

-----

A side effect of this change should be that making reports will a bit faster as calling `toString()` on nodes is slower than counting newlines.


lib/utils/report.mjs
```diff
@@ -94,13 +97,21 @@ export default function report(problem) {
 	if (start) {
 		warningProperties.start = start;
 	} else if (isNumber(index)) {
-		warningProperties.index = index;
+		if (range?.start) {
+			warningProperties.start = range.start;
+		} else {
```
romainmenke, line 102: `range.start` will now be more accurate for our use case than passing `index`.

We should use it when available.

Given that we calculate it from `index` it is also gated behind the check that `index` is a number.

romainmenke: It is impossible to have an `index` value that is accurate and not to have a `range.start`.

A rule could create a report with an `index` and `line` while omitting the `node`, but then the `index` can't be used anyway.

lib/utils/report.mjs
```diff
@@ -93,14 +96,24 @@ export default function report(problem) {
...
 		warningProperties.start = start;
-	} else if (isNumber(index)) {
-		warningProperties.index = index;
+	} else if (range?.start) {
+		warningProperties.start = range.start;
+	} else if (isNumber(startLine)) {
+		warningProperties.start = {
+			line: startLine,
+			column: 1,
+		};
```
romainmenke, line 105: It is impossible to have an `index` value that is accurate and not to have a `range.start`.

A rule could create a report with an `index` and `line` while omitting the `node`, but then the `index` can't be used anyway.

The `index` is relative to the `node`, and when the `node` is missing, the `index` is essentially bogus.

If a rule does create such a report it is better to completely ignore the `index` and use only the `line`.

types/stylelint/index.d.ts
```diff
@@ -41,6 +41,7 @@ type EmptyResult = {
 			lang?: undefined;
```
Mouvedia, line 41: @jeddy3 can we remove `nodes` and `lang`?
cf https://github.com/stylelint/stylelint/commit/36e2dbb3fc1b42d670ddc97b55ee50f883c3f0d1#diff-9a4b261e295bef1ccce813aa194ac4023a1dcf98236667889964b1ba6fbc551fR56-R60

lib/utils/report.mjs
Mouvedia: I don't like the fallback for `column`.
If it's actually better than `} else if (isNumber(index)) {` could you add a comment that explains that choice?

romainmenke: > I don't like the fallback for `column`.

Setting the report to the first column on the given line seemed like an ok fallback to me.

We could calculate it by scanning the input line by line until we encounter the line in question. This seemed like overkill to me.

> If it's actually better than `} else if (isNumber(index)) {` could you add a comment that explains that choice?

_Already explained it in detail, but second attempt :)_

`index` is relative to the `node`. `index: 0` corresponds the first position in the node itself.

It is not the first position in the original source.

However where there is no `node` the `index` becomes meaningless.
PostCSS won't do anything with it and neither will anything else.

Its only purpose is to be used in a `node.range(...)` call.
If there is no node, there is no such call.

So `index` can't be used as a fallback for `range.start`.
Either there is a `node` and then there also is a `range.start` value or there is no `node` and then we can't use anyway.

Mouvedia: > This seemed like overkill to me.

👍 

> Already explained it in detail, but second attempt :)

Sorry maybe I wasn't clear, I meant: could you add a comment _in the code_ that explains why we are using magic numbers ? (i.e. `1` and `2`)
Your initial comment—_in the PR_—was very clear.

romainmenke: I've added a comment, hopefully that is enough of a hint.
I didn't want to inline the entire explanation 😅

lib/__tests__/disableRanges-integration.test.mjs
```diff
@@ -326,3 +327,161 @@ testRule({
...
+      }
+    `,
+			fixed: `
+      .foo {
+        /* a comment */
+        z-index: 99999999999999999999
+          /* stylelint-disable-next-line plugin/declaration-value-change-length */
+          99999 /* a comment */
+          99999999999999999999;
+      }
```
romainmenke, line 370: values that grow in length shift the lines in one direction

lib/__tests__/disableRanges-integration.test.mjs
```diff
@@ -326,3 +327,161 @@ testRule({
...
+9
+9
+9
+/* stylelint-disable-next-line */
+99999999999999999
+9
+9
+9;
+}
+`,
```
romainmenke, line 440: values that reduce in length shift the lines in the other direction

types/stylelint/index.d.ts
ybiquitous: I think they are no longer necessary. Even if they're removed, the type error doesn't happen. But I think we should remove in another PR.

Mouvedia:
> This will only happen for constructed nodes that lack a source object.

@romainmenke I won't review this PR right away.
But there's one thing that might change your perspective/approach: maybe we could not allow constructed nodes that lack a source object to be passed to `report`. It would require the rules to add it themselves. This is just an idea, maybe it's unfeasible/unpractical.

romainmenke:
> But there's one thing that might change your perspective/approach: maybe we could not allow constructed nodes that lack a source object to be passed to report. It would require the rules to add it themselves. This is just an idea, maybe it's unfeasible/unpractical.

We can revisit that separately, right?

The current codebase does assume that all PostCSS nodes have an optional `source` object.
We can make this stricter in various ways and doing so would allow us to remove some conditions and fallback values.
But this seems unrelated to the problem at hand, the incorrect report ranges.

types/stylelint/index.d.ts
ybiquitous: Just opened PR #8088 for the refactoring.

lib/utils/report.mjs
```diff
@@ -47,10 +49,11 @@ export default function report(problem) {
 		);
 	}
 
-	const range = node?.rangeBy({ index, endIndex }) ?? {};
+	const range = nodeRangeBy(node, { index, endIndex });
```
ybiquitous, line 52: [question] Does it make sense to make the call of `nodeRangeBy()` lazy until it's needed actually, in order to avoid extra calculation?

Mouvedia: @ybiquitous see #8083

ybiquitous: This PR conflicts with #8083. It seems better to me to close #8083 and take the #8083 change into this PR.

Mouvedia:
> > But there's one thing that might change your perspective/approach: maybe we could not allow constructed nodes that lack a source object to be passed to report. It would require the rules to add it themselves. This is just an idea, maybe it's unfeasible/unpractical.
> 
> We can revisit that separately, right?
> 
> The current codebase does assume that all PostCSS nodes have an optional `source` object. We can make this stricter in various ways and doing so would allow us to remove some conditions and fallback values. But this seems unrelated to the problem at hand, the incorrect report ranges.

Just to confirm, we have one core rule that is affected and you chose this course of action because we cannot know if there are other plugin rules that might be exhibiting the same problems, correct?
i.e. fixing it on the rule side would be a breaking change
Based on that I'd say your approach is actually less disruptive.

If we were to change our minds on this decision, it should be batched with other controversial changes.
e.g. #8050

ybiquitous:
[question] `PostCSS.Warning` internally calls `Node#rangeBy()`, but does this change affect `Warning`?

https://github.com/postcss/postcss/blob/1cc6ac3bf43a12460d3396e2cb76f14075daea1b/lib/warning.js#L9

ybiquitous:
@romainmenke Could you reproduce this bug in our demo?

romainmenke:
I've added a link to a demo in the original post

romainmenke:
@Mouvedia I am not sure what you mean 😕 

To circle back to my original comment:

> Some of these fields are optional and might be `undefined`/`null`.
> If that is the case we do a fallback to `node.rangeBy`.
> This will only happen for constructed nodes that lack a `source` object.

`node.source` is typed as an optional field in PostCSS.
That is how it is declared there and there isn't much we can do about that from our end.

`node.source` is only not set for constructed nodes where the creator of the new nodes doesn't set a `source` value.

I only wanted to clarify that this is how it is typed and that we need to set some fallback that makes sense.

The fallback I chose was to call `node.rangeBy` as that is already the current behavior.
We can discuss if another fallback makes more sense.

I don't think there is anything controversial here :)
(We already decided against `#8050`, we will not move forwards with that.)

-------

Making it so that `node.source` is typed as a non-optional field would be very large effort.
I guess this would require changing PostCSS itself to use more generics? Somehow differentiating between nodes that were the result of parsing and those constructed by users.

This seems out of scope to me?

lib/utils/nodeRangeBy.mjs
```diff
@@ -0,0 +1,84 @@
...
+ * This is a modified version of `Node.prototype.rangeBy` from PostCSS:
+ * https://github.com/postcss/postcss/blob/1cc6ac3bf43a12460d3396e2cb76f14075daea1b/lib/node.js#L212-L267
+ *
+ * It only supports what we use and it looks at the original source.
+ * Whereas PostCSS looks at the current string representation of the given node.
+ * We require the original position to accurately disable and enable rules.
+ *
+ * @param {Node} node
+ * @param {{ index?: number, endIndex?: number }} opts
+ * @return {{start: Range['start'], end: Range['end']} | undefined}
```
Mouvedia, line 19:
```suggestion
 * @return {Range | undefined}
```

lib/utils/report.mjs
romainmenke: I agree that it should be called lazily.

But I find the current shape of the code difficult when trying to do so in a robust manner. I find it hard to be sure that each function is using the correct values and that nothing is accidentally using the original inputs as found in the report.

My preferred solution would be to have a dedicated function to normalize all inputs relevant for the source location. I've opened a separate pull request for that idea: https://github.com/stylelint/stylelint/pull/8098

---

I've made it lazily in this pull request as:

```js
	let range;

	if (!line || !start || !end) {
		range = nodeRangeBy(node, { index, endIndex });
	}
```

lib/utils/report.cjs
```diff
@@ -91,14 +93,24 @@ function report(problem) {
...
+			line: startLine,
+			column: 1, // use the start of the line when `node` is undefined
+		};
 	}
 
 	if (end) {
 		warningProperties.end = end;
-	} else if (validateTypes.isNumber(endIndex)) {
-		warningProperties.endIndex = endIndex;
+	} else if (range?.end) {
```
Mouvedia, line 111: question
nit

Since we are sure that if `range.end` is present it's always accompanied by `range.start`, should we instead add another nesting?
i.e.
```js
if (range) {
  warningProperties.start = range.start;
  warningProperties.end = range.end;
} else { …
```
Does it even matter?
I mean, do we want to allow a form of overriding by passing either `start` and/or `end` to `report`?
i.e. is this by design?

romainmenke: I am unsure.
I find the current shape of the code very hard to reason about :)

It is unclear to me which exact combinations of optional values we support and what the expected outcome is for each.

I've filed https://github.com/stylelint/stylelint/pull/8098 with a possible improvement, although that still doesn't fully address the ambiguity.

lib/utils/report.mjs
Mouvedia: I now understand why you didn't do it in this PR originally.
i.e. with `|| !start || !end` it cannot be considered a refactor anymore

lib/utils/report.cjs
Mouvedia: wait a second

It was just a nit, damnit :)
Ill take a look to the other PR.

romainmenke:
> [question] `PostCSS.Warning` internally calls `Node#rangeBy()`, but does this change affect `Warning`?
> 
> https://github.com/postcss/postcss/blob/1cc6ac3bf43a12460d3396e2cb76f14075daea1b/lib/warning.js#L9
> 
> EDIT: `Node#error()` also calls `rangeBy()` internally:
> 
> https://github.com/postcss/postcss/blob/1cc6ac3bf43a12460d3396e2cb76f14075daea1b/lib/node.js#L114

I am not entirely sure what you mean @ybiquitous

When creating warnings with `index`, `endIndex` PostCSS will calculate the positions by serializing the nodes, causing the bug.

https://github.com/postcss/postcss/blob/1cc6ac3bf43a12460d3396e2cb76f14075daea1b/lib/node.js#L243-L245

But when passing `start`, `end` those values will be used directly and exactly as provided.

https://github.com/postcss/postcss/blob/1cc6ac3bf43a12460d3396e2cb76f14075daea1b/lib/node.js#L239-L242

So by pre-calculating `start` and `end` ourselves to the correct values and passing that when creating warnings we ensure that get the expected outcome.

Is there another aspect of the warnings that you are concerned about?

lib/utils/report.cjs
romainmenke: I filed that PR before you nitted 😛 
It was something I was already thinking about.

ybiquitous:
Since `Node#rangeBy()` is used somewhere in PostCSS, I'm concerned that this PR's partial forking to `rangeBy()` may affect different locations. 🤔 

In addition, such a forking makes the code more complex, so I'd like to avoid it if possible.

It seems that a tricky way like this may imply another root cause in our code base. Unfortunately, I don't have enough time to look into this issue for now.

romainmenke:
I'll open an upstream issue in PostCSS about this.
Might be interesting to gain some insights from their perspective.

Mouvedia:
Since you stated that #8074 was blocked by #8086 should we move on with #8083 for 16.11.0?

romainmenke:
I've filed an upstream issue: https://github.com/postcss/postcss/issues/1979

Mouvedia:
> Happy to provide a PR with a proposed fix.
> Sure, send PR with change and extra test.

lib/utils/report.cjs
```diff
@@ -45,10 +46,15 @@ function report(problem) {
...
 	}
 
-	const range = node?.rangeBy({ index, endIndex }) ?? {};
+	let range;
+
+	if (!line || !start || !end) {
+		range = nodeRangeBy(node, { index, endIndex });
+	}
+
 	// If a line is not passed, use the node.rangeBy method to get the
```
Mouvedia, line 55:
```suggestion
	// If a line is not passed, use the range to get the
```

romainmenke:
This has been resolved upstream.
I've rebased this branch to the PostCSS update brnahc to demonstrate that the added tests still pass without making any changes to Stylelint itself.

Do we still want these new tests?
There also is test coverage for this aspect in PostCSS now.

ybiquitous:
@romainmenke I really appreciate your work upstream! 👏🏼 

If you think these tests in this PR are not very helpful for Stylelint to detect bugs, let's close it.

PR #8088, opened 2024-10-27
Refactor to remove direct dependencies on PostCSS internals

ybiquitous:

Ref https://github.com/stylelint/stylelint/pull/8086/files#r1818082556

This refactors using only PostCSS public API to reduce our custom code.
https://postcss.org/api/

PostCSS's `Processor#process` returns an object of `LazyResult` or `NoWorkResult`:
https://github.com/postcss/postcss/blob/8.4.47/lib/processor.js#L44-L55

### DIFF
--- a/lib/getPostcssResult.cjs
+++ b/lib/getPostcssResult.cjs
@@ -3,7 +3,6 @@
 'use strict';
 
 const promises = require('node:fs/promises');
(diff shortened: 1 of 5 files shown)


romainmenke (approved):
LGTM 👍

Thank you for working on this 🙇

PR #8090, opened 2024-10-27
Fix `benchmark-rule` script to resolve `TypeError`

ybiquitous:

Related to #8044
Related to #8085

This fixes the following `TypeError` by using the Stylelint Public API, instead of PostCSS and Stylelint internals.

```
TypeError: Cannot read properties of undefined (reading 'config')
    at validateOptions (file:///Users/masafumi/git/stylelint/stylelint/lib/utils/validateOptions.mjs:19:40)
```

> [!WARNING]
> This makes the benchmark script slower (about 2x slower in my machine), while the script will be less fragile.

In addition, this converts the script's third argument from a rule context to a lint config because the `context.fix` property is now deprecated.

### My local benchmarks

```sh-session
$ node --run benchmark-rule -- value-keyword-case '["lower", {"camelCaseSvgKeywords": true}]'
Warnings: 280
Mean: 437.2615783125 ms
Deviation: 31.228799920667353 ms
```

```sh-session
$ node --run benchmark-rule -- value-keyword-case '["lower", {"camelCaseSvgKeywords": true}]' '{"fix":true}'
Warnings: 0
Mean: 492.96302766666673 ms
Deviation: 34.28542103536553 ms
```

### DIFF
--- a/scripts/benchmark-rule.mjs
+++ b/scripts/benchmark-rule.mjs
@@ -30,12 +30,12 @@ function printHelp() {
 
 if (!ruleName || !ruleOptions) {
 	printHelp();
-	exit(-1);
+	exit(1);
 }
 
 if (!stylelint.rules[ruleName]) {
 	console.error(`Unknown rule: ${ruleName}`);
-	exit(-1);
+	exit(1);
 }
 
 const CSS_URL = 'https://cdn.jsdelivr.net/npm/bootstrap@5.3.0/dist/css/bootstrap.css';
--- a/scripts/benchmark-rule.mjs
+++ b/scripts/benchmark-rule.mjs
@@ -3,19 +3,17 @@ import { argv, exit } from 'node:process';
 
 import Benchmark from 'benchmark';
 import picocolors from 'picocolors';
-import postcss from 'postcss';
 
-import normalizeRuleSettings from '../lib/normalizeRuleSettings.mjs';
-import rules from '../lib/rules/index.mjs';
+import stylelint from '../lib/index.mjs';
 
-const { bold, yellow } = picocolors;
+const { bold, red, yellow } = picocolors;
 
-const [, , ruleName, ruleOptions, ruleContext] = argv;
+const [, , ruleName, ruleOptions, config] = argv;
 
 function printHelp() {
 	const script = 'node benchmark-rule.mjs';
 
-	console.log(`Usage: ${script} <ruleName> <ruleOptions> [ruleContext]`);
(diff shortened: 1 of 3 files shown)


Mouvedia:
I am not sure this was the intended [when the script was created](https://github.com/stylelint/stylelint/commit/52c71d20e0e53e063dc1ae79cdf496d274976e24).
You want to test the banana but you are testing the gorilla and its jungle with it.
IMO just adding

```js
	result.stylelint ||= {
		customMessages: {},
		customUrls: {},
		disabledRanges: {},
		fixersData: {},
		ruleMetadata: {},
		ruleSeverities: {},
	};
```

is simpler and more in line with the original intent.

Ill let the other chime in.

---

> this converts the script's third argument from a rule context to a lint config

This is a good change though.

romainmenke:
I think that the purpose of the benchmark script is to make sure that Stylelint is fast enough for our users. Benchmarking more of Stylelint is still in line with that purpose. It also won't get in the way of finding performance regressions.

Mouvedia:
Alright as long as we don't use it as a metric between each releases.
If we ever do, we will have to go back to a lighter/stripped version.

scripts/benchmark-rule.mjs
```diff
@@ -24,17 +22,22 @@ function printHelp() {
...
 
-if (!ruleName || !(ruleName in rules) || !ruleOptions) {
+if (!ruleName || !ruleOptions) {
 	printHelp();
 	exit(-1);
 }
 
+if (!stylelint.rules[ruleName]) {
+	console.error(`Unknown rule: ${ruleName}`);
+	exit(-1);
```
romainmenke, line 38: Should this be `exit(1)`?
Maybe the negative non-zero exit code is intentional :)

romainmenke (approved):
Thank you. LGTM 👍

I've left one question about the exit codes that are used in this script, but these already existed, so not really blocking.

scripts/benchmark-rule.mjs
ybiquitous: Ah, you're entirely right. I just copied it from L33 without any thoughts. I'll fix it. 😅

ybiquitous:
> Alright as long as we don't use it as a metric between each releases.
> If we ever do, we will have to go back to a lighter/stripped version.

As far as I know, we never used this script for that purpose.

scripts/benchmark-rule.mjs
ybiquitous: Fixed via 5a9fc2868.

romainmenke (approved):
Thank you. LGTM 👍

PR #8095, opened 2024-10-29
Add `name` configuration property under `overrides`

### LINKED ISSUE #7854: Add `name` configuration property under `overrides`

ESLint recently allowed users to write a "name" option in eslint.config.js. Reference: <https://github.com/eslint/eslint/commit/4d11e567baff575146fd267b3765ab2c788aa1e5>

Use case 1: to describe the reasons for enabling or disabling some rules

```js
import { defaultEslintRules } from '@eslint/js';

export default [
  defaultEslintRules.configs.recommended,
  {
    files: ['file1.js'],
    name: "In this case, the max-lines rule was disabled temporarily to avoid an edge-case scenario, and we plan to enable back in the future.",
    rules: {
      'max-lines': 'off',
    },
  },
  {
    files: ['file2.js'],
    name: "In this case, max-lines rule was disabled to follow our company guideline.",
    rules: {
      'max-lines': 'off',
    },
  },
  {
    files: ['file3.js'],
    name: "Rules disabled by Maria, should remain like this. Please don't touch.",
    rules: {
      'max-lines': 'off',
      'max-lines-per-function': 'off',
      'max-statements': 'off',
      'no-magic-numbers': 'off',
    },
  },
];
```

Use case 2: to name the group of rules.

```js
name: 'eslint/recommended'
```

or

```js
name: 'eslint/strict'
```

I would like to write a "name" option in stylelint.config.js too.

For example:
```js
extends: [defaultConfig],
overrides: [
  {
    files: ['src/component/star.css'],
    name: 'our-rules/strict',
    rules: {
      'property-allowed-list': null,
      'property-disallowed-list': null,
    },
  },
],
```

or

```js
extends: [defaultConfig],
overrides: [
  {
    files: ['src/component/star.css'],
    name: 'These rules are disabled because of ...',
    rules: {
      'property-allowed-list': null,
      'property-disallowed-list': null,
    },
  },
],
```

ryo-manba:

Closes #7854 

No, it's self-explanatory.

### DIFF
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -391,7 +391,7 @@ The value of the `overrides` property is an array of objects. Each object:
 
 - must contain a `files` property, which is an array of glob patterns that specify which files the configuration should be applied to
 - should contain at least one other regular configuration property, such as `customSyntax`, `rules`, `extends`, etc.
-- may include a `name` property to provide a description of the override's purpose.
+- may contain a `name` property to provide a description of the override's purpose.
 
 The `customSyntax` property will be replaced, whereas `plugins`, `extends`, `rules`, etc. will be appended.
 
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -391,6 +391,7 @@ The value of the `overrides` property is an array of objects. Each object:
 
 - must contain a `files` property, which is an array of glob patterns that specify which files the configuration should be applied to
 - should contain at least one other regular configuration property, such as `customSyntax`, `rules`, `extends`, etc.
+- may include a `name` property to provide a description of the override's purpose.
 
 The `customSyntax` property will be replaced, whereas `plugins`, `extends`, `rules`, etc. will be appended.
 
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -391,7 +391,7 @@ The value of the `overrides` property is an array of objects. Each object:
 
 - must contain a `files` property, which is an array of glob patterns that specify which files the configuration should be applied to
 - should contain at least one other regular configuration property, such as `customSyntax`, `rules`, `extends`, etc.
-- may contain a `name` property to provide a description of the override's purpose.
+- may contain a `name` property to provide a description of the override's purpose
 
 The `customSyntax` property will be replaced, whereas `plugins`, `extends`, `rules`, etc. will be appended.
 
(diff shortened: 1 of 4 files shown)


ybiquitous (changes requested):
@ryo-manba Thanks for the pull request. Can you make additional changes?

- Add a test case for the new property, which should break nothing.
- Document the property in our configuration guide.

docs/user-guide/configure.md
```diff
@@ -391,6 +391,7 @@ The value of the `overrides` property is an array of objects. Each object:
 
 - must contain a `files` property, which is an array of glob patterns that specify which files the configuration should be applied to
 - should contain at least one other regular configuration property, such as `customSyntax`, `rules`, `extends`, etc.
+- may include a `name` property to provide a description of the override's purpose.
```
ybiquitous, line 394: [suggestion] To keep consistency with the 2 lines above.

```suggestion
- may contain a `name` property to provide a description of the override's purpose.
```

docs/user-guide/configure.md
```diff
@@ -391,6 +391,7 @@ The value of the `overrides` property is an array of objects. Each object:
 
 - must contain a `files` property, which is an array of glob patterns that specify which files the configuration should be applied to
 - should contain at least one other regular configuration property, such as `customSyntax`, `rules`, `extends`, etc.
+- may contain a `name` property to provide a description of the override's purpose.
```
ybiquitous, line 394: Remove an extra period to keep consistent with the line above.

```suggestion
- may contain a `name` property to provide a description of the override's purpose
```

lib/__tests__/overrides.test.mjs
```diff
@@ -68,6 +68,32 @@ describe('single input file. all overrides are matching', () => {
...
+					},
+				],
+			},
+			configBasedir: fixturesPath,
+		});
+
+		expect(linted.results).toHaveLength(1);
+		expect(linted.results[0].source).toBe(path.join(fixturesPath, 'style.css'));
+		expect(linted.results[0].warnings).toHaveLength(2);
+		expect(linted.results[0].warnings[0].rule).toBe('block-no-empty');
```
Mouvedia, line 94:
```suggestion
		expect(linted.results[0].warnings[0].rule).toBe('block-no-empty');
		expect(linted.results[0].warnings[1].rule).toBe('color-named');
```

Mouvedia (approved):

lib/__tests__/overrides.test.mjs
Mouvedia, line 94:
```suggestion
		expect(linted.results[0].warnings[0].rule).toBe('block-no-empty');
		expect(linted.results[0].warnings[1].rule).toBe('color-named');
```

ybiquitous:
Thank you for the additional review. I'll merge this. 😃

PR #8098, opened 2024-10-30
Refactor to normalize report locations before use

romainmenke:

See: https://github.com/stylelint/stylelint/pull/8086

I am unsure if this is the right direction and I didn't want to complicate review of #8086 by suddenly introducing new things. So happy to close this without merging if this is not something we want.

While working on #8086 and getting feedback in review I wondered if maybe we are lacking a step in `report()`.

Instead of ad-hoc calculations of position values it might be better to have a dedicated normalization step.

Calling `normalizeReportSourceLocation(problem)` would ensure that `problem` has position values of an expected shape.

Individual functions and call sites wouldn't need to each make the determination to look at `line`, `index`, `start`, ...

By grouping everything related to this it is also (slightly) simpler to reason about the possible outcomes.

### DIFF
--- a/lib/utils/report.cjs
+++ b/lib/utils/report.cjs
@@ -6,7 +6,7 @@ const validateTypes = require('./validateTypes.cjs');
 const constants = require('../constants.cjs');
 const nodeRangeBy = require('./nodeRangeBy.cjs');
 
-/** @import { DisabledRangeObject, Problem, Range, RuleMessage, StylelintPostcssResult, Utils, WarningOptions } from 'stylelint' */
+/** @import { DisabledRangeObject, Problem, ProblemWithNormalizedPositions, Range, RuleMessage, StylelintPostcssResult, Utils, WarningOptions } from 'stylelint' */
 
 /**
  * Report a problem.
@@ -46,7 +46,9 @@ function report(problem) {
 		);
 	}
 
-	if (!normalizeReportSourceLocation(problem)) {
+	normalizeReportSourceLocation(problem);
+
+	if (!isProblemWithNormalizedPositions(problem)) {
 		throw new Error(
 			`The "${ruleName}" rule failed to pass either a node or a line number to the \`report()\` function.`,
 		);
@@ -109,8 +111,8 @@ function report(problem) {
 }
 
 /**
- * @param {import('stylelint').Problem} problem
- * @return {problem is import('stylelint').ProblemWithNormalizedPositions}
+ * @param {Problem} problem
+ * @return {void}
  */
 function normalizeReportSourceLocation(problem) {
 	// Use the given node to calculate the start and end positions.
@@ -123,10 +125,10 @@ function normalizeReportSourceLocation(problem) {
 		}
 	}
 
-	if (!validateTypes.isNumber(problem.line) && validateTypes.isNumber(problem.start?.line)) {
+	if (!problem.line && problem.start?.line) {
 		// Fallback the line to the line of the start position
 		problem.line = problem.start.line;
-	} else if (!problem.start && validateTypes.isNumber(problem.line)) {
+	} else if (!problem.start && problem.line) {
 		// Fallback the start position to the line
 		problem.start = {
 			line: problem.line,
@@ -134,14 +136,22 @@ function normalizeReportSourceLocation(problem) {
 		};
 	}
 
+	// `index` and `endIndex` must not be used directly
 	delete problem.index;
 	delete problem.endIndex;
+}
 
-	if (problem.start && validateTypes.isNumber(problem.line)) {
-		return true;
-	}
-
-	return false;
+/**
+ * @param {Problem} problem
+ * @return {problem is ProblemWithNormalizedPositions}
+ */
+function isProblemWithNormalizedPositions(problem) {
+	return Boolean(
+		problem.start?.line &&
+			problem.line &&
+			typeof problem.index === 'undefined' &&
+			typeof problem.endIndex === 'undefined',
+	);
 }
 
 /**
@@ -193,7 +203,7 @@ function isDisabled(ruleName, startLine, disabledRanges) {
 	return false;
 }
 
-/** @param {Problem & { line: number }} problem */
+/** @param {ProblemWithNormalizedPositions & { line: number }} problem */
 function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
 	const { disabledRanges, config = {}, fixersData } = stylelint;
 
--- a/lib/utils/report.cjs
+++ b/lib/utils/report.cjs
@@ -23,7 +23,7 @@ const nodeRangeBy = require('./nodeRangeBy.cjs');
  * @type {Utils['report']}
  */
 function report(problem) {
-	const { node, index, endIndex, line, start, end, result, ruleName, word, fix, ...rest } = problem;
+	const { node, result, ruleName, word, fix, ...rest } = problem;
 	const {
 		disabledRanges,
 		quiet,
@@ -46,28 +46,24 @@ function report(problem) {
 		);
 	}
 
-	const range = nodeRangeBy(node, { index, endIndex });
-
-	// If a line is not passed, use the node.rangeBy method to get the
-	// line number that the complaint pertains to
-	const startLine = line ?? range?.start?.line;
-
-	if (!startLine) {
+	if (!normalizeReportSourceLocation(problem)) {
 		throw new Error(
 			`The "${ruleName}" rule failed to pass either a node or a line number to the \`report()\` function.`,
 		);
 	}
 
-	if (isFixApplied({ ...problem, line: startLine })) return;
+	const { start, end, line } = problem;
 
-	if (isDisabled(ruleName, startLine, disabledRanges)) {
+	if (isFixApplied({ ...problem, line })) return;
+
+	if (isDisabled(ruleName, line, disabledRanges)) {
 		// Collect disabled warnings
 		// Used to report `needlessDisables` in subsequent processing.
 		const disabledWarnings = (result.stylelint.disabledWarnings ||= []);
 
 		disabledWarnings.push({
 			rule: ruleName,
-			line: startLine,
+			line,
 		});
 
 		if (!ignoreDisables) return;
@@ -93,24 +89,10 @@ function report(problem) {
 
 	if (start) {
 		warningProperties.start = start;
-	} else if (range?.start) {
-		warningProperties.start = range.start;
-	} else if (validateTypes.isNumber(startLine)) {
-		warningProperties.start = {
-			line: startLine,
-			column: 1, // use the start of the line when `node` is undefined
-		};
 	}
 
 	if (end) {
 		warningProperties.end = end;
-	} else if (range?.end) {
-		warningProperties.end = range.end;
-	} else if (validateTypes.isNumber(startLine)) {
-		warningProperties.end = {
-			line: startLine,
-			column: 2, // the start of the line + 1 when `node` is undefined
-		};
 	}
 
 	if (word) {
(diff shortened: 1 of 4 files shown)


lib/utils/report.mjs
```diff
@@ -129,6 +111,42 @@ export default function report(problem) {
...
+	} else if (!problem.start && isNumber(problem.line)) {
+		// Fallback the start position to the line
+		problem.start = {
+			line: problem.line,
+			column: 1,
+		};
+	}
+
+	delete problem.index;
+	delete problem.endIndex;
```
romainmenke, line 141: Deleting `index` and `endIndex` ensures that nothing after normalization depends on these values.

lib/utils/report.mjs
```diff
@@ -129,6 +111,42 @@ export default function report(problem) {
...
+
+	if (!isNumber(problem.line) && isNumber(problem.start?.line)) {
+		// Fallback the line to the line of the start position
+		problem.line = problem.start.line;
+	} else if (!problem.start && isNumber(problem.line)) {
+		// Fallback the start position to the line
+		problem.start = {
+			line: problem.line,
+			column: 1,
+		};
```
romainmenke, line 137: I didn't add a fallback for the end position.
While I previously added this fallback:

```
warningProperties.end = {
 	line: startLine,
 	column: 2, // the start of the line + 1 when `node` is undefined
};
```

I am really unsure if we need a fallback for the end position and what that should be.

lib/utils/report.mjs
```diff
@@ -129,6 +111,42 @@ export default function report(problem) {
...
+	}
+
+	delete problem.index;
```
Mouvedia, line 140: note

I thought we had `no-delete-var` enabled.
Apparently not: we have 7 instances of `delete` in the current codebase.

lib/utils/report.cjs
Mouvedia, line 1: https://github.com/stylelint/stylelint/blob/42cd4757369706178fa2600dcca054b068682bd9/types/stylelint/index.d.ts#L821

If we are doing a refactor of `report`, I would like to resolve whether we respect the current type of `node`—i.e. it's not optional—or if we prescribe that `node` can be optional.
e.g. l51, l86, l117
i.e. is it reflecting the current usage of `report` in the wild or are those unnecessary checks?

Mouvedia, line 1: minor
refactor
l80–104

```js
	/** @type {WarningOptions} */
	const warningProperties = {
		severity: ruleSeverity,
		rule: ruleName,
		node,
		start,
		end,
		word,
		customUrl
	};
```

The only difference would be that if `word === ''` it will be set.

lib/utils/report.cjs
```diff
@@ -126,6 +108,42 @@ function report(problem) {
...
+	if (problem.node && (!problem.start || !problem.end)) {
+		const range = nodeRangeBy(problem.node, { index: problem.index, endIndex: problem.endIndex });
+
+		if (range) {
+			problem.start ??= range.start;
+			problem.end ??= range.end;
+		}
+	}
+
+	if (!validateTypes.isNumber(problem.line) && validateTypes.isNumber(problem.start?.line)) {
```
Mouvedia, line 126: nit
question

No one will pass it, since line always starts at `1`, but this allows `0`.
What about just `(!problem.line && problem.start?.line)`?

lib/utils/report.cjs
romainmenke: That is out of scope for me and I won't resolve on that here.
Please open an issue to discuss that if think it is something we need to revisit.

This change doesn't have the intention of making such API changes.
I only mean to make the code more understandable and robust.

romainmenke: I'd rather keep the current conditions as that makes it easier to reason about what we pass to the warnings.

nullish vs falsy bugs are really tricky and already something I am concerned about with these changed. I'd rather be explicit about what we intend and the values we pass around

lib/utils/report.cjs
romainmenke: See: https://github.com/stylelint/stylelint/pull/8098#discussion_r1824063826

lib/utils/report.cjs
Mouvedia: It seemed neater to me but indeed I agree it would be more obtuse.

lib/utils/report.cjs
romainmenke: If passing `line === 0` is a hard error then maybe we should throw an actual error?
That would be a separate PR, but it seems reasonable to me to do such validation.

types/stylelint/index.d.ts
```diff
@@ -850,6 +850,15 @@ declare namespace stylelint {
...
 	};
 
+	/** @internal */
+	export type ProblemWithNormalizedPositions = Problem & {
+		line: number;
+		start: Position;
+		end: Position;
+		index: undefined;
+		endIndex: undefined;
+	};
```
Mouvedia, line 860:
```suggestion
	export type ProblemWithNormalizedPositions = Omit<Problem, 'index' | 'endIndex'> &
                                                     Required<Pick<Problem, 'line' | 'start' | 'end'>>;
```

lib/utils/report.cjs
Mouvedia: Overly defensive programming is not that bad, it's just confusing when we are using a type declaration file that contradicts the code.
It's like having useless `assert`s, it doesn't hurt.

romainmenke: I really do not want to discuss this here :)
I am not saying I either agree or disagree.
Only saying that this is out of scope for this change.

Please open an issue to discuss this 🙇

types/stylelint/index.d.ts
romainmenke: Cool!

I didn't know about `Omit` and `Required`

Mouvedia: https://www.typescriptlang.org/docs/handbook/utility-types.html

lib/utils/report.mjs
Mouvedia: I don't know if we should have a fallback.
But if we do, you should use `word`—if present and doesn't have `\n`—to deduce the end column.

romainmenke: I don't think we should use `word` as a fallback for the lack of `end`.

`word` already has highest "priority": https://github.com/postcss/postcss/blob/1cc6ac3bf43a12460d3396e2cb76f14075daea1b/lib/node.js#L227-L237

If `word` is provided `start` and `end` won't be used at all by PostCSS.
So we don't have to use it as the fallback, it will be the primary value used by 
PostCSS.

This however is different from the logic in Stylelint and this could lead to differences between where an error is reported (the value calculated by PostCSS) and the location used to calculate disable ranges (the value calculated by Stylelint)

This difference is not new, it existed before these PR's.
If we want to make changes to the priority we should discuss so separately from this PR.

----------

I think the scope of the question for this PR is:

`end` already receives a fallback that will be used when `word` is not provided : https://github.com/postcss/postcss/blob/1cc6ac3bf43a12460d3396e2cb76f14075daea1b/lib/node.js#L217-L225

Given that we don't use `end` ourselves, do we then need to set a fallback to make it explicit?
Or should we leave it blank and leave it up to PostCSS?

Mouvedia: > Given that we don't use end ourselves, do we then need to set a fallback to make it explicit?
> Or should we leave it blank and leave it up to PostCSS?

At first I wanted to port the fallback to be in line with PostCSS but that wouldn't be a refactor PR anymore.
Since your stance is to limit the scope, I would say let's not add a fallback in this PR.
I would even say, if you can, remove more code.

lib/utils/report.cjs
Mouvedia:
```suggestion
	if (!problem.line && problem.start?.line) {
```

lib/utils/report.cjs
```diff
@@ -126,6 +108,42 @@ function report(problem) {
...
+		// Fallback the line to the line of the start position
+		problem.line = problem.start.line;
+	} else if (!problem.start && validateTypes.isNumber(problem.line)) {
```
Mouvedia, line 129:
```suggestion
	} else if (!problem.start && problem.line) {
```

lib/utils/report.cjs
```diff
@@ -126,6 +108,42 @@ function report(problem) {
...
+		// Fallback the start position to the line
+		problem.start = {
+			line: problem.line,
+			column: 1,
```
Mouvedia, line 133: see https://github.com/stylelint/stylelint/pull/8086#discussion_r1818098602

lib/utils/report.cjs
```diff
@@ -126,6 +108,42 @@ function report(problem) {
...
+		};
+	}
+
+	delete problem.index;
+	delete problem.endIndex;
+
+	if (problem.start && validateTypes.isNumber(problem.line)) {
```
Mouvedia, line 140: minor
question

If you pass a problem to `normalizeReportSourceLocation` without a `node` but with `start`, `end` and `line`.
This `if` will return `true` without throwing

> ```The "${ruleName}" rule failed to pass either a node or a line number to the \`report()\` function.```

correct?

lib/utils/report.cjs
```diff
@@ -126,6 +108,42 @@ function report(problem) {
...
+		return true;
+	}
+
+	return false;
```
Mouvedia, line 144:
```suggestion
	if (!problem.start || !problem.line) return false;

	delete problem.index;
	delete problem.endIndex;
	return true;
```

lib/utils/report.mjs
Mouvedia: Since you 👍 Ill resolve this thread.

lib/utils/report.cjs
romainmenke: There already is a comment above stating that we fallback to the start position of the line

romainmenke:
@Mouvedia maybe a better use of our time if we discuss the merits of this change before we nitpick this many round at this level of detail? :)

Feels like wasted effort if there is no consensus that this is the right direction.

Mouvedia:
> @Mouvedia maybe a better use of our time if we discuss the merits of this change before we nitpick this many round at this level of detail? :)
> 
> Feels like wasted effort if there is no consensus that this is the right direction.

Splitting that part from `report` makes it shorter.
I think it's an overall good refactor.

The only things I don't like, but that's really nitpicking, is

1. that the function returns a boolean but is not named isX or hasX
e.g. `isSourceLocationNormalized`
2. that the error's message l54 is either incorrect or there's a missing `node` check in `normalizeReportSourceLocation`
i.e. it seems unrelated

Sorry for the long waits in between the reviews, there are many PRs queued and I am not really available nowadays.

romainmenke:
> that the function returns a boolean but is not named `isX` or `hasX`
e.g. `isSourceLocationNormalized`

That is fair.

I think it might be better to split it into two parts:
- normalize the report
- validate the report

> that the error's message l54 is either incorrect or there's a missing node check in `normalizeReportSourceLocation`

Current status on main: https://github.com/stylelint/stylelint/blob/971b61e85564975e9b2741f7c286ea248e3b243e/lib/utils/report.mjs#L55-L59

The error message and the boolean check were never aligned 1 to 1.
The boolean check does the minimal verification to determine that the inputs were not sufficient.
The error message gives a hint on possible ways to resolve the error.

I don't think there is an issue here (and definitely not a new one :) )

lib/utils/report.cjs
romainmenke: `node` is optional and we have test coverage for this.

see https://github.com/stylelint/stylelint/pull/8098#issuecomment-2453050157 for more details.

lib/utils/report.cjs
romainmenke: I don't think this is meaningfully better but I also don't have a strong opinion about it. So I've applied these nits

romainmenke:
I've made some changes, but I am unsure if these are an improvement.

lib/utils/report.cjs
```diff
@@ -193,7 +203,7 @@ function isDisabled(ruleName, startLine, disabledRanges) {
 	return false;
 }
 
-/** @param {Problem & { line: number }} problem */
+/** @param {ProblemWithNormalizedPositions & { line: number }} problem */
```
Mouvedia, line 206: If you keep `isProblemWithNormalizedPositions`, then you can remove `& { line: number }`.

romainmenke: Good catch!
Thank you

lib/utils/report.mjs
```diff
@@ -133,6 +113,50 @@ export default function report(problem) {
...
+}
+
+/**
+ * @param {Problem} problem
+ * @return {problem is ProblemWithNormalizedPositions}
+ */
+function isProblemWithNormalizedPositions(problem) {
+	return Boolean(
+		problem.start?.line &&
+			problem.line &&
```
Mouvedia, line 154: question

Shouldn't this be `(problem.start?.line || problem.line) && […]`?

Mouvedia (approved):
I don't think that the
```js
		problem.start = {
			line: problem.line,
			column: 1,
		};
```
fallback is necessary because it's not a refactor anymore.

romainmenke:
Closing as this is no longer needed

PR #8106, opened 2024-11-02
Add `lax`/`strict` values to `fix` option and CLI flag

### LINKED ISSUE #7497: Add `--fix=lax` and `--fix=strict`

Analyzing a CSS file with an invalid syntax raises a syntax error expectedly, while doing the same specifying `--fix` succeeds unexpectedly.

`test.css`:

```css
// comment
```

---

For reference, this behavior difference depends on [`postcss-safe-parser`](https://www.npmjs.com/package/postcss-safe-parser):

https://github.com/stylelint/stylelint/blob/382961fb6eb67557910a1ff7430a7a180f888bcc/lib/getPostcssResult.mjs#L127

Note: we found this problem in #7494.

```json
{"rules": {}}
```

Run the following commands in my console.

```sh-session
$ npx stylelint test.css

test.css

$ npx stylelint test.css --fix

$ echo $?
0
```

`package.json`:

```json
{
  "dependencies": {
    "stylelint": "16.2.0"
  }
}
```

Expect to raise a syntax error. E.g.

```sh-session
$ npx stylelint test.css --fix

test.css

```

### What actually happened?

A syntax error does not happen, and then Stylelint exits normally. E.g.

```sh-session
$ npx stylelint test.css --fix

$ echo $?
0
```

1. do something similar to https://github.com/noahbrenner/xo/blob/8e4f4355c4b8e62bc28f37e3673bb6872e693bb0/main.js#L119-L133

2. update [fix option](https://stylelint.io/user-guide/options/#fix)

`true` maps to `"lax"`
`false` doesn't change
`"lax"` aliases to `true`
`"strict"` added

ryo-manba:

Closes #7497 

Added `lax` and `strict` values to the `--fix` option.

- `lax` : uses `postcss-safe-parser`, attempting to fix as much as possible even when there are syntax errors. 
- `strict`: uses `postcss parser`, only fixes when there are no syntax errors.


lib/cli.mjs
```diff
@@ -468,8 +468,14 @@ export default async function main(argv) {
...
-	if (isBoolean(fix)) {
-		options.fix = fix;
+	if (isString(fix)) {
+		if (fix === 'true' || fix === '') {
+			options.fix = true;
+		} else if (fix === 'false') {
+			options.fix = false;
+		} else if (fix === 'strict' || fix === 'lax') {
+			options.fix = fix;
+		}
```
ryo-manba, line 478: It seems that this test is not being reflected in the coverage.
https://github.com/stylelint/stylelint/pull/8106/files#diff-0104e2476d83a456a5e8add0acc8334030af22abad28e606b33f9d8094c04e5dR106-R110

lib/cli.mjs
```diff
@@ -238,7 +238,7 @@ const flags = {
 		type: 'boolean',
 	},
 	fix: {
-		type: 'boolean',
+		type: 'string',
```
Mouvedia, line 241: As explained [here](https://github.com/stylelint/stylelint/issues/7497#issuecomment-2085529108), you don't have to change the type.
i.e. leave `boolean` and retrieve the value in the `if`

lib/__tests__/cli.test.mjs
```diff
@@ -35,7 +35,6 @@ describe('buildCLI', () => {
 			cache: false,
 			color: false,
 			disableDefaultIgnores: false,
-			fix: false,
```
Mouvedia, line 38: `fix` default should remain `false`
see https://github.com/stylelint/stylelint/pull/8106/files#r1826592064

lib/cli.mjs
Mouvedia, line 478:
```suggestion
	if (isBoolean(fix) || isString(fix)) {
		switch (fix) {
			case '':
			case true:
			case 'lax':
			  options.fix = true;
			  break;
			case false:
			case 'strict':
			  options.fix = fix;
			  break;
		}
```

I don't know if we want a `default` with a `throw` or not.

lib/cli.mjs
ryo-manba: I see your point, but shouldn't it be treated as a `string` rather than a `boolean`? Definig it as a `boolean` and then retrieving a `string` value seems inconsistent and relies on an upstream issue.
I'm following the approach shown here.
https://github.com/noahbrenner/xo/blob/8e4f4355c4b8e62bc28f37e3673bb6872e693bb0/main.js#L119-L133

Mouvedia: Well either is fine but if we leave it as `'boolean'`, it will reduce the diff considerably.
e.g. the default remains `false`

ybiquitous: I'm not 100% sure, but can `boolean` accept `--fix=lax`?

IMO, if the code doesn't become more complex, `string` seems more acceptable to me.

However, the change should be limited within `cli.mjs`, so either seems to have no big problems.

lib/cli.mjs
ybiquitous: No, directly calling `buildCLI()` (in the test case) doesn't call `main()`. We should believe the coverage report and add a proper case.

docs/user-guide/options.md
```diff
@@ -36,7 +36,11 @@ CLI flag: `--fix`
 
 Automatically fix, where possible, problems reported by rules.
 
-For CSS with standard syntax, Stylelint uses [postcss-safe-parser](https://www.npmjs.com/package/postcss-safe-parser) to fix syntax errors.
+Options are:
+
+- `true` or `lax` - uses [postcss-safe-parser](https://www.npmjs.com/package/postcss-safe-parser) to fix as much as possible, even when there are syntax errors
+- `strict` - uses [PostCSS Parser](https://postcss.org/api/#postcss-parser) and only fixes problems when there are no syntax errors
```
ybiquitous, line 42: [suggestion] To distinguish from boolean values, how about adding quotes like this?

```suggestion
- `true` or `"lax"` - uses [postcss-safe-parser](https://www.npmjs.com/package/postcss-safe-parser) to fix as much as possible, even when there are syntax errors
- `"strict"` - uses [PostCSS Parser](https://postcss.org/api/#postcss-parser) and only fixes problems when there are no syntax errors
```

lib/cli.mjs
```diff
@@ -238,7 +238,7 @@ const flags = {
...
 	},
```
Mouvedia, line 242: question

If you add `choices: ['lax', 'strict'],` does it throw?

lib/__tests__/cli.test.mjs
Mouvedia: see https://github.com/stylelint/stylelint/pull/8106/files#r1829415336

docs/user-guide/options.md
```diff
@@ -36,7 +36,11 @@ CLI flag: `--fix`
...
+- `false` (default) - does not fix any problems
```
ybiquitous, line 44: [suggestion] This is a just idea and not required, but what about adding the following tip?

> [!TIP]
> If you don't care about strictness now, we recommend using `true` because `"strict"` may become the default in the future.

```suggestion

> [!TIP]
> If you don't care about strictness now, we recommend using `true` because `"strict"` may become the default in the future.

```

lib/cli.mjs
ybiquitous, line 1: [suggestion] We need to update the help of `--fix` also, don't we?

https://github.com/stylelint/stylelint/blob/13933eda2c6430ea2c595f02f47799adbda771df/lib/cli.mjs#L77-L79

lib/cli.mjs
Mouvedia, line 478:
```suggestion
	if (isString(fix)) {
			switch (fix) {
				case '':
				case 'true':
				case '1':
				case 'lax':
				  options.fix = true;
				  break;
				case 'false':
				case '0':
				  options.fix = false;
				  break;
				case 'strict':
				  options.fix = 'strict';
				  break;
			}
```

types/stylelint/index.d.ts
```diff
@@ -110,7 +110,7 @@ declare namespace stylelint {
 		_processorFunctions?: Map<string, ReturnType<Processor>['postprocess']>;
 		allowEmptyInput?: boolean;
 		cache?: boolean;
-		fix?: boolean;
+		fix?: boolean | 'lax' | 'strict';
```
ybiquitous, line 113: [suggestion] The following internal properties may make the implementation a bit more understandable by reducing repetitions:

```suggestion
		fix?: boolean | 'lax' | 'strict';
		/** @internal */
		_fixLax?: boolean;
		/** @internal */
		_fixStrict?: boolean;
```

E.g.

```diff
-	const fix = stylelint._options.fix === 'lax' || stylelint._options.fix === true;
 	const syntax = await (customSyntax
 		? getCustomSyntax(customSyntax, stylelint._options.configBasedir)
-		: cssSyntax(fix));
+		: cssSyntax(stylelint._options._fixLax));
```

ybiquitous (commented):
[question] Considering consistency, should we add also a [`fix` configuration property](https://github.com/stylelint/stylelint/blob/13933eda2c6430ea2c595f02f47799adbda771df/docs/user-guide/configure.md#fix)? I'm still uncertain. 🤔

lib/cli.mjs
ryo-manba: Thanks! I've added.
[test: add cli coverage for --fix options](https://github.com/stylelint/stylelint/pull/8106/commits/8c04faae57c115370a903816af588610ec529cdc)

lib/cli.mjs
```diff
@@ -74,9 +74,15 @@ const helpText = `
         A pattern of files to ignore (in addition to those in "${DEFAULT_IGNORE_FILENAME}").
         Multiple can be set.
 
-      --fix
+      --fix <mode>
```
ybiquitous, line 77: [suggestion] Is it acceptable to mark the mode optional?

```suggestion
      --fix [<mode>]
```

lib/cli.mjs
```diff
@@ -74,9 +74,15 @@ const helpText = `
...
+
+        Automatically fix problems of certain rules. The modes are as follows:
+
+          true         automatically fixes issues, same as "lax" (default)
+          false        does not fix any problems
+          lax          attempting to fix as much as possible even with syntax errors
+          strict       only fixing when there are no syntax errors
```
ybiquitous, line 85: [suggestion] We can remove the blank line:

```suggestion
```

lib/cli.mjs
ybiquitous: Having `choices` makes sense to me. 👍🏼 

Also, can we have `default`?

Mouvedia: We can't.
see https://github.com/stylelint/stylelint/pull/8106#discussion_r1826592183

ryo-manba:
```js
fix: {
  choices: ['lax', 'strict'],
  default: 'lax',
}
```

I think we could use `default` but for compatibility, maybe it's better to include `['lax',  'strict', 'true', 'false']`?

Mouvedia: Try; as long as it doesn't `throw`.
If it doesn't remember that `0` and `1` should still be supported, as before.

Mouvedia: > What exactly does the support for 0 and 1 mean? I tested it locally, but those values didn't reflect as booleans.

I thought it was a feature of meow.
It's not documented so you can ignore my suggestion of supporting `0`/`1`.

Mouvedia: > Also, can we have `default`?

@ybiquitous check `default:` of https://github.com/stylelint/stylelint/pull/8106#discussion_r1829449217

lib/__tests__/__snapshots__/cli.test.mjs.snap
```diff
@@ -51,9 +51,14 @@ exports[`CLI --help 1`] = `
       A pattern of files to ignore (in addition to those in ".stylelintignore").
       Multiple can be set.
 
-    --fix
+    --fix [<mode>]
```
ybiquitous, line 54: [question] Should we support `--no-fix` also, as well as `--no-cache`, `--no-color`, `--no-validate`?

lib/__tests__/cli.test.mjs
```diff
@@ -586,6 +589,81 @@ describe('CLI', () => {
...
 	});
 
+	it('applies --fix=lax and fixes without error message', async () => {
+		jest.spyOn(process, 'stdin', 'get').mockImplementationOnce(() =>
+			Readable.from([
+				Buffer.from(stripIndent`
+				a { top: 0px }
+				a {
+			`),
+			]),
```
ybiquitous, line 599: [suggestion] How about correction the indentation like this?

```suggestion
			Readable.from([
				Buffer.from(stripIndent`
					a { top: 0px }
					a {
				`),
			]),
```

lib/__tests__/cli.test.mjs
```diff
@@ -586,6 +589,81 @@ describe('CLI', () => {
...
+	});
+
+	it('applies --fix=true (default) and fixes without error message', async () => {
+		jest.spyOn(process, 'stdin', 'get').mockImplementationOnce(() =>
+			Readable.from([
+				Buffer.from(stripIndent`
+				a { top: 0px }
+				a {
+			`),
+			]),
```
ybiquitous, line 625: [suggestion] Same suggestion:

```suggestion
			Readable.from([
				Buffer.from(stripIndent`
					a { top: 0px }
					a {
				`),
			]),
```

lib/__tests__/cli.test.mjs
```diff
@@ -586,6 +589,81 @@ describe('CLI', () => {
...
+	});
+
+	it('does not fix with --fix=strict and displays error message', async () => {
+		jest.spyOn(process, 'stdin', 'get').mockImplementationOnce(() =>
+			Readable.from([
+				Buffer.from(stripIndent`
+				a { top: 0px }
+				a {
+			`),
+			]),
```
ybiquitous, line 651: [suggestion] Same:

```suggestion
			Readable.from([
				Buffer.from(stripIndent`
					a { top: 0px }
					a {
				`),
			]),
```

lib/__tests__/__snapshots__/cli.test.mjs.snap
Mouvedia: meow does support it by default, the thing is we only document `--no` variant when the `false` is more relevant.
i.e. when the default is `true`
e.g. [`--no-validate`](https://stylelint.io/user-guide/cli/#--validate---no-validate)

Weirdly we have done it for [--no-color](https://stylelint.io/user-guide/cli/#--color---no-color) too.

ybiquitous: Thanks, I confirmed `--no-fix` sets `{ fix: false }` by default. But it seems a bit weird because `fix` is defined as `string`.

Also, I can't find a use case of `--no-fix`, so I think documenting `--no-fix` is unnecessary for now. Sorry for confusing.

lib/__tests__/standalone.test.mjs
```diff
@@ -466,3 +467,99 @@ describe('existing codeFilename for nested config detection', () => {
...
+
+describe('standalone with fix options for CSS syntax errors', () => {
+	const cssWithSyntaxError = '// comment';
+	const cssIncompleteRule = 'a {';
+	const cssNonSyntaxError = stripIndent`
+    a { colr: hsl(20, 10%, 30%, 5%); }
+  `;
+	const expectedFixedCode = stripIndent`
+    a { colr: hsl(20deg 10% 30% / 5%); }
+  `;
```
ybiquitous, line 479: [suggestion] Can you fix the indentation with whitespaces, not tabs? Unluckily, Prettier doesn't format code inside template literals.

```suggestion
	const cssNonSyntaxError = stripIndent`
		a { colr: hsl(20, 10%, 30%, 5%); }
	`;
	const expectedFixedCode = stripIndent`
		a { colr: hsl(20deg 10% 30% / 5%); }
	`;
```

ybiquitous: Or we can also write them with one line:

```suggestion
	const cssNonSyntaxError = 'a { colr: hsl(20, 10%, 30%, 5%); }';
	const expectedFixedCode = 'a { colr: hsl(20deg 10% 30% / 5%); }';
```

lib/cli.mjs
ybiquitous: I think throwing an error makes sense when an unknown fix mode is specified. In this case, we can control the exit code by catching the error in `buildCLI()` like this:

```js
export function buildCLI(argv) {
	try {
		return meow(helpText, {
			// ...
		});
	} catch (e) {
		handleError(e);
		process.exit();
	}
}
```

Furthermore, the suggested definition makes sense to me:

```js
fix: {
  choices: ['lax', 'strict'],
  default: 'lax',
}
```

Then, I don't think it's necessary to support `--fix=true` and `--fix=false`. The following is enough:

- `--fix` equals `--fix=lax`
- `--fix=lax` is valid
- `--fix=strict` is valid
- `--fix=unknown` throws an error: `Unknown value for flag `--fix` ...`

lib/cli.mjs
ybiquitous: If we adopt this way on <https://github.com/stylelint/stylelint/pull/8106/files#r1836690240>, such a logic using `if` or `switch` becomes unnecessary, doesn't it?

lib/__tests__/__snapshots__/cli.test.mjs.snap
Mouvedia: > But it seems a bit weird because `fix` is defined as `string`.

That's one of the reason why [I preferred to keep it at as `"boolean"`](https://github.com/stylelint/stylelint/pull/8106#discussion_r1827281601).
Since it's not documented and the default is `false`, it cannot be considered a regression if `--no-fix` stops working or throws.

lib/cli.mjs
Mouvedia: > `--fix=true`

Not supporting this one could be considered a breaking change.
I thought that we had to support `--fix=1` too.
I am not sure. You should be careful.

lib/cli.mjs
Mouvedia: Or you could add `"true"` and `"false"` as string choices.

ybiquitous: I don't know why we should also support `"true"` or `"false"`.

types/stylelint/index.d.ts
ybiquitous: Instead of a normalized autofix mode in a hidden field, it may be better to add a utility function to normalize it. For example, the function might be as follows:

```js
export function normalizeFixMode(options, config) {
  // Check options first

  // If options have no fix mode, check config
}
```

I want to combine checks like `fix === 'lax'` in one place as much as possible.

lib/__tests__/__snapshots__/cli.test.mjs.snap
ybiquitous: Similar to <https://github.com/stylelint/stylelint/pull/8106#discussion_r1836742056>, I don't think we need to support `--no-fix`. Even if it may work accidentally, it's undocumented. Considering undocumented behaviors makes things too complex.

lib/cli.mjs
Mouvedia: There are some cases in [those results](https://github.com/search?q=stylelint+%22--fix%3Dtrue%22&type=code).
That's doesn't mean we should officially support `--fix=true`.
I am just saying that it's not without consequences.

lib/cli.mjs
Mouvedia: Maybe add a changeset then?
e.g.

```
Removed: undocumented support of `--fix=true` and `--fix=false`
```

lib/__tests__/__snapshots__/cli.test.mjs.snap
Mouvedia: Ill resolve this thread.

---

off topic

I have a branch which uses `parseArgs` of `node:utils` instead of `meow`.
The complexity wasn't worth it, performance wise, so I didn't send a PR.

Now that you are saying that we don't have to support `--no-` prefix for all boolean flags, Ill come back to it.
Without all the logic necessary for `-no--` handling, the complexity burden is out of the window and the perf might be worth the refactor.

Thanks for the information.

lib/cli.mjs
ybiquitous: Thanks for providing the resources. But, I still think it's unnecessary to support `--fix=true` officially.

If we want to rescue such undocumented behaviors, we might be able to support it unofficially like this:

```js
fix: {
  choices: ['lax', 'strict', 'true'], // 'true' is unsupported officially
  default: 'lax',
}
```

However, this would increase confusion/complexity in the future, so I don't prefer it so much as a maintainer.

lib/cli.mjs
ybiquitous: I don't think it's required because these are undocumented and unsupported. Rather, such a changeset would make people confused.

types/stylelint/index.d.ts
ryo-manba: Thanks for all the suggestions! I agree that adding a utility function would make the implementation cleaner.
I'll take care of this later.

lib/cli.mjs
ryo-manba: There’s no mention of booleans in [options](https://stylelint.io/user-guide/options/#fix), but since configure specifies [`"fix": true`](https://stylelint.io/user-guide/configure/#fix), some people might think it accepts booleans. Either way, it might be better to also update the config to align with this (as you asked: 
https://github.com/stylelint/stylelint/pull/8106#pullrequestreview-2415826249).

Mouvedia: Again we can't have `default: 'lax',`; that would switch the default from `false` to `true`.

ryo-manba: Yes, that's right.

The `choices` default applies when nothing is specified or when no arguments are provided.

So, there are two issues.

```js
fix: {
  choices: ['lax', 'strict'],
  default: 'lax'
}
```
In this case, `npx stylelint *.css` would apply `--fix=lax` by default (meaning `--fix` is not left unspecified).

```js
fix: {
  choices: ['lax', 'strict', 'true', 'false'],
  default: 'false'
}
```

This looks good at first, but running `npx stylelint *.css --fix` would end up applying `--fix=false`.

The following configuration might work better.

```js
fix: {
  choices: [true, 'lax', 'strict'],
}
```

This way, if only `--fix` is specified, the value will be `true` (not a string), which aligns with expected behavior.

Mouvedia: I would be very surprised if `choices: [true, 'lax', 'strict'],` were allowed when `type: 'string',` is set.
I would even be more surprised if not setting a `type` is allowed.

IMHO if you want to use `choices` with `type: 'string'` you will probably end up with 2 alternatives:
- either `choices: ['lax', 'strict', '', 'true', 'false', '1', '0'],` + `default: 'false'`
- or `choices: ['lax', 'strict', ''],` and no `default`

If you can mix and match booleans and strings in `choices` go for it, but Id expect meow to throw.

types/stylelint/index.d.ts
ryo-manba: Applied in [this commit](https://github.com/stylelint/stylelint/pull/8106/commits/f59eade747442f97d3257ad85954c20be8badbcb).

lib/cli.mjs
ryo-manba: Applied in [this commit](https://github.com/stylelint/stylelint/pull/8106/commits/f59eade747442f97d3257ad85954c20be8badbcb).

lib/cli.mjs
```diff
@@ -74,9 +74,14 @@ const helpText = `
...
-      --fix
+      --fix [<mode>]
 
-        Automatically fix problems of certain rules.
+        Automatically fix problems of certain rules. The modes are as follows:
+
+          true         automatically fixes issues, same as "lax" (default)
+          false        does not fix any problems
+          lax          attempting to fix as much as possible even with syntax errors
+          strict       only fixing when there are no syntax errors
```
ryo-manba, line 84: @ybiquitous @Mouvedia 
[This commit](https://github.com/stylelint/stylelint/pull/8106/commits/f59eade747442f97d3257ad85954c20be8badbcb) modifies the behavior to accept only the following three patterns:

```shell
# If nothing is specified, meow treats it as true
--fix
--fix=lax
--fix=strict
```

If there are no issues with this behavior, I'll update the docs as well.

To support receiving values like the following:

```shell
--fix=true
--fix=false
```

It needs to be configured as follows:

```js
fix: {
  choices: [true, 'lax', 'strict', 'true', 'false'],
}
```

lib/cli.mjs
ybiquitous: Thank you. I now understand `choices: [true, 'lax', 'strict']` is the best. 👍🏼 

In addition, I would like to make the CLI argument spec simpler. For example, the help would output like this:

```
    --fix [<mode>]

      Automatically fix problems of certain rules. The modes are as follows:

        lax          attempting to fix as much as possible even with syntax errors (default)
        strict       only fixing when there are no syntax errors
```

Does this make sense?

lib/cli.mjs
Mouvedia: Interesting; can you keep the `type` when using a `choices` that mixes booleans and strings?
If not, let's only use `choices` and drop the type.

lib/cli.mjs
Mouvedia: style
nit

If you can try to also align the other flag values.
i.e. the second column should be aligned
```
          metadata     by metadata of a file (default)
          content      by content of a file

[…]

          string       human-readable strings (default)
          compact      similar to ESLint's compact formatter
          github       workflow commands for GitHub Actions (DEPRECATED)
          json         JSON format
          tap          TAP format
          unix         C compiler-like format
          verbose      extend "string" to include a file list and a tally
```

lib/getPostcssResult.cjs
```diff
@@ -25,9 +26,10 @@ async function getPostcssResult(stylelint, { customSyntax, filePath, code } = {}
 		return cached;
 	}
 
+	const fix = normalizeFixMode(stylelint._options, stylelint._options.config);
 	const syntax = await (customSyntax
 		? getCustomSyntax(customSyntax, stylelint._options.configBasedir)
-		: cssSyntax(stylelint._options.fix));
+		: cssSyntax(fix));
```
Mouvedia, line 32: @ybiquitous I don't think it will matter much, but this part is not a refactor.
i.e. `config.fix` used to be ignored

So either it was deliberately done and it needs a comment or it was a careless omission.

lib/cli.mjs
Mouvedia: > If nothing is specified, `fix` will not be applied. In other words, the default is `undefined`.

I saw that you had to surround `fix` with `Boolean()`; would adding `default: false,` permit to use `fix` directly like before?

ybiquitous: I agree 4 cases should be provided as you commented 👍🏼 

- unspecified
- `--fix` (equals `--fix=lax`)
- `--fix=lax`
- `--fix=strict`

I don't think it's needed to support `--fix=true` and `--fix=false`.

lib/getPostcssResult.cjs
ybiquitous: @Mouvedia Good catch. Since this is a behavior change, I think we should separate it from this PR to make it easier to track the commit history. 👍🏼 

At a glance, it seems that we should support the `fix: true` config also here, in addition to `--fix`, but I'm not 100% sure. Can you confirm how Stylelint behaves when `fix: true` config is specified?

Mouvedia: > Can you confirm how Stylelint behaves when `fix: true` config is specified?

Let's postpone that to the new issue/PR.

> I think we should separate it from this PR…

@ryo-manba you were right in trying to support both `config.fix` and `fix`, nonetheless we are trying to have commits that can be reverted as a whole. [We might also have to support `configFile` to be thorough.](https://github.com/stylelint/stylelint/pull/8106#discussion_r1844933085)
Could you only pass `LinterOptions` as a single argument to `normalizeFixMode `?
It will be fixed separately.

lib/utils/normalizeFixMode.mjs
```diff
@@ -0,0 +1,20 @@
+/**
+ * Normalize the fix mode based on options and configuration.
+ *
+ * @param {import('stylelint').LinterOptions} options
+ * @param {import('stylelint').Config | undefined} config
+ * @returns {'lax' | 'strict' | undefined}
```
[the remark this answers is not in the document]
Mouvedia: Passing the second argument is not necessary.
i.e. `config` is already part of the first argument

lib/standalone.mjs
```diff
@@ -117,6 +118,8 @@ export default async function standalone({
 
 	const formatterFunction = await getFormatter(stylelint);
 
+	const autofix = normalizeFixMode(stylelint._options) ?? config?.fix === true ? 'lax' : undefined;
```
ryo-manba, line 121: Update this section when supporting config.

lib/getPostcssResult.cjs
ryo-manba: Fixed in this commit.
[Config is not supported](https://github.com/stylelint/stylelint/pull/8106/commits/6c071ee93bd21e01f7ce9d9b59160a79f416803b)

Mouvedia: @ybiquitous see https://github.com/stylelint/stylelint/pull/8106/files/f59eade747442f97d3257ad85954c20be8badbcb..6c071ee93bd21e01f7ce9d9b59160a79f416803b#r1845332494

lib/cli.mjs
```diff
@@ -468,8 +471,10 @@ export default async function main(argv) {
 		options.cacheStrategy = cacheStrategy;
 	}
 
-	if (isBoolean(fix)) {
-		options.fix = fix;
+	if (fix === true || fix === 'lax') {
+		options.fix = 'lax';
+	} else if (fix === 'strict') {
+		options.fix = 'strict';
 	}
```
Mouvedia, line 478: note
minor

If you were to pass `stylelint._options.fix` to `normalizeFixMode` instead of `stylelint._options` you could refactor this part into

```suggestion
	options.fix = normalizeFixMode(fix);`
```

It may complicate the subsequent PR that will support `config.fix`.
@ybiquitous what do you think?

types/stylelint/index.d.ts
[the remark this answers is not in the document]
Mouvedia:
```suggestion
		fix?: boolean | FixMode;
```

lib/standalone.mjs
```diff
@@ -250,7 +256,12 @@ export default async function standalone({
 				/**
 				 * If we're fixing, save the file with changed code
 				 */
-				if (postcssResult.root && postcssResult.opts && !postcssResult.stylelint.ignored && fix) {
+				if (
+					postcssResult.root &&
+					postcssResult.opts &&
+					!postcssResult.stylelint.ignored &&
+					Boolean(autofix)
```
Mouvedia, line 263: In the same way as https://github.com/stylelint/stylelint/pull/8106#discussion_r1844933564, I think that this change, [which use `config.fix`](https://github.com/stylelint/stylelint/pull/8106/files#diff-a9211bfb8f0453e9d6e413fe613dbdcd388621ae34881d7400c93f10dafe516bR121), we should keep `&& fix` if it doesn't break your tests.
i.e. have a separate PR that takes into account `config` and `configFile`

ryo-manba: Thanks, good catch! Fixed it.
[fix: config is not yet supported](https://github.com/stylelint/stylelint/pull/8106/commits/ac7866f9cc0fb21df72353367841279bdbe03c52)

docs/user-guide/options.md
Mouvedia: refs
stylelint/stylelint.io#409
https://github.com/incentro-ecx/remark-github-admonitions-to-directives#%EF%B8%8F-customizing-the-mapping

lib/__tests__/standalone.test.mjs
```diff
@@ -466,3 +466,95 @@ describe('existing codeFilename for nested config detection', () => {
 		expect(results[0].warnings).toHaveLength(1);
 	});
 });
+
+describe('standalone with fix options for CSS syntax errors', () => {
```
ybiquitous, line 470: [suggestion] It seems reasonable that this case would be added to `standalone-fix.test.mjs`. Can you move it?

lib/cli.mjs
ybiquitous: Agree. Can you please refactor it?

lib/getPostcssResult.mjs
```diff
@@ -116,7 +118,7 @@ async function getCustomSyntax(customSyntax, basedir) {
 }
 
 /**
- * @param {boolean | undefined} fix
+ * @param {import('stylelint').Config['fix']} fix
```
ybiquitous, line 121: [suggestion] Let's simplify the type:

```suggestion
 * @param {boolean} fix
```

lib/standalone.mjs
```diff
@@ -250,7 +256,12 @@ export default async function standalone({
 				/**
 				 * If we're fixing, save the file with changed code
 				 */
-				if (postcssResult.root && postcssResult.opts && !postcssResult.stylelint.ignored && fix) {
+				if (
+					postcssResult.root &&
+					postcssResult.opts &&
+					!postcssResult.stylelint.ignored &&
+					Boolean(fix)
```
ybiquitous, line 263: [suggestion] Is this conversion to `Boolean(fix)` necessary? If not, can you revert it to reduce the diff?

ybiquitous (changes requested):
Sorry for my late response. I agree with postponing the config support to a follow-up issue/PR.

Overall LGTM, but can you address a few minor comments and resolve the conflict?

lib/standalone.mjs
ryo-manba: Fixed in this commit.
[refactor: apply suggestion](https://github.com/stylelint/stylelint/pull/8106/commits/7af47cca0e1d865258d4f5e61ff8801198ae627e)

lib/getPostcssResult.mjs
ryo-manba: Fixed in [this commit](https://github.com/stylelint/stylelint/pull/8106/commits/caccdc943017ecb2b198b12908eb33c3ed4e2317).

lib/__tests__/standalone.test.mjs
ryo-manba: Fixed in [this commit](https://github.com/stylelint/stylelint/pull/8106/commits/caccdc943017ecb2b198b12908eb33c3ed4e2317).

lib/standalone.mjs
```diff
@@ -117,6 +118,9 @@ export default async function standalone({
 
 	const formatterFunction = await getFormatter(stylelint);
 
+	const autofix =
+		(normalizeFixMode(stylelint._options.fix) ?? config?.fix === true) ? 'lax' : undefined;
```
ybiquitous, line 122: [suggestion] I wonder if the position of `()` is incorrect 🤔 The current `autofix` could be only `'lax'` or `undefined`, not `strict`. Is this intentional?

```suggestion
		normalizeFixMode(stylelint._options.fix) ?? (config?.fix === true ? 'lax' : undefined);
```

ybiquitous: Besides, if my concern is right, there might be no test cases for `'strict'` 🤔

ryo-manba: Thanks for your suggestion!
Actually, since `postcss.parse` is chosen in `lib/getPostcssResult.mjs` and a parse error results in `postcssResult` being `undefined`, there's no need for the separate `lax` and `strict` handling in `standalone.mjs`.

Fixe in this commit.
[refactor: remove unnecessary variables](https://github.com/stylelint/stylelint/pull/8106/commits/c28206ba586ddc8b0f8f2fdead689c6648293b85)

lib/__tests__/standalone-fix.test.mjs
```diff
@@ -836,3 +836,95 @@ it("doesn't return any fixed code if fixing is disabled for the rule, regardless
 
 	expect(result.code).toBeUndefined();
 });
+
+describe('standalone with fix options for CSS syntax errors', () => {
```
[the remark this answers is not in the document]
ryo-manba: Applied it!
[refactor: remove unnecessary variables](https://github.com/stylelint/stylelint/pull/8106/commits/c28206ba586ddc8b0f8f2fdead689c6648293b85)

ybiquitous (commented):
Thank you for your continuous work! Overall LGTM 👍🏼 

Lastly, can you have a look at a few minor suggestions? They're all optional.

ybiquitous (approved):

docs/user-guide/options.md
```diff
@@ -36,7 +36,13 @@ CLI flag: `--fix`
...
 Automatically fix, where possible, problems reported by rules.
 
-For CSS with standard syntax, Stylelint uses [postcss-safe-parser](https://www.npmjs.com/package/postcss-safe-parser) to fix syntax errors.
+Options are:
+
+- `"lax"` (default) - uses [postcss-safe-parser](https://www.npmjs.com/package/postcss-safe-parser) to fix as much as possible, even when there are syntax errors
+- `"strict"` - uses [PostCSS Parser](https://postcss.org/api/#postcss-parser) and only fixes problems when there are no syntax errors
+
+> [!TIP]
+> If you don't care about strictness now, we recommend using `--fix` because `"strict"` may become the default in the future.
```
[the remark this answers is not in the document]
Mouvedia:
```suggestion
> If you don't care about strictness now, we recommend using the boolean flag `--fix`—which _currently_ maps to `--fix=lax`—without any value because `"strict"` may become the new default in the future.
```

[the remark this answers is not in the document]
ryo-manba: Thank you both for your suggestions. 
Since it says `"lax" (default)`, I thought it’s unnecessary to mention that it’s mapped!

lib/cli.mjs
```diff
@@ -468,8 +472,8 @@ export default async function main(argv) {
 		options.cacheStrategy = cacheStrategy;
 	}
 
-	if (isBoolean(fix)) {
-		options.fix = fix;
+	if (fix === true || fix === 'lax' || fix === 'strict') {
+		options.fix = normalizeFixMode(fix);
 	}
```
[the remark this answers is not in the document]
ryo-manba: That's a good idea!

ybiquitous (approved):
LGTM! 😄

ybiquitous:
@jeddy3 If you're interested, can you take a look?

docs/user-guide/options.md
```diff
@@ -36,7 +36,13 @@ CLI flag: `--fix`
...
 Automatically fix, where possible, problems reported by rules.
 
-For CSS with standard syntax, Stylelint uses [postcss-safe-parser](https://www.npmjs.com/package/postcss-safe-parser) to fix syntax errors.
+Options are:
+
+- `"lax"` (default) - uses [postcss-safe-parser](https://www.npmjs.com/package/postcss-safe-parser) to fix as much as possible, even when there are syntax errors
+- `"strict"` - uses [PostCSS Parser](https://postcss.org/api/#postcss-parser) and only fixes problems when there are no syntax errors
+
+> [!TIP]
+> If you don't care about strictness now, we recommend using `--fix` without any option because `"strict"` may become the default in the future.
```
jeddy3, line 45:
```suggestion
> We recommend using `--fix=strict` if you are linting a language extension like SCSS
```

Mouvedia:
@jeddy3 what you are describing resembles stylelint/vscode-stylelint#490 which apparently has a [hackfix](https://github.com/stylelint/vscode-stylelint/issues/490#issuecomment-2156218548).

ybiquitous:
@jeddy3 Yes, when linting code supporting `//` comments like SCSS, using `--fix=strict` could be helpful.

However, the answer might be "no" when linting other syntaxes.

Also, I think we should make `strict` default in the future for usability, as I commented in <https://github.com/stylelint/stylelint/issues/7497#issuecomment-2075109495>:

> - Make `--fix=strict` default in the next major version
> - Remove `--fix=lax` in the future

So, I'd like to encourage people to use `--fix` as before to keep their Stylelint configurations as much as possible.

What do you think?

jeddy3 (approved):
@ybiquitous Thank you for the explanation. All SGTM.

lib/getPostcssResult.cjs
Mouvedia: @ryo-manba #8333

PR #8107, opened 2024-11-03
Refactor to cleanup after selector report range fixes

romainmenke:

After all the fixes for the reported ranges of selector problems there were still a few opportunities to cleanup existing code, slightly improve test coverage, and a few very minor improvements.

I bundled these because they were so minor but I also don't mind splitting this up.

I don't think a changeset is needed because most changes aren't user facing and those that are, already have a relevant changeset.

### DIFF
--- a/lib/rules/selector-class-pattern/index.mjs
+++ b/lib/rules/selector-class-pattern/index.mjs
@@ -50,19 +50,17 @@ const rule = (primary, secondaryOptions) => {
 		const normalizedPattern = isString(primary) ? new RegExp(primary) : primary;
 
 		root.walkRules((ruleNode) => {
-			const { selector, selectors } = ruleNode;
-
 			if (!isStandardSyntaxRule(ruleNode)) {
 				return;
 			}
 
-			if (selectors.some((s) => isKeyframeSelector(s))) {
+			if (ruleNode.selectors.some(isKeyframeSelector)) {
 				return;
 			}
 
 			// Only bother resolving selectors that have an interpolating &
-			if (shouldResolveNestedSelectors && hasInterpolatingAmpersand(selector)) {
-				for (const nestedSelector of resolveNestedSelector(selector, ruleNode)) {
+			if (shouldResolveNestedSelectors && hasInterpolatingAmpersand(ruleNode.selector)) {
+				for (const nestedSelector of resolveNestedSelector(getRuleSelector(ruleNode), ruleNode)) {
 					if (!isStandardSyntaxSelector(nestedSelector)) {
 						continue;
 					}
--- a/lib/rules/selector-disallowed-list/__tests__/index.mjs
+++ b/lib/rules/selector-disallowed-list/__tests__/index.mjs
@@ -203,7 +203,7 @@ testRule({
 		},
 		{
 			code: '.foo > /*comment*/.bar {}',
-			message: messages.rejected('.foo > .bar'),
+			message: messages.rejected('.foo > /*comment*/.bar'),
 			line: 1,
 			column: 1,
 			endLine: 1,
--- a/lib/rules/selector-disallowed-list/index.cjs
+++ b/lib/rules/selector-disallowed-list/index.cjs
@@ -75,38 +75,38 @@ const rule = (primary, secondaryOptions) => {
 				}
 			}
 
-			if (splitList) {
-				parseSelector(getRuleSelector(ruleNode), result, ruleNode)?.each((selector) => {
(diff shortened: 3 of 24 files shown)


lib/rules/selector-disallowed-list/__tests__/index.mjs
```diff
@@ -203,7 +203,7 @@ testRule({
 		},
 		{
 			code: '.foo > /*comment*/.bar {}',
-			message: messages.rejected('.foo > .bar'),
+			message: messages.rejected('.foo > /*comment*/.bar'),
```
romainmenke, line 206: This is user facing, but this is a minor improvement that follows on previous work to fix the reported ranges.

See: https://github.com/stylelint/stylelint/blob/971b61e85564975e9b2741f7c286ea248e3b243e/.changeset/sixty-kings-join.md?plain=1#L5

Mouvedia: Remember to add `& [@romainmenke](https://github.com/romainmenke)`

lib/rules/selector-disallowed-list/index.cjs
```diff
@@ -75,38 +75,38 @@ const rule = (primary, secondaryOptions) => {
...
+						messageArgs: [selectorStr],
 						node: ruleNode,
-						word,
+						index,
+						endIndex,
 					});
 				}
 			}
+
+			const selectors = getRuleSelector(ruleNode);
```
Mouvedia, line 97: nit

We have 3 cases of singular and one plural.
i.e. maybe rename `selector`

Mouvedia, line 97: nit

maybe rename to `selector`
it makes more sense when passing it to `parseSlector`

lib/rules/selector-type-case/index.mjs
```diff
@@ -48,8 +49,7 @@ const rule = (primary, secondaryOptions) => {
 		}
 
 		root.walkRules((ruleNode) => {
-			let hasComments = ruleNode.raws.selector && ruleNode.raws.selector.raw;
-			const selector = hasComments ? hasComments : ruleNode.selector;
+			const selector = ruleNode.selector;
```
Mouvedia, line 52: question

Is this on purpose?
i.e. didn't you want `const selector = getRuleSelector(ruleNode);` instead?

lib/rules/selector-disallowed-list/index.cjs
romainmenke: Can't, `selector` is already used

lib/rules/selector-type-case/index.mjs
romainmenke: No, this is on purpose.

Here `ruleNode.selector` is only used for fast aborts.

lib/rules/selector-disallowed-list/__tests__/index.mjs
romainmenke: I don't really need the attribution unless there is a requirement to add these? ☺️

lib/rules/selector-class-pattern/index.mjs
```diff
@@ -50,19 +50,17 @@ const rule = (primary, secondaryOptions) => {
...
-			if (selectors.some((s) => isKeyframeSelector(s))) {
+			if (ruleNode.selectors.some(isKeyframeSelector)) {
 				return;
 			}
 
 			// Only bother resolving selectors that have an interpolating &
-			if (shouldResolveNestedSelectors && hasInterpolatingAmpersand(selector)) {
-				for (const nestedSelector of resolveNestedSelector(selector, ruleNode)) {
+			if (shouldResolveNestedSelectors && hasInterpolatingAmpersand(ruleNode.selector)) {
+				for (const nestedSelector of resolveNestedSelector(getRuleSelector(ruleNode), ruleNode)) {
```
Mouvedia, line 63: The changelog will have to be edited for this one too.

lib/rules/selector-type-case/index.mjs
Mouvedia, line 1: great code reduction

Mouvedia (approved):

romainmenke:
Thank you for reviewing this and for the feedback @Mouvedia 🙇

lib/rules/selector-disallowed-list/index.cjs
Mouvedia: maybe `ruleSelector` was available
anyway, not important

PR #8108, opened 2024-11-04
Add support for profiling rule performance via the `TIMING` environment variable

### LINKED ISSUE #7350: Support profiling rule performance like ESLint

I'd like to know which rules takes a lot of time.

ESLint provides that feature by setting `TIMING=1` environment variable.
https://eslint.org/docs/latest/extend/custom-rules#profile-rule-performance

It would be happy to provide the way of profiling rule performance like ESLint.

ryo-manba:

Closes #7350

Added support for `TIMING` environment variable to track rule performance, similar to [ESLint's profiling feature](https://eslint.org/docs/latest/extend/custom-rules#profile-rule-performance).
Set `TIMING=1` to display the top 10 longest-running rules or `TIMING=all` to display all rules.

### `TIMING=1`

image

### `TIMING=all`

image


Mouvedia:
This looks cool.

Normally you ask before taking a _status: ask to implement_ issue.
That's for your own sake, having to scrap your PR because _things have changed_ would be disheartening; trust me Iv been there.
Anyway since it has 3 👍 and @jeddy3 is the one who accepted it Ill switch it to _status: ready to implement_.

lib/timing.mjs
```diff
@@ -0,0 +1,173 @@
...
+		if (enabled) {
+			data[key] = (data[key] ?? 0) + timeDiff;
+		}
+
+		return result;
+	};
+}
+
+if (enabled) {
+	process.on('exit', () => {
```
Mouvedia, line 164: question

Since you are not using the `code` argument, do you want to display the data even when there are errors?

lib/timing.mjs
```diff
@@ -0,0 +1,173 @@
...
+		widths
+			.map((width, index) => {
+				const extraAlignment = index !== 0 && index !== widths.length - 1 ? 2 : 1;
+
+				return ALIGN[index] ? ALIGN[index](':', width + extraAlignment, '-') : ':';
+			})
+			.join('|'),
+	);
+
+	console.log(table.join('\n')); // eslint-disable-line no-console -- Debugging function
```
Mouvedia, line 136:
```suggestion
	console.debug(table.join('\n'));
```

lib/timing.mjs
Mouvedia, line 1: question

I know that this is based on ESLint timing.js but couldn't we use [`table`](https://github.com/gajus/table) since we already have it in our depedencies?

lib/timing.mjs
ryo-manba: Yes, that's correct.

image

lib/timing.mjs
ryo-manba: I personally feel that `console.log` is more natural for user-facing information, while `console.debug` might be better for developer logs. By the way, ESLint seems to use `console.log`. What do you think?
https://github.com/eslint/eslint/blob/4a88a54aeb456abe94da9f5b3d5472d14b3116f1/lib/linter/timing.js#L119

ryo-manba: Ah, perhaps the comment `"Debugging function"` is unnecessary.

Mouvedia: > What do you think?

I think that if it's a "debugging function" then `console.debug` is appropriate.

lib/timing.mjs
ryo-manba: Thanks for the information. I'll give it a try!
It appears that the code you mentioned isn't included in the current codebase.

ryo-manba: I've updated. The code has been significantly reduced. Thanks!
[Use table library for formatting output](https://github.com/stylelint/stylelint/pull/8108/commits/8d607c37a2f3dd2e6a87489c2f2bd9f4c7a9c777)

image

lib/timing.mjs
ryo-manba: I've removed this comment.

lib/timing.mjs
Mouvedia: > The code has been significantly reduced.

That's what I hoped for.
Sadly it didn't appease codecov.

lib/lintPostcssResult.mjs
```diff
@@ -107,13 +108,16 @@ export default async function lintPostcssResult(stylelintOptions, postcssResult,
...
+		const ruleFn = ruleFunction(primaryOption, secondaryOptions, context);
+
+		/**
+		 * @param {import('postcss').Root} postcssRoot
+		 */
+		async function timedRuleFn(postcssRoot) {
+			return await timing.time(ruleName, async () => ruleFn(postcssRoot, postcssResult))();
+		}
+
+		performRules.push(Promise.all(postcssRoots.map(timedRuleFn)));
```
Mouvedia, line 120: question

Wouldn't it be better to use `timing.time` only if necessary?
e.g.

```js
const foo = async (postcssRoot) => ruleFn(postcssRoot, postcssResult);

/* … */

const fn = process.env.TIMING ? timedRuleFn : foo;
performRules.push(Promise.all(postcssRoots.map(fn)));
```

ryo-manba: Thanks for catching it!
I wanted to encapsulate the env access related to timing within `timing.js`, so I tried it like this.
[Changed to run only when timing is enabled](https://github.com/stylelint/stylelint/pull/8108/commits/ea546fb59c699ae9349031003c986735afe0b83d)

lib/timing.mjs
```diff
@@ -0,0 +1,125 @@
+import process from 'node:process';
+import { table } from 'table';
+
+// Inspired by ESLint's timing.js
+// https://github.com/eslint/eslint/blob/main/lib/linter/timing.js
```
ybiquitous, line 5: [suggestion] I think a permalink with a commit ID is better than the `main` branch because `main` is changeable. E.g.

```suggestion
// https://github.com/eslint/eslint/blob/09bc2a88c00aa9a93c7de505795fc4e85b2e6357/lib/linter/timing.js
```

lib/timing.mjs
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Start time measurement.
+ * @returns {[number, number]} Variable for tracking time.
+ */
+function startTime() {
+	return process.hrtime();
```
ybiquitous, line 12: [suggestion] I didn't know it until now, but `hrtime.bigint()` sounds better:

> Legacy. Use [process.hrtime.bigint()](https://nodejs.org/api/process.html#processhrtimebigint) instead.

-- https://nodejs.org/api/process.html#processhrtimetime

Mouvedia: if he does he will have to change l18, l28 and probably adjust the output to be rounded and/or converted.

ryo-manba: Thanks for your suggestion! Changed in [this commit](https://github.com/stylelint/stylelint/pull/8108/commits/9d071f7347171a3d6be33c71f5be81bb62af2796).

lib/timing.mjs
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	if (process.env.TIMING.toLowerCase() === 'all') {
+		return Number.POSITIVE_INFINITY;
+	}
+
+	const TIMING_ENV_VAR_AS_INTEGER = Number.parseInt(process.env.TIMING, 10);
+
+	return TIMING_ENV_VAR_AS_INTEGER > 10 ? TIMING_ENV_VAR_AS_INTEGER : MINIMUM_SIZE;
+}
```
Mouvedia, line 48:
```suggestion
const TIMING = process.env.TIMING;
const enabled = Boolean(TIMING);
const HEADERS = ['Rule', 'Time (ms)', 'Relative'];

/**
 * Decide how many rules to show in the output list.
 * @returns {number} The number of rules to show.
 */
function getListSize() {
	const MINIMUM_SIZE = 10;
	const displayAllRules = 'all';

	if (typeof TIMING !== 'string') {
		return MINIMUM_SIZE;
	}

	if (TIMING.toLowerCase() === displayAllRules) {
		return Number.POSITIVE_INFINITY;
	}

	const TIMING_ENV_VAR_AS_INTEGER = Number.parseInt(TIMING, 10);

	return TIMING_ENV_VAR_AS_INTEGER > 10 ? TIMING_ENV_VAR_AS_INTEGER : MINIMUM_SIZE;
}
```

lib/timing.mjs
```diff
@@ -0,0 +1,125 @@
...
+import { table } from 'table';
+
+// Inspired by ESLint's timing.js
+// https://github.com/eslint/eslint/blob/09bc2a88c00aa9a93c7de505795fc4e85b2e6357/lib/linter/timing.js
+
+/**
+ * Start time measurement.
+ * @returns {bigint} Variable for tracking time in nanoseconds.
+ */
+function startTime() {
```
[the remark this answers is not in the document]
ryo-manba: I think the existing approach of wrapping with `startTime` and `endTime` is easier to understand.
Isn't it better to keep it as is? I don't think we need to force commonization.

lib/__tests__/timing.test.mjs
```diff
@@ -0,0 +1,44 @@
...
+import timing from '../timing.mjs';
+
+describe('timing', () => {
+	describe('getListSize()', () => {
+		afterEach(() => {
+			delete process.env.TIMING;
+		});
+
+		it('should return minimum list size when TIMING is not set or has a small value', () => {
+			delete process.env.TIMING;
```
Mouvedia, line 12: Normally you do something like this:

```js
  const OLD_ENV = process.env;

  beforeEach(() => {
    jest.resetModules()
    process.env = { ...OLD_ENV };
  });

  afterEach(() => {
    process.env = OLD_ENV;
  });
```

ryo-manba: @Mouvedia 
Sorry, I'm late.
I've applied your suggestion in [this commit](https://github.com/stylelint/stylelint/pull/8108/commits/7668318f8c7bf55d048eed374f805ff10f219523).

Mouvedia: The reason why I suggested this was because of 31328ee.
i.e. you should be able to revert it now

ryo-manba: I tried to revert that commit, but the test fails.

lib/__tests__/timing.test.mjs
```diff
@@ -0,0 +1,51 @@
...
+		afterAll(() => {
+			process.env = OLD_ENV;
+		});
+
+		it('should return minimum list size when TIMING is not set or has a small value', () => {
+			delete process.env.TIMING;
+			expect(timing.getListSize()).toBe(10);
+
+			process.env.TIMING = 'true';
+			expect(timing.getListSize()).toBe(10);
```
Mouvedia, line 23: This one i covered by `'foo'`.
i.e. it can be removed

What about replacing it with `''`?
i.e. test another falsy value

docs/user-guide/cli.md
```diff
@@ -136,6 +136,46 @@ Force enable/disable the validation of the rules' options. [More info](options.m
 
 Show the currently installed version of Stylelint.
 
+## Profile Rule Performance
+
+The `TIMING` environment variable allows you to profile the performance of individual rules.
+
+When set, it displays the execution time of the longest-running rules upon linting completion. This includes both rule creation and execution time, as well as each rule's relative impact as a percentage of the total rule processing time.
+
+```shell
```
Mouvedia, line 145:
````suggestion
```console
````

File: lib/__tests__/timing.test.mjs (hunk 9 above)
Comment (ryo-manba) [reply lib/__tests__/timing.test.mjs:23]: `foo` is a meaningless value. `true` indicates a boolean was specified, which has a different meaning.

Comment (Mouvedia) [reply lib/__tests__/timing.test.mjs:23]: i.e. a test for accidental coercions that may detect regressions

File: lib/timing.mjs [hunk 11]
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Decide how many rules to show in the output list.
+ * @returns {number} The number of rules to show.
+ */
+function getListSize() {
+	const TIMING = process.env.TIMING;
+	const MINIMUM_SIZE = 10;
+
+	if (typeof TIMING !== 'string') {
```
Mouvedia, line 37: nit
style

```suggestion
	if (typeof TIMING === 'undefined') {
```

I think this would be clearer because an environment variable is always `string | undefined`.
Either way it's fine.

ryo-manba: Fixed it.
https://github.com/stylelint/stylelint/pull/8108/commits/ebbdd0deb70cd744dd77f8817854ef94e0152e5e

jeddy3 (approved):
LGTM, thank you!

It fits in nicely with our goal to identify our slower rules so that we can make them faster.

Let's wait for a 2nd approval on this before merging, though.

docs/user-guide/cli.md
```diff
@@ -136,6 +136,46 @@ Force enable/disable the validation of the rules' options. [More info](options.m
...
+╟────────────────────────────────────────────────────┼───────────┼──────────╢
+║ selector-class-pattern                             │     0.215 │     3.1% ║
+╟────────────────────────────────────────────────────┼───────────┼──────────╢
+║ length-zero-no-unit                                │     0.190 │     2.7% ║
+╟────────────────────────────────────────────────────┼───────────┼──────────╢
+║ value-keyword-case                                 │     0.154 │     2.2% ║
+╚════════════════════════════════════════════════════╧═══════════╧══════════╝
+```
+
+By default, it shows the top 10 longest-running rules. You can adjust the number of results by setting a different value for `TIMING`.
```
ybiquitous, line 172: [question] What does this "default" mean? Can we run the timing like this?

```shell
TIMING= stylelint "*.css"
```

Mouvedia: It works similarly to the `eslint` one.
If you check the tests, it will be clearer.
i.e. between 1 and 10 the table will always have a minimum of 10 rows

Since `TIMING=1` can be considered a boolean, it's not illogical.
For 2–9 it's kinda unexpected though.

ybiquitous: Thanks for the explanation. IMO, this behavior could be more explicit. I don't know why ESLint behaves like this, but `TIMING={top n}` seems more straightforward.

Any thoughts?

ybiquitous (commented):
[suggestion] This is nice-to-have, but how about adding a rank number column like this?

```
╔════╤═════════════════════════════════════════╤═══════════╤══════════╗
║  # │ Rule                                    │ Time (ms) │ Relative ║
╟────┼─────────────────────────────────────────┼───────────┼──────────╢
║  1 │ declaration-property-value-no-unknown   │    14.480 │    55.2% ║
╟────┼─────────────────────────────────────────┼───────────┼──────────╢
║  2 │ no-descending-specificity               │     2.554 │     9.7% ║
╟────┼─────────────────────────────────────────┼───────────┼──────────╢
║  3 │ no-duplicate-selectors                  │     0.946 │     3.6% ║
╟────┼─────────────────────────────────────────┼───────────┼──────────╢
```

docs/user-guide/cli.md
Mouvedia: Probably because the 2 main use cases are
- enable timing (`TIMING=1`)
- change default number of rows (i.e. more than 10)

---

@ryo-manba what happens when you have less than 10 rules enabled in your config?
Could you add a test that ensures that the empty rows are removed in that case?
e.g. TIMING=40 but you have 34 rules enabled

ryo-manba:
@ybiquitous 
Are you suggesting not just displaying an index column but also sorting the table by time in descending order?

ybiquitous:
I mean to just add the number column without the ordering (time desc).

docs/user-guide/cli.md
ryo-manba: @Mouvedia 
I think it's obvious that it's not being displayed since it's not looping over a number.
Testing I/O is going to take a bit of work, if you have a good way, please suggest it!

ryo-manba: I think users who are used to ESLint might want 10 (or more) with `TIMING=1`.
However, I don't have a strong preference, so I'm fine with @ybiquitous's suggestion as well.

ryo-manba:
@ybiquitous 
Oh, I had already implemented it to do so. (I forgot 😅)

docs/user-guide/cli.md
ryo-manba: Using the value of `TIMING` directly results in the following. 

image

Is there any problem? Please feedback.

ybiquitous: Personally, I still think `TIMING=1` is confusing. So, if anyone doesn't have strong opinions, I recommend `TIMING={top n}`.

To enable the timing, `TIMING=true` might be more straightforward for me than `TIMING=1`. However, it shouldn't be required at this time.

Mouvedia: @ybiquitous just to be clear you want to show one row for `TIMING=1`?
Just asking to be sure because, `1`/`0` are often used for bool by env variables hence you would expect an output that is exploitable.
i.e. 2 or more, which explains the minimum of 10

---

> `TIMING=0`

IMO it should not show the table as if `TIMING` wasn't passed.
For invalid values—not integer nor `'all'`—an error message could be expected or it could act like `false`/`0`.

ybiquitous: @Mouvedia I agree that `TIMING=0` doesn't make sense, but I'm not sure why `TIMING=1` is exploitable.

If we support `TIMING={n}`, I think `TIMING=0` or `TIMING=1` should be interpreted as an integer, not a boolean.

To make things simple, I suggest:

- Accept only `TIMING={n}` with `n >= 1`
- Ignore other invalid values, e.g. `TIMING=0`, `TIMING=-1`, `TIMING=`, `TIMING=foo`, etc.

ybiquitous:
I'm reconsidering the `TIMING` environment variable. If we think of the possibility of conflict with the environment variable (e.g., other tools may use the variable), it seems to make sense to provide a CLI flag like `--timing` instead. Any thoughts? 🤔

docs/user-guide/cli.md
Mouvedia: eslint/eslint#13671

Mouvedia:
If you have a separate npm script for each type of lint it shouldn't matter.
i.e. only having to prefix one environment variable to `"lint:*"` could be considered a convenience (if we share the same default for `true`/`1`)

> Any thoughts?

I think that it doesn't matter.
i.e. I don't have a preference for either one

But with the flag we will have to start discussing whether we add its configuration property and Node.js option counterparts.
The environment variable doesn't have this problem.

ybiquitous:
> But with the flag we will have to start discussing whether we add its configuration property and Node.js option counterparts.
> The environment variable doesn't have this problem.

Make sense. Agree with choosing the `TIMING` environment variable 👍🏼

Mouvedia:
This is not to be implemented in this PR but I gotta ask, what about supporting `TIMING=rule-name`?
e.g. `TIMING=no-descending-specificity`

ybiquitous:
> what about supporting `TIMING=rule-name`?

That sounds good, although it should not be a blocker of this PR 👍🏼 

Probably, supporting multiple rules would be fine, e.g., `TIMING=rule-1, rule-2`.

docs/user-guide/cli.md
ryo-manba: I think boolean support is complicated, and it’s not even clear if there’s any demand for it.
Maybe we can just go with the simple @ybiquitous proposal for now, and consider other options if requests come up in the future.

ybiquitous: I also think the boolean support is too much. And sorry, my previous comment didn't mention `all` unintentionally. So,

- Output a table for `TIMING={n}` and `n >= 1`
- Output a table for `TIMING=all`
- Do not output anything for other values

ryo-manba: Applied your suggestion!

[Remove default display size](https://github.com/stylelint/stylelint/pull/8108/commits/409f988ad8b1972d6d234b12215fc0c54430aa41)

image

lib/timing.mjs
```diff
@@ -0,0 +1,135 @@
...
+
+		total += t;
+
+		return [key, t];
+	});
+
+	rows.sort((a, b) => b[1] - a[1]);
+
+	// Limit the number of rows to display
+	const limitedRows = listSize === Number.POSITIVE_INFINITY ? rows : rows.slice(0, listSize);
```
Mouvedia, line 76:
```suggestion
	const limitedRows = listSize === Number.POSITIVE_INFINITY ? rows : rows.slice(0, Math.min(listSize, rows));
```

lib/timing.mjs
```diff
@@ -0,0 +1,135 @@
...
+		return 0;
+	}
+
+	if (TIMING.toLowerCase() === 'all') {
+		return Number.POSITIVE_INFINITY;
+	}
+
+	const parsed = Number.parseInt(TIMING, 10);
+
+	if (!isNaN(parsed) && parsed >= 1) {
```
ybiquitous, line 45: [suggestion] Minor. How about using `Number.isNaN()` instead of the global `isNaN()`? It's more robust, according to the [MDN document](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN#difference_between_number.isnan_and_global_isnan).

```suggestion
	if (!Number.isNaN(parsed) && parsed >= 1) {
```

lib/timing.mjs
```diff
@@ -0,0 +1,135 @@
...
+ */
+function time(key, fn) {
+	return function timedFunction(/** @type {any} */ ...args) {
+		const t = startTime();
+		const result = fn(...args);
+		const timeDiff = endTime(t);
+
+		if (enabled) {
+			data[key] = (data[key] ?? 0) + timeDiff;
+		}
```
ybiquitous, line 119: [suggestion] Since the `time` function is supposed to be called after the `enabled` check outside, the `enable` check inside seems unnecessary:

```js
if (timing.enabled) {
	return timing.time(ruleName, () => ruleFn(postcssRoot, postcssResult))();
}
```

```suggestion
		data[key] = (data[key] ?? 0) + timeDiff;
```

If we checked `enabled` inside too, we would have to optimize the `time` function like this:

```js
function time(key, fn) {
	if (!enabled) fn;

	return function timedFunction(/** @type {any} */ ...args) {
		// ...
```

lib/timing.mjs
ybiquitous, line 76: [suggestion] Minor. I think we can now simplify the line to avoid the redundant `Number.POSITIVE_INFINITY` check. How about this? 

```suggestion
	const limitedRows = rows.slice(0, listSize);
```

lib/timing.mjs
```diff
@@ -0,0 +1,135 @@
...
+const listSize = getListSize();
+const enabled = listSize !== 0;
+
+/**
+ * Display the timing data.
+ * @param {{ [key: string]: number }} data Data object to be displayed.
+ * @returns {void}
+ * @private
+ */
+function display(data) {
```
ybiquitous, line 61: [suggestion] How about adding a test case for the `display` function like `getListSize()`?

ybiquitous (commented):
Thanks for the changes. Overall LGTM, but I've left a few suggestions to improve the code maintainability.

lib/timing.mjs
ryo-manba: I tried not to waste slices from a performance standpoint, but I don't have a strong preference.
Let's prioritize clarity.

lib/timing.mjs
ryo-manba: Fixed it!
[Refactor applied suggestion](https://github.com/stylelint/stylelint/pull/8108/commits/6cfe6382bc03d70138000c8eaeac609628733eee)

lib/timing.mjs
ryo-manba: Fixed it!
[Refactor applied suggestion](https://github.com/stylelint/stylelint/pull/8108/commits/6cfe6382bc03d70138000c8eaeac609628733eee)

lib/timing.mjs
ryo-manba: Fixed it!
[Refactor applied suggestion](https://github.com/stylelint/stylelint/pull/8108/commits/6cfe6382bc03d70138000c8eaeac609628733eee)

ybiquitous (approved):
Thanks LGTM 👍🏼 

I probably recommend asking @jeddy3 to review it again 😃

jeddy3 (approved):
LGTM.

@ryo-manba, thank you for persevering with the requested changes. The feature is better because of them.

@Mouvedia and @ybiquitous, thank you for your attention to detail regarding code quality. Your efforts here and in other PRs have benefited the entire codebase!

ryo-manba:
@ybiquitous 
Thanks a lot for your thoughtful review!

@jeddy3 
I would be happy if you could review this when you have time.

PR #8110, opened 2024-11-04
Fix `selector-pseudo-element-no-unknown` false positives for `::scroll-marker` and `::scroll-marker-group`

Mouvedia:

N/A

- ["Layout support for ::column::scroll-marker"](https://issues.chromium.org/issues/365680822#comment11)
- https://issues.chromium.org/issues/332396355
- [explainer](https://github.com/flackr/carousel/tree/main/scroll-marker)
- [`::scroll-marker-group`](https://drafts.csswg.org/css-overflow-5/#scroll-marker-group-pseudo)
- [`::scroll-marker`](https://drafts.csswg.org/css-overflow-5/#scroll-marker-pseudo)

`::column` will be added separately

### DIFF
--- a/.changeset/spotty-dancers-divide.md
+++ b/.changeset/spotty-dancers-divide.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `selector-pseudo-element-no-unknown` false positives for `::scroll-marker` and `::scroll-marker-group`
--- a/lib/reference/selectors.cjs
+++ b/lib/reference/selectors.cjs
@@ -195,6 +195,8 @@ const pseudoElements = uniteSets(
 		'highlight',
 		'marker',
 		'placeholder',
+		'scroll-marker',
+		'scroll-marker-group',
 		'selection',
 		'shadow',
 		'slotted',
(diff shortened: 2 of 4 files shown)


romainmenke:
Maybe a bit early, I am unsure.
There are no signals from other browser vendors, no caniuse or mdn.

Should we wait a bit longer or mark them as experimental?

If they are experimental (I think they are) then we should have the leeway to remove them again without needing to do a semver major in case the names change or the feature is abandoned.

romainmenke:
I am unsure what the consensus was in this thread: https://github.com/stylelint/stylelint/issues/7510#issuecomment-2012874488

If people agree that adding and removing experimental features for `no-uknown` rules is fine, then this PR is obviously also fine :)

Mouvedia:
> we should have the leeway to remove them again without needing to do a semver major in case the names change or the feature is abandoned

👍

> I am unsure what the consensus was in this thread: https://github.com/stylelint/stylelint/issues/7510#issuecomment-2012874488

Right now it's [behind a flag](https://chromestatus.com/feature/5192332683771904).
I opened #7510 because some of our users are early adopters and that I think that being behind a flag should have the same status as prefixes back then.
i.e. it warrants addition if it's functional

romainmenke:
Right, given that #7510 never reached a conclusion and wasn't formalized in our documentation I think it is fine to add this change.

PR #8115, opened 2024-11-09
Change deprecation warning to link to `https://stylelint.io/awesome-stylelint#formatters`

### LINKED ISSUE #8111: Change `detail` sentence of `GITHUB_FORMATTER` deprecation warning

A lot of linters have their own built-in GitHub formatter, and this is also the case with stylelint

However, the GitHub formatter has been deprecated and it suggests to switch to a third-party alternative which prints deprecation messages, it is not trivial to configure (apparently), and appears to not be maintained

Reconsider the deprecation, or advise for an actively maintained alternative

The alternative,

- https://github.com/xt0rted/stylelint-actions-formatters

should be used in conjunction with (or entirely replaced by, this is not clear)

https://github.com/xt0rted/stylelint-problem-matcher

does not seem maintained. 

I didn't find a way to make it work in a GitHub action. I have some test cases here:

Ref: https://github.com/diowa/ruby3-rails7-bootstrap-heroku/pull/984

I've asked for some clarifications one month ago at xt0rted/stylelint-problem-matcher#592 and also commented at the original request to make it the official suggested formatter at https://github.com/xt0rted/stylelint-actions-formatters/issues/187#issuecomment-2399118878

---

https://github.com/stylelint/stylelint/blob/13933eda2c6430ea2c595f02f47799adbda771df/lib/standalone.mjs#L114

ryo-manba:

Closes #8111 

No, it's self-explanatory.

### DIFF
--- a/.changeset/thick-seas-fail.md
+++ b/.changeset/thick-seas-fail.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Changed: deprecation warning to suggest `@csstools/stylelint-formatter-github`
--- a/.changeset/thick-seas-fail.md
+++ b/.changeset/thick-seas-fail.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Changed: deprecation warning to suggest `@csstools/stylelint-formatter-github`
+Changed: deprecation warning to link to `https://stylelint.io/awesome-stylelint#formatters`
--- a/lib/standalone.mjs
+++ b/lib/standalone.mjs
@@ -111,7 +111,7 @@ export default async function standalone({
 		emitDeprecationWarning(
 			'"github" formatter is deprecated.',
 			'GITHUB_FORMATTER',
-			'Please use "stylelint-actions-formatters" instead.',
+			'Please use "@csstools/stylelint-formatter-github" instead.',
 		);
 	}
 
--- a/lib/standalone.mjs
+++ b/lib/standalone.mjs
@@ -111,7 +111,7 @@ export default async function standalone({
 		emitDeprecationWarning(
 			'"github" formatter is deprecated.',
 			'GITHUB_FORMATTER',
-			'See https://github.com/stylelint/awesome-stylelint/blob/main/README.md#formatters for alternative formatters.',
+			'See https://stylelint.io/awesome-stylelint#formatters for alternative formatters.',
 		);
 	}
 
--- a/lib/standalone.mjs
+++ b/lib/standalone.mjs
@@ -111,7 +111,7 @@ export default async function standalone({
 		emitDeprecationWarning(
 			'"github" formatter is deprecated.',
 			'GITHUB_FORMATTER',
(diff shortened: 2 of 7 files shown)


lib/standalone.mjs
```diff
@@ -111,7 +111,7 @@ export default async function standalone({
 		emitDeprecationWarning(
 			'"github" formatter is deprecated.',
 			'GITHUB_FORMATTER',
-			'Please use "stylelint-actions-formatters" instead.',
+			'Please use "@csstools/stylelint-formatter-github" instead.',
```
romainmenke, line 114: [question]

Should we maybe link to the list here: https://github.com/stylelint/awesome-stylelint/blob/main/README.md#formatters

Then users can pick the one they prefer and we won't need to update the message if a better formatter appears in the future.

Maybe something like:

```
See https://github.com/stylelint/awesome-stylelint/blob/main/README.md#formatters for alternative formatters
```

ryo-manba: Good suggestion! 
Linking to the list makes it more maintainable.

ryo-manba: Fixed in this commit. 
[update message](https://github.com/stylelint/stylelint/pull/8115/commits/306b314cbd1e843feec346a8035c3063354e1e2c)

lib/standalone.mjs
```diff
@@ -111,7 +111,7 @@ export default async function standalone({
 		emitDeprecationWarning(
 			'"github" formatter is deprecated.',
 			'GITHUB_FORMATTER',
-			'Please use "stylelint-actions-formatters" instead.',
+			'See https://github.com/stylelint/awesome-stylelint/blob/main/README.md#formatters for alternative formatters.',
```
Mouvedia, line 114:
```suggestion
			'See https://stylelint.io/awesome-stylelint#formatters for alternative formatters.',
```

tagliala:
Thanks for this PR, and apologies for not contributing directly. I was planning to allocate some time tomorrow.

lib/standalone.mjs
ryo-manba: Applied in this commit.
[Replace url](https://github.com/stylelint/stylelint/pull/8115/commits/eb758e3d66bf42b4e3908171b7fb3622e2a902ef)

Mouvedia (approved):

.changeset/thick-seas-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Changed: deprecation warning to suggest `@csstools/stylelint-formatter-github`
```
romainmenke, line 5:
```suggestion
Changed: deprecation warning to link to `https://stylelint.io/awesome-stylelint#formatters`
```

lib/standalone.mjs
romainmenke: Thank you!

romainmenke (approved):
Thank you. LGTM 👍

PR #8116, opened 2024-11-10
Fix `custom-property-no-missing-var-function` false negatives for fallback values within `var()`

### LINKED ISSUE #8099: Fix `custom-property-no-missing-var-function` false negatives for fallback values within `var()`

```css
height: var(--primary-height, --fallback-height);
```

```json
{
  "rules": {
    "custom-property-no-missing-var-function": true
  }
}
```

```shell
stylelint “**/*.css” —config myconfig.json
```

```json
{
        "stylelint": "^16.10.0",
        "stylelint-config-prettier-scss": "^1.0.0",
        "stylelint-config-standard-scss": "^13.1.0",
        "stylelint-declaration-strict-value": "^1.10.6",
}
```

Rule should have flagged the missing `var()` syntax on the fallback value.  The correct syntax would be:

```css
height: var(--primary-height, var(--fallback-height));
```

### What actually happened?

The test passed despite the missing `var()`

_No response_

ryo-manba:

Closes #8099 

No, it's self-explanatory.


lib/rules/custom-property-no-missing-var-function/index.cjs
```diff
@@ -65,29 +65,47 @@ const rule = (primary) => {
...
+		/**
+		 * @param {import('postcss-value-parser').Node} node
+		 * @param {import('postcss').Declaration} decl
+		 */
+		function check(node, decl) {
+			if (node.type === 'function') {
+				let args = node.nodes;
+
+				if (isVarFunction(node)) {
+					args = node.nodes.slice(1);
```
ryo-manba, line 84: Check fallback values when using `var` function.

lib/rules/custom-property-no-missing-var-function/index.mjs
```diff
@@ -62,29 +62,45 @@ const rule = (primary) => {
...
 
-			valueParser(value).walk((node) => {
-				if (isVarFunction(node)) return false;
+			valueParser(value).nodes.forEach((childNode) => {
+				check(childNode, decl);
+			});
+		});
+
+		/**
+		 * @param {import('postcss-value-parser').Node} node
```
Mouvedia, line 71: nit

Maybe share the same typedef for this line and l108
i.e. `/** @typedef {import('postcss-value-parser').Node} Node  */`

```suggestion
		 * @param {Node} node
```

lib/rules/custom-property-no-missing-var-function/__tests__/index.mjs
```diff
@@ -32,6 +32,10 @@ testRule({
 			code: 'a { color: var(--foo, red); }',
 			description: 'undeclared dashed-idents in var with fallback',
 		},
+		{
+			code: 'a { color: var(--foo, var(--bar)); }',
```
src-code, line 36: It's possible for fallback values to also have fallbacks so would be good to have a test for that as well to ensure they're recursively tested, eg:

```css 
a { color: var(--foo, var(--bar, var(--foobar, #000))); }
```

lib/rules/custom-property-no-missing-var-function/index.mjs
ryo-manba: Fixed it!
[chore: unify types](https://github.com/stylelint/stylelint/pull/8116/commits/c59d67c56dc847a58f661281563cfa2d5ae2c9af)

lib/rules/custom-property-no-missing-var-function/__tests__/index.mjs
ryo-manba: Thanks! I applied your suggestion.
[chore: add test for nested fallbacks](https://github.com/stylelint/stylelint/pull/8116/commits/b459db880c16a8525c7b3de41da85ec5c1af8b86)

jeddy3:
@ryo-manba Thanks for working on this. When you have a moment, can you please resolve the conflict? Then we'll merge this.

ryo-manba:
@jeddy3 I've fixed it!

jeddy3 (approved):
LGTM, thank you!

Issue #8100, opened 2024-11-01
Integrate `@csstools/css-syntax-patches-for-csstree` into `declaration-property-value-no-unknown`


romainmenke:
I've finished the npm package containing the syntax definition patches as scraped from all CSS specifications. 

https://github.com/csstools/postcss-plugins/tree/main/packages/css-syntax-patches-for-csstree#css-syntax-patches-for-csstree

1. we can integrate this directly into `declaration-property-value-no-unknown`
2. or we can leave it up to end users to use this package and maybe link to it from the rule docs

I think direct integration gives a better user experience for must users as they can set a simpler plugin option.

For example: 

```js
{
  "rules": {
    "declaration-property-value-no-unknown": [true, {lib: "next"}]

  }
}
```

<details>
<summary>(Demo implementation)</summary>

```js
import { createRequire } from 'node:module';

import { find, fork, parse, string } from 'css-tree';

import { isAtRule, isDeclaration } from '../../utils/typeGuards.mjs';
import { isRegExp, isString } from '../../utils/validateTypes.mjs';
import declarationValueIndex from '../../utils/declarationValueIndex.mjs';
import getDeclarationValue from '../../utils/getDeclarationValue.mjs';
import isCustomProperty from '../../utils/isCustomProperty.mjs';
import isStandardSyntaxDeclaration from '../../utils/isStandardSyntaxDeclaration.mjs';
import isStandardSyntaxProperty from '../../utils/isStandardSyntaxProperty.mjs';
import isStandardSyntaxValue from '../../utils/isStandardSyntaxValue.mjs';
import matchesStringOrRegExp from '../../utils/matchesStringOrRegExp.mjs';
import { nestingSupportedAtKeywords } from '../../reference/atKeywords.mjs';
import report from '../../utils/report.mjs';
import ruleMessages from '../../utils/ruleMessages.mjs';
import validateObjectWithArrayProps from '../../utils/validateObjectWithArrayProps.mjs';
import validateObjectWithProps from '../../utils/validateObjectWithProps.mjs';
import validateOptions from '../../utils/validateOptions.mjs';

const require = createRequire(import.meta.url);

// TODO: Replace this with Import attributes when dropping the support for older Node.js versions.
// See https://nodejs.org/api/esm.html#import-attributes
const syntaxPatches = require('@csstools/css-syntax-patches-for-csstree');

const ruleName = 'declaration-property-value-no-unknown';

const messages = ruleMessages(ruleName, {
[...]

PR #8124, opened 2024-11-13
Add `atRulesSyntax: {}` and `cssWideKeywords: []` options in `declaration-property-value-no-unknown`

### LINKED ISSUE #8113: Add `atRulesSyntax: {}` and `cssWideKeywords: []` options in `declaration-property-value-no-unknown`

`csstree` is sometimes a bit behind the most recent CSS features resulting in false positives.

For `properties` and `types` we can already pass updated definitions as rule options.
However `atRules` and `cssWideKeywords` extensions are not exposed.

I want to be able to extend any part of the CSS syntax definitions in `csstree`.

New plugin options:
- `atRulesSyntax: {}` [^1]
- `cssWideKeywords: []` [^2]

This will allow users to extend any part of the CSS syntax definitions.

[^1]: https://github.com/stylelint/stylelint/issues/8100#issuecomment-2462386218
[^2]: https://github.com/stylelint/stylelint/pull/8105#discussion_r1829506849

ryo-manba:

Closes #8113 

No, it's self-explanatory.


lib/rules/declaration-property-value-no-unknown/index.mjs
```diff
@@ -46,6 +47,8 @@ const rule = (primary, secondaryOptions) => {
 					ignoreProperties: [validateObjectWithArrayProps(isString, isRegExp)],
 					propertiesSyntax: [validateObjectWithProps(isString)],
 					typesSyntax: [validateObjectWithProps(isString)],
+					atRulesSyntax: [validateAtRulesSyntax],
```
Mouvedia, line 50: Most of our `RuleOptionsPossibleFunc` functions are rudimentary and not specialized.
i.e. can be part of utils

I am not sure if we want to start a trend which introduces validators that deeply check the shape of objects.
I don't have a strong opinion because we have `--validate` to balance any hypothetical performance hit.
I would prefer #7608 though.

Mouvedia:
@ryo-manba you maybe have to add `"@types/css-tree": "^2.3.9"` as a dev dependency to package.json?

lib/rules/declaration-property-value-no-unknown/README.md
```diff
@@ -139,3 +139,145 @@ a { top: --foo(0); }
 ```css
 a { top: --foo(10px); }
 ```
+
+### `atRulesSyntax: { "at-rule": { "prelude": "syntax", "descriptors": { "descriptor": "syntax" } } }`
```
ryo-manba, line 143: I'm unsure if it's necessary to go into this much detail.
Validation is not strictly based on this comment.
https://github.com/stylelint/stylelint/pull/8124#discussion_r1841089623

lib/rules/declaration-property-value-no-unknown/README.md
```diff
@@ -139,3 +139,145 @@ a { top: --foo(0); }
...
+
+Extend or alter the at-rules syntax dictionary. This option allows you to define custom at-rules or modify existing ones by specifying their expected syntax. Both `prelude` and `descriptors` are optional. The `prelude` is the part of the at-rule that comes after the at-rule name and before the block (if any). The `descriptors` are like properties specific to the at-rule, and are defined similarly to CSS properties.
+
+[CSS Value Definition Syntax](https://github.com/csstree/csstree/blob/master/docs/definition-syntax.md) is used to define the syntax of at-rules, their preludes, and descriptors. If a definition starts with `|` it is added to the existing definition value if any. See the [CSS syntax reference](https://csstree.github.io/docs/syntax/) for default definitions.
```
ryo-manba, line 147: I refer to the following docs.
https://github.com/csstree/stylelint-validator/blob/master/README.md#atrules

lib/rules/declaration-property-value-no-unknown/README.md
```diff
@@ -139,3 +139,138 @@ a { top: --foo(0); }
...
+```css
+a { width: unknown-value; }
+```
+
+<!-- prettier-ignore -->
+```css
+a { color: invalid-keyword; }
+```
+
+## Examples
```
Mouvedia, line 218: From this point on, it seems like duplication.

lib/rules/declaration-property-value-no-unknown/index.mjs
Mouvedia: @ybiquitous what's your opinion?

lib/rules/declaration-property-value-no-unknown/README.md
ryo-manba: Oh, I was inattentive. Fixed it.
[remove duplicate content](https://github.com/stylelint/stylelint/pull/8124/commits/a7581669c00d38e435f4ef3212f84463aab7458c)

jeddy3:
@ryo-manba Thank you for working on this PR. It has uncovered a larger problem with the scope/naming of this rule.

Am I right in thinking that we ultimately want to check that:

- property names are known (currently [`property-no-unknown`](https://stylelint.io/user-guide/rules/property-no-unknown/) using [`known-css-properties`](https://www.npmjs.com/package/known-css-properties))
- properties have known values (currently [`declaration-property-value-no-unknown`](https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown/) using [csstree](https://github.com/csstree/csstree))
- at-rule names are known (currently [`at-rule-no-unknown`](https://stylelint.io/user-guide/rules/at-rule-no-unknown/) using [our own reference](https://github.com/stylelint/stylelint/blob/main/lib/reference/atKeywords.mjs))
- at-rules have known preludes (this PR adds to `declaration-property-value-no-unknown`)
- at-rules have known descriptors (currently not checked)
- descriptors have known values (currently [`declaration-property-value-no-unknown`](https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown/)) 

And for known values and preludes, we have overlapping rules:

- [`color-no-invalid-hex`](https://stylelint.io/user-guide/rules/color-no-invalid-hex/)
- [`function-no-unknown`](https://stylelint.io/user-guide/rules/function-no-unknown/)
- [`string-no-newline`](https://stylelint.io/user-guide/rules/string-no-newline/)
- [`unit-no-unknown`](https://stylelint.io/user-guide/rules/unit-no-unknown/)

We provide different ways for users to customise the syntax or ignore unknown things, e.g. [`function-no-unknown`'s `ignorefunctions: []`](https://stylelint.io/user-guide/rules/function-no-unknown/#ignorefunctions-regex-regex-non-regex) and [`declaration-property-value-no-unknown`'s `propertiesSyntax: []`](https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown/#propertiessyntax--property-syntax-).

The [`declaration-property-value-no-unknown`](https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown/) has been in use for over a year and seems to be working well and is likely to get more robust, thanks to @romainmenke's good work in https://github.com/stylelint/stylelint/issues/8100. However, we don't currently include it in our recommended config, and it's likely many users aren't benefitting from the type-checking this rule provides. As the rule is currently marked as experimental, we can smooth off the rough edges before rolling it out more broadly.

Shall we:

- expand the rule's scope to check all types
- rename it accordingly, perhaps `no-unknown-types` or `no-unknown-syntax`
- deprecate the overlapping rules
- update the recommended config to use the expanded rule (and remove the deprecated ones)
- release the config at the same time as the deprecations to minimise disruption to users

The rule's name already miscommunicates the current scope, as the rule checks descriptors and properties but returns the same message for both. This is problematic for things like `font-weight` where the [`font-weight` property](https://csstree.github.io/docs/syntax/#Property:font-weight) accepts `<font-weight-absolute> | bolder | lighter` whereas the [`font-weight` descriptor](https://csstree.github.io/docs/syntax/#AtruleDescriptor:font-face%2Ffont-weight) accepts `<font-weight-absolute>{1,2}`.

Having a single rule for types will make it easier for users to customise the syntax in one place rather than spreading (and sometimes duplicating) the customisation across the secondary options of several rules.

I think our users will benefit greatly from easy-to-use type-checking. I may have missed some details, but @romainmenke may known more from working on https://github.com/stylelint/stylelint/issues/8100.

For example, given:

```css
@property foo {}

@counter-style bar {
  margin: 0;
}

@font-face {
  size-adjust: red;
  font-weight: 100 200;
}

a {
  font-weight: 100 200;
}
```

We can flag:

- `foo` [should be](https://csstree.github.io/docs/syntax/#Atrule:property) a `<custom-property-name>`
- `margin` is [an unknown descriptor](https://csstree.github.io/docs/syntax/#Atrule:counter-style) for `@counter-style`
- `red` is an [unknown value](https://csstree.github.io/docs/syntax/#AtruleDescriptor:font-face%2Fsize-adjust) for the `size-adjust` descriptor
- `100 200` is an [unknown value](https://csstree.github.io/docs/syntax/#Property:font-weight) for the `font-weight` property [but not for](https://csstree.github.io/docs/syntax/#AtruleDescriptor:font-face%2Ffont-weight) the `font-weight` descriptor

Mouvedia:
> the rule checks descriptors and properties
> at-rules have known descriptors (currently not checked)

Maybe it was an error of mine but `property-no-unknown` do support at least one descriptor.
see #7751

---

@jeddy3 since this rule is still experimental, can we move on with this PR and then create an issue for your proposal?
i.e. having a meta rule for all unknown rules

e.g.
```json
{
  "rules": {
		"no-unknown": {
			"syntax": {},
			"property": {},
			"animation": {},
			"mediaFeatureName": {},
			"mediaFeatureNameValue": {},
			"unit": {},
			"atRule": {},
			"function": {},
			"annotation": {},
			"selector": {
				"type": {},
				"pseudoClass": {},
				"pseudoElement": {},
			},
			"custom": {
				"media": {},
				"property": {}
			}
		}
  }
}
```

What you are saying is right and makes sense but to achieve it will require a lot of effort.
i.e. adding the options in the short term is better than having to wait 3 or 4 minor versions for this new rule to be implemented

package.json
```diff
@@ -173,7 +173,7 @@
     "@stylelint/prettier-config": "^3.0.0",
     "@stylelint/remark-preset": "^5.1.1",
     "@types/balanced-match": "^1.0.4",
-    "@types/css-tree": "^2.3.8",
+    "@types/css-tree": "^2.3.9",
```
Mouvedia, line 176: After https://github.com/stylelint/stylelint/pull/8129/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R176, it's not necessary anymore.

jeddy3:
> since this rule is still experimental, can we move on with this PR and then create an issue for your proposal?

I've opened [an issue](https://github.com/stylelint/stylelint/issues/8133) to discuss the broader proposal. I suggest we don't proceed with this PR as-is because checking and customising at-rule preludes and descriptors in a rule called `declaration-property-value-no-unknown` doesn't align with our conventions and may confuse users.

It may be that we limit this PR to adding the `cssWideKeywords` option, but let's see what comes out of the discussion first.

Mouvedia:
Sorry @ryo-manba, this PR might be in limbo for a moment until #8133 gets a resolution.

jeddy3:
@ryo-manba Thank you for working on this PR.

Shall we close it in favour of the approach outlined in https://github.com/stylelint/stylelint/issues/8133?

We can use some of the work here as the foundation for https://github.com/stylelint/stylelint/issues/8154.

ryo-manba:
Okay, let's close this.

Issue #8097, opened 2024-10-29
Refactor to improve utilities to get source location indices.


romainmenke:
See :
- https://github.com/stylelint/stylelint/pull/8066
- https://github.com/stylelint/stylelint/pull/8066#discussion_r1821578388

This change adds a couple of new utilities that get the source location indicies of various parts of AST nodes.

Having these as shared utilities has several advantages:
- shared code
- unit tests

But the utilities themselves are often quite small.
Sometimes as simple as this:

```js
/**
 * @param {import('postcss').AtRule} atRule
 * @returns {number}
 */
export default function atRuleRawsAfterNameIndex(atRule) {
	// Initial 1 is for the `@`
	return 1 + atRule.name.length;
}
```

This code is clearly trivial, but knowing that you need this code is not trivial.

I really like having these utilities.
But I can understand that the number of files and their small contents can feel out of balance.

Maybe this can be improved?

Maybe a single file with all of the utilities that have this same purpose?

ybiquitous:
Thanks for the proposal. Like [`lib/utils/validateTypes.mjs`](https://github.com/stylelint/stylelint/blob/42cd4757369706178fa2600dcca054b068682bd9/lib/utils/validateTypes.mjs), if assembling similar functions makes sense, we should do that. 👍🏼 

Also, such a refactoring should have almost no risks.

Mouvedia:
> if assembling similar functions makes sense, we should do that

👍

Mouvedia:
It seems we are all in favor of having one file that group all of these related utilities; hence I have labeled this issue as ready to implement.
You can bikeshed the file's name in the upcoming PR.

PR #8125, opened 2024-11-13
Refactor to group functions that calculate node field indices

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/8097

No, it's self-explanatory.


romainmenke:
See: https://github.com/stylelint/stylelint/pull/8066#pullrequestreview-2428518722

@ybiquitous said:

> For the declaration utils, does it make sense to combine them to one function? 

I'd rather not have a string argument to indicate the requested field.
That requires answering what should happen on an unexpected input values.

imho individual functions are more expressive and self documenting.

But I can see the appeal from the caller's perspective:

```js
/* this looks pretty neat */
const before = declarationComponentIndex(node, 'before');
const after =  declarationComponentIndex(node, 'after');
```

So happy to make changes if there is a preference.

lib/rules/function-url-quotes/index.mjs
```diff
@@ -1,5 +1,5 @@
-import atRuleParamIndex from '../../utils/atRuleParamIndex.mjs';
-import declarationValueIndex from '../../utils/declarationValueIndex.mjs';
+import { atRuleParamIndex } from '../../utils/nodeFieldIndices.mjs';
+import { declarationValueIndex } from '../../utils/nodeFieldIndices.mjs';
```
Mouvedia, line 2:
```suggestion
import { atRuleParamIndex, declarationValueIndex } from '../../utils/nodeFieldIndices.mjs';
```

lib/utils/__tests__/nodeFieldIndices.test.mjs
```diff
@@ -0,0 +1,261 @@
...
+describe('declarationRawsBetweenIndex', () => {
+	it('has no spaces before the colon', () => {
+		expect(declarationRawsBetweenIndex(decl('a { a:b }'))).toBe(1);
+	});
+
+	it('has a space before the colon', () => {
+		expect(declarationRawsBetweenIndex(decl('a { a :b}'))).toBe(1);
+	});
+
+	it('has multiple characters before the colon', () => {
```
Mouvedia, line 141: nit

```suggestion
	it('has multiple spaces before the colon', () => {
```

romainmenke: spaces are characters :)

lib/rules/function-url-quotes/index.mjs
romainmenke: Thank you, good catch!

lib/utils/nodeFieldIndices.mjs
```diff
@@ -0,0 +1,118 @@
...
+ * @returns {number}
+ */
+export function atRuleParamIndex(atRule) {
+	let index = atRuleRawsAfterNameIndex(atRule);
+
+	if (atRule.raws.afterName) {
+		index += atRule.raws.afterName.length;
+	}
+
+	return index;
```
ybiquitous, line 19: [suggestion] This is just my preference, but how about shortening the function body code like this?

```suggestion
	const index = atRuleRawsAfterNameIndex(atRule);

	return index + (atRule.raws.afterName?.length ?? 0);
```

lib/utils/nodeFieldIndices.mjs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/**
+ * @param {AtRule} node
+ * @returns {number}
+ */
+export function atRuleRawsAfterIndex(node) {
```
ybiquitous, line 26: [suggestion] As well as other at-rule functions, renaming to `atRule` seems better in terms of consistency:

```suggestion
export function atRuleRawsAfterIndex(atRule) {
```

lib/utils/nodeFieldIndices.mjs
```diff
@@ -0,0 +1,118 @@
...
+	// subtract 1 for `}`
+
+	if (!node.source?.end?.offset) return node.toString().length - 1;
+
+	if (!node.raws?.after?.length) return node.source.end.offset - 1;
+
+	return node.source.end.offset - (node.raws.after.length + 1);
```
ybiquitous, line 33: [suggestion] To make the code more readable and reduce the evaluation count, how about extracting to local variables like this?

```suggestion
	const endOffset = node.source?.end?.offset;

	if (!endOffset) return node.toString().length - 1;

	const afterLength = node.raws?.after?.length;

	if (!afterLength) return endOffset - 1;

	return endOffset - (afterLength + 1);
```

lib/utils/nodeFieldIndices.mjs
ybiquitous, line 1: [suggestion] IMO, the "Raws" word in these new function names (e.g., `atRuleRawsAfterNameIndex`) seems redundant because "raws" looks like a kind of internal expression around PostCSS. From the perspective of Stylelint, it doesn't seem to need to be aware of it.

For example, see the `atRuleParamIndex` and `declarationValueIndex` functions. They also use "raws" internally, but they don't include it in their name.

This is my personal opinion, so I'd be glad to know your thoughts. 🙏🏼

lib/utils/nodeFieldIndices.mjs
```diff
@@ -0,0 +1,118 @@
...
+ * @returns {number}
+ */
+export function declarationRawsBetweenIndex(decl) {
+	const raws = decl.raws;
+	const prop = raws.prop;
+
+	return [
+		isObject(prop) && 'prefix' in prop && prop.prefix,
+		(isObject(prop) && 'raw' in prop && prop.raw) || decl.prop,
+		isObject(prop) && 'suffix' in prop && prop.suffix,
```
ybiquitous, line 64: [suggestion] How about memoizing the function result? It may also improve the code readability a bit.

```suggestion
	const propIsObject = isObject(prop);

	return [
		propIsObject && 'prefix' in prop && prop.prefix,
		(propIsObject && 'raw' in prop && prop.raw) || decl.prop,
		propIsObject && 'suffix' in prop && prop.suffix,
```

lib/utils/nodeFieldIndices.mjs
```diff
@@ -0,0 +1,118 @@
...
+	].reduce((/** @type {number} */ count, str) => {
+		if (isString(str)) {
+			return count + str.length;
+		}
+
+		return count;
+	}, 0);
```
ybiquitous, line 71: [suggestion] How about extracting a function from this `reduce()` part for readability? For example:

```js
/**
 * @param {unknown[]} values
 * @returns {number}
 */
function countChars(values) {
	return values.reduce((/** @type {number} */ count, value) => {
		if (isString(value)) return count + value.length;

		return count;
	}, 0);
}
```

Would change to:

```js
	return countChars([
		propIsObject && 'prefix' in prop && prop.prefix,
		(propIsObject && 'raw' in prop && prop.raw) || decl.prop,
		propIsObject && 'suffix' in prop && prop.suffix,
	]);
```

lib/utils/nodeFieldIndices.mjs
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * @param {Declaration} decl
+ * @returns {number}
+ */
+export function declarationValueIndex(decl) {
+	const raws = decl.raws;
+
+	return (
+		declarationRawsBetweenIndex(decl) +
+		[raws.between || ':', raws.value && 'prefix' in raws.value && raws.value.prefix].reduce(
```
ybiquitous, line 85: [suggestion] How about extracting variables to shorten the code a bit?

```suggestion
	const { between, value } = decl.raws;

	return (
		declarationRawsBetweenIndex(decl) +
		[between || ':', value && 'prefix' in value && value.prefix].reduce(
```

lib/utils/nodeFieldIndices.mjs
```diff
@@ -0,0 +1,118 @@
...
+ */
+export function ruleRawsBetweenIndex(rule) {
+	return getRuleSelector(rule).length;
+}
+
+/**
+ * @param {Rule} node
+ * @returns {number}
+ */
+export function ruleRawsAfterIndex(node) {
```
ybiquitous, line 110: [suggestion] I have the same suggestion for the argument name consistency:

```suggestion
export function ruleRawsAfterIndex(rule) {
```

lib/utils/nodeFieldIndices.mjs
```diff
@@ -0,0 +1,118 @@
...
+	// subtract 1 for `}`
+
+	if (!node.source?.end?.offset) return node.toString().length - 1;
```
ybiquitous, line 113: [suggestion] I have the same suggestion to extract local variables also here:

```suggestion
	const endOffset = node.source?.end?.offset;

	if (!endOffset) return node.toString().length - 1;
```

ybiquitous (commented):
@romainmenke Thank you for the pull request. I love this grouping! 👍🏼 

I've left some minor suggestions, so it'd be great to have a look at them. 🙏🏼

lib/utils/nodeFieldIndices.mjs
romainmenke: I agree, this is better 🙇 

I was being too strict in following the PostCSS implementation details when naming these.

romainmenke:
Thank you for reviewing this and for the helpful feedback!
I've applied the suggestions.

ybiquitous (approved):
Thank you. LGTM 😄

PR #8144, opened 2024-11-25
Fix `font-family-no-missing-generic-family-keyword` false positives for `font-family` descriptor in `@font-palette-values`

### LINKED ISSUE #8139: Fix `font-family-no-missing-generic-family-keyword` false positives for `font-family` descriptor in `@font-palette-values`

```css
@font-palette-values --example {
  font-family: example;
  override-colors: 0 red;
}

```

```json
"rules": {
  "font-family-no-missing-generic-family-keyword": true
}
```

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIACAzCA7ALgWgA4EMAbGTTGbANyIFcYBnAAm2xgA98BbXYh4AHXQMGaLNhScAloQCe8Bm07cYAbgFCIFGACctE2NkiEIWunIAMDLTCir0AXxAAaECikwAcpziIFXYgDowOjoncAxXAHMEEH50Pkw+EC1qYjpEuVj4rJcMHHEOKWlsdAhsAuCJdAjsCJh0bQkwMUkZbABrGGkAd2ModIZMZJgBeLsBMfRQyHRIgDFjDnxMaIArOgxQ2FwQxFihRLpMaWJCSoSEBkTCJfpz8ZA7IA)

```json
  "stylelint": "^16.10.0",
  "stylelint-config-standard-scss": "^13.1.0",
```

`font-family-no-missing-generic-family-keyword` should ignore the absence of generic font-family keyword when defining font-palette-values, since the color palette definition is only valid when used for color font. 

### What actually happened?

The following problem was reported:

```
2:16-23
ERROR Unexpected missing generic font family (font-family-no-missing-generic-family-keyword)
```

_No response_

grommasdietz:

Closes #8139

No, it's self-explanatory.

### DIFF
--- a/lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.mjs
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.mjs
@@ -54,7 +54,7 @@ testRule({
 			code: "@font-palette-values --foo { font-family: bar; }",
 		},
 		{
-			code: "@FONT-PALETTE-VALUES -foo { font-family: bar; }",
+			code: "@FONT-PALETTE-VALUES --foo { font-family: bar; }",
 		},
 		{
 			code: 'a { font: inherit; }',
--- a/lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.mjs
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.mjs
@@ -51,6 +51,12 @@ testRule({
 			code: '@FONT-FACE { font-family: Gentium; }',
 		},
 		{
+			code: "@font-palette-values --foo { font-family: bar; }",
+		},
+		{
+			code: "@FONT-PALETTE-VALUES -foo { font-family: bar; }",
+		},
+		{
 			code: 'a { font: inherit; }',
 		},
 		{
--- a/lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.mjs
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.mjs
@@ -51,10 +51,10 @@ testRule({
 			code: '@FONT-FACE { font-family: Gentium; }',
 		},
 		{
-			code: "@font-palette-values --foo { font-family: bar; }",
+			code: '@font-palette-values --foo { font-family: bar; }',
 		},
 		{
-			code: "@FONT-PALETTE-VALUES --foo { font-family: bar; }",
+			code: '@FONT-PALETTE-VALUES --foo { font-family: bar; }',
 		},
 		{
 			code: 'a { font: inherit; }',
--- a/lib/rules/font-family-no-missing-generic-family-keyword/index.cjs
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/index.cjs
@@ -63,10 +63,14 @@ const rule = (primary, secondaryOptions) => {
 		}
 
 		root.walkDecls(/^font(-family)?$/i, (decl) => {
-			// Ignore @font-face
+			// Ignore @font-face and @font-palette-values
 			const parent = decl.parent;
 
 			if (parent && typeGuards.isAtRule(parent) && parent.name.toLowerCase() === 'font-face') {
+				return;
+			}
+
+			if (parent && typeGuards.isAtRule(parent) && parent.name.toLowerCase() === 'font-palette-values') {
 				return;
 			}
 
--- a/lib/rules/font-family-no-missing-generic-family-keyword/index.cjs
+++ b/lib/rules/font-family-no-missing-generic-family-keyword/index.cjs
@@ -62,12 +62,12 @@ const rule = (primary, secondaryOptions) => {
 			return;
 		}
 
-		root.walkDecls(/^font(-family)?$/i, (decl) => {
-			// Ignore @font-face and @font-palette-values
-			const parent = decl.parent;
-			const rules = ['font-face', 'font-palette-values'];
+		const ignoredAtRules = new Set(['font-face', 'font-palette-values']);
 
-			if (parent && typeGuards.isAtRule(parent) && rules.includes(parent.name.toLowerCase())) {
+		root.walkDecls(/^font(-family)?$/i, (decl) => {
+			const parent = decl.parent;
+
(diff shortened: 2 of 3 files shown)


lib/rules/font-family-no-missing-generic-family-keyword/index.cjs
```diff
@@ -63,13 +63,17 @@ const rule = (primary, secondaryOptions) => {
...
+			// Ignore @font-face and @font-palette-values
 			const parent = decl.parent;
 
 			if (parent && typeGuards.isAtRule(parent) && parent.name.toLowerCase() === 'font-face') {
 				return;
 			}
 
+			if (parent && typeGuards.isAtRule(parent) && parent.name.toLowerCase() === 'font-palette-values') {
+				return;
+			}
```
Mouvedia, line 75:
```suggestion
			const parent = decl.parent;
			const rules = ['font-face', 'font-palette-values'];

			if (parent && typeGuards.isAtRule(parent) && rules.includes(parent.name.toLowerCase())) {
				return;
			}
```

lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.mjs
```diff
@@ -50,6 +50,12 @@ testRule({
 		{
 			code: '@FONT-FACE { font-family: Gentium; }',
 		},
+		{
+			code: "@font-palette-values --foo { font-family: bar; }",
+		},
+		{
+			code: "@FONT-PALETTE-VALUES -foo { font-family: bar; }",
```
Mouvedia, line 57: question

Why `-foo` and not `--foo`?

grommasdietz:
Added missing rules on mjs. Think we are done. Adapted the term `rules` as @Mouvedia suggested, but maybe `exceptions` would be more precisely? (Don’t know how it’s handled in other parts of stylelint, though).

In general I tried to clone the repository and edited locally with Visual Studio Code, sadly the lint/prettier setup  did not work out for me. Neither in VS Code, nor via Terminal. It e.g. always used double instead of single quotation marks. So I ended up transferring all the changes manually to open the pull request. Not sure if I did something wrong, I could get it work by manually referring to the npm package, but the terminal build process still ended up in wrong formatting.

Would be happy to contribute in the future, but don’t want to crash with wrong formats or transferring suggestions manually.

lib/rules/font-family-no-missing-generic-family-keyword/index.cjs
```diff
@@ -63,10 +63,10 @@ const rule = (primary, secondaryOptions) => {
 		}
 
 		root.walkDecls(/^font(-family)?$/i, (decl) => {
-			// Ignore @font-face
+			// Ignore @font-face and @font-palette-values
 			const parent = decl.parent;
-
-			if (parent && typeGuards.isAtRule(parent) && parent.name.toLowerCase() === 'font-face') {
+			const rules = ['font-face', 'font-palette-values'];
```
Mouvedia, line 68: `npm run build` should have added the missing newline for you.

grommasdietz: v23.3.0, but as I said, I can’t really use `npm run format` and `npm run build`, since it’s formatting wrongly. Should I switch node version, does this help?

Mouvedia: > Should I switch node version, does this help?

I was just curious; 23 is fine.

> …since it’s formatting wrongly

I don't know what's going on, but since there is only one file to edit, what about doing it manually?
i.e.
```suggestion
			const rules = ['font-face', 'font-palette-values'];

```

lib/rules/font-family-no-missing-generic-family-keyword/__tests__/index.mjs
```diff
@@ -50,6 +50,12 @@ testRule({
 		{
 			code: '@FONT-FACE { font-family: Gentium; }',
 		},
+		{
+			code: "@font-palette-values --foo { font-family: bar; }",
+		},
+		{
+			code: "@FONT-PALETTE-VALUES --foo { font-family: bar; }",
+		},
```
Mouvedia, line 58:
```suggestion
		{
			code: '@font-palette-values --foo { font-family: bar; }',
		},
		{
			code: '@FONT-PALETTE-VALUES --foo { font-family: bar; }',
		},
```

lib/rules/font-family-no-missing-generic-family-keyword/index.mjs
```diff
@@ -60,10 +60,11 @@ const rule = (primary, secondaryOptions) => {
 		}
 
 		root.walkDecls(/^font(-family)?$/i, (decl) => {
```
ybiquitous, line 62:
```suggestion
		const ignoredAtRules = new Set(['font-face', 'font-palette-values']);

		root.walkDecls(/^font(-family)?$/i, (decl) => {
```

lib/rules/font-family-no-missing-generic-family-keyword/index.mjs
```diff
@@ -60,10 +60,11 @@ const rule = (primary, secondaryOptions) => {
...
-			// Ignore @font-face
+			// Ignore @font-face and @font-palette-values
```
ybiquitous, line 63: [suggestion]

```suggestion
```

lib/rules/font-family-no-missing-generic-family-keyword/index.mjs
```diff
@@ -60,10 +60,11 @@ const rule = (primary, secondaryOptions) => {
...
 			const parent = decl.parent;
+			const rules = ['font-face', 'font-palette-values'];
```
ybiquitous, line 65: [suggestion]

```suggestion
```

ybiquitous (commented):
Thanks for the pull request!

I've left some refactoring suggestions, although they're not critical. Can you have a look?

1. Move the ignored at-rules array creation outside the walk loop for performance.
2. Give a clearer name to the array variable for readability.
3. Convert the array to a set for performance.
4. Remove the redundant code comment for readability.

jeddy3 (approved):
@grommasdietz Thank you for the PR. 

Merging so we can include the fix in our upcoming release, which I'm about to start.

Issue #1957, opened 2016-10-06
False positives for CSS2 System Color keywords in value-keyword-case


ntwb:
> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

I came across this today in relation to CSS2 System Color keyword case:
> via https://www.w3.org/TR/CSS22/ui.html#system-colors
> "Although these are case-insensitive, it is recommended that the mixed capitalization shown below be used, to make the names more legible."

Based on that recommendation should these [~28 system colors](https://www.w3.org/wiki/CSS/Properties/color/keywords#System_Colors) cases be ignored?
(_Note: CSS System Colors are deprecated in CSS3_)

Possible Solutions:

1. Add a new `value-keyword-case` rule option `ignoreSystemColors` option with the ~28 keywords added to a new `systemColors` keyword set.
2. Add a new keyword set `systemColors` and ignore all ~28 by default.
3. Do nothing and have users use the existing `value-keyword-case` rule option `ignoreKeywords` to manually add the ~28 system colors to a "whitelist"

> Which rule, if any, is this issue related to?

`value-keyword-case`

> What CSS is needed to reproduce this issue?

```css
.foo {
	outline: 1px dotted ButtonText;
}
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "rules": {
    "value-keyword-case": "lower",
  }
}
```

> Which version of stylelint are you using?

`7.3.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

n/a

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No

> What did you expect to happen?

No warnings to be flagged.

> What actually happened (e.g. what warnings or errors you are getting)?

The following warnings were flagged:

```shell
test.css
1:21  Expected "ButtonText" to be "buttontext"             value-keyword-case
```

evilebottnawi:
@ntwb sgtm

jeddy3:
Lets go with option 2 - "Add a new keyword set systemColors and ignore all ~28 by default.". Deprecated or not, they're still valid values.

It's makes the rule more permissive by default, so isn't a breaking change.

PR #8146, opened 2024-11-28
Fix `value-keyword-case` false positives for vendor-prefixed system colors

Mouvedia:

#1957

- `-ms-`: https://adrianroselli.com/2021/02/whcm-and-system-colors.html#CSS2WHCM
- `-moz-`: https://developer.mozilla.org/en-US/docs/Web/CSS/Mozilla_Extensions#color_keywords

---

I didn't add non-system colors.
e.g. `-moz-activehyperlinktext`, `-moz-default-background-color`, `-moz-default-color`, `-moz-hyperlinktext`, `-moz-visitedhyperlinktext`, etc.


lib/reference/keywords.mjs
```diff
@@ -296,7 +296,49 @@ export const namedTimelineRangeKeywords = new Set([
 	'exit-crossing',
 ]);
 
-export const systemColorsKeywords = new Set([
+const prefixedSystemColorKeywords = new Set([
```
ybiquitous, line 299: [suggestion] Can you add a code comment with links to web pages to maintain this list?
E.g. https://developer.mozilla.org/en-US/docs/Web/CSS/Mozilla_Extensions#color_keywords

ybiquitous (approved):
Thanks, overall LGTM. I've left one suggestion.

lib/reference/keywords.cjs
```diff
@@ -300,7 +300,49 @@ const namedTimelineRangeKeywords = new Set([
 	'exit-crossing',
 ]);
 
-const systemColorsKeywords = new Set([
+const prefixedSystemColorKeywords = new Set([
```
Mouvedia, line 303:
```suggestion
// https://developer.mozilla.org/en-US/docs/Web/CSS/Mozilla_Extensions#color_keywords
const prefixedSystemColorKeywords = new Set([
```

lib/reference/keywords.mjs
Mouvedia:
```suggestion
// https://developer.mozilla.org/en-US/docs/Web/CSS/Mozilla_Extensions#color_keywords
const prefixedSystemColorKeywords = new Set([
```

PR #8157, opened 2024-11-30
Fix `custom-property-no-missing-var-function` false positives for `container-name`

Mouvedia:

N/A

https://developer.mozilla.org/en-US/docs/Web/CSS/container-name#browser_compatibility

### DIFF
--- a/.changeset/fair-shirts-love.md
+++ b/.changeset/fair-shirts-love.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `custom-property-no-missing-var-function` false positives for `container-name`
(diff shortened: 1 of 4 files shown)


jeddy3 (commented):
Should we also ignore the [`container`](https://drafts.csswg.org/css-conditional-5/#container-shorthand) shorthand property?

---

MDN [reference for dashed-ident container names](https://developer.mozilla.org/en-US/docs/Web/CSS/container-name), for those interested:

> The dashed ident intended to denote author-defined identifiers (e.g., `--container-name`) is permitted.

Mouvedia:
@jeddy3 I chose not to because it has an optional `container-type` which accepted values are `normal | [ [ size | inline-size ] || scroll-state ]`.
i.e. `custom-property-no-missing-var-function` may be useful for that part
e.g. `container: bar-qux / --foo;`

jeddy3 (approved):
LGTM, thank you.

PR #2445, opened 2017-03-27
Add at-page support to selector-pseudo-class-no-unknown

evilebottnawi:
> Which issue, if any, is this issue related to?

#1871

> Is there anything in the PR that needs further explanation?

No, it's self explanatory.

If the solution is appropriate, then I will add similar logic to other rules :smile:

PR #8170, opened 2024-12-04
Fix `selector-pseudo-class-no-unknown` false positives for `:recto`, `:verso` and `:nth()`

Mouvedia:

#2445

https://www.princexml.com/doc/css-at-rules/#nested-at-rules
https://pagedjs.org/documentation/5-web-design-for-print/#pseudo-class-selectors-for-pages
https://github.com/Kozea/WeasyPrint/pull/846

https://drafts.csswg.org/css-logical/#page
https://www.w3.org/TR/css-gcpm-3/#document-page-selectors

### DIFF
--- a/.changeset/curly-lies-serve.md
+++ b/.changeset/curly-lies-serve.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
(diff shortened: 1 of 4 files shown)


jeddy3 (approved):
Thank you!

Issue #1871, opened 2016-08-30
Apply selector-pseudo-class-*-list rules to at-page


alexander-akait:
- [x] `selector-pseudo-class-no-unknown`
- [x] `selector-pseudo-class-disallowed-list`
- [ ] `selector-pseudo-class-allowed-list`

> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

https://developer.mozilla.org/en-US/docs/Web/CSS/:first
https://developer.mozilla.org/en-US/docs/Web/CSS/:right
https://developer.mozilla.org/en-US/docs/Web/CSS/:left
https://developer.mozilla.org/en-US/docs/Web/CSS/:blank

cf https://github.com/stylelint/stylelint/blob/dfd1ffcabb2036d9e2e25aed512c193d3d868ae5/lib/reference/selectors.js#L197

> Which rule, if any, is this issue related to?

`selector-pseudo-class-*`

> What CSS is needed to reproduce this issue?

e.g.

``` css
@page :first {
  margin-left: 50%;
  margin-top: 50%;
}
```

> What stylelint configuration is needed to reproduce this issue?

No matter.

> Which version of stylelint are you using?

No matter.

> How are you running stylelint: CLI, PostCSS plugin, Node API?

No matter.

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

No matter.

> What did you expect to happen?

Check `pseudo-classes` in this rules

> What actually happened (e.g. what warnings or errors you are getting)?

Now we have not check.

jeddy3:
Have you run into a real world need for this?

evilebottnawi:
@jeddy3 yep, for print styles, also this include in spec 2.1, https://developer.mozilla.org/ru/docs/Web/CSS/@page

jeddy3:
> also this include in spec 2.1,

Oh wow, I had in mind that this was something new. In fact, it's spec 2.1 and [with excellent browser support](https://developer.mozilla.org/en/docs/Web/CSS/@page#Browser_compatibility).

LGTM for `future-major`.

davidtheclark:
This should be finished for v8 or closed.

jeddy3:
@evilebottnawi Am I right in thinking the following rules also need  at-page support?:

- `selector-pseudo-class-blacklist`
- `selector-pseudo-class-case`
- `selector-pseudo-class-parentheses-space-inside`
- `selector-pseudo-class-whitelist`

[...]

PR #8171, opened 2024-12-05
Add `selector-pseudo-class-disallowed-list` support for `@page`

Mouvedia:

#1871

1. `:blank` references

https://developer.mozilla.org/en-US/docs/Web/CSS/:blank
https://caniuse.com/mdn-css_selectors_-moz-only-whitespace
https://drafts.csswg.org/selectors-4/#blank-pseudo
web-platform-tests/interop#595
https://www.npmjs.com/package/css-blank-pseudo

2. I tried using a combination of `walkAtRules` and `walkRules` instead of `walk`, the performance hit was negligible.

### DIFF
--- a/.changeset/fresh-seahorses-grab.md
+++ b/.changeset/fresh-seahorses-grab.md
@@ -2,4 +2,4 @@
 "stylelint": minor
 ---
 
-Added: `@page` support to `selector-pseudo-class-disallowed-list`
+Added: `selector-pseudo-class-disallowed-list` now checks `@page` at-rules and supports `@page` pseudo-classes
--- a/.changeset/fresh-seahorses-grab.md
+++ b/.changeset/fresh-seahorses-grab.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: `@page` support to `selector-pseudo-class-disallowed-list`
--- a/lib/reference/selectors.mjs
+++ b/lib/reference/selectors.mjs
@@ -237,6 +237,7 @@ const vendorSpecificPseudoClasses = new Set([
 	'-moz-meter-optimum',
 	'-moz-meter-sub-optimum',
 	'-moz-meter-sub-sub-optimum',
+	'-moz-only-whitespace',
 	'-moz-placeholder',
 	'-moz-submit-invalid',
 	'-moz-suppressed',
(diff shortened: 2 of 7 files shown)


lib/reference/selectors.mjs
```diff
@@ -293,6 +294,10 @@ export const pseudoClasses = uniteSets(
 		'active-view-transition-type',
 		'any-link',
 		'autofill',
+		/*
+			introduced the 10/05/2016 by 8c808d0
+			as of 2024 it is still not supported by any browser
+		*/
 		'blank',
```
Mouvedia, line 301: Having both [`:blank`](https://developer.mozilla.org/en-US/docs/Web/CSS/:blank) and [`:blank`](https://drafts.csswg.org/css-page/#blank-pseudo)—line 219—could have been problematic if both were supported by at least one browser.
i.e. `['blank']` wouldn't differentiate between the two, which would be detrimental for developer

---

I am not removing it because #7510 hasn't been resolved yet.
i.e. removal policy has not been proposed
I have added its mozilla prefixed counterpart though.

.changeset/fresh-seahorses-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: `@page` support to `selector-pseudo-class-disallowed-list`
```
jeddy3, line 5:
```suggestion
Added: `selector-pseudo-class-disallowed-list` support for `@page`
```

We typically lead with the rule name so that changes to a rule are grouped in the changelog.

jeddy3 (commented):
Thank you.

Minor suggestion, otherwise LGTM.

.changeset/fresh-seahorses-grab.md
Mouvedia:
```suggestion
Added: `selector-pseudo-class-disallowed-list` now checks `@page` at-rules and supports `@page` pseudo-classes
```

copied from https://github.com/stylelint/stylelint/blob/dacc742d1ae55a9b6cef2035132c968d9057d91d/CHANGELOG.md?plain=1#L1409

jeddy3 (approved):
LGTM, thank you.

lib/rules/selector-pseudo-class-disallowed-list/index.mjs
```diff
@@ -26,51 +30,69 @@ const rule = (primary) => {
...
+				case 'atrule': {
+					if (node.name !== 'page') return;
+
+					if (!isStandardSyntaxAtRule(node)) return;
+
+					const params = getAtRuleParams(node);
+
+					if (!params.includes(':')) return;
+
+					const offset = `@page${node.raws.afterName ?? ''}`.length;
```
Mouvedia, line 74: I should have used `atRuleParamIndex`.

PR #8172, opened 2024-12-05
Fix `function-no-unknown` and `custom-property-no-missing-var-function` false positives for `running()`

Mouvedia:

#5742

https://www.w3.org/TR/css-gcpm-3/#running-syntax
https://www.princexml.com/doc/styling/#position
https://www.antenna.co.jp/AHF/help/en/ahf-css6.html#css3-functions

### DIFF
--- a/.changeset/silly-brooms-beg.md
+++ b/.changeset/silly-brooms-beg.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `function-no-unknown` false positives for `running`
--- a/.changeset/silly-brooms-beg.md
+++ b/.changeset/silly-brooms-beg.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `function-no-unknown` false positives for `running`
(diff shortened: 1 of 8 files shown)


.changeset/silly-brooms-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `function-no-unknown` false positives for `running`
```
jeddy3, line 5:
```suggestion
Fixed: `function-no-unknown` false positives for `running()`
```

jeddy3 (commented):
Thank you.

A minor suggestion, otherwise LGTM.

.changeset/silly-brooms-beg.md
Mouvedia: I was matching the previous changesets.
i.e. 078d7f8, 4d32d3

jeddy3 (approved):
LGTM, thank you.

> I was matching the previous changesets.

Oh, right. Either would've been fine then.

PR #8176, opened 2024-12-07
Add `selector-pseudo-class-allowed-list` `@page` support

### LINKED ISSUE #1871: Apply selector-pseudo-class-*-list rules to at-page

Mouvedia:

Closes #1871

I am having a hard time getting consistent results out of the _benchmark-rule_ script.
i.e. walkRules/WalkAtRules vs walk/switch

### DIFF
--- a/lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
+++ b/lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
@@ -166,6 +166,34 @@ testRule({
 
 testRule({
 	ruleName,
+	config: '/^nth/',
+
+	accept: [
+		{
+			code: 'a {}',
+		},
+		{
+			code: 'a:nth-child(5) {}',
+		},
+		{
+			code: 'a:nth-LAST-child {}',
+		},
+	],
+
+	reject: [
+		{
+			code: 'a:hover {}',
+			message: messages.rejected(':hover'),
+			line: 1,
+			column: 2,
+			endLine: 1,
+			endColumn: 8,
+		},
+	],
+});
+
+testRule({
+	ruleName,
 	config: ['-moz-first-node'],
 
 	accept: [
@@ -193,12 +221,6 @@ testRule({
 	accept: [
 		{
 			code: 'a {}',
-		},
-		{
-			code: 'a:nth-child(5) {}',
-		},
-		{
-			code: 'a:nth-LAST-child {}',
 		},
 		{
 			code: '@page :nth(1) {}',
--- a/lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
+++ b/lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
@@ -223,6 +223,9 @@ testRule({
 			code: 'a {}',
 		},
 		{
+			code: '@page foo {}',
+		},
+		{
 			code: '@page :nth(1) {}',
 		},
 	],
--- a/lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
+++ b/lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
@@ -220,7 +220,7 @@ testRule({
 
 	accept: [
 		{
-			code: 'a {}',
+			code: '@page :nth(1) {}',
 		},
 	],
 
@@ -232,6 +232,33 @@ testRule({
 			column: 2,
 			endLine: 1,
 			endColumn: 8,
+		},
+		{
+			code: '@page :first {}',
+			message: messages.rejected(':first'),
+			line: 1,
+			column: 7,
+			endLine: 1,
+			endColumn: 13,
+		},
+		{
+			code: '@page:blank:first {}',
+			warnings: [
+				{ message: messages.rejected(':blank'), line: 1, column: 6, endLine: 1, endColumn: 12 },
+				{ message: messages.rejected(':first'), line: 1, column: 12, endLine: 1, endColumn: 18 },
+			],
+			skip: true,
+		},
+		{
+			code: stripIndent`
+				@page
+				:first {}
+			`,
+			message: messages.rejected(':first'),
+			line: 2,
+			column: 1,
+			endLine: 2,
+			endColumn: 7,
 		},
 	],
 });
--- a/lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
+++ b/lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
@@ -166,34 +166,6 @@ testRule({
 
 testRule({
 	ruleName,
-	config: '/^nth/',
-
-	accept: [
-		{
-			code: 'a {}',
-		},
-		{
-			code: 'a:nth-child(5) {}',
-		},
-		{
-			code: 'a:nth-LAST-child {}',
-		},
-	],
-
-	reject: [
-		{
-			code: 'a:hover {}',
-			message: messages.rejected(':hover'),
-			line: 1,
-			column: 2,
-			endLine: 1,
-			endColumn: 8,
-		},
-	],
-});
-
-testRule({
-	ruleName,
 	config: ['-moz-first-node'],
 
 	accept: [
@@ -219,6 +191,15 @@ testRule({
 	config: /^nth/,
 
 	accept: [
+		{
+			code: 'a {}',
+		},
+		{
+			code: 'a:nth-child(5) {}',
+		},
+		{
+			code: 'a:nth-LAST-child {}',
+		},
 		{
 			code: '@page :nth(1) {}',
 		},
--- a/lib/rules/selector-pseudo-class-allowed-list/index.mjs
+++ b/lib/rules/selector-pseudo-class-allowed-list/index.mjs
@@ -1,5 +1,8 @@
 import { isRegExp, isString } from '../../utils/validateTypes.mjs';
+import { atRuleParamIndex } from '../../utils/nodeFieldIndices.mjs';
+import getAtRuleParams from '../../utils/getAtRuleParams.mjs';
 import getRuleSelector from '../../utils/getRuleSelector.mjs';
+import isStandardSyntaxAtRule from '../../utils/isStandardSyntaxAtRule.mjs';
 import isStandardSyntaxRule from '../../utils/isStandardSyntaxRule.mjs';
 import matchesStringOrRegExp from '../../utils/matchesStringOrRegExp.mjs';
 import parseSelector from '../../utils/parseSelector.mjs';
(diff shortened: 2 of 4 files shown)


lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
```diff
@@ -233,6 +233,33 @@ testRule({
...
+			endColumn: 13,
+		},
+		{
+			code: '@page:blank:first {}',
+			warnings: [
+				{ message: messages.rejected(':blank'), line: 1, column: 6, endLine: 1, endColumn: 12 },
+				{ message: messages.rejected(':first'), line: 1, column: 12, endLine: 1, endColumn: 18 },
+			],
+			skip: true,
+		},
```
Mouvedia, line 251: see https://drafts.csswg.org/css-page/#example-691ff5b9
i.e. it seems like `walkAtRules` is buggy

romainmenke: How is it buggy?
What happens, what do you expect instead?

lib/rules/selector-pseudo-class-allowed-list/index.mjs
```diff
@@ -26,55 +31,62 @@ const rule = (primary) => {
...
-			parseSelector(getRuleSelector(ruleNode), result, ruleNode)?.walkPseudos((pseudoNode) => {
-				const value = pseudoNode.value;
-				const index = pseudoNode.sourceIndex;
-				const endIndex = index + pseudoNode.value.length;
+		/**
+		 * @param {string} selector
+		 * @param {AtRule | Rule} node
+		 */
+		function check(selector, node) {
+			const offset = node.type === 'atrule' ? atRuleParamIndex(node) : 0;
```
romainmenke, line 41: In other `check` functions we pass the function to calculate the index as an argument.

Can we do the same here?

lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
```diff
@@ -164,34 +164,6 @@ testRule({
...
-		{
-			code: 'a:hover {}',
-			message: messages.rejected(':hover'),
-			line: 1,
-			column: 2,
-			endLine: 1,
-			endColumn: 8,
-		},
-	],
-});
```
romainmenke, line 193: Can we keep these tests as they were?
I am unsure why these are fine to remove/alter.

romainmenke (commented):
Thank you @Mouvedia 

I've left some questions and a suggestion, please take a look when you have time.

lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
Mouvedia: Check the tests again.
i.e. they were duplicates

lib/rules/selector-pseudo-class-allowed-list/index.mjs
Mouvedia: I am not sure why it would be desirable.
I mean if I understand correctly the 3rd argument would be a function which means to get `0`, I would have to pass `() => 0` correct?
i.e. `const offset = fn(node);`

Or did you mean pass _the result_ of the function?

lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
Mouvedia: I am not really sure, I didn't investigate much.
I would expect `@page:blank:first {}` to be valid based on that link and hence that `walkAtRules` to recognizes it as an `atrule` node and not have to skip this test.

lib/rules/selector-pseudo-class-allowed-list/index.mjs
romainmenke: No, pass the function itself.
Like we do in:

https://github.com/stylelint/stylelint/blob/2b603bbc55dc8ece7ac7022e907d415604b1fa8c/lib/rules/string-no-newline/index.mjs#L31-L36

lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
romainmenke: I think these are subtly different from other tests.
For example, we don't have other tests with a regexp as a string in the rule option.

lib/rules/selector-pseudo-class-allowed-list/index.mjs
Mouvedia: So you want us to pass `() => 0` for the `rule` case?

lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
romainmenke: Seems like a bug in PostCSS.

For: `@page:blank:first {}`

https://astexplorer.net/

Screenshot 2024-12-07 at 09 34 36


vs.

https://tabatkins.github.io/parse-css/example.html

Screenshot 2024-12-07 at 09 34 53

------

While `@media(color) {}` does produce the expected outcome.
Same for `@page#foo {}`.

But `@page.foo {}` and `@page&.foo` also seem wrong.

So I guess that the character set used by PostCSS to detect the end of the `name` is incorrect.

lib/rules/selector-pseudo-class-allowed-list/index.mjs
romainmenke: Yeah, that is indeed annoying.
Fine to keep as is, we also have rules using this exact code.

lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
Mouvedia: I see.

We can move on with this PR and we will come back to unskip the test in the eventual dependabot PR that will update `postcss`.

lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
Mouvedia: Please compare line 189–193 and 167–169.

i.e.
```
testRule({
	ruleName,
	config: /^nth/,
```

Mouvedia: You are right my bad.

romainmenke: `/^nth/` vs. `'/^nth/'`
Maybe this is fine? But it also doesn't hurt to have extra coverage, right?

It seems like an unnecessary change that isn't directly related to the change at hand, so I would prefer it if these tests were added back, if only to simplify review :)

lib/rules/selector-pseudo-class-allowed-list/__tests__/index.mjs
Mouvedia: Just a last question, what about `walkAtRules(/^page/i, () => …` instead?

romainmenke: I'd rather not :) Upstream bugs should be fixed upstream. And no one is writing `@page:first` without a separating space so there isn't much reason to add workarounds. Better to immediately fix it upstream.

Mouvedia: I am scared that minifier do write it like that though.

romainmenke: No one should be linting the output of minifiers 🤔 
If anyone is affected by this (no one will be) then they can file an issue.
But as I said before, there is a clear path to fixing this, upstream, in PostCSS.

lib/rules/selector-pseudo-class-allowed-list/index.mjs
```diff
@@ -26,55 +31,61 @@ const rule = (primary) => {
...
+
+			const selector = getRuleSelector(ruleNode);
+
+			if (!selector.includes(':')) return;
+
+			check(selector, ruleNode);
+		});
+
+		root.walkAtRules('page', (atRuleNode) => {
+			if (!isStandardSyntaxAtRule(atRuleNode)) return;
```
romainmenke, line 77: Can we add a test with a non-standard at rule?

lib/rules/selector-pseudo-class-allowed-list/index.mjs
```diff
@@ -26,55 +31,61 @@ const rule = (primary) => {
...
+
+			const params = getAtRuleParams(atRuleNode);
+
+			if (!params.includes(':')) return;
```
romainmenke, line 81: Can we also have a test with `@page wide {}`

romainmenke (commented):
Thank you for applying the feedback @Mouvedia 
I've left two suggestions to improve code coverage.

lib/rules/selector-pseudo-class-allowed-list/index.mjs
Mouvedia: I think that codecov is being overzealous with these warnings.
i.e. it would be akin to adding cases to `isStandardSyntaxAtRule.test.mjs` directly

If you have a relevant test to add to `/selector-pseudo-class-allowed-list/__tests__/index.mjs` please provide it.

---

To be clear I think that new tests could probably be added to `isStandardSyntaxAtRule.test.mjs` but I won't add them in this PR.
e.g. [margin at-rules](https://drafts.csswg.org/css-page/#margin-at-rules)

romainmenke: I think it could be fine to add a single test with non standard syntax.
Not to test the utility itself, but to encode that this rule doesn't support non-standard syntax in at rules.

But I don't think this is blocking :)

romainmenke (approved):
Thank you. LGTM 👍

jeddy3 (approved):
LGTM, thanks.

Issue #8183, opened 2024-12-10
Fix `selector-pseudo-class-no-unknown` false positives for `:open`


lukewarlow:
### What is the problem you're trying to solve?

If you use `:open` in your styles, stylelint will complain that it's an unknown pseudo class.

See demo: https://stylelint.io/demo/#N4Igxg9gJgpiBcICWA7ADgVwC7wmmKABMADoolYBGEATrDQLSQA2t8hNMUA3GQL4gANCABmSZjAByAQwC2cRDAAectBIB0YAM5ah4CCjEBzBCFJFCJEMqwEoWq+wDaZQm8sgtWAJ4TmqLCYDYwYvaRQoaTorV0IAXX49SEMkIwAxWllpLFMAKy0DPVg0XURzNysvXxh-FCxHD2ZsmC8rQVjKnz8AoJSjUKxwyOiERubWkES+IA

### What solution would you like to see?

It would be good if stylelint added support for `:open` pseudo class, so it didn't get flagged as an unknown pseudo.

See [CSS spec here](https://drafts.csswg.org/selectors-4/#open-state) and [HTML spec here](https://html.spec.whatwg.org/multipage/semantics-other.html#selector-open) for details on this new pseudo class.

Mouvedia:
related: #7510

skyclouds2001:
I've also fired https://github.com/mdn/data/pull/953 to fix it in upstream

PR #8184, opened 2024-12-10
Fix `selector-pseudo-class-no-unknown` false positives for `:open`

lukewarlow:

Closes https://github.com/stylelint/stylelint/issues/8183

See issue for more details on this new pseudo class. See https://groups.google.com/a/chromium.org/g/blink-dev/c/Pdo4oOXYKcA for the Chromium intent to ship for this.

### DIFF
--- a/.changeset/tiny-lobsters-float.md
+++ b/.changeset/tiny-lobsters-float.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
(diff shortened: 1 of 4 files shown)


jeddy3 (approved):
Thank you, LGTM.

(And thanks for including links to the relevant parts of the CSS and HTML specs in your original issue (https://github.com/stylelint/stylelint/issues/8183))

Issue #8148, opened 2024-11-29
Add `at-rule-descriptor-no-unknown`


jeddy3:
### What is the problem you're trying to solve?

A dedicated rule to flag unknown descriptors within at-rules.

For example, `margin` is not a known descriptor for `@counter-style`:

```css
@counter-style foo { margin: 0; }
```

### What solution would you like to see?

Suggested blueprint:

- Name: `at-rule-descriptor-no-unknown`
- Description: Disallow unknown descriptors for at-rules.
- Message: `Unexpected unknown descriptor "${descriptor}" for at-rule "${at-rule}"`
- Primary option: `true`
- Secondary options: n/a
- Category: Avoid errors > Unknown
- Included in sharable configs?: Yes, recommended.
- Expanded description:

This rule considers values for properties defined within the CSS specifications to be known.

Certain at-rules only allow specific descriptors. This rule checks that. For example, only the `syntax`, `inherits` and 
`initial-value` descriptors are known to the `@property` at-rule.

ybiquitous:
This idea sounds good to me 👍🏼 

But I have a concern only about the naming consistency. We already have the [`at-rule-property-required-list`](https://github.com/stylelint/stylelint/blob/ca8d5b429dadd0e0c401dd316d9e0d7da9d5cf7b/lib/rules/at-rule-property-required-list/README.md) rule, but the new rule would start with `at-rule-descriptor`. Are there some differences between `at-rule-property` and `at-rule-descriptor`?

jeddy3:
Yes, they are [different things](https://drafts.csswg.org/css-syntax/#declaration):

> Declarations are further categorized as property declarations or descriptor declarations, with the former setting CSS [properties](https://drafts.csswg.org/css-cascade-5/#css-property) and appearing most often in [qualified rules](https://drafts.csswg.org/css-syntax/#qualified-rule) and the latter setting CSS [descriptors](https://drafts.csswg.org/css-syntax/#css-descriptor), which appear only in [at-rules](https://drafts.csswg.org/css-syntax/#at-rule).

In hindsight, we probably should've named the `at-rule-property-required-list` rule `at-rule-descriptor-required-list`. I think we're fine with leaving it as it is to minimise disruption to users. Perhaps we can tweak the rule description to mention descriptors:

> Specify a list of required properties for an at-rule.

Becomes "Specify a list of required descriptors (previously labelled properties) for an at-rule."

[...]

PR #8185, opened 2024-12-11
Document `at-rule-property-required-list`  difference between properties and descriptors

ybiquitous:

This change was produced by the discussion on <https://github.com/stylelint/stylelint/issues/8148#issuecomment-2531031990>.

We probably should have named the `at-rule-property-required-list` as `at-rule-descriptor-required-list` because declarations in at-rules are semantically descriptors rather than properties.

See also <https://drafts.csswg.org/css-syntax/#declaration>.

### DIFF
--- a/lib/rules/at-rule-property-required-list/README.md
+++ b/lib/rules/at-rule-property-required-list/README.md
@@ -1,19 +1,19 @@
 # at-rule-property-required-list
 
-Specify a list of required properties for an at-rule.
+Specify a list of required descriptors (previously labelled properties) for an at-rule.
 
 <!-- prettier-ignore -->
 ```css
     @font-face { font-display: swap; font-family: 'foo'; }
 /**  ↑           ↑                   ↑
- *  At-rule and required property names */
+ *  At-rule and required descriptor names */
 ```
 
 The [`message` secondary option](../../../docs/user-guide/configure.md#message) can accept the arguments of this rule.
 
 ## Options
 
-`object`: `{ "at-rule-name": ["array", "of", "properties"]|"property" }`
+`object`: `{ "at-rule-name": ["array", "of", "descriptors"]|"descriptor" }`
 
 Given:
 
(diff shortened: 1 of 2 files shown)


ybiquitous:
@jeddy3 Do you think we should change the rule's message as well?

https://github.com/stylelint/stylelint/blob/8fe37ab4dae9b35a3aec25bdac120ea1593313d4/lib/rules/at-rule-property-required-list/index.mjs#L12

```diff
-Expected descriptor "${descriptor}" for at-rule "${atRule}"
+Expected descriptor "${descriptor}" for at-rule "${atRule}"
```

I'm a bit concerned about some impact on users. At least, we probably should make the change with another pull request.

lib/rules/at-rule-property-required-list/README.md
```diff
@@ -1,19 +1,19 @@
 # at-rule-property-required-list
 
-Specify a list of required properties for an at-rule.
+Specify a list of required descriptors (previously labelled properties) for an at-rule.
```
jeddy3, line 3:
```suggestion
Specify a list of required descriptors (previously labelled as properties) for an at-rule.
```

jeddy3 (commented):
I've left one minor suggestion, otherwise LGTM.

jeddy3:
> Do you think we should change the rule's message as well?

If it doesn't affect our public API, it should be fine to change the message as a fix.

Mouvedia:
> We probably should have named the `at-rule-property-required-list` as `at-rule-descriptor-required-list` because declarations in at-rules are semantically descriptors rather than properties.

Adding an alias is pretty easy.
I am not talking about deprecating the rule though; it would be too disrupting in my opinion.

ybiquitous:
Um... To be honest, I'm not sure if aliasing is easy. It would increase code, and more it might rather produce confusion. 🤔

Mouvedia (approved):

PR #8186, opened 2024-12-11
Fix `at-rule-property-required-list` message to use "descriptor" for accuracy

ybiquitous:

Follow-up PR #8185
See also <https://github.com/stylelint/stylelint/issues/8148#issuecomment-2531031990>

Note that it doesn't rename local variables to minimize code diffs and keep git-blame easy.


jeddy3 (approved):
Thank you, LGTM.

PR #8190, opened 2024-12-13
Fix `value-no-vendor-prefix` false negatives for vendor-prefixed `center`

Mouvedia:

N/A

https://bugzilla.mozilla.org/show_bug.cgi?id=1181130#c10
https://developer.mozilla.org/en-US/docs/Web/CSS/text-align

### DIFF
--- a/.changeset/clean-scissors-eat.md
+++ b/.changeset/clean-scissors-eat.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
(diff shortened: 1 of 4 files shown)


Mouvedia:
> `-moz-center` - Centers inline **and also block content**.
> `-moz-left` - Aligns inline **and also block content**.
> `-moz-right` - Aligns inline **and also block content**.

PR #8197, opened 2024-12-15
Add `at-rule-descriptor-no-unknown`

ryo-manba:
> Which issue, if any, is this issue related to?

Closes #8148 

> Is there anything in the PR that needs further explanation?

No, it's self-explanatory.

PR #8203, opened 2024-12-16
Fix `property-no-unknown` false positives for descriptors

### LINKED ISSUE #8155: Fix `property-no-unknown` false positives for descriptors

```json
{
  "rules": {
    "property-no-unknown": true
  }
}
```

1. **#8148 should be completed**
2. exceptions like
https://github.com/stylelint/stylelint/blob/4385d8fcbf47936f6a16236548fe106b2f9a99a3/lib/rules/property-no-unknown/index.mjs#L26-L27 
should be removed
i.e. decline #7751
3. an at-rule/descriptor map should be created for the descriptors that happen to be named after existing properties
4. the descriptors should be ignored

jeddy3:

Closes #8155 

Ref https://github.com/stylelint/stylelint/pull/8197

No, it's self-explanatory.


lib/rules/property-no-unknown/__tests__/index.mjs
```diff
@@ -378,23 +386,11 @@ testRule({
 			line: 1,
 			column: 21,
 		},
-		{
-			code: '@not-my-at-rule { foo: 1; }',
-			message: messages.rejected('foo'),
-			line: 1,
-			column: 19,
-		},
```
jeddy3, line 386: We can't tell whether declarations within custom at-rules are property or descriptor ones, so we now ignore these potential false positives.

lib/rules/property-no-unknown/__tests__/index.mjs
Mouvedia, line 1: l264–380
These are necessary anymore.

lib/rules/property-no-unknown/index.mjs
Mouvedia, line 1: l104
Should we remove [`ignoreAtRules`](https://stylelint.io/user-guide/rules/property-no-unknown#ignoreatrules-regex-regex-string)?

jeddy3:
@Mouvedia Thanks for taking a look. Can you please locate your feedback on particular lines using the tick in the square box? It makes it easier to understand what you're referencing. If possible, use suggestions for changes, including deletions, so I can commit them from here.

lib/utils/isDescriptorDeclaration.mjs
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Check whether a declaration is a descriptor one
+ *
+ * @param {import('postcss').Declaration} decl
+ * @returns {boolean}
+ */
+export default function isDescriptorDeclaration(decl) {
+	const { parent } = decl;
+
+	if (parent && isAtRule(parent) && !nestingSupportedAtKeywords.has(parent.name.toLowerCase()))
```
Mouvedia, line 13: question

if
```js
@position-try --custom-bottom-right {
  @supports (position-area: bottom right) {
    position-area: bottom right;
  }
}
```
is allowed this `if` won't be enough.
i.e. you will probably have to check all its parents which will probably degrade the perf

Mouvedia:
> Thanks for taking a look. Can you please locate your feedback on particular lines using the tick in the square box? It makes it easier to understand what you're referencing.

I couldn't because [l104](https://github.com/stylelint/stylelint/pull/8203#discussion_r1886845736) and [l264–380](https://github.com/stylelint/stylelint/pull/8203#discussion_r1886839918) were outside of the range of changes you have made.

lib/rules/property-no-unknown/__tests__/index.mjs
jeddy3: Those tests don't seem to be connected to this change?

lib/rules/property-no-unknown/index.mjs
jeddy3: It's still relevant but not to nesting supported at-rules.

We can update the tests and examples to reflect that.

Mouvedia, line 1: fix
major

l25–29
replace with `import { all as knownProperties } from 'known-css-properties';`

Mouvedia (changes requested):
see https://github.com/stylelint/stylelint/pull/8203/files#r1886886369

lib/rules/property-no-unknown/index.mjs
jeddy3: It's still relevant for properties within at-rules. We can remove the redundant tests, though.

lib/rules/property-no-unknown/__tests__/index.mjs
Mouvedia: > `{ ignoreAtRules: ['supports', /^my-/] }`

`@supports` being part of `nestingSupportedAtKeywords` we should at least remove 3 of the 5 accept tests.

lib/utils/isDescriptorDeclaration.mjs
jeddy3: Good catch, thanks!

I've pushed a commit to address. Let's favour accuracy over performance in this instance.

lib/utils/isDescriptorDeclaration.cjs
```diff
@@ -14,8 +14,15 @@ const atKeywords = require('../reference/atKeywords.cjs');
 function isDescriptorDeclaration(decl) {
 	const { parent } = decl;
 
-	if (parent && typeGuards.isAtRule(parent) && !atKeywords.nestingSupportedAtKeywords.has(parent.name.toLowerCase()))
-		return true;
+	/** @type {import('postcss').Node | undefined} */
+	let node = parent;
+
+	while (node && node.type !== 'root') {
```
Mouvedia, line 20: off-topic
remark

We have 4 rules with this condition; we could maybe add an `isNested` util.

lib/utils/isDescriptorDeclaration.mjs
```diff
@@ -0,0 +1,24 @@
...
+ * @returns {boolean}
+ */
+export default function isDescriptorDeclaration(decl) {
+	const { parent } = decl;
+
+	/** @type {import('postcss').Node | undefined} */
+	let node = parent;
+
+	while (node && node.type !== 'root') {
+		if (node && isAtRule(node) && !nestingSupportedAtKeywords.has(node.name.toLowerCase()))
```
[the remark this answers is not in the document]
Mouvedia: nit

```suggestion
		if (isAtRule(node) && !nestingSupportedAtKeywords.has(node.name.toLowerCase()))
```

lib/utils/isDescriptorDeclaration.cjs
```diff
@@ -0,0 +1,30 @@
...
+ * @returns {boolean}
+ */
+function isDescriptorDeclaration(decl) {
+	const { parent } = decl;
+
+	/** @type {import('postcss').Node | undefined} */
+	let node = parent;
+
+	while (node && node.type !== 'root') {
+		if (node && typeGuards.isAtRule(node) && !atKeywords.nestingSupportedAtKeywords.has(node.name.toLowerCase()))
```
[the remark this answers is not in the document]
Mouvedia:
```suggestion
		if (isAtRule(node) && !nestingSupportedAtKeywords.has(node.name.toLowerCase()))
```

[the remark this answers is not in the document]
Mouvedia:
```suggestion
		if (typeGuards.isAtRule(node) && !atKeywords.nestingSupportedAtKeywords.has(node.name.toLowerCase()))
)
```

lib/utils/isDescriptorDeclaration.cjs
ybiquitous: That idea could be likely. I have a rough idea for implementing more generic utilities. Like so:

```js
/** @import { Node } from 'postcss' */

/**
 * @param {Node} startNode
 * @param {(node: Node) => void | false} callback
 * @returns {void}
 */
function eachNodeUpToRoot(startNode, callback) {
	let currentNode = startNode.parent;

	while (currentNode && !isRoot(currentNode)) {
		if (callback(currentNode) === false) break;

		currentNode = currentNode.parent;
	}
}

/**
 * @param {Node} startNode
 * @param {(node: Node) => boolean} condition
 * @returns {Node | undefined}
 */
function findNodeUpToRoot(startNode, condition) {
	let found;

	eachNodeUpToRoot(startNode, (node) => {
		if (condition(node)) {
			found = node;

			return false;
		}
	});

	return found;
}
```

But I'm still unsure if adding them is worth it. Anyway, it's out of the scope of this PR. 😅

lib/utils/isDescriptorDeclaration.mjs
```diff
@@ -0,0 +1,23 @@
...
+ * @param {import('postcss').Declaration} decl
+ * @returns {boolean}
+ */
+export default function isDescriptorDeclaration(decl) {
+	const { parent } = decl;
+
+	/** @type {import('postcss').Node | undefined} */
+	let node = parent;
+
+	while (node && node.type !== 'root') {
```
ybiquitous, line 16: [suggestion] Optional. Using `isRoot` seems a bit more consistent with `isAtRule`:

```suggestion
	while (node && !isRoot(node)) {
```

lib/utils/isDescriptorDeclaration.mjs
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Check whether a declaration is a descriptor one
+ *
+ * @param {import('postcss').Declaration} decl
+ * @returns {boolean}
+ */
+export default function isDescriptorDeclaration(decl) {
+	const { parent } = decl;
+
+	/** @type {import('postcss').Node | undefined} */
```
ybiquitous, line 13: [suggestion] Minor. A bit more accurate:

```suggestion
	/** @type {import('postcss').Node['parent']} */
```

ybiquitous (approved):
I've commented on a few minor suggestions, but LGTM 👍🏼 

Just in case, please add a changelog item.

lib/utils/isDescriptorDeclaration.cjs
jeddy3: It's a good suggestion. I've opened an issue to continue the discussion https://github.com/stylelint/stylelint/issues/8221

ybiquitous (approved):

PR #8209, opened 2024-12-17
Fix `property-(dis)allowed-list` false negatives for custom properties

### LINKED ISSUE #8187: Fix `property-(dis)allowed-list` false negatives for custom properties

When using the `property-disallowed-list`, I was expecting custom properties to be disallowed too.

[Those lines](https://github.com/stylelint/stylelint/blob/b2c99cef764643f3bd9539b34cdec58af882db88/lib/rules/property-disallowed-list/index.mjs#L39-L41) seem to tell otherwise.

Was there a reason to not include custom properties in the mix?

I might misunderstand basic naming convention, so ignore custom properties may be normal.

If it is not the case, I would expect this rule to:

- Take all properties into consideration (standardised and custom).
- Have an option to ignore standardised or custom properties.

Otherwise, another rule can be specific to custom properties.

I probably can code either, so this Issue is merely to know which path to follow.

fbasmaison-lucca:

Closes #8187

* I hope tests are OK and not too much or too little.
* `@property` declarations won’t be taken into account here, but another rule might actually solve this issue already.
* I removed some `^` before `/^background/` in the READMEs, to make it clearer regexp will match anything if not constrained (I was expecting `/color/` to behave like `'color'` or `/^color$/`).
* I used and replaced `pink` by `red`, and changed properties to `color` when possible, based on the [Write tests](https://stylelint.io/developer-guide/rules#write-tests) section.


lib/rules/property-allowed-list/README.md
```diff
@@ -4,39 +4,40 @@ Specify a list of allowed properties.
...
 <!-- prettier-ignore -->
 ```css
-a { display: block; }
+a { color: red; }
 /** ↑
  * This property */
 ```
 
-This rule ignores variables (`$sass`, `@less`, `--custom-property`).
+This rule ignores preprocessor variables (`$sass`, `@less`).
```
Mouvedia, line 12:
```suggestion
This rule ignores preprocessor variables (e.g. `$sass`, `@less`).
```

″ for property-disallowed-list/README.md

lib/rules/property-allowed-list/README.md
```diff
@@ -4,39 +4,40 @@ Specify a list of allowed properties.
...
+It parses custom properties (`--foo: red`).
 
 The [`message` secondary option](../../../docs/user-guide/configure.md#message) can accept the arguments of this rule.
 
 ## Options
 
 `array|string|regex`: `["array", "of", /properties/, "regex"]|"property"|"/regex/"|/regex/`
 
-If a string is surrounded with `"/"` (e.g. `"/^background/"`), it is interpreted as a regular expression. This allows, for example, easy targeting of shorthands: `/^background/` will match `background`, `background-size`, `background-color`, etc.
+If a string is surrounded with `"/"` (e.g. `"/^background/"`), it is interpreted as a regular expression. This allows, for example, easy targeting of shorthands: `/background/` will match `background`, `background-size`, `background-color`, etc.
```
Mouvedia, line 21: I think that  `/^background/` makes sense because `/background/` would also target prefixed properties.
e.g. `-webkit-background-clip`

″ for property-disallowed-list/README.md

lib/rules/property-allowed-list/__tests__/index.mjs
```diff
@@ -100,6 +116,60 @@ testRule({
...
 			endColumn: 10,
 		},
+		{
+			code: 'a { color: red; }',
+			message: messages.rejected('color'),
+			line: 1,
+			column: 5,
+			endLine: 1,
+			endColumn: 10,
+		},
```
Mouvedia, line 126: duplicate of 95–102

lib/rules/property-allowed-list/README.md
```diff
@@ -4,39 +4,40 @@ Specify a list of allowed properties.
...
 ```css
-a { display: block; }
+a { color: red; }
 /** ↑
  * This property */
 ```
 
-This rule ignores variables (`$sass`, `@less`, `--custom-property`).
+This rule ignores preprocessor variables (e.g. `$sass`, `@less`).
+It parses custom properties (e.g. `--foo: red`).
```
jeddy3, line 13:
```suggestion
This rule ignores preprocessor variables (e.g. `$sass`, `@less`).
```

Let's keep it consistent with the other `*-property` rules that don't explicitly mention custom properties.

lib/rules/property-allowed-list/__tests__/index.mjs
fbasmaison-lucca: Thank you. This should be fixed.

lib/rules/property-allowed-list/README.md
jeddy3: Let's revert this change for now. You're welcome to open an issue to discuss it, though. If the consensus is to make the change, we'd want to roll it out consistently across all our rule READMEs.

lib/rules/property-disallowed-list/README.md
```diff
@@ -9,18 +9,21 @@ a { text-rendering: optimizeLegibility; }
  * This property */
 ```
 
+This rule ignores preprocessor variables (e.g. `$sass`, `@less`).
+It parses custom properties (e.g. `--foo: red`).
```
jeddy3, line 13:
```suggestion
This rule ignores preprocessor variables (e.g. `$sass`, `@less`).
```

lib/rules/property-allowed-list/README.md
fbasmaison-lucca: I updated the few occurrences. Thank you.

jeddy3 (commented):
@fbasmaison-lucca Thank you for working on this PR. It looks good so far.

> I removed some ^ before /^background/ in the READMEs, to make it clearer regexp will match anything if not constrained (I was expecting /color/ to behave like 'color' or /^color$/). ... replaced pink by red, changed properties to color when possible, based on the [Write tests](https://stylelint.io/developer-guide/rules#write-tests) section

Thank you for following the conventions and for wanting to make the docs more consistent. It's an attention to detail that's much appreciated.

However, we typically keep PRs focused on the changes necessary to address the issue and put refactors and documentation changes into follow-up issues and PRs to make them easier to review. 

Let's keep the value and properties changes in this case (to save you reverting change), but let's take out the regex changes as it's a bigger discussion.

I've added a (minor to be inline with [our semantic versioning](https://stylelint.io/about/semantic-versioning#minor-release)) changeset for you

lib/rules/property-allowed-list/README.md
fbasmaison-lucca: I reverted these changes.

jeddy3 (approved):
@fbasmaison-lucca Thank you for making the changes.

LGTM.

(CI is stalled at the moment, but hopefully it'll whir into action soon)

fbasmaison-lucca:
> Let's keep the value and properties changes in this case (to save you reverting the changes), but let's take out the regex change as it's something we'd want to roll out consistently across rule READMEs.

That’s understood. Thank you for the pass. :)

> I've added a (minor to be inline with [our semantic versioning](https://stylelint.io/about/semantic-versioning#minor-release)) changeset for you

Thanks for this too. I wasn’t quite sure what dolphins had to do with it!

jeddy3:
> Thanks for this too. I wasn’t quite sure what dolphins had to do with it!

It seems you got a particularly good randomly-generated changeset file for your first contribution! 😁

jeddy3 (approved):

ybiquitous (approved):
Thank you, LGTM 👍🏼

Issue #8152, opened 2024-11-29
Add `at-rule-prelude-no-invalid`


jeddy3:
### What is the problem you're trying to solve?

Catch unknown preludes for at-rules.

For example:

```css
@property foo {}
```

As `@property` [expects](https://csstree.github.io/docs/syntax/#Atrule:property) a `<custom-property-name>` for its prelude:

```css
@property --foo {}
```

### What solution would you like to see?

Suggested blueprint:

- Name: `at-rule-prelude-no-invalid`
- Description: Disallow invalid preludes for at-rules.
- Message: `Unexpected invalid prelude "${prelude}" for at-rule "${at-rule}"`
- Primary option: `true`
- Secondary options: `ignoreAtRules: []`
- Category: Avoid errors > Invalid
- Included in sharable configs?: Yes, recommended when pieces are in place
- Expanded description: This rule considers preludes for at-rules defined within the CSS specifications to be valid.

We typically use `no-unknown` for things not in the specs. However, in hindsight, we probably should have used `no-invalid` and reserved `no-unknown` for user-defined things like animations and custom properties. I suggest we continue using `no-unknown` to be consistent with the existing rules.

The `ignoreAtRules: []` secondary option to avoid overlaps with existing rules like [`media-query-no-invalid`](https://stylelint.io/user-guide/rules/media-query-no-invalid/):

```json
{
  "rules": {
    "media-query-no-invalid": true,
    "at-rule-prelude-no-invalid": [true, { "ignoreAtRules": ["media"] }]
  }
}
```

Mouvedia:
[As I explained](https://github.com/stylelint/stylelint/issues/8133#issuecomment-2486224673) in the parent issue, it should be named `at-rule-prelude-no-invalid`.

In your example—`@property foo {}`—`foo` is invalid.

i.e.
`@property` defines the custom property hence even if it was `--foo`—e.g. `@property --foo {}`—searching for a prior registration to know if it's known or not wouldn't make sense

> I suggest we continue using no-unknown to be consistent with the existing rules.

I disagree, we have 5 existing no-invalid rules as a precedent.

jeddy3:
I think `@property foo {}` was a confusing example, sorry.

[...]

PR #8211, opened 2024-12-19
Add `at-rule-descriptor-value-no-unknown`

### LINKED ISSUE #8149: Add `at-rule-descriptor-value-no-unknown`

A dedicated rule to flag unknown values for descriptors within at-rules.

For example, `bar` is not a known value for the `syntax` descriptor for `@property`:

```css
@property --foo {
  syntax: bar;
}
```

This functionality is currently rolled into [`declaration-property-value-no-unknown`](https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown/), which should only target properties rather than descriptors too.

Suggested blueprint:

- Name: `at-rule-descriptor-value-no-unknown`
- Description: Disallow unknown descriptors for at-rules.
- Message: `Unexpected unknown value "${value}" for descriptor "${descriptor}"`
- Primary option: `true`
- Secondary options: n/a
- Category: Avoid errors > Unknown
- Included in sharable configs?: Yes, recommended.
- Expanded description:

This rule considers values for properties defined within the CSS specifications to be known.

ryo-manba:

Closes #8149 

No, it's self-explanatory.


lib/rules/at-rule-descriptor-value-no-invalid/index.cjs
```diff
@@ -0,0 +1,68 @@
...
+const nodeFieldIndices = require('../../utils/nodeFieldIndices.cjs');
+const atKeywords = require('../../reference/atKeywords.cjs');
+const report = require('../../utils/report.cjs');
+const ruleMessages = require('../../utils/ruleMessages.cjs');
+const validateOptions = require('../../utils/validateOptions.cjs');
+
+const UNSUPPORTED_NESTING_REGEX = new RegExp(
+	`^((?!${[...atKeywords.nestingSupportedAtKeywords.values()].join('|')}).)*$`,
+	'i',
+);
```
ryo-manba, line 15: I'd like to unify it later.
See: https://github.com/stylelint/stylelint/pull/8197/files#diff-709b8ceefdd8d9a3b4b446f29f245ce253e36a5cc9843d695c7bfcbadb78b8d6R8-R11

jeddy3 (commented):
This is looking great. Thank you for working on it!

I've requested a few minor changes.

Having seen the reference to [`declaration-property-value-no-unknown`](https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown) and "unknown" examples in the README rule, and the upcoming [`at-rule-descriptor-no-unknown`](https://github.com/stylelint/stylelint/issues/8148) rule,  I'm more convinced that this rule should be called `at-rule-descriptor-value-no-unknown` as the consistency will benefit users.

The upcoming [`at-rule-prelude-no-invalid`](https://github.com/stylelint/stylelint/issues/8152) rule can stay as `no-invalid` to match the overlapping [`media-query-no-invalid`](https://stylelint.io/user-guide/rules/media-query-no-invalid).

What do you think?

lib/rules/at-rule-descriptor-value-no-invalid/README.md
```diff
@@ -0,0 +1,60 @@
...
+@counter-style foo {
+  system: unknown;
+}
+```
+
+<!-- prettier-ignore -->
+```css
+@property --foo {
+  syntax: "<unknown-type>";
+}
```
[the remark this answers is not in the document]
ryo-manba: Oh sorry, that wasn't an appropriate example.

lib/rules/at-rule-descriptor-value-no-invalid/__tests__/index.mjs
```diff
@@ -0,0 +1,89 @@
...
+					column: 42,
+					endLine: 1,
+					endColumn: 45,
+				},
+			],
+			description: 'multiple invalid value in @property',
+		},
+		{
+			code: '@property --foo { syntax: cyclic; }',
+			message: messages.rejected('syntax', 'cyclic'),
```
Mouvedia, line 72:
```suggestion
			code: '@property --foo { system: cyclic; }',
			message: messages.rejected('system', 'cyclic'),
```

ref https://developer.mozilla.org/fr/docs/Web/CSS/@counter-style/system#cyclic

Mouvedia (dismissed):

lib/rules/at-rule-descriptor-value-no-invalid/__tests__/index.mjs
ryo-manba: As the description says, "value known for another at-rule but not @\property," I am purposely using a different value.

Mouvedia: I know and that value is `system` but for the test to be working—i.e. the pair property/value to be valid—the property must exist on another at rule.

ryo-manba: `system: cyclic` is used in the [accept test](https://github.com/stylelint/stylelint/pull/8211/files#diff-d9fc880b1da150e8a06381ed5af062cf0934a61e0004fb7fe98a26cd88bd1660R14). 
Is this not good enough?

Mouvedia: If you don't want to change the `code` and `message` properties you will have to change the `description`.
I can't give you one because I don't know what you are trying to test if it's not a regression test that verify that other pairs aren't passing for non-matching at rules—which is the all point of the rule.

lib/rules/at-rule-descriptor-value-no-invalid/index.mjs
```diff
@@ -0,0 +1,64 @@
+import { lexer } from 'css-tree';
+
+import { declarationValueIndex } from '../../utils/nodeFieldIndices.mjs';
+import { nestingSupportedAtKeywords } from '../../reference/atKeywords.mjs';
+import report from '../../utils/report.mjs';
+import ruleMessages from '../../utils/ruleMessages.mjs';
+import validateOptions from '../../utils/validateOptions.mjs';
+
+const UNSUPPORTED_NESTING_REGEX = new RegExp(
+	`^((?!${[...nestingSupportedAtKeywords.values()].join('|')}).)*$`,
```
Mouvedia, line 10: question

Are you sure you want `.` and `*`?
Wouldn't something like `^(?:${[...nestingSupportedAtKeywords.values()].join('|')})+$` be more accurate?

lib/rules/at-rule-descriptor-value-no-invalid/__tests__/index.mjs
jeddy3: I think the test is fine. It's similar to the `@property --foo { syntax: bar; }` test but uses a value known to another descriptor rather than something made up like `bar`.

However, this discussion brings up a good point. Let's add an _accept_ `test` for:

```css
@property --foo { system: bar; }
```

With the description: "a descriptor (with an unknown value) that is unknown to the at-rule"

As checking for unknown descriptors is the responsibility of the `at-rule-descriptor-no-unknown` rule: https://github.com/stylelint/stylelint/pull/8197.

lib/rules/at-rule-descriptor-value-no-invalid/__tests__/index.mjs
```diff
@@ -0,0 +1,89 @@
...
+			line: 1,
+			column: 27,
+			endLine: 1,
+			endColumn: 33,
+			description: 'a value known for another at-rule but not @property',
```
jeddy3, line 77:
```suggestion
			description: 'a value known to a descriptor of another at-rule but not @property',
```

We can tweak this slightly be be specific.

lib/rules/at-rule-descriptor-value-no-invalid/README.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+This rule considers descriptors and values defined in the CSS Specifications, up to and including Editor's Drafts, to be known.
+
+You can filter the [CSSTree Syntax Reference](https://csstree.github.io/docs/syntax/) to find out what values are valid for a descriptor of an at-rule.
+
+The [`message` secondary option](../../../docs/user-guide/configure.md#message) can accept the arguments of this rule.
+
+This rule checks descriptor values within at-rules. You can use [`declaration-property-value-no-unknown`](../declaration-property-value-no-unknown/README.md) rule to disallow unknown values for properties within declarations.
```
jeddy3, line 20:
```suggestion
This rule checks descriptor values within at-rules. You can use [`declaration-property-value-no-unknown`](../declaration-property-value-no-unknown/README.md) to disallow unknown values for properties within declarations, and [`at-rule-descriptor-no-unknown`](../at-rule-descriptor-no-unknown/README.md) to disallow unknown descriptors for at-rules.
```

Let's add a reference to the `at-rule-descriptor-no-unknown` rule as it touches the same area, and people will likely use them in conjunction.

lib/rules/at-rule-descriptor-value-no-invalid/index.mjs
ryo-manba: Wouldn't your example end up checking only `nestingSupportedAtKeywords`? 
That pattern would cause the tests to fail.

lib/rules/at-rule-descriptor-value-no-invalid/index.cjs
```diff
@@ -0,0 +1,68 @@
...
+const rule = (primary) => {
+	return (root, result) => {
+		const validOptions = validateOptions(result, ruleName, { actual: primary });
+
+		if (!validOptions) {
+			return;
+		}
+
+		root.walkAtRules(UNSUPPORTED_NESTING_REGEX, (atRule) => {
+			atRule.walkDecls((decl) => {
```
jeddy3, line 38:
```suggestion
		root.walkAtRules(UNSUPPORTED_NESTING_REGEX, (atRule) => {
		    if (!isStandardSyntaxAtRule(atRule)) return;
			atRule.walkDecls((decl) => {
			  if (!isStandardSyntaxDeclaration(decl)) return;
```

Let's return early for non-standard at-rules and decls.

See https://github.com/stylelint/stylelint/pull/8231

jeddy3 (commented):
I've requested one change to be consistent with other rules.

>> I'm more convinced that this rule should be called at-rule-descriptor-value-no-unknown as the consistency will benefit users.

> Yes, I agree. Consistency with existing rules, this seems like a better choice.

I think we're okay to change the rule name if you and I are in agreement. Please replace "invalid" with "unknown" in the `ruleMessage`, too

lib/rules/at-rule-descriptor-value-no-invalid/index.cjs
ryo-manba: Thanks for your review! I've applied it.
[Return early for non-standard at-rules and decls](https://github.com/stylelint/stylelint/pull/8211/commits/ded4ebbc264f11a6ec36efd9b6044b282c823dd6)

ryo-manba:
The following PR needs to be merged first to prevent lint errors related to the `lexer` type definition and references to `at-rule-descriptor-no-unknown` in the README:

- https://github.com/stylelint/stylelint/pull/8197/

jeddy3 (approved):
Looks great, thank you!

types/stylelint/index.d.ts
```diff
@@ -282,6 +282,7 @@ declare namespace stylelint {
 		>;
 		'annotation-no-unknown': CoreRule<true, { ignoreAnnotations: OneOrMany<StringOrRegex> }>;
 		'at-rule-allowed-list': CoreRule<OneOrMany<string>>;
+		'at-rule-descriptor-value-no-unknown': CoreRule<true>;
```
Mouvedia, line 285: **If** #8239 gets merged before this one

```suggestion
		'at-rule-descriptor-value-no-unknown': CoreRule<
			true,
			{},
			{ rejected: (descriptor: string, value: string) => string }
		>;
```

The third argument is optional so it's not mandatory though.

ybiquitous (approved):
One point, can you add a changelog file? Otherwise LGTM 👍🏼

ryo-manba:
@Mouvedia
Thanks for adding the changeset!
I will merge it.

PR #8214, opened 2024-12-21
Add `{ "ignore": ["at-rule-preludes", "declaration-values"] }` to `string-no-newline`

ryo-manba:
> Which issue, if any, is this issue related to?

Closes #8151 

> Is there anything in the PR that needs further explanation?

No, it's self-explanatory.

Issue #7492, opened 2024-01-23
Add `at-charset-rule-no-invalid`


firefoxic:
### What is the problem you're trying to solve?

I went to fix [this bug](https://github.com/stylelint-stylistic/stylelint-stylistic/issues/13). And found out that `@charset` is not even an at-rule, but a special construct with much stricter syntax requirements than at-rules. And it doesn't seem to be about choosing a formatting style at all. After all, if the quotes are single, the parser will just pass it by without recognizing it as an encoding declaration. It is good if the encoding is `utf-8`, which is the default encoding. But if another encoding is used, it will probably cause an error.

So it's so much a matter of syntax (rather than formatting style) that the rule controlling `@charset`:
1. should be in Stylelint itself, not in the plugin;
2. should have one single option — the value `true`;
3. should be added not even to the standard, but to the recommended config.

### What solution would you like to see?

Add a separate rule to control the `@charset` syntax, which would take into account the character set of this declaration, its location in the style file, and maybe something else I may not know about yet.

ybiquitous:
@firefoxic Thanks for the proposal. Very interesting to me.

Before looking into this issue and `@charset`, I didn't know the at-rule was deprecated. The following MDN documents say:

- https://developer.mozilla.org/en-US/docs/Web/CSS/@charset
  > The CSS syntax module deprecates this rule, defining it as an unrecognized legacy rule to be dropped when a stylesheet is grammar-checked
- https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_syntax
  > Note: The module explicitly states that `@charset` is not an actual at-rule, but rather an unrecognized legacy rule that should be omitted when a stylesheet is grammar-checked.

So, I wonder if a new rule should disallow the use of `@charset`, e.g.

```css
@charset "UTF-8"; /* Unexpected deprecated "@charset" (no-charset) */
```

This rule might be expected to be put in the recommended config, as you suggest.

firefoxic:
> So, I wonder if a new rule should disallow the use of `@charset`

I'm not sure about this. The same article https://developer.mozilla.org/en-US/docs/Web/CSS/@charset also mentions this:
> but it is fully supported in all browsers.

Maybe someone can rely on such an encoding declaration.

On the other hand, it's [just one way to determine encoding](https://drafts.csswg.org/css2/#charset). And if it really is deprecated, linter would be better off not allowing it at all. Then yes, the rule would be better called `no-charset`.

[...]

PR #8215, opened 2024-12-21
Fix `atrule-*` false positives and negatives for `@charset` rule

jeddy3:

Ref https://github.com/stylelint/stylelint/issues/7492#issuecomment-1907995564
Ref https://github.com/stylelint/stylelint/pull/8214#pullrequestreview-2518665122

Fixes `@charset` being treated as an at-rule in our examples and tests.

It felt like an appropriate time to address this as we're adding new rules to shore up our capabilities around at-rules.

### DIFF
--- a/lib/utils/isStandardSyntaxAtRule.cjs
+++ b/lib/utils/isStandardSyntaxAtRule.cjs
@@ -9,6 +9,11 @@
  * @returns {boolean} If `true`, the declaration is standard
  */
 function isStandardSyntaxAtRule(atRule) {
+	// Ignore `@charset` css rule (is parsed as at-rule)
+	if (atRule.name === 'charset') {
+		return false;
+	}
+
 	// Ignore scss `@content` inside mixins
 	if (!atRule.nodes && atRule.params === '') {
 		return false;
--- a/lib/utils/isStandardSyntaxAtRule.cjs
+++ b/lib/utils/isStandardSyntaxAtRule.cjs
@@ -10,7 +10,7 @@
  */
 function isStandardSyntaxAtRule(atRule) {
 	// Ignore `@charset` css rule (is parsed as at-rule)
-	if (atRule.name === 'charset') {
+	if (atRule.name.toLowerCase() === 'charset') {
 		return false;
 	}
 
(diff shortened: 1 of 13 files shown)


romainmenke (approved):
Thank you. LGTM 👍

-----

I like that `isStandardSyntaxAtRule` is used for this.
It is both helpful in a very practical way and technically correct.

jeddy3:
@romainmenke Thanks for the review. 

I forgot to remove `@charset` from our reference library, which was masking a case bug. The latest commit fixes this.

romainmenke (approved):
LGTM!

Good catch on those last fixes :)

lib/utils/isStandardSyntaxAtRule.cjs
```diff
@@ -9,6 +9,11 @@
  * @returns {boolean} If `true`, the declaration is standard
  */
 function isStandardSyntaxAtRule(atRule) {
+	// Ignore `@charset` css rule (is parsed as at-rule)
+	if (atRule.name.toLowerCase() === 'charset') {
+		return false;
+	}
```
Mouvedia, line 15: fix
medium

> [@charset must be written literally, i.e., the 10 characters `@charset "` (lowercase, no backslash escapes), followed by the encoding name, followed by `";`.](https://drafts.csswg.org/css2/#charset)

i.e. you shouldn't lowercase `atRule.name`, it wouldn't be valid anyway

Mouvedia (changes requested):

lib/utils/isStandardSyntaxAtRule.cjs
romainmenke: Either `@charser` was written correctly and then it isn't a real `at-rule`, or it was written incorrectly, also making it not a standard `at-rule`.

In both cases it is never a standard syntax `at-rule` and could be ignored by all these rules :)

I don't think we will ever want to process `@CHARSET "utf-8";` but not one of:
- `@charset "utf-8"`
- `@charset  "utf-8"` (two spaces)
- `@charset 'utf-8'` (single quote)

Either we want to ignore any `@charset` and all its invalid variations or we want to target `@charset` specifically to make sure it was written correctly. (i.e. an `at-charset-rule-no-invalid` rule)

Issue #8126, opened 2024-11-14
Change `report` to be consistent and predictable in how it handles the provided position arguments


romainmenke:
This same issue is what I was also trying to address in : https://github.com/stylelint/stylelint/pull/8098

Currently I see these behaviors:

1. if `line` is provided:
   - use `line` as `startLine`
2. if `line` is not provided:
   - calculate `startLine` from `index` and `endIndex` (ignoring `start.line` if that was provided)
   - if `index` was not provided, PostCSS will fallback to the start of the node
3. use `startLine` to determine disable ranges
3. `startLine` is no longer used after this point
4. if `start` was provided
   - use `start` in `warningProperties`
5. if `index` was provided and is a `number`
   - use `index` in `warningProperties`
6. if `end` was provided
   - use `end` in `warningProperties`
7. if `endIndex` was provided and is a `number`
   - use `endIndex` in `warningProperties`
8. if none were provided, let PostCSS fallback to the entire range of the node

----------

This implies several inconsistencies:
1. passing only `start` and `end` will not affect disable ranges
2. passing only `line` will not be reflected in the report, even when it is used in the disable ranges

Possibly there are more bugs but I think these are the most important to resolve.

------

Issue 1 is also why it is still needed to pass a `line` argument in: https://github.com/stylelint/stylelint/pull/8074/files#diff-bd94fe99a5bd46c7e6b11e8afd5f71ca4aa1897e2d52801cc1e7e3be5de7d925R117

Changing that rule report to pass `index` and `endIndex` instead makes the `line` argument redundant.

This is very counter intuitive from a rule author perspective because currently `line` is already passed implicitly as part of `start` and yet it seems to be ignored.

-----

My preferred solution would be to have proper input validation and normalization for the position arguments of `report()`
The resulting values can then be used in a more consistent and controlled manner.

Mouvedia:
The only thing that bothers me with `report` is that `node` is/was treated as if it is optional.

https://github.com/stylelint/stylelint/blob/3bf048d2e0c7d67841e0fcc7298788ae89c5a7e9/lib/utils/report.mjs#L56

https://github.com/stylelint/stylelint/blob/3bf048d2e0c7d67841e0fcc7298788ae89c5a7e9/lib/utils/report.mjs#L52

https://github.com/stylelint/stylelint/blob/3bf048d2e0c7d67841e0fcc7298788ae89c5a7e9/lib/utils/report.mjs#L89

[...]

PR #8217, opened 2024-12-22
Change `report` to be consistent and predictable in how it handles the provided position arguments

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/8126

At this time this is only a draft.
Intended to have a clear direction based on the discussion in https://github.com/stylelint/stylelint/issues/8126#issuecomment-2482487257


lib/rules/selector-not-notation/index.mjs
```diff
@@ -114,8 +116,8 @@ const rule = (primary) => {
 					node: ruleNode,
 					result,
 					ruleName,
-					line: range.start.line,
-					...range,
+					index,
+					endIndex,
```
romainmenke, line 120: Using `index` and `endIndex` is the only way to have the correct outcome within Stylelint `16.x`.

Ideally we would use `start/end` because that is slightly more performant.

But this isn't compatible with the planned deprecation of `line` as an optional argument.

lib/rules/function-no-unknown/index.mjs
```diff
@@ -67,14 +67,18 @@ const rule = (primary, secondaryOptions) => {
 
 				if (FUNCTIONS.includes(name.toLowerCase())) return;
 
+				const declIndex = declarationValueIndex(decl);
+				const index = declIndex + node.name[2];
+				const endIndex = declIndex + node.name[3];
```
Mouvedia, line 72: note

`2` is "Start position of representation"
`3` is "End position of representation"

lib/rules/selector-not-notation/index.mjs
Mouvedia: IIRC `getRange` was necessary because, _back then_, the clones—l169—were lacking some data. If the tests pass all should be good.

lib/utils/report.mjs
```diff
@@ -90,29 +81,60 @@ export default function report(problem) {
...
+			start: problem.start,
+			end: problem.end,
+		};
 	}
 
-	if (end) {
-		warningProperties.end = end;
-	} else if (isNumber(endIndex)) {
-		warningProperties.endIndex = endIndex;
+	if ('start' in problem || 'end' in problem) {
```
Mouvedia, line 115: maybe

```
	} else if ('start' in problem || 'end' in problem) {
```

NB: suggestion doesn't work

lib/utils/report.mjs
```diff
@@ -90,29 +81,60 @@ export default function report(problem) {
...
-	if (word) {
-		warningProperties.word = word;
+	// `index/endIndex` based range
+	if (hasIndices(problem)) {
+		return problem.node.rangeBy({ index: problem.index, endIndex: problem.endIndex });
 	}
 
-	if (customUrl) {
-		warningProperties.url = customUrl;
+	if ('index' in problem || 'endIndex' in problem) {
```
Mouvedia, line 126: maybe `else if`?

types/stylelint/index.d.ts
```diff
@@ -814,6 +814,38 @@ declare namespace stylelint {
...
+				 * `index`.
+				 */
+				start: Position;
+				/**
+				 * The exclusive end position of the problem, relative to the
+				 * node's source text. If provided, this will be used instead of
+				 * `endIndex`.
+				 */
+				end: Position;
+		  }
```
Mouvedia, line 831: I suppose you didn't reuse `Range` at l251 so that you could add those comments?

types/stylelint/index.d.ts
```diff
@@ -814,6 +814,38 @@ declare namespace stylelint {
...
+				/**
+				 * The exclusive end index of the problem, relative to the node's
+				 * source text.
+				 */
+				endIndex: number;
+		  }
+		| {
+				word: string;
+		  }
+		| {};
```
Mouvedia, line 847: Why did you choose not to add
```
| {
	index: 0;
	endIndex?: number;
}
```
?

types/stylelint/index.d.ts
romainmenke: I kept the original code as it already exists, only moved it around.
See: https://github.com/stylelint/stylelint/blob/727bb1661d148cb20dc9161883d945b403e70998/types/stylelint/index.d.ts#L836-L847

types/stylelint/index.d.ts
romainmenke: I've answered this here to keep the design discussion in one place: https://github.com/stylelint/stylelint/issues/8126#issuecomment-2559300114

types/stylelint/index.d.ts
```diff
@@ -844,7 +844,7 @@ declare namespace stylelint {
 		| {
 				word: string;
 		  }
-		| {};
+		| object;
```
Mouvedia, line 847: Like I said in https://github.com/stylelint/stylelint/issues/8126#issuecomment-2559331560 you probably want `{ [key: string]: never }` or `Record<string, never>`.

ref https://biomejs.dev/linter/rules/no-banned-types/

romainmenke: Neither actually works :)
`object` does work.

Screenshot 2024-12-24 at 14 27 26

lib/utils/report.mjs
romainmenke: This function uses early returns.
So `else if` wouldn't make sense imho.

lib/utils/report.mjs
romainmenke: Also early returns here.

lib/utils/report.mjs
```diff
@@ -134,41 +107,46 @@ export default function report(problem) {
...
-	if (problem.result.stylelint.quietDeprecationWarnings) return;
-
+function getProblemRange(problem) {
 	if (!problem.node) {
-		emitDeprecationWarning(
-			`Omitting the \`node\` argument in the \`utils.report()\` function is deprecated ("${problem.ruleName}").`,
-			'REPORT_AMBIGUOUS_POSITION',
-			`Please pass a \`node\` argument in the \`utils.report()\` function of "${problem.ruleName}".`,
+		throw new Error(
+			`The "${problem.ruleName}" rule failed to provide correct position arguments to the \`utils.report()\` function. Provide a PostCSS \`node\` to the \`utils.report()\` function.`,
```
jeddy3:
```suggestion
			`The "${problem.ruleName}" rule provided incorrect position arguments to \`utils.report()\`. To fix, pass a PostCSS \`node\` to the function.`,
```

lib/utils/report.mjs
```diff
@@ -134,41 +107,46 @@ export default function report(problem) {
...
 	}
 
-	if (!hasIndices(problem) && ('index' in problem || 'endIndex' in problem)) {
-		emitDeprecationWarning(
-			`Partial position information in the \`utils.report()\` function is deprecated ("${problem.ruleName}").`,
-			'REPORT_AMBIGUOUS_POSITION',
-			`Please pass both \`index\` and \`endIndex\` as arguments in the \`utils.report()\` function of "${problem.ruleName}".`,
+	if ('start' in problem || 'end' in problem) {
+		throw new Error(
+			`The "${problem.ruleName}" rule failed to provide correct position arguments to the \`utils.report()\` function. Both \`start\` and \`end\` are required and must be valid positions.`,
```
jeddy3:
```suggestion
			`The "${problem.ruleName}" rule provided incomplete position arguments to \`utils.report()\`. To fix, provide both \`start\` and \`end\` as valid positions to the function.,
```

lib/utils/report.mjs
```diff
@@ -134,41 +107,46 @@ export default function report(problem) {
...
-			'REPORT_AMBIGUOUS_POSITION',
-			`Please pass both \`index\` and \`endIndex\` as arguments in the \`utils.report()\` function of "${problem.ruleName}" instead.`,
+	// `index/endIndex` based range
+	if (hasIndices(problem)) {
+		return problem.node.rangeBy({ index: problem.index, endIndex: problem.endIndex });
+	}
+
+	if ('index' in problem || 'endIndex' in problem) {
+		throw new Error(
+			`The "${problem.ruleName}" rule failed to provide correct position arguments to the \`utils.report()\` function. Both \`index\` and \`endIndex\` are required and must be numbers.`,
```
jeddy3:
```suggestion
			`The "${problem.ruleName}" rule provided incomplete position arguments to \`utils.report()\`. To fix, provide both \`index\` and \`endIndex\` as numbers to the function.`,
```

jeddy3 (commented):
Thank you for getting the `v17` ball rolling with the PR! It's looking great.

I've suggested some tweaked error messages to make:

- what went wrong more concise
- what needs to be done to fix it more explicit

They also soften and humanise the language a little. I suspect a thrown error will be jarring for people.

jeddy3:
If we merge this after https://github.com/stylelint/stylelint/pull/8859, let's also update the migration guide in this PR with a new heading under "Breaking changes". Something along the lines of:

---

## Changed `report` to be consistent and predictable in how it handles the provided position arguments

We've changed `report` so that it'll throw an error for the ambiguous position arguments that we deprecated in [16.13.0](https://stylelint.io/CHANGELOG#16130---2025-01-12).

You should update your plugin(s) to:

- always provide a PostCSS `node`
- provide both `start` and `end` as valid positions, if providing a [range](https://postcss.org/api/#range) within the given node
- provide both `index` and `endIndex` as numbers, if providing an index range within the given node

For example:

```diff js
root.walkRules((ruleNode) => {
  const { selector } = ruleNode;

  report({
    result,
    ruleName,
    message: messages.rejected(),
    node: ruleNode,
    index,
+   endIndex: index + 1
  });
});
```

You can learn more about the different options for specifying the location of a problem in our [developer guide](https://stylelint.io/developer-guide/rules#location-arguments-for-report).

---

(Can probably be fleshed out and made more accurate.)

romainmenke:
Thank you for these suggestions @jeddy3 🙇 
First time I'm working on a major, so I really appreciate the guidance around adding the migration docs.

docs/migration-guide/to-17.md
```diff
@@ -0,0 +1,36 @@
...
+
+This release contains significant or breaking changes.
+
+## Significant changes
+
+We've:
+
+- changed report to be consistent and predictable in how it handles the provided position arguments
+
+### Changed `report` to be consistent and predictable in how it handles the provided position arguments
```
jeddy3, line 108: Let's put these at the end of the "breaking changes" list.

jeddy3 (changes requested):
@romainmenke Thanks for making the changes. 

I think we can merge when the migration entry is moved under breaking changes and the conflicts are resolved.

romainmenke:
I've updated this branch and moved the migration docs to the end of the list :)

jeddy3 (approved):
LGTM, thank you!

PR #8222, opened 2024-12-23
Document updating types when adding an option

jeddy3:

Ref https://github.com/stylelint/stylelint/pull/8214#issuecomment-2559977052

No, it's self-explanatory.

### DIFF
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -302,6 +302,7 @@ You should:
 3. Change the rule's validation to allow for the new option.
 4. Add (as little as possible) logic to the rule to make the tests pass.
 5. Add documentation about the new option.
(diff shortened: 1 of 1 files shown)


docs/developer-guide/rules.md
```diff
@@ -302,6 +302,7 @@ You should:
 3. Change the rule's validation to allow for the new option.
 4. Add (as little as possible) logic to the rule to make the tests pass.
 5. Add documentation about the new option.
+6. Add the option to the [type definition of the rule](../../types/stylelint/index.d.ts)
```
Mouvedia, line 305:
```suggestion
6. Add the option(s) to the [type definition of the rule](../../types/stylelint/index.d.ts).
```

jeddy3: The rest of the list is singular. Let's keep this line the same to be consistent.

Issue #8208, opened 2024-12-17
Add rules to flag deprecated things


jeddy3:
### What is the problem you're trying to solve?

It'd be great if we had a consistent way of letting users know if a CSS feature has been deprecated or replaced, whether it was in an editor's draft that got changed or behind a feature flag.

At-rules, pseudo-classes, properties and media features such as:

```css
@viewport {
  min-width: 640px;
}

frame:matches(.c, .d) {
  @nest b & {
    color: red;
  }

  @apply foo;
}

@media (-ms-high-contrast: active) {
  color: red;
}

img {
  -ms-high-contrast-adjust: none;
}
```

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIACA3AljA7gBwgJwBcACYAHQDsiiBbNcgWkzSgIAt4iA2AFgAZsAHgG4KAXwoUAZngCG1GPGoyCYVjADOACgB0YADRFtUAJQkJBJOQ3EARkQBkZ8qQIvIAG3wc8MKCOcE4s4BSDLY2O4AnkSSEBD+QRRI8lBoMkSa9NTq9KxoAOas9JDkBLLqBBwyYARoKDCmZAEeXkQ+fmIS5GjU+U5UWTl5hcUQpeUE9DJQAFYArhUc5GMwCSB6IJJo7jAAcnJwiDACchEwuurq6+BjW-kIIE1UpCDHBDDkUOovHADaFFRniAKpEdu46JMSnd6BUZJ8ZHgoC8AUQALpia5QgoAMXwSgIDxm6jG11g2CuiCeRBeILBEJ+1JA7mU1heehRNIIoJg4NKo3I0Nh8MRDJezPeFWR5FEIFEQA)

### What solution would you like to see?

A set of rules to give users a consistent experience in this area that complements:

- https://github.com/stylelint/stylelint/issues/8178

For example:

- `at-rule-no-deprecated`, e.g. `@nest`, `@apply`
- `selector-no-deprecated`, e.g. `:matches` and `frame`
- `feature-no-deprecated`, e.g. `-ms-high-contrast`

Some may start small, catching just a handful of things each. 

We can either call them `*-no-deprecated` or `*-no-non-standard`. The latter is in keeping with [our existing group](https://stylelint.io/user-guide/rules#non-standard), but 'deprecated' may have a clearer meaning for people.

[...]

PR #8223, opened 2024-12-24
Add `declaration-property-value-keyword-no-deprecated`

### LINKED ISSUE #8178: Add `declaration-property-value-keyword-no-deprecated`

The keyword `overlay` is a legacy value alias for `auto` ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#auto)).

```css
.class {
  overflow: overlay;
}
```

~~Add a new rule `no-overflow-overlay` which could replace the `overlay` value with `auto`.~~

- Name: `declaration-property-value-no-deprecated`
- Description: "Disallow deprecated values for properties."
- Message: `Unexpected deprecated value "${value}" for property "${property}"`
- Primary option: `true`
- Secondary option(s): `{ ignoreValues: [] }`
- Category: Avoid errors > Deprecated
- Included in sharable configs?: Yes, recommended
- Autofixable: Partially

Mouvedia:

Closes #8178

I went for a simple implementation.

i.e.

1.
not trying to fix the `word-break: break-word;` case

2.
shorthand properties are not supported
e.g. `background: no-repeat border-box InactiveBorder;`

3.
`intrinsic` and `min-intrinsic` don't get a fix because it would require a lot of research to confirm that the standard-compliant keywords are truly equivalent in every aspect
i.e. it's unsafe

4.
This PR also remove some duplicated keywords from `systemColorsKeywords`.


lib/reference/properties.cjs
```diff
@@ -614,6 +614,32 @@ const longhandSubPropertiesOfShorthandProperties = new Map([
 	],
 ]);
 
+const singleValueColorProperties = new Set([
```
Mouvedia, line 617: `scrollbar-color` accept 2 colors

lib/reference/properties.mjs
```diff
@@ -610,6 +610,32 @@ export const longhandSubPropertiesOfShorthandProperties = new Map([
 	],
 ]);
 
+export const singleValueColorProperties = new Set([
```
Mouvedia, line 613: `scrollbar-color` accept 2 colors

docs/user-guide/rules.md
```diff
@@ -11,6 +11,16 @@ We turn on most of the rules in our [standard config](https://www.npmjs.com/pack
 
 You can avoid errors with these `no` rules.
 
+### Deprecated
+
+Disallow deprecated things with these `no-deprecated` rules.
```
Mouvedia, line 16: see #8208

lib/rules/declaration-property-value-no-deprecated/index.mjs
```diff
@@ -0,0 +1,141 @@
...
+	threedface: 'ButtonFace',
+	inactivecaptiontext: 'GrayText',
+};
+const COLOR_PROPERTIES = [...singleValueColorProperties.values()].map(
+	(value) => /** @type {[string, Record<String, string>]} */ ([value, DEPRECATED_COLORS]),
+);
+const NAME_TO_VALUE = new Map([
+	[
+		'appearance',
+		/** @type {string | string[] | Record<string, string>} */ ({
```
Mouvedia, line 51: typescript uses the first value to infer the type of the values

lib/rules/declaration-property-value-no-deprecated/index.mjs
```diff
@@ -0,0 +1,141 @@
...
+	['zoom', { reset: '1' }],
+	['text-decoration-line', 'blink'],
+	['box-sizing', 'padding-box'],
+	['image-orientation', 'flip'],
+	['min-height', ['intrinsic', 'min-intrinsic']],
+	['min-width', ['intrinsic', 'min-intrinsic']],
+	['max-height', ['intrinsic', 'min-intrinsic']],
+	['max-width', ['intrinsic', 'min-intrinsic']],
+	['height', ['intrinsic', 'min-intrinsic']],
+	['width', ['intrinsic', 'min-intrinsic']],
```
Mouvedia, line 86: Supporting arrays as keys for the map would be too impractical.
i.e.
```
	[
		['min-height', 'min-width', 'max-height', 'max-width', 'height', 'width'],
		['intrinsic', 'min-intrinsic']
	],
```

docs/user-guide/rules.md
```diff
@@ -11,6 +11,16 @@ We turn on most of the rules in our [standard config](https://www.npmjs.com/pack
...
+
+<!-- prettier-ignore-start -->
+| | | |
+| :-- | :-: | :-: |
+| [`declaration-property-value-no-deprecated`](../../lib/rules/declaration-property-value-no-deprecated/README.md)<br/>Disallow deprecated values for properties within declarations. | | 🔧 |
```
Mouvedia, line 21: I don't use standard and I don't have write access to `stylelint-config-standard` so I left it empty because I can't guarantee the release will be synched.

lib/rules/declaration-property-value-keyword-no-deprecated/README.md
```diff
@@ -0,0 +1,84 @@
...
+<!-- prettier-ignore -->
+```css
+  a { color: ThreeDDarkShadow; }
+/**     ↑         ↑
+ * property and value pairs like these */
+```
+
+The [`fix` option](../../../docs/user-guide/options.md#fix) can automatically fix some of the problems reported by this rule.
+
+The [`message` secondary option](../../../docs/user-guide/configure.md#message) can accept the arguments of this rule.
```
Mouvedia, line 14:
```suggestion
The [`message` secondary option](../../../docs/user-guide/configure.md#message) can accept `value` and `property` as arguments.
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,157 @@
...
+				possible: {
+					ignoreValues: [isString, isRegExp],
+				},
+			},
+		);
+
+		if (!validOptions) return;
+
+		root.walkDecls((decl) => {
+			const declarationValue = decl.value;
```
jeddy3, line 112:
```suggestion
			const {prop, value} = decl;
```

Let's destructure the node properties to be consistent with how we do it in other newer rules.

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,157 @@
...
+
+			if (Array.isArray(value) && !value.includes(keyword)) return;
+
+			if (isPlainObject(value)) {
+				const fixedValue = value[keyword];
+
+				if (!fixedValue) return;
+
+				fix = () => {
+					decl.value = fixedValue;
```
jeddy3, line 133:
```suggestion
					decl.value = fixedValue;
```

To fix values with comments, we may need to use the `get/setDeclarationValue` utils and the value parser.

jeddy3 (changes requested):
@Mouvedia Thank you for working on this rule, especially digging into the mappings between deprecated and new keywords! It's looking good.

lib/rules/declaration-property-value-keyword-no-deprecated/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+Disallow deprecated keywords for properties within declarations.
+
+<!-- prettier-ignore -->
+```css
+  a { color: ThreeDDarkShadow; }
+/**     ↑         ↑
+ * property and value pairs like these */
+```
+
```
jeddy3, line 11:
```suggestion

This rule flags keywords that were removed or deprecated after being in the CSS specifications, including Editor Drafts, and were either:

- shipped in a release browser version
- shipped in a pre-release browser version
- shipped but behind feature flags
- polyfilled with some adoption before any browser shipped
- given an MDN page
```

Let's add these criteria to the extended description.

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,157 @@
...
+import ruleMessages from '../../utils/ruleMessages.mjs';
+import { singleValueColorProperties } from '../../reference/properties.mjs';
+import validateOptions from '../../utils/validateOptions.mjs';
+
+const ruleName = 'declaration-property-value-keyword-no-deprecated';
+
+const messages = ruleMessages(ruleName, {
+	rejected: (value, property) =>
+		`Unexpected deprecated keyword "${value}" for property "${property}"`,
+});
```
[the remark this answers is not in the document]
Mouvedia: see https://github.com/stylelint/stylelint/pull/8223/files#diff-37ccc21a1d2eba4e9fa348a6709d4a249030fc61b658933690e950b48824a2e4R12
i.e. not all cases can be fixed

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
Mouvedia: It was on purpose.
i.e. it would conflict with https://github.com/stylelint/stylelint/pull/8223/files#diff-b1b4ecce949c46a09fbb6d105055c3c68d1da28fceaa18e1837fd92a8c4b8a1dR116

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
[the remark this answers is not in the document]
jeddy3: Let's use an `expected` message for things that can be fixed and `rejected` one for those that can't, as the `expected` message is more useful.

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
Mouvedia: I did it initially but it was not necessary.
I have compared the resulting nodes and added regression tests just in case.
see https://github.com/stylelint/stylelint/pull/8223/files#diff-652c44ce4d315110a294f6324f0b75cfda7f1ae30e1c82c9233b7d31b62fff05R33 https://github.com/stylelint/stylelint/pull/8223/files#diff-652c44ce4d315110a294f6324f0b75cfda7f1ae30e1c82c9233b7d31b62fff05R42
cf 9599633

Mouvedia: The missing `fix: true` mislead me.
Ill fix it.

lib/rules/declaration-property-value-keyword-no-deprecated/README.md
Mouvedia: Iv removed the last point because I consider it not enough until we resolve #7510.

e.g.
If [`:blank`](https://developer.mozilla.org/en-US/docs/Web/CSS/:blank#browser_compatibility) ever gets removed, only a polyfill would warrant it to be detected by `selector-no-deprecated`—not just an MDN page.
see w3c/csswg-drafts#1967

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
jeddy3: Can we rename some of the other variables to clarify some conditionals?

Destructuring `const {prop, value} = decl;` is a common pattern in our rules, so let's use that and adapt the other variables.

I've left some suggestions below.

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+			parsedValue.walk((node) => {
+				if (node.type !== 'word') return;
+
+				const keyword = node.value.toLowerCase();
+
+				if (typeof value === 'string' && value !== keyword) return;
+
+				if (Array.isArray(value) && !value.includes(keyword)) return;
+
+				const messageArgs = [declarationValue, decl.prop];
```
jeddy3, line 135:
```suggestion
				const messageArgs = [actualKeyword, prop];
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+			const value = PROPERTY_NAME_TO_KEYWORD.get(decl.prop.toLowerCase());
+
+			if (!value) return;
+
+			const parsedValue = valueParser(getDeclarationValue(decl));
+
+			parsedValue.walk((node) => {
+				if (node.type !== 'word') return;
+
+				const keyword = node.value.toLowerCase();
```
jeddy3, line 129:
```suggestion
				const actualKeyword = node.value.toLowerCase();
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+					const fixedValue = value[keyword];
+
+					if (!fixedValue) return;
+
+					messageArgs[1] = fixedValue;
+
+					fix = () => {
+						node.value = fixedValue;
+						setDeclarationValue(decl, parsedValue.toString());
+					};
```
jeddy3, line 148:
```suggestion
					const expectedKeyword = keywords[actualKeyword];

					if (!expectedKeyword) return;

					messageArgs[1] = expectedKeyword;

					fix = () => {
						node.value = expectedKeyword;
						setDeclarationValue(decl, parsedValue.toString());
					};
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+		);
+
+		if (!validOptions) return;
+
+		root.walkDecls((decl) => {
+			const declarationValue = decl.value;
+
+			if (optionsMatches(secondaryOptions, 'ignoreKeywords', declarationValue)) return;
+
+			const value = PROPERTY_NAME_TO_KEYWORD.get(decl.prop.toLowerCase());
```
jeddy3, line 120:
```suggestion
			const keywords = PROPERTY_NAME_TO_KEYWORD.get(prop.toLowerCase());
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+		},
+	],
+	['overflow', { overlay: 'auto' }],
+	['text-justify', { distribute: 'inter-character' }],
+	['text-orientation', { 'sideways-right': 'sideways' }],
+	['user-select', { element: 'contain' }],
+	['zoom', { reset: '1' }],
+	['text-decoration-line', 'blink'],
+	['box-sizing', 'padding-box'],
+	['image-orientation', 'flip'],
```
jeddy3, line 86:
```suggestion
	['text-decoration-line', ['blink']],
	['box-sizing', ['padding-box']],
	['image-orientation', ['flip']],
```

Can these be an array one value to simplify the `isArray` and `typeOf` conditionals?

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+			const parsedValue = valueParser(getDeclarationValue(decl));
+
+			parsedValue.walk((node) => {
+				if (node.type !== 'word') return;
+
+				const keyword = node.value.toLowerCase();
+
+				if (typeof value === 'string' && value !== keyword) return;
+
+				if (Array.isArray(value) && !value.includes(keyword)) return;
```
jeddy3, line 133:
```suggestion
				if (Array.isArray(keywords) && !keywords.includes(actualKeyword)) return;
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+		if (!validOptions) return;
+
+		root.walkDecls((decl) => {
+			const declarationValue = decl.value;
+
+			if (optionsMatches(secondaryOptions, 'ignoreKeywords', declarationValue)) return;
+
+			const value = PROPERTY_NAME_TO_KEYWORD.get(decl.prop.toLowerCase());
+
+			if (!value) return;
```
jeddy3, line 122:
```suggestion
			if (!keywords) return;
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+				actual: secondaryOptions,
+				possible: {
+					ignoreKeywords: [isString, isRegExp],
+				},
+			},
+		);
+
+		if (!validOptions) return;
+
+		root.walkDecls((decl) => {
```
jeddy3, line 115:
```suggestion
		root.walkDecls((decl) => {
		  if (!isStandardSyntaxDecl(decl)) return;
```

jeddy3 (changes requested):
Thank you for making the changes. The rule is shaping up. 

I've made another pass and requested some changes.

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+	['text-decoration-line', 'blink'],
+	['box-sizing', 'padding-box'],
+	['image-orientation', 'flip'],
+	['min-height', ['intrinsic', 'min-intrinsic']],
+	['min-width', ['intrinsic', 'min-intrinsic']],
+	['max-height', ['intrinsic', 'min-intrinsic']],
+	['max-width', ['intrinsic', 'min-intrinsic']],
+	['height', ['intrinsic', 'min-intrinsic']],
+	['width', ['intrinsic', 'min-intrinsic']],
+	['word-break', 'break-word'],
```
Mouvedia, line 93:
```suggestion
	['word-break', ['break-word']],
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+
+			const parsedValue = valueParser(getDeclarationValue(decl));
+
+			parsedValue.walk((node) => {
+				if (node.type !== 'word') return;
+
+				const keyword = node.value.toLowerCase();
+
+				if (typeof value === 'string' && value !== keyword) return;
+
```
Mouvedia, line 132:
```suggestion
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
Mouvedia:
```suggestion
				const messageArgs = [declarationValue, prop];
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
Mouvedia: That's a good refactor because it enables us to use the plural later on.

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
Mouvedia: I would have preferred `currentKeyword`/`fixedKeyword` but I went ahead with your proposal anyway.

lib/rules/declaration-property-value-keyword-no-deprecated/README.md
romainmenke: I suggested including `given an MDN page` because I've encountered this multiple times :)

I would find something on MDN, try it and find that it didn't work or wasn't on a standards track anymore.

Dev rel for browser engines will often contribute to MDN for features they are prototyping. It happens that the docs land, but the feature does not.

Or MDN authors will work from a spec that only has a partial implementation. For MDN authors it isn't always immediately clear what has or hasn't shipped and they might include too much too quickly.

Inclusion on MDN makes such features very visible to CSS authors and is a strong signal that something can be used. Even when the docs sometimes do not reflect the reality of implementations.

Authors might still leave the CSS thinking it will be a progressive enhancement. "It will work tomorrow".

----

I think that including features that were only documented but never shipped is still within the spirit of this rule :)

Mouvedia: > I've encountered this multiple times :)

Do you have examples of such cases?

---

> I think that including features that were only documented but never shipped is still within the spirit of this rule :)

Ill add it back since you feel strongly about it.

lib/rules/declaration-property-value-keyword-no-deprecated/README.md
```diff
@@ -0,0 +1,75 @@
...
+This rule flags keywords that were removed or deprecated after being in the CSS specifications, including editor drafts, and were subsequently either:
+
+- shipped in a stable version of a browser
+- shipped by a developer channel/edition browser
+- shipped but behind experimental flags
+- polyfilled with some adoption before any browser actually shipped
```
Mouvedia, line 17:
```suggestion
- polyfilled with some adoption before any browser actually shipped
- had an MDN page at one point in time
```

Mouvedia:
@jeddy3 is there anything left to do?

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,174 @@
...
+	infotext: 'CanvasText',
+	menutext: 'CanvasText',
+	windowtext: 'CanvasText',
+	buttonhighlight: 'ButtonFace',
+	buttonshadow: 'ButtonFace',
+	threedface: 'ButtonFace',
+	inactivecaptiontext: 'GrayText',
+};
+const COLOR_PROPERTIES = [...singleValueColorProperties.values()].map(
+	(value) => /** @type {[string, Record<String, string>]} */ ([value, DEPRECATED_COLORS]),
```
Mouvedia, line 53:
```suggestion
	(value) => /** @type {[string, Record<string, string>]} */ ([value, DEPRECATED_COLORS]),
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.cjs
```diff
@@ -0,0 +1,178 @@
...
+	infotext: 'CanvasText',
+	menutext: 'CanvasText',
+	windowtext: 'CanvasText',
+	buttonhighlight: 'ButtonFace',
+	buttonshadow: 'ButtonFace',
+	threedface: 'ButtonFace',
+	inactivecaptiontext: 'GrayText',
+};
+const COLOR_PROPERTIES = [...properties.singleValueColorProperties.values()].map(
+	(value) => /** @type {[string, Record<String, string>]} */ ([value, DEPRECATED_COLORS]),
```
Mouvedia, line 56:
```suggestion
	((value) => /** @type {[string, Record<string, string>]} */ ([value, DEPRECATED_COLORS]),
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.cjs
```diff
@@ -0,0 +1,178 @@
...
+const setDeclarationValue = require('../../utils/setDeclarationValue.cjs');
+const properties = require('../../reference/properties.cjs');
+const validateOptions = require('../../utils/validateOptions.cjs');
+
+const ruleName = 'declaration-property-value-keyword-no-deprecated';
+
+const messages = ruleMessages(ruleName, {
+	expected: (actual, expected) => `Expected "${actual}" to be "${expected}"`,
+	rejected: (keyword, property) =>
+		`Unexpected deprecated keyword "${keyword}" for property "${property}"`,
```
Mouvedia, line 22: All rules respect a certain order—for x coming first—except `media-feature-name-unit-allowed-list`.
So Ill conform to our conventions.

types/stylelint/index.d.ts
```diff
@@ -386,6 +386,10 @@ declare namespace stylelint {
 		'declaration-property-value-disallowed-list': CoreRule<
 			Record<string, OneOrMany<StringOrRegex>>
 		>;
+		'declaration-property-value-keyword-no-deprecated': CoreRule<
+			true,
+			{ ignoreKeywords: OneOrMany<StringOrRegex> }
```
Mouvedia, line 391:
```suggestion
			{ ignoreKeywords: OneOrMany<StringOrRegex> },
			AutofixMessage & { rejected: (property: string, keyword: string) => string }
```

jeddy3 (approved):
LGTM, thank you. 

Let's wait for a second approval before merging.

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,174 @@
...
+	threedface: 'ButtonFace',
+	inactivecaptiontext: 'GrayText',
+};
+const COLOR_PROPERTIES = [...singleValueColorProperties.values()].map(
+	(value) => /** @type {[string, Record<string, string>]} */ ([value, DEPRECATED_COLORS]),
+);
+const PROPERTY_NAME_TO_KEYWORDS = new Map([
+	[
+		'appearance',
+		/** @type {string[] | Record<string, string>} */ ({
```
ybiquitous, line 58: [suggestion] To simplify the logic, how about separating the constant into two objects like this?

`PROPERTY_NAME_TO_KEYWORDS` => `PROPERTY_NAME_TO_KEYWORD_MAPS` and `PROPERTY_NAME_TO_KEYWORD_LISTS`

```suggestion
const PROPERTY_NAME_TO_KEYWORD_MAPS = new Map([
	[
		'appearance',
		{
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,174 @@
...
+	['text-decoration-line', ['blink']],
+	['box-sizing', ['padding-box']],
+	['image-orientation', ['flip']],
+	['min-height', ['intrinsic', 'min-intrinsic']],
+	['min-width', ['intrinsic', 'min-intrinsic']],
+	['max-height', ['intrinsic', 'min-intrinsic']],
+	['max-width', ['intrinsic', 'min-intrinsic']],
+	['height', ['intrinsic', 'min-intrinsic']],
+	['width', ['intrinsic', 'min-intrinsic']],
+	['word-break', ['break-word']],
```
ybiquitous, line 94:
```suggestion
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,174 @@
...
+	...COLOR_PROPERTIES,
+]);
```
ybiquitous, line 96:
```suggestion
]);
const PROPERTY_NAME_TO_KEYWORD_LISTS = new Map([
	['text-decoration-line', ['blink']],
	['box-sizing', ['padding-box']],
	['image-orientation', ['flip']],
	['min-height', ['intrinsic', 'min-intrinsic']],
	['min-width', ['intrinsic', 'min-intrinsic']],
	['max-height', ['intrinsic', 'min-intrinsic']],
	['max-width', ['intrinsic', 'min-intrinsic']],
	['height', ['intrinsic', 'min-intrinsic']],
	['width', ['intrinsic', 'min-intrinsic']],
	['word-break', ['break-word']],
]);
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,174 @@
...
+		root.walkDecls((decl) => {
+			if (!isStandardSyntaxDeclaration(decl)) return;
+
+			const { prop, value } = decl;
+
+			if (optionsMatches(secondaryOptions, 'ignoreKeywords', value)) return;
+
+			const keywords = PROPERTY_NAME_TO_KEYWORDS.get(prop.toLowerCase());
+
+			if (!keywords) return;
```
ybiquitous, line 125:
```suggestion
			const propLower = prop.toLowerCase();
			const keywordMaps = PROPERTY_NAME_TO_KEYWORD_MAPS.get(propLower);
			const keywordLists = PROPERTY_NAME_TO_KEYWORD_LISTS.get(propLower);

			if (!keywordMaps && !keywordLists) return;
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,174 @@
...
+
+			const parsedValue = valueParser(getDeclarationValue(decl));
+
+			parsedValue.walk((node) => {
+				if (node.type !== 'word') return;
+
+				const actualKeyword = node.value.toLowerCase();
+
```
ybiquitous, line 133:
```js
let message;
let messageArgs;
let fix;

if (keywordMaps) {
	const expectedKeyword = keywordMaps[actualKeyword];

	if (!expectedKeyword) return;

	// Setting message, messageArgs, and fix...
} else if (keywordLists)
	if (!keywordLists.includes(actualKeyword)) return;

	// Setting message and messageArgs...
} else {
	return;
}
```

ybiquitous (commented):
Thank you for the pull request. I think there are no more big problems. 👍🏼 

I've left some minor refactoring suggestions, so can you please take a look?

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
Mouvedia: Ill partially apply this because the `else if` permits to skip the `isPlainObject(keywords)` evaluation.

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
Mouvedia: I prefer having only one map because:

- we have only one getter
- it will be simpler to reason about if we chose to support shorthands later on (2 paths instead or 3)
- I find it clearer to read them all at a glance

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,180 @@
...
+			const parsedValue = valueParser(getDeclarationValue(decl));
+
+			parsedValue.walk((node) => {
+				if (node.type !== 'word') return;
+
+				const lowercasedValue = node.value.toLowerCase();
+				let fix;
+				let message;
+				let messageArgs;
```
Mouvedia, line 139: note

I wish `eslint-config-stylelint` would allow
```js
				let fix, message, messageArgs;
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -0,0 +1,174 @@
...
+	menutext: 'CanvasText',
+	windowtext: 'CanvasText',
+	buttonhighlight: 'ButtonFace',
+	buttonshadow: 'ButtonFace',
+	threedface: 'ButtonFace',
+	inactivecaptiontext: 'GrayText',
+};
+const COLOR_PROPERTIES = [...singleValueColorProperties.values()].map(
+	(value) => /** @type {[string, Record<string, string>]} */ ([value, DEPRECATED_COLORS]),
+);
```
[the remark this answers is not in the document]
Mouvedia: I have used `singleValueColorProperties[keyof singleValueColorProperties]` instead of `string`.

ybiquitous (approved):
Thank you. The code has now become more readable. LGTM 👍🏼

lib/rules/declaration-property-value-keyword-no-deprecated/README.md
```diff
@@ -0,0 +1,76 @@
...
+- had an MDN page at one point in time
+
+The [`fix` option](../../../docs/user-guide/options.md#fix) can automatically fix some of the problems reported by this rule.
+
+The [`message` secondary option](../../../docs/user-guide/configure.md#message) accept arguments.
+
```
jeddy3, line 23:
```suggestion
Prior art:

- [@isnotdefined/no-obsolete](https://www.npmjs.com/package/@isnotdefined/stylelint-plugin)

```

To let people know that this built-in can replace some of this plugin.

Mouvedia:
```suggestion

Prior art:

- [@isnotdefined/no-obsolete](https://www.npmjs.com/package/@isnotdefined/stylelint-plugin)

```

PR #8231, opened 2024-12-26
Document use of `isStandardSyntax*` utils

jeddy3:

This PR tweaks the documentation to imply that the `isStandardSyntax*` utils should always be used.


ryo-manba (approved):
Thank you. LGTM!

PR #8232, opened 2024-12-26
Document using type import in CommonJS config example

### LINKED ISSUE #8230: Document using type import in CommonJS config example

If a user uses CJS in their project, they will most likely go straight to the [CJS example](https://stylelint.io/user-guide/configure#:~:text=and%20type%2Dcheck.-,CommonJS%20example%3A,-module.exports) and skip the ESM example. And accordingly, he will not see an example of how types can be imported.

In the CommonJS example, add a line on top:
```js
/** @type { import('stylelint').Config } */
```

rakleed:

Closes #8230 

No, it's self-explanatory.

### DIFF
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -28,6 +28,7 @@ export default {
 CommonJS example:
 
 ```js
+/** @type {import('stylelint').Config} */
 module.exports = {
   rules: {
     "block-no-empty": true


Mouvedia (approved):
concerning `Config`

AFAIK the only difference between ESM and CJS is

https://github.com/stylelint/stylelint/blob/3751cfb4c6ba8c94dbd88db75ff6bd77a1aa70bc/types/stylelint/index.d.ts#L68-L70

which is used by

https://github.com/stylelint/stylelint/blob/3751cfb4c6ba8c94dbd88db75ff6bd77a1aa70bc/types/stylelint/index.d.ts#L91

Issue #258, opened 2015-07-01
Two exceptions cancel each other out


davidtheclark:
Given this config:

```js
    'at-rule-empty-line-before': [2, 'always', {
      except: ['blockless-group', 'all-nested'],
    }],
```

and this CSS:

```css
.ht-Input {
  @mixin input-base;
  @mixin input-skin;
}
```

We have a problem. This is a nested blockless group -- so it matches two exceptions --- so those exceptions actually cancel each other out and I get a warning here.

I'm thinking that the workable solution probably is to only attend to one matched exception at a time. So if one exception matches the context and the expectation is reversed, we don't pay attention to other exceptions. Do @stylelint/owners think that's the right fix?

jeddy3:
Just to double check, you're saying that *primary* expectation is reverse if *any* of the except conditions are met.

If so, that SGTM.

jeddy3:
Oops. Just saw your PR. That's exactly what you're saying, so ignore the above comment :)

davidtheclark:
Yes. You can see the file change in the PR making this happen --- combining all the exception checks into *one* exception check, really, that can only reverse the primary expectation *once*.

davidtheclark:
Sweet. Thanks. After merging this the whole stylesheet setup at my job runs through clean! So I'm looking forward to implementing 0.1.0 when I get back from this little trip.

jeddy3:
That's awesome!

Enjoy your trip and we'll cut a release when you're back! :)

PR #8234, opened 2024-12-26
Add replaced text

romainmenke:

See: https://github.com/stylelint/stylelint/issues/7192

This is only a very rough proposal at this time.

After working on several issues related to the reported ranges I gained some new insights and saw a way forwards for https://github.com/stylelint/stylelint/issues/7192.

In all my previous ideas I was working from the perspective of rule authors with the main questions/hurdles being:
- how to compute replacement text
- how to handle escaping
- how to handle replacements spanning multiple nodes
- how to avoid having another complex callback, next to the existing `fix()` callback, that works roughly the same but still slightly different

In this proposal I am ignoring all those issues 😅 
They simply don't exist for rule authors.

Instead we ask rule authors to return a single `PostCSS.Node` that encompasses all nodes they altered in a fix callback. In most cases this will be the node they altered. In extreme cases this might be the root node.

From this node we do all needed computations for rule authors:
- are there conflicts with previous fixes?
- what is the minimal range and replacement text?

To avoid conflicts I have used a simpler `Set` approach that keeps track of altered nodes, but it should instead be done by checking for overlapping ranges. For a draft the current approach is good enough.

The compromise in this approach is that any area in a source file can only have one corresponding fix per Stylelint run. If multiple overlapping fixes need to be applied, then end users will need to run their flows multiple times. Each time fixing all issues before starting a next run.

I have no idea if this current approach works as the test framework doesn't support this new data. But I kinda like this direction for its apparent simplicity.

### DIFF
--- a/lib/rules/media-feature-range-notation/index.mjs
+++ b/lib/rules/media-feature-range-notation/index.mjs
@@ -94,6 +94,8 @@ const rule = (primary) => {
 						if (expectedMediaQueryList === atRule.params) return;
 
 						atRule.params = expectedMediaQueryList;
+
+						return atRule;
 					};
 
 					const hasFix =
--- a/lib/utils/report.cjs
+++ b/lib/utils/report.cjs
@@ -4,8 +4,10 @@
 
 const validateTypes = require('./validateTypes.cjs');
 const constants = require('../constants.cjs');
+const typeGuards = require('./typeGuards.cjs');
 
 /** @import { DisabledRangeObject, Problem, Range, RuleMessage, StylelintPostcssResult, Utils, WarningOptions } from 'stylelint' */
+/** @import { Node as PostcssNode } from 'postcss' */
 
 /**
  * Report a problem.
@@ -163,24 +165,47 @@ function isDisabled(ruleName, startLine, disabledRanges) {
 }
 
 /** @param {Problem & { line: number }} problem */
-function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
-	const { disabledRanges, config = {}, fixersData } = stylelint;
+function isFixApplied({ node, fix, line, result: { stylelint }, ruleName }) {
+	const { disabledRanges, config = {}, fixersData, fixedNodes } = stylelint;
 
-	if (!validateTypes.isFunction(fix)) {
-		addFixData({ fixersData, ruleName, fixed: false });
-
-		return false;
-	}
+	if (!validateTypes.isFunction(fix)) return false;
 
 	const shouldFix = Boolean(config.fix && !config.rules?.[ruleName][1]?.disableFix);
 	const mayFix =
 		shouldFix && (config.ignoreDisables || !isDisabled(ruleName, line, disabledRanges));
 
-	addFixData({ fixersData, ruleName, fixed: mayFix });
-
 	if (!mayFix) return false;
 
-	fix();
+	const fixedNode = fix();
+
+	if (
+		Boolean(config.recordReplacementText) &&
+		fixedNode &&
+		fixedNode.source?.start &&
+		fixedNode.source?.end
+	) {
+		// When recording replacement text we want to ensure that there is no overlap with any other fix.
+		// We only record the first fix for each node.
+		if (fixedNodes.has(node)) return true;
+
+		addFixData({
+			fixersData,
+			ruleName,
+			...reduceFixRange(fixedNode, {
+				range: {
+					start: fixedNode.source.start,
+					end: fixedNode.source.end,
+				},
+				text: fixedNode.toString(),
+			}),
+		});
+
+		addFixedNode(fixedNodes, fixedNode);
+	} else {
+		addFixData({ fixersData, ruleName });
+
+		addFixedNode(fixedNodes, node);
+	}
 
 	return true;
 }
@@ -190,13 +215,85 @@ function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
  * @param {StylelintPostcssResult['fixersData']} o.fixersData
  * @param {string} o.ruleName
  * @param {Range} [o.range]
- * @param {boolean} o.fixed
+ * @param {string} [o.text]
  * @todo stylelint/stylelint#7192
  */
-function addFixData({ fixersData, ruleName, range, fixed }) {
+function addFixData({ fixersData, ruleName, range, text }) {
 	const ruleFixers = (fixersData[ruleName] ??= []);
 
-	ruleFixers.push({ range, fixed });
+	ruleFixers.push({ range, text });
+}
+
+/**
+ * @param {WeakSet<PostcssNode>} fixedNodes
+ * @param {PostcssNode} node
+ */
+function addFixedNode(fixedNodes, node) {
+	fixedNodes.add(node);
+
+	if (typeGuards.isContainer(node)) {
+		node.walk((childNode) => {
+			fixedNodes.add(childNode);
+		});
+	}
+}
+
+/**
+ * @param {PostcssNode} node
+ * @param {{range: Range, text: string}} fixData
+ * @returns {{range: Range, text: string}}
+ */
+function reduceFixRange(node, fixData) {
+	if (!validateTypes.isNumber(node.source?.start?.offset) || !validateTypes.isNumber(node.source?.end?.offset)) {
+		return fixData;
+	}
+
+	const stringRepresentation = node.source.input.css.slice(
+		node.source?.start?.offset,
+		node.source?.end?.offset,
+	);
+
+	let replacementStartOffset = 0;
+	let startOffset = node.source?.start?.offset;
+
+	for (let i = 0; i < stringRepresentation.length; i++) {
+		const a = stringRepresentation[i];
+		const b = fixData.text[i];
+
+		if (a !== b) break;
+
+		startOffset++;
+		replacementStartOffset++;
+	}
+
+	const start = node.positionInside(startOffset);
+
+	start.offset = startOffset;
+
+	let replacementEndOffset = fixData.text.length;
+	let endOffset = node.source?.end?.offset;
+
+	for (let i = stringRepresentation.length - 1; i >= 0; i--) {
+		const a = stringRepresentation[i];
+		const b = fixData.text[i];
+
+		if (a !== b) break;
+
+		endOffset--;
+		replacementEndOffset--;
+	}
+
+	const end = node.positionInside(endOffset);
+
+	end.offset = endOffset;
+
+	return {
+		text: fixData.text.slice(replacementStartOffset, replacementEndOffset),
+		range: {
+			start,
+			end,
+		},
+	};
 }
 
 module.exports = report;
--- a/lib/utils/report.mjs
+++ b/lib/utils/report.mjs
@@ -6,8 +6,10 @@ import {
 	SEVERITY_ERROR,
 	SEVERITY_WARNING,
 } from '../constants.mjs';
+import { isContainer } from './typeGuards.mjs';
 
 /** @import { DisabledRangeObject, Problem, Range, RuleMessage, StylelintPostcssResult, Utils, WarningOptions } from 'stylelint' */
+/** @import { Node as PostcssNode } from 'postcss' */
 
 /**
  * Report a problem.
@@ -165,24 +167,47 @@ function isDisabled(ruleName, startLine, disabledRanges) {
 }
 
 /** @param {Problem & { line: number }} problem */
-function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
-	const { disabledRanges, config = {}, fixersData } = stylelint;
+function isFixApplied({ node, fix, line, result: { stylelint }, ruleName }) {
+	const { disabledRanges, config = {}, fixersData, fixedNodes } = stylelint;
 
-	if (!isFn(fix)) {
-		addFixData({ fixersData, ruleName, fixed: false });
-
-		return false;
-	}
+	if (!isFn(fix)) return false;
 
 	const shouldFix = Boolean(config.fix && !config.rules?.[ruleName][1]?.disableFix);
 	const mayFix =
 		shouldFix && (config.ignoreDisables || !isDisabled(ruleName, line, disabledRanges));
 
-	addFixData({ fixersData, ruleName, fixed: mayFix });
-
 	if (!mayFix) return false;
 
-	fix();
+	const fixedNode = fix();
+
+	if (
+		Boolean(config.recordReplacementText) &&
+		fixedNode &&
+		fixedNode.source?.start &&
+		fixedNode.source?.end
+	) {
+		// When recording replacement text we want to ensure that there is no overlap with any other fix.
+		// We only record the first fix for each node.
+		if (fixedNodes.has(node)) return true;
+
+		addFixData({
+			fixersData,
+			ruleName,
+			...reduceFixRange(fixedNode, {
+				range: {
+					start: fixedNode.source.start,
+					end: fixedNode.source.end,
+				},
+				text: fixedNode.toString(),
+			}),
+		});
+
+		addFixedNode(fixedNodes, fixedNode);
+	} else {
+		addFixData({ fixersData, ruleName });
+
+		addFixedNode(fixedNodes, node);
+	}
 
 	return true;
 }
@@ -192,11 +217,83 @@ function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
  * @param {StylelintPostcssResult['fixersData']} o.fixersData
  * @param {string} o.ruleName
  * @param {Range} [o.range]
- * @param {boolean} o.fixed
+ * @param {string} [o.text]
  * @todo stylelint/stylelint#7192
  */
-function addFixData({ fixersData, ruleName, range, fixed }) {
+function addFixData({ fixersData, ruleName, range, text }) {
 	const ruleFixers = (fixersData[ruleName] ??= []);
 
-	ruleFixers.push({ range, fixed });
+	ruleFixers.push({ range, text });
+}
+
+/**
+ * @param {WeakSet<PostcssNode>} fixedNodes
+ * @param {PostcssNode} node
+ */
+function addFixedNode(fixedNodes, node) {
+	fixedNodes.add(node);
+
+	if (isContainer(node)) {
+		node.walk((childNode) => {
+			fixedNodes.add(childNode);
+		});
+	}
+}
+
+/**
+ * @param {PostcssNode} node
+ * @param {{range: Range, text: string}} fixData
+ * @returns {{range: Range, text: string}}
+ */
+function reduceFixRange(node, fixData) {
+	if (!isNumber(node.source?.start?.offset) || !isNumber(node.source?.end?.offset)) {
+		return fixData;
+	}
+
+	const stringRepresentation = node.source.input.css.slice(
+		node.source?.start?.offset,
+		node.source?.end?.offset,
+	);
+
+	let replacementStartOffset = 0;
+	let startOffset = node.source?.start?.offset;
+
+	for (let i = 0; i < stringRepresentation.length; i++) {
+		const a = stringRepresentation[i];
+		const b = fixData.text[i];
+
+		if (a !== b) break;
+
+		startOffset++;
+		replacementStartOffset++;
+	}
+
+	const start = node.positionInside(startOffset);
+
+	start.offset = startOffset;
+
+	let replacementEndOffset = fixData.text.length;
+	let endOffset = node.source?.end?.offset;
+
+	for (let i = stringRepresentation.length - 1; i >= 0; i--) {
+		const a = stringRepresentation[i];
+		const b = fixData.text[i];
+
+		if (a !== b) break;
+
+		endOffset--;
+		replacementEndOffset--;
+	}
+
+	const end = node.positionInside(endOffset);
+
+	end.offset = endOffset;
+
+	return {
+		text: fixData.text.slice(replacementStartOffset, replacementEndOffset),
+		range: {
+			start,
+			end,
+		},
+	};
 }
--- a/types/stylelint/index.d.ts
+++ b/types/stylelint/index.d.ts
@@ -112,6 +112,7 @@ declare namespace stylelint {
 		allowEmptyInput?: boolean;
 		cache?: boolean;
 		fix?: boolean;
+		recordReplacementText?: boolean;
 		validate?: boolean;
 	};
 
@@ -149,6 +150,7 @@ declare namespace stylelint {
 		customUrls: { [ruleName: string]: string };
 		ruleMetadata: { [ruleName: string]: Partial<RuleMeta> };
 		fixersData: { [ruleName: string]: Array<FixerData> };
+		fixedNodes: WeakSet<PostCSS.Node>;
 		quiet?: boolean;
 		disabledRanges: DisabledRangeObject;
 		disabledWarnings?: DisabledWarning[];
@@ -255,7 +257,7 @@ declare namespace stylelint {
 
 	type FixerData = {
 		range?: Range;
-		fixed: boolean;
+		text?: string;
 	};
 
 	/**
@@ -851,7 +853,7 @@ declare namespace stylelint {
 		 * Optional severity override for the problem.
 		 */
 		severity?: RuleSeverity;
-		fix?: () => void | undefined | never;
+		fix?: () => PostCSS.Node | void | undefined | never;
 	};
 
 	/** @internal */
--- a/lib/formatters/__tests__/verboseFormatter.test.mjs
+++ b/lib/formatters/__tests__/verboseFormatter.test.mjs
@@ -225,8 +225,8 @@ describe('verboseFormatter', () => {
 					_postcssResult: {
 						stylelint: {
 							fixersData: {
-								'block-no-empty': [{ fixed: false }],
-								'color-hex-length': [{ fixed: true }],
+								'block-no-empty': [],
+								'color-hex-length': [{}],
 							},
 						},
 					},
@@ -254,7 +254,7 @@ describe('verboseFormatter', () => {
 		test('no warnings and plural', () => {
 			const position = { line: 0, column: 0 };
 			const range = { start: position, end: position };
-			const fixerData = { range, fixed: true };
+			const fixerData = { range };
 			const results = [
 				{
 					source: 'path/to/file.css',
@@ -291,7 +291,7 @@ describe('verboseFormatter', () => {
 					_postcssResult: {
 						stylelint: {
 							fixersData: {
-								'color-hex-length': [{ fixed: true }, { fixed: true }],
+								'color-hex-length': [{}, {}],
 							},
 						},
 					},
@@ -303,7 +303,7 @@ describe('verboseFormatter', () => {
 					_postcssResult: {
 						stylelint: {
 							fixersData: {
-								'function-name-case': [{ fixed: true }, { fixed: false }],
+								'function-name-case': [{}],
 							},
 						},
 					},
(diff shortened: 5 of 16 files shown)


lib/rules/media-feature-range-notation/__tests__/index.mjs
```diff
@@ -110,6 +110,152 @@ testRule({
...
+			column: 8,
+			endLine: 1,
+			endColumn: 24,
+			fix: {
+				range: {
+					start: { line: 1, column: 9 },
+					end: { line: 1, column: 19 },
+				},
+				text: 'width >=',
+			},
```
romainmenke, line 176: Really excited about this result.

I think it is really neat that the reported range and text is so narrow while requiring so little effort from plugin authors.

lib/rules/media-feature-range-notation/index.mjs
```diff
@@ -94,6 +94,8 @@ const rule = (primary) => {
 						if (expectedMediaQueryList === atRule.params) return;
 
 						atRule.params = expectedMediaQueryList;
+
+						return atRule;
```
romainmenke, line 98: This is the change most rule authors will need to make to ensure compatibility with this new feature.

They only need to return a PostCSS node that encloses all changes in their `fix()` callback.

Mouvedia: If it's not too perf heavy, this looks like a very elegant solution.

romainmenke: I've since updated this to a separate field `fixedNode` instead of a return value on `fix()`. This is more in line with existing values.

lib/utils/report.cjs
```diff
@@ -162,41 +173,144 @@ function isDisabled(ruleName, startLine, disabledRanges) {
...
+	const fixedNodeRange = {
+		start: fixedNode.source.start,
+		end: fixedNode.source.end,
+	};
 
-	if (!mayFix) return false;
+	// When recording replacement text we want to ensure that there is no overlap with any other fix.
+	// We only record the first fix for each node.
+	if (fixedRanges.some((fixedRange) => rangesOverlap(fixedRange, fixedNodeRange)))
+		return { applied: false };
```
Mouvedia, line 212: > We only record the first fix for each node.

Looking at https://github.com/stylelint/stylelint/pull/8234/files#diff-b429bc70b09564c65d1a7f4436fa67271400e65e9fbaf87262a9217f966eeee4R304 the description may be inaccurate.
i.e. it's not necessarily the same node (i.e. it may be a child or even a sibling if the previous fix was significant)

---

More importantly the  [`"text"`](https://github.com/stylelint/stylelint/pull/8234/files#diff-9a4b261e295bef1ccce813aa194ac4023a1dcf98236667889964b1ba6fbc551fR176) of the already stored `fixedData` has the text that corresponds to that previous state at the time of that "first" fix which takes precedence.

Hence if a latter fix falls within that range the `text` will not reflect the state after the following fixes.
i.e. before the `return` at l212 we have adjust the `text` even if we keep the `range` untouched
If you diff after the lint has been done all the matching `fixData.range` must be updated.
If you diff per report then only the following ones must have need `fixData`.

Of course all of this matters only if you are planning to store the `text` alongside its corresponding `range` for a later retrieval;
 which seems to be the case looking at https://github.com/stylelint/stylelint/pull/8234/files#diff-cba25f954347e963307879ca83b95b319e9003d9b4b51fd350d3f4110566d3a8R120

tl;dr:

I think that the condition `if ('range' in fixData)` of
```
	if ('range' in fixData) {
		warningProperties.fix = {
			range: fixData.range,
			text: fixData.text,
		};
	}
```
prevents us from updating adding `warningProperties.fix` for the overlap cases because we failed to return a new `fixData` with the updated `text`.
i.e. the non-"first" get no `warningProperties.fix` and even if they were it would be inaccurate if not modified

Did I get something wrong?
Am I overlooking something?

lib/utils/report.mjs
```diff
@@ -164,39 +175,142 @@ function isDisabled(ruleName, startLine, disabledRanges) {
...
-	addFixData({ fixersData, ruleName, fixed: mayFix });
+	const fixedNodeRange = {
+		start: fixedNode.source.start,
+		end: fixedNode.source.end,
+	};
 
-	if (!mayFix) return false;
+	// When recording replacement text we want to ensure that there is no overlap with any other fix.
+	// We only record the first fix for each node.
+	if (fixedRanges.some((fixedRange) => rangesOverlap(fixedRange, fixedNodeRange)))
```
Mouvedia, line 213: question

This implies that we never reset the initial `fixData.text` for all overlaps?
i.e. `fixedNode.toString()` is called once for that range
Wouldn't last wins be more logical?
That would require some refactors in the case that the node is different though.
e.g. sibling, child, etc.

lib/utils/report.cjs
romainmenke: > i.e. it's the `text` before any fixes


`Warning.fix.text` is the replacement text, not the original text.

- `replacement text` is any text generated by a fixer, intended to be inserted by another tool
- `original text` is the text provided to Stylelint, before any mutations

lib/utils/report.mjs
romainmenke: Why would we want the last one?
Then we would have bleed between fixers.

This is the entire concept of this change and what makes this feasible :)

By marking entire ranges as "modified" and ignoring them in subsequent `report()` calls we are sure that we only recorded a single fix from a single rule for a single region in the source code.

We must not have any overlap.

If however we constantly override we would have combined fixes.
Thereby the end user would lose all granularity.

They would receive a single `replacement text` containing the combined mutations of several rules. But all squished together in a warning for what happened to have been the last rule to run.

Mouvedia: > and _for some_ they will have the computed replacement text

That part will have to be documented so that user will have why it's missing.

romainmenke: Yes, how `replacement text` is calculated and the limitations will have to be documented.

Also see how eslint describes the same challenge: https://eslint.org/docs/latest/extend/custom-rules#conflicting-fixes

lib/utils/report.mjs
```diff
@@ -191,9 +191,9 @@ function applyFix({ fix, fixedNode, line, result: { stylelint }, ruleName }) {
 
 	if (!mayFix) return { applied: false };
 
-	fix();
-
 	if (!config.computeReplacementText) {
```
Mouvedia, line 194: once this is not a draft anymore
minor

Since we have `if (!isFn(fix))` at l185 and `if (!mayFix)` at l190, this `if` implies an `else` that throws an informative message for the user.
e.g. `"Both … have been requested resulting in a conflict; please turn off one to proceed."` (flanked by a link to the documentation)

lib/utils/report.mjs
```diff
@@ -377,3 +388,13 @@ function hasIndices(value) {
...
 }
+
+/**
+ * @param {unknown} value
+ * @returns {value is FixObject}
+ */
+function isFixObject(value) {
+	if (!isObject(value)) return false;
+
+	return 'node' in value && 'apply' in value;
```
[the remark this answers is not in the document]
romainmenke: `isNode` is from `postcss-selector-parser`, not from `postcss`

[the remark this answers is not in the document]
Mouvedia: My bad, then forget about it.

types/stylelint/index.d.ts
```diff
@@ -168,6 +170,10 @@ declare namespace stylelint {
 		severity?: Severity;
 		url?: string;
 		rule?: string;
+		fix?: {
+			range: [number, number];
+			text: string;
+		};
```
Mouvedia, line 176: Since you matched [`editInfo`](https://eslint.org/docs/latest/integrate/nodejs-api#-editinfo-type)'s shape and you have it twice, you can add

```ts
type EditInfo = {
	/** pair of 0-based indices etc. */
	range: [sourceStartOffset: number, sourceEndOffset: number];
	/** bar */
	text: string;
};
```

romainmenke: I've added the `EditInfo` type, should still be documented

types/stylelint/index.d.ts
```diff
@@ -149,6 +150,7 @@ declare namespace stylelint {
 		customMessages: { [ruleName: string]: RuleMessage };
 		customUrls: { [ruleName: string]: string };
 		ruleMetadata: { [ruleName: string]: Partial<RuleMeta> };
+		fixedRanges: Array<[number, number]>;
```
Mouvedia, line 153: > [`StylelintPostcssResult['fixersData']` could be renamed in #8234 to avoid confusion with `fixData`.
> i.e. semantic creep](https://github.com/stylelint/stylelint/pull/8269#issue-2767497294)

If you are not planning to rename `fixedRanges` nor `fixData`, I think that renaming `fixersData` to something like `fixedRules` to avoid conflicting names and to follow suit with the new `fixedX` name convention should be done in this PR.

romainmenke: > I think that renaming fixersData to something like fixedRules to avoid conflicting names and to follow suit with the new fixedX name convention should be done in this PR.

Let's do that separately to try and minimize this PR.

----

> rename `fixedRanges` nor `fixData`

Naming suggestions are welcome

lib/utils/narrowFixRange.mjs
```diff
@@ -0,0 +1,68 @@
...
+	// Try to have a context size of at least 1
+	if (startOffset === endOffset) {
+		if (startOffset > fixData.range[0]) {
+			startOffset--;
+			replacementStartOffset--;
+		} else if (endOffset < fixData.range[1]) {
+			endOffset++;
+			replacementEndOffset++;
+		}
+	}
```
romainmenke, line 62: I've noticed that the width of a range can be 0 columns. This happens on insertions. (e.g. append `;` at the end of a declaration).

An insert in the middle of a PostCSS node is unambiguous. In that case it is always clear which node should be skipped on subsequent fix attempts.

But an insert at the start or end of a node is unclear.
In such cases it might be the start of the given node, or the end of the previous node.

By widening the fix range a tiny bit we have an unambiguous range.

types/stylelint/index.d.ts
romainmenke: I've added some docs.

romainmenke:
I think this is ready for review :)

There are still some open questions:
- how do we handle users setting both `fix` and `computeReplacementText`? https://github.com/stylelint/stylelint/pull/8234#discussion_r1899946416
- bikeshedding for `computeReplacementText`?
- other things that need bikeshedding?
- should we have [standalone unit tests](https://github.com/stylelint/stylelint/blob/main/lib/__tests__/standalone.test.mjs)? I think this might be useful as standalone tests have the report and result. This will allow us to have better coverage for the resulting warnings.

_If we should have standalone unit tests, is that something you would be willing to take on @pamelalozano16?_

Thank you all in advance for taking the time to review and give feedback 🙇

types/stylelint/index.d.mts
romainmenke, line 1: Added `FixCallback` and `FixObject` but also sorted these a-z.
I don't mind reverting the sort :)

types/stylelint/index.d.ts
```diff
@@ -741,6 +745,17 @@ declare namespace stylelint {
 		source: string;
 	};
 
+	export type EditInfo = {
+		/**
+		 * The start and end offset of the edit. This is the range that is expected to be sliced out.
+		 */
+		range: [number, number];
```
Mouvedia, line 752: question

Can the end offset and start offset be equal?
i.e. `[1, 1]`

romainmenke: See:
- https://github.com/stylelint/stylelint/pull/8234#discussion_r1904713767
- https://github.com/stylelint/stylelint/blob/1b17641fe9d80efc3fa7d7aff36d2646be2bd5ca/lib/utils/rangesOverlap.mjs#L3
- https://github.com/stylelint/stylelint/blob/1b17641fe9d80efc3fa7d7aff36d2646be2bd5ca/lib/utils/narrowFixRange.mjs#L53-L62

In practice I don't see how a range could be a point with the current implementation.
This can only happen, in theory, when inserting text into a node that started out with a point range. As we can't increase the width of a range for such a node.

Nothing will or should ever produce AST nodes from empty strings as that doesn't make sense :)

pamelalozano16:
Thanks for following up on this @romainmenke, and yes! We can contribute with standalone unit tests.

types/stylelint/index.d.ts
Mouvedia: I see.

I was asking because ESLint does something peculiar when `[0]` === `[1]`.
i.e. [it inserts](https://eslint.org/docs/latest/integrate/nodejs-api#-editinfo-type)

romainmenke: Yup that is correct, but we have a slightly different implementation.
This difference in implementation is unnoticeable for both rule and tool authors.

For insertions we widen the range so that ranges are always unambiguous, as I have explained above :)

lib/createPartialStylelintResult.mjs
Mouvedia, line 1: question

Should we add `fix: {},` to l101?

lib/utils/report.mjs
```diff
@@ -213,25 +225,76 @@ function isDisabled(ruleName, startLine, disabledRanges) {
...
+		apply = fix.apply;
+		node = fix.node;
+	} else if (isFn(fix)) {
+		apply = fix;
 	}
 
-	const shouldFix = Boolean(config.fix && !config.rules?.[ruleName][1]?.disableFix);
+	if (!apply) return { applied: false };
+
+	const shouldFix = Boolean(config.fix || (config.computeReplacementText && node));
```
Mouvedia, line 251:
```suggestion
	const dryRun = config.computeReplacementText && node;
	const shouldFix = Boolean(config.fix || dryRun);
```

Mouvedia (commented):
I did a first pass; Ill check the `*.test.mjs` files later on.
Apart from the lack of documentation and the name chosen—`computeReplacementText`—it looks OK to me.

---

Do you plan to add a flag to `cli.mjs`?
e.g. `fix` option counterpart is `--fix`

lib/utils/report.mjs
romainmenke: I've written a bit about why I think we should avoid the `dry run` terminology for this feature: https://github.com/stylelint/stylelint/issues/7192#issuecomment-2585200856

lib/createPartialStylelintResult.mjs
romainmenke: I think that could be confusing as then this specific warning has an empty `fix` object while other warnings without fix data have `undefined`

lib/utils/report.mjs
Mouvedia: Leaving aside the names, do you want to create a separate variable?

romainmenke: Actually no, I applied the suggestion, but while rebuilding I found it confusing to have a variable that must not be reused.

Reusing it on line `#258` would for example be incorrect.

So I've reverted this.

----

I will take another look at this function and see if it can be improved in other ways.

romainmenke: An alternative shape:

```js
/**
 * @param {Problem & { line: number }} problem
 * @returns {boolean}
 */
function applyFix({ fix, line, result: { stylelint }, ruleName }) {
	if (!fix) return false;

	const { disabledRanges, config = {}, fixersData } = stylelint;

	if (!config.fix) return false;

	if (fixingIsDisabled(line, ruleName, config, disabledRanges)) return false;

	const apply = isFixObject(fix) ? fix.apply : fix;

	if (!apply) return false;

	apply();

	incrementFixCounter({ fixersData, ruleName });

	return true;
}

/**
 * @param {Problem & { line: number }} problem
 * @returns {{range: [number, number], text: string} | undefined}
 */
function computeReplacementText({ fix, line, result: { stylelint }, ruleName }) {
	if (!fix) return;

	const { disabledRanges, config = {}, fixedRanges } = stylelint;

	if (!config.computeReplacementText) return;

	if (fixingIsDisabled(line, ruleName, config, disabledRanges)) return;

	if (!config.ignoreDisables && isDisabled(ruleName, line, disabledRanges)) return;

	if (!isFixObject(fix) || !fix.apply || !fix.node) return;

	const { apply, node } = fix;

	if (!isNumber(node.source?.start?.offset) || !isNumber(node.source?.end?.offset)) return;

	/** @type [number, number] */
	const fixedNodeRange = [node.source.start.offset, node.source.end.offset];

	// When recording replacement text we want to ensure that there is no overlap with any other fix.
	// We only record the first fix for each node.
	if (fixedRanges.some((fixedRange) => rangesOverlap(fixedRange, fixedNodeRange))) {
		return;
	}

	// Record the current state of the node that will be fixed
	const fixedNodeClone = node.clone();

	// Apply the fix
	apply();

	// Compute the smallest range and text of the fix
	const fixData = narrowFixRange(node, {
		range: fixedNodeRange,
		text: node.toString(),
	});

	// Restore the previous state of the fixed node
	node.replaceWith(fixedNodeClone);

	// Mark the fixed range as mutated
	fixedRanges.push(fixData.range);

	return { ...fixData };
}

/**
 * @param {number} line
 * @param {string} ruleName
 * @param {Config} config
 * @returns {boolean}
 */
function fixingIsDisabled(line, ruleName, config, disabledRanges) {
	if (config.rules?.[ruleName][1]?.disableFix) return true;

	if (!config.ignoreDisables && isDisabled(ruleName, line, disabledRanges)) return true;

	return false;
}
```

Splitting fixes and computing edit info results in less confusing boolean checks I think.

Mouvedia: During my review, I wondered if I should ask you to create a separate function for the `if (config.computeReplacementText)` part.
It seemed too much since you took the time to add comments to explain each steps.

If you feel the function is too long and does too much, a refactor might be in order.
The `| { applied: boolean }` was probably hinting at that slight complexity.
i.e. several shapes being returned is sometimes a sign of something amiss
Ill let you decide.

romainmenke: I've applied my suggestion.
Please let me know if this is clearer 🙇

Mouvedia: After this refactor:

- `applyFix` should revert to its previous name (`isFixApplied`)
- `isDisabled` should be renamed `isLineDisabled` (to avoid confusion with `isFixingDisabled`)
- `fixingIsDisabled` should be renamed `isFixingDisabled`

I don't have the time right now to check everything that has changed in this commit but one thing that I noticed is that `isFn` is now only called for `fix.apply`.
cf https://github.com/stylelint/stylelint/pull/8234/files#diff-cba25f954347e963307879ca83b95b319e9003d9b4b51fd350d3f4110566d3a8L220

I like when the diff is reduced and the changes are isolated because it helps the reviewers.
It looks like a good refactor.

romainmenke: I've made some adjustments, not exactly as you suggested, please let me know if this still covers your intent?

Mouvedia: It does; `isDisabledOnLine` is much better than what I proposed.

lib/utils/report.mjs
```diff
@@ -51,7 +54,7 @@ export default function report(problem) {
 	// In quiet mode, mere warnings are ignored
 	if (quiet && ruleSeverity === SEVERITY_WARNING) return;
 
-	if (isFn(fix) && metadata && !metadata.fixable) {
+	if ((isFn(fix) || isFixObject(fix)) && metadata && !metadata.fixable) {
 		throw new Error(
 			`The "${ruleName}" rule requires "meta.fixable" to be truthy if the "fix" callback is being passed`,
```
Mouvedia, line 59: minor

This line will have to be updated to cover both `fix` and `fix.apply`.

lib/utils/report.mjs
```diff
@@ -213,27 +218,94 @@ function isDisabled(ruleName, startLine, disabledRanges) {
...
+ * @returns {{range: [number, number], text: string} | undefined}
+ */
+function computeEditInfo({ fix, line, result: { stylelint }, ruleName }) {
+	if (!fix) return;
+
+	const { disabledRanges, config = {}, fixedRanges } = stylelint;
+
+	if (!config.computeReplacementText) return;
+
+	if (isFixDisabled(line, ruleName, config, disabledRanges)) return;
```
Mouvedia, line 256: nit
opinion

I think that either [`disableFix` documentation](https://stylelint.io/user-guide/configure/#disablefix) should be updated or that `disableFix` should be ignored by `computeEditInfo` function. I am not sure what's the best course of action though.

romainmenke: Any user facing documentation still needs to happen.
I suggest we document that only those rules that are enabled both in general and for fixes will have computed edit info.

I want to avoid having a third mechanism to enable warnings while disabling some rules for edit info.

lib/utils/report.mjs
romainmenke: I think `if the "fix" callback is being passed` also covers `fix.apply` :)

We mainly want to warn that `meta.fixable` needs to be set when we detect a correct "fixer". We don't have to super technical in how we describe that fixer.

types/stylelint/index.d.ts
```diff
@@ -849,6 +865,13 @@ declare namespace stylelint {
 		column: number;
 	};
 
+	export type FixCallback = () => void | undefined | never;
+
+	export type FixObject = {
+		apply?: FixCallback;
```
Mouvedia, line 871: question

Since `fix` can also be a function and that `Function.prototype.apply` exists, it could be safer to name this `mutate`?
In practice it shouldn't matter though; I am not sure.
@ybiquitous what's your opinion?

Mouvedia, line 871: @romainmenke does this still need to be bikeshed?
e.g. `mutate`
Are you still hesitating?
Just checking so that we can move on.

romainmenke: I am in favor of `apply` :)
It's short and clear.

types/stylelint/index.d.ts
```diff
@@ -113,6 +113,7 @@ declare namespace stylelint {
 		allowEmptyInput?: boolean;
 		cache?: boolean;
 		fix?: boolean;
+		computeReplacementText?: boolean;
```
Mouvedia, line 116: Based on https://github.com/stylelint/stylelint/issues/7192#issuecomment-2585212672 and

> [To help tool makers I matched the ESLint API for the fix object on Warning.](https://github.com/stylelint/stylelint/issues/7192#issuecomment-2568284585)

It seems you are leaning towards ESLint terminology even if our implementation differs slightly.
i.e. `EditInfo` suffix is locked

Out of my suggestions—compute/generate/output/include/derive—for the verb, I think that `derive` is the most accurate, `include` would be confusing (it could be interpreted as something that accompanies fixes, i.e. it's a XOR not a AND), `output` would be confusing regarding where the data can be found (stdout/stderr, json, etc.). `add` would have the same problem as `include`.

---

tl;dr

That leaves us with `computeEditInfo`, `generateEditInfo` and `deriveEditInfo`.
For simpler alternatives you have `editInfo: true,`, `fixMetadata: true` or `replacementData: true`.
`fixMetadata` being the most the versatile for extensibility and `replacementData` the most accurate but it's unlikely that this name will permit to add new properties in the future.

ehoogeveen-medweb: > For simpler alternatives you have `editInfo: true,`, `fixMetadata: true` or `replacementData: true`. `fixMetadata` being the most the versatile for extensibility, `editInfo` the most familiar for users coming from ESLint and `replacementData` the most accurate but it's unlikely that this name will permit to add new properties in the future.

A disadvantage of `editInfo` and `fixMetadata` is that `edit` and `fix` are also verbs. In that sense `replacementData` is the least ambiguous (although I personally think it sounds a little clunky).

romainmenke: `reportEditInfo`?

Mouvedia: Users would expect supplementary elements in `results` if the boolean option was prefixed by `report` because we already have `reportDescriptionlessDisables`/`reportInvalidScopeDisables`/`reportNeedlessDisables`/`reportUnscopedDisables` that behave that way.
e.g. `descriptionlessDisables.test.mjs`

Since you liked `computeEditInfo`, let's go with that. I would have preferred something simpler like `editInfo: boolean;` but your choice is more precise.

romainmenke: I've changed it to `computeEditInfo` but happy to adjust it if there is a different preference :)

Mouvedia:
> I was hoping we could merge this separate from:
> […]
> But yes, all those things still need to happen.

Could we have these PRs be part of the same version as #8284?
Or do you want to not advertise it and include it in the next version even without the rest?

types/stylelint/index.d.ts
```diff
@@ -741,6 +745,17 @@ declare namespace stylelint {
 		source: string;
 	};
 
+	export type EditInfo = {
+		/**
+		 * The start and end offset of the edit. This is the range that is expected to be sliced out.
```
Mouvedia, line 750: Can you add that the pair are "0-based indices"?
cf https://eslint.org/docs/latest/integrate/nodejs-api#-editinfo-type

lib/utils/narrowFixRange.mjs
Mouvedia: > By widening the fix range a tiny bit we have an unambiguous range.

Iv reached these tests.
i.e. `zz` instead of `z`

I was wondering, if the first character of the next node is the same as the penultimate character of the previous node, do we have a problem?
i.e. the "context size" is too small
If we do, should we increase the number of characters checked until it differs?
i.e. 1, 2, etc. until length
Maybe it never happens or I misunderstood completely, if so you can ignore this.

lib/utils/report.mjs
```diff
@@ -213,27 +218,94 @@ function isDisabled(ruleName, startLine, disabledRanges) {
...
+	if (!fix) return false;
+
 	const { disabledRanges, config = {}, fixersData } = stylelint;
 
-	if (!isFn(fix)) {
-		return false;
-	}
+	if (!config.fix) return false;
+
+	if (isFixDisabled(line, ruleName, config, disabledRanges)) return false;
```
Mouvedia, line 232: suggestion

If you want you can reduce the number of arguments passed to `isFixDisabled`.
Marginally better? Not sure.

```suggestion
	const { config, fixersData } = stylelint;

	if (!config?.fix) return false;

	if (isFixDisabled(line, ruleName, stylelint)) return false;
```
 and inside `isFixDisabled`
```js
	const { config = {}, disabledRanges } =  stylelint;
```

types/stylelint/index.d.ts
Mouvedia, line 153:
```suggestion
		fixedRanges: Array<EditInfo['range']>;
```

lib/utils/report.mjs
romainmenke: I'd rather keep it as is.

types/stylelint/index.d.ts
romainmenke: I've replaced it fully with the ESLint wording.

types/stylelint/index.d.ts
Mouvedia: [5d9a523](https://github.com/stylelint/stylelint/commit/5d9a5235ff526344356ea63d48b48842aa7f2bb3)

lib/utils/narrowFixRange.mjs
romainmenke: The ambiguity is not one of characters being the same. It is purely mathematical.
I am not sure how to better explain it both in words and in test cases :)

romainmenke:
I really just want to avoid a change that is too large and is too complicated to review.
So I've marked the exposed parts of the new API as `@experimental`.
If we happen to resolve all parts within the same release we can remove this again.

jeddy3 (approved):
Thank you for working on this! Considerable effort has gone into it. 

Other maintainers and I may not have time to review this thoroughly. I'm reminded of the asymmetry of open-source projects; the amount of time maintainers have is vastly disproportional to the number of users. 

Rather than delay landing this or adding more friction to your continued efforts, I've approved the PR after a shallow review of the code and tests. Your features are amazingly well-considered.

romainmenke:
Thank you @jeddy3 🙇 
I agree that ideally this would be reviewed by more maintainers. Thank you for clearing the way :)

And also thank you @Mouvedia for the very thorough feedback, this really improved this change a lot 🎉

PR #8236, opened 2024-12-27
Refactor `color-hex-alpha` and `no-irregular-whitespace` to use consistent `ruleMessage` keys

Mouvedia:

N/A

I am planning to send a PR that will add the possibility to add the types of the core rule's `messages`.
i.e.
```ts
	type CoreRule<
		P extends Primary,
		S extends Secondary = CommonOptions,
		M extends RuleMessages = Messages
	> = Rule<P, S, M>;
```
This will permit to document the order and names of the arguments.
e.g.
- distinguish between `(property: string, keyword: string) => string` and `(keyword: string, property: string) => string`
- hypothetical documentation generation
- autocompletion for configurations

This preliminary PR is to prepare the codebase for that change.
i.e. these rules are the only ones that produce errors for
```ts
	type Messages = {
		[key in `expected${string}` | `rejected${string}`]: RuleMessage;
	};
```

### DIFF
--- a/lib/rules/color-hex-alpha/__tests__/index.mjs
+++ b/lib/rules/color-hex-alpha/__tests__/index.mjs
@@ -59,7 +59,7 @@ testRule({
 	reject: [
 		{
 			code: 'a { color: #ffff; }',
-			message: messages.unexpected('#ffff'),
+			message: messages.rejected('#ffff'),
 			line: 1,
 			column: 12,
 			endLine: 1,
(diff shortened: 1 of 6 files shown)


jeddy3:
Thank you for making things more consistent!

Am I right in thinking this is a refactor and doesn't impact our public API?

Mouvedia:
> Am I right in thinking this is a refactor and doesn't impact our public API?

It depends where you draw the line.
If a plugin rule is using [`checkAgainstRule`](https://stylelint.io/developer-guide/plugins/#stylelintutilscheckagainstrule) I would expect no disruption for the users.
The tests of that plugin would be broken though.

Iv opened that PR for that very reason, to enquiry whether it's safe to rename these.

jeddy3:
I think we can classify this as a refactor, then.

jeddy3 (approved):
LGTM, thank you.

Issue #7408, opened 2023-12-17
Change to flat config


jeddy3:
### What is the problem you're trying to solve?

ESLint is migrating to flat config for [some good reasons](https://eslint.org/blog/2022/08/new-config-system-part-2/#the-goals-of-flat-config), in particular:

- Logical defaults
- One way to define configs
- Use native loading for everything

One of their [main motivations is to move away from their complex cascading configs](https://eslint.org/blog/2022/08/new-config-system-part-2/#glob-based-configs-everywhere):

> Because we wanted to eliminate the config cascade of eslintrc, we had to use glob patterns to enable the same type of config overrides. We used the overrides configs as the basis for flat config.

We've never had nested configuration files that cascade. We made an early design decision to avoid that complexity, and it was, in hindsight, a good call. We've also [avoided adding competing mechanisms like `--ext`](https://github.com/stylelint/stylelint/issues/6689#issuecomment-1465130308). As such, there may not be an appetite for this change as the pain-to-gain ratio (for both our users and us) differs from ESLint's. I figured we needed a canonical issue to discuss it. 

We can pull logical defaults into a separate issue as it's not reliant on a flat configuration. This leaves us with native loading for everything and one way to define configs.

For example, the following `extends` and `overrides`:

```jss
module.exports = {
  "extends": ["stylelint-config-standard"],
  "rules": {
    "unit-allowed-list": ["rem"]
  },
  "overrides": [
    {
      "files": "**/components/*.scss"
      "rules": {
        "function-disallowed-list": ["calc"]
      }
    }
  ]
}
```

Becomes:

```js
import configStandard from "stylelint-config-standard";

export default [
  configStandard,
  {
    "rules": {
      "unit-allowed-list": ["rem"]
    }
  },
  {
    "files": "**/components/*.scss"
    "rules": {
      "function-disallowed-list": ["calc"]
    }
  }
];
```

Advantages:

- it'll simplify our [points-of-extension](https://stylelint.io/about/vision#extensible) code (read: project sustainability)
- it should improve the config resolution performance for our users (https://github.com/stylelint/stylelint/issues/6383)

The disadvantages are that it'd be a massive breaking change involving everyone shifting to JavaScript config files (from JSON and YAML) and the new syntax or a lot of effort on our part to support the old config like ESLint.

Any other thoughts?

If we do go ahead, it'll likely happen for `18.0.0` or later so that we focus on new rules in `16.x` and finishing our migration to ESM for `17.0.0`.

[...]

Issue #6593, opened 2023-01-28
Become an ESLint plugin


jeddy3:
ESLint has an exciting proposal (https://github.com/eslint/rfcs/pull/99) for Stylelint to become an ESLint plugin.

We're a small team of volunteers working on Stylelint. A couple of us share the sentiment that it would free us up to add more rules to help people rather than spend our time on file-finding, config, caching etc. It'd also let us tap into ESLint's pool of integrations, e.g. its VS Code extension. 

This issue is to draw people's attention to the RFC. Please share your thoughts here on whether you think this is a good idea or not. If you've thoughts about the particulars of the RFC itself (especially the technical nitty-gritty), please then comment [there](https://github.com/eslint/rfcs/pull/99) instead.

hudochenkov:
I'm not sure this is a good or bad idea. There is a positives, like maintenance of the core of the linter become non-issue, as it would be on ESLint.

On the other side we would need to write and maintain completely new thing — plugin for ESLint, to bridge the gap between ESLint core and Stylelint rules.

Change all rules to work with ESLint. Whatever result of that RFC would be, rules would have to be changed for new core.

Lose all of Stylelint plugins. Even if there would be a way to have plugins for plugins (see [my questions](https://github.com/eslint/rfcs/pull/99#issuecomment-1404367511) in RFC). Stylelint plugins authors would need to create ESLint version of their plugin.

User would have to navigate in all of this, change their configurations, editor settings, packages, CI.

Airkro:
I also struggled with the differences between `Stylelint` and `ESlint`.

If the team just wants to save development costs, I have an immature suggestion:

Collaborate with the `ESLint` team to create some shared `@linter-utils/xxx`  ...

ybiquitous:
FYI, the ESLint team has published an official CSS plugin: https://github.com/eslint/css

I'm curious about what the ESLint team will support for the Stylelint community. For example, add more rules, custom syntax support, third-party plugin support, etc.
(@ota-meshi may have some info about this since he's a member of ESLint)

ota-meshi:
I don't know much about that 😓

jeddy3:
@ybiquitous I'm also curious as these are interesting times.

It feels like the linting landscape has fragmented recently and seems to be becoming:

- ESLint: a linter for many languages
- Prettier: a formatter for many languages
- Biome: an everything for many languages 
- ZeroVoid (Oxc): an everything for JavaScript

[...]

Issue #5139, opened 2021-02-09
Switch to using jsonc instead of json


thernstig:
> What is the problem you're trying to solve?

When creating a `.stylelintrc.json` as per the [Getting started](https://stylelint.io/user-guide/get-started) guide, it does not allow for comments. Ideally when some rules are added you want to make a comment to explain to future readers why a rule was used or enabled. This can only be done if the `.stylelintrc.json` is interpreted as `jsonc`.

Note that `.eslintrc.json` is interpreted as `jsonc`.

> What solution would you like to see?

For `.stylelintrc.json` to be interpreted as `jsonc`.

alexander-akait:
Why do not use `js` file? Need more abilities - use more powerful format

thernstig:
The same thing can be said for ESLint, but the opted to support `jsonc`. More users are also comfortable with JSON as it is a universal format. In addition, JSON supports schemas, which we can use with http://json.schemastore.org/stylelintrc.json to make sure our `.json` file uses proper configuration keywords.

Thus using JSON is preferred. In addition, this feature request does not take away the experience from anyone to use `js`, it just adds very good options for those using `json`.

jeddy3:
@thernstig Thanks for the request and for using the template.

(The schema store stuff blows my mind every time I see it!)

We use [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) to load the configuration object. It supports [custom loaders](https://github.com/davidtheclark/cosmiconfig#loaders), so it possible. Here's Prettier [adding JSON5 config file support](https://github.com/prettier/prettier/pull/8957). However, we'll want to discuss the impact, if any, on performance,  package size and maintenance for our small team of contributors. 

Labelling as a discussion so that others can, if they like, chime in.

vasilii-kovalev:
What's the status of the discussion?

thernstig:
I do not see how this is a big deal. Most large projects (babel, ESLint, VS Code config etc.) all use jsonc (or JSON5, which is even more powerful).  It's a very small deal switching to using jsonc or JSON5.

snebjorn:
I'd appreciate json5 support. But can't that just be the default json parser? json is valid json5 so using a json5 parser to parse json should not be a problem.
[...]

PR #8237, opened 2024-12-28
Document preferred config file format

jeddy3:

It's a documentation fix that relates to https://github.com/stylelint/create-stylelint/pull/92#issuecomment-2564129562

A while back, we:

- updated many of our examples to use ESM
- decided to only implement things that would be compatible with a flat config

This PR brings the _getting started_ and _configuring_ docs in line with that by recommending people use a `stylelint.config.{js,mjs}` file.

### DIFF
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -2,7 +2,7 @@
 
 Stylelint expects a configuration object.
 
-Starting from the current working directory, Stylelint searches upwards until it finds a `stylelint.config.{js,mjs}` file that exports one. Or, you can use the [`--config`](cli.md#--config--c) or [`configFile`](options.md#configfile) option to short-circuit the search.
+Starting from the current working directory, Stylelint searches upwards until it finds a `stylelint.config.js` (or `stylelint.config.mjs`, `stylelint.config.cjs`) file that exports one. Or, you can use the [`--config`](cli.md#--config--c) or [`configFile`](options.md#configfile) option to short-circuit the search.
 
 > [!NOTE]
 > Stylelint currently supports other configuration formats, but we may remove these in the future:
--- a/docs/user-guide/configure.md
+++ b/docs/user-guide/configure.md
@@ -2,19 +2,9 @@
 
 Stylelint expects a configuration object.
 
-Starting from the current working directory, Stylelint searches upwards until it finds a `stylelint.config.js` (or `stylelint.config.mjs`, `stylelint.config.cjs`) file that exports one. Or, you can use the [`--config`](cli.md#--config--c) or [`configFile`](options.md#configfile) option to short-circuit the search.
+Starting from the current working directory, Stylelint searches upwards until it finds a `stylelint.config.js` file that exports one. You can use the [`--config`](cli.md#--config--c) or [`configFile`](options.md#configfile) options to short-circuit the search.
 
-> [!NOTE]
-> Stylelint currently supports other configuration formats, but we may remove these in the future:
->
-> - `.stylelintrc` file in JSON or YAML format.
-> - `.stylelintrc.json` file or `stylelint` property in `package.json`
-> - `.stylelintrc.yml` or `.stylelintrc.yaml` file
-> - `.stylelintrc.js` file using `export default` or `module.exports`
-> - `.stylelintrc.mjs` file using `export default`
-> - `stylelint.config.cjs` or `.stylelintrc.cjs` file using `module.exports`
->
-> For `.js` files, the module system depends on your [default module system configuration](https://nodejs.org/api/packages.html#determining-module-system) for Node.js (e.g., `"type": "module"` in `package.json`).
+The style of export depends on your [default module system configuration](https://nodejs.org/api/packages.html#determining-module-system) for Node.js, e.g., `"type": "module"` in your `package.json` file. You can use the `stylelint.config.mjs` or `stylelint.config.cjs` filename to be explicit.
 
 Example `stylelint.config.mjs` file:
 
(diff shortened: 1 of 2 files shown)


docs/user-guide/configure.md
```diff
@@ -1,53 +1,32 @@
...
-  rules: {
-    "block-no-empty": true
-  }
-};
-```
+Starting from the current working directory, Stylelint searches upwards until it finds a `stylelint.config.{js,mjs}` file that exports one. Or, you can use the [`--config`](cli.md#--config--c) or [`configFile`](options.md#configfile) option to short-circuit the search.
 
 > [!NOTE]
-> The `@type` JSDoc annotation enables Typescript to autocomplete and type-check.
+> Stylelint currently supports other configuration formats, but we may remove these in the future:
```
Mouvedia, line 8: > …but we may remove these in the future

I was pretty much on page with what you said before:

> [It probably makes sense if we delay making potentially disruptive changes (e.g. #7408) as our users likely appreciate the stability of Stylelint.](https://github.com/stylelint/stylelint/issues/6593#issuecomment-2470778938)

Hence I don't agree with removing the support of these formats and even I would like for us to go even further and support `.jsonc`.
see #5139

jeddy3: In that post, I suggested delaying potentially disruptive changes until the dust settles. However, we may still adopt a flat configuration later. 

People can still use all the formats. The documentation change is to recommend one over the others. The one that'll be most compatible with a flat config. 

This change arguably makes the onboarding experience easier for first-time users, too; a recommended happy path through the configuration process rather than frontloading with all the choices.

We did something similar with the getting started page when we introduced the init script and simplified the steps for linting CSS.

romainmenke, line 8: As this list also includes `.mjs` and `.js` it isn't clear which format is unlikely to be removed in the future.

Should we split this up in two separate notes?
- one note listing all currently supported formats (maybe with `.mjs` and `.js` appearing first)
- one not listing all formats that are potential candidates for deprecation

docs/user-guide/configure.md
```diff
@@ -1,53 +1,32 @@
...
+>
+> - `.stylelintrc` file in JSON or YAML format.
+> - `.stylelintrc.json` file or `stylelint` property in `package.json`
+> - `.stylelintrc.yml` or `.stylelintrc.yaml` file
+> - `.stylelintrc.js` file using `export default` or `module.exports`
+> - `.stylelintrc.mjs` file using `export default`
```
romainmenke, line 14: Listing `.mjs` and `.js` first might make it more likely that users will pick that. As some users might not have a preference of their own and will just pick the first in the list.

romainmenke (commented):
Thank you for working on this @jeddy3
I've left a few suggestions.

docs/user-guide/configure.md
jeddy3: I've reworked the section in https://github.com/stylelint/stylelint/pull/8237/commits/58b0544653c4910d61af631529da72ade5952fb9 to:

- frontload all the bits relevant to the recommended route
- clarify that the note relates to formats and locations (and users should probably avoid using either)

Is this clearer, or do we still need to split the notes? I found it tricky to split formats and locations (e.g., filenames, and the `package.json` property) in a way that made sense.

jeddy3:
@ybiquitous & @ybiquitous Thank you for the reviews. I've committed your suggestions and reworked bits to address your other comments.

docs/user-guide/configure.md
romainmenke: I agree :)
This version looks good to me, thank you!

romainmenke (approved):
Thank you. LGTM 👍

docs/user-guide/configure.md
Mouvedia: > People can still use all the formats.

ESLint has the `ESLINT_USE_FLAT_CONFIG` env variable hence I would expect for us to do the same and maintain the compatibility for a while. If its just a recommendation, it's fine.

docs/user-guide/get-started.md
```diff
@@ -89,16 +91,17 @@ If you want to lint more than one language or container, you can use the [`overr
 
 For example, to lint CSS files and the CSS within Lit Elements you can update your configuration to:
 
-```json
-{
-  "extends": ["stylelint-config-standard"],
-  "overrides": [
+```js
+/** @type {import('stylelint').Config} */
```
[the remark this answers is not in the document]
ybiquitous: `@type` seems still necessary: 👇🏼 

image

image

If we want to use `@import`, both comments seem needed, but it looks a bit verbose 🤔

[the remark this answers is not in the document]
romainmenke: Good catch @ybiquitous, I hadn't thought of that 🙇 
I agree with keeping the original notation for brevity.

docs/user-guide/get-started.md
```diff
@@ -38,12 +38,13 @@ For example, to lint SCSS you can extend the [SCSS community config](https://www
 npm install --save-dev stylelint stylelint-config-standard-scss
 ```
 
-2\. Create a `.stylelintrc.json` configuration file in the root of your project with the following content:
+2\. Create a `stylelint.config.mjs` configuration file in the root of your project with the following content:
```
ybiquitous, line 41: [suggestion] `.mjs` or `.cjs` depends on the `type` value in `package.json`, so `.js` may be a bit more accurate in many cases.

```suggestion
2\. Create a `stylelint.config.js` configuration file in the root of your project with the following content:
```

Or if you want to use `.mjs`, some notes may be necessary, e.g.

```
Note: this example assumes that your project works under ESM.
```

Although that seems a bit verbose.

jeddy3:
Thanks for the reviews. I've committed the suggestions and addressed the latest comments.

ybiquitous (approved):
Thank you. LGTM 👍🏼

Issue #8242, opened 2024-12-29
Deprecate ambiguous position arguments in `utils.report()`


romainmenke:
### What is the problem you're trying to solve?

see: 
- https://github.com/stylelint/stylelint/issues/8126
- https://github.com/stylelint/stylelint/pull/8217

As discussed in https://github.com/stylelint/stylelint/issues/8126 we want to deprecate various ways of providing ambiguous or invalid position arguments to `report()`.

### What solution would you like to see?

We want to emit a warning each time a rule author does not adhere to the type described in:

https://github.com/stylelint/stylelint/blob/8a054dda6c176c0e4b42114bc6bbf95cfabdfb4b/types/stylelint/index.d.ts#L817-L846

```ts
type ProblemLocation = {
	/**
	* The inclusive start position of the problem, relative to the
	* node's source text. If provided, this will be used instead of
	* `index`.
	*/
	start: { line: number, column: number };
	/**
	* The exclusive end position of the problem, relative to the
	* node's source text. If provided, this will be used instead of
	* `endIndex`.
	*/
	end: { line: number, column: number };
} | {
	/**
	* The inclusive start index of the problem, relative to the node's
	* source text.
	*/
	index: number;
	/**
	* The exclusive end index of the problem, relative to the node's
	* source text.
	*/
	endIndex: number;
} | {
	word: string;
} | object;
```

This deprecation is intended to improve the accuracy of lint warnings and to reduce bugs.
The actual change will likely be made in Stylelint `v17`.
Please reach out if we overlooked something.

PR #8244, opened 2024-12-29
Deprecate ambiguous position arguments in `utils.report()`

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/8242

No, it's self-explanatory.


lib/utils/report.mjs
```diff
@@ -115,6 +125,45 @@ export default function report(problem) {
 	result.warn(warningMessage, warningProperties);
 }
 
+/**
+ * @param {Problem} problem
+ */
+function problemRangeDeprecations(problem) {
+	// TODO: `config.quietDeprecationWarnings` is not available in `problem`
```
ybiquitous, line 132: [question] Is it possible to use `config.quietDeprecationWarnings` from `result.stylelint`?

lib/utils/report.mjs
```diff
@@ -115,6 +125,45 @@ export default function report(problem) {
...
+
+	if (!problem.node) {
+		emitDeprecationWarning(
+			`Omitting \`node\` in \`report\` is deprecated ("${problem.ruleName}").`,
```
ybiquitous, line 136: [suggestion] How about clarifying `node` and `report` like this? It may be a bit longer, though:

```suggestion
			`Omitting the \`node\` argument in the \`report()\` function is deprecated ("${problem.ruleName}").`,
```

For example:

https://github.com/stylelint/stylelint/blob/003bbc416c48e5bf774f32b833aaa0d2259950bc/lib/utils/report.mjs#L66

lib/utils/report.mjs
```diff
@@ -115,6 +125,45 @@ export default function report(problem) {
 	result.warn(warningMessage, warningProperties);
 }
 
+/**
+ * @param {Problem} problem
+ */
+function problemRangeDeprecations(problem) {
+	// TODO: `config.quietDeprecationWarnings` is not available in `problem`
+
+	if (!problem.node) {
```
ybiquitous, line 134: [suggestion] To check the case when `node` is falsy like `undefined`, does the following make sense?

```suggestion
	if ('node' in problem) {
```

lib/utils/validateTypes.mjs
```diff
@@ -118,3 +120,41 @@ export function assertString(value) {
 	// eslint-disable-next-line no-console
 	console.assert(isString(value), `"${value}" must be a string`);
 }
+
+/**
+ * @param {any} value
```
ybiquitous, line 125: [suggestion] Can we use `unknown` like other functions?

```suggestion
 * @param {unknown} value
```

.changeset/good-crews-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Deprecated: ambiguous position arguments in `report`
```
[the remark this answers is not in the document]
romainmenke: Thank you for suggesting this.

`utils.report()` seems good to me!

lib/utils/report.mjs
romainmenke: Unfortunately not.
`quietDeprecationWarnings` is not included there: https://github.com/stylelint/stylelint/blob/3a903800248fcccd4968e8e0dc4a76a4d8b88ff4/types/stylelint/index.d.ts#L89-L116

As I understand it, the value does exist higher up the stack, but isn't passed to the `config` object present on `result.stylelint`

ybiquitous: Ah, from `config.quietDeprecationWarnings`, I misunderstood `quietDeprecationWarnings` would be in `Config`, sorry.

Instead, is it possible to add `quietDeprecationWarnings` to `StylelintPostcssResult` like `quiet`?

jeddy3:
@romainmenke thank you for working on this. I greatly appreciate all your efforts around reporting and granular fixes!

Judging by https://github.com/AndyOGo/stylelint-declaration-strict-value/issues/379, our last deprecation (related to `context.fix`) may have caused some disruption. The deprecation was intended to inform custom rule authors but also impacted end-users. Are there things we can do to minimise their disruption this time around? 

Perhaps we can:

- ensure the deprecation warning is only displayed once (I think the last time the message was repeated many times)
- add the ability to quiet warnings via a config property (to round out the CLI flag and Node.js options) 

If we think these are worth pursuing, I can open issues to discuss.

romainmenke:
> ensure the deprecation warning is only displayed once

This definitely seems important and doable to me :)
We should indeed avoid impact on end users and only emit the minimal amount of messages so that rule/plugin authors can be informed.

lib/utils/report.mjs
romainmenke: Good suggestion!
I've made some changes, please let me know if these are sufficiently clear now.

lib/utils/report.mjs
romainmenke: I am unsure 🤔
A value of `undefined`, `null`, ... would also be an invalid argument for our purposes.

Maybe we need to validate both:
- that the `node` argument was provided
- that the provided value is a `PostCSS` node

It is not inconceivable that a rule author might try to pass a `postcss-selector-parser` node to `utils.report()`.

lib/utils/report.mjs
ybiquitous: Thank you. For consistency, could you change `report`s in the `emitDeprecationWarnings` third arguments, too?

lib/utils/report.mjs
ybiquitous: Oh, I completely misunderstood! 🙇🏼 

The current check is no problem 👍🏼

lib/utils/validateTypes.mjs
```diff
@@ -118,3 +120,45 @@ export function assertString(value) {
 	// eslint-disable-next-line no-console
 	console.assert(isString(value), `"${value}" must be a string`);
 }
+
+/**
+ * @param {unknown} value
+ * @returns {value is (Range['start'] | Range['end'])}
```
ybiquitous, line 126: [suggestion] It seems possible to import and use the `Position` type here:

```suggestion
 * @returns {value is Position}
```

lib/utils/validateTypes.mjs
```diff
@@ -118,3 +120,45 @@ export function assertString(value) {
...
+	if (!isNumber(value.column)) return false;
+
+	return true;
+}
+
+/**
+ * @param {unknown} value
+ * @returns {value is Range}
+ */
+export function isRange(value) {
```
ybiquitous, line 142: [note] If we wanted to make `isPosition` and `isRange` more strict, we might also need to check that `value` should contain only specific properties (e.g., `line`/`column` or `start`/`end`).

But that may be too strict for now. I think we can leave it as-is.

lib/utils/validateTypes.mjs
```diff
@@ -118,3 +120,45 @@ export function assertString(value) {
...
+	if (!isPosition(value.end)) return false;
+
+	return true;
+}
+
+/**
+ * @param {unknown} value
+ * @returns {value is { index: number, endIndex: number }}
+ */
+export function hasIndices(value) {
```
ybiquitous, line 156: [suggestion] `hasIndices` seems too specific to be added to this `validateTypes` since the name starts with `has`, not `is`.

Instead, how about moving `hasIndices` to `report.mjs` as a private function?

lib/utils/validateTypes.mjs
romainmenke: That might also make things harder for rule authors.
The `postcss` `Position` type for example also has an `offset` field.

Making the type too strict would require rule authors to destructure these to ensure that extra fields are erased.

Agree to leave it as-is for now :)

lib/utils/report.mjs
romainmenke: I've added `quietDeprecationWarnings` like `quiet` and added test coverage.

ybiquitous (approved):
Thank you, LGTM 👍🏼

We probably should release this PR along with #8248 fix.

lib/utils/report.mjs
```diff
@@ -115,6 +125,45 @@ export default function report(problem) {
...
+ */
+function problemRangeDeprecations(problem) {
+	if (problem.result.stylelint.quietDeprecationWarnings) return;
+
+	if (!problem.node) {
+		emitDeprecationWarning(
+			`Omitting the \`node\` argument in the \`utils.report()\` function is deprecated ("${problem.ruleName}").`,
+			'REPORT_AMBIGUOUS_POSITION',
+			`Please pass a \`node\` argument in the \`utils.report()\` function of "${problem.ruleName}".`,
+		);
```
Mouvedia, line 139: suggestion
minor

We already had `report` throwing when `node` wasn't passed.
I don't think this `emitDeprecationWarning` is necessary.
It should remain a simple `throw`.

romainmenke: The current code throws when neither `node` or `line` are provided.
According to that logic it would be acceptable to pass only `node` or only `line`.
Rule authors might today omit `node` and provide `line`.

We need to communicate that `node` is now required and that `line` is deprecated.

This shape of the code also aligns with the intended form for the next major: https://github.com/stylelint/stylelint/blob/b0525066e26cd5a1f45a2a746be945c8150046df/lib/utils/report.cjs#L99-L103

In the next major we do:
- remove the existing `throw`
- `throw` instead of warn about deprecations

So I intentionally added this throw when `node` is missing as this is both the most informative and the most direct path forwards.

lib/utils/validateTypes.mjs
Mouvedia: Objects are extensible by default in TS.
It's actually really hard to restrict them compared to flow.
i.e. [`{| /* … */ |}`](https://flow.org/en/docs/types/objects/#exact-and-inexact-object-types)
Passing an object with supplementary properties shouldn't be a problem.

ref: microsoft/TypeScript#12936

lib/utils/__tests__/report.test.mjs
```diff
@@ -550,3 +563,304 @@ test('without node nor line', () => {
...
+	report(v);
+
+	expect(emitWarning).toHaveBeenCalledWith(
+		expect.stringMatching(
+			/Providing the `line` argument in the `utils.report\(\)` function is deprecated \("foo"\)\./,
+		),
+		expect.objectContaining({
+			code: 'stylelint:007',
+			detail:
+				'Please pass both `index` and `endIndex` as arguments in the `utils.report()` function of "foo" instead.',
```
Mouvedia, line 637: question

Maybe _to the_ instead of _in the_?
e.g. pass x as y to the z

lib/utils/__tests__/report.test.mjs
```diff
@@ -550,3 +563,304 @@ test('without node nor line', () => {
...
+			expect.objectContaining({
+				code: 'stylelint:007',
+				detail:
+					'Please pass both a valid `start` and `end` argument in the `utils.report()` function of "foo".',
+				type: 'DeprecationWarning',
+			}),
+		);
+	});
+
+	it('has an invalid end', () => {
```
Mouvedia, line 832: question

It seems that you have covered many cases, is it our responsibility if `start` comes after `end`?
e.g.
```
start: {
	line: 5,
	column: 3,
},
end: {
	line: 1,
	column: 2,
},
```

lib/utils/report.mjs
```diff
@@ -115,6 +125,45 @@ export default function report(problem) {
 	result.warn(warningMessage, warningProperties);
 }
 
+/**
+ * @param {Problem} problem
+ */
+function problemRangeDeprecations(problem) {
```
Mouvedia, line 131: question

What about a name which would convey an action?
e.g. `checkPositions` or `enforceConsistency`

lib/utils/emitDeprecationWarning.mjs
```diff
@@ -8,6 +8,7 @@ const CODES = {
 	GITHUB_FORMATTER: '004',
 	CONTEXT_FIX: '005',
 	RULE: '006', // either removal or renaming
+	REPORT_AMBIGUOUS_POSITION: '007',
```
Mouvedia, line 11: suggestion
nit

`AMBIGUOUS_POSITION` should be sufficient
I know you wanted to scope it to `report`.
You could leave it in if you think we may have other cases of _ambiguous positions_ elsewhere in the future.

types/stylelint/index.d.ts
Mouvedia, line 1: question
minor

Can you add `/** @deprecated */` before l851?

lib/utils/__tests__/report.test.mjs
romainmenke: I think that is out of scope :)

We could investigate how PostCSS handles such arguments, both with `start/end` and `index/endIndex`. Can you open an issue about this?

lib/utils/__tests__/report.test.mjs
romainmenke: I think `in` is also fine.
Let's keep it as-is, as this is only temporary text.

lib/utils/report.mjs
romainmenke: I've changed it to `checkProblemRangeDeprecations`.
This function is only temporary.

It will change to something like `problemRange` or `getProblemRange`.
I am trying to maintain a parallel with the future name.

lib/utils/emitDeprecationWarning.mjs
romainmenke: Let's keep this as-is.

types/stylelint/index.d.ts
romainmenke: Good suggestion! Thank you 🙇

lib/utils/__tests__/report.test.mjs
Mouvedia: On `postcss` repository?

romainmenke: I would first open an issue here to investigate. Then we don't block this PR as this question is out of scope. If we do find that improvements can be made in such cases then we can see where and how to submit patches.

On the other hand I think it is a fair assumption that `start` always comes before `end`. If rule authors get it wrong then they will see this in their test outcomes. We already have systems in place to validate the exact ranges.

So honestly I think this is a non-issue :)

ybiquitous (approved):

jeddy3 (approved):

romainmenke:
Thank you for the feedback everyone 🙇

Issue #8243, opened 2024-12-29
Refactor core rules to have unambiguous position arguments


romainmenke:
see: https://github.com/stylelint/stylelint/issues/8126

Given the planned deprecations we should refactor core rules so that these do not emit deprecation warnings.

Affected rules:
- [x] https://github.com/stylelint/stylelint/pull/8245
- [x] https://github.com/stylelint/stylelint/pull/8246
- [x] https://github.com/stylelint/stylelint/pull/8247

PR #8245, opened 2024-12-29
Refactor `block-no-empty` to have unambiguous position arguments

romainmenke:

See https://github.com/stylelint/stylelint/issues/8243

No, it's self-explanatory.

### DIFF
--- a/lib/rules/block-no-empty/index.mjs
+++ b/lib/rules/block-no-empty/index.mjs
@@ -75,6 +75,7 @@ const rule = (primary, secondaryOptions, context) => {
 				message: messages.rejected,
 				node: statement,
 				start: statement.positionBy({ index }),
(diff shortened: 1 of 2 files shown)


lib/rules/block-no-empty/index.mjs
```diff
@@ -75,6 +75,7 @@ const rule = (primary, secondaryOptions, context) => {
 				message: messages.rejected,
 				node: statement,
 				start: statement.positionBy({ index }),
+				end: statement.rangeBy({}).end,
```
romainmenke, line 78: This isn't particularly pretty or elegant but it is an exact match of what happens implicitly when only providing `start` and omitting `end`.

jeddy3 (approved):

Issue #8133, opened 2024-11-19
Add support for checking more unknown things


jeddy3:
### What is the problem you're trying to solve?

We currently have:

- gaps in our checks for unknown things
- overlaps between rules that lead to problems being flagged twice
- some incorrectly scoped rules

It'd be great to address these and make checking of unknown values widely available to our users through our standard config.

### What solution would you like to see?

I originally suggested a new mega-rule for unknown at-rules, properties and descriptors in https://github.com/stylelint/stylelint/pull/8124#issuecomment-2484169262. In hindsight, that was rash as:

- we'll want to [limit the disruption](https://github.com/stylelint/stylelint/issues/6593#issuecomment-2470778938) to our users (including those using language extensions like SCSS)
- we have [a diverse set](https://github.com/stylelint/stylelint/pull/8124#issuecomment-2484227591) of unknown rules powered by different tools

We can fill the gaps, workaround overlaps, and correct the scope of rules while minimising disruption by:

- adding new rules with limited scopes
- adding new options to existing rules
- fixing the default behaviour of some rules to be more permissive, i.e. non-breaking

This approach aligns with our design decision to have standalone rules, which has helped us manage complexity. However, it means users must continue to customise syntax in multiple places. I suspect customizing the syntax is an uncommon use case, though. 

Top-level overview:

- add `at-rule-prelude-no-unknown` to catch `foo` in `@property foo {}`
- add `at-rule-descriptor-no-unknown` to catch `margin` in `@counter-style bar { margin: 0; }`
- add `at-rule-descriptor-value-no-unknown` to catch `red` in `@font-face { size-adjust: red; }`
- add `ignore: ["values"]` to `color-no-invalid-hex`, `function-no-unknown`, `string-no-newline` and `unit-no-unknown`
- fix `property-no-unknown` and `declaration-property-value-no-unknown` to only check properties and not descriptors
- refactor `property-no-unknown` and `at-rule-no-unknown` to use csstree data

In our shared config, we can then turn on our `*-value-no-unknown` rules and use the new secondary options to avoid overlaps:

```diff js
module.exports = {
	rules: {
        /* … */
+       'declaration-property-value-no-unknown': true,
-       'color-no-invalid-hex': true,
+	'color-no-invalid-hex': [true, { ignore: ["values"] }],
-	'function-linear-gradient-no-nonstandard-direction': true,
-	'function-no-unknown': true,
+	'function-no-unknown': [true, { ignore: ["values"] }],
[...]

PR #8251, opened 2024-12-29
Add `at-rule-no-deprecated`

### LINKED ISSUE #8224: Add `at-rule-no-deprecated`

I want to be warned about at-rules that were once in a spec (including editor ones) but are now removed or deprecated.

A new rule to flag `@viewport`, `@apply`, `@nest` and `@document`:

https://github.com/stylelint/stylelint/blob/6cccbdcc0315c4a5ebfb9033c20d838dc2c2260b/lib/reference/atKeywords.mjs#L49-L69

- Name: `at-rule-no-deprecated`
- Description: Disallow deprecated at-rules.
- Message: `Unexpected deprecated at-rule "${at-rule}"`
- Primary option: `true`
- Secondary options: `ignoreAtRules: []`
- Autofix: No
- Category: Avoid errors > Deprecated
- Included in sharable configs?: Yes, recommended when pieces are in place
- Expanded description: This rule flags at-rules that were once in the CSS specifications, including Editor Drafts, but were removed or deprecated.

jeddy3:

Closes #8224 

I figured I'd quickly create this one, as it feels like part of our efforts to improve the coverage of at-rules in https://github.com/stylelint/stylelint/issues/8133. We may get all the parts of that issue out in one release!

It's been a while since I created a rule from scratch, and I'm thankful for the improved DX since my last one. I was prompted to do the [three wiring-up steps](https://stylelint.io/developer-guide/rules/#wire-up-the-rule) through type checks and failing tests rather than having to remember myself. Thank you if you had a hand in this improvment.


docs/user-guide/rules.md
```diff
@@ -11,6 +11,16 @@ We turn on most of the rules in our [standard config](https://www.npmjs.com/pack
 
 You can avoid errors with these `no` rules.
 
+### Deprecated
+
+Disallow deprecated things with these `no-deprecated` rules.
```
Mouvedia, line 16: note

Already part of #8223.

lib/rules/at-rule-no-deprecated/README.md
```diff
@@ -0,0 +1,70 @@
+# at-rule-no-deprecated
+
+Disallow deprecated at-rules.
+
+<!-- prettier-ignore -->
+```css
+    @document url(http://www.w3.org/);
```
Mouvedia, line 7: You should probably pick another one as an example.
i.e. AFAIK non-vendor-prefixed `@document` was never supported by any browser

ref: https://developer.mozilla.org/en-US/docs/Web/CSS/@document#browser_compatibility

lib/rules/at-rule-no-deprecated/index.mjs
```diff
@@ -0,0 +1,64 @@
...
+
+			const atName = `@${name}`;
+
+			report({
+				message: messages.rejected,
+				messageArgs: [atName],
+				node: atRule,
+				ruleName,
+				result,
+				word: atName,
```
Mouvedia, line 54: IIRC this is not recommended anymore.
@romainmenke should we continue using `word` internally?

romainmenke:
```suggestion
				index: 0,
				endIndex: atName.length,
```

This would be better.

------

The `at-rule` `name` is maybe the one instance where `word` is acceptable :)
It is always at `index` `0`, so it is always first and always singular.
Using `word` can only find the first instance of multiple problems.

But calculating position information from `word` is slower as it requires serializing the entire current node.

This might be a large amount of AST nodes as `at-rules` can span many style rules each containing many declarations.

jeddy3: Thanks for the explanation. It all makes sense.

lib/rules/at-rule-no-deprecated/index.mjs
```diff
@@ -0,0 +1,65 @@
...
+
+			report({
+				message: messages.rejected,
+				messageArgs: [atName],
+				node: atRule,
+				ruleName,
+				result,
+				index: 0,
+				endIndex: atName.length,
+			});
```
romainmenke, line 56:
```suggestion
				endIndex: atName.length,
				fix: name.toLowerCase() === 'nest' ? () => {
					// implies `import postcss from 'postcss';` at the top
					const styleRule = postcss.rule({
						selector: atRule.params,
						source: atRule.source,
					});

					if (atRule.nodes) {
						atRule.nodes.forEach((node) => {
							styleRule.append(node);
						});
					}

					atRule.replaceWith(styleRule);
				} : undefined
			});
```

[suggestion] `@nest` can be auto fixed by changing the node to a regular rule.

See: https://github.com/csstools/postcss-plugins/blob/main/plugins-stylelint/no-at-nest-rule/index.mjs

_The suggested code is not in the ideal form or entirely following our conventions. It is only intended as an illustration._

romainmenke (commented):
Nice addition!
I've left a suggestion to add a fix for `@nest {}` but we could also ship an initial version without a fixer.

lib/rules/at-rule-no-deprecated/index.mjs
Mouvedia: > I figured I'd quickly create this one…

If we want to expedite this PR should we separate the fixer from the addition of the rule itself?

jeddy3:
I've added the fixer, but have tripped on unhelpful test results:

```
TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property 'parent' -> object with constructor 'Object'
    |     property 'nodes' -> object with constructor 'Array'
    --- index 0 closes the circle
    at stringify (<anonymous>)
    at writeChannelMessage (node:internal/child_process/serialization:159:20)
    at process.target._send (node:internal/child_process:851:17)
    at process.target.send (node:internal/child_process:751:19)
    at reportSuccess (/Users/jeddy3/Projects/stylelint/node_modules/jest-worker/build/workers/processChild.js:82:11)

Node.js v22.11.0
 FAIL  lib/rules/at-rule-no-deprecated/__tests__/index.mjs
  ● Test suite failed to run

    Jest worker encountered 4 child process exceptions, exceeding retry limit

      at ChildProcessWorker.initialize (node_modules/jest-worker/build/workers/ChildProcessWorker.js:181:21)

```

Has anyone seen this before? I can investigate, but I figured I'd ask before I do.

I also documented the prior art plugins.

lib/rules/at-rule-no-deprecated/README.md
```diff
@@ -19,6 +19,11 @@ This rule flags at-rules that were removed or deprecated after being in the CSS
 
 The [`message` secondary option](../../../docs/user-guide/configure.md#message) accept arguments.
```
Mouvedia, line 20:
```suggestion
The [`fix` option](../../../docs/user-guide/options.md#fix) can automatically fix some of the problems reported by this rule.

The [`message` secondary option](../../../docs/user-guide/configure.md#message) accept arguments.
```

docs/user-guide/rules.md
```diff
@@ -11,6 +11,16 @@ We turn on most of the rules in our [standard config](https://www.npmjs.com/pack
...
+
+<!-- prettier-ignore-start -->
+| | | |
+| :-- | :-: | :-: |
+| [`at-rule-no-deprecated`](../../lib/rules/at-rule-no-deprecated/README.md)<br/>Disallow deprecated at-rules. | | |
```
Mouvedia, line 21:
```suggestion
| [`at-rule-no-deprecated`](../../lib/rules/at-rule-no-deprecated/README.md)<br/>Disallow deprecated at-rules. | | 🔧 |
```

Mouvedia:
> I've added the fixer, but have tripped on unhelpful test results […]

I don't know if that will help but you could try moving the `atNestFixer` function out of `root.walkAtRules((atRule) => { … }`.
i.e. `(atRule: AtRule) => void`

Mouvedia:
@jeddy3 just guessing there but

```
				styleRule.append(atRule.nodes);

				atRule.replaceWith(styleRule);
```

I suppose you want to clone `atRule` before append and use the clone instead.

romainmenke:
@jeddy3 I've filed https://github.com/stylelint/jest-preset-stylelint/issues/130 to improve the error feedback in such cases

jeddy3:
I've added autofix for `nest`. Ready for another review.

romainmenke (approved):
Thank you. LGTM 👍

lib/rules/at-rule-no-deprecated/index.mjs
```diff
@@ -0,0 +1,82 @@
...
+					selector: atRule.params,
+					source: atRule.source,
+				});
+
+				styleRule.append(atRule.nodes);
+
+				atRule.replaceWith(styleRule);
+			};
+
+			const fix = name === 'nest' ? atNestFixer : undefined;
```
ybiquitous, line 62: [question] Is the normalization of the name necessary here? E.g.

```js
const fix = name.toLowerCase() === 'nest' ? atNestFixer : undefined;
```

If needed, avoiding repetition of the `toLowerCase()` calls might be better. E.g.

```js
const normalizedName = name.toLowerCase();

if (!deprecatedAtKeywords.has(normalizedName)) return;

const fix = normalizedName === 'nest' ? atNestFixer : undefined;
```

jeddy3: Good catch. It was necessary. Fixed in https://github.com/stylelint/stylelint/pull/8251/commits/54b11d3fa9ef7e9537b4267fb99cecf07a563788.

ybiquitous (approved):
Thank you for the quick action. LGTM 👍🏼

PR #8252, opened 2024-12-30
Add `messageArgs` to `selector-pseudo-element-colon-notation` and `media-feature-name-unit-allowed-list`

Mouvedia:

#4117

N/A

### DIFF
--- a/.changeset/short-lobsters-end.md
+++ b/.changeset/short-lobsters-end.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: `messageArgs` to `selector-pseudo-element-colon-notation` and `media-feature-name-unit-allowed-list`
--- a/lib/rules/media-feature-name-unit-allowed-list/README.md
+++ b/lib/rules/media-feature-name-unit-allowed-list/README.md
@@ -9,6 +9,8 @@ Specify a list of allowed name and unit pairs within media features.
  * This media feature name and these units */
 ```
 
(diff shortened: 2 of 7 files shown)


types/stylelint/index.d.ts
```diff
@@ -279,20 +279,48 @@ declare namespace stylelint {
...
+
+	type ExpectedMessage<T> = {
+		expected: (
+			...args: T extends keyof MessageTypes ? MessageTypes[T] : T extends [string] ? T : never
+		) => string;
+	};
+
+	type PropertyMessage<T = never> = {
+		rejected: (
+			...args: [T] extends [never]
```
Mouvedia, line 296: To avoid the distributive nature of TS conditional types we need a non-naked type parameter.
i.e. `T extends never` wouldn't work

ybiquitous (changes requested):
Thanks for the pull request. For the support for `messageArgs`, LGTM. (as the PR title says)

However, I'm against mixing the `messageArgs` changes and polishing the types. If you want to discuss the types, you should separate a PR.

For the types, I'm worried that the types would be too complex to maintain. Do we really need such types?

Mouvedia:
> For the types, I'm worried that the types would be too complex to maintain. Do we really need such types?

Right now it's not completely warranted once I type the other half—which is around 50 more rules—it will be evident.

> …you should separate a PR.

I can but that part won't have a changeset anyway, it's just an enhancement.

ybiquitous (approved):
Thank you for simplifying this PR. LGTM 👍🏼 

Let's discuss the types on a follow-up PR.

PR #7983, opened 2024-09-03
Fix code block in Markdown documents

ybiquitous:
> Which issue, if any, is this issue related to?

None, as it's a documentation fix.

> Is there anything in the PR that needs further explanation?

Summary:
- Fix syntax errors (thanks to [`remark-lint-code-block-syntax`](https://www.npmjs.com/package/remark-lint-code-block-syntax))
- Make code examples more consistent in `user-guide/configure.md`

To verify syntax, temporarily update `package.json` and run `npm i && npm run lint:md`:

```diff
--- a/package.json
+++ b/package.json
@@ -118,7 +118,8 @@
   },
   "remarkConfig": {
     "plugins": [
-      "@stylelint/remark-preset"
+      "@stylelint/remark-preset",
+      "remark-lint-code-block-syntax"
     ]
   },
   "jest": {
@@ -239,6 +240,7 @@
     "postcss-sass": "^0.5.0",
     "postcss-scss": "^4.0.9",
     "remark-cli": "^12.0.1",
+    "remark-lint-code-block-syntax": "^0.10.0",
     "rollup": "^4.21.2",
     "sugarss": "^4.0.1",
     "typescript": "^5.5.4"
```

Or run only the commands:

```shell
npm i -D remark-lint-code-block-syntax
npm pkg set 'remarkConfig.plugins[1]=remark-lint-code-block-syntax'
npm run lint:md
```

However, we have to accept the following errors for now:

```
lib/rules/no-empty-source/README.md
6:1-10:4  warning Invalid CSS: <css input>:1:3: Unknown word code-block-syntax remark-lint
26:1-28:4 warning Invalid CSS: <css input>:1:1: Unknown word code-block-syntax remark-lint
31:1-33:4 warning Invalid CSS: <css input>:1:1: Unknown word code-block-syntax remark-lint
```

PR #8257, opened 2024-12-30
Document best practices for position arguments in `utils.report()`

jeddy3:

Closes https://github.com/stylelint/stylelint/issues/8261

If someone has time, it may be worth adding a small section to the "writing rules" about the `report()` function to give examples of how `node`, `index` and `endIndex` are likely used.

### DIFF
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -54,7 +54,8 @@ const ruleFunction = (primary, secondaryOptions, context) => {
         ruleName,
         message: messages.rejected(selector),
         node: ruleNode,
-        word: selector
+        index: 0,
+        endIndex: selector.length
       });
     });
   };
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -54,8 +54,7 @@ const ruleFunction = (primary, secondaryOptions, context) => {
         ruleName,
         message: messages.rejected(selector),
         node: ruleNode,
-        index: 0,
-        endIndex: selector.length
+        word: selector
       });
     });
   };
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -152,9 +152,9 @@ root.walkRules((ruleNode) => {
 
 Using `word` is the slowest approach to finding a location within a node, especially on nodes with many children, e.g. at-rules.
 
-When using offsets and positions, you can use the `nodeFieldIndices` utilities, e.g. `declarationValueIndex`, to get the index of part of a node. These utilities account for the `raw` fields in the PostCSS AST.
+When using offsets and positions, you can use the `nodeFieldIndices` utilities, e.g. `declarationValueIndex()`, to get the index of part of a node. These utilities account for the `raw` fields in the PostCSS AST.
 
-In the following example of using `index`, `endIndex` and the `declarationValueIndex` utility, the location spans the value of the declaration:
+In the following example of using `index`, `endIndex` and the `declarationValueIndex()` utility, the location spans the value of the declaration:
 
 ```js
 root.walkDecls((declNode) => {
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -178,7 +178,7 @@ This approach also works well when using construct-specific parsers:
 
 ```js
 root.walkDecls((declNode) => {
-  const { prop, value as declValue } = declNode;
+  const { prop, value: declValue } = declNode;
 
   valueParser(declValue).walk(({ value, sourceIndex }) => {
 
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -212,6 +212,8 @@ function rule(primary, secondary) {
       ruleName,
       message,
       node,
+      index,
+      endIndex,
 +     fix
     });
   };
--- a/docs/developer-guide/rules.md
+++ b/docs/developer-guide/rules.md
@@ -123,7 +123,7 @@ When using `report()`, you can specify the location of a problem in various ways
 - only `node` - implicitly span the entire range of the given node
 - `word` - the first instance of a word in the serialized given node
 - `index` and `endIndex` offsets - an index range within the given node
-- `start` and `end` positions - a position range within the given node
+- `start` and `end` positions - a [range](https://postcss.org/api/#range) within the given node
 
 Each approach has pros and cons concerning:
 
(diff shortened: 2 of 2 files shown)


docs/developer-guide/plugins.md
```diff
@@ -54,7 +54,8 @@ const ruleFunction = (primary, secondaryOptions, context) => {
         ruleName,
         message: messages.rejected(selector),
         node: ruleNode,
-        word: selector
+        index: 0,
+        endIndex: selector.length
```
Mouvedia, line 58: We want to not recommend `word` but the previous code was actually better.
see https://github.com/postcss/postcss/blob/99da2f28aeded88c5e0fa3b4dd6fb44230710f57/lib/node.js#L261-L273

We would need to be certain that it's always `0` for us to make that change.
I think to leave it in for now, only `line` is being deprecated.
see https://github.com/stylelint/stylelint/issues/8126#issuecomment-2482487257

romainmenke:
> If someone has time, it may be worth adding a small section to the "writing rules" about the report() function to give examples of how node, index and endIndex are likely used.

https://github.com/stylelint/stylelint/issues/8261

I've opened an issue for this and added some initial thoughts

romainmenke (commented):
I am a bit unsure about this change.

On the one hand your change is obviously correct and it is a direction I think we want to promote for core rules. But `word` is also very convenient for new rule authors.

Should we keep the documentation as-is for now and await the outcome of https://github.com/stylelint/stylelint/issues/8261 ?

https://github.com/stylelint/stylelint/pull/8251 this change might have just been an unlucky combo of linting at-rule names, where all the other and similar rules still use `word`, despite being core rules :)

But I don't have a very strong opinion on this and happy with this change if you feel it is generally helpful to authors to already promote `index`/`endIndex`.

Mouvedia:
> I am a bit unsure about this change.

I agree with @romainmenke, `word` is sometimes OKish/convenient.
i.e. this PR is probably not necessary
If anything, it's `line` that will need a `> [!WARNING]` for its deprecation.

jeddy3:
> Should we keep the documentation as-is for now and await the outcome of https://github.com/stylelint/stylelint/issues/8261 ?

SGTM. Let's rework this PR for that issue.

jeddy3:
I've reworked the docs based on @romainmenke excellent write-up in https://github.com/stylelint/stylelint/issues/8261#issue-2763583610.

I tried to give rule authors a sense of how the location arguments work without getting into overwhelming detail so that they have some grounding when they look at the code of existing rules.

I've given a couple of examples of what seem to be our most common patterns:

- using `word`
- using `index` and `endIndex` (often with a construct parser)

docs/developer-guide/rules.md
```diff
@@ -113,7 +113,89 @@ Stylelint has [utility functions](https://github.com/stylelint/stylelint/tree/ma
...
+- `report()` utility to report lint problems
+
+##### Location arguments for `report()`
+
+When using `report()`, you can specify the location of a problem in various ways:
+
+- only `node` - implicitly span the entire range of the given node
+- `word` - the first instance of a word in the serialized given node
+- `index` and `endIndex` offsets - an index range within the given node
+- `start` and `end` positions - a position range within the given node
```
jeddy3, line 126: I wasn't sure how to describe these. Any suggestions?

Mouvedia: If you are talking about `start`/`end`, I would link to the PostCSS API directly.

i.e.
```suggestion
- `index` and `endIndex` offsets - an index range within the given node
- `start` and `end` positions - a [range](https://postcss.org/api/#range) within the given node
```

docs/developer-guide/rules.md
```diff
@@ -113,7 +113,89 @@ Stylelint has [utility functions](https://github.com/stylelint/stylelint/tree/ma
...
+    message: messages.rejected(),
+    node: ruleNode,
+    word: selector
+  });
+});
+```
+
+Using `word` is the slowest approach to finding a location within a node, especially on nodes with many children, e.g. at-rules.
+
+When using offsets and positions, you can use the `nodeFieldIndices` utilities, e.g. `declarationValueIndex`, to get the index of part of a node. These utilities account for the `raw` fields in the PostCSS AST.
```
Mouvedia: If we are recommending our internal utilities we should probably add them to the `Utils` type and to the `stylelint.utils` chapter of `plugins.md`.

We didn't bikeshed about the name of the file in #8125 because we thought it would remain private.

jeddy3: Let's move that discussion to another issue. This doc is primarily for rule authors.

docs/developer-guide/rules.md
```diff
@@ -113,7 +113,89 @@ Stylelint has [utility functions](https://github.com/stylelint/stylelint/tree/ma
...
+
+In the following example of using `index`, `endIndex` and the `declarationValueIndex` utility, the location spans the value of the declaration:
```
ybiquitous, line 157: [suggestion] This is minor, but how about making the consistent form for the function like `report()`?

```suggestion
When using offsets and positions, you can use the `nodeFieldIndices` utilities, e.g. `declarationValueIndex()`, to get the index of part of a node. These utilities account for the `raw` fields in the PostCSS AST.

In the following example of using `index`, `endIndex` and the `declarationValueIndex()` utility, the location spans the value of the declaration:
```

docs/developer-guide/rules.md
```diff
@@ -113,7 +113,89 @@ Stylelint has [utility functions](https://github.com/stylelint/stylelint/tree/ma
...
+    endIndex:
+  });
+});
+```
+
+This approach also works well when using construct-specific parsers:
+
+```js
+root.walkDecls((declNode) => {
+  const { prop, value as declValue } = declNode;
```
ybiquitous, line 181: [suggestion] We should fix the invalid syntax for `as`:

```suggestion
  const { prop, value: declValue } = declNode;
```

ybiquitous (commented):
Thank you, almost LGTM. Can you check my minor comments?

docs/developer-guide/rules.md
jeddy3: Oops, we totally should! Good catch.

jeddy3:
@ybiquitous Commit your suggestions. Thank you for your attention to detail.

docs/developer-guide/rules.md
Mouvedia: Should we use `remark-lint-code-block-syntax`?
related: #7983

jeddy3: Let's open an issue to discuss.

romainmenke (approved):
Thank you for working on this @jeddy3. LGTM 👍

PR #8258, opened 2024-12-30
Document rules becoming stable

jeddy3:

Ref: https://github.com/stylelint/stylelint/issues/8133

The rules have been used for a while, and we're readying to use them in our config.


romainmenke (approved):
Thank you. LGTM 👍

PR #8262, opened 2024-12-30
Refactor `FileCache` to support `file-entry-cache` version 10

romainmenke:

See https://github.com/stylelint/stylelint/pull/8256

The latest version of `file-entry-cache` considers any file with either a changed check sum or changed meta data as changed. This is different from the previous version.

As I understood the current code and tests we want to skip linting files that still have the same content, even when a timestamp or other metadata has changed.

In support of that I added a `hasFileCheckSumChanged` method to `FileCache` and updated `hasFileChanged` to restore the old behavior.

### DIFF
--- a/lib/__tests__/standalone-cache.test.mjs
+++ b/lib/__tests__/standalone-cache.test.mjs
@@ -405,10 +405,7 @@ describe('standalone cache uses cacheStrategy', () => {
 
 		// No content change, but file metadata id changed
 		// Note: needs a time offset for fast test machines
-		const time = new Date();
-
-		time.setSeconds(time.getSeconds() + 10);
-		await utimes(newFileDest, time, time);
+		await utimes(newFileDest, Date.now(), Date.now() + 1000);
 
 		const { results: resultsCached } = await standalone(getConfig({ cacheStrategy }));
 
--- a/lib/__tests__/standalone-cache.test.mjs
+++ b/lib/__tests__/standalone-cache.test.mjs
@@ -405,7 +405,7 @@ describe('standalone cache uses cacheStrategy', () => {
 
 		// No content change, but file metadata id changed
 		// Note: needs a time offset for fast test machines
-		await utimes(newFileDest, Date.now() / 1000, Date.now() / 1000 + 1000);
+		await utimes(newFileDest, Date.now() / 1000, Date.now() / 1000 + 1);
 
 		const { results: resultsCached } = await standalone(getConfig({ cacheStrategy }));
 
--- a/lib/__tests__/standalone-cache.test.mjs
+++ b/lib/__tests__/standalone-cache.test.mjs
@@ -404,7 +404,12 @@ describe('standalone cache uses cacheStrategy', () => {
 		expect(results.some((file) => isChanged(file, newFileDest))).toBe(true);
 
 		// No content change, but file metadata id changed
-		await utimes(newFileDest, new Date(), new Date());
+		// Note: needs a time offset for fast test machines
+		const time = new Date();
+
+		time.setSeconds(time.getSeconds() + 10);
+		await utimes(newFileDest, time, time);
+
 		const { results: resultsCached } = await standalone(getConfig({ cacheStrategy }));
 
 		expect(resultsCached.some((file) => isChanged(file, newFileDest))).toBe(true);
--- a/lib/__tests__/standalone-cache.test.mjs
+++ b/lib/__tests__/standalone-cache.test.mjs
@@ -405,7 +405,7 @@ describe('standalone cache uses cacheStrategy', () => {
 
 		// No content change, but file metadata id changed
 		// Note: needs a time offset for fast test machines
-		await utimes(newFileDest, Date.now(), Date.now() + 1000);
+		await utimes(newFileDest, Date.now() / 1000, Date.now() / 1000 + 1000);
 
 		const { results: resultsCached } = await standalone(getConfig({ cacheStrategy }));
 
--- a/lib/utils/FileCache.mjs
+++ b/lib/utils/FileCache.mjs
@@ -59,6 +59,10 @@ export default class FileCache {
 	hasFileChanged(absoluteFilepath) {
 		// Get file descriptor compares current metadata against cached
 		// one and stores the result to "changed" prop.w
+
+		const fileCheckSumChanged = this._useCheckSum
+			? this.hasFileCheckSumChanged(absoluteFilepath)
+			: false;
 		const descriptor = this._fileCache.getFileDescriptor(absoluteFilepath);
 
 		descriptor.meta.data ??= {};
@@ -70,7 +74,7 @@ export default class FileCache {
 		let changed = false;
 
 		if (this._useCheckSum) {
-			changed = configChanged || this.hasFileCheckSumChanged(absoluteFilepath);
+			changed = configChanged || fileCheckSumChanged;
 		} else {
 			changed = configChanged || Boolean(descriptor.changed);
 		}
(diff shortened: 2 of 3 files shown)


lib/__tests__/standalone-cache.test.mjs
```diff
@@ -404,7 +404,12 @@ describe('standalone cache uses cacheStrategy', () => {
 		expect(results.some((file) => isChanged(file, newFileDest))).toBe(true);
 
 		// No content change, but file metadata id changed
-		await utimes(newFileDest, new Date(), new Date());
+		// Note: needs a time offset for fast test machines
+		const time = new Date();
+
+		time.setSeconds(time.getSeconds() + 10);
+		await utimes(newFileDest, time, time);
```
romainmenke, line 411: This test was failing on my machine, also on main.

I am assuming that the precision of timestamps on mac is not very high and that modern macs (M1 or later) are just very fast.

This combo means there just isn't enough time between creating the file, running Stylelint once and calling `utimes()`.

Artificially increasing the time delta makes this test pass consistently.

lib/utils/FileCache.mjs
```diff
@@ -60,23 +60,54 @@ export default class FileCache {
...
+		}
+
+		const buffer = readFileSync(absoluteFilepath);
+
+		if (this._fileCache.getHash(buffer) !== metaCache.hash) {
+			return true;
+		}
+
+		return false;
+	}
```
romainmenke, line 109: This is based on : https://github.com/jaredwray/cacheable/blob/5ec1858bdd07aa5efc30e87f14253a951fbc83b5/packages/file-entry-cache/src/index.ts#L243-L310

lib/utils/FileCache.mjs
```diff
@@ -60,23 +60,54 @@ export default class FileCache {
 		// Get file descriptor compares current metadata against cached
 		// one and stores the result to "changed" prop.w
 		const descriptor = this._fileCache.getFileDescriptor(absoluteFilepath);
-		/** @type {CacheMetadata} */
-		const meta = descriptor.meta || {};
-		const changed = descriptor.changed || meta.hashOfConfig !== this._hashOfConfig;
+
+		descriptor.meta.data ??= {};
```
romainmenke, line 64: See: https://github.com/jaredwray/cacheable/tree/5ec1858bdd07aa5efc30e87f14253a951fbc83b5/packages/file-entry-cache#changes-from-v9-to-v10

> On `FileEntryDescriptor.meta` if using typescript you need to use the `meta.data` to set additional information. This is to allow for better type checking and to avoid conflicts with the meta object which was any.

lib/__tests__/standalone-cache.test.mjs
ybiquitous: That makes sense. 👍🏼 

In addition, we could simplify the code (`Date.now()` returns milliseconds) since [`fs.utimes()`](https://nodejs.org/api/fs.html#fsutimespath-atime-mtime-callback) can also receive numbers. Like so:

```js
await utimes(newFileDest, Date.now(), Date.now() + 1000);
```

Also, it seems possible to change also in another case: 👇🏼 

https://github.com/stylelint/stylelint/blob/0500ce6e40a8f9a978e3fe0ae360626bdb13d930/lib/__tests__/standalone-cache.test.mjs#L421-L422

lib/utils/FileCache.mjs
```diff
@@ -59,24 +59,59 @@ export default class FileCache {
...
+			? this.hasFileCheckSumChanged(absoluteFilepath)
+			: false;
 		const descriptor = this._fileCache.getFileDescriptor(absoluteFilepath);
-		/** @type {CacheMetadata} */
-		const meta = descriptor.meta || {};
-		const changed = descriptor.changed || meta.hashOfConfig !== this._hashOfConfig;
+
+		descriptor.meta.data ??= {};
+
+		assert(isPlainObject(descriptor.meta.data));
```
ybiquitous, line 70: [suggestion] How about making the following assertion unnecessary like this?

```suggestion
		/** @type {{ hashOfConfig?: string; }} */
		const metadata = descriptor.meta.data ??= {};
```

The following changes would be necessary too:

```diff
- descriptor.meta.data.hashOfConfig
+ metadata.hashOfConfig
```

lib/utils/FileCache.mjs
```diff
@@ -59,24 +59,59 @@ export default class FileCache {
...
 		}
 
 		return changed;
 	}
 
+	/**
+	 * @param {string} absoluteFilepath
+	 * @returns {boolean}
+	 */
+	hasFileCheckSumChanged(absoluteFilepath) {
```
ybiquitous, line 99: [suggestion] How about making the method private if you don't want to expose it?

```suggestion
	#hasFileCheckSumChanged(absoluteFilepath) {
```

lib/utils/FileCache.mjs
```diff
@@ -59,24 +59,59 @@ export default class FileCache {
...
+		descriptor.meta.data ??= {};
+
+		assert(isPlainObject(descriptor.meta.data));
+
+		const configChanged = descriptor.meta.data.hashOfConfig !== this._hashOfConfig;
+
+		let changed = false;
+
+		if (this._useCheckSum) {
+			changed = configChanged || fileCheckSumChanged;
```
ybiquitous, line 77: [suggestion] Since `getFileDescriptor()` already calculates a file hash if `useCheckSum` is true, `#hasFileCheckSumChanged()` seems to be simplified like this:

https://github.com/jaredwray/cacheable/blob/5ec1858bdd07aa5efc30e87f14253a951fbc83b5/packages/file-entry-cache/src/index.ts#L267-L271

```suggestion
			changed = configChanged || this.#hasFileCheckSumChanged(descriptor.key, descriptor.meta.hash);
```

```js
	/**
	 * @param {string} key
	 * @param {string | undefined} fileHash
	 * @returns {boolean}
	 */
	#hasFileCheckSumChanged(key, fileHash) {
		if (!fileHash) {
			return true;
		}

		const metaCache = this._fileCache.cache.getKey(key);

		if (fileHash !== metaCache?.hash) {
			return true;
		}

		return false;
	}
```

ybiquitous (commented):
@romainmenke Thank you for working on the `file-entry-cache` update. The direction looks good to me, but I've left a few suggestions. Could you take a look?

lib/__tests__/standalone-cache.test.mjs
Mouvedia: > I've also seen periodic failures of this test in CI.

Thanks for the fix.

lib/utils/FileCache.mjs
romainmenke: perfect! thank you!

lib/utils/FileCache.mjs
romainmenke: Thank you for this!

I suspected there might be a way to simplify by reusing the data already in `descriptor` but didn't quite see it yesterday :)

Much better like this 🙇

lib/__tests__/standalone-cache.test.mjs
romainmenke: `fs.utimes` [expects time in seconds](https://github.com/nodejs/node/issues/33227), i.e. `Date.now() / 1000`.

I don't think `Date.now() / 1000 + 1` is the most readable, but this is probably fine as it is only in a test and it is accompanied by a comment.

ybiquitous: Oh, I missed the unit difference between `Date.now()` (milliseconds) and `fs.utimes()` (seconds). This actually surprised me. Good catch 👍🏼 

> Values can be either numbers representing Unix epoch time in seconds

ybiquitous (approved):
Thank you. LGTM! 👍🏼

Issue #8248, opened 2024-12-29
Fix deprecation warnings to only display once per (custom) rule


jeddy3:
### What minimal example or steps are needed to reproduce the bug?

Run Stylelint on multiple files using a (custom) rule that triggers deprecation warnings.

Ref: https://github.com/stylelint/stylelint/pull/8244#issuecomment-2564728166

### What minimal configuration is needed to reproduce the bug?

```js
/** @type {import('stylelint').Config} */
export default {
  plugins: ["stylelint-declaration-strict-value"],
  rules: {
    "scale-unlimited/declaration-strict-value": "color",
  },
};
```

### How did you run Stylelint?

```shell
npx stylelint "code/*.css"   
```

### Which Stylelint-related dependencies are you using?

```
"stylelint": "latest",
"stylelint-declaration-strict-value": "latest"
```

### What did you expect to happen?

One deprecation warning per rule is displayed.

### What actually happened?

A warning for each file checked is displayed.

```console
npx stylelint "code/*.css"   
(node:48662) [stylelint:005] DeprecationWarning: `context.fix` is being deprecated.
Please pass a `fix` callback to the `report` utility of "scale-unlimited/declaration-strict-value" instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
(node:48662) [stylelint:005] DeprecationWarning: `context.fix` is being deprecated.
Please pass a `fix` callback to the `report` utility of "scale-unlimited/declaration-strict-value" instead.
```

### Do you have a proposal to fix the bug?

Display one warning and, if possible, clean it up to:

> [stylelint:005] DeprecationWarning: `context.fix` is being deprecated. Please pass a `fix` callback to the `report` utility of "scale-unlimited/declaration-strict-value" instead.

romainmenke:
Would something like this be sufficient?

```js
const emittedWarnings = new Set();

/**
 * Use this function for deprecation warnings, instead of `process.emitWarning()`.
 *
 * @param {string} message
 * @param {keyof CODES} codeKey
 * @param {string} detail
 * @returns {void}
 */
export default function emitDeprecationWarning(message, codeKey, detail) {
	const code = CODES[codeKey];

	const key = JSON.stringify({ message, options: { type: 'DeprecationWarning', code: `stylelint:${code}`, detail } })

	if (emittedWarnings.has(key)) return;

	emittedWarnings.add(key);

[...]

PR #8265, opened 2024-12-31
Fix deprecation warnings to only display once per (custom) rule

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/8248

Not especially happy with the addition of the `clearEmittedDeprecationWarnings` function.
If anyone has a good alternative?


Mouvedia:
> If anyone has a good alternative?

If we had `--quiet-deprecation-warnings=[002,006]` I would recommend a `class`.
e.g. `clear`, `emit` and `remove`/`init` methods
i.e. `init` to filter before or `remove` to filter after
Also if we supported adding custom deprecations in user land then `add` would be necessary too.
i.e. preferring our flag over `NODE_OPTIONS='--no-deprecation'`

Since we are doing neither a `class` would be overkill for just `clear` and `emit`.

lib/lintPostcssResult.cjs
```diff
@@ -100,7 +100,7 @@ async function lintPostcssResult(stylelintOptions, postcssResult, config) {
 			// due to the sheer number of rules in the wild that rely on it
 			get fix() {
 				if (shouldWarn && !warningFlag) {
```
Mouvedia, line 102: Can you check whether [this test](https://github.com/stylelint/stylelint/pull/8265/files#diff-a5338b3127d4362334c745fc36f7eb15c3b029c550b326b6a4c9ac8f578c85b7R437-R466) still passes if you remove the `warningFlag`?

.changeset/fluffy-waves-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: deprecation warnings to only display once per (custom) rule
```
Mouvedia, line 5: suggestion

```suggestion
Fixed: deprecation warnings to only display once per lint
```

lib/utils/emitDeprecationWarning.mjs
```diff
@@ -19,8 +21,27 @@ const CODES = {
  * @param {string} detail
  * @returns {void}
  */
-export default function emitDeprecationWarning(message, codeKey, detail) {
+export function emitDeprecationWarning(message, codeKey, detail) {
```
ybiquitous, line 24: [suggestion] Is it possible to keep `export default` to prevent the increase of trivial changes?

```suggestion
export default function emitDeprecationWarning(message, codeKey, detail) {
```

ybiquitous:
> Not especially happy with the addition of the `clearEmittedDeprecationWarnings` function.
> If anyone has a good alternative?

It seems we cannot avoid exposing the internals. 🤔 

But, to reduce code duplication, how about using the [`setupFilesAfterEnv`](https://jestjs.io/docs/configuration#setupfilesafterenv-array) option of Jest and clearing emitted warnings in a global `afterEach`? E.g.

`emitDeprecationWarning.mjs`:

```js
/**
 * Expose for testing.
 * @private
 */
export const _emittedWarnings = new Set();
```

`jest.setupAfterEnv.mjs` (new file):

```js
import { _emittedWarnings } from "./lib/utils/emitDeprecationWarning.mjs";

afterEach(() => {
	_emittedWarnings.clear();
});
```

`package.json`:

```json
{
  "setupFilesAfterEnv": [
    "<rootDir>/jest.setupAfterEnv.mjs"
  ],
}
```

.changeset/fluffy-waves-ring.md
romainmenke: It is "once per unique warning each time Stylelint is loaded" but I am unsure how specific we have to be here :)

Maybe a bit more vague is actually better?

> Fixed: only display unique deprecation warnings

lib/lintPostcssResult.cjs
romainmenke: I am unsure what you mean :)

Are you concerned that we don't have sufficient test coverage?
Is there a test we should add?

.changeset/fluffy-waves-ring.md
Mouvedia: Either vague or more specific, as you want.

lib/lintPostcssResult.cjs
Mouvedia: Ill checkout your branch and verify.

Mouvedia: OK I have confirmed my assumption.
i.e. `main` fails without `warningFlag` but not `fix-deprecation-warnings-to-only-display-once-per-custom-rule--unassuming-burmese-7f9819439e`

Could you please modify l93–106 of `lib/lintPostcssResult.mjs` to be
```
		const context = {
			...ctx,
			// context.fix is unlikely to be removed in the foreseeable future
			// due to the sheer number of rules in the wild that rely on it
			get fix() {
				if (shouldWarn) {
					emitDeprecationWarning(
						'`context.fix` is being deprecated.',
						'CONTEXT_FIX',
						`Please pass a \`fix\` callback to the \`report\` utility of "${ruleName}" instead.`,
					);
				}
```
?

i.e. `warningFlag` is not necessary anymore

Mouvedia (dismissed):
see https://github.com/stylelint/stylelint/pull/8265#discussion_r1900428412

lib/lintPostcssResult.cjs
romainmenke: I see what you are hinting at now.

This specific deprecation warning also had a mechanic to emit a specific warning only once. This mechanic is now obsolete and should be removed.

I've made the needed changes to remove this flag.

.changeset/fluffy-waves-ring.md
romainmenke: I think I actually prefer the original message :)
I think it best conveys what was changed to Stylelint users.

Mouvedia: The thing is we have deprecation warnings that are not related to rules.
e.g. `GITHUB_FORMATTER`
i.e. hence `per (custom) rule` is not really accurate IMO

romainmenke: But from a user's perspective those warnings would not have changed.
Only the deprecations that were very repetitive have been reduced.

So even if the code change was more "global", the message is still informative and useful for Stylelint users.

But I also don't feel strongly about this :)
The original message seems more to the point to me.

I'll let @jeddy3 chime in, having described the original issue.

ybiquitous (approved):
Thank you, LGTM 👍🏼

PR #8266, opened 2025-01-01
Fix `FileCache` false negatives when only contents have changed

romainmenke:

See https://github.com/stylelint/stylelint/pull/8256

It seems we made a small mistake in https://github.com/stylelint/stylelint/pull/8262 and that we didn't have sufficient test coverage.

`this._fileCache.getFileDescriptor(absoluteFilepath);` also writes the new content hash to `this._fileCache.cache` even though it is a getter method.

Any reads from the cache to compare the file hash after already calling `getFileDescriptor` would return the new hash, not the previous one.

So to compare content hashes we need to first read from the cache and only after call `getFileDescriptor`.

### DIFF
--- a/lib/utils/FileCache.cjs
+++ b/lib/utils/FileCache.cjs
@@ -57,6 +57,7 @@ class FileCache {
 		// Get file descriptor compares current metadata against cached
 		// one and stores the result to "changed" prop.w
 
+		const metaCache = this._fileCache.cache.getKey(this._fileCache.createFileKey(absoluteFilepath));
 		const descriptor = this._fileCache.getFileDescriptor(absoluteFilepath);
 
 		/** @type {{ hashOfConfig?: string; }} */
@@ -67,7 +68,7 @@ class FileCache {
 		let changed = false;
 
 		if (this._useCheckSum) {
-			changed = configChanged || this.#hasFileCheckSumChanged(descriptor.key, descriptor.meta.hash);
+			changed = configChanged || !metaCache?.hash || metaCache?.hash !== descriptor.meta.hash;
 		} else {
 			changed = configChanged || Boolean(descriptor.changed);
 		}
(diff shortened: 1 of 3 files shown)


lib/utils/FileCache.cjs
```diff
@@ -85,25 +86,6 @@ class FileCache {
...
-	#hasFileCheckSumChanged(key, fileHash) {
-		if (!fileHash) {
-			return true;
-		}
-
-		const metaCache = this._fileCache.cache.getKey(key);
-
-		if (fileHash !== metaCache?.hash) {
-			return true;
-		}
```
romainmenke, line 102: This condition would never match and this function would never return `true` because `fileHash` and `metaCache?.hash` are always the same value.

lib/utils/FileCache.mjs
```diff
@@ -59,6 +59,7 @@ export default class FileCache {
 		// Get file descriptor compares current metadata against cached
 		// one and stores the result to "changed" prop.w
 
+		const metaCache = this._fileCache.cache.getKey(this._fileCache.createFileKey(absoluteFilepath));
```
ybiquitous, line 62: [suggestion] I see `metaCache` is an `any` type in my IDE, so how about typing the variable like this?

```suggestion
		/** @type {import('file-entry-cache').FileDescriptorMeta | undefined} */
		const metaCache = this._fileCache.cache.getKey(this._fileCache.createFileKey(absoluteFilepath));
```

lib/utils/FileCache.mjs
```diff
@@ -69,7 +70,7 @@ export default class FileCache {
 		let changed = false;
```
ybiquitous, line 70: [suggestion] Minor. This assigning seems unnecessary because the `changed` variable is set within the next `if` block;

```suggestion
		let changed;
```

lib/utils/FileCache.mjs
```diff
@@ -69,7 +70,7 @@ export default class FileCache {
...
 
 		if (this._useCheckSum) {
-			changed = configChanged || this.#hasFileCheckSumChanged(descriptor.key, descriptor.meta.hash);
+			changed = configChanged || !metaCache?.hash || metaCache?.hash !== descriptor.meta.hash;
```
ybiquitous, line 73: [suggestion] How about polishing the condition like this? The `metaCache?` repetition seems a bit less readable to me:

```suggestion
			changed = configChanged || !metaCache?.hash || metaCache.hash !== descriptor.meta.hash;
```

ybiquitous (commented):
Thank you for the fix and test case addition. I missed the logic in the previous PR. 😅 

Already almost LGTM, but could you take a look at my minor suggestions?

ybiquitous (approved):
Thank you for your quick action. LGTM 👍🏼

Mouvedia:
@romainmenke I dunno if it's confirmed, but could you take a look at webpack-contrib/stylelint-webpack-plugin#364?

Issue #8233, opened 2024-12-26
Refactor to remove `fixed` from `FixerData`


romainmenke:
see: https://github.com/stylelint/stylelint/blob/3751cfb4c6ba8c94dbd88db75ff6bd77a1aa70bc/lib/utils/report.mjs#L168-L202

While working on https://github.com/stylelint/stylelint/issues/7192 I found that the `fixed` flag in `FixerData` is confusing and limiting.

I assume the intention was to be able to differentiate between a fix that was purely run to compute the replacement text and a fix that was also applied on the final output.

However I don't think there is a use case for that.

Either users want all fixes to be applied by Stylelint or they want no fixes to be applied at all.
Optionally users might want a report of potential fixes that could have been applied.

But I don't think users ever want both interleaved, or am I overlooking something?

------

I suggest we remove this flag and refactor `report` to only record fixes in `fixersData`.

i.e.

```js
/** @param {Problem & { line: number }} problem */
function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
	const { disabledRanges, config = {}, fixersData } = stylelint;

	if (!isFn(fix)) {
		return false;
	}

	const shouldFix = Boolean(config.fix && !config.rules?.[ruleName][1]?.disableFix);
	const mayFix =
		shouldFix && (config.ignoreDisables || !isDisabled(ruleName, line, disabledRanges));

	if (!mayFix) return false;

	fix();
	addFixData({ fixersData, ruleName });

	return true;
}
```

_This also requires refactoring the `verbose` formatter so that it simply reports all fixers data. Currently it filters entries where `fixed == true`._

-----

To resolve https://github.com/stylelint/stylelint/issues/7192 we will eventually have to calculate replacement text, which will likely require running `fix()`.

By the current implementation this would set the `fixed` flag to true.

But for https://github.com/stylelint/stylelint/issues/7192 it is also important that no fixers are actually applied.

I think the fixed flag is mainly ambiguous in its meaning:
- a fix was computed?
- a fix was applied to your source files?

Removing it altogether is a simple way to avoid the ambiguity.

-------

I do think there might a need for several other flags:
- were some fixes skipped because of configuration comments?
- are there any fixes left that couldn't be computed in a first pass?

romainmenke:
see : https://github.com/stylelint/stylelint/blob/cf17793d6e4bd159352d78d142cbcda645ee952a/lib/utils/report.mjs#L178-L221

[...]

PR #8267, opened 2025-01-02
Refactor to remove `fixed` from `FixerData`

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/8233

`fixersData` is not part of our public API and the verbose formatter still produces the same reports as before.

### DIFF
--- a/lib/utils/report.mjs
+++ b/lib/utils/report.mjs
@@ -218,8 +218,6 @@ function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
 	const { disabledRanges, config = {}, fixersData } = stylelint;
 
 	if (!isFn(fix)) {
-		addFixData({ fixersData, ruleName, fixed: false });
-
 		return false;
 	}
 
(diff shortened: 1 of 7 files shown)


lib/utils/report.mjs
```diff
@@ -241,13 +239,12 @@ function isFixApplied({ fix, line, result: { stylelint }, ruleName }) {
  * @param {StylelintPostcssResult['fixersData']} o.fixersData
  * @param {string} o.ruleName
  * @param {Range} [o.range]
- * @param {boolean} o.fixed
  * @todo stylelint/stylelint#7192
```
Mouvedia, line 242: Are you planning to eventually pass the range to `addFixData`?
e.g.

```
const range = {
  start: fixedNode.source.start,
  end: fixedNode.source.end,
};

addFixData({ fixersData, ruleName, range });
```

So that we can close #4664 since we will be able to add the range+path to the fixed section of the `verbose` formatter's output.
i.e. click in the terminal, get to the line in your IDE and eventually scroll horizontally to reach the range

romainmenke: See: https://github.com/stylelint/stylelint/issues/4664#issuecomment-2568437149

ybiquitous (approved):
Thanks, LGTM 👍🏼

PR #8269, opened 2025-01-03
Refactor to remove `FixerData` type

Mouvedia:

follow-up of #8267

1.
With the removal of `fixed`, it was absurd to keep `FixerData` if we were not even planning to add `range` to it.
see https://github.com/stylelint/stylelint/pull/8267#pullrequestreview-2528238317

2.
`StylelintPostcssResult['fixersData']` could be renamed in #8234 to avoid confusion with [`fixData`](https://github.com/stylelint/stylelint/pull/8234/files#diff-cba25f954347e963307879ca83b95b319e9003d9b4b51fd350d3f4110566d3a8R72).
i.e. semantic creep


romainmenke (approved):
Thank you for this @Mouvedia. LGTM 👍

PR #8270, opened 2025-01-03
Test invalid options for all rules

ybiquitous:

None.

This increases the test coverage when invalid options are passed to all the rules.

```js
if (!validOptions) {
	return; //<- This line is now covered.
}
```

See the coverage on Codecov: https://app.codecov.io/gh/stylelint/stylelint/pull/8270

### DIFF
--- a/lib/rules/__tests__/index.test.mjs
+++ b/lib/rules/__tests__/index.test.mjs
@@ -5,6 +5,8 @@ import { fileURLToPath } from 'node:url';
 import { join } from 'node:path';
 import { readFile } from 'node:fs/promises';
 
+import postcss from 'postcss';
+
 import rules from '../index.mjs';
 
 const require = createRequire(import.meta.url);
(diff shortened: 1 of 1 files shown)


lib/rules/__tests__/index.test.mjs
```diff
@@ -74,6 +76,25 @@ describe('custom message option', () => {
...
 
+describe('invalid options', () => {
+	test.each(ruleNames)('"%s" should handle invalid options', async (name) => {
+		const rule = await rules[name];
+		const primaryOption = false;
+		const secondaryOptions = {};
+		const context = {};
+		const root = postcss.root();
+		const result = Object.assign(root.toResult(), {
+			stylelint: { config: { validate: true } },
```
Mouvedia, line 87: note

`validate` configuration property wasn't added
cf #8087

ybiquitous: `config.validate` is still present. Should we remove it?

https://github.com/stylelint/stylelint/blob/a0c6dc43d32f3c93e7bcef090c06aefd59373e08/lib/utils/validateOptions.mjs#L19-L21

lib/rules/__tests__/index.test.mjs
```diff
@@ -74,6 +76,25 @@ describe('custom message option', () => {
 	);
 });
 
+describe('invalid options', () => {
+	test.each(ruleNames)('"%s" should handle invalid options', async (name) => {
+		const rule = await rules[name];
+		const primaryOption = false;
```
Mouvedia, line 82: note

We have a related test named "with options (boolean)" in `validateOptions.test.mjs`.

Issue #1437, opened 2016-06-12
Support prefixed property and at-rules in all rules.


evilebottnawi:
Two step:
- [x] Add tests to some rules for check they support vendor prefix value (merge in master)
- [x] Add support prefixed property, value and etc in all rules (merge in v7)
  - keyframe-declaration-no-important
  - no-unknown-animations
  - time-no-imperceptible

For other issue and PR:
 - [x] Recheck on support case insensitivity
  - number-max-precision (import)
  - function-linear-gradient-no-nonstandard-direction (add flag `i` to `isStandardDirection`) and to check `deg`, `grad`, `rad`, `turn` (line 34)

I think it is breaking changes and should completed in v7.
Example https://github.com/stylelint/stylelint/blob/master/src/rules/no-unknown-animations/index.js#L28 we see only `keyframes`, `kEyFrAmEs`, `KEYFRAMES`, but not see on `-webkit-keyframes` and etc.

jeddy3:
This issue is for the work started in https://github.com/stylelint/stylelint/pull/1278, right?

evilebottnawi:
@jeddy3 right, in order not to forget and put v7 milestone.

jeddy3:
Agreed.

evilebottnawi:
@jeddy3 @ntwb @davidtheclark I checked all the rules twice, but still fear miss, it would be great if you ran bu code the second eyes, see first post (above) that i found

evilebottnawi:
Done in https://github.com/stylelint/stylelint/pull/1591#issuecomment-230872130

PR #8272, opened 2025-01-03
Refactor to commonize regex patterns

ryo-manba:

See https://github.com/stylelint/stylelint/pull/8211#discussion_r1892243036

I have commonized regex patterns related to at-rules, including the section mentioned in [#8211 (comment)](https://github.com/stylelint/stylelint/pull/8211#discussion_r1892243036) and other similar patterns.


lib/utils/atRuleRegex.mjs
Mouvedia, line 1: suggestion
minor

atRuleRegex.mjs > atRuleRegexes.mjs

Mouvedia:
Utilities are normally for complicate things, I find `/^media$/i` pretty simple.
If we were to go all the way you would have `MEDIA_REGEX` for many more rules than `at-rule-*` ones.
@ybiquitous I don't know, what do you think?
Maybe only export UNSUPPORTED_NESTING_REGEX and KEYFRAMES_REGEX and rename the util `regexex.mjs`?

lib/utils/atRuleRegex.mjs
```diff
@@ -0,0 +1,9 @@
+import { nestingSupportedAtKeywords } from '../reference/atKeywords.mjs';
+
+export const MEDIA_REGEX = /^media$/i;
+export const KEYFRAMES_REGEX = /^(-(o|moz|ms|webkit)-)?keyframes$/i;
```
Mouvedia, line 4: off-topic

@jeddy3 @ybiquitous 
Should `keyframes-name-pattern` use this regex?

https://github.com/stylelint/stylelint/blob/8ef1792d08755282b74216ead1508af94556538d/lib/rules/keyframes-name-pattern/index.mjs#L32

ref #1437

Mouvedia:
`no-unknown-animations` never had a `^`; should it use `KEYFRAMES_REGEX` too?

ybiquitous:
For <https://github.com/stylelint/stylelint/pull/8272#issuecomment-2570123993>.

@Mouvedia I understand your thought, but if we introduce such utilities, I think it's unavoidable that we also want to contain other regexes, even if some are pretty simple.

One of the benefits of implementing regexes as utilities is that we can quickly test them.

So I generally agree with these changes. 👍🏼

lib/utils/atRuleRegex.mjs
ybiquitous: Great catch. I think we should use this regex for the rule, too. :+1:

lib/utils/atRuleRegex.mjs
ybiquitous: Agree. I prefer `regexes.mjs` and want to introduce nesting for future regexes.

Also, this is not a strong opinion, but adopting camelcase naming might be good.

E.g.

`lib/utils/regexes.mjs`:

```js
export const atRuleRegexes = {
  media: /^media$/i,
  keyframes: /^(-(o|moz|ms|webkit)-)?keyframes$/i,
  // ...
};
```

Usage:

```js
import { atRuleRegexes } from '../utils/regexes.mjs';

root.walkAtRules(atRuleRegexes.media, (node) => {
  // ...
});
```

ybiquitous:
For <https://github.com/stylelint/stylelint/pull/8272#issuecomment-2570727285>.

Good catch. Yes, I agree `no-unknown-animations` has `KEYFRAMES_REGEX`. 👍🏼 

https://github.com/stylelint/stylelint/blob/8ef1792d08755282b74216ead1508af94556538d/lib/rules/no-unknown-animations/index.mjs#L29

lib/utils/atRuleRegex.mjs
Mouvedia: Since it's part of the test it can be considered a refactor.
i.e. could be done in this PR

lib/utils/atRuleRegex.mjs
ryo-manba: Looks good! Changed in the following commit.
[Rename atRuleRegex to regexes](https://github.com/stylelint/stylelint/pull/8272/commits/ee9a2fdba0646f826e7165b2d0d398ccb0046796)

lib/utils/atRuleRegex.mjs
ryo-manba: Good suggestion! I've fixed in the following commit.
[Refactor to use regexes](https://github.com/stylelint/stylelint/pull/8272/commits/25c8276b8dbaf6eea9aba96d7d77d40bf6e85e7d)

lib/utils/regexes.mjs
```diff
@@ -0,0 +1,11 @@
+import { nestingSupportedAtKeywords } from '../reference/atKeywords.mjs';
+
+export const atRulesRegexes = {
```
ybiquitous, line 3: [suggestion] This is quite a personally opinion and just bike-shedding, but I feel `s` at the end of `atRules` is a bit less readable since the name already ends with `s` in `Regexes`.

So, how about removing a doubled `s` like this?

```suggestion
export const atRuleRegexes = {
```

This is just my opinion, so if you are against it, you don't need to address it. 😃

ybiquitous (approved):
Thank you, LGTM 👍🏼 

I've left a few minor suggestions, but they aren't a blocker. 😃 

---

(optional) Do you think adding a regex also for `@import` makes sense?

https://github.com/stylelint/stylelint/blob/3562dd1acfa69abd708acb1ae6c06601565aaa6a/lib/rules/import-notation/index.mjs#L44

https://github.com/stylelint/stylelint/blob/3562dd1acfa69abd708acb1ae6c06601565aaa6a/lib/rules/no-duplicate-at-import-rules/index.mjs#L31

PR #8273, opened 2025-01-03
Document how to handle emitted deprecation warnings

### LINKED ISSUE #8249: Document how to handle emitted deprecation warnings

To quiet deprecation warnings in a shared config rather than changing multiple packages to use the CLI flag.

Ref: https://github.com/stylelint/stylelint/pull/8244#issuecomment-2564728166
Ref: https://github.com/AndyOGo/stylelint-declaration-strict-value/issues/379#issuecomment-2464946770

Be able to quiet deprecation warnings from the config.

```js
/** @type {import('stylelint').Config} */
export default {
  quietDeprecationWarnings: true,
  plugins: ["stylelint-declaration-strict-value"],
  rules: {
    "scale-unlimited/declaration-strict-value": "color",
  },
};
```

jeddy3:

Closes #8249 

I'd also like to add something for plugin authors that describes https://github.com/stylelint/stylelint/issues/8249#issuecomment-2564755567, i.e. handling warning events, but I could only silence all warnings in a plugin using:

```js
process.removeAllListeners("warning");
```

I thought I'd be able to silence individual warnings using:

```js
process.on("warning", (warning) => {
  if (warning.code === "stylelint:005") return;
});
```

But that didn't work. Am I missing something?

### DIFF
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -396,7 +396,7 @@ When publishing such configurations, use the `stylelint-config` keyword within y
 
 We sometimes introduce deprecation warnings that impact plugin authors. You can individually quiet these in your plugin by handling the `emitWarning` event.
 
-For example, to quiet the deprecationg warning for `context.fix`:
+For example, to quiet the deprecation warning for `context.fix`:
 
 ```diff js
 const ruleName = "plugin/foo-bar-qux";
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -402,7 +402,9 @@ For example, to quiet the deprecation warning for `context.fix`:
 const ruleName = "plugin/foo-bar-qux";
 
 + const original = process.emitWarning;
-+ process.emitWarning = function (message, options) {
++ process.emitWarning = function emitWarning(warning, type, code, ctor) {
++   const options = arguments[1];
++ 
 +  if (
 +    options &&
 +    typeof options === "object" &&
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -133,22 +133,23 @@ We sometimes introduce deprecation warnings that impact plugin authors. You can 
 
 For example, to quiet the deprecationg warning for `context.fix`:
 
-```js
-const ruleName = "foo-bar-qux";
-const original = process.emitWarning;
-process.emitWarning = function (message, options) {
-  if (
-    options &&
-    typeof options === "object" &&
-    options.type === "DeprecationWarning" &&
-    options.code === "stylelint:005" &&
-    options.detail.includes(ruleName)
-  ) {
-    return;
-  }
+```diff js
+const ruleName = "plugin/foo-bar-qux";
 
-  original.apply(process, arguments);
-};
++ const original = process.emitWarning;
++ process.emitWarning = function (message, options) {
++  if (
++    options &&
++    typeof options === "object" &&
++    options.type === "DeprecationWarning" &&
++    options.code === "stylelint:005" &&
++    options.detail.includes(ruleName)
++  ) {
++    return;
++  }
++
++  original.apply(process, arguments);
++ };
 ```
 
 > [!WARNING]
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -140,7 +140,8 @@ process.emitWarning = function (message, options) {
     options &&
     typeof options === "object" &&
     options.type === "DeprecationWarning" &&
-    options.code === "stylelint:005"
+    options.code === "stylelint:005" &&
+    options.detail.includes(ruleName)
   ) {
     return;
   }
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -134,6 +134,7 @@ We sometimes introduce deprecation warnings that impact plugin authors. You can 
 For example, to quiet the deprecationg warning for `context.fix`:
 
 ```js
+const ruleName = "foo-bar-qux";
 const original = process.emitWarning;
 process.emitWarning = function (message, options) {
   if (
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -137,9 +137,8 @@ For example, to quiet the deprecationg warning for `context.fix`:
 const original = process.emitWarning;
 process.emitWarning = function (message, options) {
   if (
-    message &&
     options?.type === "DeprecationWarning" &&
-    options.code === "stylelint:005"
+    options?.code === "stylelint:005"
   )
     return;
 
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -143,7 +143,7 @@ process.emitWarning = function (message, options) {
     return;
   }
 
-  original.call(process, message, options);
+  original.apply(process, arguments);
 };
 ```
 
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -404,7 +404,7 @@ const ruleName = "plugin/foo-bar-qux";
 + const original = process.emitWarning;
 + process.emitWarning = function emitWarning(warning, type, code, ctor) {
 +   const options = arguments[1];
-+ 
++
 +  if (
 +    options &&
 +    typeof options === "object" &&
--- a/docs/developer-guide/plugins.md
+++ b/docs/developer-guide/plugins.md
@@ -127,34 +127,6 @@ const ruleFunction = (primary, secondaryOptions) => {
 };
 ```
 
-## Quiet deprecation warnings
-
-We sometimes introduce deprecation warnings that impact plugin authors. You can individually quiet these in your plugin by handling the `emitWarning` event.
-
-For example, to quiet the deprecationg warning for `context.fix`:
-
-```diff js
-const ruleName = "plugin/foo-bar-qux";
-
-+ const original = process.emitWarning;
-+ process.emitWarning = function (message, options) {
-+  if (
-+    options &&
-+    typeof options === "object" &&
-+    options.type === "DeprecationWarning" &&
-+    options.code === "stylelint:005" &&
-+    options.detail.includes(ruleName)
-+  ) {
-+    return;
-+  }
-+
-+  original.apply(process, arguments);
-+ };
-```
-
-> [!WARNING]
-> Note that this monkey patch might have side effects.
-
 ## Testing
 
 You can use either:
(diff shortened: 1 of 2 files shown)


Mouvedia:
> I thought I'd be able to silence individual warnings using:
> 
> ```js
> process.on("warning", (warning) => {
>   if (warning.code === "stylelint:005") return;
> });
> ```
> 
> But that didn't work. Am I missing something?

Try

```js
const original = process.emitWarning;
process.emitWarning = function (message, options) {
  if (message && options?.type === 'DeprecationWarning' && options.code === 'stylelint:005') return;

  return original.call(process, message, options);
};
```

Mouvedia:
> Am I missing something?

Can you `console.dir(warning);` to check what properties it holds?

docs/user-guide/options.md
```diff
@@ -259,4 +259,9 @@ CLI flag: `--quiet-deprecation-warnings`
 
 Ignore deprecation warnings.
 
+> [!TIP]
+> For Node.js 20.11.0+, you can disable individual deprecation warnings using the Node.js [`--disable-warning`](https://nodejs.org/api/cli.html#--disable-warningcode-or-type) mechanism, e.g.:
+>
+> `NODE_OPTIONS='--disable-warning=stylelint:005' npx stylelint "**/*.css"`
```
ybiquitous, line 265: [suggestion] How about converting it to a block like this? It can be easily copied via the button and supports syntax highlighting.

```suggestion
> ```shell
> NODE_OPTIONS='--disable-warning=stylelint:005' npx stylelint "**/*.css"
> ```
```

docs/developer-guide/plugins.md
```diff
@@ -127,6 +127,26 @@ const ruleFunction = (primary, secondaryOptions) => {
...
+
+```js
+const original = process.emitWarning;
+process.emitWarning = function (message, options) {
+  if (
+    message &&
+    options?.type === "DeprecationWarning" &&
+    options.code === "stylelint:005"
+  )
+    return;
```
ybiquitous, line 143: [suggestion] This is just my opinion, but I'd recommend `return` is surrounded by `{}` for clarity:

```suggestion
  ) {
    return;
  }
```

docs/developer-guide/plugins.md
```diff
@@ -127,6 +127,26 @@ const ruleFunction = (primary, secondaryOptions) => {
...
+
+  original.call(process, message, options);
```
ybiquitous, line 145: [suggestion] I'd recommend using the `arguments` object, considering several arguments patterns:

```suggestion
  original.apply(process, arguments);
```

jeddy3:
```console
{
  type: 'DeprecationWarning',
  code: 'stylelint:005',
  detail: 'Please pass a `fix` callback to the `report` utility of "plugin/test" instead.'
}
```

I've reduced the example down.

docs/developer-guide/plugins.md
```diff
@@ -127,6 +127,25 @@ const ruleFunction = (primary, secondaryOptions) => {
...
+We sometimes introduce deprecation warnings that impact plugin authors. You can individually quiet these in your plugin by handling the `emitWarning` event.
+
+For example, to quiet the deprecationg warning for `context.fix`:
+
+```js
+const original = process.emitWarning;
+process.emitWarning = function (message, options) {
+  if (
+    options?.type === "DeprecationWarning" &&
+    options?.code === "stylelint:005"
```
ybiquitous, line 141: [suggestion] Seeing `@types/node`, `emitWarning` supports several patterns for the arguments:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/486660167625287171c6094685c0edac5600510e/types/node/process.d.ts#L790-L793

So, I suggest checking if `options` is an object or not. E.g.

```suggestion
    options &&
    typeof options === "object" &&
    options.type === "DeprecationWarning" &&
    options.code === "stylelint:005"
```

docs/developer-guide/plugins.md
```diff
@@ -127,6 +127,25 @@ const ruleFunction = (primary, secondaryOptions) => {
 };
 ```
 
+## Quiet deprecation warnings
+
+We sometimes introduce deprecation warnings that impact plugin authors. You can individually quiet these in your plugin by handling the `emitWarning` event.
```
Mouvedia, line 132: question

> You can individually quiet these in your plugin by handling the `emitWarning` event.

Wouldn't the code at l137–146 be global?
i.e. it would disable the deprecation warning for the other plugins as well

ybiquitous (commented):
Sorry, I misunderstood `process.on('warning')` could suppress warnings. 🙇🏼 

I also think overriding `process.emitWarning` is a unique way to do that, but I'm a bit worried that such a monkey-patching might have side effects since it rewrites a global function.

So, how about adding a simple note too, just in case? E.g.

```markdown
> [!WARNING]
> Note that this monkey patch might have side effects.
```

docs/developer-guide/plugins.md
jeddy3: We typically include the `ruleName` in the details for deprecations that impact rule authors.

005:

```
Please pass a `fix` callback to the `report` utility of "plugin/test" instead.
```

007:

```
Please pass both `index` and `endIndex` as arguments in the `utils.report()` function of "plugin/test" instead.
```

As such, this seems to limit the quieting to individual rules?:

```js
const original = process.emitWarning;
process.emitWarning = function (message, options) {
  if (
    options &&
    typeof options === "object" &&
    options.type === "DeprecationWarning" &&
    options.code === "stylelint:007" &&
    options.detail.includes(ruleName)
  ) {
    return;
  }

  original.apply(process, arguments);
};
```

Works locally when tested.

I think we can assume most rule authors will update their code to remove deprecations, and only a few will use this monkey-patching approach to maintain compatibility until the next major release, so it doesn't need to be eloquent.

docs/developer-guide/plugins.md
```diff
@@ -127,6 +127,29 @@ const ruleFunction = (primary, secondaryOptions) => {
...
+
+For example, to quiet the deprecationg warning for `context.fix`:
+
+```js
+const original = process.emitWarning;
```
Mouvedia, line 137:
````suggestion
const ruleName = "foo-bar-qux";
const original = process.emitWarning;
```

Mouvedia:
> Sorry, I misunderstood `process.on('warning')` could suppress warnings. 🙇🏼

see https://github.com/stylelint/stylelint/pull/8273#issuecomment-2571276364
Can't you use `message` instead of `code`?
I didn't check though, I am just guessing that a plain `Error` is being passed.

ybiquitous:
Ah, `options.detail` isn't perfect, but it seems acceptable now. 😅 

Please note that `message` or `options.detail` are less stable than `code`. These aren't Stylelint's public API, though.

I think maintaining such a monkey patch is still the responsibility of plugin authors, and we should emphasize that the patching should be a kind of workaround.

ybiquitous:
> we should emphasize that the patching should be a kind of workaround.

Considering the importance, the "Quiet deprecation warnings" section probably should be put at the end of the document (`plugins.md`).

jeddy3:
> Please note that message or options.detail are less stable than code. These aren't Stylelint's public API, though.

I agree. I think we can guarantee the `code` will be stable and that `details` will contain the rule name.

I've moved the section to the bottom of the file.

ybiquitous (approved):
LGTM, thank you 👍🏼

docs/developer-guide/plugins.md
```diff
@@ -391,3 +391,31 @@ When publishing such configurations, use the `stylelint-config` keyword within y
...
+
+We sometimes introduce deprecation warnings that impact plugin authors. You can individually quiet these in your plugin by handling the `emitWarning` event.
+
+For example, to quiet the deprecation warning for `context.fix`:
+
+```diff js
+const ruleName = "plugin/foo-bar-qux";
+
++ const original = process.emitWarning;
++ process.emitWarning = function (message, options) {
```
romainmenke, line 405:
```suggestion
+ process.emitWarning = function emitWarning(warning, type, code, ctor) {
```

So that `process.emitWarning` retains the correct arity of `4` and a name of `emitWarning`

Mouvedia: We are using the other signature.

i.e. https://nodejs.org/api/process.html#processemitwarningwarning-options vs https://nodejs.org/api/process.html#processemitwarningwarning-type-code-ctor

I think we should not have this section at all because it will become too complicated.
If someone wants to overwrite a global Node.js function—which is probably a bad practice unless it's in a test—it shouldn't be recommended by us.

romainmenke: You can test this with `console.log(process.emitWarning.name)` and `console.log(process.emitWarning.length)`

There you will see that the name should be `emitWarning` and the arity should be `4`, even if newer node versions also support a variant where the second argument is an object.

Overriding the function with a new function with only two arguments without a name changes the signature. It will now have no name and an arity of `2`

Mouvedia: What I meant is, that [this problem](https://github.com/stylelint/stylelint/pull/8273#discussion_r1903098170) can be addressed only if you have access to `detail`, correct?
i.e. how do you address it without `detail`?

romainmenke: That is unrelated to the arity or name of a function.

Please just try the code, I am 100% certain that my suggestion is correct and follows best practices for wrapping platform/global functions. I am commenting from years of experience providing polyfills for the web as part of my work on `polyfill-library`.

Mouvedia: I get it now, I didn't understand what you were suggesting.
It makes perfect sense.

Still it's kind of a bad practice but the warning will do I guess?

romainmenke (approved):
Thank you. LGTM 👍

docs/developer-guide/plugins.md
```diff
@@ -391,3 +391,33 @@ When publishing such configurations, use the `stylelint-config` keyword within y
...
+
+We sometimes introduce deprecation warnings that impact plugin authors. You can individually quiet these in your plugin by handling the `emitWarning` event.
+
+For example, to quiet the deprecation warning for `context.fix`:
+
+```diff js
+const ruleName = "plugin/foo-bar-qux";
+
++ const original = process.emitWarning;
++ process.emitWarning = function emitWarning(warning, type, code, ctor) {
```
AndyOGo, line 405: Thank you very much for looking at this @jeddy3, @ybiquitous, @romainmenke, @Mouvedia 

Sorry for joining a bit too late, but I would prefer to not monkey patch the global `process.emitWarning` API because it comes, as you commented, with unknown side effects.

I found the pitfalls of monkey patching pointed out here very useful:
- https://en.wikipedia.org/wiki/Monkey_patch#Pitfalls
- https://kettanaito.com/blog/why-patching-globals-is-harmful

---

Regarding my plugin I was faced with a rather big commotion of end users regarding the deprecation of `context.fix` API warnings.
https://github.com/AndyOGo/stylelint-declaration-strict-value/issues/379

romainmenke: @AndyOGo I do understand your concerns and hesitation to author a patch to temporarily work around too many deprecation warnings for end users.

But I don't think any of those listed concerns hold true in this case.
It is not a patch to add non-standard behavior, but rather works like middleware. This, in general, is a lot safer and less prone to issues. Especially when written correctly.

I am definitely not saying that monkey patching is good :), but rather that it can be an ok compromise to improve a temporary pain point.

-----

If you do have suggestions, please comment on [this issue](https://github.com/stylelint/stylelint/issues/8249), rather than on this PR as discussions should happen in issues :)
We'd love to hear from plugin authors on what can be done to improve things and what could help you migrate more quickly.

AndyOGo: thx @romainmenke 
I followed up there
https://github.com/stylelint/stylelint/issues/8249#issuecomment-2573961684

PR #8277, opened 2025-01-05
Fix `at-rule-(dis)allowed-list`, `at-rule-no-vendor-prefix`, `at-rule-property-required-list` message argument

Mouvedia:
> Which PR, if any, is this related to?

follow-up of #8236 and #8252

1.
`at-rule-descriptor-no-unknown`, `at-rule-no-deprecated`, `at-rule-prelude-no-invalid` and `at-rule-no-unknown` pass `'@foo'`.
`at-rule-(dis)allowed-list`, `at-rule-no-vendor-prefix`, `at-rule-property-required-list` used to pass `'foo'`.

2.
`at-rule-no-vendor-prefix` had the right message but the wrong argument.
i.e. the @ should not be added in the message itself

3.
13 rules that have a message that starts with `Unexpected unknown` use the `rejected` key for it.
`no-unknown-custom-media` used `expected`.

4.
The argument passed should respect the current style of the user's codebase.
i.e. [`"at-rule-name-case": "upper"`](https://github.com/stylelint-stylistic/stylelint-stylistic/tree/main/lib/rules/at-rule-name-case)

### DIFF
--- a/.changeset/perfect-pots-destroy.md
+++ b/.changeset/perfect-pots-destroy.md
@@ -1,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Changed: `at-rule-(dis)allowed-list`, `at-rule-no-vendor-prefix`, `at-rule-property-required-list` message argument
--- a/.changeset/perfect-pots-destroy.md
+++ b/.changeset/perfect-pots-destroy.md
@@ -1,5 +1,5 @@
 ---
-"stylelint": minor
+"stylelint": patch
 ---
 
-Changed: `at-rule-(dis)allowed-list`, `at-rule-no-vendor-prefix`, `at-rule-property-required-list` message argument
+Fixed: `at-rule-(dis)allowed-list`, `at-rule-no-vendor-prefix`, `at-rule-property-required-list` message argument
--- a/lib/rules/at-rule-allowed-list/__tests__/index.mjs
+++ b/lib/rules/at-rule-allowed-list/__tests__/index.mjs
@@ -71,7 +71,7 @@ testRule({
 			column: 7,
 			endLine: 2,
 			endColumn: 13,
-			message: messages.rejected('mixin'),
+			message: messages.rejected('@mixin'),
 			description: '@rule not from an allowed list; independent rule.',
 		},
 	],
@@ -126,7 +126,7 @@ testRule({
 			code: `
       @mixin name ($p) {}
     `,
-			message: messages.rejected('mixin'),
+			message: messages.rejected('@mixin'),
 			line: 2,
 			column: 7,
 			endLine: 2,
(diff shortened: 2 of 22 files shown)


Mouvedia:
@jeddy3 is the argument change a fix?
If so Ill add a changeset.

ybiquitous (commented):
Thank you, I love this change 👍🏼 

I think we should add a changelog item for the fixed rule messages.

.changeset/perfect-pots-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Changed: `at-rule-(dis)allowed-list`, `at-rule-no-vendor-prefix`, `at-rule-property-required-list` message argument
```
Mouvedia, line 5: We have used _Changed_ outside of majors before, check the changelog.
I think it's more accurate.
But still if you want to use _Fixed_ you can apply

```suggestion
"stylelint": patch
---

Fixed: `at-rule-(dis)allowed-list`, `at-rule-no-vendor-prefix`, `at-rule-property-required-list` message argument
```

ybiquitous: I think we can consider this change as a bugfix for the messages, such as #8186.

See also:
https://github.com/stylelint/stylelint/blob/3562dd1acfa69abd708acb1ae6c06601565aaa6a/docs/about/semantic-versioning.md#L7-L9

In addition, the change is related to the messages themselves rather than "message argument", so how about the following? For example:

```diff
-Fixed: `at-rule-(dis)allowed-list`, `at-rule-no-vendor-prefix`, `at-rule-property-required-list` message argument
+Fixed: `at-rule-(dis)allowed-list`, `at-rule-no-vendor-prefix`, `at-rule-property-required-list` messages to include `@` for consistency
```

Mouvedia: > In addition, the change is related to the messages themselves rather than "message argument", so how about the following? For example:
> 
> ```diff
> -Fixed: `at-rule-(dis)allowed-list`, `at-rule-no-vendor-prefix`, `at-rule-property-required-list` message argument
> +Fixed: `at-rule-(dis)allowed-list`, `at-rule-no-vendor-prefix`, `at-rule-property-required-list` messages to include `@` for consistency
> ```

If we were to change to that we would have to remove `at-rule-no-vendor-prefix`, as explained in [OP](https://github.com/stylelint/stylelint/pull/8277#issue-2769453813) it has the right message but the wrong argument which matters for users of [custom messages](https://stylelint.io/user-guide/configure/#message).
i.e. the argument change is more important

> I think we can consider this change as a bugfix for the messages, such as https://github.com/stylelint/stylelint/pull/8186.

That PR was just changing the message, not the argument itself.
You can apply the suggested change in https://github.com/stylelint/stylelint/pull/8277#discussion_r1903672123 if you want.

ybiquitous: Understood, can you apply the suggestion? The commit button doesn't appear for me.

ybiquitous (approved):
LGTM, thank you 👍🏼

PR #8282, opened 2025-01-06
Suppress diffs in generated `.cjs` files

ybiquitous:

None. Just for developer experience.

Generated `.cjs` files often make it hard to review diffs in pull requests.

Ref https://github.com/github-linguist/linguist/blob/63cfd70d54ee8f76c41a73fe56689ed8229c9622/docs/overrides.md#generated-code

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,5 @@
 * text eol=lf
 *.png binary
+
+# See https://github.com/github-linguist/linguist/blob/63cfd70d54ee8f76c41a73fe56689ed8229c9622/docs/overrides.md#generated-code
+lib/**/*.cjs linguist-generated


ybiquitous:
The test via f86230543 passed ✅ 

image

If reviewers check it, I'll revert the commit.

.gitattributes
```diff
@@ -1,2 +1,5 @@
 * text eol=lf
 *.png binary
+
+# See https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github
+lib/**/*.cjs linguist-generated
```
romainmenke, line 5: We also have some `.cjs` files in tests, e.g. `lib/__tests__/fixtures/plugin-warn-about-foo.cjs`.

I am assuming that this would also hide those?
Kan we keep them instead?

romainmenke (commented):
Really nice find, the `linguist-generated` feature 🎉

.gitattributes
ybiquitous: Oh, great catch! I'll find a way to exclude some test files. 👀

ybiquitous: I've just pushed f48b46f83 to exclude test files. How about it?

romainmenke: Looks awesome! Thank you 🙇

ybiquitous:
I've just reverted test commits. Now we can review the actual code changes.

romainmenke (approved):
Really nice addition, because `cjs` appears before `mjs` I find myself often reviewing the wrong file :)

PR #8284, opened 2025-01-06
Document overlapping rules

jeddy3:

Although we handle this for most users via our shared config, we should document the overlaps for people who craft their own config.

### DIFF
--- a/lib/rules/at-rule-descriptor-no-unknown/README.md
+++ b/lib/rules/at-rule-descriptor-no-unknown/README.md
@@ -15,6 +15,8 @@ This rule considers descriptors defined in the CSS Specifications, up to and inc
 
 You can filter the [CSSTree Syntax Reference](https://csstree.github.io/docs/syntax/) to find out what descriptors are known for an at-rule.
 
+This rule is only appropriate for CSS. You should not turn it on for CSS-like languages, such as SCSS or Less.
+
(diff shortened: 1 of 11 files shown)


lib/rules/at-rule-descriptor-no-unknown/README.md
```diff
@@ -15,6 +15,8 @@ This rule considers descriptors defined in the CSS Specifications, up to and inc
 
 You can filter the [CSSTree Syntax Reference](https://csstree.github.io/docs/syntax/) to find out what descriptors are known for an at-rule.
 
+This rule is only appropriate for CSS. You should not turn it on for CSS-like languages, such as SCSS or Less.
```
ybiquitous, line 18: [suggestion] Do you think marking the paragraph as a warning `[!WARNING]` makes sense?

ybiquitous (approved):
Thank you, LGTM 👍 
I've left a minor suggestion, but it's not a blocker.

lib/rules/at-rule-descriptor-no-unknown/README.md
jeddy3: Let's leave it flat and informational because there are other bits in the extended description, like overlaps or what the rule ignores, that could be considered warnings.

Issue #8134, opened 2024-11-20
Fix `named-grid-areas-no-invalid` false positives for mix of tabs and spaces


EyedPeas:
### What minimal example or steps are needed to reproduce the bug?

``` css
.test-grid {
	grid-template:
		"a a a"     /* 3 whitspaces */
		"b b	b"; /* 2 whitespaces 1 tab */
}
```

### What minimal configuration is needed to reproduce the bug?

```json
"rules": {
  "named-grid-areas-no-invalid": true
}
```

### How did you run Stylelint?

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIB0AXGBnFBaA5gJwEsoACYAHQDsyV8is0BbABwBsBDNeS6nkN4-mzIhuKXgCNi46tJABuSgF9uFEABoQAMwIsYAOTYM4iGAA9DrGEjDp068BAracCEOQrFiws2gpR0wvDEANqUnp7CmACeuiwEFNiQTgQ4WJhsfmx4UMJhxAC6SvZJzgBiEHgMHK4AVuiO9rBMdojuESDRsfFiCF4g7GiYwmp5kSgxMHEJWCUpaSgZUFk5vcIDGD1FikA)

### Which Stylelint-related dependencies are you using?

```json
{		
  "stylelint": "^15.10.3",
  "stylelint-config-standard-scss": "^11.0.0"
}

```

### What did you expect to happen?

No error "Expected same number of cell tokens in each string  (named-grid-areas-no-invalid)"
=> ignoring tabs

### What actually happened?

Error "Expected same number of cell tokens in each string  (named-grid-areas-no-invalid)"  was reported

### Do you have a proposal to fix the bug?

ignore tabs

romainmenke:
is `\` a trash token?

No, it is not a trash token and this is indeed preprocessed.
Or rather there is double tokenization step.

The tokenizing algorithm described in https://drafts.csswg.org/css-grid/#valdef-grid-template-areas-string is expected to be applied the "normalized" value of the string and not the raw value as it appears in source.

The tokenizer algorithm will see: `"b b b"`, as if the newline was never there.

------

I was curious and implemented this algorithm in `css-tokenizer`.
It's a good fit to include it there.

But it does reveal some interesting issues.
[...]

Issue #6966, opened 2023-06-24
Drop experimental from `message` option's function support


ybiquitous:
Currently, many built-in rules support the `message` option's function feature:

https://github.com/stylelint/stylelint/blob/113888c0ddea330d9083efc6c5ff081a1b97069b/docs/user-guide/configure.md?plain=1#L114-L137

Rough calculation:

```console
$ git grep -l 'messageArgs: \[' lib/rules/*/index.js | wc -l
      92
```

And people have never complained about the feature spec, as I remember.

So, I suggest dropping *experimental* from this feature and officially publishing it.

Here're tasks:

- Add the function support to all rules except for deprecated ones.
- Fix the test:
  - https://github.com/stylelint/stylelint/blob/113888c0ddea330d9083efc6c5ff081a1b97069b/lib/rules/__tests__/index.test.js#L62-L63
- Remove the sentence from all the rule README:
  - https://github.com/stylelint/stylelint/blob/113888c0ddea330d9083efc6c5ff081a1b97069b/lib/rules/alpha-value-notation/README.md?plain=1#L14
- Remove "Experimental feature" wording from the doc:
  - https://github.com/stylelint/stylelint/blob/113888c0ddea330d9083efc6c5ff081a1b97069b/docs/user-guide/configure.md?plain=1#L114

jeddy3:
SGTM.

mattxwang:
I'm doing this work now! I have a handful of cases that I'm not exactly sure how to proceed:

- for rules like `block-no-empty`, there is no set of arguments to pass to the message (it doesn't reference any elements). Should I:
    - make the error message a parameterless (constant) function, and then pass an empty `[]` into `messageArgs`? The pro here is that we consistently use the same method to create custom messages, but it feels a bit strange/potentially inefficient.
    - not add `messageArgs`?
    - something else?
- for rules like `color-named`, the `expected` and `rejected` functions have different numbers of arguments. Should I:
    - just have different input lengths in `messageArgs`
    - create a falsy parameter so that `messageArgs` always has the same number of elements (or standardize the functions to throw away an argument, etc.)
    - something else?

In #7194, I have included demos of:

- making no-argument message functions constant (see: `block-no-empty`)
- proceeding to have variable-length functions when the expected/rejected disagree with each other (see: `color-named`)

I may also be missing some context (I've read through #4117, #6312, #6589 but may have missed something). Let me know what we think!

mattxwang:
[...]

PR #8285, opened 2025-01-06
Add `messageArgs` to `declaration-property-value-no-unknown`, `font-family-name-quotes`, `font-family-no-duplicate-names`, `function-calc-no-unspaced-operator`, `import-notation` and `selector-attribute-quotes`

Mouvedia:

#4117
#6966

1.
~~`function-url-quotes` didn't pass `messageArgs` to `report` hence the argument replacement is a refactor.~~
see https://github.com/stylelint/stylelint/pull/8285#discussion_r1905542680

2.
I could add `messageArgs` to `named-grid-areas-no-invalid` but I consider it not future-proof.
i.e. Ill postpone until #8134 has been fixed

### DIFF
--- a/lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
+++ b/lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
@@ -136,20 +136,17 @@ const rule = (primary, secondaryOptions) => {
 				const lowercasedValue = node.value.toLowerCase();
 				let fix;
 				let message;
-				let messageArgs;
 
 				if (Array.isArray(keywords)) {
 					if (!keywords.includes(lowercasedValue)) return;
 
-					message = messages.rejected;
-					messageArgs = [prop, value];
+					message = messages.rejected(prop, value);
 				} else {
 					const expectedKeyword = keywords[lowercasedValue];
 
 					if (!expectedKeyword) return;
 
-					message = messages.expected;
-					messageArgs = [value, expectedKeyword];
+					message = messages.expected(value, expectedKeyword);
 					fix = () => {
 						node.value = expectedKeyword;
 						setDeclarationValue(decl, parsedValue.toString());
@@ -161,7 +158,6 @@ const rule = (primary, secondaryOptions) => {
 
 				report({
 					message,
-					messageArgs,
 					node: decl,
 					result,
 					ruleName,
--- a/lib/rules/function-url-quotes/index.mjs
+++ b/lib/rules/function-url-quotes/index.mjs
@@ -12,8 +12,8 @@ import validateOptions from '../../utils/validateOptions.mjs';
 const ruleName = 'function-url-quotes';
 
 const messages = ruleMessages(ruleName, {
-	expected: (functionName) => `Expected quotes around "${functionName}" function argument`,
-	rejected: (functionName) => `Unexpected quotes around "${functionName}" function argument`,
+	expected: `Expected quotes around url function argument`,
+	rejected: `Unexpected quotes around url function argument`,
 });
 
 const meta = {
@@ -95,10 +95,14 @@ const rule = (primary, secondaryOptions) => {
 
 				if (object) {
 					const { messageType, ...rest } = object;
-					const message = messages[messageType](funcNode.value.toLowerCase());
+					const message = messages[messageType];
+					const {
+						nodes: [string = { value: '' }],
+					} = funcNode;
+					const messageArgs = [string.value];
 					const fix = getFix(node, funcNode, messageType, parsedValue);
 
-					report({ ...rest, node, fix, message, result, ruleName });
+					report({ ...rest, node, fix, message, messageArgs, result, ruleName });
 				}
 			});
 		}
--- a/lib/rules/function-url-quotes/index.mjs
+++ b/lib/rules/function-url-quotes/index.mjs
@@ -12,8 +12,8 @@ import validateOptions from '../../utils/validateOptions.mjs';
 const ruleName = 'function-url-quotes';
 
 const messages = ruleMessages(ruleName, {
-	expected: `Expected quotes around url function argument`,
-	rejected: `Unexpected quotes around url function argument`,
+	expected: (functionName) => `Expected quotes around "${functionName}" function argument`,
+	rejected: (functionName) => `Unexpected quotes around "${functionName}" function argument`,
 });
 
 const meta = {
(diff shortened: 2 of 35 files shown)


lib/rules/no-duplicate-at-import-rules/index.mjs
```diff
@@ -9,7 +9,7 @@ import validateOptions from '../../utils/validateOptions.mjs';
 const ruleName = 'no-duplicate-at-import-rules';
 
 const messages = ruleMessages(ruleName, {
-	rejected: (atImport) => `Unexpected duplicate @import rule ${atImport}`,
+	rejected: (url) => `Unexpected duplicate @import url "${url}"`,
```
Mouvedia, line 12: @jeddy3 the message doesn't match the rule name anymore; do you have a suggestion?

ybiquitous:
@Mouvedia Can you take the #8282 change into this branch? It can suppress `.cjs` file diffs.

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -161,7 +158,6 @@ const rule = (primary, secondaryOptions) => {
 
 				report({
 					message,
-					messageArgs,
```
ybiquitous, line 164: [question] Why doesn't this rule use `messageArgs`?

lib/rules/function-url-quotes/index.mjs
```diff
@@ -12,8 +12,8 @@ import validateOptions from '../../utils/validateOptions.mjs';
 const ruleName = 'function-url-quotes';
 
 const messages = ruleMessages(ruleName, {
-	expected: (functionName) => `Expected quotes around "${functionName}" function argument`,
-	rejected: (functionName) => `Unexpected quotes around "${functionName}" function argument`,
+	expected: `Expected quotes around url function argument`,
+	rejected: `Unexpected quotes around url function argument`,
```
ybiquitous, line 16: [question] I'm worried that this change may break any customized message functions:
https://stylelint.io/user-guide/configure/#message

lib/rules/no-duplicate-at-import-rules/index.mjs
jeddy3: How about:

```
Unexpected duplicate @import with URL "${url}"
```

Mouvedia:
```suggestion
	rejected: (url) => `Unexpected duplicate @import with URL "${url}"`,
```

lib/rules/no-duplicate-at-import-rules/index.cjs
```diff
@@ -12,7 +12,7 @@ const validateOptions = require('../../utils/validateOptions.cjs');
 const ruleName = 'no-duplicate-at-import-rules';
 
 const messages = ruleMessages(ruleName, {
-	rejected: (atImport) => `Unexpected duplicate @import rule ${atImport}`,
+	rejected: (url) => `Unexpected duplicate @import url "${url}"`,
```
Mouvedia, line 15:
```suggestion
	rejected: (url) => `Unexpected duplicate @import with URL "${url}"`,
```

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
ybiquitous: Understood. But if we need to consider the differentiation, we should keep this rule's `messageArgs` as-is in this PR.

In addition, the "Experimental feature" expression should also be kept because this feature is still unstable. For now, we cannot ensure the feature stability.

ybiquitous:
@Mouvedia I'm afraid, but as I have told you multiple times, it seems you often include more things than a PR title says. I think this PR contains more than what the title "Add `messageArgs` to ..." says, such as removing `messageArgs` or changing some rules' messages.

In general, when a PR has many things, it becomes hard to track the commit history or why a change was performed.

lib/rules/function-url-quotes/index.mjs
Mouvedia: > should we include `funcNode.value` in the messages instead?

What matters for the users is the argument of the url function, which is what I have added as an argument.
I don't see what the value of passing `'url'` or `'URL'` would bring.

I didn't use for our `messages` because the URL may be very long.
i.e. it would be inconvenient

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
Mouvedia: > the "Experimental feature" expression should also be kept because this feature is still unstable

OK.

Mouvedia:
> changing some rules' messages.

Ill remove the consistency message enhancements.

> such as removing `messageArgs`

Ill keep the removal of `messageArgs` of `declaration-property-value-keyword-no-deprecated` because the goal of this PR is to pave the way for the type PR. Before the release of 16.3.0 it's a refactor and doesn't impact anything.

i.e.
No rules will be modified in the upcoming type PR and all rules which should have `messageArgs` do have it.

lib/rules/function-url-quotes/index.mjs
ybiquitous: > I don't see what the value of passing `'url'` or `'URL'` would bring.

I still think including a part of an actual source (e.g. `url` or `URL`) makes sense for the message's clarity.

> I didn't use the argument for our `messages` because the URL may be very long.

If so, we shouldn't specify a `url()` argument to `messageArgs`, should we?

Additionally, I think the changed `messages` implementation not to use functions is very confusing:

```js
	expected: `Expected quotes around url function argument`,
	rejected: `Unexpected quotes around url function argument`,
```

Because:
- At first glance, it appears unsupported of the message functions.
- It is hard to understand that the first argument passed to message functions is a `url()` argument.

This change is clearly out of the scope of what this PR title says. We should discuss this in another PR or issue only about the `function-url-quotes` messages.

ybiquitous (changes requested):
> Ill keep the removal of declaration-property-value-keyword-no-deprecated's messageArgs because the goal of this PR is to pave the way for the type PR. Before the release of 16.13.0 it's a refactor and doesn't impact anything.

Okay 👍🏼 

> I do when it's tightly related and considered a minor enhancement/refactor.

Please keep in mind my comment above when you do so. You often create a pretty big PR.

ybiquitous (approved):
Thanks for addressing my comments. LGTM 👍🏼

Issue #8291, opened 2025-01-09
Add `exceptWithoutPropertyFallback: []` to `function-allowed-list`


AlexanderShushunov:
### What is the problem you're trying to solve?

The `function-allowed-list` rule is a powerful tool for ensuring that only CSS functions supported by our target browsers are used. For example, if we aim to support Chrome 47, we might want to block the use of the `min()` function, which is supported starting from Chrome 79.

However, in many cases, developers can provide fallbacks to ensure compatibility with older browsers while enhancing the user experience for modern ones. For instance:

```css
  width: 200px; /* Fallback for older browsers */
  width: min(50%, 300px); /* Advanced functionality for modern browsers */
```

Currently, `function-allowed-list` prohibits this usage even when a fallback is present. This limits developers from leveraging modern CSS capabilities responsibly.

### What solution would you like to see?

Allow the use of restricted functions if a valid fallback is provided

romainmenke:
Hi @AlexanderShushunov,

Thank you for opening an issue about this.

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYAdAdilB3AllAC4AW8KATAAxUAOAHgNwoD0AVCgGJIA2PARkjABrFADMIAJxQQesaf0kRcAZxiSVKNiyw4CxMigC2+TAAoArFQCkAGhQBmGgwCUzdigCCUAG5JMYDBQ4gCuAUT4EJi8+EQAnuJSxtDq2IrKahpaOpgAvlhY-Gi6eISk5NR0TKwc3HyCIonSsvIo6arqmtol+uXGppY29k5VbjUoKvE8MDymRAC0UPgqSPzT87OYMKHhkZjzvDzKQRvLRCjz8yi0SgDmkjAqKvg+2zCYJP6BRu-n3XlYEC2EBifDTABySB+CBAMHoUNo0wAdGAnkDwFFQbcYRhsCh0LD6ER3lAVATyABtEo4AmTOLTTYLSCYLHzSb+KBISRQAklAC6thKBMkIWmZIQxTxNJBYTAESiBz4xygp0m5JQFL5JXyeXRzKxnCkRiQRBhACsVFF0bBaCocUKQHSGXN1QSeCbHkQCYK8bSpjM5vN9fhbmyiByuTyJW6PWqQFhciBckA)

Thank you also for providing a PR with a proposed fix.
We normally first discuss the issue and then mark it as "ready to implement".
Let's also hear from others about this :)

[...]

PR #8292, opened 2025-01-09
`function-allowed-list` allows a function if a fallback is provided

AlexanderShushunov:
#### Which issue, if any, is this issue related to?

Closes https://github.com/stylelint/stylelint/issues/8291 .

#### Is there anything in the PR that needs further explanation?

No, it's self-explanatory.

#### Summary of Changes

This PR enhances the functionality of the `function-allowed-list` rule to support the use of restricted functions when valid fallbacks are provided. For example:

```css
  width: 200px; /* Fallback for older browsers */
  width: min(50%, 300px); /* Advanced functionality for modern browsers */
```

Previously, such usage was disallowed, even with a proper fallback. This update ensures that developers can leverage modern CSS features responsibly while maintaining compatibility with older browsers.

### DIFF
--- a/lib/rules/function-allowed-list/README.md
+++ b/lib/rules/function-allowed-list/README.md
@@ -73,3 +73,11 @@ a {
     -moz-linear-gradient(45deg, blue, red);
 }
 ```
+
+<!-- prettier-ignore -->
+```css
+a {
+  color: darkseagreen;
+  color: hsla(170, 50%, 45%, 1);
+}
+```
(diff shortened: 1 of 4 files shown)


Mouvedia:
Hello @AlexanderShushunov

We usually wait for the label _status: ready to implement_ to be set on an issue to start working on the corresponding PR.
Let's wait for a resolution in the issue.

AlexanderShushunov:
@AlexanderShushunov, do not worry about the PR. I have never worked with CSS processing before. This code helped me understand the limits of the technology's applicability. It’s okay if I need to rewrite or delete it. I’ve already learned something new and worked with a good piece of code ;)

PR #8239, opened 2024-12-28
Fix `*-no-vendor-prefix` message ambiguity

Mouvedia:
> Which issue, if any, is this issue related to?

followup of #8236

> Is there anything in the PR that needs further explanation?

1.
I didn't add the third argument to all rules.
96 rules would need it.
This PR enables the possibility to eventually do it.
i.e. for now the default is sufficient

2.
Messages of the form
```
(actual, expected) => `Expected "${actual}" to be "${expected}"`
```
which key weren't `expected` are fixed to match our conventions.

3.
https://github.com/stylelint/stylelint/pull/8236#issuecomment-2564142598

PR #8294, opened 2025-01-10
Add core rules' `messages` types

Mouvedia:
> Which PR, if any, is this PR related to?

#8239

1.
I expect the review to be long and tedious.
I am sorry :)

2.
This PR won't touch the rules.
i.e. the erroneous messages and/or badly named parameters can be corrected in followup PRs

3.
We may add rejected message types if we feel like it later on.
e.g.
```
{
  name: [name: string]
  property: [property: string]
  selector: [selector: string]
  // etc.
}
```

4.
The medium term goal would be to flow the type to [`message`](https://stylelint.io/user-guide/configure/#message) so that it may eventually be exploited by your IDE when using js/mjs/cjs configs.

It can already be achieved through other means.
e.g. https://github.com/search?q=repo%3Astylelint-types%2Fstylelint-define-config+message%3F&type=code

5.
In the short term, it has surfaced many inconsistencies/oversights that have been or will be corrected.


types/stylelint/index.d.ts
```diff
@@ -394,35 +465,51 @@ declare namespace stylelint {
...
 		'declaration-property-value-disallowed-list': CoreRule<
-			Record<string, OneOrMany<StringOrRegex>>
+			Record<string, OneOrMany<StringOrRegex>>,
+			{},
+			RejectedMessage<[property: string, value: string]>
 		>;
 		'declaration-property-value-keyword-no-deprecated': CoreRule<
 			true,
-			{ ignoreKeywords: OneOrMany<StringOrRegex> },
-			AutofixMessage & { rejected: (property: string, keyword: string) => string }
```
Mouvedia: see https://github.com/stylelint/stylelint/pull/8285#discussion_r1905170837

types/stylelint/index.d.ts
```diff
@@ -315,28 +351,44 @@ declare namespace stylelint {
...
+			ExpectedMessage<[hex: string]> & RejectedMessage<[hex: string]>
+		>;
+		'color-hex-length': CoreRule<'short' | 'long', {}, ExpectedMessage<'autofix'>>;
 		'color-named': CoreRule<
 			'never' | 'always-where-possible',
-			{ ignoreProperties: OneOrMany<StringOrRegex>; ignore: OneOrMany<'inside-function'> },
-			{
-				expected: (expected: string, actual: string) => string;
-				rejected: (keyword: string) => string;
-			}
```
Mouvedia, line 475: `messageArgs` wasn't added to rules that require for the user to check the arity.
e.g.
```
message: (a, b) => {
  if (typeof b === 'string') {
    /* … */
  } else { /* … */ }
}
```

It may be added so that #6966 can be resolved but that's out of scope for this PR.
The main rationale behind the type's addition is that it can eventually flow to `message`.
i.e. not for `messages` itself (for that we have the `Messages` default)

Since `color-named` has the arguments backward for `AutofixMessage` that's a blessing in disguise.

Mouvedia:
I guess this one is too big to review.
@ybiquitous what about splitting it in 3?
i.e. 3 PRs

ybiquitous:
> I guess this one is too big to review.
> @ybiquitous what about splitting it in 3?
> i.e. 3 PRs

I'm okay with 1 PR.

For another point, this PR is a refactoring of the types, right? If so, can you please update the title to clarify the intention?

types/stylelint/index.d.ts
```diff
@@ -275,26 +275,62 @@ declare namespace stylelint {
...
+		T,
+		'expected',
+		{
+			autofix: [actual: string, expected: string];
+			pattern: [input: string, pattern: StringOrRegex];
+			max: [object: string, maximum: number];
+		}
+	>;
+
+	type RejectedMessage<T> = Message<T, 'rejected', {}>;
```
ybiquitous, line 294: [suggestion] IMO, these new types are too tricky for me, and they seem so hard to maintain in the future. Like the following examples, allowing some duplications makes the code more straightforward.

```typescript
type AutofixMessages = {
	expected: (actual: string, expected: string) => string;
};
type PatternMessages = {
	expected: (input: string, pattern: StringOrRegex) => string;
};
type MaximumMessages = {
	expected: (target: string, maximum: number) => string;
};
type RejectedMessages<T extends unknown[]> = {
	rejected: (...args: T) => string;
};
```

ybiquitous: Also, I think the `RejectedMessages` name is more accurate than `RejectedMessage` because they are objects that can have multiple message functions.

Mouvedia: @ybiquitous thanks for the review.
Sorry for the delay, Ill try to come up with something simpler when Ill have time.

This PR is not blocking anything.
I am planning to eventually generate the documentation automatically but that's not urgent.

Mouvedia: I have refactored the types.
It's now easier to maintain if you are not a TS expert.

types/stylelint/index.d.ts
```diff
@@ -620,34 +721,69 @@ declare namespace stylelint {
...
 		'media-feature-name-no-vendor-prefix': CoreRule<true>;
-		'media-feature-name-unit-allowed-list': CoreRule<Record<string, OneOrMany<string>>>;
-		'media-feature-name-value-allowed-list': CoreRule<Record<string, OneOrMany<StringOrRegex>>>;
-		'media-feature-name-value-no-unknown': CoreRule<true>;
-		'media-feature-range-notation': CoreRule<'prefix' | 'context'>;
-		'media-query-no-invalid': CoreRule<true, { ignoreFunctions: OneOrMany<StringOrRegex> }>;
+		'media-feature-name-unit-allowed-list': CoreRule<
+			Record<string, OneOrMany<string>>,
+			{},
+			// FIXME should be [name: string, unit: string]
```
Mouvedia:
```suggestion
```

Mouvedia:
@ybiquitous I have simplified the types, please take a look when you will have the time.

types/stylelint/index.d.ts
```diff
@@ -413,13 +413,24 @@ declare namespace stylelint {
 		[key in `expected${string}` | `rejected${string}`]: RuleMessage;
 	};
 
-	type AutofixMessage = { expected: (actual: string, expected: string) => string };
+	type ExpectedMessage<T extends unknown[]> = { expected: (...args: T) => string };
+	type RejectedMessage<T extends unknown[]> = { rejected: (...args: T) => string };
+	type AutofixMessage = ExpectedMessage<[actual: string, expected: string]>;
+	type PatternMessage = ExpectedMessage<[input: string, pattern: StringOrRegex]>;
+	type MaximumMessage = ExpectedMessage<[object: string, maximum: number]>;
```
ybiquitous: [suggestion] `object` looks like a kind of type. Instead, what about `target`?

```suggestion
	type MaximumMessage = ExpectedMessage<[target: string, maximum: number]>;
```

Mouvedia: b085634

ybiquitous (approved):
Thanks, LGTM 👍🏼

types/stylelint/index.d.ts
```diff
@@ -620,34 +725,68 @@ declare namespace stylelint {
...
 		'media-feature-name-no-vendor-prefix': CoreRule<true>;
-		'media-feature-name-unit-allowed-list': CoreRule<Record<string, OneOrMany<string>>>;
-		'media-feature-name-value-allowed-list': CoreRule<Record<string, OneOrMany<StringOrRegex>>>;
-		'media-feature-name-value-no-unknown': CoreRule<true>;
-		'media-feature-range-notation': CoreRule<'prefix' | 'context'>;
-		'media-query-no-invalid': CoreRule<true, { ignoreFunctions: OneOrMany<StringOrRegex> }>;
+		'media-feature-name-unit-allowed-list': CoreRule<
+			Record<string, OneOrMany<string>>,
+			{},
+			RejectedMessage<[unit: string, name: string]>
```
Mouvedia, line 750: note

should be `[name: string, unit: string]` but that would be a breaking change

Issue #8290, opened 2025-01-08
Document sub sections in changelog


jeddy3:
### Describe the documentation issue

We're inconsistent with the keepachange format, which we normally lean on for changelog direction.

Ref: https://github.com/stylelint/stylelint/pull/8289

### What solution would you like to see?

Adopt sub sections as they improve readability. This will involve updating our changesets script.

We'll also want to retroactively apply this to the changelog to be consistent. This can be done manually, but it may be easier to rustle up a codemod as a remark plugin to do the heavy lifting; it may well just be the case of:

1. inserting a heading when the prefix changes
2. removing the prefixes
3. patching up anything amiss

Mouvedia:
@jeddy3 I made a [putout](https://github.com/coderaiser/putout) plugin just for fun; it's quite easy to grasp.
The blocker for this issue is not converting the CHANGELOG.md, it's updating our scripts.
e.g. [`rewrite-changelog.mjs`](https://github.com/stylelint/stylelint/blob/main/.changeset/rewrite-changelog.mjs)

jeddy3:
Thank you for rustling up a codemod.

Mouvedia:
You can commit on the draft PR and alter the scripts if you think it's worth pursuing.
I was just curious to try putout.

ybiquitous:
Do you think it makes sense to adopt the [changesets](https://github.com/changesets/changesets) format? Like [this](
https://github.com/changesets/changesets/blob/a3acd12146e888abb104a6a41399be8bba0f6169/packages/cli/CHANGELOG.md).

Mouvedia:
Personally I prefer keepachangelog sub-sections because it's more descriptive for the consumers, but if we were to use `Major Changes`/`Minor Changes`/`Patch Changes` I think we should keep our message formats—i.e. changing the list items would be too disruptive.

ybiquitous:
Continuing to keep the *keepachangelog* format makes sense. We have already documented it here:

https://github.com/stylelint/stylelint/blob/f5055feb3613eaf0f8a790d553112f28fa78fb3c/docs/maintainer-guide/issues.md?plain=1#L38

It doesn't seem that the *changesets* package supports the *keepachangelog* format, so we need some script to rewrite the generated changelog by *changesets*.

[...]

PR #8295, opened 2025-01-11
Update CHANGELOG

Mouvedia:

#8290

generated using this putout plugin

```js
const report = ({ type }) => {
	switch (type) {
		case 'move':
			return 'qux';
			break;
		case 'remove':
			return 'bar';
			break;
	}
};

export default {
	name: 'foo',
	fix,
	traverse,
	report,
};

function fix({ index, node, listItems, type }, tree) {
	switch (type) {
		case 'move':
			// add sud-heading
			tree.children.splice(index, 0, node);

			// remove prefixes
			listItems.forEach(({ children: [{ children: [line] }	] }) => {
					line.value = line.value.split(': ')[1];
			});

			// add corresponding list
			tree.children.splice(index + 1, 0, {
				type: 'list',
				spread: false,
				children: listItems,
			});
			break;
		case 'remove':
			tree.children.splice(index, 1);
			break;
	}
}

function traverse(tree, { push }) {
	let delta = 0;

	for (const [index, node] of tree.children.entries()) {
		if (node.type === 'list') {
			const types = new Set();

			node.children.forEach((listItem) => {
				const line = listItem.children[0].children[0];
				const [type, tail] = line.value.split(': ');

				if (!types.has(type)) {
					types.add(type);
					push({
						node: {
							type: 'heading',
							depth: 3,
							children: [{ type: 'text', value: type }],
						},
						index: index + delta,
						listItems: node.children.filter(({ children: [paragraph] }) => {
							const [prefix] = paragraph.children[0].value.split(': ');

							return prefix === type;
						}),
						type: 'move',
					});
					delta = delta + 2;
				}
			});
			push({
				index: index + delta,
				type: 'remove',
			});
			delta--;
		}
	}
}
```

```diff
@@ -1495,7 +1495,7 @@

 ### Fixed

-- "fatal[#3915](https://github.com/stylelint/stylelint/pull/3915)).
+- "fatal: Not a git repository" error ([#3915](https://github.com/stylelint/stylelint/pull/3915)).
 - unintended increase in package size ([#3915](https://github.com/stylelint/stylelint/pull/3915)).

 ## 9.10.0 - 2019-01-16
@@ -1859,18 +1859,14 @@

 This release is accompanied by:

-### A new
-
-- [semantic version policy](docs/about/semantic-versioning.md). The use of the tilde (`~`) in `package.json` is now recommended, e.g. `"stylelint": "~8.0.0"`, to guarantee the results of your builds ([#1865](https://github.com/stylelint/stylelint/issues/1865)).
-- [VISION document](docs/about/vision.md), complemented by ([#2704](https://github.com/stylelint/stylelint/pull/2704)):
+- A new [semantic version policy](docs/about/semantic-versioning.md). The use of the tilde (`~`) in `package.json` is now recommended, e.g. `"stylelint": "~8.0.0"`, to guarantee the results of your builds ([#1865](https://github.com/stylelint/stylelint/issues/1865)).
+- A new [VISION document](docs/about/vision.md), complemented by ([#2704](https://github.com/stylelint/stylelint/pull/2704)):
   - The restructuring of the [list of rules](docs/user-guide/rules.md) into three groups:
     - Possible errors
     - Limit language features
     - Stylistic issues
   - The release of a new sharable config, [`stylelint-config-recommended`](https://github.com/stylelint/stylelint-config-recommended). This config only turns on the possible error rules. [`stylelint-config-standard`](https://github.com/stylelint/stylelint-config-standard) now builds on top of the recommended config by turning on over 60 additional stylistic rules.

-Changes:
-
 ### Removed

 - the 21 rules deprecated in [`7.8.0`](https://github.com/stylelint/stylelint/releases/tag/7.8.0) & [`7.12.0`](https://github.com/stylelint/stylelint/releases/tag/7.12.0) ([#2422](https://github.com/stylelint/stylelint/issues/2422) & [#2693](https://github.com/stylelint/stylelint/issues/2693)).
@@ -2737,11 +2733,8 @@
 - `rule-no-shorthand-property-overrides` rule. Use the new `declaration-block-no-shorthand-property-overrides` rule instead.
 - `rule-properties-order` rule. Use the new `declaration-block-properties-order` rule instead.
 - `rule-trailing-semicolon` rule. Use the new `declaration-block-trailing-semicolon` rule instead.
-- `"always"` option for `font-weight-notation`. Use the new `always-where-possible` option instead.
-
-### Removed
-
 - `true` option for `emptyLineBefore` when using property groups in `rule-properties-order`. Use the new `"always"` or `"never"` option instead.
+- `"always"` option for `font-weight-notation`. Use the new `always-where-possible` option instead.

 ### Added

@@ -2774,11 +2767,8 @@
 - `rule-no-shorthand-property-overrides` rule. Use the new `declaration-block-no-shorthand-property-overrides` rule instead.
 - `rule-properties-order` rule. Use the new `declaration-block-properties-order` rule instead.
 - `rule-trailing-semicolon` rule. Use the new `declaration-block-trailing-semicolon` rule instead.
-- `"always"` option for `font-weight-notation`. Use the new `always-where-possible` option instead.
-
-### Deprecated
-
 - `true` option for `emptyLineBefore` when using property groups in `rule-properties-order`. Use the new `"always"` or `"never"` option instead.
+- `"always"` option for `font-weight-notation`. Use the new `always-where-possible` option instead.

 ### Added
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,829 +2,1172 @@
 
 ## 16.12.0 - 2024-12-15
 
-- Added: `selector-pseudo-class-allowed-list` now checks `@page` pseudo-classes ([#8176](https://github.com/stylelint/stylelint/pull/8176)) ([@Mouvedia](https://github.com/Mouvedia)).
-- Added: `selector-pseudo-class-disallowed-list` now checks `@page` pseudo-classes ([#8171](https://github.com/stylelint/stylelint/pull/8171)) ([@Mouvedia](https://github.com/Mouvedia)).
-- Fixed: `at-rule-property-required-list` message to use "descriptor" for accuracy ([#8186](https://github.com/stylelint/stylelint/pull/8186)) ([@ybiquitous](https://github.com/ybiquitous)).
-- Fixed: `custom-property-no-missing-var-function` false positives for `container-name` ([#8157](https://github.com/stylelint/stylelint/pull/8157)) ([@Mouvedia](https://github.com/Mouvedia)).
-- Fixed: `custom-property-no-missing-var-function` false positives for custom properties passed to `running()` ([#8172](https://github.com/stylelint/stylelint/pull/8172)) ([@Mouvedia](https://github.com/Mouvedia)).
-- Fixed: `function-no-unknown` false positives for `running()` ([#8172](https://github.com/stylelint/stylelint/pull/8172)) ([@Mouvedia](https://github.com/Mouvedia)).
-- Fixed: `selector-pseudo-class-no-unknown` false positives for `:open` ([#8184](https://github.com/stylelint/stylelint/pull/8184)) ([@lukewarlow](https://github.com/lukewarlow)).
-- Fixed: `selector-pseudo-class-no-unknown` false positives for `:recto`, `:verso` and `:nth()` ([#8170](https://github.com/stylelint/stylelint/pull/8170)) ([@Mouvedia](https://github.com/Mouvedia)).
-- Fixed: `selector-pseudo-class-no-unknown` false positives for some `moz-*` vendor-prefixed pseudo-classes ([#8188](https://github.com/stylelint/stylelint/pull/8188)) ([@Mouvedia](https://github.com/Mouvedia)).
-- Fixed: `selector-pseudo-element-no-unknown` false positives for `::details-content` ([#8174](https://github.com/stylelint/stylelint/pull/8174)) ([@lukewarlow](https://github.com/lukewarlow)).
-- Fixed: `selector-type-no-unknown` false positives for idents in functional pseudo-classes ([#8191](https://github.com/stylelint/stylelint/pull/8191)) ([@elskhn](https://github.com/elskhn)).
-- Fixed: `value-keyword-case` false negatives ([#8158](https://github.com/stylelint/stylelint/pull/8158)) ([@Mouvedia](https://github.com/Mouvedia)).
-- Fixed: `value-keyword-case` false positives for vendor-prefixed system colors ([#8146](https://github.com/stylelint/stylelint/pull/8146)) ([@Mouvedia](https://github.com/Mouvedia)).
+### Added
+
+- `selector-pseudo-class-allowed-list` now checks `@page` pseudo-classes ([#8176](https://github.com/stylelint/stylelint/pull/8176)) ([@Mouvedia](https://github.com/Mouvedia)).
+- `selector-pseudo-class-disallowed-list` now checks `@page` pseudo-classes ([#8171](https://github.com/stylelint/stylelint/pull/8171)) ([@Mouvedia](https://github.com/Mouvedia)).
+
+### Fixed
+
+- `at-rule-property-required-list` message to use "descriptor" for accuracy ([#8186](https://github.com/stylelint/stylelint/pull/8186)) ([@ybiquitous](https://github.com/ybiquitous)).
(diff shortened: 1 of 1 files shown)


scripts/update-changelog.mjs
Mouvedia, line 1: @ybiquitous can you try it out?

I have never used `changelog-stylelint.cjs` before; is it still needed?

ybiquitous: The `changeset add` command internally uses `changelog-stylelint.cjs` 😃 

https://github.com/stylelint/stylelint/blob/2c639ff31ff4c04b63736edd0ebde601bb108e9a/.changeset/config.json?plain=1#L4

Mouvedia: Well then, I just need to remove the changesets from the `.changetset` folder at the end of the process and I should be done here. I don't have the time to finish it right now though.

The only bug that I detected is regarding nested lists in release descriptions, which currently only impacts version 8.0.0. That would require to either drop the `keep-a-changelog` dependency or open an issue/PR on their repository. In both cases, it's not a big deal if we are using `git add -p`.

scripts/update-changelog.mjs
```diff
@@ -0,0 +1,82 @@
...
+const execAsync = util.promisify(exec);
+
+let releaseType = 'patch';
+
+for (const {
+	releases: [{ type }],
+	summary,
+	id,
+} of changesets) {
+	const [change, item] = summary.split(': ');
```
Mouvedia, line 23: todo

use regex

scripts/update-changelog.mjs
```diff
@@ -0,0 +1,82 @@
+import { exec } from 'node:child_process';
+import fs from 'node:fs';
+import process from 'node:process';
+import util from 'node:util';
+
+import { Release, parser } from 'keep-a-changelog';
+import increment from 'semver/functions/inc.js';
+import read from '@changesets/read';
+
+import pkg from '../package.json' assert { type: 'json' };
```
Mouvedia, line 10: todo

use require

scripts/update-changelog.mjs
```diff
@@ -0,0 +1,83 @@
...
+	const [issue] = message.match(/#[0-9]{4,}/);
+	const PR = `https://github.com/stylelint/stylelint/pull/${issue.slice(1)}`;
+	const account = `https://github.com/${user}`;
+
+	groups[change.toLowerCase()].push(`${item} ([${issue}](${PR})) ([@${user}](${account})).`);
+
+	if (releaseType !== 'major' && type !== 'patch') releaseType = type;
+}
+
+class CustomRelease extends Release {
```
Mouvedia, line 45: note

required because the default order or the subheaders is different from ours
see https://github.com/oscarotero/keep-a-changelog/blob/master/src/Release.ts#L24-L31

scripts/update-changelog.mjs
```diff
@@ -0,0 +1,83 @@
...
+		super(...args);
+
+		this.changes = new Map([
+			['security', []],
+			['removed', []],
+			['changed', []],
+			['deprecated', []],
+			['added', []],
+			['fixed', []],
+		]);
```
Mouvedia, line 56: note

`new Map(Object.entries(structuredClone(changes)));` would be less performant

ybiquitous:
@Mouvedia Can we close this old PR?

Mouvedia:
> @Mouvedia Can we close this old PR?

Ill close it and come back to it later.

PR #8297, opened 2025-01-12
Add `languageOptions` configuration property

### LINKED ISSUE #8154: Add `languageOptions` configuration property

There are two connected and upcoming problems:

- duplicated syntax configuration 
- performance cost of forking CSSTree more than once

The [`declaration-property-value-no-unknown`](https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown/#propertiessyntax--property-syntax-) rule currently has:

- two options to extend the syntax
- [a PR](https://github.com/stylelint/stylelint/pull/8124) to add a `cssWideKeywords` one

The new rules in https://github.com/stylelint/stylelint/issues/8133 will add more.

Consolidate syntax configuration into a single option and fork CSSTree only once.

```json
{
  "languageOptions": {
    "syntax": {
      "atRules": {
        "example": {
          "prelude": "<custom-ident>",
          "descriptors": {
            "foo": "<number>",
            "bar": "<color>"
          }
        }
      },
      "cssWideKeywords": ["my-global-value"],
      "properties": { "top": "| <--foo()>" },
      "types": { "--foo()": "--foo( <length-percentage> )" }
    }
  },
  "rules": {
    "at-rule-descriptor-no-unknown": true,
    "at-rule-descriptor-value-no-unknown": true,
    "at-rule-prelude-no-unknown": true,
  }
}
```

We'd deprecate the `propertiesSyntax` and `typesSyntax` options in [declaration-property-value-no-unknown](https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown/#propertiessyntax--property-syntax-). I think this is fine as it's an experimental rule.

Ref: https://github.com/stylelint/stylelint/issues/8133#issuecomment-2488108205

ryo-manba:

Closes #8154 

`lexer` is now generated only when required by the rules that utilize it, effectively reducing the performance overhead associated with forking CSSTree multiple times. 

`declaration-property-value-no-unknown` rule remains unchanged to ensure compatibility with existing configurations.


jeddy3:
@ryo-manba Thank you for starting to work on this! It's an exciting addition that already [meets a request](https://github.com/stylelint/stylelint/issues/8303).

Let's remember to add links to each of the rule READMEs to the `languageOptions` configuration property within the `configure.md` page. See: https://github.com/stylelint/stylelint/issues/8303#issuecomment-2587177985

docs/user-guide/configure.md
```diff
@@ -254,6 +254,119 @@ But the following pattern would be reported as a warning:
...
+```
+
+The following rules are configured via the `languageOptions` property:
+
+- [`at-rule-descriptor-no-unknown`](../../lib/rules/at-rule-descriptor-no-unknown/README.md)
+- [`at-rule-descriptor-value-no-unknown`](../../lib/rules/at-rule-descriptor-value-no-unknown/README.md)
+- [`at-rule-prelude-no-invalid`](../../lib/rules/at-rule-prelude-no-invalid/README.md)
+<!-- TODO: Add 'declaration-property-value-no-unknown' after options are deprecated -->
+
+#### `AtRules`
```
ryo-manba, line 305: Would it be better to provide more detailed explanations for each option?

types/stylelint/index.d.ts
```diff
@@ -84,6 +86,22 @@ declare namespace stylelint {
...
 
+	type LanguageOptions = {
+		syntax?: {
+			atRules?: Record<
+				string,
+				{
+					comment?: string;
+					prelude?: string;
+					descriptors?: Record<string, string>;
+				}
```
ryo-manba, line 97: Used [csstree's patch.json](https://github.com/csstree/csstree/blob/v3.1.0/data/patch.json) as a reference.

docs/user-guide/configure.md
```diff
@@ -254,6 +254,119 @@ But the following pattern would be reported as a warning:
...
+
+- `atRules`: Define custom at-rules with specific `prelude` and `descriptors` syntax
+- `cssWideKeywords`: Extend the list of CSS-wide keywords with custom values
+- `properties`: Customize the syntax of specific properties
+- `types`: Extend or modify type definitions used in property values
+
+```json
+{
+  "languageOptions": {
+    "syntax": {
```
Mouvedia, line 273: note

As to why `languageOptions` has only one sub-property right now see

> [and tie into a Baseline set in `languageOptions`](https://github.com/stylelint/stylelint/issues/8291#issuecomment-2587132762)

i.e. it won't be the only one

lib/lintSource.mjs
```diff
@@ -111,6 +122,23 @@ export default async function lintSource(stylelint, options = {}) {
...
 }
 
+const rulesUsingLexer = [
+	'at-rule-descriptor-no-unknown',
+	'at-rule-descriptor-value-no-unknown',
+	'at-rule-prelude-no-invalid',
+
+	// TODO: Add after options deprecation
+	// 'declaration-property-value-no-unknown',
+];
```
Mouvedia, line 132: This should be discussed in the issue itself but Ill give my opinion here.
This should be a boolean on `rule.meta`.
e.g. `RuleMeta['lexer']`

For a POC I wouldn't mind but as an official feature I would prefer if we could avoid having to maintain another list buried in a random file.
The name of the boolean will have to bikeshed though.

lib/rules/at-rule-descriptor-value-no-unknown/index.mjs
```diff
@@ -36,6 +34,10 @@ const rule = (primary) => {
 
 				const { prop, value } = decl;
 
+				const lexer = result.stylelint.lexer;
+
+				if (!lexer) throw new Error('Expected a "lexer" object');
```
Mouvedia, line 39: Once we have new meta boolean field, we can check this earlier.
i.e. this check should not be done in the rules themselves

Whether the check is necessary or not is another discussion; I am just talking about the location.

jeddy3 (commented):
Thanks again for working on this substantial feature. I've asked a question that hopefully someone will be able to answer. Once that's confirmed, we can pick up the rest of the review. It's looking fantastic so far, btw!

lib/lintSource.mjs
ryo-manba: Thanks for your comment! I've added a metadata. 
[Add lexer info to metadata](https://github.com/stylelint/stylelint/pull/8297/commits/1d19d07be3cbc2d202620313847176579784d347)

I tested the performance as a precaution, but there was no significant difference between the original implementation and the metadata-based implementation.

<details>
<summary>Original implementation</summary>

benchmark-original-implementation

</details>


<details>
<summary>Metadata based implementation</summary>

benchmark-metadata-implementation

</details>

Mouvedia:
I personally think that having `RuleMeta['usesLexer']` is better than having to maintain a list of rules in some file.
i.e. 1d19d07
Before we go further—i.e. add it to the documentation—I would like some feedback regarding the name.
What I mean by that is that if we were to add several lexers the meta field type will have to be changed from `boolean` to `string`.
i.e.
```js
rule.meta = {
  /* … */
  lexer: 'foo',
};
```
i.e. if we had to change the type I wish the chosen name wouldn't have to be changed too

@jeddy3 @ybiquitous
What do you think?

lib/rules/at-rule-descriptor-value-no-unknown/index.mjs
ryo-manba: I added a getter, but I don't fully understand the significance of "early checks" in this context.

[Add getLexer](https://github.com/stylelint/stylelint/pull/8297/commits/41d9857043804ee7344adb6ea53620cee6a05951)

docs/user-guide/configure.md
jeddy3: Let's add some common examples to each option, e.g. supporting a new function for `<color>` type.

lib/rules/at-rule-descriptor-value-no-unknown/index.mjs
Mouvedia: I meant that we want to have this code once in the codebase.
i.e. if you have access to the enabled rules that have the new meta field you would run that function once or in an existing loop
```
// example
function foo(filteredRules, ruleMetadata) {
  // 1. loop on rules
  // 2. switch/case on lexers string ENUM (will only be an if since we only have one)
  // 3. throw
    // 3.1 wrong lexer
    // 3.2 missing lexer
}
```

Based on that knowledge the function should probably be called somewhere where you have these.
i.e. rules and meta
That file is probably lintPostcssResult.mjs.
e.g.
```js
/* […] */
// after line 89
checkLexers(ruleFunction.meta); // or checkLexer(ruleFunction.meta['bar']);
```
i.e. before you reach l110-123

docs/user-guide/configure.md
ryo-manba: I've added explanations to convey their connection to the rules.
[Add more examples](https://github.com/stylelint/stylelint/pull/8297/commits/57ff9a3fb45587ededf7d495e85984ff841f4b16)

docs/user-guide/configure.md
```diff
@@ -254,6 +254,186 @@ But the following pattern would be reported as a warning:
...
+      "atRules": {
+        "example": {
+          "prelude": "<custom-ident>",
+          "descriptors": {
+            "foo": "<number>",
+            "bar": "<color>"
+          }
+        }
```
jeddy3, line 321:
```suggestion
        "foo": {
          "prelude": "<custom-ident>",
          "descriptors": {
            "bar": "<number>",
            "baz": "<color>"
          }
        }
```

docs/user-guide/configure.md
```diff
@@ -254,6 +254,186 @@ But the following pattern would be reported as a warning:
...
+    }
+  }
+}
+```
+
+When `at-rule-descriptor-no-unknown` is enabled, the following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+@example custom-ident { baz: 10; }
```
jeddy3, line 332:
```suggestion
@foo custom-ident { qux: 10; }
```

docs/user-guide/configure.md
```diff
@@ -254,6 +254,186 @@ But the following pattern would be reported as a warning:
...
+```
+
+When `at-rule-descriptor-value-no-unknown` is enabled, the following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+@example custom-ident { foo: red; }
```
jeddy3, line 339:
```suggestion
@foo custom-ident { bar: red; }
```

docs/user-guide/configure.md
```diff
@@ -254,6 +254,186 @@ But the following pattern would be reported as a warning:
...
+```
+
+In both case, the following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+@example custom-ident { foo: 10; bar: red; }
```
jeddy3, line 346:
```suggestion
@foo custom-ident { bar: 10; baz: red; }
```

docs/user-guide/configure.md
```diff
@@ -254,6 +254,186 @@ But the following pattern would be reported as a warning:
...
+<!-- prettier-ignore -->
+```css
+a { color: foo; }
+```
+
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+a { color: my-global-value; }
```
jeddy3, line 375:
```suggestion
a { color: -moz-initial; }
```

docs/user-guide/configure.md
```diff
@@ -254,6 +254,186 @@ But the following pattern would be reported as a warning:
...
+#### `properties`
+
+You can extend or modify the syntax for specific properties.
+For example, to support a custom `bar` property that accepts `<color>` values:
+
+```json
+{
+  "languageOptions": {
+    "syntax": {
+      "properties": { "bar": "<color>" }
```
jeddy3, line 387:
```suggestion
      "properties": { "foo": "<color>" }
```

docs/user-guide/configure.md
```diff
@@ -254,6 +254,186 @@ But the following pattern would be reported as a warning:
...
+    }
+  }
+}
+```
+
+When `declaration-property-value-no-unknown` is enabled, The following patterns are considered problems:
+
+<!-- prettier-ignore -->
+```css
+a { bar: 10px; }
```
jeddy3, line 397:
```suggestion
a { foo: 10px; }
```

docs/user-guide/configure.md
```diff
@@ -254,6 +254,186 @@ But the following pattern would be reported as a warning:
...
+```
+
+The following patterns are _not_ considered problems:
+
+<!-- prettier-ignore -->
+```css
+a { bar: red; }
```
jeddy3, line 404:
```suggestion
a { foo: red; }
```

jeddy3:
> Do we give us this possibility in the future?

There's no harm in giving ourselves this possibility in the future.

jeddy3 (commented):
@ryo-manba This is looking fantastic, thank you.

I've added some minor doc requests in addition to @Mouvedia suggestion to use `meta.lexer = "csstree"`.

Let's wait a little longer to see if anyone has time to chime in on https://github.com/stylelint/stylelint/pull/8297#discussion_r1923590960.

docs/user-guide/configure.md
```diff
@@ -254,6 +254,186 @@ But the following pattern would be reported as a warning:
...
+
+<!-- prettier-ignore -->
+```css
+@example custom-ident { foo: 10; bar: red; }
+```
+
+#### `cssWideKeywords`
+
+You can add custom global keywords that are accepted as valid property values.
+For example, to add `-moz-initial` to the wide keywords:
```
Mouvedia, line 352:
```suggestion
For example, to add `-moz-initial` to the [wide keywords](https://drafts.csswg.org/css-values-4/#css-wide-keywords):
```

docs/user-guide/configure.md
```diff
@@ -254,6 +254,186 @@ But the following pattern would be reported as a warning:
...
+
+<!-- prettier-ignore -->
+```css
+a { color: my-global-value; }
+```
+
+#### `properties`
+
+You can extend or modify the syntax for specific properties.
+For example, to support a custom `bar` property that accepts `<color>` values:
```
Mouvedia, line 381: To go along with https://github.com/stylelint/stylelint/pull/8297/files#r1929561055

```suggestion
For example, to support a custom `foo` property that accepts `<color>` values:
```

lib/utils/hasRulesUsingLexer.mjs
```diff
@@ -0,0 +1,19 @@
+import getStylelintRule from './getStylelintRule.mjs';
+
+/**
+ * @param {import('stylelint').Config} config
+ * @returns {Promise<boolean>}
+ */
+export default async function hasRulesUsingLexer(config) {
+	const ruleNames = config.rules ? Object.keys(config.rules) : [];
```
Mouvedia, line 8: Maybe return `false` if `rules?.length` is falsy?

types/stylelint/index.d.mts
```diff
@@ -1,6 +1,7 @@
 export {
 	default,
 	Config,
+	ConfigRules,
```
Mouvedia, line 4: You are importing `ConfigRules` [here](https://github.com/stylelint/stylelint/pull/8297/files#diff-a885e25af8f2e8f600542d13e286eacdf8d80b1a527b58a0a6e88cc24063f7b5R19) but not using it.
If you are not planing to use it, can you revert this change and https://github.com/stylelint/stylelint/pull/8297/files#diff-9a4b261e295bef1ccce813aa194ac4023a1dcf98236667889964b1ba6fbc551fL11-L14?

lib/lintSource.mjs
```diff
@@ -89,6 +91,17 @@ export default async function lintSource(stylelint, options = {}) {
 			customSyntax: config.customSyntax,
 		}));
 
+	const usesLexer = await hasRulesUsingLexer(config);
+
+	const lexer = usesLexer
+		? fork({
+				atrules: config.languageOptions?.syntax?.atRules || {},
```
Mouvedia, line 98: question

Since `config` is common to all the rules that use that lexer—i.e. we only have one lexer right now—what about only forking once?
i.e. memoize `lexer`

lib/lintSource.mjs
```diff
@@ -97,6 +107,7 @@ export default async function lintSource(stylelint, options = {}) {
 			ruleMetadata: {},
 			fixersData: {},
 			disabledRanges: {},
+			lexer,
```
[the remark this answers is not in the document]
Mouvedia: `PostcssResult` is internal hence `PostcssResult['stylelint']` is also internal.
We may have to restructure `StylelintPostcssResult` at one point but I think that it is still manageable.
i.e. it's not surprising that he thought it would be appropriate to put it here

The only question that I would have would be regarding the nesting.
i.e. are we planning to add things that may be related to `lexer` in the future?
e.g.
```ts
	/** @internal */
	export type StylelintPostcssResult = {
		ruleSeverities: { [ruleName: string]: RuleSeverity };
		customUrls: { [ruleName: string]: string };
		rangesOfComputedEditInfos: Array<EditInfo['range']>;
		fixersData: { [ruleName: string]: number };
		quiet?: boolean;
		quietDeprecationWarnings?: boolean;
		disabledRanges: DisabledRangeObject;
		disabledWarnings?: DisabledWarning[];
		ignored?: boolean;
		stylelintError?: boolean;
		stylelintWarning?: boolean;
		config?: Config;
		foo: {
		  bar: {};
		  lexer:? Lexer;
		}
	};
```

types/stylelint/index.d.ts
```diff
@@ -84,6 +86,22 @@ declare namespace stylelint {
 		severity?: Severity;
 	};
 
+	type LanguageOptions = {
```
Mouvedia, line 89: question

Since you patch `css-tree`, can't you use that type instead?
e.g.
```ts
import type { fork } from 'foo';

type LanguageOptions = {
  syntax: Parameters<fork>[0];
};
```

Mouvedia (commented):
I did a first pass hastily.
I didn't check the tests yet.
It's looking good so far.

Iv left some suggestions, questions and comments.

types/stylelint/index.d.ts
ryo-manba: In that case, the stylelint side should also be `atrules` instead of `atRules`.

lib/utils/hasRulesUsingLexer.mjs
ryo-manba: Fixed in this commit!
[Refactor check rule length and early return](https://github.com/stylelint/stylelint/pull/8297/commits/6997f04434228e05fb129fc0e9bf7844cf3cf52f)

types/stylelint/index.d.mts
ryo-manba: Good catch! Fixed in this commit.
[Refactor remove unused import](https://github.com/stylelint/stylelint/pull/8297/commits/0fbcbe41586facbde6e0ddfcf954f2c2ecc52bc3)

lib/lintSource.mjs
ryo-manba: It is forked only once when lint is run, so I guess it's not a problem.

types/stylelint/index.d.ts
Mouvedia: Since you are already patching it, does it matter?
i.e. will the patch stay?

lib/lintSource.mjs
Mouvedia: IIRC we have new lint per file?
i.e. if you pass `a.css b.css` wouldn't `lexer` be created twice for the same config?

I guess it's safer because we have the `--config` flag.
What I mean by that—_if we do create a new lexer per lint—memoizing it would only be allowed if it's the same config each time.
I dunno, maybe a perf todo comment will do for now.

types/stylelint/index.d.mts
Mouvedia: You forgot to revert the changes in `index.d.ts`.

types/stylelint/index.d.ts
```diff
@@ -161,6 +180,7 @@ declare namespace stylelint {
 		stylelintError?: boolean;
 		stylelintWarning?: boolean;
 		config?: Config;
+		lexer?: Lexer;
```
Mouvedia, line 183: @jeddy3 in response to https://github.com/stylelint/stylelint/pull/8297#pullrequestreview-2574034022

If it's very likely to happen, shouldn't we opt for
```suggestion
		lexers: {
		  csstree?: Lexer;
		}
```
instead?

types/stylelint/index.d.ts
```diff
@@ -251,6 +271,7 @@ declare namespace stylelint {
 		url: string;
 		deprecated?: boolean;
 		fixable?: boolean;
+		lexer?: 'csstree';
```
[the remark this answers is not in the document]
Mouvedia: @jeddy3 do you think a rule will use 2 or more lexers?
I feel like I am being too careful; I am not sure.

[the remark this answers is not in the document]
jeddy3: > do you think a rule will use 2 or more lexers?

I don't know, but if adding that future-proofing requires little effort and overhead, then I think we can go with an array.

types/stylelint/index.d.ts
ryo-manba: If the options become `atrules`, would that be okay? 
I don't have a strong preference, so I'm okay either way.
https://github.com/stylelint/stylelint/blob/8d8e75b54770da0711977641a50888b38dab2e5a/docs/user-guide/configure.md?plain=1#L314

Mouvedia: You would need to remap it based on the other object and use [`as`](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#key-remapping-via-as).
This sounds too complicated.
i.e. not worth the effort
I wanted not to have to maintain our own type just for a camelcase difference.

Ill ask another maintainer just in case.
@ybiquitous @jeddy3 @romainmenke 
Should we base our type from theirs or not?

types/stylelint/index.d.ts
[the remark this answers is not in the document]
Mouvedia: @ryo-manba based on Richard's response I think you can go ahead and rename `lexer` to `lexers` for `meta`.

I leave it you whether you want to go ahead with https://github.com/stylelint/stylelint/pull/8297#discussion_r1929658975 suggestion or not, since it's internal it can be done at a later time.
i.e. if we add another lexer

docs/user-guide/configure.md
```diff
@@ -254,6 +254,186 @@ But the following pattern would be reported as a warning:
...
+#### `types`
+
+You can extend or modify the syntax for specific types.
+For example, to support a new function `--foo()` that accepts `<length-percentage>` values and use it within the top property:
+
+```json
+{
+  "languageOptions": {
+    "syntax": {
+      "types": { "--foo()": "--foo( <length-percentage> )" },
```
Mouvedia, line 416: question

Can you create new custom "syntax strings" using `types`?
e.g. `"types": { "<foo>": "<length-percentage>" },`

types/stylelint/index.d.ts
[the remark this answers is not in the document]
ryo-manba: Is there a realistic scenario where this would actually be required? 
I feel it's better to keep the implementation simple and focus on addressing real needs as they arise, rather than designing for an uncertain future.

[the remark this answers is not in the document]
Mouvedia: I am not certain either way; what I know is that if we want to add new lexers in the future and one rule do support multiple lexers, which is unlikely, we will have to keep `lexer` and add `lexers`.
i.e. removing `lexer` would be a breaking change
e.g.
```
	export type RuleMeta = {
		/* … */
		lexer?: string; // not removed
		lexers?: string[]; // added
	};
```

If it's acceptable to you, we can move on and not rename `lexer`.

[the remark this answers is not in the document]
jeddy3: > I feel it's better to keep the implementation simple and focus on addressing real needs as they arise, rather than designing for an uncertain future.

Let's go with your feeling as the rule's author and keep the implementation simple.

lib/lintSource.mjs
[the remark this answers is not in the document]
romainmenke: I think `context` could also be a good position.

This already holds `newline`, `fix` (deprecated), ...
These are all bits of information from the engine to individual rules.

romainmenke (commented):
Thank you for working on this @ryo-manba!
Very excited about this!

lib/lintSource.mjs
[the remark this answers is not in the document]
Mouvedia: That's an alternative indeed. The code would have to be put in `lintPostcssResult`.
The only problem I see with `context` is that any change to the field will be a breaking change.
e.g. cannot rename `RuleContext['lexer']` easily because `RuleContext` is not internal.

lib/lintSource.mjs
```diff
@@ -89,6 +91,17 @@ export default async function lintSource(stylelint, options = {}) {
 			customSyntax: config.customSyntax,
 		}));
 
+	const usesLexer = await hasRulesUsingLexer(config);
```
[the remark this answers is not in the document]
Mouvedia: > memoize the result

👍 

see also https://github.com/stylelint/stylelint/pull/8297/files#r1929578482

lib/lintSource.mjs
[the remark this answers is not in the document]
romainmenke: Indeed, can only be an optional field in a minor.

types/stylelint/index.d.ts
Mouvedia: What I was trying to get at is, whether we create that type from theirs and rename `atrules` to `atRules` using `as` or just maintain our own just due to that one difference.

lib/lintSource.mjs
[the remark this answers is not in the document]
Mouvedia: Looking at `hasRulesUsingLexer`, all rules share the same `stylelintPostcssResult` per lint.
Hence putting it in `RuleContext ` would be more granular.

So I would be in favour of either:
- `PostcssResult['stylelint']['lexers']`
  - internal
  - future proof
  - gives access to lexers to formatters 
- or `RuleContext['lexer']`/`RuleContext['lexers']`
  - granular
  - expected place to find it
  - cannot be removed easily (CON)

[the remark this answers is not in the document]
ryo-manba: @romainmenke @Mouvedia 
Thanks for the suggestions and for clearly outlining the pros and cons!

I’ve moved `lexer` to `context` as suggested.
[Refactor move lexer to context](https://github.com/stylelint/stylelint/pull/8297/commits/b32e2e506ff724cd289a784eea0366b6f74ac8a3)

Following this [comment](https://github.com/stylelint/stylelint/pull/8297#discussion_r1935934065), I also added caching for the `lexer`, which allowed us to drop the related meta data.

Let me know if you see further improvements!

types/stylelint/index.d.ts
ryo-manba: If the types don’t match exactly, it might be better to have our own format. 
Using `as` to rename and extend could make things harder to understand.

docs/user-guide/configure.md
ryo-manba: Yes, for example:

```js
properties: { top: '| <bar>' },
types: { bar: '<color>' },
```

With this setup, `a { top: red; }` won’t trigger a warning.

Mouvedia: off-topic

So the follow-up question is should we accept those custom types in `syntax-string-no-invalid`?

types/stylelint/index.d.ts
Mouvedia: Alright, we can reconsider this decision in the PR that will remove the patch.

.changeset/fuzzy-spiders-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": minor
+---
+
+Added: languageOptions configuration property
```
Mouvedia, line 5:
```suggestion
Added: `languageOptions` configuration property
```

lib/lintSource.mjs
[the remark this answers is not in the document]
Mouvedia: The lexer is now accessible to all rules.
i.e. it's global

jeddy3 (commented):
> Very excited about this!

Likewise! Thanks for working on this @ryo-manba .

Are there any other changes to make? It LGTM.

lib/lintPostcssResult.mjs
```diff
@@ -154,3 +157,23 @@ function warnDeprecatedRule(result, ruleName) {
...
+		cachedLexer = fork({
+			atrules: config.languageOptions?.syntax?.atRules || {},
+			properties: config.languageOptions?.syntax?.properties || {},
+			types: config.languageOptions?.syntax?.types || {},
+			cssWideKeywords: config.languageOptions?.syntax?.cssWideKeywords || [],
+		}).lexer;
+	}
+
+	return cachedLexer;
+}
```
romainmenke, line 179: This now uses a "per process" cache.

In practice this might lead to unexpected behavior because a single process might lint multiple files, each with different config and different values for `languageOptions`.

Instead we should cache per call to `lintPostcssResult`.

This still has the benefit of only creating one lexer fork for all rules for each lint.

romainmenke (commented):
Hi @ryo-manba,
Thank you for applying the suggestions 👍 

There are two aspects that I think can be improved:
- `context.lexer` should be typed as an optional field so that this can be shipped in a minor (a new required field on this object would be a breaking change)
- we should avoid "cache pollution" issues

Unfortunately I suddenly have less time available than I would like and than is needed to give the most helpful feedback.
I hope that it is somewhat clear what I mean 🙇

lib/lintPostcssResult.mjs
ryo-manba: Does this mean that if `a.css` and `b.css` have different configs, the first `lexer` could be reused for both and cause issues? I wasn't sure how common it is for a single process to lint files with different configs, but it seems like the possibility isn't zero.

Your suggestion to move the cache to the function level makes sense to avoid this. 
I'll implement that change. Thanks for the insightful feedback!

ryo-manba: I changed the cache to be managed using the `syntax` as the key. What do you think?
[Change lexer cache to use syntax-based key to prevent cache pollution](https://github.com/stylelint/stylelint/pull/8297/commits/ee23855da4906c15e9f8b7e3ffdc2878c04aec79)

types/stylelint/index.d.ts
```diff
@@ -237,6 +255,7 @@ declare namespace stylelint {
 		configurationComment?: string | undefined;
 		fix?: boolean | undefined;
 		newline?: string | undefined;
+		lexer: Lexer;
```
Mouvedia: https://github.com/stylelint/stylelint/pull/8297#discussion_r1936022646

```suggestion
		lexer?: Lexer | undefined;
```

jeddy3 (approved):
LGTM.

@ryo-manba Thank you for working on this big addition!

lib/lintPostcssResult.mjs
ryo-manba: @romainmenke 
Can you see if there is any problem with this change?

jeddy3:
As we collectively lack time, I think we can leverage the community by merging and releasing new features like this, even if they don't have two reviews, in the hope that the community will identify (and maybe help fix) any issues in subsequent patch releases. 

Unless there are objections, I'll merge this and do a release in the next couple of days.

Issue #8304, opened 2025-01-13
Fix `ignore.default is not a function`


pkudryashov:
I updated `stylelint` to version `16.13` (`node v20.11.0`)  and I get the following error on startup

> TypeError: ignore.default is not a function
>     at getFileIgnorer (file:///D:/work/webui/node_modules/stylelint/lib/utils/getFileIgnorer.mjs:14:32)
>     at standalone (file:///D:/work/webui/node_modules/stylelint/lib/standalone.mjs:81:13)
>     at main (file:///D:/work/webui/node_modules/stylelint/lib/cli.mjs:528:9)
>     at file:///D:/work/webui/node_modules/stylelint/bin/stylelint.mjs:7:1
>     at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
>     at async ModuleLoader.import (node:internal/modules/esm/loader:329:24)
>     at async loadESM (node:internal/process/esm_loader:28:7)
>     at async handleMainPromise (node:internal/modules/run_main:113:12)

If fix in the file `node_modules\stylelint\lib\utils\getFileIgnorer.mjs`
```diff
-const ignorer = ignore.default();
+const ignorer = ignore();
```
error disappears

goapunk:
as a workaround for now you can also pin `ignore` to `7.0.0` in `package.json`

romainmenke:
@pkudryashov thank you for reporting this!

As I've updated `ignore` to `7.0.0` I took a short look at this issue and prepared a patch here: https://github.com/stylelint/stylelint/pull/8305

----

To fellow maintainers:

I'm unavailable for the next few hours so feel free to use the proposed changes if there is a need for an urgent patch :)

Mister-Hope:
> as a workaround for now you can also pin `ignore` to `7.0.0` in `package.json`

A better workaround should be pining stylelint, otherwise you will get yourself break again once #8305 is merged.

Mouvedia:
@jeddy3 @ybiquitous

Mouvedia:
1. `7.0.1` is not even listed in the [releases page](https://github.com/kaelzhang/node-ignore/releases) nor the [changelog file](https://github.com/kaelzhang/node-ignore/blob/HEAD/CHANGELOG.md)
2. the version `7.0.1` of `ignore` was published the 13/01/2025 but the latest Stylelint version was published the 12/01/2025
i.e. we couldn't detect the problem in advance

[...]

PR #8305, opened 2025-01-13
Fix `ignore.default is not a function`

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/8304

No, it's self-explanatory.

### DIFF
--- a/.changeset/pink-panthers-fix.md
+++ b/.changeset/pink-panthers-fix.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `ignore.default is not a function`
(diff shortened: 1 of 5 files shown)


Mister-Hope:
It would be great to have this fixed and releases ASAP, as overriding for ignore now will make users project break again once this PR is merged and a new version of stylelint is released.

romainmenke:
> ~~It would be great to have this fixed and releases ASAP, as overriding ignore to `7.0.0` now will make users project break again once this PR is merged and a new version of stylelint is released.~~
> 
> 
> 
> Edited: I believe it's better to pin the ignore to `7.0.0` instead of applying `7.0.1` change, it's ignore that is not following semver and it will  possibly reverting the previous behavior in next patch and raise another major.

See the two commits I made :) 

I am anticipating a potential upstream revert and made sure that both versions work with the changes here in Stylelint.

jeddy3 (approved):
> I am anticipating a potential upstream revert and made sure that both versions work with the changes here in Stylelint.

Nicely done. 

I think I've enough time to do a release. We may as well release to minimise the length of disruption to our users, regardless of what action is taken upstream.

Issue #8308, opened 2025-01-13
Fix `RuleContext['fix']` type


Mouvedia:
### Details

https://github.com/stylelint/stylelint/blob/8a173d4ad4164aae9160b69b248de573bd2235d9/types/stylelint/index.d.ts#L234

does not reflect

https://github.com/stylelint/stylelint/blob/8a173d4ad4164aae9160b69b248de573bd2235d9/lib/lintPostcssResult.mjs#L92

https://github.com/stylelint/stylelint/blob/8a173d4ad4164aae9160b69b248de573bd2235d9/lib/lintPostcssResult.mjs#L106

see also https://github.com/stylelint/stylelint/pull/8106#discussion_r1845183846

### Which Stylelint-related dependencies are you using?

```json
{
  "stylelint": "16.13.0"
}
```

### What did you expect to happen?

no semantic error TS2322

### What actually happened?

It was detected the 20/12/2024 by [this job](https://github.com/stylelint/stylelint-ecosystem-tester/actions/runs/12423419685/job/34686910033) but we overlooked it.
i.e. `stylelint-ecosystem-tester` did its job

```
 semantic error TS2322: Type '{ (properties: PrimaryOptions, options: SecondaryOptions, context?: RuleContext): (root: Root, result: PostcssResult) => void; primaryOptionArray: boolean | undefined; ruleName: string; messages: stylelint.RuleMessages; meta: stylelint.RuleMeta | undefined; }' is not assignable to type 'StylelintPlugin<PrimaryOptions, SecondaryOptions>'.
  Type '{ (properties: PrimaryOptions, options: SecondaryOptions, context?: RuleContext): (root: Root, result: PostcssResult) => void; primaryOptionArray: boolean | undefined; ruleName: string; messages: stylelint.RuleMessages; meta: stylelint.RuleMeta | undefined; }' is not assignable to type 'RuleBase<PrimaryOptions, SecondaryOptions>'.
    Types of parameters 'context' and 'context' are incompatible.
      Type 'import("/…/index").RuleContext' is not assignable to type 'RuleContext'.
        Types of property 'fix' are incompatible.
          Type 'boolean | FixMode | undefined' is not assignable to type 'boolean | undefined'.
            Type '"lax"' is not assignable to type 'boolean | undefined'.
```

### Do you have a proposal to fix the bug?

1. [x] revert to `fix?: boolean | undefined;`
[...]

PR #8309, opened 2025-01-13
Fix `RuleContext['fix']` type

Mouvedia:

#8308
doesn't c l o s e it (a new enhancement issue must be created)

1.
guilty commit: 0f92394
We failed to detect the problem during the review.

2.
So that it doesn't happen again here are some suggestions:

- improve `type-test.ts`
- review the errors reported by `stylelint-ecosystem-tester` more carefully during the pre-release phase
- increase the number of approvals needed when the number of `mjs` files reach a certain threshold—e.g. 3 approvals

These are just suggestions—i.e. overcompensating can be bad too—maybe it was simply an oversight and it can't be helped.

### DIFF
--- a/.changeset/lucky-garlics-compare.md
+++ b/.changeset/lucky-garlics-compare.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `RuleContext['fix']` type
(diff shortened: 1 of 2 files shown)


ryo-manba (approved):
LGTM, thanks for fixing my mistake.

Mouvedia:
[`stylelint-declaration-strict-value` next test passes](https://github.com/stylelint/stylelint-ecosystem-tester/actions/runs/12839219175)

PR #8310, opened 2025-01-13
Move --fix flag to last

### LINKED ISSUE #8301: Fix unexpected error `Unknown value for flag --fix`

```sh
npm exec stylelint -- --fix "foo.css" 
```

_any configuration_

```sh
npm exec stylelint -- --fix "foo.css" 
```

`stylelint@16.13.0`

I expected `--fix` to work as a boolean flag, as it did before.

### What actually happened?

```
Error: Unknown value for flag `--fix`: `foo.css`. Value must be one of: [`true`, `lax`, `strict`]
```

related to: 
- https://github.com/stylelint/stylelint/issues/7497
- https://github.com/sindresorhus/meow/issues/44

Even if `--fix` was the last flag before the list of inputs it worked fine up to version `16.12.0` because `--fix` was a boolean flag.

Now that `--fix` also supports string keywords as inputs the flag has become ambiguous.
It treats the input option as an incorrect flag keyword.

I suspect that I am not alone in having scripts all over with `--fix` right before the inputs 😅 .
Is there anything we can do to improve the current implementation?

### Workaround

Move the `--fix` flag before any other flag or after the list of inputs.

```diff
- npm exec stylelint -- --max-warnings=0 --fix "foo.css" 
+ npm exec stylelint -- --fix --max-warnings=0 "foo.css"
```

```diff
- npm exec stylelint -- --fix "foo.css" 
+ npm exec stylelint -- "foo.css" --fix 
```

immitsu:
Hi!
PR closes #8301
Looks like a workaround but works

### DIFF
--- a/lib/__tests__/cli.test.mjs
+++ b/lib/__tests__/cli.test.mjs
@@ -106,6 +106,7 @@ describe('buildCLI', () => {
 		expect(buildCLI(['--fix']).flags.fix).toBe(true);
 		expect(buildCLI(['--fix=lax']).flags.fix).toBe('lax');
 		expect(buildCLI(['--fix=strict']).flags.fix).toBe('strict');
+		expect(buildCLI(['--fix', '/path/to/file']).flags.fix).toBe(true);
 	});
 
 	it('flags.validate', () => {
--- a/lib/cli.mjs
+++ b/lib/cli.mjs
@@ -638,6 +638,17 @@ function importCustomFormatter(fileOrModulePath) {
  * @param {string[]} argv
  */
 export function buildCLI(argv) {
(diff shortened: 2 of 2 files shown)


lib/__tests__/cli.test.mjs
Mouvedia, line 1: @ybiquitous
Do or should `--fix=lax` be equivalent to `--fix lax`?
I am asking because his hackfix will have to modified to ignore both `--fix lax` and `--fix strict` if that's the case.

ybiquitous: Absolutely yes. I just opened PR #8313, which should fix the problem.

PR #8313, opened 2025-01-14
Fix `--fix` CLI flag to raise unknown value error

### LINKED ISSUE #8301: Fix unexpected error `Unknown value for flag --fix`

ybiquitous:

Closes #8301

For example, when `--fix foo.css` is given, `foo.css` was previously interpreted as a `--fix` argument instead of a file path.

This change tries to consider `foo.css` as an appropriate file path.


romainmenke (approved):
Thank you @ybiquitous. LGTM 👍

Nice to see the increased test coverage as well.

jeddy3 (approved):
LGTM, thank you!

I can do the release now.

lib/cli.mjs
```diff
@@ -472,7 +473,16 @@ export default async function main(argv) {
...
 
-	options.fix = normalizeFixMode(fix);
+	if (isString(fix)) {
+		options.fix = normalizeFixMode(fix);
+
+		// If the fix argument is unknown, consider it as an input path.
+		// TODO: `false` is for backward compatibility. We can remove it in the next major (17.0.0).
+		if (options.fix === undefined && fix !== 'false') {
+			options.fix = DEFAULT_FIX_MODE;
+			cli.input.push(fix);
```
Mouvedia, line 483: question
nit

`input` is "non-flag arguments" which IIUC means that `--fix foo.css --bar=0 qux.css` would now be interpreted as `--fix --bar=0 qux.css foo.css` correct?
Is there any reason why we would want to `unshift` instead?
i.e. start vs end
Would it make a difference?

lib/cli.mjs
```diff
@@ -242,7 +243,7 @@ const flags = {
 		type: 'boolean',
 	},
 	fix: {
-		choices: [true, 'lax', 'strict'],
+		type: 'string',
```
Mouvedia, line 246: note

This avoids ``Error: Unknown value for flag `--fix`: `false`.``.
It also permits us to easily support `=0`/`=1` _if it were requested_ without having to maintain a list.
i.e. it used to be unofficially supported and could be considered a downgrade

PR #8319, opened 2025-01-14
Fix `at-rule-descriptor-no-unknown` false positives for page and margin context properties in `@page`

### LINKED ISSUE #8316: Fix `at-rule-descriptor-no-unknown` false positives for page and margin context properties in `@page`

```css
@page {
	margin: 2cm;
}
```

```json
{
  "extends": [
    "stylelint-config-standard"
  ]
}
```

[Demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcIACAHAhgcxgAmAHQDs8AXAWzQCcMBLA+bAJjFIG5CBfEAGhADNqANjAByaUnEQwAHmJRCAdGADOS7uAgF+GBCHwFs2PCGnEYBKEqP0A2oQMGjS4gE8hA2sQC0kTdQyenNHNKKCM7bABdDjUfLQAxCApyYh0AKyUNNVgUVUQ9BxAnVxh3AmIrQxABNFMnIy5wxxc3D28NLQDiIKgQiqNq2vKQaPYgA)

```json
{
  "stylelint": "16.13.2",
  "stylelint-config-standard": "37.0.0"
}
```

No problems to be reported.

### What actually happened?

```
2:2-8 error Unexpected unknown descriptor "margin" for at-rule "@page" at-rule-descriptor-no-unknown
```

This seems to be valid CSS according to https://developer.mozilla.org/en-US/docs/Web/CSS/@page

jeddy3:

Closes #8316 

I've used the `content` property in some tests because it's one of the margin context properties that isn't a page context one.


lib/reference/properties.mjs
```diff
@@ -644,3 +644,146 @@ export const longhandTimeProperties = new Set([
...
+	'margin-top',
+	'margin-right',
+	'margin-bottom',
+	'margin-left',
+	'margin',
+	'outline-width',
+	'outline-style',
+	'outline-color',
+	'outline',
+	'padding-top',
```
Mouvedia, line 698: question
minor

Are `orphans` and `widows` missing from the list?
I am not really sure.

Here are some links that seems to confirm it:

https://olevelexam.com/css-property-orphans
https://web.archive.org/web/20150510080955/https://developer.mozilla.org/en-US/docs/Web/CSS/@page
https://publishing-project.rivendellweb.net/css-paged-media/#controlling-widows-and-orphans

I couldn't find any examples for `page-break-before`/`page-break-after`/`page-break-inside` though.

@romainmenke you might know more?
Are these examples invalid?

jeddy3: Let's start with the ones from the spec in this PR. If a user reports `orphans` and `widows` we can handle those in another PR.

lib/reference/properties.mjs
```diff
@@ -644,3 +644,146 @@ export const longhandTimeProperties = new Set([
...
+	'text-transform',
+	'white-space',
+	'word-spacing',
+	'vertical-align',
+	'visibility',
+	'width',
+	'min-width',
+	'max-width',
+	'z-index',
+]);
```
Mouvedia, line 789: If you import `uniteSets` you could reduce it to:

```suggestion
export const marginContextProperties = uniteSets(pageContextProperties, new Set([
	'unicode-bidi',
	'content',
	'overflow',
	'z-index',
	'vertical-align',
]));
```

I am not sure if that's better to create such a relation though.
i.e. `pageContextProperties` may contain properties that are not part of `marginContextProperties` one day in the future

lib/rules/at-rule-descriptor-no-unknown/__tests__/index.mjs
```diff
@@ -85,5 +97,32 @@ testRule({
...
+			message: messages.rejected('@page', 'content'),
+			line: 1,
+			column: 9,
+			endLine: 1,
+			endColumn: 16,
+			description: 'a margin context property in @page',
+		},
+		{
+			code: '@page { @top-center { bar: 0; } }',
+			message: messages.rejected('@page', 'bar'),
```
Mouvedia, line 120: question

Will it warn twice?
i.e. `@top-center` also contains descriptors
e.g.
```
warnings: [
  {
    /* … */
    message: messages.rejected('@page', 'bar'),
  },
  {
    /* … */
    message: messages.rejected('@top-center', 'bar'),
  },
]
```

jeddy3: It checks `@top-conter` but in the context of `@page` as CSSTree only has descriptors for `@page`: https://github.com/stylelint/stylelint/pull/8319/commits/34da558ee366c1d9d4d2dd0fc373ea65e290f555.

lib/rules/at-rule-descriptor-no-unknown/__tests__/index.mjs
```diff
@@ -28,6 +28,18 @@ testRule({
...
+			code: '@page { size: landscape; }',
+			description: 'known descriptor in @page',
+		},
+		{
+			code: '@page { color: red; }',
+			description: 'known page context property in @page',
+		},
+		{
+			code: '@page { @top-center { content: "foo"; } }',
+			description: 'known margin context property within margin at-rule',
```
ryo-manba, line 41: 👍

lib/reference/properties.mjs
```diff
@@ -644,3 +644,146 @@ export const longhandTimeProperties = new Set([
 ]);
 
 export const shorthandTimeProperties = new Set(['transition', 'animation']);
+
+export const pageContextProperties = new Set([
```
ryo-manba, line 648: memo: https://drafts.csswg.org/css-page/#page-property-list

lib/reference/properties.mjs
```diff
@@ -644,3 +644,146 @@ export const longhandTimeProperties = new Set([
...
+	'text-transform',
+	'white-space',
+	'word-spacing',
+	'visibility',
+	'width',
+	'min-width',
+	'max-width',
+]);
+
+export const marginContextProperties = new Set([
```
ryo-manba, line 717: memo: https://drafts.csswg.org/css-page/#margin-property-list

lib/rules/at-rule-descriptor-no-unknown/__tests__/index.mjs
```diff
@@ -85,5 +97,51 @@ testRule({
...
+			code: '@page { @top-center { bar: 0; } }',
+			message: messages.rejected('@page', 'bar'),
+			line: 1,
+			column: 23,
+			endLine: 1,
+			endColumn: 26,
+			description: 'an unknown descriptor or margin context property in a margin at-rule',
+		},
+		{
+			code: '@page { foo: 0; @top-center { bar: 0; } }',
```
ryo-manba, line 128: That's a good test!

ryo-manba (approved):
LGTM! Thank you for the quick fix!

lib/rules/at-rule-descriptor-no-unknown/__tests__/index.mjs
Mouvedia: https://csstree.github.io/docs/syntax/#discovery&q=ZGljdC5bdHlwZT0iQXRydWxlIl0%3D

lib/rules/at-rule-descriptor-no-unknown/__tests__/index.mjs
Mouvedia, line 1: suggestion
minor

Can you add a reject case for `@page { margin-block: 2px 1px; }`?
i.e. it accepts `margin` but not its logical counterparts

lib/rules/at-rule-descriptor-no-unknown/__tests__/index.mjs
Mouvedia: I agree.
i.e. once `css-tree` will support margin at-rules, it will be triggered

PR #8322, opened 2025-01-15
Fix `declaration-property-value-no-unknown` false positives for `calc-size` and `shape` functions

### LINKED ISSUE #8320: Unexpected unknown value "calc-size(auto, size)" for property "height"  declaration-property-value-no-unknown

```
import { css } from '@emotion/react';
import styled from '@emotion/styled';

export const StyledAlertAnimatedExpandable = styled.div<{ isExpanded: boolean }>(
  ({ isExpanded }) => {
    return css`
      display: none;
      height: 0;
      overflow: hidden;
      transition:
        height 300ms ease-out,
        display 300ms;
      transition-behavior: allow-discrete;

      @starting-style {
        height: 0;
      }

      ${isExpanded &&
      css`
        display: block;
        height: calc-size(auto, size);

        @supports not (height: calc-size(auto, size)) {
          height: auto;
        }
      `}
    `;
  }
);
```

it fails for line: `height: calc-size(auto, size);`

```json
{
  "extends": ["stylelint-config-standard"],
  "plugins": ["stylelint-prettier"],
  "ignoreFiles": ["build/**"],
  "overrides": [
    {
      "files": ["**/*.js", "**/*.cjs", "**/*.mjs", "**/*.jsx", "**/*.ts", "**/*.tsx"],
      "customSyntax": "postcss-styled-syntax"
    }
  ]
}
```

`stylelint 'src/**/*.{tsx,css}'`

```json
"postcss-styled-syntax": "^0.7.0",
"stylelint": "^16.13.2",
"stylelint-config-standard": "^37.0.0",
"stylelint-prettier": "^5.0.2"
```

According to https://developer.mozilla.org/en-US/docs/Web/CSS/calc-size this property is not unknown and is valid.

### What actually happened?

```
22:17  ✖  Unexpected unknown value "calc-size(auto, size)" for property "height"  declaration-property-value-no-unknown
```

_No response_

Mouvedia:

Closes #8320

https://github.com/stylelint/stylelint/issues/8320#issuecomment-2592683396
https://csstree.github.io/docs/validator.html

### DIFF
--- a/lib/rules/declaration-property-value-no-unknown/index.mjs
+++ b/lib/rules/declaration-property-value-no-unknown/index.mjs
@@ -202,15 +202,15 @@ const rule = (primary, secondaryOptions) => {
 /**
  * @see csstree/csstree#164 min, max, clamp
  * @see csstree/csstree#245 env
+ * @todo create csstree issue for shape and calc-size
  * @param {import('css-tree').CssNode} cssTreeNode
  * @returns {boolean}
  */
 function containsUnsupportedFunction(cssTreeNode) {
+	const functions = ['clamp', 'min', 'max', 'env', 'shape', 'calc-size'];
+
 	return Boolean(
-		find(
-			cssTreeNode,
-			(node) => node.type === 'Function' && ['clamp', 'min', 'max', 'env'].includes(node.name),
-		),
+		find(cssTreeNode, (node) => node.type === 'Function' && functions.includes(node.name)),
 	);
 }
 
--- a/.changeset/fast-bikes-visit.md
+++ b/.changeset/fast-bikes-visit.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `declaration-property-value-no-unknown` false positives for `calc-size()` and `shape()`
--- a/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
+++ b/lib/rules/declaration-property-value-no-unknown/__tests__/index.mjs
@@ -30,6 +30,12 @@ testRule({
 			code: 'a { top: env(titlebar-area-height); }',
 		},
 		{
+			code: 'a { height: calc-size(auto, round(up, size, 1px)); }',
+		},
+		{
+			code: 'a { clip-path: shape(nonzero from 0 0, line to 1px 1px); }',
+		},
+		{
 			code: 'a { foo: 1px; }',
 		},
 		{
--- a/lib/rules/declaration-property-value-no-unknown/index.cjs
+++ b/lib/rules/declaration-property-value-no-unknown/index.cjs
@@ -204,15 +204,15 @@ const rule = (primary, secondaryOptions) => {
 /**
  * @see csstree/csstree#164 min, max, clamp
  * @see csstree/csstree#245 env
+ * @todo create csstree issue for shape and calc-size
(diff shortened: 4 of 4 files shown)


lib/rules/declaration-property-value-no-unknown/index.mjs
```diff
@@ -202,15 +202,15 @@ const rule = (primary, secondaryOptions) => {
 /**
  * @see csstree/csstree#164 min, max, clamp
  * @see csstree/csstree#245 env
+ * @todo create csstree issue for shape and calc-size
  * @param {import('css-tree').CssNode} cssTreeNode
  * @returns {boolean}
  */
 function containsUnsupportedFunction(cssTreeNode) {
+	const functions = ['clamp', 'min', 'max', 'env', 'shape', 'calc-size'];
```
romainmenke, line 210: Can you omit `shape` both from the implementation and the titles?

This was not part of the original bug report and isn't actually the same scenario.

`shape()` is an extremely new spec and isn't a math function. If you add the syntax definitions validation also works just fine.

romainmenke (commented):
Thank you for this pull request @Mouvedia

I am not sure we want to fix the issue in this manner.
`calc-size()` works almost fine if you add the syntax definition.
The sole exception is the `size` keyword. So adding it to a list of unsupported functions seems incorrect to me.

Screenshot 2025-01-15 at 14 12 27

Maybe we should instead focus on fixing validation for keywords which are only valid in specific contexts?

Mouvedia:
> Maybe we should instead focus on fixing validation for keywords which are only valid in specific contexts?

Are you thinking of a solution that would cover #7241 as well?

> I am not sure we want to fix the issue in this manner.
> `calc-size()` works almost fine if you add the syntax definition.
> The sole exception is the `size` keyword.

Can we merge this as a temporary solution?
I have removed _Closes_ from [OP](https://github.com/stylelint/stylelint/pull/8322#issue-2789771115).

lib/rules/declaration-property-value-no-unknown/index.mjs
Mouvedia: > If you add the syntax definitions validation also works just fine.

That's important.

> Can you omit `shape` both from the implementation and the titles?

Will do.

romainmenke:
> > Maybe we should instead focus on fixing validation for keywords which are only valid in specific contexts?
> Are you thinking of a solution that would cover https://github.com/stylelint/stylelint/issues/7241 as well?

Yes :)

Fixing `size` in `calc-size` will be a lot simpler than fixing relative color syntax because the syntax of `calc-size()` is much less varied. We only have to look for `size` in the second argument position and replace it with a `<length>` node before doing syntax validation.

We could start with only adding a fix for `size` in `calc-size` and then build on that for relative color syntax.

Such a fix can still be implemented relatively quickly because of the lower complexity.

Mouvedia:
> We could start with only adding a fix for `size` in `calc-size` and then build on that for relative color syntax.

So not a one-size-fits-all approach.

> Such a fix can still be implemented relatively quickly because of the lower complexity.

> Can we merge this as a temporary solution?

Do you want me to close this PR?

romainmenke:
Maybe we can repurpose it? If you want to work on it?
If not, then maybe indeed beter to close and take it back to the issue?

Issue #8327, opened 2025-01-16
Fix `declaration-property-value-keyword-no-deprecated` false positives


Stanzilla:
### What minimal example or steps are needed to reproduce the bug?

```typescript
const SensitiveContentInner = styled.div`
  background-color: ${palette(({ background }) => background.primary)};
  border: 1px solid ${palette(({ border }) => border.primary)};
  display: flex;
  flex-direction: column;
  align-items: center;
  justify-content: center;
  height: 100%;
  width: 100%;
`;

const RewardNoticeHolder = styled.div`
  position: absolute;
  background: ${palette(({ background }) => background.tertiary)};
  border-radius: ${shape(({ wrapperBorderRadius }) => wrapperBorderRadius)};
  border-top-right-radius: 0;
  padding: ${spacing(0.5)} ${spacing(1)};
  text-align: right;
  &::before {
    content: '';
    position: absolute;
    right: 0;
    bottom: 100%;
    width: 0;
    height: 0;
    border-style: solid;
    border-width: 0 0 12px 16px;
    border-color: transparent transparent
      ${palette(({ background }) => background.tertiary)} transparent;
  }
`;
```

### What minimal configuration is needed to reproduce the bug?

```js
/** @type {import('stylelint').Config} */
module.exports = {
  extends: ['stylelint-config-recommended'],
  customSyntax: 'postcss-styled-syntax',
  plugins: ['stylelint-plugin-logical-css'],
  rules: {
    'plugin/use-logical-properties-and-values': [
      true,
      {
        severity: 'warning',
        ignore: [
          'height',
          'max-height',
          'min-height',
          'width',
          'max-width',
          'min-width',
          'contain-intrinsic-height',
          'contain-intrinsic-width',
          'overflow-x',
          'overflow-y',
          'overscroll-behavior-x',
          'overscroll-behavior-y',
          'box-orient: vertical',
          'box-orient: horizontal',
          '-webkit-box-orient: vertical',
          '-webkit-box-orient: horizontal',
          'resize: horizontal',
          'resize: vertical'
        ]
      }
    ],
    'plugin/use-logical-units': [true, { severity: 'warning' }],
    'value-no-vendor-prefix': true,
    'property-no-vendor-prefix': true,
    'no-empty-source': null,
    'unit-no-unknown': [
      true,
      {
        ignoreUnits: ['svh']
      }
    ],
    'selector-type-no-unknown': [
      true,
      {
        ignoreTypes: [
[...]

PR #8328, opened 2025-01-16
Fix `declaration-property-value-keyword-no-deprecated` false positives for `border-color`

Mouvedia:

#8327

1.
`border-color` is not a single value color property.
i.e. it accepts 1, 2, 3 and 4
This was missed during the review of #8223.

2.
[shorthand properties are not supported](https://github.com/stylelint/stylelint/pull/8223#issue-2758167186)

### DIFF
--- a/lib/rules/declaration-property-value-keyword-no-deprecated/__tests__/index.mjs
+++ b/lib/rules/declaration-property-value-keyword-no-deprecated/__tests__/index.mjs
@@ -7,7 +7,7 @@ testRule({
 	fix: true,
 	accept: [
 		{
-			code: 'a { color: red; }',
+			code: 'a { border-color: background; }',
 		},
 	],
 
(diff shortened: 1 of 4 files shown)


lib/rules/declaration-property-value-keyword-no-deprecated/__tests__/index.mjs
```diff
@@ -7,7 +7,7 @@ testRule({
 	fix: true,
 	accept: [
 		{
-			code: 'a { color: red; }',
+			code: 'a { border-color: background; }',
```
jeddy3, line 10: Let's add a new test case here with a description rather than change an existing one.

jeddy3 (commented):
Thanks for opening the PR. I've requested a change to the test.

lib/rules/declaration-property-value-keyword-no-deprecated/__tests__/index.mjs
Mouvedia: [c1c408b](https://github.com/stylelint/stylelint/pull/8328/commits/c1c408ba8bb94e34d432ff676ef27f937f477b83)

jeddy3 (approved):
LGTM, thank you.

PR #8329, opened 2025-01-16
Fix `declaration-property-value-keyword-no-deprecated` false positives for non-standard syntax values

### LINKED ISSUE #8327: Fix `declaration-property-value-keyword-no-deprecated` false positives

Mouvedia:

Closes #8327

`${foo(({ background }) => background.bar)}` cannot be tested in the `accept` array because it gives a `CssSyntaxError` without `postcss-styled-syntax` which shan't be added as a dependency.
i.e. no `customSyntax: 'postcss-styled-syntax',` testRule
see discussion in #6556

ref: https://styled-components.com/docs/advanced#tagged-template-literals

### DIFF
--- a/.changeset/violet-shirts-dream.md
+++ b/.changeset/violet-shirts-dream.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `declaration-property-value-keyword-no-deprecated` false positives in styled component interpolations
+Fixed: `declaration-property-value-keyword-no-deprecated` false positives for non-standard syntax values
--- a/.changeset/violet-shirts-dream.md
+++ b/.changeset/violet-shirts-dream.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `declaration-property-value-keyword-no-deprecated` false positives in styled component interpolations
--- a/lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
+++ b/lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
@@ -4,6 +4,7 @@ import { isRegExp, isString } from '../../utils/validateTypes.mjs';
 import { declarationValueIndex } from '../../utils/nodeFieldIndices.mjs';
 import getDeclarationValue from '../../utils/getDeclarationValue.mjs';
 import isStandardSyntaxDeclaration from '../../utils/isStandardSyntaxDeclaration.mjs';
+import isStandardSyntaxValue from '../../utils/isStandardSyntaxValue.mjs';
 import optionsMatches from '../../utils/optionsMatches.mjs';
 import report from '../../utils/report.mjs';
 import ruleMessages from '../../utils/ruleMessages.mjs';
@@ -123,6 +124,8 @@ const rule = (primary, secondaryOptions) => {
 			const { prop, value } = decl;
 
 			if (optionsMatches(secondaryOptions, 'ignoreKeywords', value)) return;
+
+			if (!isStandardSyntaxValue(value)) return;
 
 			const keywords = PROPERTY_NAME_TO_KEYWORDS.get(prop.toLowerCase());
 
--- a/lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
+++ b/lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
@@ -125,13 +125,18 @@ const rule = (primary, secondaryOptions) => {
 
 			if (optionsMatches(secondaryOptions, 'ignoreKeywords', value)) return;
 
-			if (!isStandardSyntaxValue(value)) return;
-
 			const keywords = PROPERTY_NAME_TO_KEYWORDS.get(prop.toLowerCase());
 
 			if (!keywords) return;
 
 			const parsedValue = valueParser(getDeclarationValue(decl));
+			let isStandardKeyword = true;
+
+			parsedValue.walk(({ type, value: keyword }) => {
+				if (type === 'word' && !isStandardSyntaxValue(keyword)) isStandardKeyword = false;
+			});
+
+			if (!isStandardKeyword) return;
 
 			parsedValue.walk((node) => {
 				if (node.type !== 'word') return;
--- a/lib/rules/declaration-property-value-keyword-no-deprecated/__tests__/index.mjs
+++ b/lib/rules/declaration-property-value-keyword-no-deprecated/__tests__/index.mjs
@@ -83,3 +83,23 @@ testRule({
 		},
 	],
 });
+
+testRule({
+	ruleName,
+	customSyntax: 'postcss-scss',
+	config: true,
+	accept: [
+		{
+			code: `
+				$prefix: info;
+				a { background-color: $prefix + background; }
+			`,
+		},
+		{
+			code: `
+				$postfix: text;
+				a { background-color: menu + $postfix; }
+			`,
+		},
+	],
+});
--- a/lib/rules/declaration-property-value-keyword-no-deprecated/index.cjs
+++ b/lib/rules/declaration-property-value-keyword-no-deprecated/index.cjs
@@ -7,6 +7,7 @@ const validateTypes = require('../../utils/validateTypes.cjs');
 const nodeFieldIndices = require('../../utils/nodeFieldIndices.cjs');
 const getDeclarationValue = require('../../utils/getDeclarationValue.cjs');
 const isStandardSyntaxDeclaration = require('../../utils/isStandardSyntaxDeclaration.cjs');
+const isStandardSyntaxValue = require('../../utils/isStandardSyntaxValue.cjs');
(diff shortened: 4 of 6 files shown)


lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -129,10 +129,17 @@ const rule = (primary, secondaryOptions) => {
 			if (!keywords) return;
 
 			const parsedValue = valueParser(getDeclarationValue(decl));
+			let flag = false;
 
 			parsedValue.walk((node) => {
+				if (flag) return;
```
Mouvedia, line 135: @romainmenke I tried returning `false` + setting bubble to `true`; how do you break from `walk`?

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -129,10 +129,17 @@ const rule = (primary, secondaryOptions) => {
...
+
 				if (node.type !== 'word') return;
 
+				// FIXME once we support shorthands
+				// e.g. border-color: green red yellow ActiveBorder;
+				flag = true;
```
Mouvedia, line 141: @ybiquitous what's better in your opinion a flag like I did or `try`/`catch`?

romainmenke: A `throw` + `try/catch` is extremely slow.
When an actual exception needs to happen or needs to caught, then this is fine. But using it as a way to escape early will most likely be slower than just letting the `walk` run to completion.

Mouvedia: > A `throw` + `try/catch` is extremely slow.

I know…

lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
Mouvedia: > Can you add failing tests?

I can't.
see https://github.com/stylelint/stylelint/issues/8327#issuecomment-2596623482

But you can try `npm i -D postcss-styled-syntax` +
```js
testRule({
	ruleName,
	customSyntax: 'postcss-styled-syntax',
	config: [true],

	accept: [
		{
			code: `
				export const s = styled.a\`
					background-color: \${palette(({ background }) => background.primary)};
				\`;
			`,
		},
	],
});
```

romainmenke: Shouldn't we just return early when the value isn't a standard syntax value?
Currently we only check the property, not the value.

Mouvedia: I think that not running the code unnecessarily is still a good thing.
It will also skip cases that are invalid.

e.g.
someone who erroneously used `background-color` instead of `background-color`
`background-color: border-box WindowFrame;`

---

> Ill try to come up with a non css-in-js test.

I can't come up with a non-invalid CSS test case because we don't support shorthands yet.
i.e. the properties are filtered
see also #8328

romainmenke: If there is no failing test or real world use case then I can't really advise :/

romainmenke: If the concern is purely performance and that this change isn't intended to change the behavior then I would suggest adding an early check that the value at least contains one of the candidates. see: https://github.com/stylelint/stylelint/issues/6869

Skipping the parsing step will be much more impactful than stopping the walk early.

If however there is an intended behavior change, then there should be some illustration of why the current behavior isn't ideal :)
I am still trying to understand what the issue is and how this change relates to it.
Given that the title is also `TBD` I really have too little information to be helpful.
My advice is likely wrong purely because I don't have sufficient information.

Mouvedia: > Shouldn't we just return early when the value isn't a standard syntax value?

Ill just do that.

.changeset/violet-shirts-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `declaration-property-value-keyword-no-deprecated` false positives in styled component interpolations
```
romainmenke, line 5:
```suggestion
Fixed: `declaration-property-value-keyword-no-deprecated` false positives for non-standard syntax values
```

Because the change will also prevent false positives in other kinds of non-standard syntax.

Can you also change the title?

romainmenke (commented):
Looks good! Thank you!
I've made a suggestion for a more general title and changelog entry.

romainmenke:
> i.e. no `customSyntax: 'postcss-styled-syntax'`, hence codecov warning can be ignored

But we can have a test for a simpler value containing `$`. We don't have to test `postcss-styles-syntax`, we only have to test any non-standard value. Most (all?) rules with such conditions also have some samples of non-standard syntax.

For example: `lib/rules/declaration-block-no-duplicate-properties/__tests__/index.mjs`

```
		{
			code: 'a { $scss: 0; $scss: $scss + 1; }',
		},
		{
			code: 'a { @less: 0; @less: @less + 1; }',
		},
```

.changeset/violet-shirts-dream.md
romainmenke, line 5:
```suggestion
Fixed: `declaration-property-value-keyword-no-deprecated` false positives for non-standard syntax values
```

I think this was lost when you force pushed?

romainmenke (approved):
Thank you. LGTM 👍

PR #6390, opened 2022-10-03
Document list of rules by what-they-do groups

jeddy3:
> Which issue, if any, is this issue related to?

Ref: https://github.com/stylelint/stylelint/issues/6376#issuecomment-1263543829

> Is there anything in the PR that needs further explanation?

This pull request groups our list of rules by what they do rather than what they're applied to.

I believe this approach helps our users more easily:

- understand what the built-in rules do
- find related built-in rules

You can [view the page here](https://github.com/stylelint/stylelint/blob/5111971b8b97c996a1792504b02d9e936df0b300/docs/user-guide/rules.md).

I think it does a better job of communicating the number of powerful rules we have for enforcing non-stylistic conventions that are unique to Stylelint.

This change also surfaced a lot of inconsistencies in our rule descriptions, which I've fixed in this pull request.

Grouping by the thing the rule applies to was mostly redundant as the rule name starts with it. I've still used those sub groups for the larger `no-*` and `*-list` group for readability, though. And I kept the old system for the stylistic rules as they'll be removed at some point.

The list of rules page now leads with just enough information for users to understand what they're looking at, i.e.:
- what makes a Stylelint rule
- and how it's named

I've moved other information that lived on the "About rules" page to the "Working on rules" page in the developer guide, which feel like a better home for it as it's more useful for people writing rules than using them.

PR #8334, opened 2025-01-20
Fix `tap` formatter outputting invalid YAML blocks for rules containing double quotes in their warnings' messages

### LINKED ISSUE #8321: Fix TAP formatter generating invalid YAML blocks for rules containing double quotes in messages

1. Run stylelint
2. Examine generated Test Anything Protocol file

{
  "extends": ["stylelint-config-standard-scss"],
  "plugins": ["stylelint-use-logical"],
  "rules": {
    "csstools/use-logical": "always"
  }
}

`npx stylelint "**/*.scss" --formatter tap --output-file result.tap`

{
  "extends": ["stylelint-config-standard-scss"],
  "plugins": ["stylelint-use-logical"]
}

The generated Test Anything Protocol file should contain valid YAML blocks.
If a message contains double quotes they should be escaped like in the following example:

```
TAP version 14
1..1
not ok 1 - path/to/file.css
  ---
  csstools/use-logical:
    - message: "Unexpected \"width\" property. Use \"inline-size\". (csstools/use-logical)"
      severity: error
      line: 13
      column: 3
      endLine: 13
      endColumn: 15
  ...

```

### What actually happened?

Double quotes in YAML block are not escaped:

```
TAP version 14
1..1
not ok 1 - path/to/file.css
  ---
  csstools/use-logical:
    - message: "Unexpected "width" property. Use "inline-size". (csstools/use-logical)"
      severity: error
      line: 13
      column: 3
      endLine: 13
      endColumn: 15
  ...
```

_No response_

Mouvedia:

Closes #8321

https://stackoverflow.com/q/19109912/248058

### DIFF
--- a/.changeset/sweet-mirrors-lie.md
+++ b/.changeset/sweet-mirrors-lie.md
@@ -1,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `tap` formatter not escaping `#` in its test point descriptions
--- a/.changeset/sweet-mirrors-lie.md
+++ b/.changeset/sweet-mirrors-lie.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `tap` formatter not escaping `#` and `/` in its test point descriptions
+Fixed: `tap` formatter not escaping `#` and `\` in its test point descriptions
--- a/.changeset/sweet-mirrors-lie.md
+++ b/.changeset/sweet-mirrors-lie.md
@@ -2,4 +2,4 @@
 "stylelint": patch
 ---
 
-Fixed: `tap` formatter not escaping `#` in its test point descriptions
+Fixed: `tap` formatter not escaping `#` and `/` in its test point descriptions
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -21,6 +21,12 @@ export default [
 		rules: {
 			'import/extensions': ['error', 'ignorePackages'],
 			'jest/no-standalone-expect': ['error', { additionalTestBlockFunctions: ['testFn'] }],
+			'n/no-extraneous-import': [
+				'error',
+				{
+					allowModules: ['js-yaml'],
+				},
+			],
 		},
 	},
 ];
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -21,12 +21,6 @@ export default [
 		rules: {
 			'import/extensions': ['error', 'ignorePackages'],
 			'jest/no-standalone-expect': ['error', { additionalTestBlockFunctions: ['testFn'] }],
-			'n/no-extraneous-import': [
-				'error',
-				{
-					allowModules: ['js-yaml'],
-				},
-			],
 		},
 	},
 ];
--- a/lib/formatters/__tests__/tapFormatter.test.mjs
+++ b/lib/formatters/__tests__/tapFormatter.test.mjs
@@ -92,7 +92,7 @@ ok 1 - path/to/file.css
 not ok 1 - path/to/file1.css
   ---
   baz:
-    - message: Unexpected bar
+    - message: "Unexpected bar"
       severity: error
       line: 2
       column: 1
@@ -102,20 +102,20 @@ not ok 1 - path/to/file1.css
 not ok 2 - path/to/file2.css
   ---
   bar:
-    - message: Unexpected foo
+    - message: "Unexpected foo"
       severity: error
       line: 1
       column: 1
       endLine: 2
       endColumn: 3
-    - message: Unexpected foo
+    - message: "Unexpected foo"
       severity: error
       line: 4
       column: 1
       endLine: 5
       endColumn: 3
   bar2:
-    - message: Unexpected foo 2
+    - message: "Unexpected foo 2"
       severity: error
       line: 10
       column: 1
@@ -196,7 +196,7 @@ not ok 2 - path/to/file2.css
 not ok 1 - path/to/file1.css
   ---
   baz:
-    - message: '"a":"1"'
+    - message: "\\"a\\":\\"1\\""
       severity: error
       line: 2
       column: 1
@@ -206,7 +206,7 @@ not ok 1 - path/to/file1.css
 not ok 2 - octothorp/%23/file2.css\\#qux
   ---
   bar:
-    - message: message
+    - message: "message"
       severity: error
       line: 1
       column: 1
@@ -216,7 +216,7 @@ not ok 2 - octothorp/%23/file2.css\\#qux
 not ok 3 - backslash/file3.css\\\\\\#qux
   ---
   bar:
-    - message: message
+    - message: "message"
       severity: error
       line: 1
       column: 1
@@ -226,7 +226,7 @@ not ok 3 - backslash/file3.css\\\\\\#qux
 not ok 4 - number/file4.css
   ---
   bar:
-    - message: '1'
+    - message: "1"
       severity: error
       line: 1
       column: 1
@@ -264,7 +264,7 @@ not ok 4 - number/file4.css
 not ok 1 - path/to/file.css
   ---
   bar:
-    - message: Unexpected foo
+    - message: "Unexpected foo"
       severity: error
       line: 1
       column: 1
@@ -304,7 +304,7 @@ not ok 1 - path/to/file.css
 not ok 1 - path/to/file.css
   ---
   bar-very-very-very-very-very-long:
-    - message: Unexpected very very very very very very very very very very very very very long foo
+    - message: "Unexpected very very very very very very very very very very very very very long foo"
       severity: error
       line: 1
       column: 1
@@ -361,12 +361,12 @@ ok 1 - ${results[0].source} # SKIP ignored
 not ok 1 - path/to/file.css
   ---
   parseError:
-    - message: Cannot parse selector (parseError)
+    - message: "Cannot parse selector (parseError)"
       severity: error
       line: 1
       column: 1
   unknown:
-    - message: Unexpected foo
+    - message: "Unexpected foo"
       severity: error
       line: 2
       column: 1
--- a/lib/formatters/__tests__/tapFormatter.test.mjs
+++ b/lib/formatters/__tests__/tapFormatter.test.mjs
@@ -92,7 +92,7 @@ ok 1 - path/to/file.css
 not ok 1 - path/to/file1.css
   ---
   baz:
-    - message: "Unexpected bar"
+    - message: Unexpected bar
       severity: error
       line: 2
       column: 1
@@ -102,20 +102,20 @@ not ok 1 - path/to/file1.css
 not ok 2 - path/to/file2.css
   ---
   bar:
-    - message: "Unexpected foo"
+    - message: Unexpected foo
       severity: error
       line: 1
       column: 1
       endLine: 2
       endColumn: 3
-    - message: "Unexpected foo"
+    - message: Unexpected foo
       severity: error
       line: 4
       column: 1
       endLine: 5
       endColumn: 3
   bar2:
-    - message: "Unexpected foo 2"
+    - message: Unexpected foo 2
       severity: error
       line: 10
       column: 1
@@ -138,12 +138,12 @@ not ok 2 - path/to/file2.css
 						endColumn: 3,
 						rule: 'baz',
 						severity: 'error',
-						text: '"foo"',
+						text: '"a":"1"',
 					},
 				],
 			},
 			{
-				source: 'path/to/file2.css#qux',
+				source: 'octothorp/%23/file2.css#qux',
 				errored: true,
 				warnings: [
 					{
@@ -157,26 +157,76 @@ not ok 2 - path/to/file2.css
 					},
 				],
 			},
+			{
+				source: 'backslash/file3.css\\#qux',
+				errored: true,
+				warnings: [
+					{
+						line: 1,
+						column: 1,
+						endLine: 2,
+						endColumn: 3,
+						rule: 'bar',
+						severity: 'error',
+						text: 'message',
+					},
+				],
+			},
+			{
+				source: 'number/file4.css',
+				errored: true,
+				warnings: [
+					{
+						line: 1,
+						column: 1,
+						endLine: 2,
+						endColumn: 3,
+						rule: 'bar',
+						severity: 'error',
+						text: '1',
+					},
+				],
+			},
 		];
 
 		const output = tapFormatter(results);
 
 		expect(output).toBe(`TAP version 14
-1..2
+1..4
 not ok 1 - path/to/file1.css
   ---
   baz:
-    - message: "\\"foo\\""
+    - message: '"a":"1"'
       severity: error
       line: 2
       column: 1
       endLine: 3
       endColumn: 3
   ...
-not ok 2 - path/to/file2.css%23qux
+not ok 2 - octothorp/%23/file2.css\\#qux
   ---
   bar:
-    - message: "message"
+    - message: message
+      severity: error
+      line: 1
+      column: 1
+      endLine: 2
+      endColumn: 3
+  ...
+not ok 3 - backslash/file3.css\\\\\\#qux
+  ---
+  bar:
+    - message: message
+      severity: error
+      line: 1
+      column: 1
+      endLine: 2
+      endColumn: 3
+  ...
+not ok 4 - number/file4.css
+  ---
+  bar:
+    - message: '1'
       severity: error
       line: 1
       column: 1
@@ -214,7 +264,7 @@ not ok 2 - path/to/file2.css%23qux
 not ok 1 - path/to/file.css
   ---
   bar:
-    - message: "Unexpected foo"
+    - message: Unexpected foo
       severity: error
       line: 1
       column: 1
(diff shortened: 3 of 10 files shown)


lib/formatters/tapFormatter.mjs
```diff
@@ -37,7 +37,7 @@ export default function tapFormatter(results) {
 
 				for (const { text, severity, line, column, endLine, endColumn } of warnings) {
 					lines.push(
-						`    - message: "${text}"`,
+						`    - message: "${text.replace(/"/g, '\\"')}"`,
```
Mouvedia, line 40: or

```suggestion
						`    - message: "${text.replaceAll('"', '\\"')}"`,
```

romainmenke:
> I don't think that https://github.com/stylelint/stylelint/security/code-scanning/71 is relevant.

I think we should consider this warning accurate.
A regex also isn't the proper way to do escaping :)

See: https://testanything.org/tap-version-14-specification.html

This should be done with a loop over the characters and adding backslashes where needed.

Mouvedia:
> This should be done with a loop over the characters and adding backslashes where needed.

@romainmenke just to clarify, you want to loop over the string itself instead of using `replaceAll`?
i.e.
```js
// this is an example
for (const index in str) {
  if (str[index] === '\\') { str.substring(0, index) + '\\"' + str.substring(index); }
}
```
instead of simply `str.replaceAll('"', '\\"')`?

Mouvedia:
OK I think what you wanted to convey is: there are other characters that need escaping, look at the specification.
i.e. `replaceAll` is not sufficient because we have more than one character to replace

Thanks, I missed the `#` character indeed.

Mouvedia:
After reviewing the [Escaping chapter](https://testanything.org/tap-version-14-specification.html#escaping) of the specification, `#` will have to be escaped for the `LintResult['source']`.

i.e.

> Sometimes a user may include a `#` character in a **Test Point description**, Plan comment, Bailout reason, or TODO/SKIP reason.

- the SKIP reason, if included, is always `ignored`
- no TODO
- no bailout
- no plan comment

---

@romainmenke we are escaping `"` for only one reason, we are surrounding the message with double quotes by default.
i.e. it has more to do with YAML than with the the TAP specification
I think that `replaceAll` is sufficient and appropriate for that purpose.

Ill add escaping for `source` and corresponding tests.

lib/formatters/tapFormatter.mjs
```diff
@@ -57,3 +58,22 @@ export default function tapFormatter(results) {
...
+ * with its escape sequence
+ * in the file path or unique ID of the result's source
+ * @param {LintResult} result
+ * @modifies {result}
+ * @see {@link https://testanything.org/tap-version-14-specification.html#escaping TAP 14 specification}
+ */
+function escapeOctothorp(result) {
+	if (!result.source) return;
+
+	result.source = result.source.replaceAll('#', '%23');
```
Mouvedia, line 73: note

We don't use `\#` because:

1. postcss doesn't use a [custom alphabet](https://github.com/ai/nanoid#custom-alphabet-or-size) for the id
https://github.com/postcss/postcss/blob/main/lib/input.js#L61
i.e. `#` shouldn't be part of the id generated
2. in consequence, `#` may only be part of a file path
i.e. `encodeURIComponent('#') === '%23'`

lib/formatters/preprocessWarnings.mjs
```diff
@@ -7,9 +7,9 @@ import { DEFAULT_SEVERITY, SEVERITY_ERROR } from '../constants.mjs';
 
 /**
  * Preprocess warnings in a given lint result.
- * Note that this function has a side-effect.
  *
  * @param {LintResult} result
+ * @modifies {result}
```
Mouvedia, line 12: see jsdoc/jsdoc#1709

Mouvedia:
> Part is that we don't implement escaping.
> To resolve this issue we need to implement this and this should be done correctly.

@romainmenke
This is done partially.
see [58bec57](https://github.com/stylelint/stylelint/pull/8334/commits/58bec57a22d370476099a61e9530056b0ff40efa)
The escaping of `\` can be done in a separate PR.
i.e. it's off topic for 8321
It will require UNC path file tests which are much more than what I signed for :)
The CodeQL alerts are fixed.

romainmenke:
> The CodeQL alerts are fixed.

How were they fixed? The current code doesn't seem meaningfully different 🤔 

-----

I can't review this properly.

My guidance is to not use regex for this case but to implement proper serialization and escaping for the specific encoding. If that is Yaml, then use a Yaml serializer, same for TAP.

It seems trivial to me to do this right once.

Other maintainers might have different opinions and that is fine :)

lib/formatters/preprocessWarnings.mjs
romainmenke: This seems unrelated.

lib/formatters/tapFormatter.mjs
```diff
@@ -1,15 +1,16 @@
...
 /** @typedef {Array<Omit<Warning, 'rule'>>} Warnings */
 
-/** @type {import('stylelint').Formatter} */
+/** @type {Formatter} */
 export default function tapFormatter(results) {
 	const lines = ['TAP version 14', `1..${results.length}`];
 
 	for (const [index, result] of results.entries()) {
 		preprocessWarnings(result);
+		escapeOctothorp(result);
```
romainmenke, line 13: I don't think it is correct to mutate the `result`.
We only want to escape the output from this formatter we don't want to modify the `result.source` in ways that could be observable.

lib/formatters/tapFormatter.mjs
romainmenke, line 73: This seems incorrect.
Why is url encoding used and not `\` as dictated by the TAP spec?

romainmenke (commented):
@Mouvedia I've gone over this change again and I am still concerned. I don't think it is correct and I also don't think it uses the correct techniques.

I think it is important that we try to fix issues as best as we can, not as quickly as we can :)

Let's take it back to the issue: https://github.com/stylelint/stylelint/issues/8321#issuecomment-2613924041

lib/formatters/tapFormatter.mjs
Mouvedia: Because as explained in https://github.com/stylelint/stylelint/pull/8334#discussion_r1923831061:

1. it will ultimately be part of a path that may be copy/pasted or clicked on (not an id)
2. `%23` is `#`
3.  I consider that having a superfluous step is not necessary
i.e. `#` > `\#` > `%23`
4. we don't want to `encodeURIComponent` everything

lib/formatters/tapFormatter.mjs
Mouvedia: `preprocessWarnings` does it too.
i.e. mutating result

I could return a copy but that would require to refactor `preprocessWarnings` and I didn't want to do that in this PR.

lib/formatters/preprocessWarnings.mjs
Mouvedia: see
1. https://github.com/stylelint/stylelint/pull/8334#discussion_r1929527807
2. https://github.com/stylelint/stylelint/pull/8334/files#diff-6c2b58f14ee9189aabe9078bedf28bc0b5e62756820c2bd84422c8fc87a1096dR67

lib/formatters/tapFormatter.mjs
romainmenke: We can avoid extra mutations without refactoring `preprocessWarnings`.

lib/formatters/tapFormatter.mjs
romainmenke: Do you have any supporting references that this is correct?
This is not according to the specification.

A TAP formatter is expected to be used with other tools that can interpret the TAP format.

Mouvedia: If you want we can add that step but I consider that it will require another step.
i.e. since it's a path we don't want to leave `\#` unencoded

romainmenke: I think we are trying to solve different problems.

Currently we don't follow the TAP format.
Encoding as I described, as the specification describes is needed to correctly implement a TAP reporter.

uri encoding `#` makes is less correct, not more so.

What I think you are trying to achieve is that you can click on a TAP report text and open the corresponding file. Even when that TAP report contains a file name with `#` in the `description` field. I don't think that is correct given how TAP dictates that descriptions must be escaped.

lib/formatters/tapFormatter.mjs
Mouvedia: Either result is not mutated at all or it can be.
It's an opinion.
i.e. I value consistency

Let's discuss this again, after Iv done the changes that you requested in https://github.com/stylelint/stylelint/issues/8321#issuecomment-2613924041.

lib/formatters/tapFormatter.mjs
Mouvedia: Let's take a real example to illustrate my point.

for `/home/lang/C#/foo/file.css` as the `source`
if we the formatter escapes `#` it now has the responsibility to output a valid URI
i.e. `/home/lang/C\#/foo/file.css` wouldn't be exploitable anymore
More over if it's an UNC file path which uses `\` instead of `/`.

What I mean is, I am not against escaping the characters, but if it introduces a regression it needs to be taken into account.
i.e. `\#` will have to be encoded back to `%23` if it's not a delimiter for the fragment identifier of the URI

So there's more to it than just a `replaceAll`, you were right about that.
Let's discuss this again once Ill make the changes that you have requested.

lib/formatters/tapFormatter.mjs
```diff
@@ -37,7 +39,7 @@ export default function tapFormatter(results) {
 
 				for (const { text, severity, line, column, endLine, endColumn } of warnings) {
 					lines.push(
-						`    - message: "${text}"`,
+						`    - message: ${yaml.dump(`${text}`, { quotingType: '"', forceQuotes: true, lineWidth: -1 }).trim()}`,
```
Mouvedia, line 42: note

`lineWidth` added to avoid the strip block chomping indicator (default is 80)

lib/formatters/tapFormatter.mjs
```diff
@@ -37,7 +39,7 @@ export default function tapFormatter(results) {
```
Mouvedia, line 39: todo

The Names chapter was removed in #6390.

> [made up of lowercase words separated by hyphens](https://github.com/stylelint/stylelint/pull/6390/files#diff-7a62ae7246fc278776a7129004bcc382e494619fcda082e35c94b8a858122680)

@jeddy3 we need to put this back in the documentation.
It has the advantage of discarding `linebreaks`, `indentation` and the `report*` options as valid rules.

lib/formatters/tapFormatter.mjs
Mouvedia, line 1: todo
line 38

The Names chapter was removed in #6390.

> [made up of lowercase words separated by hyphens](https://github.com/stylelint/stylelint/pull/6390/files#diff-7a62ae7246fc278776a7129004bcc382e494619fcda082e35c94b8a858122680)

@jeddy3 we need to put this back in the documentation.
It has the advantage of discarding `linebreaks`, `indentation` and the `report*` options as valid rules.

lib/formatters/tapFormatter.mjs
```diff
@@ -57,3 +59,29 @@ export default function tapFormatter(results) {
...
+ */
+function getEscapedSource(source) {
+	if (!source) return source;
+
+	let output = '';
+
+	for (const char of source) {
+		switch (char.codePointAt(0)) {
+			case 0x005c: // `\`
+			case 0x0023: // `#`
```
Mouvedia, line 76: note

This function assumes that PostCSS passes us an exploitable URI when the source isn't an id.
i.e. when the hash is not the delimiter of the fragment identifier it is properly encoded to %23

lib/formatters/tapFormatter.mjs
```diff
@@ -37,7 +39,7 @@ export default function tapFormatter(results) {
 
 				for (const { text, severity, line, column, endLine, endColumn } of warnings) {
 					lines.push(
-						`    - message: "${text}"`,
+						`    - message: ${yaml.dump(`${text}`, { quotingType: '"', forceQuotes: true, lineWidth: -1 }).trim()}`,
 						`      severity: ${severity}`,
```
Mouvedia, line 43: note

https://github.com/stylelint/stylelint/blob/32c9dded2b6d68ee5a435859d39b091b7f4b0a07/types/stylelint/index.d.ts#L64

lib/formatters/__tests__/tapFormatter.test.mjs
Mouvedia, line 1: note

Using [`dump`](https://github.com/nodeca/js-yaml#dump-object---options-) had the consequence of removed extraneous double quotes.

eslint.config.mjs
```diff
@@ -21,6 +21,12 @@ export default [
 		rules: {
 			'import/extensions': ['error', 'ignorePackages'],
 			'jest/no-standalone-expect': ['error', { additionalTestBlockFunctions: ['testFn'] }],
+			'n/no-extraneous-import': [
+				'error',
+				{
+					allowModules: ['js-yaml'],
```
Mouvedia, line 27: note

We have 6 dependencies that use `js-yaml`.

lib/formatters/tapFormatter.mjs
romainmenke: Can you open an issue about this?
I am unsure what you mean.

lib/formatters/__tests__/tapFormatter.test.mjs
romainmenke: Can you add `"js-yaml": "^4.1.0",` as direct dependency?
Not only is that needed anyway but `forceQuotes` only shipped in `4.1.0` and it seems that our indirect dependency is currently an earlier version.

lib/formatters/tapFormatter.mjs
```diff
@@ -17,6 +18,7 @@ export default function tapFormatter(results) {
 		const ignored = result.ignored ? ' # SKIP ignored' : '';
 		const error = result.errored ? 'not ' : '';
 
+		result.source = getEscapedSource(result.source);
 		lines.push(`${error}ok ${index + 1} - ${result.source}${ignored}`);
```
romainmenke, line 22:
```suggestion
		lines.push(`${error}ok ${index + 1} - ${getEscapedSource(result.source)}${ignored}`);
```

Let's avoid this mutation.
It can only be a source of bugs and confusion.
It can never be a good approach.

Anything that needs to use the value of `result.source` beyond this point will get a TAP escaped value.

That is only useful for TAP reporters, such as this formatter.

Anything else must get the original value.

lib/formatters/tapFormatter.mjs
romainmenke, line 42:
```suggestion
						`    - message: ${yaml.dump(text, { quotingType: '"', forceQuotes: true, lineWidth: -1 }).trim()}`,
```

romainmenke, line 42: I only think of this now but `yml` is a super set of `json`.

I think that `JSON.stringify` might actually also work.
This has the (small) benefit of not requiring us to change our list of direct dependencies.

`json` already always quotes strings and uses double quotes. It also doesn't do line wrapping.

romainmenke (commented):
Thank you for applying feedback @Mouvedia,
I've left a few suggestions.

lib/formatters/tapFormatter.mjs
Mouvedia: Actually you also have `String.prototype.quote`.
I would know since I created `@string/prototype.inspect` package.
Let's try `JSON.stringify`.

lib/formatters/tapFormatter.mjs
Mouvedia: That comment is directed to Richard.
In short I am asking something related to the rule names (ours and plugin ones) which would prove that we don't have to alter that line.

lib/formatters/tapFormatter.mjs
```diff
@@ -1,10 +1,10 @@
 import { isNumber } from '../utils/validateTypes.mjs';
 import preprocessWarnings from './preprocessWarnings.mjs';
 
-/** @import {Warning} from 'stylelint' */
+/** @import {Warning, Formatter, LintResult} from 'stylelint' */
```
romainmenke, line 4:
```suggestion
/** @import {Warning, Formatter} from 'stylelint' */
```

I didn't see `LintResult` used, so maybe can be omitted?
Or maybe I overlooked it?

romainmenke (commented):
One final note :)
Thank you for working on this @Mouvedia!

Mouvedia:
In retrospect I should have split the PR in two right away, once we uncovered the related TAP escaping bug.
Sorry for the long wait.

---

It surfaced that we don't document rule name guidelines anymore, that's a good thing.

.changeset/sweet-mirrors-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
+---
+
+Fixed: `tap` formatter not escaping `#` and `/` in its test point descriptions
```
Mouvedia, line 5:
```suggestion
Fixed: `tap` formatter not escaping `#` and `\` in its test point descriptions
```

PR #8340, opened 2025-01-20
Fix `declaration-property-value-keyword-no-deprecated` false negatives for `overflow-x`, `overflow-y` and warning message for `overflow`

Mouvedia:

#8178

1.
This should be followed up by adding a new enhancement issue to add support for autofixing shorthands.
e.g. `overflow: hidden overlay` to `overflow: hidden auto`

2.
This PR should be part of the same version as #8328.

3.
[demo](https://stylelint.io/demo/#N4Igxg9gJgpiBcICGACYAdAdugLhAbjAE4BmANhAO7woAWAllLJigcWUgJ4DcWAvlhAAaECXpkYAOSQBbOIhgAPWQAcJAOjABnLcPARMYgOYIQGFinQgiAVwlarNcyheWQsMByJIc9AwFoVIggVYhxOf3wkMhsYfwBrGE5KCCIof0wIf1ggmDAfGChHFBxbGCwXAUw+PUhDeiMAMVSZH1MAKy0DPRzdRGc3LXCJMnpMHGKrAEYANnUpgGZ1ACYrIQrB4ZhR8f86438hpEwoJDTJkA4cGCGrfhA+IA)


.changeset/few-needles-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"stylelint": patch
```
romainmenke, line 2:
```suggestion
"stylelint": minor
```

lib/rules/declaration-property-value-keyword-no-deprecated/__tests__/index.mjs
```diff
@@ -81,6 +90,30 @@ testRule({
...
+			endColumn: 29,
+		},
+		{
+			code: 'a { overflow: hidden /* comment */ overlay; }',
+			message: messages.rejected('overflow', 'overlay'),
+			line: 1,
+			column: 36,
+			endLine: 1,
+			endColumn: 43,
+		},
```
jeddy3, line 116: Let's move these test cases to the previous `testRule` block, as they are unrelated to the testing the secondary option. If they're unfixable, you can use [`unfixable: true`](https://github.com/stylelint/jest-preset-stylelint/blob/0b550662b23ad1449e5fdb7227c175cf6f46ce86/index.d.ts#L76).

Let's also open an issue (or do a quick follow-up PR) to add `fix: true` to this `testRule` block, as we typically do that in every block for rules that support autofix.

jeddy3 (commented):
Thanks for working on this. I've requested a small change to the tests.

.changeset/few-needles-call.md
Mouvedia: Wait that's the wrong one.

lib/rules/declaration-property-value-keyword-no-deprecated/__tests__/index.mjs
Mouvedia: > Let's also open an issue (or do a quick follow-up PR) to add fix: true to this testRule block, as we typically do that in every block for rules that support autofix.

Ill do that in the PR that will add support for `border-color` and `scrollbar-color`.
This PR needs to be merged first.

jeddy3 (approved):
LGTM, thank you.

PR #8346, opened 2025-01-21
Add `declaration-property-value-keyword-no-deprecated` autofix for shorthands

Mouvedia:
> Which issue, if any, is this issue related to?

follow-up of #8340 and #8328

> Is there anything in the PR that needs further explanation?

N/A

PR #8351, opened 2025-01-25
Fix `declaration-property-value-keyword-no-deprecated` false negatives for `text-decoration: blink`

Mouvedia:
> Which PR, if any, is this PR related to?

#8346

N/A


lib/rules/declaration-property-value-keyword-no-deprecated/index.mjs
```diff
@@ -91,6 +91,7 @@ const PROPERTY_NAME_TO_KEYWORDS = new Map([
 	['zoom', { reset: '1' }],
 	['border-color', DEPRECATED_COLORS],
 	['scrollbar-color', DEPRECATED_COLORS],
+	['text-decoration', ['blink']],
 	['text-decoration-line', ['blink']],
```
Mouvedia, line 95: I don't know which browser support `text-decoration-line` **and** does not disable `blink` effect.
It's probably one or the other but to err on the side of caution I won't remove line 95.

ryo-manba (approved):
LGTM, thanks!

jeddy3 (approved):

Issue #8358, opened 2025-01-27
Fix incompatible cache file format


larouxn:
### What minimal example or steps are needed to reproduce the bug?

1. Have an application on v16.12.0 of `stylelint` with cache enabled.
2. Run `stylelint` such as `stylelint ./app/ui/**/*.scss --allow-empty-input --cache --cache-location './tmp/cache/assets/stylelint/' --cache-strategy content`
3. Upgrade to v16.13.0 or later of `stylelint`.
4. Run `stylelint` command with cache flags again.
5. Observe the following error log.

```console
TypeError: Cannot read properties of undefined (reading 'length')
    at CacheableMemory.hashKey (file:///Users/larouxn/src/github.com/repo-redacted/node_modules/.pnpm/cacheable@1.8.8/node_modules/cacheable/dist/index.js:661:29)
    at CacheableMemory.getStore (file:///Users/larouxn/src/github.com/repo-redacted/node_modules/.pnpm/cacheable@1.8.8/node_modules/cacheable/dist/index.js:607:24)
    at CacheableMemory.set (file:///Users/larouxn/src/github.com/repo-redacted/node_modules/.pnpm/cacheable@1.8.8/node_modules/cacheable/dist/index.js:467:24)
    at FlatCache.loadFile (file:///Users/larouxn/src/github.com/repo-redacted/node_modules/.pnpm/flat-cache@6.1.5/node_modules/flat-cache/dist/index.js:153:21)
    at createFromFile (file:///Users/larouxn/src/github.com/repo-redacted/node_modules/.pnpm/flat-cache@6.1.5/node_modules/flat-cache/dist/index.js:371:9)
    at create (file:///Users/larouxn/src/github.com/repo-redacted/node_modules/.pnpm/file-entry-cache@10.0.5/node_modules/file-entry-cache/dist/index.js:24:30)
    at FileEntryDefault.createFromFile (file:///Users/larouxn/src/github.com/repo-redacted/node_modules/.pnpm/file-entry-cache@10.0.5/node_modules/file-entry-cache/dist/index.js:9:10)
    at new FileCache (file:///Users/larouxn/src/github.com/repo-redacted/node_modules/.pnpm/stylelint@16.13.0_typescript@5.7.3/node_modules/stylelint/lib/utils/FileCache.mjs:37:36)
[...]

PR #8359, opened 2025-01-27
Fix incompatible cache file format

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/8358

No, it's self-explanatory.

### DIFF
--- a/lib/utils/FileCache.mjs
+++ b/lib/utils/FileCache.mjs
@@ -1,5 +1,5 @@
+import { readFileSync, rmSync } from 'node:fs';
 import process from 'node:process';
-import { readFileSync } from 'node:fs';
 import { resolve } from 'node:path';
 
 import createDebug from 'debug';
@@ -34,7 +34,16 @@ export default class FileCache {
 		const useCheckSum = cacheStrategy === CACHE_STRATEGY_CONTENT;
 
 		debug(`Cache file is created at ${cacheFile}`);
-		this._fileCache = fileEntryCache.createFromFile(cacheFile, useCheckSum, undefined);
+
+		try {
+			this._fileCache = fileEntryCache.createFromFile(cacheFile, useCheckSum, undefined);
+		} catch {
+			debug(`Cache file might be corrupt, attempting to remove and recreate the cache file`);
+
+			rmSync(cacheFile);
+			this._fileCache = fileEntryCache.createFromFile(cacheFile, useCheckSum, undefined);
+		}
+
 		this._hashOfConfig = '';
 		this._useCheckSum = useCheckSum;
 	}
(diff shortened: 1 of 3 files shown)


lib/utils/FileCache.mjs
```diff
@@ -34,7 +34,16 @@ export default class FileCache {
...
 		debug(`Cache file is created at ${cacheFile}`);
-		this._fileCache = fileEntryCache.createFromFile(cacheFile, useCheckSum, undefined);
+
+		try {
+			this._fileCache = fileEntryCache.createFromFile(cacheFile, useCheckSum, undefined);
+		} catch {
+			debug(`Cache file might be corrupt, attempting to remove and recreate the cache file`);
+
+			rmSync(cacheFile);
+			this._fileCache = fileEntryCache.createFromFile(cacheFile, useCheckSum, undefined);
```
romainmenke, line 44: If the issue is unrelated to the cache file format this will throw again.

lib/utils/FileCache.mjs
```diff
@@ -34,7 +34,16 @@ export default class FileCache {
...
 
 		debug(`Cache file is created at ${cacheFile}`);
-		this._fileCache = fileEntryCache.createFromFile(cacheFile, useCheckSum, undefined);
+
+		try {
+			this._fileCache = fileEntryCache.createFromFile(cacheFile, useCheckSum, undefined);
+		} catch {
+			debug(`Cache file might be corrupt, attempting to remove and recreate the cache file`);
+
+			rmSync(cacheFile);
```
ybiquitous, line 43: [suggestion] Not confident, but does giving the `force` option make sense?

```suggestion
			rmSync(cacheFile, { force: true });
```

romainmenke: https://nodejs.org/api/fs.html#fsrmsyncpath-options
> When true, exceptions will be ignored if path does not exist. 

I think it is indeed good to add this, thank you for the suggestion.

Then we avoid throwing with a more confusing error when the initial throw was unrelated to an existing cache file and no cache file exists.

ybiquitous (approved):
LGTM, this PR is straightforward to review. Thank you 👍🏼

lib/utils/FileCache.mjs
```diff
@@ -34,7 +34,16 @@ export default class FileCache {
...
 
 		debug(`Cache file is created at ${cacheFile}`);
-		this._fileCache = fileEntryCache.createFromFile(cacheFile, useCheckSum, undefined);
+
+		try {
+			this._fileCache = fileEntryCache.createFromFile(cacheFile, useCheckSum, undefined);
+		} catch {
+			debug(`Cache file might be corrupt, attempting to remove and recreate the cache file`);
+
+			rmSync(cacheFile, { force: true });
```
Mouvedia, line 43: note

The `recursive` option is false by default and _if_ `cacheLocation` is a directory, Stylelint creates a cache file inside the specified folder.

Issue #8362, opened 2025-01-28
Add `*` support for computing `EditInfo`


romainmenke:
### What is the problem you're trying to solve?

See: https://github.com/stylelint/stylelint/issues/7192

We added support for computing `EditInfo` to Stylelint in https://github.com/stylelint/stylelint/pull/8234
Currently no rules yet support this feature.

### What solution would you like to see?

Rules should be updated with support:
- [ ] `alpha-value-notation`
- [ ] `at-rule-empty-line-before`
- [ ] `at-rule-no-deprecated`
- [ ] `at-rule-no-vendor-prefix`
- [ ] `color-function-notation`
- [ ] `color-hex-length`
- [ ] `comment-empty-line-before`
- [ ] `comment-whitespace-inside`
- [ ] `custom-property-empty-line-before`
- [ ] `declaration-block-no-duplicate-properties`
- [ ] `declaration-block-no-redundant-longhand-properties`
- [ ] `declaration-empty-line-before`
- [ ] `declaration-property-value-keyword-no-deprecated`
- [ ] `font-family-name-quotes`
- [ ] `font-weight-notation`
- [ ] `function-calc-no-unspaced-operator`
- [ ] `function-name-case`
- [ ] `function-url-quotes`
- [ ] `hue-degree-notation`
- [ ] `import-notation`
- [ ] `keyframe-selector-notation`
- [ ] `length-zero-no-unit`
- [ ] `lightness-notation`
- [ ] `media-feature-name-no-vendor-prefix`
- [ ] `media-feature-range-notation`
- [ ] `property-no-vendor-prefix`
- [ ] `rule-empty-line-before`
- [ ] `selector-attribute-quotes`
- [ ] `selector-no-vendor-prefix`
- [ ] `selector-not-notation`
- [ ] `selector-pseudo-element-colon-notation`
- [ ] `selector-type-case`
- [ ] `shorthand-property-no-redundant-values`
- [ ] `value-keyword-case`
- [ ] `value-no-vendor-prefix`

romainmenke:
If you have time, please consider contributing to this issue.
I've added a first change as an example but the list of rules to update is quite long :)

jeddy3:
A big thank you to @ryo-manba and @pamelalozano16 for adding `EditInfo` support to the rules. 

We can officially release the feature once the docs and tests are finalised:

- https://github.com/stylelint/stylelint/issues/8341
- https://github.com/stylelint/stylelint/issues/8342
[...]

PR #8363, opened 2025-01-28
Add `declaration-block-no-duplicate-properties` support for computing `EditInfo`

romainmenke:

See: https://github.com/stylelint/stylelint/issues/8362

This is the first PR in a series.

This rule didn't have any tests without `fix: true` in the test config.
I opted to only add a few new tests as the fixer itself has good coverage.

The needed change is simpler if a rule has existing tests without `fix: true`.
In such a case we could add `computeEditInfo: true` to the test config and add the `fix` expects.

### DIFF
--- a/lib/rules/declaration-block-no-duplicate-properties/__tests__/index.mjs
+++ b/lib/rules/declaration-block-no-duplicate-properties/__tests__/index.mjs
@@ -253,6 +253,66 @@ testRule({
 
 testRule({
 	ruleName,
+	config: [true],
+	computeEditInfo: true,
+
+	accept: [],
+
+	reject: [
+		{
+			code: 'a { color: pink; color: orange }',
+			message: messages.rejected('color'),
+			line: 1,
+			column: 5,
+			endLine: 1,
+			endColumn: 10,
+			fix: { range: [11, 24], text: '' },
+		},
+		{
+			code: 'a { COlOr: pink; coLOR: pink; color: pink }',
+			description: 'multiple duplicates - same',
+			warnings: [
+				{
+					message: messages.rejected('COlOr'),
+					line: 1,
+					column: 5,
+					endLine: 1,
+					endColumn: 10,
+					fix: { range: [4, 17], text: '' },
+				},
+				{
+					message: messages.rejected('coLOR'),
+					line: 1,
+					column: 18,
+					endLine: 1,
+					endColumn: 23,
+					fix: undefined,
+				},
+			],
+		},
+		{
+			code: 'a { color: pink; background: orange; color: orange }',
+			message: messages.rejected('color'),
+			line: 1,
+			column: 5,
+			endLine: 1,
+			endColumn: 10,
+			fix: { range: [4, 17], text: '' },
+		},
+		{
+			code: 'a { color: pink; background: orange; background: pink; }',
+			message: messages.rejected('background'),
+			line: 1,
+			column: 18,
+			endLine: 1,
+			endColumn: 28,
+			fix: { range: [29, 49], text: '' },
+		},
+	],
+});
+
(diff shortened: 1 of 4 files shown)


lib/rules/declaration-block-no-duplicate-properties/__tests__/index.mjs
```diff
@@ -251,6 +251,66 @@ testRule({
 	],
 });
 
+testRule({
```
Mouvedia, line 254: note

#8356

lib/rules/declaration-block-no-duplicate-properties/__tests__/index.mjs
```diff
@@ -251,6 +251,66 @@ testRule({
...
+			code: 'a { COlOr: pink; coLOR: pink; color: pink }',
+			description: 'multiple duplicates - same',
+			warnings: [
+				{
+					message: messages.rejected('COlOr'),
+					line: 1,
+					column: 5,
+					endLine: 1,
+					endColumn: 10,
+					fix: { range: [4, 17], text: '' },
```
Mouvedia, line 281: question

Since `17` is exclusive—i.e. `c` is not removed—the next warning shouldn't be considered as overlapping unless it started at `16`, right?

romainmenke: We reviewed the internals in detail in https://github.com/stylelint/stylelint/pull/8234

I will try again to explain this :)

--------

First off:

```js
var sourceCodeText = 'a { COlOr: pink; coLOR: pink; color: pink }';
var edit = { range: [4, 17], text: '' };

sourceCodeText.slice(0, edit.range[0]) + edit.text + sourceCodeText.slice(edit.range[1]);

// 'a { coLOR: pink; color: pink }'

sourceCodeText.slice(edit.range[0], edit.range[1]);

// 'COlOr: pink; '
```

That outcome is correct and matches the regular fixer.

-------

I think you are misunderstanding how the overlap calculation works.

We are not comparing final ranges.
We are not running each fixer and then calculating the most minimal range for each after which we compare final ranges.

In such a scenario there would be much fewer instances of overlap, but that is not what we are doing.

Instead we are comparing the final ranges of already computed edit info's with the full range of the mutated nodes.

In this case the mutated node is the `a {}` rule, not the `COlOr: pink;` declaration.
So any attempted fix after the first is considered overlapping.

Not because it is overlapping with the next edit info's, but instead because it is overlapping with the node that will be mutated.

This approach was key to getting this entire feature unblocked and why we no longer have to worry about https://github.com/stylelint/stylelint/issues/7836

Part of calculating the edit info is serializing the mutated node.
The serialized node can only contain a single mutation.
If it contains anything else the edit info values will be incorrect.
They might be incorrect in such a way that the ranges don't appear to overlap with anything else even if the values and text are mangled.

-----

That said, I think that a pure deletion of a single node should also be possible with using the node to be deleted as the mutated node.

Currently we need to use the parent node, but I am not sure if that is best.
I'll open an issue about that.

-----

Did this answer your question?

Mouvedia: > but instead because it is overlapping with the node that will be mutated

I totally forgot about that indeed.
That was way more than I was expecting for an answer but I think it might be useful for someone else that wasn't aware of what transpired in #8234.

---

> That said, I think that a pure deletion of a single node should also be possible with using the node to be deleted as the mutated node.
> I'll open an issue about that.

👍

jeddy3 (approved):
Amazing to see the first of these! ❤️ 

LGTM, thank you.

romainmenke:
Thank you for reviewing 🙇

Issue #8364, opened 2025-01-29
Refactor `utils.report` to not restore mutated nodes in `computeEditInfo`


romainmenke:
See: https://github.com/stylelint/stylelint/blob/19ee47d0a8afde625672a1fa6c76ac3d60015bd8/lib/utils/report.mjs#L272-L289

I added `node.replaceWith(fixedNodeClone);` because I thought it would be cleaner if the final AST matched the original.

But this isn't actually needed for any current feature.

I found that this detail makes it slightly less intuitive to use the `FixObject` API.

```ts
type FixObject = {
	apply?: FixCallback;
	node?: PostCSS.Node;
};
```

A removal of a node currently requires passing the parent instead of the node itself.
A replacement of a node also requires passing the parent.
This isn't obvious and we can't warn users when they get it wrong.

Current requirements:

| scenario | mutated node in `FixObject` |
| --- | ---- |
| edit | node itself |
| replace | parent node |
| remove | parent node |
| sort | parent node |
| add before/after | parent node |

The computed edit info will also be correct even when users pass the node itself instead of the parent.
Only restoring the AST will not work and we don't leverage that anyway.

I suggest that we remove that implementation detail entirely.
This will simplify what users should pass as `node` in `FixObject`:

| scenario | mutated node in `FixObject` |
| --- | ---- |
| edit | node itself |
| replace | node itself |
| remove | node itself |
| sort | parent node |
| add before/after | parent node |

A side effect of dropping this unused feature is that we avoid overlap in more cases and can collect more edit info's in a single run. Anytime the parent is used the range will obviously span a lot more surface. Using the node itself in more scenarios seems better. 

-----

If we do ever need a clean AST at the end of collecting warnings we can better clone the original before doing any work.

------

Thoughts?

romainmenke:
> In these cases, will you throw a custom message if the `parent` is absent?

No.

If `FixObject.node` is missing the edit info won't be computed.
It is equivalent to not providing `FixObject`.

It is up to rule authors to have correct implementations and to have sufficient test coverage.

Mouvedia:
> It is up to rule authors to have correct implementations…

Follow-up question then, if `fix.node` is not provided should we use the `node` passed by to `report` by default?
i.e. it may often be the same IIUC

romainmenke:
Definitely not :)
This will only lead to bugs.

[...]

PR #8367, opened 2025-01-29
Refactor `utils.report` to not restore mutated nodes in `computeEditInfo`

romainmenke:

Closes https://github.com/stylelint/stylelint/issues/8364

No, it's self-explanatory.

### DIFF
--- a/lib/utils/report.mjs
+++ b/lib/utils/report.mjs
@@ -270,9 +270,6 @@ function computeEditInfo({ fix, line, result: { stylelint }, ruleName }) {
 		return;
 	}
 
-	// Record the current state of the node that will be fixed
-	const fixedNodeClone = node.clone();
-
 	// Apply the fix
 	apply();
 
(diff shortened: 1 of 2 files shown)


lib/utils/report.mjs
```diff
@@ -282,13 +279,10 @@ function computeEditInfo({ fix, line, result: { stylelint }, ruleName }) {
 		text: node.toString(),
 	});
 
-	// Restore the previous state of the fixed node
-	node.replaceWith(fixedNodeClone);
-
 	// Mark the fixed range as mutated
 	rangesOfComputedEditInfos.push(fixData.range);
 
-	return { ...fixData };
```
romainmenke, line 291: I think this destructering was a left over from a previous iteration where there were more fields to be assignment.

lib/utils/report.mjs
Mouvedia, line 1: nit

You could replace line 247 by `* @returns {EditInfo | undefined}`

romainmenke: `EditInfo` is not currently exposed.
Maybe it should be, but seems out of scope.

ybiquitous (approved):
This change makes sense to me. 👍🏼

PR #8369, opened 2025-01-30
Add `alpha-value-notation` support for computing `EditInfo`

pamelalozano16:

See: https://github.com/stylelint/stylelint/issues/8362

I added some cases with multiple warnings in `fix: true` to make sure the semicolon after declarations wasn't being removed.


lib/utils/report.mjs
```diff
@@ -276,7 +276,7 @@ function computeEditInfo({ fix, line, result: { stylelint }, ruleName }) {
 	// Compute the smallest range and text of the fix
 	const fixData = narrowFixRange(node, {
 		range: fixedNodeRange,
-		text: node.toString(),
+		text: node.parent?.raws.semicolon ? `${node.toString()};` : node.toString(),
```
ybiquitous, line 279: [suggestion] It seems better to move this change to another PR so that we can easily track the commit history later.

For example, PR #8367 is dedicated to one refactoring.

Any thoughts? @romainmenke

ybiquitous (changes requested):
@pamelalozano16 Thank you for the pull request and adding many test cases for better coverage. 👍🏼 

I've left a few comments on the maintenance, can you have a look? Thanks.

ybiquitous:
@pamelalozano16 I updated the PR title to match #8363. In addition, can you add a `minor` changelog item like #8363?

https://github.com/stylelint/stylelint/blob/86d2c1708fc4d3ad6e6101ebb558a544d27bd9cd/.changeset/nine-needles-grow.md?plain=1#L1-L5

lib/utils/report.mjs
ybiquitous: Oh sorry, I missed #8370. We should merge it first. 😅

.changeset/nine-needles-grow.md
Mouvedia, line 1: guidance

Type `npx changeset` in your terminal to create a new changeset.
i.e. you shan't edit existing ones

ref: https://stylelint.io/maintainer-guide/pull-requests/#merging

lib/rules/alpha-value-notation/index.mjs
```diff
@@ -112,7 +112,10 @@ const rule = (primary, secondaryOptions) => {
 					endIndex,
 					result,
 					ruleName,
-					fix,
+					fix: {
+						apply: fix(),
```
Mouvedia, line 116: if you revert the change l102 or rename the function l102, I suggest either

```suggestion
						apply: fix,
```
or
```suggestion
						apply,
```

lib/rules/alpha-value-notation/__tests__/index.mjs
```diff
@@ -541,3 +581,218 @@ testRuleConfigs({
...
+
+	reject: [
+		{
+			code: 'a { opacity: 0.1; }',
+			message: messages.expected('0.1', '10%'),
+			line: 1,
+			column: 14,
+			endLine: 1,
+			endColumn: 17,
+			fix: { range: [13, 16], text: '10%' },
```
Mouvedia, line 717: off-topic
minor

@romainmenke I have noticed that we don't throw if this line is missing.
i.e. that would be expected since testRule throws if `fixed` is missing when `fix` is `true`

lib/rules/alpha-value-notation/__tests__/index.mjs
```diff
@@ -541,3 +581,218 @@ testRuleConfigs({
...
+					endColumn: 17,
+					fix: { range: [13, 16], text: '10%' },
+				},
+				{
+					message: messages.expected('0.5', '50%'),
+					line: 1,
+					column: 42,
+					endLine: 1,
+					endColumn: 45,
+					fix: undefined,
```
Mouvedia, line 784:
```suggestion
					fix: { range: [41, 44], text: '50%' },
```

see https://github.com/stylelint/stylelint/pull/8369/files#r1970638806 for details

Mouvedia (approved):
Apart for the missing fix object this LGTM.

lib/rules/alpha-value-notation/__tests__/index.mjs
romainmenke: Can you open an issue about this in the most relevant repository?
I don't want to cause confusion or excess noise in this thread.

Thank you

Mouvedia: This is minor, I don't feel an issue is necessary.
I just thought you might be interested since you were the one who updated `testRule` IIRC.

lib/rules/alpha-value-notation/__tests__/index.mjs
```diff
@@ -541,3 +581,218 @@ testRuleConfigs({
...
+testRule({
+	ruleName,
+	config: ['percentage'],
+	computeEditInfo: true,
+
+	accept: [
+		{
+			code: `@import "foo.css";`,
+		},
+	],
```
romainmenke, line 707:
```suggestion
	accept: [],
```

lib/rules/alpha-value-notation/__tests__/index.mjs
```diff
@@ -541,3 +581,218 @@ testRuleConfigs({
...
+testRule({
+	ruleName,
+	config: ['number'],
+	computeEditInfo: true,
+
+	accept: [
+		{
+			code: `@import "foo.css";`,
+		},
+	],
```
romainmenke, line 595:
```suggestion
	accept: [],
```

romainmenke (commented):
Two small suggestions to simplify the test cases.
I think those will also work fine.

lib/rules/alpha-value-notation/__tests__/index.mjs
romainmenke: I am always a very confused and feel blocked by such off topic comments where you do not want to open an issue to actually discuss the matter.

I don't understand why such comments are made on unrelated thread where they ping others and cause signal fatigue.

I personally pick between:
- something is on topic and it should be discussed here before proceeding with a change
- something is off topic and it shouldn't be mentioned at all, not even as an aside

And also:
- something is important and should be discussed. Then I will place a comment in the most appropriate place (possibly a new issue)
- something is unimportant and then I don't mention it at all.

If I were reviewing I might also be unsure if something is important or on topic and then I would phrase it as a question. But I wouldn't make a known to be off topic statement.

So I don't really understand making off topic comments and not wanting to open a dedicated issue. Because I don't understand I become very hesitant and don't really know how to address such comments.

I also try to place myself in the position of infrequent contributors who would be less confident to handle such off topic comments.

Can we try to move such remarks to issues?
I think this will work better and remove friction.
I also think it will lead to better answers and more shared knowledge.

Mouvedia: I will open an issue but I don't feel strongly about it.
It will prevent errors in the future so it might be worth considering.
We have already many issues open so I usually avoid opening nice-to-have issues.

Ill stop pinging you in PR comments from now on.

romainmenke (approved):
Thank you. LGTM 👍

